PR #8158, opened 2015-05-13
Removes Spy's ability to disguise as Attack Dog or Giant Ant.

deniz1a:
Fixes #7478. Also currently when spy disguises as giant ant, it has glitchy idle animation.

PR #8216, opened 2015-05-22
Fix Disguise oversight.

phrohdoh:


OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -115,7 +115,14 @@ void DisguiseAs(Actor self, Actor target)
 				var tooltip = target.TraitsImplementing<IToolTip>().FirstOrDefault();
 				AsTooltipInfo = tooltip.TooltipInfo;
 				AsPlayer = tooltip.Owner;
-				AsSprite = target.Trait<RenderSprites>().GetImage(target);
+
+				// Take the image of the target's disguise, if it exist.
+				// E.g., SpyA is disguised as a dog, SpyB then targets SpyA: We should use the dog image.
```
deniz1a: Spies can't disguise as dogs anymore: #8158

OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -115,7 +115,14 @@ void DisguiseAs(Actor self, Actor target)
 				var tooltip = target.TraitsImplementing<IToolTip>().FirstOrDefault();
 				AsTooltipInfo = tooltip.TooltipInfo;
 				AsPlayer = tooltip.Owner;
```
matija-hustic: This guy should also copy from targetDisguise.AsPlayer

OpenRA.Mods.RA/Traits/Disguise.cs
LipkeGu: This is not Dog related he delivers only an Example :)

deniz1a: Yeah but shouldn't the example be correct?

deniz1a:
I found the following issue: When spy A disguises as spy B of the same player which is in turn disguised as enemy rifle infantry, spy A gets own rifle infantry sprites instead of enemy's.

Phrohdoh:
I fixed this locally yesterday but haven't had the chance to push.

OpenRA.Mods.RA/Traits/Disguise.cs
Phrohdoh: That would be incorrect.

matija-hustic: Why? :) As far as I can see, that's what you did.
But only in one branch of the if-block, instead of removing this line and having each branch set the correct `AsPlayer`.

pchote: It would make sense to move this into the else branch for symmetry and to avoid double-assigning it.

Phrohdoh:
Fixed.

deniz1a:
I tested it, fixes the issue.

matija-hustic:
Works fine over here :+1: 
(Just that comment in the code with the dog example seems to rub some people the wrong way)

OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -112,10 +112,22 @@ void DisguiseAs(Actor self, Actor target)
 
 			if (target != null)
 			{
-				var tooltip = target.TraitsImplementing<IToolTip>().FirstOrDefault();
-				AsTooltipInfo = tooltip.TooltipInfo;
-				AsPlayer = tooltip.Owner;
-				AsSprite = target.Trait<RenderSprites>().GetImage(target);
+				// Take the image of the target's disguise, if it exist.
+				// E.g., SpyA is disguised as a dog. SpyB then targets SpyA. We should use the dog image.
```
matija-hustic: Thisn right here. Replace `dog` with something else. Like `engineer`

Phrohdoh:
The example isn't incorrect at all.
Just because that particular target isn't valid in the RA mod doesn't mean the example doesn't explain the code.

deniz1a:
Can spies disguise as dogs in other mods?

deniz1a:
Ok nevermind, example is not important. So Needs +2?

OpenRA.Mods.RA/Traits/Disguise.cs
cjshmyr: Probably, I don't get why it's a pain to update it to something that the mod can actually do.

OpenRA.Mods.RA/Traits/Disguise.cs
cjshmyr: `exists`

abcdefg30:
:+1: / :white_check_mark: 

Merging as this has already :+1:'s from @deniz1a and @matija-hustic.
I think the example is not a blocker and everyone will understand what is meant even though the described case wouldn't work out.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5949fbfedb366453ec98fa6feccd1ee1e7e76869...e7e39f3d904979012939a3d1b6e2d5e411ca771a)

PR #8220, opened 2015-05-23
Fix a crash in the ParaBomb weapon defintion

### LINKED ISSUE #8020: System.InvalidOperationException: Unit `bomblet` does not have a sequence named `open`
Happens when you use a parabomb support power (from crates) in Fort Lonestar.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12073835-system-invalidoperationexception-unit-bomblet-does-not-have-a-sequence-named-open?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #8219: Crash in survival02 when the timer runs out
```
Exception of type `System.InvalidOperationException`: Unit `bomblet` does not have a sequence named `open`
```

Also the bridges cannot be repaired anymore. Might not be mission-specific, haven't tested that yet.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/16973672-crash-in-survival02-when-the-timer-runs-out?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

abcdefg30:
Fixes #8219.
Fixes #8020.

### DIFF
--- a/mods/ra/maps/fort-lonestar/map.yaml
+++ b/mods/ra/maps/fort-lonestar/map.yaml
@@ -837,6 +837,7 @@ Weapons:
 		Report: CHUTE1.AUD
 		Projectile: GravityBomb
 			Image: BOMBLET
+			-OpenSequence:
 		Warhead: SpreadDamage
 			Spread: 426
 			Versus:
--- a/mods/ra/maps/survival02/map.yaml
+++ b/mods/ra/maps/survival02/map.yaml
@@ -1162,6 +1162,7 @@ Weapons:
 		Report: CHUTE1.AUD
 		Projectile: GravityBomb
 			Image: BOMBLET
+			-OpenSequence:
 		Warhead@1Dam: SpreadDamage
 			Spread: 150
 			Damage: 3500


obrakmann:
Confirmed that this fixed the crash on survival02. :+1:

reaperrr:
I think we should rather fix this properly in GravityBomb.cs, it shouldn't need yaml-level changes to avoid crashes.

obrakmann:
I don't know my way around those plumbing parts much. Do you mean to revert the null check introduced [here](../commit/6b7ffdc1b72ce935796d9abd1c84e8a9a94c29d3) and replace it with `anim.HasSequence()` again?

reaperrr:
> BTW, reading that code I'd expect it not to crash anyway. Nothing sets OpenSequence afaics, so why doesn't the check work?

I'm asking myself that, too.

That being said, using ```if (anim.HasSequence(info.OpenSequence))``` is probably safer anyway, because even if someone specifies an OpenSequence but forgets to define (or mistypes) the sequence itself in a sequence.yaml, it will simply not show up in-game rather than crash.

pchote:
Crashing with a useful exception is better than silently failing if someone mistypes the sequence.

penev92:
I'd actually rather get an exception that tells me my YAML is wrong, rather than it just now working for no obvious reason, which may result in me spending hours trying to figure out what is wrong, just because I forgot to define/didn't know I have to define/misspelled the definition of the sequence.

reaperrr:
Ok, got it.

The question why it fails still stands, something must be wrong with the null check in ```GravityBomb```.

penev92:
Because as I understand, it currently inherits "open", so the null check doesn't do anything. Adding `OpenSequence:` should override that "open" and set it to null, which should fix the problem. My guess is the current `-OpenSequence:` works because the FieldLoader just ignores the '-'.

reaperrr:
Tested, confirmed that this fixes the crash and that the '-' isn't needed.

:+1: once the '-'es have been removed.

pchote:
The `-` tells the yaml merger to remove the inherited nodes (which makes the game fall back to the default `null` value).  Removing it is equivalent to overriding `null` to `""`, which the code should also handle.

Mailaender:
Confirmed the fix at Fort Lonestar with parabomb crates. :white_check_mark: The `-` syntax is actually a lot clearer in this case. We just didn't know it can work in this case.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/4682a084fb831ba3824915aa3f3ac644887b7aa1...5cab437a66e67abbd11a53c8100fa0d5c250a500)

PR #8221, opened 2015-05-23
Introduce a class for efficient tiled sprite rendering.

pchote:
`TerrainSpriteLayer` wraps a vertex buffer and state management that can be used for sprite map layers (terrain, resources, shroud, etc).  It also removes the assumptions about static map bounds, which is a step towards supporting map-scripted bound changes (needed for TS/RA2) and cordon manipulation in the new editor.

This pr introduces the new class and hooks it up for the terrain renderer.  There shouldn't be any significant perf or gameplay changes.  The real benefits will come in the next few prs that will use this for the shroud, resources, etc.

Pinging @chrisforbes and @RoosterDragon who are the best suited to review this.

### DIFF
--- a/OpenRA.Game/Graphics/TerrainSpriteLayer.cs
+++ b/OpenRA.Game/Graphics/TerrainSpriteLayer.cs
@@ -1,0 +1,134 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.IO;
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Graphics
+{
+	public sealed class TerrainSpriteLayer : IDisposable
+	{
+		readonly Sprite emptySprite;
+
+		readonly IVertexBuffer<Vertex> vertexBuffer;
+		readonly Vertex[] vertices;
+		readonly HashSet<int> dirtyRows = new HashSet<int>();
+		readonly int rowStride;
+
+		readonly WorldRenderer worldRenderer;
+		readonly Map map;
+
+		readonly Sheet sheet;
+		readonly BlendMode blendMode;
+
+		float paletteIndex;
+
+		public TerrainSpriteLayer(World world, WorldRenderer wr, Sheet sheet, BlendMode blendMode, PaletteReference palette)
+		{
+			worldRenderer = wr;
+			this.sheet = sheet;
+			this.blendMode = blendMode;
+			paletteIndex = palette.TextureIndex;
+
+			map = world.Map;
+			rowStride = 4 * map.MapSize.X;
+
+			vertices = new Vertex[rowStride * map.MapSize.Y];
+			vertexBuffer = Game.Renderer.Device.CreateVertexBuffer(vertices.Length);
+			emptySprite = new Sprite(sheet, Rectangle.Empty, TextureChannel.Alpha);
+
+			wr.PaletteInvalidated += () =>
+			{
+				paletteIndex = palette.TextureIndex;
+
+				// Everything in the layer uses the same palette,
+				// so we can fix the indices in one pass
+				for (var i = 0; i < vertices.Length; i++)
+				{
+					var v = vertices[i];
+					vertices[i] = new Vertex(v.X, v.Y, v.Z, v.U, v.V, paletteIndex, v.C);
+				}
+
+				for (var row = 0; row < map.MapSize.Y; row++)
+					dirtyRows.Add(row);
+			};
+		}
+
+		public void Update(CPos cell, Sprite sprite)
+		{
+			var pos = worldRenderer.ScreenPosition(map.CenterOfCell(cell)) + sprite.Offset - 0.5f * sprite.Size;
+			Update(cell.ToMPos(map.TileShape), sprite, pos);
+		}
+
+		public void Update(MPos uv, Sprite sprite, float2 pos)
+		{
+			if (sprite != null)
+			{
+				if (sprite.Sheet != sheet)
+					throw new InvalidDataException("Attempted to add sprite from a different sheet");
+
+				if (sprite.BlendMode != blendMode)
+					throw new InvalidDataException("Attempted to add sprite with a different blend mode");
+			}
+			else
+				sprite = emptySprite;
+
+			var offset = rowStride * uv.V + 4 * uv.U;
+			Util.FastCreateQuad(vertices, pos, sprite, paletteIndex, offset, sprite.Size);
+
+			dirtyRows.Add(uv.V);
+		}
+
+		public void Draw(Viewport viewport)
+		{
+			var cells = viewport.VisibleCells;
+
+			// Only draw the rows that are visible.
+			// VisibleCells is clamped to the map, so additional checks are unnecessary
+			var firstRow = cells.TopLeft.ToMPos(map).V;
+			var lastRow = cells.BottomRight.ToMPos(map).V + 1;
+
+			Game.Renderer.Flush();
+
+			// Flush any visible changes to the GPU
+			for (var row = firstRow; row <= lastRow; row++)
+			{
+				if (!dirtyRows.Remove(row))
+					continue;
+
+				var rowOffset = rowStride * row;
+
+				unsafe
+				{
+					// The compiler / language spec won't let us calculate a pointer to
+					// an offset inside a generic array T[], and so we are forced to
+					// calculate the start-of-row pointer here to pass in to SetData.
+					fixed (Vertex* vPtr = &vertices[0])
+						vertexBuffer.SetData((IntPtr)(vPtr + rowOffset), rowOffset, rowStride);
+				}
+			}
+
+			Game.Renderer.WorldSpriteRenderer.DrawVertexBuffer(
+				vertexBuffer, rowStride * firstRow, rowStride * (lastRow - firstRow),
+				PrimitiveType.QuadList, sheet, blendMode);
+
+			Game.Renderer.Flush();
+		}
+
+		public void Dispose()
+		{
+			vertexBuffer.Dispose();
+		}
+	}
+}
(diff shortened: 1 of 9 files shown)


OpenRA.Game/Graphics/TerrainSpriteLayer.cs
```diff
@@ -0,0 +1,135 @@
...
+			// Only draw the rows that are visible.
+			// VisibleCells is clamped to the map, so additional checks are unnecessary
+			var firstRow = cells.TopLeft.ToMPos(map).V;
+			var lastRow = cells.BottomRight.ToMPos(map).V + 1;
+
+			Game.Renderer.Flush();
+
+			// Flush any visible changes to the GPU
+			var toUpdate = dirtyRows.Where(v => v >= firstRow && v <= lastRow)
+				.ToArray();
```
RoosterDragon: You can avoid the intermediate array here.

```c#
for (var v = firstRow; v <= lastRow; v++)
{
    if (!dirtyRows.Remove(v))
        continue;
    // Copy row to GPU.
}
```

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
```diff
@@ -0,0 +1,135 @@
...
+				}
+
+				dirtyRows.Remove(v);
+			}
+
+			Game.Renderer.WorldSpriteRenderer.DrawVertexBuffer(
+				vertexBuffer, rowStride * firstRow, rowStride * (lastRow - firstRow),
+				PrimitiveType.QuadList, sheet, blendMode);
+
+			Game.Renderer.Flush();
```
RoosterDragon: Is this second flush actually needed?

RoosterDragon:
Looks reasonable to me.

Once you start trying to attach the resource and shroud layers to it I'll have a lot more perf comments depending on how you go about it, but for the terrain layer everything is fine since it's fixed.

pchote:
We're going to need at least one more pr before then, adding proper dispose support for actors/traits.

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
pchote: Perhaps not, but we've always had it so I don't really want to risk breaking things by removing now.

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
pchote: Good idea.  Fixed.

OpenRA.Game/Graphics/Vertex.cs
```diff
@@ -15,24 +15,19 @@ namespace OpenRA.Graphics
...
-		#pragma warning restore
+		public readonly float X, Y, Z, U, V, P, C;
 
 		public Vertex(float2 xy, float u, float v, float p, float c)
-		{
-			this.x = xy.X; this.y = xy.Y; this.z = 0;
-			this.u = u; this.v = v;
-			this.p = p; this.c = c;
-		}
+			: this(xy.X, xy.Y, 0, u, v, p, c) { }
```
chrisforbes: This is getting a bit silly -- how many ctors with no interesting behavior do we need?

OpenRA.Game/Graphics/Vertex.cs
```diff
@@ -15,24 +15,19 @@ namespace OpenRA.Graphics
 	[StructLayout(LayoutKind.Sequential)]
 	public struct Vertex
 	{
-		// TODO Workaround for unused field warnings in mono 2.10
-		#pragma warning disable 414
-		float x, y, z, u, v;
-		float p, c;
-		#pragma warning restore
+		public readonly float X, Y, Z, U, V, P, C;
```
chrisforbes: We need to fix this vertex format at some point. 7 floats is not sane.

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
```diff
@@ -0,0 +1,135 @@
...
+
+				// Everything in the layer uses the same palette,
+				// so we can fix the indices in one pass
+				for (var i = 0; i < vertices.Length; i++)
+				{
+					var v = vertices[i];
+					vertices[i] = new Vertex(v.X, v.Y, v.Z, v.U, v.V, paletteIndex, v.C);
+				}
+
+				for (var v = 0; v < map.MapSize.Y; v++)
```
chrisforbes: Pick another letter for this. 'v' already has a different meaning above.

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
```diff
@@ -0,0 +1,135 @@
...
+				unsafe
+				{
+					// The compiler / language spec won't let us calculate a pointer to
+					// an offset inside a generic array T[], and so we are forced to
+					// calculate the start-of-row pointer here to pass in to SetData.
+					fixed (Vertex* vPtr = &vertices[0])
+						vertexBuffer.SetData((IntPtr)(vPtr + rowOffset), rowOffset, rowStride);
+				}
+
+				dirtyRows.Remove(v);
```
chrisforbes: You already did this above.

OpenRA.Renderer.Sdl2/VertexBuffer.cs
```diff
@@ -48,6 +48,16 @@ public void SetData(T[] data, int start, int length)
 			ErrorHandler.CheckGlError();
 		}
 
+		public void SetData(IntPtr data, int start, int length)
+		{
+			Bind();
```
chrisforbes: We should really have a lighter-weight bind that doesnt set up the vertex format.

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
```diff
@@ -0,0 +1,135 @@
...
+					continue;
+
+				var rowOffset = rowStride * v;
+
+				unsafe
+				{
+					// The compiler / language spec won't let us calculate a pointer to
+					// an offset inside a generic array T[], and so we are forced to
+					// calculate the start-of-row pointer here to pass in to SetData.
+					fixed (Vertex* vPtr = &vertices[0])
```
chrisforbes: Interesting. This works, but we're not winning anything from this genericity. See the innards of VertexBuffer.cs -- it knows about the exact layout of the vertex type, to set up the attribs!

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Drawing;
+using System.IO;
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Graphics
+{
+	public sealed class TerrainSpriteLayer : IDisposable
+	{
+		static readonly Sprite EmptySprite = new Sprite(new Sheet(new Size(1, 1)), Rectangle.Empty, TextureChannel.Alpha);
```
chrisforbes: This is a bit bogus. It's in its own weird sheet, so if you end up using it, you violate the one-sheet-only invariant. (benignly)

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
```diff
@@ -0,0 +1,135 @@
...
+
+			// Only draw the rows that are visible.
+			// VisibleCells is clamped to the map, so additional checks are unnecessary
+			var firstRow = cells.TopLeft.ToMPos(map).V;
+			var lastRow = cells.BottomRight.ToMPos(map).V + 1;
+
+			Game.Renderer.Flush();
+
+			// Flush any visible changes to the GPU
+			for (var v = firstRow; v <= lastRow; v++)
```
chrisforbes: Can we call this `row` ?

chrisforbes:
I added a bunch of notes for nitpicks.

The broad strokes are fine.

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
pchote: Indeed.  It looks like you originally set it up that way to avoid a cycling dependency between `Vertex` (in Game.exe) and `IVertexBuffer` implementations in the renderer dlls.

OpenRA.Game/Graphics/Vertex.cs
pchote: We'll likely need to grow an 8th float when somebody gets around to doing the TS lighting stuff.

OpenRA.Game/Graphics/Vertex.cs
pchote: I considered removing these, but we then need to add a bunch of intermediate variables and manually extract the components out of the float2 / float[3].  That makes the callers even uglier than they already are.

pchote:
Fixed nits in the `TerrainSpriteLayer` implementation.  The other comments aren't really actionable within the scope of this pr (unless you had any specific changes in mind @chrisforbes?).

OpenRA.Renderer.Sdl2/VertexBuffer.cs
WolfGaming: if we had one buffer for each renderer we would only need to set up the vertex format at initialization.

pchote:
Rebased. I count no objections (but also no real :+1:s) from both @RoosterDragon and @chrisforbes.

RoosterDragon:
No remaining objections here. :+1:

obrakmann:
No idea what's happening in here, but I didn't notice any regressions in-game, and with Rooster's and chrisf's review, I'll think I'll just merge it.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/13d6898a29a75787d45fd843f34320aebbb85a1e%5E...13d6898a29a75787d45fd843f34320aebbb85a1e)

OpenRA.Renderer.Sdl2/VertexBuffer.cs
chrisforbes: @WolfGaming you need separate VAOs too -- vertex format state is context state (<=2.1) or VAO state (>=3.0), not part of the buffer object.

chrisforbes: @pchote does your GL implementation expose ARB_vertex_array_object in a legacy context, or do we need to go core to have VAOs? I'd rather not use APPLE_vertex_array_object, its semantics are subtly bogus.

OpenRA.Game/Graphics/Vertex.cs
chrisforbes: I mean we should be packing the vertices better. floats are overkill. Half-float, unsigned byte, 10/10/10/2 are all suitable options. Big vertices carry a hefty memory bandwidth cost.

OpenRA.Renderer.Sdl2/VertexBuffer.cs
pchote: Need to go core.

Issue #6636, opened 2014-10-01
AI sends paradrops above water


ghost:
![sans titre](https://cloud.githubusercontent.com/assets/8930382/4483628/5a4746ee-49b3-11e4-87ad-32a0aa7cd9a0.png)
##

UberWaffe:
The easiest way to reduce the chance of this happening is by reducing https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/rules/ai.yaml#L82 from ```CheckRadius: 10c0``` to something smaller, say ```CheckRadius: 5c0```.

The best solution would be to define a new type of consideration for terraintype, that can assign an attractiveness to the type of terrain of the target cell. (I.e. then you could say for paratroopers Water has an attractiveness of -100, or some such.)

Mailaender:
Why don't we just exclude ships here? https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/rules/ai.yaml#L85 It seems to attract the AI in the case shown in the screenshot.

UberWaffe:
@Mailaender : That consideration has a negative attraction value, so it is actually attempting to avoid those.
Or at least it should, from just the screenshot it really does look as if it had not attempted to. From just the picture I cannot tell you what exactly went wrong.

Mailaender:
I think you are misunderstanding me. My suggestion is the following: https://github.com/OpenRA/OpenRA/pull/7922.

UberWaffe:
Ah, I see. Yes, that will help.

abcdefg30:
The AI sending paratroopers against naval units was resolved a while ago.

PR #8222, opened 2015-05-23
Fixed the AI not avoiding boats when searching for viable paradrop targets

Mailaender:
Tries to lessen #6636 with the adjusting screws we already got.


pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/a160b72c582756bee465e0629921f98dc7c971fe%5E...a160b72c582756bee465e0629921f98dc7c971fe

PR #7475, opened 2015-02-14
Convert the float health percentage to an int percentage

abcdefg30:
As wished by @pchote.

Should fix #7426 as well.

Issue #7426, opened 2015-02-03
Transforming with low hp causes invulnerability


abcdefg30:
See https://github.com/OpenRA/OpenRA/pull/7130#issuecomment-72467231.

Steps:
1. Set ```mcv``` [Health](https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/rules/vehicles.yaml#L298) to ```16```.
2. Set ```M1Carbine``` [chance](https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/weapons.yaml#L331) against ```light``` to ```100```.
3. Go ingame and build one mcv and one rifle man.
4. Shoot a single time on the mcv.
5. Deploy and undeploy it.

penev92:
Is it possible this might be linked to the ConYard issues we've been hearing about lately, @obrakman?

PR #8225, opened 2015-05-23
Fix broken bridges being initialized with a non-zero HP value

obrakmann:
Fixes the second half of #8219. For an explanation, see https://github.com/OpenRA/OpenRA/issues/8219#issuecomment-104926466

### DIFF
--- a/OpenRA.Game/Traits/Health.cs
+++ b/OpenRA.Game/Traits/Health.cs
@@ -44,8 +44,6 @@ namespace OpenRA.Traits
 			MaxHP = info.HP;
 
 			hp = init.Contains<HealthInit>() ? init.Get<HealthInit, int>() * MaxHP / 100 : MaxHP;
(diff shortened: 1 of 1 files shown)


OpenRA.Game/Traits/Health.cs
```diff
@@ -44,7 +44,7 @@ public Health(ActorInitializer init, HealthInfo info)
 			MaxHP = info.HP;
 
 			hp = init.Contains<HealthInit>() ? init.Get<HealthInit, int>() * MaxHP / 100 : MaxHP;
-			if (hp <= 0)
+			if (hp < 0)
```
pchote: This was the most important part of #7475, and changing it will reintroduce #7426.

obrakmann:
Eh, damn it. This solution won't work either. I've got nothing then.

PR #8226, opened 2015-05-23
Dispose traits when destroying an actor.

pchote:
This addresses some long-standing bogosity with our world/actor teardown logic.  Traits that implement `IDisposable` should now be cleaned up whenever their actor is permanently removed from the world.  Several traits that were previously forced to hook only death have been updated to use disposing instead.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
+++ b/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
@@ -48,7 +48,7 @@ namespace OpenRA.Mods.Common.Traits
 		bool IOccupySpaceInfo.SharesCell { get { return false; } }
 	}
 
-	public class Aircraft : IFacing, IPositionable, ISync, INotifyKilled, IIssueOrder, IOrderVoice, INotifyAddedToWorld, INotifyRemovedFromWorld
+	public class Aircraft : IFacing, IPositionable, ISync, IIssueOrder, IOrderVoice, INotifyAddedToWorld, INotifyRemovedFromWorld, INotifyActorDisposing
 	{
 		static readonly Pair<CPos, SubCell>[] NoCells = { };
 
@@ -157,7 +157,7 @@ namespace OpenRA.Mods.Common.Traits
 			}
 		}
 
-		public void Killed(Actor self, AttackInfo e)
+		public void Disposing(Actor self)
 		{
 			UnReserve();
 		}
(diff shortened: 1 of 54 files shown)


pchote:
Note: I expect this will trigger a new wave of "trait from destroyed object" crashes in cases that were always bogus, but previously hidden.

RoosterDragon:
If we're going to piggyback onto the Dispose mechanism, we should also do the following:
- Implement the pattern correctly, the easiest solution here is probably to mark each of the classes that have become disposable as ```sealed```.
- Ensure it is safe to call Dispose more than once without bad things happening.

pchote:
Reworked this a bit.  It now uses an `INotifyActorDisposing` interface to be more consistent with our other notifications and to make it clearer that we are disposing the actor, and not just the trait.

pchote:
Rebased.  This and #8221 together block the next stage of TS development.

RoosterDragon:
:+1:

LipkeGu:
an Year ago we fixed alot of crashes (related to hidden traits) ... this PR is welcome... :)

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -157,7 +157,7 @@ public void UnReserve()
 			}
 		}
 
-		public void Killed(Actor self, AttackInfo e)
+		public void Disposing(Actor self)
 		{
 			UnReserve();
```
obrakmann: Isn't disposing meant for cleaning up held references, memory and things like that? If so, I wouldn't expect gameplay logic in a `Disposing` method. In other words, what's wrong with keeping this in the `Killed` method?

(I just picked this line as an example, same thing applies to others.)

pchote: The trait isn't being disposed - the actor that owns it is.

The problem with keeping this in `Killed` is that the reservation isn't removed if the actor is removed from the world without being explicitly killed (e.g. by selling it or using Lua).

obrakmann:
Code makes sense, and I saw no regressions in a medium-length AI game. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f1907518618ee755a96407ba647d4bcb8f44b378%5E...f1907518618ee755a96407ba647d4bcb8f44b378)

PR #8227, opened 2015-05-24
Fix broken bridges being initialized with a non-zero HP value

obrakmann:
Fixes the second half of #8219. For an explanation, see https://github.com/OpenRA/OpenRA/issues/8219#issuecomment-104926466

As a side effect this also replaces invalid initial health values with 1 HP instead of 1% of MaxHP.


Mailaender:
I can confirm that this makes bridges repairable again in Survival 02. :+1:

penev92:
Well the bridge got repaired. Let's hope this is the only thing affected :+1:

PR #8228, opened 2015-05-24
Added render trait documentation

Mailaender:
I wrote some short descriptions to our various render https://github.com/OpenRA/OpenRA/wiki/Traits. Turned out what they do is not obvious to new modders.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithMakeAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithMakeAnimation.cs
@@ -17,6 +17,7 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
+	[Desc("Replaces the sprite during construction.")]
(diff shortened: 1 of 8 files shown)


penev92:
Nice. :+1: 
Sounds like some of these traits really need to be generalized/trashed.

OpenRA.Mods.Common/Traits/Render/WithMakeAnimation.cs
```diff
@@ -17,6 +17,7 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
+	[Desc("Replaces the sprite during build up.")]
```
obrakmann: "build up" -> "construction".  The german translation for "build up" would be "anhäufen", "ansammeln", so doesn't quite fit.

obrakmann:
Apart from that one niggle, :+1:

Mailaender:
Fixed.

PR #8229, opened 2015-05-24
Added a NuGet replacement script as fallback

### LINKED ISSUE #7386: Is the NuGet requirement justified?
In Gentoo, =games-strategy/openra-20150118 requires dev-dotnet/nuget
What is this new addition which justifies having your users install a new package manager (!) which is not in portage, and how does one build without it?
The ebuild should have a USE flag to disable the requirement for NuGet.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8031786-is-the-nuget-requirement-justified?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/7386.

Added a simple (no dependency resolution) `noget.sh` based upon https://gist.github.com/Phrohdoh/a4aae3982824e87fefb6.

NuGet on Unix is still a problem. It is either not packaged or broken: the certificate storage is fiddly to get right and the error messages that produces are confusing. Also somehow it is not behaving properly. For unexplained reasons the cloud servers often run into trouble breaking the builds so I enforce the replacement now until https://github.com/travis-ci/travis-ci/issues/3940 is resolved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
 # Check source code with StyleCop
 # call OpenRA to check for YAML errors
 script:
+ - sudo apt-get remove nuget # https://github.com/travis-ci/travis-ci/issues/3940
  - travis_retry make all-dependencies
  - make all
  - make check
--- a/thirdparty/fetch-thirdparty-deps-windows.sh
+++ b/thirdparty/fetch-thirdparty-deps-windows.sh
@@ -8,30 +8,39 @@ download_dir="${0%/*}/download/windows"
 mkdir -p "${download_dir}"
 cd "${download_dir}"
 
+function get()
+{
+	if which nuget >/dev/null; then
+		nuget install $1 -Version $2 -ExcludeVersion
+	else
+		../../noget.sh $1 $2
+	fi
+}
+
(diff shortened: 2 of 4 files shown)


pchote:
Works under OSX (after commenting out the actual nuget case). :+1:

obrakmann:
works on linux, too :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/574ebc2fa92a8d3226350bb2c01eafef68148147%5E...574ebc2fa92a8d3226350bb2c01eafef68148147)

Phrohdoh:
Nice job cutting down the cruft. :+1:

Mailaender:
@svenstaro may be happy about that.

svenstaro:
NICE

PR #8230, opened 2015-05-24
Made projectiles accept empty to disable image or sequence

reaperrr:
In addition to accepting null.

I agree that syntax-wise negating a value with `-` looks cleaner, but considering even some developers (including myself) were not aware that this is even possible, I think it would be better if both work, since quite a few parts of the engine (and official mod yaml code) allow empty strings to disable inherited properties.

Follow-up on #8220.


Mailaender:
:+1: This is more robust in any case.

PR #8149, opened 2015-05-11
Automatized GeoIP database download

Mailaender:
I always forget to update this one before releases and it is a binary that should not be stored here. According to http://dev.maxmind.com/geoip/geoip2/geolite2/ it is updated every first Tuesday of the month. I need some help with PowerShell to update the AppVeyor cache. The rule can probably be simplified to just "every Tuesday" or maybe even better some other day of the month to not stress the Maxmind servers when everybody runs his update. It does not need to be bleeding edge, but still update every once in a while.

PR #8236, opened 2015-05-24
Fixed a GeoLite2-Country DB filename in Windows fetch scripts.

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/8149.


obrakmann:
:+1: , but let's wait for appveyor.

PR #8237, opened 2015-05-24
Fixed make dependencies not copying the cached GeoLite2-Country.mmdb.gz into the root.

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/8236.


Mailaender:
In fact the same is required on Linux, too.

obrakmann:
:+1:

Mailaender:
Also sorry for this oversight. I only realized that I forgot about caching when updating https://build.opensuse.org/package/show/games:openra/playtest

PR #8245, opened 2015-05-25
Added Coverity static code analysis

Mailaender:
See https://scan.coverity.com/travis_ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,18 @@ cache:
 # Environment variables
 env:
   secure: "C0+Hlfa0YGErxUuWV00Tj6p45otC/D3YwYFuLpi2mj1rDFn/4dgh5WRngjvdDBVbXJ3duaZ78jPHWm1jr7vn2jqj9yETsCIK9psWd38ep/FEBM0SDr6MUD89OuXk/YyvxJAE+UXF6bXg7giey09g/CwBigjMW7ynET3wNAWPHPs="
+  global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,6 +14,18 @@ cache:
...
+   - secure: "VRSP4PUJFfBt/5goL87B7BjCbq9Bw5GwCAHaVUdLzjGeeWAuziXeQLDd9etV2920fnSFW6vAfv96BOChgtZgYdb5vurmsl+QfrsvDBCAj9g2dT7TPJRewolBKHF9yopUFXIgDIs20WiLIcx9lIIC7cGkwwyRQCrQxVunx60kCJw="
+
+addons:
+  coverity_scan:
+    project:
+      name: "OpenRA/OpenRA"
+      description: "An open-source implementation of the Command & Conquer: Red Alert engine using .NET/Mono and OpenGL."
+    notification_email: coverity@openra.net
+    build_command_prepend: "make dependencies"
+    build_command:   "make all"
```
pchote: We need to do additional configuration to tell their service about how our build works.  I haven't worked out just what that configuration is yet, though.

pchote:
Coverity only support C# analysis using the Microsoft C# compiler, so there is no way to make this work from travis (should work from appveyor, though).

Mailaender:
https://scan.coverity.com/faq#frequency is also a good reason against setting up automation first.

PR #8247, opened 2015-05-25
Added a UNIX man page

Mailaender:
`rpmlint` kept complaining and we got so many command line options scattered through different places so I decided to write one. First time I did this. It wasn't that hard. My main focus was server admins here. You can preview this locally using `man ./packaging/linux/openra.6`. Also did a test run at https://build.opensuse.org/package/show/games:openra/development

### DIFF
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -40,7 +40,6 @@ namespace OpenRA
 		public bool Dedicated = false;
 		public bool DedicatedLoop = true;
 		public bool LockBots = false;
-		public bool AllowVersionMismatch = false;
 		public string TimestampFormat = "HH:mm";
 
 		public ServerSettings() { }
@@ -64,7 +63,6 @@ namespace OpenRA
 			Dedicated = other.Dedicated;
 			DedicatedLoop = other.DedicatedLoop;
 			LockBots = other.LockBots;
-			AllowVersionMismatch = other.AllowVersionMismatch;
 		}
 	}
 
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -23,23 +23,55 @@ namespace OpenRA
 
 	public class ServerSettings
 	{
+		[Desc("Sets the server name.")]
 		public string Name = "OpenRA Game";
+
+		[Desc("Sets the internal port.")]
 		public int ListenPort = 1234;
+
+		[Desc("Sets the port advertised to the master server.")]
 		public int ExternalPort = 1234;
+
+		[Desc("Reports the game to the master server list.")]
 		public bool AdvertiseOnline = true;
+
+		[Desc("Locks the game with a password.")]
 		public string Password = "";
+
 		public string MasterServer = "http://master.openra.net/";
-		public bool DiscoverNatDevices = false; // Allow users to disable NAT discovery if problems occur
-		public bool AllowPortForward = true; // let the user disable it even if compatible devices are found
+
+		[Desc("Allow users to enable NAT discovery for external IP detection and automatic port forwarding.")]
+		public bool DiscoverNatDevices = false;
+
+		[Desc("Set this to false to disable UPnP even if compatible devices are found.")]
+		public bool AllowPortForward = true;
+
 		public bool NatDeviceAvailable = false; // internal check if discovery succeeded
-		public int NatDiscoveryTimeout = 1000; // ms to search for UPnP enabled NATs
-		public bool VerboseNatDiscovery = false; // print very detailed logs for debugging
+
+		[Desc("Time in miliseconds to search for UPnP enabled NAT devices.")]
+		public int NatDiscoveryTimeout = 1000;
+
+		[Desc("Print very detailed logs for debugging issues with routers.")]
+		public bool VerboseNatDiscovery = false;
+
+		[Desc("Starts the game with a default map. Input as hash that can be obtained by the utility.")]
 		public string Map = null;
+
+		[Desc("Takes a comma separated list of IP addresses that are not allowed to join.")]
 		public string[] Ban = { };
+
+		[Desc("Value in miliseconds when to terminate the game. Needs to be at least 10000 (10 s) to enable the timer.")]
 		public int TimeOut = 0;
+
+		[Desc("Run in headless mode with an empty renderer and without sound output.")]
 		public bool Dedicated = false;
+
+		[Desc("Automatically restart when a game ends. Disable this when something else already takes care about it.")]
 		public bool DedicatedLoop = true;
+
+		[Desc("Disallow AI bots.")]
 		public bool LockBots = false;
+
 		public string TimestampFormat = "HH:mm";
 
 		public ServerSettings() { }
@@ -122,6 +154,7 @@ namespace OpenRA
 
 	public class GameSettings
 	{
+		[Desc("Load a specific mod on startup. Shipped ones include: ra, cnc and d2k")]
 		public string Mod = "modchooser";
 		public string PreviousMod = "ra";
 
--- a/OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
+++ b/OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
@@ -20,19 +20,7 @@ namespace OpenRA.Mods.Common.LoadScreens
 {
 	public class BlankLoadScreen : ILoadScreen
 	{
-		public class LaunchArgs
-		{
-			[Desc("Connect to the following server given as IP:PORT on startup.")]
-			public string Connect;
-
-			[Desc("Connect to the unified resource identifier openra://IP:PORT on startup.")]
-			public string URI;
-
-			[Desc("Automatically start playing the given replay file.")]
-			public string Replay;
-		}
-
-		public LaunchArgs Launch = new LaunchArgs();
+		public LaunchArguments Launch;
 
 		public virtual void Init(Manifest m, Dictionary<string, string> info) { }
 
@@ -48,10 +36,7 @@ namespace OpenRA.Mods.Common.LoadScreens
 
 		public void StartGame(Arguments args)
 		{
-			foreach (var f in Launch.GetType().GetFields())
-				if (args.Contains("Launch" + "." + f.Name))
-					FieldLoader.LoadField(Launch, f.Name, args.GetValue("Launch" + "." + f.Name, ""));
-
+			Launch = new LaunchArguments(args);
 			Ui.ResetAll();
 			Game.Settings.Save();
 
@@ -80,14 +65,7 @@ namespace OpenRA.Mods.Common.LoadScreens
 			}
 
 			// Join a server directly
-			var connect = string.Empty;
-
-			if (!string.IsNullOrEmpty(Launch.Connect))
-				connect = Launch.Connect;
-
-			if (!string.IsNullOrEmpty(Launch.URI))
-				connect = Launch.URI.Replace("openra://", "").TrimEnd('/');
-
+			var connect = Launch.GetConnectAddress();
 			if (!string.IsNullOrEmpty(connect))
 			{
 				var parts = connect.Split(':');
--- a/OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
+++ b/OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
@@ -20,6 +20,20 @@ namespace OpenRA.Mods.Common.LoadScreens
 {
 	public class BlankLoadScreen : ILoadScreen
 	{
+		public class LaunchArgs
+		{
+			[Desc("Connect to the following server given as IP:PORT on startup.")]
+			public string Connect;
+
+			[Desc("Connect to the unified resource identifier openra://IP:PORT on startup.")]
+			public string URI;
+
+			[Desc("Automatically start playing the given replay file.")]
+			public string Replay;
+		}
+
+		public LaunchArgs Launch = new LaunchArgs();
+
 		public virtual void Init(Manifest m, Dictionary<string, string> info) { }
 
 		public virtual void Display()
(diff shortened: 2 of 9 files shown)


packaging/linux/openra.6
```diff
@@ -0,0 +1,68 @@
+.TH OPENRA 6
+.SH NAME
+openra \- An open-source implementation of the Command & Conquer: Red Alert engine using .NET/Mono and OpenGL.
```
pchote: OpenRA isn't an implementation of the C&C:RA engine.  This is just asking for trouble from those who are angry that OpenRA isn't a drop-in bug compatible reimplementation.

Mailaender: That is a copy & paste from https://github.com/OpenRA/OpenRA...

pchote: That description was pretty much correct when it was originally written back in 2010 :)

I suggest changing this to "An open-source modernization of the early 2D Command & Conquer games." (people running the games don't need more info, and updating the repo description to the more complete "The OpenRA project recreates and modernizes the early 2D Command & Conquer games on top of a modular open source RTS game engine.  It uses .NET/Mono, SDL2, and OpenGL, and runs on Windows, Mac OS X, and Linux."

packaging/linux/openra.6
```diff
@@ -0,0 +1,68 @@
...
+.SH SYNOPSIS
+.B openra
+[\fB\Game.Mod=\fR\fImodchooser\fR]
+.SH DESCRIPTION
+.B openra
+starts the game.
+.SH OPTIONS
```
pchote: These will inevitably bitrot as people adjust the settings without realizing that they are manually documented here.

Could we please move these descriptions to `[Desc()]`s in the code, and then generate the man page using a simple utility command?

Phrohdoh:
No reason this shouldn't be packaged for OS X also.

pchote:
Man pages are only used for commandline executables.  Our .app shouldn't package one.

packaging/linux/openra.6
Mailaender: I suggest "classic Westwood real-time strategy games like C&C Red Alert, Tiberian Dawn and Dune 2000" to not exclude Dune 2000, explain what RA stands for and avoid the Command & Conquer trademark.

packaging/linux/openra.6
Mailaender: I thought about it already and rejected the idea. Most of the https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Settings.cs are not very interesting and having them all exposed will add too much bloat. Also these settings are pretty stable. It is way quicker to do it by hand now. This can be done later in another pull request.

pchote: The bloat can be solved by only extracting settings that have descriptions.  We can also mirror the output to a wiki page, which will address a long-standing documentation request.

As for stability, we've been actively planning on changing the way that servers work (and thereby breaking all of the Server.* options included here) over the last couple of weeks.  If we can't keep this in sync automatically, then could we delay this until after that change has been done?

Mailaender:
Now moved to man page auto-generation as requested.

OpenRA.Game/Settings.cs
```diff
@@ -20,23 +20,61 @@ namespace OpenRA
...
+		[Desc("Takes a comma separated list of IP addresses that are not allowed to join.")]
 		public string[] Ban = { };
 		public int TimeOut = 0;
+
+		[Desc("Run in headless mode.")]
 		public bool Dedicated = false;
+
+		[Desc("Automatically restart when a game ends.")]
 		public bool DedicatedLoop = true;
 		public bool LockBots = false;
```
obrakmann: Why no documentation for `LockBots`?

OpenRA.Game/Settings.cs
```diff
@@ -20,23 +20,61 @@ namespace OpenRA
...
+
 		public bool NatDeviceAvailable = false; // internal check if discovery succeeded
+
 		public int NatDiscoveryTimeout = 1000; // ms to search for UPnP enabled NATs
-		public bool VerboseNatDiscovery = false; // print very detailed logs for debugging
+
+		[Desc("Print very detailed logs for debugging port forwarding in routers.")]
+		public bool VerboseNatDiscovery = false;
+
+		[Desc("Starts the game with a default map. Input as hash.")]
```
obrakmann: Maybe add a sentence saying how to get the hash from the utility.

obrakmann:
Looks good otherwise. :+1: with those two nits fixed.

OpenRA.Game/Settings.cs
```diff
@@ -20,23 +20,61 @@ namespace OpenRA
 {
 	public enum MouseScrollType { Disabled, Standard, Inverted }
 
+	public class LaunchSettings
```
pchote: Please, no.  These are configured as launch-only flags instead of settings because we should not be persisting them across launches!

Mailaender: We aren't. I immediately empty them.

pchote: Sorry that my initial advice didn't account for these launch-only args, but hacking them into the Settings just so that they get auto-generated man page entries is really not very nice.

Mailaender: My idea was that keeping track of them in a central place outweighs the hackiness of it.

pchote: IMO it doesn't.  You could define these somewhere else in the Game namespace, but it shouldn't be here.

Mailaender:
Updated as requested.

OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
```diff
@@ -61,24 +61,21 @@ public void StartGame(Arguments args)
...
-					if (connect != null)
-					{
-						connect = connect.Replace("openra://", "");
-						connect = connect.TrimEnd('/');
-					}
-				}
+			if (!string.IsNullOrEmpty(Game.Launch.URI))
+			{
+				connect = Game.Launch.URI.Replace("openra://", "").TrimEnd('/');
+				Game.Launch.URI = null;
```
pchote: My main issue from my earlier comments were that you were persisting these values as static state, and then manually clearing them.  This hasn't really changed.

OpenRA.Game/Game.cs
```diff
@@ -190,6 +191,10 @@ public static void InitializeSettings(Arguments args)
 
 		internal static void Initialize(Arguments args)
 		{
+			foreach (var f in Launch.GetType().GetFields())
```
pchote: This shouldn't be mixed in with the Settings code - launch args aren't settings.

OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
Mailaender: How would you solve it instead?

pchote: Don't persist it.  Pass the object to the things that want it, then throw it away when you're done.

Mailaender: That is again very vague. It is part of Game where most of the fields are static. Will try to put this onto the BlankLoadScreen now.

Mailaender:
Updated.

abcdefg30:
Needs a rebase.

Mailaender:
Rebased.

OpenRA.Game/Settings.cs
```diff
@@ -20,26 +20,66 @@ namespace OpenRA
 {
 	public enum MouseScrollType { Disabled, Standard, Inverted }
 
+	public class LaunchArgs
```
penev92: Is it me, or are these duplicating the `openra://` protocol and `.orarep` OS integration?

pchote: This should be moved out of Settings.

Mailaender: > Is it me, or are these duplicating the `openra://` protocol and `.orarep` OS integration?

Not duplicating, these ARE used for OS integration

> This should be moved out of Settings.

Moved.

penev92: I'm not sure I get it. Last I heard, OS integration worked on Linux as well.
Not sure I'm too comfortable with this, but as long as it doesn't break Windows integration...

Mailaender: You don't seem to get what @pchote was enforcing here. He wanted me to auto-generate those parameters for a man-page so first they became settings so I could add `[Desc()]` tags, then I had to invent something new (non-persisting settings read from args) and now I am stuck in rebase and refactor hell. Will try to fix this shortly.

pchote: To be more correct, I wanted you to autogenerate the setting code but didn't forsee the problems with these launch args.  I then suggested that the LaunchArgs args was moved away from the settings code (but not removed) as a compromise instead of rewriting that part of the code a second time.  I'm convinced that the settings automation is the best approach, but not convinced at all about the LaunchArgs stuff.

Mailaender:
Updated.

OpenRA.Game/Settings.cs
Mailaender: Yeah well, I don't care as long as I don't have to move, redo and rebase stuff again. It's been over 2 months again. I agree that manually setting LaunchSettings to `null` wasn't elegant. Should be FIXED now.

OpenRA.Game/Settings.cs
```diff
@@ -23,24 +23,52 @@ public enum MouseScrollType { Disabled, Standard, Inverted }
 
 	public class ServerSettings
 	{
+		[Desc("Sets the server name.")]
 		public string Name = "OpenRA Game";
+
+		[Desc("Sets the internal port.")]
 		public int ListenPort = 1234;
+
+		[Desc("Sets the advertised port.")]
```
pchote: This description isn't helpful unless the reader already knows what 'advertised' means in this context.

OpenRA.Game/Settings.cs
```diff
@@ -23,24 +23,52 @@ public enum MouseScrollType { Disabled, Standard, Inverted }
...
 		public bool AdvertiseOnline = true;
+
+		[Desc("Locks the game with a password.")]
 		public string Password = "";
+
 		public string MasterServer = "http://master.openra.net/";
-		public bool DiscoverNatDevices = false; // Allow users to disable NAT discovery if problems occur
-		public bool AllowPortForward = true; // let the user disable it even if compatible devices are found
+
+		[Desc("Allow users to disable NAT discovery if problems occur.")]
```
pchote: This is incorrect - the setting *enables* the use of the NAT discovery.  It just happens to be disabled by default.

OpenRA.Game/Settings.cs
```diff
@@ -23,24 +23,52 @@ public enum MouseScrollType { Disabled, Standard, Inverted }
...
+		public bool DiscoverNatDevices = false;
+
+		[Desc("Let the user disable UPnP even if compatible devices are found.")]
```
pchote: Again, this is describing the opposite of the setting.

OpenRA.Game/Settings.cs
```diff
@@ -23,24 +23,52 @@ public enum MouseScrollType { Disabled, Standard, Inverted }
...
+
 		public bool NatDeviceAvailable = false; // internal check if discovery succeeded
+
 		public int NatDiscoveryTimeout = 1000; // ms to search for UPnP enabled NATs
-		public bool VerboseNatDiscovery = false; // print very detailed logs for debugging
+
+		[Desc("Print very detailed logs for debugging port forwarding in routers.")]
+		public bool VerboseNatDiscovery = false;
+
+		[Desc("Starts the game with a default map. Input as hash. Can be optained by the utility.")]
```
pchote: optained

pchote: It would be better if this gave the utility command.

OpenRA.Game/Settings.cs
```diff
@@ -23,24 +23,52 @@ public enum MouseScrollType { Disabled, Standard, Inverted }
...
 		public string Map = null;
+
+		[Desc("Takes a comma separated list of IP addresses that are not allowed to join.")]
 		public string[] Ban = { };
 		public int TimeOut = 0;
```
pchote: This is an obscure parameter, so really wants a description too.

OpenRA.Game/Settings.cs
```diff
@@ -23,24 +23,52 @@ public enum MouseScrollType { Disabled, Standard, Inverted }
...
+
+		[Desc("Run in headless mode.")]
```
pchote: Please describe what headless mode means in this context.

OpenRA.Game/Settings.cs
```diff
@@ -23,24 +23,52 @@ public enum MouseScrollType { Disabled, Standard, Inverted }
...
 		public bool Dedicated = false;
+
+		[Desc("Automatically restart when a game ends.")]
```
pchote: I've never understood why this is even a parameter... why would somebody ever want this to be false?

OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
```diff
@@ -63,21 +81,12 @@ public void StartGame(Arguments args)
...
-				{
-					connect = args.GetValue("Launch.URI", null);
-					if (connect != null)
-					{
-						connect = connect.Replace("openra://", "");
-						connect = connect.TrimEnd('/');
-					}
-				}
-			}
+			if (!string.IsNullOrEmpty(Launch.Connect))
```
pchote: If we're going to have the LaunchArgs class, then this logic really should be encapsulated inside it.

OpenRA.Game/Settings.cs
Mailaender: That might bit rot though.

OpenRA.Game/Settings.cs
Mailaender: A shell script can also take care of restarting the server I guess.

Mailaender:
Added the requested changes.

abcdefg30:
Although I can't really test it: Looks good to me. :+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f769ff639a64dbc5616adc0875189087b9726459...9723a56fb72865d5fcc35587e61635e056c0db70)

PR #8176, opened 2015-05-16
Load textures using premultiplied alpha.

pchote:
There is a bunch of literature online about why premultiplied alpha is a good thing, so I won't repeat it here.  It also lets us simplify and improve our world sprite blend modes.

Improved harvester animation:
harv

Improved Deviator gas (same panel order as above):
deviator

In the premultiplied color space additive blend mode is the same as the alpha blend mode with Alpha: 0 and Premultiply: false.  You can create various soft mixed blends by setting Alpha part way between 0 and 1.

Fixes #8168.

PR #8249, opened 2015-05-25
Fix Alpha Blend for Path Search Debug Overlay

MutterOberin:
Pull #8176 introduced errors in debug overlay rendering of costs for path search.

Fixed by changing color of overlays.


MutterOberin:
Not at #8258

PR #8256, opened 2015-05-25
Fixed a dereference of an explicit null value in VoxelRenderer

Mailaender:
Coverity detected that code paths can lead there where this is explicitly set to null. Not sure how likely it is. The code a few lines above already has this null check.

### DIFF
--- a/OpenRA.Game/Graphics/VoxelRenderer.cs
+++ b/OpenRA.Game/Graphics/VoxelRenderer.cs
@@ -315,7 +315,8 @@ namespace OpenRA.Graphics
 				v.Second();
 			}
 
-			DisableFrameBuffer(fbo);
+			if (fbo != null)
+				DisableFrameBuffer(fbo);
 		}
 
 		public Sheet AllocateSheet()


Phrohdoh:
I'm not sure this is a dereference (but I may have terminology wrong), either way :+1:

obrakmann:
It's not here, but `DisableFrameBuffer` calls `fbo.Unbind()`. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/fea5c5a2f57062f441ad4ef1f30373313478a14f%5E...fea5c5a2f57062f441ad4ef1f30373313478a14f)

OpenRA.Game/Graphics/VoxelRenderer.cs
```diff
@@ -315,7 +315,8 @@ public void EndFrame()
 				v.Second();
 			}
 
-			DisableFrameBuffer(fbo);
+			if (fbo != null)
+				DisableFrameBuffer(fbo);
```
penev92: Are we sure this shouldn't still execute
```
			Game.Renderer.Flush();
			Game.Renderer.Device.DisableDepthBuffer();
```

I guess it never happens anyway, or we would have heard about a crash, but still...

Issue #8260, opened 2015-05-25
ISpeedModifier does not work anymore


Mailaender:
Both `TakeCover` and `ScaredyCat` are affected. Tested on RA and TS.

penev92:
That *might* have something to do with my 2 warhead PRs (actually the first one).

Mailaender:
Shall we revert them or do you want to fix it?

penev92:
Turns out my changes had nothing to do with it. Still, I filed a fix.

PR #8261, opened 2015-05-25
Added the Tiberian Sun cyborg logic

Mailaender:
See http://modenc.renegadeprojects.com/Cyborg for reference.

### DIFF
--- a/OpenRA.Mods.TS/OpenRA.Mods.TS.csproj
+++ b/OpenRA.Mods.TS/OpenRA.Mods.TS.csproj
@@ -52,6 +52,7 @@
     <Compile Include="Activities\VoxelHarvesterDockSequence.cs" />
     <Compile Include="SpriteLoaders\TmpTSLoader.cs" />
     <Compile Include="Traits\Buildings\TiberianSunRefinery.cs" />
+    <Compile Include="Traits\Render\WithPermanentInjury.cs" />
     <Compile Include="Traits\Render\WithVoxelWalkerBody.cs" />
     <Compile Include="Traits\Render\WithVoxelUnloadBody.cs" />
     <Compile Include="Traits\Render\WithVoxelWaterBody.cs" />
--- a/OpenRA.Mods.TS/Traits/Render/WithPermanentInjury.cs
+++ b/OpenRA.Mods.TS/Traits/Render/WithPermanentInjury.cs
@@ -1,0 +1,46 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.TS.Traits
+{
+	[Desc("Change the sprite after a certain amount of damage is taken, even when the hitpoints are regenerated.")]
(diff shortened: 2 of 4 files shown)


abcdefg30:
> 21:32 (abcdefg30) ah btw, could we have a damagemodifier on WithPermanentInjury?
21:32 (abcdefg30) actually the damaged cyborg is lying so it should take less damage as with takecover

pchote:
If this granted an upgrade instead of hardcoding a specific modifier(s) then we (and modders) could add any effect that they wanted while damaged.

penev92:
I agree this would be much more useful as `UpgradeOn(Critical)Damage`. We already can change sprites and movement speed on upgrade. We can also change voxels, overlays, weapons, (soon armor, hopefully), and a ton of other stuff.

Currently it is basically hardcoded to do the TS Cyborg behavior. :/

abcdefg30:
Although it was discussed on IRC that we actually want to have an implementation using upgrades,
I'll :+1: / :white_check_mark: this. As @Mailaender argued this can be replaced easily if someone is up to make an upgrade-based implementation and this works for now.

Note to reviewers: This is suffering from #8260.

Mailaender:
I found no evidence at http://modenc.renegadeprojects.com/Cyborg that injured cyborgs take less damage luck prone infantry (shall I really add it?) and I am also not sure if they are really walking slower (shall I remove it?) although that seemed to make sense to me.

abcdefg30:
> I found no evidence at http://modenc.renegadeprojects.com/Cyborg that injured cyborgs take less damage like prone infantry (shall I really add it?)

Yes please, then.

Mailaender:
Removed the speed modifier. Also won't add the damage modifier. I don't intend to re-balance the mod. Also won't implement any upgrade plugs. This should just fix the rendering.

LipkeGu:
SpeedModifider should be go to an extra PR together with the other thing :/

chrisforbes:
:+1:

OpenRA.Mods.TS/Traits/WithPermanentInjury.cs
```diff
@@ -0,0 +1,46 @@
...
+ * see COPYING.
+ */
+#endregion
+
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.TS.Traits
+{
+	[Desc("Change the sprite and movement speed permanently after a certain amount of damage is taken, even when the hitpoints are regenerated.")]
```
penev92: No more "and movement speed".

Mailaender: Corrected.

SMIFFGIG:
I think this should be a % chance of them losing their legs. Rather than a hardcoded 100% chance when they reach low health.

Not only does it not make sense that they lose specifically their legs 100% of the time they get damaged. But in the cyborg animation sequence there are multiple death frames for when they cyborg is in standing non prone position (same applies to cyborg commando) which go totally unused in TS.

Unit158:
Adding RNG on a specific unit is a :-1: from me.

Unit158:
I could, however, understand the legs being gibbed by certain weapons.

SMIFFGIG:
Also by making this happen 100% of the time. All of the death animations for the cyborg+cyborg commando will go totally unused.
Seems to make little sense that say a flamethrower could remove the legs of the unit and unfortunate if the specific flame death animation for this unit would also go unused as a consequence.
An oversight from Westwood in the original TS which I hope isn't transferred to OpenRA TS.

GraionDilach:
Could you please open a new issue about this? Think we already discussed this.

PR #7967, opened 2015-04-17
Add descriptions to all TS units

abcdefg30:
Descriptions "inspired" by the articles found at
http://cnc.wikia.com/wiki/Category:Tiberian_Sun_Nod_Arsenal and
http://cnc.wikia.com/wiki/Category:Tiberian_Sun_GDI_Arsenal .

Suggestions for improvement wanted. :)

PR #8277, opened 2015-05-27
Update all ra tooltips

abcdefg30:
"Follow-up" of #7967.

### DIFF
--- a/mods/ra/rules/aircraft.yaml
+++ b/mods/ra/rules/aircraft.yaml
@@ -86,7 +86,7 @@ MIG:
 		Cost: 2000
 	Tooltip:
 		Name: Mig Attack Plane
-		Description: Fast Ground-Attack Plane.\n  Strong vs Buildings\n  Weak vs Infantry, Light Vehicles
+		Description: Fast Ground-Attack Plane.\n  Strong vs Buildings, Tanks\n  Weak vs Infantry, Light armor, Aircraft
 	Health:
 		HP: 70
 	Armor:
@@ -139,7 +139,7 @@ YAK:
 		Cost: 1350
 	Tooltip:
 		Name: Yak Attack Plane
-		Description: Anti-Tanks & Anti-Infantry Plane.\n  Strong vs Infantry, Tanks\n  Weak vs Buildings
+		Description: Anti-Tanks & Anti-Infantry Plane.\n  Strong vs Infantry, Vehicles\n  Weak vs Aircraft
 	Health:
 		HP: 60
 	Armor:
@@ -240,7 +240,7 @@ HELI:
 		Cost: 2000
 	Tooltip:
 		Name: Longbow
-		Description: Helicopter Gunship with AG Missiles.\n  Strong vs Tanks\n  Weak vs Infantry
+		Description: Helicopter gunship armed\nwith multi-purpose missiles.\n  Strong vs Tanks, Aircraft\n  Weak vs Infantry
 	Health:
 		HP: 120
 	Armor:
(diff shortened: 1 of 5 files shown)


mods/ra/rules/infantry.yaml
```diff
@@ -220,7 +220,7 @@ SPY:
 	DisguiseToolTip:
 		Name: Spy
 		GenericName: Soldier
-		Description: Infiltrates enemy structures for intel or\nsabotage. Exact effect depends on the\nbuilding infiltrated.\n  Strong vs Nothing\n  Weak vs Everything\n  Special Ability: Disguised
+		Description: Infiltrates enemy structures for intel or\nsabotage. Exact effect depends on the\nbuilding infiltrated.\nLooses disguise when attacking.\n  Strong vs Infantry\n  Weak vs Vehicles, Aircraft\n  Special Ability: Disguised
```
ScottNZ: Loses

mods/ra/rules/aircraft.yaml
```diff
@@ -230,7 +230,7 @@ HELI:
 		Cost: 2000
 	Tooltip:
 		Name: Longbow
-		Description: Helicopter Gunship with AG Missiles.\n  Strong vs Tanks\n  Weak vs Infantry
+		Description: Helicopter Gunship with AG and AA Missiles.\n  Strong vs Tanks, Aircraft\n  Weak vs Infantry
```
penev92: how about "Helicopter gunship armed with multi-purpose missiles."

mods/ra/rules/aircraft.yaml
```diff
@@ -277,7 +277,7 @@ HIND:
 		Cost: 1350
 	Tooltip:
 		Name: Hind
-		Description: Helicopter Gunship with Chainguns.\n  Strong vs Infantry, Light Vehicles.\n  Weak vs Tanks
+		Description: Helicopter Gunship with Chainguns.\n  Strong vs Infantry, Light armor.\n  Weak vs Tanks, Aircraft
```
penev92: Drop a few upper-case letters and maybe use "armed with dual chainguns".

mods/ra/rules/infantry.yaml
```diff
@@ -75,7 +75,7 @@ E2:
 		Cost: 160
 	Tooltip:
 		Name: Grenadier
-		Description: Infantry armed with grenades. \n  Strong vs Buildings, Infantry\n  Weak vs Vehicles
+		Description: Infantry armed with grenades. \n  Strong vs Buildings, Infantry\n  Weak vs Vehicles, Aircraft
```
penev92: Redundant ' ' before '\n'.

obrakmann:
It would be nice if all those instances where a unit is said to be "weak" against something when it can't even shoot at it at all could be cleaned up. Like "weak against aircraft" for the yak, for example. That implies to me that it does at least a little damage, not that it doesn't do anything at all.

Phrohdoh:
I've made that argument before but it was outvoted.

mods/ra/rules/structures.yaml
```diff
@@ -977,7 +977,7 @@ SILO:
 		Cost: 150
 	Tooltip:
 		Name: Silo
-		Description: Stores excess refined Ore and gems.
+		Description: Stores excess refined\nOre and gems.
```
Mailaender: Gems is in title case in the line above.

Mailaender:
:+1: This corrects the tooltip it in several cases.

mods/ra/rules/structures.yaml
abcdefg30: Fixed.

penev92:
Looks good to me :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/813394dcbe4be2a11bf07e5d6ce073a7bae2ff54...cb45ff5db15d738468c3e3c2320d35b1cbba5121)

PR #8187, opened 2015-05-18
Brought back the Windows setup deployment via Travis CI

Mailaender:
Closes #8153
- Travis CI will do the final Windows deployment so every official release CLI get's compiled by the same compiler using the same flags to avoid possible desyncs.
- We switch back the legacy Travis CI non-Docker environment, because we need sudo to install NSIS. ~~The nuget connectivity problems should be gone now thanks to caching.~~ Caching is not supported on the non-container based infrastructure. :(
- I decided to still keep the AppVeyor Setup creation so we keep our nightly builds and CI checks for the NSIS scripts which tend to bit rot faster, because everything has to be listed and the syntax is error prone.

Issue #8153, opened 2015-05-12
AppVeyor Windows build significantly slower than Mono build running on Microsoft Windows


reaperrr:
My numbers on Win7x64, i5-2500 CPU, RA shellmap:

OSX release 20150424: avg ~4.0-4.5 Tick / ~2.5-3.0 Render
Win release 20150424: avg ~5.5-6.0 Tick / ~4.0-4.5 Render

So the mono build is at least roughly 30-50% faster with identical compile settings (Debug mode, no optimization).

pchote:
```
23:27 < pchote> I was really hoping that the release builds under windows were just that much better
23:27 < pchote> but oh well
23:27 < pchote> do you want to file an issue with some numbers?
23:28 < pchote> we can resurrect antares's branch that shifts the windows builds back to travis
23:29 < pchote> and this gives us some evidence stronger than "pchote has the heebie jeebies" to merge it
```

obrakmann:
If anyone else wants to compare, here is current bleed (git-7154ce4) [compiled on Appveyor](https://ci.appveyor.com/api/buildjobs/lpv6k3b1d675q8iy/artifacts/OpenRA-.exe), and here's bleed [compiled with mono](https://seacloud.cc/d/5d1adbebce/files/?p=/openra/OpenRA-mono-20150513.exe)

Mailaender:
That is terrible news. Is the slowness undone if you compile OpenRA with msbuild/VS in Release mode on Windows?

Mailaender:
I tested the AppVeyor Windows build (setup extracted using Wine, adding only the .dll.conf files) vs. a self-compiled Makefile build running on Linux/Mono and I can't find a significant performance difference. My assumption is that it is not the generated IL code that is slow or too different, but some kind of flag that is set which only the original .NET runtime seems to adhere.

reaperrr:
That flag is probably only set in debug mode then, because while I haven't tested with 20150424 in particular yet, both my previous tests as well as @RoosterDragon's Release config branch showed a massive performance advantage for Release config over Debug config when compiling with msbuild/VS.

RoosterDragon:
I can confirm similar speedups as @reaperrr reports. Running on Windows the OSX build has noticeably lower times than the Windows build. I tried both the MS and mono runtimes and they produce similar numbers for both binaries.

Compiling the 20150424 release (via Visual Studio) in release mode with optimizations on produces results comparable to the OSX binary.

[...]

PR #8282, opened 2015-05-28
Enable compiler optimizations

RoosterDragon:
Adds a release config to the solution with optimizations enabled, and changes the makefile to use these flags too.

We should test this a fair amount to ensure it doesn't introduce any desyncs. I suggest we collect some replays from the latest release (ideally from a variety of OSes) and have some contributors replay them using an optimized build on their system. Any new desyncs would indicate we have a problem.

I don't think the risk of desyncs is too high though, I present some more detailed reasoning below.

---

There's been a worry for some time that enabling compiler optimizations might provoke desyncs due to differences in how the mono and MS runtimes might optimize things. However, it seems things are less risky than we thought because it appears that both runtimes are already applying runtime optimizations!.

We currently build with the `-optimize-` flag passed to the compiler, which disables compiler optimizations.

The compiler doesn't really do any optimizations though - the real optimization happens at runtime when the jitter generates the actual code your machine runs.

If you build via MSBuild and run the binary on Windows, it appears the MS runtime uses the compile time flag to decide if it's going to optimize or not.

The mono runtime has a separate command line flag that control optimizations at runtime. If you run a binary (built by either MSBuild or mono), the performance is dependant on this flag, and the compiler flag is essentially irrelevant.

@Phrohdoh was kind enough to demonstrate. I get similar results on Windows (running an MSBuild binary with compiler optimizations disabled).

Unspecified flag: Run with the default set of optimizations.
`--optimize=all`: Run with all optimizations.
`--optimize=-all`: Run with no optimizations.

```
[20:01] RoosterDragon   Run OpenRA via mono with a couple of different flags and see what the ticks times are like on the shellmap.
[20:01] Phroh|orca  Which flags? Which mod?
[20:01] RoosterDragon   Let's say RA since it has a busy shellmap.
[20:01] RoosterDragon   Please try these three:
[20:02] RoosterDragon   'mono OpenRA.Game.exe'
[20:02] RoosterDragon   'mono --optimize=all OpenRA.Game.exe'
[20:02] RoosterDragon   'mono --optimize=-all OpenRA.Game.exe'
[20:02] RoosterDragon   Just eyeball the numbers.
[20:03] Phroh|orca  12.2-15.4 for mono openra.game.exe
[20:03] Phroh|orca  ~700 ticks
[20:04] Phroh|orca  12-15.6 for optimize=all
[20:05] Phroh|orca  20-34 for optimize=-all
```

This would explain why the AppVeyor build was so much slower in #8153. Mono has been optimizing all along but the MS runtime was not. In #8187, we reintroduced a mono-built binary to Windows. I speculate that the binary built by mono does not contain a "don't optimize" flag that the MS runtime understands, and so it optimizes by default. I think this because an optimized MSBuild runs at a similar speed as the mono build.

Anyway, assume the hopefully reasonable speculation to explain the observed performance of everything is correct, then in reality we're deploying a mono build with compiler optimizations disabled, but which both runtimes were optimizing at runtime anyway.

Therefore, if we can bring back a release config:
- We could consider using AppVeyor for Windows builds again, as there wouldn't be a perf hit.
- People building via Visual Studio can build using the release config and get the same perf as everybody else.
- Whatever minor benefits compiler optimizations would enable are applied.

There is of course still a small risk this could introduce desyncs somehow, so we should do some testing before moving forward.


pchote:
Sounds reasonable, so i'm willing to give this a shot.  Needs a rebase, though.

I still don't trust appveyor for windows packaging, so lets keep that on travis for now.

OpenRA.sln
```diff
@@ -1,6 +1,8 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
+# Visual Studio 2013
+VisualStudioVersion = 12.0.31101.0
+MinimumVisualStudioVersion = 10.0.40219.1
```
pchote: This appears to bump the minimum requires VS version?

OpenRA.Test/OpenRA.Test.csproj
```diff
@@ -19,7 +19,17 @@
     <DebugType>full</DebugType>
     <PlatformTarget>x86</PlatformTarget>
     <ErrorReport>prompt</ErrorReport>
-    <CodeAnalysisRuleSet>ManagedMinimumRules.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisRuleSet>AllRules.ruleset</CodeAnalysisRuleSet>
```
pchote: What's the difference between these?

OpenRA.sln
```diff
@@ -40,17 +42,17 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tiberian Dawn Lua scripts",
...
 		mods\ra\maps\allies-03b\allies03b.lua = mods\ra\maps\allies-03b\allies03b.lua
-		mods\ra\maps\allies-05a\AI.lua = mods\ra\maps\allies-05a\AI.lua
 		mods\ra\maps\allies-05a\allies05a.lua = mods\ra\maps\allies-05a\allies05a.lua
-		mods\ra\maps\soviet-01\soviet01.lua = mods\ra\maps\soviet-01\soviet01.lua
 		mods\ra\maps\desert-shellmap\desert-shellmap.lua = mods\ra\maps\desert-shellmap\desert-shellmap.lua
-		mods\ra\maps\intervention\intervention.lua = mods\ra\maps\intervention\intervention.lua
 		mods\ra\maps\fort-lonestar\fort-lonestar.lua = mods\ra\maps\fort-lonestar\fort-lonestar.lua
+		mods\ra\maps\intervention\intervention.lua = mods\ra\maps\intervention\intervention.lua
 		mods\ra\maps\monster-tank-madness\monster-tank-madness.lua = mods\ra\maps\monster-tank-madness\monster-tank-madness.lua
+		mods\ra\maps\soviet-01\soviet01.lua = mods\ra\maps\soviet-01\soviet01.lua
```
penev92: I think we'll want to keep the order of the scripts by missions.

OpenRA.sln
RoosterDragon: I'm being extremely lazy in this PR and including all the stuff VS liked to change arbitrarily.

I could happily leave it out, on the other hand it would be nice to eliminate some of the stuff the IDE changes all the time.

I don't know what effect this setting has, but given that VS is free these days and people can get 2013 (and soon 2015) easily enough, it probably doesn't matter?

OpenRA.Test/OpenRA.Test.csproj
RoosterDragon: This affects the ruleset used when VS (and other tools?) do static code analysis.

I am changing only to make it consistent with all the others.

OpenRA.sln
RoosterDragon: VS likes to automatically alphabetise these whenever it updates the solution file.

I don't mind keeping the order, but if people don't mind the VS imposed order you'd remove a small nuisance from my day having to reset them all the time.

RoosterDragon:
I was intentionally leaving this based on the last release for ease of testing, if we want to do some testing.

Would you prefer to just merge it and see what happens?

OpenRA.sln
penev92: This sets the minimum supported version to Visual Studio 2010, so I think we can live with that. I should check how the line above this affects my Visual Studio 2012 though, because you are setting the current version to 2013.

Mailaender:
Can you file the pull request against master then? That will make the pull request mergeable and enable the AppVeyor CI preview builds of the pull request.

pchote:
Filing this against master would break our release processes.  This appears to just add back the release profiles, so as long as the default stays as debug I don't see any reason not to merge it.  Tests can then be done with self-compiled builds vs future playtests/releases.

abcdefg30:
For the record: I watched a bunch of replays from the release (around 20) and didn't encounter any desyncs. :)

RoosterDragon:
To clarify: Is the current request to rebase this on bleed?

pchote:
@RoosterDragon: That's my request, at least.  I'm assuming that this still keeps the debug build as the default.

RoosterDragon:
Ok, rebased to bleed.

I assume the solution config will default to Debug just by nature of it being listed first, might depend on the IDE though.

I have however just replaced the flags in the makefile to release versions. If we need the makefile to have a debug & release mode somebody who knows make-fu will need to help me out with that.

obrakmann:
I made a pr for the Makefile changes: https://github.com/RoosterDragon/OpenRA/pull/1.

Not sure if we want to enable release builds on travis.

RoosterDragon:
Merged. Is is convention to rebase in PRs of PRs or will a merge suffice?

obrakmann:
In the past we've usually used rebase to get rid of the merge commit.

RoosterDragon:
Ok rebased too.

obrakmann:
We discussed it in the channel and decided to drop the last commit.

OpenRA.Game/OpenRA.Game.csproj
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Mailaender: Is this adding or removing a byte order mark?

RoosterDragon: This would be adding a BOM. VS adds them in automatically any time you touch these files, so this makes in consistent with the rest at least.

RoosterDragon:
Ok dropped the commit the made Travis use the release build.

Mailaender:
As a followup we may want to make use of `mono --debug` to actually get stacktraces with line numbers upon crashes. https://github.com/RoosterDragon/OpenRA/blob/release-config/Makefile#L424

obrakmann:
:+1: I'm gonna merge this now. This doesn't actually change anything wrt to our builds, you have to manually intervene to select the release profile or disable debug mode in the Makefile.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4d6b0b5f9bfcdd7c25133ba1740eb73457cefaf1%5E...4d6b0b5f9bfcdd7c25133ba1740eb73457cefaf1)

PR #8284, opened 2015-05-28
Moves Disguise and related traits from Mods.RA to Mods.Common.

deniz1a:
Also fixes some issues.

### DIFF
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -293,6 +293,7 @@
     <Compile Include="Traits\CustomSellValue.cs" />
     <Compile Include="Traits\Demolishable.cs" />
     <Compile Include="Traits\DetectCloaked.cs" />
+    <Compile Include="Traits\Disguise.cs" />
     <Compile Include="Traits\EjectOnDeath.cs" />
     <Compile Include="Traits\EmitInfantryOnSell.cs" />
     <Compile Include="Traits\EngineerRepair.cs" />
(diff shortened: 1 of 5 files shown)


deniz1a:
And satellite code should probably also move to Mods.Common since RA2 will use it too.

penev92:
Not yet.

deniz1a:
After satellite code refactor?

penev92:
After it is actually needed.

deniz1a:
But it will be needed?

deniz1a:
Also currently IgnoresDisguise.cs is in Mods.Common. So this patch would move the rest of the disguise stuff there too.

penev92:
Moving of files is pretty easy to follow, but what exactly are you fixing in Disguise.cs?

deniz1a:
* Incorrect example in comments. Spies can't disguise as dogs anymore.
* Radar color and displayed owner are corrected. Now all states of current player are checked and the real owner is shown to allies.

reaperrr:
:-1: to moving RA-specific stuff to Common just because the RA2 mod will use it. If it's just the RA2 mod that will use it (which is likely), the RA2 mod should simply use the Mods.RA.dll, in my opinion. Otherwise we might as well ditch the concept of mod-specific dlls.

Phrohdoh:
FWIW RA2 does not currently reference the RA assembly.

Just Common and TS and eventually RA2.

Mailaender:
I agree this is a very specific (and hacky) Red Alert feature. We didn't move it to Mods.Common on purpose.

deniz1a:
What if a custom mods wants to have disguise trait?

reaperrr:
> What if a custom mods wants to have disguise trait?

Then it should use the Mods.RA assembly.

Phrohdoh:
To be fair we did move the voxel stuff to Common.

GraionDilach:
To be fair, TS also uses the Disguise trait already.

PR #8290, opened 2015-05-29
Give D2k its own mission browser

penev92:
Since D2k has different requirements about the size and shape of some components than RA, it gets its own mission browser.
Feel free to compare with bleed and decide which looks better ;)

### DIFF
--- a/mods/d2k/chrome/missionbrowser.yaml
+++ b/mods/d2k/chrome/missionbrowser.yaml
@@ -1,0 +1,146 @@
+Background@MISSIONBROWSER_PANEL:
+	Logic: MissionBrowserLogic
+	X: (WINDOW_RIGHT - WIDTH)/2
+	Y: (WINDOW_BOTTOM - HEIGHT)/2
+	Width: 960
+	Height: 690
+	Children:
+		Label@MISSIONBROWSER_TITLE:
+			Y: 20
+			Width: PARENT_RIGHT
+			Height: 25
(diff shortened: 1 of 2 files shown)


abcdefg30:
Well, this is just a bigger version, isn't it? :D
Anyway looks good to me except this overlap, but I'm not sure if this is a issue:
d2kmissionoverlap

abcdefg30:
> Anyway looks good to me except this overlap, [...]

Doesn't seem like an issue to anyone. So :+1: / :white_check_mark:

Mailaender:
It adjusts it to the bigger map size. :+1: 

> Anyway looks good to me except this overlap

Should be easy to fix. Just move the performance graph further to the left.

image

penev92:
This overlap depends on the windows size you have OpenRA.
While the mission browser is static, the performance graph is relative to the bottom right corner.
So for people with big screens there will be no overlap. And for people with small screens the performance graph may be completely hidden.
With small enough window size this would happen with every single windows across every mod - lobby, settings menu, asset browser.

abcdefg30:
I count :+2:, therefore merging now.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/139a5760a535bab13f32c1d40e6b634289b6d84a...a5f880bbed1b593414764ef47ef0802d56dc6960)

PR #7709, opened 2015-03-22
Added checks for missing sequence entries

Mailaender:
Closes #2283.

I experienced two crashes recently: 
- The sietch not having make animations while testing #4496.
- Deployed Tiberian Sun units not having death animations, a regression from https://github.com/OpenRA/OpenRA/pull/7687.

I extended the rule to also cover sequence definition on the actor from all traits (including overlays) and discovered some more inconsistencies in seldom used units. Also took some effort to clean things up.

PR #8296, opened 2015-05-30
Moved the sequence image check to RenderSprites level

Mailaender:
This detected several unnecessary RenderSprites traits in the Tiberian Sun mods. Split from https://github.com/OpenRA/OpenRA/pull/7709.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckSequences.cs
+++ b/OpenRA.Mods.Common/Lint/CheckSequences.cs
@@ -20,14 +20,21 @@ namespace OpenRA.Mods.Common.Lint
 		public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 		{
 			var sequences = MiniYaml.MergeLiberal(map.SequenceDefinitions, Game.ModData.Manifest.Sequences.Select(s => MiniYaml.FromFile(s)).Aggregate(MiniYaml.MergeLiberal));
+			var races = map.Rules.Actors["world"].Traits.WithInterface<CountryInfo>().Select(c => c.Race);
 
 			foreach (var actorInfo in map.Rules.Actors)
-				foreach (var renderInfo in actorInfo.Value.Traits.WithInterface<RenderSimpleInfo>())
+			{
+				foreach (var renderInfo in actorInfo.Value.Traits.WithInterface<RenderSpritesInfo>())
 				{
-					var image = renderInfo.Image ?? actorInfo.Value.Name;
-					if (!sequences.Any(s => s.Key == image.ToLowerInvariant()) && !actorInfo.Value.Name.Contains("^"))
-						emitWarning("Sprite image {0} from actor {1} has no sequence definition.".F(image, actorInfo.Value.Name));
+					foreach (var race in races)
+					{
+						var image = renderInfo.GetImage(actorInfo.Value, map.Rules.Sequences[map.Tileset], race);
+						if (!sequences.Any(s => s.Key == image.ToLowerInvariant()) && !actorInfo.Value.Name.Contains("^"))
+							emitWarning("Sprite image {0} from actor {1} on tileset {2} using race {3} has no sequence definition."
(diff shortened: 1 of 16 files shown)


OpenRA.Mods.Common/Traits/Render/WithCrateBody.cs
```diff
@@ -20,8 +20,6 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Renders crates with both water and land variants.")]
 	class WithCrateBodyInfo : ITraitInfo, Requires<RenderSpritesInfo>, IQuantizeBodyOrientationInfo, IRenderActorPreviewSpritesInfo
 	{
-		public readonly string[] Images = { "crate" };
```
pchote: This breaks the ability to have randomized crate variants, if we still want that.

Mailaender: It was never used. Instead we define different crate types nowadays more closely towards the originals.

OpenRA.Mods.Common/Traits/Render/WithCrateBody.cs
```diff
@@ -46,7 +44,8 @@ public WithCrateBody(Actor self, WithCrateBodyInfo info)
 		{
 			this.self = self;
 			var rs = self.Trait<RenderSprites>();
-			var images = info.XmasImages.Any() && DateTime.Today.Month == 12 ? info.XmasImages : info.Images;
+			var image = rs.GetImage(self);
+			var images = info.XmasImages.Any() && DateTime.Today.Month == 12 ? info.XmasImages : new string[] { image };
```
ScottNZ: Use `new[]` instead of `new string[]`

chrisforbes:
:+1: Looks OK to me.

OpenRA.Mods.Common/Lint/CheckSequences.cs
```diff
@@ -22,12 +22,14 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 			var sequences = MiniYaml.MergeLiberal(map.SequenceDefinitions, Game.ModData.Manifest.Sequences.Select(s => MiniYaml.FromFile(s)).Aggregate(MiniYaml.MergeLiberal));
 
 			foreach (var actorInfo in map.Rules.Actors)
-				foreach (var renderInfo in actorInfo.Value.Traits.WithInterface<RenderSimpleInfo>())
+			{
+				foreach (var renderInfo in actorInfo.Value.Traits.WithInterface<RenderSpritesInfo>())
 				{
 					var image = renderInfo.Image ?? actorInfo.Value.Name;
```
penev92: I believe this is what @pchote meant by
>This needs to use RenderSpritesInfo and query the image name from GetImage(). The render traits moved on from renderInfo.Image ?? actorInfo.Value.Name ages ago.

?
Although I don't see why `RenderSimpleInfo` needs to change to `RenderSpritesInfo` to use `.GetImage()`.

Mailaender: The check will cover some more rendering traits that are lower level such as crates.

OpenRA.Mods.Common/Traits/Render/WithCrateBody.cs
penev92: RA seems to be using it in a number of places.

penev92:
:+1:  from me with the above comment.

penev92:
Now also needs a rebase.
This also already has +2, so would be nice to get it merged.

Mailaender:
Updated.

penev92:
Help wanted with changelog entries.

Phrohdoh:
Add this onto the original changelog if this didn't make it into release.

Otherwise it is just a general improvement of organization, IMO.

Mailaender:
We can document it at as part of https://github.com/OpenRA/OpenRA/issues/7709.

PR #8298, opened 2015-05-30
Corrects displayed owner of disguised actors.

deniz1a:
Now allies also see the real owner of disguised units.

### DIFF
--- a/OpenRA.Mods.RA/Traits/Disguise.cs
+++ b/OpenRA.Mods.RA/Traits/Disguise.cs
@@ -48,10 +48,16 @@ namespace OpenRA.Mods.RA.Traits
 		{
 			get
 			{
-				if (disguise.Disguised)
-					return self.Owner == self.World.LocalPlayer ? self.Owner : disguise.AsPlayer;
+				if (!disguise.Disguised)
+					return self.Owner;
 
-				return self.Owner;
+				Player p = null;
+				if (self.World.LocalPlayer != null && self.World.LocalPlayer.WinState == WinState.Undefined)
+					p = self.World.LocalPlayer;
+				else
+					p = self.World.RenderPlayer;
+
+				return self.Owner.IsAlliedWith(p) ? self.Owner : disguise.AsPlayer;
 			}
 		}
 	}
@@ -100,7 +106,16 @@ namespace OpenRA.Mods.RA.Traits
 
 		public Color RadarColorOverride(Actor self)
 		{
-			if (!Disguised || self.Owner.IsAlliedWith(self.World.RenderPlayer))
+			if (!Disguised)
+				return self.Owner.Color.RGB;
+
+			Player p = null;
+			if (self.World.LocalPlayer != null && self.World.LocalPlayer.WinState == WinState.Undefined)
+				p = self.World.LocalPlayer;
+			else
+				p = self.World.RenderPlayer;
+
+			if (self.Owner.IsAlliedWith(p))
 				return self.Owner.Color.RGB;
 
 			return AsPlayer.Color.RGB;
@@ -112,8 +127,8 @@ namespace OpenRA.Mods.RA.Traits
 
 			if (target != null)
 			{
-				// Take the image of the target's disguise, if it exist.
-				// E.g., SpyA is disguised as a dog. SpyB then targets SpyA. We should use the dog image.
+				// Take the image of the target's disguise, if it exists.
+				// E.g., SpyA is disguised as a rifle infantry. SpyB then targets SpyA. We should use the rifle infantry image.
 				var targetDisguise = target.TraitOrDefault<Disguise>();
 				if (targetDisguise != null && targetDisguise.Disguised)
 				{


OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -48,10 +48,16 @@ public Player Owner
...
 			get
 			{
-				if (disguise.Disguised)
-					return self.Owner == self.World.LocalPlayer ? self.Owner : disguise.AsPlayer;
+				if (!disguise.Disguised)
+					return self.Owner;
 
-				return self.Owner;
+				Player p = null;
+				if (self.World.LocalPlayer != null && self.World.LocalPlayer.WinState == WinState.Undefined)
```
pchote: These probably want to be using `var p = self.World.RenderPlayer ?? self.World.LocalPlayer;` to be consistent with our other observer-aware checks.

deniz1a:
But that wouldn't be correct according to this diagram: https://cloud.githubusercontent.com/assets/6599106/7490202/078aa156-f3e7-11e4-9033-d183480a1e75.png

If RenderPlayer is null, there are three possibilities:
1) spectator with disable shroud view
2) player in game with disable shroud cheat
3) player observing after game end with disable shroud view

``` var p = self.World.RenderPlayer ?? self.World.LocalPlayer; ``` would get LocalPlayer for the third option. But it should be null. Because self.Owner might not be allied with LocalPlayer, but since we're observing after game end, that shouldn't matter.

**Comment by pchote** [comment]
That 3) case will also have the same problem for team health colors and multi-tap actor selection.  It's important to use the same code patterns for these cases that do the same queries, otherwise we devolve into a mess of edge cases.

**Comment by deniz1a** [comment]
Ok then shouldn't this version be used and other code changed?

**Comment by matija-hustic** [comment]
With things as they are here,
* spectators viewing as `Everyone` and
* players with the disable shroud cheat enabled
would not see the disguised spy's real owner and color on radar which I find counter-intuitive.

Two unrelated points that might be worth considering in this disguise-related PR:
* It is allowed to guard and auto-heal enemy disguised units. Is this behavior desired?
* RA2 periodically changes disguised spies' rendering for the owner (might also be nice to have it change for allies, too), i.e., the owner sees the spy's disguise 1/6 of the time and the spy's original sprite 5/6 of the time. These two sprites are exchanged through a fade-out / fade-in effect.

**Comment by penev92** [comment]
- I see no problem in tricking an enemy medic to heal your spy (same for guarding).
 - While the fade effect is cool (at first, then I find it annoying later on), #7466 would make that redundant.

File: OpenRA.Mods.RA/Traits/Disguise.cs (hunk 1 above)
Comment (obrakmann) [inline OpenRA.Mods.RA/Traits/Disguise.cs]: The idiom that should be used is this:

```csharp
var p = self.World.RenderPlayer ?? self.World.LocalPlayer;
if (p == null || p.Spectating)
        // return spectator stuff
else
     return p.IsAlliedWith(self.Owner) ? // <stuff for allies> : <stuff for enemies>
```

matija-hustic: @obrakmann this would put the case of the player with the disable shroud cheat into `<stuff for allies>`, right? I'd assume that using that cheat should allow me to see disguised spies even for enemies.

obrakmann: Yep, you're right. Argh, so many special cases!

It would have to be
```csharp
var p = self.World.RenderPlayer ?? self.World.LocalPlayer;
if (p == null || p.Spectating || self.World.RenderPlayer == null)
        // return spectator stuff
else
     return p.IsAlliedWith(self.Owner) ? // <stuff for allies> : <stuff for enemies>
```
then if you wanted to do that. That looks kinda ugly now, though. Surely there's a way to express those conditions better, but I'm not seeing it atm.

matija-hustic: When I thought about this the other day, I figured that there's just two fundamental cases
* when `RenderPlayer.InternalName == "Everyone" || self.Owner.IsAlliedWith(RenderPlayer)` then you ought to see through the disguise (allied case: allies in game, allies spectators, everyone spectator, players with disable shroud cheat),
* and otherwise just show the disguise (enemy case).

deniz1a: With player with disabled shroud cheat you mean allied player, right? Because enemy player should not see through disguise even with disable shroud cheat.

Also, ```RenderPlayer.InternalName == "Everyone" ||``` is not necessary since that is already allied with all players, the second part would cover it.

matija-hustic: I'd argue that any player with the disable shroud cheat should see thru the disguise since that cheat enables them already too see things like ore load in a harvester and refinery, as well as ammo loads etc.

And are you sure about `Everyone` being an ally to all players? If I remember correctly, it is neutral and that makes that second check necessary.

deniz1a: Disable shroud cheat makes enemy pips (ammo, refinery...) visible? I think that's a bug.

And yes, "Everyone" is allied with everyone heh. Look here:
https://github.com/OpenRA/OpenRA/blob/12966129a4e412fd2c53eb05b310f30557058521/OpenRA.Mods.Common/Traits/World/CreateMPPlayers.cs#L49

penev92: #NotABug
It is a cheat that allows you to see everything.

deniz1a: Then it should be renamed I think. Disable shroud suggests only disabling shroud...

matija-hustic: @deniz1a, our misunderstanding comes from the fact that stances are not a symmetric relation. `Everyone` considers every player an ally, but all players consider `Everyone` as neutral. So when making this check, you have to pay attention whether you call `self.Owner.IsAlliedWith` or `RenderPlayer.IsAlliedWith` because they will return different values.
You can either use `RenderPlayer.InternalName == "Everyone" || self.Owner.IsAlliedWith(RenderPlayer)` or `RenderPlayer != null && RenderPlayer.IsAlliedWith(self.Owner)`.

Phrohdoh: Disabling shroud should not implicitly mean seeing *everything* (pips, disguise, etc).

penev92:
OK if I understand this correctly, this seems to be held up on a simple check, the likes of which we've seen several times before.
Please refer to @cjshmyr's team healthbars PR (merged) and @obrakmann's range circles for spectators PR (currently open), which should be doing the same checks (there was at least one more but I can't remember now) and let's get this moving.

deniz1a:
How about adding a shared method to OpenRA.Game where these checks are done, and everything else could use that?

obrakmann:
A shared method indeed does make sense. I'll add it to the range circle PR.

deniz1a:
Are these checks needed in OpenRA.Game too or only in Mods.Common?

deniz1a:
They are also used in Mods.RA in #8589...

deniz1a:
Oh ok, every project can use OpenRA.Game, it's the OpenRA namespace ?

penev92:
Yes. You can now use the changes from that PR (`.IsAlliedWith()`).

penev92:
Can you please finish this so we can finally merge it?

deniz1a:
I'm not sure about this anymore. RenderPlayer will probably change (#8647). I'll close this now and look at it later. Also having a shared method which returns current player or player state would probably better than this.

Issue #5249, opened 2014-05-04
Shroud outside map borders should have a different color


deniz1a:
It is difficult to tell where the map edges are without scouting. I think it would be better if the shroud that is outside the map borders had a different color, such as dark gray instead of black.

pchote:
The changes over the last several months to dramatically improve the shroud performance and support TS/RA2 make this unfeasible to implement.  The edge of the map, edges of the rendered shroud, and scissor rectangle do not line up cleanly in the way implied/requested in the mockups above.

bairesearch:
I would recommend always (only if possible; ie on sufficiently large maps) preventing the user from centring their view on a position which would result in them seeing content (shroud) outside of the map. This would require readjusting the view position during zoom-out, if the user is currently zoomed in on a position that is bordering the edge of the map.

PR #8299, opened 2015-05-30
Distinguish between all map cells and cells inside map bounds.

pchote:
This is the next step towards scripted map bounds changes (for missions), and makes the next set of shroud changes a bit cleaner.

The original `Map.Cells` has been renamed to `Map.CellsInsideBounds`, and should be used for anything that cares about the _current_ map state.  `Map.AllCells` covers the entire map (including the cordon cells), and should be used when initializing map state.

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -239,7 +239,8 @@ namespace OpenRA
 		public SequenceProvider SequenceProvider { get { return Rules.Sequences[Tileset]; } }
 
 		public WVec[][] CellCorners { get; private set; }
-		[FieldLoader.Ignore] public CellRegion Cells;
+		[FieldLoader.Ignore] public CellRegion CellsInsideBounds;
+		[FieldLoader.Ignore] public CellRegion AllCells;
 
 		public static Map FromTileset(TileSet tileset)
 		{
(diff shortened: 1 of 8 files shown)


OpenRA.Game/Map/Map.cs
```diff
@@ -392,12 +393,16 @@ void PostInit()
...
+			AllCells = new CellRegion(TileShape, tl, br);
+
+			var btl = new MPos(Bounds.Left, Bounds.Top).ToCPos(this);
+			var bbr = new MPos(Bounds.Right - 1, Bounds.Bottom - 1).ToCPos(this);
+			CellsInsideBounds = new CellRegion(TileShape, btl, bbr);
 
 			CustomTerrain = new CellLayer<byte>(this);
-			foreach (var uv in Cells.MapCoords)
+			foreach (var uv in AllCells.MapCoords)
 				CustomTerrain[uv] = byte.MaxValue;
```
penev92: Is there a reason this is not done inside `Resize()` as well?

pchote: Resize is specific to the legacy editor, which plays by its own rules. This isn't related to this pr, and that method will be going away soon.

penev92:
Ingame map editor broke.
This is what I get when I try to create a new map:


For some reason I can't copy the call stack, so...

pchote:
Fixed. The radar widget didn't validate that cells were inside the footprint it cared about, which lead to a buffer under/overflows.

obrakmann:
Code looks good and I didn't spot any regressions in-game, in either of the editors, or while importing a map. :+1:

Mailaender:
Needs a rebase.

pchote:
Rebased.

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
```diff
@@ -278,7 +278,7 @@ void CacheMapBorderShroud()
 		{
 			// Cache the whole of the map border shroud ahead of time, since it never changes.
 			Func<MPos, bool> mapContains = map.Contains;
-			foreach (var uv in CellRegion.Expand(map.Cells, 1).MapCoords)
+			foreach (var uv in map.AllCells.MapCoords)
```
Mailaender: Shouldn't this be `map.AllCells - map.CellsInsideBounds` then?

Mailaender:
Also can't spot any regressions in OpenRA.Game.exe nor OpenRA.Editor.exe :white_check_mark:

Mailaender:
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Lint/CheckMapCordon.cs also still works.

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
Mailaender: or just `foreach (var uv in CellRegion.Expand(map.AllCells, 1).MapCoords)`?

x-a-n-a-x:
#5249

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
pchote: No, those don't make sense here.  We need to regenerate all the cells to clear any shroud sprites inside the visible bounds (ruling out the first), and the TerrainSpriteLayer can't/doesn't need to store data outside the AllCells area (ruling out the second).  `AllCells` is guaranteed to be at least as large as the old `CellRegion.Expand(map.Cells, 1).MapCoords` because of the >= 1 cell cordon.

penev92:
Well looks like this already has +2, so might as well stop beating around the bush...
Code seems legit and I didn't notice anything obviously wrong ingame :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/ea51742343b8eafe324c76f3e57d18519292bd01...139a5760a535bab13f32c1d40e6b634289b6d84a)

PR #8210, opened 2015-05-21
Veteran unit production & unlocking through spies.

matija-hustic:
A stab at #7585.

~~Sample veteran unit (E1.upgraded) is defined (using GainsExperience trait, InitialExperience property) and its production made available through tech tree prerequisite (barracks.upgraded under Player.ProvidesPrerequisite (1)) that is normally disabled. The prerequisite gets enabled when a spy infiltrates the appropriate building (BARR or TENT) that has the InfiltrateForUpgrade trait which references the prerequisite trait definition (1) under Player.~~

~~In case the idea is OK, I'd remove the yaml entries and the improvised cameo, leaving just the logic, until the upgraded units' cameos are all available.~~

This PR basically introduces three things.
- Upgrading icons in the production queue.
- Spawning upgraded units.
- Triggering both these upgrades with spies.

Issue #7585, opened 2015-03-04
Support (icon) sequence name variant for veterancy.


Phrohdoh:
RA2 needs this for most trainable actors. [RA2 icon list](http://xhp.xwis.net/ra2_icons/).
`e2icon` -> `e2uico`.

WolfGaming:
So this will basically be icon change on upgrade right?

Since we also need to activate an upgrade to instantly increase the veterancy on the produced unit I think that would be the best way.

Phrohdoh:
Correct.

pchote:
Option 2 of #7469 would allow this, but that is still a bit of a hack.

Mailaender:
You may want to solve this with an overlay instead. https://github.com/OpenRA/OpenRA/issues/8210

> many of the RA2 units used beta variants of the cameos for upgraded (Crazy Ivan), was missing (Harrier, defences, navy can also gain veterancy in your logic, this was impossible even in Ares-enhanced YR), or went bogus (Cosmonaut to Laser Rocketeer).

GraionDilach:
Take the overlay included in the RA1 mod - itself being a highre quality remake of the RA2 chevron featured on the RA2 cameos.

Issue #7551, opened 2015-02-26
Replace Rank and Tags and make more use of WithDecoration


penev92:
Basically what @Phrohdoh tried to do for #7466 (as discussed on IRC).
This can be extended and be used for:
 - [x] The spy insignia (#7466)
 - [ ] Unit stance decoration (#3437)
 - [x] Unit rank (#8300)
 - [ ] Adding a faction logo to MCVs (since most games use the same MCVs for all factions, a logo to distinguish what MCV is the one you are looking at might be nice). Same for ConYards.
 - [x] "Unit waiting for carryall", as seen in the original Dune 2000
 - [ ] Sum of all `IDamageModifier`s (shield if < 100% damage, for instance)
 - [ ] Primary Building indicator
 - [ ] "Hero unit" decoration (be it a star, like Mental Omega, or a skull like Tiberium wars), again mentioned in #7466
 - [ ] Fake building indicator

penev92:
>[14:20] <+pchote> the current rank implementation is horrible
[14:20] <+pchote> an artefact from ancient days
[14:20] <+pchote> the first step will be to fix that
[14:21] <+pchote> this probably ties into the (also horrible) pips and tags code
[14:21] < penev> afaik that's what Phrohdoh started with
[14:21] <+pchote> they are all basically the same, except they get drawn in different places

Phrohdoh:
My attempt compiled and ran fine and I threw some debugs in for good measure but nothing 'new' was ever drawn to screen.

:+1: for someone telling me what I was doing wrong or someone doing it properly.

Phrohdoh:
Added ` Sum of all IDamageModifiers (shield if < 100% damage, for instance)` to list of use-cases.

penev92:
Adding the Primary Building "star" tag.

penev92:
Added fake buildings.

pchote:
Refactoring of our existing decorations is complete, and the other ones are probably out of scope.  Closing as fixed.

PR #8300, opened 2015-05-30
Rank effect moved to new trait

matija-hustic:
`Rank` effect moved to new trait that implements `IRender`.

Third item in #7551
Prerequisite for #7585 (PR #8210)


Phrohdoh:
Why are you duplicating `WithDecoration` instead of subclassing it?

matija-hustic:
Three reasons
1) `WithDecoration` does not support freezing a single frame of a sequence,
2) it requires the actor to be in the world and
3) it can only draw decorations that are offset from the top left corner of the actor sprite, whereas the chevron needs to be drawn offset with respect to the bottom right corner.

It'd be impossible to do this with subclassing without first modifying `WithDecoration` and I've been instructed in IRC to implement a new `With*` trait instead of modifying `WithDecoration`.

pchote:
I suggest looking at the `SelectionDecorations` trait and in particular the `UISpriteRenderable` class that was designed specifically to deal with absolute offsets.

matija-hustic:
Now `WithRank` derives from `WithDecoration` and the chevron changes through an upgrade that `GainsExperience` grants.

mods/ra/rules/defaults.yaml
```diff
@@ -11,6 +11,17 @@
 
 ^GainsExperience:
 	GainsExperience:
+	WithRank:
```
pchote: Don't forget the other three mods!

OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
```diff
@@ -42,6 +45,9 @@ public class WithDecorationInfo : UpgradableTraitInfo, ITraitInfo
 		[Desc("Should this be visible to enemy players?")]
 		public readonly bool ShowToEnemies = false;
 
+		[Desc("Should this only be rendered when actor is in world?")]
+		public readonly bool OnlyWhenInWorld = true;
```
pchote: What is the usecase for this?  We generally don't update the actors world position when they are inside transports etc, so you will probably end up with orphaned chevrons in weird places.

OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
```diff
@@ -27,7 +27,10 @@ public class WithDecorationInfo : UpgradableTraitInfo, ITraitInfo
 		[Desc("Palette to render the sprite in. Reference the world actor's PaletteFrom* traits.")]
 		public readonly string Palette = "chrome";
 
-		[Desc("Pixel offset relative to the top-left point of the actor's bounds.")]
+		[Desc("Point in the actor's bounding box used as reference for offsetting the decoration image. 0 for top left, 1 for bottom right.")]
+		public readonly int ReferencePoint = 0;
```
pchote: Please make this an enum instead of using magic numbers.

OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
matija-hustic: I primarily had paradropped units with decorations in mind when writing this. They are not in the world, but still need rendering. How do you propose dealing with this?

pchote: Fixing paradrop (which has always been a bit of a hack) to either use a parachute actor and `WithCargo` or an `ActorPreview` would be two better options.

mods/ra/rules/defaults.yaml
penev92: This is going to need an upgrade rule.

matija-hustic:
Fixed what was commented on. I'll do the upgrade rules then once the code is OK.

mods/ra/rules/defaults.yaml
```diff
@@ -11,6 +11,17 @@
...
+		Image: rank
+		Sequence: rank
+		Palette: effect
+		ReferencePoint: BottomRight
+		Offset: 2, 2
+		OnlyWhenInWorld: false
```
penev92: This doesn't seem to be referencing an existing field.

Mailaender:
The `d2k` and `ts` mod also need a MiniYAML rule update.

matija-hustic:
.yaml files upgraded and rule inserted into `UpgradeRules.cs`.

OpenRA.Mods.Common/Traits/Render/WithRankDecoration.cs
```diff
@@ -0,0 +1,34 @@
...
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using OpenRA.Graphics;
+using OpenRA.Traits;
```
Mailaender: None of these are actually used.

matija-hustic: Gone.

Mailaender:
> .yaml files upgraded and rule inserted into `UpgradeRules.cs`.

You forgot the `ts` mod if I am not mistaken.

matija-hustic:
Ah right @Mailaender, I misread your comment as "`d2k` and `cnc`". Now all the mods are upgraded.

mods/ts/rules/defaults.yaml
```diff
@@ -1,3 +1,20 @@
+^GainsExperience:
+	GainsExperience:
+		ChevronPalette: ra
+		Upgrades:
+			500: firepower, damage, speed, reload
+			1000: firepower, damage, speed, reload, selfheal, eliteweapon
+	WithRankDecoration:
+		Image: rank
+		Sequence: rank
+		Palette: effect
```
Mailaender: This has to be `ra`.

mods/ts/rules/defaults.yaml
```diff
@@ -1,3 +1,20 @@
+^GainsExperience:
+	GainsExperience:
+		ChevronPalette: ra
+		Upgrades:
+			500: firepower, damage, speed, reload
+			1000: firepower, damage, speed, reload, selfheal, eliteweapon
```
Mailaender: I think you are missing `, rank` here.

mods/ts/rules/defaults.yaml
```diff
@@ -1,3 +1,20 @@
+^GainsExperience:
+	GainsExperience:
+		ChevronPalette: ra
```
Mailaender: You removed `ChevronPalette`.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1085,6 +1085,36 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+						{
+							var nodeChPal = ge.Value.Nodes.FirstOrDefault(n => n.Key == "ChevronPalette");
+							var chPal = nodeChPal != null && !string.IsNullOrEmpty(nodeChPal.Value.Value) ? nodeChPal.Value.Value : "effect";
+
+							node.Value.Nodes.Add(new MiniYamlNode("WithRankDecoration", null, new List<MiniYamlNode>
+							{
+								new MiniYamlNode("Image", "rank"),
+								new MiniYamlNode("Sequence", "rank"),
+								new MiniYamlNode("Palette", chPal),
+								new MiniYamlNode("ReferencePoint", "BottomRight"),
```
Mailaender: I think it was `TopLeft` before.

matija-hustic: Are you sure? The `Rank` effect has drawn the chevron in the bottom right corner...

Mailaender: You are right. I was mistaken.

OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
```diff
@@ -15,6 +15,8 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
+	public enum ReferencePoints { TopLeft, BottomRight }
```
obrakmann: Would it be too much to ask for `TopRight` and `BottomLeft`, too?

OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
```diff
@@ -27,7 +29,10 @@ public class WithDecorationInfo : UpgradableTraitInfo, ITraitInfo
 		[Desc("Palette to render the sprite in. Reference the world actor's PaletteFrom* traits.")]
 		public readonly string Palette = "chrome";
 
-		[Desc("Pixel offset relative to the top-left point of the actor's bounds.")]
+		[Desc("Point in the actor's bounding box used as reference for offsetting the decoration image.")]
+		public readonly ReferencePoints ReferencePoint = ReferencePoints.TopLeft;
```
obrakmann: With enums I find it's always nice to mention the possible values in the `Desc`. Maybe that would be better done with a change to the utility's `--docs` command, though, so feel free to ignore my rambling.

OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
matija-hustic: As far as the coding is concerned, not at all. Even stuff like `Center`, `TopCenter`, `BottomCenter`, `CenterLeft` and `CenterRigh` crossed my mind. The only question would be the usefulness of such a thing. Do you have ideas of use cases?

OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
matija-hustic: I'll adhere to whatever the convention is with OpenRA. You just have to tell me what it is :)

OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
obrakmann: Well, there's the planned spy insignia, I'm currently working on adding fuel support, which will have a low-on-fuel decoration, similarly one could think of an out-of-ammo decoration and so on. Also you never know what modders might come up with :)

obrakmann:
Works as advertised, :+1: 

Would be nice if the other positions for decorations were added, but isn't strictly necessary.

obrakmann:
Oh, you should probably squash some of those commits, though.

OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
Phrohdoh: I'd say add it to the desc for now.
Later we can refactor and expand `--docs`.

OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
pchote: These options would be useful for a future PR that could use the same positioning options for pips (default: BottomLeft) and tags (default: TopCenter) too.

matija-hustic: How do you feel about making this field then of type `Tuple<ReferencePointY, ReferencePointX>`? Apart from the FieldLoader addition, it would make for nicer code w.r.t. the conditional branching.

pchote: A bitfield (with the `[Flags]` attribute) would be better than a pair of points.

matija-hustic: I meant a pair of enums, but a bitfield sounds nicer anyways.

matija-hustic:
Added the suggested reference points. Squashed.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1085,6 +1085,40 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+						var upgrades = nodeUpgrades != null ? nodeUpgrades.Value.Nodes.Count() : 4;
+
+						if (upgrades != 0)
+						{
+							foreach (var nodeUpgrade in nodeUpgrades.Value.Nodes)
+								nodeUpgrade.Value.Value = "rank" + (string.IsNullOrEmpty(nodeUpgrade.Value.Value) ? null : ", ") + nodeUpgrade.Value.Value;
+
+							var nodeChPal = ge.Value.Nodes.FirstOrDefault(n => n.Key == "ChevronPalette");
+							var chPal = nodeChPal != null && !string.IsNullOrEmpty(nodeChPal.Value.Value) ? nodeChPal.Value.Value : "effect";
+							ge.Value.Nodes.Remove(nodeChPal);
```
penev92: Perhaps the `ChevronPalette` removal (these 3 lines) should not depend on `upgrades` being `!= 0` (which I find quite unlikely to happen, considering your fallback value of `4`), but instead be moved before the `if`.

matija-hustic: `upgrades == 0` is only possible if the `Upgrades` dictionary is empty. This renders ´GainsExperience` useless and, as far as I can understand, should be equivalent to simply using `-GainsExperience`. I've seen this in `mods/ra/maps/desert-shellmap/map.yaml` and I'm not sure why it is done this way...

So if I get this right, in this case not only that the palette is no longer needed, as you say, but the whole node can be replaced by `-GainsExperience`...

penev92:
Replaces `Rank` with a very simple trait, enhances `WithDecoration` and enhances the YAML for 3 of the official mods by that `^GainsExperience` default :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4d6b0b5f9bfcdd7c25133ba1740eb73457cefaf1...32c0443416b0f91fa92319851f87cd538b174dcc)

PR #8302, opened 2015-05-30
Fix an invalid owner init entry in nod05.

obrakmann:
Interestingly, the game deals with it fine, but the in-game editor crashes when loading it.

@42foobar42 / @alexander-boll: was this just an oversight, or did you have something special in mind with that building?

I guess it's an oversight, since that Owner existed already in the original .ini, but wasn't wired to any trigger, so didn't do anything.

### DIFF
--- a/mods/cnc/maps/nod05/map.yaml
+++ b/mods/cnc/maps/nod05/map.yaml
@@ -229,7 +229,7 @@ Actors:
 		Owner: Neutral
 	Actor66: v32
 		Location: 19,41
-		Owner: Special
+		Owner: Neutral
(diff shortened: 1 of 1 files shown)


Phrohdoh:
:+1:

penev92:
:+1:

Mailaender:
> Interestingly, the game deals with it fine

Well it won't crash. The game will just ignore the actor and not spawn it. I created a lint test to detect those: https://github.com/OpenRA/OpenRA/pull/8308.

42foobar42:
@obrakmann You were right. Just an oversight.

PR #8307, opened 2015-05-31
reorder infantry build palette in TS mod

wenzeslaus:
The order is not the same as the original one (influenced by dependencies) but the main point is to have standard infantry (E1) first, E2 second, ENGINEER third, and the special units on last places. The biggest difference to the original is GDI MEDIC as last (which also applies when both GDI and NOD units are available).


Mailaender:
:+1: Thanks.

Phrohdoh:
:+1: I like it.

Phrohdoh:
I like it. :+1:

PR #8308, opened 2015-05-31
Added a lint check for invalid owners

Mailaender:
To automatically detect the issue resolved by https://github.com/OpenRA/OpenRA/pull/8302.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckPlayers.cs
+++ b/OpenRA.Mods.Common/Lint/CheckPlayers.cs
@@ -51,6 +51,14 @@ namespace OpenRA.Mods.Common.Lint
 				if (map.SpawnPoints.Value.Distinct().Count() != spawnCount)
 					emitError("Duplicate spawn point locations detected.");
 			}
+
+			foreach (var kv in map.ActorDefinitions)
+			{
+				var actorReference = new ActorReference(kv.Value.Value, kv.Value.ToDictionary());
+				var ownerName = actorReference.InitDict.Get<OwnerInit>().PlayerName;
+				if (!playerNames.Contains(ownerName))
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Lint/CheckPlayers.cs
```diff
@@ -51,6 +51,14 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 				if (map.SpawnPoints.Value.Distinct().Count() != spawnCount)
 					emitError("Duplicate spawn point locations detected.");
 			}
+
+			foreach (var kv in map.ActorDefinitions)
+			{
+				var actorReference = new ActorReference(kv.Value.Value, kv.Value.ToDictionary());
+				var ownerName = actorReference.InitDict.Get<OwnerInit>().PlayerName;
+				if (!playerNames.Contains(ownerName))
+					emitWarning("Actor {0} is owned by unknown player {1}.".F(actorReference.Type, ownerName));
```
obrakmann: Already mentioned in the chat, but for completeness' sake: please turn this into an `emitError`, so that Travis will properly flag the build as failed.

Mailaender: Well, I went for error now as there is no excuse for this yaml problem and it crashes the editor (though that problably should get fixed). However `emitError` because Travis is not a reasonable argument. Just make it also exit with != 0 on warnings, maybe using a `--pedantic` flag or do it by default. The linter has almost no false positives.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d64a571c2c072302fa0040d45329e0f75ade160e%5E...d64a571c2c072302fa0040d45329e0f75ade160e)

PR #8310, opened 2015-05-31
Fixed a possible NRE in WRange.TryParse

Mailaender:
Coverity detected that FieldLoader could forward a null value here. This adds extra safety.


chrisforbes:
:+1: looks OK to me.

Unit158:
:+1:

PR #8311, opened 2015-05-31
Fixed a resource leak in PngLoader

Mailaender:
As detected by Coverity. It doesn't really matter though as this was only used in our command line utility.


chrisforbes:
The *caller* should dispose it.

Mailaender:
Updated to dispose in the calling method.

PR #8312, opened 2015-05-31
Fixed a cryptic error message when feeding non-paletted PNGs into the command line utility

Mailaender:
This was detected by Coverity as accessing a null array which will throw. This throws before this can happen with an error message that is easier to understand.


chrisforbes:
:+1: looks good to me.

penev92:
Sounds about right :+1:

PR #8313, opened 2015-05-31
Automatized Coverity Scan with AppVeyor

Mailaender:
Note: this is not final. I just have no other way to test it than file a pull request.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,6 +2,10 @@ version: 1.0.{build}
 
 environment:
   NSIS_ROOT: C:\nsis
+  CoverityProjectToken:
+    secure: BYP5qnptMazSCwTwgiFHlfqZd7BM3bwTRq2Cbs++ofg=
+  CoverityNotificationEmail:
+    secure: BRYcOAGiOqBV406Oh9C/F/bp+OrowEvljq0/CeOddww=
 
 install:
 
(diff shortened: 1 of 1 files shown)


csMACnz:
@Mailaender just looking this over and I see two issues.

First, you are only compressing the files from the current implementation, not publishing them as well.

publish example are https://github.com/GitTools/GitReleaseManager/blob/develop/BuildScripts/default.ps1#L553 and https://github.com/csMACnz/BCLExtensions/blob/master/build.tasks.ps1#L107.

Second, the AppVeyor secure environment variables are not accessible from a PR build. This means your PR cannot actually run the publish from the secure tokens. You have to add conditional code to only run this either on schedules builds (as the two examples above do) or exclude pull requests. Since there is a daily limit to how many reports are submitted to Coverity Scan, I prefer the schedule build version.  This does make setup of Coverity tricky but I recommend using a branch to get it right.

Also, you can set up appveyor against your fork for testing. Just be aware the secure tokens will probably differ.

Mailaender:
> Second, the AppVeyor secure environment variables are not accessible from a PR build.

That explains a lot. Thanks!

csMACnz:
Also FYI I've been working on a fix for the 401 issues I've noticed everyone having. looks like something might have changed on their end but I think i've fixed it now. 
Latest release (0.10.0) should be publishing again.

pchote:
What's the plan here @Mailaender?

Mailaender:
I need to figure out AppVeyor scheduled builds work.

csMACnz:
@Mailaender checkout: http://www.appveyor.com/docs/build-configuration#scheduled-builds

Note that it builds whatever your default branch is.
image

You use crontab timing to set the schedule. I use daily on the 0th minute on the 18th hour (UTC) every day, 
image

Mailaender:
Automatized it as suggested. This will now build, store and published every day on 4 am once merged.

penev92:
That sounds good.
:+1: with my quite limited understanding of the changes.

GraionDilach:
i never really could access the logs of these Coverity scans but any change leading to a more stable code gets a :+1: from me.

Mailaender:
Invited you to https://scan.coverity.com/projects/3650. I actually want this to be publicly available, but that doesn't seem to work.

GraionDilach:
Thank you.

obrakmann:
Looks good to me :+1: 

Note that the scheduled build is at 4am UTC.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c326a9ce388f74de5d20f973708e02443d7945e3%5E...c326a9ce388f74de5d20f973708e02443d7945e3)

PR #8316, opened 2015-05-31
Added missing disposings of the *LoadScreen base class

Mailaender:
`BlankLoadScreen.Dispose` is empty at the moment, so nothing bad has happened yet. Still Coverity marks it as a class hierarchy inconsistency that will lead to errors when expanded.


penev92:
Sounds legit :+1:

obrakmann:
:+1:

PR #8320, opened 2015-05-31
Fixed a possible NRE in FlyAttack.Tick

Mailaender:
For some reason AttackPlane is optional here (TraitOrDefault). Either way we can't check at compile time if the trait dependency is full-filled so another null check will avoid a crash. Discovered by Coverity.


penev92:
:+1:

obrakmann:
:+1:

PR #8321, opened 2015-05-31
Add a lint check for actor upgrades

### LINKED ISSUE #8272: Lint rule to check actor upgrades
https://github.com/OpenRA/OpenRA/wiki/Actor-Upgrades can't be checked at compile time. String types of several independent traits have to match. This is error prone as typos can occur and there will be no warning just silent breakage. A rule could add checks against unused `UpgradeTypes` or undefined `Upgrades` and spit out a warning which may catch at least some sloppy yaml mistakes early.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/17935645-lint-rule-to-check-actor-upgrades?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

penev92:
While this doesn't check whether actor A that grants upgrade X and actor B that uses it can actually interact ingame, it does check whether:
- All upgrades that can be granted by various sources have at least one actor that uses them.
- All upgrades that are consumed by any actor have at least one source to grant them.

It also adds a proxy trait that allows Lua scripts to declare in YAML what custom upgrades they will be using (this enables validation of Lua-granted upgrades).

The main purpose is to close #8272 by checking for unused upgrades, spelling errors and upgrades that nothing can grant.

Closes #8272.

I strongly recommend reviewing per-commit!

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckUpgrades.cs
+++ b/OpenRA.Mods.Common/Lint/CheckUpgrades.cs
@@ -1,0 +1,153 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Lint
+{
+	public class CheckUpgrades : ILintPass
+	{
+		public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
+		{
+			CheckUpgradesValidity(emitError, map);
+			CheckUpgradesUsage(emitError, emitWarning, map);
+		}
+
+		private static void CheckUpgradesValidity(Action<string> emitError, Map map)
+		{
+			var upgradesGranted = GetAllGrantedUpgrades(emitError, map).ToHashSet();
+
+			foreach (var actorInfo in map.Rules.Actors)
+			{
+				foreach (var trait in actorInfo.Value.Traits)
+				{
+					var fields = trait.GetType().GetFields();
+					foreach (var field in fields.Where(x => x.HasAttribute<UpgradeUsedReferenceAttribute>()))
+					{
+						var values = LintExts.GetFieldValues(trait, field, emitError);
+						foreach (var value in values.Where(x => !upgradesGranted.Contains(x)))
+							emitError("Actor type `{0}` uses upgrade `{1}` that is not granted by anything!".F(actorInfo.Key, value));
+					}
+				}
+			}
+		}
+
+		private static void CheckUpgradesUsage(Action<string> emitError, Action<string> emitWarning, Map map)
+		{
+			var upgradesUsed = GetAllUsedUpgrades(emitError, map).ToHashSet();
+
+			// Check all upgrades granted by traits.
+			foreach (var actorInfo in map.Rules.Actors)
+			{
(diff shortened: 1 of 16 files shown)


Phrohdoh:
You don't need the !Any check as the following foreach would be empty and not execute.

OpenRA.Mods.Common/Lint/CheckUpgrades.cs
```diff
@@ -0,0 +1,96 @@
...
+
+		private static void CheckUpgradesValidity(Action<string> emitError, Map map)
+		{
+			var upgradesGranted = GetAllGrantedUpgrades(map).ToHashSet();
+
+			foreach (var actor in map.Rules.Actors)
+			{
+				var upgradesUsed = actor.Value.Traits.WithInterface<UpgradableTraitInfo>().SelectMany(x => x.UpgradeTypes).ToHashSet();
+				if (!upgradesUsed.Any())
+					continue;
```
penev92: Redundant `Any()` check.
Also that means redundant `.ToHashSet()`.

mods/d2k/rules/misc.yaml
```diff
@@ -102,10 +102,6 @@ crate:
 		SelectionShares: 0
 		NoBaseSelectionShares: 9001
 		Units: mcv
-	GrantUpgradeCrateAction@cloak:
-		SelectionShares: 5
-		Effect: cloak
-		Upgrades: cloak
```
Mailaender: You may also fix it the other way round, because this was added on purpose in https://github.com/OpenRA/OpenRA/pull/6856. It also proofs my point about the weak spot about https://github.com/OpenRA/OpenRA/wiki/Actor-Upgrades. They are easy to do wrong without anyone noticing. This lint checks are long overdue. Thanks!

pchote:
This probably should be using attributes like our existing `[ActorReference]`, `[WeaponReference]`, `[SequenceReference]`s.

penev92:
Added field attributes.

OpenRA.Mods.Common/Scripting/Properties/UpgradeProperties.cs
```diff
@@ -17,35 +18,41 @@ namespace OpenRA.Mods.Common.Scripting
...
 		{
 			um = self.Trait<UpgradeManager>();
+			validUpgrades = self.World.WorldActor.TraitOrDefault<ScriptUpgradesCache>();
 		}
 
 		[Desc("Grant an upgrade to this actor.")]
 		public void GrantUpgrade(string upgrade)
 		{
-			um.GrantUpgrade(Self, upgrade, this);
+			if (validUpgrades.Info.Upgrades.Contains(upgrade))
```
Phrohdoh: I feel we ought to retain the failing logic here in the case of a bogus upgrade, instead of failing silently.
If the author does not run a lint pass that is their fault and should be reported to them (via crashes).

penev92:
Made granting/revoking of upgrades from a Lua script throw if the `ScriptUpgradeCache` isn't set up properly.

Mailaender:
This depends on https://github.com/OpenRA/OpenRA/pull/8412 for the LintExt just like #7709.

penev92:
Rebased.

obrakmann:
Looks ok to me, aside from that one spurious change to the project. :+1:

penev92:
Updated.

Mailaender:
Thanks.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/b232a7de8e2b4c3c694d5e3daba3c0242961ac4a...775dd24b2745364201bf2d4c29220d3519fd6e3a)

PR #7331, opened 2015-01-13
TerrainModifiesDamage trait.

phrohdoh:
^Infantry in D2k now take 80% damage if they are standing on Rough terrain.

Closes #7328 
Even if this doesn't belong/exist in D2k it will be nice to have.

I chose to allow negative percentages because if someone wants an actor healed when damaged on a specific tile they should have the ability to do so.

/cc @MicroJOo and @penev92

Note: I won't integrate this with #7110 as they have different purposes.

PR #8322, opened 2015-05-31
Fixed an explicit NRE in TerrainModifiesDamage

Mailaender:
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Activities/Demolish.cs#L70 sets warhead to null so it will crash here. I believe this is a real vulnerability introduced by https://github.com/OpenRA/OpenRA/pull/7331.


penev92:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c689e63efb81b5f7ba0c773728b8fbf1ff7c961e%5E...c689e63efb81b5f7ba0c773728b8fbf1ff7c961e)

PR #7112, opened 2014-12-10
Bogus yaml checking

MunWolf:
Allows classes that implement ITraitInfo to specify a method that is called after the Info class has been made and checks for some TraitSpecific stuff like if two arrays should be the same size.

PR #8325, opened 2015-05-31
Fixed a division through zero

Mailaender:
This is probably just a theoretical problem as zero occupied cells generally doesn't make much sense.

### DIFF
--- a/OpenRA.Mods.D2k/Traits/Buildings/DamagedWithoutFoundation.cs
+++ b/OpenRA.Mods.D2k/Traits/Buildings/DamagedWithoutFoundation.cs
@@ -52,7 +52,10 @@ namespace OpenRA.Mods.D2k.Traits
 					safeTiles++;
 			}
 
-			damageThreshold = (info.DamageThreshold * health.MaxHP + (100 - info.DamageThreshold) * safeTiles * health.MaxHP / totalTiles) / 100;
+			if (totalTiles > 0)
+				damageThreshold = (info.DamageThreshold * health.MaxHP + (100 - info.DamageThreshold) * safeTiles * health.MaxHP / totalTiles) / 100;
+			else
+				damageThreshold = health.HP;
 
 			// Actors start with maximum damage applied
 			var delta = health.HP - damageThreshold;


Unit158:
Considering zero occupied cells is bogus, we should actually crash. :-1: for adding the check and not throwing an exception if it is false.

Mailaender:
That would then throw an exception for things like mines which use `Immobile.OccupiesSpaces = false`.

Unit158:
In this specific case, it should be throwing if it is dividing by a negative number or zero. I don't care how it is done, but the check should not cause silent failures.

Mailaender:
A division through zero exception is never the proper way to throw.

OpenRA.Mods.D2k/Traits/Buildings/DamagedWithoutFoundation.cs
```diff
@@ -52,7 +52,8 @@ public void AddedToWorld(Actor self)
 					safeTiles++;
 			}
 
-			damageThreshold = (info.DamageThreshold * health.MaxHP + (100 - info.DamageThreshold) * safeTiles * health.MaxHP / totalTiles) / 100;
+			if (totalTiles > 0)
```
Unit158: This will now cause the actor to lose 100 health on spawn if they have no cells occupied. You should add an else statement that throws an InvalidArgumentException.

penev92:
Just do an `else damageThreshold = health.HP;`, which would make actors that don't occupy any cells not take any damage because of the check in `Tick()`.

Mailaender:
I went for the else instead of crashing on purpose. Game engine should be forgiving imho. You could write a lint rule that spits out a warning that the setups doesn't make much sense in this case.

OpenRA.Mods.D2k/Traits/Buildings/DamagedWithoutFoundation.cs
```diff
@@ -52,7 +52,10 @@ public void AddedToWorld(Actor self)
...
+				damageThreshold = (info.DamageThreshold * health.MaxHP + (100 - info.DamageThreshold) * safeTiles * health.MaxHP / totalTiles) / 100;
+			else
+				damageThreshold = health.HP;
```
Unit158: Now they'll die.

penev92: ?

Unit158: Look at how the code works.

WolfGaming: No they wont.
```
			var delta = health.HP - damageThreshold;
			if (delta > 0)
				health.InflictDamage(self, self.World.WorldActor, delta, null, false);
```
the delta is the damage that will be taken so if damageThreshold is equal to health.HP then delta will be 0 and no damage will happen.

Unit158:
:+1: after a lint check for occupiedcells > 0 is made.

Mailaender:
That is out of scope for this pull request.

Unit158:
Well, you must make a follow-up then. I'm not going to condone bad behaviour, this engine already fails silently too much of the time. :|

Mailaender:
That would delay this pull request unnecessarily as it depends on the abandoned https://github.com/OpenRA/OpenRA/pull/7112 then.

penev92:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/0f70e2aa8a6acec0dbb9333134a31188eea06ae3%5E...0f70e2aa8a6acec0dbb9333134a31188eea06ae3)

PR #8328, opened 2015-05-31
Made the tech description label optional

Mailaender:
It is inconsistent to use `GetOrNull` and then crash on the null state which was picked up by Coverity.


penev92:
:+1:

obrakmann:
It's a bit silly to have a tech level dropdown without a description. Personally, I'd have changed the `GetOrNull` to just `Get`. Anyway, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9dfcbb95a300e2aa0a880815d3d8b6ffb86321e7%5E...9dfcbb95a300e2aa0a880815d3d8b6ffb86321e7)

Mailaender:
You can also have it self-explanatory with `{ "Techlevel 1", "Techlevel 2" }` but yes I agree those Coverity fixes are not that too impressive so I just hope to get through it quickly.

PR #8330, opened 2015-05-31
Fixed a NullReferenceException in GameTimeLogic

Mailaender:
The `GAME_TIMER` label was meant to be optional as seen in the code above. Never crashed, because everyone copy & pasted the RA chrome MiniYaml definitions.


obrakmann:
:+1:

PR #8331, opened 2015-05-31
Added more useful exceptions to the command line utility replay parser

Mailaender:
This was picked up by Coverity as a "Dereference null return value" violation. As NREs are often cryptic to read by users I decided to add more appropriate ones with additional context.


obrakmann:
:+1:

PR #8332, opened 2015-05-31
Fixed an explicit NullReferenceException in ExtractLanguageStringsCommand

Mailaender:
I should probably give this a second go now that we have more possibly interested non-native English speaking developers for testing and review.


chrisforbes:
:+1:

PR #8333, opened 2015-05-31
Fixed an unintentional integer overflow in AttackBase

Mailaender:
Not sure how bad this is, but it surely doesn't sound good. http://cwe.mitre.org/data/definitions/190.html


Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/8338.

PR #8334, opened 2015-05-31
Fixed an unintentional integer overflow in Aircraft.GetRepulsionForce

Mailaender:
Detected by Coverity.


pchote:
All these theoretical integer overflows come from the same code pattern.  I recommend introducing a `(long)WRange.RangeSquared` method to simplify all these cases.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/8338.

PR #8335, opened 2015-05-31
Fixed a missing call to base clase in OpenRA.Editor.OnPaint

Mailaender:
This doesn't seem to fix anything. Still it was violating the API contract.


chrisforbes:
Sure, the base class's OnPaint does nothing, but this is harmless.

chrisforbes:
:+1:

PR #8337, opened 2015-05-31
Simplify and improve shroud rendering performance.

pchote:
This pr uses #8221 and the supporting changes in #8226 and #8299 to dramatically simplify the shroud rendering code.  This should also give a significant performance boost.  Only the last commit is part of the actual pr.

### DIFF
--- a/OpenRA.Game/Map/CellLayer.cs
+++ b/OpenRA.Game/Map/CellLayer.cs
@@ -19,6 +19,7 @@ namespace OpenRA
 	public class CellLayer<T> : IEnumerable<T>
 	{
 		public readonly Size Size;
+		readonly Rectangle bounds;
 		public readonly TileShape Shape;
 		public event Action<CPos> CellEntryChanged = null;
 
@@ -30,6 +31,7 @@ namespace OpenRA
 		public CellLayer(TileShape shape, Size size)
 		{
 			Size = size;
+			bounds = new Rectangle(0, 0, Size.Width, Size.Height);
 			Shape = shape;
 			entries = new T[size.Width * size.Height];
 		}
@@ -121,6 +123,11 @@ namespace OpenRA
 		IEnumerator IEnumerable.GetEnumerator()
 		{
 			return GetEnumerator();
+		}
+
+		public bool Contains(MPos uv)
+		{
+			return bounds.Contains(uv.U, uv.V);
 		}
 	}
 
--- a/OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
+++ b/OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
@@ -10,7 +10,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Drawing;
+using System.Drawing.Imaging;
+using System.IO;
+using System.Linq;
 using OpenRA.Graphics;
+using OpenRA.Primitives;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
@@ -43,7 +48,7 @@ namespace OpenRA.Mods.Common.Traits
 		public object Create(ActorInitializer init) { return new ShroudRenderer(init.World, this); }
 	}
 
-	public class ShroudRenderer : IRenderShroud, IWorldLoaded
+	public sealed class ShroudRenderer : IRenderShroud, IWorldLoaded, INotifyActorDisposing
 	{
 		[Flags]
 		enum Edges : byte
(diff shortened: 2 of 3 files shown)


LipkeGu:
is the heightmap shroud fix also included? :)

pchote:
No.

penev92:
All dependencies have been merged.

pchote:
Rebased.

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
```diff
@@ -170,14 +163,28 @@ public void WorldLoaded(World w, WorldRenderer wr)
...
+			var fogSheet = fogSprites[0].Sheet;
+			if (fogSprites.Any(s => s.Sheet != fogSheet))
+				throw new InvalidDataException("Fog sprites span multiple sheets. Try loading their sequences earlier.");
+
+			var fogBlend = fogSprites[0].BlendMode;
+			if (fogSprites.Any(s => s.BlendMode != fogBlend))
+				throw new InvalidDataException("Fog sprites must all use the same blend mode.");
+
+			shroudLayer = new TerrainSpriteLayer(w, wr, shroudSheet, shroudBlend, wr.Palette(info.ShroudPalette));
+			fogLayer = new TerrainSpriteLayer(w, wr, fogSheet, shroudBlend, wr.Palette(info.FogPalette));
```
penev92: Something tells me this wants to be `fogBlend` instead of `shroudBlend`.

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
```diff
@@ -209,177 +216,71 @@ Edges GetEdges(MPos uv, Func<MPos, bool> isVisible)
 			return info.UseExtendedIndex ? edge ^ ucorner : edge & Edges.AllCorners;
 		}
 
-		public void RenderShroud(WorldRenderer wr, Shroud shroud)
+		void UpdateCells(IEnumerable<CPos> cells)
```
penev92: If I understand this correctly, this method just marks the given cells as dirty. If so, maybe a better name is in order? Or at least a `<summary>`. We should make more use of those.

penev92:
Well, the shroud in TS is different...


Code seems legit and works well at first glance, so :+0.5: for now.

Mailaender:
I can't measure any noticeable performance improvement. Tested the Intervention mission.

Can't spot any rendering regressions (or improvements, the TS glitches just look different now). :white_check_mark:

pchote:
Updated.  That top-of-map cutoff happens because the current set of TS maps don't have the correct cordon setup.  The next prs will improve this.

pchote:
I found a couple more issues that will need to be fixed.

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
```diff
@@ -389,5 +290,16 @@ Sprite GetSprite(Sprite[] sprites, Edges edges, int variant)
 
 			return sprites[variant * variantStride + edgesToSpriteIndexOffset[(byte)edges]];
 		}
+
+		bool disposed;
```
RoosterDragon: You don't need this protective flag in this case.

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
```diff
@@ -389,5 +290,16 @@ Sprite GetSprite(Sprite[] sprites, Edges edges, int variant)
...
+		public void Disposing(Actor self)
```
RoosterDragon: I think we might want to consider implementing the usual ```IDisposable``` interface in this case, since we have actual disposable fields, and have this method simply delegate to it.

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
```diff
@@ -209,177 +216,71 @@ Edges GetEdges(MPos uv, Func<MPos, bool> isVisible)
...
-				var location = tileInfo.ScreenPosition - 0.5f * size;
-				OpenRA.Graphics.Util.FastCreateQuad(
-					vertices, location + sprite.FractionalOffset * size,
-					sprite, palette.TextureIndex, offset, size);
-			}
-
-			spriteLayer[uv] = sprite;
-			return sprite;
+			fogLayer.Draw(wr.Viewport);
+			shroudLayer.Draw(wr.Viewport);
```
RoosterDragon: It used to be that fog was rendered after the shroud layer. Might not actually affect things in practise but might be worth keeping that ordering.

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
```diff
@@ -209,177 +216,71 @@ Edges GetEdges(MPos uv, Func<MPos, bool> isVisible)
...
+					shroud.CellsChanged += DirtyCells;
 
-			if (currentShroud != null)
-			{
-				mapBorderShroudIsCached = false;
-
-				// We need to mark newly dirtied areas of the shroud.
-				// Expand the dirty area to cover the neighboring cells, since shroud is affected by neighboring cells.
-				foreach (var cell in cellsDirty)
+					// Needs the anonymous function to ensure the correct overload is chosen
```
RoosterDragon: This comment doesn't seem to be true.

pchote: Mono gives runtime errors if I assign `currentShroud.IsExplored` directly.

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
pchote: That ordering was a bit bogus, so I think its worth fixing it here.

OpenRA.Mods.Common/Traits/World/ShroudRenderer.cs
pchote: I don't see the benefit of implementing all the disposable plumbing here when it will never be used.

OpenRA.Game/Map/CellLayer.cs
```diff
@@ -122,6 +122,12 @@ IEnumerator IEnumerable.GetEnumerator()
 		{
 			return GetEnumerator();
 		}
+
+		public bool Contains(MPos uv)
+		{
+			return new Rectangle(0, 0, Size.Width, Size.Height)
```
obrakmann: Is it worth it to store that rectangle somewhere so you save yourself the object creation for every cell in every tick?

obrakmann:
Code looks fine to me, didn't notice any regressions either. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/5467fd47e64b8c1d11d89c745605b583c4ac2e88%5E...5467fd47e64b8c1d11d89c745605b583c4ac2e88)

PR #8338, opened 2015-05-31
Fixed all squared range related integer overflows

Mailaender:
Even those not detected by Coverity.


chrisforbes:
:+1:

penev92:
These are some pretty small cuts, but this still reminded me of the "death by a thousand cuts", regarding our performance.
`WRange.RangeSquared` is calculated on each call, and it is being called from inside a lot of LINQs.
In at least half of the places you can store it in a local variable before the LINQ to avoid doing the same calculation 50 times.

Mailaender:
I expect the JIT runtime to cache those by itself.

PR #8339, opened 2015-05-31
Step in the direction of RA2 paratroopers

### LINKED ISSUE #2988: Infantry on parachutes makes idle actions in the air.
From a comment on http://www.moddb.com/games/openra/news/playtest-20130407

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/141903-infantry-on-parachutes-makes-idle-actions-in-the-air?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

matija-hustic:
`Parachute` effect replaced by an in-world actor implementing the `Parachute` activity. That activity grants a `parachute` upgrade, changes the actor's targetability and moves the actor towards the ground. Once on the ground, the `parachute` upgrade is revoked.

The `parachute` upgrade is consumed by `WithParachute`, an upgradable trait that renders the parachute (similar to `WithIdleOverlay`) and the shadow (using `IRender`).

Parachuted actors are made uncrushable and unable to crush.

Two further small adjustments have been made to the `SetPosition` placement functions on `Mobile` and `Crate` traits.

Edit: Fixes #2988.


OpenRA.Mods.Common/Activities/Parachute.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace OpenRA.Mods.Common.Activities
+{
+	public class Parachute : Activity
+	{
+		WPos currentPosition;
+		WVec fallVector;
+
+		public Parachute(Actor self, WPos dropPosition)
+		{
+			var um = self.TraitOrDefault<UpgradeManager>();
```
pchote: In general, activities may be queued and not run until an arbitrary time after they are constructed.  It is therefore best practice to not do anything that changes world state from their constructor.  You can call these from the first `Tick()`.

matija-hustic: Fixed locally. I'll push after other remarks.

LipkeGu:
we need this for TS to o.o" -> drop pods

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -573,6 +576,9 @@ public bool CanMoveFreelyInto(CPos cell, Actor ignoreActor = null, bool checkTra
 
 		public void EnteringCell(Actor self)
 		{
+			// Only make actor crush if it is on the ground
+			if (self.CenterPosition.Z != 0)
+				return;
```
abcdefg30: Can we have a new line after this?

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -581,6 +587,9 @@ public void EnteringCell(Actor self)
 
 		public void FinishedMoving(Actor self)
 		{
+			// Only make actor crush if it is on the ground
+			if (self.CenterPosition.Z != 0)
+				return;
```
abcdefg30: Same here.

OpenRA.Mods.Common/Traits/Crates/Crate.cs
```diff
@@ -125,8 +125,39 @@ public void Tick(Actor self)
...
+		{
+			CenterPosition = pos;
+			if (self.IsInWorld)
+			{
+				self.World.ScreenMap.Update(self);
+				self.World.ActorMap.UpdatePosition(self, this);
+			}
+		}
+
+		private void SetLocation(Actor self, CPos cell, SubCell subCell = SubCell.Any)
```
abcdefg30: Seems like you forgot to comment here as you did for the other methods?

OpenRA.Mods.Common/Traits/Crates/Crate.cs
```diff
@@ -155,21 +186,11 @@ public bool CanEnterCell(CPos a, Actor ignoreActor = null, bool checkTransientAc
...
-				self.World.ActorMap.UpdatePosition(self, this);
-				self.World.ScreenMap.Update(self);
-			}
-		}
-
 		public bool CrushableBy(string[] crushClasses, Player owner)
 		{
+			// Crate can only be crushed if it is not in the air or underground
+			if (CenterPosition.Z != 0)
+				return false;
```
abcdefg30: And maybe a new line here too?

OpenRA.Mods.Common/Activities/Parachute.cs
```diff
@@ -0,0 +1,77 @@
...
+			if (um != null)
+				um.GrantUpgrade(self, "parachute", this);
+
+			var tpu = self.TraitOrDefault<TargetableParachutableUnit>();
+			if (tpu != null)
+				tpu.IsParachuting = true;
+
+			var p = self.Info.Traits.GetOrDefault<ParachutableInfo>();
+			if (p != null)
+				fallVector = new WVec(0, 0, p.FallRate);
```
abcdefg30: Can we get a default for this or require the `Parachutable` trait?
If a unit doesn't have the `Parachutable` trait it'll just stay in air all the time (in this PR).

penev92: This is in `Activities/Parachute`, so it should require the `Parachutable` trait.
Otherwise we can have a `Freefall` activity for actors that don't have it :D

abcdefg30:
You should also set the paradrop-traits at `^Tank` in ra up or remove them from `^Tank` as we don't have the `ArmorDrop` support power anymore.

abcdefg30:
We should also fix infantry going prone in air and the death animation somewhen.

OpenRA.Mods.Common/Traits/Crates/Crate.cs
matija-hustic: I didn't forget, I just thought clarification wasn't necessary with this function. I'll add the comments tho.

OpenRA.Mods.Common/Activities/Parachute.cs
matija-hustic: So do I seriously go for this `Freefall`? :D

matija-hustic:
Fixed as commented and rebased. Only remains to see about
* the `Parachutable` trait being a prerequisite / adding the `Freefall` activity and
* the paratrooper animation (not going prone + death).

Maybe I can some more input on what would be the right way to go with this PR and what should be left for an additional PR?

GraionDilach:
I did a quick glampse on the code and while I do appreciate this there's one thing I noticed:

RA2 parachute has a closing sequence as well. It doesn't disappear like the RA1 one, but fall to the ground then play a decayanim. Notsure how to implement this via the upgrade tho.

OpenRA.Mods.Common/Traits/Render/WithParachute.cs
```diff
@@ -0,0 +1,116 @@
...
+	[Desc("Renders a parachute on units.")]
+	public class WithParachuteInfo : UpgradableTraitInfo, ITraitInfo, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>, Requires<IBodyOrientationInfo>
+	{
+		[Desc("The image that contains the prachute sequences.")]
+		public readonly string Image = "parach";
+
+		[Desc("Prachute opening sequence.")]
+		public readonly string OpeningSequence = "open";
+
+		[Desc("Prachute idle sequence.")]
```
GraionDilach: Typo (there's also one at ShadowSequence's description along with these two).

matija-hustic:
@GraionDilach, it isn't too problematic to implement the closing sequence. The only question is - seeing how the closing sequence needs to play while the unit is already able to move around, should the animation move with him or play out on the drop location?

GraionDilach:
In RA2 it moved along with it, it was attached to the unit like the EMP or the mindcontrol animation.

(Actually, WW even had a reason to do that although they made the anim entirely to work exactly with their flying height. The unit had a doesitstillexist check upon the parachute anim and the result was used to decide between applying [General]ParachuteMaxFallRate or NoParachuteMaxFallRate, but since FallRate is directly customizable here, your solution is better than theirs.)

penev92:
I think moving to the north would look bad if the parachute moves with the infantry (and keeps its relative position), so probably drop the parachute when you hit the ground (a.k.a. play it where they landed). But that would be trickier at best, on top of having people go "that's not what the original did!", so...

matija-hustic:
So which one should it be? With the current code it would follow the unit.

GraionDilach:
I'd say current will do in this case.

OpenRA.Mods.Common/Traits/Crates/Crate.cs
Phrohdoh: No explicit private.

OpenRA.Mods.Common/Activities/Parachute.cs
```diff
@@ -0,0 +1,96 @@
...
+			this.dropPosition = dropPosition;
+		}
+
+		Activity FirstTick(Actor self)
+		{
+			triggered = true;
+
+			var um = self.TraitOrDefault<UpgradeManager>();
+			if (um != null)
+				um.GrantUpgrade(self, "parachute", this);
```
obrakmann: It would be nice to un-hardcode the name of the upgrade and make it into a field in `ParachutableInfo`.

Also, you need the `UpgradeManager` twice, so it's probably already worth caching it in the constructor.

OpenRA.Mods.Common/Activities/Parachute.cs
```diff
@@ -0,0 +1,96 @@
...
+
+			var tpu = self.TraitOrDefault<TargetableParachutableUnit>();
```
obrakmann: Please cache this, too

OpenRA.Mods.Common/Activities/Parachute.cs
```diff
@@ -0,0 +1,96 @@
...
+			if (tpu != null)
+				tpu.IsParachuting = true;
+
+			// Parachutable trait is a prerequisite for running this activity
+			var p = self.Info.Traits.Get<ParachutableInfo>();
+			fallVector = new WVec(0, 0, p.FallRate);
+
+			// Place the actor and retrieve its visual position (CenterPosition)
+			self.Trait<IPositionable>().SetPosition(self, dropPosition);
```
obrakmann: and `IPositionable` as well

OpenRA.Mods.Common/Traits/ParaDrop.cs
```diff
@@ -82,7 +82,11 @@ public void Tick(Actor self)
 			droppedAt.Add(self.Location);
 
 			var a = cargo.Unload(self);
-			self.World.AddFrameEndTask(w => w.Add(new Parachute(a, self.CenterPosition)));
+			self.World.AddFrameEndTask(w =>
+			{
+				w.Add(a);
+				a.QueueActivity(new Activities.Parachute(a, self.CenterPosition));
```
obrakmann: You can drop the `Activities` here, you're already including the namespace.

OpenRA.Mods.Common/Traits/EjectOnDeath.cs
```diff
@@ -66,7 +66,11 @@ public void Killed(Actor self, AttackInfo e)
 			{
 				if (cp.Z > 0)
 				{
-					self.World.AddFrameEndTask(w => w.Add(new Parachute(pilot, cp)));
+					self.World.AddFrameEndTask(w =>
+					{
+						w.Add(pilot);
+						pilot.QueueActivity(new Activities.Parachute(pilot, cp));
```
obrakmann: I think you can clean up the `using`s here, too, and then remove the `Activities`.

Bonus points if you add a `Desc()` that states the traits the `PilotActor` needs in the Info class.

obrakmann:
And I'm afraid these changes call for an upgrade rule.

matija-hustic:
I'll put the upgrade rule and rest of YAML changes at the very end as @Mailaender once suggested. Otherwise I'll probably need to rebase alot.

Now also prevents paratroopers from idling while in air.

OpenRA.Mods.Common/Activities/Parachute.cs
```diff
@@ -0,0 +1,99 @@
...
+
+namespace OpenRA.Mods.Common.Activities
+{
+	public class Parachute : Activity
+	{
+		WPos dropPosition;
+		WPos currentPosition;
+		WVec fallVector;
+		bool triggered = false;
+		UpgradeManager um;
```
obrakmann: You can make all of these cached traits `readonly`. (And iirc you'll have to move them to the top then to conform to the coding style. Run StyleCop to see if it complains about anything.)

OpenRA.Mods.Common/Traits/EjectOnDeath.cs
```diff
@@ -8,6 +8,7 @@
  */
 #endregion
 
+using OpenRA.Mods.Common.Activities;
 using OpenRA.Mods.Common.Effects;
```
obrakmann: Unless I missed something, I think you don't need the `Effects` anymore.

OpenRA.Mods.Common/Traits/Render/WithParachute.cs
```diff
@@ -0,0 +1,133 @@
...
+			var offset = body.LocalToWorld(Offset.Rotate(orientation));
+			yield return new SpriteActorPreview(anim, offset, offset.Y + offset.Z + 1, p, rs.Scale);
+		}
+	}
+
+	public class WithParachute : UpgradableTrait<WithParachuteInfo>, IRender
+	{
+		Animation shadow;
+		AnimationWithOffset anim;
+		WithParachuteInfo info;
```
obrakmann: These can all be `readonly`, too.

OpenRA.Mods.Common/Traits/Render/WithInfantryBody.cs
```diff
@@ -113,16 +113,17 @@ public virtual void Tick(Actor self)
...
-				DefaultAnimation.PlayFetchIndex(NormalizeInfantrySequence(self, Info.StandSequences.Random(Game.CosmeticRandom)), () => 0);
-			}
-			else if ((state != AnimationState.Moving || dirty) && move.IsMoving)
+			if ((state != AnimationState.Moving || dirty) && move.IsMoving)
 			{
 				state = AnimationState.Moving;
 				DefaultAnimation.PlayRepeating(NormalizeInfantrySequence(self, Info.MoveSequence));
 			}
+			else if (((state == AnimationState.Moving || dirty) && !move.IsMoving)
+			 || ((state == AnimationState.Idle || state == AnimationState.IdleAnimating) && !self.IsIdle))
```
obrakmann: Looks like a space instead of a tab at the beginning of the line.

obrakmann:
Aside from those codestyle niggles, this is looking otherwise fine to me. I'll give my :+1: when the upgrade rules are added.

matija-hustic:
* Replaced `TargetableParachutableUnit` with an upgradable `TargetableUnit`
* Added an upgrade rule
* Upgraded all mods
* Rebased and squashed

OpenRA.Mods.Common/Traits/TargetableUnit.cs
```diff
@@ -52,5 +63,26 @@ public virtual IEnumerable<WPos> TargetablePositions(Actor self)
...
+		}
+
+		public void UpgradeLevelChanged(Actor self, string type, int oldLevel, int newLevel)
+		{
+			if (newLevel > 0)
+			{
+				if (info.UpgradedTargetTypes.Contains("!"))
+					targetTypes = info.UpgradedTargetTypes.Where(utt => !utt.StartsWith("!")).ToArray();
+				else
+					targetTypes = targetTypes.Except(info.UpgradedTargetTypes.Where(utt => utt.StartsWith("!")).Select(utt => utt.Substring(1)))
```
abcdefg30: Sorry if I missunderstood something here, but aren't you checking for `info.UpgradedTargetTypes.Contains("!")` above, so `info.UpgradedTargetTypes.Where(utt => utt.StartsWith("!")).Select(...` is redundant here?
(As something that doesn't contain a "!" shouldn't start with it :p)

matija-hustic: `info.UpgradedTargetTypes` is a collection of strings, `utt` is a string from this collection

`info.UpgradedTargetTypes.Contains()` checks if a string is present in the collection
`utt.StartsWith()` checks if a string begins by a substring

:)

obrakmann: No, that's correct. The first `if` checks whether a *single* ! exists in `UpgradedTargetTypes`. In that case, `TargetTypes` will be *replaced* with all entries in `UpgradedTargetTypes` that don't start with an !.

The `else` case instead *modifies* the existing `TargetTypes` and removes all entries that start with an !, and adds all others.

obrakmann:
Looks ok to me, and works as expected in-game. :+1:

OpenRA.Mods.Common/Traits/TargetableUnit.cs
```diff
@@ -20,20 +21,30 @@ public class TargetableUnitInfo : ITraitInfo, ITargetableInfo
 		public readonly string[] TargetTypes = { };
 		public string[] GetTargetTypes() { return TargetTypes; }
 
+		[Desc("Upgrades that, when granted, trigger target type modification.")]
+		public readonly string[] UpgradeTypes = { };
+
+		[Desc("List of targets to add or remove when at least one upgrade is granted. ",
+			"Prefix target types to remove with an exclamation point (!). ",
+			"Use the exclamation point alone to remove all target types.")]
```
pchote: We tend to avoid magic values where possible, so lets make this a full list of types instead, please.  You could default this to null, and then have it fall back to TargetTypes in that case.

reaperrr:
Needs a rebase.

matija-hustic:
Changed as commented. Rebased.

GraionDilach:
I don't like the TargetableUnit upgrade implementation. It's okay for this particular case, but it limits us in the long run. There will be cases when multiple TargetType upgrades each having separate effects will be desired. (RA2 paratroopers upgrade the TargetType to air-only while Chrono Legionnaires should upgrade it to Temporal-only, reimplementing the tiberiumheal and tiberiumproof veteraneffects in TS due to both effects rely on TargetTypes in OpenRA to apply their effects as well).

It should be refactored to allow using IdentifierSuffixes in a separate PR.

OpenRA.Mods.Common/Traits/Render/WithParachute.cs
```diff
@@ -0,0 +1,134 @@
...
+	public class WithParachuteInfo : UpgradableTraitInfo, ITraitInfo, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>, Requires<IBodyOrientationInfo>
+	{
+		[Desc("The image that contains the parachute sequences.")]
+		public readonly string Image = "parach";
+
+		[Desc("Parachute opening sequence.")]
+		public readonly string OpeningSequence = "open";
+
+		[Desc("Parachute idle sequence.")]
+		public readonly string Sequence = "idle";
```
Mailaender: You forgot all the `[SequenceReference("Image")]` during the refactor. See https://github.com/openra/openra/pull/7709.

matija-hustic:
Fixed as @Mailaender commented. Rebased.

@GraionDilach, I'm not sure I followed what you meant with you remark on `TargetableUnit`, but if it can be done in a separate PR, as you said, I'd say go for it.

Mailaender:
This spits out a lot of MiniYaml errors and warnings: https://travis-ci.org/OpenRA/OpenRA/builds/68995836#L1177

matija-hustic:
Does d2k use parachuting units?

Mailaender:
Yes, but without parachute sequence in the Carryall drop.

matija-hustic:
Is that how things ought to be or do they need a parachute sequence?

Mailaender:
They need to have none. My previous traits did not hard-code the sequences and had null checks to omit the animations in that case.

pchote:
The carryall drop will be removed if we ever proceed to the next stage of the d2k mod.

matija-hustic:
OK, so basically the lint checks, that have been added since this PR was open, have forced me to do some new stuff. Seeing how `WithParachute` doesn't support rendering the shadow as just the actor's usual sprite with the shadow palette, any parachutable actors that have relied on their shadow being rendered this way will no longer have a shadow when automatically upgrading the mods. I've added the nicer shadow `parach-shadow` manually for crates and d2k paratroops (this doesn't happen automatically as it potentially requires copying the image file and adding a sequence entry).

And, parachuting without a parachute sequence is now also handled with null-checks as before.

Mailaender:
Needs a rebase.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1306,11 +1306,11 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 
 							if (ibfNode.Any())
 								rh.Value.Nodes.RemoveAll(n => ibfNode.Contains(n));
-								Console.WriteLine("The 'ImagesByFullness' property from the removed RenderHarvester trait has been");
```
Mailaender: You fixed the indention error the wrong way here. https://github.com/OpenRA/OpenRA/pull/8644/ did it properly already.

matija-hustic: This was automatic when I rebased last night. It'll go away when I rebase now.

Mailaender: Be sure to leave a message when you do. I saw some of your pull request being silently rebased so they weren't merged already, because no one noticed.

matija-hustic:
Rebased.

OpenRA.Mods.Common/Traits/TargetableUnit.cs
```diff
@@ -9,28 +9,33 @@
...
 
 using System.Collections.Generic;
+using System.Linq;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Actor can be targeted.")]
-	public class TargetableUnitInfo : ITraitInfo, ITargetableInfo
+	public class TargetableUnitInfo : UpgradableTraitInfo, ITraitInfo, ITargetableInfo
```
Mailaender: This conflicts with https://github.com/OpenRA/OpenRA/pull/8662 and somehow broke things. The defending allied units in the shellmap don't retaliate anymore.

pchote:
What's the situation with this?  It seems a bit unfair that newer prs that conflict with this are being preferentially merged.

matija-hustic:
@GraionDilach and @Mailaender didn't how the `TargetableUnit` was modified so I'm waiting for #8662 to get merged before I rebase and use it.

GraionDilach:
#8662 is a step towards removing all `TargetableUnit` subclasses. That's something I feel preferable over the implementation provided here.

matija-hustic:
Rebased and using #8662. Shellmap works normally.

OpenRA.Mods.Common/Traits/Parachutable.cs
```diff
@@ -30,25 +30,11 @@ class ParachutableInfo : ITraitInfo
...
 		public readonly int FallRate = 13;
 
-		[Desc("Alternative to ParachuteShadowPalette which disables it and allows to set a custom sprite sequence instead.")]
-		public readonly string ShadowSequence = null;
-
-		[Desc("Optional, otherwise defaults to the palette the actor is using.")]
-		public readonly string ShadowPalette = null;
+		[UpgradeGrantedReference]
+		[Desc("Upgrade to grant to this actor when parachuting. Normally used to render the parachute using the WithParachute trait.")]
+		public readonly string ParachuteUpgrade = "parachute";
```
GraionDilach: Could this be a list, please?

mods/ra/rules/defaults.yaml
```diff
@@ -648,10 +658,17 @@
...
 		CustomBounds: 20,20
+	WithParachute:
+		UpgradeTypes: parachute
+		UpgradeMinEnabledLevel: 1
+		Image: parach
+		Sequence: idle
+		OpeningSequence: open
+		ShadowImage: parach-shadow
+		ShadowSequence: idle
+	UpgradeManager:
```
GraionDilach: Something implies me this was caused by auto-upgrading - `^Soldier` inherits both two required traits from `^ExistsInWorld`.

GraionDilach:
I think I can safely give my :+1: when the above nits are resolved.

mods/ra/rules/defaults.yaml
matija-hustic: This is a `^Crate` YAML here. Expand the code above ^^

GraionDilach: Oh, sorry. I misread it as the D2K infantry for some reason.

obrakmann:
Sorry, needs a rebase.

matija-hustic:
Rebased.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/db824811bcabdc6f2cf5c011790022ea678abf0f%5E...db824811bcabdc6f2cf5c011790022ea678abf0f)

mods/d2k/rules/defaults.yaml
```diff
@@ -225,6 +227,11 @@
 			Rough: 80
 	Voiced:
 		VoiceSet: InfantryVoice
+	WithParachute:
+		UpgradeTypes: parachute
+		UpgradeMinEnabledLevel: 1
+		ShadowImage: parach-shadow
+		ShadowSequence: idle
```
penev92: This somehow broke D2k:
>Unit `parach-shadow` does not have a sequence named `idle`

GraionDilach: The shadow is in TS SHP format instead of TD SHP format. >_>

matija-hustic: Godamn it. What do you guys use for SHP editing? I've downloaded this one program and I can't get anything done with it properly

matija-hustic: Altho, I don't remember messing around with this particular SHP. It was just a copy paste if I remember correctly

GraionDilach: I tend to go XCC Mixer (being part of XCC Utilities) from http://xhp.xwis.net/utilities/ due to that being practically the swiss knife WW modding tool. It has some bugs - only read pals within mixes, TS and RA2 folders are read from registry - but people can easily create a mix and feed it's folder as TD or RA1 install folder instead. It works under Wine.

PR #8351, opened 2015-06-01
Adjust the dogs generic name

abcdefg30:
Before it looked like this:
genericdog
Now:
genericdog2


obrakmann:
Well, that's hard to argue with. :+1:

PR #8352, opened 2015-06-01
Disable EnemyWatcher in all ra missions

abcdefg30:
IMHO it does not make sense in some of then.
(E.g. in allies05a when your spy drives with the truck into the enemy base, you get notifications...)

### DIFF
--- a/mods/ra/maps/allies-01/map.yaml
+++ b/mods/ra/maps/allies-01/map.yaml
@@ -470,6 +470,7 @@ Rules:
 		-ConquestVictoryConditions:
 		MissionObjectives:
 			EarlyGameOver: true
+		-EnemyWatcher:
 	World:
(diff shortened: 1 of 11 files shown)


abcdefg30:
Removed `EnemyWatcher` from all ra missions as "discussed" on IRC.

Mailaender:
It would make sense to keep it in the survival missions, because you are mostly defending there. However I also noticed some overlapping with other announcements there so :+1: anyway.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/5bc609b346259ddb124254f457616d0d619d529e%5E...5bc609b346259ddb124254f457616d0d619d529e)

PR #8356, opened 2015-06-01
make TD mission objective punctuation consistent

clemty:
some mission objectives end with ".", most don't - this removes the punctuation for consistency

some nod objectives end with an exclamation mark, I kept that even though it is inconsistent through nod missions as well

edit: I addded fullstops to all objectives and replaced exclamation marks in the process


Unit158:
How about we end them all with periods instead of leaving some with exclamation points?

clemty:
I assumed that the default is "no punctuation" since that was the case in the majority of objectives, but a quick online search shows that punctuation like in sentences is apparently good style so you're right and there should be punctuation:

http://www.onlinegrammar.com.au/punctuation-in-lists/
http://www.getitwriteonline.com/archive/101406VerticalLists.htm

abcdefg30:
Somehow I'd prefer fullstops, but above sounds reasonable, so :+1: / :white_check_mark:

obrakmann:
I'm in the "proper punctuation at all times, please"-camp.

pchote:
Everyone seems to agree that punctuation is required.

obrakmann:
You missed one in gdi03, but that's the objective for the AI player which is never shown anywhere, so :+1:

obrakmann:
Thanks!

[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/1926b4322d6cbea5f8f67139cf00cd88a5396620%5E...1926b4322d6cbea5f8f67139cf00cd88a5396620)

Issue #4451, opened 2014-01-11
Color picker area is not filled with colors.


xaionaro:
When I start openra instance and trying to pick the color, I see next window:

![Colorpicker_bug0](http://ut.mephi.ru/imgs/other/openra-colorpicker-bug0.png "color picker bug")

Bug is constantly repeated on my system. How can I help with debugging?

Related: https://github.com/OpenRA/OpenRA/pull/3169

penev92:
This was fixed in #8618.

PR #8380, opened 2015-06-05
Less permissive color validator

### LINKED ISSUE #7902: Player color validation: still not enough contrast to terrain [$22]
Despite the color validation colors that are too close to terrain can still be chosen (and are actually even auto-suggested after color validation "strike").

hard_2_see

Fixing it would also boost significance of the stealth units.

Random color suggestions need work too (example: first suggest colors of all available hues before repeating a hue ...)

<bountysource-plugin>
---
There is a **[$22 open bounty](https://www.bountysource.com/issues/10702919-player-color-validation-still-not-enough-contrast-to-terrain?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Enforces larger difference to terrain color and higher minimum lumination.

Closes #7902.

### DIFF
--- a/OpenRA.Mods.Common/ServerTraits/ColorValidator.cs
+++ b/OpenRA.Mods.Common/ServerTraits/ColorValidator.cs
@@ -22,8 +22,8 @@ namespace OpenRA.Mods.Common.Server
 	public class ColorValidator : ServerTrait, IClientJoined
 	{
 		// The bigger the color threshold, the less permitive is the algorithm
-		const int ColorThreshold = 0x40;
-		const byte ColorLowerBound = 0x33;
+		const int ColorThreshold = 0x70;
+		const byte ColorLowerBound = 0x80;
 		const byte ColorHigherBound = 0xFF;
 
 		static bool ValidateColorAgainstForbidden(Color askedColor, IEnumerable<Color> forbiddenColors, out Color forbiddenColor)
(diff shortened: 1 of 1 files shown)


Mailaender:
:+1: The new values seem to yield better results.

obrakmann:
That's better, but it still allows setting this:

screenshot from 2015-06-05 22 26 40

I played around a bit and got satisfying results with `0xA0` as threshold (and `0x80` for lower bound). That would allow setting colors like below, but that's about the closest you get:

screenshot from 2015-06-05 22 30 52

reaperrr:
Updated, changed the values to ```0x70``` for threshold and ```0x80``` for lower bound. As mentioned on irc, the higher the threshold, the more difficult it becomes to choose a GDI-ish color on TD desert maps, so I tried to find a middle-ground. This should still be significantly better than the values I used initially, maybe enough that players trying to exploit colors might get annoyed about the rejected colors and give up eventually (or settle for less advantageous colors).

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/8a9a960cfedc63c6912595b9c91f4b56ab0b341e%5E...8a9a960cfedc63c6912595b9c91f4b56ab0b341e)

x-a-n-a-x:
Is it normal that the "color chooser" hides the swatches after picking a color?

screenshot001

Mailaender:
No, that is a glitch: https://github.com/OpenRA/OpenRA/issues/4451. I am not sure, but https://github.com/OpenRA/OpenRA/pull/8425 might fix it.

PR #8382, opened 2015-06-05
Check also for 2012 and 2013 Visual C runtimes libaries

LipkeGu:
freetype6.dll is accepting also 2012 and 2013 revisions of the Visual C runtimes.

PR #8383, opened 2015-06-05
Download and install the required dependencies for freetype6

LipkeGu:
Add Visual C 2010 SP1 ...

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -121,11 +121,11 @@ elseif ($command -eq "dependencies")
 	cd ..
 	echo "Dependencies copied."
 	
-	$dep = "Microsoft Visual C++ 2010"
+	$dep = "Microsoft Visual C++ 201"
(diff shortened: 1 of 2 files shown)


LipkeGu:
Browser session was out of Sync X.x

penev92:
As a user I really hate when programs decide they can download and install random stuff whenever they feel like it.
I am OK with the current fetch script, because that only downloads a few DLLs, and doesn't affect your system in any way.

The fact that you download and install this no matter what in your current implementation (the check has *nothing* to do with the download as you've wrote it, it tries to download an .exe from a link that nobody can guarantee won't change. This isn't a really good thing by itself.
Also there is a ton of stuff that can go wrong with the installation - the user might not have permissions, the installation process can fail due to whatever reason, the script can throw when trying to delete the file that you just started...

:-1: for automatic installation (of anything)

LipkeGu:
@penev92 #8382 <- we could use this PR

penev92:
Yes, that one I agree with.

PR #8387, opened 2015-06-05
Create only one scratch vertex buffer for rendering

RoosterDragon:
Previously several buffers were created but only one was ever used.


Mailaender:
Can't spot any regressions. Tested unit rendering, the grid overlay and A* debug layer. :white_check_mark:

pchote:
Looks good to me :+1:

chrisforbes:
ARGH, NAK NAK NAK.

This looks unused, but it's a performance workaround for lousy drivers that fail to `rename` our BO memory and instead decide to stall.

chrisforbes:
[It looks like someone broke the actual behavior of this some time ago. I don't have time to do the archaeology to untangle it now]

Mailaender:
Next time, better leave a comment in the source code explaining those.

PR #8389, opened 2015-06-06
Fixed units stopping when making them move while they are already moving.

irungentoo:
This fixes an annoying issue where units stop for a moment if you give them an order to move to the same place they are moving to.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Move/Move.cs
+++ b/OpenRA.Mods.Common/Activities/Move/Move.cs
@@ -150,12 +150,6 @@ namespace OpenRA.Mods.Common.Activities
 
 			if (path == null)
 			{
-				if (mobile.TicksBeforePathing > 0)
-				{
-					--mobile.TicksBeforePathing;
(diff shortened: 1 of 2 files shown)


Unit158:
How does this effect performance? Can you get some numbers please?

irungentoo:
@Unit158 The function after the check gets called once every time you order a unit to move, same thing as before the PR.

I'm not sure what this code was supposed to fix but removing it only seems to do good things.

Mailaender:
This was added by @chrisforbes way back in the past https://github.com/irungentoo/OpenRA/commit/e9d95dba72e4aa862fe1dd523ee7955f65ef087e to avoid all units pathing at once hence the random delay. It avoids lag when a large group of units start moving.

Unit158:
:-1: This emphasizes our already slow pathfinder.

Mailaender:
The proper fix would be something like https://github.com/OpenRA/OpenRA/issues/3241.

PR #8392, opened 2015-06-06
Fixed a resource leak in BigFile

Mailaender:
This isn't used anywhere except @ScottNZ's abandoned CCG demake mod so not too severe. Still it is code that could serve as a template to others so it should dispose streams properly. Detected by Coverity.

### DIFF
--- a/OpenRA.Game/FileSystem/BigFile.cs
+++ b/OpenRA.Game/FileSystem/BigFile.cs
@@ -26,26 +26,27 @@ namespace OpenRA.FileSystem
 			Name = filename;
 			Priority = priority;
 
-			var s = GlobalFileSystem.Open(filename);
-
-			if (s.ReadASCII(4) != "BIGF")
-				throw new InvalidDataException("Header is not BIGF");
-
-			// Total archive size.
-			s.ReadUInt32();
-
-			var entryCount = s.ReadUInt32();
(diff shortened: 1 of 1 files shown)


Mailaender:
Suggested way to review: https://github.com/OpenRA/OpenRA/pull/8392/files?w=1

ScottNZ:
I don't think this is going to work. It's going to dispose the stream at the end of the constructor, except we still need to use the stream later. `s` should be pulled out into a field and disposed in a destructor/IDisposable implementation.

Mailaender:
`Entry.GetData` gets the stream from the `Entry` constructor.

pchote:
The stream from the `Entry` constructor is the same stream that you're disposing, so it is bogus by the time `Entry.GetData` uses it.

RoosterDragon:
Try storing the stream in a field and implementing ```IDisposable``` on ```BigFile```, you can throw away the stream there. Check out MixFile.cs which does a similar thing.

For bonus points on top of that change you should make sure the constructor is exception-safe. You'd need to do roughly the following:
```
s = GlobalFileSystem.Open(filename);
try
{
    // Other ctor stuff that might throw.
}
catch
{
    Dispose();
    throw;
}
```

Mailaender:
Probably not worth the effort with zero progress on the Generals demake.

PR #8393, opened 2015-06-06
Fixed a dereference null return value in ZipFile.GetContent

Mailaender:
I believe this could blow up easily which should be avoided. Detected by Coverity.

### DIFF
--- a/OpenRA.Game/FileSystem/ZipFile.cs
+++ b/OpenRA.Game/FileSystem/ZipFile.cs
@@ -58,7 +58,11 @@ namespace OpenRA.FileSystem
 
 		public Stream GetContent(string filename)
 		{
-			using (var z = pkg.GetInputStream(pkg.GetEntry(filename)))
+			var entry = pkg.GetEntry(filename);
+			if (entry == null)
+				return null;
+
+			using (var z = pkg.GetInputStream(entry))
 			{
 				var ms = new MemoryStream();
 				z.CopyTo(ms);


OpenRA.Game/FileSystem/ZipFile.cs
```diff
@@ -58,7 +58,11 @@ public ZipFile(string filename, int priority, Dictionary<string, byte[]> content
 
 		public Stream GetContent(string filename)
 		{
-			using (var z = pkg.GetInputStream(pkg.GetEntry(filename)))
+			var entry = pkg.GetEntry(filename);
+			if (entry == null)
+				return Stream.Null;
```
RoosterDragon: Other ```GetContent``` implementations appear to return ```null``` rather than ```Stream.Null``` for non-existent file-names.

Mailaender: You are right. We never use and expect Stream.Null so I better changed it.

RoosterDragon:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/0d00bc4f7c3aec7bf9a766d5820b6a5c04046e7d%5E...0d00bc4f7c3aec7bf9a766d5820b6a5c04046e7d)

PR #8394, opened 2015-06-06
Fixed a null reference exception in Map.Save

Mailaender:
Discovered via Coverity.

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -484,7 +484,7 @@ namespace OpenRA
 			}
 
 			// Saving the map to a new location
-			if (toPath != Path)
+			if (toPath != Path || Container == null)
 			{
 				Path = toPath;
 


pchote:
"If we don't have an existing container in `Save()` then don't actually save the file" doesn't seem like a proper fix to me.

OpenRA.Game/Map/Map.cs
```diff
@@ -493,7 +493,8 @@ public void Save(string toPath)
 			}
 
 			// Update existing package
-			Container.Write(entries);
+			if (Container != null)
```
pchote: A better fix would probably to add an `|| Container == null` case to the conditional above.

Mailaender:
Updated.

OpenRA.Game/Map/Map.cs
```diff
@@ -483,10 +483,11 @@ public void Save(string toPath)
 				}
 			}
 
-			// Saving the map to a new location
-			if (toPath != Path)
+			if (toPath != Path || Container == null)
 			{
-				Path = toPath;
+				// Saving the map to a new location
+				if (Path != toPath)
```
pchote: This check is unnecessary.  If `Path == toPath` then reassigning the variable doesn't change anything.

Mailaender: Removed the superfluous check.

pchote:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/57db26c9f25667b8afcef570f657e30cd5490e24%5E...57db26c9f25667b8afcef570f657e30cd5490e24)

PR #8395, opened 2015-06-06
Fixed null reference exception when processing the Pause order

Mailaender:
Discovered by Coverity.

### DIFF
--- a/OpenRA.Game/Network/UnitOrders.cs
+++ b/OpenRA.Game/Network/UnitOrders.cs
@@ -117,7 +117,7 @@ namespace OpenRA.Network
 						if (client != null)
 						{
 							var pause = order.TargetString == "Pause";
-							if (orderManager.World.Paused != pause && !world.LobbyInfo.IsSinglePlayer)
+							if (orderManager.World.Paused != pause && world != null && !world.LobbyInfo.IsSinglePlayer)
 							{
 								var pausetext = "The game is {0} by {1}".F(pause ? "paused" : "un-paused", client.Name);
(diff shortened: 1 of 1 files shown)


OpenRA.Game/Network/UnitOrders.cs
```diff
@@ -117,7 +117,7 @@ public static void ProcessOrder(OrderManager orderManager, World world, int clie
 						if (client != null)
 						{
 							var pause = order.TargetString == "Pause";
-							if (orderManager.World.Paused != pause && !world.LobbyInfo.IsSinglePlayer)
+							if (orderManager.World.Paused != pause && world != null && !world.LobbyInfo.IsSinglePlayer)
```
pchote: This looks bogus - you're mixing `orderManager.World` and `world`.

Mailaender: I also thought about that, but we are using `world` in the methods above, too.

Mailaender: > you're mixing `orderManager.World` and `world`.

Not me, this was actually introduced by yourself in https://github.com/OpenRA/OpenRA/pull/2991 so I assume you knew what you were doing back then. If you spotted a mistake of your own, you need to elaborate.

abcdefg30:
:+1: / :white_check_mark:

penev92:
Well this silences ReSharper and Coverity, so :+1: 
The `world` that is passed to the method as argument is used in a bunch of other places, while `orderManager.World` is used only here, so `¯\_(ツ)_/¯`

Issue #2441, opened 2012-10-20
MixFile.RaHeader, duplicate values found


dn150:
On map room-warzone with release-20121019, playing as Allies, closed as game began. exception.log contains:

```
Exception of type `System.ArgumentException`: MixFile.RaHeader, duplicate values found for the following keys: 2811818068: [(offs=1376304610, len=1436818881),(offs=1303428578, len=3844756581),(offs=3932274146, len=504285086),(offs=261406178, len=2011127635),(offs=3772169698, len=2458156051),(offs=4124556770, len=1657300316),(offs=394575330, len=2763129078),(offs=3464150498, len=1912530027),(offs=3887775202, len=1535573085),(offs=1667349986, len=1651050265)], 3255518541: [(offs=1099208783, len=3303102605),(offs=3604191311, len=3972225307),(offs=2693699663, len=316463550),(offs=2821888079, len=2932147912)], 2925018180: [(offs=118932477, len=331163447),(offs=2042545149, len=4248835109)], 3575071090: [(offs=2624743082, len=3797053306),(offs=1751213738, len=1182934561)], 1243613079: [(offs=1637127797, len=3841171263),(offs=3865482869, len=2023497237)], 3284888157: [(offs=687789427, len=358779927),(offs=3138704755, len=4230776020),(offs=1887884659, len=3736988869)], 465091051: [(offs=1882287750, len=3730501625),(offs=3090378374, len=4279611615)], 3851858380: [(offs=197200840, len=1913283067),(offs=83561416, len=2673869643)], 3683159305: [(offs=867278684, len=1087770530),(offs=2663358300, len=3130547281),(offs=1722326876, len=1312637092)], 400405198: [(offs=2385359528, len=400405198),(offs=1218818728, len=1142595513)]
   at OpenRA.Exts.ToDictionaryWithConflictLog[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, String debugName, Func`2 logKey, Func`2 logValue)
   at OpenRA.Exts.ToDictionaryWithConflictLog[TSource,TKey](IEnumerable`1 source, Func`2 keySelector, String debugName, Func`2 logKey, Func`2 logValue)
   at OpenRA.FileFormats.MixFile..ctor(String filename, Int32 priority)
   at OpenRA.FileFormats.FileSystem.OpenPackage(String filename, Int32 order)
   at OpenRA.FileFormats.FileSystem.OpenPackage(String filename)
[...]

PR #8399, opened 2015-06-06
Fixed EndOfStreamException in MixFile.ReadBlocks

### LINKED ISSUE #5866: Attempted to read past the end of the stream
I closed a game on my localhost, by clicking X from window and got this:

exception.log:
```
Red Alert Mod at Version release-20140608
Runtime Version: .NET CLR 4.0.30319.18444
Exception of type `System.IO.EndOfStreamException`: Attempted to read past the end of the stream.
   at OpenRA.StreamExts.ReadBytes(Stream s, Int32 count)
   at OpenRA.FileSystem.MixFile.ReadBlocks(Stream s, Int64 offset, Int32 count)
   at OpenRA.FileSystem.MixFile.DecryptHeader(Stream s, Int64 offset, Int64& headerEnd)
   at OpenRA.FileSystem.MixFile..ctor(String filename, PackageHashType type, Int32 priority)
   at OpenRA.FileSystem.GlobalFileSystem.OpenPackage(String filename, String annotation, Int32 order)
   at OpenRA.FileSystem.GlobalFileSystem.<Mount>c__AnonStorey29.<>m__7()
   at OpenRA.FileSystem.GlobalFileSystem.Mount(String name, String annotation)
   at OpenRA.FileSystem.GlobalFileSystem.LoadFromManifest(Manifest manifest)
   at OpenRA.ModData.PrepareMap(String uid)
   at OpenRA.Game.StartGame(String mapUID, Boolean isShellmap)
   at OpenRA.Game.LoadShellMap()
   at OpenRA.Mods.RA.Widgets.Logic.IngameMenuLogic.<IngameMenuLogic>c__AnonStorey111.<>m__30A()
   at System.Action.Invoke()
   at OpenRA.Primitives.ActionQueue.PerformActions()
   at OpenRA.Game.Tick(OrderManager orderManager)
   at OpenRA.Game.Run()
```

server.log:
```
Order lag is now 3 frames.
Order lag is now 1 frames.
Order lag is now 3 frames.
Order lag is now 1 frames.
Order lag is now 3 frames.
Order lag is now 1 frames.
Order lag is now 3 frames.
Order lag is now 1 frames.
Initial mod: ra
Initial map: 18517ca00f89a4266044acdb2e4ec5569b335cec
Client 0: Accepted connection from 127.0.0.1:60832.
Order lag is now 3 frames.
Order lag is now 1 frames.
Player @127.0.0.1:60832 is NotReady
Dropping client 0 because reading the data failed: System.Net.Sockets.SocketException (0x80004005): An existing connection was forcibly closed by the remote host
   at System.Net.Sockets.Socket.Receive(Byte[] buffer, Int32 offset, Int32 size, SocketFlags socketFlags)
   at OpenRA.Server.Connection.ReadDataInner(Server server)

```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3004854-attempted-to-read-past-the-end-of-the-stream?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mailaender:
Closes #5866 by replacing the EndOfStreamException with more useful ones for debugging like we do in StreamExts. I am not sure what is happening. It seems to be a Windows .NET only issue. Probably related to https://github.com/OpenRA/OpenRA/issues/2441.

### DIFF
--- a/OpenRA.Game/FileSystem/MixFile.cs
+++ b/OpenRA.Game/FileSystem/MixFile.cs
@@ -122,6 +122,15 @@ namespace OpenRA.FileSystem
 
 		static uint[] ReadBlocks(Stream s, long offset, int count)
 		{
+			if (offset < 0)
+				throw new ArgumentOutOfRangeException("offset", "Non-negative number required.");
+
+			if (count < 0)
+				throw new ArgumentOutOfRangeException("count", "Non-negative number required.");
+
+			if (offset + (count * 2) > s.Length)
+				throw new ArgumentException("Bytes to read {0} and offset {1} greater than stream length {2}.".F(count * 2, offset, s.Length));
+
 			s.Seek(offset, SeekOrigin.Begin);
 
 			// A block is a single encryption unit (represented as two 32-bit integers)


OpenRA.Game/FileSystem/MixFile.cs
```diff
@@ -122,8 +122,14 @@ static MemoryStream Decrypt(uint[] h, Blowfish fish)
 
 		static uint[] ReadBlocks(Stream s, long offset, int count)
 		{
+			if (offset > s.Length)
+				throw new ArgumentOutOfRangeException("offset", "Offset {0} greater than stream length {1}.".F(offset, s.Length));
+
 			s.Seek(offset, SeekOrigin.Begin);
 
+			if (count < 0)
```
RoosterDragon: You should move this check above the seek.

Mailaender: Even though Seek does not use it?

RoosterDragon: So you don't change the stream position until you've validated all the arguments.

Mailaender: Okay, though I don't want anyone to catch these exceptions.

OpenRA.Game/FileSystem/MixFile.cs
```diff
@@ -122,6 +122,12 @@ static MemoryStream Decrypt(uint[] h, Blowfish fish)
 
 		static uint[] ReadBlocks(Stream s, long offset, int count)
 		{
+			if (offset > s.Length)
+				throw new ArgumentOutOfRangeException("offset", "Offset {0} greater than stream length {1}.".F(offset, s.Length));
```
RoosterDragon: Looking at this more closely, we may wish to throw an ```ArgumentException``` if offset + bytes to read is greater than the stream length.

We would also want to ensure ```offset``` is non-negative, just like count.

Mailaender:
I added some more checks as requested.

OpenRA.Game/FileSystem/MixFile.cs
```diff
@@ -122,6 +122,18 @@ static MemoryStream Decrypt(uint[] h, Blowfish fish)
 
 		static uint[] ReadBlocks(Stream s, long offset, int count)
 		{
+			if (offset < 0)
+				throw new ArgumentOutOfRangeException("offset", "Non-negative number required.");
+
+			if (offset > s.Length)
+				throw new ArgumentOutOfRangeException("offset", "Offset {0} greater than stream length {1}.".F(offset, s.Length));
```
RoosterDragon: Sorry I meant to imply we can do without this check, since the offset + bytes count you've now added checks the same thing (and more) anyway.

Mailaender: Removed it.

OpenRA.Game/FileSystem/MixFile.cs
```diff
@@ -122,6 +122,15 @@ static MemoryStream Decrypt(uint[] h, Blowfish fish)
 
 		static uint[] ReadBlocks(Stream s, long offset, int count)
 		{
+			if (offset < 0)
+				throw new ArgumentOutOfRangeException("offset", "Non-negative number required.");
+
+			if (count < 0)
```
Phrohdoh: `<=`?

Mailaender: That won't be a problem. http://stackoverflow.com/questions/3102353/what-happens-if-i-initialize-an-array-to-size-0

RoosterDragon:
:+1:

PR #8405, opened 2015-06-06
POSIX compliance

moviuro:
I want to run a dedicated on FreeBSD, hence the 3 issues and this PR

### DIFF
--- a/thirdparty/configure-linux-native-deps.sh
+++ b/thirdparty/configure-linux-native-deps.sh
@@ -2,7 +2,7 @@
 # Use Linux system dependencies where possible, but take into account different .so names.
 
 os="$(uname -s)"
-if [ "$os" == 'Linux' ] || [ "$os" == 'FreeBSD' ]; then
+if [ "$os" = 'Linux' ] || [ "$os" = 'FreeBSD' ]; then
 	locations="/lib /lib64 /usr/lib /usr/lib64 /usr/lib/i386-linux-gnu /usr/lib/x86_64-linux-gnu /usr/local/lib /opt/lib"
 	sonames="liblua.so.5.1.5 liblua5.1.so.5.1 liblua5.1.so.0 liblua.so.5.1 liblua-5.1.so liblua5.1.so"
 
--- a/thirdparty/configure-linux-native-deps.sh
+++ b/thirdparty/configure-linux-native-deps.sh
@@ -1,18 +1,18 @@
-#!/bin/bash
+#!/bin/sh
 # Use Linux system dependencies where possible, but take into account different .so names.
 
-os=`uname`
-if [ "$os" == 'Linux' ]; then
-	locations=(/lib /lib64 /usr/lib /usr/lib64 /usr/lib/i386-linux-gnu /usr/lib/x86_64-linux-gnu /usr/local/lib /opt/lib)
-	sonames=(liblua.so.5.1.5 liblua5.1.so.5.1 liblua5.1.so.0 liblua.so.5.1 liblua-5.1.so liblua5.1.so)
+os="$(uname -s)"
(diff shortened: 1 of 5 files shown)


Mailaender:
:+1: Still works fine for me under Linux. :white_check_mark:

moviuro:
@Mailaender That is the point of POSIX-compliance :stuck_out_tongue_closed_eyes: work on Linux, BSDs, OSX and whatnot...

obrakmann:
the travis log shows this:
```
./thirdparty/configure-linux-native-deps.sh: 5: [: Linux: unexpected operator
```

thirdparty/configure-linux-native-deps.sh
```diff
@@ -1,18 +1,18 @@
-#!/bin/bash
+#!/bin/sh
 # Use Linux system dependencies where possible, but take into account different .so names.
 
-os=`uname`
-if [ "$os" == 'Linux' ]; then
-	locations=(/lib /lib64 /usr/lib /usr/lib64 /usr/lib/i386-linux-gnu /usr/lib/x86_64-linux-gnu /usr/local/lib /opt/lib)
-	sonames=(liblua.so.5.1.5 liblua5.1.so.5.1 liblua5.1.so.0 liblua.so.5.1 liblua-5.1.so liblua5.1.so)
+os="$(uname -s)"
+if [ "$os" == 'Linux' ] || [ "$os" == 'FreeBSD' ]; then
```
Mailaender: This line seems to fail under Ubuntu.

moviuro: I can't begin to understand why :open_mouth:

Mailaender: Just remove it. We don't really need it anymore. The real check moved to the https://github.com/OpenRA/OpenRA/blob/bleed/Makefile#L78

thirdparty/configure-linux-native-deps.sh
```diff
@@ -1,18 +1,18 @@
-#!/bin/bash
+#!/bin/sh
 # Use Linux system dependencies where possible, but take into account different .so names.
 
-os=`uname`
-if [ "$os" == 'Linux' ]; then
-	locations=(/lib /lib64 /usr/lib /usr/lib64 /usr/lib/i386-linux-gnu /usr/lib/x86_64-linux-gnu /usr/local/lib /opt/lib)
-	sonames=(liblua.so.5.1.5 liblua5.1.so.5.1 liblua5.1.so.0 liblua.so.5.1 liblua-5.1.so liblua5.1.so)
+os="$(uname -s)"
+if [ "$os" = 'Linux' ] || [ "$os" = 'FreeBSD' ]; then
```
Mailaender: Turns out your latest change completely breaks it for Ubuntu. Notice the missing line "Detected Lua 5.1 library" at https://travis-ci.org/OpenRA/OpenRA/builds/65695051

moviuro: I don't have an Ubuntu at hand. I'd appreciate it if someone had one and could actually post only the result of ```sh ./thirdparty/configure-linux-native-deps.sh``` after adding ```set -x``` right under the shebang.
It works on Archlinux and FreeBSD.

obrakmann: @Mailaender: the missing "Detected Lua" line is ok, lua isn't installed in the travis vm. (Which is ok, because the generated Eluant.dll.config isn't used for the debian package.)

obrakmann:
:+1: works fine here too

obrakmann:
Thanks!

[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/e7447e1318e9c7ae0f10d106744107801b053212%5E...e7447e1318e9c7ae0f10d106744107801b053212)

Issue #6743, opened 2014-10-12
Crash in new playtest: Object reference not set to an instance of an object at Graphics.SpriteFont.Measure


cerebrum:
https://gist.github.com/cerebrum/2e833be15a7129168819

PR #8409, opened 2015-06-06
Fixed NullReferenceExceptions in SpawnSelectorTooltipLogic

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/6743.


RoosterDragon:
Won't this leave us with the old widths when the string is null or empty? Could we do something like ```font.Measure(text ?? "")``` to avoid the nulls but still get correct widths?

Mailaender:
In that case I guess https://github.com/OpenRA/OpenRA/pull/8457 makes more sense.

Issue #4926, opened 2014-03-20
D2k Devastator Self Destruct


Sircrashalot:
Could you guys add an ability to self destruct or is that already possible?
##

Mailaender:
At first I tried to refactor MADTank into re-usable traits for like `PreventsTeleportOnCondition`, `EjectDriverOnCondition` and  `ShakeOnCondition`, but `ShakeOnCondition` would require another `ITick` for a periodical shake. The split of `Suicides` and `AttackSuicides` was tried first as it made sense to separate two different order types https://github.com/Mailaender/OpenRA/tree/devastator-suicide but it didn't play well with `GrantConditionOnDeploy` to glue together the rest which didn't activate when using the right-click deploy. Only the hotkey hack and command bar triggers all of them at once. So I abandoned this approach.

I went with a simpler version https://github.com/Mailaender/OpenRA/tree/devastator-overload in #13884 where only one state controls everything. Once the reactor overload is triggered the devastator will take damage over time, display an overlay and the explosion type changes to something more "devastating". The original might have had a one-click boom (don't know) but I also assume nuclear reactor meltdowns don't work that way and this is better controllable as you maneuver your tank into an enemy blob.

PR #8410, opened 2015-06-06
Refactored detonate behaviour and added Devastator logic.

Unit158:
~~This isn't quite ready to merge yet, as I am not sure how to do the overlay properly.~~

Closes https://github.com/OpenRA/OpenRA/issues/4926.


OpenRA.Mods.Common/Traits/Detonates.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Text;
+using OpenRA.Activities;
+using OpenRA.Mods.Common.Orders;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Traits;
+
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public class DetonatesInfo : ITraitInfo, Requires<RenderUnitInfo>
```
penev92: RenderUnit has a deprecation notice. Please use the trait that's replacing it.

OpenRA.Mods.Common/Traits/Detonates.cs
```diff
@@ -0,0 +1,128 @@
...
+
+		protected void StartDetonationSequence()
+		{
+			if (Deployed)
+				return;
+
+			Deployed = true;
+
+			if (info.ChargeSequence != null)
+				renderUnit.PlayCustomAnimationRepeating(self, info.ChargeSequence);
```
penev92: Something tells me WithChargeAnimation makes this redundant.

OpenRA.Mods.Common/Traits/Detonates.cs
```diff
@@ -0,0 +1,128 @@
...
+
+			if (target.Type == TargetType.FrozenActor)
+				return new Order(order.OrderID, self, queued) { ExtraData = target.FrozenActor.ID };
+
+			return new Order(order.OrderID, self, queued) { TargetActor = target.Actor };
+		}
+
+		public string VoicePhraseForOrder(Actor self, Order order)
+		{
+			return "Attack";
```
penev92: Please move this to the info class so it can be configurable via YAML.

OpenRA.Mods.Common/Traits/Detonates.cs
Unit158: One provides an overlay, the other changes the unit's sequence.

OpenRA.Mods.Common/Traits/Detonates.cs
```diff
@@ -0,0 +1,128 @@
...
+			{
+				if (info.DetonationWeapon != null)
+				{
+					var weapon = self.World.Map.Rules.Weapons[info.DetonationWeapon.ToLowerInvariant()];
+
+					// Use .FromPos since this actor is killed. Cannot use Target.FromActor
+					weapon.Impact(Target.FromPos(self.CenterPosition), self, Enumerable.Empty<int>());
+				}
+
+				self.Kill(self);
```
penev92: This seems redundant for the devastator because of the explosion's third warhead.
We should remove it so this trait can be used by stuff like emp tanks, yuri, chaos drones... and anything that doesn't suicide.

OpenRA.Mods.Common/Traits/Detonates.cs
penev92: Sorry, WithChargeOverlay. I remember @Mailaender had a pr for that for the TS obelisk.

OpenRA.Mods.Common/Traits/Detonates.cs
Unit158: I'll add a killself flag

penev92: The weapon should already be handling that.

OpenRA.Mods.Common/Traits/Detonates.cs
```diff
@@ -0,0 +1,130 @@
...
+
+			self.QueueActivity(new Wait(info.ChargeDelay, false));
+			self.QueueActivity(new CallFunc(() => Sound.Play(info.ChargeSound, self.CenterPosition)));
+			self.QueueActivity(new Wait(info.DetonationDelay, false));
+			self.QueueActivity(new CallFunc(Detonate));
+		}
+
+		public void ResolveOrder(Actor self, Order order)
+		{
+			if (order.OrderString == "DetonateAttack")
```
penev92: You should probably check Deployed (rename to IsDeployed) here and return if true.

penev92: *Before* here. No resolving orders once we've started!

OpenRA.Mods.Common/Traits/Render/WithChargeAnimation.cs
```diff
@@ -0,0 +1,74 @@
...
+
+		public object Create(ActorInitializer init) { return new WithChargeAnimation(init, this); }
+	}
+
+	public class WithChargeAnimation : INotifyCharging, INotifyDamageStateChanged, INotifySold
+	{
+		readonly RenderUnit renderUnit;
+		readonly WithChargeAnimationInfo info;
+
+		bool charging;
```
Unit158: I'm not sure how I can implement this in a similar fashion to WithChargeOverlay

Unit158:
Help wanted.

Phrohdoh:
What do you need help with?

OpenRA.Mods.Common/Traits/Render/WithChargeAnimation.cs
```diff
@@ -0,0 +1,74 @@
...
+		public void FinishedCharging(Actor self)
+		{
+		}
+
+		public void DamageStateChanged(Actor self, AttackInfo e)
+		{
+			if (info.Repeating)
+				wsb.PlayCustomAnimationRepeating(self, wsb.NormalizeSequence(self, info.Sequence));
+			else
+				wsb.PlayCustomAnimationRepeating(self, wsb.NormalizeSequence(self, info.Sequence));
```
Phrohdoh: These are the exact same.

Unit158:
I don't know why travis is failing :(

Phrohdoh:
Travis tells you:
```
Testing Tiberian Sun mod MiniYAML...

Testing map: A River Runs Near It

OpenRA.Utility(1,1): Error: Missing Type: DemoTruckInfo

OpenRA.Utility(1,1): Error: Missing Type: DemoTruckInfo
```

Unit158:
Does that map have custom yaml?

Phrohdoh:
Did you fix the ts rules at all in this PR?

Phrohdoh:
Did you fix the ts rules at all in this PR?

penev92:
You still haven't said what it is you need help with?
Also you are removing a trait and replacing it with another. This will need an upgrade rule. And if someone isn't convinced - there are maps that need upgrading that were missed.

Unit158:
I actually mentioned a while back what it was I needed help with. It's gone now though.

mods/ra/rules/vehicles.yaml
```diff
@@ -691,7 +691,10 @@ DTRK:
 	Explodes:
 		Weapon: MiniNuke
 		EmptyWeapon: MiniNuke
-	DemoTruck:
+	Detonates:
+		ChargeDelay: 0
```
GraionDilach: Why do you specify the ChargeDelay and DetonationDelay here, when the default is also 0?

Unit158: It wasn't originally 0, I just forgot to change it back.

OpenRA.Mods.Common/Traits/Detonates.cs
```diff
@@ -0,0 +1,138 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using System.Text;
```
obrakmann: I don't think this is really used.

OpenRA.Mods.Common/Traits/Detonates.cs
```diff
@@ -0,0 +1,138 @@
...
+			if (order.OrderString == "DetonateAttack")
+			{
+				var target = self.ResolveFrozenActorOrder(order, Color.Red);
+				if (target.Type != TargetType.Actor)
+					return;
+
+				if (!order.Queued)
+					self.CancelActivity();
+
+				self.SetTargetLine(target, Color.Red);
```
obrakmann: I know we're doing this everywhere, but this should really be in a `CallFunc`. I guess I'll have to go all over those places if I ever get around to redoing `Activity`.

obrakmann:
Looks good to me overall. Needs an upgrade rule.

OpenRA.Mods.Common/Traits/Detonates.cs
Unit158: What should be done? SetTargetLine?

penev92:
@Unit158, I believe you were going to update this?

Unit158:
some day.

I have the code ready minus a flaw with the upgrade rule so ill push today if when i get the chance

OpenRA.Mods.Common/Traits/Detonates.cs
```diff
@@ -0,0 +1,147 @@
...
+using OpenRA.Mods.Common.Orders;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public class DetonatesInfo : ITraitInfo
+	{
+		[Desc("Measured in ticks.")]
+		public readonly int ChargeDelay = 0;
+		public readonly string ChargeSequence = null;
```
Mailaender: Needs a `[SequenceReference]`.

OpenRA.Mods.Common/Traits/Detonates.cs
```diff
@@ -0,0 +1,147 @@
...
+		public readonly string ChargeSound = null;
+
+		[Desc("Measured in ticks.")]
+		public readonly int DetonationDelay = 0;
+		[WeaponReference]
+		public readonly string DetonationWeapon = null;
+
+		public readonly string OrderVoice = "Attack";
```
Mailaender: Needs a `[VoiceReference]` and probably wants to be `"Action"` nowadays.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1204,9 +1204,45 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 						if (editorSprite != null)
 							node.Value.Nodes.Remove(editorSprite);
 					}
-				}
```
Mailaender: You resolved the rebase conflicts incorrectly. This brace needs to stay.

OpenRA.Mods.Common/Traits/Detonates.cs
```diff
@@ -0,0 +1,149 @@
...
+		public readonly string ChargeSequence = null;
+		public readonly string ChargeSound = null;
+
+		[Desc("Measured in ticks.")]
+		public readonly int DetonationDelay = 0;
+		[WeaponReference]
+		public readonly string DetonationWeapon = null;
+
+		[VoiceReference]
+		public readonly string OrderVoice = "Action";
```
abcdefg30: Why `OrderVoice` instead of just `Voice` like we do elsewhere?

abcdefg30:
When I "deploy" a `Devastator`, I get:
`BUG: in order targeter - decided to Detonate but then didn't order`

abcdefg30:
ra seems to work fine. :white_check_mark:

Unit158:
That bug is probably unrelated to Detonates, but someone who knows the order code would know better than I.

Mailaender:
> That bug is probably unrelated to Detonates, but someone who knows the order code would know better than I.

That means your code will desync in multiplayer, because you circumvented the network order system!

OpenRA.Mods.Common/Traits/Render/WithChargeOverlay.cs
```diff
@@ -25,6 +25,8 @@ public class WithChargeOverlayInfo : ITraitInfo, Requires<RenderSpritesInfo>
 		[Desc("Custom palette is a player palette BaseName")]
 		public readonly bool IsPlayerPalette = false;
 
+		public readonly bool Repeating = false;
```
penev92: A `Desc` would be nice.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1380,6 +1380,43 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 						})));
 				}
 
+				// Detonates replaces DemoTruck and changes MadTank
+				if (engineVersion < 20150629)
```
penev92: Looks like you have this upgrade rule twice?

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1607,6 +1644,43 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 					}
 				}
 
+				// Detonates replaces DemoTruck and changes MadTank
+				if (engineVersion < 20150629)
```
penev92: This should have its date updated.

Unit158:
That makes no sense, because as far as I can see, no code is ran outside the orders. Please read the code beforehand.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
Unit158: Nobody noticed that the bigger issue is that AttackCharge will break this class, so I'll have to find another solution or fix my current solution.

Unit158:
This will also need a way for the devastator to discern from attacking and moving to detonate. I'm just lucky it didn't bite me in the ass in testing.

Unit158:
Fixed the order targeter issue and added a hack to WithChargeOverlay to keep it above the unit.

Unit158:
This should be in it's final state.

OpenRA.Mods.Common/Traits/Render/WithChargeOverlay.cs
```diff
@@ -44,14 +47,27 @@ public WithChargeOverlay(ActorInitializer init, WithChargeOverlayInfo info)
 
 			overlay = new Animation(init.World, renderSprites.GetImage(init.Self));
 
-			renderSprites.Add(new AnimationWithOffset(overlay, null, () => !charging),
-				info.Palette, info.IsPlayerPalette);
+			// HACK: WithChargeOverlay added before sprite body.
+			init.World.AddFrameEndTask(w =>
+				renderSprites.Add(new AnimationWithOffset(overlay, null, () => !charging),
```
penev92: Use [this constructor](https://github.com/OpenRA/OpenRA/blob/12966129a4e412fd2c53eb05b310f30557058521/OpenRA.Game/Graphics/AnimationWithOffset.cs#L30) instead and set the `zOffset`. That should take care of the issue you are working around with this.

Mailaender:
Works as promised in both RA and D2k. :white_check_mark: Also a nice cleanup. :+1:

OpenRA.Mods.Common/Traits/Render/WithChargeOverlay.cs
```diff
@@ -44,14 +47,25 @@ public WithChargeOverlay(ActorInitializer init, WithChargeOverlayInfo info)
 
 			overlay = new Animation(init.World, renderSprites.GetImage(init.Self));
 
-			renderSprites.Add(new AnimationWithOffset(overlay, null, () => !charging),
+			renderSprites.Add(new AnimationWithOffset(overlay, null, () => !charging, 1),
```
pchote: Please define the z offset in the sequences instead of hardcoding it here.

OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -36,7 +36,7 @@ public interface IPlaceBuildingDecoration
 	}
 
 	public interface INotifyAttack { void Attacking(Actor self, Target target, Armament a, Barrel barrel); }
-	public interface INotifyCharging { void Charging(Actor self, Target target); }
+	public interface INotifyCharging { void Charging(Actor self, Target target); void FinishedCharging(Actor self); }
```
pchote: Please expand this out so there is one signature per line.

pchote:
Please split the devastator changes into a new commit.  As a general rule you want each commit to do one self-contained thing.

mods/ra/rules/vehicles.yaml
```diff
@@ -714,6 +714,12 @@ QTNK:
 	SelectionDecorations:
 		VisualBounds: 44,38,0,-4
 	MadTank:
+	Detonates:
+		ChargeDelay: 96
+		DetonationDelay: 42
+		ChargeSound: MADCHRG2.AUD
+		DetonationSound: MADEXPLO.AUD
```
GraionDilach: The sounds should be lowercase imo.

Unit158: That's unrelated to this pr

GraionDilach: It's not. It was lowercase in the MadTank trait you're removing. That's your change.

Unit158: If you want to change this everywhere, go ahead. It is actually more consistent with the rest of the mod this way. I'm not changing it.

Unit158: I'm not going to change this. Someone else can pick it up as a part of a greater cleanup if it means that much to them.

Unit158:
Someone else can complete the work here if there is any interest for this feature, I don't feel like it has been a valuable use of my time. There's not much left to do. Thanks for the help anyway.

mods/ra/rules/vehicles.yaml
pchote: There's no opinion here: these aud files are named lower case on disk, so changing the case breaks things for people with case sensitive file systems.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1607,6 +1607,43 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+					}
+
+					if (node.Value.Nodes.Any(n => n.Key == "MadTank"))
+					{
+						node.Value.Nodes.Add(new MiniYamlNode("Detonates", "", new List<MiniYamlNode>
+						{
+							new MiniYamlNode("ChargeDelay", "96"),
+							new MiniYamlNode("ChargeSound", "madchrg2.aud"),
+							new MiniYamlNode("DetonationDelay", "42"),
+							new MiniYamlNode("DetonationSound", "MadExplo.aud"),
```
pchote: Likewise here: you must be consistent with filename case on disk.  We don't all run windows!

Unit158: This is directly copy pasted from the original.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/8997.

Issue #7136, opened 2014-12-16
Lint failed missing audio


kyrreso:
http://resource.openra.net/maps/1895
Missed actor, did not fail lint...

PR #8388, opened 2015-06-05
Selectable bounds/selection box refactor

reaperrr:
- Introduced `VisualBounds` in `SelectionDecorations`, these are responsible for the visible selection box now (independent from actual selectable area)
- Moved `SelectionDecorations` and `SelectionBoxRenderable` to Mods.Common.
- selection boxes and selection bars can now be enabled/disabled individually, selection box color is now configurable.
- Set tighter `Selectable.Bounds` for many actors, to dramatically reduce issues where it was difficult to click on a cell or actor hidden behind another actor's selectable area.
- Removed `Selectable` boolean from `Selectable` trait.

This should pretty much fix #5552 and #3871 for the shipping mods. TS buildings will need isometric bounds to fix this completely.

PR #8412, opened 2015-06-06
Added a lint rule to detect missing voice definitions

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/7136.

### DIFF
--- a/OpenRA.Game/Selection.cs
+++ b/OpenRA.Game/Selection.cs
@@ -57,9 +57,21 @@ namespace OpenRA
 				}
 			}
 
-			var voicedActor = actors.FirstOrDefault(a => a.Owner == world.LocalPlayer && a.IsInWorld && a.HasTrait<IVoiced>());
-			if (voicedActor != null)
-				voicedActor.PlayVoice("Select");
+			// Play the selection voice from one of the selected actors
+			// TODO: This probably should only be considering the newly selected actors
+			// TODO: Ship this into an INotifySelection trait to remove the engine dependency on Selectable
+			foreach (var actor in actors)
+			{
+				if (actor.Owner != world.LocalPlayer || !actor.IsInWorld)
+					continue;
+
+				var selectable = actor.Info.Traits.GetOrDefault<SelectableInfo>();
+				if (selectable == null || !actor.HasVoice(selectable.Voice))
+					continue;
+
+				actor.PlayVoice(selectable.Voice);
+				break;
+			}
 
 			foreach (var a in newSelection)
 				foreach (var sel in a.TraitsImplementing<INotifySelected>())
--- a/OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
+++ b/OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
@@ -1,0 +1,64 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+using System.Reflection;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Lint
+{
+	public class CheckVoiceReferences : ILintPass
+	{
+		public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
+		{
+			foreach (var actorInfo in map.Rules.Actors)
+			{
+				foreach (var traitInfo in actorInfo.Value.Traits.WithInterface<ITraitInfo>())
+				{
+					var fields = traitInfo.GetType().GetFields().Where(f => f.HasAttribute<VoiceSetReferenceAttribute>());
+					foreach (var field in fields)
+					{
+						var voiceSets = LintExts.GetFieldValues(traitInfo, field, emitError);
+						foreach (var voiceSet in voiceSets)
+						{
+							if (string.IsNullOrEmpty(voiceSet))
+								continue;
+
+							CheckVoices(actorInfo.Value, emitError, map, voiceSet);
+						}
+					}
+				}
+			}
+		}
+
+		void CheckVoices(ActorInfo actorInfo, Action<string> emitError, Map map, string voiceSet)
+		{
+			var soundInfo = map.Rules.Voices[voiceSet.ToLowerInvariant()];
+
+			foreach (var traitInfo in actorInfo.Traits.WithInterface<ITraitInfo>())
+			{
+				var fields = traitInfo.GetType().GetFields().Where(f => f.HasAttribute<VoiceReferenceAttribute>());
+				foreach (var field in fields)
+				{
+					var voices = LintExts.GetFieldValues(traitInfo, field, emitError);
+					foreach (var voice in voices)
+					{
+						if (string.IsNullOrEmpty(voice))
+							continue;
+
+						if (!soundInfo.Voices.Keys.Contains(voice))
+							emitError("Actor {0} using voice set {1} does not define {2} voice required by {3}.".F(actorInfo.Name, voiceSet, voice, traitInfo));
+					}
+				}
+			}
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Lint/LintExts.cs
+++ b/OpenRA.Mods.Common/Lint/LintExts.cs
@@ -1,0 +1,33 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Reflection;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Lint
+{
+	public class LintExts
+	{
+		public static string[] GetFieldValues(object ruleInfo, FieldInfo fieldInfo, Action<string> emitError)
+		{
+			var type = fieldInfo.FieldType;
+			if (type == typeof(string))
+				return new string[] { (string)fieldInfo.GetValue(ruleInfo) };
+			if (type == typeof(string[]))
+				return (string[])fieldInfo.GetValue(ruleInfo);
+
+			emitError("Bad type for reference on {0}.{1}. Supported types: string, string[]"
+				.F(ruleInfo.GetType().Name, fieldInfo.Name));
+
+			return new string[] { };
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Traits/AttackMove.cs
+++ b/OpenRA.Mods.Common/Traits/AttackMove.cs
@@ -18,6 +18,8 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Provides access to the attack-move command, which will make the actor automatically engage viable targets while moving to the destination.")]
 	class AttackMoveInfo : ITraitInfo
 	{
+		[VoiceReference] public readonly string Voice = "Action";
+
 		public object Create(ActorInitializer init) { return new AttackMove(init.Self, this); }
 	}
 
@@ -27,16 +29,18 @@ namespace OpenRA.Mods.Common.Traits
 		public CPos? TargetLocation = null;
 
 		readonly IMove move;
+		readonly AttackMoveInfo info;
 
 		public AttackMove(Actor self, AttackMoveInfo info)
 		{
 			move = self.Trait<IMove>();
+			this.info = info;
 		}
 
 		public string VoicePhraseForOrder(Actor self, Order order)
 		{
 			if (order.OrderString == "AttackMove")
-				return "AttackMove";
+				return info.Voice;
 
 			return null;
 		}
--- a/OpenRA.Mods.Common/Traits/Harvester.cs
+++ b/OpenRA.Mods.Common/Traits/Harvester.cs
@@ -22,22 +22,34 @@ namespace OpenRA.Mods.Common.Traits
 	public class HarvesterInfo : ITraitInfo
 	{
 		public readonly string[] DeliveryBuildings = { };
+
 		[Desc("How much resources it can carry.")]
 		public readonly int Capacity = 28;
+
 		public readonly int LoadTicksPerBale = 4;
+
 		[Desc("How fast it can dump it's carryage.")]
 		public readonly int UnloadTicksPerBale = 4;
+
 		[Desc("How many squares to show the fill level.")]
 		public readonly int PipCount = 7;
+
 		public readonly int HarvestFacings = 0;
+
 		[Desc("Which resources it can harvest.")]
 		public readonly string[] Resources = { };
+
 		[Desc("Percentage of maximum speed when fully loaded.")]
 		public readonly int FullyLoadedSpeed = 85;
+
 		[Desc("Initial search radius (in cells) from the refinery that created us.")]
 		public readonly int SearchFromProcRadius = 24;
+
 		[Desc("Search radius (in cells) from the last harvest order location to find more resources.")]
 		public readonly int SearchFromOrderRadius = 12;
+
+		[VoiceReference] public readonly string HarvestVoice = "Action";
+		[VoiceReference] public readonly string DeliverVoice = "Action";
 
 		public object Create(ActorInitializer init) { return new Harvester(init.Self, this); }
 	}
@@ -270,7 +282,13 @@ namespace OpenRA.Mods.Common.Traits
 
 		public string VoicePhraseForOrder(Actor self, Order order)
 		{
-			return (order.OrderString == "Harvest" || (order.OrderString == "Deliver" && !IsEmpty)) ? "Move" : null;
+			if (order.OrderString == "Harvest")
+				return info.HarvestVoice;
+
+			if (order.OrderString == "Deliver" && !IsEmpty)
+				return info.DeliverVoice;
+
+			return null;
 		}
 
 		public void ResolveOrder(Actor self, Order order)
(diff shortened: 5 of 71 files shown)


OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
```diff
@@ -0,0 +1,58 @@
...
+		{
+			foreach (var actorInfo in map.Rules.Actors)
+			{
+				var voiceSet = string.Empty;
+
+				foreach (var traitInfo in actorInfo.Value.Traits.WithInterface<ITraitInfo>())
+				{
+					var fields = traitInfo.GetType().GetFields();
+					foreach (var field in fields)
+					{
```
pchote: Can drop these braces.

OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
```diff
@@ -0,0 +1,58 @@
...
+						if (field.HasAttribute<VoiceSetReferenceAttribute>())
+							voiceSet = LintExts.GetFieldValues(traitInfo, field, emitError).FirstOrDefault();
```
pchote: Should we give an error if an actor has more than one voice set defined?

OpenRA.Mods.Common/Traits/AttackMove.cs
```diff
@@ -18,6 +18,8 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Provides access to the attack-move command, which will make the actor automatically engage viable targets while moving to the destination.")]
 	class AttackMoveInfo : ITraitInfo
 	{
+		[VoiceReference] public readonly string AttackMoveVoice = "Move";
```
pchote: All these prefixes are redundant, since they exist inside their own trait namespace.  I suggest uniformly using `Voice`.

OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
Mailaender: Hadn't thought about that. Maybe a warning?

pchote:
Looks sensible overall, but lets converge on a consistent naming scheme for these new parameters before merging.

OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
penev92: Voices can easily be switched on upgrade, along with artwork. The code doesn't care about any voice set other than the first one at the moment, but could easily be changed to handle that case.
So multiple voice sets on a single actor sounds perfectly plausible to me.

pchote: Good point @penev92.  We will need to do the checks below for each voice set on the actor, then.

OpenRA.Mods.Common/Traits/Voiced.cs
```diff
@@ -19,7 +19,9 @@ namespace OpenRA.Mods.Common.Traits
 	public class VoicedInfo : ITraitInfo
 	{
 		[Desc("Which voice set to use.")]
-		[VoiceReference] public readonly string VoiceSet = null;
+		[VoiceSetReference] public readonly string VoiceSet = null;
+
+		[VoiceReference] public readonly string SelectVoice = "Select";
```
reaperrr: Maybe move this...

OpenRA.Mods.Common/Traits/Voiced.cs
```diff
@@ -37,6 +39,7 @@ public Voiced(Actor self, VoicedInfo info)
 		}
 
 		public string VoiceSet { get { return Info.VoiceSet; } }
+		public string SelectVoice { get { return Info.SelectVoice; } }
```
reaperrr: ...and this to the Selectable trait instead? Then you won't need the prefix, either.

penev92: Nah. We just split the voices off. I'd personally rather keep the voices in `Voiced`, even this one.

pchote: I agree with @reaperrr.  There's no need to make this a special case.

Mailaender:
Updated with the suggestions made.

OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
```diff
@@ -0,0 +1,56 @@
...
+			foreach (var actorInfo in map.Rules.Actors)
+			{
+				var voiceSet = string.Empty;
+
+				foreach (var traitInfo in actorInfo.Value.Traits.WithInterface<ITraitInfo>())
+				{
+					var fields = traitInfo.GetType().GetFields();
+					foreach (var field in fields)
+						if (field.HasAttribute<VoiceSetReferenceAttribute>())
+							voiceSet = LintExts.GetFieldValues(traitInfo, field, emitError).FirstOrDefault();
```
pchote: This still doesn't look right.  Any traits that are defined in the yaml before the `Voiced` trait will not be checked.

Mailaender: There is only the voiced trait so it does not matter.

pchote: Let me rephrase: The linter will not test any traits with `VoiceReference` fields that are defined on the actor above the `Voiced` trait definition.  The following will not be detected by the linter (try it yourself!):
```
AnyTraitThatDefinesVoices:
	Voice: Bogus
Voiced:
	VoiceSet: Whatever
```

Mailaender: VoiceSetReferenceAttribute != VoiceReferenceAttribute and I am looping through the fields twice here getting the VoiceSet first.

Mailaender: Oh wait, I am only looping through the fields of the same trait.

pchote: Looping over the fields twice only helps if the same trait defines both `VoiceSetReferenceAttribute` and `VoiceReferenceAttribute`.  The problem occurs for *traits* that are parsed before the first VoiceSetReference.

Try it yourself and see: add `Voice: Bogus` to the TS Wolverine's `AttackFrontal` trait.  The linter won't detect it because it is above `Voiced`, and the game will crash as soon as you try to attack something.

Mailaender: That was stupid. Updated accordingly.

Mailaender: You commented the wrong line. That confused me and is also the reason this comment won't hide although I fixed it.

Mailaender:
Had to rebase this due to MiniYaml file conflicts.

OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace OpenRA.Mods.Common.Lint
+{
+	public class CheckVoiceReferences : ILintPass
+	{
+		public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
+		{
+			foreach (var actorInfo in map.Rules.Actors)
+			{
+				var voiceSet = string.Empty;
```
penev92: This here (actually the line that sets it below, but that is too overcrowded already) assumes the actor will have only one voice set, and will ignore all but the first if there are multiple. Shouldn't be too hard to remove the `.FirstOrDefault()` from [line 31](https://github.com/OpenRA/OpenRA/pull/8412/files#diff-53fde2db49577802d0410cae9f4d5c2dR31) and add a `foreach(var voiceSet in voiceSets)` around lines [44](https://github.com/OpenRA/OpenRA/pull/8412/files#diff-53fde2db49577802d0410cae9f4d5c2dR44) through 53 (after you rename this to `voiceSets`, of course).

penev92:
All right! I like the cleanup of hardcoded strings very much.
Code looks good. I still haven't reviewed the YAML changes in detail.
Three things I noticed that need attention:
 - Fixing the upgrade rule.
 - Making the check for all voice sets on the actor.
 - Fixing `DeathSounds` in all 4 mods - you have an upgrade rule, but you didn't run it and the YAML still contains definitions like `DeathSound: Zapped`, which should now be `Voice: Zapped`.

Pinging all available reviewers, as this is a fairly big PR (don't let the numbers GitHub shows fool you, it's not *that* big) and conflicts with at least several other PRs, so would be better to get it merged ASAP.

OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
pchote: I second this request (actually this is exactly what I asked for in the first review pass, so you're seconding me :)).

Mailaender:
Updated.

Mailaender:
Rebased.

OpenRA.Game/Selection.cs
```diff
@@ -59,7 +59,11 @@ public void Combine(World world, IEnumerable<Actor> newSelection, bool isCombine
 
 			var voicedActor = actors.FirstOrDefault(a => a.Owner == world.LocalPlayer && a.IsInWorld && a.HasTrait<IVoiced>());
 			if (voicedActor != null)
-				voicedActor.PlayVoice("Select");
+			{
+				var selectable = voicedActor.TraitOrDefault<Selectable>();
+				if (selectable != null)
```
pchote: This should be merged into the `voicedActor` query to ensure that the selected actor will always play a voice.

OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
```diff
@@ -0,0 +1,69 @@
...
+		public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
+		{
+			foreach (var actorInfo in map.Rules.Actors)
+			{
+				foreach (var traitInfo in actorInfo.Value.Traits.WithInterface<ITraitInfo>())
+				{
+					var fields = traitInfo.GetType().GetFields();
+					foreach (var field in fields)
+					{
+						if (field.HasAttribute<VoiceSetReferenceAttribute>())
```
pchote: Style nit: You can avoid one block level here and in `CheckVoices` by adding `.Where(f => f.HasAttribute<VoiceSetReferenceAttribute>())` to the end of the `fields` definition.

OpenRA.Mods.Common/Lint/LintExts.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System;
+using System.Reflection;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Lint
+{
+	public class LintExts
+	{
+		public static string[] GetFieldValues(ITraitInfo traitInfo, FieldInfo fieldInfo, Action<string> emitError)
```
pchote: It doesn't look like you depend on anything `ITraitInfo` specific in here, so maybe change this to `object`?

mods/cnc/rules/infantry.yaml
```diff
@@ -15,6 +15,7 @@ E1:
 	Armament:
 		Weapon: M16
 	AttackFrontal:
+		Voice: Move
```
pchote: It may be cleaner to define an Attack voice instead of overriding this everywhere.

pchote: Otherwise, `Move` should really be renamed to something more generic like `Action`.

OpenRA.Mods.Common/Lint/LintExts.cs
penev92: Yes, I actually copied this method for my upgrades lint check and had to change this to `object`.
So feel free to change or leave it - it's going to change anyway.

OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -270,7 +281,7 @@ public Order IssueOrder(Actor self, IOrderTargeter order, Target target, bool qu
 
 		public string VoicePhraseForOrder(Actor self, Order order)
 		{
-			return (order.OrderString == "Harvest" || (order.OrderString == "Deliver" && !IsEmpty)) ? "Move" : null;
+			return (order.OrderString == "Harvest" || (order.OrderString == "Deliver" && !IsEmpty)) ? info.Voice : null;
```
GraionDilach: Does this mean you'd have the same (say, "Harvest") definition played when you order the harvester to harvest a resource and when you order this harvester to return with said resource? RA2 already had both Harvest and Return separated.

penev92: For the moment - yes. Looks like it won't be any problem separating them, but I'll leave that up to @Mailaender, as this PR isn't small or simple, and that's not really related.

mods/cnc/rules/infantry.yaml
Mailaender: > It may be cleaner to define an Attack voice instead of overriding this everywhere.

That will add redundancy.

> Otherwise, `Move` should really be renamed to something more generic like `Action`.

I decided against it for compatibility reasons and pull request scope.

OpenRA.Mods.Common/Lint/LintExts.cs
Mailaender: Changed it to `object` already.

OpenRA.Mods.Common/Traits/Harvester.cs
Mailaender: I separated those while I was at it.

Mailaender:
Updated as requested.

GraionDilach:
TD mod still has DeathSound defined.

Mailaender:
Fixed.

GraionDilach:
Cannot find any other issues.

mods/cnc/rules/infantry.yaml
pchote: You're arguing 21 that duplicated `Voice: Move` definitions is less redundant than vs 1 duplicated voice definition.

OpenRA.Game/Selection.cs
```diff
@@ -57,9 +57,13 @@ public void Combine(World world, IEnumerable<Actor> newSelection, bool isCombine
...
-			var voicedActor = actors.FirstOrDefault(a => a.Owner == world.LocalPlayer && a.IsInWorld && a.HasTrait<IVoiced>());
-			if (voicedActor != null)
-				voicedActor.PlayVoice("Select");
+			var voicedActors = actors.Where(a => a.Owner == world.LocalPlayer && a.IsInWorld && a.HasTrait<IVoiced>());
+			var selectableActors = voicedActors.ToDictionary(a => a, v => v.TraitOrDefault<Selectable>());
+			foreach (var selectable in selectableActors.Where(v => v.Value != null))
+			{
+				selectable.Key.PlayVoice(selectable.Value.Info.Voice);
+				break;
+			}
```
pchote: This can be simplified to 
```csharp
// Play the selection voice from one of the selected actors
// TODO: This probably should only be considering the newly selected actors
// TODO: Ship this into an INotifySelection trait to remove the engine dependency on Selectable
foreach (var a in actors)
{
	if (a.Owner != world.LocalPlayer || !a.IsInWorld)
		continue;

	var selectable = a.Info.Traits.GetOrDefault<SelectableInfo>();
	if (selectable == null || !a.HasVoice(selectable.Voice))
		continue;

	a.PlayVoice(selectable.Voice); 
	break;
}
```

penev92: +1 to that.
But change `a in actors` to `a in newSelection` and remove the first `TODO`

OpenRA.Mods.Common/Lint/CheckVoiceReferences.cs
```diff
@@ -0,0 +1,63 @@
...
+				var fields = traitInfo.GetType().GetFields().Where(f => f.HasAttribute<VoiceReferenceAttribute>());
+				foreach (var field in fields)
+				{
+					var voices = LintExts.GetFieldValues(traitInfo, field, emitError);
+					foreach (var voice in voices)
+					{
+						if (string.IsNullOrEmpty(voice))
+							continue;
+
+						var soundInfo = map.Rules.Voices[voiceSet.ToLowerInvariant()];
```
penev92: Minor nit: This can easily go to the top of the method.

penev92:
I like the code changes and cleanup.
But after reading the entire diff I have to agree with @pchote's earlier comment about the `Attack` voice - would look better to have such a voice in TD and TS (and bring it back for D2k and RA). A lot of YAML duplication aside, it defeats the purpose of having a default value if **everything** is going to override it.
Lint check works, and I can still hear stuff ingame (wow, what a surprise :D ), so :+1: with the above comments.

pchote:
If none of the mods use separate attack and move voices for the generic units (heros are of course an exception) then my earlier suggestion of changing the defaults to "Action" and removing all the overrides seems like the best approach.

penev92:
Needs a rebase now that #8388 was merged. Sorry.

GraionDilach:
TS uses separate voices for Attack and Move. Notsure on D2k.

Mailaender:
Rebased and special cased Tiberian Sun instead of Dune 2000, Red Alert, Tiberian Dawn. :sweat:

reaperrr:
Looks good to me. I didn't test it thoroughly, but the move/attack differentiation in TS worked, and I didn't hear any obvious regressions in TD, D2k and RA either.

:+0.9: from me.

pchote:
I haven't tested this, but this looks like a really nice improvement over what we currently have.

Mailaender:
In summary: it removes magic behavior, adds automatic checks so all the rule refactors that almost always introduced hard to spot regressions will get tested. I would like to see this in before we do more large scale MiniYaml changes.

obrakmann:
Looks ok to me, :+1: 

A future PR could also add checks to see if the VoiceSet exists at all (currently there wouldn't be any message at all).

Mailaender:
I think we already do that: https://github.com/OpenRA/OpenRA/pull/8412/files#diff-2908d570caac7d047ea2dd3ea940cb28R40

obrakmann:
Whoops, missed that. Even better then.

[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/fbdc527380279b68739b43a61be3a2c7b4eb956b%5E...fbdc527380279b68739b43a61be3a2c7b4eb956b)

PR #8416, opened 2015-06-07
Added an Infiltrate function to lua

abcdefg30:


OpenRA.Mods.Common/Scripting/Properties/CaptureProperties.cs
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	[ScriptPropertyGroup("Targeting")]
+	public class ExternalCaptureProperties : ScriptActorProperties, Requires<ExternalCapturesInfo>
+	{
+		public ExternalCaptureProperties(ScriptContext context, Actor self)
+			: base(context, self) { }
+
+		[Desc("Captures the target actor using external-capturing logic.")]
+		public void ExternalCapture(Actor target)
```
pchote: It would be nicer if this was also named `Capture`, the same way that `Mobile` and `Aircraft` expose alternative `Move` script properties, and the different versions of `Attack`.

abcdefg30:
Updated.

OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
```diff
@@ -198,6 +198,7 @@
     <Compile Include="Scripting\Global\FacingGlobal.cs" />
     <Compile Include="Scripting\Global\HSLColorGlobal.cs" />
     <Compile Include="Scripting\Global\UserInterfaceGlobal.cs" />
+    <Compile Include="Scripting\Properties\CaptureProperties.cs" />
```
Mailaender: I suggest to move this into a separate pull request as it doesn't provide a test case yet and is independent from infiltrate.

OpenRA.Mods.Common/Scripting/Properties/CaptureProperties.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System.Linq;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Scripting;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Scripting
+{
+	[ScriptPropertyGroup("Targeting")]
```
obrakmann: Grouping capturing and infiltration under "Targeting" somehow rubs me the wrong way, but I don't have any better suggestions, either.

OpenRA.Mods.RA/Scripting/Properties/InfiltrateProperties.cs
```diff
@@ -0,0 +1,29 @@
...
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.Linq;
```
obrakmann: I think this isn't needed.

obrakmann:
Looks good to me otherwise. :+1:

OpenRA.Mods.RA/Scripting/Properties/InfiltrateProperties.cs
abcdefg30: Fixed.

OpenRA.Mods.Common/Scripting/Properties/CaptureProperties.cs
Mailaender: We call it special abilities in the tooltips so maybe `Ability`?

penev92:
I have to agree with the naming nit above about "Targeting".

abcdefg30:
Updated. Renamed `Targeting` to `Ability`.
> I suggest to move this into a separate pull request as it doesn't provide a test case yet and is independent from infiltrate.

Done: #8614

Mailaender:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/cb06d675a075bb7fe43ed1aa65b7841257550944...1a1d0246e9d44e9d6b86daba4c1412e4ec3ed2f2)

PR #8417, opened 2015-06-07
Added DisguiseAs and DisguiseAsType functions to lua

abcdefg30:


OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -140,6 +140,24 @@ void DisguiseAs(Actor self, Actor target)
 				t.OnEffectiveOwnerChanged(self, oldEffectiveOwner, AsPlayer);
 		}
 
+		public void DisguiseAs(string actorType, Player newOwner)
```
pchote: The API would be a bit cleaner if this took an `ActorInfo` instead of a `string`.

pchote: This should also have an `Actor self` argument for consistency with everything else.

Phrohdoh:
You're asking for an NRE but not doing a null check with TraitOrDefault.

OpenRA.Mods.RA/Scripting/Properties/DisguiseProperties.cs
```diff
@@ -0,0 +1,35 @@
...
+		[Desc("Disguises as the target actor.")]
+		public void DisguiseAs(Actor target)
+		{
+			Self.World.IssueOrder(new Order("Disguise", Self, true) { TargetActor = target });
+		}
+
+		[Desc("Disguises as the target type with the specified owner.")]
+		public void DisguiseAsType(string actorType, Player newOwner)
+		{
+			Self.TraitOrDefault<Disguise>().DisguiseAs(actorType, newOwner);
```
pchote: This should use .Get (no OrDefault) and the class Requires on this trait.

abcdefg30:
Updated.

abcdefg30:
Well, updated. But not sure if it's like we want it now. :worried:

OpenRA.Mods.RA/Scripting/Properties/DisguiseProperties.cs
```diff
@@ -0,0 +1,42 @@
...
+#endregion
+
+using System.Linq;
+using OpenRA.Mods.RA.Traits;
+using OpenRA.Scripting;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.RA.Scripting
+{
+	[ScriptPropertyGroup("Targeting")]
```
penev92: I believe this was changed to `Ability` for the capturing and infiltration.

penev92:
This will actually `[Sync]` the wrong thing when disguising the spy via the
```public void DisguiseAs(Actor self, ActorInfo actorInfo, Player newOwner)```
method (used from Lua), because that does not set `[Sync] Actor targetActor;` (and how could it, there is no target in that case).
I see two ways out of this:
 1. Make `[Sync]` work with strings. (I hear that it doesn't, but I have no idea why it wouldn't).
 2. Drop the `DisguiseAsType` method and use an actual actor to disguise as. (The much simpler way out, but much less preferable. However, syncing of strings is not for this PR.)

**Comment by abcdefg30** [comment]
Rebased and removed the `Sync` stuff.

**Comment by penev92** [comment]
Looks good to me :+1:

File: OpenRA.Game/GameRules/ActorInfo.cs [hunk 4]
```diff
@@ -18,7 +18,7 @@ namespace OpenRA
 {
 	/// <summary>
 	/// A unit/building inside the game. Every rules starts with one and adds trait to it.
-	/// Special actors like world or player are usually defined in system.yaml and affect everything.
+	/// Special actors like world or player are usually defined in world.yaml and player.yaml and affect everything.
```
pchote: Lets please remove this line.  Engine code shouldn't be commenting about individual mod file layout choices.

pchote:
Looks fine to me aside from that one nit.

abcdefg30:
Fixed and rebased.

mods/ra/maps/allies-05a/allies05a.lua
```diff
@@ -247,9 +245,8 @@ InitTriggers = function()
 		if a == Truk then
 			Trigger.RemoveFootprintTrigger(id)
 
-			CloakProvider = Actor.Create("CloakUpgrade", true, { Owner = greece, Location = Prison.Location })
-
 			Spy = Actor.Create("spy", true, { Owner = greece, Location = TrukWaypoint5.Location })
+			Spy.DisguiseAsType("e1", ussr)
```
obrakmann: The spy doesn't show that recently added decoration at this point.

abcdefg30: Should be fixed now.

OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -177,6 +177,14 @@ void DisguiseAs(Actor target)
 			}
 		}
 
+		public void DisguiseAs(ActorInfo actorInfo, Player newOwner)
+		{
+			var renderSprites = actorInfo.Traits.GetOrDefault<RenderSpritesInfo>();
+			AsSprite = renderSprites == null ? null : renderSprites.GetImage(actorInfo, self.World.Map.SequenceProvider, newOwner.Country.Race);
+			AsPlayer = newOwner;
+			AsTooltipInfo = actorInfo.Traits.WithInterface<TooltipInfo>().FirstOrDefault();
+		}
```
obrakmann: I guess you need to duplicate/refactor into a new method all the stuff that the other `DisguiseAs` method does in addition to changing the `As*` variables.

OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -177,6 +177,34 @@ void DisguiseAs(Actor target)
...
+				{
+					if (!um.Value.AcknowledgesUpgrade(self, u))
+						continue;
+
+					if (Disguised)
+						um.Value.GrantUpgrade(self, u, this);
+					else
+						um.Value.RevokeUpgrade(self, u, this);
+				}
+			}
```
obrakmann: Can you factor the duplicated parts out into a separate private method, please? Thanks.

abcdefg30:
Updated again.

Mailaender:
Looking good. :white_check_mark: I wasn't even aware of that hack and wonder why it was necessary before.

abcdefg30:
>  I wasn't even aware of that hack and wonder why it was necessary before.

Otherwise the "guards" would have started to hunt/attack the spy after he exited the truck.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/012b02ec4d3c7247676f8cbd23393689843cd6a5...00d9e0b844665a52bdefae206a174d32a3154aaa)

PR #8418, opened 2015-06-07
Added even more lua functions

abcdefg30:
This inculdes functions to:
- set beacons
- enter transports
- ~~demolish actors~~ #8576

### DIFF
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -213,6 +213,7 @@
     <Compile Include="Scripting\Media.cs" />
     <Compile Include="Scripting\ScriptTriggers.cs" />
     <Compile Include="Scripting\Global\ActorGlobal.cs" />
+    <Compile Include="Scripting\Global\BeaconGlobal.cs" />
     <Compile Include="Scripting\Global\CameraGlobal.cs" />
     <Compile Include="Scripting\Global\CoordinateGlobals.cs" />
     <Compile Include="Scripting\Global\DateTimeGlobal.cs" />
--- a/OpenRA.Mods.Common/Scripting/Global/BeaconGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/BeaconGlobal.cs
@@ -1,0 +1,47 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.Mods.Common.Effects;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Scripting;
+
+namespace OpenRA.Mods.Common.Scripting
+{
+	[ScriptGlobal("Beacon")]
+	public class BeaconGlobal : ScriptGlobal
+	{
+		readonly RadarPings radarPings;
+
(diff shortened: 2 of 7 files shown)


OpenRA.Mods.Common/Scripting/Global/BeaconGlobal.cs
```diff
@@ -0,0 +1,31 @@
...
+	public class BeaconGlobal : ScriptGlobal
+	{
+		public BeaconGlobal(ScriptContext context) : base(context) { }
+
+		[Desc("Creates a new beacon that stays for the specified time at the specified WPos." +
+			"Does not remove player set beacons, nor gets removed by placing them.")]
+		public Beacon New(Player owner,  WPos position, int duration = 30 * 25, string palettePrefix = "player")
+		{
+			var playerBeacon = new Beacon(owner, position, duration, palettePrefix);
+			owner.PlayerActor.World.AddFrameEndTask(w => w.Add(playerBeacon));
```
pchote: This will also miss the radar ping and "press spacebar to go to beacon" input.

abcdefg30: Hm, I should fix the radar ping, but do we want "press spacebar to go to beacon" (as there could be multiple beacons and just jumping to the last one sounds unpolished to me)?

Mailaender:
Possible test cases:
* Demolition: the jeep in the beginning of soviet-01
* Beacon: allies-02 when the trucks arrive.

OpenRA.Mods.Common/Scripting/Global/BeaconGlobal.cs
abcdefg30: Whatever, you are able to use the spacebar jump when using the radar pings.

abcdefg30:
Updated.

> Possible test cases:
- Demolition: the jeep in the beginning of soviet-01
- Beacon: allies-02 when the trucks arrive.

- The bridge is actually blown up by the barrel explosion there. I think using the jeep to blow it up would be a bigger hack then just calling `bridge.Kill()`.
- Done (also in survival01).

mods/ra/maps/allies-02/allies02.lua
```diff
@@ -70,6 +70,7 @@ SendTrucks = function()
 	if not ConvoyOnSite then
 		ConvoyOnSite = true
 		ticked = 0
+		Beacon.New(player, TruckEntryPoint.CenterPosition + WVec.New(0, 3 * 1024, 0))
```
obrakmann: This is ok as a testcase, but I think a bit pointless without a radar.

obrakmann:
besides that, :+1:

abcdefg30:
Updated. Removed the beacon from allies02 and added one to survival02.

OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
```diff
@@ -230,6 +231,7 @@
     <Compile Include="Scripting\Properties\GuardProperties.cs" />
     <Compile Include="Scripting\Properties\MissionObjectiveProperties.cs" />
     <Compile Include="Scripting\Properties\MobileProperties.cs" />
+    <Compile Include="Scripting\Properties\DemolitionProperties.cs" />
```
Mailaender: Then again just put this into a separate pull request so we can merge it later with a test case.

abcdefg30: Done: #8576

Mailaender:
Works as promised. :white_check_mark: Again, nice expansions to the Lua API.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/da62830b04488d0e2e62abc2d1a0f48814128ffb...78002d14d8191d3f42e701733a5bb0d0a9c51467)

PR #8421, opened 2015-06-07
Fixed a bad bit shift operation in PriorityQueue

Mailaender:
`--level` can be -1 here. This leads to undefined behavior. https://msdn.microsoft.com/en-us/library/f96c63ed.aspx

Discovered using Coverity.

### DIFF
--- a/OpenRA.Game/Primitives/PriorityQueue.cs
+++ b/OpenRA.Game/Primitives/PriorityQueue.cs
@@ -83,7 +83,7 @@ namespace OpenRA.Primitives
 
 			var ret = At(0, 0);
 			BubbleInto(0, 0, Last());
-			if (--index < 0)
+			if (level > 0 && --index < 0)
 				index = (1 << --level) - 1;
 
 			return ret;


abcdefg30:
:+1: / :white_check_mark:

Phrohdoh:
This can't happen now.
If index is 1 (>0) --index is 0 (== 0).

Mailaender:
This is about `level` not index. See https://scan8.coverity.com/reports.htm#v24414/p10888/fileInstanceId=6103897&defectInstanceId=2339176&mergedDefectId=111058

Phrohdoh:
Oh, ignore me then, sorry.

penev92:
For some reason I couldn't get Coverity to log me in with my GH account.
Can you give the details of the warning here?
I am not convinced this is the best fix, given the `if () throw` above.

obrakmann:
This is a false positive. [This code](https://github.com/OpenRA/OpenRA/blob/12966129a4e412fd2c53eb05b310f30557058521/OpenRA.Game/Primitives/PriorityQueue.cs#L54-L57) in `Add()` ensures that as long as the queue has at least one entry, `level` is greater than 0.

PR #3727, opened 2013-08-22
Colorpicker crash fix

pchote:
This replaces the front and back bitmaps with raw byte arrays. It doesn't address the underlying threading issue in #3696 (I don't see why that fails and can't reproduce it), but this changes the failure mode to be at worst a visual tear instead of crashing.

PR #8425, opened 2015-06-07
Fixed a bad choice of lock objects in ColorMixerWidget

Mailaender:
~~Reverts https://github.com/OpenRA/OpenRA/pull/3727 and~~ fixes it properly. Detected as a high impact outstanding issue by Coverity due to nondeterministic locking behavior.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ColorMixerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ColorMixerWidget.cs
@@ -31,7 +31,8 @@ namespace OpenRA.Mods.Common.Widgets
 		bool isMoving;
 
 		bool update;
-		object syncWorker = new object();
+		readonly object syncWorker = new object();
+		readonly object bufferSync = new object();
 		Thread workerThread;
 		bool workerAlive;
 
@@ -111,7 +112,7 @@ namespace OpenRA.Mods.Common.Widgets
 						}
 					}
 
-					lock (front)
+					lock (bufferSync)
 					{
 						var swap = front;
 						front = back;
@@ -123,7 +124,7 @@ namespace OpenRA.Mods.Common.Widgets
 
 		public override void Draw()
 		{
-			if (Monitor.TryEnter(front))
+			if (Monitor.TryEnter(bufferSync))
 			{
 				try
 				{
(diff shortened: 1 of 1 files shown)


pchote:
We don't use or need any of the overhead of the `Bitmap`.  Could you please keep the byte arrays?

Mailaender:
Removed `Bitmap` again.

chrisforbes:
Unconvinced, from a quick inspection. I need to look at this more closely.

Mailaender:
See https://scan8.coverity.com/reports.htm#v24413/p10888/fileInstanceId=6104409&defectInstanceId=2339175&mergedDefectId=111052

RoosterDragon:
I don't have the benefit of seeing the Coverity results, but is this being flagged because we're assigning to the ```front``` and ```back``` variables whilst holding locks on them? I mean, that smells fishy to me at least.

Assuming it is something along those lines, the dedicated lock object makes sense, but the locking on the ```back``` array would still seem out of place.

Mailaender:
According to https://scan.coverity.com/projects/3650?tab=project_settings everyone should have access:

image

RoosterDragon:
I assume this is because I lack an account on their site (I'll live, I'm sure).

OpenRA.Mods.Common/Widgets/ColorMixerWidget.cs
```diff
@@ -32,6 +32,7 @@ public class ColorMixerWidget : Widget
 
 		bool update;
 		object syncWorker = new object();
+		object frontWorker = new object();
```
RoosterDragon: I would rename this, maybe ```bufferLock``` or ```bufferSync``` but I'm open to other suggestions. ```frontWorker``` makes little sense though.

RoosterDragon: Maybe throw some ```readonly``` modifiers on this and ```syncWorker``` too since dedicated lock objects never change.

RoosterDragon:
This change makes sense. Reassigning the lock object from within the lock is pure madness. You need a dedicated object.

:+1: after style comments are addressed/dismissed.

Mailaender:
Updated.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d05507e86e66498a3a45464f0ad5107bf32c56f3%5E...d05507e86e66498a3a45464f0ad5107bf32c56f3)

PR #8429, opened 2015-06-07
Configure dl agent

moviuro:
`curl(1)` is not available in a base FreeBSD install. As it has a counterpart (`fetch(1)`), we use it if curl is not available.

Tested and fully functionnal on FreeBSD 10.1

### DIFF
--- a/thirdparty/noget.sh
+++ b/thirdparty/noget.sh
@@ -1,8 +1,10 @@
 #!/bin/sh
 # fallback without dependency resolution if nuget is not present on the system
 
+. ../configure-utils.sh
+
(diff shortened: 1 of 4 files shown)


thirdparty/noget.sh
```diff
@@ -1,8 +1,10 @@
 #!/bin/sh
 # fallback without dependency resolution if nuget is not present on the system
 
+. ../configure-utils.sh
```
Mailaender: > ../../noget.sh: 4: .: Can't open ../configure-utils.sh

moviuro:
Abandonned

PR #8440, opened 2015-06-08
Expanded the bug reporting guidelines

Mailaender:
A humble wish.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,10 @@
 
 * Have you read the [FAQ](https://github.com/OpenRA/OpenRA/wiki/FAQ)?
 * Add the appropriate log files on crashes.
+* Mention the version you have tested against.
 * Please be specific on how to reproduce the problem.
+* Do not submit duplicate issues. Search first.
+* Avoid including multiple requests/bugs into a single issue.
 
 ## Patches
 


CONTRIBUTING.md
```diff
@@ -5,6 +5,8 @@
 * Have you read the [FAQ](https://github.com/OpenRA/OpenRA/wiki/FAQ)?
 * Add the appropriate log files on crashes.
 * Please be specific on how to reproduce the problem.
+* Do not submit duplicate issues. Search first.
+* Avoid including few requests/bugs into a single issue.
```
penev92: "few" -> "several"

penev92:
:+1: although I doubt most people read those. :/

Phrohdoh:
s/few/multiple

clemty:
Perhaps tell people to say which version they are using? (to avoid reports for long-fixed bugs in bleed because people use the last release)

Mailaender:
Updated.

> Perhaps tell people to say which version they are using?

The situation of not mentioning the version has improved since https://github.com/OpenRA/OpenRA/pull/3938, but nonetheless that is important and often forgotten and a problem when no logs are provided. Added.

obrakmann:
:+1:

PR #8441, opened 2015-06-08
Adds a chat logging feature.

### LINKED ISSUE #8136: Log ingame chat
Ingame chat could be logged in a text file in \OpenRA\Logs folder.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/14103114-log-ingame-chat?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

deniz1a:
Ingame chat is logged in chat.log file with the following format:
10:50:42 <Server> hello has joined the game.
10:51:01 <hello> hello.

Fixes #8136.

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -204,6 +204,7 @@ namespace OpenRA
 			Log.AddChannel("sound", "sound.log");
 			Log.AddChannel("graphics", "graphics.log");
 			Log.AddChannel("geoip", "geoip.log");
+			Log.AddChannel("chat", "chat.log");
 
 			if (Settings.Server.DiscoverNatDevices)
 				UPnP.TryNatDiscovery();
@@ -665,6 +666,9 @@ namespace OpenRA
 
 		public static void AddChatLine(Color color, string name, string text)
 		{
+			var t = DateTime.Now;
+			var time = "{0:D4}-{1:D2}-{2:D2} {3:D2}:{4:D2}:{5:D2}".F(t.Year, t.Month, t.Day, t.Hour, t.Minute, t.Second);
+			Log.Write("chat", time + " <" + name + "> " + text);
 			OrderManager.AddChatLine(color, name, text);
 		}
 


Phrohdoh:
Please add Y-M-d

penev92:
Personally I'd prefer the ability to extract chat from a replay for several reasons:
- No huge (think gigabytes) log files that people don't even know about (and usually won't need).
- No additional slowdown for people with less-than-fast HDDs. (Writing to disk is usually a relatively slow task)

Phrohdoh:
Is there a real use for this?
Most users don't know the logs exist (much less where they are at).

penev92:
One day it may be useful for reporting players, I guess (when we have authentication etc.)

Phrohdoh:
Unfortunately there is no validity to a plaintext file.

penev92:
>penev|m: no, but if you report via replay and the moderator can extract the chat, that would be helpful
penev|m: Which is actually not doable with on-the-spot logging

deniz1a:
Added Year-Month-Day. Now the log has this format:
2015-06-08 10:50:42 \<Server\> hello has joined the game.
2015-06-08 10:51:01 \<hello\> hello.

deniz1a:
* Can this chat.log get to a gigabyte size? How could its size be restricted with a limit?
* There could be an option in settings menu to turn chat logging on and off.
* To make users aware of log files, there could be a log files menu under extras that displays contents of log files. Or there could be a detailed option in settings menu about logging.
* Would writing chat lines to a text file affect performance in any noticable way?
* Extraction of chat content from replays could be a separate feature.

x-a-n-a-x:
IMO displaying the month, year and day is useless.. You already have this information in the Replay filename. Plus, it would take quite a significant space if you add them. Please, only display time.

:-1: for me if Year-Month-Day is added.

deniz1a:
@x-a-n-a-x : This PR is not about replays but for general chat logging in game.

x-a-n-a-x:
I know, my argument was especially about the lobby chat. It would take too much space with the date.

x-a-n-a-x:
Hmm.. ok

deniz1a:
Ok, should it display Y-M-D or not? I'm ok with either.

cjshmyr:
Is this for dedicated server owners? Would it make more sense to just be spitting the chat messages in a `Console.Writeline` so anything that's reading the stdout of the process can see it real time?

deniz1a:
I actually thought this for normal players as a convenience feature. So you can look at the log of what was said. But dedicated servers could also use it if there is a need. I don't know if it currently logs in dedicated server mode.

cjshmyr:
> Can this chat.log get to a gigabyte size? How could its size be restricted with a limit?

Clear it before a game begins

From a player perspective, this is possibly fine (albeit I've never cared about a chat log after hundreds of games)
From a server owner perspective, this log will be very hard to use since we spin up multiple instances of the same executable, resulting in multiple chat.log files. This gets *a lot* more complicated when we decide we want to log more than just the chat.
Lastly from a project maintainer perspective, I want to avoid design that only satisfies one of these two user bases if possible. I'm sorry I never gave this feedback on the original ticket, so I remain indifferent on this for now.

deniz1a:
Those issues could probably be fixed by a detailed logging section in settings menu. Individual logging could be turned on and off, logging behavior could be set (reset after each game, custom size limit, where to save the log etc.).

deniz1a:
Is the only problem with dedicated servers that there would be multiple chat.log files? That's the case for all log files.

OpenRA.Game/Game.cs
```diff
@@ -665,6 +666,9 @@ public static void Restart()
 
 		public static void AddChatLine(Color color, string name, string text)
 		{
+			var t = DateTime.Now;
+			var time = "{0:D4}-{1:D2}-{2:D2} {3:D2}:{4:D2}:{5:D2}".F(t.Year, t.Month, t.Day, t.Hour, t.Minute, t.Second);
```
Mailaender: You can make this easier to read by using `DateTime.Now.ToString("YYYY-MM-dd HH:mm:ss")`. See https://msdn.microsoft.com/en-us/library/zdtaw1bw(v=vs.110).aspx or https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Game.cs#L59

deniz1a:
Ok, @pchote and @penev92 suggested adding a feature to the utility that extracts chat content from replays. Since replays already contain this information that's probably better. The only issue with that would be that the replays containing only lobby chat and no game are not shown in the replay browser. Maybe there should be a checkbox saying "only show replays containing games" or something...

deniz1a:
I'm closing this now and will try to implement chat extraction from replays in a different PR.

PR #8196, opened 2015-05-20
Fixed StyleCop on Mono 4

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/8015.

We need MSBuild in the .NET Framework 3.5 version. Mono 4 dropped those so we need to manually install them as a dependency now. I consider this a workaround until we replace this with something modern probably https://github.com/dotnet/roslyn based in the future.

PR #8445, opened 2015-06-08
Package the *.exe.config files on all platforms

obrakmann:
I just noticed that the packaging of the .exe.config files was inconsistent.

Windows and Mac OS X only shipped the Game.exe.config, not the Utility.exe.config.
Debian shipped neither.

The Game.exe.config is currently only used on Windows, while the Utility.exe.config is only used on systems using mono.

I opted to ship all .configs on all platforms. If someday someone decides to add something to those files, it would just be forgotten to add them to the packages.


Mailaender:
I first thought they are useless on platforms other than Windows, but I myself filed https://github.com/OpenRA/OpenRA/pull/8196 although not too as these StyleCop runs are only useful during development. Whatever. :+1:

obrakmann:
Oh well, I just looked at the title of the commit and didn't remember that it had to do with running StyleCop. It doesn't matter then.

PR #8430, opened 2015-06-07
Display and support editing the area outside the map bounds

pchote:
Fixes one of the few limitations that the new editor has vs the old.  Also fixes the missing tiles that used to appear at the right/bottom edges of the screen in TS (I was sure we had a ticket on this, but can't find it).

Depends on #8337.

PR #8447, opened 2015-06-08
Offset the TS shroud sequences.

pchote:
For some reason WW thought it was a good idea to offset the shroud/fog sprites by 1px vertically relative to the terrain tiles (see screenshot).  This applies the opposite offset in our sequences to line them up again.  The easiest way to test this is to look at the edges of the map in #8430.

offset


obrakmann:
:+1:

PR #8101, opened 2015-05-07
Cleanup unused DLL references

penev92:
With the effort of moving shared code to `Mods.Common` apparently done, there are now **no** (at least explicit) class dependencies between any of the mod DLLs.
So here is a little unused references cleanup.

PR #8455, opened 2015-06-09
Remove old dependencies from Makefile.

phrohdoh:
Followup of https://github.com/OpenRA/OpenRA/pull/8101.


penev92:
Seems reasonable :+1:

Mailaender:
Well spotted.

PR #8460, opened 2015-06-11
Enabled and documented DEBUG option on Unix startup scripts

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/8282.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -392,7 +392,11 @@ install-linux-appdata:
 install-linux-scripts:
 	@echo "#!/bin/sh" > openra
 	@echo 'cd "$(gameinstalldir)"' >> openra
+ifeq ($(DEBUG), $(filter $(DEBUG),false no n off 0))
 	@echo 'mono OpenRA.Game.exe "$$@"' >> openra
+else
+	@echo 'mono --debug OpenRA.Game.exe "$$@"' >> openra
+endif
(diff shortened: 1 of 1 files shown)


pchote:
Wouldn't it be better to just include the `--debug` all the time?

Mailaender:
It defaulting to true should be enough.

Makefile
```diff
@@ -1,10 +1,10 @@
 ############################# INSTRUCTIONS #############################
 #
 # to compile, run:
-#   make
+#   make [DEBUG=true]
```
obrakmann: People might assume that the default is false when you write it like that. I'd probably either invert that example usage, or add a line that states the default value.

Mailaender: Inverted.

obrakmann:
:+1:

Mailaender:
Rebased after https://github.com/OpenRA/OpenRA/pull/8511 was merged prior to this.

pchote:
Looks fine to me, although these changes don't apply to my platform

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d14896d84c8762d71afba17d5cd169233494dcf6%5E...d14896d84c8762d71afba17d5cd169233494dcf6

Issue #8434, opened 2015-06-07
Using alternate palettes for actor icons.


scskwirl:
There is a hard-coded chrome palette that runs the icons.

When using assets between mods, there is no current method to override this and read from another palette to keep the correct color.

Use a trait inside Buildable: for each item that will need the effect?

Phrohdoh:
I can't find a hard-coded palette anywhere for production icons.

GraionDilach:
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Widgets/WidgetUtils.cs#L30 sounds like such to me.

Issue #3309, opened 2013-05-17
PCX support for individual cameos


Deformat:
Basically, allow a cameo to have a specific palette, or allow it to be loaded as a .PCX file.

Mailaender:
> allow a cameo to have a specific palette

Reverse duplicate of https://github.com/OpenRA/OpenRA/issues/8464.

pchote:
We now support per-icon palettes, which should be sufficient.

Mailaender:
I guess the specific request for PCX is because of legacy compatibility with RA2 mods?

GraionDilach:
Kinda but those RA2 modders should know that any cameo PCX they have can be separated into an SHP-PAL combination with XCC Mixer in four clicks. I'm fine with this.

PR #8464, opened 2015-06-12
Dehardcode the chrome palette from icons

GraionDilach:
Adds an IconPalette property to Buildable and SupportPower traits. 

Implements #8434.
Related to #3309.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ObserverProductionIconsWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ObserverProductionIconsWidget.cs
@@ -75,15 +75,16 @@ namespace OpenRA.Mods.Common.Widgets
 				var rsi = actor.Traits.Get<RenderSpritesInfo>();
 				var icon = new Animation(world, rsi.GetImage(actor, world.Map.SequenceProvider, race));
 				icon.Play(actor.Traits.Get<TooltipInfo>().Icon);
+				var bi = actor.Traits.Get<BuildableInfo>();
 				var location = new float2(RenderBounds.Location) + new float2(queue.i * (IconWidth + IconSpacing), 0);
-				WidgetUtils.DrawSHPCentered(icon.Image, location + 0.5f * iconSize, worldRenderer, 0.5f);
+				WidgetUtils.DrawSHPCentered(icon.Image, location + 0.5f * iconSize, worldRenderer.Palette(bi.IconPalette), 0.5f);
 
 				var clock = clocks[queue.Trait];
 				clock.PlayFetchIndex("idle",
 					() => current.TotalTime == 0 ? 0 : ((current.TotalTime - current.RemainingTime)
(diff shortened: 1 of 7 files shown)


penev92:
Looks good to me.
I like unhardcoding stuff :+1:

Phrohdoh:
s/sidebar/production palette/, please.

GraionDilach:
I don't see what you mean. In sequence it's called icon. (I do know I call it cameo in my commit but that's just my Westwood-based reflex). I think naming is to something else besides IconPalette wouldn't be consistent.

Phrohdoh:
Replace "sidebar" with "production palette", I didn't mention " icon" at all.

OpenRA.Mods.Common/Widgets/ObserverProductionIconsWidget.cs
```diff
@@ -75,15 +75,16 @@ public override void Draw()
 				var rsi = actor.Traits.Get<RenderSpritesInfo>();
 				var icon = new Animation(world, rsi.GetImage(actor, world.Map.SequenceProvider, race));
 				icon.Play(actor.Traits.Get<TooltipInfo>().Icon);
+				var bi = actor.Traits.GetOrDefault<BuildableInfo>();
```
Mailaender: You are using `GetOrDefault`, but forgot the `bi == null` check. Either add it or simply use `Get`.

OpenRA.Mods.Common/Widgets/ProductionPaletteWidget.cs
```diff
@@ -312,12 +313,15 @@ public void RefreshIcons()
 				var icon = new Animation(World, rsi.GetImage(item, World.Map.SequenceProvider, race));
 				icon.Play(item.Traits.Get<TooltipInfo>().Icon);
 
+				var bi = item.Traits.GetOrDefault<BuildableInfo>();
```
Mailaender: Essentially the same thing here.

OpenRA.Mods.Common/Widgets/ObserverProductionIconsWidget.cs
pchote: Use `Get`.  We are guaranteed that it won't fail because  non-`Buildable` actors can't show up here.

GraionDilach:
Resolved all issues aforementioned.

Mailaender:
:white_check_mark: Works as promised.

GraionDilach:
Removed testcase.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5f8a5a855fee148d9e3d635377c296baab5b891c...813394dcbe4be2a11bf07e5d6ce073a7bae2ff54)

Mailaender:
Well done.

PR #8465, opened 2015-06-12
Adds a feature to Utility that extracts chat content from replay files.

deniz1a:
This is still work in progress and not finished. I opened this PR to get some feedback.

### DIFF
--- a/OpenRA.Game/FileFormats/ReplayReport.cs
+++ b/OpenRA.Game/FileFormats/ReplayReport.cs
@@ -179,8 +179,11 @@ namespace OpenRA.FileFormats
 
 			text.AppendLine();
 			text.AppendLine("#### Game Ended ####");
-			text.AppendLine();
-			text.AppendLine("Results:");
+			if (playersByTeam.Any())
+			{
+				text.AppendLine();
+				text.AppendLine("Results:");
+			}
 
 			foreach (var team in playersByTeam)
 			{
--- a/OpenRA.Game/FileFormats/ReplayReport.cs
+++ b/OpenRA.Game/FileFormats/ReplayReport.cs
@@ -187,9 +187,7 @@ namespace OpenRA.FileFormats
 			}
 			catch (Exception ex)
 			{
-				// OutOfMemoryException can be thrown if the replay file is corrupt.
-				if (ex is ArgumentException || ex is NotSupportedException || ex is IOException
-					|| ex is UnauthorizedAccessException || ex is OutOfMemoryException)
+				if (ex is ArgumentException || ex is NotSupportedException || ex is IOException || ex is UnauthorizedAccessException)
 				{
 					Log.Write("debug", ex.ToString());
 					return null;
--- a/OpenRA.Game/FileFormats/ReplayReport.cs
+++ b/OpenRA.Game/FileFormats/ReplayReport.cs
@@ -205,6 +205,11 @@ namespace OpenRA.FileFormats
 			text.AppendLine();
 			text.AppendLine("Results:");
 
+			// Generate results table
+			var maxPlayerLength = gameInfo.Players.Max(p => p.Name.Length);
+			var maxOutcomeLength = gameInfo.Players.Max(p => p.Outcome.ToString().Length);
+			var paddingLength = 2;
+
 			foreach (var team in playersByTeam)
 			{
 				text.AppendLine();
@@ -212,7 +217,16 @@ namespace OpenRA.FileFormats
 				text.AppendLine(teamLabel);
 				text.AppendLine(new string('-', teamLabel.Length));
 				foreach (var player in team)
-					text.AppendLine(player.Name + " - " + player.Outcome);
+				{
+					var result = player.Outcome.ToString();
+					if (player.Outcome != WinState.Undefined)
+					{
+						var timeDiff = player.OutcomeTimestampUtc.Subtract(gameInfo.StartTimeUtc);
+						result += new string(' ', maxOutcomeLength - result.Length + paddingLength) + timeDiff;
+					}
+
+					text.AppendLine(player.Name + new string(' ', maxPlayerLength - player.Name.Length + paddingLength) + result);
+				}
 			}
 
 			text.AppendLine();
--- a/OpenRA.Game/FileFormats/ReplayReport.cs
+++ b/OpenRA.Game/FileFormats/ReplayReport.cs
@@ -19,9 +19,6 @@ namespace OpenRA.FileFormats
 {
 	public class ReplayReport
 	{
-		// Engine constant (ms)
-		static int networkTickDuration = 120;
-
 		public static string Read(ReplayMetadata metadata)
 		{
 			if (metadata == null || metadata.FilePath == null)
@@ -157,7 +154,7 @@ namespace OpenRA.FileFormats
 
 						// Time stamps are in game time.
 						if (gameStarted)
-							time = new TimeSpan(0, 0, networkTickCount * networkTickDuration / 1000) + " ";
+							time = new TimeSpan(0, 0, networkTickCount * Game.NetTickScale * Game.Timestep / 1000) + " ";
 
 						foreach (var o in orders)
 						{
--- a/OpenRA.Game/FileFormats/ReplayReport.cs
+++ b/OpenRA.Game/FileFormats/ReplayReport.cs
@@ -155,8 +155,6 @@ namespace OpenRA.FileFormats
 				using (var fs = File.OpenRead(metadata.FilePath))
 				{
 					var networkTickCount = 0;
-					if (!gameInfo.IsSinglePlayer)
-						GeoIP.Initialize();
 
 					while (fs.Position < metadata.MetaStartMarkerPosition)
 					{
--- a/OpenRA.Game/FileFormats/ReplayReport.cs
+++ b/OpenRA.Game/FileFormats/ReplayReport.cs
@@ -19,7 +19,8 @@ namespace OpenRA.FileFormats
 {
 	public class ReplayReport
 	{
-		static int networkTickDuration = 120; // Engine constant (ms)
+		// Engine constant (ms)
+		static int networkTickDuration = 120;
 
 		public static string Read(ReplayMetadata metadata)
 		{
--- a/OpenRA.Game/FileFormats/ReplayReport.cs
+++ b/OpenRA.Game/FileFormats/ReplayReport.cs
@@ -63,7 +63,6 @@ namespace OpenRA.FileFormats
 						text.AppendLine("Start Time UTC: " + gameInfo.StartTimeUtc.ToString("yyyy-MM-dd HH:mm:ss"));
 						text.AppendLine();
 
-						GeoIP.Initialize();
 						foreach (var team in playersByTeam)
 						{
 							var teamLabel = team.Key == 0 ? "No Team:" : "Team " + team.Key + ":";
@@ -74,17 +73,17 @@ namespace OpenRA.FileFormats
 								text.AppendLine(player.Name);
 								var client = session.ClientWithIndex(player.ClientIndex);
 								if (client.IsAdmin && !gameInfo.IsSinglePlayer)
-									text.AppendLine("\t[Admin]");
+									text.AppendLine("[Admin]");
 
-								text.AppendLine("\tHuman: " + player.IsHuman);
-								text.AppendLine("\tFaction: " + player.FactionName);
-								text.AppendLine("\tRandom faction: " + player.IsRandomFaction);
-								text.AppendLine("\tSpawn point: " + player.SpawnPoint);
-								text.AppendLine("\tRandom spawn point: " + player.IsRandomSpawnPoint);
+								text.AppendLine("Human: " + player.IsHuman);
+								text.AppendLine("Faction: " + player.FactionName);
+								text.AppendLine("Random faction: " + player.IsRandomFaction);
+								text.AppendLine("Spawn point: " + player.SpawnPoint);
+								text.AppendLine("Random spawn point: " + player.IsRandomSpawnPoint);
 
 								if (player.IsHuman && !gameInfo.IsSinglePlayer)
 									foreach (var line in NetworkInfo(session, client))
-										text.AppendLine("\t" + line);
+										text.AppendLine(line);
 
 								text.AppendLine();
 							}
@@ -102,10 +101,10 @@ namespace OpenRA.FileFormats
 								if (!gameInfo.IsSinglePlayer)
 								{
 									if (spec.IsAdmin)
-										text.AppendLine("\t[Admin]");
+										text.AppendLine("[Admin]");
 
 									foreach (var line in NetworkInfo(session, spec))
-										text.AppendLine("\t" + line);
+										text.AppendLine(line);
 								}
 
 								text.AppendLine();
@@ -156,6 +155,8 @@ namespace OpenRA.FileFormats
 				using (var fs = File.OpenRead(metadata.FilePath))
 				{
 					var networkTickCount = 0;
+					if (!gameInfo.IsSinglePlayer)
+						GeoIP.Initialize();
 
 					while (fs.Position < metadata.MetaStartMarkerPosition)
 					{
--- a/OpenRA.Game/FileFormats/ReplayReport.cs
+++ b/OpenRA.Game/FileFormats/ReplayReport.cs
@@ -105,7 +105,7 @@ namespace OpenRA.FileFormats
 							}
 						}
 
-						AddSettingsDisplay(session.GlobalSettings, text);
+						AddSettingsDisplay(session.GlobalSettings, gameInfo.Mod, text);
 						text.AppendLine();
 						gameStarted = true;
 					}
@@ -223,10 +223,15 @@ namespace OpenRA.FileFormats
 			text.AppendLine("\tLocation:    " + GeoIP.LookupCountry(client.IpAddress));
 		}
 
-		static void AddSettingsDisplay(Session.Global settings, StringBuilder text)
+		static void AddSettingsDisplay(Session.Global settings, string mod, StringBuilder text)
 		{
 			var setDisp = new List<string>();
 			var defaultSettings = new Session.Global();
+			if (ModMetadata.AllMods.ContainsKey(mod))
+			{
+				var modData = new ModData(mod);
+				FieldLoader.Load(defaultSettings, modData.Manifest.LobbyDefaults);
+			}
 
 			setDisp.Add("Default game settings");
 
--- a/OpenRA.Game/FileFormats/ReplayReport.cs
+++ b/OpenRA.Game/FileFormats/ReplayReport.cs
@@ -27,10 +27,10 @@ namespace OpenRA.FileFormats
 				return null;
 
 			var text = new StringBuilder();
+			var session = new Session();
 
 			var gameInfo = metadata.GameInfo;
 			var playersByTeam = gameInfo.Players.GroupBy(p => p.Team).OrderBy(g => g.Key);
-			var session = new Session();
 
 			var packetClientIndex = 0;
 			var gameStarted = false;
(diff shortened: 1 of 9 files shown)


deniz1a:
Here's an example output:
Replay file: https://github.com/deniz1a/file-uploads/raw/master/OpenRA-2015-05-05T214007Z.orarep
Output of the --replay-chat command: https://github.com/deniz1a/file-uploads/blob/master/chat-output.txt

It currently only outputs to command line.

deniz1a:
Now I have some questions. To get time stamps for ingame chat, I convert tick count to seconds by dividing it by 25. This should normally give the relative game time, which would progress slower than real time if there was lag. But the actual output has a discrepancy with real time that I think is not explainable by lag. In the example I posted above, the real game duration is 57 minutes but the calculated last time stamp is at 15 minutes... Why is that so? Are frames skipped in orders? Is there a way to get the real tick count of network orders?

deniz1a:
Does ReplayMetadata.GameInfo.StartTimeUtc give the starting time of the game or the recording, which begins in the lobby?

deniz1a:
I will also add saving the output to a file and printing teams at game start features.

deniz1a:
Ok start time is the game start and not recording start:
https://github.com/OpenRA/OpenRA/blob/ce1eed6bcc29eb4033ff82589a0d2c9e5afb7735/OpenRA.Game/World.cs#L218

deniz1a:
Ok I fixed the timestamps. Now the output looks like this: https://github.com/deniz1a/file-uploads/blob/master/fixed-chat-output.txt

I was using wrong type of ticks to calculate time. There is game tick, which is 40 ms or 25 per second. And then there is network tick which is 120 ms or once every 3 game ticks. The frame number from replay packets corresponds to network tick count. And network tick * 3 / 25 gives the game time of packets in seconds.

deniz1a:
Updated the code. Now it shows team configuration at game start and results at game end. Here's the output: https://github.com/deniz1a/file-uploads/blob/master/chat2.txt

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
```diff
@@ -0,0 +1,167 @@
...
+using System.Linq;
+using OpenRA.FileFormats;
+using OpenRA.Network;
+
+namespace OpenRA.Mods.Common.UtilityCommands
+{
+	class ReplayChatCommand : IUtilityCommand
+	{
+		public string Name { get { return "--replay-chat"; } }
+		readonly int gameTicksPerSecond = 25;		// Engine
```
Unit158: Do we already have this defined somewhere else? This seems silly.

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
```diff
@@ -0,0 +1,167 @@
...
+using System.IO;
+using System.Linq;
+using OpenRA.FileFormats;
+using OpenRA.Network;
+
+namespace OpenRA.Mods.Common.UtilityCommands
+{
+	class ReplayChatCommand : IUtilityCommand
+	{
+		public string Name { get { return "--replay-chat"; } }
```
Unit158: Can this be const or readonly?

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
```diff
@@ -0,0 +1,167 @@
...
+		readonly int gameTicksPerSecond = 25;		// Engine
+		readonly int gameTicksPerNetworkTick = 3;	// constants
+
+		[Desc("REPLAYFILE", "Extract chat content from a replay file.")]
+		public void Run(ModData modData, string[] args)
+		{
+			if (args.Length == 1)
+			{
+				Console.WriteLine("Path argument for a replay file is missing.");
```
Unit158: We should be displaying the use of this command here instead.

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
```diff
@@ -0,0 +1,167 @@
...
+			var path = args[1];
+			if (!File.Exists(path))
+			{
+				Console.WriteLine("Replay file could not be found.");
+				return;
+			}
+
+			var metadata = ReplayMetadata.Read(path);
+			if (metadata == null)
+				throw new InvalidDataException("Failed to read replay meta data");
```
Unit158: meta data -> metadata, and add a period at the end of the line please.

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
```diff
@@ -0,0 +1,167 @@
...
+			{
+				while (fs.Position < fs.Length - 4)
+				{
+					var client = fs.ReadInt32();
+					if (client == ReplayMetadata.MetaStartMarker)
+						break;
+
+					var packetLen = fs.ReadInt32();
+					var packet = fs.ReadBytes(packetLen);
+					if (packet.Length == 5 && packet[4] == 0xBF)
```
Unit158: Do we have enums for these?

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
penev92: It is implementing a property from `IUtilityCommand`, that only has a getter, so no to both.

abcdefg30:
(Not a real request) Would it be possible to also show messages like "xy is defeated"?

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
Phrohdoh: What about `!= 1` lengths?

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
Phrohdoh: No, we just use `25` as necessary.

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
deniz1a: I don't know, I copied them from here: https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Network/ReplayConnection.cs#L58

deniz1a:
It shows all the server messages. Is that sent as a server message? Player disconnected messages are shown for example. And the end result is also shown at the end of the output.

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
deniz1a: The Run method is called with at least one argument. I will add checks for too many arguments.

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
penev92: I don't believe so, which is unfortunate.

abcdefg30:
It's a debug message... [ConquestVictoryConditions Line 57](https://github.com/OpenRA/OpenRA/blob/12966129a4e412fd2c53eb05b310f30557058521/OpenRA.Mods.Common/Traits/Player/ConquestVictoryConditions.cs#L57)

deniz1a:
Here are the outputs of these replays: 
https://www.dropbox.com/s/m9f5mn4pcwn69oa/15%20games.rar?dl=0

outputs:
https://github.com/deniz1a/file-uploads/tree/master/15_games

deniz1a:
Maybe I should also add batch extraction parameter... So it could extract all .orarep files in a directory.

deniz1a:
Don't review this yet, it's not finished. I may rewrite this.

deniz1a:
I will definitely rewrite this, currently it's not using engine's Session.Deserialize method and it's not organized well.. I am thinking of creating a new class in OpenRA.FileFormats called ReplayChat, like the one for replay metadata. So that it can be reused in other places like the replay browser. Also how's this idea: replays should be zip archives like .oramap files. It would contain three files: the actual replay, a text file containing replay metadata and another text file containing chat output... The text files would be automatically generated when the replay is saved...

deniz1a:
Also, there are at least two places in engine code where network packets are read. Wouldn't it be better to have a common Packet class or something so it would be unified and reused? And if in the future there were any changes to network code, it would be enough to change it in one place and everything that uses it would be automatically updated...

deniz1a:
This has become more of a replay report than just the chat... It contains the complete information about the game.

deniz1a:
Updated. It's still not finished but has all the features. --replay-report command has three operation modes:
* Extract from single replay and print the report to the console.
* Extract from single replay and save the report file to the given path.
* Batch extract from a source directory and save report files in given path. If no output path is given, a default one is created.

Operation mode is determined by the type of arguments. If source path is a directory, batch mode is executed.

The progress display made the code a little more complicated but I think it's useful. Extracting a large number of replays can take a while and it is better to give a visual feedback so the program doesn't look like it's frozen.

Again, it's not finished and so not ready for full review. But quick feedbacks would be appreciated.

deniz1a:
OK I think this PR is finished now. It needs review.

OpenRA.Mods.Common/UtilityCommands/ReplayChatCommand.cs
WolfGaming: Yeah, our replay code is full of magic values :/

deniz1a:
@abcdefg30 : I think debug messages are not sent to other players, so they're not in replays. I don't know if that could be extracted without running the game.

deniz1a:
Here is a compilation of replays I found on the forum, along with their extracted reports. They are organized in folders with links to the forum pages.

https://drive.google.com/file/d/0B_dMvsHPqPZkYW1tUGpxQlAyTFE/view?usp=sharing

deniz1a:
Metadata contains timestamp for the outcome. But it's in real time. It would be possible to insert that into the chat but the timestamp would not be accurate because game time flows slower than real time...

deniz1a:
I could add some checks to ReplayReport.cs to make it more robust against corrupted replay files. Currently it can throw uncaught exception if some fields are missing... But it didn't happen for the 900+ replays i tried this with... So I don't know if that would be necessary...

OpenRA.Game/FileFormats/ReplayReport.cs
```diff
@@ -0,0 +1,290 @@
...
+
+							orderHandler[o.OrderString](o.TargetString);
+						}
+					}
+				}
+			}
+			catch (Exception ex)
+			{
+				// OutOfMemoryException can be thrown if the replay file is corrupt.
+				if (ex is ArgumentException || ex is NotSupportedException || ex is IOException
```
Unit158: Only catch what you handle.
Catching Exception is foolish.

deniz1a:
Here's an example report if you don't want to download the 300 MB file:
https://gist.github.com/deniz1a/a4e852d064fa30e49c74

OpenRA.Game/FileFormats/ReplayReport.cs
deniz1a: I throw it again a couple of lines below.

Unit158: That's even worse.

deniz1a: Catching Exception and then rethrowing it is also done elsewhere in the game. Here for example:
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Utility/Program.cs#L56

deniz1a: Other examples:
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Exts.cs#L448
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Renderer.Sdl2/Sdl2GraphicsDevice.cs#L153
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/GameInformation.cs#L73

deniz1a: Also, how else could I catch multiple exceptions and do the same thing for all of them?

deniz1a: Only, I'm not sure if OutOfMemoryException should also be caught here... That happens when replay file is corrupt but could also happen for other reasons, right?

deniz1a:
Made some improvements. Now the results are better formatted and show outcome timestamps. These are in real time and not game time as opposed to chat timestamps. Also, batch extractions should be faster now since GeoIP.Initialize() is called only once in ReplayReportCommand and not for every file in ReplayReport.

deniz1a:
Because of the formatting of the results section, reports are best viewed with a monospace font.

Mailaender:
Pinging @ihptru who might be interested in this for http://resource.openra.net

deniz1a:
Does it look better with text alignment or not?

OpenRA.Game/FileFormats/ReplayReport.cs
```diff
@@ -0,0 +1,269 @@
...
+using System.IO;
+using System.Linq;
+using System.Text;
+using OpenRA.Network;
+
+namespace OpenRA.FileFormats
+{
+	public class ReplayReport
+	{
+		static int networkTickDuration = 120; // Engine constant (ms)
```
abcdefg30: Comments on a new line, please.

deniz1a:
Any opinions on whether texts should be aligned? I think it should be aligned for results but not the other stuff. :waits for opinions emoticon:

abcdefg30:
Looks good to me :+1: / :white_check_mark:
after you "fixed" the aligment and the results being empty when "playing" with spectators only.

deniz1a:
I fixed empty Results: section. And about the alignment, it looks ok to me... If there are noobjections, I'll leave it as it is.

OpenRA.Game/Network/ReplayRecorderConnection.cs
```diff
@@ -50,8 +50,7 @@ void StartSavingReplay(byte[] initialContent)
 			var mod = Game.ModData.Manifest.Mod;
 			var dir = Platform.ResolvePath("^", "Replays", mod.Id, mod.Version);
 
-			if (!Directory.Exists(dir))
-				Directory.CreateDirectory(dir);
+			Directory.CreateDirectory(dir);
```
GraionDilach: Why this change?

OpenRA.Game/Network/Session.cs
```diff
@@ -185,7 +185,7 @@ public class Global
 			public bool Fog = true;
 			public bool AllyBuildRadius = true;
 			public int StartingCash = 5000;
-			public string TechLevel = "none";
+			public string TechLevel = "Unrestricted";
```
GraionDilach: Why is this?

OpenRA.Utility/Program.cs
```diff
@@ -28,7 +28,6 @@ static void Main(string[] args)
 			AppDomain.CurrentDomain.AssemblyResolve += GlobalFileSystem.ResolveAssembly;
 
 			Log.AddChannel("perf", null);
-			Log.AddChannel("debug", null);
```
GraionDilach: Why do you remove this?

OpenRA.Game/Network/ReplayRecorderConnection.cs
deniz1a: Because the check for the existance of directory is not necessary. If it exists, Directory.CreateDirectory() does nothing.

OpenRA.Game/Network/Session.cs
deniz1a: Apparently it was incorrect. The default value is "Unrestricted".

OpenRA.Utility/Program.cs
deniz1a: I think because --replay-metadata and --replay-report are the only utility functions that use that? So I moved it to those classes.

OpenRA.Game/Network/Session.cs
pchote: This doesn't have anything to do with extracting chat content from replay files, so please move it to another pr.

OpenRA.Utility/Program.cs
deniz1a: No, sorry. The reason I removed that is once a channel is added, the subsequent calls to add the same channel are ignored. So once it is added as null, it is not possible to change that channel to debug.log.

OpenRA.Game/Network/ReplayRecorderConnection.cs
GraionDilach: I assumed it was there due to some Mono-.NET difference but cannot find any information regarding such, so okay.

OpenRA.Game/Network/Session.cs
deniz1a: But if I don't change this, games with unrestricted tech setting show up as non-default. I didn't change the default setting, it is the same.

pchote: That means your reporting is wrong.  You need to be comparing against the mod-defined defaults (specified in mod.yaml), which override these global defaults.  Replacing these global placeholders with mod-specific values is the wrong approach.

deniz1a: Oh ok, I'll change that.

pchote:
This looks to have scope-creeped massively since the original proposal.  I'd like to see an explanation for why this is sufficiently useful to merge.  Because it does all its own order parsing, I can easily see this bitrotting and becoming a maintenance burden for us in the future.

deniz1a:
Isn't order parsing required to extract chat? How could it be done otherwise? And about expanding chat output to complete reports, I thought this would give context to the conversations.

OpenRA.Utility/Program.cs
penev92: The utility is not supposed to use that channel anyway.
Looks to me like you are removing this from here so you can add in in the replay parsing because you copied that part from somewhere else, which is another issue that @pchote's comment below is talking about.

deniz1a: But ReplayMetadata and ReplayReport write error messages to debug log. Shouldn't the utility use that?

deniz1a:
I fixed the default settings. Now it uses mod yaml for comparison. And while doing that, I realized that this utility command is mod independent. So there is no need to give a mod argument at the command line because it reads all information from the replay file. It could process a directory full of replays from different mods.

If this PR gets merged, a separate one could change the utility program so that commands that are common to all mods can be run without mod argument.

deniz1a:
--replay-metadata command is also mod independent, right?

OpenRA.Game/Network/ReplayRecorderConnection.cs
deniz1a: I don't know about that... The docs for .Net Framework and Mono both say that it does nothing if directory already exists:
https://msdn.microsoft.com/en-us/library/54a0at6s%28v=vs.110%29.aspx
http://docs.go-mono.com/?link=M%3aSystem.IO.Directory.CreateDirectory%28System.String%29

pchote:
This appears to have been superseded by http://www.gamereplays.org/openra/

deniz1a:
How does that replace this?

pchote:
It provides a user-friendly replay report, which is what this pr appears to have evolved into.

deniz1a:
I can't find any replay reports on that site? It only shows player names... Am I missing something?

pchote:
It shows the player name, map name/preview, game duration, and they are working on fixing the winner/loser display.

deniz1a:
Why don't they use this utility feature?

deniz1a:
Also, the reports of this utility command show much more... Chat for example.

penev92:
After a lengthy discussion on IRC two days ago, it became apparent people aren't interested.
([IRC Logs](http://logs.openra.net/index.php?year=2015&month=07&day=17), somewhere around 22:00-22:30).
Looking for any dev to express any interest, because otherwise I'll close this as unwanted/unneeded/redundant to what [GameReplays](http://www.gamereplays.org/portals.php) offer.

deniz1a:
So, extracting chat, displaying game settings etc aren't wanted?

obrakmann:
:-1:. I'm all for closing. Chat extraction is a bad idea, and --replay-metadata already exists and offers enough information.

deniz1a:
Ok, then maybe I will add lobby settings to metadata.

PR #8154, opened 2015-05-13
Added lint rules to detect bad spawn points in maps.

Mailaender:
This will detect faulty maps with invalid spawn points (they can't share the same location for obvious reasons) such as those described in #8115 and show a FIXME warning as well as disable download at resource.openra.net. Another rule will detect maps where the Multi\* player definitions don't match the spawn point count. This happens when you forgot to click "Setup default players" in the OpenRA.Editor.exe which is a common pitfall. Only the not enough spawn points case is really problematic though.

PR #8473, opened 2015-06-13
Fixed an ArgumentException in ClosestProc

### LINKED ISSUE #8467: Crash in the harvester logic
Happened while testing soviet-05 (see also #8446).

```
Red Alert Mod at Version {DEV_VERSION}
Exception of type `System.ArgumentException`: An item with the same key has already been added. (Ein Element mit dem gleichen Schlüssel wurde bereits hinzugefügt.)
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey](IEnumerable`1 source, Func`2 keySelector)
   at OpenRA.Mods.Common.Traits.Harvester.ClosestProc(Actor self, Actor ignore)
   at OpenRA.Mods.Common.Traits.Harvester.ChooseNewProc(Actor self, Actor ignore)
   at OpenRA.Mods.Common.Traits.Harvester.<>c__DisplayClass4.<.ctor>b__2()
   at OpenRA.Activities.CallFunc.Tick(Actor self) in \OpenRA\OpenRA.Game\Activities\CallFunc.cs:Line 29.
   at OpenRA.Traits.Util.RunActivity(Actor self, Activity act) in \OpenRA\OpenRA.Game\Traits\Util.cs:Line 83.
   at OpenRA.Actor.Tick() in \OpenRA\OpenRA.Game\Actor.cs:Line 121.
   at OpenRA.World.Tick() in \OpenRA\OpenRA.Game\World.cs:Line 302.
   at OpenRA.Game.InnerLogicTick(OrderManager orderManager) in \OpenRA\OpenRA.Game\Game.cs:Line 456.
   at OpenRA.Game.LogicTick() in \OpenRA\OpenRA.Game\Game.cs:Line 480.
   at OpenRA.Game.Loop() in \OpenRA\OpenRA.Game\Game.cs:Line 596.
   at OpenRA.Game.Run() in \OpenRA\OpenRA.Game\Game.cs:Line 636.
   at OpenRA.Program.Run(String[] args) in \OpenRA\OpenRA.Game\Support\Program.cs:Line 111.
   at OpenRA.Program.Main(String[] args) in \OpenRA\OpenRA.Game\Support\Program.cs:Line 39.

```
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/21749489-crash-in-the-harvester-logic?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mailaender:
This intends to fix #8467.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Harvester.cs
+++ b/OpenRA.Mods.Common/Traits/Harvester.cs
@@ -117,7 +117,8 @@ namespace OpenRA.Mods.Common.Traits
 				from r in self.World.ActorsWithTrait<IAcceptResources>()
 				where r.Actor != ignore && r.Actor.Owner == self.Owner && IsAcceptableProcType(r.Actor)
 				let linkedHarvs = self.World.ActorsWithTrait<Harvester>().Count(a => a.Trait.LinkedProc == r.Actor)
-				select new { Location = r.Actor.Location + r.Trait.DeliveryOffset, Actor = r.Actor, Occupancy = linkedHarvs }).ToDictionary(r => r.Location);
+				select new { Location = r.Actor.Location + r.Trait.DeliveryOffset, Actor = r.Actor, Occupancy = linkedHarvs })
+				.Distinct().ToDictionary(r => r.Location);
 
 			// Start a search from each refinery's delivery location:
 			var mi = self.Info.Traits.Get<MobileInfo>();


penev92:
While this will fix the crash, I think it only hides the problem.
I imagine a reproduce-case would be to place 2 refineries at the exact same spot (last I checked the game won't complain) either via YAML or the map editor. Your fix would just hide the (actually bogus) fact that there are 2 refineries at those coordinates, not 1.
I don't consider #8467 a crash that needs fixing unless we can reproduce and debug. It is on a work-in-progress map/mission, so that doesn't help its case much either.

pchote:
Do we want a lint check that looks for invalid footprint overlaps? (note: there may be cases where maps do want this, such as for trees?)

penev92:
We definitely do. Don't we already have an issue open about that?

Mailaender:
We only check invalid overlapping spawn points, because there is no way to recover from that. https://github.com/OpenRA/OpenRA/pull/8154

obrakmann:
I agree that #8467 does not really need fixing. I think we shouldn't make the code fail-safe for situations we consider to be bogus in the first place. But :+1: on a lint check for actor collisions, which will fix the problem at its core.

Mailaender:
> Do we want a lint check that looks for invalid footprint overlaps? (note: there may be cases where maps do want this, such as for trees?)

Even just checking the location and not the whole footprints seems to trigger a massive amount of actors: https://github.com/OpenRA/OpenRA/pull/11240 so I am not really sure if the decision not to take the fix, but have a spammy lint check instead is a good idea.

PR #8474, opened 2015-06-13
Remove region assumptions from fast shroud tests

pchote:
The next step towards heightmap-aware shroud and scripted map bound changes.  This removes the region tests by making _all_ cases use the fast path.

This is possible because of two changes:
1. Coordinates are pre-filtered to ensure they are inside the map area (note: this includes the cordon outside the current map bounds).
2. When updating the current visibility any cells outside the current bounds are forced to be hidden.

### DIFF
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -28,7 +28,7 @@ namespace OpenRA.Traits
 		public readonly WPos CenterPosition;
 		public readonly Rectangle Bounds;
 		readonly Actor actor;
-		readonly Func<MPos, bool> isVisibleTest;
+		readonly Shroud shroud;
 
 		public Player Owner;
 
@@ -45,7 +45,7 @@ namespace OpenRA.Traits
 		public FrozenActor(Actor self, MPos[] footprint, Shroud shroud)
 		{
 			actor = self;
-			isVisibleTest = shroud.IsVisibleTest;
+			this.shroud = shroud;
 
 			// Consider all cells inside the map area (ignoring the current map bounds)
 			Footprint = footprint
@@ -79,6 +79,7 @@ namespace OpenRA.Traits
 		void UpdateVisibility()
 		{
 			var wasVisible = Visible;
+			var isVisibleTest = shroud.IsVisibleTest;
 
 			// We are doing the following LINQ manually for performance since this is a hot path.
 			// Visible = !Footprint.Any(isVisibleTest);
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -42,12 +42,16 @@ namespace OpenRA.Traits
 		public bool NeedRenderables;
 		public bool IsRendering { get; private set; }
 
-		public FrozenActor(Actor self, MPos[] footprint, CellRegion footprintRegion, Shroud shroud)
+		public FrozenActor(Actor self, MPos[] footprint, Shroud shroud)
 		{
 			actor = self;
-			isVisibleTest = shroud.IsVisibleTest(footprintRegion);
+			isVisibleTest = shroud.IsVisibleTest;
 
-			Footprint = footprint;
+			// Consider all cells inside the map area (ignoring the current map bounds)
+			Footprint = footprint
+				.Where(m => shroud.Contains(m))
+				.ToArray();
+
 			CenterPosition = self.CenterPosition;
 			Bounds = self.Bounds;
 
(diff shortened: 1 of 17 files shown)


OpenRA.Game/Traits/World/Shroud.cs
```diff
@@ -406,5 +404,10 @@ public bool HasFogVisibility()
 		{
 			return fogVisibilities.Value.Any(f => f.HasFogVisibility(self.Owner));
 		}
+
+		public bool Contains(MPos uv)
+		{
+			return explored.Contains(uv);
```
penev92: This isn't too complicated, but still got me confused. Please add a comment why you're checking `explored`.

pchote:
Updated and rebased.

obrakmann:
Code looks good at first glance, but I think you might have missed a place or two that needs updating:

ra-shroud
ts-shroud

OpenRA.Game/Traits/World/Shroud.cs
```diff
@@ -129,6 +125,11 @@ void AddVisibility(Actor a, ref CPos[] visible)
 			foreach (var c in visible)
 			{
 				var uv = c.ToMPos(map);
+
+				// Force cells outside the visible bounds invisible
+				if (!map.Contains(uv))
+					continue;
```
RoosterDragon: This is going to break the corresponding ```RemoveVisibility``` call since it will remove vis on cells that you're filtering out here.

Maybe expose the ```allowOutsideBounds``` parameter in the ```FindVisibleTiles``` method?

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -42,12 +42,16 @@ public class FrozenActor
 		public bool NeedRenderables;
 		public bool IsRendering { get; private set; }
 
-		public FrozenActor(Actor self, MPos[] footprint, CellRegion footprintRegion, Shroud shroud)
+		public FrozenActor(Actor self, MPos[] footprint, Shroud shroud)
 		{
 			actor = self;
-			isVisibleTest = shroud.IsVisibleTest(footprintRegion);
+			isVisibleTest = shroud.IsVisibleTest;
```
RoosterDragon: This is my own fault, but I've just realised this was an invalid optimization - you can't hold onto this delegate for so long. If somebody uses the dev setting to toggle shroud the results become incorrect since you're now applying the wrong delegate.

It needs to become a local variable in ```UpdateVisibility``` again.

RoosterDragon:
Perhaps we should document that ```IsExploredTest``` returns undefined results for cells outside the map and isn't valid to be cached across ticks, whilst ```IsExplored``` will do the right thing at the expense of performance (same for IsVisible, etc).

OpenRA.Game/Traits/World/Shroud.cs
pchote: I'll add a similar check to the remove case.  We need to store all of the cells in the visibility cache so that we can generate the correct visibility if the map bounds are changed.

pchote:
Updated and rebased.

obrakmann:
Looks ok to me now, didn't find any more regressions in-game either. :+1:

OpenRA.Game/Traits/World/Shroud.cs
```diff
@@ -112,7 +108,7 @@ static IEnumerable<CPos> FindVisibleTiles(World world, CPos position, WRange rad
 			var limit = radius.RangeSquared;
 			var pos = map.CenterOfCell(position);
 
-			foreach (var cell in map.FindTilesInCircle(position, r))
+			foreach (var cell in map.FindTilesInCircle(position, r, true))
```
penev92: I don't understand while this needs to include tiles outside the current map bounds.

OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
```diff
@@ -146,7 +146,7 @@ public void Render(WorldRenderer wr)
 
 			Dirty.Clear();
 
-			foreach (var uv in wr.Viewport.VisibleCells.MapCoords)
+			foreach (var uv in wr.Viewport.VisibleCellsInsideBounds.MapCoords)
```
penev92: This (`EditorResourceLayer`) is only used inside the editor, so why not make things clearer and use `AllVisibleCells`?

penev92:
I don't know what's happening here, but... (not my harvester):


Otherwise :+1: with the above comment.

pchote:
> I don't know what's happening here, but... (not my harvester):

The harvester is probably about to move in to the visible area, which means it can be revealed by either of its current or next cells.

OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
pchote: The two are the same in editor mode.

pchote: Actually, this is a good idea.  I updated this and the terrain renderer, and then removed the magic Viewport behavior.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/dc0d32ea72ac256ce3ec62f2a5dda8ee6d07d34c...67f36b4c558de9ca9b58b27360527b77e11a1e9d)

PR #8477, opened 2015-06-13
Move actor visibility checks out of Shroud

pchote:
This moves the responsibility of checking actor visibility from Shroud on to the actor itself.  This is another small step towards heightmapped shroud support.

The new-found configurability is used to hide your own delivery aircraft under the fog in TD and D2K: these actors do not reveal any shroud, and having them visible is inconsistent with the idea that the fogged regions of the map are "blind" to you -- you should not have any idea what is happening under it.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckDefaultVisibility.cs
+++ b/OpenRA.Mods.Common/Lint/CheckDefaultVisibility.cs
@@ -1,0 +1,36 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+using OpenRA.GameRules;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Lint
+{
+	class CheckDefaultVisibility : ILintPass
+	{
+		public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
+		{
+			foreach (var actorInfo in map.Rules.Actors)
+			{
+				if (actorInfo.Key.StartsWith("^"))
+					continue;
+
+				var count = actorInfo.Value.Traits.WithInterface<IDefaultVisibilityInfo>().Count();
+
+				if (count == 0)
+					emitError("Actor type `{0}` does not define a default visibility type!".F(actorInfo.Key));
+				else if (count > 1)
+					emitError("Actor type `{0}` defines multiple default visibility types!".F(actorInfo.Key));
+			}
+		}
+	}
+}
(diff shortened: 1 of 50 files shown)


OpenRA.Mods.Common/Lint/CheckDefaultVisibility.cs
```diff
@@ -0,0 +1,36 @@
...
+		{
+			foreach (var actorInfo in map.Rules.Actors)
+			{
+				if (actorInfo.Key.StartsWith("^"))
+					continue;
+
+				var count = actorInfo.Value.Traits.WithInterface<IDefaultVisibilityInfo>().Count();
+
+				if (count == 0)
+					emitError("Actor type `{0}` does not define a default visibility type!".F(actorInfo.Key));
```
Phrohdoh: Newline, please.

pchote: Made it an `else if` instead.

pchote:
Updated.  This now also moves the GPS-related mess out of Shroud (but doesn't attempt to fix it - that is too big of a job).

mods/cnc/rules/defaults.yaml
```diff
@@ -651,6 +652,7 @@
 	ScriptTriggers:
 
 ^Crate:
+	HiddenUnderShroud:
```
penev92: Now that we finally have the chance to fix this, let's make it HiddenUnderFog?
Same for other mods of course.

penev92:
Well diff looks OK. :+0.5:
Would be nice to finally properly hide crates though.

pchote:
Rebased.

pchote:
Rebased again.

OpenRA.Mods.Common/Traits/Modifiers/AlwaysVisible.cs
```diff
@@ -0,0 +1,23 @@
...
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public class AlwaysVisibleInfo : TraitInfo<AlwaysVisible>, IDefaultVisibilityInfo { }
```
obrakmann: Please add a `[Desc()]`

obrakmann:
That one nit aside, this looks good to me. :+1:

pchote:
Done.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/731415923bd447ba22518e91b6528584b80de975%5E...731415923bd447ba22518e91b6528584b80de975)

abcdefg30:
This fails travis in #8484:
[`Actor type `e5` does not define a default visibility type!`](https://travis-ci.org/OpenRA/OpenRA/builds/67603289#L1273)
Even though it only modifies `e5` this way (that shouldn't break anything):
```yaml
E5:
	Buildable:
		Prerequisites: ~disabled
```
?

abcdefg30:
My bad. There is no `e5` in ra...

Issue #7466, opened 2015-02-11
Friendly and owned spy insignia (from Mental Omega).


Phrohdoh:
The popular RA2 mod Mental Omega gives spys a special insignia that denote that the actor is:
a) A spy
b) An allied unit (or owned by you)

This insignia is not visible to enemies and cannot be 'discovered' even if the disguise is discovered.

![Screenshot](http://mentalomega.com/images/news_img/item_insignia.jpg)

MustaphaTR:
:+1: It will be good, if we'll use this on OpenRA.

Unit158:
Have you contacted the author to make sure we can indeed use his resource? (assuming the license doesn't already say so.)

cjshmyr:
I'd only ask if you had it already coded, if this is going to happen.

Phrohdoh:
I don't think we should use the art from MO.

Phrohdoh:
I'm going to try to implement this.

penev92:
Opening again because support for this has been added, but a decoration is still missing.

x-a-n-a-x:
As requested on IRC (http://logs.openra.net/?year=2015&month=05&day=23#16:06:09), here are some possible symbols:

![7466_2](https://cloud.githubusercontent.com/assets/1732773/7785126/12629f90-0182-11e5-82a2-229b411790bd.png)

And here are 4 and 9 (rgb non-indexed PNG files):
![star7466](https://cloud.githubusercontent.com/assets/1732773/7785132/238ada94-0182-11e5-8ca1-9f7dc98bb62e.png)

![eye7466](https://cloud.githubusercontent.com/assets/1732773/7785130/1fb5d310-0182-11e5-92e0-e523939ac333.png)

x-a-n-a-x:
Now that #8478 is merged for spies, is there a plan to add a symbol (a star: see my previous message) for allied/friendly units?

GraionDilach:
That star was used for hero identification in MO (practically Tanya/Commandoes had it). I don't know if we're interested in such.

penev92:
I like the idea for hero units. Added it to the checklist at #7551.

GraionDilach:
I'll do that tomorrow then.

PR #8478, opened 2015-06-13
Implements a disguise decoration to Spies. 

GraionDilach:
Deprecates WithDecorationDisguised in favor of a refactor in Disguise.
spypip
Opened eye when disguised (animated, thanks @x-a-n-a-x ), closed eye when undisguised.

Implements #7466.


Phrohdoh:
There's no reason to use `pipdis` as the filename, we aren't limited in character count.

OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -108,6 +124,9 @@ public Color RadarColorOverride(Actor self)
 
 		void DisguiseAs(Actor self, Actor target)
 		{
+			var um = self.TraitOrDefault<UpgradeManager>();
```
Phrohdoh: Use `Requires<>` and assign this in the ctor.

OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -128,16 +147,33 @@ void DisguiseAs(Actor self, Actor target)
...
 				AsTooltipInfo = null;
 				AsPlayer = null;
 				AsSprite = null;
+
 			}
 
 			foreach (var t in self.TraitsImplementing<INotifyEffectiveOwnerChanged>())
 				t.OnEffectiveOwnerChanged(self, oldEffectiveOwner, AsPlayer);
+
+			if ((Disguised != oldDisguiseSetting) && (um != null))
```
Phrohdoh: Unnecessary parenthesis pairs.

GraionDilach:
Will fix these - and Travis - on Monday, sorry.

OpenRA.Mods.RA/Traits/Disguise.cs
ScottNZ: `Lazy<T>` would be more suitable if UpgradeManager is actually optional

epicelite:
How about no eye at all when undisguised?

Phrohdoh:
That was the original plan.

chrisforbes:
That seems like a better plan. The open/closed eyes are hard to distinguish.

GraionDilach:
Concerns addressed - renamed pipdis to pip-disguise, removed closed eye, fixed StyleCop and code issues.

mods/ra/rules/infantry.yaml
```diff
@@ -214,11 +214,21 @@ SPY:
...
+		Offset: 4, -2
+		ZOffset: 256
+		UpgradeTypes: disguise
+		UpgradeMinEnabledLevel: 1
 	Armament:
 		Weapon: SilencedPPK
 	AttackFrontal:
 	Voiced:
 		VoiceSet: SpyVoice
+
```
Mailaender: Spurious newline.

Mailaender:
Don't forget to add yourself to https://github.com/OpenRA/OpenRA/blob/bleed/AUTHORS

mods/ra/rules/infantry.yaml
GraionDilach: Ah, yeah, thanks.

GraionDilach:
I believe in the rule of three - and this is only the second one.

OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -138,8 +158,22 @@ void DisguiseAs(Actor self, Actor target)
...
 				t.OnEffectiveOwnerChanged(self, oldEffectiveOwner, AsPlayer);
+
+			if (Disguised != oldDisguiseSetting && um.Value != null)
+			{
+				foreach (var u in info.Upgrades)
+				{
+					if (!um.Value.AcceptsUpgrade(self, u))
+						continue;
+
+					if (Disguised)
```
Mailaender: Currently the revoking does not work. 
```csharp
if (Disguised && AsPlayer != self.Owner)
```

might fix it.

GraionDilach: How did you tested this out? I honestly couldn't even figure out an ingame way to drop the disguise...

Mailaender: Right-click on the spy again or shoot an enemy (or force fire on the ground).

GraionDilach: Thanks.

reaperrr:
Needs rebase.

OpenRA.Mods.RA/Traits/Disguise.cs
GraionDilach: Turned out that's not the issue. Due to UpgradeMaxAcceptsLevel defaults to 1, the AcceptsUpgrade check notices the Spy cannot taker another Disguise upgrade and returns false even before this line.

GraionDilach:
Rebased and fixed the issue about the eye never disappearing, see my last comment upon the outdated diff/second commit's message.

GraionDilach:
BTW, does this need an upgrade rule?

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -125,6 +125,14 @@ public void RevokeUpgrade(Actor self, string upgrade, object source)
 			NotifyUpgradeLevelChanged(s.Traits, self, upgrade, -1);
 		}
 
+		//Returns true if the actor can work with the upgrade. Does not check the actual level of the upgrade.
```
Phrohdoh: What does "can work with the upgrade" mean?
Also these comments should be `<summary>`.

OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -61,18 +62,14 @@ class DisguiseInfo : ITraitInfo
 	{
 		[VoiceReference] public readonly string Voice = "Action";
 
+		[Desc("Grants upgrades when disguised.")]
```
Phrohdoh: Suggested English: `Upgrades to grant when disguised.`

penev92:
The trait has been released for a week now, and was not even used in the default mods. I'd say no need.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
GraionDilach: That the upgrade is defined on the actor.

I didn't see many such descriptions in summary, sorry about that.

penev92: "uses the given upgrade" ?

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -125,6 +125,14 @@ public void RevokeUpgrade(Actor self, string upgrade, object source)
...
+		public bool AcknowledgesUpgrade(Actor self, string upgrade)
+		{
+			UpgradeState s;
+			return upgrades.Value.TryGetValue(upgrade, out s);
```
penev92: Looks like this can be simplified to `return upgrades.Value.ContainsKey(upgrade);`

GraionDilach:
Tomorrow I'll fix the issues. >_>

GraionDilach:
Resolved all issues aforementioned.

obrakmann:
Code looks good and works as advertised. Not sure if I like the art, though. It's very hard to make out unless your're zoomed in in my opinion, and the eye moves too fast for my liking. But I'll go with the majority's opinion here. :+1:

chrisforbes:
:+1:

GraionDilach:
I can give another try at art, I don't mind that.

GraionDilach:
Remade art, now it's a bit bigger, has an outer edge and slower (both because of smoother movement and Tick increased from 250 to 300).

I really hope it's okay now.

newopenraspy

OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -61,18 +62,14 @@ class DisguiseInfo : ITraitInfo
 	{
 		[VoiceReference] public readonly string Voice = "Action";
 
+		[Desc("Upgrades to grant when disguised.")]
+		public readonly string[] Upgrades = { "disguise" };
```
Mailaender: This needs a `[UpgradeGrantedReference]` now. See https://github.com/OpenRA/OpenRA/pull/8321/

Mailaender:
> Remade art, now it's a bit bigger, has an outer edge and slower (both because of smoother movement and Tick increased from 250 to 300).

Looks much better now! :eyes:

GraionDilach:
Rebased, added reference and the same asset to the TS Chameleon Spy as being an additional Disguise trait user.

Mailaender:
I guess we can live without an upgrade rule for the `WithDecorationDisguised` as it was never used.

Mailaender:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/78002d14d8191d3f42e701733a5bb0d0a9c51467...a084fa5b429f447257ee5e77df3a6f6d44e757be)

Mailaender:
Well done!

Phrohdoh:
@GraionDilach Nice job on this!

Issue #7916, opened 2015-04-10
Delay growing back the shroud when an actor gets killed


obrakmann:
So things like demo truck explosions and sandworm attacks will be visible to the player

obrakmann:
```
15:50:41) antares79: pchote: how would you do that? spawn a camera where the actor was, or add some magic to the shroud/frozen actor layer?
(15:50:53) pchote: magic to shroud
(15:51:15) pchote: we could then use the same magic for revealing firing locations and getting rid of the camera actors
(15:51:42) antares79: for super powers, you mean?
(15:51:49) pchote: and scripts
(15:51:59) pchote: creating a specific actor for each reveal range is dumb
(15:52:16) pchote: it would be much nicer if maps and support powers could just say "reveal shroud at X with radius R"
(15:53:56) pchote: easiest way is probably to add an abstraction
(15:54:06) pchote: make it so shroud doesn't deal with actors directly, but some kind of shroud proxy
(15:54:16) pchote: RevealsShroud then deals with manipulating the proxy for an actor
(15:54:24) pchote: and scripts and superweapons can make their own proxies
(15:54:42) pchote: we already kind of do that already
(15:54:51) pchote: except instead of an explicit proxy we just pass around a reference to a list of cells
```

penev92:
This has the same idea that  #7621 tried to make, so I guess close that one as reverse duplicate?

Phrohdoh:
I'll need this for a mod I have in mind.

Issue #5061, opened 2014-04-06
Static defenses should reveal their location after firing


psydev:
Currently if a turret opens fire on you, but it is under fog, you cannot see its location. I think that it would be better if turrets would become visible after they fire upon you, like in the original. 
##

Unit158:
This could probably use some more discussion, or close.

penev92:
This can definitely use some implementing. I am also told that with the shroud refactoring it is now quite doable (I even see the PR that mentions is just above your comment).

GraionDilach:
This existed as a weapon flag in TS/RA2 - which is why I assume it'll be needed in the long run anyway.

StexoO:
Will not it be too easy to detect and destroy all enemy defenses ? Or simply avoid them ?
Just send a single guy to detect one of them or a curtained APC to detect all of them, then build 3 artilleries / V2s to get rid of all of them ?
I know you all do it in all the RTS where defenses are revealed when they attack you. :smirk:
I don't know if :+1: because the players will build more vehicles / air units and less defenses, so the game will be less static, or if :-1: because we will see less defense, as they will be almost always destroyed by long range vehicles or air units.

penev92:
All the more reason to reveal everything that fires on you, not just buildings.

abcdefg30:
Support for this was added in #12557. `RevealOnFire` is enabled for TS and D2k (#12808), but I don't see this happening for RA and/or TD. (At least not without much discussion, which should happen on the forums.)

psydev:
I can't remember, but I think one of the reasons I suggested this was to help deal with basewalking and being fired upon by newly built structures that you couldn't see.

I don't know where the game meta is right now or if basewalking is still a problem, but I would suggest further discussion on possible implementation of this if basewalking is still an issue.

pchote:
@AoAGeneral: This may have flown under your radar.  How do you think the community would react to the idea of introducing this to TD?

AoAGeneral:
Short discussion in IRC. I think its fine to have it on base defenses only if to be introduced. Having them on units is a good idea but it could very well turn into arty vs arty games or V2 vs V2 in both mods. Also having RevealOnFire with aircraft could be an issue as well since MSAMs or other AA in RA could be very strong without having to spot the unit. (Specially since in RA E3 have shorter ranges.)

Having it on defenses only for TD won't mean to much since once you scout it you scout it anyways and can lob shells on it.

[...]

PR #8479, opened 2015-06-13
Move more shroud logic to RevealsShroud / GeneratesShroud

pchote:
This moves the last of the actor-specific visibility code from `Shroud` on to the actors themselves.  It is an ugly set of changes with no visible in-game effects, but sets the scene for future PRs that will introduce new features.  Depends on #8474, #8477.
- `Shroud` is now (if we ignore the fragile alliances workaround) focused on a much narrower scope of shroud queries.  Logic related to working out the shroud revealing/generation area now lives in `RevealsShroud`/`GeneratesShroud` and logic related to determining actor visibility is in `IDefaultVisibility` implementations (from #8477).
- This prepares for (but stops short of) splitting `GetVisOrigins` into separate logic for "cells I reveal from" and "cells that reveal me" (needed for heightmap shroud and faithful D2K shroud revealing).
- This prepares for (but stops short of) adding support for non-actor and non-circular shroud reveal sources (#5061, #7916, etc)
- This takes us yet another step towards heightmap-aware shroud.

### DIFF
--- a/OpenRA.Mods.Common/Traits/RevealsShroud.cs
+++ b/OpenRA.Mods.Common/Traits/RevealsShroud.cs
@@ -9,30 +9,30 @@
 #endregion
 
 using System.Linq;
+using OpenRA.Traits;
 
-namespace OpenRA.Traits
+namespace OpenRA.Mods.Common.Traits
 {
-	public class CreatesShroudInfo : ITraitInfo
+	public class RevealsShroudInfo : ITraitInfo
 	{
 		public readonly WRange Range = WRange.Zero;
 
-		public object Create(ActorInitializer init) { return new CreatesShroud(init.Self, this); }
+		public object Create(ActorInitializer init) { return new RevealsShroud(init.Self, this); }
 	}
 
-	public class CreatesShroud : ITick, ISync, INotifyAddedToWorld, INotifyRemovedFromWorld
+	public class RevealsShroud : ITick, ISync, INotifyAddedToWorld, INotifyRemovedFromWorld
 	{
-		readonly CreatesShroudInfo info;
+		readonly RevealsShroudInfo info;
 		readonly bool lobbyShroudFogDisabled;
 		[Sync] CPos cachedLocation;
-		[Sync] bool cachedDisabled;
 
-		public CreatesShroud(Actor self, CreatesShroudInfo info)
+		public RevealsShroud(Actor self, RevealsShroudInfo info)
 		{
 			this.info = info;
 			lobbyShroudFogDisabled = !self.World.LobbyInfo.GlobalSettings.Shroud && !self.World.LobbyInfo.GlobalSettings.Fog;
 		}
 
-		CPos[] ShroudedTiles(Actor self)
+		CPos[] VisibleTiles(Actor self)
 		{
 			return Shroud.GetVisOrigins(self)
 				.SelectMany(o => Shroud.FindVisibleTiles(self.World, o, Range))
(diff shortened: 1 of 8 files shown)


pchote:
Rebased and addressed @penev92's comments.

OpenRA.Mods.Common/Traits/RevealsShroud.cs
```diff
@@ -0,0 +1,103 @@
...
+		{
+			this.info = info;
+			lobbyShroudFogDisabled = !self.World.LobbyInfo.GlobalSettings.Shroud && !self.World.LobbyInfo.GlobalSettings.Fog;
+
+			add = (p, c) => p.Shroud.AddVisibility(self, c);
+			remove = p => p.Shroud.RemoveVisibility(self);
+			isDisabled = () => false;
+		}
+
+		static IEnumerable<CPos> FindVisibleTiles(World world, CPos position, WRange radius)
```
penev92: I'm not sure about the name here. This is used to find both tiles that the actor reveals and tiles that the actor hides (when generating shroud).
How about something like `FindTilesInRange`, as that is exactly what it's doing.

OpenRA.Mods.Common/Traits/RevealsShroud.cs
```diff
@@ -0,0 +1,103 @@
...
+		public void Tick(Actor self)
+		{
+			if (lobbyShroudFogDisabled || !self.IsInWorld)
+				return;
+
+			var disabled = isDisabled();
+			if (cachedLocation == self.Location)
+				return;
+
+			var visible = Tiles(self);
```
penev92: Let's go with `var tiles` here, again because of the dual purpose of the code (I would have asked for `visibleTiles` otherwise anyway ;) ).

OpenRA.Mods.Common/Traits/RevealsShroud.cs
```diff
@@ -0,0 +1,103 @@
...
+				add(p, visible);
+			}
+
+			cachedLocation = self.Location;
+			cachedDisabled = disabled;
+		}
+
+		public void AddedToWorld(Actor self)
+		{
+			var visible = Tiles(self);
```
penev92: Same here.

penev92:
Well the code looks good, aside from those small naming nits.

pchote:
Fixed all nits, and moved the shroud cell helper into a ShroudExts class so that it can be reused elsewhere.

Mailaender:
Shroud revealing seems to work fine. :white_check_mark: Not sure about gap generators. They seem to do nothing.

pchote:
> Not sure about gap generators. They seem to do nothing.
Gap is only visible to enemies, so you will need to run with two clients to test that.

penev92:
OK so the `MPStartLocations` changes woke up my spidy sence (apparently for the wrong reason, but ... oh well).
I reduced the MCV `RevealsShroud` range to 1, and this is what I got:

Is this supposed to happen?

penev92:
Here is a map I put together to test the gap generation:
[Link to map](https://www.dropbox.com/s/fq3idvu5ngz3zto/visTest.oramap?dl=0)
And here is a problem it revealed:
[Link to replay](https://www.dropbox.com/s/gk3e9uwr95jtsr7/OpenRA-2015-06-21T085723Z.orarep?dl=0)

pchote:
>  Is this supposed to happen?

Yes - this is just d2k artwork being weird.  The same behaviour is present in the last release.

>  And here is a problem it revealed:

This was a problem with stale data that only affected map-placed buildings.  Good catch!  Moving the `cachedX = X` assignments above the cell queries fixed it.  Also rebased against latest bleed.

penev92:
Looks good to me :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/21d6bc6e643a8fffdfa69051341d2f317f341c9f...c2f088e8f2b5efd301515a48b50901f8232a91eb)

PR #8212, opened 2015-05-21
Added selection classes

matija-hustic:
Option 3 from #7469 that would support the type of functionality proposed by @MustaphaTR:

http://ares-developers.github.io/Ares-docs/new/typeselect.html

Basically whenever the standard select-by-type functionality is supposed to be extended to select any custom class of selectable actors, these custom classes can be defined using the Class property of the Selectable trait.

If the basic idea is fine, I'd define any remaining selection classes in the YAMLs.

PR #8483, opened 2015-06-14
Fixes regression with selection classes

matija-hustic:
Fixes regression from #8212, as @Phrohdoh remarked:
Select all units on screen should use selection priorities to determine who actually gets selected. I have falsely understood the function of this command to be selecting all units that belong to the current player no matter the selection priorities.

### DIFF
--- a/OpenRA.Game/OpenRA.Game.csproj
+++ b/OpenRA.Game/OpenRA.Game.csproj
@@ -156,6 +156,7 @@
     <Compile Include="Primitives\MergedStream.cs" />
     <Compile Include="Primitives\SpatiallyPartitioned.cs" />
     <Compile Include="Primitives\ConcurrentCache.cs" />
+    <Compile Include="SelectableExts.cs" />
     <Compile Include="Selection.cs" />
     <Compile Include="Server\Connection.cs" />
     <Compile Include="Server\Exts.cs" />
--- a/OpenRA.Game/SelectableExts.cs
+++ b/OpenRA.Game/SelectableExts.cs
@@ -1,0 +1,67 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OpenRA.Traits
+{
+	public static class SelectableExts
+	{
+		public static int SelectionPriority(this ActorInfo a)
+		{
+			var selectableInfo = a.Traits.GetOrDefault<SelectableInfo>();
+			return selectableInfo != null ? selectableInfo.Priority : int.MinValue;
+		}
+
+		const int PriorityRange = 30;
+
+		public static int SelectionPriority(this Actor a)
+		{
+			var basePriority = a.Info.Traits.Get<SelectableInfo>().Priority;
+			var lp = a.World.LocalPlayer;
+
+			if (a.Owner == lp || lp == null)
+				return basePriority;
+
+			switch (lp.Stances[a.Owner])
+			{
+				case Stance.Ally: return basePriority - PriorityRange;
+				case Stance.Neutral: return basePriority - 2 * PriorityRange;
+				case Stance.Enemy: return basePriority - 3 * PriorityRange;
+
+				default:
+					throw new InvalidOperationException();
+			}
+		}
+
+		public static Actor WithHighestSelectionPriority(this IEnumerable<Actor> actors)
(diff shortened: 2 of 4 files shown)


OpenRA.Game/OpenRA.Game.csproj
```diff
@@ -156,6 +156,7 @@
     <Compile Include="Primitives\MergedStream.cs" />
     <Compile Include="Primitives\SpatiallyPartitioned.cs" />
     <Compile Include="Primitives\ConcurrentCache.cs" />
+    <Compile Include="SelectableExts.cs" />
```
penev92: I think this would be better suited in the `Traits` folder. (especially since it's in the `Traits` namespace)

matija-hustic: I know you suggested this, but as far as I've noticed, these `Traits` folders usually contain only traits, right? Stuff you can attach to actors through the rule YAML files. That's why I opted for the `Game` folder. Did I understand this incorrectly about only traits being in the `Traits` folder?

penev92: Well, in theory... yes.
In practice, however, we have tons of exceptions for every rule. And this is tightly coupled with the `Selectable` trait.
Also I don't consider it to be an engine thing.
Anyway, let's hear a third opinion.

GraionDilach: I personally don't even see the gain of extracting it to a file on it's own over just adding it to the end of the Selectable's file.

penev92: Convention is one class per file, because it makes navigation much easier.

reaperrr:
I have a hunch that if this is merged first, rebasing #8388 would become a real nightmare. The reverse case doesn't seem as bad, so if possible, I'd like my PR to be merged first...

reaperrr:
Needs rebase.

matija-hustic:
Rebased.

obrakmann:
Looks fine to me, selection works as expected again. :+1:

pchote:
Could you also please squash the last three commits together?  We try to avoid unnecessary rewrite noise within an individual changeset.

matija-hustic:
Style related issues fixed, commits squashed. Do I look into the `PriorityExts`?

pchote:
Sorry, I didn't notice that you were just moving existing code around there.  The priority issue can be left to another PR.

OpenRA.Game/SelectableExts.cs
```diff
@@ -0,0 +1,81 @@
...
+		public static IEnumerable<Actor> SubsetWithHighestSelectionPriority(this IEnumerable<Actor> actors)
+		{
+			return actors.GroupBy(x => x.SelectionPriority())
+				.OrderByDescending(g => g.Key)
+				.Select(g => g.AsEnumerable())
+				.DefaultIfEmpty(NoActors)
+				.FirstOrDefault();
+		}
+
+		public static IEnumerable<Actor> WithPlayerAndClass(this IEnumerable<Actor> actors, Player player, IEnumerable<string> selectionClasses)
```
pchote: `IEnumerable<Actor>`s are used for a bunch of non-selection things too.  Could we rename this to something like `WhereSelectableWithPlayerAndClass` (or ideally something thats less of a mouthful but the same meaning).

matija-hustic:
And btw, I've taken a quick look, that function `SelectableExts.SelectionPriority(this Actor)` is used in two cases.
* Selecting all actors on the screen that belong to the current (render) player. In this case it never gets called for enemy players.
* Selecting units with a dragbox, where multiple selection remains possible even for enemy units, but allied units always have priority.

OpenRA.Game/SelectableExts.cs
pchote: This is a rather specific method, so an alternative would be to keep it in `WorldInteractionControllerWidget` where it can't leak out.

matija-hustic: To me it seems nicer or cleaner to use the LINQ-like syntax with an extension `IEnumerable<Actor>.WithPlayerAndSelectionClass()` than to have a private method in `WorldInteractionControllerWidget` that requires doing this by calling `SelectByPlayerAndSelectionClass(IEnumerable<Actor>)`

Would the current implementation with a change of name to `WithPlayerAndSelectionClass` be OK?

matija-hustic:
OK, fixed as commented.

penev92:
Two things:
 - What regression are we talking about?
 - You should have stayed with the 3 commit split. Separate commits make everything easier (especially review) if they are properly separated logical units. (to keep in mind for next time)

matija-hustic:
First thing: https://github.com/OpenRA/OpenRA/pull/8483#issue-88163521
Second thing: https://github.com/OpenRA/OpenRA/pull/8483#issuecomment-113937954

penev92:
Code looks OK and it works ingame, so :+1:

PR #8490, opened 2015-06-15
Add campaign.psd for D2k mission previews [Superseded]

penev92:
Courtesy of @x-a-n-a-x.
Really helps with making a mission preview for missions.


abcdefg30:
:+1: / :white_check_mark: i guess.

x-a-n-a-x:
Don't you think we should compress (zip or better 7z) the PSD file? 
It would break the github PSD preview but it would reduce the file size from 2 mb to 909 kb.

After all, it's not like everybody will want to use the file often. On another hand... many people regulary download the source code to build it.. 

Soon or later, I also plan to release the PSD files I use to create the UIs but it's the same problem: they are about 3 mb to 7mb uncompressed. I think I will have to 7z them (but @pchote seemed reluctant when I talked to him about this file size issue).

Mailaender:
Please submit it to https://github.com/OpenRA/ArtSrc instead.

Mailaender:
Superseded by https://github.com/OpenRA/ArtSrc/pull/1.

Issue #8444, opened 2015-06-08
Elite weapon(s) in TS broken


reaperrr:
I don't know yet whether this is a yaml or code issue, but in any case elite weapons don't work.

The Nod bike should get the HoverMissile @ elite, but it doesn't.

matija-hustic:
This is a trait construction order issue at its core.

The problem is that when `UpgradeManager.GrantUpgrade` gets called to grand the `eliteweapons` upgrade, the `Armament` traits are not present in the lazy `UpgradeManager.upgrades` dictionary. This is because this dictionary enumerates all constructed traits implementing `IUpgradable` at the time of its first use. And in this particular case, when `^Vehicle.TimedUpgradeBar` gets constructed, it accesses `UpgradeManager.upgrades` thus enumerating only traits that have been constructed up until that moment. And it just so happens that `Armaments` haven't yet been constructed.

If the `IUpgradable` interface is here to stay and this cannot be resolved through `Requires<>`, then a constraint on the use of `UpgradeManager` should be made, forbidding any access to `UpgradeManager.upgrades` during trait construction.

matija-hustic:
If that's OK, I can take care of this in #8496

PR #8496, opened 2015-06-16
Fixed upgrade level tracking

matija-hustic:
Fixes `UpgradeManager` keeping track of traits' upgrade levels cumulatively without keeping track of the individual upgrades' levels.

This has made `GainsStatUpgrades` grant maximum upgrades after only two rank ups.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
@@ -51,7 +51,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		readonly List<TimedUpgrade> timedUpgrades = new List<TimedUpgrade>();
 		readonly Lazy<Dictionary<string, UpgradeState>> upgrades;
-		readonly Dictionary<IUpgradable, int> levels = new Dictionary<IUpgradable, int>();
+		readonly Dictionary<IUpgradable, Dictionary<string, int>> levels = new Dictionary<IUpgradable,Dictionary<string,int>>();
 
 		public UpgradeManager(ActorInitializer init)
 		{
@@ -87,18 +87,19 @@ namespace OpenRA.Mods.Common.Traits
 		{
 			foreach (var up in traits)
 			{
-				var oldLevel = levels.GetOrAdd(up);
-				var newLevel = levels[up] = oldLevel + levelAdjust;
+				var oldLevel = levels.GetOrAdd(up).GetOrAdd(upgrade);
+				var newLevel = levels[up][upgrade] = oldLevel + levelAdjust;
 
 				// This will internally clamp the levels to its own restricted range
 				up.UpgradeLevelChanged(self, upgrade, oldLevel, newLevel);
 			}
 		}
 
-		int GetOverallLevel(IUpgradable upgradable)
+		int GetOverallLevel(IUpgradable upgradable, string upgrade)
 		{
 			int level;
-			return levels.TryGetValue(upgradable, out level) ? level : 0;
+			Dictionary<string, int> upgrades;
+			return levels.TryGetValue(upgradable, out upgrades) && upgrades.TryGetValue(upgrade, out level) ? level : 0;
 		}
 
 		public void GrantUpgrade(Actor self, string upgrade, object source)
@@ -131,7 +132,7 @@ namespace OpenRA.Mods.Common.Traits
 			if (!upgrades.Value.TryGetValue(upgrade, out s))
 				return false;
 
-			return s.Traits.Any(up => up.AcceptsUpgradeLevel(self, upgrade, GetOverallLevel(up) + 1));
+			return s.Traits.Any(up => up.AcceptsUpgradeLevel(self, upgrade, GetOverallLevel(up, upgrade) + 1));
 		}
 
 		public void RegisterWatcher(string upgrade, Action<int, int> action)


pchote:
We need to be a bit careful about this.  For some traits we may indeed want to track the aggregate level across multiple types instead of having them tracked separately.  Somebody needs to take charge of this feature and write up something specifying exactly how upgrades and their levels should interact with traits in all the different cases, and then we can work from there.

matija-hustic:
The current state of the upgrade system:
https://github.com/OpenRA/OpenRA/wiki/The-upgrade-system

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -51,7 +51,7 @@ class UpgradeState
 
 		readonly List<TimedUpgrade> timedUpgrades = new List<TimedUpgrade>();
 		readonly Lazy<Dictionary<string, UpgradeState>> upgrades;
-		readonly Dictionary<IUpgradable, int> levels = new Dictionary<IUpgradable, int>();
+		readonly Dictionary<IUpgradable, Dictionary<string, int>> levels = new Dictionary<IUpgradable,Dictionary<string,int>>();
```
obrakmann: OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs:L54: [SA1001] Invalid spacing around the comma.

matija-hustic: I've already fixed the compile error locally. I'm waiting for @pchote to get a chance to review the description of the current upgrade system :)

obrakmann:
Code looks ok to me. As expected, it takes a while longer to get fully-ranked units now.

As for the regressions, I don't think we even have any traits that make use of upgrade levels directly. Even `GainsStatUpgrades` implements its own levels on top of upgrades.

GraionDilach:
Does this fixes #8444 ?

matija-hustic:
It shouldn't. This only concerns traits which implement `IUpgradable` directly (only `GainsStatUpgrades`). Descendants of `UpgradableTrait` like `Armament` should function properly with things as they are.

GraionDilach:
I'm also a bit confused about this PR and the regressions it'd bring - I don't really grasp the difference in this scenario.

I guess it's related to the following: The doc at https://github.com/OpenRA/OpenRA/wiki/Actor-Upgrades implies me that I'm allowed to use the @ mark with, say `GainsStatUpgrades` to trigger them at `GrantUpgrade` warheads - (tho I'm uncertain if that actually works atm - probably due to #8444) - should that be a valid implication?

matija-hustic:
With things as they are, `GainsStatUpgrades` doesn't work as one would expect it to. What one expects is that this trait implements four independent upgrades: e.g. if I grant to a unit from TS `firepower` and `speed`, the actor should only inflict 10% more damage when attacking and move 20% more quickly. However, right now, if you just grant each of those two upgrades once, you will get full elite behavior: 30% more firepower, 66% damage taken from attacks, 50% quicker movement and 75% the reload time. This is clearly not what this trait is meant to do. Now, in order to properly fix this, a blueprint of the upgrade system's logic is necessary and needs to be agreed upon. So this is what I'm waiting for right now.

`UpgradableTrait` however (and all traits that inherit from it), works correctly with things as they are.

WolfGaming:
I was under the assumption that GainsStatUpgrades was one upgrade that had multiple effects, is that wrong?

matija-hustic:
@WolfGaming, `GainsStatUpgrades` currently consumes 5 different upgrades: `firepower`, `damage`, `speed`, `reload` and `inaccuracy`

atlimit8:
Actually, I plan on splitting GainsStatUpgrades into separate traits and have since about the time of the introduction of UpgradableTrait. The problem with GainsStatUpgrades is that it clumps functionality together (fire-power modifier, damage modifier,...). One reason for the bug may be this bad design making functionality non-obvious. If I remember correctly, I'm not the only one who thinks it should be split.

Since GainsStatUpgrades should be replaced with separate traits and this seems to have no other use (so far), This probably should be closed for now at least and possibly an issue opened regarding this functionality to figure out what to do first.

penev92:
We have a bug. We have a fix for it. Why would we close the PR with the fix, just to open an issue about the bug? o.O

The bug at hand here has nothing to do with any potential refactoring of the trait.

matija-hustic:
@penev92, it was @pchote that wished to first establish what the desired functionality of the upgrade system is before actually fixing anything. If what @atlimit8 is saying should be done, i.e., if the system should be designed so that each trait uses at most one parameter (sum of all upgrade levels) instead of multiple parameters (independent upgrade levels for different upgrades), then
* `GainsStatUpgrades` indeed needs refactoring and
* I'd say the notifications, that the `UpgradeManager` sends, should contain only information on this one parameter (upgrade level sum) instead of also saying which upgrade triggered the latest aggregate level change.
That would make this PR unwanted.

If, on the other hand, traits should be allowed to provide upgrades that depend on multiple independent parameters (upgrade levels), then this PR would be OK, but the `GainsStatUpgrades` would be the only trait using this functionality and the `UpgradableTrait` should be updated (this is what I've done locally).

Basically the question is - single or multiple parameters for upgrades?

GraionDilach:
Honestly, now that I see @atlimit8's implementation, I'd vote for his one to axe `GainsStatUpgrades`.

penev92:
Once his PR gets merged - sure.

matija-hustic:
I'll try and see then if this effective preference for single parameter upgrades means IUpgradable can be removed and replaced exclusively with UpgradableTrait

abcdefg30:
Travis fails at the moment:
```
OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs:L54: [SA1001] Invalid spacing around the comma.

matija-hustic:
#8572 and #8586 took care of whatever I mean to do with this PR.

penev92:
:-1: 

 - You just closed a 6-line bugfix PR in favor of a giant refactoring PR that may or may not fix the problem.
 - That big refactoring PR also may or may not get merged.
 - Last I checked, neither of those PRs actually had anything to do with the bug being fixed here.

matija-hustic:
Well, the decision has been implicitly made that upgradable traits will use the sum of all upgrade levels instead of individual upgrade levels. That means that this PR is not needed.

And I'm pretty sure #8572 will take care of the problem with `GainsStatUpgrades`. The only thing, that might be needed in the future, is some explicit documentation on the agreements and conventions regarding the upgrade system so that regressions like this don't happen again. But I'm not sure if that even needs its own PR.

pchote:
I still think the best choice is to remove the "all upgrade levels" from UpgradeableTrait (i.e. array &rarr; single type), and avoid the ambiguity completely.

penev92:
From what I've seen, the bug is inside `UpgradeManager`, with `GrantStatsUpgrades` being the only one invoking it. That PR removes `GrantStatsUpgrades`, effectively hiding the bug. It does nothing to fix it.

atlimit8:
I plan on adding traits that act as expressions of upgrades to grant/revoke upgrades.
Perhaps, those should be the only ones to accept multiple upgrades.

matija-hustic:
@penev92 whether or not the bug exists depends on what you want the `UpgradeManager` to do. If you want it to track aggregate levels, there is no bug. If you want it to track individual upgrade levels, then this PR might be needed. But I thought the point of #8572 is to make a transition onto upgradable traits that only use aggregate levels.

PR #8499, opened 2015-06-17
D2K - Fix Barracks Footprint

MustaphaTR:
Current one loks like only true for Atreides Barracks and looks weird for Harkoonen and Ordos ones.


GraionDilach:
This fix is already a part of #8388.

MustaphaTR:
Closing then.

PR #8506, opened 2015-06-18
Fix TS maps.

pchote:
- Fixes the incorrectly named TEMPERAT tileset to match the original game.
- Fixes a bug in Map.CenterOfCell that breaks the rendering of heighmapped cells outside the map bounds.
- Reimports the original TS maps using an improved importer that understands resources and cells outside the map bounds.


pchote:
I found the full list of overlaytypes in FinalSun, and will reimport the maps again after implementing it.

pchote:
Updated.  This now includes a placeholder for veins (which future PRs can build on to add proper vein rendering/growth/attacking), proper import of rocks, and two new maps.  I've removed the old test map, because we now support heightmaps well enough to not need it.

pchote:
I suggest testing this on https://github.com/pchote/OpenRA/tree/test-working-clip, which has proper shroud behaviour.

Mailaender:
:+1: on these changes.

> Reimports the original TS maps using an improved importer that understands resources and cells outside the map bounds.

I would like to see the code for the improved importer probably alongside. This is just showing it's result.

mods/ts/rules/trees.yaml
```diff
@@ -94,6 +94,20 @@ TREE24:
 TREE25:
 	Inherits: ^Tree
 
+VEINHOLEDUMMY:
+	HiddenUnderShroud:
+	RadarColorFromTerrain:
+		Terrain: Tiberium
```
Mailaender: This should probably be `Terrain: Veins` now.

pchote:
Updated the veinhole radar color and fixed the hot springs ambient lighting.

> I would like to see the code for the improved importer probably alongside. This is just showing it's result.

The updated importer is where it has always been: my ts-map-importer branch.

penev92:
I'm still not comfortable with removing the blank map:
 - None of the heightmap stuff is finished yet, and this gives us a nice flat baseline for comparison.
 - It is still helpful for easy testing/debugging.
 - It doesn't hurt anyone.

Let's start a petition to save Blank map! :D

pchote:
Rebased and removed the legacy editor cruft.

pchote:
> None of the heightmap stuff is finished yet, and this gives us a nice flat baseline for comparison.

Enough of the heightmap stuff is finished that we don't need a baseline to compare against.

> It is still helpful for easy testing/debugging.

Example?

> It doesn't hurt anyone.

This is my main reason for removing it.  For some reason whenever I switch mods (and invalidate the cached map hash) TS will *always* default to the blank map, and so I always need to manually switch to one of our other maps to test whatever i'm aiming to test.  I hate the blank map with a burning passion (okay, maybe not... but it annoys the crap out of me).

Mailaender:
> I'm still not comfortable with removing the blank map.

Just back it up in your user map folder if you still need it badly for testing.

reaperrr:
The veinhole visuals are off by 48 vertical pixels (2 cells), but I can deal with that in #8513.
:+1:

PR #8518, opened 2015-06-20
Fix screen bounds calculations

pchote:
This fixes the screen bounds calculation (in diamond coords there is only _one_ height step per map row), drops the arbitrary padding factors that tried to work around that bug, and adds comments explaining each step.

### DIFF
--- a/OpenRA.Game/Graphics/Viewport.cs
+++ b/OpenRA.Game/Graphics/Viewport.cs
@@ -271,7 +271,7 @@ namespace OpenRA.Graphics
 			{
 				if (cellsDirty)
 				{
-					cells = CalculateVisibleCells(false);
+					cells = CalculateVisibleCells(true);
 					cellsDirty = false;
 				}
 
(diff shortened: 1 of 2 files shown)


pchote:
I recommend testing this by loading the "a river runs near it" map in ts, clearing the shroud, and looking around the edges of the map on https://github.com/pchote/OpenRA/tree/test-working-clip versus https://github.com/pchote/OpenRA/tree/test-broken-clip

Mailaender:
https://github.com/pchote/OpenRA/tree/test-working-clip seems to work fine while https://github.com/pchote/OpenRA/tree/test-broken-clip adds a cut-off line to the top of the map. This pull request however only seems to add additional artifacts to the TS cliffs so I am not sure what how to judge this in isolation.

pchote:
> This pull request however only seems to add additional artifacts to the TS cliffs so I am not sure what how to judge this in isolation.

In isolation this should introduce no visible clipping at the screen edge when inside the map bounds, or anywhere in TD/RA/D2K.  The map-edge behaviour cannot be tested with TS bleed because the bleed shroud is not compatible with heightmaps.  The test-broken-clip branch contains the heightmap-aware shroud, which exposes the problem you mentioned above (line along the top of the map, also a cliff sticking out the left edge).  The test-working-clip branch contains this fix on top of test-broken-clip, and fixes the problem as you mentioned above.

Mailaender:
:+1:

pchote:
Fixed scissoring in the editor.

obrakmann:
yup, it's fine now. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/53a25e7b7a747fbf7b5e6072ec9f43958fb38e57%5E...53a25e7b7a747fbf7b5e6072ec9f43958fb38e57)

Issue #8493, opened 2015-06-15
DllNotFoundException: Unable to load DLL 'freetype6.dll'


RedAlert2:
Just installed the 20150614 release, but could not launch it.

From exception.log:
Inner Exception of type `System.DllNotFoundException`: Unable to load DLL 'freetype6.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)

The file 'freetype6.dll' is in the OpenRA.exe directory.
OS: Windows 7

penev92:
Please make sure you have Microsoft Visual C++ 2010 Redistributable or later.

RedAlert2:
Thanks. Installed Microsoft Visual C++ 2010 Redistributable and it is now working perfectly.

Mailaender:
I intend to bring back the check during setup: https://github.com/OpenRA/OpenRA/pull/7609 as there seem to be Windows 7 systems in the wild that don't have the 2010 version installed. See also https://github.com/OpenRA/OpenRA/pull/8360.

GraionDilach:
:+1: for bringing in such a check... there's even a report of this by a guest on the ModDB profile who fixed it via getting a version from sourceforge...?

Mailaender:
You can test https://ci.appveyor.com/project/OpenRA/openra/build/1.0.3936/artifacts and give your thumps up at https://github.com/OpenRA/OpenRA/pull/7609.

Mailaender:
Added FAQ entry ([diff](https://github.com/OpenRA/OpenRA/wiki/FAQ/_compare/ea6343f9c1baaf039c6453a258994646043f405f...74811b1a9400ca68cadb395f454393756864e45f)) as this is going to be a problem. Failed to do the check at install time. https://github.com/OpenRA/OpenRA/issues/7609 We may want to do the check on game start instead which gives us more flexibility.

RedAlert2:
Was installing OpenRA on another machine, did a search for the C++ package but a slightly different link came up, turns out there are 4 different package available for download:

Microsoft Visual C++ 2010 Redistributable Package (x86)
Date Published: 4/12/2010
https://www.microsoft.com/en-us/download/details.aspx?id=5555

Microsoft Visual C++ 2010 Redistributable Package (x64) 
Date Published: 4/12/2010
https://www.microsoft.com/en-us/download/details.aspx?id=14632

Microsoft Visual C++ 2010 SP1 Redistributable Package (x86) 
Date Published: 3/3/2011
[...]

PR #7609, opened 2015-03-07
Added a check for Visual C++ Redistributable during install time.

Mailaender:
as proposed in https://github.com/OpenRA/OpenRA/pull/7579#issuecomment-77186857 by @obrakmann.

Took https://github.com/UberEnt/NSISInstallers/blob/master/VCRedist11.nsh as a reference.

PR #8523, opened 2015-06-21
Added a check for Visual Studio runtime at game start

Mailaender:
Another attempt to solve https://github.com/OpenRA/OpenRA/issues/8493.

### DIFF
--- a/OpenRA.GameMonitor/GameMonitor.cs
+++ b/OpenRA.GameMonitor/GameMonitor.cs
@@ -15,16 +15,44 @@ using System.IO;
 using System.Media;
 using System.Reflection;
 using System.Windows.Forms;
+using Microsoft.Win32;
 
 namespace OpenRA
 {
 	class GameMonitor
 	{
+		static readonly string[] VisualStudioRuntimeVersions = { "10.0", "11.0", "12.0" };
+
 		static Process gameProcess;
 
 		[STAThread]
 		static void Main(string[] args)
 		{
+			if (Platform.CurrentPlatform == PlatformType.Windows)
+			{
+				var visualStudioRuntimeInstalled = 0;
+
+				foreach (var version in VisualStudioRuntimeVersions)
+				{
+					var key = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\DevDiv\VC\Servicing\{0}\RuntimeMinimum\".F(version), false);
+					if (key != null && visualStudioRuntimeInstalled == 0)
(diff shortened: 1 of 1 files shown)


penev92:
OK, this seems to mostly work, with 2 comments:

OpenRA.GameMonitor/GameMonitor.cs
```diff
@@ -15,16 +15,44 @@
...
+			{
+				var visualStudioRuntimeInstalled = 0;
+
+				foreach (var version in VisualStudioRuntimeVersions)
+				{
+					var key = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\DevDiv\VC\Servicing\{0}\RuntimeMinimum\".F(version), false);
+					if (key != null && visualStudioRuntimeInstalled == 0)
+						visualStudioRuntimeInstalled = (int)key.GetValue("Install", 0);
+
+					key.Close();
```
penev92: If you fail to open (find) the key, you get an NRE here. You do have a null check above though.

OpenRA.GameMonitor/GameMonitor.cs
```diff
@@ -15,16 +15,44 @@
...
+				}
+
+				if (visualStudioRuntimeInstalled == 0)
+				{
+					var result = MessageBox.Show("OpenRA needs the Microsoft Visual C++ 2010 (or above) x86 Redistributable installed.",
+							"Visual C++ runtime not found.", MessageBoxButtons.OK, MessageBoxIcon.Error);
+
+					if (result == DialogResult.OK)
+						Process.Start("http://www.microsoft.com/en-us/download/details.aspx?id=5555");
```
penev92: As a user, I really hate it when programs decide they can open random links whenever they feel like it. This is great to help people find what they need, but I'd rather it be optional.
Please replace `MessageBoxButtons.OK` above with `MessageBoxButtons.YesNo` and check if `result == Yes`.
(And add a "Do you want OpenRA to guide you to the download link now?" or something to the message.)

pchote:
Doing this in our game (launcher) code At runtime is madness.  This kind of thing is what installers are for.

Mailaender:
What I was asked for isn't possible in NSIS afaik, see https://github.com/OpenRA/OpenRA/pull/7609 but anyway go fix it yourself then.

penev92:
The check was perfectly fine here. Actually made sure you didn't uninstall the prerequisites.
We even have a dedicated launcher for the occasion.

PR #8527, opened 2015-06-21
Introduce new options for shroud visibility and revealing.

pchote:
This splits the `GetVisOrigins` query into different configurable options for shroud visibility and revealing.

`HiddenUnderShroud.Type = CenterPosition` will be used in TS to ensure that units are revealed correctly when they are on an odd-height tile (where the shroud tiles are staggered over position tiles, and units would otherwise be revealed if any of the 4 nearby tiles are revealed).  It does not change anything in the classic mods. 

`RevealsShroud.Type = CenterPosition` is used for units that should reveal in a circle around their center instead of from every cell in their footprint.  This is used for aircraft (which don't have any cells in their footprint), and can be used to replicate the original games' building shroud revealing behavior.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Modifiers/HiddenUnderShroud.cs
+++ b/OpenRA.Mods.Common/Traits/Modifiers/HiddenUnderShroud.cs
@@ -15,32 +15,37 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
+	public enum VisibilityType { Footprint, CenterPosition }
+
 	[Desc("The actor stays invisible under the shroud.")]
 	public class HiddenUnderShroudInfo : ITraitInfo, IDefaultVisibilityInfo
 	{
 		[Desc("Players with these stances can always see the actor.")]
 		public readonly Stance AlwaysVisibleStances = Stance.Ally;
 
+		[Desc("Possible values are CenterPosition (reveal when the center is visible) and ",
+			"Footprint (reveal when any footprint cell is visible).")]
+		public readonly VisibilityType Type = VisibilityType.Footprint;
+
 		public virtual object Create(ActorInitializer init) { return new HiddenUnderShroud(this); }
 	}
 
 	public class HiddenUnderShroud : IDefaultVisibility, IRenderModifier
 	{
-		readonly HiddenUnderShroudInfo info;
+		protected readonly HiddenUnderShroudInfo Info;
 
 		public HiddenUnderShroud(HiddenUnderShroudInfo info)
 		{
-			this.info = info;
-		}
-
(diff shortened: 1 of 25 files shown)


pchote:
I guess I'll need to write an upgrade rule for the aircraft adjustment...

pchote:
Added upgrade rule.

OpenRA.Mods.Common/Traits/Modifiers/HiddenUnderShroud.cs
```diff
@@ -15,32 +15,36 @@
...
 {
+	public enum VisibilityType { CenterPosition, Footprint }
+
 	[Desc("The actor stays invisible under the shroud.")]
 	public class HiddenUnderShroudInfo : ITraitInfo, IDefaultVisibilityInfo
 	{
 		[Desc("Players with these stances can always see the actor.")]
 		public readonly Stance AlwaysVisibleStances = Stance.Ally;
 
+		[Desc("Possible values are Point (reveal when the center is visible) and ",
```
penev92: Probably outdated description - no value `Point` is available.

OpenRA.Mods.Common/Traits/RevealsShroud.cs
```diff
@@ -18,6 +18,10 @@ public class RevealsShroudInfo : ITraitInfo
 	{
 		public readonly WRange Range = WRange.Zero;
 
+		[Desc("Possible values are Point (measure range from the center) and ",
```
penev92: Again `Point`.

pchote:
Fixed.

RoosterDragon:
Do we need to do something special to adapt ```Immobile``` for this change - since you can set it to return an empty set of occupied cells too?

Are we worried about any other traits that may have been relying on the previous behaviour where an empty set of occupied cells fell back to using the center position?

pchote:
> Do we need to do something special to adapt Immobile for this change - since you can configure it to return an empty set of occupied cells too?

Apparently yes - I have broken at least the support power cameras.

pchote:
Rebased and fixed.  `VisibilityType.Footprint` now retains the legacy behaviour, where actors with no footprint fall back to `CenterPosition`.

obrakmann:
Code looks good to me and works in-game. Didn't spot any regressions, either. :+1:

pchote:
Went back to the previous implementation without the fallback.  Enjoy the new upgrade rules and manual rule workarounds ;)

penev92:
Looks OK to me :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4d1e87e59fe2c860daccf2487007f8bd339a91e1...b1cf0b3dd7047bd3493c371242dee2123290dfee)

PR #7930, opened 2015-04-12
Added a hotkey to take screenshots

Mailaender:
Closes #5317
Fixes #7624

I added support for a wide range of file formats:

image

To test press `CTRL` + `P` (this is configurable in the hotkey settings).

PR #8538, opened 2015-06-21
Stripped down the platform dependencies

Mailaender:
I added https://github.com/OpenRA/SDL2-CS but stripped it to only add what we need, so the unused SDL2_image.cs, SDL2_mixer.cs, SDL2_ttf.cs, AudioCapture.cs, but most importantly GL.cs and GLCore.cs have been reduced (~100x less methods) which I dubbed it MicroTK now. Inpired by https://github.com/flibitijibibo/SDL2-CS/issues/20#issuecomment-67949509.

The benefits are:
- 61 OpenGL extensions loaded instead of EVERYTHING
- 2,1 MB size reduction of our binary package
- easier packaging for distros that require to build completely from source

I also removed the OpenAL (and subsequent OpenTK) dependency from OpenRA.Game and placed it into the renderer DLLs, which in fact do a lot more than just rendering nowadays (input, window creation, etc.) so I renamed things to reflect this. This became a bigger job than I thought. Hope it was worth it.


reaperrr:
This doesn't create the OpenRA.exe anymore, which is needed to launch on Windows.

pchote:
It would be more maintainable if we could keep the SDL2-CS dependency (using their main branch without the MiniTK cruft - fetched from nuget), and then had our own OpenRA.MicroTK.dll library in a separate repository for the gl bindings.

Mailaender:
> It would be more maintainable if we could keep the SDL2-CS dependency (using their main branch without the MiniTK cruft - fetched from nuget)

That was rejected upstream. https://github.com/flibitijibibo/SDL2-CS/issues/81

Mailaender:
> and then had our own OpenRA.MicroTK.dll library in a separate repository for the gl bindings.

That will be a pain for Linux packaging again and not more maintainable, because you need to keep those in sync without direct compiler check connection. I tried that at first, but it was nearly impossible.

pchote:
We are going to need to find some compromise, because, to be blunt, I don't think it is reasonable for you to ask us to add nearly 30000 lines of code (thats 30% of the entire project!) to save 2MB in our binary packages.

Unit158:
This is impossible to review, my phone's browser crashes trying to load the diff :-/

Mailaender:
I am also not yet quite happy with it. Will see if I can strip this further and split into several commits, but I wanted to hear feedback before I dump another 1/2 weekend on it. So far I didn't hear destructive criticism only so I may drop the idea entirely.

> I don't think it is reasonable for you to ask us to add nearly 30000 lines of code (thats 30% of the entire project!) and maintain an incompatible fork of our low-level platform integration to save 2MB in our binary packages.

The main reason is not really file size especially in these days, but runtime performance. We are already maintaining an incompatible dead-end fork at https://github.com/OpenRA/SDL2-CS/tree/openra This just makes it easier for downstream packagers to build it.

Mailaender:
This also depends on https://github.com/OpenRA/OpenRA/pull/7930 which needs to go in first as it needs additional OpenGL API. That is also another reason against it. The benefit of @OpenTK is that you can browse and explore all those functions in your IDE and not write a wrapper at first.

Mailaender:
Rebased and stripped this down even further. All the math stuff and unused SDL2 hooks were removed. I also drastically removed line numbers by purging the XML docs which mostly contained redundant information.

pchote:
This does a couple of long-planned things (renaming the renderer dlls and moving the openal integration there), but overall my opinion is still :-1: on merging the SDL bindings into our main code base.  That feels like a big backwards step after all the work you did to split out our dependencies, and i dont feel that the justification is strong enough considering the architectural scope of this change.

Mailaender:
I tend to agree now. Superseded by https://github.com/OpenRA/OpenRA/pull/8632.

PR #8501, opened 2015-06-17
Dispose of a MemoryStream in PackageEntry

penev92:
There are probably a couple hundred `MemoryStreams` opened and never disposed of each time a mod loads (later research suggests also when a map is being loaded).

PR #8548, opened 2015-06-23
Revert "Dispose of a MemoryStream in PackageEntry"

obrakmann:
Reverts OpenRA/OpenRA#8501


Mailaender:
:+1: I am also affected. Discovered after a `make clean`.

PR #8554, opened 2015-06-24
Dispose of a MemoryStream in PackageEntry Take 2

Mailaender:
https://github.com/OpenRA/OpenRA/pull/8501


RoosterDragon:
Doesn't crash like #8501 (maybe it was an endianness issue) and seems reasonable otherwise. :+1:

PR #8558, opened 2015-06-25
Add target validation to GrantUpgrade warheads.

GraionDilach:
This seems like a bug to me.


GraionDilach:
Forget this, I overlooked it.

GraionDilach:
It's okay now.

chrisforbes:
:+1:

chrisforbes:
Can you send another PR that removes the 0% assumptions from the mods?

GraionDilach:
If it doesn't conflict with @penev92"s plans, I can look into it.

penev92:
I actually don't know what those comments refer to.

GraionDilach:
In the Westwood games, 0% versus on the warhead would not just make the target immune to the weapon, but also untargetable. Achieving this via TargetTypes only would be a lot more elegant.

penev92:
Looks reasonable :+1:

PR #8560, opened 2015-06-25
Fix accessing a disposed world when music ends after changing maps

### LINKED ISSUE #8559: Bleed TD crashes due to the Media.PlayMusic Lua script
This seems to be a regression of #8529 due to PlayMusic bleeding into a skirmish game.

Practically, if one has shellmap music - map1.aud, part of transit.mix - then after starting a skirmish game, the player must skip a track manually otherwise the game crashes when map1.aud ends.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23020777-bleed-td-crashes-due-to-the-media-playmusic-lua-script?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
Fixes #8559.

EDIT by @penev:
This is in fact a bandaid for the crash in #8559 for the sake of getting a playtest moving. The code being changed here is still a remnant of the previous world, running in the new one, which is conflicting with the recently introduced world disposing.

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
@@ -68,6 +68,9 @@ namespace OpenRA.Mods.Common.Scripting
 				var f = func.CopyReference() as LuaFunction;
 				onComplete = () =>
 				{
+					if (world.Disposing)
+						return;
+
 					try
 					{
 						using (f)


penev92:
With my changes to your comment above, and with another "oh god, this is so wrong on so many levels"... :+1:  for a playtest.

GraionDilach:
:-1: If it's "so wrong on so many levels", then it shouldn't be accepted, now that I filed #8562.

pchote:
Stopping any currently playing music when tearing down the world would be a better fix, and fits in with our old / long term plans to keep music world-local.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/8568.

Issue #8033, opened 2015-04-26
Music plays for one track, then doesn't continue


r34ch:
In RA the music starts to play on the splash screen and continues when you are ingame. Once the first track finishes however, it does not continue to play the rest of the tracks.

Release 20150424
Ubuntu 14.10

penev92:
Did you tick the "Loop" checkbox in the music menu?

reaperrr:
I noticed that, too. It just repeats that one track from the shellmap and doesn't cycle through the playlist.

r34ch:
@penev92 Loop is not enabled. I assume it means loop the *individual* track? (There is no tool-tip to clarify)

@reaperrr After joining a match, when the track ends I just get silence.

obrakmann:
Fixed with #8562.

PR #8528, opened 2015-06-21
TD mission script fixes

clemty:
nod06 (a, b, c): transport helicopter no longer selectable

nod06c would crash if you moved to the extraction point without first picking up the crate
easiest way to reproduce is to edit the map to move to the extraction point immediately

edit: changes added to also fix #8537
edit: changes added to also fix #8534
edit: gdi05b: player's tanks would crush the sandbags in the abandoned GDI base because their owner was set to neutral - set all sbags to either gdi or nod
edit: nod04a: set initial camera position; fixes #8532
edit: fixed sbag owner in gdi04b and nod05 
edit: music now starts playing when a mission starts

is there a way to make the helicopter land in nod06? currently it moves to the flare, then stays there

PR #6880, opened 2014-11-01
Added Media.PlayMusic to the Lua API

Mailaender:
and make use of it on the shellmaps to fix https://github.com/OpenRA/OpenRA/issues/6071.

PR #8562, opened 2015-06-26
Refactors the music player.

### LINKED ISSUE #8671: Music player stops after first song
RA mod, no shuffle, no loop selected. When game launches, music starts to play automatically. But it stops after the first song. I have the full scores.mix installed.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23813885-music-player-stops-after-first-song?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

GraionDilach:
Proper fix for #8559.
Closes  #8671, #8587 and #8033.

<del>ATM it has one regression: due to the playlist getting set up before the shellmap's lua scripts, the game plays a maximum of two seconds maybe before the shellmap music at game start, depending on settings. That will probably need a StartingMusic trait.</del>

### DIFF
--- a/OpenRA.Game/Traits/World/MusicPlaylist.cs
+++ b/OpenRA.Game/Traits/World/MusicPlaylist.cs
@@ -44,8 +44,7 @@ namespace OpenRA.Traits
 
 			random = playlist.Shuffle(Game.CosmeticRandom).ToArray();
 
-			if (Game.Settings.Sound.MapMusic
-				&& !string.IsNullOrEmpty(info.StartingMusic)
+			if (!string.IsNullOrEmpty(info.StartingMusic)
 				&& world.Map.Rules.Music.ContainsKey(info.StartingMusic)
 				&& world.Map.Rules.Music[info.StartingMusic].Exists)
 			{
--- a/OpenRA.Game/Traits/World/MusicPlaylist.cs
+++ b/OpenRA.Game/Traits/World/MusicPlaylist.cs
@@ -1,0 +1,150 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+using OpenRA.GameRules;
+
+namespace OpenRA.Traits
+{
+	[Desc("Trait for music handling. Attach this to the world actor.")]
+	public class MusicPlaylistInfo : ITraitInfo
+	{
+		public readonly string StartingMusic = null;
+		public readonly bool LoopStartingMusic = false;
+
+		public object Create(ActorInitializer init) { return new MusicPlaylist(init.World, this); }
+	}
+
+	public class MusicPlaylist : INotifyActorDisposing
+	{
+		readonly MusicInfo[] random;
+		readonly MusicInfo[] playlist;
+
+		public readonly bool IsMusicAvailable;
+
+		MusicInfo currentSong;
+		bool repeat;
+
+		public MusicPlaylist(World world, MusicPlaylistInfo info)
+		{
+			IsMusicAvailable = world.Map.Rules.InstalledMusic.Any();
+
+			playlist = world.Map.Rules.InstalledMusic.Select(a => a.Value).ToArray();
+
+			if (!IsMusicAvailable)
+				return;
+
+			random = playlist.Shuffle(Game.CosmeticRandom).ToArray();
+
+			if (Game.Settings.Sound.MapMusic
+				&& !string.IsNullOrEmpty(info.StartingMusic)
+				&& world.Map.Rules.Music.ContainsKey(info.StartingMusic)
+				&& world.Map.Rules.Music[info.StartingMusic].Exists)
+			{
+				currentSong = world.Map.Rules.Music[info.StartingMusic];
+				repeat = info.LoopStartingMusic;
+			}
+			else
+			{
+				currentSong = Game.Settings.Sound.Shuffle ? random.First() : playlist.First();
+				repeat = Game.Settings.Sound.Repeat;
+			}
+
+			Play();
+		}
+
+		public MusicInfo CurrentSong()
+		{
+			return currentSong;
+		}
+
+		public MusicInfo[] AvailablePlaylist()
+		{
+			// TO-DO: add filter options for Race-specific music
+			return playlist;
+		}
+
+		void Play()
+		{
+			if (currentSong == null || !IsMusicAvailable)
+				return;
+
+			Sound.PlayMusicThen(currentSong, () =>
+			{
+				if (!repeat)
+					currentSong = GetNextSong();
+
+				Play();
+			});
+		}
+
+		public void Play(MusicInfo music)
+		{
+			if (music == null || !IsMusicAvailable)
+				return;
+
+			currentSong = music;
+			repeat = Game.Settings.Sound.Repeat;
+
+			Sound.PlayMusicThen(music, () =>
+			{
+				if (!repeat)
+					currentSong = GetNextSong();
+
+				Play();
+			});
+		}
+
+		public void Play(MusicInfo music, Action onComplete)
+		{
+			if (music == null || !IsMusicAvailable)
+				return;
+
+			currentSong = music;
+			Sound.PlayMusicThen(music, onComplete);
+		}
+
+		public MusicInfo GetNextSong()
+		{
+			return GetSong(false);
+		}
+
+		public MusicInfo GetPrevSong()
+		{
+			return GetSong(true);
+		}
+
+		MusicInfo GetSong(bool reverse)
+		{
+			if (!IsMusicAvailable)
+				return null;
+
+			var songs = Game.Settings.Sound.Shuffle ? random : playlist;
+
+			return reverse ? songs.SkipWhile(m => m != currentSong)
+				.Skip(1).FirstOrDefault() ?? songs.FirstOrDefault() :
+				songs.Reverse().SkipWhile(m => m != currentSong)
+				.Skip(1).FirstOrDefault() ?? songs.Reverse().FirstOrDefault();
+		}
+
+		public void Stop()
+		{
+			currentSong = null;
+			Sound.StopMusic();
+		}
+
+		public void Disposing(Actor self)
+		{
+			if (currentSong != null)
+				Stop();
+		}
+	}
+}
--- a/OpenRA.Game/World.cs
+++ b/OpenRA.Game/World.cs
@@ -388,7 +388,6 @@ namespace OpenRA
 			frameEndActions.Clear();
 
 			Sound.StopAudio();
-			Sound.StopMusic();
 			Sound.StopVideo();
 
 			// Dispose newer actors first, and the world actor last
--- a/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
@@ -57,7 +57,7 @@ namespace OpenRA.Mods.Common.Scripting
 		[Desc("Play track defined in music.yaml or keep it empty for a random song.")]
 		public void PlayMusic(string track = null, LuaFunction func = null)
 		{
-			if (!Game.Settings.Sound.MapMusic || !playlist.IsMusicAvailable)
+			if (!playlist.IsMusicAvailable)
 				return;
 
 			var musicInfo = !string.IsNullOrEmpty(track) ? world.Map.Rules.Music[track]
--- a/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
@@ -11,26 +11,28 @@
 using System;
 using System.Drawing;
 using System.IO;
-using System.Linq;
 using Eluant;
 using OpenRA.Effects;
 using OpenRA.FileFormats;
 using OpenRA.FileSystem;
-using OpenRA.GameRules;
 using OpenRA.Graphics;
 using OpenRA.Mods.Common.Effects;
 using OpenRA.Scripting;
+using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Scripting
 {
 	[ScriptGlobal("Media")]
 	public class MediaGlobal : ScriptGlobal
 	{
-		World world;
+		readonly World world;
+		readonly MusicPlaylist playlist;
+
 		public MediaGlobal(ScriptContext context)
 			: base(context)
 		{
 			world = context.World;
+			playlist = world.WorldActor.Trait<MusicPlaylist>();
 		}
 
 		[Desc("Play an announcer voice listed in notifications.yaml")]
@@ -51,23 +53,15 @@ namespace OpenRA.Mods.Common.Scripting
 			Sound.Play(file);
 		}
 
-		MusicInfo previousMusic;
 		Action onComplete;
 		[Desc("Play track defined in music.yaml or keep it empty for a random song.")]
 		public void PlayMusic(string track = null, LuaFunction func = null)
 		{
-			if (!Game.Settings.Sound.MapMusic)
-				return;
-
-			var music = world.Map.Rules.InstalledMusic.Select(a => a.Value).ToArray();
-			if (!music.Any())
+			if (!Game.Settings.Sound.MapMusic || !playlist.IsMusicAvailable)
 				return;
 
 			var musicInfo = !string.IsNullOrEmpty(track) ? world.Map.Rules.Music[track]
-			: Game.Settings.Sound.Repeat && previousMusic != null ? previousMusic
-			: Game.Settings.Sound.Shuffle ? music.Random(Game.CosmeticRandom)
-			: previousMusic == null ? music.First()
-			: music.SkipWhile(s => s != previousMusic).Skip(1).First();
+			: playlist.GetNextSong();
 
 			if (func != null)
 			{
@@ -84,19 +78,17 @@ namespace OpenRA.Mods.Common.Scripting
 						Context.FatalError(e.Message);
 					}
 				};
+
+				playlist.Play(musicInfo, onComplete);
 			}
 			else
-				onComplete = () => { };
-
-			Sound.PlayMusicThen(musicInfo, onComplete);
-
-			previousMusic = Sound.CurrentMusic;
+				playlist.Play(musicInfo);
 		}
 
 		[Desc("Stop the current song.")]
 		public void StopMusic()
 		{
-			Sound.StopMusic();
+			playlist.Stop();
 		}
 
 		Action onCompleteFullscreen;
--- a/OpenRA.Mods.Common/Widgets/Logic/MusicPlayerLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/MusicPlayerLogic.cs
@@ -18,29 +18,24 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 {
 	public class MusicPlayerLogic
 	{
-		readonly Ruleset modRules;
+		readonly ScrollPanelWidget musicList;
+		readonly ScrollItemWidget itemTemplate;
 
-		bool installed;
+		readonly MusicPlaylist musicPlaylist;
 		MusicInfo currentSong = null;
-		MusicInfo[] music;
-		MusicInfo[] random;
-		ScrollPanelWidget musicList;
-
-		ScrollItemWidget itemTemplate;
 
 		[ObjectCreator.UseCtor]
 		public MusicPlayerLogic(Widget widget, Ruleset modRules, World world, Action onExit)
 		{
-			this.modRules = modRules;
-
 			var panel = widget.Get("MUSIC_PANEL");
 
 			musicList = panel.Get<ScrollPanelWidget>("MUSIC_LIST");
 			itemTemplate = musicList.Get<ScrollItemWidget>("MUSIC_TEMPLATE");
+			musicPlaylist = world.WorldActor.Trait<MusicPlaylist>();
 
 			BuildMusicTable();
 
-			Func<bool> noMusic = () => !installed;
+			Func<bool> noMusic = () => !musicPlaylist.IsMusicAvailable;
 			panel.Get("NO_MUSIC_LABEL").IsVisible = noMusic;
 
 			var playButton = panel.Get<ButtonWidget>("BUTTON_PLAY");
@@ -54,15 +49,15 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			pauseButton.IsVisible = () => Sound.MusicPlaying;
 
 			var stopButton = panel.Get<ButtonWidget>("BUTTON_STOP");
-			stopButton.OnClick = Sound.StopMusic;
+			stopButton.OnClick = () => { musicPlaylist.Stop(); };
 			stopButton.IsDisabled = noMusic;
 
 			var nextButton = panel.Get<ButtonWidget>("BUTTON_NEXT");
-			nextButton.OnClick = () => { currentSong = GetNextSong(); Play(); };
+			nextButton.OnClick = () => { currentSong = musicPlaylist.GetNextSong(); Play(); };
 			nextButton.IsDisabled = noMusic;
 
 			var prevButton = panel.Get<ButtonWidget>("BUTTON_PREV");
-			prevButton.OnClick = () => { currentSong = GetPrevSong(); Play(); };
+			prevButton.OnClick = () => { currentSong = musicPlaylist.GetPrevSong(); Play(); };
 			prevButton.IsDisabled = noMusic;
 
 			var shuffleCheckbox = panel.Get<CheckboxWidget>("SHUFFLE");
@@ -111,11 +106,13 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 		public void BuildMusicTable()
 		{
-			music = modRules.InstalledMusic.Select(a => a.Value).ToArray();
-			random = music.Shuffle(Game.CosmeticRandom).ToArray();
-			currentSong = Sound.CurrentMusic;
+			if (!musicPlaylist.IsMusicAvailable)
+				return;
+
+			var music = musicPlaylist.AvailablePlaylist();
+			currentSong = musicPlaylist.CurrentSong();
 			if (currentSong == null && music.Any())
-				currentSong = Game.Settings.Sound.Shuffle ? random.First() : music.First();
+				currentSong = musicPlaylist.GetNextSong();
 
 			musicList.RemoveChildren();
 			foreach (var s in music)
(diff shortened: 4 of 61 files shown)


Mailaender:
This now always play the first song in the playlist, instead of the one specified. https://github.com/OpenRA/OpenRA/blob/bleed/mods/cnc/maps/shellmap/shellmap.lua#L30

> it has one regression: due to the playlist getting set up before the shellmap's lua scripts, the game plays a maximum of two seconds maybe before the shellmap music at game start

I don't see that as a problem. We want to keep that order.

pchote:
This is going in the wrong direction by promoting world-state leakage into a feature, instead of a bug.

GraionDilach:
@pchote: That sounded as a feature to me. Resolved.

@Mailaender: Such happens only if you don't have a map1.aud (which is **not** in scores.mix) in your game. I guess transit.mix isn't among the files transferred from an online source.

Mailaender:
I have `map1.aud` and checked again that it works on bleed. Now it plays 2 seconds "Act on Instinct" (first in the playlist) until it switches to `map1.aud` in the cnc mod which is still bogus.

Mailaender:
If I disable `Game.Settings.Sound.MapMusic` it still plays, but only "Act on Instinct".

GraionDilach:
That's exactly why I was thinking about a StartingMusic trait. Alright.

OpenRA.Game/MusicPlaylist.cs
```diff
@@ -0,0 +1,105 @@
...
+
+			music = world.Map.Rules.InstalledMusic.Select(a => a.Value).ToArray();
+			if (!installed)
+				return;
+
+			random = music.Shuffle(Game.CosmeticRandom).ToArray();
+
+			currentSong = Game.Settings.Sound.Shuffle ? random.First() : music.First();
+
+			Play();
```
Mailaender:
```csharp
if (Game.Settings.Sound.MapMusic)
	Play();
```

Mailaender: Or not at all. Why does the playlist decide to play music on it's own?

Mailaender:
> That's exactly why I was thinking about a StartingMusic trait.

We removed in favor of Lua. https://github.com/OpenRA/OpenRA/pull/6880

OpenRA.Game/MusicPlaylist.cs
GraionDilach: MapMusic is the ShellMap / Mission Music checkbox in the UI and even the current bleed tracks silenced music. If I want to disable music overall then the slider is the option.

GraionDilach:
Lua should not be used for defining the music at map start due to the speed condition. I cannot solve that without a trait.

pchote:
My thoughts on the music situation:
```
13:49 <+pchote> the way i'd like to see the music work is that the game will automatically play music if there is no overrides
13:49 <+pchote> lua and/or traits then have the ability to do two things:
13:49 <+pchote> lock the music to a specific track (disabling user choice), and set the suggested track
13:50 <+pchote> so we can then have missions start with a specific track before continuing on as normal (e.g. AOI in GDI01)
13:50 <+pchote> and we can also have missions and MP force-play the score music at the end of the game
13:51 <+pchote> we could possibly drop the "suggestion" lua API by having it defined in the map.yaml instead
13:51 <+pchote> like we do for videos
```

GraionDilach:
Rebased. ATM only the win-lose song is unimplemented but I'd do it in a followup PR.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,139 @@
...
+	{
+		public readonly string StartingMusic = null;
+		public readonly bool Loop = false;
+
+		public object Create(ActorInitializer init) { return new MusicPlaylist(init.World, this); }
+	}
+
+	public class MusicPlaylist
+	{
+		readonly MusicPlaylistInfo info;
```
Mailaender: > The private field `OpenRA.Traits.MusicPlaylist.info` is assigned but its value is never used

GraionDilach: Right. It'll be probably used in the followup PRs, but for now let's fix Travis.

GraionDilach:
Was the Requested changes label on this PR stayed for the score music implementation?

Mailaender:
Please remove the deleted shellmap.lua from https://github.com/GraionDilach/OpenRA/blob/music-refactor/OpenRA.sln, too. Otherwise it seems to retain the behavior now. :white_check_mark:

GraionDilach:
Done. Also did the TODO: in the playlist widget and renamed the Loop to StartingMusicLoop to prevent confusion.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Linq;
+using OpenRA.GameRules;
+
+namespace OpenRA.Traits
+{
+	[Desc("Trait for music handling. Attach this to the world actor.")]
+	public class MusicPlaylistInfo : ITraitInfo
+	{
+		public readonly string StartingMusic = null;
+		public readonly bool StartingMusicLoop = false;
```
penev92: Let's go with `LoopStartingMusic`.

GraionDilach: Went with that then.

OpenRA.Mods.Common/Widgets/Logic/MusicPlayerLogic.cs
```diff
@@ -111,11 +106,10 @@ public MusicPlayerLogic(Widget widget, Ruleset modRules, World world, Action onE
 
 		public void BuildMusicTable()
 		{
-			music = modRules.InstalledMusic.Select(a => a.Value).ToArray();
-			random = music.Shuffle(Game.CosmeticRandom).ToArray();
+			var music = musicPlaylist.GetPlaylist;
 			currentSong = Sound.CurrentMusic;
-			if (currentSong == null && music.Any())
-				currentSong = Game.Settings.Sound.Shuffle ? random.First() : music.First();
+			if (currentSong == null && musicPlaylist.GetInstalled)
```
GraionDilach: Oh wait, this is a bug.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,135 @@
...
+		{
+			if (!installed)
+				return null;
+
+			var songs = Game.Settings.Sound.Shuffle ? random : music;
+			return songs.Reverse().SkipWhile(m => m != currentSong)
+				.Skip(1).FirstOrDefault() ?? songs.Reverse().FirstOrDefault();
+		}
+
+		public void Dispose()
```
penev92: I believe you actually want `public void Disposing(Actor self)` from the `INotifyActorDisposing` interface here. Otherwise this never gets called.
Another issue here is that the new world is initialized (and plays its own music) before the old one is disposed of, meaning this here will never do anything, because by the time it is called, the music has already changed to play whatever the new world specified.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,135 @@
...
+				currentSong = world.Map.Rules.Music[info.StartingMusic];
+				repeat = info.LoopStartingMusic;
+			}
+			else
+			{
+				currentSong = Game.Settings.Sound.Shuffle ? random.First() : music.First();
+				repeat = Game.Settings.Sound.Repeat;
+			}
+
+			Play();
```
penev92: May I suggest
```
			// Fade out any previous music before we replace it with new one
			var volume = Sound.MusicVolume;

			while (Sound.MusicVolume > 0.0f)
				Sound.MusicVolume -= 0.0000005f;

			Sound.MusicVolume = volume;
```
before this line.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,135 @@
...
+	{
+		bool installed = false;
+		bool repeat;
+
+		MusicInfo currentSong = null;
+		MusicInfo[] music;
+		MusicInfo[] random;
+
+		public MusicInfo[] GetPlaylist { get { return music; } }
+		public bool GetInstalled { get { return installed; } }
```
penev92: How about simplifying things to
```
		public MusicInfo[] Playlist { get; private set; }
		public bool IsInstalled { get; private set; }
```

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	public class MusicPlaylist
+	{
+		bool installed = false;
+		bool repeat;
+
+		MusicInfo currentSong = null;
+		MusicInfo[] music;
+		MusicInfo[] random;
```
penev92: Can be `readonly`.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,135 @@
...
+
+		public object Create(ActorInitializer init) { return new MusicPlaylist(init.World, this); }
+	}
+
+	public class MusicPlaylist
+	{
+		bool installed = false;
+		bool repeat;
+
+		MusicInfo currentSong = null;
```
penev92: Redundant default value.

OpenRA.Game/Traits/World/MusicPlaylist.cs
GraionDilach: There's only one stream for the music. I'd guess that would do a fade-out within a single tick. Which makes no difference for the ear.

penev92: No, it actually may prolong the tick, but it sounds quite nice. ( I did test it )

penev92:
Depends on #8607 now for proper disposing.

GraionDilach:
Rebased, addressed most issues (excluding fadeout which will be a debate, I feel), removed a pre-#8560 workaround to deal with leaking music (shellmap-to-modchooser).

**Due to #8607 fixing world leaking and this making music world-dependant, we lost loadscreen music.**

I am not certain if I should call that bug or feature, because while the default mods have small loadtimes, nothing stops a mod to grow enough large for a noticable silence.

OpenRA.Game/Traits/World/MusicPlaylist.cs
GraionDilach: Since I don't have to do a check on the music now, I refactored this into a StopMusic call, which I call explicitly in World.Dispose - this StopMusic is also the StopMusic Lua and GUI will use.

penev92: The `World.Dispose()` method already notifies all relevant traits to stop what they are doing and clean up, so I'd rather we go with that instead of hacking a call to `Playlist` inside `World.Dispose()`. (The current `Sound.StopMusic()` in there was a quick hack to patch things up.)

GraionDilach: That would be cool having the fadeout, but the fadeout cannot be in a Dispose... I'd like to evade that discussion for now.

penev92: I'm not talking about the fadeout. I'm talking about `Playlist.Stop();` being out of place inside `World.Dispose()`.
And we even have an interface for that - `INotifyActorDisposing`, or whatever it was called.

penev92: P.S.: That actually removes any references from `World` to `Playlist`, which is a very good thing. Not only it spares us some bogus (from an architectural standpoint) class/trait/whateveryouwant dependencies, it lets people change Playlist with their own trait in their mods, as there is no longer a hardcoded reference.

GraionDilach: Except that World does a Trait check on MusicPlaylist and not TraitOrDefault, or should I change that as well?

GraionDilach: PS.: TBH, after I'm done, I'll see no reason why would someone go with a new trait instead of MusicPlaylist because I might implement all the footwork of a faction-specific dynamic music system - the dynamic part is the one I'm wondering about for now, but whatever.

GraionDilach: PS2.: Added the interface implementation.

OpenRA.Game/World.cs
```diff
@@ -169,6 +169,8 @@ public bool ToggleInputMode<T>() where T : IOrderGenerator, new()
 			}
 		}
 
+		public readonly MusicPlaylist Playlist;
```
pchote: This is only accessed from two places (`MediaGlobal` and `MusicPlayerLogic`).  It would be better if you queried the trait in those two places and removed this reference.

GraionDilach: The World does not implement the self.Trait<T>() interface.

pchote: In the `MediaGlobal` constructor add `playlist = world.WorldActor.Trait<MusicPlaylist>()` and a `readonly MusicPlaylist playlist;` instance variable.  You can do something similar in `MusicPlayerLogic`.

GraionDilach: Thanks, for some reason IntelliSense hid that option at first. Done.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -27,10 +26,12 @@ namespace OpenRA.Mods.Common.Scripting
 	public class MediaGlobal : ScriptGlobal
 	{
 		World world;
+		MusicPlaylist playlist;
```
pchote: This and the world above should be readonly, and then a newline before the constructor please.

OpenRA.Mods.Common/Widgets/Logic/MusicPlayerLogic.cs
```diff
@@ -18,29 +18,24 @@ namespace OpenRA.Mods.Common.Widgets.Logic
...
 	public class MusicPlayerLogic
 	{
-		readonly Ruleset modRules;
-
-		bool installed;
 		MusicInfo currentSong = null;
-		MusicInfo[] music;
-		MusicInfo[] random;
-		ScrollPanelWidget musicList;
+		MusicPlaylist musicPlaylist;
```
pchote: Readonly here and for the two widgets below please.  The general rule is that if it's set in a constructor and isn't supposed to change, then it should be marked readonly.

GraionDilach: Done.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,150 @@
...
+		public object Create(ActorInitializer init) { return new MusicPlaylist(init.World, this); }
+	}
+
+	public class MusicPlaylist : INotifyActorDisposing
+	{
+		readonly MusicInfo[] random;
+		readonly MusicInfo[] playlist;
+
+		MusicInfo currentSong;
+		bool isMusicInstalled;
```
pchote: This can be readonly and moved below `playlist`.

pchote: You can then make this public and remove the `IsInstalled()` method.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,150 @@
...
+	{
+		public readonly string StartingMusic = null;
+		public readonly bool LoopStartingMusic = false;
+
+		public object Create(ActorInitializer init) { return new MusicPlaylist(init.World, this); }
+	}
+
+	public class MusicPlaylist : INotifyActorDisposing
+	{
+		readonly MusicInfo[] random;
```
pchote: These can become public too, and then remove the methods that return them.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,150 @@
...
+				return;
+
+			Sound.PlayMusicThen(currentSong, () =>
+			{
+				if (!isMusicInstalled)
+					return;
+
+				if (!repeat)
+					currentSong = GetNextSong();
+				Play();
```
pchote: nit: newline above here please.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,150 @@
...
+			repeat = Game.Settings.Sound.Repeat;
+
+			Sound.PlayMusicThen(music, () =>
+			{
+				if (!isMusicInstalled)
+					return;
+
+				if (!repeat)
+					currentSong = GetNextSong();
+				Play();
```
pchote: nit: newline above here please.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,150 @@
...
+		{
+			if (!isMusicInstalled)
+				return null;
+
+			var songs = Game.Settings.Sound.Shuffle ? random : playlist;
+			return songs.SkipWhile(m => m != currentSong)
+				.Skip(1).FirstOrDefault() ?? songs.FirstOrDefault();
+		}
+
+		public MusicInfo GetPrevSong()
```
pchote: You could remove a bit of duplication by having a private `GetNextSong(IEnumerable<MusicInfo> tracks)` that gets called with random/playlist from `GetNextSong` and (random/playlist).Reverse() from `GetPrevSong`.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -53,15 +54,11 @@ public void PlayMusic(string track = null, LuaFunction func = null)
...
-			if (!music.Any())
+			if (!playlist.IsInstalled())
 				return;
 
 			var musicInfo = !string.IsNullOrEmpty(track) ? world.Map.Rules.Music[track]
-			: Game.Settings.Sound.Repeat && previousMusic != null ? previousMusic
-			: Game.Settings.Sound.Shuffle ? music.Random(Game.CosmeticRandom)
-			: previousMusic == null ? music.First()
-			: music.SkipWhile(s => s != previousMusic).Skip(1).First();
+			: playlist.GetNextSong();
```
pchote: Please throw an error if someone asks for a bogus track instead of silently falling back to a random song.

mods/cnc/maps/shellmap/map.yaml
```diff
@@ -994,6 +994,9 @@ Rules:
 			Effect: Desaturated
 		LuaScript:
 			Scripts: shellmap.lua
+		MusicPlaylist:
+			StartingMusic: map1
```
pchote: You're leaking state somewhere.  If you start TD without scores.mix it will play map1.  Go Extras &rarr; Music &rarr; Install Music &rarr; Back and it will sometimes play the score screen music.

pchote: Actually, it seems to be ignoring this completely, and picking a random track at startup.

pchote:
Looks good overall, but there are a few style nits and you'll need to fix whatever is stopping `StartingMusic` from working on the TD shellmap.

OpenRA.Game/Traits/World/MusicPlaylist.cs
GraionDilach: Outer classes don't need the random - only the music player GUI uses playlist alone to list the available songs - and I plan to filter the songs in the return method in a followup PR aimed for faction-specific OST.

pchote: Yeah, I noticed that after commenting.  It still applies to `playlist`, though.

mods/cnc/maps/shellmap/map.yaml
GraionDilach: Yeah, https://github.com/OpenRA/OpenRA/pull/8562/files#diff-0612b1f741b8612131b70560bfabc77aR57 is a check I carried over which causes it. I guess it should be removed there.

OpenRA.Game/Traits/World/MusicPlaylist.cs
GraionDilach: Added one above the return parameter.

GraionDilach:
On a second check, noidea what caused the bug in the TD shellmap, but I got it working now. Also addressed the nits.

OpenRA.Game/Traits/World/MusicPlaylist.cs
GraionDilach: I got confused in the syntax, so I just made the two public functions into wrappers of a private one instead.

OpenRA.Game/Traits/World/MusicPlaylist.cs
GraionDilach: Done.

GraionDilach:
Note: This will fix #8033 - I was completely unaware that was reported.

penev92:
Now that #8528 was merged, you'll have to account for a few more missions with the music changes, sorry.

GraionDilach:
No problem, I actually hoped for this result - didn't wanted to get clemty into trouble of a rebase.

GraionDilach:
Rebased. I think no changes are requested at the moment - pchote's changes were already resolved before I went out during the weekend.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,149 @@
...
+				currentSong = world.Map.Rules.Music[info.StartingMusic];
+				repeat = info.LoopStartingMusic;
+			}
+			else
+			{
+				currentSong = Game.Settings.Sound.Shuffle ? random.First() : playlist.First();
+				repeat = Game.Settings.Sound.Repeat;
+			}
+
+			Play();
```
obrakmann: This isn't correct, it completely ignores `Game.Settings.Sounds.MapMusic` (try starting the game with music enabled, then disable the checkbox, then switch mods). You'd need to do it like this:

```csharp
if (!Game.Settings.Sound.MapMusic)
    return;

if (!string.IsNullOrEmpty(info.StartingMusic) && ...)
    ...
else
    ...

Play()
```

pchote: Why don't we just remove MapMusic?  If people don't want music, then they can mute it.

OpenRA.sln
```diff
@@ -66,7 +66,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Red Alert Lua scripts", "Re
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Dune 2000 Lua scripts", "Dune 2000 Lua scripts", "{06B1AE07-DDB0-4287-8700-A8CD9A0E652E}"
 	ProjectSection(SolutionItems) = preProject
-		mods\d2k\maps\shellmap\shellmap.lua = mods\d2k\maps\shellmap\shellmap.lua
```
obrakmann: That looks like an accidental removal.

OpenRA.Game/Traits/World/MusicPlaylist.cs
obrakmann: Not sure if I'd like that. The game would constantly do something without any benefit. That feels unclean.

OpenRA.sln
penev92: Iirc the only thing that script did was play the music, which is now done by the new trait, so the scrip is now redundant.

OpenRA.Game/Traits/World/MusicPlaylist.cs
pchote: The setting was added long ago to work around the type of bugs that this pr is finally fixing properly.  It's mostly redundant now, and fouls up the logic around end-of-game scripted music.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -0,0 +1,149 @@
...
+
+		MusicInfo currentSong;
+		bool repeat;
+
+		public MusicPlaylist(World world, MusicPlaylistInfo info)
+		{
+			IsMusicAvailable = world.Map.Rules.InstalledMusic.Any();
+
+			if (!IsMusicAvailable)
+			{
```
abcdefg30: I think you can drop the braces here.

OpenRA.sln
GraionDilach: This might need a revert anyway due to the scripted lighting pr.

OpenRA.Game/Traits/World/MusicPlaylist.cs
GraionDilach: @obrakmann I fail to see what's wrong with that. I interpreted MapMusic as the allow override of the music playlist. The sole thing which fails is to carry over the active song, but due to music being part of the world, I don't think that's a bug.

OTOH decide if you still want that checkbox afterall.

obrakmann: Really not sure about all this, but I'll defer the decision to you guys, as I clearly have the minority opinion here. I'd probably vote for removing the checkbox then.

GraionDilach: Removing the checkbox sounds like the general agreement - even I was considering that after this. Done.

OpenRA.sln
GraionDilach: Yes, I reverted removing the file from the solution to not get a conflict with #8684.

OpenRA.Game/Traits/World/MusicPlaylist.cs
GraionDilach: Dropped.

GraionDilach:
Also did a bugfix to fall back to available music from the playlist if the music selected in trait is defined in the yamls, but missing from the user PC (TD missions without music pack would not play silence anymore).

obrakmann:
Just got this in d2k:
```
Inner Exception of type `System.ArgumentNullException`: Argument cannot be null.
Parameter name: source
    at System.Linq.Check.Source (System.Object source) [0x00000] in <filename unknown>:0 
  at System.Linq.Enumerable.Any[MusicInfo] (IEnumerable`1 source) [0x00000] in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.Logic.MusicPlayerLogic.BuildMusicTable () [0x0002a] in /home/oliver/devel/openra/git/OpenRA.Mods.Common/Widgets/Logic/MusicPlayerLogic.cs:111 
  at OpenRA.Mods.Common.Widgets.Logic.MusicPlayerLogic..ctor (OpenRA.Widgets.Widget widget, OpenRA.Ruleset modRules, OpenRA.World world, System.Action onExit) [0x0007b] in /home/oliver/devel/openra/git/OpenRA.Mods.Common/Widgets/Logic/MusicPlayerLogic.cs:36 
```

GraionDilach:
Whoops, fixed.

obrakmann:
Thanks. :+1: when the last two (or at least the last) commits are squashed into the first.

GraionDilach:
Squashed the last commit. I wouldn't squash the MapMusic one to make sure that change doesn't get overlooked - it just loosely fits to this PR anyway.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/128f27f87edf1803baa45ae7cb7ff05ab24746dc%5E...128f27f87edf1803baa45ae7cb7ff05ab24746dc)

Issue #2330, opened 2012-07-24
Have SHOK boost tesla coils


chrisforbes:
Allow firing again immediately after zapping from SHOK -- and possibly slight damage boost

obrakmann:
Added in #8563.

PR #8563, opened 2015-06-26
Tesla boost for RA2 mod

matija-hustic:
Functionality needed for the RA2 mod. Also described in #2330

Introduces a new upgrade called `tesla`. Its consumers would be
- `RequiresPower` that would disable itself (thus not necessitating power for proper functioning of the actor) when enough `tesla` upgrades  have been granted
- `Armament` that would increase range, damage and reload speed and modify rendering for a juiced-up tesla coil

Normally tesla coils would use these traits.

The upgrade's provider would be the `TeslaBoost` activity which requires the actor, that wishes to use this activity, to have the `GivesTeslaBoost` trait. Normally, tesla tanks and tesla troopers would use this trait.

I've tried to make things flexible by allowing
- the modders to decide whether or not `GivesTeslaBoost` activates the `TeslaBoost` when the units are idle in the vicinity of a receiving tesla coil
- the player to force the tesla boost by explicitly giving an order to a tesla tank or tesla trooper to target a particular tesla coil
- the `TeslaBoost` activity to be used with mobile units

untitled


Unit158:
:-1: This should be with the RA2 mod.

matija-hustic:
Ah they don't share common code here?

GraionDilach:
The RA2 mod currently does not have it's own DLL repo - @Phrohdoh plans to maintain one on his profile, but tbh I'm not a fan of scattering DLL repos.

Besides that, we'll need the premature upgrade dictionary initialization to fix #8444.

matija-hustic:
I thought that RA2 mod on @Phrohdoh's profile was only there for art copyright reasons, but that all the code resides in this repository.

GraionDilach:
I haven't heard a concrete word on that yet and tbh yes, I'd welcome the DLL code residing in this repository too.

GraionDilach:
http://logs.openra.net/?year=2015&month=06&day=26#16:03:55

pchote:
This has been a long-planned feature for our RA mod, so the code is in the right place.

matija-hustic:
I'll see what I can do with @GraionDilach's suggestions. In the meantime, this should be marked as having changes requested.

OpenRA.Mods.RA/Activities/TeslaBoost.cs
```diff
@@ -0,0 +1,138 @@
...
+
+			// Should the activity be canceled or target moves out of range, cancel the activity
+			if (IsCanceled || (self.CenterPosition - target.CenterPosition).LengthSquared > range.RangeSquared)
+			{
+				if (upgraded)
+				{
+					upgraded = false;
+					if (um != null)
+						foreach (var upgrade in upgrades)
+							for (var u = 0; u < grantMultiple; u++)
```
atlimit8: If `grantMultiple` is to be used as a cap, the it should have a name like `maxGrantCount` instead of one that implies boolean value.

OpenRA.Mods.RA/Traits/GivesTeslaBoost.cs
```diff
@@ -0,0 +1,196 @@
...
+
+namespace OpenRA.Mods.RA.Traits
+{
+	public class GivesTeslaBoostInfo : ITraitInfo
+	{
+		[Desc("The upgrades to grant.")]
+		public readonly string[] Upgrades = { "tesla" };
+
+		[Desc("Make this actor grant the upgrades multiple times.")]
+		public readonly int GrantMultiple = 1;
```
atlimit8: As mentioned above, this looks like it is interpreted as a predicate value.
`"Maximum number of upgrades to grant."`
`MaxGrantCount`

matija-hustic: This guy is here to make Tesla tanks grant `tesla` multiple times because if I remember correctly from RA2, one tank should give more of a boost than a single trooper

atlimit8: Regardless of its existence, its name is misleading.

matija-hustic: I'm open to suggestions ;)

atlimit8:
Since this could be used for supression attacks and possibly Chrono Legionnaire attacks, perhaps `GivesTeslaBoost` should be `SupportsUpgrade` or `SupportUpgradeAttack` and `TeslaBoost` should be `SupportUpgrade` or `SupportUpgradeActivity`.

matija-hustic:
I'll see how much of these even stay when I try to do this using upgrade granting warheads. But `GrantMultiple` functionality will probably have to stay. So a preferred name would be appreciated

atlimit8:
Is there a reason this can not be done with the existing code?

Perhaps:
  * `GrantMultiple` => `GrantCount`
  * `GivesTeslaBoost` => `GrantsUpgradeDuringAttack`
  * `TeslaBoost`  =>`UpgradeDuringAttack`

atlimit8:
People on the IRC channel question, perhaps the reason should by appended to the description.

matija-hustic:
Upgrade granting warheads is existing code, so this would make this pr smaller. Plus it's more logical.

See discussion in

http://logs.openra.net/?year=2015&month=06&day=26#16:03:55

matija-hustic:
I asked GraionDilach to write the reasons here, but we thought a link to the discussion would suffice

atlimit8:
Do you plan to merge it with the current GrantUpgradeWarhead?

matija-hustic:
Don't know yet, didn't have the time to look into it in detail. I'll do my best to do it elegantly, if possible without modifying existing functionality

GraionDilach:
@atlimit8: Chrono Legionnaire should not use this at all. Chrono Legionnaire should have two armaments, with one only used for upgrading the target's TargetType from casual to Temporal only along with DisableUpgrade recoloring the target to light white, a WithDecoration giving the temporal sparkles and the another one only affecting Temporal TargetType'd targets, maintaining this upgrade and delivering the damage entirely. It has no relation to this atm.... the Paratroopers refactor is more related due to the current TargetType-upgrade proposal disallowing to have different TargetType setup for paratroopers and for temporal weapons (and for RA2 IronCurtain because in RA2, AI units didn't even wasted shots for Ironcurtained units, something OpenRA does).

Both the TeslaBoost and GivesTeslaBoost trait can be replaced via an upgradable Armament on the coil with a weapon having GrantUpgrade warheads to get it upgraded on SHOK's side, with both of SHOK's armaments sharing the AmmoPool to prevent SHOK charging both the coil and attacking at the same time. This needs no change in the code at GrantUpgrade, only YAML level modding - Armaments needs a change just because of #8444 alone.

The sole thing which shall stay from the current state of this PR is the Upgradable implementation of RequiresPower. Everything else can be axed and if @matija-hustic does not have time, I gladly fill a PR against his PR tomorrow to do this YAML modding.

OpenRA.Mods.RA/Traits/GivesTeslaBoost.cs
GraionDilach: Tesla Tanks couldn't charge coils in RA2. The original code hardcoded the charger to be an infantry.

matija-hustic: Are you sure about this?

matija-hustic:
Go ahead and take care of this @GraionDilach

OpenRA.Mods.RA/Traits/GivesTeslaBoost.cs
GraionDilach: https://github.com/Ares-Developers/YRpp/blob/master/BuildingClass.h#L230

matija-hustic: Ah ok. I must be getting senile :D

GraionDilach:
Did a PR against your PR. That should solve 90% of the nits everybody can have against the implementation.

mods/ra/rules/infantry.yaml
```diff
@@ -634,7 +635,7 @@ Ant:
 	Armament:
 		Weapon: mandible
 	TargetableUnit:
-		TargetTypes: Ground, Infantry
+		TargetTypes: Ground, Infantry, TeslaBoost
```
GraionDilach: I didn't touched this in my PR - why is this change?

matija-hustic:
It seems that timed upgrades cannot go over level 1. This means that without changing this behavior, the tesla powerup cannot be implemented using upgrade granting warheads...
Apart from that, `AttackBase` only handles `DamageWarheads`. This means that a new trait would still be required to control the tesla powerup `GrantUpgradeWarhead`. It would nevertheless be more elegant with the warheads, but... Maybe see if the timed upgrades could be changed in a separate PR?
In the meantime, what should I do with this PR?

penev92:
I started on a warhead refactoring quest a while back (see 7997), and I still need to do one more PR. That may handle the `GrantUpgradeWarhead` side of the problem (or at least pave the road for it).
The `TimedUpgrade`s can be refactored to have levels (which would most likely want its own PR).

If you are hopelessly stuck on those two points, I would suggest closing this PR until they are resolved.

GraionDilach:
I honestly can't find any reason in the code why this wouldn't work.

matija-hustic:
I've tried doing this thing locally using higher level timed upgrades and `GrantUpgradeWarhead`.

When the upgrade granting warhead's weapon's reload time is less than the upgrade duration, it's not the prettiest of sights when
* you put one tesla trooper next to a coil and then have the coil's range or
* you put two tesla troopers next to a coil without power

tesla1
tesla2

If you put the upgrade granting warhead's weapon's reload and the upgrade durationto be exactly equal, the situation is visually nicer, but risks breaking certain behaviors due to race conditions... One example is ordering the coil to attack ground - for a certain time it does but after a random number of zaps it just cancels the `ChargeAttack` activity because the coil's upgraded armament briefly becomes disabled.

matija-hustic:
Idea: leave how the timed upgrade granting works right now, just make the timed upgrades source dependent (one source can grant at most one upgrade level, if it grants a new one before the old one is revoked, the old one's duration is just refreshed).

GraionDilach:
I think that jumpy behaviour is acceptable tbh - but if you want to add that GrantUpgadeWarhead change, do it as a variable which modders can disable.

matija-hustic:
Depends on #8658.

GraionDilach:
#8658 has been merged.

I also made a new litning SHP which might look better, @matija-hustic, if you're interested, I add it to my PR against yours.
openra-2015-07-07t200145z

matija-hustic:
OK, so here's the version using upgrade granting warheads.

I was unable to take advantage of `AttackBase` and its subclasses since at most one of those per actor is allowed. And this means that it is effectively impossible to use `AttackBase` to implement a powerup weapon with a different cursor from the attack cursor used for other weapons.

This is why I've stuck to `GivesTeslaBoost`.

Also, for more natural behavior (when putting idle troopers next to a coil, they'll power it up, but stop powering up and attack as soon as enemies are present; on the other hand, when issuing an explicit order to power up a coil, the trooper will not stop powering up even if enemies approach), I've also kept `TeslaBoost`.

Maybe somehow modifying `AttackBase` could eliminate the need for these two classes?

matija-hustic:
@GraionDilach, sure, could you please add your artwork so that I can merge it?

penev92:
Well maybe we're at the point where we should have `AttackBase` upgradable as well?
You should open an issue for what you said above about it.

GraionDilach:
The main issue isn't with `AttackBase` but `AttackMedic` and `AutoHeal` duplicating the entire Attack* logic for themselves.

OpenRA.Mods.RA/Traits/GivesTeslaBoost.cs
```diff
@@ -0,0 +1,200 @@
...
+		[Desc("Should the actor boost nearby targets when idling?")]
+		public readonly bool BoostOnIdle = true;
+
+		public object Create(ActorInitializer init) { return new GivesTeslaBoost(init.Self, this); }
+	}
+
+	public class GivesTeslaBoost : INotifyAddedToWorld, INotifyRemovedFromWorld, ITick, IIssueOrder, IResolveOrder, IOrderVoice, INotifyIdle
+	{
+		readonly string orderName = "TeslaBoost";
+		readonly System.Drawing.Color targetLineColor = System.Drawing.Color.BlueViolet;
```
abcdefg30: Why not `using System.Drawing;` and then drop it here?

OpenRA.Mods.RA/Traits/GivesTeslaBoost.cs
```diff
@@ -0,0 +1,200 @@
...
+		Actor preferredReceiver;
+		List<Actor> nearbyReceivers = new List<Actor>();
+
+		public GivesTeslaBoost(Actor self, GivesTeslaBoostInfo info)
+		{
+			this.self = self;
+			this.info = info;
+
+			mobile = self.Trait<Mobile>();
+			armament = self.TraitsImplementing<Armament>().Where(a => a.Info.Name == info.Armament).First();
```
abcdefg30: `self.TraitsImplementing<Armament>().First(a => a.Info.Name == info.Armament);`?

OpenRA.Mods.RA/Activities/TeslaBoost.cs
```diff
@@ -0,0 +1,58 @@
...
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.Activities;
+using OpenRA.GameRules;
+using OpenRA.Mods.Common.Activities;
```
abcdefg30: This using and `OpenRA.GameRules` aren't used.

matija-hustic:
@abcdefg30, thanks for reviewing, but this PR depends on #8768. With it, `TeslaBoost.cs` and `GivesTeslaBoost.cs` will become unnecessary. This PR should be marked as having dependencies...

abcdefg30:
Oh, heh. Added the label.

Unit158:
This will need a rebase

matija-hustic:
So finally #8768 was merged and this PR is updated. I still need to test that everything works well with autotargeting and that nothing has regressed.
Special thanks to @GraionDilach for helping me out a lot!

mods/ra/weapons/other.yaml
```diff
@@ -177,6 +208,22 @@ TTankZap:
...
+	Report: TESLACHARGE.AUD
+	Charges: yes
+	ValidStances: Ally
+	ValidTargets: TeslaBoost
+	Projectile: TeslaZap
+		Image: litningupg
+	Warhead@charge: GrantUpgrade
+		Range: 42
+		Duration: 125
+		Upgrades: tesla, tesla1, tesla2
```
pchote: Why do you have three different upgrade types that all do the same thing?

matija-hustic: Because of timed upgrades again...
Only one levelup is permitted per timed upgrade source. This is OK for shock troopers since they only give one tesla up each, but it is a little problematic for tesla tanks since they need to give three tesla ups (obviously all with the same source).

Simply allowing more than one levelup per timed upgrade source is kinda problematic - take a shock trooper that grants a 75 tick upgrade every 70 ticks. If you allow multiple upgrades per source, then for a brief period of 5 ticks, the shock trooper would give two tesla ups to the tesla coil. If you make the shock trooper grant a 70 tick upgrade every 70 ticks, then the upgrade level can at times actually drop to 0 on the 70th tick (which causes e.g. that the tesla coil loses its target and other weird stuff).

But yeah, this isn't too pretty. Maybe it'd be nicer if tesla tanks gave a different kind of upgrade, that would count as three shock trooper upgrades?

pchote: How difficult would it be to add an `UpgradeLevels:` parameter that we set to 3 for the tesla tank?

matija-hustic: I've made the `UpgradeManager` support requests to grant a timed upgrade multiple times from a single source. The maximum number of times it will grant needs to be supplied by the caller (in this instance `GrantUpgradeWarhead`), otherwise it is assumed to be 1.

What `GrantUpgradeWarhead` then does when it gets `Upgrades == { "tesla", "tesla", "tesla" }`, it requests granting of each of these tesla ups, but notifying that it should be granted up to three times from this particular source.

How does this sound?

pchote:
Code looks good now, and this is feels really well integrated with the game :+1:

The only possible downside is that this may make the shock troopers even more OP, but we can deal with that balance issue during playtesting.

obrakmann:
Looks fine to me. But yeah, we should probably nerf the shock troopers some more, since they've now gained a defensive capability as well. I'd suggest increasing the price and lowering the damage somewhat. But that's for another PR. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/be9347a9f8426d6a597ff31645e7be0dffd5a797)

PR #8567, opened 2015-06-27
Added multiplier modifier traits, removing GainsStatUpgrades, InvulnerabilityUpgrade, & Invulnerable.

atlimit8:
A branch I created about 7 months ago and just rebased.

Adds:
- DamageMultiplier
- FirepowerMultiplier
- InaccuracyMultiplier
- ReloadTimeMultiplier
- SpeedMultiplier
- PowerMultiplier

Removes as redundant:
- Invulnerable
- GainsStatUpgrades
- InvulnerabilityUpgrade

This should remove GainsStatUpgrades bug(s) by simply replacing it.


GraionDilach:
Looks good in the first sight. It'd be nice if Range could get such a modifier as well in the long run. Some InvulnerabilityUpgrade-replacements have two 0s and others have only one, please make that consistent.

mods/cnc/rules/defaults.yaml
```diff
@@ -40,7 +40,16 @@
...
+	FirepowerMultiplier@EXPERIENCE:
+		UpgradeTypes: firepower
+	DamageMultiplier@EXPERIENCE:
+		UpgradeTypes: damage
+	SpeedMultiplier@EXPERIENCE:
+		UpgradeTypes: speed
+	ReloadTimeMultiplier@EXPERIENCE:
+		UpgradeTypes: reload
+	InaccuracyMultiplier@EXPERIENCE:
+		UpgradeTypes: inaccuarcy
```
reaperrr: Typo, should be inaccuracy.

mods/cnc/rules/defaults.yaml
```diff
@@ -98,7 +107,16 @@
...
```
reaperrr: Same here.

mods/cnc/rules/defaults.yaml
```diff
@@ -141,7 +159,16 @@
...
```
reaperrr: Same here.

mods/ts/rules/defaults.yaml
```diff
@@ -209,11 +209,15 @@
...
-		ReloadModifier: 90, 75
+	FirepowerMultiplier:
+		Modifier: 110, 130
+	DamageMultiplier:
+		Modifier: 83, 66
+	SpeedMultiplier:
+		Modifier: 120, 150
+	ReloadTimeMultiplier:
+		Modifier: 90, 75
+	InaccuracyMultiplier:
```
reaperrr: TS does not (yet) modify inaccuracy, so this trait shouldn't be needed yet.

penev92:
I don't usually do this, but since this has the potential to bring a lot of regressions, I have to request that we mark it as depending on #8321.
[Case and point](https://github.com/OpenRA/OpenRA/pull/8567#discussion_r33412482)

OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
```diff
@@ -324,6 +323,13 @@
     <Compile Include="Traits\Parachutable.cs" />
     <Compile Include="Traits\ParaDrop.cs" />
     <Compile Include="Traits\Passenger.cs" />
+    <Compile Include="Traits\Multipliers\UpgradableMultiplierTrait.cs" />
+    <Compile Include="Traits\Multipliers\DamageMultiplier.cs" />
+    <Compile Include="Traits\Multipliers\FirepowerMultiplier.cs" />
+    <Compile Include="Traits\Multipliers\InaccuracyMultiplier.cs" />
+    <Compile Include="Traits\Multipliers\ReloadTimeMultiplier.cs" />
```
reaperrr: Nit: The weapon property is called ```ReloadDelay```, so it would be more consistent if this was called ```ReloadDelayModifier```.

mods/cnc/rules/defaults.yaml
```diff
@@ -240,7 +267,16 @@
...
```
reaperrr: Here.

mods/cnc/rules/defaults.yaml
```diff
@@ -352,7 +388,16 @@
...
```
reaperrr: Here.

mods/cnc/rules/defaults.yaml
```diff
@@ -399,6 +443,16 @@
...
```
reaperrr: Here.

mods/d2k/rules/defaults.yaml
```diff
@@ -38,7 +38,16 @@
...
+	FirepowerMultiplier@EXPERIENCE:
+		UpgradeTypes: firepower
+	DamageMultiplier@EXPERIENCE:
+		UpgradeTypes: damage
+	SpeedMultiplier@EXPERIENCE:
+		UpgradeTypes: speed
+	ReloadTimeMultiplier@EXPERIENCE:
+		UpgradeTypes: reload
+	InaccuracyMultiplier@EXPERIENCE:
+		UpgradeTypes: inaccuarcy
```
reaperrr: Typo in this file, too.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1016,6 +1016,111 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 					}
 				}
 
+				if (engineVersion < 20150626)
```
penev92: This is going to need (several) comments about what's going on.

OpenRA.Mods.Common/Traits/Multipliers/UpgradableMultiplierTrait.cs
```diff
@@ -0,0 +1,75 @@
...
+		// Allow subclasses to specify defaults.
+		protected UpgradeMultiplierTraitInfo(int[] modifiers, string[] upgradeTypes = null, int baseLevel = 1)
+		{
+			UpgradeTypes = upgradeTypes ?? new string[0];
+			Modifier = modifiers;
+			BaseLevel = UpgradeTypes.Length > 0 ? baseLevel : 0;
+		}
+	}
+
+	public abstract class UpgradeMultiplierTrait : IUpgradable, IDisabledTrait, ISync
```
penev92: Is there a reason why you are reinventing `UpgradableTrait` ?

OpenRA.Mods.Common/Traits/Multipliers/UpgradableMultiplierTrait.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public abstract class UpgradeMultiplierTraitInfo
+	{
+		[Desc("Accepted upgrade types.")]
+		public readonly string[] UpgradeTypes = { };
```
Mailaender: This needs a `[UpgradeGrantedReference]` I guess since we have https://github.com/OpenRA/OpenRA/pull/8321/ now.

OpenRA.Mods.Common/Traits/Multipliers/UpgradableMultiplierTrait.cs
atlimit8: Basing `UpgradableMultiplierTrait` would require explicit setting of `UpgradeMaxAcceptedLevel = UpgradeMinEnabledLevel + Modifier.Length`, making mod authors do unnessary counting and adding should be unacceptable and the changes needed, otherwise, to make `UpgradableMultiplierTrait` fit onto `UpgradableTrait` would clutter the code with special condition(s) to override `UpgradableTrait` such as allowing special values and virtual hooks to override the base behavior (to add that calculation dynamically) while almost all subclasses would have no need for the functionality added (`PowerMultiplier` requires notice to call `power.UpdateActor(self)` when the level changes).

It might be worthwhile to add an abstract base class with the common subset, possibly renaming `UpgradableTrait` to better describe its slightly more specialized functionality (of explicit capping).


`UpgradableMultiplierTrait` might be the reason why I first made `IUpgradable`.

OpenRA.Mods.Common/Traits/Multipliers/UpgradableMultiplierTrait.cs
penev92: Wait, isn't it `[UpgradeUsedReference]` ?

OpenRA.Mods.Common/Traits/Multipliers/DamageMultiplier.cs
```diff
@@ -0,0 +1,34 @@
...
+
+		public object Create(ActorInitializer init) { return new DamageMultiplier(this); }
+	}
+
+	public class DamageMultiplier : UpgradeMultiplierTrait, IDamageModifier
+	{
+		public DamageMultiplier(DamageMultiplierInfo info)
+			: base(info) { }
+
+		public int GetDamageModifier(Actor attacker, DamageWarhead warhead) { return GetModifier(); }
```
WolfGaming: is there any special reason as to why you have all of these getters that call another getter?
why not just replace all of them with GetModifier()?

atlimit8: Those are for Implementing interfaces that may be implemented elsewhere as well.

PR #8572, opened 2015-06-27
Added multiplier modifier traits, removing GainsStatUpgrades, InvulnerabilityUpgrade, & Invulnerable.

atlimit8:
A branch I created about 7 months ago and just rebased.

Adds:
- DamageMultiplier
- FirepowerMultiplier
- InaccuracyMultiplier
- ReloadDelayMultiplier
- SpeedMultiplier
- PowerMultiplier

Removes as redundant:
- Invulnerable
- GainsStatUpgrades
- InvulnerabilityUpgrade

This should remove GainsStatUpgrades bug(s) by simply replacing it.

(Hopefully this is the end of the PR madness.)


Mailaender:
Looking good at first sight, but needs an upgrade rule.

GraionDilach:
https://github.com/OpenRA/OpenRA/pull/8572/files#diff-58e597d337be43c2ec48a0334d6bd447R1209 is an upgrade rule.

atlimit8:
I even tested the upgrade rules using `git checkout HEAD~1 ...`

abcdefg30:
Needs a rebase.

GraionDilach:
Looks solid to me. :+1:/ :white_check_mark:

OpenRA.Mods.Common/Traits/Multipliers/DamageMultiplier.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Damage taken by this actor is multiplied based on upgrade level.",
+		"Decrease to increase actor's apparent strength.",
+		"Use 0 to make actor invulnerable.")]
+	public class DamageMultiplierInfo : UpgradeMultiplierTraitInfo, ITraitInfo
+	{
+		public DamageMultiplierInfo()
+			: base(new int[] { 91, 87, 83, 65 }) { }
```
Phrohdoh: How are these values to be changed in yaml?

Phrohdoh:
Needs a rebase.

OpenRA.Mods.Common/Traits/Multipliers/DamageMultiplier.cs
GraionDilach:
```
DamageMultiplier@IAmZeExample:
UpgradeTypes: example
UpgradeMinEnabledLevel: 1 
Modifier: 85, 66, 45, 22
```

penev92: Oh god please move those to the YAMLs.

Phrohdoh: Okay, I see now. Thanks.

Not a fan of passing values straight to the base's ctor, but I don't see any other way.

pchote: I agree with @penev92 here - there's no reason why these traits can't default to 100 (no modifier), and then let mods define their own curves.

atlimit8: A default of 0 would be better since 100 acts as no modifier which is the same as disabling the trait and cases such as invulnerability use 0 as the modifier.
That would reduce:
```
	DamageMultiplier@INVULNERABLE:
		Modifier: 0
```
to
```
	DamageMultiplier@INVULNERABLE:
```

penev92: 100% of the original value sounds like a perfectly reasonable default to me.

atlimit8: 0 causes different behaviour from being disabled which could help debug mods if mistakes are made by making the error (missing `Modifier` field) apparent during game-play, while 100 would require noticing that the multiplier is present and its `Modifier` field is missing.

In short, 0 makes a mistake more apparent to anyone who can play the mod with a distinct error.

Phrohdoh: 100% is proper, IMO, for the stated reasons.

penev92: I actually wouldn't object to setting the default to -1 and throwing in the constructor if it's not overridden in YAML. That would ensure people override the default, and not having random default values that are used instead of being overridden in half of the places makes potential future upgrade rules much easier.

atlimit8: Actually I'm using an empty array now.

atlimit8:
Now it throws when actor is created if the list of modifiers is empty.

atlimit8:
This would handle the `Modifier` property better if #8703 is merged first by using `[FieldLoader.Require]`.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/4b081ee9775085e9f34c9d5d55afe038ccf6f44f...480db3023ca28b1fda6b81b60c490634798431e0)

Mailaender:
This regressed badly. https://github.com/OpenRA/OpenRA/issues/8751

PR #8575, opened 2015-06-27
Make RequiresPower upgradable

atlimit8:
While this could be used for Tesla boosting, it could have over uses such as adding an internal (backup) generator.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Power/RequiresPower.cs
+++ b/OpenRA.Mods.Common/Traits/Power/RequiresPower.cs
@@ -13,23 +13,24 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Needs power to operate.")]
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Power/RequiresPower.cs
```diff
@@ -13,23 +13,24 @@
...
+		public RequiresPower(Actor self, RequiresPowerInfo info)
+			: base(info)
 		{
 			playerPower = self.Owner.PlayerActor.Trait<PowerManager>();
 		}
 
 		public bool Disabled
 		{
-			get { return playerPower.PowerProvided < playerPower.PowerDrained; }
+			get { return playerPower.PowerProvided < playerPower.PowerDrained && !IsTraitDisabled; }
```
penev92: This seems wrong.

PR #8577, opened 2015-06-27
Replace INotifyCreated with a priority queue & block upgrades until manager is notified

### LINKED ISSUE #8444: Elite weapon(s) in TS broken

atlimit8:
This replaces INotifyCreated with a priority queue and gives UpgradeManager a much more immediate priority.

Should be what was requested in #8444.
Assuming TimedUpgradeBar is the problem, this fixes #8444.

### DIFF
--- a/mods/cnc/maps/nod05/nod05.lua
+++ b/mods/cnc/maps/nod05/nod05.lua
@@ -37,12 +37,12 @@ Gdi12Waypoints = { waypoint0, waypoint1, waypoint3, waypoint11, waypoint12 }
 
 AllWaypoints = { Gdi1Waypoints, Gdi2Waypoints, Gdi3Waypoints, Gdi5Waypoints, Gdi11Waypoints, Gdi12Waypoints }
 
-HuntActorTriggerActivator = { Tower1, Tower2, Radar, Silo1, Silo2, Silo3, Refinery, Barracks, Plant1, Plant2, Yard, Factory }
+PrimaryTargets = { Tower1, Tower2, Radar, Silo1, Silo2, Silo3, Refinery, Barracks, Plant1, Plant2, Yard, Factory }
 
 GDIStartUnits = { }
 
 SendGDIAirstrike = function()
-	if not Radar.IsDead then
+	if not Radar.IsDead and Radar.Owner == GDI then
 		local target = getAirstrikeTarget()
 
 		if target then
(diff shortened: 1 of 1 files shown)


GraionDilach:
Something like that, as I seen in this diff. https://github.com/OpenRA/OpenRA/compare/bleed...matija-hustic:upgradable_construction_order

atlimit8:
That might cause problems if the actor is constructed without entering the world; passengers for example.

GraionDilach:
Well, I don't take a side in this particular PR. This is something you, @matija-hustic  and someone with a lot more experience about trait-load-orders should discuss.

atlimit8:
Is that in a PR?

penev92:
You can simplify (and beautify) this by having an `IsInitialized` field on `UpgradeManager`, that is set by `INotifyCreated`.
However, the whole idea that we need such a workaround doesn't sit well with me.

matija-hustic:
I had a rough idea of removing `IUpgradable` in favor of using only `UpgradableTrait` seeing how basically all upgrades can be derived just from that one trait (the only exception being `GainsStatUpgrades` and its descendants now with #8572). If this were the case, the construction order might be able to be resolved by simply using `Requires<UpgradableTraitInfo>`. Haven't looked into this more deeply tho. Just an idea

atlimit8:
UpgradeManager may not be the first trait to recieve `INotifyCreated.Create(this)`; I considered using it.
However, It must unlock on (or before) the first call after all of the traits are created even if that is *another_trait*`.Created(this)`.
The only other solution that comes to mind is integrating UpgradeManager into Actor, which is probably undesirable.

atlimit8:
Well, another interface could be constructed, but that would be introducing degrees of Created.

atlimit8:
I suppose a `INotifyTraitManagerOfTraitsCreaeted` that is implemented only by trait managers and those managers are careful about accessing each other.

pchote:
This looks super overcomplicated. Just run the offending code in UpgradeManager from a FrameEndTask instead.

PR #8579, opened 2015-06-28
Added new map El Dorado for Tiberian Dawn

Mailaender:
It is using the JUNGLE tileset which we have none shipped so far and the gameplay promises to focus a lot on quick expansion. I also really like the details like the crashed A10, church and the creeps defending the middle.


GraionDilach:
tdjunglemap

The center blossom trees favor bottom. Both top trees should be placed a cell left. I'm also uncertain about the bottom defender Guard Tower's placement, the top can defend both Turrets better.

That's all my nitpicks, I can see why you like this map.

GraionDilach:
Why did you deleted this? I was about to suggest that I'd do the edits myself.

Mailaender:
You can get it from http://resource.openra.net/maps/2907/

PR #8581, opened 2015-06-28
Enable the EMP overlay animation in TS.

GraionDilach:
NB: This must use `WithDecoration` because `WithIdleOverlay` forces the Image be taken from the source actor, requiring it specified on every sequence - this is a more maintainable solution.

This is mostly aimed to @pchote after the discussion last night.

### DIFF
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -357,6 +357,16 @@
 	TimedUpgradeBar@EMPDISABLE:
 		Upgrade: empdisable
 		Color: 255,255,255
+	WithDecoration@EMPDISABLE:
+		Image: emp
+		Sequence: idle
+		ReferencePoint: HCenter, VCenter
+		Palette: effect
+		UpgradeTypes: empdisable
+		ShowToEnemies: true
+		ZOffset: 512
+		UpgradeMinEnabledLevel: 1
+		UpgradeMaxAcceptedLevel: 2
 	Cloak@CLOAKGENERATOR:
 		UpgradeTypes: cloakgenerator
 		UpgradeMinEnabledLevel: 1
--- a/OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
@@ -143,7 +143,10 @@ namespace OpenRA.Mods.Common.Traits
 			}
 
 			pxPos += info.Offset;
+
+			// HACK: Because WorldRenderer.Position() does not care about terrain height at the location
 			var renderPos = wr.Position(pxPos);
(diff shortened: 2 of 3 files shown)


Mailaender:
:+1: It is a bit hard to spot on dark SNOW maps.

GraionDilach:
I could make it better via BlendMode: Additive (also applied that to the EMP Cannon projectile) but something implies me that the ZOffset issues I have with this on Hot Springs are related to the aircrafts' height issue.

obrakmann:
I guess it needs more ZOffset. It works on some units, but on the Mammoth Mk II, for example, it is below the actor:

screenshot from 2015-06-29 21 17 58

GraionDilach:
Had to increase the ZOffset to 8192(!) to get it working on what I guess was the highest elevation of River Raid. It's a joke, but it does the trick for now.

mods/ts/rules/defaults.yaml
```diff
@@ -357,6 +357,16 @@
...
 		Upgrade: empdisable
 		Color: 255,255,255
+	WithDecoration@EMPDISABLE:
+		Image: emp
+		Sequence: idle
+		ReferencePoint: HCenter, VCenter
+		Palette: effect
+		UpgradeTypes: empdisable
+		ShowToEnemies: true
+		ZOffset: 8192
```
pchote: This is rather bogus.  Can we fix it properly, or please add a # TODO: FIXME comment.

GraionDilach: Considering that I know why it is bogus (WithDecoration does not check for the actor's current ZOffset), I am working on a fix already.

GraionDilach:
I got stuck. WithDecoration is bogus when it comes to zoffsets, and I cannot figure out a solution. Someone please decide the fate of this - this is pretty much a minor thing which can return later anyway.

mods/ts/rules/defaults.yaml
Mailaender: Then just remove the workaround so we don't hide the problem.

penev92:
Just change `WithDecoration.cs`, line 150:
```
yield return new SpriteRenderable(img, renderPos, WVec.Zero, info.ZOffset, wr.Palette(info.Palette), info.Scale, true);
```
to
```
yield return new SpriteRenderable(img, renderPos, WVec.Zero, self.CenterPosition.Z + info.ZOffset, wr.Palette(info.Palette), info.Scale, true);
```
(add `self.CenterPosition.Z + ` to the ZOffset parameter)
and you should be fine. Then you can fix the ZOffset on the sequence to something reasonable.

GraionDilach:
Fixed Travis.

penev92:
Works on all elevations now :+1:

PR #8585, opened 2015-06-28
Modified the text inside "Install Required" dialogs to be consistent across all mods

### LINKED ISSUE #7389: TD "Install Game Content" dialog incorrectly states music cannot be downloaded
playtest: 20150118 / win32

TD "Install Game Content" dialog incorrectly states music cannot be downloaded but has to be installed from original media when it has been possible to download it easily from within openra for some time now

edit: I realized after submitting that the text means that music will not be downloaded using this dialog, however it may be worthwhile to point out the music can be easily downloaded later

tdmusic

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8075953-td-install-game-content-dialog-incorrectly-states-music-cannot-be-downloaded?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

suvjunmd:
Fixes #7389.
### Install Required

Before:
td install before
d2k install before
After:
td install after
d2k install after
ra install
ts install
### Installing from CD

Before:
td cd before
d2k cd before
ra cd before
ts cd before
After:
td cd after
d2k cd after
ra cd after
ts cd after
### Downloading Content

Before:
td download before
After:
td download after
d2k download
ra download
ts donwload

### DIFF
--- a/mods/cnc/chrome/install.yaml
+++ b/mods/cnc/chrome/install.yaml
@@ -37,7 +37,7 @@ Container@INSTALL_PANEL:
 			Font: BigBold
 			Contrast: true
 			Align: Center
-			Text: Install Game Content
+			Text: Install Required
 		Background@bg:
 			Width: 640
 			Height: 150
@@ -50,10 +50,10 @@ Container@INSTALL_PANEL:
 					ImageName: install
 				Label@INFO:
 					X: 170
-					Y: 35
+					Y: 50
 					Width: PARENT_RIGHT-30
 					Height: 25
-					Text: OpenRA requires the original C&C content files.
+					Text: OpenRA requires the original Command & Conquer game content.
 					Font: Bold
 				Label@INFO2:
 					X: 170
@@ -61,7 +61,7 @@ Container@INSTALL_PANEL:
 					Width: PARENT_RIGHT-185
 					Height: 25
 					WordWrap: true
-					Text: OpenRA can download these files (excluding music and videos) from the internet, or you can install from the original C&C CDs.
+					Text: Content can be downloaded, or copied from the install CD.
 					Font: Bold
 		Button@BACK_BUTTON:
 			Y: 149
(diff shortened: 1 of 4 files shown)


penev92:
Hm. RA, D2k and TS all share the same UI (for the most part), so they are pretty much identical.
TD however has it's own UI with a quite different style (absolutely intended), and that is perfectly fine. You can find such discrepencies in every aspect of TD's UI.

While I'm all for consistency, this is intentional, and it was done to set the mods apart. Might be even better to have unique styles for D2k and TS (which now inherit RA's UI), again to highlight that they are their own mods and not just copies of RA.

pchote:
Note that TD follows the convention that buttons that take you "forward" are positioned on the right, and back are on the left. Changing the installer button layout therefore makes things *less* consistent.

suvjunmd:
I will return the buttons to their original positions. What do you think about the modified text ?

penev92:
I'm OK with the text, but I'm pretty sure Dune 2000 had only one CD.

suvjunmd:
Updated

penev92:
Looks good :+1: 
But since we're polishing stuff, you might want to move the header labels on the TD dialog a few pixels lower, because these screens show them way too high.

suvjunmd:
How about this ?

obrakmann:
Looks fine. :+1:

GraionDilach:
This was already your second contribution. You should add yourself to AUTHORS.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/a722a2508f9f871aaab457e26bbf9f70684635ca%5E...a722a2508f9f871aaab457e26bbf9f70684635ca)

suvjunmd:
> This was already your second contribution. You should add yourself to AUTHORS.

Thanks for letting me know :wink:

PR #8586, opened 2015-06-28
UpgradeManager after IUpgradables

### LINKED ISSUE #8444: Elite weapon(s) in TS broken

atlimit8:
Fixes #8444 as a consequence.

### DIFF
--- a/OpenRA.Mods.Cnc/Traits/PoisonedByTiberium.cs
+++ b/OpenRA.Mods.Cnc/Traits/PoisonedByTiberium.cs
@@ -15,12 +15,12 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Cnc.Traits
 {
-	class PoisonedByTiberiumInfo : UpgradableTraitInfo, ITraitInfo
+	class PoisonedByTiberiumInfo : UpgradableTraitInfo
 	{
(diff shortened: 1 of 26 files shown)


atlimit8:
Someone should check to see if @matija-hustic is correct in #8444 by checking it against this.

GraionDilach:
My elite Armaments work wonderfully with this.

penev92:
Will this wait for *all* traits that implement `IUpgradableInfo` to be initialized before it initializes `UpgradeManager`, or would one of those suffice?

atlimit8:
`ActorInfo.TraitsInConstructOrder` sorts required traits before, so it should unless their is a bug.

Issue #3318, opened 2013-05-19
The 'Jump to base' hotkey is useless if the Construction Yard is destroyed


x-a-n-a-x:
In some cases, there is a small problem with the "Jump to base" hotkey : 

If you lose your main Construction Yard, you can't use this hotkey anymore...
It would be nice to go back to the barracks, war factory or outpost ?

bryanwilbur:
I submitted a fix which works like the original Red Alert.  The hot key will jump to the oldest non-destroyed building.

PR #8600, opened 2015-06-29
The 'Jump to base' hotkey is useless if the Construction Yard is dest…

bryanwilbur:
…royed #3318

Now works like original Red Alert.  When there are no more Construction Yards the 'Jump to base' hotkey goes to the oldest undestroyed building.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Crates/Crate.cs
+++ b/OpenRA.Mods.Common/Traits/Crates/Crate.cs
@@ -65,9 +65,6 @@ namespace OpenRA.Mods.Common.Traits
 
 			var crateActions = self.TraitsImplementing<CrateAction>();
 
-			self.Dispose();
-			collected = true;
-
 			if (crateActions.Any())
 			{
 				var shares = crateActions.Select(a => Pair.New(a, a.GetSelectionSharesOuter(crusher)));
@@ -79,7 +76,11 @@ namespace OpenRA.Mods.Common.Traits
 				{
 					if (n < s.Second)
 					{
+						if (!crusher.HasTrait<Armament>() && s.First.GetEffectName().Equals("firepower"))
+							return;
 						s.First.Activate(crusher);
+						self.Dispose();
+						collected = true;
 						return;
 					}
 					else
--- a/OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
@@ -191,28 +191,24 @@ namespace OpenRA.Mods.Common.Widgets
 				.Select(b => b.Actor)
 				.ToList();
 
-            var noBase = false;
-
-            // If no base found the nondestroyed buildings are selected.
+            // If no base is found all existing buildings are selected.
             if (!bases.Any())
             {
-                bases = world.ActorsWithTrait<TargetableBuilding>()
+				bases = world.ActorsWithTrait<TargetableBuilding>()
                 .Where(a => a.Actor.Owner == world.LocalPlayer)
                 .Select(b => b.Actor)
                 .ToList();
-                noBase = true;
-            }
-
-            if (!bases.Any())
-                return true;
+                if (!bases.Any())
+					return true;
+            }   
 
 			var next = bases
 				.SkipWhile(b => !world.Selection.Actors.Contains(b))
 				.Skip(1)
 				.FirstOrDefault();
 
-            // If no construction yard exists then next is set to only the oldest non-destroyed building
-			if (next == null || noBase)
+            // If no construction yard exists then next is set to only the oldest existing building
+			if (next == null || !next.HasTrait<BaseBuilding>())
 				next = bases.First();
 
 			world.Selection.Combine(world, new Actor[] { next }, false, true);
--- a/OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
@@ -191,15 +191,27 @@ namespace OpenRA.Mods.Common.Widgets
 				.Select(b => b.Actor)
 				.ToList();
 
-			if (!bases.Any())
-				return true;
+            var noBase = false;
+
+            // If no base found the first nondestroyed building is selected.
+            if (!bases.Any())
+            {
+                bases = world.ActorsWithTrait<TargetableBuilding>()
+                .Where(a => a.Actor.Owner == world.LocalPlayer)
+                .Select(b => b.Actor)
+                .ToList();
+                noBase = true;
+            }
+
+            if (!bases.Any())
+                return true;
 
 			var next = bases
 				.SkipWhile(b => !world.Selection.Actors.Contains(b))
 				.Skip(1)
 				.FirstOrDefault();
 
-			if (next == null)
+			if (next == null || noBase)
 				next = bases.First();
 
 			world.Selection.Combine(world, new Actor[] { next }, false, true);
--- a/OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
@@ -191,23 +191,23 @@ namespace OpenRA.Mods.Common.Widgets
 				.Select(b => b.Actor)
 				.ToList();
 
-            // If no base is found all existing buildings are selected.
-            if (!bases.Any())
-            {
+			// If no base is found all existing buildings are selected.
+			if (!bases.Any())
+			{
 				bases = world.ActorsWithTrait<TargetableBuilding>()
-                .Where(a => a.Actor.Owner == world.LocalPlayer)
-                .Select(b => b.Actor)
-                .ToList();
-                if (!bases.Any())
+				.Where(a => a.Actor.Owner == world.LocalPlayer)
+				.Select(b => b.Actor)
+				.ToList();
+				if (!bases.Any())
 					return true;
-            }   
+			}
 
 			var next = bases
 				.SkipWhile(b => !world.Selection.Actors.Contains(b))
 				.Skip(1)
 				.FirstOrDefault();
 
-            // If no construction yard exists then next is set to only the oldest existing building
+			// If no construction yard exists then next is set to only the oldest existing building
 			if (next == null || !next.HasTrait<BaseBuilding>())
 				next = bases.First();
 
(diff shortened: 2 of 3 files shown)


OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
```diff
@@ -191,15 +191,27 @@ bool CycleBases()
...
+            if (!bases.Any())
+                return true;
 
 			var next = bases
 				.SkipWhile(b => !world.Selection.Actors.Contains(b))
 				.Skip(1)
 				.FirstOrDefault();
 
-			if (next == null)
+			if (next == null || noBase)
```
GraionDilach: Why is this change?

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
```diff
@@ -191,15 +191,27 @@ bool CycleBases()
 				.Select(b => b.Actor)
 				.ToList();
 
-			if (!bases.Any())
-				return true;
+            var noBase = false;
```
GraionDilach: I think you don't even need this boolean at all. Merely the function below is enough.

GraionDilach:
Also you apparently used wrong line endings in the commit. If you're using Visual Studio, you can either fix the file by using File->Advanced Save Options... and select Unix line endings there or use the Strip'em plugin: http://www.grebulon.com/software/stripem.php

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
```diff
@@ -191,15 +191,28 @@ bool CycleBases()
...
+            if (!bases.Any())
+            {
+                bases = world.ActorsWithTrait<TargetableBuilding>()
+                .Where(a => a.Actor.Owner == world.LocalPlayer)
+                .Select(b => b.Actor)
+                .ToList();
+                noBase = true;
+            }
+
+            if (!bases.Any())
```
WolfGaming: You can move this if inside the other if so you don't check it twice if its false in both cases.

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
WolfGaming: indeed the noBase variable shouldnt be needed.

GraionDilach:
That and the snappet uses spaces for indenting instead of tabs. See https://msdn.microsoft.com/en-us/library/ms165330%28v=vs.90%29.aspx how to change that.

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
```diff
@@ -191,15 +191,28 @@ bool CycleBases()
...
 				.ToList();
 
-			if (!bases.Any())
-				return true;
+            var noBase = false;
+
+            // If no base found the nondestroyed buildings are selected.
+            if (!bases.Any())
+            {
+                bases = world.ActorsWithTrait<TargetableBuilding>()
```
penev92: Is there a particular reason this uses `TargetableBuilding`?
That makes me uncomfortable for some reason. Can you use `Building`?

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
```diff
@@ -191,15 +191,28 @@ bool CycleBases()
 				.Select(b => b.Actor)
 				.ToList();
 
-			if (!bases.Any())
-				return true;
+            var noBase = false;
+
+            // If no base found the nondestroyed buildings are selected.
```
penev92: "nondestroyed" -> "existing"
"If no base found" -> "If no base is found,"
aaand probably "the" -> "all" :D

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
WolfGaming: Wouldn't it be weirder if this command allowed you to select buildings that aren't targetable?

penev92: We're talking about player-owned buildings here, so I don't see a case where you'd own something that is not targetable.

WolfGaming: We don't want to introduce edge cases here if a custom map assigns a non selectable building to a player.

bryanwilbur:
Updated so boolean variable is removed and second if statement moved inside the first one.

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
```diff
@@ -191,15 +191,24 @@ bool CycleBases()
...
+            }   
 
 			var next = bases
 				.SkipWhile(b => !world.Selection.Actors.Contains(b))
 				.Skip(1)
 				.FirstOrDefault();
 
-			if (next == null)
+            // If no construction yard exists then next is set to only the oldest existing building
+			if (next == null || !next.HasTrait<BaseBuilding>())
```
GraionDilach: Why is this? From what I see, this change isn't needed. bases will contain all buildings if no conyards and only conyards when conyards, since you reused the variable above. This check was fine as-is.

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
```diff
@@ -191,15 +191,24 @@ bool CycleBases()
...
+            // If no base is found all existing buildings are selected.
+            if (!bases.Any())
+            {
+				bases = world.ActorsWithTrait<TargetableBuilding>()
+                .Where(a => a.Actor.Owner == world.LocalPlayer)
+                .Select(b => b.Actor)
+                .ToList();
+                if (!bases.Any())
+					return true;
+            }
```
GraionDilach: > OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs:L203: [SP2000] Invalid spacing at the end of the line.

You should run make check to notice issues like having additional whitespace at the end of a line.

penev92: Just in case you didn't understand that - OpenRA's Utility has a `make check` function, that checks the code for style errors and warns you.
It is what our CI system (Travis) is using, and this is why it's marked the build as failing below.

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
```diff
@@ -191,15 +191,24 @@ bool CycleBases()
...
 
-			if (!bases.Any())
-				return true;
+            // If no base is found all existing buildings are selected.
+            if (!bases.Any())
+            {
+				bases = world.ActorsWithTrait<TargetableBuilding>()
+                .Where(a => a.Actor.Owner == world.LocalPlayer)
+                .Select(b => b.Actor)
+                .ToList();
```
WolfGaming: You seem to be using spaces here instead of tabs.

pchote:
@bryanwilbur please join us in our IRC channel (http://www.openra.net/community/ - "Getting in Touch") if you'd like some help sorting out the indentation and git issues so that we can merge this.

bryanwilbur:
I used make check and the tab issue and the extra spaces should be resolved.

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
```diff
@@ -191,15 +191,24 @@ bool CycleBases()
...
 				.ToList();
 
+			// If no base is found all existing buildings are selected.
 			if (!bases.Any())
-				return true;
+			{
+				bases = world.ActorsWithTrait<TargetableBuilding>()
+				.Where(a => a.Actor.Owner == world.LocalPlayer)
+				.Select(b => b.Actor)
+				.ToList();
```
penev92: Please add one more level of indentation on the last 3 lines, and then an empty line after this.

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
```diff
@@ -191,15 +191,24 @@ bool CycleBases()
...
+			}
 
 			var next = bases
 				.SkipWhile(b => !world.Selection.Actors.Contains(b))
 				.Skip(1)
 				.FirstOrDefault();
 
-			if (next == null)
+			// If no construction yard exists then next is set to only the oldest existing building
+			if (next == null || !next.HasTrait<BaseBuilding>())
```
penev92: What is the purpose of this additional check for `BaseBuilding`?

penev92:
Forks :+1:

OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs
bryanwilbur: Without that check using the hotkey will cycle through all buildings.  The check results in going to the oldest existing building which is more inline with how Red Alert handles returning to base without a construction yard.

---- On Sat, 04 Jul 2015 15:01:39 -0600 Pavel Penev &lt;notifications@github.com&gt; wrote ---- 

In OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs:
 &gt; &gt; var next = bases &gt; .SkipWhile(b =&gt; !world.Selection.Actors.Contains(b)) &gt; .Skip(1) &gt; .FirstOrDefault(); &gt; &gt; - if (next == null) &gt; + // If no construction yard exists then next is set to only the oldest existing building &gt; + if (next == null || !next.HasTrait&lt;BaseBuilding&gt;())  What is the purpose of this additional check for BaseBuilding?
 —
Reply to this email directly or view it on GitHub.
    
 
 




​

OpenRA.Mods.Common/Traits/Crates/Crate.cs
```diff
@@ -76,10 +73,14 @@ public void OnCrush(Actor crusher)
 				var n = self.World.SharedRandom.Next(totalShares);
 
 				foreach (var s in shares)
-				{
+				{
 					if (n < s.Second)
 					{
+						if (!crusher.HasTrait<Armament>() && s.First.GetEffectName().Equals("firepower"))
```
pchote: This doesn't appear to be related to the hotkey changes, and has its own issues that are best dealt with in its own pr.  Could you please split it off from this one?

penev92:
Bump.
You should remove the changes to `Crate`, because those are not related to this PR, and address my style comment. Then we can merge this easily.

bryanwilbur:
I am in the middle of moving, I will remove that commit next week.

Thanks,


Bryan

---- On Tue, 14 Jul 2015 13:28:07 -0600 Pavel Penev &lt;notifications@github.com&gt; wrote ---- 

Bump.
 You should remove the changes to Crate, because those are not related to this PR, and address my style comment. Then we can merge this easily.
 —
Reply to this email directly or view it on GitHub.
    
 
 





​

pchote:
Thanks for the patches.  I have filed #8777 so that we can get this feature merged in time for the next playtest.

Issue #8587, opened 2015-06-28
Music ends unexpectedly when pressing "Start Game" in the lobby


obrakmann:
Introduced in #8568.

reaperrr:
The way I understand it, stopping it when the world changes is not wrong. I think It should restart when starting a game, though.

penev92:
@GraionDilach is already on it, as far as I know.

pchote:
I consider this a feature, but it would be nice if we could fade it out instead of stopping abruptly.

pchote:
In 99% of cases you will be transitioning from either the shellmap or score music (once somebody finally decides to implement it), so there's not much value to persisting the music state.

GraionDilach:
It's already fixed in #8562.

GraionDilach:
This is caused by the fact that worlds overlap in the memory, the garbage collector will only trigger stopping the music after the new world has already been constructed.

obrakmann:
Closing as it's expected behaviour.

PR #8607, opened 2015-06-30
Dispose of the old world before creating a new one

penev92:
As discussed on IRC - [link](http://logs.openra.net/?year=2015&month=06&day=30#11:26:44).
The problem was that the new world was initialized before the old one was destroyed, so they overlapped, which led to a bunch of issues that we dealt with in different ways (unfortunately).

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -135,6 +135,10 @@ namespace OpenRA
 		public static event Action BeforeGameStart = () => { };
 		internal static void StartGame(string mapUID, WorldType type)
 		{
+			// Dispose of the old world before creating a new one.
+			if (worldRenderer != null)
+				worldRenderer.Dispose();
+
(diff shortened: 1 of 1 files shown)


pchote:
> a bunch of issues that we dealt with in different ways (unfortunately).

What kind of ways?  Can they be removed now?

penev92:
I was referring to the recent PR that @obrakmann did to sanitize the Lua triggers, which was probably also a result of this. Can't say if that can be changed without testing though.

GraionDilach:
:+1: for this but noidea what PRs are affected. If this works, this should already fix #8587.

obrakmann:
Looks good and doesn't seem to cause any breakage. :+1: 

@GraionDilach: it doesn't fix #8587, `Sound.StopMusic` is still run from `World::Dispose()`, just a bit earlier.

PR #8609, opened 2015-06-30
Fixed another bad choice of lock objects in ColorMixerWidget

Mailaender:
A followup of https://github.com/OpenRA/OpenRA/pull/8425 as Coverity is still complaining. The problem may be that `back` is changed in another lock/thread which is twitchy and can race.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ColorMixerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ColorMixerWidget.cs
@@ -98,26 +98,23 @@ namespace OpenRA.Mods.Common.Widgets
 					hue = H;
 				}
 
-				lock (back)
+				unsafe
 				{
-					unsafe
+					// Generate palette in HSV
+					fixed (byte* cc = &back[0])
 					{
-						// Generate palette in HSV
-						fixed (byte* cc = &back[0])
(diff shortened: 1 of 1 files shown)


GraionDilach:
:+1: Definitely that's the issue here. Let's prevent racing.

OpenRA.Mods.Common/Widgets/ColorMixerWidget.cs
```diff
@@ -31,6 +31,7 @@ public class ColorMixerWidget : Widget
 		bool isMoving;
 
 		bool update;
+		readonly object backWorker = new object();
```
pchote: We now have three sync objects, each using a different naming scheme.

RoosterDragon:
Actually I think we can drop this lock altogether. Only the worker thread writes to the back buffer, and the other lock ensures the front and back are swapped safely.

chrisforbes:
:mask: this is all far too subtle and prone to failure.

Mailaender:
Removed the unnecessary lock object: https://github.com/OpenRA/OpenRA/pull/8609/files?w=1

RoosterDragon:
Works for me :+1:

PR #8610, opened 2015-06-30
Fixed a possible NRE in Hotkey.TryParse

Mailaender:
Discovered by Coverity.

### DIFF
--- a/OpenRA.Game/Input/Hotkey.cs
+++ b/OpenRA.Game/Input/Hotkey.cs
@@ -24,6 +24,9 @@ namespace OpenRA
 		public static bool TryParse(string s, out Hotkey result)
 		{
 			result = Invalid;
+			if (string.IsNullOrWhiteSpace(s))
+				return false;
+
 			var parts = s.Split(' ');
 
 			Keycode key;


OpenRA.Game/Input/Hotkey.cs
```diff
@@ -24,6 +24,9 @@ public bool IsValid()
 		public static bool TryParse(string s, out Hotkey result)
 		{
 			result = Invalid;
+			if (string.IsNullOrWhiteSpace(s))
```
penev92: Are you sure you don't want `NullOrEmpty` here?
What about pressing Spacebar?

Mailaender: .Split is actually the offender below and it will also make ` ` an impossible parse.

abcdefg30:
:+1:

OpenRA.Game/Input/Hotkey.cs
pchote: Spacebar corresponds to the string `SPACE`, so there's no problem here.

pchote:
This won't fix any real crashes, but that doesn't mean we shouldn't fix this :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/cc48afa23705add3002e89c6e9c34d7445f3070e%5E...cc48afa23705add3002e89c6e9c34d7445f3070e

PR #8611, opened 2015-06-30
Fixed another null reference exception in PngLoader

Mailaender:
Discovered by Coverity. Just like https://github.com/OpenRA/OpenRA/pull/8312 catching it and re-throw gives a more helpful error message.


GraionDilach:
Oh, nice. :+1:

PR #8614, opened 2015-07-01
Added a Capture function to lua

abcdefg30:
Split from #8416 as requested: https://github.com/OpenRA/OpenRA/pull/8416#discussion_r32369236

Has already one :+1: from https://github.com/OpenRA/OpenRA/pull/8416#issuecomment-113252033.

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Properties/CaptureProperties.cs
+++ b/OpenRA.Mods.Common/Scripting/Properties/CaptureProperties.cs
@@ -1,0 +1,47 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using Eluant;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Scripting;
+using OpenRA.Traits;
+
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Scripting/Properties/CaptureProperties.cs
```diff
@@ -0,0 +1,44 @@
...
+		[Desc("Captures the target actor.")]
+		public void Capture(Actor target)
+		{
+			var normalCapturable = target.Info.Traits.GetOrDefault<CapturableInfo>();
+			var externalCapturable = target.Info.Traits.GetOrDefault<ExternalCapturableInfo>();
+
+			if (normalInfo != null && normalCapturable != null && normalInfo.CaptureTypes.Contains(normalCapturable.Type))
+				Self.QueueActivity(new CaptureActor(Self, target));
+			else if (externalInfo != null && externalCapturable != null && externalInfo.CaptureTypes.Contains(externalCapturable.Type))
+				Self.QueueActivity(new ExternalCaptureActor(Self, Target.FromActor(target)));
```
penev92: How about an `else throw new InvalidOperationException("Actor X cannot capture")` or something, so when someone screws up it doesn't fail silently.

abcdefg30:
Added `else throw new LuaException("Actor '{0}' cannot capture actor '{1}'!".F(Self, target));`.

penev92:
Thanks.
Looks good to me now :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4d1f5bc59a646c0134e6c1a6d8954a3a14ac5162...074e440714b2aad490d09f8cdf40f9d6dc8f0625)

Issue #8613, opened 2015-06-30
Soviet mission 05


Townes27:
Tried out the latest playtest.
There are several issues with this mission.
1. The AI is completely passive and does not even harvest on the main island despite having a harvester.
2. The AI doesn't care whether it's being attacked or not.
3. Either the mission does not want to end or I'm too impatient, but after capturing the enemy buildings, destroying all allied units and having waited for about 20 minutes for the end, I decided it's time to write this bug report.

Townes27:
Replayed it on all difficulty levels.
Same three issues as above and also got this on Normal and Hard:

<pre><code>Fatal Lua Error: [string "BindingSupport.lua"]:30: Uncaught CLR exception at Lua->CLR boundary: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentNullException: Argument cannot be null.
Parameter name: source
  at System.Linq.Check.Source (System.Object source) [0x00000] in <filename unknown>:0 
  at System.Linq.Enumerable.ToList[Actor] (IEnumerable`1 source) [0x00000] in <filename unknown>:0 
  at OpenRA.Mods.Common.Scripting.TriggerGlobal.OnAllKilled (OpenRA.Actor[] actors, Eluant.LuaFunction func) [0x00000] in <filename unknown>:0 
  at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <filename unknown>:0 
  --- End of inner exception stack trace ---
  at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <filename unknown>:0 
  at System.Reflection.MethodBase.Invoke (System.Object obj, System.Object[] parameters) [0x00000] in <filename unknown>:0 
  at OpenRA.Scripting.ScriptMemberWrapper.Invoke (Eluant.LuaVararg args) [0x00000] in <filename unknown>:0 
[...]

PR #8615, opened 2015-07-01
Fix a crash in soviet05

abcdefg30:
Fixes the crash reported in #8613.
(See also http://pastebin.com/KRCStqfK and http://pastebin.com/MH9vV6Y9 .)


Mailaender:
:+1: Very likely to fix it as `unit` does not exist in that context.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/39752050ea58fdf51bbb48eaddcd01740800711a...806985ab5028aaf60bc15d3a6b45c3818ddd4b8e)

Issue #2901, opened 2013-03-30
Additional Spy Functionality


Surrealistik:
Spies should be able to detect and kill each other with their silenced pistols (Spy vs Spy/counterintelligence). A Spy only attacks in this way if it's disguised as an allied unit (or undisguised), and not an enemy one to avoid blowing its cover.

As per the following thread:

http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=16145

As an alternative to these infiltration effects, you can have Spies place remotely detonated explosives in buildings it infiltrates, which also includes production facilities, power plants, super weapons, tech centres and defensive structures (takes about as long as an Engineer capture to set explosives), reemerging after the explosives are set. You gain sight of the infiltrated structure. The Spy can 'deploy' in order to detonate all placed explosives, which instantly destroys the buildings they're planted in after a short delay (~3 seconds). This is similar to capabilities he demonstrated in Soviet Mission 3 from the original campaign. This effect is in addition to existing infiltration effects for the Radar Dome and Ore Refinery. Just before the buildings detonate, the victimized player will hear the triggering Spy proclaim: "For king and country" in his suave Connery voice.

If the Spy is killed before it can detonate its explosives, those explosives become harmless, and the Spy's controller loses sight of the structures it has infiltrated.

Spies can have their price increased (+100%) to compensate for this increased versatility. To help keep them useful for stealing credits, you could reimburse 50% of their cost on infiltration of Ore Refineries.

CCob:
I agree, there is little to counter a spy if you are allied v allied, short of spotting the little fella with your own eyes, which his difficult on large maps

Unit158:
Dogs were implemented for a reason...

PR #8616, opened 2015-07-01
Added counter intelligence to spies

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/2901 as the only balancing change everyone agreed on there is that Allies vs. Allies match-up needs a counter to the spy. It will matter more when we actually make the infiltration useful like in https://github.com/OpenRA/OpenRA/pull/8210.

### DIFF
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -186,7 +186,7 @@ SPY:
 	DisguiseToolTip:
 		Name: Spy
 		GenericName: Soldier
-		Description: Infiltrates enemy structures for intel or\nsabotage. Exact effect depends on the\nbuilding infiltrated.\nLoses disguise when attacking.\n  Strong vs Infantry\n  Weak vs Vehicles, Aircraft\n  Special Ability: Disguised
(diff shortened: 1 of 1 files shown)


GraionDilach:
I would've make the Kennel/Dog shared, but that might introduce more issues in the long run so :+1:.

chrisforbes:
:+1:

penev92:
Doesn't this make usage of spies against Allies a bit impossible?
When you know I will have dogs guarding, you can pick them off before you get the spy to my base.
But when you know any one of my soldiers can be a disguised spy that can detect your spy... what can you do about it?

Unit158:
Kill ALL the soldiers :D

Besides, counter intelligence works both ways :)

penev92:
Eh, what the heck.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f84b55ff54fd7f83cf74b590493974164ae1d675...3cff87701b4584d16f37fb183c769152f692245d)

PR #7737, opened 2015-03-25
Improve support power timer display.

deniz1a:
Improves support power countdown timer text display on top left of the screen. Fixes #4798. Also see #6282.

This commit will change the following:
1) Player's own countdown timers aren't shown in the text. This is redundant since it is already shown in the icons.
2) Timers are divided into two groups: Allies and Enemies.
3) Each group has a header which can be configured in ingame.yaml under SupportPowerTimer@SUPPORT_POWER_TIMER:.
4) The old Order parameter in ingame.yaml is repurposed to configure which group will be shown on top. Default is EnemiesFirst, I thought enemy timers would be more important. Old Order parameter wasn't working for Ascending value.
5) Player names are also shown in front of timer text. When the timer is zero, only the description text flashes white and not the player name.

Here's what it looks like:
timer

I tested this and seems to work.

Issue #7855, opened 2015-04-04
SupportPowers are outside the screen


Kaito-Kid:
If you have to many SupportPowers they are outside the display.
It would be better if OpenRA creates new columns (depending on display resolution).

![1](https://cloud.githubusercontent.com/assets/5711422/6993179/b5e7799a-daec-11e4-8cab-0e8c83d35702.jpg)

![2](https://cloud.githubusercontent.com/assets/5711422/6993180/bb99e85a-daec-11e4-9f76-81e70d674ead.jpg)

UberWaffe:
I'd vote for handling it the C&C:Generals way. In that multiple powers of the same type only get one icon, and the 'progress' of that icon shows the minimum time left before being ready for that group. Multiple ready ones show up as a number on the icon. (I.e. x3 = Means 3 atomic bombs ready for launch.)

That will allow for up to 12 types of support powers before a second column is needed, even if you have 50 of each of those 12 types.

SilverlineDev:
12 with that particular vertical resolution, I assume you mean.

Kaito-Kid:
I think UberWaffe's idea is the best.

sinf:
I'm also a fan of UberWaffe's idea.

PR #8619, opened 2015-07-02
Let's improve support power timer display.

deniz1a:
Hello. I made a new PR. This is a simpler reimplementation of #7737. It organizes support power countdown timers in teams. Thank you.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
@@ -32,19 +32,20 @@ namespace OpenRA.Mods.Common.Widgets
 		// Represents one line of text
 		struct Text
 		{
-			public Color Color;
-			public string Team;		// Used only for team label
+			public Color PlayerColor;
 			public string Player;
 			public string Power;
 			public string RemainingTime;
+			public bool Ready;
+			public string Team;		// Used only for team label
 		}
 
 		readonly List<Text> display;
 		readonly bool noPowers;
-		bool alternateColor, needTeamLabel, showPlayer;
+		bool flashing, needTeamLabel, showPlayer;
+		Color timerColor;
 
 		readonly World world;
-		int lastTick;
 
 		[ObjectCreator.UseCtor]
 		public SupportPowerTimerWidget(World world)
@@ -85,11 +86,9 @@ namespace OpenRA.Mods.Common.Widgets
 
 			// Generate texts once in every 10 WorldTicks,
 			// which is 0.4 seconds in game time.
-			if (world.WorldTick - lastTick > 9)
+			if (world.WorldTick % 10 == 0)
 			{
-				lastTick = world.WorldTick;
 				display.Clear();
-				alternateColor = Game.LocalTick % 50 > 25;
 				foreach (var team in playersByTeam)
 				{
 					needTeamLabel = true;
@@ -113,7 +112,8 @@ namespace OpenRA.Mods.Common.Widgets
 						foreach (var power in powers)
 						{
 							var timer = new Text();
-							timer.Color = power.Ready && alternateColor ? Color.White : player.Color.RGB;
+							timer.PlayerColor = player.Color.RGB;
+							timer.Ready = power.Ready;
 							if (showPlayer)
 							{
 								timer.Player = player.PlayerName;
@@ -131,6 +131,7 @@ namespace OpenRA.Mods.Common.Widgets
 			if (!display.Any())
 				return;
 
+			flashing = Game.LocalTick % 50 < 25;
 			var widgetBounds = new float2(Bounds.Location);
 			var position = new float2(0, 0);
 			foreach (var text in display)
@@ -143,12 +144,13 @@ namespace OpenRA.Mods.Common.Widgets
 				}
 
 				if (text.Player != null)
-					font.DrawTextWithContrast(text.Player, widgetBounds + position, text.Color, Color.Black, 1);
+					font.DrawTextWithContrast(text.Player, widgetBounds + position, text.PlayerColor, Color.Black, 1);
 
+				timerColor = text.Ready && flashing ? Color.White : text.PlayerColor;
 				position.X = powerPos;
-				font.DrawTextWithContrast(text.Power, widgetBounds + position, text.Color, Color.Black, 1);
+				font.DrawTextWithContrast(text.Power, widgetBounds + position, timerColor, Color.Black, 1);
 				position.X = timePos;
-				font.DrawTextWithContrast(text.RemainingTime, widgetBounds + position, text.Color, Color.Black, 1);
+				font.DrawTextWithContrast(text.RemainingTime, widgetBounds + position, timerColor, Color.Black, 1);
 
 				position.X = 0;
 				position.Y += yIncrement;
--- a/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
@@ -11,55 +11,148 @@
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
+using OpenRA.Graphics;
 using OpenRA.Mods.Common.Traits;
-using OpenRA.Primitives;
 using OpenRA.Widgets;
 
 namespace OpenRA.Mods.Common.Widgets
 {
 	public class SupportPowerTimerWidget : Widget
 	{
-		public readonly string Font = "Bold";
-		public readonly string Format = "{0}: {1}";
-		public readonly TimerOrder Order = TimerOrder.Descending;
+		[Translate] public readonly string NoTeamText = "No Team";
+		[Translate] public readonly string TeamText = "Team {0}";
 
-		readonly IEnumerable<SupportPowerInstance> powers;
-		Pair<string, Color>[] texts;
+		readonly Dictionary<Player, SupportPowerManager> spManagers;
+		readonly SortedDictionary<int, Player[]> playersByTeam;
+
+		readonly string fontName = "Bold";
+		readonly SpriteFont font;
+		readonly float yIncrement, powerPos, timePos;
+
+		// Represents one line of text
+		struct Text
+		{
+			public Color Color;
+			public string Team;		// Used only for team label
+			public string Player;
+			public string Power;
+			public string RemainingTime;
+		}
+
+		readonly List<Text> display;
+		readonly bool noPowers;
+		bool alternateColor, needTeamLabel, showPlayer;
+
+		readonly World world;
+		int lastTick;
 
 		[ObjectCreator.UseCtor]
 		public SupportPowerTimerWidget(World world)
 		{
-			powers = world.ActorsWithTrait<SupportPowerManager>()
-				.Where(p => !p.Actor.IsDead && !p.Actor.Owner.NonCombatant)
-				.SelectMany(s => s.Trait.Powers.Values)
-				.Where(p => p.Instances.Any() && p.Info.DisplayTimer && !p.Disabled);
-		}
+			spManagers = new Dictionary<Player, SupportPowerManager>();
+			foreach (var tp in world.ActorsWithTrait<SupportPowerManager>()
+				.Where(tp => !tp.Actor.IsDead && !tp.Actor.Owner.NonCombatant))
+				spManagers[tp.Actor.Owner] = tp.Trait;
 
-		public override void Tick()
-		{
-			texts = powers.Select(p =>
-			{
-				var time = WidgetUtils.FormatTime(p.RemainingTime, false);
-				var text = Format.F(p.Info.Description, time);
-				var color = !p.Ready || Game.LocalTick % 50 < 25 ? p.Instances[0].Self.Owner.Color.RGB : Color.White;
-				return Pair.New(text, color);
-			}).ToArray();
+			var availablePowers = world.Map.Rules.Actors.Values
+				.SelectMany(ai => ai.Traits.WithInterface<SupportPowerInfo>())
+				.Where(i => i.DisplayTimer).ToArray();
+
+			noPowers = !spManagers.Keys.Any() || !availablePowers.Any();
+			if (noPowers)
+				return;
+
+			playersByTeam = new SortedDictionary<int, Player[]>();
+			foreach (var team in spManagers.Keys.GroupBy(p => world.LobbyInfo.ClientWithIndex(p.ClientIndex).Team))
+				playersByTeam[team.Key] = team.ToArray();
+
+			font = Game.Renderer.Fonts[fontName];
+			yIncrement = font.Measure(" ").Y + 5;
+			var padding = font.Measure("  ").X;
+			var maxPlayerWidth = spManagers.Keys.Max(p => font.Measure(p.PlayerName).X);
+			var maxPowerWidth = availablePowers.Max(i => font.Measure(i.Description).X);
+			powerPos = maxPlayerWidth + padding;
+			timePos = powerPos + maxPowerWidth + padding;
+
+			display = new List<Text>();
+			this.world = world;
 		}
 
 		public override void Draw()
 		{
-			if (!IsVisible() || texts == null)
+			if (!IsVisible() || noPowers)
 				return;
 
-			var y = 0;
-			foreach (var t in texts)
+			// Generate texts once in every 10 WorldTicks,
+			// which is 0.4 seconds in game time.
+			if (world.WorldTick - lastTick > 9)
 			{
-				var font = Game.Renderer.Fonts[Font];
-				font.DrawTextWithContrast(t.First, new float2(Bounds.Location) + new float2(0, y), t.Second, Color.Black, 1);
-				y += (font.Measure(t.First).Y + 5) * (int)Order;
+				lastTick = world.WorldTick;
+				display.Clear();
+				alternateColor = Game.LocalTick % 50 > 25;
+				foreach (var team in playersByTeam)
+				{
+					needTeamLabel = true;
+					foreach (var player in team.Value)
+					{
+						var powers = spManagers[player].Powers.Values
+							.Where(i => i.Instances.Any() && i.Info.DisplayTimer && !i.Disabled).ToArray();
+
+						if (!powers.Any())
+							continue;
+
+						showPlayer = true;
+						if (needTeamLabel)
+						{
+							var teamLabel = new Text();
+							teamLabel.Team = team.Key == 0 ? NoTeamText : string.Format(TeamText, team.Key);
+							display.Add(teamLabel);
+							needTeamLabel = false;
+						}
+
+						foreach (var power in powers)
+						{
+							var timer = new Text();
+							timer.Color = power.Ready && alternateColor ? Color.White : player.Color.RGB;
+							if (showPlayer)
+							{
+								timer.Player = player.PlayerName;
+								showPlayer = false;
+							}
+
+							timer.Power = power.Info.Description;
+							timer.RemainingTime = WidgetUtils.FormatTime(power.RemainingTime, false);
+							display.Add(timer);
+						}
+					}
+				}
+			}
+
+			if (!display.Any())
+				return;
+
+			var widgetBounds = new float2(Bounds.Location);
+			var position = new float2(0, 0);
+			foreach (var text in display)
+			{
+				if (text.Team != null)
+				{
+					font.DrawTextWithContrast(text.Team, widgetBounds + position, Color.White, Color.Black, 1);
+					position.Y += yIncrement;
+					continue;
+				}
+
+				if (text.Player != null)
+					font.DrawTextWithContrast(text.Player, widgetBounds + position, text.Color, Color.Black, 1);
+
+				position.X = powerPos;
+				font.DrawTextWithContrast(text.Power, widgetBounds + position, text.Color, Color.Black, 1);
+				position.X = timePos;
+				font.DrawTextWithContrast(text.RemainingTime, widgetBounds + position, text.Color, Color.Black, 1);
+
+				position.X = 0;
+				position.Y += yIncrement;
 			}
 		}
-
-		public enum TimerOrder { Ascending = -1, Descending = 1 }
 	}
 }
--- a/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
@@ -94,6 +94,9 @@ namespace OpenRA.Mods.Common.Widgets
 					needTeamLabel = true;
 					foreach (var player in team.Value)
 					{
+						if (!player.IsAlliedWith(world.RenderPlayer) || player == world.RenderPlayer)
+							continue;
+
 						var powers = spManagers[player].Powers.Values
 							.Where(i => i.Instances.Any() && i.Info.DisplayTimer && !i.Disabled).ToArray();
 
--- a/mods/d2k/chrome/ingame.yaml
+++ b/mods/d2k/chrome/ingame.yaml
@@ -19,10 +19,6 @@ Container@INGAME_ROOT:
 				StrategicProgress@STRATEGIC_PROGRESS:
 					X: WINDOW_RIGHT/2
 					Y: 40
-				SupportPowerTimer@SUPPORT_POWER_TIMER:
-					X: 80
-					Y: 34
-					Order: Descending
 				Container@PLAYER_ROOT:
 				Container@PERF_ROOT:
 		Container@MENU_ROOT:
--- a/mods/ra/chrome/ingame-observer.yaml
+++ b/mods/ra/chrome/ingame-observer.yaml
@@ -1,5 +1,8 @@
 Container@OBSERVER_WIDGETS:
 	Children:
+		SupportPowerTimer@SUPPORT_POWER_TIMER:
+			X: 10
+			Y: 10
 		Image@SIDEBAR_BACKGROUND_TOP:
 			X: WINDOW_RIGHT - 250
 			Y: 10
--- a/mods/ra/chrome/ingame-player.yaml
+++ b/mods/ra/chrome/ingame-player.yaml
@@ -25,6 +25,9 @@ Container@PLAYER_WIDGETS:
 							IgnoreMouseOver: true
 							ImageCollection: sidebar
 							ImageName: background-supportoverlay
+		SupportPowerTimer@SUPPORT_POWER_TIMER:
+			X: 80
+			Y: 10
 		Image@SIDEBAR_BACKGROUND_TOP:
 			Logic: AddRaceSuffixLogic
 			X: WINDOW_RIGHT - 250
--- a/mods/ra/chrome/ingame.yaml
+++ b/mods/ra/chrome/ingame.yaml
@@ -19,10 +19,6 @@ Container@INGAME_ROOT:
 				StrategicProgress@STRATEGIC_PROGRESS:
 					X: WINDOW_RIGHT/2
 					Y: 40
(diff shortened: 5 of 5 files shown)


penev92:
Have you given up on the idea of getting the maximum name length in the constructor and using that? Because you have code commented out there.

deniz1a:
I tried getting maxPowerNameWidth in constructor but couldn't do it. maxPlayerNameWidth could be in constructor but this way the texts align dynamically. So when a player with a long name doesn't build a support power, it won't be considered in alignment. But to do that this calculates maxPlayerNameWidth every tick.. So if dynamic alignment is not necessary, it could be done using the commented line in constructor.

Which one is better?

Unit158:
Dynamic alignment is better in this case.

Can we get a better name for Line and change it to a struct please?

deniz1a:
How should it be named? I thought line is the most fitting because that's what it represents. Any other suggestions?

Also, isn't class better than struct because it's passed by reference? Although it isn't passed to any method here, just added to a list. So it would probably make no difference? Should I change it to struct?

deniz1a:
I could rename Line to Text...

deniz1a:
And actually you can't change teams during game. Stance change in diplomacy window doesn't change player's team, only individual stances to other players... So I can move playersByTeam definition to the constructor.

penev92:
This looks like a more complicated version of [this](https://github.com/penev92/OpenRA/blob/ba6b3b8d4315750056db5fa816078289b87c4d53/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs).
What are the benefits of this one over that?

deniz1a:
I don't know, I haven't tested that.

deniz1a:
Can Draw() execute before the first Tick()?

penev92:
OK, I'm going to be blunt.

You already had a PR about this that was way more complicated that it needed to be. Because of that a lot of developers spent *a lot* of time on it, including myself. *Then*, at one point, it reached a dead end. Instead of just closing it, I decided to give it a last chance, and spent around four hours of my time to rewrite it in a much simpler way. (Please keep in mind that even before I did so my opinion was that enough time was wasted on the PR by so many people for no gain.) Ultimately the PR was closed because you didn't want to use my suggested solution.

Now you are opening a new PR about the same thing (which is more than likely unfinished feature-wise). The code looks like a slightly more complicated version of the code I suggested for your previous PR, plus a (unnecessary?) `Tick()` method.
You say that you "haven't even tested" my suggested code. (?)

In light of that, I am going to -1 any solution that is more complicated than the one I proposed, unless you can explain how it is better than mine.

deniz1a:
Why didn't you open a new pr with your version?

penev92:
Because I don't consider this as something important, and definitely not something that justifies so many people wasting so much time on. And I still don't. So either we resolve this in a reasonable timeframe this time around, or I will close it.
Also because everyone else seems to be OK with the current state of the timers.

deniz1a:
Are Tick() and Draw() methods synchronized? What happens when the game is paused? Tick() stops but Draw() continues, right? So it is better to put game state dependent code in Tick() and leave only drawing code in Draw() to minimize calculations.

deniz1a:
If you don't consider this important and don't think this feature should be implemented, then why did you waste so much time, as you put it? Why did you write your own version and then not open a pull request?

penev92:
> Instead of just closing it, I decided to give it a last chance

I figured that was the polite thing to do.
Don't get me wrong, I wanted to just close it and stop wasting everyone's time. And I definitely didn't want to spend 2 hours trying to figure out what your code does and 2 more simplifying it.

deniz1a:
Ok, so you have the code that implements team grouping? Open a PR and if that's better than this one, it can be reviewed and merged...

deniz1a:
I think this PR is finished and needs review. It does two things:
1) Separates SUPPORT_POWER_TIMER widget definition into ingame-player and ingame-observer. This makes it possible to define different bound offsets for observers who don't have support power icons.
2) Organizes support power countdown timers in player teams. It looks like this:
sptimer-simple

Power names and remaining times are aligned dynamically.
This map can be used to quickly test the timer display:
http://resource.openra.net/maps/3506/

Possible changes/improvements that could be done:
* Change ```class Text``` to ```struct``` as suggested by @Unit158.
* Calculate ```maxPlayerWidth``` once in the constructor for all players and not every tick for the players that have a support power.
* Calculate ```maxPowerWidth``` once in the constructor for all buildable support powers in current map or mod, instead of every tick. (I don't know how to do this.)
* Store ```playersByTeam``` in a data structure like Dictionary<> instead of querying it every tick. Or cache ```IOrderedEnumerable```. (I don't know how to do this.)

I can't think of anything else... OK, let's discuss!

deniz1a:
Here are Visual Studio profiling results (.vsps files) for bleed and this PR:
https://drive.google.com/folderview?id=0B_dMvsHPqPZkfnJhR1Ryck44Szh4aDhJMjNHd2oySnpsbDNxdjd4UEc1MVhzY0NVcUE0RHM&usp=sharing

abcdefg30:
Looks ok to me. :+1: / :white_check_mark:

pchote:
Do we want this feature?  The reception has been luke-warm at best.

deniz1a:
Doesn't this improve the timers considerably? Without team grouping and player names, I find the timers to be not that useful.

pchote:
That is the open question which nobody has really asked yet.  It adds more details, but it also adds a lot more visual clutter to the screen.  If people don't feel that extra detail is needed, then the additional content could be taken as an overall negative instead of improvement.

Unit158:
I'm going to day that player names are definitely a good thing as our colour validation is simply ass. Teams are a bit overkill but whatever.

penev92:
Popular opinion has been that we do not want this.
Also :-1: until you can convince me this really needs `Tick()` *and* `Draw()` methods.

GraionDilach:
I'm worried about that this can and will also run out of screen, increasing #7855 just even further.

abcdefg30:
If we remove the `Team` stuff this is kinda just a bigger version of #6282.

deniz1a:
I don't think team labels add that much to the length of the text. There are usually two teams... If that's a problem, team labels could be omitted if total number of teams exceeds a set number.

It is better to have separate Tick and Draw methods. It allows for separation of tasks. Tick does game state dependent logic, generates the actual text that will be drawn. Draw is only responsible for drawing on screen. It gets a list of texts and draws them. This has benefits for the following cases:
* When there is lag and Tick frequency is lower than Draw.
* When game is paused. Tick stops, Draw continues
* If in the future, Tick and Draw frequencies are decoupled, Draw could run at much higher rate than Tick. I think currently it runs with the same frequency as Tick (when there's no lag).

The other reason you listed is that this is more complicated than the version you implemented. If that is so, you should open a PR for it so we have a better implementation. It was nice of you to write that code to help me and I got some ideas from it. But what do you want now?

deniz1a:
Apart from all that, the more important issue with support power timers is that drawing that text consumes 15% of CPU time according to the profiling results I posted above. If those numbers are correct, there is a problem with text drawing code. Maybe until that is fixed, countdown timers could be shown in a menu widget?

deniz1a:
OpenRA.Graphics.SpriteFont.DrawTextWithContrast() consumes more than 14%.

deniz1a:
I ran another profiling test and Draw() of SupportPowerTimerWidget uses 18.46% :worried:

deniz1a:
I think Tick() and Draw() always run synchronized, so separating logic doesn't improve performance. But still it's better, right?

deniz1a:
Widget's Tick() runs with LocalTick, which is separate from world tick, right? But it's not necessary to run the code in this widget's Tick() when world state hasn't changed. So how can I make it run with world tick instead of local tick?

deniz1a:
Another issue with this code is that when game is paused, ready timers don't flash. Because colors are calculated in Tick() which is skipped if world tick hasn't changed. But I think this is not important.

deniz1a:
Ok I think the game works like this:
There are three different frequencies or timesteps.
* **world.WorldTick** is the game progress. This stops when game is paused and slows down when there's lag. This is synchronized for all players. Runs every 40 ms or 25 times per second when there's no lag.
* **orderManager.LocalFrameNumber** is increased when the Tick() methods of all the classes are called. This is also 25 per second. I think this is not affected by game progress, so constant.
* **Game.RenderFrame** is the frame number of screen drawing. The frequency of this is as high as the computer can render or the frame limit value in settings menu, whichever is lower. This is increased when the Draw() methods of all the classes are called.

Then there's also network tick which has 1/3 the frequency of LocalFrameNumber. That's when packets are sent I think.
The frequency order is: 
RenderFrame > LocalFrameNumber > WorldTick > network tick

Is all this correct?

So, for this PR, we want to generate timer texts every WorldTick and draw them every RenderFrame which is when Draw() is called.

deniz1a:
Although I just tested it and the RenderFrame frequency is around 250 fps and is not limited by frame limit setting...

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,146 @@
...
 		}
 
 		public override void Tick()
 		{
-			texts = powers.Select(p =>
+			if (lastTick == world.WorldTick)
+				return;
+
+			lastTick = world.WorldTick;
+			display = new List<Text>();
```
pchote: Creating a new list every tick creates work for the garbage collector.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,146 @@
...
+				var teamDisplay = new List<Text>();
+				foreach (var player in team.Value)
+				{
+					var spiList = spManagers[player].Powers.Values
+						.Where(i => i.Instances.Any() && i.Info.DisplayTimer && !i.Disabled);
+
+					if (!spiList.Any())
+						continue;
+
+					var playerWidth = font.Measure(player.PlayerName).X;
```
pchote: This won't change at runtime, so you can do it once in the constructor and then cache it.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,146 @@
...
+					if (playerWidth > maxPlayerWidth)
+						maxPlayerWidth = playerWidth;
+
+					var showPlayer = true;
+					foreach (var spi in spiList)
+					{
+						var powerWidth = font.Measure(spi.Info.Description).X;
```
pchote: Likewise here - you know all the possible support power names before you start, so you can build a dictionary at startup instead of measuring everything every tick.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
pchote: If you precalculate the power and player names it becomes cheap to calculate the lengths, and you can then drop this memory churning caching completely.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,146 @@
...
 		{
-			powers = world.ActorsWithTrait<SupportPowerManager>()
-				.Where(p => !p.Actor.IsDead && !p.Actor.Owner.NonCombatant)
-				.SelectMany(s => s.Trait.Powers.Values)
-				.Where(p => p.Instances.Any() && p.Info.DisplayTimer && !p.Disabled);
+			spManagers = new Dictionary<Player, SupportPowerManager>();
+			foreach (var tp in world.ActorsWithTrait<SupportPowerManager>().Where(tp => !tp.Actor.Owner.NonCombatant))
+				spManagers[tp.Actor.Owner] = tp.Trait;
+
+			playersByTeam = new SortedDictionary<int, List<Player>>();
```
pchote: What is the purpose of the `SortedDictionary` here?

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,146 @@
...
+			foreach (var team in spManagers.Keys.GroupBy(p => world.LobbyInfo.ClientWithIndex(p.ClientIndex).Team))
+			{
+				playersByTeam[team.Key] = new List<Player>();
+				playersByTeam[team.Key].AddRange(team);
```
pchote: You can do this in one line (and then drop the braces) with something like `playersByTeam[team.Key] = team.ToList()`.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,146 @@
...
+
+			font = Game.Renderer.Fonts[fontName];
+			yIncrement = font.Measure(" ").Y + 5;
+			padding = font.Measure("  ").X;
+
+			display = new List<Text>();
+			this.world = world;
 		}
 
 		public override void Tick()
```
pchote: Widget ticking is decoupled from world ticking, so there is no benefit to implementing this in `Tick` compared with inside a `lastTick == world.WorldTick` check inside Render.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
deniz1a: So the teams are returned in order. Normal dictionary returns items in undefined order, right?

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,146 @@
...
 		[ObjectCreator.UseCtor]
 		public SupportPowerTimerWidget(World world)
 		{
-			powers = world.ActorsWithTrait<SupportPowerManager>()
-				.Where(p => !p.Actor.IsDead && !p.Actor.Owner.NonCombatant)
-				.SelectMany(s => s.Trait.Powers.Values)
-				.Where(p => p.Instances.Any() && p.Info.DisplayTimer && !p.Disabled);
+			spManagers = new Dictionary<Player, SupportPowerManager>();
+			foreach (var tp in world.ActorsWithTrait<SupportPowerManager>().Where(tp => !tp.Actor.Owner.NonCombatant))
+				spManagers[tp.Actor.Owner] = tp.Trait;
```
deniz1a: Should I cache a list of all support power instances that have DisplayTimer true? Instead of sp managers? Does that change during game?

pchote: That won't work, because the instances will change as the game progresses.  What you *can* do is query all the SupportPowerInfos and then store a dictionary of Info to width.

deniz1a: The list of SupportPowerInstance of a player changes during game? How do I get a list of all available powers then?

deniz1a: Does this change during game?
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/SupportPowers/SupportPowerManager.cs#L28

pchote: You can query them from the game rules.

pchote: > Does this change during game?

Yes.  Look at how it is used in that file, and you'll see why/how.

deniz1a: Oh ok thanks.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
deniz1a: I can't get the names of all possible support powers... How can I do that?

pchote: `<Map rules>.<ActorInfos>.SelectMany(ai => ai.Traits.Where(t => t is SupportPowerInfo))`

deniz1a: I used this and it threw TargetInvocationException:
```
maxPowerWidth = world.Map.Rules.Actors.Values
				.Select(ai => ai.Traits.GetOrDefault<SupportPowerInfo>())
				.Max(i => font.Measure(i.Description).X);
```

pchote: It will, yes. Your version is completely different to my hint above, and assumes that every actor type has exactly one support power. That assumption fails, so it throws.

deniz1a: Ok, this doesn't crash but the list is empty:
```
maxPowerWidth = world.Map.Rules.Actors.Values
				.SelectMany(ai => ai.Traits.WithInterface<SupportPowerInfo>())
				.Max(i => font.Measure(i.Description).X);
```

deniz1a:
Refactored the code. Minimized object creation. But I get an NRE when I sell allied tech center.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,144 @@
...
+				lastTick = world.WorldTick;
+				display.Clear();
+				alternateColor = Game.LocalTick % 50 > 25;
+				foreach (var team in playersByTeam)
+				{
+					needTeamLabel = true;
+					foreach (var player in team.Value)
+					{
+						var powers = spManagers[player].Powers.Values
+							.Where(i => i.Info.DisplayTimer && i.Instances.Any() && !i.Disabled).ToArray();
```
deniz1a: Does this change while it is still being queried and cause the NRE?

deniz1a:
Fixed NRE, added comments.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
Unit158: no. the support power will be destroyed before it is removed from the list.

deniz1a:
Looks finished to me, anything else that can be done?

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,148 @@
...
 
 namespace OpenRA.Mods.Common.Widgets
 {
 	public class SupportPowerTimerWidget : Widget
 	{
-		public readonly string Font = "Bold";
-		public readonly string Format = "{0}: {1}";
-		public readonly TimerOrder Order = TimerOrder.Descending;
+		[Translate] public readonly string NoTeamText = "No Team";
+		[Translate] public readonly string TeamText = "Team {0}";
```
GraionDilach: Exposing a direct C# format string to YAML is a bad idea. You sure there isn't a better way?

pchote: Unfortunately not, but that's a discussion for our wider translation efforts and is way out of scope of this pr.

GraionDilach: Alright then.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,148 @@
...
+			// Generate texts once in every 10 WorldTicks,
+			// which is 0.4 seconds in game time.
+			if (world.WorldTick - lastTick > 9)
 			{
-				var font = Game.Renderer.Fonts[Font];
-				font.DrawTextWithContrast(t.First, new float2(Bounds.Location) + new float2(0, y), t.Second, Color.Black, 1);
-				y += (font.Measure(t.First).Y + 5) * (int)Order;
+				lastTick = world.WorldTick;
+				display.Clear();
+				alternateColor = Game.LocalTick % 50 > 25;
```
GraionDilach: I see you carried this over but why not using 60 and 30 as magic values instead? Your changes turn ready texts from a balanced 50:50 blinking to one which prefers the player's color.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,148 @@
...
 		{
-			if (!IsVisible() || texts == null)
+			if (!IsVisible() || noPowers)
 				return;
 
-			var y = 0;
-			foreach (var t in texts)
+			// Generate texts once in every 10 WorldTicks,
+			// which is 0.4 seconds in game time.
+			if (world.WorldTick - lastTick > 9)
```
GraionDilach: This should be replaced with a (world.WorldTick % 10 == 0) check and lastTick should be removed then. The kind of optimization you're doing here should be done by the VM if needed.

deniz1a: I tried that and it seems to skip some ticks.. The countdown seconds look irregular, I don't know why?

GraionDilach: Ah, I see, it must be a race condition when Tick is faster than rendering - in that scenario your code might be a better option indeed.

GraionDilach: Another option worths a try is to use Game.LocalTick here to prevent that race condition.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
deniz1a: I think 50 and 25 were used because local tick frequency is 25 per second, if there is no lag. So this would ideally show player color and flashing color for one second each.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,150 @@
...
 		{
-			powers = world.ActorsWithTrait<SupportPowerManager>()
-				.Where(p => !p.Actor.IsDead && !p.Actor.Owner.NonCombatant)
-				.SelectMany(s => s.Trait.Powers.Values)
-				.Where(p => p.Instances.Any() && p.Info.DisplayTimer && !p.Disabled);
+			public Color PlayerColor;
+			public string Player;
+			public string Power;
+			public string RemainingTime;
+			public bool Ready;
```
deniz1a: Instead of having a separate Ready variable here, I could compare RemainingTime to zero time. I don't know which one is better.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -11,55 +11,150 @@
...
 		{
-			if (!IsVisible() || texts == null)
+			if (!IsVisible() || noPowers)
 				return;
 
-			var y = 0;
-			foreach (var t in texts)
+			// Generate texts once in every 10 WorldTicks,
+			// which is 0.4 seconds in game time.
+			if (world.WorldTick % 10 == 0)
```
deniz1a: The game cannot skip a world tick, right? Because that could cause this to not update often enough.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
GraionDilach: "Ideally", but then you only check every "0.4 seconds" leading to 1.2 for one color and 0.8 for the other.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
GraionDilach: >it must be a race condition when Tick is faster than rendering

Between your draws, the Ticks already pass. Either that, either the UI is separated from the world in this case.

Considering that the ready-text-blinking code uses Game.LocalTick instead, I'd assume that's much reliable than this. As I said it already.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
deniz1a: I changed that, now text flashing is done in draw part of the method. So it will flash even when the game is paused.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
deniz1a: But that's different. Ready text blinking and countdown timer flashing are only visual indications. They don't depend on world state.

pchote:
This has been modified since the original +1, so removing the needs +2 label (needs a full re-review).

obrakmann:
:-1: I don't want this.

In fact, I'd rather see the support power timers reduced to your own (ie. just the icons), and if need be your allies. Why should my enemies know when my spy satellite launches?

GraionDilach:
SpySat is a bad use-case. Knowing your enemy's Nuke/IonCannon is important. Not sure if we have a tag for hiding the countdown.

deniz1a:
Actually I also think the countdown timers should be shown only for allies. But with this PR I was trying to improve the current display without changing any behaviour. Should I close this PR and open one that reduces timers to only those belonging to allies? Or this PR could still be useful for observers. So I could add that feature to this one.

deniz1a:
Ok I removed timers that belong to players who are not allied with current player. Also removed player's own timers, they are already shown in icons.

abcdefg30:
Sorry, but :-1: for only showing timers of allied players.

deniz1a:
It's more logical and also improves gameplay I think.

obrakmann:
OK, it seems we can't come to an agreement over this, so it's probably best to stay at the status quo.

deniz1a:
I reverted the commit that hides enemy timers. I think there is an agreement that this would improve the game, the only part with disagreement was hiding enemy timers and I removed it.

deniz1a:
I'm keeping my branch upto date with bleed, in case an agreement is reached that this is better than current timers. Code is here:
https://github.com/deniz1a/OpenRA/tree/countdown-simple

GraionDilach:
Add something like a DisplayTimerToEnemies, set it false on GPS, true on AtomBomb and I think this is good then.

deniz1a:
Should I open a PR with those changes? Why should atom bomb timer be shown to enemies but GPS not?

GraionDilach:
Honestly I don't even understand why the GPS timer is even shown to the player, but meh. I'd consider the GPS as a support power (equal of Parabombs, Paratroopers or the Sonar Pulse) which does not majorly affect the game that much, compared to Chronosphere, Iron Curtain or the A-Bomb.

deniz1a:
I think it affects the gameplay but those timers should be shown only to allies. I'll add DisplayTimerToEnemies and maybe open another PR.

deniz1a:
I added ````DisplayTimerToEnemies```` to ````SupportPower````, so now it's configurable in yaml.

I think if there should be countdown timers in game, my countdown-simple branch is the way to do it. Even though the code is a lot more complex than the current implementation, there is no simpler way of doing it right. And despite the added logic, it is very efficient. It might even be faster to run than the current code, aside from any performance issues in font drawing.

In conclusion, I present to you this marvel of software engineering which improves the player experience significantly. The open question now is, if countdown timers are an important enough feature to justify maintaining this code. I think they are.

pchote:
Please do not open another pr.  For the reasons that we have been over (and over and over...) across the last two PRs and the IRC channel this is not as straightforward as you claim. I think I can speak for everyone on the reviewing side by saying that we are not willing to reshash these arguments again.

In conclusion, no.  Lets not improve the support power display.  The timers are not important enough to justify the amount of hair that you are making us pull out in frustration over this.

PR #8620, opened 2015-07-02
Allied units will not attack a building being captured

### LINKED ISSUE #6170: Allied units shouldn't attack a building which is being captured [$20]
Allied units shouldn't attack a building if an engineer is in the process of capturing it.

<bountysource-plugin>
---
There is a **[$20 open bounty](https://www.bountysource.com/issues/3478400-allied-units-shouldn-t-attack-a-building-which-is-being-captured?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

DrMelon:
Fixes #6170.
Units will still attack if the intent was to sabotage the building.

### DIFF
--- a/OpenRA.Mods.Common/Traits/AutoTarget.cs
+++ b/OpenRA.Mods.Common/Traits/AutoTarget.cs
@@ -45,11 +45,14 @@ namespace OpenRA.Mods.Common.Traits
 		readonly AutoTargetInfo info;
 		readonly AttackBase attack;
 		readonly AttackFollow at;
-		[Sync] int nextScanTime = 0;
+		[Sync]
+		int nextScanTime = 0;
 
 		public UnitStance Stance;
-		[Sync] public Actor Aggressor;
-		[Sync] public Actor TargetedActor;
+		[Sync]
+		public Actor Aggressor;
+		[Sync]
+		public Actor TargetedActor;
 
 		// NOT SYNCED: do not refer to this anywhere other than UI code
 		public UnitStance PredictedStance;
@@ -153,15 +156,19 @@ namespace OpenRA.Mods.Common.Traits
 
 			return inRange
 				.Where(a =>
+				{
+					var target = Target.FromActor(a);
+					return target.IsValidFor(self) &&
 					a.AppearsHostileTo(self) &&
-					!a.HasTrait<AutoTargetIgnore>() &&
 					attack.HasAnyValidWeapons(Target.FromActor(a)) &&
-					self.Owner.CanTargetActor(a))
+					self.Owner.CanTargetActor(a) &&
+					a.TraitsImplementing<IPreventsAutoTarget>().Any(t => t.CanAutoTarget(a, self));
+				})
 				.ClosestTo(self);
 		}
 	}
 
 	[Desc("Will not get automatically targeted by enemy (like walls)")]
 	class AutoTargetIgnoreInfo : TraitInfo<AutoTargetIgnore> { }
-	class AutoTargetIgnore { }
+	class AutoTargetIgnore : IPreventsAutoTarget { public bool CanAutoTarget(Actor self, Actor target) { return false; } }
 }
--- a/OpenRA.Mods.Common/Traits/ExternalCapturable.cs
+++ b/OpenRA.Mods.Common/Traits/ExternalCapturable.cs
@@ -49,10 +49,12 @@ namespace OpenRA.Mods.Common.Traits
 		public object Create(ActorInitializer init) { return new ExternalCapturable(init.Self, this); }
 	}
 
-	public class ExternalCapturable : ITick
+	public class ExternalCapturable : ITick, IPreventsAutoTarget
 	{
-		[Sync] public int CaptureProgressTime = 0;
-		[Sync] public Actor Captor;
+		[Sync]
+		public int CaptureProgressTime = 0;
+		[Sync]
+		public Actor Captor;
 		private Actor self;
 		public ExternalCapturableInfo Info;
 		public bool CaptureInProgress { get { return Captor != null; } }
@@ -65,6 +67,8 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void BeginCapture(Actor captor)
 		{
+			// Stops actors currently targeting this actor from targeting it.
+			self.Generation++;
 			var building = self.TraitOrDefault<Building>();
 			if (building != null)
 				building.Lock();
@@ -91,5 +95,7 @@ namespace OpenRA.Mods.Common.Traits
 			else
 				CaptureProgressTime++;
 		}
+
+		public bool CanAutoTarget(Actor self, Actor attacker) { return !CaptureInProgress; }
 	}
 }
--- a/OpenRA.Mods.Common/TraitsInterfaces.cs
+++ b/OpenRA.Mods.Common/TraitsInterfaces.cs
@@ -96,4 +96,6 @@ namespace OpenRA.Mods.Common.Traits
 		void MovementCancelled(Actor self);
 		void RequestTransport(CPos destination, Activity afterLandActivity);
 	}
+
+	public interface IPreventsAutoTarget { bool CanAutoTarget(Actor self, Actor attacker); }
 }


penev92:
Hi there!
Thanks for the patch. I see you are having the same issues as most new contributors regarding our style requirements:
 - Use tabs instead of spaces
 - Use `var` wherever possible instead of explicit types
 - Don't put braces (`{ }`) around a single line

OpenRA.Mods.Common/Activities/Attack.cs
```diff
@@ -64,6 +64,18 @@ protected virtual Activity InnerTick(Actor self, AttackBase attack)
 			if (!attack.Info.IgnoresVisibility && type == TargetType.Actor && !Target.Actor.HasTrait<FrozenUnderFog>() && !self.Owner.CanTargetActor(Target.Actor))
 				return NextActivity;
 
+            // Drop the target if it is being captured by our own team.
+            if(Target.Actor.HasTrait<Building>() && Target.Actor.HasTrait<Capturable>())
+            {
+                Capturable capturable = Target.Actor.TraitOrDefault<Capturable>();
+                CapturesInfo capturesinfo = Target.Actor.TraitOrDefault<CapturesInfo>();
```
Mailaender: Do those before and check against null to avoid expensive and unnecessary trait lookups.

DrMelon:
Hi there!
Sorry about that. I am using tabs (or at least, the tab key in VS2013), so I'll fix up those issues right away.

penev92:
Visual Studio (as most normal text editor) inserts N spaces when you press the Tab key. You can change that *somewhere* in the settings.

DrMelon:
Found it - I had it set to tabs in VS2012, annoying that it'd change back after updating to 2013. Changes added.

DrMelon:
Mailaender: I've made the changes you requested.

OpenRA.Mods.Common/Activities/Attack.cs
```diff
@@ -64,6 +64,12 @@ protected virtual Activity InnerTick(Actor self, AttackBase attack)
 			if (!attack.Info.IgnoresVisibility && type == TargetType.Actor && !Target.Actor.HasTrait<FrozenUnderFog>() && !self.Owner.CanTargetActor(Target.Actor))
 				return NextActivity;
 
+			// Drop the target if it is being captured by our own team.
+			var capturable = Target.Actor.TraitOrDefault<Capturable>();
+			var capturesinfo = Target.Actor.TraitOrDefault<CapturesInfo>();
+			if ((capturable != null && capturesinfo != null) && (capturable.BeingCaptured && !capturesinfo.Sabotage))
```
Mailaender: You forgot about `ExternalCapturable.CaptureInProgress` here.

OpenRA.Mods.Common/Activities/Attack.cs
```diff
@@ -64,6 +64,12 @@ protected virtual Activity InnerTick(Actor self, AttackBase attack)
 			if (!attack.Info.IgnoresVisibility && type == TargetType.Actor && !Target.Actor.HasTrait<FrozenUnderFog>() && !self.Owner.CanTargetActor(Target.Actor))
 				return NextActivity;
 
+			// Drop the target if it is being captured by our own team.
+			var capturable = Target.Actor.TraitOrDefault<Capturable>();
+			var capturesinfo = Target.Actor.TraitOrDefault<CapturesInfo>();
```
penev92: `Captures` would be on the engineer or whatever, not on the building, so checking for `CapturesInfo` won't give you anything (and I don't think you can check for sabotage with the current setup (unless we add a reference inside `Capturable` to the actor that is performing the capturing)).

DrMelon:
ExternalCapturable behaviour should be handled now too.

OpenRA.Mods.Common/Activities/Attack.cs
Mailaender: Yep, this is always true or false. You will have to check against `CaptureThreshold` and hit points. I am not sure if the complication is really required here. When someone wants to capture, just assume everyone should stop firing.

OpenRA.Mods.Common/Activities/Attack.cs
```diff
@@ -64,6 +64,14 @@ protected virtual Activity InnerTick(Actor self, AttackBase attack)
 			if (!attack.Info.IgnoresVisibility && type == TargetType.Actor && !Target.Actor.HasTrait<FrozenUnderFog>() && !self.Owner.CanTargetActor(Target.Actor))
 				return NextActivity;
 
+			// Drop the target if it is being captured by our own team.
+			// Cannot check for sabotage-in-progress without reference inside capturable object to the object doing the capturing.
+			var capturable = Target.Actor.TraitOrDefault<Capturable>();
+			var externalcapturable = Target.Actor.TraitOrDefault<ExternalCapturable>();
+			
+			if ((capturable != null || externalcapturable != null) && (capturable.BeingCaptured || externalcapturable.CaptureInProgress))
```
Mailaender: This will cause null reference exceptions this way as you don't chain the null check to the corresponding field.

```csharp
if ((capturable != null && capturable.BeingCaptured) || (externalcapturable != null && externalcapturable.CaptureInProgress))
```

is what you want here. Also far more readable.

DrMelon: Thanks, I missed that completely.

DrMelon:
Commits squashed.

OpenRA.Mods.Common/Activities/Attack.cs
```diff
@@ -64,6 +64,14 @@ protected virtual Activity InnerTick(Actor self, AttackBase attack)
 			if (!attack.Info.IgnoresVisibility && type == TargetType.Actor && !Target.Actor.HasTrait<FrozenUnderFog>() && !self.Owner.CanTargetActor(Target.Actor))
 				return NextActivity;
 
+			// Drop the target if it is being captured by our own team.
+			// Cannot check for sabotage-in-progress without reference inside capturable object to the object doing the capturing.
+			var capturable = Target.Actor.TraitOrDefault<Capturable>();
```
penev92: Looking at how `Capturable` works, `BeingCaptured` is true for only about half a tick, so you might as well skip that (the capture is instant and the check doesn't give you anything). The one that you want is `ExternalCapturable`, which takes a lot longer.

penev92:
OK, this works, but has a nasty side effect - once someone starts to capture a building, nobody can attack it. Not even by force-firing.
You might be able to resolve that by moving this check from the `Attack` activity (where it is checked on every tick, so even if you force-attack the building, the check would tell your units to stop on the first tick) to the code that scans for targets when the actor wants to find a target to attack - `AutoTarget.ScanForTarget()`. That way you can still order your units to attack the building if you want (imagine Player A is capturing Player B's building, and you are player C and just want to blow it up), but they won't attack it unless you tell them to.

DrMelon:
That does seem like a more elegant solution. I'll patch that in tomorrow.

pchote:
Increment the Target generation and the engine will handle the invalidation for you.

OpenRA.Mods.Common/Activities/Attack.cs
```diff
@@ -64,6 +64,14 @@ protected virtual Activity InnerTick(Actor self, AttackBase attack)
...
+			var externalcapturable = Target.Actor.TraitOrDefault<ExternalCapturable>();
```
pchote: You will need to be careful here. Actors may have multiple traits of these types.

DrMelon:
Right, I've shifted the fix into AutoTarget.ChooseTarget, as suggested - I have tried to keep the code mostly the same as the code that came before, but this has resulted in two calls to TraitOrDefault<>(). Does this raise performance/optimisation concerns?

penev92:
You can change that to something like
```
			return inRange.Where(a =>
				{
					var trait = a.TraitOrDefault<ExternalCapturable>();
					return a.AppearsHostileTo(self) &&
						!a.HasTrait<AutoTargetIgnore>() &&
						attack.HasAnyValidWeapons(Target.FromActor(a)) &&
						self.Owner.CanTargetActor(a)
						&& (trait == null || trait.CaptureInProgress);
				})
				.ClosestTo(self);
```
Which of the two would be faster is highly situational though.

DrMelon:
Wouldn't checking `(trait == null || trait.CaptureInProgress)` be the incorrect behaviour in this instance? As the condition returns true when something is allowed to be targeted, and so returning true if the capture was in progress would mean that the auto-targeting behaviour would begin? 
It would be better if it were `(trait == null || !trait.CaptureInProgress)` - but I've just realized that such behavior would mean that auto-targeting would fail on a building that had that trait but was not being captured. 
Ideally the check should return true if the building is not being captured, and so checking if the trait exists or not shouldn't be part of the overall condition, but only there to make sure that there won't be a null reference exception when checking for `.CaptureInProgress`.
This might mean I have to have two separate conditions, one for the case where there is such a trait, and one without... I can't figure out a way to unify the cases without introducing issues.

penev92:
Well I was just suggesting a syntax change to allow you to have a single lookup. I don't insist that it's correct.
Anyway, you can leave it as it is, my suggestion isn't really any better.
You should change yours to 
`a.HasTrait<ExternalCapturable>() && a.Trait<ExternalCapturable>().CaptureInProgress`
but @pchote had a different idea about implementation, which I'll let him explain.

pchote:
Here's how I would implement this:

1. Add `self.Generation++;` to `ExternalCapturable.BeginCapture`. This will make any actors that are currently targeting the building stop attacking it.
2. Introduce an `IPreventsAutoTarget` interface with a method `bool CanAutoTarget(Actor self, Actor attacker)` and make `ExternalCapturable` implement it and return false if it is being captured.
3. Add a check like `!a.TraitsImplementing<IPreventsAutoTarget>().Any(t => t.CanAutoTarget(a, self))` to `AutoTarget.ChooseTarget`.
4. Make `AutoTargetIgnore` implement `IPreventsAutoTarget` and always return true, then remove the explicit `!a.HasTrait<AutoTargetIgnore>()` from `ChooseTarget`.

pchote:
It looks like `AutoTarget` uses a direct reference to an actor instead of using a `Target`, so my suggestion would also need a "0.  Make AutoTarget use a proper Target" in order for the target dropping to work.

DrMelon:
Alright! I'll make those changes, @pchote. @penev92, I wasn't suggesting you were providing a complete & correct implementation, it was more like thinking aloud, bouncing ideas around!

DrMelon:
I forgot to run the latest change by StyleCop, hold on.

abcdefg30:
Could you squash the commits, please?

DrMelon:
Commits squashed.

DrMelon:
Or, apparently not. Hold on.

DrMelon:
Ok, I've screwed this up big-time. I'm going to reset this branch and recreate the solution.

DrMelon:
Done.

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -153,15 +156,19 @@ Actor ChooseTarget(Actor self, WRange range)
 
 			return inRange
 				.Where(a =>
+				{
+					var target = Target.FromActor(a);
```
pchote: Above I meant for `TargetedActor` to be a target, because we want that to become invalidated if it is captured (or teleported, or a few other conditions).  After a second look, that field is actually unused, which is a bit odd.

You don't need to worry about the target here, because it won't be invalidated in the course of this query.

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -45,11 +45,14 @@ public class AutoTarget : INotifyIdle, INotifyDamage, ITick, IResolveOrder, ISyn
 		readonly AutoTargetInfo info;
 		readonly AttackBase attack;
 		readonly AttackFollow at;
-		[Sync] int nextScanTime = 0;
+		[Sync]
```
abcdefg30: Could you revert those new line changes, please?

abcdefg30:
Capturing won't stop anymore after aborting the process:
captureabort

penev92:
Bump.
Are you around, @DrMelon?

DrMelon:
Oh, hey. Sorry I wasn't able to get in touch sooner. We might have to put this PR on hold, or hand off to someone else - some life stuff has come up for me and I can't work on this just now. Apologies for the mess.

penev92:
Well it has been a while now.
Closing for the moment to not clutter the PR queue. Feel free to reopen once you get some time.

PR #8621, opened 2015-07-02
Fixed an unneccessary and frequent trait lookup

Mailaender:
Those are actually constant values. No need to look up the underlying trait info each path check.


penev92:
:+1:

DrMelon:
:+1:

PR #7658, opened 2015-03-13
[Proof-of-Concept] Improved the performance and intelligence of resource harvesting

Rydra:
Improved the performance and intelligence of resource harvesting by refactoring the Harvesters' pathfinding. Now in first place they assess which is the closest resource inside their search area (previously it was quite "randomized" by the heuristic) and then a path is calculated. In fact, this change allowed me to get rid of the Harvester heuristic, which is great. This should improve performance as well.

I would appreciate help in testing that everything is in place, no regressions have appeared and any feedback you consider. This is a proof-of-concept and a work under development, so it would be cool if I require lot of input for it to improve performance and correctness.

Issue #7092, opened 2014-12-07
AI harvesters will worse case path find every second when there are no resources


Mailaender:
![image](https://cloud.githubusercontent.com/assets/756669/5332699/35b3429c-7e60-11e4-877a-7ed5d852d688.png)

reaperrr:
According to my tests this is THE main showstopper performance issue on larger maps, far worse than the AI base builder and regular pathfinding. Adding to Next Release milestone.

#7658 not only improves FindResources performance massively according to my tests (by a factor of about 20 in some cases, down from 234(!) milliseconds worstcase delay to at most 10 on RA Seaside 2 map), the performance improvement would also allow us to increase the search radius so harvesters a) don't run into this infinite "search resource - find none - try again - find none..." loop that easily, and b) find a new resource field more easily instead of getting stuck.

reaperrr:
Fixed in #8622.

Issue #6269, opened 2014-08-18
Harvesters refuse to harvest sometimes


VagueVigor:
I click to the ore, but it automatically goes somewhere else as if it couldnt find any ore from there:

http://i1.minus.com/ilava7nkKIyvB.jpg

I think this bug sometimes fixes itself during the game.

reaperrr:
This was before the harvester search refactor, and I haven't noticed this specific issue since then, so I guess this can be closed.

Issue #3089, opened 2013-04-17
Harvesters can get stuck on unloading


ScottNZ:
They seem to get stuck at the refinery when multiple harvesters are about to unload at the same refinery. The harvester that should unload does not unload and only unloads after you explicitly order it to. It happens pretty often during games where you have multiple harvesters attached to one refinery.
##

x-a-n-a-x:
It seems the problem has resurrected in the latest bleed: https://github.com/OpenRA/OpenRA/commit/43c3621ef697cbe3158e105eee5efd7a9844aa93
Tomsons26 told me that his harvesters were stuck in front of a Refinery. Chokepoint map.

![weee2pnt_o](https://cloud.githubusercontent.com/assets/1732773/6489799/ae7c9400-c2a1-11e4-83ef-16306bb1fe28.png)

![haviechoke](https://cloud.githubusercontent.com/assets/1732773/6489804/bb6060ca-c2a1-11e4-9fe4-f91f2a179442.png)

pchote:
This indeed appears to be much worse (tending towards unplayable) on latest bleed.  This will be a regression from the recent pathfinder changes.

Mailaender:
See also https://github.com/OpenRA/OpenRA/issues/7610.

Mailaender:
https://github.com/OpenRA/OpenRA/pull/7618 comes with a proposed fix.

MicroJOo:
@Mailaender #7618  is not a fix for this problem ... i allready gave it a test ...

Rydra:
My proposed solution should only solve a crash that may happen on a crowded situation, but does not solve the problem of blocked harvesters.

Can anyone point out which parts of the code deal with those harvester jams?

pchote:
@Rydra: the harvester problem probably originates from a behavioral change in `MobileInfo.CanEnterCell`.  This appears to be the same as the problem (the "obvious bug fix" that broke harvesters) that I mentioned in #7565.  I would double check your changes to that code, and also look at the nudging code that uses it (which is where the real problem probably lives).

Rydra:
Weird because I haven't touched anything relevant on nuding or mobile. I'll check it out and debug what happens.

Rydra:
[...]

PR #8622, opened 2015-07-02
Refactored harvester resource search

reaperrr:
Rebase of #7658, with some minor additions by me.

Fixes or at least massively improves #7092, #6269 and #3089.

### DIFF
--- a/OpenRA.Mods.Common/Activities/FindResources.cs
+++ b/OpenRA.Mods.Common/Activities/FindResources.cs
@@ -8,7 +8,6 @@
  */
 #endregion
 
-using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
@@ -58,91 +57,108 @@ namespace OpenRA.Mods.Common.Activities
 			if (harv.IsFull)
 				return Util.SequenceActivities(deliver, NextActivity);
 
+			var closestHarvestablePosition = ClosestHarvestablePos(self);
+
+			// If no harvestable position could be found, either deliver the remaining resources
+			// or get out of the way and do not disturb.
+			if (!closestHarvestablePosition.HasValue)
+			{
+				if (!harv.IsEmpty)
+					return deliver;
+
+				var cachedPosition = self.Location;
+				harv.UnblockRefinery(self);
+
+				// Only do this if UnblockRefinery did nothing.
+				if (self.Location == cachedPosition)
+				{
+					var unblockCell = harv.LastHarvestedCell ?? (self.Location + new CVec(0, 4));
+					var moveTo = mobile.NearestMoveableCell(unblockCell, 2, 5);
+					self.QueueActivity(mobile.MoveTo(moveTo, 1));
+					self.SetTargetLine(Target.FromCell(self.World, moveTo), Color.Gray, false);
+				}
+
+				var randFrames = self.World.SharedRandom.Next(100, 175);
+				return Util.SequenceActivities(NextActivity, new Wait(randFrames), this);
+			}
+			else
+			{
+				var next = this;
+
+				// Attempt to claim a resource as ours
+				if (territory != null)
+				{
+					if (!territory.ClaimResource(self, closestHarvestablePosition.Value))
+						return Util.SequenceActivities(new Wait(25), next);
+				}
+
+				// If not given a direct order, assume ordered to the first resource location we find:
+				if (!harv.LastOrderLocation.HasValue)
+					harv.LastOrderLocation = closestHarvestablePosition;
+
+				self.SetTargetLine(Target.FromCell(self.World, closestHarvestablePosition.Value), Color.Red, false);
+
+				var notify = self.TraitsImplementing<INotifyHarvesterAction>();
+
+				foreach (var n in notify)
+					n.MovingToResources(self, closestHarvestablePosition.Value, next);
+
+				return Util.SequenceActivities(mobile.MoveTo(closestHarvestablePosition.Value, 1), new HarvestResource(self), next);
+			}
+		}
+
+		bool IsHarvestable(Actor self, CPos pos)
+		{
+			var resType = resLayer.GetResource(pos);
+			if (resType == null)
+				return false;
+
+			// Can the harvester collect this kind of resource?
+			if (!harvInfo.Resources.Contains(resType.Info.Name))
+				return false;
+
+			if (territory != null)
+			{
+				// Another harvester has claimed this resource:
+				ResourceClaim claim;
+				if (territory.IsClaimedByAnyoneElse(self as Actor, pos, out claim))
+					return false;
+			}
+
+			return true;
+		}
+
+		/// <summary>
+		/// Finds the closest harvestable pos between the current position of the harvester
+		/// and the last order location
+		/// </summary>
+		CPos? ClosestHarvestablePos(Actor self)
+		{
 			// Determine where to search from and how far to search:
 			var searchFromLoc = harv.LastOrderLocation ?? (harv.LastLinkedProc ?? harv.LinkedProc ?? self).Location;
 			var searchRadius = harv.LastOrderLocation.HasValue ? harvInfo.SearchFromOrderRadius : harvInfo.SearchFromProcRadius;
 			var searchRadiusSquared = searchRadius * searchRadius;
 
-			// Find harvestable resources nearby:
-			var path = pathFinder.FindPath(
-				PathSearch.Search(self.World, mobileInfo, self, true)
-					.WithHeuristic(loc =>
-					{
-						// Avoid this cell:
-						if (avoidCell.HasValue && loc == avoidCell.Value)
-							return EstimateDistance(loc, searchFromLoc) + Constants.CellCost;
-
-						// Don't harvest out of range:
-						var distSquared = (loc - searchFromLoc).LengthSquared;
-						if (distSquared > searchRadiusSquared)
-							return EstimateDistance(loc, searchFromLoc) + Constants.CellCost * 2;
-
-						// Get the resource at this location:
-						var resType = resLayer.GetResource(loc);
-						if (resType == null)
-							return EstimateDistance(loc, searchFromLoc) + Constants.CellCost;
-
-						// Can the harvester collect this kind of resource?
-						if (!harvInfo.Resources.Contains(resType.Info.Name))
-							return EstimateDistance(loc, searchFromLoc) + Constants.CellCost;
-
-						if (territory != null)
-						{
-							// Another harvester has claimed this resource:
-							ResourceClaim claim;
-							if (territory.IsClaimedByAnyoneElse(self, loc, out claim))
-								return EstimateDistance(loc, searchFromLoc) + Constants.CellCost;
-						}
-
-						return 0;
-					})
-					.FromPoint(self.Location));
-
-			var next = this;
-
-			if (path.Count == 0)
-			{
-				if (!harv.IsEmpty)
-					return deliver;
-				else
+			var search = PathSearch.Search(self.World, mobileInfo, self, true,
+				loc => IsHarvestable(self, loc))
+				.WithCustomCost(loc =>
 				{
-					// Get out of the way if we are:
-					harv.UnblockRefinery(self);
-					var randFrames = self.World.SharedRandom.Next(90, 160);
-					if (NextActivity != null)
-						return Util.SequenceActivities(NextActivity, new Wait(randFrames), next);
-					else
-						return Util.SequenceActivities(new Wait(randFrames), next);
-				}
-			}
+					if ((avoidCell.HasValue && loc == avoidCell.Value) ||
+						(loc - self.Location).LengthSquared > searchRadiusSquared)
+						return int.MaxValue;
 
-			// Attempt to claim a resource as ours:
-			if (territory != null)
-			{
-				if (!territory.ClaimResource(self, path[0]))
-					return Util.SequenceActivities(new Wait(25), next);
-			}
+					return 0;
+				})
+				.FromPoint(self.Location)
+				.FromPoint(searchFromLoc);
 
-			// If not given a direct order, assume ordered to the first resource location we find:
-			if (harv.LastOrderLocation == null)
-				harv.LastOrderLocation = path[0];
+			// Find any harvestable resources:
+			var path = pathFinder.FindPath(search);
 
-			self.SetTargetLine(Target.FromCell(self.World, path[0]), Color.Red, false);
+			if (path.Count > 0)
+				return path[0];
 
-			var notify = self.TraitsImplementing<INotifyHarvesterAction>();
-			foreach (var n in notify)
-				n.MovingToResources(self, path[0], next);
-
-			return Util.SequenceActivities(mobile.MoveTo(path[0], 1), new HarvestResource(self), next);
-		}
-
-		// Diagonal distance heuristic
-		static int EstimateDistance(CPos here, CPos destination)
-		{
-			var diag = Math.Min(Math.Abs(here.X - destination.X), Math.Abs(here.Y - destination.Y));
-			var straight = Math.Abs(here.X - destination.X) + Math.Abs(here.Y - destination.Y);
-
-			return Constants.CellCost * straight + (Constants.DiagonalCellCost - 2 * Constants.CellCost) * diag;
+			return null;
 		}
 
 		public override IEnumerable<Target> GetTargets(Actor self)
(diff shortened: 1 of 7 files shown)


OpenRA.Mods.Common/Activities/DeliverResources.cs
```diff
@@ -46,8 +58,8 @@ public override Activity Tick(Actor self)
 			if (harv.LinkedProc == null || !harv.LinkedProc.IsInWorld)
 				harv.ChooseNewProc(self, null);
 
-			if (harv.LinkedProc == null)	// no procs exist; check again in 1s.
-				return Util.SequenceActivities(new Wait(25), this);
+			if (harv.LinkedProc == null)	// no refineries exist; check again in 5s.
+				return Util.SequenceActivities(new Wait(125), this);
```
Mailaender: Bonus points for storing these magic values in HarvesterInfo.

OpenRA.Mods.Common/Activities/DeliverResources.cs
```diff
@@ -17,6 +17,14 @@ namespace OpenRA.Mods.Common.Activities
...
 	{
+		IMove movement;
+		Harvester harv;
+
+		// TODO: This is a hack because this class is sometimes
+		// constructed by reflection and cannot add an Actor variable in
+		// the constructor. Eventually remove!
+		bool loadedVars = false;
+		
 		const int NextChooseTime = 100;
```
Mailaender: [SA1203] All constants must be placed before all non-constants.

OpenRA.Mods.Common/Activities/DeliverResources.cs
```diff
@@ -17,6 +17,14 @@ namespace OpenRA.Mods.Common.Activities
...
 	public class DeliverResources : Activity
 	{
+		IMove movement;
+		Harvester harv;
+
+		// TODO: This is a hack because this class is sometimes
+		// constructed by reflection and cannot add an Actor variable in
+		// the constructor. Eventually remove!
+		bool loadedVars = false;
+
```
Mailaender: [SP2000] Invalid spacing at the end of the line.

OpenRA.Mods.Common/Activities/FindResources.cs
```diff
@@ -21,6 +20,18 @@ namespace OpenRA.Mods.Common.Activities
 {
 	public class FindResources : Activity
 	{
+		ResourceLayer resLayer;
+		HarvesterInfo harvInfo;
+		ResourceClaimLayer territory;
+		Mobile mobile;
+		Harvester harv;
+		MobileInfo mobileInfo;
+		IPathFinder pathFinder;
```
Mailaender: You can put a round of `readonly` before those.

reaperrr: They are assigned later in Tick (see comment on ```loadVars``` below), so that's not possible in this case.

OpenRA.Mods.Common/Activities/DeliverResources.cs
```diff
@@ -46,8 +60,8 @@ public override Activity Tick(Actor self)
 			if (harv.LinkedProc == null || !harv.LinkedProc.IsInWorld)
 				harv.ChooseNewProc(self, null);
 
-			if (harv.LinkedProc == null)	// no procs exist; check again in 1s.
-				return Util.SequenceActivities(new Wait(25), this);
+			if (harv.LinkedProc == null)	// no refineries exist; check again in 5s.
```
Mailaender: The value is not hard-coded anymore so better remove the comment.

OpenRA.Mods.Common/Activities/FindResources.cs
```diff
@@ -34,103 +45,118 @@ public FindResources(CPos avoidCell)
...
 			{
 				if (!harv.IsEmpty)
 					return new DeliverResources();
+
+				harv.UnblockRefinery(self);
+				var moveTo = harv.LastHarvestedCell ?? (self.Location + new CVec(0, 4));
+				self.QueueActivity(mobile.MoveTo(moveTo, 1));
+				self.SetTargetLine(Target.FromCell(self.World, moveTo), Color.Gray, false);
+
+				var randFrames = 125 + self.World.SharedRandom.Next(-35, 35);
```
Mailaender: I know this carried over, but now I wonder why this isn't just `self.World.SharedRandom.Next(90, 160);`

OpenRA.Mods.Common/Activities/DeliverResources.cs
penev92: The comment is good. Just account for the modifiable time.

OpenRA.Mods.Common/Activities/DeliverResources.cs
```diff
@@ -18,6 +18,15 @@ namespace OpenRA.Mods.Common.Activities
 	public class DeliverResources : Activity
 	{
 		const int NextChooseTime = 100;
+
+		IMove movement;
+		Harvester harv;
+		HarvesterInfo harvInfo;
+
+		// TODO: This is a hack because this class is sometimes
+		// constructed by reflection and cannot add an Actor variable in
```
pchote: Where is it constructed by reflection?  This doesn't sound right to me.

OpenRA.Mods.Common/Activities/FindResources.cs
```diff
@@ -21,6 +20,18 @@ namespace OpenRA.Mods.Common.Activities
...
+		ResourceLayer resLayer;
+		HarvesterInfo harvInfo;
+		ResourceClaimLayer territory;
+		Mobile mobile;
+		Harvester harv;
+		MobileInfo mobileInfo;
+		IPathFinder pathFinder;
+
+		// TODO: This is a hack because this class is sometimes
+		// constructed by reflection and cannot add an Actor variable in
```
pchote: I would prefer it if we could fix this first in a separate pr.  We can replace the `InitialActivity` hack on `FreeActor` with code in `Harvester` that will queue this activity if it isn't suppressed by an ActorInit or a rules flag (and maybe tie this in to whatever logic makes harvs auto-harvest when built normally?)

OpenRA.Mods.Common/Activities/FindResources.cs
```diff
@@ -34,103 +45,118 @@ public FindResources(CPos avoidCell)
...
-			if (path.Count == 0)
+			// If no harvestable position could be found, either deliver the remaining resources
+			// or get out of the way and do not disturb.
+			if (!closestHarvestablePosition.HasValue)
 			{
 				if (!harv.IsEmpty)
 					return new DeliverResources();
+
+				harv.UnblockRefinery(self);
+				var moveTo = harv.LastHarvestedCell ?? (self.Location + new CVec(0, 4));
```
pchote: This looks like an awfully arbitrary constant.  It would be nice if it were at least randomizing the direction.

OpenRA.Mods.Common/Activities/DeliverResources.cs
obrakmann: That comment is about the `InitialActivity` actor init in `FreeActor`

reaperrr: I just didn't look up the reason for this approach in ```FindResources``` and wrongly assumed it would apply here, too.

OpenRA.Mods.Common/Activities/FindResources.cs
```diff
@@ -34,103 +45,118 @@ public FindResources(CPos avoidCell)
...
+				self.QueueActivity(mobile.MoveTo(moveTo, 1));
+				self.SetTargetLine(Target.FromCell(self.World, moveTo), Color.Gray, false);
+
+				var randFrames = 125 + self.World.SharedRandom.Next(-35, 35);
+				if (NextActivity != null)
+					return Util.SequenceActivities(NextActivity, new Wait(randFrames), this);
 				else
```
obrakmann: If I'm not mistaken, the `else` branch will never be taken since you're queuing up a `Move` activity just a few lines above.

OpenRA.Mods.Common/Activities/FindResources.cs
```diff
@@ -34,103 +45,118 @@ public FindResources(CPos avoidCell)
...
+					return Util.SequenceActivities(new Wait(randFrames), this);
 			}
 
-			// Attempt to claim a resource as ours:
+			// Attempt to claim a resource as ours
 			if (territory != null)
 			{
-				if (!territory.ClaimResource(self, path[0]))
-					return Util.SequenceActivities(new Wait(25), new FindResources());
+				if (!territory.ClaimResource(self, closestHarvestablePosition.Value))
```
obrakmann: Don't you need to add a check whether `closestHarvestablePosition` has a value here? Or make it the `else` branch of the previous `if` block.

OpenRA.Mods.Common/Pathfinder/PathFinderCacheDecorator.cs
```diff
@@ -68,19 +68,7 @@ public List<CPos> FindUnitPathToRange(CPos source, SubCell srcSub, WPos target,
...
-				if (cachedPath != null)
-					return cachedPath;
-
-				var pb = pathFinder.FindPath(search);
-
-				cacheStorage.Store(key, pb);
-
-				return pb;
-			}
+				return pathFinder.FindPath(search);
```
obrakmann: Why are you removing all the caching?

OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -19,10 +19,13 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
-	public class HarvesterInfo : ITraitInfo
+	public class HarvesterInfo : ITraitInfo, Requires<MobileInfo>
 	{
 		public readonly string[] DeliveryBuildings = { };
 
+		[Desc("How long to wait until (re-)checking for a nearby DeliveryBuilding if not yet assigned to one.")]
```
obrakmann: Please mention that the value is in ticks

OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -174,23 +181,18 @@ public void UnblockRefinery(Actor self)
 				if (self.Location == deliveryLoc)
 				{
 					// Get out of the way:
-					var mobile = self.Trait<Mobile>();
-					var harv = self.Trait<Harvester>();
-
-					var moveTo = harv.LastHarvestedCell ?? (deliveryLoc + new CVec(0, 4));
-					self.QueueActivity(mobile.MoveTo(moveTo, 1));
-					self.SetTargetLine(Target.FromCell(self.World, moveTo), Color.Gray, false);
+					var moveTo = LastHarvestedCell ?? (deliveryLoc + new CVec(0, 4));
```
obrakmann: Same comment as above about the hard-coded offset

OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -301,52 +302,46 @@ public void ResolveOrder(Actor self, Order order)
...
 
-					self.QueueActivity(mobile.MoveTo(loc.Value, 0));
-					self.SetTargetLine(Target.FromCell(self.World, loc.Value), Color.Red);
+				self.QueueActivity(new FindResources());
+				self.SetTargetLine(Target.FromCell(self.World, loc.Value), Color.Red);
 
-					LastOrderLocation = loc;
-				}
+				var notify = self.TraitsImplementing<INotifyHarvesterAction>();
+				var next = new FindResources();
```
obrakmann: Please move this up a bit and use `self.QueueActivity(next)`, like you do in `UnblockRefinery`.

reaperrr:
Rebased and updated. Kept each step I took in separate commits for now, will squash them when this gets thumb ups.

reaperrr:
One issue I came across myself: It appears that FindResources occasionally runs into an infinite loop within a tick, effectively freezing the game for good. That's a showstopper, we need to find and fix that bug before we can consider merging this.

reaperrr:
> [00:37] <+antares79> reaperrr: ClosestHarvestablePos seems to return self.Location during the freeze (even though searchFromLoc is not self.Location)

reaperrr:
Updated.

OpenRA.Mods.Common/Activities/FindResources.cs
```diff
@@ -21,171 +20,153 @@ namespace OpenRA.Mods.Common.Activities
 {
 	public class FindResources : Activity
 	{
+		readonly Harvester harv;
+		readonly HarvesterInfo harvInfo;
+		readonly Mobile mobile;
+		readonly MobileInfo mobileInfo;
+		ResourceLayer resLayer;
+		ResourceClaimLayer territory;
+		IPathFinder pathFinder;
```
obrakmann: These three can be readonly as well

OpenRA.Mods.Common/Activities/HarvestResource.cs
```diff
@@ -0,0 +1,75 @@
...
+			{
+				if (territory != null)
+					territory.UnclaimByActor(self);
+				return NextActivity;
+			}
+
+			// Turn to one of the harvestable facings
+			if (harvInfo.HarvestFacings != 0)
+			{
+				var facing = self.Trait<IFacing>().Facing;
```
obrakmann: I guess this trait lookup could be cached.

obrakmann:
I noticed that in d2k, harvesters eventually stop harvesting (at least when carryalls are used)

reaperrr:
Rebased.

Note: I also squashed to 4 original commits together. Commits 1+2 had issues that commit 3 fixed, and commit 4 was mostly about cosmetic stuff and caching trait lookups/activities. I believe in this particular case keeping them separate had no adavantage anymore and would only have made rebasing even harder.

abcdefg30:
Can't really comment on the code. But ingame was ok. :+1: / :white_check_mark:

pchote:
Have the infinite loop and d2k issues been fixed yet?

abcdefg30:
I didn't encounter any infinite loops, but the bug in d2k is not fixed, my bad.

reaperrr:
The loop issue should be fixed by the 2nd commit (thanks to @obrakmann), but I haven't figured out what causes the d2k issue yet.

OpenRA.Mods.Common/Activities/FindResources.cs
```diff
@@ -58,91 +57,111 @@ public override Activity Tick(Actor self)
...
+				foreach (var n in notify)
+					n.MovingToResources(self, closestHarvestablePosition.Value, next);
+
+				return Util.SequenceActivities(mobile.MoveTo(closestHarvestablePosition.Value, 1), new HarvestResource(self), next);
 			}
+		}
+
+		bool IsHarvestable(Actor self, CPos pos)
+		{
+			if (pos == self.Location)
```
pchote: The harvester actively harvests from the cell that it is in, so returning false unconditionally isn't right.  This was the same kind of situation that caused units to blow up if they were standing too close to a bridge that is destroyed or repaired (checking if they can enter the cell they are already in returns false, so the bridge thinks they fell in the water and kills them).

I wouldn't be surprised if this is causing your d2k carryall issue.

reaperrr:
Updated.

I changed the fix.
Rather than always considering the own position non-harvestable (problematic when dropped by carryall), it now first checks if `self.Location` is harvestable, and if it is, skips the path search completely and directly returns `self.Location` as `ClosestHarvestablePos`. I still saw a D2k harvester stop once (enabled debug options indicate that it ran into some kind of infinite loop as well, as the target point kept flickering on its own cell), but it doesn't crash and d2k harvesters appear to get stuck far less. This needs confirmation form other testers, of course.

reaperrr:
After some more testing, my fix doesn't really do anything different, so d2k harvesters still get stuck.

What I can say for sure now is that the issue - both for the freeze and the stopping d2k harvesters - is that the `ClosestHarvestablePos` check is looped infinitely in the same tick, so the problem is not in the pathfinder, but rather how and when `FindResources` checks `ClosestHarvestablePos`.

reaperrr:
Updated, managed to fix (or at least work around) the d2k issue. Fix is a bit hacky, but in my opinion the performance improvement justifies a little hackyness.

Promoting back to Next Release, would go well together with the AI fixes.

reaperrr:
Updated.

obrakmann:
Looks fine to me now. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7cb7bcf96078e2ccf07d1125c11f80b11602eaf7%5E...7cb7bcf96078e2ccf07d1125c11f80b11602eaf7)

PR #8625, opened 2015-07-03
Refactor TypeDictionary

RoosterDragon:
Fix a couple of potential bugs:
- When removing entries, ensure the dictionary entry is removed is the list is emptied, to prevent incorrect lookups later.
- Prevent NRE when calling GetOrDefault<T> where T is a value type and the default is returned.

General code cleanup:
- Use a single dictionary for simplicity and improved lookup performance (since there is only one dictionary to check).
- De-duplicate code in Get and GetOrDefault.

### DIFF
--- a/OpenRA.Game/Primitives/TypeDictionary.cs
+++ b/OpenRA.Game/Primitives/TypeDictionary.cs
@@ -17,8 +17,8 @@ namespace OpenRA.Primitives
 {
 	public class TypeDictionary : IEnumerable
 	{
-		Dictionary<Type, object> dataSingular = new Dictionary<Type, object>();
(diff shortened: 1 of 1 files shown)


Mailaender:
Can't spot regressions in-game. :white_check_mark:

penev92:
Code looks good and game still works :+1: 

I have no idea how to changelog this though.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/0195c258a9e1d1a8b51f78a8a51755f8dc49f4f5...ffd18a21f6dbdcddc30daacea320c4d03d216291)

PR #8629, opened 2015-07-03
Touch up Server.cs

penev92:
The first commit contains what I believe is important logging on error (it is very annoying when you try to start a skirmish game and you keep getting kicked from the server as soon as you try to connect, with _no_ message and _no_ logs).
The second commit contains exactly what its name says. If anyone has any requests for something else in that file that they would like to see beautified, now is the time.

### DIFF
--- a/OpenRA.Game/Server/Server.cs
+++ b/OpenRA.Game/Server/Server.cs
@@ -10,6 +10,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Net;
@@ -24,7 +25,7 @@ using XTimer = System.Timers.Timer;
 
 namespace OpenRA.Server
 {
-	public enum ServerState : int
+	public enum ServerState
 	{
 		WaitingPlayers = 1,
 		GameStarted = 2,
@@ -35,8 +36,6 @@ namespace OpenRA.Server
 	{
 		public readonly IPAddress Ip;
 		public readonly int Port;
-
-		int randomSeed;
 		public readonly MersenneTwister Random = new MersenneTwister();
 
 		// Valid player connections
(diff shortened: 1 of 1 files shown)


Mailaender:
:+1:

OpenRA.Game/Server/Server.cs
```diff
@@ -362,34 +374,36 @@ void ValidateClient(Connection newConn, string data)
...
-					"and may desynchronize the game state if they have incompatible rules.");
+						"and may desynchronize the game state if they have incompatible rules.");
 
 				SetOrderLag();
 			}
-			catch (Exception) { DropClient(newConn); }
+			catch (Exception ex)
+			{
+				Log.Write("server", "Dropping connection {0} because an error occured:{1}{2}",
+					newConn.Socket.RemoteEndPoint, Environment.NewLine, ex);
```
obrakmann: Log messages should not contain new lines, IMHO.

obrakmann:
Looks ok otherwise, :+1: . For the log, I suggest two messages:
```
Log.Write("server", "Dropping connection {0} because an error occurred. See exception.log for the details.", newConn.Socket.RemoteEndPoint)
Log.Write("exception", ex)
```
(not sure if the syntax works on that second line)

penev92:
Changed.

penev92:
Changed again.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/6d6bb2076c24ecbdf0ab497f9845d7033ca1d1a2%5E...6d6bb2076c24ecbdf0ab497f9845d7033ca1d1a2)

OpenRA.Game/Server/Server.cs
```diff
@@ -673,13 +674,10 @@ public void StartGame()
 				DropClient(c);
 
 			// Drop any players who are not ready
-			foreach (var c in Conns.ToArray())
+			foreach (var c in Conns.Where(c => GetClient(c).IsInvalid))
```
RoosterDragon: This will break when ```DropClient``` removes somebody from the ```Conns``` list!

PR #8515, opened 2015-06-20
Cleaned RA yaml rules

reaperrr:
Sanitizes defaults and removes some duplication.

PR #8634, opened 2015-07-04
Fix issues with civilian infantry in RA

obrakmann:
Introduced in #8515.

### DIFF
--- a/mods/ra/maps/allies-01/map.yaml
+++ b/mods/ra/maps/allies-01/map.yaml
@@ -542,6 +542,9 @@ Rules:
 	^CivInfantry:
 		RevealsShroud:
 			Range: 0c0
+	C8:
+		Inherits@2: ^ArmedCivilian
+		AutoTarget:
 	JEEP:
 		Cargo:
 			Types: Infantry, Einstein
--- a/mods/ra/rules/defaults.yaml
+++ b/mods/ra/rules/defaults.yaml
@@ -248,6 +248,8 @@
 	Armament:
 		Weapon: Pistol
 	AttackFrontal:
+	WithInfantryBody:
+		AttackSequence: shoot
 
 ^Ship:
 	Inherits@1: ^ExistsInWorld


mods/ra/maps/allies-01/map.yaml
```diff
@@ -539,9 +539,12 @@ Rules:
 			GenericVisibility: Enemy, Ally, Neutral
 			GenericStancePrefix: false
 			ShowOwnerRow: false
-	^CivInfantry:
+	^ArmedCivilian:
 		RevealsShroud:
 			Range: 0c0
+		AutoTarget:
```
abcdefg30: Any reason why you put this on `^ArmedCivlian` and not on the actor rules itself as [#8515(diff)](https://github.com/OpenRA/OpenRA/pull/8515/files#diff-0e19aa72642d1c8ffcf6cc9c47e4471cR36) did?

mods/ra/maps/allies-01/map.yaml
```diff
@@ -539,9 +539,12 @@ Rules:
 			GenericVisibility: Enemy, Ally, Neutral
 			GenericStancePrefix: false
 			ShowOwnerRow: false
-	^CivInfantry:
+	^ArmedCivilian:
 		RevealsShroud:
 			Range: 0c0
```
abcdefg30: This doesn't work anymore.
civreveal

mods/ra/maps/allies-01/map.yaml
obrakmann: Because the second civilian is a `C7` (which already is an `^ArmedCivilian`), so I only need to have that line once ;)

abcdefg30: `C7` has already `AutoTarget`: [#8515(diff)](https://github.com/OpenRA/OpenRA/pull/8515/files#diff-0e19aa72642d1c8ffcf6cc9c47e4471cR36)

obrakmann:
Whoops, fixed.

abcdefg30:
Although `Range: 0c0` on `RevealsShroud` causes weird stuff, like Healthbars in the nowhere, or civilians visible under Fog of War, it doesn't look like we can fix that on the yaml level, so :+1: / :white_check_mark:

Mailaender:
Thanks for the quick fix.

PR #8635, opened 2015-07-04
Move Warheads from the engine to Mods.Common

penev92:
Introduces `IWarhead` to enable moving `Warhead` and `DamageWarhead` outside of the engine and into `OpenRA.Mods.Common.Warheads`, where they belong.
Split into several commits to make reviewing easier.
This is the last step before `Armor` can be moved to `Mods.Common` and made upgradable.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Explodes.cs
+++ b/OpenRA.Mods.Common/Traits/Explodes.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System.Linq;
+using OpenRA.Mods.Common.Warheads;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
@@ -40,7 +41,8 @@ namespace OpenRA.Mods.Common.Traits
 			if (self.World.SharedRandom.Next(100) > info.Chance)
 				return;
 
-			if (info.DeathType != null && e.Warhead != null && !info.DeathType.Intersect(e.Warhead.DamageTypes).Any())
+			var warhead = e.Warhead as DamageWarhead;
+			if (info.DeathType != null && warhead != null && !info.DeathType.Intersect(warhead.DamageTypes).Any())
 				return;
 
 			var weaponName = ChooseWeaponForExplosion(self);
--- a/OpenRA.Mods.Common/Traits/Sound/DeathSounds.cs
+++ b/OpenRA.Mods.Common/Traits/Sound/DeathSounds.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System.Linq;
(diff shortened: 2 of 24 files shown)


OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -40,7 +41,8 @@ public void Killed(Actor self, AttackInfo e)
 			if (self.World.SharedRandom.Next(100) > info.Chance)
 				return;
 
-			if (info.DeathType != null && e.Warhead != null && !info.DeathType.Intersect(e.Warhead.DamageTypes).Any())
+			var warhead = e.Warhead as DamageWarhead;
+			if (info.DeathType != null && e.Warhead != null && !info.DeathType.Intersect(warhead.DamageTypes).Any())
```
pchote: This (and the other code like it) need to be checking `warhead != null` to avoid a NRE if `e.Warhead` isn't a DamageWarhead.

OpenRA.Mods.Common/Traits/Sound/DeathSounds.cs
```diff
@@ -38,10 +39,11 @@ public class DeathSounds : INotifyKilled
 		public void Killed(Actor self, AttackInfo e)
 		{
 			// Killed by some non-standard means
-			if (e.Warhead == null)
+			if (e.Warhead == null || !(e.Warhead is DamageWarhead))
 				return;
 
-			if (info.DeathTypes.Intersect(e.Warhead.DamageTypes).Any())
+			var warhead = e.Warhead as DamageWarhead;
```
pchote: Move this above the above check, then do `warhead == null`.

OpenRA.Mods.Common/Traits/Explodes.cs
penev92: This is the warhead that killed the actor, so how can it not have dealt damage?

pchote: "this won't happen in practice" is no excuse for incorrect code :)

Mailaender:
Everything still shoots, gets damaged, spawns visceroids and deteriorate without concrete. :white_check_mark:

abcdefg30:
Needs a rebase.

penev92:
Rebased.

reaperrr:
Code looks good to me and warheads still work (including special cases like spawning visceroid and changing owner).
:+1:

reaperrr:
changelog:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/00686635c8d67d6ecb514f651735d0bb81bc0026

PR #8636, opened 2015-07-04
Added the in-game IRC client again

Mailaender:
This time using http://smartirc4net.meebey.net by @Meebey.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -151,6 +151,9 @@ Motmans and distributed under the MIT license.
 Using ICSharpCode.SharpZipLib initially by Mike
 Krueger and distributed under the GNU GPL terms.
 
+Using SmartIrc4Net developed by Mirco Bauer
+distributed under the LGPL version 2.1 or later.
+
 
 Finally, special thanks goes to the original teams
 at Westwood Studios and EA for creating the classic
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -288,6 +288,17 @@ namespace OpenRA
 		}
 	}
 
+	public class IrcSettings
+	{
+		public string[] Hostname = { "irc.openra.net" };
+		public int Port = 6667;
+		public string Channel = "lobby";
+		public string Nickname = "Newbie";
+		public string QuitMessage = "Battle control terminated!";
+		public string TimestampFormat = "HH:mm";
+		public bool ConnectAutomatically = false;
+	}
+
 	public class Settings
 	{
 		string settingsFile;
@@ -299,6 +310,7 @@ namespace OpenRA
 		public ServerSettings Server = new ServerSettings();
 		public DebugSettings Debug = new DebugSettings();
 		public KeySettings Keys = new KeySettings();
+		public IrcSettings Irc = new IrcSettings();
 
 		public Dictionary<string, object> Sections;
 
@@ -314,6 +326,7 @@ namespace OpenRA
 				{ "Server", Server },
 				{ "Debug", Debug },
 				{ "Keys", Keys },
+				{ "Irc", Irc }
 			};
 
 			// Override fieldloader to ignore invalid entries
--- a/OpenRA.Game/Support/Log.cs
+++ b/OpenRA.Game/Support/Log.cs
@@ -61,6 +61,18 @@ namespace OpenRA
 				catch (IOException) { }
 		}
 
+		public static void Write(string channel, string value)
+		{
+			ChannelInfo info;
+			if (!Channels.TryGetValue(channel, out info))
+				throw new Exception("Tried logging to non-existant channel " + channel);
+
+			if (info.Writer == null)
+				return;
+
+			info.Writer.WriteLine(value);
+		}
+
 		public static void Write(string channel, string format, params object[] args)
 		{
 			ChannelInfo info;
--- a/OpenRA.Mods.Common/Irc.cs
+++ b/OpenRA.Mods.Common/Irc.cs
@@ -1,0 +1,19 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using Meebey.SmartIrc4net;
+
+namespace OpenRA.Mods.Common
+{
+	public static class Irc
+	{
+		public static volatile IrcClient Client;
+	}
+}
--- a/OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
@@ -1,0 +1,374 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using Meebey.SmartIrc4net;
+using OpenRA.Widgets;
+
+namespace OpenRA.Mods.Common.Widgets.Logic
+{
+	class IrcLogic
+	{
+		readonly TextFieldWidget inputBox;
+		readonly TextFieldWidget nicknameBox;
+		readonly Widget connectBG;
+		readonly Widget ircContainer;
+
+		readonly ScrollPanelWidget historyPanel;
+		readonly LabelWidget historyTemplate;
+
+		readonly ScrollPanelWidget nicknamePanel;
+		readonly LabelWidget nicknameTemplate;
+
+		bool pingSent;
+		Channel channel;
+
+		[ObjectCreator.UseCtor]
+		public IrcLogic(Widget widget)
+		{
+			Log.AddChannel("irc", "irc.log");
+
+			historyPanel = widget.Get<ScrollPanelWidget>("HISTORY_PANEL");
+			historyTemplate = widget.Get<LabelWidget>("HISTORY_TEMPLATE");
+			nicknamePanel = widget.Get<ScrollPanelWidget>("NICKNAME_PANEL");
+			nicknameTemplate = widget.Get<LabelWidget>("NICKNAME_TEMPLATE");
+
+			inputBox = widget.Get<TextFieldWidget>("INPUT_BOX");
+			inputBox.OnEnterKey = EnterPressed;
+			inputBox.IsDisabled = () => Irc.Client == null;
+
+			if (Game.Settings.Irc.Nickname == new IrcSettings().Nickname)
+				Game.Settings.Irc.Nickname += Game.CosmeticRandom.Next(100, 999);
+
+			nicknameBox = widget.Get<TextFieldWidget>("NICKNAME_BOX");
+			nicknameBox.Text = SanitizedName(Game.Settings.Irc.Nickname);
+			nicknameBox.OnTextEdited = () =>
+			{
+				nicknameBox.Text = SanitizedName(nicknameBox.Text);
+				Game.Settings.Irc.Nickname = nicknameBox.Text;
+				Game.Settings.Save();
+			};
+
+			connectBG = widget.Get("IRC_CONNECT_BG");
+			ircContainer = widget.Get("IRC_CONTAINER");
+
+			var disconnectButton = widget.Get<ButtonWidget>("DISCONNECT_BUTTON");
+			disconnectButton.IsDisabled = () => Irc.Client == null;
+			disconnectButton.OnClick = Disconnect;
+
+			MaybeShowConnectPanel();
+		}
+
+		static string SanitizedName(string dirty)
+		{
+			if (string.IsNullOrEmpty(dirty))
+				return null;
+
+			// TODO: some special chars are allowed as well, but not at every position
+			var clean = new string(dirty.Where(c => char.IsLetterOrDigit(c)).ToArray());
+
+			if (string.IsNullOrEmpty(clean))
+				return null;
+
+			if (char.IsDigit(clean[0]))
+				return SanitizedName(clean.Substring(1));
+
+			// Source: https://tools.ietf.org/html/rfc2812#section-1.2.1
+			if (clean.Length > 9)
+				clean = clean.Substring(0, 9);
+
+			return clean;
+		}
+
+		void MaybeShowConnectPanel()
+		{
+			if (Irc.Client != null && Irc.Client.IsConnected)
+			{
+				ircContainer.Visible = true;
+				connectBG.Visible = false;
+
+				Initialize();
+
+				if (Irc.Client.JoinedChannels.Count > 0)
+					channel = Irc.Client.GetChannel(Irc.Client.JoinedChannels[0]);
+
+				SyncNicknamePanel();
+
+				return;
+			}
+
+			if (Game.Settings.Irc.ConnectAutomatically)
+			{
+				ircContainer.Visible = true;
+				connectBG.Visible = false;
+				Connect();
+				return;
+			}
+
+			ircContainer.Visible = false;
+			connectBG.Visible = true;
+
+			var connectAutomaticallyCheckBox = connectBG.Get<CheckboxWidget>("CONNECT_AUTOMATICALLY_CHECKBOX");
+			connectAutomaticallyCheckBox.IsChecked = () => Game.Settings.Irc.ConnectAutomatically;
+			connectAutomaticallyCheckBox.OnClick = () => Game.Settings.Irc.ConnectAutomatically ^= true;
+
+			var connectButton = connectBG.Get<ButtonWidget>("CONNECT_BUTTON");
+			connectButton.IsDisabled = () => string.IsNullOrEmpty(nicknameBox.Text);
+			connectButton.OnClick = () =>
+			{
+				ircContainer.Visible = true;
+				connectBG.Visible = false;
+
+				Game.Settings.Irc.ConnectAutomatically = connectAutomaticallyCheckBox.IsChecked();
+				Game.Settings.Save();
+				Connect();
+			};
+		}
+
+		void Initialize()
+		{
+			Irc.Client.OnConnected += OnConnected;
+			Irc.Client.OnError += OnError;
+			Irc.Client.OnRawMessage += OnRawMessage;
+			Irc.Client.OnJoin += OnJoin;
+			Irc.Client.OnChannelActiveSynced += OnChannelActiveSynced;
+			Irc.Client.OnNickChange += OnNickChange;
+			Irc.Client.OnPart += OnPart;
+			Irc.Client.OnQuit += OnQuit;
+			Irc.Client.OnChannelMessage += OnChannelMessage;
+			Irc.Client.OnPong += OnPong;
+		}
+
+		void Connect()
+		{
+			Irc.Client = new IrcClient();
+			Irc.Client.Encoding = System.Text.Encoding.UTF8;
+			Irc.Client.SendDelay = 100;
+			Irc.Client.ActiveChannelSyncing = true;
+
+			Initialize();
+
+			Game.OnQuit += Disconnect;
+
+			try
+			{
+				AddChatLine("Connecting to {0}...".F(Game.Settings.Irc.Hostname));
+				Irc.Client.Connect(Game.Settings.Irc.Hostname, Game.Settings.Irc.Port);
+			}
+			catch (Exception e)
+			{
+				AddChatLine("Connection error: {0}".F(e.Message));
+				Game.RunAfterTick(() =>
+				{
+					Log.Write("irc", e.ToString());
+				});
+			}
+
+			new Thread(Irc.Client.Listen) { Name = "IrcListenThread" }.Start();
+		}
+
+		void OnPong(object sender, PongEventArgs e)
+		{
+			if (pingSent)
+			{
+				AddChatLine("PONG recieved after {0} ms.".F(e.Lag.Milliseconds));
+				pingSent = false;
+			}
+			else
+			{
+				Game.RunAfterTick(() =>
+				{
+					Log.Write("irc", "PONG sent after {0} ms.".F(e.Lag.Milliseconds));
+				});
+			}
+		}
+
+		Widget MakeLabelWidget(LabelWidget template, string item)
+		{
+			var widget = (LabelWidget)template.Clone();
+			var font = Game.Renderer.Fonts[widget.Font];
+			item = WidgetUtils.WrapText(item, widget.Bounds.Width, font);
+			widget.Bounds.Height = font.Measure(item).Y;
+			widget.GetText = () => item;
+			widget.Id = item;
+			return widget;
+		}
+
+		void AddChatLine(string text)
+		{
+			Game.RunAfterTick(() =>
+			{
+				Log.Write("irc", text);
+
+				var scrolledToBottom = historyPanel.ScrolledToBottom;
+
+				var newChild = MakeLabelWidget(historyTemplate, text);
+				historyPanel.AddChild(newChild);
+
+				if (scrolledToBottom)
+					historyPanel.ScrollToBottom(smooth: true);
+			});
+		}
+
+		bool EnterPressed()
+		{
+			if (inputBox.Text.Length == 0)
+				return true;
+
+			var text = inputBox.Text;
+			inputBox.Text = "";
+
+			if (text.StartsWith("/nick "))
+			{
+				var nick = text.Replace("/nick ", string.Empty);
+				if (Rfc2812.IsValidNickname(nick))
+					Irc.Client.RfcNick(nick);
+				else
+					AddChatLine("Invalid nickname.");
+			}
+			else if (text.StartsWith("/ping "))
+			{
+				Irc.Client.RfcPing(Irc.Client.GetIrcUser(text.Replace("/ping ", string.Empty)).Host);
+				pingSent = true;
+			}
+			else if (text.StartsWith("/"))
+				AddChatLine("Unknown command.");
+			else
+			{
+				AddChatLine("[{0}] <{1}> {2}".F(DateTime.Now.ToString(Game.Settings.Irc.TimestampFormat), Irc.Client.Nickname, text));
+				Irc.Client.SendMessage(SendType.Message, "#" + Game.Settings.Irc.Channel, text);
+			}
+
+			return true;
+		}
+
+		void OnConnected(object sender, EventArgs e)
+		{
+			AddChatLine("Connected.");
+
+			if (!Rfc2812.IsValidNickname(Game.Settings.Irc.Nickname))
+			{
+				AddChatLine("Invalid nickname. Can't login.");
+				return;
+			}
+
+			Irc.Client.Login(new[] { Game.Settings.Irc.Nickname }, "in-game IRC client", 0, "OpenRA");
+
+			Irc.Client.RfcJoin("#" + Game.Settings.Irc.Channel);
+		}
+
+		void OnError(object sender, ErrorEventArgs e)
+		{
+			AddChatLine("Error: " + e.ErrorMessage);
+			Game.RunAfterTick(() =>
+			{
+				Log.Write("irc", e.ToString());
+			});
+		}
+
+		void OnRawMessage(object sender, IrcEventArgs e)
+		{
+			Game.RunAfterTick(() =>
+			{
+				Log.Write("irc", e.Data.RawMessage);
+			});
+		}
+
+		void OnChannelMessage(object sender, IrcEventArgs e)
+		{
+			AddChatLine("[{0}] <{1}> {2}".F(DateTime.Now.ToString(Game.Settings.Irc.TimestampFormat), e.Data.Nick, e.Data.Message));
+		}
+
+		void OnJoin(object sender, JoinEventArgs e)
+		{
+			if (e.Who == Irc.Client.Nickname)
+				return;
+
+			AddChatLine("{0} joined channel {1}.".F(e.Who, e.Channel));
+			channel = Irc.Client.GetChannel(e.Channel);
+			SyncNicknamePanel();
+		}
+
+		void OnChannelActiveSynced(object sender, IrcEventArgs e)
+		{
+			channel = Irc.Client.GetChannel(e.Data.Channel);
+
+			AddChatLine("{0} users online".F(channel.Users.Count));
+
+			if (!string.IsNullOrEmpty(channel.Topic))
+				AddChatLine("*** Topic: {0}".F(channel.Topic));
+
+			SyncNicknamePanel();
+		}
+
+		void OnNickChange(object sender, NickChangeEventArgs e)
+		{
+			AddChatLine("{0} is now known as {1}.".F(e.OldNickname, e.NewNickname));
+			SyncNicknamePanel();
+		}
+
+		void SyncNicknamePanel()
+		{
+			if (channel == null)
+				return;
+
+			var users = channel.Users;
+			Game.RunAfterTick(() =>
+			{
+				nicknamePanel.RemoveChildren();
+
+				foreach (DictionaryEntry user in users)
+				{
+					var channeluser = (ChannelUser)user.Value;
+					var prefix = channeluser.IsOp ? "@" : channeluser.IsVoice ? "+" : "";
+					var newChild = MakeLabelWidget(nicknameTemplate, prefix + channeluser.Nick);
+					nicknamePanel.AddChild(newChild);
+				}
+			});
+		}
+
+		void OnQuit(object sender, QuitEventArgs e)
+		{
+			AddChatLine("{0} quit.".F(e.Who));
+		}
+
+		void OnPart(object sender, PartEventArgs e)
+		{
+			AddChatLine("{0} left {1}.".F(e.Who, e.Data.Channel));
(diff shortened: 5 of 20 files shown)


OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,265 @@
...
+		}
+
+		void Connect()
+		{
+			client = new IrcClient();
+			client.Encoding = System.Text.Encoding.UTF8;
+			client.SendDelay = 100;
+			client.ActiveChannelSyncing = true;
+
+			client.OnConnected += new EventHandler(OnConnected);
```
ScottNZ: This can just be `client.OnConnected += OnConnected;`
Similar for the other lines below

Mailaender: Okay. Pushed it with a bunch of other improvements.

AUTHORS
```diff
@@ -149,6 +149,9 @@ Motmans and distributed under the MIT license.
 Using ICSharpCode.SharpZipLib initially by Mike
 Krueger and distributed under the GNU GPL terms.
 
+Using SmartIrc4Net developed by Mirco Bauer
+distributed under the LGPL version 2.1.
```
meebey: version 2.1 or later. That can be important for GPLv3 compatibility.

Mailaender: Fixed.

x-a-n-a-x:
@devs, please review this PR.
A global IRC channel will help players communicate, organize games and tournaments, spread news, share maps...

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,266 @@
...
+				AddChatLine("[{0}] <{1}> {2}".F(DateTime.Now.ToString(Game.Settings.Irc.TimestampFormat), client.Nickname, text));
+				client.SendMessage(SendType.Message, "#" + Game.Settings.Irc.Channel, text);
+			}
+
+			return true;
+		}
+
+		public void OnError(object sender, ErrorEventArgs e)
+		{
+			AddChatLine("Errror: ".F(e.ErrorMessage));
```
abcdefg30: Typo: `Errror`

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,266 @@
...
+		bool EnterPressed()
+		{
+			if (!inputBox.Text.Any())
+				return true;
+
+			var text = inputBox.Text;
+			inputBox.Text = "";
+
+			if (text.StartsWith("/"))
+				AddChatLine("Unknown command");
```
abcdefg30: If we don't want users to perform commands, why not print something like
`You're not allowed to perform commands/this command!` out?

abcdefg30:
We need to do something about player names too:
Your name can be too long:
ircnamelength
Or too short (empty):
ircemptyname
Latter also crashes when disconnecting:
```
Tiberian Sun Mod at Version {DEV_VERSION}
Exception of type `Meebey.SmartIrc4net.NotConnectedException`: The connection could not be disconnected because there is no active connection
   at Meebey.SmartIrc4net.IrcConnection.Disconnect()
   at OpenRA.Mods.Common.Widgets.Logic.IrcLogic.Disconnect()
   at OpenRA.Mods.Common.Widgets.ButtonWidget.<.ctor>b__d(MouseInput _)
   at OpenRA.Mods.Common.Widgets.ButtonWidget.HandleMouseInput(MouseInput mi)
   at OpenRA.Widgets.Widget.HandleMouseInputOuter(MouseInput mi) in OpenRA\OpenRA.Game\Widgets\Widget.cs:Line 354.
   at OpenRA.Widgets.Ui.HandleInput(MouseInput mi) in OpenRA\OpenRA.Game\Widgets\Widget.cs:Line 87.
   at OpenRA.DefaultInputHandler.<>c__DisplayClass7.<OnMouseInput>b__6() in OpenRA\OpenRA.Game\Input\InputHandler.cs:Line 49.
   at OpenRA.Sync.CheckSyncUnchanged[T](World world, Func`1 fn) in OpenRA\OpenRA.Game\Sync.cs:Line 197.
   at OpenRA.DefaultInputHandler.OnMouseInput(MouseInput input) in OpenRA\OpenRA.Game\Input\InputHandler.cs:Line 49.
   at OpenRA.Renderer.Sdl2.Sdl2Input.PumpInput(IInputHandler inputHandler) in OpenRA\OpenRA.Renderer.Sdl2\Sdl2Input.cs:Line 105.
   at OpenRA.Renderer.Sdl2.Sdl2GraphicsDevice.PumpInput(IInputHandler inputHandler) in OpenRA\OpenRA.Renderer.Sdl2\Sdl2GraphicsDevice.cs:Line 359.
   at OpenRA.Graphics.Renderer.EndFrame(IInputHandler inputHandler) in OpenRA\OpenRA.Game\Graphics\Renderer.cs:Line 153.
   at OpenRA.Game.RenderTick() in OpenRA\OpenRA.Game\Game.cs:Line 546.
   at OpenRA.Game.Loop() in OpenRA\OpenRA.Game\Game.cs:Line 653.
   at OpenRA.Game.Run() in OpenRA\OpenRA.Game\Game.cs:Line 671.
   at OpenRA.Program.Run(String[] args) in OpenRA\OpenRA.Game\Support\Program.cs:Line 111.
   at OpenRA.Program.Main(String[] args) in OpenRA\OpenRA.Game\Support\Program.cs:Line 39.

**Comment by GraionDilach** [comment]
It also throws no message if you try it with a playername containing invalid chars, e. g. space. It even shows an empty channel then, and the player won't pop up in the actual channel.

File: OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs [hunk 5]
```diff
@@ -0,0 +1,266 @@
...
+				var newChild = MakeLabelWidget(nicknameTemplate, prefix + channeluser.Nick);
+				nicknamePanel.AddChild(newChild);
+			}
+		}
+
+		public void OnConnected(object sender, EventArgs e)
+		{
+			AddChatLine("Connected.");
+
+			AddChatLine("Loggin in...");
```
ScottNZ: As I mentioned previously, this message is unnecessary (and also spelled incorrectly fwiw)

Mailaender:
Added nickname changing and sanity checks.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,305 @@
...
+			return true;
+		}
+
+		public void OnError(object sender, ErrorEventArgs e)
+		{
+			AddChatLine("Error: ".F(e.ErrorMessage));
+			Log.Write("irc", "Error: {0}".F(e.ErrorMessage));
+		}
+
+		public void OnRawMessage(object sender, IrcEventArgs e)
```
Unit158: No ping or VERSION?

Mailaender: The IrcClient from the library automatically answers those.

Mailaender:
Added a `/ping` command.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,316 @@
...
+				AddChatLine("Connecting to {0}...".F(Game.Settings.Irc.Hostname));
+				client.Connect(Game.Settings.Irc.Hostname, Game.Settings.Irc.Port);
+			}
+			catch (Exception e)
+			{
+				AddChatLine("Connection error: {0}".F(e.Message));
+				Log.Write("irc", e.ToString());
+			}
+
+			listenThread = new Thread(new ThreadStart(Listen));
```
RoosterDragon: Can you give the thread a name - I find them handy when distinguishing between threads during debugging.

OpenRA.Mods.Common/Widgets/ScrollPanelWidget.cs
```diff
@@ -167,7 +167,7 @@ public override void DrawOuter()
 			Game.Renderer.EnableScissor(drawBounds);
 
 			drawBounds.Offset((-ChildOrigin).ToPoint());
-			foreach (var child in Children)
+			foreach (var child in Children.ToArray())
```
RoosterDragon: Did you add this because the event handlers are being raised on the listening thread and not the UI thread? If so I think you'll to better to invoke UI code on the UI thread or you will likely run into other threading nastiness.

Mailaender: Hmm, I tried adding `Game.RunAfterTick(() => { });` everywhere, but that kinda had an opposite effect.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
Mailaender: Gave it a thread name.

OpenRA.Mods.Common/Widgets/ScrollPanelWidget.cs
pchote: Manipulating the UI from an external thread is a very bad idea, and will (and has in the past) cause(d) a bunch of hard to find regressions.

abcdefg30:
The game freezes when sending `{0}` or when logging in with an invalid nickname and clicking on `Disconnect` after that.

Mailaender:
Fixed the remaining exploits and the UI threading issues. Removing my workarounds also improved responsiveness a lot.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,346 @@
...
+
+		public void OnError(object sender, ErrorEventArgs e)
+		{
+			AddChatLine("Error: ".F(e.ErrorMessage));
+			Log.Write("irc", "Error: {0}".F(e.ErrorMessage));
+		}
+
+		public void OnRawMessage(object sender, IrcEventArgs e)
+		{
+			Log.Write("irc", e.Data.RawMessage.Replace("{", string.Empty).Replace("}", string.Empty));
```
pchote: You should escape these instead of just removing them.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,346 @@
...
+	class IrcLogic
+	{
+		IrcClient client;
+
+		TextFieldWidget inputBox;
+		TextFieldWidget nicknameBox;
+		Widget connectBG;
+		Widget ircContainer;
+
+		ScrollPanelWidget historyPanel;
```
pchote: Most of these should be readonly.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,346 @@
...
+			var font = Game.Renderer.Fonts[widget.Font];
+			item = WidgetUtils.WrapText(item, widget.Bounds.Width, font);
+			widget.Bounds.Height = font.Measure(item).Y;
+			widget.GetText = () => item;
+			return widget;
+		}
+
+		void AddChatLine(string text)
+		{
+			Log.Write("irc", text.Replace("{", string.Empty).Replace("}", string.Empty));
```
pchote: You should escape these instead of removing them.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,346 @@
...
+			AddChatLine("Error: ".F(e.ErrorMessage));
+			Log.Write("irc", "Error: {0}".F(e.ErrorMessage));
+		}
+
+		public void OnRawMessage(object sender, IrcEventArgs e)
+		{
+			Log.Write("irc", e.Data.RawMessage.Replace("{", string.Empty).Replace("}", string.Empty));
+
+			if (e.Data.Type == ReceiveType.ChannelMessage)
+				Game.RunAfterDelay(client.Lag.Milliseconds * 5, () =>
```
pchote: What is the significance of this delay?

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,346 @@
...
+		{
+			if (channel == null)
+				return;
+
+			var users = channel.Users;
+			Game.RunAfterTick(() =>
+			{
+				nicknamePanel.RemoveChildren();
+
+				foreach (DictionaryEntry user in users)
```
pchote: `var`?

Mailaender: Not sure. This is casting form a hash table.

OpenRA.Game/Settings.cs
```diff
@@ -288,6 +288,19 @@ Hotkey GetKey(Func<KeySettings, Hotkey>[] keys, int index)
 		}
 	}
 
+	public class IrcSettings
+	{
+		public string[] Hostname = { "irc.openra.net" };
+		public int Port = 6667;
+		public string Nickname = "Newbie";
+		public string Username = "OpenRA";
```
Phrohdoh: Do we have a reason to actually expose both of these?
Changing nickname in IRC (if possible) ought to change username, which shouldn't matter at all to players / people chatting.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,346 @@
...
+				else
+					AddChatLine("Invalid nickname.");
+			}
+			else if (text.StartsWith("/ping "))
+			{
+				client.RfcPing(client.GetIrcUser(text.Replace("/ping ", string.Empty)).Host);
+				pingSent = true;
+			}
+			else if (text.StartsWith("/"))
+				AddChatLine("Unknown command");
```
Phrohdoh: This is a bit bogus.
What if someone wants to tell someone else to use `/nick foo`?
There's no reason we should respond with `Unknown command` if we don't have a list of commands to check against.

Mailaender: The list of supported commands is in the statements above and doesn't preovent someone from texting you "use /nick to change your name".

Phrohdoh: Certainly not, but my point was `/nick foo` as the entire message.
While not allowing that and requiring `use /nick foo` is valid I feel the message for any other `/`-prefixed message is not accurate.

OpenRA.Game/Settings.cs
Mailaender: I think you are confused by IRC terminology here. The username is part of the ident which will be "~OpenRA@host86-159-185-244.range86-159.btcentralplus.com` when you connect with the in-game IRC client. It is usually configurable in IRC clients and not automatically updated with a nickname change.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
pchote: Why do we need `/command`s at all?  Names should be changeable through the UI, and I don't see a use for ping in our global game chat.  We should universally prefix any message starting with '/' with a space.

Mailaender:
Updated with the hints from @ScottNZ and @pchote.

OpenRA.Game/Settings.cs
pchote: I agree with @Phrohdoh.  We are using IRC as the backend for our global chat, not writing a general purpose IRC client.  There's no benefit to making this configurable.

pchote: The same applies to the properties below too.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
Mailaender: It was wished by @Unit158 and we have command line support for all other chat windows and somehow everyone expects it in IRC clients even simple ones made exclusively for games like @cncnet.

OpenRA.Game/Settings.cs
Mailaender: Nobody noticed that `Username` was in fact never used here. Removed it.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
Mailaender: > We should universally prefix any message starting with '/' with a space.

That is a really bad idea if people start trying out things like:
`/register personalemail@address.com password` etc.

pchote: I very much doubt that would be a real problem.  If somebody does try it, then PEBCAK for thinking that our global chat was a general purpose IRC client.

Mailaender:
Updated again to stay connected even after closing widgets and switching mod like the original https://github.com/OpenRA/OpenRA/pull/3859.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
Unit158: This actually can't happen. We need a way to communicate with the IRC network's command as well. We dont want somebody sending '/ns IDENTIFY JoeBlo2345 SomeCornyPassword' in chat!

Phrohdoh: We don't support `/ns` as-is from my understanding.

Mailaender: So everyone wants to all `/command` gone and `/` being escaped with space?

Unit158: You're misunderstanding how IRC works. I'll explain later.

Unit158: Only if we provide a warning in big flaming letters that you can't use chat line commands.

pchote: @Unit158: Players shouldn't even *know* that it is using IRC behind the scenes.  There's no reason for them to try to use IRC commands, so no need to warn them.

Unit158: Alright, that's good enough for me. If there are any issues, I will point a finger.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,355 @@
...
+		}
+
+		void Disconnect()
+		{
+			Irc.Client.RfcQuit(Game.Settings.Irc.QuitMessage);
+			AddChatLine("Disconnecting from {0}...".F(Irc.Client.Address));
+			Irc.Client.Disconnect();
+			nicknamePanel.RemoveChildren();
+			Game.Settings.Irc.ConnectAutomatically = false;
+			Irc.ListenThread.Abort();
```
RoosterDragon: Is there a way to gracefully end the thread rather than having to resort to ```Abort```?

abcdefg30:
If I connect with an empty nickname and then disconnect, there is a huge lag/freeze.

Also crashes when I quit the game:
```
Tiberian Sun Mod at Version {DEV_VERSION}
Exception of type `Meebey.SmartIrc4net.NotConnectedException`: The connection could not be disconnected because there is no active connection    
   at Meebey.SmartIrc4net.IrcConnection.Disconnect()
   at OpenRA.Mods.Common.Widgets.Logic.IrcLogic.Disconnect()
   at System.Action.Invoke()
   at OpenRA.Game.Run() in OpenRA\OpenRA.Game\Game.cs:Line 686.
   at OpenRA.Program.Run(String[] args) in OpenRA\OpenRA.Game\Support\Program.cs:Line 111.
   at OpenRA.Program.Main(String[] args) in OpenRA\OpenRA.Game\Support\Program.cs:Line 39.

**Comment by Mailaender** [comment]
Fixed the crash due to invalid user name input, replaced all remaining raw message parsers with event handlers and tried to get the thread to self-eliminate.

**Comment by abcdefg30** [comment]
```
OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs:L150: [SA1008] Invalid spacing around the opening parenthesis.
OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs:L155: [SA1008] Invalid spacing around the opening parenthesis.
OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs:L162: [SA1008] Invalid spacing around the opening parenthesis.

Mailaender:
Fixed.

OpenRA.Game/Support/Log.cs
```diff
@@ -61,6 +61,18 @@ public static void AddChannel(string channelName, string baseFilename)
 				catch (IOException) { }
 		}
 
+		public static void Write(string channel, string format)
```
RoosterDragon: The ```format``` parameter needs to be renamed here.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,363 @@
...
+			else
+				Log.Write("irc", "PONG sent after {0} ms.".F(e.Lag.Milliseconds));
+		}
+
+		void Listen()
+		{
+			while (Irc.Client != null)
+				Irc.Client.ListenOnce();
+
+			return;
```
RoosterDragon: ```return``` is unnecessary.

Mailaender:
Updated with the suggested changes and fixed the nickname panel not being initialized when leaving the window without disconnecting.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,364 @@
...
+			}
+			catch (Exception e)
+			{
+				AddChatLine("Connection error: {0}".F(e.Message));
+				Log.Write("irc", e.ToString());
+			}
+
+			Irc.ListenThread = new Thread(new ThreadStart(Listen));
+			Irc.ListenThread.Name = "IrcListenThread";
+			Irc.ListenThread.Start();
```
RoosterDragon: You don't use the ```ListenThread``` member any more so you can remove it.

I also checked and the library with return from Listen when you call Disconnect, this means you can simply this whole chunk down to:

```new Thread(Irc.Client.Listen) { Name = "IrcListenThread" }.Start();```

File: OpenRA.Mods.Common/Irc.cs [hunk 20]
```diff
@@ -0,0 +1,21 @@
...
+#endregion
+
+using System.Threading;
+using Meebey.SmartIrc4net;
+
+namespace OpenRA.Mods.Common
+{
+	public static class Irc
+	{
+		public static IrcClient Client;
```
RoosterDragon: We'll need to access this in a thread-safe way - although the class is itself thread-safe because we sometimes do several operations in a row on it, it's state may have changed. For example, we may do a null check, find it not-null. Then it might disconnect under our feet, and when we access some property on it we find it's suddently become null again.

A static object to use in a bunch of lock statements would help with that. We probably want to make both ```Client``` and this lock object private to the ```IrcLogic``` to avoid exposing them elsewhere (somewhat nasty as that is).

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,364 @@
...
+
+			Irc.Client.Login(new[] { Game.Settings.Irc.Nickname }, "in-game IRC client", 0, "OpenRA");
+
+			Irc.Client.RfcJoin("#" + Game.Settings.Irc.Channel);
+		}
+
+		void OnError(object sender, ErrorEventArgs e)
+		{
+			AddChatLine("Error: ".F(e.ErrorMessage));
+			Log.Write("irc", "Error: {0}".F(e.ErrorMessage));
```
RoosterDragon: Our Log.Write isn't thread-safe - this is an issue with Log.Write rather than the IRC stuff but we'd still need to rectify it if we're going to start calling it from the IRC thread.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,364 @@
...
+			AddChatLine("{0} is now known as {1}.".F(e.OldNickname, e.NewNickname));
+			SyncNicknamePanel();
+		}
+
+		void SyncNicknamePanel()
+		{
+			if (channel == null)
+				return;
+
+			var users = channel.Users;
```
RoosterDragon: Is this safe to enumerate on a different thread, or do we need to take a copy first?

RoosterDragon:
I noticed the if I joined in one mod, disconnected and then set it connect automatically, that when I opened the TS mod (and it joined automatically) the user list wouldn't be updated with other users, only me.

This only seems to happen on this mod, and the list works if you disconnect and join on it usually, or if you remain connected whilst switching mods. Not sure what's up with that.

Seems to work otherwise though :)

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
Mailaender: This had a tendency to become null when done after delay on a different thread or change during enumeration so caching it seemed like the best way.

RoosterDragon: Sorry I meant: do we need to copy the whole collection in order to do this safely, rather than just caching the variable?

What happens if the IRC thread changes the collection whilst we're enumerating it, for example?

OpenRA.Mods.Common/Irc.cs
Mailaender: This doesn't seem to be a problem in practice. As I can't properly dispose it, the only way for it to become null is when I explicitly set it in `Disconnect` myself.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
Mailaender: Updated also on all other occasions.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
Mailaender: Done.

OpenRA.Mods.Common/Irc.cs
RoosterDragon: That was just an example, any operation where we check multiple things on the UI thread if they could change in the meantime might go wrong. Another example:
```
if (Irc.Client.JoinedChannels.Count > 0)
    channel = Irc.Client.GetChannel(Irc.Client.JoinedChannels[0]);
```

Let's say the IRC thread leaves the channel after our check but before we get the channel. We'd hit an exception at this point. Something needs to be done to ensure race conditions like this don't bite us.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
RoosterDragon: Nevermind - I checked the source and it clones it internally - you don't need to do anything here.

OpenRA.Mods.Common/Irc.cs
RoosterDragon: Please make this ```volatile``` so both threads definitely see the right value.

RoosterDragon: I looked through for other thread safety issues - and actually the comment I made above using the channels as an example was the only one I could see.

The library doesn't seem to provide a thread-safe way to query ```JoinedChannels``` - but in reality I doubt we'll ever actually run into this problem so I can live with it.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,377 @@
...
+			else
+			{
+				Game.RunAfterTick(() =>
+				{
+					Log.Write("irc", "PONG sent after {0} ms.".F(e.Lag.Milliseconds));
+				});
+			}
+		}
+
+		void Listen()
```
RoosterDragon: This method is unused - you can remove it.

RoosterDragon:
This looks good to me, if you can fix the two code issues I commented on inline, as well as the bug from my [other comment](https://github.com/OpenRA/OpenRA/pull/8636#issuecomment-135524404).

OpenRA.Mods.Common/Irc.cs
Mailaender: Made it `volatile`. I only need `JoinedChannels` once at first to initialize the user list when returning to the widget. In all other cases I use the channel provided by the event args.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
Mailaender: Removed.

Mailaender:
As I can't reproduce https://github.com/OpenRA/OpenRA/pull/8636#issuecomment-135524404 I assume your thread safety hints already squished that bug.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,371 @@
...
+
+		void Disconnect()
+		{
+			if (Irc.Client == null)
+				return;
+
+			Irc.Client.RfcQuit(Game.Settings.Irc.QuitMessage);
+
+			AddChatLine("Disconnecting from {0}...".F(Irc.Client.Address));
+			Irc.Client.Disconnect();
```
RoosterDragon: If you're not connected here (for example, due to an error) this will throw with "The connection could not be disconnected because there is no active connection".

I was able to recreate this by just connecting and disconnecting really fast until the IRC server threw an error at me rather than connecting me.

OpenRA.Mods.Common/Widgets/Logic/IrcLogic.cs
```diff
@@ -0,0 +1,371 @@
...
+			}
+
+			Irc.Client.Login(new[] { Game.Settings.Irc.Nickname }, "in-game IRC client", 0, "OpenRA");
+
+			Irc.Client.RfcJoin("#" + Game.Settings.Irc.Channel);
+		}
+
+		void OnError(object sender, ErrorEventArgs e)
+		{
+			AddChatLine("Error: ".F(e.ErrorMessage));
```
RoosterDragon: You're missing a ```{0}``` in your format string.

Phrohdoh: On that note, why are you using `.F` instead of direct concatenation?

RoosterDragon:
I found the problem for my issue: If the ```OnJoin``` and ```OnChannelActiveSynced``` fire within the same tick, the following happens:
- OnJoin calls SyncNicknamePanel with 1 user.
- OnChannelActiveSynced SyncNicknamePanel with all the users.

These both use Game.RunAfterTick to run stuff on the UI thread, but that doesn't preserve order correctly (see #9286). If they both get queued in the same tick then they get processed in reverse - you'll sync the list with all the users, then sync it with the one user.

OpenRA.Mods.Common/Irc.cs
```diff
@@ -0,0 +1,20 @@
...
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.Threading;
```
RoosterDragon: Unused using.

OpenRA.Mods.Common/Irc.cs
```diff
@@ -0,0 +1,20 @@
...
+using Meebey.SmartIrc4net;
+
+namespace OpenRA.Mods.Common
+{
+	public static class Irc
+	{
+		public volatile static IrcClient Client;
```
RoosterDragon: StyleCop wants ```public static volatile```.

Mailaender:
Fixed.

RoosterDragon:
:+1: Works nice.

Unit158:
I'm not a fan of the char limit of the nickname. I can't use two of my most common nicknames that would otherwise fit well with the UI.

penev92:
If one client dies, the other (I was testing with 2) gets this:

ScottNZ:
Quits aren't even associated with channels, so this is hardly surprising.

pchote:
The AddChatLine above that also passes two arguments to a one-parameter format string.

Mailaender:
Fixed. The API isn't really fool proof in that regard.

abcdefg30:
Looks good to me, :+1: / :white_check_mark: 
except that you can write messages that exceed the message box thing:
ircexceed

Mailaender:
The same thing happens to the Lobby and any other chat window. Our word wrap only works if you write proper words. WONTFIX for now, but you can make a ticket if it annoys you.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/e3a5d6c18616edd7d661d0d1d8dab947dd539095...5b92fb8bf6b3edc373a214327258614cff4419a3)

abcdefg30:
Good work!

Mailaender:
@ihptru Can you also run https://github.com/ihptru/orabot at irc.openra.net #lobby now?

ihptru:
@Mailaender added orabot to irc.openra.net #lobby

Mailaender:
Thanks! @baxxxster can you give OP to orabot and maybe automatize that?

image

baxxxster:
@Mailaender There's no services up who could do that, though, I can look into auto-op script in my irssi client

baxxxster:
Irssi: #lobby: Now auto-opping: orabot@openra-E0EEB61.ihptru.net - Can be tested by re-joining the channel

PR #8640, opened 2015-07-04
Fixed Warning: Sprite image delphi from actor DELPHI does not define sequence shoot

Mailaender:
Revealed by https://github.com/OpenRA/OpenRA/pull/8634.


penev92:
:+1:

PR #8641, opened 2015-07-04
Added exception handling to ExportCharacterSeparatedRules

### LINKED ISSUE #8546: Utility doesn't save file if user doesn't have write access
When OpenRA.Utility.exe is in a directory where the user doesn't have write access, such as C:\Program Files (x86)\, the command --generate-dps-table (and possibly others) doesn't produce an output file and the program doesn't show any errors. In this case it is necessary to open the command line with administrator privilages.

It should show a permission denied error. Or maybe put the output file in the directory where all the replays and logs are located...

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/22771694-utility-doesn-t-save-file-if-user-doesn-t-have-write-access?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

suvjunmd:
Closes #8546

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/ExportCharacterSeparatedRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/ExportCharacterSeparatedRules.cs
@@ -23,10 +23,18 @@ namespace OpenRA.Mods.Common.UtilityCommands
 			Game.ModData = modData;
 			var table = ActorStatsExport.GenerateTable();
 			var filename = "{0}-mod-dps.csv".F(Game.ModData.Manifest.Mod.Id);
-			using (var outfile = new StreamWriter(filename))
-				outfile.Write(table.ToCharacterSeparatedValues(";", true));
-			Console.WriteLine("{0} has been saved.".F(filename));
-			Console.WriteLine("Open as values separated by semicolon.");
+
+			try
+			{
+				using (var outfile = new StreamWriter(filename))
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/UtilityCommands/ExportCharacterSeparatedRules.cs
```diff
@@ -23,10 +23,18 @@ public void Run(ModData modData, string[] args)
...
-			Console.WriteLine("{0} has been saved.".F(filename));
-			Console.WriteLine("Open as values separated by semicolon.");
+
+		    try
+		    {
+                using (var outfile = new StreamWriter(filename))
+                    outfile.Write(table.ToCharacterSeparatedValues(";", true));
+                Console.WriteLine("{0} has been saved.".F(filename));
+                Console.WriteLine("Open as values separated by semicolon.");
+            }
```
penev92: You are using spaces instead of tabs.

suvjunmd:
Replaced spaces with tabs

Mailaender:
Cheat, but effective. Why not. :+1:

penev92:
:+1:

PR #8642, opened 2015-07-04
Added separate CloseSequence to WithLandingCraftAnimation

reaperrr:
Closing sequence of landing craft is no longer hardcoded to be a reversed opening sequence.

This also fixes (or let's say avoids) the visual glitch that you see when moving an open landing craft away from the shore.
The underlying problem behind the above glitch appears to be an issue with playing animations backwards, which is now no longer used by this trait.

Additionally, modified `WithAttackAnimation` and `WithMoveAnimation` to work with `WithSpriteBody` as well, not just `WithFacingSpriteBody`.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
@@ -13,7 +13,7 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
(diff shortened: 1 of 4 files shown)


Mailaender:
That looks a lot smoother in-game. :white_check_mark:

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/81ac059880d5aaadeffef8c9b30315156b2bf630...0b2d5e63222bb3ea127c954b07a6852c5cbad644)

PR #8643, opened 2015-07-04
Removed bogus Selectable from RA crate

reaperrr:
And increased bounds for tooltips etc. on shipping mods.


Mailaender:
OK. :+1:

PR #8644, opened 2015-07-05
Fixed missing braces breaking the upgrade rules

Mailaender:
This happened again. I think I already complained last time that we should not enforce the unwritten omit braces for one liners style rule which leads to problems when those conditions are later expanded. The good news is: Coverity detects these problems!


abcdefg30:
:+1:

PR #8645, opened 2015-07-05
Fixed missing null checks in FieldLoader

Mailaender:
Coverity only highlights one case, but most functions are affected that do string operations before try-parsing the value. This time I also looked at the greyed out lines to fix this defect once and for all.

### DIFF
--- a/OpenRA.Game/FieldLoader.cs
+++ b/OpenRA.Game/FieldLoader.cs
@@ -157,69 +157,81 @@ namespace OpenRA
 			else if (fieldType == typeof(float))
 			{
 				float res;
-				if (float.TryParse(value.Replace("%", ""), NumberStyles.Float, NumberFormatInfo.InvariantInfo, out res))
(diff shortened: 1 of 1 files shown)


Mailaender:
Best reviewed with https://github.com/OpenRA/OpenRA/pull/8645/files?w=1

penev92:
Seems like `if (value == null) return InvalidValueAction(value, fieldType, fieldName);` would have been a nicer solution. 
Maybe even move it to the very first line of the method?
Not something I'd like to argue over though.
:+1:

Mailaender:
I already thought about it, but some methods allow null and just return empty arrays or throw their own more specialized exception.

obrakmann:
Looks fine and still works, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/3e741235bcb1768671da2425ed6ce3c7408aa661%5E...3e741235bcb1768671da2425ed6ce3c7408aa661)

PR #8649, opened 2015-07-05
Render resources using TerrainSpriteLayers

pchote:
This changes ResourceLayer to use a set of TerrainSpriteLayers instead of generating a pile of SpriteRenderables each tick.  This should cut out a bunch of rendering work (ping @RoosterDragon) and removes a dependency on shroud / bounds that complicates the future heightmap-shroud code.

This is mainly a proof of concept: if it's viable, then followup prs will apply the same approach to `EditorResourceLayer`, `SmudgeLayer`, and `BuildableTerrainLayer`.

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/ResourceLayer.cs
+++ b/OpenRA.Mods.Common/Traits/World/ResourceLayer.cs
@@ -17,19 +17,30 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Attach this to the world actor.", "Order of the layers defines the Z sorting.")]
-	public class ResourceLayerInfo : TraitInfo<ResourceLayer>, Requires<ResourceTypeInfo> { }
+	public class ResourceLayerInfo : ITraitInfo, Requires<ResourceTypeInfo>, Requires<BuildingInfluenceInfo>
+	{
+		public virtual object Create(ActorInitializer init) { return new ResourceLayer(init.Self); }
+	}
 
 	public class ResourceLayer : IRenderOverlay, IWorldLoaded, ITickRender
 	{
 		static readonly CellContents EmptyCell = new CellContents();
 
-		World world;
+		readonly World world;
+		readonly BuildingInfluence buildingInfluence;
+		readonly List<CPos> dirty = new List<CPos>();
 
-		BuildingInfluence buildingInfluence;
+		protected readonly CellLayer<CellContents> Content;
+		protected readonly CellLayer<CellContents> RenderContent;
 
-		protected CellLayer<CellContents> content;
-		protected CellLayer<CellContents> render;
-		List<CPos> dirty;
+		public ResourceLayer(Actor self)
+		{
+			world = self.World;
+			buildingInfluence = world.WorldActor.Trait<BuildingInfluence>();
+
+			Content = new CellLayer<CellContents>(world.Map);
+			RenderContent = new CellLayer<CellContents>(world.Map);
+		}
 
 		public void Render(WorldRenderer wr)
 		{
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.Common/Traits/World/ResourceLayer.cs
```diff
@@ -99,15 +109,45 @@ public void WorldLoaded(World w, WorldRenderer wr)
...
+				}
+				else
+				{
+					layer = new TerrainSpriteLayer(w, wr, sheet, blendMode, pal, wr.World.Type != WorldType.Editor);
+					spriteLayers.Add(pal, layer);
 				}
+
+				// Validate that sprites are compatible with this layer
+				if (r.Value.Variants.Any(kv => kv.Value.Any(s => s.Sheet != sheet)))
+					throw new InvalidDataException("Resource sprites span multiple sheets. Try loading their sequences earlier.");
```
pchote: There are a couple of places where we do checks like this now.  It's out of scope for this PR, but a future one can introduce lint rules to detect these cases.

penev92:
Well code looks OK and it still works ingame, so :+1: 
Let's hear what profiling has to say about performance.

RoosterDragon:
Can confirm render time has reduced by a large margin - this will be well worth applying to the other layers.

:+1: from me after minor issues addressed.

OpenRA.Mods.Common/Traits/World/ResourceLayer.cs
```diff
@@ -17,35 +18,52 @@
...
+			foreach (var kv in spriteLayers)
 			{
-				if (shroudObscured(uv))
-					continue;
-
-				var c = render[uv];
-				if (c.Sprite != null)
-					new SpriteRenderable(c.Sprite, wr.World.Map.CenterOfCell(uv.ToCPos(world.Map)),
-						WVec.Zero, -511, c.Type.Palette, 1f, true).Render(wr); // TODO ZOffset is ignored
+				if (resource.Sprite != null && resource.Type.Palette == kv.Key)
```
RoosterDragon: This null check isn't needed - the logic is the same regardless.

OpenRA.Mods.Common/Traits/World/ResourceLayer.cs
```diff
@@ -17,35 +18,52 @@
...
+					kv.Value.Update(cell, resource.Sprite);
+				else
+					kv.Value.Update(cell, null);
 			}
 		}
 
+		public void Render(WorldRenderer wr)
+		{
+			foreach (var kv in spriteLayers)
```
RoosterDragon: Consider enumerating ```spriteLayers.Values```.

OpenRA.Mods.Common/Traits/World/ResourceLayer.cs
```diff
@@ -84,30 +94,60 @@ public void WorldLoaded(World w, WorldRenderer wr)
...
+					dirty.Add(cell);
 				}
 			}
+
+			// Build the sprite layer dictionary for rendering resources
+			// All resources that have the same palette must also share a sheet and blend mode
+			foreach (var r in resources)
+			{
+				var pal = r.Value.Palette;
+				TerrainSpriteLayer layer;
```
RoosterDragon: I would declare the ```layer``` next to the ```TryGetValue```.

OpenRA.Mods.Common/Traits/World/ResourceLayer.cs
pchote: Not quite - if `resource.Sprite` is null then `resource.Type` is meaningless at best, and sometimes null.

pchote: I can add a comment explaining that.

RoosterDragon: That would be good in case somebody else comes to the same conclusion I just did.

pchote:
Updated based on @RoosterDragon's comments and rebased.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/dee280bc26c0ea8fe6666da5a3dd8f817a2620a0...ac9faab63761cb31dbbb05e7908dfc8f74a69672)

Issue #8651, opened 2015-07-05
Indication of map with custom rules in multiplayer lobby


Basti504:
It would be nice if one could see that a map contains custom rules by some kind of a indicator in the multiplayer lobby. Something similar to the "Advanced" label on the Resource Center. A player would know in front that the game may behave in a totally different way.

Especially if people are creating identical named maps with cheating-like rules and crappy trol-sounds, as I encountered this morning.

Example:
http://resource.openra.net/maps/3525 (clone)
http://resource.openra.net/maps/64/ (origional)

Mailaender:
I think @cncnet has this, too. Added a similar warning to our lobby. https://github.com/OpenRA/OpenRA/issues/8652

PR #8652, opened 2015-07-05
Added a warning about custom game rules

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/8651.

### DIFF
--- a/OpenRA.Game/Server/Server.cs
+++ b/OpenRA.Game/Server/Server.cs
@@ -359,7 +359,9 @@ namespace OpenRA.Server
 					t.ClientJoined(this, newConn);
 
 				SyncLobbyInfo();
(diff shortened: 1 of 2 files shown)


penev92:
Works :+1:

Unit158:
:+1:

pchote:
In the longer term I think this would be better represented with some kind of icon on or near the minimap display, but this solves the problem acceptably in the short term.

penev92:
Yeah, a visual representation on the map would be nice.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/6dc7b0f05ebeaafd0365829e516e7aafe3743b97...ec4cee53d9d35d50aad40cf901e1f54bcf2dba53)

PR #8654, opened 2015-07-05
Move maximum terrain height definition to mod.yaml.

pchote:
One of the upcoming heightmap-related PRs needs to use this value from a place where it is not convenient to access a TileSet.  There is no reason for this value to be tileset-specific, so it has been moved to the Manifest.


Mailaender:
No change visible. :white_check_mark:

penev92:
Looks a lot more sensible than the current situation :+1: 

Would be even better to have a dynamic max height to not limit mappers though. :/
And to not force modders to do some magical estimates and put a random limit for their mod.

PR #8658, opened 2015-07-05
Higher level timed upgrades

matija-hustic:
Right now timed upgrades are not allowed to get past level 1, i.e., every new granting of the upgrade just refreshes the duration of the previous upgrade instead of increasing its level.

I'd like to loosen this constraint to allow timed upgrades to reach higher levels if different sources grant the same timed upgrade. Per source, however, the highest level would still remain constrained to 1, refreshing duration with every new granting instead of increasing the level. That's the first commit.

This would be used for #8563 through `GrantUpgradeWarhead`. Adapting this warhead to register itself as a source of the upgrade is what commit number two is about.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
@@ -28,18 +28,37 @@ namespace OpenRA.Mods.Common.Traits
 	{
 		class TimedUpgrade
 		{
+			public class UpgradeSource
+			{
+				public readonly object Source;
+				public int Remaining;
+
+				public UpgradeSource(int duration, object source)
+				{
+					Remaining = duration;
+					Source = source;
+				}
+			}
+
 			public readonly string Upgrade;
 			public readonly int Duration;
-			public int Remaining;
+			public readonly HashSet<UpgradeSource> Sources;
+			public int Remaining; // Equal to maximum of all Sources.Remaining
 
-			public TimedUpgrade(string upgrade, int duration)
+			public TimedUpgrade(string upgrade, int duration, object source)
 			{
 				Upgrade = upgrade;
 				Duration = duration;
 				Remaining = duration;
+				Sources = new HashSet<UpgradeSource> { new UpgradeSource(duration, source) };
 			}
 
-			public void Tick() { Remaining--; }
+			public void Tick()
+			{
+				Remaining--;
+				foreach (var source in Sources)
+					source.Remaining--;
+			}
 		}
 
 		class UpgradeState
@@ -66,17 +85,36 @@ namespace OpenRA.Mods.Common.Traits
 			});
 		}
 
-		public void GrantTimedUpgrade(Actor self, string upgrade, int duration)
+		/// <summary>Upgrade level increments are limited to one per source, i.e., if a single source
+		/// attempts granting multiple upgrades, they will not accumulate. They will replace each other
+		/// instead, leaving only the last granted upgrade active. An upgrade from each new distinct
+		/// source will increment the upgrade's level until AcceptsUpgrade starts returning false. Then,
+		/// when no new levels are accepted, the upgrade source with the shortest remaining upgrade
+		/// duration will be replaced by the new source.</summary>
+		public void GrantTimedUpgrade(Actor self, string upgrade, int duration, object source = null)
 		{
 			var timed = timedUpgrades.FirstOrDefault(u => u.Upgrade == upgrade);
 			if (timed == null)
 			{
-				timed = new TimedUpgrade(upgrade, duration);
+				timed = new TimedUpgrade(upgrade, duration, source);
 				timedUpgrades.Add(timed);
 				GrantUpgrade(self, upgrade, timed);
+				return;
+			}
+
+			var src = timed.Sources.FirstOrDefault(s => s.Source == source);
+			if (src == null)
+			{
+				timed.Sources.Add(new TimedUpgrade.UpgradeSource(duration, source));
+				if (AcceptsUpgrade(self, upgrade))
+					GrantUpgrade(self, upgrade, timed);
+				else
+					timed.Sources.Remove(timed.Sources.OrderByDescending(s => s.Remaining).Last());
 			}
 			else
-				timed.Remaining = Math.Max(duration, timed.Remaining);
+				src.Remaining = duration;
+
+			timed.Remaining = Math.Max(duration, timed.Remaining);
 		}
 
 		// Different upgradeable traits may define (a) different level ranges for the same upgrade type,
@@ -150,13 +188,20 @@ namespace OpenRA.Mods.Common.Traits
 			s.Watchers.Add(action);
 		}
 
+		/// <summary>Watchers will be receiving notifications while the upgrade's level is nonzero.
+		/// They will also be provided with the number of ticks before the level returns to zero,
+		/// as well as the duration in ticks of the timed upgrade (provided in the first call to
+		/// GrantTimedUpgrade).</summary>
 		public void Tick(Actor self)
 		{
 			foreach (var u in timedUpgrades)
 			{
 				u.Tick();
-				if (u.Remaining <= 0)
-					RevokeUpgrade(self, u.Upgrade, u);
+				foreach (var source in u.Sources)
+					if (source.Remaining <= 0)
+						RevokeUpgrade(self, u.Upgrade, u);
+
+				u.Sources.RemoveWhere(source => source.Remaining <= 0);
 
 				foreach (var a in upgrades.Value[u.Upgrade].Watchers)
 					a(u.Duration, u.Remaining);
--- a/OpenRA.Mods.Common/Traits/Crates/GrantUpgradeCrateAction.cs
+++ b/OpenRA.Mods.Common/Traits/Crates/GrantUpgradeCrateAction.cs
@@ -47,7 +47,8 @@ namespace OpenRA.Mods.Common.Traits
 		bool AcceptsUpgrade(Actor a)
 		{
 			var um = a.TraitOrDefault<UpgradeManager>();
-			return um != null && info.Upgrades.Any(u => um.AcceptsUpgrade(a, u));
+			return um != null && (info.Duration > 0 ?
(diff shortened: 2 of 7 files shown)


Mailaender:
I guess nothing can go wrong here and I can't wait to see #8563. :+1:

WolfGaming:
Does the Deviator mind control thingy in Dune 2000 still work with multiple Deviators on a single target?

matija-hustic:
This will only change behavior for warheads using `GrantUpgradeWarhead` and currently only TS EMP is using it. And even if other traits used this, I don't believe this would adversely affect them since all of them use (or should be using because higher level timed upgrades weren't available before) the implicit `UpgradableTrait.UpgradeMaxAcceptedLevel == 1`.

pchote:
You'll need to make sure that this doesn't break EMP - multiple hits are supposed to reset the timer instead of adding new levels.  Somebody appears to have broken the EMP targeting on bleed, so I couldn't test this myself.

matija-hustic:
Oh yeah, you're right @pchote, this breaks stuff.

I falsely assumed all temporarily upgradable traits used `UpgradableTrait.UpgradeMaxAcceptedLevel == 1` which is not true. `DisableUpgrade` is currently assuming `UpgradeMaxAcceptedLevel == 2` just so that `UpgradeManager.AcceptsUpgrade` would always return `true`.

With this PR if you have `n` cannons firing at the same time at the sime target you should have `UpgradeMaxAcceptedLevel == n + 1` to ensure that `UpgradeManager.AcceptsUpgrade == true`.

Plus, the watcher notification system gets broken since multiple entries in `timedUpgrades` for the same upgrade means multiple notifications per tick to the watchers. And in the case of `TimedUpgradeBar` this means a race in which upgrade determines the value in the progress bar.

I'd like to have higher level upgrades, but I'll need to rethink my approach.

GraionDilach:
>I falsely assumed all temporarily upgradable traits used `UpgradableTrait.UpgradeMaxAcceptedLevel == 1`. `DisableUpgrade` is currently assuming `UpgradeMaxAcceptedLevel == 2` just so that `UpgradeManager.AcceptsUpgrade` would always return true.

Replace the calls with `UpgradeManager.AcknowledgesUpgrade` where needed in that case, I added that function exactly to prevent such workarounds when I was at the disguise decoration. I was unaware that workaround was already in use.

matija-hustic:
@GraionDilach nice! Then I'm only left with the broken notifications, but I think I've got an idea for that, too.

matija-hustic:
Currently the `UpgradeManager`, when it notifies watchers of remaining upgrade time, also provides the total upgrade duration. This duration is simply taken to be the duration of the first upgrade that was granted. E.g. if for some reason the first EMP upgrade had a duration of 250 and a subsequent 300 ticks, the `UpgradeManager` would send a notification to the watcher that the progress is 300/250 (possibly causing bad stuff). This kind of situations are implicitly assumed to never happen by the existing code, and I'll go with the same assumption to keep this PR's size to a minimum.

Also, seeing how currently timed upgrades are either 0 or 1, notifications are sent only when upgrades are at the nonzero (1) level. The remaining time in the notifications represents how much longer the upgrade will be at the nonzero level. Again, to minimize the scope of this PR, I have chosen to do the analogous thing - only notify when at nonzero level and say how much time is left until back to zero level. The actual level info is not included in the notification.

WRT the implementation, I wanted to keep LINQ out of `UpgradeManager.Tick()' so I've kept individual upgrades in `timedUpgrades`, but having each upgrade keep track of its granting sources.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -66,17 +85,31 @@ public UpgradeManager(ActorInitializer init)
...
 				GrantUpgrade(self, upgrade, timed);
 			}
 			else
+			{
+				if (source != null)
+				{
+					var src = timed.Sources.FirstOrDefault(s => s.Source == source);
+					if (src == null)
+					{
+						timed.Sources.Add(new TimedUpgrade.UpgradeSource(duration, source));
```
GraionDilach: This needs an AcceptsUpgrade check, otherwise you can pass MaxAcceptable via having more actor sources than it.

GraionDilach: You could also add a "replace oldest iteration of this upgrade coming from a different actor" in the else section of that AcceptsUpgrade check.

matija-hustic: Wasn't that the point of replacing `AcceptsUpgrade` with `AcknowledgesUpgrade`?

GraionDilach: https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs#L128

matija-hustic: Hm... This would then break the convention that you call `AcceptsUpgrade` before you grant the upgrade. Thus making the `GrantTimeUpgrade` effectively a `TryGrantTimeUpgrade`. If `UpgradeMaxAcceptedLevel` shouldn't be broken for timed upgrades, it might make more sense to modify `AcceptsUpgrade` so that it takes into account the source...

matija-hustic: Or make a new `AcceptsTimedUpgrade`

matija-hustic: Ah no, if you replace the oldest upgrade, then it's basically not a try. It'll always succeed. I'll do that then.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -66,17 +85,36 @@ public UpgradeManager(ActorInitializer init)
...
+			}
+
+			var src = timed.Sources.FirstOrDefault(s => s.Source == source);
+			if (source == null || src == null)
+			{
+				timed.Sources.Add(new TimedUpgrade.UpgradeSource(duration, source));
+				if (AcceptsUpgrade(self, upgrade))
+					GrantUpgrade(self, upgrade, timed);
+				else
+					timed.Sources.Remove(timed.Sources.OrderByDescending(s => s.Remaining).Last());
```
GraionDilach: The new upgrade doesn't get applied in this branch.

matija-hustic: It doesn't need to. It replaces an existing one. That was your idea, right?

GraionDilach: I rephrase: You only remove the oldest instance, without adding the current instance there.

matija-hustic: I'm not following. So let's say I call `GrantTimedUpgrade` from tesla trooper C's upgrade granting warhead. At that time the level is already at `UpgradeMaxAcceptedLevel == 2` due to let's say two warheads on tesla trooper A and tesla trooper B. What should happen during the call to `GrantTimedUpgrade`?

GraionDilach: Your intention is to replace A with C. What this code does, however, is removing A, leaving the target alone with B's upgrade until the next upgrade granting weapon, which will add the upgrade alongside B.

matija-hustic: @GraionDilach, in that case this code doesn't touch the upgrades (neither `GrantUpgrade` nor `RevokeUpgrade` get called). Notice that `GrantUpgrade` and `RevokeUpgrade` use `TimedUpgrade timed` as the upgrade source which grants a certain number of upgrades. And `timed` handles its own housekeeping (sources and remaining time for individual upgrades) separately.
This means that `GrantTimedUpgrade` just updates the information on to whom the upgrades are related. Instead of the two granted upgrades being related to A and B, they will be updated to relate to B and C.

GraionDilach: Oh, right, I see it now, instead of recreating the upgrade itself, you just recreate the watcher. Sorry.

GraionDilach:
:+1: from me.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -150,13 +188,19 @@ public void RegisterWatcher(string upgrade, Action<int, int> action)
...
 		{
 			foreach (var u in timedUpgrades)
 			{
 				u.Tick();
-				if (u.Remaining <= 0)
-					RevokeUpgrade(self, u.Upgrade, u);
+				foreach (var source in u.Sources)
+					if (source.Remaining <= 0)
+						RevokeUpgrade(self, u.Upgrade, u);
+				u.Sources.RemoveWhere(source => source.Remaining <= 0);
```
abcdefg30: Could we get a new line here?

obrakmann:
This looks fine, although a bit confusing with the two distinct `Remaining` fields in `TimedUpgrade` and `TimedUpgrade.UpgradeSource` that are kept in lock-step.

The regular timed upgrade case also still works :+1:

matija-hustic:
The newline is there + the comment on the `Remaining` field.

I've added a third commit that updates users of timed upgrades to work well with the changes from this PR. Plus it gets rid of the workaround described in https://github.com/OpenRA/OpenRA/pull/8658#issuecomment-118853622

Sorry to do this, but could you re-review @GraionDilach, @obrakmann, please?

GraionDilach:
My :+1: still stands. The additional changes looks good - I'm a fan of removing unnecessary workarounds.

obrakmann:
Yep, still looks fine. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c50d1b7a4a22cb35ca8d886865c5280284b393ff%5E...c50d1b7a4a22cb35ca8d886865c5280284b393ff). Hope the entry is ok like that.

PR #8660, opened 2015-07-05
Fix a bug in Server

penev92:
I derped in #8629...


RoosterDragon:
:+1:

deniz1a:
Is that needed?

penev92:
Yes. Long story short, `DropClient()` modifies `Conns` by removing a connection from it, and you don't want to be modifying a collection that you are enumerating at the moment.

Mailaender:
Underp accepted.

deniz1a:
Ok thanks.

PR #8662, opened 2015-07-06
Upgradable TargetableUnit & apply to Allies 03

atlimit8:


Mailaender:
Allies03 as well as submarines and aircraft targeting still work. :white_check_mark:

GraionDilach:
That's exactly the implementation I wanted to see in #8339. This man deserves a medal.

OpenRA.Mods.Common/Traits/Air/TargetableAircraft.cs
```diff
@@ -35,7 +35,8 @@ public override string[] TargetTypes
 		{
 			get
 			{
-				return (self.CenterPosition.Z > 0) ? info.TargetTypes
+				return IsTraitDisabled ? new string[] { }
+					: (self.CenterPosition.Z > 0) ? info.TargetTypes
 					: info.GroundedTargetTypes;
```
penev92: Indentation here isn't consistent. Some parentheses around the outer `else` would be good for readability, and maybe have in on one line.

OpenRA.Mods.Common/Traits/Air/TargetableAircraft.cs
```diff
@@ -35,8 +35,12 @@ public override string[] TargetTypes
...
 			get
 			{
-				return (self.CenterPosition.Z > 0) ? info.TargetTypes
-					: info.GroundedTargetTypes;
+				return IsTraitDisabled ?
+						new string[] { }
+					: (self.CenterPosition.Z > 0) ?
+						info.TargetTypes
+					:
+						info.GroundedTargetTypes;
```
penev92: How about
```
return IsTraitDisabled ? new string[] { } 
        : (self.CenterPosition.Z > 0 ? info.TargetTypes : info.GroundedTargetTypes);
```

OpenRA.Mods.RA/Traits/TargetableSubmarine.cs
```diff
@@ -34,7 +34,7 @@ public override string[] TargetTypes
 		{
 			get
 			{
-				return cloak.Cloaked ? info.CloakedTargetTypes : info.TargetTypes;
+				return IsTraitDisabled ? new string[] { } : cloak.Cloaked ? info.CloakedTargetTypes : info.TargetTypes;
```
penev92: `return IsTraitDisabled ? new string[] { } : (cloak.Cloaked ? info.CloakedTargetTypes : info.TargetTypes);`
please.

pchote: Bonus points if you extract that empty array into a static readonly variable instead of creating a new object each time.

atlimit8:
This prevents reimplementation of the upgradable logic in subclasses.

GraionDilach:
Those should be reimplemented with some dummy upgrades involved - Helicopter granting one when in-air, etc.

atlimit8:
The point of `TargetableUnit.GetTargetableTypes()` is to separate the upgrade logic from subclasses. As long as the upgrade logic works for TargetableUnit and the enabled behavior of subclasses work, then the subclasses should also have the correct disabled behavaior.

That said, `TargetableAircraft` and `TargetableSubmarine` can be tested for correct disabled behavior by adding a nonempty dummy `UpgradeTypes` (such as `UpgradeTypes: dummy_upgrade`) and `UpgradeMinEnabledLevel: 1` to any `TargetableAircraft` and `TargetableSubmarine`.

Example that disables `TargetableAircraft` and `TargetableSubmarine` in the ra mod:
```diff
diff --git a/mods/ra/rules/defaults.yaml b/mods/ra/rules/defaults.yaml
index 2807979..fb9e8f3 100644
--- a/mods/ra/rules/defaults.yaml
+++ b/mods/ra/rules/defaults.yaml
@@ -304,6 +304,8 @@
        TargetableAircraft:
                TargetTypes: Air
                GroundedTargetTypes: Ground, Repair
+               UpgradeTypes: dummy_upgrade
+               UpgradeMinEnabledLevel: 1
        HiddenUnderFog:
                Type: CenterPosition
        AttackMove:
diff --git a/mods/ra/rules/ships.yaml b/mods/ra/rules/ships.yaml
index 88a30d1..4057867 100644
--- a/mods/ra/rules/ships.yaml
+++ b/mods/ra/rules/ships.yaml
@@ -23,6 +23,8 @@ SS:
        TargetableSubmarine:
                TargetTypes: Ground, Water, Repair
                CloakedTargetTypes: Underwater, Repair
+               UpgradeTypes: dummy_upgrade
+               UpgradeMinEnabledLevel: 1
        Cloak:
                CloakTypes: Underwater
                InitialDelay: 0
@@ -72,6 +74,8 @@ MSUB:
        TargetableSubmarine:
                TargetTypes: Ground, Water, Repair
                CloakedTargetTypes: Underwater, Repair
+               UpgradeTypes: dummy_upgrade
+               UpgradeMinEnabledLevel: 1
        Cloak:
                CloakTypes: Underwater
                InitialDelay: 0
```
The planes fly with impunity on the shellmap.

atlimit8:
Made the test a gist: https://gist.github.com/atlimit8/e37f47d2d5658f5b5996.

OpenRA.Mods.Common/Traits/Air/TargetableAircraft.cs
```diff
@@ -31,13 +31,9 @@ public TargetableAircraft(Actor self, TargetableAircraftInfo info)
 			this.self = self;
 		}
 
-		public override string[] TargetTypes
+		protected override string[] GetTargetableTypes()
```
pchote: This plumbing change makes the already complicated targeting plumbing a step worse, IMO.  Keep it as a property without any inner methods, and do the disabled check in these two extra places.

abcdefg30:
Could you also please remove https://github.com/atlimit8/OpenRA/blob/UpgradableTargetableUnit/mods/ra/maps/allies-03b/map.yaml#L1309-L1314 ? It was part of the workaround in allies03b.

Other than that :+1: / :white_check_mark:

atlimit8:
Changed.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3e7456db1b23e0297ea8eee7bd086e3ef6a5e3e9...d3281d33d1c34de58e4e4381c0700e2e063f6484)

Mailaender:
Sadly this added a new bug https://github.com/OpenRA/OpenRA/issues/8722.

GraionDilach:
@atlimit8 is already aware of that, see #8727.

PR #8663, opened 2015-07-06
Fixed a crash when disguising the spy as infantry without stand2

Mailaender:
Some recent sequence cleanups must have undone my workaround in https://github.com/OpenRA/OpenRA/pull/4265. Time to fix it properly. http://resource.openra.net/maps/3367/ is a good test case. It has c2 in the creep base near the weapon factory.

### DIFF
--- a/OpenRA.Game/Graphics/Animation.cs
+++ b/OpenRA.Game/Graphics/Animation.cs
@@ -10,6 +10,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Support;
 
 namespace OpenRA.Graphics
 {
@@ -198,6 +200,11 @@ namespace OpenRA.Graphics
 		public ISpriteSequence GetSequence(string sequenceName)
 		{
 			return sequenceProvider.GetSequence(name, sequenceName);
+		}
+
+		public string GetRandomExistingSequence(string[] sequences, MersenneTwister random)
+		{
+			return sequences.Where(s => HasSequence(s)).RandomOrDefault(random);
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Traits/Render/WithInfantryBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithInfantryBody.cs
@@ -70,10 +70,17 @@ namespace OpenRA.Mods.Common.Traits
 
 			DefaultAnimation = new Animation(init.World, rs.GetImage(self), RenderSprites.MakeFacingFunc(self));
 			rs.Add(new AnimationWithOffset(DefaultAnimation, null, () => IsTraitDisabled));
+			PlayStandAnimation(self);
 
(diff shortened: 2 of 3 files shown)


pchote:
It would be better if this randomized the stand sequences that are available instead of picking one and doing nothing if that doesnt exist.

Mailaender:
Fixed both the code duplication and missing back-fall.

OpenRA.Mods.Common/Traits/Render/WithInfantryBody.cs
```diff
@@ -70,12 +70,19 @@ public WithInfantryBody(ActorInitializer init, WithInfantryBodyInfo info)
...
 
-			DefaultAnimation.PlayFetchIndex(NormalizeInfantrySequence(init.Self, info.StandSequences.Random(Game.CosmeticRandom)), () => 0);
 			state = AnimationState.Waiting;
 			move = init.Self.Trait<IMove>();
 		}
 
+		public void PlayStandAnimation(Actor self)
+		{
+			var sequence = DefaultAnimation.GetRandomExistingSequence(Info.StandSequences);
+			if (!string.IsNullOrEmpty(sequence))
```
pchote: A null check is sufficient here.  The earlier HasSequences will filter out any bogus (i.e. empty) sequences.

OpenRA.Mods.RA/Traits/Render/WithDisguisingInfantryBody.cs
```diff
@@ -39,7 +39,9 @@ public override void Tick(Actor self)
 			if (disguise.AsSprite != intendedSprite)
 			{
 				intendedSprite = disguise.AsSprite;
-				DefaultAnimation.ChangeImage(intendedSprite ?? rs.GetImage(self), info.StandSequences.Random(Game.CosmeticRandom));
+				var sequence = DefaultAnimation.GetRandomExistingSequence(info.StandSequences);
+				if (!string.IsNullOrEmpty(sequence))
```
pchote: Null check is sufficient here too.

Mailaender:
Updated.

obrakmann:
Fixes the crash :+1:

OpenRA.Game/Graphics/Animation.cs
```diff
@@ -199,5 +200,10 @@ public ISpriteSequence GetSequence(string sequenceName)
 		{
 			return sequenceProvider.GetSequence(name, sequenceName);
 		}
+
+		public string GetRandomExistingSequence(string[] sequences)
```
pchote: This should take the random generator as an argument.  In its current state, somebody might accidentally use this for synced state (e.g. via PlayAnimationThen) and introduce a desync.

Mailaender: Hadn't thought about that. Updated.

abcdefg30:
:+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8a43dffd01119147292067df86e452ad648d05ce...d6085b41f496b48504896560b7c3f18bac4df4c0)

PR #8664, opened 2015-07-06
Removed code duplication from Mobile.cs

Mailaender:
Detected by dupFinder.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Mobile.cs
+++ b/OpenRA.Mods.Common/Traits/Mobile.cs
@@ -221,34 +221,7 @@ namespace OpenRA.Mods.Common.Traits
 			if ((movementCost = MovementCostForCell(world, cell)) == int.MaxValue)
 				return false;
 
-			if (SharesCell && world.ActorMap.HasFreeSubCell(cell))
-				return true;
-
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -221,32 +221,7 @@ public bool CanEnterCell(World world, Actor self, CPos cell, out int movementCos
...
-						return false;
-					var crushables = a.TraitsImplementing<ICrushable>();
-					if (!crushables.Any())
-						return false;
-					foreach (var crushable in crushables)
-						if (!crushable.CrushableBy(Crushes, self.Owner))
-							return false;
-				}
-			}
+			CanMoveFreelyInto(world, self, cell, ignoreActor, check);
```
chrisforbes: Do you mean `return CanMoveFreelyInto(...` ?

As is, this isn't equivalent at all.

Mailaender: Yes, fixed. Thanks.

abcdefg30:

obrakmann:
Looks fine, stuff still moves. :+1:

PR #8665, opened 2015-07-06
Removed unused duplicated code in Warhead.cs

Mailaender:
This is a left-over from https://github.com/OpenRA/OpenRA/pull/6030 which now lives in WeaponInfo.cs for quite some time.


chrisforbes:
:+1:

penev92:
:+1:

PR #8666, opened 2015-07-06
Reverted friendly units' immunity to EMP.

GraionDilach:
Also removed the explicit enemy vehicle targeting.

I acknowledge, I introduced this, but tbh, I consider these features and not bugs.

### DIFF
--- a/mods/ts/weapons/superweapons.yaml
+++ b/mods/ts/weapons/superweapons.yaml
@@ -90,7 +90,6 @@ EMPulseCannon:
 		Range: 3c0
 		Duration: 250
 		Upgrades: empdisable
-		ValidTargets: Vehicle
 
 ClusterMissile:
 	ValidTargets: Ground, Water, Air


penev92:
Didn't EMP in TS affect all units regardless of player stance?
Also it seems pretty accurate to not care about it.

GraionDilach:
I'm unsure, I never really used them in TS. The main reason why I filed the PR back at the time was that my mobile EMPs frequently EMPd themselves at the AI and it pretty much broke stuff.

pchote:
You should be able to set a flag on the warhead to prevent it from affecting the actor that fired it.

GraionDilach:
Yes, one can. Only since #8558 tho, which added this issue in the first case.

mods/ts/weapons/superweapons.yaml
```diff
@@ -90,7 +90,7 @@ EMPulseCannon:
 		Range: 3c0
 		Duration: 250
 		Upgrades: empdisable
-		ValidTargets: Vehicle
+		ValidStances: Ally, Neutral, Enemy
```
abcdefg30: You don't need to explicitly set this, as `ValidStances` has this values as defaults.

GraionDilach: Oh, right. For some reason I remembered having only Neutral and Enemy.

GraionDilach:
Removed unnecessary line.

abcdefg30:
Now EMPs target also buildings etc., but that is your intention? So :+1: / :white_check_mark:

GraionDilach:
For now, yes. We don't have a separate TargetType defined for terrain alone (Ground also lets building to be targeted).

It was also the status before #8558.

obrakmann:
:+1:

Issue #5274, opened 2014-05-09
Renamings and Reorganisation


Mailaender:
A collection from stuff suggested in IRC. These could be done once the pull request queue is a little less crowded https://github.com/OpenRA/OpenRA/issues/5263.

- [x] OpenRA.Support.Random clashes with System.Random https://github.com/OpenRA/OpenRA/pull/5359/
- [x] WRange → WDist ~~https://github.com/OpenRA/OpenRA/pull/4701~~ https://github.com/OpenRA/OpenRA/pull/8667
- [ ] OpenRA.Renderer.* is actually more like OpenRA.Platform.* because it is graphics + input and could also include OpenRA.Sound which is currently in OpenRA.Game so we save linking to SDL2 there. ~~https://github.com/OpenRA/OpenRA/issues/5787~~ ~~https://github.com/OpenRA/OpenRA/pull/8538~~ https://github.com/OpenRA/OpenRA/pull/8632
- [x] Many traits in OpenRA.Mods.RA should actually be in OpenRA.Mods.Common as they are shared in all three mods. Same goes for system rules every mod needs or shared chrome definitions which could also be moved to `mods/common/` for clarity. ~~https://github.com/OpenRA/OpenRA/pull/5321~~ https://github.com/OpenRA/OpenRA/pull/6511
- [x] Don't call factions countries or races. https://github.com/OpenRA/OpenRA/issues/7552

PR #8667, opened 2015-07-06
Renamed WRange to WDist

Mailaender:
We planned to do that last year. Almost last remainder of https://github.com/OpenRA/OpenRA/issues/5274.

### DIFF
--- a/OpenRA.Game/WDist.cs
+++ b/OpenRA.Game/WDist.cs
@@ -20,43 +20,43 @@ namespace OpenRA
 	/// <summary>
 	/// 1d world distance - 1024 units = 1 cell.
 	/// </summary>
-	public struct WRange : IComparable, IComparable<WRange>, IEquatable<WRange>, IScriptBindable, ILuaAdditionBinding, ILuaSubtractionBinding, ILuaEqualityBinding, ILuaTableBinding
+	public struct WDist : IComparable, IComparable<WDist>, IEquatable<WDist>, IScriptBindable, ILuaAdditionBinding, ILuaSubtractionBinding, ILuaEqualityBinding, ILuaTableBinding
 	{
 		public readonly int Range;
 		public long RangeSquared { get { return (long)Range * (long)Range; } }
 
-		public WRange(int r) { Range = r; }
-		public static readonly WRange Zero = new WRange(0);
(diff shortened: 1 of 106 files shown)


OpenRA.Game/WDist.cs
```diff
@@ -20,43 +20,43 @@ namespace OpenRA
 	/// <summary>
 	/// 1d world distance - 1024 units = 1 cell.
 	/// </summary>
-	public struct WRange : IComparable, IComparable<WRange>, IEquatable<WRange>, IScriptBindable, ILuaAdditionBinding, ILuaSubtractionBinding, ILuaEqualityBinding, ILuaTableBinding
+	public struct WDist : IComparable, IComparable<WDist>, IEquatable<WDist>, IScriptBindable, ILuaAdditionBinding, ILuaSubtractionBinding, ILuaEqualityBinding, ILuaTableBinding
 	{
 		public readonly int Range;
```
pchote: These will probably want to change too.

Mailaender: I thought this was a nice opportunity to get rid of the `WRange.Range` duplication.

pchote: If you're really bothered by it you could implement the explicit operator and do `(int)myWDist` when you need it explicitly.  `WDist.Range` makes less sense than `WRange.Range`.

pchote: `.Length` would be a potential alternative.

Mailaender: Okay. For some reason MonoDevelop had be do this conversion manually. :anger:

penev92:
This has @Phrohdoh's :+1:  btw.

chrisforbes:
:+1: but needs a rebase. Let's get this merged before it bitrots.

Mailaender:
Rebased without conflicts and updated the trait documentation generator.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/337118cb457633756fa8b503189d48bdbab19211...28ee922f6302ec58fc4b661d516dcde9efb920a2)

Mailaender:
Created a followup for the non-internal changes: https://github.com/OpenRA/OpenRA/issues/8691.

PR #8669, opened 2015-07-06
Ensure FieldLoader is left in its original state when loading Settings

RoosterDragon:
If something throws, ensure we don't leave FieldLoader in a borked state on the way out. Discovered by Coverity. [Diff ignoring whitespace.](https://github.com/OpenRA/OpenRA/pull/8669/files?w=1)


Mailaender:
Oh nice. I never really understood how to fix that Coverity defect. :+1:

penev92:
Seems reasonable :+1:

chrisforbes:
I don't like this. If the FieldLoader has mangleable state, we should have instances. You're currently hosed if you use it from multiple threads.

PR #8672, opened 2015-07-06
Convert remaining terrain layers to use TerrainSpriteLayer

pchote:
Extends #8649 to the remaining types.

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
+++ b/OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
@@ -28,13 +28,15 @@ namespace OpenRA.Mods.Common.Traits
 		public virtual object Create(ActorInitializer init) { return new EditorResourceLayer(init.Self); }
 	}
 
-	public class EditorResourceLayer : IWorldLoaded, IRenderOverlay
+	public class EditorResourceLayer : IWorldLoaded, IRenderOverlay, INotifyActorDisposing
 	{
 		protected readonly Map Map;
 		protected readonly TileSet Tileset;
 		protected readonly Dictionary<int, ResourceType> Resources;
 		protected readonly CellLayer<CellContents> Tiles;
 		protected readonly HashSet<CPos> Dirty = new HashSet<CPos>();
+
+		readonly Dictionary<PaletteReference, TerrainSpriteLayer> spriteLayers = new Dictionary<PaletteReference, TerrainSpriteLayer>();
 
 		public EditorResourceLayer(Actor self)
 		{
@@ -58,6 +60,28 @@ namespace OpenRA.Mods.Common.Traits
 
 			foreach (var cell in Map.AllCells)
 				UpdateCell(cell);
+
+			// Build the sprite layer dictionary for rendering resources
+			// All resources that have the same palette must also share a sheet and blend mode
+			foreach (var r in Resources)
+			{
+				var res = r;
+				var layer = spriteLayers.GetOrAdd(r.Value.Palette, pal =>
+				{
+					var first = res.Value.Variants.First().Value.First();
+					return new TerrainSpriteLayer(w, wr, first.Sheet, first.BlendMode, pal, false);
+				});
+
+				// Validate that sprites are compatible with this layer
+				var sheet = layer.Sheet;
+				if (res.Value.Variants.Any(kv => kv.Value.Any(s => s.Sheet != sheet)))
+					throw new InvalidDataException("Resource sprites span multiple sheets. Try loading their sequences earlier.");
+
+				var blendMode = layer.BlendMode;
+				if (res.Value.Variants.Any(kv => kv.Value.Any(s => s.BlendMode != blendMode)))
+					throw new InvalidDataException("Resource sprites specify different blend modes. "
+						+ "Try using different palettes for resource types that use different blend modes.");
+			}
 		}
 
 		public void UpdateCell(CPos cell)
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
```diff
@@ -58,6 +60,27 @@ public void WorldLoaded(World w, WorldRenderer wr)
 
 			foreach (var cell in Map.AllCells)
 				UpdateCell(cell);
+
+			// Build the sprite layer dictionary for rendering resources
+			// All resources that have the same palette must also share a sheet and blend mode
```
penev92: `ResourceLayer` already has the exact same piece of code. Seems like SmudgeLayer isn't too far off either.
Looks like they duplicate a bunch of other code as well. I can't suggest a place at the moment, but those duplicate code snippets should be extracted to a common location and called from there.

OpenRA.Mods.D2k/Traits/World/BuildableTerrainLayer.cs
```diff
@@ -15,22 +15,30 @@
...
+		}
 
 		public void WorldLoaded(World w, WorldRenderer wr)
 		{
 			theater = wr.Theater;
-			tileset = w.TileSet;
-			map = w.Map;
-			tiles = new Dictionary<CPos, Sprite>();
-			dirty = new Dictionary<CPos, Sprite>();
+			render = new TerrainSpriteLayer(w, wr, theater.Sheet, BlendMode.Alpha, wr.Palette("terrain"), wr.World.Type != WorldType.Editor);
```
penev92: Bonus points if you unhardcode the palette.

OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
pchote: `EditorResourceLayer` duplicates a bit of `ResourecLayer`, but there is significantly more code in there which it doesn't (and shouldn't) use.  I made the best of two bad choices by splitting it off into a different trait with some duplication vs shoehorning more complexity into `ResourceLayer` by forcing it to also work with the editor.  Unless you have a specific suggestion there's not much I can do here.

pchote: `SmudgeLayer` is horrible, and needs to be thrown out and replaced, so a similar comment applies there too.

Mailaender: > ClearSides FindClearSides(ResourceType t, CPos p)

is something our dupFinder (see https://ci.appveyor.com/project/OpenRA/openra/build/1.0.4556/artifacts for example) spits out, but they are not too easy to unduplicate. They look identical from static source code analysis, but the `CellContains` is completely different.

OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
```diff
@@ -141,18 +164,39 @@ public void Render(WorldRenderer wr)
...
 			Dirty.Clear();
 
-			foreach (var uv in wr.Viewport.AllVisibleCells.MapCoords)
-			{
-				var t = Tiles[uv];
-				if (t.Sprite != null)
-					new SpriteRenderable(t.Sprite, wr.World.Map.CenterOfCell(uv.ToCPos(Map)),
-						WVec.Zero, -511, t.Type.Palette, 1f, true).Render(wr);
-			}
+			foreach (var kv in spriteLayers.Values)
```
RoosterDragon: Minor quibble: This variable is not a key-value.

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
```diff
@@ -125,12 +137,14 @@ public void AddSmudge(CPos loc)
 
 		public void TickRender(WorldRenderer wr, Actor self)
 		{
-			var remove = new List<CPos>();
+			var remove = new HashSet<CPos>();
```
RoosterDragon: This doesn't need to be a set, only unique items will be added in the first place.

pchote:
Updated.

OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
```diff
@@ -58,6 +60,27 @@ public void WorldLoaded(World w, WorldRenderer wr)
...
 			foreach (var cell in Map.AllCells)
 				UpdateCell(cell);
+
+			// Build the sprite layer dictionary for rendering resources
+			// All resources that have the same palette must also share a sheet and blend mode
+			foreach (var r in Resources)
+			{
+				var layer = spriteLayers.GetOrAdd(r.Value.Palette, pal =>
+				{
+					var first = r.Value.Variants.First().Value.First();
```
penev92: for `r`.

OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
```diff
@@ -58,6 +60,27 @@ public void WorldLoaded(World w, WorldRenderer wr)
...
+					return new TerrainSpriteLayer(w, wr, first.Sheet, first.BlendMode, pal, wr.World.Type != WorldType.Editor);
```
penev92: Why the world type check? Won't this always be `false`, as this being the `EditorResourceLayer`?

OpenRA.Mods.D2k/Traits/World/BuildableTerrainLayer.cs
```diff
@@ -15,22 +15,35 @@
...
+		}
 
 		public void WorldLoaded(World w, WorldRenderer wr)
 		{
 			theater = wr.Theater;
-			tileset = w.TileSet;
-			map = w.Map;
-			tiles = new Dictionary<CPos, Sprite>();
-			dirty = new Dictionary<CPos, Sprite>();
+			render = new TerrainSpriteLayer(w, wr, theater.Sheet, BlendMode.Alpha, wr.Palette(info.Palette), wr.World.Type != WorldType.Editor);
```
penev92: Can we unhardcode the blend mode?

penev92:
Works :+1: 
Adding `Changes requested` *at least* for the loop closure thing.

RoosterDragon:
Can confirm a speedup from profiling. We're using a bit more memory than before but it's easily a worthwhile tradeoff.

OpenRA.Mods.D2k/Traits/World/BuildableTerrainLayer.cs
pchote: Technically, sure, but it doesn't make a lot of sense.  This trait renders normal terrain tiles which by definition use the alpha blend mode.

OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
pchote: This isn't a problem in practice because the closure is evaluated immediately.  Fixed it to silence the warning, though.

pchote:
Updated.

obrakmann:
Seems fine to me, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c063d4b82f51fc438c275547b5eaab3ff6bf4c5a%5E...c063d4b82f51fc438c275547b5eaab3ff6bf4c5a)

PR #8675, opened 2015-07-07
Enabled UI localisation for the Red Alert mod

Mailaender:
To get started on this. Using the automation from https://github.com/OpenRA/OpenRA/pull/5152. Also reverts https://github.com/OpenRA/OpenRA/commit/a99d7fee8233baea649936ee425f9313dfe80e94.


pchote:
My main complaint from the last two versions of this PR still hasn't applies: We are getting way ahead of ourselves by replacing the yaml strings without having any mechanism in place for the giant bundle of hardcoded UI strings.  I'm still against against doing this until we at least have a mechanism to translate those.

pchote:
Please lets have at least two PRs finishing our translation support before we run the chrome extraction command:

1. Adjusting the translation code to support directly translating ids from the chrome logic files, and then extract these IDs into mods.common.
2. Adjust the server code to send untranslated string ids to the player, and then translate them server side (with strings also extracted to mods.common).

This chrome string translation is automated, so can be done at any time.  By doing it now, before the translation support is ready, you are just forcing additional overheads on everyone who touches the yaml files and we won't even have proper translations to justify that extra work.

mods/ra/languages/english.yaml
```diff
@@ -1,2 +1,450 @@
 english:
 	english: English
+
+# mods/ra/chrome/install.yaml:
```
pchote: Most of these will need to live in Mods.Common.

Mailaender:
I added translation support for most of the internal widget logic strings now.

mods/ra/languages/english.yaml
Mailaender: No, `install.yaml` is very mod specific.

Mailaender: Also it seems language files don't merge so this is not possible yet.

pchote: I am referring to the whole file, not specifically the install strings.

pchote:
> Also it seems language files don't merge so this is not possible yet.

This falls under point 1, and will need to be fixed before we do the extraction.  Otherwise you are forcing more trouble and manual work down the road.

OpenRA.Game/Sound/Sound.cs
```diff
@@ -86,8 +86,8 @@ public static SoundDevice[] AvailableDevices()
 		{
 			var defaultDevices = new[]
 			{
-				new SoundDevice("AL", null, "Default Output"),
-				new SoundDevice("Null", null, "Output Disabled")
+				new SoundDevice("AL", null, FieldLoader.Translate("SOUND-DEFAULT")),
```
pchote: As I mentioned in IRC before you did any of these changes - FieldLoader has nothing conceptually to do with translations (except that it uses them), so it is not an appropriate class to do the translations.  This should be done via ModData instead.

pchote:
We're also going to need to think about how we deal with map-provided translations (for scripting via Lua, and maybe also map names?), as that will impact the design of our ModData Translate functions.

PR #8118, opened 2015-05-09
Improves some text field behavior.

deniz1a:
This patch improves text field behavior in game lobby, settings menu and asset browser.

PR #8680, opened 2015-07-07
Some text field improvements.

deniz1a:
This is split from #8118.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
@@ -108,9 +108,29 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				panel.Get<ColorBlockWidget>("COLORBLOCK").GetColor = () => Game.Settings.Player.Color.RGB;
 			}
 
+			Action closeAssetBrowser = () =>
+			{
+				if (isVideoLoaded)
+					player.Stop();
+				Ui.CloseWindow();
+				onExit();
+			};
+
 			filenameInput = panel.Get<TextFieldWidget>("FILENAME_INPUT");
+			filenameInput.TakeKeyboardFocus();
 			filenameInput.OnTextEdited = () => ApplyFilter(filenameInput.Text);
-			filenameInput.OnEscKey = filenameInput.YieldKeyboardFocus;
+			filenameInput.OnEscKey = () =>
+			{
+				if (filenameInput.Text.Length == 0)
+					closeAssetBrowser();
+				else
+				{
+					filenameInput.Text = "";
+					filenameInput.OnTextEdited();
+				}
+
+				return true;
+			};
 
 			var frameContainer = panel.GetOrNull("FRAME_SELECTOR");
 			if (frameContainer != null)
@@ -218,13 +238,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 			var closeButton = panel.GetOrNull<ButtonWidget>("CLOSE_BUTTON");
 			if (closeButton != null)
-				closeButton.OnClick = () =>
-				{
-					if (isVideoLoaded)
-						player.Stop();
-					Ui.CloseWindow();
-					onExit();
-				};
+				closeButton.OnClick = closeAssetBrowser;
 		}
 
 		void SelectNextFrame()
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
@@ -523,9 +523,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 			chatLabel = lobby.Get<LabelWidget>("LABEL_CHATTYPE");
 			var chatTextField = lobby.Get<TextFieldWidget>("CHAT_TEXTFIELD");
-
 			chatTextField.TakeKeyboardFocus();
-
 			chatTextField.OnEnterKey = () =>
 			{
 				if (chatTextField.Text.Length == 0)
@@ -538,6 +536,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				chatTextField.Text = "";
 				return true;
 			};
+
 			chatTextField.OnTabKey = () =>
 			{
 				var previousText = chatTextField.Text;
@@ -549,6 +548,8 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				else
 					return true;
 			};
+
+			chatTextField.OnEscKey = () => { chatTextField.Text = ""; return true; };
 
 			chatPanel = lobby.Get<ScrollPanelWidget>("CHAT_DISPLAY");
 			chatTemplate = chatPanel.Get("CHAT_TEMPLATE");
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
@@ -257,23 +257,35 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			name.IsDisabled = () => orderManager.LocalClient.IsReady;
 
 			name.Text = c.Name;
+			var escPressed = false;
 			name.OnLoseFocus = () =>
 			{
+				if (escPressed)
+				{
+					escPressed = false;
+					return;
+				}
+
 				name.Text = name.Text.Trim();
 				if (name.Text.Length == 0)
 					name.Text = c.Name;
-
-				if (name.Text == c.Name)
-					return;
-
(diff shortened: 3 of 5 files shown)


Phrohdoh:
What do we gain from this? What is the reasoning behind it?

chrisforbes:
This seems vaguely OK to me, but I'd rather you pull the shared handler out into its own (possibly anonymous) function rather than invoking OnClick of another widget.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -109,8 +119,20 @@ public AssetBrowserLogic(Widget widget, Action onExit, World world)
...
+			filenameInput.TakeKeyboardFocus();
 			filenameInput.OnTextEdited = () => ApplyFilter(filenameInput.Text);
-			filenameInput.OnEscKey = filenameInput.YieldKeyboardFocus;
+			filenameInput.OnEscKey = () =>
+			{
+				if (filenameInput.Text.Length == 0)
+					closeButton.OnClick();
+				else
+				{
+					filenameInput.Text = null;
```
ScottNZ: This seems dangerous to me. In general you want to use "" over null when it comes to UI text, and it looks to me like this could break if this handler runs twice.

penev92: Yup, always go for `string.Empty`.

deniz1a:
I put the shared closing statements in an Action variable and used that.

```filenameInput.Text = null;``` works because the set method of Text assigns empty string if the value is null.
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Widgets/TextFieldWidget.cs#L23

But assigning an empty string directly is better I guess.

**Comment by deniz1a** [comment]
This PR does the following:
* Asset browser: filter text field gets focus by default. Pressing Esc doesn't yield focus. Instead, if text field is empty, it closes the window. If text is not empty, it resets text (makes it empty) and updates filter.
* Lobby: Pressing Esc resets chat text field if it's not empty, otherwise it closes lobby window. In player name field, when Esc is pressed, text returns to unedited value and focus is yielded.
* Settings menu Display tab: For both frame limit and player name fields: pressing Esc resets text field and yields focus.

File: OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs [hunk 2]
```diff
@@ -108,9 +108,28 @@ public AssetBrowserLogic(Widget widget, Action onExit, World world)
 				panel.Get<ColorBlockWidget>("COLORBLOCK").GetColor = () => Game.Settings.Player.Color.RGB;
 			}
 
+			Action closeAssetBrowser = () =>
+				{
+					if (isVideoLoaded)
+						player.Stop();
+					Ui.CloseWindow();
+					onExit();
+				};
```
pchote: Style nits: indented one level too far, and newline after the `};` please.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -218,13 +237,7 @@ public AssetBrowserLogic(Widget widget, Action onExit, World world)
...
 			var closeButton = panel.GetOrNull<ButtonWidget>("CLOSE_BUTTON");
 			if (closeButton != null)
-				closeButton.OnClick = () =>
-				{
-					if (isVideoLoaded)
-						player.Stop();
-					Ui.CloseWindow();
-					onExit();
-				};
+				closeButton.OnClick = () => closeAssetBrowser();
```
pchote: You don't need the wrapper: `closeButton.OnClick = closeAssetBrowser;` does what you want.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -515,17 +515,16 @@ void CloseWindow()
 					"fog {0}".F(!orderManager.LobbyInfo.GlobalSettings.Fog)));
 			}
 
+			Action closeLobby = () => { CloseWindow(); onExit(); };
 			var disconnectButton = lobby.Get<ButtonWidget>("DISCONNECT_BUTTON");
-			disconnectButton.OnClick = () => { CloseWindow(); onExit(); };
+			disconnectButton.OnClick = () => closeLobby();
```
pchote: `disconnectButton.OnClick = closeLobby`.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -549,6 +548,15 @@ void CloseWindow()
 				else
 					return true;
 			};
+			chatTextField.OnEscKey = () =>
```
pchote: Style nit: newline between `};` and this line.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
```diff
@@ -257,23 +257,28 @@ public static void SetupEditableNameWidget(Widget parent, Session.Slot s, Sessio
 			name.IsDisabled = () => orderManager.LocalClient.IsReady;
 
 			name.Text = c.Name;
-			name.OnLoseFocus = () =>
+			name.OnEnterKey = () =>
```
pchote: Removing the `OnLoseFocus` leaves the widget in an inconsistent state if you unfocus by clicking on the chat text field.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
```diff
@@ -257,23 +257,28 @@ public static void SetupEditableNameWidget(Widget parent, Session.Slot s, Sessio
...
-
-				name.Text = Settings.SanitizedPlayerName(name.Text);
-
-				orderManager.IssueOrder(Order.Command("name " + name.Text));
-				Game.Settings.Player.Name = name.Text;
-				Game.Settings.Save();
+				name.YieldKeyboardFocus();
+				return true;
+			};
+			name.OnEscKey = () =>
```
pchote: Newline here too

OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
```diff
@@ -190,14 +190,21 @@ Action InitDisplayPanel(Widget panel)
...
-			frameLimitTextfield.OnLoseFocus = () =>
+			frameLimitTextfield.OnEnterKey = () =>
 			{
 				int fps;
 				Exts.TryParseIntegerInvariant(frameLimitTextfield.Text, out fps);
 				ds.MaxFramerate = fps.Clamp(1, 1000);
 				frameLimitTextfield.Text = ds.MaxFramerate.ToString();
+				frameLimitTextfield.YieldKeyboardFocus();
+				return true;
+			};
```
pchote: And after every other pseudo-block in this pr.

OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
```diff
@@ -190,14 +190,21 @@ Action InitDisplayPanel(Widget panel)
 
 			var frameLimitTextfield = panel.Get<TextFieldWidget>("FRAME_LIMIT_TEXTFIELD");
 			frameLimitTextfield.Text = ds.MaxFramerate.ToString();
-			frameLimitTextfield.OnLoseFocus = () =>
+			frameLimitTextfield.OnEnterKey = () =>
```
pchote: This has the same issue as the lobby.  If you unfocus using the mouse the widget text will no longer match the underlying data.

pchote:
There are some nice usability improvements here, but there are a few style issues and a couple of bugs that will need to be fixed before we can merge it.  Please also split this into individual commits for lobby / settings / asset browser with commit messages that describe what you said above.

deniz1a:
I fixed all the issues.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -550,6 +550,16 @@ void CloseWindow()
 					return true;
 			};
 
+			chatTextField.OnEscKey = () =>
+			{
+				if (chatTextField.Text.Length == 0)
+					closeLobby();
```
pchote: We deliberately chose not to do this because leaving the lobby is a high-impact action for something that can be so easily triggered.

obrakmann:
Looks fine now. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f3f5bdd6f80d3edf77bfdaa967e680cf969bc837%5E...f3f5bdd6f80d3edf77bfdaa967e680cf969bc837)

Issue #8522, opened 2015-06-20
GDI Upgrade center prerequisites are wrong


pchote:
It is visible in the Nod build palette.

PR #8681, opened 2015-07-08
Fixed GDI upgrade center being visible in Nod build palette

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/8522.


GraionDilach:
:+1:

PR #8684, opened 2015-07-08
Exposed GlobalPaletteEffect to the Lua API for day/night cycles

Mailaender:
This is inspired by how you can set up lighting triggers to create such effects in Tiberian Sun.

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Global/EffectGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/EffectGlobal.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System.Collections.Generic;
+using Eluant;
 using OpenRA.Mods.Common.Traits;
 using OpenRA.Scripting;
 
@@ -18,11 +19,13 @@ namespace OpenRA.Mods.Common.Scripting
 	public class EffectGlobal : ScriptGlobal
 	{
 		readonly IEnumerable<FlashPaletteEffect> fpes;
+		readonly GlobalLightingPaletteEffect lighting;
 
 		public EffectGlobal(ScriptContext context)
 			: base(context)
 		{
 			fpes = context.World.WorldActor.TraitsImplementing<FlashPaletteEffect>();
+			lighting = context.World.WorldActor.TraitOrDefault<GlobalLightingPaletteEffect>();
 		}
 
 		[Desc("Controls the `FlashPaletteEffect` trait.")]
@@ -31,6 +34,18 @@ namespace OpenRA.Mods.Common.Scripting
 			foreach (var fpe in fpes)
 				if (fpe.Info.Type == type)
 					fpe.Enable(ticks);
+		}
+
+		[Desc("Dynamically adjusts the `GlobalLightingPaletteEffect` trait.")]
+		public void ChangeLighting(double red, double green, double blue, double ambient)
+		{
+			if (lighting == null)
+				throw new LuaException("GlobalLightingPaletteEffect needs to be added to the World actor first.");
+
+			lighting.Red = (float)red;
+			lighting.Green = (float)green;
+			lighting.Blue = (float)blue;
+			lighting.Ambient = (float)ambient;
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs
```diff
@@ -98,6 +98,12 @@ public int RandomInteger(int low, int high)
 			return Context.World.SharedRandom.Next(low, high);
 		}
 
+		[Desc("Returns a random floating point value.")]
```
pchote: This needs a giant warning that it shouldn't be used for gameplay code.

GraionDilach:
The shellmap should have a way to keep the color changing in sync - now after a while it'll end up at 1 1 1 1, losing the default tint afterall.

OpenRA.Mods.Common/Scripting/Global/EffectGlobal.cs
```diff
@@ -32,5 +35,17 @@ public void Flash(string type = null, int ticks = -1)
 				if (fpe.Info.Type == type)
 					fpe.Enable(ticks);
 		}
+
+		[Desc("Dynamically adjusts the `GlobalLightingPaletteEffect` trait.")]
+		public void ChangeLighting(double red, double green, double blue, double ambient)
```
pchote: Could we make this accept an HSLColor (which is what we use for the other color-using methods), and then add something to the HSLColor table to create a color from the specified RGB?

pchote: This produces a more consistent API as well as removing the need to expose potentially dangerous floating point stuff to scripters.

GraionDilach: The trait use RGB. All WW ways to change lighting - map, scripts, superweapons - use RGB. Having this on HSL would be odd to me.

pchote: This could then look something like `public void ChangeLighting(double ambient, HSLColor? tint)`.

OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs
Mailaender: I think I will better remove it then. It was just used as a test case where I created an epileptic disco world and understand how Lua works with floats.

OpenRA.Mods.Common/Scripting/Global/EffectGlobal.cs
pchote: @GraionDilach: Limitations of the .NET api mean we can't expose the standard Color type directly.  That's why I suggested we produce an api for creating a color object from RGB values.

Mailaender:
> The shellmap should have a way to keep the color changing in sync - now after a while it'll end up at 1 1 1 1, losing the default tint afterall.

The shellmap didn't use any tints before so this resets to the defaults (which is no modifiers).

OpenRA.Mods.Common/Scripting/Global/EffectGlobal.cs
Mailaender: Experimented a bit, but can't really get it to work with HSLColor. It will also introduce inconsistency with GlobalLightingPaletteEffect and make conversions from legacy triggers harder.

GraionDilach:
I'd still prefer to stop adjusting the lighting when one parameter reaches 1 instead of waiting until all four reaches it, but nothing else I can nitpick on.

abcdefg30:
Needs a rebase.

Mailaender:
Rebased.

abcdefg30:
Looks also good to me. :+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a0745c71dc9c1acfb7545052de74a11219ea722f...8a43dffd01119147292067df86e452ad648d05ce)

PR #8685, opened 2015-07-08
Added projectile sprite testing to the CheckYaml lint rules

Mailaender:
A followup of https://github.com/OpenRA/OpenRA/pull/7709 for projectile sprites. This can detect crashes like #8219 and #8020 beforehand.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckSequences.cs
+++ b/OpenRA.Mods.Common/Lint/CheckSequences.cs
@@ -12,6 +12,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using OpenRA.GameRules;
 using OpenRA.Mods.Common.Traits;
 using OpenRA.Traits;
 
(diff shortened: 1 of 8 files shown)


abcdefg30:
Needs a rebase.

Mailaender:
Rebased.

reaperrr:
Needs another rebase.

reaperrr:
Doesn't appear to work properly. I added `Sequence: idle2` to TD `70mm` cannon, which correctly crashed the game when firing, but it wasn't caught by `make test`.
I then did the same for the `Rockets` weapon, that one was caught but only on 'Take Bialystok', where the weapon is customised.

Mailaender:
Rebased. Not sure why the weapon rule map yaml merging does not work.

GraionDilach:
Can it be related to Inherits doesn't work in weapons.yaml neither?

abcdefg30:
Except the thing @reaperrr pointed out, this seems to work. :+1: / :white_check_mark:

pchote:
Needs a rebase.

OpenRA.Mods.RA/Graphics/TeslaZapRenderable.cs
```diff
@@ -35,13 +35,15 @@ struct TeslaZapRenderable : IRenderable, IFinalizedRenderable
...
+		readonly string dimSequence;
+		readonly string brightSequence;
 		readonly int brightZaps, dimZaps;
 
 		WPos cachedPos;
 		WVec cachedLength;
 		IEnumerable<IFinalizedRenderable> cache;
 
-		public TeslaZapRenderable(WPos pos, int zOffset, WVec length, string image, int brightZaps, int dimZaps, string palette)
+		public TeslaZapRenderable(WPos pos, int zOffset, WVec length, string image, int brightZaps, int dimZaps, string palette, string dim, string bright)
```
pchote: The order of the arguments here is inconsistent: it specifies brightZaps followed by dimZaps, but then dim followed by bright.

To improve readability please change this to `public TeslaZapRenderable(WPos pos, int zOffset, WVec length, string image, string brightZapsSequence, int brightZapsCount, string dimZapsSequence, int dimZapsCount, string palette)`

Mailaender:
Rebased. I don't understand why https://github.com/OpenRA/OpenRA/pull/8732 was merged before or pulled at all. It seemed to only include my bullet cleanup commit.

pchote:
Confirmed @reaperrr's discovery that the lint test doesn't work for Bullets.  That will need to be fixed before we can merge this.

Mailaender:
Updated.

Mailaender:
I can't reproduce any problems. My test cases were #8219 and #8020.

reaperrr:
It works on maps that customize weapons.

It does not work for the weapon yamls.

pchote:
My test case is the same as what @reaperrr mentioned above: Add `Sequence: bogus` to a projectile in the weapon yaml, and observe that the tests fail.  Launch the game, and observe that it crashes when they are spawned.  There is no merging going on.

pchote:
After additional testing I can narrow this down even further: the lint will pick up the bogus sequences *only* if a map modifies the same weapon.  Regular maps with no overrides do not detect the problem.

Mailaender:
Are you sure that Image is != null for your Sequence: bogus?

Mailaender:
As already told: I don't think there is anything I can do about this limitation.

reaperrr:
I suspect that weapon rules are simply not treated as part of general rules, but anything custom on maps is treated as map rules hence why it works there.

OpenRA.Mods.Common/Lint/CheckSequences.cs
```diff
@@ -89,6 +90,48 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 						}
 					}
 				}
+
+				foreach (var weaponInfo in map.Rules.Weapons)
+				{
+					var projectileInfo = weaponInfo.Value.Projectile;
+					if (projectileInfo == null)
+						return;
```
pchote: Some fairly straightforward debugging pointed me here.  This cancels the lint pass as soon as you encounter a weapon without a projectile.

pchote: The very first weapon definition in TD triggers this case, so the pass is not run at all.

Mailaender: Good catch. This was stupid.

reaperrr:
Works now :+1:

Mailaender:
:raised_hands:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/550856c1b1281206b40e42d36e5ec157579beb73...d41f55e18f5f73c34eb9cb7064cdc9844ce2a17f)

PR #8688, opened 2015-07-08
Unhardcode sound and delay in PlaceBuilding.

phrohdoh:
https://github.com/OpenRA/OpenRA/pull/8688/files?w=1

Also stores `order.OrderString` in a variable.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Player/PlaceBuilding.cs
+++ b/OpenRA.Mods.Common/Traits/Player/PlaceBuilding.cs
@@ -15,118 +15,136 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
-	[Desc("Allows to execute build orders.", " Attach this to the player actor.")]
-	class PlaceBuildingInfo : TraitInfo<PlaceBuilding>
+	[Desc("Allows the player to execute build orders.", " Attach this to the player actor.")]
+	public class PlaceBuildingInfo : ITraitInfo
 	{
 		[Desc("Palette to use for rendering the placement sprite.")]
 		public readonly string Palette = "terrain";
+
+		[Desc("Play NewOptionsNotification this many ticks after building placement.")]
+		public readonly int NewOptionsNotificationDelay = 10;
+
+		[Desc("Notification to play after building placement if new construction options are available.")]
+		public readonly string NewOptionsNotification = "NewOptions";
+
+		public object Create(ActorInitializer init) { return new PlaceBuilding(this); }
 	}
 
-	class PlaceBuilding : IResolveOrder
+	public class PlaceBuilding : IResolveOrder
 	{
+		readonly PlaceBuildingInfo info;
+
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Player/PlaceBuilding.cs
```diff
@@ -15,118 +15,136 @@
...
+		{
+			this.info = info;
+		}
 
-					if (queue == null)
-						return;
+		public void ResolveOrder(Actor self, Order order)
+		{
+			var os = order.OrderString;
+			if (os != "PlaceBuilding" ||
```
cjshmyr: If you invert the comparison use `&&` instead

Phrohdoh: Oh, right. Thanks.

Phrohdoh:
Fixed broken inverted logical check.

GraionDilach:
Uhm, NewOptions is not a building sound, but the New Construction Options EVA announcement. The Desc is wrong.

OpenRA.Mods.Common/Traits/Player/PlaceBuilding.cs
```diff
@@ -15,118 +15,136 @@
...
+	public class PlaceBuildingInfo : ITraitInfo
 	{
 		[Desc("Palette to use for rendering the placement sprite.")]
 		public readonly string Palette = "terrain";
+
+		[Desc("Play the NewOptionsSound this many ticks after building placement.")]
+		public readonly int NewOptionsSoundDelay = 10;
+
+		[Desc("Sound to play after building placement.")]
+		public readonly string NewOptionsSound = "NewOptions";
```
pchote: Sound &rarr; Notification here and above.

Phrohdoh: Can do.

Phrohdoh:
@GraionDilach I never said it was a building sound.
It is a sound to be played after the build action.

pchote:
@GraionDilach was referring to the "Sound to play after building placement." description.  This seems to imply that it is played after every building placement, but it actually only plays if the placement changes the available production options.

Phrohdoh:
Okay, that makes sense. I'll reword that too.

GraionDilach:
Honestly, the current wording would imply me Building->BuildSounds. Which that definitely isn't.

abcdefg30:
:+1: / :white_check_mark: from me after the descriptions are fixed.

Phrohdoh:
Updated the field names and descriptions.

GraionDilach:
:+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/128f27f87edf1803baa45ae7cb7ff05ab24746dc...a0745c71dc9c1acfb7545052de74a11219ea722f)

PR #7921, opened 2015-04-11
Move ra silo to defense queue

Amelia55555:
4 reasons:
- Consistency with the TD mod
- Lets you build real structures in parallel with the silo, so players that want to build them aren't penalised as much
- The defense queue is actually the support queue, and the silo is a support structure
- People keep asking for it (to me personally, and general discussion on the forum)

PR #8692, opened 2015-07-09
Move silo from defense to building queue

benletchford:
Love the game. I cannot stand the fact that the silos are in the defence category! Can someone convince me why that makes sense?

Cheers,

Ben


ScottNZ:
#7921

benletchford:
Thanks for that. I guess the issue I have is with the queue name and the icon for it. Turning that category into "support" is an interesting idea.

alexanderdickson:
Is it labeled *defence* or just the shield icon to denote it? Been a while since I played

Issue #7750, opened 2015-03-28
Debug option for showing cell coordinates


penev92:
Being able to see which cell is which have been useful in a multitude of situations when debugging.
I suppose this will want to use something like the performance text for useability reasons, printing the coordinates of the mouse cursor.
I consider this also a pretty basic requirement for a map editor.
##

Phrohdoh:
:+1:
![world](https://camo.githubusercontent.com/b081cd0dd2a901f5637d8051cc29ced0649bf53a/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f343936353438372f323236323130312f37383035386532342d396535362d313165332d393939382d3739613262303063303438392e676966)
![cells](https://camo.githubusercontent.com/054764a57a7a646501cff85b57ff69e2a239aa41/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f343936353438372f323236323130332f37383137396266302d396535362d313165332d386632362d3763643332363932636662662e676966)

penev92:
As long as it doesn't become too visually polluting or too render-taxing, the second image would be the most helpful. :+1:

Mailaender:
Now that https://github.com/OpenRA/OpenRA/issues/8694 has been rejected is https://github.com/OpenRA/OpenRA/pull/9089 sufficient?

penev92:
No, actually my original request was to have this ingame when "Show Terrain Geometry" is enabled, for debugging purposes. So this is still not fully resolved.

teinarss:
Closed by #16442

PR #8694, opened 2015-07-09
Added a map coordinates overlay for the editor

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/7750.

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/MapCoordinatesOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/World/MapCoordinatesOverlay.cs
@@ -1,0 +1,85 @@
+#region Copyright & License Information
+/*
+  * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+  * This file is part of OpenRA, which is free software. It is made
+  * available to you under the terms of the GNU General Public License
+  * as published by the Free Software Foundation. For more information,
+  * see COPYING.
+  */
+#endregion
+
+using System.Collections.Generic;
+using System.Drawing;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Commands;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Renders a debug overlay showing cell coordinates. Attach this to the world actor.")]
+	public class MapCoordinatesOverlayInfo : ITraitInfo
+	{
+		public readonly string Font = "TinyBold";
+
+		public readonly Color Color = Color.White;
+
+		public object Create(ActorInitializer init) { return new MapCoordinatesOverlay(this); }
+	}
+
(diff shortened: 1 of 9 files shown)


OpenRA.Mods.Common/Traits/World/MapCoordinatesOverlay.cs
```diff
@@ -0,0 +1,82 @@
...
+				Enabled ^= true;
+		}
+
+		public IEnumerable<IRenderable> Render(Actor self, WorldRenderer wr)
+		{
+			if (!Enabled)
+				yield break;
+
+			var mouseCell = wr.Viewport.ViewToWorld(Viewport.LastMousePos).ToMPos(wr.World.Map).ToCPos(shape);
+			var cells = AllCells ? wr.Viewport.AllVisibleCells : new CellRegion(shape, mouseCell, mouseCell);
```
Phrohdoh: `AllCells` is never set, so is always false.

Mailaender: I decided to tie this up to the chat command now.

Phrohdoh: Okay, that is undocumented though.

Mailaender: Not really:

> Optional parameter: 'allcells'"

Phrohdoh: I had not seen that, sorry.

pchote:
The useability of this could be improved a lot by combining it with the terrain grid visualization.

pchote:
Elaborating on my earlier comment: a much simpler (both code an UX) implementation would be to show a text string "Cursor position: (x, y)" in a label widget at the top-left of the screen when the terrain grid is enabled.

Phrohdoh:
I don't see why we can't show more than a single tile's coordinates, so not a fan of that requested change.

pchote:
@Phrohdoh: Have you tested this ingame?  Our tile size is smaller than theme hospital, and our fonts are a lot chunkier, so the grid layout you posted in #7750 doesn't work in OpenRA.  To make them fit, @Mailaender displays the x and y coordinates as separate numbers (on separate rows), which hurts the usability of the feature.

Mailaender:
https://github.com/OpenRA/OpenRA/issues/7750 wasn't about labels, but a grid overlay. I don't intend to change that.

pchote:
The changes requested tag is to improve the poor usability of the feature:  The coordinates are not displayed in the standard x,y format, they are hidden by the mouse cursor, and become completely unreadable when you zoom out.  I have suggested a simple way to fix that, but I don't know why I bothered.

Mailaender:
You told me in IRC afterwards that you did it yourself already combined in the terrain overlay so go ahead than. I don't have time for this UI bike shedding.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/9089.

PR #8697, opened 2015-07-09
Made all body, turret and barrel render traits upgradable

reaperrr:
Another milestone towards implementing TS deployed units (artillery, tick tank etc.) via `DeployToUpgrade`.

Allows actors to change appearance on upgrade/levelup too, of course.

### DIFF
--- a/OpenRA.Game/Traits/BodyOrientation.cs
+++ b/OpenRA.Game/Traits/BodyOrientation.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System;
+using System.Linq;
 
 namespace OpenRA.Traits
 {
@@ -62,7 +63,7 @@ namespace OpenRA.Traits
 				if (info.QuantizedFacings >= 0)
 					return info.QuantizedFacings;
 
-				var qboi = self.Info.Traits.GetOrDefault<IQuantizeBodyOrientationInfo>();
+				var qboi = self.Info.Traits.WithInterface<IQuantizeBodyOrientationInfo>().FirstOrDefault();
 				if (qboi == null)
 					throw new InvalidOperationException("Actor type '" + self.Info.Name + "' does not define a quantized body orientation.");
 
--- a/OpenRA.Mods.Common/Traits/Render/WithBarrel.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithBarrel.cs
@@ -17,24 +17,25 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Renders barrels for units with the Turreted trait.")]
-	class WithBarrelInfo : ITraitInfo, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>, Requires<IBodyOrientationInfo>
+	public class WithBarrelInfo : UpgradableTraitInfo, IRenderActorPreviewSpritesInfo, Requires<TurretedInfo>,
+		Requires<ArmamentInfo>, Requires<RenderSpritesInfo>, Requires<IBodyOrientationInfo>
 	{
-		[Desc("Sequence name to use")]
+		[Desc("Sequence name to use.")]
 		[SequenceReference] public readonly string Sequence = "barrel";
 
-		[Desc("Armament to use for recoil")]
+		[Desc("Armament to use for recoil.")]
 		public readonly string Armament = "primary";
 
-		[Desc("Turreted 'Barrel' key to display")]
-		public readonly string Barrel = "first";
-
-		[Desc("Visual offset")]
+		[Desc("Visual offset.")]
 		public readonly WVec LocalOffset = WVec.Zero;
 
-		public object Create(ActorInitializer init) { return new WithBarrel(init.Self, this); }
+		public override object Create(ActorInitializer init) { return new WithBarrel(init.Self, this); }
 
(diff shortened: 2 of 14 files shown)


mods/ra/weapons/largecaliber.yaml
```diff
@@ -170,7 +170,7 @@ TurretGun:
 		ValidImpactTypes: Water
 
 8Inch:
-	ReloadDelay: 250
+	ReloadDelay: 50#250
```
Phrohdoh: Why this change?

reaperrr: Just for this testcase, to make it upgrade faster.

reaperrr:
Updated, fixed commit message and added `SequenceReferences` to voxel traits.

mods/ra/rules/ships.yaml
```diff
@@ -150,7 +150,7 @@ CA:
 		BuildPaletteOrder: 50
 		Prerequisites: ~syrd, atek, ~techlevel.unrestricted
 	Valued:
-		Cost: 2400
+		Cost: 500#2400
```
Phrohdoh: Why this change?

OpenRA.Mods.Common/Traits/Render/WithBarrel.cs
```diff
@@ -17,24 +17,25 @@
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Renders barrels for units with the Turreted trait.")]
-	class WithBarrelInfo : ITraitInfo, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>, Requires<IBodyOrientationInfo>
+	public class WithBarrelInfo : UpgradableTraitInfo, ITraitInfo, IRenderActorPreviewSpritesInfo, Requires<TurretedInfo>,
```
atlimit8: `UpgradableTraitInfo, ITraitInfo` is redundant. Use just `UpgradableTraitInfo`.

reaperrr:
Updated.

OpenRA.Game/Traits/BodyOrientation.cs
```diff
@@ -62,7 +63,7 @@ public BodyOrientation(ActorInitializer init, BodyOrientationInfo info)
 				if (info.QuantizedFacings >= 0)
 					return info.QuantizedFacings;
 
-				var qboi = self.Info.Traits.GetOrDefault<IQuantizeBodyOrientationInfo>();
+				var qboi = self.Info.Traits.WithInterface<IQuantizeBodyOrientationInfo>().FirstOrDefault();
```
pchote: This is bogus for two reasons:

1.  You're unconditionally taking the first implementation of this trait, regardless of whether it is active or not.
2.  Fixing 1. doesn't help, because a bunch of things implicitly depend on `IQuantizeBodyOrientationInfo` not changing.

A better idea might be to move `IQuantizeBodyOrientationInfo` onto `RenderSprites` / `RenderVoxels`, but i'll leave it to you to work out how to make `RenderSprites` work out the number of facings it should use.

OpenRA.Mods.Common/Traits/Render/WithBarrel.cs
```diff
@@ -51,38 +52,43 @@ public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer i
...
-	class WithBarrel
+	public class WithBarrel : UpgradableTrait<WithBarrelInfo>
 	{
-		WithBarrelInfo info;
-		Actor self;
-		Armament armament;
-		Turreted turreted;
-		IBodyOrientation body;
-		Animation anim;
+		public readonly Animation DefaultAnimation;
```
pchote: This doesn't appear to be used anywhere, so why make it public?

If you need to e.g. run an animation from an external trait then that should be done via an appropriate interface.

OpenRA.Mods.Common/Traits/Render/WithVoxelBody.cs
```diff
@@ -18,16 +18,19 @@
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Also returns a default selection size that is calculated automatically from the voxel dimensions.")]
-	public class WithVoxelBodyInfo : ITraitInfo, IQuantizeBodyOrientationInfo, IRenderActorPreviewVoxelsInfo, Requires<RenderVoxelsInfo>
+	public class WithVoxelBodyInfo : UpgradableTraitInfo, IQuantizeBodyOrientationInfo, IRenderActorPreviewVoxelsInfo, Requires<RenderVoxelsInfo>, Requires<IBodyOrientationInfo>
 	{
-		public readonly string Sequence = "idle";
+		[SequenceReference] public readonly string Sequence = "idle";
```
pchote: These are all voxels, so `[SequenceReference]` is wrong.

OpenRA.Game/Traits/BodyOrientation.cs
penev92: In what situation would this be a problem and why would `RenderSprites`/`RenderVoxels` deal with the facings?

reaperrr:
Closing this for now.

Refactoring the `IQuantizeBodyOrientation` implementation should be done in its own PR first, as discussed on IRC. 
However, a large portion of traits which currently implement `IQuantizeBodyOrientation` will go away when `RenderSimple`/`RenderBuilding` and their derivatives are replaced by `WithSpriteBody` + XY, so it makes far more sense to do the latter first.

In summary, this PR would likely sit dead for a few more weeks, so there's not point in occupying the PR queue until then.

PR #8698, opened 2015-07-10
Make GainsExperience upgrades explicit in MiniYaml.

atlimit8:
Applies a similar policy as requested in comments for #8572 to GainsExperience.


reaperrr:
Just the other day while writing my testcases for #8697 I thought we should do exactly this, glad to see it adressed so quickly.

Looks good and works in-game, so :+1:

Phrohdoh:
We could mark these fields with an attribute instead of including it in the desc.

Mailaender:
Thanks.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/76c5c119651bd9023b67e47aed2117039930dc10...e9fb1d46dccbb301177fd2a002e51106955c6dff)

PR #8699, opened 2015-07-10
Shows real owner of disguised units to allies.

deniz1a:
Continuation of #8298.

I changed the code. If disable shroud cheat should actually give the player observer view, then it is enough to only check for RenderPlayer as ally. I think then that debug option should be renamed to "Get observer view" or similar. Also, then it is much simpler than how I was trying to do it. No need for shared method to get current player...

This PR does the following:
- In tooltip, shows the real owner of disguised units to allies. Now it matches their radar color. Previously it was shown only to the owner.
- Corrects some comments.

### DIFF
--- a/OpenRA.Mods.RA/Traits/Disguise.cs
+++ b/OpenRA.Mods.RA/Traits/Disguise.cs
@@ -49,10 +49,10 @@ namespace OpenRA.Mods.RA.Traits
 		{
 			get
 			{
-				if (disguise.Disguised)
-					return self.Owner == self.World.LocalPlayer ? self.Owner : disguise.AsPlayer;
+				if (!disguise.Disguised || self.Owner.IsAlliedWith(self.World.RenderPlayer))
+					return self.Owner;
 
-				return self.Owner;
+				return disguise.AsPlayer;
 			}
 		}
 	}
@@ -135,8 +135,8 @@ namespace OpenRA.Mods.RA.Traits
 
 			if (target != null)
 			{
-				// Take the image of the target's disguise, if it exist.
-				// E.g., SpyA is disguised as a dog. SpyB then targets SpyA. We should use the dog image.
+				// Take the image of the target's disguise, if it exists.
+				// E.g., SpyA is disguised as a rifle infantry. SpyB then targets SpyA. We should use the rifle infantry image.
 				var targetDisguise = target.TraitOrDefault<Disguise>();
 				if (targetDisguise != null && targetDisguise.Disguised)
 				{


OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -49,10 +49,10 @@ public Player Owner
 		{
 			get
 			{
-				if (disguise.Disguised)
-					return self.Owner == self.World.LocalPlayer ? self.Owner : disguise.AsPlayer;
+				if (!disguise.Disguised || self.Owner.IsAlliedWith(self.World.RenderPlayer))
```
pchote: Using RenderPlayer from synced code is asking for desyncs.

deniz1a: Ok how should it be done?

deniz1a: This also uses RenderPlayer, to determine radar colors: https://github.com/deniz1a/OpenRA/blob/disguise-owner/OpenRA.Mods.RA/Traits/Disguise.cs#L125

But is that not a problem because it doesn't affect game state?

GraionDilach: Revert it to LocalPlayer maybe?

deniz1a: But that wouldn't work for spectators who choose a player's view.

deniz1a:
So, does this cause desync? I tested it and nothing happened. How should it be changed?

Unit158:
You'll need to play with other clients to test for desyncs.

deniz1a:
I tested it by launching multiple instances of the game.

deniz1a:
I tested this again, and seems to work, no desyncs.

OpenRA.Mods.RA/Traits/Disguise.cs
abcdefg30: I think
> This also uses RenderPlayer, to determine radar colors: https://github.com/deniz1a/OpenRA/blob/disguise-owner/OpenRA.Mods.RA/Traits/Disguise.cs#L125
But is that not a problem because it doesn't affect game state?

is a valid point, isn't it?
Removing `PR: Changes requested`.

abcdefg30:
:+1:

OpenRA.Mods.RA/Traits/Disguise.cs
penev92: Not really, because the radar color may be as wrong as it likes, it doesn't affect the game state at all.

Phrohdoh: It does matter if you see a disguised actor in a different player color.

penev92:
Perhaps this calls for some multiplayer testing, to make sure we're not introducing desync time bombs?

penev92:
Tested, no desyncs detected.
Also works :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f8725717d769ad59223a07639b5160e14d852a1f...3be2f3d5d17a5000be35800d05ebef61f925f34e)

PR #8700, opened 2015-07-10
Renames "Disable Shroud & Fog" debug option to "Observer view".

deniz1a:
Because it sets RenderPlayer to null which affects more than shroud and fog.

### DIFF
--- a/OpenRA.Mods.Common/Commands/DevCommands.cs
+++ b/OpenRA.Mods.Common/Commands/DevCommands.cs
@@ -36,7 +36,7 @@ namespace OpenRA.Mods.Common.Commands
 				help.RegisterHelp(name, helpText);
 			};
 
-			register("disableshroud", "toggles shroud and minimap.");
+			register("visibility", "toggles visibility checks and minimap.");
 			register("givecash", "gives the default or specified amount of money.");
 			register("givecashall", "gives the default or specified amount of money to all players and ai.");
 			register("instantbuild", "toggles instant building.");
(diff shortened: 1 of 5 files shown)


pchote:
I think "Disable visibility checks" is more accurate terminology than "observer view", which doesn't say what it actually does.

Phrohdoh:
That is definitely more accurate, but it also disables shroud and fog rendering.

deniz1a:
Ok then how should the internal stuff be named?
DevShroudDisable
disableshroud
DISABLE_SHROUD

To DevDisableVisibility...?

abcdefg30:
Any progress here?

deniz1a:
I'm waiting for naming suggestions. Is DevDisableVisibility ok?

abcdefg30:
I think `DevDisableVisibilityChecks` is better, but it might be a bit long.

deniz1a:
Shroud and fog could be considered part of visibility checks, so ```Disable visibility checks``` would cover it.

Made requested changes, rebased.

OpenRA.Mods.Common/Commands/DevCommands.cs
```diff
@@ -36,7 +36,7 @@ public void WorldLoaded(World w, WorldRenderer wr)
 				help.RegisterHelp(name, helpText);
 			};
 
-			register("disableshroud", "toggles shroud and minimap.");
+			register("disablevisibilitychecks", "toggles observer view of the map and radar.");
```
abcdefg30: Could you adjust the description too, please?

deniz1a:
Ok, the debug checkbox says "Disable visibility checks" and the dev command is /visibility. Also fixed the dev command description.

But now when I think about it, "Spectator view" would be the best description for this command. Because players know what a spectator is but they might not know what visibility checks are. Spectator view is player oriented and visibility check is more of an engine code terminology. What do you think?

deniz1a:
And this debug option gives spectator view... So it is also accurate description. There wouldn't be a need to write "Give spectator view" because the checkbox already implies that what the description says is enabled when the checkbox is ticked.

abcdefg30:

deniz1a:
But it would look even better with "Spectator view", right?

abcdefg30:
I think it is ok as is.
Also @pchote was already against "Spectator/observer view":
> I think "Disable visibility checks" is more accurate terminology than "observer view", which doesn't say what it actually does.

deniz1a:
Ok.

PR #8703, opened 2015-07-11
Require trait properties in MiniYaml

atlimit8:
Allows for marking certain trait properties as required using `[FieldLoader.Require]` or `[FieldLoader.LoadUsing("...", true)]`.

Based on comment from @Phrohdoh on #8698.

To test, just comment out one of the required trait properties.

### DIFF
--- a/OpenRA.Game/FieldLoader.cs
+++ b/OpenRA.Game/FieldLoader.cs
@@ -24,6 +24,26 @@ namespace OpenRA
 {
 	public static class FieldLoader
 	{
+		public class MissingFieldsException : YamlException
+		{
+			public readonly string[] Missing;
+			public readonly string Header;
+			public override string Message
+			{
+				get
+				{
+					return (string.IsNullOrEmpty(Header) ? "" : Header + ": ") + Missing[0]
+						+ string.Concat(Missing.Skip(1).Select(m => ", " + m));
+				}
+			}
+
+			public MissingFieldsException(string[] missing, string header = null, string headerSingle = null) : base(null)
+			{
+				Header = missing.Length > 1 ? header : headerSingle ?? header;
+				Missing = missing;
+			}
+		}
+
 		public static Func<string, Type, string, object> InvalidValueAction = (s, t, f) =>
 		{
 			throw new InvalidOperationException("FieldLoader: Cannot parse `{0}` into `{1}.{2}` ".F(s, f, t));
@@ -45,6 +65,7 @@ namespace OpenRA
 		public static void Load(object self, MiniYaml my)
 		{
 			var loadInfo = TypeLoadInfo[self.GetType()];
+			var missing = new List<string>();
 
 			Dictionary<string, MiniYaml> md = null;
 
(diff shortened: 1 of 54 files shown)


pchote:
This is a great idea.  The implementation looks sensible, but I haven't gone over the changes line by line and can't run the game to test anything.

pchote:
The [trait docs](https://github.com/OpenRA/OpenRA/wiki/Traits) will need to be updated to reflect this.  I suggest replacing the default value (which is meaningless) with something like **[Required]** for these properties.

atlimit8:
That should be done by the changes to `OpenRA.Mods.Common/UtilityCommands/ExtractTraitDocsCommand.cs`, except its more like *(required)*.

Phrohdoh:
What if something is required but the default value is the desired value?

atlimit8:
That is a contradiction. Either the MiniYaml must overwrite the default making it meaningless, or the MiniYaml is optional and the default value is used if the MiniYaml is not present.

atlimit8:
`[FieldLoader.Require]` and `[FieldLoader.LoadUsing("...", true)]` force the default value to be overwritten or throw an exception stopping the engine. Either way, the hard coded value is not used.

I may be misunderstanding your question or you may misunderstand what `require` means in this context,

pchote:
Oh sorry, when I skimmed through the code I completely missed that you had already done it.

obrakmann:
Looks good to me and works as expected, :+1: 

I guess we will fine-tune these required annotations as we come across them. I think there are more candidates for it than this PR already adds (`Queue` in `Buildable` comes to mind, or `PanelName` in `ObjectivesPanel`.)

OpenRA.Game/GameRules/ActorInfo.cs
```diff
@@ -94,7 +103,16 @@ static ITraitInfo LoadTraitInfo(string traitName, MiniYaml my)
...
 			var info = Game.CreateObject<ITraitInfo>(traitName + "Info");
-			FieldLoader.Load(info, my);
+			try
+			{
+				FieldLoader.Load(info, my);
+			}
+			catch (FieldLoader.MissingFieldsException e)
+			{
+				throw new FieldLoader.MissingFieldsException("Trait name " + traitName + ": "
+					+ (e.Message.Contains(',') ? "Required properties missing: " : "Required properties missing: ") + e.Message);
```
pchote: Typo: that second branch should be "Required property missing: ".

OpenRA.Game/FieldLoader.cs
```diff
@@ -52,19 +58,33 @@ public static void Load(object self, MiniYaml my)
...
+							missing.Add(fli);
 						continue;
+					}
 				}
 
 				fli.Field.SetValue(self, val);
 			}
+
+			if (missing.Any())
+				throw new MissingFieldsException(missing.First().YamlName + string.Concat(missing.Skip(1).Select(r => ", " + r.YamlName)));
```
pchote: If you added a `string[]` property for the missing fields then you won't need to detect on `,` when handling it (which is a bit gross).

atlimit8:
I know I tried some more that looked like they should be required, but `make test` would fail. I decided it would be better for them to be handled by those more familiar and/or while those traits are being changed.

OpenRA.Game/FieldLoader.cs
atlimit8: Yea, I knew it was a hack. OK.

atlimit8:
`FieldLoader.MissingFieldsException` now takes a `string[]` of missing properties and an optional header `string`.

abcdefg30:
Looks good to me. :+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d3281d33d1c34de58e4e4381c0700e2e063f6484...4b081ee9775085e9f34c9d5d55afe038ccf6f44f)

PR #8705, opened 2015-07-12
Removes "Start Game" button from lobby.

deniz1a:
Now only ready checkboxes determine game start. Also adds ready checkboxes to spectators.


ScottNZ:
I really don't see a point in adding checkboxes for spectators. They aren't even playing, so they have no need to be ready to play.

deniz1a:
I added that because otherwise the logic to determine when a spectator needs a checkbox would be too complex. So I decided to add checkboxes to specs too.

deniz1a:
Since there is no start game button, specs would need ready checkboxes for the following situations:
* single player game
* multiplayer with only specs

Unit158:
:-1:
Why do we need to remove this? It's sort of found in every decent RTS. If we want to remove the rudeness factor from it, we should have it start a countdown in the lobby before the game starts.

deniz1a:
Reasons to remove it:
* It is redundant. Checkboxes accomplish the same task.
* It makes space for possible future buttons or design elements.
* It prevents game start if not all players clicked ready. If someone is not clicking ready, it is better to kick that player than to start the game with someone who's not responding.

GraionDilach:
:-1: There are cases when the host looks aside and your way forces the game launching without him. Also when many people wait for each other, there are cases when people just spam the ready checkbox without checking the situation neither.

I'd even say that due to the short loadtimes the default mods have, this will mess up the early games of those who does not monitor the lobby screen like crazy.

deniz1a:
Then they should only click ready when they are ready to start the game. Isn't that better?

GraionDilach:
The host is always ready in your implementation and I'm not certain if one can unready him/herself.

So yes, definitely not better. Manual launching works for ages now, no need to change it.

deniz1a:
No, the host also has to click ready. I didn't change that, only added checkboxes to specs.

Phrohdoh:
:-1: for removing it.
I shouldn't be locked out of my servers because someone is just sitting around and not readying-up.

deniz1a:
Then kick that player.

PR #8709, opened 2015-07-13
Removes force start and auto start from lobby.

deniz1a:
Instead of #8705. I think this way it is clearer and simpler.
- No auto start. Host has to manually click Start Game button.
- No force start. Start Game button is disabled if not all players are ready.


Phrohdoh:
What if everyone is read by the afk host?

deniz1a:
Then they would either have to wait for the host or go to another server...

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
```diff
@@ -438,9 +438,9 @@ public static void SetupSpawnWidget(Widget parent, Session.Slot s, Session.Clien
 		public static void SetupEditableReadyWidget(Widget parent, Session.Slot s, Session.Client c, OrderManager orderManager, MapPreview map)
 		{
 			var status = parent.Get<CheckboxWidget>("STATUS_CHECKBOX");
-			status.IsChecked = () => orderManager.LocalClient.IsReady || c.Bot != null;
+			status.IsChecked = () => orderManager.LocalClient.IsReady;
 			status.IsVisible = () => true;
```
deniz1a: STATUS_CHECKBOX is defined as Visible: false in yaml and set as true here. Is there a reason for not removing both of them?

Unit158: :-1: You didn't fix the issue at all.

Unit158: To elaborate: we need more logic to implement what we discussed earlier. Just don't do the OS specific code part.

deniz1a: I was going to do the start countdown too but decided to break it up into smaller PRs so they can be reviewed separately.

Unit158: Alright, I'm fine with this then. I have yet to do a code review and test, however.

dan9550:
I'm all for having a start game button, the current way games are started is WAYY to jarring. At the very least we should have a countdown when everyone is ready (that is is we have no "Start Game" button).

pchote:
> No force start. Start Game button is disabled if not all players are ready.

We discussed this (at length!) in IRC, and this goes against both the consensus as well as the years of player feedback while we developed the lobby UI.  :-1: until you please actually listen to what other people are saying.

deniz1a:
So, should ready checkboxes be removed for that situation? If ticking ready is not required for game start, then shouldn't the checkboxes be removed?

pchote:
No.

deniz1a:
If clicking ready doesn't mean anything, then why is it there?

pchote:
Most of this comes back to the same point (which we have repeatedly explained) - our lobby UI is set up the way it is for consistency across the different cases within OpenRA (skirmish, multiplayer, players with and without the installed maps, etc) and with the larger community of RTS and other online games.  Please stop trying to break this.

deniz1a:
But the UI could change depending on the current case. If it is not required for players to click ready, checkboxes could be removed. This would make it clearer and would also indicate the options that are available to the players.

I think checkboxes shouldn't be removed and instead force start should be removed.

dan9550:
Here my little compromise on this idea.
The check boxes stay whether or not we think they do anything or are needed they give at the very least a visible indication that a player is ready to start the game.

While the force start isn't something everyone likes there are cases where having it available is appropriate. I would like to see the force start not become available as an option to the host until a majority of players have marked themselves as ready. Although this may not be for everyone, i believe it could work and kind of makes sense.

5 second or so countdown before the game starts :+1:

deniz1a:
How about 2/3 majority?

Phrohdoh:
The countdown is useless unless players can unready and cancel the countdown, in which case we still have a griefing issue.

MicroJOo:
:-1:

GraionDilach:
I am also :-1: against this. Having the checkboxes alone are unintuitive in a skirmish screen. In fact, I don't think a change is even required here, only the addition of a countdown.

However, seeing @deniz1a's recent habits of dismissing general consensus/opening PRs before even planning out exactly what the PR is supposed to do, I'm not certain if I'd like to see this pushed further by him...

deniz1a:
This PR doesn't remove start button. Host has to click Start Game.

obrakmann:
:-1: This does not implement the solution we discussed at length in IRC. As experience shows, this also has little chance to ever get to that point, so I'll close it.

deniz1a:
Would it be considered if I added countdown to this?

Unit158:
It would be considered if you did what we asked and agreed upon. :\

penev92:
The consensus from the lengthy discussion on IRC was just adding countdown, and possibly removing the autostart. Both checkboxes and the `Start game` button have their purpose, as does the force start.

deniz1a:
What's the purpose of force start?

Unit158:
So the host may force the game to start without ready up. We've already covered this several times.

deniz1a:
But what's the point of starting the game with afk players?

penev92:
That is up to the host.
Please let's end the discussion already. We've been over every single point of it multiple times by now.

deniz1a:
Ok.

PR #8710, opened 2015-07-13
Implements victory and defeat music.

GraionDilach:
Followup of #8562.

Due to map loading themes became useless to us, I repurposed them into defeat music at the RA, D2K and the TS mod. In C&C, GDI and Nod had separate loadthemes, so I repurposed the Nod one.
I also changed the D2K shellmap to use the network theme instead of the score track.

The first commit is the logic, the second commit might need it's own PR, because the Nod victory/map themes were unlisted previously alongside three other tracks.

### DIFF
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -352,6 +352,8 @@ namespace OpenRA.Traits
 		void OnObjectiveFailed(Player player, int objectiveID);
 	}
 
+	public interface IGameOver { void GameOver(World world); }
+
 	public interface IWarhead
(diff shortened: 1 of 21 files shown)


Mailaender:

obrakmann:
This works very nicely indeed and also finally fixes the weird overlap where the score screen music would start to play a split second before the video starts. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/38e152a66f69f10a1912eb6af15bd5c504268fe7%5E...38e152a66f69f10a1912eb6af15bd5c504268fe7)

PR #8711, opened 2015-07-13
ResourceType.MinDensityToHarvest

phrohdoh:
This isn't really the effect I wanted but works well enough so I figured we could use it.

What I'd like to do is set and use this value on a per-cell resource instance basis.
If we'd rather have that and anyone has pointers, please let me know.


GraionDilach:
What is the benefit of having this on a per-cell instance basis? I assume you already have an example case of that in mind - at the moment I cannot think of the gain of the latter (loss would be certainly a new control interface for it alone).

Phrohdoh:
I'd like to work on a seed/farming system where planted resources grow but cannot be harvested until fully grown, but this PR has no particular use right now.

If this is ever desired it'll be pretty simple to reopen or reimplement.
Closing as unwanted for now.

PR #8712, opened 2015-07-14
Rename ProduceableWithLevel to ProducibleWithLevel.

phrohdoh:


Unit158:
Needs an upgrade rule then :+1:

GraionDilach:
Doesn't need the upgrade rule yet imo, due to the trait haven't been featured in a playtest before. I'm neutral of this change though.

pchote:
The current `produceable` matches our use of `upgradeable`, so i'm not convinced about this.

Unit158:
We should be defaulting to US English however, as our own contributing guidelines/rules state.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/63b3fc288d7a61695587618c983736737ef5a69d%5E...63b3fc288d7a61695587618c983736737ef5a69d)

PR #8713, opened 2015-07-14
Fixes Empty filter in server browser.

deniz1a:
Currently, the Empty filter in server browser doesn't work. Waiting covers empty servers too, so when only Empty is selected, nothing is shown. Deselecting Waiting shouldn't filter out empty servers.

I haven't tested this code but when only Empty is checked, it should show servers with zero players.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/ServerBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/ServerBrowserLogic.cs
@@ -381,7 +381,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			if ((game.State == (int)ServerState.GameStarted) && !showStarted)
 				return true;
 
-			if ((game.State == (int)ServerState.WaitingPlayers) && !showWaiting)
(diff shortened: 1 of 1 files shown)


deniz1a:
Ok now I tested it and it works.

abcdefg30:
Now only selecting `Waiting` doesn't show anything.

deniz1a:
It works for me? When only Waiting is selected, it should show servers with non-zero players that have not started. Maybe there was no such server? Also did you change the mod version in yaml so you can see servers for the release?

abcdefg30:
Ah, I thought `Waiting` only selected show also show *empty* servers that are *waiting* for players. But  idk. This works then. :+1: / :white_check_mark:

obrakmann:
Behaviour now matches www.openra.net/games, so I guess this makes sense. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7a209e77db92f3bd17ad031e11452f2ae5beefcb%5E...7a209e77db92f3bd17ad031e11452f2ae5beefcb)

Issue #7509, opened 2015-02-20
Missile code needs refactoring


reaperrr:
It needs at least the following improvements, in my opinion:

- proper 3D math
- ROT should apply to vertical movement as well (maybe as separate VerticalROT)
- customisable launch angle
- support for lazy curve flight behavior (like Bullets with Angle, but with target tracking)
- support for straight flight path instead of the current "move to target's Z height as quickly as possible" behavior
- support for Airburst (as in TS)
- support for voxel artwork

These would be nice to have as well:
- acceleration (instead of constant speed from launch to impact)
- customisable cruise altitude
- customisable explosion altitude (might be better than just Airburst)

pchote:
We also need support for voxel artwork for TS/RA2.

Phrohdoh:
Nice list. I'll try and help over any free time I get (whether that be code or discussions on IRC).

reaperrr:
> We also need support for voxel artwork for TS/RA2.

True.

Additionally, I think 
- ```TurboBoost``` should define an integer percentage instead of just a bool with hardcoded value
- ```High``` should be renamed to ```Blockable```, and 
- related: ```BlocksBullets``` should be renamed to ```BlocksProjectiles``` and define a customisable ```BlockHeight``` value which should then be compared to the projectile's altitude when it passes that actor (applies to all projectile types, of course).

reaperrr:
Supercedes #3736.

GraionDilach:
>ROT should apply to vertical movement as well (maybe as separate VerticalROT)
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Effects/Missile.cs#L32

reaperrr:
```MaximumPitch``` works very different than a ```VerticalROT``` as I mean it would. For starters, it's Tan so the curve gets flatter the closer the missile is to the desired height. This looks fine for MiGs' Mavericks and anti-air missiles, but not bazookas and other surface-to-surface missiles. It's also more difficult to get right than a VerticalROT would be. Set the value too low, and the missile might miss its target because it didn't reach pos.Z = 0 quickly enough. Set it too high and the missile moves to zero height too quickly, flying at zero altitude most of the way which looks especially bad in isometric perspective (no visible shadow).

pchote:
Fixed by #8717 and followups.

PR #8717, opened 2015-07-15
Some work on the missile projectile

matija-hustic:
An attempt at tackling #7509.

Introduces:
- vertical rate of turn
- customizable vertical launch angle
- delayed activation of the homing mechanism / propulsion
- freefall while propulsion deactivated
- customizable explosion altitude (for airburst)
- customizable cruise altitude (when target out of range)
- height checks for terrain impact and shadow rendering
- acceleration (instead of constant speed from launch to impact)

Not sure about what was meant in #7509 by
- proper 3D math

Still missing and out of scope of this PR (at least currently)
- support for voxel artwork

### DIFF
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -34,17 +34,23 @@ namespace OpenRA.Mods.Common.Effects
 		[Desc("Should the projectile's shadow be rendered?")]
 		public readonly bool Shadow = false;
 
-		[Desc("Projectile speed in WDist / tick")]
-		public readonly WDist InitialSpeed = new WDist(8);
+		[Desc("Minimum vertical launch angle (pitch).")]
+		public readonly WAngle MinimumLaunchAngle = new WAngle(-128);
 
-		[Desc("Vertical launch angle (pitch).")]
-		public readonly WAngle LaunchAngle = WAngle.Zero;
+		[Desc("Maximum vertical launch angle (pitch).")]
+		public readonly WAngle MaximumLaunchAngle = new WAngle(64);
+
+		[Desc("Minimum launch speed in WDist / tick")]
+		public readonly WDist MinimumLaunchSpeed = WDist.Zero;
+
+		[Desc("Maximum launch speed in WDist / tick")]
+		public readonly WDist MaximumLaunchSpeed = new WDist(8);
 
 		[Desc("Maximum projectile speed in WDist / tick")]
 		public readonly WDist MaximumSpeed = new WDist(512);
 
 		[Desc("Projectile acceleration when propulsion activated.")]
-		public readonly WDist Acceleration = WDist.Zero;
+		public readonly WDist Acceleration = new WDist(5);
 
 		[Desc("How many ticks before this missile is armed and can explode.")]
 		public readonly int Arm = 0;
@@ -62,7 +68,7 @@ namespace OpenRA.Mods.Common.Effects
 		public readonly int HorizontalRateOfTurn = 5;
 
 		[Desc("Vertical rate of turn.")]
-		public readonly int VerticalRateOfTurn = 5;
+		public readonly int VerticalRateOfTurn = 6;
 
 		[Desc("Run out of fuel after being activated this many ticks. Zero for unlimited fuel.")]
 		public readonly int RangeLimit = 0;
@@ -83,7 +89,7 @@ namespace OpenRA.Mods.Common.Effects
 		public readonly string TrailImage = null;
 
 		[Desc("Smoke sequence name.")]
-		[SequenceReference("Trail")] public readonly string TrailSequence = "idle";
+		[SequenceReference("TrailImage")] public readonly string TrailSequence = "idle";
 
 		[Desc("Palette used to render the smoke sequence.")]
 		[PaletteReference("TrailUsePlayerPalette")] public readonly string TrailPalette = "effect";
@@ -124,32 +130,47 @@ namespace OpenRA.Mods.Common.Effects
 
 	public class Missile : IEffect, ISync
 	{
+		enum States
+		{
+			Freefall,
+			Homing,
+			Hitting
+		}
+
 		readonly MissileInfo info;
 		readonly ProjectileArgs args;
 		readonly Animation anim;
 
+		// NOTE: Might be desirable to unhardcode the number -10
 		readonly WVec gravity = new WVec(0, 0, -10);
+
+		int ticks;
 
 		int ticksToNextSmoke;
 		ContrailRenderable contrail;
 		string trailPalette;
-		int terrainHeight;
+
+		States state;
+		bool targetPassedBy;
+		bool lockOn = false;
+
+		WPos targetPosition;
+		WVec offset;
+
+		WVec tarVel;
+		WVec predVel;
 
 		[Sync] WPos pos;
-		[Sync] WVec velocity;
+		WVec velocity;
+		int speed;
+		int loopRadius;
+
+		int renderFacing;
 		[Sync] int hFacing;
 		[Sync] int vFacing;
-		[Sync] bool activated;
-		[Sync] int speed;
 
-		[Sync] WPos targetPosition;
-		[Sync] WVec offset;
-		[Sync] int ticks;
-
-		[Sync] bool lockOn = false;
-
-		[Sync] public Actor SourceActor { get { return args.SourceActor; } }
-		[Sync] public Target GuidedTarget { get { return args.GuidedTarget; } }
+		public Actor SourceActor { get { return args.SourceActor; } }
+		public Target GuidedTarget { get { return args.GuidedTarget; } }
 
 		public Missile(MissileInfo info, ProjectileArgs args)
 		{
@@ -158,19 +179,9 @@ namespace OpenRA.Mods.Common.Effects
 
 			pos = args.Source;
 			hFacing = args.Facing;
-			vFacing = info.LaunchAngle.Angle / 4;
-
-			speed = info.InitialSpeed.Length;
-			velocity = new WVec(WDist.Zero, -info.InitialSpeed, WDist.Zero)
-				.Rotate(new WRot(WAngle.FromFacing(vFacing), WAngle.Zero, WAngle.Zero))
-				.Rotate(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(hFacing)));
-
 			targetPosition = args.PassiveTarget;
 
 			var world = args.SourceActor.World;
-
-			if (world.SharedRandom.Next(100) <= info.LockOnProbability)
-				lockOn = true;
 
 			if (info.Inaccuracy.Length > 0)
 			{
@@ -178,9 +189,18 @@ namespace OpenRA.Mods.Common.Effects
 				offset = WVec.FromPDF(world.SharedRandom, 2) * inaccuracy / 1024;
 			}
 
+			DetermineLaunchSpeedAndAngle(world, out speed, out vFacing);
+
+			velocity = new WVec(0, -info.MaximumLaunchSpeed.Length, 0)
+				.Rotate(new WRot(WAngle.FromFacing(vFacing), WAngle.Zero, WAngle.Zero))
+				.Rotate(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(hFacing)));
+
+			if (world.SharedRandom.Next(100) <= info.LockOnProbability)
+				lockOn = true;
+
 			if (!string.IsNullOrEmpty(info.Image))
 			{
-				anim = new Animation(world, info.Image, () => hFacing);
+				anim = new Animation(world, info.Image, () => renderFacing);
 				anim.PlayRepeating(info.Sequence);
 			}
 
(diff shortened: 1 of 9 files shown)


reaperrr:
> acceleration (instead of constant speed from launch to impact)

Starting with low speed and gradually become faster, rather than flying at constant speed.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -141,52 +200,106 @@ public void Tick(World world)
...
+				// Compute which direction the projectile should be facing
+				var desiredHFacing = OpenRA.Traits.Util.GetFacing(dist, hFacing);
+				var desiredVFacing = OpenRA.Traits.Util.GetFacing(vDist, vFacing);
+
+				// Check whether the homing mechanism is jammed
+				var jammed = info.Jammable && world.ActorsWithTrait<JamsMissiles>().Any(JammedBy);
+				if (jammed)
+				{
+					desiredHFacing = hFacing + world.SharedRandom.Next(-20, 21);
+					desiredVFacing = vFacing + world.SharedRandom.Next(-20, 21);
```
reaperrr: Might be a good opportunity to unhardcode this. Something like "JammedDiversionBounds" (used for both horizontal and vertical, but there might be a better name for that) with Desc "Defines range of facings by which jammed missiles can stray off original path".

ScottNZ: It's jamming the *current path* instead of the original path. This gives a compounding jamming effect on the path of the missile.

reaperrr: Ok, fixed.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -22,40 +22,89 @@ namespace OpenRA.Mods.Common.Effects
...
 		[Desc("Maximum offset at the maximum range")]
 		public readonly WDist Inaccuracy = WDist.Zero;
+
 		[Desc("Probability of locking onto and following target.")]
 		public readonly int LockOnProbability = 100;
-		[Desc("In n/256 per tick.")]
-		public readonly int RateOfTurn = 5;
-		[Desc("Explode when following the target longer than this many ticks.")]
+
+		[Desc("Horizontal rate of turn in 360�/256 per tick.")]
```
pchote: Facing and WAngle units are pretty standard across the code, so i dont think its really appropriate to explain the conversions here.

On a side note, converting the remaining facing units to WAngle is the next ugly giant refactor after height maps, as the current mess is stopping us from doing unit orientations on slopes etc.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -141,52 +200,106 @@ public void Tick(World world)
...
 			{
-				desiredFacing = facing + world.SharedRandom.Next(-20, 21);
-				desiredAltitude = world.SharedRandom.Next(-43, 86);
+				// If target is not within range and cruise altitudes are used,
+				// aim at the same target but at cruise altitude above current terrain height.
+
+				// Target is considered in range if it would take the missile more time to
+				// travel the distance dist.HorizontalLength at speed info.Speed than to
+				// reach ground using maximum vertical turning downwards starting with the
+				// current vertical facing.
```
pchote: Shouldnt this be ground &rarr; target?  Not all targets are on the ground.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -141,52 +200,106 @@ public void Tick(World world)
...
 				ticksToNextSmoke = info.TrailInterval;
 			}
 
 			if (info.ContrailLength > 0)
 				contrail.Update(pos);
 
-			var cell = world.Map.CellContaining(pos);
-
-			var shouldExplode = (pos.Z < 0) // Hit the ground
+			var shouldExplode = (pos.Z <= terrainHeight) // Hit the ground
```
pchote: A future pr may need to look for cliffs ahead in the flight path and start climbing before it hits them.

matija-hustic: Well I've tweaked the in range condition now, so that if you put the cruise altitude and vertical rate of turn large enough, you can get the missiles to climb. It's not a smooth climb tho, but might be possible to smooth it out by using only moderate vertical rate of turn instead of a really big one.

commandos

I guess this would be one of those things that need a lot of testing

matija-hustic:
I'll play around with the acceleration next. I guess that that would mean 
* an additional upper speed limit would probably be needed
* too high speeds might adversely affect accuracy

OpenRA.Mods.Common/Effects/Missile.cs
pchote: It will be steppier than it needs to be right now because terrain height doesn't yet account for slopes (I have code that fixes this, but i'm waiting for the other heightmap stuff to be integrated first).  Don't worry *too* much about TS right now - as long as you don't do anything that we know is incompatible, then we can build on it in future prs.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -22,40 +22,98 @@ namespace OpenRA.Mods.Common.Effects
...
+		public readonly WDist AirburstAltitude = WDist.Zero;
+
+		[Desc("Cruise altitude. Zero means no cruise altitude used.")]
+		public readonly WDist CruiseAltitude = new WDist(512);
+
+		[Desc("Activate homing mechanism after this many ticks.")]
+		public readonly int ActivationDelay = 0;
+
+		[Desc("Image that contains the smoke animation.")]
+		public readonly string Trail = null; // Rename to SmokeImage
```
GraionDilach: I don't think these should be renamed. The tag was named as such in TS/RA2  and theirmods tend to use fire/toxic weapons with them too.

GraionDilach:
Sorry, I took only a quick glampse so far.

What is freefall mode you refer to? Does that apply to missiles which already passed their RangeLimit but has ExplodeWhenEmpty=false and refers that they fall to the ground without doing(?) damage?

Does ActivationDelay means RA2 CourseLockDuration - missile is forced into the initial facing for x frames and only then starts to guide itself to target?

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -141,52 +212,138 @@ public void Tick(World world)
...
+
+				// Target is considered in range if after an additional tick of accelerated motion
+				// the horizontal distance from the target would be less than
+				// the diameter of the circle that the missile travels along when
+				// turning vertically at the maximum possible rate.
+				// This should work because in the worst case, the missile will have to make
+				// a semi-loop before hitting the target.
+				var hLenCurr = dist.HorizontalLength; // Get current distance from target
+				var hLenNext = (long)(hLenCurr - speed - info.Acceleration.Length).Clamp(0, hLenCurr); // Get distance from target in next tick
+				bool inRange = hLenNext * hLenNext * info.VerticalRateOfTurn * info.VerticalRateOfTurn * 3.14 * 3.14
```
GraionDilach: I haven't tested this yet, true, but unless I overlooked something/didn't grasped the concept, you completely disable acceleration when the firer directly fires upside.

pchote: You'll need to rewrite this using integer math.  Floating point math can be non-deterministic, and causes hard-to-reproduce desyncs.

OpenRA.Mods.Common/Effects/Missile.cs
reaperrr: Agreed. It's also used for the RA flamethrowers, so it should stay Trail, in my opinion.

OpenRA.Mods.Common/Effects/Missile.cs
GraionDilach: Missiles in TS/RA2 are similarly steppy if there's a height difference of at least a two cliffs (8 cells) anyway, as far as I remember.

OpenRA.Mods.Common/Effects/Missile.cs
Phrohdoh: So remove the comment?

GraionDilach: The current code have them mixed - new tags and the code use Smoke* already with only the old tags aren't renamed - this should be reverted back entirely to use Trail*.

OpenRA.Mods.Common/Effects/Missile.cs
matija-hustic: This condition might change, but @GraionDilach, right now it works well with any firing angle.

matija-hustic:
@GraionDilach, you have understood the `RangeLimit` and `ExplodeWhenEmpty == false` correctly. But the missile explodes when it touches ground ultimately.

The `ActivationDelay` also uses freefall (not constant velocity) before the guiding starts. This means that the missile does not just travel in a given direction, but that it follows a ballistic trajectory. You would want to use the `LaunchAngle` and `InitialSpeed` here.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -47,28 +58,59 @@ class MissileInfo : IProjectileInfo
 		[Desc("Probability of locking onto and following target.")]
 		public readonly int LockOnProbability = 100;
 
-		[Desc("In n/256 per tick.")]
-		public readonly int RateOfTurn = 5;
+		[Desc("Horizontal rate of turn.")]
+		public readonly int RateOfTurn = 5; // Rename to HorizontalRateOfTurn
```
reaperrr: Nit: Add a 'TODO:' before Rename, easier to keep track of these if you have a buzzword to search for.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -47,28 +58,59 @@ class MissileInfo : IProjectileInfo
...
-		[Desc("Interval in ticks between each spawned Trail animation.")]
-		public readonly int TrailInterval = 2;
+		[Desc("Altitude above terrain below which to explode. Zero effectively deactivates airburst.")]
+		public readonly WDist AirburstAltitude = WDist.Zero;
+
+		[Desc("Cruise altitude. Zero means no cruise altitude used.")]
+		public readonly WDist CruiseAltitude = new WDist(512);
+
+		[Desc("Activate homing mechanism after this many ticks.")]
+		public readonly int ActivationDelay = 0;
```
reaperrr: Just 'Activation' is a little too ambigous, in my opinion. Some modders might confuse this with `Arm`.
`HomingActivationDelay` would leave no room for misunderstandings.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -47,28 +58,59 @@ class MissileInfo : IProjectileInfo
...
+
+		[Desc("Image that contains the smoke animation.")]
+		public readonly string Trail = null; // Rename to TrailImage
```
reaperrr: Nit: Add a 'TODO:' here, too.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -157,52 +212,144 @@ public void Tick(World world)
...
+				// the horizontal distance from the target would be less than
+				// the diameter of the circle that the missile travels along when
+				// turning vertically at the maximum possible rate.
+				// This should work because in the worst case, the missile will have to make
+				// a semi-loop before hitting the target.
+
+				// Get current distance from target
+				var hLenCurr = dist.HorizontalLength;
+
+				// Get underestimate of distance from target in next tick ( => inRange turnes true earlier)
```
reaperrr: I assume `turnes` should be `turned`

matija-hustic: This was meant as a consequence. Get an underestimate of the distance, in which case `inRange` would become `true` a little sooner than the theoretical "in range" condition is met. I'll reformulate it for clarity.

Unit158: It should be turns, in that case.

matija-hustic:
OK, updated. The TODO comments will be taken care of when the YAML upgrade rule is in place.

abcdefg30:
`OpenRA.Mods.Common/Effects/Missile.cs(382,36): error CS0103: The name "shadowPos" does not exist in the current context`

matija-hustic:
I've removed the missile shadow part cause it's done in #8923

reaperrr:
Gave this some ingame testing, works nice for the most part!

One thing I noticed, though: increasing the AirburstAltitude does not only increase the altitude above, but also distance from target the explosion will happen. With a TS Nod bazooka, an Airburst value of 1024 made the missile explode after little more than half the maximum distance (and RangeLimit was disabled, so that can't be the problem).

:+1: from me after that is adressed.

penev92:
Bump.
@matija-hustic you still around?

matija-hustic:
@penev Yep, just without my computer for another ten days. I'll get to the
PRs as soon as I can. But all of the requested changes are minor if I
recall correctly
Em 20/08/2015 10:53, "Pavel Penev" <notifications@github.com> escreveu:

Unit158:
Any progress?

reaperrr:
@matija-hustic if you're still around, can you give an estimation when you'll be able to update and rebase this?

matija-hustic:
OK, so I've updated this. Hopefully the airburst works a little better now?
@reaperrr, I may not understand 100% what airburst should do. Without airburst, the missile is supposed to explode when inside a ball of radius `CloseEnough` arond the target. With airburst, I've made the missle explode when inside a cilinder of radius `CloseEnough` and height `AirburstAltitude`. Is this approximately what should happen?

Previously airburst involved only height and velocity (to explode only when on the downward part of the trajectory) checks, no distance checks

reaperrr:
> Is this approximately what should happen?

@matija-hustic That's the right direction, but not quite there yet: In TS, airburst missiles would *not* descent towards the target, instead they'd fly towards and (visually at least, the damage was still dealt to the ground but that's out of this PR's scope) explode at targetpos.Z + AirburstAltitude (original TS used some magic hardcoded number of cells above ground, I think).

That being said, I digged in my memory and the only Airburst weapon in TS without the Firestorm addon is the voxel cluster missile, so this feature is just a nice-to-have and far from urgent for sprite missiles, at least for now. If you feel like it, you can try to reproduce the behavior described above of course, but maybe it would be better to split Airburst off to its own PR depending on this one, so this PR won't get blocked over any airburst-related issues (I suspect the warheads might need some fiddling as well to reproduce the behavior of the original).

matija-hustic:
Ah OK! That should then be very easy to do. I'll get to it today.

matija-hustic:
How about now @reaperrr? The missile now aims at exactly what you said, but I've left the condition that it explodes when in that airburst cylinder (if you e.g. launch a missile from a bazooka and it has a low VROT, then the missile wouldn't have to spiral upwards to reach the exact airburst altitude, but can explode even if it's under that altitude). Or should I just remove that condition?

reaperrr:
Yes, this is what I meant. Unless someone else says otherwise, I think that condition is fine.

The rest already worked fine before, so :+1:

GraionDilach:
>  they'd fly towards and (visually at least, the damage was still dealt to the ground but that's out of this PR's scope) explode at targetpos.Z + AirburstAltitude (original TS used some magic hardcoded number of cells above ground, I think).

No, the airburst delivered the damage where it exploded. The second stages are out of this PR's scope though. I'll check the code tomorrow.

matija-hustic:
Now the upgrade rule is in there, too, as well as the changed yamls it yields

reaperrr:
Last changes and upgrade rules look good, but it would be nice if you could make the title and description of the main commit more informative.

Calling it "Missile projectile refactor" and then copy-pasting the list of changes from the PR description would be enough.

matija-hustic:
Done

GraionDilach:
This looks good pretty much overall, however the current setup disallows TS rockets to fire up a cliff - Rocket Infantry can't even fire down if it doesn't stand at the edge.

The cause of that is the low `CruiseAltitude`, the default works on flat terrain, but since a cliff is 4 cell (2048 wu) high in TS the altitude cannot adapt to such a sudden jump immediately. 2048 works nicely with `VerticalROT` 10 on the Rocket Infantry, without being such prevalent jumpiness as your old gif.

I just assume though that HorizontalROT = VerticalROT in TS looks the best.

:+1: when CruiseAltitude and VerticalROT are upped in the TS weapons.

pchote:
In the longer term the missiles will need to probe the ground a few ticks-worth of travel ahead, and start to climb early if there is a sudden change.

matija-hustic:
@GraionDilach, the current VROT and the current cruise altitude of the rocket infantry actually suffices to guide the missile up the cliff (unless the unit is really close to the cliff, but then maybe something should rather be done with the initial launch angle?) as well as to its bottom.

The problem is ultimately in the "in range" condition. It was triggering too soon (so as to leave enough time and place to accommodate a 180° vertical loop before hitting the target; I've done this intentionally as a provision for the "worst case" that could arrive when the missile is guided). I've now changed the trigger to "only" accommodate 90° vertical turns (ascent or descent). This wouldn't work in the mentioned worst case scenario, but will leave enough time for the missile to face and hit its target in most typical cases. Making this condition flexible (i.e. taking into account the exact current vertical facing) would involve a bit more computational complexity (calling an inverse trigonometric function) and would just complicate the code and ideas even more than they already are.

@pchote, I've already considered this before, but thought it might be too computationally expensive. But now I've added a simple look-ahead to the code and I've made the missile target the computed height as soon as possible (when target is out of range).

pchote:
Note that we have lookup-table based trig functions, so they aren't going to be particularly expensive.

reaperrr:
The last change(s) regressed the flight behavior, in my opinion. The missile now sometimes climbs up 100% vertically, which looks very weird, and on flat terrain against targets the missile "in range" check now seems to trigger very late, leading to a very sudden and steep descend towards the target.

> This looks good pretty much overall, however the current setup disallows TS rockets to fire up a cliff - Rocket Infantry can't even fire down if it doesn't stand at the edge.

In my opinion this should be solved by tweaking LaunchAngle and CruiseAltitude instead. I don't like the side-effects the latest revision has introduced.

reaperrr:
Ok, after some more testing it's not as bad as I thought at first.

However, that check for height triggers even if the targeted cell is just close to, but not on top of the cliff. This makes the missile fly like this when targeting a cell close to a cliff: _|\ even if the target is on the same level as the shooter.

matija-hustic:
@reaperrr, do I close this?

reaperrr:
Unless you'd like to take it over again this can be closed, yes.

matija-hustic:
Well I have a new version of the lookahead. When I test it, it avoids almost all obstacles (when physically possible) even without increasing VROT. Plus it tries decelerating to get the vertical turn radius smaller and the missile more easily manoeuvrable.

I might have made it too complex tho... I can push it to this PR so you guys can see if it's something that'd be desired or not

matija-hustic:
I've put the code up if anyone feels like testing it out. It still needs simplification and comments, but the functionality is there...

reaperrr:
*Much* better :)

The only thing I noticed is that sometimes the TS bazooka flies past the target cell if it's up a cliff and performs a circle towards it, but that should be fixable by tweaking VerticalROF and speed.

:+1: with the TrailImage reference fixed.

matija-hustic:
I'm making the code simpler, smaller and cleaner. I'm also trying to get the maximum out of the default VROT. I'll increase it then if it's impossible to avoid the missiles circling.

OpenRA.Game/WAngle.cs
```diff
@@ -61,6 +61,24 @@ public int Tan()
 			return new WAngle(Angle - 512).Tan();
 		}
 
+		public static WAngle ArcCos(int cos)
+		{
+			if (cos < -1024 || cos > 1024)
+				return new WAngle();
```
pchote: Return `WAngle.Zero` instead of constructing new ones.

OpenRA.Game/WAngle.cs
```diff
@@ -61,6 +61,24 @@ public int Tan()
...
+
+			var add = 0;
+			if (cos < 0)
+			{
+				cos = -cos;
+				add = 256;
+			}
+
+			for (var angle = 0; angle < 1024; angle++)
```
pchote: `CosineTable` only has 256 entries.

pchote:
This looks pretty good ingame (based on *very* limited testing).  One suggestion, which can be left to a later PR if you don't want to do it now, is to account for the vertical angle in the sprite facing calculation like we do for bullets so that the missile image is aligned with the apparent direction of travel.

matija-hustic:
The code is a little cleaner now. There are a couple of remaining TODOs.
Two of them are important and I need help with them:
* How do I decide which variables should have [Sync] next to them?
* How do I get an integer square root of an integer (deterministically)?

The other ones include
* possibly unhardcoding some values from the homing code,
* dynamically selected initial speed and launch angle (better interaction of units firing near cliffs),
* tweaking the missile's rendering to account for the vertical angle.
These can be left for later if there's a hurry to get the PR off the queue.

pchote:
> How do I decide which variables should have [Sync] next to them?

As few as you can get away with while still catching any desyncs early.  The position and facing are probably sufficient.

> How do I get an integer square root of an integer (deterministically)?

`Exts.ISqrt`

matija-hustic:
>The position and facing are probably sufficient.

So I can remove `[Sync]` from the other ones (e.g. `targetPosition`, `offset`, etc. that were there before I modified `Missile.cs`)?

penev92:
I see this was silently updated.

obrakmann:
Alright, I won't try to understand all the logic behind the calculations in this PR, but it certainly looks good in-game. The cliff avoidance in TS works, and the missiles in RA still do what they're supposed to do (jammers work, ballistic missiles from subs still work, the rests, too).

Still, there are two issues (well, probably just one): the missiles from the Nod bike and stealth tank aren't rendered when the target is below or at the same height as the unit. They do show up when the target is higher. That should be fixed, then it gets my :+1: as well.

reaperrr:
Needs rebase as well.

matija-hustic:
Making this thing know how to avoid obstacles is proving to be a handful. As it turns out, the previous version of the lookahead didn't handle firing *downhill* properly...
Now it should finally handle both uphill and downhill firing with some intelligence. But I'll need some more time for testing and going through the code.

reaperrr:
Retested since there have been a few updates since I last looked at this.

The look-ahead works fine now, the vertical angle is factored into the sprite facing and the rest still works. So I'm renewing my :+1:

obrakmann:
Yes, works fine now, :+1:

matija-hustic:
I wanted to clean up the code some more. And make sure no other problems appear... Are you sure you want to merge it in the state it is in?

Can I file a followup PR then? In particular I'm afraid that some of my calls to `ISqrt` might pass a negative number.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/e93b940ca138a356ba4d292f7bea89cc4d98d6b7)

obrakmann:
@matija-hustic: oh sorry, I wasn't aware of that. But sure, file a follow-up. It all seemed to work nicely for me, though.

matija-hustic:
OK, I'll do that asap. And next time I'll ask for a "Has dependencies" label. I just hope the `ISqrt` calls doesn't cause problems.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -152,44 +381,402 @@ bool JammedBy(TraitPair<JamsMissiles> tp)
...
+					if (!WillClimbWithinDistance(vFac, loopRadius, predClfDist, diffClfMslHgt)
+						&& !(predClfDist <= loopRadius * (1024 - WAngle.FromFacing(vFac).Sin()) / 1024
+							&& WillClimbAroundInclineTop(vFac, loopRadius, predClfDist, diffClfMslHgt, speed)))
+					{
+						desiredVFacing = vFac + 1;
+						break;
+					}
+
+			// Attained height after ascent as predicted from upper part of incline surmounting manoeuvre
+			var predAttHght = loopRadius * (1024 - WAngle.FromFacing(vFacing).Cos()) / 1024 - diffClfMslHgt;
```
pchote: If you do a followup pr, could you please put in some more readable names?  We have plenty of spare vowels available for use in variables ;)

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -152,44 +381,402 @@ bool JammedBy(TraitPair<JamsMissiles> tp)
...
+			// Missile will not climb terrAltDiff w-units within hHeightChange w-units
+			// all the while ending the ascent with vertical facing 0
+			else if (!WillClimbWithinDistance(vFacing, loopRadius, predClfDist, diffClfMslHgt))
+
+				// Find smallest vertical facing, attainable in the next tick,
+				// for which the missile will be able to climb terrAltDiff w-units
+				// within hHeightChange w-units all the while ending the ascent
+				// with vertical facing 0
+				for (var vFac = System.Math.Min(vFacing + info.VerticalRateOfTurn - 1, 63); vFac >= vFacing; vFac--)
+					if (!WillClimbWithinDistance(vFac, loopRadius, predClfDist, diffClfMslHgt)
```
pchote: Also avoid hanging blocks - if you have an inner statement using braces, then please also put braces around the outer blocks.

OpenRA.Mods.Common/Effects/Missile.cs
matija-hustic: Sure :) I would have already left the vowels before were I not afraid of the names being excessively long and ugly. E.g. `predAttHght` to `predictedAttainedHeight` basically doubles the length

GraionDilach:
I'm just here to state that @matija-hustic is a hero:tm:, no more, no less.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -81,27 +128,51 @@ public class MissileInfo : IProjectileInfo
...
+			Homing,
+			Hitting
+		}
+
 		readonly MissileInfo info;
 		readonly ProjectileArgs args;
 		readonly Animation anim;
 
+		// NOTE: Might be desirable to unhardcode the number -10
+		readonly WVec gravity = new WVec(0, 0, -10);
```
reaperrr: If you do unhardcode this in a follow-up (I like the idea), personally I'd suggest leaving the minus here and only unhardcode the number, I think that would be more intuitive for modders (higher value = stronger gravity).

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -152,44 +381,402 @@ bool JammedBy(TraitPair<JamsMissiles> tp)
...
+								desiredVFacing = (sbyte)OpenRA.Traits.Util.GetFacing(vDist, vFacing);
+								if (desiredVFacing < 0 && info.VerticalRateOfTurn < (sbyte)vFacing)
+									desiredVFacing = 0;
+							}
+						}
+					}
+					else
+					{
+						// Aim for the target
+						var vDist = new WVec(-relTarHgt, -relTarHorDist * (targetPassedBy ? -1 : 1), 0);
```
Mailaender: Coverity detected dead code here. https://scan8.coverity.com/reports.htm#v24414/p10888/fileInstanceId=8951757&defectInstanceId=3159905&mergedDefectId=131343

hypercube33:
This seriously junked up the SAM sites for NOD / CnC Gold FYI. The SAMs now shoot missles into the ground short of their targets.

PR #7873, opened 2015-04-05
Replace old "country" and "race" name in the code with the generic "faction"

penev92:
I finally got around to doing this.
I replaced all occurrences of "country", "countries", "race" and "races" with "faction" in file names, type names, code and UI-related YAML.
Still need to add **a ton** of YAML upgrade code to upgrade the actor rules.

I will probably squash the commits once everything is done, because they are _a lot_.

Closes  #7552.

PR #8718, opened 2015-07-15
Rename Country to Faction in Player.cs

penev92:
Yet another piece of #7873.
Split into nice small commits for easy reviewing.

### DIFF
--- a/OpenRA.Game/Player.cs
+++ b/OpenRA.Game/Player.cs
@@ -80,7 +80,7 @@ namespace OpenRA
 			return factions.FirstOrDefault(f => f.InternalName == factionName) ?? factions.First();
 		}
 
-		public Player(World world, Session.Client client, Session.Slot slot, PlayerReference pr)
+		public Player(World world, Session.Client client, PlayerReference pr)
 		{
 			string botType;
 
--- a/OpenRA.Game/Player.cs
+++ b/OpenRA.Game/Player.cs
@@ -51,42 +51,42 @@ namespace OpenRA
 
 		readonly IFogVisibilityModifier[] fogVisibilities;
 
-		FactionInfo ChooseCountry(World world, string name, bool requireSelectable = true)
+		static FactionInfo ChooseFaction(World world, string name, bool requireSelectable = true)
 		{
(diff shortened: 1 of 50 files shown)


OpenRA.Game/Player.cs
```diff
@@ -193,7 +191,7 @@ public void OnScriptBind(ScriptContext context)
 		public LuaValue Equals(LuaRuntime runtime, LuaValue left, LuaValue right)
 		{
 			Player a, b;
-			if (!left.TryGetClrValue<Player>(out a) || !right.TryGetClrValue<Player>(out b))
```
Mailaender: This seems unrelated.

OpenRA.Game/Player.cs
```diff
@@ -193,7 +191,7 @@ public void OnScriptBind(ScriptContext context)
...
+			if (!left.TryGetClrValue(out a) || !right.TryGetClrValue(out b))
```
Mailaender: I am just curious why you did this change.

Mailaender:
Looks good. :+1: Seems to be just internal renamings.

Mailaender:
The Lua API also needs an update. https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Scripting/Properties/PlayerProperties.cs#L34 Probably again keeping the old one printing a note that this will be removed soon.

OpenRA.Game/Player.cs
penev92: Those were marked as redundant, because the expected type is already specified on the line above: `Player a, b;`.

penev92:
I'll address that in the next PR, as it is an omission from the previous one, not this one.

abcdefg30:
Looks good to me too. :+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/db824811bcabdc6f2cf5c011790022ea678abf0f...9e58de684b24bb02307d6405d1971b3480fed5de)

PR #7733, opened 2015-03-24
Disable legacy sequence parsing in D2K.

pchote:
This disables the magic sprite filename resolution (#5272) for D2K and removes the unused parachute sequence definition.  `OpenRA.Utility.exe d2k --check-sequence-sprites` now passes for D2K.

PR #8719, opened 2015-07-15
Fixes D2K not finding paratrooper shadow sequence

matija-hustic:
Fixes regression from #8339

https://github.com/OpenRA/OpenRA/pull/8339/files#r34637358

The engine wasn't able to locate the sequence. I haven't gone deeply into it, but it seems all sequences in d2k/sequences/misc.yaml specify the exact file where the sequence resides...

### DIFF
--- a/mods/d2k/sequences/misc.yaml
+++ b/mods/d2k/sequences/misc.yaml
@@ -386,5 +386,5 @@ sandcraters:
 		Offset: -16,-16
 
 parach-shadow:
-	idle:
+	idle: parach-shadow.shp
 		Length: *


penev92:
Works :+1:

reaperrr:
I think the d2k mod defaults to `.r8` extension, that's why it didn't find the `.shp`.

pchote:
>  it seems all sequences in d2k/sequences/misc.yaml specify the exact file where the sequence resides...

This has been the case ever since #7733.

matija-hustic:
Changed the name

abcdefg30:
Looks good to me two. :+1: / :white_check_mark: 
I get another crash when paradroping though.
I filed another followup PR: #8720

PR #8725, opened 2015-07-16
Step towards removing RenderSimple and its derivatives (Part 1)

reaperrr:
This PR gets some of the easy stuff out of the way to reduce the diff and improve reviewability of the follow-ups.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithSpriteBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithSpriteBody.cs
@@ -81,8 +81,8 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void PlayCustomAnimationRepeating(Actor self, string name)
 		{
-			DefaultAnimation.PlayThen(name,
-				() => PlayCustomAnimationRepeating(self, name));
+			var sequence = NormalizeSequence(self, name);
+			DefaultAnimation.PlayThen(sequence, () => PlayCustomAnimationRepeating(self, sequence));
 		}
 
 		public void PlayCustomAnimationBackwards(Actor self, string name, Action after = null)
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -404,7 +404,7 @@
     <Compile Include="Traits\Render\RenderBuilding.cs" />
     <Compile Include="Traits\Render\RenderBuildingCharge.cs" />
     <Compile Include="Traits\Render\RenderBuildingTurreted.cs" />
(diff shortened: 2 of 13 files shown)


GraionDilach:
I'd say the YAML changes of the mines should be extracted out to a new PR with removing the RA1 mod defaults.

OpenRA.Mods.Common/Traits/Render/WithSpriteBody.cs
```diff
@@ -81,7 +81,7 @@ public void PlayCustomAnimation(Actor self, string name, Action after = null)
 
 		public void PlayCustomAnimationRepeating(Actor self, string name)
 		{
-			DefaultAnimation.PlayThen(name,
+			DefaultAnimation.PlayThen(NormalizeSequence(self, name),
```
pchote: You normalize this one, but not the one below?

reaperrr:
Updated.

Mailaender:
Looks sensible. :+1: Editor only rendering still works. :white_check_mark:

OpenRA.Mods.Common/Traits/Render/WithSpriteBody.cs
```diff
@@ -81,8 +81,8 @@ public void PlayCustomAnimation(Actor self, string name, Action after = null)
 
 		public void PlayCustomAnimationRepeating(Actor self, string name)
 		{
-			DefaultAnimation.PlayThen(name,
-				() => PlayCustomAnimationRepeating(self, name));
+			DefaultAnimation.PlayThen(NormalizeSequence(self, name),
```
pchote: Harvesting animations now appear to be double-normalized.  You should be able to adjust WithHarvestAnimation to account for these upstream checks (which I think are a good idea).

reaperrr:
Updated.
For the record, I did test the harvester normalization in combination with damage stages, seems to work as intended.

reaperrr:
Updated, removed the `RenderSimple` commit. It will later return in its own PR, where `RenderSimple` will actually be removed from code as well. That PR will follow the complete removal of `RenderBuilding`.

reaperrr:
Updated.

penev92:
Needs a rebase.

penev92:
Looks and works OK.
:+1: when my comment on the last commit is addressed. Also when rebased, of course.

reaperrr:
Updated and rebased.

reaperrr:
Updated.

penev92:
Thanks.
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f307594884e7c8cd2552a8e857bade75ae05d0e6...c9903db9f8a964dc11006a3306022f18e3827045)

Issue #8722, opened 2015-07-15
Allies03a: Imprisoned infantry does not get killed by exploding barrels


obrakmann:

mistikal:
For what it is worth, I attempted to reproduce the issue, but mine seem to die as expected. Tested with the medic, dies - tested with the engineer and the infantry, both die.

obrakmann:
Nope, it's really bugged. It was introduced with acc4b307b63ee76c229051143fb28fcbab1a7af4. It seems that disabling `TargetableUnit` is a bad idea (ie. it makes units pretty much invulnerable)

abcdefg30:
Seems like the problem is even bigger:

> 21:21 (+pchote) antares79: don't we cache TargetableUnit etc in a bunch of cases? 
21:21 (+pchote) that won't play nicely with turning them on/off 
21:22 (+antares79) pchote: with 'disable' I meant disabling using upgrades. that would be fine with caching, unless I'm terribly mistaken 
21:22 (+pchote) if you disable one without enabling another then you will have no targeting trait 
21:23 (+pchote) which means nothing can attack it 
21:23 (+antares79) hence the bug 
21:23 (+pchote) yup 
21:23 (+pchote) ...although, unless i'm missing something, I would call it intentional design 
21:23 (GraionDilach) ah yeah, multiple TargetableUnits still doesn't work lol 
21:23 (+antares79) yeah. it's just the wrong use case ;) 
21:24 (+pchote) what is that mission supposed to be doing? 
21:26 (+antares79) there are some infantry units of yours that are held captive by surrounding enemy units. those enemy units must be prevented from shooting at the captives somehow. previously, Cloak was used, since a few days ago it's TargetableUnit 
21:26 (+pchote) just make them not autotargetable 
21:26 (+pchote) via an upgrade 
21:27 (+antares79) yeah. or revert back to cloaking, that worked well. I don't know why it was changed. 
21:28 (+pchote) I think i remember seeing some comments about that being considered a hack 
21:55 (abcdefg30) anyway, i'm for support of multiple targetable unit traits 
21:55 (abcdefg30) we'll need that anyway for the targeting of airborne?

PR #8727, opened 2015-07-16
TargetTypes by Upgrade and Replacement of Targetable* by Targetable and ITargetablePositions

### LINKED ISSUE #8724: Proposal to refactor TargetType handling
Currently we have like 4-5 traits which handle TargetTypes on actors.
TargetableBuilding returns the cells
TargetableUnit works on units
and multiple subclasses of TargetableUnit, all working around the fact that an actor can have only one Targetable* trait.

I honestly didn't really looked into the code much, but from what I saw, merging TargetableBuilding and TargetableUnit is possible (via moving the Building() check in the latter) and then instead of having TargetableAircraft etc, we could set up their source traits granting an upgrade to enable/disable targeting/affection. (Helicopter granting an upgrade at height 200 for example)

This will be important for the RA2 mod because at there unit-targeting could change on the fly (Temporal weapons made the target only affected by such, an already mindcontrolled unit cannot be remindcontrolled, IronCurtained units weren't attacked by the AI, units in-air (paradropped or flying) couldn't get targeted by strictly ground weapons, etc).

This would also render map scripting easier.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24400080-proposal-to-refactor-targettype-handling?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

atlimit8:
Closes #8724, unless there is something else to it.


atlimit8:
Closing for replacement.

atlimit8:
Note: the first commit is being replaced - different system.

GraionDilach:
Cannot this be extended to TargetableBuilding as well?

atlimit8:
I simply missed that one, but plan on including it now.

pchote:
See my point (1) in #8724.  Merging TargetableBuilding into this is asking for trouble.

atlimit8:
My plan is create ITargetablePositions and have Mobile, Building, and Aircraft implement it.

atlimit8:
I know the TargetTypes is updated by upgrade changes. I used Console.WriteLine to double check. However, I'm not sure the targeting system always recognizes the changes.

Please check whether landing aircraft has an effect on targetability.

OpenRA.Mods.Common/Activities/Air/HeliFly.cs
```diff
@@ -53,6 +53,9 @@ public override Activity Tick(Actor self)
 			if (IsCanceled || !target.IsValidFor(self))
 				return NextActivity;
 
+			if (helicopter.CenterPosition.Z == 0)
```
pchote: #8866 introduced an `IsAtGroundLevel()` that works correctly for mods that use heightmaps.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -188,6 +192,8 @@ public void AddedToWorld(Actor self)
 			self.World.ActorMap.AddInfluence(self, this);
 			self.World.ActorMap.AddPosition(self, this);
 			self.World.ScreenMap.Add(self);
+			if (CenterPosition.Z > 0)
```
pchote: These z checks should be replaced with IsAtGroundLevel too.

atlimit8:
Manages airborne upgrades via Aircraft.IsAirborne property now.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -172,6 +204,7 @@ public void SetPosition(Actor self, WPos pos)
 			{
 				self.World.ScreenMap.Update(self);
 				self.World.ActorMap.UpdatePosition(self, this);
+				IsAirborne = self.World.Map.DistanceAboveTerrain(CenterPosition).Length != 0;
```
Phrohdoh: `> 0`, just in the off chance that someone makes a subterranean aircraft that doesn't break before this.

atlimit8:
Added MinAirborneAltitude to Aircraft trait to set airborne minimum altitude.

abcdefg30:
Needs a rebase.

atlimit8:
Rebased.

abcdefg30:

As a (nice) side-effect you can't force fire at the prisoners in allies03 anymore.

atlimit8:
Changed.

Mailaender:
Needs a rebase.

abcdefg30:
Needs a rebase again.

atlimit8:
Rebased.

abcdefg30:
Needs a rebase again.

atlimit8:
Rebased.
It should be noted that this PR replaces the prisoner cloak hack in the Allies 03 missions but without causing #8722 like #8662 did.

atlimit8:
Rebased again.
Thanks #9236.

Mailaender:
Aircraft and submarine targeting still works. I can also confirm the fix for Allies 03. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/2cfc5c677533d94d8b2a47e8b419da945dd8740c...2f356e99955907302f98381ac31fd4856fa215f1)

PR #8728, opened 2015-07-16
Replaced RenderGunboat with WithGunboatBody

reaperrr:
For consistency with other actor render traits.

Note: I'm aware that this is slightly hacky, but I've already spent far too much time on this special case that will likely never be used for anything but the TD gunboat, so unless someone can give me a quick and easy solution to get rid of the work-arounds, I don't intend to make any more time-consuming changes.


pchote:
Quick and slightly less hacky solution: Have a Tick method that checks for facing changes and then changes the active body and wake animations when it changes direction.

reaperrr:
Updated.

pchote:
Part two (which I should have been explicit about, sorry) is to use `DefaultAnimation` instead of `body` and drop the `HideDefaultSequence` stuff.

reaperrr:
> Part two (which I should have been explicit about, sorry) is to use DefaultAnimation instead of body and drop the HideDefaultSequence stuff.

How do I make `DefaultAnimation` use turret facing?

pchote:
Hmm, good point.  This is also a problem for turrets (the base defense), so I guess we'll have to live with the yaml hack.

OpenRA.Mods.Cnc/Traits/Render/WithGunboatBody.cs
```diff
@@ -0,0 +1,80 @@
...
+			wake.PlayRepeating(info.WakeLeftSequence);
+			rs.Add(new AnimationWithOffset(wake, null, null, -87));
+		}
+
+		public void Tick(Actor self)
+		{
+			if (facing.Facing <= 128)
+			{
+				if (body.CurrentSequence.Name != NormalizeSequence(self, info.LeftSequence))
+					body.ReplaceAnim(NormalizeSequence(self, info.LeftSequence));
```
Phrohdoh: Newline, please. And below.

OpenRA.Mods.Common/Traits/Render/WithSpriteBody.cs
```diff
@@ -26,6 +26,9 @@ public class WithSpriteBodyInfo : UpgradableTraitInfo, IRenderActorPreviewSprite
 		[Desc("Animation to play when the actor is idle.")]
 		[SequenceReference] public readonly string Sequence = "idle";
 
+		[Desc("Should the above default Sequence be hidden.")]
```
Phrohdoh: This should end with `?` as it is a question.

reaperrr:
> Hmm, good point. This is also a problem for turrets (the base defense), so I guess we'll have to live with the yaml hack.

Actually no, 'turrets' was a good buzzword. I had already replaced `RenderBuildingTurreted` on another branch and the solution from there works here, too, so I was able to get rid of a good portion of hackyness.

OpenRA.Mods.Cnc/Traits/Render/WithGunboatBody.cs
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+
+	class WithGunboatBody : WithSpriteBody, ITick
+	{
+		readonly WithGunboatBodyInfo info;
+		readonly Animation wake;
+		readonly RenderSprites rs;
+		readonly IFacing facing;
+		Turreted turret;
```
Mailaender: This can be readonly as well I guess.

OpenRA.Mods.Cnc/Traits/Render/WithGunboatBody.cs
```diff
@@ -0,0 +1,88 @@
...
+
+using System;
+using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Cnc.Traits
+{
+	class WithGunboatBodyInfo : WithSpriteBodyInfo, Requires<IBodyOrientationInfo>, Requires<IFacingInfo>
```
Mailaender: `init.Self.TraitsImplementing<Turreted>().First` makes me think this also needs a `Requires<Turreted>`.

reaperrr:
Updated.

Mailaender:
Gunboat still works. :white_check_mark:

OpenRA.Mods.Cnc/Traits/Render/WithGunboatBody.cs
```diff
@@ -0,0 +1,88 @@
...
+			wake = new Animation(init.World, name);
+			wake.PlayRepeating(info.WakeLeftSequence);
+			rs.Add(new AnimationWithOffset(wake, null, null, -87));
+		}
+
+		public void Tick(Actor self)
+		{
+			if (facing.Facing <= 128)
+			{
+				if (DefaultAnimation.CurrentSequence.Name != NormalizeSequence(self, info.LeftSequence))
```
pchote: You can pull out a `var leftSequence = NormalizeSequence(self, info.LeftSequence)` to avoid doing this twice if it does change.  Same for `rightSequence` below.

reaperrr:
Updated.

pchote:
Code looks fine, and no regressions ingame :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c040b3c970c2be0591890b6f0f045bdcae7db3a9%5E...c040b3c970c2be0591890b6f0f045bdcae7db3a9

PR #8729, opened 2015-07-16
Reduce allocations

RoosterDragon:
A set of changes that avoid allocating memory where it can be avoided during the main game loop to reduce GC pressure. However I have attempted to avoid making any "clever" changes.

See individual commits for details - these changes combined save ~16% in memory allocations in the main game loop (if you have something with a circle range highlight selected).

### DIFF
--- a/OpenRA.Game/Graphics/LineRenderer.cs
+++ b/OpenRA.Game/Graphics/LineRenderer.cs
@@ -20,6 +20,7 @@ namespace OpenRA.Graphics
 
 		readonly Renderer renderer;
 		readonly IShader shader;
+		readonly Action renderAction;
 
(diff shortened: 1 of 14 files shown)


Mailaender:
Changes look safe. :+1: Passenger and target lines still behave properly. :white_check_mark:

pchote:
Changes look good, and no visible changes ingame :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/ef54947f068eb5ed9fb5783c370354d0e1cd9467%5E...ef54947f068eb5ed9fb5783c370354d0e1cd9467

PR #8731, opened 2015-07-17
Simplifies Start button logic in lobby

deniz1a:
This does the same thing but is simpler.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
@@ -297,15 +297,8 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 					orderManager.LobbyInfo.Slots.Any(sl => sl.Value.Required && orderManager.LobbyInfo.ClientInSlot(sl.Key) == null);
 				startGameButton.OnClick = () =>
 				{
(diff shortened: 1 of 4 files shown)


Unit158:
:-1: this still doesn't fix the issue.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -297,15 +297,8 @@ void CloseWindow()
...
-					{
-						var cl = orderManager.LobbyInfo.ClientInSlot(sl.Key);
-
-						// Bots and admins don't count
-						return cl != null && !cl.IsAdmin && cl.Bot == null && !cl.IsReady;
-					};
-
-					if (orderManager.LobbyInfo.Slots.Any(notReady))
+					// Bots and admins don't count
+					if (orderManager.LobbyInfo.Clients.Any(c => c.Slot != null && !c.IsAdmin && c.Bot == null && !c.IsReady))
```
pchote: This doesn't do the same thing.  The old code excluded observers, which don't have a slot associated with them.

deniz1a: What about ```c => c.Slot != null```?

abcdefg30:
Looks good. :+1:

RoosterDragon:
:+1: Spectators are handled correctly.

PR #8733, opened 2015-07-17
Restore the original setup of the menu music in the RA mod.

GraionDilach:


mods/ra/rules/world.yaml
```diff
@@ -4,6 +4,8 @@
 	ActorMap:
 	ScreenMap:
 	MusicPlaylist:
+		StartingMusic: intro
```
pchote: I guess that this is in transit.mix, because I don't have it.  Some of our content mirrors don't include transit.mix (but some do, IIRC!), so we're going to need to fix that and then add it to our content file check in mod.yaml.

pchote:
:+1: to the idea here, but we're going to need to fix our content distribution before we can do this!

mods/ra/rules/world.yaml
GraionDilach: There are no transit.mixes in RA afaik. It's in my scores.mix

Mailaender:
You are confusing multiple things here. `transmit.mix` missing is a problem of https://github.com/OpenRA/OpenRAWeb/blob/master/content/packages/cnc-mirrors.txt and @ihptru already fixed his package.

Mailaender:
This doesn't just change the menu music, but also plays it in-game during skirmishes and missions.

pchote:
@Mailaender: Like I said above, the changes requested tag was because my content download did not not include the "intro" song.  I suggested that it might be transit.mix, but apparently it is some other issue.

mods/ra/rules/world.yaml
Mailaender: We only have hell march in the initial scores.mix, but I don't see how that is a problem here.

pchote: I had most of the songs, but not intro.  It would then (as expected) pick a random track (not hellmarch) at startup.

dan9550:
We should really consolidate our mix files here. There are heaps of different variations of modified scores.mix files floating around :\
We should i guess aim to support those from the original game CDs and from our own mirrors.

pchote:
#8746 outlines a plan to fix the issue I brought up above.

pchote:
Superseded by #8876.

PR #8734, opened 2015-07-17
Hide the songs which were hidden in the base games.

GraionDilach:


OpenRA.Game/GameRules/MusicInfo.cs
```diff
@@ -26,6 +27,8 @@ public MusicInfo(string key, MiniYaml value)
 
 			var nd = value.ToDictionary();
 			var ext = nd.ContainsKey("Extension") ? nd["Extension"].Value : "aud";
+			bool t;
+			Hidden = nd.ContainsKey("Hidden") && bool.TryParse(nd["Hidden"].Value, out t) ? t : false;
```
pchote: This can be simplified to
```
if (nd.ContainsKey("Hidden"))
	bool.TryParse(nd["Hidden"].Value, out Hidden);
```

pchote:
This interacts badly with the music player, which allows you to control the shellmap music but otherwise shows an empty list in TD if you don't have scores.mix installed.

pchote:
Crash when starting a skirmish in TD with transit.mix but not scores.mix:
```
Exception of type `System.InvalidOperationException`: Sequence contains no elements
  at System.Linq.Enumerable.First[TSource] (IEnumerable`1 source) [0x00067] in /private/tmp/source-mono-mac-crypto-32/bockbuild-xamarin/profiles/mono-mac-xamarin/build-root/mono-4.3.0/external/referencesource/System.Core/System/Linq/Enumerable.cs:930 
  at OpenRA.Traits.MusicPlaylist..ctor (OpenRA.World world, OpenRA.Traits.MusicPlaylistInfo info) [0x00158] in /Users/paul/src/OpenRA/OpenRA.Game//Traits/World/MusicPlaylist.cs:75 
  at OpenRA.Traits.MusicPlaylistInfo.Create (OpenRA.ActorInitializer init) [0x00008] in /Users/paul/src/OpenRA/OpenRA.Game//Traits/World/MusicPlaylist.cs:38 
  at OpenRA.Actor..ctor (OpenRA.World world, System.String name, OpenRA.Primitives.TypeDictionary initDict) [0x000c8] in /Users/paul/src/OpenRA/OpenRA.Game//Actor.cs:95 
  at OpenRA.World.CreateActor (Boolean addToWorld, System.String name, OpenRA.Primitives.TypeDictionary initDict) [0x00004] in /Users/paul/src/OpenRA/OpenRA.Game//World.cs:245 
  at OpenRA.World.CreateActor (System.String name, OpenRA.Primitives.TypeDictionary initDict) [0x00005] in /Users/paul/src/OpenRA/OpenRA.Game//World.cs:240 
  at OpenRA.World..ctor (OpenRA.Map map, OpenRA.Network.OrderManager orderManager, WorldType type) [0x00134] in /Users/paul/src/OpenRA/OpenRA.Game//World.cs:187 
  at OpenRA.Game.StartGame (System.String mapUID, WorldType type) [0x00074] in /Users/paul/src/OpenRA/OpenRA.Game//Game.cs:150 
  at OpenRA.Network.UnitOrders.ProcessOrder (OpenRA.Network.OrderManager orderManager, OpenRA.World world, Int32 clientId, OpenRA.Order order) [0x004da] in /Users/paul/src/OpenRA/OpenRA.Game//Network/UnitOrders.cs:110 
  at OpenRA.Network.OrderManager.TickImmediate () [0x00119] in /Users/paul/src/OpenRA/OpenRA.Game//Network/OrderManager.cs:130 
  at OpenRA.Sync+<CheckSyncUnchanged>c__AnonStorey0.<>m__0 () [0x00007] in /Users/paul/src/OpenRA/OpenRA.Game//Sync.cs:168 
  at OpenRA.Sync.CheckSyncUnchanged[T] (OpenRA.World world, System.Func`1 fn) [0x00008] in /Users/paul/src/OpenRA/OpenRA.Game//Sync.cs:176 
  at OpenRA.Sync.CheckSyncUnchanged (OpenRA.World world, System.Action fn) [0x0001b] in /Users/paul/src/OpenRA/OpenRA.Game//Sync.cs:168 
  at OpenRA.Game.InnerLogicTick (OpenRA.Network.OrderManager orderManager) [0x0011c] in /Users/paul/src/OpenRA/OpenRA.Game//Game.cs:462 
  at OpenRA.Game.LogicTick () [0x0004e] in /Users/paul/src/OpenRA/OpenRA.Game//Game.cs:507 
  at OpenRA.Game.Loop () [0x000d4] in /Users/paul/src/OpenRA/OpenRA.Game//Game.cs:631 
  at OpenRA.Game.Run () [0x00042] in /Users/paul/src/OpenRA/OpenRA.Game//Game.cs:671 
  at OpenRA.Program.Run (System.String[] args) [0x00011] in /Users/paul/src/OpenRA/OpenRA.Game//Support/Program.cs:111 
  at OpenRA.Program.Main (System.String[] args) [0x00050] in /Users/paul/src/OpenRA/OpenRA.Game//Support/Program.cs:39 
```

Unit158:
why do we want this?

clemty:
> why do we want this?

in TD you have the shellmap theme, and a short "win" jingle which would eventually play if you set the background music to random

OpenRA.Mods.Common/Widgets/Logic/MusicPlayerLogic.cs
```diff
@@ -41,12 +41,12 @@ public MusicPlayerLogic(Widget widget, Ruleset modRules, World world, Action onE
 			var playButton = panel.Get<ButtonWidget>("BUTTON_PLAY");
 			playButton.OnClick = Play;
 			playButton.IsDisabled = noMusic;
-			playButton.IsVisible = () => !Sound.MusicPlaying;
+			playButton.IsVisible = () => { !Sound.MusicPlaying && currentSong != null; };
```
GraionDilach: Could someone tell me how to write this right? Travis/Appveyor both throw a syntax error here but my MonoDevelop compiles it - I guess it's a mono bug and I don't want to feel more stupid than I am and do x amount of commits to guess out the proper syntax.

pchote: This should be disabling the buttons (playButton.IsDisabled) instead of hiding them.  As for the actual syntax error, just drop the `{}`.  This change covers part of #8746.

GraionDilach: Wouldn't implementing #8746 as part of this PR fell into scope-creeping or still acceptable?

pchote: I mean that this change implements part of that ticket.

OpenRA.Mods.Common/Widgets/Logic/MusicPlayerLogic.cs
```diff
@@ -36,29 +36,30 @@ public MusicPlayerLogic(Widget widget, Ruleset modRules, World world, Action onE
 			BuildMusicTable();
 
 			Func<bool> noMusic = () => !musicPlaylist.IsMusicAvailable;
+			Func<bool> disableControls = () => !musicPlaylist.IsMusicAvailable || musicPlaylist.AmbientMusic;
```
GraionDilach: My debugger throws garble to this - although it compiles, from my tests, I'd assume the code only checks for `IsMusicAvailable`. Dun ask me why, I think I really can't grasp this `Func<bool>`.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -45,8 +45,11 @@ public class MusicPlaylist : INotifyActorDisposing, IGameOver
 		readonly MusicInfo[] random;
 		readonly MusicInfo[] playlist;
 
+		public readonly bool IsMusicInstalled;
 		public readonly bool IsMusicAvailable;
 
+		public bool AmbientMusic;
```
pchote: I would define this as `public bool CurrentSongIsAmbient { get; private set; }` to make its purpose clearer.

This is going a step further than we needed for this pr, but since it's here we might as well set it up in preparation for the next changes.

OpenRA.Mods.Common/Widgets/Logic/MusicPlayerLogic.cs
pchote: You can think of this as defining a function:
```
bool _DisableControls()
{
	 return !musicPlaylist.IsMusicAvailable || musicPlaylist.AmbientMusic;
}
```

which is then assigned to the `disableControls` variable.   Executing `disableControls()` is equivalent to running this implicit (anonymous, in programming lingo) function `_DisableControls()`.

GraionDilach: Then this should work, yet if you go into the TD mod's music menu with music installed, you can pretty much pick any song to replace the music.

Thinking of it, I also forgot to override the click-on-songname-to-play-it-action.

pchote: I think my earlier comments must have confused you somewhere.  We can discuss this on IRC later tonight.

OpenRA.Game/Traits/World/MusicPlaylist.cs
GraionDilach: This must be exposed to Lua. I'm not certain if it should be passed along with the PlayMusic calls or should have an own function to be set.

penev92: *`bool IsCurrentSongAmbient`

pchote:
I didn't intend for you to start on the ambient music fixes here (my initial comment was an observation, not a request!), but since you've started then let's get it right here.

The idea is that this is completely separate to the StartingMusic, and defines what should be played in the absence of any other music.  It doesn't make sense to have a lua API to actively play an ambient track, and you definitely don't want outside code directly manipulating the playlists IsCurrentSongAmbient property.  The ambient track is then played automatically when the stop function is called, or via the song completed handler if there isn't a new song to queue up.

The public API for this (both to code and to lua) should be a SetAmbientSong method, which then sets the the internal reference for the ambient track, and changes the song only if the (old) ambient track is already playing.  If code wants to set a new ambient track and then force it to play it should call playlist.SetAmbientSong and then playlist.Stop().

OpenRA.Game/Traits/World/MusicPlaylist.cs
pchote: See my comment below with respect to outside code manipulating this flag.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -69,9 +74,13 @@ public MusicPlaylist(World world, MusicPlaylistInfo info)
 			{
 				currentSong = world.Map.Rules.Music[info.StartingMusic];
 				repeat = info.LoopStartingMusic;
+				AmbientMusic = true;
```
pchote: See my comment below about mixing up StartingMusic and AmbientMusic.

OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -126,13 +137,16 @@ public void GameOver(World world)
...
 				return;
 
 			Sound.PlayMusicThen(currentSong, () =>
 			{
-				if (!repeat)
-					currentSong = GetNextSong();
+					if (!repeat)
+					{
+						currentSong = GetNextSong();
+						AmbientMusic = false;
```
pchote: See my comment below with respect to separating ambient from regular music.

GraionDilach:
How would the game know which song is the ambient track?

pchote:
Via the `AmbientMusic` property or the Lua API.

GraionDilach:
So that would be in MusicInfo?

pchote:
`MusicPlaylist`.  I expect these tracks would then also have `Hidden: true` in the music info.

pchote:
I've implemented the background music concept in #8876.  Once that's merged we can rebase this on top for the hidden songs flag.

pchote:
I've folded the changes from here into #8876 so that can fix everything in one go.

PR #7565, opened 2015-02-28
Refactored MobileInfo, Mobile and Move

Rydra:
Refactored MobileInfo, Mobile and Move to make them testable in isolation, easier to understand what they do by commenting and restructuring the code making sure to not cause undesired side-effects.

The fact of having to break dependencies on a highly coupled system forced me to modify and create several interfaces, that at the same time forced me to change several classes directly and indirectly related to Move, Mobile and MobileInfo.

Additionally, I added several unit tests to try how testable they've become.

I've been comenting throughoutfully the code, I've been manipulating, yet to sum up:
- By no means this refactor is complete. It should be completed through the span of several PR. If you take a look at the MoveTests, you'll see that I had to mock a load of dependencies in order to make the tests pass successfully. This is a clear smell of Demeter violation. The construction of Move instances should be delegated to some kind of Move factory that can resolve all the move dependencies for him. This, utimately, will make Move easier to undertand.
- There is no clear barrier and meaning between Mobile and MobileInfo. There are lot of functions that Mobile could deal by himself and yet are delegated to MobileInfo...
- The crush, block, enterCell, finishmove notifications... are implemented in a way that couples actors to each other. We should think in a way to make this work on a subscribe/publish like the Observer Pattern: http://dotnetcodr.com/2013/08/01/design-patterns-and-practices-in-net-the-observer-pattern/ . A message bus implementation of this would be really smart for the future, and would skyroof the performance.

PR #8735, opened 2015-07-17
Minor pathfinder-related changes (preparation for mobile refactor)

reaperrr:
This is a bunch of presumably "safe", easy to review changes from #7565.

### DIFF
--- a/OpenRA.Mods.Common/Pathfinder/PathGraph.cs
+++ b/OpenRA.Mods.Common/Pathfinder/PathGraph.cs
@@ -23,7 +23,7 @@ namespace OpenRA.Mods.Common.Pathfinder
 		/// <summary>
 		/// Gets all the Connections for a given node in the graph
 		/// </summary>
-		ICollection<GraphConnection> GetConnections(CPos position);
+		IEnumerable<GraphConnection> GetConnections(CPos position);
 
 		/// <summary>
 		/// Retrieves an object given a node in the graph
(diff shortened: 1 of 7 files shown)


Mailaender:
Looks good. :+1: Can't notice any changes in-game. :white_check_mark:

OpenRA.Game/Actor.cs
```diff
@@ -172,6 +172,11 @@ public void QueueActivity(Activity nextActivity)
 				currentActivity.Queue(nextActivity);
 		}
 
+		public bool IsAlliedWith(Actor actor)
```
pchote: I'm not convinced about this.  Alliances are defined in terms of players, not actors, and this just adds a duplicate code path.

reaperrr:
Updated. Travis error is not my fault.
Removed `IsAlliedWith` and additionally removed an `as World` and `as Actor` from PathFinder which I had previously overlooked.

OpenRA.Mods.Common/Pathfinder/PathGraph.cs
```diff
@@ -127,17 +127,17 @@ int GetCostToNode(CPos destNode, CVec direction)
...
 				out movementCost,
-				IgnoredActor as Actor,
+				IgnoredActor,
 				checkConditions) && !(CustomBlock != null && CustomBlock(destNode)))
 			{
 				return CalculateCellCost(destNode, direction, movementCost);
 			}
 
-			return InvalidNode;
+			return Constants.InvalidNode;
```
penev92: If you are replacing the `InvalidNode` from this class with `Constants.InvalidNode`, why are you not removing the old one?

reaperrr: I think the original refactor PR forgot about that too (or I just overlooked it). Anyway, fixed.

obrakmann:
:+1:

reaperrr:
Huh? Does github auto-rebase now? Because this definitely had some changes overlapping with #8622.

PR #8741, opened 2015-07-18
Remove shellmap script from d2k

obrakmann:


pchote:
:+1: Looks good to me.

deniz1a:
What was that code in shellmap.lua even doing?

Phrohdoh:
`Effect.ChangeLighting(red, green, blue, ambient)` is the bit that actually changed something.

deniz1a:
So it changed lighting? I looked at the d2k shellmap and couldn't notice anything.

Mailaender:
> No description provided.

What is the reason behind this and why is this a release blocker?

obrakmann:
The reason is that the effect doesn't do a good job of conveying a sunrise to me. It looks more like a fade-in that takes ages to complete, and as a user I'm constantly wondering, "is it still changing, or is something wrong with my eyes"?

I'm not opposed to having the effect at all, but it needs some more work. It would probably be better to start off with regular lighting, then segue into night and back again.

pchote:
I was under the impression that this effect was a testcase and wasn't supposed to be merged in the first place.  There are several places where this effect fits naturally and improves the experience ([e.g. IRC logs](http://logs.openra.net/?year=2015&month=07&day=08#12:36:56)), but it doesn't benefit anything on the d2k shellmap.

reaperrr:
Needs rebase.

Mailaender:
I didn't know it was perceived that bad.

obrakmann:
Rebased

reaperrr:
:+1:

PR #8742, opened 2015-07-18
Support multiple WithDeathAnimation traits

abcdefg30:
We need that if we want different palettes for different death types.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Crushable.cs
+++ b/OpenRA.Mods.Common/Traits/Crushable.cs
@@ -49,7 +49,8 @@ namespace OpenRA.Mods.Common.Traits
 		public void OnCrush(Actor crusher)
 		{
 			Sound.Play(info.CrushSound, crusher.CenterPosition);
-			var wda = self.TraitOrDefault<WithDeathAnimation>();
+			var wda = self.TraitsImplementing<WithDeathAnimation>()
+				.FirstOrDefault(s => s.Info.CrushedSequence != null);
 			if (wda != null)
 			{
 				var palette = wda.Info.CrushedSequencePalette;
--- a/OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
@@ -82,8 +82,7 @@ namespace OpenRA.Mods.Common.Traits
 				var warhead = e.Warhead as DamageWarhead;
 				var damageType = warhead.DamageTypes.Intersect(Info.DeathTypes.Keys).FirstOrDefault();
 				if (damageType == null)
-					throw new Exception("Actor type `{0}` does not define a death animation for weapon with damage types `{1}`!"
-						.F(self.Info.Name, string.Join(", ", warhead.DamageTypes)));
+					return;
 
 				sequence += Info.DeathTypes[damageType];
 			}
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -227,13 +227,21 @@
 		TargetTypes: Ground, Infantry
 	RenderSprites:
 	WithInfantryBody:
-	WithDeathAnimation:
+	WithDeathAnimation@normal:
 		DeathTypes:
 			BulletDeath: 1
 			SmallExplosionDeath: 2
 			ExplosionDeath: 3
-			InfantryExplodeDeath: 4 #not used by any warhead; TODO: check what should use this type and add it!
 			FireDeath: 5
+	WithDeathAnimation@explosion:
+		DeathSequencePalette: effect
+		DeathPaletteIsPlayerPalette: False
+		DeathTypes:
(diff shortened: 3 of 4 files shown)


GraionDilach:
Incase anyone wonders why these changes.

Our die3 was unused in TS - and the animation is a headshot at 90% of the units with only hijacker and cyborg commando excluded - so it's better if we refer that as such.
Especially considering that the actual ExplosionDeath used s_bang34.
EnergyDeath is the RA1 tesla death animation, the only difference is being the format converted from RA1 SHP to TS SHP. TS was shipped with that animation using wrong palette.

This will mean that we enable a unused death animation inserted before TS die3, die4 and die5.

Mailaender:
Looking good. :+1: Infantry still dies properly in Tiberian Sun. :white_check_mark:

OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
```diff
@@ -80,11 +80,15 @@ public void Killed(Actor self, AttackInfo e)
 			if (Info.UseDeathTypeSuffix)
 			{
 				var warhead = e.Warhead as DamageWarhead;
-				var damageType = warhead.DamageTypes.Intersect(Info.DeathTypes.Keys).FirstOrDefault();
-				if (damageType == null)
+				var wda = self.TraitsImplementing<WithDeathAnimation>().FirstOrDefault(s => s.Info.DeathTypes.Keys.Any(k => warhead.DamageTypes.Contains(k)));
```
pchote: This feels really ugly, and is now covered by a lint check.  How about we just remove it?

OpenRA.Mods.Common/Traits/Crushable.cs
```diff
@@ -49,7 +49,7 @@ public void WarnCrush(Actor crusher)
 		public void OnCrush(Actor crusher)
 		{
 			Sound.Play(info.CrushSound, crusher.CenterPosition);
-			var wda = self.TraitOrDefault<WithDeathAnimation>();
+			var wda = self.TraitsImplementing<WithDeathAnimation>().FirstOrDefault(s => s.Info.CrushedSequence != null);
```
pchote: Style nit: Could you please put the `.FirstOrDefault` on a new line, indented by one additional level?

abcdefg30:
Updated.

OpenRA.Mods.Common/Traits/Crushable.cs
penev92: I've noticed that you like linebreaks, but that's not long enough to need one. :/

mods/ts/rules/defaults.yaml
```diff
@@ -227,13 +227,21 @@
...
 	RenderSprites:
 	WithInfantryBody:
-	WithDeathAnimation:
+	WithDeathAnimation@normal:
 		DeathTypes:
 			BulletDeath: 1
 			SmallExplosionDeath: 2
-			ExplosionDeath: 3
-			InfantryExplodeDeath: 4 #not used by any warhead; TODO: check what should use this type and add it!
+			HeadshotDeath: 3 #not used by any warhead; TODO: check what should use this type and add it!
```
penev92: Something is wrong here.
You are removing the currently unused death type, but keeping the "This is not used!" comment on the used type.

penev92:
You are removing the `throw` that is there as a safety measure to ensure this doesn't break silently, so please make sure the lint check for this does not need changing. Would be pretty bad if we break the lint check.

mods/ts/rules/defaults.yaml
GraionDilach: No, something is getting fixed there. Our die3 did not existed in TS and before this comment, our die4 had no sequence.

TS die3 was InfantryExplode=S_BANG34. (See your tibsun.mix/local.mix/rules.ini along with http://modenc.renegadeprojects.com/InfDeath). When @reaperrr managed to sneak in said unused (infdie.shp) asset into the game, he inserted it before the last 3 used deathanims (explosion, flameguy, electro)

infdie is a headshot animation. What he did there is to restore the used death type and rename the unused one. ExplosionDeath got shifted from 3 to 4 and the other got a new name. The to-do still stands.

abcdefg30:
Lint still works.

mods/ts/rules/defaults.yaml
penev92: OK, can we replace "TODO: check what should use this type and add it!" with "TODO: should be used by/for X and Y"?

GraionDilach: Umagon comes into my mind as a use-case.

abcdefg30: I think I should do that in a follow-up PR then.

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f3f5bdd6f80d3edf77bfdaa967e680cf969bc837...75546a0ff8d02e60ba1c4bd3a24acad33f5af217)

PR #8743, opened 2015-07-19
Remove Outtakes (the blooper song) from C&C.

GraionDilach:
I thought I'll get the hiding-music-from-UI done in time, but that didn't happened.

Outtakes shouldn't pop up in a playtest.

### DIFF
--- a/mods/cnc/audio/music.yaml
+++ b/mods/cnc/audio/music.yaml
@@ -58,6 +58,5 @@ target: Target (Mechanical Man)
 j1: Untamed Land
 valkyrie: Ride of the Valkyries
 voic226m: Voice Rhythm
-outtakes: Outtakes
 nod_map1: Nod Map Theme
 nod_win1: Nod Win Theme


dan9550:
This could be implemented as some kind of unlockable or since we have a campaign now, we could have a credits sequence at the end of that which it could play on.

mods/cnc/audio/music.yaml
```diff
@@ -58,6 +58,5 @@ target: Target (Mechanical Man)
 j1: Untamed Land
 valkyrie: Ride of the Valkyries
 voic226m: Voice Rhythm
-outtakes: Outtakes
```
Mailaender: Maybe out-comment it instead.

obrakmann:
It *would* be nice if this could play in the credits dialog, if we could ever make that work somehow. For the time being, :+1:

pchote:
My scores.mix (from the original C&C95 disk) doesn't even have this track.  :+1: in any case.

Issue #7035, opened 2014-11-30
Use upgrades to untangle implicit Building trait dependencies


pchote:
A number of traits include workarounds to disable their functionality until the building make animation has completed.  These all require querying the Building trait, and many of them do it without properly accounting for the trait construct order which causes them to break if the order that the traits are specified in the yaml is changed.

We can now use @atlimit8's new upgrade/disabled trait system to replace these hardcoded hacks:
1. Introduce a new trait that implements `INotifyBuildComplete` and grants an upgrade when the construction is complete / revokes the upgrade before selling.
2. Set `UpgradeMinEnabledLevel` for the traits that should only be enabled once the construction is complete. 
##

reaperrr:
Note to anyone who might take this on (including myself):
The way #9124 implements the deploy animation looks like a good example of how to do this.

reaperrr:
We should try to get this into Next+1, as this is a prerequisite to fix some long-standing issues.

pchote:
We now have support for granting a condition while the make animation plays (`WithMakeAnimation.Condition`), so we now just need to adapt the mod rules and fix any traits that work with `INotifyBuildComplete` but not conditions.

pchote:
There are now only 4 places left that do this - we should aim to finish cleaning this up alongside the building lock code

PR #8753, opened 2015-07-20
Replace RenderBuilding derivatives (RenderSimple removal Part 2)

reaperrr:
This replaces the four traits inheriting `RenderBuilding` with solutions based on `WithSpriteBody`.

NOTE: The `WithSpriteMakeAnimation` trait is a temporary solution that was necessary to avoid having to replace `RenderBuilding` and its four derivatives in one massive PR. It will be removed (or rather, replace the current `WithMakeAnimation`) together with `RenderBuilding`'s removal, hence why there's no upgrade rule for this.


reaperrr:
Updated.
@penev92 : Adressed everything you pointed out, pushed the changes to separate commits for easier reviewing. Tell me when I can squash/fixup them.

pchote:
Repeating my comment from above (where it's not very obvious), I agree with @penev92's concerns about pushing too much building-specific behaviour onto `WithSpriteBody`.  Don't fall into the trap of "everything is a (must use a) WithSpriteBody"!

You could solve this in a cleaner way by having a `WithBuildingSpriteBody` trait like you suggested above, but then going a step further by moving the make animation code into this trait directly and removing the `INotifyBuildComplete` interface.  That interface has always been a horrible hack that has been prone to initialization-ordering bugs, and its use case (disabling traits until the make animation is complete) can be done more robustly using the upgrades code.

reaperrr:
I'm a bit at a loss, not sure how to proceed from here.

If I remove `INotifyBuildComplete` from `WithSpriteBody`, firing offsets on turrets are broken (unless the actor has WithSpriteMakeAnimation, which implements it as well).

If I move the make anim stuff to a `WithSpriteBuildingBody`, I don't know how to make facings work for traits inheriting that (in this case `WithTurretedSpriteBody`).

And I don't fully understand how exactly #7035 would work, either. How would the approach described in #7035 know about waiting for the make anim to finish if we don't implement it 'the old way'?

penev92:
Short version - have a `WithMakeAnimation` that grants an upgrade when the animation is done.
Have all other traits depend on that upgrade.

reaperrr:
> Short version - have a WithMakeAnimation that grants an upgrade when the animation is done.
Have all other traits depend on that upgrade.

What about buildings without a make animation (d2k turrets, for example)? If that upgrade is supposed to replace INotifyBuildComplete, all buildings need it, regardless of whether they have a make animation.

pchote:
That isn't necessary if the upgrade is granted (to disable the traits) while the animation plays.

abcdefg30:
Needs a rebase.

OpenRA.Mods.Common/Traits/Render/WithSpriteBody.cs
```diff
@@ -69,6 +72,15 @@ public string NormalizeSequence(Actor self, string sequence)
 			return RenderSprites.NormalizeSequence(DefaultAnimation, self.GetDamageState(), sequence);
 		}
 
+		public virtual void BuildingComplete(Actor self)
```
penev92: Can you change this to `AddedToWorld()`, using the `INotifyAddedToWorld` interface instead of `INotifyBuildComplete`?

penev92:
OK, back to this. Sorry it has taken a while...
I think only the `INotifyAddedToWorld` is valid now that I see your last 3 commits.

reaperrr:
OK, will try to adress everything this weekend.

reaperrr:
Updated. After some cooked brain cells and discussion on IRC, I'd like to leave that "upgrades instead of INotifyBuildComplete" for later (after my render PRs are merged), as I'm struggling with the implementation anyway and don't want this PR to become any bigger.

reaperrr:
Rebased.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1782,6 +1782,136 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+							rb.Value.Nodes.RemoveAll(n => wsbNodes.Contains(n));
+						}
+
+						var rrb = node.Value.Nodes.FirstOrDefault(n => n.Key.StartsWith("-RenderBuildingCharge"));
+						if (rrb != null)
+							rrb.Key = "-WithChargeAnimation";
+					}
+				}
+
+				if (engineVersion < 20150719)
```
GraionDilach: IMO this should be redated - there was even a playtest since.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1782,6 +1782,136 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+							rb.Value.Nodes.RemoveAll(n => wsbNodes.Contains(n));
+						}
+
+						var rrb = node.Value.Nodes.FirstOrDefault(n => n.Key.StartsWith("-RenderBuildingSilo"));
+						if (rrb != null)
+							rrb.Key = "-WithSiloAnimation";
+					}
+				}
+
+				if (engineVersion < 20150720)
```
GraionDilach: And again.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1782,6 +1782,136 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+							rb.Value.Nodes.RemoveAll(n => rsNodes.Contains(n));
+						}
+
+						var rrb = node.Value.Nodes.FirstOrDefault(n => n.Key.StartsWith("-RenderBuildingTurreted"));
+						if (rrb != null)
+							rrb.Key = "-WithTurretedSpriteBody";
+					}
+				}
+
+				if (engineVersion < 20150721)
```
GraionDilach: This too.

reaperrr:
Updated.

reaperrr:
Rebased.

Mailaender:
This breaks RA silos by constantly looping through all fillness states and not having auto selection sizes: 
image

Mailaender:
This also breaks the tesla coil make animation.

Mailaender:
Popup turret also has no selection boundaries:

image

reaperrr:
Updated.
Fixed issues by reverting to `INotifyBuildComplete`, fixed selection sizes (note: these would have been fixed in #9189 at the latest, because there `AutoSelectionSize` is inherited from defaults).

Mailaender:
Tesla coils, obelisks and sam sites still don't have a make animation.

reaperrr:
Ok, it seems that splitting the removal of `RenderBuilding` in two PRs wasn't such a good idea afterall. I have trouble tracking down the make anim issues in this PR, but everything works fine in #9189, so I'm closing this in favor of the larger, but complete and more importantly now-functional #9189.

PR #8754, opened 2015-07-20
Modify upgrade rules to match #8658 for rules older than 2015-07-11

atlimit8:
#8658 made the extra `0` in `Modifier: 0, 0` unnecessary.

It may be more difficult to fix rules in between without the information lost by using more generic traits as of #8572.


abcdefg30:
:+1:

Issue #4717, opened 2014-02-23
Performance issues with AI building placement [$75]


reaperrr:
One of the biggest AI-related performance issues besides its problems with pathfinding is when it tries to place a building. In particular if it fails to find any room for it within the max base perimeter because the base has become too large or the building type cannot be built on the available ground type (example: ship yard when there is no water nearby).

In my case the AI wanted to build an advanced power plant, then couldn't find space for it, started its production again, failed to find space for it again, and so on, each time resulting in a massive tick_time lag spike.

There's a minimum of aspects to this that need to be fixed before I consider this issue solved:
- [ ] performance of the space check in general needs to be improved significantly, the lag spike when searching for space is simply too large - #8770 + #8759
- [x] if the AI had to abort placement of a building due to lack of space once, it should stop trying to build and place a structure again until it loses a structure (sold, destroyed, captured) - #8759

PR #8759, opened 2015-07-21
Improved AI BaseBuilder

### LINKED ISSUE #8116: Red Alert AI: don't build naval yards when they are impossible to deploy
Version: Release 20150424. Steps to reproduce:

1) Start a bot-only skirmish on the map "Mass Confliction" -- one normal, turtle and rush AI. Factions and spawn positions don't matter, though the spawn points should be as far apart as possible.
2) Open the production view, wait, and observe the AIs attempt to build naval yards on this land-only map.

This is different from #4672 -- the AI should not be attempting to build naval yards when it cannot deploy them, though proposed solution 2 would solve both issues.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13873290-red-alert-ai-don-t-build-naval-yards-when-they-are-impossible-to-deploy?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
This improves 3 things:

**1. The low power check.** 
Previously, it would check whether players' `ExcessPower` was less than `MinimumExcessPower` or whether the sum of all `Power` traits on the structure it wanted to build _exceeded_ the players' `ExcessPower`, and if either was true, build a power plant instead.
The catch: power _draw_ is represented by _negative_ numbers, so the only structures that would trigger the 2nd check were - power plants.
In other words, as long as current `ExcessPower` was equal or higher than `MinimumExcessPower`, the whole check would effectively fail, hence why the AI was out of power about half the time before I introduced `MinimumExcessPower` recently.
Now the 2nd check properly evaluates whether current `ExcessPower` + to-build actors' `Power` would be _lower_ than `MinimumExcessPower`, and in that case build a power plant instead.
Additionally, unlike before, it now also performs the "do I have enough power" check for the higher-priority overrides (refineries, production buildings, silos).
Last but not least, it now also gives powerplants priority if already on low power (no idea why it was different before).

**2. Repeated placement failures.**
Previously, the AI would constantly and stubbornly retry to build and place a structure even if out of space.
Now, after 3 consecutive failed attempts to place a structure, it is assumed that the AI has run out of build space and further production of structures is paused for a minute (default value of `StructureProductionResumeDelay`) and only resumed if the number of buildings decreased (or number of construction yards increased) in the meantime.

**3. Placement failure of naval structures.**
Previously, the AI would try to build naval structures even if there was no water within the buildable area. Now it first checks whether there is enough water in the base perimeter, and disables production of naval buildings if not. 
If there is actually such water space, when deciding whether to build a naval structure it checks if a building providing buildable area is close enough to that water. Only if that condition is met as well, the naval structure will be produced.

In summary, this should considerably reduce lag spikes from repeated building placement failures (particularly if it runs out of space) and finally prevent the AI reliably from driving itself into low power.

Fixes #8116, mostly adresses #4717.

### DIFF
--- a/OpenRA.Mods.Common/AI/BaseBuilder.cs
+++ b/OpenRA.Mods.Common/AI/BaseBuilder.cs
@@ -130,22 +130,27 @@ namespace OpenRA.Mods.Common.AI
 			return available.RandomOrDefault(ai.Random);
 		}
 
+		bool HasSufficientPowerForActor(ActorInfo actorInfo)
+		{
+			return (actorInfo.Traits.WithInterface<PowerInfo>().Where(i => i.UpgradeMinEnabledLevel < 1)
+				.Sum(p => p.Amount) + playerPower.ExcessPower) >= ai.Info.MinimumExcessPower;
+		}
+
 		ActorInfo ChooseBuildingToBuild(ProductionQueue queue)
 		{
 			var buildableThings = queue.BuildableItems();
 
+			// This gets used quite a bit, so let's cache it here
+			var power = GetProducibleBuilding("Power", buildableThings,
+				a => a.Traits.WithInterface<PowerInfo>().Where(i => i.UpgradeMinEnabledLevel < 1).Sum(p => p.Amount));
+
 			// First priority is to get out of a low power situation
 			if (playerPower.ExcessPower < ai.Info.MinimumExcessPower)
 			{
-				var power = GetProducibleBuilding("Power", buildableThings, a => a.Traits.WithInterface<PowerInfo>().Where(i => i.UpgradeMinEnabledLevel < 1).Sum(p => p.Amount));
 				if (power != null && power.Traits.WithInterface<PowerInfo>().Where(i => i.UpgradeMinEnabledLevel < 1).Sum(p => p.Amount) > 0)
 				{
-					// TODO: Handle the case when of when we actually do need a power plant because we don't have enough but are also suffering from a power outage
-					if (playerPower.PowerOutageRemainingTicks <= 0)
-					{
-						HackyAI.BotDebug("AI: {0} decided to build {1}: Priority override (low power)", queue.Actor.Owner, power.Name);
-						return power;
-					}
+					HackyAI.BotDebug("AI: {0} decided to build {1}: Priority override (low power)", queue.Actor.Owner, power.Name);
+					return power;
 				}
 			}
 
@@ -153,10 +158,16 @@ namespace OpenRA.Mods.Common.AI
 			if (!ai.HasAdequateProc() || !ai.HasMinimumProc())
 			{
 				var refinery = GetProducibleBuilding("Refinery", buildableThings);
-				if (refinery != null)
+				if (refinery != null && HasSufficientPowerForActor(refinery))
 				{
 					HackyAI.BotDebug("AI: {0} decided to build {1}: Priority override (refinery)", queue.Actor.Owner, refinery.Name);
 					return refinery;
+				}
+
+				if (power != null && refinery != null && !HasSufficientPowerForActor(refinery))
+				{
+					HackyAI.BotDebug("{0} decided to build {1}: Priority override (would be low power)", queue.Actor.Owner, power.Name);
+					return power;
 				}
 			}
 
@@ -164,10 +175,16 @@ namespace OpenRA.Mods.Common.AI
 			if (ai.Info.NewProductionCashThreshold > 0 && playerResources.Resources > ai.Info.NewProductionCashThreshold)
 			{
 				var production = GetProducibleBuilding("Production", buildableThings);
-				if (production != null)
+				if (production != null && HasSufficientPowerForActor(production))
 				{
 					HackyAI.BotDebug("AI: {0} decided to build {1}: Priority override (production)", queue.Actor.Owner, production.Name);
 					return production;
+				}
+
+				if (power != null && production != null && !HasSufficientPowerForActor(production))
+				{
+					HackyAI.BotDebug("{0} decided to build {1}: Priority override (would be low power)", queue.Actor.Owner, power.Name);
+					return power;
 				}
 			}
 
@@ -175,10 +192,16 @@ namespace OpenRA.Mods.Common.AI
 			if (playerResources.AlertSilo)
 			{
 				var silo = GetProducibleBuilding("Silo", buildableThings);
-				if (silo != null)
+				if (silo != null && HasSufficientPowerForActor(silo))
 				{
 					HackyAI.BotDebug("AI: {0} decided to build {1}: Priority override (silo)", queue.Actor.Owner, silo.Name);
 					return silo;
+				}
+
+				if (power != null && silo != null && !HasSufficientPowerForActor(silo))
+				{
+					HackyAI.BotDebug("{0} decided to build {1}: Priority override (would be low power)", queue.Actor.Owner, power.Name);
+					return power;
 				}
 			}
 
@@ -200,23 +223,18 @@ namespace OpenRA.Mods.Common.AI
 					continue;
 
 				// Will this put us into low power?
-				var actor = world.Map.Rules.Actors[frac.Key];
-				var pis = actor.Traits.WithInterface<PowerInfo>().Where(i => i.UpgradeMinEnabledLevel < 1);
-				if (playerPower.ExcessPower < ai.Info.MinimumExcessPower || playerPower.ExcessPower < pis.Sum(pi => pi.Amount))
+				var actor = world.Map.Rules.Actors[name];
+				if (playerPower.ExcessPower < ai.Info.MinimumExcessPower || !HasSufficientPowerForActor(actor))
 				{
 					// Try building a power plant instead
-					var power = GetProducibleBuilding("Power",
-						buildableThings, a => a.Traits.WithInterface<PowerInfo>().Where(i => i.UpgradeMinEnabledLevel < 1).Sum(pi => pi.Amount));
 					if (power != null && power.Traits.WithInterface<PowerInfo>().Where(i => i.UpgradeMinEnabledLevel < 1).Sum(pi => pi.Amount) > 0)
 					{
-						// TODO: Handle the case when of when we actually do need a power plant because we don't have enough but are also suffering from a power outage
 						if (playerPower.PowerOutageRemainingTicks > 0)
-							HackyAI.BotDebug("AI: {0} is suffering from a power outage; not going to build {1}", queue.Actor.Owner, power.Name);
+							HackyAI.BotDebug("{0} decided to build {1}: Priority override (is low power)", queue.Actor.Owner, power.Name);
 						else
-						{
 							HackyAI.BotDebug("{0} decided to build {1}: Priority override (would be low power)", queue.Actor.Owner, power.Name);
-							return power;
-						}
+
+						return power;
 					}
 				}
 
--- a/OpenRA.Mods.Common/AI/BaseBuilder.cs
+++ b/OpenRA.Mods.Common/AI/BaseBuilder.cs
@@ -28,6 +28,8 @@ namespace OpenRA.Mods.Common.AI
 
 		int waitTicks;
 		Actor[] playerBuildings;
+		int failCount;
+		int failRetryTicks;
 
 		public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, PlayerResources pr)
 		{
@@ -37,10 +39,15 @@ namespace OpenRA.Mods.Common.AI
 			playerPower = pm;
 			playerResources = pr;
 			this.category = category;
+			failRetryTicks = ai.Info.StructureProductionResumeDelay;
 		}
 
 		public void Tick()
 		{
+			// If failed to place something N consecutive times, wait M ticks until resuming building production
+			if (failCount >= ai.Info.MaximumFailedPlacementAttempts && --failRetryTicks <= 0)
+				failCount = 0;
+
 			// Only update once per second or so
 			if (--waitTicks > 0)
 				return;
@@ -63,7 +70,7 @@ namespace OpenRA.Mods.Common.AI
 			var currentBuilding = queue.CurrentItem();
 
 			// Waiting to build something
-			if (currentBuilding == null)
+			if (currentBuilding == null && failCount < ai.Info.MaximumFailedPlacementAttempts)
 			{
 				var item = ChooseBuildingToBuild(queue);
 				if (item == null)
@@ -77,6 +84,7 @@ namespace OpenRA.Mods.Common.AI
 				// Production is complete
 				// Choose the placement logic
 				// HACK: HACK HACK HACK
+				// TODO: Derive this from BuildingCommonNames instead
 				var type = BuildingType.Building;
 				if (world.Map.Rules.Actors[currentBuilding.Item].Traits.Contains<AttackBaseInfo>())
 					type = BuildingType.Defense;
@@ -88,9 +96,11 @@ namespace OpenRA.Mods.Common.AI
 				{
 					HackyAI.BotDebug("AI: {0} has nowhere to place {1}".F(player, currentBuilding.Item));
 					ai.QueueOrder(Order.CancelProduction(queue.Actor, currentBuilding.Item, 1));
+					failCount += failCount;
 				}
 				else
 				{
+					failCount = 0;
 					ai.QueueOrder(new Order("PlaceBuilding", player.PlayerActor, false)
 					{
 						TargetLocation = location.Value,
--- a/OpenRA.Mods.Common/AI/BaseBuilder.cs
+++ b/OpenRA.Mods.Common/AI/BaseBuilder.cs
@@ -33,6 +33,7 @@ namespace OpenRA.Mods.Common.AI
 		int failCount;
 		int failRetryTicks;
 		int cachedBases;
+		int cachedBuildings;
 
 		public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, PlayerResources pr)
 		{
@@ -49,7 +50,23 @@ namespace OpenRA.Mods.Common.AI
 		{
 			// If failed to place something N consecutive times, wait M ticks until resuming building production
 			if (failCount >= ai.Info.MaximumFailedPlacementAttempts && --failRetryTicks <= 0)
-				failCount = 0;
+			{
+				var currentBuildings = world.ActorsWithTrait<Building>()
+					.Where(a => a.Actor.Owner == player)
+					.Count();
+
+				var baseProviders = world.ActorsWithTrait<BaseBuilding>()
+					.Where(a => a.Actor.Owner == player)
+					.Count();
+
+				// Only bother resetting failCount if either a) the number of buildings has decreased since last failure M ticks ago,
+				// or b) number of BaseProviders (construction yard or similar) has increased since then.
+				// Otherwise reset failRetryTicks instead to wait again.
+				if (currentBuildings < cachedBuildings || baseProviders > cachedBases)
+					failCount = 0;
+				else
+					failRetryTicks = ai.Info.StructureProductionResumeDelay;
+			}
 
 			// Only update once per second or so
 			if (--waitTicks > 0)
@@ -119,6 +136,18 @@ namespace OpenRA.Mods.Common.AI
 					HackyAI.BotDebug("AI: {0} has nowhere to place {1}".F(player, currentBuilding.Item));
 					ai.QueueOrder(Order.CancelProduction(queue.Actor, currentBuilding.Item, 1));
 					failCount += failCount;
+
+					// If we just reached the maximum fail count, cache the number of current structures
+					if (failCount == ai.Info.MaximumFailedPlacementAttempts)
+					{
+						cachedBuildings = world.ActorsWithTrait<Building>()
+							.Where(a => a.Actor.Owner == player)
+							.Count();
+
+						cachedBases = world.ActorsWithTrait<BaseBuilding>()
+							.Where(a => a.Actor.Owner == player)
+							.Count();
+					}
 				}
 				else
 				{
--- a/OpenRA.Mods.Common/AI/BaseBuilder.cs
+++ b/OpenRA.Mods.Common/AI/BaseBuilder.cs
@@ -101,7 +101,11 @@ namespace OpenRA.Mods.Common.AI
 				if (TickQueue(queue))
 					active = true;
 
-			waitTicks = active ? ai.Info.StructureProductionActiveDelay : ai.Info.StructureProductionInactiveDelay;
+			// Add a random factor so not every AI produces at the same tick early in the game.
+			// Minimum should not be negative as delays in HackyAI could be zero.
+			var randomFactor = world.SharedRandom.Next(0, ai.Info.StructureProductionRandomBonusDelay);
+			waitTicks = active ? ai.Info.StructureProductionActiveDelay + randomFactor
+				: ai.Info.StructureProductionInactiveDelay + randomFactor;
 		}
 
 		bool TickQueue(ProductionQueue queue)
--- a/OpenRA.Mods.Common/AI/BaseBuilder.cs
+++ b/OpenRA.Mods.Common/AI/BaseBuilder.cs
@@ -28,12 +28,20 @@ namespace OpenRA.Mods.Common.AI
 
 		int waitTicks;
 		Actor[] playerBuildings;
-		bool waterAvailable;
-		bool checkedWater;
 		int failCount;
 		int failRetryTicks;
+		int checkForBasesTicks;
 		int cachedBases;
 		int cachedBuildings;
+
+		enum Water
+		{
+			NotChecked,
+			EnoughWater,
+			NotEnoughWater
+		}
+
+		Water waterState = Water.NotChecked;
 
 		public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, PlayerResources pr)
 		{
@@ -55,7 +63,7 @@ namespace OpenRA.Mods.Common.AI
 					.Where(a => a.Actor.Owner == player)
 					.Count();
 
-				var baseProviders = world.ActorsWithTrait<BaseBuilding>()
+				var baseProviders = world.ActorsWithTrait<BaseProvider>()
 					.Where(a => a.Actor.Owner == player)
 					.Count();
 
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -153,33 +167,92 @@ ActorInfo ChooseBuildingToBuild(ProductionQueue queue)
...
+				if (refinery != null
+					&& (refinery.Traits.WithInterface<PowerInfo>().Where(i => i.UpgradeMinEnabledLevel < 1).Sum(p => p.Amount) + playerPower.ExcessPower)
+					>= ai.Info.MinimumExcessPower)
 				{
 					HackyAI.BotDebug("AI: {0} decided to build {1}: Priority override (refinery)", queue.Actor.Owner, refinery.Name);
 					return refinery;
 				}
+
+				if (power != null && refinery != null
+					&& (refinery.Traits.WithInterface<PowerInfo>().Where(i => i.UpgradeMinEnabledLevel < 1).Sum(p => p.Amount) + playerPower.ExcessPower)
```
pchote: This check is done a bunch.  Could you pull it out into a helper function?

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,10 +40,19 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
 			playerResources = pr;
 			this.category = category;
+			waterAvailable = true;
+			failCount = 0;
+			failRetryTicks = ai.Info.StructureProductionResumeDelay;
 		}
 
 		public void Tick()
 		{
+			// If failed to place something 3 consecutive times, wait 1 minute until retrying to build something
```
pchote: Not necessarily 1 minute (depends on yaml).

pchote: Also, 3 is a rather magic number.  Can we pull that out into a parameter too?

reaperrr:
Updated.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,10 +40,19 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
 			playerResources = pr;
 			this.category = category;
+			waterAvailable = true;
+			failCount = 0;
+			failRetryTicks = ai.Info.StructureProductionResumeDelay;
 		}
 
 		public void Tick()
 		{
+			// If failed to place something N consecutive times, wait N ticks until resuming building production
```
pchote: Nit: that second N should be an M (it is a different arbitrary number to the first one).

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,10 +40,19 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
+			if (failCount >= ai.Info.MaximumFailedPlacementAttempts && --failRetryTicks > 0)
+				--failRetryTicks;
```
pchote: Do you really intend on decrementing `failRetryTicks` twice per tick?  You probably want to drop the `--` inside the if check.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,10 +40,19 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
+			else if (failCount >= ai.Info.MaximumFailedPlacementAttempts && --failRetryTicks <= 0)
```
pchote: and here too.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -130,22 +145,31 @@ ActorInfo GetProducibleBuilding(string commonName, IEnumerable<ActorInfo> builda
 			return available.RandomOrDefault(ai.Random);
 		}
 
+		bool SufficientPowerForActor(ActorInfo actorInfo)
```
pchote: `HasSufficientPowerForActor`?

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -153,33 +177,76 @@ ActorInfo ChooseBuildingToBuild(ProductionQueue queue)
...
+				{
+					var navalBuilding = navalproduction.ToString();
+					var buildLoc = ai.ChooseBuildLocation(navalBuilding, true, BuildingType.Building);
+					if (buildLoc != null)
+					{
+						HackyAI.BotDebug("AI: {0} decided to build {1}: Priority override (navalproduction)", queue.Actor.Owner, navalproduction.Name);
+						return navalproduction;
+					}
+					else
+						waterAvailable = false;
```
pchote: If you haven't expanded to the full range of your base radius then it may be possible to place a naval yard later even if you can't place one right now.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -199,24 +266,34 @@ ActorInfo ChooseBuildingToBuild(ProductionQueue queue)
 				if (ai.Info.BuildingLimits.ContainsKey(name) && ai.Info.BuildingLimits[name] <= count)
 					continue;
 
+				if (!waterAvailable && ai.Info.BuildingCommonNames.ContainsKey("NavalProduction")
+					&& ai.Info.BuildingCommonNames["NavalProduction"].Contains(name))
+					continue;
+
+				if (waterAvailable && ai.Info.BuildingCommonNames.ContainsKey("NavalProduction")
```
pchote: This is mostly duplicating the check from above - how about pulling it out into a temporary variable above?

OpenRA.Mods.Common/AI/BaseBuilder.cs
pchote: A more accurate way to test this would be to count the number of water cells within the current available base ranges.  If it's < 5 (or some other arbitrary number that should be large enough to justify building a naval yard) then turn it off.  Turn it back on and recalculate if new bases become available.

obrakmann:
Haven't looked at the code yet, only played two test games. In both games, I noticed that several AIs would either not produce any buildings at all, or stop relatively early in the tech tree (they only had power, a refinery and a silo).

The replays are [here](https://seacloud.cc/d/5d1adbebce/files/?p=/OpenRA-2015-07-22T130350Z.orarep) and [here](https://seacloud.cc/d/5d1adbebce/files/?p=/OpenRA-2015-07-22T140921Z.orarep). 

But the low power issue seems to be gone indeed. Did not see that at all in the replays.

reaperrr:
OK, I suspect the last chunk with the `canPlaceNavalProduction` check is bollocks and blocks production, I'll test that first.

clemty:
>  if no space is found for the naval structure, disables construction of naval structures permanently for the rest of that game.
wouldn't that be overkill in case the AI would somehow acquire another construction yard near water? (that might not be the case now, but what if AI gets good enough to spawn a second base sometime?)

reaperrr:
Updated.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,14 +42,44 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
+			cachedBases = 0;
+			waterAvailable = false;
+			checkedWater = false;
+			failCount = 0;
+			failRetryTicks = ai.Info.StructureProductionResumeDelay;
 		}
 
 		public void Tick()
 		{
+			// If failed to place something N consecutive times, wait M ticks until resuming building production
```
pchote: This can be simplified to
```csharp
// If failed to place something N consecutive times, wait M ticks until resuming building production
if (failCount >= ai.Info.MaximumFailedPlacementAttempts && --failRetryTicks <= 0)
    failCount = 0;

// Only update once per second or so
```

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,14 +42,44 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
 			if (--waitTicks > 0)
 				return;
 
+			if (!checkedWater)
+			{
+				waterAvailable = ai.EnoughWaterToBuildNaval();
+				checkedWater = true;
+			}
+
+			if (checkedWater && !waterAvailable)
```
pchote: You set checkedWater, then check if you should unset it.  This logic is twistier than it needs to be.

pchote: Since you're doing this check in Tick anyway, you might as well have it check against the currently available build area, and reset it each time a building is placed or lost.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -84,13 +120,26 @@ bool TickQueue(ProductionQueue queue)
 					type = BuildingType.Refinery;
 
 				var location = ai.ChooseBuildLocation(currentBuilding.Item, true, type);
-				if (location == null)
+				var navalBuilding = ai.Info.BuildingCommonNames.ContainsKey("NavalProduction")
+					&& ai.Info.BuildingCommonNames["NavalProduction"].Contains(currentBuilding.Item);
+
+				// Don't count naval buildings towards regular failCount
```
pchote: You could then remove this special case hack.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -153,33 +211,69 @@ ActorInfo ChooseBuildingToBuild(ProductionQueue queue)
...
 				}
+
+				if (power != null && production != null && !HasSufficientPowerForActor(production))
+				{
+					HackyAI.BotDebug("{0} decided to build {1}: Priority override (would be low power)", queue.Actor.Owner, power.Name);
+					return power;
+				}
+			}
+
+			// Make sure that we can can spend as fast as we are earning
```
pchote: The comment doesn't match what this is doing.  You can increase spending by building any new production structure - there's nothing special about naval.

penev92: Also "can can".

reaperrr:
Updated.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -231,6 +238,30 @@ public void Activate(Player p)
...
+					&& !a.HasTrait<Mobile>());
+
+			foreach (var b in baseBuildings)
+			{
+				// TODO: Unhardcode terrain type
+				// TODO2: Only count cells that have all 8 adjacent cells free and of matching terrain type, then anything > 0 would be fine
+				var countWaterTiles = Map.FindTilesInCircle(b.Location, Info.MaxBaseRadius)
+					.Where(c => Player.World.Map.Contains(c)
+						&& Player.World.Map.GetTerrainInfo(c).Type == "Water")
+					.Count();
```
obrakmann: Shouldn't this check if the water is in buildable range, too (`BuildingInfo.Adjacent`)? Otherwise the AI would still try to build naval structures even before it could actually place them just because there is water somewhere within the base radius. Not sure, really. Maybe this is good enough.

obrakmann:
Works nicely now, otherwise.

reaperrr:
Updated, refactored the naval check once more.

Rather than just counting an arbitrary number of water cells, it now checks whether there is a least one water cell where all 8 adjacent cells are water cells, too.

Additionally, if the above is the case, when deciding what structure to build next it then checks whether such a cell is within a radius of N (defaults to 8 since that's the Adjacent value of RA naval buildings) around any current building providing buildable area, and only starts production of a naval building if this is the case as well.

This should be far more robust now, but needs some testing, of course.

reaperrr:
There's one more small thing I'd like to add, so don't merge this yet even if everything works fine.

reaperrr:
Updated.
The `failCount` is now only reset after a certain amount of time if the number of buildings changed, otherwise `failRetryTicks` is reset instead.

This prevents the AI from performing 3 failing production attempts every minute or so in the situation where the failures were indeed caused by lack of space, and no buildings have been sold or destroyed in the meantime (and no construction yard added, either).

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,14 +43,59 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
 			playerPower = pm;
 			playerResources = pr;
 			this.category = category;
+			cachedBases = 0;
```
pchote: 0 / false are the default values for the data types anyway, so these are redundant.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -130,22 +195,31 @@ ActorInfo GetProducibleBuilding(string commonName, IEnumerable<ActorInfo> builda
 			return available.RandomOrDefault(ai.Random);
 		}
 
+		bool HasSufficientPowerForActor(ActorInfo actorInfo)
+		{
+			if ((actorInfo.Traits.WithInterface<PowerInfo>()
+				.Where(i => i.UpgradeMinEnabledLevel < 1).Sum(p => p.Amount) + playerPower.ExcessPower)
+				>= ai.Info.MinimumExcessPower)
+				return true;
```
Mailaender: You can simplify this to 
```csharp
return actorInfo.Traits.WithInterface<PowerInfo>().Where(i => i.UpgradeMinEnabledLevel < 1)
	.Sum(p => p.Amount) + playerPower.ExcessPower) >= ai.Info.MinimumExcessPower)
```

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,14 +43,54 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
 		public void Tick()
 		{
+			// If failed to place something N consecutive times, wait M ticks until resuming building production
+			if (failCount >= ai.Info.MaximumFailedPlacementAttempts && --failRetryTicks <= 0)
+			{
+				var currentBuildings = world.ActorsWithTrait<Building>()
+					.Where(a => a.Actor.Owner == player)
+					.Count();
+
+				var baseProviders = world.ActorsWithTrait<BaseBuilding>()
```
pchote: `BaseBuilding` (a tag trait used for the 'go to base' hotkey) != `BaseProvider` (the trait that grants base area)!

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,14 +43,54 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
 
+			if (!checkedWater)
+			{
+				waterAvailable = ai.EnoughWaterToBuildNaval();
+				checkedWater = true;
+			}
+
+			if (!waterAvailable)
+			{
+				var currentBases = world.ActorsWithTrait<BaseBuilding>()
```
pchote: Similarly here.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,14 +43,54 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
+					failRetryTicks = ai.Info.StructureProductionResumeDelay;
+			}
+
 			// Only update once per second or so
 			if (--waitTicks > 0)
 				return;
 
+			if (!checkedWater)
+			{
+				waterAvailable = ai.EnoughWaterToBuildNaval();
```
pchote: Managing the state of two flags feels kind of gross.  Can we use an enum with three values instead?

OpenRA.Mods.Common/AI/BaseBuilder.cs
pchote: It looks like this query is run every tick, which is a bit overkill.  Can we rate limit it?

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -231,6 +246,58 @@ public void Activate(Player p)
 				resourceTypeIndices.Set(World.TileSet.GetTerrainIndex(t.TerrainType), true);
 		}
 
+		// TODO: Possibly give this a more generic name when terrain type is unhardcoded
+		public bool EnoughWaterToBuildNaval()
+		{
+			var baseBuildings = World.Actors.Where(
+				a => a.Owner == Player
+					&& a.HasTrait<BaseBuilding>()
```
pchote: `BaseProvider`.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,14 +51,57 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
+			if (waterState == Water.NotChecked)
+			{
+				if (ai.EnoughWaterToBuildNaval())
+					waterState = Water.EnoughWater;
+				else
+					waterState = Water.NotEnoughWater;
+					checkForBasesTicks = ai.Info.CheckForNewBasesDelay;
+			}
+
+			if (waterState == Water.NotEnoughWater && --checkForBasesTicks <= 0)
```
obrakmann: Due to the `if (--waitTicks > 0) return;` above, this counter will not decrease every tick, delaying re-checking considerably.

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -37,14 +51,57 @@ public BaseBuilder(HackyAI ai, string category, Player p, PowerManager pm, Playe
...
 			if (--waitTicks > 0)
 				return;
 
+			if (waterState == Water.NotChecked)
+			{
+				if (ai.EnoughWaterToBuildNaval())
+					waterState = Water.EnoughWater;
+				else
+					waterState = Water.NotEnoughWater;
+					checkForBasesTicks = ai.Info.CheckForNewBasesDelay;
```
obrakmann: missing braces on the `else` block

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -153,33 +234,69 @@ ActorInfo ChooseBuildingToBuild(ProductionQueue queue)
...
+
+				if (power != null && production != null && !HasSufficientPowerForActor(production))
+				{
+					HackyAI.BotDebug("{0} decided to build {1}: Priority override (would be low power)", queue.Actor.Owner, power.Name);
+					return power;
+				}
+			}
+
+			// Only consider building this if there is enough water inside the base perimeter and there are close enough adjacent buildings
+			if (waterState == Water.EnoughWater && ai.CloseEnoughToWater()
```
obrakmann: Can you move the `ai.CloseEnoughToWater()` call to the end of the conditions? It's the most costly.

obrakmann:
Just a remark: the AI would probably be much easier to program if it implemented all the INotify* interfaces, and all the events would then actually notify not only the actor, but its owner, too. Then we could get rid of this whole freaking state machine.

Phrohdoh:
re: @obrakmann's above comment; see [this](https://github.com/Phrohdoh/ModularAI).

obrakmann:
I think this should be ready once the inline comments are addressed. Looks good overall.

reaperrr:
Updated.

Decided to keep the checkForBasesTicks, and kept the commit separate since I didn't feel like wading through rebase conflicts due to the commits in between.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -52,12 +52,25 @@ public sealed class HackyAIInfo : IBotInfo, ITraitInfo
...
 		public readonly int StructureProductionInactiveDelay = 125;
 
-		[Desc("How long to wait (in ticks) between structure production checks ticks when actively building things.")]
+		[Desc("Minimum delay (in ticks) between structure production checks when actively building things.")]
 		public readonly int StructureProductionActiveDelay = 10;
 
+		[Desc("A random delay (in ticks) of up to this is added to production delays.")]
+		public readonly int StructureProductionRandomBonusDelay = 10;
+
+		[Desc("How long to wait (in ticks) until retrying to build structure after the last 3 consecutive attempts failed.")]
```
GraionDilach: You replaced "How long to wait" with "Minimum delay" above... consistency is for the weak? :D

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -52,12 +52,25 @@ public sealed class HackyAIInfo : IBotInfo, ITraitInfo
...
+		public readonly int StructureProductionResumeDelay = 1500;
+
+		[Desc("After how many failed attempts to place a structure should AI give up and wait",
+		"for StructureProductionResumeDelay before retrying.")]
+		public readonly int MaximumFailedPlacementAttempts = 3;
+
+		[Desc("How long to wait (in ticks) until rechecking for new BaseProviders.")]
```
GraionDilach: Again.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -82,6 +95,11 @@ public sealed class HackyAIInfo : IBotInfo, ITraitInfo
 		[Desc("Radius in cells around the center of the base to expand.")]
 		public readonly int MaxBaseRadius = 20;
 
+		[Desc("Radius in cells around each building with ProvideBuildableArea",
+		"to check for a 3x3 area of water where naval structures can be built.",
```
GraionDilach: I'm not certain if I like the sound of this. RA2 naval yards are definitely bigger than 3x3 and IIRC that GDI navalyard c'n'p for TS is also 4x3. Can't you customize this that instead of a fixed size, it checks for the biggest foundation of the buildings listed in NavalProduction?

penev92: While I agree this is bad (too RA-specific), IMO we should limit the scope creeping already.
Also as far as I see the code that uses this radius doesn't really care about the footprint, so I'd say just remove the size from the description here.

OpenRA.Mods.Common/AI/HackyAI.cs
reaperrr: StructureProductionRandomBonusDelay is added to the production delays, but not this, hence why minimum for the above and not for these. So yes this is intentional.

OpenRA.Mods.Common/AI/HackyAI.cs
reaperrr: Factoring in bulding footprints (or generally more than 3x3 cells) would need another refactor, and I don't feel like doing that in this pr. This is already far better than what we had before.

OpenRA.Mods.Common/AI/HackyAI.cs
GraionDilach: This is `StructureProductionResumeDelay`'s description.

reaperrr: Yes, and this does not add a random bonus delay, so it represents the exact delay value. The two Active and Inactive delays above add a random bonus delay, though, so they only represent the minimum delay. So the descriptions are correct, though it seems they can be misunderstood, so I'll try to word them better.

OpenRA.Mods.Common/AI/HackyAI.cs
GraionDilach: @penev92: It explicitly does not check the footprint indeed, however it counts how many water tiles surrounded by fully buildable water tiles exist, which ultimately does end up with a 3x3 buildable water area.

@reaperrr: Please add TODOs into `EnoughWaterToBuildNaval()` and `CloseEnoughToWater()` then at least.

reaperrr:
Rebased & updated.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -49,17 +49,32 @@ public sealed class HackyAIInfo : IBotInfo, ITraitInfo
...
 
+		[Desc("Minimum portion of pending orders to issue each tick (e.g. 5 issues at least 1/5th of all pending orders). Excess orders remain queued for subsequent ticks.")]
+		public readonly int MinOrderQuotientPerTick = 5;
+
 		[Desc("Minimum excess power the AI should try to maintain.")]
 		public readonly int MinimumExcessPower = 0;
 
-		[Desc("How long to wait (in ticks) between structure production checks when there is no active production.")]
+		[Desc("Delay (in ticks) between structure production checks when there is no active production.",
+		"A StructureProductionRandomBonusDelay is added to this.")]
```
obrakmann: Please remember to add spaces in front of a string continuation. Affects almost all `Desc`s here.

obrakmann:
Aside from that style nit, this is looking good to me now. :+1:

OpenRA.Mods.Common/AI/HackyAI.cs
reaperrr: Well, I misunderstood your comment anyway and tabbed lines 2+ on the descriptions instead ;)

reaperrr:
Updated.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -250,6 +270,60 @@ public void Activate(Player p)
...
+					&& !a.HasTrait<Mobile>());
+
+			foreach (var a in areaProviders)
+			{
+				// TODO: Unhardcode terrain type
+				// TODO2: Properly check building foundation rather than 3x3 area
+				var playerWorld = Player.World;
+				var adjacentWater = Map.FindTilesInCircle(a.Location, Info.CheckForWaterRadius)
+					.Where(c => playerWorld.Map.Contains(c)
+						&& playerWorld.Map.GetTerrainInfo(c).Type == "Water"
```
Mailaender: You can use `.IsWater` instead which at least is cheaper than a string comparison.

reaperrr:
Updated.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -250,6 +270,60 @@ public void Activate(Player p)
...
+		public bool CloseEnoughToWater()
+		{
+			var areaProviders = World.Actors.Where(
+				a => a.Owner == Player
+					&& a.HasTrait<GivesBuildableArea>()
+					&& !a.HasTrait<Mobile>());
+
+			foreach (var a in areaProviders)
+			{
+				// TODO: Unhardcode terrain type
```
Mailaender: Is this resolved now?

reaperrr: No, the idea is that in the future this whole check could be generalized to check for terrain type(s) of the modders choice, so the TODO still stands.

Mailaender:
Tested a bot match on a map with small ponds and limited space to build which the AI seems to solve a bit better now. :white_check_mark: Only noticed one blackout due to not building a power plant first.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/0cfe138c44851a619b8166a2a734cb223c4057dd...6c0ddced2783473302d62733e9503e3491670eec)

OpenRA.Mods.Common/AI/BaseBuilder.cs
```diff
@@ -55,7 +114,11 @@ public void Tick()
 				if (TickQueue(queue))
 					active = true;
 
-			waitTicks = active ? ai.Info.StructureProductionActiveDelay : ai.Info.StructureProductionInactiveDelay;
+			// Add a random factor so not every AI produces at the same tick early in the game.
+			// Minimum should not be negative as delays in HackyAI could be zero.
+			var randomFactor = world.SharedRandom.Next(0, ai.Info.StructureProductionRandomBonusDelay);
```
Mailaender: Don't use shared random here in future patches. See https://github.com/OpenRA/OpenRA/pull/8870.

PR #8761, opened 2015-07-21
Added RenderUpgradeRangeCircle trait & applied it to Cloak Generator

atlimit8:


OpenRA.Mods.Common/Traits/Render/RenderUpgradeRangeCircle.cs
```diff
@@ -0,0 +1,84 @@
...
+	class RenderUpgradeRangeCircleInfo : ITraitInfo, IPlaceBuildingDecoration, Requires<UpgradeActorsNearInfo>
+	{
+		public readonly string RangeCircleType = null;
+
+		[Desc("Color to use")]
+		public readonly Color Color = Color.BlueViolet;
+
+		public IEnumerable<IRenderable> Render(WorldRenderer wr, World w, ActorInfo ai, WPos centerPosition)
+		{
+			var infos = ai.Traits.WithInterface<UpgradeActorsNearInfo>().Where(t => t.RangeCircleType == RangeCircleType);
```
Mailaender: This is duplicated from the constructor. Can you cache the trait lookup for performance reasons?

OpenRA.Mods.Common/Traits/Render/RenderUpgradeRangeCircle.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Draw a circle indicating my upgrade of nearby range.")]
```
penev92: This doesn't sound quite right around the end.

OpenRA.Mods.Common/Traits/Render/RenderUpgradeRangeCircle.cs
```diff
@@ -0,0 +1,94 @@
...
+	class RenderUpgradeRangeCircleInfo : ITraitInfo, IPlaceBuildingDecoration, Requires<UpgradeActorsNearInfo>
+	{
+		public readonly string RangeCircleType = null;
```
penev92: Simple description please.

OpenRA.Mods.Common/Traits/Render/RenderUpgradeRangeCircle.cs
```diff
@@ -0,0 +1,94 @@
...
+
+		[Desc("Color to use")]
+		public readonly Color Color = Color.BlueViolet;
+
+		IEnumerable<UpgradeActorsNearInfo> infos = null;
```
penev92: Looks like this can be a local variable inside `GetInfos()`.

OpenRA.Mods.Common/Traits/Render/RenderUpgradeRangeCircle.cs
```diff
@@ -0,0 +1,94 @@
...
+
+		public IEnumerable<UpgradeActorsNearInfo> GetInfos(ActorInfo ai)
+		{
+			if (infos == null)
+				infos = ai.Traits.WithInterface<UpgradeActorsNearInfo>()
```
penev92: Can this check for `RenderUpgradeRangeCircleInfo` instead? I'd really like to see `UpgradeActorsNearInfo` not having a `RangeCircleType`.

penev92:
I like the overall idea, but I'd rather `UpgradesActorsNear` know about `RenderUpgradeCircle`, otherwise that mostly defeats the purpose of having them as separate traits.
I'd rather:
 - Either have `RenderUpgradeCircle` define a `Range` and let modders set it to the same range as that on `UpgradesActorsNear` (very, very simple, but somewhat prone to human errors).
 - Or do a 1:1 mapping between the traits with trait instance names like `Armament` has. Then `RenderUpgradeCircle` can have a reference to the corresponding `UpgradeActorsNear` and the latter can have a public `Range` property that the first can get. (IMO this is the best option because then the `Range` of UAN can easily upgradable, and RUC will still render at the current range without needing manual changing).
I'd be OK with either of the above, or another option if someone has ideas.

atlimit8:
Low priority right now and requires #9317.

PR #8762, opened 2015-07-21
Release file handles when unmounting file system

RoosterDragon:
Make `IFolder` implement `IDisposable`, and fix all concrete classes to dispose any streams they open.

When a folder is unmounted from `GlobalFileSystem`, dispose the folder to ensure all the streams are released in a timely manner.

This prevents us eating up file handles longer than we should be, but as a secondary effect also means we'll avoid having multiple open file streams to the same file. It's a long shot, that might help with #2441 if having multiple streams is somehow causing occasional garbage reads (I doubt it - but on the other hand the only-occasionally-broken weirdness of #2441 is so crazy that maybe the solution is more crazy).

Tested by loading the game and switching between all the mods.

[Diff ignoring whitespace](https://github.com/OpenRA/OpenRA/pull/8762/files?w=1)

### DIFF
--- a/OpenRA.Game/FileSystem/Folder.cs
+++ b/OpenRA.Game/FileSystem/Folder.cs
@@ -13,7 +13,7 @@ using System.IO;
 
 namespace OpenRA.FileSystem
 {
-	public class Folder : IFolder
+	public sealed class Folder : IFolder
 	{
 		readonly string path;
 		readonly int priority;
@@ -78,5 +78,7 @@ namespace OpenRA.FileSystem
 				using (var writer = new BinaryWriter(dataStream))
 					writer.Write(file.Value);
 		}
+
+		public void Dispose() { }
 	}
 }
--- a/OpenRA.Game/FileSystem/GlobalFileSystem.cs
+++ b/OpenRA.Game/FileSystem/GlobalFileSystem.cs
@@ -17,7 +17,7 @@ using OpenRA.Primitives;
 
 namespace OpenRA.FileSystem
 {
-	public interface IFolder
+	public interface IFolder : IDisposable
 	{
 		Stream GetContent(string filename);
 		bool Exists(string filename);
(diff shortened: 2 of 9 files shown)


Mailaender:
Resource loading still works properly across all mods on Linux. :white_check_mark:

OpenRA.Game/FileSystem/Folder.cs
```diff
@@ -78,5 +78,9 @@ public void Write(Dictionary<string, byte[]> contents)
 				using (var writer = new BinaryWriter(dataStream))
 					writer.Write(file.Value);
 		}
+
+		public void Dispose()
+		{
+		}
```
abcdefg30: `public void Dispose() { }`, please.

OpenRA.Game/FileSystem/GlobalFileSystem.cs
```diff
@@ -127,6 +127,8 @@ public static void Mount(string name, string annotation = null)
 
 		public static void UnmountAll()
 		{
+			foreach (var folder in MountedFolders)
+				folder.Dispose();
```
Phrohdoh: Newline after this, and below, please.

Phrohdoh:
I do not know or understand the dispose pattern / implementation, but this code looks logical and the asset browser, map editor, and skirmish works in all mods under OS X for me. :+1:

There are some changes being requested though.

OpenRA.Game/FileSystem/GlobalFileSystem.cs
RoosterDragon: Added.

OpenRA.Game/FileSystem/Folder.cs
RoosterDragon: Amended.

pchote:
Looks good to me :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/e2afffae8e48b50d7464cb45e7764051cf945f0b%5E...e2afffae8e48b50d7464cb45e7764051cf945f0b

GraionDilach:
This entry in the changelog ended up typoed according to https://github.com/OpenRA/OpenRA/wiki/Changelog%20%28bleed%29

abcdefg30:
Thanks, [fixed](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f47f0bc9551731c902f0a91355410da2f1b135b2...419c3fa437ab82b107766f41b347c8fd61a92532).

PR #8303, opened 2015-05-30
TD Balance Changes

AoAGeneral:
http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=17430

List of changes on forums.
#8244 is now out of date due to technical issues. Please close when possible/delete.

Ltnk: Changes
HP 350 from 360
damage vs heavy 90 from 100

Bike: Changes
Price 500 from 600

Artillery: Changes
Chance to explode 0 from 75

Chem Warrior: Changes
reload delay 65 from 70

Genadier: Changes
Grenade projectile speed 140 from 119

Hummer/Buggy: Changes
Damage vs heavy 15 from 20

PR #8765, opened 2015-07-24
Improved Explodes trait and how artillery explodes

reaperrr:
Added `LoadedChance` to `Explodes`, changed how RA/TD artillery explodes.

See individual commits for more detailed descriptions.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Explodes.cs
+++ b/OpenRA.Mods.Common/Traits/Explodes.cs
@@ -14,14 +14,22 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
+	[Desc("This actor explodes when killed.")]
 	public class ExplodesInfo : ITraitInfo
 	{
-		[WeaponReference]
+		[WeaponReference, Desc("Weapon to use for explosion if ammo/payload is loaded.")]
 		public readonly string Weapon = "UnitExplode";
-		[WeaponReference]
+
+		[WeaponReference, Desc("Weapon to use for explosion if no ammo/payload is loaded.")]
 		public readonly string EmptyWeapon = "UnitExplode";
 
+		[Desc("Chance that the explosion will use Weapon if the actor has ammo/payload.")]
+		public readonly int LoadedChance = 100;
+
+		[Desc("Chance that this actor will explode at all.")]
 		public readonly int Chance = 100;
+
+		[Desc("DeathType(s) to apply upon explosion.")]
 		public readonly string[] DeathType = null;
 
 		public object Create(ActorInitializer init) { return new Explodes(this); }
@@ -60,7 +68,8 @@ namespace OpenRA.Mods.Common.Traits
 		string ChooseWeaponForExplosion(Actor self)
 		{
 			var shouldExplode = self.TraitsImplementing<IExplodeModifier>().All(a => a.ShouldExplode(self));
-			return shouldExplode ? info.Weapon : info.EmptyWeapon;
+			var useFullExplosion = self.World.SharedRandom.Next(100) <= info.LoadedChance;
+			return (shouldExplode && useFullExplosion) ? info.Weapon : info.EmptyWeapon;
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


reaperrr:
I don't want to make that call on my own, but I think this should go into next playtest/release.

pchote:
> Reduced TD UnitExplode damage to 1/10 of old value.

This seems a bit drastic.  What's the balance motivation for such a large change?

reaperrr:
> This seems a bit drastic. What's the balance motivation for such a large change?

That this explosion currently isn't used at all in TD since #8303 removed it from artillery precisely because it dealt way too much (splash) damage.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -14,14 +14,22 @@
...
 {
+	[Desc("This actor explodes when killed.")]
 	public class ExplodesInfo : ITraitInfo
 	{
-		[WeaponReference]
+		[WeaponReference, Desc("Weapon to use for explosion if ammo/payload is loaded.")]
 		public readonly string Weapon = "UnitExplode";
-		[WeaponReference]
+
+		[WeaponReference, Desc("Weapon to use for explosion if no ammo/payload.")]
```
abcdefg30: "`if no ammo/payload` ... is loaded."?

abcdefg30:
Some thing about `LoadedChance` seems wrong to me...
If you set it to 0, the Weapon explosion will always occur.

Unit158:
As it should. Loaded chance only changes the likeliness of a different weapon being used.

abcdefg30:
`[Desc("Chance that the explosion will use Weapon if the actor has ammo/payload.")]`
Sounds to me that `Weapon` shouldn't be used if the chance is 0.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -60,7 +68,8 @@ public void Killed(Actor self, AttackInfo e)
 		string ChooseWeaponForExplosion(Actor self)
 		{
 			var shouldExplode = self.TraitsImplementing<IExplodeModifier>().All(a => a.ShouldExplode(self));
-			return shouldExplode ? info.Weapon : info.EmptyWeapon;
+			var useFullExplosion = self.World.SharedRandom.Next(100) > info.LoadedChance;
```
Unit158: > should be <= :)

reaperrr:
Updated.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -14,14 +14,22 @@
...
-		[WeaponReference]
+		[WeaponReference, Desc("Weapon to use for explosion if ammo/payload is loaded.")]
 		public readonly string Weapon = "UnitExplode";
-		[WeaponReference]
+
+		[WeaponReference, Desc("Weapon to use for explosion if no ammo/payload is loaded.")]
 		public readonly string EmptyWeapon = "UnitExplode";
 
+		[Desc("Chance that the explosion will use Weapon if the actor has ammo/payload.")]
+		public readonly int LoadedChance = 100;
```
Mailaender: This changes the previous default.

Mailaender: Or wait probably not. This is just confusingly complicated.

Mailaender:
Stuff still explodes properly. :white_check_mark:

obrakmann:
The artillery in RA now makes the 'artillery firing sound' when it explodes. TD is fine.

reaperrr:
Updated.

- Reverted RA changes to `UnitExplode`, since it's the default explosion used by some traits (crashing aircraft, for example). This was not necessary for TD since it uses custom `HeliCrash` and `HeliExplode` for helicopters.
- Introduced an `ArtilleryExplosion` to RA instead, which is now used as `Weapon` when artillery explodes. 
- Artillery `EmptyWeapon` is now `UnitExplodeSmall` like for all other vehicles.

obrakmann:
Looks ok now. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/82f45c2533da91fd4efb8e4f3de023331c2ec8c4%5E...82f45c2533da91fd4efb8e4f3de023331c2ec8c4)

Issue #8763, opened 2015-07-22
AttackMedic and AutoHeal duplicates the logic of AttackFrontal and AutoTarget


GraionDilach:
Currently we have a duplicate logic for healers. This means that if a unit wants to be a healers/active buffer with a damaging weapon as well, the unit can either resort to force-fire to heal/buff (along with forcing the damaging weapons to not work on allies to make sure it won't get fired along with the bufferweapon) via using the right Attack* or just split the functionality of and use purely `AttackMedic` etc.

This also means that healer/buffers are hardcoded to have "`AttackFrontal`".

What I'd propose that instead of the duplication logic, there would be some kind of a `Benefical` tag on the weapon and `AutoTarget` would look at that - if `Benefical=yes`, allied autotargeting is allowed with that weapon, if no, enemy autotargeting is allowed on that weapon.

#8563 is a case where this becomes a problem.

matija-hustic:
I've worked on this a little locally and got these ideas that I'd like to discuss.
* Right now at most one subclass of `AttackBase` is allowed per actor. Also, exactly one pair of `Cursor` and `OutsideRangeCursor` per `AttackBase` is used. The former is not necessarily a problem, but the latter probably is - different cursors for attack and buff, different cursors for ground attack and air attack for example... I'd suggest moving `Cursor` and `OutsideRangeCursor` to `Armament`.
* Related to this, @GraionDilach, is there any specific reason why you'd want the `Beneficial` on the weapon instead of the `Armament`?
* I imagine then that targeting with the force modifier would in some instances have to discriminate between different available armaments. Right now the force modifier doesn't affect the choice of the armament. It only affects whether or not "a target is targetable". However, an armed medic for example might want to normally heal allied units, but when using the force modifier, shoot them. This would be the case with #8563, a tesla trooper would normally want to powerup a tesla coil, but when force targeting it, he might want to zap it.

GraionDilach:
>Related to this, @GraionDilach, is there any specific reason why you'd want the `Beneficial` on the weapon instead of the `Armament`?
There practically wasn't, it was a draft I didn't give much thought about. Either way if we'd go with `Benefical` - regardless of place - then indeed, force-attack should ignore it.

Moving the cursors to Armament-level sounds odd to me - that same logic could apply to voices (Yuri's Revenge had such for Magnetron, Boris) however I'm not certain how feasible would either be with OpenRA's system.

matija-hustic:
[...]

Issue #8184, opened 2015-05-17
AttackMove is broken for some unit types with "Defend" stance


obrakmann:
While working on https://github.com/OpenRA/OpenRA/compare/bleed...obrakmann:default-stance-defend I noticed that AttackMove is broken for E3s when their stance is set to "Defend" (in ra at least, didn't test the other mods or other units yet). They acquire a target, but aren't close enough to fire and won't move in range. E1s don't seem to have that problem.

![screenshot from 2015-05-17 20 34 06](https://cloud.githubusercontent.com/assets/4331210/7671571/c15d7fb6-fcd6-11e4-96b3-89d543922590.png)
![screenshot from 2015-05-17 20 31 43](https://cloud.githubusercontent.com/assets/4331210/7671570/c15d77b4-fcd6-11e4-98c3-3d3ee9475ecb.png)

I can reproduce this in release-20141029, so it's not really a recent bug.

obrakmann:
I found the problem. E3 has two weapons, a bazooka against ground targets with a range of five cells, and a MANPAD with a range of 7.5 cells. When a target gets within the MANPADs range, the E3 stops to attack, but can't because it's not a suitable weapon for the target. It can't go any closer though, either, because it's stance is "Defend", so it cannot move during the attack.

matija-hustic:
Could you check if #8768 solved this?

abcdefg30:
I was.

PR #8768, opened 2015-07-24
Some work on Attack*

matija-hustic:
Some rewiring (not a full refactor) in the attacking mechanism to accommodate using different cursors and voices with different armaments, as well as combining beneficial (buffs) and damaging weapons for a single actor.

Eliminates the need for the medic to subclass `AttackFrontal` and `AutoHeal` or use `Heal`.

I'll type up the upgrade rule after review of other code.

Discussed to some extent in #8763. Prerequisite for #8563.


OpenRA.Mods.Common/Traits/Armament.cs
```diff
@@ -62,6 +62,19 @@ public class ArmamentInfo : UpgradableTraitInfo, Requires<AttackBaseInfo>
 		[Desc("Use multiple muzzle images if non-zero")]
 		public readonly int MuzzleSplitFacings = 0;
 
+		[Desc("Is it possible to use the armament when force-attacking?")]
+		public readonly bool ForceAttacks = true;
+
+		[Desc("Is the armament beneficial (should it be used on allies)?")]
+		public readonly bool Beneficial = false;
```
Phrohdoh: Why not name this `CanTargetAllies`?

GraionDilach: Force-attack ignores this tag. Keep in mind this is used for target acquisition to define it. Also, `CanTargetAllies: false` wouldn't suggest that the weapon is used on enemies.

penev92: No, it wouldn't, as those are different things. `Beneficial` doesn't sound very good to me too. Let's try to come up with an alternative.
`IsUsedOnAllies` ?  `AutoTargetsAllies` ?

matija-hustic: I was thinking along the lines of `MeantForAllies` or `CannotTargetEnemies`

GraionDilach: See https://github.com/Ares-Developers/Ares/blob/master/src/Ext/Techno/Hooks.cpp#L632 why any `*Allies` are wrong.

penev92: All `*Allies` can't be wrong since this weapon is going to only be used on allies.
So something like `IsUsedOnlyOnAllies` ( + `WhenAutoTargeting`).

GraionDilach: `IntendedForAllies` then. But really, any `Allies` tag would mean - with the long term plan to remove the fragile alliances - that there should be an `Enemies` counterpart which I'm definitely not a fan of.

matija-hustic: This is the only place where `Beneficial` is used

https://github.com/OpenRA/OpenRA/pull/8768/files#diff-26d2cbcfb01852edd1172678f8116784R244

Right now,
* if an armament has `Beneficial == false`,
  * it can be used to target enemies as usual
    * without force-attack,
    * and with it if additionally `ForceAttacks == true`
  * and allies
    * only using force-attack if additionally `ForceAttacks == true`
* if `Beneficial == true`,
  * only allies can be targeted with it
    * without force-attack,
    * and with it if additionally `ForceAttacks == true`
  * enemies cannot be targeted at all
    * neither with
    * nor without force-attack

UberWaffe:
A suggestion:

"NotHarmful" instead of "Beneficial" ?
That would imply that a unit wouldn't care about getting hit by it.

A healing serum would be not harmful, while a targeting laser would be harmful (even though it doesn't directly cause harm, it still results in harm.)

OpenRA.Mods.Common/Traits/Armament.cs
UberWaffe: A suggestion:

"NotHarmful" instead of "Beneficial" ?
That would imply that a unit wouldn't care about getting hit by it, regardless of its diplomatic stance.

A healing serum would be not harmful, while a targeting laser would be harmful (even though it doesn't directly cause harm, it still results in harm.)

abcdefg30:
RA medic and mechanic still work as well as the testcase. :+1: / :white_check_mark:

OpenRA.Mods.Common/Traits/Armament.cs
abcdefg30: So, what about this now? Leave as-is or change to a suggestion?

Unit158:
Needs a rebase

OpenRA.Mods.Common/Traits/Armament.cs
pchote: You can remove the confusion and make it more consistent with our other traits by defining:
```csharp
public readonly Stance Targets: Stance.Enemy;
public readonly Stance ForceTargets: Stance.Enemy | Stance.Neutral | Stance.Ally;
```

OpenRA.Mods.Common/Warheads/DamageWarhead.cs
```diff
@@ -48,13 +48,27 @@ public int DamageVersus(ActorInfo victim)
...
 				return false;
 
-			return DamageVersus(victim) > 0;
+			if (Damage < 0 && victim.GetDamageState() == DamageState.Undamaged)
+				return false;
+
+			return DamageVersus(victim.Info) > 0;
+		}
+
+		// TODO: This can be removed after the legacy and redundant 0% = not targetable
```
pchote: This assumption has now been removed.

pchote:
This is a good step towards a much-needed cleanup, so its a shame to see it sit idle.  Are you able to work on this any time soon @matija-hustic, or should somebody else take this over?

obrakmann:
I think [these two lines](https://github.com/matija-hustic/OpenRA/blob/attackbase_refactor/OpenRA.Mods.Common/Traits/AutoTarget.cs#L131-L132) need to be changed to as well to fix #8481 and #9266. If `currentTarget` is not null, you'll want `range` to be the range of a valid weapon against it, not `attack.GetMaximumRange()`.

matija-hustic:
Hey guys, I'm still in the middle of moving to a new city and I'm gonna need at least another week before I'm able to do anything. The time it's taking has surpassed even my own initial predictions :( It's probably going to take another week. If you're in a hurry to get my PRs off the queue, have someone else take over what ever they can.

matija-hustic:
I've updated this. #8184 should no longer happen. I haven't checked #9266, though.

matija-hustic:
I've made `ChooseArmamentsForTarget` ignore disabled armaments and now it seems #9266 is gone, too. At least I haven't been able to reproduce it.

Phrohdoh:
Yes that would be one fix for that.

FWIW

I'd personally prefer a field like such on AttackBase:
```csharp
    EnabledArmaments { return Armaments.Where(a => !a.IsTraitDisabled); }
```

Or perhaps a `GetEnabledArmaments()` method to indicate work being done.

matija-hustic:
@Phrohdoh, how about i put a `bool Enabled = true` in the `ChooseArmamentsForTarget`'s arguments?
Or do you need a separate function?

Phrohdoh:
That ought to suffice though I think a more fitting name would be `onlyEnabled` or something to that effect.

matija-hustic:
Done

reaperrr:
I'm putting `Has Dependencies` on this so it doesn't get accidentally merged with testcase and without upgrade rules.

reaperrr:
Also needs a rebase now.

matija-hustic:
Rebased. Upgrade rule now in there. I'll leave the `TEST Medic` commit in there until the final review.

Otherwise, I've checked and #9266 should be gone, too, now, not only #8184.

OpenRA.Mods.Common/Activities/Attack.cs
```diff
@@ -64,8 +67,16 @@ protected virtual Activity InnerTick(Actor self, AttackBase attack)
...
 				return NextActivity;
 
+			// Drop the target once none of the weapons are effective against it
+			var armaments = attack.ChooseArmamentsForTarget(Target, forceAttack);
+			if (!armaments.Any())
+				return NextActivity;
+
+			// Update ranges
+			minRange = armaments.MaxBy(a => a.Weapon.MinRange.LengthSquared).Weapon.MinRange;
+			maxRange = armaments.MinBy(a => a.Weapon.Range.LengthSquared).MaxRange();
```
pchote: You need to order by `a.MaxRange()`, which includes range modifiers, instead of `a.Weapon.Range`.

pchote: `armaments.Min(a => a.MaxRange())` should work, with a similar simplification for `minRange`.

matija-hustic:
Changed as commented

OpenRA.Mods.Common/Traits/Armament.cs
```diff
@@ -65,6 +65,17 @@ public class ArmamentInfo : UpgradableTraitInfo, IRulesetLoaded, Requires<Attack
 		public WeaponInfo WeaponInfo { get; private set; }
 		public WDist ModifiedRange { get; private set; }
 
+		public readonly Stance Targets = Stance.Enemy;
```
atlimit8: `Targets` could be confused with target types. Perhaps `TargetAllegiances` or possibly just `Allegiances`.

pchote: Our convention is to use "stance" when referring to stances/allegiances.

atlimit8:
The "PR: Needs +2" was outdated anyways.

matija-hustic:
Changed

obrakmann:
Works, doesn't break anything afaics, and code looks ok, too :+1:

atlimit8:
:+1: - Test commit needs to be remove before merging,

reaperrr:
Also needs a rebase now.

matija-hustic:
OK, so the test commit is out and the branch rebased

penev92:
Can you (if possible) change the first commit's message to something more descriptive than `Some work on Attack*`? This is a lot of changes, so I assume there is a lot to put in there.

matija-hustic:
Ah yes, I forgot that the commits should be more descriptive. Is it better now, @penev92 ?

penev92:
I can't verify that what it says is correct because I haven't reviewed this, but it looks a lot better, thanks ;)

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/6dbb6a0ce3301ce0e346b2ba1881be09e264d4e2)

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2198,6 +2198,52 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 						node.Key = "-WithSpriteRotorOverlay";
 				}
 
+				if (engineVersion < 20151005)
+				{
+					// Move certain properties from Parachutable to new WithParachute trait
+					// Add dependency traits to actors implementing Parachutable
+					// Make otherwise targetable parachuting actors untargetable
```
Mailaender: This comment seems to be unrelated copy & pasta from https://github.com/matija-hustic/OpenRA/blob/69696fc154ebdf596cbb77c438f4f091b552d2d5/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs#L1569

matija-hustic: grrrrr

https://github.com/matija-hustic/OpenRA/commit/9927057c0f76058959141dcecf5816886398f83f

Mailaender: Thanks for the quick fix in https://github.com/OpenRA/OpenRA/pull/9581.

PR #8769, opened 2015-07-24
Remove ESC hotkey from lobby 'Back' button

obrakmann:


Phrohdoh:
Why?

obrakmann:
https://github.com/OpenRA/OpenRA/pull/8680#discussion-diff-35460324

penev92:
Well as far as I know the Esc key has never actually worked in the lobby until today, so I guess no harm done.
:+1:

pchote:
Works as advertised :+1:.  Skipping the changelog entry because this is effectively fixing a regression from #8680.

PR #8770, opened 2015-07-24
Throttle AI order speed to prevent lag spikes

### LINKED ISSUE #4584: very bad lag spikes due to pathfinding
After about 10 or 15 mins the game starts to freeze and play fine every few secs. I turned on the graph and the (white) pathingfind and (darkblue) Tick_time start spiking and eventually start going off my screen. none of my components are maxing. Was is the cause so i can turn it off or fix it. I have a intel i5 radeon 280x 16gb ram win 7 ult

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1400010-very-bad-lag-spikes-due-to-pathfinding?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #8384: Jittery game play
I get jittery game play after I've been playing for a while. I'm on Ubuntu-MATE 15.04.
video demo https://www.youtube.com/watch?v=WGkGbKDsIeQ
The issue happens even in the latest version.

I don't get this in the Windows version.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/20627094-jittery-game-play?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #8466: Intermittent freezing in AI matches with >6 players
open-ra_1

I have game slowdowns on maps with more than 4 players, but the problem is more prevalent with 6 or more in a map. Every 1-2 seconds the game will freeze for about 1-1.5 seconds before unfreezing, after which it freezes again and the process repeats itself.

I'm running a HP Pavilion g4 laptop, with a Intel Core i5-3210M CPU, 4GBs of RAM, and a Radeon HD 7670M with integrated Intel HD Graphics 4000 as my graphics card.

Please do advise, thank you.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/21696355-intermittent-freezing-in-ai-matches-with-6-players?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RoosterDragon:
Fixes #8466
Fixes #8384
Fixes #4584
## Issue

In games with AI, as the game progresses there will start to be lag spikes every ~1 second. The spikes get worse with more AI.
## Cause

AIs re-evaluate their current state at set intervals. For example [assigning of roles](https://github.com/OpenRA/OpenRA/blob/7d56209f3b772ea847cc8c9decb4c33cc4210e7d/OpenRA.Mods.Common/AI/HackyAI.cs#L40-L41) and [updating of squads](https://github.com/OpenRA/OpenRA/blob/7d56209f3b772ea847cc8c9decb4c33cc4210e7d/OpenRA.Mods.Common/AI/HackyAI.cs#L46-L47). You'll notice both of these are 20/30 ticks - which are both roughly 1 second.

Luckily, evaluating AI state is fast.

What happens then is that the AIs issue new orders to be carried out. In a typical case they might issue say 10 orders to one of the squads to attack.

Luckily, issuing orders is fast.

[These orders are processed during the tick by the order manager](https://github.com/OpenRA/OpenRA/blob/7d56209f3b772ea847cc8c9decb4c33cc4210e7d/OpenRA.Game/Network/OrderManager.cs#L212).

Luckily, processing orders is fast.

[Each time an order is processed, a hash of the world state must be generated for sync purposes](https://github.com/OpenRA/OpenRA/blob/7d56209f3b772ea847cc8c9decb4c33cc4210e7d/OpenRA.Game/Network/OrderManager.cs#L213). This allows clients to identify the order responsible in the event of a desync.

Unluckily, hashing the world state is extremely slow.

In games with several AI, a few AI will decide to send orders to their squads, each squad needing maybe a dozen orders. They do this all at the exact same tick. This results in a flood of orders (say about 50 orders instead of the more usual 0-5) to be processed in a single tick. Hashing the world state so many times as a result of processing the orders is the reason for the lag spike.
## The Fix

I address this issue by limiting the ability of the AI to generate a flood of orders in a single tick. Instead, I spread out the orders over several ticks.

The first commit makes each AI in the game operate on different ticks. In games with lots of AI, this prevents them teaming up to make one tick very slow.

The second limits the number of orders an AI can issue per tick. Extra orders are queued up and processed on subsequent ticks. This prevents spikes from issuing orders to entire squads at once by forcing the AI to order each unit on separate ticks.

This throttling means that orders might be outdated by the time they are issued. This seems to work in my testing but somebody who knows how orders work better than I do should confirm I'm not introducing a massive chance for crashes and/or desyncs.
## Results

Perf graphs before and after (we care about the teal line representing tick_time)

Perf Graph Comparion

You can check this yourself by just spectating a game with a large number of AI. Wait a few minutes until they start attacking each other and the lag spikes should appear. With this PR they should be prevented.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -58,6 +58,9 @@ namespace OpenRA.Mods.Common.AI
 		[Desc("How long to wait (in ticks) between structure production checks ticks when actively building things.")]
 		public readonly int StructureProductionActiveDelay = 10;
 
+		[Desc("Minimum portion of pending orders to issue each tick (e.g. 5 issues at least 1/5th of all pending orders). Excess orders remain queued for subsequent ticks.")]
+		public readonly int MinOrderQuotientPerTick = 5;
+
 		[Desc("Minimum range at which to build defensive structures near a combat hotspot.")]
 		public readonly int MinimumDefenseRadius = 5;
 
@@ -189,7 +192,7 @@ namespace OpenRA.Mods.Common.AI
 		// Units that the ai already knows about. Any unit not on this list needs to be given a role.
 		List<Actor> activeUnits = new List<Actor>();
 
-		public const int FeedbackTime = 30;		// ticks; = a bit over 1s. must be >= netlag.
+		public const int FeedbackTime = 30; // ticks; = a bit over 1s. must be >= netlag.
 
 		public readonly World World;
 		public Map Map { get { return World.Map; } }
@@ -199,6 +202,8 @@ namespace OpenRA.Mods.Common.AI
 		int assignRolesTicks;
 		int attackForceTicks;
 		int minAttackForceDelayTicks;
+
+		readonly Queue<Order> orders = new Queue<Order>();
 
 		public HackyAI(HackyAIInfo info, ActorInitializer init)
 		{
@@ -243,6 +248,11 @@ namespace OpenRA.Mods.Common.AI
 			resourceTypeIndices = new BitArray(World.TileSet.TerrainInfo.Length); // Big enough
 			foreach (var t in Map.Rules.Actors["world"].Traits.WithInterface<ResourceTypeInfo>())
 				resourceTypeIndices.Set(World.TileSet.GetTerrainIndex(t.TerrainType), true);
+		}
+
+		public void QueueOrder(Order order)
+		{
+			orders.Enqueue(order);
 		}
 
 		ActorInfo ChooseRandomUnitToBuild(ProductionQueue queue)
@@ -449,6 +459,10 @@ namespace OpenRA.Mods.Common.AI
 
 			foreach (var b in builders)
 				b.Tick();
+
+			var ordersToIssueThisTick = Math.Min((orders.Count + Info.MinOrderQuotientPerTick - 1) / Info.MinOrderQuotientPerTick, orders.Count);
+			for (var i = 0; i < ordersToIssueThisTick; i++)
+				World.IssueOrder(orders.Dequeue());
 		}
 
 		internal Actor ChooseEnemyTarget()
@@ -592,7 +606,7 @@ namespace OpenRA.Mods.Common.AI
 					continue;
 
 				// Tell the idle harvester to quit slacking:
-				World.IssueOrder(new Order("Harvest", a, false));
+				QueueOrder(new Order("Harvest", a, false));
 			}
 		}
 
@@ -606,7 +620,7 @@ namespace OpenRA.Mods.Common.AI
 			foreach (var a in newUnits)
 			{
 				if (a.HasTrait<Harvester>())
-					World.IssueOrder(new Order("Harvest", a, false));
+					QueueOrder(new Order("Harvest", a, false));
 				else
 					unitsHangingAroundTheBase.Add(a);
 
@@ -702,7 +716,7 @@ namespace OpenRA.Mods.Common.AI
 					!IsRallyPointValid(rp.Trait.Location, rp.Actor.Info.Traits.GetOrDefault<BuildingInfo>())).ToArray();
 
 			foreach (var a in buildings)
-				World.IssueOrder(new Order("SetRallyPoint", a.Actor, false) { TargetLocation = ChooseRallyLocationNear(a.Actor), SuppressVisualFeedback = true });
+				QueueOrder(new Order("SetRallyPoint", a.Actor, false) { TargetLocation = ChooseRallyLocationNear(a.Actor), SuppressVisualFeedback = true });
 		}
 
 		// Won't work for shipyards...
@@ -734,7 +748,7 @@ namespace OpenRA.Mods.Common.AI
 				// Don't transform the mcv if it is a fact
 				// HACK: This needs to query against MCVs directly
 				if (mcv.HasTrait<Mobile>())
-					World.IssueOrder(new Order("DeployTransform", mcv, false));
+					QueueOrder(new Order("DeployTransform", mcv, false));
 			}
 			else
 				BotDebug("AI: Can't find BaseBuildUnit.");
@@ -759,8 +773,8 @@ namespace OpenRA.Mods.Common.AI
 				if (desiredLocation == null)
 					continue;
 
-				World.IssueOrder(new Order("Move", mcv, true) { TargetLocation = desiredLocation.Value });
-				World.IssueOrder(new Order("DeployTransform", mcv, true));
+				QueueOrder(new Order("Move", mcv, true) { TargetLocation = desiredLocation.Value });
+				QueueOrder(new Order("DeployTransform", mcv, true));
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


Phrohdoh:
This shows great improvement in a single game of TS (sorry I can't test more now).
We'll need to stress-test this quite a bit in all mods, but so far bravo and thanks @RoosterDragon!
https://drive.google.com/file/d/0B4Mag-oGBQNCVmNuWXpDVEtwa3M/view?usp=sharing

chrisforbes:
Seems like a good idea to me.

Real players can cause similar problems with lots of units selected, but AI
is much better at doing it consistently.

pchote:
This plus #8759 appear to completely fix the common lag spikes.  Great Job!

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -58,6 +58,9 @@ public sealed class HackyAIInfo : IBotInfo, ITraitInfo
 		[Desc("How long to wait (in ticks) between structure production checks ticks when actively building things.")]
 		public readonly int StructureProductionActiveDelay = 10;
 
+		[Desc("Limits the AI to issuing this many orders per tick. Orders that cannot be issued in this tick are queued for the next tick.")]
+		public readonly int MaxOrdersPerTick = 1;
```
pchote: The order queue grows unboundedly as soon as the average order count per tick gets larger than this.  I found that the AI would grind to a halt in the late game until I turned this up.

That obviously isn't a great solution (since it just defers the problem until they have sufficiently more units, and also hurts perf in the early game) - can we dynamically scale this based on the current queue size instead?

RoosterDragon:
I've implemented a parameter to process a portion of the queue size rather than a fixed amount per turn in order to reduce the risk of AIs queueing orders at a faster rate than they can issue them as @pchote noted. The default value might need some tuning.

Mailaender:
I played this with two local clients which desynced:

```
Exception of type `System.InvalidOperationException`: Out of sync in frame 4705: Blame Tick.
 Compare syncreport.log with other players.
  at OpenRA.Network.OrderManager.OutOfSync (Int32 frame, System.String blame) [0x0001e] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Network/OrderManager.cs:63 
  at OpenRA.Network.OrderManager.CheckSync (System.Byte[] packet) [0x00074] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Network/OrderManager.cs:161 
  at OpenRA.Network.OrderManager+<TickImmediate>c__AnonStorey0.<>m__0 (Int32 clientId, System.Byte[] packet) [0x00050] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Network/OrderManager.cs:119 
  at OpenRA.Network.ReplayRecorderConnection+<Receive>c__AnonStorey0.<>m__0 (Int32 client, System.Byte[] data) [0x00094] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Network/ReplayRecorderConnection.cs:95 
  at OpenRA.Network.EchoConnection.Receive (System.Action`2 packetFn) [0x00055] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Network/Connection.cs:103 
  at OpenRA.Network.ReplayRecorderConnection.Receive (System.Action`2 packetFn) [0x00027] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Network/ReplayRecorderConnection.cs:82 
  at OpenRA.Network.OrderManager.TickImmediate () [0x000c0] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Network/OrderManager.cs:112 
  at OpenRA.Sync+<CheckSyncUnchanged>c__AnonStorey0.<>m__0 () [0x00007] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Sync.cs:168 
  at OpenRA.Sync.CheckSyncUnchanged[Boolean] (OpenRA.World world, System.Func`1 fn) [0x00044] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Sync.cs:185 
  at OpenRA.Sync.CheckSyncUnchanged (OpenRA.World world, System.Action fn) [0x0001b] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Sync.cs:168 
  at OpenRA.Game.InnerLogicTick (OpenRA.Network.OrderManager orderManager) [0x0011c] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:462 
  at OpenRA.Game.LogicTick () [0x0004e] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:507 
  at OpenRA.Game.Loop () [0x000d4] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:631 
  at OpenRA.Game.Run () [0x00042] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:671 
  at OpenRA.Program.Run (System.String[] args) [0x00011] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Support/Program.cs:111 
  at OpenRA.Program.Main (System.String[] args) [0x00050] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Support/Program.cs:39
```

I only got 1 syncreport.log for obvious reasons: https://gist.github.com/Mailaender/ff047a4d04d175ca0cab

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -435,6 +459,10 @@ public void Tick(Actor self)
 
 			foreach (var b in builders)
 				b.Tick();
+
+			var ordersToIssueThisTick = Math.Min((int)Math.Ceiling((double)orders.Count / Info.MinOrderQuotientPerTick), orders.Count);
```
Mailaender: Can you do this with integer math only?

pchote:
> I played this with two local clients and many bots which desynced the game

This PR changes the code for issuing bot orders, which only runs on one client.  It shouldn't be able to desync.

OpenRA.Mods.Common/AI/HackyAI.cs
pchote: `(orders.Count + Info.MinOrderQuotientPerTick - 1) / Info.MinOrderQuotientPerTick` is the equivalent integer calculation.

obrakmann:
FWIW, I just had a 45 minute AI match with a second client as spectator and didn't get any desyncs.

obrakmann:
This latest iteration works very well. In several test games I didn't notice any mis-behaving, passive AIs anymore. The code changes are straight-forward, too. :+1:

Mailaender:
In that case I stumbled upon another d2k mod exclusive desync I guess :(

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8e6edb08a71aa67ebf46eb07a246517d847ae84f...29c0bbfc83d4d7a62269e8d2603dd440e3b40cde)

hxazgalor:
That's good to hear. Has this been implemented in the latest playtest release (20150808)? Thanks again!

Mailaender:
Yes, though we are probably not done here.

PR #8773, opened 2015-07-25
Alter install logic to convert case of file names

DanAE111:
I had issues with music not being available after installing the TS content from the original disc. Made some simple changes to the YAML so case sensitive OSes should work too!

### DIFF
--- a/OpenRA.Mods.Common/InstallUtils.cs
+++ b/OpenRA.Mods.Common/InstallUtils.cs
@@ -102,7 +102,7 @@ namespace OpenRA.Mods.Common
 
 				foreach (var file in directory.Value)
 				{
-					var dest = Path.Combine(destPath, targetDir, file);
+					var dest = Path.Combine(destPath, targetDir, file.ToLowerInvariant());
 					if (File.Exists(dest))
 					{
 						if (overwrite)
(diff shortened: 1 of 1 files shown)


abcdefg30:
Looks simple enough, but I can't test it. :+0.5:

mods/ts/mod.yaml
```diff
@@ -20,7 +20,7 @@ MapFolders:
 
 Packages:
 # Tiberian Sun
-	~scores.mix:CRC32
+	~SCORES.MIX:CRC32
```
Mailaender: :-1: This will clash with the filename when downloading. https://github.com/OpenRA/OpenRAWeb/blob/master/content/packages/ts-music-mirrors.txt

pchote: We need to fix the music packages then.

Mailaender: Or simple `ToLowerInvariant` when extracting/copying those from CD.

dan9550: Given we have two solutions to the issue, what is the preferred fix?

Mailaender: We have lower-case package lists in all other C&C mod so let's keep it that way for consistency.

penev92:
Is something happening with this?

dan9550:
Updated, tested on both a Windows 10 x64 machine and Ubuntu 15.04 x64. Installing Tiberian Sun assets from the Nod disc.

abcdefg30:
You should squash the commits.

Mailaender:
:+1: Looking good.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/22fb500391ef9d24a2e168d55e6cc8284c75ac50...07df7436b23566883f369e4e542a34ba184c1879)

PR #8774, opened 2015-07-25
Add StartSequence to WithIdleOverlay & change V19.Husk

atlimit8:
V19.Husk uses RenderBuilding & WithSpriteBody which implement IQuantizeBodyOrientation.
The problem is that `self.Info.Traits.GetOrDefault<IQuantizeBodyOrientationInfo>()` is called in BodyOrientation. While this probably has no noticeable affect on the game, it can be fixed simply by adding these two lines.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
@@ -19,6 +19,9 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Renders a decorative animation on units and buildings.")]
 	public class WithIdleOverlayInfo : UpgradableTraitInfo, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>, Requires<IBodyOrientationInfo>
 	{
+		[Desc("Animation to play when the actor is created.")]
+		[SequenceReference] public readonly string StartSequence = null;
+
 		[Desc("Sequence name to use")]
 		[SequenceReference] public readonly string Sequence = "idle-overlay";
 
@@ -67,7 +70,11 @@ namespace OpenRA.Mods.Common.Traits
 
 			buildComplete = !self.HasTrait<Building>(); // always render instantly for units
 			overlay = new Animation(self.World, rs.GetImage(self));
-			overlay.PlayRepeating(RenderSprites.NormalizeSequence(overlay, self.GetDamageState(), info.Sequence));
+			if (info.StartSequence != null)
+				overlay.PlayThen(RenderSprites.NormalizeSequence(overlay, self.GetDamageState(), info.StartSequence),
+					() => overlay.PlayRepeating(RenderSprites.NormalizeSequence(overlay, self.GetDamageState(), info.Sequence)));
+			else
+				overlay.PlayRepeating(RenderSprites.NormalizeSequence(overlay, self.GetDamageState(), info.Sequence));
 
 			var anim = new AnimationWithOffset(overlay,
 				() => body.LocalToWorld(info.Offset.Rotate(body.QuantizeOrientation(self, self.Orientation))),
--- a/OpenRA.Mods.Common/Traits/Render/WithSpriteBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithSpriteBody.cs
@@ -57,11 +57,11 @@ namespace OpenRA.Mods.Common.Traits
 			DefaultAnimation = new Animation(init.World, rs.GetImage(init.Self), baseFacing);
 			rs.Add(new AnimationWithOffset(DefaultAnimation, null, () => IsTraitDisabled));
 
-			if (Info.StartSequence != null)
-				PlayCustomAnimation(init.Self, Info.StartSequence,
-					() => DefaultAnimation.PlayRepeating(NormalizeSequence(init.Self, Info.Sequence)));
+			if (info.StartSequence != null)
+				PlayCustomAnimation(init.Self, info.StartSequence,
+					() => PlayCustomAnimationRepeating(init.Self, info.Sequence));
 			else
-				DefaultAnimation.PlayRepeating(NormalizeSequence(init.Self, Info.Sequence));
+				DefaultAnimation.PlayRepeating(NormalizeSequence(init.Self, info.Sequence));
 		}
 
 		public string NormalizeSequence(Actor self, string sequence)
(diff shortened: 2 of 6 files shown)


reaperrr:
> While this probably has no noticeable affect on the game

I doubt it, without your fix the game would likely crash if either actor appears on map.
So :+1:

pchote:
> While this probably has no noticeable affect on the game

The body of the oil derrick husk is no longer rendered in TD.

reaperrr:
True, should have tested ingame. 
If `RenderBuilding` is removed, you need to add `RenderSprites` (and possibly `AutoSelectionSize` if it has no custom bounds) as well.

reaperrr:
This is messed up completely, actually. Whoever set this up originally probably didn't test either, otherwise he would notice that the `fire-` sequences should be an overlay and are not part of the building sprite itself.

But since `WithIdleOverlay` has no equivalent to `StartSequence` yet, I'd suggest to add an `WithIdleOverlay@UNDERLAY` to the actor with `Sequence: idle`. Then check ingame whether the Z sorting is correct, otherwise give `idle` sequence a `ZOffset: -1`.

atlimit8:
Made changes based on comments and tested in "Allies 03a: Dead End".

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
```diff
@@ -19,6 +19,9 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Renders a decorative animation on units and buildings.")]
 	public class WithIdleOverlayInfo : UpgradableTraitInfo, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>, Requires<IBodyOrientationInfo>
 	{
+		[Desc("Animation to play when the actor is created.")]
+		[SequenceReference] public readonly string StartSequence = null;
```
reaperrr: This doesn't do anything on its own, look at the implementation in `WithSpriteBody` to see how this is played before the looping animation, and then change line 73 accordingly.

reaperrr:
:+1: once my last comment has been adressed (and confirmed to work).

atlimit8:
Actually, I did make those changes, but forgot to save OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs.

Fixed now.

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
```diff
@@ -67,7 +70,11 @@ public WithIdleOverlay(Actor self, WithIdleOverlayInfo info)
 
 			buildComplete = !self.HasTrait<Building>(); // always render instantly for units
 			overlay = new Animation(self.World, rs.GetImage(self));
-			overlay.PlayRepeating(RenderSprites.NormalizeSequence(overlay, self.GetDamageState(), info.Sequence));
+			if (Info.StartSequence != null)
```
reaperrr: Shouldn't this Info actually be lower-case?

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
```diff
@@ -67,7 +70,11 @@ public WithIdleOverlay(Actor self, WithIdleOverlayInfo info)
...
+				overlay.PlayThen(Info.StartSequence,
```
reaperrr: same here

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
atlimit8: `base(info)` sets `Info = info`, so it does the same thing. I could change it, but is it worth it. I doubt it would make an impact on performance to get from a field instead of an argument.

atlimit8: Actually, `WithSpriteBody` does the same thing.

reaperrr:
Works now :+1:

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
reaperrr: Ok, would be more consistent with `info.Sequence` but not a blocker.

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
Mailaender: Isn't this redundant to WithMakeAnimation.cs?

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
```diff
@@ -19,6 +19,9 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Renders a decorative animation on units and buildings.")]
 	public class WithIdleOverlayInfo : UpgradableTraitInfo, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>, Requires<IBodyOrientationInfo>
 	{
+		[Desc("Animation to play when the actor is created.")]
```
Mailaender: Isn't this redundant to WithMakeAnimation.cs?

reaperrr: > Isn't this redundant to WithMakeAnimation.cs?

No, WithMakeAnimation only works with RenderBuilding, and in this case it's the overlay that has a StartSequence, not the building 'body'.

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
pchote: This sequence needs to be normalized to avoid breakage when actors are spawned with < 100% health.  Please also fix that in `WithSpriteBody` if that needs it, and you can earn some bonus points if you fix `Info` &rarr; `info` while you're in these parts of the code.

pchote:
:+1: after fixing the sequence normalization and `info` refs.

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
atlimit8: `WithSpriteBody` has its own helpers for normalization.

atlimit8:
Changed.
`WithSpriteBody` has its own helpers for normalization.

Mailaender:
Idle overlays still work in TS and V19.husk is rendered properly in RA/TD. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/29c0bbfc83d4d7a62269e8d2603dd440e3b40cde...1e4c081fa0456593317fdf17ff4d4f91dd170e91)

PR #8781, opened 2015-07-25
Add plumbing for heighmap-aware shroud and map bounds checks.

pchote:
This introduces a new coordinate system ("projected map cells", or `PPos`) for code that deals with the map border and shroud.  For mods that don't use terrain height `PPos` maps identically to `MPos`.

From the first commit message:

```
PPos is best thought of as a cell grid applied in
screen space.  Multiple cells with different
terrain heights may be projected to the same PPos,
or to multiple PPos if they do not align with the
screen grid (i.e. odd-height or sloped).

PPos coordinates are used primarily for map edge
checks and shroud / visibility queries.
```

For the sake of keeping the PR as reviewable as possible I have limited this to what we need for the shroud.  Future PRs will deal with the map-edge coordinate helpers which are fiddly and harder to review.

Expected behaviour with this PR applied is for there to be no visible changes in TD/RA/D2K, and for the TS shroud rendering to behave logically.  Edge of map behaviour in TS should be no worse than in current bleed.

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -739,10 +739,10 @@ namespace OpenRA
 
 		public bool Contains(MPos uv)
 		{
-			// The first check ensures that the cell is within the valid map region, avoiding
-			// potential crashes in deeper code.  All CellLayers have the same geometry, and
-			// CustomTerrain is convenient (cellProjection may be null and others are Lazy).
-			return CustomTerrain.Contains(uv) && ProjectedCellsCovering(uv).All(containsTest);
+			// TODO: Checking against the bounds excludes valid parts of the map if MaxTerrainHeight > 0.
+			// Unfortunatley, doing this properly leads to memory corruption issues in the (unsafe) radar
+			// rendering code.
+			return Bounds.Contains(uv.U, uv.V) && ProjectedCellsCovering(uv).All(containsTest);
 		}
 
 		public bool Contains(PPos puv)
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Drawing.Imaging;
@@ -245,6 +246,8 @@ namespace OpenRA
 		[FieldLoader.Ignore] public Lazy<CellLayer<byte>> MapHeight;
 
 		[FieldLoader.Ignore] public CellLayer<byte> CustomTerrain;
+		[FieldLoader.Ignore] CellLayer<PPos[]> cellProjection;
+		[FieldLoader.Ignore] CellLayer<List<MPos>> inverseCellProjection;
 
 		[FieldLoader.Ignore] Lazy<TileSet> cachedTileSet;
 		[FieldLoader.Ignore] Lazy<Ruleset> rules;
@@ -252,8 +255,10 @@ namespace OpenRA
 		public SequenceProvider SequenceProvider { get { return Rules.Sequences[Tileset]; } }
 
 		public WVec[][] CellCorners { get; private set; }
-		[FieldLoader.Ignore] public CellRegion CellsInsideBounds;
+		[FieldLoader.Ignore] public ProjectedCellRegion ProjectedCellBounds;
 		[FieldLoader.Ignore] public CellRegion AllCells;
+
+		readonly Func<PPos, bool> containsTest;
 
 		void AssertExists(string filename)
 		{
@@ -268,6 +273,8 @@ namespace OpenRA
 		/// </summary>
 		public Map(TileSet tileset, int width, int height)
 		{
+			containsTest = Contains;
+
 			var size = new Size(width, height);
 			var tileShape = Game.ModData.Manifest.TileShape;
 			var tileRef = new TerrainTile(tileset.Templates.First().Key, (byte)0);
@@ -307,6 +314,8 @@ namespace OpenRA
 		/// <summary>Initializes a map loaded from disk.</summary>
 		public Map(string path)
 		{
+			containsTest = Contains;
+
 			Path = path;
 			Container = GlobalFileSystem.OpenPackage(path, null, int.MaxValue);
 
@@ -409,8 +418,8 @@ namespace OpenRA
 			var br = new MPos(MapSize.X - 1, MapSize.Y - 1).ToCPos(this);
 			AllCells = new CellRegion(TileShape, tl, br);
 
-			var btl = new MPos(Bounds.Left, Bounds.Top);
-			var bbr = new MPos(Bounds.Right - 1, Bounds.Bottom - 1);
+			var btl = new PPos(Bounds.Left, Bounds.Top);
+			var bbr = new PPos(Bounds.Right - 1, Bounds.Bottom - 1);
 			SetBounds(btl, bbr);
 
 			CustomTerrain = new CellLayer<byte>(this);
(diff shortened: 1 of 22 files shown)


pchote:
I won't try and force it, but I think we should aim to get this in for the next release.  We now have several community mods building off TS, and this would be a big benefit to them.

pchote:
I decided to include the proper projection code, as "shroud versus map edge helpers" is a better split for testing than "plumbing versus fiddly calculations".  Updated the original description to match.

MicroJOo:
exception.log
http://pastebin.com/td0XV5z6
graphics.log
http://pastebin.com/Tf51kKyP

pchote:
Fixed the above crash, and we gave this a stress-test with an awesome 2v2 match (which ended prematurely thanks to #8789).

Replay is available at https://www.dropbox.com/s/sli7p8iec6d5btc/OpenRA-pr8781-33e6f4e.orarep?dl=0 if anyone wants to hunt for bugs in it.

pchote:
> Expected behaviour with this PR applied is for there to be no visible changes in TD/RA/D2K.

This isn't quite true: aircraft visibility in these mods is now centered on the aircraft itself, instead of the ground below it.  When we originally discussed this change (months ago) this was considered a good thing.

GraionDilach:
>aircraft visibility in these mods is now centered on the aircraft itself, instead of the ground below it

You're a hero.

penev92:
Just putting it out there - that aircraft visibility sucks.

GraionDilach:
It always did. I guess we can amend the game later anyway.

To me, that falls into the same basket as the Mammoth MK2 having a working Burst: 2 on it's railgun/ Devils Tongue has a fixed rate of fire on the flamethrower - while yes, according to the INI, these were intended, but due to how the engine works, any unit which fires a particle-based weapon in TS/RA2 (railgun/flamesare) cannot refire that weapon until the previous instance still has particles of it (railgun spiral effect, fire streams), leading to DT having a slower rate of fire on bigger ranges and MMkII railgun never bursting.

OpenRA.Game/Map/Map.cs
```diff
@@ -695,6 +785,34 @@ public CPos CellContaining(WPos pos)
...
+			if (MaximumTerrainHeight == 0)
+				return new PPos[] { (PPos)uv };
+
+			if (!cellProjection.Contains(uv))
+				return NoProjectedCells;
+
+			return cellProjection[uv];
+		}
+
+		public MPos[] Unproject(PPos puv)
```
RoosterDragon: This is currently unused, will you be needing it later?

OpenRA.Game/Map/ProjectedCellRegion.cs
```diff
@@ -0,0 +1,195 @@
...
+				u = r.TopLeft.U - 1;
+				v = r.TopLeft.V;
+			}
+
+			public PPos Current { get { return current; } }
+			object IEnumerator.Current { get { return Current; } }
+			public void Dispose() { }
+		}
+
+		public struct MapCoordsRegion : IEnumerable<MPos>
```
RoosterDragon: We should probably extract a common version of this.

OpenRA.Game/Traits/World/Shroud.cs
```diff
@@ -262,25 +270,26 @@ public bool IsExplored(MPos uv)
 			if (!map.Contains(uv))
 				return false;
 
+			return map.ProjectToShroudPlane(uv).Any(IsExplored);
```
RoosterDragon: Cache this delegate please.

OpenRA.Game/Traits/World/Shroud.cs
```diff
@@ -294,39 +303,40 @@ bool IsExploredCore(MPos uv)
...
+			return IsVisible(cell.ToMPos(map));
 		}
 
 		public bool IsVisible(MPos uv)
 		{
-			if (!map.Contains(uv))
+			if (!visibleCount.Contains(uv))
 				return false;
 
+			return map.ProjectToShroudPlane(uv).Any(IsVisible);
```
RoosterDragon: Cache this delegate please.

OpenRA.Game/Map/Map.cs
```diff
@@ -652,7 +737,12 @@ public bool Contains(CPos cell)
 
 		public bool Contains(MPos uv)
 		{
-			return Bounds.Contains(uv.U, uv.V);
+			return Bounds.Contains(uv.U, uv.V) && ProjectedCellsCovering(uv).All(Contains);
```
RoosterDragon: Cache this delegate or de-LINQ this please.

OpenRA.Game/Map/Map.cs
```diff
@@ -428,6 +430,85 @@ void PostInit()
...
+
+			foreach (var puv in projected)
+				inverseCellProjection[(MPos)puv].Add(uv);
+		}
+
+		public IEnumerable<PPos> ProjectToShroudPlane(MPos uv)
+		{
+			var mapHeight = MapHeight.Value;
+			if (!mapHeight.Contains(uv))
+				yield break;
```
RoosterDragon: Avoid the compiler generated enumerable and return Enumerable.Empty<PPos>() on this fast path.

OpenRA.Game/Map/Map.cs
```diff
@@ -428,6 +430,85 @@ void PostInit()
...
+
+			var height = mapHeight[uv];
+			if (height == 0)
+			{
+				yield return (PPos)uv;
+				yield break;
```
RoosterDragon: Avoid the compiler generated enumerable and return a new array on this fast path.

The rest of the method can retain the yield statements, but this improves memory allocations on flat maps significantly.

RoosterDragon:
Played a couple of test games on TS and RA and the shroud seems to work just fine on both. Awesome. Overall it seems reasonably performant too.

I think we might do well to add some much explanatory descriptions about the difference between MPos/CPos/PPos/WPos and their regions though - we've got quite a few co-ordinates systems declared in code and if you're not already familiar with them they're pretty scary.

OpenRA.Game/Map/Map.cs
pchote: Yeah, this will be used in future PRs that resolve the `// TODO: Account for terrain height` through Map.

pchote:
Rebased, cached delegates, and fixed a major derp where the shroud was manually projecting everything instead of using the cached mapping.  These changes improve performance without affecting the ingame logic - the replay above still plays without desyncing.

pchote:
Updated to remove `MapCoordsRegion` duplication.

pchote:
Fixed.

Mailaender:
Works as promised: :white_check_mark:

pchote:
Updated to add a Contains check to Map.Unproject and correct the bounds check in Map.Contains(MPos).

pchote:
Turns out that correcting the bounds check in Map.Contains exposes massive problems in RadarWidget.  I've reverted that change (so it is now back to the same code that was :+1:'d above), but left it in its own commit for future revertability and wrote a descriptive comment.

obrakmann:
Something is wrong when you're trying to place the advanced power plant, it looks like this now, was ok before. Other buildings were ok, though, so I guess it's rather a bug in the setup of the actor than with this code.

screenshot from 2015-07-28 00 06 13

obrakmann:
Also, in the editor it looks like this now. I suppose that is expected? (Wouldn't know what else to do there when showing the area inside the cordon. Probably show it fogged, though. Either way, not sure if that should be part of this PR.)
screenshot from 2015-07-28 00 16 05

Phrohdoh:
I tested this on RA2 and noticed no difference at all, positive or negative.
I'm not sure if that is good.

pchote:
> Also, in the editor it looks like this now. I suppose that is expected? 

Yes.  This is a consequence of our fixed-size `CellLayer`, and we would need a completely new layer type just for shroud to fix this cosmetic editor issue.

pchote:
> Something is wrong when you're trying to place the advanced power plant, it looks like this now, was ok before.

I can reproduce this on bleed, so its not introduced by this pr.  It only happens on "A River Runs Near It".

obrakmann:
Shroud now works in ts, and continues to work in the regular mods. Code seems fine as far as I can be a judge of that. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/bce1ad43726735b35379e14fe47900febac9fb00%5E...bce1ad43726735b35379e14fe47900febac9fb0)

PR #8783, opened 2015-07-25
Re-enable mission replays

obrakmann:
Replays for missions have been disabled in #6442 due to the issues mentioned there.

The first of these has been fixed a while ago when FMV playback was moved into the UI.

The second is fixed by the first commit in this PR.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/ObserverShroudSelectorLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/ObserverShroudSelectorLogic.cs
@@ -19,9 +19,11 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 {
 	public class ObserverShroudSelectorLogic
 	{
+		readonly CameraOption combined, disableShroud;
+		readonly IOrderedEnumerable<IGrouping<int, CameraOption>> teams;
+		readonly bool limitViews;
+
 		CameraOption selected;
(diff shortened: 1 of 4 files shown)


pchote:
It might be a good idea to disable the statistics panel too.

OpenRA.Mods.Common/Widgets/Logic/Ingame/ObserverShroudSelectorLogic.cs
```diff
@@ -22,6 +22,7 @@ public class ObserverShroudSelectorLogic
 		CameraOption selected;
 		CameraOption combined, disableShroud;
 		IOrderedEnumerable<IGrouping<int, CameraOption>> teams;
+		bool limitViews;
```
pchote: This and maybe some of the other things can be readonly.

pchote:
Looks good and works well, so :+1: after we resolve my points above.

obrakmann:
Updated

Mailaender:
Nicely solved without introducing crashes or spoilers. :white_check_mark: This will be really useful.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a2d37ed05509576c3b152efe5da8dfb51a837d7e...0a730c243e41a1799d50f1955b40e2f96f7d1eed)

PR #7904, opened 2015-04-08
Add TerrainHeightAt to Map.

phrohdoh:
Adds height-awareness to Cargo and some other traits that shouldn't use `Z == 0`, like we currently do.

To test this try un/loading some passengers on elevated terrain in TS.

PR #8784, opened 2015-07-25
Added helicopter takeoff and landing sound

Mailaender:
Used for Tiberian Sun dropships.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Air/HeliFly.cs
+++ b/OpenRA.Mods.Common/Activities/Air/HeliFly.cs
@@ -21,6 +21,7 @@ namespace OpenRA.Mods.Common.Activities
 		readonly Target target;
 		readonly WDist maxRange;
 		readonly WDist minRange;
+		bool playedSound;
 
 		public HeliFly(Actor self, Target t)
 		{
@@ -52,6 +53,12 @@ namespace OpenRA.Mods.Common.Activities
 		{
 			if (IsCanceled || !target.IsValidFor(self))
 				return NextActivity;
+
+			if (!playedSound && helicopter.Info.TakeoffSound != null && self.IsAtGroundLevel())
+			{
+				Sound.Play(helicopter.Info.TakeoffSound);
+				playedSound = true;
+			}
 
 			if (AdjustAltitude(self, helicopter, helicopter.Info.CruiseAltitude))
 				return this;
--- a/mods/ts/rules/aircraft.yaml
+++ b/mods/ts/rules/aircraft.yaml
@@ -144,6 +144,11 @@ ORCATRAN:
 		Cost: 1200
 	Tooltip:
 		Name: Orca Transport
+	Buildable:
+		Queue: Air
+		BuildPaletteOrder: 10
+		Prerequisites: ~disabled
+	RenderSprites:
 	Helicopter:
 		LandWhenIdle: yes
 		ROT: 5
--- a/mods/ts/rules/aircraft.yaml
+++ b/mods/ts/rules/aircraft.yaml
@@ -45,6 +45,8 @@ DSHP:
 		Speed: 168
 		InitialFacing: 0
 		LandableTerrainTypes: Clear
+		TakeoffSound: dropup1.aud
+		LandingSound: dropdwn1.aud
 	Health:
 		HP: 200
 	Armor:
@@ -155,6 +157,8 @@ ORCATRAN:
 		Speed: 84
 		InitialFacing: 0
 		LandableTerrainTypes: Clear
+		TakeoffSound: dropup1.aud
+		LandingSound: dropdwn1.aud
 	Health:
 		HP: 200
 	Armor:
@@ -186,6 +190,9 @@ TRNSPORT:
 		Speed: 149
 		InitialFacing: 0
 		LandableTerrainTypes: Clear
+		TakeoffSound: dropup1.aud
+		LandingSound: dropdwn1.aud
+		AltitudeVelocity: 64
 	Health:
 		HP: 175
 	Armor:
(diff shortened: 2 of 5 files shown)


OpenRA.Mods.Common/Activities/Air/HeliFly.cs
```diff
@@ -35,6 +35,7 @@ public HeliFly(Actor self, Target t, WDist minRange, WDist maxRange)
 			this.minRange = minRange;
 		}
 
+		static int previousdz;
```
pchote: Making this static shares it over all helicopters!

OpenRA.Mods.Common/Activities/Air/HeliFly.cs
```diff
@@ -45,6 +46,18 @@ public static bool AdjustAltitude(Actor self, Helicopter helicopter, WDist targe
 			var dz = (targetAltitude.Length - altitude).Clamp(-delta, delta);
 			helicopter.SetPosition(self, helicopter.CenterPosition + new WVec(0, 0, dz));
 
+			if (dz > 0 && previousdz <= 0 && helicopter.Info.TakeoffSound != null)
```
pchote: These checks will play the take off and landing sounds for normal altitude changes as helicopters move over bumpy terrain (once #7904 or a similar PR implements proper height handling for aircraft).

OpenRA.Mods.Common/Activities/Air/HeliFly.cs
Mailaender: Woops!

pchote:
Please change this to explicitly check for takeoff and landing.  Attaching the logic to any height changes will make the sounds play when the helicopters move over bumpy terrain (after #7904 or a similar pr fixes aircraft height).

Mailaender:
I tied this together differently now. The main problem is that helicopters never properly fire the TakeOff activity. Not sure why.

OpenRA.Mods.Common/Activities/Air/HeliFly.cs
```diff
@@ -53,6 +54,12 @@ public override Activity Tick(Actor self)
 			if (IsCanceled || !target.IsValidFor(self))
 				return NextActivity;
 
+			if (!playedSound && helicopter.Info.TakeoffSound != null && helicopter.CenterPosition.Z == helicopter.Info.LandAltitude.Length)
```
penev92: This would be a lot clearer if it uses `LandAltitude.Z` instead of `LandAltitude.Length`, which is the same, because X and Y are supposed to be 0.

penev92:
I'd say we're not ready for this (the aircraft needs a lot of work before this can work properly), but maybe we choose the "complete solution to this and related issues" path way too often.
This adds a nice effect (although incomplete) and will serve as a reminder so these don't get forgotten.
I'd also say this should go to `Aircraft` (you can still either expose its info or cast the helicopter's info), but I don't think we have airplane use-cases for now, so meh.

In summary: I'm not a huge fan of the way this is implemented, but I don't see an alternative without asking you to refactor the entire aircraft code. So I'll go with "better have it and polish later" this time :+1:

OpenRA.Mods.Common/Activities/Air/HeliFly.cs
Mailaender: `LandAltitude` is `WDist` (1D so there is no X, Y or Z)

Mailaender:
Your suggestions are very vague, but I already tried those (sticking it to `Aircraft` or `Helicopter` and other methods). The places I chose in the end seem to be the best in the current code base. Otherwise the sound gets fired too often (or needs boolean guards to suppress it) or it doesn't properly triggered.

abcdefg30:
> 14:27 (abcdefg30) i mean, when you produce an aircraft, should the takeoff sound be played?
14:31 (+antares79) if it takes off, then sure

The takeoff sound is also played when loading units off the orca transport.

mods/ts/rules/aircraft.yaml
```diff
@@ -144,12 +146,19 @@ ORCATRAN:
 		Cost: 1200
 	Tooltip:
 		Name: Orca Transport
+	Buildable:
```
pchote: This is unrelated to takeoff and landing sounds, so needs to be at least in its own commit.

Mailaender: Separated.

abcdefg30:
The bug with unloading is gone now, but no sound is played when units take off after being produced.

penev92:
OK so I have `helicopter.CenterPosition.Z == helicopter.Info.LandAltitude.Length` -> `1024 == 0`, which is `False` for obvious reasons, so the sound isn't played.
Perhaps using the newly-introduced (today) `IsAtGroundLevel()` instead would fix that.

Mailaender:
Updated to make use of the new function from https://github.com/OpenRA/OpenRA/pull/8866.

pchote:
That may not have been the best idea because aircraft remain completely incompatible with terrain height.  I'll be surprised if `IsAtGroundLevel` works for them.

penev92:
IMO this PR has done all it can with the current aircraft code. The sounds are played when produced (once, ofc) and when landing (every time). This is good enough for me until we have a fix for the aircraft's apparently broken `self.CenterPosition.Z`.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/2fb00570c781e5d07b3ef0892576ca418f3528a9...744371d0da5c33c9ca4c5f6fe5e0a204d414b0cf)

PR #8785, opened 2015-07-25
Singleton Traits - Implicit & Interface

atlimit8:
Requires #8848.
This is how I discovered the issue addressed by #8774.

Allows for singleton traits, that coordinate or otherwise manage other traits, such as `UpgradeManager` to be included automatically and allows checking for inclusion of multiple traits implementing the same singleton interface (usually accessed by `.Trait[OrDefault]` or `.Traits.Get[OrDefault]`).
Is part of my effort to improve inter-trait handling (#8786).

### DIFF
--- a/OpenRA.Game/GameRules/ActorInfo.cs
+++ b/OpenRA.Game/GameRules/ActorInfo.cs
@@ -22,6 +22,13 @@ namespace OpenRA
 			: base("Multiple occurances of singleton trait "
 				+ info.Name.Substring(0, info.Name.Length - 4)
 				+ " when there only can be one.") { }
+
+		public MultipleOfSingletonTraitException(Type iface, string actor, Type first, Type next)
+			: base("Multiple occurances of singleton trait interface "
+				+ iface.Name.Substring(0, iface.Name.Length - 4) + " including "
+				+ first.Name.Substring(0, first.Name.Length - 4) + " & "
+				+ next.Name.Substring(0, next.Name.Length - 4)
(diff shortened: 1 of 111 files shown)


pchote:
There's some good ideas here, but i'm not convinced about the specific abstraction that you're proposing in #8786.

I believe a cleaner appropach would be to start by splitting `Requires` into two interfaces that handle its two (actually quite seperable) jobs:
*  `Requires<>` (keeping the same name) - used when this trait doesn't make sense without the named trait.  Used primarily by lint tests to check for bogus rule setups and docs generation.  Does not affect the trait load order.
*  `InitializeAfter<>` - used when this trait accesses another trait (if it is available) from the constructor.  Affects the load order, and can be used together with `Requires<>` if a trait is needed from a constructor, or by itself if the dependency is optional.  There are a bunch of cases where we currently use Lazy to work around the lack of this capability which could be cleaned up.

Once this is in place, we can then add (in a second PR) `RequiresSingleton<>` and `Uses<>` and cover all of your use cases with a smaller number of clearly named tags.

Would you be willing to do these changes as a prerequisite PR first?

atlimit8:
@pchote: Started.

atlimit8:
I'll just scavenge from my `LoadAfterIfAvailableAndRequires` branch and then rebase.

atlimit8:
FYI: that won't handle mutually cached traits; that is why I created this design (#8786).

penev92:
Depends on #8848 now.

atlimit8:
WIP

PR #8792, opened 2015-07-26
Changes shroud lobby setting from disabling shroud to revealing it.

### LINKED ISSUE #6123: RA1 Gap technology fails to work when Shroud is disabled
What I mean is that if Shroud is turned off in the game settings (most of my friends like to play with it off because seeing the terrain is a huge thing if you don't remember the map), then units/buildings like the Gap Generator or the vehicle that does the same thing don't work at all. We tried and it basically did nothing.

Could this be fixed in any way so that the Allies don't lose 2 units when one game setting is turned off? :O

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3426756-ra1-gap-technology-fails-to-work-when-shroud-is-disabled?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

deniz1a:
Disabling shroud in lobby will reveal shroud at game start. It does the same thing as clear shroud debug option. But shroud as a gameplay element will still be available. This means that gap generators will work.

Fixes #6123.

### DIFF
--- a/OpenRA.Game/Traits/World/Shroud.cs
+++ b/OpenRA.Game/Traits/World/Shroud.cs
@@ -282,7 +282,7 @@ namespace OpenRA.Traits
 			return explored.Contains(uv) && explored[uv] && (generatedShroudCount[uv] == 0 || visibleCount[uv] > 0);
 		}
 
-		public bool ShroudEnabled { get { return !Disabled && self.World.LobbyInfo.GlobalSettings.Shroud; } }
+		public bool ShroudEnabled { get { return !Disabled; } }
 
 		/// <summary>
 		/// Returns a fast exploration lookup that skips the usual validation.
--- a/OpenRA.Game/World.cs
+++ b/OpenRA.Game/World.cs
@@ -180,6 +180,10 @@ namespace OpenRA
 				MapUid = Map.Uid,
 				MapTitle = Map.Title
 			};
+
+			if (!LobbyInfo.GlobalSettings.Shroud)
+				foreach (var player in Players)
+					player.Shroud.ExploreAll(this);
 		}
 
 		public void LoadComplete(WorldRenderer wr)
--- a/OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
+++ b/OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
@@ -440,8 +440,8 @@ namespace OpenRA.Mods.Common.Server
 
 						bool.TryParse(s, out server.LobbyInfo.GlobalSettings.Shroud);
 						server.SyncLobbyGlobalSettings();
-						server.SendMessage("{0} {1} Shroud."
-							.F(client.Name, server.LobbyInfo.GlobalSettings.Shroud ? "enabled" : "disabled"));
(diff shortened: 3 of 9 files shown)


ScottNZ:
This change means 'Disable shroud' won't actually disable the shroud. It's still enabled, it just needs to be generated. How about calling it 'Start with explored map' (but maybe worded more concisely)?

deniz1a:
You mean rename the lobby option text? Currently it only says Shroud.

deniz1a:
This isn't about the Disable shroud debug option. This only changes the lobby option Shroud.

ScottNZ:
I'm referring to the lobby option.

deniz1a:
Ok, should I change it from Shroud to Reveal shroud? This does the same thing as the Clear shroud debug option, so maybe calling the lobby setting Clear shroud would be better. Or simply keeping it as it is would also be fine I think.

deniz1a:
"Explored map" sounds better actually. Should I change it?

chrisforbes:
Yes please.

deniz1a:
Renamed to Explored map. Maps will need upgrading I think, because this also changes MapOptions.

pchote:
Can we please keep the map interface as it was?  Introducing map format changes changes this from a relatively straightforward gameplay fix to a major and potentially risky change.

deniz1a:
Reverted renaming in MapOptions.

deniz1a:
Ok, I reverted all non-UI renaming.

Mailaender:
Works as promised :white_check_mark: and looks like an elegant fix for #6123.

pchote:
Also fixes #8827.

deniz1a:
Yes, I tested it and this also fixes map edges. I don't know why though heh.

pchote:
Please lets diagnose the actual cause of #8827 before we merge this then.
If you don't know why this fixes it then there's a significant chance that you're just fixing one of the symptoms and sweeping the real bug under the carpet.

deniz1a:
This wasn't intended to fix that bug. I didn't even know about it. I opened this PR just to make gap gens work when shroud is off. So this PR fixes #6123. Can fixing #8827 be done in another PR if this doesn't already fix it?

Unit158:
It will be, but this PR will have to wait.

Mailaender:
I guess https://github.com/OpenRA/OpenRA/pull/8920 needs to go in first then if you insist.

Unit158:
@Mailaender The fix wasn't discovered when I posted this six hours ago. If the fix has been found, I see no reason to delay this further :)

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d1bfc0a0c494c3913003474636ef00a6baf14899...f3a6b6a6b023fa3622fef222664b315a41ae0ad0)

PR #8793, opened 2015-07-26
Remove unnecessary Lazy<>s from Sandworm.cs

atlimit8:
Without looking at the commit history, I don't know why they're there. They were probably added before `Requires<MobileInfo>, Requires<WithSpriteBodyInfo>, Requires<AttackBaseInfo>` was.


Mailaender:
Yeah strange, they are superfluous as they are initialized in the constructor. :+1:

pchote:
:+1:

PR #7510, opened 2015-02-20
Added uninstaller check for running game

MorMund:
The uninstaller now checks if the game or editor is running, fixes #7142.
This requires a new [NSIS plugin(NsProcess)](http://nsis.sourceforge.net/NsProcess_plugin).
I also fixed a dll not getting removed by the uninstaller.

PR #7311, opened 2015-01-10
Switched to the new the Docker-based Travis CI infrastructure

Mailaender:
This has been suggested in https://github.com/travis-ci/travis-ci/issues/3077#issuecomment-69458787 to get timeouts under control.

PR #8796, opened 2015-07-26
Enabled the container based Travis CI architecture

Mailaender:
After some back https://github.com/OpenRA/OpenRA/pull/7311 and forth https://github.com/OpenRA/OpenRA/pull/8187 we finally need to upgrade. I had to revert https://github.com/OpenRA/OpenRA/pull/7510 because that is not available nor install-able in the new @Travis-CI environment. http://docs.travis-ci.com/user/migrating-from-legacy

### DIFF
--- a/packaging/windows/buildpackage.sh
+++ b/packaging/windows/buildpackage.sh
@@ -8,11 +8,6 @@ OUTPUTDIR="$4"
 if [ -x /usr/bin/makensis ]; then
     pushd "$SRCDIR" >/dev/null
     popd >/dev/null
-    if [[ ! -f /usr/share/nsis/Include/nsProcess.nsh &&  ! -f /usr/share/nsis/Plugin/nsProcess.dll ]]; then
-        echo "Installing NsProcess NSIS plugin in /usr/share/nsis"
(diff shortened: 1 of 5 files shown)


packaging/windows/buildpackage.sh
```diff
@@ -8,11 +8,6 @@ OUTPUTDIR="$4"
 if [ -x /usr/bin/makensis ]; then
     pushd "$SRCDIR" >/dev/null
     popd >/dev/null
```
obrakmann: These two lines can go away. I probably messed up there when copying&pasting.

obrakmann:
Local builds still work fine, :+1:

packaging/windows/buildpackage.sh
Mailaender: Agreed. Removed.

pchote:
Have we tested that the nsis packages generated by this still install and uninstall properly?

reaperrr:
> Have we tested that the nsis packages generated by this still install and uninstall properly?

They do, under Win7x64 at least.

reaperrr:
:+1:

Issue #6353, opened 2014-08-27
Different Notifications Narrator/Sounds for different teams


scskwirl:
There is only one narrator for any faction in any mod, with notifications.yaml only accepting one .aud file at a time.

What about using the same suffix effect in the voices.yaml for narrating too? Make it sound like each faction's interface has different narration and sound effects.

Phrohdoh:
@scskwirl does [the way RA2 does it](https://github.com/Phrohdoh/oramod-ra2/blob/master/notifications.yaml#L1-L12) resolve this issue?

scskwirl:
Totally forgot about this thread. lol

Looks like it does. I didn't test it yet.

But yes, have a prefix string that then points to which set of sound effects to use. So if I wanted to, I could have different voices for each specific team. That looks like it here.

pchote:
I think it would be a good idea to rework our audio definition parsing to match our new sequence parsing (#7631).  Each mod can then define its own way to optimally resolve the filename for each notification type, without conflicting with other mods.

PR #8799, opened 2015-07-26
Added support to specify the package/folder in the virtual file system for Tiberian Sun Cabal/EVA voices and Nod/GDI icons

### LINKED ISSUE #8803: Remap TS refinery/silo icon color to player color
Tiberian Sun uses the same production icon with different "player" color (Yellow for GDI, Red for Nod) for several structures ([source](http://xhp.xwis.net/ts_icons/)):
 - Refinery
 - Silo
 - EMP Cannon

Since that is a bit restrictive, and we have the technology, I propose we remap those to the actual color the player is.
Those aren't building decoration colors. Those parts of the buildings are remapped to the player color. So why not remap them on the icons as well?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24986082-remap-ts-refinery-silo-icon-color-to-player-color?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/6353. 
Closes #8803. See http://xhp.xwis.net/ts_icons/ for reference.

### DIFF
--- a/OpenRA.Game/FileSystem/GlobalFileSystem.cs
+++ b/OpenRA.Game/FileSystem/GlobalFileSystem.cs
@@ -185,10 +185,23 @@ namespace OpenRA.FileSystem
 			return s;
 		}
 
-		public static bool TryOpen(string filename, out Stream s)
+		public static bool TryOpen(string name, out Stream s)
 		{
-			// Check the cache for a quick lookup
-			if (filename.IndexOfAny(new char[] { '/', '\\' }) == -1)
+			var filename = name;
+			var foldername = string.Empty;
+
+			// Used for faction specific packages; rule out false positive on Windows C:\ drive notation
+			var explicitFolder = name.Contains(':') && !Directory.Exists(Path.GetDirectoryName(name));
+			if (explicitFolder)
+			{
+				var divide = name.Split(':');
+				foldername = divide.First();
+				filename = divide.Last();
+			}
+
+			// Check the cache for a quick lookup if the folder name is unknown
+			// TODO: This disables caching for explicit folder requests
+			if (filename.IndexOfAny(new char[] { '/', '\\' }) == -1 && !explicitFolder)
 			{
 				s = GetFromCache(PackageHashType.Classic, filename);
 				if (s != null)
@@ -200,9 +213,11 @@ namespace OpenRA.FileSystem
 			}
 
 			// Ask each package individually
-			var folder = MountedFolders
-				.Where(x => x.Exists(filename))
-				.MaxByOrDefault(x => x.Priority);
+			IFolder folder;
+			if (explicitFolder && !string.IsNullOrEmpty(foldername))
+				folder = MountedFolders.Where(x => x.Name == foldername).MaxByOrDefault(x => x.Priority);
+			else
+				folder = MountedFolders.Where(x => x.Exists(filename)).MaxByOrDefault(x => x.Priority);
 
 			if (folder != null)
 			{
@@ -214,7 +229,19 @@ namespace OpenRA.FileSystem
 			return false;
 		}
 
-		public static bool Exists(string filename) { return MountedFolders.Any(f => f.Exists(filename)); }
+		public static bool Exists(string name)
+		{
+			var explicitFolder = name.Contains(':') && !Directory.Exists(Path.GetDirectoryName(name));
+			if (explicitFolder)
+			{
+				var divide = name.Split(':');
+				var foldername = divide.First();
+				var filename = divide.Last();
+				return MountedFolders.Where(n => n.Name == foldername).Any(f => f.Exists(filename));
+			}
+			else
+				return MountedFolders.Any(f => f.Exists(name));
+		}
 
 		static Dictionary<string, Assembly> assemblyCache = new Dictionary<string, Assembly>();
 
--- a/mods/ts/sequences/infantry.yaml
+++ b/mods/ts/sequences/infantry.yaml
@@ -1,5 +1,5 @@
-e1:
-	Defaults:
+e1.gdi:
+	Defaults: e1
 		Tick: 80
 	stand:
 		Facings: 8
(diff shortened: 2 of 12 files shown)


OpenRA.Mods.Cnc/CncLoadScreen.cs
```diff
@@ -41,7 +43,9 @@ public override void Init(Manifest m, Dictionary<string, string> info)
 			r = Game.Renderer;
 			if (r == null) return;
 
-			sheet = new Sheet(Platform.ResolvePath(loadInfo["Image"]));
+			var image = Platform.ResolvePath(info["Image"]);
+			GlobalFileSystem.Mount(Path.GetDirectoryName(image));
```
pchote: This looks like a workaround / bandaid for something bogus.  Why are all these additional mounts needed?

Mailaender: It is absolute vs. relative path inconsistencies. The game loads and labels a relative one, but the code here wants an absolute one. Never mattered as we simply searched through the whole virtual file system for the one file. This adds redundancy, but saves lookup time.

pchote: Changing global state as a side-effect of this otherwise unrelated code is really bad practice, so we'll need to find a better way of solving this problem.

Mailaender:
I went for a package:filename separator that doesn't clash with real folders anymore now.

Phrohdoh:
This works nicely, though I'd like a chance to write code for sequence merging to avoid duplication. :+1:

mods/ts/sequences/vehicles.yaml
```diff
@@ -1,13 +1,21 @@
-mcv:
-	icon: mcvicon
+mcv.gdi:
+	icon: sidec01.mix/mcvicon
```
GraionDilach: You forgot to update this with your new system?

Mailaender: Yep, well spotted. Fixed.

mods/ts/sequences/infantry.yaml
```diff
@@ -1,37 +1,37 @@
-e1:
+e1.gdi:
 	Defaults:
 		Tick: 80
-	stand:
+	stand: e1
```
GraionDilach: Why not specifying this -and other infantry filenames - on the Defaults instead of the current override on every subsequence?

OpenRA.Game/FileSystem/GlobalFileSystem.cs
```diff
@@ -185,10 +185,23 @@ public static Stream Open(string filename)
...
+			{
+				var divide = name.Split(':');
+				foldername = divide.First();
+				filename = divide.Last();
+			}
+			else
+				filename = name;
+
+			// Check the cache for a quick lookup if the folder name is unknown
+			if (filename.IndexOfAny(new char[] { '/', '\\' }) == -1 && string.IsNullOrEmpty(foldername))
```
pchote: This forces these faction-specific sprites to take the slow path, which kind of sucks.
The whole filesystem code needs a rewrite, so I won't block the pr over that, but please add a `// TODO` about this.

mods/ts/sequences/infantry.yaml
pchote: I agree. Can you please fix this?

Mailaender:
Updated.

Phrohdoh:
:+1: works for me. Let's test it in a game?

OpenRA.Game/FileSystem/GlobalFileSystem.cs
```diff
@@ -185,10 +185,23 @@ public static Stream Open(string filename)
...
+			var explicitFolder = name.Contains(':') && !Directory.Exists(Path.GetDirectoryName(name));
+			if (explicitFolder)
+			{
+				var divide = name.Split(':');
+				foldername = divide.First();
+				filename = divide.Last();
+			}
+
+			// Check the cache for a quick lookup if the folder name is unknown
+			// TODO: explicit folders will always take the slow path
```
pchote: I suggest: `// TODO: This disables caching for explicit folder requests` to make this more clear.

OpenRA.Game/FileSystem/GlobalFileSystem.cs
```diff
@@ -214,7 +229,19 @@ public static bool TryOpen(string filename, out Stream s)
 			return false;
 		}
 
-		public static bool Exists(string filename) { return MountedFolders.Any(f => f.Exists(filename)); }
+		public static bool Exists(string name)
+		{
+			var explicitFolder = name.Contains(':');
```
pchote: This will need the same windows safety check as above.

Mailaender: Right. Fixed.

OpenRA.Game/FileSystem/GlobalFileSystem.cs
penev92: :+1:

Mailaender: I replaced the `string.IsNullOrEmpty(folder)` with `if (explicitFolder)` which should be faster and essentially say that in code.

pchote: The comment refers to skipping the `GetFromCache` calls inside this block, not anything to do the evaluation of the conditions.

Mailaender: Updated.

pchote:
Code changes look fine to me, and this is a great polish improvement ingame :+1:

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/5eb464bce4f4040ac6382d066df97d2f649eb97b...9d442967474fbe1c07617b9d67ac85f08992dddc)

PR #8801, opened 2015-07-26
Removes attack cursor on disguised spies.

deniz1a:
The attack cursor will not be shown on enemy spies disguised as own unit. Force firing is required to attack them. I also want to make them selectable like player's own units but couldn't find the code where it's determined.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Heal.cs
+++ b/OpenRA.Mods.Common/Activities/Heal.cs
@@ -21,7 +21,13 @@ namespace OpenRA.Mods.Common.Activities
 
 		protected override Activity InnerTick(Actor self, AttackBase attack)
 		{
-			if (!Target.IsValidFor(self) || !self.Owner.IsAlliedWith(Target.Actor.Owner))
+			if (!Target.IsValidFor(self))
+				return NextActivity;
+
+			var disguise = Target.Actor.EffectiveOwner;
+			var targetOwner = disguise != null && disguise.Disguised && !Target.Actor.Owner.IsAlliedWith(self.Owner) ?
+				disguise.Owner : Target.Actor.Owner;
+			if (!self.Owner.IsAlliedWith(targetOwner))
 				return NextActivity;
 
 			if (Target.Type == TargetType.Actor && Target.Actor.GetDamageState() == DamageState.Undamaged)
--- a/OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
@@ -246,8 +246,17 @@ namespace OpenRA.Mods.Common.Traits
 
 				var targetableRelationship = negativeDamage ? Stance.Ally : Stance.Enemy;
 
-				var owner = target.Type == TargetType.FrozenActor ? target.FrozenActor.Owner : target.Actor.Owner;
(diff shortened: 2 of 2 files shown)


deniz1a:
Actually this does more. Medics also have heal cursor on disguised spies now. But they don't heal.

obrakmann:
I really don't like special-casing spies all over the code base. :/

deniz1a:
How could it be done otherwise?

deniz1a:
Could this be added to Actor.Owner?

deniz1a:
Or add a new property to Actor called EffectiveOwner?

obrakmann:
My opinion on the matter:
```
(21:24:09) antares79: deniz1a: I don't really think spies need to be perfect copies of the units they're impersonating
(21:24:39) antares79: so if there are little things like a cursor acting different over a spy than your own unit, then so be it
(21:26:50) antares79: especially the perf overhead, while negligible on its own, is often required in places that are sometimes even called multiple times per tick, but would only be used in really rare edge cases. it's just not worth it
```

deniz1a:
OK. I changed the code to use ```Actor.EffectiveOwner```. I don't think it can be simplified more than this.

So, would this cause noticable performance decrease? Spies are often used in multiplayer games, I don't think this is an edge case. If this can be implemented in a way that doesn't impact performance, I could change it to that. But other than that, I'm just making a feature suggestion by opening a PR. Of course it is up to the maintainers to decide if it is merged or not... So if this is not wanted, the PR can be closed.

penev92:
>Spies are often used in multiplayer games, I don't think this is an edge case.

Spies are *sometimes* used when playing **1** of **N** games (mods). Don't forget that.

Also `Heal` really has no business knowing about disguise IMO.

GraionDilach:
Noticing spies via the attack cursor is a valid and legit tactic imo - considering that Allies only have spies right now to counter them.

deniz1a:
I thought when a spy is disguised, it should behave like that unit. And the only way to detect them, aside from gameplay mechanics like using dogs, would be by checking if a unit follows orders. I think disguised enemy spies should also be selectable and play selection voice and have order cursor. But when order it, it doesn't follow. And that's how you realize that's an enemy spy.

GraionDilach:
:-1: Spies are enough useful as-is, now with the veterancy, no need for undermining counter-spy tactics like this at all.

penev92:
- The "you can reveal enemy spies with a mouse-over" issue was brought up a while back and the consensus was to leave it on account of "if someone is willing to invest the time and effort in discovering spies like that they should be able to.
 - Having more and more of the engine know about spies is just bad. While it may make some sense for `AttackBase` (still not very nice), `Heal` has no business knowing about that, as I said above.

And as for your previous question about performace:
While `Heal.InnerTick()` won't be called all that often, `AttackBase.CanTargetActor()` can literally be called thousands of times in a team match lategame scenario. Your changes don't do anything to hurt performance significantly though.

penev92:
Err, wrong button. I did mean to leave this open in case someone else wanted to weigh in.

Mailaender:
Consensus seems to be that manual spy detection by the player is not a bug, but a feature. Also these exceptions are hacked into general function and not part of the Disguise trait. That is not your fault though. It was never properly cleaned up. Only renamed to EffectiveOwner to sound generic. Also the reason it regresses frequently.

Issue #7472, opened 2015-02-14
Occupancy of enemy APCs is visible


deniz1a:
When selecting an enemy APC, its occupancy bibs are shown. These should not be visible to enemies.

pevers:
I was playing around with the OpenRA code and fixed this. I'll come with a PR very soon.

PR #8805, opened 2015-07-26
Fix not being able to select enemy units

### LINKED ISSUE #8802: Selection-Box doesnt work on enemy units
selection box doesnt show up anymore when doing a left-click/rubberband-selection on enemy units 
something is broken here ... (reproducible only with fog/shroud "ON") 

http://logs.openra.net/?year=2015&month=06&day=29#16:13:38

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24985820-selection-box-doesnt-work-on-enemy-units?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
Closes #8802.

I liked it, though :/

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
+++ b/OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
@@ -71,7 +71,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr)
 		{
-			if (!self.Owner.IsAlliedWith(self.World.RenderPlayer) || self.World.FogObscures(self))
+			if (self.World.FogObscures(self))
 				yield break;
 
 			if (Info.RenderSelectionBox)
@@ -79,6 +79,9 @@ namespace OpenRA.Mods.Common.Traits
 
 			if (Info.RenderSelectionBars)
 				yield return new SelectionBarsRenderable(self);
+
+			if (!self.Owner.IsAlliedWith(wr.World.RenderPlayer))
+				yield break;
 
 			if (self.World.LocalPlayer != null && self.World.LocalPlayer.PlayerActor.Trait<DeveloperMode>().PathDebug)
 				yield return new TargetLineRenderable(ActivityTargetPath(), Color.Green);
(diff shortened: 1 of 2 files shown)


MicroJOo:
:+1: this works

Mailaender:
This reverts https://github.com/OpenRA/OpenRA/commit/329d6105773cb0f545793d3a727da818e3fff89c.

Mailaender:
This regresses so far that https://github.com/OpenRA/OpenRA/issues/7472 will be back:

image

You can see the empty pips from the APC.

pchote:
It looks like the check needs to be moved to the pip rendering/tag/other decoration rendering.

obrakmann:
Right, good catch. Fixed.

During tested I noticed that the chronoshift selection bar was also visible to enemy players. I fixed that too.

Here's a table of all selection bars and their visibility. If we want to change any of those, now would be a good opportunity. Looks all pretty sane to me now, though.

Selection bar | Visible to non-allies?
---|---
TemporaryOwnerManager | true
Chronoshiftable | false
BaseProvider | false
ExternalCapturableBar | true
AffectedByPowerOutage | true
ProductionBar | false
SupportPowerChargeBar | false
TimedUpgradeBar | false

penev92:
The only one of these that actually makes sense is `ExternalCapturableBar`. Half of the rest don't make sense unless you have been infiltrated by a spy, and the other half don't make sense at all.

GraionDilach:
`TimedUpgradeBar` should've been customizable on trait-level imo due to they can be used for debuff too - smoke grenade, tear gas, godknowswhat.
Same applies for `SupportPowerChargeBar`, having it on all makes no sense but tracking enemy super weapons (Nuke, Ion Cannon) is a must.
`TemporaryOwnerManager` is a must. The player should know when he'll get his unit back.

penev92:
That is a good set of points.

pchote:
I like the idea of making this configurable for most traits, but that's definitely out of scope for this pr.

Mailaender:
Works as promised now. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1e4c081fa0456593317fdf17ff4d4f91dd170e91...a19ba683de0d374c8fef5fe707d372872480f06c) (I assume https://github.com/OpenRA/OpenRA/issues/8802 was an unreleased regression.)

PR #8105, opened 2015-05-07
Removes LeaveMap window.

deniz1a:
So that it is possible to look at the map after game has ended. The map is frozen, it is possible to select units but not order them. I added buttons to ra, td and d2k mods.

PR #8808, opened 2015-07-27
Removes unnecessary code from IngameChatLogic.cs.

deniz1a:
Cleanup after #8105. Reverts part of #6473.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameChatLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameChatLogic.cs
@@ -37,7 +37,6 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 		bool disableTeamChat;
 		bool teamChat;
-		bool inDialog;
 
 		[ObjectCreator.UseCtor]
 		public IngameChatLogic(Widget widget, OrderManager orderManager, World world, Ruleset modRules)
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameChatLogic.cs
```diff
@@ -56,13 +55,8 @@ public IngameChatLogic(Widget widget, OrderManager orderManager, World world, Ru
...
 			var chatPanel = (ContainerWidget)widget;
 			chatOverlay = chatPanel.GetOrNull<ContainerWidget>("CHAT_OVERLAY");
-			if (chatOverlay != null)
-			{
-				chatOverlayDisplay = chatOverlay.Get<ChatDisplayWidget>("CHAT_DISPLAY");
-				chatOverlay.Visible = false;
-			}
-			else
-				inDialog = true;
+			chatOverlayDisplay = chatOverlay.Get<ChatDisplayWidget>("CHAT_DISPLAY");
```
Mailaender: `chatOverlay` may be null as it is GetOrNull before. This will trigger a Coverity warning. Either keep the null check or change GetOrNull to Get above.

deniz1a: I changed it to Get.

abcdefg30:

pchote:
Looks fine to me too. :+1:

PR #8809, opened 2015-07-27
Improved RA mod AI (yaml tweaks)

### LINKED ISSUE #8564: AI has trouble finding coarse attack location for paratroopers
I get this reproducably in the bot debug messages in RA if there's an AI with soviet paratroopers.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23145847-ai-has-trouble-finding-coarse-attack-location-for-paratroopers?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
- Removed kennels from AI.
- Tweaked Turtle AI to build more defenses and less excess adv. power plants, works better in conjunction with #8759.
- Fixed soviet paratroops target evaluation so it doesn't always fail.
- Removed obsolete sniper entry.

Fixes #8564.

### DIFF
--- a/mods/ra/rules/ai.yaml
+++ b/mods/ra/rules/ai.yaml
@@ -75,13 +75,13 @@ Player:
 				Types: Structure
 				Attractiveness: 1
 				TargetMetric: None
-				CheckRadius: 10c0
+				CheckRadius: 8c0
 			Consideration@2:
 				Against: Enemy
-				Types: Vehicle, Tank, Infantry, Defense, Water
-				Attractiveness: -10
+				Types: Water
+				Attractiveness: -5
 				TargetMetric: None
-				CheckRadius: 10c0
+				CheckRadius: 8c0
 		SupportPowerDecision@parabombs:
 			OrderName: UkraineParabombs
 			MinimumAttractiveness: 1
@@ -199,13 +199,13 @@ Player:
 				Types: Structure
 				Attractiveness: 1
 				TargetMetric: None
-				CheckRadius: 10c0
+				CheckRadius: 8c0
 			Consideration@2:
 				Against: Enemy
-				Types: Vehicle, Tank, Infantry, Defense, Water
-				Attractiveness: -10
+				Types: Water
+				Attractiveness: -5
 				TargetMetric: None
-				CheckRadius: 10c0
+				CheckRadius: 8c0
 		SupportPowerDecision@parabombs:
 			OrderName: UkraineParabombs
 			MinimumAttractiveness: 1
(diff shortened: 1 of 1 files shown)


reaperrr:
Since this doesn't touch code and is relatively easy to review/test, I'm adding this to release milestone as bugfix for (#8564).

reaperrr:
Updated. Decided to disable kennels completely for all AIs, and removed the third consideration from para troops (they still work without it).

obrakmann:
The AI still has trouble with the paratroopers. Parabombs work, though. The difference between the two is that the parabombs rule has no second (negative) consideration. I'm just testing if removing that from the paratroopers rule makes them work for the AI.

reaperrr:
We need a negative consideration to make the AI avoid larger troop concentrations and for Water in general, but we could increase the Attractiveness of structures to 2 so the check succeeds more easily.

obrakmann:
But the first consideration doesn't even target units, why does it need a negative consideration then in order to avoid troop concentrations?

I can see the point regarding water though, so that the AI doesn't paradrop onto a shipyard, but wouldn't the rule below suffice to achieve that (-5 because 5 is the MinimumAttractiveness value)? I'm not sure it's even needed, since you'd need at least four other buildings in the same 8 cell radius for the location to be considered.

```yaml
			Consideration@2:
				Against: Enemy
				Types: Water
				Attractiveness: -5
				TargetMetric: None
				CheckRadius: 8c0
```

Anyway, my test without a second consideration was successful. I didn't see a single failure, and all paradrops happened in the middle of an enemy base.

GraionDilach:
I don't really like the kennel removal, but I know suggesting a UnitLimits into HackyAI is out of scope for this PR.

obrakmann:
Eh.... I'm not sure why I added the +2 label above. I could have sworn there was comment from @Mailaender with a :+1:, but I can't find any trace of it anymore.

Anyway, now it's got my :+1:

reaperrr:
@GraionDilach yeah, a unit limit would be the best solution. Wouldn't hurt to limit AI harvesters too, for example. This PR is more a stop-gap solution in case we don't come up with something better before next release.

@obrakmann that was probably a comment on a commit and got lost because I fixup'd the changes.

Mailaender:
Filed it as https://github.com/OpenRA/OpenRA/issues/8823 so we don't forget.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/67427d9d862c2dd9a0b77ba0c420de0f85b84590...0cfe138c44851a619b8166a2a734cb223c4057dd)

PR #8811, opened 2015-07-27
Use HeadshotDeath instead of BulletDeath for Umagon's rifle

abcdefg30:
"Follow-up" of #8742, see https://github.com/OpenRA/OpenRA/pull/8742#discussion_r34966625.

### DIFF
--- a/mods/ts/weapons/smallguns.yaml
+++ b/mods/ts/weapons/smallguns.yaml
@@ -263,5 +263,4 @@ Sniper:
 			Light: 5
 			Heavy: 5
 			Concrete: 5
-		DamageTypes: Prone100Percent, TriggerProne, BulletDeath
-
+		DamageTypes: Prone100Percent, TriggerProne, HeadshotDeath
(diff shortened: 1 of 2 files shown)


mods/ts/weapons/smallguns.yaml
```diff
@@ -263,5 +263,4 @@ Sniper:
 			Light: 5
 			Heavy: 5
 			Concrete: 5
-		DamageTypes: Prone100Percent, TriggerProne, BulletDeath
-
+		DamageTypes: Prone100Percent, TriggerProne, HeadshotDeath
```
abcdefg30: Somehow I still have another line below this one locally...

GraionDilach: Windows vs Linux line ending issue?

GraionDilach:
:+1: though.

penev92:
For the record, this doesn't look like a bullet-induced death to me (type 3). More like chemically-induced.
But there isn't much point to argue about that here and now so :+1: otherwise.

PR #8812, opened 2015-07-27
Fix TS Repair weapons versus values

abcdefg30:
[Rules.ini](https://github.com/LipkeGu/TSIniFiles/blob/master/rules.ini#L8583)

### DIFF
--- a/mods/ts/weapons/healweapons.yaml
+++ b/mods/ts/weapons/healweapons.yaml
@@ -17,17 +17,13 @@ Repair:
 	ReloadDelay: 80
 	Range: 1c819
 	Report: REPAIR11.AUD
+	ValidTargets: Repair
 	Projectile: Bullet
 		Speed: 1c682
 	Warhead@1Dam: SpreadDamage
 		Spread: 213
 		Damage: -50
-		Versus:
-			None: 0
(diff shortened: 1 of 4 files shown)


GraionDilach:
We shouldn't even have `Versus` defined on that one tbh, just give it a `TargetType: Repair` á lá RA1 mechanic.

pchote:
I agree with @GraionDilach.  This is just propagating the broken legacy "0% means untargetable" logic

abcdefg30:
Updated.

mods/ts/weapons/healweapons.yaml
```diff
@@ -17,17 +17,19 @@ Repair:
...
 	Projectile: Bullet
 		Speed: 1c682
 	Warhead@1Dam: SpreadDamage
 		Spread: 213
 		Damage: -50
+		ValidTargets: Repair
 		Versus:
 			None: 0
-			Wood: 0
+			Wood: 100
```
GraionDilach: Remove Versus altogether instead of refining it. :D

abcdefg30:
Updated.

GraionDilach:
:+1:

PR #8813, opened 2015-07-27
Replace map.Contains with proper array bounds checks.

pchote:
This fixes a bunch of cases where map.Contains is used as a guard against index out of range exceptions.  These cases now use the Contains check on the CellLayer itself.


obrakmann:
Makes sense, I guess. Looks fine in-game. :+1:

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/a7c2beaf9ed67bdaa983a5119a7cef9c3dda23db...1ed009b666571b47a890aa17427b9de0ea420bdb)

PR #8817, opened 2015-07-28
Touch up a few files

penev92:
Just a small bunch of simple code touch-ups.

### DIFF
--- a/OpenRA.Game/Traits/World/ActorMap.cs
+++ b/OpenRA.Game/Traits/World/ActorMap.cs
@@ -42,18 +42,16 @@ namespace OpenRA.Traits
 
 		class CellTrigger
 		{
-			public readonly int Id;
 			public readonly CPos[] Footprint;
 			public bool Dirty;
 
-			Action<Actor> onActorEntered;
-			Action<Actor> onActorExited;
+			readonly Action<Actor> onActorEntered;
+			readonly Action<Actor> onActorExited;
 
 			IEnumerable<Actor> currentActors = Enumerable.Empty<Actor>();
 
-			public CellTrigger(int id, CPos[] footprint, Action<Actor> onActorEntered, Action<Actor> onActorExited)
+			public CellTrigger(CPos[] footprint, Action<Actor> onActorEntered, Action<Actor> onActorExited)
 			{
-				Id = id;
 				Footprint = footprint;
 
 				this.onActorEntered = onActorEntered;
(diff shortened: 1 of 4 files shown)


Phrohdoh:
Why are you removing `ProximityTrigger.Id`?

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -63,13 +61,13 @@ public CellTrigger(int id, CPos[] footprint, Action<Actor> onActorEntered, Actio
...
 
-			public void Tick(ActorMap am)
+			public void Tick(ActorMap actorMap)
 			{
 				if (!Dirty)
 					return;
 
-				var oldActors = currentActors;
-				currentActors = Footprint.SelectMany(c => am.GetUnitsAt(c)).ToList();
+				var oldActors = currentActors.ToList();
```
pchote: This forces a(n unnecessary) copy of the list.  More than just a simple touchup!

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -129,7 +123,7 @@ public void Tick(ActorMap am)
 				if (!Dirty)
 					return;
 
-				var oldActors = currentActors;
+				var oldActors = currentActors.ToList();
```
pchote: This forces a(n unnecessary) copy of the list. More than just a simple touchup!

penev92:
Reverted the `.ToList()`s.

Mailaender:
:+1:

penev92:
@Phrohdoh Because it's unused. Maybe a leftover from something old?

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -88,24 +86,20 @@ public void Tick(ActorMap am)
 
 		class ProximityTrigger : IDisposable
 		{
-			public readonly int Id;
-			public WPos Position { get; private set; }
-			public WDist Range { get; private set; }
```
pchote: This is useful info that makes sense to be public, even if nothing is using it right now.

Phrohdoh: You and @Mailaender have told me many times not to make something public unless it is being used by something.

pchote: That's a valid point I guess.

pchote:
Looks like a reasonable set of code cleanups :+1:

PR #8824, opened 2015-07-28
Implement heightmap-aware bounds clamping.

pchote:
Followup 1 / 2 of #8781.  This does two things:
- Changes the radar code to draw the full map into the backing sprite.  This is one of the final prerequisite steps for scripted map bounds changes, and dramatically simplifies the required sanity checking.
- Implements a heightmap-aware algorithm for clamping map cells to the projected bounds.

There are two easy ways to test this ingame:
- Try ordering units near and beyond the edges of the map, and observe that they now behave properly.
- Enable the terrain grid in the map editor, and compare the cursor cell with the clamped cell (white).

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -739,10 +739,10 @@ namespace OpenRA
 
 		public bool Contains(MPos uv)
 		{
-			// TODO: Checking against the bounds excludes valid parts of the map if MaxTerrainHeight > 0.
-			// Unfortunatley, doing this properly leads to memory corruption issues in the (unsafe) radar
-			// rendering code.
-			return Bounds.Contains(uv.U, uv.V) && ProjectedCellsCovering(uv).All(containsTest);
+			// The first check ensures that the cell is within the valid map region, avoiding
+			// potential crashes in deeper code.  All CellLayers have the same geometry, and
+			// CustomTerrain is convenient (cellProjection may be null and others are Lazy).
+			return CustomTerrain.Contains(uv) && ProjectedCellsCovering(uv).All(containsTest);
 		}
 
 		public bool Contains(PPos puv)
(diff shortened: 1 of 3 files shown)


abcdefg30:
```
OpenRA.Mods.Common/Traits/World/TerrainGeometryOverlay.cs:L117: [SA1005] The comment must start with a single space. To ignore this error when commenting out a line of code, begin the comment with '////' rather than '//'.
OpenRA.Mods.Common/Traits/World/TerrainGeometryOverlay.cs:L117: [SP2100] Code line must be no longer than 180 characters (now 205).
OpenRA.Mods.Common/Widgets/RadarWidget.cs:L83: [SA1508] A closing curly bracket must not be preceded by a blank line.

penev92:
>Changes the radar code to draw the full map into the backing sprite.

I see several issues with that. (Aside from "shouldn't be done")
 - Only the top and left cordons are drawn. Right and bot aren't.
 - As shown in [this screenshot](http://i.imgur.com/r4G645G.png) you can see the cordons on the radar, but not on the map (of course, this is ingame). Turns out the actor positioning doesn't account for the borders, so it shows the war factory is *inside* the cordon.
 - With scripted map size changes, the player should not be shown anything other than the current map (shouldn't even know that it can expand) - hence the "shouldn't be done" (ingame). It's very nice for the map editor, of course.

pchote:
You've completely misinterpreted the purpose of that commit and how it works.  The key words are "backing sprite" (oops - I actually mean backing *sheet*).  This is an implementation detail, and does not change what is actually rendered to the player.

Updated to fix the terrain offset and commit wording.

penev92:
Well that's what I was thinking before testing, but the bounds being drawn kind of threw me off.

penev92:
:+1:

obrakmann:
:+1:

pchote:
Thanks, but you weren't supposed to merge this until I removed the test commit!

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/1ed009b666571b47a890aa17427b9de0ea420bdb...187ff42debd453082a8bda507a8cc868d2ad6170)

PR #8836, opened 2015-07-29
Removed redundant ISync and ITick from Attack* traits

reaperrr:
`ISync` is implemented by `AttackBase`, so there's no need for other `Attack*` traits that inherit it to implement it again.
`AttackTurreted` inherits `AttackFollow`, which already implements `ITick`.


penev92:
:+1:

abcdefg30:
:+1: / :white_check_mark:

PR #8839, opened 2015-07-29
Test sync implementations

RoosterDragon:
Added tests to verify classes use both the ISync interface and the Sync attribute and don't forget one or the other when implementing syncable traits.

Fixed some traits that failed the tests.

### DIFF
--- a/OpenRA.Game/Actor.cs
+++ b/OpenRA.Game/Actor.cs
@@ -30,7 +30,7 @@ namespace OpenRA
 
 		public readonly uint ActorID;
 
-		[Sync] public Player Owner { get; set; }
+		public Player Owner { get; set; }
 
 		public bool IsInWorld { get; internal set; }
 		public bool Disposed { get; private set; }
--- a/OpenRA.Game/Traits/World/Shroud.cs
+++ b/OpenRA.Game/Traits/World/Shroud.cs
@@ -20,7 +20,7 @@ namespace OpenRA.Traits
 		public object Create(ActorInitializer init) { return new Shroud(init.Self); }
 	}
 
-	public class Shroud
+	public class Shroud : ISync
 	{
 		[Sync] public bool Disabled;
 
--- a/OpenRA.Mods.Common/Effects/GravityBomb.cs
+++ b/OpenRA.Mods.Common/Effects/GravityBomb.cs
@@ -38,7 +38,7 @@ namespace OpenRA.Mods.Common.Effects
 		public IEffect Create(ProjectileArgs args) { return new GravityBomb(this, args); }
 	}
 
-	public class GravityBomb : IEffect
+	public class GravityBomb : IEffect, ISync
 	{
 		readonly GravityBombInfo info;
 		readonly Animation anim;
--- a/OpenRA.Mods.Common/Lint/CheckSyncAnnotations.cs
+++ b/OpenRA.Mods.Common/Lint/CheckSyncAnnotations.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using OpenRA.Traits;
@@ -19,23 +20,43 @@ namespace OpenRA.Mods.Common.Lint
 	{
 		public void Run(Action<string> emitError, Action<string> emitWarning)
 		{
-			/* first, check all the types implementing ISync */
-			foreach (var t in Game.ModData.ObjectCreator.GetTypesImplementing<ISync>())
-				if (!HasAnySyncFields(t))
-					emitWarning("{0} has ISync but nothing marked with [Sync]".F(t.Name));
+			var modTypes = Game.ModData.ObjectCreator.GetTypes();
+			CheckTypesWithSyncableMembersImplementSyncInterface(modTypes, emitWarning);
+			CheckTypesImplementingSyncInterfaceHaveSyncableMembers(modTypes, emitWarning);
 		}
 
-		bool HasAnySyncFields(Type t)
+		static readonly Type SyncInterface = typeof(ISync);
+
+		static bool TypeImplementsSync(Type type)
 		{
-			var flags = BindingFlags.Public | BindingFlags.NonPublic
-				| BindingFlags.Instance;
+			return type.GetInterfaces().Contains(SyncInterface);
+		}
 
-			var fs = t.GetFields(flags);
-			var ps = t.GetProperties(flags);
+		static bool AnyTypeMemberIsSynced(Type type)
+		{
+			const BindingFlags Flags = BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly;
+			while (type != null)
+			{
+				if (((MemberInfo[])type.GetFields(Flags)).Concat(type.GetProperties(Flags)).Any(x => x.HasAttribute<SyncAttribute>()))
+					return true;
+				type = type.BaseType;
+			}
 
-			return fs.Any(f => f.HasAttribute<SyncAttribute>()) ||
-				ps.Any(p => p.HasAttribute<SyncAttribute>()) ||
-				HasAnySyncFields(t.BaseType);
+			return false;
+		}
+
+		static void CheckTypesWithSyncableMembersImplementSyncInterface(IEnumerable<Type> types, Action<string> emitWarning)
+		{
+			foreach (var type in types)
+				if (!TypeImplementsSync(type) && AnyTypeMemberIsSynced(type))
+					emitWarning("{0} has members with the Sync attribute but does not implement ISync".F(type.FullName));
+		}
+
+		static void CheckTypesImplementingSyncInterfaceHaveSyncableMembers(IEnumerable<Type> types, Action<string> emitWarning)
+		{
+			foreach (var type in types)
+				if (TypeImplementsSync(type) && !AnyTypeMemberIsSynced(type))
+					emitWarning("{0} implements ISync but does not use the Sync attribute on any members.".F(type.FullName));
 		}
 	}
 }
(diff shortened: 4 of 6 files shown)


Mailaender:
I guess you can drop https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Lint/CheckSyncAnnotations.cs then.

OpenRA.Test/OpenRA.Game/SyncImplTest.cs
```diff
@@ -0,0 +1,67 @@
...
+			return false;
+		}
+
+		[TestCase(TestName = "Types with syncable members implement sync interface")]
+		public void TypesWithSyncableMembersImplementSyncInterface()
+		{
+			foreach (var type in AllModTypes)
+				if (!TypeImplementsSync(type) && AnyTypeMemberIsSynced(type))
+					Assert.Fail("{0} has members with the Sync attribute but does not implement ISync".F(type.FullName));
+			Assert.Pass();
```
Mailaender: https://ci.appveyor.com/project/OpenRA/openra/build/1.0.5255/tests failed with an exception here.

RoosterDragon:
```
at NUnit.Framework.Assert.Pass(String message, Object[] args)
at NUnit.Framework.Assert.Pass()
at OpenRA.Test.SyncImplTest.TypesImplementingSyncInterfaceHaveSyncableMembers() in c:\projects\openra\OpenRA.Test\OpenRA.Game\SyncImplTest.cs:line 64
```

What does this mean? Does it think Assert.Pass() indicates a failure or something?

OpenRA.Test/OpenRA.Test.csproj
```diff
@@ -43,12 +43,14 @@
       <Private>False</Private>
     </Reference>
     <Reference Include="System" />
+    <Reference Include="System.Drawing" />
```
Mailaender: What do you need this for?

RoosterDragon:
Didn't know there was a lint pass.

Should I implement it as a lint pass? That would allow 3rd party mods to check themselves too right?

Mailaender:
We just didn't have NUnit support back then. 3rd party mods should also have access to unit tests.

OpenRA.Test/OpenRA.Test.csproj
RoosterDragon: For some reason it needs the reference in order to build. Otherwise I get these errors:
```
Error	CS0012	The type 'RectangleF' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.	OpenRA.Test	\OpenRA\OpenRA.Test\OpenRA.Game\SyncImplTest.cs	33
Error	CS0012	The type 'Rectangle' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.	OpenRA.Test	\OpenRA\OpenRA.Test\OpenRA.Game\SyncImplTest.cs	33
```

RoosterDragon:
Will the search for ```OpenRA.Mods.*.dll``` pick them up correctly?

OpenRA.Test/OpenRA.Test.csproj
Phrohdoh: Why are we using `RectangleF` is a sync test implementation?

RoosterDragon: ```return type.GetInterfaces().Contains(typeof(ISync))``` produces these errors (until you include the dll) - don't ask me!

A loop seems to avoid the problem - I'll use that.

RoosterDragon:
Updated:
- Also check and fix traits in OpenRA.Game.exe
- Remove the lint pass.
- Removed need to reference System.Drawing.

OpenRA.Test/OpenRA.Game/SyncImplTest.cs
```diff
@@ -0,0 +1,71 @@
...
+			return false;
+		}
+
+		[TestCase(TestName = "Types with syncable members implement sync interface")]
+		public void TypesWithSyncableMembersImplementSyncInterface()
+		{
+			foreach (var type in AllTypes)
+				if (!TypeImplementsSync(type) && AnyTypeMemberIsSynced(type))
+					Assert.Fail("{0} has members with the Sync attribute but does not implement ISync".F(type.FullName));
+			Assert.Pass();
```
Mailaender: http://www.nunit.org/index.php?p=utilityAsserts&r=2.5

> The Assert.Pass method allows you to immediately end the test, recording it as successful. Since it causes an exception to be thrown, it is more efficient to simply allow the test to return.

chrisforbes:
I like this *much* better as a lint pass, so you can apply it to a
particular set of DLLs.

Mailaender:
Why not just parametrize the DLL name?

RoosterDragon:
Perhaps I can do both (and have the test just run the lint pass or something like that).

Mailaender:
I am against a lint pass as it will run the test for each map (and this repository already has a lot of them) which is very wasteful and would wreck my current effort in reducing the Travis CI build times.

Mailaender:
These ISync annotation checks have nothing to do with MiniYaml. I added a workaround in https://github.com/OpenRA/OpenRA/pull/8936 to avoid doing duplicate checks for every map, but it just shows the insanity of using the inappropriate linter system.

OpenRA.Test/OpenRA.Game/SyncImplTest.cs
RoosterDragon: Ok, I'll just let it return then.

RoosterDragon:
OK I've rebased and fixed the issues.

I agree with @Mailaender regarding the lint stuff - so I've stuck with using a test and removing the lint pass for now.

Mailaender:
Can you resubmit again? I had to manually increment the @AppVeyor build number again.

Mailaender:
:+1:

RoosterDragon:
I got an exception: "Actor type 'camera.paradrop' does not define a quantized body orientation." whilst running a game with this PR.

I suspect this is because the ```QuantizedFacings``` property in ```BodyOrientation``` is lazy and nothing was reading it before for this actor. However trying to sync it means it is now forcibly read and thus throwing the error.

Should we avoid syncing this? Or only sync it if something already generated it?

RoosterDragon:
I have decided to keep the current behaviour for ```BodyOrientation`` and avoid attempting to sync it since it may be problematic.

This resolves the crash.

penev92:
Needs a rebase.

RoosterDragon:
Rebased, and also leave in syncing of ```BodyOrientation``` since #9193 fixed the underlying problem.

penev92:
Sorry, needs a rebase again.

penev92:
Sorry, needs a rebase again.

RoosterDragon:
Rebased.

OpenRA.Game/Actor.cs
```diff
@@ -22,7 +22,7 @@
 
 namespace OpenRA
 {
-	public sealed class Actor : IScriptBindable, IScriptNotifyBind, ILuaTableBinding, ILuaEqualityBinding, ILuaToStringBinding, IEquatable<Actor>, IDisposable
+	public sealed class Actor : IScriptBindable, IScriptNotifyBind, ILuaTableBinding, ILuaEqualityBinding, ILuaToStringBinding, ISync, IEquatable<Actor>, IDisposable
```
pchote: I'm pretty sure we don't sync-test Actors (only traits), so this isn't very useful.

OpenRA.Test/OpenRA.Game/SyncImplTest.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Reflection;
+using NUnit.Framework;
+
+namespace OpenRA.Test
+{
+	[TestFixture]
+	class SyncImplTest
+	{
+		static readonly IEnumerable<Type> AllTypes =
+			Directory.EnumerateFiles("mods", "OpenRA.Mods.*.dll", SearchOption.AllDirectories)
```
pchote: This isn't very nice: mods are free to load dlls from wherever they want (this becomes even more important after #8992).

OpenRA.Game/Actor.cs
pchote: We have a special-case path for syncing actors, but that only includes the actor ID.

OpenRA.Test/OpenRA.Game/SyncImplTest.cs
RoosterDragon: Where can I find the existing mechanism that does this loading? I'll just use that.

pchote: The linter does it per-mod, using the mod.yaml definitions, but I don't think you can use that directly here.  Perhaps enumerate over all of the installed mods and build a list of all the possible mod dlls?

pchote: This kind of trouble was why we decided to use lint passes for these things.

OpenRA.Game/Actor.cs
RoosterDragon: There is a [Sync] attribute on the player through, which seems like a reasonable thing to check. I'm kinda leaving it around as a mini-reminder to fix it eventually so we sync the actor correctly but until then I need ISync here to make the lint pass for now.

I could remove it if preferred.

RoosterDragon:
OK we're back to a lint pass. It seems to be doing the right thing but I'm not super-sure it that means I have it implemented right.

Probably needs a review from fresh.

OpenRA.Game/Actor.cs
pchote: I'd prefer it if you could remove the [Sync] attribute from the owner, since it currently has no purpose except to be be misleading.

RoosterDragon: Ok - done.

pchote:
Thanks. Looks good to me now :+1:.  Confirmed that things break if I remove ISync or add it where it isn't needed.

penev92:
:+1:

PR #8840, opened 2015-07-29
Removes spacing between dropdown menu items.

### LINKED ISSUE #8829: There is a gap between dropdown menu items
Sometimes I click on the gap between dropdown menu items and the menu closes with no action. I think there shouldn't be any gap between menu item click areas.

Clicking here closes the dropdown menu without selecting 1 or 2:
menu-gap

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25100848-there-is-a-gap-between-dropdown-menu-items?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

deniz1a:
Fixes #8829. Adds a new TopBottomSpacing variable that controls the padding at the top and the bottom of the list. ItemSpacing is now added only between two menu items and not at the beginning and at the end of a list. So it is arranged like this:

**TopBottomSpacing**
Menu Item 1
**ItemSpacing**
Menu Item 2
**ItemSpacing**
Menu Item 3
**TopBottomSpacing**

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ListLayout.cs
+++ b/OpenRA.Mods.Common/Widgets/ListLayout.cs
@@ -38,7 +38,8 @@ namespace OpenRA.Mods.Common.Widgets
 					widget.ContentHeight += w.Bounds.Height + widget.ItemSpacing;
 			}
 
-			// The loop above appended an extra widget.ItemSpacing after the last item. Replace it with proper bottom spacing.
+			// The loop above appended an extra widget.ItemSpacing after the last item.
+			// Replace it with proper bottom spacing.
 			widget.ContentHeight += widget.TopBottomSpacing - widget.ItemSpacing;
 		}
 	}
--- a/OpenRA.Mods.Common/Widgets/ListLayout.cs
+++ b/OpenRA.Mods.Common/Widgets/ListLayout.cs
@@ -21,22 +21,25 @@ namespace OpenRA.Mods.Common.Widgets
 		public void AdjustChild(Widget w)
 		{
 			if (widget.Children.Count == 0)
-				widget.ContentHeight = widget.ItemSpacing;
+				widget.ContentHeight = 2 * widget.TopBottomSpacing - widget.ItemSpacing;
 
-			w.Bounds.Y = widget.ContentHeight;
+			w.Bounds.Y = widget.ContentHeight - widget.TopBottomSpacing + widget.ItemSpacing;
 			if (!widget.CollapseHiddenChildren || w.IsVisible())
 				widget.ContentHeight += w.Bounds.Height + widget.ItemSpacing;
 		}
 
 		public void AdjustChildren()
 		{
-			widget.ContentHeight = widget.ItemSpacing;
+			widget.ContentHeight = widget.TopBottomSpacing;
 			foreach (var w in widget.Children)
 			{
 				w.Bounds.Y = widget.ContentHeight;
 				if (!widget.CollapseHiddenChildren || w.IsVisible())
 					widget.ContentHeight += w.Bounds.Height + widget.ItemSpacing;
 			}
+
+			// The loop above appended an extra widget.ItemSpacing after the last item. Replace it with proper bottom spacing.
+			widget.ContentHeight += widget.TopBottomSpacing - widget.ItemSpacing;
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Widgets/GridLayout.cs
+++ b/OpenRA.Mods.Common/Widgets/GridLayout.cs
@@ -24,14 +24,14 @@ namespace OpenRA.Mods.Common.Widgets
 		{
 			if (widget.Children.Count == 0)
 			{
-				widget.ContentHeight = widget.ItemSpacing;
-				pos = new int2(widget.ItemSpacing, widget.ItemSpacing);
+				widget.ContentHeight = 2 * widget.TopBottomSpacing;
+				pos = new int2(widget.ItemSpacing, widget.TopBottomSpacing);
 			}
 
-			if (pos.X + widget.ItemSpacing + w.Bounds.Width > widget.Bounds.Width - widget.ScrollbarWidth)
+			if (pos.X + w.Bounds.Width + widget.ItemSpacing > widget.Bounds.Width - widget.ScrollbarWidth)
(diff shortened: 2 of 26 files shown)


deniz1a:
Oh of course, the actual thing this PR does is it sets ItemSpacing to zero. So there is no dead space between menu items.

deniz1a:
And I could also remove ItemSpacing. Because there is no use for it now. Should I remove it?

Mailaender:
I fail to see the visual improvement to be honest.

deniz1a:
This is not for improving the appearance but the mouse interaction. Currently, there is a gap between dropdown menu items and when you click there, the menu closes without selecting any item. I posted a screenshot in the issue #8829. And I don't think it looks bad without the gap:
nogap

deniz1a:
This is how it currently looks:
withgap

OpenRA.Mods.Common/Widgets/ListLayout.cs
```diff
@@ -21,22 +21,24 @@ public class ListLayout : ILayout
 		public void AdjustChild(Widget w)
 		{
 			if (widget.Children.Count == 0)
-				widget.ContentHeight = widget.ItemSpacing;
+				widget.ContentHeight = 2 * widget.TopBottomSpacing - widget.ItemSpacing;
```
pchote: This think `widget.ContentHeight = widget.TopBottomSpacing` would be a more sensible choice.

OpenRA.Mods.Common/Widgets/ListLayout.cs
```diff
@@ -21,22 +21,24 @@ public class ListLayout : ILayout
...
-			widget.ContentHeight = widget.ItemSpacing;
+			widget.ContentHeight = widget.TopBottomSpacing;
 			foreach (var w in widget.Children)
 			{
 				w.Bounds.Y = widget.ContentHeight;
 				if (!widget.CollapseHiddenChildren || w.IsVisible())
 					widget.ContentHeight += w.Bounds.Height + widget.ItemSpacing;
 			}
+
+			widget.ContentHeight = widget.ContentHeight - widget.ItemSpacing + widget.TopBottomSpacing;
```
pchote: Please replace this with:
```csharp
// The loop above appended an extra widget.ItemSpacing after the last item.  Replace it with proper bottom spacing.
widget.ContentHeight += widget.TopBottomSpacing - widget.ItemSpacing;
```

OpenRA.Mods.Common/Widgets/ListLayout.cs
deniz1a: But that wouldn't do the same thing. When adding the first item, both top and bottom spacings should be added.

pchote:
A couple more issues:
* This (incorrectly) removes the padding from the top of the lobby player bin.
* There are a whole bunch of dropdowns across the game that still use item spacing.  If we're going to do this, then we need to do them all.

pchote:
This is a nice polish improvement aside from the above issues.

OpenRA.Mods.Common/Widgets/ListLayout.cs
pchote: I now see that this is a hacky workaround for the line below (so that the final y is indeed `widget.TopBottomSpacing`).

deniz1a:
Ok there are lots of places in yaml where custom values are assigned to ItemSpacing. Instead of adding a line for TopBottomSpacing in all those yaml files, should I make it a property that returns the same value as ItemSpacing if it's not zero? This would fix them all but it would not be possible to adjust TopBottomSpacing separately in yaml.

pchote:
ItemSpacing probably wants to be set to 0 in those cases.

I expect that this will indeed need a bunch of yaml tweaks - please don't be lazy by using a code workaround.

deniz1a:
I added TopBottomSpacing definitions above all ItemSpacing values in yaml. I used the following regular expression to find and replace:
Find: (\s+)ItemSpacing: (\d+)
Replace with: $1TopBottomSpacing: $2$1ItemSpacing: $2

deniz1a:
https://imgs.xkcd.com/comics/regular_expressions.png

deniz1a:
regular_expressions

pchote:
Hang on, how does that work? Changing the TopBottom spacing shouldn't have anything to do with the spacing between items?

deniz1a:
It doesn't. This just adds the same amount of TopBottomSpacing where a custom ItemSpacing is defined.

pchote:
My request was to remove the spacing between adjacent list items.  Otherwise we have inconsistencies between the lobby dropdowns that you have shown above and every other dropdown in the game.

deniz1a:
But normal dropdowns don't have custom spacings in yaml, do they? This only affects ScrollPanelWidgets that are not simple dropdown lists. Like the player bin in lobby.

deniz1a:
The default value in code is zero for ItemSpacing

pchote:
Did you change that default in the latest set of changes? (you only mentioned the TopBottomSpacing change).  There were several regular dropdowns that had non-zero spacing when I tested the earlier version of the PR.

deniz1a:
No, ItemSpacing was zero in the first commit. Which dropdowns were they?

deniz1a:
I also have to adjust the grid layout I think. So this is not yet finished.

deniz1a:
Ok I changed GridLayout to use TopBottomSpacing. Review pls.

abcdefg30:
:+1:

deniz1a:
I renamed some stuff in the constructors but now I don't know if those were part of how the general widget constructor parameters were named. It looks better with descriptive names but I can revert them if widget constructors have a standard way of naming parameters.

penev92:
Yes please revert the renaming. We use "widget" everywhere, and also I don't see "sp" being any clearer than "widget".

deniz1a:
Reverted variable renamings.

penev92:
Looks good :+1:

deniz1a:
This doesn't have a changelog entry?

PR #8848, opened 2015-07-30
Split Requires<> into Requires<> and InitializeAfter<>

atlimit8:
Requires<> - Trait needs other trait
InitializeAfter<> - Trait constructed after other trait.


OpenRA.Game/GameRules/ActorInfo.cs
```diff
@@ -125,33 +125,43 @@ public IEnumerable<ITraitInfo> TraitsInConstructOrder()
...
+				Requisites = RequisitesOf(i).ToList()
 			}).ToList();
 
-			var resolved = source.Where(s => !s.Dependencies.Any()).ToList();
+			var resolved = source.Where(s => !s.Predecessors.Any()).ToList();
 			var unresolved = source.Except(resolved);
 
 			var testResolve = new Func<Type, Type, bool>((a, b) => a == b || a.IsAssignableFrom(b));
-			var more = unresolved.Where(u => u.Dependencies.All(d => resolved.Exists(r => testResolve(d, r.Type))));
+			var more = unresolved.Where(u => u.Predecessors.All(d => !unresolved.Any(r => testResolve(d, r.Type))));
```
pchote: This is the magic line in the resolving procedure.  While you're here, could you please add a comment explaining how it works?

pchote:
I searched through all of our traits that declare at least one `Requires` and found a couple of inconsistencies:

The following traits need to declare `InitializeAfter`s:
* `WithRoof` (`RenderSpritesInfo`).
* `WithCrumbleOverlay` (`RenderSpritesInfo`)

There are a few more instances of lady / deferred loading that look like they could be replaced with `InitializeAfter`:
* `Armament` (`Turreted`, `IBodyOrientation`).
* `AttackGarrisoned` (`IBodyOrientation`).
* `AutoHeal` (`AttackBase`).
* `SupportPowerManager` (`RadarPings`)
* `Turreted` (`IFacing`, `IBodyOrientation`)

Things with `Requires` that may not need them?:
* `Carryall` (`IBodyOrientation`).
* `Crate` (`RenderSprites`).
* `ClassicProductionQueue` (`TechTree`).

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -154,7 +154,7 @@ Actor ChooseTarget(Actor self, WDist range)
 			return inRange
 				.Where(a =>
 					a.AppearsHostileTo(self) &&
-					!a.HasTrait<AutoTargetIgnore>() &&
+					!a.Info.Traits.Contains<AutoTargetIgnoreInfo>() &&
```
pchote: Why this change?

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
```diff
@@ -26,15 +27,15 @@ public class CombatDebugOverlay : IPostRender, INotifyDamage
 	{
 		readonly DeveloperMode devMode;
 
-		Lazy<AttackBase> attack;
-		Lazy<IBodyOrientation> coords;
-		Lazy<Health> health;
+		AttackBase attack;
```
pchote: `readonly` for these three please.

pchote:
There are also a set of `HasTrait<Building>()`s that will need to change to `Info.Contains<BuildingInfo>()`.

pchote:
Or perhaps we could change the definition of `HasTrait` to check the info instead?  That would fix those other cases for free, and let you change the Info.Contains back to HasTrait.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -19,7 +19,7 @@
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Attach this to a unit to enable dynamic upgrades by warheads, experience, crates, support powers, etc.")]
-	public class UpgradeManagerInfo : ITraitInfo, Requires<IUpgradableInfo>
+	public class UpgradeManagerInfo : ITraitInfo, InitializeAfter<IUpgradableInfo>
```
pchote: This might cause us dependency loops in the future (if we wanted to e.g. make `DeployToUpgrade` itself upgradeable).  How about we use `INotifyCreated` instead of the Lazy?

atlimit8: I was planning on changing some traits to use `INotifyCreated` in a separate PR, but I guess I  can do this one here.

pchote:
Looks great overall - this is something that I have wanted for a few years now.  Thanks!

Please look into the following things:
* Check my list of inconsistencies above and fix anything that is real.
* Use INotifyCreated to avoid the potential dependency loop in `UpgradeManager`.
* Fix the inconsistent use of `HasTrait` (either by removing it completely, or by making it safe to use from the constructor and using it everywhere)

atlimit8:
Requires #8927.

abcdefg30:
Needs a rebase now, too.

abcdefg30:

PR #8853, opened 2015-07-31
Expose the WorldActor to Lua via Map.WorldActor.

phrohdoh:
The only testcase for this I have is [ra2#weather2](https://github.com/Phrohdoh/OpenRA/commit/ee52c3b7c90d46ca10810b539812b27324246f53), but eventually it could be used to query (certain?) `GlobalLightingPaletteEffect`s to determine time of day, or other things.


pchote:
World being an actor is an implementation detail, and our goal has been to hide things like that from the map scripters.  Our standard so far has been to use global tables to expose things that affect the map globally.

obrakmann:
Yep, I agree. This will want to be in `EffectGlobal`.

Phrohdoh:
I was working on this... but ok.

PR #8858, opened 2015-07-31
Sync members of base classes

RoosterDragon:
The method that generates sync hashing functions for traits was missing out non-public members in base classes of that trait. This meant the hash function might not be syncing over all the members you had tagged.

Now, if a trait has a base class with syncable members, we ensure these are included when hashing the state of the trait. This is verified by a test.

I also include a test to guard against unintended changes of the output from the sync mechanism in the future.

### DIFF
--- a/OpenRA.Game/Sync.cs
+++ b/OpenRA.Game/Sync.cs
@@ -20,18 +20,22 @@ namespace OpenRA
 {
 	[AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
 	public sealed class SyncAttribute : Attribute { }
-	public interface ISync { }	/* marker interface */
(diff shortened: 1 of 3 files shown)


Mailaender:
:+1:

pchote:
Looks sensible, so :+1: after fixing the namespaces (either pulling the sync up, or I guess more likely pushing the other ones down).

RoosterDragon:
Updated to keep namespace consistent with existing tests.

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/2fb00570c781e5d07b3ef0892576ca418f3528a9%5E...2fb00570c781e5d07b3ef0892576ca418f3528a9

PR #8865, opened 2015-08-01
Rename EffectGlobal to LightingGlobal and expose individual channels.

phrohdoh:
~~Without the `Lighting` prefix the property names are too generic in my opinion.~~

A new take on #8853. My effort was not to do this yet but rather to expose anything on the world.
Guess that's not the optimal way.

I can spin up a test-case in RA2 tomorrow if desired.


Mailaender:
Seems rather redundant.

pchote:
This fixes a limitation from #8684 where you can't query the current lighting state.

Looks sensible, but needs a testcase.

Mailaender:
I guess you can get rid of ChangeLighting then.

OpenRA.Mods.Common/Scripting/Global/EffectGlobal.cs
```diff
@@ -1,51 +0,0 @@
...
-
-		public EffectGlobal(ScriptContext context)
-			: base(context)
-		{
-			fpes = context.World.WorldActor.TraitsImplementing<FlashPaletteEffect>();
-			lighting = context.World.WorldActor.TraitOrDefault<GlobalLightingPaletteEffect>();
-		}
-
-		[Desc("Controls the `FlashPaletteEffect` trait.")]
-		public void Flash(string type = null, int ticks = -1)
```
Mailaender: This unnecessary refactoring will break https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/maps/fort-lonestar/fort-lonestar.lua#L153.

Phrohdoh: Sorry I had not caught that `Flash`. There is no need to have two tables that do the same thing.

Phrohdoh:
Fixed.

OpenRA.Mods.Common/Scripting/Global/LightingGlobal.cs
```diff
@@ -0,0 +1,66 @@
...
+		readonly GlobalLightingPaletteEffect lighting;
+
+		public LightingGlobal(ScriptContext context)
+			: base(context)
+		{
+			flashPaletteEffects = context.World.WorldActor.TraitsImplementing<FlashPaletteEffect>();
+			lighting = context.World.WorldActor.TraitOrDefault<GlobalLightingPaletteEffect>();
+
+			if (lighting == null)
+				throw new LuaException("World actor must have GlobalLightingPaletteEffect to use the Lighting table.");
```
pchote: Global tables are always loaded, so this breaks all scripted maps!
RA/TD/D2K now all crash when trying to load their shellmaps.

pchote: You can either return 0 or throw in the individual properties instead.

pchote:
This changes a so-far-unreleased Lua API, so we need to take this before the playtest.

OpenRA.Mods.Common/Scripting/Global/LightingGlobal.cs
Phrohdoh: Hmm checking for null (or a bool) in each property access is a bit distasteful, but okay.

pchote: Less distasteful than completely breaking our RA and TD mods!

Phrohdoh: I didn't say I wouldn't do it or ever compare it to that.

Phrohdoh:
Fixed the breakage of TD and RA.

obrakmann:
Looks good to me :+1:

pchote:
:+1: after removing the test case.

Phrohdoh:
Testcase removed.

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/2870d06a00486575aab336a89980fe25445a3306%5E...2870d06a00486575aab336a89980fe25445a3306

OpenRA.Mods.Common/Scripting/Global/LightingGlobal.cs
```diff
@@ -0,0 +1,64 @@
...
+		public void Flash(string type = null, int ticks = -1)
+		{
+			foreach (var effect in flashPaletteEffects)
+				if (effect.Info.Type == type)
+					effect.Enable(ticks);
+		}
+
+		public double Red
+		{
+			get { return hasLighting ? lighting.Red : 0d; }
```
Mailaender: Shouldn't this be `1d` for unchanged lighting. I think 0 would be black.

Phrohdoh: My idea was to indicate that this wasn't usable (getting a bogus value back).
But perhaps 1d would have been a better default.

Mailaender: `null`/`nil` would be a quite unfriendly and `0d` a bit confusing. ;)

Issue #8780, opened 2015-07-25
Crates are not collectable in TS


pchote:
Introduced by https://github.com/OpenRA/OpenRA/pull/8339/files#diff-8e8bb64e0c97a0eb2f07e03046a5628fR67.  This needs to do a proper ground check.

Issue #8573, opened 2015-06-27
Tiberian Sun infantry can't board APCs anymore


Mailaender:
Reproducible on Hot Springs mpa. Might be a height related issue.

penev92:
It would be pretty easy to verify if it's height-related (btw you wanted an example of where the blank map would be useful).
But as far as I remember, we've always had this issue - cargo (among other things) can't deal with height and expects everything to be at height 0.

Phrohdoh:
See #7904

Issue #8463, opened 2015-06-12
Cell selection on diamond (iso) perspective broken again


reaperrr:
Horizontal axis is fine, but vertically it is off by a half cell. Moving the cursor on the lower half of a cell selects the cell below (easy to see when using the terrain geometry debug overlay in TS).

pchote:
Can you bisect to find what regressed this?

reaperrr:
Finally found the time to track this down.

This seems to have been (re-)introduced by #7847.

pchote:
This was actually introduced by #7748.  I've found the problem, but need to sit down and work out the correct maths for the fix.

PR #8866, opened 2015-08-01
Implement DistanceAboveTerrain

phrohdoh:
#7904 sans-aircraft.

cc @pchote, @penev92

Test by (un)loading cargo on elevated terrain in TS/RA2.

Fixes https://github.com/OpenRA/OpenRA/issues/8573.
Fixes https://github.com/OpenRA/OpenRA/issues/8780.

### DIFF
--- a/OpenRA.Mods.Common/ActorExts.cs
+++ b/OpenRA.Mods.Common/ActorExts.cs
@@ -18,6 +18,25 @@ namespace OpenRA.Mods.Common
 {
 	public static class ActorExts
 	{
+		public static bool IsAtGroundLevel(this Actor self)
+		{
+			if (self.IsDead)
+				return false;
+
+			if (!self.HasTrait<IPositionable>())
+				return false;
+
+			if (!self.IsInWorld)
+				return false;
+
+			var map = self.World.Map;
+
+			if (!map.Contains(self.Location))
+				return false;
+
+			return map.DistanceAboveTerrain(self.CenterPosition).Length == 0;
+		}
+
 		public static bool AppearsFriendlyTo(this Actor self, Actor toActor)
 		{
 			var stance = toActor.Owner.Stances[self.Owner];
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -775,6 +775,13 @@ namespace OpenRA
 			return CenterOfCell(cell);
 		}
 
+		public WDist DistanceAboveTerrain(WPos pos)
+		{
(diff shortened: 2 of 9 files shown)


Mailaender:

pchote:
I don't really like the `TerrainHeightAt` abstraction because you're saying "at" but then completely ignoring the z coordinate.  I was working on this feature plus proper slope height yesterday (before I hit and was forced to fix #8463), and decided on `DistanceAboveTerrain` which uses everything in the given WPos, matches the `DistanceToEdge` function, and returns a result that is directly useful for `actor.IsAtGroundLevel` (check if == 0) and `VoxelRenderable` (just subtract it from the position).

pchote:
Could you also please use `IsAtGroundLevel` at Crate.csL193 to fix #8780?

GraionDilach:
It should be at Crushable imo, not just crates, but the whole crushing logic has that z check.

Phrohdoh:
Updated to reflect the requested changes.

reaperrr:
The `isAir` and `isWater` checks in `CreateEffectWarhead` need to be adapted as well, otherwise they will fail on elevated TS terrain.

OpenRA.Mods.Common/ActorExts.cs
```diff
@@ -18,6 +18,28 @@ namespace OpenRA.Mods.Common
...
+
+			if (!self.IsInWorld)
+				return false;
+
+			var map = self.World.Map;
+
+			if (!map.Contains(self.Location))
+				return false;
+
+			var centerPos = self.CenterPosition;
```
pchote: You aren't doing anything interesting with these temporary variables, so just `return map.DistanceAboveTerrain(self.CenterPosition) == 0`.

Phrohdoh:
Okay, fixed.
But do note:

>[12:16:44] Phrohdoh: okay, i can fix crates as long as you're not going to ask for a bunch of other fixes
[12:16:57] Phrohdoh: the point was to do as little as possible so it doesn't become the other pr
[12:17:12] pchote: those are my only requests
[12:17:16] Phrohdoh: okay

GraionDilach:
Well, I have no further requests regarding this.

pchote:
Confirmed that this fixes transports, but not crates.

pchote:
There's z == 0 check at Mobile.cs:L568 that is breaking crates.  Can you please fix that too?

Phrohdoh:
Thanks, done.

pchote:
Confirmed fixes :+1:

PR #8868, opened 2015-08-01
Make trait fields readonly

pchote:


Phrohdoh:
:+1:

OpenRA.Mods.Common/Effects/Contrail.cs
```diff
@@ -36,9 +36,9 @@ class ContrailInfo : ITraitInfo, Requires<IBodyOrientationInfo>
 
 	class Contrail : ITick, IRender
 	{
-		ContrailInfo info;
-		ContrailRenderable trail;
-		IBodyOrientation body;
+		readonly ContrailInfo info;
+		readonly ContrailRenderable trail;
```
reaperrr: This seems to break contrails (example: badgers on ra shellmap).

reaperrr:
:+1: after contrails are fixed.

OpenRA.Mods.Common/Effects/Contrail.cs
pchote: Oh yes - readonly mutable structs are bad news!

pchote:
Fixed.

PR #8873, opened 2015-08-01
Remove the balance-breaking unit duplication crate.

### LINKED ISSUE #8058: Remove/delay/wait TD clone-function in crates
Hi.

Most of the games in TD are played with crates off. That is because the players say it will end the game early.. Even when we play 3 vs 3... When you then loose you construction yard, often all hope is lost. 
Can the "clone units" have 5-10 min waiting time? Can crates wait before they arrive? Can anything be done? At least the player AoA want a change in crates, and he play TD alot. 
please help us find a way to make players use it again...

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12413234-remove-delay-wait-td-clone-function-in-crates?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Closes #8058.

### DIFF
--- a/mods/cnc/rules/misc.yaml
+++ b/mods/cnc/rules/misc.yaml
@@ -20,11 +20,6 @@ CRATE:
 		SelectionShares: 5
 		Effect: cloak
 		Upgrades: cloak
-	DuplicateUnitCrateAction:
-		SelectionShares: 10
-		MaxAmount: 5
-		MinAmount: 1
-		MaxDuplicateValue: 1250
 	GiveMcvCrateAction:
 		SelectionShares: 0
 		NoBaseSelectionShares: 120


Phrohdoh:
All for it. I've won a few games by unfairly getting 4 or 5 trikes at game start. :+1:

Mailaender:
Can you remove it from https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/rules/misc.yaml#L85 as well?

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7848131f762e11911fd7fdc6963a62a9db9d2c42%5E...7848131f762e11911fd7fdc6963a62a9db9d2c42)

In RA it's not as critical, since it doesn't have starting units by default, the maps are usually bigger, and it has more crate variants.

kyrreso:
They still turn crates off in TD, so this is needed :+1: Boring when you cannot get mcv again after they rushed....

PR #8874, opened 2015-08-01
Fixed RA grenadier prone-stand sequences

reaperrr:
On bleed, RA grenadiers use the wrong frames for prone-stand(2).

### DIFF
--- a/mods/ra/sequences/infantry.yaml
+++ b/mods/ra/sequences/infantry.yaml
@@ -476,11 +476,11 @@ e2:
 			DESERT: TEMPERAT
 			INTERIOR: TEMPERAT
 	prone-stand:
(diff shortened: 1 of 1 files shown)


pchote:
Confirmed the fix :+1:

Note that this is the normal stationary prone animation, and not the (so far unused) standing up transition animation.

penev92:
Seems legit :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4a1a3fef8bb5d1b0b4b10b2be26fc726564b1748...8ca9f7811c17274375b2386e0c3af99b5a83a906)

PR #8875, opened 2015-08-01
Remove bounties from the TS mod.

pchote:
As discussed in IRC.


Phrohdoh:
:+1: This will be happening in RA2 as well.

GraionDilach:
:+1:

Issue #8746, opened 2015-07-19
Distinguish between 'ambient' and regular music


pchote:
There are still a few cases that are poorly handled by our music playlist code because we are trying to apply a one-size-fits-all approach to two different use cases.

Consider the shellmap as an example: main menu music is a standard feature across almost all games, and is usually a low-key ambient music loop.  In our current implementation we treat the shellmap music as a regular track and (based on user preferences) cycle between all the available songs or stop playing anything after the song has completed its first playthrough.  Even worse, if the player has the default RA content download (which only includes one track) then the main menu will loop Hell March - one of our least low-key ambient tracks!  We have similar problems with the Ion Storm ambient track in TS, and the end-game victory/loss tracks.

A simple (and I believe the best) way to handle these cases is to introduce an `AmbientMusic` property on the `MusicPlaylist` trait.  If this is set, then it is looped (ignoring any player music preferences) as the default music choice.  If the file is defined but not present (music content not installed) then it should play nothing instead of selecting an alternative track.  Player selected music (and the InitialMusic trait property) take precident over the ambient music, but the ambient track should start again as soon as the regular playlist runs out of songs or is manually stopped in the music player.

Ambient tracks should also be displayed slightly different in the music player: they show the name and progress in the bottom player, but the playback controls (play/pause/next/prev) should be disabled.

TheRaffy:
Is this related to trait world AmbientSound?

GraionDilach:
No, this is related to the music handling cleanup I've started a while ago. It's 80% aimed towards me anyway.

pchote:
Implemented in #8876.

Issue #8747, opened 2015-07-19
Disable music autoplay if there are insufficient installed tracks.


pchote:
If the player has less than two regular tracks available (ignoring shellmap and score tracks, which are going to be hidden from the player) then we should not autoplay them during regular games.  This will avoid the issue where we loop hell march continuously in RA.

I'm sure we used to do this in the past, but it must have been lost in one of the refactors.

GraionDilach:
After the music-hiding gets done, we could use ShippedSoundtracks for this - it wouldn't count hidden music and as long as there isn't more than SS (oh my), it'd stop autoplaying the playlist.

pchote:
Fixed by #8876.

PR #8876, opened 2015-08-01
Refine the music playlist code.

pchote:
This implements the plan from #8746 with some adaptations.
Also fixes a crash when opening the music player from the TD lobby, and randomizes the default starting song so that it doesn't always play the first song in music.yaml if shuffle is disabled.

### DIFF
--- a/OpenRA.Game/Traits/World/MusicPlaylist.cs
+++ b/OpenRA.Game/Traits/World/MusicPlaylist.cs
@@ -61,10 +61,11 @@ namespace OpenRA.Traits
 
 			random = playlist.Shuffle(Game.CosmeticRandom).ToArray();
 
+			// Always start with a random song
+			currentSong = random.FirstOrDefault();
+
 			if (SongExists(info.StartingMusic))
-			{
 				currentSong = world.Map.Rules.Music[info.StartingMusic];
-			}
 			else if (SongExists(info.BackgroundMusic))
 			{
 				currentSong = currentBackgroundSong = world.Map.Rules.Music[info.BackgroundMusic];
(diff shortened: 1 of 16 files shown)


OpenRA.Game/Traits/World/MusicPlaylist.cs
```diff
@@ -41,44 +36,57 @@ public class MusicPlaylistInfo : ITraitInfo
...
 		public MusicPlaylist(World world, MusicPlaylistInfo info)
 		{
 			this.info = info;
-
-			IsMusicAvailable = world.Map.Rules.InstalledMusic.Any();
+			this.world = world;
 
 			playlist = world.Map.Rules.InstalledMusic.Select(a => a.Value).ToArray();
 
+			IsMusicAvailable = playlist.Length > 1;
```
Phrohdoh: Why is music not available if we only have 1 song?

pchote: This was inherited from @obrakmann's original implementation.  I've removed it for now so we can argue about it in a followup PR.  See #8747.

pchote:
I have scope-creeped this to fix the rest of our music issues too.

GraionDilach:
Oh.... Oh.

What does `ShippedSoundtracks` check now? I might messed it up back then with the first music commit, via enabling the other two tracks from `transit.mix`.

On the other hand, now I can see what got me confused - I wouldn't have added the `BackgroundMusic` system separately.

Also really sorry that you had to do this. >_>

Mailaender:
Seems to work as intended. :white_check_mark: Can't test d2k/ts due to https://github.com/OpenRA/OpenRA/issues/8884, but that didn't regress here.

obrakmann:
Looks fine. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d518fb3c6a845b82caf9157941ce06a9bb794b6a%5E...d518fb3c6a845b82caf9157941ce06a9bb794b6a)

PR #8881, opened 2015-08-02
Fix Aircraft.GetActorBelow() altitude checking

atlimit8:
`Map.DistanceAboveTerrain(WPos pos)` is called directly because `Aircraft` is an `IPositionable` trait and all calls to `Aircraft.GetActorBelow()` occur in `Tick` methods.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
+++ b/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
@@ -127,7 +127,9 @@ namespace OpenRA.Mods.Common.Traits
 
 		public Actor GetActorBelow()
 		{
-			if (self.CenterPosition.Z != 0)
+			// Map.DistanceAboveTerrain(WPos pos) is called directly because Aircraft is an IPositionable trait
+			// and all calls occur in Tick methods.
+			if (self.World.Map.DistanceAboveTerrain(CenterPosition).Length != 0)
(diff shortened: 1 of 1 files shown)


Mailaender:
Looks sensible. :+1:

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -127,7 +127,7 @@ public WVec GetRepulsionForce(Actor other)
 
 		public Actor GetActorBelow()
 		{
-			if (self.CenterPosition.Z != 0)
+			if (self.World.Map.DistanceAboveTerrain(CenterPosition).Length != 0)
```
pchote: Could you please add a comment here with your explanation about not using IsAtGroundLevel?

Phrohdoh: `> 0` for clarity?

atlimit8: @Phrohdoh this is for when an aircraft is docked. A trait similar to `Aircraft` could be used to go underground in which case you would what `GetActorBelow()` to return `null` as well if not at ground level.

atlimit8:
Comment added.

abcdefg30:
Looks also good to me. :+1: / :white_check_mark:

Issue #8855, opened 2015-07-31
Trees and rocks show white on the radar


pchote:
These should use a terrain color or not show on the radar at all.

PR #8882, opened 2015-08-02
Fixed undefined white decoration radar colors

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/8855.

### DIFF
--- a/mods/ts/tilesets/snow.yaml
+++ b/mods/ts/tilesets/snow.yaml
@@ -55,6 +55,10 @@ Terrain:
 		Type: Veins
 		Color: 226, 103, 0
 		TargetTypes: Ground
+	TerrainType@Tree:
+		Type: Tree
+		TargetTypes: Ground
+		Color: 255,28,32,36
 
 # Automatically generated. DO NOT EDIT!
 Templates:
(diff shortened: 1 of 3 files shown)


pchote:
Looks much better.  Thanks :+1:

mods/ts/tilesets/snow.yaml
```diff
@@ -55,6 +55,14 @@ Terrain:
 		Type: Veins
 		Color: 226, 103, 0
 		TargetTypes: Ground
+	TerrainType@Tree:
+		Type: Tree
+		TargetTypes: Ground
+		Color: 255,28,32,36
```
pchote: These dark colours look quite out of place on the bright snow maps.

Mailaender: Propose an alternative. (It is the value from https://github.com/OpenRA/OpenRA/blob/bleed/mods/cnc/tilesets/snow.yaml#L49)

pchote:
This is still an improvement over what we have now, so we can adjust the snow terrain colors later. :+1:

obrakmann:
:+1:

PR #8883, opened 2015-08-02
Remove ProximityCaptor from TS camera.

pchote:
Removes some more rules uglyness that leaked from RA.


Mailaender:
We don't even have it anymore in RA.

PR #8892, opened 2015-08-02
Use background music for Fort Lonestar rain.

pchote:
This improves the ambience (as the rain is not masked with music by default) and prevents a conflict with the game music if the player does choose to play something manually.  The volume of the rain effect is now also tied to the music volume.

### DIFF
--- a/mods/ra/audio/music.yaml
+++ b/mods/ra/audio/music.yaml
@@ -25,7 +25,7 @@ twin: Twin Cannon
 vector1a: Vector
 work226m: Workmen
 
-#Counterstrike tracks
+# Counterstrike tracks
 2nd_hand: The Second Hand
 backstab: Backstab
 shut_it: Shut It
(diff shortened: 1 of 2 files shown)


Phrohdoh:
:+1: works for me and is a bit nicer as stated.

GraionDilach:
:+1:

Mailaender:
Isn't defining ambient sounds as music an abuse?

pchote:
As I mentioned in IRC: no, and this use case was one of the reasons for implementing the background music feature in its current form.  This is the same way that the original games did it.

Mailaender:
However now if you start music in the player, the rain suddenly stops which is weird.

Mailaender:
Well, the background "music" feature is already in and stopping music in the player makes it rain again so everything is working as intended I guess. :white_check_mark:

obrakmann:
I really don't think this is a good idea, for two reasons:

1. Players will have to choose between playing music and having the full complete storm effect.
2. There already is the AmbientSound trait, which fulfills this use case exactly. Or should we remove that now in favor of background music? I'd be very much against that idea.

Mailaender:
Indeed. In practice it doesn't work at all, because music and ambient "music" clashes. https://github.com/OpenRA/OpenRA/pull/9550

PR #8893, opened 2015-08-02
LightingGlobal default of 1d.

phrohdoh:
Followup of #8865 per https://github.com/OpenRA/OpenRA/pull/8865#discussion_r36044159.


pchote:
:+1:

PR #8896, opened 2015-08-02
Fix missile - ground collision checking

atlimit8:


pchote:
It might be better to wait for #8717 first, otherwise this risks breaking missiles completely in TS (since they don't understand how to fly above the terrain yet).

atlimit8:
Well it looked like a simple oversight.

PR #8902, opened 2015-08-02
Makes dropdown menus auto-scroll to selected item.

### LINKED ISSUE #8832: Dropdown menu doesn't scroll to currently selected item
When a menu item that is not visible without scrolling is selected, the next time menu is clicked, it doesn't scroll to show the selected item. This is annoying when the menu has a long list of items. For example in the map editor.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25102883-dropdown-menu-doesn-t-scroll-to-currently-selected-item?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

deniz1a:
Fixes #8832. This is best tested with the tile selector in Tiberian Sun map editor.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/DropDownButtonWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/DropDownButtonWidget.cs
@@ -94,6 +94,10 @@ namespace OpenRA.Mods.Common.Widgets
 				oldBounds.Width,
 				oldBounds.Height);
 			panelRoot.AddChild(panel);
+
+			var scrollPanel = panel as ScrollPanelWidget;
+			if (scrollPanel != null)
+				scrollPanel.ScrollToSelectedItem();
 		}
 
 		public void ShowDropDown<T>(string panelTemplate, int maxHeight, IEnumerable<T> options, Func<T, ScrollItemWidget, ScrollItemWidget> setupItem)
--- a/OpenRA.Mods.Common/Widgets/ScrollPanelWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ScrollPanelWidget.cs
@@ -214,17 +214,8 @@ namespace OpenRA.Mods.Common.Widgets
 			get { return targetListOffset == Math.Min(0, Bounds.Height - ContentHeight) || ContentHeight <= Bounds.Height; }
 		}
 
-		public void ScrollToItem(string itemKey, bool smooth = false)
+		void ScrollToItem(Widget item, bool smooth = false)
 		{
-			var item = Children.FirstOrDefault(c =>
-			{
-				var si = c as ScrollItemWidget;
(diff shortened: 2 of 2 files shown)


OpenRA.Mods.Common/Widgets/ScrollItemWidget.cs
```diff
@@ -57,6 +57,7 @@ public static ScrollItemWidget Setup(ScrollItemWidget template, Func<bool> isSel
 			w.IsVisible = () => true;
 			w.IsSelected = isSelected;
 			w.OnClick = onClick;
+			w.ItemKey = Convert.ToString(Game.CosmeticRandom.Next());
```
deniz1a: I wanted to use the menu item text as ItemKey but didn't know how to do it. This also works.

Phrohdoh: Why is this being done?

deniz1a: So that every menu item has a unique ItemKey.

Phrohdoh: That defeats the purpose of a hotkey (which is meant to be consistent), as it will be new each time.

deniz1a: ItemKey is not a hotkey but an identifying text for the button. Map chooser window uses map UID as ItemKey for example.

Unit158: Why are you converting the key to a string? In any case, we shouldn't just be assigning random numbers as keys here.

deniz1a: Because ItemKey is a string? It is enough that the keys are unique, it doesn't matter what they are. What should I use as ItemKey then?

deniz1a:
Ok I changed the implementation. There is no need to assign a unique ItemKey. I added a new method to ScrollPanelWidget called ScrollToSelectedItem.

Mailaender:
Works as promised. :white_check_mark: Thanks. This was annoying.

pchote:
This is a real nice polish improvement that also helps things like the team selector in the game lobby. Code also looks fine :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/89121320959b944723e64572dacd0b7d35b3ed1a%5E...89121320959b944723e64572dacd0b7d35b3ed1a

PR #8905, opened 2015-08-03
WithProductionOverlay: replace ITick => INotifyCreated & add INotifyOwnerChanged

atlimit8:
Allows WithProductionOverlay to update which `ProductionQueue` is associated with it, including after owner changes, based on the first enabled `ProductionQueue`.

Replaces the use of `ITick.Tick(Actor self)` with `INotifyCreated.Created(Actor self)` and `INotifyBuildComplete.BuildingComplete(Actor self)` for executing initialization code.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithProductionOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithProductionOverlay.cs
@@ -17,7 +17,7 @@ namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Renders an animation when the Production trait of the actor is activated.",
 		"Works both with per player ClassicProductionQueue and per building ProductionQueue, but needs any of these.")]
-	public class WithProductionOverlayInfo : ITraitInfo, Requires<RenderSpritesInfo>, Requires<IBodyOrientationInfo>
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Render/WithProductionOverlay.cs
```diff
@@ -50,7 +51,8 @@ public WithProductionOverlay(Actor self, WithProductionOverlayInfo info)
 			var rs = self.Trait<RenderSprites>();
 			var body = self.Trait<IBodyOrientation>();
 
-			buildComplete = !self.HasTrait<Building>(); // always render instantly for units
+			buildComplete = !self.Info.Traits.Contains<BuildingInfo>(); // always render instantly for units
+			production = self.Info.Traits.Get<ProductionInfo>();
```
Mailaender: Needs a `Requires<ProductionInfo>` then I guess.

atlimit8: Added.

Unit158:
Why do we want this?

atlimit8:
The watched `ProductionQueue` changes on owner change based on first enabled `ProductionQueue` after owner change and replacing the use of `ITick` for initialization is rather obvious.

atlimit8:
Updated description above.

Mailaender:
Can't spot regressions in d2k. :white_check_mark:

RoosterDragon:
:+1: Looks good to me.

Issue #8863, opened 2015-08-01
FrozenActorLayer: Out of sync


deniz1a:
Bleed, shroud off, crates off, debug on. Map: http://resource.openra.net/maps/1715/
````
Red Alert Mod at Version {DEV_VERSION}
Operating System: Windows (Microsoft Windows NT 6.1.7601 Service Pack 1)
Runtime Version: .NET CLR 4.0.30319.42000
Exception of type `System.InvalidOperationException`: Out of sync in frame 829: Blame Tick.
 Compare syncreport.log with other players.
   at OpenRA.Network.OrderManager.OutOfSync(Int32 frame, String blame) in d:\Dev\openra_dev\OpenRA\OpenRA.Game\Network\OrderManager.cs:line 63
   at OpenRA.Network.OrderManager.CheckSync(Byte[] packet) in d:\Dev\openra_dev\OpenRA\OpenRA.Game\Network\OrderManager.cs:line 161
   at OpenRA.Network.OrderManager.<>c__DisplayClass9.<TickImmediate>b__5(Int32 clientId, Byte[] packet) in d:\Dev\openra_dev\OpenRA\OpenRA.Game\Network\OrderManager.cs:line 119
   at OpenRA.Network.ReplayRecorderConnection.<>c__DisplayClass3.<Receive>b__2(Int32 client, Byte[] data) in d:\Dev\openra_dev\OpenRA\OpenRA.Game\Network\ReplayRecorderConnection.cs:line 95
   at OpenRA.Network.EchoConnection.Receive(Action`2 packetFn) in d:\Dev\openra_dev\OpenRA\OpenRA.Game\Network\Connection.cs:line 103
   at OpenRA.Network.ReplayRecorderConnection.Receive(Action`2 packetFn) in d:\Dev\openra_dev\OpenRA\OpenRA.Game\Network\ReplayRecorderConnection.cs:line 82
   at OpenRA.Network.OrderManager.TickImmediate() in d:\Dev\openra_dev\OpenRA\OpenRA.Game\Network\OrderManager.cs:line 112
   at OpenRA.Sync.<>c__DisplayClass5.<CheckSyncUnchanged>b__4() in d:\Dev\openra_dev\OpenRA\OpenRA.Game\Sync.cs:line 168
   at OpenRA.Sync.CheckSyncUnchanged[T](World world, Func`1 fn) in d:\Dev\openra_dev\OpenRA\OpenRA.Game\Sync.cs:line 185
[...]

PR #8908, opened 2015-08-03
Fixed a desync when a player surrenders

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/8863.


pchote:
Good job tracking this down!

I can confirm (by reading the code) that the line you removed would have caused desyncs, and that the line you have added should fix the issue.  I haven't tested this ingame yet, however.

penev92:
Seems to be working, too.
Good job! :+1:

chrisforbes:
:+1:

Issue #8462, opened 2015-06-11
The map script has encountered a fatal error


IJKPrinciple:
When playing GDI mission 3, where you have to destroy the SAM sites (secondary objective) and the Nod base, my game ended, when a fatal error was encountered. It seems that I killed all units on the map but not all buildings were destroyed yet.
Here's lua.log:
```
Fatal Lua Error: [string "BindingSupport.lua"]:30: Actor 'e6' does not define a property 'AttackMove'
   v OpenRA.Scripting.ScriptContext.FatalError(String message)
   v OpenRA.Mods.Common.Scripting.ScriptTriggers.TickIdle(Actor self)
   v OpenRA.World.Tick()
   v OpenRA.Game.InnerLogicTick(OrderManager orderManager)
   v OpenRA.Game.LogicTick()
   v OpenRA.Game.Loop()
   v OpenRA.Game.Run()
   v OpenRA.Program.Run(String[] args)
   v OpenRA.Program.Main(String[] args)

Fatal Lua Error: [string "BindingSupport.lua"]:30: Actor 'e6' does not define a property 'AttackMove'
   v OpenRA.Scripting.ScriptContext.FatalError(String message)
   v OpenRA.Mods.Common.Scripting.ScriptTriggers.TickIdle(Actor self)
   v OpenRA.World.Tick()
   v OpenRA.Game.InnerLogicTick(OrderManager orderManager)
   v OpenRA.Game.LogicTick()
   v OpenRA.Game.Loop()
   v OpenRA.Game.Run()
   v OpenRA.Program.Run(String[] args)
   v OpenRA.Program.Main(String[] args)

Fatal Lua Error: [string "BindingSupport.lua"]:30: Actor 'e6' does not define a property 'AttackMove'
   v OpenRA.Scripting.ScriptContext.FatalError(String message)
   v OpenRA.Mods.Common.Scripting.ScriptTriggers.TickIdle(Actor self)
   v OpenRA.World.Tick()
   v OpenRA.Game.InnerLogicTick(OrderManager orderManager)
   v OpenRA.Game.LogicTick()
   v OpenRA.Game.Loop()
   v OpenRA.Game.Run()
   v OpenRA.Program.Run(String[] args)
   v OpenRA.Program.Main(String[] args)

Fatal Lua Error: [string "BindingSupport.lua"]:30: Actor 'e6' does not define a property 'AttackMove'
   v OpenRA.Scripting.ScriptContext.FatalError(String message)
[...]

PR #8909, opened 2015-08-03
Fixed the Lua API demanding AttackBase for AttackMove Combat properties

Mailaender:
Since https://github.com/OpenRA/OpenRA/pull/4464 every unit can AttackMove whether it has Attack traits or not. Our Lua API doesn't reflect hat which leads to crashes. Fixes https://github.com/OpenRA/OpenRA/issues/8462.

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
+++ b/OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
@@ -29,7 +29,7 @@ namespace OpenRA.Mods.Common.Scripting
 			: base(context, self)
 		{
 			move = self.Trait<IMove>();
-			attackBase = self.Trait<AttackBase>();
+			attackBase = self.TraitOrDefault<AttackBase>();
 		}
 
 		[ScriptActorPropertyActivity]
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
```diff
@@ -86,7 +86,10 @@ public void Attack(Actor targetActor, bool allowMove = true)
 			if (!targetActor.HasTrait<FrozenUnderFog>() && !Self.Owner.CanTargetActor(targetActor))
 				Log.Write("lua", "{1} is not revealed for player {0}!", Self.Owner, targetActor);
 
-			attackBase.AttackTarget(target, true, allowMove);
+			if (attackBase == null)
+				Log.Write("lua", "{0} does not have an Attack* trait. Ignoring order to attack {1}.".F(Self, targetActor));
```
pchote: Our convention is to throw a script error if the map tries to do something bogus.

Mailaender: However there is no mean to check if the actor has Attack* traits in Lua so this will crash and the scripter can't fix it. The checks above also just log stuff.

pchote:
This won't fix #8462 because the script object explicitly `Requires<AttackBaseInfo>` and so the `AttackMove` command is not exported for those actors.  Please test that potential fixes actually work, or say "should fix" instead.

OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
pchote: We do have checks in place: `myLuaActor.HasProperty("Attack")`.

PR #8912, opened 2015-08-03
Added the Nod repair bay icon

reaperrr:
Set up Nod repair bay icon for cases where Nod captures a GDI construction yard.

### DIFF
--- a/mods/ts/rules/gdi-structures.yaml
+++ b/mods/ts/rules/gdi-structures.yaml
@@ -239,6 +239,11 @@ GADEPT:
 	ProvidesPrerequisite@buildingname:
 	SelectionDecorations:
 		VisualBounds: 98, 68, -6, -6
+	RenderBuilding:
+		Image: gadept.gdi
+		RaceImages:
+			gdi: gadept.gdi
+			nod: gadept.nod
 
 GARADR:
 	Inherits: ^Building
--- a/mods/ts/sequences/structures.yaml
+++ b/mods/ts/sequences/structures.yaml
@@ -1031,8 +1031,8 @@ gasilo.nod:
 		Offset: 0, 0
 		UseTilesetCode: false
 
-gadept:
-	Defaults:
+gadept.gdi:
+	Defaults: gtdept
 		Offset: 0, -36
 		UseTilesetCode: true
 	idle:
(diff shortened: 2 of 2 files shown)


mods/ts/rules/palettes.yaml
```diff
@@ -22,7 +22,6 @@
 		Name: chrome
 		Filename: cameo.pal
 		AllowModifiers: false
-		ShadowIndex: 242
```
Mailaender: Sadly this breaks the clock:

image

GraionDilach: The proper fix of this would be to decouple the clock palette from the underlaying icon. I had started to work on this, but my last attempt broke support powers. >_>

pchote:
Do we have a GDI icon for the Nod repair bay?  If not, then I think we should keep it as-is for symmetry and to indicate which buildings are captured tech (or go through and remap all of the rest ourselves).

mods/ts/rules/palettes.yaml
pchote: The easiest fix would be to specify the clock palette on the palette / support power widgets themselves.  There's no need for a per-icon clock sequence or clock sequence palette.

reaperrr:
> Do we have a GDI icon for the Nod repair bay?

There is no Nod repair bay, Nod only has a mobile repair vehicle.

mods/ts/rules/palettes.yaml
reaperrr: Damn...

Anyway, the problem here is that Westwood did not really use regular shadow, but rather translucency, and iirc TS clock uses `anim.pal` rather than `cameo.pal`, too. So I guess the correct solution will be to add a `ClockPalette` to `Buildable`.

GraionDilach: Indeed that was my plan as well - along with per-widget clock sequences - the whole thing was a PEBKAC.

pchote: As I mentioned above, put it on the widgets instead of `Buildable` please.

GraionDilach:
:+1: on the Nod repair bay though.

Mailaender:
> Nod players could capture a GDI construction yard, so let's use the Nod version of its icon.

If you follow that logic, you also need Nod versions for the tech center, turrets etc.

reaperrr:
> If you follow that logic, you also need Nod versions for the tech center, turrets etc.

I meant that in the way of "the nod icon is there and we already use all other faction-specific icons, so let's use this one, too".

Mailaender:
Well, but to use it, you need to duplicate it with a fixed up version so I am not sure that counts. https://github.com/reaperrr/OpenRA/blob/ts-icons1/mods/ts/bits/nfixicon.shp ;)

reaperrr:
Using a custom sequence offset would work, too. I just felt this was a cleaner solution.

reaperrr:
Updated.

- Decided to remove the custom icon shp and used offsets to correct position ingame.
- Added UI unhardcoding of icon clock and darkening.

OpenRA.Mods.Common/Widgets/ProductionPaletteWidget.cs
```diff
@@ -46,6 +48,13 @@ public enum ReadyTextStyleOptions { Solid, AlternatingColor, Blinking }
 		public readonly string TooltipContainer;
 		public readonly string TooltipTemplate = "PRODUCTION_TOOLTIP";
 
+		public readonly string ClockAnimation = ChromeMetrics.Get<string>("ProductionClockAnimation");
+		public readonly string ClockSequence = ChromeMetrics.Get<string>("ProductionClockSequence");
+		public readonly string ClockPalette = ChromeMetrics.Get<string>("ProductionClockPalette");
+		public readonly string NotBuildableAnimation = ChromeMetrics.Get<string>("ProductionNotBuildableAnimation");
+		public readonly string NotBuildableSequence = ChromeMetrics.Get<string>("ProductionNotBuildableSequence");
+		public readonly string NotBuildablePalette = ChromeMetrics.Get<string>("ProductionNotBuildablePalette");
```
Mailaender: Why use chrome metrics here? Just put it on the widget and leave the defaults to work with RA/TD.

GraionDilach: I'd say these could be moved to `ProductionQueue`/`SupportPowerManager` instead.

Mailaender: No, they are UI related just like ReadyText and HoldText.

reaperrr: @pchote said on irc I should use ChromeMetrics...

Mailaender: We only dropped that there for generic re-usable parts of the UI that is why I was asking. Also this will break every custom mods as you don't have upgrade rules to add the missing bits.

reaperrr:
Removed everything except the Nod repair bay icon changes, the rest will come back later in a separate PR when I have more time.

reaperrr:
Counted Graion's +1 for the icon.

Mailaender:
I submitted https://github.com/OpenRA/OpenRA/pull/9129 with my suggested changes.

PR #8913, opened 2015-08-03
Fixed bright dots on TS temperate terrain water

reaperrr:
Temperate water tiles use palette index #1, which is also used for shadow on SHPs.

### DIFF
--- a/mods/ts/rules/palettes.yaml
+++ b/mods/ts/rules/palettes.yaml
@@ -17,6 +17,8 @@
 		Filename: depth.pal
 	PaletteFromCurrentTileset:
 		Name: terrain
+	PaletteFromCurrentTileset@decoration:
+		Name: terraindecoration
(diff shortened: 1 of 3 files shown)


mods/ts/rules/palettes.yaml
```diff
@@ -17,6 +17,8 @@
 		Filename: depth.pal
 	PaletteFromCurrentTileset:
 		Name: terrain
+	PaletteFromCurrentTileset@decoration:
+		Name: decoterrain
```
pchote: Can we call this `terraindecoration` instead?

reaperrr:
Updated.

Mailaender:
Nicely solved. :+1: This was very visible on the TEMPERAT waterfalls.

GraionDilach:
:+1:

PR #8921, opened 2015-08-03
Added Tiberian Sun map Pit Or Plateau

Mailaender:
This map is open and does not contain (vital) bridges. The layout is balanced and conservative. See http://ra2maps.zzattack.org/Tiberian%20Sun/Official%20Multiplayer/getmap.php?map=Pit+or+Plateau


mods/ts/maps/pitplats/map.yaml
```diff
@@ -0,0 +1,837 @@
+MapFormat: 7
+
+RequiresMod: ts
+
+Title: Pit Or Plateau
+
+Description: Describe your map here
```
Unit158: !

Mailaender: Added some descriptions.

GraionDilach:
One of my favourite maps from TS. (y) ofc.

pchote:
If this is the same as the map we played online, then the player setup for Neutral isn't correct - their structures show up in the "All Players" observer mode.

pchote:
This also needs a larger cordon on the north edge of the map to prevent the glitches at the top of the minimap

screen shot 2015-08-04 at 22 36 52

pchote:
The southern spawn is right on the map border. IIRC from the other maps you need to add 4 to the map height to fix the bottom edge.  You'll be able to compare the painted terrain with the cordon in the editor to find the right value.

abcdefg30:
> If this is the same as the map we played online, then the player setup for Neutral isn't correct - their structures show up in the "All Players" observer mode.

Looks like a regression to me, as this happens in all maps with actors owned by neutral.
(See `Hot Springs`, `Tread Lightly` or `Center of Attention redux 2` [oil derricks there].)

Mailaender:
Fixed in https://github.com/OpenRA/OpenRA/pull/8964.

Mailaender:
If I change map size or bounds slightly the game crashes so I am giving up here. It was uploaded at http://resource.openra.net/maps/3953 for playtesting.

pchote:
Unfortunately my WIP importer still needs a bunch of work before it will be able to import maps that don't need manual tweaking.

PR #8922, opened 2015-08-03
Fix a divide by zero crash in TabCompletionLogic.

pchote:
Repro case: type "/alld" into the chat, press tab, delete the last character, press tab again.


Mailaender:
:+1:

PR #8933, opened 2015-08-04
Reduced the bloated Travis CI Mono installation

Mailaender:
Due to https://github.com/travis-ci/travis-ci/issues/4571 and http://docs.travis-ci.com/user/languages/csharp/ including F#, Visual Basic and other stuff we don't need, I will try to reduce the VM setup by manually specifying just the dependencies we need.

:no_entry: **DON'T MERGE JUST YET. THIS IS A TEST UNTIL THIS MESSAGE IS REMOVED!**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 # Travis-CI Build for OpenRA
 # see travis-ci.org for details
 
-language: csharp
-mono: 4.0.0
+language: c
 
 # http://docs.travis-ci.com/user/migrating-from-legacy
 sudo: false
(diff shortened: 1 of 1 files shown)


abcdefg30:
Waiting for travis-ci/apt-source-whitelist#94.

akoeplinger:
I'm curious to know how much this really shaves off from the Mono installation time on Travis, as we had some discussions initially whether to split the languages or keep all in one and decided for the latter as we thought it's not a huge deal and makes building multi-lang solutions easier :)

.travis.yml
```diff
@@ -13,7 +12,17 @@ cache:
 
 addons:
   apt:
+    sources:
+    - mono-4.0
     packages:
+    - mono-dmcs
```
akoeplinger: dmcs is deprecated, just use mono-mcs here

Mailaender:
Sadly I can't give this a test run until https://github.com/travis-ci/apt-source-whitelist/pull/94 is reviewed and merged.

akoeplinger:
You could try it on "sudo: required" (might be a bit slower to start though).

Btw. since you're going through the trouble of whitelisting, why not white list the normal non-snapshot repo as well so you don't need to do it again when Mono 4.2 comes out?

Mailaender:
It is 40 s on installing manually on the legacy infrastructure vs. I guess ~30 s (it doesn't display it properly) on the new infrastructure. Looks like I overestimated the gain and actual problem. As https://github.com/travis-ci/apt-source-whitelist/ looks pretty much unmaintained I am giving up here.

PR #8936, opened 2015-08-04
Improved the --check-yaml performance

Mailaender:
This speeds up the linter by aborting early. It checks either mod or map and skips redundant checks such as unchanged rules.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckActorReferences.cs
+++ b/OpenRA.Mods.Common/Lint/CheckActorReferences.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System;
+using System.Linq;
 using System.Reflection;
 using OpenRA.Traits;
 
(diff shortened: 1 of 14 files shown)


penev92:
Having all these checks in `CheckYaml`'s `foreach` would be much more concise and would mean future lint checks won't do bad stuff like ignoring the constraint implemented here for the others.

abcdefg30:

Mailaender:
> Having all these checks in CheckYaml's foreach would be much more concise and would mean future lint checks won't do bad stuff like ignoring the constraint implemented here for the others.

As discussed in IRC the checks are too individual to generalize.

pchote:
I'm also not a fan of the duplicated boilerplate at the top of each pass.
A much cleaner solution will be to split our generic `ILintPass` into separate `ILintRulesPass`, `ILintSequencesPass`, `ILintMapPass`, etc, so that we *can* pull out the generic boilerplate.

Most of the changes in this pr are removing map references from the passes, and collecting the boilerplate at the top of the pass.  This seems like a reasonable step towards the longer term fix, so i'm happy to :+1: this as-is.

Mailaender:
I submitted https://github.com/OpenRA/OpenRA/pull/9083 as a followup with your suggestions.

Issue #6879, opened 2014-11-01
Automatically check the Lua scripts for errors


Mailaender:
With the old Lua API the missions broke a lot. The situation has improved, but I fear breakage will come in the future and with more missions, manual testing is not feasible. My idea is some kind of OpenRA.Lint rule to execute Lua script and check for breakage. It should at least go through WorldLoaded or find other obvious problem such as syntax errors and broken actor references.

I started at https://github.com/Mailaender/OpenRA/compare/lint-map-scripts but it does not work at all.

Mailaender:
You are right. You can use the Lua compiler itself to do syntax checks. http://stackoverflow.com/a/5347221 This might also help during development as it is faster than starting the mission and searching the stacktrace for useable information.

Issue #5490, opened 2014-05-30
Port missions to the new Lua API


Mailaender:
The new https://github.com/OpenRA/OpenRA/wiki/Lua-API has landed in https://github.com/OpenRA/OpenRA/pull/5190. Time to update the maps and enhance the standard library. Some shared deficiencies of both old and new Lua API:

- [x] #4400 Missing cell triggers
- [x] #4541 Lua integration of objectives panel 

Red Alert
------------

| Mission                          | Basis        | Assignee       | PR      | Status
|-------------------------------|--------------|------------------|----------|---------|
| allies-01-classic             | old Lua     | @Mailaender | #6377 | done |
| allies-02-classic             | old Lua     | @Mailaender | #6379 | done |
| ~~allies01~~                  | legacy C# |                       |            | redundant |
| allies02 (co-op)             | legacy C# | @obrakmann | #11316 | done |
| allies03 (co-op)             | legacy C# | @Mailaender | #11679 | done |
| allies04 (co-op)             | legacy C# | @Mailaender |#11702 | done |
| desert-shellmap             | old Lua    | @pchote        | #5190 | done  |
| fort-lonestar                   | legacy C# | @Mailaender | #5469 | done |
| intervention                   | old Lua     | @obrakmann | #6762 | done  |
| monster-tank-madness | legacy C# | @abcdefg30  | #6552 | done |
| soviet01-classic            | legacy C# | @Mailaender | #6796 | done |
| survival01                     | legacy C# |  @abcdefg30 | #6934 | done |
| survival02                     | legacy C# |  @abcdefg30 | #7164 | done |

Tiberian Dawn
------------------

| Mission      | Basis        | Assignee       | PR      | Status
|---------------|--------------|------------------|----------|---------|
| gdi01         | old Lua     | @Mailaender | #5782 | done |
| gdi02         | old Lua     | @obrakmann | #6324 | done |
| gdi03         | old Lua     | @Mailaender | #6795 | done |
| gdi04a       | old Lua     | @obrakmann | #6762 |  done  |
| gdi04b       | old Lua     | @obrakmann | #6762 |  done  |
| gdi04c       | old Lua     | @obrakmann | #6325 | done |
| nod01       | old Lua     | @obrakmann | #6326 | done |
| nod03a     | old Lua     | @obrakmann | #6327 | done |
| nod03b     | old Lua     | @obrakmann | #6328 | done |
| shellmap    | old Lua     | @pchote       | #5190 | done |

[...]

PR #8939, opened 2015-08-04
Added checks for syntax errors in Lua scripts

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/6879.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ cache:
 addons:
   apt:
     packages:
+    - lua5.1
     - nsis
     - nsis-common
     - dpkg
@@ -31,6 +32,7 @@ script:
  - travis_retry make all-dependencies
  - make all
  - make check
+ - make check-scripts
  - make test
 
 # Automatically update the trait documentation and Lua API
--- a/Makefile
+++ b/Makefile
@@ -170,6 +170,12 @@ mod_ts_LIBS = $(COMMON_LIBS) $(STD_MOD_LIBS) $(mod_common_TARGET)
 PROGRAMS += mod_ts
 mod_ts: $(mod_ts_TARGET)
 
+check-scripts:
+	@echo
+	@echo "Checking for Lua syntax errors..."
+	@luac -p $(shell find mods/*/maps/* -iname '*.lua')
+	@luac -p $(shell find lua/* -iname '*.lua')
+
 check: utility mods
 	@echo
 	@echo "Checking for code style violations in OpenRA.Game..."
--- a/make.ps1
+++ b/make.ps1
@@ -35,6 +35,7 @@ if ($args.Length -eq 0)
 	echo "                  'dependencies' commands to restore removed files."
 	echo "  test            Tests the default mods for errors."
 	echo "  check           Checks .cs files for StyleCop violations."
+	echo "  check-scripts   Checks .lua files for syntax errors."
 	echo "  docs            Generates the trait and Lua API documentation."
 	echo ""
 	$command = (Read-Host "Enter command").Split(' ', 2)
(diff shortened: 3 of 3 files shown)


Unit158:
This will require a wiki update stating that Lua 5.1 is now a requirement for make check.

abcdefg30:
Could you also add the check to `make.ps1`?

pchote:
Please either make this optional, or use a separate command for lua vs code checks.  I basically never modify lua files in a way that need syntax checks, so I don't want to be forced to install a system-wide lua package in order to continue using the code style checks.  Likewise, most people working on scripted maps are generally not going to be working on C# code changes at the same time.

Mailaender:
> Could you also add the check to `make.ps1`?

No, I don't have Windows nor do I know how Lua distributions there work and which are common.

Mailaender:
The "I don't want to install Lua (on Mac)" argument is weak. However you are right that mixing those StyleCop checks, which also run a lot longer, with missions together doesn't make much sense so I separated those.

Phrohdoh:
It is not weak at all.
I don't want to install Lua to check C# is perfectly valid.

Mailaender:
> This will require a wiki update stating that Lua 5.1 is now a requirement for make check.

Not anymore. I separated it into `check-scripts` as requested.

GraionDilach:
`make.ps1` still needs to be expanded though.

Mailaender:
Not in this pull request and not by me though.

pchote:
Can you please push a test commit that shows that this fixes

> It should at least go through WorldLoaded or find other obvious problem such as syntax errors and broken actor references.

Mailaender:
Wow, I am astonished that it the `sudo zypper install lua` is such a Linux exclusive luxury. Added a test case for you. It obviously can't check actor references.

pchote:
You're proposing a change to our travis test setup, so it is unreasonable to expect us to merge this without verification that that actually works.  There have been cases in the past where changes where checks were added without testing and only later found to be completely ineffectual because the check did not return 1 on error.

The attitude is unnecessary.

Mailaender:
Don't worry, I locally tested the exit codes. Shall I remove the test commit again?

Phrohdoh:
`luac` being sandboxed is terrible at giving the real error, unfortunately:

```
Checking for Lua syntax errors...
luac: mods/ra/maps/desert-shellmap/desert-shellmap.lua:40: '(' expected near 'if'
make: *** [check-scripts] Error 1
The command "make check-scripts" exited with 2.
```

Mailaender:
Nothing is sandboxed here. This is the real Lua compiler doing it's syntax checks. It spits out the line and position of the error. This is really useful for me and I am annoyed by this simple pull request being delayed and bad-mouthed here. I wanted to work on https://github.com/OpenRA/OpenRA/issues/5490 but made only slow progress due to having to reload the game only to get a crash from a simple syntax error.

Phrohdoh:
The real error is a missing `(a)` not just a `(` preceding the `if` in the function body.

Mailaender:
Static source code analysis isn't trivial for Lua due to the flexibility of the language.

Unit158:
I think it is important to note that Lua is so powerful that you can build an entire other language on it without a huge amount of effort. I'm not talking about a new interpreter either. Just a language that is a script.

Phrohdoh:
@Unit158 I'm not understanding how that is relevant (especially since this can be done with many languages).

Unit158:
It is relevant because Lua can't see the actual "structure" of the program as so much as it sees the words.

penev92:
:+1: from me.
If travis checks all PR-ed/merged scripts and mappers can check them if they want, that's good enough. Kind of sucks that it doesn't include Windows, but `¯\_(ツ)_/¯`
@abcdefg30 can look into it if he feels it can benefit him.

Mailaender:
Removed the test case again. See https://travis-ci.org/OpenRA/OpenRA/builds/75810433 for the recorded run.

abcdefg30:
:+1: / :white_check_mark: with https://github.com/Mailaender/OpenRA/pull/55.

Mailaender:
Thanks!

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/4ab2848666ce29abc31b2318937caa2c33a17f09...22fb500391ef9d24a2e168d55e6cc8284c75ac50)

Issue #8940, opened 2015-08-04
KeyNotFoundException at OpenRA.Mods.Common.Traits.Explodes.Killed(Actor self, AttackInfo e)


OmegaBolt:
http://pastebin.com/1PAPxGAP

```
<+pchote> that is probably related to damage types
<+pchote> i'll blame reaperr for that
<+pchote> can you please file a bug with that - this is something that we should at least be giving a sensible error message about
<+pchote> you can copy these four irc lines into the ticket if you want, in case someone else tells you "this is you doing x - not a bug"
<OmegaBolt___> so it could be nothing to do with what I did? :P
<+pchote> no, it probably is something you did
<+pchote> but openra sucks for not telling you that in a nicer way
```

penev92:
Closing this. We can now reopen with a more descriptive exception if it happens again.

PR #8942, opened 2015-08-04
Add specific exception message when Explodes cannot find a weapon

reaperrr:
Additionally, accept empty string to 'disable' `Explodes` weapons.

Last but not least, since we define `UnitExplode` as default for both filled and empty weapon, the d2k grenadier used it when reloading(!), causing massive damage.

Adresses #8940 (or at least helps debugging that crash).

### DIFF
--- a/OpenRA.Mods.Common/Traits/Explodes.cs
+++ b/OpenRA.Mods.Common/Traits/Explodes.cs
@@ -8,6 +8,7 @@
  */
 #endregion
 
+using System;
 using System.Linq;
 using OpenRA.Mods.Common.Warheads;
 using OpenRA.Traits;
@@ -17,7 +18,7 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("This actor explodes when killed.")]
 	public class ExplodesInfo : ITraitInfo
 	{
-		[WeaponReference, Desc("Weapon to use for explosion if ammo/payload is loaded.")]
+		[WeaponReference, FieldLoader.Require, Desc("Weapon to use for explosion if ammo/payload is loaded.")]
 		public readonly string Weapon = "UnitExplode";
 
 		[WeaponReference, Desc("Weapon to use for explosion if no ammo/payload is loaded.")]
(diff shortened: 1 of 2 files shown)


reaperrr:
Note: I'm not sure about defaulting to 'null' (3rd commit). Generally speaking I think this is better because those magic, not-visible-in-yaml defaults are always troublesome for modders.

On the other hand, it slightly increases the risk of regressions in the shipping mods, unless we catch any existing problems before this is merged.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -32,14 +33,20 @@ public class ExplodesInfo : ITraitInfo
...
 	public class Explodes : INotifyKilled
 	{
 		readonly ExplodesInfo info;
 
-		public Explodes(ExplodesInfo info) { this.info = info; }
+		public Explodes(ActorInitializer init, ExplodesInfo info)
+		{
+			this.info = info;
+
+			if (info.Weapon == null && info.EmptyWeapon == null)
```
pchote: Could we use `[Require]` on the traitinfo instead?  I don't think it would be that big of a deal to force modders to define both the empty and loaded versions, and that gives us static guarantees.

pchote: Or we could `[Require]` only the `Weapon`, and leave `EmptyWeapon` optional and defaulting to null.

reaperrr:
Updated.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -54,9 +55,13 @@ public void Killed(Actor self, AttackInfo e)
 				return;
 
 			var weaponName = ChooseWeaponForExplosion(self);
-			if (weaponName == null)
+			if (string.IsNullOrEmpty(weaponName))
 				return;
 
+			if (!e.Attacker.World.Map.Rules.Weapons.ContainsKey(weaponName.ToLowerInvariant()))
```
pchote: The `WeaponReference` already gives us this, so i'm not convinced we need this.  I guess this is going against my comments in #8940 however, so feel free to ignore me.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -17,11 +18,11 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("This actor explodes when killed.")]
 	public class ExplodesInfo : ITraitInfo
 	{
-		[WeaponReference, Desc("Weapon to use for explosion if ammo/payload is loaded.")]
+		[WeaponReference, FieldLoader.Require, Desc("Weapon to use for explosion if ammo/payload is loaded.")]
 		public readonly string Weapon = "UnitExplode";
 
 		[WeaponReference, Desc("Weapon to use for explosion if no ammo/payload is loaded.")]
-		public readonly string EmptyWeapon = "UnitExplode";
+		public readonly string EmptyWeapon = null;
```
abcdefg30: Are we sure that this change doesn't want an upgrade rule?

abcdefg30: There are several places in our default mods yaml where we don't explicitly give an `EmptyWeapon`.

abcdefg30:

reaperrr:
Currently on't have the time and 'head-bandwidth' to deal with an upgrade rule, so I reverted the null-default for EmptyWeapon and fixed the d2k grenadier manually instead.

penev92:
:+1:

PR #8957, opened 2015-08-05
Added Tiberian Sun decorational actors

Mailaender:
including boxes/crates, barrels and palettes which for some reasons are overlays in Tiberian Sun. We will make them normal actors. Kept the names compatible with the conversion code at https://github.com/OpenRA/OpenRA/compare/bleed...pchote:ts-map-importer#diff-0ae9cae5772783a57dba2df98e2bde26R82

### DIFF
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -565,3 +565,19 @@
 	Tooltip:
 		Name: Rock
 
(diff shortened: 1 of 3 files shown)


GraionDilach:
:+1:

RoosterDragon:
:+1:

Issue #8961, opened 2015-08-05
Actors belonging to the Neutral player reveal fog and shroud for spectators


penev92:
Happens when you select the "All Players" visibility option, and my guess is it's pretty new.

obrakmann:
Looks like that was my fault again: git bisect found fc8a34a7a2684dbe317e6dc25a4b7ccde65812e1

Mailaender:
Proposed fix at https://github.com/OpenRA/OpenRA/issues/8964.

PR #8964, opened 2015-08-05
Fixed neutral actors being visible in the All Players observer shroud mode

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/8961.

### DIFF
--- a/OpenRA.Game/Player.cs
+++ b/OpenRA.Game/Player.cs
@@ -137,8 +137,8 @@ namespace OpenRA
 		public Dictionary<Player, Stance> Stances = new Dictionary<Player, Stance>();
 		public bool IsAlliedWith(Player p)
 		{
-			// Observers are considered as allies
-			return p == null || Stances[p] == Stance.Ally || p.Spectating;
+			// Observers are considered allies to active combatants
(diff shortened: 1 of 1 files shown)


OpenRA.Game/Player.cs
```diff
@@ -137,8 +137,8 @@ public override string ToString()
 		public Dictionary<Player, Stance> Stances = new Dictionary<Player, Stance>();
 		public bool IsAlliedWith(Player p)
 		{
-			// Observers are considered as allies
-			return p == null || Stances[p] == Stance.Ally || p.Spectating;
+			// Human observers are considered as allies
```
pchote: This isn't quite right.  You're attributing humanity to the wrong player, but then that's confusing with respect to the AI...

How about `// Observers are considered allies to active combatants`

Mailaender: Hadn't thought about the bots. Updated.

pchote:
Note for testing: the Spectating flag is there mainly for players who die, so make sure to test this case and not just regular observers.

penev92:
Works :+1:

abcdefg30:
:+1: / :white_check_mark:

PR #8749, opened 2015-07-19
Added a lint rule for palette references

Mailaender:
Closes #8689

PR #8965, opened 2015-08-05
Fixed the RallyPoint trait for Tiberian Sun polish

Mailaender:
The sprites are not hard-coded, but configurable, self-documented and partly optional. Makes it lint testable regarding sprites and also future palette references https://github.com/OpenRA/OpenRA/pull/8749 as well.

### DIFF
--- a/OpenRA.Mods.Common/Effects/RallyPointIndicator.cs
+++ b/OpenRA.Mods.Common/Effects/RallyPointIndicator.cs
@@ -20,22 +20,22 @@ namespace OpenRA.Mods.Common.Effects
 	{
 		readonly Actor building;
 		readonly RallyPoint rp;
-		readonly string palettePrefix;
+		readonly string paletteName;
 		readonly Animation flag;
 		readonly Animation circles;
 
-		public RallyPointIndicator(Actor building, string palettePrefix)
+		public RallyPointIndicator(Actor building, string paletteName)
 		{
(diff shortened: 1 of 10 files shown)


Phrohdoh:
The point in our `Desc` attribute is to provide descriptions even in the case of self-documentation.

OpenRA.Mods.Common/Effects/RallyPointIndicator.cs
```diff
@@ -20,33 +20,44 @@ class RallyPointIndicator : IEffect
...
+		readonly string paletteName;
 		readonly Animation flag;
 		readonly Animation circles;
 
-		public RallyPointIndicator(Actor building, string palettePrefix)
+		public RallyPointIndicator(Actor building, string palette)
 		{
 			this.building = building;
-			this.palettePrefix = palettePrefix;
+			this.paletteName = palette;
```
abcdefg30: I don't think you need the `this.` here anymore.

abcdefg30:
TS looks good. But in RA are the circles gone:
rally-nocircle

Mailaender:
Woops, reverted some parts to fix this.

abcdefg30:

RoosterDragon:
:+1: Needs a rebase.

Mailaender:
Rebased.

PR #8966, opened 2015-08-05
Makes map script error display more noticeable.

### LINKED ISSUE #8962: Lua exception dialog no longer works properly
The PR that removed the forced end-game screen regressed our handling of script errors.

The game will now throw up the regular game menu when the script crashes, with the error message showing as plain text below the center panel tabs (which are visually more prominent, so make the error less obvious).  The player can return to the game which then remains paused with no obvious indication why.

The return to game button should be disabled if the game was ended due to a script error, and the briefing tab should be hidden so that the error message takes priority.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25409750-lua-exception-dialog-no-longer-works-properly?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

deniz1a:
The script error display in the ingame menu is now more noticeable. This PR does the following:
- Changes the text of Objectives tab button to Script Error if there was an error. And also flashes the button.
- Changes the title text color of error message from white to red.

Fixes #8962.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/GameInfoLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/GameInfoLogic.cs
@@ -8,7 +8,6 @@
  */
 #endregion
 
-using System;
 using System.Linq;
 using OpenRA.Mods.Common.Scripting;
 using OpenRA.Traits;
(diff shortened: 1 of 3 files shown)


Unit158:
:-1: This is not what the consensus ruled.

pchote:
I agree: This is still significantly less clear than the original error dialog.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/8969. I found some more weaknesses in the code which are easier to fix myself than to explain.

PR #8967, opened 2015-08-06
disable beacon button if game is in single player mode

### LINKED ISSUE #8535: UI: single player missions should not have a button for beacon placement
there's nobody who can see the signal

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/22664037-ui-single-player-missions-should-not-have-a-button-for-beacon-placement?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

small-fishes:
Check to see if the game is in single player mode and if it is disable the beacon button.

Closes #8535

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/OrderButtonsChromeLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/OrderButtonsChromeLogic.cs
@@ -68,6 +68,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			if (beacon != null)
 			{
 				beacon.GetKey = _ => Game.Settings.Keys.PlaceBeaconKey;
+				beacon.IsDisabled = () => world.LobbyInfo.IsSinglePlayer;
 				OrderButtonsChromeUtils.BindOrderButton<BeaconOrderGenerator>(world, beacon, "beacon");
 			}
 		}


pchote:
I like to use beacons to point out crates to myself in singleplayer, so that I can select a unit and then jump immediately back to order them to pick it up.

Mailaender:
No consensus. Closing both this and the issue then. Sorry.

pchote:
I intended that as a discussion point (written wearing my occasional-player hat) rather than a review comment!  If my use case is out of the ordinary or something we don't want to support then there may well be a case for disabling this in singleplayer (note: this includes both skirmishes as well as missions).

abcdefg30:
> 14:12 (+Phroh|work) #8967 was not nearly given enough time or discussion
14:13 (+Phroh|work) Can someone reopen that please?
14:13 (+Phroh|work) I'm all for that PR.

abcdefg30:
Personally I'm not really convinced, but I'm not using beacons in singleplayer either.
Code looks good anyway and this works ingame. :white_check_mark: 

But we probably want to gray the button out (or sth similar), otherwise this could look like a bug to users.

GraionDilach:
Beacons would be useful to mark mission objectives as well.  My vote goes for keep.

abcdefg30:
> Beacons would be useful to mark mission objectives as well. My vote goes for keep.

That can be done via lua and shouldn't be affect by this change.
(Assuming you refer to singeplayer missions/campaigns.)

MicroJOo:
keep

nalidixic:
"But we probably want to gray the button out (or sth similar), otherwise this could look like a bug to users." 

What about just removing the button all together instead of disabling it in single player mode?

abcdefg30:
> What about just removing the button all together instead of disabling it in single player mode?

Hm, you could try using `IsVisible` instead of `IsDisabled`, yeah.

pchote:
That will look weird in D2K, which has the beacon button near the middle.

dan9550:
as @pchote said i don't mind using beacons in single player also just for little reminders and such.

nalidixic:
You guys won't hurt my feelings if you close this... 

This seems like it's just not an issue to most people.

I'll find another bug to work on.

Phrohdoh:
Now closing with a pretty clear consensus.

PR #8969, opened 2015-08-06
Fixed the script error UI not shutting down the game properly

### LINKED ISSUE #8962: Lua exception dialog no longer works properly

Mailaender:
Fixes #8962.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/GameInfoLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/GameInfoLogic.cs
@@ -39,7 +39,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				numTabs++;
 				var objectivesTabButton = widget.Get<ButtonWidget>(string.Concat("BUTTON", numTabs.ToString()));
 				objectivesTabButton.GetText = () => "Objectives";
-				objectivesTabButton.IsVisible = () => lp != null && numTabs > 1;
+				objectivesTabButton.IsVisible = () => lp != null && numTabs > 1 && !hasError;
 				objectivesTabButton.OnClick = () => activePanel = IngameInfoPanel.Objectives;
 				objectivesTabButton.IsHighlighted = () => activePanel == IngameInfoPanel.Objectives;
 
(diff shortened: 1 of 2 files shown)


penev92:
Looks good :+1: 
I'm not sure we want to disable `resumeButton` though. It might pay off to be able to return to looking at the map once you see there was a crash.

Mailaender:
Enabled the `resumeButton` again.

OpenRA.Mods.Common/Widgets/Logic/Ingame/GameInfoLogic.cs
```diff
@@ -78,7 +78,7 @@ public GameInfoLogic(Widget widget, World world, IngameInfoPanel activePanel)
 				numTabs++;
 				var debugTabButton = widget.Get<ButtonWidget>(string.Concat("BUTTON", numTabs.ToString()));
 				debugTabButton.Text = "Debug";
-				debugTabButton.IsVisible = () => lp != null && world.LobbyInfo.GlobalSettings.AllowCheats && numTabs > 1;
+				debugTabButton.IsVisible = () => lp != null && world.LobbyInfo.GlobalSettings.AllowCheats && numTabs > 1  && !hasError;
```
abcdefg30: Double space before the `&&`.

Mailaender: Woops. Fixed.

abcdefg30:
:+1: / :white_check_mark:

PR #8971, opened 2015-08-06
Removed executable bits from random files in the repository

Mailaender:
This will give rpmlint and lintian warnings which will be irritating. I think Ubuntu exposes them to the user and reports "bad quality" packages.


GraionDilach:
Whoops, sorry about the last two. :+1: though.

abcdefg30:
:+1:

PR #8979, opened 2015-08-07
Remove producer-is-in-world assumptions from Production

penev92:
Justification:

> D2k upgrades need to be actors that never actor the world, produced by the PlayerActor, who is also not in the world, but the Production trait expects the producer to be in the world even if it has `Exit.MoveIntoWorld = False`, allowing the producees to not enter the world.

Use-case resides on @OmegaBolt's [d2k balancing branch](https://github.com/OmegaBolt/OpenRA).
Anything currently in bleed should work exactly the same as it does now.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Production.cs
+++ b/OpenRA.Mods.Common/Traits/Production.cs
@@ -35,39 +35,50 @@ namespace OpenRA.Mods.Common.Traits
 		public ProductionInfo Info;
 		public string Faction { get; private set; }
 
+		readonly bool occupiesSpace;
+
 		public Production(ActorInitializer init, ProductionInfo info)
 		{
 			Info = info;
+			occupiesSpace = init.Self.Info.Traits.WithInterface<IOccupySpaceInfo>().Any();
 			rp = Exts.Lazy(() => init.Self.IsDead ? null : init.Self.TraitOrDefault<RallyPoint>());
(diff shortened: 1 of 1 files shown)


OmegaBolt:
Tested this on the my branch and the upgrades now finish production and 'exit' without a crash, as expected.

Mailaender:
Current d2k updates and spy induced veterancy still works. :white_check_mark: Also no regressions on the current exit point behavior.

OpenRA.Mods.Common/Traits/Production.cs
```diff
@@ -35,39 +35,50 @@ public class Production
 		public ProductionInfo Info;
 		public string Faction { get; private set; }
 
+		readonly bool occupiesSpace;
+
 		public Production(ActorInitializer init, ProductionInfo info)
 		{
 			Info = info;
+			occupiesSpace = init.Self.Info.Traits.WithInterface<IOccupySpaceInfo>().Any();
```
pchote: This could have been simplified to `.Contains<>`.

OpenRA.Mods.Common/Traits/Production.cs
```diff
@@ -107,7 +118,13 @@ public virtual bool Produce(Actor self, ActorInfo producee, string factionVarian
 			if (Reservable.IsReserved(self))
 				return false;
 
-			// pick a spawn/exit point pair
+			if (!occupiesSpace)
```
pchote: This unnecessarily duplicates the block below.

PR #8627, opened 2015-07-03
Added Spiceblooms.

Unit158:
~~I currently need help rendering the spicebloom properly.~~ Closes https://github.com/OpenRA/OpenRA/issues/4462

Issue #4243, opened 2013-12-06
Disallow Spice grow overlaps on cliffs


Mailaender:
![spice-cliff-glitches](https://f.cloud.github.com/assets/756669/1691183/90e23a3a-5e5c-11e3-9f39-070e271c3d70.png)

arrakis.yaml needs some tweaking.

PR #8987, opened 2015-08-08
Added spice blooms

### LINKED ISSUE #4462: Spiceblooms
Dune2000 is currently missing "spice blooms", which is an important part of the game. I am currently working on it.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1377662-spiceblooms?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mailaender:
https://github.com/OpenRA/OpenRA/pull/8627 with unrelated stuff moved into it's own commit and removed the `Health.Resurect` hack.

Closes #4462.


Unit158:
With the delayed action, you don't need to use the counter when it is dead.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -0,0 +1,169 @@
...
+		int counter;
+		int respawnTime;
+		int growTime;
+
+		public SpiceBloom(ActorInitializer init, SpiceBloomInfo info)
+		{
+			this.info = info;
+			self = init.Self;
+
+			resType = self.World.WorldActor.TraitsImplementing<ResourceType>().First(t => t.Info.Name == info.ResourceType);
```
pchote: Might want to think about using *OrDefault and then manually throwing a more understandable exception if there are no matches (same for the line below).

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -0,0 +1,169 @@
...
+					self.Kill(self);
+				else
+				{
+					var index = info.GrowthSequences.Length * counter / growTime;
+					anim.Animation.Play(info.GrowthSequences[index]);
+				}
+			}
+			else
+			{
+				counter++;
```
pchote: I thought we were going to do this with aggregated damage so that attacking the spawned would accelerate the spawning instead of having two separate mechanisms that don't interact.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -0,0 +1,169 @@
...
+
+				if (counter >= growTime)
+					self.Kill(self);
+				else
+				{
+					var index = info.GrowthSequences.Length * counter / growTime;
+					anim.Animation.Play(info.GrowthSequences[index]);
+				}
+			}
+			else
```
pchote: Is this branch actually used?

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
Unit158: This code and the code you mentioned below are obsolete as I mentioned above.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
Unit158: Nope.

Mailaender:
Cleaned this up further. Fixed the crushing and adapted the balancing so that early (forced) explosion don't give you as much as pieces as fully grown once so you can only speed up, but not generate more by manual targeting.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -0,0 +1,169 @@
...
+						if (projectile != null)
+							self.World.Add(projectile);
+
+						if (args.Weapon.Report != null && args.Weapon.Report.Any())
+							Sound.Play(args.Weapon.Report.Random(self.World.SharedRandom), self.CenterPosition);
+					}
+				});
+			}
+
+			self.World.AddFrameEndTask(w => w.Add(new DelayedAction(respawnTicks, () =>
```
GraionDilach: It'd be nice if this would be optional, then this could be used for a crop logic - using minelayers for seeding.

Mailaender: This isn't really a general purpose trait designed to be re-usable. Better write your own. I also don't quite get what you have in mind here to be honest.

GraionDilach: Alright.

Unit158: You can actually already do that. Explodes, SelfHeals (negative heal) and a spreads resource warhead or whatever it's called.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -0,0 +1,168 @@
...
+					new CenterPositionInit(self.CenterPosition),
+					new OwnerInit(self.Owner),
+					new FactionInit(self.Owner.Faction.InternalName),
+					new SkipMakeAnimsInit()
+				};
+				self.World.CreateActor(self.Info.Name, td);
+			})));
+		}
+
+		public void WarnCrush(Actor crusher) { }
```
Unit158: This should be removed and replaced by the generic Crushable trait, no?

Mailaender: That didn't work out. As discussed in IRC that is quite infantry specific with `WarnCrush` overhead etc.

pchote: The `Crushable` trait works fine for walls.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
GraionDilach: Good point on SelfHeal, I totally forgot that could be used for this.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
Mailaender: Actually it doesn't. I made a ticket for it. https://github.com/OpenRA/OpenRA/issues/9011

pchote: Fixed by #9014.

Mailaender: Rebased and fixed the Devastator crush class while I was at it.

mods/d2k/weapons.yaml
```diff
@@ -717,8 +717,15 @@ Shrapnel:
 		ImpactSound: EXPLLG5.WAV
 
 SpiceExplosion:
+	Report: EXPLMD1.WAV
+	Projectile: Bullet
+		Speed: 50, 75
+		High: true
+		Angle: 91, 264
+		Trail: large_trail
+		Image: large_trail
```
Unit158: This may or may not want to be changed to the "null" sprite. I think It looks a little bit better like this, however.

abcdefg30:
Needs a rebase.

Mailaender:
Rebased.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -0,0 +1,154 @@
...
+			respawnTicks = self.World.SharedRandom.Next(info.RespawnDelay[0], info.RespawnDelay[1]);
+			growTicks = self.World.SharedRandom.Next(info.GrowthDelay[0], info.GrowthDelay[1]);
+			anim.Animation.Play(info.GrowthSequences[0]);
+		}
+
+		public void Tick(Actor self)
+		{
+			if (!self.IsDead)
+			{
+				ticks++;
```
pchote: I'm still not a big fan of having a separate tick counter for this.  Like I mentioned earlier, this means that there is no interaction between damage and the timer - they are both all or nothing.  If this instead decremented the health then shooting the spice bloom to half health would halve the spawn time, which I believe is more intuitive.

Unit158: Ok, but that isn't the original behaviour. This replicates the original behaviour (nearly) to a T, minus some heuristics that were included in the original to only spawn the spiceblooms when there was a certain amount of spice needed.

abcdefg30:
This causes weird glitches:
spicebloomglitch

penev92:
Regarding that glitch - @OmegaBolt mentioned on IRC yesterday that he's planning a PR to change the terrain type of sand decorations so that spice doesn't grow on them and replace them. Maybe this is related.

Mailaender:
Yes, those are just terrain definitions errors like https://github.com/OpenRA/OpenRA/issues/4243.

Mailaender:
Updated according to the review comments.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
pchote: Ok, fair enough.

pchote:
Adding [Changes Requested] for the MCV/Harv fixes, and needs a rebase.

Mailaender:
Rebased and fixed the crushable regressions.

abcdefg30:

reaperrr:
Needs rebase.

Mailaender:
Rebased.

reaperrr:
Sorry, needs another rebase.

Mailaender:
Rebased.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -0,0 +1,153 @@
...
+			render.Add(anim);
+
+			respawnTicks = self.World.SharedRandom.Next(info.RespawnDelay[0], info.RespawnDelay[1]);
+			growTicks = self.World.SharedRandom.Next(info.GrowthDelay[0], info.GrowthDelay[1]);
+			anim.Animation.Play(info.GrowthSequences[0]);
+		}
+
+		public void Tick(Actor self)
+		{
+			if (!self.IsDead)
```
penev92: The trait shouldn't tick if the actor is dead, so this should not be necessary.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -0,0 +1,153 @@
...
+				{
+					var index = info.GrowthSequences.Length * ticks / growTicks;
+					anim.Animation.Play(info.GrowthSequences[index]);
+				}
+			}
+		}
+
+		public void Killed(Actor self, AttackInfo e)
+		{
+			var weapon = self.World.Map.Rules.Weapons[info.Weapon.ToLowerInvariant()];
```
penev92: You can get rid of the local variable as it is only used once anyway.

mods/d2k/sequences/misc.yaml
```diff
@@ -334,19 +334,20 @@ crate:
 		ZOffset: -511
 		Offset: -16,-16
 
-# TODO: keep the redundant spicebloom.shp for now for the awful WinForms maps editor
```
penev92: Perhaps with the removal of this comment (and the long-gone WinForms map editor) you should remove that SHP?

penev92:
Also the actor that crushes the spicebloom is unharmed. Not sure if it should be instakilled or dealt a specific amount of damage, but there should be damaging involved.

Aside from that this looks OK. :+0.5: for the moment.

mods/d2k/sequences/misc.yaml
Mailaender: I already did that.

Mailaender:
Updated.

penev92:
Sorry this still doesn't harm the actor that crushes the spicebloom.

obrakmann:
Yep. The problem seems to be that the SpreadDamageWarhead is fired at the cells where the spice is spread to, not at the location where the spice bloom was.

penev92:
Added this to the Next release milestone because it will compliment the D2k rebalancing nicely and is almost ready IMO.

GraionDilach:
SpiceExplosion's warhead needs AffectsParent for the crusher to be harmed.

pchote:
Needs a rebase.

Mailaender:
Updated.

reaperrr:
There are still quite a few differences compared to the original, but I think this is now good enough to build upon. :+1:

pchote:
What are the differences?

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/0976c2833a4a6662620cff9005d16d12e78348e7...2253cc95eb70037f241edf9e0d1113a079d52ad7)

> What are the differences?

Spice debris were flying slower and on much lower angles (and more of them, I think) and spice blooms were only growing when there was no spice on that tile. They were growing slower, too.

I don't mind filing a follow-up myself, though.

reaperrr:
@Mailaender Actually, I didn't notice this before due to the large variance in Angle, but all fired spice debris target the same cell, something isn't right with the 'target' randomization.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -0,0 +1,148 @@
...
+
+			var pieces = self.World.SharedRandom.Next(info.Pieces[0], info.Pieces[1]) * ticks / growTicks;
+			for (var i = 0; pieces > i; i++)
+			{
+				var cells = OpenRA.Traits.Util.RandomWalk(self.Location, self.World.SharedRandom);
+				var cell = cells.Take(info.Range).SkipWhile(p => resLayer.GetResource(p) == resType && resLayer.IsFull(p)).Cast<CPos?>().RandomOrDefault(self.World.SharedRandom);
+				if (cell == null)
+					cell = cells.Take(info.Range).Random(self.World.SharedRandom);
+
+				args.PassiveTarget = self.World.Map.CenterOfCell(cell.Value);
```
pchote: This is why.  This gets reset each time through the loop, so all the projectiles will be spawned with the last value that is assigned.

Mailaender: Actually this is not the problem. This has to be set for each projectile. The randomization a few lines before just doesn't work, because `cell` is always null.

Issue #5884, opened 2014-07-08
Strange Render(Infantry) attack anim errors


reaperrr:
- [ ] 1) Sometimes the last frame will be omitted. As a result, the last frame with a muzzle flash is shown instead of the last frame of the animation (so if the attack anim is 6 frames long, frame 5 is shown at the end).
- [x] 2) Even after being ordered to stop, infantry tends to not go back into idle state, at least it doesn't go back to idle animations.

This screenshot shows both problems (the rifle infantry has been ordered to stop and actually stopped shooting quite a while ago):

![render-error](https://cloud.githubusercontent.com/assets/2857877/3512958/964e7924-06be-11e4-93b7-a8a896b9bb67.png)

For the record, it has been like that for ages, I just kept forgetting to report it.

penev92:
Can someone confirm this is still present issue after #7483 was merged?

reaperrr:
Issue 2) was fixed by #7090, but issue 1) still persists. Easily reproducable by letting TS rifle infantry or ghost stalker shoot to the right.

That being said, most other infantry is not affected, and the TS minigunner and stalker have identical sequences (number of frames), maybe it's something in the sequence code?

Phrohdoh:
RA2 Conscripts stuck on the last frame of `shoot` while they are actually idling:
![image](http://i.imgur.com/eFBUAOs.png)

matjaeck:
I can reproduce a similar error in playtest-20180729, but I'm not sure if this one is congruent with this issue, as the original comment indicates the error occurred regardless of an issued stop order:

![101](https://user-images.githubusercontent.com/37185497/44034500-8a3a90f0-9f0d-11e8-9826-553474b8f779.gif)

In this case, the stop order was issued during the frame(s) that shows the muzzle flash and the stop order caused to stop the sequence at this frame.

However, this does not always happen: Sometimes the "fixed" muzzle flash frame will only last for a couple of seconds after the unit stopped firing, then it goes over to the idle sequence. As the gif above shows, it is also possible to that the muzzle flash frame stays "fixed" (maybe only when it is stopped in the last frame?).

reaperrr:
Fixed by https://github.com/OpenRA/OpenRA/pull/15701.

PR #8988, opened 2015-08-08
Fix for WithAttackAnimation breaking WithMoveAnimation

### LINKED ISSUE #8977: Vehicle Move animation freeze after perform attack.
I have test TS mod on bleed version of openra and try mod something on it but I notice this problem which not only happen on my unit but also effect with official unit like Wolverine.

Unit move animation work fine until I order him to attack then their move animation become freezing and can't recover for entire game. 

img

[info]
- Unit still able to switch animation base on their activity .
- As if unit away stuck at first frame of move animation on each direction .
- My unit use similar code to Openra Wolverine.
- Wolverine also suffer from this glitch
- Not sure if Vehicle with turret also effect by this but Titan doesn't seem to effect with this bug..
- This bug not effect with Titan not sure if it effect only with non-turret vehicle.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25474654-vehicle-move-animation-freeze-after-perform-attack?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Closes #8977.

Test cases: TS wolverine (does it properly restore move animation after attack?), RA V2 launcher (does it still work as before?).

Fix works according to my own testing.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
@@ -49,7 +49,7 @@ namespace OpenRA.Mods.Common.Traits
 		public void Attacking(Actor self, Target target, Armament a, Barrel barrel)
 		{
 			if (!string.IsNullOrEmpty(info.AttackSequence))
-				wsb.PlayCustomAnimation(self, info.AttackSequence);
+				wsb.PlayCustomAnimation(self, info.AttackSequence, () => wsb.CancelCustomAnimation(self));
 		}
 
 		public void Tick(Actor self)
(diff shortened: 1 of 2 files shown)


pchote:
Does this affect anything in our shipping mods (that is, does this need a PR: For stable tag)?

penev92:
Only the RA V2 uses that trait, but since it doesn't have any other animations it should be unaffected.

Mailaender:
:+1: from https://github.com/OpenRA/OpenRA/issues/8977#issuecomment-129030350.

Mailaender:
Works as promised. :white_check_mark: It seems to also resolve https://github.com/OpenRA/OpenRA/issues/5884.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/bf07e57c7504d063822166538cc924cff8a054d1...6b9a00038c9b8bd31b507c785d20e0172bef16aa)

Issue #4164, opened 2013-11-21
Fix Helicopters shadows


x-a-n-a-x:
Here is an example of how helicopters shadows should look ingame (on the left):
![untitled-1](https://f.cloud.github.com/assets/1732773/1589508/b2ff3a6a-5271-11e3-864c-76692c9514e3.gif)

I think helicopter blades should blend with the helicopter body.

pchote:
Reverse-duplicate of #6019.

PR #8996, opened 2015-08-08
Added pre-multiplied alpha palettes for cloak effects

Mailaender:
This implements the cloak effect known from Tiberian Sun and also applies it to Red Alert. Remember the initial "what are the black holes in the ground" impression when we added cloaking to the camo pillbox.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Cloak.cs
+++ b/OpenRA.Mods.Common/Traits/Cloak.cs
@@ -34,7 +34,9 @@ namespace OpenRA.Mods.Common.Traits
 
 		public readonly string CloakSound = null;
 		public readonly string UncloakSound = null;
-		[PaletteReference] public readonly string Palette = "cloak";
+
+		public readonly string Palette = "cloak";
+		public readonly bool IsPlayerPalette = false;
 
 		public readonly HashSet<string> CloakTypes = new HashSet<string> { "Cloak" };
 
@@ -112,10 +114,11 @@ namespace OpenRA.Mods.Common.Traits
 
 			if (Cloaked && IsVisible(self, self.World.RenderPlayer))
 			{
-				if (string.IsNullOrEmpty(Info.Palette))
+				var palette = string.IsNullOrEmpty(Info.Palette) ? null : Info.IsPlayerPalette ? wr.Palette(Info.Palette + self.Owner.InternalName) : wr.Palette(Info.Palette);
+				if (palette == null)
 					return r;
 				else
-					return r.Select(a => a.WithPalette(wr.Palette(Info.Palette)));
+					return r.Select(a => a.WithPalette(palette));
 			}
 			else
 				return SpriteRenderable.None;
--- a/mods/d2k/rules/infantry.yaml
+++ b/mods/d2k/rules/infantry.yaml
@@ -135,6 +135,7 @@ fremen:
 		CloakDelay: 250
 		CloakSound: STEALTH1.WAV
 		UncloakSound: STEALTH2.WAV
+		IsPlayerPalette: true
 	-MustBeDestroyed:
 	Voiced:
 		VoiceSet: FremenVoice
--- a/mods/d2k/rules/palettes.yaml
+++ b/mods/d2k/rules/palettes.yaml
@@ -26,12 +26,6 @@
 		G: 0
 		B: 0
 		A: 140
-	PaletteFromRGBA@cloak:
-		Name: cloak
-		R: 0
-		G: 0
-		B: 0
-		A: 140
 	PaletteFromRGBA@highlight:
 		Name: highlight
 		R: 255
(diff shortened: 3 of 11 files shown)


pchote:
I've already explained on multiple occasions why this is not a feasible approach for TS (the sprites need to be blended while opaque and the opacity applied to the result).  Countering my comments with a PR that adds this to both TS *and* RA feels deliberately malicious.

Mailaender:
I am just going to ignore the mindless personal attack. You didn't make the problem as clear as you probably thought you had. Extended this to the `d2k` mod where that effect is already in use. Still fail to see any breakage. Can you maybe post a screenshot?

penev92:
So as far as I understand (I didn't pay much attention at the time of the original PR), stacking sprites somewhat counter the transparency effect. I think the concern here is that whether or not you've managed to get TS to look OK, the engine hasn't changed for the better (that's just YAML changes) and we can't guarantee that thirdparty mods will be OK.

GraionDilach:
@penev92: This is the same issue the RA1 chinook rotor shadows are having - see #4164, but with transparency instead of darkening practically.

Mailaender:
In contrast to the fixed black color palette, using pre-multiplied alpha does not only look better, but it also allows us to hide this engine limitation a bit by choosing a very low value for the alpha percentage so that actors still look cloaked even when the sprites are stacked.

pchote:
I suggest that you re-read the explanation in the previous PR, which I consider to be a clear explanation.  If you don't understand the technical details, then please ask me specific questions.

If other people want a quick but insufficient band-aid then so be it, but this really kills my enthusiasm to work on the actual underlying issue.

Mailaender:
I suggest you open a new ticket describing the problem and your proposed solutions then so someone else can work on it in the future.

Mailaender:
I just did some research and the original seemed to have the same problem with sprite stacking undoing parts of the transparency most notably for the construction yard crane:

image

GraionDilach:
That's true, YR had the DoubleThick tag on sequences which meant that the animation should be rendered 0.5 less transparent when transparent, as a bandaid.

abcdefg30:
Needs a rebase.

Phrohdoh:
If this isn't going to be accepted by the person who will do it properly eventually then there is no need to ask the author to update it.

abcdefg30:
The code doesn't look to bad to me, and @pchote didn't express any interest in working on this.
"So I think this is better than nothing."

Mailaender:
Rebased.

I guess adding this feature in the same limited way of the original is a first step in improving on it.

GraionDilach:
Woah, what happened with Travis?! D:

Mailaender:
I resolved the rebase conflict in the wrong direction. Fixed.

GraionDilach:
Since Travis passed, and we already have the similar issue with shadows stacking, I think this can get my (y).

abcdefg30:
You forgot about the `phase tank` (`stnk`).
`Exception of type "System.InvalidOperationException": Palette "cloak" does not exist`

penev92:
I've only tested this in TS, and won't comment on RA, but I think this is out of place in D2k. Our current effect is much closer to what Dune 2000 had. (I'm not sure it suits RA either).
Also 0.25 is way too low for TS.

Mailaender:
Fixed the phase transport, adjusted the alpha value for TS and fixed remap issues.

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -34,7 +34,9 @@ public class CloakInfo : UpgradableTraitInfo
 
 		public readonly string CloakSound = null;
 		public readonly string UncloakSound = null;
-		[PaletteReference] public readonly string Palette = "cloak";
+
+		public readonly string Palette = "cloak";
+		public readonly bool IsPlayerPalette = false;
```
Phrohdoh: These need `desc`s.

Mailaender:
Rebased.

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -34,7 +34,9 @@ public class CloakInfo : UpgradableTraitInfo
 
 		public readonly string CloakSound = null;
 		public readonly string UncloakSound = null;
-		[PaletteReference] public readonly string Palette = "cloak";
+
+		public readonly string Palette = "cloak";
```
Phrohdoh: Doesn't this have some sort of lint attrib?

Mailaender: We don't have https://github.com/OpenRA/OpenRA/issues/9101 yet.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/74077bedcb9e825f0b6a67b9cd4a4057c13b3d60...88ea7e9aa98cdea79a16090fef71bcdc8468a49b)

PR #9002, opened 2015-08-08
Add an etiquette notice to our guidelines.

Unit158:
This may not be necessary, but it is a worthwhile reminder if people do read this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,7 @@
 # OpenRA Contributing Guidelines
 
+### Please stay calm and polite when submitting or commenting on a bug or patch.
+
 ## Bug reports
 
 * Have you read the [FAQ](https://github.com/OpenRA/OpenRA/wiki/FAQ)?


deniz1a:
What if I'm an amateur?

Unit158:
We've already had this discussion on IRC.

deniz1a:
So, amateur programmers should not file bug reports?

Phrohdoh:
@deniz1a you know this discussion has already happened, please do not duplicate it here for no gain.
If you feel it is important simply copy and paste from the logs into a code-block (```).

CONTRIBUTING.md
```diff
@@ -8,6 +8,7 @@
 * Please be specific on how to reproduce the problem.
 * Do not submit duplicate issues. Search first.
 * Avoid including multiple requests/bugs into a single issue.
+* Stay calm and professional when submitting a bug. Bad attitude will drive developers away.
```
Mailaender: Maybe remove the second sentence to avoid breaking the scheme.

pchote: I agree with @Mailaender.
You could also add "or commenting on a pull request" to the end of the first sentence.

Mailaender: In that case maybe even move this up so it includes Bugs and Patches.

Unit158:
Updated.

abcdefg30:
So, is there consensus in adding this?

> 21:20 (+abcdefg30) hm, am i the only one who thinks that https://github.com/Unit158/OpenRA/blob/patch-1/CONTRIBUTING.md#please-stay-calm-and-polite-when-submitting-or-commenting-on-a-bug-or-patch doesn't look good on two lines?
21:20 (+hamb) wow, we had to put that in contributing? bummer
21:21 (+Phroh|work) hamb: note that that isn't upstream
21:21 (+hamb) but probably soon to be?
21:22 (+Phroh|work) it shouldn't be imo
21:22 (+Phroh|work) it is a bit of a bummer, and if someone is going to rage they're going to do so either way
21:22 (+hamb) it's kind of a given on courtesy and i agree with you
21:23 (+hamb) rather i assume that message is there to target some people who never followed instructions anyway? ;)

penev92:
"... when typing." would fix the two-line issue.
But I two don't see this helping for the reasons @Phrohdoh gave. :/

Unit158:
I would close this if I could.

PR #9004, opened 2015-08-08
Introduce QuantizeFacingsFromSequence

### LINKED ISSUE #8889: Decouple facing/orientation quantization from render traits
This is necessary to allow an actor having multiple `With*Body` traits (for purposes like upgrading, deploying etc)., and therefore prerequisite for implementing deployable TS units (tick tank, sensor array, artillery etc) via `DeployToUpgrade`.

While not strictly a requirement, #8848 would make it a lot easier to implement this in a robust way.
Render traits don't necessarily require `IQuantizeBodyOrientation`, but in the presence of a trait that implements it they need to be loaded after that trait.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25281375-decouple-facing-orientation-quantization-from-render-traits?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
... and related changes.
Makes it possible to have multiple `With*Body` traits on a single actor.

Closes #8889.

### DIFF
--- a/OpenRA.Mods.Common/Traits/QuantizeFacingsFromSequence.cs
+++ b/OpenRA.Mods.Common/Traits/QuantizeFacingsFromSequence.cs
@@ -1,0 +1,43 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Derive facings from sprite body sequence.")]
+	public class QuantizeFacingsFromSequenceInfo : UpgradableTraitInfo, IQuantizeBodyOrientationInfo, Requires<RenderSpritesInfo>
+	{
+		[Desc("Defines sequence to derive facings from."), SequenceReference]
+		public readonly string Sequence = "idle";
+
+		public int QuantizedBodyFacings(ActorInfo ai, SequenceProvider sequenceProvider, string race)
+		{
+			if (string.IsNullOrEmpty(Sequence))
+				throw new InvalidOperationException("Actor " + ai.Name + " is missing sequence to quantize facings from.");
+
+			var rsi = ai.Traits.Get<RenderSpritesInfo>();
+			return sequenceProvider.GetSequence(rsi.GetImage(ai, sequenceProvider, race), Sequence).Facings;
+		}
+
+		public override object Create(ActorInitializer init) { return new QuantizeFacingsFromSequence(this); }
+	}
+
+	public class QuantizeFacingsFromSequence : UpgradableTrait<QuantizeFacingsFromSequenceInfo>
+	{
+		public QuantizeFacingsFromSequence(QuantizeFacingsFromSequenceInfo info)
+			: base(info) { }
+	}
+}
--- a/OpenRA.Game/OpenRA.Game.csproj
+++ b/OpenRA.Game/OpenRA.Game.csproj
@@ -191,7 +191,6 @@
     <Compile Include="Graphics\VoxelRenderer.cs" />
     <Compile Include="Graphics\VoxelLoader.cs" />
     <Compile Include="Graphics\VoxelProvider.cs" />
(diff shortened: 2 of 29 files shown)


reaperrr:
Note: I kept the yaml changes in a separate commit for now to make this a little easier to review.

Haven't written an upgrade rule yet, since I first want some feedback on this approach.

reaperrr:
Updated.

penev92:
```
Object reference not set to an instance of an object.
  Source=OpenRA.Mods.Common
  StackTrace:
       at OpenRA.Mods.Common.Traits.BodyOrientation.<>c__DisplayClass4_1.<.ctor>b__0()
       at System.Lazy`1.CreateValue()
```
When building a Nod Airstrip in TD.
`qboi` is `null` and `isb` is `false`.

penev92:
Also got this in D2k:
```
Additional information: Unit `rifle` does not have a sequence named `idle`
```

I think it happens when they try to fire, because it's Lazy and doesn't get initialized beforehand.

reaperrr:
TD airstrip crash: WithIdleOverlay needs facing information (since it's not just for buildings). Actually it's better for buildings to have QFFS or
`
BodyOrientation:
   QuantizedFacings: 1
`
anyway, I just forgot to set that for TD.

D2k crash: just happened because I forgot to set `Sequence: stand` on QFFS since infantry has no `idle` sequence.

reaperrr:
Fixed crashes and split yaml changes for each mod.

These need to be squashed into the first commit before merging, though

OpenRA.Mods.Common/Traits/QuantizeFacingsFromSequence.cs
```diff
@@ -0,0 +1,43 @@
...
+		[Desc("Defines sequence to derive facings from."), SequenceReference]
+		public readonly string Sequence = "idle";
+
+		public int QuantizedBodyFacings(ActorInfo ai, SequenceProvider sequenceProvider, string race)
+		{
+			if (string.IsNullOrEmpty(Sequence))
+				throw new InvalidOperationException("Actor " + ai.Name + " is missing sequence to quantize facings from.");
+
+			var rsi = ai.Traits.Get<RenderSpritesInfo>();
+			return sequenceProvider.GetSequence(rsi.GetImage(ai, sequenceProvider, race), Sequence).Facings;
```
penev92: Won't this get the number of facings for the actor's "idle" sequence? (unless the default is overridden, of course)
But in any case, we need the number of facings to be able to change depending on the currently used sequence.
Perhaps the sequence name needs to be passed from outside? So that callers can specify what sequence they want to know about?

P.S.: This is only used for syncing?!?  (and ActorPreviews)

pchote: > But in any case, we need the number of facings to be able to change depending on the currently used sequence.

That's a significant scope creep, and i'm not convinced we need it.

abcdefg30:
Needs a rebase, too.

OpenRA.Mods.Common/Traits/QuantizeFacingsFromSequence.cs
penev92: Well now that I've seen where it's used I'm not convinced as well.

reaperrr:
Updated and rebased.

Added upgrade rule.

OpenRA.Mods.Common/Traits/QuantizeFacingsFromSequence.cs
reaperrr: > Won't this get the number of facings for the actor's "idle" sequence? (unless the default is overridden, of course)

Yes, and that's the idea since `idle` is the most common sequence and usually has the correct number of facings for that actor.

penev92:
:+1:

Mailaender:
I guess this will make modding a bit harder as stuff won't work out of the box anymore, but that can be solved with documentation. Can't spot regressions. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5064e96a77cbf5d3d3ef0acdcdd10e44d2ccb4b2...32719840f55a3055ab344add590d62c4a2e5e708)

GraionDilach:
@Mailaender It's never a bad sign to have tags/traits which involve RTFM. Believe me, such intelligence checks are quite beneficial for everyone - developers have more chance to spot people incapable of reading that and wasting developer time with their demand to be spoonfed instead.

Sorry if that comes harsh though. But I pretty much welcome every obstacle with such a result.

PR #9006, opened 2015-08-09
Remove legacy 0% = not targetable assumption

reaperrr:
None of the first-party mods use the above assumption anymore as far as I could tell, so we might as well remove it now.

### DIFF
--- a/mods/d2k/weapons.yaml
+++ b/mods/d2k/weapons.yaml
@@ -57,6 +57,8 @@ Sniper:
 	ReloadDelay: 100
 	Range: 8c512
 	Report: FREMODD1.WAV
+	ValidTargets: Ground, Infantry
+	InvalidTargets: Vehicle, Structure, Wall
 	Projectile: Bullet
 		Speed: 1c896
 		ContrailLength: 6
@@ -66,19 +68,14 @@ Sniper:
 	Warhead@1Dam: SpreadDamage
 		Spread: 32
 		Damage: 60
-		Versus:
-			None: 100
-			Wood: 0
-			Light: 1
-			Heavy: 0
-			Concrete: 0
+		ValidTargets: Infantry
 		DamageTypes: Prone50Percent, TriggerProne, BulletDeath
 
 Vulcan:
 	ReloadDelay: 30
 	Range: 5c768
 	Report: VULCAN.AUD
-	ValidTargets: Ground
+	InvalidTargets: Structure
 	Projectile: Bullet
 		Speed: 1c256
 		ContrailLength: 3
@@ -88,12 +85,10 @@ Vulcan:
 	Warhead@1Dam: SpreadDamage
 		Spread: 96
 		Damage: 30
-		ValidTargets: Ground
+		InvalidTargets: Structure
 		Versus:
-			Wood: 0
 			Light: 60
 			Heavy: 10
-			Concrete: 0
 		DamageTypes: Prone50Percent, TriggerProne, BulletDeath
 	Warhead@2Eff: CreateEffect
 		Explosion: piffs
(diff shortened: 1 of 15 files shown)


Mailaender:
Can you make the utility spit out a warning on upgrade?

GraionDilach:
You should remove the 0% verses if they are unused anyway.
https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/weapons/smallcaliber.yaml#L10
https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/weapons/smallcaliber.yaml#L290
https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/weapons/smallcaliber.yaml#L341
https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/weapons/other.yaml#L189

Also MadTank seemingly use it
https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/weapons/other.yaml#L281

reaperrr:
> You should remove the 0% verses if they are unused anyway.

My bad, I messed up when searching for these cases (I forgot we use integer math now and no longer use the '%').

I'll update this later.

reaperrr:
Updated.

mods/d2k/weapons.yaml
```diff
@@ -66,19 +67,14 @@ Sniper:
...
-			Concrete: 0
+		ValidTargets: Infantry
 		DamageTypes: Prone50Percent, TriggerProne, BulletDeath
 
 Vulcan:
 	ReloadDelay: 30
 	Range: 5c768
 	Report: VULCAN.AUD
-	ValidTargets: Ground
+	InvalidTargets: Structure
```
GraionDilach: Wouldn't this render the weapon anti-air capable?

reaperrr: `ValidTargets` defaults to `Ground` and `Water`, you'll find that many normal weapons don't explicitly define `ValidTargets` at all, so that entry was simply redundant (though you could argue it removed `Water` since that doesn't exist in d2k).

penev92: Yay for random default values, eh?

reaperrr: *shrugs* Legacy from when OpenRA was just RA, I guess. Might be worth dealing with at some point, but not in this PR.

abcdefg30:
:+1:

abcdefg30:
Oh wait, some units like `Tanya`, the `Sniper` or the `Spy` can't force fire on the ground anymore.

reaperrr:
Updated.

Fortunately that was just a yaml issue, the `Ground` target type must be valid to allow force-firing on ground.

mods/d2k/weapons.yaml
```diff
@@ -88,12 +85,10 @@ Vulcan:
 	Warhead@1Dam: SpreadDamage
 		Spread: 96
 		Damage: 30
-		ValidTargets: Ground
+		InvalidTargets: Buildings
```
GraionDilach: It's Buildings or Structure?

penev92: Looks like it's "Structure" everywhere else.

GraionDilach:
:+1: after my above nit gets fixed.

mods/d2k/weapons.yaml
reaperrr: Good catch, fixed.

reaperrr:
Updated, can be merged now.

penev92:
Changelog, anyone?
I feel this is somewhat important for modders to know about...

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/82edd49b510786fedfbc7c7c4ccc01bbd41caa4e...f80567f63accba61f29b991376abe393a5b2c3e7)

PR #9007, opened 2015-08-09
Made Armor upgradable

reaperrr:
This allows to change armor on levelup, and even add additional armor types.

### DIFF
--- a/OpenRA.Mods.Common/Warheads/DamageWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/DamageWarhead.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System.Collections.Generic;
+using System.Linq;
 using OpenRA.Mods.Common.Traits;
 using OpenRA.Traits;
 
@@ -34,13 +35,13 @@ namespace OpenRA.Mods.Common.Warheads
 				: new Dictionary<string, int>();
 		}
 
-		public int DamageVersus(ActorInfo victim)
+		public int DamageVersus(Actor victim)
 		{
-			var armor = victim.Traits.GetOrDefault<ArmorInfo>();
-			if (armor != null && armor.Type != null)
+			var armor = victim.TraitsImplementing<Armor>().Where(a => !a.IsTraitDisabled && a.Info.Type != null);
+			foreach (var a in armor)
 			{
 				int versus;
-				if (Versus.TryGetValue(armor.Type, out versus))
+				if (Versus.TryGetValue(a.Info.Type, out versus))
 					return versus;
 			}
 
@@ -60,7 +61,7 @@ namespace OpenRA.Mods.Common.Warheads
 
 		public virtual void DoImpact(Actor victim, Actor firedBy, IEnumerable<int> damageModifiers)
 		{
-			var damage = Util.ApplyPercentageModifiers(Damage, damageModifiers.Append(DamageVersus(victim.Info)));
+			var damage = Util.ApplyPercentageModifiers(Damage, damageModifiers.Append(DamageVersus(victim)));
 			victim.InflictDamage(firedBy, damage, this);
 		}
 	}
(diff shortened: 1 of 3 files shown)


reaperrr:
Note: While not strictly a requirement, it would be better if #9006 was merged first so I can get rid of the `CanTargetActor` work-arounds.

Mailaender:
You could set up the crates. The animation sequences have already been identified:

* https://github.com/OpenRA/OpenRA/blob/bleed/mods/cnc/sequences/misc.yaml#L168
* https://github.com/OpenRA/OpenRA/blob/bleed/mods/ts/sequences/misc.yaml#L52
* https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/sequences/misc.yaml#L312

reaperrr:
> You could set up the crates. The animation sequences have already been identified:

The 'armor' crates in the originals simply halved the damage taken, which is already possible (and easier to do) with the `DamageMultiplier` trait.

I don't mind setting up armor crates, but not as part of this PR.

pchote:
FYI we used to have the separate stat crates in RA and TD, but we decided to replace them all with the veterancy crate because individual stat upgrades don't really make sense for us (remember: the original games didn't have any veterancy).

GraionDilach:
TS/RA2 had both. The crate-given individual stat was bigger than the stat one would gain from ranking.

reaperrr:
Rebased.

OpenRA.Mods.Common/Warheads/DamageWarhead.cs
```diff
@@ -34,13 +35,13 @@ public static object LoadVersus(MiniYaml yaml)
...
 
-		public int DamageVersus(ActorInfo victim)
+		public int DamageVersus(Actor victim)
 		{
-			var armor = victim.Traits.GetOrDefault<ArmorInfo>();
-			if (armor != null && armor.Type != null)
+			var armor = victim.TraitsImplementing<Armor>().Where(a => !a.IsTraitDisabled && a.Info.Type != null);
+			foreach (var a in armor)
 			{
 				int versus;
```
GraionDilach: Move the creation of this out from the foreach, with a default value of 100 and multiply it here to let more armors stack with each other, also move the return out from the foreach too.

reaperrr: I've now wasted 4 hours trying to make this work with no success, so unless someone else comes up with some concrete, working code, there's no telling when this will be finished.

reaperrr:
I don't know how to handle IEnumerable and Versus.TryGetValue in order to make armor effects stack. Wasted over 5 hours so far without success.

GraionDilach:
Give me tomorrow.

GraionDilach:
Filed https://github.com/reaperrr/OpenRA/pull/9.

reaperrr:
> Filed reaperrr#9.

Looks so simple...
Thanks, updated.

Also updated test case (disabled veterancy damage modifier at first two rank-ups, so any changes to received damage are entirely caused by the armor).

reaperrr:
Explanation on test case for reviewers:

The correct behavior when shooting at the allied jeep should be

- medium/heavy tank deal 75% of max. damage before rank-up, 100% after both 1st and 2nd rank-up (at 2nd rank-up, the jeep has two 'layers' of heavy armor, but cannons have 100% versus heavy, so 100% * 100% = 100%).

- light tank does 100% damage befor rank-up, 40% after 1st rank-up, 16% after 2nd rank-up (40% * 40%).

GraionDilach:
I don't know how my :+1: would be valid after that patch... but yea, I say :+1:

penev92:
This removes the `.TryGetValue()` safety net, but I'm OK with that.
Works :+1:

reaperrr:
> I don't know how my :+1: would be valid after that patch... but yea, I say :+1:

Well, I reviewed and tested your patch, and I assume you tested it as well, so in my opinion this PR has +2 already.

penev92:
Good enough for me. I'll merge when you remove the test commits.

reaperrr:
Test case removed.

penev92:
Thanks.

PR #9016, opened 2015-08-09
TS vanilla rocket explosions

OmegaBolt:
-- Uses vanilla TS explosions when impacting the ground
-- All use twlt variants in the air with their vanilla sound effects (this is not vanilla but "<+Mailaender> It think it is required to indicate hits on aircraft." and it's not a bad idea)

### DIFF
--- a/mods/ts/weapons/missiles.yaml
+++ b/mods/ts/weapons/missiles.yaml
@@ -35,9 +35,9 @@ Bazooka:
 		ImpactSound: ssplash3.aud
 		ValidImpactTypes: Water
 	Warhead@4EffAir: CreateEffect
-		Explosion: small_twlt
-		ImpactSound: expnew12.aud
 		ValidImpactTypes: Air, AirHit
+		Explosion: tiny_twlt
+		ImpactSound: expnew05.aud
 	Warhead@5Smu: LeaveSmudge
 		SmudgeType: SmallCrater
 
@@ -80,8 +80,8 @@ HoverMissile:
 		ValidImpactTypes: Water
 	Warhead@4EffAir: CreateEffect
 		Explosion: small_twlt
-		ImpactSound: expnew12.aud
 		ValidImpactTypes: Air, AirHit
+		ImpactSound: expnew06.aud
 	Warhead@5: LeaveSmudge
 		SmudgeType: SmallCrater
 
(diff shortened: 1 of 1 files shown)


GraionDilach:
Wait... this doesn't work that way. Nw Effect2 will be only played when the missile hit a ground __target__ alone.

OmegaBolt:
What are you refering to? Previously the air impacts overrode the supposed ground explosions, now they only happen in the air.

GraionDilach:
Oh wait, I got it wrong, although I'm worried that this is still buggy. If only one effect is played from a warhead then now you broke water impacts instead of ground.

mods/ts/weapons/missiles.yaml
```diff
@@ -35,9 +35,9 @@ Bazooka:
 		ImpactSound: ssplash3.aud
 		ValidImpactTypes: Water
 	Warhead@4EffAir: CreateEffect
-		Explosion: small_twlt
-		ImpactSound: expnew12.aud
-		ValidImpactTypes: Air, AirHit
+		Explosion: tiny_twlt
+		ImpactSound: expnew05.aud
+		InvalidImpactTypes: Ground
```
reaperrr: This doesn't look right to me, with the original setup, the explosion should only display when exploding in the air. With this change, it would play when hitting ground targets or anything on water as well.

If you say that's not the case, this appears to be a code-level bug.

OmegaBolt: Just play the TS in bleed. The air anim plays whenever it hits anything on the ground too.

reaperrr: Then something is broken in the code.

reaperrr:
@OmegaBolt Sorry, but this PR doesn't fix the actual issue.

The real issue is that the Missile projectile code still makes the outdated `pos.Z < 0` check for "hitting ground", which doesn't factor in height levels. #8717 will fix that.

Your other changes (replacing `small_twlt` with `tiny_twlt` and so on) are fine, but you have to remove the changes to `ImpactTypes` before we can merge this.

pchote:
Any progress here @OmegaBolt?

OmegaBolt:
Do you still want this then, with just the animation changes?

reaperrr:
Sure, as I said:
> Your other changes (replacing small_twlt with tiny_twlt and so on) are fine, but you have to remove the changes to ImpactTypes before we can merge this.

abcdefg30:
:+1:

reaperrr:
:+1:

Issue #8603, opened 2015-06-29
TD Ranking


AoAGeneral:
Ranking I think is a good addition but the ranking is way to powerful. Rank 1 and 2 in example for recon bikes and infantry causing them to move way to fast for any vehicle to catch up to them. At max rank bike they move faster then the A10 planes.

Suggestion: Make small gains for ranking rather then big upgrades

Unit158:
Another idea: We could change how things are upgraded when they rank up. ATM, most "AA" weapons are too slow to be very effective versus air, so we could see about improving the weapon when the unit ranks up, and reduce the speed (for example).

AoAGeneral:
This could be a good idea. However, some people may prefer to have faster speed and regret the rank up system instead.

penev92:
Addressed in #9017.

PR #9017, opened 2015-08-09
Rebalanced stat gains from rank-up

reaperrr:
It was mentioned on IRC that the first, relatively powerful rank-up allows jeeps, for example, to safely bring infantry like spies and engineers to enemy buildings.

This is mainly because the firepower, armor and speed bonuses are about 10% each on first rank-up, whereas the next two, harder to reach rank-ups only add 4-5%.

This lowers the initial buffs, generally flattens the 'armor' curve for level-ups, and lowers the maximum bonus for armor and speed.

### DIFF
--- a/mods/cnc/rules/defaults.yaml
+++ b/mods/cnc/rules/defaults.yaml
@@ -17,13 +17,13 @@
 			1600: firepower, damage, speed, reload, inaccuracy, rank, eliteweapon, selfheal
 	FirepowerMultiplier@EXPERIENCE:
 		UpgradeTypes: firepower
-		Modifier: 110, 115, 120, 130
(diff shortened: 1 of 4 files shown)


Mailaender:
:+1: I can follow the reasoning. Leveled up yaks can also out-run SAM missiles etc.

pchote:
Fixes #8603.

penev92:
Seems legit :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c811593fe0400ab73c96f5546a2603c1543b6829...a8017c81acbfd688120d6de89209773fc9f2f7f0)

Issue #8932, opened 2015-08-04
DeployToUpgrade does not provide any animations or sounds.


Unit158:

penev92:
[Proof of the oposite](http://i.imgur.com/pBfXvvA.gif).

Unit158:
Note: the above is not currently in bleed.

Mailaender:
So are you going to add it? At the moment `DeployToUpgrade` is completely unused.

Mailaender:
I did experiments https://github.com/Mailaender/OpenRA/compare/change-sequence some time ago.

PR #9019, opened 2015-08-09
Added the Thumper infantry

Mailaender:
See http://dune2k.com/Duniverse/Games/Dune2000/Details/Units for reference.

### DIFF
--- a/OpenRA.Mods.D2k/Traits/Sandworm.cs
+++ b/OpenRA.Mods.D2k/Traits/Sandworm.cs
@@ -99,7 +99,8 @@ namespace OpenRA.Mods.D2k.Traits
 			targetCountdown = Info.TargetRescanInterval;
 
 			// If close enough, we don't care about other actors.
-			var target = self.World.FindActorsInCircle(self.CenterPosition, Info.IgnoreNoiseAttackRange).FirstOrDefault(x => x.HasTrait<AttractsWorms>());
+			var target = self.World.FindActorsInCircle(self.CenterPosition, Info.IgnoreNoiseAttackRange)
+				.FirstOrDefault(x => attackTrait.HasAnyValidWeapons(Target.FromActor(x)));
 			if (target != null)
 			{
 				self.CancelActivity();
--- a/OpenRA.Mods.D2k/Traits/Thumper.cs
+++ b/OpenRA.Mods.D2k/Traits/Thumper.cs
@@ -1,0 +1,122 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Mods.Common.Orders;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.D2k.Traits
+{
+	[Desc("Lures the `Sandworm` to its position when deployed. To be used in conjunction with an `AttractsWorm` upgrade.")]
+	class ThumperInfo : ITraitInfo, Requires<MobileInfo>, Requires<UpgradeManagerInfo>
+	{
+		[SequenceReference(null, true)] public readonly string UndeployedPrefix = "undeployed-";
+
+		public readonly int ThumpInterval = 40;
+		public readonly string ThumpSound = "THUMPER1.WAV";
+
+		public readonly string[] DeployTerrainTypes = { "Sand", "Dune", "Spice" };
+		public readonly string DeployCursor = "deploy";
+		public readonly string DeployBlockedCursor = "deploy-blocked";
+
+		[VoiceReference] public readonly string Voice = "Action";
+
+		[UpgradeGrantedReference]
+		[Desc("The upgrade to grant when deploying and revoke when undeploying.")]
+		public readonly string Upgrade = "WormAttraction";
+
+		public object Create(ActorInitializer init) { return new Thumper(init.Self, this); }
+	}
+
+	class Thumper : ITick, IIssueOrder, IResolveOrder, IOrderVoice, IRenderInfantrySequenceModifier
+	{
+		readonly Actor self;
+		readonly ThumperInfo info;
+		readonly Mobile mobile;
+		readonly UpgradeManager manager;
+
+		bool deployed;
+		bool isUpgraded;
+		int tick;
+
+		public bool IsModifyingSequence { get { return !deployed; } }
+		public string SequencePrefix { get { return info.UndeployedPrefix; } }
+
+		public Thumper(Actor self, ThumperInfo info)
+		{
+			this.self = self;
+			this.info = info;
+			mobile = self.Trait<Mobile>();
+			manager = self.Trait<UpgradeManager>();
+		}
+
+		public void Tick(Actor self)
+		{
+			if (!deployed)
+				return;
+
+			if (++tick >= info.ThumpInterval)
+			{
+				tick = 0;
+				Sound.Play(info.ThumpSound, self.CenterPosition);
+			}
+		}
+
+		bool CanDeploy()
+		{
+			var terrainType = self.World.Map.GetTerrainInfo(self.Location).Type;
+			return info.DeployTerrainTypes.Contains(terrainType) && !mobile.IsMoving;
+		}
+
+		public IEnumerable<IOrderTargeter> Orders
+		{
+			get
+			{
+				yield return new DeployOrderTargeter("DeployThumper", 5,
+					() => CanDeploy() ? info.DeployCursor : info.DeployBlockedCursor);
+			}
+		}
+
+		public Order IssueOrder(Actor self, IOrderTargeter order, Target target, bool queued)
+		{
+			if (order.OrderID == "DeployThumper")
+				return new Order(order.OrderID, self, queued);
+
+			return null;
+		}
+
+		public string VoicePhraseForOrder(Actor self, Order order)
+		{
+			return (order.OrderString == "DeployThumper") ? info.Voice : null;
+		}
+
+		public void ResolveOrder(Actor self, Order order)
+		{
+			if (!CanDeploy())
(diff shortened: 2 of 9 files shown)


OpenRA.Mods.D2k/Traits/Sandworm.cs
```diff
@@ -71,7 +71,13 @@ public override void OnBecomingIdle(Actor self)
 			if (withSpriteBody.DefaultAnimation.CurrentSequence.Name != Info.IdleSequence)
 				withSpriteBody.DefaultAnimation.PlayRepeating(Info.IdleSequence);
 
-			base.OnBecomingIdle(self);
+			// The thumper lures the worms, but doesn't attract it.
```
Phrohdoh: The difference between luring and attracting is..?

OpenRA.Mods.D2k/Traits/Thumper.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Linq;
+using OpenRA.Mods.Common.Orders;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.D2k.Traits
+{
+	class ThumperInfo : ITraitInfo, Requires<MobileInfo>
+	{
+		[SequenceReference(null, true)] public readonly string UndeployedPrefix = "undeployed-";
```
Phrohdoh: These fields need `Desc`s.

OpenRA.Mods.D2k/Traits/Sandworm.cs
Mailaender: I added a description to the trait to make it more clear for everyone.

OpenRA.Mods.D2k/Traits/Thumper.cs
```diff
@@ -0,0 +1,98 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Mods.Common.Orders;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.D2k.Traits
+{
+	[Desc("Lures the `Sandworm` to it's position when deployed, but doesn't attract it to avoid becoming a target.")]
```
pchote: Did you try to fix this targeting issue?  That would be a much better solution than hacking around it.

ScottNZ: Grammar nit: it's -> its

Mailaender: There is no "targeting issue". This is just describing the workings of the trait.

OpenRA.Mods.D2k/Traits/Sandworm.cs
```diff
@@ -71,7 +71,12 @@ public override void OnBecomingIdle(Actor self)
 			if (withSpriteBody.DefaultAnimation.CurrentSequence.Name != Info.IdleSequence)
 				withSpriteBody.DefaultAnimation.PlayRepeating(Info.IdleSequence);
 
-			base.OnBecomingIdle(self);
+			var thumper = self.World.FindActorsInCircle(self.CenterPosition, Info.MaxSearchRadius)
```
pchote: This doesn't account for proximity or multiple thumpers, so the worm will behave strangely if you have more than one in an area.

pchote: This is the hack that I was referring to in the other comment thread: We built the attraction code specifically to gracefuly handle situations like this.  The attraction does not need to be tied directly to targeting, and so fixing that would produce cleaner code and more consistent behaviour than this "pick a thumper and move".

pchote: This is also easy to do: add `&& attackTrait.HasAnyValidWeapons(Target.FromActor(x))` to the `FirstOrDefault` in the target check, and then give it a target type that the worm isn't valid against.

pchote: This block of code can then disappear.

OpenRA.Mods.D2k/Traits/Thumper.cs
```diff
@@ -0,0 +1,98 @@
...
+		public readonly string[] DeployTerrainTypes = { "Sand", "Dune", "Spice" };
+		public readonly string DeployCursor = "deploy";
+		public readonly string DeployBlockedCursor = "deploy-blocked";
+
+		[VoiceReference] public readonly string Voice = "Action";
+
+		public object Create(ActorInitializer init) { return new Thumper(init.Self, this); }
+	}
+
+	class Thumper : IRenderInfantrySequenceModifier, ITick, IIssueOrder, IResolveOrder, IOrderVoice
```
pchote: You can then avoid almost all of the duplication in this trait by making it subclass `DeployToUpgrade` and add the `IRenderInfantrySequenceModifier` and sound code on top.  The `AttractsWorms` trait already supports using upgrades to switch to a version with much higher attraction when deployed.

OpenRA.Mods.D2k/Traits/Sandworm.cs
pchote: You could make the target selection more openra-ey by removing the `x.HasTrait<AttractsWorms>()` check and using target types exclusively:

```csharp
// If close enough, we don't care about other actors.
var target = self.World.FindActorsInCircle(self.CenterPosition, Info.IgnoreNoiseAttackRange)
	.FirstOrDefault(x => attackTrait.HasAnyValidWeapons(Target.FromActor(x)));
```

OpenRA.Mods.D2k/Traits/Thumper.cs
pchote: I just discovered that we have a ticket asking for the same behaviour more generally: #8932.

Mailaender: That sounds good in theory, but can't be done in practice. I had to disable the hotkey deploy and disable deployment during movement to get the behavior right and glitch free as well as reverse the deployed check to get the infantry sequences animating to work.

pchote: Ok.  If you implement this as a superset/duplicate of that behaviour for now we can work on fixing those issues and merging them later.

Mailaender:
Updated this to work with `AttractsWorms`, upgrades and target types.

OpenRA.Mods.D2k/Traits/Thumper.cs
```diff
@@ -0,0 +1,116 @@
...
+
+		public void Tick(Actor self)
+		{
+			if (!deployed)
+				return;
+
+			if (++tick >= info.ThumpInterval)
+			{
+				tick = 0;
+				Sound.Play(info.ThumpSound, self.CenterPosition);
```
pchote: This could potentially be handled by an improved `AmbientSound` trait (with upgrade support + repeat delay).

OpenRA.Mods.D2k/Traits/Thumper.cs
```diff
@@ -0,0 +1,116 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Mods.Common.Orders;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.D2k.Traits
+{
+	[Desc("Lures the `Sandworm` to its position when deployed.")]
```
pchote: This should be updated to reflect that it now uses upgrades.

OpenRA.Mods.D2k/Traits/Thumper.cs
pchote: I'm okay with leaving this to a future PR that merges this trait with `DeployToUpgrade`.

OpenRA.Mods.D2k/Traits/Thumper.cs
pchote: Why did you have to disable hotkeys?

Mailaender: They don't care if the unit is moving or on the right terrain and fire the order anyway.

pchote: Oh, that is a bug/flaw in these traits: you *must* validate conditions before resolving the order.  Forgetting to validate orders leads to build anywhere / infinite support power exploits.

Mailaender: I had my `CanDeploy()` check misplaced the whole time so I got those BUG in order targeter messages. Fixed it and removed the workarounds.

OpenRA.Mods.D2k/Traits/Thumper.cs
Mailaender: Yeah, it is easier if we do this in a properly connected trait here and then try to generalize more traits (like MadTank and RA2 GIs) at once, otherwise this will end up not covering all use cases.

Mailaender:
Updated with hotkey support.

pchote:
Needs a rebase.

Mailaender:
Rebased.

pchote:
`DeployToUpgrade` has received a lot of upgrades since this was first written, so it should be possible to use that now.

Mailaender:
It is known to crash and no one has any idea how to fix it. https://github.com/OpenRA/OpenRA/issues/9242 I doubt we should use it in our shipped mods.

Phrohdoh:
@Mailaender I pointed out how to fix it just after making that ticket.

abcdefg30:
Needs a rebase.

abcdefg30:
I also couldn't undeploy thumpers using the deploy command.

Mailaender:
I did that on purpose for better usability. Moving it will automatically undeploy.

abcdefg30:
Hm, then we should show the `deploy-blocked` cursor.
But anyway, what do we loose if we allow manual undeploying?

pchote:
Preventing regular undeploying goes against player expections, but this sounds like a good idea for an optional flag (`UndeployOnOrder`) on `DeployToUpgrade`.

Mailaender:
This is getting dragged in a direction where I have to redo this over and over again. Also the end result isn't really satisfying. I think this unit is essentially useless from a gameplay point of view.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/9771/.

Issue #9008, opened 2015-08-09
Playtest AI issues


xp65:
Latest Playtest 2015 08 08

First of all congratulations with this update, it makes many things better. Thanks for the hard work!

The problems I came across while testing:

AI makes 18 (some are out of screen) harvesters with only 2 refineries, but not a single unit to defend himself: (his base builiding was destroyed earlier)
![image](https://cloud.githubusercontent.com/assets/6690499/9152840/9b33db74-3e39-11e5-8a7b-90f820f823da.png)

Subs from AI are dropping performance heavily.
The game was very slow, so I started to look for the reason (I first suspected the wall across the map, but it wasn't it). Once I killed all units on land except a few buildings, I started at sea, each boat I destroyed the game got back faster and faster. I have a modern fast computer, SSD hard drive, 16GB Ram, etc so on a slower pc the game wouldn't have been playable at all.

To give an idea: Sub Pen + 6 subs remaining: huge drop in performance, all other units and buildings from that party destroyed. After destroying the sub pen (boats get destroyed with it), performance is back normal. (at the end part of the game, the green player upper right side of the map)

I couldn't wait till the end in the replay to take screenshots at the end, as due to the slow performance it took takes, but see spikes herein performance due to boats:
![image](https://cloud.githubusercontent.com/assets/6690499/9152912/597dd546-3e3d-11e5-8ffa-9e98ef7a904b.png)

Replay: https://drive.google.com/file/d/0BynHp26MoAw-eVpmME5VS0dDQ0E/view?usp=sharing (you can download the file at the top center of that page) and then extract it with an application like 7zip.
The map was: http://resource.openra.net/maps/758/

Thanks again for the great update and thanks for looking into this.

Phrohdoh:
Hey @xp65, thanks for the detailed ticket!
These types of reports make it much easier to begin looking into an issue which means we can resolve it quicker.

Performance is of course quite important so this will definitely need to be resolved before the stable release is out.

penev92:
Nice, I like this report :)
Do you happen to have the `perf.log` from that game, or from watching the replay? My laptop is significantly slower so it might take a while before I can generate my own logs.

xp65:
Thanks guys! I'll let my pc run the replay again later today to make the perf.log.

xp65:
I did some digging and it seems you guys store all the games in the perf.log until closing, thanks to that, I still had the original log file, I cutted out the actual game and here it is: 
[...]

PR #9022, opened 2015-08-09
Sync only once per tick

RoosterDragon:
This greatly improves performance by not syncing the world state for every single order processed as this becomes very expensive, at the cost of being unable to directly pinpoint the order that causes a desync. Instead the granularity of detecting desyncs is reduced to the tick level.

As noted in https://github.com/OpenRA/OpenRA/issues/9008#issuecomment-129202470, in the worst case scenario of running a max speed replay with high orders rates (due to many players, particularly AI in this case), this could represent upto 50% of CPU!

Only syncing once per tick will benefit tick times when:
- Overall orders rates are high (e.g. due to many players or players with high APM (such as AI))
- An order is given to large groups of units (this should reduce the lag spike significantly)

The disadvantage is that in the event of a desync we won't know what specific order caused it, instead we will only know the frame the desync occurred in. This is hopefully a worthwhile tradeoff.

### DIFF
--- a/OpenRA.Game/Network/FrameData.cs
+++ b/OpenRA.Game/Network/FrameData.cs
@@ -19,6 +19,11 @@ namespace OpenRA.Network
 		{
 			public int Client;
 			public Order Order;
+
+			public override string ToString()
+			{
+				return "ClientId: {0} {1}".F(Client, Order);
+			}
 		}
 
 		readonly Dictionary<int, int> clientQuitTimes = new Dictionary<int, int>();
--- a/OpenRA.Game/Network/OrderIO.cs
+++ b/OpenRA.Game/Network/OrderIO.cs
@@ -35,14 +35,13 @@ namespace OpenRA.Network
 			return ret;
 		}
 
-		public static byte[] SerializeSync(this List<int> sync)
+		public static byte[] SerializeSync(int sync)
 		{
 			var ms = new MemoryStream();
 			using (var writer = new BinaryWriter(ms))
 			{
 				writer.Write((byte)0x65);
-				foreach (var s in sync)
-					writer.Write(s);
+				writer.Write(sync);
 			}
 
 			return ms.ToArray();
--- a/OpenRA.Game/Network/OrderManager.cs
+++ b/OpenRA.Game/Network/OrderManager.cs
@@ -43,8 +43,6 @@ namespace OpenRA.Network
 		public bool GameStarted { get { return NetFrameNumber != 0; } }
 		public IConnection Connection { get; private set; }
 
-		public readonly int SyncHeaderSize = 9;
-
 		List<Order> localOrders = new List<Order>();
 
 		List<ChatLine> chatCache = new List<ChatLine>();
@@ -55,12 +53,8 @@ namespace OpenRA.Network
 
 		static void OutOfSync(int frame)
 		{
+			syncReport.DumpSyncReport(frame);
 			throw new InvalidOperationException("Out of sync in frame {0}.\n Compare syncreport.log with other players.".F(frame));
-		}
-
-		static void OutOfSync(int frame, string blame)
-		{
-			throw new InvalidOperationException("Out of sync in frame {0}: Blame {1}.\n Compare syncreport.log with other players.".F(frame, blame));
 		}
 
 		public void StartGame()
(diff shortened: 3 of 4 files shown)


Phrohdoh:
Please dump a list of orders given in the desync frame so we have some sort of information to go off of.

Mailaender:
This still desyncs properly when watching old replays. :white_check_mark: I am also okay with the trait off. Regarding desync hunting the blame tick was never important to me. I am not sure how this will look like with a real desync.log as a more diverged world state with lot's of diffs might make root-cause analysis more difficult, but this is just speculation. I still favor this trade-off towards performance. We should give it a try. :+1:

pchote:
This will definitely need some test games  (preferrably with some intentionally introduced desyncs) to explore the effect it has on real games.

RoosterDragon:
Orders issued in the out of sync frame will now also be added to the sync report.

chrisforbes:
99% of the desyncs I've looked at have been in Tick, so this seems OK to me.

reaperrr:
Even ordering ~400 infantry around repeatedly is now still playable (unlike before), with this change we're now almost purely pathfinder-bottlenecked when issuing move/attack orders to large numbers of actors.
Singleplayer ingame gets my :+1: , but let's wait for additional feedback on multiplayer/desync behavior before merging this.

Mailaender:
I changed HoverMissile inaccuracy and it still pretty much instantly desynced properly: https://gist.github.com/Mailaender/a1d3e9e6abb2ecf53f8e/revisions

Mailaender:
I changed Transforms.cs to provoke an order related desync:

```patch
@@ -73,7 +73,7 @@ public string VoicePhraseForOrder(Actor self, Order order)
 		bool CanDeploy()
 		{
 			var building = self.TraitOrDefault<Building>();
-			if (building != null && building.Locked)
+			if (building == null)
 				return false;
 
 			return buildingInfo == null || self.World.CanPlaceBuilding(info.IntoActor, buildingInfo, self.Location + info.Offset, self);
```

Result looks okay: https://gist.github.com/Mailaender/7451de1c6d4187f4d49d/revisions

Mailaender:
For comparison: https://gist.github.com/Mailaender/c9ea8ec6ef85197defdc/revisions without those changes on current bleed.

Mailaender:
```patch
@@ -137,7 +137,10 @@ public void DeployTransform(bool queued)
 		public void ResolveOrder(Actor self, Order order)
 		{
 			if (order.OrderString == "DeployTransform")
+			{
+				System.Console.WriteLine(self.World.SharedRandom.Next());
 				DeployTransform(order.Queued);
+			}
 		}
 	}
 }
```

created:

```
[ERROR] FATAL UNHANDLED EXCEPTION: System.InvalidOperationException: Out of sync in frame 4.
 OrderString: "DeployTransform" 
	 Subject: "mcv 45". 
	 TargetActor: "" 
	 TargetLocation: 0,0.
	 TargetString: "".
	 IsImmediate: False.
	 Player(PlayerName): abcdefg30
```

https://gist.github.com/Mailaender/e440c3485ebbf5789a6e/revisions on bleed.

Mailaender:
Which becomes https://gist.github.com/Mailaender/0956817999c1c5d622f7/revisions on this branch, but only due to https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Widgets/WorldCommandWidget.cs#L134

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/6094c3120b9dc7e1664c7aeb0aebcecf2bcc01be...5dd08158d19c77ccb0703e4677aa593240b05cfa)

GraionDilach:
Note to changelog: this affects single player as well, not just network from what I noticed.

penev92:
Yeah, I did test @reaperrr's comment myself, with ~350 infantry:
>Even ordering ~400 infantry around repeatedly is now still playable (unlike before)

Just issuing the orders in single player, there is still a huge difference.

Mailaender:
[Revised changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/263729c227af442822b2b08286f91369d5ec60d7...82edd49b510786fedfbc7c7c4ccc01bbd41caa4e).

PR #9026, opened 2015-08-10
Changed TimestampFormat for server logs

Baxxster:
Thanks for implementing the function in the first place :)

I'd like to change it to this pattern, because I got really confused when I opened my dedicated server window and all I saw was different numbers, there wasn't a lot of activity on my servers that day, so first I thought this was game timer or something, but with this it's easier to see exactly when things happen. This is also good for people who like to make scripts with "grep" and create stats out of that :)

### DIFF
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -72,7 +72,7 @@ namespace OpenRA
 		[Desc("Disallow AI bots.")]
 		public bool LockBots = false;
 
-		public string TimestampFormat = "HH:mm";
+		public string TimestampFormat = "s";
 
 		public ServerSettings() { }
 


OpenRA.Game/Settings.cs
```diff
@@ -72,7 +72,7 @@ public class ServerSettings
 		[Desc("Disallow AI bots.")]
 		public bool LockBots = false;
 
-		public string TimestampFormat = "HH:mm";
+		public string TimestampFormat = "dd/MM - HH:mm:ss";
```
pchote: Could we use [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) (`YYYY-MM-ddTHH:mm:ssZ`, but there's probably a short code for it) instead?

ScottNZ: https://msdn.microsoft.com/en-us/library/az4se3k1%28v=vs.110%29.aspx#Roundtrip
Looks like it's "o"

baxxxster: Yes, I was afraid of suggesting too long string for prettiness-sake, "s" seems to be: YYYY-MM-ddTHH:mm:ss (No Z at the end, but could probably be added if needed)

baxxxster: "o" seems to also include the microseconds which might be "too much" again, there shouldn't be many events on the server that requires deeper logging than seconds afaik?

baxxxster:
Updated the format as requested

RoosterDragon:
's' seems very reasonable.

:+1:

Mailaender:
Indeed. :+1: It is not seconds like I first thought. https://msdn.microsoft.com/en-us/library/az4se3k1%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396#Sortable

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f3a6b6a6b023fa3622fef222664b315a41ae0ad0...002d75c202aa99a3d3643ed24b463b8fb3153e35)

PR #9039, opened 2015-08-11
TD Balance Additions.

AoAGeneral:
The following additions being added. Reference from http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=17430&start=15  |||

Infantry: Detect cloak range 3 from 1. Evade chance 75 from 67.

MLRS: Price reduction 900 from 1200.

Repair Pad: power reduced 10 from 30. HP increased 600 from 400.

Samsite: cost reduced 650 from 750. Build time reduced 32 from 35.

Blue Tiberium: Value decreased 180 from 225.

Rocket Infantry: Damage vs armor increase 105 from 100

Apache and Orca: Price reduced 1100 from 1200. Build time reduced 27 from 29

Visceroids: Now move at a slower rate to prevent rush base attacks.

---

Let me know if I goofed yet again (As I did previously.)

### DIFF
--- a/mods/cnc/rules/aircraft.yaml
+++ b/mods/cnc/rules/aircraft.yaml
@@ -46,7 +46,7 @@ TRAN:
 HELI:
 	Inherits: ^Helicopter
 	Valued:
-		Cost: 1200
+		Cost: 1100
 	Tooltip:
 		Name: Apache Longbow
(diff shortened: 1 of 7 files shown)


AoAGeneral:
Remembered about the upcoming release. Waiting till release then figure out the issues.

PR #9040, opened 2015-08-12
Mobile and Move refactor

reaperrr:
Rebased and cleaned the rest from #7565.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Move/Move.cs
+++ b/OpenRA.Mods.Common/Activities/Move/Move.cs
@@ -135,12 +135,15 @@ namespace OpenRA.Mods.Common.Activities
 
 		public override Activity Tick(Actor self)
 		{
+			// Is the actor disabled?
 			if (moveDisablers.Any(d => d.MoveDisabled(self)))
 				return this;
 
+			// Is the actor at its destination?
 			if (destination == mobile.ToCell)
 				return NextActivity;
 
+			// If no path calculated, do it now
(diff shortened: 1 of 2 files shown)


GraionDilach:
Did you accidentally pushed this into an OpenRA/OpenRA branch too? It got a mobile-refactor1 branch... https://github.com/OpenRA/OpenRA/tree/mobile-refactor1

reaperrr:
Hm yeah, seems something went wrong with my initial push. Removed that from OpenRA/OpenRA.

reaperrr:
This stuff is mostly out of my league, so I'd actually prefer if someone with better understanding of the code ( @pchote ? @RoosterDragon ? ) took over (provided we still want this to begin with, now that we have #9141 and its dependencies).

RoosterDragon:
We still want most of this, since it's nice cleanup (although #9141 would indeed negate a few small refactorings) so we should keep this around until we can iron out the few remaining issues.

reaperrr:
I'm leaving this as-is since I'm afraid I might break something if I try to rebase this.

RoosterDragon:
Actually - on taking a second look at this it's probably rotted too much to justify the effort to rebase it in. :/

Unless somebody else desperately wants to grab it I think you could safely close this, @reaperrr.

reaperrr:
Ok. I won't delete the branch for now so anyone still interested in it can scavenge for salvagable bits.

PR #9047, opened 2015-08-12
TD tooltip fixes

clemty:
- Mention Apache being strong vs aircraft.
- Mention engineers will damage buildings, not just capture them.
- Removed whitespace before line breaks

reverted:
- Sentences are now (mostly) lowercase, end with full-stop.
- Changed "defence" tab tooltip to "defense".

### DIFF
--- a/mods/cnc/rules/aircraft.yaml
+++ b/mods/cnc/rules/aircraft.yaml
@@ -47,7 +47,7 @@ HELI:
 		Cost: 1200
 	Tooltip:
 		Name: Apache Longbow
-		Description: Helicopter Gunship with Chainguns.\n  Strong vs Infantry, Light Vehicles\n  Weak vs Tanks
+		Description: Helicopter Gunship with Chainguns.\n  Strong vs Infantry, Light Vehicles and\n  Aircraft\n  Weak vs Tanks
 	Buildable:
 		BuildPaletteOrder: 20
 		Prerequisites: hpad, anyhq, ~techlevel.medium
@@ -181,7 +181,7 @@ A10:
 	Inherits: ^Plane
 	Tooltip:
 		Name: A10 Bomber
-		Description: Used to deliver Napalm strikes.
+		Description: Used to deliver air strikes.
 	Valued:
 		Cost: 2000
 	Plane:
--- a/mods/cnc/rules/infantry.yaml
+++ b/mods/cnc/rules/infantry.yaml
@@ -136,7 +136,7 @@ E6:
 		Cost: 500
 	Tooltip:
 		Name: Engineer
-		Description: Infiltrates and captures enemy structures.\n  Strong vs Nothing\n  Weak vs Everything
+		Description: Damages and captures enemy structures.\n  Unarmed
 	Buildable:
 		BuildPaletteOrder: 30
 		Queue: Infantry.GDI, Infantry.Nod
@@ -203,7 +203,7 @@ PVICE:
 		Queue: Biolab
 		BuildPaletteOrder: 40
 	Tooltip:
-		Description: Mutated abomination that spits liquid tiberium.\n  Strong vs Infantry, Buildings\n  Weak vs Aircraft
+		Description: Mutated abomination that spits liquid Tiberium.\n  Strong vs Infantry, Buildings\n  Weak vs Aircraft
 	ActorLostNotification:
 
 STEG:
--- a/mods/cnc/rules/structures.yaml
+++ b/mods/cnc/rules/structures.yaml
@@ -125,7 +125,7 @@ NUK2:
 		Cost: 800
 	Tooltip:
 		Name: Advanced Power Plant
-		Description: Provides more power, cheaper than the \nstandard Power Plant
+		Description: Provides more power, cheaper than the\nstandard Power Plant
 	ProvidesPrerequisite:
 		Prerequisite: anypower
 	Buildable:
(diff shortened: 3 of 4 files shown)


mods/cnc/rules/aircraft.yaml
```diff
@@ -47,7 +47,7 @@ HELI:
 		Cost: 1200
 	Tooltip:
 		Name: Apache Longbow
-		Description: Helicopter Gunship with Chainguns.\n  Strong vs Infantry, Light Vehicles\n  Weak vs Tanks
+		Description: Helicopter gunship with chainguns.\n  Strong vs infantry, light vehicles &\n  aircraft.\n  Weak vs tanks.
```
abcdefg30: `Strong vs Infantry, Light vehicles, Aircraft\n  Weak vs Tanks`, to be consistent with #8277 and #7967, please.

mods/cnc/rules/aircraft.yaml
```diff
@@ -4,7 +4,7 @@ TRAN:
 		Cost: 750
 	Tooltip:
 		Name: Chinook Transport
-		Description: Fast Infantry Transport Helicopter.\n  Unarmed
+		Description: Fast infantry transport helicopter.\n  Unarmed.
```
abcdefg30: No `.` after `Unarmed`, please.

mods/cnc/rules/aircraft.yaml
```diff
@@ -98,7 +98,7 @@ ORCA:
 		Cost: 1200
 	Tooltip:
 		Name: Orca
-		Description: Helicopter Gunship with AG Missiles.\n  Strong vs Buildings, Tanks\n  Weak vs Infantry
+		Description: Helicopter gunship with AG missiles.\n  Strong vs buildings, tanks.\n  Weak vs infantry.
```
abcdefg30: `Strong vs Buildings, Tanks\n  Weak vs Infantry`

abcdefg30:
Above affects all `Strong/Weak vs`. Sorry for making you revert that.
Otherwise we would get inconsistent with the other mods.
(Or do we want `TD` to be special again?)

GraionDilach:
Shouldn't the second tab be named as Support? Or do I remember wrong that it was treat as such...

clemty:
>Above affects all Strong/Weak vs. Sorry for making you revert that.
>Otherwise we would get inconsistent with the other mods.
>(Or do we want TD to be special again?)

I didn't intend to make spelling inconsistent with the other mods, it's just that the current way of capitalizing some units/unit types seemed incorrect to me

Making unit types upper-case in the strong/weak sections creates another inconsistency:
- Fast infantry armed with grenades. / Anti-tank/Anti-aircraft infantry. ... Weak vs Infantry

also:
- Strong vs Buildings, slow-moving targets

why "Buildings" but not "Targets" ?

>No . after Unarmed, please.

This would create a case where most lines end in "." and only "Unarmed" does not - according to http://www.edufind.com/english-grammar/period-full-stop-or-point/ it is ok to put "." after a single word.

>Shouldn't the second tab be named as Support? Or do I remember wrong that it was treated as such...

That's a good point - the Tiberium silo is not a defense structure

Phrohdoh:
A description for an actor is not a sentence and therefor, technically, does not need to end with a full-stop.

clemty:
updated

mods/cnc/rules/aircraft.yaml
```diff
@@ -47,7 +47,7 @@ HELI:
 		Cost: 1200
 	Tooltip:
 		Name: Apache Longbow
-		Description: Helicopter Gunship with Chainguns.\n  Strong vs Infantry, Light Vehicles\n  Weak vs Tanks
+		Description: Helicopter Gunship with Chainguns.\n  Strong vs Infantry, Light Vehicles &\n  Aircraft\n  Weak vs Tanks
```
GraionDilach: I don't really like the ampersand here.

mods/cnc/rules/infantry.yaml
```diff
@@ -136,7 +136,7 @@ E6:
 		Cost: 500
 	Tooltip:
 		Name: Engineer
-		Description: Infiltrates and captures enemy structures.\n  Strong vs Nothing\n  Weak vs Everything
+		Description: Infiltrates and captures enemy structures.\n  Strong vs Buildings\n  Weak vs Everything
```
abcdefg30: `Infiltrates and captures enemy structures.\n  Unarmed`

mods/cnc/rules/structures.yaml
```diff
@@ -914,7 +914,7 @@ BRIK:
 		Value: 0
 	Tooltip:
 		Name: Concrete Barrier
-		Description: Stop units.
+		Description: Stops infantry & most tanks.
```
abcdefg30: `&` -> `and`, please.

clemty:
updated.

>I don't really like the ampersand here.

Neither do I, but it is used that way in vehicles.yaml (4 hits) and structures.yaml (5 hits).

>& -> and, please.

I changed all occurrences of "&" into "and"

abcdefg30:
Thanks. :+1:

mods/cnc/rules/vehicles.yaml
```diff
@@ -502,7 +502,7 @@ STNK:
 		Cost: 900
 	Tooltip:
 		Name: Stealth Tank
-		Description: Long-range missile tank that can cloak. \nCan attack Aircraft. \nHas weak armor. Can be spotted by infantry.\n  Strong vs Vehicles, Tanks\n  Weak vs Infantry.
+		Description: Long-range missile tank that can cloak.\nCan attack Aircraft.\nHas weak armor. Can be spotted by infantry.\n  Strong vs Vehicles, Tanks\n  Weak vs Infantry.
```
Unit158: They can be spotted by static defence as well.

PR #9048, opened 2015-08-12
Add IsSuicide to Explodes trait for kill credit

### LINKED ISSUE #9044: RA - Unit that destroyed the Demo Truck is counted as source of damage.
Enemy units (to destroyer not to Demo Truck) take damage from explosion of Demo Truck starts to attack unit that destroyed the Demo Truck if not destroyed by explosion.

atlimit8:


pchote:
#9044 applies for all explosion kills (artillery, etc), and not just the demo truck.  Adding a specific flag just for the demo truck feels like a new hack on top of an old hack.

pchote:
What do you want to do with this @atlimit8?

atlimit8:
Without any feedback and the direction for fixing #9044 being unclear, I'll just close this.

PR #9054, opened 2015-08-13
Merge Plane and Helicopter traits

penev92:
First steps toward having a single aircraft type that can handle different situations.
Ultimate goals are:
- Provide modders with as much flexibility regarding aircraft as possible.
- Reduce the 10 traits and 15 activities we currently have to a sensible number.
- Fix some aircraft issues along the way.

This PR merges `Plane` and `Helicopter` into their base `Aircraft` trait.
It does some cleanup on `Aircraft` as well, and some minor side cleanups along the way.
The last commit also fixes an issue currently present on bleed.

While the commits here aren't as good (self-contained) as I want them, I do suggest reviewing by commit.
Ingame result should be that nothing changes behavior-wise (apart from the bugfix from the last commit). It's all pretty easy to test, actually.


OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -305,41 +457,123 @@ public string VoicePhraseForOrder(Actor self, Order order)
...
+			}
+			else if (order.OrderString == "ReturnToBase")
+			{
+				UnReserve();
+				self.CancelActivity();
+				if (isPlane)
+					self.QueueActivity(new ReturnToBase(self));
+				else
+					self.QueueActivity(new HeliReturnToBase(self));
+				self.QueueActivity(new ResupplyAircraft(self));
```
abcdefg30: Can we get a new line before this one, please?

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -305,41 +457,123 @@ public string VoicePhraseForOrder(Actor self, Order order)
...
+			}
+			else
+			{
+				// Game.Debug("Unreserve due to unhandled order: {0}".F(order.OrderString));
```
abcdefg30: Could you remove the outcommented code here and drop the braces then?

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -305,41 +457,123 @@ public string VoicePhraseForOrder(Actor self, Order order)
...
-		public bool OverrideSelection { get { return false; } }
+		public void ResolveOrder(Actor self, Order order)
+		{
+			if (order.OrderString == "Move")
+			{
+				var cell = self.World.Map.Clamp(order.TargetLocation);
+				var explored = self.Owner.Shroud.IsExplored(cell);
 
-		readonly AircraftInfo info;
+				if (!explored && !Info.MoveIntoShroud)
```
abcdefg30: Could you change this to `if (!Info.MoveIntoShroud && !self.Owner.Shroud.IsExplored(cell))` and drop the `var` above? Then we don't need the `IsExplored` lookup when `MoveIntoShroud` is `true`.

mods/ts/rules/defaults.yaml
```diff
@@ -261,7 +261,7 @@
 		Probability: 10
 		OwnerType: InternalName
 		InternalOwner: Creeps
-		DeathType: EnergyDeath # TODO: FIX ME! (Tiberium currently uses the wrong damage type!)
+		DeathType: EnergyDeath
```
abcdefg30: This change looks unrelated and should be reverted.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -305,41 +457,123 @@ public string VoicePhraseForOrder(Actor self, Order order)
...
+					}
+				}
+			}
+			else if (order.OrderString == "Stop")
+			{
+				self.CancelActivity();
+				if (GetActorBelow() != null)
+				{
+					self.QueueActivity(new ResupplyAircraft(self));
+					return;
```
abcdefg30: Can't we just use
```
if (GetActorBelow() != null)
	return;
```
straight at the beginning instead of throwing away the activity and creating a new one?

Also what about checking if `Reservation` is `null` instead of using `GetActorBelow`?

abcdefg30:
It also seems like you broke reserving for helicopter.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
penev92: We can't, in case there are queued orders.

I'll see about `Reservation`, but I doubt it. That part is so meh... So I doubt it.

abcdefg30:
Also needs a rebase.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
penev92: Actually I'm going to uncomment it.

penev92:
Rebased, fixed the "helicopter" reservations and fixed some style issues.

Mailaender:
Needs a rebase. Sorry.

penev92:
Rebased.

penev92:
Rebased.

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -18,22 +18,28 @@ namespace OpenRA.Mods.Common.Activities
...
 	{
-		readonly Plane plane;
-		readonly PlaneInfo planeInfo;
+		readonly Aircraft plane;
+		readonly AircraftInfo planeInfo;
 		bool isCalculated;
 		Actor dest;
 		WPos w1, w2, w3;
 
+		public ReturnToBase(Actor self)
```
pchote: Can you default `dest = null` below instead of duplicating this ctor?

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -19,7 +20,8 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
-	public class AircraftInfo : ITraitInfo, IFacingInfo, IOccupySpaceInfo, ICruiseAltitudeInfo, UsesInit<LocationInit>, UsesInit<FacingInit>
+	public class AircraftInfo : IFacingInfo, IOccupySpaceInfo, IMoveInfo, ICruiseAltitudeInfo,
```
pchote: Looks like you've dropped the `ITraitInfo` from here?

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -46,27 +48,55 @@ public class AircraftInfo : ITraitInfo, IFacingInfo, IOccupySpaceInfo, ICruiseAl
...
+		INotifyAddedToWorld, INotifyRemovedFromWorld, INotifyActorDisposing
 	{
 		static readonly Pair<CPos, SubCell>[] NoCells = { };
 
-		readonly AircraftInfo info;
+		public readonly AircraftInfo Info;
 		readonly Actor self;
 
+		// Return To Base Path Hash is a leftover for debugging an old desync with aircraft
+		[Sync] public WPos RTBPathHash;
```
pchote: This can be removed.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -75,30 +105,41 @@ public Aircraft(ActorInitializer init, AircraftInfo info)
 				SetPosition(self, init.Get<CenterPositionInit, WPos>());
 
 			Facing = init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : info.InitialFacing;
+
+			// TODO: HACK: This is a hack until we can properly distinquish between airplane and helicopter!
+			// Or until the activities get unified enough so that it doesn't matter.
+			isPlane = !info.CanHover;
```
pchote: Kind of ugly, but ok.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -108,24 +149,37 @@ public void Repulse()
...
 				return WVec.Zero;
 
-			return self.World.FindActorsInCircle(self.CenterPosition, info.IdealSeparation)
-				.Where(a => !a.IsDead && a.HasTrait<Aircraft>() && a.Info.Traits.Get<AircraftInfo>().CruiseAltitude == info.CruiseAltitude)
+			var repulsionForce = self.World.FindActorsInCircle(self.CenterPosition, Info.IdealSeparation)
+				.Where(a => !a.IsDead && a.HasTrait<Aircraft>() && a.Info.Traits.Get<AircraftInfo>().CruiseAltitude == Info.CruiseAltitude)
 				.Select(GetRepulsionForce)
 				.Aggregate(WVec.Zero, (a, b) => a + b);
+
+			if (!isPlane)
```
pchote: You can check against `CanHover` directly here.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -240,64 +302,133 @@ public SubCell GetAvailableSubCell(CPos a, SubCell preferredSubCell = SubCell.An
...
+
+		public CPos NearestMoveableCell(CPos cell) { return cell; }
+
+		public bool IsMoving { get { return self.CenterPosition.Z > 0; } set { } }
+
+		public bool CanEnterTargetNow(Actor self, Target target)
+		{
+			if (target.Positions.Any(p => self.World.ActorMap.GetUnitsAt(self.World.Map.CellContaining(p)).Any(a => a != self && a != target.Actor)))
+				return false;
+			var res = target.Actor.TraitOrDefault<Reservable>();
```
pchote: Nit: newline before this please.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -240,64 +302,133 @@ public SubCell GetAvailableSubCell(CPos a, SubCell preferredSubCell = SubCell.An
...
+			if (res == null)
+				return true;
+			UnReserve();
```
pchote: Here too.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
pchote: Lets maintain consistency with the original implementations for safety.  We can address this in its own smaller pr later.

pchote:
I haven't tested this, but the code changes look sensible (and not nearly as intimidating as the line change count indicates - it is mostly just renames).  Could you please squash down some of the commits?

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
penev92: Yes. One (or more) of these others inherit it, so it's redundant.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
penev92: Yes, but I opted for consistency and decided to use the same hack everywhere.

pchote: Its not a hack here: This code deals with the different cases for units that can hover (allowing repulsion in any direction) vs those that must continue flying forward always.

penev92:
Updated.

pchote:
I don't like the "Apply changes from PR XXX" commits: they should be squashed down into the prs that moved the original code.

pchote:
Likewise for the "Some fixes and adjustments".

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -143,24 +184,37 @@ public void Repulse()
...
-				.Where(a => !a.IsDead && a.HasTrait<Aircraft>() && a.Info.Traits.Get<AircraftInfo>().CruiseAltitude == info.CruiseAltitude)
+			var repulsionForce = self.World.FindActorsInCircle(self.CenterPosition, Info.IdealSeparation)
+				.Where(a => !a.IsDead && a.HasTrait<Aircraft>() && a.Info.Traits.Get<AircraftInfo>().CruiseAltitude == Info.CruiseAltitude)
 				.Select(GetRepulsionForce)
 				.Aggregate(WVec.Zero, (a, b) => a + b);
+
+			if (Info.CanHover)
+				return repulsionForce;
+
+			var currentDir = FlyStep(Facing);
```
pchote: Can we get a comment here explaining that non-hovering aircraft must always move forward.

OpenRA.Mods.Common/Traits/Cargo.cs
```diff
@@ -66,7 +66,7 @@ public class Cargo : IPips, IIssueOrder, IResolveOrder, IOrderVoice, INotifyCrea
 
 		int totalWeight = 0;
 		int reservedWeight = 0;
-		Helicopter helicopter;
+		Aircraft helicopter;
```
pchote: Can we please rename this to aircraft?

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2001,7 +2002,9 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 
 						// Add `WhileCloakedUpgrades: underwater` to Cloak trait if `CloakTypes: Underwater`
 						var cloak = node.Value.Nodes.FirstOrDefault(n => (n.Key == "Cloak" || n.Key.StartsWith("Cloak@"))
-							&& n.Value.Nodes.Any(p => p.Key == "CloakTypes" && p.Value.Value == "Underwater"));
+						                                                 &&
```
pchote: This indentation here looks to have gone a bit wonky.

mods/cnc/rules/aircraft.yaml
```diff
@@ -9,14 +9,13 @@ TRAN:
...
 		Queue: Aircraft.GDI, Aircraft.Nod
-	Helicopter:
+	Aircraft:
 		LandWhenIdle: true
 		ROT: 5
 		Speed: 140
 		InitialFacing: 0
 		LandableTerrainTypes: Clear,Rough,Road,Ore,Beach,Tiberium,BlueTiberium
 		AltitudeVelocity: 0c100
-		AirborneUpgrades: airborne
```
pchote: Removing this looks like a regression.

mods/cnc/rules/aircraft.yaml
```diff
@@ -53,11 +52,10 @@ HELI:
 		BuildPaletteOrder: 20
 		Prerequisites: hpad, anyhq, ~techlevel.medium
 		Queue: Aircraft.Nod
-	Helicopter:
+	Aircraft:
 		RearmBuildings: hpad
 		ROT: 4
 		Speed: 186
-		AirborneUpgrades: airborne
```
pchote: Same here and elsewhere.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -53,23 +55,51 @@ public class AircraftInfo : ITraitInfo, IFacingInfo, IOccupySpaceInfo, ICruiseAl
...
+		[Desc("Does this actor need to turn before landing?")]
+		public readonly bool TurnToLand = false;
+
+		public readonly WDist LandAltitude = WDist.Zero;
+
+		[Desc("How fast this actor ascends or descends when using horizontal take off/landing.")]
+		public readonly WAngle MaximumPitch = WAngle.FromDegrees(10);
+
+		[Desc("How fast this actor ascends or descends when using vertical take off/landing.")]
+		public readonly WDist AltitudeVelocity = new WDist(43);
```
pchote: Velocity implies a vector... could we rename this to `ClimbSpeed`?

pchote: This can probably be left for a future PR, since it is moved from existing code.

OpenRA.Mods.Common/Scripting/Global/ReinforcementsGlobal.cs
```diff
@@ -137,7 +137,7 @@ public Actor[] Reinforce(Player owner, string[] actorTypes, CPos[] entryPath, in
 			}
 			else
 			{
-				var heli = transport.TraitOrDefault<Helicopter>();
+				var heli = transport.TraitOrDefault<Aircraft>();
```
pchote: This has an implicit dependency on the helicopter logic below.  Can you please add a check on the temporary isPlane flag?

pchote: Also `heli` &rarr; `aircraft.

OpenRA.Mods.Common/Scripting/Properties/HelicopterProperties.cs
```diff
@@ -16,7 +16,7 @@
 namespace OpenRA.Mods.Common.Scripting
 {
 	[ScriptPropertyGroup("Movement")]
-	public class HelicopterProperties : ScriptActorProperties, Requires<HelicopterInfo>
+	public class HelicopterProperties : ScriptActorProperties, Requires<AircraftInfo>
```
pchote: The `Move` property is going to conflict with the one in `PlaneProperties`.  Could you please rename `PlaneProperties` &rarr; `AircraftProperties` (won't affect the lua compatibility, as that isn't exposed), and then add the heli/plane checks there?

OpenRA.Mods.Common/Scripting/Global/ReinforcementsGlobal.cs
Phrohdoh: If this has an implicit dependency on heli logic I see no reason to rename the variable.
`heli` is a hint while `aircraft` is just repeating the type for consistency.

OpenRA.Mods.D2k/Activities/DeliverUnit.cs
```diff
@@ -23,7 +23,7 @@ public class DeliverUnit : Activity
 		readonly IMove movement;
 		readonly Carryable carryable;
 		readonly Carryall carryall;
-		readonly Helicopter helicopter;
+		readonly Aircraft helicopter;
```
pchote: `helicopter` &rarr; `aircraft` please.

OpenRA.Mods.D2k/Activities/PickupUnit.cs
```diff
@@ -22,7 +22,7 @@ public class PickupUnit : Activity
 		readonly IMove movement;
 		readonly Carryable carryable;
 		readonly Carryall carryall;
-		readonly Helicopter helicopter;
+		readonly Aircraft helicopter;
```
pchote: Here too.

mods/cnc/rules/aircraft.yaml
penev92: Not really. The default all of these inherit has that upgrade, so this is just removing junk yaml left by an upgrade rule.

pchote: Ah, I see.

OpenRA.Mods.Common/Scripting/Global/ReinforcementsGlobal.cs
pchote: This line queries it as an Aircraft.  The next line then needs to check if that Aircraft is also a Helicopter.

abcdefg30:
Needs a rebase.

penev92:
Rebased and took care of the Lua API part.

atlimit8:
Needs another rebase.

penev92:
Rebased.

atlimit8:
Needs another rebase.

OpenRA.Mods.Common/Scripting/Properties/AircraftProperties.cs
```diff
@@ -16,23 +16,34 @@
...
+	public class AircraftProperties : ScriptActorProperties, Requires<AircraftInfo>
 	{
-		public PlaneProperties(ScriptContext context, Actor self)
-			: base(context, self) { }
+		readonly bool isPlane;
+
+		public AircraftProperties(ScriptContext context, Actor self)
+			: base(context, self)
+		{
+			isPlane = !self.Trait<Aircraft>().Info.CanHover;
```
pchote: This should use `Trait<Aircraft>().isPlane` (which will need to become public).

penev92:
Updated.

abcdefg30:
It looks like you replace `AltitudeVelocity` by `ClimbSpeed`. Why?
Also it needs an upgrade rule.

penev92:
Rebased, reverted the `AltitudeVelocity` -> `ClimbSpeed` rename (was requested by @pchote because the name makes more sense) and snuck a little behavioral change to enable scripted reinforcements by airplane (so far they were only doable with a helicopter) as requested by @abcdefg30.

Good luck to whoever tries to to the changelog entries for this, if it ever gets to that.

obrakmann:
Looks fine to me, no regressions spotted, either. :+1:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/70d67a23b91ded183a11a1aedda1e21feb34da76...bda42c89c86eab4f84574dc1bced343e59c8a983)

PR #9057, opened 2015-08-13
Fixed a likely null reference exception when checking mod sequence definitions

Mailaender:
Detected by Coverity.


abcdefg30:
:+1:

PR #9058, opened 2015-08-13
Reduce allocations in the main game loop

RoosterDragon:
- Cache the shroud projection even for flat maps to avoid allocating single element arrays.
- Avoid LINQ in shroud and map projection queries to avoid enumerator allocations.
- Avoid LINQ in calculation of sync values.
- Cache enumerables in ProductionQueue.

Mostly straightforward, but if testing, I would double-check I didn't break the shroud on both the rectangular and isometric maps.

Overall, this reduces allocation in the main game loop by about 20%.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
+++ b/OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
@@ -68,14 +68,16 @@ namespace OpenRA.Mods.Common.Traits
 		public readonly ProductionQueueInfo Info;
 		readonly Actor self;
 
+		// A list of things we could possibly build
+		readonly Dictionary<ActorInfo, ProductionState> produceable = new Dictionary<ActorInfo, ProductionState>();
+		readonly List<ProductionItem> queue = new List<ProductionItem>();
+		readonly IEnumerable<ActorInfo> allProduceables;
+		readonly IEnumerable<ActorInfo> buildableProduceables;
+
 		// Will change if the owner changes
 		PowerManager playerPower;
 		PlayerResources playerResources;
 		protected DeveloperMode developerMode;
-
-		// A list of things we could possibly build
-		Dictionary<ActorInfo, ProductionState> produceable;
-		List<ProductionItem> queue = new List<ProductionItem>();
 
 		public Actor Actor { get { return self; } }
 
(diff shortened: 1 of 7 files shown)


OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -68,15 +68,17 @@ public class ProductionQueue : IResolveOrder, ITick, ITechTreeElement, INotifyOw
 		public readonly ProductionQueueInfo Info;
 		readonly Actor self;
 
+		// A list of things we could possibly build
+		readonly Dictionary<ActorInfo, ProductionState> produceable = new Dictionary<ActorInfo, ProductionState>();
+		readonly List<ProductionItem> queue = new List<ProductionItem>();
+		readonly IEnumerable<ActorInfo> allProducables;
+		readonly IEnumerable<ActorInfo> buildableProducables;
```
ScottNZ: Inconsistent spelling: `buildableProducables`/`allProducables` here but `produceable` above.

RoosterDragon: Oddly I went for the correct spelling of producible initially but decided to change it back rather than introduce inconsistency/change everything else. Instead I managed to spell it wrong but differently to the rest of the wrong spellings. Heh.

penev92:
I find it a disturbing coincidence that I decided to test this PR and OpenRA got to 1.34 GB ram when loading TS, then I got `System.OutOfMemoryException` at `Map.cs`, line 451.

RoosterDragon:
I made a seemingly harmless change before pushing that resulting in a infinite loop. Whoops!

Corrected that and retested the shroud on RA and TS - should be fine now.

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
RoosterDragon: Adjusted for consistency with existing names.

Mailaender:
:+1: Looks like the memory leak is gone. I can't spot any regressions for shroud and production. :white_check_mark:

penev92:
I can't confirm the memory win, but the changes do look quite reasonable, so at least in theory it should be there :P
Everything seems to work :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/081b1e03b769aa59db22453fb28c38e3fd58c7aa...3c22226763a298fcfbe9d13d561622a77cc608fb)

PR #9072, opened 2015-08-14
Fix capitalisation of the "Explored Map" option.

pchote:
Fixes a silly polish regression that shouldn't have made it through review.


Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/8792.

PR #9073, opened 2015-08-14
Minor Production trait cleanups.

pchote:
Fixes a handful of minor nits that were missed in #8979.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Production.cs
+++ b/OpenRA.Mods.Common/Traits/Production.cs
@@ -32,7 +32,7 @@ namespace OpenRA.Mods.Common.Traits
 	{
 		readonly Lazy<RallyPoint> rp;
 
-		public ProductionInfo Info;
+		public readonly ProductionInfo Info;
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.Common/Traits/Production.cs
```diff
@@ -118,17 +118,11 @@ public virtual bool Produce(Actor self, ActorInfo producee, string factionVarian
 			if (Reservable.IsReserved(self))
 				return false;
 
-			if (!occupiesSpace)
```
penev92: I put this here for a reason. It is the `CanUseExit()` below that crashes, so moving this check won't work.

pchote: See the other change in this patch, which prevents modders from defining exits on actors that don't occupy space.

RoosterDragon:
:+1: Looks good to me.

abcdefg30:
Looks also good to me. :+1:

PR #9075, opened 2015-08-15
Removes LocalPlayer check from HideMapCrateAction and RevealMapCrateAction.

### LINKED ISSUE #9063: Hide-map effect does not work in replays
The hide-map effect of crates doesn't work in replays, meaning the scouted area merely remains hidden behind fog of war, not newly created shroud as it should.

Here's a replay and a map (provided by Zjordy):
http://s000.tinyupload.com/index.php?file_id=59247814826002024943

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25745757-hide-map-effect-does-not-work-in-replays?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #9127: Chrono crash, explored terrain using crate
release-20150614

The game crashed after I used the chrono, but to a location I normally could not, but the terrain was already explored using a special crate. It resulted in "Out of sync" with AdvancedChronoshift

```
Red Alert Mod at Version release-20150614
Exception of type `System.InvalidOperationException`: Out of sync in frame 5876.
 OrderString: "AdvancedChronoshift" 
	 Subject: "player 4". 
	 TargetActor: "" 
	 TargetLocation: 29,18.
	 TargetString: "".
	 IsImmediate: False.
	 Player(PlayerName): General

 Compare syncreport.log with other players.
   bij OpenRA.Network.OrderManager.OutOfSync(Int32 frame, Int32 index)
   bij OpenRA.Network.OrderManager.CheckSync(Byte[] packet)
   bij OpenRA.Network.OrderManager.<TickImmediate>c__AnonStorey0.<>m__0(Int32 clientId, Byte[] packet)
   bij OpenRA.Network.ReplayRecorderConnection.<Receive>c__AnonStorey0.<>m__0(Int32 client, Byte[] data)
   bij OpenRA.Network.EchoConnection.Receive(Action`2 packetFn)
   bij OpenRA.Network.ReplayRecorderConnection.Receive(Action`2 packetFn)
   bij OpenRA.Network.OrderManager.TickImmediate()
   bij OpenRA.Sync.<CheckSyncUnchanged>c__AnonStorey0.<>m__0()
   bij OpenRA.Sync.CheckSyncUnchanged[T](World world, Func`1 fn)
   bij OpenRA.Sync.CheckSyncUnchanged(World world, Action fn)
   bij OpenRA.Game.InnerLogicTick(OrderManager orderManager)
   bij OpenRA.Game.LogicTick()
   bij OpenRA.Game.Loop()
   bij OpenRA.Game.Run()
   bij OpenRA.Program.Run(String[] args)
   bij OpenRA.Program.Main(String[] args)
```

All files including replay can be found here: https://drive.google.com/folderview?id=0BynHp26MoAw-fjJxVklYTmQzTjNvZzhUbEQ4MTJoV2d1bkVZVlBMU25QcnA0eU1FR1BnTzg&usp=sharing

Thanks for looking into this

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26065409-chrono-crash-explored-terrain-using-crate?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

deniz1a:
Fixes #9063.
Fixes #9127.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Crates/HideMapCrateAction.cs
+++ b/OpenRA.Mods.Common/Traits/Crates/HideMapCrateAction.cs
@@ -32,9 +32,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		public override void Activate(Actor collector)
 		{
-			if (collector.Owner == collector.World.LocalPlayer)
-				collector.Owner.Shroud.ResetExploration();
-
+			collector.Owner.Shroud.ResetExploration();
 			base.Activate(collector);
 		}
 	}
--- a/OpenRA.Mods.Common/Traits/Crates/RevealMapCrateAction.cs
+++ b/OpenRA.Mods.Common/Traits/Crates/RevealMapCrateAction.cs
@@ -8,8 +8,6 @@
  */
 #endregion
 
-using OpenRA.Traits;
-
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Reveals the entire map.")]
@@ -31,18 +29,15 @@ namespace OpenRA.Mods.Common.Traits
 			this.info = info;
 		}
 
-		bool ShouldReveal(Player collectingPlayer)
-		{
-			if (info.IncludeAllies)
-				return collectingPlayer.World.LocalPlayer != null &&
-					collectingPlayer.Stances[collectingPlayer.World.LocalPlayer] == Stance.Ally;
-
-			return collectingPlayer == collectingPlayer.World.LocalPlayer;
-		}
-
 		public override void Activate(Actor collector)
 		{
-			if (ShouldReveal(collector.Owner))
+			if (info.IncludeAllies)
+			{
+				foreach (var player in collector.World.Players)
+					if (collector.Owner.IsAlliedWith(player))
+						player.Shroud.ExploreAll(player.World);
+			}
+			else
 				collector.Owner.Shroud.ExploreAll(collector.World);
 
 			base.Activate(collector);


Mailaender:
:+1: This check wasn't removed when we introduced per-player shrouds and is now bogus. https://github.com/deniz1a/OpenRA/commit/f41fb32d60fbdd44e0a28156c0cb222326399f2e#diff-25b65f46503e3e43f1aa30d8af29a2a1

deniz1a:
Ok then I should also remove that check from ````RevealMapCrateAction````...

Mailaender:
This pull request changed completely. My initial :+1: only belongs to https://github.com/deniz1a/OpenRA/commit/82473ae0c88f09c364577df0d2ee9ef22438f7ec.

deniz1a:
I expanded this PR a bit. It does the following:
* Fixes #9063 by removing an unnecessary LocalPlayer check in HideMapCrateAction.
* Doesn't set the chance of getting hide-map crate as zero when player has fog visibility. Instead, player can get it but it has no effect. This works better with maps that have custom selection shares.
* Adds an IncludeAllies variable to HideMapCrateAction.
* Sets IncludeAllies variables of hide-map and reveal-map crates as true. I think it is better if they also affect allies by default. Otherwise they cause difference in shroud visibility between allies.
* Fixes a similar issue in replays for reveal-map crate.

Unit158:
:-1: for everything but the fix. It worked fine except for the replays.

deniz1a:
Are you also against adding IncludeAllies variable to HideMapCrateAction?

Unit158:
I stand by my words.

deniz1a:
Ok :koala:

abcdefg30:
I think we should default to `false` for `IncludeAllies`, or we need an upgrade rule.

deniz1a:
Ok, I'll revert that. What about the other changes?

OpenRA.Mods.Common/Traits/Crates/HideMapCrateAction.cs
```diff
@@ -8,34 +8,42 @@
...
 		}
 
 		public override void Activate(Actor collector)
 		{
-			if (collector.Owner == collector.World.LocalPlayer)
-				collector.Owner.Shroud.ResetExploration();
-
 			base.Activate(collector);
+
+			// don't ever hide the map for people who have GPS.
```
abcdefg30: Please let this commet start with an uppercase letter, as you already move it.

abcdefg30:
Code looks good to me otherwise.

deniz1a:
Are there any maps that depend on these crates not affecting allies?

deniz1a:
It is not practical to separate the bug fix from the other changes I made. So it is better to review this as is and I can revert changes that are unwanted.

deniz1a:
Reverted everything except the bugfix.

deniz1a:
For the other changes, I'll open a separate PR after this one is merged.

OpenRA.Mods.Common/Traits/Crates/RevealMapCrateAction.cs
```diff
@@ -31,18 +31,12 @@ public RevealMapCrateAction(Actor self, RevealMapCrateActionInfo info)
...
-					collectingPlayer.Stances[collectingPlayer.World.LocalPlayer] == Stance.Ally;
-
-			return collectingPlayer == collectingPlayer.World.LocalPlayer;
-		}
-
 		public override void Activate(Actor collector)
 		{
-			if (ShouldReveal(collector.Owner))
+			if (info.IncludeAllies)
+				foreach (var ally in collector.World.Players.Where(collector.Owner.IsAlliedWith))
```
deniz1a: This could also be done without the LINQ query, by checking IsAlliedWith with an if inside foreach. Which one is better?

Mailaender:
:+1: again.

penev92:
This most definitely does *not* fix #9127

penev92:
Also the same issue as #9063 is present for `RevealMapCrateAction`.

penev92:
OK, seems to fix both issues (and the unreported one), sorry.
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/3ee6ea5bfb348d368c9de31b9de4dfaca39839ee...2e57c45545103ec965a6a9dc107926a825effc1d)

PR #9079, opened 2015-08-15
Dont crash when Documents\OpenRA is hidden

LipkeGu:
It fixes the hidden openra folder crash.

I decided to put a warning on console to the User!

Problems when the openRA runs readonly:
- no settings will be written
- no Log files will be created
- Map editor is not functiontional

Tested so far: when Documents\OpenRA is HIDDEN
-> Download and installing content works as expected
-> Settings and Logs are not written
-> Screenshots will be saved correctly (the files are not hidden)

### DIFF
--- a/OpenRA.Game/MiniYaml.cs
+++ b/OpenRA.Game/MiniYaml.cs
@@ -22,7 +22,11 @@ namespace OpenRA
 	{
 		public static void WriteToFile(this MiniYamlNodes y, string filename)
 		{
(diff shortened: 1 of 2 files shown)


pchote:
As a general rule if you don't think a pr is useful, then please don't file it.  It is a drain on our overstretched resources, and generally makes the reviewers sad (we don't like rejecting prs).

LipkeGu:
this PR makes some parts of the game unsusable (map editor for example)

PR #9080, opened 2015-08-15
Replace icon veterancy overlay with new artwork.

pchote:
This replaces the RA2-style artwork with a much more fitting overlay made by MicroBit.

vet


Phrohdoh:
I cannot tell a difference so I don't care either way, :+1: since others seem to like it.

pchote:
For comparison the old overlay looked like:

GraionDilach:
Looks a bit small imo, but I can see the reason behind it.

Phrohdoh:
Oh, in that case :+1: the new art looks much more fitting.

MicroJOo:
its as small as the infantry-icons and has the same color as the bullet-shell from the rifle infantry , so it suits pretty much @ScottNZ 's idea of how it should look like for Ra1

cjshmyr:
:+1:

penev92:
The bigger, blended one may fit RA2, but I agree this is better for RA (although a bit small).

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/8fb1e8b3eed36abe9be2a4ccc4a96125690492b3...5ca740ffdde999858c391b50b6fc17b13d789dc7)

Mailaender:
The RA2 style overlays are not gone. Found them in https://github.com/Phrohdoh/OpenRA/commit/b07091e772a18fdd3e2f7628b2b8cf326a128cc9.

Phrohdoh:
I'm not sure why you'd expect RA2-style art to not be in RA2.

Mailaender:
I wanted @GraionDilach to know that his contribution isn't lost, only just hidden in a work-in-progress branch with a non-descriptive commit message "Holy mother-of-git-diffs, Batman! Import the RA2 mod files." which doesn't make it obvious to find.

Phrohdoh:
That commit message also doesn't note every actor, upgrade rule, artwork change, etc.
I think it is fine. There is a reason it isn't being merged into upstream.

GraionDilach:
I was already aware of that being there.

Note: I don't feel bad for seeing it replaced, I just still think that the new artwork blends in too well and not enough noticeable that "hey, you succeeded" - which is why I suggested the size tweaks. But with it getting in as is, I can't do but ¯\_(ツ)_/¯.

MicroJOo:
size should be fine , considering that i did all the ranks ... (and they have to fit too)

ranks2

PR #9083, opened 2015-08-16
Cleaned up the ILint boiler plate code

Mailaender:
as suggested in https://github.com/OpenRA/OpenRA/pull/8936#issuecomment-128945242.


abcdefg30:
:+1:

OpenRA.Mods.Common/Lint/CheckSequences.cs
```diff
@@ -18,7 +18,7 @@
 
 namespace OpenRA.Mods.Common.Lint
 {
-	class CheckSequences : ILintPass
+	class CheckSequences : ILintMapPass
```
penev92: Isn't this supposed to be an `ILintRulesPass`?
Also I see the `map != null` check is still present below, unlike all the other lint checks, why is that?

OpenRA.Mods.Common/UtilityCommands/CheckYaml.cs
```diff
@@ -52,7 +52,21 @@ public void Run(ModData modData, string[] args)
 				var maps = new List<Map>();
 				if (args.Length < 2)
 				{
-					maps.Add(null);
+					Console.WriteLine("Testing mod: {0}".F(Game.ModData.Manifest.Mod.Title));
+					CheckGameRules(Game.ModData.DefaultRules);
```
penev92: Please add a comment like "Run all rule checks on the default mod rules." (although somewhat obvious), and perhaps move it after the `foreach`?

OpenRA.Mods.Common/UtilityCommands/CheckYaml.cs
```diff
@@ -52,7 +52,21 @@ public void Run(ModData modData, string[] args)
...
+					foreach (var customPassType in Game.ModData.ObjectCreator.GetTypesImplementing<ILintPass>())
```
penev92: Please add a comment like "Run all generic (not mod-level) checks here."

OpenRA.Mods.Common/UtilityCommands/CheckYaml.cs
```diff
@@ -63,29 +77,21 @@ public void Run(ModData modData, string[] args)
...
-						Console.WriteLine("Testing map: {0}".F(testMap.Title));
-						testMap.PreloadRules();
-					}
-					else
-					{
-						Console.WriteLine("Testing mod: {0}".F(Game.ModData.Manifest.Mod.Title));
-					}
+					Console.WriteLine("Testing map: {0}".F(testMap.Title));
+					testMap.PreloadRules();
+					if (testMap.RuleDefinitions.Any() || testMap.VoiceDefinitions.Any() || testMap.WeaponDefinitions.Any())
```
penev92: This probably wants to account also for `SequenceDefinitions`, `VoxelSequenceDefinitions`, `NotificationDefinitions` and `SmudgeDefinitions`?

OpenRA.Mods.Common/UtilityCommands/CheckYaml.cs
```diff
@@ -63,29 +77,21 @@ public void Run(ModData modData, string[] args)
...
+						CheckGameRules(testMap.Rules, testMap);
 
-					foreach (var customPassType in Game.ModData.ObjectCreator
-						.GetTypesImplementing<ILintPass>())
+					foreach (var customMapPassType in Game.ModData.ObjectCreator.GetTypesImplementing<ILintMapPass>())
```
penev92: Please add a comment like "Run all map-level checks here."

OpenRA.Mods.Common/UtilityCommands/CheckYaml.cs
```diff
@@ -102,5 +108,22 @@ public void Run(ModData modData, string[] args)
 				Environment.Exit(1);
 			}
 		}
+
+		void CheckGameRules(Ruleset rules, Map map = null)
```
penev92: This is used for both mod and map rules, so how about just "CheckRules"?

OpenRA.Mods.Common/Lint/CheckSequences.cs
Mailaender: Sequences are rule, sequence and (map) tileset specific.

OpenRA.Mods.Common/UtilityCommands/CheckYaml.cs
Mailaender: No, that would be wrong wrong. Not all checks are run. Especially not the `ILintMapPass`.

OpenRA.Mods.Common/UtilityCommands/CheckYaml.cs
Mailaender: We don't add comments for obvious stuff according to our guidelines and we don't have any lint checks for the above definitions using `CheckGameRules`.

OpenRA.Mods.Common/UtilityCommands/CheckYaml.cs
Mailaender: And add a comment again that this is about MiniYaml game rules. ;)

OpenRA.Mods.Common/UtilityCommands/CheckYaml.cs
Mailaender: That isn't really describing anything. It already says `customMapPassType` and `ILintMapPass`. I also don't want to say when to use these, because people may come up with creative lint checks in the future.

penev92:
Well... I want those comments before I can +1...
I don't care that the code is "readable". I'd like to not spend 10 minutes on trying to figure out what it does. I'd like to be able to scroll through it and see what happens where.

Mailaender:
Please stay civil. It might make sense to add summaries on to the LintAttributes.

Mailaender:
Rebased to include https://github.com/OpenRA/OpenRA/pull/8749 and added some comments.

penev92:
Thanks! :+1: 
Very nice cleanup.

PR #9085, opened 2015-08-16
RA mod balance changes

ghost:
In response to the ongoing discussion at http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=18117
This is the first of a series of incremental balance changes as proposed by various members of the community.

Hijacker;
-cost increased from $500 to $600
-speed increased to 71 
-stealthed when not moving

Shock trooper;
-cost increased from $400 to $500
-health reduced to 80 from 100

Heavy tank
-cost reduced from $1150 to $1050
-health increased from 550 to 575

Yak
-chaingun spread increased from 128 to 298

### DIFF
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -431,14 +431,22 @@ HIJACKER:
 		BuildPaletteOrder: 110
 		Prerequisites: ~barr, fix, ~techlevel.medium
 	Valued:
-		Cost: 500
+		Cost: 600
 	Tooltip:
 		Name: Hijacker
 		Description: Hijacks enemy vehicles.\n  Unarmed
 	Health:
 		HP: 50
+	Mobile:
+		Speed: 71
 	RevealsShroud:
 		Range: 5c0
+	Cloak:
+		InitialDelay: 250
+		CloakDelay: 120
+		CloakSound:
+		UncloakSound:
+		UncloakOnMove: yes
 	Passenger:
 		PipType: Yellow
 	Captures:
@@ -459,12 +467,12 @@ SHOK:
 		BuildPaletteOrder: 130
 		Prerequisites: ~barr, stek, tsla, ~infantry.russia, ~techlevel.unrestricted
 	Valued:
-		Cost: 400
+		Cost: 500
 	Tooltip:
 		Name: Shock Trooper
 		Description: Elite infantry with portable tesla coils.\n  Strong vs Infantry, Vehicles\n  Weak vs Aircraft
 	Health:
-		HP: 100
+		HP: 80
 	Mobile:
 		Voice: Move
 	RevealsShroud:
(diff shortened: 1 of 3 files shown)


ScottNZ:
What is the motivation for increasing the heavy tank health slightly? It seems pretty insignificant.

r34ch:
It was a suggestion by Prince Blueblood with a number of people agreeing. The hp increase may not be enough, but that can be tweaked later.

Mailaender:
Syntax looks good. :+1:

pchote:
I'm not convinced about [PR: For stable].  The last time we sneaked in a simple-looking balance change late in the playtesting cycle (#6825) turned into a disaster .

MicroJOo:
i am not sure about the shocktrooper ... because he seems to be well balanced for "allies vs soviet" ...
i did read that there is a problem with ukraine vs russia ... because ukraine has problems defending against russia shocktroopers ... 

but we could find a different solution for this , other then nerfing shocktroopers and making them less good against allies too 
__________________________________

im fine with the hijacker buff .. this needs to be done 
__________________________________
heavy tank buff is like @ScottNZ said "insignificant" ... and i wouldnt recommend buffing him without buffing medium tanks too ... 
__________________________________
not sure about yak ... i have to test this to see the difference ... but if its a gently buff then i fine with it

PrinceBlueblood:
Heavy tank health get increased very sightly because some people keep complaining about the modified map when Heavy Tanks cost only $1000 and they still complains about how heavy tanks is too weak compared to the original RA heavy tanks in 550. Already tried 600 and people complaining about how it was too OP now it sucks, so I do an idiot's solution by put heavy tank health in 575 and everyone (at least in that fora) is happy now.

MicroJOo:
1. if people complain about the balance on a modified map , that is nothing we have to care about ... talk to the owner of the map
2. im sorry , but i didnt notice anyone complaining about the heavy-tanks ... i know that using tanks isnt that common anymore , but this affects both allies "Medium Tanks" and soviets "Heavy Tanks" ... 
this ... "buff" ... if you could call it that ... doesnt solve anything regarding enhancements on the tank-gameplay
3. a lot of units are different from the original ra , its not "OpenRA's" duty to recreate the original (unblanced) gameplay ...

r34ch:
I would urge discussion to occur in the forum thread to keep thoughts in one place.

MicroJOo:
speaking of ukraine vs russia ... 

ukraineruss

r34ch:
@MicroJOo Actually I completely agree. I was going to change the PR to nerf shockie health to a more sensible level (60hp) but also increase the charge time of the german chronosphere to 3 minutes, as both factions are simply too good.

After I want to tweak country specific units (buff phase tank dmg, demo truck speed, fake structures damage)

The final PR would be all the 'underused' units of the game (light tank, longbow, hijaker etc)


I will close this for now.

penev92:
It's actually much better to not do ALL.THE.UNITS! in one PR.
The majority seems to agree that shock troopers need to be nerfed, and perhaps you should PR that so it can get merged while you work on changing other things.

MicroJOo:
@penev92 : i agree

@r34ch : and you should make different poll - threads for every "problem" there is ... not discuss the whole ra-balancing in your shocktrooper-thread ...

Issue #8899, opened 2015-08-02
Tooltips leaking information.


Phrohdoh:
![screen shot 2015-08-02 at 11 50 32 am](https://cloud.githubusercontent.com/assets/4861023/9027192/f7bcc442-3910-11e5-9de2-39d4de940007.png)
![screen shot 2015-08-02 at 11 51 44 am](https://cloud.githubusercontent.com/assets/4861023/9027191/f7bbd6b8-3910-11e5-91cf-49503f2050ba.png)
![screen shot 2015-08-02 at 12 05 12 pm](https://cloud.githubusercontent.com/assets/4861023/9027193/f7be335e-3910-11e5-8229-8dd840de1331.png)

PR #9090, opened 2015-08-16
Fixed cloaked units getting revealed by tooltips

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/8899.

Test map: http://resource.openra.net/maps/4162/

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
@@ -123,6 +123,10 @@ namespace OpenRA.Mods.Common.Widgets
 
 			if (frozen != null)
 			{
+				var actor = frozen.Actor;
+				if (actor != null && actor.TraitsImplementing<IVisibilityModifier>().Any(t => !t.IsVisible(actor, world.RenderPlayer)))
+					return;
+
 				FrozenActorTooltip = frozen;
 				if (frozen.Actor != null)
 					ActorTooltipExtra = frozen.Actor.TraitsImplementing<IProvideTooltipInfo>().ToArray();


pchote:
Does this need a [PR: For stable]?

deniz1a:
What if RenderPlayer is null because of disable shroud cheat?

pchote:
If RenderPlayer is null then there will be no visible FrozenActors, so it doesn't matter.

OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
```diff
@@ -118,7 +118,7 @@ public void UpdateMouseover()
 			}
 
 			var frozen = world.ScreenMap.FrozenActorsAt(world.RenderPlayer, worldRenderer.Viewport.ViewToWorldPx(Viewport.LastMousePos))
-				.Where(a => a.TooltipInfo != null && a.IsValid)
+				.Where(a => a.TooltipInfo != null && a.IsValid && !a.Actor.TraitsImplementing<IVisibilityModifier>().Any(t => !t.IsVisible(a.Actor, world.RenderPlayer)))
```
RoosterDragon: Maybe there should be a method for this check (maybe an extension method), which this and ```CanViewActor``` could both use.

Mailaender: Good idea. Done.

OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
```diff
@@ -118,7 +118,7 @@ public void UpdateMouseover()
 			}
 
 			var frozen = world.ScreenMap.FrozenActorsAt(world.RenderPlayer, worldRenderer.Viewport.ViewToWorldPx(Viewport.LastMousePos))
-				.Where(a => a.TooltipInfo != null && a.IsValid)
+				.Where(a => a.TooltipInfo != null && a.IsValid && (a.Actor != null && !a.Actor.IsInvisibleTo(world.RenderPlayer)))
```
RoosterDragon: Extra parenthesis are not needed.
Also, a few lines down there is a ```if (frozen.Actor != null)``` check you can remove, because you are now filtering out null actors here.

RoosterDragon:
:+1: with minor comments addressed.

OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
Mailaender: Wait a moment. That will change behavior. Frozen actors can be dead (then return null on Actor) so if I filter them here, then you can use the tooltip to find out if they are which would add another exploit.

RoosterDragon: Ah - that's a very good point.

I'm at a bit of a loss as to how to solve that though.

Mailaender: What needs to be resolved anyway?

RoosterDragon: Here's my understanding (correct me if I've got this wrong as I'm somewhat confused right now):
- We want to run ```!a.Actor.IsInvisibleTo(world.RenderPlayer)``` to stop the tooltip appearing.
- We don't want to do ```a.Actor != null``` as that would introduce an issue if the frozen actor was dead.
- But we can't call ```IsInvisibleTo``` without checking if the actor is null first, so we can't satisfy both of the above requirements?

Mailaender: Shouldn't the bracket solve that? `(a.Actor != null && !a.Actor.IsInvisibleTo(world.RenderPlayer)`

RoosterDragon: The brackets don't change the logic here. ```a && (b && c)``` is equivalent to ```a && b && c```. All three conditions must be true for the whole thing to be true.

pchote: Why can't you just filter on whether the frozen actor has sprites?

Mailaender:
Updated the check logic.

OpenRA.Game/Actor.cs
```diff
@@ -295,6 +295,11 @@ public bool IsDisabled()
 			return false;
 		}
 
+		public bool IsInvisibleTo(Player player)
```
pchote: I don't like the idea of adding a new method that almost duplicates our existing `Player.CanViewActor`, but inverts everything about it.

Mailaender: I can't use `CanViewActor` as it includes fog visibility so this isn't duplication.

pchote: Then please make it consistent with the existing code.

Mailaender: I need the negative selector for filtering in both the ViewportControllerWidget and CanViewActor.

pchote: Please move this to Player to make it more consistent with the existing code.

pchote: Or perhaps better - just inline this where you need it.  This is a special case workaround, not a general-use function.

Mailaender: Removed. This essentially reverts what @RoosterDragon asked for.

abcdefg30:

obrakmann:
This happened in TD when I killed a FACT while hovering the mouse over it
```
  at OpenRA.Mods.Common.Widgets.ViewportControllerWidget.<UpdateMouseover>m__2 (OpenRA.Traits.FrozenActor a) [0x0003f] in /home/oliver/devel/openra/git/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs:121 
  at System.Linq.Enumerable+<CreateWhereIterator>c__Iterator1E`1[OpenRA.Traits.FrozenActor].MoveNext () [0x00000] in <filename unknown>:0 
  at OpenRA.Exts.CompareBy[FrozenActor,Int32] (IEnumerable`1 ts, System.Func`2 selector, Int32 modifier, Boolean throws) [0x00010] in /home/oliver/devel/openra/git/OpenRA.Game/Exts.cs:206 
  at OpenRA.Exts.MaxByOrDefault[FrozenActor,Int32] (IEnumerable`1 ts, System.Func`2 selector) [0x00005] in /home/oliver/devel/openra/git/OpenRA.Game/Exts.cs:196 
  at OpenRA.Traits.SelectableExts.WithHighestSelectionPriority (IEnumerable`1 actors) [0x00001] in /home/oliver/devel/openra/git/OpenRA.Game/SelectableExts.cs:53 
  at OpenRA.Mods.Common.Widgets.ViewportControllerWidget.UpdateMouseover () [0x00110] in /home/oliver/devel/openra/git/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs:122 
  at OpenRA.Mods.Common.Widgets.ViewportControllerWidget.Draw () [0x00002] in /home/oliver/devel/openra/git/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs:90 
  at OpenRA.Widgets.Widget.DrawOuter () [0x00013] in /home/oliver/devel/openra/git/OpenRA.Game/Widgets/Widget.cs:418 
  at OpenRA.Widgets.Widget.DrawOuter () [0x00033] in /home/oliver/devel/openra/git/OpenRA.Game/Widgets/Widget.cs:420 
  at OpenRA.Widgets.Widget.DrawOuter () [0x00033] in /home/oliver/devel/openra/git/OpenRA.Game/Widgets/Widget.cs:420 
  at OpenRA.Widgets.Widget.DrawOuter () [0x00033] in /home/oliver/devel/openra/git/OpenRA.Game/Widgets/Widget.cs:420 
  at OpenRA.Widgets.Ui.Draw () [0x00006] in /home/oliver/devel/openra/git/OpenRA.Game/Widgets/Widget.cs:77 
  at OpenRA.Game.RenderTick () [0x000b0] in /home/oliver/devel/openra/git/OpenRA.Game/Game.cs:522 
  at OpenRA.Game.Loop () [0x0013e] in /home/oliver/devel/openra/git/OpenRA.Game/Game.cs:641 
  at OpenRA.Game.Run () [0x00042] in /home/oliver/devel/openra/git/OpenRA.Game/Game.cs:659 
  at OpenRA.Program.Run (System.String[] args) [0x00011] in /home/oliver/devel/openra/git/OpenRA.Game/Support/Program.cs:117 
  at OpenRA.Program.Main (System.String[] args) [0x00050] in /home/oliver/devel/openra/git/OpenRA.Game/Support/Program.cs:39 
```

Mailaender:
Updated.

obrakmann:
:+1: 

Next thing to fix would need to be the attack cursors, then

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/23d61753b14fbbf84f483118c524432984d9e614)

Issue #9082, opened 2015-08-16
German localized TFD disc support


LipkeGu:
List of German CabFileIDs for all Mods

```
Cabinet: data1.hdr

File group: Tiberian Sun
MAPS01.MIX - 452
MAPS02.MIX - 453
MOVIES01.MIX - 455
MOVIES02.MIX - 456
SCORES.MIX - 464
TIBSUN.MIX - 473
e01scd01.mix - 483
e01scd02.mix - 484
expand01.mix - 485
gmenu.mix - 504
multi.mix - 506
patch.mix - 507
scores01.mix - 508
wdt.mix - 509
wdtvox.mix - 510

File group: Red Alert
EXPAND.MIX - 232
EXPAND2.MIX - 233
HIRES1.MIX - 235
LORES1.MIX - 238
REDALERT.MIX - 251
main.mix - 257

 File group: CnC
AUD.MIX - 39
CCLOCAL.MIX - 42
CONQUER.MIX - 45
DESEICNH.MIX - 46
DESERT.MIX - 47
GENERAL.MIX - 48
LOCAL.MIX - 49
SC-000.MIX - 53
SC-001.MIX - 54
SCORES.MIX - 55
SOUNDS.MIX - 56
SPEECH.MIX - 57
TEMPERAT.MIX - 58
TEMPICNH.MIX - 59
TRANSIT.MIX - 62
UPDATE.MIX - 63
UPDATEC.MIX - 64
WINTER.MIX - 66
WINTICNH.MIX - 67
covert\AUD.MIX - 215
covert\CONQUER.MIX - 216
covert\DESERT.MIX - 217
covert\GENERAL.MIX - 218
covert\MOVIES.MIX - 221
covert\SCORES.MIX - 225
covert\SOUNDS.MIX - 226
covert\TEMPERAT.MIX - 227
covert\WINTER.MIX - 228

File group: Red Alert 2
language.mix - 610
maps01.mix - 613
maps02.mix - 614
movies01.mix - 615
movies02.mix - 616
multi.mix - 618
ra2.mix - 623
theme.mix - 628
wdt.mix - 629

File group: Yury
expandmd01.mix - 737
langmd.mix - 739
mapsmd03.mix - 742
movmd03.mix - 743
multimd.mix - 745
ra2md.mix - 749
thememd.mix - 751

```

LipkeGu:
[...]

PR #8170, opened 2015-05-14
TFD Installer for Tiberian dawn, Red Alert and Tiberian Sun

DSUK:
Installs Tiberian dawn from the The First Decade Disk. Closes https://github.com/OpenRA/OpenRA/issues/3863.

PR #9093, opened 2015-08-16
Improve TFD Installation logic

LipkeGu:
This Pr adds more flexibilty to the TFD Installer when installing content from TFD discs. by using index lookup. Closes https://github.com/OpenRA/OpenRA/issues/9082. 

also this Pr pays attention to:  https://github.com/OpenRA/OpenRA/pull/8170#issuecomment-129621166

### DIFF
--- a/OpenRA.Game/FileSystem/InstallShieldCABExtractor.cs
+++ b/OpenRA.Game/FileSystem/InstallShieldCABExtractor.cs
@@ -40,7 +40,7 @@ namespace OpenRA.FileSystem
 			public FileGroup(Stream reader, long offset)
 			{
 				var nameOffset = reader.ReadUInt32();
-				/*   unknown  */ reader.ReadBytes(18);
+				/*  unknown  */ reader.ReadBytes(18);
 				FirstFile = reader.ReadUInt32();
 				LastFile = reader.ReadUInt32();
 
@@ -127,12 +127,12 @@ namespace OpenRA.FileSystem
 			{
 				reader.Seek(commonHeader.CabDescriptorOffset + 12, SeekOrigin.Begin);
 				FileTableOffset = (long)reader.ReadUInt32();
-				/*    unknown  */ reader.ReadUInt32();
+				/*  unknown  */ reader.ReadUInt32();
 				FileTableSize = reader.ReadUInt32();
 
 				FileTableSize2 = reader.ReadUInt32();
 				DirectoryCount = reader.ReadUInt32();
-				/*   unknown  */ reader.ReadBytes(8);
+				/*  unknown  */ reader.ReadBytes(8);
 				FileCount = reader.ReadUInt32();
 
 				FileTableOffset2 = (long)reader.ReadUInt32();
@@ -141,37 +141,37 @@ namespace OpenRA.FileSystem
 
 		struct FileDescriptor
 		{
-			public readonly ushort	Flags;
-			public readonly uint	ExpandedSize;
-			public readonly uint	CompressedSize;
-			public readonly uint	DataOffset;
+			public readonly ushort Flags;
+			public readonly uint ExpandedSize;
+			public readonly uint CompressedSize;
+			public readonly uint DataOffset;
 
-			public readonly byte[]	MD5;
-			public readonly uint	NameOffset;
-			public readonly ushort	DirectoryIndex;
-			public readonly uint	LinkToPrevious;
+			public readonly byte[] MD5;
+			public readonly uint NameOffset;
+			public readonly ushort DirectoryIndex;
+			public readonly uint LinkToPrevious;
 
-			public readonly uint	LinkToNext;
-			public readonly byte	LinkFlags;
-			public readonly ushort	Volume;
-			public readonly string	Filename;
+			public readonly uint LinkToNext;
+			public readonly byte LinkFlags;
+			public readonly ushort Volume;
+			public readonly string Filename;
 
 			public FileDescriptor(Stream reader, long tableOffset)
 			{
 				Flags = reader.ReadUInt16();
 				ExpandedSize = reader.ReadUInt32();
-				/*    unknown   */ reader.ReadUInt32();
+				/*  unknown  */ reader.ReadUInt32();
 				CompressedSize = reader.ReadUInt32();
 
-				/*  unknown */ reader.ReadUInt32();
-				DataOffset   = reader.ReadUInt32();
-				/*  unknown */ reader.ReadUInt32();
-				MD5          = reader.ReadBytes(16);
+				/*  unknown  */	reader.ReadUInt32();
+				DataOffset = reader.ReadUInt32();
+				/*  unknown  */ reader.ReadUInt32();
+				MD5 = reader.ReadBytes(16);
 
-				/*   unknown  */ reader.ReadBytes(16);
-				NameOffset     = reader.ReadUInt32();
+				/*  unknown  */ reader.ReadBytes(16);
+				NameOffset = reader.ReadUInt32();
 				DirectoryIndex = reader.ReadUInt16();
-				/*   unknown  */ reader.ReadBytes(12);
+				/*  unknown  */ reader.ReadBytes(12);
 				LinkToPrevious = reader.ReadUInt32();
 				LinkToNext = reader.ReadUInt32();
 
@@ -305,7 +305,7 @@ namespace OpenRA.FileSystem
 						fileOffset = head.FirstFileOffset;
 					}
 					else
-						throw new Exception("Cannot Resolve Remaining Stream");
+						throw new IOException("Cannot Resolve Remaining Stream");
 				}
 				else
 				{
@@ -352,7 +352,7 @@ namespace OpenRA.FileSystem
 
 			commonHeader = new CommonHeader(hdrFile);
 			cabDescriptor = new CabDescriptor(hdrFile, commonHeader);
-			/*    unknown   */ hdrFile.ReadBytes(14);
+			/*  unknown  */ hdrFile.ReadBytes(14);
 
 			for (var i = 0U; i < MaxFileGroupCount; ++i)
 				fileGroupOffsets.Add(hdrFile.ReadUInt32());
@@ -383,9 +383,10 @@ namespace OpenRA.FileSystem
 				for (var index = fileGroup.FirstFile; index <= fileGroup.LastFile; ++index)
 				{
 					AddFileDescriptorToList(index);
+					var dirSeperator = Path.DirectorySeparatorChar;
 					var fileDescriptor = fileDescriptors[index];
-					var fullFilePath   = "{0}\\{1}\\{2}".F(fileGroup.Name, DirectoryName((uint)fileDescriptor.DirectoryIndex), fileDescriptor.Filename);
-					fileLookup.Add(fullFilePath, index);
+					var fullFilePath = "{0}{3}{1}{3}{2}".F(fileGroup.Name, DirectoryName((uint)fileDescriptor.DirectoryIndex), fileDescriptor.Filename, dirSeperator);
+					fileLookup.Add(fullFilePath.Replace("\\\\", "{0}".F(dirSeperator)).ToLowerInvariant(), index);
 				}
 			}
 		}
@@ -404,9 +405,25 @@ namespace OpenRA.FileSystem
 			return test;
 		}
 
+		public uint GetIndexByFilename(string directory, string filename)
+		{
+			var index = uint.MinValue;
+
+			foreach (var entry in fileLookup.Keys)
+			{
+				if (entry.StartsWith(directory) && entry.EndsWith(filename))
+				{
+					index = fileLookup[entry];
+					break;
+				}
+			}
+
+			return index;
+		}
+
 		public bool Exists(string filename)
 		{
-			return fileLookup.ContainsKey(filename);
+			return fileLookup.ContainsKey(filename.ToLowerInvariant());
 		}
 
 		public uint DirectoryCount()
@@ -461,7 +478,7 @@ namespace OpenRA.FileSystem
 		{
 			var fileDes = fileDescriptors[index];
 			if ((fileDes.Flags & FileInvalid) != 0)
-				throw new Exception("File Invalid");
+				throw new InvalidDataException("File Invalid");
 
 			if ((fileDes.LinkFlags & LinkPrev) != 0)
 				return GetContentById(fileDes.LinkToPrevious);
@@ -475,7 +492,7 @@ namespace OpenRA.FileSystem
 				reader.CopyTo(output);
 
 			if (output.Length != fileDes.ExpandedSize)
-				throw new Exception("Did not fully extract Expected = {0}, Got = {1}".F(fileDes.ExpandedSize, output.Length));
+				throw new IOException("Did not fully extract! (Expected = {0}, Got = {1})".F(fileDes.ExpandedSize, output.Length));
 
 			output.Position = 0;
 			return output;
@@ -485,7 +502,7 @@ namespace OpenRA.FileSystem
 		{
 			var fileDes = fileDescriptors[index];
 			if ((fileDes.Flags & FileInvalid) != 0)
-				throw new Exception("File Invalid");
+				throw new InvalidDataException("File Invalid");
 
 			if ((fileDes.LinkFlags & LinkPrev) != 0)
 			{
@@ -500,7 +517,7 @@ namespace OpenRA.FileSystem
 				reader.CopyTo(output);
 
 			if (output.Length != fileDes.ExpandedSize)
-				throw new Exception("Did not fully extract Expected = {0}, Got = {1}".F(fileDes.ExpandedSize, output.Length));
+				throw new IOException("Did not fully extract! (Expected = {0}, Got = {1})".F(fileDes.ExpandedSize, output.Length));
 		}
 
 		public Stream GetContent(string fileName)
--- a/OpenRA.Mods.Common/InstallUtils.cs
+++ b/OpenRA.Mods.Common/InstallUtils.cs
@@ -36,11 +36,11 @@ namespace OpenRA.Mods.Common
 		public readonly string MusicPackageMirrorList = null;
 		public readonly int ShippedSoundtracks = 0;
 
-		/// <summary> InstallShield .cab File Ids, used to extract Mod specific files. </summary>
-		public readonly int[] InstallShieldCABFileIds = { };
+		/// <summary> InstallShield PackageName, used to extract Mod specific files </summary>
+		public readonly string InstallShieldPackageName;
 
-		/// <summary> InstallShield .cab File Ids, used to extract Mod specific archives and extract contents of ExtractFilesFromCD. </summary>
-		public readonly string[] InstallShieldCABFilePackageIds = { };
+		/// <summary> InstallShield Package file list, used to lookup the index of Mod specific files </summary>
+		public readonly string[] ExtractFromInstallShieldPackage = { };
 	}
 
 	public static class InstallUtils
--- a/OpenRA.Mods.Common/Widgets/Logic/Installation/InstallFromCDLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Installation/InstallFromCDLogic.cs
@@ -52,15 +52,9 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			return installData.DiskTestFiles.All(f => File.Exists(Path.Combine(diskRoot, f)));
 		}
 
-		bool IsTFD(string diskpath)
+		bool IsTheFirstDecadeDisk(string diskpath)
 		{
-			var test = File.Exists(Path.Combine(diskpath, "data1.hdr"));
-			var i = 0;
-
-			while (test && i < 14)
-				test &= File.Exists(Path.Combine(diskpath, "data{0}.cab".F(++i)));
-
-			return test;
+			return File.Exists(Path.Combine(diskpath, "data1.hdr"));
 		}
 
 		void CheckForDisk()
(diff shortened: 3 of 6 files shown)


GraionDilach:
:-1:. You were told why supporting multiple languages at TFD is a bad idea.

LipkeGu:
It changes only the behaver Oo how the files gets detected. it adds no language support

Unit158:
Looks reasonable. Will test tomorrow.

OpenRA.Game/FileSystem/InstallShieldCABExtractor.cs
```diff
@@ -404,6 +404,16 @@ public string DirectoryName(uint index)
 			return test;
 		}
 
+		public uint GetIndexFromFilename(string directory, string filename)
+		{
+			var path = directory + "\\" + filename;
```
Mailaender: I am not sure as this is inside a CAB file, but `\` is the Windows path separator. Hardcoding it will break everywhere else. I suggest you `Path.DirectorySeparatorChar` instead.

LipkeGu: https://gist.github.com/DSUK/e243103217bac918b175 yes they are inside... you say in this case give me the IDs of the files... not give me the files of the IDs.

LipkeGu:
updated

mods/cnc/mod.yaml
```diff
@@ -153,7 +153,8 @@ ContentInstaller:
 		.: conquer.mix, desert.mix, general.mix, scores.mix, sounds.mix, temperat.mix, winter.mix
 	ShippedSoundtracks: 4
 	MusicPackageMirrorList: http://www.openra.net/packages/cnc-music-mirrors.txt
-	InstallShieldCABFileIds: 66, 45, 42, 65, 68, 67, 71, 47, 49, 60, 75, 73, 53
+	InstallShieldCABDirectory: CnC
+	FiletoExtractFromCab: SPEECH.MIX, CONQUER.MIX:CRC32, CCLOCAL.MIX:CRC32, SOUNDS.MIX, TEMPICNH.MIX, TEMPERAT.MIX, TRANSIT.MIX, DESERT.MIX, GENERAL.MIX, SCORES.MIX, WINTER.MIX, UPDATEC.MIX
```
pchote: What is the purpose of these `CRC32`s?

LipkeGu: hash checks which was already there :(

LipkeGu: should i remove them?

mods/ra/mod.yaml
```diff
@@ -152,8 +152,8 @@ ContentInstaller:
 		.: INSTALL/REDALERT.MIX
 	ShippedSoundtracks: 2
 	MusicPackageMirrorList: http://www.openra.net/packages/ra-music-mirrors.txt
-	InstallShieldCABFilePackageIds: 105
-	InstallShieldCABFileIds: 116
+	InstallShieldCABDirectory: Red Alert
+	FiletoExtractFromCab: redalert.mix:crc32, main.mix
```
Mailaender: CRC32 encryption is TS/RA2 era mix files.

LipkeGu:
updated

OpenRA.Mods.Common/InstallUtils.cs
```diff
@@ -39,11 +39,8 @@ public class ContentInstaller : IGlobalModData
 		public readonly string MusicPackageMirrorList = null;
 		public readonly int ShippedSoundtracks = 0;
 
-		/// <summary> InstallShield .cab File Ids, used to extract Mod specific files </summary>
-		public readonly int[] InstallShieldCABFileIds = { };
-
-		/// <summary> InstallShield .cab File Ids, used to extract Mod specific archives and extract contents of ExtractFilesFromCD </summary>
-		public readonly string[] InstallShieldCABFilePackageIds = { };
+		public readonly string InstallShieldCABDirectory = null;
+		public readonly string[] FiletoExtractFromCab = { };
```
penev92: Please keep with the `InstallShieldCAB*` naming convention set for these.
Also don't remove the `<summary>`-ies.

Oh, and the ` = null` is not necessary.

mods/ra/mod.yaml
```diff
@@ -152,8 +152,8 @@ ContentInstaller:
 		.: INSTALL/REDALERT.MIX
 	ShippedSoundtracks: 2
 	MusicPackageMirrorList: http://www.openra.net/packages/ra-music-mirrors.txt
-	InstallShieldCABFilePackageIds: 105
-	InstallShieldCABFileIds: 116
+	InstallShieldCABDirectory: Red Alert
+	FiletoExtractFromCab: redalert.mix, main.mix
```
penev92: I see 1 ID above turned into 2 names here. What's going on?

mods/cnc/mod.yaml
```diff
@@ -153,7 +153,8 @@ ContentInstaller:
 		.: conquer.mix, desert.mix, general.mix, scores.mix, sounds.mix, temperat.mix, winter.mix
 	ShippedSoundtracks: 4
 	MusicPackageMirrorList: http://www.openra.net/packages/cnc-music-mirrors.txt
-	InstallShieldCABFileIds: 66, 45, 42, 65, 68, 67, 71, 47, 49, 60, 75, 73, 53
+	InstallShieldCABDirectory: CnC
+	FiletoExtractFromCab: SPEECH.MIX, CONQUER.MIX, CCLOCAL.MIX, SOUNDS.MIX, TEMPICNH.MIX, TEMPERAT.MIX, TRANSIT.MIX, DESERT.MIX, GENERAL.MIX, SCORES.MIX, WINTER.MIX, UPDATEC.MIX
```
penev92: 13 IDs turned into 12 file names here?

mods/ra/mod.yaml
LipkeGu: index lookup for this files

mods/cnc/mod.yaml
LipkeGu: CabinetExtrator: Implement GetIndexByFilename

LipkeGu:
has dependencies

abcdefg30:
On what?

mods/cnc/mod.yaml
DSUK: it's called MOVIES.MIX on TFD

mods/ra/mod.yaml
penev92: Which TFD disks?

mods/cnc/mod.yaml
LipkeGu: @DSUK : thats why there is no movies.mix either movies***.mix Europe Disks has extracted vqa files

penev92: Which is exactly why people told you on GH and IRC that we can't support them.

LipkeGu:
updated

LipkeGu:
dependencies was related to RA2 but are now solved

OpenRA.Game/FileSystem/InstallShieldCABExtractor.cs
```diff
@@ -404,9 +405,44 @@ public string DirectoryName(uint index)
 			return test;
 		}
 
+		public string GetFirstSubDirectory(string package)
+		{
+			var dir = "";
+			var delimeter = "{0}".F(Path.DirectorySeparatorChar);
+
+			foreach (var key in fileLookup.Keys)
+				if (key.StartsWith(package.ToLowerInvariant()) && key.Contains(delimeter))
```
pchote: Style nit: avoid hanging blocks.  Put braces around this please.

OpenRA.Game/FileSystem/InstallShieldCABExtractor.cs
```diff
@@ -404,9 +405,44 @@ public string DirectoryName(uint index)
...
+			return dir;
+		}
+
+		public uint GetIndexByFilename(string directory, string filename)
+		{
+			var path = directory + Path.DirectorySeparatorChar + filename;
+
+			if (Exists(path.ToLowerInvariant()))
+				return fileLookup[path.ToLowerInvariant()];
+			else
```
pchote: Style nit: this `else` is unnecessary (you can unconditionally throw if it doesn't return).

OpenRA.Mods.Common/Widgets/Logic/Installation/InstallFromCDLogic.cs
```diff
@@ -91,44 +85,63 @@ void InstallTFD(string source)
...
+
+						dest = Platform.ResolvePath("^", "Content", Game.ModData.Manifest.Mod.Id, filename.ToLowerInvariant());
+						cabExtractor.ExtractFile(fileindex, dest);
 						progressBar.Percentage += installPercent;
 					}
 
-					var ArchivesToExtract = installData.InstallShieldCABFilePackageIds.Select(x => x.Split(':'));
+					progressBar.Indeterminate = true;
+
+					var ArchivesToExtract = installData.ExtractFromInstallShieldPackage.Select(x => x.Split(':'));
```
pchote: style nit: local variables shouldn't start with a capital letter.

mods/cnc/mod.yaml
```diff
@@ -153,7 +153,8 @@ ContentInstaller:
 		.: conquer.mix, desert.mix, general.mix, scores.mix, sounds.mix, temperat.mix, winter.mix
 	ShippedSoundtracks: 4
 	MusicPackageMirrorList: http://www.openra.net/packages/cnc-music-mirrors.txt
-	InstallShieldCABFileIds: 66, 45, 42, 65, 68, 67, 71, 47, 49, 60, 75, 73, 53
+	InstallShieldPackageName: CnC
+	ExtractFromInstallShieldPackage: SPEECH.MIX, CONQUER.MIX, CCLOCAL.MIX, SOUNDS.MIX, TEMPICNH.MIX, TEMPERAT.MIX, TRANSIT.MIX, DESERT.MIX, GENERAL.MIX, SCORES.MIX, WINTER.MIX, UPDATEC.MIX
```
pchote: The case changes across mods are worrying.  Somebody with a case sensitive filesystem will need to test this before we merge.

mods/ts/mod.yaml
```diff
@@ -190,18 +190,18 @@ LoadScreen: LogoStripeLoadScreen
...
-	TestFiles: cache.mix, conquer.mix, isosnow.mix, isotemp.mix, local.mix, sidec01.mix, sidec02.mix, sno.mix, snow.mix, sounds.mix, speech01.mix, tem.mix, temperat.mix
+	TestFiles: cache.mix, conquer.mix, isosnow.mix, isotemp.mix, local.mix, sidec01.mix, sidec02.mix, sno.mix, snow.mix, sounds.mix, speech01.mix, speech02.mix, tem.mix, temperat.mix
 	PackageMirrorList: http://www.openra.net/packages/ts-mirrors.txt
 	DiskTestFiles: MULTI.MIX, INSTALL/TIBSUN.MIX
 	CopyFilesFromCD:
 		.: INSTALL/TIBSUN.MIX, SCORES.MIX, MULTI.MIX
 	PackageToExtractFromCD: INSTALL/TIBSUN.MIX:CRC32
 	ExtractFilesFromCD:
-		.: cache.mix, conquer.mix, isosnow.mix, isotemp.mix, local.mix, sidec01.mix, sidec02.mix, sno.mix, snow.mix, sounds.mix, speech01.mix, tem.mix, temperat.mix
+		.: cache.mix, conquer.mix, isosnow.mix, isotemp.mix, local.mix, sidec01.mix, sidec02.mix, sno.mix, snow.mix, sounds.mix, speech01.mix, speech02.mix, tem.mix, temperat.mix
```
pchote: These speech02 changes aren't related to the TFD changes, so should live in their own commit or (preferably, assuming they are fixing a bug) in their own PR.

OpenRA.Game/FileSystem/InstallShieldCABExtractor.cs
```diff
@@ -404,9 +405,44 @@ public string DirectoryName(uint index)
 			return test;
 		}
 
+		public string GetFirstSubDirectory(string package)
```
pchote: What is the purpose of this method?  The implementation looks to be DIYing some path manipulation that the standard library should provide methods for.

penev92: >German disks has 13 cabinets. so i decided to check for data1.hdr only.

TLDR he's hacking it because German and non-german TFD contents is different.
Which, again, is the exact reason why people have been saying we shouldn't try to support them.

pchote: I'm deliberately ignoring the motivations because replacing the opaque file IDs with filenames is a good idea.

pchote: Ah, I see.  That should not be included in this PR.  Replacing the opaque and arbitrary file IDs with filenames is good, but we are not ready to support non-english disks OR RA2 right now.

penev92:
OK let's make the purpose of this PR clear, because "Improve TFD Installation logic" and "adds more flxibility" aren't.

This should replace the file IDs in YAML with file names.
This should **not** add support for non-english disks.
This should **not** try to add support for RA2, if that includes any hacks in the code.

GraionDilach:
@penev92: Except that it does _exactly_ what you listed, nothing.

mods/ts/mod.yaml
LipkeGu: no it was a problem in the first tfd PR :/ but nobody found the problem... so i added it here also :(

penev92: Can you explain this? Because this line has nothing to do with installation from a TFD DVD, and was not touched by the PR that introduced support for TFD.

LipkeGu:
@pchote i compared the us disk with the german disk and there are no differences. all filenames are the same. but only the folder names are different. They are Binary Compatible also mailaender sayed that (he is an german guy) and for years i play with german TS disks ...

The intentions to improve that Installer was not to add GERMAN support it was mainly an idea to add generic support for ALL TFD Disks. What Im not understandig is we say in the News.

"Added support for TFD disks"... that means for german users "support for all disks" we did not say "ONLY FOR US Disks" but now we have in the first PR 2 comments about that german disks will not be able to use because of the IDs" because we wrote "support for TFD disks".

When we did not support german disks TFD disks why we does we support german versions of RA,TS and CNC?!?! since years... ?!?! many peoples are playing with german disks and there is no problems on this site. binary compatible

In this case when we add support for TFD then we should not limit this support for US only instead we should support EVERY TFD disk ... not only for I2 types of disks. we have all our Legals and rules, but it should not be limited to this statement: they reason why we does not support german disks are: german has oil instead of blood so thats why we doesnt support german disks... Oo
that statement was written... in the issue where i posted the german IDs of the files...

OpenRA.Game/FileSystem/InstallShieldCABExtractor.cs
LipkeGu: that else handles conditions like when files can not be found in the file table... it must return indexe i can not return 0 because zero stays for a another file in the lookup table

mods/cnc/mod.yaml
LipkeGu: why we have to test this?! the filenames will always lowercase

https://github.com/LipkeGu/OpenRA/commit/54f878993b02c2d8e3ddcd98ba01dcd74734c056#diff-d7d7b22ed1d8f9159c5bc9aa83fdb4d0R437

mods/ts/mod.yaml
LipkeGu: you self sayed speech02 is missing on TS

OpenRA.Game/FileSystem/InstallShieldCABExtractor.cs
LipkeGu: thats does not enable german support

LipkeGu:
closing for now because there is no intention for this pr

penev92:
Well OK, since we discussed it on IRC last night and figured out the misunderstanding I don't know why you'd close it before someone tested it, but OK.

LipkeGu:
as dicussed in IRC

<penev> we keep the ID lookup by name (which is nice); we add a check if this is an english DVD; if not, we issue a warning along the lines of "non-english versions of TFD are not supported. Install at your own risk!"
<GraionDilach> that's acceptable```
```

**Comment by LipkeGu** [comment]
updated

**Comment by GraionDilach** [comment]
I can't find where it throws the warning. That change is still requested.

**Comment by LipkeGu** [comment]
That notice should be in the changelog

File: OpenRA.Game/FileSystem/InstallShieldCABExtractor.cs [hunk 13]
```diff
@@ -475,7 +492,7 @@ public Stream GetContentById(uint index)
 				reader.CopyTo(output);
 
 			if (output.Length != fileDes.ExpandedSize)
-				throw new Exception("Did not fully extract Expected = {0}, Got = {1}".F(fileDes.ExpandedSize, output.Length));
+				throw new IOException("Did not fully extract Expected = {0}, Got = {1}".F(fileDes.ExpandedSize, output.Length));
```
penev92: Fullstop or exclamation mark after "extract" please.

OpenRA.Game/FileSystem/InstallShieldCABExtractor.cs
```diff
@@ -500,7 +517,7 @@ public void GetContentById(uint index, Stream output)
...
```
penev92: And here.

penev92:
No. We were over this on IRC. A warning in the changelog makes no difference.
In case @pchote and @Mailaender are OK with the suggestion, it can go in the game. Otherwise we can't take this PR, as discussed on IRC yesterday.

LipkeGu:
yes then we need message box support in OpenRA =(

GraionDilach:
You can add a label to the content copy process dialog where you can write the warning message in capslock.

LipkeGu:
@GraionDilach : Only RA needs this warning: http://www.schnittberichte.com/schnittbericht.php?ID=3406

there is an missing movie ... tiberian sun has an changed movie and generals is completely changed :(

LipkeGu:
updated

Mailaender:
I don't get your obsessions with warnings about non-existing incompatibilities. Those assets are binary compatible. Even the Generals censorship could be undone with content patches only. Localisation is often done after the game has gone gold by non-technical people or it is outsourced. That is the reason we don't get into trouble here.

This unhardcodes ugly file offsets to make this work with any TFD disc. :+1: It makes the whole thing much more robust and sane. Can't test as I don't own the DVD myself.

GraionDilach:
Binary compatible in Westwood terms means that the executable, art.ini, rules.ini, sound.ini and the map is the same. Everything else can differ.

The problem is that OpenRA does not react to the same for missing assets as the Westwood games. If an asset lost a frame, got removed because being unused, then the WW games will deal with it, but OpenRA throws a crash.

I still haven't seen any of you actually checking out to see if each file still exists within the localized mixes and that's just the German version. There might be more versions as well.

Mailaender:
Please stop spreading that FUD. We have never received a crash report so far when someone used localized mix files even though there are many and even fan made ones out on the net. The worst thing that could happen here is spelling mistakes in icons and terrible voice actors.

LipkeGu:
Changelog could be:

```
The compatibility with the Content Installer and "The First Decade" disks has been improved. Please keep in mind that the content in Europe can differ.
```

Mailaender:
> yes then we need message box support in OpenRA =(

We already have `ConfirmationDialogs.PromptConfirmAction`.

LipkeGu:
and depends label please because the asset code is moving....

penev92:
Doesn't look like there will be any conflicts between the two PRs other than that moving `ContentInstaller` to a new file and this one changing 2 lines in it, which is a very simple rebase regardless of which one gets merged first.

LipkeGu:
Otherwise we can move this together with your PR

penev92:
Reasons against doing that:
 - My PR is big enough as it is and I wouldn't want to force more stuff on reviewers than is necessary.
 - I am not familiar with the changes in this PR, so I'm not comfortable with merging it in my own.
 - Both PRs are perfectly fine on their own as they have very little in common.

Mailaender:
This just needs a +1 from people who have a TFD disc and can confirm that this works properly on every platform.

LipkeGu:
On Windows it works so as far i can tell...

LipkeGu:
updated and merged into one commit!

OpenRA.Mods.Common/Widgets/Logic/Installation/InstallFromCDLogic.cs
```diff
@@ -91,44 +85,74 @@ void InstallTFD(string source)
 			{
 				using (var cabExtractor = new InstallShieldCABExtractor(source))
 				{
-					var denom = installData.InstallShieldCABFileIds.Length;
+					var dest = "";
+					var files = installData.ExtractFromInstallShieldPackage.Select(x => x.Split(':'));
 					var extractFiles = installData.ExtractFilesFromCD;
+					var installPercent = 100 / files.Count();
+					var fileindex = (uint)0;
```
Phrohdoh: `0u`.

Mailaender:
Needs a rebase.

pchote:
This pr actually does three different things, so it would be useful if you could separate this into three commits:

1. Code style changes in `InstallShieldCABExtractor`.
2. Replacing index &rarr; filename.
3. Fixing speech02.mix.

steelphase:
It may be better as a separate PR, but this could fix #9480.

penev92:
I think this is worth someone taking over at this point if @LipkeGu is missing.
Even if just for the ID -> file name change.

DSUK:
tried installing this on my machine and got:
Exception of type `System.IO.DirectoryNotFoundException`: Could not find a part of the path "/home/james/.openra/content/ra/redalert.mix".

(Note: content has lowercase c here, There is Content with an uppercase C in my ~/.openra directory)

LipkeGu:
Im not missing im looking at this at the Weekend :)

abcdefg30:

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/11375.

PR #9096, opened 2015-08-17
Hides crates under fog.

deniz1a:
Crates are hidden under fog of war like units. So to find them, scouting of the map is required. Also adds a satellite icon for crates. They are shown as white empty squares.

crate_icon

This is the enlarged view of the image I used for the icon:
crate_icon_source

What I don't understand however is that gpsdot.shp got smaller after I added the crate image. How did that happen?

### DIFF
--- a/mods/cnc/rules/defaults.yaml
+++ b/mods/cnc/rules/defaults.yaml
@@ -691,7 +691,7 @@
 
 ^Crate:
 	Inherits@1: ^SpriteActor
-	HiddenUnderShroud:
+	HiddenUnderFog:
 	Tooltip:
 		Name: Crate
 		GenericName: Crate
--- a/mods/d2k/rules/misc.yaml
+++ b/mods/d2k/rules/misc.yaml
@@ -1,5 +1,5 @@
 crate:
-	HiddenUnderShroud:
+	HiddenUnderFog:
 	Tooltip:
 		Name: Crate
 	Crate:
--- a/mods/ra/rules/defaults.yaml
+++ b/mods/ra/rules/defaults.yaml
@@ -671,7 +671,7 @@
 
 ^Crate:
 	Inherits@1: ^SpriteActor
-	HiddenUnderShroud:
+	HiddenUnderFog:
 	Tooltip:
 		Name: Crate
 		GenericName: Crate
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -157,7 +157,7 @@
 		RequireTilesets: TEMPERATE
 
 ^Crate:
-	HiddenUnderShroud:
+	HiddenUnderFog:
 	Tooltip:
 		Name: Crate
 	Crate:


deniz1a:
But they are not easy to see on snow:
crate_icon_snow

GraionDilach:
The idea ain't bad. The colors make no sense though. Be consistent, use the same colors as the other GPS dots.

deniz1a:
GPS dots don't have a color by themselves. They are remapped with the player's color. Crates are neutral and neutral units have white color, that's why the icons are white. If there was a custom map with player owned crates then those icons would have player colors too.

GraionDilach:
You can use non-remapable colors as well though. Color #124, for example would work.

penev92:
OK, as far as I know, the situation on whether crates should be hidden under fog is controversial. So we may or may not agree on that this time around.
But having them show on the GPS is a no-no.

deniz1a:
Why not? Satellite shows every unit and building. Crates are like neutral units.

penev92:
Also if people do agree to hide them, we should hide them in all mods.

deniz1a:
Ok, I'll add that to all other mods.

deniz1a:
D2k doesn't have crates.

MicroJOo:
i do not like the idea to hide the crates ... it could be important for a player , who lost his mcv ... to keep track (atleast) of crates on a scouted map ... so he may reestablish (still needs luck though) his base ...

i see crates as neutral support-objects ... therefor , as long as they are vital for players to stay in the game ... i would like to keep this exception for them , being visible under the fog 

:-1:

penev92:
So far we have @MicroJOo voting for no changes, and me and @pchote voting for hiding under fog, and GPS not detecting them.
While it would be nice more people to share an opinion, changes are definitely wanted here.
D2k **does** have crates.

GraionDilach:
Have to say, I fully support this PR.

Crates are neutral objects and therefore __not__ eligible to be seen through fog. Even new buildings doesn't pop up, then why should crates?

Regarding GPS, either the crates either the plane dropping them would pop up. Having an exception with that Badger again sounds odd.

I acknowledge that I never looked at OpenRA MP, but back at RA2 MP, crates were never enabled for serious games - they process a random factor which is independent of the player's skill level, which rendered them out from tourny rules pretty much.

TLDR: Crates are not that special that they should be seen through the fog, although the GPS should be aware of at least the plane dropping them.

deniz1a:
Ok I also think badger planes should have GPS icons. But then so should parabomb badgers, to be consistent. And I too think that crates shouldn't be enabled in tournaments. But if crates are enabled, they are important, can be game changing. When you lose construction yard, there is a high chance of getting an MCV from a crate.

LipkeGu:
Before anyonebegins to murdering each other we could make this optional...

MicroJOo:
would like to wait for @pchote 's +1 here ...

xp65:
As crates early on in the game can quickly change the balance, I vote for hiding it below the fog.

pi99y:
I vote for hidden crates without GPS detection. 
 
I do sympathize with @MicroJOo's point - so maybe game could even reveal crates (perhaps even with signal beacons) to a player who loses their last MCV.

MicroJOo:
i have changed my mind on this ... @GraionDilach 's : "Crates are neutral objects and therefore not eligible to be seen through fog. Even new buildings doesn't pop up, then why should crates?" convinced me ...  i guess im just having a hard time to break habits 

but the satellite-icon for the crate looks ugly ... !

deniz1a:
Ok I will separate this into three PRs:
1) Hide crates under fog
2) Add satellite icon to crates
3) Add satellite icon to badger planes

deniz1a:
I'll open the PR to add satellite icon for crates after this one is merged.

r34ch:
Please do not give crates a GPS icon. The badger planes yes, but little wooden boxes on the ground? Besides, from a pure gameplay balancing point of view, having crates on GPS simply rewards the allied faction for constructing a building.

I don't really mind the current implementation. From a canon point of view I view the supply drops as coming from Red Alerts variant of the UN. UN generally informs all factions of their presence in an area, so to see neutral badger supply drops (and the crates) through FoW was fine for me.

A factions badger bomber or badger paradrop plane should be hidden though except for the GPS dots.

penev92:
"No" to 2) and 3)
I think we've already argued about those.

GraionDilach:
@r34ch:
>From a canon point of view I view the supply drops as coming from Red Alerts variant of the UN. UN generally informs all factions of their presence in an area, so to see neutral badger supply drops (and the crates) through FoW was fine for me.

The UN born after WW2 and the precedessor Leauge of Nations omitted USSR, US and a lot more. Your "canon POV" is nonsense.

@penev92 : Why no to 3)? From what I see, 2) gets rejected alone but not 3.

r34ch:
@GraionDilach Fair point. Although I did state RAs _equivalent_ to the UN, I guess I was inventing my own canon...

deniz1a:
So, changes requested label can be removed and needs +2 label can be added now, right?

deniz1a:
Here's a test map for this: http://resource.openra.net/maps/4446

pchote:
:+1: for the `HiddenUnderShroud` &rarr; `HiddenUnderFog` change to crates.  This review is nullified if the PR is rewritten or changed.

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/cfff98fbde4c9ea88a4180253a1c78fabe9d620a...0dad6a3eec01f65fd61602df64a4d42fb785cbde)

PR #9097, opened 2015-08-17
Fix TerrainTypes for walls and crates

LipkeGu:
This Pr fixes a bug where walls can not be placed on some terrainTypes 

good example are the new TS maps. where we cant place walls on terrain types like `Sand, Green, Road, Pavement, Rough` and so on.


abcdefg30:
:+1:

Mailaender:
Thanks.

PR #9099, opened 2015-08-17
Fix WithCargo desync.

### LINKED ISSUE #9094: Desync when Cargo.InitialUnits is used
Game Crash when i build modded Plane
 on this version: "Civillian Outrage BETA 48a"   http://resource.openra.net/maps/4168
out of sync 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25838582-desync-when-cargo-initialunits-is-used?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Adjusting world state from render code is a bad idea.
Closes #9094.

### DIFF
--- a/OpenRA.Mods.Cnc/Traits/Render/WithCargo.cs
+++ b/OpenRA.Mods.Cnc/Traits/Render/WithCargo.cs
@@ -29,7 +29,7 @@ namespace OpenRA.Mods.Cnc.Traits
 		public object Create(ActorInitializer init) { return new WithCargo(init.Self, this); }
 	}
 
-	public class WithCargo : IRenderModifier
+	public class WithCargo : IRenderModifier, ITick
 	{
 		readonly Cargo cargo;
 		readonly IFacing facing;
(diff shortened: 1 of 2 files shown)


Mailaender:
This probably solves the problem I had when I adding manual Carryalls to Dune 2000. :+1:

OpenRA.Mods.Cnc/Traits/Render/WithCargo.cs
```diff
@@ -45,6 +45,16 @@ public WithCargo(Actor self, WithCargoInfo info)
 			body = self.Trait<IBodyOrientation>();
 		}
 
+		public void Tick(Actor self)
+		{
+			foreach (var c in cargo.Passengers)
+			{
+				var cargoFacing = c.TraitOrDefault<IFacing>();
+				if (facing != null && cargoFacing != null)
```
RoosterDragon: Can we lift the null check on facing out of the loop?

Mailaender: How? The trait lookup is done on the passengers who are looped upon.

pchote: `facing` is a trait on the transport.

Mailaender: Right. You can actually just skip the whole loop then. Mind to do that simple change so we can release?

Mailaender:
This is taking too long already. Superseded by https://github.com/OpenRA/OpenRA/pull/9172.

PR #9100, opened 2015-08-17
stop "Unit Ready" sound spaming when Instant Buildtime is enabled

LipkeGu:
By adding the missing condition... Normal mode works as expected...


abcdefg30:
Now I don't get any `unit ready` notifications when building with the dev cheats.

LipkeGu:
thats because every unit is done under 1 second :)

Unit158:
I'm really not convinced this is the correct way to fix this issue. I have an idea that might work a little bit better.

LipkeGu:
lets hear

PR #9102, opened 2015-08-18
Makes trees destructable.

deniz1a:
They have nice damaged state images and death animations which were not used. They can be targeted using force fire.

### DIFF
--- a/mods/cnc/rules/defaults.yaml
+++ b/mods/cnc/rules/defaults.yaml
@@ -576,6 +576,11 @@
 		HP: 500
 	Armor:
 		Type: Wood
+	WithDeathAnimation:
+		DeathSequence: dead
+		UseDeathTypeSuffix: false
+	TargetableBuilding:
+		TargetTypes: Ground
 	AutoTargetIgnore:
 	BodyOrientation:
 	FrozenUnderFog:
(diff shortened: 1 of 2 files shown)


pchote:
This is a bad idea.  We briefly had targetable trees in an older release, but had to disable them again after they introduced several regressions.  Map authors also use trees as part of the static map design, either hiding flaws in the underlying tiles or to create choke points.   Allowing players to destroy and remove them will break that use case.

deniz1a:
But they have nice animations that are not used... Shouldn't maps use other elements to block units, like cliffs?

deniz1a:
Map authors could be asked to update their maps...

reaperrr:
Updating dozens of maps with potential negative gameplay impact just for some nice animations doesn't sound worth the hassle, in my opinion.

deniz1a:
Ok, so what if updating maps was a prerequisite for this? This could be in next+1 milestone and in the meantime we could gather all the maps that need to be updated and ask the authors to change their maps until next+1?

Because this feature should eventually be included, right? After maps are made compatible with it. For example recently, the gem mines were changed and maps needed to be updated. Couldn't this be done like that?

abcdefg30:
To be fair, I think this is a good idea for custom maps and thus created [this wiki page](https://github.com/OpenRA/OpenRA/wiki/How-to-add-destructible-trees-to-your-map).

PR #9103, opened 2015-08-18
Refactored WaterPaletteRotation into RotationPaletteEffect

### LINKED ISSUE #8872: Water color cycle broken for RA ships on desert maps
The desert tileset modifies different palette indexes, so it doesn't animate on ships (and possibly sub pen/ship yard) on desert tileset.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25254220-water-color-cycle-broken-for-ra-ships-on-desert-maps?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Moves rotation base index from tileset to palette effect.
Unhardcodes some other aspects.

Fixes #8872.

### DIFF
--- a/OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
+++ b/OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
@@ -24,13 +24,13 @@ namespace OpenRA.Mods.Common.Traits
 
 		[Desc("Defines for which tileset IDs this effect should be loaded.",
 			"If none specified, it applies to all tileset IDs not explicitly excluded.")]
-		public readonly string[] Tilesets = null;
+		public readonly HashSet<string> Tilesets = new HashSet<string>();
 
 		[Desc("Defines which palettes should be excluded from this effect.")]
 		public readonly HashSet<string> ExcludePalettes = new HashSet<string>();
 
 		[Desc("Don't apply the effect for these tileset IDs.")]
-		public readonly string[] ExcludeTilesets = null;
+		public readonly HashSet<string> ExcludeTilesets = new HashSet<string>();
 
 		[Desc("Palette index of first RotationRange color.")]
 		public readonly int RotationBase = 0x60;
@@ -47,25 +47,41 @@ namespace OpenRA.Mods.Common.Traits
 	class RotationPaletteEffect : ITick, IPaletteModifier
 	{
 		readonly RotationPaletteEffectInfo info;
-		readonly World world;
 		readonly uint[] rotationBuffer;
+		readonly bool validTileset;
+		readonly string tilesetId;
 		float t = 0;
 
 		public RotationPaletteEffect(World world, RotationPaletteEffectInfo info)
 		{
-			this.world = world;
 			this.info = info;
-
 			rotationBuffer = new uint[info.RotationRange];
+			tilesetId = world.TileSet.Id;
+
+			validTileset = IsValidTileset();
+		}
+
+		bool IsValidTileset()
+		{
+			if (info.Tilesets.Count == 0 && info.ExcludeTilesets.Count == 0)
+				return true;
+
+			return info.Tilesets.Contains(tilesetId) && !info.ExcludeTilesets.Contains(tilesetId);
 		}
 
 		public void Tick(Actor self)
 		{
+			if (!validTileset)
+				return;
+
 			t += info.RotationStep;
 		}
 
 		public void AdjustPalette(IReadOnlyDictionary<string, MutablePalette> palettes)
 		{
+			if (!validTileset)
+				return;
+
 			var rotate = (int)t % info.RotationRange;
 			if (rotate == 0)
 				return;
@@ -73,9 +89,7 @@ namespace OpenRA.Mods.Common.Traits
 			foreach (var kvp in palettes)
 			{
 				if ((info.Palettes.Count > 0 && !info.Palettes.Any(kvp.Key.StartsWith))
-					|| (info.Tilesets != null && !info.Tilesets.Contains(world.TileSet.Id))
-					|| (info.ExcludePalettes.Count > 0 && info.ExcludePalettes.Any(kvp.Key.StartsWith))
-					|| (info.ExcludeTilesets != null && info.ExcludeTilesets.Contains(world.TileSet.Id)))
+					|| (info.ExcludePalettes.Count > 0 && info.ExcludePalettes.Any(kvp.Key.StartsWith)))
 					continue;
 
 				var palette = kvp.Value;
--- a/OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
+++ b/OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
@@ -48,16 +48,15 @@ namespace OpenRA.Mods.Common.Traits
 	{
 		readonly RotationPaletteEffectInfo info;
 		readonly World world;
+		readonly uint[] rotationBuffer;
 		float t = 0;
-		uint[] temp;
 
 		public RotationPaletteEffect(World world, RotationPaletteEffectInfo info)
 		{
 			this.world = world;
 			this.info = info;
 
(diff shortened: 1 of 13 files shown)


pchote:
Nice.  This has been on my extended todo list for ages.

OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
```diff
@@ -0,0 +1,80 @@
...
+		readonly World world;
+		float t = 0;
+		uint[] temp;
+
+		public RotationPaletteEffect(World world, RotationPaletteEffectInfo info)
+		{
+			this.world = world;
+			this.info = info;
+
+			temp = new uint[info.RotationRange]; /* allocating this on the fly actually hurts our profile */
```
Phrohdoh: Comment above this line and with `//` please.

abcdefg30: Also start the comment with an uppercase char, please.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -1760,6 +1760,13 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 					}
 				}
 
+				// WaterPaletteRotation renamed to RotationPaletteEffect
+				if (engineVersion < 20150818)
+				{
+					if (depth == 1 && parentKey == "^Palettes" && node.Key == "WaterPaletteRotation")
```
abcdefg30: Why the `parentKey == "^Palettes"`?

reaperrr:
Updated.
Also added 'water' to @keys to make it more clear that the effect is for water in RA and TD (as the trait name itself is now generic).

Mailaender:
Nice cleanup and fix. :+1:

GraionDilach:
TBH, I'd be happier if this would have a positive logic instead of the negative logic it has right now in the long run, but I guess I'm a minority here.

reaperrr:
I've tried to add a positive logic, but somehow it doesn't work as it should (likely something simple and silly). If someone helps me debugging and fixing it, I'll include it, otherwise I'd rather leave that for later when RA2 or some third-party mod needs it.

Mailaender:
You probably also have to set the Exclude* arrays to null and check against that to make it work. As all palette effects currently use negative filters this is just fine and consistent for the moment.

reaperrr:
Updated, now supports both excluding and explicitly loading palettes/tilesets.

Mailaender:
Needs a rebase.

reaperrr:
Rebased.

reaperrr:
Rebased.

Unit158:
Simple enough :+0.5: I have yet to test it.

OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
```diff
@@ -0,0 +1,92 @@
...
+		readonly World world;
+		float t = 0;
+		uint[] temp;
+
+		public RotationPaletteEffect(World world, RotationPaletteEffectInfo info)
+		{
+			this.world = world;
+			this.info = info;
+
+			// Allocating this on the fly actually hurts our profile
```
RoosterDragon: This comment can go.

OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
```diff
@@ -0,0 +1,92 @@
...
+
+		public object Create(ActorInitializer init) { return new RotationPaletteEffect(init.World, this); }
+	}
+
+	class RotationPaletteEffect : ITick, IPaletteModifier
+	{
+		readonly RotationPaletteEffectInfo info;
+		readonly World world;
+		float t = 0;
+		uint[] temp;
```
RoosterDragon: ```readonly``` and maybe rename to ```rotationBuffer``` or something less silly.

OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
```diff
@@ -0,0 +1,92 @@
...
+	[Desc("Palette effect used for sprinkle \"animations\".")]
+	class RotationPaletteEffectInfo : ITraitInfo
+	{
+		[Desc("Defines to which palettes this effect should be applied to.",
+			"If none specified, it applies to all palettes not explicitly excluded.")]
+		public readonly string[] Palettes = null;
+
+		[Desc("Defines for which tileset IDs this effect should be loaded.",
+			"If none specified, it applies to all tileset IDs not explicitly excluded.")]
+		public readonly string[] Tilesets = null;
```
RoosterDragon: This can now be a ```HashSet<string>```.

OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
```diff
@@ -0,0 +1,92 @@
...
+
+		[Desc("Defines which palettes should be excluded from this effect.")]
+		public readonly string[] ExcludePalettes = null;
+
+		[Desc("Don't apply the effect for these tileset IDs.")]
+		public readonly string[] ExcludeTilesets = null;
```
RoosterDragon: ```HashSet<string>```.

RoosterDragon:
:+1: after addressing the minor refactorings.

reaperrr:
Updated and rebased.

OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
```diff
@@ -0,0 +1,106 @@
...
+		{
+			this.info = info;
+
+			rotationBuffer = new uint[info.RotationRange];
+
+			tilesetId = world.TileSet.Id;
+			validTileset = IsValidTileset();
+		}
+
+		public bool IsValidTileset()
```
RoosterDragon: This method you've added - this doesn't seem to need to be public?

OpenRA.Mods.Common/Traits/PaletteEffects/RotationPaletteEffect.cs
```diff
@@ -0,0 +1,106 @@
...
+
+		public object Create(ActorInitializer init) { return new RotationPaletteEffect(init.World, this); }
+	}
+
+	class RotationPaletteEffect : ITick, IPaletteModifier
+	{
+		readonly RotationPaletteEffectInfo info;
+		readonly uint[] rotationBuffer;
+		float t = 0;
+		bool validTileset = true;
```
RoosterDragon: This could be ```readonly``` and doesn't need initializing to ```true``` since the constructor will set the value.

reaperrr:
Updated.

RoosterDragon:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5a179fa592d7d1ba382ec291bb28685aa9c87f10...b9f95a4a354072e39f47ed01faac0a3ea2182951)

mods/ra/rules/palettes.yaml
```diff
@@ -1,7 +1,8 @@
 ^Palettes:
-	PlayerPaletteFromCurrentTileset:
+	PaletteFromFile@player:
 		Name: player
-		ShadowIndex: 3,4
```
Mailaender: I think this change is responsible for https://github.com/OpenRA/OpenRA/issues/9989. Can you elaborate why this was necessary?

reaperrr: I just didn't know this served a purpose, since only index 4 is used for shadows in the original.

PR #9105, opened 2015-08-18
Mobile Blocking Refactor

RoosterDragon:
This is a refactoring in Mobile.cs of the logic used to determine if actors block movement.

This centralises some duplicated logic in `CanMoveFreelyInto` and into `GetAvailableSubCell` into a new method and makes some minor changes to squeeze out a bit more performance as the pathfinder calls `CanMoveFreelyInto` frequently.

This is intended to have no impact on the output of these functions. This means for testing you could record a game on bleed and replaying it with this PR should not desync. That's how I tested it.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Mobile.cs
+++ b/OpenRA.Mods.Common/Traits/Mobile.cs
@@ -29,6 +29,14 @@ namespace OpenRA.Mods.Common.Traits
 		All = TransientActors | BlockedByMovers
 	}
 
+	public static class CellConditionsExts
+	{
+		public static bool HasCellCondition(this CellConditions c, CellConditions cellCondition)
+		{
+			return (c & cellCondition) == cellCondition;
+		}
+	}
+
 	[Desc("Unit is able to move.")]
 	public class MobileInfo : IMoveInfo, IOccupySpaceInfo, IFacingInfo, UsesInit<FacingInit>, UsesInit<LocationInit>, UsesInit<SubCellInit>
 	{
@@ -189,9 +197,9 @@ namespace OpenRA.Mods.Common.Traits
 			if (SharesCell && world.ActorMap.HasFreeSubCell(cell))
 				return true;
 
-			if (check.HasFlag(CellConditions.TransientActors))
+			if (check.HasCellCondition(CellConditions.TransientActors))
 			{
-				var canIgnoreMovingAllies = self != null && !check.HasFlag(CellConditions.BlockedByMovers);
+				var canIgnoreMovingAllies = self != null && !check.HasCellCondition(CellConditions.BlockedByMovers);
 				var needsCellExclusively = self == null || Crushes == null || !Crushes.Any();
 				foreach (var a in world.ActorMap.GetUnitsAt(cell))
 				{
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -186,36 +194,48 @@ public bool CanEnterCell(World world, Actor self, CPos cell, Actor ignoreActor =
...
+
+			// If the check allows: we are not blocked by allied units moving in our direction.
+			if (!check.HasCellCondition(CellConditions.BlockedByMovers) &&
+				self != null &&
+				self.Owner.Stances[otherActor.Owner] == Stance.Ally &&
+				IsMovingInMyDirection(self, otherActor))
+				return false;
+
+			// If we cannot crush the other actor in our way, we are blocked.
+			if (self == null || Crushes == null || Crushes.Length == 0)
```
WolfGaming: The thing that irks me about this line is that we are checking this multiple times in CanMoveFreelyInto even though it doesn't ever change... I guess it is negligible but it still irks me.

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -186,36 +194,48 @@ public bool CanEnterCell(World world, Actor self, CPos cell, Actor ignoreActor =
...
+				return true;
+
+			// If the other actor in our way cannot be crushed, we are blocked.
+			var crushables = otherActor.TraitsImplementing<ICrushable>();
+			if (!crushables.Any())
+				return true;
```
WolfGaming: If I remember correctly our styling says we should add an empty line after a single if statement?

OpenRA.Mods.Common/Traits/Mobile.cs
RoosterDragon: Whilst it's a touch irksome, it's certainly not a performance issue compared to grovelling through the various dictionary lookups.

OpenRA.Mods.Common/Traits/Mobile.cs
RoosterDragon: I prefer to group it since it's all related to crushability and the comment applies to the whole chunk of code, but I'll add a newline if that's the preference (we should try and find a StyleCop rule that does that, if it is indeed supposed to be the convention).

Mailaender: It is not convention and we also tend to group that in various other location as it helps with readability vs. following a blind preference. Let's try to avoid commenting on minor style issues.

Phrohdoh: A minor style issue is still a style issue and we ought to enforce consistency.

Mailaender:
Nudging and crushing seems to work as it did before. :white_check_mark:

chrisforbes:
:+1: good to go.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f80567f63accba61f29b991376abe393a5b2c3e7...c5fe2c8f7597976da42bddc601fe72033637e061)

PR #9106, opened 2015-08-18
D2k vanilla balance pass

OmegaBolt:
-- All costs, prerequisites, CustomBuildTimeValue, health, armour, weapon damage, range, warhead values and sight ranges should be D2k accurate
-- Made aircraft and sandworms visible through fog because it is quite oppressive (especially for carryalls)
-- Removed paradrop power, bounty, veterancy
-- Unit names were partially copied from D2k until they got a bit silly, such as upgrade.high_tech_factory
-- Unit speeds were worked out using "unit speed- (([speed value] / 32) / 10) x 1024 / 25", then multiplied by 2 to be a playable speed
-- ROT is a guess, but should be respectful of unit's ROT relative to each other
-- Warhead spreads are as accurate as I could get them through trial and error
-- Gave husks an explosion on death
-- Buildings are no longer automatically targeted (except turrets)
-- Buildings no longer emit Troopers on sale
-- Worms no longer target infantry, spawn less frequently and always disappear after an attack
-- Fixed a couple of cliff tiles that were set to Sand instead of Cliff
-- All Starport units can now be gotten with no other prerequisites
-- The Upgrades queue is now independent of the Con Yard
    - Gave the Building production trait to the player actor as well, so that it will be the first tab to open after deploying a Con Yard. Added construction_yard as prerequisite to necessary buildings. The side effect is that if you use the debug menu to enable the construction of everything you will be able to build and place structures without a Con Yard.

### DIFF
--- a/mods/d2k/rules/aircraft.yaml
+++ b/mods/d2k/rules/aircraft.yaml
@@ -1,21 +1,23 @@
 carryall.reinforce:
 	Inherits: ^Plane
 	Valued:
-		Cost: 1200
+		Cost: 1100
+	CustomBuildTimeValue:
+		Value: 648
 	Tooltip:
 		Name: Carryall
 		Description: Large winged, planet-bound ship\n  Automatically lifts harvesters.
 	Health:
-		HP: 250
+		HP: 4800
 	Armor:
-		Type: Light
+		Type: light
 	Helicopter:
 		CruiseAltitude: 2100
 		InitialFacing: 0
-		ROT: 4
-		Speed: 160
+		ROT: 7
+		Speed: 262	##131
 		LandableTerrainTypes: Sand, Rock, Transition, Spice, Dune
-		RepairBuildings: repair
+		RepairBuildings: repair_pad
 		RearmBuildings:
 		Repulsable: False
 		LandAltitude: 100
@@ -26,6 +28,10 @@ carryall.reinforce:
 		Automatic: False
 	RenderSprites:
 		Image: carryall
+	SelfHealing:
+		Step: 5
+		Ticks: 3
+		HealIfBelow: 50%
 
 carryall:
 	Inherits: carryall.reinforce
@@ -35,35 +41,6 @@ carryall:
 		Queue: Aircraft
 		BuildPaletteOrder: 120
 
-carryall.infantry:
-	Inherits: ^Plane
-	ParaDrop:
-		DropRange: 5c0
-		ChuteSound:
-	Health:
-		HP: 200
-	Armor:
-		Type: Light
-	RevealsShroud:
-		Range: 12c0
-		Type: CenterPosition
-	Plane:
-		ROT: 4
-		Speed: 280
-		RepairBuildings: repair
-		RearmBuildings:
-		Repulsable: False
-	Cargo:
-		MaxWeight: 5
-		Types: Infantry
-	Tooltip:
-		Name: Carryall
-	SpawnActorOnDeath:
-		Actor: carryall.infantry.husk
-	RejectsOrders:
-	RenderSprites:
-		Image: carryall
-
 frigate:
 	Inherits: ^Plane
 	ParaDrop:
@@ -72,16 +49,17 @@ frigate:
 		Name: Frigate
 		Description: Supply spacecraft
 	Plane:
-		ROT: 5
-		Speed: 350
-		RepairBuildings: repair
+		CruiseAltitude: 1800
+		ROT: 7
+		Speed: 296	##148
+		RepairBuildings: repair_pad
 		RearmBuildings:
 		Repulsable: False
 	Health:
 		HP: 500
 	-AppearsOnRadar:
 	Armor:
-		Type: Heavy
+		Type: heavy
 	Cargo:
 		MaxWeight: 20
 		PipCount: 10
@@ -90,7 +68,7 @@ frigate:
 	FlyAwayOnIdle:
 	RejectsOrders:
 
-orni:
+ornithopter:
 	Inherits: ^Plane
 	Valued:
 		Cost: 1000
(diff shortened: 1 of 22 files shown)


mods/d2k/weapons.yaml
```diff
@@ -9,74 +9,95 @@ LMG:
 		ContrailDelay: 0
 		ContrailUsePlayerColor: true
 	Warhead@1Dam: SpreadDamage
-		Spread: 96
-		Damage: 15
+		Spread: 48
+		Falloff: 100, 100, 65, 30, 0, 0, 0
```
reaperrr: You don't need to define 0 multiple times, the last two can simply be removed (applies to all other weapons, too).

mods/d2k/rules/defaults.yaml
```diff
@@ -131,13 +123,16 @@
...
-		AllowAllies: yes
-		CaptureThreshold: 1.0
-	TransformOnCapture:
-		ForceHealthPercentage: 25
+#	Capturable:
+#		Type: husk
+#		AllowAllies: yes
+#		CaptureThreshold: 1.0
+#	TransformOnCapture:
+#		ForceHealthPercentage: 25
```
reaperrr: Our policy is to not leave commented out lines in, but rather delete them, especially in cases like this (unless someone explicitly said otherwise on IRC).

OmegaBolt: I know that but I wasn't sure if people really want certain features removed. I do, but to save more work if there are complaints I left it in. If people are willing to go along with it then the comments can all be removed.

mods/d2k/weapons.yaml
OmegaBolt: I see, all examples of Falloff in the yamls seemed to have the same amount of numbers so I stuck to that.

mods/d2k/rules/defaults.yaml
pchote: We probably want to keep this in to avoid confusion (since this is available in the other mods).

Mailaender:
Needs a rebase. See https://github.com/OpenRA/OpenRA/wiki/Contributing#keeping-up-to-date-rebasing on what to do.

MicroJOo:
i have tried it , and i think its worth a playtest ... but we should keep in mind ... that stuff like the "speed" of the units ... and "range" ... is based on environment with no "Fog of War" ...

penev92:
We could have the lobby settings default to Fog: Off for the time being, and see what the feedback is.

mods/d2k/rules/aircraft.yaml
```diff
@@ -72,16 +49,17 @@ frigate:
 		Name: Frigate
 		Description: Supply spacecraft
 	Plane:
-		ROT: 5
-		Speed: 350
-		RepairBuildings: repair
+		CruiseAltitude: 1800
+		ROT: 7
+		Speed: 296	##148
```
Unit158: What's with the commented speeds?

OmegaBolt: The "original" converted values. The ones used are multiplied to be a playable speed. I kept the comment so it's easier to go back to the original if people disagree with the amount it was increased.

Unit158: I'm going to disagree right of the bat, in that case.

OmegaBolt: Disagree with what? You don't want the mod at middle D2k game speed, it's ludicrously slow and no one (especially in multiplayer) plays it at that speed.

Unit158: I want it at the default d2k speed, and this is faster than that :)

MicroJOo: its near impossible to destroy trikes as long as they are moving ... which makes scouting really easy , im just saying ... you could probably explore all  maps within 1 minute ...

penev92: Actually this is still a lot slower than the original at max speed, which is how I hear people play it. But I do agree that trikes are unkillable while moving.
I tried the commented-out values though, and it is indeed ridiculously slow.

penev92:
I'm not convinced about removing veterancy by the way.
It's a cute little system with no real application in a meat grinder mod like this. That said, it won't impact gameplay much, but can still heavily impact player experience.

OmegaBolt:
I don't think veterancy should be added for the sake of it. It could be kept as purely a visual queue like "oh this unit has survived a lot of combat" but adding veterancy benefits does effect gameplay and it should be done properly, with proper balancing, testing and thought IF done at all.

Since, as you said, Dune is pretty spammy and quick and all weapons have very effective damage spread so units die rapidly, I think its better not to have it. Having a feature for the sake of it is just messy. I think a cleaner, simpler, purer game is always better (not just to stick to vanilla).

mods/d2k/rules/aircraft.yaml
OmegaBolt: Well I really don't want default speed, it's not fun. There could always be a compromise of something in the middle, 1.5x or 1.75x the speed but I'm definitely not doing default.

MicroJOo:
i agree with @OmegaBolt about the veterancy

reaperrr:
In any case, I think this will be a nightmare to review and rebase as a single PR, so this should be done bit by bit spanning over multiple smaller PRs.

GraionDilach:
The main issue for this PR is that I don't think it's easy to split. Maybe to commits alone, but decoupling it to separate PRs sounds tedious to me, considering that it does not touch the codebase per-se.

Maybe it should be merged as-is (ofc after rebase and a compromise at speeds reached) at the start of a release cycle and then dealing with the regressions one-by-one later.

penev92:
Right now what @GraionDilach is suggesting seems like the only way to get this moving again, unless @OmegaBolt or someone else want to do some git magic and split this into several commits (then those can be PR-ed in chunks if needed).

MicroJOo:
theres currently nothing that could stop a moving trike ... no weapon can hit em
and with this speed , you have almost every map scouted completly in less then 20 seconds

having d2k on default speed isnt ridiculously slow ... it would be close to our RedAlert (our most popular mod)

GraionDilach:
One of the constant criticism towards the Red Alert mod that it __is__ slow. A gamespeed change sounds convincing both as an additional point to further depart from the "RA with sand" POV and also as a gesture towards "the game is slooow why you went with this slooow" haters.

That being said, the 2x modifier really seems too fast. I am opting for 1.5x.

MicroJOo:
lol , "constant criticism" towards RedAlert ... by whom ? cncnet ? just increasing the speed of a mod doesnt mean it adds a additional point to depart from "RA with sand" ... its kind of weird having mods all with their own different gamespeed ... that said ... switching mod and playing a game of RedAlert after playing a D2k-game must feel like youre playing in slow-motion .. 

"Game-Speed"  shouldnt be a factor to distinguish between our mods ... and RedAlert our "slowest mod" is our best horse , theres daily 30-40+ people playing it ... so the lack of speed  doesnt bother players 

on the other hand , what has been a constant criticism towards "OpenRA" .. is the lack of a "GameSpeed-slider" to control the gamespeed for specific game

but hey , if anyone thinks this will help luring "especially" new people on D2k , giving it a try ... 
then i shall not give a shit

Mailaender:
We have dynamic game speed now. https://github.com/OpenRA/OpenRA/pull/9255 

This needs a rebase since https://github.com/OpenRA/OpenRA/issues/9278 was integrated.

reaperrr:
I'm closing this since most of this has already been merged or at least filed in separate PRs, often with several corrections, and any remaining bits can be extracted easily.
Thanks for the work, this helped a lot!

PR #9108, opened 2015-08-19
Cactus improvement

deniz1a:
Solves the highly debated cactus tooltip issue in Red Alert desert maps. They are finally identified as what they are: cacti. :cactus:

### DIFF
--- a/mods/ra/rules/decoration.yaml
+++ b/mods/ra/rules/decoration.yaml
@@ -24,6 +24,8 @@ T03:
 
 T04:
 	Inherits: ^Tree
+	Tooltip:
+		Name: Cactus
 	Building:
 		Footprint: __ x_
 		Dimensions: 2,2


deniz1a:
Changelog entry suggestion: Improved cacti.

Unit158:
:+1:

reaperrr:
This will make T04 appear as "Cactus" on temperate/snow terrain, which is wrong because it's a small fir or spruce there, so :-1: unless you find a solution for that.

deniz1a:
Oh T04 represents multiple actors? I don't know then.

pchote:
Best to close this, then.

Unit158:
What is going on here? This behaviour isn't obvious when looking at the YAML.

GraionDilach:
@Unit158 Tree IDs are reused between theaters but does not mean the same artwork in all theaters. In this case `T04` is a cactus in `DESERT` but a regular tree in the other theaters - a proper fix would involve tooltips per theater.

PR #9115, opened 2015-08-20
Dehardcode the scale-inaccuracy option.

GraionDilach:
In current bleed, inaccuracy behaves odd, it's hardcoded to scale with the range at bullets and not on missiles.

I can understand this is benefical in the current mods, but besides not being consistent with itself, there are also cases where the softcoding is preferred.
- In TS and RA2, Westwood inaccuracy never scaled. Although I can't remember if TS used this, the RA2 Flak weaponry was notorious of having around 0c300 inaccuracy even at point-blank range while being a Bullet weapon.
- Modders might try to use high-burst inaccurate bullet-based weapons, which end up impossible to balance - perfect example is @DoGyAUT's Wolverine at http://media.moddb.com/images/mods/1/15/14166/wolverine.gif - resulting either being overpowered in close-range either underpowered at max range.

This PR adds a `ScaleInaccurancyWithRange` option to bullets and missiles, with a default value of `true`, while also brings an upgrade rule for missiles to not change their behaviour.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Bullet.cs
+++ b/OpenRA.Mods.Common/Effects/Bullet.cs
@@ -25,8 +25,9 @@ namespace OpenRA.Mods.Common.Effects
 		[Desc("Projectile speed in WDist / tick, two values indicate variable velocity.")]
 		public readonly WDist[] Speed = { new WDist(17) };
 
-		[Desc("Maximum offset at the maximum range.")]
-		public readonly WDist Inaccuracy = WDist.Zero;
+		[Desc("Maximum inaccuracy offset. Can have two values.",
+			"First value determinates a fixed limit, second value scales along with weapon's range.")]
+		public readonly WDist[] Inaccuracy = { WDist.Zero };
 
 		[Desc("Image to display.")]
 		public readonly string Image = null;
@@ -105,10 +106,13 @@ namespace OpenRA.Mods.Common.Effects
 				speed = info.Speed[0];
 
 			target = args.PassiveTarget;
-			if (info.Inaccuracy.Length > 0)
+
+			if (info.Inaccuracy.Length > 1 || info.Inaccuracy[0].Length > 0)
 			{
-				var inaccuracy = OpenRA.Traits.Util.ApplyPercentageModifiers(info.Inaccuracy.Length, args.InaccuracyModifiers);
-				var maxOffset = inaccuracy * (target - pos).Length / args.Weapon.Range.Length;
+				var inaccuracy = info.Inaccuracy.Length > 1
+					? info.Inaccuracy[1].Length * (target - pos).Length / args.Weapon.Range.Length + info.Inaccuracy[0].Length
+					: info.Inaccuracy[0].Length;
+				var maxOffset = OpenRA.Traits.Util.ApplyPercentageModifiers(inaccuracy, args.InaccuracyModifiers);
 				target += WVec.FromPDF(world.SharedRandom, 2) * maxOffset / 1024;
 			}
 
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -2285,6 +2285,17 @@ namespace OpenRA.Mods.Common.UtilityCommands
 					}
 				}
 
+				// update inaccuracy modes
+				if (engineVersion < 20150902)
+				{
+					if (node.Key.StartsWith("Projectile") && node.Value.Value == "Bullet")
+					{
+						var inaccuracynode = node.Value.Nodes.FirstOrDefault(x => x.Key == "Inaccuracy");
+						if (inaccuracynode != null)
+							inaccuracynode.Value.Value = "0, " + inaccuracynode.Value.Value;
+					}
+				}
+
 				UpgradeWeaponRules(engineVersion, ref node.Value.Nodes, node, depth + 1);
 			}
 		}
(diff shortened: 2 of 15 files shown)


pchote:
It would make more sense to default it to false, since you're just disabling it for our default mods anyway.  If you're worried about backwards compatibility for external mods you can have the upgrade rule define it as true, and then remove the line from the default mods in the second commit.

GraionDilach:
The upgrade rule would affect the default mods if scaling behaviour is disabled by default, because then we have to upgrade inaccurate bullets instead, and that would be a bigger diff: https://github.com/OpenRA/OpenRA/compare/bleed...GraionDilach:dehardcode-inaccuracy-scaling-false.

pchote:
Ok, fair enough.

UberWaffe:
Would it make sense to perhaps create a separate future issue to change inaccuracy into a min and max? Or is there no real benefit?

Ex. Weapon Range: 0 -> 1000
Inaccuracy.Max: 200
Inaccuracy.Min: 100
Would mean that the inaccuracy starts at 100 and scales to 200 at the max range of 1000. (So 110 at 100, 120 at 200, ... 150 at 500, ..., etc.)

Having Inaccuracy.Min == Inaccuracy.Max would be the same as ScaleInaccuracyWithRange = false.

GraionDilach:
I'm neutral on that one, but if someone says that should be done, I'm okay with repurposing this PR into that.

OpenRA.Mods.Common/Effects/Bullet.cs
```diff
@@ -28,6 +28,9 @@ public class BulletInfo : IProjectileInfo
 		[Desc("Maximum offset at the maximum range.")]
```
Mailaender: The inline documentation needs to be adapted as well.

GraionDilach: I'd wait with that before someone decides if @UberWaffe's suggestion should be incorporated into this one or not.

UberWaffe:
@GraionDilach If you are willing to go through the effort now, I would vote for yes.

penev92:
On one hand it would make things even more configurable (flexible). On the other hand modders will have to do some math if they want to keep the current behaviour (not sure anyone would do it though).
So if I understand the suggestion right, I think it's a good one.

GraionDilach:
In @UberWaffe's concept, all bleed inaccurate bullets would have 0 `MinAccuracy` and all inaccurate missiles would have `MaxAccuracy` `MinAccuracy` with `Accuracy` getting renamed to `MaxAccuracy`.

It sounds as a really simple modder-friendly system imo.

GraionDilach:
Tried doing @UberWaffe's method at https://github.com/OpenRA/OpenRA/compare/bleed...GraionDilach:rescale-inaccuracy but it made everything more accurate around the max ranges and I can't tell why. >_>

GraionDilach:
Okay, thrown out the boolean, and properly merged the two behaviours now, with `Inaccuracy` having two fields, first being the constant (ex-missile) and second being the scaled (ex-bullet) behaviour.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2249,6 +2249,17 @@ internal static void UpgradeWeaponRules(int engineVersion, ref List<MiniYamlNode
 					}
 				}
 
+				// update inaccuracy modes
+				if (engineVersion < 20150828)
```
Unit158: Invalid indentation

GraionDilach: A line below. Fixed.

Mailaender:
Less realism for easier legacy compatible balancing makes sense. :+1: I also can't spot regressions. :white_check_mark: This might need an upgrade rule now to retain behavior for everyone including thirdparty mods.

GraionDilach:
The upgrade rule is there.

GraionDilach:
Rebased.

GraionDilach:
Rebased again because Travis didn't run in the last instance. :eyes:

OpenRA.Mods.Common/Effects/Bullet.cs
```diff
@@ -25,8 +25,9 @@ public class BulletInfo : IProjectileInfo
 		[Desc("Projectile speed in WDist / tick, two values indicate variable velocity.")]
 		public readonly WDist[] Speed = { new WDist(17) };
 
-		[Desc("Maximum offset at the maximum range.")]
-		public readonly WDist Inaccuracy = WDist.Zero;
+		[Desc("Maximum inaccuracy offset. Can have two values.",
+			"First value determinates a fixed limit, second value scales along with weapon's range.")]
+		public readonly WDist[] Inaccuracy = { WDist.Zero };
```
obrakmann: I feel the description needs a bit more elaboration. Could you mention that the second value is optional, and that the first value is evaluated as well when the second is given?

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2285,6 +2285,17 @@ internal static void UpgradeWeaponRules(int engineVersion, ref List<MiniYamlNode
 					}
 				}
 
+				// update inaccuracy modes
+				if (engineVersion < 20150902)
```
obrakmann: This needs updating. Probably also the reason this needs a rebase.

obrakmann:
Looks otherwise fine to me. :+1:

GraionDilach:
I'm currently waiting for the missile PR being finished, notsure how they'd conflict - that and i again forgot what was the main reason this got the Chnages requested tag to begin with - explicitly crashing if Inaccuracy != 2maybe IIRC.

abcdefg30:

PR #9117, opened 2015-08-20
Prevent infiltrating allied targets.

### LINKED ISSUE #9116: Prevent spies from infiltrating ally buildings
Seems to defeat the purpose/reward of using of a spy

> Phoenix___: Hello all! I've been looking at the change logs and nothing seems to have been done about spies infiltrating their ally's buildings; wont this make allies a bit OP?
> Phoenix___: spies being able to*
> hamb: i don't play RA, but that sounds a bit broken
> hamb: also unintentional
> Phoenix___: Yeah; allows you to get sonar pulse pretty easily and because british spies are $250 and the minimum you can get from a refinery is $500 you can get infinite money if you and your teammate work together
> Phoenix___: plus in the new update it will allow you get veteran units very quickly.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25990572-prevent-spies-from-infiltrating-ally-buildings?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

GraionDilach:
Fixes #9116.

### DIFF
--- a/OpenRA.Mods.RA/Activities/Infiltrate.cs
+++ b/OpenRA.Mods.RA/Activities/Infiltrate.cs
@@ -10,6 +10,7 @@
 
 using OpenRA.Mods.Common.Activities;
 using OpenRA.Mods.Common.Traits;
+using OpenRA.Mods.RA.Traits;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.RA.Activities
@@ -20,17 +21,25 @@ namespace OpenRA.Mods.RA.Activities
 
 		readonly Cloak cloak;
 
+		readonly Infiltrates infiltrates;
+
 		public Infiltrate(Actor self, Actor target)
 			: base(self, target)
 		{
 			this.target = target;
 
 			cloak = self.TraitOrDefault<Cloak>();
+
+			infiltrates = self.TraitOrDefault<Infiltrates>();
 		}
 
 		protected override void OnInside(Actor self)
 		{
-			if (target.IsDead || target.Owner == self.Owner)
+			if (target.IsDead)
+				return;
+
+			var stance = self.Owner.Stances[target.Owner];
+			if (!infiltrates.Info.ValidStances.HasStance(stance))
 				return;
 
 			if (cloak != null && cloak.Info.UncloakOnInfiltrate)
--- a/OpenRA.Mods.RA/Traits/Infiltration/Infiltrates.cs
+++ b/OpenRA.Mods.RA/Traits/Infiltration/Infiltrates.cs
@@ -24,21 +24,24 @@ namespace OpenRA.Mods.RA.Traits
 
 		[VoiceReference] public readonly string Voice = "Action";
 
+		[Desc("What diplomatic stances can be infiltrated by this actor.")]
+		public readonly Stance ValidStances = Stance.Neutral | Stance.Enemy;
+
 		public object Create(ActorInitializer init) { return new Infiltrates(this); }
 	}
 
 	class Infiltrates : IIssueOrder, IResolveOrder, IOrderVoice
 	{
-		readonly InfiltratesInfo info;
+		public readonly InfiltratesInfo Info;
 
 		public Infiltrates(InfiltratesInfo info)
 		{
-			this.info = info;
+			Info = info;
 		}
 
 		public IEnumerable<IOrderTargeter> Orders
 		{
-			get { yield return new TargetTypeOrderTargeter(info.Types, "Infiltrate", 7, "enter", true, false); }
+			get { yield return new InfiltrationOrderTargeter(Info); }
 		}
 
 		public Order IssueOrder(Actor self, IOrderTargeter order, Target target, bool queued)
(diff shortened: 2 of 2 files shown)


OpenRA.Mods.RA/Traits/Infiltration/Infiltrates.cs
```diff
@@ -58,6 +58,10 @@ bool IsValidOrder(Actor self, Order order)
 			if (order.ExtraData == 0 && order.TargetActor == null)
 				return false;
 
+			// Should not infiltrate own or allied
+			if (order.TargetActor.Owner.IsAlliedWith(self.Owner))
```
abcdefg30: Could we get a field in the info class for this?

GraionDilach: Not certain what you mean - that sounds pointless/abusable to me. Issuing orders before breaking a fragile alliance for infiltration would be the sole case I'd see that field being used and I consider that an abuse.

Mailaender: I guess he means a bool in the trait info which I don't think we need as setting it allow this only enables exploits.

Phrohdoh: :+1: for the bool.
It isn't an exploit if the mod author wants to allow it.

penev92: I think he means that if you have a field in the Info (set to not allow infiltrating allies by default, and none of our mods changing the default, ofc) some thirdparty mod can change it if they like, or custom maps can still enable it for their crazy gameplay.

Mailaender:
I can still force fire on an ally although now the order is not resolved. As Infiltrates already disallows targetAllyUnits via TargetTypeOrderTargeter the bug probably lies in UnitOrderTargeter instead.

OpenRA.Mods.RA/Traits/Infiltration/Infiltrates.cs
Phrohdoh: That is still not an exploit then.
I see no reason not to allow it, as long as we disable it where we want to.

OpenRA.Mods.RA/Activities/Infiltrate.cs
```diff
@@ -30,7 +30,7 @@ public Infiltrate(Actor self, Actor target)
 
 		protected override void OnInside(Actor self)
 		{
-			if (target.IsDead || target.Owner == self.Owner)
+			if (target.IsDead || target.Owner.IsAlliedWith(self.Owner))
```
Phrohdoh: This should be removed entirely IMO as the Order ought to take care of proper targeting.

OpenRA.Mods.RA/Traits/Infiltration/Infiltrates.cs
GraionDilach: This was reported as a bug. This was considered a bug even in RA2. I still don't see the gain of putting this under a boolean of say, `AllowInfiltrateAllies`.

OpenRA.Mods.RA/Activities/Infiltrate.cs
penev92: The owner can change after issuing the order though.

GraionDilach: No, it shouldn't. Fragile alliances. Allying inbetween the order and the spy reaching the target building should not be rewarded.

OpenRA.Mods.RA/Traits/Infiltration/Infiltrates.cs
penev92: We try to give modders and mappers as much flexibility for whatever crazy ideas they might have as possible. `¯\_(ツ)_/¯ `

penev92:
:+1: 
The cursor is not disabled, but it uses the same logic as the attack cursor, so forcing it *will* give the enabled one.

OpenRA.Mods.RA/Activities/Infiltrate.cs
Phrohdoh: Then you'll need to pass in the requested bool, too.

GraionDilach:
Okay, let me go away crying to the corner for adding a tag to support an exploit. >_>

penev92:
Removing the "Changes requested" label as the flag was added.
If someone has a suggestion on how to disable the cursor, please share and add the label again.

OpenRA.Mods.RA/Traits/Infiltration/Infiltrates.cs
```diff
@@ -24,21 +24,24 @@ class InfiltratesInfo : ITraitInfo
 
 		[VoiceReference] public readonly string Voice = "Action";
 
+		[Desc("Enable this if you want the actor be allowed to infiltrate allied buildings.")]
+		public readonly bool AllowInfiltrateAllies = false;
```
pchote: Our convention is to use a Stance enum instead of bools.

GraionDilach: Fixed, sorry about that.

abcdefg30:
I agree with @Mailaender that showing an "enter building" cursor should be avoided.

Phrohdoh:
That is definitely something that needs to be adjusted.

GraionDilach:
I had a headache all day and I'm not sure if this should be the right way but I made it.

Mailaender:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/d96faf08fc0f0ccc42f54e20ce6cde89960dc0ef...b2a3348d602248b46a28a3a8ba4f8af16737947f)

PR #9118, opened 2015-08-20
N64 Special Ops - GDI 1

clemty:
N64 GDI 1, mostly as found in Nyerguds' CNC 1.06
differences:
- AI will not rebuild destroyed base buildings
- looking at the original scripts, the AI would replace the Harvester if buildings were destroyed (?) - changed to maintain a level of two harvesters
- AI will build units more frequently than in the original

edit: map.png blanked (used to contain copyrighted N64 logo)


mods/cnc/maps/n64gdi01/n64gdi01.lua
```diff
@@ -0,0 +1,195 @@
...
+	end)
+
+	Trigger.AfterDelay(DateTime.Seconds(150), AutoCreateTeam)
+
+	NodInfantryProduction()
+
+	Camera.Position = UnitsRally.CenterPosition
+	Actor.Create('flare', true, { Owner = player, Location = Flare.Location })		
+	Reinforce(CommandoReinforcements)
+
```
abcdefg30: Unnecessary new line.

mods/cnc/maps/n64gdi01/n64gdi01.lua
```diff
@@ -0,0 +1,195 @@
...
+			unit.Hunt()
+		end)
+	end
+end
+
+NodInfantryProduction = function()
+	if HandOfNod.IsDead or not HandOfNod.Owner == enemy then
+		return
+	end
+	HandOfNod.Build(inf1,SquadHunt)
```
abcdefg30: One space after the `,`, please.

mods/cnc/maps/n64gdi01/n64gdi01.lua
```diff
@@ -0,0 +1,195 @@
...
+	Trigger.AfterDelay(DateTime.Seconds(15), function() NodInfantryProduction() end)
```
abcdefg30: This can be `Trigger.AfterDelay(DateTime.Seconds(15), NodInfantryProduction)`

mods/cnc/maps/n64gdi01/n64gdi01.lua
```diff
@@ -0,0 +1,195 @@
...
+end
+
+NodVehicleProduction = function(Squad)
+	if Airfield.IsDead or not Airfield.Owner == enemy then
+		return
+	end
+	Airfield.Build(Squad,SquadHunt)
```
abcdefg30: Missing space here, too.

mods/cnc/maps/n64gdi01/map.yaml
```diff
@@ -0,0 +1,813 @@
+MapFormat: 7
+
+RequiresMod: cnc
+
+Title: N64 Special Ops - GDI 1
```
Mailaender: N64 is also a workmark so try to avoid it as well. https://trademarks.justia.com/750/77/n64-75077622.html

GraionDilach: We could refer it as C&C64 instead.

LipkeGu: :+1:  for C&C64

clemty:
updated with requested changes

added polish:
- flare will go out after Obelisk is destroyed
- flare now reveals whole Obelisk (like in original)
- Obelisk has full hitpoints (now gets destroyed OnDamaged)
- Airstrike capability is lost after the Obelisk is destroyed
- added two Apaches so the Nod Airpads have a reason to exist (afaik Apaches existed in the original, I think they get created on mission start when Airpads exist; the import utility does not take that into account)

GraionDilach:
Maybe the foldername should also be changed to cnc64 from n64.

mods/cnc/maps/n64gdi01/n64gdi01.lua
```diff
@@ -0,0 +1,201 @@
...
+	if not unit.IsDead then
+		Trigger.OnIdle(unit, function()
+			unit.AttackMove(UnitsRally.Location, 50)
+			unit.Hunt()
+		end)
+	end
+end
+
+NodInfantryProduction = function()
+	if HandOfNod.IsDead or not HandOfNod.Owner == enemy then
```
Mailaender: This doesn't work. It keeps building infantry after I captured it. Try 

```lua
if NodBarracks.Owner == player then
	return
```

instead

clemty:
- renamed folder/files to cnc64gdi01
- Hand of Nod production now stops when captured

Mailaender:
Seems to work fine now. :white_check_mark:

GraionDilach:
Is this supposed to have the Ion Cannon?

GraionDilach:
Never mind me it's supposed to have. Although according to an N64 replay I found, the Airstrike never disappeared it wasn't one-shot.

https://youtu.be/z-SAWwI4Yts?t=2m

clemty:
I disabled airstrikes on purpose - it didn't feel right to have them in, if I remember correctly, in the original storyline GDI missions you wouldn't get airstrike capability until all Nod SAM sites were destroyed. now you have to at least build the comm center

it would be more fair to disable the ion cannon (maybe in higher difficulty levels?) - ion cannon & airstrike make it easy to win the mission without having to go through the enemy base: airstrike to scout the position of the biolab, ion cannon to instakill it

GraionDilach:
Yeah, true. Well, the map plays quite nice anyway - couldn't really break it although I'm not a LUa guy to properly debug. So atm I'd say :+0.5:

clemty:
looking at the video it seems I got the mission mechanics wrong: you're supposed to weaken the power supply with the airstrike, then blow up the obelisk with the commando, whereas I thought you're supposed to kill the obelisk with the airstrike

also currently the mission won't end if you capture the biolab (only way to win after that is to ion cannon it)

clemty:
Biolab will now self-destruct upon capture, winning the mission
I'm keeping the kill-obelisk-with-airstrike mechanic for now unless a change is requested (I could disable the Obelisk via some dark upgrade-rule magic, but something like #9291 would be handier)

mods/cnc/maps/cnc64gdi01/cnc64gdi01.lua
```diff
@@ -0,0 +1,205 @@
...
+
+AutocreateSquads = { auto1, auto2, auto3, auto4, auto5, auto6, auto7, auto8 }
+
+AttackTriggers = { AttackTrigger1, AttackTrigger2, AttackTrigger3, AttackTrigger4 }
+
+harvester = { "harv" }
+
+SamSites = { SAM01, SAM02 }
+Obelisk = { Obelisk01 }
+BiotechCenter = { Biolab }
```
obrakmann: You don't need these, you can reference `Obelisk01` and `Biolab` directly (as you already do in some places).

mods/cnc/maps/cnc64gdi01/cnc64gdi01.lua
```diff
@@ -0,0 +1,205 @@
...
+Tick = function()
+	if DateTime.GameTime > DateTime.Seconds(5) and player.HasNoRequiredUnits() then
+		player.MarkFailedObjective(destroyBiotechCenterObjective)		
+	end
+end
+
+Reinforce = function(units)
+	Media.PlaySpeechNotification(player, "Reinforce")
+	ReinforceWithLandingCraft(units, lstStart.Location, lstEnd.Location, UnitsRally.Location)
+end
```
obrakmann: You can replace this with the snippet below and then remove the `ReinforceWithLandingCraft` function. We have built-in functions for this now (gdi01, where you probably copied this from, just hasn't been updated to use them).

```lua
Reinforce = function(units)
	Media.PlaySpeechNotification(player, "Reinforce")
	local cargo = Reinforcements.ReinforceWithTransport(player, "oldlst", units, { lstStart.Location, lstEnd.Location }, { lstStart.Location })
	Utils.Do(cargo[2], function(unit) unit.Move(UnitsRally.Location) end)
end
```

mods/cnc/maps/cnc64gdi01/map.yaml
```diff
@@ -0,0 +1,823 @@
...
+		LockSpawn: True
+		LockTeam: True
+		Allies: GDI
+		Enemies: Nod, Creeps
+	PlayerReference@Neutral:
+		Name: Neutral
+		OwnsWorld: True
+		NonCombatant: True
+		Faction: gdi
+	PlayerReference@Creeps:
```
obrakmann: You can remove this player. You set `Creeps: False` in the map options above, so it isn't needed.

obrakmann:
You should also add the lua file to the solution file `OpenRA.sln` in the project's root. Also needs a preview image.

Nice mission otherwise. It has a lot of potential for tweaking, difficulty-wise (for example, removing AutoTarget from the commando or disabling the ion cannon).

mods/cnc/maps/cnc64gdi01/cnc64gdi01.lua
pchote: Actually, we need to do it manually in GDI01 and 02 because `ReinforceWithTransport` does not play well with the open-topped landing crafts.  Fixing that generally would make the function unnecessarily complicated compared with faking it in lua in the missions that need it.

clemty:
updated.
- new map.png
- added mission to openra.sln
- referencing biolab and obelisk01 directly
- disabled Ion Cannon
- helis will come hunt you after 5 minutes

mods/cnc/maps/cnc64gdi01/cnc64gdi01.lua
```diff
@@ -0,0 +1,209 @@
...
+
+AIRebuildHarvesters = function(ai)
+	if AIHarvesterCount == NIL or AIHarvesterCount == 0 then
+		AIHarvesterCount = #ai.GetActorsByType("harv")
+		IsBuildingHarvester = false
+	end
+
+	local CurrentHarvesterCount = #ai.GetActorsByType("harv")
+
+	if CurrentHarvesterCount < AIHarvesterCount and not IsBuildingHarvester then
```
obrakmann: I think you'll need to add a `and not Airfield.IsDead` check here

mods/cnc/maps/cnc64gdi01/cnc64gdi01.lua
```diff
@@ -0,0 +1,209 @@
...
+end
+
+HeliHunt = function()
+	local helicopters = enemy.GetActorsByType("heli")
+	SquadHunt(helicopters)
+end
+
+SquadHunt = function(actors)
+	Utils.Do(actors, function(actor)
+		Trigger.OnIdle(actor, function()
```
obrakmann: This will need an `IsDead` check as well. It could happen that the first unit of a squad has already been killed while the last one is still being produced. As soon as it is finished, this would crash.

obrakmann:
One more check to add, then this is ready to go in. :+1:

mods/cnc/maps/cnc64gdi01/cnc64gdi01.lua
```diff
@@ -0,0 +1,209 @@
+CommandoReinforcements = { "rmbo" }
+MCVReinforcements = { "mcv" }
+
+inf1 = { "e4" }
+auto1 = { "stnk", "stnk" }
```
abcdefg30: I'd suggest using
```
AutocreateSquads =
{
	{ "stnk", "stnk" },
	{ "ftnk", "ftnk" },
	{ "ltnk", "ltnk", "bike" },
	{ "arty", "arty", "bike", "bike" },
	{ "ltnk", "ltnk" },
	{ "stnk", "stnk" },
	{ "ltnk", "ltnk" },
	{ "arty", "arty" }
}
```
It should work the same way and we avoid this `autoX` stuff.

mods/cnc/maps/cnc64gdi01/cnc64gdi01.lua
abcdefg30: (and a check for the ownership.)

clemty:
updated

thanks for the suggestions!
- added hovercraft logo (#9466)
- checking if nod vehicle production hasn't been destroyed
- AutoCreateSquads cleanup

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7594a1b00d6286f327945db5592a113d156f6805)

Biofreak1987:
Building a communications center at the beginning of the mission gives you the airstrike power. Is this indented? It seems to make the mission way too easy.

clemty:
When playing it in C&C95, you'd constantly have the airstrike even without the commcenter (I don't remember how it was on the N64)

I agree, it'd be better to disable airstrikes until all SAMs are destroyed; I've been told it is possible to do that by granting an upgrade, but I don't know how to do that

GraionDilach:
The vid I posted a while ago here shows that N64 also had the airstrike all the time.

jtyle6:
I remember that iron canon was available.

acidtonic:
I used to play this mission *all the time* as a kid as it was super fun to watch the sam sites appear out of the darkness and shoot at the airstrike you could use over and over.

I remember some rather vivid details of this mission. For one, when the commando destroyed the sam sites, a minigunner or two would appear out of the rubble and instantly try to shoot the commando. I would reload the save until luck made each one only release one so the commando wouldn't have most of his health down before the mission began.

The nod base to the bottom right, would pretty much like clockwork, send a single flamethrower trooper every few minutes to attack your main base so I would leave the commando there to stop them without having to worry about it.

There would be a decent attack force from Nod after a few minutes into the base building part and they would send a few light tanks with rocket soldiers and flame tank or two after that.

The GDI side in the N64 game could not build the Turret defense building, they could only build the guard tower and advanced guard tower. Also GDI couldn't build the rocket soldier unless it was multiplayer in most of the maps. Only Nod had rocket soldiers. Nod couldn't build guard towers either.

Anyways it really made my day seeing this map in the list. Noticed the AI still doesn't seem to really do much after the initial 4-helicopter strike they do early on.

PR #8953, opened 2015-08-05
Force helicopters to land before transforming.

pchote:
Fixes the issue shown in http://www.ppmforums.com/viewtopic.php?p=534110

This does not need a flag because Transforms is not to be used for aircraft -> aircraft transformations.  We have the upgrade system for intra-actor state changes.

PR #9124, opened 2015-08-21
Make use of DeployToUpgrade

### LINKED ISSUE #8932: DeployToUpgrade does not provide any animations or sounds.

penev92:
Enhance `DeployToUpgrade` and make use of it for the TS Mobile Sensor Array.

This is split into nice small commits that explain what is happening and why.
While reviewing by commits is not necessary here, it would likely be easier.

First three commits set the scene, next 4 commits add the necessary logic to `DeployToUpgrade`, and finally we have a use-case for finally using this trait.
(_[Awesomer Mobile Sensor Array](http://i.imgur.com/pBfXvvA.gif) is back!!_)

Notes:
- [x] The Mobile Sensor Array's deployed version (along with the deploy animation) has half a cell vertical offset. I noticed this after rebasing, and as shown in the gif, it wasn't an issue several months ago. Also not present in the normal `gadpsa`.
- [x] The original has `RadarInvisible=yes` as shown [here](https://github.com/LipkeGu/TSIniFiles/blob/4cc5adac88b012e27a6a82ba7c93d182451d9d64/rules.ini#L2073), so I've [removed](https://github.com/OpenRA/OpenRA/pull/9124/files#diff-20162b6ff85d8da38c40f5095744bad2R152) the `AppearsOnRadar` trait. But do we actually want this?

When this is all over I will remove the current mobile sensor array in favor of the new one (but keep the original name and tooltip, of course).
After that Tick tanks and Nod Artilleries can be done as well, thanks to @reaperrr's efforts (and mine :P ) toward making the necessary traits upgradable. (Ofc RA2 GIs can also be done now.)

Closes  #8932.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
@@ -11,6 +11,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using OpenRA.Activities;
 using OpenRA.Mods.Common.Orders;
 using OpenRA.Traits;
 
@@ -32,6 +33,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Cursor to display when unable to (un)deploy the actor.")]
 		public readonly string DeployBlockedCursor = "deploy-blocked";
 
+		[SequenceReference, Desc("Animation to play for deploying/undeploying.")]
+		public readonly string DeployAnimation = null;
+
 		public object Create(ActorInitializer init) { return new DeployToUpgrade(init.Self, this); }
 	}
 
@@ -41,6 +45,7 @@ namespace OpenRA.Mods.Common.Traits
 		readonly DeployToUpgradeInfo info;
 		readonly UpgradeManager manager;
 		readonly bool checkTerrainType;
+		readonly Lazy<ISpriteBody> body;
 
 		bool isUpgraded;
 
@@ -50,6 +55,7 @@ namespace OpenRA.Mods.Common.Traits
 			this.info = info;
 			manager = self.Trait<UpgradeManager>();
 			checkTerrainType = info.AllowedTerrainTypes.Length > 0;
+			body = Exts.Lazy(self.TraitOrDefault<ISpriteBody>);
 		}
 
 		public IEnumerable<IOrderTargeter> Orders
@@ -75,11 +81,40 @@ namespace OpenRA.Mods.Common.Traits
 				return;
 
 			if (isUpgraded)
-				foreach (var up in info.Upgrades)
-					manager.RevokeUpgrade(self, up, this);
+			{
+				// Play undeploy animation and after that revoke the upgrades
+				self.QueueActivity(new CallFunc(() =>
+				{
+					if (string.IsNullOrEmpty(info.DeployAnimation))
+					{
+						RevokeUpgrades();
+						return;
+					}
+
+					if (body.Value != null)
+						body.Value.PlayCustomAnimationBackwards(self, info.DeployAnimation, RevokeUpgrades);
+					else
+						RevokeUpgrades();
+				}));
+			}
 			else
-				foreach (var up in info.Upgrades)
-					manager.GrantUpgrade(self, up, this);
+			{
+				self.CancelActivity();
+
+				// Grant the upgrade
+				self.QueueActivity(new CallFunc(GrantUpgrades));
+
+				// Play deploy animation
+				self.QueueActivity(new CallFunc(() =>
+				{
+					if (string.IsNullOrEmpty(info.DeployAnimation))
+						return;
+
+					if (body.Value != null)
+						body.Value.PlayCustomAnimation(self, info.DeployAnimation,
+							() => body.Value.PlayCustomAnimationRepeating(self, "idle"));
+				}));
+			}
 
 			isUpgraded = !isUpgraded;
 		}
@@ -97,6 +132,18 @@ namespace OpenRA.Mods.Common.Traits
 			var terrainType = tileSet[tileSet.GetTerrainIndex(tiles[self.Location])].Type;
 
 			return info.AllowedTerrainTypes.Contains(terrainType);
+		}
+
+		void GrantUpgrades()
+		{
+			foreach (var up in info.Upgrades)
+				manager.GrantUpgrade(self, up, this);
+		}
+
+		void RevokeUpgrades()
+		{
+			foreach (var up in info.Upgrades)
+				manager.RevokeUpgrade(self, up, this);
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
@@ -95,7 +95,7 @@ namespace OpenRA.Mods.Common.Traits
 			if (isUpgraded)
 			{
 				// Play undeploy animation and after that revoke the upgrades
-				self.QueueActivity(new CallFunc(() =>
+				self.QueueActivity(false, new CallFunc(() =>
 				{
 					if (!string.IsNullOrEmpty(info.UndeploySound))
 						Sound.Play(info.UndeploySound, self.CenterPosition);
--- a/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
@@ -40,6 +40,12 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Facing that the actor must face before deploying. Set to -1 to deploy regardless of facing.")]
 		public readonly int Facing = -1;
 
+		[Desc("Sound to play when deploying.")]
+		public readonly string DeploySound = null;
+
+		[Desc("Sound to play when undeploying.")]
+		public readonly string UndeploySound = null;
+
 		public object Create(ActorInitializer init) { return new DeployToUpgrade(init.Self, this); }
 	}
 
@@ -91,6 +97,9 @@ namespace OpenRA.Mods.Common.Traits
 				// Play undeploy animation and after that revoke the upgrades
 				self.QueueActivity(new CallFunc(() =>
 				{
+					if (!string.IsNullOrEmpty(info.UndeploySound))
+						Sound.Play(info.UndeploySound, self.CenterPosition);
+
 					if (string.IsNullOrEmpty(info.DeployAnimation))
 					{
 						RevokeUpgrades();
@@ -114,9 +123,12 @@ namespace OpenRA.Mods.Common.Traits
 				// Grant the upgrade
 				self.QueueActivity(new CallFunc(GrantUpgrades));
 
-				// Play deploy animation
+				// Play deploy sound and animation
 				self.QueueActivity(new CallFunc(() =>
 				{
+					if (!string.IsNullOrEmpty(info.DeploySound))
+						Sound.Play(info.DeploySound, self.CenterPosition);
+
 					if (string.IsNullOrEmpty(info.DeployAnimation))
 						return;
 
--- a/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
@@ -12,6 +12,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Activities;
+using OpenRA.Mods.Common.Activities;
 using OpenRA.Mods.Common.Orders;
 using OpenRA.Traits;
 
@@ -36,6 +37,9 @@ namespace OpenRA.Mods.Common.Traits
 		[SequenceReference, Desc("Animation to play for deploying/undeploying.")]
 		public readonly string DeployAnimation = null;
 
+		[Desc("Facing that the actor must face before deploying. Set to -1 to deploy regardless of facing.")]
+		public readonly int Facing = -1;
+
 		public object Create(ActorInitializer init) { return new DeployToUpgrade(init.Self, this); }
 	}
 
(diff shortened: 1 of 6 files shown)


OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -32,6 +34,18 @@ public class DeployToUpgradeInfo : ITraitInfo, Requires<UpgradeManagerInfo>
 		[Desc("Cursor to display when unable to (un)deploy the actor.")]
 		public readonly string DeployBlockedCursor = "deploy-blocked";
 
+		[Desc("Facing that the actor must face before deploying. Set to -1 to not care about facing.")]
```
Unit158: "not care" -> "ignore"?

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -32,6 +34,18 @@ public class DeployToUpgradeInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
+		public readonly int Facing = 0;
+
+		[Desc("Animation to play for deploying/undeploying.")]
+		public readonly string DeployAnimation = "";
```
Unit158: Needs a sequence reference.

GraionDilach:
Animation and sounds should be optional imo.

mods/ts/rules/shared-vehicles.yaml
```diff
@@ -130,6 +130,59 @@ LPST:
...
+			gdi: lpst2.gdi
+			nod: lpst2.nod
+	WithSpriteBody@deployed:
+		StartSequence: make
+		UpgradeTypes: deployed
+		UpgradeMinEnabledLevel: 1
+	DisableUpgrade:
+		UpgradeTypes: deployed
+		UpgradeMinEnabledLevel: 1
+		UpgradeMaxAcceptedLevel: 2
```
GraionDilach: You shouldn't need these MaxAccepted: 2 statements - you're not even checking for them at all during revoking anyway.

GraionDilach:
The deployed sensor bug seems to be caused by the asset positioning differences between buildings and units, unit artworks are always centered on the actual positioon while buildings center their top cell's top point there You have to tweak the offset in yaml for that, I fear.

Compare the following two shots: 
deploysensor vs 
wolverine

mods/ts/rules/shared-vehicles.yaml
```diff
@@ -130,6 +130,59 @@ LPST:
...
+	Health:
+		HP: 600
+	Armor:
+		Type: Light
+	Mobile:
+		Speed: 85
+		ROT: 5
+	RevealsShroud:
+		Range: 10c0
+	-AppearsOnRadar:	# TODO: Discuss whether we want to remove the trait. The original had RadarInvisible=yes
```
Mailaender: Let's spare the balancing discussion for now and just keep original Tiberian Sun behavior while we are still working on the technical parts.

GraionDilach:
#8953 could be added in too - YR Siege Chopper comes into my mind.

penev92:
The code here is still too young, and the aircraft code is undergoing a major overhaul (I have a PR that starts it, hopefully it will get finished eventually), so I'd like to skip this for now. Right now there is no non-hacky way of doing it.

GraionDilach:
Okay, I'm fine with that for now.

mods/ts/rules/shared-vehicles.yaml
penev92: I'll keep the comment for now, until I'm sure there will be no objections.

penev92:
Made deploy animation and sounds optional (as they should have been) and removed `UpgradeMaxAcceptedLevel`.
Will keep looking into the offset, but will most likely need assistance with that.

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -32,6 +34,18 @@ public class DeployToUpgradeInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
+		public readonly int Facing = 0;
+
+		[Desc("Animation to play for deploying/undeploying.")]
+		[SequenceReference] public readonly string DeployAnimation = "";
+
+		[Desc("Sound to play when deploying.")]
+		public readonly string DeploySound = "";
+
+		[Desc("Sound to play when undeploying.")]
+		public readonly string UnDeploySound = "";
```
GraionDilach: Undeploy, imo.

GraionDilach:
Remove the defined Offsets in `gadpsa`'s sequence to fix that offset issue, just checked.

GraionDilach:
Copy the offset of `gadpsa:Defaults` to `lpst2` to fix the offset issue. I'm stupid. >_>

penev92:
That fixed the offset for the actor, but screwed up the production icon :S

penev92:
Removed the comment from `-AppearsOnRadar:` and renamed `UnDeploySound`.

penev92:
Changed default values of `DeployAnimation`, `DeploySound`, `UndeploySound` and `Facing` to `null`.

LipkeGu:
@penev92 : sorry this was an bad asume that lpst is using place2 as deploy... not all units are using place2 as deploy sound. the lpst is using clicky other tanks are using place2

OpenRA.Game/Traits/BodyOrientation.cs
```diff
@@ -69,7 +70,7 @@ public BodyOrientation(ActorInitializer init, BodyOrientationInfo info)
 				if (info.QuantizedFacings >= 0)
 					return info.QuantizedFacings;
 
-				var qboi = self.Info.Traits.GetOrDefault<IQuantizeBodyOrientationInfo>();
+				var qboi = self.Info.Traits.WithInterface<IQuantizeBodyOrientationInfo>().FirstOrDefault();
```
pchote: Is this still needed?

pchote: Taking the first IQBOI is a bad idea, but IIRC @reaperrr has already worked around needing to do this for sprite code.

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -32,6 +34,18 @@ public class DeployToUpgradeInfo : ITraitInfo, Requires<UpgradeManagerInfo>
 		[Desc("Cursor to display when unable to (un)deploy the actor.")]
 		public readonly string DeployBlockedCursor = "deploy-blocked";
 
+		[Desc("Facing that the actor must face before deploying. Will deploy disregarding facing by default.")]
+		public readonly int? Facing = null;
```
pchote: IIRC convention is to use `-1` as the magic value for automatic / no facing.

OpenRA.Mods.Common/Widgets/ProductionPaletteWidget.cs
```diff
@@ -309,7 +309,7 @@ public void RefreshIcons()
 				var y = DisplayedIconCount / Columns;
 				var rect = new Rectangle(rb.X + x * (IconSize.X + IconMargin.X), rb.Y + y * (IconSize.Y + IconMargin.Y), IconSize.X, IconSize.Y);
 
-				var rsi = item.Traits.Get<RenderSpritesInfo>();
+				var rsi = item.Traits.WithInterface<RenderSpritesInfo>().First(t => t.IsTraitEnabled());
```
pchote: Actors should never have more than one `RenderSprites` trait, so this is not necessary.

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
penev92: It was, but @Phrohdoh made a somewhat good case for using `int?` instead, and thus not needing to use (the otherwise valid value) -1 as a magic invalid default.

pchote: IMO it would be better to keep this consistent with the rest of our code.  Somebody could argue for changing all of them in a different pr, but the other parts of the code may run into performance regressions due to unboxing.

GraionDilach: I'm not certain if that can be disabled when inherited though.

penev92: Yeah that is a problem which I noticed when testing the suggestion. Will revert to -1  :/

OpenRA.Game/Traits/BodyOrientation.cs
pchote: See #9004.

penev92:
I guess this depends on #9004 so I can remove the `BodyOrientation` change :/

reaperrr:
Needs rebase (#9004 was merged).

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -32,6 +34,18 @@ public class DeployToUpgradeInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
+
+		[Desc("Animation to play for deploying/undeploying.")]
+		[SequenceReference] public readonly string DeployAnimation = null;
```
Phrohdoh: Nitpick: Could you combine these two attributes?

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -75,11 +89,52 @@ public void ResolveOrder(Actor self, Order order)
...
+					if (!string.IsNullOrEmpty(info.UndeploySound))
+						Sound.Play(info.UndeploySound, self.CenterPosition);
+
+					if (string.IsNullOrEmpty(info.DeployAnimation))
+					{
+						RevokeUpgrades();
+						return;
+					}
+
+					var render = self.TraitOrDefault<ISpriteBody>();
```
Phrohdoh: Why are you now restricting this to Sprites?

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -75,11 +89,52 @@ public void ResolveOrder(Actor self, Order order)
...
+				}));
+			}
 			else
-				foreach (var up in info.Upgrades)
-					manager.GrantUpgrade(self, up, this);
+			{
+				self.CancelActivity();
+
+				// Turn
+				if (info.Facing.HasValue && self.HasTrait<IFacing>())
```
Phrohdoh: This `HasTrait` can probably be cached.

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
GraionDilach: Why?

Phrohdoh: So the `DeployAnimation` is on the same indentation level as the fields.

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
penev92: Nobody said this is the final version of the trait, but without a test case I can't reliably make it support every possible scenario. Also it would be much easier to digest if it's not all at once, I assume.

Phrohdoh: I never implied it was the final version.
But this cannot then be used on most of our deployers (voxels), as far as I know.

penev92: All of them do Voxel -> Sprite -> Voxel (with Thumpers and GIs being the Sprite -> Sprite exception), so this will actually be good enough for the official mods (custom mods may want Voxel -> Voxel, with a custom MakeAnimation on the deployed voxel, but all of our cases will work.

GraionDilach: Siege Chopper was voxel-voxel, but YR is out of the scope for now I guess.

penev92:
- Reverted `Facing` to being `int`.
 - Beautified `DeployAnimation`'s attributes.
 - Removed the second commit "Don't crash on multiple `RenderSprites` traits" (an actor should not have multiple of those).
 - Removed the third commit "Don't crash on multiple traits implementing `IQuantizeBodyOrientationInfo`" (now that #9004 is merged there is no more need).
 - Switched from inheriting `^Vehicle` to `^VoxelVehicle`.
 - Added shadow to the deployed version.
 - Cached trait lookups for both `IFacing` and `ISpriteBody`.

Mailaender:
`LPST2` does not show a cloak reveal range circle. Seems to be otherwise fine.

GraionDilach:
The circle issue needs #8761 I guess.

penev92:
It does display the reveal range. It's just a lot bigger than the one on `LPST` - 25 cells instead of 6. [Reference](https://github.com/LipkeGu/TSIniFiles/blob/4cc5adac88b012e27a6a82ba7c93d182451d9d64/rules.ini#L4252).
Also #8761 is not related to the `LPST`.  ;)

GraionDilach:
Wait, that's just 18 (17.6) cells in OpenRA terms!

Again, OpenRA measures the diagonals, while WW measured the sides...

penev92:
I knew there was something in the range conversion I was missing, but there was nobody online to ask at the time...
Thanks, will fix shortly.

reaperrr:
The move order support is not a blocker, :+1:  once the yaml changes have been made.

penev92:
Fixed the ranges (detection *and* sight) and removed the duplicate actor and sequences.
Haven't addressed the automatic undeploy comment as I'm currently not sure how to proceed with that.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/afdcdb050cea0b41ab9f94c6117e41768ead181f...2efcf22959d4fa2bbec1de835aece1b6c4fbb94c)

PR #9130, opened 2015-08-22
Replaced the blank shellmap with something more interesting

Mailaender:
http://ra2maps.zzattack.org/Tiberian%20Sun/GDI%20Campaign/getmap.php?map=GDI+04+-+Defend+Crash+Site to be exact.


penev92:
Overall I like it.
@dan9550 was working on an active (with tons of action) shellmap, using this exact same map as base, but that can be merged when it's done.
Two issues I have with this:
 - [ ] You should reduce the camera horizontal movement so that it doesn't go (partially) outside of the map from the East and West edges.
 - [ ] I feel like you should reduce the green in the lighting.

After that I can +1.

Mailaender:
Updated.

penev92:
:+1:

reaperrr:
:+1:

PR #9131, opened 2015-08-22
Added a Tiberian Sun loadscreen and shellmap logo

Mailaender:
Just like https://github.com/OpenRA/OpenRA/pull/9130 this isn't meant to be final in any way although they look quite nice together. This replaces the last bit of Tiberian Dawn artwork from the mod. Editable image source: https://github.com/OpenRA/ArtSrc/pull/2.


LipkeGu:
We should keep the preview image to show that this Mod is in still in development stage :)

penev92:
Sorry, I'm not convinced that it looks better.

Mailaender:
It adds consistency and uses a more subtle placeholder like https://github.com/OpenRA/OpenRA/pull/7186. From what I read in IRC there are more ambitious plans for crafting and coding the in-game UI, but those have a tendencies to take very long and be abandoned quickly so this is better than nothing when only ideas and mock-ups exist I guess.

GraionDilach:
It looks much more fitting than that GDI logo for sure. :+1:

Issue #8323, opened 2015-05-31
Ingame Map Editor : Forbid to place several actors on the same location


x-a-n-a-x:
The ingame map editor allows to place several actors on the same location, resulting in having overlapping tanks for example.

While it could be interesting to have 3 or 5 infantry units at the same location, it doesn't make sense for tanks or other vehicles.. Maybe the sidebar could offer: 1 infantry unit, 3 infantry units and 5 infantry units to be placed?

http://logs.openra.net/?year=2015&month=05&day=31#16:34:06

Mailaender:
I assume we don't want this following the discussion at https://github.com/OpenRA/OpenRA/pull/9138.

PR #9138, opened 2015-08-23
Added checks for occupied cells in the map editor

Mailaender:
This disallows stacking actors in occupied cells and subcells.

Closes https://github.com/OpenRA/OpenRA/issues/8323.

### DIFF
--- a/OpenRA.Mods.Common/EditorBrushes/EditorActorBrush.cs
+++ b/OpenRA.Mods.Common/EditorBrushes/EditorActorBrush.cs
@@ -107,11 +107,28 @@ namespace OpenRA.Mods.Common.Widgets
 				newActorReference.Add(new LocationInit(cell));
 
 				var ios = Actor.Traits.GetOrDefault<IOccupySpaceInfo>();
-				if (ios != null && ios.SharesCell)
+				if (ios != null)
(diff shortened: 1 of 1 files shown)


pchote:
There are cases (such as placing dense forests) where it is necessary to overlap actors.

LipkeGu:
We could allow this by YAML key such like "AllowOccupy:"

GraionDilach:
Is there any particular reason why trees are using building code? Having them taking infantry subcells would be a lot more elegant imo - and still would let them be a dense forest.

I'm aware that TD/RA do have some tree artwork where there are more trees in a single shp (where I understand infantry trees are not a solution), but this does not apply to the later games.

penev92:
Because of FrozenActor and others requiring Immobile stuff to be Buildings... It's kind of a legacy mess. I had this talk recently with @pchote on IRC, search for "Immobile" in the logs if you're interested.

Mailaender:
I think trees are always 1x1 cells or you can adjust the footprint to make it not occupy everything. However yes this will restrict creativity a bit. It may be impossible to put i.e. a santa clause infantry on an ice shelf for example as the ice actor is then already occupying the cell.

PR #9089, opened 2015-08-16
Added a coordinate display to the map editor

Mailaender:
See the label widget at the top of the screen.

PR #9139, opened 2015-08-23
Added a total resource cash counter to the map editor

Mailaender:
We had this in the legacy WinForms application https://github.com/OpenRA/OpenRA/pull/2894. This together with https://github.com/OpenRA/OpenRA/pull/9089 will give it feature parity with what we already had displayed in the status bar.

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
+++ b/OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
@@ -38,6 +38,8 @@ namespace OpenRA.Mods.Common.Traits
 
 		readonly Dictionary<PaletteReference, TerrainSpriteLayer> spriteLayers = new Dictionary<PaletteReference, TerrainSpriteLayer>();
 
+		public int NetWorth { get; private set; }
+
 		public EditorResourceLayer(Actor self)
 		{
 			if (self.World.Type != WorldType.Editor)
@@ -88,6 +90,10 @@ namespace OpenRA.Mods.Common.Traits
 		{
 			var uv = cell.ToMPos(Map);
 			var tile = Map.MapResources.Value[uv];
+
+			var t = Tiles[cell];
+			if (t.Density > 0)
+				NetWorth -= t.Density * t.Type.Info.ValuePerUnit;
 
 			ResourceType type;
 			if (Resources.TryGetValue(tile.Type, out type))
@@ -149,8 +155,12 @@ namespace OpenRA.Mods.Common.Traits
 				return t;
 			}
 
+			NetWorth -= t.Density * type.Info.ValuePerUnit;
+
 			// Set density based on the number of neighboring resources
 			t.Density = ResourceDensityAt(c);
+
+			NetWorth += t.Density * type.Info.ValuePerUnit;
 
 			var sprites = type.Variants[t.Variant];
 			var frame = int2.Lerp(0, sprites.Length - 1, t.Density - 1, type.Info.MaxDensity);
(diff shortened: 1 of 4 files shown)


RoosterDragon:
:+1:

OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
```diff
@@ -113,6 +115,17 @@ public void UpdateCell(CPos cell)
 				Dirty.Add(cell + d);
 		}
 
+		void CalculateNetWorth()
```
pchote: This is a bit ugly, and makes for a lot of unnecessary calculations.  You should instead update the state incrementally from inside UpdateCell.

Mailaender: That won't work as I need to take the density into account.

pchote: All the information you need is available from inside `UpdateDirtyTile`, so use that instead of `UpdateCell`.

Mailaender: I tried a few more complex and simple things like

```patch
+			NetWorth -= t.Density * type.Info.ValuePerUnit;
+
 			// Set density based on the number of neighboring resources
 			t.Density = ResourceDensityAt(c);
 
+			NetWorth += t.Density * type.Info.ValuePerUnit;
+
```

which will always end in the same result: miscalculations when removing a lone cell without neighbors and unknown previous contents.

As the sum over the whole resource layer is an easy calculation my guess is that things like

```csharp
var previousCells = CVec.Directions.Select(d => d + cell).Where(c => Map.Contains(c));
foreach (var previousCell in previousCells)
{
	var t = Tiles[previousCell];
	if (t.Density > 0)
		NetWorth -= t.Density * t.Type.Info.ValuePerUnit;
}
```

might actually be more costly performance wise.

Mailaender:
Updated.

OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
Mailaender: Ah, *selfslap* `CVec.Directions` does not contain 0,0. No wonder this doesn't work.

OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
```diff
@@ -89,6 +91,14 @@ public void UpdateCell(CPos cell)
 			var uv = cell.ToMPos(Map);
 			var tile = Map.MapResources.Value[uv];
 
+			var previousCells = CVec.Directions.Select(d => d + cell).Where(c => Map.Contains(c)).Append(cell);
```
pchote: This is unnecessary.

Mailaender: How so?

pchote:
You can simplify this further.  See https://github.com/pchote/OpenRA/commit/cbe0bbbc89067d0ad35eccbb2f64d469e5d3d56c

penev92:
:+1:

penev92:
[Changelog is in here somewhere.](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/6e1538a7bfc79b70770194c736f8ae39ef59b1d1...98625547054b9d6ae8bb8135573f9bfe1a6da407)

PR #9140, opened 2015-08-23
Improve pathfinder performance

RoosterDragon:
A number of commits intended to make the pathfinder faster by reducing small inefficiencies in the parts of the engine it relies on.

This PR should not affect the pathfinder output at all. For testing I recommend recording a replay on bleed and replaying it with this PR - there should be no desyncs unless I've messed up.

Overall, these changes save ~21% in pathfinding time.

PR #9141, opened 2015-08-23
Improve pathfinder performance (paths may change)

RoosterDragon:
Depends on #9140.

Introduces two changes which improve pathfinder performance, but which can change the resulting paths. The changes should be minor and the new paths shouldn't be drastically different, but we should keep an eye out just in case.

Overall, these changes save ~18% in pathfinding time.

### DIFF
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -508,7 +508,7 @@
     <Compile Include="Pathfinder\CellInfoLayerManager.cs" />
     <Compile Include="Pathfinder\Constants.cs" />
     <Compile Include="Pathfinder\PathGraph.cs" />
-    <Compile Include="Pathfinder\PathFinderCacheDecorator.cs" />
+    <Compile Include="Pathfinder\PathFinderUnitPathCacheDecorator.cs" />
     <Compile Include="Pathfinder\PathCacheStorage.cs" />
     <Compile Include="Traits\World\PathFinder.cs" />
     <Compile Include="Traits\World\PathfinderDebugOverlay.cs" />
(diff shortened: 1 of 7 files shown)


reaperrr:
What I noticed is that actors seem to more likely choose a different (potentially much longer) path when blocked on narrow paths (fords, bridges etc.), but that is still better than actors giving up and sitting idle in any case.

Performance improvement over bleed is quite noticable.

reaperrr:
Needs rebase.

RoosterDragon:
Rebased.

Mailaender:
The calculated paths are still valid. I can't spot negative side effects. :white_check_mark:

pchote:
I keep on running into QuantizedBodyOrientation crashes while trying to test this.  Can you please rebase on top of the latest bleed?

pchote:
The code changes look fine, so :+0.5: until I am able to properly test this.

reaperrr:
To a glance at the code and saw nothing wrong, though that doesn't mean much in my case.

I rebased on latest bleed before testing and didn't run into any crashes.

The paths look fine to me, not much different to bleed.

As for performance:
- The spikes at the beginning of RA's shellmap are notably smaller now.
Generally In RA and TD, the pathing performance is now acceptable even on larger maps with larger amounts of rivers, cliffs and so on. The very first move order given to a large army tends to be larger, consecutive move orders are usually barely noticable. Rendering the path previews in DevMode is now more expensive than the pathfinding itself, which can probably be considered a good thing.

- In D2k, pathfinding is now pretty much a non-issue, since there aren't as many obstacles and bottlenecks, never ran into any pf-related lag.

- Pathfinding performance on TS maps is still disappointingly bad, but fixing that will probably require some more disruptive changes like group pathfinding. It's probably better than on bleed, but for large armies the spike can still go off-screen on first move order so I can't tell by how much. Out of scope for this PR anyway.

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/bace266959041fc0c0e1a2f17b25bf9ec9a331d6...c6aa850e3cae3e2f2606513544eed70b7a131a94)

LipkeGu:
this together with the other pf pr is improving the PF performance drastically ... in TS :O

PR #9143, opened 2015-08-23
LoadScreen: Do not crash when Image or Text is not defined.

LipkeGu:
This crash can be reproduced without this PR by commenting out the loading image entry and loading text entry in mod.yaml. In this case the Pr adds a fallback loading message.

IMO In this unexpected case it is better to display `Loading...` instead of crashing the game.

### DIFF
--- a/OpenRA.Mods.Common/LoadScreens/LogoStripeLoadScreen.cs
+++ b/OpenRA.Mods.Common/LoadScreens/LogoStripeLoadScreen.cs
@@ -25,7 +25,7 @@ namespace OpenRA.Mods.Common.LoadScreens
 		float2 logoPos;
 		Sheet sheet;
 		Sprite stripe, logo;
-		string[] messages;
+		string[] messages = { "Loading..." };
 
 		public override void Init(Manifest m, Dictionary<string, string> info)
 		{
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/LoadScreens/LogoStripeLoadScreen.cs
```diff
@@ -35,12 +36,19 @@ public override void Init(Manifest m, Dictionary<string, string> info)
...
 				return;
 
-			messages = info["Text"].Split(',');
-			sheet = new Sheet(Platform.ResolvePath(info["Image"]));
-			logo = new Sprite(sheet, new Rectangle(0, 0, 256, 256), TextureChannel.Alpha);
-			stripe = new Sprite(sheet, new Rectangle(256, 0, 256, 256), TextureChannel.Alpha);
-			stripeRect = new Rectangle(0, r.Resolution.Height / 2 - 128, r.Resolution.Width, 256);
-			logoPos = new float2(r.Resolution.Width / 2 - 128, r.Resolution.Height / 2 - 128);
+			if (!info.ContainsKey("Text"))
+				messages = defaultMessage.Split(',');
```
Mailaender: Nothing to split there.

LipkeGu: Thats expected, so you can set more default messages

OpenRA.Mods.Common/LoadScreens/LogoStripeLoadScreen.cs
```diff
@@ -25,6 +25,7 @@ public sealed class LogoStripeLoadScreen : BlankLoadScreen
 		float2 logoPos;
 		Sheet sheet;
 		Sprite stripe, logo;
+		string defaultMessage = "Loading...";
```
Mailaender: What I mean is `string[] defaultMessage = { "Loading..."; };` would make more sense.

OpenRA.Mods.Common/LoadScreens/LogoStripeLoadScreen.cs
Mailaender: and then do `messages = defaultMessage` which is less of a hack.

Mailaender:
Another thing: this is the `LogoStripeLoadScreen` so having `Image` stay empty kinda defeats the purpose. You could make a `TextOnlyLoadScreen` and then maybe inherit it.

OpenRA.Mods.Common/LoadScreens/LogoStripeLoadScreen.cs
penev92: Not really. This is a hardcoded fallback string. Nobody is going to add more things to it.

LipkeGu: Its only fall back that the Game doesnt crash when no text is defined.

LipkeGu:
but this is not fixing the crash

Mailaender:
I agree. It definitely doesn't hurt to make those optional. :+1:

OpenRA.Mods.Common/LoadScreens/LogoStripeLoadScreen.cs
```diff
@@ -35,12 +35,17 @@ public override void Init(Manifest m, Dictionary<string, string> info)
...
 				return;
 
-			messages = info["Text"].Split(',');
-			sheet = new Sheet(Platform.ResolvePath(info["Image"]));
-			logo = new Sprite(sheet, new Rectangle(0, 0, 256, 256), TextureChannel.Alpha);
-			stripe = new Sprite(sheet, new Rectangle(256, 0, 256, 256), TextureChannel.Alpha);
-			stripeRect = new Rectangle(0, r.Resolution.Height / 2 - 128, r.Resolution.Width, 256);
-			logoPos = new float2(r.Resolution.Width / 2 - 128, r.Resolution.Height / 2 - 128);
+			if (info.ContainsKey("Text"))
+				messages = info["Text"].Split(',');
```
Mailaender: > [SP2000] Invalid spacing at the end of the line.

LipkeGu: allready seen this ^^

RoosterDragon:
:+1:

PR #9144, opened 2015-08-23
Optimized FindResources worse case scenario path finding

Mailaender:
This will help with simple cases like no resources on the map, but also more complex ones such as no resources left in path-able range on a large island maps:

image

The map was http://resource.openra.net/maps/2928/ and removed due to the unbearable late-game lag.

### DIFF
--- a/OpenRA.Mods.Common/Activities/FindResources.cs
+++ b/OpenRA.Mods.Common/Activities/FindResources.cs
@@ -27,6 +27,7 @@ namespace OpenRA.Mods.Common.Activities
 		readonly ResourceLayer resLayer;
 		readonly ResourceClaimLayer territory;
 		readonly IPathFinder pathFinder;
+		readonly DomainIndex domainIndex;
 
 		CPos? avoidCell;
 
@@ -39,6 +40,7 @@ namespace OpenRA.Mods.Common.Activities
 			resLayer = self.World.WorldActor.Trait<ResourceLayer>();
 			territory = self.World.WorldActor.TraitOrDefault<ResourceClaimLayer>();
 			pathFinder = self.World.WorldActor.Trait<IPathFinder>();
+			domainIndex = self.World.WorldActor.Trait<DomainIndex>();
 		}
 
 		public FindResources(Actor self, CPos avoidCell)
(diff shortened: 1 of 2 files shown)


LipkeGu:
How will this behaving in TS on the Veins?

Mailaender:
No change actually. `IsHarvestable` already did the check before submitting the path search.

reaperrr:
AI uses `Harvester.FindNextResourceForBot`, could you implement at least the 2nd fix there as well?

Mailaender:
I removed `Harvester.FindNextResourceForBot` completely as it seems to be not only redundant, but also expensive doing another full map path search just to simulate where a human would click to initiate the real search for resources.

reaperrr:
:+1:

OpenRA.Mods.Common/Traits/World/ResourceLayer.cs
```diff
@@ -35,6 +35,8 @@ public class ResourceLayer : IRenderOverlay, IWorldLoaded, ITickRender, INotifyA
 		protected readonly CellLayer<CellContents> Content;
 		protected readonly CellLayer<CellContents> RenderContent;
 
+		public bool IsEmpty { get { return !Content.Any(); } }
```
RoosterDragon: A CellLayer is just the set of cells on the map, so it's not ever going to be empty. Did you intend this to check if there were any resources on the map?

Mailaender: Yes.

Mailaender: Removed.

penev92:
Seems to work quite well.
:+1: 

On a side note, that map with 4 AIs had relatively very little lag. We should consider bringing some of those nice huge maps back after all the performance improvements that were done in the past few months.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/b0c8737009859e1bfae3e236599d77cfab88eb0b...51a557be36ba5098a4c3f698f7ea78d8de584a95)

Mailaender:
[Optimized changelog specificity to make it more accessible to players.](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/51a557be36ba5098a4c3f698f7ea78d8de584a95...263729c227af442822b2b08286f91369d5ec60d7)

Mailaender:
This needs some tweaks though. https://github.com/OpenRA/OpenRA/issues/9232 This just removed the ugly workarounds that made things work in practice. It didn't replace them with more clever anti-deadlock resource searching. I didn't expect it to regress this hard, but watching a mid game RA bot match reveals the weakness. Sorry.

PR #9147, opened 2015-08-23
Added the current map to the exception.log

Mailaender:
as an increasing number of crashes are just broken modded maps which isn't always obvious.


penev92:
:+1: nice

reaperrr:
:+1:

PR #8632, opened 2015-07-04
Removed the OpenAL/OpenGL/SDL2 dependency from the game engine

Mailaender:
I removed the OpenAL (and subsequent OpenTK) dependency from OpenRA.Game and placed it into the renderer DLLs, which in fact do a lot more than just rendering nowadays (input, window creation, etc.) so I renamed things to reflect this.

Our SDL2/OpenGL/OpenAL combination works fine, but isn't far from ideal. Now it is possible for anyone to create a DirectX platform DLL or something more light weight new based on GLFW or other stripped down OpenGL bindings etc.

Closes #5274.

PR #9150, opened 2015-08-23
Make lower-level Sound not require the server state directly

penev92:
As pointed out in https://github.com/OpenRA/OpenRA/pull/8632#discussion_r37711021

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -275,7 +275,7 @@ namespace OpenRA
 			Settings.Game.Mod = mod;
 
 			Sound.StopVideo();
-			Sound.Initialize(Settings.Sound, Settings.Server);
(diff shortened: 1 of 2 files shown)


abcdefg30:
:+1:

OpenRA.Game/Game.cs
```diff
@@ -275,7 +275,7 @@ public static void InitializeMod(string mod, Arguments args)
 			Settings.Game.Mod = mod;
 
 			Sound.StopVideo();
-			Sound.Initialize(Settings.Sound, Settings.Server);
+			Sound.Initialize(Settings.Sound, Settings.Server.Dedicated);
```
pchote: Pass the engine name directly.

penev92:
Passed the engine name directly.

PR #9152, opened 2015-08-24
Renderer: Do not use (Pseudo-)Fullscreen when Debugger is attached.

LipkeGu:
This Pr implements a check which forces the Renderer to run OpenRA in Window Mode when a Debugger is attached. This elimimnates also the problem that you cant switch back to the IDE when the IDE freezes the Process ( in such cases Windows user has to log out and re login.

The actions are the following:
- User is runnig OpenRA with attached Debugger (Visual studio, etc) = OpenRA is running in a Window
- User runs OpenRA without Debugger = The selected mode from settings will be used.

### DIFF
--- a/OpenRA.Game/Renderer.cs
+++ b/OpenRA.Game/Renderer.cs
@@ -10,6 +10,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Drawing;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


Unit158:
:-1: There have been fullscreen related bugs in the past.

LipkeGu:
Did you realy understand the Description?

Unit158:
Yes, and it is invalid at least for me. Windows 7 with VS 2013. This can also potentially affect our MonoDevelop users.

LipkeGu:
on my laptop with VS 2013 with Windows 7 i dont have problems to select window mode im talking about PseudoFullScreen not real Fullscreen.

LipkeGu:
What are the Problems on your site?

penev92:
Yes, playing in fullscreen ends up being a bad idea when the game crashes in debug. Yes, I have been running OpenRA in a tiny window for the last 9 months because otherwise debugging just doesn't work.
But that is no reason to hardcode such an aggressive limitation for no good reason. People can just select windowed mode if they wish. And sometimes they may need to see as much of the game as possible.
So :-1:

LipkeGu:
@penev92  That is what the check Does?!

Unit158:
We need pseudo-fullscreen for testing bugs with the pseudo-fullscreen feature. I have never had a problem so severe that I've had to log out, and I am quite sure that I can provide steps to get out of a mess as such if need be. You may use this code in your own version (please feel free to write another dll to include this feature in for yourself) but I do not agree that this should be a forced limitation. Especially since it can be remedied otherwise.

LipkeGu:
I will make this selectable via settings....

PR #9159, opened 2015-08-25
Add PlugInit

penev92:
Add `PlugInit` for use with map-placed structures in TS (or whatever custom mod decides to use plugs). There have been a couple of requests by people who wanted to have plugs on their maps.
I also figured I should use this to make the towers on the new TS shellmap look better.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Pluggable.cs
+++ b/OpenRA.Mods.Common/Traits/Pluggable.cs
@@ -9,40 +9,44 @@
 #endregion
 
 using System.Collections.Generic;
-using System.Linq;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
-	public class PluggableInfo : ITraitInfo, Requires<UpgradeManagerInfo>
+	public class PluggableInfo : ITraitInfo, Requires<UpgradeManagerInfo>, UsesInit<PlugsInit>
 	{
 		[Desc("Footprint cell offset where a plug can be placed.")]
 		public readonly CVec Offset = CVec.Zero;
 
-		[FieldLoader.LoadUsing("LoadUpgrades", true)]
-		[Desc("Upgrades to grant for each accepted plug type.")]
+		[FieldLoader.Require, Desc("Upgrades to grant for each accepted plug type.")]
 		public readonly Dictionary<string, string[]> Upgrades = null;
 
-		static object LoadUpgrades(MiniYaml y)
-		{
-			return y.ToDictionary()["Upgrades"].Nodes.ToDictionary(
-				kv => kv.Key,
-				kv => FieldLoader.GetValue<string[]>("(value)", kv.Value.Value));
-		}
-
-		public object Create(ActorInitializer init) { return new Pluggable(init.Self, this); }
+		public object Create(ActorInitializer init) { return new Pluggable(init, this); }
 	}
 
-	public class Pluggable
+	public class Pluggable : INotifyCreated
 	{
 		public readonly PluggableInfo Info;
+
+		readonly string initialPlug;
 		readonly UpgradeManager upgradeManager;
+
 		string active;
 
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Traits/Pluggable.cs
```diff
@@ -39,10 +39,13 @@ public class Pluggable
 		readonly UpgradeManager upgradeManager;
 		string active;
 
-		public Pluggable(Actor self, PluggableInfo info)
+		public Pluggable(ActorInitializer init, PluggableInfo info)
 		{
 			Info = info;
-			upgradeManager = self.Trait<UpgradeManager>();
+			upgradeManager = init.Self.Trait<UpgradeManager>();
+			var plugInit = init.Contains<PlugInit>() ? init.Get<PlugInit, string[]>() : new string[0];
```
Phrohdoh: Newline after this please.

penev92:
Things took a turn for the worse when I decided to experiment with the GDI Upgrade Center.
I had to find a way for the `PlugInit` to specify on which slot of the host to be "placed". I must say it's not the prettiest code I've written :/
I added a test commit to showcase all available plugs in all possible combinations, to give people a better idea of how this would work in a scenario where the map author actually needs to put more than one plug on an actor.

Mailaender:
Works as promised in the shellmap. :white_check_mark: Seems to be ignored in the map editor as in not rendered, not just no support for adding them. There are some StyleCop issues that needs to be resolved to please Travis.

Phrohdoh:
Could you make this a `Dictionary<int2, string>` so we could have a sub-tree like so:

```yaml
Plugs:
    0, 0: plug.ioncannon
    1, 0: plug.powr
```

penev92:
Sorry, not possible until `FieldLoader` learns how to deal with Dictionaries.
Squashed the second commit into the first one.

pchote:
FieldLoader learned how to deal with dictionaries years ago.  Check out things like Mobile and GainsExperience.

penev92:
Those do their own FieldLoading, which is by far not the same thing. Also doesn't seem feasable for ActorInits.

penev92:
This now depends on #9182. Sorry I did try to do dictionary support when I filed this PR, but failed at the time.
On the bright side, the code for this PR looks much better now.

penev92:
This should be good to go now.

OpenRA.Mods.Common/Traits/Pluggable.cs
```diff
@@ -71,4 +75,13 @@ public void DisablePlug(Actor self, string type)
 				upgradeManager.RevokeUpgrade(self, u, this);
 		}
 	}
+
+	public class PlugInit : IActorInit<Dictionary<CVec, string>>
```
Phrohdoh: `PlugsInit`, please.

OpenRA.Mods.Common/Traits/Pluggable.cs
```diff
@@ -9,40 +9,44 @@
...
 
 namespace OpenRA.Mods.Common.Traits
 {
-	public class PluggableInfo : ITraitInfo, Requires<UpgradeManagerInfo>
+	public class PluggableInfo : ITraitInfo, Requires<UpgradeManagerInfo>, UsesInit<PlugInit>
 	{
 		[Desc("Footprint cell offset where a plug can be placed.")]
 		public readonly CVec Offset = CVec.Zero;
 
-		[FieldLoader.LoadUsing("LoadUpgrades", true)]
```
pchote: `[FieldLoader.Required]`

penev92:
Renamed to `PlugsInit`, added `FieldLoader.Require`, rebased and removed the test commit.

pchote:
:+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/2cfc5c677533d94d8b2a47e8b419da945dd8740c

PR #9162, opened 2015-08-25
[D2K] Add Music Mirrors

LipkeGu:


GraionDilach:
There's nothing there. I don't think that1s even applicable in the long run, since D2k is abandonware....

penev92:
He filed https://github.com/OpenRA/OpenRAWeb/pull/256 to add that file, but I do agree that we don't ship the music for a reason.

Mailaender:
Works as promised. :white_check_mark: Technically it isn't shipped, but an external download.

LipkeGu:
and what is basic on the legals can we get problems?

Mailaender:
The statement from http://www.frankklepacki.com/ via e-Mail is that EA owns all the rights including the music and we have to ask them.

LipkeGu:
what will you mean?

Mailaender:
I will try to contact an EA copyright agent via http://www.ea.com/1/legal-notices

TheRaffy:
good luck with that! I'll keep my fingers crossed.

LipkeGu:
reopened so far...

jaZzKCS:
So is this the reason why the music content has actually been on the content mirror for ages, albeit with a renamed filename so that the client wont have it as an option? I have always been wondering about the weird filenaming.

LipkeGu:
yes... but is now available again

penev92:
Last time we discussed this the consensus was against providing the D2k music for download.
As far as I know nothing has changed.

Mailaender:
As far as I understand it the Dune (2000) license is in total legal limbo. You can't even buy it from Electronic Arts own web shop. It is also pretty impossible to contact anyone in charge at EA if you are a nobody to them so I don't expect some kind of official message any time soon.

When we hosted the d2k assets my idea was: well, they released the Dune 2000 demo and 1.06 patch files for free, so technically we can harvest those like we did Red Alert and Tiberian Dawn/Sun which are still also accompanied with a very restrictive EULA.

I am not a lawyer so don't expect proper legal advice and I also don't host the mirrors so the responsibilities is off there as well. When I presented the d2k mod there were some brave people who uploaded the ZIP files and many mirrors followed. Let's say we can't really stop you from doing the same with the music, but we also can't protect you from potential DMCA take downs.

obrakmann:
My view on the matter is that we should steer clear of this. While the license (for using the Dune universe) is indeed in limbo, the music has nothing to do with that, it's only coupled to the game. So theoretically EA could probably pull it off to sell the Dune 2000 soundtrack, for example (that assumption depends on what exactly the license entailed, and they might not be able to use the Dune 2000 name for it, but still). We'd be too far out of the legal grey area for my liking if we did it.

But IANAL and all that.

pchote:
It looks like we still have a consensus to not do this &rarr; closing.

PR #9168, opened 2015-08-26
Allowed turrets to randomly realign when idle.

Unit158:
This should also fix the popup turrets.

### DIFF
--- a/OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
+++ b/OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
@@ -15,7 +15,7 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Cnc.Traits
 {
 	[Desc("Actor's turret rises from the ground before attacking.")]
-	class AttackPopupTurretedInfo : AttackTurretedInfo, Requires<BuildingInfo>, Requires<RenderBuildingInfo>
+	class AttackPopupTurretedInfo : AttackFollowInfo, Requires<BuildingInfo>, Requires<RenderBuildingInfo>
 	{
 		[Desc("How many game ticks should pass before closing the actor's turret.")]
 		public int CloseDelay = 125;
@@ -27,7 +27,7 @@ namespace OpenRA.Mods.Cnc.Traits
 		public override object Create(ActorInitializer init) { return new AttackPopupTurreted(init, this); }
 	}
 
-	class AttackPopupTurreted : AttackTurreted, INotifyBuildComplete, INotifyIdle, IDamageModifier
+	class AttackPopupTurreted : AttackFollow, INotifyBuildComplete, INotifyIdle, IDamageModifier
 	{
 		enum PopupState { Open, Rotating, Transitioning, Closed }
 
@@ -43,7 +43,7 @@ namespace OpenRA.Mods.Cnc.Traits
 			: base(init.Self, info)
 		{
 			this.info = info;
-			turret = turrets.FirstOrDefault();
+			turret = init.Self.Trait<Turreted>();
 			rb = init.Self.Trait<RenderBuilding>();
 			skippedMakeAnimation = init.Contains<SkipMakeAnimsInit>();
 		}
@@ -68,6 +68,9 @@ namespace OpenRA.Mods.Cnc.Traits
 				return false;
 			}
 
+			if (state != PopupState.Open)
+				return false;
+
 			return turret.FaceTarget(self, target);
 		}
 
@@ -75,7 +78,7 @@ namespace OpenRA.Mods.Cnc.Traits
 		{
 			if (state == PopupState.Open && idleTicks++ > info.CloseDelay)
 			{
-				turret.DesiredFacing = info.DefaultFacing;
+				turret.Face(self, info.DefaultFacing);
 				state = PopupState.Rotating;
 			}
 			else if (state == PopupState.Rotating && turret.TurretFacing == info.DefaultFacing)
@@ -85,7 +88,7 @@ namespace OpenRA.Mods.Cnc.Traits
 				{
 					state = PopupState.Closed;
 					rb.PlayCustomAnimRepeating(self, "closed-idle");
-					turret.DesiredFacing = null;
+					turret.Face(self, null);
 				});
 			}
 		}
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -23,6 +23,9 @@ public class TurretedInfo : ITraitInfo, UsesInit<TurretFacingInit>, Requires<IBo
 		[Desc("Number of ticks before turret is realigned. (-1 turns off realignment)")]
 		public readonly int RealignDelay = 40;
 
+		[Desc("Does the turret realign to a random facing while idle?")]
+		public readonly bool RandomRealign = true;
```
penev92: This should be false to keep current behavior.

Unit158: This is for testing :)

OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
```diff
@@ -43,7 +43,7 @@ public AttackPopupTurreted(ActorInitializer init, AttackPopupTurretedInfo info)
 			: base(init.Self, info)
 		{
 			this.info = info;
-			turret = turrets.FirstOrDefault();
+			turret = init.Self.Trait<Turreted>();
```
penev92: This will throw for actors with more than one `Turreted` trait.

Unit158: This can't be used for actors with more than one Turreted trait anyway.

Unit158:
wrt the pop-up turrets
`[21:29]	Unit|SASL	penev: they dont close on release iirc
[21:29]	Unit|SASL	and it fixes a visual glitch where the turret would rotate "instantly" due to animation stuffs`

penev92:
OK, so there are Realign(), MoveTurret(), Face() and FaceTarget() in Turreted.
Is it me, or does that feel excessive?
(Also no Turn() or Rotate(), which would make the most sense IMO.)

Unit158:
Ok, with this PR, the turreted code has hit a wall.

The code can work as it does, with the chances of race conditions (these will exist no matter what, really) or we can postpone these changes and completely rewrite the turreted code as a step towards the Attack* refactor. I haven't lost very much time from this now, so I want a clear and concise decision before I continue any further.

Unit158:
FaceTarget can be removed, I keep it there to keep changes to a minimum. Realign is something completely different. MoveTurret should probably go, too.

OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -62,48 +67,95 @@ public Turreted(ActorInitializer init, TurretedInfo info)
...
 		{
-			DesiredFacing = Util.GetFacing(target.CenterPosition - self.CenterPosition, TurretFacing);
+			realignTick = realign ? 0 : realignTick;
+			desiredFacing = facing;
+
 			MoveTurret();
-			return TurretFacing == DesiredFacing.Value;
+
+			if (facing == null)
+				return TurretFacing == (facing != null ? this.facing.Facing : TurretFacing);
```
abcdefg30: The `facing != null` check seems redundant to me, as you are already checking `facing == null` above.

abcdefg30:
It also looks like `Realign` is duplicating `Tick`?

abcdefg30:
```
OpenRA.Mods.Common/Traits/Turreted.cs:L99: [SA1513] Statements or elements wrapped in curly brackets must be followed by a blank line.
OpenRA.Mods.Common/Traits/Turreted.cs:L128: [SA1513] Statements or elements wrapped in curly brackets must be followed by a blank line.

File: OpenRA.Mods.Common/Traits/Turreted.cs [hunk 4]
```diff
@@ -62,48 +67,95 @@ public Turreted(ActorInitializer init, TurretedInfo info)
 		{
 			this.info = info;
 			TurretFacing = GetInitialTurretFacing(init, info.InitialFacing);
+
+			if (info.RandomRealign)
+				randomFacing = init.Self.World.SharedRandom.Next(0, 256);
```
GraionDilach: Can't this get a parameter specifying the max random angles compared to body orientation? This method looks really bad on tanks.

GraionDilach: Meh, I miswrote, I meant body facing instead of orientation. >_>

Unit158: If it used angles relative to the body, yes. Unfortunately, that isn't what it does and was not what I signed up to do (yet).

When the turret code gets rewritten for the Attack refactor, I'll think about it. Right now that is out of scope.

penev92: How is that any harder though

Unit158: I have to grab the facing each tick (or whenever it changes), do some arithmetic to make sure it is within the bounds defined, rotate it within these bounds. If the tank moves while rotating, I'd need to get another value within the limit before the rotation has finished. It would really be easier if I could do this with WAngles or proper angle support over facings.

GraionDilach: Why?

Sounds overly complicated to me. It should work exactly like how `WithInfantryBody` picks an idle animation, and moving/attacking should switch back to face-where-you're-going/attacking behaviour. Generals and beyond did that, using turret realigning as an idle anim.

Unit158: Because facings aren't animations o.O

GraionDilach: The action to turn the turret to a random angle is pretty much an equivalent to an animation in this case to me. The point was in the _when-to-do-this-action_ aspect and not in the _what-is-this_ aspect, iow that you could just implement `INotifyIdle` and `INotifyBecomingIdle` adding a timer, and move all the `randomFacing` changes to `TickIdle` after the timer expired.

penev92: Yeah, unfortunately @Unit158 is right here. I tried doing it with `INotifyBecomingIdle` today, it won't work :/

GraionDilach: Probably https://github.com/OpenRA/OpenRA/pull/6503#issuecomment-56280668 still applies then. Alright then... although I'm not certain what is the gain of this realigning logic if it ends up such blatantly stupid.

Unit158:
No consensus has been reached with regards to a refactor of the turret code, thus I am closing this PR.

PR #9174, opened 2015-08-26
Customizable impact victim scan radius

### LINKED ISSUE #9173: Something is off about warhead spread/health radius interaction
zero-damage

Health radius `1c0` on the cyard, `Spread: 96` on the `QuadRockets`, gave `BlocksProjectiles` trait to cyard, missiles explodes barely within health radius, zero damage. 
As far as I understand the concept of the health radius and spread, it should deal full damage within 96 wdist units (3 pixel in d2k) of the outer border of the health radius, 37% at 192 wdist/6px etc. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26212526-something-is-off-about-warhead-spread-health-radius-interaction?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Fixes #9173.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Bullet.cs
+++ b/OpenRA.Mods.Common/Effects/Bullet.cs
@@ -164,8 +164,14 @@ namespace OpenRA.Mods.Common.Effects
 			if (info.ContrailLength > 0)
 				contrail.Update(pos);
 
-			if (ticks++ >= length || (info.Blockable && world.ActorMap
-				.GetUnitsAt(world.Map.CellContaining(pos)).Any(a => a.Info.HasTraitInfo<IBlocksProjectilesInfo>())))
+			var cell = world.Map.CellContaining(pos);
+			var height = world.Map.DistanceAboveTerrain(pos);
+
+			var shouldExplode = height.Length <= 0 // Hit the ground
+				|| ticks++ >= length // Flight length reached/exceeded
+				|| (info.Blockable && world.ActorMap.GetUnitsAt(cell).Any(a => a.Info.HasTraitInfo<IBlocksProjectilesInfo>())); // Hit a wall or other blocking obstacle
+
+			if (shouldExplode)
 				Explode(world);
 		}
 
--- a/OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs
@@ -19,6 +19,9 @@ namespace OpenRA.Mods.Common.Warheads
 		[Desc("Range between falloff steps.")]
 		public readonly WDist Spread = new WDist(43);
 
+		[Desc("Extra search radius beyond maximum spread. Required to ensure damage to actors with large health radius.")]
+		public readonly WDist TargetExtraSearchRadius = new WDist(2048);
+
 		[Desc("Damage percentage at each range step")]
 		public readonly int[] Falloff = { 100, 37, 14, 5, 2, 1, 0 };
 
@@ -46,7 +49,10 @@ namespace OpenRA.Mods.Common.Warheads
 				InitializeRange();
 
 			var world = firedBy.World;
-			var hitActors = world.FindActorsInCircle(pos, Range[Range.Length - 1]);
+
+			// This only finds actors where the center is within the search radius,
+			// so we need to search beyond the maximum spread to account for actors with large health radius
+			var hitActors = world.FindActorsInCircle(pos, Range[Range.Length - 1] + TargetExtraSearchRadius);
 
 			foreach (var victim in hitActors)
 			{
--- a/OpenRA.Mods.Common/Lint/CheckTargetHealthRadius.cs
+++ b/OpenRA.Mods.Common/Lint/CheckTargetHealthRadius.cs
@@ -1,0 +1,55 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Mods.Common.Warheads;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Lint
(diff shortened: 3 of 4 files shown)


OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs
```diff
@@ -19,6 +19,9 @@ public class SpreadDamageWarhead : DamageWarhead
 		[Desc("Range between falloff steps.")]
 		public readonly WDist Spread = new WDist(43);
 
+		[Desc("Extra search radius beyond maximum spread. Required to ensure damage to actors with large health radius.")]
+		public readonly WDist VictimExtraSearchRadius = new WDist(2048);
```
pchote: Can we get a lint rule that checks all of these vs all of the health radii?

reaperrr: Makes sense, I guess. I'll give it a try.

Unit158:
Why would the raycaster help? I'll be happy to get that code up and running again for you. I'll just need a little bit of your brain in exchange :p

reaperrr:
Rebased and added lint rule.

Note: The Travis error is caused by the lint testcase to demonstrate that the rule works.

OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs
reaperrr: Done.

reaperrr:
Updated, renamed `Arm` to `ArmDelay` in `Bullet` and fixed its height check in `shouldExplode`.

reaperrr:
I've removed the lint testcase since the failed travis annoyed me, but you can easily test it:

- give any `SpreadDamageWarhead` a custom `VictimExtraSearchRadius` with a value of less than `426`, but more than `128` (which is the `Health.Radius` default), then `make test`. It should fail on any non-infantry actor.
- give any actor a `Health.Radius` of more than `2c0`, then the test will fail on every `SpreadDamageWarhead` on that actor since the default `VictimExtraSearchRadius` is less than that.

OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs
pchote: Can we call this `Target` instead of `Victim` for consistency?

pchote:
Looks sensible overall, but I haven't tested it.  Needs a rebase.

reaperrr:
Rebased.

reaperrr:
I'm beginning to think I should go for a `ExplodeOnContactType` instead, with (initial) available types being 
- `ContactType.Blocker` (actors with BlocksProjectiles, would replace `Blockable` boolean), 
- `ContactType.EnemyHealthRadius` (if you want this to trigger only on enemies), 
- `ContactType.AlliedHealthRadius` (if you want this to only trigger on allies), 
- `ContactType.HealthRadius` (if you want this to trigger on any health radius). 

We could then later add more types like `.Shield` (makes the projectile blow up outside the health radius and only damage the shield).


If I do that, it would probably be better to split that off to its own PR and only deal with the warhead target check radius in this PR.

Thoughts?

UberWaffe:
@reaperrr Would you need ray-casting between two WPos to do some of these?
There is a working & tested (but only by myself) ray-casting function in the pending pull request  #9357.

It currently only takes the actor's health radius into account, along with the 'width' of the beam, but it has proven to work well (touch wood) for making the Line Impact Projectiles stop / impact blocking actors regardless of the speed of the Line Impact Projectile.
See disruptor beam (slow moving) and ghost stalker railgun (instant hit) for test cases.
I'm not entirely happy with how I 'redirect' them when being blocked, but that is not a fault of the ray-caster itself, but rather the projectile code.

atlimit8:
Needs to be rebased.

reaperrr:
Rebased.

Additionally, I decided to split off the ExplodeOnContact logic (will file a PR later) since that might take longer, and this here is both a simpler feature and a bugfix as well, so it should not be delayed any longer than necessary.

OpenRA.Mods.Common/Effects/Bullet.cs
```diff
@@ -164,8 +164,14 @@ public void Tick(World world)
 			if (info.ContrailLength > 0)
 				contrail.Update(pos);
 
-			if (ticks++ >= length || (info.Blockable && world.ActorMap
-				.GetUnitsAt(world.Map.CellContaining(pos)).Any(a => a.Info.HasTraitInfo<IBlocksProjectilesInfo>())))
+			var cell = world.Map.CellContaining(pos);
+			var height = world.Map.DistanceAboveTerrain(pos);
+
+			var shouldExplode = (height.Length <= 0) // Hit the ground
```
RoosterDragon: Unnecessary brackets.

reaperrr:
Updated.

penev92:
Looks reasonable :+1:

RoosterDragon:
:+1: Code looks good and confirmed this fixes your reproduction case from #9173..

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4b93e7fdf229588fd7c18a1f45de7485db7223fc...8e31fb4bbbfbb7ef7a202a821fca4d3dc02f25e6)

reaperrr:
The Bullet's new `ArmDelay` breaks several instant-hit weapons (just watch RA shellmap), I'll file a fix shortly.

Mailaender:
This check breaks for https://github.com/OpenRA/OpenRA/pull/10613. I need help as I don't understand it.

GraionDilach:
`TargetExtraSearchRadius` has been decreased to 1536 since then, but since the gates have a diagonal of sqrt2*1.5 cells (2173 wu), all warhead TESR which can affect it needs to be bigger than the aforementioned 2173.

reaperrr:
> EDIT: Oh wait, Radius, these needs to be doubled to 4346 as the minimum, -_-

Nope, `TargetExtraSearchRadius` is compared against `HitShape.OuterRadius` so it's radius vs. radius, no need to double them.

Also, `2048` is enough according to my tests.

PR #9175, opened 2015-08-26
Update MaxMind.GeoIP2 version to 2.3.1

penev92:
Fixes an "Operation did not complete successfully because the file contains a virus." message when making dependencies.

### DIFF
--- a/thirdparty/fetch-thirdparty-deps.sh
+++ b/thirdparty/fetch-thirdparty-deps.sh
@@ -49,10 +49,10 @@ fi
 
 if [ ! -f MaxMind.GeoIP2.dll ]; then
 	echo "Fetching MaxMind.GeoIP2 from NuGet"
-	get MaxMind.Db 1.0.0.0
-	get Newtonsoft.Json 6.0.5
-	get RestSharp 105.0.1
-	get MaxMind.GeoIP2 2.1.0
+	get Newtonsoft.Json 7.0.1
+	get MaxMind.Db 1.1.0.0
+	get RestSharp 105.2.3
+	get MaxMind.GeoIP2 2.3.1
 	cp ./MaxMind.Db/lib/net40/MaxMind.Db.* .
 	rm -rf MaxMind.Db
 	cp ./MaxMind.GeoIP2/lib/net40/MaxMind.GeoIP2* .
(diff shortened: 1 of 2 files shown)


Mailaender:
This is not enough.

> Unable to resolve dependency 'MaxMind.Db (≥ 1.1.0)'.

thirdparty/fetch-thirdparty-deps.sh
```diff
@@ -52,7 +52,7 @@ if [ ! -f MaxMind.GeoIP2.dll ]; then
 	get MaxMind.Db 1.0.0.0
 	get Newtonsoft.Json 6.0.5
 	get RestSharp 105.0.1
```
Mailaender: You need to update those as well. They are not automatically resolved and manually specified as noget has no dependency resolution of it's own.

penev92:
Updated the versions of `MaxMind.GeoIP2`'s dependencies.
I'd like to have at least one person on each OS test this, so we're sure this is not breaking stuff.

thirdparty/fetch-thirdparty-deps.sh
```diff
@@ -49,10 +49,10 @@ fi
 
 if [ ! -f MaxMind.GeoIP2.dll ]; then
 	echo "Fetching MaxMind.GeoIP2 from NuGet"
-	get MaxMind.Db 1.0.0.0
-	get Newtonsoft.Json 6.0.5
-	get RestSharp 105.0.1
-	get MaxMind.GeoIP2 2.1.0
+	get MaxMind.Db 1.1.0.0
+	get Newtonsoft.Json 6.0.8
```
Mailaender: > Unable to resolve dependency 'Newtonsoft.Json (≥ 7.0.1)'.

Please change this accordingly https://www.nuget.org/packages/MaxMind.Db/

penev92:
Should be good now.

Mailaender:
Still hitting the error. You need to re-order the dependencies:

>	get Newtonsoft.Json 7.0.1
>	get MaxMind.Db 1.1.0.0

Mailaender:
Works. :white_check_mark: Thanks and sorry for the hazzle. We will also need to clear the Travis CI build cache manually as it seems.

RoosterDragon:
Works for me on Windows.

Mailaender:
Cleaned the Travis cache and rebuild. `noget` variant also works. :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c5fe2c8f7597976da42bddc601fe72033637e061...a1bc3e252cb73bc5347be9a5997ff224aae4990d)

PR #9178, opened 2015-08-27
Move duplicate Tick() code from Plane and Helicopter to Aircraft.

LavenderMoon:


penev92:
Hello.
Our CI system is complaining that
```
OpenRA.Mods.Common/Traits/Air/Aircraft.cs:L86: [SA1515] A single-line comment must be preceded by a blank line or another single-line comment, or must be the first item in its scope. To ignore this error when commenting out a line of code, begin the comment with '////' rather than '//'.
```
You can fix it by adding a new line before the comment.
Other than that the changes look good :+1:

LavenderMoon:
Whoops, didn't think to run StyleCop. Will keep that in mind in the future.

RoosterDragon:
Can you squash the fix into the original commit please?

LavenderMoon:
Will do. Just gotta figure out this rebase thing.

penev92:
Start an interactive rebase, then either squash, or fixup the second commit into the first one.
Also you can join us on our IRC channel :)

LavenderMoon:
Took me a while to figure out that I was using vim to edit the interactive rebase. It should be good now.

RoosterDragon:
:+1:

PR #9179, opened 2015-08-27
Fix radar not showing frozen actors

### LINKED ISSUE #9155: Mini-map/radar not showing enemy structures under shroud
(playtest 20150808)
Let's say you scout an enemy base and retreat... the mini-map does not show enemy structures when you lose vision of the base.

Same behaviour with GPS satelite. 

Is this intended? 
If so **it might be helpful to at least get a marker/dot/flag on the mini-map meaning "enemy base found here".**

Maybe I kind of like it... :)

penev92:
Fixes #9155.

Also saves trait lookups on every render tick in `AppearsOnRadar`.

Individual commits should be pretty simple to review on a 1-by-1 basis.
Not sure if we want this for stable.

### DIFF
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -39,7 +39,11 @@ namespace OpenRA.Traits
 		public int HP;
 		public DamageState DamageState;
 
-		public bool Visible = true;
+		/// <summary>
+		/// Is the frozen preview of the real actor currently visible?
+		/// That means the real actor is currently under fog.
+		/// </summary>
+		public bool IsVisible = true;
 		public bool NeedRenderables;
 		public bool IsRendering { get; private set; }
 
@@ -85,22 +89,22 @@ namespace OpenRA.Traits
 
 		void UpdateVisibility()
 		{
-			var wasVisible = Visible;
+			var wasVisible = IsVisible;
 			var isVisibleTest = shroud.IsVisibleTest;
 
 			// We are doing the following LINQ manually for performance since this is a hot path.
-			// Visible = !Footprint.Any(isVisibleTest);
-			Visible = true;
+			// IsVisible = !Footprint.Any(isVisibleTest);
+			IsVisible = true;
 			foreach (var uv in Footprint)
 			{
 				if (isVisibleTest(uv))
 				{
-					Visible = false;
+					IsVisible = false;
 					break;
 				}
 			}
 
-			if (Visible && !wasVisible)
+			if (IsVisible && !wasVisible)
 				NeedRenderables = true;
 		}
 
@@ -188,7 +192,7 @@ namespace OpenRA.Traits
 
 				if (frozenActor.ShouldBeRemoved(owner))
 					remove.Add(kvp.Key);
-				else if (frozenActor.Visible)
+				else if (frozenActor.IsVisible)
 					VisibilityHash += hash;
 				else if (frozenActor.Actor == null)
 					remove.Add(kvp.Key);
@@ -204,7 +208,7 @@ namespace OpenRA.Traits
 		public virtual IEnumerable<IRenderable> Render(Actor self, WorldRenderer wr)
 		{
 			return world.ScreenMap.FrozenActorsInBox(owner, wr.Viewport.TopLeft, wr.Viewport.BottomRight)
-				.Where(f => f.Visible)
+				.Where(f => f.IsVisible)
 				.SelectMany(ff => ff.Render(wr));
 		}
 
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -12,6 +12,7 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
 using OpenRA.Graphics;
+using OpenRA.Primitives;
 
 namespace OpenRA.Traits
 {
@@ -47,6 +48,9 @@ namespace OpenRA.Traits
 		public ActorInfo Info { get { return actor.Info; } }
 		public Actor Actor { get { return !actor.IsDead ? actor : null; } }
 
+		public Color RadarColor;
+		public HashSet<CPos> OccupiedCells;
+
 		public FrozenActor(Actor self, PPos[] footprint, Shroud shroud)
 		{
 			actor = self;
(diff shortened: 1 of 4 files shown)


OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -211,5 +220,14 @@ public FrozenActor FromID(uint id)
...
 			return ret;
 		}
+
+		public IEnumerable<Pair<CPos, Color>> RadarSignatureCells(Actor self)
+		{
+			// Filter only actors that have been discovered by the player
+			// HasRenderables is only True once the actor enters the viewport, so is not enough for the minimap
+			foreach (var frozenActor in frozen.Values.Where(f => f.NeedRenderables || f.HasRenderables))
+				foreach (var cell in frozenActor.OccupiedCells)
+					yield return Pair.New(cell, frozenActor.RadarColor);
```
RoosterDragon: Maybe you can cache the radar signature for each frozen actor (since it tends to remain fixed) so you don't have to re-calculate this all the time.

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -211,5 +220,14 @@ public FrozenActor FromID(uint id)
 
 			return ret;
 		}
+
+		public IEnumerable<Pair<CPos, Color>> RadarSignatureCells(Actor self)
+		{
+			// Filter only actors that have been discovered by the player
+			// HasRenderables is only True once the actor enters the viewport, so is not enough for the minimap
+			foreach (var frozenActor in frozen.Values.Where(f => f.NeedRenderables || f.HasRenderables))
```
RoosterDragon: Maybe we should rename ```NeedsRenderables``` to ```HasBeenSeen``` or something like that, since it's not strictly about renderables any more.

OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
```diff
@@ -120,6 +123,13 @@ public void Tick(Actor self)
 					frozenActor.TooltipInfo = tooltip.Value.TooltipInfo;
 					frozenActor.TooltipOwner = tooltip.Value.Owner;
 				}
+
+				if (ownerHasChanged[player])
+				{
+					var mod = self.TraitsImplementing<IRadarColorModifier>().FirstOrDefault();
+					frozenActor.RadarColor = mod != null ? mod.RadarColorOverride(self) : self.Owner.Color.RGB;
+					ownerHasChanged[player] = false;
+				}
```
RoosterDragon: It is possible to do all this when the owner changes rather than checking on every tick?

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -50,21 +62,18 @@ public FrozenActor(Actor self, PPos[] footprint, Shroud shroud)
...
-				.Where(m => shroud.Contains(m))
-				.ToArray();
+			Footprint = footprint.Where(shroud.Contains).ToArray();
 
 			CenterPosition = self.CenterPosition;
 			Bounds = self.Bounds;
 
+			var mod = self.TraitsImplementing<IRadarColorModifier>().FirstOrDefault();
+			RadarColor = mod != null ? mod.RadarColorOverride(self) : self.Owner.Color.RGB;
+			OccupiedCells = new HashSet<CPos>(footprint.SelectMany(uv => self.World.Map.Unproject(uv).Select(m => m.ToCPos(self.World.Map))));
```
RoosterDragon: Does this have to be a set? Might be more efficient to do ```.Distinct.ToArray()``` (unless we need to do set queries on this anywhere, but I don't see any).

OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
penev92: We don't want to change the color until the player for who that actor is frozen reveals it again and sees that it was captured.
Otherwise frozen actors will change colors under the fog, giving away information that you shouldn't know. (The commit message describes that, maybe I should have put a comment in there as well).

RoosterDragon: No that's fine. :)

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
RoosterDragon: Sorry, to clarify: Since we don't do any queries on it, it would be simpler if we stored ```OccupiedCells``` as an array, rather than a set.

penev92: Got it.

pchote: Unprojecting seems a little silly.  I suggest looking a step up the chain and checking whether it is worth passing the original OccupiedCells, and then calculate the footprint in here.

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
pchote: Note that actors can, in principle, return different colors for different cells in their footprint.

pchote: How about you call `RadarSignatureCells` on the original actor when it is visible, and cache what that returns instead?

OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
pchote: Owner changes aren't the only things that can change the color.  Cloaking is another specific example, and there are arbitrary other things that may also do it.

penev92:
This is potentially (quite) prone to regressions, so it would be dangerous  to have into the stable branch. Also I tested the current release (*20150614*!) and also **release_20150424** and the same thing is present there, which means this isn't a recent regression that we *have* to fix now.
In light of that I suggest we move ahead with the playtest and leave this fix for the next development cycle (actually... the one running now on bleed).

abcdefg30:
Also needs a rebase.

obrakmann:
Works as expected. Tested capturing and the stealth generator. It'd be nice if this was picked up again and the other review comments were resolved. The bug is pretty bad. :+1:

RoosterDragon:
I think my earlier comments are largely irrelevant, so you probably just need to address @pchote's.

abcdefg30:

PR #9181, opened 2015-08-27
Support HashSet in FieldLoader/FieldSaver

RoosterDragon:
Adds support to load and save a `HashSet` directly.

This means classes can directly use a set if they need set-based operations, rather than suffering the cost of querying an array.

Lots of places may benefit from changing to sets eventually, for the moment I have limited it to simplifying `WeaponInfo` which was already creating a set for performance reasons.


chrisforbes:
:+1:

penev92:
:+1:

Phrohdoh:
I think documentation generation should have been updated as well.

PR #9182, opened 2015-08-28
Add support for Dictionary to FieldLoader and FieldSaver

penev92:
Inspired by #9181, this makes `FieldLoader` capable of loading `Dictionary<,>` directly, instead of traits relying on their own code to load them in a number of different ways.
Also `FieldSaver` is not left out, and is able to save them as YAML (for saving maps) and potentially as string, for trait documentation generation.

As usual, split into nice small easy to review commits.

Easiest way to test this is with my `PlugInit` PR, because that uses both the `FieldLoader` and `FieldSaver` changes.

### DIFF
--- a/OpenRA.Game/FieldLoader.cs
+++ b/OpenRA.Game/FieldLoader.cs
@@ -648,6 +648,7 @@ namespace OpenRA
 			public string YamlName;
 			public string Loader;
 			public bool FromYamlKey;
+			public bool DictionaryFromYamlKey;
 			public bool Required;
 
 			public SerializeAttribute(bool serialize = true, bool required = false)
@@ -707,6 +708,17 @@ namespace OpenRA
 		public FieldFromYamlKeyAttribute()
 		{
 			FromYamlKey = true;
+		}
+	}
+
+	// Special-cases FieldFromYamlKeyAttribute for use with Dictionary<K,V>.
+	[AttributeUsage(AttributeTargets.Field)]
+	public sealed class DictionaryFromYamlKeyAttribute : FieldLoader.SerializeAttribute
+	{
+		public DictionaryFromYamlKeyAttribute()
+		{
+			FromYamlKey = true;
+			DictionaryFromYamlKey = true;
 		}
 	}
 
--- a/OpenRA.Game/FieldLoader.cs
+++ b/OpenRA.Game/FieldLoader.cs
@@ -110,13 +110,8 @@ namespace OpenRA
 			if (!md.TryGetValue(yamlName, out yaml))
 				return false;
 
-			if (yaml.Nodes.Count == 0)
-			{
-				ret = GetValue(field.Name, field.FieldType, yaml.Value, field);
-				return true;
-			}
-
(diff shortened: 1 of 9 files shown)


OpenRA.Game/FieldLoader.cs
```diff
@@ -441,6 +442,20 @@ public static object GetValue(string fieldName, Type fieldType, string value, Me
 					addMethod.Invoke(set, new[] { GetValue(fieldName, fieldType.GetGenericArguments()[0], parts[i].Trim(), field) });
 				return set;
 			}
+			else if (fieldType.IsGenericType && fieldType.GetGenericTypeDefinition() == typeof(Dictionary<,>))
+			{
+				var dict = Activator.CreateInstance(fieldType);
+				var addMethod = fieldType.GetMethod("Add", fieldType.GetGenericArguments());
```
pchote: You call `fieldType.GetGenericArguments()` `2 * <number of children> + 1` times.

OpenRA.Mods.Common/Traits/GainsExperience.cs
```diff
@@ -20,7 +19,6 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("This actor's experience increases when it has killed a GivesExperience actor.")]
 	public class GainsExperienceInfo : ITraitInfo, Requires<ValuedInfo>, Requires<UpgradeManagerInfo>
 	{
-		[FieldLoader.LoadUsing("LoadUpgrades", true)]
```
pchote: The `true` indicates that this is a required field.  You will need to add the appropriate attribute to not change behavior here.

OpenRA.Mods.Common/Traits/Infantry/TerrainModifiesDamage.cs
```diff
@@ -17,7 +16,6 @@ namespace OpenRA.Mods.Common.Traits
 {
 	public class TerrainModifiesDamageInfo : ITraitInfo
 	{
-		[FieldLoader.LoadUsing("LoadPercents", true)]
```
pchote: And here.

OpenRA.Game/FieldLoader.cs
penev92: OK, does that give me anything, because the current setup is a lot clearer than that would be I think.

penev92:
Added `[FieldLoader.Require]` to not change behaviour and avoided calling `fieldType.GetGenericArguments()` repeatedly.

Mailaender:
Nice cleanup and can't spot regressions. :white_check_mark:

OpenRA.Game/FieldLoader.cs
```diff
@@ -588,7 +604,7 @@ static FieldLoadInfo[] BuildTypeLoadInfo(Type type)
 
 				var loader = sa.GetLoader(type);
 				if (loader == null && sa.FromYamlKey)
-					loader = (yaml) => GetValue(yamlName, field.FieldType, yaml.Value, field);
+					loader = (yaml) => GetValue(yamlName, field.FieldType, yaml, field);
```
RoosterDragon: Just since we're here:
```(yaml)``` -> ```yaml```.

RoosterDragon:
:+1:

PR #9185, opened 2015-08-28
D2k fix for cliff passability

### LINKED ISSUE #9166: Carryalls disappear under FoW in d2k
This is on bleed. The only problem I have with this is that the carryall makes off with my harvester and unless I am constantly defending my spice patches, I can't see if someone set up a rocket tank to snipe the carryalls.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26174870-carryalls-disappear-under-fow-in-d2k?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Fixes passability on some d2k cliffs.

### DIFF
--- a/mods/d2k/tilesets/arrakis.yaml
+++ b/mods/d2k/tilesets/arrakis.yaml
@@ -309,10 +309,10 @@ Templates:
 		Size: 2,2
 		Category: Sand-Cliff
 		Tiles:
-			0: Rough
-			1: Rough
-			2: Rough
-			3: Rough
+			0: Cliff
+			1: Cliff
+			2: Cliff
+			3: Cliff
 	Template@23:
 		Id: 23
 		Images: BLOXBASE.R8
@@ -320,10 +320,10 @@ Templates:
 		Size: 2,2
 		Category: Sand-Cliff
 		Tiles:
-			0: Rough
-			1: Rough
-			2: Rough
-			3: Rough
+			0: Cliff
+			1: Cliff
+			2: Cliff
+			3: Cliff
 	Template@24:
 		Id: 24
 		Images: BLOXBASE.R8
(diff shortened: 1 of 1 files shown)


mods/d2k/rules/defaults.yaml
```diff
@@ -237,7 +239,7 @@
 	Inherits@1: ^ExistsInWorld
 	AppearsOnRadar:
 		UseLocation: yes
-	HiddenUnderFog:
+	HiddenUnderShroud:
```
pchote: This makes them visible for *everyone* under the fog, but we want them to be visible to only the owner and their allies.

reaperrr: Oh, I see.

pchote:
After refreshing my memory on the history behind #9166, I think we shouldn't fix it.

reaperrr:
Updated.

> After refreshing my memory on the history behind #9166, I think we shouldn't fix it (its a feature, not a bug).

#9106 wants to "fix" this as well though, meaning we've now had at least two people who considered it a bug (or at least inconvenience), so we should at least keep an eye on this.

GraionDilach:
Although that PR also has this included, but I guess this can go without that so :+1:

reaperrr:
It may take a while until that is mergable, and this here might still make it into the playtest/release.

Mailaender:
Thanks.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a1bc3e252cb73bc5347be9a5997ff224aae4990d...5064e96a77cbf5d3d3ef0acdcdd10e44d2ccb4b2)

PR #9186, opened 2015-08-28
Combine Shrapnel into Bullet

LavenderMoon:
Not quite satisfied with this, and am looking for feedback. I'm not familiar enough with the codebase to know the best way to profile performance and memory usage and ensure that this doesn't have any negative impacts. I'm also not sure if moving the `Sync`ed properties to a parent class messes anything up, or if making anything readonly or protected risks messing with `Sync`. I also haven't been able to test it in multiplayer. Again, comments and criticism welcome.


pchote:
The `Shrapnel` projectile looks to be almost completely redundant.  IMO it would be better to move the random image support to `Bullet` and then delete it.

LavenderMoon:
That was my gut feeling, but I assumed there was some performance reason it was separate. I'll combine them real quick.

LavenderMoon:
As far as I can tell, everything is working with this.

OpenRA.Mods.Common/Effects/Bullet.cs
```diff
@@ -171,6 +171,9 @@ public void Tick(World world)
 
 		public IEnumerable<IRenderable> Render(WorldRenderer wr)
 		{
+			if (anim == null || ticks >= length)
```
pchote: This is duplicating an existing check just below.

LavenderMoon: Whoops.

OpenRA.Mods.Common/Effects/Bullet.cs
```diff
@@ -171,12 +171,12 @@ public void Tick(World world)
 
 		public IEnumerable<IRenderable> Render(WorldRenderer wr)
 		{
-			if (info.ContrailLength > 0)
-				yield return contrail;
-
```
reaperrr: I haven't tested this, but I think the idea here was that bullet projectiles should be able to display a contrail even if the bullet itself has no image (I think RA snipers used this).

OpenRA.Mods.Common/Effects/Bullet.cs
```diff
@@ -31,8 +31,8 @@ public class BulletInfo : IProjectileInfo
 		[Desc("Image to display.")]
 		public readonly string Image = null;
 
-		[Desc("Loop this sequence of Image while this projectile is moving.")]
-		[SequenceReference("Image")] public readonly string Sequence = "idle";
+		[Desc("Loop these sequences of Image while this projectile is moving.")]
+		[SequenceReference("Image")] public readonly string[] Sequences = { "idle" };
```
reaperrr: While it seems none of the shipping mods use anything else than 'idle' (except the former shrapnel weapons), it would be better if there was an upgrade rule for this property rename anyway, for 3rd-party mods.

reaperrr:
Could you squash the commits together? On their own they would look odd in the commit history.

OpenRA.Mods.Common/Effects/Bullet.cs
LavenderMoon: Good point, will fix.

OpenRA.Mods.Common/Effects/Bullet.cs
LavenderMoon: Yeah, I didn't have any test cases that broke, so I figured that was handled automatically by reflection. I'll write an upgrade rule.

LavenderMoon:
I'll make the suggested changes and squash the commits together.

OpenRA.Mods.Common/Effects/Bullet.cs
GraionDilach: Disagree. I'd say this is unneeded. Shrapnel IIRC haven't hit a playtest.

pchote: This is a change to Bullet, not Shrapnel.

GraionDilach: Whoops, indeed. I'll get my dunce cap. The upgrade rule does seem justified then.

LavenderMoon: Good to note, though. I was going to make it upgrade `Projectile: Shrapnel` to `Projectile: Bullet`, but if it hasn't reached playtest, the only users would likely be the official OpenRA mods.

LavenderMoon:
Alright, changes made, upgrade rules added, and commits squashed.

GraionDilach:
:+1:

LavenderMoon:
Finally figured out how to rename a pull request.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2249,6 +2249,14 @@ internal static void UpgradeWeaponRules(int engineVersion, ref List<MiniYamlNode
 					}
 				}
 
+				if (engineVersion < 20150828)
+				{
+					if (depth == 2 && parentKey == "Projectile" && !node.Key.StartsWith("Sequences") && node.Key.StartsWith("Sequence"))
```
Mailaender: This does not check for `Bullet` as a value so it will break `Missile` and all the other projectiles which still use `Sequence`.

LavenderMoon: Ah, didn't realize there were projectiles other than `Bullet` and `Shrapnel`. Glad you caught that.

LavenderMoon:
Alright, fixed the bug where it also effected `Missile`s.

Mailaender:
Can you please run it on the mods in this repository?

LavenderMoon:
Alright. Going off how that removes all comments from the files, I assume this is just a sanity check?

Mailaender:
Don't commit those unrelated changes.

penev92:
You need to be careful what you commit when running the utility, as it has the nasty habit of changing **everything**, even if you don't want it to.

LavenderMoon:
Those are the only changes, unless they also effect the .oramaps.

GraionDilach:
You have to stage **only** the changes your upgrade rule explicitly did, (the actual Shrapnel->Bullet renamings), not filling your PR with dummy changes at all maps, removed comments etc.

Use your tool's diff viewer during commiting, and not just stage everything changed.

penev92:
Packed maps (*.oramap) don't contain custom rules, which means the utility isn't actually changing them. You can just skip those when checking the diff and deciding what to commit.

LavenderMoon:
I thought Mailaender meant something else because I'd already done that.

LavenderMoon:
GraionDilach, my upgrade rule only changes the `Bullet`'s `Sequence` key to `Sequences` because it was determined earlier that there's not likely any mods using `Shrapnel` outside of this repo. Should I just go ahead and add that rule, to be extra safe?

GraionDilach:
Oh, no sorry, I'm in a middle of a heat wave, the rule itself doesn't need to be changed.

LavenderMoon:
No worries. :)

Mailaender:
I also didn't really notice that no one really uses the bullet sequence(s). :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3be2f3d5d17a5000be35800d05ebef61f925f34e...f793edd6e31d5379bbf86a60573860d05f2c3e97)

PR #9187, opened 2015-08-28
Fix culture in FieldSaver for collections

RoosterDragon:
Ensure arrays and sets format all elements in their collections. This ensures they use the invariant culture and other formatting correctly.

Currently, the formatting of a type depends on if it is in a collection or not, which is not ideal.


Mailaender:
Floating point values still get properly saved into settings.yaml on a German machine. :white_check_mark:

penev92:
:+1:

PR #9188, opened 2015-08-28
Friendly name for HashSet in docs

RoosterDragon:
Requested in https://github.com/OpenRA/OpenRA/pull/9181#issuecomment-135579391.


Mailaender:
:+1:

Phrohdoh:
Why do we need separate names for a set vs array, etc?

penev92:
I guess it's a compromise between giving technical information like `HashSet` vs `string[]` and not being technical enough (people might like to know why their YAML isn't working when it is due to `HashSet` swallowing the duplicate values).

Phrohdoh:
Do we have any situations where duplicate values *do* work?

PR #9191, opened 2015-08-29
Fix TS regression from #9004

reaperrr:
No idea how this passed the lint checks...


Mailaender:
I tested TS infantry manually and actually Travis did complain, but only with warnings not errors. https://travis-ci.org/OpenRA/OpenRA/builds/76903828

Mailaender:
I thought I had already replaced warnings with errors, but apparently I didn't. Filed https://github.com/OpenRA/OpenRA/pull/9194.

PR #9193, opened 2015-08-29
Fixed BodyOrientation refactor regressions

Mailaender:
Another followup of https://github.com/OpenRA/OpenRA/pull/9004 which fixes problems discovered by Coverity and myself.

### DIFF
--- a/OpenRA.Mods.Common/Traits/BodyOrientation.cs
+++ b/OpenRA.Mods.Common/Traits/BodyOrientation.cs
@@ -73,12 +73,11 @@ namespace OpenRA.Mods.Common.Traits
 					return info.QuantizedFacings;
 
 				var qboi = self.Info.Traits.GetOrDefault<IQuantizeBodyOrientationInfo>();
-				var isb = self.HasTrait<ISpriteBody>();
 
 				// If a sprite actor has neither custom QuantizedFacings nor a trait implementing IQuantizeBodyOrientationInfo, throw
 				if (qboi == null)
 				{
-					if (isb)
+					if (self.HasTrait<ISpriteBody>())
 						throw new InvalidOperationException("Actor" + self.Info.Name + "has a sprite body but no facing quantization."
 							+ " Either add the QuantizeFacingsFromSequence trait or set custom QuantizedFacings on BodyOrientation.");
 					else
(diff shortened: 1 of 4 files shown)


RoosterDragon:
I get "Actor type 'camera.paradrop' does not define a quantized body orientation." when using paratroopers with the PR, just like I got with #8839. That's why I removed the sync check in mine.

Mailaender:
Fixed.

OpenRA.Mods.Common/Traits/BodyOrientation.cs
```diff
@@ -76,9 +76,14 @@ public BodyOrientation(ActorInitializer init, BodyOrientationInfo info)
 				var isb = self.HasTrait<ISpriteBody>();
```
RoosterDragon: Looks like you can move this after the ```qboi == null``` check and avoid the lookup most of the time.

Mailaender: Good idea. Done.

penev92:
:+1:

OpenRA.Mods.Common/Traits/BodyOrientation.cs
```diff
@@ -73,12 +73,16 @@ public BodyOrientation(ActorInitializer init, BodyOrientationInfo info)
...
-				var isb = self.HasTrait<ISpriteBody>();
 
 				// If a sprite actor has neither custom QuantizedFacings nor a trait implementing IQuantizeBodyOrientationInfo, throw
-				if (qboi == null && isb)
-					throw new InvalidOperationException("Actor" + self.Info.Name + "has a sprite body but no facing quantization."
-						+ " Either add the QuantizeFacingsFromSequence trait or set custom QuantizedFacings on BodyOrientation.");
+				if (qboi == null)
+				{
+					if (self.HasTrait<ISpriteBody>())
+						throw new InvalidOperationException("Actor" + self.Info.Name + "has a sprite body but no facing quantization."
```
penev92: Please change the beginning to 
>("Actor " + self.Info.Name + "`` has a ...

for easier readability. Right now this results in "Actorlpst2has a sprite body".

Mailaender: Right. Done.

reaperrr:
:+1:

PR #9194, opened 2015-08-29
Fixed undefined sequence lint checks not emitting errors

Mailaender:
These problems are always serious so by definition they should be errors not optional warnings. It will also exit with 1 then, which will notify us via CI build checks earlier on.


GraionDilach:
:+1:

penev92:
:+1:

PR #9195, opened 2015-08-29
Fixed a resource leak in MapPreview

Mailaender:
This was detected by Coverity.

### DIFF
--- a/OpenRA.Game/Map/MapPreview.cs
+++ b/OpenRA.Game/Map/MapPreview.cs
@@ -210,20 +210,22 @@ namespace OpenRA
 				// Request the filename from the server
 				// Run in a worker thread to avoid network delays
 				var mapUrl = Game.Settings.Game.MapRepository + Uid;
+				var mapPath = string.Empty;
 				try
 				{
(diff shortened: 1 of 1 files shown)


OpenRA.Game/Map/MapPreview.cs
```diff
@@ -253,6 +253,7 @@ public void Install()
 				{
 					Console.WriteLine(e.Message);
 					Status = MapStatus.DownloadError;
+					download.Cancel();
```
penev92: An exception can potentially occur before line 250, which means that `download` can be `null` here. Please add a null check for extra safety.

RoosterDragon:
The Coverity issue is reporting [```res``` is being leaked](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Map/MapPreview.cs#L217). It just needs a ```using``` on it as far as I can tell.

I don't think the current change of cancelling the download makes sense.

Mailaender:
Updated to fix the real issue.

RoosterDragon:
:+1:

penev92:
:+1:

PR #9199, opened 2015-08-29
Fixed TD shellmap mammoth turrets drawn below body

reaperrr:
On bleed, the turrets of the mammoth tanks on the TD shellmap are drawn below the body. This is an initialization order issue, different ZOffsets don't help.

Note1: This is a band-aid, we need #8848 to fix this properly.
Note2: This was probably introduced in #9004, but we should check the release branch just in case.


deniz1a:
There are mammoth tanks on TD shellmap?

Mailaender:
This does not effect the release branch.

Issue #9183, opened 2015-08-28
Improve/refactor WithCargo


reaperrr:
... to cache passenger facings on transport enter/exit instead of `Tick` (mentioned in #9172), possibly rewrite it to use `ActorPreview` instead of `ModifyRender`.

reaperrr:
Adressed by #11426.

PR #9202, opened 2015-08-30
Improved WithCargo performance

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/9183.

Also moved it to Mods.Common as this can be used everywhere.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithCargo.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithCargo.cs
@@ -12,10 +12,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Graphics;
-using OpenRA.Mods.Common.Traits;
 using OpenRA.Traits;
 
-namespace OpenRA.Mods.Cnc.Traits
+namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Renders the cargo loaded into the unit.")]
 	public class WithCargoInfo : ITraitInfo, Requires<CargoInfo>, Requires<IBodyOrientationInfo>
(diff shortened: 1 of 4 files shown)


OpenRA.Mods.Common/Traits/Render/WithCargo.cs
```diff
@@ -29,13 +28,16 @@ public class WithCargoInfo : ITraitInfo, Requires<CargoInfo>, Requires<IBodyOrie
...
-	public class WithCargo : IRenderModifier, ITick
+	public class WithCargo : IRenderModifier, ITick, INotifyPassengerEntered, INotifyPassengerExited
 	{
 		readonly Cargo cargo;
 		readonly IFacing facing;
 		readonly WithCargoInfo cargoInfo;
 		readonly IBodyOrientation body;
 
+		List<IFacing> passengerFacings;
+		Dictionary<Actor, PassengerInfo> passengerInfos;
```
RoosterDragon: ```readonly``` please.

reaperrr:
Needs rebase.

Mailaender:
Rebased and updated.

RoosterDragon:
:+1: Looks good to me.

reaperrr:
Needs rebase.

Mailaender:
Rebased.

penev92:
This causes some very strange passenger behavior when unloading, which is quite visible on the first GDI mission.

reaperrr:
Also needs rebase now.

Mailaender:
Sadly this regresses hilariously and I am not sure why it happens.

obrakmann:
The reason for the regression is that the cargo is added twice to the `passengerFacings` (and `passengerInfos`, probably) list. If you create a transport actor and then immediately load cargo into it, `PassengerEntered` is called twice, once from `Cargo.Tick` (because `initialized` is still false), and once from `Cargo.Load`. Setting `initialized` to true in `Cargo.Load` would be one way to fix that, another more correct way would be to use `Cargo.Load` in the constructor (otherwise potential upgrades provided by the cargo wouldn't be applied).

pchote:
passengerFacings probably wants to be a set rather than a list, independent of the other fixes.

Mailaender:
Doesn't really matter anyway as `WithCargo` isn't used extensively.

PR #9203, opened 2015-08-30
Added tunnel top rendering

Mailaender:
Before:
image

After:
image

Only polishes the rendering. Does not add support for http://modenc.renegadeprojects.com/Tubes.

### DIFF
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -694,7 +694,7 @@
 		Pieces: 3, 7
 		Range: 2c0, 5c0
 
-^Railway:
+^TerrainOverlay:
 	AlwaysVisible:
 	Immobile:
 		OccupiesSpace: False
@@ -704,3 +704,11 @@
 	BodyOrientation:
 		UseClassicPerspectiveFudge: False
 		QuantizedFacings: 1
+
+^Railway:
+	Inherits: ^TerrainOverlay
+
+^Tunnel:
+	Inherits: ^TerrainOverlay
+	CustomSelectionSize:
+		CustomBounds: 220,220
--- a/mods/ts/maps/arivruns/map.yaml
+++ b/mods/ts/maps/arivruns/map.yaml
@@ -1286,6 +1286,18 @@ Actors:
 	Actor422: trock02
 		Location: 147,106
 		Owner: Neutral
+	Actor423: tuntop04
+		Owner: Neutral
+		Location: 145,5
+	Actor424: tuntop04
+		Owner: Neutral
+		Location: 116,9
+	Actor425: tuntop02
+		Owner: Neutral
+		Location: 144,-19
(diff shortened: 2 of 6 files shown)


mods/ts/rules/defaults.yaml
```diff
@@ -609,6 +609,17 @@
 	Tooltip:
 		Name: Palette
 
+^Tunnel:
```
GraionDilach: Waterfalls could also use this (booth are pretty much the same in WW logics), so `^Tunnel` might not be the best option.

Mailaender: Identified the sprites as `WA01X` etc. but can't get them o work as easily by just pasting them:

image

It results into a pixelish mess that probably has to be resolved differently.

Also in the future tunnel related logic traits might be attached to this.

GraionDilach: Alright.

mods/ts/rules/defaults.yaml
```diff
@@ -609,6 +609,17 @@
...
+	AlwaysVisible:
+	Immobile:
```
pchote: This leads to fun rendering glitches when the actor's CenterPosition moves outside the viewport:

tunnel

You'll need to make these Buildings instead.

pchote: You can then also make the inner-most row of cells impassible, which will stop units from pathing through the tunnel to the ground above.

Mailaender: I can't as buildings are heightmap aware and occupy space so they can't be placed on cliff edges:

image

image

Mailaender:
The viewport rendering problem of actors spanning multiple cells on isometric terrain is unrelated to this.

pchote:
Regular buildings ingame do not show this problem, so :-1: to removing the tag and merging a broken feature.

Mailaender:
I probably need to rethink this then. Might need an `TerrainAnimationLayer` and do it like the original.

pchote:
There is a simpler fix: Add the `CustomSelectionSize` trait with appropriate bounds to the tunnel actors ( `AutoSelectionSize` doesn't generate a large enough rectangle, so there is still flickering).

> The viewport rendering problem of actors spanning multiple cells on isometric terrain is unrelated to this.

It sounds like your other recent decoration PRs have also missed these important traits.

Mailaender:
When we last tested https://github.com/OpenRA/OpenRA/pull/9130 with the moving viewport the BodyOrientation and sprite rendering refactor that backfired a lot wasn't in so things may have changed again.

Mailaender:
Tweaked the tunnel tops disappearing too early by introducing `CustomSelectionSize` bounds.

Mailaender:
Rebased.

abcdefg30:

reaperrr:
:+1:

PR #9189, opened 2015-08-28
Remove RenderBuilding

reaperrr:
Replaces the four traits inheriting `RenderBuilding` with solutions based on `WithSpriteBody`.
Replaces `RenderBuilding` with `WithSpriteBody` + `RenderSprites` + `AutoSelectionSize`.

Supercedes #8753.

PR #9204, opened 2015-08-30
Fix TD tower selection boxes and improve TS sequence offsets

reaperrr:
- Fixes that the Obelisk and AGT in TD only have 24x24 visual selection boxes.
- Improves offsets of TS Adv. Nod Power Plant and GDI Comm. Center. This does NOT fix the offset issues on "A River Runs Near It", but makes them more accurate everywhere else.

### DIFF
--- a/mods/cnc/rules/structures.yaml
+++ b/mods/cnc/rules/structures.yaml
@@ -720,6 +720,8 @@ OBLI:
 		Dimensions: 1,2
 	Selectable:
 		Bounds: 24,24,0,12
+	SelectionDecorations:
+		VisualBounds: 22,42
 	RequiresPower:
 	DisabledOverlay:
 	Health:
@@ -799,6 +801,8 @@ ATWR:
 		Dimensions: 1,2
 	Selectable:
 		Bounds: 24,24,0,12
+	SelectionDecorations:
+		VisualBounds: 22,40,0,4
 	RequiresPower:
 	DisabledOverlay:
 	Health:
--- a/mods/ts/sequences/structures.yaml
+++ b/mods/ts/sequences/structures.yaml
@@ -232,7 +232,7 @@ napowr:
 
 naapwr:
 	Defaults:
-		Offset: 12, -30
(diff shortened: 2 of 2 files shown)


Mailaender:
:+1:

Mailaender:
Regarding https://github.com/OpenRA/OpenRA/pull/9189#issuecomment-136099529 I guess there is still code which isn't really height-map aware.

pchote:
The upgrade center / advanced power plants have nothing to do with heightmaps, and only occur on one theater each.  I suspect that the tileset-specific artwork might have bogus offsets defined in the artwork.

GraionDilach:
Disagree. I've put a day to fix two buildings having that same footprint in my mod, and in both cases I
 - either managed to have the cell position markers synced with the art, with both floating in air like 75 pixels above they should
 - either managed to make the building placed in it's actual position so-so but then the placement position cell art went off

I can't think anything else besides that footprint being bogus for some reason.

pchote:
@GraionDilach: Do you have screenshots of those issues?  I don't see how your description disagrees with my comment (if that's what it was aimed at).

GraionDilach:
Because if it'd be the tileset-specific art, then it would happen with all buildings. It also affects my powerplant and barracks on these pictures which has one common thing next to the affected ones - they are all 2x3 buildings. Which is why I think this specific foundation being bugged.

openra-2015-08-30t224723z
openra-2015-08-30t225235z

pchote:
@GraionDilach: the problem there is that the buildings are rendered fine in the snow tileset.  The only thing that is different between snow and temperate is the specific building sprite file.

reaperrr:
@pchote: No, the offset issues do not appear on both temperate maps, only on "A river runs near it", not on "River Raid".

I just now changed the normal Nod power plant to a 2x3 footprint without any other changes and indeed, offset is fine on "River Raid", wrong on "A river blah".

So it's definitely not an art issue.

penev92:
:+1:

PR #9205, opened 2015-08-30
Added the Tiberium active overlay animation

Mailaender:
blowing spores in the air.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithActiveAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithActiveAnimation.cs
@@ -21,6 +21,7 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Sequence name to use")]
 		[SequenceReference] public readonly string Sequence = "active";
 
+		[Desc("Measured in ticks.")]
 		public readonly int Interval = 750;
 
 		public readonly bool PauseOnLowPower = false;
--- a/OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
@@ -36,6 +36,9 @@ namespace OpenRA.Mods.Common.Traits
 
 		public readonly bool PauseOnLowPower = false;
 
+		[Desc("Set this to pause the sequence playback periodically. Measured in ticks.")]
+		public readonly int Interval = 0;
+
 		public override object Create(ActorInitializer init) { return new WithIdleOverlay(init.Self, this); }
 
 		public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer init, RenderSpritesInfo rs, string image, int facings, PaletteReference p)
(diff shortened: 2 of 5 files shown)


reaperrr:
In both TD and TS, this animation actually only played when the tib tree actually tried to spawn tiberium, it wasn't just independent cosmetics.
Not a blocker, but something to consider.

GraionDilach:
Why does this need an additional trait? Just asking of the difference.

reaperrr:
Animations replace the body, overlays are added on top.

GraionDilach:
Ah, right.

pchote:
I agree with both @reaperrr and @GraionDilach.  The Interval property could be added to `WithIdleOverlay` to avoid unnecessary duplication.

Mailaender:
They were talking about `WithActiveAnimation` and `WithActiveOverlay` though.

reaperrr:
Yes, but generally speaking, the most logical solution would be to add `Interval` to `Idle` anims/overlays (defaulting to 0, of course) so it's usable by any kind of actor, and introduce `WithResourceSpawnAnimation`/-`Overlay` triggered whenever the actor spawns a resource, with a minimum delay property to avoid anims playing every 55 ticks.

Mailaender:
I don't think this kind of feature creep demand is fair on this pull request.

pchote:
Its not really a feature creep, as far as I can see.  There have been two major points, which are largely separate but both apply to your original implementation instead of asking for new things:

* Adding a generic `WithActiveOverlay` is redundant, as this duplicates the existing `WithIdleOverlay` trait.  The only difference is the `Interval` property, which can be added to `WithActiveAnimation` instead.
* In any case, its not correct to use either `WithActiveOverlay` or `WithIdleOverlay` for tiberium trees, because they should only animate when spawning.  To make this work as it should you would need to introduce a new interface, and then rename your generic `WithActiveOverlay` to a more specific `WithResourceSeedingOverlay` that implements the interface.

penev92:
This hasn't seen any activity in almost a month and a half.
I don't know if `WithActiveOverlay` would be useful or not, but I do agree with the `WithResourceSeedingOverlay` part.
If this is a #wontfix then we should close this.

Mailaender:
This can be consolidated later. The behavioral changes are also out of scope of this pull request.

pchote:
The conservative option may be to close this for now, and then somebody else can revisit either the `WithResourceSeedingOverlay` trait or adding an `Interval` parameter to `WithIdleOverlay` in the future.

Mailaender:
I merged `WithActiveOverlay` into `WithIdleOverlay` although I don't think it is a good idea to have every idle overlay tick for performance reasons.

pchote:
It may be better to go with the `WithResourceSeedingOverlay` approach in that case.

abcdefg30:
Needs a rebase.

Mailaender:
Too many cooks...

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/commit/8ce274ed791f19e81745509afd31d0143dbbe5b1 and https://github.com/OpenRA/OpenRA/pull/11297.

PR #9207, opened 2015-08-30
Added damaged, light and water animations to the Tiberian Sun hydroelectric dam

Mailaender:
Seems to be disabled in the original, however I don't see any palette problems. https://github.com/LipkeGu/TSIniFiles/blob/master/art.ini#L1049


GraionDilach:
The palette problems in the original happens because the water uses colors above #240 and in TS those slots were unaffected by map lighting.

We have our own, finer controls over that so we are unaffected. :+1:

penev92:
Looks OK to me :+1:

Issue #8916, opened 2015-08-03
Support ground-level bridges in TS


pchote:
These are not fundamentally different to the "long bridges" in the RA mod, so it should be possible to support these after some straightforward adjustments:
- Add support for multiple / randomized template variants in the `Bridge` trait.  This will allow us compress the 28 bridge overlay types into 6 actor definitions.  It might pay to split the rendering code out into its own `WithSpriteBridgeBody` while there.
- Add support for map-placed bridges in `BridgeLayer`.

pchote:
Discussion from #10764 before it was closed:

> we actually have all the logic you need to implement low bridge destruction. The only bits we're still missing for complete low bridges are the random art selection (this could be done with a random upgrade trait switching different WithSpriteBody traits) and demolition/repair via bridge huts.

> You can implement destruction by setting up two "normal" actors (one for each bridge orientation), two "dead" husk actors, and the four ramps (set up like the normal actors). The normal actors have regular and damaged art states, and spawn the "dead" actors on death. The "dead" actors use WithWallSpriteBody to render the damaged-end overhang when they are next to or between undamaged bridge segments. The artwork for the different adjacency cases are in different files, which you assemble into a single sequence for the wall trait using an aggregate sequence.

Mailaender:
That approach will create unpassable bridges that don't render correctly: https://github.com/Mailaender/OpenRA/tree/combined-lowbridges

pchote:
That branch does not appear to implement the approach that I have outlined.

pchote:
More discussion from IRC:

```
[19:02:52]	<Mailaender> our current bridges work exactly the same way except that the actors are tilset templates for each damage state as well
[19:03:10]	<pchote> actually bridges work the way i'm asking
[19:03:38]	<Mailaender> I probably also don't get you right, but I have the impression you thought about it in theory, but didn't make any attempts to try out the approach you demand shall not be deviated from.
[19:04:46]	<pchote> the wall logic / "spawn actors at the edge upon destruction" is exactly what the Destroyed*Template stuff on Bridge is
[19:05:08]	<pchote> the only difference is that it is done using existing code and sequences instead of a pile of hardcoded logic referencing tile templates
[19:05:30]	<pchote> I know it works because I spent a long time with ^
[19:05:35]	<Mailaender> but how do you define the endings
[19:05:44]	<Mailaender> walls are 1x1 tiles not 3x1 or 1x3
[...]

PR #9211, opened 2015-08-30
Added Tiberian Sun railroad track rendering

Mailaender:
For now this is just doing the railroad track rendering. Already comes with a test case (see shellmap) and I updated https://github.com/Mailaender/OpenRA/compare/ts-map-import as well to show off that this is legacy compatible:

image

### DIFF
--- a/mods/ts/sequences/misc.yaml
+++ b/mods/ts/sequences/misc.yaml
@@ -489,3 +489,83 @@ palet03:
 palet04:
 	idle:
 
+tracks01:
+	idle:
+		UseTilesetExtension: true
+		ZOffset: -1c0
+
+tracks02:
+	idle:
+		UseTilesetExtension: true
+		ZOffset: -1c0
+
+tracks03:
+	idle:
+		UseTilesetExtension: true
+		ZOffset: -1c0
+
+tracks04:
+	idle:
+		UseTilesetExtension: true
+		ZOffset: -1c0
+
+tracks05:
+	idle:
+		UseTilesetExtension: true
+		ZOffset: -1c0
+
+tracks06:
+	idle:
+		UseTilesetExtension: true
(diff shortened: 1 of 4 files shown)


pchote:
This doesn't implement the custom terrain type that will be needed for train pathing.  I don't think that should block us from importing these as cosmetic map props, so :+0.5: (haven't tested this ingame, but I'm okay with the concept).

Having multiple custom terrain layers for different prop types kind of sucks, so I suggest we leave that until after #8916, which I expect (for the implementation I have in my head at least) will make it possible to use the same layer for both bridges and railroads.

pchote:
The z-offsets need adjusting so that these render below actors.

screen shot 2015-09-01 at 05 38 56

Mailaender:
Rebased and fixed the ZOffset.

mods/ts/sequences/misc.yaml
```diff
@@ -489,3 +489,83 @@ palet03:
 palet04:
 	idle:
 
+tracks01:
+	idle:
+		UseTilesetExtension: true
+		ZOffset: -1c511
```
pchote: Why 1.5 cells? 0.5 should be sufficient.

Mailaender: 0.5 isn't:

image

1.5 is:

image

dan9550:
Hooray, i won't have track bits mission on my shellmap after this.

penev92:
:+1:

mods/ts/sequences/misc.yaml
pchote: What about 1.0?

Mailaender: Seems to be enough. Updated.

abcdefg30:

PR #9212, opened 2015-08-30
Added the unloaded truck to Tiberian Sun

Mailaender:
Just noticed

> Ignoring unknown actor type: `trucka`

when trying to import http://ra2maps.zzattack.org/Tiberian%20Sun/Official%20Multiplayer/getmap.php?map=Tunnel+Training

### DIFF
--- a/mods/ts/rules/civilian-vehicles.yaml
+++ b/mods/ts/rules/civilian-vehicles.yaml
@@ -34,7 +34,7 @@
 	WithVoxelBarrel:
 	WithMuzzleFlash:
 
-TRUCKB:
+^TRUCK:
 	Inherits: ^CivilianVoxelVehicle
 	Valued:
 		Cost: 500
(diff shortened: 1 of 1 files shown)


GraionDilach:
I'm actually not certain about this one. In TS, there were two map scripts which changed `TRUCKA` to `TRUCKB` respectively - yes, they only did that, and even the IDs were hardcoded for that and nothing more - so it might be better if we just treat them as one actor with say, `*B` being a `WithVoxelBody` upgrade.

Although it probably needs an `UpgradeInit` trait or something then.

penev92:
Nah, this makes it much easier for mappers and modders to make use of it for stuff that are != than vanilla TS. It's also only 10 lines of YAML. That said, can't one truck inherit the other one directly, thus only adding 2 lines of YAML?

GraionDilach:
Right. We could also add a tag into `Transforms` that it can't be ordered by the players to recreate the map scripts' logic with transforming them back and forth strictly via map LUA at a later point.

Mailaender:
I guess map decoration != mission scripting. If you do unloading with transforms, upgrade voxel sequences or whatever doesn't matter right now. Made them inherit the same MiniYaml though.

LipkeGu:
you pushed this to the upstream o.o"

Mailaender:
Sorry, I must have misclicked.

penev92:
:+1:

PR #9214, opened 2015-08-31
Fixed the TS mod crashing on startup and oh trains!

Mailaender:
:bullettrain_side:


pchote:
I like trains. :+1:

LipkeGu:
:+1:

Phrohdoh:
Can you give us a test commit?
Or should this depend on #9211?

Mailaender:
Use the map editor to place the units into a map. There is nothing else to test yet.

PR #9215, opened 2015-09-01
 Move NAT Logic from ClientTooltipLogic into its own Function at LobbyUtils.

### LINKED ISSUE #8787: UPNP ClientTooltip NRE
I expect `UPnP.NatDevice.GetExternalIP()` is returning null.

```
Inner Exception of type `System.NullReferenceException`: Object reference not set to an instance of an object
    at OpenRA.Mods.Common.Widgets.Logic.ClientTooltipLogic..ctor (OpenRA.Widgets.Widget widget, OpenRA.Mods.Common.Widgets.TooltipContainerWidget tooltipContainer, OpenRA.Network.OrderManager orderManager, Int32 clientIndex) [0x00296] in /Users/paul/src/OpenRA/OpenRA.Mods.Common//Widgets/Logic/Lobby/ClientTooltipLogic.cs:85 
  at (wrapper managed-to-native) System.Reflection.MonoCMethod:InternalInvoke (System.Reflection.MonoCMethod,object,object[],System.Exception&)
 ```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24949022-upnp-clienttooltip-nre?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### PR DESCRIPTION
**Pull request by LipkeGu**
This PR is fixing a Crash when the Client has problems to get the IP-Adress from the NAT Devices.

fixes #8787

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/ClientTooltipLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/ClientTooltipLogic.cs
@@ -79,10 +79,8 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			var ping = orderManager.LobbyInfo.PingFromClient(client);
 			latency.GetText = () => LobbyUtils.LatencyDescription(ping);
 			latency.GetColor = () => LobbyUtils.LatencyColor(ping);
-			var address = orderManager.LobbyInfo.ClientWithIndex(clientIndex).IpAddress;
-			if (clientIndex == orderManager.LocalClient.Index && UPnP.NatDevice != null
-				&& address == IPAddress.Loopback.ToString())
-				address = UPnP.NatDevice.GetExternalIP().ToString();
+
+			var address = LobbyUtils.GetExternalIP(clientIndex, orderManager);
 			var cachedDescriptiveIP = LobbyUtils.DescriptiveIpAddress(address);
 			ip.GetText = () => cachedDescriptiveIP;
 			var cachedCountryLookup = GeoIP.LookupCountry(address);
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
@@ -482,5 +482,18 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			};
 			playerName.GetColor = () => player.Color.RGB;
 		}
+
+		public static string GetExternalIP(int clientIndex, OrderManager orderManager)
+		{
+			var address = orderManager.LobbyInfo.ClientWithIndex(clientIndex).IpAddress;
+			if (clientIndex == orderManager.LocalClient.Index && address == IPAddress.Loopback.ToString() && UPnP.NatDevice != null)
+			{
+				var externalIP = UPnP.NatDevice.GetExternalIP();
+				if (externalIP != null)
+					address = externalIP.ToString();
+			}
+
+			return address;
+		}
 	}
 }

### CODE REVIEW

File: OpenRA.Mods.Common/Widgets/Logic/Lobby/ClientTooltipLogic.cs [hunk 1]
```diff
@@ -79,10 +79,15 @@ public ClientTooltipLogic(Widget widget, TooltipContainerWidget tooltipContainer
...
 			latency.GetColor = () => LobbyUtils.LatencyColor(ping);
+
 			var address = orderManager.LobbyInfo.ClientWithIndex(clientIndex).IpAddress;
-			if (clientIndex == orderManager.LocalClient.Index && UPnP.NatDevice != null
-				&& address == IPAddress.Loopback.ToString())
-				address = UPnP.NatDevice.GetExternalIP().ToString();
+			if (clientIndex == orderManager.LocalClient.Index && address == IPAddress.Loopback.ToString())
+			{
+				var externalIP = IPAddress.None;
+				if (UPnP.NatDevice != null && IPAddress.TryParse(UPnP.NatDevice.GetExternalIP().ToString(), out externalIP))
```
Mailaender: This doesn't make much sense to me. You are converting to string and then parsing the string back into an IP address only to save it as string below.

LipkeGu: ```out externalIP``` requires ```System.net.IPAddress```

LipkeGu:
@Mailaender : The complete NAT logic in this TooltipLogic has no much sense and should be moved to UPNP.cs :/ and make a general function from that snipped.

LipkeGu:
what has to be changed? the idea or the code?

Mailaender:
In my eyes you are just complicating things and not fixing the actual `NullReferenceException`.

Mailaender:
Feel free to move this to LobbyUtils if you want.

LipkeGu:
updated

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
```diff
@@ -482,5 +482,17 @@ public static void AddPlayerFlagAndName(ScrollItemWidget template, Player player
 			};
 			playerName.GetColor = () => player.Color.RGB;
 		}
+
+		public static string GetExternalIP(int clientIndex, OrderManager orderManager)
+		{
+			var address = orderManager.LobbyInfo.ClientWithIndex(clientIndex).IpAddress;
+			if (clientIndex == orderManager.LocalClient.Index && address == IPAddress.Loopback.ToString())
+			{
+				if (UPnP.NatDevice != null && UPnP.NatDevice.GetExternalIP().ToString() != null)
```
Mailaender: I expect I expect `UPnP.NatDevice.GetExternalIP()` is returning null and not the `ToString()` afterwards.

LipkeGu: when UPnP.NatDevice.GetExternalIP() then the string is also zero... when i check UPnP.NatDevice.GetExternalIP() for null then it will crash anyway

Mailaender: This is strange. When I call .ToString() on a null IPAddress it crashes like in https://github.com/OpenRA/OpenRA/issues/8787:

image

Mailaender: Also you are running UPnP.NatDevice.GetExternalIP() or UPnP.NatDevice.GetExternalIP().ToString() twice. Do it once. Cache the result. Check for null. Then set address and return.

LipkeGu:
updated

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
```diff
@@ -482,5 +482,17 @@ public static void AddPlayerFlagAndName(ScrollItemWidget template, Player player
 			};
 			playerName.GetColor = () => player.Color.RGB;
 		}
+
+		public static string GetExternalIP(int clientIndex, OrderManager orderManager)
+		{
+			var address = orderManager.LobbyInfo.ClientWithIndex(clientIndex).IpAddress;
+			if (clientIndex == orderManager.LocalClient.Index && address == IPAddress.Loopback.ToString())
+			{
```
Phrohdoh: Unnecessary brackets.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
```diff
@@ -482,5 +482,17 @@ public static void AddPlayerFlagAndName(ScrollItemWidget template, Player player
...
+				if (UPnP.NatDevice != null && UPnP.NatDevice.GetExternalIP() != null)
+					address = UPnP.NatDevice.GetExternalIP().ToString();
```
Mailaender: You are still doing the `UPnP.NatDevice.GetExternalIP()` lookup twice.

LipkeGu:
Can anyone restart the travis part?

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyUtils.cs
```diff
@@ -8,16 +8,16 @@
...
-using System.Collections.Generic;
-using System.Drawing;
-using System.Linq;
-using System.Net;
 using OpenRA.Graphics;
 using OpenRA.Network;
 using OpenRA.Primitives;
 using OpenRA.Traits;
 using OpenRA.Widgets;
+using System;
```
Mailaender: This will trigger a StyleCop error.

LipkeGu: Productivity power tools has messed up this part i will fix this :(

LipkeGu:
updated & rebased without PPT

Mailaender:
:+1:

abcdefg30:
Looks good to me. :+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/4d4a2414a917a7982e0624f3f74bc4663f096f4e...c6259c6cecd9ddd67793f93c06e139f7090b405e)

PR #9217, opened 2015-09-01
Ensure that we have a working OpenGL Context.

LipkeGu:
This PR adds an missing error check to the SDL2/OpenGL context creation. When something goes wrong with SDL and OpenGL it will crash...

```
Platform is Windows
Using SDL 2 with OpenGL renderer
Desktop resolution: 1920x1080
No custom resolution provided, using desktop resolution
Using resolution: 1920x1080
Detected OpenGL version: 4.5
```

### DIFF
--- a/OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
+++ b/OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
@@ -75,7 +75,9 @@ namespace OpenRA.Platforms.Default
 			}
 
 			context = SDL.SDL_GL_CreateContext(window);
-			SDL.SDL_GL_MakeCurrent(window, context);
+			if (context == IntPtr.Zero || SDL.SDL_GL_MakeCurrent(window, context) < 0)
+				throw new InvalidOperationException("Can not create OpenGL context. (Error: {0})".F(SDL.SDL_GetError()));
+
 			GraphicsContext.CurrentContext = context;
 
 			GL.LoadAll();


pchote:
I wasn't aware that we had any issues with context creation.  What is this solving?

LipkeGu:
#7622 it was your suggestion to implement this

pchote:
Checking the context alone should be sufficient.

LipkeGu:
What is with the other things for example Doublebuffer and DEPTH_SIZE (later) something can go wrong for example you have 16 bit @1024x768... and you will try to set 8 for each color mode

pchote:
It seems very unlikely to me that a machine that otherwise supports OpenRA would die on those.

LipkeGu:
updated

OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
```diff
@@ -75,7 +75,9 @@ public Sdl2GraphicsDevice(Size windowSize, WindowMode windowMode)
 			}
 
 			context = SDL.SDL_GL_CreateContext(window);
-			SDL.SDL_GL_MakeCurrent(window, context);
+			if (context == IntPtr.Zero || SDL.SDL_GL_MakeCurrent(window, context) < 0)
+				throw new InvalidOperationException("Can not create OpenGL context. (Error: {0}".F(SDL.SDL_GetError()));
```
Phrohdoh: This will print out as `(Error: foobarbaz` without a closing parenthesis.

LipkeGu:
updated

OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
```diff
@@ -75,7 +75,9 @@ public Sdl2GraphicsDevice(Size windowSize, WindowMode windowMode)
 			}
 
 			context = SDL.SDL_GL_CreateContext(window);
-			SDL.SDL_GL_MakeCurrent(window, context);
+			if (context == IntPtr.Zero || SDL.SDL_GL_MakeCurrent(window, context) < 0)
```
Unit158: Does SDL.SDL_GL_MakeCurrent return an unsigned integer? If not, should this be != 0?

Mailaender:
Still runs fine for me :white_check_mark: and this is the way to do such SDL error checks according to https://wiki.libsdl.org/SDL_GetError so :+1: probably doesn't hurt and we don't have enough information on the strange context creation failures of some Windows users with very modern hardware.

RoosterDragon:
:+1: - Just needs the commit message amending as Mailaender pointed out.

obrakmann:
Re-submitted as #9474.

PR #9218, opened 2015-09-01
Red Alert: Adds satellite icons to badger planes.

deniz1a:
They drop crates, paratroopers and parabombs. This is split from #9096.

### DIFF
--- a/mods/ra/rules/aircraft.yaml
+++ b/mods/ra/rules/aircraft.yaml
@@ -32,7 +32,6 @@ BADR:
 		Offset: -432,-560,0
 		Interval: 2
 	-EjectOnDeath:
-	-GpsDot:
 	RejectsOrders:
 
 BADR.Bomber:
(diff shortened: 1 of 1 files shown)


Mailaender:
Seems like a good compromise. :+1:

GraionDilach:
Indeed. :+1:

penev92:
I'd say this renders Soviet support powers completely useless.

GraionDilach:
Care to elaborate how?

It only shows that a plane is coming. It does not tell where it will drop it's content. In RA2 MP Allies had a Spy Sat which revealed all shroud - it still made America's Paradrop among the most useful things in the game.

Mailaender:
Nuke and iron curtain are unaffected by this and as the badgers are slow you often can see them coming in advance via revealed fog if you pay attention. However the balancing chance might be worth discussing once we see how this plays out in practice.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/7eb0241836c22689d1fb82ef1a9fc14a70677348...17d72c9af83a76ba296a30e642d73fbe89f94298)

PR #9219, opened 2015-09-01
Adds IncludeAllies option to hide-map crate.

deniz1a:
1) Adds an IncludeAllies variable to HideMapCrateAction.
2) Doesn't set the chance of getting hide-map crate as zero when player has fog visibility. Instead, player can get it but it has no effect. This works better with maps that have custom selection shares.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Crates/HideMapCrateAction.cs
+++ b/OpenRA.Mods.Common/Traits/Crates/HideMapCrateAction.cs
@@ -13,17 +13,25 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Hides the entire map in shroud.")]
 	class HideMapCrateActionInfo : CrateActionInfo
 	{
+		[Desc("Should the map also be hidden for the allies of the collector's owner?")]
+		public readonly bool IncludeAllies = false;
+
 		public override object Create(ActorInitializer init) { return new HideMapCrateAction(init.Self, this); }
 	}
 
 	class HideMapCrateAction : CrateAction
 	{
+		readonly HideMapCrateActionInfo info;
+
 		public HideMapCrateAction(Actor self, HideMapCrateActionInfo info)
-			: base(self, info) { }
+			: base(self, info)
+		{
+			this.info = info;
+		}
 
 		public override int GetSelectionShares(Actor collector)
 		{
-			// don't ever hide the map for people who have GPS.
+			// Don't ever hide the map for people who have GPS.
 			if (collector.Owner.HasFogVisibility)
 				return 0;
 
@@ -32,7 +40,15 @@ namespace OpenRA.Mods.Common.Traits
 
 		public override void Activate(Actor collector)
 		{
-			collector.Owner.Shroud.ResetExploration();
+			if (info.IncludeAllies)
+			{
+				foreach (var player in collector.World.Players)
+					if (player.IsAlliedWith(collector.Owner))
+						player.Shroud.ResetExploration();
+            }
+			else
+				collector.Owner.Shroud.ResetExploration();
+
 			base.Activate(collector);
 		}
 	}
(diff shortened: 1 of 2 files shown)


penev92:
2) sounds like it reverts some polish.

deniz1a:
What do you mean?

deniz1a:
Actually, other crates change selection shares too, depending on collector unit or player. So I don't know if the second commit is that good.

Should all crates be changed so that selection shares aren't modified and you can get crates that are useless?

pchote:
No.  The hide map cate shouldnt either.

deniz1a:
Ok.

deniz1a:
Removed the second commit.

OpenRA.Mods.Common/Traits/Crates/HideMapCrateAction.cs
```diff
@@ -13,13 +13,21 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Hides the entire map in shroud.")]
 	class HideMapCrateActionInfo : CrateActionInfo
 	{
+		[Desc("Should the map also be hidden for the allies of the collector's owner.")]
```
Phrohdoh: This is a question not a statement.

OpenRA.Mods.Common/Traits/Crates/HideMapCrateAction.cs
```diff
@@ -13,13 +13,21 @@ namespace OpenRA.Mods.Common.Traits
...
 
 	class HideMapCrateAction : CrateAction
 	{
+		readonly HideMapCrateActionInfo info;
+
 		public HideMapCrateAction(Actor self, HideMapCrateActionInfo info)
-			: base(self, info) { }
+			: base(self, info)
+		{
+			this.info = info;
```
Unit158: D:

deniz1a: ¯\_(ツ)_/¯

Phrohdoh: Use `Info`?

OpenRA.Mods.Common/Traits/Crates/HideMapCrateAction.cs
penev92: End it with a question mark.

OpenRA.Mods.Common/Traits/Crates/HideMapCrateAction.cs
deniz1a: So I should change all *CrateAction.cs files? It's like this in all of them.

penev92: Leave it. If you use `Info` you'd likely need to cast it to use it, and that makes it not worth the trouble.

deniz1a:
This can be tested with http://resource.openra.net/maps/4446

abcdefg30:

pchote:
I think it would be better (more consistent with other traits/weapons/etc, and more configurable) to have this use a Stances field instead of hardcoding a bool check for allies.

deniz1a:
But RevealMapCrateAction also has a check for allies.

penev92:
We are trying to move away from such allies checks and toward using valid stances. See #9382 for example.

GraionDilach:
Stances might not work here - we don't have an Owner stance.

deniz1a:
So how should I do this?

deniz1a:
@GraionDilach: I think this could be done by checking stances between **player** and **collector.Owner**.

penev92:
OK stances don't seem to be feasable here as this needs more fine-grained control.
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/a5e78939162b803503972262c39a6a5d6279cb45...4b93e7fdf229588fd7c18a1f45de7485db7223fc)

PR #9220, opened 2015-09-01
Add customizable properties to observer clocks.

GraionDilach:
I need help here though - the properties ain't read from the yaml (neither does the `Icon*`s, checked).

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ObserverProductionIconsWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ObserverProductionIconsWidget.cs
@@ -29,6 +29,10 @@ namespace OpenRA.Mods.Common.Widgets
 		public int IconHeight = 24;
 		public int IconSpacing = 8;
 
+		public string ClockAnimation = "clock";
+		public string ClockSequence = "idle";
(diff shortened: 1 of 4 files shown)


RoosterDragon:
The protected constructors need to copy across these properties (i.e. add all the missing ```Blah = other.Blah``` lines) - then it'll work. Might as well fix it for the icon bits too since you're here.

I have tested with the above fix, and it works fine so :+1: once applied.

GraionDilach:
Knew it would be something seriously simple, thanks! Done.

RoosterDragon:
You need to do the same thing in ```ObserverSupportPowerIconsWidget.cs```.

GraionDilach:
I'm stupid.

RoosterDragon:
Looks fine now. :)

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/318ceea6a4015e59d9f5b9cd003a450dae0946a3...6e1538a7bfc79b70770194c736f8ae39ef59b1d1)

PR #9221, opened 2015-09-01
Fix veinholes.

GraionDilach:
Another `BodyOrientation`-related regression, this time only A River Runs Near It is affected - no other maps have veins atm.


penev92:
:+1:

Mailaender:
We probably need a lint rule for this now.

Mailaender:
Thanks.

PR #9222, opened 2015-09-01
 Implement HackyAI->UnitLimits.

### LINKED ISSUE #6695: AI Build Harversters for life time
the AI never stops building harversters. that makes the game lag
and i get the fealing that the AI is not fully explored? 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5115839-ai-build-harversters-for-life-time?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #8823: Unit limit to cap production for AI bots
Already exists for buildings. Would solve https://github.com/OpenRA/OpenRA/issues/6695.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25086032-unit-limit-to-cap-production-for-ai-bots?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

GraionDilach:
Fixes #8823.
Fixes #6695.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -116,6 +116,9 @@ namespace OpenRA.Mods.Common.AI
 		[Desc("What units to the AI should build.", "What % of the total army must be this type of unit.")]
 		public readonly Dictionary<string, float> UnitsToBuild = null;
 
+		[Desc("What units should the AI have a maximum limit to train.")]
+		public readonly Dictionary<string, int> UnitLimits = null;
+
 		[Desc("What buildings to the AI should build.", "What % of the total base must be this type of building.")]
 		public readonly Dictionary<string, float> BuildingFractions = null;
 
@@ -125,7 +128,7 @@ namespace OpenRA.Mods.Common.AI
 		[Desc("Tells the AI what building types fall under the same common name.")]
 		public readonly Dictionary<string, string[]> BuildingCommonNames = null;
 
-		[Desc("What buildings should the AI have max limits n.", "What is the limit of the building.")]
+		[Desc("What buildings should the AI have a maximum limit to build.")]
 		public readonly Dictionary<string, int> BuildingLimits = null;
 
 		// TODO Update OpenRA.Utility/Command.cs#L300 to first handle lists and also read nested ones
@@ -984,7 +987,11 @@ namespace OpenRA.Mods.Common.AI
 				ChooseRandomUnitToBuild(queue) :
 				ChooseUnitToBuild(queue);
 
-			if (unit != null && Info.UnitsToBuild.Any(u => u.Key == unit.Name))
+			if (unit != null
+				&& Info.UnitsToBuild.ContainsKey(unit.Name)
+				&& (!Info.UnitLimits.ContainsKey(unit.Name)
+					|| World.Actors.Count(a => a.Info.Name == unit.Name && a.Owner == Player) < Info.UnitLimits[unit.Name]))
+
 				QueueOrder(Order.StartProduction(queue.Actor, unit.Name, 1));
 		}
 
--- a/mods/cnc/rules/ai.yaml
+++ b/mods/cnc/rules/ai.yaml
@@ -52,7 +52,7 @@ Player:
 			e2: 10%
 			e4: 7%
 			e5: 5%
-			harv: 1%
+			harv: 10%
 			bggy: 5%
 			bike: 40%
 			ltnk: 25%
@@ -65,6 +65,8 @@ Player:
 			htnk: 50%
 			heli: 5%
 			orca: 5%
+		UnitLimits:
+			harv: 8
 		SquadSize: 15
 		SupportPowerDecision@Airstrike:
 			OrderName: AirstrikePowerInfoOrder
@@ -171,7 +173,7 @@ Player:
 			e3: 30%
 			e4: 30%
 			e5: 30%
-			harv: 1%
+			harv: 10%
 			bggy: 10%
 			ftnk: 10%
 			arty: 40%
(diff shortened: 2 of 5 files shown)


OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -1005,7 +1010,10 @@ void BuildUnit(string category, bool buildRandom)
 				ChooseRandomUnitToBuild(queue) :
 				ChooseUnitToBuild(queue);
 
-			if (unit != null && Info.UnitsToBuild.Any(u => u.Key == unit.Name))
+			if (unit != null && Info.UnitsToBuild.Any(u => u.Key == unit.Name
+				&& (!Info.UnitLimits.ContainsKey(unit.Name)
+					|| World.Actors.Where(a => a.Owner == Player).Count(a => a.Info.Name == unit.Name) < Info.UnitLimits[unit.Name])))
```
penev92: This will be quite bad for performance.
Consider counting and storing the number of units per type when they are built (would need to be updated when they are destroyed, which is the tricky part).

GraionDilach: Why is this bad for perf? The LINQ is only checked if the unit exists in `UnitLimits`. In most cases, it just gets expanded with a dictionary lookup.

Caching all actors in HackyAI isn't hard per se, but that would bring in a bigger refactor to update all places where counting units are used.

penev92: Iterating over all world actors every time the AI decides to build a unit won't be fun.

Also at the very least this should be simplified to
```
if (unit != null
	&& Info.UnitsToBuild.ContainsKey(unit.Name)
	&& (!Info.UnitLimits.ContainsKey(unit.Name)
		|| World.Actors.Count(a => a.Owner == Player && a.Info.Name == unit.Name) < Info.UnitLimits[unit.Name]))

```

penev92:
Needs a rebase, and to take into accout https://github.com/penev92/OpenRA/commit/256929073c0faa8a1a373e10364f773970d97b34

GraionDilach:
Rebased.

penev92:
Of course it heavily depends on the map and the matchup, but 5 harvesters isn't that many. Especially for a turtling AI, I think (depending on the number of refs, ofc) up to 8 would be good.

GraionDilach:
Increased harvlimits to 8.

penev92:
:+1:

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -984,7 +987,11 @@ void BuildUnit(string category, bool buildRandom)
 				ChooseRandomUnitToBuild(queue) :
 				ChooseUnitToBuild(queue);
 
-			if (unit != null && Info.UnitsToBuild.Any(u => u.Key == unit.Name))
+			if (unit != null
+				&& Info.UnitsToBuild.ContainsKey(unit.Name)
+				&& (!Info.UnitLimits.ContainsKey(unit.Name)
+					|| World.Actors.Count(a => a.Info.Name == unit.Name && a.Owner == Player) < Info.UnitLimits[unit.Name]))
```
RoosterDragon: It feels like it'd be more sensible if ```ChooseRandomUnitToBuild``` and ```ChooseUnitToBuild``` took the unit limits into account (as well as ```Info.UnitsToBuild``` to be honest - might be worth seeing if you can move that too).

Mailaender: Sorry I did not see your comment while I was testing this in practice.

RoosterDragon:
Arbitrary number bikesheding which you may ignore if you like:
- I feel like the dog cap could be raised, to say 6 or 8.
- Introducing a fairly high limit on subs might be a good idea (say 12-15) to prevent an AI building runaway subs. For example in #9008 the leftover AI built runaway subs in the endgame since it had little else it could do. AI is pretty bad at subs anyway so this would let it build comfortably without being able to nuke perf by building silly amounts.

OpenRA.Mods.Common/AI/HackyAI.cs
RoosterDragon: No worries - it all works so it's not a big problem really :)

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/17d72c9af83a76ba296a30e642d73fbe89f94298...aa61fc501ff970808aed9f0c7fa9afebcb625d26)

Mailaender:
This is a small change that will great improve AI behavior tweaking. Thanks!

OpenRA.Mods.Common/AI/HackyAI.cs
GraionDilach: @penev92 suggested that HackyAI should cache the amount of it's units into a `Dictionary` pretty much, which would probably end up even more elegant - it's just massive and I didn't felt such a scopecreeping would be nice here.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -984,7 +987,11 @@ void BuildUnit(string category, bool buildRandom)
 				ChooseRandomUnitToBuild(queue) :
 				ChooseUnitToBuild(queue);
 
-			if (unit != null && Info.UnitsToBuild.Any(u => u.Key == unit.Name))
+			if (unit != null
+				&& Info.UnitsToBuild.ContainsKey(unit.Name)
+				&& (!Info.UnitLimits.ContainsKey(unit.Name)
```
Phrohdoh: This needs a null check.

https://github.com/Phrohdoh/OpenRA/issues/15#issuecomment-141710307

penev92: `// TODO: Open an issue about this.`

PR #9225, opened 2015-09-02
Fix Namespace definition

LipkeGu:
This Pr fixes a typo in a namespace definition.

### DIFF
--- a/OpenRA.Mods.RA/Scripting/Properties/InfiltrateProperties.cs
+++ b/OpenRA.Mods.RA/Scripting/Properties/InfiltrateProperties.cs
@@ -11,13 +11,14 @@
 using OpenRA.Mods.RA.Activities;
 using OpenRA.Scripting;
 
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.RA/Scripting/Properties/InfiltrateProperties.cs
```diff
@@ -11,7 +11,7 @@
 using OpenRA.Mods.RA.Activities;
 using OpenRA.Scripting;
 
-namespace OpenRA.RA.Common.Scripting
+namespace OpenRA.Mods.Common.Scripting
```
penev92: This should actually be OpenRA.Mods.RA.Scripting to match the folder.

penev92:
:+1:

GraionDilach:
:+1:

penev92:
Thanks.

PR #9230, opened 2015-09-02
Unhardcode Cursor definitions.

LipkeGu:
Unhardcode the Cursors and documentate them.

### DIFF
--- a/OpenRA.Game/Orders/UnitOrderGenerator.cs
+++ b/OpenRA.Game/Orders/UnitOrderGenerator.cs
@@ -17,6 +17,9 @@ namespace OpenRA.Orders
 {
 	class UnitOrderGenerator : IOrderGenerator
 	{
+		private const string SelectCursor = "select";
+		private const string DefaultCursor = "default";
+
 		public IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
 		{
 			var underCursor = world.ScreenMap.ActorsAt(mi)
@@ -84,7 +87,7 @@ namespace OpenRA.Orders
 
 			var cursorOrder = ordersWithCursor.MaxByOrDefault(o => o.Order.OrderPriority);
 
-			return cursorOrder != null ? cursorOrder.Cursor : (useSelect ? "select" : "default");
+			return cursorOrder != null ? cursorOrder.Cursor : (useSelect ? SelectCursor : DefaultCursor);
 		}
 
 		// Used for classic mouse orders, determines whether or not action at xy is move or select
--- a/OpenRA.Mods.Common/Orders/DeployOrderTargeter.cs
+++ b/OpenRA.Mods.Common/Orders/DeployOrderTargeter.cs
@@ -16,10 +16,12 @@ namespace OpenRA.Mods.Common.Orders
 {
 	public class DeployOrderTargeter : IOrderTargeter
 	{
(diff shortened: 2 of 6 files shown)


OpenRA.Mods.Common/Orders/DeployOrderTargeter.cs
```diff
@@ -16,10 +16,13 @@ namespace OpenRA.Mods.Common.Orders
 {
 	public class DeployOrderTargeter : IOrderTargeter
 	{
+		[Desc("The Cursor that appears when units can be deployed.")]
+		public static readonly string DeployUnitCursor = "deploy";
```
pchote: `DeployCursor` please.

penev92:
I'm all for unhardcoding stuff.
But these aren't traits, so the values won't be read from YAML anyway. You might as well make them `const`, and also `private`.
`Desc` is used for trait documentation generation, so it won't help here. Might be better to switch those to regular comments to avoid confusion. Other than that I like the idea.

pchote:
The names are self-descriptive, so the comments/descs aren't needed at all.

OpenRA.Game/Orders/UnitOrderGenerator.cs
```diff
@@ -17,6 +17,12 @@ namespace OpenRA.Orders
 {
 	class UnitOrderGenerator : IOrderGenerator
 	{
+		[Desc("The Cursor that appears when Actors can be selected.")]
+		public static readonly string SelectActorCursor = "select";
```
pchote: `SelectCursor` please.

LipkeGu:
updated

Phrohdoh:
This isn't actually unhardcoding anything. You'd need to setup default values in the constructors that you then override for this to actually make a difference to modders.

OpenRA.Game/Orders/UnitOrderGenerator.cs
```diff
@@ -17,6 +17,9 @@ namespace OpenRA.Orders
 {
 	class UnitOrderGenerator : IOrderGenerator
 	{
+		private const string SelectCursor = "select";
```
pchote: You don't need to (and shouldn't) explicitly write `private`.

Phrohdoh: This also ought not be `const` if you want it to be configurable in the way I stated earlier in a top-level comment.

LipkeGu: ok then im set them back to public

Phrohdoh: My point is that you can pass these in through the ctor (using default values) via traits that create this.

That is how you can read the values in from YAML.

LipkeGu: ah ok :)

obrakmann:
Closing due to inactivity. Also, as @Phrohdoh mentioned, this doesn't unhardcode the cursor definitions in a sense that would be useful, ie. exposing them to yaml.

Mailaender:
Yep, those should be moved to `TraitInfo` classes instead otherwise they aren't exposed to documentation. Then you could also add a `CursorReference` to them and add a `CheckCursors` lint check.

Issue #9232, opened 2015-09-02
AI harvesters don't search far away enough for new resources


Mailaender:
https://github.com/Mailaender/OpenRA/commit/2a21c7d1f2f8f10578cb71ce07147886e1dbec55 has the downside that the AI harvesters will start running circles when the next ore patch is too far away. The FindResources has to increase the search range in those cases.

RoosterDragon:
Given that:
- #9144 prevents harvesters sucking up perf searching for unreachable tiles.
- Mailaender@2a21c7d improves performance by saving a whole pathsearch.
- #9140 and #9141 will improve pathfinder perf a bit
- #9222 limits AIs from building too many harvesters

I'd say we could happily triple the search radius without too much worry of performance problems.

In fact, would it be worth removing the limit entirely? It'd be much better to pay one high pathfinder cost to reach far away ore rather than do a moderately costly search that we'll then just have to repeat all the time because it keeps failing.

pchote:
The limit was there primarily for gameplay reasons, not perf.  You don't want your harvesters running to the other side of the map if the local ore field becomes depleted.

Mailaender:
I did some tests and the harvester behavior is still to search for the nearest available resource even without resource radius restrictions. Given the many manual overrides we need to workaround problems of these range checks https://github.com/OpenRA/OpenRA/search?q=SearchFromProcRadius I vote for complete removal.

LipkeGu:
@RoosterDragon says we have to repeat that steps again when maps will be bigger ...

complete removal: :+1:

pchote:
:-1: to removing the limit, which regresses gameplay.  If your harvesters run out of local fields there is no benefit for them to travelling hundreds of cells to find a new field.  If they don't die on the way (very likely if they pick a field near an enemy base) then the time it takes for them to travel there and back would be comparable to the time it would take for the local field to regenerate.

penev92:
- I have always found it retarded when harvesters tried to harvest from the enemy base all the way across the map.
 - If you want them to, you can order them manually, which is still easier than keeping an eye of for them doing it on their own as they please and having to stop them or defend them.
 - The AI has no way of dealing with either the situation where you need to stop them and send them back to base, or the situation where you want to let them do their stuff but you need to protect them.

Mailaender:
[...]

PR #9233, opened 2015-09-02
Removed harvester search radius restrictions

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/9232.


pchote:
:-1: to removing the limit.  This regresses gameplay.

If your harvesters run out of local fields there is no benefit for them to travelling hundreds of cells to find a new field. If they don't die on the way (very likely if they pick a field near an enemy base) then the time it takes for them to travel there and back would be comparable to the time it would take for the local field to regenerate.

reaperrr:
I'm also against this. The current defaults are certainly too low, but removing the code entirely just robs us of any kind of fine-grained control.

Mailaender:
You will have to tweak this for every map as ore patch to next ore patch distances vary. This is actually what is currently being done in the Lua scripted maps. I highly advise against it.

LipkeGu:
We could override this limit via rules in the map

reaperrr:
> We could override this limit via rules in the map

@Mailaender's point is that it would be too much work to do that for each map individually.

Issue #9234, opened 2015-09-03
[TS] Game Crashed when using Ion Cannon


steadyfield:
Here is exception.log
```
Tiberian Sun Mod at Version {DEV_VERSION}
on map 1a5a76ed0d34fbbeb9b3f7909743f491f27527d2 (A River Runs Near It by Westwood Studios).
Operating System: Windows (Microsoft Windows NT 6.2.9200.0)
Runtime Version: .NET CLR 4.0.30319.34209
Exception of type `System.InvalidOperationException`: Actor type 'camera' does not define a quantized body orientation.
   at OpenRA.Mods.Common.Traits.BodyOrientation.<>c__DisplayClass3.<.ctor>b__0()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at OpenRA.Mods.Common.Traits.BodyOrientation.get_QuantizedFacings()
   at hash_BodyOrientation(Object )
   at OpenRA.Sync.CalculateSyncHash(Object obj)  d:\git\OpenRA-bleed-new\OpenRA.Game\Sync.cs:Line 33
   at OpenRA.World.SyncHash()   d:\git\OpenRA-bleed-new\OpenRA.Game\World.cs:Line 331
   at OpenRA.Network.OrderManager.Tick()  d:\git\OpenRA-bleed-new\OpenRA.Game\Network\OrderManager.cs:Line 179
   at OpenRA.Game.InnerLogicTick(OrderManager orderManager)  d:\git\OpenRA-bleed-new\OpenRA.Game\Game.cs:Line 465
   at OpenRA.Game.LogicTick()   d:\git\OpenRA-bleed-new\OpenRA.Game\Game.cs:Line 497
   at OpenRA.Game.Loop()   d:\git\OpenRA-bleed-new\OpenRA.Game\Game.cs:Line 621
   at OpenRA.Game.Run()   d:\git\OpenRA-bleed-new\OpenRA.Game\Game.cs:Line 661
   at OpenRA.Program.Run(String[] args)   d:\git\OpenRA-bleed-new\OpenRA.Game\Support\Program.cs:Line 117
   at OpenRA.Program.Main(String[] args)   d:\git\OpenRA-bleed-new\OpenRA.Game\Support\Program.cs:Line 39
```

PR #9237, opened 2015-09-03
Fixed BodyOrientation in 3rd Allies mission

Mailaender:
Fixed

> Actor type 'prison' does not define a quantized body orientation.

Another followup of #9004.

### DIFF
--- a/mods/ra/maps/allies-03a/map.yaml
+++ b/mods/ra/maps/allies-03a/map.yaml
@@ -1407,7 +1407,6 @@ Rules:
 			Type: CenterPosition
 		Immobile:
 			OccupiesSpace: false
-		BodyOrientation:
 		UpgradeActorsNear:
 			Upgrades: jail
 			Range: 1c0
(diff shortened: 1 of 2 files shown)


pchote:
Can you fix #9234 too?

The number of regressions from that pr is getting a bit ridiculous now.

Mailaender:
I thought the `Game.ModData.RulesetCache.Load(map);` lint check would automatically detect those.

reaperrr:
> The number of regressions from that pr is getting a bit ridiculous now.

In hindsight I should have introduced some kind of ^SpriteActor default that every actor default with art inherits (pretty much everything except things like bridge huts and camera actors).

I'll file a PR later today.

reaperrr:
That being said, several invisible actors incorrectly had BodyOrientation before my PR already, so in this particular case my PR just exposed something that was already wrong on bleed.

Mailaender:
> Can you fix #9234 too?

See https://github.com/OpenRA/OpenRA/pull/9241.

RoosterDragon:
:+1: (and we'll make #9241 the catch all for remaining ```BodyOrientation``` fixes)

reaperrr:
:+1:

PR #7800, opened 2015-03-30
Push actors outside of the to-place footprint, if possible.

phrohdoh:
ra
d2k
ts
ra2

Closes #3682.

PR #9239, opened 2015-09-03
Move actors outside of the desired footprint when placing a building

### LINKED ISSUE #3682: Smart Infantry Units : they could move to let the player install a building [$30]
What about infantry units that automatically move if the player tries to install a building where they are standing ?

It would just be a short move, just to let the building being placed.

<bountysource-plugin>
---
There is a **[$30 open bounty](https://www.bountysource.com/issues/792109-smart-infantry-units-they-could-move-to-let-the-player-install-a-building-15?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

penev92:
What the title says.
Spiritual successor to #7800.
Closes #3682.

Tested with 3 clients in a 2v2 match (+ 1 AI) and couldn't get anything to break or behave exploit-y.
You **cannot** tell allied units to move. This is a very basic engine design choice to prevent exploits - orders are accepted only if the client that issued them is the one that has the owner of the order's subject. See [here](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Traits/ValidateOrder.cs#L37).

### DIFF
--- a/OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
+++ b/OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
@@ -12,9 +12,12 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Graphics;
+using OpenRA.Mods.Common.Activities;
 using OpenRA.Mods.Common.Graphics;
 using OpenRA.Mods.Common.Traits;
 using OpenRA.Primitives;
+using OpenRA.Traits;
+using Util = OpenRA.Traits.Util;
 
 namespace OpenRA.Mods.Common.Orders
 {
@@ -65,7 +68,11 @@ namespace OpenRA.Mods.Common.Orders
 				world.CancelInputMode();
 
 			var ret = InnerOrder(world, xy, mi).ToArray();
-			if (ret.Length > 0)
+
+			// If there was a successful placement order
+			if (ret.Any(o => o.OrderString == "PlaceBuilding"
+				|| o.OrderString == "LineBuild"
+				|| o.OrderString == "PlacePlug"))
 				world.CancelInputMode();
 
 			return ret;
@@ -96,6 +103,9 @@ namespace OpenRA.Mods.Common.Orders
 					if (!world.CanPlaceBuilding(building, buildingInfo, topLeft, null)
 						|| !buildingInfo.IsCloseEnoughToBase(world, producer.Owner, building, topLeft))
 					{
+						foreach (var order in ClearBlockersOrders(world, topLeft))
+							yield return order;
+
 						Game.Sound.PlayNotification(world.Map.Rules, producer.Owner, "Speech", "BuildingCannotPlaceAudio", producer.Owner.Faction.InternalName);
 						yield break;
 					}
@@ -210,5 +220,28 @@ namespace OpenRA.Mods.Common.Orders
 		}
 
 		public string GetCursor(World world, CPos xy, MouseInput mi) { return "default"; }
+
+		IEnumerable<Order> ClearBlockersOrders(World world, CPos topLeft)
+		{
+			var allTiles = FootprintUtils.Tiles(world.Map.Rules, building, buildingInfo, topLeft).ToArray();
+			var neightborTiles = Util.ExpandFootprint(allTiles, true).Except(allTiles)
+				.Where(world.Map.Contains).ToList();
+
+			var blockers = allTiles.SelectMany(world.ActorMap.GetUnitsAt)
+				.Where(a => a.Owner == producer.Owner && a.IsIdle)
+				.Select(a => new TraitPair<Mobile> { Actor = a, Trait = a.TraitOrDefault<Mobile>() });
+
+			foreach (var blocker in blockers.Where(x => x.Trait != null))
+			{
+				var availableCells = neightborTiles.Where(t => blocker.Trait.CanEnterCell(t)).ToList();
+				if (availableCells.Count == 0)
+					continue;
+
+				yield return new Order("Move", blocker.Actor, false)
+				{
+					TargetLocation = blocker.Actor.ClosestCell(availableCells)
+				};
+			}
+		}
 	}
 }
--- a/OpenRA.Mods.Common/ActorExts.cs
+++ b/OpenRA.Mods.Common/ActorExts.cs
@@ -118,7 +118,8 @@ namespace OpenRA.Mods.Common
 			NotifyBlocker(self, positions.SelectMany(p => self.World.ActorMap.GetUnitsAt(p)));
 		}
 
-		public static bool CanHarvestAt(this Actor self, CPos pos, ResourceLayer resLayer, HarvesterInfo harvInfo, ResourceClaimLayer territory)
+		public static bool CanHarvestAt(this Actor self, CPos pos, ResourceLayer resLayer, HarvesterInfo harvInfo,
+			ResourceClaimLayer territory)
 		{
 			var resType = resLayer.GetResource(pos);
 			if (resType == null)
@@ -137,6 +138,11 @@ namespace OpenRA.Mods.Common
 			}
 
 			return true;
+		}
+
+		public static CPos ClosestCell(this Actor self, IEnumerable<CPos> cells)
+		{
+			return cells.MinByOrDefault(c => (self.Location - c).LengthSquared);
 		}
 	}
 }


OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
```diff
@@ -65,7 +67,11 @@ public IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
 				world.CancelInputMode();
 
 			var ret = InnerOrder(world, xy, mi).ToArray();
-			if (ret.Length > 0)
+
+			// If there was a successful placement order
+			if (ret.Any(o => o.OrderString == "PlaceBuilding"
+				|| o.OrderString == "LineBuild"
```
Phrohdoh: Personal preference (and I'm pretty sure we do this elsewhere in the code): can you move the `||`s to the end of the preceding line?

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
```diff
@@ -210,5 +219,25 @@ public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world)
 		}
 
 		public string GetCursor(World world, CPos xy, MouseInput mi) { return "default"; }
+
+		IEnumerable<Order> TryClearBlockers(World world, CPos topLeft)
+		{
+			var allTiles = FootprintUtils.Tiles(world.Map.Rules, building, buildingInfo, topLeft).ToArray();
+			var neightborTiles = Util.ExpandFootprint(allTiles, true).Except(allTiles)
+				.Where(t => world.Map.Contains(t)).ToList();
```
Phrohdoh: You ought to be able to remove the lambda from this and have `.Where(world.Map.Contains)` for a tiny bit less overhead.

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
```diff
@@ -210,5 +219,25 @@ public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world)
...
+
+			var blockers = allTiles.SelectMany(t => world.ActorMap.GetUnitsAt(t))
+				.Where(a => a.Owner == producer.Owner && a.HasTrait<Mobile>());
```
Phrohdoh: Is `Mobile` necessary, or can we reduce this to `IMove`?

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
```diff
@@ -210,5 +219,25 @@ public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world)
...
+
+			foreach (var blocker in blockers)
+			{
+				var availableCells = neightborTiles.Where(t => blocker.Trait<Mobile>().CanEnterCell(t));
+
+				yield return new Order("Move", blocker, false)
+				{
+					TargetLocation = blocker.ClosestCell(availableCells)
```
Phrohdoh: One issue I ran into was a cell becoming full while actors were still trying to find a destination so they ended up hanging around still.

Of course you can spam click to eventually solve this.

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
penev92: Ah right. That had more stuff that got removed, but the lambda remained.

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
penev92: Yeah, when dealing with more than a few actors you sometimes need to issue the build order more than once :/

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
penev92: I started with `IMove`, but `Mobile` gives me the `.CanEnterCell(t)` that I use below, so I figured this won't affect aircraft anyway, might as well use `Mobile`.

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
penev92: I think it's easier to both see how far the expression extends (which is not always obvious at first glance thanks to our indentation rules) and what the operators are. Just makes for easier readability IMO.

Unit158:
I believe you could allow ally units to move if you use PlaceBuildingOrderGenerator to move them to the exact same place on all the clients, but this would be very hacky. This should still be investigated (moving allied units) in the future, I'll open an issue after this is merged.

Phrohdoh:
You could allow it but we shouldn't.
Giving orders to actors that aren't yours is not a polite thing to do.

penev92:
I **could** change `ValidateOrder` to allow orders from allies, but that opens the gates to tons of potential exploits and regressions.

Unit158:
From a gameplay PoV, not having allied soldiers move makes this feature seem half baked, that's all I'm saying.

Unit158:
Another possibility would be to add exceptions for certain orders somehow (like a firewall), but that would be dangerous (and potentially slower) too.

Unit158:
@chrisforbes any opinion on the discussion above?

Mailaender:
> From a gameplay PoV, not having allied soldiers move makes this feature seem half baked, that's all I'm saying.

If you can move your team mates soldiers by continuously placing structures on them, that sounds like it will only get exploited. In most cases it is your own units that just came out of the production buildings to clunk up your base, so I think this is fine as is.

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
```diff
@@ -210,5 +219,25 @@ public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world)
...
 		public string GetCursor(World world, CPos xy, MouseInput mi) { return "default"; }
+
+		IEnumerable<Order> TryClearBlockers(World world, CPos topLeft)
+		{
+			var allTiles = FootprintUtils.Tiles(world.Map.Rules, building, buildingInfo, topLeft).ToArray();
+			var neightborTiles = Util.ExpandFootprint(allTiles, true).Except(allTiles)
+				.Where(world.Map.Contains).ToList();
+
+			var blockers = allTiles.SelectMany(world.ActorMap.GetUnitsAt)
+				.Where(a => a.Owner == producer.Owner && a.HasTrait<Mobile>());
```
Mailaender: Can you add a `IsMoving` check here to not stop an army marching away already or harvesters stopping their business?

Mailaender:
Works as promised and doesn't desync in multiplayer. :white_check_mark:

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
penev92: Sure. I thought I had that.

pchote: `IsIdle` would be more appropriate than `IsMoving`.

penev92:
Updated to not bother actors that aren't idle.

abcdefg30:
Needs a rebase.

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
```diff
@@ -210,5 +220,25 @@ public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world)
 		}
 
 		public string GetCursor(World world, CPos xy, MouseInput mi) { return "default"; }
+
+		IEnumerable<Order> TryClearBlockers(World world, CPos topLeft)
```
Phrohdoh: Could this be renamed to reflect its return type, please?
The `Try` prefix is typically reserved for methods such as `TryParse*` and is a `void`.

Perhaps `ClearBlockersOrders`.

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
```diff
@@ -210,5 +220,25 @@ public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world)
...
+			var allTiles = FootprintUtils.Tiles(world.Map.Rules, building, buildingInfo, topLeft).ToArray();
+			var neightborTiles = Util.ExpandFootprint(allTiles, true).Except(allTiles)
+				.Where(world.Map.Contains).ToList();
+
+			var blockers = allTiles.SelectMany(world.ActorMap.GetUnitsAt)
+				.Where(a => a.Owner == producer.Owner && a.IsIdle && a.HasTrait<Mobile>());
+
+			foreach (var blocker in blockers)
+			{
+				var availableCells = neightborTiles.Where(t => blocker.Trait<Mobile>().CanEnterCell(t));
```
Phrohdoh: Can you check if this is empty, and `continue` if so?
Not doing so is causing auditory feedback that is actually invalid.

>Moving out!
When there is in fact nowhere to go.

Phrohdoh:
The comment regarding an empty `availableCells` is the change I am requesting.
I'd like to see the other in too, but we have a way of ignoring the comments we don't agree with (or like to label as bikeshedding).

After that, this has my :+1: assuming nothing else is changed.
I haven't been able to test this over a network game, and I think that should be done before merging.

penev92:
Rebased and addressed some more comments.

OpenRA.Mods.Common/Orders/PlaceBuildingOrderGenerator.cs
```diff
@@ -210,5 +220,28 @@ public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world)
...
+		{
+			var allTiles = FootprintUtils.Tiles(world.Map.Rules, building, buildingInfo, topLeft).ToArray();
+			var neightborTiles = Util.ExpandFootprint(allTiles, true).Except(allTiles)
+				.Where(world.Map.Contains).ToList();
+
+			var blockers = allTiles.SelectMany(world.ActorMap.GetUnitsAt)
+				.Where(a => a.Owner == producer.Owner && a.IsIdle)
+				.Select(a => new TraitPair<Mobile> { Actor = a, Trait = a.TraitOrDefault<Mobile>() });
+
+			foreach (var blocker in blockers.Where(x => x.Trait != null))
```
pchote: Recent perf Prs suggest that there is a non-negligible improvement from using an if () continue; in the loop instead.

RoosterDragon: It's a tiny saving that's only worth thinking about in code that gets called many times per second. Placing a building doesn't begin to come close - I wouldn't bother.

obrakmann:
Works as expected in SP, MP and in replays. Code looks sensible as well. :+1: 

We should probably mention somewhere that this will only work for idle actors.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/2326c2b26cd3aff1c4be4a9ddd50380f7f84001f)

PR #9240, opened 2015-09-03
Fix crash when aircraft is rearming at a helipad in TS

penev92:
Caused by both helipads missing an "active" animation.
TLDR - https://github.com/OpenRA/OpenRA/pull/9240/files?w=1

### DIFF
--- a/OpenRA.Mods.Common/Activities/Rearm.cs
+++ b/OpenRA.Mods.Common/Activities/Rearm.cs
@@ -56,7 +56,9 @@ namespace OpenRA.Mods.Common.Activities
 				if (!pool.GiveAmmo())
 					continue;
 
-				hostBuilding.Trait<WithSpriteBody>().PlayCustomAnimation(hostBuilding, "active");
+				var wsb = hostBuilding.Trait<WithSpriteBody>();
+				if (wsb.DefaultAnimation.HasSequence("active"))
+					wsb.PlayCustomAnimation(hostBuilding, "active");
 
 				var sound = pool.Info.RearmSound;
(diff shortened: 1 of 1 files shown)


Phrohdoh:
:+1: sure, thanks. :)

OpenRA.Mods.Common/Traits/Render/WithSpriteBody.cs
```diff
@@ -78,12 +78,13 @@ public virtual void BuildingComplete(Actor self)
...
 		public void PlayCustomAnimation(Actor self, string name, Action after = null)
 		{
-			DefaultAnimation.PlayThen(NormalizeSequence(self, name), () =>
-			{
-				DefaultAnimation.Play(NormalizeSequence(self, Info.Sequence));
-				if (after != null)
-					after();
-			});
+			if (DefaultAnimation.HasSequence(name))
+				DefaultAnimation.PlayThen(NormalizeSequence(self, name), () =>
```
pchote: We want this to throw because the outside code has asked this low level method to do something bogus.  The chexk should be in the reload code instead.

penev92:
Changed as suggested.

Phrohdoh:
My :+1: stands, but in a follow-up I'd like to see that value unhardcoded (out of scope here).

Unit158:
Can you reword the commit summary to "Don't try to play an animation that doesn't exist"

penev92:
That was the original commit message, but it got changed with the requested code change above because it's not generic any more. It's now very specific to rearming.

PR #9245, opened 2015-09-03
Load HashSets in OpenRA.Mods.Common/Warheads/Warhead.cs directly from MiniYaml

atlimit8:
`OpenRA.Mods.Common/Warheads/Warhead.cs` used `HashSet<string>`s created from `string[]`s loaded from MiniYaml, but #9181 makes that separation unnecessary.


RoosterDragon:
:+1: Would've done this originally if I hadn't overlooked it.

penev92:
:+1:

PR #9246, opened 2015-09-03
Change some arrays to sets

RoosterDragon:
In places where arrays were being treated as a set, just create a set directly. This reveals the intention of such collections better, and also improves performance by allowing set based methods to be used.

(The follow up of  #9181, #9245).


atlimit8:
Would have done these myself otherwise, so :+1:.

Mailaender:
Thanks!

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/2f356e99955907302f98381ac31fd4856fa215f1...4caa9359c8d13aba0a6cd62277dba3ffbf5788f2)

PR #9248, opened 2015-09-03
Lower shocktrooper health

ghost:
(Had to create a new fork and PR as I am unfamiliar with keeping a forked repo synced.)

Result of the ongoing discussion on the forums and feedback from the previous closed PR; http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=18117&postdays=0&postorder=asc&start=0 

This simply nerfs the shock-trooper health from 100hp to 60hp. They now die in one pillbox burst and from parabombs.

### DIFF
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -464,7 +464,7 @@ SHOK:
 		Name: Shock Trooper
 		Description: Elite infantry with portable tesla coils.\n  Strong vs Infantry, Vehicles\n  Weak vs Aircraft
 	Health:
-		HP: 100
+		HP: 60
 	Mobile:
 		Voice: Move
 	RevealsShroud:


Mailaender:
:+1: They seemed to be very tanky.

pchote:
The lack of interest in this from other reviewers is concerning, so at this point it looks unlikely that this will make it into the next stable release.  Sorry about that... hopefully we'll address the problem (if it still exists) in the next cycle.

I won't personally :+1: it because I think it is a bad idea to take any balance changes immediately before a release without adequate testing with the other playtest changes.  The last time we took a "simple" change without testing it completely broke the C&C mod balance, so I've learned my lesson there.

In fact, I haven't seen *any* serious discussion about our playtest balance.  So far the only balance-related comment i've seen was @Unit158 saying that he thought shock troopers had been nerfed too much (despite the fact that they haven't yet changed at all).

HenrytheSlav:
I definitely support this, large stacks of Stroopers are just pain to stop by any means. Exact amount is open for discussion to me but I was testing with 65/75 and would go for 65 or smthing like that. :+1: for me

MicroJOo:
im not sure about the one-burst-kill ... so for me ...this is too much 

it took 2 bursts for our previous pillbox to kill "any" infantry ... now it pretty much one-hit-kills every infantry except shocktroopers ... this was fine to me ... since the shocktrooper is a faction-unique ... 

if you could adjust the hp , so the pillbox still needs 2 bursts to kill it ... i would be happy ... otherwise 
:-1:

r34ch:
@pchote I completely agree with testing balance changes in playtests. I don't know how people go about organising these tests though if the PR doesn't get merged into bleed? Or do we checkout the PR branch. Can you explain?

@MicroJOo Perhaps we are playing a different build of oRA, but I have to completely disagree. 75hp allows shock troopers to survive a __direct__ parabomb hit with low health on bleed ~7/10 times. Test it.

MicroJOo:
@r34ch : still better then 10/10 times :.. as they do in current release ...

r34ch:
@MicroJOo Ok well lets keep the hp open for discussion until player feedback since everyone has different ideas. 

@Mailaender What needs done with this PR to get it into playtests?

obrakmann:
Considering that the other combat infantry units have 40-50 HP, I think 60 is well enough for the Shocks. Should be enough for people not to base their entire strategy on them. If it's too much we can still jack it up to 70 in the second playtest.

:+1:

abcdefg30:
Looks simple enough. :+1: 

> but fine , lets do a playtest with it

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/21a2dbf7a40fa423aa8ee7cfdee1db405b29dc27...d92454dfb6c56c256fbab5978f8dc791c6caf30a)

MicroJOo:
obrakmann : shocktroopers are a "faction special" ... so they are different then normal infantry ... you need a tech centre and a tesla-coil to be able to build it ... they have pretty serious damage-output and  high range ... "together" with a good amount of hp ... they are pretty good against allies ... 

and 70hp would not help anything ... did you even read what i wrote ? probably not ...

Issue #9244, opened 2015-09-03
Don't treat trees like regular actors


Aaron-Lloyd:
So it's pretty well known that maps with a very large number of trees (or any other type of actor of course) tend to introduce quite a bit of lag to games. But because trees are indestructible and unmoving, is there any way that trees could be treated more like terrain elements than actors in order to avoid the lag that they introduce?

Thanks.

penev92:
Hi. Everything that is physically present in the world (and even some things that aren't) are actors. It's how the engine is set up from the foundations and that won't change.
But do you have any evidence of this, because this is the first time I hear the idea that trees can cause lag?
Trees don't actually do anything other than sit around, so I find that quite odd.

Aaron-Lloyd:
You can try this map here. It lags in singleplayer too quite a bit, but is extremely noticeable in multiplayer from my testing.

http://resource.openra.net/maps/4414

About 4780 actors in total FYI.

Aaron-Lloyd:
Also, if I recall correctly, Materia's had to reduce tree numbers to fix some lag issues on some of his maps.

pchote:
We will at some point implement (visual only - they wont be cleared) forest fires, so they will need to stay as actors.

RoosterDragon:
The sheer number of trees stresses some of the slower parts of the engine:
- Trees are frozen under fog. This means that they all have to be constantly tracked to see if they're within the sight range of the player. Each player has a different view of the map, so this gets slower just by adding more players. 
- The radar needs to constantly track every actor which should be showing up on it.
- The number of actors make our sync checks slow - because there's more stuff to check to make sure it's all in sync.

Some of my immediate musings to improve each of these are:
- For permanent and unchanging features of the map - we should avoid making them freezeable somehow - if they can't possibly change when moving it or out of the sight of the player then we can avoid the checks to see if they're in range.
- There are probably some things we can do to improve the speed of the radar generally.
- If trees weren't freezeable that would make sync checks a bit faster since we wouldn't need to check their frozen state was in sync.

LavenderMoon:
[...]

PR #9250, opened 2015-09-03
Make rocks and trees HiddenUnderShroud instead of FrozenUnderFog

reaperrr:
This improves performance significantly when many trees and rocks are on the map and shroud/fog are enabled.

Should reduce the performance issues described in #9244.

### DIFF
--- a/mods/cnc/rules/defaults.yaml
+++ b/mods/cnc/rules/defaults.yaml
@@ -591,8 +591,7 @@
 	Armor:
 		Type: Wood
 	AutoTargetIgnore:
-	FrozenUnderFog:
-		StartsRevealed: true
+	HiddenUnderShroud:
 	ScriptTriggers:
 
 ^TibTree:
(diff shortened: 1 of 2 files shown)


Phrohdoh:
This will be undone when destruction is implemented for these actors.

Mailaender:
Needs a rebase.

reaperrr:
> This will be undone when destruction is implemented for these actors.

But we don't know when that will be.
On a test map with lots of trees this saved ~10ms per tick even on my relatively fast system (5ms/tick vs. 15ms/tick on bleed, with just three local instances of OpenRA), that's too much to just leave it be.

And at least rocks and tiberium trees won't change in the future.

pchote:
I'm okay with this as long as somebody doesn't try to use "it hurts perf" as an excuse to block the tree burning feature when it does eventually happen.

Mailaender:
:+1:

RoosterDragon:
:+1:

This is quite a nice saving, so I'm just going to cross my fingers and hope we don't have to undo it for other future features - or that by the time we do we've improved frozen actor perf enough that it's not such a bad thing.

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4caa9359c8d13aba0a6cd62277dba3ffbf5788f2...56e0fc955c9774c20f6d463db2e911acb16119e6)

PR #9255, opened 2015-09-04
Add a game speed dropdown to the lobby options.

pchote:
All the recent perf improvements have made it (hopefully) feasible to implement a game speed modifier by adjusting the timestep and order latency values.  This will need testing in real network games (not just toy tests) before we consider merging it.

### DIFF
--- a/OpenRA.Game/Network/Session.cs
+++ b/OpenRA.Game/Network/Session.cs
@@ -172,6 +172,7 @@ namespace OpenRA.Network
 		{
 			public string ServerName;
 			public string Map;
+			public int Timestep = 40;
 			public int OrderLatency = 3; // net tick frames (x 120 = ms)
 			public int RandomSeed = 0;
 			public bool FragileAlliances = false; // Allow diplomatic stance changes after game start.
--- a/OpenRA.Game/Network/Session.cs
+++ b/OpenRA.Game/Network/Session.cs
@@ -188,6 +188,7 @@ namespace OpenRA.Network
 			public int StartingCash = 5000;
 			public string TechLevel = "none";
 			public string StartingUnitsClass = "none";
+			public string GameSpeedType = "default";
 			public bool ShortGame = true;
 			public bool AllowVersionMismatch;
 			public string GameUid;
--- a/OpenRA.Game/Widgets/WidgetUtils.cs
+++ b/OpenRA.Game/Widgets/WidgetUtils.cs
@@ -158,14 +158,14 @@ namespace OpenRA.Widgets
 				DrawRGBA(cornerBottomRight, new float2(bounds.Right - cornerBottomRight.Size.X, bounds.Bottom - cornerBottomRight.Size.Y));
 		}
 
-		public static string FormatTime(int ticks)
+		public static string FormatTime(int ticks, int timestep)
 		{
-			return FormatTime(ticks, true);
+			return FormatTime(ticks, true, timestep);
 		}
 
-		public static string FormatTime(int ticks, bool leadingMinuteZero)
+		public static string FormatTime(int ticks, bool leadingMinuteZero, int timestep)
 		{
-			var seconds = (int)Math.Ceiling(ticks / 25f);
+			var seconds = (int)Math.Ceiling(ticks * timestep / 1000f);
 			return FormatTimeSeconds(seconds, leadingMinuteZero);
 		}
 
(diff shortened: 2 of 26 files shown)


OpenRA.Mods.Common/UtilityCommands/ActorStatsExport.cs
```diff
@@ -79,7 +80,9 @@ public static DataTable GenerateTable()
 							var burst = weapon.Burst;
 							var delay = weapon.BurstDelay;
 							var damage = warhead.Damage;
-							var damagePerSecond = (1000f / Game.Timestep) * (damage * burst) / (delay + burst * rateOfFire);
+
+							// Assumes the default game speed (25 ticks / sec)
+							var damagePerSecond = (1000f / 40) * (damage * burst) / (delay + burst * rateOfFire);
```
Mailaender: I suggest you don't leave a magic variable here as you just made it configurable and instead try:

```csharp
var gameSpeed = Game.ModData.Manifest.Get<GameSpeeds>().First();
```

penev92:
Played some test games with @Mailaender and noticed that the support power timers get scaled with the speed (for example Chronosphere takes 1 minute to charge) but the max times don't. So the tooltip says `01:00 / 04:00` when it starts to charge.

Mailaender:
Otherwise it was playable. It is more demanding, but a 2v2 comp stomp is still possible performance wise.

Mailaender:
Also replays work fine. :white_check_mark:

OpenRA.Game/Widgets/WidgetUtils.cs
```diff
@@ -158,14 +158,14 @@ public static void DrawPanelPartial(string collection, Rectangle bounds, PanelSi
 				DrawRGBA(cornerBottomRight, new float2(bounds.Right - cornerBottomRight.Size.X, bounds.Bottom - cornerBottomRight.Size.Y));
 		}
 
-		public static string FormatTime(int ticks)
+		public static string FormatTime(int ticks, int timestep = 40)
```
Mailaender: Better not allow those overloads with magic values and instead fix all occurrences.

OpenRA.Mods.Common/UtilityCommands/ActorStatsExport.cs
pchote: This would report DPS values for the slowest game speed, which isn't what users expect.

Mailaender: A `bool GameSpeed.IsDefault` might help you with finding the right one.

pchote: I'm not prepared to add core engine code just to work around a utility rule that its not clear whether anyone uses.

Mailaender: It is certainly cleaner for the lobby code as well. Currently you are enforcing a hidden string comparison. If someone does not define a game speed named "Default", but maybe call it "Normal" like in the originals, it will also break.

pchote:
Updated.

OpenRA.Mods.Common/UtilityCommands/ActorStatsExport.cs
pchote: The names are separated from the ids to support future translation, so that isn't true.

OpenRA.Game/Network/Session.cs
```diff
@@ -187,6 +188,7 @@ public class Global
 			public int StartingCash = 5000;
 			public string TechLevel = "none";
 			public string StartingUnitsClass = "none";
+			public string GameSpeedType = "default";
```
Mailaender: What I mean is that this default type is hard-coded to a specific undocumented ID.

OpenRA.Mods.Common/UtilityCommands/ActorStatsExport.cs
Mailaender: If you don't want to change your system and add a boolean than just let it pick the one with the "default" ID.

pchote:
Updated.  I changed the actor stats export to report damage per tick to remove the arbitrary ticks / second variable.  In the process I discovered that the stats export command doesn't even work, so I have also filed #9272.

Mailaender:
Needs a rebase.

penev92:
Can't find anything obviously wrong with it, so :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/1eef037432d479349852bc7769c5ced7fe3ca862...f7402b30a11a0d52ce44f9643b73786a6c03397d)

LavenderMoon:
In the future, could this be set to allow changing game speed while in-game for singleplayer and skirmish games (and maybe co-op, where it could use the slowest speed of all players like in Introversion's *DEFCON*)?

PR #9256, opened 2015-09-04
Added AI search for new resource patches avoiding enemies

### LINKED ISSUE #9232: AI harvesters don't search far away enough for new resources

Mailaender:
It is a similar check I removed from https://github.com/OpenRA/OpenRA/pull/9144 however it doesn't care about the resource claim layer, because that will run in the FindResource activity limited radius fine grained path search so we can skip it. Running two path finder searches is expensive so this is only the last resort when the current patch has dried out, not the default for every AI harvesting operation.

This also  brings back the enemy avoidance check which has been removed in https://github.com/OpenRA/OpenRA/pull/6986. I believe it didn't work, because it was comparing length of cell position with world distances. I fixed that when rebasing the code snippet.

Closes https://github.com/OpenRA/OpenRA/issues/9232.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -13,6 +13,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Mods.Common.Activities;
+using OpenRA.Mods.Common.Pathfinder;
 using OpenRA.Mods.Common.Traits;
 using OpenRA.Primitives;
 using OpenRA.Support;
@@ -105,6 +106,9 @@ namespace OpenRA.Mods.Common.AI
 			"Should match maximum adjacency of naval structures.")]
 		public readonly int CheckForWaterRadius = 8;
 
+		[Desc("Avoid enemy actors nearby when searching for a new resource patch. Should be somewhere near the max weapon range.")]
+		public readonly WDist HarvesterEnemyAvoidanceRadius = WDist.FromCells(8);
+
 		[Desc("Production queues AI uses for producing units.")]
 		public readonly string[] UnitQueues = { "Vehicle", "Infantry", "Plane", "Ship", "Aircraft" };
 
@@ -171,6 +175,11 @@ namespace OpenRA.Mods.Common.AI
 
 		public Player Player { get; private set; }
 
+		readonly DomainIndex domainIndex;
+		readonly ResourceLayer resLayer;
+		readonly ResourceClaimLayer territory;
+		readonly IPathFinder pathfinder;
+
 		readonly Func<Actor, bool> isEnemyUnit;
 		Dictionary<SupportPowerInstance, int> waitingPowers = new Dictionary<SupportPowerInstance, int>();
 		Dictionary<string, SupportPowerDecision> powerDecisions = new Dictionary<string, SupportPowerDecision>();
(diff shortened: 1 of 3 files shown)


reaperrr:
Wouldn't it be better to move/copy the threat avoidance radius to `Harvester` and the check to `FindResources`, so non-bot mission 'AIs' and human players benefit as well? Then we could afford increasing the default search radius, too.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -105,6 +106,9 @@ public sealed class HackyAIInfo : IBotInfo, ITraitInfo
 			"Should match maximum adjacency of naval structures.")]
 		public readonly int CheckForWaterRadius = 8;
 
+		[Desc("Avoid enemy actors nearby when searching for a new resource patch. Should be somewhere near the max weapon range.")]
+		public readonly WDist HarvesterEnemyAvoidanceRadius = new WDist(8);
```
reaperrr: I doubt a radius of ~0.07 cells is enough... better use `.FromCells`.

Mailaender: Done.

Mailaender:
I fear that will change behavior. You order your harvester to go for that tile, but the activity will spot some enemies you haven't even scouted yet and re-route it somewhere else which will irritate and annoy human players. Making harvesters on themselves smarter in that regard requires a fair bit of tweaking and experimenting. I just want this to solve https://github.com/OpenRA/OpenRA/issues/9232 and get back to what we had before for now (and this pull request).

RoosterDragon:
I agree with Mailaender - it should be the higher power making higher-level decisions like this (be it the player or HackyAI) and not the unit-level AI.

RoosterDragon:
I tested this and can confirm the AI harvesters will now start expanding their search when needed, and the enemy avoidance works too.

However, I did notice the AI harvesters seemed to largely prefer continuing to harvest tiny amounts of regrown ore rather than move to a new field. I think you may need to keep the claim check so there's a bit more motivation for the AI harvesters to more to new fields rather than fighting over scraps.

Mailaender:
Well, harvesting slowly regrowing ore is safer and also what human players usually do. I also don't get why that doesn't already work. It is returning to it's previous ore patch, find resources is triggered so it will try an unclaimed patch, find none and become idle so the AI overlord will force them to re-route their search somewhere else. I guess in your test cases an ore patch wasn't over-harvested enough so there still was a chunk left.

Mailaender:
Nevermind, the AI overlord will then re-route it back to the neared ore patch as it doesn't care for the harv claimer. This can actually dead lock.

Mailaender:
I was also missing the domain index check which is crucial for performance. Solved it with a new extension method to avoid redundancy as well.

Mailaender:
Also managed to cache several trait lookups for extra perf.

LipkeGu:
Question: What is happen when every Tiberium field has an enemy refinery or tank next to it? will this logic be using the fall back (default logic without enemy scan)?

penev92:
That would mean defaulting to the "stupid" case of suicide harvesters, so I'm hoping no.

Mailaender:
It will return CPos.Zero which will then initiate a local FindResources scan. That finding nothing will result in harvesters circling like in https://github.com/OpenRA/OpenRA/issues/9232. The cycle will repeat until it finds an ore patch free of enemy units.

RoosterDragon:
This seems to have good behaviour all round:
- When a patch runs out, excess harvesters will seek new patches, but a couple will continue working the regrowing ore.
- Enemies are avoided.
- If harvesters are cut off from new patches by enemies, they just cycle round those they do have access to.

:+1:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4d900ffc20091a23823474ecda08226355346a78...e3a5d6c18616edd7d661d0d1d8dab947dd539095)

PR #9257, opened 2015-09-04
Increased harvester search radii

reaperrr:
Even before performance of harvesters was improved, I felt the search radius was a little too low.

This PR slightly-tomoderately increases the search radii in all mods.

This should improve the situation in cases where #9256 doesn't help.

### DIFF
--- a/mods/cnc/rules/vehicles.yaml
+++ b/mods/cnc/rules/vehicles.yaml
@@ -57,8 +57,8 @@ HARV:
 		Capacity: 20
 		LoadTicksPerBale: 12
 		UnloadTicksPerBale: 6
-		SearchFromProcRadius: 24
-		SearchFromOrderRadius: 12
+		SearchFromProcRadius: 25
+		SearchFromOrderRadius: 15
 	Mobile:
 		Speed: 85
 	Health:
--- a/mods/d2k/rules/vehicles.yaml
+++ b/mods/d2k/rules/vehicles.yaml
@@ -61,8 +61,8 @@ harvester:
 		HarvestFacings: 8
 		Resources: Spice
 		UnloadTicksPerBale: 5
-		SearchFromProcRadius: 24
-		SearchFromOrderRadius: 12
+		SearchFromProcRadius: 30
+		SearchFromOrderRadius: 15
 	Health:
 		HP: 1000
 	Armor:
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -254,8 +254,8 @@ HARV:
 		Capacity: 20
 		Resources: Ore,Gems
 		UnloadTicksPerBale: 1
-		SearchFromProcRadius: 24
-		SearchFromOrderRadius: 12
+		SearchFromProcRadius: 30
+		SearchFromOrderRadius: 15
 	Health:
 		HP: 600
 	Armor:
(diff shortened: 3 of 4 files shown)


OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -52,10 +52,13 @@ public class HarvesterInfo : ITraitInfo, Requires<MobileInfo>
...
 
 		[Desc("Initial search radius (in cells) from the refinery that created us.")]
-		public readonly int SearchFromProcRadius = 24;
+		public readonly int SearchFromProcRadius = 30;
 
 		[Desc("Search radius (in cells) from the last harvest order location to find more resources.")]
-		public readonly int SearchFromOrderRadius = 12;
+		public readonly int SearchFromOrderRadius = 15;
+
+		[Desc("Adds this value to search radius when owner is not a human player.")]
```
RoosterDragon: I feel the description should mention this doesn't apply to bots like you have said in your PR comment.

Phrohdoh: >when owner is not a human player.

Mailaender:
Tweaking this on a per mod/map basis which has different (default) and individual resource patch distances between each other and the refinery allows a more fine grained control so I don't think we want a generalized https://github.com/reaperrr/OpenRA/commit/f2ed3c9fcc4398e64c1891ae20a92f982ba73662.

reaperrr:
Ok, any suggestions where to put that? Should I make it a world trait?

> Tweaking this on a per mod/map basis which has different (default) and individual resource patch distances between each other and the refinery allows a more fine grained control so I don't think we want a generalized reaperrr@f2ed3c9.

reaperrr:
Updated.
Reverted the changes to `Harvester`, made the AI radius bonus a world trait.

Mailaender:
> Ok, any suggestions where to put that? Should I make it a world trait?

No. I mean why would you want this. This just adds extra overhead and complexity. Now you need to know about and tweak two values. Also your definition of AI is also incompatible with multiplayer bots whose trait names also says AI which will only confuse.

reaperrr:
> Also your definition of AI is also incompatible with multiplayer bots whose trait names also says AI which will only confuse.

I already changed it so that it applies to bots as well, so it applies to all AIs now.

> No. I mean why would you want this. This just adds extra overhead and complexity. Now you need to know about and tweak two values.

Because one value cannot produce satisfying results for all 3 cases, human players, bots and dumb singleplayer AI. 
Singleplayer AI is completely incapable of "manually" sending harvesters to farther away resources, so it needs a (much) larger search radius.
Human-controlled harvesters are supposed to have a low search radius so they don't accidentally move too close to an enemy or otherwise inconvenient location.
Bots have a fallback, but performance-wise, particularly on large maps - where the fallback search is more expensive AND kicks in more often due to larger distances - the regular search is still much, much better.

How else do you want to solve this? The only other thing I can think of is copy-pasting your bot fallback to a world trait and modify it so it only applies to singleplayer AI and then add that to singleplayer map rules, but in my opinion that's not really any better (or even as flexible) as my approach.

Mailaender:
You could give the Lua API an additional overload `FindResources(int range)` or `FindResources(CPos target)` if you want to fix this properly for single-player missions.

reaperrr:
I don't feel like dealing with Lua at the moment, so I stripped this PR down to just increased search radii.

abcdefg30:
Seems simple enough. :+1:

GraionDilach:
:+1:

PR #9258, opened 2015-09-04
Added stepless mouse wheel actor rotation to the editor

Mailaender:
This had a hard-coded limit of 8 facings which would be sufficient for Dune II units, but not later Westwood game era sprites and especially not voxels.

### DIFF
--- a/OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
+++ b/OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
@@ -53,9 +53,9 @@ namespace OpenRA.Mods.Common.Widgets
 
 		public bool HandleMouseInput(MouseInput mi)
 		{
-			// Exclusively uses left and right mouse buttons, but nothing else
+			// Exclusively uses mouse wheel and right mouse buttons, but nothing else
 			// Mouse move events are important for tooltips, so we always allow these through
-			if (mi.Button != MouseButton.Left && mi.Button != MouseButton.Right && mi.Event != MouseInputEvent.Move)
+			if (mi.Button != MouseButton.Right && mi.Event != MouseInputEvent.Move && mi.Event != MouseInputEvent.Scroll)
 				return false;
 
 			var cell = worldRenderer.Viewport.ViewToWorld(mi.Location);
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
```diff
@@ -55,7 +55,7 @@ public bool HandleMouseInput(MouseInput mi)
 		{
 			// Exclusively uses left and right mouse buttons, but nothing else
 			// Mouse move events are important for tooltips, so we always allow these through
-			if (mi.Button != MouseButton.Left && mi.Button != MouseButton.Right && mi.Event != MouseInputEvent.Move)
+			if (mi.Button != MouseButton.Left && mi.Button != MouseButton.Right && mi.Event != MouseInputEvent.Move && mi.Event != MouseInputEvent.Scroll)
```
pchote: You will also need to update the comment.

pchote: This brush no longer uses the Left mouse button.

Mailaender: Updated.

reaperrr:
This goes through all 256 internal facings, which might be a nice checkbox-feature for voxels (though I doubt it would get used much even for them) but makes it quite slow and tiresome to rotate actors, even more so when they have only 16 or 8 visible facings.

I'd prefer if you multiplied the delta with a value of 8, so that voxels and sprite actors with 32 facings change visible facing by 1/32 for every mousewheel 'step', actors with 16 facings 1 every 2 steps, actors with 8 facings 1 every 4 steps. An upper limit of 32 facings should be sufficient for both 1st-party and 3rd-party mods and would make rotation much faster and more responsive.

Maybe you could expose the wheelstep multiplier to yaml so we can set it per mod and use 8 as default value, that would be even better.

Mailaender:
I would have liked to scale this with the available facing of each sprite as they will still be wrong for special cases like say the TD boats. However that failed as I can't look that up without access to the traits dictionary so I stayed with keeping this to 1.

pchote:
The current value works nicely with scrolling on trackpads, FWIW

reaperrr:
> The current value works nicely with scrolling on trackpads, FWIW

Ok, I don't want to cause a regression for trackpads (or mousewheels less precise than mine).

It's better than bleed in any case, so :+1:

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/19aecb76f61968e0cbfe1c6009a96c2fac74163a...1eef037432d479349852bc7769c5ced7fe3ca862)

PR #9259, opened 2015-09-04
AI counts all aircraft rearm buildings

atlimit8:
Currently the AI only counts `hpad` (Helipad) or whatever is the first entry of `RearmBuildings` for aircraft when deciding whether to build rearm buildings for aircraft, ignoring other entries such as `afld` (Airfield).

~~If count matching behaviour is unwanted for certain aircraft, then a `NeedsRearmBuilding`/`ShouldHaveRearmBuilding` field, `[Recommended]RearmBuildingRatio` field, or the like should be added.~~

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -421,7 +421,7 @@ namespace OpenRA.Mods.Common.AI
 			if (ammoPoolsInfo.Any(x => !x.SelfReloads))
 			{
 				var countOwnAir = CountUnits(actorInfo.Name, Player);
-				var countBuildings = CountBuilding(aircraftInfo.RearmBuildings.FirstOrDefault(), Player);
+				var countBuildings = aircraftInfo.RearmBuildings.Sum(b => CountBuilding(b, Player));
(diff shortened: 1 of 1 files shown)


Mailaender:
:+1: Maybe leave a TODO, but this isn't a problem in practice as the AI can't capture enemy buildings and one side can only build the right rearm buildings.

atlimit8:
Well if anyone does not want the count matching AI logic to apply to all aircraft with `RearmBuildings`, they can create a new issue.

If this causes any problems to appear, then it probably just exposing them.

Phrohdoh:
@Mailaender it is an issue as a map author can give AIs multiple rearm buildings. (pre-placed, custom rules)

atlimit8:
@Phrohdoh if they are listed under `RearmBuildings`, then they will be counted. Are you saying that those other rearm buildings should not be counted? If so, then something like `PreferredRearmBuilding` could be used.

Phrohdoh:
I'm referring to the current code which only takes the first building into account.

atlimit8:
@Phrohdoh so you prefer this PR's behaviour then?

Phrohdoh:
Yes.

RoosterDragon:
:+1:

PR #7757, opened 2015-03-28
Add support for TS-style random tile variants.

pchote:
Adds more cheap polish for our TS and RA2 mods, and fixes the last major unhandled item on #5350.

PR #9260, opened 2015-09-04
Fixed the legacy isometric tileset importer

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/7757.

### DIFF
--- a/OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
+++ b/OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
@@ -87,7 +87,7 @@ namespace OpenRA.Mods.TS.UtilityCommands
 									images.Add(variant);
 							}
 
-							Console.WriteLine("\t\tImage: {0}", images.JoinWith(", "));
+							Console.WriteLine("\t\tImages: {0}", images.JoinWith(", "));
 
 							var templateWidth = s.ReadUInt32();
 							var templateHeight = s.ReadUInt32();


OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
```diff
@@ -61,7 +61,7 @@ public void Run(ModData modData, string[] args)
 					var section = file.GetSection("TileSet{0:D4}".F(tilesetGroupIndex));
 
 					var sectionCount = int.Parse(section.GetValue("TilesInSet", "1"));
-					var sectionFilename = section.GetValue("FileName", "");
+					var sectionFilename = section.GetValue("FileName", "").ToLowerInvariant();
```
pchote: This affects the two TS tilesets.  Can you please reimport them?

Mailaender: I doubt it is a problem. As not defining `Images` leads to weird error messages I was mostly guessing. This doesn't break anything in practice.

penev92:
:+1:

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
GraionDilach: It does not affect the current tilesets because mixes are case-insensitive in all OS. It affects loose files alone.

reaperrr:
This must have been broken for ages (definitely before the last 2 playtests), should this go into stable? Personally I'd vote yes (it doesn't affect ingame code, so it wouldn't require another playtest).

Mailaender:
We don't ship OpenRA.Mods.TS.dll in releases so it doesn't matter.

reaperrr:
Nevermind, seems like I wasn't fully awake yet, overlooked both the "isometric" part and dll this is in.

pchote:
Did you mean "images" in the commit message?

Mailaender:
No.

pchote:
Ok, then please give this an unambiguous commit message.  "this changed to the plural form" is meaningless when viewed in the git history.

Mailaender:
Updated.

abcdefg30:
Looks good to me. :+1:

PR #9263, opened 2015-09-04
Move asset installation to the ModChooser

### LINKED ISSUE #8990: Move asset installation code and UI into the mod chooser.
The original plan for our mod chooser has always included asset management, but this second phase has not yet happened.  This is going to be a relatively big job to do fully, but this is a simple first step to get the ball rolling.

I expect the following steps to be involved:
* Cache the `ContentInstaller` mod data inside the `ModMetadata` objects that are read at game start.
* Move the mod installation check from `BlankLoadScreen` to `ModBrowserLogic`, and run it when selecting a new mod in the UI.
* Replace the "Play" button text in the mod chooser with "Install" if installation is required, and open the installation dialog instead of launching the mod.
* Keep a check for missing content in `BlankLoadScreen`, but return to the mod chooser if it fails instead of showing a dialog.
* Adjust the music installation code to also go via the mod chooser.
* Adjust the dialog theme to fit our website-inspired UI.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25517869-move-asset-installation-code-and-ui-into-the-mod-chooser?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

penev92:
Closes #8990.

That ticket pretty much explains what is going on here and why.

The progress bar will probably need polishing. Other than that I think it's done.

I **strongly** suggest reviewing the code by commits and not the entire diff.


OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
```diff
@@ -159,5 +194,41 @@ void LoadMod(ModMetadata mod)
...
+
+		static bool IsModInstalled(ModMetadata mod)
+		{
+			var manifest = new Manifest(mod.Id);
+			var folders = manifest.Folders.Where(f => f.Contains("^")).Select(f => f.Substring(f.IndexOf("^", StringComparison.Ordinal)));
+
+			var filesList = new List<string>();
+			foreach (var folder in folders)
+			{
+				var path = Platform.ResolvePath(folder).Replace("/", "\\");
```
Mailaender: This hard-codes the directory separator to the Windows only `\` which breaks everywhere else so please change this to
```csharp
var path = Platform.ResolvePath(folder);
```

OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
```diff
@@ -159,5 +194,41 @@ void LoadMod(ModMetadata mod)
...
+				if (!Directory.Exists(path))
+					continue;
+
+				var filesInDir = Directory.GetFiles(path).Select(f => f.ToLower());
```
Mailaender: This isn't just the files, but the file + folder. It is `~/.openra/Content/` on Unix so making this `~/.openra/content/` will break on case sensitive file systems. Change it to:

```csharp
var filesInDir = Directory.GetFiles(path);
```

OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
```diff
@@ -159,5 +194,41 @@ void LoadMod(ModMetadata mod)
...
+				filesList.AddRange(filesInDir);
+			}
+
+			var files = filesList.ToHashSet();
+
+			return mod.AssetInformation.TestFiles.All(f => files.Contains(Platform.ResolvePath(f).Replace("/", "\\").ToLower()));
```
Mailaender: This combines both mistakes of above and breaks the detection on non-Windows systems completely. The following will work:

```charp
return mod.AssetInformation.TestFiles.All(f => files.Contains(Platform.ResolvePath(f)));
```

penev92:
Fixed the path check workarounds.

OpenRA.Mods.Common/Widgets/Logic/Installation/DownloadPackagesLogic.cs
```diff
@@ -52,9 +57,9 @@ void ShowDownloadDialog()
 
 			var cancelButton = panel.Get<ButtonWidget>("CANCEL_BUTTON");
 
-			var mirrorsFile = Platform.ResolvePath("^", "Content", Game.ModData.Manifest.Mod.Id, "mirrors.txt");
+			var mirrorsFile = Platform.ResolvePath("^", "Content", "mirrors.txt");
```
Mailaender: We probably want to store this in `Path.GetTempPath()` instead.

Mailaender:
From a usability standpoint if I run the game with `OpenRA.Game.exe Game.Mod=ts` it will now start the `modchooser` and highlights the first entry (which is `d2k`). As we store `Game.PreviousMod` in the settings we could highlight it in the mod chooser in that case so you immediately see the install assets button of the mod you wanted to start.

Mailaender:
Content detection and download works fine on Linux. :white_check_mark:

LipkeGu:
@penev92 you could also apply the TFD changes when you are on it so it is done in one way

OpenRA.Game/ContentInstaller.cs
```diff
@@ -0,0 +1,39 @@
...
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+
+namespace OpenRA
+{
+	public class ContentInstaller : IGlobalModData
```
pchote: Could you please add a comment `// Referenced from ModMetadata, so needs to be in OpenRA.Game :(` before this?

OpenRA.Game/ModMetadata.cs
```diff
@@ -24,6 +24,7 @@ public class ModMetadata
 		public string Version;
 		public string Author;
 		public bool Hidden;
+		public ContentInstaller AssetInformation;
```
pchote: Can we call the variable `Installer` or something else more closely related to the type?

OpenRA.Mods.Common/InstallUtils.cs
```diff
@@ -93,6 +68,10 @@ public static string GetMountedDisk(Func<string, bool> isValidDisk)
 						}
 					}
 
+					var dir = Path.GetDirectoryName(dest);
+					if (!Directory.Exists(dir))
+						Directory.CreateDirectory(dir);
```
pchote: It looks like you can now remove the `Directory.CreateDirectory(destPath);` at the start of the method.

OpenRA.Mods.Common/Widgets/Logic/Installation/DownloadPackagesLogic.cs
```diff
@@ -21,24 +20,30 @@ namespace OpenRA.Mods.Common.Widgets.Logic
...
 		{
 			this.mirrorListUrl = mirrorListUrl;
 			this.afterInstall = afterInstall;
+			this.modId = modId;
 
 			panel = widget.Get("INSTALL_DOWNLOAD_PANEL");
 			progressBar = panel.Get<ProgressBarWidget>("PROGRESS_BAR");
 			statusLabel = panel.Get<LabelWidget>("STATUS_LABEL");
 
+			var modName = ModMetadata.AllMods[modId].Title;
```
pchote: You could improve separation by passing the modName (`modTitle`?) instead of the modId and doing this lookup.

pchote: Oh, nevermind.  I see that you use `modId` in other places too.

OpenRA.Mods.Common/Widgets/Logic/Installation/DownloadPackagesLogic.cs
pchote: It can have a `Path.GetRandomFileName()` too (and don't forget to remove it when you're done).

pchote: Or even better, we could go with my suggestion (which was ignored) from the original PR, and use the version of the `Download` constructor that returns a string instead of saving a file to disk.

pchote: See `MapCache.QueryRemoteMapDetails` for an example on how to use that version.

OpenRA.Mods.Common/Widgets/Logic/Installation/InstallFromCDLogic.cs
```diff
@@ -74,6 +76,9 @@ void CheckForDisk()
 					InstallTFD(Platform.ResolvePath(path, "data1.hdr"));
 			else
 			{
+				var info = insertDiskContainer.Get<LabelWidget>("INFO2");
```
pchote: `INFO2` isn't a great ID.  Can we call this `MOD_TITLE` instead?

OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
```diff
@@ -140,18 +147,46 @@ void RebuildModList()
 			}
 		}
 
-		void SelectMod(ModMetadata mod)
+		void SelectMod(ModMetadata mod, bool? isInstalled)
 		{
+			if (isInstalled == null)
```
pchote: This is pretty ugly.  You can make this cleaner and also save some unnecessary checks when scrolling through mods by defining a `Cache<string, bool> modInstallStatus`.

OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
```diff
@@ -140,18 +147,46 @@ void RebuildModList()
...
 				modOffset = selectedIndex - 4;
+
+			loadButton.Text = isInstalled.Value ? "Load Mod" : "Install Assets";
 		}
 
-		void LoadMod(ModMetadata mod)
+		void LoadMod(ModMetadata mod, bool? isInstalled)
 		{
+			if (isInstalled == null)
+				isInstalled = IsModInstalled(mod);
```
pchote: Likewise here.

OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
```diff
@@ -140,18 +147,46 @@ void RebuildModList()
...
+
+				var widgetArgs = new WidgetArgs
+				{
+					{ "continueLoading", () =>
+						Game.RunAfterTick(() => Game.InitializeMod(Game.Settings.Game.Mod, new Arguments())) },
+					{ "mirrorListUrl", installData.PackageMirrorList },
+					{ "modId", mod.Id }
+				};
+
+				modChooserPanel.Get<BackgroundWidget>("bg").Visible = false;
```
pchote: Widget IDs in yaml should be in upper case.

mods/cnc/chrome.yaml
```diff
@@ -146,31 +145,6 @@ textfield-focused: chrome.png
 	corner-bl: 64,254,2,2
 	corner-br: 126,254,2,2
 
-# A copy of button
```
pchote: Is this deliberate?  Progress bars are used in more places than just the installer!

mods/d2k/chrome.yaml
```diff
@@ -173,30 +173,6 @@ scrollbar: buttons.png
 	up_arrow: 32,140,16,16
 	up_pressed: 32,140,16,16
 
-# A copy of dialog3 (pressed button)
```
pchote: Likewise here.

OpenRA.Game/ContentInstaller.cs
```diff
@@ -0,0 +1,39 @@
...
+	{
+		public readonly string MenuWidget = null;
```
pchote: These three `*Widget` variables are now unused.

mods/ra/chrome.yaml
```diff
@@ -502,30 +502,6 @@ scrollbar: buttons.png
 	up_arrow: 132,140,16,16
 	up_pressed: 132,140,16,16
 
-# A copy of dialog3 (pressed button)
```
pchote: Here too.

pchote:
The buttons in the main mod chooser UI are 35px high.  All your installer buttons are 32px high.

pchote:
After the asset download completes it takes you back to the mod chooser.  It should try to load the mod instead (which can then drop back to the chooser itself if something is busted).

mods/cnc/chrome.yaml
Mailaender: I checked the map download but it vanishes so fast that I am not sure.

pchote:
The "Back" button from the installation panel will reload the mod chooser.  That should no longer be necessary.

OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
pchote: You can / should adjust the mod chooser to open these as a Window so that you don't need to manually hide/show them.

pchote:
I volunteer to use a YAML polish pass, so you can focus on the code nits for now.

mods/cnc/chrome.yaml
pchote: You can confirm that it is used there by checking the lobby yaml.

penev92: Yeah, didn't think of map download. Will bring it back.

OpenRA.Mods.Common/InstallUtils.cs
LipkeGu: Line 72 can also be removed...

penev92:
Rebased and addressed all requests, minus the button sizes.

OpenRA.Game/ModMetadata.cs
```diff
@@ -46,6 +47,9 @@ public class ModMetadata
 				var mod = FieldLoader.Load<ModMetadata>(nd["Metadata"]);
 				mod.Id = m;
 
+				if (nd.ContainsKey("ContentInstaller"))
+					mod.Content = FieldLoader.Load<ContentInstaller>(nd["ContentInstaller"]);
```
pchote: It might pay to wrap this in a try/catch.  You don't want to kill the whole game if someone has one bogus mod installed.

OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
```diff
@@ -40,27 +40,13 @@ public void StartGame(Arguments args)
 			Ui.ResetAll();
 			Game.Settings.Save();
 
-			// Check whether the mod content is installed
-			// TODO: The installation code has finally been beaten into shape, so we can
-			// finally move it all into the planned "Manage Content" panel in the modchooser mod.
 			var installData = Game.ModData.Manifest.Get<ContentInstaller>();
-			var installModContent = !installData.TestFiles.All(f => GlobalFileSystem.Exists(f));
-			var installModMusic = args != null && args.Contains("Install.Music");
+			var isModContentInstalled = installData.TestFiles.All(f => GlobalFileSystem.Exists(f.Substring(f.LastIndexOf('/') + 1)));
```
pchote: `Path.GetFileName`?

OpenRA.Game/ModMetadata.cs
penev92: Actually the `if` should take care of that. I also just reverted locally the addition of `ContentInstaller:` to `modchooser\mod.yaml`, because it's no longer needed (was needed in the earlier stages of this branch).

OpenRA.Mods.Common/Widgets/Logic/Installation/DownloadPackagesLogic.cs
```diff
@@ -21,24 +20,30 @@ namespace OpenRA.Mods.Common.Widgets.Logic
...
+			var text = panel.Get<LabelWidget>("TITLE").Text;
+			panel.Get<LabelWidget>("TITLE").Text = text.F(modName);
```
pchote: This adds an edge case that will complicate the future translation string extraction (if this ever happens).  Can you hardcode the format string in the logic file instead for now? (this is what we do everywhere else).

OpenRA.Mods.Common/Widgets/Logic/Installation/DownloadPackagesLogic.cs
pchote: You missed this in the last pass.

OpenRA.Mods.Common/Widgets/Logic/Installation/InstallLogic.cs
```diff
@@ -17,31 +15,28 @@ namespace OpenRA.Mods.Common.Widgets.Logic
...
+				{ "afterInstall", () => { Game.InitializeMod(modId, new Arguments()); } },
+				{ "mirrorListUrl", mirrorListUrl },
+				{ "modId", modId }
 			};
 
-			panel.Get<ButtonWidget>("DOWNLOAD_BUTTON").OnClick = () =>
-				Ui.OpenWindow("INSTALL_DOWNLOAD_PANEL", widgetArgs);
+			var mod = ModMetadata.AllMods[modId];
+			var text = panel.Get<LabelWidget>("DESC1").Text;
+			panel.Get<LabelWidget>("DESC1").Text = text.F(mod.Title);
```
pchote: Likewise here.

OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
```diff
@@ -148,10 +155,29 @@ void SelectMod(ModMetadata mod)
 			var selectedIndex = Array.IndexOf(allMods, mod);
 			if (selectedIndex - modOffset > 4)
 				modOffset = selectedIndex - 4;
+
+			loadButton.Text = modInstallStatus[mod] ? "Load Mod" : "Install Assets";
```
pchote: The screen that pops up after "Install Assets" says "Game assets not found!". This wording assumed that we kept the "Load Mod" button text.  Can you please either keep the button text, or change the title of the dialog?

OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
```diff
@@ -148,10 +155,29 @@ void SelectMod(ModMetadata mod)
...
 		}
 
 		void LoadMod(ModMetadata mod)
 		{
+			if (!modInstallStatus[mod])
+			{
+				var contentData = mod.Content;
```
pchote: Just use `mod.Content` directly below.

OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
pchote: My preference is to change the dialog title to "Install Assets", which then matches "Install Music".

OpenRA.Game/ModMetadata.cs
pchote: The `ContentInstaller` block may exist, but have bogus content that makes FieldLoader throw.

pchote:
A handful of UI nits:
* Please reduce the height of the buttons on the main mod chooser panel so that they are 32px everywhere.
* The progress bar bar should be pure grey, without the inner red border (I can help with art).
* The layout of the download (and I assume also the copy from CD) dialog is bogus.  I suggest making the progress bar thinner to give the extra space you need (see point above).
* Please increase the height of all of these dialogs (keeping them equal) so that there is an equal margin to the right and below the buttons.
* Please take 4d740e897bf1f6ec05093eb221ffd6ad9f8d923e and then set an appropriate BaseLine to center the text in the mod chooser buttons.
* You can take 2affae44beccb4852351be0bcbb5c66a13d19b47 to fix the broken overlaps that happen at low percentages.


screen shot 2015-09-08 at 21 24 04

LipkeGu:
#9136 was also a try to make the progressbars more easier to configure :)

pchote:
b25aafa96d766109355468e0f45d257d7b7593bd gives you a new progress bar to play with.

screen shot 2015-09-08 at 22 22 15

penev92:
That should be all now.

pchote:
There are a few more nits:

* Use the download ctor that doesn't need a temporary on-disk file (see discussion above).
* You forgot to set ButtonBaseLine: 1 for the modchooser mod.
* The download panel layout is still off: move the text down by a few px to match the second line in the other panels, and then center the progress bar in the gap.
* The button margins and the horizontal rule margin on the main browser panel don't match the dialogs.  Fixing this means tweaking the whole layout of the browser panel, so I won't insist on fixing that now.

penev92:
Changed as requested.

Mailaender:
You added new stylecop errors.

mods/modchooser/metrics.yaml
```diff
@@ -3,10 +3,12 @@
 Metrics:
 	ButtonDepth: 0
 	ButtonFont: Bold
+	ButtonBaseLine: 1
 	ButtonTextColor: 255,255,255
 	ButtonTextColorDisabled: 128,128,128
 	ButtonTextContrast: false
 	ButtonTextContrastColor: 0,0,0
+	ButtonBaseLine: 0
```
pchote: This resets the BaseLine back to 0!

pchote:
Confirmed that my requests above have been fixed, except for the `ButtonBaseLine` (again!).  :+1: after fixing that and the stylecop errors.  Good work!

penev92:
OK they should be centered now.

pchote:
Good job :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/1442fb34743e0196e966b725f10a680261512f80

PR #9270, opened 2015-09-05
Made BlocksProjectiles upgradable

reaperrr:
Needed for TS firestorm defense and laser fences.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Bullet.cs
+++ b/OpenRA.Mods.Common/Effects/Bullet.cs
@@ -167,12 +167,8 @@ namespace OpenRA.Mods.Common.Effects
 			if (info.ContrailLength > 0)
 				contrail.Update(pos);
 
-			var cell = world.Map.CellContaining(pos);
-			var height = world.Map.DistanceAboveTerrain(pos);
-
-			var shouldExplode = height.Length <= 0 // Hit the ground
-				|| ticks++ >= length // Flight length reached/exceeded
-				|| BlockedByActor(world, pos); // Hit a wall or other blocking obstacle
+			var shouldExplode = ticks++ >= length // Flight length reached/exceeded
+				|| (info.Blockable && BlocksProjectiles.AnyBlockingActorAt(world, pos)); // Hit a wall or other blocking obstacle
 
 			if (shouldExplode)
 				Explode(world);
@@ -211,12 +207,6 @@ namespace OpenRA.Mods.Common.Effects
 			world.AddFrameEndTask(w => w.Remove(this));
 
 			args.Weapon.Impact(Target.FromPos(pos), args.SourceActor, args.DamageModifiers);
-		}
-
-		bool BlockedByActor(World world, WPos pos)
-		{
-			return info.Blockable && world.ActorMap.GetUnitsAt(world.Map.CellContaining(pos))
-					.Any(a => a.TraitsImplementing<BlocksProjectiles>().Any(Exts.IsTraitEnabled));
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Effects/Bullet.cs
+++ b/OpenRA.Mods.Common/Effects/Bullet.cs
@@ -16,6 +16,7 @@ using OpenRA.Effects;
 using OpenRA.GameRules;
 using OpenRA.Graphics;
 using OpenRA.Mods.Common.Graphics;
+using OpenRA.Mods.Common.Traits;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Effects
@@ -171,7 +172,7 @@ namespace OpenRA.Mods.Common.Effects
 
 			var shouldExplode = height.Length <= 0 // Hit the ground
 				|| ticks++ >= length // Flight length reached/exceeded
-				|| (info.Blockable && world.ActorMap.GetUnitsAt(cell).Any(a => a.Info.HasTraitInfo<IBlocksProjectilesInfo>())); // Hit a wall or other blocking obstacle
+				|| BlockedByActor(world, pos); // Hit a wall or other blocking obstacle
 
 			if (shouldExplode)
 				Explode(world);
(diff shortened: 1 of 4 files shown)


OpenRA.Mods.Common/Effects/Bullet.cs
```diff
@@ -164,8 +165,11 @@ public void Tick(World world)
 			if (info.ContrailLength > 0)
 				contrail.Update(pos);
 
-			if (ticks++ >= length || (info.Blockable && world.ActorMap
-				.GetUnitsAt(world.Map.CellContaining(pos)).Any(a => a.HasTrait<IBlocksProjectiles>())))
+			if (ticks++ >= length
+				|| (info.Blockable && world.ActorMap
+					.GetUnitsAt(world.Map.CellContaining(pos))
+					.Any(a => a.TraitsImplementing<BlocksProjectiles>()
+					.Any(b => !b.IsTraitDisabled))))
```
atlimit8: Please add a level of indentation to show that it is inside the previous Any.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -201,7 +201,10 @@ public void Tick(World world)
 			var shouldExplode = (pos.Z < 0) // Hit the ground
 				|| (dist.LengthSquared < info.CloseEnough.LengthSquared) // Within range
 				|| (info.RangeLimit != 0 && ticks > info.RangeLimit) // Ran out of fuel
-				|| (info.Blockable && world.ActorMap.GetUnitsAt(cell).Any(a => a.HasTrait<IBlocksProjectiles>())) // Hit a wall or other blocking obstacle
+				|| (info.Blockable && world.ActorMap
+					.GetUnitsAt(world.Map.CellContaining(pos))
+					.Any(a => a.TraitsImplementing<BlocksProjectiles>()
+					.Any(b => !b.IsTraitDisabled))) // Hit a wall or other blocking obstacle
```
atlimit8: Please add a level of indentation to show that it is inside the previous Any as also requested above.

reaperrr:
Updated.

OpenRA.Mods.Common/Effects/Bullet.cs
```diff
@@ -164,8 +165,11 @@ public void Tick(World world)
 			if (info.ContrailLength > 0)
 				contrail.Update(pos);
 
-			if (ticks++ >= length || (info.Blockable && world.ActorMap
-				.GetUnitsAt(world.Map.CellContaining(pos)).Any(a => a.HasTrait<IBlocksProjectiles>())))
+			if (ticks++ >= length
```
Unit158: You'll have to check two times per tick now. The current location and the next location.

reaperrr: Why and how?

Unit158: it actually shouldn't matter in practice due to order latency, so never mind :)

OpenRA.Mods.Common/Effects/Bullet.cs
```diff
@@ -164,8 +165,11 @@ public void Tick(World world)
...
+				|| (info.Blockable && world.ActorMap
+					.GetUnitsAt(world.Map.CellContaining(pos))
+					.Any(a => a.TraitsImplementing<BlocksProjectiles>()
+						.Any(b => !b.IsTraitDisabled))))
```
pchote: This is a bit of a monster to have inside a conditional check, so could you please pull this out into an `IsBlockedByActor()` method?

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -201,7 +201,10 @@ public void Tick(World world)
 			var shouldExplode = (pos.Z < 0) // Hit the ground
 				|| (dist.LengthSquared < info.CloseEnough.LengthSquared) // Within range
 				|| (info.RangeLimit != 0 && ticks > info.RangeLimit) // Ran out of fuel
-				|| (info.Blockable && world.ActorMap.GetUnitsAt(cell).Any(a => a.HasTrait<IBlocksProjectiles>())) // Hit a wall or other blocking obstacle
+				|| (info.Blockable && world.ActorMap
```
pchote: Likewise here

pchote:
Needs a rebase, sorry.

reaperrr:
@atlimit8 Is there or will there be something to replace

`.Any(a => a.TraitsImplementing<BlocksProjectiles>().Any(b => !b.IsTraitDisabled))`

with an Info equivalent?

atlimit8:
If you mean check if the actor type has a `BlocksProjectiles` trait that is enabled by default, then yes; however, that (being enabled/disabled) is part of the traits state (may change).

In an upcoming PR I plan on allowing something like:
```
.WhereType(projectileBlockerTypes).Any(a => a.TraitsImplementing<BlocksProjectiles>().Any(Exts.IsTraitEnabled))
```
where `projectileBlockerTypes` is a cached `HashSet<uint>` of `ActorTypeID`s : 
```
projectileBlockerTypes = self.World.Map.Rules.ActorsWhereTraitInfo<BlocksProjectilesInfo>();
```

That would allow for filtering out all non-blockers with an amortized cost of O(N) where N is the number of actors so the cost (O(N * log T) where N is the number of possible blockers and T is the number of trait pairs in the dictionary) of the
```
.Any(a => a.TraitsImplementing<BlocksProjectiles>().Any(Exts.IsTraitEnabled))
```
will only be applied for actors that might block (usually 0, otherwise generally 1).

Basically, that should change the cost from O(N * log T) to an amortized O(N + B * log T) where N is the number of actors, B is the number of possible blockers and T is the number of trait pairs in the dictionary. 


I am working on a branch that will add `WhereOwner*` filters to replace common delegates and then basing the aforementioned branch on that using comment feedback to iron out naming style.

atlimit8:
In case you missed it in my previous post:
You should replace `b => !b.IsTraitDisabled` with `Exts.IsTraitEnabled`

reaperrr:
> You should replace `b => !b.IsTraitDisabled` with `Exts.IsTraitEnabled`

Done.

RoosterDragon:
Is the duplicated ```BlockedByActor``` method something we can move into one place, maybe on ```BlocksProjectiles```?

abcdefg30:
Looks good to me. :+1:

penev92:
Maybe have the `if` (both of them) as
```
if (ticks++ >= length || (info.Blockable && AnyBlockersAt(world, pos)))
```

and move `BlockedByActor()` to `BlocksProjectiles` as
`public static bool AnyBlockersAt(World world, WPos pos)` ?

reaperrr:
Rebased and updated.

penev92:
Works :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/8e31fb4bbbfbb7ef7a202a821fca4d3dc02f25e6...1f39ddebb0b97deb8495003e4983a93379f9ef08)

PR #9272, opened 2015-09-05
Remove the broken actor stats export command.

pchote:
The DPS values are calculated / reported for only the first warhead on the last defined weapon on an actor.  This means that it reports the garrisoned weapon stats for most infantry, the elite weapon stats for many TS actors, and arbitrary values for units that define multiple weapons (rocket troopers, mamoth tanks, etc).  These values aren't very meaningful.  We now also support upgradeable armors.

This appears to have been broken for quite some time, so the fact that nobody has discovered or reported it before now shows that it is probably not used or worth the (significant) effort of making it work properly.


penev92:
I can see how that would bitrot very quickly. I am also unaware of any use for it, so :+1:  I guess.

clemty:
isn't this used by http://aaronfranks.com/openra_chart/ ?
(or how else are these values generated?)

pchote:
@clemty: That appears to use an automated yaml-to-json converter, and then parse the values itself: http://aaronfranks.com/openra_chart/js/data/release-20150424.js

Mailaender:
I build it for @psydev who then vanished and probably also never used it. Thinking about how much time it cost me to create it satisfying everyone and now just throwing it in the garbage is disgusting, but I guess we have to move along.

PR #9279, opened 2015-09-05
More HashSet trait fields & Lint support

atlimit8:
More `string[]` => `HashSet<string>` for trait fields with Lint reference support for `HashSet<string>` trait fields.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Explodes.cs
+++ b/OpenRA.Mods.Common/Traits/Explodes.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Mods.Common.Warheads;
 using OpenRA.Traits;
@@ -31,7 +32,7 @@ namespace OpenRA.Mods.Common.Traits
 		public readonly int Chance = 100;
 
 		[Desc("DeathType(s) to apply upon explosion.")]
-		public readonly string[] DeathType = null;
+		public readonly HashSet<string> DeathType = new HashSet<string>();
 
 		public object Create(ActorInitializer init) { return new Explodes(this); }
 	}
@@ -51,7 +52,7 @@ namespace OpenRA.Mods.Common.Traits
 				return;
 
 			var warhead = e.Warhead as DamageWarhead;
-			if (info.DeathType != null && warhead != null && !warhead.DamageTypes.Overlaps(info.DeathType))
+			if (warhead != null && !warhead.DamageTypes.Overlaps(info.DeathType))
 				return;
 
 			var weaponName = ChooseWeaponForExplosion(self);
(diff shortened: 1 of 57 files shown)


OpenRA.Mods.Common/Warheads/LeaveSmudgeWarhead.cs
```diff
@@ -38,7 +38,7 @@ public override void DoImpact(Target target, Actor firedBy, IEnumerable<int> dam
 			// Draw the smudges:
 			foreach (var sc in allCells)
 			{
-				var smudgeType = world.Map.GetTerrainInfo(sc).AcceptsSmudgeType.FirstOrDefault(t => SmudgeType.Contains(t));
+				var smudgeType = world.Map.GetTerrainInfo(sc).AcceptsSmudgeType.Intersect(SmudgeType).FirstOrDefault();
```
RoosterDragon: This first is better as Intersects with create a new set internally for elements in the source collection. You could lose the lambda though ```FirstOrDefault(SmudgeType.Contains)```.

RoosterDragon:
Please don't change```.Length > 0``` or ```.Count > 0``` to ```.Any()``` as the former are faster and just as obvious.

atlimit8:
Changed.

RoosterDragon:
:+1:

Mailaender:
Thanks.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f3bf0e36cbd3ded7b1592e335a977db0477b09a6...f2353d79b565531dd4f64fe62fb08609fff09514)

reaperrr:
This broke the `Explodes` trait.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -51,7 +52,7 @@ public void Killed(Actor self, AttackInfo e)
 				return;
 
 			var warhead = e.Warhead as DamageWarhead;
-			if (info.DeathType != null && warhead != null && !warhead.DamageTypes.Overlaps(info.DeathType))
+			if (warhead != null && !warhead.DamageTypes.Overlaps(info.DeathType))
```
reaperrr: This isn't right, `info.DeathType != null` needs a replacement (`.Count > 0`, I assume).

atlimit8: `HashSet.Overlaps(...)` only return true if there is at least one element in the intersection ([HashSet(T).Overlaps Method (System.Collections.Generic)](https://msdn.microsoft.com/en-us/library/bb355623%28v=vs.110%29.aspx)). That is `HashSet.Overlaps(...)` is always false if either is empty (`.Count == 0`).

reaperrr: The problem is that the whole check - including the return - is supposed to be skipped if no DeathType is specified (see how in the old code, DeathType == null would skip the rest of the check including the return). 
The way the check is now, it will instead `return` if DeathType is empty (which is the default), because there's nothing to overlap.

Basically, this changes the default behavior from "ignore this check and continue with the following lines" to "return".

atlimit8: I see, its negated.
OK, I'll fix it.

atlimit8: Pushed #9311.

PR #8162, opened 2015-05-14
Remove bounties from d2k

abcdefg30:
As requested by @pchote and @penev92 on IRC.

Btw, this has actually :-1: from me, but it's needed for #8117.

PR #9282, opened 2015-09-06
Removed bounties from D2k

reaperrr:


Unit158:
:+1:

Unit158:
By posting here with a :-1: and causing the same stuff that happened with #8162, you waive the right to not get beat with a roughly cylindrical stick no longer than one meter and with a radius no larger than an inch.

penev92:
Ignoring the comment above, which, despite its playful tone, has no place on GitHub, I guess if we're aiming for a "full" recreation that would be attractive to the original game's fans, bounties should go.
At the very least until the dusts settles from all the other rebalancing.
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f2353d79b565531dd4f64fe62fb08609fff09514...9bdd9ae22e0a5d43cda0d256ebde839441202aa9)

PR #9284, opened 2015-09-06
TD: Add additional desert tiles (work originally by psydev)

clemty:
from https://github.com/psydev/OpenRA/commits/desert (last updated 2014-05-25):

> There is some art in RA Desert tileset that is not in TD. It looks as though it belongs
> to the original, but not sure.
> They look really good, so including them in TD. There were some "pink
> templates" (lacking image) which now have an image (br5.des and
> br10.des). Were they missing? They look like they belong in the original
> game.

this looks like a worthwhile addition and doesn't look as if it was ever pulled into the main OpenRA code; it also doesn't look as if psydev is still working on it.

I think it is ok I submit this (I haven't been in contact with psydev)

beforeafter


Mailaender:
To clarify: RA has a few more tiles than TD. So when DESERT was imported from TD https://github.com/OpenRA/OpenRA/pull/2445 our artists filled the gaps trying to retain the original style.

reaperrr:
:+1:

obrakmann:
:+1:

penev92:
Those two :+1:s are good enough to me.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d92454dfb6c56c256fbab5978f8dc791c6caf30a...8a2b47da40c6e99b3a8e79f3cee919c5dc764fca)

PR #9288, opened 2015-09-06
Fix ActionQueue ordering

### LINKED ISSUE #9286: Game.RunAfterTick & RunAfterDelay are badly sequenced
Consider the following:

```c#
Game.RunAfterDelay(0, () => Console.WriteLine("A0"));
Game.RunAfterDelay(0, () => Console.WriteLine("B0"));
Game.RunAfterDelay(0, () => Console.WriteLine("C0"));
Game.RunAfterDelay(1000, () => Console.WriteLine("A1"));
Game.RunAfterDelay(1000, () => Console.WriteLine("B1"));
Game.RunAfterDelay(1000, () => Console.WriteLine("C1"));
Game.RunAfterDelay(2000, () => Console.WriteLine("A2"));
Game.RunAfterDelay(2000, () => Console.WriteLine("B2"));
Game.RunAfterDelay(2000, () => Console.WriteLine("C2"));
```

This produces:
```
B0
C0
A0
C1
A1
B1
B2
A2
C2
```

whereas I would expect (and prefer) queue-like behaviour:
```
A0
B0
C0
A1
B1
C1
A2
B2
C2
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26485762-game-runaftertick-runafterdelay-are-badly-sequenced?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RoosterDragon:
Fixes #9286.

This enforces a well defined ordering for ActionQueue to ensure it runs actions in the the order one would expect.

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -384,9 +384,9 @@ namespace OpenRA
 
 		// Note: These delayed actions should only be used by widgets or disposing objects
 		// - things that depend on a particular world should be queuing them on the worldactor.
-		static ActionQueue delayedActions = new ActionQueue();
-		public static void RunAfterTick(Action a) { delayedActions.Add(a); }
-		public static void RunAfterDelay(int delay, Action a) { delayedActions.Add(a, delay); }
+		static volatile ActionQueue delayedActions = new ActionQueue();
+		public static void RunAfterTick(Action a) { delayedActions.Add(a, Game.RunTime); }
+		public static void RunAfterDelay(int delayMilliseconds, Action a) { delayedActions.Add(a, Game.RunTime + delayMilliseconds); }
 
 		static void TakeScreenshotInner()
 		{
@@ -487,7 +487,7 @@ namespace OpenRA
 
 		static void LogicTick()
 		{
-			delayedActions.PerformActions();
+			delayedActions.PerformActions(Game.RunTime);
 
 			if (OrderManager.Connection.ConnectionState != lastConnectionState)
 			{
--- a/OpenRA.Game/Primitives/ActionQueue.cs
+++ b/OpenRA.Game/Primitives/ActionQueue.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System;
+using System.Collections.Generic;
 
 namespace OpenRA.Primitives
 {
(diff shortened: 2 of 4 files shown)


OpenRA.Game/Primitives/ActionQueue.cs
```diff
@@ -17,37 +18,56 @@ namespace OpenRA.Primitives
...
+		readonly List<DelayedAction> actions = new List<DelayedAction>();
 
-		public void Add(Action a) { Add(a, 0); }
-		public void Add(Action a, int delay)
+		public void Add(Action a, int desiredTime)
 		{
-			lock (syncRoot)
-				actions.Add(new DelayedAction(a, Game.RunTime + delay));
+			if (a == null)
+				throw new ArgumentNullException("a");
```
Unit158: D:

OpenRA.Test/OpenRA.Game/ActionQueueTest.cs
```diff
@@ -0,0 +1,42 @@
...
+			queue.Add(() => list.Add(2), 0);
+			queue.Add(() => list.Add(3), 0);
+			queue.Add(() => list.Add(9), 2);
+			queue.Add(() => list.Add(5), 1);
+			queue.Add(() => list.Add(6), 1);
+			queue.PerformActions(1);
+			queue.PerformActions(2);
+			queue.PerformActions(3);
+			if (!list.SequenceEqual(new[] { 1, 2, 3, 4, 5, 6, 7, 8, 9 }))
+				Assert.Fail("Actions were not perform in the correct order. Actual order was: " + string.Join(", ", list));
```
pchote: perform &rarr; performed.

OpenRA.Game/Primitives/ActionQueue.cs
Mailaender: The first string is the name of the argument. This will automatically translate into a suitable error message, because of the exception type. http://www.dotnetperls.com/argumentexception

OpenRA.Game/Game.cs
```diff
@@ -383,9 +383,9 @@ static string ChooseShellmap()
 
 		// Note: These delayed actions should only be used by widgets or disposing objects
 		// - things that depend on a particular world should be queuing them on the worldactor.
-		static ActionQueue delayedActions = new ActionQueue();
-		public static void RunAfterTick(Action a) { delayedActions.Add(a); }
-		public static void RunAfterDelay(int delay, Action a) { delayedActions.Add(a, delay); }
+		static volatile ActionQueue delayedActions = new ActionQueue();
+		public static void RunAfterTick(Action a) { delayedActions.Add(a, Game.RunTime); }
+		public static void RunAfterDelay(int delay, Action a) { delayedActions.Add(a, Game.RunTime + delay); }
```
pchote: Minor nit/request: Could you please rename `delay` &rarr; `delayMilliseconds` or similar.  I had to check to the code because I initially assumed that delay would be in ticks.

pchote:
Code makes sense, and verified that the surrender &rarr; exit game flow (which is the main thing using RunAfterDelay) works correctly.  :+1:

OpenRA.Game/Game.cs
RoosterDragon: Added.

Mailaender:
Thanks!

PR #7559, opened 2015-02-27
Added Campaign for CnC-mod

Sithil-F:
Hello,
as part of a student project we worked on OpenRA (the same project as the mission-teams)
Our team implemented a campaign menu for the CnC:TD mod.
It features:
    - a new entry in the singleplayer menu
    - Faction selection
        -- the nostalgic transmission selection for new games
        -- continue with the campaign of GDI or Nod
    - playing missions of one faction in chronological order
    - a selection if there are alternative missions (i.e. a-,b-,c-missions)
        -- if multiple path exist an overview-map with all paths is shown (image is taken out of the a-mission-folder)
    - country information, if specified in the mission yaml (optional)
    - saving progress after each successful mission
        -- nod and gdi progress is tracked independently
    - a congratulation-message after the last available mission was won

ToDos:
- [x] make variables readonly if possible
- [x] fix incorrect namespacing
- [x] add a continue button after sucessful mission
- [x] skip to faction-selection if no saves exist
- [x] use "var" where possible
- [x] make fields "const" where possible
- [x] better support for branching in campaigns
- [x] split and move code from CampaignWorldLogic's constructor
- [x] keep the codechanges inside OpenRA.Mods.Cnc or OpenRA.Mods.Common
- [ ] use MapCache instead of manual preview
- [ ] implement SaveProcessFlag
- [ ] use Lua API for setting/querying campaign status
- [ ] Generalisation for use in other mods (low priority)

PR #9289, opened 2015-09-06
Ported CnC-Campaign from our student project (now generalised)

Sithil-F:
Hello,
we are back with our campaign menu, but since we changed a lot of things, we decided to open a new pull request.
The original PR can be found here(https://github.com/OpenRA/OpenRA/pull/7559).
Here are the changes/additions from the last PR:
- Faction selection
  - the .vqa files are not mandatory anymore and we can remove them, if that is preferred
  - a custom recreation of the transmission-select screen if no vqa is found
- alternative mission selection (Campaign-Preview, i.e. a-,b-,c-missions)
  - the mission following a victory is declared in the yaml of the map (“Next Mission”), which allows for branching campaign-paths
  - If no preview-picture is available the faction logo appears
- Generalization, so it should work for all mods
  - currently active in the cnc and ra mods
- Other requests
  - use MapCache instead of manual preview
  - implement SaveProgressFlag
  - used OnPlayerWon Event to save the campaign progress


Unit158:
You will need to split this up into manageable commits for the reviewers.

pchote:
Adding new fields to Map and new game-level plumbing shouldn't be needed to support this.  We have created the mission.yaml files as a stub for holding this kind of info, and they should be expanded on instead.

pchote:
The static CampaignProgress doesn't feel like a good fit.  See #7213 for some earlier discussion about recording our campaign progression.  I still think that the most flexible way to implement this will be via a Lua API for saving arbitrary data, which then lets us implement mission-objective specific progression (e.g. the C&C commando mission, which branched based on which structure you destroyed), and persisting information such as base layout across missions.

pchote:
Can we please keep the windowed FMV playback (using buttons off the bottom of the campaign info panel) instead of reverting to the one-shot fullscreen videos?  We moved away from those because they do not look very good played at full size on modern high-resolution monitors, and because it is useful to be able to repeat the briefing if you miss something the first time.

Sithil-F:
Thank you for your feedback. We will take care of the issues.
However, we will need help with the Lua API. 
(Could someone please mark this with the “PR: Help wanted” label?)
We will try to get the other 2 changes done within a week.

One Question about the outsourcing of the campaign paths:
So the possible paths should be defined in the missions.yaml, right?
Is the Country Information acceptable where it is, or should we move it too?

OpenRA.Game/Manifest.cs
```diff
@@ -70,7 +70,7 @@ public class Manifest
 		readonly string[] reservedModuleNames = { "Metadata", "Folders", "MapFolders", "Packages", "Rules",
 			"Sequences", "VoxelSequences", "Cursors", "Chrome", "Assemblies", "ChromeLayout", "Weapons",
 			"Voices", "Notifications", "Music", "Translations", "TileSets", "ChromeMetrics", "Missions",
-			"ServerTraits", "LoadScreen", "LobbyDefaults", "Fonts", "TileSize", "MaximumTerrainHeight",
+			"FactionMedia",	"Congratulations",	"ServerTraits", "LoadScreen", "LobbyDefaults", "Fonts", "TileSize", "MaximumTerrainHeight",
```
abcdefg30: Seems like you're using tabs instead of spaces after the commas.

OpenRA.Mods.Cnc/Widgets/Logic/CncMainMenuLogic.cs
```diff
@@ -1,4 +1,4 @@
-#region Copyright & License Information
+﻿#region Copyright & License Information
```
abcdefg30: Can you undo the changes made to this file, please?

OpenRA.Mods.Common/CampaignProgress.cs
```diff
@@ -0,0 +1,119 @@
+using System;
```
abcdefg30: The "Copyright & License Information" region is missing here.

OpenRA.Mods.Common/Widgets/Logic/CampaignCongratulationLogic.cs
```diff
@@ -0,0 +1,87 @@
...
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+
```
abcdefg30: Please remove this new line here.

OpenRA.Mods.Common/Widgets/Logic/CampaignFactionLogic.cs
```diff
@@ -0,0 +1,312 @@
...
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+
```
abcdefg30: Here too.

penev92:
>However, we will need help with the Lua API. 

What kind of help?

>So the possible paths should be defined in the missions.yaml, right?

Yes.

penev92:
Bump.

Freddiie:
Hello there,
we will begin in the coming days the implementation work on the Lua API. It would be helpful to know what kind of files and components to do with leaving a map in order to spare us familiarization time. Please have patience.

penev92:
I'm going to close this now. Do reopen it if you find the time to finish it.

obrakmann:
FWIW, I don't think this PR should be re-openend as-is. It's a massive pain to review such a huge thing. Especially the UI changes alone will take ages to get reviewed and be corrected. I suggest to pick and choose small parts of this and then pr them one-by-one.

PR #9290, opened 2015-09-06
Dispose earlier in FixClassicTilesets.

RoosterDragon:
Ensure we dispose the stream we opened right away to avoid leaving it open when an exception occurs.

Discovered by Coverity.


Mailaender:
Not so serious in command line utility stuff that exits when done, but also doesn't hurt I guess. :+1:

penev92:
:+1:

PR #9292, opened 2015-09-06
Fix an infinite loop that occurs when a silo is captured

obrakmann:
Introduced in #9189's 69d062495cd46fff35de1fe229e16cab53b797d2


Mailaender:
*ugh* :+1:

Mailaender:
Thanks.

PR #9294, opened 2015-09-06
Remove Shroud.IsVisibleTest and IsExploredTest

RoosterDragon:
These are no longer any faster than the regular checks, so there's no need to keep them.


Mailaender:
:+1: Trusting you on this. Is there a way to measure it without VS?

RoosterDragon:
If you follow the chain of calls, you'll find the delegate it returns just calls the method anyway! It's just a roundabout way of doing the same thing.

It used to return a more specific delegate that skipped some checks, but we lost that when projected shroud came about and we don't really need it any more.

penev92:
:+1:

PR #9296, opened 2015-09-06
Speed up Player.CanViewActor

RoosterDragon:
Create Actor.CanBeViewedByPlayer and simply call this instead. The actor can cache all trait lookups on construction to avoid them being repeated for every visibility check.

This approximately doubles the speed of Player.CanViewActor, which is useful since it gets called a lot from the radar widget (which checks all actors with a radar signature) and when "Always Show Health Bars" is enabled (which checks all selectable actors).

### DIFF
--- a/OpenRA.Game/Actor.cs
+++ b/OpenRA.Game/Actor.cs
@@ -71,6 +71,8 @@ namespace OpenRA
 		readonly IRenderModifier[] renderModifiers;
 		readonly IRender[] renders;
 		readonly IDisable[] disables;
+		readonly IVisibilityModifier[] visibilityModifiers;
+		readonly IDefaultVisibility defaultVisibility;
 
 		internal Actor(World world, string name, TypeDictionary initDict)
 		{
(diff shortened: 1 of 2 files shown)


Phrohdoh:
Is this a case that `Actor` really cares about?
Note the `TODO` over `ChangeOwner`.

If this fits in `Actor` and isn't an abstraction, then okay.

Mailaender:
This is generic enough so it doesn't belong to ActorExts if that is your concern. :+1: Trait caching is always good especially in hot paths. I tested this and can't spot any ordering related trait lookup regression or any other behavioral changes. :white_check_mark:

RoosterDragon:
@Phrohdoh: I see your point in the general case, but for this specific case I figure either ```Actor```, ```Player``` or some visibility specific class should handle it - so I'm basically just choosing to shift stuff into actor for the performance benefits. Hopefully that's reasonable.

penev92:
Looks good to me :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/0dad6a3eec01f65fd61602df64a4d42fb785cbde...19aecb76f61968e0cbfe1c6009a96c2fac74163a)

PR #9298, opened 2015-09-07
Create abstract class AffectsShroud[Info] with RevealsShroud core logic also used by CreatesShroud

### LINKED ISSUE #9297: CreatesShroud inherits from RevealsShroud, but are functionally disjoint.
`CreatesShroud` inherits from `RevealsShroud`, but every actor with `CreatesShroud` also has  `RevealsShroud` in the MiniYaml (`GAP` in `mods/ra/rules/structures.yaml:78-83`, `MGG` in `mods/ra/rules/vehicles.yaml:504-507`).

Separating the two would be the most flexible option.

`CreatesShroud` *currently* derives from `RevealsShroud`:
```c#
	public class CreatesShroud : RevealsShroud
	{
		public CreatesShroud(Actor self, CreatesShroudInfo info)
			: base(self, info)
 		{
```

I'm proposing that that be change to:
```c#
	public class CreatesShroud : AffectsShroud
	{
		public CreatesShroud(Actor self, CreatesShroudInfo info)
			: base(self, info)
 		{
```
where `AffectsShroud` is an absract base class for `RevealsShroud` & `CreatesShroud`.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26493508-createsshroud-inherits-from-revealsshroud-but-are-functionally-disjoint?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

atlimit8:
Closes  #9297.

### DIFF
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -695,6 +695,7 @@
     <Compile Include="Traits\Modifiers\HiddenUnderShroud.cs" />
     <Compile Include="Lint\CheckDefaultVisibility.cs" />
     <Compile Include="Traits\Modifiers\AlwaysVisible.cs" />
+    <Compile Include="Traits\AffectsShroud.cs" />
     <Compile Include="Traits\CreatesShroud.cs" />
     <Compile Include="Traits\RevealsShroud.cs" />
     <Compile Include="Lint\CheckRevealFootprint.cs" />
--- a/OpenRA.Mods.Common/Traits/AffectsShroud.cs
+++ b/OpenRA.Mods.Common/Traits/AffectsShroud.cs
@@ -1,0 +1,101 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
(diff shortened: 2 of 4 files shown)


penev92:
Please either put all 3 classes in 1 file or put each in its own file. The current separation doesn't make sense any more.

atlimit8:
Split into 3 files now.

atlimit8:
`git diff` seems to have trouble seeing that I renamed `OpenRA.Mods.Common/Traits/RevealsShroud.cs` => ` OpenRA.Mods.Common/Traits/AffectsShroud.cs` and created a new `OpenRA.Mods.Common/Traits/RevealsShroud.cs`.

RoosterDragon:
It isn't obvious that ```addCellsToPlayerShroud```, ```removeCellsFromPlayerShroud``` and ```isDisabled``` need to be assigned.

Try adding a constructor argument in ```AffectsShroud``` for them, or maybe change them into abstract methods instead.

atlimit8:
Changed.

RoosterDragon:
:+1: Looks reasonable to me and revealing and gap gens still work.

pchote:
Looks good, and confirmed that gap gens and sight still work :+1:.

PR #9299, opened 2015-09-07
TD: support TriggerPowerOutage()

### LINKED ISSUE #9291: TD: TriggerPowerOutage() not supported
This seems to be a RA-specific LUA function, but would be handy for TD C&C64 GDI01: power down the Obelisk for a few seconds after the Powerplants are hit by the airstrike

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26489062-td-triggerpoweroutage-not-supported?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

clemty:
fixes #9291

### DIFF
--- a/mods/cnc/rules/structures.yaml
+++ b/mods/cnc/rules/structures.yaml
@@ -117,6 +117,7 @@ NUKE:
 	Bib:
 	Power:
 		Amount: 100
+	AffectedByPowerOutage:
 	ScalePowerWithHealth:
 
 NUK2:
@@ -142,6 +143,7 @@ NUK2:
 	Bib:
 	Power:
 		Amount: 200
+	AffectedByPowerOutage:
 	ScalePowerWithHealth:
 
 PROC:


Mailaender:
You should add that to mission specific rules as there are no units that trigger this.

clemty:
I can do that - I just thought it'd be an easy fix since it is already exposed to Lua, and apparently no other mission uses that command either?

Mailaender:
Just add it to https://github.com/OpenRA/OpenRA/issues/9118. I also didn't know we already have `TriggerPowerOutage` but there is no need to waste performance on unused traits as there are no spies or EMP weapons in the TD mod.

PR #9300, opened 2015-09-07
Removed a redundant bot production debug message

Mailaender:
This is spammy and not really useful, because a) you can already see this via the observer tools and b) it is announced in the line before which tracks the prioritizing and decision making `AI: {0} decided to build {1}: Priority override` or `{0} decided to build {1}: Desired is {2} ({3} / {4}); current is {5} / {4}` so we can just skip this and keep an overview in the already crowded bot debug lines.


RoosterDragon:
:+1: Sounds reasonable.

PR #9309, opened 2015-09-09
Add margin to CreateEffectWarhead air check

### LINKED ISSUE #9308: Missiles sometimes trigger Air impact type when targeting ground
This is what #9016 attempted to fix.
On bleed, the issue is only really noticable in TS, and it doesn't happen all the time. The best shot at reproducing it is shooting at various directions on the ground with bazooka infantry.
It can be reproduced more reliably with my `BallisticMissile` projectile in TDX (even the old betas from May), where it actually happens more often. That also means it is not a recent regression, but probably never (visibly) triggers in the shipping mods because in the 3 shipping mods missiles usually start at rather low altitude already and reach Z = 0 before "CloseEnough" triggers due to the Tan flight curve.
But the higher the projectile speed and starting altitude (or the lower the MaximumPitch), the more likely the issue comes up.

The issue becomes much more frequent and reproducable with #8717 as well, due to the different flight curve.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26569525-missiles-sometimes-trigger-air-impact-type-when-targeting-ground?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Instead of considering the ImpactType to be "Air" as soon as the warhead triggers even a single Z WDist above 0 (relative to terrain height), there is now a small margin. This fixes the issue that #9016 originally attempted to fix, namely that missiles sometimes show the "air" explosion even though they were shot at a ground target.

Fixes #9308.

### DIFF
--- a/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
@@ -39,8 +39,13 @@ namespace OpenRA.Mods.Common.Warheads
 
 		public static ImpactType GetImpactType(World world, CPos cell, WPos pos)
 		{
+			// Missiles need a margin because they sometimes explode a little above ground
+			// due to their explosion check triggering slightly too early (because of CloseEnough).
+			// TODO: Base ImpactType on target altitude instead of explosion altitude.
+			var airMargin = new WDist(128);
+
 			var dat = world.Map.DistanceAboveTerrain(pos);
-			var isAir = dat.Length > 0;
+			var isAir = dat.Length > airMargin.Length;
 			var isWater = dat.Length <= 0 && world.Map.GetTerrainInfo(cell).IsWater;
 			var isDirectHit = GetDirectHit(world, cell, pos);
 


pchote:
This feels a little hacky... What do you think of the idea of basing the explosion type on the altitude of the target, rather than the projectile?

pchote:
Note that that would require a bigger rework than this, so may be a longer term prospect.  Another idea could be to force the explosion to the same altitude as the target, but that may have other unwanted side effects.

reaperrr:
> Another idea could be to force the explosion to the same altitude as the target, but that may have other unwanted side effects.

I'm uncomfortable with that approach, and it would presumably make implementing Airburst in #8717 harder.

> Note that that would require a bigger rework than this, so may be a longer term prospect.

Could we compromise on accepting this hack as temporary fix for Next+1 (maybe minus the yaml exposing so we won't need an upgrade rule later, but instead a TODO explaining the issue and pointing at the proper fix)? The issue is annoying enough in TS and TDX (and #8717) that I'd like to see a quick fix, but I'm not sure I'll have the time for that bigger rework in the next weeks.

For what it's worth, I did test this and it works well in practice. A mere 2 world units margin doesn't make a difference in any other situation than this. Bullets explode at the right altitude anyway, and air targets fly much higher than 2 or 3 world units above ground so it can't regress AA missiles either. This hack shouldn't cause any regressions and would be easy to replace/revert later.

pchote:
> Could we compromise on accepting this hack as temporary fix for Next+1 (maybe minus the yaml exposing so we won't need an upgrade rule later, but instead a TODO explaining the issue and pointing at the proper fix)? The issue is annoying enough in TS and TDX (and #8717) that I'd like to see a quick fix, but I'm not sure I'll have the time for that bigger rework in the next weeks.

Sorry, I meant for that to be implicit in my comment above.  I'm fine with taking this as a temporary fix.

OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
```diff
@@ -37,10 +37,15 @@ public class CreateEffectWarhead : Warhead
...
+		// Needs a default slightly above 0 because missiles sometimes seem to explode 2 or 3 world units above ground
+		// due to their explosion check triggering slightly too early (probably because of CloseEnough).
+		[Desc("Highest altitude above ground before the explosion is considered an air explosion.")]
+		public readonly WDist AirMargin = new WDist(2);
+
+		public ImpactType GetImpactType(World world, CPos cell, WPos pos)
 		{
 			var dat = world.Map.DistanceAboveTerrain(pos);
-			var isAir = dat.Length > 0;
+			var isAir = (dat.Length - AirMargin.Length) > 0;
```
pchote: `dat.Length > AirMargin.Length` is a bit clearer.

reaperrr:
Updated.

reaperrr:
Updated, with #8717 in mind I increased the value to 128 world units. This is still way below any aircraft's cruise altitude, but is needed to improve the situation once #8717 is merged.

pchote:
Looks good to me :+1:

penev92:
Can we unhardcode this and put it in the weapon/projectile/warhead/whatever info?

reaperrr:
> Can we unhardcode this and put it in the weapon/projectile/warhead/whatever info?

Originally I actually did that, but this is supposed to be a temporary fix. As @pchote mentioned above, the proper fix would be to base the explosion type on the altitude of the target, rather than the projectile, but I don't feel like adressing that in the next weeks (at least not in time for the feature-freeze).

But if I unhardcode it now, we'll need another upgrade rule when that proper fix happens, something I'd like to avoid. Exposing this to yaml in the upcoming release and then removing it again in Next+1 in favor of something else would just cause unnecessary headaches.

penev92:
Umm... the projectile can be designed to explode at altitude Z above the target(s), so I'm not sure I get the "base the explosion type on the altitude of the target, rather than the projectile" idea.

GraionDilach:
Such designed projectiles would have a tag specifying that Z altitude - `Airburst` - while this is about projectiles without such behaviour - in their case the long-term fix would be a snap-to-Z when reaching `CloseEnough` at Missile and something similar on `Bullet`.

abcdefg30:
:+1:

PR #9311, opened 2015-09-09
DeathType Count check in Explodes

### LINKED ISSUE #9310: Demo trucks don't explode when destroyed
They only explode when deployed. This is in current bleed.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26570907-demo-trucks-don-t-explode-when-destroyed?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

atlimit8:
Fixes #9310.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Explodes.cs
+++ b/OpenRA.Mods.Common/Traits/Explodes.cs
@@ -52,7 +52,7 @@ namespace OpenRA.Mods.Common.Traits
 				return;
 
 			var warhead = e.Warhead as DamageWarhead;
-			if (warhead != null && !warhead.DamageTypes.Overlaps(info.DeathType))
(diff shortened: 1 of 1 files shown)


RoosterDragon:
I hate it when null checks look like null checks but turn out to be key control flows instead. Bloody nulls.

:+1: Confirmed fixes the regression - and checking the count makes it less of a minefield than the original check so that's a nice bonus.

reaperrr:
:+1:

PR #9313, opened 2015-09-10
Eagerly cache traits in Actor

RoosterDragon:
In this case, removing the lazy generation buys us more performance in removing the overhead of accessing these values than we lose from avoided computation when these values are never accessed.

Caching these traits in the constructor is essentially free and this saves us 1% in CPU by not having to access everything through the indirection of Lazy<T>.Value.

[Diff ignoring whitespace](https://github.com/OpenRA/OpenRA/pull/9313/files?w=1)

### DIFF
--- a/OpenRA.Game/Actor.cs
+++ b/OpenRA.Game/Actor.cs
@@ -40,34 +40,38 @@ namespace OpenRA
 		public Group Group;
 		public int Generation;
 
-		Lazy<Rectangle> bounds;
-		Lazy<Rectangle> visualBounds;
-		Lazy<IFacing> facing;
-		Lazy<Health> health;
-		Lazy<IOccupySpace> occupySpace;
-		Lazy<IEffectiveOwner> effectiveOwner;
-
-		public Rectangle Bounds { get { return bounds.Value; } }
-		public Rectangle VisualBounds { get { return visualBounds.Value; } }
-		public IOccupySpace OccupiesSpace { get { return occupySpace.Value; } }
-		public IEffectiveOwner EffectiveOwner { get { return effectiveOwner.Value; } }
+		public Rectangle Bounds { get; private set; }
+		public Rectangle VisualBounds { get; private set; }
+		public IEffectiveOwner EffectiveOwner { get; private set; }
+		public IOccupySpace OccupiesSpace
+		{
+			get
+			{
+				if (occupySpace == null)
+					occupySpace = Trait<IOccupySpace>();
+				return occupySpace;
+			}
+		}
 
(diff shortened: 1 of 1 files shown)


OpenRA.Game/Actor.cs
```diff
@@ -40,34 +40,38 @@ public sealed class Actor : IScriptBindable, IScriptNotifyBind, ILuaTableBinding
...
+		public IEffectiveOwner EffectiveOwner { get; private set; }
+		public IOccupySpace OccupiesSpace
+		{
+			get
+			{
+				if (occupySpace == null)
+					occupySpace = Trait<IOccupySpace>();
+				return occupySpace;
+			}
+		}
```
RoosterDragon: Minor behaviour change here. Accessing this now throws if IOccupySpace is missing, rather than returning null. None of our existing code is affected by this change.

pchote: If we cared about the behaviour change (I don't think we do) you could use a sentinel boolean and TraitAndDefault.

OpenRA.Game/Actor.cs
```diff
@@ -40,34 +40,38 @@ public sealed class Actor : IScriptBindable, IScriptNotifyBind, ILuaTableBinding
...
 			get
 			{
 				// TODO: Support non-zero pitch/roll in IFacing (IOrientation?)
-				var facingValue = facing.Value != null ? facing.Value.Facing : 0;
+				var facingValue = facing != null ? facing.Facing : 0;
 				return new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(facingValue));
 			}
 		}
 
+		IOccupySpace occupySpace;
```
pchote: Why not merge this with the getter above, and then initialize it via a TraitOrDefault at the end of the ctor?

RoosterDragon: Several traits access properties that depend on ```IOccupySpace``` during construction. So it needs to be available at that time.

pchote: Ah, I see.

reaperrr:
Needs rebase.

RoosterDragon:
Rebased.

pchote:
Looks good to me :+1:

Mailaender:
Can't spot negatives side effects in-game. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/61a807fbfed338263bd1af8b88dbe694889cf34f...13cfd02eb35eeb95f150fb76baa85d2360590165)

PR #9317, opened 2015-09-11
Generic WithRangeCircle

atlimit8:
Replaces `RenderRangeCircle`, `RenderDetectionCircle`, `RenderShroudCircle`, and `RenderJammerCircle` as well as providing range circles for `BaseProvider`, weapons (through armaments), and armaments (regardless of whether attacking is enabled).

Also gives control over `Color` and `ContrastColor` of the range circles providing color scheme control by modders.

Uses provider type and option variant filter for selecting ranges provided by other traits. The `Name` field controls the circle group - range circles (with the same name) of other actors are shown when placing.

Provided types:

```
weapon: range of weapon from armament with name matching variant
armament: range of armament(s)
    [with name matching variant]
targets: range of armaments
    [with a certain valid target type]
attack-armament: like armament except show only if able to attack
attacks: like targets except show only if able to attack
detect: range of cloak detection
    [with variant as one of the cloak types detected]
shroud: range of shroud generated
jams: range of all jamming
    missiles: range of missile jamming
    radar: range of radar jamming
```

The fields `Name`, `Type`, and `Variant` might be more understandable with different names:

```
Name => Circles | CircleGroup | Group
Type => Filter | Provider
Variant => Subfilter | Filter
```


pchote:
I think it would be worth checking out how the selection bars are implemented - you should be able to get almost as much flexibility but with much less code by adapting this to use a similar approach.

atlimit8:
Alright.

abcdefg30:
Needs a rebase.

atlimit8:
Low priority right now.

PR #8659, opened 2015-07-05
Rename PlayerReference.Race to Faction

penev92:
This is the second step in a series of PRs toward unblocking #7873.
Renames `PlayerReference.Race` and `PlayerReference.LockRace` and updates ALL THE MAPS!
Only the TS shellmap is spared.
Should be pretty easy to review, actually.

PR #9324, opened 2015-09-12
Hex Color Parsing

### LINKED ISSUE #8984: OpenRA.Utility.exe Color Picker
Parse in the form *base_color* `,` *saturation* `,` *lightness* `,` *alpha* and output rgb[a]/hsl[a].
Where:
  * *base_color* is one of:
    * `rgb` `,` *red* `,` *green* `,` *blue*
    * [`hsl` `,`] *hue* : HSL (hue, saturation, lightness) values (*hue*, 255, 255)
    * *knownColor* : parsed by `Enum<KownColor>.TryParse` (See System.Drawing.KnownColor)
  * *saturation* is multiplied by *base_color* saturation to get the final saturation (colourfulness - 0 is grey)
  * *lightness* is multiplied by *base_color* lightness to get the final lightness (brightness - 0 is black)
  * *alpha* is the opacity/alpha channel

Options for shortening:
  * Default missing to 255 filling from alpha. Ex: `Blue, 127` is blue with 50% opacity/alpha
    * This would allow for using rgba semantics like `rgb, 0, 0, 255, 127` to represent blue with 50% opacity/alpha
  * Default empty to 255. Ex: `Blue, , , 127` is blue with 50% opacity/alpha
    * The number of components is more deterministic

Since there have been remarks about the HSL used by OpenRA being counterintuitive, this might help:
  * Component text: *numerator* `/` *denominator*
    * Component value = `255 *` *numerator* `/` *denominator*
    * Except *hue* is `256 *` *numerator* `/` *denominator*
    * Allows for `hsl, 180/360, 50/100, 50/100`
  * Component text: *percentage* `%`

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25506314-openra-utility-exe-color-picker?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #8986: Unified MiniYaml Color Syntax
Currently, MiniYaml parses `Color` using [a]rgb format and `HSLColor` using hsl[a] format.
They should be one of these:
  * *red* `,` *green* `,` *blue* [`,` *alpha*]
  * *hue* `,` *saturation* `,` *lightness* [`,` *alpha*]
  * Either by prefix:
    * `hsl,` *hue* `,` *saturation* `,` *lightness* [`,` *alpha*]
    * `rgb,` *red* `,` *green* `,` *blue* [`,` *alpha*]

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25511460-unified-miniyaml-color-syntax?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

atlimit8:
Closes #8986 - Unified MiniYaml Color Syntax
Closes #8984 - OpenRA.Utility.exe Color Picker - no need because any color picker that gives hex values will be usable.


penev92:
This will either need an upgrade rule, or we'll want to keep the default (no prefix) to HSL.

atlimit8:
It still parses the prefix-less format for `HSLColor` (and `Color`). The `hsl,` prefixes were added so that modders would see that the colors are in the HSL color space.

atlimit8:
Currently, this allows for a gradual transition by saving in the new format and writing in the new format once documentation is added. Another approach would be to change everything and use an upgrade rule as mentioned.

penev92:
Fortunately #8659 added support for that not too long ago, so an upgrade rule might be the better option. This does break map compatibility unless they get upgraded, but I don't think that is that much of an issue.

atlimit8:
OK, I'll add upgrade rules unless someone objects.

That said, should prefix-less values still be supported then?

pchote:
Can you pull out utility methods for parsing a `Color` or `HSLColor` from a string?  This will reduce a bit of duplication and make `FieldLoader` much cleaner.

atlimit8:
* Common color loading and upgrade rule methods
* name color space added because it requires little additional code and they are easier to read which will be helpful as more colors are un-hardcoded.

Still not sure about including default color spaces for when the color space prefix is absent.

abcdefg30:
Needs a rebase.

penev92:
IMO we whouldn't allow for the "identifier is missing" case. That would just make it harder to read. And to upgrade in the future.

atlimit8:
Rebased (a while ago).

atlimit8:
Rebased again.

abcdefg30:
Needs a rebase again...

atlimit8:
Rebased.

abcdefg30:
Needs a rebase again.

atlimit8:
Rebased yet  again. Hopefully, nothing conflicts for a while. I was hoping `git` would not have trouble with this and #9444.

reaperrr:
Spotted no issues and in-game all mods looked normal.
:+1:

penev92:
Sorry, needs rebasing again.

reaperrr:
Seems like one of the new maps needs to be updated.

atlimit8:
I noticed and already fixed, hopefully.

atlimit8:
I forgot to run `make test` *after* the default color spaces were removed.

pchote:
I haven't been that much of a fan of the mandatory prefixes, but haven't had time to think about alternatives until now.  I think a better approach would be to use html style colour codes (e.g. CC0000 for dark red) for rgb, and then keep the current 3-part syntax for hsl.  That leaves no ambiguity in parsing the code.

atlimit8:
Changed to hex color parsing.

reaperrr:
I won't block a merge over this if the majority favors this approach, but personally I'd say from a modders perspective, the prefix was still *far* more user-friendly than html could ever be. Every modder understands what 255,120,44 means in rgb, but with html, many (most?) would have to use a dec-hex converter, bye-bye to getting the desired color quickly and without much thinking.

atlimit8:
Since I have written both ways already, I could do either. The prefix system could gain the advantages of the HTML color system by adding a `hex` prefix and replacing `,` with spaces (helpful with parsing Color[]).

Another option would be to add the space splitting and parse `(hhhhhh(hh | ' ' a)? | r g b [a])`:
```
hhhhhh    - 1 part, length of 6
hhhhhhhh  - 1 part, length of 8
hhhhhh a  - 2 parts
r g b     - 3 parts
r g b a   - 4 parts
```
* The format is unique to the number of parts with the exception of `hhhhhh[hh]` which is unique by length.
* This flexible format also fits in comma delimited lists for `Color[]` loading.
* Hex color pickers could be used allowing for closure of #8984.
* Deprecates normalized HSL color space, which is considered confusing by many.
* If modders have an easier time with `r, g, b`, then `r g b` will also be easier.

Opinions?

pchote:
r,g,b[,a] triples (quads) might be a bit easier for modders (personally, I find the other way - I can quite happily visualize hex color codes, but need to concentrate or look up anything except crude r,g,b triples) but makes it significantly more complicated on our end because of the ambiguity of alpha/non-alpha in arrays, and with h,s,l colors.  Going with hex codes is a simple solution to something that shouldn't need to be complicated.

atlimit8:
@reaperrr:
What about adding a rgb[a] to to conversion to `OpenRA.Utility.exe`?

atlimit8:
Adding something like:
```
$ [mono] OpenRA.Utility.exe all --rgba2hex 255,255,255 96,96,96,128
FFFFFF, 60606080
```

reaperrr:
That would help I guess.

Generally speaking, I know I'd be able to adapt to it, but I wanted to point out that some modders might complain.

Anyway, needs rebase.

abcdefg30:

abcdefg30:
Reopening as discussed on IRC.

atlimit8:
Finally rebased.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2795,10 +2892,43 @@ internal static void UpgradePlayers(int engineVersion, ref List<MiniYamlNode> no
...
+
+				UpgradeChromeMetrics(engineVersion, ref node.Value.Nodes, node, depth + 1);
+			}
+		}
+
+		internal static void UpgradeChromeLayout(int engineVersion, ref List<MiniYamlNode> nodes, MiniYamlNode parent, int depth)
+		{
+			foreach (var node in nodes)
+			{
+				if (engineVersion < 20151018)
```
abcdefg30: Please add braces here.

OpenRA.Mods.Common/UtilityCommands/Rgba2Hex.cs
```diff
@@ -0,0 +1,202 @@
...
+		}
+
+		bool PrintUsage()
+		{
+			Console.WriteLine("");
+			Console.WriteLine("Usage:");
+			Console.WriteLine("\tOpenRA.Utility.exe [MOD] --rgba2hex r1,g1,b1");
+			Console.WriteLine("\tOpenRA.Utility.exe [MOD] --rgba2hex r1,g1,b1,a1");
+			Console.WriteLine("\tOpenRA.Utility.exe [MOD] --rgba2hex r1,g1,b1 r2,g2,b2,a2");
+			Console.WriteLine("\tOpenRA.Utility.exe [MOD] --rgba2hex r1,g1,b1 r2,g2,b2,a2");
```
abcdefg30: This just duplicates the line above?

OpenRA.Mods.Common/UtilityCommands/Rgba2Hex.cs
```diff
@@ -0,0 +1,202 @@
...
+			for (int i = 1; i < args.Length; i++)
+			{
+				var parts = args[i].Split(Comma, StringSplitOptions.RemoveEmptyEntries);
+				if (parts.Length != 3 && parts.Length != 4)
+				{
+					invalid = true;
+					Console.WriteLine("Invalid color (argument " + i + "): " + args[i]);
+				}
+				else
+					foreach (var part in parts)
```
abcdefg30: Please add braces here and for the `else`, too.

OpenRA.Mods.Common/UtilityCommands/Rgba2Hex.cs
```diff
@@ -0,0 +1,202 @@
...
```
abcdefg30: Same here.

abcdefg30:
> 15:41:10 (pchote) abcdefg30: map format bump might be a good idea

abcdefg30:
Crashes in the modchooser.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
atlimit8: Added.

atlimit8:
Added `MapFormat` upgrade change.

abcdefg30:
Looks good to me. :+1: 
But needs a rebase again.

Do we want to merge this before the next release?
(It's quite a big change.)

Also, what about https://github.com/OpenRA/OpenRA/pull/9324#issuecomment-149216473?

atlimit8:
This change has been discussed at length for some time to reach this point and the vast majority of the changes in the patch come from applying the upgrade rules.

I did add [that feature](https://github.com/OpenRA/OpenRA/pull/9324#issuecomment-149216473) along with auto updating map format.
[See OpenRA.Game/Map/Map.cs](https://github.com/OpenRA/OpenRA/pull/9324/files#diff-ad4437e58b6bba75e4be119fc69224c3R347)

penev92:
Definitely not before the release.

abcdefg30:
> I did add [that feature](https://github.com/OpenRA/OpenRA/pull/9324#issuecomment-149216473) along with auto updating map format.
[See OpenRA.Game/Map/Map.cs](https://github.com/OpenRA/OpenRA/pull/9324/files#diff-ad4437e58b6bba75e4be119fc69224c3R347)

Ops, sry, didn't see that.

mods/cnc/maps/gdi06/map.yaml
```diff
@@ -1165,6 +1161,10 @@ Weapons:
 
 Voices:
 
+Music:
+	rain-ambient: Rain (ambient)
+		Hidden: true
```
obrakmann: Rebase error?

obrakmann:
Looks good. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c54ae8f104609f5b643907b75077ade000b20fa1)

Issue #9322, opened 2015-09-11
Bridges appear undamaged at the screen's edge


abcdefg30:
... although the bridge is actually destroyed.

This happenes when the bridge is at the edge of my screen:
![bridgebuga](https://cloud.githubusercontent.com/assets/7704140/9827501/19013772-58e3-11e5-93ae-5eebb5b2e51a.png)
And this when the bridge is not at the edge (like it should be at the edge too):
![bridgebugb](https://cloud.githubusercontent.com/assets/7704140/9827502/1a84938c-58e3-11e5-8249-f9f3095e8048.png)

pchote:
I bet that adding a `CustomSelectionSize` trait will fix this.

Mailaender:
I filed https://github.com/OpenRA/OpenRA/issues/9327. Not sure which change is responsible for these regressions.

reaperrr:
Buildings completely disappear from screen once a row of cells is off-screen as well, I suspect this is a similar issue (and I doubt selection size has to do with it, unless the game wants the *whole* selection size to be on-screen).

reaperrr:
As mentioned on IRC, the previous releases (I tested June and April releases from this year) had the same issue, so it's not a recent regression and doesn't seem to be caused by the Render/SelectionSize refactorings.

reaperrr:
Fixed by #9327.

PR #9327, opened 2015-09-12
Fixed bridge rendering glitches

Mailaender:
As suggested in https://github.com/OpenRA/OpenRA/issues/9322.

### DIFF
--- a/OpenRA.Mods.Common/Traits/CustomSelectionSize.cs
+++ b/OpenRA.Mods.Common/Traits/CustomSelectionSize.cs
@@ -13,9 +13,7 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
-	[Desc("Special case trait for unselectable actors that need to define targetable area bounds",
-	"for special cases like C4, engineer repair and tooltips.",
-	"Examples: bridge huts and crates.")]
+	[Desc("Special case trait for actors that need to define targetable area and screen map bounds manually.")]
 	public class CustomSelectionSizeInfo : ITraitInfo
 	{
 		[FieldLoader.Require]
--- a/OpenRA.Mods.Common/Traits/Render/AutoSelectionSize.cs
+++ b/OpenRA.Mods.Common/Traits/Render/AutoSelectionSize.cs
@@ -13,6 +13,7 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
+	[Desc("Automatically calculates the targetable area and screen map boundaries from the sprite size.")]
 	public class AutoSelectionSizeInfo : ITraitInfo, Requires<RenderSpritesInfo>
 	{
 		public object Create(ActorInitializer init) { return new AutoSelectionSize(this); }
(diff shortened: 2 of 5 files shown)


abcdefg30:
I can still reproduce #9322.

Mailaender:
Are you sure as I can't.

abcdefg30:
bridgebug

Mailaender:
Okay, do we still want this?

abcdefg30:
Well, the changes still look reasonable to me. :+1:

OpenRA.Mods.Common/Traits/CustomSelectionSize.cs
```diff
@@ -13,9 +13,7 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
-	[Desc("Special case trait for unselectable actors that need to define targetable area bounds",
-	"for special cases like C4, engineer repair and tooltips.",
-	"Examples: bridge huts and crates.")]
+	[Desc("Special case trait for actors that need to define targetable area and rendering bounds manually.")]
```
reaperrr: Actually, to my knowledge this does *only* affect targetable area, not rendering bounds.

OpenRA.Mods.Common/Traits/Render/AutoSelectionSize.cs
```diff
@@ -13,6 +13,7 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
+	[Desc("Automatically calculates the targetable area and rendering boundaries from the sprite size.")]
```
reaperrr: Same here.

OpenRA.Mods.Common/Traits/CustomSelectionSize.cs
Mailaender: That is what I thought as well until I was lectured by @pchote in https://github.com/OpenRA/OpenRA/pull/9203 so I wanted to share that knowledge.

pchote: It affects everything that queries the ScreenMap, which includes selecting the actors that should be rendered.

reaperrr: You learn something new every day... ok then.

pchote: Note that that *is* different to the rendering bounds.

Mailaender: How would you call it instead here then?

Mailaender:
Let's not merge this yet. I am not yet sure about all effects this has or if it is needed at all.

pchote:
This "doesn't work" because you haven't changed all the bridges.

Mailaender:
Should be fixed now.

reaperrr:
Can confirm this fixes the issue now :+1:

RoosterDragon:
Much better :+1:

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/5ed9f6b9f50d5e11d157eafd2646374ece2bde73...4f882ebb9d8369666a4bf68d2767aeb38c8d186c)

PR #8927, opened 2015-08-03
Trait [Info] LINQ-like Access Changes

atlimit8:
_TODO: rewrite_

PR #9332, opened 2015-09-13
TraitInfo through ActorInfo and HasTrait Removal

atlimit8:
Split and rebased from #8927; the reason for taking the earlier commits is the discussion about LINQ-like methods in the comments for #8927.

_selfOrOtherActor_`.HasTrait<`_Trait_`>()` => _selfOrOtherActor_`.Info.HasTraitInfo<`_TraitInfo_`>`
 // with ITraitInfo restriction

_actorInfo_`.Traits.Contains` => _actorInfo_`.HasTraitInfo` // with ITraitInfo restriction

_actorInfo_`.Traits.Get` => _actorInfo_`.TraitInfo` // with ITraitInfo restriction

_actorInfo_`.Traits.GetOrDefault` => _actorInfo_`.TraitInfoOrDefault` // with ITraitInfo restriction

Foundation for a future PR for reducing trait access and allocation where trait state is not needed.

Separated because a few of these changes were put in #8848 to reduce ordering (`InitializeAfter<>`) and requirement (`Requires<>`) requirements and @pchote requested the changes be made everywhere before the split (`Requires<>` => `Requires<>` and `InitializeAfter<>`) in #8848.

Almost all of this is just `for f in $(git grep -lF ...); do sed -i s/.../.../g $f; done` substitutions with 1 method removed from `Actor` and 3 wrapper methods with `where T : ITraitInfo` added to `ActorInfo`.

@RoosterDragon profile this one.

### DIFF
--- a/OpenRA.Mods.Common/AI/AttackOrFleeFuzzy.cs
+++ b/OpenRA.Mods.Common/AI/AttackOrFleeFuzzy.cs
@@ -176,7 +176,7 @@ namespace OpenRA.Mods.Common.AI
 			var sumOfHp = 0;
 			foreach (var a in actors)
 			{
-				if (a.Info.Traits.Contains<HealthInfo>())
+				if (a.Info.HasTraitInfo<HealthInfo>())
 				{
 					sumOfMaxHp += a.Trait<Health>().MaxHP;
 					sumOfHp += a.Trait<Health>().HP;
@@ -228,7 +228,7 @@ namespace OpenRA.Mods.Common.AI
 		{
 			var sum = 0;
 			foreach (var a in actors)
-				if (a.Info.Traits.Contains<TTraitInfo>())
+				if (a.Info.HasTraitInfo<TTraitInfo>())
 					sum += getValue(a);
 
 			return sum;
@@ -240,7 +240,7 @@ namespace OpenRA.Mods.Common.AI
 			var countActors = 0;
 			foreach (var a in actors)
 			{
-				if (a.Info.Traits.Contains<TTraitInfo>())
+				if (a.Info.HasTraitInfo<TTraitInfo>())
 				{
 					sum += getValue(a);
 					countActors++;
--- a/OpenRA.Mods.Common/AI/AttackOrFleeFuzzy.cs
+++ b/OpenRA.Mods.Common/AI/AttackOrFleeFuzzy.cs
@@ -208,7 +208,7 @@ namespace OpenRA.Mods.Common.AI
 
 		protected float RelativeSpeed(IEnumerable<Actor> own, IEnumerable<Actor> enemy)
 		{
-			return RelativeValue(own, enemy, 100, Average<MobileInfo>, (Actor a) => a.Info.Traits.Get<MobileInfo>().Speed);
+			return RelativeValue(own, enemy, 100, Average<MobileInfo>, (Actor a) => a.Info.TraitInfo<MobileInfo>().Speed);
 		}
 
 		protected static float RelativeValue(IEnumerable<Actor> own, IEnumerable<Actor> enemy, float normalizeByValue,
--- a/OpenRA.Mods.Common/AI/AttackOrFleeFuzzy.cs
+++ b/OpenRA.Mods.Common/AI/AttackOrFleeFuzzy.cs
@@ -176,7 +176,7 @@ namespace OpenRA.Mods.Common.AI
 			var sumOfHp = 0;
 			foreach (var a in actors)
 			{
-				if (a.HasTrait<Health>())
+				if (a.Info.Traits.Contains<HealthInfo>())
 				{
 					sumOfMaxHp += a.Trait<Health>().MaxHP;
 					sumOfHp += a.Trait<Health>().HP;
@@ -191,7 +191,7 @@ namespace OpenRA.Mods.Common.AI
 
 		protected float RelativePower(IEnumerable<Actor> own, IEnumerable<Actor> enemy)
 		{
-			return RelativeValue(own, enemy, 100, SumOfValues<AttackBase>, a =>
+			return RelativeValue(own, enemy, 100, SumOfValues<AttackBaseInfo>, a =>
 			{
 				var sumOfDamage = 0;
 				var arms = a.TraitsImplementing<Armament>();
@@ -208,7 +208,7 @@ namespace OpenRA.Mods.Common.AI
 
 		protected float RelativeSpeed(IEnumerable<Actor> own, IEnumerable<Actor> enemy)
 		{
-			return RelativeValue(own, enemy, 100, Average<Mobile>, (Actor a) => a.Trait<Mobile>().Info.Speed);
+			return RelativeValue(own, enemy, 100, Average<MobileInfo>, (Actor a) => a.Trait<Mobile>().Info.Speed);
 		}
 
 		protected static float RelativeValue(IEnumerable<Actor> own, IEnumerable<Actor> enemy, float normalizeByValue,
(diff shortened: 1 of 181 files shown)


RoosterDragon:
Can confirm this provides a small speedup for changes where we can look up the trait info rather than the trait. Nothing major but all the little savings tend to add up for trait access so it's always welcome.

Bikesheding which may be ignored if desired: ```TraitInfosAny``` still seems like an obtuse name to me. I think ```HasTraitInfo``` would be better.

atlimit8:
@pchote:
@RoosterDragon thinks `HasTraitInfo` is better than `TraitInfosAny`.
Should it just be an exception to using Linq-like names?

pchote:
`HasTraitInfo` is fine by me.

atlimit8:
OK, I didn't really like it myself either.
It works a little differently in that more specific to a particular kind of trait info rather than actor info in general.

atlimit8:
Renamed `TraitInfosAny` => `HasTraitInfo`.

OpenRA.Mods.Common/AI/AttackOrFleeFuzzy.cs
```diff
@@ -224,23 +224,23 @@ protected float RelativeSpeed(IEnumerable<Actor> own, IEnumerable<Actor> enemy)
 			return relative.Clamp(0.0f, 999.0f);
 		}
 
-		protected float SumOfValues<Trait>(IEnumerable<Actor> actors, Func<Actor, int> getValue)
+		protected float SumOfValues<TraitInfo>(IEnumerable<Actor> actors, Func<Actor, int> getValue) where TraitInfo : ITraitInfo
```
RoosterDragon: Since you're here: This should be ```TTraitInfo```.

OpenRA.Mods.Common/AI/AttackOrFleeFuzzy.cs
```diff
@@ -224,23 +224,23 @@ protected float RelativeSpeed(IEnumerable<Actor> own, IEnumerable<Actor> enemy)
...
 			foreach (var a in actors)
-				if (a.HasTrait<Trait>())
+				if (a.Info.HasTraitInfo<TraitInfo>())
 					sum += getValue(a);
 
 			return sum;
 		}
 
-		protected float Average<Trait>(IEnumerable<Actor> actors, Func<Actor, int> getValue)
+		protected float Average<TraitInfo>(IEnumerable<Actor> actors, Func<Actor, int> getValue) where TraitInfo : ITraitInfo
```
RoosterDragon: ```TTraitInfo``` here too.

OpenRA.Mods.Common/Activities/Air/HeliReturn.cs
```diff
@@ -17,20 +17,16 @@ namespace OpenRA.Mods.Common.Activities
...
 		{
-			aircraftInfo = self.Info.Traits.Get<AircraftInfo>();
 			heli = self.Trait<Helicopter>();
-			heliInfo = self.Info.Traits.Get<HelicopterInfo>();
 		}
 
 		public static Actor ChooseHelipad(Actor self)
 		{
-			var rearmBuildings = self.Info.Traits.Get<HelicopterInfo>().RearmBuildings;
+			var rearmBuildings = self.Info.TraitInfo<HelicopterInfo>().RearmBuildings;
```
RoosterDragon: Should this also be using ```heli.Info```?

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
```diff
@@ -26,15 +26,15 @@ public class CombatDebugOverlay : IPostRender, INotifyDamage
 	{
 		readonly DeveloperMode devMode;
 
+		readonly HealthInfo health;
 		Lazy<AttackBase> attack;
 		Lazy<IBodyOrientation> coords;
-		Lazy<Health> health;
```
RoosterDragon: Is removing the laziness here safe without some sort of ordering guarantee of the trait construction order?

OpenRA.Mods.D2k/Traits/Carryall.cs
```diff
@@ -50,7 +50,8 @@ public Carryall(Actor self, CarryallInfo info)
 
 			IsBusy = false;
 			IsCarrying = false;
-			carryHeight = self.Trait<Helicopter>().Info.LandAltitude;
+			var helicopter = self.Info.TraitInfoOrDefault<HelicopterInfo>();
+			carryHeight = helicopter != null ? helicopter.LandAltitude : WDist.Zero;
```
RoosterDragon: For the attention of people who know better: Should Carryall require the helicopter trait instead of making this change, or is this change OK?

pchote: See #9054.

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
atlimit8: It is not removing the laziness from a (lazy loaded) trait; rather, it is replacing a trait with its trait info which is initialized before actor creation starts.

atlimit8: The trait state is not used at all.

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
```diff
@@ -26,15 +26,15 @@ public class CombatDebugOverlay : IPostRender, INotifyDamage
 	{
 		readonly DeveloperMode devMode;
 
+		readonly HealthInfo health;
```
pchote: can we call this `healthInfo` to make it clear it is an Info?

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
RoosterDragon: You're right of course :)

atlimit8:
Changed.

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
```diff
@@ -96,11 +96,8 @@ public void Damaged(Actor self, AttackInfo e)
 			if (devMode == null || !devMode.ShowCombatGeometry || e.Damage == 0)
 				return;
 
-			var health = self.TraitOrDefault<Health>();
-			if (health == null)
-				return;
-
-			var damageText = "{0} ({1}%)".F(-e.Damage, e.Damage * 100 / health.MaxHP);
+			var maxHP = (healthInfo != null && healthInfo.HP > 0) ? healthInfo.HP : 1;
+			var damageText = "{0} ({1}%)".F(-e.Damage, e.Damage * 100 / maxHP);
```
RoosterDragon: This looks wrong. You'll now execute this code if ```healthInfo``` is null rather than doing nothing, and your ```maxHP``` variable is grabbing the ```HP``` from the health, not that ```MaxHP```.

atlimit8: I'm changing it to:
```c#
if (healthInfo == null)
	return;

var maxHP = healthInfo.HP > 0 ? healthInfo.HP : 1;
```

`Health.MaxHP` is set once to `info.HP > 0 ? info.HP : 1` so this will be equivalent minus the trait look up.

reaperrr:
Needs rebase.

atlimit8:
Rebased.

RoosterDragon:
:+1:

reaperrr:
Needs another rebase, :+1: after that.

atlimit8:
Rebased.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d47d7607d2ffbb95ecab7aba7df9fd9a68d32ab5...4d4a2414a917a7982e0624f3f74bc4663f096f4e)

PR #9333, opened 2015-09-13
Make DuplicateUnitCrateAction only check enabled targetability

atlimit8:
Make DuplicateUnitCrateAction only use enabled `ITargetable` traits for valid target check.
Missed this in #8727.


abcdefg30:
:+1:

penev92:
:+1:

PR #9334, opened 2015-09-13
Add support for EditorTemplateOrder auto-generation

Mailaender:
As the TS/RA2 tilesets are big, it is more convenient to generate this automatically as well.

### DIFF
--- a/OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
+++ b/OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
@@ -12,6 +12,8 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
+using System.Text;
 using OpenRA.Mods.Common.FileFormats;
 
 namespace OpenRA.Mods.TS.UtilityCommands
@@ -25,7 +27,7 @@ namespace OpenRA.Mods.TS.UtilityCommands
 			return args.Length >= 3;
 		}
 
-		[Desc("FILENAME", "TEMPLATEEXTENSION", "Convert a legacy tileset to the OpenRA format.")]
+		[Desc("FILENAME", "TEMPLATEEXTENSION", "[TILESETNAME]", "Convert a legacy tileset to the OpenRA format.")]
 		public void Run(ModData modData, string[] args)
 		{
 			// HACK: The engine code assumes that Game.modData is set.
@@ -57,7 +59,21 @@ namespace OpenRA.Mods.TS.UtilityCommands
 				"Cliff" // TS defines this as "Rock"
 			};
 
+			var metadata = new StringBuilder();
+			metadata.AppendLine("General:");
+
+			var name = args.Length > 3 ? args[3] : Path.GetFileNameWithoutExtension(args[2]);
+			metadata.AppendLine("\tName: {0}".F(name));
+			metadata.AppendLine("\tId: {0}".F(name.ToUpperInvariant()));
+			metadata.AppendLine("\tPalette: iso{0}.pal".F(extension));
+			metadata.AppendLine("\tHeightDebugColors:  00000080, 00004480, 00008880, 0000CC80, 0000FF80, 4400CC80," +
+				" 88008880, CC004480, FF110080, FF550080, FF990080, FFDD0080, DDFF0080, 99FF0080, 55FF0080, 11FF0080");
+
 			// Loop over template sets
+			var data = new StringBuilder();
+			data.AppendLine("Templates:");
+			var definedCategories = new HashSet<string>();
+			var usedCategories = new HashSet<string>();
 			try
 			{
 				for (var tilesetGroupIndex = 0;; tilesetGroupIndex++)
@@ -67,6 +83,8 @@ namespace OpenRA.Mods.TS.UtilityCommands
 					var sectionCount = int.Parse(section.GetValue("TilesInSet", "1"));
 					var sectionFilename = section.GetValue("FileName", "").ToLowerInvariant();
 					var sectionCategory = section.GetValue("SetName", "");
+					if (!string.IsNullOrEmpty(sectionCategory) && sectionFilename != "blank")
+						definedCategories.Add(sectionCategory);
 
 					// Loop over templates
 					for (var i = 1; i <= sectionCount; i++, templateIndex++)
@@ -77,9 +95,11 @@ namespace OpenRA.Mods.TS.UtilityCommands
 
 						using (var s = modData.DefaultFileSystem.Open(templateFilename))
 						{
-							Console.WriteLine("\tTemplate@{0}:", templateIndex);
-							Console.WriteLine("\t\tCategory: {0}", sectionCategory);
-							Console.WriteLine("\t\tId: {0}", templateIndex);
+							data.AppendLine("\tTemplate@{0}:".F(templateIndex));
+							data.AppendLine("\t\tCategory: {0}".F(sectionCategory));
+							usedCategories.Add(sectionCategory);
+
+							data.AppendLine("\t\tId: {0}".F(templateIndex));
 
 							var images = new List<string>();
 
@@ -91,7 +111,7 @@ namespace OpenRA.Mods.TS.UtilityCommands
 									images.Add(variant);
 							}
 
-							Console.WriteLine("\t\tImages: {0}", images.JoinWith(", "));
+							data.AppendLine("\t\tImages: {0}".F(images.JoinWith(", ")));
 
 							var templateWidth = s.ReadUInt32();
 							var templateHeight = s.ReadUInt32();
@@ -101,8 +121,8 @@ namespace OpenRA.Mods.TS.UtilityCommands
 							for (var j = 0; j < offsets.Length; j++)
 								offsets[j] = s.ReadUInt32();
 
-							Console.WriteLine("\t\tSize: {0}, {1}", templateWidth, templateHeight);
-							Console.WriteLine("\t\tTiles:");
+							data.AppendLine("\t\tSize: {0}, {1}".F(templateWidth, templateHeight));
+							data.AppendLine("\t\tTiles:");
 
 							for (var j = 0; j < offsets.Length; j++)
 							{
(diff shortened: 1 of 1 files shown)


abcdefg30:
:+1:

pchote:
This adds to the already existing confusion about how to use this utility.  I suggest either adding some explanatory text saying that you must manually copy the templates definitions into a new file, write a new header, and then copy the `EditorTemplateOrder` definitions from the bottom of the output to the header at the top (a gross workaround, but better than nothing), or to have this write out a minimal header *before* the templates.

Mailaender:
Hmm, not sure how important polishing this utility command output really is. It never worked out of the box, but only generated certain parts. To me it already served it's purpose. See https://github.com/Mailaender/OpenRA/commit/2eb83d65127be6a6c7bfd16e3a53178d155ccf6b for the RA2 editor categories.

GraionDilach:
I'd say this worths the :+1: even as it is now, because RA2 terrain is still small compared to some mod terrains (YR with Terrain Expansion, D-Day, Twisted Insurrection, Project Phantom, maybe Robot Storm too) and since terrain conversion isn't a daily task, the overhead would still be small if it's dealt on a per-user basis.

Maybe a YAML comment at the generated output's top is sufficient to explain the steps need to be finalized.

pchote:
Indeed, that is a simple change to make this mergeable.

abcdefg30:

Mailaender:
I fixed the ordering and added stubs for the header definitions. The output is now a proper tileset definition file that should run on it's own to ease the conversion during first time setups where modders are usually not yet too experienced with the engine.

reaperrr:
> OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs:L76: [SP2100] Code line must be no longer than 180 characters (now 216).

Mailaender:
Sorry. My local StyleCop didn't complain about it. Fixed.

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
```diff
@@ -137,6 +157,41 @@ public void Run(ModData modData, string[] args)
...
+			foreach (var category in categories.Skip(1))
+				metadata.Append(", " + category);
+
+			metadata.AppendLine();
+			metadata.AppendLine("\tSheetSize: 2048");
+			metadata.AppendLine("\tEnableDepth: true");
+			metadata.AppendLine();
+
+			metadata.AppendLine("Terrain:");
+			terrainTypes.Distinct();
```
pchote: Distinct returns a new enumerable (it doesn't edit the collection in-place), so this does nothing.
This causes a whole pile of duplicated terrain types to be written when importing the TS temperate tileset.

Mailaender: Should be fixed now and for categories as well.

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
```diff
@@ -137,6 +157,41 @@ public void Run(ModData modData, string[] args)
 			{
 				// GetSection will throw when we run out of sections to import
 			}
+
+			metadata.Append("\tEditorTemplateOrder: ");
+
+			categories = categories.Distinct().ToList();
+			metadata.Append(categories.First());
+			foreach (var category in categories.Skip(1))
```
pchote: This is overkill. Something like `metadata.Append("\tEditorTemplateOrder: " + string.Join(", ", categories.Distinct()));` should work fine.

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
```diff
@@ -67,6 +85,8 @@ public void Run(ModData modData, string[] args)
 					var sectionCount = int.Parse(section.GetValue("TilesInSet", "1"));
 					var sectionFilename = section.GetValue("FileName", "").ToLowerInvariant();
 					var sectionCategory = section.GetValue("SetName", "");
+					if (!string.IsNullOrEmpty(sectionCategory) && sectionFilename != "blank")
+						categories.Add(sectionCategory);
```
pchote: This includes categories for tiles that don't exist (e.g. "MM Ramps" and "Obsolete Height Pieces").  Please move this below the loop, and use a flag that is set inside the loop by the first valid tile to ensure that only valid categories are added to the list.

pchote:
These imported tilesets won't run out of the box (since they don't include the resource terrain types), but I don't think we should be adding those here - they are not part of the legacy tilesets.  This works well except for the minor nits above, so :+1: after they are fixed.

pchote:
Removing @abcdefg30's :+1: which was left over from before this was completely rewritten.

Mailaender:
Updated.

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
Mailaender: Indeed way better.

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
Mailaender: I never intended this to be create a file that requires no edit, but this seems easy to fix.

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
```diff
@@ -137,6 +162,37 @@ public void Run(ModData modData, string[] args)
 			{
 				// GetSection will throw when we run out of sections to import
 			}
+
+			var usedCategories = categoriesUsage.Where(categoryUsed => categoryUsed.Value).Select(c => c.Key);
+			metadata.Append("\tEditorTemplateOrder: " + string.Join(", ", usedCategories));
```
Phrohdoh: `usedCategories.JoinWith(", ")`

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
```diff
@@ -57,7 +59,23 @@ public void Run(ModData modData, string[] args)
...
+			metadata.AppendLine("\tName: {0}".F(name));
+			metadata.AppendLine("\tId: {0}".F(name.ToUpperInvariant()));
+			metadata.AppendLine("\tPalette: iso{0}.pal".F(extension));
+			metadata.AppendLine("\tHeightDebugColors:  00000080, 00004480, 00008880, 0000CC80, 0000FF80, 4400CC80," +
+				" 88008880, CC004480, FF110080, FF550080, FF990080, FFDD0080, DDFF0080, 99FF0080, 55FF0080, 11FF0080");
+
 			// Loop over template sets
+			var data = new StringBuilder();
+			data.AppendLine("Templates:");
+			var categoriesUsage = new Dictionary<string, bool>();
```
pchote: If you want to go this route (which is fine I guess) you should just use a `HashSet<string>`.

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
```diff
@@ -67,6 +85,8 @@ public void Run(ModData modData, string[] args)
 					var sectionCount = int.Parse(section.GetValue("TilesInSet", "1"));
 					var sectionFilename = section.GetValue("FileName", "").ToLowerInvariant();
 					var sectionCategory = section.GetValue("SetName", "");
+					if (!string.IsNullOrEmpty(sectionCategory) && sectionFilename != "blank" && !categoriesUsage.ContainsKey(sectionCategory))
```
pchote: You can then remove these changes

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
```diff
@@ -77,9 +97,14 @@ public void Run(ModData modData, string[] args)
 
 						using (var s = modData.DefaultFileSystem.Open(templateFilename))
 						{
-							Console.WriteLine("\tTemplate@{0}:", templateIndex);
-							Console.WriteLine("\t\tCategory: {0}", sectionCategory);
-							Console.WriteLine("\t\tId: {0}", templateIndex);
+							data.AppendLine("\tTemplate@{0}:".F(templateIndex));
+							data.AppendLine("\t\tCategory: {0}".F(sectionCategory));
+							if (categoriesUsage.ContainsKey(sectionCategory))
```
pchote: And replace this with an unconditional addition to the set (sets can't contain duplicates).

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
```diff
@@ -137,6 +162,37 @@ public void Run(ModData modData, string[] args)
 			{
 				// GetSection will throw when we run out of sections to import
 			}
+
+			var usedCategories = categoriesUsage.Where(categoryUsed => categoryUsed.Value).Select(c => c.Key);
```
pchote: This filtering is then unnecessary.

Mailaender:
Finally understood the magic of hash sets.

OpenRA.Mods.TS/UtilityCommands/LegacyTilesetImporter.cs
```diff
@@ -57,7 +59,24 @@ public void Run(ModData modData, string[] args)
 				"Cliff" // TS defines this as "Rock"
 			};
 
+			var metadata = new StringBuilder();
+			metadata.AppendLine("General:");
+
+			var name = Path.GetFileNameWithoutExtension(args[1]);
```
pchote: Can we please make this a proper commandline argument instead of hacking it from the filename (which then fails 50% of the time in ts!)?

Mailaender: This now relies on the knowledge of temperat translates to Temperate, but it makes it more flexible I guess.

pchote:
Looks fine now :+1:.  Thanks.

pchote:
We're still only at 1 :+1: because @abcdefg30's initial one was negated by all the rewrites.  The tag shouldn't have been readded earlier.

Mailaender:
The "retracting" is confusing. Tools like https://zappr.opensource.zalan.do/ will count 6 :+1: from 3 individuals in this pull request.

pchote:
Thats a shame.  It should be straightforward for those tools to only count thumbs given after the last commit changes.

GraionDilach:
Could you add a black default to terraintype colors? Listing them as black does less harm than not even listing them and feels more intuitive for modders IMO.

After that, I'll gladly add the second thumb.

Mailaender:
Added:

```patch
+metadata.AppendLine("\t\tColor: 000000");
```

GraionDilach:
:+1:

Mailaender:
> That's a shame. It should be straightforward for those tools to only count thumbs given after the last commit changes.

Seeing it in action at https://github.com/zalando/zappr/pulls I might be mistaken:

>  zappr: This PR has 1/1 approvals since the last commit.

It might help us count the :+1: more easily.

PR #9339, opened 2015-09-13
Makes only concrete walls block fire

### LINKED ISSUE #9320: Firing over barbed wire and sandbags
In the original games, I believe units could fire over barbed wire and sandbags, but that it would block the paths of light vehicles and infantry.

In oRA, these seem to block projectiles too, making them more like cheap weak walls. I think it would be  more interesting to allow units to shoot through them but still block the paths of light vehicles and infantry.

That is unless it is actually a ray-caster issue that these units can't fire over them?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26618699-firing-over-barbed-wire-and-sandbags?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
And clean up TS `^Wall` default.

Closes #9320.

### DIFF
--- a/mods/cnc/rules/defaults.yaml
+++ b/mods/cnc/rules/defaults.yaml
@@ -556,7 +556,6 @@
 	Crushable:
 		CrushClasses: wall
 		CrushSound: sandbag2.aud
-	BlocksProjectiles:
 	LineBuild:
 		Range: 8
 		NodeTypes: wall
--- a/mods/cnc/rules/structures.yaml
+++ b/mods/cnc/rules/structures.yaml
@@ -892,7 +892,7 @@ BRIK:
 		Value: 0
 	Tooltip:
 		Name: Concrete Barrier
-		Description: Stops infantry and most tanks.
+		Description: Stops infantry and most tanks.\nBlocks some projectiles.
 	Buildable:
 		BuildPaletteOrder: 30
 		Prerequisites: vehicleproduction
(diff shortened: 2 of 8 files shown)


Mailaender:
You will also need to adapt the descriptions in
* https://github.com/OpenRA/OpenRA/pull/9339/files#diff-83e2e01e165e3b565b57237ce194b720R1497
* https://github.com/OpenRA/OpenRA/pull/9339/files#diff-83e2e01e165e3b565b57237ce194b720R1522

reaperrr:
Updated.

Phrohdoh:
What is the inspiration behind this?
The community or some discussion over IRC?

reaperrr:
As mentioned above, #9320 and the fact that this matches the behavior in the original games.

Mailaender:
:+1:

pchote:
Looks good, and makes a lot of sense :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/1499304129624af3c97ed3af1ecfbbd484391c81

Issue #9101, opened 2015-08-17
Lint check player palette references


Mailaender:
to complete https://github.com/OpenRA/OpenRA/issues/8689.

reaperrr:
I'm not sure whether this approach will work in cases where a single property defines the palette name and a boolean property then defines whether that palette should be treated as player palette.

@pchote's suggestion from #8749 sounds more sensible to me.
`I suggest adding overloads for the palette reference that let you specify as a bool whether or not it is a player palette, or a string that points at the bool field that determines at runtime whether it is a player palette.`

Mailaender:
It will be easier if we used:

```csharp
[PaletteReference] public readonly string Palette = null
[PlayerPaletteReference] public readonly string PlayerPalette = null
```

everywhere (at some places we already do) and using what is != null instead of a boolean switch.

pchote:
That leads to ambiguities when the mod defines both.  A single string and a bool specifying straight / player avoids that.

atlimit8:
What about a new reusable `struct` such as `OpenRA.Primatives.TaggedString<EnumType>` that is loaded from the MiniYaml code:

1. *enumValue*`,`*stringValue*
`player, player`, `fixed, terrain`, `fixed, chrome`

2. Another style is to use *stringValue[*`@`*enumValue]* with the default enum vaule if the `@` is missing:
`player@player`, `terrain`, `chrome`

3. If the `@` is confusing because of its current usage, perhaps *stringValue[*`?`*enumValue]* with the default enum vaule if the `?` is missing:
`player?player`, `terrain`, `chrome`

pchote:
That feels more complicated than it needs to be.  We are already using the pattern i've been asking for for sequences / images.

Mailaender:
You will then need to explain how your new string@string or string?string filetype works in https://github.com/OpenRA/OpenRA/wiki/Traits and I don't see how that will be easier to lint.

```csharp
[PaletteReference("IsPlayerPalette")] public readonly string Palette = null
public readonly string IsPlayerPalette = false
```

will work after some fiddling with the lint rule and the resulting configuration will be understood immediately or at least be more resistant against wrong input.

PR #9340, opened 2015-09-13
Changed D2k health, armor types and weapon stats to match original

reaperrr:
A large chunk from #9106.


mods/d2k/weapons.yaml
```diff
@@ -83,68 +106,92 @@ M_LMG:
...
+		ImpactSound: EXPLSML1.WAV
 
 HMG:
-	ReloadDelay: 30
-	Range: 5c0
-	Burst: 2
-	BurstDelay: 5
+	ReloadDelay: 20
+	Range: 3c0
+#	Burst: 3	# Actually fires 3 times in D2k but deals damage all at once, purely visual/audio effect
```
GraionDilach: Whatever uses this could have the weapon defined three times on it instead then with 1/3 damage.

Mailaender:
You need to remove that `[PaletteReference]` from https://github.com/reaperrr/OpenRA/blob/d2k-bal-weapons/OpenRA.Mods.Common/Effects/Missile.cs#L62 as that can be a player palette reference which is not yet supported. https://github.com/OpenRA/OpenRA/issues/9101

mods/d2k/weapons.yaml
reaperrr: That wouldn't reproduce the original behavior either.

I opted fro `Burst: 3` and `BurstDelay: 2` with `Damage: 60` per burst instead. That looks and sounds pretty much like the original.

reaperrr:
Updated.

Additionally improved weapons after looking at D2k original weapon stats myself:
- split Trike/Raider muzzle to separate weapon (for 3 muzzle flashes)
- reduced raider weapon range from 5 to 3 cells to match trike and original
- removed redundant contrail definitions from various MGs
- removed explosion sound from tank cannons, less annoying and matches original
- changed some other firing and explosion sounds and effects to be less annoying and better match original
- tweaked projectile speeds, should match original even more closely now
- reduced devastator weapon range from 4 to 3 cells to match original
- added projectile shadow to grenadier's grenade

OmegaBolt:
DevBullet has range 128 in D2k which should be 4 cells.

reaperrr:
> DevBullet has range 128 in D2k which should be 4 cells.

Which version of D2k? I looked at both the Gruntmod edition advertised by CnCNet as well as a seemingly vanilla edition from a abandonware site, and in both cases TibEd claims a range of 96.

OmegaBolt:
Hmm, mine should just be the original release with 1.06 patch.

penev92:
I currently have the Gruntsmods version installed and TibEd says 128:

reaperrr:
Then I don't know what's going on here. I've tried reinstalling TibEd, I've tried reinstalling Gruntmod, I've unzipped vanilla 1.06 and used the bins from that - same result all the time, TibEd stubbornly shows 96 for me.

The only explanation I can think of is that TibEd somehow somewhere stored some kind of temporary file with these values, and now reloads that all the time instead of directly reading the bin files.

OmegaBolt:
AFAIK TibEd does just that, under Data\bin in the D2k folder.

reaperrr:
I've overwritten all .bin files with the originals from the official 1.06 patch, no change.

penev92:
Well I have no idea what's going on, but let's not get hung up on that for the next 3-5 weeks. I say pick one (the higher, as the tank is slow enough already) and go with it.

reaperrr:
Done.

reaperrr:
I wanted to know for certain and did some ingame testing. It appears that the range was 3 cells in the original release, but later increased to 4 in one of the official patches (it's definitely 4 with 1.06). So it seems my TibEd just doesn't refresh some cached(?) value. 4 cells is correct in any case.

reaperrr:
Updated. Fixed:
- Grenadier projectile image
- Bazooka trail and Quad rocket explosion
- Siege Tank 155mm explosion

Mailaender:
The death hand missile is not doing any damage anymore and the repair pad doesn't allow vehicle repairs (might have regressed earlier). I also noticed that weapon ranges are extremely short now. Not sure if that really was original game balancing or is a conversion error.

reaperrr:
They really were that short in the original. However, our MoveIntoRange logic probably moves units 1 cell closer to target than the original did, so it might feel even shorter.

reaperrr:
Updated death hand damage. Should now kill most buildings when they are hit directly unless they have full health or are conyards.

mods/d2k/sequences/misc.yaml
```diff
@@ -261,6 +266,11 @@ bombs:
 		Start: 3280
 		Length: 4
 
+grenade:
+	idle: grenade.shp
```
Mailaender: Bonus points if you document which frame number you extracted those.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -59,7 +59,8 @@ class MissileInfo : IProjectileInfo
 		[Desc("Interval in ticks between each spawned Trail animation.")]
 		public readonly int TrailInterval = 2;
 
-		[PaletteReference] public readonly string TrailPalette = "effect";
+		// TODO: Re-add PaletteReference as soon as it handles player palettes
+		public readonly string TrailPalette = "effect";
```
Mailaender: I just filed https://github.com/OpenRA/OpenRA/pull/9384 so if you want to wait.

pchote:
From a technical point this gets my :+1: with or without the player palette lint pr.  I haven't made any real effort to verify that the stats match the original game, except to check that the trooper seems consistently OP in both :wink:

mods/d2k/rules/structures.yaml
```diff
@@ -49,9 +49,9 @@ construction_yard:
 	Selectable:
 		Bounds: 96,64
 	Health:
-		HP: 1000
+		HP: 3000
 	Armor:
-		Type: Concrete
+		Type: cy
```
penev92: TibEd says this uses "LIGHT" ?

penev92: OK looking at the "Damage versus" section of the rules (in TibEd), that doesn't make any sense. Also TibEd does say that "CY" stands for "Construction Yard". So I guess it's quite possible the "LIGHT" value is an error.

penev92:
Looks fine to me :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/711b5c646f895921c629fc47a6e4bb3d26b1718e...fc04f185052181dbe82f062970606394d718e4c0)

PR #9342, opened 2015-09-13
Organize map data in Manifest

penev92:
This is step one toward the so-called "diamond maps", starting off with some simple, barely related refactoring.
The point of this PR is combining `TileSize`, `TileShape` and `SubCells` from `Manifest` (`mod.yaml`) into a `MapGrid` field.
The results are most visible in TS's `mod.yaml`.

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -151,8 +151,7 @@ namespace OpenRA
 		};
 
 		public const int MaxTilesInCircleRange = 50;
-		public readonly TileShape TileShape;
-		public readonly byte MaximumTerrainHeight;
+		public readonly MapGrid Grid;
 
 		[FieldLoader.Ignore] public readonly WVec[] SubCellOffsets;
 		public readonly SubCell DefaultSubCell;
@@ -275,7 +274,7 @@ namespace OpenRA
 		public Map(TileSet tileset, int width, int height)
 		{
 			var size = new Size(width, height);
-			var tileShape = Game.ModData.Manifest.TileShape;
+			Grid = Game.ModData.Manifest.Get<MapGrid>();
 			var tileRef = new TerrainTile(tileset.Templates.First().Key, (byte)0);
 
 			Title = "Name your map here";
@@ -287,29 +286,27 @@ namespace OpenRA
 			Videos = new MapVideos();
 			Options = new MapOptions();
 
-			MapResources = Exts.Lazy(() => new CellLayer<ResourceTile>(tileShape, size));
+			MapResources = Exts.Lazy(() => new CellLayer<ResourceTile>(Grid.Type, size));
 
 			MapTiles = Exts.Lazy(() =>
 			{
-				var ret = new CellLayer<TerrainTile>(tileShape, size);
+				var ret = new CellLayer<TerrainTile>(Grid.Type, size);
 				ret.Clear(tileRef);
-				if (MaximumTerrainHeight > 0)
+				if (Grid.MaximumTerrainHeight > 0)
 					ret.CellEntryChanged += UpdateProjection;
 				return ret;
 			});
 
 			MapHeight = Exts.Lazy(() =>
 			{
-				var ret = new CellLayer<byte>(tileShape, size);
+				var ret = new CellLayer<byte>(Grid.Type, size);
 				ret.Clear(0);
-				if (MaximumTerrainHeight > 0)
+				if (Grid.MaximumTerrainHeight > 0)
 					ret.CellEntryChanged += UpdateProjection;
 				return ret;
 			});
 
 			SpawnPoints = Exts.Lazy(() => new CPos[0]);
-			TileShape = tileShape;
-			MaximumTerrainHeight = Game.ModData.Manifest.MaximumTerrainHeight;
 
 			PostInit();
 		}
@@ -374,12 +371,11 @@ namespace OpenRA
 			MapResources = Exts.Lazy(LoadResourceTiles);
 			MapHeight = Exts.Lazy(LoadMapHeight);
 
-			TileShape = Game.ModData.Manifest.TileShape;
-			MaximumTerrainHeight = Game.ModData.Manifest.MaximumTerrainHeight;
+			Grid = Game.ModData.Manifest.Get<MapGrid>();
 
-			SubCellOffsets = Game.ModData.Manifest.SubCellOffsets;
+			SubCellOffsets = Grid.SubCellOffsets;
 			LastSubCell = (SubCell)(SubCellOffsets.Length - 1);
-			DefaultSubCell = (SubCell)Game.ModData.Manifest.SubCellDefaultIndex;
+			DefaultSubCell = (SubCell)Grid.SubCellDefaultIndex;
 
 			if (Container.Exists("map.png"))
 				using (var dataStream = Container.GetContent("map.png"))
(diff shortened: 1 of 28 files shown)


OpenRA.Game/Map/Map.cs
```diff
@@ -151,8 +151,7 @@ public class Map
 		};
 
 		public const int MaxTilesInCircleRange = 50;
-		public readonly TileShape TileShape;
-		public readonly byte MaximumTerrainHeight;
+		public readonly MapGridData MapGrid;
```
pchote: We're already in `Map`, so you can simplify this down to `Grid` to remove duplication everywhere that uses it.

OpenRA.Game/Map/MapGridData.cs
```diff
@@ -0,0 +1,61 @@
...
+
+using System;
+using System.Drawing;
+using System.IO;
+
+namespace OpenRA
+{
+	public enum TileShape { None, Rectangle, Diamond }
+
+	public class MapGridData
```
pchote: Is there any specific reason for having separate `MapGridData` and `MapGrid` types?
If you combine the two you can use `FieldLoader` instead of manually loading the fields from yaml.

penev92:
Changed.

OpenRA.Game/Map/MapGrid.cs
```diff
@@ -0,0 +1,67 @@
...
+			new WVec(-299, -256, 0), // top left - index 1
+			new WVec(256, -256, 0),  // top right - index 2
+			new WVec(0, 0, 0),       // center - index 3
+			new WVec(-299, 256, 0),  // bottom left - index 4
+			new WVec(256, 256, 0),   // bottom right - index 5
+		};
+
+		readonly Dictionary<string, string> Data;
+
+		public MapGrid(MiniYaml yaml)
```
pchote: If you remove this ctor the engine will automatically load the data using FieldLoader.  You can move the subcell stuff into a [LoadUsing].

penev92:
Removed the new reasonable default for Type, sticking with an arbitrary one.

OpenRA.Game/Map/MapGrid.cs
```diff
@@ -0,0 +1,67 @@
...
+		{
+			new WVec(0, 0, 0),       // full cell - index 0
+			new WVec(-299, -256, 0), // top left - index 1
+			new WVec(256, -256, 0),  // top right - index 2
+			new WVec(0, 0, 0),       // center - index 3
+			new WVec(-299, 256, 0),  // bottom left - index 4
+			new WVec(256, 256, 0),   // bottom right - index 5
+		};
+
+		readonly Dictionary<string, string> Data;
```
Mailaender: > [SA1306] Variable names and private field names must start with a lower-case letter: Data.

pchote: This can be removed completely.

OpenRA.Game/Map/MapGrid.cs
penev92: Turned out to not be worth the hassle. It's the `SubCellDefaultIndex` that has the extra initialization logic, and those two are pretty tightly coupled.

penev92:
Updated.

OpenRA.Mods.Common/Widgets/MapPreviewWidget.cs
```diff
@@ -138,8 +140,7 @@ public override void MouseExited()
 		public int2 ConvertToPreview(CPos cell)
 		{
 			var preview = Preview();
-			var tileShape = Game.ModData.Manifest.TileShape;
-			var point = cell.ToMPos(tileShape);
+			var point = cell.ToMPos(Game.ModData.Manifest.Get<MapGrid>().Type);
```
pchote: Looks like you can use `shape` here to avoid extra lookups.

OpenRA.Game/Map/MapGrid.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace OpenRA
+{
+	public enum TileShape { Rectangle, Diamond }
+
+	public class MapGrid : IGlobalModData
+	{
+		public readonly TileShape Type;
+		public readonly Size TileSize = new Size(24, 24);
+		public readonly byte MaximumTerrainHeight;
+		public readonly byte SubCellDefaultIndex;
```
pchote: Please default this to `byte.MaxValue`

OpenRA.Game/Map/MapGrid.cs
```diff
@@ -0,0 +1,63 @@
...
+		{
+			new WVec(0, 0, 0),       // full cell - index 0
+			new WVec(-299, -256, 0), // top left - index 1
+			new WVec(256, -256, 0),  // top right - index 2
+			new WVec(0, 0, 0),       // center - index 3
+			new WVec(-299, 256, 0),  // bottom left - index 4
+			new WVec(256, 256, 0),   // bottom right - index 5
+		};
+
+		public MapGrid(MiniYaml yaml)
```
pchote: You can the replace this with the much simpler code:
```csharp
public MapGrid(MiniYaml yaml)
{
	FieldLoader.Load(this, yaml);

	// The default subcell index defaults to the middle entry
	if (SubCellDefaultIndex == byte.MaxValue)
		SubCellDefaultIndex = (byte)(SubCellOffsets.Length / 2);
	else if (SubCellDefaultIndex < (SubCellOffsets.Length > 1 ? 1 : 0) || SubCellDefaultIndex >= SubCellOffsets.Length)
		throw new InvalidDataException("Subcell default index must be a valid index into the offset triples and must be greater than 0 for mods with subcells");
}
```

pchote:
This is a nice cleanup, and doesn't break anything that I could see.  :+1: after the simple cleanups I have suggested above.

OpenRA.Mods.Common/Widgets/MapPreviewWidget.cs
penev92: No, that breaks stuff hard. I guess it's not yet initialized here?

OpenRA.Mods.Common/Widgets/MapPreviewWidget.cs
```diff
@@ -83,7 +84,8 @@ public MapPreviewWidget()
 			spawnContrastColor = ChromeMetrics.Get<Color>("SpawnContrastColor");
 			spawnLabelOffset = ChromeMetrics.Get<int2>("SpawnLabelOffset");
 
-			cellWidth = Game.ModData.Manifest.TileShape == TileShape.Diamond ? 2 : 1;
+			shape = Game.ModData.Manifest.Get<MapGrid>().Type;
```
pchote: You also need to initialize these in the other ctor.  I suspect this may be the cause of the hard breakage below.

OpenRA.Mods.Common/Widgets/MapPreviewWidget.cs
pchote: Maybe try again after fixing my nit above

OpenRA.Mods.Common/Widgets/MapPreviewWidget.cs
penev92: Yup, that would be it.

penev92:
Fixed.

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/58fbd154ca050c08abda097688100f2870add726...c894fe3f230eea75e89a323319e2e81885c9da0d)

PR #9344, opened 2015-09-14
Fixed RotationPaletteEffect tileset validity check

reaperrr:
Fixes a regression from #9103, RA temperate/snow terrain water effect is currently broken on bleed.

If `Tilesets` was empty but `ExcludeTilesets` was not, it would always return `false` even though that combination is supposed to mean only the excluded tileset(s) is/are not valid for this effect.


GraionDilach:
:+1:

Mailaender:
Thanks for the quick followup fix.

PR #9347, opened 2015-09-14
Retry loading of mix file

RoosterDragon:
Binaries built by mono and running on Windows exhibit a bug where the parsing the the mix file will occasionally produce garbage results. Retrying the operation is sufficient to mitigate the issue.

---

This should prevent any more crashes due to #2441 for now, until there is time for somebody to investigate the root cause of the problem and have a proper fix implemented.

(Many thanks to @pi99y for their tireless efforts in helping to investigate the issue)

### DIFF
--- a/OpenRA.Game/FileSystem/MixFile.cs
+++ b/OpenRA.Game/FileSystem/MixFile.cs
@@ -56,7 +56,28 @@ namespace OpenRA.FileSystem
 			this.priority = priority;
 			this.type = type;
 
-			s = GlobalFileSystem.Open(filename);
+			// HACK: Binaries built by mono that run on Windows exhibit an issue where parsing this file sometimes
+			// returns garbage results. This usually manifests in an exception trying to build the index as there will
+			// be many duplicate entries. Retrying is generally sufficient for a successful parse.
+			const int Attempts = 3;
+			for (var i = 1; i <= Attempts; i++)
+			{
+				try
+				{
+					LoadMixFile(filename, out s, out index, out dataStart);
+					break;
+				}
+				catch (ArgumentException)
+				{
+					if (i == Attempts)
+						throw;
+				}
+			}
+		}
+
+		static void LoadMixFile(string filename, out Stream stream, out Dictionary<uint, PackageEntry> index, out long dataStart)
+		{
+			var s = GlobalFileSystem.Open(filename);
 			try
 			{
 				// Detect format type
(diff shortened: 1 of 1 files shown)


OpenRA.Game/FileSystem/MixFile.cs
```diff
@@ -56,7 +56,27 @@ public MixFile(string filename, PackageHashType type, int priority)
 			this.priority = priority;
 			this.type = type;
 
-			s = GlobalFileSystem.Open(filename);
+			// HACK: Binaries built by mono that run on Windows exhibit an issue where parsing this file sometimes
+			// returns garbage results. This usually manifests in an exception trying to build the index as there will
+			// be many duplicate entries. Retrying is generally sufficient for a successful parse.
+			const int Attempts = 3;
+			for (var i = 1; i <= Attempts; i++)
```
pchote: Minor style nit: `i = 0; i < Attempts;` please.

OpenRA.Game/FileSystem/MixFile.cs
```diff
@@ -56,7 +56,27 @@ public MixFile(string filename, PackageHashType type, int priority)
...
+			{
+				try
+				{
+					LoadMixFile(filename, out s, out index, out dataStart);
```
pchote: You'll need to break on success.

RoosterDragon: Derp - done.

OpenRA.Game/FileSystem/MixFile.cs
RoosterDragon: This would require ```i == Attempts - 1``` below. Are you sure?

pchote: I guess this is fine, then.  Both options are about equally style-nitty.

RoosterDragon: I'll leave it as-is for now.

Mailaender:
:+1: Still works fine on Linux. I am actually fine with "Closes https://github.com/OpenRA/OpenRA/issues/2441" as the problem may be solved in practice although this clearly is the brute force method.

RoosterDragon:
Probably best to leave it open for now - this isn't a 100% guaranteed fix.

pchote:
I agree - lets keep this on hold for now, and we can always apply it later in the release cycle if we decide we don't want to debug this further.

RoosterDragon:
To clarify my comment: I still think this is worth merging as it is likely (although not guaranteed) to prevent the crashes - but we should keep the issue open until:
- We're happy this is actually helping.
- We either fix this properly, or decide it isn't worth the effort to debug and fix properly.

penev92:
I'm going to say it's probably not worth the time and effort to look for a proper fix.

reaperrr:
I think unless someone manages to fix the bug properly before that, we should definitely merge this before the next release, a work-around is still better than no fix at all.
At some point we should rather ditch using .mix files directly and instead extract and repack files into .zips, until then this should be good enough.

pi99y:
I agree with @reaperrr, please include it in the next playtest/release. Thanks.

pchote:
I have tracked down the actual problem (a miscompile related to the `fixed` statement), which confirms that this does not fix the problem.  I will file a PR that fixes the underlying problem.

RoosterDragon:
Sweet.

GraionDilach:
This should be removed from the Future milestone as well?

Issue #9349, opened 2015-09-15
Husks don't chronoshift back to original location


abcdefg30:

pchote:
This regression may have been introduced by the husk/viceroid spawner merge?

abcdefg30:
Shouldn't, as I didn't touch that code at all: [Diff](https://github.com/OpenRA/OpenRA/pull/8744/files#diff-72f5025743864a1bd2ce4b95283f1a7bR112)

Will test that though.

abcdefg30:
The bug seems to be already present in 1bc84275fd36a3ad2575a04947855338ecd40ac4, the commit before #8744.

obrakmann:
Is this really a regression? In which release did it use to work? Any chance that this is just a duplicate of #8670?

PR #9350, opened 2015-09-15
Add the flameguy death animation

abcdefg30:
Note one: The first commit also suffers from #9349.
Note two: Does someone have better names for `IHuskModifier` and `IHuskCreated`?
(That are just internals, but we removed `husk` from the rest of the code, too.)

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
@@ -61,18 +61,23 @@ namespace OpenRA.Mods.Common.Traits
 	public class WithDeathAnimation : INotifyKilled
 	{
 		public readonly WithDeathAnimationInfo Info;
+		readonly Crushable crushable;
 		readonly RenderSprites rs;
 
 		public WithDeathAnimation(Actor self, WithDeathAnimationInfo info)
 		{
 			Info = info;
 			rs = self.Trait<RenderSprites>();
+			crushable = self.TraitOrDefault<Crushable>();
 		}
 
 		public void Killed(Actor self, AttackInfo e)
 		{
-			// Killed by some non-standard means. This includes being crushed
-			// by a vehicle (Actors with Crushable trait will spawn CrushedSequence instead).
+			// Actors with Crushable trait will spawn CrushedSequence.
+			if (crushable != null && crushable.Crushed)
+				return;
+
+			// Killed by some non-standard means.
 			if (e.Warhead == null || !(e.Warhead is DamageWarhead))
 				return;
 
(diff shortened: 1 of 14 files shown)


OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs
```diff
@@ -45,11 +45,15 @@ public class SpawnActorOnDeathInfo : ITraitInfo
 		[Desc("Should an actor only be spawned when the 'Creeps' setting is true?")]
 		public readonly bool RequiresLobbyCreeps = false;
 
+		[Desc("Should no actor be spawned when the parent actor is crushed?")]
+		public readonly bool IgnoreCrushes = true;
```
pchote: Can we do this via a DeathType filter instead of a special case?

GraionDilach:
I almost asked if we need FallbackAnimation, but then I realized that negative SelfHealing/Burns also gives no TargetType filtering, so yeah, those three (+ crushing) cases justify.

OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
```diff
@@ -46,6 +46,9 @@ public class WithDeathAnimationInfo : ITraitInfo, Requires<RenderSpritesInfo>
 			"Is only used if UseDeathTypeSuffix is `True`.")]
 		public readonly Dictionary<string, int> DeathTypes = new Dictionary<string, int>();
 
+		[Desc("Sequence to use when the actor is killed by some non-standard means (e.g. suicide).")]
+		public readonly string FallbackSequence = null;
```
Mailaender: `[SequenceReference]` please.

Mailaender:
The lint check goes crazy https://travis-ci.org/OpenRA/OpenRA/builds/80408749#L1089 Not sure if this is a false positive. Either way it needs to be handled.

OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs
```diff
@@ -57,6 +61,7 @@ public SpawnActorOnDeath(ActorInitializer init, SpawnActorOnDeathInfo info)
 		{
 			this.info = info;
 
+			crushable = init.Self.TraitOrDefault<Crushable>();
```
Mailaender: This may not work when the trait ordering doesn't match `Crushable` before `SpawnActorOnDeath` so either follow @pchote's advice or do the trait lookup in `Killed`.

abcdefg30:
I already started on the fix @pchote suggested.
Closing this for now to clean the PR-queue up.

PR #9355, opened 2015-09-16
Catch exception when hosting server on busy port

### LINKED ISSUE #9354: Crash when hosting a game if selected port is not free
In any mod, if one attempts to host a game on a busy port, we crash with:
```
Exception of type `System.Net.Sockets.SocketException`: Address already in use
  at System.Net.Sockets.Socket.Bind (System.Net.EndPoint local_end) [0x00000] in <filename unknown>:0 
  at System.Net.Sockets.TcpListener.Start (Int32 backlog) [0x00000] in <filename unknown>:0 
  at System.Net.Sockets.TcpListener.Start () [0x00000] in <filename unknown>:0 
  at OpenRA.Server.Server..ctor (System.Net.IPEndPoint endpoint, OpenRA.ServerSettings settings, OpenRA.ModData modData) [0x00000] in <filename unknown>:0 
  at OpenRA.Game.CreateServer (OpenRA.ServerSettings settings) [0x00000] in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.Logic.ServerCreationLogic.CreateAndJoin () [0x00000] in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.ButtonWidget.<ButtonWidget>m__A (MouseInput _) [0x00000] in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.ButtonWidget.HandleMouseInput (MouseInput mi) [0x00000] in <filename unknown>:0 
  at OpenRA.Widgets.Widget.HandleMouseInputOuter (MouseInput mi) [0x00000] in <filename unknown>:0 
  at OpenRA.Widgets.Ui.HandleInput (MouseInput mi) [0x00000] in <filename unknown>:0 
  at OpenRA.DefaultInputHandler+<OnMouseInput>c__AnonStorey2.<>m__0 () [0x00000] in <filename unknown>:0 
  at OpenRA.Sync.CheckSyncUnchanged[Boolean] (OpenRA.World world, System.Func`1 fn) [0x00000] in <filename unknown>:0 
  at OpenRA.DefaultInputHandler.OnMouseInput (MouseInput input) [0x00000] in <filename unknown>:0 
  at OpenRA.Renderer.Sdl2.Sdl2Input.PumpInput (IInputHandler inputHandler) [0x00000] in <filename unknown>:0 
  at OpenRA.Renderer.Sdl2.Sdl2GraphicsDevice.PumpInput (IInputHandler inputHandler) [0x00000] in <filename unknown>:0 
  at OpenRA.Graphics.Renderer.EndFrame (IInputHandler inputHandler) [0x00000] in <filename unknown>:0 
  at OpenRA.Game.RenderTick () [0x00000] in <filename unknown>:0 
  at OpenRA.Game.Loop () [0x00000] in <filename unknown>:0 
  at OpenRA.Game.Run () [0x00000] in <filename unknown>:0 
  at OpenRA.Program.Run (System.String[] args) [0x00000] in <filename unknown>:0 
  at OpenRA.Program.Main (System.String[] args) [0x00000] in <filename unknown>:0 
AL lib: (EE) alc_cleanup: 1 device not closed
```
PR incoming.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26716255-crash-when-hosting-a-game-if-selected-port-is-not-free?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ChaoticMind:
First PR, haven't coded in C# before. But the changes should be fairly simple.

Closes #9354

Suggested changelog: "Now displays an error when attempting to host a game on a busy port instead of crashing".

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs
@@ -119,13 +119,16 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			{
 				Game.CreateServer(settings);
 			}
-			catch (System.Net.Sockets.SocketException)
+			catch (System.Net.Sockets.SocketException e)
 			{
-				ConfirmationDialogs.CancelPrompt(
-					"Server Creation Failed",
-					"Could not listen on port {0}.\n\nCheck if the port is already being used.".F(Game.Settings.Server.ListenPort),
-					cancelText: "OK");
+				var err_msg = "Could not listen on port {0}.".F(Game.Settings.Server.ListenPort);
+				if (e.ErrorCode == 10048) { // AddressAlreadyInUse (WSAEADDRINUSE)
+					err_msg += "\n\nCheck if the port is already being used.";
+				} else {
+					err_msg += "\n\nError is: \"{0}\" ({1})".F(e.Message, e.ErrorCode);
+				}
 
+				ConfirmationDialogs.CancelPrompt("Server Creation Failed", err_msg, cancelText: "OK");
 				return;
 			}
 
--- a/OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs
@@ -115,7 +115,20 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			var settings = new ServerSettings(Game.Settings.Server);
 
 			// Create and join the server
-			Game.CreateServer(settings);
+			try
+			{
+				Game.CreateServer(settings);
+			}
+			catch (System.Net.Sockets.SocketException)
+			{
+				ConfirmationDialogs.CancelPrompt(
+					"Server Creation Failed",
+					"Could not listen on port {0}.\n\nCheck if the port is already being used.".F(Game.Settings.Server.ListenPort),
+					cancelText: "OK");
+
+				return;
+			}
+
 			Ui.CloseWindow();
 			ConnectionLogic.Connect(IPAddress.Loopback.ToString(), Game.Settings.Server.ListenPort, password, onCreate, onExit);
 		}


OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs
```diff
@@ -115,9 +115,39 @@ void CreateAndJoin()
...
 			// Create and join the server
-			Game.CreateServer(settings);
+			try
+			{
+				Game.CreateServer(settings);
+			}
+			catch (System.Net.Sockets.SocketException)
+			{
+				// Console.WriteLine("Couldn't listen on port {0}, socket already in use?", settings.ListenPort);
+				Ui.OpenWindow("CREATESERVERFAIL_PANEL", new WidgetArgs()
```
Mailaender: You can use the pre-existing `ConfirmationDialogs.PromptConfirmAction` or `CancelPrompt` for this.

ChaoticMind: Thanks. I'm now using CancelPrompt(). It's a much smaller diff now. I'm still glad I did it the other way first because it forced me to figure out how the prompts work.

ChaoticMind:
Let me know if I should squash the commits before merging

AUTHORS
```diff
@@ -158,4 +159,3 @@ distributed under the LGPL version 2.1 or later.
 Finally, special thanks goes to the original teams
 at Westwood Studios and EA for creating the classic
 games which OpenRA aims to reimagine.
-
```
pchote: Please don't remove the end-of-file newlines.

pchote:
The travis tests failed:
```
Checking for code style violations in OpenRA.Mods.Common...

OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs:L126: [SA1118] The parameter spans multiple lines. If the parameter is short, place the entire parameter on a single line. Otherwise, save the contents of the parameter in a temporary variable and pass the temporary variable as a parameter. 
```

ChaoticMind:
Oops, could have sworn I checked it before pushing. Updated.

OpenRA.Mods.Common/Widgets/Logic/ConnectionLogic.cs
```diff
@@ -142,4 +142,4 @@ public ConnectionFailedLogic(Widget widget, OrderManager orderManager, Action on
 			}
 		}
 	}
-}
\ No newline at end of file
+}
```
pchote: Please also undo the end-of-file newline removal here and in the other files.

pchote: note: it always pays to click on the "Files changed" tab above and check that you've only committed things that are directly related to the pr.

mods/ra/chrome/connection.yaml
```diff
@@ -69,7 +69,7 @@ Background@CONNECTING_PANEL:
 		Label@CONNECTING_TITLE:
 			X: 0
 			Y: 20
-			Width: 450
+			Width: PARENT_RIGHT
```
pchote: This doesn't change anything in practice, so you may as well drop it too.

OpenRA.Mods.Common/Widgets/Logic/ConnectionLogic.cs
ChaoticMind: Actually, here it's adding an end-of-line newline (which didn't exist). I assume you want this, or? (since both git in diffs and github usually complain/warn about that).

The other two are actually two new lines at end of file (instead of just one), that my editor automatically removed. I initially kept them cause why not, but I'll get rid of them.

pchote: There are various automatic processes that like to fight over the end-of-file whitespace style, so our policy is for PRs to leave them as they found them, unless the pr is specifically dealing with that on a project-wide basis :)

ChaoticMind:
squashed!

RoosterDragon:
[The documentation](https://msdn.microsoft.com/en-us/library/xt2x8705%28v=vs.110%29.aspx) suggests the exception you're catching here could be for any error trying to set up the connection - not just the port being in use.

You may want to generalise the error message a bit in case the cause was something else. You might also consider adding the ```Message``` property from the exception into the error text which hopefully will allow you to report the specific problem.

ChaoticMind:
Good call. Updated. 

I'm not very satisfied with the error message. I prefer something like "The port is already being used by another application.", but that requires a larger panel.. I initially had its own panel implemented before I was pointed out to reusing CancelPrompt(), which is better.
I can generalize the widget and make it (somehow) dynamically set its size by taking optional weight/height parameters, but that should probably be a different PR, if you'll take it. It seems valuable to have a versatile generic widget for presenting simple messages such as this.

Mailaender:

image

Easiest way to reproduce is starting two intances of OpenRA and start a multiplayer server.

abcdefg30:
:+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c6259c6cecd9ddd67793f93c06e139f7090b405e...4bb14e4bdc5efc6658854e2d5152d8f69c6a8e22)

PR #9359, opened 2015-09-16
Fixed pooling of layers used for pathfinding

RoosterDragon:
The previous implementation:
- Was failing to dispose of pooled layers.
- Was using a finalizer to allow undisposed layers to be reused.

This means all pooled layers are kept alive indefinitely until the map changes. If the finalizer is slow for any reason then the pathfiinder will allocate new layers when the pool runs out. Since these new layers are eventually stuffed back into the pool when the finalizer does run, this can theoretically leak unbounded memory until the pool goes out of scope. In practice it would leak tens of megabytes.

The new implementation ensures layers are disposed and pooled correctly to allow proper memory reuse. It also introduces some safeguards against memory leaks:
- A cap is set on the number of pooled layers. If more concurrent layers are needed than this, then the excess layers will not be pooled but instead be allowed to be garbage collected.
- No finalizer. An implementation that fails to call dispose simply allows the layer to be garbage collected instead.

---

Proposed changelog: Lump in with "Improved in-game performance and memory usage."


chrisforbes:
:+1:

Mailaender:
Review assistance: https://github.com/OpenRA/OpenRA/pull/9359/files?w=1

Can't spot any regressions in-game. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d2e9fc724c1938ae8247652a6bbaeade768a06d2...d47d7607d2ffbb95ecab7aba7df9fd9a68d32ab5)

PR #9361, opened 2015-09-17
Extends MapChooserLogic.mapFilter to match against PlayerCount

visav:
`MapChooserLogic.EnumerateMaps` currently matches `mapFilter` against map's `Title` and `Author`.
This PR extends `Where` predicate to match against `PlayerCount`. 

When choosing a map I often like to browse maps by number of players.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/MapChooserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/MapChooserLogic.cs
@@ -200,11 +200,16 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 		void EnumerateMaps(MapClassification tab, ScrollItemWidget template)
 		{
+			int playerCountFilter;
+			if (!int.TryParse(mapFilter, out playerCountFilter))
+				playerCountFilter = -1;
+
 			var maps = tabMaps[tab]
 				.Where(m => gameMode == null || m.Type == gameMode)
 				.Where(m => mapFilter == null ||
 					m.Title.IndexOf(mapFilter, StringComparison.OrdinalIgnoreCase) >= 0 ||
-					m.Author.IndexOf(mapFilter, StringComparison.OrdinalIgnoreCase) >= 0)
+					m.Author.IndexOf(mapFilter, StringComparison.OrdinalIgnoreCase) >= 0 ||
+					m.PlayerCount == playerCountFilter)
 				.OrderBy(m => m.PlayerCount)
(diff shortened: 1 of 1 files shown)


Mailaender:
Hi, thanks for making your first pull request here. :confetti_ball: 

Took me a while to understand this, but it works as promised. :white_check_mark: 

image

The false positive is inevitable when filtering through all fields.

visav:
> The false positive is inevitable when filtering through all fields.

True. 

I have not tested against all reasonable counts yet. I think that those false positives (riderr3) are acceptable, because the search text is found in map data presented to the user. That said, __inputs such as "8 players" will fail__ with the currently proposed changes.

Adding support for filter parameters like in github issues page could be an option if one wanted exact results. To me there's too much overhead for the user in doing so.

GraionDilach:
Another option in this case could be having another dropdown list between Filter and GameMode where people could select playercount.

pchote:
The current approach seems fine to me - I don't see much advantage to getting bogged down with discussions about doing major rewrites to give minor improvements.

OpenRA.Mods.Common/Widgets/Logic/MapChooserLogic.cs
```diff
@@ -200,11 +200,15 @@ void SetupGameModeDropdown(MapClassification tab, DropDownButtonWidget gameModeD
 
 		void EnumerateMaps(MapClassification tab, ScrollItemWidget template)
 		{
+			int playerCountFilter = -1;
+			int.TryParse(mapFilter, out playerCountFilter);
```
pchote: `playerCountFilter` is set to 0 if the conversion fails, which introduces an edge case: If you have a map installed with no players (e.g. a map that you're working on in the editor) it will show up for all non-numeric search terms.

You can fix this using the following:
```csharp
int playerCountFilter;
if (!int.TryParse(mapFilter, out playerCountFilter))
	playerCountFilter = -1;
```

pchote:
This works well ingame.  :+1: after fixing the issue I pointed out above.

visav:
Fixed as proposed by @pchote.

atlimit8:
Works.

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/dbe2066114345fdb66cd8e31c2578e6d06542cab...318ceea6a4015e59d9f5b9cd003a450dae0946a3)

PR #9363, opened 2015-09-17
Fix husks of chronosphered actors not being teleported back

### LINKED ISSUE #9349: Husks don't chronoshift back to original location

obrakmann:
Fixes #9349

This may regress horribly. I'm not sure in which other cases `teleporter` was null before. All of those (if any) would be affected by this change.

Edit: it looks like using the `Chronoshift*Init`s would be the only case where `teleporter` could be null, which means only husks and custom maps using those inits would be affected.

### DIFF
--- a/OpenRA.Mods.RA/Activities/Teleport.cs
+++ b/OpenRA.Mods.RA/Activities/Teleport.cs
@@ -102,8 +102,7 @@ namespace OpenRA.Mods.RA.Activities
 
 		CPos? ChooseBestDestinationCell(Actor self, CPos destination)
 		{
-			if (teleporter == null)
-				return null;
+			var shroudOwner = teleporter != null ? teleporter.Owner : self.Owner;
 
 			var restrictTo = maximumDistance == null ? null : self.World.Map.FindTilesInCircle(self.Location, maximumDistance.Value);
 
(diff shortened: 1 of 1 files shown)


Mailaender:
I can't spot negative side effects in-game. :white_check_mark:

atlimit8:
This case should be tested:
1. Player A chronoshifts actor C from somewhere unexplored by player B using a chronosphere.
2. Player B captures actor C and never explores actor chronoshift C's origin.
3. Chronoshift expires.

This change would abort the return because Player B never saw the origin of actor C's chronoshift.

obrakmann:
We could cache the initiating player in `Chronoshiftable`, then capturing wouldn't matter. Would probably conflict with your new PR though.

atlimit8:
See IRC logs for discussion.

obrakmann:
I'm closing this for now in favour of #9415. My take-away from the IRC logs is that we want to remove the shroud checks from `Chronoshiftable` completely, eventually. They are most likely redundant as the OrderTargeter (iirc) already contains the same checks.

PR #9365, opened 2015-09-18
HealUnitsCrateAction.Activate() use ActorsWithTrait<Health>()

atlimit8:
`ActorsWithTrait<Health>()` only accesses actors with a `Health` trait as opposed to going through looking for those with a `Health` trait.
It is the only place I found using `World.Actors` and `.Trait[OrDefault]<TTrait>` instead of `ActorsWithTrait`.


RoosterDragon:
:+1:

Mailaender:
Soviet 01 is a good test case here. There is one hidden in the church. :white_check_mark:

PR #9368, opened 2015-09-18
Replace only use of Linq syntax

atlimit8:
`Harvester.ClosestProc(Actor self, Actor ignore)` is the only place where Linq syntax is used instead of writing calls to the Linq methods directly.

This should fit better with the rest of the code base.


RoosterDragon:
:+1:

LavenderMoon:
Wow, I didn't even know that syntax was possible.

PR #9369, opened 2015-09-18
Added thread-affinity checks to SDL2 renderer

RoosterDragon:
If a call is made into a graphics resource that has thread-affinity, from a thread other than the one that created the graphics device, an exception will now be thrown to make debugging easier.


pchote:
Good idea.  Code changes look good to me, and the game still runs so :+1:.

Mailaender:
Thanks.

PR #9370, opened 2015-09-18
Adds FindActorsInAnnulus and tweaks AI scan for enemies

reaperrr:
An attempt to improve performance of AI scans for nearby enemies by only checking the last third of the radius if nothing was found in the first 2/3.

Additionally gave ChooseBuildLocation check a minimum distance of 2 cells, as the center cell and the 8 adjacent cells will normally be occupied by the construction yard (or its bib) anyway.

### DIFF
--- a/OpenRA.Game/WorldUtils.cs
+++ b/OpenRA.Game/WorldUtils.cs
@@ -32,12 +32,25 @@ namespace OpenRA
 
 		public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origin, WDist r)
 		{
-			using (new PerfSample("FindUnitsInCircle"))
+			using (new PerfSample("FindActorsInCircle"))
 			{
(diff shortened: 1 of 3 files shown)


RoosterDragon:
Does this improve performance in practise? I'm wondering if the extra cost of sometimes running a second ```FindActorsInAnnulus``` outweighs the savings of doing an initially smaller search.

reaperrr:
I was assuming that the performance impact primarily depends on the number of cells that are checked. If that's the case, performance for the smaller search should be a decent bit better and both searches combined should perform the same or only slightly slower than the original, large search.

Admittedly I'm not sure how much overhead a 2nd search might add regardless of the number of checked cells, though.

reaperrr:
As benefits are not guaranteed and I have enough other stuff on my to-do list, I'm closing this for now. Might return later.

PR #9371, opened 2015-09-18
Changed make.ps1 test order

reaperrr:
- changed make.ps1 yaml test order to match makefile/Travis
- shortened `dependencies` command to `deps`


penev92:
I guess you need to update the AppVeyor build script, because it tried to `make dependencies`.

Phrohdoh:
This introduces inconsistency between `make.ps1` and `Makefile`.

reaperrr:
Removed the dep shortening, the changed test order was more important anyway.

Mailaender:
:+1:

PR #9373, opened 2015-09-19
Prep 1508

marcos111:
Hi! 

Is possible add the hotkey for "move to selected point and deploy" for the chooper, APC .....? I think it will be very helpful.

thanks


Mailaender:
Please use https://github.com/OpenRA/OpenRA/issues/new instead.

Phrohdoh:
Sorry, I deleted this branch without reading the name.
I thought this was an accidental PR that pulled in random commits.

PR #9377, opened 2015-09-19
Fixed TS crane overlay's last frame hiding idle overlay

### LINKED ISSUE #9319: Last (or first) frame of WithBuildingPlacedOverlay overlays idle overlay
Currently only visibly affects the TS construction yard.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26617779-last-or-first-frame-of-withbuildingplacedoverlay-overlays-idle-overlay?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
TS crane overlay's last frame no longer overlaps idle overlay. `WithBuildingPlacedOverlay` is now only visible while playing the animation, rather than sticking around.

Fixes #9319.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithBuildingPlacedOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithBuildingPlacedOverlay.cs
@@ -35,6 +35,7 @@ namespace OpenRA.Mods.Common.Traits
 	{
 		readonly Animation overlay;
 		bool buildComplete;
+		bool visible;
 
 		public WithBuildingPlacedOverlay(Actor self, WithBuildingPlacedOverlayInfo info)
 		{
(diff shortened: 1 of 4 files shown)


pchote:
Are all the other overlay traits going to have similar problems?

OpenRA.Mods.Common/Traits/Render/WithBuildingPlacedOverlay.cs
```diff
@@ -35,6 +35,7 @@ public class WithBuildingPlacedOverlay : INotifyBuildComplete, INotifySold, INot
 	{
 		readonly Animation overlay;
 		bool buildComplete;
+		bool hide;
```
pchote: Might be a bit more consistent with other code to call this `visible` and invert the checks.

reaperrr:
> Are all the other overlay traits going to have similar problems?

Possibly. I'll take a look.

OpenRA.Mods.Common/Traits/Render/WithBuildingPlacedOverlay.cs
reaperrr: Ok.

reaperrr:
Fixed `WithRepairOverlay` as well. Looks like all other overlays have something comparable in place already.

Mailaender:
:+1: As a nice side effect this also makes the cloaking look better.

pchote:
Confirmed fixes :+1:.  Thanks.

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/bf15da7fc151ed12b745d08315a37e74f1e07b04

PR #9384, opened 2015-09-19
Added a lint check for player palette references.

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/9101.

### DIFF
--- a/OpenRA.Game/Traits/LintAttributes.cs
+++ b/OpenRA.Game/Traits/LintAttributes.cs
@@ -45,9 +45,28 @@ namespace OpenRA.Traits
 	public sealed class UpgradeUsedReferenceAttribute : Attribute { }
 
 	[AttributeUsage(AttributeTargets.Field)]
-	public sealed class PaletteDefinitionAttribute : Attribute { }
+	public sealed class PaletteDefinitionAttribute : Attribute
+	{
+		public readonly bool IsPlayerPalette;
+		public PaletteDefinitionAttribute(bool isPlayerPalette = false)
+		{
+			IsPlayerPalette = isPlayerPalette;
+		}
+	}
 
-	// TODO: differentiate player palettes
 	[AttributeUsage(AttributeTargets.Field)]
-	public sealed class PaletteReferenceAttribute : Attribute { }
+	public sealed class PaletteReferenceAttribute : Attribute
+	{
+		public readonly bool IsPlayerPalette;
+		public PaletteReferenceAttribute(bool isPlayerPalette = false)
+		{
+			IsPlayerPalette = isPlayerPalette;
+		}
+
+		public readonly string PlayerPaletteReferenceSwitch;
+		public PaletteReferenceAttribute(string playerPaletteReferenceSwitch)
+		{
+			PlayerPaletteReferenceSwitch = playerPaletteReferenceSwitch;
+		}
+	}
 }
--- a/OpenRA.Mods.Common/Lint/CheckPalettes.cs
+++ b/OpenRA.Mods.Common/Lint/CheckPalettes.cs
@@ -18,9 +18,12 @@ namespace OpenRA.Mods.Common.Lint
 {
 	class CheckPalettes : ILintRulesPass
 	{
+		List<string> palettes = new List<string>();
+		List<string> playerPalettes = new List<string>();
+
 		public void Run(Action<string> emitError, Action<string> emitWarning, Ruleset rules)
 		{
-			var palettes = GetPalettes(emitError, rules).ToList();
+			GetPalettes(emitError, rules);
 
 			foreach (var actorInfo in rules.Actors)
 			{
@@ -29,14 +32,36 @@ namespace OpenRA.Mods.Common.Lint
 					var fields = traitInfo.GetType().GetFields();
 					foreach (var field in fields.Where(x => x.HasAttribute<PaletteReferenceAttribute>()))
 					{
+						var isPlayerPalette = false;
+
+						var paletteReference = field.GetCustomAttributes<PaletteReferenceAttribute>(true).FirstOrDefault();
+						if (paletteReference != null)
+						{
+							if (!string.IsNullOrEmpty(paletteReference.PlayerPaletteReferenceSwitch))
+							{
+								var fieldInfo = fields.First(f => f.Name == paletteReference.PlayerPaletteReferenceSwitch);
+								isPlayerPalette = (bool)fieldInfo.GetValue(traitInfo);
+							}
+							else
+								isPlayerPalette = paletteReference.IsPlayerPalette;
+						}
+
 						var references = LintExts.GetFieldValues(traitInfo, field, emitError);
 						foreach (var reference in references)
 						{
 							if (string.IsNullOrEmpty(reference))
 								continue;
 
-							if (!palettes.Contains(reference))
-								emitError("Undefined palette reference {0} detected at {1}".F(reference, traitInfo));
+							if (isPlayerPalette)
+							{
+								if (!playerPalettes.Contains(reference))
+									emitError("Undefined player palette reference {0} detected at {1} for {2}".F(reference, traitInfo, actorInfo.Key));
+							}
+							else
+							{
+								if (!palettes.Contains(reference))
+									emitError("Undefined palette reference {0} detected at {1} for {2}".F(reference, traitInfo, actorInfo.Key));
+							}
 						}
 					}
 				}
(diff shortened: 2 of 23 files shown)


OpenRA.Mods.Common/Traits/Render/RenderSprites.cs
```diff
@@ -36,7 +36,7 @@ public class RenderSpritesInfo : IRenderActorPreviewInfo, ITraitInfo
 		[PaletteReference] public readonly string Palette = null;
 
 		[Desc("Custom PlayerColorPalette: BaseName")]
-		public readonly string PlayerPalette = "player";
+		[PaletteDefinition(true)] public readonly string PlayerPalette = "player";
```
reaperrr: Shouldn't this be `[PaletteReference(true)]`?

Mailaender: Fixed.

reaperrr:
:+1:

OpenRA.Mods.Common/Traits/Render/RenderSprites.cs
```diff
@@ -36,7 +36,9 @@ public class RenderSpritesInfo : IRenderActorPreviewInfo, ITraitInfo
 		[PaletteReference] public readonly string Palette = null;
 
 		[Desc("Custom PlayerColorPalette: BaseName")]
-		public readonly string PlayerPalette = "player";
+		[PaletteReference("IsPlayerPalette")] public readonly string PlayerPalette = "player";
```
pchote: Calling this `PlayerPalette` doesn't make sense after adding the `IsPlayerPalette` property.

OpenRA.Mods.Common/Traits/World/PaletteFromPlayerPaletteWithAlpha.cs
```diff
@@ -21,11 +21,13 @@ class PaletteFromPlayerPaletteWithAlphaInfo : ITraitInfo
 	{
 		[FieldLoader.Require]
 		[Desc("The prefix for the resulting player palettes")]
-		public readonly string BaseName = null;
+		[PaletteDefinition(true)] public readonly string BaseName = null;
 
 		[FieldLoader.Require]
 		[Desc("The name of the player palette to base off.")]
-		public readonly string BasePalette = null;
+		[PaletteReference("IsPlayerPalette")] public readonly string BasePalette = null;
```
pchote: The description will need to be updated to reflect the now-optional player-paletteness.

OpenRA.Mods.Common/Traits/Render/RenderSprites.cs
reaperrr: Yeah, I said it should be "true" for a reason.

We should probably merge this with the `Palette` property above, but that can be done in a follow-up PR.

Mailaender:
Updated.

OpenRA.Mods.Common/Traits/Render/RenderSprites.cs
```diff
@@ -36,7 +36,7 @@ public class RenderSpritesInfo : IRenderActorPreviewInfo, ITraitInfo
 		[PaletteReference] public readonly string Palette = null;
 
 		[Desc("Custom PlayerColorPalette: BaseName")]
-		public readonly string PlayerPalette = "player";
+		[PaletteReference(null, true)] public readonly string PlayerPalette = "player";
```
atlimit8: Awkward, should be `[PaletteReference(true)]`.

OpenRA.Mods.Common/Traits/World/PaletteFromPlayerPaletteWithAlpha.cs
```diff
@@ -21,11 +21,11 @@ class PaletteFromPlayerPaletteWithAlphaInfo : ITraitInfo
 	{
 		[FieldLoader.Require]
 		[Desc("The prefix for the resulting player palettes")]
-		public readonly string BaseName = null;
+		[PaletteDefinition(true)] public readonly string BaseName = null;
 
 		[FieldLoader.Require]
 		[Desc("The name of the player palette to base off.")]
-		public readonly string BasePalette = null;
+		[PaletteReference(null, true)] public readonly string BasePalette = null;
```
atlimit8: Same as above.

OpenRA.Game/Traits/LintAttributes.cs
```diff
@@ -45,9 +45,24 @@ public sealed class UpgradeGrantedReferenceAttribute : Attribute { }
...
+	}
 
-	// TODO: differentiate player palettes
 	[AttributeUsage(AttributeTargets.Field)]
-	public sealed class PaletteReferenceAttribute : Attribute { }
+	public sealed class PaletteReferenceAttribute : Attribute
+	{
+		public readonly string PlayerPaletteReferenceSwitch;
+		public readonly bool IsPlayerPalette;
+		public PaletteReferenceAttribute(string playerPaletteReferenceSwitch = null, bool isPlayerPalette = false)
```
atlimit8: This should be 
```c#
public PaletteReferenceAttribute(bool isPlayerPalette = false)
...
public PaletteReferenceAttribute(string playerPaletteReferenceSwitch)
```

penev92:
@atlimit8's suggestions make sense, and should probably be applied if possible.
This also needs a rebase.
Also #9340 was merged, so this should account for [this TODO](https://github.com/OpenRA/OpenRA/pull/9340/files#diff-f3450a5dcb328e0ab733404f5c74fc32R62).

OpenRA.Game/Traits/LintAttributes.cs
Mailaender: Good idea. Done.

Mailaender:
Updated.

OpenRA.Mods.Common/Lint/CheckPalettes.cs
```diff
@@ -75,9 +120,15 @@ static IEnumerable<string> GetPalettes(Action<string> emitError, Ruleset rules)
 					var fields = traitInfo.GetType().GetFields();
 					foreach (var field in fields.Where(x => x.HasAttribute<PaletteDefinitionAttribute>()))
 					{
+						var paletteDefinition = field.GetCustomAttributes<PaletteDefinitionAttribute>(true).FirstOrDefault();
```
penev92: `.First()` please. the `.Where()` just above ensures that there is one, and the `if` below will stop complaining about a "Possible null reference exception".

OpenRA.Mods.Common/Lint/CheckPalettes.cs
```diff
@@ -29,14 +32,36 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Ruleset ru
...
 					foreach (var field in fields.Where(x => x.HasAttribute<PaletteReferenceAttribute>()))
 					{
+						var isPlayerPalette = false;
+
+						var paletteReference = field.GetCustomAttributes<PaletteReferenceAttribute>(true).FirstOrDefault();
+						if (paletteReference != null)
+						{
+							if (!string.IsNullOrEmpty(paletteReference.PlayerPaletteReferenceSwitch))
+							{
+								var fieldInfo = fields.FirstOrDefault(f => f.Name == paletteReference.PlayerPaletteReferenceSwitch);
```
penev92: This should probably also be `.First()`, just to get a slightly better exception if it crashes (with or without the `OrDefault` part it will crash if there is no such field; that is good because it will indicate a major error in the trait's Info class that needs to be addressed).

OpenRA.Mods.Common/Lint/CheckPalettes.cs
```diff
@@ -48,16 +73,36 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Ruleset ru
...
+					var paletteReference = field.GetCustomAttributes<PaletteReferenceAttribute>(true).FirstOrDefault();
+					if (paletteReference != null)
 					{
-						var references = LintExts.GetFieldValues(projectileInfo, field, emitError);
-						foreach (var reference in references)
+						if (!string.IsNullOrEmpty(paletteReference.PlayerPaletteReferenceSwitch))
 						{
-							if (string.IsNullOrEmpty(reference))
-								continue;
+							var fieldInfo = fields.FirstOrDefault(f => f.Name == paletteReference.PlayerPaletteReferenceSwitch);
```
penev92: Same here.

penev92:
:+1: with the simple nits above.

Mailaender:
Updated.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/a494331b25ece3aeaa514da814054863bd21e202...a12bb503fc1505b1d77f769af0f543c8820acc6c)

PR #9386, opened 2015-09-19
Finish wrapping ActorInfo.Traits

atlimit8:
With this, _actorInfo_`.Traits` is no longer accessible preventing direct access issues with a compile-time error.

Added:

``` c#
        public IEnumerable<ITraitInfo> TraitInfos() { return traits.WithInterface<ITraitInfo>(); }
        public IEnumerable<T> TraitInfos<T>() where T : ITraitInfo { return traits.WithInterface<T>(); }
```

### DIFF
--- a/OpenRA.Game/GameRules/ActorInfo.cs
+++ b/OpenRA.Game/GameRules/ActorInfo.cs
@@ -67,6 +67,13 @@ namespace OpenRA
 			}
 		}
 
+		public ActorInfo(string name, params ITraitInfo[] traitInfos)
+		{
+			Name = name;
+			foreach (var t in traitInfos)
+				Traits.Add(t);
+		}
+
 		static Dictionary<string, MiniYaml> GetParents(MiniYaml node, Dictionary<string, MiniYaml> allUnits)
 		{
 			return node.Nodes.Where(n => n.Key == "Inherits" || n.Key.StartsWith("Inherits@"))
--- a/OpenRA.Game/GameRules/ActorInfo.cs
+++ b/OpenRA.Game/GameRules/ActorInfo.cs
@@ -187,5 +187,6 @@ namespace OpenRA
 		public bool HasTraitInfo<T>() where T : ITraitInfo { return Traits.Contains<T>(); }
 		public T TraitInfo<T>() where T : ITraitInfo { return Traits.Get<T>(); }
 		public T TraitInfoOrDefault<T>() where T : ITraitInfo { return Traits.GetOrDefault<T>(); }
+		public IEnumerable<T> TraitInfos<T>() where T : ITraitInfo { return Traits.WithInterface<T>(); }
 	}
 }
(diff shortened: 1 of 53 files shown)


pchote:
I'm not a big fan of folding folding / hiding `ITraitInfo` under the `I*Info` tag interfaces.  I understand why you've done it (to satisfy the query constraint), but the naming inconsistencies are a bit irksome.  This isn't a request, just a moan :)

OpenRA.Game/GameRules/ActorInfo.cs
```diff
@@ -184,8 +184,13 @@ static IEnumerable<Type> PrerequisitesOf(ITraitInfo info)
...
-		public T TraitInfo<T>() where T : ITraitInfo { return Traits.Get<T>(); }
-		public T TraitInfoOrDefault<T>() where T : ITraitInfo { return Traits.GetOrDefault<T>(); }
+		public bool HasTraitInfo<T>() where T : ITraitInfo { return traits.Contains<T>(); }
+		public T TraitInfo<T>() where T : ITraitInfo { return traits.Get<T>(); }
+		public T TraitInfoOrDefault<T>() where T : ITraitInfo { return traits.GetOrDefault<T>(); }
+		public IEnumerable<ITraitInfo> TraitInfos() { return traits.WithInterface<ITraitInfo>(); }
+		public IEnumerable<T> TraitInfos<T>() where T : ITraitInfo { return traits.WithInterface<T>(); }
+
+		// For testing only
+		public void AddTraitInfo(ITraitInfo traitInfo) { traits.Add(traitInfo); }
```
pchote: This is a bit ugly.  Can you make the unit test use the standard mechanism for adding traitinfos, or hack it in that one place using reflection?

OpenRA.Game/GameRules/ActorInfo.cs
```diff
@@ -184,8 +184,13 @@ static IEnumerable<Type> PrerequisitesOf(ITraitInfo info)
 					i.Name.Replace("Init", ""), i));
 		}
 
-		public bool HasTraitInfo<T>() where T : ITraitInfo { return Traits.Contains<T>(); }
-		public T TraitInfo<T>() where T : ITraitInfo { return Traits.Get<T>(); }
-		public T TraitInfoOrDefault<T>() where T : ITraitInfo { return Traits.GetOrDefault<T>(); }
+		public bool HasTraitInfo<T>() where T : ITraitInfo { return traits.Contains<T>(); }
+		public T TraitInfo<T>() where T : ITraitInfo { return traits.Get<T>(); }
+		public T TraitInfoOrDefault<T>() where T : ITraitInfo { return traits.GetOrDefault<T>(); }
+		public IEnumerable<ITraitInfo> TraitInfos() { return traits.WithInterface<ITraitInfo>(); }
```
pchote: In some places you use `TraitInfos()` and in others you use `TraitInfos<ITraitInfo>()`.  I think it would be clearer to use `TraitInfos<ITraitInfo>()` everywhere and then remove this version.

pchote:
Looks good overall, but see the two comments above.

atlimit8:
Changed.
Although, the utility mod support does seem a little out of place being stashed in this PR.

pchote:
Was the utility stuff there just so that you can access the ObjectCreator?  If so, I think a much better approach would be to pass it in to the ActorInfo via the ctor, and have the test create its own instead of initializing a fake mod.

atlimit8:
Changed.

pchote:
Thanks for working through my nits. Looks good to me now :+1:

penev92:
:+1:

PR #9387, opened 2015-09-19
Add game speed dropdown to the mission browser.

pchote:
screen shot 2015-09-19 at 20 37 27

This exposes a game speed dropdown in the mission browser, and makes the difficulty dropdown visible for all mods.  Also renames Default &rarr; Normal speed to be consistent with the difficulties.


OpenRA.Mods.Common/Widgets/Logic/MissionBrowserLogic.cs
```diff
@@ -186,27 +189,58 @@ void SelectMap(Map map)
 			descriptionPanel.ScrollToTop();
 			descriptionPanel.Layout.AdjustChildren();
 
-			difficultyButton.IsVisible = () => map.Options.Difficulties.Any();
-			if (!map.Options.Difficulties.Any())
-				return;
```
reaperrr: > OpenRA.Mods.Common/Widgets/Logic/MissionBrowserLogic.cs:L192: [SA1507] The code must not contain multiple blank lines in a row.

pchote:
Updated.

clemty:
looks good! But why does the player need to set the speed before mission start? Original TD had the speed setting on the same ingame options screen as the volume.

GraionDilach:
A) In TS and RA2, triggers could malfunction on nondefault game speeds which is why they had that slider disabled during a campaign.
B) IIRC, the dynamic changing of game speed settings are still not supported.

Mailaender:
:+1: Works as promised. I see two UI polish issues, but I don't want to derail the discussion here.

pchote:
What are the polish issues?

Mailaender:
https://github.com/OpenRA/OpenRA/issues/9392 and https://github.com/OpenRA/OpenRA/issues/9393 but you don't need to address them here.

pchote:
I think I should fix #9393 here first.

pchote:
Updated.  I have trimmed away the starfield at the top and bottom of the d2k mission previews so that we have enough room to display the mission description.  This conflicts with #9402.

pchote:
Updated again to fix the fmv issue.

I've also reduced the width of the panel, and have fixed the aspect ratio of the FMVs:

screen shot 2015-09-27 at 13 58 32

screen shot 2015-09-27 at 13 58 21

obrakmann:
Looks fine :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/e6de1eee8d83de1e12b697f486e393bad1fc004b)

AoAGeneral:
Tested this in the current bleed version. A nice little addition but im not sure if this is supposed to happen or not.

All the units and buildings have increased time speeds. IE: normal gameplay weapon factories build at 48 seconds. Fastest they build in 24 seconds. This means buggies build in nearly 6 seconds. And since planes fly faster as well they get dropped off nearly instantly which can easily overpower GDI.

Again im not sure if its supposed to speed up production speeds or if its supposed to just speed up the game itself.

pchote:
The game speed is a global modifier.  *Everything* runs faster or slower by the same factor.

AoAGeneral:
Aah I gotcha!

Issue #6241, opened 2014-08-12
Lag causes replays to stop [$58]


ghost:
ive noticed this in several replays , if the (real) game got slowed down because of lags (but the whole game went fine) it can cause replays to stop the playback after a certain amount of minutes (being watched) and remain that way ...

so basicly , you can loose (good?) replays if there was connection problems involved during the real game

bnemer:
Same happens when a player disconnects, or is dropped after a minute of disconnection.

fgdzfggsfgsfgsf:
Here are a few example replays where this happens.
https://www.dropbox.com/s/fyjvioc2vkyvm3g/Openra%2020150424%20release%20replays%20freezing.zip?dl=0

grouzen:
Why this bug is still open? ;)

Unit158:
I understand you're being slightly sarcastic, but none of the devs so far have wanted to do the debugging to figure out why this happens. Just because there is money associated doesn't mean it's worth the time to do.

Herve-M:
As fist see it's impossible to load an old replay with the last release or bleed version.
Even if I work with the good version (for replay from @fgdzfggsfgsfgsf release-20150424), I can't debug (attach to process) the released version without the debug symbol.

So is it possible to load a replay from release into a dev version ? (as using same release / dev version)

penev92:
You can `git checkout release-20150424` and build that. Then you will need to change the mod's version in mod.yaml to `release-20150424` and you can then load the replay.

Herve-M:
@penev92 Thanks for info !

PR #9389, opened 2015-09-19
Fix replay freeze

Herve-M:
This PR normally fix the freezing problem when a player disconnect from game.
- Get the list of client 
  - Get all clients + last frame [ReplayConnection.cs#L59-L65](https://github.com/Herve-M/OpenRA/blob/fix-replay/OpenRA.Game/Network/ReplayConnection.cs#L59-L65)
- Correct the datas
  - Parsing (2nd pass) [ReplayConnection.cs#L97-L112](https://github.com/Herve-M/OpenRA/blob/fix-replay/OpenRA.Game/Network/ReplayConnection.cs#L97-L112)
  - Using 4 first byte from packet to store frame n° [ReplayConnection.cs#L108](https://github.com/Herve-M/OpenRA/blob/fix-replay/OpenRA.Game/Network/ReplayConnection.cs#L108)

More info's [here](https://github.com/OpenRA/OpenRA/issues/6241#issuecomment-140373166)
Was tested on release-20150424 (depend of given replay files)

### DIFF
--- a/OpenRA.Game/Network/ReplayConnection.cs
+++ b/OpenRA.Game/Network/ReplayConnection.cs
@@ -27,6 +27,7 @@ namespace OpenRA.Network
 		Queue<Chunk> chunks = new Queue<Chunk>();
 		List<byte[]> sync = new List<byte[]>();
 		int ordersFrame;
+		Dictionary<int, int> lastClientsFrame = new Dictionary<int, int>();
 
 		public int LocalClientId { get { return 0; } }
 		public ConnectionState ConnectionState { get { return ConnectionState.Connected; } }
@@ -55,6 +56,14 @@ namespace OpenRA.Network
 					var frame = BitConverter.ToInt32(packet, 0);
 					chunk.Packets.Add(Pair.New(client, packet));
 
+                    if (frame != int.MaxValue)
+                    {
+                        if (!lastClientsFrame.ContainsKey(client))
+                            lastClientsFrame[client] = frame;
+                        else if (frame > lastClientsFrame[client])
+                            lastClientsFrame[client] = frame;
+                    }
+
 					if (packet.Length == 5 && packet[4] == 0xBF)
 						continue; // disconnect
 					else if (packet.Length >= 5 && packet[4] == 0x65)
@@ -79,6 +88,25 @@ namespace OpenRA.Network
 						chunk = new Chunk();
 
 						TickCount = Math.Max(TickCount, frame);
+					}
+				}
+
+				// 2nd parse : replace all disconnect packets without frame with real
+				// disconnect frame
+				// NOTE: to modify/remove if a reconnect feature is set
+				foreach (var tmpChunk in chunks)
+				{
+					foreach (var tmpPacketPair in tmpChunk.Packets)
+					{
+						var client = tmpPacketPair.First;
+						var packet = tmpPacketPair.Second;
+						if (packet.Length == 5 && packet[4] == 0xBF)
+						{
+							int lastClientFrame = lastClientsFrame[client];
+							byte[] lastFramePacket = BitConverter.GetBytes(lastClientFrame);
+							if (lastFramePacket.Length == 4)
+								Array.Copy(lastFramePacket, packet, lastFramePacket.Length);
+						}
 					}
 				}
 			}
--- a/AUTHORS
+++ b/AUTHORS
@@ -59,6 +59,7 @@ Also thanks to:
     * Gordon Martin (Happy0)
     * Guido Lipke (LipkeGu)
     * Gyula Zimmermann (Graion Dilach)
+    * Hervé Matysiak (Herve-M)
     * Huw Pascoe
     * Ian T. Jacobsen (Smilex)
     * Imago


Herve-M:
Squash done.

OpenRA.Game/Network/ReplayConnection.cs
```diff
@@ -55,6 +59,14 @@ public ReplayConnection(string replayFilename)
 					var frame = BitConverter.ToInt32(packet, 0);
 					chunk.Packets.Add(Pair.New(client, packet));
 
+					if (!clientsFrame.ContainsKey(client))
+					{
+						clientIds.Add(client);
+						clientsFrame[client] = new List<int>();
+					}
+
+					clientsFrame[client].Add(frame);
```
RoosterDragon: It looks like you could directly update ```lastClientsFrame``` here with the highest frame for each client, since below you're only picking out the highest frame from the whole list that you're putting into this ```clientsFrame``` dictionary.

OpenRA.Game/Network/ReplayConnection.cs
```diff
@@ -81,6 +93,45 @@ public ReplayConnection(string replayFilename)
...
+							{
+								Array.Copy(lastFramePacket, packet, lastFramePacket.Length);
+							}
+						}
+					}
+
+					tmpChunks.Enqueue(tmpChunk);
+				}
+
+				chunks = tmpChunks;
```
RoosterDragon: It doesn't look like you need the temporary ```tmpChunks``` queue. It appears like you could just update the ```packet``` in place and everything would be fine. Please let me know if I've missed something that means I'm wrong.

Herve-M: You are right, I was using the Dequeue func. before I saw that Queue impl. IEnumerable. Forget to remove it..

Herve-M:
@RoosterDragon Done changes.

OpenRA.Game/Network/ReplayConnection.cs
```diff
@@ -81,6 +90,27 @@ public ReplayConnection(string replayFilename)
...
+					foreach (var tmpPacketPair in tmpChunk.Packets)
+					{
+						var client = tmpPacketPair.First;
+						var packet = tmpPacketPair.Second;
+						if (packet.Length == 5 && packet[4] == 0xBF)
+						{
+							int lastClientFrame = lastClientsFrame[client];
+							byte[] lastFramePacket = BitConverter.GetBytes(lastClientFrame);
+							if (lastFramePacket.Length == 4)
+							{
```
pchote: Style nit: We omit braces around single line blocks

OpenRA.Game/Network/ReplayConnection.cs
```diff
@@ -81,6 +90,25 @@ public ReplayConnection(string replayFilename)
...
+				{
+					foreach (var tmpPacketPair in tmpChunk.Packets)
+					{
+						var client = tmpPacketPair.First;
+						var packet = tmpPacketPair.Second;
+						if (packet.Length == 5 && packet[4] == 0xBF)
+						{
+							int lastClientFrame = lastClientsFrame[client];
+							byte[] lastFramePacket = BitConverter.GetBytes(lastClientFrame);
+							if (lastFramePacket.Length == 4)
```
RoosterDragon: This condition is always true - did you mean to check something else?

OpenRA.Game/Network/ReplayConnection.cs
```diff
@@ -55,6 +56,14 @@ public ReplayConnection(string replayFilename)
 					var frame = BitConverter.ToInt32(packet, 0);
 					chunk.Packets.Add(Pair.New(client, packet));
 
+                    if (frame != int.MaxValue)
+                    {
+                        if (!lastClientsFrame.ContainsKey(client))
+                            lastClientsFrame[client] = frame;
+                        else if (frame > lastClientsFrame[client])
+                            lastClientsFrame[client] = frame;
```
RoosterDragon: Since both these branches do the same thing, you should probably combine the condition into one statement. i.e. ` if (!lastClientsFrame.ContainsKey(client) || frame > lastClientsFrame[client])`

RoosterDragon:
I created my own replay with a disconnection via timeout by starting 3 clients locally (one under a debugger) and running a game. I paused in the debugger to force one client to disconnect via a timeout due to "lag".

I can confirm this PR fixed the replays from the perspective of the players that continued the game (which would otherwise freeze without this PR).

The replay of the player who was disconnected was interesting. It also no longer freezes and continues! The game desyncs from reality at this point but keeps going regardless - the replay just happily continues past 100%. No crashes or anything. Some of the units were fighting and continued their attacks - but not in the way the real game played out obviously.

That probably shouldn't happen :)

Herve-M:
Interesting ! Normally for the player who disconnected the replay must end when he goes out.
I will see that.

OpenRA.Game/Network/ReplayConnection.cs
Herve-M: It was to be sure that lastFramePacket always be 4, no more no less. Otherwise replacing a arbitrary value to the packet can cause crash. 
But totally miss-viewed that BitConverter.GetBytes(int32) return every time a 4 byte array.

RoosterDragon:
I feel we should detect it somehow and prevent the replay continuing past that point.

RoosterDragon:
We could fix the replay continuing too far to another PR actually - since it doesn't crash or anything.

:+1: once my inline comments are fixed.

OpenRA.Game/Network/ReplayConnection.cs
```diff
@@ -81,6 +90,25 @@ public ReplayConnection(string replayFilename)
...
+				// NOTE: to modify/remove if a reconnect feature is set
+				foreach (var tmpChunk in chunks)
+				{
+					foreach (var tmpPacketPair in tmpChunk.Packets)
+					{
+						var client = tmpPacketPair.First;
+						var packet = tmpPacketPair.Second;
+						if (packet.Length == 5 && packet[4] == 0xBF)
+						{
+							int lastClientFrame = lastClientsFrame[client];
```
pchote: Please use `var` instead of explicit types.

pchote:
Fixing the replay continuing too far should be easy: find the last client to disconnect, and don't rewrite their disconnect packet.

RoosterDragon:
@Herve-M: We've decided to merge this a bit early and fix the minor things in a followup PR since we don't want you to miss out on hacktoberfest credit and it'll help us push our release milestone forward. Thanks for contributing!

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/104e9a726a56b24d6c1afc0f1ec6e5d639c235d6...3a79ae11560f0faa92511a96a0e554567f68ce8f)

Herve-M:
@RoosterDragon Ok, sorry for the delay, University took me to many time..

PR #9390, opened 2015-09-19
Renamed WithMuzzleFlash to WithMuzzleOverlay

reaperrr:


reaperrr:
Note: yaml diff was entirely produced with the provided upgrade rules.

OpenRA.Mods.Common/Traits/Render/WithTileFootprintOverlay.cs
```diff
@@ -15,17 +15,17 @@
...
 		[PaletteReference] public readonly string Palette = "terrain";
 		public readonly bool HasMinibib = false;
 
-		public object Create(ActorInitializer init) { return new Bib(init.Self, this); }
+		public object Create(ActorInitializer init) { return new WithTileFootprintOverlay(init.Self, this); }
 
 		public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer init, RenderSpritesInfo rs, string image, int facings, PaletteReference p)
 		{
-			if (init.Contains<HideBibPreviewInit>() && init.Get<HideBibPreviewInit, bool>())
+			if (init.Contains<HideWithTileFootprintOverlayPreviewInit>() && init.Get<HideWithTileFootprintOverlayPreviewInit, bool>())
```
pchote: Can we add a flag or another way to turn this off?  This trait has other uses (such as the d2k structures), and they won't want to be affected by this.

OpenRA.Mods.Common/Traits/Render/WithTileFootprintOverlay.cs
```diff
@@ -15,17 +15,17 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
-	public class BibInfo : ITraitInfo, Requires<BuildingInfo>, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>
+	public class WithTileFootprintOverlayInfo : ITraitInfo, Requires<BuildingInfo>, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>
```
pchote: We could make this much more general purpose by defining a Size and Offset here, and then we can remove the Minibib stuff (which is just a normal bib but with a different size and offset).

OpenRA.Mods.Common/Traits/Render/WithTileFootprintOverlay.cs
pchote: We could possibly work around this by using upgrades to hide / show the bibs as appropriate, but the interaction with actor previews is going to be fiddly.

pchote:
Thanks. Looks good to me :+1:

atlimit8:
Needs rebase.

reaperrr:
Rebased.

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/b0f43215e1694fada9d15a62766bce19f5725586...2ef12983644f24b047bd0c8be721ddcf6b6ae345)

PR #9396, opened 2015-09-20
Added a stub Red Alert 2 mod

Mailaender:
This only adds auto-generated tileset definitions and a very basic mod which allows asset exploration and new map creation. The game rules contributed by the RA2 team in external repositories and branches can be rebased and added one after another in separate pull requests to ease the review and to keep original authorship.

Changes from https://github.com/Phrohdoh/OpenRA include:
- Added gems and avoid Tiberium/BlueTiberium naming towards players.
- Shroud and fog is working.
- Blank shellmap doesn't use repeated tiles and non-isometric borders.
- The C&C: Red Alert 2 logo and wordmark is avoided.
- Uses the UI skin made by @x-a-n-a-x.
- Icon based on the retrace from @Tomsons26 instead of RA2 cover art.
- No unused fake test trait in the RA2 mod DLL.
- Tilesets use the correct RA2 categories and unit palettes.


pchote:
I think it is extremely premature to merge this, so this gets a :-1: from me.

* It is bad for the RA2 mod because it forces the high standards of our main repository guidelines onto the mod development.
* It is bad for engine development because all new features and refactorings need to be aware of the RA2 mod (this pr doesn't yet enable the travis checks to enforce this, I expect that will change).
* It is bad for contributors, because if their PRs break RA2 then they are forced to find and install paywall-blocked assets so that they can test and fix their PRs.

@Mailaender responded in IRC that these points do not matter, and that we don't need to enforce our quality rules for the RA2 mod (i.e. it is okay if a pr breaks it because somebody can fix the mod later).  If we make that exception however, then I see there being no benefit to hosting the mod here vs in its own repository.

The benefits that have been claimed for this are:

* @Mailaender doesn't like the way that @Phrohdoh managed his fork, and decided that a fresh start would be more maintainable.
* There aren't enough reviewers in the main ra2 mod repository, and maintaining it here encourages others to be involved.
* Maintaining it in a separate repository makes it hard to keep it updated with regards to the main engine changes, and having multiple developer-specific forks means that there is confusion about the canonical upstream version.

I personally do not think that the benefits of these points outweigh the negatives from above.
A better compromise would be to maintain this in a separate OpenRA/RA2 repository instead of a developer-specific fork.

Mailaender:
I certainly don't approve PRs breaking the RA2 mod. In fact I just setup travis rule checking for the RA2 mod. Indeed I think the mod would benefit from re-using our existing community and infrastructure. Red Alert 2 is not behind a paywall anymore. It is now available for free* https://www.origin.com/en-us/store/free-games/on-the-house which invalidates the paywall argument. This is meant to help @Phrohdoh & friends to contribute their efforts into the official repository, because I think it is about time now. As you stated yourself publicly the core isometric engine changes are done and only gameplay features are missing. https://github.com/OpenRA/OpenRAWeb/pull/258/files#diff-ab4e0928a7aeca60ff22a50710be5099R44

pchote:
> Red Alert 2 is not behind a paywall anymore. It is now available for free* https://www.origin.com/en-us/store/free-games/on-the-house which invalidates the paywall argument.

From the linked website:

> On the House specials can appear and disappear at any time.

You cannot guarantee that RA2 will stay free, and it is very likely that it won't.

Mailaender:
I don't have to. It is normal for an OS game clone to rely on assets that are not provided with the engine.

TheRaffy:
Maybe asking EA about if there's a deadline or duration of the free ra2 offer could help.

Mailaender:
It will be permanently tied to your Origin account. Think of it as a limited time 100% discount on the game. You can grab it now and download the assets. Safe them for later. A storage place like `%USERPROFILE%\My Documents\OpenRA\Content\ra2` could prove handy.

GraionDilach:
Trademark issues exist with D2k as well unless I'm mistaken - I am unaware of an official D2k abandonware status, and the D2k Gruntmods thingy seems as official to me as the XWIS RA2 - which works fine with the OpenRA mod as well, since it only has the campaign videos and maybe the OST missing.

I have to acknowledge that I agree with the benefit points at @pchote's post. I for example easily lose the track of the @Phrohdoh repository, because github does not have subscriptions on commits alone - nevertheless I disagree with @Phrohdoh's "purist" POV - the Kirov change is a major change affecting gameplay for example.

My counterargument right now would be the scope-creeping - many people have already expressed interest at implementing RA2 features which might hurt the pacing of the TS/FS feature implementation and even D2k requires polish to be acceptable right now feature wise - Starports comes into my mind.

Overall, I'm +0.5 on this, because tracking the RA2 mod development via the PRs would be really helpful for people IMO.

reaperrr:
My main concerns are the 3 points @pchote made. I'd +1 his OpenRA/RA2 idea (and I don't mind helping with keeping it up-to-date with bleed), but for moving this to bleed itself, I think it's simply too early and would probably have a negative impact on development of the other mods and the engine*.

Not in terms of features since people would certainly contribute quite an influx of RA2 logic, but in terms of regressions, upgrade rules and review bandwidth (RA2 has a lot more stuff than TS both in terms of logic and actor rules).

Mailaender:
My aim was not to have everyone write very specialized wacky RA2 actor logic, but present something playable ASAP to hop on the RA2 is free* band waggon. The TS mod and the "release when it's (completely) done" dogma essentially resulted in development hell with no end in sight. We could have a simplified RA2 mod which can serve the casual medium tank battles on flat terrain just like http://cncnet.org done in a few weeks so we can release a preview version just before EA switches their freebie time limited promotional offer to "Candy Crush: Zombie Edition" again.

I don't have time for creating and maintaining OpenRA/RA2 nor do I think a mod outside this volatile development repository can survive as well as be kept stable and compatible. So if that is what everybody prefers then someone else has to do the step.

pchote:
> My aim was not to have everyone write very specialized wacky RA2 actor logic, but present something playable ASAP to hop on the RA2 is free* band waggon.

This can be done just as easily and flexibly from a separate repo, unless you are intending for us to ship RA2 as an official mod in the main distribution (which would be an even more controversial proposal).

penev92:
>There aren't enough reviewers in the main ra2 mod repository, and maintaining it here encourages others to be involved.

No, I think you made a typo here. Let me correct that - "maintaining it here forces others to be involved".

I won't go into arguments and counter-arguments. Those have been discussed *over* and *over* and *over* again.
I'll just leave my :-1: here. 

>[15:39:37] < penev|m> I think it's great to help get RA2 in shape
[15:39:52]	< penev|m> I've done some helping myself
[15:40:22]	< penev|m> But I stand firmly behind the "don't merge ra2" notion 

The main argument is "it would be easier to maintain the mod if it is in the official repo". Sure. But have you asked yourself "why would it be easier?" ? I'll tell you - because the mod's maintainers won't have to rebase and update it themselves. They will be delegating that work to contributors. Yes, it will be up to whoever files a PR to also update RA2 (as part of the official mods). I do **not** see that as a fair tradeoff.

Also I said earlier today on IRC that I see **no** benefit to either OpenRA or RA2 to be gained by such a merge. I still haven't been shown any.  (Btw I think the idea to get a preview mod for the free offer hype is an interesting one, but it's still not worth it. Also that can be done without merging the mod.)

(Oh, look at that, I got into the "arguments and counter-arguments" thing a bit.)


P.S.: Also please don't make this about "persuading the lead programmer", because while this doesn't run well as a democracy, there is still a sizable chunk of the (admittedly quite small) dev team that is against this.

reaperrr:
> My aim was not to have everyone write very specialized wacky RA2 actor logic, but present something playable ASAP to hop on the RA2 is free* band waggon.

In addition to @pchote's and @penev92's points, I'm not sure we should rush this just because it might get a little more attention now than it would in a few months.
The thing is, once we hop onto that bandwagon there's no jumping down, at least not without people constantly barraging us with "full RA2 when?", "screw TS, I want RA2!", "how is RA2 progress?", "I'm disappointed, your RA2 is still missing feature X, unit Y and building Z and performance is soooo poor :(", "original RA2/YR over CnCNet is much better, lol", "why no Yuri?" and so on.

And if we add it to the main repo, I fully expect our issue/PR lists to explode well beyond 1000/100 respectively in no time, worst case we'll have to reject dozens of them to keep the whole project somewhat maintainable which might scare/p*ss off contributors.

Mailaender:
I thought the discussion here would be more on topic focusing on the actual pull request, but no one even bothered. This is a meritocracy so I will subordinate. It was a strength of this project to keep everything under the hood and implement as general and re-usable as possible. If this approach will be taken and we separate all mods from the engine coordinating releases will become a real challenge and slow everything down even further.

reaperrr:
I have a fixed version of @Phrohdoh's master branch including some fixes from this PR, can be found [here](https://github.com/reaperrr/OpenRA/tree/ra2%23master).

Mailaender:
I added the missing bits from my https://github.com/Mailaender/OpenRA/tree/ra2 in https://github.com/Phrohdoh/OpenRA/pull/18.

Mailaender:
> A better compromise would be to maintain this in a separate OpenRA/RA2 repository instead of a developer-specific fork.

See https://github.com/OpenRA/ra2

Mailaender:
> * It is bad for contributors, because if their PRs break RA2 then they are forced to find and install paywall-blocked assets so that they can test and fix their PRs.

Technically the assets from http://xwis.net/forums/index.php/topic/163831-how-to-play/ are freely available so no paywall, but still you need to find and install them.

Issue #8442, opened 2015-06-08
Crash when required arguments are not given / passed.


LipkeGu:
it is able to crash the Utitlity by simpliy omitting required arguments :/

```
Received args: --map-import
System.IndexOutOfRangeException: Der Index war außerhalb des Arraybereichs.
   bei OpenRA.Mods.Common.UtilityCommands.ImportLegacyMapCommand.Run(ModData modData, String[] args)
   bei OpenRA.Utility.Program.Main(String[] args) in d:\Projekte\OpenRA\OpenRA.Utility\Program.cs:Zeile 54.

```

Phrohdoh:
This is a completely normal part of applications that require arguments.

Unless you are suggest we just give a warning and catch any exceptions (or implement nargs).

LipkeGu:
we should redirect the user to the Usage function in this case

penev92:
This is a completely normal part of applications that require arguments.

And also it is unreasonable to add argument checks for every utility
command (they will each need a separate check).
On Jun 8, 2015 3:10 PM, "Taryn Hill" <notifications@github.com> wrote:

This is a completely normal part of applications that require arguments.

Unless you are suggest we just give a warning and catch any exceptions (or
implement nargs).

—
Reply to this email directly or view it on GitHub
<https://github.com/OpenRA/OpenRA/issues/8442#issuecomment-109970222>.

Phrohdoh:
@penev92 Well, as I was saying, we could add `int nargs = -1` (default to 'any'), and check the number of arguments based on the `IUtilityCommand`'s `ArgumentCount` field.

pchote:
> And also it is unreasonable to add argument checks for every utility command (they will each need a separate check).

I disagree.  It is completely reasonable to expect each command to check that it has been given proper input.

obrakmann:
Fixed by #9397

PR #9397, opened 2015-09-20
Add basic arguments validation

Herve-M:
This PR is the first step of validation of argument(s) passed through OpenRA.Utility. Target issue #8442
- Add a method [IUtilityCommand](https://github.com/Herve-M/OpenRA/blob/feat-utility-polish/OpenRA.Game/IUtilityCommand.cs#L20)
- Test Argument before calling command [OpenRA.Utility/Program.cs](https://github.com/Herve-M/OpenRA/blob/feat-utility-polish/OpenRA.Utility/Program.cs#L59-L72)
- Impl. basic validation (on args.Length) on all [Utility.Commands](https://github.com/Herve-M/OpenRA/commit/ff82276c2103ab2c06e5926f37170346a7ee8795?diff=unified)

If a command is called without validating, the description is showed.

```
λ OpenRA.Utility.exe ra --shp
Invalid arguments for '--shp'
  --shp PNGFILE [PNGFILE ...]
  Combine a list of PNG images into a SHP
```

If a inexisting command is used, a System.ArgumentException is thrown without crashing / showing it. 

```
λ OpenRA.Utility.exe ra --sh
No such command '--sh'
```

**Special attention for validation** : see if UnityCommands:ValidateArguments are just.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/CheckCodeStyle.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/CheckCodeStyle.cs
@@ -9,12 +9,7 @@
 #endregion
 
 using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Linq;
-using System.Reflection;
-using OpenRA.FileSystem;
-using OpenRA.Traits;
 using StyleCop;
 
 namespace OpenRA.Mods.Common.UtilityCommands
@@ -23,6 +18,11 @@ namespace OpenRA.Mods.Common.UtilityCommands
 	{
 		public string Name { get { return "--check-code-style"; } }
 		int violationCount;
+
+		public bool ValidateArguments(string[] args)
+		{
+			return args.Length >= 2;
+		}
 
 		[Desc("DIRECTORY", "Check the *.cs source code files in a directory for code style violations.")]
 		public void Run(ModData modData, string[] args)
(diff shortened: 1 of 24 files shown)


OpenRA.Mods.Common/UtilityCommands/CheckCodeStyle.cs
```diff
@@ -24,6 +24,16 @@ class CheckCodeStyle : IUtilityCommand
...
 		int violationCount;
 
+		public bool ValidateArguments(string[] args)
+		{
+			if (args.Length >= 2)
+			{
+				return true;
+			}
+
+			return false;
```
Phrohdoh: `return args.Length >= 2;`

RoosterDragon:
Looks good generally.

- Can you squash your fixes into your original commit please.
- If invalid arguments are given to a command, can you print out a message saying something like `Invalid arguments for '--shp'` and then print the usage to make it clear there was a problem.
- If the command doesn't exist, rather than crashing or showing an exception, I think just outputting an error message (e.g. `No such command '--sh'`) would be sufficient.

Herve-M:
@RoosterDragon All done.

penev92:
Looks good to me :+1:

obrakmann:
:+1: Works and looks fine.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/dae2dfc646f525cbd4ba66e4f79a418c7fa3e46f)

PR #9398, opened 2015-09-20
Renamed Bib to WithTileFootprintOverlay

reaperrr:
Split from #9390.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithTileFootprintOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithTileFootprintOverlay.cs
@@ -15,17 +15,17 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
-	public class BibInfo : ITraitInfo, Requires<BuildingInfo>, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>
+	public class WithTileFootprintOverlayInfo : ITraitInfo, Requires<BuildingInfo>, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>
 	{
 		[SequenceReference] public readonly string Sequence = "bib";
 		[PaletteReference] public readonly string Palette = "terrain";
 		public readonly bool HasMinibib = false;
 
-		public object Create(ActorInitializer init) { return new Bib(init.Self, this); }
+		public object Create(ActorInitializer init) { return new WithTileFootprintOverlay(init.Self, this); }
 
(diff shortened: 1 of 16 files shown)


OpenRA.Mods.Common/Traits/Render/WithTileFootprintOverlay.cs
```diff
@@ -15,17 +15,17 @@
...
 		[PaletteReference] public readonly string Palette = "terrain";
 		public readonly bool HasMinibib = false;
 
-		public object Create(ActorInitializer init) { return new Bib(init.Self, this); }
+		public object Create(ActorInitializer init) { return new WithTileFootprintOverlay(init.Self, this); }
 
 		public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer init, RenderSpritesInfo rs, string image, int facings, PaletteReference p)
 		{
-			if (init.Contains<HideBibPreviewInit>() && init.Get<HideBibPreviewInit, bool>())
+			if (init.Contains<HideWithTileFootprintOverlayPreviewInit>() && init.Get<HideWithTileFootprintOverlayPreviewInit, bool>())
```
reaperrr: > Can we add a flag or another way to turn this off? This trait has other uses (such as the dune 2 structures and bridges), and they won't want to be affected by this.

OpenRA.Mods.Common/Traits/Render/WithTileFootprintOverlay.cs
```diff
@@ -15,17 +15,17 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
-	public class BibInfo : ITraitInfo, Requires<BuildingInfo>, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>
+	public class WithTileFootprintOverlayInfo : ITraitInfo, Requires<BuildingInfo>, IRenderActorPreviewSpritesInfo, Requires<RenderSpritesInfo>
```
reaperrr: > We could make this much more general purpose by defining a Size and Offset here, and then we can remove the Minibib stuff (which is just a normal bib but with a different size and offset).

atlimit8: :+1:

OpenRA.Mods.Common/Traits/Render/WithTileFootprintOverlay.cs
atlimit8: Sounds reasonable.

atlimit8:
Needs to be rebased.

abcdefg30:

PR #9400, opened 2015-09-21
Fixed massive AI base builder performance issue

### LINKED ISSUE #4717: Performance issues with AI building placement [$75]

reaperrr:
HackyAI causes occasional BaseBuilder-related lag spikes of anywhere between 200 and 1200(!) milliseconds.

After throwing around with PerfTimers, I narrowed it down to the refinery placement.
In worst-case scenarios, the AI would iterate through all cells with resources inside MaxBaseRadius and perform a full findPos check for each of them.

Now it simply takes the first randomly chosen cell with resources inside MaxBaseRadius and places the refinery in that direction, if possible, and otherwise falls back to a normal placement check.

Closes #4717.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -500,11 +500,11 @@ namespace OpenRA.Mods.Common.AI
 					// Try and place the refinery near a resource field
 					var nearbyResources = Map.FindTilesInCircle(baseCenter, Info.MaxBaseRadius)
 						.Where(a => resourceTypeIndices.Get(Map.GetTerrainIndex(a)))
-						.Shuffle(Random);
+						.Shuffle(Random).ToList();
 
-					foreach (var c in nearbyResources)
+					if (nearbyResources.Count > 0)
 					{
-						var found = findPos(c, baseCenter, 0, Info.MaxBaseRadius);
+						var found = findPos(nearbyResources.First(), baseCenter, 0, Info.MaxBaseRadius);
 						if (found != null)
 							return found;
 					}
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -101,6 +101,9 @@ namespace OpenRA.Mods.Common.AI
 		[Desc("Radius in cells around a factory scanned for rally points by the AI.")]
 		public readonly int RallyPointScanRadius = 8;
 
+		[Desc("Minimum distance in cells from center of the base when checking for building placement.")]
+		public readonly int MinBaseRadius = 2;
+
 		[Desc("Radius in cells around the center of the base to expand.")]
 		public readonly int MaxBaseRadius = 20;
 
@@ -501,22 +504,22 @@ namespace OpenRA.Mods.Common.AI
 				case BuildingType.Refinery:
 
 					// Try and place the refinery near a resource field
-					var nearbyResources = Map.FindTilesInCircle(baseCenter, Info.MaxBaseRadius)
+					var nearbyResources = Map.FindTilesInAnnulus(baseCenter, Info.MinBaseRadius, Info.MaxBaseRadius)
 						.Where(a => resourceTypeIndices.Get(Map.GetTerrainIndex(a)))
 						.Shuffle(Random).Take(Info.MaxResourceCellsToCheck);
 
 					foreach (var r in nearbyResources)
 					{
-						var found = findPos(r, baseCenter, 0, Info.MaxBaseRadius);
+						var found = findPos(r, baseCenter, Info.MinBaseRadius, Info.MaxBaseRadius);
 						if (found != null)
 							return found;
 					}
 
 					// Try and find a free spot somewhere else in the base
-					return findPos(baseCenter, baseCenter, 0, Info.MaxBaseRadius);
+					return findPos(baseCenter, baseCenter, Info.MinBaseRadius, Info.MaxBaseRadius);
 
 				case BuildingType.Building:
-					return findPos(baseCenter, baseCenter, 0, distanceToBaseIsImportant ? Info.MaxBaseRadius : Map.MaxTilesInCircleRange);
+					return findPos(baseCenter, baseCenter, Info.MinBaseRadius, distanceToBaseIsImportant ? Info.MaxBaseRadius : Map.MaxTilesInCircleRange);
 			}
 
 			// Can't find a build location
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -497,19 +497,10 @@ bool HasAdequateAirUnitReloadBuildings(ActorInfo actorInfo)
 
 				case BuildingType.Refinery:
 
-					// Try and place the refinery near a resource field
-					var nearbyResources = Map.FindTilesInCircle(baseCenter, Info.MaxBaseRadius)
-						.Where(a => resourceTypeIndices.Get(Map.GetTerrainIndex(a)))
-						.Shuffle(Random);
+					// Old code to place closer to resources was removed for performance reasons.
+					// TODO: Reimplement it in a way to doesn't cause performance issues.
```
ScottNZ: to -> that?

reaperrr:
Updated.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -502,9 +502,9 @@ bool HasAdequateAirUnitReloadBuildings(ActorInfo actorInfo)
 						.Where(a => resourceTypeIndices.Get(Map.GetTerrainIndex(a)))
 						.Shuffle(Random);
 
-					foreach (var c in nearbyResources)
+					if (nearbyResources.Any())
 					{
-						var found = findPos(c, baseCenter, 0, Info.MaxBaseRadius);
+						var found = findPos(nearbyResources.First(), baseCenter, 0, Info.MaxBaseRadius);
```
RoosterDragon: This will enumerate the nearby resources twice. If you're just going to pick any random nearby resource try:
```c#
var randomNearbyResource = Map.FindTilesInCircle(baseCenter, Info.MaxBaseRadius)
    .Where(a => resourceTypeIndices.Get(Map.GetTerrainIndex(a)))
    .RandomOrDefault(Random);
```

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -499,12 +499,11 @@ bool HasAdequateAirUnitReloadBuildings(ActorInfo actorInfo)
...
 					var nearbyResources = Map.FindTilesInCircle(baseCenter, Info.MaxBaseRadius)
-						.Where(a => resourceTypeIndices.Get(Map.GetTerrainIndex(a)))
-						.Shuffle(Random);
+						.Where(a => resourceTypeIndices.Get(Map.GetTerrainIndex(a)));
 
-					foreach (var c in nearbyResources)
+					if (nearbyResources.Any())
 					{
-						var found = findPos(c, baseCenter, 0, Info.MaxBaseRadius);
+						var found = findPos(nearbyResources.RandomOrDefault(Random), baseCenter, 0, Info.MaxBaseRadius);
```
RoosterDragon: We're back to the double enumeration problem. You should consider a ```ToList``` or my earlier suggestion in IRC to cast to a ```CPos?``` so RandomOrDefault works better for you.

reaperrr:
@RoosterDragon Is this what you meant? I'm fishing a little in the dark about how to implement your suggestions.

RoosterDragon:
Yeah that looks fine.

I feel failing after one random choice may be a bit too restrictive. Maybe we could allow it to try `n` random choices before trying the fallback? (could be done by doing `.Shuffle(Random).Take(MaxResourcesCellsToCheck).ToList()` and keeping the `foreach`) That way we can strike a balance between the lag spike and smartness of placement.

That's entirely a feels driven opinion so I might try a test or two to see if I can come up with a sensible number on that.

reaperrr:
Done.

Additionally added `MinBaseRadius` so we can skip cells likely to be occupied by the construction yard anyway.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -498,22 +504,22 @@ bool HasAdequateAirUnitReloadBuildings(ActorInfo actorInfo)
 				case BuildingType.Refinery:
 
 					// Try and place the refinery near a resource field
-					var nearbyResources = Map.FindTilesInCircle(baseCenter, Info.MaxBaseRadius)
+					var nearbyResources = Map.FindTilesInAnnulus(baseCenter, Info.MinBaseRadius, Info.MaxBaseRadius)
 						.Where(a => resourceTypeIndices.Get(Map.GetTerrainIndex(a)))
-						.Shuffle(Random);
+						.Shuffle(Random).ToList().Take(Info.MaxResourceCellsToCheck);
```
RoosterDragon: You can drop the `ToList` now, so we're only going to enumerate it once. Sorry for the back and forth.

RoosterDragon:
On my system, a single search sometimes took up to 3ms. With the loop overhead the 3 searches can take about 10ms. (Though often will be faster)

So if we throw another 50% on that to be conservative for slower systems, that's about 15ms or basically a single frame at 60 fps.

Therefore I think the default of 3 here is pretty good.

:+1: once you address the unnecessary `ToList`.

reaperrr:
> once you address the unnecessary `ToList`.

Done.

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d9f32d0a250f2ec8e43d9c80b8ca47f3600044d5...69a79d8244e14b61caddfb97eff6cc4079757e7f)

PR #9406, opened 2015-09-22
Fixed D2k armament offsets and some cleanup

reaperrr:


mods/d2k/rules/structures.yaml
```diff
@@ -593,7 +593,7 @@ large_gun_turret:
 		QuantizedFacings: 32
 	Armament:
 		Weapon: TowerMissile
-		LocalOffset: 256,384,768, 256,-384,768
+		LocalOffset: 256,0,768
```
penev92: What is the reason for this? This looks like it is removing some polish.
Same for the missile tank.

GraionDilach: This one - haven't checked the missile tank yet - doesn't have burst on the weapon, which is why the swapping launchers back and forth is changed to a "fire between the two" look. Another polish could be halving missile damage and giving it Burst: 2 to maintain the image of using both launchers.

penev92: That would be great, but I guess it's out of scope here.

reaperrr: Applies to missile tank as well.

We could also leave damage as it is, but still use Burst: 2 and make the BurstDelay identical to ReloadDelay, then the rate of fire would remain the same but it would alternate between left and right.

penev92: Alternating would be great, as long as we don't change the effective DPS of the unit (and don't make it fire 2 missiles at once).

reaperrr:
Updated:
- Reverted the offset changes and made the missile weapons alternate between left and right.
- Fixed quad rockets (they were anti-ground only in the original).
- Corrected missile tank report and weapon explosion anim, also used a less noticable explosion sound (the original had none at all, but that felt wrong, too) and removed armament FireDelay.
- Corrected missile tower weapon explosion and sound to match missile tank.
- Corrected siege tank weapon explosion sound.

penev92:
Looks good to me :+1: 
Thanks for all the cleanups.

pchote:
Looks fine to me, too. :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7a79164cbdc008a98730522b6219bad36d412ee4

PR #9408, opened 2015-09-22
Added original D2k debris

reaperrr:
...and removed our Shrapnel weapon.

Additionally made first step towards splitting weapons.yaml.

### DIFF
--- a/mods/d2k/mod.yaml
+++ b/mods/d2k/mod.yaml
@@ -106,6 +106,7 @@ ChromeLayout:
 
 Weapons:
 	./mods/d2k/weapons.yaml
+	./mods/d2k/weapons/debris.yaml
 
 Voices:
 	./mods/d2k/audio/voices.yaml
--- a/mods/d2k/rules/defaults.yaml
+++ b/mods/d2k/rules/defaults.yaml
@@ -281,7 +281,7 @@
 	Demolishable:
 	DamagedWithoutFoundation:
 	ThrowsShrapnel:
-		Weapons: shrapnel
+		Weapons: Debris, Debris2, Debris3, Debris4
 		Pieces: 3, 7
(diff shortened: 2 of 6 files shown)


penev92:
This seems like quite the overkill ingame. You should perhaps lower the amount of pieces.
Also it feels like they fly somewhat slower than they should. Not sure what to do about that.

reaperrr:
Original d2k *was* overkill in terms of spawning debris, though. But yeah, I guess it should be tweaked more to match the individual structures' size.

penev92:
Well you may be right - https://www.youtube.com/watch?v=6aOen-y6K1g&feature=youtu.be&t=600
But they definitely go away faster, and don't appear to fly that far.

LavenderMoon:
Debris are implemented as `Bullet`s and could be made to do the same in OpenRA without any additional C# code, right? (Bare with me, I'm still getting used to this codebase.)

reaperrr:
Updated, reduced the range and slightly increased max speed.

@LavenderMoon Yes, that can be done entirely in yaml.

GraionDilach:
Is there a reason why the first debris doesn't have trail?

penev92:
They still feel a bit slow, but I guess it's nothing to cry over, so :+1:

Mailaender:
On normal game speed this feels like playing on a low/no gravity and the debris is also very deadly to infantry on direct hit, but I assume this is the original balancing which will be improved later on. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/7f3c87ae1c762704615c86e843cf4672db4dbae4...01097bc17f82b5d5a296374bf935281a071d317b)

PR #9414, opened 2015-09-23
Support loading multiple turret facings

atlimit8:
Adds `TurretFacingsInit` to support loading/saving the turret facing of each turret.
Moved `TurretFacingInit` since it should have been in `OpenRA.Mods.Common/Traits/Turreted.cs` all along.
Added `IActorInitializer` so that `Turreted.TurretFacingsInit` can use `ActorPreviewInitializer`s instead of `ActorInitializer`s when called from `RenderPreview*`.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Turreted.cs
+++ b/OpenRA.Mods.Common/Traits/Turreted.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System;
+using System.Collections.Generic;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
@@ -47,8 +48,15 @@ namespace OpenRA.Mods.Common.Traits
 		public WVec Offset { get { return info.Offset + localOffset; } }
 		public string Name { get { return info.Turret; } }
 
-		public static int GetInitialTurretFacing(ActorInitializer init, int def)
+		public static int GetInitialTurretFacing(IActorInitializer init, int def, string turret = null)
 		{
+			if (turret != null && init.Contains<TurretFacingsInit>())
+			{
+				int facing;
+				if (init.Get<TurretFacingsInit, Dictionary<string, int>>().TryGetValue(turret, out facing))
+					return facing;
+			}
+
 			if (init.Contains<TurretFacingInit>())
 				return init.Get<TurretFacingInit, int>();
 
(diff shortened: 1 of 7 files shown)


OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -47,8 +48,15 @@ public class Turreted : ITick, ISync, INotifyCreated
...
 		public string Name { get { return info.Turret; } }
 
-		public static int GetInitialTurretFacing(ActorInitializer init, int def)
+		public static int GetInitialTurretFacing(IActorInitializer init, int def, string turret = null)
 		{
+			if (turret != null && init.Contains<TurretFacingsInit>())
+			{
+				int facing;
+				if (init.Get<TurretFacingsInit, Dictionary<string, int>>().TryGetValue(turret, out facing))
+					return init.Get<TurretFacingInit, int>();
```
pchote: I think this needs to be `return facing;`?

atlimit8: Yes, copy and paste change failure.

atlimit8:
Fixed.

OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -128,4 +136,21 @@ public WRot LocalOrientation(Actor self)
...
+
+	public class TurretFacingInit : IActorInit<int>
+	{
+		[FieldFromYamlKey] readonly int value = 128;
+		public TurretFacingInit() { }
+		public TurretFacingInit(int init) { value = init; }
+		public int Value(World world) { return value; }
+	}
+
+	public class TurretFacingsInit : IActorInit<Dictionary<string, int>>
```
penev92: Is there a reason you're keeping the old (and now obsolete) singlular init?

pchote: Most units only have one turret, so it would be a bit ugly to require them to use the more verbose TurretFacingsInit.

abcdefg30:
Looks good to me and works with the RA cruiser. :+1: / :white_check_mark:

obrakmann:
Looks fine. :+1: 

For documentation purposes, you'd use it like this:
```yaml`
	Cruiser: ca
		Location: 52,129
		Owner: Allies
		TurretFacings:
			primary: 64
			secondary: 224
```

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/de6dd63b6782c97e4cf5ebc85e75b623af7cec8c)

Issue #8549, opened 2015-06-23
Bogus use of Map.TopLeft and .BottomRight in mission scripts 


pchote:
Queries like `Map.ActorsInBox(Map.TopLeft, Map.BottomRight, function(self) ... end)` are used through most of our map scripts as a method of finding the actors in a map that match an arbitrary filter.

This is bogus for two reasons:

* It abuses an API designed for one thing (finding the actors within a specific part of the map) to do something something completely different (finding all the actors on the map matching a condition).  This is a design smell.
* It doesn't work in TS or RA2 (or any other mods that decide to use terrain height).  The top and bottom edges of the map are not lines of constant WPos.Y, and so you will miss actors near the top of the map if they are at a lower elevation than the top-left cell, and miss actors at the bottom edge of the map if they are at a higher elevation than the bottom-right cell.
* (bonus point) The implementation of `.TopLeft` and `.BottomRight` is basically completely bogus, and returns the wrong points for isometric maps.  We can fix this at least, but its not clear whether these positions are actually useful to scripters after we fix the above two points.

PR #9417, opened 2015-09-23
reduce use of Map.Topleft and Map.Bottomright

clemty:
as requested in #8549

### DIFF
--- a/mods/cnc/maps/gdi01/gdi01.lua
+++ b/mods/cnc/maps/gdi01/gdi01.lua
@@ -2,6 +2,7 @@ MCVReinforcements = { "mcv" }
 InfantryReinforcements = { "e1", "e1", "e1" }
 VehicleReinforcements = { "jeep" }
 NodPatrol = { "e1", "e1" }
+GDIBaseBuildings = { "pyle", "fact", "nuke" }
 
 SendNodPatrol = function()
 	Reinforcements.Reinforce(enemy, NodPatrol, { nod0.Location, nod1.Location }, 15, function(soldier)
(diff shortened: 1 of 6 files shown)


mods/cnc/maps/gdi01/gdi01.lua
```diff
@@ -49,11 +49,10 @@ Reinforce = function(units)
...
 
 CheckForBase = function()
-	baseBuildings = Map.ActorsInBox(Map.TopLeft, Map.BottomRight, function(actor)
-		return actor.Type == "fact" or actor.Type == "pyle" or actor.Type == "nuke"
-	end)
-
-	return #baseBuildings >= 3
+	local barracks = player.GetActorsByType("pyle")
+	if #barracks > 0 then
+		return true
```
abcdefg30: `return #barracks > 0`

mods/cnc/maps/nod05/nod05.lua
```diff
@@ -270,10 +270,10 @@ IdleHunt = function(unit)
 end
 
 CheckForSams = function(player)
-	local baseBuildings = Map.ActorsInBox(Map.TopLeft, Map.BottomRight, function(actor)
-		return actor.Owner == Nod and actor.Type == 'sam' end)
-
-	return #baseBuildings >= 3
+	local sams = Nod.GetActorsByType("sam")
+	if #sams >= 3 then
+		return true
```
abcdefg30: `return #sams >= 3`

mods/cnc/maps/gdi01/gdi01.lua
GraionDilach: Can't it just be `return player.GetActorsByType("pyle") > 0`?

abcdefg30: With a `#` before `player` (`return #player.GetActorsByType("pyle") > 0`), but yes, that is actually better.

clemty: I find the current code easier to read - I can change it, but is it worth it in terms of runtime efficiency? (that function is only called once per second, and the GetActorsByType call is probably the bottleneck there)

abcdefg30:
I'd also like it more if we keep `NodBaseBuildings = { "hand", "fact", "nuke" } ` and use this CheckForBase method:
```
CheckForBase = function(player)
	local buildings = 0

	Utils.Do(NodBaseBuildings, function(name)
		if #player.GetActorsByType(name) > 0 then
			buildings = buildings + 1
		end
	end)

	return buildings == #NodBaseBuildings
end

clemty:
updated

@abcdefg30 : thanks for the code! I updated gdi01 as well to use that

clemty:
updated with whitespace fixes (thanks @GraionDilach )
removed one instance of map.topleft/map.bottomright from allies03a

abcdefg30:
Looks good to me and works. :+1: / :white_check_mark: 
Also thanks for the cleanup in allies03a.

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/554c9dec5ba65f162a21d4b1734b9ef2b15886e7...cc998c71ca4e1357e5fafe3ed71874181f8e9f8f)

PR #9418, opened 2015-09-23
Apply original D2k actor speed

reaperrr:


pchote:
These seem reasonable in game, but I haven't verified that they match the original.

mods/d2k/rules/aircraft.yaml
```diff
@@ -52,16 +52,16 @@ frigate:
...
-		ROT: 5
-		Speed: 350
+		ROT: 1
+		Speed: 126
 		RepairBuildings: repair_pad
 		RearmBuildings:
 		Repulsable: False
 	Health:
-		HP: 500
+		HP: 9990
```
penev92: The frigate should be indestructable, so best remove `Health:` and `Armor:` completely, and `Targetable:` if it has it.

penev92:
Other than that :+1:

Mailaender:
Feels a bit (probably authentic) slow, but that is not an issue anymore since https://github.com/OpenRA/OpenRA/pull/9255. :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8623c79550221167fbd2bee6ae828d541ef5c4ac...61a807fbfed338263bd1af8b88dbe694889cf34f)

PR #9419, opened 2015-09-23
Chronoshifted are force returned to origin with damage by collisions

### LINKED ISSUE #9349: Husks don't chronoshift back to original location

atlimit8:
More or less what was wanted on IRC.
Maybe the inflicted damage should be different for the both have `Health` case; I averaged current and max HP of each as rough structural strength based damage values.

~~Fixes #9349~~

### DIFF
--- a/OpenRA.Mods.RA/Activities/Teleport.cs
+++ b/OpenRA.Mods.RA/Activities/Teleport.cs
@@ -28,9 +28,10 @@ namespace OpenRA.Mods.RA.Activities
 		CPos destination;
 		bool killCargo;
 		bool screenFlash;
+		bool force;
 		string sound;
 
-		public Teleport(Actor teleporter, CPos destination, int? maximumDistance, bool killCargo, bool screenFlash, string sound)
+		public Teleport(Actor teleporter, CPos destination, int? maximumDistance, bool killCargo, bool screenFlash, string sound, bool force = false)
 		{
 			if (maximumDistance > MaxCellSearchRange)
 				throw new InvalidOperationException("Teleport cannot be used with a maximum teleport distance greater than {0}.".F(MaxCellSearchRange));
@@ -40,6 +41,7 @@ namespace OpenRA.Mods.RA.Activities
 			this.maximumDistance = maximumDistance;
 			this.killCargo = killCargo;
 			this.screenFlash = screenFlash;
+			this.force = force;
 			this.sound = sound;
 		}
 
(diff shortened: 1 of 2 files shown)


penev92:
TL;DR what does this do? Fixes the "actor X didn't chronoshift back" issue?
If it fixes it reliably then you'd want to reference the issue in the PR description.

OpenRA.Mods.RA/Activities/Teleport.cs
```diff
@@ -53,16 +55,59 @@ public override Activity Tick(Actor self)
...
 
 			Game.Sound.Play(sound, self.CenterPosition);
 			Game.Sound.Play(sound, self.World.Map.CenterOfCell(destination));
 
 			self.Trait<IPositionable>().SetPosition(self, destination);
+			if (force)
+			{
+				if (self.Info.HasTraitInfo<HealthInfo>())
+				{
+					var health = self.Trait<Health>();
```
abcdefg30: Is there a reason why you didn't use `TraitOrDefault` and did a `null`-check above (instead of `HasTraitInfo`)?

abcdefg30:
It works as advertised. :+1: / :white_check_mark: 
Causes stuff like a tank and a husk being on the same cell, but I guess this is a good tradeoff.

(PS: I don't really like the code duplication, but we can't really do something about that, can we?)

deniz1a:
This just destroys tanks when they return from chrono, even when there is nothing on the return cell.

OpenRA.Mods.RA/Activities/Teleport.cs
```diff
@@ -53,16 +55,59 @@ public override Activity Tick(Actor self)
...
+					var givenDamage = (health.HP + health.MaxHP) / 2;
+					foreach (var a in self.World.ActorMap.GetUnitsAt(destination))
```
pchote: The filtering here isn't sufficient. There are certain actor types (e.g. bridges, building bibs) that they should be able to return on without damage).  Use IPositionable.CanEnterCell (might be on a different interface) to check.

deniz1a:
Here's what happens when I chrono a tank:
https://drive.google.com/file/d/0B_dMvsHPqPZkdzIzZnA0MUhYbkE/view

atlimit8:
Works correctly on bibs and bridges - only actual collisions cause damage now.
I accidentally forgot to filter out itself when checking for collisions.

atlimit8:
Merged `foreach` loops.

OpenRA.Mods.RA/Activities/Teleport.cs
```diff
@@ -53,16 +55,71 @@ public override Activity Tick(Actor self)
...
+				destination = bestCell.Value;
+			}
 
 			Game.Sound.Play(sound, self.CenterPosition);
 			Game.Sound.Play(sound, self.World.Map.CenterOfCell(destination));
 
-			self.Trait<IPositionable>().SetPosition(self, destination);
+			var pos = self.Trait<IPositionable>();
+			if (pos.CanEnterCell(destination))
+				force = false;
```
Phrohdoh: Newline here please.

OpenRA.Mods.RA/Activities/Teleport.cs
```diff
@@ -53,16 +55,71 @@ public override Activity Tick(Actor self)
...
+			pos.SetPosition(self, destination);
+			if (force)
+			{
+				var subcell = pos.OccupiedCells().First(p => p.First == destination).Second;
+				var health = self.TraitOrDefault<Health>();
+				var givenDamage = health != null ? (health.HP + health.MaxHP) / 2 : 0;
+
+				foreach (var a in self.World.ActorMap.GetActorsAt(destination))
```
Phrohdoh: Name this `actor` please.
There is no need to use a single-character variable name.

OpenRA.Mods.RA/Activities/Teleport.cs
```diff
@@ -53,16 +55,71 @@ public override Activity Tick(Actor self)
...
+
+					var otherHealth = a.TraitOrDefault<Health>();
+					if (otherHealth == null)
+						continue;
+
+					if (health != null)
+					{
+						var maxHealthRadius = health.Info.Radius;
+						if (otherHealth.Info.Radius > maxHealthRadius)
+							maxHealthRadius = otherHealth.Info.Radius;
```
Phrohdoh: Newline.

OpenRA.Mods.RA/Activities/Teleport.cs
```diff
@@ -53,16 +55,71 @@ public override Activity Tick(Actor self)
...
+						if ((a.CenterPosition - self.CenterPosition).LengthSquared > maxHealthRadius.LengthSquared)
+							continue;
+
+						var otherGivenDamage = (otherHealth.HP + otherHealth.MaxHP) / 2;
+						otherHealth.InflictDamage(a, self, givenDamage, null, true);
+						health.InflictDamage(self, a, otherGivenDamage, null, true);
```
Phrohdoh: Shouldn't the attacker be the `other` (whether that be self or `actor`)?

atlimit8: They are colliding with each other causing each other damage.

Phrohdoh: Sorry, I read the arguments in the wrong order.

atlimit8:
Changed.

abcdefg30:
Something is wrong now, returning actors don't necessarily cause damage (on buildings):
chronoinweap

I also noticed two other things (not sure if we want to fix them [in this PR]):
- Returning tanks damage a repair pad, although they can move on it
- If a unit is has the Iron Curtain invulnerability, it still gets killed

pchote:

PR #9435, opened 2015-09-25
Create disable movement trait

ABrandau:
This trait is a clone of the `DisableUpgrade` trait, the only difference is that it only disables movement. It's needed for units that transform into static variants using deploy to upgrade, as those depend on the "disableupgrade" trait to keep them from moving, that obviously makes them unable to fire.
deploytoupgrade
This image is just a test, where I changed the EMP upgrade for this new one.
titan

### DIFF
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -485,7 +485,8 @@
     <Compile Include="Traits\Turreted.cs" />
     <Compile Include="Traits\ProducibleWithLevel.cs" />
     <Compile Include="Traits\Upgrades\DeployToUpgrade.cs" />
-    <Compile Include="Traits\Upgrades\DisableUpgrade.cs" />
+    <Compile Include="Traits\Upgrades\DisableOnUpgrade.cs" />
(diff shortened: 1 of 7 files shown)


penev92:
:+1: 
Of course the commits will have to be squashed when reviewing is done.

atlimit8:
Might be better just to make `Mobile` upgradable. Then `IDisableMove` could be removed.

penev92:
Making `Mobile` upgradable is it's own can of worms and while it would be nice to eventually get that (and then maybe remove this if it becomes redundant), I wouldn't request it in this PR.

atlimit8:
I don't expect upgradable `Mobile` to be in this PR.

abcdefg30:
:+1: once squashed.

obrakmann:
As far as I see, this conflicts with and would be made redundant by #9438. I have nothing against the code in particular, but it seems odd to merge this only to have replace it a few days later.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c09a06894ee61854bc8846c4c1691e910df67233...389c4e1bc566c5c97120543baa6c3bcd0575345d)

PR #9441, opened 2015-09-26
Removed the included MOUSE.R8

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/9263. The file is unmodified. It was just shipped, because we had to.


pchote:
The same should be true (with slight caveats, like we can now go back to the original d2 format shp cursor for TD) for the cursor shps and pals in the other mods.

pchote:
Likewise for the d2k.pal that we ship in bits, too.

GraionDilach:
I'd vote for keeping the D2k pal though, modders might need it for their SHP tools.

pchote:
The palette is shipped (as `PALETTE.BIN`) in the d2k content download / cd install.

penev92:
Btw that palette file, whatever it is, while downloaded with the asset packages, is not installed from CD (I suppose it's not even a D2k palette?), so is likely completely unused. Or at the very least redundant.

abcdefg30:
Cursors in d2k still work. :+1: / :white_check_mark:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/e359b0160a1aceb14036c4bc0c64ee2cc02cd983...a5e78939162b803503972262c39a6a5d6279cb45)

Issue #9385, opened 2015-09-19
Harvesters transported by Carryalls are ordered to find resources by the AI


Mailaender:
Technically it counts as waiting and idle so it constantly forces path searches for new resources which is expensive and needless. Revealed by the debug message I added in https://github.com/OpenRA/OpenRA/pull/9256.

LavenderMoon:
What would be the best way to handle this? Should we add a state for being transported, or add a specific check to the AI against this?

Mailaender:
Adding `IsInWorld` checks might be sufficient.

PR #9442, opened 2015-09-26
Fixed AI bots trying to assign orders to units not in the world

Mailaender:
The AI was ordering units around which are not in the world. This is futile and probably responsible for some performance regressions whenever units are airlifted in. Fixes https://github.com/OpenRA/OpenRA/issues/9385.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -182,7 +182,7 @@ namespace OpenRA.Mods.Common.AI
 		readonly IPathFinder pathfinder;
 
 		readonly Func<Actor, bool> isEnemyUnit;
-		readonly Predicate<Actor> unitIsDeadOrHasNewOwner;
+		readonly Predicate<Actor> unitCannotBeOrdered;
 		Dictionary<SupportPowerInstance, int> waitingPowers = new Dictionary<SupportPowerInstance, int>();
 		Dictionary<string, SupportPowerDecision> powerDecisions = new Dictionary<string, SupportPowerDecision>();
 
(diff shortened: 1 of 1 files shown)


RoosterDragon:
Won't this cause the AI to lose track of the squads for units that are only temporarily "out of the world"? Could that be bad?

Mailaender:
The AI doesn't know how to use troop transports so I can't think of a case where this effects it. It re-scans for new units periodically (as new ones are produced) and assigns them to squads or does the special case orders like MCV deployment and resource to harvest search. Otherwise Dune 2000 harvesters that are brought in by carryall and only after the dropdown in the world won't work which they now do without the order spam during flight.

atlimit8:
Needs a rebase.

Mailaender:
Rebased.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -182,7 +182,7 @@ public CPos GetRandomBaseCenter()
 		readonly IPathFinder pathfinder;
 
 		readonly Func<Actor, bool> isEnemyUnit;
-		readonly Predicate<Actor> unitIsDeadOrHasNewOwner;
+		readonly Predicate<Actor> unitInWorldIsDeadOrHasNewOwner;
```
atlimit8: Should be something like `unitOutOfWorldIsDeadOrHasNewOwner` since it checks `!a.IsInWorld`.

obrakmann: I'd probably go with `unitCannotBeOrdered` or something.

obrakmann:
Looks fine. :+1: once atlimit8's comment is addressed.

OpenRA.Mods.Common/AI/HackyAI.cs
Mailaender: Renamed.

pchote:
Confirmed that this fixes the issue, thanks. :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/ea2be71805e30e30244e03fdc92277c1e3cb828f

PR #9444, opened 2015-09-26
Remove overridden ButtonBaseLine line in mods/modchooser/metrics.yaml

atlimit8:


abcdefg30:
Looks simple enough. :+1:

Mailaender:
Strange that this didn't error with `key already in dictionary` as it usually does.

PR #9445, opened 2015-09-26
Provide an early exit in RunActivity when there is no activity

RoosterDragon:
Provides a tiny speedup for actors that are idle, as we can avoid calling `Stopwatch.GetTimestamp()`.


abcdefg30:
:+1:

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4f882ebb9d8369666a4bf68d2767aeb38c8d186c...554c9dec5ba65f162a21d4b1734b9ef2b15886e7)

PR #9262, opened 2015-09-04
Speed up FrozenUnderFog.Tick

RoosterDragon:
FrozenUnderFog.Tick gets called a lot, and most of that time is spent doing the two dictionary lookups.

This change stuffs everything into one dictionary. This means only one lookup and saves 40% of the time spent in this method.

PR #9452, opened 2015-09-26
Move FrozenActor creation to Created in FrozenUnderFog

atlimit8:
Move `FrozenActor` creation from `FrozenUnderFog.Tick` to `FrozenUnderFog` constructor as discussed in #9262 after merge.

### DIFF
--- a/OpenRA.Game/Primitives/PlayerDictionary.cs
+++ b/OpenRA.Game/Primitives/PlayerDictionary.cs
@@ -11,7 +11,7 @@ namespace OpenRA.Primitives
 		readonly T[] valueByPlayerIndex;
 		readonly Dictionary<Player, T> valueByPlayer;
 
-		public PlayerDictionary(World world, Func<Player, T> valueFactory)
+		public PlayerDictionary(World world, Func<Player, int, T> valueFactory)
 		{
 			var players = world.Players;
 			if (players.Length == 0)
@@ -24,11 +24,15 @@ namespace OpenRA.Primitives
 			for (var i = 0; i < players.Length; i++)
 			{
 				var player = players[i];
-				var state = valueFactory(player);
+				var state = valueFactory(player, i);
 				valueByPlayerIndex[i] = state;
 				valueByPlayer[player] = state;
 			}
 		}
+
+		public PlayerDictionary(World world, Func<Player, T> valueFactory)
+			: this(world, (player, _) => valueFactory(player))
+		{ }
 
 		/// <summary>Gets the value for the specified player. This is slower than specifying a player index.</summary>
 		public T this[Player player] { get { return valueByPlayer[player]; } }
--- a/OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
+++ b/OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
@@ -28,7 +28,7 @@ namespace OpenRA.Mods.Common.Traits
 		public object Create(ActorInitializer init) { return new FrozenUnderFog(init, this); }
 	}
 
-	public class FrozenUnderFog : IRenderModifier, IDefaultVisibility, ITick, ITickRender, ISync
+	public class FrozenUnderFog : IRenderModifier, IDefaultVisibility, ITick, ITickRender, ISync, INotifyCreated
 	{
 		[Sync] public int VisibilityHash;
 
@@ -39,7 +39,6 @@ namespace OpenRA.Mods.Common.Traits
 		PlayerDictionary<FrozenState> frozenStates;
 		ITooltip tooltip;
 		Health health;
-		bool initialized;
 		bool isRendering;
 
 		class FrozenState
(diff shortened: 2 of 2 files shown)


RoosterDragon:
Note to reviewers: I *think* there may be some calls made to `IsVisible` before the first ticks occurs which would previously return false but may now return true if `StartsRevealed` was set. Not actually sure if that affects anything but worth checking in case that changes behaviour.

OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
```diff
@@ -95,20 +100,10 @@ public void Tick(Actor self)
 			{
 				FrozenActor frozenActor;
 				bool isVisible;
```
RoosterDragon: You can now merge these declarations with their assignments.

atlimit8:
Changed.

RoosterDragon:
The problem I noted in my earlier comment applies. If visibility is queried for something that starts revealed, hasn't had a chance to tick yet, and is under the fog, it will now be visible.

This means newly created buildings created by other players in their fog will flash up on the screen and radar for a single frame, since they haven't had a chance to tick yet and become invisible.

Not sure how to fix that at the moment.

atlimit8:
Restored `initialized` flag.

RoosterDragon:
You can remove the use of the flag in `ModifyRender` since your change now means that won't explode.

:+1: LGTM after that.

atlimit8:
Changed.

obrakmann:
Targeting frozen actors doesn't work anymore on this branch.

atlimit8:
Are they visible but not targetable?

obrakmann:
Yes, correct.

RoosterDragon:
Any plans on tackling this? I feel like it might be easier to just leave everything as-is.

atlimit8:
At least until the right bug(s) are fixed.

atlimit8:
#9672 might help.

atlimit8:
#9672 helped, but `AddedToWorld` is required to get targeting working, suggesting a possible bug in targeting.

atlimit8:
Forgot to click reopen.

RoosterDragon:
:+1: Doesn't regress the targeting and looks a bit cleaner so sure.

OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
```diff
@@ -62,10 +62,26 @@ public FrozenUnderFog(ActorInitializer init, FrozenUnderFogInfo info)
...
 			footprint = footprintCells.SelectMany(c => map.ProjectedCellsCovering(c.ToMPos(map))).ToArray();
+
 			tooltip = Exts.Lazy(() => init.Self.TraitsImplementing<ITooltip>().FirstOrDefault());
 			health = Exts.Lazy(() => init.Self.TraitOrDefault<Health>());
 		}
 
+		public void AddedToWorld(Actor self)
+		{
+			if (initialized)
+				return;
```
penev92: Perhaps we should get rid of this now?
`AddedToWorld()` will be called only once under normal circumstances. In more special cases, people could end up wanting to change the state of the actor when making it disappear and reappear again (fancy weapons or support powers maybe?).

atlimit8: If this is removed and circumstances change, then that bug would be exposed. One easy case for exposing such is to give an actor both `FrozenUnderFog` and `Passenger` traits such as a mod or a map that uses `FrozenUnderFog` for showing a last seen location of a transportable actor.

obrakmann:
For some reason the `StartsRevealed` setting doesn't work for civilian buildings (start a game with shroud off). I can't quite see why yet, though.

atlimit8:
That is probably caused by the order things happen. More plumbing will probably have to be reworked for this to work.

RoosterDragon:
I think I worked out the key to this one.

Currently, on the first tick `initialized` is false and thus the visibility isn't changed. This mean the initial visibility (which is `startsRevealed` is used).

When you moved it out, the check was removed from tick. Thus, on the first tick, it immediately overwrites the `startsRevealed` visibility and that never gets a chance to be taken into account.

I think if, when you first create the frozen actor, you immediately set all the state if `startsRevealed` is true (the owner, tooltip and health) - that should be enough.

atlimit8:
Reworked.

OpenRA.Game/Primitives/PlayerDictionary.cs
```diff
@@ -11,7 +11,7 @@ public struct PlayerDictionary<T> : IReadOnlyList<T>, IReadOnlyDictionary<Player
 		readonly T[] valueByPlayerIndex;
 		readonly Dictionary<Player, T> valueByPlayer;
 
-		public PlayerDictionary(World world, Func<Player, T> valueFactory)
+		public PlayerDictionary(World world, Func<Player, int, T> valueFactory)
```
RoosterDragon: Could you provide a constructor that still has the old signature for convenience and just have it delegate to this one.

atlimit8: Sure, I actually did but saw only one other occurrence.

OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
```diff
@@ -64,6 +64,50 @@ public FrozenUnderFog(ActorInitializer init, FrozenUnderFogInfo info)
...
+			{
+				frozenActor.TooltipInfo = tooltip.TooltipInfo;
+				frozenActor.TooltipOwner = tooltip.Owner;
+			}
+		}
+
+		public void AddedToWorld(Actor self)
+		{
+			if (initialized)
+				return;
```
RoosterDragon: Will using `INotifyCreated` instead work for this? Would be nice to avoid having to have this flag.

OpenRA.Game/Primitives/PlayerDictionary.cs
RoosterDragon: #10324 will add another usage, if that helps.

atlimit8:
`INotifyCreated.Created` appears to work.

OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
```diff
@@ -64,6 +64,44 @@ public FrozenUnderFog(ActorInitializer init, FrozenUnderFogInfo info)
 			footprint = footprintCells.SelectMany(c => map.ProjectedCellsCovering(c.ToMPos(map))).ToArray();
 		}
 
+		public void Created(Actor self)
+		{
+			tooltip = self.TraitsImplementing<ITooltip>().FirstOrDefault();
+			health = self.TraitOrDefault<Health>();
+
+			VisibilityHash = 0;
```
RoosterDragon: No need to set this here, it'll be still be 0.

OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
```diff
@@ -64,6 +64,44 @@ public FrozenUnderFog(ActorInitializer init, FrozenUnderFogInfo info)
...
+			frozenStates = new PlayerDictionary<FrozenState>(self.World, (player, playerIndex) =>
+			{
+				var frozenActor = new FrozenActor(self, footprint, player.Shroud, startsRevealed);
+				if (startsRevealed)
+					UpdateFrozenActor(self, frozenActor, playerIndex);
+				player.PlayerActor.Trait<FrozenActorLayer>().Add(frozenActor);
+				return new FrozenState(frozenActor) { IsVisible = startsRevealed };
+			});
+
+			initialized = true;
```
RoosterDragon: Now you've set all the state right away, I think you can eradicate this variable wholesale.

atlimit8:
Changed.

OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
```diff
@@ -64,13 +63,47 @@ public FrozenUnderFog(ActorInitializer init, FrozenUnderFogInfo info)
 			footprint = footprintCells.SelectMany(c => map.ProjectedCellsCovering(c.ToMPos(map))).ToArray();
 		}
 
+		public void Created(Actor self)
```
pchote: Note that `Created` is *not* equivalent to `AddedToWorld`.  An actor may be added or removed from the world multiple times between its creation and disposal.

pchote: Shouldn't be a problem in this case, just making a note for anyone who might be following the notification chain.

RoosterDragon:
:+1: Civ buildings are correctly revealed at the start of matches with the explored map option now, and frozen actors seem otherwise fine still.

Mailaender:
Also can't spot any regressions from this refactor. :white_check_mark:

PR #9453, opened 2015-09-27
Change D2k Cost and CustomBuildValues to match original

reaperrr:
Additionally changed queue build speeds and set prerequisite for all starport units to just starport.

Another chunk from #9106.


penev92:
Looks OK to me :+1:

pchote:
Seems reasonable, and nobody else has complained so :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/b9700fd444271676c607c81aa65b546a58e26ccf

PR #9454, opened 2015-09-27
TD Additional Balance Changes

AoAGeneral:
Additional fixes and ideas:

Infantry stealth detection: Increase sight range of stealth detection to
2 from 1.

MLRS: Reduce price to 900 from 1200

Repair pad: Power to 10 from 30.

Blue Tiberium: Reduce the value income to 75-90% from 110%

Infantry evade chance: 75 from 67

Samsites: cost reduced to 700 from 750

Airstrikes: Increase cool down timer to 3:30 from 3:00.

Forums location: http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=17430&start=30

Might add a bit more (If I don't screw anything else up)


mods/cnc/rules/civilian.yaml
```diff
@@ -430,7 +430,7 @@ C10:
 VICE:
 	Inherits: ^Viceroid
 	AttackWander:
-		WanderMoveRadius: 2
-		MinMoveDelayInTicks: 25
-		MaxMoveDelayInTicks: 45
+		WanderMoveRadius: 1
+		MinMoveDelayInTicks: 1500
```
pchote: This looks excessive (visceroids will now move only once cell per minute!), and isn't mentioned in your changelog.

mods/cnc/rules/defaults.yaml
```diff
@@ -223,7 +223,7 @@
 		Upgrades: hospitalheal
 		Prerequisites: hosp
 	DetectCloaked:
-		Range: 1c0
+		Range: 3c0
```
pchote: What is the logic behind this change?  It makes stealth tanks much less stealthy.

AoAGeneral:
The movement one cell per minute prevents visceroids on maps such as Hour Glass and Lexington from attacking your starting location bases right away. Previously two releases ago they stood still and did not wander at all. But a change happened somewhere where someone added it to make them wander and they wander way to frequently. (Also true that it isnt added in the notes I will add that now)

As for the Stealth detection the range on Stealth Tanks themselves outranges the mammoth tanks so infantry still have to move closer (Since infantry fire range is 4c0) which means stealth tanks are still rather stealthed. Also, using stealth tanks by themselves is not the best counter as mixing them together with a backing army makes them more deadly. Using stealth tanks alone should not be the all counterable meta. Adding stealth detection to infantry also initiates more infantry play.

pchote:
You're effective just turning off the visceroid movement (in a half hearted way), which doesn't sit well with me.  If there's a problem then we should identify and fix it, and not just plaster over it.

AoAGeneral:
Fair enough. Both the mentioned maps above are not played anymore due to visceroids for example. So can work on doing something about it. I will see about reverting the changes to visceroids to defaults.

How about the stealth armada?

pchote:
No real comment - it seems a bit extreme (why not 2?), but that is based on feel rather than gameplay experience. Future playtesting (as long as you guys actually start using the playtests!) will reveal whether or not the change was appropriate.

AoAGeneral:
Myself and Anjew used the playtests before this last release testing light tanks. Worked out rather well. They will be tested on more closely. 

As for range of 2 I thought about doing such but it gives E3 less chance for a response vs stealth tanks. They move rather quickly before they can fire at multiple targets and then they are either crushed or escaped. Also keeping in mind, the E3 missiles will go dead lock when the tanks go stealth. Which means if they leave the 2 cell range the missiles go dead. Makes the E3 ineffective.

AoAGeneral:
Reverted the visceroids change. Up for discussion for the fix on those two maps (and possible other future ones/edit the maps on removal of the creatures.)

Following Pchote's advice on going with 2 with stealth detection for testing then later try 3 if 2 isn't enough. Main concern is the E3 dead lock missiles. But will see if minigunners and other infantry is enough.

abcdefg30:
:+1: 

Could you squash the commits?

AoAGeneral:
well I just screwed that up. Awesome.

PR #9458, opened 2015-09-27
Change D2k shroud revealing ranges to match original

reaperrr:

PR #9459, opened 2015-09-27
Change D2k power draw and various other aspects to match original

reaperrr:
After #9453, #9458 and this there should be only some minor loose ends remaining (Palace infantry, tweaking debris etc.).

### DIFF
--- a/mods/d2k/rules/aircraft.yaml
+++ b/mods/d2k/rules/aircraft.yaml
@@ -38,6 +38,10 @@ carryall.reinforce:
 		Automatic: False
 	RenderSprites:
 		Image: carryall
+	SelfHealing:
+		Step: 5
+		Ticks: 3
+		HealIfBelow: 50%
 
 carryall:
 	Inherits: carryall.reinforce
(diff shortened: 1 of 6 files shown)


penev92:
Making infantry not attract worms will make them a whole lot more difficult to deal with, so I'm hesitant on that one.
Otherwise looks good :+1: 
Adding `Changes requested` for the engineers' `CaptureTypes`.

reaperrr:
> Making infantry not attract worms will make them a whole lot more difficult to deal with, so I'm hesitant on that one.

That's what Thumpers should be for once they're implemented (which should be easy now with DeployToUpgrade and then enabling AttractWorms via upgrade on them).

reaperrr:
Updated.

pchote:
Needs a rebase.

reaperrr:
Rebased.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d328e377beb374b133bd97ab98acfc5a0bbe28b0...d78cef09b26889be39fb63508d15e85e691426bc)

PR #9460, opened 2015-09-27
Exposed pips to trait documentation and lint testing

Mailaender:
This is a common source for errors. Makes a start at cleaning it up.

### DIFF
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -19,9 +19,8 @@ using OpenRA.Primitives;
 
 namespace OpenRA.Traits
 {
-	// depends on the order of pips in WorldRenderer.cs!
+	// Depends on the order of pips in SelectionDecorations.cs!
 	public enum PipType { Transparent, Green, Yellow, Red, Gray, Blue, Ammo, AmmoEmpty }
-	public enum TagType { None, Fake, Primary }
 
 	[Flags]
 	public enum Stance
@@ -225,7 +224,6 @@ namespace OpenRA.Traits
 	public interface ILoadsPlayerPalettes { void LoadPlayerPalettes(WorldRenderer wr, string playerName, HSLColor playerColor, bool replaceExisting); }
 	public interface IPaletteModifier { void AdjustPalette(IReadOnlyDictionary<string, MutablePalette> b); }
 	public interface IPips { IEnumerable<PipType> GetPips(Actor self); }
-	public interface ITags { IEnumerable<TagType> GetTags(); }
 	public interface ISelectionBar { float GetValue(); Color GetColor(); }
 
 	public interface IPositionableInfo : ITraitInfo { }
--- a/OpenRA.Mods.Common/Traits/Buildings/PrimaryBuilding.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/PrimaryBuilding.cs
@@ -10,6 +10,7 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using OpenRA.Graphics;
 using OpenRA.Mods.Common.Orders;
 using OpenRA.Traits;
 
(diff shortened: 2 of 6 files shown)


OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
```diff
@@ -33,17 +33,23 @@ public class SelectionDecorationsInfo : ITraitInfo, ISelectionDecorationsInfo
...
+
 		public object Create(ActorInitializer init) { return new SelectionDecorations(init.Self, this); }
 
+		[Desc("Don't change the order of these sequences."), SequenceReference("Image")]
+		public readonly string[] PipSequences = { "pip-empty", "pip-green", "pip-yellow", "pip-red", "pip-gray", "pip-blue", "pip-ammo", "pip-ammoempty" };
+
+		[Desc("Don't change the order of these sequences."), SequenceReference("Image")]
+		public readonly string[] TagSequences = { "", "tag-fake", "tag-primary" };
+
+		// PipSequences and TagSequences depends on the order of pips in TraitsInterfaces.cs!
```
abcdefg30: > OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs:L46: [SA1512] A single-line comment must not be followed by a blank line. To ignore this error when commenting out a line of code, begin the comment with '////' rather than '//'.

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -19,7 +19,7 @@
 
 namespace OpenRA.Traits
 {
-	// depends on the order of pips in WorldRenderer.cs!
+	// depends on the order of pips in SelectionDecorations.cs!
```
abcdefg30: As you're already editing this, start the comment with an uppercase letter, please.

pchote:
For what its worth, the tag code is basically redundant, and can be ripped out after a simple patch changing them to use upgrades instead.

Mailaender:
Updated.

OpenRA.Mods.Common/Traits/Buildings/PrimaryBuilding.cs
```diff
@@ -82,5 +96,21 @@ public void SetPrimaryProducer(Actor self, bool state)
...
 		}
+
+		public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr)
+		{
+			if (!isPrimary)
+				yield break;
+
+			var tagImages = new Animation(wr.World, info.Image);
+			var pal = wr.Palette(info.Palette);
+			var tagxyOffset = new int2(0, 6);
```
abcdefg30: Why is this generated everytime new?
Can't we just have a const. or something like that?
(PS: What about unhardcoding this?)

pchote: See my earlier comment about how this code can be removed completely.

abcdefg30:
Tags still work ingame. :+1: / :white_check_mark:

OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
```diff
@@ -33,17 +33,18 @@ public class SelectionDecorationsInfo : ITraitInfo, ISelectionDecorationsInfo
 
 		public readonly Color SelectionBoxColor = Color.White;
 
+		public readonly string Image = "pips";
+
 		public object Create(ActorInitializer init) { return new SelectionDecorations(init.Self, this); }
 
+		[Desc("Don't change the order of these sequences."), SequenceReference("Image")]
```
pchote: Asking politely isn't a very good approach.  You can enforce this by adding a [LoadUsing] method that ignores any attempted yaml overrides and returns the default sequences.

atlimit8:
Needs a rebase.

OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
pchote: Or alternatively, just don't expose this.

penev92:
Bump.

pchote:
#10104 adds the first half of what i was asking for above, and once that is merged i will file a pr to bleed that removes the tag code completely. It would be better if this could wait until both those prs are merged to avoid unnecessary code churn.

OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
GraionDilach: Although this needs to be dehardcoded in the long term, iirc TS and RA2 had more pips.

pchote: In the long term (or possibly even sooner) this will be replaced with something similar to `WithDecoration` that allows for arbitrary placement and sprite choice for pips.

pchote:
#10216 implements the second half of what I was asking for above (with some other improvements thrown in).

pchote:
It would be better if we could expose pips to lint testing and documentation by implementing a `WithPipsDecoration` trait, following my lead from #10104 and #10216.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/10216/.

Issue #6820, opened 2014-10-22
Add support for rectangular hitboxes [$100]


reaperrr:
The health radius is suboptimal for visually rectangular and/or long units (tanks, ships, some buildings). While it might complicate the damage calculations compared to a simple radius, it would allow for hitboxes matching the visual shape, and might even be a first step towards directional armor.

huwpascoe:
I'm taking on this issue as discussed in irc. I've decided on
Radius for infantry
Axis-aligned rectangle for structures
Capsule (line segment with radius) for vehicles
Polygons which I can't honestly find a good use case for but am including them anyway.

Will deliver a PR soon.

pchote:
The polygon (note: we already have helper functions for polygon intersection testing) covers the vehicle and structure case, so you could avoid some work and duplication.

TheRaffy:
Is this related to any melee attack`?

huwpascoe:
PRs merged. I claim this bounty!

pchote:
Unfortunately this broke our unit tests #10325.

PR #9465, opened 2015-09-27
Added warhead visualization to combat debug overlay.

huwpascoe:
Step one of https://github.com/OpenRA/OpenRA/issues/6820
whdbg

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/WarheadDebugOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/World/WarheadDebugOverlay.cs
@@ -1,0 +1,87 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using System.Drawing;
+using OpenRA.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Part of the combat overlay from DeveloperMode. Attach this to the world actor.")]
+	public class WarheadDebugOverlayInfo : ITraitInfo
+	{
+		public readonly int DisplayDuration = 25;
+
+		public object Create(ActorInitializer init) { return new WarheadDebugOverlay(this); }
+	}
+
+	public class WarheadDebugOverlay : IPostRender
+	{
+		class WHImpact
+		{
+			public readonly WPos CenterPosition;
+			public readonly WDist[] Range;
+			public int Time;
+
+			public WDist OuterRange
+			{
+				get { return Range[Range.Length - 1]; }
+			}
+
+			public WHImpact(WPos pos, WDist[] range, int time)
+			{
+				CenterPosition = pos;
+				Range = range;
+				Time = time;
+			}
+		}
+
+		readonly WarheadDebugOverlayInfo info;
+		readonly List<WHImpact> impacts = new List<WHImpact>();
+
+		public WarheadDebugOverlay(WarheadDebugOverlayInfo info)
+		{
+			this.info = info;
+		}
+
+		public void AddImpact(WPos pos, WDist[] range)
+		{
+			impacts.Add(new WHImpact(pos, range, info.DisplayDuration));
+		}
+
+		public void RenderAfterWorld(WorldRenderer wr, Actor self)
+		{
+			foreach (var i in impacts)
+			{
+				var alpha = 255.0f * i.Time / info.DisplayDuration;
+				var rangeStep = alpha / i.Range.Length;
+
+				wr.DrawRangeCircle(i.CenterPosition, i.OuterRange, Color.FromArgb((int)alpha, Color.Red));
+
+				foreach (var r in i.Range)
+				{
+					var tl = wr.ScreenPosition(i.CenterPosition - new WVec(r.Length, r.Length, 0));
+					var br = wr.ScreenPosition(i.CenterPosition + new WVec(r.Length, r.Length, 0));
+					var rect = RectangleF.FromLTRB(tl.X, tl.Y, br.X, br.Y);
+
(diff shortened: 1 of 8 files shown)


obrakmann:
You've got a typo in the file name.

Also, am I the only one who tried the nuke first thing? Awesome.

OpenRA.Mods.Common/Traits/World/WartheadDebugOverlay.cs
```diff
@@ -0,0 +1,69 @@
...
+using OpenRA.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Part of the combat overlay from DeveloperMode. Attach this to the world actor.")]
+	public class WarheadDebugOverlayInfo : TraitInfo<WarheadDebugOverlay> { }
+	public class WarheadDebugOverlay : IRenderOverlay
+	{
+		const int DisplayDuration = 20;
```
pchote: This should move into the Info where it can be configured.

reaperrr:
@huwpascoe Could you maybe add a thin circle line around the most outer border of the spread? That would help seeing at what distance the warhead stops dealing damage (very low Falloff values like 1 or 2 will otherwise be very hard to see).

pchote:
This certainly looks neat, but it might be more useful to draw circles at constant damage radii (say every 10 or 20%) instead?

huwpascoe:
@reaperrr @pchote I couldn't find a circle drawing function. Is there any code for that laying around that I could stuff into the linerenderer?

huwpascoe:
...I say that while looking at the unit radius in the screenshots. Durrrr. Scratch the last comment.

huwpascoe:
Changes made

huwpascoe:
Small fix, changed to an IPostRender

huwpascoe:
Resolved conflict with rooster's merged PR.

reaperrr:
:+1:

OpenRA.Mods.Common/Warheads/HealthPercentageDamageWarhead.cs
```diff
@@ -22,6 +23,10 @@ public class HealthPercentageDamageWarhead : DamageWarhead
 		public override void DoImpact(WPos pos, Actor firedBy, IEnumerable<int> damageModifiers)
 		{
 			var world = firedBy.World;
+
+			if (world.AllowDevCommands && world.LocalPlayer.PlayerActor.Trait<DeveloperMode>().ShowCombatGeometry)
```
penev92: This crashes in replays because `LocalPlayer` is null.
Could just go for a null check because these aren't supposed to be visible in replays and for spectators in live games.

OpenRA.Mods.Common/Traits/World/WarheadDebugOverlay.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Drawing;
+using OpenRA.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Part of the combat overlay from DeveloperMode. Attach this to the world actor.")]
+	public class WarheadDebugOverlayInfo : ITraitInfo
+	{
+		public readonly int DisplayDuration = 20;
```
penev92: Up this to a full second please (maybe even longer, but nukes and such already take too long due to their 50 warheads).

penev92:
This looks pretty cool, and according to feedback I got from some modders, it will be quite useful.
Code looks good as well.

huwpascoe:
requested changes applied (I'm assuming the update loop runs at 25 ticks per second)

OpenRA.Mods.Common/Traits/World/WarheadDebugOverlay.cs
```diff
@@ -0,0 +1,85 @@
...
+			public readonly WDist OuterRange;
+			public int Time;
+
+			public WHImpact(WPos pos, WDist[] range, int time)
+			{
+				CenterPosition = pos;
+				Range = range;
+				Time = time;
+
+				OuterRange = range[range.Length - 1];
```
atlimit8: Is there a particular reason to store another copy of the outermost range instead of using `range[range.Length - 1]` directly.
If you are trying to avoid calculating `range.Length - 1]` each time, you could do this:
```c#
var alpha = 0f
var rangeStep = 255.0f * i.Time / info.DisplayDuration / i.Range.Length;
...
	alpha += rangeStep;
```
 \- OR for some shading to outermost ellipse -
```c#
var rangeStep = 255.0f * i.Time / info.DisplayDuration / (i.Range.Length - 1);
var alpha = rangeStep;
...
	alpha += rangeStep;
```

OpenRA.Mods.Common/Warheads/HealthPercentageDamageWarhead.cs
```diff
@@ -22,6 +23,11 @@ public class HealthPercentageDamageWarhead : DamageWarhead
 		public override void DoImpact(WPos pos, Actor firedBy, IEnumerable<int> damageModifiers)
 		{
 			var world = firedBy.World;
+
+			var devMode = world.LocalPlayer.PlayerActor.TraitOrDefault<DeveloperMode>();
```
penev92: This still crashes because `LocalPlayer` is `null`.
Put this whole thing in a `if (world.LocalPlayer != null)`.

OpenRA.Mods.Common/Traits/World/WarheadDebugOverlay.cs
huwpascoe: I figured it was more readable but you're right I should probably be using a getter or something instead.

huwpascoe:
Fixed things

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/389c4e1bc566c5c97120543baa6c3bcd0575345d...70d67a23b91ded183a11a1aedda1e21feb34da76)

pchote:
> (I'm assuming the update loop runs at 25 ticks per second)

This assumption does not hold anymore.  We support multiple game speeds now, each with a different number of ticks per second.  Not a huge deal for this PR, though.

deniz1a:
Mechanics and medics also have red damage overlays. How about making them green?

PR #9467, opened 2015-09-27
Refactor Missile projectile

reaperrr:
Essentially the latest version of #8717, only without the last commit (the in-advance height-level check it adds is plagued by some issues, let's deal with that later) and rebased.


reaperrr:
Rebased.

PR #9478, opened 2015-09-28
PNG file size reduction

clemty:
run pngcrush -brute on *.png

before:
- 70 File(s)      3.455.140 bytes

after:
- 70 File(s)      2.616.088 bytes

that's a 800kb size reduction that should translate almost identically to download file size

(edit: pngcrush is lossless)


reaperrr:
I haven't looked at them, but I remember we've had issues with 24 bit vs. 32 bit in the past, so we have to make sure these don't change whatever we're using (32 bit, iirc).

clemty:
according to the windows explorer, the bit depths of both sets (before and after) are identical:
the majority of files are __32bit__, with two exceptions:
* mods\d2k\maps\atreides-01a\map.png __(24bit)__
* mods\d2k\maps\atreides-01b\map.png __(8bit)__

afaik pngcrush does not change bit depth (from pngcrush -h):
```
  -bit_depth depth (deprecated)
             This option no longer does anything.
```

clemty:
updated, now all files are 32 bit

RoosterDragon:
:+1:

obrakmann:
Looks fine to me. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/303d85fc75864b6af5749c2c9259db0acaee007b)

PR #9483, opened 2015-09-29
Clean up assets

### LINKED ISSUE #9482: Remove the rest of the unused installation assets
#9441 made a start, but there are a bunch of other assets that were left behind.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27039230-remove-the-rest-of-the-unused-installation-assets?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Fixes #9482.


OpenRA.Game/ModData.cs
```diff
@@ -74,10 +74,7 @@ public ModData(string mod, bool useLoadScreen = false)
 			SpriteSequenceLoader = (ISpriteSequenceLoader)ctor.Invoke(new[] { this });
 			SpriteSequenceLoader.OnMissingSpriteError = s => Log.Write("debug", s);
 
-			// HACK: Mount only local folders so we have a half-working environment for the asset installer
-			GlobalFileSystem.UnmountAll();
```
pchote: You may need to keep this line - check that things don't break when switching between mods.

reaperrr:
I'm puzzled by these Travis errors, no idea what to do about them.

reaperrr:
Updated.

OpenRA.Game/ModData.cs
```diff
@@ -74,10 +74,8 @@ public ModData(string mod, bool useLoadScreen = false)
 			SpriteSequenceLoader = (ISpriteSequenceLoader)ctor.Invoke(new[] { this });
 			SpriteSequenceLoader.OnMissingSpriteError = s => Log.Write("debug", s);
 
-			// HACK: Mount only local folders so we have a half-working environment for the asset installer
+			// We cannot put GlobalFileSystem.LoadFromManifest(Manifest) here because Travis doesn't like that
 			GlobalFileSystem.UnmountAll();
```
pchote: We probably don't need this or the comment above anymore.  The reasonable default assumption is that constructors don't have ugly side-effects on global state.

reaperrr:
Updated again.

mods/d2k/cursors.yaml
```diff
@@ -1,7 +1,7 @@
 ShadowIndex: 1
 
 Palettes:
-	mouse: d2k.pal
+	mouse: palette.bin
```
GraionDilach: These were uppercase in the disc and uppercase in my content folder as well. You're breaking Linux here.

penev92:
Seems reasonable :+1: 
But shouldn't you update the palette file in `palettes.yaml` and `arrakis.yaml` to reflect the change @GraionDilach requested for `cursors.yaml `?

reaperrr:
> But shouldn't you update the palette file in palettes.yaml and arrakis.yaml to reflect the change @GraionDilach requested for cursors.yaml?

That needs some kind of an update, yes. 
However, personally I'm rather inclined to make all palette traits load files with `.ToLowerInvariant()` to avoid "accidents" in the future (for 3rd-party modders as well).

I need to check whether that's possible for the cursor special case, if it is, that applies there, too.

GraionDilach:
That should be out of scope and would actually make the whole thing worse than how it is right now. Although I'll open an issue about it.

reaperrr:
Updated, changed D2k to upper-case `PALETTE.BIN` throughout.

obrakmann:
Looks ok. All mods still work, and switching between mods as well. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/98c6ec7668489446f1fc90a2b31d5a45ad629df0)

Mailaender:
You sadly forgot to add `palette.bin` to `TestFiles` and `CopyFilesFromCD` at https://github.com/OpenRA/OpenRA/blob/bleed/mods/d2k/mod.yaml#L129 so now CD installations will crash quite early due to the file missing.

PR #9485, opened 2015-09-29
Fix crash in cnc64gdi01

clemty:
apparently helicopters can't .Hunt()
I introduced the bug late in testing, so it didn't trigger in review.
On the plus side, AI will now attack you from a random direction

### DIFF
--- a/mods/cnc/maps/cnc64gdi01/cnc64gdi01.lua
+++ b/mods/cnc/maps/cnc64gdi01/cnc64gdi01.lua
@@ -15,6 +15,12 @@ AutocreateSquads =
 	{ "arty", "arty" }
 }
 
+HeliPatrolPaths =
+{
+	{ HeliPatrol1.Location, HeliPatrol2.Location, HeliPatrol3.Location, HeliPatrol4.Location, HeliPatrol5.Location, HeliPatrol6.Location },
+	{ HeliPatrol5.Location, HeliPatrol4.Location, HeliPatrol3.Location, HeliPatrol2.Location, HeliPatrol1.Location, HeliPatrol6.Location }
+}
+
 AttackTriggers = { AttackTrigger1, AttackTrigger2, AttackTrigger3, AttackTrigger4 }
 
 harvester = { "harv" }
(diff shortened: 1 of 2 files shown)


obrakmann:
Oh, I shot them down before they started the hunting, that's why I didn't see the crash. Anyway, there's no reason at all why they shouldn't be able to hunt.

clemty:
the crash wouldn't be triggered if they come across an enemy and start attacking him on the way ther - it only triggered for me since I build the base to the right near the tiberium fields, and the helis reached their waypoint before starting .Hunt()

obrakmann:
Looks fine :+1: . I just noticed that the biolab still has the owner row in the tooltip. You might want to fix that as well, while you're at it.

penev92:
@obrakmann is right
>there's no reason at all why they shouldn't be able to hunt.

I can't see any either. We should probably get to the bottom of that.

mods/cnc/maps/cnc64gdi01/cnc64gdi01.lua
```diff
@@ -176,7 +176,16 @@ end
 
 HeliHunt = function()
 	local helicopters = enemy.GetActorsByType("heli")
-	SquadHunt(helicopters)
+	local patrolpath = Utils.Random({
```
abcdefg30: Can we get that extracted to a global variable at the beginning of the script? Like
```
HeliPatrolPaths =
{
	{ HeliPatrol1.Location, HeliPatrol2.Location, HeliPatrol3.Location, HeliPatrol4.Location, HeliPatrol5.Location, HeliPatrol6.Location },
	{ HeliPatrol5.Location, HeliPatrol4.Location, HeliPatrol3.Location, HeliPatrol2.Location, HeliPatrol1.Location, HeliPatrol6.Location } 
}
```
Then just use `local patrolpath = Utils.Random(HeliPatrolPaths)` here.

clemty:
updated

obrakmann:
> I can't see any either. We should probably get to the bottom of that.

@penev92: see #9489. Basically because our activities suck. Traits and activities regularly need to call `self.CancelActivitiy()` and stomp all over the activity queue to make themselves work.

@clemty: I swear I saw the owner row on the biolab, but maybe I'm going crazy. Should have taken a screenshot to assure myself of my sanity.

Ah, it was the husk: 
screenshot from 2015-10-01 21 17 52

clemty:
updated

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d8f62211d9bef94032d212a17ff5fd5781b1fb39)

PR #9486, opened 2015-09-29
Fix AI queueing structure build orders twice at higher game speeds due to lag

### LINKED ISSUE #9403: AI seems to have two BaseBuilders per construction yard
At least in TD. It has a strong tendency to build two radars and two adv. comm. centers even with just 1 construction yard.

So either they have one more basebuilder than they should, or base builders don't stick to their queue type (base or support).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26845392-ai-seems-to-have-two-basebuilders-per-construction-yard?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Fixes #9403.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -56,13 +56,13 @@ namespace OpenRA.Mods.Common.AI
 		[Desc("Minimum excess power the AI should try to maintain.")]
 		public readonly int MinimumExcessPower = 0;
 
-		[Desc("Delay (in ticks) between structure production checks when there is no active production.",
-			"A StructureProductionRandomBonusDelay is added to this.")]
+		[Desc("Additional delay (in ticks) between structure production checks when there is no active production.",
+			"StructureProductionRandomBonusDelay is added to this.")]
 		public readonly int StructureProductionInactiveDelay = 125;
 
-		[Desc("Delay (in ticks) between structure production checks when actively building things.",
-			"A StructureProductionRandomBonusDelay is added to this.")]
-		public readonly int StructureProductionActiveDelay = 10;
+		[Desc("Additional delay (in ticks) added between structure production checks when actively building things.",
+			"Note: The total delay is gamespeed OrderLatency x 4 + this + StructureProductionRandomBonusDelay.")]
+		public readonly int StructureProductionActiveDelay = 0;
 
 		[Desc("A random delay (in ticks) of up to this is added to active/inactive production delays.")]
 		public readonly int StructureProductionRandomBonusDelay = 10;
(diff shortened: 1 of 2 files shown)


penev92:
This is a straight nerf for the AI :(
Can we just remove this property and replace
`waitTicks = active ? ai.Info.StructureProductionActiveDelay + randomFactor`
with
`waitTicks = active ? 3 * world.OrderLatency + randomFactor` ?

reaperrr:
> waitTicks = active ? 3 * world.OrderLatency + ai.Info.StructureProductionActiveDelay + randomFactor

Done. Although I set the default to 2 instead of 0 because it (very rarely) still happens at fastest speed.

penev92:
You should revert/update the Desc on the property now. And maybe put it as a comment on the `3 * ...` to explain what that magic number is.

reaperrr:
Updated.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -56,13 +56,14 @@ public sealed class HackyAIInfo : IBotInfo, ITraitInfo
...
+		[Desc("Additional delay (in ticks) between structure production checks when there is no active production.",
+			"StructureProductionRandomBonusDelay is added to this.")]
 		public readonly int StructureProductionInactiveDelay = 125;
 
-		[Desc("Delay (in ticks) between structure production checks when actively building things.",
-			"A StructureProductionRandomBonusDelay is added to this.")]
-		public readonly int StructureProductionActiveDelay = 10;
+		[Desc("Additional delay (in ticks) added between structure production checks when actively building things.",
+			"Note: The total delay is gamespeed OrderLatency x 3 + this + StructureProductionRandomBonusDelay.",
+			"A value of at least 2 or 3 is recommended to avoid AI issuing duplicate build orders.")]
```
pchote: Is this last sentence still true? That should be completely solved by the OrderLatency change.

pchote:
:+1: after addressing my question above.

reaperrr:
@pchote:
> Is this last sentence still true? That should be completely solved by the OrderLatency change.

Unfortunately it was still true, I still encountered it at fastest gamespeed with OrderLatency * 3. Only very rarely, but not never.

However, after some pondering I've now removed that part of the description, changed the active delay default to 0 instead of 2, and instead increased the hardcoded minimum to 4*OrderLatency.

Compared to before this update, this only increases the total delay by 1 tick at slow and normal speed (4x3 instead of 3x3 +2), 2 ticks at 'faster' (4x4 instead of 3x4 +2) and 4 at 'fastest' where the issue showed up most (6x4 instead of 6x3 +2). Compared to bleed it's +2 / +2 / +6 / +14, respectively.

That way the effect on AI efficiency should be negligible at slow and normal and offset by human player's limits at faster and fastest, and the duplication issue should be gone for good*.

*fingers crossed.

abcdefg30:
Seems to fix the issue. :+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1a216bd80f339fca04b3ff291b23604b8abca872...28c9216915c3e86cbeff49b246524725d491e2e9)

PR #9488, opened 2015-09-29
Map Music

### LINKED ISSUE #9472: Allow maps to define their own music
Right now maps like Fort Lonestar rely on definitions in the mod-shipped music.yaml for their ambient track.  It is a simple to change to add a Music: block to map.yaml.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27005761-allow-maps-to-define-their-own-music?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

atlimit8:
Load additional/replacement music from `map.yaml` with support for loading music from map assets.
Closes  #9472.


OpenRA.Mods.Common/Traits/World/MusicPlaylist.cs
```diff
@@ -52,13 +53,26 @@ public class MusicPlaylist : INotifyActorDisposing, IGameOver
 		public MusicPlaylist(World world, MusicPlaylistInfo info)
 		{
 			this.info = info;
-			this.world = world;
 
-			IsMusicInstalled = world.Map.Rules.InstalledMusic.Any();
+			// Build InstalledMusic preferring existing in Map.Music over Rules.InstalledMusic
```
pchote: Convention is to do this in `Ruleset.Load`, where we are currently explicitly passing in an empty set of map rules to merge in.

OpenRA.Game/Map/Map.cs
```diff
@@ -206,6 +207,19 @@ static object LoadVideos(MiniYaml y)
 			return videos;
 		}
 
+		[FieldLoader.LoadUsing("LoadMusic")] public ReadOnlyDictionary<string, MiniYaml> Music;
```
pchote: Are we able to use the auto dictionary loading for this kind of thing yet?

atlimit8: Not with a `MiniYaml` type (and only with `Dictionary<,>`).
I could add that somewhere.

atlimit8:
Uses `RulsetCache.Load` now.

OpenRA.Mods.Common/Traits/World/MusicPlaylist.cs
```diff
@@ -54,6 +55,10 @@ public MusicPlaylist(World world, MusicPlaylistInfo info)
 			this.info = info;
 			this.world = world;
 
+			// Reload music with map assets mounted
+			foreach (var entry in world.Map.Rules.Music)
+				entry.Value.Reload();
```
pchote: This forces the music tracks to be reloaded / reparsed, so I expect this will introduce a jank in the first tick of the game (but I haven't tested this to verify it).  If we absolutely must support this, then it should be done early enough to be hidden by the load screen.

pchote:
What's the use case for the `OnlyMapMusic` option?

atlimit8:
2 actually (use cases for the `OnlyMapMusic` option):
 1. Allow maps to use a more fitting list of music for that particular map without being restricted to one piece of music.
 2. Maps that function as mini-mods. (arguably hacks)

* Another option would be selectable list.
  * Except that would require listing each piece id twice for map specific (technically all) entries instead of copy and paste to import mod level pieces.
  * However, a list could be used order the playlist
  * After thinking about it, I am open to doing this instead.

Overriding `Hidden` to be `true` would miss any added to the mod after the map was last edited.

OpenRA.Mods.Common/Traits/World/MusicPlaylist.cs
atlimit8: Calling `Reload` for each is needed since any could be overridden or added by assets in the map. I am looking into somewhere else.

atlimit8:
Moved late loading to `ModData.PrepareMap`.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -55,14 +56,26 @@ public void PlaySound(string file)
...
 		public void PlayMusic(string track = null, LuaFunction func = null)
 		{
 			if (!playlist.IsMusicAvailable)
 				return;
 
-			var musicInfo = !string.IsNullOrEmpty(track) ? world.Map.Rules.Music[track]
-			: playlist.GetNextSong();
+			MusicInfo musicInfo;
+			if (string.IsNullOrEmpty(track))
+			{
```
abcdefg30: Seems like you can omit the braces here and at the `else if` below.

OpenRA.Mods.Common/Widgets/Logic/MusicPlayerLogic.cs
```diff
@@ -36,7 +39,10 @@ public MusicPlayerLogic(Widget widget, Ruleset modRules, World world, Action onE
 			BuildMusicTable();
 
 			Func<bool> noMusic = () => !musicPlaylist.IsMusicAvailable || musicPlaylist.CurrentSongIsBackground || currentSong == null;
-			panel.Get("NO_MUSIC_LABEL").IsVisible = () => !musicPlaylist.IsMusicAvailable;
+			panel.Get("NO_MUSIC_INSTALLED_LABEL").IsVisible =
+				!world.Map.Options.OnlyMapMusic && !musicPlaylist.IsMusicAvailable ? True : False;
```
abcdefg30: Why aren't you using `panel.Get("NO_MUSIC_INSTALLED_LABEL").IsVisible = () => !world.Map.Options.OnlyMapMusic && !musicPlaylist.IsMusicAvailable;` here (and below)?

atlimit8: The expressions have constant values once the map is loaded. If this is too out of place, I could replace it.

pchote:
I'm still not convinced about / a fan of `OnlyMapMusic` and the backend changes that support it, but f6180d7fa4caacfa0783ae4645a72a77ebf118f4 and eddf439682ce41a9511da9959892b1d49cb43c85 are unconversial and I want to get them in before the next feature freeze.  Can you please split those into their own pr?

atlimit8:
Removed `OnlyMapMusic`.

mods/ra/maps/fort-lonestar/map.yaml
```diff
@@ -20,6 +20,11 @@ Type: Minigame
 
 Videos:
 
+Music:
+	# Special effects
```
pchote: Minor nit: this comment is now redundant.

OpenRA.Game/ModData.cs
```diff
@@ -171,6 +171,11 @@ public Map PrepareMap(string uid)
 			using (new Support.PerfTimer("Map.SequenceProvider.Preload"))
 				map.SequenceProvider.Preload();
 
+			// Load music with map assets mounted
+			using (new Support.PerfTimer("Music"))
```
pchote: This should be "Map.Music" to be consistent with the other times.

pchote:
Looks good to me now, aside from those minor nits. :+1:

atlimit8:
Nits applied.

abcdefg30:
Works ingame and the code looks good to me. :+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d5b9424abe8c86d020f784d6d39f61c9a6260706...eda01069712ff831f9b93e952e54a976ce2e7148)

PR #9489, opened 2015-09-29
Fix the hunt activity to work with planes and helicopters

obrakmann:
(I'm not sure how planes will behave when `Wait()`ing. They'll probably just be suspended in mid-air)

### DIFF
--- a/OpenRA.Mods.Common/Activities/Hunt.cs
+++ b/OpenRA.Mods.Common/Activities/Hunt.cs
@@ -18,10 +18,12 @@ namespace OpenRA.Mods.Common.Activities
 {
 	public class Hunt : Activity
 	{
+		readonly IMove move;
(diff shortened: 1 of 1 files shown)


pchote:
I remember writing a special activity for making aircraft not suspend in mid air when doing stuff like this.  Lets please verify this before merging, and fix it if necessary.

obrakmann:
Hmm. I just made the yaks in soviet01 hunt after spawning, and they just RTB after the first kill. Not quite sure what happens there. I suspect their activity queue gets cleared at some point and then `ReturnOnIdle` triggers, but I haven't found yet where it happens.

(Later) Yep, removing `ReturnOnIdle` triggers the mid-air suspension.

Helicopters don't have `ReturnOnIdle`, but their activity queue stills gets cleared, so they still don't really hunt. Guess this isn't as easy as I thought.

PR #9491, opened 2015-09-30
Fix incorrect "In progress" time shown in server browser

cjshmyr:
This fixes a couple problems. The in progress time should be using `TotalMinutes`, and we should be treating the local machine's current time as UTC to match the master server.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/ServerBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/ServerBrowserLogic.cs
@@ -363,15 +363,13 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 			if (game.State == (int)ServerState.GameStarted)
 			{
-				var label = "";
-				try
+				var label = "In progress";
+
+				DateTime startTime;
+				if (DateTime.TryParse(game.Started, out startTime))
 				{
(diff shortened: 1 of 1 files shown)


obrakmann:
Looks ok and works for me. :+1:

RoosterDragon:
You might want to consider using `DateTime.TryParse` instead of `string.IsNullOrEmpty` to guard against invalid stuff sent by the server (since you've removed the try-catch).

atlimit8:
I agree.

cjshmyr:
I think that try/catch was spurious to begin with (it was there to catch the NRE I replaced it with), shouldn't we fix bad dates if there are on the master server instead?

The master server sets dates like using php's `date('Y-m-d H:i:s')` which seems okay, though I won't argue if I'm the minority here

atlimit8:
It is both better form to use `TryParse*` where parsing may fail and more efficient since parsing has to check for those cases anyways.

cjshmyr:
Updated.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/37b16d2ead6a0eca235cedb48ab9408cae2b072b...9c1e73d685038896a4ed72ced7b5d23b6c6c6110)

PR #9494, opened 2015-09-30
Add lua support for playing background music

### LINKED ISSUE #9473: Expose background music configration to Lua
This rounds out the functionality introduced by #8892. See also #9472.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27005807-expose-background-music-configration-to-lua?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

abcdefg30:
Fixes #9473.

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
@@ -20,7 +20,6 @@ using OpenRA.Graphics;
 using OpenRA.Mods.Common.Effects;
 using OpenRA.Mods.Common.Traits;
 using OpenRA.Scripting;
-using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Scripting
 {
--- a/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
@@ -95,6 +95,18 @@ namespace OpenRA.Mods.Common.Scripting
 				playlist.Play(musicInfo);
 		}
 
+		[Desc("Play track defined in music.yaml or map.yaml as background music." +
+			" If music is already playing use Media.StopMusic() to stop it" +
+			" and the background music will start automatically." +
+			" Keep the track empty to disable background music.")]
+		public void SetBackgroundMusic(string track = null)
+		{
+			if (!playlist.IsMusicAvailable)
+				return;
+
+			playlist.SetBackgroundMusic(string.IsNullOrEmpty(track) ? null : world.Map.Rules.Music[track]);
+		}
+
 		[Desc("Stop the current song.")]
 		public void StopMusic()
 		{
--- a/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
@@ -61,16 +61,8 @@ namespace OpenRA.Mods.Common.Scripting
 			if (!playlist.IsMusicAvailable)
 				return;
 
-			MusicInfo musicInfo;
-			if (string.IsNullOrEmpty(track))
-				musicInfo = playlist.GetNextSong();
-			else if (world.Map.Rules.Music.ContainsKey(track))
-				musicInfo = world.Map.Rules.Music[track];
-			else
-			{
-				Log.Write("lua", "Missing music track: " + track);
-				return;
-			}
+			var musicInfo = !string.IsNullOrEmpty(track) ? GetMusicTrack(track)
+				: playlist.GetNextSong();
 
 			if (func != null)
 			{
@@ -103,7 +95,17 @@ namespace OpenRA.Mods.Common.Scripting
 			if (!playlist.IsMusicAvailable)
 				return;
 
-			playlist.SetBackgroundMusic(string.IsNullOrEmpty(track) ? null : world.Map.Rules.Music[track]);
+			playlist.SetBackgroundMusic(string.IsNullOrEmpty(track) ? null : GetMusicTrack(track));
+		}
+
+		MusicInfo GetMusicTrack(string track)
+		{
+			var music = world.Map.Rules.Music;
+			if (music.ContainsKey(track))
+				return music[track];
+
+			Log.Write("lua", "Missing music track: " + track);
+			return null;
 		}
 
 		[Desc("Stop the current song.")]
--- a/OpenRA.Mods.Common/Traits/World/MusicPlaylist.cs
+++ b/OpenRA.Mods.Common/Traits/World/MusicPlaylist.cs
@@ -168,6 +168,14 @@ namespace OpenRA.Mods.Common.Traits
 			Game.Sound.PlayMusicThen(music, onComplete);
 		}
 
+		public void SetBackgroundMusic(MusicInfo music)
+		{
+			currentBackgroundSong = music;
+
+			if (CurrentSongIsBackground)
+				Stop();
+		}
+
 		public MusicInfo GetNextSong()
 		{
 			return GetSong(false);


pchote:
Can you please also include a way to cancel the background music (either as a separate function, or by passing a nil value)

abcdefg30:
Done.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -86,12 +86,28 @@ public void PlayMusic(string track = null, LuaFunction func = null)
...
 		[Desc("Stop the current song.")]
 		public void StopMusic()
 		{
 			playlist.Stop();
 		}
 
+		[Desc("Stop the current background music.")]
+		public void StopBackgroundMusic()
+		{
+			playlist.Play(playlist.GetNextSong());
```
pchote: You'll also need to reset `CurrentSongIsBackground`.

pchote: Actually hmm, something doesn't look right here...

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -86,12 +86,28 @@ public void PlayMusic(string track = null, LuaFunction func = null)
 				playlist.Play(musicInfo);
 		}
 
+		[Desc("Play track defined in music.yaml as background music.")]
+		public void PlayBackgroundMusic(string track)
+		{
+			if (!playlist.IsMusicAvailable)
+				return;
+
+			playlist.Play(world.Map.Rules.Music[track]);
```
pchote: I don't think this will work.  You need to set the `currentBackgroundSong` variable in MusicPlaylist in order to get the right behaviour.  I suggest adding a new SetBackgroundMusic method to MusicPlaylist to set that and CurrentSongAsBackground (and then change that back to private set).

abcdefg30: I tested this and it works.
As `Play(MusicInfo music)` sets `CurrentSongIsBackground` to `false`, I set it to `true` below.
`Play(MusicInfo music)` calls `Play()` and this uses
```
Game.Sound.PlayMusicThen(currentSong, () =>
{
	if (!CurrentSongIsBackground && !Game.Settings.Sound.Repeat)
		currentSong = GetNextSong();

	Play();
});
```
By then `CurrentSongIsBackground` is `true` (because of my line below) and the song is repeated.

abcdefg30:
Updated.

OpenRA.Mods.Common/Traits/World/MusicPlaylist.cs
```diff
@@ -164,6 +164,14 @@ public void Play(MusicInfo music, Action onComplete)
 			Game.Sound.PlayMusicThen(music, onComplete);
 		}
 
+		public void SetBackgroundMusic(MusicInfo music)
+		{
+			currentSong = currentBackgroundSong = music;
+			CurrentSongIsBackground = music != null;
+
+			Play();
```
pchote: As mentioned above, I don't think this should override the currently playing song.  If you want to force this behavior from a script, then you can instead do it by stopping the playing song.

abcdefg30: This seems a bit unintentional though...
To have to call `Media.Stop()` after `Media.PlayBackgroundMusic`.
I'll just add `playlist.Stop()` to `Media.PlayBackgroundMusic` with an extra bool parameter.

abcdefg30:
Updated.

abcdefg30:
Updated again as discussed on IRC.

abcdefg30:
Added checks for missing music tracks.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -86,6 +86,27 @@ public void PlayMusic(string track = null, LuaFunction func = null)
...
+			" Keep the track empty to disable background music.")]
+		public void SetBackgroundMusic(string track = null)
+		{
+			if (!playlist.IsMusicAvailable)
+				return;
+
+			playlist.SetBackgroundMusic(string.IsNullOrEmpty(track) ? null : CheckForTrackExistence(track));
+		}
+
+		MusicInfo CheckForTrackExistence(string track)
```
pchote: This name implies that it should return a `bool`.  Can you change this to `GetMusicTrack`?

abcdefg30:
Updated.

OpenRA.Mods.Common/Traits/World/MusicPlaylist.cs
```diff
@@ -164,6 +164,12 @@ public void Play(MusicInfo music, Action onComplete)
 			Game.Sound.PlayMusicThen(music, onComplete);
 		}
 
+		public void SetBackgroundMusic(MusicInfo music)
+		{
+			currentBackgroundSong = music;
+			CurrentSongIsBackground = music != null;
```
pchote: This line is related to the music playing, so shouldn't be set here.  Just to make this explicit (since the meaning of my earlier suggestions may have been lost in the confusion): SetBackgroundMusic should only have one line, which sets.  All the other state is managed by the (normal) music playback functions.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -86,6 +86,27 @@ public void PlayMusic(string track = null, LuaFunction func = null)
...
+				return;
+
+			playlist.SetBackgroundMusic(string.IsNullOrEmpty(track) ? null : GetMusicTrack(track));
+		}
+
+		MusicInfo GetMusicTrack(string track)
+		{
+			var music = world.Map.Rules.Music;
+			if (!music.ContainsKey(track))
+				throw new LuaException("Track '{0}' is not defined in music.yaml!".F(track));
```
pchote: Oh sorry, I just realized we can't do this.  This will make maps that use `PlayMusic` unusable for people who don't have the music installed.

pchote: #9488 makes `PlayMusic` generate a soft error (writing a message to lua.log) instead.

abcdefg30:
Updated.

Mailaender:
Needs a rebase.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -86,6 +86,28 @@ public void PlayMusic(string track = null, LuaFunction func = null)
 				playlist.Play(musicInfo);
 		}
 
+		[Desc("Play track defined in music.yaml as background music." +
```
atlimit8: "Play track defined in music.yaml or map.yaml as background music."

Now that #9488 has been merged, `map.yaml` can include music track(s) for that map.

abcdefg30:
Updated.

obrakmann:
Works as expected. :+1: 

Testcase:
```diff
diff --git a/mods/ra/maps/fort-lonestar/fort-lonestar.lua b/mods/ra/maps/fort-lonestar/fort-lonestar.lua
index 5001e0b..8a3c7b5 100644
--- a/mods/ra/maps/fort-lonestar/fort-lonestar.lua
+++ b/mods/ra/maps/fort-lonestar/fort-lonestar.lua
@@ -176,5 +176,7 @@ WorldLoaded = function()
 
        Media.DisplayMessage("Defend Fort Lonestar at all costs!")
 
+       Media.SetBackgroundMusic("rain")
+
        SendWave()
 end
diff --git a/mods/ra/maps/fort-lonestar/map.yaml b/mods/ra/maps/fort-lonestar/map.yaml
index 4544cc2..fcee04b 100644
--- a/mods/ra/maps/fort-lonestar/map.yaml
+++ b/mods/ra/maps/fort-lonestar/map.yaml
@@ -525,8 +525,6 @@ Rules:
                        Green: 0.85
                        Blue: 1.5
                        Ambient: 0.35
-               MusicPlaylist:
-                       BackgroundMusic: rain
                FlashPaletteEffect@LIGHTNINGSTRIKE:
                        Type: LightningStrike
                LuaScript:
```

Btw, another thing I dislike about using the rain as background music is that you actually need to set a non-zero music volume to get the benefit of this effect.

pchote:
Thanks :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/126b3c59e4285413e0dd29e3b2a8e9bc8fbe09c2

PR #9505, opened 2015-10-02
Makes hide-map and reveal-map crates affect allied players.

deniz1a:
This way, allied players will have the same shroud visibility.


abcdefg30:
So, do we want this now?

deniz1a:
yes

penev92:
I don't think so.

Phrohdoh:
What is the justification for changing our fallback default value, which may change thirdparty mods without the author knowing why?

I do not believe this is desired. :-1:

MustaphaTR:
:-1: I agree with @Phrohdoh

atlimit8:
:-1: - Closing due to unpopularity.

@deniz1a Try opening an issue or make your case on the IRC channel if you think you can make a good argument for this.

PR #9402, opened 2015-09-21
Add atreides02a

abcdefg30:

PR #9512, opened 2015-10-03
Add atreides02b 

abcdefg30:
Basically the same as #9402.


abcdefg30:
Updated with the fix from #9514.

mods/d2k/maps/atreides-02b/map.yaml
```diff
@@ -0,0 +1,212 @@
+MapFormat: 7
+
+RequiresMod: d2k
+
+Title: Atreides 02b
+
+Description: Infiltrate the Imperial Basin and build up our forces until they are strong enough to eradicate the local Harkonnen presence.\n\nThe Harkonnen are reinforcing their troops by air, so be on your guard. Use the Outpost's radar to detect attacks from unexpected quarters.\n\nBe careful when minig the Spice. Spice mounds grow out of the sand. While a vital source of Spice, Spice mounds can damage or destroy any unit that blunders into them.\n\nGood luck.\n
```
clemty: you wrote `minig`, it should be `mining` (in `atreides-01b` too)

abcdefg30: You probably mean `atreides-02a`, but yeah, thanks.

abcdefg30:
Fixed.

Biofreak1987:
The Harkonen wind traps and barracks are missing a concrete ground and thus just end up with 50% of their original health, same applies for the already merged atreides 02a.

abcdefg30:
Updated.

Mailaender:
Works :white_check_mark: however the Harkonnen are not making any attempt at defending their base and their radar is unpowered.

abcdefg30:
> and their radar is unpowered.

Are you sure? https://github.com/OpenRA/OpenRA/pull/9512/files#diff-31fc9236939b858cb9df8991a57f2fdfR200

abcdefg30:
> however the Harkonnen are not making any attempt at defending their base 

I think I'll add that and base repairing (also in atreides02a) in a follow up PR.
This one grew already big.

obrakmann:
Looks fine.

One thing I just noticed in all the d2k missions, though, is that the worm doesn't attack anything. This is due to a missing `Enemies:` entry for the `Creeps` player reference. Please change that, then :+1: 

Units also don't automatically open fire on sandworms. Again due to a missing `Enemies:` entry. I'm overall fine with that behaviour, but perhaps it would be better to stick an `AutoTargetIgnore` on the sandworm instead.

abcdefg30:
Updated.

mods/d2k/maps/atreides-02b/map.yaml
```diff
@@ -0,0 +1,213 @@
...
+			Prerequisites: ~disabled
+	outpost:
+		Buildable:
+			Prerequisites: barracks
+	outpostnopower:
+		Inherits: outpost
+		Buildable:
+			Prerequisites: ~disabled
+		Power:
+			Amount: 0
```
obrakmann: Oh, one other thing: the output is still powered down. Make this either `Amount: 100`, or use `-RequiresPower:` instead. Since there's no production happening, I'd go with the latter.

abcdefg30:
Fixed.

obrakmann:
Alright, now we're done here. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/1b3a9ed13cf2c8c61857cb12e6201ff8015fea39)

PR #9518, opened 2015-10-03
Tiberian Sun style right-click-and-drag scrolling implementation

JoppyFurr:
Rarr :3

```
This patch introduces support for the right-click-and-drag scrolling that
is available in Tiberian Sun and Red Alert 2. It can be enabled by
selecting "TS" scrolling in the Input settings.

A side-effect of this is that events previously tied to right clicks on
the world are now based on the release of the click rather than the press.

"Middle-Mouse Scrolling:" in the menu is also changed to "Scrolling Method:"
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -70,6 +70,7 @@ Also thanks to:
     * Jeff Harris (jeff_1amstudios)
     * Jes
     * Joakim Lindberg (booom3)
+    * Joppy Furr
     * Kanar
     * Kenny Hoxworth (hoxworth)
     * Krishnakanth Mallik
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -19,7 +19,7 @@ using OpenRA.Traits;
 
 namespace OpenRA
 {
-	public enum MouseScrollType { Disabled, Standard, Inverted }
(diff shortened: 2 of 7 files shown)


atlimit8:
Do you know how to use `git rebase -i bleed`? It would be nice if the style fixes were in the original commits instead of a follow-up commit.

JoppyFurr:
Okay, I'll give them a squash.

atlimit8:
I did not notice any problems with using mouse up/release instead.
:+1:

penev92:
This works great with both right-click and left-click orders!
But I get a feeling your changes to `WorldInteractionControllerWidget` should go inside `ViewportControllerWidget` because that already handles the other two types of scrolling.

Other than that this is awesome! Thanks!

JoppyFurr:
Okies - I've moved the changes into ViewportControllerWidget.

penev92:
Everything still works :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/98c6ec7668489446f1fc90a2b31d5a45ad629df0...0fcab595f8f0ea3a142880799259c6f4b8efd7e7)

PR #9521, opened 2015-10-03
Implement actor-producing support powers

pchote:
Implements the original spawning for fremen and saboteur in d2k, and the hunter-seeker spawning in TS (note: we don't yet support the proper H-S targeting, so it is left as a player-controlled unit).

### DIFF
--- a/mods/d2k/rules/structures.yaml
+++ b/mods/d2k/rules/structures.yaml
@@ -782,6 +782,12 @@ palace:
 	ProvidesPrerequisite@nuke:
 		Prerequisite: palace.nuke
 		Factions: harkonnen
+	ProvidesPrerequisite@fremen:
+		Prerequisite: palace.fremen
+		Factions: atreides
+	ProvidesPrerequisite@saboteur:
+		Prerequisite: palace.saboteur
+		Factions: ordos
 	NukePower:
 		Cursor: nuke
 		Icon: deathhand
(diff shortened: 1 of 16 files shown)


abcdefg30:
> OpenRA.Utility(1,1): Error: hunter.AircraftInfo.RepairBuildings: Missing actor `fix`.

pchote:
Fixed.

mods/d2k/rules/structures.yaml
```diff
@@ -800,6 +806,39 @@ palace:
 		DisplayRadarPing: True
 		CameraActor: camera
 		ActivationSequence:
+	ProduceActorPower@fremen:
+		Description: Recruit Fremen
+		LongDesc: Elite sniper infantry unit                       \n  Strong vs Infantry\n  Weak vs Vehicles\n  Special Ability: Invisibility
```
abcdefg30: This doesn't look right.
Same for the `LongDesc` below.

pchote: Ah yes - this is a bit of a hack.  The spaces forces the dialog wider so that the timer in the tooltip doesn't collide with the name.

pchote: That is a long standing bug that i'm not willing to fix in this pr.

abcdefg30:
Looks good to me, except some minor things (that we don't necessarily need to address here):
- The HunterSeeker doesn't display a explosion when detonating.
- We may want to allow to set a primary building, as at the moment the spawned actors will always come out of the first one built.

:+1: / :white_check_mark:

mods/d2k/rules/structures.yaml
abcdefg30: Are you sure that this issue exists?
I couldn't reproduce it.

pchote: I am

abcdefg30: spawnactorpowers

Mailaender:
I tried to verify that spy infiltration and sonar still works, but it seems to be completely busted. Easy to reproduce on http://resource.openra.net/maps/3367/ I haven't checked if the regression is already in bleed.

pchote:
@Mailaender: this shouldn't touch anything related to the sonar pulse.  That is a different power.

mods/d2k/rules/structures.yaml
abcdefg30: > 21:48 (+pchote) oh, I see 
21:48 (+pchote) abcdefg30: remove your hotkey bindings 
21:48 (+pchote) which aren't there by default 
21:48 (+pchote) they are adding the padding for you

abcdefg30: #9543

pchote:
Change H-S back to the vehicle target types.

mods/ts/rules/gdi-structures.yaml
```diff
@@ -375,6 +375,18 @@ GAPLUG:
...
 		CameraActor: camera
+	ProduceActorPower:
+		UpgradeTypes: plug.hunterseeker
+		UpgradeMinEnabledLevel: 1
+		Description: Hunter Seeker
+		LongDesc: Releases a drone that will acquire and destroy an enemy target.
+		Icon: hunterseeker
+		Actors: hunter
+		Type: HunterSeeker
+		ChargeTime: 720
```
obrakmann: I assume 12 minutes was really the original value? It's seems quite long.

mods/ts/rules/aircraft.yaml
```diff
@@ -287,3 +287,44 @@ APACHE:
...
+		Name: Hunter-Seeker Droid
+	Health:
+		HP: 500
+	Armor:
+		Type: Light
+	DemoTruck:
+	Explodes:
+		Weapon: SuicideBomb
+		EmptyWeapon: SuicideBomb
+	Aircraft:
```
obrakmann: The H-S doesn't reveal shroud, but can be moved into it. Don't know how the original handled it, but should it either have `RevealsShroud` or `MoveIntoShroud` set to false on `Aircraft`?

obrakmann:
Looks fine otherwise. :+1:

mods/ts/rules/aircraft.yaml
pchote: The H-S shouldn't be controllable at all, but rewriting our AutoTarget and DemoTruck logic is way out of scope for this PR.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/091ada86ced834cc9bd3b393f6b20a0529a05186)

PR #9523, opened 2015-10-03
Fix TS component towers

### LINKED ISSUE #9428: TS component towers are broken
Astor reported 2 issues with the TS component towers (discovered in his own mod):
As it turns out, 3 weeks ago 21186c10b6a37b72402229ca0ac03444eea27dd3 broke the component tower's make animation (possibly for other structures as well).
And 9da56f51e21064088f29e14d90839c27e10f240a broke their firing.

I was unable to find any other broken actors.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26935950-ts-component-towers-are-broken?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

penev92:
As the commit message suggests, `WithWallSpriteBody` was hijacking `WithSpriteBody`'s `DefaultAnimation` before/during the "make" animation, and the `Building` trait is counting on the make animation to finish and notify it that the building is complete. But that never happens, so the building is "stuck" with `Building.BuildComplete == false`, which then causes all sorts of trouble.

Fixes #9428.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
@@ -17,7 +17,7 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Render trait for actors that change sprites if neighbors with the same trait are present.")]
-	class WithWallSpriteBodyInfo : WithSpriteBodyInfo
+	class WithWallSpriteBodyInfo : WithSpriteBodyInfo, Requires<BuildingInfo>
 	{
 		public readonly string Type = "wall";
 
@@ -68,7 +68,7 @@ namespace OpenRA.Mods.Common.Traits
 		}
 	}
 
-	class WithWallSpriteBody : WithSpriteBody, INotifyAddedToWorld, INotifyRemovedFromWorld, ITick
+	class WithWallSpriteBody : WithSpriteBody, INotifyRemovedFromWorld, ITick
 	{
 		readonly WithWallSpriteBodyInfo wallInfo;
 		int adjacent = 0;
@@ -78,7 +78,6 @@ namespace OpenRA.Mods.Common.Traits
 			: base(init, info, () => 0)
 		{
 			wallInfo = info;
-			DefaultAnimation.PlayFetchIndex(NormalizeSequence(init.Self, Info.Sequence), () => adjacent);
 		}
 
 		public override void DamageStateChanged(Actor self, AttackInfo e)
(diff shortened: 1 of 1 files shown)


obrakmann:
Not the first time that the `MakeAnimation`<->`BuildComplete` dependency has caused bugs.

:+1:

OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
```diff
@@ -131,8 +133,26 @@ static void UpdateNeighbours(Actor self)
 
 		public void AddedToWorld(Actor self)
 		{
-			DefaultAnimation.PlayFetchIndex(NormalizeSequence(self, Info.Sequence), () => adjacent);
-			UpdateNeighbours(self);
+			var building = self.Trait<Building>();
```
pchote: `Requires<BuildingInfo>` too please.

OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
```diff
@@ -131,8 +133,26 @@ static void UpdateNeighbours(Actor self)
...
+				DefaultAnimation.PlayFetchIndex(NormalizeSequence(self, Info.Sequence), () => adjacent);
+				UpdateNeighbours(self);
+			}
+			else
+			{
+				// Wall adjacency rendering has lower priority.
+				// Wait for the "make" animation to finish first instead of overriding DefautAnimation.
+				self.QueueActivity(Util.SequenceActivities(
+					new WaitFor(() => building.BuildComplete),
+					new CallFunc(() =>
```
pchote: I believe we have a notification for when the make animation finishes - this should live there instead.

pchote: It's the same one that sets `building.BuildComplete` true.

penev92: Yes, the one one `WithMakeAnimation`. I can't introduce such an ugly dependency. :X

pchote: I'm talking about `INotifyBuildComplete`, which we use in a bunch of the other render traits.

penev92:
Well derp. This sure is simpler.

abcdefg30:

PR #9526, opened 2015-10-03
Rename TileShape to fit its role better

penev92:
This is the second small step toward "diamond" maps.
It just does some renaming to enable the addition of an _Isometric_ grid type.
Also `TileShape` isn't exactly proper, as it's used for much more than that.

### DIFF
--- a/OpenRA.Game/CPos.cs
+++ b/OpenRA.Game/CPos.cs
@@ -47,9 +47,9 @@ namespace OpenRA
 			return ToMPos(map.Grid.Type);
 		}
 
-		public MPos ToMPos(TileShape shape)
+		public MPos ToMPos(MapGridType gridType)
 		{
-			if (shape == TileShape.Rectangle)
+			if (gridType == MapGridType.Rectangle)
 				return new MPos(X, Y);
 
 			// Convert from diamond cell (x, y) position to rectangular map position (u, v)
--- a/OpenRA.Game/CPos.cs
+++ b/OpenRA.Game/CPos.cs
@@ -49,7 +49,7 @@ namespace OpenRA
 
 		public MPos ToMPos(MapGridType gridType)
 		{
-			if (gridType == MapGridType.Rectangle)
+			if (gridType == MapGridType.Rectangular)
 				return new MPos(X, Y);
 
 			// Convert from diamond cell (x, y) position to rectangular map position (u, v)
--- a/OpenRA.Game/CPos.cs
+++ b/OpenRA.Game/CPos.cs
@@ -52,7 +52,7 @@ namespace OpenRA
 			if (gridType == MapGridType.Rectangular)
 				return new MPos(X, Y);
 
-			// Convert from diamond cell (x, y) position to rectangular map position (u, v)
+			// Convert from RectangularIsometric cell (x, y) position to rectangular map position (u, v)
 			//  - The staggered rows make this fiddly (hint: draw a diagram!)
 			// (a) Consider the relationships:
 			//  - +1x (even -> odd) adds (0, 1) to (u, v)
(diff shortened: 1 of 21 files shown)


OpenRA.Game/CPos.cs
```diff
@@ -47,12 +47,12 @@ public MPos ToMPos(Map map)
...
 
-		public MPos ToMPos(TileShape shape)
+		public MPos ToMPos(MapGridType gridType)
 		{
-			if (shape == TileShape.Rectangle)
+			if (gridType == MapGridType.Rectangular)
 				return new MPos(X, Y);
 
-			// Convert from diamond cell (x, y) position to rectangular map position (u, v)
+			// Convert from isometric cell (x, y) position to rectangular map position (u, v)
```
pchote: RectangularIsometric.  You won't need to do this for DiagonalIsometric.

OpenRA.Game/Graphics/Viewport.cs
```diff
@@ -240,10 +240,9 @@ ProjectedCellRegion CalculateVisibleCells(bool insideBounds)
 			var tl = (PPos)map.CellContaining(worldRenderer.ProjectedPosition(TopLeft)).ToMPos(map);
 			var br = (PPos)map.CellContaining(worldRenderer.ProjectedPosition(BottomRight)).ToMPos(map);
 
-			// Diamond tile shapes don't have straight edges, and so we need
-			// an additional cell margin to include the cells that are half
-			// visible on each edge.
-			if (map.Grid.Type == TileShape.Diamond)
+			// Isometric tiles don't have straight edges, and so we need an additional
```
pchote: RectangularIsometric

OpenRA.Game/MPos.cs
```diff
@@ -41,12 +41,12 @@ public CPos ToCPos(Map map)
...
 
-		public CPos ToCPos(TileShape shape)
+		public CPos ToCPos(MapGridType gridType)
 		{
-			if (shape == TileShape.Rectangle)
+			if (gridType == MapGridType.Rectangular)
 				return new CPos(U, V);
 
-			// Convert from rectangular map position to diamond cell position
+			// Convert from rectangular map position to isometric cell position
```
pchote: I think you get the idea by now, so I'll stop here...

penev92:
Fixed the comments.

huwpascoe:
:+1: for better clarity this brings

atlimit8:
Needs rule upgrade in `OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs`.

OpenRA.Game/Map/MapGrid.cs
```diff
@@ -13,11 +13,11 @@
 
 namespace OpenRA
 {
-	public enum TileShape { Rectangle, Diamond }
+	public enum MapGridType { Rectangular, RectangularIsometric }
```
Phrohdoh: Do we have anything other than these two?
`RectangularIsometric` looks like a change for the sake of making a change.

atlimit8: I suspect that there is a plan to include `DiamondIsometric` which would be for diamond shaped isometric maps instead of rectangular (isometric) maps.

penev92: Exactly.

penev92:
Needs an upgrade rule for what? The upgrade rules don't touch `map.yaml` (which is the only thing this change affects), and they don't really need to.

atlimit8:
Well it looks like the rule upgrade system does not have a way to upgrade `MapGrid` or any other stuff in `mod.yaml`.
I'm OK with not adding an rule upgrade if this change is announced another way. `MapGrid` will have to be explained as well since these changes are not automatically applied.

obrakmann:
:+1: @penev92, please rebase so I can merge this.

penev92:
Oh, fun. I have to track down and fix all the stuff #9744 broke first.

penev92:
Updated

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/e662cb58d69423e98c11bb9893ad2a7335b77d5f...b36dfe1fd23a6f386951a285b64910dd0242f321)

PR #9533, opened 2015-10-04
Copy palette.bin from disk and check it exists.

pchote:
This should fix the issue mentioned in #9483.  It will need to be tested by someone with the disk before we can merge.


penev92:
Fixes it :+1:

PR #9536, opened 2015-10-04
Experience added to Player Actor, Statistics

Phoib:
In order to eventually facilitate Support Powers as featured in Generals, I added the Trait PlayerExperience. This was inspired by the Pull Request #6221.

Experience is as of yet based on cost.

### DIFF
--- a/OpenRA.Game/Traits/Player/PlayerExperience.cs
+++ b/OpenRA.Game/Traits/Player/PlayerExperience.cs
@@ -1,0 +1,35 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+
+namespace OpenRA.Traits
+{
+	public class PlayerExperienceInfo : ITraitInfo
+	{
+		public object Create(ActorInitializer init) { return new PlayerExperience(init.Self, this); }
+	}
+
+	public class PlayerExperience : ISync
+	{
+		[Sync] public int Experience { get; private set; }
+
+		public PlayerExperience(Actor self, PlayerExperienceInfo info)
+		{
+			Experience = 0;
+		}
+
+		public void GivePlayerExperience(int num)
+		{
+			Experience += num;
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Traits/GivesExperience.cs
+++ b/OpenRA.Mods.Common/Traits/GivesExperience.cs
@@ -41,14 +41,16 @@ namespace OpenRA.Mods.Common.Traits
 
 			var valued = self.Info.TraitInfoOrDefault<ValuedInfo>();
 
-			// Default experience is 100 times our value
 			var exp = info.Experience >= 0
 				? info.Experience
-				: valued != null ? valued.Cost * 100 : 0;
+				: valued != null ? valued.Cost : 0;
 
 			var killer = e.Attacker.TraitOrDefault<GainsExperience>();
 			if (killer != null)
 				killer.GiveExperience(exp);
+
+			if (info.Experience > 0)
+				e.Attacker.Owner.PlayerActor.Trait<PlayerExperience>().GivePlayerExperience(info.Experience);
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Traits/Player/PlayerStatistics.cs
+++ b/OpenRA.Mods.Common/Traits/Player/PlayerStatistics.cs
@@ -47,6 +47,8 @@ namespace OpenRA.Mods.Common.Traits
 		public int BuildingsKilled;
 		public int BuildingsDead;
 
+		public int ExperienceGained;
+
 		public PlayerStatistics(Actor self)
 		{
 			world = self.World;
@@ -124,6 +126,10 @@ namespace OpenRA.Mods.Common.Traits
 				attackerStats.KillsCost += cost;
 				defenderStats.DeathsCost += cost;
 			}
+
+			var experienceInfo = self.Info.TraitInfo<GivesExperienceInfo>();
+			if (experienceInfo != null && experienceInfo.Experience > 0)
+				attackerStats.ExperienceGained += experienceInfo.Experience;
 		}
 	}
 }
--- a/mods/cnc/rules/aircraft.yaml
+++ b/mods/cnc/rules/aircraft.yaml
@@ -2,6 +2,8 @@ TRAN:
 	Inherits: ^Helicopter
 	Valued:
 		Cost: 750
+    GivesExperience:
+        Experience: 750
 	Tooltip:
 		Name: Chinook Transport
 		Description: Fast Infantry Transport Helicopter.\n  Unarmed
@@ -45,6 +47,8 @@ HELI:
 	Inherits: ^Helicopter
 	Valued:
 		Cost: 1200
+    GivesExperience:
+        Experience: 1200
 	Tooltip:
 		Name: Apache Longbow
 		Description: Helicopter Gunship with Chainguns.\n  Strong vs Infantry, Light Vehicles and\n  Aircraft\n  Weak vs Tanks
@@ -96,6 +100,8 @@ ORCA:
 	Inherits: ^Helicopter
 	Valued:
 		Cost: 1200
+    GivesExperience:
+        Experience: 1200
 	Tooltip:
 		Name: Orca
 		Description: Helicopter Gunship with AG Missiles.\n  Strong vs Buildings, Tanks\n  Weak vs Infantry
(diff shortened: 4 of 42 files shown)


OpenRA.Mods.Common/Traits/Player/PlayerStatistics.cs
```diff
@@ -124,6 +126,12 @@ public void Killed(Actor self, AttackInfo e)
 				attackerStats.KillsCost += cost;
 				defenderStats.DeathsCost += cost;
 			}
+
+			if (self.Info.HasTraitInfo<GivesExperienceInfo>() && self.Info.TraitInfo<GivesExperienceInfo>().GivesPlayerExperience)
+		        {
```
abcdefg30: One level of indentation less here, please.
```
			if (self.Info.HasTraitInfo<GivesExperienceInfo>() && self.Info.TraitInfo<GivesExperienceInfo>().GivesPlayerExperience)
			{
				var exp = self.Info.TraitInfo<ValuedInfo>().Cost;
				attackerStats.ExperienceGained += exp;
			}

File: OpenRA.Mods.Common/Traits/Player/PlayerStatistics.cs [hunk 2]
```diff
@@ -124,6 +126,12 @@ public void Killed(Actor self, AttackInfo e)
 				attackerStats.KillsCost += cost;
 				defenderStats.DeathsCost += cost;
 			}
+
+			if (self.Info.HasTraitInfo<GivesExperienceInfo>() && self.Info.TraitInfo<GivesExperienceInfo>().GivesPlayerExperience)
```
Mailaender: Please cache the trait lookup with `self.Info.TraitInfo<GivesExperienceInfo>()` and a `!= null` check to makes this less wasteful regarding performance.

OpenRA.Mods.Common/Traits/GivesExperience.cs
```diff
@@ -49,6 +52,11 @@ public void Killed(Actor self, AttackInfo e)
 			var killer = e.Attacker.TraitOrDefault<GainsExperience>();
 			if (killer != null)
 				killer.GiveExperience(exp);
+
+			if (info.GivesPlayerExperience)
+			{
+				e.Attacker.Owner.PlayerActor.Trait<PlayerExperience>().GivePlayerExperience(exp / 100);
```
Mailaender: Why the `/ 100` here?

Phoib: // Default experience is 100 times our value

OpenRA.Mods.Common/Traits/Player/PlayerStatistics.cs
```diff
@@ -124,6 +126,13 @@ public void Killed(Actor self, AttackInfo e)
 				attackerStats.KillsCost += cost;
 				defenderStats.DeathsCost += cost;
 			}
+
+			var experienceInfo = self.Info.TraitInfo<GivesExperienceInfo>();
+			if (experienceInfo != null && experienceInfo.GivesPlayerExperience)
+		    {
+				var exp = self.Info.TraitInfo<ValuedInfo>().Cost;
```
Mailaender: Line https://github.com/OpenRA/OpenRA/pull/9536/files#diff-d10fad95a9ee072ed5c227a38d1e35ffR125 already does the valued info lookup so you could move the local variable up and re-use that as well for bonus points.

OpenRA.Mods.Common/Traits/GivesExperience.cs
pchote: I think it might make more sense to have this configurable, with a default value of 0.  This then also covers the `GivesPlayerExperience` parameter.

Phoib: Reuse the already declared Experience? (Has a default value of -1, but the same idea applies)

OpenRA.Mods.Common/Traits/GivesExperience.cs
```diff
@@ -21,6 +21,9 @@ class GivesExperienceInfo : ITraitInfo
 		[Desc("Grant experience for team-kills.")]
 		public readonly bool FriendlyFire = false;
 
+		[Desc("Grant player experience for kills.")]
+		public readonly bool GivesPlayerExperience = false;
```
GraionDilach: What is the reason behind the default value of this?

OpenRA.Mods.Common/Traits/Player/PlayerStatistics.cs
Phoib: But that line is under a HasTraitInfo check. 

I'll take Paul's comment about experience into account though, so it will be a newly declared variable anyways.

OpenRA.Mods.Common/Traits/GivesExperience.cs
Phoib: My impression was non combat objects outnumber combat objects, hence making it false by default.

GraionDilach: Noncombat objects don't even have this trait IMO.

Phoib: Replacing it with an Experience > 0 check anyways.

Phoib:
*cough* added a new commit, the Experience label is now used. This supersedes the GivesPlayerExperience.

OpenRA.Game/Traits/Player/PlayerExperience.cs
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	public class PlayerExperience : ISync
+	{
+		public PlayerExperience(Actor self, PlayerExperienceInfo info)
+		{
+			Experience = 0;
+		}
+
+		[Sync] public int Experience;
```
pchote: Style nit: This should be `int Experience { get; private set; }` to prevent outside code from modifying the value directly.  Please also move this above the constructor.

mods/cnc/rules/aircraft.yaml
```diff
@@ -2,6 +2,8 @@ TRAN:
 	Inherits: ^Helicopter
 	Valued:
 		Cost: 750
+    GivesExperience:
+        Experience: 750
```
pchote: Did you mean for all of these to be x 100 to balance out the change in the trait?

pchote: Nope, you seem to have handled that in the GainsExperience trait instead. What's the motivation for all these additions?

OpenRA.Mods.Common/Traits/GivesExperience.cs
```diff
@@ -41,14 +41,18 @@ public void Killed(Actor self, AttackInfo e)
...
 				? info.Experience
-				: valued != null ? valued.Cost * 100 : 0;
+				: valued != null ? valued.Cost : 0;
 
 			var killer = e.Attacker.TraitOrDefault<GainsExperience>();
 			if (killer != null)
 				killer.GiveExperience(exp);
+
+			if (info.Experience > 0)
+			{
```
pchote: Note that our code style is to omit braces around single-line blocks.  Please remove them from here and the addition to PlayerStatistics.

mods/cnc/rules/aircraft.yaml
Phoib: Keeping the traits separate, and these are placeholder numbers. 

They might want to be changed, I just put them as the same value as the cost, (except for the British spy)

Phoib:
Should I rebase to a single commit?

penev92:
Yes please. Also please be patient, we are working hard to get the release out now :)

Phoib:
Patience for people who work on this in their free time, without compensation? Outrageous! 

I mean, sure, 3 simple edits took me 4 weeks as well ;-)

mods/cnc/rules/aircraft.yaml
GraionDilach: But these are still the defaults - why respecify them then?

abcdefg30:
We still need to resolve https://github.com/OpenRA/OpenRA/pull/9536#discussion-diff-41701588.

I also think that the experience should be shown in the statistics tab.
Not (only) when you open the ingame menu.

OpenRA.Mods.Common/Traits/GivesExperience.cs
```diff
@@ -41,14 +41,16 @@ public void Killed(Actor self, AttackInfo e)
...
 			var exp = info.Experience >= 0
 				? info.Experience
-				: valued != null ? valued.Cost * 100 : 0;
+				: valued != null ? valued.Cost : 0;
 
 			var killer = e.Attacker.TraitOrDefault<GainsExperience>();
 			if (killer != null)
 				killer.GiveExperience(exp);
+
+			if (info.Experience > 0)
```
GraionDilach: Oh, so this is why you duplicated all the experience values!  This should be `exp > 0` though.

OpenRA.Mods.Common/Traits/GivesExperience.cs
```diff
@@ -41,14 +41,16 @@ public void Killed(Actor self, AttackInfo e)
...
+				e.Attacker.Owner.PlayerActor.Trait<PlayerExperience>().GivePlayerExperience(info.Experience);
```
GraionDilach: Also this should use `TraitOrDefault` and have a nullcheck.

pchote:

Phoib:
Fixed the comment from abcdefg30, but cannot build it on my own, #10248 is blocking this.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/10358.

PR #9537, opened 2015-10-04
Fix crash when launching the game directly into a mod with missing assets

penev92:
From the commit description:

> This fixes a crash that happens when you try to launch a mod directly while its assets aren't installed.
> Additionally should reduce the overhead for dedicated servers (in theory).

This only moves things around.
- The asset validation check was moved to an earlier point, because on bleed the game crashes because it tries to load the assets before checking whether they are available.
- The asset validation check now uses `File.Exists()` instead of `GlobalFileSystem.Exists()` because at that point in time the files aren't loaded yet.
- All asset loading and initializing is moved _after_ the dedicated server check. In theory that should save some unnecessary loading for those.

**Note:** I am unsure whether removing this from the dedicated server loading will have any unwanted effects, so someone more familiar with that should take a look:

```
                PerfHistory.Items["render"].HasNormalTick = false;
                PerfHistory.Items["batches"].HasNormalTick = false;
                PerfHistory.Items["render_widgets"].HasNormalTick = false;
                PerfHistory.Items["render_flip"].HasNormalTick = false;

                JoinLocal();
```

I did try setting up a dedicated server and joining it (jaZz_KCS set it up) and it all worked fine.

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -281,13 +281,29 @@ namespace OpenRA
 			Sound.Initialize();
 
 			ModData = new ModData(mod, !Settings.Server.Dedicated);
-			ModData.MountFiles();
-			ModData.InitializeLoaders();
-			if (!Settings.Server.Dedicated)
-				Renderer.InitializeFonts(ModData.Manifest);
 
 			using (new PerfTimer("LoadMaps"))
 				ModData.MapCache.LoadMaps();
+
+			if (Settings.Server.Dedicated)
+			{
+				RunDedicatedServer();
+				Environment.Exit(0);
+			}
+
+			var installData = ModData.Manifest.Get<ContentInstaller>();
+			var isModContentInstalled = installData.TestFiles.All(f => File.Exists(Platform.ResolvePath(f)));
+
+			// Mod assets are missing!
+			if (!isModContentInstalled)
+			{
+				InitializeMod("modchooser", new Arguments());
+				return;
+			}
+
+			ModData.MountFiles();
+			ModData.InitializeLoaders();
+			Renderer.InitializeFonts(ModData.Manifest);
 
 			if (Cursor != null)
 				Cursor.Dispose();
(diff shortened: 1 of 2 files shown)


Mailaender:
Dedicated server still runs and the crash is gone. :white_check_mark:

steelphase:
:+1: this does resolve the crash at launch with missing assets

OpenRA.Game/Game.cs
```diff
@@ -352,7 +317,55 @@ public static void InitializeMod(string mod, Arguments args)
...
+				// Mod assets are missing!
+				if (!isModContentInstalled)
+				{
+					InitializeMod("modchooser", new Arguments());
+					return;
+				}
+
+				ModData.MountFiles();
+				ModData.InitializeLoaders();
+				if (!Settings.Server.Dedicated)
```
pchote: This is now guaranteed to be true, so you don't need the if.

OpenRA.Game/Game.cs
```diff
@@ -352,7 +317,55 @@ public static void InitializeMod(string mod, Arguments args)
...
+					Cursor = new SoftwareCursor(ModData.CursorProvider);
+
+				PerfHistory.Items["render"].HasNormalTick = false;
+				PerfHistory.Items["batches"].HasNormalTick = false;
+				PerfHistory.Items["render_widgets"].HasNormalTick = false;
+				PerfHistory.Items["render_flip"].HasNormalTick = false;
+
+				JoinLocal();
+
 				ModData.LoadScreen.StartGame(args);
```
pchote: Managing mod assets was the whole reason to have this in the (mod-code) loadscreen.  I'm not sure it's worth keeping this now that it has all been moved to the mod chooser or engine code.

penev92: What do you suggest?

OpenRA.Game/Game.cs
```diff
@@ -352,7 +317,54 @@ public static void InitializeMod(string mod, Arguments args)
 				Environment.Exit(0);
 			}
 			else
+			{
```
pchote: If Environment.Exit works the way I assume it does (terminating the application) then you could keep the following in the main flow instead of inside the else.  That would minimise the future change when the dedicated server code is pulled into its own executable.

OpenRA.Game/Game.cs
pchote: If it doesn't cause follow on issues (I'm assuming that all our load screens inherit that implementation) I suggest moving the rest of the BlankLoadScreen.StartGame code here too, and then remove that method from the load screen interface.

penev92: `ModChooserLoadScreen` has it's own implementation that opens one window or another depending on whether it's going to install music or not.

penev92:
Beautified some more.

Mailaender:
Still works. Thanks for fixing this yourself.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/29921d4f580101406cc9baa5a1c12c1c3437bfcb...b5e73cd8b7bb4bfa6b9ef3d1a9033dcfc17f81dc)

PR #9539, opened 2015-10-04
Prepare for lobby layout overhaul.

pchote:
This makes some simple changes to the lobby layout to make it much more maintainable.
- Split out the different top panels into their own files.
- Removed the special case handling of the player bin header.
- Automatically size the bin contents instead of hardcoding everything.

There should be no visible changes ingame, and I suggest using a diff tool to review - the real changes are only a handful of lines.


pchote:
I recommend reviewing by commit, where the changes (a few lines each commit) aren't mixed in with the rename.  That saves you from having to wade through thousands of lines of identical content.

abcdefg30:
:+1: / :white_check_mark:

Mailaender:
Notifying @Phrohdoh as this change will break his RA2 mod during rebase and there is no automatic upgrade rule to fix it.

PR #9546, opened 2015-10-05
Add panel-transparent to TD

pchote:
This adds a new panel type to TD in preparation for the lobby changes, and saves a bit of performance by not tiling transparent sprites over the screen in the shellmap/ingame menu.


Mailaender:
Still works. :white_check_mark: What is the panel-transport for?

pchote:
> What is the panel-transport for?

The new music panel, to avoid the dark panels from stacking:

PR #9550, opened 2015-10-05
Add gdi06

clemty:
map.png is currently too bright, otherwise the mission should be playable
I disabled autotargeting in a different way than it was done for Tanya in allies03 - it seems to work though


pchote:
It might change the feel a bit much, but I think it would be cool if we changed this to a stormy night-time mission (like Fort Lonestar in RA) to match the FMV and show off our cool new weather effects.

mods/cnc/maps/gdi06/gdi06.lua
```diff
@@ -0,0 +1,159 @@
...
+		Reinforcements.ReinforceWithTransport(player, 'tran', nil, { lstStart.Location, TransportRally.Location }, nil, nil, nil)
+	end)
+
+	Trigger.OnKilled(CivFleeTrigger, function()
+		Civilian.Patrol(CivFleeRoute, false)
+	end)
+
+	Trigger.OnEnteredFootprint({ CivFleeRoute3.Location }, function(a, id)
+		if a.Owner == civilian then
+			HuntTriggerFunction()
```
pchote: Ouch, this is pretty brutal in-game.  They all make a bee-line straight for the commando, wherever he is (was is like this in the original game)?

pchote:
It would be nice if you could add an easy difficulty level that applies an upgrade to the commando that enables an armor bonus (since we can do that now).

mods/cnc/maps/gdi06/gdi06.lua
clemty: In the original mission: Commando shoots a guard in the village which can be easily avoided. Civilian next to him runs off and stops a bit outside the village.

here: Commando shoots easily avoided guard; civilian runs off all the way to the north-west, then east until he reaches the Hand of Nod and informs Nod forces - plenty of time for the Commando to neutralize the civilian

maybe make the cat&mouse game more fun: instead of Hunt(), send them (AttackMove) to the location where the commando was when he killed the unit? That way Commando & Chinook could lure forces away from the base

abcdefg30: > maybe make the cat&mouse game more fun: instead of Hunt(), send them (AttackMove) to the location where the commando was when he killed the unit? That way Commando & Chinook could lure forces away from the base

Seems like a good idea.

clemty:
Updated:
- [x] Weather effects (from Fort Lonestar): night time mood, rain (no lightning, no thunder)
- [x] 3 difficulty levels:
 - easy: Commando will return fire when attacked, will self heal like Mammoth Tank
 - normal: Commando will return fire when attacked
 - hard: Commando will not return fire when attacked

todo:
- [ ] Replace Hunt() with AttackMove()
- [ ] Fix map preview
- [ ] Add more Nod bazooka soldiers so the player can't just fly to the north edge of the map using the Chinook

It'd be good if we could agree on a set of traits for the Commando in Easy/Normal/Hard so the missions having Commandos are consistent

clemty:
updated (current status above)

mods/cnc/maps/gdi06/gdi06.lua
```diff
@@ -0,0 +1,239 @@
+CommandoReinforcements = { "rmbo" }
+IslandSamSites = { SAM01, SAM02 }
+NodBase = { PowerPlant1, PowerPlant2, PowerPlant3, PowerPlant4, PowerPlant5, Refinery, HandOfNod, Silo1, Silo2, Silo3, Silo4, ConYard, CommCenter}
```
abcdefg30: Add a space before the closing brace, please.

mods/cnc/maps/gdi06/gdi06.lua
```diff
@@ -0,0 +1,239 @@
...
+			Utils.Do(AttackUnits, function(unit)
+				unit.AttackMove(waypoint10.Location)
+			end)
+			Trigger.RemoveFootprintTrigger(id)
+		end
+	end)
+
+	Utils.Do(HuntTriggerActivator, function(unit)
+		Trigger.OnDamaged(unit, function()
+			HuntTriggerFunction()
```
abcdefg30: `Trigger.OnDamaged(unit, HuntTriggerFunction)`

mods/cnc/maps/gdi06/gdi06.lua
```diff
@@ -0,0 +1,239 @@
...
+
+	Utils.Do(NodBase, function(structure)
+		Trigger.OnKilled(structure, function()
+			player.MarkCompletedObjective(destroyObjective)
+		end)
+	end)
+
+	Trigger.OnAllKilled(IslandSamSites, function()
+		TransportFlare = Actor.Create('flare', true, { Owner = player, Location = Flare.Location })
+		Reinforcements.ReinforceWithTransport(player, 'tran', nil, { lstStart.Location, TransportRally.Location }, nil, nil, nil)
```
abcdefg30: You don't need to set the last three parameters to `nil`, they already default to `nil`.

mods/cnc/maps/gdi06/gdi06.lua
```diff
@@ -0,0 +1,239 @@
...
+		end
+	end)
+
+	Utils.Do(HuntTriggerActivator, function(unit)
+		Trigger.OnDamaged(unit, function()
+			HuntTriggerFunction()
+		end)
+	end)
+
+	Trigger.AfterDelay(5,NodKillCounter)
```
abcdefg30: Add a space after the `,`, please.

mods/cnc/maps/gdi06/gdi06.lua
```diff
@@ -0,0 +1,239 @@
...
+	end)
+
+	transport.Wait(5)
+	transport.ScriptedMove(transportStart)
+	transport.Destroy()
+end
+
+NodKillCounter = function()
+	local enemyUnits
+	enemyUnits = enemy.GetGroundAttackers()
```
abcdefg30: `local enemyUnits = enemy.GetGroundAttackers()`

mods/cnc/maps/gdi06/gdi06.lua
```diff
@@ -0,0 +1,239 @@
...
+	local list = enemy.GetGroundAttackers()
+	Utils.Do(list, function(unit)
+		IdleHunt(unit)
+	end)
+end
+
+IdleHunt = function(unit)
+	if not unit.IsDead then
+		Trigger.OnIdle(unit, function()
+			unit.Hunt()
```
abcdefg30: `Trigger.OnIdle(unit, unit.Hunt)`

mods/cnc/maps/gdi06/map.yaml
```diff
@@ -0,0 +1,1168 @@
...
+	Crates: False
+	Creeps: False
+	Fog: True
+	Shroud: True
+	AllyBuildRadius: False
+	FragileAlliances: False
+	StartingCash: 0
+	ConfigurableStartingUnits: False
+	Difficulties: Easy, Normal, Hard
+	ShortGame: True
```
abcdefg30: Why not `false`?

mods/cnc/maps/gdi06/map.yaml
```diff
@@ -0,0 +1,1168 @@
...
+		Inherits: LST
+		-WithRoof:
+		-Selectable:
+		RejectsOrders:
+	SAM:
+		SpawnActorOnDeath:
+			Actor: e1
+	RMBO:
+		MustBeDestroyed:
+			RequiredForShortGame: true
```
abcdefg30: As then you can also drop this.

mods/cnc/maps/gdi06/map.yaml
```diff
@@ -0,0 +1,1168 @@
...
+		AutoTarget:
+			TargetWhenIdle: false
+			TargetWhenDamaged: true
+		Health:
+			HP: 150
+	RMBO.easy:
+		Inherits: RMBO
+		AutoTarget:
+			TargetWhenIdle: false
```
abcdefg30: This is the same as on `RMBO`, intentional?

abcdefg30:

mods/cnc/maps/gdi06/map.yaml
clemty: the ShortGame settings are used so that the mission ends as soon as the RMBO dies (even if the Chinook is still alive) - without it, the mission would continue with the Chinook

abcdefg30: You could use `Trigger.OnKilled(RMBO...`.

clemty: can I do that without getting the actual actor via GetActorsByType before? RMBO is just a unit type, and not a specific unit, so I get something like
```System.NullReferenceException: Object reference not set to an instance of an object.``` (when trying to use RMBO)
or
```Unable to convert parameter 0 to Actor``` (when trying to use CommandoType)

Creating the Actor via the Landing Craft doesn't return handle for the unit either

Comment (abcdefg30) [inline mods/cnc/maps/gdi06/map.yaml]: Yeah you need to actor, sorry for being unclear.
The actor is created [here](https://github.com/OpenRA/OpenRA/pull/9550/files#diff-59f77a61dfe126788e68a1fe2c353166R190).

Comment (abcdefg30) [inline mods/cnc/maps/gdi06/map.yaml]: The change would be indeed a bit comlicated, just leave it as is at the moment then.

File: mods/cnc/maps/gdi06/map.yaml (hunk 10 above)
Comment (clemty) [inline mods/cnc/maps/gdi06/map.yaml]: I *think* that was intentional to make it clear what the differences between easy/normal/hard are, but duplicating all the values does not increase readability so I'll remove them

File: mods/cnc/maps/gdi06/gdi06.lua [hunk 11]
```diff
@@ -0,0 +1,233 @@
...
+		KillCounterHuntThreshold = 30
+	elseif Map.Difficulty == "Hard" then
+		CommandoType = "rmbo.hard"
+		KillCounterHuntThreshold = 15
+	else
+		CommandoType = "rmbo"
+		KillCounterHuntThreshold = 20
+	end
+
+	Media.PlayMusic("rain")
```
Mailaender: Use https://github.com/OpenRA/OpenRA/wiki/Traits#musicplaylist `StartingMusic` instead. This seems to otherwise break `BackgroundMusic`.

mods/cnc/maps/gdi06/gdi06.lua
```diff
@@ -0,0 +1,233 @@
...
+		end)
+	end)
+
+	Trigger.OnAllKilled(IslandSamSites, function()
+		TransportFlare = Actor.Create('flare', true, { Owner = player, Location = Flare.Location })
+		Reinforcements.ReinforceWithTransport(player, 'tran', nil, { lstStart.Location, TransportRally.Location })
+	end)
+
+	Trigger.OnKilled(CivFleeTrigger, function()
+		Civilian.Move(CivHideOut.Location)
```
Mailaender:
```
Fatal Lua Error: [string "BindingSupport.lua"]:30: Actor 'c8 (dead)' does not define a property 'Move'
```

when I kill the civilian before the guard soldier.

clemty: good catch!

mods/cnc/maps/gdi06/gdi06.lua
clemty: odd - I was told to manually start the music that way on IRC. it doesn't work for me either way (see #9577), does it work for anyone now after I change it?

mods/cnc/maps/gdi06/map.yaml
```diff
@@ -0,0 +1,1166 @@
...
+		-SpawnMPUnits:
+		-MPStartLocations:
+		-CrateSpawner:
+		LuaScript:
+			Scripts: gdi06.lua
+		ObjectivesPanel:
+			PanelName: MISSION_OBJECTIVES
+		MusicPlaylist:
+			BackgroundMusic: ambientrain
+			StartingMusic: rain
```
GraionDilach: Uhm... this looks fishy. Are you sure you need both?

clemty: I added BackgroundMusic after Mailänder's comment (see above); the trait documentation says nothing about the *Music traits being mutually exclusive.
The idea is to play the Rain in the Night track but have the ambient music in case the track is stopped - does it work for you?

abcdefg30: That way `BackgroundMusic` won't be set ever.
So no wonder that you can't hear anything after stopping the current track.

Mailaender: Hmm, try

```yaml
AmbientSound:
	SoundFile: ambientrain.aud
```

instead of the broken by design `BackgroundMusic` then.

pchote: Please don't do that. If there is a bug in that code then the productive thing to do is to report it so that somebody can fix it.

pchote: I chased @abcdefg30's comment back to the source of the problem, and filed #9666.

mods/cnc/audio/music.yaml
```diff
@@ -66,3 +66,7 @@ nod_map1: Nod Map Theme
 	Hidden: true
 nod_win1: Nod Win Theme
 	Hidden: true
+
+# Special effects
```
pchote: #9488 lets you move these into the map file itself, so adding [has dependencies] for that and #9666.

Mailaender: Dependencies have been merged. Needs a rebase to make use of them.

clemty:
updated & rebased

Mailaender:
This is so awesome with the rain effects! :umbrella:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c912668d65322b882a398cab05ad1a732b5a2b08...f1fd62af3e567a17c8981650bb3ba31819bd30fd)

Issue #9561, opened 2015-10-06
Soviet 07


r34ch:
Even after freeing the dogs and engineers, if your rifle soldiers from the start of the mission die, you loose the mission.

Probably makes more sense to end the mission if all your engineers die before reaching the cooling stations, rewiring the security or the actual reactor.

On bleed.

PR #9562, opened 2015-10-06
Fix early game over in soviet07

ghost:
- Fix https://github.com/OpenRA/OpenRA/issues/9561
- Add a missing healcrate

### DIFF
--- a/mods/ra/maps/soviet-07/soviet07.lua
+++ b/mods/ra/maps/soviet-07/soviet07.lua
@@ -238,8 +238,12 @@ IntroSequence = function()
 		Media.PlaySpeechNotification(player, "TimerStarted")
 		timerStarted = true
 	end)
+
+	-- Trigger a game over if the player lost all human units before the security system has been deactivated
 	Trigger.OnAllKilled(StartingUnits, function()
-		enemy.MarkCompletedObjective(alliedObjective)
+		if not controlCenterTrigger then
+			enemy.MarkCompletedObjective(alliedObjective)
+		end
 	end)
 end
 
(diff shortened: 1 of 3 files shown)


mods/ra/maps/soviet-07/soviet07.lua
```diff
@@ -239,7 +239,9 @@ IntroSequence = function()
 		timerStarted = true
 	end)
 	Trigger.OnAllKilled(StartingUnits, function()
```
abcdefg30: Couldn't we just drop this entirely?
As you have a `player.HasNoRequiredUnits()` check in `Tick`.

Biofreak1987: The point of it was that you need a human unit to interact with the security system and can't use an animal for that job. I'll remove that Trigger if the majority wants it gone.

abcdefg30:
:+1:

mods/ra/maps/soviet-07/soviet07.lua
abcdefg30: Didn't think about that, good point.

pchote: Please add a comment to the code explaining this.

Biofreak1987:
Added a comment for the early game over and an OpenRA.sln entry for the lua script as pointed out by clemty

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/8be49ecdf28b5ee2d1a6b27de4980129561153d5)

PR #9578, opened 2015-10-09
Add a CanDeployOnRamps property to DeployToUpgrade

penev92:
So stuff like the Mobile Sensor Array can't deploy on ramps (slopes). (As seen in @SoScared's recent TS stream).

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
@@ -28,6 +28,9 @@ namespace OpenRA.Mods.Common.Traits
 			"Leave empty to allow any.")]
 		public readonly HashSet<string> AllowedTerrainTypes = new HashSet<string>();
 
+		[Desc("Can this actor deploy on slopes?")]
+		public readonly bool CanDeployOnRamps = false;
+
 		[Desc("Cursor to display when able to (un)deploy the actor.")]
 		public readonly string DeployCursor = "deploy";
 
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -143,6 +146,11 @@ public void ResolveOrder(Actor self, Order order)
 
 		bool OnValidTerrain()
 		{
+			return CheckTerrain() && CheckRamps();
+		}
+		
+		bool CheckTerrain()
```
pchote: `OnValidTerrainType` would be more descriptive.

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -156,6 +164,23 @@ bool OnValidTerrain()
 			return info.AllowedTerrainTypes.Contains(terrainType);
 		}
 
+		bool CheckRamps()
```
pchote: `OnValidRampType` would be more descriptive.

abcdefg30:

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -28,6 +28,9 @@ public class DeployToUpgradeInfo : ITraitInfo, Requires<UpgradeManagerInfo>
 			"Leave empty to allow any.")]
 		public readonly HashSet<string> AllowedTerrainTypes = new HashSet<string>();
 
+		[Desc("Can this actor deploy on slopes?")]
+		public readonly bool CanDeployOnRamps = false;
```
Phrohdoh: The description and field name are inconsistent.

Mailaender: It is just not redundant which is actually better. Explaining terms with the same word isn't helping.

Mailaender:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/6dbb6a0ce3301ce0e346b2ba1881be09e264d4e2...0b2ca66206d1851a474e213af9178766a6037e86)

Issue #8301, opened 2015-05-30
ctrl+backspace doesn't delete last word in chat


ChaoticMind:
In the lobby as well as in-game, `ctrl`+`backspace` (or its osx equivalent) doesn't remove the last word when chatting. It's useful in case the user wants to just quickly clear the text (or the last n words).

x-a-n-a-x:
Similar to #2794

Issue #2794, opened 2013-03-19
Word jump and selection with CTRL (+SHIFT) + arrow key in lobby


xan2622:
In the lobby chat, the user should be able to perform several actions related to words and selection:
- [x] jump to the start of the sentense with CTRL+Home
- [x] jump to the end of the sentense with CTRL+End
- [x] jump to previous word with CTRL+left arrow
- [x] jump to next word with CTRL+right arrow
- [x] select the whole sentense (and see the selection) with CTRL+SHIFT+Home 
- [x] select the whole sentense (and see the selection) with CTRL+SHIFT+End
- [x] select the previous word (and see the selection) with CTRL+SHIFT+left arrow
- [x] select the next word (and see the selection) with CTRL+SHIFT+right arrow

<a href="http://www.hostingpics.net/viewer.php?id=270758ScreenShot001.png"><img src="http://img15.hostingpics.net/thumbs/mini_270758ScreenShot001.png" alt="Heberger image" /></a>

pchote:
This is controversial, because these shortcuts are windows-specific.  OSX, for example, uses ctrl+arrow keys for switching workspaces, and uses alt+arrow(+modifiers) for jumping words.

x-a-n-a-x:
Hmm... Can't OpenRA detect the kind of OS it runs on, and according to this, manage the shortcuts accordingly ?

ChaoticMind:
@x-a-n-a-x yes it can detect, at least it can easily do so now, not sure about at the time of posting.
btw, the first two checkboxes are already implemented (home and end, no need for ctrl, or?)

joealam:
It looks like the word-jumping part of this is done, but the selection isn't.

I can take a look, if this is still a feature request?

pchote:
This was fixed by #14980.

PR #9582, opened 2015-10-10
Implement some keyboard hotkeys for chat (TextFieldWidget)

ChaoticMind:
This implements #8301 and parts of #2794.

The first commits implements the features, the second is a refactor to clarify the code, remove redundancies and make it easier to modify/add to in the future. It's also more explicit in its behavior, especially with regards to some functions that were being called from conditionals.
Because of the refactor, I recommend reviewing the commits separately.

The implemented keyboard hotkeys in the text areas are:
- ctrl + left/right (navigate words) (cmd + left/right on osx)
- ctrl + backspace/delete (backspaces/deletes words) (alt + backspace/delete on osx)
- ctrl + u / ctrl + k (removes rest of line from cursor) (cmd backspace/delete on osx)
- ctrl + x (clears the whole line) (cmd + x on osx) (this is optional but I often use it, especially in text editors)

I don't have access to OSX, so someone has to test that it's working fine on OSX, and that the hotkeys are sane/expected behavior (coherent with how interface usually behaves in other apps/in the os)

Also, does home/end exist on OSX? should cmd + left/cmd + right behave like home/end instead? (edit: did the latter)

A third commit resets the cursor blinking which I think makes the UX much better
A fourth commit makes ctrl+x cut to clipboard instead of cutting to the ether.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
@@ -123,15 +123,15 @@ namespace OpenRA.Mods.Common.Widgets
 			Func<int> getPrevWhitespaceIndex = () =>
 				Text.Substring(0, CursorPosition).TrimEnd().LastIndexOf(' ') + 1;
 
-			Func<int> getNextWhitespaceIndex = () => {
+			Func<int> getNextWhitespaceIndex = () =>
+			{
 				var substr = Text.Substring(CursorPosition);
 				var substrTrimmed = substr.TrimStart();
 				var trimmedSpaces = substr.Length - substrTrimmed.Length;
 				var nextWhitespace = substrTrimmed.IndexOf(' ');
 				if (nextWhitespace == -1)
 					return Text.Length;
-				else
-					return CursorPosition + trimmedSpaces + nextWhitespace;
+				return CursorPosition + trimmedSpaces + nextWhitespace;
 			};
 
 			var isOSX = Platform.CurrentPlatform == PlatformType.OSX;
@@ -161,21 +161,24 @@ namespace OpenRA.Mods.Common.Widgets
 				case Keycode.LEFT:
 					ResetBlinkCycle();
 					if (CursorPosition > 0)
-						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) ||
-						(isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
+					{
+						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) || (isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
 							CursorPosition = getPrevWhitespaceIndex();
 						else
 							CursorPosition--;
+					}
+
 					break;
 
 				case Keycode.RIGHT:
 					ResetBlinkCycle();
 					if (CursorPosition <= Text.Length - 1)
-						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) ||
-						(isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
+					{
+						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) || (isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
 							CursorPosition = getNextWhitespaceIndex();
 						else
 							CursorPosition++;
+					}
 
 					break;
 
@@ -214,10 +217,10 @@ namespace OpenRA.Mods.Common.Widgets
 
 				case Keycode.X:
 					ResetBlinkCycle();
-					if (((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) ||
-						 (isOSX && e.Modifiers.HasModifier(Modifiers.Meta))) &&
-						 (!string.IsNullOrEmpty(Text)))
+					if (((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) || (isOSX && e.Modifiers.HasModifier(Modifiers.Meta))) &&
+						!string.IsNullOrEmpty(Text))
 					{
+						Game.Renderer.SetClipboardText(Text);
 						Text = Text.Remove(0);
 						CursorPosition = 0;
 						OnTextEdited();
@@ -230,8 +233,7 @@ namespace OpenRA.Mods.Common.Widgets
 					ResetBlinkCycle();
 					if (CursorPosition < Text.Length)
 					{
-						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) ||
-						(isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
+						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) || (isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
 							Text = Text.Substring(0, CursorPosition) + Text.Substring(getNextWhitespaceIndex());
 						else if (isOSX && e.Modifiers.HasModifier(Modifiers.Meta))
 							Text = Text.Remove(CursorPosition);
@@ -248,8 +250,7 @@ namespace OpenRA.Mods.Common.Widgets
 					ResetBlinkCycle();
 					if (CursorPosition > 0)
 					{
-						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) ||
-						(isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
+						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) || (isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
 						{
 							var prev_whitespace = getPrevWhitespaceIndex();
 							Text = Text.Substring(0, prev_whitespace) + Text.Substring(CursorPosition);
@@ -273,8 +274,7 @@ namespace OpenRA.Mods.Common.Widgets
 
 				case Keycode.V:
 					ResetBlinkCycle();
-					if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) ||
-						 (isOSX && e.Modifiers.HasModifier(Modifiers.Meta)))
+					if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) || (isOSX && e.Modifiers.HasModifier(Modifiers.Meta)))
 					{
 						var clipboardText = Game.Renderer.GetClipboardText();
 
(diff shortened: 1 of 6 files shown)


OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -115,82 +115,169 @@ public override bool HandleKeyPress(KeyInput e)
...
+				var substr_trimmed = Text.Substring(CursorPosition).TrimStart();
+				var trimmed_spaces = substr_len - substr_trimmed.Length;
+				var next_whitespace = substr_trimmed.IndexOf(' ');
+				if (next_whitespace == -1)
+					return Text.Length;
+				else
+					return CursorPosition + trimmed_spaces + next_whitespace;
+			};
+
+			Func<bool> isOSX = () => Platform.CurrentPlatform == PlatformType.OSX;
```
pchote: This won't change at runtime, so it can be a plain bool.

ChaoticMind: right, ofc :)

Mailaender:
Works as promised on Linux. :white_check_mark: I didn't know about most of the hotkeys, but I checked at least Kate Editor also has them.

OpenRA.Platforms.Default/Sdl2Input.cs
```diff
@@ -20,6 +20,7 @@ class Sdl2Input
 		MouseButton lastButtonBits = (MouseButton)0;
 
 		public string GetClipboardText() { return SDL.SDL_GetClipboardText(); }
+		public int SetClipboardText(string text) { return SDL.SDL_SetClipboardText(text); }
```
pchote: Please change this to return a bool: `SDL.SDL_SetClipboardText(text) == 0`.  The "non-zero integer means an error" status code isn't conventional in C#.

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -115,82 +120,180 @@ public override bool HandleKeyPress(KeyInput e)
...
 
-				return true;
-			}
+				case Keycode.BACKSPACE:
+					// cmd+backspace is equivalent to ctrl+u on non-osx
+					ResetBlinkCycle();
+					if (CursorPosition > 0)
+					{
+						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) ||
+						(isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
```
pchote: Please indent this by an additional tab

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -115,82 +120,180 @@ public override bool HandleKeyPress(KeyInput e)
...
-				var nl = clipboardText.IndexOf('\n');
-				if (nl > 0)
-					clipboardText = clipboardText.Substring(0, nl);
+				case Keycode.DELETE:
+					// cmd+delete is equivalent to ctrl+k on non-osx
+					ResetBlinkCycle();
+					if (CursorPosition < Text.Length)
+					{
+						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) ||
+						(isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
```
pchote: Additional tab here too.

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -115,82 +120,180 @@ public override bool HandleKeyPress(KeyInput e)
...
-			if (e.Key == Keycode.V &&
-				((Platform.CurrentPlatform != PlatformType.OSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) ||
-				 (Platform.CurrentPlatform == PlatformType.OSX && e.Modifiers.HasModifier(Modifiers.Meta))))
-			{
-				var clipboardText = Game.Renderer.GetClipboardText();
+				case Keycode.X:
+					ResetBlinkCycle();
+					if (((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) ||
+						 (isOSX && e.Modifiers.HasModifier(Modifiers.Meta))) &&
+						 (!string.IsNullOrEmpty(Text)))
```
pchote: You don't need the parens around the `!string.IsNullOrEmpty()`

pchote: Having the `||` and `&&` lined up is a bit misleading - the or actually applies to the first two groups, and the and to the result of that comparison and the third group.  You could make this a lot clearer by pulling out a temporary `cutHotkeyPressed` variable before the check.

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -115,82 +120,180 @@ public override bool HandleKeyPress(KeyInput e)
...
+							CursorPosition = getPrevWhitespaceIndex();
+						else
+							CursorPosition--;
+					break;
 
-				return true;
-			}
+				case Keycode.RIGHT:
+					ResetBlinkCycle();
+					if (CursorPosition <= Text.Length - 1)
```
pchote: Please add braces around the body of this outer if.  While the if/else is a single statement (so its not technically wrong), the potential for confusion about what the else belongs to means it always pays to be explicit in these cases.

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -115,82 +120,180 @@ public override bool HandleKeyPress(KeyInput e)
...
+						return true;
+					break;
 
-			if (e.Key == Keycode.LEFT)
-			{
-				if (CursorPosition > 0)
-					CursorPosition--;
+				case Keycode.LEFT:
+					ResetBlinkCycle();
+					if (CursorPosition > 0)
```
pchote: Same here.

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -115,82 +120,180 @@ public override bool HandleKeyPress(KeyInput e)
...
+				Text.Substring(0, CursorPosition).TrimEnd().LastIndexOf(' ') + 1;
+
+			Func<int> getNextWhitespaceIndex = () => {
+				var substr = Text.Substring(CursorPosition);
+				var substrTrimmed = substr.TrimStart();
+				var trimmedSpaces = substr.Length - substrTrimmed.Length;
+				var nextWhitespace = substrTrimmed.IndexOf(' ');
+				if (nextWhitespace == -1)
+					return Text.Length;
+				else
```
pchote: The if removes, so the body of the else doesn't need to be inside the else (just run it unconditionally).

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
ChaoticMind: done, for the record.

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -115,82 +120,183 @@ public override bool HandleKeyPress(KeyInput e)
 			if (!HasKeyboardFocus)
 				return false;
 
-			if ((e.Key == Keycode.RETURN || e.Key == Keycode.KP_ENTER) && OnEnterKey())
-				return true;
+			Func<int> getPrevWhitespaceIndex = () =>
```
pchote: I don't see any particular need to have these defined internally, so you should pull them out into helper methods so that they aren't reallocated every key event.

pchote:
This is close, but not quite right under OSX:

* In your implementation Cmd + left/right jumps to the start/end of the textbox instead of navigating by word (this is correct behavior on OSX, so don't change it.  Your description above is just wrong)
* In your implementation Cmd + backspace deletes the entire line before the cursor (this is correct behaviour in OSX, so don't change it.  Your description above is just wrong)
* Under OSX ctrl+backspace should delete the last word, like you have described for the other platforms (please add this).
* Under OSX ctrl+k should delete the rest of the line forward from the cursor, like you have described for the other platforms (please add this).
* Under OSX ctrl+u makes the word that the cursor is inside UPPERCASE.  Please don't add this, we don't want people to abuse it.


See also http://osxdaily.com/2012/02/23/keyboard-shortcuts-to-navigate-select-text-mac-os-x/ and http://www.amsys.co.uk/2012/08/useful-mac-text-editing-shortcuts/

ChaoticMind:
I updated this back then, forgot to comment. I think it should be proper behavior now.

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -74,15 +80,14 @@ public override bool HandleMouseInput(MouseInput mi)
...
-			showCursor = true;
+			ResetBlinkCycle();
 			CursorPosition = ClosestCursorPosition(mi.Location.X);
 			return true;
 		}
 
 		protected virtual string GetApparentText() { return text; }
 
-		public int ClosestCursorPosition(int x)
+		private int ClosestCursorPosition(int x)
```
abcdefg30: Please remove the explicit `private` here and below.

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -106,6 +111,22 @@ public int ClosestCursorPosition(int x)
...
+
+		private int GetNextWhitespaceIndex()
+		{
+			var substr = Text.Substring(CursorPosition);
+			var substrTrimmed = substr.TrimStart();
+			var trimmedSpaces = substr.Length - substrTrimmed.Length;
+			var nextWhitespace = substrTrimmed.IndexOf(' ');
+			if (nextWhitespace == -1)
+				return Text.Length;
+			return CursorPosition + trimmedSpaces + nextWhitespace;
```
abcdefg30: Can we get a new line above this line, please?

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
```diff
@@ -115,82 +136,178 @@ public override bool HandleKeyPress(KeyInput e)
...
+					break;
+
+				case Keycode.BACKSPACE:
+					// cmd+backspace is equivalent to ctrl+u on non-osx
+					ResetBlinkCycle();
+					if (CursorPosition > 0)
+					{
+						if ((!isOSX && e.Modifiers.HasModifier(Modifiers.Ctrl)) || (isOSX && e.Modifiers.HasModifier(Modifiers.Alt)))
+						{
+							var prev_whitespace = GetPrevWhitespaceIndex();
```
abcdefg30: `prevWhitespace`, please.

abcdefg30:
Looks good to me. :white_check_mark: 

:+1: after those style nits are resolved.

ChaoticMind:
Cleaned up the style nits

OpenRA.Mods.Common/Widgets/TextFieldWidget.cs
ChaoticMind: Pythonic habits :)

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/398488fa5aa34598b1eb748e74c5c11787479be9...ffbee94f5224281d315c687c26ff45e39d0be80d)

PR #9583, opened 2015-10-10
Enables Explored Map by default in lobby settings.

deniz1a:
This can be done now since #6123 is fixed.

### DIFF
--- a/OpenRA.Game/Network/Session.cs
+++ b/OpenRA.Game/Network/Session.cs
@@ -182,7 +182,7 @@ namespace OpenRA.Network
 			public string Difficulty;
 			public bool Crates = true;
 			public bool Creeps = true;
-			public bool Shroud = true;
+			public bool Shroud = false;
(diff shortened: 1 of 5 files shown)


penev92:
Why exactly would we want this?

deniz1a:
Because explored map is better.

MicroJOo:
nope :-1:

jaZzKCS:
Why would we want to change defaults?
This is just your personal preference.
In almost every RTS, shroud is an essential game element and always present.

Shroud should always be enabled by default to guarantee a certain game experience, and should only be disabled if there is a significant portion of players on the server that does NOT know the map, to make it more fair.

But in general, having shroud enabled is and will always be the standard.

obrakmann:
:-1: I agree with all the previous comments. This is already configurable, no need to change the default.

abcdefg30:
I don't think we're getting anywhere here.

PR #9584, opened 2015-10-10
Fix case sensitivity issue in d2k sequence definitions

obrakmann:
Game would crash on systems with case-sensitive filesystems when a support-power beacon appeared.


Mailaender:
:+1: Followup of https://github.com/OpenRA/OpenRA/pull/9441.

PR #9585, opened 2015-10-10
Make "select all units" work across screen/map again

ChaoticMind:
I noticed in the new release that 'q' no longer selects all units across the map (only across the screen). I couldn't find the commit that was responsible for this regression, but this PR addresses that.

I also made 'w' do less work if no units/classes are selected (no debug msg appears either)

The behavior now adheres to what's stated in the [wiki](https://github.com/OpenRA/OpenRA/wiki/Hotkeys). Updated the line in settings to better represent what 'q' does.

Although I incorrectly stated that it selects "military units" instead of "highest priority".

Do you think 'q' should instead select "military units" ie a specific priority instead of simply the highest one?

Note: this makes use of "Game.debug()" which might not be desirable. I could remove that or replace it with some placeholder (e.g. Game.WriteLine()) that calls Game.debug() until a better solution is available. I'm personally OK with leaving it as is (using Game.debug()).

### DIFF
--- a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
+++ b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
@@ -235,11 +235,25 @@ namespace OpenRA.Widgets
 				else if (key == Game.Settings.Keys.SelectAllUnitsKey && !World.IsGameOver)
 				{
 					// Select actors on the screen which belong to the current player
-					var ownUnitsOnScreen = SelectActorsOnScreen(World, worldRenderer, null, player).SubsetWithHighestSelectionPriority();
+					var ownUnitsOnScreen = SelectActorsOnScreen(World, worldRenderer, null, player).SubsetWithHighestSelectionPriority().ToList();
+
+					// Check if selecting actors on the screen has selected new units
+					if (ownUnitsOnScreen.Count > World.Selection.Actors.Count())
+						Game.Debug("Selected across screen");
+					else
+					{
+						// Select actors in the world that have highest selection priority
+						ownUnitsOnScreen = SelectActorsInWorld(World, null, player).SubsetWithHighestSelectionPriority().ToList();
+						Game.Debug("Selected across map");
+					}
+
 					World.Selection.Combine(World, ownUnitsOnScreen, false, false);
 				}
 				else if (key == Game.Settings.Keys.SelectUnitsByTypeKey && !World.IsGameOver)
 				{
+					if (!World.Selection.Actors.Any())
+						return false;
+
 					// Get all the selected actors' selection classes
 					var selectedClasses = World.Selection.Actors
 						.Where(x => !x.IsDead && x.Owner == player)
@@ -250,7 +264,7 @@ namespace OpenRA.Widgets
 					var newSelection = SelectActorsOnScreen(World, worldRenderer, selectedClasses, player).ToList();
 
 					// Check if selecting actors on the screen has selected new units
-					if (newSelection.Count() > World.Selection.Actors.Count())
+					if (newSelection.Count > World.Selection.Actors.Count())
 						Game.Debug("Selected across screen");
 					else
 					{
--- a/OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
@@ -394,7 +394,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 					{ "CycleBaseKey", "Jump to base" },
 					{ "ToLastEventKey", "Jump to last radar event" },
 					{ "ToSelectionKey", "Jump to selection" },
-					{ "SelectAllUnitsKey", "Select all units on screen" },
+					{ "SelectAllUnitsKey", "Select all combat units" },
 					{ "SelectUnitsByTypeKey", "Select units by type" },
 
 					{ "PlaceBeaconKey", "Place beacon" },


ChaoticMind:
I have a couple of tangentially related questions: 
Where is `WorldInteractionControllerWidget` instantiated? `git grep WorldInteractionControllerWidget` only shows it in `OpenRA.Mods.Common/Widgets/RadarWidget.cs` under `var controller = Ui.Root.Get<WorldInteractionControllerWidget>(WorldInteractionController);` whose syntax I can't really claim I understand (it looks more like some sort of singleton getter?).

If there's indeed just a single instance of `WorldInteractionControllerWidget`, should it have "world" as a static variable instead? This removes a bunch of parameter-passing to the `WorldInteractionControllerWidget` static methods, such as `SelectActorsOnScreen` and sisters.

RoosterDragon:
Widgets are all loaded via reflection, see `WidgetLoader.cs`. All their config is stored in YAML files which is parsed and applied at runtime.

`Ui.Root.Get`: The `Root` widget is the top-level widget - the whole UI tree is attached to this guy. The method itself finds a widget by a string ID that matches the type. In this case, the radar wants to send some input to this widget so it can move the location of the screen in the world.

Static variables are a bad idea - if you ever wanted more than one `WorldInteractionControllerWidget` with a different world it would all go horribly wrong. In this case a better idea might be to unstatic the methods so they can make use of the instance variables (but I don't think we need to tackle that in this PR or anything).

Hope that helps.

ChaoticMind:
Great, thanks!

OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
```diff
@@ -236,10 +236,24 @@ public override bool HandleKeyPress(KeyInput e)
 				{
 					// Select actors on the screen which belong to the current player
 					var ownUnitsOnScreen = SelectActorsOnScreen(World, worldRenderer, null, player).SubsetWithHighestSelectionPriority();
```
RoosterDragon: You'll want to stick a `ToList` on this (like `SelectUnitsByTypeKey` block does) to prevent it being enumerated from scratch multiple times.

OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
```diff
@@ -236,10 +236,24 @@ public override bool HandleKeyPress(KeyInput e)
...
+
+					// Check if selecting actors on the screen has selected new units
+					if (ownUnitsOnScreen.Count() > World.Selection.Actors.Count())
+						Game.Debug("Selected across screen");
+					else
+					{
+						// Select actors in the world that have highest selection priority
+						ownUnitsOnScreen = SelectActorsInWorld(World, null, player).SubsetWithHighestSelectionPriority();
```
RoosterDragon: `ToList` here too.

OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
```diff
@@ -394,7 +394,7 @@ Action InitInputPanel(Widget panel)
 					{ "CycleBaseKey", "Jump to base" },
 					{ "ToLastEventKey", "Jump to last radar event" },
 					{ "ToSelectionKey", "Jump to selection" },
-					{ "SelectAllUnitsKey", "Select all units on screen" },
+					{ "SelectAllUnitsKey", "Select all military units" },
```
RoosterDragon: I don't have any particular opinions on this - but I'm commenting here to draw attention for other reviewers. :)

RoosterDragon:
:+1: This seems reasonable.

Would be nice to get some opinions on the descriptive text for the hotkey before merging though.

OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
pchote: Not sure of the context behind this change (i.e. whether we want to change it at all), but the convention is for "combat units" rather than military.

ChaoticMind: The context is that it's no longer just "on screen" but alternates between "on screen" and "across the entire map".

Phrohdoh:
@ChaoticMind are you still around for this?

ChaoticMind:
Yeah, sorry, I've been quite busy. I'll follow up on my PRs during the weekend!

ChaoticMind:
Updated.

OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
```diff
@@ -235,11 +235,25 @@ public override bool HandleKeyPress(KeyInput e)
 				else if (key == Game.Settings.Keys.SelectAllUnitsKey && !World.IsGameOver)
 				{
 					// Select actors on the screen which belong to the current player
-					var ownUnitsOnScreen = SelectActorsOnScreen(World, worldRenderer, null, player).SubsetWithHighestSelectionPriority();
+					var ownUnitsOnScreen = SelectActorsOnScreen(World, worldRenderer, null, player).SubsetWithHighestSelectionPriority().ToList();
```
pchote: `.ToArray` and then `.Length` below?

ChaoticMind: How about keeping it .ToList() and replacing .Count() with .Count instead?

pchote: Ok.

ChaoticMind:
Updated

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3bc138b6527d4a4edac9c55c05f31b86560a8679...c9fd827198d8545c5c99dab2b56eabcad327079a)

PR #9586, opened 2015-10-10
Overhaul IRC in preparation for the global chat UI

pchote:
The existing code didn't scale to multiple panels, so I have reorganized and simplified it in preparation for the lobby and ingame integration.

This mainly focuses on the backend code, but there are a few visible changes:
- Polished topic and op/voice status indicators.
- Correctly handles kicks and errors.
- Distinguishes between chat and notifications.


pchote:
The dispose commit will need to be redone.  It currently disposes even when a window is temporarily hidden, which then breaks the chat when it is added to the game lobby (doesn't affect testing in the server brwoser, though).

OpenRA.Mods.Common/Widgets/Logic/ServerBrowserLogic.cs
```diff
@@ -135,14 +135,7 @@ public ServerBrowserLogic(Widget widget, Action onStart, Action onExit, string d
...
 
-			try
-			{
-				Game.LoadWidget(null, "SERVERBROWSER_IRC", panel.Get("IRC_ROOT"), new WidgetArgs());
-			}
-			catch
-			{
-				Log.Write("debug", "Failed to load server browser IRC chrome layout");
-			}
+			Game.LoadWidget(null, "GLOBALCHAT_PANEL", panel.Get("GLOBALCHAT_ROOT"), new WidgetArgs());
```
Mailaender: This will make the IRC feature mandatory for every mod. Attention @Phrohdoh this will break your RA2 branch.

pchote:
Fixed the disposing logic.

OpenRA.Mods.Common/Widgets/Logic/ServerBrowserLogic.cs
penev92: I don't think we should be making any widget mandatory, and we shouldn't force (our!) IRC on thirdparty mods.

OpenRA.Mods.Common/Widgets/Logic/GlobalChatLogic.cs
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		Widget MakeHistoryWidget(object o)
+		{
+			var message = (ChatMessage)o;
+			var widget = (LabelWidget)historyTemplate.Clone();
+			var font = Game.Renderer.Fonts[widget.Font];
+
+			var color = message.Type == ChatMessageType.Notification ?
+				Color.LightGray : Color.White;
```
Mailaender: :+1: I like the subtle color differentiation, but I guess this wants to live in `metrics.yaml` instead.

OpenRA.Mods.Common/Widgets/Logic/ServerBrowserLogic.cs
pchote: Mods that don't want this can define a dummy `GLOBALCHAT_PANEL` widget to disable it.  The try/catch makes it impossible to debug any errors in the IRC logic.

pchote:
This can crash if the IRC thread adds a message to the History or Users collections while the UI thread is enumerating them.  I'll need to ensure that all dictionary accesses are done on the main thread - running just the widget updates in a RunAfterTick is not enough.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,342 @@
...
+
+		public ChatUser(string name, bool isOp, bool isVoice)
+		{
+			Name = name;
+			IsOp = isOp;
+			IsVoiced = isVoice;
+		}
+	}
+
+	public sealed class ChatUserComparer : IComparer<string>
```
RoosterDragon: Use one on the static members on `System.StringComparer` and save yourself this class.

pchote:
Fixed.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,358 @@
...
+			if (Type == ChatMessageType.Notification)
+				return "{0} {1}".F(time, Message);
+
+			return "{0} {1}: {2}".F(time, Nick, Message);
+		}
+	}
+
+	public class GlobalChat : IDisposable
+	{
+		volatile IrcClient client;
```
RoosterDragon: `volatile` is pointless here.

pchote: This was carried over from the earlier code, where I believe you asked for it yourself :P.
But you're right - this is only ever accessed from the worker thread now.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,358 @@
...
+		public readonly ObservableSortedDictionary<string, ChatUser> Users = new ObservableSortedDictionary<string, ChatUser>(StringComparer.InvariantCultureIgnoreCase);
+		public readonly ObservableList<ChatMessage> History = new ObservableList<ChatMessage>();
+		public string Topic { get; private set; }
+		public ChatConnectionStatus ConnectionStatus { get; private set; }
+
+		static GlobalChat()
+		{
+			// Set a random default nick
+			if (Game.Settings.Chat.Nickname == new ChatSettings().Nickname)
+				Game.Settings.Chat.Nickname += Game.CosmeticRandom.Next(100, 999);
```
RoosterDragon: Modifying globals from a random static constructor is just asking for trouble with initialization order. You should move this somewhere more deterministic.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,358 @@
...
+
+	public class GlobalChat : IDisposable
+	{
+		volatile IrcClient client;
+		Channel channel;
+
+		public readonly ObservableSortedDictionary<string, ChatUser> Users = new ObservableSortedDictionary<string, ChatUser>(StringComparer.InvariantCultureIgnoreCase);
+		public readonly ObservableList<ChatMessage> History = new ObservableList<ChatMessage>();
+		public string Topic { get; private set; }
+		public ChatConnectionStatus ConnectionStatus { get; private set; }
```
RoosterDragon: You're reading/writing this from both the UI and IRC threads without any syncing.

pchote: Made the backing field for both of these volatile.

pchote:
Updated.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,353 @@
...
+			connectionStatus = ChatConnectionStatus.Connecting;
+		}
+
+		void OnConnected(object sender, EventArgs e)
+		{
+			AddNotification("Connected.");
+			connectionStatus = ChatConnectionStatus.Connected;
+
+			// Guard against settings.yaml modification
+			Game.Settings.Chat.Nickname = SanitizedName(Game.Settings.Chat.Nickname);
```
RoosterDragon: This write isn't thread-safe (since we usually only muck with globals on the UI thread).

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,353 @@
...
+			else
+				connectionStatus = ChatConnectionStatus.Disconnecting;
+
+			if (!client.IsConnected)
+				return;
+
+			client.RfcQuit(Game.Settings.Chat.QuitMessage);
+
+			AddNotification("Disconnecting from {0}...".F(client.Address));
+			Game.Settings.Chat.ConnectAutomatically = false;
```
RoosterDragon: This write isn't thread-safe.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,353 @@
...
+				if (Users.TryGetValue(whom, out user))
+					user.IsOp = isOp;
+			});
+		}
+
+		void SetUserVoiced(string whom, bool isVoiced)
+		{
+			Game.RunAfterTick(() =>
+			{
+					ChatUser user;
```
RoosterDragon: You have an extra indent here.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,353 @@
...
+
+		public bool IsValidNickname(string name)
+		{
+			return Rfc2812.IsValidNickname(name);
+		}
+
+		public void SendMessage(string text)
+		{
+			AddMessage(client.Nickname, text);
+			client.SendMessage(SendType.Message, channel.Name, text);
```
RoosterDragon: Accessing channel here isn't thread-safe.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,353 @@
...
+			if (Type == ChatMessageType.Notification)
+				return "{0} {1}".F(time, Message);
+
+			return "{0} {1}: {2}".F(time, Nick, Message);
+		}
+	}
+
+	public class GlobalChat : IDisposable
+	{
+		IrcClient client;
```
RoosterDragon: `readonly`.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,353 @@
...
+		public readonly string UID;
+
+		public ChatMessage(DateTime time, ChatMessageType type, string nick, string message)
+		{
+			Time = time;
+			Type = type;
+			Nick = nick;
+			Message = message;
+
+			UID = (++nextUID).ToString();
```
RoosterDragon: Maybe use `Interlocked.Increment` so if anybody else ever decides to use this class they don't get duplicate IDs.

pchote:
Fixed some more.  Also fixed a jank when first connecting to the IRC sever.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,369 @@
...
+					client.Connect(Game.Settings.Chat.Hostname, Game.Settings.Chat.Port);
+				}
+				catch (Exception e)
+				{
+					connectionStatus = ChatConnectionStatus.Error;
+					AddNotification(e.Message);
+					Game.RunAfterTick(() => Log.Write("irc", e.ToString()));
+				}
+
+				client.Listen();
```
RoosterDragon: If the connect fails, do we still want to listen?

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,369 @@
...
+				Game.RunAfterTick(() => Game.Settings.Chat.Nickname = nick);
+
+			client.Login(nick, "in-game IRC client", 0, "OpenRA");
+			client.RfcJoin("#" + Game.Settings.Chat.Channel);
+		}
+
+		void OnDisconnecting(object sender, EventArgs e)
+		{
+			if (connectionStatus != ChatConnectionStatus.Error)
+				connectionStatus = ChatConnectionStatus.Connecting;
```
abcdefg30: Is this intentionally not `ChatConnectionStatus.Disconnecting`?

abcdefg30:
Chat still works. :white_check_mark: / :+1:

pchote:
Updated.

OpenRA.Game/GlobalChat.cs
```diff
@@ -0,0 +1,371 @@
...
+		void OnChannelActiveSynced(object sender, IrcEventArgs e)
+		{
+			channel = client.GetChannel(e.Data.Channel);
+			AddNotification("{0} users online".F(channel.Users.Count));
+			connectionStatus = ChatConnectionStatus.Joined;
+
+			foreach (DictionaryEntry user in channel.Users)
+			{
+				var u = (ChannelUser)user.Value;
+				Game.RunAfterTick(() => Users.Add(u.Nick, new ChatUser(u.Nick, u.IsOp, u.IsVoice)));
```
obrakmann: I guess this is supposed to sync the server's and client's state, right? If so, wouldn't you need to use `Users.Clear()` before repopulating the list?

pchote: I haven't read the docs on this callback, but @Mailaender set it up as if it were used exclusively on first join to the channel.  If it may be called multiple times, it would indeed be necessary to clear the dict.

OpenRA.Game/OpenRA.Game.csproj
```diff
@@ -84,6 +84,9 @@
       <HintPath>..\thirdparty\download\MaxMind.Db.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="SmarIrc4net">
+      <HintPath>..\thirdparty\download\SmarIrc4net.dll</HintPath>
```
obrakmann: If we ever get to extracting a dedicated server into a single .exe, this would probably mean that it would depend on this .dll, right?

OpenRA.Game/GlobalChat.cs
obrakmann: Looks like it only happens that once. At least it doesn't happen when users join or leave. Could be that it's run once every hour or so, but I didn't test that long.

obrakmann:
Looks and works fine. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/6fb4b9513b834e06ff2e5dab7d611cee923070ff)

Issue #8584, opened 2015-06-28
Check code for missing Requires<>


reaperrr:
If we managed to find and add all missing `Requires<>` throughout the code, the number of potential inheritance issues and crashes would go down.
##

Mailaender:
That probably doesn't work as tracker ticket as this can never be completed. It is a never-ending task. Also writing a lint rule like we have for ISync may not be easy.

Unit158:
This should go in contributing if it isn't already there.

Phrohdoh:
Could we have Travis launch the engine? (which would fail on a missing Requires)

Mailaender:
It won't always fail because the missing requires only happens when a trait is activated and the order is wrong so you would kinda need to start a replay in headless mode or something.

penev92:
That is pretty much way out of scope for a CI environment. Would it be OK if we exploited either Travis or AppVeyor like that?

Mailaender:
I doubt it won't be much fun to run a game in a CI build VM. The builds will time out or we will simply get banned I guess. Establishing some automated tests via scripted gameplay instead of stupid unit tests is a good idea. I think @cjshmyr had it before, but I don't think he has implemented anything.

cjshmyr:
Can confirm that I hadn't implemented anything. What @Mailaender is referring to is [here](https://github.com/OpenRA/OpenRA/commit/8d6a82540704bcf250f125560c3411b51dcdad4f), and @pchote had a comment about this as well @ https://github.com/OpenRA/OpenRA/pull/6201#issuecomment-52410395

reaperrr:
I guess this isn't fixable in any permanent way, and right now there don't seem to be any major problems stemming from this, so I'm closing this.

PR #9588, opened 2015-10-10
Fix sprite turret recoil.

### LINKED ISSUE #9587: Recoil:  isnt working anymore
i really miss moving turrets while firing with tanks..

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27315557-recoil-isnt-working-anymore?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #9587.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithTurret.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithTurret.cs
@@ -18,7 +18,7 @@ namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Renders turrets for units with the Turreted trait.")]
 	public class WithTurretInfo : UpgradableTraitInfo, IRenderActorPreviewSpritesInfo,
-		Requires<RenderSpritesInfo>, Requires<TurretedInfo>, Requires<BodyOrientationInfo>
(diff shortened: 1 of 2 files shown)


obrakmann:
This fixes it for tanks and ships. But didn't the gun turret also have recoil?

Mailaender:
The RA and TD can't have recoil, because the artwork is just one SHP and the D2k gun turret does not have it set up in the Armament so I think the issue is resolved in general here.

Mailaender:
Thanks for investigating this. Another reason for something like https://github.com/OpenRA/OpenRA/issues/8584.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d78cef09b26889be39fb63508d15e85e691426bc...38d529f50cacc39fc61eec21797ad7f5e8d1fccc)

Issue #6082, opened 2014-07-27
New shortcut: CTRL+rClick -> instant cancel unit that is being built [$20]


VagueVigor:
I really hate the slowness of canceling some unit from building; first it makes it on hold, and then after 10 clicks (because game lags), it will finally cancel it.

It would be super awesome if there was shortcut to insta-cancel anything.

So if you CTRL+rightclick, it would CANCEL the unit (you get money back from whatever was used for it before). Not just remove 1 from queue like the current cancel behaviour acts now; in current way its really hard to cancel something if you have 100 on queue.

r34ch:
This would make more sense if it instantly cancelled all queued units of that type. Rather than mashing shift+rclick

Mailaender:
Implemented in https://github.com/OpenRA/OpenRA/pull/9595. @ChaoticMind please collect your bounty.

PR #9595, opened 2015-10-11
Middle click to cancel directly (skips onhold) & ctrl+cancel to cancel entire queue

ChaoticMind:
Implemented a fix to a pet peeve which is having to go through "onhold" before cancelling an item. Especially frustrating during lag spikes (and misclicks of adding extra units/structures). If we middle click instead of right click, we can now directly cancel an item without interrupting the currently building/training item. This also gives us a lot more versatility in fixing up the queue for long term construction projects. 

I also implemented `ctrl + click` to cancel an entire queue, resolving #6082.
- `Ctrl + leftclick` doesn't do anything (trains one unit, except if it's `ctrl + shift + leftclick` which behaves as `shift + leftclick`.
- `Ctrl + middleclick`: immediately cancels the whole queue (without going to "onhold")
- `Ctrl + rightclick`: goes to onhold if not already there. If already in onhold, cancels the entire queue.

Just a note: If someone gets the bright idea of implementing modifiers for keyboard hotkeys (like shift+f1 to train 5 items, or ctrl+f1 to cancel an item or something, then it's an exercise in futility since ctrl+f1 won't match the hotkey for the building at all, since it could be binded to something else anyway :p). It's debatable if we care enough to look for some solution to this.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ProductionPaletteWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ProductionPaletteWidget.cs
@@ -182,15 +182,11 @@ namespace OpenRA.Mods.Common.Widgets
 			if (icon == null)
 				return false;
 
-			// Only support left and right clicks
-			if (mi.Button != MouseButton.Left && mi.Button != MouseButton.Right)
-				return false;
-
 			// Eat mouse-up events
 			if (mi.Event != MouseInputEvent.Down)
 				return true;
 
-			return HandleEvent(icon, mi.Button == MouseButton.Left, mi.Modifiers.HasModifier(Modifiers.Shift));
+			return HandleEvent(icon, mi.Button, mi.Modifiers);
 		}
 
 		protected bool PickUpCompletedBuildingIcon(ProductionIcon icon, ProductionItem item)
@@ -216,7 +212,7 @@ namespace OpenRA.Mods.Common.Widgets
 			}
 		}
 
-		bool HandleLeftClick(ProductionItem item, ProductionIcon icon, bool handleMultiple)
+		bool HandleLeftClick(ProductionItem item, ProductionIcon icon, int handleCount)
 		{
 			if (PickUpCompletedBuildingIcon(icon, item))
 			{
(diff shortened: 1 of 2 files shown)


ChaoticMind:
I had some questions: 
- In `OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs`, `CancelProduction()` gets called twice, why? (The second time, `queue` is an empty list which is why it doesn't actually cancel two items). This looks like a possible benign bug?
- Why are things like `QueueLength` and `CurrentRemainingCost` etc in `sync`? Where are threads playing role here?
- What's with the whole serialization stuff in `Order` anyway?

GraionDilach:
The last two are for multiplayer. [Sync]ed variables are synced between players and observer statistics/infiltration/etc plays the requirement for syncing production properties, Order is also serialized to not cause mismatches.

ChaoticMind:
Right, makes sense. So I'm guessing since cancelling production orders is always on the local queues, I don't need to worry about getting an invalid value for `QueueLength` [here](https://github.com/ChaoticMind/OpenRA/commit/d82bf4d3caeec14eee567836aeb38eddd7db58f3#diff-6b8b708073f7d1d3ef01dd77c279c596R282).

I guess it could become an issue if an action is ordered locally and we won't know if it succeeds or fails until a roundtrip with peers/server, for example if a building gets destroyed or something like that. It that happens though, (as long as it doesn't crash the game), it will never have wrong behavior since the queue can't increase due to a remote action.

But I'm not sure if I should have some sort of safety around accessing QueueLength in case it becomes invalid (due to a building being destroyed), therefore crashing the game.

pchote:
>  [Sync]ed variables are synced between computers and observer statistics/infiltration/etc plays the requirement for syncing production properties

You need to be careful with your wording here, because the word "sync" is ambigous.
The `[Sync]` attribute is for *verifying* that variables are in sync over the network: if the variable has different values on different clients the game will generate a desync error.  It has nothing to do with actually copying the value between clients (that is done via orders).

RoosterDragon:
:+1: Works for me and doesn't appear to break the other hotkey combos. Code looks fine.

OpenRA.Mods.Common/Widgets/ProductionPaletteWidget.cs
```diff
@@ -269,11 +263,28 @@ bool HandleRightClick(ProductionItem item, ProductionIcon icon, bool handleMulti
 			return true;
 		}
 
-		bool HandleEvent(ProductionIcon icon, bool isLeftClick, bool handleMultiple)
+		bool HandleMiddleClick(ProductionItem item, ProductionIcon icon, int handleCount)
+		{
+			if (item == null)
+				return false;
+
+			// directly cancel, skipping "on-hold"
```
abcdefg30: Comments start with a captial letter, so `Dircetly`, please.

abcdefg30:
Looks good to me, too.
:+1: / :white_check_mark: with that tiny nit fixed.

OpenRA.Mods.Common/Widgets/ProductionPaletteWidget.cs
Phrohdoh: `Directly`

abcdefg30: Yeah, I meant that. *facepalms*

ChaoticMind:
Added self to authors and capitalized the comment.

ChaoticMind:
Added self to authors and capitalized the comment.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/137f3f6cf52f3c61afda0537f5002c76a6a953e9...dd15919198e79ba3f314184a13a4948bb44243ff)

abcdefg30:
Note: We probably want to update https://github.com/OpenRA/OpenRA/wiki/Hotkeys when this gets into a release.

Issue #9427, opened 2015-09-24
"Enable Pixel Doubling" on small 4k monitor isn't enough...


bobosch:
Tested with Release 20150919 on linux

I'm using a 28" 4k monitor. The units are really small. Even with enabled pixel doubling. Would be nice if this option would be replaced by a "Zoom" (none, 2x, 3x, 4x) option or possibility to use the mouse wheel to zoom during the game.

By the way: The build menu on top right isn't affected by pixel doubling :-(

penev92:
All I can say to this is "tough...".
A 20-year-old game's assets are naturally not meant to be visible on a miniature screen with 4k resolution. Neither are modern-day games, for that matter.
The widgets (radar, build palette, etc.) are intentionally unaffected by pixel doubling.

And "zooming" has been discussed a lot and has been decided it's just not feasable with those assets. Also I'm pretty sure we already have an issue open with a feature request for "zooming", so if you don't mind I'll close this now.

GraionDilach:
A workaround is to specify a custom fullscreen resolution in the user settings - even though the UI hides the width/height boxes when fullscreen is selected, it still works when specified manually.

bobosch:
@GraionDilach The workaround doesn't work here :-( Yes, a possibility to specify the fullscreen resolution would do the trick.
@penev92 Sorry, I found no issues like this. #6405 requests a configurable fullscreen resolution. Would be fine for me, too.

PR #9597, opened 2015-10-11
Mouse wheel zooming

nicholas-ochoa:
This change adds support for mousewheel zooming, with a minimum zoom level of 0.5 (half of the current) and a maximum of 4.0 (double the current).

### DIFF
--- a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
+++ b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
@@ -144,7 +144,35 @@ namespace OpenRA.Widgets
 				}
 			}
 
-			lastMousePosition = xy;
+            if (mi.Event == MouseInputEvent.Scroll && mi.ScrollDelta == +1)
+            {
+                var zoomLevel = worldRenderer.Viewport.Zoom + 0.5f;
+
+                //maximum level = 4
+                if (zoomLevel > 4f)
+                {
+                    zoomLevel = 4f;
+                }
+
+                worldRenderer.Viewport.Zoom = zoomLevel;
+                Game.Settings.Graphics.PixelDouble = true;
+            }
+
+            if (mi.Event == MouseInputEvent.Scroll && mi.ScrollDelta == -1)
+            {
+                var zoomLevel = worldRenderer.Viewport.Zoom - 0.5f;
+
+                //minimum level = 0.5
+                if (zoomLevel < 0.5f)
+                {
+                    zoomLevel = 0.5f;
+                }
+
+                worldRenderer.Viewport.Zoom = zoomLevel;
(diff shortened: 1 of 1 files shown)


kagaku:
Not sure if there is a better place to store the minimum/maximum zoom levels; but these seem to be sane defaults.

steelphase:
pr does exactly what is says, and works pretty well. I'd like to see an option to enable or disable it at least. Also because it doesn't have a modifier key to activate it, it could potentially interfere with other actions that involve the scroll bar.

x-a-n-a-x:
I also want this feature (and other players already requested it in the past). 
Thank you @kagaku for your contribution.

I agree with @steelphase, there needs to be a modifier (CTRL, ALT, Shift ...) key to activate the zoom. A modifier and a way to change the hotkey.

ChaoticMind:
I was thinking of implementing this, but according to [this comment](https://github.com/OpenRA/OpenRA/issues/9427#issuecomment-143050457), it's not entirely clear whether the general consensus is to have this or not.

If this is implemented, it might be desirable to get rid of the concept of "pixel doubling" and replace its [hotkey](https://github.com/OpenRA/OpenRA/commit/9c483a912ec4605f061ee653e10675c5d3d9bebb) with two (e.g. +/-) and as @x-a-n-a-x said, have a modifier and also make that configurable.

x-a-n-a-x:
After testing the branch, here is my feeling:

The mousewheel should let zooming out much farer (currently, it only zooms out by 0.5) while it's possible to zoom in closer than 4.0. Imo, there is more interest in zooming out than zooming closer than 2:1.

The point of zooming out is to be able to view a large part of the battlefield. 0.5 is not enough. It currently doesn't feel like SupremeCommander.

Also, I noticed some graphic glitches sometimes:
screenshot001

pchote:
We originally wrote the zoom code with support for non-integer zooms (which is why you were able to enable this so easily), but we then had to disable it because this introduces unacceptable tradeoffs.  Like @x-a-n-a-x pointed out above there can be serious graphical glitches (such as 1 pixel wide holes in the map) on certain GPUs based on the way that they round the fractional numbers.  Fixing those requires some much deeper changes in the rendering code that are still planned, but just haven't happened yet.

kagaku:
Thanks for the feedback, hopefully this makes it in someday. :grin:

x-a-n-a-x:
Again, thank you for contributing to Openra, @kagaku. We really appreciate your help.

If you want, you can contact us on our IRC channel to live chat with us : 
http://www.openra.net/community/

We always appreciate to meet and discuss with new developers :-)

digitalheir:
(This still works in replay mode?)

In my opinion, as a user, the glitches are acceptable if you put a warning next to it. Because the glitches are incidental and usually disappear after zooming in and out a couple of times.

image
image

Issue #9598, opened 2015-10-11
Rename "Show Bot Debug Messages" & expose debug messages to Lua


clemty:
it would be useful to have toggle-able debug messages in single-player missions which could remain in the code after merging - this could serve as a way to comment the mission code as well

This could be tied to the "Show Debug Messages"-Setting (maybe prefix the messges with "Lua: "?), but in that case the setting should be renamed to something like "Show Bot/Lua Debug Messages")

(currently I'm using Media.DisplayMessage for debugging and removing them before pushing)

Mailaender:
https://github.com/OpenRA/OpenRA/pull/9599/ implemented it although retaining "bot debug messages" as a separate button.

PR #9599, opened 2015-10-11
Add a Lua wrapper to display debug messages (shows if setting enabled)

ChaoticMind:
Addresses #9598.

Can be tested in lua with `Media.Debug()`

### DIFF
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -101,6 +101,7 @@ namespace OpenRA
 	public class DebugSettings
 	{
 		public bool BotDebug = false;
+		public bool LuaDebug = false;
 		public bool PerfText = false;
 		public bool PerfGraph = false;
 		public float LongTickThresholdMs = 1;
--- a/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
@@ -177,6 +177,15 @@ namespace OpenRA.Mods.Common.Scripting
 			Game.AddChatLine(c, prefix, text);
 		}
 
+		[Desc("Displays a debug message to the player, if \"Show Map Debug Messages\" is checked in the settings.")]
+		public void Debug(string text)
+		{
+			if (string.IsNullOrEmpty(text) || !Game.Settings.Debug.LuaDebug)
+				return;
+
+			Game.Debug(text);
+		}
+
 		[Desc("Display a text message at the specified location.")]
 		public void FloatingText(string text, WPos position, int duration = 30, HSLColor? color = null)
 		{
--- a/mods/cnc/chrome/settings.yaml
+++ b/mods/cnc/chrome/settings.yaml
@@ -531,6 +531,13 @@ Container@SETTINGS_PANEL:
 							Height: 20
 							Font: Regular
 							Text: Check Sync around Unsynced Code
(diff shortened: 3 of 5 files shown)


OpenRA.Game/Settings.cs
```diff
@@ -101,6 +101,7 @@ public ServerSettings(ServerSettings other)
 	public class DebugSettings
 	{
 		public bool BotDebug = false;
+		public bool LUADebug = false;
```
pchote: Lua, not LUA.  (http://www.lua.org/about.html even has a special section about this).

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -177,6 +177,15 @@ public void DisplayMessage(string text, string prefix = "Mission", HSLColor? col
 			Game.AddChatLine(c, prefix, text);
 		}
 
+		[Desc("Displays a debug message to the player, if \"LUADEBUG_CHECKBOX\" is checked in the settings.")]
```
pchote: Please reference it by the player-visible name rather than the internal ID.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -177,6 +177,15 @@ public void DisplayMessage(string text, string prefix = "Mission", HSLColor? col
...
+		public void Debug(string text)
+		{
+			if (string.IsNullOrEmpty(text) || !Game.Settings.Debug.LUADebug)
+				return;
+
+				Game.Debug(text);
```
pchote: This is indented too far.

mods/cnc/chrome/settings.yaml
```diff
@@ -543,4 +550,3 @@ Container@SETTINGS_PANEL:
 			Width: 140
 			Height: 35
 			Text: Reset
-
```
pchote: Please keep the end-of-file newlines.

mods/ra/chrome/settings.yaml
```diff
@@ -532,4 +532,10 @@ Background@SETTINGS_PANEL:
 					Height: 20
 					Font: Regular
 					Text: Check Sync around Unsynced Code
-
```
pchote: Likewise here (keep the newline after your final line of new yaml).

mods/cnc/chrome/settings.yaml
ChaoticMind: Actually there were two newlines at end of file, my editor automatically removes trailing newlines and keeps only one (on-save). You can tell because github otherwise marks it in a red "return-style" arrow if the last newline is removed. 
I don't mean to nitpick, but it's actually really annoying (and prone to be forgotten) to have to open up nano/different editor and manually add however many extra trailing newlines were in the file pre-save (and have to do it again whenever I amend something :()
So I'm hoping this is ok, since I don't think trailing newlines are desirable anyway, and they were probably there by accident.

pchote: Ah okay, in that case its fine.

abcdefg30:

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/398488fa5aa34598b1eb748e74c5c11787479be9)

Issue #4690, opened 2014-02-20
Prevent UI sounds from 'overlapping'


NinovanderMark:
Currently, if you click on a UI button repeatedly, like when building multiple units, the sample for that action is played as many times as the button is clicked. This is pretty crude and a better way to handle that scenario should be implemented.

My suggestion would be that only 1 of each UI sample may be played at the same time. Don't playback the sample if it's already playing.
##

Mailaender:
We talked about it in IRC some time. What we want is probably a `NotificationManager` trait that queues up the speech and sounds, then only plays a distinct lists. It could also go along with a `NotificationReference` lint test to check for erroneous entries in other traits.

deleted-user-1:
Bump to Next lol.

dragunoff:
I took a stab at fixing this but I stumbled at the question of how to handle cases when a notification overlaps itself? This has been discussed a bit in the thread above but I want to elaborate a bit more.

Currently there are no checks so all notifications play, for example:
```
Unit lost.
Unit lost.
 Unit lost.
  Unit lost.
    Unit lost.
          Unit lost.
```

One way to handle this is to stop the audio that's playing and then start it again:
```
U
 U
  Un
    Unit l
          Unit lost.
```

Another way is to play a new notification only if there's none of the same playing at the moment:
```
Unit lost.
          Unit lost.
```

The above problem also applies to unit voices and it's solved in the first manner (stopping the sound and starting again). But that results in a stutter "A-A-Affirmative", "U-U-Un-Unit lost". That's why I lean more towards the latter solution (playing the audio in full before starting a new one) :thinking: I wonder how other games handle that.

NinovanderMark:
@dragunoff My suggestion would be to have this depend on who initiates the triggering of the sample.

If they player does this directly by ordering a unit, or building something, starting the sample from the start is good auditory feedback that their command is registered. This is how the old C&C games used to do it.

If the trigger is something happening in the game, its function is to aid the player's situational awareness and restarting the sample doesn't add anything. We don't need to know exactly how many units are dying, only that a battle is happening somewhere.

PR #9605, opened 2015-10-11
Added support for speech variants to CanPowerDown

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/9602 and cleans up the surroundings.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Power/CanPowerDown.cs
+++ b/OpenRA.Mods.Common/Traits/Power/CanPowerDown.cs
@@ -24,6 +24,12 @@ namespace OpenRA.Mods.Common.Traits
 
 		[PaletteReference] public readonly string IndicatorPalette = "chrome";
 
+		public readonly string PowerupSound = null;
+		public readonly string PowerdownSound = null;
+
+		public readonly string PowerupSpeech = null;
+		public readonly string PowerdownSpeech = null;
+
 		public override object Create(ActorInitializer init) { return new CanPowerDown(init.Self, this); }
 	}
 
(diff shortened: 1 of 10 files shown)


OpenRA.Mods.Common/Traits/Power/CanPowerDown.cs
```diff
@@ -20,9 +20,13 @@ public class CanPowerDownInfo : UpgradableTraitInfo, Requires<PowerInfo>
 		public readonly bool CancelWhenDisabled = false;
 
 		public readonly string IndicatorImage = "poweroff";
-		public readonly string IndicatorSequence = "offline";
+		[SequenceReference("IndicatorImage")] public readonly string IndicatorSequence = "offline";
 
-		public readonly string IndicatorPalette = "chrome";
+		[PaletteReference] public readonly string IndicatorPalette = "chrome";
+
+		public readonly string NotificationType = "Sounds";
```
pchote: It would be better to define separate PowerdownNotification vs PowerdownSounds (and similar for Powerup) for consistency with our other traits that can use sounds and/or notifications.

pchote: This should be fixed in PlaceBeacon too.  Having a Notification that might actually be a Sound breaks our naming conventions, and limits the configuration options (mods might want to use both a notification and a sound effect).

Mailaender: The inconsistency in our code base isn't nice, but I think we should not hard-code it, but have it configurable and documented everywhere. Same as we do the sprite image/sequence setup and then run a lint test with reflection.

pchote: Not sure what you mean there. My request makes this *more* configurable than you currently have it, and making it match all the other traits helps with documentation.

Mailaender: I mean hard-coding it and having lot's of boiling plate code spaghetti code like:

```csharp
if (Info.PowerupNotification != null && disabled)
	Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Notification", PowerdownNotification, self.Owner.Faction.InternalName);

if (Info.PowerupNotification != null && !disabled)
	Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Notification", PowerupNotification, self.Owner.Faction.InternalName);

if (Info.PowerupSpeech != null && disabled)
	Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Speech", PowerdownSpeech, self.Owner.Faction.InternalName);

if (Info.PowerupSpeech != null && !disabled)
	Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Speech", PowerupSpeech, self.Owner.Faction.InternalName);
```

is not improving anything.

pchote: This isn't a problem in literally every other trait (except PlaceBeacon) that uses speech and/or notifications.

In the current pr it is not possible to play a powerdown/powerup sound (like machinery turning on/off) at the structure you are toggling at the same time as playing an EVA notification.  The changes that I have requested makes this situation doable – an improvement – as well as bringing the code in line with the rest of our traits.

Mailaender: This is a theoretical feature that won't even work due to https://github.com/OpenRA/OpenRA/issues/4690 but I want get this through.

Mailaender: My preference to go for the PlaceBeacon style everywhere.

abcdefg30:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/838653d408dcff582a4322c83fb71cc1b2c856b0...cea6ae7e41e426ec2338feab0fa1657988340bb0)

PR #9606, opened 2015-10-11
Add ReloadAmmo trait, remove Rearm activity and rename Repairable

reaperrr:
This is the first out of 2 (maybe 3) PRs to replace the `Rearm` and `Repair` activities with upgrade-based solutions (#9613).

Note1: Although the commits should work on their own, the later commits make improvements over the first in some cases, keep that in mind while reviewing.

Note2: I'll write upgrade rules for both the `Repairable` -> `Resupplyable` rename and the `AmmoPool.SelfReloads` -> `ReloadAmmo` once I'm sure there won't be further changes to trait names or larger changes requested.


reaperrr:
Updated.

abcdefg30:
In RA, helis/planes don't rearm on their pad/airfield now.

reaperrr:
> In RA, helis/planes don't rearm on their pad/airfield now.

Minelayers, too.
It looks like (right from the first commit, seems some 'fixup' turned into a 'breakup') things break as soon as `ReloadsAmmo` requires an upgrade, but I have no clue why. Probably something about how I look up the `ReloadsAmmo` trait in `AmmoPool`, but so far I haven't been able to pin it down.

atlimit8:
Needs a rebase.

reaperrr:
Rebased and adressed @abcdefg30's comment.

Still don't know why reloading is broken, though. I might need help with that.

OpenRA.Mods.Common/Traits/AmmoPool.cs
```diff
@@ -105,23 +101,40 @@ public void Attacking(Actor self, Target target, Armament a, Barrel barrel)
...
 			// Resets the tick counter if ammo was fired.
-			if (Info.ResetOnFire && GetAmmoCount() < PreviousAmmo)
+			if (reload.Info.ResetOnFire && GetAmmoCount() < PreviousAmmo)
 			{
-				RemainingTicks = Info.SelfReloadTicks;
+				RemainingTicks = reload.Info.ReloadTicks;
 				PreviousAmmo = GetAmmoCount();
 			}
 
 			if (!FullAmmo() && --RemainingTicks == 0)
```
obrakmann: Reloading aircraft started working for me when I changed this to `... && --RemainingTicks <= 0)`

OpenRA.Mods.Common/Traits/AmmoPool.cs
```diff
@@ -71,7 +61,13 @@ public AmmoPool(Actor self, AmmoPoolInfo info)
 			else
 				CurrentAmmo = Info.Ammo;
 
-			RemainingTicks = Info.SelfReloadTicks;
+			reload = self.TraitsImplementing<ReloadAmmo>()
+				.Where(a => a.Info.ReloadAmmoPools.Contains(Info.Name))
+				.FirstOrDefault();
```
obrakmann: This line here is the reason why the minelayer doesn't reload. In the minelayer's yaml, the `ReloadsAmmo` trait comes after `AmmoPool`. You need to implemented `INotifyCreated` move this line into the `Created()` method.

OpenRA.Mods.Common/Traits/AmmoPool.cs
WolfGaming: Yeah it seems to decrement RemainingTicks even if !FullAmmo() is false, so if !FullAmmo stays false for a while then RemainingTicks eventually goes below 0.

reaperrr:
Updated.
TODO:
- [ ] adapt Aircraft to use ReloadAmmo.RearmBuildings instead of its own property
- [ ] write upgrade rules
- [ ] possibly things I might have overlooked

pchote:

Issue #9592, opened 2015-10-11
Tanya can't easily target a bridge to use C4


steelphase:
there only seems to be specific locations where you can place the c4 on the bridge. this only seems to affect bleed.

Mailaender:
Probably the same underlying problem as https://github.com/OpenRA/OpenRA/issues/9192.

steelphase:
#9327 looks to be the first commit where this breaks, just adding this for memory sake.

Issue #9192, opened 2015-08-29
Long bridge engineer repair is broken


Mailaender:
It seems the path finder does not allow it to enter the invisible bridge hut. Test map: Poland raid.

Mailaender:
On coastal influence it only works from the north side, but never from the south side.

reaperrr:
Maybe the placement of the invisible bridge hut is the problem. We should test whether #9327 fixes this.

Mailaender:
I tested it and sadly it doesn't. Something else must have regressed here. I assume the method which decides which bridge tiles and bridge huts belong together. Pretty sure that used to work.

reaperrr:
This was actually introduced in #8639, more specifically 26ce7b5.

PR #9607, opened 2015-10-11
Fixed long bridge demolition

Mailaender:
Reverts https://github.com/OpenRA/OpenRA/commit/c23ee1be2e58bcc512ceb106d56a619bab48eb33#diff-0e19aa72642d1c8ffcf6cc9c47e4471c.

Fixes https://github.com/OpenRA/OpenRA/issues/9592.

### DIFF
--- a/mods/ra/rules/civilian.yaml
+++ b/mods/ra/rules/civilian.yaml
@@ -493,8 +493,9 @@ BRIDGEHUT:
 	Building:
 		Footprint: __ __
 		Dimensions: 2,2
-	CustomSelectionSize:
-		CustomBounds: 48,48
+	Selectable:
+		Bounds: 48,48
+		Priority: 2
 	BridgeHut:
 	Targetable:
 		TargetTypes: BridgeHut, C4
@@ -504,8 +505,9 @@ BRIDGEHUT.small:
 	Building:
 		Footprint: _
 		Dimensions: 1,1
-	CustomSelectionSize:
-		CustomBounds: 24,24
+	Selectable:
+		Bounds: 24,24
+		Priority: 2
 	BridgeHut:
 	Targetable:
 		TargetTypes: BridgeHut, C4


steelphase:
:+1: this resolve the bridge issue for Tanya. It also looks like the bridges render correctly at the edge of the screen.

reaperrr:
Ultimately this means `CustomSelectionSize` is not working properly, so we should at least open a ticket for that.

steelphase:
i also don't think this resolves #9192. I still can't use an engineer to repair the bridge.

reaperrr:
Same here. #9592 works, #9192 still doesn't.

reaperrr:
Actually, #9592 already works with `CustomSelectionSize` for me.

Mailaender:
Can you give me the name and a screenshot of the bridge, because I tested this and it worked.

reaperrr:
Any bridge on the Burlesca map in RA mod (didn't notice any differences between them).

pchote:
The bridges in Burlesca are bogus - one of the end tiles on the bridge is missing (and theres a road tile there instead).  This confuses the bridge logic, and this appears to confuse the engineer repairing code.  The same issue exists in the last release.

pchote:
The southern ends of the two bridges in Styrian Mountains also won't repair, and also have missing tiles (and also had the same problem in the last release).

pchote:
Confirmed that this fixes #9592, but i'm removing the reference to #9192, which predates the Selectable change.

pchote:
Unfortunately this does not address the real problem, it just hides it.  So i'll have to :-1: it in its current form.

The real regression for the bridge issues was introduced by 26ce7b5e1c3e6f8c7eaf359e14faaedf9954e470, not c23ee1b.

pchote:
I have identified the real issue, and will file a pr later tonight.

PR #8861, opened 2015-07-31
Add `nunit` NUnit make target to Makefile.

atlimit8:
This allows for overrides from the `make nunit ...` command line:
- Use NUNIT_CONSOLE if nunit[2]-console is not in bin search paths
- Use NUNIT_LIBS_PATH if NUnit libs are not in search paths. Include trailing /
- Use NUNIT_LIBS if NUnit libs have different names (such as including a prefix or suffix)

Closes https://github.com/OpenRA/OpenRA/issues/8849.

Issue #8849, opened 2015-07-30
Completely replace OpenRA.Test with OpenRA.Utility.exe or Support It


atlimit8:
It seems that `OpenRA.Test` does not provide much functionality and I have issues with running it on my system:
`dmcs -t:library -out:OpenRA.Test.dll OpenRA.Test/OpenRA.Game/*.cs -r:mods/common/OpenRA.Mods.Common.dll -r:OpenRA.Game.exe -r:nunit.framework.dll -r:System.dll -r:System.Core.dll -r:System.Data.dll -r:System.Data.DataSetExtensions.dll -r:System.Drawing.dll -r:System.Xml.dll -r:thirdparty/download/ICSharpCode.SharpZipLib.dll -r:thirdparty/download/FuzzyLogicLibrary.dll -r:thirdparty/download/Mono.Nat.dll -r:thirdparty/download/MaxMind.Db.dll -r:thirdparty/download/MaxMind.GeoIP2.dll -r:thirdparty/download/Eluant.dll && nunit-console OpenRA.Test.dll`
results in:
```
OpenRA.Test/OpenRA.Game/ActorInfoTest.cs(67,9): warning CS0219: The variable `i' is assigned but its value is never used
OpenRA.Test/OpenRA.Game/ActorInfoTest.cs(88,9): warning CS0219: The variable `i' is assigned but its value is never used
Compilation succeeded - 2 warning(s)
NUnit version 2.4.8
Copyright (C) 2002-2007 Charlie Poole.
Copyright (C) 2002-2004 James W. Newkirk, Michael C. Two, Alexei A. Vorontsov.
Copyright (C) 2000-2002 Philip Craig.
All Rights Reserved.

Runtime Environment - 
   OS Version: Unix 4.1.3.2
  CLR Version: 4.0.30319.17020 ( 4.0.2 (Stable 4.0.2.5/c99aa0c Mon Jul 13 18:57:35 UTC 2015) )

Symbol file /tmp/nunit20/ShadowCopyCache/8873_635738788612624820/Tests/assembly/shadow/0f6609dd/157c8393_1a1a8a4e_00000001/OpenRA.Test.dll.mdb doesn't match image (null)

Tests run: 0, Failures: 0, Not run: 0, Time: 0.014 seconds


```
That probably is something with my system, but it is another possible dependency / point of failure.

pchote:
[...]

PR #9610, opened 2015-10-11
Tab completion improvement/rewrite

ChaoticMind:
Tab completion can now properly handle elements (most likely usernames) with whitespace. Previously, if someone had a nickname "I like lamp", and one types "I li+tab", it wouldn't work. Another pet peeve tackled!

I'm not super satisfied with it because I would have liked to avoid rebuilding `relevantSubstrings`, but I didn't really find a more elegant solution. (Suggestions welcome). It might be impossible though because of the immutability of strings and the not-worthiness of having a "startsWith()" helper function that works on char arrays.
It's O(n*m) (worst case) where m is the average length of the possible elements (nicknames).

I rewrote a good section of the class to increasing modularity and reduce the need to handle corner cases (although some remnants linger). Commands and Names are now in different `TabCompletionLogic` objects.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/TabCompletionLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/TabCompletionLogic.cs
@@ -20,60 +20,89 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 		IList<string> candidates = new List<string>();
 		int currentCandidateIndex = 0;
 		string lastCompleted;
+		public readonly string PrefixOnStart;
+		public readonly string SuffixOnStart;
 		string prefix;
 		string suffix;
 
-		public IList<string> Commands { get; set; }
+		int longestElementLength;
+		IList<string> completions;
+		public IList<string> Completions
+		{
+			get
+			{
+				return completions;
+			}
 
-		public IList<string> Names { get; set; }
+			set
+			{
+				completions = value;
+				longestElementLength = completions.Max(element => element.Length);
+			}
+		}
+
+		public TabCompletionLogic(string prefixOnStart = "", string suffixOnStart = ": ")
+		{
+			this.PrefixOnStart = prefixOnStart;
+			this.SuffixOnStart = suffixOnStart;
+		}
 
 		public string Complete(string text)
 		{
-			if (string.IsNullOrWhiteSpace(text))
+			if (string.IsNullOrEmpty(text) || completions == null)
 				return text;
 
-			if (lastCompleted == text && candidates.Any())
+			// Check if tab is cycled without text modification
+			if (lastCompleted == text)
 			{
+				if (!candidates.Any())
+					return text;
+
 				lastCompleted = prefix + candidates[++currentCandidateIndex % candidates.Count] + suffix;
 				return lastCompleted;
 			}
 
-			var toComplete = "";
-			if (text.StartsWith("/") && Commands != null)
-			{
-				prefix = "/";
-				suffix = "";
-				toComplete = text.Substring(1);
-				candidates = Commands.Where(x => x.StartsWith(toComplete, StringComparison.InvariantCultureIgnoreCase)).ToList();
-			}
-			else if (Names != null)
-			{
-				var oneWord = text.Contains(' ');
-				if (oneWord)
-				{
-					prefix = text.Substring(0, text.LastIndexOf(' ') + 1);
-					suffix = "";
-					toComplete = text.Substring(prefix.Length);
-				}
-				else
-				{
-					prefix = "";
-					suffix = ": ";
-					toComplete = text;
-				}
-
-				candidates = Names.Where(x => x.StartsWith(toComplete, StringComparison.InvariantCultureIgnoreCase)).ToList();
-			}
+			GenerateCompletionCandidates(text);
+			if (candidates.Count == 0)
+				lastCompleted = text;
 			else
-				return text;
+				lastCompleted = prefix + candidates[currentCandidateIndex] + suffix;  // Alternatively, we could fill the shortest common substring of candidates
+			return lastCompleted;
+		}
+
+		void GenerateCompletionCandidates(string text)
+		{
+			suffix = "";
+			candidates = new List<string>();
+			var relevantSubstrings = Enumerable.Range(1, Math.Min(text.Length, longestElementLength)).Select(i => text.Substring(text.Length - i));
+
+			prefix = "";
+			var possibleCandidates = new List<Tuple<string, int>>();
+			foreach (string x in completions)
+			{
+				var firstValid = relevantSubstrings.Take(x.Length).Reverse().FirstOrDefault(e => x.StartsWith(e, StringComparison.InvariantCultureIgnoreCase));
+
+				if (!string.IsNullOrEmpty(firstValid))
+				{
+					possibleCandidates.Add(Tuple.Create(x, firstValid.Length));
+				}
+			}
+
+			if (possibleCandidates.Any())
+			{
+				var maxMatch = possibleCandidates.Max(e => e.Item2);  // Longest common substring length
+				candidates = possibleCandidates.Where(e => e.Item2 == maxMatch).Select(e => e.Item1).ToList();
+				prefix = text.Substring(0, text.Length - maxMatch);
+			}
+
+			// Add prefix/suffix if first word
+			if (prefix == "")
+			{
+				prefix = PrefixOnStart;
+				suffix = SuffixOnStart;
+			}
 
 			currentCandidateIndex = 0;
-
-			if (candidates.Count == 0)
-				return text;
-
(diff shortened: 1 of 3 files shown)


ChaoticMind:
This class is a nice candidate to write tests on because it has some annoying corner cases that we want to make sure are still working if modifications are made (e.g. one character tab completion, input longer than longest tab completion object, etc). And it's a fairly isolated object with simple string inputs/outputs. I couldn't however figure out how to set up the tests with mono (cmdline linux), any tips?
@RoosterDragon, maybe you can help?

RoosterDragon:
I don't think we support that yet sadly: #8849.

Mailaender:
Better don't run it on the command line. You are not a CI bot. Use @MonoDevelop: https://github.com/OpenRA/OpenRA/issues/8849#issuecomment-126804827

OpenRA.Mods.Common/Widgets/Logic/TabCompletionLogic.cs
```diff
@@ -19,58 +19,83 @@ public class TabCompletionLogic
...
+		{
+			get
+			{
+				return universe;
+			}
+
+			set
+			{
+				universe = value;
+				longestElementLength = universe.Aggregate("", (max, cur) => max.Length > cur.Length ? max : cur).Length;
```
RoosterDragon: `universe.Max(element => element.Length)` should suffice.

OpenRA.Mods.Common/Widgets/Logic/TabCompletionLogic.cs
```diff
@@ -19,58 +19,83 @@ public class TabCompletionLogic
...
 		int currentCandidateIndex = 0;
 		string lastCompleted;
+		public readonly string PrefixOnStart;
+		public readonly string SuffixOnStart;
 		string prefix;
 		string suffix;
 
-		public IList<string> Commands { get; set; }
+		private int longestElementLength;
+		private IList<string> universe;
```
RoosterDragon: Our style guide says the default accessibility modifier should be omitted. i.e. You should remove the explicit `private` here.

OpenRA.Mods.Common/Widgets/Logic/TabCompletionLogic.cs
```diff
@@ -19,58 +19,83 @@ public class TabCompletionLogic
...
+
+			set
+			{
+				universe = value;
+				longestElementLength = universe.Aggregate("", (max, cur) => max.Length > cur.Length ? max : cur).Length;
+			}
+		}
 
-		public IList<string> Names { get; set; }
+		public TabCompletionLogic(string prefixOnStart = "", string suffixOnStart = ": ")
```
RoosterDragon: I feel like the empty string would be a better default for the suffix, too.

Phrohdoh: Most clients I've used either append ": " or ", " so one of those two is what I vote for.

atlimit8:
> I couldn't however figure out how to run the current tests with mono (command line on Linux)

See #8861 (reopened)

atlimit8:
> Better don't run it on the command line. You are not a CI bot. Use @MonoDevelop: #8849 (comment)

@Mailaender and what would you put in `git rebase -i --exec 'make ...' bleed` to run the unit tests?

OpenRA.Mods.Common/Widgets/Logic/TabCompletionLogic.cs
RoosterDragon: To be clear, I just mean the default for `TabCompletionLogic` constructor should be an empty string. The caller should explicitly pass the suffix they're after.

ChaoticMind:
Rebased and updated (bugfix)

Covers the following cases, for both starting a string with a nick or not:

1. nicks with and without space
2. nicks with common substring (space or not)
3. nicks with repeating substring (take longest match) (also both with and without space)

Example for the third point, if two nicks are "cabcabal" and "cabal", typing "cabca + tab" will only match the first nick. This is important to cover because it would otherwise get confusing in cases of nicks with spaces.

I had to keep track of "longest match" for each hit, which I did in a Tuple<string, int> (the `string` is the proposed candidate/nick, and the `int` is the length of the longest common substring between the candidate and the input text). I then filtered on the longest match (largest second element).

Another way would be a dictionary of lists, indexed by the "longest common substring length" and then only take the list for the highest index. Unsure if that would be more readable, probably equivalent.

OpenRA.Mods.Common/Widgets/Logic/TabCompletionLogic.cs
```diff
@@ -20,60 +20,91 @@ public class TabCompletionLogic : ChromeLogic
...
 		string lastCompleted;
+		public readonly string PrefixOnStart;
+		public readonly string SuffixOnStart;
 		string prefix;
 		string suffix;
 
-		public IList<string> Commands { get; set; }
+		int longestElementLength;
+		IList<string> universe;
+		public IList<string> Universe
```
pchote: What is the significance of this name?

ChaoticMind: Set theory [nomenclature](https://en.wikipedia.org/wiki/Universe_%28mathematics%29), seems appropriate.

pchote: This is potentially confusing here, since we are simulating `Actor`s, and `World`s.  Long ago we were going to have a `Universe` class that wrapped the concept of the world and other bits.

atlimit8: `Pool` or `Completions` should work.

pchote: `Completions` sounds good to me.  It is descriptive and unambigous.

ChaoticMind: Sure thing, I'll wait for the rest of the review and change it.

OpenRA.Mods.Common/Widgets/Logic/TabCompletionLogic.cs
```diff
@@ -20,60 +20,91 @@ public class TabCompletionLogic : ChromeLogic
...
+		}
 
 		public string Complete(string text)
 		{
-			if (string.IsNullOrWhiteSpace(text))
+			if (string.IsNullOrEmpty(text) || Universe == null)
 				return text;
 
-			if (lastCompleted == text && candidates.Any())
+			// check if tab is cycled without text modification
```
abcdefg30: Start comments with an uppercase letter, please.

ChaoticMind:
Updated, rebased, and squashed.

OpenRA.Mods.Common/Widgets/Logic/TabCompletionLogic.cs
```diff
@@ -20,60 +20,91 @@ public class TabCompletionLogic : ChromeLogic
...
+			return lastCompleted;
+		}
+
+		void GenerateCompletionCandidates(string text)
+		{
+			suffix = "";
+			candidates = new List<string>();
+			var relevantSubstrings =
+				from i in Enumerable.Range(1, Math.Min(text.Length, longestElementLength))
+				select text.Substring(text.Length - i);
```
Mailaender: > [17:40] pchote: functional syntax, please

Mailaender:

OpenRA.Mods.Common/Widgets/Logic/TabCompletionLogic.cs
ChaoticMind: fixed

RoosterDragon:
Any chance of some unit tests for this now  #8861 is merged?

atlimit8:
Considering the concern above, it would be nice for the unit tests to be added now rather than later or possibly never. @ChaoticMind use `make nunit` to run NUnit tests in a Linux terminal.

abcdefg30:
Needs a rebase.

obrakmann:
Looks good for the most part. I found one issue: when you type "/og", for example, and press tab, it will auto-complete to "/ogiveall", which is obviously wrong.

Phrohdoh:
@ChaoticMind Are you still working on this?

ChaoticMind:
I haven't worked on this since last commit. It's still on my radar but I can't do it before early February. If you prefer to close it and reopen it when I push an update, that's OK for me. Sorry :(

@obrakmann Yeah, I was planning on making that adjustment in a PR after this one (to keep them smaller). But I guess I may as well just include it in this one.

pchote:
Any updates here?

reaperrr:
Since there hasn't been any activity for two months, we're closing this to free the PR queue.
@ChaoticMind fell free to reopen it when you get around to update it.

PR #9611, opened 2015-10-11
Change the default stance to Defend for human players

### LINKED ISSUE #3285: Change default unit stance to defend for land units and returnfire for cloaked units [$48]
Nobody actually wants them to default to AttackAnything.

<bountysource-plugin>
---
There is a **[$48 open bounty](https://www.bountysource.com/issues/335981-change-default-unit-stance-to-defend-for-land-units-and-returnfire-for-cloaked-units?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
Exceptions are dogs and stealthy things.

Closes #3285.

Thanks to @matija-hustic for laying the groundwork that makes this work properly in the first place!

### DIFF
--- a/OpenRA.Mods.Common/Traits/AutoTarget.cs
+++ b/OpenRA.Mods.Common/Traits/AutoTarget.cs
@@ -24,8 +24,12 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Set to a value >1 to override weapons maximum range for this.")]
 		public readonly int ScanRadius = -1;
 
-		[Desc("Possible values are HoldFire, ReturnFire, Defend and AttackAnything.")]
-		public readonly UnitStance InitialStance = UnitStance.AttackAnything;
+		[Desc("Possible values are HoldFire, ReturnFire, Defend and AttackAnything.",
+			"Used for computer-controlled players, both Lua-scripted and regular Skirmish AI alike.")]
+		public readonly UnitStance InitialStanceAI = UnitStance.AttackAnything;
+
+		[Desc("Possible values are HoldFire, ReturnFire, Defend and AttackAnything. Used for human players.")]
+		public readonly UnitStance InitialStance = UnitStance.Defend;
 
 		[Desc("Allow the player to change the unit stance.")]
 		public readonly bool EnableStances = true;
@@ -63,7 +67,7 @@ namespace OpenRA.Mods.Common.Traits
 		{
 			this.info = info;
 			attack = self.Trait<AttackBase>();
-			Stance = info.InitialStance;
+			Stance = self.Owner.IsBot || !self.Owner.Playable ? info.InitialStanceAI : info.InitialStance;
 			PredictedStance = Stance;
 			at = self.TraitOrDefault<AttackFollow>();
 		}
--- a/OpenRA.Mods.Common/Traits/AutoTarget.cs
+++ b/OpenRA.Mods.Common/Traits/AutoTarget.cs
@@ -16,7 +16,7 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("The actor will automatically engage the enemy when it is in range.")]
-	public class AutoTargetInfo : ITraitInfo, Requires<AttackBaseInfo>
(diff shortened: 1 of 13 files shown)


mods/cnc/rules/vehicles.yaml
```diff
@@ -151,6 +151,7 @@ ARTY:
 	WithMuzzleOverlay:
 	AutoTarget:
 		InitialStance: Defend
```
abcdefg30: This can be removed then?

mods/cnc/rules/vehicles.yaml
```diff
@@ -491,6 +492,7 @@ MLRS:
 		AmmoPoolName: primary
 	AutoTarget:
 		InitialStance: Defend
```
abcdefg30: Same here.

pchote:
I'm still very wary about this, but we can always revert it if (when?) our non-pro players complain / file confused bug reports about broken unit AI.

obrakmann:
Updated. The date in the upgrade rules will probably need to be changed before merging.

pchote:
Adding to the milestone because if we don't take this before the first playtest then I will insist that we leave it to the next cycle.  Slipping this in later in the testing cycle is asking for trouble.

obrakmann:
Rebased onto bleed, also dropped the commit that set the `Playable` flag on the scripted players to `false`, since that's already the default.

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -24,8 +24,11 @@ public class AutoTargetInfo : ITraitInfo, Requires<AttackBaseInfo>
 		[Desc("Set to a value >1 to override weapons maximum range for this.")]
 		public readonly int ScanRadius = -1;
 
-		[Desc("Possible values are HoldFire, ReturnFire, Defend and AttackAnything.")]
-		public readonly UnitStance InitialStance = UnitStance.AttackAnything;
+		[Desc("Possible values are HoldFire, ReturnFire, Defend and AttackAnything. Used for computer-controlled players.")]
```
pchote: Could you clarify this a bit wrt mission players?

pchote:
Code changes look good, but I haven't tested this so :+0.5:.

OpenRA.Mods.Common/Traits/AutoTarget.cs
obrakmann: Hmm, like this? Not sure what kind of clarification you're asking for. I updated the `Desc` and renamed the variable to `InitialStanceAI`.

pchote: That's better, thanks.

penev92:
You missed fremen with the "stealthy things".

Other than that :+1:

obrakmann:
Added default stance for fremen. Also changed the stances of all stealthy units to "HoldFire" for human players and "ReturnFire" for the AI, as discussed in IRC.

mods/ra/maps/fort-lonestar/map.yaml
```diff
@@ -654,7 +654,7 @@ Rules:
 		Health:
 			HP: 200
 		AutoTarget:
-			InitialStance: Defend
+			InitialBotStance: Defend
```
Mailaender: I guess this has to be `InitialStanceAI` to match the code although `InitialBotStance` has better readability.

obrakmann: Well spotted! I only ran sed on the rules directory and forgot about this.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/9de9115f976ea6a64baeb94602ddde442fd785e7...8a09b406e165173328c7edc4c0f60de77828cc4a)

PR #3467, opened 2013-06-23
AutoTarget unarmed buildings again to help AI crush it's enemy

Mailaender:
This reverts #3337 for ra and the changes introduced by @psydev for d2k/cnc in that regard. The AI needs AutoTarget because it solely uses AttackMove to fight it's enemies.

PR #9615, opened 2015-10-11
Remaining D2k changes to restore original balancing/gameplay

reaperrr:
#9458 + this should wrap up the implementation of @OmegaBolt's PR.

Don't let the number of commits intimidate you, I split everything I could into individual commits for easier reviewing.

### DIFF
--- a/mods/d2k/rules/arrakis.yaml
+++ b/mods/d2k/rules/arrakis.yaml
@@ -31,12 +31,12 @@ sandworm:
 		Name: Sandworm
 		Description: Attracted by vibrations in the sand.\nWill eat units whole and has a large appetite.
 	Health:
-		HP: 1200
-		Radius: 3
+		HP: 9990
+		Radius: 256
 	Armor:
 		Type: None
 	Mobile:
-		Speed: 50
+		Speed: 42
 		TerrainSpeeds:
 			Sand: 100
 			Dune: 100
--- a/mods/d2k/rules/defaults.yaml
+++ b/mods/d2k/rules/defaults.yaml
@@ -236,6 +236,8 @@
 	SelectionDecorations:
 	Selectable:
 		Priority: 2
+	RevealsShroud:
+		VisibilityType: CenterPosition
 	Targetable:
 		TargetTypes: Ground, C4, Structure
 	Building:
--- a/mods/d2k/rules/defaults.yaml
+++ b/mods/d2k/rules/defaults.yaml
@@ -254,7 +254,7 @@
 	WithBuildingExplosion:
 	RepairableBuilding:
 	EmitInfantryOnSell:
-		ActorTypes: light_inf,light_inf,light_inf,light_inf,light_inf,trooper,trooper,trooper,engineer
+		ActorTypes: light_inf
 	MustBeDestroyed:
 		RequiredForShortGame: true
 	FrozenUnderFog:
--- a/mods/d2k/rules/defaults.yaml
+++ b/mods/d2k/rules/defaults.yaml
@@ -175,7 +175,7 @@
 			Rough: 80
 	SelectionDecorations:
 	Selectable:
-		Bounds: 12,18,0,-6
+		Bounds: 12,20,0,-4
 	Targetable:
 		TargetTypes: Ground, Infantry
 	QuantizeFacingsFromSequence:
--- a/mods/d2k/rules/defaults.yaml
+++ b/mods/d2k/rules/defaults.yaml
@@ -124,13 +124,10 @@
 	Targetable:
 		TargetTypes: Ground, Vehicle
 		RequiresForceFire: yes
-	Capturable:
-		Type: husk
-		AllowAllies: yes
-		CaptureThreshold: 1.0
-	TransformOnCapture:
-		ForceHealthPercentage: 25
 	DisabledOverlay:
+	Explodes:
+		Weapon: UnitExplodeMed
+		EmptyWeapon: UnitExplodeMed
 
 ^AircraftHusk:
 	Inherits: ^Husk
--- a/mods/d2k/rules/defaults.yaml
+++ b/mods/d2k/rules/defaults.yaml
@@ -273,8 +273,8 @@
 	DamagedWithoutFoundation:
 	ThrowsShrapnel:
 		Weapons: Debris, Debris2, Debris3, Debris4
-		Pieces: 3, 7
-		Range: 2c0, 5c0
+		Pieces: 2, 5
+		Range: 1c512, 4c0
 	WithMakeAnimation:
 
 ^Defense:
@@ -291,9 +291,9 @@
 	LineBuildNode:
 		Types: turret
 	ThrowsShrapnel:
-		Weapons: Debris, Debris2, Debris3, Debris4
-		Pieces: 2, 5
-		Range: 2c0, 5c0
+		Weapons: Debris, Debris3
+		Pieces: 2, 2
+		Range: 2c0, 4c0
 	MustBeDestroyed:
 		RequiredForShortGame: false
 
--- a/mods/d2k/rules/defaults.yaml
+++ b/mods/d2k/rules/defaults.yaml
@@ -22,22 +22,22 @@
 			1600: firepower, damage, speed, reload, inaccuracy, rank, eliteweapon, selfheal
 	FirepowerMultiplier@EXPERIENCE:
 		UpgradeTypes: firepower
-		Modifier: 105, 110, 120, 130
+		Modifier: 105, 110, 115, 125
 	DamageMultiplier@EXPERIENCE:
 		UpgradeTypes: damage
-		Modifier: 95, 90, 85, 75
+		Modifier: 96, 92, 88, 80
 	SpeedMultiplier@EXPERIENCE:
 		UpgradeTypes: speed
-		Modifier: 105, 110, 120, 140
+		Modifier: 105, 110, 115, 125
 	ReloadDelayMultiplier@EXPERIENCE:
 		UpgradeTypes: reload
-		Modifier: 95, 90, 85, 75
+		Modifier: 96, 92, 88, 80
 	InaccuracyMultiplier@EXPERIENCE:
 		UpgradeTypes: inaccuracy
 		Modifier: 90, 80, 70, 50
 	SelfHealing@ELITE:
 		Step: 2
-		Ticks: 100
+		Ticks: 125
 		HealIfBelow: 1
 		DamageCooldown: 125
 		UpgradeTypes: selfheal
--- a/mods/d2k/rules/structures.yaml
+++ b/mods/d2k/rules/structures.yaml
@@ -16,6 +16,7 @@
 		RemoveInstead: true
 	Buildable:
 		Queue: Building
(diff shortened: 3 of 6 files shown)


mods/d2k/rules/defaults.yaml
```diff
@@ -267,15 +273,16 @@
...
 	Guardable:
 		Range: 3c0
 	WithCrumbleOverlay:
 	Demolishable:
 	DamagedWithoutFoundation:
 	ThrowsShrapnel:
-		Weapons: Debris, Debris2, Debris3, Debris4
-		Pieces: 3, 7
-		Range: 2c0, 5c0
+		Weapons: debris, debris2, debris3, debris4
```
GraionDilach: Is this intended?

reaperrr: No, will fix it.

reaperrr:
Updated.

- left veterancy enabled, but toned down stat gains particularly at higher levels
- removed commit that disabled AllowMovement on infantry AutoTarget
- removed commit that disabled some crate actions
- fixed debris typo

OmegaBolt:
Why would you keep veterancy and allow movement?

As was discussed in some other commit, veterancy could be purely visual, which I think is the best idea personally so it doesn't interfere with balance and yet gives people that personal attachment to long lasting units.

Allowing movement from autotargeting is one of the absolute most annoying and stupid ideas to put in a game about unit control. The original games never did it, except in TS and RA2 where your anti-aircraft units would chase aircraft and that's commonly considered by the community an annoying bug. I do not want my infantry to ever chase down anything unless I've set it to guard mode, and even then it shouldn't hunt a unit across the whole map.

GraionDilach:
Regarding the latter, #9611 would make it mostly redundant.

penev92:
I'm for keeping veterancy. D2k is a mod where your units are dead meat from the moment they exit the factory anyway. Unit preservation is almost non-existent (and that's fine). So I don't think veterancy will have a big (if any) effect on the gameplay (as long as we keep the bonuses small).

mods/d2k/rules/defaults.yaml
```diff
@@ -267,6 +272,7 @@
 	ShakeOnDeath:
 	Sellable:
 		SellSounds: BUILD1.WAV
+	AutoTargetIgnore:
```
pchote: Isn't this going to break the AI?

mods/d2k/rules/arrakis.yaml
```diff
@@ -47,6 +47,7 @@ sandworm:
 		Weapon: WormJaw
 	Sandworm:
 		WanderMoveRadius: 5
+		ChanceToDisappear: 100
```
pchote: It would be better to change this in the trait instead, IMO.

mods/d2k/rules/defaults.yaml
```diff
@@ -124,13 +124,16 @@
 	Targetable:
 		TargetTypes: Ground, Vehicle
 		RequiresForceFire: yes
-	Capturable:
-		Type: husk
-		AllowAllies: yes
-		CaptureThreshold: 1.0
-	TransformOnCapture:
-		ForceHealthPercentage: 25
+#	Capturable:
```
pchote: Please don't leave commented out dead code in the rules.

mods/d2k/rules/world.yaml
```diff
@@ -35,6 +35,10 @@
 		Name: Corrino
 		InternalName: corrino
 		Selectable: false
+	Faction@Smuggler:
```
pchote: This doesn't appear to be used yet, but if you want to keep it will need entries set up in the building `FactionImages` dictionaries.

pchote: IIRC smugglers should use ordos(?) artwork for everything except their custom starport, which we currently have set up as corrino.

mods/d2k/rules/world.yaml
```diff
@@ -59,6 +63,8 @@ World:
 	ProductionQueueFromSelection:
 		ProductionPaletteWidget: PRODUCTION_PALETTE
 	WormManager:
+		Minimum: 0
+		SpawnInterval: 6000
```
pchote: The defaults could be changed instead.

mods/d2k/rules/structures.yaml
```diff
@@ -59,8 +60,9 @@ construction_yard:
 	RevealsShroud:
 		Range: 10c0
 	Production:
-		Produces: Building, Upgrade
+		Produces: Building
 	Exit:
```
pchote: I don't think we need this anymore, so this trait can be removed (it was only used for upgrades, which have now moved to the player).

mods/d2k/rules/world.yaml
penev92: That is correct.

mods/d2k/rules/defaults.yaml
```diff
@@ -222,7 +225,7 @@
 	Inherits@2: ^SpriteActor
 	AppearsOnRadar:
 		UseLocation: yes
-	HiddenUnderFog:
+	HiddenUnderShroud:
```
pchote: This will make aircraft visible to *everyone* under the fog.

mods/d2k/rules/world.yaml
reaperrr: Are smugglers the same as mercenaries? IIRC the AI-controlled faction allied to Ordos in the campaign had a different artwork for their heavy factory, too.

pchote: Nope, mercenaries were yet another faction that copied one of the base factions for everything except their custom heavy factory.

pchote: IMO it would make sense to have a dedicated PR for setting up the smuggers/mercenaries(/fremen?) factions properly.

penev92: Smugglers used Ordos units and structures and had their own Starport.
Mercenaries used Ordos units and structures and had their own Heavy Factory.
Fremen used Atreides units and structures.
House Corrino (the Imperial family) used Harkonnen units and structures and had their own Palace.

Might as well leave those for a separate PR.

mods/d2k/rules/defaults.yaml
reaperrr: The AI still attacked my buildings normally in a test game I had.

mods/d2k/rules/defaults.yaml
reaperrr: IIRC the main reason @OmegaBolt did this was that Carryalls and Ornithopters would otherwise be much harder to shoot down than in the original.

reaperrr:
Updated.

Not sure about the aircraft HiddenUnderFog, should I leave that in or remove it?

mods/d2k/rules/defaults.yaml
pchote: I assumed that it was done to make them visible to their owner under the fog.  That should be done by removing the `AlwaysVisibleStances` line below instead.

reaperrr:
Removed the aircraft visibility change. I guess we should let people play some games, if lots of them complain we can still change it.

mods/d2k/rules/defaults.yaml
pchote: Strange.  AFAIK we haven't changed anything relevant to AutoTarget or the AI since the failure from #3337 / #3467.

mods/d2k/rules/world.yaml
MustaphaTR: According to Tibed in orginal game Corrino has own CYard and War Factory. But they are same as Harkonnen ones as graphics and code. Possibly used in missions to separate them from Harkonnen ones.

penev92:
- [ ] Please include mission `atreides 02b` to the first commit as well.
 - [ ] I am now able to build any structure apart from concrete slabs without a ConYard.

mods/d2k/maps/atreides-01a/map.yaml
```diff
@@ -145,8 +147,8 @@ Rules:
 			Minimum: 1
 			Maximum: 1
 	construction_yard:
-		Production:
-			Produces: Building
+		Buildable:
+			Prerequisites: ~disabled
```
abcdefg30: Shouldn't `construction_yard` be unabuildable by default?
If so, please remove this (also from the other missions).

reaperrr:
@penev92 @abcdefg30 fixed prerequisites and missions.

pchote:
Changes look ok, and didn't notice anything that looked too bogus while spectating a bot match with these changes, so :+1:.

mods/d2k/rules/defaults.yaml
penev92: But this *does* break scripted reinforcements in the missions, which can sit idly by my ConYard, chilling.
I think it would be simplest to just not change that here. Can be done in a later PR if you want.

penev92:
Adding "Changes requested" for building `AutoTargetIgnore` breaking scripted reinforcements.

reaperrr:
Updated.

- removed AutoTargetIngore commit
- removed warhead falloff changes from debris commit, this is properly taken care of in my warhead PR now

mods/d2k/maps/atreides-01a/map.yaml
```diff
@@ -134,6 +134,8 @@ Rules:
 		-ConquestVictoryConditions:
 		MissionObjectives:
 			EarlyGameOver: true
+		Production:
+			Produces: Building
```
penev92: Is this still necessary? Selling my ConYard has an odd side-effect.
While that's not an issue, you need to just exclude Upgrades here, not move Building to the player, right?

mods/d2k/rules/arrakis.yaml
```diff
@@ -32,11 +32,11 @@ sandworm:
 		Description: Attracted by vibrations in the sand.\nWill eat units whole and has a large appetite.
 	Health:
 		HP: 1200
```
penev92: Partially related: it takes 2 trikes to kill a worm before it can eat one of them.
TibEd says Sandworm has 9990 HP, so this definitely needs a buff.

reaperrr:
Updated.

- Fixed Production issue
- Increased worm HP, but increased target rescan delay from 32 to 125 ticks and reduced search radius from 27 to 20 cells

mods/d2k/rules/structures.yaml
```diff
@@ -895,9 +895,6 @@ conyard.harkonnen:
 conyard.ordos:
 	Inherits: construction_yard
 	Buildable:
-		Queue: Building
-		BuildPaletteOrder: 1000
-		Prerequisites: ~disabled
```
Mailaender: I guess this makes the construction yard appear twice with cheats on:

image

reaperrr:
Updated, fixed the cyard double appearance.

penev92:
Adding "Changes requested" for the first commit as discussed on IRC.

reaperrr:
Updated.

obrakmann:
I'll relieve this one of its misery as well. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9107ff42356ced738edf4a126be541fc8d9b2eaa)

PR #9618, opened 2015-10-12
Increase planes' CruiseAltitude in cnc

abcdefg30:
In order to fix `A10`'s pushing `orcas`/`apaches` out of the way.
(See [Sleipnir's Stuff - TD balance thread](http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&p=292015#292015).)

PR #9619, opened 2015-10-12
Fix aircraft repulsing other aircraft when not being repulsable themselves

abcdefg30:


OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -200,6 +200,7 @@ public virtual WVec GetRepulsionForce()
 
 			var repulsionForce = self.World.FindActorsInCircle(self.CenterPosition, Info.IdealSeparation)
 				.Where(a => !a.IsDead && a.Info.HasTraitInfo<AircraftInfo>()
+					&& a.Info.TraitInfo<AircraftInfo>().Repulsable
 					&& a.Info.TraitInfo<AircraftInfo>().CruiseAltitude == Info.CruiseAltitude)
```
pchote: This now does three queries for `AircraftInfo`, which isn't very efficient.  Can you change this to something like the following?

```csharp

.Where(a =>
{
    if (a.IsDead)
        return false;
    var ai = a.Info.TraitInfoOrDefault<AircraftInfo>();
    return ai != null && ai.Repulsable && ai.CruiseAltitude == Info.CruiseAltitude;
})
```

abcdefg30:
Updated.

ScottNZ:
The current behaviour was intentional. What does this solve?

penev92:
I guess we can turn off their repulsion when trying to land? Currently having more than one helicopter and just 1 helipad means they won't rearm 99% of the time, unless micro-ed individually.

pchote:
@ScottNZ maybe intended as a more general solution to #9618?

abcdefg30:
> The current behaviour was intentional. What does this solve?

This is indeed "a more general solution to #9618".
It seems a bit odd to me too, if an actor that is not repulsable itself repulses other actors...

penev92:
:+1:

Mailaender:
Can't spot regressions :white_check_mark: and appreciate the trait lookup save.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/08771babf362d235225cc3849b1f1f8d40bd886f...8258f7f661e433103a39a60a650cc008928171d5)

PR #9623, opened 2015-10-12
More RA-Balance changes for the next playtest

Micr0Bit:
Change Hijackers speed from 56 to 85 (same speed as a medium tank and harvester)

Change Hijackers "WarnProbability" from 75 to 95 so its less likely that he gets crushed

Add the "Cloak-Trait" to the Hijacker which makes him invisible when idle ... ever Infantry-Unit can detect him along with every other "Cloak-detecting" unit or structure

Change Chinooks shroud-revealing from 12 to 10 .. so he is equally good as the Hind when it comes to scouting

### DIFF
--- a/mods/ra/rules/aircraft.yaml
+++ b/mods/ra/rules/aircraft.yaml
@@ -210,7 +210,7 @@ TRAN:
 	Armor:
 		Type: Light
 	RevealsShroud:
-		Range: 12c0
+		Range: 10c0
 		Type: CenterPosition
 	Aircraft:
 		RearmBuildings: hpad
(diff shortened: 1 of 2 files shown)


abcdefg30:
Looks good to me. :+1: 

Two minor things though:
- Can't we move the `DetectCloaked: CloakTypes: Hijacker` stuff onto `^Soldier`?
- Should hijackers be able to detect other hijackers (would be true if we did above)?

MicroJOo:
sry but im deciding to back away from this ... since i dont know what other things might get in ... adding additional weight on one of the factions ... 
if there is no communication when it comes to balance ... then im not interested doing this here

PR #9627, opened 2015-10-13
Avoid redundant copies of mix files in memory

RoosterDragon:
The mix file on disk stores all the game assets. For each asset in the file, we copy the portion of the file we need into memory. This works nicely but is a bit wasteful.

With this PR, we don't make copies, instead we just return a "view" into the overall file so we don't have to copy everything. This allows us to leave the file sitting on disk without copying whole chunks of it into memory and never letting go.

Once we load into the shellmap, we make the following memory savings:
- RA saves ~16 MiB.
- RA2 saves ~ 253 MiB.

### DIFF
--- a/OpenRA.Game/FileSystem/MixFile.cs
+++ b/OpenRA.Game/FileSystem/MixFile.cs
@@ -14,6 +14,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using OpenRA.FileFormats;
+using OpenRA.Primitives;
 
 namespace OpenRA.FileSystem
 {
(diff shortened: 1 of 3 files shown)


Mailaender:
I measured the following with KSysGuard:

* TS (bleed) 353.3 MB
* TS (this) 325.3 MB

which might just be measurement uncertainty here, but I didn't notice regressions. :white_check_mark:

Phrohdoh:
Here are some interesting results measured with OS X's Activity Monitor:

* TS (this) 342.6 MB
* TS (bleed) 290.8 MB

RoosterDragon:
@Phrohdoh: I am unable to recreate your TS figures - but I noticed that after loading the TS shellmap the memory usage is fairly high for a dozen seconds or so before dropping by 50-100 MB. Maybe if you grabbed your figures from different sides of this drop that would explain the "increase" you saw.

Phrohdoh:
All of these values came from after running the shellmap for approximately 20 seconds.
I'll try to get better numbers for you soon, as I can't explain the TS increase.

pchote:
Activity monitor tracks several different figures for memory, so you need to be careful which one you're looking at before interpreting them.

Phrohdoh:
These values came from the 'Real Mem' column.
The values in 'Mem' were ~40-50 MB less than in Real Mem.

huwpascoe:
Visual Studio profiler revealed a significant improvement in memory usage over time, difference ranging from 10s to 100s of megabytes depending on what you're doing with the program. :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/43a016d3b7d461fdba1a7d4363f61253fb3b9d7e...5e6181116f9650f54abcb22f887b2fd939970bbb)

PR #9628, opened 2015-10-13
Fix Utils.FormatTime not working properly on different game speeds

abcdefg30:


OpenRA.Mods.Common/Scripting/Global/DateTimeGlobal.cs
```diff
@@ -35,7 +39,7 @@ public int GameTime
 		[Desc("Converts the number of seconds into game time (ticks).")]
 		public int Seconds(int seconds)
 		{
-			return seconds * 25;
+			return (int)Math.Ceiling(seconds * 1000f / world.Timestep);
```
pchote: You can't use floating point math in simulation code.

pchote:
I don't this this is a good idea. Keeping the same real-time delays (mission timers etc) at different game speed risks completely breaking game balance - e.g. may not be possible to complete timed objectives at slow game speeds because everything goes slower in real time.

pchote:
For what it's worth I believe I made this same point as an argument against adding these functions in the first place, but we are stuck with them now!  I suggest instead changing the api description to include "at the default game speed" instead.

penev92:
I don't think I understand your first comment - keeping the same real-time delay on all speeds breaks the missions on speeds different than the default one, so this *is* a good idea IMO.

Sticking with this as it is means we say that any scripted timers will have to be made to work with a specific game speed, and are almost guaranteed to not work on any other speed (in most cases that would be the default one).

GraionDilach:
@pchote is right, this should not be changed. Even though the tickrate can be changed nowadays, the missions are made with a fixed amount of ticks planned in them and decoupling that into a fixed RL time would be missionbreaker. TS/RA2 hid the game speed bar on missions due to this very reason, since all timing was frame based and sometimes faster speed would demand for an insane APM at cases.

pchote:
@penev92: you have mixed up the current / proposed changes to the code.  Right now `Seconds` and the other time wrappers return a fixed number of ticks that is the same for all game speeds.  This PR makes it different for each speed.

ChaoticMind:
I agree with changing the description, and possibly even the function names (SecondsToTicks) (?). Otherwise it sounds like it's converting from ticks to seconds, to me.

obrakmann:
I discovered that changing this also causes desyncs in replays, so it's probably a bad idea to do.

What we *could* do, however, is to scale the in-game timer display with the game speed setting. So increasing the game speed would actually speed up in-game time compared to real wall-clock time (as it already happens in replays, funnily enough). That would at least remove inconsistencies between countdowns in missions and the elapsed time displayed by the game. It would also mean that  production times in the build palette tooltips would be shown as the same regardless of game speed setting.

pchote:
> It would also mean that production times in the build palette tooltips would be shown as the same regardless of game speed setting.

Most of the effort that went into the game speed PR was in making this work in the first place.  I don't really want to throw that out.

abcdefg30:
This is on hold until #9640 is merged to avoid conflicts in the lua files.

abcdefg30:
Updated.

OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs
```diff
@@ -106,7 +103,8 @@ public int RandomInteger(int low, int high)
 		[Desc("Returns the ticks formatted to HH:MM:SS.")]
 		public string FormatTime(int ticks, bool leadingMinuteZero = true)
 		{
-			return WidgetUtils.FormatTime(ticks, leadingMinuteZero, world.Timestep);
```
pchote: Replacing the `world.Timestep` here with 25 would be a simpler fix.

abcdefg30: https://github.com/OpenRA/OpenRA/pull/9628#discussion_r41915794

abcdefg30: Well nvm.

abcdefg30:
Updated.

OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs
pchote: Sorry for misleading you: this needs to be 40 (1000 / 25) instead of 25.  Timestep measures the number of milliseconds per tick, not ticks per second.

pchote:
:+1: with that fix.

abcdefg30:
Fixed.

Mailaender:
Works as promised in Survival 02. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/669cf43b6322eb58559b1d82097d38a7aba8d6c3...0952cd58cdf4db993bcfef0ef467dcd901f0fd13)

PR #9633, opened 2015-10-14
Added the movement classes of the domain index to the debug logs

Mailaender:
I was just told that the domain indices are calculated per movement class and wanted to share that insight somehow.


abcdefg30:
:+1:

obrakmann:
Well, just the number doesn't tell you much, except that there are indeed different movement classes, so the benefit is really small. Only advantage it gives is to remove confusion why the logs have the same line twice. Still, doesn't hurt anyone. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/e729607de9c66248586ff898ef45dbc14ce62b3a)

Mailaender:
I wasn't able to find a way to translate the numerical movement class into something human readable.

PR #9634, opened 2015-10-14
Fix bridge destruction and repair.

### LINKED ISSUE #9192: Long bridge engineer repair is broken

### LINKED ISSUE #9592: Tanya can't easily target a bridge to use C4

pchote:
The bridge hut spawning is sensitive to the order that the bridges were spawned on the map: it assumes that southern bridges were spawned before northern bridges, which only worked because we used to index the map in the bridge spawner in a non-standard way (column-major instead of row-major).  When 26ce7b5e1c3e6f8c7eaf359e14faaedf9954e470 fixed the map enumeration the code ended attaching bridges to the wrong ends of eachother, and this flowed on to the bridge huts being connected wrong.

This fix inverts the assumption above so that bridges are assumed to spawn north-to-south (which matches what we do), but the whole method is bogus and really should be rewritten at some point.

Fixes #9592.
Fixes #9192.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Buildings/Bridge.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/Bridge.cs
@@ -152,6 +152,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		internal void AddHut(BridgeHut hut)
 		{
+			// TODO: This method is incomprehensible and fragile, and should be rewritten.
 			if (huts[0] == huts[1])
 				huts[1] = hut;
 			if (Hut == null)
@@ -161,7 +162,7 @@ namespace OpenRA.Mods.Common.Traits
 					huts[0] = hut; // Set only first time
 				for (var d = 0; d <= 1; d++)
 					for (var b = neighbours[d]; b != null; b = b.Hut == null ? b.neighbours[d] : null)
-						b.huts[1 - d] = hut;
+						b.huts[d] = hut;
 			}
 			else
 				Hut = null;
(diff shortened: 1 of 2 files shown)


reaperrr:
:+1:

abcdefg30:
It still looks a bit off to me:
bridgerepaircursoroff

reaperrr:
That looks more like a (yaml) spawn offset issue.

Mailaender:
Long bridge repair at the northern hut is still broken on both Poland Raid and Coastal Influence:

image

Mailaender:
https://github.com/OpenRA/OpenRA/issues/9192#issuecomment-139739432
> On coastal influence it only works from the north side, but never from the south side.

You just inverted the breakage.

pchote:
The north bridge issue was caused by selection priorities.  Merged in the commit from #9607 to address that.

abcdefg30:

PR #9635, opened 2015-10-14
Speed up checks for IOccupySpace trait

RoosterDragon:
Eagerly load the trait (if it exists) in Actor, and use this reference to avoid having to perform `self.Info.HasTraitInfo<IOccupySpaceInfo>()` checks.

The main benefit is for `IsAtGroundLevel`, which is used during pathfinding for crushability checks. Therefore this saves us a lookup for every actor we need to check for crushability on our path.

### DIFF
--- a/OpenRA.Game/Actor.cs
+++ b/OpenRA.Game/Actor.cs
@@ -43,15 +43,7 @@ namespace OpenRA
 		public Rectangle Bounds { get; private set; }
 		public Rectangle VisualBounds { get; private set; }
 		public IEffectiveOwner EffectiveOwner { get; private set; }
-		public IOccupySpace OccupiesSpace
-		{
-			get
-			{
-				if (occupySpace == null)
-					occupySpace = Trait<IOccupySpace>();
-				return occupySpace;
-			}
-		}
+		public IOccupySpace OccupiesSpace { get; private set; }
 
 		public bool IsIdle { get { return currentActivity == null; } }
 		public bool IsDead { get { return Disposed || (health != null && health.IsDead); } }
@@ -69,7 +61,6 @@ namespace OpenRA
 			}
 		}
 
-		IOccupySpace occupySpace;
 		readonly IFacing facing;
 		readonly Health health;
 		readonly IRenderModifier[] renderModifiers;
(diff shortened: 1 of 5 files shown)


abcdefg30:
:+1:

OpenRA.Game/Actor.cs
```diff
@@ -96,7 +87,14 @@ internal Actor(World world, string name, TypeDictionary initDict)
 
 				Info = world.Map.Rules.Actors[name];
 				foreach (var trait in Info.TraitsInConstructOrder())
+				{
 					AddTrait(trait.Create(init));
+
+					// Some traits rely on properties provided by IOccupySpace in their initialization,
+					// so we must ready it now, we cannot wait until all traits have finished construction.
+					if (trait is IOccupySpaceInfo)
+						OccupiesSpace = Trait<IOccupySpace>();
```
penev92: Eww.

RoosterDragon: It's the same eww as before, just moved around a bit.

atlimit8: Should be something like:
```c#
	foreach (var trait in Info.TraitsInConstructOrder())
	{
		var instance = trait.Create(init);
		AddTrait(instance);

		// Some traits rely on properties provided by IOccupySpace in their initialization,
		// so we must ready it now, we cannot wait until all traits have finished construction.
		if (instance is IOccupySpace)
		{
			if (OccupiesSpace == null)
				OccupiesSpace = instance;
			else
				throw InvalidDataException("Multiple OccupySpace traits");
		}
	}
```
The trait lookup is not needed when the trait instance is created in the same block.

This is also, currently, the only way catch multiple traits implementing `IOccupySpace`.

penev92: There is actually nothing wrong with having multiple traits implementing this on the same actor.
Now that "everything is an upgrade" we even have use-cases for that.

atlimit8: Except only one with be used this way.

atlimit8: This only caches one trait (the last) the implements `IOccupySpace`.

RoosterDragon: `Trait<Blah>` will throw for duplicates too.

atlimit8: OK, but it is still an unnecessary lookup.

RoosterDragon: I don't want to duplicate the code for throwing errors on duplicates here. The lookup literally doesn't matter performance wise - it's only once per actor and that's nothing compared to creating the traits and populating the trait dictionary in the first place. Even if we really did care about the lookup, there's another 10 such lookups just below this loop to worry about as well :)

Simpler is better here.

atlimit8: Alright; I'll just revisit this when adding `ISingleOccurrenceTrait` and `IMultipleOccurrenceTrait`

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/ef71ecd1cc932737b2086a47b30e44a15ffde035...5f6c48eb5cc420feb24eef85d890536e716917e5)

Issue #4772, opened 2014-03-03
Chat window allows switching to team speak for dead players with no effect


deniz1a:
After a player is defeated and is still in the game as a spectator, the chat window still allows switching to team speak mode but what is typed is seen by everyone. This is misleading, there shouldn't be any team speak while dead.

This might be related to dead players being able to pause-spam the game.
##

rob-v:
There are various related issues to this one. According to my test, Spectator can chat with his [Team] - other spectators or [All]. Player can chat with [Team] - allied players or [All]. There is still bug reported in this issue. Defeated player can only chat with all - spectators and players with both options [Team], [All].
The bug is that [Team] is for defeated player in fact [All]. That can be funny and also bad - you think you still chat with your team, though now all see it.

Is there an agreement how to fix this? 
1. disable [Team] option so defeated player can only chat to [All]
2. change [Team] to team of spectators (can be still bad but it is better than [All])
3. ?

gdavegdave:
I think a defeated player should be able to speak to global [All] or to Spectators. He shouldn't be able to speak exclusively to his Team anymore as that could be exploited.

kyrreso:
Is it possible to tag this (future), as this and spec-ping is really wanted by many players?

dragunoff:
IMO it also makes sense to change the input field prefix for spec chat. Right now it says `Team` which can be confusing, especially in the case of defeated players. The UI should make it clear that their message is going out to `Spectators` and not their original team. I know that after sending the message there is a `[Spectators]` prefix but during typing that indication is not there.

abcdefg30:
This was fixed by #15615.

PR #9639, opened 2015-10-15
Add team chat for spectators

### LINKED ISSUE #2062: Spectator Team for Chat [$15]
Logically, spectators should be a different team so Team Chat in spectator mode would represent it.

<bountysource-plugin>
---
There is a **[$15 open bounty](https://www.bountysource.com/issues/98303-spectator-team-for-chat?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

abcdefg30:
Closes #2062.

### DIFF
--- a/OpenRA.Game/Network/UnitOrders.cs
+++ b/OpenRA.Game/Network/UnitOrders.cs
@@ -81,13 +81,13 @@ namespace OpenRA.Network
 							else
 							{
 								var player = world.FindPlayerByClient(client);
-								if (player == null) return;
-
-								if ((world.LocalPlayer != null && player.Stances[world.LocalPlayer] == Stance.Ally) || player.WinState == WinState.Lost)
+								if (player != null && ((world.LocalPlayer != null && player.Stances[world.LocalPlayer] == Stance.Ally) || player.WinState == WinState.Lost))
 								{
 									var suffix = player.WinState == WinState.Lost ? " (Dead)" : " (Team)";
 									Game.AddChatLine(client.Color.RGB, client.Name + suffix, order.TargetString);
 								}
+								else if (orderManager.LocalClient != null && orderManager.LocalClient.IsObserver && client.IsObserver)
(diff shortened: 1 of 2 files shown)


deniz1a:
When a player loses and becomes spectator, will he have the same team chat indicator as he did during game? Because that could be confusing, team chat would no longer mean the same thing.

abcdefg30:
Unfortunately it doesn't do that, but it's also another issue, I think: #4772.

pchote:
We should distinguish between spectator and team chat, in that case.

abcdefg30:
Updated. Using `(Spectators)` instead of `(Team)` now.

Mailaender:

MicroJOo:
does that mean , we want "defeated players" (therefor having fog and shroud turned off for him) ,  to still help their team by telling them in the teamchat what the enemy is doing ? 

so nobody else can read that ?

pchote:
If the existing (before this pr) behaviour does that, then I don't think we need hold this PR up over it.  We should definitely switch players from team chat to observer chat at some point, though.

MicroJOo:
true

pchote:
The defeated team-mate case is best solved by #3210:
* If a player is on a team with at least one active player, then use the team-spec mode (no vis controls, regular team chat)
* If a player is not on a team, or everyone in their team is defeated then they all drop to the full spec mode (vis controls, spec chat).

obrakmann:
:+1: works as advertised. As team-chat already doesn't work for dead players and cheating is thus not possible, I won't hold this up to require that dead players get switched to spec chat as well. #3210 indeed sounds like a better solution to that problem.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/478230707498f9c5de15e9da04514d86442816d1)

PR #9644, opened 2015-10-16
Update gdi-vehicles.yaml

hypercube33:
Changed because the thing was stronger than a Mammoth Mk 2

Previously set to 800.

Tiberian Sun value was set to 175

Changed to 175 to match Tiberian Sun. Verified other values matched Rules.ini


cjshmyr:
:+1:

hypercube33:
Tested and confirmed this works 2015-10-15 with another user. Plays well now.

penev92:
:+1:  Thanks.

Issue #9629, opened 2015-10-13
Five minute initial crate spawn delay is probably too long


pchote:
I forgot about this until watching https://www.youtube.com/watch?v=nKePUk5DBuQ, where fiveaces and @SoScared comment that they think this is a bad change.  I agree, this seems far too long.  A TD game can be half way finished by five minutes.  A 30 - 60 second (max) delay to prevent spawn-time crates would be much more reasonable.

AoAGeneral:
Originally I agreed with someone about the 5 min delay. But keeping track of the timeframes I agree with this here.

Ive actually removed the idea from my post and mentioned another idea to FiveAces and SoScared which is having an option for tournament crates. The original crates dont need to be edited. So the original crates can have things like spawning a visceroid (They want some love) or the old fashioned nuke crate (those were awesome).

While tournament crates can go from 500-1000 money give and maybe giving an extra unit while having the 30-60 second delay factor.

r34ch:
I believe I suggested disabling crates for X minutes as a simple bandaid for a deeper issue that really affects all mods - it was just more pronounced on TD due to how 'effective' dupe crates were.

I feel a better option is to group crate 'outcomes' into categories and have the crate outcomes available after (mod specific?) times

0-2 mins - cat 1
2-5 mins - cat 1, 2 (including vehicle crates)
5+ mins  - cat 1, 2, 3 (including MCV crates)

This way you can tune exactly what is dropped in the early stages of the game while still rewarding early map control.

AoAGeneral:
The dupe crates however was removed in the current release and no longer spawns.

r34ch:
RA you can still get MCV crates in the first few minutes of a game. Likewise a vehicle before you have a WF and dupe crates to boot. 5 enemy rangers on ore lord controlling most of the map LoS (and hence future crate drops) before you have even finished your WF? Invisibility for your starting APC on TD? (dunno if that one is still true to be fair)

A timing fix only works on maps of a certain size. And also doesn't reward very early map control. But I guess since we have the 'enable crates' check-box this will most likely not get looked into unless more players kick up fuss.

pchote:
In TD you can only get a MCV crate if you've lost all of your MCVs / conyards.  Has anyone thought of doing the same thing in RA?

MicroJOo:
i remember voting for "crates off" on default .... at one point ... 

now i see the results , having them "on" by default ... people constantly change the content of the crates and now you want to give them a time-limit ...

[...]

PR #9648, opened 2015-10-16
mcv from crates only when a player has no base

ghost:
completely removes the chance of getting an MCV from a crate 
in the RA mod unless the player has no base remaining

Stems from a comment from @pchote in #9629


cjshmyr:
:+1:

Issue #4873, opened 2014-03-15
Make approach direction of paratrooper plane controllable


deniz1a:
There is no way (or no easy way) of telling from which direction the paratroopers will arrive. Often times the plane flies right towards anti-air defenses and gets shot down. There should be a way to at least determine from which direction the plane approaches the drop site in order to avoid AA and make paratroopers more useful. This could be done by drawing a vector for example.

huwpascoe:
I'm taking on this bounty. Will submit a PR soon

huwpascoe:
Question: The direction of the aircraft is typically randomized. So what exactly is being asked here, an indicator of the direction after the order is sent, or a change in the random behavior?

```
if (randomize)
	dropFacing = Util.QuantizeFacing(self.World.SharedRandom.Next(256), info.QuantizedFacings) * (256 / info.QuantizedFacings);
```

UberWaffe:
A change in the random behavior. The title refers to it being controllable.
Previous discussions I have seen about this was to have it so that you first select the target position, and then select a second point.
The plane will enter the map in a straight line with point 1 and point 2, in such a way that it will first arrive at point 1, start dropping paratroopers while continuing straight towards point 2.

C&C generals particle cannon control is a good idea of what to aim for.

pchote:
We decided that a smoother direction select is to use a click-drag instead of two distinct clicks.  Mouse down like you would normally to select the target point, and then with the mouse held move the cursor in a direction, and you'll get an arrow cursor indicating the direction that the plane should come from (like our new joystick scroll on the rmb, except that doesn't have the cursors implemented yet).

penev92:
Selecting an entry point yourself is just stupid. Just saying...

huwpascoe:
Made a working prototype. PR probably ready in a few days.
![a](https://cloud.githubusercontent.com/assets/1812087/10525400/7f087752-737c-11e5-91d0-6dba85dd0b0c.png)
![b](https://cloud.githubusercontent.com/assets/1812087/10525402/80f8e60a-737c-11e5-936e-1d3ead69cfaa.png)

pchote:
Is that using two separate clicks, or click-drag?  Please see my comment above about click-drag, which makes it suitable for Airstrikes and keeps the interaction model simpler for people who don't care about direction.

huwpascoe:
@pchote yeah that way is implemented

Mouse down, up = random direction
[...]

PR #9652, opened 2015-10-17
Refactored WorldInteractionControllerWidget

huwpascoe:
Class is now much easier to read
Moved static methods to WorldUtils

Part 1 of 2 for https://github.com/OpenRA/OpenRA/issues/4873

### DIFF
--- a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
+++ b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
@@ -38,7 +38,7 @@ namespace OpenRA.Widgets
 			if (!IsDragging)
 			{
 				// Render actors under the mouse pointer
-				foreach (var u in SelectActorsInBoxWithDeadzone(World, lastMousePosition, lastMousePosition))
+				foreach (var u in WorldUtils.SelectActorsInBoxWithDeadzone(World, lastMousePosition, lastMousePosition))
 					worldRenderer.DrawRollover(u);
 
 				return;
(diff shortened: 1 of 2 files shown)


OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
```diff
@@ -23,150 +24,169 @@ public class WorldInteractionControllerWidget : Widget
...
 
 		[ObjectCreator.UseCtor]
 		public WorldInteractionControllerWidget(World world, WorldRenderer worldRenderer)
 		{
 			this.World = world;
 			this.worldRenderer = worldRenderer;
+
+			hotkeyMap = new Dictionary<Hotkey, Func<bool>>
+			{
+				{ Game.Settings.Keys.PauseKey, Pause },
```
pchote: Caching these breaks the ability to change the during a game.

huwpascoe:
fixed cache bug

OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
```diff
@@ -201,108 +211,80 @@ void ApplyOrders(World world, MouseInput mi)
...
-				if (key == Game.Settings.Keys.PauseKey && World.LocalPlayer != null) // Disable pausing for spectators
-					World.SetPauseState(!World.Paused);
-				else if (key == Game.Settings.Keys.SelectAllUnitsKey && !World.IsGameOver)
-				{
-					// Select actors on the screen which belong to the current player
-					var ownUnitsOnScreen = SelectActorsOnScreen(World, worldRenderer, null, player).SubsetWithHighestSelectionPriority();
-					World.Selection.Combine(World, ownUnitsOnScreen, false, false);
-				}
-				else if (key == Game.Settings.Keys.SelectUnitsByTypeKey && !World.IsGameOver)
+				var hotkeyMap = new Dictionary<Hotkey, Func<bool>>
```
RoosterDragon: A `switch` statement would be preferable here.

huwpascoe: C# won't let me
"Error CS0151 A switch expression or case label must be a bool, char, string, integral, enum, or corresponding nullable type"

RoosterDragon: My derp - can we have an if/else chain then? Anything to avoid creating and throwing away this dictionary on every keypress.

huwpascoe:
changed

RoosterDragon:
This is some nice cleanup.

I'm not sure on moving the helper methods to WorldUtils though - I don't think other places will get much use of out them. I can't imagine much of the code cares about selecting actors on screen, for example.

What do you think?

huwpascoe:
Rather than about reuse it's about making sense of the code. World Utils seems like the best place to put them I think.

huwpascoe:
merged with upstream, still passes.

huwpascoe:
rebased

huwpascoe:
resolved merge conflict

huwpascoe:
Split up the commit into smaller parts.

huwpascoe:
I'm still not happy with my work for this one, I'm gonna close this temporarily.

PR #9661, opened 2015-10-17
Update the UID stored in the map file after saving.

### LINKED ISSUE #9617: Map preview duplication
Map browser shows preview duplicates after each save within the editor if there were any changes made (does not apply if you just keep saving the map without making changes to it). This does not create extra files, it fills the menu with extra previews which all point to the same file.

testmap_ra

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27346913-map-preview-duplication?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #9617.

That error occurred when saving a map more than two times.  Because the UID was not being updated, it would always invalidate the first loaded preview, instead of the currently active one.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Editor/SaveMapLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Editor/SaveMapLogic.cs
@@ -145,9 +145,6 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 				map.Save(combinedPath);
 
-				// Reload map to calculate new UID
-				map = new Map(combinedPath);
-
 				// Update the map cache so it can be loaded without restarting the game
 				var classification = mapDirectories[directoryDropdown.Text];
 				Game.ModData.MapCache[map.Uid].UpdateFromMap(map, classification);
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -603,6 +603,9 @@ namespace OpenRA
 
 			// Update existing package
 			Container.Write(entries);
+
+			// Update UID to match the newly saved data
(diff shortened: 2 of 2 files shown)


OpenRA.Mods.Common/Widgets/Logic/Editor/SaveMapLogic.cs
```diff
@@ -143,10 +143,8 @@ public SaveMapLogic(Widget widget, Action<string> onSave, Action onExit, Map map
 				if (map.Uid != null && combinedPath == map.Path)
 					Game.ModData.MapCache[map.Uid].Invalidate();
 
-				map.Save(combinedPath);
```
abcdefg30: Here are two blank lines now.

pchote:
Note: updating the UID at runtime may have unforseen consequences.  I *think* we're safe, because we only ever call `map.Save` from the editor (which needs this) and the utility (which then doesn't touch the map again).

abcdefg30:
Fixes the issue indeed. :+1: / :white_check_mark:

Mailaender:
Can't spot any negative side effects. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/b1540f7409c20a25037971371c664df2a7cd33ec...e36551a304fbc8e6b5636903e080b90a77067721)

PR #9665, opened 2015-10-18
Change display-mode without restart

huwpascoe:
Working prototype of display mode change without restarting. (Live display adjustments https://github.com/OpenRA/OpenRA/issues/3774) Type the new res in the display settings and go back/to another settings tab.

Need help to finish:
Not all ingame widgets are initialized properly
Nitpicking the display settings UI - I think a dropdown resolution changer would be better
Probably don't need help with but haven't included yet: fullscreen switching, hardware cursor switching

### DIFF
--- a/OpenRA.Game/Graphics/IGraphicsDevice.cs
+++ b/OpenRA.Game/Graphics/IGraphicsDevice.cs
@@ -55,6 +55,8 @@ namespace OpenRA
 		IShader CreateShader(string name);
 
 		Size WindowSize { get; }
+		WindowMode WindowMode { get; }
+		void SetWindowSize(Size size, WindowMode windowMode);
 
 		void Clear();
 		void Present();
@@ -135,6 +137,6 @@ namespace OpenRA
 	{
 		Windowed,
 		Fullscreen,
-		PseudoFullscreen,
+		NativeFullscreen,
 	}
 }
--- a/OpenRA.Game/Graphics/Viewport.cs
+++ b/OpenRA.Game/Graphics/Viewport.cs
@@ -116,6 +116,12 @@ namespace OpenRA.Graphics
 
 			Zoom = Game.Settings.Graphics.PixelDouble ? 2 : 1;
 			tileSize = grid.TileSize;
+
+			Game.Renderer.OnResolutionChange += (windowSize) =>
+			{
+				screenClip = new Rectangle(Point.Empty, windowSize);
+				Zoom = zoom;
+			};
 		}
 
 		public CPos ViewToWorld(int2 view)
@@ -214,7 +220,7 @@ namespace OpenRA.Graphics
 		}
 
 		// Rectangle (in viewport coords) that contains things to be drawn
-		static readonly Rectangle ScreenClip = Rectangle.FromLTRB(0, 0, Game.Renderer.Resolution.Width, Game.Renderer.Resolution.Height);
+		Rectangle screenClip = Rectangle.FromLTRB(0, 0, Game.Renderer.Resolution.Width, Game.Renderer.Resolution.Height);
 		public Rectangle GetScissorBounds(bool insideBounds)
 		{
 			// Visible rectangle in world coordinates (expanded to the corners of the cells)
@@ -224,8 +230,8 @@ namespace OpenRA.Graphics
 			var cbr = map.CenterOfCell(((MPos)bounds.BottomRight).ToCPos(map)) + new WVec(512, 512, 0);
 
 			// Convert to screen coordinates
-			var tl = WorldToViewPx(worldRenderer.ScreenPxPosition(ctl - new WVec(0, 0, ctl.Z))).Clamp(ScreenClip);
-			var br = WorldToViewPx(worldRenderer.ScreenPxPosition(cbr - new WVec(0, 0, cbr.Z))).Clamp(ScreenClip);
+			var tl = WorldToViewPx(worldRenderer.ScreenPxPosition(ctl - new WVec(0, 0, ctl.Z))).Clamp(screenClip);
+			var br = WorldToViewPx(worldRenderer.ScreenPxPosition(cbr - new WVec(0, 0, cbr.Z))).Clamp(screenClip);
 
 			// Add an extra one cell fudge in each direction for safety
 			return Rectangle.FromLTRB(tl.X - tileSize.Width, tl.Y - tileSize.Height,
--- a/OpenRA.Game/Renderer.cs
+++ b/OpenRA.Game/Renderer.cs
@@ -45,9 +45,11 @@ namespace OpenRA
 		ITexture currentPaletteTexture;
 		IBatchRenderer currentBatchRenderer;
 
+		public event Action<Size> OnResolutionChange;
+
 		public Renderer(GraphicSettings graphicSettings, ServerSettings serverSettings)
 		{
-			var resolution = GetResolution(graphicSettings);
+			var resolution = new Size(graphicSettings.WindowedSize.X, graphicSettings.WindowedSize.Y);
 
 			var rendererName = serverSettings.Dedicated ? "Null" : graphicSettings.Renderer;
 			var rendererPath = Platform.ResolvePath(".", "OpenRA.Platforms." + rendererName + ".dll");
@@ -72,14 +74,6 @@ namespace OpenRA
 			tempBuffer = Device.CreateVertexBuffer(TempBufferSize);
 		}
 
-		static Size GetResolution(GraphicSettings graphicsSettings)
-		{
-			var size = (graphicsSettings.Mode == WindowMode.Windowed)
-				? graphicsSettings.WindowedSize
-				: graphicsSettings.FullscreenSize;
-			return new Size(size.X, size.Y);
-		}
-
 		static IGraphicsDevice CreateDevice(Assembly platformDll, int width, int height, WindowMode window)
 		{
 			foreach (PlatformAttribute r in platformDll.GetCustomAttributes(typeof(PlatformAttribute), false))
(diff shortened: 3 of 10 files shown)


pchote:
Unfortunately changing the resolution is the easy part.  95% of the work in this feature is fixing the widget code to support relayout.  I suggest doing this by having code to adjust the existing widget bounds.  There is too much internal state for throwing out and rebuilding the widget tree to work unless you wanted to limit this feature to only work from the settings menu on the shellmap, which defeats the purpose.

penev92:
> Nitpicking the display settings UI - I think a dropdown resolution changer would be better

Please no. I run it in windowed mode at 1024x736, which is something you won't find in any dropdown in any game.

OpenRA.Game/Game.cs
```diff
@@ -69,6 +69,15 @@ static void JoinInner(OrderManager om)
 			ConnectionStateChanged(OrderManager);
 		}
 
+		public static void ChangeRes(Size size)
```
pchote: This isn't going to work in the long term: you can't know ahead of time everything that needs to respond to resolution changes.

You should instead set up an `OnResolutionChange` event in the Renderer class that everything else attaches callbacks to.

OpenRA.Game/Graphics/IGraphicsDevice.cs
```diff
@@ -55,6 +55,7 @@ public interface IGraphicsDevice : IDisposable
 		IShader CreateShader(string name);
 
 		Size WindowSize { get; }
+		void ChangeRes(Size size, WindowMode windowMode);
```
pchote: `SetWindowSize(Size size)` please.  This doesn't use the `windowMode` arg, and there's no need to shorten "resolution" to "res" everywhere

OpenRA.Game/Graphics/Viewport.cs
```diff
@@ -214,7 +214,14 @@ public void Scroll(float2 delta, bool ignoreBorders)
 		}
 
 		// Rectangle (in viewport coords) that contains things to be drawn
-		static readonly Rectangle ScreenClip = Rectangle.FromLTRB(0, 0, Game.Renderer.Resolution.Width, Game.Renderer.Resolution.Height);
+		static Rectangle ScreenClip
+		{
+			get
+			{
+				return Rectangle.FromLTRB(0, 0, Game.Renderer.Resolution.Width, Game.Renderer.Resolution.Height);
```
pchote: This will create additional GC pressure by creating and discarding many more objects.  You can keep this cached, and then use the OnResolutionChange event to update its value.

huwpascoe:
@penev92 Please no. I run it in windowed mode at 1024x736, which is something you won't find in any dropdown in any game. Well lets have both a dropdown and a manual input : )

x-a-n-a-x:
That wouldn't shock me: https://cloud.githubusercontent.com/assets/1732773/10564093/4c0f86c0-75a6-11e5-8773-ebd4104ecc10.png

huwpascoe:
There, got all the display modes working too!
openrares

The UI seems to be able to adjust to the changes now thanks to OnChangeResolution() although there are a few more bugs to find I'm sure.

I don't really understand what the hardware cursor is doing though if anyone has an explanation/ideas?

OpenRA.Game/Renderer.cs
```diff
@@ -91,6 +85,24 @@ static IGraphicsDevice CreateDevice(Assembly platformDll, int width, int height,
 			throw new InvalidOperationException("Renderer DLL is missing RendererAttribute to tell us what type to use!");
 		}
 
+		public readonly Size MinimumResolution = new Size(800, 600);
+
+		public readonly Size[] SuggestedResolutions = new Size[]
```
pchote: These are all common screen resolutions, so it doesn't make sense to offer them.  There is no room for the window decorations!

OpenRA.Game/Graphics/IGraphicsDevice.cs
```diff
@@ -135,6 +137,6 @@ public enum WindowMode
 	{
 		Windowed,
 		Fullscreen,
-		PseudoFullscreen,
+		NativeFullscreen,
```
pchote: You've introduced a terminology change here.  This should be at least in its own commit, but preferably in its own pr.

OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
```diff
@@ -182,13 +173,25 @@ Action InitDisplayPanel(Widget panel)
 			var cursorDoubleIsChecked = cursorDoubleCheckbox.IsChecked;
 			cursorDoubleCheckbox.IsChecked = () => !cursorDoubleCheckbox.IsDisabled() && cursorDoubleIsChecked();
 
-			panel.Get("WINDOW_RESOLUTION").IsVisible = () => ds.Mode == WindowMode.Windowed;
+			panel.Get("WINDOW_RESOLUTION").IsVisible = () => ds.Mode != WindowMode.NativeFullscreen;
```
pchote: Erm, what's going on here?  The resolution control only makes sense for the windowed mode.

OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
```diff
@@ -98,6 +86,47 @@ public Sdl2GraphicsDevice(Size windowSize, WindowMode windowMode)
 
 			SDL.SDL_SetModState(SDL.SDL_Keymod.KMOD_NONE);
 			input = new Sdl2Input();
+
+			Clear();
+			Present();
+			SetWindowSize(WindowSize, windowMode);
+		}
+
+		public void SetWindowSize(Size windowSize, WindowMode windowMode)
```
pchote: You're now handling the window mode as well, so calling this `SetWindowSize` doesn't make sense.  It should be changed to `SetWindowMode`.

OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
```diff
@@ -98,6 +86,47 @@ public Sdl2GraphicsDevice(Size windowSize, WindowMode windowMode)
...
+				SDL.SDL_SetWindowFullscreen(window, (uint)SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN);
+			}
+
+			if (windowMode == WindowMode.NativeFullscreen)
+				SDL.SDL_SetWindowFullscreen(window, (uint)SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN_DESKTOP);
+
+			if (windowMode == WindowMode.Windowed)
+				SDL.SDL_SetWindowPosition(window, SDL.SDL_WINDOWPOS_CENTERED, SDL.SDL_WINDOWPOS_CENTERED);
+			else
+				SDL.SDL_SetWindowPosition(window, 0, 0);
```
pchote: Be careful about / please carry over the commented workaround for OSX from the code you've removed.

pchote:
This is already starting to turn into a monster pr with multiple distinct changes.  Please keep the scope small, and split out smaller prs as needed.

OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
GraionDilach: What? Definitely no, fullscreen (or pseudo-notsure of a sudden which I used) resolution change can be done via editing the settings.yaml on bleed. It makes sense as a form of full zooming in.

pchote:
> The UI seems to be able to adjust to the changes now thanks to OnChangeResolution() although there are a few more bugs to find I'm sure.

The loadscreen geometry is wrong, and the ingame UI breaks if you change resolution from inside a game.

OpenRA.Game/Widgets/Widget.cs
```diff
@@ -237,6 +241,15 @@ public virtual void Initialize(WidgetArgs args)
 								   height);
 		}
 
+		protected void OnResolutionChange()
+		{
+			if (widgetArgs != null)
+				Initialize(widgetArgs);
```
pchote: Widgets do more than just lay out their geometry in Inititialize.  Calling this multiple times is a bad idea.  You should instead implement a new `OnResolutionChangeInner` method that each widget can implement only the things it needs on resize (most widgets will be able to use a default implementation on the base widget).

pchote:
Its going to take a bit more work to get this mergable, but its a promising start!

pchote:
Changing to a custom resolution in Fullscreen (not native) under OSX gives bogus geometry, chopping off part of the screen and giving different coordinates for rendering and inputs (so there is a constant offset between where the cursor is rendered and where click events are fired).  This is probably an SDL bug rather than something you did.

OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
huwpascoe: True full screen mode allows resolution change, this is a highly desired feature!

OpenRA.Game/Renderer.cs
huwpascoe: That gives me an idea actually. I'll get back to that one

OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
huwpascoe: This is no longer a workaround but required by all systems. Should I still grab the comment?

huwpascoe:
Can you provide a screenshot of the osx bug?

OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
huwpascoe: Er required for correct mode changing I mean

pchote:
screen shot 2015-10-18 at 23 05 36

Note that the (hardware rendered) cursor was positioned near the "OpenRA" text, while the game thought it was over the Multiplayer button.  This vertical offset matches the size of the black bar at the top of the screen.

huwpascoe:
Closing this for now, it relies on another dependency which I'll probably make another PR for.

pchote:
Adding to the future milestone so that we can find this again if you don't have time to reopen it.

PR #9668, opened 2015-10-18
RA-Balance changes for the next playtest

abcdefg30:
Next try after #9623 was closed.

### DIFF
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -39,6 +39,7 @@ DOG:
 		StandSequences: stand
 	IgnoresDisguise:
 	DetectCloaked:
+		CloakTypes: Cloak, Hijacker
 		Range: 5c0
 	Voiced:
 		VoiceSet: DogVoice
@@ -222,6 +223,7 @@ SPY:
 		UpgradeMinEnabledLevel: 1
 	IgnoresDisguise:
 	DetectCloaked:
+		CloakTypes: Cloak, Hijacker
 		Range: 5c0
 	Armament:
 		Weapon: SilencedPPK
@@ -452,6 +454,12 @@ HIJACKER:
 		StandSequences: stand
 	Crushable:
 		WarnProbability: 95
+	Cloak:
+		InitialDelay: 250
+		CloakDelay: 120
+		UncloakOnMove: true
+		CloakTypes: Cloak, Hijacker
+		IsPlayerPalette: true
 
 SHOK:
 	Inherits: ^Soldier
@@ -540,6 +548,7 @@ SNIPER:
 		UncloakOnMove: true
 		IsPlayerPalette: true
 	DetectCloaked:
+		CloakTypes: Cloak, Hijacker
 		Range: 6c0
 	-MustBeDestroyed:
 	ProducibleWithLevel:
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -40,7 +40,6 @@ DOG:
 	IgnoresDisguise:
 	DetectCloaked:
 		CloakTypes: Cloak, Hijacker
-		Range: 5c0
 	Voiced:
 		VoiceSet: DogVoice
 	-TakeCover:
@@ -224,7 +223,6 @@ SPY:
 	IgnoresDisguise:
 	DetectCloaked:
 		CloakTypes: Cloak, Hijacker
-		Range: 5c0
 	Armament:
 		Weapon: SilencedPPK
 	AttackFrontal:
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -450,6 +450,8 @@ HIJACKER:
 	WithInfantryBody:
 		-IdleSequences:
 		StandSequences: stand
+	Crushable:
+		WarnProbability: 95
 
 SHOK:
 	Inherits: ^Soldier
--- a/mods/ra/rules/misc.yaml
+++ b/mods/ra/rules/misc.yaml
@@ -144,15 +144,9 @@ camera.spyplane:
 		CloakTypes: Cloak, Hijacker
 
 SONAR:
-	AlwaysVisible:
-	Immobile:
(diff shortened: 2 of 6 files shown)


MicroJOo:
also the flaktrucks require further tweaking ... they have been reported "extremely" usefull against "Hinds" ... this values here , shows why this is a "real issue"
_____________________________________
FLAKTRUCK:
        RevealsShroud:
	Range: 6c0
Weapon:
        FLAK-23:
	Range: 8c0
_____________________________________
HIND:
        RevealsShroud:
	Range: 10c0
Weapon:
        ChainGun:
	Range: 5c0

i can see what @ScottNZ did here ... the idea in theory was to make Flaktrucks work together with other (scout) units ... to make it more effective
but experiencing this on the battlefield is quite sobering , flaks are agile and fast ... and quickly to produce with the cost of only 600 (even a light tank costs 700) ... apart from that , its possible to "curtain" them ... this is what soviet-players do when they think its "time to clean up" on the front ... destroying artys/v2s and all aircrafts ....

i suggest to here to start fixing this with : 

Weapon:
        FLAK-23:
	Range: 8c0 -> 6c0

and lets see how this will turn out in playtests

r34ch:
You could also buff the light tank to be a flak trucks direct counter. Lower the cost to $600 and buff the RoT since the speed for both are already the same. Perhaps buff the damage against 'Heavy' targets to also make them useful for early OT harassment and mid game hit and run.

MicroJOo:
thats not gonna work ... in order to be a direct counter , we have to buff also the range of the ligh-tanks ...
there is allready plans to change the "ROT" of mediums and ligh-tanks ... but we should wait for this until we have  something to "trade" in , otherwise people could get mad 

better we keep it simple , and just see what the decreased flaktruck-range will do

abcdefg30:
Added the `FLAK-23` range decrease.

abcdefg30:
Updated.

r34ch:
@MicroJOo I don't see why the light tank range would need increased if the flak truck range is being decreased? If that happened along with the RoT and price changes then the light tank would dominate the flak truck with its much heavier armour, far higher DPM and equal speed.

But anyway, like you say lets test the flak truck change first. I think it will be a good change but I'm curious about defending sub pens from air...

MicroJOo:
r34ch : that you cant really defend a subpen from air-units hasnt been changed ... even when having FLAK-23 on Range: 8c0 .. it cant reach a hind from a straight coast ... but thats a different story , we all know something has to be changed on the naval stuff

about the light tank ... i dont see why we should give him the same speed as the flaktruck ...its still a "tank" ... and has other advantages like "leaving husks" , "crushing infantry" ... you can do some nice harrassement on powerplants and so on ... 

as said , ill up the ROT ... cause thats one thing that would actually make sense ... (its hard to explain though , why out of the sudden all the allies tanks get a buff ... so im for waiting on this

reaperrr:
wrt the flak truck, if the range nerf turns out to be a little too large vs. aircraft, a possibility would be to split the weapon into AA-only and AG-only and then only re-increase range of the AA variant.

mods/ra/rules/infantry.yaml
```diff
@@ -450,6 +450,14 @@ HIJACKER:
 	WithInfantryBody:
 		-IdleSequences:
 		StandSequences: stand
+	Crushable:
+		WarnProbability: 95
+	Cloak:
+		InitialDelay: 250
+		CloakDelay: 120
+		UncloakOnMove: true
+		CloakTypes: Cloak, Hijacker
```
obrakmann: Why isn't just `Hijacker` not enough?

mods/ra/rules/misc.yaml
```diff
@@ -126,10 +126,6 @@ CAMERA:
 	RevealsShroud:
 		Range: 10c0
 		Type: CenterPosition
-	ProximityCaptor:
-		Types: Camera
-	DetectCloaked:
-		Range: 10c0
```
obrakmann: This was actually on purpose, see here: fc2c46a2c25aad0c5c05a869d18e37762f673fe8

But cameras are used in too many places now for this to not potentially have bad side effects.

obrakmann:
Confirmed that hijackers can still be crushed by harvesters :p

I hope we'll get some feedback on how this turns out during the playtest. This is quite a buff for the hijacker. Also not sure how the weapon range decrease for the flak truck will really affect things. But yeah, they are annoying things.

Oh well. :+1:, pending an answer on the `CloakTypes` question.

MicroJOo:
@obrakmann : about your CloakTypes questio 

i would like to have also "base defenses" detect him ... and not rely completely on infantry to detect him , thats why "cloak" too , beside "hijacker"

considering i didnt up the price of the unit , this should be fair enough

GraionDilach:
Something is definitely off here though, in the original PR of @MicroJOo IIRC there were stuff which could only detect the hijacker but now there isn't. `^Soldier` or `^Defense` should get a `DetectCloaked: Hijacker` then.

abcdefg30:
> `^Soldier` or `^Defense` should get a `DetectCloaked: Hijacker` then.

That is `^Soldier`:
https://github.com/OpenRA/OpenRA/pull/9668/files#diff-5b0780203fcfc47fcb88e0b8d6d306b9R271

GraionDilach:
Ah, tight, ignore me being stupid.

mods/ra/rules/misc.yaml
abcdefg30: Added a new `camera.spyplane`.

abcdefg30:
Updated.

MicroJOo:
:+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/629e1b74f2de01fe3ba3a104225402f91d915660

r34ch:
Well fwiw I think the chinook sight range and hijacker buffs are good. 

I can't comment fully on the flak changes because I didn't really play many naval maps but the changes don't seem to break any of the main strategies on land with them.

PR #9676, opened 2015-10-18
Tweaked D2k warhead falloffs

reaperrr:
I recently played the original quite a bit and I think our area of effect on several weapons is over the top even compared to the original, which is one of the reasons why debris are too powerful on bleed (they are being dealt with in #9615).

I tried to keep things as simple and consistent as possible, but this might need more testing and fine-tuning.

### DIFF
--- a/mods/d2k/weapons.yaml
+++ b/mods/d2k/weapons.yaml
@@ -9,13 +9,11 @@ LMG:
 		Falloff: 100, 50, 25, 0
 		Damage: 125
 		Versus:
-			none: 100
 			wall: 10
 			building: 25
 			wood: 75
 			light: 40
 			heavy: 20
-			concrete: 100
 			invulnerable: 0
 			cy: 20
 			harvester: 25
@@ -48,8 +46,6 @@ Bazooka:
 			building: 40
 			wood: 45
 			light: 70
-			heavy: 100
-			concrete: 100
 			invulnerable: 0
 			cy: 20
 			harvester: 50
@@ -71,13 +67,11 @@ Fremen_S:
 		Falloff: 100, 50, 25, 0
 		Damage: 125
 		Versus:
-			none: 100
 			wall: 10
 			building: 25
 			wood: 75
 			light: 40
 			heavy: 20
-			concrete: 100
 			invulnerable: 0
 			cy: 20
 			harvester: 25
@@ -99,13 +93,11 @@ M_LMG:
 		Damage: 125
 		ValidTargets: Ground
 		Versus:
-			none: 100
 			wall: 10
 			building: 25
 			wood: 75
 			light: 40
 			heavy: 20
-			concrete: 100
 			invulnerable: 0
 			cy: 20
 			harvester: 25
@@ -125,12 +117,9 @@ M_HMG:
 		Damage: 250
 		Versus:
 			none: 25
-			wall: 100
 			building: 50
 			wood: 65
-			light: 100
 			heavy: 50
-			concrete: 100
 			invulnerable: 0
 			cy: 20
 			harvester: 50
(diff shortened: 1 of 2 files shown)


mods/d2k/weapons.yaml
```diff
@@ -5,8 +5,8 @@ LMG:
 	Projectile: Bullet
 		Speed: 1c256
 	Warhead@1Dam: SpreadDamage
-		Spread: 48
-		Falloff: 100, 100, 65, 30, 0
+		Spread: 96
+		Falloff: 100, 50, 25, 12, 6, 0
```
pchote: Is there a specific reason why you didn't go back to our default falloff model for these? (the two are basically equal if you scale the spread by a factor of ~1.4 and don't mind the wider negligible-damage tail).

reaperrr:
Updated.

- Using default Falloff in favor of larger Spread values
- increased machinegun spread a bit more to account for infantry's smaller health radius
- added some cleanup commits

pchote:
I spectated an all bot match and didn't spot anything bogus, so :+1:

penev92:
Damage spreads are ridiculously high.
Quad rockets deal damage in a 3-cell diameter. Trike machine gun deals damage in a 4 (!) cell diameter.
Infantry's rifles deal damage in a 2-cell diameter and bazookas - in 3.

This both looks ridiculous when you enable Combat Geometry in the debug menu and causes infantry to go prone while not even being fired at.

reaperrr:
I guess this needs a custom Falloff afterall.

reaperrr:
Added back custom falloffs.

- for machineguns, I left Spread untouched and only added a custom Falloff of `100, 40, 16, 0`, which means damage is slightly higher at first and second spread, but then falls off a cliff compared to internal default to avoid adjacent infantry going prone.
- for anti-tank weapons, I also left Spread untouched and added a custom Falloff of `100, 50, 25, 0`, which means damage is moderately higher at first two spread ranges but then falls off a cliff, too.
- for high-explosive weapons, I slightly reduced Spread a bit in some cases (siege tank 155mm from 448 to 384) and added a custom Falloff of `100, 60, 30, 15, 0`, so the total area of effect is still significantly lower than with default Falloff + previous Spread, but damage falloff *within* area of effect is not as steep, to better reflect original behavior.

Don't merge this yet, there are some minor things I noticed which I want to adress first.

reaperrr:
Updated.

- fine-tuned machine gun falloffs a bit more
- swapped explosion animation of small and medium unit explosion to match original
- fixed unit explosion sounds to match original as well
- removed spice explosion impact sound, was way too annoying

penev92:
I noticed two things, the first one of which is unrelated to this PR:
 - Combat tanks are surprisingly inaccurate;
 - The Siege tank, while having a good spread, perhaps has too much of a Falloff, because the spread damage is negledgeable.

Other than that, the spreads look a lot more sane now.

reaperrr:
Updated.

- Reduced Siege Tank and Grenadier `Falloff` by 5 points at stages 2-4.
- Slightly re-increased Siege Tank `Spread` from 384 (12 pixel) to 416 (13 pixel), to stretch out falloff slightly more.

penev92:
Looks fine now :+1: 

Although I expect the grenadiers will need a damage buff.
Not merging because not sure of @pchote's +1

pchote:
My :+1: above was nullified by the complete change (also by my insufficient testing not spotting those major issues).

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/b054cd351ea6c67cb1a1472421026233a8a09173)

PR #9681, opened 2015-10-19
Fix D2K refinery flames.

### LINKED ISSUE #9670: d2k: smoking chimneys on refineries don't work anymore
Probably a bug in the fragment shader, introduced with 83949b250a07b8829b5e3b3bb2ad95644c7fec17 in #9476.

Removing [these two lines](https://github.com/OpenRA/OpenRA/blob/bleed/glsl/shp.frag#L16-L17) makes it work again, but that's most likely not the correct solution.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27500497-d2k-smoking-chimneys-on-refineries-don-t-work-anymore?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #9670.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithDockingOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithDockingOverlay.cs
@@ -53,7 +53,7 @@ namespace OpenRA.Mods.Common.Traits
 
 			anim = new AnimationWithOffset(overlay,
 				() => body.LocalToWorld(info.Offset.Rotate(body.QuantizeOrientation(self, self.Orientation))),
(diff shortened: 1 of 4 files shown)


Mailaender:
:+1: This also fixes the problem where the flash target effect also shows the flames.

abcdefg30:

PR #9683, opened 2015-10-19
Add *.xml to gitignore

reaperrr:
`SmarIrc4net.xml` shows up as new file all the time, which is quite annoying.

The only other .xml in the main folder is the StyleCop one, so this shouldn't cause any problems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ mods/*/*.mdb
 /*.pdb
 /*.mdb
 /*.exe
+/*.xml
 thirdparty/download/*
 *.mmdb.gz
 


pchote:
I think it would be better to fix whatever is causing that xml file to appear.  On my machine that file only exists in the thirdparty download folder - is the windows make script copying too much into the main dir?

reaperrr:
Seems that way. What I don't get is why this happens during compile (`all`) rather than `dependencies`.

clemty:
note that there is a 't' missing in the filename (a google search for that string brings up only OpenRA)

cjshmyr:
Is packaged that way in NuGet -- @Mailaender did they ever get around to fixing that?

clemty:
PR which removes the .XML while running 'make dependencies': #9694

Mailaender:
Those XML files are required for IDE inline documentation to work so you shouldn't remove them.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/issues/9697.

Issue #9451, opened 2015-09-26
Re-design server browser and lobby


cjshmyr:
I think the server browser (and to some extent the lobby) are in need of an overhaul. Here's a mockup I created fairly quickly with a few concepts highlighted. I'm not a self proclaimed design expert at all :), please chime in!

Server
- Show up to 10 servers at a time now with a flatter lobby list. Currently with the chat we only see 3!
- Show a panel on the right for more detailed information, we don't really care about some of this stuff for every single server (minimap preview, ip, version, and potentially players)
- Potentially allow players to sort by clicking on the server list columns
- Move the server filter into its own popup panel, which would show over the server list
- Change the chat to be tabbed - this means once you join a server, you can alternate between global and lobby chat

Lobby
- Added a Music button and moved the Change Map button. The Music panel would cover the player bin when opened.
- Changed the chat to be tabbed - let you switch between lobby and global chat.

Mockups (the OpenRA window is 1280x1024)
- It was suggested we keep the geometry of the server browser and game lobby the same size
- This has to fit the user interface's design limitation of 1024x768 size resolution

![serverbrowser](https://cloud.githubusercontent.com/assets/566742/10119065/8dd007e8-6448-11e5-8279-f255c61d3c95.png)
![lobby](https://cloud.githubusercontent.com/assets/566742/10119050/1ab872a4-6448-11e5-9dd2-706239636f42.png)

These mockups are available as a Balsamiq file so anyone can tweak them

obrakmann:
Those mock-ups look purdy. Does the tabbed design mean that people would be able to connect to multiple servers at once? If so, that might be exploited by people with malicious intent to hog player slots or even whole servers.

Another thing that would be nice is to show more information on the map in the lobby (whether it has custom rules, the description text and so on).

Also, just as food for thought: maybe the players and options bin could be implemented as tabs as well. (And the (extended) map info could then be a third tab, maybe.). I always found it odd to change the panels with a button. That might waste too much vertical space, though.

MicroJOo:
i want this !

cjshmyr:
> Those mock-ups look purdy. Does the tabbed design mean that people would be able to connect to multiple servers at once? If so, that might be exploited by people with malicious intent to hog player slots or even whole servers.

Nope, like usual you have to leave your current lobby to join another

> Another thing that would be nice is to show more information on the map in the lobby (whether it has custom rules, the description text and so on).

I want this too!

[...]

PR #9692, opened 2015-10-20
Integrate global chat with the lobby.

pchote:
The next step towards #9451.  A future PR will add unread message counts to the tab labels, but that depends on some wider widget refactoring that will come in the next PR.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
@@ -37,6 +37,9 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 		enum PanelType { Players, Options, Music, Kick, ForceStart }
 		PanelType panel = PanelType.Players;
 
+		enum ChatPanelType { Lobby, Global }
+		ChatPanelType chatPanel = ChatPanelType.Lobby;
+
 		readonly Widget lobby;
 		readonly Widget editablePlayerTemplate;
 		readonly Widget nonEditablePlayerTemplate;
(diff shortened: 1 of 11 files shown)


obrakmann:
The two 'disconnect' buttons right next to each other irk me. Can we change the label on one of them (I'd probably go with 'Leave [match|game|lobby]')?
screenshot from 2015-10-20 21 13 24

pchote:
In TD we use "Back" for skirmish, and "Leave Game" for MP.  I never noticed that RA used different terminology.  I'll change them to match TD, and then use "Leave Chat" for the IRC button.

pchote:
Updated.

MicroJOo:
this works :+1:

pchote:
Rebased.

obrakmann:
What do you think about making the buttons all the same width? It pleases my OCD ;)

screenshot from 2015-10-21 20 47 25

Diff:
```diff
index bf0c684..de45eb0 100644
--- a/mods/ra/chrome/lobby.yaml
+++ b/mods/ra/chrome/lobby.yaml
@@ -124,16 +124,16 @@ Background@SERVER_LOBBY:
                        Width: PARENT_RIGHT - 40
                        Height: 229
                Button@START_GAME_BUTTON:
-                       X: PARENT_RIGHT - WIDTH - 150
+                       X: PARENT_RIGHT - WIDTH - 205
                        Y: PARENT_BOTTOM - HEIGHT - 20
-                       Width: 120
+                       Width: 172
                        Height: 25
                        Text: Start Game
                        Font: Bold
                Button@DISCONNECT_BUTTON:
                        X: PARENT_RIGHT - WIDTH - 20
                        Y: PARENT_BOTTOM - HEIGHT - 20
-                       Width: 120
+                       Width: 172
                        Height: 25
                        Text: Leave Game
                        Font: Bold
```

pchote:
The button widths are matched to the map chooser, and (will be matched to) the new server browser, so that ends up being a rather major style change.  The current button widths / positions are consistent across most of the UI panels.

obrakmann:
hmm... ok. maybe we can come up with another solution eventually. for now, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/427df889772cc4a4ab35c4d386d7e2063d3b7545)

PR #9695, opened 2015-10-20
Fix solution file

RoosterDragon:
VS2015 was producing a warning during loading of the solution stating it could not load some of the properties. This rectifies it.


penev92:
Thanks!

abcdefg30:
Ops, thanks.

PR #9696, opened 2015-10-20

RoosterDragon:
VS like to streamroll over the format of this file all the time - keeping the version in the repo in line with its arbitrary whims likes removes a mild nuisance for VS users as it won't dirty the working directory (which then requires it to be reset when rebasing, etc).

### DIFF
--- a/OpenRA.sln
+++ b/OpenRA.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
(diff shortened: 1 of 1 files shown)


Mailaender:
Still opens with MonoDevelop. :white_check_mark:

huwpascoe:
:+1: Yes please

pchote:
What is the minimum OS requirement for that version of VS, and whatever version of MD is required to support it?

RoosterDragon:
The minimum version of VS is 2010 (given by the `MinimumVisualStudioVersion` line) and has not changed.

PR #9697, opened 2015-10-20
Don't force SmarIrc4net to be copied to output

RoosterDragon:
Instead, it will be implicitly included since it is a reference. This avoids pulling in the documentary .xml file.

Supersedes #9683.


Mailaender:
`make clean && make dependencies` then build and start with MonoDevelop works without problems. :white_check_mark:

penev92:
Fixes it on Windows :+1:

Issue #9443, opened 2015-09-26
Spatially partition IEffects


pchote:
We render every effect in the world, including those that are outside the screen bounds.  Adding a new `SpatiallyPartitioned<IEffect> partitionedEffects` to ScreenMap for use by the WorldRenderer should give us a reasonable rendering performance win.  As an added bonus this should cut the rendering workload to the point where it is feasible to interleave actor and effect renderables, which fixes a big chunk of polish issues.

RoosterDragon:
Is the interleaving as simple as ensuring both the actors and effects are affected by the sort in `WorldRenderer.GenerateRenderables`? If so we can already afford that on the RA shellmap easily.

Mailaender:
I tried a simple approach without partiion IEffects https://github.com/Mailaender/OpenRA/tree/effects-viewport but it seems sprites have a different definition of screen bounds compared to all other renderables

pchote:
Your definition of the viewport bounds are incorrect in that branch.  That will cause you to render effects on only 1/4 of the visible screen.

pchote:
Fixed by #13979.

PR #9699, opened 2015-10-20
Added ScreenMap.partitionedEffects

Mailaender:
Some kind of first try to tackle https://github.com/OpenRA/OpenRA/issues/9443. I don't know how to measure performance to see if it worked. There is a problem that effects don't have bounds and effects can be visible, but can also be something completely different like `DelayedAction` that renders nothing. Not sure about best practices on this.

### DIFF
--- a/OpenRA.Game/Traits/World/ScreenMap.cs
+++ b/OpenRA.Game/Traits/World/ScreenMap.cs
@@ -13,6 +13,7 @@ using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
+using OpenRA.Effects;
 using OpenRA.Graphics;
 using OpenRA.Primitives;
 
@@ -33,6 +34,7 @@ namespace OpenRA.Traits
 		readonly Func<Actor, bool> actorIsInWorld = a => a.IsInWorld;
 		readonly Cache<Player, SpatiallyPartitioned<FrozenActor>> partitionedFrozenActors;
 		readonly SpatiallyPartitioned<Actor> partitionedActors;
+		readonly SpatiallyPartitioned<IEffect> partitionedEffects;
 		WorldRenderer worldRenderer;
 
 		public ScreenMap(World world, ScreenMapInfo info)
@@ -43,6 +45,7 @@ namespace OpenRA.Traits
 			partitionedFrozenActors = new Cache<Player, SpatiallyPartitioned<FrozenActor>>(
 				_ => new SpatiallyPartitioned<FrozenActor>(width, height, info.BinSize));
 			partitionedActors = new SpatiallyPartitioned<Actor>(width, height, info.BinSize);
+			partitionedEffects = new SpatiallyPartitioned<IEffect>(width, height, info.BinSize);
 		}
 
 		public void WorldLoaded(World w, WorldRenderer wr) { worldRenderer = wr; }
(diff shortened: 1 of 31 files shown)


pchote:
> effects don't have bounds

This was implicit in the issue ticket, sorry.  Implementing this properly requires a `.VisualBounds` property to be added to IEffect.

> effects can be visible, but can also be something completely different

Effects that don't render can return `Rectangle.Empty` for their bounds, and the caching code can check for this and exclude them from the partition.

Mailaender:
Added bounds to everything.

OpenRA.Game/Traits/World/ScreenMap.cs
```diff
@@ -67,6 +70,16 @@ public void Add(Player viewer, FrozenActor fa)
 			partitionedFrozenActors[viewer].Add(fa, FrozenActorBounds(fa));
 		}
 
+		public void Add(IEffect effect, WPos position)
```
pchote: Effects can (and do: projectiles) move.  The effects themselves will need to tell the screenmap when they are added, moved, or removed from the world in the same way that actors do.

Mailaender: Added an `Update(IEffect effect, WPos position)` function.

pchote:
Mixing add/remove at the world level and updating at the activity level isn't that nice.  Again, please make this work like actors where those that care about positions themselves update the screen map.

pchote:
Any updates here?

Mailaender:
Rebased and rewrote it so that effects aren't automatically added/removed.

OpenRA.Game/Effects/IEffect.cs
```diff
@@ -18,5 +19,6 @@ public interface IEffect
 	{
 		void Tick(World world);
 		IEnumerable<IRenderable> Render(WorldRenderer r);
+		Rectangle VisualBounds(WorldRenderer wr);
```
pchote: If you pass the bounds into the ScreenMap.Add call then you won't need this on the interface, and can avoid having the dummy implementations on the effects that don't interact with the ScreenMap.

OpenRA.Game/Traits/World/ScreenMap.cs
```diff
@@ -68,6 +71,16 @@ public void Add(Player viewer, FrozenActor fa)
...
+		{
+			var bounds = effect.VisualBounds(worldRenderer);
+			if (bounds.Width > 0 && bounds.Height > 0)
+			{
+				var screenPos = worldRenderer.ScreenPxPosition(position);
```
pchote: `screenPos` gives the middle of the sprite, but you treat it as the top-left below.  This will offset everything in the map by `(width, height) / 2`.

OpenRA.Game/Traits/World/ScreenMap.cs
```diff
@@ -83,11 +96,30 @@ public void Update(Actor a)
...
 
+		public void Update(IEffect effect, WPos position)
+		{
+			if (!partitionedEffects.Contains(effect))
+				return;
+
+			var bounds = effect.VisualBounds(worldRenderer);
+			if (bounds.Width > 0 && bounds.Height > 0)
+			{
+				var screenPos = worldRenderer.ScreenPxPosition(position);
```
pchote: Likewise here.

Mailaender:
Rewrote this again and corrected the offset.

reaperrr:
Needs rebase.

Mailaender:
Rebased, but lasers and teslas got fucked up during all the rewritings and I don't have an immediate fix.

reaperrr:
I'm closing this for now to clear the PR queue, but it would still be nice to get the performance improvement so please reopen it when you have time to debug and rebase it.

Mailaender:
I made a new attempt: https://github.com/OpenRA/OpenRA/pull/11676

PR #9700, opened 2015-10-20
Chronotank uses extra bar when charging.

### LINKED ISSUE #9558: Replace chronotank blue charging pips to pink charging bar
Currently the chronotank has a single 'chronoshift' / 'blink' ability. Although I would *love* to see it have a double blink ability, I think using pips to display a 'support power' charging is inconsistent with the other support powers which display a pink bar as they charge. This could confuse newer players where pips represent passengers (or ore harvested in ore trucks)

The hue of the blue is also hard to see against temperate terrain.

ctank

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27201236-replace-chronotank-blue-charging-pips-to-pink-charging-bar?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ghost:
The two blue pips have been replaced by
a magenta charge bar.

ctank

Closes #9558 if accepted.

### DIFF
--- a/OpenRA.Mods.RA/Traits/PortableChrono.cs
+++ b/OpenRA.Mods.RA/Traits/PortableChrono.cs
@@ -32,9 +32,6 @@ namespace OpenRA.Mods.RA.Traits
 		[Desc("Sound to play when teleporting.")]
 		public readonly string ChronoshiftSound = "chrotnk1.aud";
 
-		[Desc("Display rectangles indicating the current charge status")]
-		public readonly int Pips = 2;
-
 		[Desc("Cursor to display when able to deploy the actor.")]
 		public readonly string DeployCursor = "deploy";
 
@@ -46,7 +43,7 @@ namespace OpenRA.Mods.RA.Traits
 		public object Create(ActorInitializer init) { return new PortableChrono(this); }
 	}
 
-	class PortableChrono : IIssueOrder, IResolveOrder, ITick, IPips, IOrderVoice, ISync
+	class PortableChrono : IIssueOrder, IResolveOrder, ITick, ISelectionBar, IOrderVoice, ISync
 	{
 		[Sync] int chargeTick = 0;
 		public readonly PortableChronoInfo Info;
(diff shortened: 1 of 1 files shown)


Mailaender:
:+1:

OpenRA.Mods.RA/Traits/PortableChrono.cs
```diff
@@ -108,19 +105,12 @@ public bool CanTeleport
...
-			{
-				if ((1 - chargeTick * 1.0f / (25 * Info.ChargeTime)) * Info.Pips < i + 1)
-				{
-					yield return PipType.Transparent;
-					continue;
-				}
-
-				yield return PipType.Blue;
-			}
+			return (float)((chargeTick - (25 * Info.ChargeTime)) * (Info.ChargeTime * -0.0001));
```
penev92: Tick rate is no longer hardcoded to 25 per second. You should use `World.LobbyInfo.GlobalSettings.Timestep` instead to account for different game speeds.

Also why the `* -0.0001` ?

Mailaender: https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.RA/Traits/SupportPowers/ChronoshiftPower.cs#L55 In fact the chronoshift charge time is still hard-coded.

penev92: That is a different issue that needs a separate fix before we release.

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/191b3fd2e3c06cb0f74d824659b6ef58cb5e824d...15db90b46543046437d796a793af365a03fa630a)

penev92:
[Fixed changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/15db90b46543046437d796a793af365a03fa630a...3a345642756b7d0beb4b45b1cf152addffa676b8)

OpenRA.Mods.RA/Traits/PortableChrono.cs
pchote: Scaling the charge time based on game speed doesn't make any sense (this breaks the game balance at different speeds).  The proper fix is to convert the charge time to ticks so that it is unambiguously measured relative to the world.

pchote: It is correct for `CustomBuildTime` and `Animation.defaultTick` to have hardcoded values.  Scaling those with game speed would also break the game at non-standard speeds.

PR #9701, opened 2015-10-20
Add a common ChromeLogic base class and lint verification.

pchote:
This makes all of our chrome logic classes inherit from a new `ChromeLogic`, and adds a lint pass that checks that all the referenced logics exist and are of the correct type.

This puts the foundations in place for the next PR, which will replace the conceptually ugly `LogicTickerWidget` with a `Tick` function on the chrome logic classes themselves.

### DIFF
--- a/OpenRA.Game/Widgets/Widget.cs
+++ b/OpenRA.Game/Widgets/Widget.cs
@@ -140,6 +140,12 @@ namespace OpenRA.Widgets
 		}
 	}
 
+	public class ChromeLogic : IDisposable
+	{
+		public void Dispose() { Dispose(true); GC.SuppressFinalize(this); }
+		protected virtual void Dispose(bool disposing) { }
+	}
+
 	public abstract class Widget
 	{
 		public readonly List<Widget> Children = new List<Widget>();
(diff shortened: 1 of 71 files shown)


Mailaender:
A very good idea. :+1:

OpenRA.Game/Widgets/Widget.cs
```diff
@@ -140,6 +140,11 @@ public static void ResetTooltips()
 		}
 	}
 
+	public class ChromeLogic : IDisposable
+	{
+		public virtual void Dispose() { }
```
RoosterDragon: Dispose pattern please.

```
public void Dispose() { Dispose(true); GC.SuppressFinalize(this); }
protected virtual void Dispose(bool disposing) { }
```

pchote:
Updated.

penev92:
OK :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/427df889772cc4a4ab35c4d386d7e2063d3b7545...21ad227b55cbd04a8ee6f9f89859f67673b34ebd)

Issue #9680, opened 2015-10-19
d2k aircraft speeds are far too slow.


pchote:
I suspect the values defined in the original d2k were before the acceleration bonuses were applied.  As they are in OpenRA it feels like carryalls are actually slower than letting harvesters drive themselves, and the starport deliveries slow to the point that I first though somebody had broken the production traits.

reaperrr:
> I suspect the values defined in the original d2k were before the acceleration bonuses were applied.

I don't think so. Aircraft in D2k were not particularly fast. It *could* be that our aircraft-specific conversion formula is at fault (for D2k it's effectively original speed [ / 64] * 7, whereas the formula for ground units gives original speed [ / 64] * 10.67), but I don't know how much of a difference there is between our aircraft speed and our ground unit speed. I assume there's a reason why the formulas differ like that.

>  As they are in OpenRA it feels like carryalls are actually slower than letting harvesters drive themselves

That applied to the original as well, at least for shorter distances. It was actually worse, because they were slower at picking up and dropping them. Carryalls were only really beneficial over longer distances (when the pick/drop overhead had less of an effect) or when the direct path was blocked by cliffs or other obstacles (D2k pathfinding was more performant, but also dumber than ours). And for partial protection from worms and enemies.

> and the starport deliveries slow to the point that I first though somebody had broken the production traits.

The real problem here is that they should fly in from the closest map border cell, like they did in the original. The TD-style approach works poorly, especially on large maps.


Anyway, filed #9702.

pchote:
Fixed by #9702.

PR #9702, opened 2015-10-21
Increase D2k aircraft speed

reaperrr:
While I'm relatively sure that the current values are correctly or at least closely replicating the original speeds (normalized to our gamespeed), I agree with #9680 that they somewhat hamper gameplay.

I tried to chose values that don't make carryalls and ornithopters too hard to hit by missiles. The frigate's speed can be reverted once we can make them appear from the closest map border cell.

### DIFF
--- a/mods/d2k/rules/aircraft.yaml
+++ b/mods/d2k/rules/aircraft.yaml
@@ -15,7 +15,7 @@ carryall.reinforce:
 		CruiseAltitude: 2100
 		InitialFacing: 0
 		ROT: 4
-		Speed: 112
+		Speed: 144 # 112 * ~1.3 for balance reasons
 		LandableTerrainTypes: Sand, Rock, Transition, Spice, Dune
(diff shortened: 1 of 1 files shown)


abcdefg30:
Do we want the comments in our yaml files?
Also why didn't you add comments to the last 2 changes? (Because those are the same as already done changes?)

abcdefg30:
:+1: otherwise.

reaperrr:
I'm not sure about the comments myself, but I'm afraid without comments, in a few months nobody (maybe not even myself) would remember the reason for these values.

pchote:
Feels more playable to me :+1:.  We can iterate balance further in a future pr once we add acceleration.

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/18c79321cd1d3e57ce92b41ab8cdc00942ccc990

PR #9703, opened 2015-10-21
Fix Tanya's Colt targeting buildings

### LINKED ISSUE #9673: Tanya shoots at enemy structures
Tanya shoots at enemy structures instead of holding fire.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27504011-tanya-shoots-at-enemy-structures?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
This should fix #9673.

### DIFF
--- a/mods/ra/rules/fakes.yaml
+++ b/mods/ra/rules/fakes.yaml
@@ -58,7 +58,7 @@ SYRF:
 		GenericVisibility: Enemy
 		GenericStancePrefix: False
 	Targetable:
-		TargetTypes: Ground, Water
+		TargetTypes: Ground, Water, Structure
 	Building:
 		Footprint: xxx xxx xxx
 		Dimensions: 3,3
@@ -74,7 +74,7 @@ SYRF:
 SPEF:
 	Inherits: ^FakeBuilding
 	Targetable:
-		TargetTypes: Ground, Water
+		TargetTypes: Ground, Water, Structure
 	Buildable:
 		BuildPaletteOrder: 910
 		Queue: Defense
--- a/mods/ra/rules/structures.yaml
+++ b/mods/ra/rules/structures.yaml
@@ -101,7 +101,7 @@ SPEN:
 		BuildPaletteOrder: 50
 		Prerequisites: anypower, ~structures.soviet, ~techlevel.low
 	Targetable:
-		TargetTypes: Ground, Water, C4, DetonateAttack, SpyInfiltrate
+		TargetTypes: Ground, Water, Structure, C4, DetonateAttack, SpyInfiltrate
 	Building:
 		Footprint: xxx xxx xxx
 		Dimensions: 3,3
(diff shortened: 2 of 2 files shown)


reaperrr:
Updated.

After some thinking I've come to the conclusion that the 'safest' fix is to simply replace `Ground` with `Barrel` on `ValidTargets`, like already done on the warhead. 

The only reason I previously left Ground there and instead tried to disable building targeting via InvalidTargets was so players still had the option to force-fire on (empty) ground, but I guess in practice that's pretty pointless anyway.

pchote:
All the other non-melee units in all the games are able to force-fire on the ground.  It's just part of the game!  I'm not a fan of the idea of disabling it for tanya.

If there is a bug in the way InvalidTargets works, then we need to fix that instead.

reaperrr:
Ok, I've done some more testing and I think I found the real problem.

- Adding `C4` to the list of `InvalidTargets` 'fixed' it
- Changing the order of `TargetTypes` on `^BasicBuilding` to list `Structure` *before* `C4` 'fixed' it

Since the order shouldn't matter, I took a look at `WeaponInfo`.
I *believe* the problem is that line 130 is missing a `.ToHashSet()`, and a quick test seems to confirm it.

Updated accordingly.

pchote:
That is a bit worrying, because it sounds like the `.Overlaps` may be behaving differently for collections that contain duplicate entries vs collections that don't... that would be quite bogus.

RoosterDragon:
This seems dodgy - why would need need to explicitly remove the duplicates here before letting everything else chow down on them? Are you sure this isn't, say, a side-effect of the reordering of the elements? Does `.Distinct()` (which preserves the source order) have different behaviour?

There's also a several places that enumerate the target types in the same way that we'll want to check, if you current hypothesis correct

reaperrr:
> This seems dodgy - why would need need to explicitly remove the duplicates here before letting everything else chow down on them? Are you sure this isn't, say, a side-effect of the reordering of the elements?

No idea. I mostly tried this because `FrozenActor.TargetTypes` uses `.ToHashSet()` as well.

RoosterDragon:
That's probably so it can use the Overlaps method in other places, for performance.

RoosterDragon:
I was unable to see any difference when adding a `ToHashSet` myself - however if you had other changes in your YAML that may have affected things.

I posted some more info back on #9673.

reaperrr:
Ouch... somehow I completely overlooked the possibility that the individual buildings' `TargetTypes` might differ from the inherited default.

So yes, this actually isn't a code issue, our yaml definitions are flawed. Several buildings - power plants among them - simply didn't list `Structure` under `TargetTypes`.

Updated.

GraionDilach:
Another option could be giving a `TargetTypes@Spyable: ValidTargets SpyInfiltrate` to these ones.

obrakmann:
:+1: Tanya behaves now.

abcdefg30:

PR #9705, opened 2015-10-21
Misc yaml formatting cleanup

reaperrr:
Purpose of this is to reduce the 'noise' produced by applying upgrade rules.

Adds `Music:` to all maps, additionally fixes an issue with the `WithMuzzleFlash` -> `WithMuzzleOverlay` upgrade rule.

### DIFF
--- a/mods/ts/weapons/energyweapons.yaml
+++ b/mods/ts/weapons/energyweapons.yaml
@@ -124,8 +124,7 @@ Proton:
 		ImpactSound: ssplash3.aud
 		ValidImpactTypes: Water
 
-# Obelisk laser
-LaserFire:
+ObeliskLaserFire:
 	ReloadDelay: 120
 	Range: 10c512
 	Charges: true
@@ -137,8 +136,7 @@ LaserFire:
 		Damage: 250
 		DamageTypes: Prone60Percent, TriggerProne, EnergyDeath
 
-# Laser turret
-LaserFire2:
+TurretLaserFire:
 	ReloadDelay: 40
 	Range: 5c512
 	Report: LASTUR1.AUD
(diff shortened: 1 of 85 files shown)


mods/ts/weapons/energyweapons.yaml
```diff
@@ -124,7 +124,6 @@ Proton:
 		ImpactSound: ssplash3.aud
 		ValidImpactTypes: Water
 
-# Obelisk laser
```
atlimit8: This comment should be kept or `LaserFire` should be renamed to `ObeliskLaser`[`Fire`].

mods/ts/weapons/energyweapons.yaml
```diff
@@ -137,7 +136,6 @@ LaserFire:
 		Damage: 250
 		DamageTypes: Prone60Percent, TriggerProne, EnergyDeath
 
-# Laser turret
```
atlimit8: This comment should be kept or `LaserFire2` should be renamed to `LaserTurret`[`Fire`].

atlimit8:
I have to deal with this (and more) when rebasing #9324.
:+1: after changes.

mods/ts/weapons/energyweapons.yaml
penev92: To be fair we don't have any comments on any other weapon across 4 mods, so keeping it is silly. And we make a point of retaining the original names of actors and weapons as they were in the original games (as much as they suck sometimes).

atlimit8:
I guess if others are OK with dropping those comments, then there is no need for the changes.

mods/ts/weapons/energyweapons.yaml
Mailaender: Hmm, the internal weapon names don't have to match for map importing so I think this can be safely changed. Westwood's habit of enumerating weapon names isn't really a great smart idea worth copying.

reaperrr:
I'm leaning towards renaming them.

Renamed to `ObeliskLaserFire` and `TurretLaserFire` (to keep name format consistent).

atlimit8:
:+1:

abcdefg30:
Shouldn't hurt and looks ok to me. :+1:

PR #9706, opened 2015-10-21
Add Height property to BlocksProjectiles

reaperrr:
`Blockable` projectiles now compare the height of potential blockers with their own distance above ground.

Default `Height` is intentionally high so the shipping mods (or mods based on them) won't be affected, unless modders consciously make use of this.
Used to give the Tick Tank the ability to shoot over walls from certain positions (reproduces original behavior).

Possible TODO (might need help with that): Visualize height in `CombatDebugOverlay`.

### DIFF
--- a/OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
+++ b/OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
@@ -13,10 +13,11 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
-	// TODO: Add functionality like a customizable Height that is compared to projectile altitude
 	[Desc("This actor blocks bullets and missiles with 'Blockable' property.")]
 	public class BlocksProjectilesInfo : UpgradableTraitInfo
 	{
+		public readonly WDist Height = WDist.FromCells(1);
+
 		public override object Create(ActorInitializer init) { return new BlocksProjectiles(init.Self, this); }
 	}
 
@@ -27,8 +28,11 @@ namespace OpenRA.Mods.Common.Traits
 
 		public static bool AnyBlockingActorAt(World world, WPos pos)
 		{
+			var dat = world.Map.DistanceAboveTerrain(pos);
 			return world.ActorMap.GetActorsAt(world.Map.CellContaining(pos))
-				.Any(a => a.TraitsImplementing<BlocksProjectiles>().Any(Exts.IsTraitEnabled));
+				.Any(a => a.TraitsImplementing<BlocksProjectiles>()
+					.Where(t => t.Info.Height.Length >= dat.Length)
+					.Any(Exts.IsTraitEnabled));
 		}
 	}
 }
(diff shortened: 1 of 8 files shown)


atlimit8:
Needs a rebase.

reaperrr:
Rebased.

OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
```diff
@@ -13,10 +13,11 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
-	// TODO: Add functionality like a customizable Height that is compared to projectile altitude
 	[Desc("This actor blocks bullets and missiles with 'Blockable' property.")]
 	public class BlocksProjectilesInfo : UpgradableTraitInfo
 	{
+		public readonly WDist Height = WDist.FromCells(6);
```
penev92: Is there a reason for this seemingly random default value?

pchote: > Default Height is intentionally high so the shipping mods (or mods based on them) won't be affected, unless modders consciously make use of this.

penev92: Pretty sure a height of 1 WDist should be enough for the first 3 games.

reaperrr: > Pretty sure a height of 1 WDist should be enough for the first 3 games.

Unless you meant 1 cell (1024 WDist), no, definitely not. For example, the cannon armament's LocalOffset Z value for mammoth tanks is 340 WDist above ground, so even at half the distance towards target the altitude is still 170 WDist above ground. The weapon's launch offset is factored in, so even straight-flying projectiles usually won't fly that low until right in front of the target.

That being said, in hindsight 6 cells is certainly over the top, I guess 1 cell/1024 WDist should be enough.

penev92: Yeah I meant 1 cell.

pchote: Note that one cell will be less high than some arcing projectiles (artillery, etc).

reaperrr: > Note that one cell will be less high than some arcing projectiles (artillery, etc).

Arcing projectiles in the shipping mods have `Blockable: false` anyway, so that shouldn't matter.

reaperrr:
Updated, reduced default to 1 cell / 1024 WDist and added debug overlay visualization (vertical orange line).

RoosterDragon:
:+1: Appears to work nicely.

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/35c198dfa927eced69446d925cc0151cb8a2e5df...54cb033786750eacfc9593e57dc56752c31ab069)

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/35c198dfa927eced69446d925cc0151cb8a2e5df...54cb033786750eacfc9593e57dc56752c31ab069)

PR #9708, opened 2015-10-21
Charge now using ticks and fixed extra bar time formula.

ghost:
ChargeTime has been changed to ChargeDelay and is
measured in ticks rather than seconds.

The formula for working out the extra bar time has  
been fixed to work properly at different charge delays.

This addresses the comments from @penev92 and @pchote in #9700. Extra bar correctly works at different charge times.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -2243,6 +2243,19 @@ namespace OpenRA.Mods.Common.UtilityCommands
 					}
 				}
 
+				// ChargeTime is now replaced by ChargeDelay.
+				// ChargeDelay uses 500 as a default now.
+				if (engineVersion < 20151022)
+				{
+					if (depth == 2 && parentKey == "PortableChrono" && node.Key == "ChargeTime")
+					{
+						node.Key = "ChargeDelay";
+
+						if (node.Value.Value != null)
+							node.Value.Value = (Exts.ParseIntegerInvariant(node.Value.Value) * 25).ToString();
+					}
+				}
+
 				UpgradeActorRules(engineVersion, ref node.Value.Nodes, node, depth + 1);
 			}
 		}
--- a/OpenRA.Mods.RA/Traits/PortableChrono.cs
+++ b/OpenRA.Mods.RA/Traits/PortableChrono.cs
@@ -20,8 +20,8 @@ namespace OpenRA.Mods.RA.Traits
 {
 	class PortableChronoInfo : ITraitInfo
 	{
-		[Desc("Cooldown in seconds until the unit can teleport.")]
-		public readonly int ChargeTime = 20;
+		[Desc("Cooldown in ticks until the unit can teleport.")]
+		public readonly int ChargeDelay = 500;
 
 		[Desc("Can the unit teleport only a certain distance?")]
 		public readonly bool HasDistanceLimit = true;
@@ -97,7 +97,7 @@ namespace OpenRA.Mods.RA.Traits
 
 		public void ResetChargeTime()
 		{
-			chargeTick = 25 * Info.ChargeTime;
+			chargeTick = Info.ChargeDelay;
 		}
 
 		public bool CanTeleport
@@ -107,7 +107,7 @@ namespace OpenRA.Mods.RA.Traits
 
 		public float GetValue()
 		{
-			return (float)((chargeTick - (25 * Info.ChargeTime)) * (Info.ChargeTime * -0.0001));
+			return (float)(Info.ChargeDelay - chargeTick) / Info.ChargeDelay;
 		}
 
 		public Color GetColor() { return Color.Magenta; }


OpenRA.Mods.RA/Traits/PortableChrono.cs
```diff
@@ -97,7 +99,7 @@ public string VoicePhraseForOrder(Actor self, Order order)
 
 		public void ResetChargeTime()
 		{
-			chargeTick = 25 * Info.ChargeTime;
+			chargeTick = Self.World.LobbyInfo.GlobalSettings.Timestep * Info.ChargeTime;
```
pchote: We talked about this in IRC, and @r34ch is working on a fix.

OpenRA.Mods.RA/Traits/PortableChrono.cs
```diff
@@ -107,7 +109,7 @@ public bool CanTeleport
 
 		public float GetValue()
 		{
-			return (float)((chargeTick - (25 * Info.ChargeTime)) * (Info.ChargeTime * -0.0001));
+			return (float)((chargeTick - Info.ChargeDelay) * (1f / -Info.ChargeDelay));
```
pchote: This can simplify this down to `(float)(Info.ChargeDelay - chargeTick) / Info.ChargeDelay;`

OpenRA.Mods.RA/Traits/PortableChrono.cs
```diff
@@ -40,17 +40,19 @@ class PortableChronoInfo : ITraitInfo
...
 
-		public object Create(ActorInitializer init) { return new PortableChrono(this); }
+		public object Create(ActorInitializer init) { return new PortableChrono(this, init.Self); }
 	}
 
 	class PortableChrono : IIssueOrder, IResolveOrder, ITick, ISelectionBar, IOrderVoice, ISync
 	{
 		[Sync] int chargeTick = 0;
 		public readonly PortableChronoInfo Info;
+		public readonly Actor Self;
```
pchote: This shouldn't need to be public, so use `readonly Actor self;` and `this.self = self` in the constructor.

pchote: Actually, this doesn't look to be used now, so can be removed.

pchote:
This will also need an upgrade rule before merging, which somebody should be able to help you with a bit later.

r34ch:
I have added an upgrade rule. I don't know if it works or how to really test it. 

I just ran `mono OpenRA.Utility.exe ra --upgrade-mod 20150919` which returned no errors

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2243,6 +2243,17 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
 
+				// ChargeTime is now replaced by ChargeDelay.
+				// ChargeDelay uses 500 as a default value from ChargeTime 25.
+				if (engineVersion < 20151021)
+				{
+					if (depth == 2 && parentKey == "PortableChrono" && node.Key == "ChargeTime")
+						node.Key = "ChargeDelay";
+
+					if (depth == 2 && parentKey == "PortableChrono" && node.Key == "ChargeTime" && node.Value.Value != null)
+						node.Value.Value = (Convert.ToInt32(node.Value.Value) * 25).ToString();
```
atlimit8: Use `Exts.ParseIntegerInvariant(node.Value.Value)` instead of `Convert.ToInt32(node.Value.Value)`.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2243,6 +2243,17 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
 				}
 
+				// ChargeTime is now replaced by ChargeDelay.
+				// ChargeDelay uses 500 as a default value from ChargeTime 25.
+				if (engineVersion < 20151021)
+				{
+					if (depth == 2 && parentKey == "PortableChrono" && node.Key == "ChargeTime")
+						node.Key = "ChargeDelay";
+
+					if (depth == 2 && parentKey == "PortableChrono" && node.Key == "ChargeTime" && node.Value.Value != null)
```
atlimit8: This should be folded into the condition above:
```c#
					if (depth == 2 && parentKey == "PortableChrono" && node.Key == "ChargeTime")
					{
						node.Key = "ChargeDelay";

						if (node.Value.Value != null)
							node.Value.Value = (Exts.ParseIntegerInvariant(node.Value.Value) * 25).ToString();
					}
```

r34ch:
Thanks @atlimit8, updated.

atlimit8:
:+1: works.

r34ch:
Fixed comment.

abcdefg30:
Looks good to me. :+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/14244c0577fd165f241fe59b3518e9a7ac19b87e...2acc9fef3356ef81b493bd58eda60d2517ce5b85)

PR #9709, opened 2015-10-21
Rename ActorMap *UnitsAt* occurences to *ActorsAt*

reaperrr:
'Units' usually implies only mobile, non-building actors.


penev92:
:+1:

atlimit8:
> 'Units' usually implies only mobile, non-building actors.
I agree :+1: .

Issue #8086, opened 2015-05-04
Negating double inherited traits does not work


abcdefg30:
If a trait is on the default (`^Helicopter` e.g.) and on the actor (`TRAN`) then simply negating it (with `-`) won't work. Tested with `LeavesHusk` on the TD `TRAN` (`LeavesHusk` is on `TRAN` but not on the default. Adding it to default also breaks the custom removal on a map).
Also noticeable with `TRAN` in #8048 and `Selectable`. [Another example.](https://gist.github.com/abcdefg30/0e6a3f33e099f5e0efd2)

Sorry if I'm too confusing. ([Here](https://gist.github.com/abcdefg30/bb8959d4c9b9a7640a50) is my IRC spam regarding this.)

pchote:
Does it work if you add the removal twice?  The first removal will probably be removing the inherited override, so that your actor then ends up with the unmodified parent rules.

abcdefg30:
> Does it work if you add the removal twice?

No. Either the map is disabled or `Exception of type "System.InvalidOperationException": No free spawnpoint.` shows up.

pchote:
Ok.  This will require a rework of my rework.  It won't be particularly hard, but also isn't trivial.

pchote:
I have a branch that probably fixes this (https://github.com/pchote/OpenRA/commits/yaml-removals), but I don't really feel like writing unit tests or doing ingame tests for it just yet.

Mailaender:
Looks like this is already fixed? At least the `TRAN` example in nod04b https://github.com/OpenRA/OpenRA/commit/c23ee1be2e58bcc512ceb106d56a619bab48eb33#diff-cc6e2c8cab65841973b60343a8c32f0eR548

PR #9710, opened 2015-10-21
Fix yaml removals

### LINKED ISSUE #9196: yaml inheritance breaks in conjunction with yaml removals
Adding the following piece of yaml to a map.yaml...

```yaml
^A:
        ProximityCaptor:
DOG:
        Inherits@a: ^A
```

... causes breakage:
```
OpenRA.Utility(1,1): Error: Actor type dog: Bogus yaml removals: TakeCover
```

It works as expected on https://github.com/pchote/OpenRA/tree/yaml-removals

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26296390-yaml-inheritance-breaks-in-conjunction-with-yaml-removals?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
Fixes #9196

### DIFF
--- a/OpenRA.Test/OpenRA.Game/ActorInfoTest.cs
+++ b/OpenRA.Test/OpenRA.Game/ActorInfoTest.cs
@@ -28,6 +28,10 @@ namespace OpenRA.Test
 	class MockEInfo : MockTraitInfo, Requires<MockFInfo> { }
 	class MockFInfo : MockTraitInfo, Requires<MockDInfo> { }
 
+	class MockA2Info : MockTraitInfo { }
+	class MockB2Info : MockTraitInfo { }
+	class MockC2Info : MockTraitInfo { }
+
 	[TestFixture]
 	public class ActorInfoTest
 	{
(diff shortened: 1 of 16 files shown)


RoosterDragon:
Does it also fix #8086?

pchote:
It should do - that was the initial goal of the branch.

atlimit8:
NUnit test case(s) should be added as well to catch future breaking.

atlimit8:
I'll relent on insisting on unit test for now.
See #9733.

RoosterDragon:
I would like to add some unit tests for this.

I have a branch for bleed and a rebase of this PR on, https://github.com/RoosterDragon/OpenRA/tree/bleed-yaml-test and https://github.com/RoosterDragon/OpenRA/tree/yaml-removals-test respectively.

There is a test for #8086 (TraitCanBeRemovedAfterMultipleInheritance) which fails on bleed, but passes with this PR - so that's good.

I have created a test for #9196 (TraitInheritanceAndRemovalCanBeComposed), but it passes on both branches and I can't see where I'm messing up. I'd like to prove it fails on bleed the same as reported in the issue. Can anybody help?

Mailaender:
I am under the impression that https://github.com/OpenRA/OpenRA/issues/8086#issuecomment-158775732 is already fixed or at least I am having trouble finding test cases for this as they have been worked around one way or another.

RoosterDragon:
The unit test I wrote for it fails on bleed, if you want to try that.

Mailaender:
I suggest you add the unit tests to this pull request then.

obrakmann:
Added Rooster's unit tests. The `TraitInheritanceAndRemovalCanBeComposed()` test already finishes successfully on bleed, though, so something's not quite right somewhere.

RoosterDragon:
I tried to recreate the yaml used in the example from #9196, but maybe I messed something up or didn't include enough of it to replicate the problem.

If anybody has any suggestions on tweaking the test case so it fails on bleed, now would be a good time. :)

I assume #9196 hasn't been accidentally fixed prior to this PR?

obrakmann:
No, it hasn't.

I still get `Failed to load rules for 02:   Five to one with error Actor type dog: Bogus yaml removals: TakeCover` with this diff:
```diff
diff --git a/mods/ra/maps/allies-02/map.yaml b/mods/ra/maps/allies-02/map.yaml
index a1b3ec1..f260c5f 100644
--- a/mods/ra/maps/allies-02/map.yaml
+++ b/mods/ra/maps/allies-02/map.yaml
@@ -826,6 +826,10 @@ Rules:
        PDOF:
                Buildable:
                        Prerequisites: ~disabled
+       ^A:
+               ProximityCaptor:
+       DOG:
+               Inherits@a: ^A
 
 Sequences:
 
```

RoosterDragon:
I sent you a fix in obrakmann/OpenRA#3.

obrakmann:
Updated with the mentioned fix.

RoosterDragon:
This is ready to go then.

abcdefg30:
Needs a rebase though.

RoosterDragon:
FWIW the passing tests are effectively my +1, but since I wrote them I won't make that proper.

RoosterDragon:
On second thought, the tests aren't that complex or anything. So they justify this PR for me. :+1:

OpenRA.Test/OpenRA.Game/ActorInfoTest.cs
```diff
@@ -88,5 +92,65 @@ public void TraitsInConstructOrderC()
...
+			var overrideYaml = @"
+Actor:
+	-MockA2
+";
+
+			var actorInfo = CreateActorInfoFromYaml("Actor", null, baseYaml, overrideYaml);
+			Assert.IsFalse(actorInfo.HasTraitInfo<MockA2Info>(), "Actor should not have the MockA2 trait, but does.");
+		}
+
+		// This needs to match the logic used in RulesetCache.LoadYamlRules
```
Mailaender: Ugh, I guess this will bit rot fast. No way to re-use it?

obrakmann: My expectation is that it wouldn't bitrot fast. This part of the code is not supposed to change often anymore (as compared to, say, six years ago). And if it does, chances are that the test would fail. Also the code isn't copied verbatim, but just does the same thing in spirit, so to speak, so reusing it would require rewriting the code in `RulesetCache.LoadYamlRules` to make it reusable first. And that would be too risky for too little gain, IMO.

Mailaender: Sounds convincing.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5647f3329ad13648122132a1e52532e1bbba9368...9d166f4759d56dc31415296884c15bb9e213fbe8)

abcdefg30:
It looks to me like this broke Fort Lonestar?

abcdefg30:
> 17:19:22 (abcdefg30) anyway, i wonder how yaml inheritance affects that insta win
17:19:48 (abcdefg30) oh
17:19:51 (abcdefg30) hehe
17:20:01 (abcdefg30) snipers have -MustBeDestroyed or sth like that probably
17:20:08 (abcdefg30) because of their cloak
17:20:46 (abcdefg30) yep
17:20:55 (abcdefg30) that is stupid
17:22:05 (Phroh|work) abcdefg30: What is stupid about that?
17:22:29 (abcdefg30) you have MustBeDestroyed: on ^Infantry
17:22:37 (abcdefg30) then -MustBeDestroyed: on SNIPER
17:22:49 (abcdefg30) and then MustBeDestroyed: on SNIPER.soviets
17:23:03 (abcdefg30) which should give us at least one MustBeDestroyed:

abcdefg30:
Anyway & for the record: I have a fix ready but won't push it until I'm done with the rebalancing.

PR #9714, opened 2015-10-22
Fix misleading Utility message.

### LINKED ISSUE #9713: Utility incorrectly reports that a command does not exist.
screen shot 2015-10-21 at 9 23 39 pm

The command *does* exist, but an `ArgumentException` is being raised by a malformed `Substring`.

```csharp
			catch (Exception e)
			{
				Log.AddChannel("utility", "utility.log");
				Log.Write("utility", "Received args: {0}", args.JoinWith(" "));
				Log.Write("utility", "{0}", e);

			    if (e is ArgumentException)
			        Console.WriteLine("No such command '{0}'", args[0]);
			    else
			    {
                    Console.WriteLine("Error: Utility application crashed. See utility.log for details");
                    throw;
			    }
			}
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27600859-utility-incorrectly-reports-that-a-command-does-not-exist?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

phrohdoh:
Closes #9713.

### DIFF
--- a/OpenRA.Utility/Program.cs
+++ b/OpenRA.Utility/Program.cs
@@ -66,11 +66,12 @@ namespace OpenRA.Utility
 
 			try
 			{
-				if (!actions.ContainsKey(args[0]))
-					throw new NoSuchCommandException(args[0]);
+				var command = args[0];
+				if (!actions.ContainsKey(command))
+					throw new NoSuchCommandException(command);
 
-				var action = actions[args[0]].Key;
-				var validateActionArgs = actions[args[0]].Value;
+				var action = actions[command].Key;
+				var validateActionArgs = actions[command].Value;
 
 				if (validateActionArgs.Invoke(args))
 				{
@@ -78,8 +79,8 @@ namespace OpenRA.Utility
 				}
 				else
 				{
-					Console.WriteLine("Invalid arguments for '{0}'", args[0]);
-					GetActionUsage(args[0], action);
+					Console.WriteLine("Invalid arguments for '{0}'", command);
+					GetActionUsage(command, action);
 				}
 			}
 			catch (Exception e)
--- a/OpenRA.Utility/Program.cs
+++ b/OpenRA.Utility/Program.cs
@@ -15,6 +15,16 @@ using OpenRA.FileSystem;
 
 namespace OpenRA.Utility
 {
+	public class NoSuchCommandException : Exception
+	{
+		public readonly string Command;
+		public NoSuchCommandException(string command)
+			: base("No such command '{0}'".F(command))
+		{
+			Command = command;
+		}
+	}
+
 	class Program
 	{
 		static void Main(string[] args)
@@ -57,7 +67,7 @@ namespace OpenRA.Utility
 			try
 			{
 				if (!actions.ContainsKey(args[0]))
-					throw new ArgumentException();
+					throw new NoSuchCommandException(args[0]);
 
 				var action = actions[args[0]].Key;
 				var validateActionArgs = actions[args[0]].Value;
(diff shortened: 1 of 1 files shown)


atlimit8:
I don't like identifying exception type by checking the message string.
A subclass should be used:
```c#
public class NoSuchCommandException : ArgumentException
{
	public NoSuchCommandException(string command)
		: base("No such command '{0}'", command) { }
}
```
```diff
@@ -57,7 +57,7 @@ static void Main(string[] args)
 			try
 			{
 				if (!actions.ContainsKey(args[0]))
-					throw new ArgumentException();
+					throw new NoSuchCommandException(args[0]);
 
 				var action = actions[args[0]].Key;
 				var validateActionArgs = actions[args[0]].Value;
@@ -78,8 +78,10 @@ static void Main(string[] args)
 				Log.Write("utility", "Received args: {0}", args.JoinWith(" "));
 				Log.Write("utility", "{0}", e);
 
-			    if (e is ArgumentException)
-			        Console.WriteLine("No such command '{0}'", args[0]);
+				if (e is NoSuchCommandException)
+				{
+					Console.WriteLine(e.Message);
+				}
 			    else
 			    {
                     Console.WriteLine("Error: Utility application crashed. See utility.log for details");
```

Phrohdoh:
That's fair.
I'll fix this soon, it is already 12:15am here.

Phrohdoh:
Updated.

atlimit8:
:+1: 
The `command` variable also makes the code a little more readable.

OpenRA.Utility/Program.cs
```diff
@@ -15,6 +15,16 @@
 
 namespace OpenRA.Utility
 {
+	public class NoSuchCommandException : ArgumentException
```
RoosterDragon: This should inherit plain `Exception`.

OpenRA.Utility/Program.cs
```diff
@@ -78,13 +89,13 @@ static void Main(string[] args)
...
 
-			    if (e is ArgumentException)
-			        Console.WriteLine("No such command '{0}'", args[0]);
-			    else
-			    {
-                    Console.WriteLine("Error: Utility application crashed. See utility.log for details");
-                    throw;
-			    }
+				if (e is NoSuchCommandException)
+					Console.WriteLine(e.Message);
```
RoosterDragon: This cleanup enables you to `catch (NoSuchCommandException)` and do the error handling there, which would be better.

OpenRA.Utility/Program.cs
atlimit8: How is `NoSuchCommandException` not an `ArgumentException` when it is about an argument.

RoosterDragon: `ArgumentException` defines `ParamName` which hasn't been set at all here, so either that's an oversight or `ArgumentException` doesn't apply in the first place.

ScottNZ: Furthermore, ArgumentExceptions are used for method arguments/parameters, not arguments for commands in OpenRA.Utility

atlimit8: OK.

OpenRA.Utility/Program.cs
atlimit8: Then log output would have to be copied or placed in a method.
```c#
 				Log.Write("utility", "Received args: {0}", args.JoinWith(" "));
 				Log.Write("utility", "{0}", e);
```

Phrohdoh: I'd rather keep the amount of changes introduced here to a minimum.

RoosterDragon: Good points.

OpenRA.Utility/Program.cs
Phrohdoh: Fixed.

Phrohdoh:
`NoSuchCommandException` now inherits from `Exception`.

RoosterDragon:
:+1:

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/25ef0d4f364aa966dea233e24d7335df8630009f...3d276d29317d7ac22af2e4e3ece69bc3842f0ca7)

PR #9716, opened 2015-10-22
Disable hovering for landed aircraft

teees:


penev92:
Looks good :+1: 

Note that aircraft still hovers while landing/taking off, but that can be handled through the Aircraft trait once it gets its long-needed overhaul. I'll open an issue about that.

atlimit8:
@penev92 perhaps setting `MinAirborneAltitude` > 1 would suffice. I didn't change any behavior when I added that property, figuring that it would be tuned later.

Mailaender:

Mailaender:
Thanks!

PR #9718, opened 2015-10-22
Check HVA files validity at load time

penev92:
Currently if you provide the game with a voxel with bogus HVA file, the game will crash [here](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Graphics/VoxelRenderer.cs#L203) after `Util.MatrixInverse(t)` returns `null` and `Util.MatrixMultiply()` chokes on it.
That, of course, happens ingame when you create the actor, which is far from ideal.

The reason in the case I debugged ( reported by @ABrandau ) was a bogus transformation matrix.
The exception, of course, was an NRE with no information whatsoever.

This throws at load time (as soon as you start up the game), and it gives a proper exception, which will hopefully help modders find their mistake.

### DIFF
--- a/OpenRA.Game/FileFormats/HvaReader.cs
+++ b/OpenRA.Game/FileFormats/HvaReader.cs
@@ -9,7 +9,10 @@
 
 #endregion
 
+using System;
 using System.IO;
+using System.Linq;
+using OpenRA.Graphics;
 
 namespace OpenRA.FileFormats
 {
@@ -19,7 +22,7 @@ namespace OpenRA.FileFormats
 		public readonly uint LimbCount;
 		public readonly float[] Transforms;
 
-		public HvaReader(Stream s)
+		public HvaReader(Stream s, string fileName)
 		{
 			// Index swaps for transposing a matrix
 			var ids = new byte[] { 0, 4, 8, 12, 1, 5, 9, 13, 2, 6, 10, 14 };
@@ -31,6 +34,8 @@ namespace OpenRA.FileFormats
 			// Skip limb names
 			s.Seek(16 * LimbCount, SeekOrigin.Current);
 			Transforms = new float[16 * FrameCount * LimbCount];
+
+			var testMatrix = new float[16];
 			for (var j = 0; j < FrameCount; j++)
 				for (var i = 0; i < LimbCount; i++)
 				{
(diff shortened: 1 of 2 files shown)


OpenRA.Game/FileFormats/HvaReader.cs
```diff
@@ -43,13 +46,19 @@ public HvaReader(Stream s)
 
 					for (var k = 0; k < 12; k++)
 						Transforms[c + ids[k]] = s.ReadFloat();
+
+					var testMatrix = new float[16];
+					Array.Copy(Transforms, 16 * (LimbCount * j + i), testMatrix, 0, 16);
+					if (Util.MatrixInverse(testMatrix) == null)
+						throw new InvalidDataException("The transformation matrix for HVA file `{0}` limb {1} frame {2} is invalid!"
```
RoosterDragon: Maybe "...is invalid because it is not invertible"?

GraionDilach: Please use section instead of limb - latter makes no sense.

penev92:
Changed.

OpenRA.Game/FileFormats/HvaReader.cs
```diff
@@ -43,13 +46,20 @@ public HvaReader(Stream s)
 
 					for (var k = 0; k < 12; k++)
 						Transforms[c + ids[k]] = s.ReadFloat();
+
+					var testMatrix = new float[16];
```
RoosterDragon: Can you declare this outside the loop so we don't create a bunch of throwaway arrays? (Sorry for missing this on the first review)

RoosterDragon:
:+1: LGTM after the minor fix.

penev92:
Like so?

RoosterDragon:
Yup.

abcdefg30:
Code looks good to me, too. :+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3a79ae11560f0faa92511a96a0e554567f68ce8f...1f83e1350ebc05b4b013d32f5440260da83d47b4)

Mailaender:
For some reason Coverity now 2 new "Dereference null return value" problems. It looks like those which this pull request tried to catch and re-throw. I suggest you investigate and fix or mark them as false positives.

PR #9720, opened 2015-10-22
Fix a crash in CheckPlayers when an actor doesn't define an Owner

penev92:


OpenRA.Mods.Common/Lint/CheckPlayers.cs
```diff
@@ -59,7 +59,14 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 			foreach (var kv in map.ActorDefinitions)
 			{
 				var actorReference = new ActorReference(kv.Value.Value, kv.Value.ToDictionary());
-				var ownerName = actorReference.InitDict.Get<OwnerInit>().PlayerName;
+				var ownerInit = actorReference.InitDict.GetOrDefault<OwnerInit>();
+				if (ownerInit == null)
+				{
+					emitError("Actor {0} is not owned by any player.".F(kv.Key));
+					continue;
```
pchote: This is a bit ugly.  You can make the check below an `else if` and avoid the continue.

penev92:
Changed.

OpenRA.Mods.Common/Lint/CheckPlayers.cs
```diff
@@ -59,9 +59,17 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 			foreach (var kv in map.ActorDefinitions)
 			{
 				var actorReference = new ActorReference(kv.Value.Value, kv.Value.ToDictionary());
-				var ownerName = actorReference.InitDict.Get<OwnerInit>().PlayerName;
-				if (!playerNames.Contains(ownerName))
-					emitError("Actor {0} is owned by unknown player {1}.".F(actorReference.Type, ownerName));
+				var ownerInit = actorReference.InitDict.GetOrDefault<OwnerInit>();
+				if (ownerInit == null)
+				{
```
abcdefg30: You can drop the braces around this `if`, I think.

abcdefg30:

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f1fd62af3e567a17c8981650bb3ba31819bd30fd...61badc4d227112d91d19d9ee72382b91fccf73e4)

PR #9722, opened 2015-10-22
use right cursor for ion and nuke

teees:
nuke and ion cursors where switched;
ion cannon was using default cursor


reaperrr:
:+1:

PR #9723, opened 2015-10-22
flip or mirror sprites with negative width and height

evgeniysergeev:
currently sprite with negative width or height looks on the screen the same as with normal width and height.
This fix allow to use negative width and/or height to flip/mirror sprite. 
It assign correct points for textures in a sheet.


reaperrr:
Code looks good to me and spotted no issues on any of the 4 shellmaps :+1:

penev92:
:+1:

PR #9724, opened 2015-10-22
Add 'Flip/Mirror' functionality for sequences

evgeniysergeev:
example

```
quad:                                                    
        idle: UNITS.SHP                                        
                Frames:  0, 1, 2, 3, 4,  3,  2,  1             
                FlipX:   1, 1, 1, 1, 1, -1, -1, -1             
                Facings: -8                                    
        icon: SHAPES.SHP                                       
                Start: 74                                      
                Offset: 0,0                                    
```

### DIFF
--- a/OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
+++ b/OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
@@ -131,26 +131,27 @@ namespace OpenRA.Mods.Common.Graphics
 				// Apply offset to each sprite in the sequence
 				// Different sequences may apply different offsets to the same frame
 				var src = GetSpriteSrc(modData, tileSet, sequence, animation, info.Value, d);
-				sprites = cache[src].Select(
+				Sprite[] cacheSprites = cache[src];
+				sprites = cacheSprites.Select(
 					s => new Sprite(s.Sheet, s.Bounds, s.Offset + offset, s.Channel, blendMode)).ToArray();
 
 				if (FlipX != null)
 				{
-					spritesFlipX = cache[src].Select(
+					spritesFlipX = cacheSprites.Select(
 						s => new Sprite(s.Sheet, Rectangle.FromLTRB(s.Bounds.Right, s.Bounds.Top, s.Bounds.Left, s.Bounds.Bottom),
 							s.Offset + offset, s.Channel, blendMode)).ToArray();
 				}
 
 				if (FlipY != null)
 				{
-					spritesFlipY = cache[src].Select(
+					spritesFlipY = cacheSprites.Select(
 						s => new Sprite(s.Sheet, Rectangle.FromLTRB(s.Bounds.Left, s.Bounds.Bottom, s.Bounds.Right, s.Bounds.Top),
 							s.Offset + offset, s.Channel, blendMode)).ToArray();
 				}
 
 				if (FlipX != null && FlipY != null)
 				{
-					spritesFlipXY = cache[src].Select(
+					spritesFlipXY = cacheSprites.Select(
 						s => new Sprite(s.Sheet, Rectangle.FromLTRB(s.Bounds.Right, s.Bounds.Bottom, s.Bounds.Left, s.Bounds.Top),
 							s.Offset + offset, s.Channel, blendMode)).ToArray();
 				}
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
```diff
@@ -126,6 +134,27 @@ public DefaultSpriteSequence(ModData modData, TileSet tileSet, SpriteCache cache
 				sprites = cache[src].Select(
 					s => new Sprite(s.Sheet, s.Bounds, s.Offset + offset, s.Channel, blendMode)).ToArray();
 
+				if (FlipX != null)
```
reaperrr: Wouldn't it be better to add `&& FlipY == null` here?

OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
```diff
@@ -126,6 +134,27 @@ public DefaultSpriteSequence(ModData modData, TileSet tileSet, SpriteCache cache
...
+				{
+					spritesFlipX = cache[src].Select(
+						s => new Sprite(s.Sheet, Rectangle.FromLTRB(s.Bounds.Right, s.Bounds.Top, s.Bounds.Left, s.Bounds.Bottom),
+							s.Offset + offset, s.Channel, blendMode)).ToArray();
+				}
+
+				if (FlipY != null)
```
reaperrr: And `&& FlipX == null` here?

OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
```diff
@@ -126,6 +134,27 @@ public DefaultSpriteSequence(ModData modData, TileSet tileSet, SpriteCache cache
...
+				{
+					spritesFlipY = cache[src].Select(
+						s => new Sprite(s.Sheet, Rectangle.FromLTRB(s.Bounds.Left, s.Bounds.Bottom, s.Bounds.Right, s.Bounds.Top),
+							s.Offset + offset, s.Channel, blendMode)).ToArray();
+				}
+
+				if (FlipX != null && FlipY != null)
```
reaperrr: Because otherwise all 3 don't exclude each other, wouldn't that lead to a total or 3 sheets being created?

reaperrr:
```
OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs(170,27): error CS0103: The name 'src' does not exist in the current context

OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs(177,27): error CS0103: The name 'src' does not exist in the current context

OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs(184,28): error CS0103: The name 'src' does not exist in the current context
```

OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
evgeniysergeev: no. both will be created. 
Yes memory overhead here. but current sprites array also produce a lot of unused entries in it

OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
evgeniysergeev: both 3 will be created. that will allow to use something like this:

FlipX: 1, 1, -1, -1
FlipY: 1, -1, -1, 1

(all combinations: no flip, flipY, mirror, flipX)

reaperrr: Oh, I see.

pchote:
I think it would be cleaner to use a (sub-)sequence wide  `FlipX` boolean combined with aggregate sequences:
```yaml
	idle:
		Combine:
			a: UNITS.SHP
			 	Start: 0
				Length: 5
			b: UNITS.SHP
				Frames: 3, 2, 1
				FlipX: true
        Facings: -8                                    
	icon: SHAPES.SHP                                       
        Start: 74
        Offset: 0, 0
```

evgeniysergeev:
ok. this looks better

evgeniysergeev:
closed. solution will be in another pr

PR #9730, opened 2015-10-23
Fake structures detonate

ghost:
First commit adds the `HealthPercentage` property to the `Explodes` class. This allows actors to be killed once they fall below a certain percentage of their health.

The second commit applies this functionality to RA fake structures. Most fakes now explode at 10% of their real counterparts max health.

Before
norm_fake

After
new_fake

Previously a single rifle or flak burst would show a large chunk out of the visual health bar. Now fakes take damage at the same rate as their real counterparts. In the gifs above, both structures are destroyed after 100hp has been dealt (two grenades in this case). The gifs show the damage done from a single grenade.

WF 100hp -> 150hp
CY 100hp ->150hp
SPEN 100hp -
NY 100hp -
DOME 100hp -
ATEK 100hp -> 40hp (perhaps this could do with a buff?)

I have however buffed the fake msilo and fake pdox to explode at 50% since only 1 of each can be built, and it is pretty easy to get a couple rockets off from a paradrop now that you can select them in the air.

PDOX 100hp -> 500hp
MSILO 100hp -> 500hp

The one issue I have noticed that I discussed briefly on IRC, is that when a players fake structure is destroyed by an opponent, that player gains both a kill and a loss while the opponent who killed the actor gains nothing.

If anyone has ideas on how to solve this, please detail.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Explodes.cs
+++ b/OpenRA.Mods.Common/Traits/Explodes.cs
@@ -18,7 +18,7 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("This actor explodes when killed.")]
-	public class ExplodesInfo : ITraitInfo, IRulesetLoaded
+	public class ExplodesInfo : ITraitInfo, IRulesetLoaded, Requires<HealthInfo>
 	{
 		[WeaponReference, FieldLoader.Require, Desc("Weapon to use for explosion if ammo/payload is loaded.")]
 		public readonly string Weapon = "UnitExplode";
@@ -32,13 +32,16 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Chance that this actor will explode at all.")]
 		public readonly int Chance = 100;
 
+		[Desc("Health level at which actor will explode.")]
+		public readonly int DamageThreshold = 0;
+
 		[Desc("DeathType(s) to apply upon explosion.")]
 		public readonly HashSet<string> DeathType = new HashSet<string>();
 
 		public WeaponInfo WeaponInfo { get; private set; }
 		public WeaponInfo EmptyWeaponInfo { get; private set; }
 
-		public object Create(ActorInitializer init) { return new Explodes(this); }
+		public object Create(ActorInitializer init) { return new Explodes(this, init.Self); }
 		public void RulesetLoaded(Ruleset rules, ActorInfo ai)
 		{
 			WeaponInfo = string.IsNullOrEmpty(Weapon) ? null : rules.Weapons[Weapon.ToLowerInvariant()];
@@ -46,11 +49,17 @@ namespace OpenRA.Mods.Common.Traits
 		}
 	}
 
-	public class Explodes : INotifyKilled
+	public class Explodes : INotifyKilled, INotifyDamage
 	{
 		readonly ExplodesInfo info;
 
-		public Explodes(ExplodesInfo info) { this.info = info; }
+		readonly Health health;
+
+		public Explodes(ExplodesInfo info, Actor self)
+		{
+			this.info = info;
+			health = self.Trait<Health>();
+		}
 
 		public void Killed(Actor self, AttackInfo e)
 		{
@@ -80,6 +89,15 @@ namespace OpenRA.Mods.Common.Traits
 			var shouldExplode = self.TraitsImplementing<IExplodeModifier>().All(a => a.ShouldExplode(self));
 			var useFullExplosion = self.World.SharedRandom.Next(100) <= info.LoadedChance;
 			return (shouldExplode && useFullExplosion) ? info.WeaponInfo : info.EmptyWeaponInfo;
+		}
+
+		public void Damaged(Actor self, AttackInfo e)
+		{
+			if (info.DamageThreshold == 0)
+				return;
+
+			if (health.HP * 100 < info.DamageThreshold * health.MaxHP)
+				self.World.AddFrameEndTask(w => self.Kill(e.Attacker));
 		}
 	}
 }
--- a/mods/ra/rules/fakes.yaml
+++ b/mods/ra/rules/fakes.yaml
@@ -19,6 +19,10 @@ FACF:
 		Image: FACT
 	Valued:
 		Cost: 250
+	Health:
+		HP: 1500
+	Armor:
+		Type: Wood
 
 WEAF:
 	Inherits: ^FakeBuilding
@@ -43,6 +47,10 @@ WEAF:
 		Sequence: build-top
 	Valued:
 		Cost: 200
+	Health:
+		HP: 1500
+	Armor:
+		Type: Wood
 
 SYRF:
 	Inherits: ^FakeBuilding
@@ -68,6 +76,10 @@ SYRF:
 		Image: SYRD
 	Valued:
 		Cost: 100
+	Health:
+		HP: 1000
+	Armor:
+		Type: Light
 	EditorTilesetFilter:
 		ExcludeTilesets: INTERIOR
 
@@ -95,6 +107,10 @@ SPEF:
 		Image: SPEN
 	Valued:
 		Cost: 100
+	Health:
+		HP: 1000
+	Armor:
+		Type: Light
 	EditorTilesetFilter:
 		ExcludeTilesets: INTERIOR
 
@@ -119,6 +135,10 @@ DOMF:
 		Image: DOME
 	Valued:
 		Cost: 180
+	Health:
+		HP: 1000
+	Armor:
+		Type: Wood
 	RequiresPower:
 	DisabledOverlay:
 
@@ -143,6 +163,10 @@ ATEF:
 		Image: ATEK
 	Valued:
 		Cost: 150
+	Health:
+		HP: 400
+	Armor:
+		Type: Wood
 	RequiresPower:
 	DisabledOverlay:
 
@@ -169,6 +193,12 @@ PDOF:
 		HasMinibib: Yes
 	Valued:
 		Cost: 150
+	Health:
+		HP: 1000
+	Armor:
+		Type: Wood
+	Explodes:
+		DamageThreshold: 50
 	RequiresPower:
 	DisabledOverlay:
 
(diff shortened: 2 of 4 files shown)


cjshmyr:
Great idea and solves the true problem with fake structures. I suggest implementing `INotifyDamaged` rather than ticking to check health

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -81,5 +84,16 @@ WeaponInfo ChooseWeaponForExplosion(Actor self)
 			var useFullExplosion = self.World.SharedRandom.Next(100) <= info.LoadedChance;
 			return (shouldExplode && useFullExplosion) ? info.WeaponInfo : info.EmptyWeaponInfo;
 		}
+
+		public void Tick(Actor self)
+		{
+			if (info.HealthPercentage != 0)
+			{
+				var health = self.Trait<Health>();
```
penev92: Make this a class member variable and get the trait in the constructor.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -81,5 +84,16 @@ WeaponInfo ChooseWeaponForExplosion(Actor self)
 			var useFullExplosion = self.World.SharedRandom.Next(100) <= info.LoadedChance;
 			return (shouldExplode && useFullExplosion) ? info.WeaponInfo : info.EmptyWeaponInfo;
 		}
+
+		public void Tick(Actor self)
+		{
+			if (info.HealthPercentage != 0)
```
penev92: Please invert this to
```
if (info.HealthPercentage == 0)
    return;
```
Perhaps even do `<= 0`.

atlimit8:
Still should use `INotifyDamaged` instead of `ITick`.

r34ch:
Thanks @cjshmyr, @penev92 - updated.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -46,11 +49,17 @@ public void RulesetLoaded(Ruleset rules, ActorInfo ai)
...
 	{
 		readonly ExplodesInfo info;
 
-		public Explodes(ExplodesInfo info) { this.info = info; }
+		readonly Health health;
+
+		public Explodes(ExplodesInfo info, Actor self)
+		{
+			this.info = info;
+			health = self.Trait<Health>();
```
pchote: You must declare `Requires<HealthInfo>` on the TraitInfo if you use this from the constructor.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -81,5 +90,14 @@ WeaponInfo ChooseWeaponForExplosion(Actor self)
 			var useFullExplosion = self.World.SharedRandom.Next(100) <= info.LoadedChance;
 			return (shouldExplode && useFullExplosion) ? info.WeaponInfo : info.EmptyWeaponInfo;
 		}
+
+		public void Damaged(Actor self, AttackInfo e)
+		{
+			if (info.HealthPercentage == 0)
+				return;
+
+			if ((((float)health.HP / (float)health.MaxHP) * 100) < 100 - info.HealthPercentage)
```
pchote: You can't use floating point calculations anywhere in simulation code.  This leads to non-deterministic behaviour and desyncs.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -32,13 +32,16 @@ public class ExplodesInfo : ITraitInfo, IRulesetLoaded
 		[Desc("Chance that this actor will explode at all.")]
 		public readonly int Chance = 100;
 
+		[Desc("Health level at which actor will explode.")]
+		public readonly int HealthPercentage = 0;
```
pchote: I suggest changing this to `DamageThreshold` to match `DamagedWithoutFoundation`.

OpenRA.Mods.Common/Traits/Explodes.cs
pchote: You can drop the floats by using `if (100 * health.HP < info.DamageThreshold * health.MaxHP)`

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -81,5 +90,14 @@ WeaponInfo ChooseWeaponForExplosion(Actor self)
...
 			return (shouldExplode && useFullExplosion) ? info.WeaponInfo : info.EmptyWeaponInfo;
 		}
+
+		public void Damaged(Actor self, AttackInfo e)
+		{
+			if (info.HealthPercentage == 0)
+				return;
+
+			if ((((float)health.HP / (float)health.MaxHP) * 100) < 100 - info.HealthPercentage)
+				self.Kill(self);
```
pchote: You can use (IIRC) `e.Attacker` to attribute the kill correctly.

r34ch:
Thanks everyone for all the input. 

The remaining issues now are that an enemy scores 2 kills on a fake building while you receive 2 deaths and buildings are demolished with a stronger screen shake than all other buildings.

I imagine both of these issues are hinting at the fact that the actor is being killed twice?

pchote:
Adding to the release milestone, as this should help address the balance concerns about france being underpowered, without risking any major side-effects.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -46,11 +49,17 @@ public void RulesetLoaded(Ruleset rules, ActorInfo ai)
 		}
 	}
 
-	public class Explodes : INotifyKilled
+	public class Explodes : INotifyKilled, INotifyDamage, Requires<HealthInfo>
```
pchote: `Requires<HealthInfo>` needs to be on `ExplodesInfo` above.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -81,5 +90,14 @@ WeaponInfo ChooseWeaponForExplosion(Actor self)
 			var useFullExplosion = self.World.SharedRandom.Next(100) <= info.LoadedChance;
 			return (shouldExplode && useFullExplosion) ? info.WeaponInfo : info.EmptyWeaponInfo;
 		}
+
+		public void Damaged(Actor self, AttackInfo e)
+		{
+			if (info.DamageThreshold == 0)
+				return;
+
+			if ((health.HP * 100) < (100 - info.DamageThreshold) * health.MaxHP)
```
pchote: This makes the damage threshold a count down (e.g. "explode when i'm damaged more than 10%"), while our convention and the description in the TraitInfo is to specify the *remaining* health.  Can you change this to `info.DamageThreshold * health.MaxHP`, and then adjust the rules to 90?

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -81,5 +90,14 @@ WeaponInfo ChooseWeaponForExplosion(Actor self)
...
+				self.Kill(e.Attacker);
```
pchote: > The remaining issues now are that an enemy scores 2 kills on a fake building while you receive 2 deaths and buildings are demolished with a stronger screen shake than all other buildings.

These happen because of the way that we detect how weapons have killed a target (turns out that our damage handling code is not reentrant).  The short version is: replace this line with `self.World.AddFrameEndTask(w => self.Kill(e.Attacker));` to fix the quoted problems.

pchote:
I've tested that my above suggestions fix the double-killing and the crash in TD (which nobody else has reported yet).  Should be good to go once they are fixed.

r34ch:
Thanks @pchote, updated.

mods/ra/rules/fakes.yaml
```diff
@@ -169,6 +193,13 @@ PDOF:
 		HasMinibib: Yes
 	Valued:
 		Cost: 150
+	Health:
+		HP: 1000
+	Armor:
+		Type: Wood
+	Explodes:
+		Weapon: Demolish
```
pchote: This is already specified in the default, so you don't need to repeat it here

mods/ra/rules/fakes.yaml
```diff
@@ -193,6 +224,13 @@ MSLF:
 		Image: MSLO
 	Valued:
 		Cost: 250
+	Health:
+		HP: 1000
+	Armor:
+		Type: Wood
+	Explodes:
+		Weapon: Demolish
```
pchote: This one too.

OpenRA.Mods.Common/Traits/Explodes.cs
```diff
@@ -81,5 +90,14 @@ WeaponInfo ChooseWeaponForExplosion(Actor self)
 			var useFullExplosion = self.World.SharedRandom.Next(100) <= info.LoadedChance;
 			return (shouldExplode && useFullExplosion) ? info.WeaponInfo : info.EmptyWeaponInfo;
 		}
+
+		public void Damaged(Actor self, AttackInfo e)
+		{
+			if (info.DamageThreshold == 0)
+				return;
+
+			if ((health.HP * 100) < (info.DamageThreshold * health.MaxHP))
```
pchote: Unnecessary parens.  You can simplify this to `if (health.HP * 100 < info.DamageThreshold * health.MaxHP)`

r34ch:
Updated. Fixed related issue on mission Monster Tank Madness.

pchote:
Looks good to me now, thanks! :+1:

Could you please squash the last two commits together? We try to keep it so that each commit works on its own independently.

Suggested changelog (under the RA balance section): Fake structures are no longer revealed by their health bar.

r34ch:
Squashed and updated the commit message.

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f1dacc5b4722a629fcc3f1bd7ecbd6edffad3f1d...109a970bab5d1a86f32ef166d4b6dcd2e6dea0be)

epicelite:
Ok but like...

Why is there a fake sub pen?

ScottNZ:
Because it was present in the original game and had icon art. It isn't normally buildable in OpenRA though.

SoScared:
These health bar changes are fantastic. The health bar buff on the Nuke Silo and Chrono (PDOX?) will make France more resilient in the late game. Good job!

PR #9731, opened 2015-10-23
Remove tesla boost from RA

### LINKED ISSUE #9647: Shock troopers powering tesla coils will "ignore" attack moves
(on bleed)
When a shock trooper is "attached" to a tesla coil you can't move him away with an attack move.
He **doesn't really ignore** the command but he defaults back to "attacking" the coil (probably because it's the nearest thing) which is a bit annoying.

Edit:
* will however prefer attacking nearby enemy units but only **if he's not already "attached"** to the coil (probably a wanted behavior)
* won't return fire when he's "stuck" on coil (debatable behavior)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27459339-shock-troopers-powering-tesla-coils-will-ignore-attack-moves?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #9727: Boosted tesla will break the balance
i dont even need a playtest to figure that one out 

they do get more range and damage when being boosted by shocktroopers , this is totally inacceptable beside the fact that it is a "RA2" feature ... i dont get the point why we should add every new feature to the "RA1"-mod ... as if our RA-Mod is the laboratory monkey here  for all the feature experiments

the Teslas are allready badass defense-structures , with the highest range and damage of "all" defense structures ... the ONLY thing you encourage by allies , is to spam even more artys ... 

why ? you ask ... cause allies have the weakest tanks ... medium-tanks and light-tanks are cannon fodder and teslas just have a "Grill-Fest" when allies tanks roll up ... thats why nobody is using them in a serious attack ... and thats why everyone spams artys instead 

now with the shocktroopers boosting teslas ... you dont need to be einstein to see that this is inappropriated

https://github.com/OpenRA/OpenRA/issues/2330
https://github.com/OpenRA/OpenRA/issues/4589
https://github.com/OpenRA/OpenRA/pull/8563

@Mailaender : please , we all know your "#this-is-not-true-to-the-original"-philosophy ... when theres controversy about a new feature .... but people play "Red Alert1" because it is called "Red Alert1" ... so lets "keep" it Red Alert1 atleast slightly

@pchote @obrakmann : 
we should consider to remove the "stat boost" and just have the Tesla "powered" when there is a PowerOutage ... with the help of shocktroopers ... i think that would be fair enough , cause Allies can cause a PowerOutage

@Unit158 @Phrohdoh : .... thx

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27637301-boosted-tesla-will-break-the-balance?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
Fixes #9647.
Fixes #9727.


atlimit8:
Note: The underlying bug of #9647 will still exist.

obrakmann:
@atlimit8: depends on how you look at it. AttackMove itself is working just as it should in this case, so it's not a bug. If you consider the behaviour a bug, then target priorities will fix it, for which there already is a separate ticket.

atlimit8:
:+1: Alright then.

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/2acc9fef3356ef81b493bd58eda60d2517ce5b85...25ef0d4f364aa966dea233e24d7335df8630009f)

pchote:
We'll be able to take this again in the future, after the targeting issues have been fixed and we have had a proper balance discussion (and when we aren't trying to push for a time constrained release).

Mailaender:
Added it to https://github.com/OpenRA/ra2/pull/83 where it belongs.

PR #9734, opened 2015-10-23
Improve AutoTarget performance

RoosterDragon:
I have attempted to improve the performance of several key areas used by the AutoTarget functionality, since this is used by many units. As a bonus, this reduces some code duplication too.

On the RA shellmap, total CPU time spent in AutoTarget.ScanForTarget drops from 7.1% to 3.4% with these changes.

### DIFF
--- a/OpenRA.Game/Actor.cs
+++ b/OpenRA.Game/Actor.cs
@@ -44,6 +44,7 @@ namespace OpenRA
 		public Rectangle VisualBounds { get; private set; }
 		public IEffectiveOwner EffectiveOwner { get; private set; }
 		public IOccupySpace OccupiesSpace { get; private set; }
+		public ITargetable[] Targetables { get; private set; }
 
 		public bool IsIdle { get { return currentActivity == null; } }
 		public bool IsDead { get { return Disposed || (health != null && health.IsDead); } }
@@ -107,6 +108,7 @@ namespace OpenRA
 			disables = TraitsImplementing<IDisable>().ToArray();
 			visibilityModifiers = TraitsImplementing<IVisibilityModifier>().ToArray();
 			defaultVisibility = Trait<IDefaultVisibility>();
+			Targetables = TraitsImplementing<ITargetable>().ToArray();
 		}
 
 		Rectangle DetermineBounds()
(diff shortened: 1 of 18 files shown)


reaperrr:
Needs rebase.

RoosterDragon:
Rebased.

abcdefg30:
For some reason `Tanya` is auto targeting explosive barrels now. (Testcase: `allies03a`)

RoosterDragon:
Pop quiz: What's invalid about this transformation?

Before
```c#
foreach (var actor in actorsInRange)
{
    if (actor.TraitsImplementing<IPreventsAutoTarget>().Any(pat =>
        pat.PreventsAutoTarget(actor, self))
        continue;

    // etc...
}
```

After
```c#
foreach (var actor in actorsInRange)
{
    foreach (var pat in actor.TraitsImplementing<IPreventsAutoTarget>())
        if (pat.PreventsAutoTarget(actor, self))
            continue;

    // etc...
}
```

Answer: The `continue` now applies to the inner loop, rather than the outer loop. This effectivly stops the prevents auto target from working. Oops! I extracted a method to fix this, thanks for finding it.

OpenRA.Game/Actor.cs
```diff
@@ -320,6 +322,30 @@ public bool CanBeViewedByPlayer(Player player)
 			return defaultVisibility.IsVisible(this, player);
 		}
 
+		public IEnumerable<string> GetAllTargetTypes()
```
penev92: I'm thinking these should go into some Actor exts, or (since targeting code is also in the engine, not in mod code) get their own region, because this class is getting pretty crowded.

RoosterDragon: Both seem reasonable to me to avoid overcluttering the `Actor` class. Any thoughts on where an extension method should be defined if I was going to do that?

RoosterDragon: On second thought, since I can't move it to ActorExts, I think I'll leave it for now.

I don't know if a region just for these will help. Maybe another PR can look into reorganising/refactoring the Actor class a bit.

RoosterDragon:
Added some perf comments, similar to #10208.

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -167,43 +164,57 @@ void Attack(Actor self, Actor targetActor, bool allowMove)
...
+						target.IsInRange(self.CenterPosition, arm.MaxRange()) &&
+						!target.IsInRange(self.CenterPosition, arm.Weapon.MinRange));
+
+				var armament = armaments.FirstOrDefault();
+				if (armament == null)
+					continue;
+
+				List<Actor> actors;
+				if (actorsByArmament.TryGetValue(armament, out actors))
+					actors.Add(actor);
```
abcdefg30: Somehow I don't get the logic here... Why are we adding stuff to a field that isn't used anymore after?
Should this be `actorsByArmament` instead of `actors`?

RoosterDragon: You mean `actors`? That's stored as a value in the `actorsByArmament` dictionary.

A bit further down we pull out one of the lists in this dictionary. So there's where it would get used again.

abcdefg30: Ah now I got it, thanks.

abcdefg30:

obrakmann:
Code looks good to me, didn't see anything obviously wrong in a couple of games, either. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7d67c6a64c87a24f15fb5ce9a07a5a50970f19e3)

PR #9738, opened 2015-10-23
Fix the legacy map importer not working

abcdefg30:
On current bleed, I can't import red alert maps anymore.
The bad commit seems to be 95709f11578747dfc9e1869f074b8d686e645336.

(Not sure if this is the correct fix tho.)


penev92:
Something tells me a number of other utility commands will need the same fix.

abcdefg30:
Which ones? I didn't find one while skimming throught the other commands.
E.g. `D2kMapImporter` uses `File.OpenRead(filename)` and not `GlobalFileSystem.Open(iniFile)`.

Mailaender:
I solved it with https://github.com/OpenRA/OpenRA/commit/1bff654754b66777729726c76f6c916716202366 for the ts map import.

penev92:
Well I can't think of other examples so OK :+1:

PR #9743, opened 2015-10-24
Fix RA-uibits dependency in D2k and TS

reaperrr:
Copies ui elements to D2k and TS instead of loading them from RA.


Mailaender:
:-1: on the duplication. Move it to `mods/common` instead.

reaperrr:
I'll wait with updating until #9744 is merged.

reaperrr:
Updated.

Moved the art to each mods' buttons.png instead and removed the separate files from RA. I know I could have packed them more densely, but I wanted to keep position consistent across mods.

Mailaender:
Hmmkay... I guess saving sprite sheets on runtime is more beneficial than avoiding duplicated artwork. Still works. :white_check_mark:

obrakmann:
Seems ok. Couldn't spot any misalignments in any mods during a quick check. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9905ff86d3c68fd30a89c07b62fc0e804da71f80)

PR #9744, opened 2015-10-24
Redesign the MP server browser.

### LINKED ISSUE #9451: Re-design server browser and lobby

pchote:
rtk6npk

Now displays up to 10 games at a time (up from three) and streamlines the flow for direct connect and creating servers.

Closes https://github.com/OpenRA/OpenRA/issues/9451.


obrakmann:
Looking good so far.

One style issue: in !cnc, the tab buttons have gaps between them, whereas in the lobby they don't.

And one suggestion: the smallprint in direct ip mode could be moved into the tab proper. There's a lot of empty space there, so it wouldn't look as cramped below the minimap, and not as bland in main content area. Also you could choose a bigger font size to improve readability for that text. As I said, just a suggestion, not a request. Do with it what thou wilt.

pchote:
Fixed the tabs, well spotted.  Not sure what to do about the description - I wrote that specifically to fill the gap between the map preview and the join button.  Could move it to the main bin, but we'd then need to come up with something to fill the new gap, and also the create server panel.

obrakmann:
There's one more problem with the spawnpoints display:
screenshot from 2015-10-24 21 48 35

Fixing that would require some surgery in `MapPreviewWidget`, which I think is out of scope for this PR. We could ignore it for now, since TS isn't shipped so the bug wouldn't be exposed, or we could remove the spawnpoints display until it is fixed. I'd probably go with the second option. Enough people are now checking out TS themselves for them to notice bugs.

Other than that, :+1:

pchote:
Yeah, I also noticed that (and commented about it in IRC) while testing.  That issue has always been there, but was previously hidden by not showing the spawn points in the map chooser.  Fixing that will require changes to both the widget and the master server (because we don't necessarily know what map grid style another mod uses), so can't be fixed here.

pchote:
Updated:
* Rebased.
* Fixed the width of the refresh button in !td.
* Fixed cross-mod spawn rendering (depends on https://github.com/OpenRA/OpenRA-Resources/issues/265).

Mailaender:
It suffers from the same layout problems mentioned in https://github.com/OpenRA/OpenRA/issues/9451#issuecomment-149784520 with a lot of empty space in the row of the back button as well as the direct IP and create server panels.

image

No doubt the multiplayer browser scales far better with the number of servers we have nowadays. Great job! This can be polished in later releases / pull requests.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/126b3c59e4285413e0dd29e3b2a8e9bc8fbe09c2...669cf43b6322eb58559b1d82097d38a7aba8d6c3)

Mailaender:
@Phrohdoh Keep these chrome.yaml changes in mind when rebasing your RA2 mod.

penev92:
Already done.

PR #9646, opened 2015-10-16
Temporary fix for missiles launched from zero height

matija-hustic:
Fixes choosing minimum launch speed for missiles launched at or below terrain height.

Takes care of #9645.

PR #9746, opened 2015-10-24
Revert missile explosion condition to strictly less than zero

matija-hustic:
Makes broken submarine torpedos work again.

https://github.com/OpenRA/OpenRA/pull/9646#issuecomment-150830239

### DIFF
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -788,7 +788,7 @@ namespace OpenRA.Mods.Common.Effects
 			// NOTE: High speeds might cause the missile to miss the target or fly through obstacles
 			//       In that case, big moves should probably be decomposed into multiple smaller ones with hit checks
 			var height = world.Map.DistanceAboveTerrain(pos);
-			var shouldExplode = (height.Length <= 0) // Hit the ground
+			var shouldExplode = (height.Length < 0) // Hit the ground
 				|| (relTarDist < info.CloseEnough.Length) // Within range
 				|| (info.ExplodeWhenEmpty && info.RangeLimit != 0 && ticks > info.RangeLimit) // Ran out of fuel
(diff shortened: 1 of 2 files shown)


abcdefg30:
Torpedos work again and the TD SAM works too. :+1: / :white_check_mark:

obrakmann:
Torpedoes are ok again. Thanks for the quick fix!

Unfortunately I just noticed another issue: the advanced guard tower in TD has a problem with its missiles (on bleed, too, not due to this change). Could you take a look at that, too, please?

matija-hustic:
@obrakmann the advanced guard tower gives the missile [initial position of -85](https://github.com/OpenRA/OpenRA/blob/bleed/mods/cnc/rules/structures.yaml#L821)

Up until now it has been these boundary cases of exactly zero launch height, but what do I do here? The launch height is -85, one tick later, the missile climbs up to -36 and then it explodes. Is a YAML change justified here? Or do I still need to handle this somehow in the code?

pchote:
> The launch height is -85

This sounds bogus to me.

matija-hustic:
I've just checked with the old missile code and there it didn't blow up because the missile was able to climb to zero height within one tick (there was only the maximum pitch and no VROT to constrain it).

obrakmann:
I just played with the offsets some, and 75 for the Z seems to give results similar to last release. 0 doesn't help at all, and lower values (I tried 10 for example) only work on longer distances.

matija-hustic:
@obrakmann zero works for me...

reaperrr:
> This sounds bogus to me.

@pchote Probably a side-effect of using a 1x2 foundation with `_ x` instead of 1x1 foundation with art position fixed via sequence offsets.

obrakmann:
@matija-hustic: right, sorry. I was testing on bleed. It's fine now. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/2206b3597f77fa5b9dff37f26989069f20427eb9)

Thanks again for the quick fix!

Issue #9420, opened 2015-09-24
Crash when using map search if a map doesn't have an author.


Raymond-cnc:
The crash happens immediately as you type into the search box. http://resource.openra.net/maps/5923 This map was causing the issue for me as it doesn't have an author.

Here is exception: http://pastebin.com/NDPWHxMw

Occured in latest September 2015 release.

PR #9748, opened 2015-10-24
Add a lint pass for map format/author/title.

pchote:
Inspired by #9420 and others.  It doesn't make any sense for a map to not define a title or author, so this will flag the maps when they are uploaded to the resource center.

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -115,6 +115,8 @@ namespace OpenRA
 
 	public class Map
 	{
+		public const int MinimumSupportedMapFormat = 6;
+
 		static readonly int[][] CellCornerHalfHeights = new int[][]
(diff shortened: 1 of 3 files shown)


obrakmann:
I think it would be worth checking Map.Type, too, just to see if it's set. If it's unset, it would never appear in the map chooser, possibly leading to confusion on the part of the mapper (not to mention players if it was uploaded to the resource site).

pchote:
Good idea.  Updated.

abcdefg30:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/75a6c5dd902b19b0ecebfd03aa38bdcd03d43f3d)

Issue #9508, opened 2015-10-02
Map Editor: Screenshot hotkey doesn't work


clemty:
there's no way to make a screenshot of the Lobby either
##

Mailaender:
I submitted a "fix" for it https://github.com/OpenRA/OpenRA/issues/9755.

rob-v:
Fixed by #13186

PR #9755, opened 2015-10-25
Added a minimal debug text overlay for UI without chat

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/9508.

Background: actually there is a way to take a screenshot. The hotkey `CTRL` + `P` works, but there was no debug text giving feedback. This adds the missing debug text to the main menu and map editor.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
@@ -37,6 +37,9 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			rootMenu = widget;
 			rootMenu.Get<LabelWidget>("VERSION_LABEL").Text = Game.ModData.Manifest.Mod.Version;
 
+			// TODO: This will disapear when navigating into sub menus.
+			Game.LoadWidget(world, "CHAT_PANEL", rootMenu, new WidgetArgs());
+
 			// Menu buttons
 			var mainMenu = widget.Get("MAIN_MENU");
 			mainMenu.IsVisible = () => menuType == MenuType.Main;
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameChatLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameChatLogic.cs
@@ -109,6 +109,9 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				if (e.Event == KeyInputEvent.Up)
 					return false;
 
+				if (world.Type != WorldType.Regular)
+					return false;
+
 				if (!chatChrome.IsVisible() && (e.Key == Keycode.RETURN || e.Key == Keycode.KP_ENTER))
 				{
 					OpenChat();
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/LoadMapEditorLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/LoadMapEditorLogic.cs
@@ -19,6 +19,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 		{
 			var editorRoot = widget.Get("WORLD_ROOT");
 			Game.LoadWidget(world, "EDITOR_WORLD_ROOT", editorRoot, new WidgetArgs());
+			Game.LoadWidget(world, "CHAT_PANEL", editorRoot, new WidgetArgs());
 		}
 	}
 }


OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameDebugLogic.cs
```diff
@@ -0,0 +1,62 @@
...
+		readonly ChatDisplayWidget debugOverlayDisplay;
+
+		[ObjectCreator.UseCtor]
+		public IngameDebugLogic(Widget widget, OrderManager orderManager, World world, Ruleset modRules)
+		{
+			this.orderManager = orderManager;
+			this.modRules = modRules;
+
+			var chatPanel = (ContainerWidget)widget;
+			debugOverlay = chatPanel.Get<ContainerWidget>("DEBUG_OVERLAY");
```
Phrohdoh: Are these widgets necessary, or will this force them on mods?

abcdefg30:
Thanks. :+1: / :white_check_mark:

obrakmann:
It seems to stop working when you've opened some of the other widgets (lobby, replay browser, ...). The screenshots still get saved, but the notification is missing.

Mailaender:
*ugh* I don't really have an idea why this happens nor how to fix this. Only promised to get it to work in the map editor. :P

abcdefg30:
It looks to me like this is caused by creating a new world when opening the lobby.

Mailaender:
Fixed.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameDebugLogic.cs
```diff
@@ -0,0 +1,62 @@
...
+			orderManager.AddChatLine -= AddDebugLineWrapper;
+			Game.BeforeGameStart -= UnregisterEvents;
+		}
+
+		public void AddDebugLineWrapper(Color c, string from, string text)
+		{
+			AddDebugLine(c, from, text, false);
+		}
+
+		void AddDebugLine(Color c, string from, string text, bool replayCache)
```
RoosterDragon: You can simplify this method by removing `replayCache`. You can then hook it to the event handler directly, and remove the loops that fills in the history from `orderManager.ChatCache` since this is all ignored anyway.

Mailaender: Done. Thanks.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameDebugLogic.cs
```diff
@@ -0,0 +1,54 @@
...
+		{
+			this.orderManager = orderManager;
+			this.modRules = modRules;
+
+			var chatPanel = (ContainerWidget)widget;
+			debugOverlay = chatPanel.Get<ContainerWidget>("DEBUG_OVERLAY");
+			debugOverlayDisplay = debugOverlay.Get<ChatDisplayWidget>("DEBUG_DISPLAY");
+
+			foreach (var chatLine in orderManager.ChatCache)
+				AddDebugLine(chatLine.Color, chatLine.Name, chatLine.Text);
```
RoosterDragon: I'm not sure about this. In `IngameChatLogic` when it runs through the cache it *doesn't* add lines or notifications for anything from the cache.

RoosterDragon:
If I open the shellmap and take a few screenshots (so there's some text) and then open the skirmish menu to trigger the creation of the new world (you can tell when the happens as the tick count gets confused and stuck at 0) - the text will never go away. I assume the old widget is stuck on screen.

OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
```diff
@@ -341,12 +347,17 @@ void RemoveShellmapUI()
...
+			Game.LoadWidget(world, "DEBUG_TEXT_PANEL", rootMenu, new WidgetArgs());
+		}
+
 		void OpenSkirmishLobbyPanel()
 		{
 			menuType = MenuType.None;
 			Game.OpenWindow("SERVER_LOBBY", new WidgetArgs
 			{
-				{ "onExit", () => { Game.Disconnect(); menuType = MenuType.Singleplayer; } },
+				{ "onExit", () => { Game.Disconnect(); menuType = MenuType.Singleplayer; AddShellmapUI(); } },
```
pchote: This will add new widgets each time, never removing the old ones.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameDebugLogic.cs
pchote: Why can't this use Game.AddChatLine?

OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
Mailaender: Should be fixed now.

RoosterDragon:
I can still recreate my issue, if you open the skirmish menu any notifications still on screen will be stuck there. I assume the creation of the new world is messing with the removal.

Mailaender:
I simplified this by re-using the regular in-game chat chrome instead of copy/pasting parts. The entry window is now simply disabled when you shouldn't chat or use developer commands. This will fix it for the shellmap entry screen, but not sub-menus and the map editor, which was the original request.

RoosterDragon:
This new approach still suffers from the exact same issue where the widget won't get cleaned up if you do something that triggers a new world being loaded (opening the skirmish or multiplayer menu).

Mailaender:
This is leading no where I guess then.

Issue #7520, opened 2015-02-22
Z-Buffer support [$15]


Mailaender:
~~Corpses being rendered above vehicles is looking glitchy and bugging me for quite some time. Somehow not easy to fix. The ZOffset is ignored. I tried adding it to sequences and hard-coded into the corpse effect.~~

Already fixed by https://github.com/OpenRA/OpenRA/issues/10119.

pchote:
This applies to all Effects, which aren't z-sorted for performance reasons.  This will eventually be fixed when we redo the renderer with proper z-buffer support (which is a prerequisite for proper TS support).

pchote:
We merged #7968 in the last release, and I have just filed #9476.  The next step is to fix our treatment of the vertex z coordinates, and it then becomes easy to construct a depth image for the terrain:

![screen shot 2015-09-28 at 21 15 02](https://cloud.githubusercontent.com/assets/167819/10147824/3678beae-6628-11e5-971d-e44d3c132361.png)

pchote:
Once we have the depth image, it then becomes straightforward (but not trivial) to enable proper depth filtering:

![screen shot 2015-09-28 at 22 09 27](https://cloud.githubusercontent.com/assets/167819/10148831/b9c62c74-662d-11e5-8879-273692a62da6.png)

Mailaender:
Looks impressive! Is the missing door overlay a regression due to wrong Z sorting?

pchote:
It is a regression due to wrong z sorting because of the quick and dirty way it was implemented.  The proper implementation (which depends on some refactoring that hasn't happened yet) won't have that problem.

pchote:
I now have a big chunk of this feature completed and ready for review/merge:

* #11058 puts in place the initial groundwork and adds depth buffer support for terrain.
* [shroud-vertex-depth](https://github.com/pchote/OpenRA/tree/shroud-vertex-depth) adds depth buffer support for regular sprites and fixes shroud rendering.
* [tileset-depth](https://github.com/pchote/OpenRA/tree/tileset-depth) makes some depth metadata configurable and sets up the correct offsets for TS.

[depthbuffer-part-two](https://github.com/pchote/OpenRA/tree/depthbuffer-part-two) is a bucket of prototype changes that will grow into more PRs after the earlier ones are merged.

PR #9756, opened 2015-10-25
Made Dune 2000 corpses vanish faster

Mailaender:
This hides the glitch from https://github.com/OpenRA/OpenRA/issues/7520 a bit better by playing the animation faster and keeping the blood splattering on top of the vehicle for a shorter amount of time.

### DIFF
--- a/mods/d2k/sequences/infantry.yaml
+++ b/mods/d2k/sequences/infantry.yaml
@@ -53,7 +53,7 @@ light_inf:
 	die-crushed: DATA.R8
 		Frames: 386, 393, 400, 407, 414, 421, 428, 435
 		Length: 8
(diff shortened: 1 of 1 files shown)


pchote:
I'm not convinced about this.  This makes the crushed corpses 20 times faster than RA and TD, which both use values of 1600.

obrakmann:
#10119 fixes this in a better way. :-1:

abcdefg30:
> 17:15:27 (penev) abcdefg30: what are you reviewing
17:15:38 (abcdefg30) #9756 kinda xD
17:15:43 (abcdefg30) was just going to later Mailaender
17:16:27 (pchote|alt) I think we can close that one now

PR #9759, opened 2015-10-25
Add ActorsHavingTrait<TTrait>()

atlimit8:
Gets actors (distinct) that have trait `TTrait` - won't repeat actor if it has multiple occurrences of `TTrait`.
Should be equivalent to `ActorsWithTrait<TTrait>().Select(tp => tp.Actor).Distinct()` except probably faster with less allocation.

### DIFF
--- a/OpenRA.Game/Graphics/WorldRenderer.cs
+++ b/OpenRA.Game/Graphics/WorldRenderer.cs
@@ -185,8 +185,7 @@ namespace OpenRA.Graphics
 
 			if (World.Type == WorldType.Regular && Game.Settings.Game.AlwaysShowStatusBars)
 			{
-				foreach (var g in World.Actors.Where(a => !a.Disposed
-					&& a.Info.HasTraitInfo<SelectableInfo>()
+				foreach (var g in World.ActorsHavingTrait<Selectable>().Where(a => !a.Disposed
 					&& !World.FogObscures(a)
 					&& !World.Selection.Actors.Contains(a)))
 
--- a/OpenRA.Game/TraitDictionary.cs
+++ b/OpenRA.Game/TraitDictionary.cs
@@ -107,6 +107,16 @@ namespace OpenRA
 			return InnerGet<T>().All();
 		}
 
+		public IEnumerable<Actor> ActorsHavingTrait<T>()
+		{
+			return InnerGet<T>().Actors();
+		}
+
+		public IEnumerable<Actor> ActorsHavingTrait<T>(Func<T, bool> predicate)
+		{
+			return InnerGet<T>().Actors(predicate);
+		}
+
 		public void RemoveActor(Actor a)
 		{
 			foreach (var t in traits)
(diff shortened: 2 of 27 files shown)


pchote:
> probably faster with less allocation.

Can we get some numbers showing these?  Without evidence showing an improvement, this seems like more plumbing for the sake of it.

atlimit8:
Actually, the distinct part is more important since actors can have multiple occurrences of traits; currently, code using `ActorsWithTrait<TTrait>().Select(tp => tp.Actor)` assumes that `TTrait` will only occur once per actor.

RoosterDragon:
I think this is worth having on its own merits - and it will certainly be faster than grovelling through every actor in the world and checking the traits of all of them.

Phrohdoh:
On the topic of English, `ActorsHavingTrait` doesn't sound proper nor does it lend itself to being understandably different from `ActorsWithTrait`.

atlimit8:
What about `ActorsThatHaveTrait<TTrait>`, it is a little more verbose but is a valid English noun phrase converted to PascalCase C#.

atlimit8:
@RoosterDragon perhaps you can measure it if you have the time.

RoosterDragon:
You might consider renaming `ActorsWithTrait<T>` - since that is really more like `AllTraitsOfType<T>`.

atlimit8:
It's more like `AllTraitPairs<TTrait>`.

RoosterDragon:
I'm fine with whatever names, if it fits in overall and doesn't create too much confusion.

RoosterDragon:
:+1: The changes are useful and I'm happy with the current naming scheme - but if anybody has a better suggestion for names I'm fine with changing it.

atlimit8:
Profiling using `time`
```
Map:          Costal Influence
Bots:         8 (2 of each)
Command line: time mono OpenRA.Game.exe Launch.Replay=$HOME/.openra/Replays/ra/\{DEV_VERSION\}/OpenRA-2015-10-28T231319Z.orarep

            bleed:	mono OpenRA.Game.exe   2206.08s user 20.00s system 99% cpu 37:19.95 total
ActorsHavingTrait:	mono OpenRA.Game.exe   2084.19s user 12.70s system 99% cpu 34:59.48 total

user/application:	0.944748150566 => 5.52518494343% reduction in application code execution
system/kernal:   	0.635 => 36.5% reduction from less allocation
total:           	(2084.19s + 12.70s = 2096.89s) / (2206.08s+20.00s = 2226.08s) =
					= 0.941965248329 => 5.80347516711% total reduction
```
So about a 5-6% reduction in total AI execution and allocation cost.

obrakmann:
Looks ok to me. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/a360fa7101fa80fcb9108d815724d2529481bc96). I lumped this into the "general performance improvements" entry, hope that's ok.

atlimit8:
Yup.

PR #9760, opened 2015-10-25
Fix harvesters teleporting through doors

pchote:
The `Harvester` trait has an `INotifyBlockingMove` implementation that cancels the activity queue if the current activity is `Wait`.  This interacts badly with the weapons factories in TD and TS, which make the units wait while the door opens.  On current bleed cancelling the wait and queued MoveIntoWorld makes the units instantly teleport to the exit cell, before continuing on.  This isn't visible in RA or D2K because they don't define production waits.

This works around the issue by making the production wait uncancellable.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Production.cs
+++ b/OpenRA.Mods.Common/Traits/Production.cs
@@ -87,7 +87,7 @@ namespace OpenRA.Mods.Common.Traits
 					if (exitinfo.MoveIntoWorld)
 					{
 						if (exitinfo.ExitDelay > 0)
-							newUnit.QueueActivity(new Wait(exitinfo.ExitDelay));
+							newUnit.QueueActivity(new Wait(exitinfo.ExitDelay, false));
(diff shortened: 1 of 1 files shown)


Mailaender:
I don't see a difference. Does the bug only occur when the factory exit is blocked?

pchote:
It happens if another actor nudges the exit cell that an actor is going to move in to.  There are several ways that this could happen, but the easiest for the sake of testing is to enable the instant build cheat and build multiple harvesters.

obrakmann:
:+1:

atlimit8:
:+1: Simple and looks like it works.

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f9fb5707971c138c09310cbd9314255a97bc88bc...9de9115f976ea6a64baeb94602ddde442fd785e7)

PR #9763, opened 2015-10-25
Attempt to reduce SpreadDamageWarhead CPU overhead

reaperrr:
I'm not sure how large the total positive effect on performance is (assuming there is any), but I think this is worth testing.

This basically does ~two~ three things:
- removes the 2% and 1% stages from default `Falloff`. For the majority of weapons in RA, TD and TS these stages resulted in zero damage or 1 damage point against most actors anyway, while D2k now gets its own, custom Falloffs. This reduces situations where infantry would go prone due to a single damage point, and effectively reduces the default radius that is scanned for victims from 6x to 4x `Spread`, reducing the average number of iterated actors and number of 1-damage-point `Health.InflictDamage` ops.
- `hitActors` now only enumerates actors with `HealthInfo`, which avoids iterating over and performing a pointless `DoImpact` on invulnerable actors without `Health` trait (which might actually yield significant savings when weapons with large area of effect - like nukes - are fired into an area with lots of rocks or invulnerable trees etc., but for now that's just a guess).
- reduces default `TargetExtraSearchRadius` to largest feasible health radius on any shipping mod (1c512). A larger scan radius will only be needed when we increase health radii / health boxes, and only on weapons with relatively low area of effect.

### DIFF
--- a/OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs
@@ -22,7 +22,7 @@ namespace OpenRA.Mods.Common.Warheads
 		public readonly WDist Spread = new WDist(43);
 
 		[Desc("Extra search radius beyond maximum spread. Required to ensure damage to actors with large health radius.")]
-		public readonly WDist TargetExtraSearchRadius = new WDist(2048);
+		public readonly WDist TargetExtraSearchRadius = new WDist(1536);
 
 		[Desc("Damage percentage at each range step")]
 		public readonly int[] Falloff = { 100, 37, 14, 5, 0 };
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs
```diff
@@ -58,7 +59,7 @@ public override void DoImpact(WPos pos, Actor firedBy, IEnumerable<int> damageMo
 
 			// This only finds actors where the center is within the search radius,
 			// so we need to search beyond the maximum spread to account for actors with large health radius
-			var hitActors = world.FindActorsInCircle(pos, Range[Range.Length - 1] + TargetExtraSearchRadius);
+			var hitActors = world.FindActorsInCircle(pos, Range[Range.Length - 1] + TargetExtraSearchRadius).Where(a => a.Info.HasTraitInfo<HealthInfo>());
```
pchote: This introduces a second trait(info) lookup per actor.  It would be better to keep the one query in the loop, but change the conditional to `if (healthInfo == null) continue;`

reaperrr:
Updated.

RoosterDragon:
:+1:

GraionDilach:
:+1:

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/de6dd63b6782c97e4cf5ebc85e75b623af7cec8c...62a3311c2ea10ccab42bc49d9099c0b1ee9a1674)

Issue #9737, opened 2015-10-23
Sidebar button keyboard shortcuts are triggered by repeating rate input rather than physical key press.


RichSuther:
Holding down keyboard keys assigned to the sidebar buttons such as Main Menu(Esc), Place Beacon(B), Sell(Z), Repair(C). Build queue short cuts and builds themselves (Function keys). Are trigged by keyboard auto repeat rate rather than physical press.

These are triggered by the OS Keyboard Repeat Rate settings and if User has these set high can effect the game. As described in issue #7941

atlimit8:
Would you prefer that they be triggered on key down or up?

RichSuther:
Wasn't sure so I checked in the original Command and Conquer(95), looks like they went for the on key down approach, so cant argue with that. :)

PunkPun:
Is this still an issue?

PR #9765, opened 2015-10-25
Added filter to only register first key press

RichSuther:
Fix for #9737. Currently holding keyboard short cut keys triggers every key repeat, this changes to only register on physical key press.


OpenRA.Platforms.Default/Sdl2Input.cs
```diff
@@ -161,7 +161,10 @@ public void PumpInput(IInputHandler inputHandler)
 								Platform.CurrentPlatform == PlatformType.Windows)
 								Game.Exit();
 							else
-								inputHandler.OnKeyInput(keyEvent);
+
```
Mailaender: You are missing `{` here or the else should go away.

Phrohdoh:
You'll need to rebase to remove a2e8d16.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/9791.

PR #9767, opened 2015-10-25
Change aircraft's maximum pitch in TD to match the increased cruise altitude

obrakmann:
Without this change, the C17 would not manage to get all the way down to the airfield's level, leading to a sudden drop in height, seemingly teleporting down to ground level.


RoosterDragon:
:+1:

penev92:
This looks very unnatural, but it is a much-needed fix so :+1:

pchote:
Production airdrop has a hardcoded 9-cell land range.  An alternative fix would be to unhardcode and increase that so that a shallower rate works.

PR #9771, opened 2015-10-26
Implement D2k thumper

reaperrr:
Thumper replaces Medic.

`AttractsWorms` now only attracts them, while actors with `WormFodder` also get eaten.

### DIFF
--- a/OpenRA.Mods.D2k/Traits/Sandworm.cs
+++ b/OpenRA.Mods.D2k/Traits/Sandworm.cs
@@ -99,7 +99,8 @@ namespace OpenRA.Mods.D2k.Traits
 			targetCountdown = Info.TargetRescanInterval;
 
 			// If close enough, we don't care about other actors.
-			var target = self.World.FindActorsInCircle(self.CenterPosition, Info.IgnoreNoiseAttackRange).FirstOrDefault(x => x.Info.HasTraitInfo<AttractsWormsInfo>());
+			var target = self.World.FindActorsInCircle(self.CenterPosition, Info.IgnoreNoiseAttackRange)
+				.FirstOrDefault(x => attackTrait.HasAnyValidWeapons(Target.FromActor(x)));
 			if (target != null)
 			{
 				self.CancelActivity();
--- a/mods/d2k/rules/infantry.yaml
+++ b/mods/d2k/rules/infantry.yaml
@@ -74,34 +74,48 @@ trooper:
 	WithInfantryBody:
 		AttackSequence: shoot
 
-medic:
+thumper:
 	Inherits: ^Infantry
+	-AutoTarget:
 	Buildable:
 		Queue: Infantry
 		BuildPaletteOrder: 60
-		Prerequisites: ~barracks.medics, upgrade.barracks, ~techlevel.high
+		Prerequisites: upgrade.barracks, ~techlevel.high
 	Valued:
 		Cost: 200
 	CustomBuildTimeValue:
 		Value: 108
 	Tooltip:
-		Name: Medic
-		Description: Heals nearby infantry\n  Strong vs Nothing\n  Weak vs Everything
+		Name: Thumper
+		Description: Attracts nearby worms\n  Unarmed
 	Health:
 		HP: 375
 	RevealsShroud:
 		Range: 2c768
 	Mobile:
-		Speed: 42
-	Armament:
-		Weapon: Heal
-		Cursor: ability
-		OutsideRangeCursor: ability
-		TargetStances: Ally
-		ForceTargetStances: None
-	AttackFrontal:
+		Speed: 43
+	DeployToUpgrade:
+		Upgrades: deployed
+		Facing: 128
+		AllowedTerrainTypes: Sand, Spice, Dune
 	WithInfantryBody:
-		AttackSequence: heal
+		UpgradeTypes: deployed
+		UpgradeMaxEnabledLevel: 0
+	WithSpriteBody@DEPLOYED:
+		Sequence: thump
+		UpgradeTypes: deployed
+		UpgradeMinEnabledLevel: 1
+	WithIdleOverlay@DEPLOYED:
+		Sequence: thump-sand
+		UpgradeTypes: deployed
+		UpgradeMinEnabledLevel: 1
+	AttractsWorms:
+		Intensity: 1000
+		UpgradeTypes: deployed
+		UpgradeMinEnabledLevel: 1
+	DisableOnUpgrade:
+		UpgradeTypes: deployed
+		UpgradeMinEnabledLevel: 1
 	Passenger:
 		PipType: Blue
 	Voiced:
--- a/mods/d2k/rules/infantry.yaml
+++ b/mods/d2k/rules/infantry.yaml
@@ -109,6 +109,11 @@ thumper:
 		Sequence: thump-sand
 		UpgradeTypes: deployed
 		UpgradeMinEnabledLevel: 1
+	AmbientSound:
+		SoundFile: THUMPER1.WAV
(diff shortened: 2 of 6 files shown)


mods/d2k/rules/infantry.yaml
```diff
@@ -70,32 +70,46 @@ trooper:
...
+		Prerequisites: upgrade.barracks, ~techlevel.high
 	Valued:
 		Cost: 200
 	CustomBuildTimeValue:
 		Value: 108
 	Tooltip:
-		Name: Medic
-		Description: Heals nearby infantry\n  Strong vs Nothing\n  Weak vs Everything
+		Name: Thumper
+		Description: Attracts nearby worms\n  Strong vs Nothing\n  Weak vs Everything
```
abcdefg30: `Description: Attracts nearby worms\n  Unarmed`

mods/d2k/rules/infantry.yaml
```diff
@@ -70,32 +70,46 @@ trooper:
...
-	WithInfantryBody:
-		AttackSequence: heal
+		Speed: 43
+	DeployToUpgrade:
+		Upgrades: deployed
+#		DeployAnimation:
+		Facing: 128
+		AllowedTerrainTypes: Sand, Spice, Dune
+		DeploySound:
+		UndeploySound:
```
abcdefg30: Both sounds already default to `null`?

mods/d2k/rules/infantry.yaml
```diff
@@ -70,32 +70,45 @@ trooper:
...
-		Cursor: ability
-		OutsideRangeCursor: ability
-		TargetStances: Ally
-		ForceTargetStances: None
-	AttackFrontal:
-	WithInfantryBody:
-		AttackSequence: heal
+		Speed: 43
+	DeployToUpgrade:
+		Upgrades: deployed
```
pchote: We'll also want an `AmbientSound` definition for the thump audio.

reaperrr: Done.

GraionDilach:
I have a feeling that ambientsounds with Interval=0 cannot be controlled via disabling the trait.

OpenRA.Mods.D2k/Traits/AttractsWorms.cs
```diff
@@ -20,10 +20,10 @@ public class AttractsWormsInfo : UpgradableTraitInfo
 		public readonly int Intensity = 0;
 
 		[Desc("Noise percentage at Range step away from the actor.")]
-		public readonly int[] Falloff = { 100, 100, 25, 11, 6, 4, 3, 2, 1, 0 };
+		public readonly int[] Falloff = { 100, 100, 25, 12, 6, 3, 2, 1, 0 };
```
Phrohdoh: At this point is it safe to change default values (which mods may rely on), instead of changing our yaml?

reaperrr:
Updated.

MicroJOo:
works as expected :) ... :+1:

problem though is that you can trap him "forever" ... basicly take him out of a game
a solution could be , that he can get used only for a limited time ... and needs to recharge then
thats just an idea , of course that would be not true to the original ... but i expect the current implementation of the thumper to cause controversy in the future ... sooner or later 

however , this is really close to the original , i like it

OpenRA.Mods.D2k/Traits/Sandworm.cs
```diff
@@ -99,7 +99,7 @@ void RescanForTargets(Actor self)
 			targetCountdown = Info.TargetRescanInterval;
 
 			// If close enough, we don't care about other actors.
-			var target = self.World.FindActorsInCircle(self.CenterPosition, Info.IgnoreNoiseAttackRange).FirstOrDefault(x => x.Info.HasTraitInfo<AttractsWormsInfo>());
+			var target = self.World.FindActorsInCircle(self.CenterPosition, Info.IgnoreNoiseAttackRange).FirstOrDefault(x => x.Info.HasTraitInfo<WormFodderInfo>());
```
Mailaender: @pchote suggested to use target types instead. This would become `.FirstOrDefault(x => attackTrait.HasAnyValidWeapons(Target.FromActor(x)));` then.

mods/d2k/sequences/infantry.yaml
```diff
@@ -196,10 +196,15 @@ medic: # actually thumper
 		Facings: -8
 		Transpose: true
 		Tick: 120
-	heal: DATA.R8
+	thump: DATA.R8
 		Start: 1458
 		Length: 5
 		Tick: 480
```
Mailaender: Something isn't right here. I think you are missing `Transpose: true` because on your branch the deployed thumper is still holding his (other?) thumping device:

image

However I believe you want it to look like

image

compare https://github.com/OpenRA/OpenRA/pull/9019/.

penev92: This screenshot suggests that probably the idle animation isn't removed when deploying.

reaperrr: Ah, my bad. The `@DEFAULT` added to `WithInfantryBody` meant it didn't override the `WithInfantryBody` inherited from the infantry default.

reaperrr:
Rebased and updated.

obrakmann:
Looks and works fine. Just two comments:

1. I think there's a typo in the second to last commit's commit message: "Align thumper _sand_..."

2. To address @GraionDilach's comment regarding controlling looped AmbientSounds via upgrades, I suggest the following code. I think that should fix it:

```csharp
class AmbientSound : UpgradableTrait<AmbientSoundInfo>, ITick
{
	ISound currentSound;
	bool wasDisabled = true;
	int interval;

	public AmbientSound(Actor self, AmbientSoundInfo info)
		: base(info)
	{
		interval = info.Interval;
	}

	public void Tick(Actor self)
	{
		if (IsTraitDisabled)
		{
			Game.Sound.StopSound(currentSound);
			currentSound = null;
			wasDisabled = true;
			return;
		}

		if (wasDisabled && Info.Interval == 0)
		{
			if (self.OccupiesSpace != null)
				currentSound = Game.Sound.PlayLooped(Info.SoundFile, self.CenterPosition);
			else
				currentSound = Game.Sound.PlayLooped(Info.SoundFile);
		}

		wasDisabled = false;

		if (Info.Interval == 0)
			return;

		if (interval-- > 0)
			return;

		interval = Info.Interval;

		if (self.OccupiesSpace != null)
			Game.Sound.Play(Info.SoundFile, self.CenterPosition);
		else
			Game.Sound.Play(Info.SoundFile);
	}
}
```

reaperrr:
@obrakmann 
> I think there's a typo in the second to last commit's commit message: "Align thumper *sand*..."

Unless you mean the missing `-` I'm not sure what you mean, there is a `thumper-sand` overlay which I tried to better align with the `thump` animation by reordering the frames. But I agree the commit message is a bit unprecise, I'll reword it a little.

obrakmann:
Oh nevermind. I probably should have looked at the actual commit. I assumed it was supposed to say "sound" instead of "sand". It's fine then.

reaperrr:
Updated.

Only minor change compared to @obrakmann's code: I changed `Info.Interval == 0` to `Info.Interval <= 0` so `-1` or some bogus higher negative value will have the same effect.

obrakmann:
:+1:

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/6a6de7f1f8c81a030cdb871e189caa984c8b7bf5...777757b055ea8c3f9dfb74a0d606ac1653fc82ea)

PR #9772, opened 2015-10-26
Fixed filtersButton = null return value from GetOrNull<DropDownButtonWidget>

Mailaender:
Another inconsistency from https://github.com/OpenRA/OpenRA/pull/9744. Not really a regression, but Coverity detects it, so I made this non-optional for a clearer crash if the chrome.yaml isn't adapted.


penev92:
:+1:

obrakmann:
:+1: doesn't make much sense to just make this one button optional.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f9fb5707971c138c09310cbd9314255a97bc88bc)

PR #9549, opened 2015-10-05
Add a music player to the game lobby.

pchote:
The first big part of #9451.  Fixes #9153. Depends on #9546.

PR #9773, opened 2015-10-26
Fixed musicTab = null return value from GetOrNull<ButtonWidget>

Mailaender:
Same as https://github.com/OpenRA/OpenRA/pull/9772 essentially to silence Coverity. This tab/button isn't really optional at all just like the others. Introduced by https://github.com/OpenRA/OpenRA/pull/9549.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
@@ -296,7 +296,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			playersTab.IsDisabled = () => panel == PanelType.Kick || panel == PanelType.ForceStart;
 			playersTab.OnClick = () => panel = PanelType.Players;
 
-			var musicTab = lobby.GetOrNull<ButtonWidget>("MUSIC_TAB");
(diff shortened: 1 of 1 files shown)


penev92:
:+1:

Phrohdoh:
Or do a proper null check instead of forcing mods to have a music tab.

obrakmann:
I think I agree with adding a proper null check. Since we only support the original soundtracks and not .mp3 or something, mods for which there isn't any soundtrack would have an unused button in their UI.

atlimit8:
>do a proper null check instead of forcing mods to have a music tab.
Agreed.

pchote:
Personally, i'm not convinced.  If a mod wants to remove all references to the music player, then there are at least two more `.Get<>()`'s that will need to be changed (the main and ingame menus).  This feels like a scope creep to me.

Phrohdoh:
@pchote We're not insisting on scope creep here, we're saying it'd be better to open a new PR (if this one refuses to change) that does away with requiring a music player.

penev92:
FWIW there are at least a dozen other cases like this, so it makes very little sense to bikeshed here.

Phrohdoh:
Asking for a different approach that doesn't force UI *is not* bikeshedding.

Mailaender:
Making parts of the lobby optional isn't a bad idea. Feel free to send a patch yourself.

PR #9775, opened 2015-10-27
Use verb-noun form of commands in make.ps1 instead of aliases

phrohdoh:


ScottNZ:
Why?

Phrohdoh:
No real reason other than following the Powershell best-practice of being verbose for others that may read/rewrite your code.

This is very simple and doesn't change behaviour, so if it isn't desired I'm not willing to argue for it.

ScottNZ:
Fair enough. I'm pretty indifferent to this.

RoosterDragon:
Needs a rebase.

Phrohdoh:
Rebased.

make.ps1
```diff
@@ -184,7 +184,7 @@ elseif ($command -eq "check-scripts")
 {
 	if ((Get-Command "luac.exe" -ErrorAction SilentlyContinue) -ne $null)
 	{
-		echo "Testing Lua scripts..."
+		Write-Output "Testing Lua scripts..."
 		foreach ($script in ls "mods/*/maps/*/*.lua")
```
atlimit8: Shouldn't `ls` be `Get-ChildItem` then?

Phrohdoh: It could be, yes.

penev92:
:+0.5: until I can test this, with @atlimit8's comment.

Issue #9035, opened 2015-08-11
Repulsion prevents helicopters from landing on helipad


reaperrr:
When you have more than 1 helicopter selected and order them to enter a helipad, the heli who reserves the pad more often than not won't be able to land because repulsion from the surrounding helicopters prevents that.

MicroJOo:
confirmed

GraionDilach:
Was that confirmation in latest bleed? The aircraft-terrain-height PR resulted with Starcraft I-like stacking of helis - not just like RA2, where they can be stacked during moving but they also stack after stopped/attacked.

MicroJOo:
i only know it from our current release ...

penev92:
Yes, having them land on bleed is hard-at-best. Also the aircraft repulsion should prevent any stacking.

pchote:
Fixed by #11393.

PR #9786, opened 2015-10-27
Reserve fixes

### LINKED ISSUE #9783: Attempted to finalize an undisposed DisposableAction. heli (355) reserved fix (77)
http://pastebin.com/zPePbUm8

git-549f8a9 RA

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27746265-attempted-to-finalize-an-undisposed-disposableaction-heli-355-reserved-fix-77?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RoosterDragon:
Probably fixes #9783 and strengthens the API slightly to help prevent this problem in future.

I have called out all potential behaviour changes inline for the attention of reviewers.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Air/TakeOff.cs
+++ b/OpenRA.Mods.Common/Activities/Air/TakeOff.cs
@@ -30,12 +30,7 @@ namespace OpenRA.Mods.Common.Activities
 			if (NextActivity == null)
 				self.CancelActivity();
 
-			var reservation = aircraft.Reservation;
-			if (reservation != null)
-			{
-				reservation.Dispose();
-				reservation = null;
-			}
+			aircraft.UnReserve();
 
 			var host = aircraft.GetActorBelow();
 			var hasHost = host != null;
--- a/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
+++ b/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
@@ -543,7 +543,10 @@ namespace OpenRA.Mods.Common.Traits
 					{
 						var res = order.TargetActor.TraitOrDefault<Reservable>();
 						if (res != null)
+						{
+							UnReserve();
 							Reservation = res.Reserve(order.TargetActor, self, this);
+						}
 
 						Action enter = () =>
 						{
--- a/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
+++ b/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
@@ -92,11 +92,11 @@ namespace OpenRA.Mods.Common.Traits
 		readonly Actor self;
 
 		UpgradeManager um;
+		IDisposable reservation;
 
 		[Sync] public int Facing { get; set; }
 		[Sync] public WPos CenterPosition { get; private set; }
 		public CPos TopLeft { get { return self.World.Map.CellContaining(CenterPosition); } }
-		public IDisposable Reservation;
 		public int ROT { get { return Info.ROT; } }
 
 		bool airborne;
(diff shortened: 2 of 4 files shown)


OpenRA.Mods.Common/Activities/Air/TakeOff.cs
```diff
@@ -30,12 +30,7 @@ public override Activity Tick(Actor self)
 			if (NextActivity == null)
 				self.CancelActivity();
 
-			var reservation = aircraft.Reservation;
-			if (reservation != null)
-			{
-				reservation.Dispose();
-				reservation = null;
-			}
+			aircraft.UnReserve();
```
RoosterDragon: This will now null out the `Reservation` field on `aircraft`, rather than the `reservation` local. This is intended to be a fix.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -541,9 +538,7 @@ public void ResolveOrder(Actor self, Order order)
 					}
 					else
 					{
-						var res = order.TargetActor.TraitOrDefault<Reservable>();
-						if (res != null)
-							Reservation = res.Reserve(order.TargetActor, self, this);
+						MakeReservation(order.TargetActor);
```
RoosterDragon: This will now mean `UnReserve` is called before setting a new reservation and is the likely fix for #9783.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -448,12 +450,7 @@ public bool CanEnterTargetNow(Actor self, Target target)
 			if (target.Positions.Any(p => self.World.ActorMap.GetActorsAt(self.World.Map.CellContaining(p)).Any(a => a != self && a != target.Actor)))
 				return false;
 
-			var res = target.Actor.TraitOrDefault<Reservable>();
-			if (res == null)
-				return true;
-
-			UnReserve();
-			Reservation = res.Reserve(target.Actor, self, this);
+			MakeReservation(target.Actor);
```
RoosterDragon: *PLEASE CHECK THIS*

The old code would bail early if the target was unreservable. This means any old reservation it had would not be cancelled, even though presumably it was now heading to a new target. This seems highly suspicious and so I have changed it.

Frankly - a method named `CanEnterTargetNow` is something I expect to be side-effect free - whereas it will actually make a reservation of the target you are checking. :/

RoosterDragon: OK I double checked this myself and I believe this to be a sensible change - anything calling this is expecting the aircraft to move to this target - so I think changing the behaviour so the aircraft releases any old reservation is fine.

RoosterDragon:
I think this is ready for review in its current form - I don't have any further changes to make.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -257,22 +257,24 @@ protected void ReserveSpawnBuilding()
...
-			var res = afld.TraitOrDefault<Reservable>();
+			MakeReservation(afld);
+		}
 
-			if (res != null)
-			{
-				UnReserve();
-				Reservation = res.Reserve(afld, self, this);
-			}
+		public void MakeReservation(Actor target)
```
penev92: More like TryMakeReservation() ?
Because it isn't unlikely for the reservation to fail/it's not guaranteed to happen.

pchote: Bikeshedding: If we rename it to `Try*` it should also return a bool.

penev92: True.

atlimit8:
Actually, `public bool MakeReservation(Actor target)` could be used to help close #9035.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
RoosterDragon: `Reservable.Reserve` never fails. If you call it on an already reserved target the old reservation is simply overwritten without warning - this probably leads to bad things but that's how it currently works.

All the current callers use `Reservable.IsReserved` to filter out already reserved locations before trying to make a reservation.

RoosterDragon:
The current reservation API is not structured to work in such a manner - I think any such rework would be best left for another PR.

Mailaender:
Airfield reservation still works in a robust way. :white_check_mark:

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/333551fe4037c33874ddc253a0d5c4bd64d4e387...49ac2f104172d750b830f9a6aa9eb11b1d33fc63)

PR #9791, opened 2015-10-28
Added Keyboard input filter to only register first key press.

### LINKED ISSUE #9737: Sidebar button keyboard shortcuts are triggered by repeating rate input rather than physical key press.

RichSuther:
Fix for #9737 to prevent key repeats triggering command keys multiple times.

Currently holding keyboard short cut keys triggers every key repeat, this changes to only register on physical key press.

Closes #9737.

### DIFF
--- a/OpenRA.Platforms.Default/Sdl2Input.cs
+++ b/OpenRA.Platforms.Default/Sdl2Input.cs
@@ -161,7 +161,11 @@ namespace OpenRA.Platforms.Default
 								Platform.CurrentPlatform == PlatformType.Windows)
 								Game.Exit();
 							else
-								inputHandler.OnKeyInput(keyEvent);
+							{
+								// Only register initial key press
+								if (e.key.repeat == 0)
+									inputHandler.OnKeyInput(keyEvent);
+							}
 
 							break;
 						}


Mailaender:
This works way better. :+1:

UberWaffe:
:+1:
Tested both this branch and current bleed on C&C TD mod.
This definitely works much better.

OpenRA.Platforms.Default/Sdl2Input.cs
```diff
@@ -161,7 +161,11 @@ public void PumpInput(IInputHandler inputHandler)
 								Platform.CurrentPlatform == PlatformType.Windows)
 								Game.Exit();
 							else
-								inputHandler.OnKeyInput(keyEvent);
+							{
```
atlimit8: Our convention is to use braces when there is more than one statement.

OpenRA.Platforms.Default/Sdl2Input.cs
```diff
@@ -161,7 +161,11 @@ public void PumpInput(IInputHandler inputHandler)
...
+								// Only register initial key press
+								if (e.key.repeat == 0)
```
atlimit8: Why not `else if`?

RichSuther: I'm following the initial implementation, specifically the Special case workaround for windows users If this is not needed then the normal programme flow of a key press whatever that is, is  followed. - I would argue an `else` is a better use here to show that.

It would also allow in future if more of these special cases where needed they could be added as an `else if` while the normal flow is the else.

That was my thinking behind it anyway.

penev92: Since the 2 ifs aren't logically connected, that would be confusing.

obrakmann:
It seems this broke the backspace key in text fields. Keeping the backspace key pressed will only erase one character and stop then instead of removing the whole line.

OpenRA.Platforms.Default/Sdl2Input.cs
atlimit8: An exception for the backspace key can be made here.

pchote: I'm not a fan of hardcoding a rather high level implementation detail (backspace in a textfield should repeat) in our lowest level platform code.

atlimit8: @pchote: how about replacing this with `public bool KeyInput.Repeated` and have `Hotkey.FromKeyInput(KeyInput ki)` return `Hotkey.Invalid` if `ki.Repeated` is true?

RichSuther: The chat itself still can handle repeating keys it just seems an exception to the backspace input has been made, which i think is due to how the widgets handle keyboard shortcuts. - This means that individual widgets require an individual filter as per requirement not the low level platform code.

I'm think this could be done via adding the `e.key.repeat` value into the `KeyInput` datatype, this would then mean each widget would need to be refactored to handle whether or not it should respond to key repeat or should be a singular press.

So I agree with @pchote Its far to low level of an implementation unfortunately, and it should be the widgets that make the decision I think that would be the ideal solution. but it does require a bit of refactoring.

So back to the drawing board on this one, could we un-merge and i will work on a better solution.

RichSuther:
The chat itself still can handle repeating keys it just seems an exception to the backspace input has been made, which i think is due to how the widgets handle keyboard shortcuts. - This means that individual widgets require an individual filter as and when required not within the low level platform code.

I'm think this could be done via adding the e.key.repeat value into the KeyInput datatype, this would then mean each widget would need to be refactored to handle whether or not it should respond to key repeat or should be a singular press.

So I agree with @pchote Its far to low level of an implementation unfortunately, and it should be the widgets that make the decision I think that would be the ideal solution. but it does require a bit of refactoring.

So back to the drawing board on this one, could we un-merge and i will work on a better solution.

PR #9798, opened 2015-10-29
Lower GDI medic self heal speed

teees:
Current self heal speed is way too fast (you need >3 e1 to kill one selfhealing medic)

### DIFF
--- a/mods/ts/rules/gdi-infantry.yaml
+++ b/mods/ts/rules/gdi-infantry.yaml
@@ -51,6 +51,7 @@ MEDIC:
 	WithInfantryBody:
 		AttackSequence: heal
 	SelfHealing:
+		Ticks: 60
 	Passenger:
 		PipType: Red
 


Phrohdoh:
Is it a good time to start attempting to balance an incomplete mod?
If others are fine with this, then sure :+1:.

penev92:
TS doesn't care about releases, and the playtest/release don't depend on TS in any way, so this is harmless.
We do want to properly balance it at some point, and I don't think importing the proper stats for actors needs to wait for the mod to be feature-complete.

pchote:
How does this new value compare with the original game?

teees:
Really hard to say, as self healing was hard coded. Feels closer to original though.

pchote:
In the original game it takes 31 minigunner bursts to kill a medic.  With this PR it now takes 34 minigunner bursts.  Using `Ticks: 60` matches the original game.

pchote:
:+1: after setting Ticks: 60.

teees:
changed and rebased

abcdefg30:
:+1:

PR #9357, opened 2015-09-16
Added a LineImpactProjectile (Attempt 2)

UberWaffe:
Attempt number 2 at  #6252

New projectile can create multiple weapon impacts along its travel vector.
(i.e. apply all warheads effects at multiple positions as it travels).

It has a leading head position, and a following tail position, and renders a beam (LaserZap) between these two positions. Weapon impacts are created (iteratively over the distance) between the head and tail, at definable intervals.

The tail position (origin of beam) tracks the firing actor until it detaches and travels on its own. This means that if you move the firing actor while the beam is firing, it does track correctly.

You can currently tell it to:
- Define how many ticks the tail must lag the head (i.e. the length of the travelling beam).
- Create an impact every WRange, every X ticks, between head and tail.
- Only create an impact if at least WRange away from its source.
- Only create an impact if at least within WRange of its target.
- Keep traveling for up to WRange beyond its target.
- Have damage falloff (essentially passed as a firepowermodifier to the weapon during impact) that
  - Has a WRange before it starts
  - Has a WRange at which it reaches maximum falloff
  - Has a maximum falloff.

Was based off mixture of Bullet and LaserZap.

Does NOT support
- A proper rendering engine to make it look like something other than a basic beam. (Currently looks fine for instant-hit beams. But slow-moving beams are ugly.)

I still have to do:
- The target is a static position, so if you fire at a moving actor, even while the firing actor is still busy emitting the beam, the head does not 'track' the target. Not sure how to do this. (Would like to make this an Info setting.)
- ~~Add Ray-casting, finding all actors within the bounds of the beam's length and width. #8186~~
- ~~Make warheads apply to actors directly, ignoring warhead spreads.~~ (Causes crashes and unexpected behavior. Will require some warhead refactoring, or at least a new warhead type.)
- ~~When the actor is ordered to stop, the beam should stop emitting if still busy.~~
- ~~When the actor dies, the beam should stop emitting if still busy.~~

Still to be tested:
- ~~That walls properly block the beam if the beam is blockable.~~
- ~~That the projectile properly works for 'short' beams, such as D2K sonic ball.~~ Works, though very short beams render as practically invisible.
- ~~That the projectile properly works for brief instant-hit beams, such as TS railguns.~~ Works

Closes  #4458.

I've attempted to apply it to the D2K sonic tank, and the TS disruptor tank, but balancing needs work.

[...]

PR #9800, opened 2015-10-29
Hitshape

huwpascoe:
Final part of https://github.com/OpenRA/OpenRA/issues/6820, new shapes added: Circle, Capsule, Rectangle.

demonstration with arbitrary values:

```
MCV:
    Health:
        HP: 600
        Shape: Capsule
            Radius: 426
            Vertices: 0, -512, 0, 512
```

hit-a
hit-b

```
4TNK:
    Health:
        HP: 900
        Shape: Rectangle
            Vertices: -512, -512, 512, 512
```

```
AFLD:
    Health:
        HP: 1000
        Shape: Rectangle
            Vertices: -1280,-384, 1280, 1024
```

hit-c

### DIFF
--- a/OpenRA.Mods.Common/HitShapes/Capsule.cs
+++ b/OpenRA.Mods.Common/HitShapes/Capsule.cs
@@ -1,0 +1,93 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Drawing;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
+
+namespace OpenRA.Mods.Common.HitShapes
+{
+	public class CapsuleShape : IHitShape
+	{
+		public WDist OuterRadius { get; private set; }
+
+		[FieldLoader.Require]
+		public readonly int2 PointA;
+
+		[FieldLoader.Require]
+		public readonly int2 PointB;
+
+		public readonly WDist Radius = new WDist(426);
+
+		int2 ab;
+		int abLenSq;
+
+		public CapsuleShape() { }
+
+		public CapsuleShape(int2 a, int2 b, WDist radius)
+		{
+			PointA = a;
+			PointB = b;
+			Radius = radius;
+		}
+
+		public void Initialize()
+		{
+			ab = PointB - PointA;
+			abLenSq = ab.LengthSquared / 1024;
+
+			if (abLenSq == 0)
+				throw new YamlException("This Capsule describes a circle. Use a Circle HitShape instead.");
+
+			OuterRadius = Radius + new WDist(Math.Max(PointA.Length, PointB.Length));
+		}
+
+		public WDist DistanceFromEdge(WVec v)
+		{
+			var p = new int2(v.X, v.Y);
+
+			var t = int2.Dot(p - PointA, ab) / abLenSq;
+
+			if (t < 0)
+				return new WDist(Math.Max(0, (PointA - p).Length - Radius.Length));
+			if (t > 1024)
+				return new WDist(Math.Max(0, (PointB - p).Length - Radius.Length));
+
+			var projection = PointA + new int2(
+				(ab.X * t) / 1024,
+				(ab.Y * t) / 1024);
+
+			var distance = (projection - p).Length;
+
+			return new WDist(Math.Max(0, distance - Radius.Length));
+		}
+
+		public WDist DistanceFromEdge(WPos pos, Actor actor)
+		{
+			return DistanceFromEdge((pos - actor.CenterPosition).Rotate(-actor.Orientation));
+		}
+
+		public void DrawCombatOverlay(WorldRenderer wr, RgbaColorRenderer wcr, Actor actor)
+		{
+			var a = actor.CenterPosition + new WVec(PointA.X, PointA.Y, 0).Rotate(actor.Orientation);
+			var b = actor.CenterPosition + new WVec(PointB.X, PointB.Y, 0).Rotate(actor.Orientation);
+
+			var offset = new WVec(a.Y - b.Y, b.X - a.X, 0);
+			offset = offset * Radius.Length / offset.Length;
+
+			var c = Color.Yellow;
+			RangeCircleRenderable.DrawRangeCircle(wr, a, Radius, 1, c, 0, c);
+			RangeCircleRenderable.DrawRangeCircle(wr, b, Radius, 1, c, 0, c);
+			wcr.DrawLine(new[] { wr.ScreenPosition(a - offset), wr.ScreenPosition(b - offset) }, 1, c);
+			wcr.DrawLine(new[] { wr.ScreenPosition(a + offset), wr.ScreenPosition(b + offset) }, 1, c);
+		}
+	}
+}
--- a/OpenRA.Mods.Common/HitShapes/Circle.cs
+++ b/OpenRA.Mods.Common/HitShapes/Circle.cs
@@ -1,0 +1,46 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Drawing;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
+
+namespace OpenRA.Mods.Common.HitShapes
+{
+	public class CircleShape : IHitShape
+	{
+		public WDist OuterRadius { get { return Radius; } }
+
+		[FieldLoader.Require]
+		public readonly WDist Radius = new WDist(426);
+
+		public CircleShape() { }
+
+		public CircleShape(WDist radius) { Radius = radius; }
+
+		public void Initialize() { }
+
+		public WDist DistanceFromEdge(WVec v)
+		{
+			return new WDist(Math.Max(0, v.Length - Radius.Length));
+		}
+
+		public WDist DistanceFromEdge(WPos pos, Actor actor)
+		{
+			return DistanceFromEdge(pos - actor.CenterPosition);
+		}
+
+		public void DrawCombatOverlay(WorldRenderer wr, RgbaColorRenderer wcr, Actor actor)
+		{
+			RangeCircleRenderable.DrawRangeCircle(wr, actor.CenterPosition, Radius, 1, Color.Yellow, 0, Color.Yellow);
+		}
+	}
+}
--- a/OpenRA.Mods.Common/HitShapes/IHitShape.cs
+++ b/OpenRA.Mods.Common/HitShapes/IHitShape.cs
@@ -1,0 +1,25 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using OpenRA.Graphics;
+
+namespace OpenRA.Mods.Common.HitShapes
+{
+	public interface IHitShape
+	{
+		WDist OuterRadius { get; }
+
+		WDist DistanceFromEdge(WVec v);
+		WDist DistanceFromEdge(WPos pos, Actor actor);
+
+		void Initialize();
+		void DrawCombatOverlay(WorldRenderer wr, RgbaColorRenderer wcr, Actor actor);
+	}
+}
--- a/OpenRA.Mods.Common/HitShapes/Rectangle.cs
+++ b/OpenRA.Mods.Common/HitShapes/Rectangle.cs
@@ -1,0 +1,80 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Drawing;
+using System.Linq;
+using OpenRA.Graphics;
+
+namespace OpenRA.Mods.Common.HitShapes
+{
+	public class RectangleShape : IHitShape
+	{
+		public WDist OuterRadius { get; private set; }
+
+		[FieldLoader.Require]
+		public readonly int2 TopLeft;
+
+		[FieldLoader.Require]
+		public readonly int2 BottomRight;
+
+		int2 quadrantSize;
+		int2 center;
+
+		WVec[] combatOverlayVerts;
+
+		public RectangleShape() { }
+
+		public RectangleShape(int2 tl, int2 br)
+		{
+			TopLeft = tl;
+			BottomRight = br;
+		}
+
+		public void Initialize()
+		{
+			if (TopLeft.X >= BottomRight.X || TopLeft.Y >= BottomRight.Y)
+				throw new YamlException("TopLeft and BottomRight points are invalid.");
+
+			quadrantSize = (BottomRight - TopLeft) / 2;
+			center = TopLeft + quadrantSize;
+
+			OuterRadius = new WDist(Math.Max(TopLeft.Length, BottomRight.Length));
+
+			combatOverlayVerts = new WVec[]
+			{
+				new WVec(TopLeft.X, TopLeft.Y, 0),
+				new WVec(BottomRight.X, TopLeft.Y, 0),
+				new WVec(BottomRight.X, BottomRight.Y, 0),
+				new WVec(TopLeft.X, BottomRight.Y, 0)
+			};
+		}
+
+		public WDist DistanceFromEdge(WVec v)
+		{
+			var r = new int2(
+				Math.Max(Math.Abs(v.X - center.X) - quadrantSize.X, 0),
+				Math.Max(Math.Abs(v.Y - center.Y) - quadrantSize.Y, 0));
+
+			return new WDist(r.Length);
+		}
+
+		public WDist DistanceFromEdge(WPos pos, Actor actor)
+		{
+			return DistanceFromEdge((pos - actor.CenterPosition).Rotate(-actor.Orientation));
+		}
+
+		public void DrawCombatOverlay(WorldRenderer wr, RgbaColorRenderer wcr, Actor actor)
+		{
+			var verts = combatOverlayVerts.Select(v => wr.ScreenPosition(actor.CenterPosition + v.Rotate(actor.Orientation)));
+			wcr.DrawPolygon(verts.ToArray(), 1, Color.Yellow);
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Traits/Health.cs
+++ b/OpenRA.Mods.Common/Traits/Health.cs
@@ -8,7 +8,9 @@
  */
 #endregion
 
+using System;
 using System.Linq;
+using OpenRA.Mods.Common.HitShapes;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
@@ -17,12 +19,38 @@ namespace OpenRA.Mods.Common.Traits
 	{
 		[Desc("HitPoints")]
 		public readonly int HP = 0;
-
-		[Desc("Physical size of the unit used for damage calculations. Impacts within this radius apply full damage.")]
-		public readonly WDist Radius = new WDist(426);
-
 		[Desc("Trigger interfaces such as AnnounceOnKill?")]
 		public readonly bool NotifyAppliedDamage = true;
+
+		[FieldLoader.LoadUsing("LoadShape")]
+		public readonly IHitShape Shape;
+
+		static object LoadShape(MiniYaml yaml)
+		{
+			IHitShape ret;
+
+			var shapeNode = yaml.Nodes.Find(n => n.Key == "Shape");
+			var shape = shapeNode != null ? shapeNode.Value.Value : string.Empty;
+
+			if (!string.IsNullOrEmpty(shape))
+			{
+				ret = Game.CreateObject<IHitShape>(shape + "Shape");
+
+				try
+				{
+					FieldLoader.Load(ret, shapeNode.Value);
+				}
+				catch (YamlException e)
+				{
+					throw new YamlException("HitShape {0}: {1}".F(shape, e.Message));
+				}
+			}
+			else
+				ret = new CircleShape();
+
+			ret.Initialize();
+			return ret;
+		}
 
 		public virtual object Create(ActorInitializer init) { return new Health(init, this); }
 	}
--- a/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
@@ -78,7 +78,7 @@ namespace OpenRA.Mods.Common.Warheads
 					continue;
 
 				// If the impact position is within any actor's health radius, we have a direct hit
-				if ((unit.CenterPosition - pos).LengthSquared <= healthInfo.Radius.LengthSquared)
+				if ((unit.CenterPosition - pos).LengthSquared <= healthInfo.Shape.OuterRadius.LengthSquared)
 					return true;
 			}
 
--- a/OpenRA.Test/OpenRA.Mods.Common/ShapeTest.cs
+++ b/OpenRA.Test/OpenRA.Mods.Common/ShapeTest.cs
@@ -1,0 +1,95 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using NUnit.Framework;
+using OpenRA.Mods.Common.HitShapes;
+
+namespace OpenRA.Test
+{
+	[TestFixture]
+	public class ShapeTest
+	{
+		IHitShape shape;
+
+		[TestCase(TestName = "CircleShape reports accurate distance")]
+		public void Circle()
+		{
+			shape = new CircleShape(new WDist(1234));
+			shape.Initialize();
+
+			Assert.That(shape.DistanceFromEdge(new WVec(100, 100, 0)).Length,
+				Is.EqualTo(0));
+
+			Assert.That(shape.DistanceFromEdge(new WVec(1000, 0, 0)).Length,
+				Is.EqualTo(0));
+
+			Assert.That(shape.DistanceFromEdge(new WVec(2000, 2000, 0)).Length,
+				Is.EqualTo(1594));
+
+			Assert.That(new CircleShape(new WDist(73))
+				.DistanceFromEdge(new WVec(150, -100, 0)).Length,
+				Is.EqualTo(107));
+
+			Assert.That(new CircleShape(new WDist(55))
+				.DistanceFromEdge(new WVec(30, -45, 0)).Length,
+				Is.EqualTo(0));
+		}
+
+		[TestCase(TestName = "CapsuleShape report accurate distance")]
+		public void Capsule()
+		{
+			shape = new CapsuleShape(new int2(-50, 0), new int2(500, 235), new WDist(50));
+			shape.Initialize();
+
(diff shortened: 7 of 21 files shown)


huwpascoe:
This PR not only detects a direct hit, but also computes damage falloff

Like this (visualization from stack overflow)

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -17,14 +18,27 @@ public class HealthInfo : ITraitInfo, UsesInit<HealthInit>
...
+		public readonly int2[] Vertices = null;
 
-		[Desc("Trigger interfaces such as AnnounceOnKill?")]
-		public readonly bool NotifyAppliedDamage = true;
+		public IHitShape ShapeInstance { get; private set; }
 
-		public virtual object Create(ActorInitializer init) { return new Health(init, this); }
+		public virtual object Create(ActorInitializer init)
+		{
+			switch (Shape)
```
pchote: Use `IRulesetLoaded` / `RulesetLoaded` instead of putting this here (where it is invoked every time a new actor is created).

OpenRA.Mods.Common/HitShapes/IHitShape.cs
```diff
@@ -0,0 +1,23 @@
...
+using OpenRA.Graphics;
+
+namespace OpenRA.Mods.Common.HitShapes
+{
+	public enum HitShape { Circle, Rectangle, Capsule }
+
+	public interface IHitShape
+	{
+		int OuterRadius { get; }
+		int DistanceFromEdge(int2 p);
```
pchote: This should take a `WVec` and return a `WDist` to cleanly integrate with the rest of our world coordinates.

OpenRA.Mods.Common/HitShapes/IHitShape.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using OpenRA.Graphics;
+
+namespace OpenRA.Mods.Common.HitShapes
+{
+	public enum HitShape { Circle, Rectangle, Capsule }
+
+	public interface IHitShape
+	{
+		int OuterRadius { get; }
```
pchote: `WDist` here too.

OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
```diff
@@ -70,7 +70,8 @@ public static bool GetDirectHit(World world, CPos cell, WPos pos)
 					continue;
 
 				// If the impact position is within any actor's health radius, we have a direct hit
-				if ((unit.CenterPosition - pos).LengthSquared <= healthInfo.Radius.LengthSquared)
+				var or = healthInfo.ShapeInstance.OuterRadius;
```
pchote: You can then use `.LengthSquared` instead of `or*or`.

OpenRA.Mods.Common/HitShapes/Capsule.cs
```diff
@@ -0,0 +1,74 @@
...
+			this.radius = radius;
+
+			outerRadius = radius + Math.Max(v.Length, w.Length);
+		}
+
+		public int DistanceFromEdge(int2 p)
+		{
+			var vec = w - v;
+
+			var lenSq = vec.LengthSquared >> 10;
```
pchote: Just write `/ 1024` and let the compiler do its job.

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -17,14 +18,27 @@ public class HealthInfo : ITraitInfo, UsesInit<HealthInit>
...
 		[Desc("HitPoints")]
 		public readonly int HP = 0;
+		[Desc("Trigger interfaces such as AnnounceOnKill?")]
+		public readonly bool NotifyAppliedDamage = true;
 
-		[Desc("Physical size of the unit used for damage calculations. Impacts within this radius apply full damage.")]
+		[Desc("Shape of the unit used to calculate warhead damage")]
+		public readonly HitShape Shape = HitShape.Circle;
 		public readonly WDist Radius = new WDist(426);
+		public readonly int2[] Vertices = null;
```
Phrohdoh: Can we get descriptions for these new fields?

OpenRA.Mods.Common/Traits/Health.cs
pchote: Actually, I would have annotate `ShapeInstance` with a custom `[LoadUsing]` and then make the vertices / other info a sub-tree of the field.  You can then drop the Vertices definition here, where it doesn't really belong.

pchote:
I haven't looked closely, but it looks mostly sensible overall.  Cool!

OpenRA.Mods.Common/Traits/Health.cs
huwpascoe: Yay I wondered if there was something to do that

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -8,21 +8,36 @@
...
 
-		[Desc("Trigger interfaces such as AnnounceOnKill?")]
-		public readonly bool NotifyAppliedDamage = true;
+		public IHitShape ShapeInstance { get; private set; }
+
+		public void RulesetLoaded(Ruleset rules, ActorInfo info)
```
pchote: (copied from the previous tree, where it is now marked as outdated):



> Actually, I would annotate ShapeInstance with a custom [LoadUsing] and then make the vertices / other info a sub-tree of the field. You can then drop the Vertices from the parent Health, where it doesn't really belong.

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -17,13 +19,36 @@ public class HealthInfo : ITraitInfo, UsesInit<HealthInit>
...
+			if (nodes.ContainsKey("Radius"))
+				radius = FieldLoader.GetValue<int>("Radius", nodes["Radius"].Value);
+
+			int2[] verts = null;
+			if (nodes.ContainsKey("Vertices"))
+				verts = FieldLoader.GetValue<int2[]>("Vertices", nodes["Vertices"].Value);
+
+			switch (shape)
+			{
+				case "Rectangle": return new Rectangle(verts[0], verts[1]);
```
atlimit8: This should include:
```c#
if (Vertices == null || Vertices.Length != 2)
	throw new YamlException("Rectangles require top-left and bottom-right vertices");
```

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -17,13 +19,36 @@ public class HealthInfo : ITraitInfo, UsesInit<HealthInit>
...
+				case "Capsule": return new Capsule(verts[0], verts[1], radius);
```
atlimit8: This should include:
```c#
if (Vertices == null || Vertices.Length != 2)
	throw new YamlException("Capsules require top-left and bottom-right inner rectangle vertices");
```

atlimit8:
You replaced RulesetLoaded?
Why?

atlimit8:
I wrote it for situations like this.

huwpascoe:
@atlimit8 to quote pchote

"Actually, I would annotate ShapeInstance with a custom [LoadUsing] and then make the vertices / other info a sub-tree of the field. You can then drop the Vertices from the parent Health, where it doesn't really belong."

I applied that change. Which should I go for then, IRulesetLoaded or LoadUsing?

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -17,13 +19,36 @@ public class HealthInfo : ITraitInfo, UsesInit<HealthInit>
...
+		static object LoadShape(MiniYaml yaml)
+		{
+			var nodes = yaml.ToDictionary();
+
+			string shape = "Circle";
+			if (nodes.ContainsKey("Shape"))
+				shape = FieldLoader.GetValue<string>("Shape", nodes["Shape"].Value);
+
+			int radius = 426;
+			if (nodes.ContainsKey("Radius"))
```
atlimit8: Is this what it is supposed to parse?
```Yaml
MCV:
    Health:
        HP: 600
        Shape: Capsule
            Radius: 426
            Vertices: 0, -512, 0, 512
```
If so, then `nodes = nodes["Shape"].Value`.

atlimit8:
LoadUsing should be used for
```Yaml
MCV:
    Health:
        HP: 600
        Shape: Capsule
            Radius: 426
            Vertices: 0, -512, 0, 512
```
because recursive MinYaml loading is not supported (especially custom object creation dispatch).

huwpascoe:
@atlimit8 Is this better?

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -17,13 +19,40 @@ public class HealthInfo : ITraitInfo, UsesInit<HealthInit>
...
+				if (nodes.ContainsKey("Radius"))
+					radius = FieldLoader.GetValue<int>("Radius", nodes["Radius"].Value);
+
+				if (nodes.ContainsKey("Vertices"))
+					verts = FieldLoader.GetValue<int2[]>("Vertices", nodes["Vertices"].Value);
+			}
+
+			switch (shape)
+			{
+				case "Rectangle": return new Rectangle(verts[0], verts[1]);
```
pchote: If you use reflection to create these and have the constructor accept a list of miniyaml nodes then you can avoid hardcoding the types, and modders could add new ones.

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -17,13 +19,40 @@ public class HealthInfo : ITraitInfo, UsesInit<HealthInit>
...
+
+			var shapeNode = yaml.Nodes.Find(n => n.Key == "Shape");
+			if (shapeNode != null)
+			{
+				shape = shapeNode.Value.Value;
+
+				var nodes = shapeNode.Value.ToDictionary();
+
+				if (nodes.ContainsKey("Radius"))
+					radius = FieldLoader.GetValue<int>("Radius", nodes["Radius"].Value);
```
atlimit8: `WDist` type should be kept here instead of loading an `int` for clarity of the length of the radius.

OpenRA.Mods.Common/Traits/Health.cs
atlimit8: This should include:
```c#
if (Vertices == null || Vertices.Length != 2)
	throw new YamlException("Rectangel requires top-left and bottom-right vertices");
```
- or better -
Use @pchote's suggestion and put these checks in the constructors.

atlimit8:
Put those two checks in the cases or in the constructors.

OpenRA.Mods.Common/Traits/Health.cs
huwpascoe: Can you show me an example? Reflection isn't my strong point

atlimit8: Add `void Loaded()` to `IHitShape`:
```c#
	public interface IHitShape
	{
		WDist OuterRadius { get; }
		WDist DistanceFromEdge(WVec p);
		void DrawCombatOverlay(WorldRenderer wr, LineRenderer wlr, WPos center, WRot angle);
		void Loaded(); // Called once loaded from MiniYaml
	}
```

---
Modify hit shapes to something like:
```c#
	public class CircleHitShape : IHitShape
	{
		public WDist OuterRadius { get { return Radius; } }
		public readonly WDist Radius = new WDist(426);

		public void Loaded() { }

		...
	}
```
```c#
	public class CapsuleHitShape : IHitShape
	{
		public WDist OuterRadius { get { return outerRadius; } }
		[FieldLoader.required]
		public readonly int2 TopLeft;
		[FieldLoader.required]
		public readonly int2 BottomRight;
		public readonly WDist Radius = new WDist(426);
		WDist outerRadius;

		public void Loaded()
		{
			outerRadius = new WDist(radius + Math.Max(v.Length, w.Length));
		}

		...
	}
```
```c#
	public class RectangleHitShape : IHitShape
	{
		public WDist OuterRadius { get { return outerRadius; } }
		[FieldLoader.required]
		public readonly int2 Size;
		public readonly int2 Center = new int2(0, 0);
		public readonly WDist Radius = new WDist(426);
		WDist outerRadius;
		readonly WVec[] combatOverlayVerts;

		public void Loaded()
		{
			if (Size.X <= 0 || Size.Y <= 0)
				throw new YamlException("Size must be positive values");

			var tl = Center - Size / 2;
			var br = Center + Size / 2;
			combatOverlayVerts = new WVec[]
			{
				new WVec(tl.X, tl.Y, 0),
				new WVec(br.X, tl.Y, 0),
				new WVec(br.X, br.Y, 0),
				new WVec(tl.X, br.Y, 0),
				new WVec(tl.X, tl.Y, 0)
			};

			outerRadius = new WDist(Math.Max(tl.Length, br.Length));
		}

		...
	}
```
Note: each shape can have its own MiniYaml fields

---
Loader code:
```c#
		static object LoadShape(MiniYaml yaml)
		{
			var shapeNode = yaml.Nodes.Find(n => n.Key == "Shape");
			var shape = shapeNode == null || string.IsNullOrEmpty(shapeNode.Value) ? "Circle" : shapeNode.Value;
			var ret = Game.CreateObject<IHitShape>(shape + "HitShape");
			if (shapeNode != null)
			{
				try
				{
					FieldLoader.Load(ret, shapeNode);
				}
				catch (YamlException ex)
				{
					throw new YamlException("Shape " + shape + ": " + ex.Message));
				}
			}

			ret.Loaded();
			return ret;
		}
```

OpenRA.Mods.Common/HitShapes/IHitShape.cs
```diff
@@ -0,0 +1,21 @@
...
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using OpenRA.Graphics;
+
+namespace OpenRA.Mods.Common.HitShapes
+{
+	public interface IHitShape
```
pchote: Can you please also add implementations for line-vs-shape intersections so that this can play nicely with #9357?

OpenRA.Mods.Common/HitShapes/Capsule.cs
```diff
@@ -0,0 +1,75 @@
...
+			this.w = w;
+			this.radius = radius;
+
+			outerRadius = new WDist(radius + Math.Max(v.Length, w.Length));
+		}
+
+		public WDist DistanceFromEdge(WVec pv)
+		{
+			var p = new int2(pv.X, pv.Y);
+			var vec = w - v;
```
pchote: `w` and `v` are both constant, so you can calculate `vec` once in the constructor.

huwpascoe:
https://github.com/OpenRA/OpenRA/pull/9800#discussion_r43492030
I've just had a short attempt at doing it but ray->rectangle, ray->capsule is a lot more difficult so this will take some time.

@atlimit8 Thanks, it works well. I'll have the next PR soon

OpenRA.Mods.Common/HitShapes/IHitShape.cs
UberWaffe: tl;dr: ```WDist DistanceFromEdge(WVec p);``` might already suit the needs of #9357.

!tl;dr
With regards to #9357, does the implementations of the line below ```WDist DistanceFromEdge(WVec p);``` give you the distance from the actor center position to the edge of its shape along a vector?

If so, in #9357 the [actor width line] (https://github.com/OpenRA/OpenRA/pull/9357/files#diff-5f66c498eea5ea840b8d556a7285788eR49) can be adjusted to use this instead of assuming circle radius (i.e. equal radius in all directions).

Basically, I could use it to get the actor's 'width' along the vector from its center point to the closest point on the line.

UberWaffe:
From the [comment] (https://github.com/OpenRA/OpenRA/pull/9800#discussion_r43573052) above:

@huwpascoe:
tl;dr: WDist DistanceFromEdge(WVec p); might already suit the needs of #9357.

!tl;dr
With regards to #9357, does the implementations of the line below WDist DistanceFromEdge(WVec p); give you the distance from the actor's center position to the edge of its shape along a vector? (it appears to do that)

If so, in #9357 the actor width line can be adjusted to use WDist DistanceFromEdge(WVec p); instead of assuming circle radius (i.e. equal radius in all directions).

Basically, I could use it to get the actor's 'width' along the vector from its center point to the closest point on the line.
The code already has both WPos, so it should be straightforward.

I think the following should be what it ends up as in #9357

reaperrr:
> OpenRA.Mods.Common/HitShapes/Circle.cs(24,7): error CS0414: Warning as Error: The private field `OpenRA.Mods.Common.HitShapes.HitShapeCircle.radiusSq' is assigned but its value is never used

huwpascoe:
@reaperrr Don't worry I just noticed it : P (you'd think visual studio would have picked up on that first)

@UberWaffe If you think you can make do with the dist methods then I'll post as is. I got as far as line/circle and line/rectangle (line/capsule still stuck on) so if it turns out you want those methods later I can post them separately.

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -17,13 +19,39 @@ public class HealthInfo : ITraitInfo, UsesInit<HealthInit>
...
+		static object LoadShape(MiniYaml yaml)
+		{
+			IHitShape ret;
+
+			var shapeNode = yaml.Nodes.Find(n => n.Key == "Shape");
+			var shape = shapeNode != null ? shapeNode.Value.Value : string.Empty;
+
+			if (!string.IsNullOrEmpty(shape))
+			{
+				ret = Game.CreateObject<IHitShape>("HitShape" + shape);
```
atlimit8: The idiom is to add a suffix instead of a prefix:
```
OpenRA.Game/GameRules/ActorInfo.cs:112:    var info = Game.CreateObject<ITraitInfo>(traitName + "Info");
OpenRA.Game/GameRules/WeaponInfo.cs:78:    var ret = Game.CreateObject<IProjectileInfo>(proj.Value + "Info");
OpenRA.Game/GameRules/WeaponInfo.cs:88:    var ret = Game.CreateObject<IWarhead>(node.Value.Value + "Warhead");
OpenRA.Game/Map/ActorReference.cs:46:      var info = Game.CreateObject<IActorInit>(traitName + "Init");
OpenRA.Game/Widgets/WidgetLoader.cs:77:    return Game.ModData.ObjectCreator.CreateObject<Widget>(widgetType + "Widget", args);

```
It also sounds better.

huwpascoe:
suffix'd "Shape" to match the rest of the project

huwpascoe:
Added unit tests and fixed a bunch of stupid mistakes.

huwpascoe:
Fixed merge conflict

huwpascoe:
Merged with upstream, still passes.

penev92:
You need to rebase onto bleed to get rid of the "Merge" commits.

huwpascoe:
lol mybad. Rebased.

pchote:
We're really low on review bandwidth right now, and the time that we do have is mainly focused towards the next release.  It may be a while longer before we can continue reviewing this, but please don't take that as a slight against you personally!

huwpascoe:
No worries!

OpenRA.Mods.Common/HitShapes/Circle.cs
```diff
@@ -0,0 +1,40 @@
...
+using OpenRA.Graphics;
+
+namespace OpenRA.Mods.Common.HitShapes
+{
+	public class CircleShape : IHitShape
+	{
+		public WDist OuterRadius { get { return Radius; } }
+
+		[FieldLoader.Require]
+		public readonly WDist Radius = new WDist(426);
```
penev92: Is it me, or is this 0.416 cells? Please default it to 512 or something.

OpenRA.Mods.Common/HitShapes/Circle.cs
```diff
@@ -0,0 +1,40 @@
...
+
+		public CircleShape() { }
+
+		public CircleShape(WDist radius) { Radius = radius; }
+
+		public void Initialize() { }
+
+		public WDist DistanceFromEdge(WVec p)
+		{
+			return new WDist(Math.Max(0, p.Length - Radius.Length));
```
penev92: I don't get it, why the 0?

OpenRA.Mods.Common/HitShapes/Circle.cs
reaperrr: @penev92: The internal default for health radius was also 426 ( = 10 pixel in RA/TD).

penev92: Wow. YAY for random defaults...
Never mind then...

OpenRA.Mods.Common/HitShapes/Circle.cs
huwpascoe: It returns the distance relative to the edge of the hitshape, <= 0 means inside the hitshape, and so a direct hit. Since negative distance makes no sense, needs to be clamped to 0.

penev92: Ah, so you only care for values outside the shape. OK then.

huwpascoe:
fixed build error with bleed

OpenRA.Mods.Common/HitShapes/IHitShape.cs
```diff
@@ -0,0 +1,24 @@
...
+	{
+		WDist OuterRadius { get; }
+
+		WDist DistanceFromEdge(WVec p);
```
pchote: This requires `p` to be already rotated into the local coordinate system of the victim (see SpreadDamageWarhead:L69) which breaks the abstraction a bit, IMO.

How about we make this `WDist DistanceFromEdge(Actor self, WPos position)`, and then calculate the vector, rotate it, and return the distance all internally.  The circle shape can then just return the vector distance directly.

huwpascoe: Seems kinda repetitive, but I'll give it a try

huwpascoe:
Okay replacing the vectors with Actor is a lot cleaner after all

OpenRA.Test/OpenRA.Test.csproj
```diff
@@ -58,6 +59,11 @@
       <Name>OpenRA.Game</Name>
       <Private>False</Private>
     </ProjectReference>
+    <ProjectReference Include="..\OpenRA.Mods.Common\OpenRA.Mods.Common.csproj">
```
UberWaffe: Are these additions in `csproj` needed?

UberWaffe:
Did a quick test in TS.
The rectangle hitshape does not appear to take the isometric perspective into account.

OpenRA.Test/OpenRA.Test.csproj
pchote: You'll also need to fix the makefile:
```
paul@Mjolnir:OpenRA+pr/9800$ make nunit
CSC OpenRA.Test.dll
OpenRA.Test//OpenRA.Game/ShapeTest.cs(13,14): error CS0234: The type or namespace name `Mods' does not exist in the namespace `OpenRA'. Are you missing an assembly reference?
OpenRA.Test//OpenRA.Game/ShapeTest.cs(20,3): error CS0246: The type or namespace name `IHitShape' could not be found. Are you missing an assembly reference?
Compilation failed: 2 error(s), 0 warnings
```

OpenRA.Test/OpenRA.Game/ShapeTest.cs
```diff
@@ -0,0 +1,96 @@
...
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.IO;
+using NUnit.Framework;
+using OpenRA.Mods.Common.HitShapes;
+
+namespace OpenRA.Test
```
pchote: This appears to be in the wrong place.  The file path implies that it is a Game test, but it then depends on things from mods.common.  Can you put this in a new Mods.Common dir instead?

OpenRA.Test/OpenRA.Game/ShapeTest.cs
```diff
@@ -0,0 +1,96 @@
...
+			shape.Initialize();
+
+			Assert.That(shape.DistanceFromEdge(new WVec(100, 100, 0)).Length,
+				Is.EqualTo(0));
+
+			Assert.That(shape.DistanceFromEdge(new WVec(1000, 0, 0)).Length,
+				Is.EqualTo(0));
+
+			Assert.That(shape.DistanceFromEdge(new WVec(2000, 2000, 0)).Length,
+				Is.InRange(1593, 1595));
```
pchote: Why not `Is.Equals(1594)` and so on everywhere else?

pchote:
Code looks good to me aside from those remaining test nits.  I haven't had a chance to test this ingame.

OpenRA.Test/OpenRA.Game/ShapeTest.cs
huwpascoe: The result is fractional so there's going to be a margin of error with these integer based maths.

pchote: It is deterministic with well defined rules: you will always get the same result.

OpenRA.Test/OpenRA.Test.csproj
huwpascoe: I can't reproduce in windows

pchote: Windows doesn't use the makefile.

huwpascoe:
I've made changes (although I can't test the makefile works)

reaperrr:
Needs a rebase.

pchote:
Unit tests now work.

huwpascoe:
Rebased

OpenRA.Test/OpenRA.Mods.Common/ShapeTest.cs
```diff
@@ -0,0 +1,95 @@
...
+			Assert.That(shape.DistanceFromEdge(new WVec(2000, 2000, 0)).Length,
+				Is.InRange(1593, 1595));
+
+			Assert.That(new CircleShape(new WDist(73))
+				.DistanceFromEdge(new WVec(150, -100, 0)).Length,
+				Is.InRange(106, 108));
+
+			Assert.That(new CircleShape(new WDist(55))
+				.DistanceFromEdge(new WVec(30, -45, 0)).Length,
+				Is.InRange(-1, 1));
```
RoosterDragon: I swear there was an earlier comment about this, but I agree that these tests should specify exact values and not ranges. The output must be deterministic or else we'll get desyncs.

RoosterDragon:
Do we want to tackle actually applying this anywhere in this PR, or is somebody going to follow up with some changes?

Also, needs a rebase.

reaperrr:
> Do we want to tackle actually applying this anywhere in this PR, or is somebody going to follow up with some changes?

I wouldn't mind doing a follow-up if nobody else volunteers. Using rectangular hitboxes for buildings, for example, will likely not only result in a large yaml diff, but also have an effect on balance, so I believe it would be better to tackle that in a separate PR (or multiple ones).

huwpascoe:
Refactored for the new RgbaColorRenderer.

This PR is for the bounty of "Add support for rectangular hitboxes" which I've delivered. Rule changes are another issue entirely. : V

reaperrr:
This PR would currently break any custom health radius (it actually does for infantry on all mods) and needs an upgrade rule that checks whether a custom `Health.Radius` is defined in yaml rules, and if yes, replaces `Radius: xx` with `Shape: Circle` -> `Radius: xx`.

Luckily for you, I had nothing better to do ;)
https://github.com/reaperrr/OpenRA/tree/hitshape-ur

Just pick the last commit.

Other than that and the minor rectangle overlay issue, everything seems to work fine in-game.
:+0.5: for ingame from me once those are adressed, since this is a bit out of my league in terms of reviewing code.

OpenRA.Mods.Common/HitShapes/Rectangle.cs
```diff
@@ -0,0 +1,81 @@
...
+			TopLeft = tl;
+			BottomRight = br;
+		}
+
+		public void Initialize()
+		{
+			if (TopLeft.X >= BottomRight.X || TopLeft.Y >= BottomRight.Y)
+				throw new YamlException("TopLeft and BottomRight points are invalid.");
+
+			size = (BottomRight - TopLeft) / 2;
```
atlimit8: nit: This is a quadrant size rather than the actual dimensions. Perhaps `quadrantSize`.

huwpascoe:
@reaperrr Thanks, that saved me a lot of work

Nits picked and bugs fixed.

RoosterDragon:
:+1: Code looks fine. Debug visuals seem fine. Works fine in TS's isometric perspective. Damage is being dealt correctly in accordance with the debug visuals. Tests look fine and also pass.

I have no problem leaving actual rule changes to a later PR.

Revised YAML from the opening post for anybody else doing a test run:
```
MCV:
	Health:
		HP: 600
		Shape: Capsule
            Radius: 426
            PointA: 0, -512
			PointB: 0, 512
```

```
4TNK:
	Health:
		HP: 900
		Shape: Rectangle
            TopLeft: -512, -512
			BottomRight: 512, 512
```

```
AFLD:
	Health:
		HP: 1000
		Shape: Rectangle
            TopLeft: -1280, -384
			BottomRight: 1280, 1024
```

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/349ef44f49580951edbc855c4827104149b11c6d...61ee467173d19078eb221d21c9a15ba167f9ad89)

pchote:
Unfortunately this broke our unit tests #10325.

PR #9801, opened 2015-10-29
Display team colors on radar.

ghost:
Adds an option to display actors on radar in colors denoting their diplomatic stance toward the player.
team-radar

This means players who use colors similar to the terrain are on equal terms with players who opt for bright contrasting colours.

Currently hotkey is 'm'.

### DIFF
--- a/OpenRA.Game/Player.cs
+++ b/OpenRA.Game/Player.cs
@@ -10,6 +10,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Drawing;
 using System.Linq;
 using Eluant;
 using Eluant.ObjectBinding;
@@ -179,6 +180,24 @@ namespace OpenRA
 
 				return false;
 			}
+		}
+
+		public Color PlayerStanceColor(Actor a, bool isForSelectionBar = false)
+		{
+			var player = a.World.RenderPlayer ?? a.World.LocalPlayer;
+			if (player != null && !player.Spectating)
+			{
+				if (a.Owner == player)
+					return System.Drawing.Color.LimeGreen;
+
+				if (a.Owner.IsAlliedWith(player))
+					return System.Drawing.Color.Yellow;
+
+				if (!a.Owner.NonCombatant)
+					return System.Drawing.Color.Red;
+			}
+
+			return isForSelectionBar ? System.Drawing.Color.Tan : System.Drawing.Color.WhiteSmoke;
 		}
 
 		#region Scripting interface
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -168,7 +168,7 @@ namespace OpenRA
 
 		public bool UseClassicMouseStyle = false;
 		public StatusBarsType StatusBars = StatusBarsType.Standard;
-		public bool TeamHealthColors = false;
+		public bool UsePlayerStanceColors = false;
 		public bool DrawTargetLine = true;
 
 		public bool AllowDownloading = true;
@@ -213,6 +213,7 @@ namespace OpenRA
 
 		public Hotkey CycleStatusBarsKey = new Hotkey(Keycode.COMMA, Modifiers.None);
 		public Hotkey TogglePixelDoubleKey = new Hotkey(Keycode.PERIOD, Modifiers.None);
+		public Hotkey TogglePlayerStanceColorsKey = new Hotkey(Keycode.COMMA, Modifiers.Ctrl);
 
 		public Hotkey DevReloadChromeKey = new Hotkey(Keycode.C, Modifiers.Ctrl | Modifiers.Shift);
 		public Hotkey HideUserInterfaceKey = new Hotkey(Keycode.H, Modifiers.Ctrl | Modifiers.Shift);
--- a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
+++ b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
@@ -304,6 +304,8 @@ namespace OpenRA.Widgets
 					return TogglePixelDouble();
 				else if (key == Game.Settings.Keys.ToggleMuteKey)
 					return ToggleMute();
+				else if (key == Game.Settings.Keys.TogglePlayerStanceColorsKey)
+					return TogglePlayerStanceColors();
 			}
 
 			return false;
@@ -360,6 +362,7 @@ namespace OpenRA.Widgets
 		{
 			Game.Settings.Graphics.PixelDouble ^= true;
 			worldRenderer.Viewport.Zoom = Game.Settings.Graphics.PixelDouble ? 2 : 1;
+
 			return true;
 		}
 
@@ -377,6 +380,13 @@ namespace OpenRA.Widgets
 				Game.Sound.UnmuteAudio();
 				Game.Debug("Audio unmuted");
 			}
+
+			return true;
+		}
+
+		bool TogglePlayerStanceColors()
+		{
+			Game.Settings.Game.UsePlayerStanceColors ^= true;
 
 			return true;
 		}
--- a/OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
+++ b/OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
@@ -40,7 +40,9 @@ namespace OpenRA.Mods.Common.Traits
 
 		public IEnumerable<Pair<CPos, Color>> RadarSignatureCells(Actor self)
 		{
-			var color = modifier != null ? modifier.RadarColorOverride(self) : self.Owner.Color.RGB;
+			var color = Game.Settings.Game.UsePlayerStanceColors ? self.Owner.PlayerStanceColor(self) : self.Owner.Color.RGB;
+			if (modifier != null)
+				color = modifier.RadarColorOverride(self, color);
 
 			if (info.UseLocation)
 				return new[] { Pair.New(self.Location, color) };
--- a/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
@@ -43,7 +43,14 @@ namespace OpenRA.Mods.Common.Widgets
 			{
 				var time = WidgetUtils.FormatTime(p.RemainingTime, false, world.Timestep);
 				var text = Format.F(p.Info.Description, time);
-				var color = !p.Ready || Game.LocalTick % 50 < 25 ? p.Instances[0].Self.Owner.Color.RGB : Color.White;
+				var self = p.Instances[0].Self;
+				var playerColor = self.Owner.Color.RGB;
+
+				if (Game.Settings.Game.UsePlayerStanceColors)
+					playerColor = self.Owner.PlayerStanceColor(self);
+
+				var color = !p.Ready || Game.LocalTick % 50 < 25 ? playerColor : Color.White;
+
 				return Pair.New(text, color);
 			}).ToArray();
 		}
--- a/OpenRA.Mods.RA/Traits/Disguise.cs
+++ b/OpenRA.Mods.RA/Traits/Disguise.cs
@@ -120,12 +120,12 @@ namespace OpenRA.Mods.RA.Traits
 			return order.OrderString == "Disguise" ? info.Voice : null;
 		}
 
-		public Color RadarColorOverride(Actor self)
+		public Color RadarColorOverride(Actor self, Color color)
 		{
 			if (!Disguised || self.Owner.IsAlliedWith(self.World.RenderPlayer))
-				return self.Owner.Color.RGB;
+				return color;
 
-			return AsPlayer.Color.RGB;
+			return color = Game.Settings.Game.UsePlayerStanceColors ? AsPlayer.PlayerStanceColor(self) : AsPlayer.Color.RGB;
 		}
 
 		public void DisguiseAs(Actor target)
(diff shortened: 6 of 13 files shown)


OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
```diff
@@ -40,7 +40,22 @@ public void Created(Actor self)
 
 		public IEnumerable<Pair<CPos, Color>> RadarSignatureCells(Actor self)
 		{
-			var color = modifier != null ? modifier.RadarColorOverride(self) : self.Owner.Color.RGB;
+			var player = self.World.RenderPlayer ?? self.World.LocalPlayer;
+			var color = self.Owner.Color.RGB;
+
+			if (Game.Settings.Game.AllianceDisplay)
+				if (self.Owner.PlayerName != "Neutral")
```
pchote: Not a good idea to compare player names.  Check out the health bars or other parts of the UI to see how to compare against stances instead.

pchote: There's also a flag for non-combatant players (on Player), which might be what you are after instead.

OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
```diff
@@ -40,7 +40,22 @@ public void Created(Actor self)
...
+				{
```
pchote: Style nit: avoid hanging blocks.  If you can't combine these nested conditionals, you should add braces around the outer block too.

r34ch:
@pchote thanks, updated.

OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
```diff
@@ -317,5 +319,11 @@ bool TogglePixelDouble()
 			worldRenderer.Viewport.Zoom = Game.Settings.Graphics.PixelDouble ? 2 : 1;
 			return true;
 		}
+
+		bool ToggleAllianceDisplay()
```
Phrohdoh: Why is this a `bool`?

penev92: I guess `HandleKeyPress()` above needs to return a `bool`.

OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
```diff
@@ -40,7 +40,17 @@ public void Created(Actor self)
...
 		public IEnumerable<Pair<CPos, Color>> RadarSignatureCells(Actor self)
 		{
-			var color = modifier != null ? modifier.RadarColorOverride(self) : self.Owner.Color.RGB;
+			var player = self.World.RenderPlayer ?? self.World.LocalPlayer;
+			var color = self.Owner.Color.RGB;
+
+			if (Game.Settings.Game.AllianceDisplay && !self.Owner.NonCombatant)
+			{
+				color = self.Owner == player ? Color.LimeGreen :
+					self.Owner.IsAlliedWith(player) ? Color.Yellow : Color.Red;
```
Phrohdoh: This should check EffectiveOwner I bet.

Phrohdoh:
This doesn't work for me:
screen shot 2015-11-01 at 2 02 18 pm
screen shot 2015-11-01 at 2 02 21 pm

OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
cjshmyr: Agreed, may want to use similar logic to the health bar coloring if it's meant to behave exact: https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Graphics/SelectionBarsRenderable.cs#L77

Phrohdoh: If we do use that we should probably pull it out into a utility/helper class.

r34ch:
Updated. Now works in TS.

Will need testing with spies as civilians and cloaking I would guess.

OpenRA.Game/Settings.cs
```diff
@@ -171,7 +171,7 @@ public class GameSettings
 
 		public bool UseClassicMouseStyle = false;
 		public bool AlwaysShowStatusBars = false;
-		public bool TeamHealthColors = false;
+		public bool AllianceDisplay = false;
```
pchote: How about `UseAllianceColors` with `ToggleAllianceColorsKey` for the hotkey?  This makes it a bit more self descriptive.

OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
```diff
@@ -40,7 +40,31 @@ public void Created(Actor self)
 
 		public IEnumerable<Pair<CPos, Color>> RadarSignatureCells(Actor self)
 		{
-			var color = modifier != null ? modifier.RadarColorOverride(self) : self.Owner.Color.RGB;
+			var player = self.World.RenderPlayer ?? self.World.LocalPlayer;
+			var color = self.Owner.Color.RGB;
+
+			if (modifier != null)
+				color = modifier.RadarColorOverride(self);
```
pchote: I personally preferred the single-line definition of `color`.

OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
```diff
@@ -40,7 +40,31 @@ public void Created(Actor self)
...
+
+			if (Game.Settings.Game.AllianceDisplay && player != null && !player.Spectating)
```
pchote: Can you define a helper method instead of duplicating this code between here and the health bars?

pchote:
This works nicely ingame :)  Because this is simply copying (and hopefully soon reusing) the same logic as the team health bars I don't think we need to be too worried about regressions.

OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
r34ch: @pchote where should I put this helper method so it can be shared across here and health bars?

pchote: Hmm... how about on Player itself?

r34ch:
Updated and rebased. Now color codes the support weapon timers too.

OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
```diff
@@ -42,6 +42,14 @@ public void Created(Actor self)
 		{
 			var color = modifier != null ? modifier.RadarColorOverride(self) : self.Owner.Color.RGB;
 
+			if (Game.Settings.Game.UsePlayerStanceColors)
+			{
+				color = self.Owner.PlayerStanceColor(self, modifier);
+
+				if (color == Color.Tan)
```
pchote: Eww :(

pchote: Could we check against the player stance (maybe via an out parameter?) instead of checking against a magic value?

OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
```diff
@@ -42,6 +42,14 @@ public void Created(Actor self)
 		{
 			var color = modifier != null ? modifier.RadarColorOverride(self) : self.Owner.Color.RGB;
```
pchote: It would be better if we passed the color (either owner or team) into the modifier and then let these traits decide how best to deal with it.  The cloak trait could then apply the translucency, and the terrain color trait could check against UsePlayerStanceColors and do nothing if its set.

OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
pchote: Handling the modifier as suggested above may also make this workaround unnecessary.  The civilian actors that don't implement RadarColorFromTerrain probably *do* want to be shown as tan.

OpenRA.Mods.Common/Traits/AppearsOnRadar.cs
r34ch: @pchote Can you expand on this a little more please? I am failing to see how I can pass color to the modifier.

Also would I also need to check against UsePlayerStanceColors in disguise.cs in this case?

pchote: I would change the method signature on the interfaces so that you can use this as:
```csharp
var color = Game.Settings.Game.UsePlayerStanceColors ? self.Owner.PlayerStanceColor(self) : self.Owner.Color.RGB;
if (modifier != null)
    color = modifier.RadarColorOverride(self, color);
```

r34ch:
I have re-factored this following @pchote 's advice. Still unsure of a clean way of making civilians (their shade of) white on the radar while keeping the nicer tan health bars.

As a bonus it looks like the foundations have been laid for extending this further for 'economy/tactical radar' (black terrain, dark blue water, coloured resources) which would solve the color blur of red enemies on desert maps.

r34ch:
Rebased. 

If people don't mind the civilian tan health bars being changed to white when player stance colors are enabled this should be ready to merge.

RoosterDragon:
Needs another rebase.

r34ch:
Rebased.

r34ch:
Rebased.

Added a way to have tan health bars for civilian actors. Not sure about the implementation.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -43,7 +43,13 @@ public override void Tick()
 			{
 				var time = WidgetUtils.FormatTime(p.RemainingTime, false, world.Timestep);
 				var text = Format.F(p.Info.Description, time);
-				var color = !p.Ready || Game.LocalTick % 50 < 25 ? p.Instances[0].Self.Owner.Color.RGB : Color.White;
+				var playerColor = p.Instances[0].Self.Owner.Color.RGB;
```
RoosterDragon: `var self = p.Instances[0].Self;` might be useful in a few places.

OpenRA.Game/Player.cs
```diff
@@ -181,6 +182,24 @@ public bool HasFogVisibility
 			}
 		}
 
+		public Color PlayerStanceColor(Actor a, bool tan = false)
```
RoosterDragon: I would suggest renaming `tan` to `isForSelectionBar` or something like that, to indicate that the intention is to tweak the colour for the bar.

RoosterDragon:
FWIW I think the white bar is acceptable if you do wish to remove the override code to fall back to tan. I'll leave that up to you.

:+1: Looks good to me.

r34ch:
Fixed up suggestions from @RoosterDragon, kept tan health bars for now.

pchote:
Why not make neutral actors tan on the minimap too?  Looks fine when I tested this, and it removes the need for a special case.

pchote:
We probably want to extract the ally/neutral/enemy color definitions into ChromeMetrics so that mods can change them, and as a workaround for colorblind players.

r34ch:
@pchote Something like this within metrics.yaml? http://pastebin.com/fp84ZpRU

I have noticed that lots of other places in each mod use `System.Drawing.Color` for various status', so I assume eventually they will use a similar approach (metrics) to cater for colour-blind players? 

With this in mind, what would be a sensible naming convention going forward? Should they be named as generic colors, or should I keep to status names? I'm just thinking of things like ping, building health, power status etc etc

pchote:
Indeed.  The metrics is really just a stop-gap for now so that mods or players can at least adjust it themselves.  In the future I imagine we will replace the metrics setup with something a bit better.

OpenRA.Game/Settings.cs
```diff
@@ -213,6 +213,7 @@ public class KeySettings
 
 		public Hotkey CycleStatusBarsKey = new Hotkey(Keycode.COMMA, Modifiers.None);
 		public Hotkey TogglePixelDoubleKey = new Hotkey(Keycode.PERIOD, Modifiers.None);
+		public Hotkey TogglePlayerStanceColorsKey = new Hotkey(Keycode.COMMA, Modifiers.Ctrl);
```
obrakmann: `,` is already used, two lines above your change.

MustaphaTR: This is `CTRL+,` one obe is just `,` isn't it.

OpenRA.Mods.RA/Traits/Disguise.cs
```diff
@@ -120,12 +120,12 @@ public string VoicePhraseForOrder(Actor self, Order order)
...
 
-		public Color RadarColorOverride(Actor self)
+		public Color RadarColorOverride(Actor self, Color color)
 		{
 			if (!Disguised || self.Owner.IsAlliedWith(self.World.RenderPlayer))
-				return self.Owner.Color.RGB;
+				return color;
 
-			return AsPlayer.Color.RGB;
+			return color = Game.Settings.Game.UsePlayerStanceColors ? AsPlayer.PlayerStanceColor(self) : AsPlayer.Color.RGB;
```
obrakmann: This doesn't work, because you dropped the whole `apparentOwner` stuff when you moved most of `SelectionBarsRenderable.GetHealthColor` to `Player.PlayerStanceColor`.

OpenRA.Game/Settings.cs
obrakmann: Whoops, you're right.

obrakmann:
Superseded by #10649.

r34ch:
@obrakmann Thanks for not letting this rot while I was busy

Issue #5841, opened 2014-07-06
Rocket infantry stays out of firing range


reaperrr:
I noticed that when you order a TD rocket trooper to fire on te ground outside of its weapon range, it will only move closer but not start firing. Looks like it stays about 1 cell out of weapon range.
##

obrakmann:
I can't reproduce this on current bleed. @reaperrr, can you please check again?

reaperrr:
Can't reproduce it anymore and haven't noticed it recently, either.

reaperrr:
Actually just witnessed this again, still not fixed.

reaperrr:
![example1](https://cloud.githubusercontent.com/assets/2857877/10865081/bdb4ce34-8000-11e5-9a5c-c2b4a0c2db15.png)

The bazooka inf in the lower-right corner was ordered to shoot at the same cell as the rest, but neither shoots nor moves closer.

RoosterDragon:
@reaperrr has diagnosed the cause of the problem - https://github.com/OpenRA/OpenRA/pull/9802#issuecomment-152755390. If the unit is in a subcell that is too far away it is still out of range - however `MoveWithinRange` doesn't care because it's still on a cell that's within range so it doesn't move it any closer. If the unit was on a closer subcell it would be ok.

reaperrr:
It would probably help if we had something like 'FindSubCellsInAnnulus'.

GraionDilach:
I am worried if the random targetoffsets will increase the exposition of this one though.

reaperrr:
Probably, but it's something we should have fixed sooner anyway. Moving to Next milestone.

pchote:
We already used targetablepositions for range calculations, and I wouldn't expect the small position tweaks to change things here *that* much.

PR #9802, opened 2015-10-29
Fix MoveWithinRange not moving far enough for min range checks

### LINKED ISSUE #9789: Units with minimum range can't find their way to shoot.
Version: release-20150919

Steps to reproduce: Open an RA game, and build an artillery or V2 launcher. Place it directly next to an enemy building and order it to attack.

Expected behaviour: Unit will move away from the building until it is in range, and then attack.

Observed behaviour: Unit will move side to side a bunch but never attack.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27755657-units-with-minimum-range-can-t-find-their-way-to-shoot?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RoosterDragon:
MoveWithinRange was only moving the unit far enough away so the center of the target was above the minimum distance. However, the min range checks in the attack code require all positions on the target to be above the minimum distance. For large targets (e.g. buildings) this means some of the target was still too close, and the unit would get stuck in a loop.

Now MoveWithinRange uses the same range checks in order to ensure units are moved the correct distances.

Fixes #9789.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Move/MoveWithinRange.cs
+++ b/OpenRA.Mods.Common/Activities/Move/MoveWithinRange.cs
@@ -32,14 +32,12 @@ namespace OpenRA.Mods.Common.Activities
 		{
 			// We are now in range. Don't move any further!
 			// HACK: This works around the pathfinder not returning the shortest path
-			var cp = self.CenterPosition;
-			return Target.IsInRange(cp, maxRange) && !Target.IsInRange(cp, minRange);
+			return AtCorrectRange(self.CenterPosition);
 		}
 
 		protected override bool ShouldRepath(Actor self, CPos oldTargetPosition)
 		{
-			var cp = self.CenterPosition;
-			return targetPosition != oldTargetPosition && (!Target.IsInRange(cp, maxRange) || Target.IsInRange(cp, minRange));
+			return targetPosition != oldTargetPosition && !AtCorrectRange(self.CenterPosition);
 		}
 
 		protected override IEnumerable<CPos> CandidateMovementCells(Actor self)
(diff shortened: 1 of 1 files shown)


scshunt:
AtCorrectRange should be private, no?

This does not fully fix #9789. There is still a way to reproduce if there are buildings in the way.

scshunt:
My bad, it does appear to fix it. For some reason the changes weren't being reflected in my game...

RichSuther:
Works well, in open ground takes a nice path away and takes the shot.

Made a little 'assault course' to test, the V2 takes the path round then fires (red arrow)
ok

RoosterDragon:
@scshunt: OpenRA code style is to omit the default accessibility modifiers - so I have not put an explicit `private` since that is already the case.

scshunt:
Ahh, ok. Thanks!

OpenRA.Mods.Common/Activities/Move/MoveWithinRange.cs
```diff
@@ -48,15 +46,13 @@ protected override IEnumerable<CPos> CandidateMovementCells(Actor self)
...
+		}
 
-			return map.FindTilesInAnnulus(targetPosition, minCells + 1, maxCells).Where(c =>
-			{
-				var dxSq = (map.CenterOfCell(c) - center).HorizontalLengthSquared;
-				return dxSq >= innerSq && dxSq <= outerSq;
-			});
+		bool AtCorrectRange(WPos origin)
+		{
+			return Target.IsInRange(origin, maxRange) && !Target.IsInRange(origin, minRange);
```
atlimit8: `minRange` should be `minRange + new WDist(1)`

RoosterDragon: The rest of the targeting code doesn't make such an adjustment - so that would break everything again because we could end up too close.

atlimit8: Sadly that means `minRange` is not min.

reaperrr:
@RoosterDragon while you're at it: 
Sometimes infantry on the furthest-away subcell neither starts shooting nor moves closer to the target, probably because the `WDist` between subcell and `target.CenterPosition` is slightly too large for the weapon to shoot, but the `MoveWithinRange` code thinks it's close enough in terms of cells.

RoosterDragon:
@reaperrr: Your diagnoses is correct - however I don't know the best way to go about fixing that one.

Mailaender:
Can't spot regressions :white_check_mark: and the code looks cleaner. :+1:

obrakmann:
I can still quite reliably reproduce #9789 on this branch. [Here's a replay.](https://seacloud.cc/d/5d1adbebce/files/?p=/ora-pr9802.orarep)

RoosterDragon:
Your replay shows #5841 clearly - which this PR indeed does not fix.

This PR fixes the range checks at the cell level which is sufficient to fix #9789 for units like the artillery - but the subcell level is still broken (for which #5841 remains open).

obrakmann:
:+1: Now that I was looking for the correct thing, this does indeed fix what it says it fixes. Sorry for the confusion on my part earlier.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7c7e964452d6e3881a83a126db6686da49bac571)

PR #9808, opened 2015-10-30
Several fixes for cloak behavior

teees:
- Reworked the cloak class
  - Removed code duplication when granting upgrades
  - Use explicit interface implementations where possible
  - Added optional `UncloakOnDamage` which uncloaks actors when at least one tick damage taken
  - Play sound when being cloaked by an upgrade
  - Don't play sounds when cloaked by another cloaking trait
  - Uncloak actors when `UncloakOnAttack==true` and they are charging
  - Refactored all loose `UncloakOn*` booleans and merged them into one enum
- Extended/Reworked TS cloaking behavior
  - Cloak Aircraft via cloaking generator
  - Uncloak actors and buildings on damage taken
  - Added Sounds for cloaking

### DIFF
--- a/OpenRA.Mods.Common/Activities/UnloadCargo.cs
+++ b/OpenRA.Mods.Common/Activities/UnloadCargo.cs
@@ -59,7 +59,7 @@ namespace OpenRA.Mods.Common.Activities
 			if (IsCanceled || cargo.IsEmpty(self))
 				return NextActivity;
 
-			if (cloak != null && cloak.Info.UncloakOnUnload)
+			if (cloak != null && cloak.Info.UncloakOn.HasFlag(UncloakType.Unload))
 				cloak.Uncloak();
 
 			var actor = cargo.Peek(self);
--- a/OpenRA.Mods.Common/Traits/Cloak.cs
+++ b/OpenRA.Mods.Common/Traits/Cloak.cs
@@ -17,6 +17,18 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
+	[Flags]
+	public enum UncloakType
+	{
+		None = 0,
+		Attack = 1,
+		Move = 2,
+		Unload = 4,
+		Infiltrate = 8,
+		Demolish = 16,
+		Damage = 32
+	}
+
 	[Desc("This unit can cloak and uncloak in specific situations.")]
 	public class CloakInfo : UpgradableTraitInfo
 	{
@@ -26,11 +38,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Measured in game ticks.")]
 		public readonly int CloakDelay = 30;
 
-		public readonly bool UncloakOnAttack = true;
-		public readonly bool UncloakOnMove = false;
-		public readonly bool UncloakOnUnload = true;
-		public readonly bool UncloakOnInfiltrate = true;
-		public readonly bool UncloakOnDemolish = true;
+		[Desc("Events leading to the actor getting uncloaked. Possible values are: Attack, Move, Unload, Infiltrate, Demolish and Damage")]
+		public readonly UncloakType UncloakOn = UncloakType.Attack
+			| UncloakType.Unload | UncloakType.Infiltrate | UncloakType.Demolish;
 
 		public readonly string CloakSound = null;
 		public readonly string UncloakSound = null;
@@ -44,70 +54,48 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("The upgrades to grant to self while cloaked.")]
 		public readonly string[] WhileCloakedUpgrades = { };
 
-		public override object Create(ActorInitializer init) { return new Cloak(init.Self, this); }
+		public override object Create(ActorInitializer init) { return new Cloak(this); }
 	}
 
-	public class Cloak : UpgradableTrait<CloakInfo>, IRenderModifier, INotifyDamageStateChanged, INotifyAttack, ITick, IVisibilityModifier, IRadarColorModifier, INotifyCreated
+	public class Cloak : UpgradableTrait<CloakInfo>, IRenderModifier, INotifyDamageStateChanged,
+		INotifyAttack, ITick, IVisibilityModifier, IRadarColorModifier, INotifyCreated
 	{
 		[Sync] int remainingTime;
 		[Sync] bool damageDisabled;
 		UpgradeManager upgradeManager;
 
-		Actor self;
 		CPos? lastPos;
+		bool wasCloaked = false;
 
-		public Cloak(Actor self, CloakInfo info)
+		public Cloak(CloakInfo info)
 			: base(info)
 		{
-			this.self = self;
-
 			remainingTime = info.InitialDelay;
 		}
 
-		public void Created(Actor self)
+		void INotifyCreated.Created(Actor self)
 		{
 			upgradeManager = self.TraitOrDefault<UpgradeManager>();
-			if (remainingTime == 0)
-			{
-				if (upgradeManager != null)
-					foreach (var u in Info.WhileCloakedUpgrades)
-						upgradeManager.GrantUpgrade(self, u, this);
-			}
-		}
-
-		protected override void UpgradeDisabled(Actor self)
-		{
-			Uncloak();
-			remainingTime = Info.InitialDelay;
-		}
-
-		public void Uncloak() { Uncloak(Info.CloakDelay); }
-
-		public void Uncloak(int time)
-		{
 			if (Cloaked)
-			{
-				Game.Sound.Play(Info.UncloakSound, self.CenterPosition);
-				if (upgradeManager != null)
-					foreach (var u in Info.WhileCloakedUpgrades)
-						upgradeManager.RevokeUpgrade(self, u, this);
-			}
-
-			remainingTime = Math.Max(remainingTime, time);
+				GrantUpgrades(self);
 		}
-
-		public void Attacking(Actor self, Target target, Armament a, Barrel barrel) { if (Info.UncloakOnAttack) Uncloak(); }
 
 		public bool Cloaked { get { return !IsTraitDisabled && remainingTime <= 0; } }
 
-		public void DamageStateChanged(Actor self, AttackInfo e)
+		public void Uncloak() { Uncloak(Info.CloakDelay); }
+
+		public void Uncloak(int time) { remainingTime = Math.Max(remainingTime, time); }
+
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel) { if (Info.UncloakOn.HasFlag(UncloakType.Attack)) Uncloak(); }
+
+		void INotifyDamageStateChanged.DamageStateChanged(Actor self, AttackInfo e)
 		{
 			damageDisabled = e.DamageState >= DamageState.Critical;
-			if (damageDisabled)
+			if (damageDisabled || Info.UncloakOn.HasFlag(UncloakType.Damage))
 				Uncloak();
 		}
 
-		public IEnumerable<IRenderable> ModifyRender(Actor self, WorldRenderer wr, IEnumerable<IRenderable> r)
+		IEnumerable<IRenderable> IRenderModifier.ModifyRender(Actor self, WorldRenderer wr, IEnumerable<IRenderable> r)
 		{
 			if (remainingTime > 0 || IsTraitDisabled)
 				return r;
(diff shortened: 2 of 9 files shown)


teees:
I know, this makes them cloakable by stealth gen, not by themselves

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -31,6 +31,7 @@ public class CloakInfo : UpgradableTraitInfo
 		public readonly bool UncloakOnUnload = true;
 		public readonly bool UncloakOnInfiltrate = true;
 		public readonly bool UncloakOnDemolish = true;
+		public readonly bool UncloakOnDamaged = true;
```
pchote: This should be `UncloakOnDamage` to match the tense of the other options.

teees: thank you, changed

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -77,10 +80,30 @@ public void Created(Actor self)
...
+			}
+
 			remainingTime = Info.InitialDelay;
 		}
 
+		protected override void UpgradeEnabled(Actor self)
+		{
+			if (IsTraitDisabled != wasDisabled)
+			{
+				Game.Sound.Play(Info.CloakSound, self.CenterPosition);
```
pchote: Won't this make the sound play even if the InitialDelay means it doesn't actually cloak?

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -77,10 +80,30 @@ public void Created(Actor self)
...
+				if (upgradeManager != null)
+					foreach (var u in Info.WhileCloakedUpgrades)
+						upgradeManager.RevokeUpgrade(self, u, this);
```
pchote: Shouldn't this be *granting* upgrades?

pchote: Actually... i'm not sure this should be doing anything, since the actual cloaking should be happening later.

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -77,10 +80,30 @@ public void Created(Actor self)
 
 		protected override void UpgradeDisabled(Actor self)
 		{
-			Uncloak();
+			if (IsTraitDisabled != wasDisabled)
+			{
+				Game.Sound.Play(Info.UncloakSound, self.CenterPosition);
```
pchote: Can you keep this calling `Uncloak` to avoid the code duplication?

teees:
I am currently reworking the class to be more readable and with less code duplication

pchote:
Just a thought (at the risk of making more work): we could merge all these flags into a single enum / bitfield, and then define e.g. `UncloakOn: Damage, Attack, Unload`

teees:
seems like a good idea, I'll look into that

teees:
Done, also added a `Default` enum for convenience

OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -126,4 +126,17 @@ public interface IPreventsAutoTarget
...
+	public enum UncloakType
+	{
+		None = 0,
+		Attack = 1,
+		Move = 2,
+		Unload = 4,
+		Infiltrate = 8,
+		Demolish = 16,
+		Damage = 32,
+		Default = Attack | Unload | Infiltrate | Demolish
```
pchote: This is actually less convenient: if you leave it explicit in the trait the generated docs will list all the types, while 'default' is opaque to modders.

OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -126,4 +126,17 @@ public interface IPreventsAutoTarget
 	{
 		bool PreventsAutoTarget(Actor self, Actor attacker);
 	}
+
+	[Flags]
+	public enum UncloakType
```
pchote: Might be better to keep this in cloak.cs

OpenRA.Mods.Common/TraitsInterfaces.cs
Phrohdoh: This line inspired me to write http://pastebin.com/raw.php?i=vQEKvGYV for situations where this may exist / be merged in the future.

pchote: Nice!  We can build that into the docs generation and then remove the (IMO ugly) cases where we list the enum values in the [Desc].

Phrohdoh: Agreed. :)

OpenRA.Mods.Common/TraitsInterfaces.cs
teees: When I placed it there, I used `ImpactType` as a reference. I am a bit confused on which should go where. Will put it back to Cloak.cs as initially dont

OpenRA.Mods.Common/Activities/UnloadCargo.cs
```diff
@@ -59,7 +59,7 @@ public override Activity Tick(Actor self)
 			if (IsCanceled || cargo.IsEmpty(self))
 				return NextActivity;
 
-			if (cloak != null && cloak.Info.UncloakOnUnload)
+			if (cloak != null && cloak.Info.UncloakOn.HasFlag(UncloakType.Demolish))
```
abcdefg30: This wants to be `UncloakType.Unload` instead of `Demolish`, I guess.

teees: Fixed, thanks

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -47,67 +57,45 @@ public class CloakInfo : UpgradableTraitInfo
 		public override object Create(ActorInitializer init) { return new Cloak(init.Self, this); }
```
abcdefg30: Do we still need the `init.Self` parameter?

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -124,27 +112,38 @@ public IEnumerable<IRenderable> ModifyRender(Actor self, WorldRenderer wr, IEnum
...
 
-			if (Info.UncloakOnMove && (lastPos == null || lastPos.Value != self.Location))
+				if (Info.UncloakOn.HasFlag(UncloakType.Move) && (lastPos == null || lastPos.Value != self.Location))
+				{
+					Uncloak();
+					lastPos = self.Location;
+				}
+			}
+
+			bool isCloaked = Cloaked;
```
abcdefg30: var?

abcdefg30:
Won't this also need an upgrade rule?

teees:
Added the upgrade rule

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -17,6 +17,20 @@
...
+	public enum UncloakType
+	{
+		None = 0,
+		Attack = 1,
+		Move = 2,
+		Unload = 4,
+		Infiltrate = 8,
+		Demolish = 16,
+		Damage = 32,
+		[Desc("Uncloaks on Attack, Unload, Infiltrate, Demolish")]
```
Phrohdoh: `, or Demolish` please.

teees: I chose this notation on purpose, since it will look that way when you define those flags (it's a list of flags after all, not an english sentence)

Phrohdoh: I realize you chose this on purpose, but it is an english sentence because this is documentation.
As a reader I don't know if this is AND (imagine that scenario) or OR.

Common sense says it is an or but there's no reason not be a bit more helpful.

Perhaps `Uncloaks on any of ...` then?

pchote: I still think we should remove this `Default` type, and make it explicit in the trait.

Phrohdoh: I agree, but that wasn't received at all.

teees: The problem I see arising when maintaining a mod is that most of the time you will want default behavior + your additional uncloak types. If that list gets extended in the future and you have to explicitely state all the default types, you will have to carefully adapt all those units using Cloak.
Anyways, removed that pseudo flag and rebased

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2683,6 +2683,52 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
 				}
 
+				if (engineVersion < 20151214 && depth == 1 && node.Key.StartsWith("Cloak"))
+				{
+					var defaultCloakType = Traits.UncloakType.Attack
+						| Traits.UncloakType.Unload | Traits.UncloakType.Infiltrate | Traits.UncloakType.Demolish;
+
+					// Merge Uncloak types
+					var t = defaultCloakType;
+					for (int i = node.Value.Nodes.Count - 1; i >= 0; i--)
```
Phrohdoh: `var i`

abcdefg30:
Looks good to me now. :+1:

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2694,6 +2694,52 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 					}
 				}
 
+				if (engineVersion < 20151214 && depth == 1 && node.Key.StartsWith("Cloak"))
```
abcdefg30: We probably want to update this number here before merge.

Mailaender:
Needs a rebase.

teees:
rebased and changed the date

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -26,11 +38,9 @@ public class CloakInfo : UpgradableTraitInfo
 		[Desc("Measured in game ticks.")]
 		public readonly int CloakDelay = 30;
 
-		public readonly bool UncloakOnAttack = true;
-		public readonly bool UncloakOnMove = false;
-		public readonly bool UncloakOnUnload = true;
-		public readonly bool UncloakOnInfiltrate = true;
-		public readonly bool UncloakOnDemolish = true;
+		[Desc("Events leading to the actor getting uncloaked.")]
```
Mailaender: You need to document the possible values here for our auto-generated https://github.com/OpenRA/OpenRA/wiki/Traits documentation. It is inconvenient. I filed https://github.com/OpenRA/OpenRA/issues/10378.

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -44,70 +54,48 @@ public class CloakInfo : UpgradableTraitInfo
...
+	public class Cloak : UpgradableTrait<CloakInfo>, IRenderModifier, INotifyDamageStateChanged,
+		INotifyAttack, ITick, IVisibilityModifier, IRadarColorModifier, INotifyCreated
 	{
 		[Sync] int remainingTime;
 		[Sync] bool damageDisabled;
 		UpgradeManager upgradeManager;
 
-		Actor self;
 		CPos? lastPos;
+		bool wasCloaked = false;
```
Mailaender: Is this worth checking `[Sync]` as well?

OpenRA.Mods.Common/Traits/Cloak.cs
```diff
@@ -44,70 +54,48 @@ public class CloakInfo : UpgradableTraitInfo
 		[Desc("The upgrades to grant to self while cloaked.")]
 		public readonly string[] WhileCloakedUpgrades = { };
 
-		public override object Create(ActorInitializer init) { return new Cloak(init.Self, this); }
+		public override object Create(ActorInitializer init) { return new Cloak(this); }
 	}
 
-	public class Cloak : UpgradableTrait<CloakInfo>, IRenderModifier, INotifyDamageStateChanged, INotifyAttack, ITick, IVisibilityModifier, IRadarColorModifier, INotifyCreated
+	public class Cloak : UpgradableTrait<CloakInfo>, IRenderModifier, INotifyDamageStateChanged,
+		INotifyAttack, ITick, IVisibilityModifier, IRadarColorModifier, INotifyCreated
```
Mailaender: `ISync` is missing here. Not really your fault. Thought we had a lint test for it.

Mailaender:
Otherwise works as promised. :white_check_mark:

OpenRA.Mods.Common/Traits/Cloak.cs
pchote: Excessive `[Sync]`ing just hurts perf.

Mailaender: Okay, then nevermind: don't `[Sync]` it.

abcdefg30:
Also needs a rebase again.

OpenRA.Mods.Common/Traits/Cloak.cs
teees: ISync should already be present since Cloak uses UpgradableTrait as base class

teees:
Added the missing documentation, rebased and adapted date in UpgradeRules

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/14d5faeebc80a5dc439acd4b63c52fe8fd427871...c82346ddedc96113571f4954df3c7c7ac2255f17)

PR #9821, opened 2015-10-31
Increase the speed of HIJACKER

abcdefg30:
Adds the speed buff missed in #9668.

### DIFF
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -460,6 +460,8 @@ HIJACKER:
 		UncloakOnMove: true
 		CloakTypes: Cloak, Hijacker
 		IsPlayerPalette: true
+	Mobile:
+		Speed: 85
 
 SHOK:
 	Inherits: ^Soldier


MicroJOo:
:+1:

obrakmann:
Even with the speed change, hijackers still can't catch up to moving/harvesting harvesters. they get too them in the refinery, though.

MicroJOo:
should be possible to get them while they are "harvesting" ... too

reaperrr:
I'd increase it to 96, tbh, then it would be a little faster than harvs (assuming same terrain modifier).

reaperrr:
After discussing this on irc, 85 should be good enough for the playtest, we can still tweak it later if necessary. :+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/ad735cd3eec6d138d9c09086c1fb99fc0c53355c...1ca40f4b03768ae85440bcba4e5758097f82fd15)

Issue #9820, opened 2015-10-31
Stealth Detection is broken/not functional


DoGyAUT:
DetectCloaked seems not to be working at all.

It also looks like RenderDetectionCircle doesn't work too.

reaperrr:
What version and mod did you test?

Because on latest bleed, both still work in the cnc mod.

penev92:
There is no detection circle in TS on latest bleed fwiw.

DoGyAUT:
Check TS mod, Sensor doesn't have a detection circle.

My Dogs can't detect my scouts (custom mod) - it worked in older builds (early oct)

reaperrr:
You're in all likelyness affected by the same issue as the TS sensor:

`DetectCloaked`'s `Range` was recently converted from cells to world units (where `1024` or `1c0` equals 1 cell), adding `c0` to your detection range will likely fix both (as it does on TS' sensor array).

DoGyAUT:
Yes seems that was the problem!

PR #9822, opened 2015-10-31
Fix TS sensor array DetectCloaked range

reaperrr:


Phrohdoh:
Why are we not throwing on this type of issue?

GraionDilach:
Because 0c18 is technically a valid value.

penev92:
Closes #9820.
:+1:

PR #9825, opened 2015-10-31
Add infantry stealth detection to D2k

reaperrr:
So they can detect fremen, saboteurs and stealth raiders.

Note:
- the radius needs to be a bit larger than 1 cell to make sure the subcell doesn't matter that much


GraionDilach:
:+1:

PR #9827, opened 2015-10-31
Implement the TS EMPulse SW.

GraionDilach:
I'm not certain how viable is this though, because Armament requires AttackBase but this doesn't use that.

Also it needs an estimation of the projectile reaching target for camera/beacon purposes, see comment on diff.

### DIFF
--- a/OpenRA.Mods.TS/Traits/SupportPowers/FireArmamentPower.cs
+++ b/OpenRA.Mods.TS/Traits/SupportPowers/FireArmamentPower.cs
@@ -1,0 +1,220 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using OpenRA.Effects;
+using OpenRA.GameRules;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Mods.Common.Effects;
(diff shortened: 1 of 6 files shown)


OpenRA.Mods.TS/Traits/SupportPowers/FireArmamentPower.cs
```diff
@@ -0,0 +1,220 @@
...
+				Game.Sound.Play(Info.LaunchSound);
+			else
+				Game.Sound.Play(Info.IncomingSound);
+
+			target = Target.FromCell(self.World, order.TargetLocation);
+
+			enabled = true;
+
+			// TODO: Estimate the projectile travel time somehow
+			estimatedTicks = armament.Value.FireDelay;
```
GraionDilach: I thought about doing this via doing a rough estimation in the projectile constructors and then sending that back via an interface - since this is needed to be calculated at runtime, due to `Missile`/`GravityBomb` etc. having acceleration - but wouldn't that affect perf heavily?

matija-hustic: As far as I see, the EMP uses the bullet projectile and in that case, the projectile's travel duration is easily obtainable:
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Effects/Bullet.cs#L119

GraionDilach: Yes, for this particular case - but I want this to be used for the Cluster Missile SW in a followup.

matija-hustic: Ahhh, yeah then I can tell you for the missile, it's practically impossible to compute exactly... But you don't need the exact remaining time, do you? Just the progress status `FractionComplete` basically that doesn't necessarily have to decrease linearly with time? Then you could use remaining distance / total distance instead...

GraionDilach:
I'll return this when I'll have time to come up with something for the beacon.

PR #9838, opened 2015-10-31
Per-warhead WarheadDebugOverlay colors

phrohdoh:
This allows any warhead to provide its own color, such as healing warheads being green.
The test commit changes the E1 in TS to use a cyan-ish color.

Inspired by @deniz1a's comment [here](https://github.com/OpenRA/OpenRA/pull/9465#issuecomment-148508283).

Outdated gif
newer

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/WarheadDebugOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/World/WarheadDebugOverlay.cs
@@ -29,6 +29,7 @@ namespace OpenRA.Mods.Common.Traits
 		{
 			public readonly WPos CenterPosition;
 			public readonly WDist[] Range;
+			public readonly Color Color;
 			public int Time;
 
 			public WDist OuterRange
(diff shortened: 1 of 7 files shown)


GraionDilach:
This is something which would be so damn useful if weapons could inherit. Without it it's a niche, but I guess people could find use for it even.

:+1:

pchote:
Adding [Has Dependencies] so that we remember to remove the test commit before merging.

RoosterDragon:
Seems like a nice idea. I would like to see the colour adjusted for healing units adjusted as part of this PR, so we're actually putting this to use rather than just adding dead code.

Phrohdoh:
Updated.

RoosterDragon:
:+1:

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d798fa71ec89c0e75ed621edba0d350d66b70902)

PR #9842, opened 2015-11-01
Fix (HealthPercentage)DamageWarheads ignoring stances.

### LINKED ISSUE #9107: MAD tanks destroy trees
Trees are normally indestructable. They are untargetable. Nuke doesn't destroy them.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25926604-mad-tanks-destroy-trees?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

GraionDilach:
How come noone noticed this before, I wonder... target validity was never checked at these two compared to `SpreadDamageWarhead`.

Fixes  #9107.

### DIFF
--- a/OpenRA.Mods.Common/Warheads/HealthPercentageDamageWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/HealthPercentageDamageWarhead.cs
@@ -42,6 +42,9 @@ namespace OpenRA.Mods.Common.Warheads
 
 		public override void DoImpact(Actor victim, Actor firedBy, IEnumerable<int> damageModifiers)
 		{
+			if (!IsValidAgainst(victim, firedBy))
+				return;
+
 			var healthInfo = victim.Info.TraitInfoOrDefault<HealthInfo>();
 			if (healthInfo == null)
 				return;
--- a/OpenRA.Mods.Common/Warheads/DamageWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/DamageWarhead.cs
@@ -65,6 +65,9 @@ namespace OpenRA.Mods.Common.Warheads
 
 		public virtual void DoImpact(Actor victim, Actor firedBy, IEnumerable<int> damageModifiers)
 		{
+			if (!IsValidAgainst(victim, firedBy))
+				return;
+
 			var damage = Util.ApplyPercentageModifiers(Damage, damageModifiers.Append(DamageVersus(victim)));
 			victim.InflictDamage(firedBy, damage, this);
 		}
--- a/OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs
@@ -62,9 +62,6 @@ namespace OpenRA.Mods.Common.Warheads
 
 			foreach (var victim in hitActors)
 			{
-				if (!IsValidAgainst(victim, firedBy))
-					continue;
-
 				var healthInfo = victim.Info.TraitInfoOrDefault<HealthInfo>();
 				if (healthInfo == null)
 					continue;


RoosterDragon:
You can remove the check in `SpreadDamageWarhead.DoImpact` if I'm reading this right, since it will now be done by `DamageWarhead.DoImpact` instead.

GraionDilach:
You're right, I didn't noticed that `SpreadDamageWarhead` inherits `DamageWarhead` while I was debugging this.

Phrohdoh:
Isn't that also true in the case of `public class HealthPercentageDamageWarhead : DamageWarhead`?

GraionDilach:
`HealthPercentageDamageWarhead` skips `DamageWarhead.DoImpact`. `SpreadDamageWarhead` calls it at the end of it's own override.

Phrohdoh:
Okay, that makes sense.
We could avoid these computations by keeping that check. Food for thought.

Phrohdoh:
Okay this code makes sense so :+1:, but I cannot manufacture a situation to actually notice the fix.

reaperrr:
Needs rebase.

GraionDilach:
Rebased.

OpenRA.Mods.Common/Warheads/HealthPercentageDamageWarhead.cs
```diff
@@ -42,6 +42,9 @@ public override void DoImpact(WPos pos, Actor firedBy, IEnumerable<int> damageMo
 
 		public override void DoImpact(Actor victim, Actor firedBy, IEnumerable<int> damageModifiers)
 		{
+			if (!IsValidAgainst(victim, firedBy))
+				return;
```
reaperrr: This warhead also implements `DamageWarhead`, so this shouldn't be needed.

Phrohdoh: See https://github.com/OpenRA/OpenRA/pull/9842#issuecomment-152852839

GraionDilach: It does not call `DamageWarhead`'s relevant `DoImpact` and goes directly into `InflictDamage`. See https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Warheads/SpreadDamageWarhead.cs#L76 vs https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Warheads/HealthPercentageDamageWarhead.cs#L51

reaperrr: Ok, overlooked that. Nevermind then.

RoosterDragon:
The code change looks fine at this point.

I'm wondering if there are any possible negative side effects? Are there any warheads dealing damage to invalid units that we might actually want to keep? I'm not familiar enough with the game mechanics to know if there might be such a scenario.

GraionDilach:
Currently only the MAD Tank uses these kind of warhead in our mods - which even has InvalidTargets set up for it to mimic the RA1 behaviour, see https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/weapons/other.yaml#L261 - so I guess that was a bug in our case. The [For stable] notice was @penev92's idea though, I'm also unsure about that.

penev92:
From what I heard this fixes a bug that would let actors hurt targets they aren't supposed to be able to. 
>Currently only the MAD Tank uses these kind of warhead in our mods

So the only change would be that MAD tanks won't destroy stuff that they aren't supposed to any more, like trees. (Yes, people will complain. They always do. This still fixes a bug.)

RoosterDragon:
That seems reasonable to me.

Confirmed this fixes the mad tank issue, so :+1:

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/1f6d68dd372dbaa2a1b61d79f989a5b09a29e97d...53efa822f8396150c3b64f1e3be0b008b90d1d1c)

PR #9848, opened 2015-11-01
Add LoS to Bio lab and prison.

ghost:
Both of these tech buildings are missing LoS when captured.

Noticed on PizzaAtomicas Mad Science MP map.


GraionDilach:
The changes seem consistent with Hospital and Oil Derrick, so :+1:.

PR #9852, opened 2015-11-01
Detach some event handlers

### LINKED ISSUE #9841: TerrainSpriteLayer memory sink
Whilst profiling https://github.com/OpenRA/OpenRA/pull/9627, I came across this hard-to-miss memory sink. It happens each time you start another map. Here's the difference from baseline to after jumping between skirmish and title screen for a while:

OpenRA.Graphics.TerrainSpriteLayer
+114 instances
+209,198,208 bytes

Almost all of that memory is composed of OpenRA.Graphics.Vertex[] inside the TerrainSpriteLayer.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27865703-terrainspritelayer-memory-sink?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RoosterDragon:
Some classes attach event handlers to classes that live longer than they do, thus they end up extending their lifetime beyond what we normally expect because the event handler reference keeps them alive.

This leads to a few "memory leak" style scenarios where we don't reclaim memory because it still has references. Explicitly removing the event handlers when these classes are done with prevents this problem.

Fixes #9841.

### DIFF
--- a/OpenRA.Game/Graphics/TerrainSpriteLayer.cs
+++ b/OpenRA.Game/Graphics/TerrainSpriteLayer.cs
@@ -33,7 +33,7 @@ namespace OpenRA.Graphics
 		readonly WorldRenderer worldRenderer;
 		readonly Map map;
 
-		float paletteIndex;
+		readonly PaletteReference palette;
 
 		public TerrainSpriteLayer(World world, WorldRenderer wr, Sheet sheet, BlendMode blendMode, PaletteReference palette, bool restrictToBounds)
 		{
@@ -41,7 +41,7 @@ namespace OpenRA.Graphics
 			this.restrictToBounds = restrictToBounds;
 			Sheet = sheet;
 			BlendMode = blendMode;
-			paletteIndex = palette.TextureIndex;
+			this.palette = palette;
 
 			map = world.Map;
 			rowStride = 4 * map.MapSize.X;
@@ -50,21 +50,21 @@ namespace OpenRA.Graphics
 			vertexBuffer = Game.Renderer.Device.CreateVertexBuffer(vertices.Length);
 			emptySprite = new Sprite(sheet, Rectangle.Empty, TextureChannel.Alpha);
 
-			wr.PaletteInvalidated += () =>
+			wr.PaletteInvalidated += UpdatePaletteIndices;
+		}
+
+		void UpdatePaletteIndices()
+		{
+			// Everything in the layer uses the same palette,
+			// so we can fix the indices in one pass
+			for (var i = 0; i < vertices.Length; i++)
 			{
-				paletteIndex = palette.TextureIndex;
+				var v = vertices[i];
+				vertices[i] = new Vertex(v.X, v.Y, v.Z, v.U, v.V, palette.TextureIndex, v.C);
+			}
 
-				// Everything in the layer uses the same palette,
-				// so we can fix the indices in one pass
-				for (var i = 0; i < vertices.Length; i++)
-				{
-					var v = vertices[i];
-					vertices[i] = new Vertex(v.X, v.Y, v.Z, v.U, v.V, paletteIndex, v.C);
-				}
-
-				for (var row = 0; row < map.MapSize.Y; row++)
-					dirtyRows.Add(row);
-			};
+			for (var row = 0; row < map.MapSize.Y; row++)
+				dirtyRows.Add(row);
 		}
 
 		public void Update(CPos cell, Sprite sprite)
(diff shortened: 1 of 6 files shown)


OpenRA.Mods.Common/Widgets/RadarWidget.cs
```diff
@@ -87,6 +87,7 @@ public override void Initialize(WidgetArgs args)
 			foreach (var cell in world.Map.AllCells)
 				UpdateTerrainCell(cell);
 
+			// TODO: Ideally, we need to detach these event handlers when the radar widget is finished with.
```
pchote: We have `Widget.Removed()` for when a widget is removed (not just hidden) from view.  I'm not 100% sure this is used as a dispose-equivalent, but it may well be.

RoosterDragon: Exactly the magic method I needed! Updated.

huwpascoe:
:+1: Memory usage doesn't grow at all now.

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
```diff
@@ -32,6 +32,7 @@ public sealed class TerrainSpriteLayer : IDisposable
 
 		readonly WorldRenderer worldRenderer;
 		readonly Map map;
+		readonly Action changePaletteIndexes;
```
atlimit8: Is there a reason an explicit delegate is used instead of a private method?

OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
```diff
@@ -27,6 +27,7 @@ protected enum MenuType { Main, Singleplayer, Extras, MapEditor, None }
 		readonly ScrollPanelWidget newsPanel;
 		readonly Widget newsTemplate;
 		readonly LabelWidget newsStatus;
+		readonly Action<string, int> onRemoteDirectConnect;
```
atlimit8: Same as above.

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
RoosterDragon: Not really, it was just the quickest thing to do.

atlimit8:
Since `readonly Action changePaletteIndexes` and `readonly Action<string, int> onRemoteDirectConnect` are defined in those classes (inside the respective constructors), they should be replaced with `void ChangePaletteIndexes()` and `void OnRemoteDirectConnect(string host, int port)`
Otherwise, :+1:.

RoosterDragon:
Updated per comments.

huwpascoe:
:+1:

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
```diff
@@ -50,21 +50,21 @@ public TerrainSpriteLayer(World world, WorldRenderer wr, Sheet sheet, BlendMode
...
+		}
 
-				// Everything in the layer uses the same palette,
-				// so we can fix the indices in one pass
-				for (var i = 0; i < vertices.Length; i++)
-				{
-					var v = vertices[i];
-					vertices[i] = new Vertex(v.X, v.Y, v.Z, v.U, v.V, paletteIndex, v.C);
-				}
+		void ChangePaletteIndexes()
```
pchote: We use "Indices" in other parts of the code, but TIL that "Indexes" is valid / preferred in en-us.
It would be nice to keep this consistent with our other code.

pchote: Could you change this to `Update*` to match the callbacks in the other classes touched by this pr?

pchote:
Looks good to me aside from those naming nits.

RoosterDragon:
Updated naming.

OpenRA.Game/Graphics/TerrainSpriteLayer.cs
Phrohdoh: `Indices` is used at work and is more proper IMO, so another vote for that.

RoosterDragon: We use a mix in our code right now, so we may want to consistantify this at some point.

obrakmann:
Looks ok to me. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/2aa07b9b9a72849e9de44955018abaf4ec79df2e)

Issue #8992, opened 2015-08-08
Support mods loaded from the support directory


pchote:
This is another fundamental engine feature for third party mods.  Expecting players to install mods in the main game dir is dangerous, and causes trouble when we release new engine versions.  Loading mods from the support dir fixes most of these problems.

We should support both unpacked mods (in a directory) and packaged mods in a zip renamed to oramod.
Because these mods are effectively data files in a user-owned directory there are no fundamental problems with supporting in-game installation and upgrades for them.

The tricky part is dealing with file paths, as our current mod.yaml syntax defines only `.` for "relative to game exe" and `^` for "relative to support dir".  I propose introducing a `<mod id>:` prefix for referencing files relative to the current or another named mod dir.

e.g. a snippet of the d2k mod may look like:
```yaml
Assemblies:
	common: OpenRA.Mods.Common.dll
	ra: OpenRA.Mods.RA.dll
	cnc: OpenRA.Mods.Cnc.dll
	d2k: OpenRA.Mods.D2k.dll

ChromeLayout:
	d2k: chrome/install.yaml
	d2k: chrome/ingame.yaml
	ra: chrome/ingame-chat.yaml
	ra: chrome/ingame-diplomacy.yaml
	ra: chrome/ingame-fmvplayer.yaml
	d2k: chrome/ingame-menu.yaml
	ra: chrome/ingame-info.yaml
	ra: chrome/ingame-infoscripterror.yaml
```

penev92:
As a long-term solution I'd like for the mod chooser to be able to install mods (not assets).
This will include browsing mods hosted on http://www.resource.openra.net/ and downloading a selected mod into the `OpenRA\mods\` directory.
That would also depend on #8991, of course.

This could work as a short-term solution though.

pchote:
Installing mods into the main installation directory is probably never going to happen.  Operating system permissions prevent us from writing or modifying files there, and even if we did implement permission escalation for all our supported platforms it violates the platform standards under at least Linux.

penev92:
OK, I'd still like to see that happen though, even if the mods get installed into the user directory.

GraionDilach:
@penev92 Bad idea, third-party mods can be really large at filesize. The bigger TS/YR mods nowadays (Mental Omega, Twisted Insurrection, Robot Storm, Project Phantom, D-Day, etc) are atleast 200-300 MB, which has the danger of running the resource server out of it's quotas.

penev92:
[...]

PR #9853, opened 2015-11-01
Unglobal file system

penev92:
The next step toward thirdparty mod support (last step before my PR for #8992).

### DIFF
--- a/OpenRA.Game/FileSystem/BagFile.cs
+++ b/OpenRA.Game/FileSystem/BagFile.cs
@@ -39,14 +39,14 @@ namespace OpenRA.FileSystem
 
 			// Build the index and dispose the stream, it is no longer needed after this
 			List<IdxEntry> entries;
-			using (var indexStream = GlobalFileSystem.Open(indexFilename))
+			using (var indexStream = Game.ModData.ModFiles.Open(indexFilename))
 				entries = new IdxReader(indexStream).Entries;
 
 			index = entries.ToDictionaryWithConflictLog(x => x.Hash,
 				"{0} (bag format)".F(filename),
 				null, x => "(offs={0}, len={1})".F(x.Offset, x.Length));
 
-			s = GlobalFileSystem.Open(filename);
+			s = Game.ModData.ModFiles.Open(filename);
 		}
 
 		public int Priority { get { return 1000 + bagFilePriority; } }
@@ -159,9 +159,9 @@ namespace OpenRA.FileSystem
 		public IEnumerable<string> AllFileNames()
 		{
 			var lookup = new Dictionary<uint, string>();
-			if (GlobalFileSystem.Exists("global mix database.dat"))
+			if (Game.ModData.ModFiles.Exists("global mix database.dat"))
 			{
-				var db = new XccGlobalDatabase(GlobalFileSystem.Open("global mix database.dat"));
+				var db = new XccGlobalDatabase(Game.ModData.ModFiles.Open("global mix database.dat"));
 				foreach (var e in db.Entries)
 				{
 					var hash = IdxEntry.HashFilename(e, PackageHashType.CRC32);
@@ -175,7 +175,7 @@ namespace OpenRA.FileSystem
 
 		public void Write(Dictionary<string, byte[]> contents)
 		{
-			GlobalFileSystem.Unmount(this);
+			Game.ModData.ModFiles.Unmount(this);
 			throw new NotImplementedException("Updating bag files unsupported");
 		}
 
--- a/OpenRA.Game/FileSystem/BagFile.cs
+++ b/OpenRA.Game/FileSystem/BagFile.cs
@@ -27,11 +27,13 @@ namespace OpenRA.FileSystem
 		readonly Stream s;
 		readonly int bagFilePriority;
 		readonly Dictionary<uint, IdxEntry> index;
+		readonly FileSystem context;
 
-		public BagFile(string filename, int priority)
+		public BagFile(FileSystem context, string filename, int priority)
 		{
 			bagFilename = filename;
 			bagFilePriority = priority;
+			this.context = context;
 
 			// A bag file is always accompanied with an .idx counterpart
 			// For example: audio.bag requires the audio.idx file
@@ -39,14 +41,14 @@ namespace OpenRA.FileSystem
 
 			// Build the index and dispose the stream, it is no longer needed after this
 			List<IdxEntry> entries;
-			using (var indexStream = Game.ModData.ModFiles.Open(indexFilename))
+			using (var indexStream = context.Open(indexFilename))
 				entries = new IdxReader(indexStream).Entries;
 
 			index = entries.ToDictionaryWithConflictLog(x => x.Hash,
 				"{0} (bag format)".F(filename),
 				null, x => "(offs={0}, len={1})".F(x.Offset, x.Length));
 
-			s = Game.ModData.ModFiles.Open(filename);
+			s = context.Open(filename);
 		}
 
 		public int Priority { get { return 1000 + bagFilePriority; } }
@@ -159,9 +161,9 @@ namespace OpenRA.FileSystem
 		public IEnumerable<string> AllFileNames()
 		{
 			var lookup = new Dictionary<uint, string>();
-			if (Game.ModData.ModFiles.Exists("global mix database.dat"))
+			if (context.Exists("global mix database.dat"))
 			{
-				var db = new XccGlobalDatabase(Game.ModData.ModFiles.Open("global mix database.dat"));
+				var db = new XccGlobalDatabase(context.Open("global mix database.dat"));
 				foreach (var e in db.Entries)
 				{
 					var hash = IdxEntry.HashFilename(e, PackageHashType.CRC32);
@@ -175,7 +177,7 @@ namespace OpenRA.FileSystem
 
 		public void Write(Dictionary<string, byte[]> contents)
 		{
-			Game.ModData.ModFiles.Unmount(this);
+			context.Unmount(this);
 			throw new NotImplementedException("Updating bag files unsupported");
 		}
 
--- a/OpenRA.Game/FileSystem/FileSystem.cs
+++ b/OpenRA.Game/FileSystem/FileSystem.cs
@@ -28,14 +28,14 @@ namespace OpenRA.FileSystem
 		Cache<uint, List<IFolder>> crcHashIndex = new Cache<uint, List<IFolder>>(_ => new List<IFolder>());
 		Cache<uint, List<IFolder>> classicHashIndex = new Cache<uint, List<IFolder>>(_ => new List<IFolder>());
 
-		public static IFolder CreatePackage(string filename, int order, Dictionary<string, byte[]> content)
+		public IFolder CreatePackage(string filename, int order, Dictionary<string, byte[]> content)
 		{
 			if (filename.EndsWith(".mix", StringComparison.InvariantCultureIgnoreCase))
-				return new MixFile(filename, order, content);
+				return new MixFile(this, filename, order, content);
 			if (filename.EndsWith(".zip", StringComparison.InvariantCultureIgnoreCase))
-				return new ZipFile(filename, order, content);
+				return new ZipFile(this, filename, order, content);
 			if (filename.EndsWith(".oramap", StringComparison.InvariantCultureIgnoreCase))
-				return new ZipFile(filename, order, content);
+				return new ZipFile(this, filename, order, content);
 			if (filename.EndsWith(".RS", StringComparison.InvariantCultureIgnoreCase))
 				throw new NotImplementedException("The creation of .RS archives is unimplemented");
 			if (filename.EndsWith(".Z", StringComparison.InvariantCultureIgnoreCase))
@@ -50,7 +50,7 @@ namespace OpenRA.FileSystem
 			return new Folder(filename, order, content);
 		}
 
-		public static IFolder OpenPackage(string filename, string annotation, int order)
+		public IFolder OpenPackage(string filename, string annotation, int order)
 		{
 			if (filename.EndsWith(".mix", StringComparison.InvariantCultureIgnoreCase))
 			{
@@ -58,25 +58,25 @@ namespace OpenRA.FileSystem
 					? PackageHashType.Classic
 					: FieldLoader.GetValue<PackageHashType>("(value)", annotation);
 
-				return new MixFile(filename, type, order);
+				return new MixFile(this, filename, type, order);
 			}
 
 			if (filename.EndsWith(".zip", StringComparison.InvariantCultureIgnoreCase))
-				return new ZipFile(filename, order);
+				return new ZipFile(this, filename, order);
 			if (filename.EndsWith(".oramap", StringComparison.InvariantCultureIgnoreCase))
-				return new ZipFile(filename, order);
+				return new ZipFile(this, filename, order);
 			if (filename.EndsWith(".RS", StringComparison.InvariantCultureIgnoreCase))
-				return new D2kSoundResources(filename, order);
+				return new D2kSoundResources(this, filename, order);
 			if (filename.EndsWith(".Z", StringComparison.InvariantCultureIgnoreCase))
-				return new InstallShieldPackage(filename, order);
+				return new InstallShieldPackage(this, filename, order);
 			if (filename.EndsWith(".PAK", StringComparison.InvariantCultureIgnoreCase))
-				return new PakFile(filename, order);
+				return new PakFile(this, filename, order);
 			if (filename.EndsWith(".big", StringComparison.InvariantCultureIgnoreCase))
-				return new BigFile(filename, order);
+				return new BigFile(this, filename, order);
 			if (filename.EndsWith(".bag", StringComparison.InvariantCultureIgnoreCase))
-				return new BagFile(filename, order);
+				return new BagFile(this, filename, order);
 			if (filename.EndsWith(".hdr", StringComparison.InvariantCultureIgnoreCase))
-				return new InstallShieldCABExtractor(filename, order);
+				return new InstallShieldCABExtractor(this, filename, order);
 
 			return new Folder(filename, order);
 		}
--- a/OpenRA.Game/FileSystem/FileSystem.cs
+++ b/OpenRA.Game/FileSystem/FileSystem.cs
@@ -17,34 +17,16 @@ using OpenRA.Primitives;
 
 namespace OpenRA.FileSystem
 {
-	public static class GlobalFileSystem
+	public class FileSystem
 	{
-		public static List<IFolder> MountedFolders = new List<IFolder>();
-		static Cache<uint, List<IFolder>> classicHashIndex = new Cache<uint, List<IFolder>>(_ => new List<IFolder>());
-		static Cache<uint, List<IFolder>> crcHashIndex = new Cache<uint, List<IFolder>>(_ => new List<IFolder>());
+		public readonly List<string> FolderPaths = new List<string>();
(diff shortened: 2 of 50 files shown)


OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -111,7 +87,12 @@ public static IFolder OpenPackage(string filename, string annotation, int order)
 			return new Folder(filename, order);
 		}
 
-		public static void Mount(string name, string annotation = null)
+		public void Mount(IFolder mount)
+		{
+			if (!MountedFolders.Contains(mount)) MountedFolders.Add(mount);
```
Phrohdoh: Please move the instruction down to the next line.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -129,18 +110,26 @@ public static void Mount(string name, string annotation = null)
...
-				folder.Dispose();
+			MountedFolders.Add(folder);
 
-			MountedFolders.Clear();
-			FolderPaths.Clear();
-			classicHashIndex = new Cache<uint, List<IFolder>>(_ => new List<IFolder>());
-			crcHashIndex = new Cache<uint, List<IFolder>>(_ => new List<IFolder>());
+			foreach (var hash in folder.ClassicHashes())
+			{
+				var l = classicHashIndex[hash];
```
Phrohdoh: Better variable name please.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -129,18 +110,26 @@ public static void Mount(string name, string annotation = null)
...
+				if (!l.Contains(folder))
+					l.Add(folder);
+			}
+
+			foreach (var hash in folder.CrcHashes())
+			{
+				var l = crcHashIndex[hash];
```
Phrohdoh: Same.

OpenRA.Game/Map/Map.cs
```diff
@@ -602,7 +602,7 @@ public void Save(string toPath)
 				Path = toPath;
 
 				// Create a new map package
-				Container = GlobalFileSystem.CreatePackage(Path, int.MaxValue, entries);
+				Container = FileSystem.FileSystem.CreatePackage(Path, int.MaxValue, entries);
```
Phrohdoh: This is nasty namespacing.

OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
```diff
@@ -41,6 +39,16 @@ public void StartGame(Arguments args)
 			Ui.ResetAll();
 			Game.Settings.Save();
 
+			var installData = Game.ModData.Manifest.Get<ContentInstaller>();
+			var isModContentInstalled = installData.TestFiles.All(f => Game.ModData.ModFiles.Exists(Path.GetFileName(f)));
+
+			// Mod assets are missing!
```
Phrohdoh: The check explains this pretty well.

OpenRA.Platforms.Default/Shader.cs
```diff
@@ -25,11 +24,9 @@ class Shader : ThreadAffine, IShader
...
 		protected int CompileShaderObject(ShaderType type, string name)
 		{
-			string ext = type == ShaderType.VertexShader ? "vert" : "frag";
-			string filename = "glsl{0}{1}.{2}".F(Path.DirectorySeparatorChar, name, ext);
-			string code;
-			using (var file = new StreamReader(GlobalFileSystem.Open(filename)))
-				code = file.ReadToEnd();
+			var ext = type == ShaderType.VertexShader ? "vert" : "frag";
+			var filename = "glsl{0}{1}.{2}".F(Path.DirectorySeparatorChar, name, ext);
+			var code = File.ReadAllText(filename);
```
Phrohdoh: Is this subtly missing a catch (or check) for a `FileNotFoundException`?

Phrohdoh:
I cannot spot any regressions and all mods load fine for me (tried switching between them all 3 times).
:+1: after my nits are addressed.

OpenRA.Game/Map/Map.cs
penev92: Yes, it is. Quite unfortunate.

OpenRA.Platforms.Default/Shader.cs
penev92: I think we'd want to throw if the file is missing.

Phrohdoh: You can throw in one spot and catch in another to present useful information.
This was my point.

penev92:
Updated.

OpenRA.Game/FileSystem/BagFile.cs
```diff
@@ -39,14 +39,14 @@ public BagFile(string filename, int priority)
 
 			// Build the index and dispose the stream, it is no longer needed after this
 			List<IdxEntry> entries;
-			using (var indexStream = GlobalFileSystem.Open(indexFilename))
+			using (var indexStream = Game.ModData.ModFiles.Open(indexFilename))
```
pchote: The next step will be to pass the `ModFiles` as an argument in the ctor.  Fixing that isn't required for my :+1:, but IMO you may as well do it now.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -17,47 +17,23 @@
...
-					l.Add(folder);
-			}
-
-			foreach (var hash in folder.CrcHashes())
-			{
-				var l = crcHashIndex[hash];
-				if (!l.Contains(folder))
-					l.Add(folder);
-			}
+			FolderPaths = new List<string>();
```
pchote: These can be merged with the initial declarations above.

pchote: You can then remove the ctor definition completely.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -243,31 +239,36 @@ public static bool Exists(string name)
...
 				if (assembly.FullName == e.Name)
 					return assembly;
 
-			var frags = e.Name.Split(',');
+			return Game.ModData.ModFiles.ResolveAssemblyInner(e.Name);
+		}
+
+		Assembly ResolveAssemblyInner(string assemblyName)
+		{
+			// TODO: This doesn't work, but nothing actually uses it and fixing it is out of scope.
```
pchote: Can you elaborate on what part of this doesn't work?

OpenRA.Game/GameRules/MusicInfo.cs
```diff
@@ -36,11 +35,11 @@ public MusicInfo(string key, MiniYaml value)
 
 		public void Load()
 		{
-			if (!GlobalFileSystem.Exists(Filename))
+			if (!Game.ModData.ModFiles.Exists(Filename))
```
pchote: As for the packages, it would be better if this referenced a passed `ModData` (as opposed to `ModFiles`, which feels a little bit too specific IMO) so that we can unstatic this now instead of in a future pass.

OpenRA.Game/Graphics/Sheet.cs
```diff
@@ -49,7 +49,7 @@ public Sheet(SheetType type, ITexture texture)
 
 		public Sheet(SheetType type, string filename)
 		{
-			using (var stream = GlobalFileSystem.Open(filename))
+			using (var stream = File.Exists(filename) ? File.OpenRead(filename) : Game.ModData.ModFiles.Open(filename))
```
pchote: This looks a bit iffy.  What case causes `Game.ModData.ModFiles.Open` to fail?

It might be cleaner to pass a `Stream` into the ctor instead of a filename, and then this logic can be special-cased in the couple of cases that need it.

OpenRA.Game/Graphics/SpriteLoader.cs
```diff
@@ -66,7 +65,7 @@ public static Sprite[] GetSprites(string filename, ISpriteLoader[] loaders, Shee
 
 		public static ISpriteFrame[] GetFrames(string filename, ISpriteLoader[] loaders)
 		{
-			using (var stream = GlobalFileSystem.Open(filename))
+			using (var stream = Game.ModData.ModFiles.Open(filename))
```
pchote: Fixing this one is a further out of scope than my other requests, so i'm happy to leave this to the next PR.

OpenRA.Game/MiniYaml.cs
```diff
@@ -224,7 +223,7 @@ static string SplitAtColon(ref string t)
 
 		public static List<MiniYamlNode> FromFileInPackage(string path)
 		{
-			using (var stream = GlobalFileSystem.Open(path))
+			using (var stream = Game.ModData.ModFiles.Open(path))
```
pchote: This method isn't even used.  How about we just delete it?

OpenRA.Mods.Common/InstallUtils.cs
```diff
@@ -46,9 +45,9 @@ public static string GetMountedDisk(Func<string, bool> isValidDisk)
 				Directory.CreateDirectory(destPath);
 
 			Log.Write("debug", "Mounting {0}".F(srcPath));
-			GlobalFileSystem.Mount(srcPath);
+			Game.ModData.ModFiles.Mount(srcPath);
```
pchote: We can now fix the TODO about, but this should be left to the next PR.

OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
```diff
@@ -41,6 +39,15 @@ public void StartGame(Arguments args)
 			Ui.ResetAll();
 			Game.Settings.Save();
 
+			var installData = Game.ModData.Manifest.Get<ContentInstaller>();
```
pchote: Didn't we just ship this out of here?  Rebase failure?

OpenRA.Mods.Common/UtilityCommands/LegacyMapImporter.cs
```diff
@@ -130,7 +129,7 @@ enum IniMapFormat { RedAlert = 3 } // otherwise, cnc (2 variants exist, we don't
 
 		public void ConvertIniMap(string iniFile)
 		{
-			using (var stream = GlobalFileSystem.Open(iniFile))
+			using (var stream = Game.ModData.ModFiles.Open(iniFile))
```
pchote: Shouldn't this be using the standard (not our vfs) calls?

pchote: Hang on... this shouldn't be using our vfs, should it?

OpenRA.Mods.Common/LoadScreens/BlankLoadScreen.cs
penev92: I guess so.

OpenRA.Game/FileSystem/BagFile.cs
penev92: In whose constructor?

pchote: `public BagFile(ModData modData, string filename, int priority)` and so on for the other package types.  They are all created from the same one place in the code, so this isn't a lot of work.

penev92:
Updated and added one extra commit for the package file formats.

OpenRA.Mods.Common/UtilityCommands/LegacyMapImporter.cs
penev92: *shrug*
In theory it should work with `File.Open()`, but I won't change it until someone who can test it confirms no side-effects.

OpenRA.Game/FileSystem/BagFile.cs
```diff
@@ -159,9 +159,9 @@ public IEnumerable<uint> CrcHashes()
 		public IEnumerable<string> AllFileNames()
 		{
 			var lookup = new Dictionary<uint, string>();
-			if (GlobalFileSystem.Exists("global mix database.dat"))
+			if (Game.ModData.ModFiles.Exists("global mix database.dat"))
```
pchote: These need to use the same `FileSystem` that you pass into the ctor.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -17,51 +17,21 @@
...
+		Cache<uint, List<IFolder>> classicHashIndex = new Cache<uint, List<IFolder>>(_ => new List<IFolder>());
 
-		public static IFolder CreatePackage(string filename, int order, Dictionary<string, byte[]> content)
+		public IFolder CreatePackage(string filename, int order, Dictionary<string, byte[]> content)
 		{
 			if (filename.EndsWith(".mix", StringComparison.InvariantCultureIgnoreCase))
-				return new MixFile(filename, order, content);
+				return new MixFile(this, filename, order, content);
 			if (filename.EndsWith(".zip", StringComparison.InvariantCultureIgnoreCase))
 				return new ZipFile(filename, order, content);
```
pchote: Can you pass `this` to ZipFile too?  This will make life easier when we extract an interface for package types.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -96,22 +66,28 @@ public static IFolder OpenPackage(string filename, string annotation, int order)
 			if (filename.EndsWith(".oramap", StringComparison.InvariantCultureIgnoreCase))
 				return new ZipFile(filename, order);
```
pchote: Likewise here.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -243,31 +233,13 @@ public static bool Exists(string name)
...
 
-		static Dictionary<string, Assembly> assemblyCache = new Dictionary<string, Assembly>();
-
 		public static Assembly ResolveAssembly(object sender, ResolveEventArgs e)
 		{
 			foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
 				if (assembly.FullName == e.Name)
 					return assembly;
 
-			var frags = e.Name.Split(',');
```
pchote: I'd like to work out what this is actually used for before we remove it.  I guess some docs-reading is in order.

pchote:
Code changes look good except for the points noted above.  Haven't tested anything yet, however.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -243,31 +233,13 @@ public static bool Exists(string name)
...
-				using (var s = Open(filename))
-				{
-					var buf = s.ReadBytes((int)s.Length);
-					a = Assembly.Load(buf);
-					assemblyCache.Add(filename, a);
-					return a;
-				}
-
-			return null;
+			throw new InvalidOperationException("Unable to resolve assembly {0}!".F(e.Name));
```
pchote: According to the docs this should continue to `return null;` instead of throwing.

OpenRA.Game/FileSystem/FileSystem.cs
pchote: Okay, so this gets called when you have a mod dll (which are loaded manually elsewhere) that links against an additional dll that isn't referenced by the mod.yaml.  This isn't a problem for our mod dlls, but it may be if they want to link against an external library... might need a test case for this.

pchote: In any case, i'm pretty sure that the check above is effectively a no-op, so if we do remove this code below then we may as well remove the method and the event entirely.

pchote: But I think the safest action for this pr is to just convert this method to use the new FS methods and leave it otherwise unchanged.

abcdefg30:
> OpenRA.Game/FileSystem/FileSystem.cs(25,41): error CS0414: Warning as Error: The private field `OpenRA.FileSystem.FileSystem.assemblyCache' is assigned but its value is never used

penev92:
Updated.

OpenRA.Game/FileSystem/ZipFile.cs
```diff
@@ -19,6 +19,7 @@ namespace OpenRA.FileSystem
 {
 	public sealed class ZipFile : IFolder
 	{
+		readonly FileSystem context;
```
pchote: You don't need to store this - it triggers an unused variable error.

abcdefg30:
Travis fails again:
> OpenRA.Game/FileSystem/ZipFile.cs(22,23): error CS0414: Warning as Error: The private field `OpenRA.FileSystem.ZipFile.context' is assigned but its value is never used

penev92:
Silenced Travis for good this time.

abcdefg30:
Almost.
> OpenRA.Game/FileSystem/FileSystem.cs:L25: [SA1311] Static readonly fields must start with an upper-case letter: assemblyCache.

penev92:
> Silenced Travis for good this time.

abcdefg30:
Needs a rebase.

penev92:
Rebased.

obrakmann:
This looks ok overall, but introduces a problem during music installation (at least in RA). The regular ra assets already contain a scores.mix file, which gets overwritten during music installation. That works fine on bleed, but on this branch I gets this exception (no crash though):

```
Exception of type `System.IO.IOException`: Sharing violation on path /home/oliver/.openra/Content/ra/scores.mix
  at System.IO.FileStream..ctor (System.String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, Boolean anonymous, FileOptions options) <0x7f4dde95d060 + 0x005ed> in <filename unknown>:0 
  at System.IO.FileStream..ctor (System.String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize) <0x7f4dde95cd20 + 0x0004d> in <filename unknown>:0 
  at (wrapper remoting-invoke-with-check) System.IO.FileStream:.ctor (string,System.IO.FileMode,System.IO.FileAccess,System.IO.FileShare,int)
  at System.IO.File.Create (System.String path, Int32 bufferSize) <0x7f4dde959da0 + 0x0003b> in <filename unknown>:0 
  at System.IO.File.Create (System.String path) <0x7f4dde959d80 + 0x00011> in <filename unknown>:0 
  at OpenRA.Mods.Common.InstallUtils.ExtractZip (ICSharpCode.SharpZipLib.Zip.ZipInputStream z, System.String destPath, System.Collections.Generic.List`1 extracted, System.Action`1 onProgress) [0x0005d] in /home/oliver/devel/openra/git/OpenRA.Mods.Common/InstallUtils.cs:160 
  at OpenRA.Mods.Common.InstallUtils.ExtractZip (System.String zipFile, System.String dest, System.Action`1 onProgress, System.Action`1 onError) [0x00058] in /home/oliver/devel/openra/git/OpenRA.Mods.Common/InstallUtils.cs:135 
  at OpenRA.Mods.Common.Widgets.Logic.DownloadPackagesLogic+<ShowDownloadDialog>c__AnonStorey3.<>m__3 (System.ComponentModel.AsyncCompletedEventArgs i, Boolean cancelled) [0x0009a] in /home/oliver/devel/openra/git/OpenRA.Mods.Common/Widgets/Logic/Installation/DownloadPackagesLogic.cs:116 
```

penev92:
Nice catch. I had to unmount the mod files when finishing with a mod.

obrakmann:
Yup, music installation works again. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/ff2bf2534f1b7d0f4197edab89a4cec3618c728b). I tacked the changelog entry onto the end of the entry for the inter-mod dependencies PR. I hope that the rest of the mod packaging PRs will get in as well, then that whole line can be changed to "added support for mod packages".

PR #9854, opened 2015-11-01
prevent allied units from autoattack building which is being captured…

### LINKED ISSUE #6170: Allied units shouldn't attack a building which is being captured [$20]

Arcitectus:
…. resolves #6170.

### DIFF
--- a/OpenRA.Mods.Common/Traits/AutoTarget.cs
+++ b/OpenRA.Mods.Common/Traits/AutoTarget.cs
@@ -168,7 +168,8 @@ namespace OpenRA.Mods.Common.Traits
 		Actor ChooseTarget(Actor self, WDist range, bool allowMove)
 		{
 			var inRange = self.World.FindActorsInCircle(self.CenterPosition, range)
-				.Where(a => !a.Info.HasTraitInfo<AutoTargetIgnoreInfo>());
+				.Where(a =>
+					!a.TraitsImplementing<IPreventsAutoTarget>().Any(t => t.PreventsAutoTarget(a, self)));
 
 			// Armaments are enumerated in attack.Armaments in construct order
 			// When autotargeting, first choose targets according to the used armament construct order
@@ -207,7 +208,13 @@ namespace OpenRA.Mods.Common.Traits
 
 	[Desc("Will not get automatically targeted by enemy (like walls)")]
 	class AutoTargetIgnoreInfo : TraitInfo<AutoTargetIgnore> { }
-	class AutoTargetIgnore { }
+	class AutoTargetIgnore : IPreventsAutoTarget
+	{
+		public bool PreventsAutoTarget(Actor self, Actor attacker)
+		{
+			return true;
+		}
+	}
 
 	public class StanceInit : IActorInit<UnitStance>
 	{
--- a/OpenRA.Mods.Common/Traits/ExternalCapturable.cs
+++ b/OpenRA.Mods.Common/Traits/ExternalCapturable.cs
@@ -24,6 +24,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Seconds it takes to change the owner.", "You might want to add a ExternalCapturableBar: trait, too.")]
 		public readonly int CaptureCompleteTime = 15;
 
+		[Desc("Whether to prevent autotargeting this actor while it is being captured by an ally.")]
+		public readonly bool PreventsAutoTarget = true;
+
 		public bool CanBeTargetedBy(Actor captor, Player owner)
 		{
 			var c = captor.Info.TraitInfoOrDefault<ExternalCapturesInfo>();
@@ -49,7 +52,7 @@ namespace OpenRA.Mods.Common.Traits
 		public object Create(ActorInitializer init) { return new ExternalCapturable(init.Self, this); }
 	}
 
-	public class ExternalCapturable : ITick, ISync
+	public class ExternalCapturable : ITick, ISync, IPreventsAutoTarget
 	{
 		[Sync] public int CaptureProgressTime = 0;
 		[Sync] public Actor Captor;
@@ -90,6 +93,11 @@ namespace OpenRA.Mods.Common.Traits
 				CaptureProgressTime = 0;
 			else
 				CaptureProgressTime++;
+		}
+
+		public bool PreventsAutoTarget(Actor self, Actor attacker)
+		{
+			return Info.PreventsAutoTarget && Captor != null && attacker.AppearsFriendlyTo(Captor);
 		}
 	}
 }
--- a/AUTHORS
+++ b/AUTHORS
@@ -90,6 +90,7 @@ Also thanks to:
     * Matthijs Benschop (Nerdie)
     * Max621
     * Max Ugrumov (katzsmile)
+    * Michael Rätzel
     * Michael Sztolcman (s1w_)
     * Mustafa Alperen Seki (MustaphaTR)
     * Neil Shivkar (havok13888)
--- a/OpenRA.Mods.Common/TraitsInterfaces.cs
+++ b/OpenRA.Mods.Common/TraitsInterfaces.cs
@@ -121,4 +121,9 @@ namespace OpenRA.Mods.Common.Traits
 	{
 		void ModifyDeathActorInit(Actor self, TypeDictionary init);
 	}
+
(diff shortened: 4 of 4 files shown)


OpenRA.Mods.Common/Traits/ExternalCapturable.cs
```diff
@@ -91,5 +91,15 @@ public void Tick(Actor self)
...
 		}
+
+		public bool PreventAutoTarget(Actor self, Actor attacker)
+		{
+			if (Captor == null)
+			{
+				return false;
+			}
+
+			return attacker.AppearsFriendlyTo(Captor);
```
Phrohdoh: Could you change this entire block to `return Captor != null && attacker.AppearsFriendlyTo(Captor);`, please?

Phrohdoh:
:+1: after that change.
This works and is pretty neat.

GraionDilach:
I request a tag to make this optional.

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -167,15 +167,17 @@ void Attack(Actor self, Actor targetActor, bool allowMove)
 
 		Actor ChooseTarget(Actor self, WDist range, bool allowMove)
 		{
-			var inRange = self.World.FindActorsInCircle(self.CenterPosition, range)
-				.Where(a => !a.Info.HasTraitInfo<AutoTargetIgnoreInfo>());
+			var inRange = self.World.FindActorsInCircle(self.CenterPosition, range);
```
pchote: This isn't used anywhere else, so why split it out?

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -168,7 +168,8 @@ void Attack(Actor self, Actor targetActor, bool allowMove)
 		Actor ChooseTarget(Actor self, WDist range, bool allowMove)
 		{
 			var inRange = self.World.FindActorsInCircle(self.CenterPosition, range)
-				.Where(a => !a.Info.HasTraitInfo<AutoTargetIgnoreInfo>());
+				.Where(a =>
+				!a.TraitsImplementing<IPreventsAutoTarget>().Any(t => t.PreventAutoTarget(a, self)));
```
reaperrr: Our style convention is to indent cases like this 2nd line with another tab (to make it more visible that the line belongs to the `Where` above).

Phrohdoh:
My comment was read before being updated.
The original was this:

Could you add a bool AutoTargetIgnores to the ExternalCapturableInfo class so this can be (un)set per-actor type then change the entire block to `return Info.AutoTargetIgnores && Captor != null && attacker.AppearsFriendlyTo(Captor);`, please?

penev92:
I'd like to hear opinions on two things:
IPreventsAutoTarget -> IPreventAutoTarget
and
bool PreventAutoTarget() -> bool IsPreventingAutoTarget()

GraionDilach:
IPreventAutoTarget and PreventAutoTarget imo.

Phrohdoh:
@penev92 I agree on both naming changes, and have another request now that the info field has been added.

OpenRA.Mods.Common/Traits/ExternalCapturable.cs
```diff
@@ -21,6 +21,7 @@ public class ExternalCapturableInfo : ITraitInfo
 		public readonly bool AllowAllies = false;
 		public readonly bool AllowNeutral = true;
 		public readonly bool AllowEnemies = true;
+		public readonly bool AutoTargetIgnores = false;
```
Phrohdoh: Please add a `desc` (like the one below) to this trait property describing the circumstance that this is tied to.

GraionDilach: Also a newline between `AllowEnemies` and the `desc` imo.

Arcitectus:
I applied the suggested changes, default is now the old behaviour and the change can be activated by setting `AutoTargetIgnores`.

GraionDilach:
:+1:

OpenRA.Mods.Common/Traits/ExternalCapturable.cs
```diff
@@ -24,6 +24,9 @@ public class ExternalCapturableInfo : ITraitInfo
 		[Desc("Seconds it takes to change the owner.", "You might want to add a ExternalCapturableBar: trait, too.")]
 		public readonly int CaptureCompleteTime = 15;
 
+		[Desc("Whether autotargeting ignores this actor while it is being captured by an ally.")]
+		public readonly bool AutoTargetIgnores = false;
```
Mailaender: I think you can safely set this to `true`. At least for the monster tank madness mission.

Arcitectus: @Mailaender 
> I think you can safely set this to true. At least for the Monster Tank Madness mission.

I am not sure what that means in relation to the PR.
I thought that this was the wrong place to implement a setting for a specific mission. Which led me to believe that your comment was just meant as a note to users of the field. Is that correct?
Or do you think changes should be made to the PR?

Phrohdoh: I think changing individual missions is out of scope for this PR.

penev92:
Needs a rebase.

OpenRA.Mods.Common/Traits/ExternalCapturable.cs
abcdefg30: I'm with @Mailaender here. If you don't set this to true, this doesn't really fix #6170 (except you add some yaml to the default mods), as the behaviour ingame won't change.

Phrohdoh:
This was silently rebased.

Phrohdoh:
I suspect something is missing from this, I'll investigate and post my findings.

Phrohdoh:
~delete~

Arcitectus:
@Phrohdoh 
I changed the names of interface and its method to match the examples you cited.

OpenRA.Mods.Common/Traits/ExternalCapturable.cs
```diff
@@ -24,6 +24,9 @@ public class ExternalCapturableInfo : ITraitInfo
 		[Desc("Seconds it takes to change the owner.", "You might want to add a ExternalCapturableBar: trait, too.")]
 		public readonly int CaptureCompleteTime = 15;
 
+		[Desc("Whether autotargeting ignores this actor while it is being captured by an ally.")]
+		public readonly bool AutoTargetIgnores = true;
```
pchote: Can we change this to `PreventsAutoTarget` to match the other recent changes?

Phrohdoh:
Thanks Arcitectus! Sorry for all of the back and forth.

Just one more thing to change then this should be on its way. :)

Arcitectus:
> Just one more thing to change

done.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/740cd9a2afd1905e25ad03a8121503aab2a17426).

@Arcitectus, don't forget to collect your bounty!

r34ch:
@Arcitectus Currently on bleed I have found two issues;

1. If a unit is attacking a building and then a friendly engineer starts capturing it, that unit will not stop attacking until ordered to move; with the exception of

2. turreted units will continue to attack the building being captured even when moved, including moving out of range and back into range.

Both attacking issues can be stopped by manually hitting the S hotkey.

Arcitectus:
@r34ch I will look for a way to invalidate the targets for those scenarios.

GraionDilach:
2) is pretty much beyond this one's scope a lot, forceattack terrain works that way and prolly a lot of stuff the recent targeting PRs are touching.

PR #9861, opened 2015-11-02
Allow for cash refineries.

ghost:
RA2 does not use resources and rather only has cash---that is, there are no limits on resource possession and engineers cannot steal resources.

The easiest and most flexible way to implement this is to allow cash-only refineries, which would allow RA2 to not use resources at all. It would also open up the possibility for future modders to have refineries which give cash rather than resources.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Buildings/Refinery.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/Refinery.cs
@@ -35,6 +35,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("In how many steps to perform the dragging?")]
 		public readonly int DragLength = 0;
 
+		[Desc("Does this refinery grant cash instead of resources?")]
+		public readonly bool GivesCash = false;
+
 		public readonly bool ShowTicks = true;
 		public readonly int TickLifetime = 30;
 		public readonly int TickVelocity = 2;
@@ -84,11 +87,17 @@ namespace OpenRA.Mods.Common.Traits
 				.Where(a => a.Trait.LinkedProc == self);
 		}
 
-		public bool CanGiveResource(int amount) { return playerResources.CanGiveResources(amount); }
+		public bool CanGiveResource(int amount)
+		{
+			return info.GivesCash ? true : playerResources.CanGiveResources(amount);
+		}
 
 		public void GiveResource(int amount)
 		{
-			playerResources.GiveResources(amount);
+			if (info.GivesCash)
+				playerResources.GiveCash(amount);
+			else
+				playerResources.GiveResources(amount);
 			if (info.ShowTicks)
 				currentDisplayValue += amount;
 		}


OpenRA.Mods.Common/Traits/Buildings/Refinery.cs
```diff
@@ -84,11 +87,17 @@ public IEnumerable<TraitPair<Harvester>> GetLinkedHarvesters()
 				.Where(a => a.Trait.LinkedProc == self);
 		}
 
-		public bool CanGiveResource(int amount) { return playerResources.CanGiveResources(amount); }
+		public bool CanGiveResource(int amount)
+		{
+			return info.GivesCash ? true : playerResources.CanGiveResources(amount);
```
Phrohdoh: `return !info.GivesCash && playerResources.CanGiveResources(amount);`

pchote:
I'm not too thrilled by this: it's a short term hack around something that we know needs to be completely redone in the medium term.  The current two hardcoded resource types are far too restrictive, and don't cover the needs of... pretty much anything except our default RA and TD mods.  As you've mentioned: RA2 only has a single resource type, TS needs a new resource type (veins), and we've had requests from people who wanted to implement AOE-style resources.

We need to break out individual resource type traits, and then have the traits that reference them (`Refinery`, `GiveCashCrateAction`, etc) specify the resource type that they provide / consume.

pchote:
I'm not too thrilled by this: it's a short term hack around something that we know needs to be completely redone in the medium term.  The current two hardcoded resource types are far too restrictive, and don't cover the needs of... pretty much anything except our default RA and TD mods.  As you've mentioned: RA2 only has a single resource type, TS needs a new resource type (veins), and we've had requests from people who wanted to implement AOE-style resources.

We need to break out individual resource type traits, and then have the traits that reference them (`Refinery`, `GiveCashCrateAction`, etc) specify the resource type that they provide / consume.

GraionDilach:
Dammit, I just wanted to do this!

@pchote I thought that would be done via running multiple `PlayerResource` traits (one per each different resource), differing them via, say a `Name` tag and everything which uses `PlayerResource` would specify which one it uses.

pchote:
@GraionDilach: indeed.

OpenRA.Mods.Common/Traits/Buildings/Refinery.cs
GraionDilach: `return info.GivesCash || playerResources.CanGiveResources(amount)` sounds more friendly to me.

OpenRA.Mods.Common/Traits/Buildings/Refinery.cs
```diff
@@ -84,11 +87,17 @@ public IEnumerable<TraitPair<Harvester>> GetLinkedHarvesters()
...
+		}
 
 		public void GiveResource(int amount)
 		{
-			playerResources.GiveResources(amount);
+			if (info.GivesCash)
```
GraionDilach:
```
info.GivesCash ? playerResources.GiveCash(amount)
	: playerResources.GiveResources(amount);
```

Phrohdoh:
@GraionDilach If you want to start implementing this chat with me. I've done exactly what you described before, but at the time it was more than I could chew so wasn't completed.

GraionDilach:
Not for now for sure. I already picked a task to chew on afterall.

SeanAtIgloo:
Ok, going to close this and make an issue with design.

PR #9871, opened 2015-11-02
Fix hex colors upgrade rules engine version

atlimit8:
The playtest came out (20151031) before hex colors were merged but the upgrade rules use version 20151027.


penev92:
That PR didn't get into the playtest (and so is not on the prep branch), so no need for the "For stable" label and the "Next release" milestone.

GraionDilach:
:+1:

obrakmann:
Yeah, sorry. I just checked that there wasn't any newer engine version referenced in the upgrade rules already. Didn't occur to me at all that there was a release in-between.

PR #9873, opened 2015-11-03
Handle mouse up event in default editor brush (instead of mouse down)

### LINKED ISSUE #9836: Joystick scrolling doesn't work in the map editor
<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27861777-joystick-scrolling-doesn-t-work-in-the-map-editor?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

teees:
Let the mouse down event bubble up, fixes #9836
EDIT:
- applied this behaviour to all brushes (only affects right clicks), so you can also joystick scroll with with all other brushes

### DIFF
--- a/OpenRA.Mods.Common/EditorBrushes/EditorActorBrush.cs
+++ b/OpenRA.Mods.Common/EditorBrushes/EditorActorBrush.cs
@@ -89,8 +89,13 @@ namespace OpenRA.Mods.Common.Widgets
 
 			if (mi.Button == MouseButton.Right)
 			{
-				editorWidget.ClearBrush();
-				return true;
+				if (mi.Event == MouseInputEvent.Up)
+				{
+					editorWidget.ClearBrush();
+					return true;
+				}
+
+				return false;
 			}
 
 			var cell = worldRenderer.Viewport.ViewToWorld(mi.Location);
--- a/OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
+++ b/OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
@@ -55,12 +55,11 @@ namespace OpenRA.Mods.Common.Widgets
 		{
 			// Exclusively uses mouse wheel and right mouse buttons, but nothing else
 			// Mouse move events are important for tooltips, so we always allow these through
-			if (mi.Button != MouseButton.Right && mi.Event != MouseInputEvent.Move && mi.Event != MouseInputEvent.Scroll)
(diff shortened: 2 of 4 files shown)


pchote:
I suspect that switching from mouse down to mouse up may regress something else - please check that carefully.

teees:
You were right, if you right clicked with another editor brush, you got an unwanted mouseup event on the default brush.
I fixed that, as a side effect you can now also joystick scroll in all other brushes

OpenRA.Mods.Common/EditorBrushes/EditorActorBrush.cs
```diff
@@ -89,8 +89,13 @@ public bool HandleMouseInput(MouseInput mi)
...
 			if (mi.Button == MouseButton.Right)
 			{
-				editorWidget.ClearBrush();
-				return true;
+				if (mi.Event == MouseInputEvent.Up)
+				{
+					editorWidget.ClearBrush();
+					return true;
+				}
+				else
```
pchote: Style nit: you can drop the else statement and have the `return false` unconditional because the if branch returns.

pchote:
> You were right, if you right clicked with another editor brush, you got an unwanted mouseup event on the default brush.

IIRC I was asked specifically to add this to make it match the behavior of the original editor prototype.  I have no issue with changing it back, especially since it conflicts with other desired behaviour.

OpenRA.Mods.Common/EditorBrushes/EditorActorBrush.cs
teees: will do, rebasing

teees:
As I see it, it is now (more) canon with the ingame build and place behaviour, where a right mouse up event cancels your place order, while a mousedown won't do anything

penev92:
Adding this to the milestone because the issue it fixes needs to be resolved. Hopefully this fixes it in a way we can accept (I haven't read the code or tested).

obrakmann:
Looks good to me. I couldn't find any breakage due to this change, and josystick scrolling does indeed work now in the editor. :+1:

pchote:
Doesn't regress anything, code looks fine, and works as promised :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/91c136ad888ec8eac34fe72574ab759de79e49aa

PR #9874, opened 2015-11-03
Drop targets out of range if we can't move

### LINKED ISSUE #6185: AA and Sam site should face their targets only when they're within range
They seem to follow enemy aircraft far beyond their range. But not always, I haven't tested it but in the early game they seem to face aircraft anywhere on the map.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3567417-aa-and-sam-site-should-face-their-targets-only-when-they-re-within-range?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #9837: AI defenses don't stop tracking units even if InitialStanceAI is set to Defend
They do switch once a new target comes into range, but as long as that doesn't happen, AI turrets continue to track a target even if it's out of range.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27861950-ai-defenses-don-t-stop-tracking-units-even-if-initialstanceai-is-set-to-defend?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #9849: Camo Pillboxes dont cloak.
Found out in a 1V1 game, other player said he could totally see them so it wasn't just me. Tested in a solo game, same results. Doubt this is intentional.
In playtest 20151031

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27871822-camo-pillboxes-dont-cloak?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

teees:
I noticed that my mammy in guard stance was trying to shoot a far away harvester while letting himself kill by infantry.

Closes #9837.
Closes #6185.
Closes #9849.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Attack.cs
+++ b/OpenRA.Mods.Common/Activities/Attack.cs
@@ -76,9 +76,13 @@ namespace OpenRA.Mods.Common.Activities
 			minRange = armaments.Max(a => a.Weapon.MinRange);
 			maxRange = armaments.Min(a => a.MaxRange());
 
-			// Try to move within range
-			if (move != null && (!Target.IsInRange(self.CenterPosition, maxRange) || Target.IsInRange(self.CenterPosition, minRange)))
+			if (!Target.IsInRange(self.CenterPosition, maxRange) || Target.IsInRange(self.CenterPosition, minRange))
+			{
+				// Try to move within range, drop the target otherwise
+				if (move == null)
+					return NextActivity;
 				return Util.SequenceActivities(move.MoveWithinRange(Target, minRange, maxRange), this);
+			}
 
 			var desiredFacing = Util.GetFacing(Target.CenterPosition - self.CenterPosition, 0);
 			if (facing.Facing != desiredFacing)
--- a/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
@@ -96,8 +96,11 @@ namespace OpenRA.Mods.Common.Traits
 
 					if (move != null)
 						return Util.SequenceActivities(move.MoveFollow(self, target, weapon.Weapon.MinRange, maxRange), this);
+					if (target.IsInRange(self.CenterPosition, weapon.MaxRange()) && !target.IsInRange(self.CenterPosition, weapon.Weapon.MinRange))
+						return this;
 				}
 
+				attack.Target = Target.Invalid;
 				return NextActivity;
 			}
 		}


pchote:
Note that this will only affect non-turreted units.  Turreted units use the `Follow` activity instead (which may not have this problem).

GraionDilach:
Pretty sure something similar is needed for them too, think of the TD SAM tracking planes-outside-range-and-not-closing-due-to.-this issue.

teees:
This affects turreted units now as well. AttackFollow was bugged for immovable actors

teees:
fixes #9837 
fixes #6185
And possibly also #9849

obrakmann:
This does indeed fix all of those bugs, good job!

Unfortunately, it breaks the TD gunboat (see the first GDI mission).

penev92:
Uhh not that gunboat again

GraionDilach:
Remove `AutoTarget>AllowMovement: false` from the Gunboat and it works without a problem.

pchote:
That will then break its fixed path movement.

GraionDilach:
I just finished gdi01 and it didn't looked like that to me (probably because it has `RejectsOrders` and owned by the player - thereby having the `Defend` stance - instead of owned by an AI house with `AttackAnything`).

Confirming the fix works though.

teees:
I'm guessing the map is currently using broken behavior

> Actor will follow units until in range to attack them.

I'm guessing the Trait is now working as expected (?)

pchote:
Removing `AllowMovement` from the gunboat is not sufficient - that is just hiding that that feature is now broken.

teees:
Hmm so some thoughts about this - the current level design is using `AttackMove` on an actor with `AutoTarget>AllowMovement:false` defined, meaning that `AutoTarget`should not alter the movement.

Wouldn't it be more sane to have a move command that did that instead of using this hacky behaviour? So move-and-attack-anything-you-come-across-without-stopping

Currently this is only explicitely done, so if you give an attack order to a turreted unit and order it to move, it will keep attacking. Wouldn't it be useful for units to do that automatically according to their stance? (iirc turreted units in the original ts did that)

obrakmann:
This patch on top of yours will make the gunboat work again (it basically undoes the changes from your second commit if `OnRails` is set). This is ugly, but at least it keeps all the `OnRails` hackery in the same place (`AttackFollows`).

```diff
diff --git a/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs b/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
index 6280de3..3582f5e 100644
--- a/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
@@ -56,6 +56,7 @@ class AttackActivity : Activity
                {
                        readonly AttackFollow attack;
                        readonly IMove move;
+                       readonly bool onRailsHack;
                        readonly Target target;
                        readonly bool forceAttack;
 
@@ -64,10 +65,13 @@ public AttackActivity(Actor self, Target target, bool allowMove, bool forceAttac
                                attack = self.Trait<AttackFollow>();
                                move = allowMove ? self.TraitOrDefault<IMove>() : null;
 
-                               // HACK: Mobile.OnRails is horrible
+                               // HACK: Mobile.OnRails is horrible. Blergh.
                                var mobile = move as Mobile;
                                if (mobile != null && mobile.Info.OnRails)
+                               {
                                        move = null;
+                                       onRailsHack = true;
+                               }
 
                                this.target = target;
                                this.forceAttack = forceAttack;
@@ -96,11 +100,15 @@ public override Activity Tick(Actor self)
 
                                        if (move != null)
                                                return Util.SequenceActivities(move.MoveFollow(self, target, weapon.Weapon.MinRange, maxRange), this);
-                                       if (target.IsInRange(self.CenterPosition, weapon.MaxRange()) && !target.IsInRange(self.CenterPosition, weapon.Weapon.MinRange))
+                                       if (!onRailsHack &&
+                                               target.IsInRange(self.CenterPosition, weapon.MaxRange()) &&
+                                               !target.IsInRange(self.CenterPosition, weapon.Weapon.MinRange))
                                                return this;
                                }
 
-                               attack.Target = Target.Invalid;
+                               if (!onRailsHack)
+                                       attack.Target = Target.Invalid;
+
                                return NextActivity;
                        }
                }
```

RoosterDragon:
FWIW - I would very much like to take this change, and I think @obrakmann's workaround is a sensible approach to keep the hack working as intended (I'm just assuming it works - haven't tested it).

GraionDilach:
FWIW I'd even say deprecating AllowMovement in favor of a unchangeable Defend stance on affected actors works better even.

obrakmann:
Superseded by #10750.

PR #9875, opened 2015-11-03
D2k, fixed infantry moving too slow on dunes

OmegaBolt:
This isn't a 'scientific' fix because I don't believe we know the exact terrain movement stats, but playing the original, infantry definitely seem to move faster on dunes than vehicles and don't have as a slow a speed reduction as there is here.

Adding to that, it does IMO make sense that infantry wouldn't slow down as much and it gives them another nice boost to usability on the battlefield.


Mailaender:
:+1: for a step towards less annoying gameplay.

penev92:
:+1: good catch.

PR #9877, opened 2015-11-03
Show unread message count on lobby chat tabs.

### LINKED ISSUE #9833: Lobby chat tabs should show unread message count
Right now there is no way to tell when new messages come in.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27861754-lobby-chat-tabs-should-show-unread-message-count?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Closes #9833.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
@@ -62,6 +62,10 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 		readonly LabelWidget chatLabel;
 		bool teamChat;
 
+		int lobbyChatUnreadMessages;
+		int globalChatLastReadMessages;
+		int globalChatUnreadMessages;
+
 		// Listen for connection failures
 		void ConnectionStateChanged(OrderManager om)
 		{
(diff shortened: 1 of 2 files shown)


RoosterDragon:
If I join the global chat from the multiplayer tab, then switch to the skirmish tab, all messages seen on the multiplayer tab are regarded as unread.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -62,6 +62,10 @@ enum ChatPanelType { Lobby, Global }
 		readonly LabelWidget chatLabel;
 		bool teamChat;
 
+		int lobbyChatUnreadMessages;
+		int globalChatLastReadMessages;
```
obrakmann: Making this `static` fixes the thing that RoosterDragon mentioned. I think I'd like it better that way, too.

You just need a way to reset it to 0 when you press the `leave chat` button.

pchote: Setting the initial value in the lobbylogic ctor should be an even easier fix

obrakmann: Derp. Yeah, it is.

pchote:
Fixed.

penev92:
:+1:

obrakmann:
Works nicely now. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/2db89075bbe535542f5d84625aa4edbfa791588d)

PR #9881, opened 2015-11-04
Polish ProximityCapturable.

GraionDilach:
Added descriptions and an Influenced field for gamemodes where the capturing actors only would initiate the capturing process.

Practically everything which was agreed on the IRC yesterday. I didn't removed the inline comments though - should that also happen?

### DIFF
--- a/OpenRA.Mods.Common/Traits/ProximityCapturable.cs
+++ b/OpenRA.Mods.Common/Traits/ProximityCapturable.cs
@@ -18,10 +18,21 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Actor can be captured by units in a specified proximity.")]
 	public class ProximityCapturableInfo : ITraitInfo
 	{
+		[Desc("If set, the actor can only be captured via this logic once.")]
 		public readonly bool Permanent = false;
+
+		[Desc("Maximum range at a ProximityCaptor actor can initiate the capture.")]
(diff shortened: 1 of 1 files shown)


pchote:
It's out of scope for this PR, but the capture range checks should also be rewritten using proximity triggers (`ActorMap.AddProximityTrigger`, `ActorMap.RemoveProximityTrigger`) instead of DIYing its own queries each tick.

GraionDilach:
Asking for the Has Dependencies label. I'll do that.

The trait has instant capture right now. >_>

obrakmann:
This hasn't seen any activity in a while so I took a shot at it, since I'll need this for my mod: https://github.com/OpenRA/OpenRA/compare/bleed...obrakmann:koth-refactor

@GraionDilach: let me know how we want to go from here.

GraionDilach:
I have a complete rewrire in mind tbh, but I couldn't find time to implement it before Wednesday... I have no idea what to do now.

obrakmann:
OK. I suggest you check out my branch then and see how closely it matches what you were going to do, and we'll either go with it if you don't need any more changes, or you can use it as a base.

RoosterDragon:
Do we maybe want to drop the addition of `Influenced` and just take the extra descriptions? Seems silly to delay this for a rewrite that may or may not happen any time soon.

GraionDilach:
Then please take @obrakmann 's polish.

PR #9886, opened 2015-11-04
Fix the C17 delivery plane doing a steep dive when trying to land

penev92:
Yes, it sucks that the land range is hardcoded. Yes, I will unhardcode it. No, not in this PR. It will be in my PR that completely rewrites this trait to enable proper Starport logic (or it already is, I don't remember; if it isn't request in on that PR next time I reopen it).

The frigate's MaximumPitch is of no consequence.
`MaximumPitch: 56` was removed from the TD `^Plane` because only C17 lands anyway.

### DIFF
--- a/OpenRA.Mods.Cnc/Traits/Buildings/ProductionAirdrop.cs
+++ b/OpenRA.Mods.Cnc/Traits/Buildings/ProductionAirdrop.cs
@@ -63,7 +63,7 @@ namespace OpenRA.Mods.Cnc.Traits
 					new FacingInit(64)
 				});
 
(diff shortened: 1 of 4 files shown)


GraionDilach:
Okay, I checked it out and now I know what you mean under odd. :+1:

RoosterDragon:
:+1:

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/fb0c5229ebb735d7e422d8e5c6719aa287ce51a3)

PR #9887, opened 2015-11-04
Move blossom trees south one cell on cnc64gdi01.

ghost:
ame_21 46 02

Pretty sure this is a misplaced blossom tree on the TD GDI N64 mission.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -44,6 +44,7 @@ Also thanks to:
     * clem
     * Cody Brittain (Generalcamo)
     * D2k Sardaukar
+    * D'Arcy Rush (r34ch)
     * Daniel Derejvanik (Harisson)
     * Danny Keary (Dan9550)
     * David Jiménez (Rydra)
--- a/mods/cnc/maps/cnc64gdi01/map.yaml
+++ b/mods/cnc/maps/cnc64gdi01/map.yaml
@@ -293,7 +293,7 @@ Actors:
 		Location: 38,27
 		Owner: Neutral
 	Actor77: split3
-		Location: 31,27
+		Location: 31,28
 		Owner: Neutral
 	Actor78: tc01
 		Location: 29,27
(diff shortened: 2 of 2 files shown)


GraionDilach:
https://youtu.be/z-SAWwI4Yts?t=20m45s It was there.

r34ch:
...Maybe misplaced by the original N64 devs? I'll let people decide in that case.

pchote:
Why not shift it south by a cell or two?

GraionDilach:
I favor the shifting it down a cell option as well.

r34ch:
Shifted down one cell and updated commit message.

r34ch:
A more logical edit; thanks @GraionDilach 

(I shouldn't file PRs when tired..)

GraionDilach:
:+1:

pchote:
It looks like all of the tib trees need to move one cell south on that map, not just that one.  Can anyone else confirm?

reaperrr:
> It looks like all of the tib trees need to move one cell south on that map, not just that one. Can anyone else confirm?

Yes. This was caused by my foundation change from 1x2 `_ x` to 1x1 waaay back when I didn't know how to write and run proper upgrade rules.

There might be a few newer maps where this is already corrected, though, so we need to be careful about which maps to update.

r34ch:
All blossom trees have been moved south one cell.

obrakmann:
When looking at the map in the editor, the shifted locations become really apparent. :+1:

Mailaender:
Looks like you forgot to setup your e-Mail address in your git client or you are using a different one on your GitHub account which causes the commits to not match.

pchote:
Looks fine to me, too.  We talked in IRC about how to associate additional emails with @r34ch's github account.

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/8bd549cc625b82cedfd62f0b66596b8b83e553a2

PR #9888, opened 2015-11-04
Improve D2k spiceblooms

reaperrr:
This adresses most remaining issues left by #8987.
- slower and more random growth rate
- spice 'debris' no longer fly to the same cell
- much more sane projectile angles
- crushing actors are now damaged properly
- `SpiceExplosion` now has a visible projectile. It had a darker one in the original, but I haven't found the image yet, so I'm using `120mm` for now
- make sure that at least the minimum of pieces is spawned and slightly tweak defaults

### DIFF
--- a/OpenRA.Mods.D2k/Traits/SpiceBloom.cs
+++ b/OpenRA.Mods.D2k/Traits/SpiceBloom.cs
@@ -91,21 +91,6 @@ namespace OpenRA.Mods.D2k.Traits
 
 		public void Killed(Actor self, AttackInfo e)
 		{
-			var args = new ProjectileArgs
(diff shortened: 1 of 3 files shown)


obrakmann:
Works very well now. :+1:

Mailaender:
Don't quite understand yet why, but indeed it works well. Great job!

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/bda56db0bc84634e1826e876970f53ead6d8fc4a...f5f4dbc26a9893187e3f2f10003fa9ab96c3d568)

PR #9890, opened 2015-11-05
Target type support for CreateEffectWarhead

### LINKED ISSUE #9815: Attacking an invulnerable unit should display an animation/play a sound
In RA2, attacking an iron curtained unit displays an animation/plays a sound to indicate that it is invulnerable.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27849475-attacking-an-invulnerable-unit-should-display-an-animation-play-a-sound?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

atlimit8:
Adds `TargetHit` and `TargetTerrain` `ImpactType`s to `CreateEffectWarhead` that use `ValidTargets` and `InvalidTargets`

The test commit adds a noticeable sound when an actor under the Iron Curtain effect is hit.

@scshunt and @GraionDilach does this fix #9815?

### DIFF
--- a/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
@@ -35,15 +35,19 @@ namespace OpenRA.Mods.Common.Warheads
 		[Desc("What impact types should this effect NOT apply to.", "Overrides ValidImpactTypes.")]
 		public readonly ImpactType InvalidImpactTypes = ImpactType.None;
 
-		public static ImpactType GetImpactType(World world, CPos cell, WPos pos)
+		public ImpactType GetImpactType(World world, CPos cell, WPos pos, Actor firedBy)
 		{
 			// Missiles need a margin because they sometimes explode a little above ground
 			// due to their explosion check triggering slightly too early (because of CloseEnough).
 			// TODO: Base ImpactType on target altitude instead of explosion altitude.
 			var airMargin = new WDist(128);
 
+			// Matching target actor
+			if (ValidImpactTypes.HasFlag(ImpactType.TargetHit) && GetDirectHit(world, cell, pos, firedBy, true))
+				return ImpactType.TargetHit;
+
 			var dat = world.Map.DistanceAboveTerrain(pos);
-			var isDirectHit = GetDirectHit(world, cell, pos);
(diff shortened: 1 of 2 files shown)


GraionDilach:
The splash sound was a bad choice - you could've took out something from TD zounds.mix (since I guess that's only for test commit). For a while I was completely uncertain if it works at all due to that. But after a check with RA2 IC hit sound, I can confirm this fixes the issue.

:+1:

atlimit8:
Yea, I know it is not the best sound, but it is just for testing.
@GraionDilach if you can point out a better sound (in the repo) for testing, I wouldn't mind using it in the test commit to make testing easier.

GraionDilach:
`toney10.aud` should sound out.

atlimit8:
@GraionDilach that sound is easier to notice, thanks.

Mailaender:
Works as promised, but I assume the sound effect wants to be removed.

GraionDilach:
My :+1: stood for the logic. I also assume that the testcase was only for the feature to be tested and not intended to be merged.

OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
```diff
@@ -35,36 +35,46 @@ public class CreateEffectWarhead : Warhead
...
-			else if (isWater && !isDirectHit)
-				return ImpactType.Water;
-			else if (isAir && isDirectHit)
-				return ImpactType.AirHit;
-			else if (isWater && isDirectHit)
-				return ImpactType.WaterHit;
-			else if (isDirectHit)
+			var isDirectHit = GetDirectHit(world, cell, pos, firedBy);
+
+			// Air
```
Phrohdoh: The return values explain this just fine.

atlimit8:
> I assume the sound effect wants to be removed.
Yup and done.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c9fd827198d8545c5c99dab2b56eabcad327079a...5aa1800562d02d2e5424c476f69b1f99225f403a)

PR #9893, opened 2015-11-05
Polish the Tiberian Sun Construction Yard and MCV deployment

teees:
- Use the undeploy cursor for the construction yard
- Use the correct deploy sound for the mcv


Mailaender:
:+1:

atlimit8:
Since the ts mod is still an unreleased developmental mod, no changelog entry was added.

PR #9900, opened 2015-11-05
Animations tick after first render

### LINKED ISSUE #9243: Fix first sequence frame of animations not showing when Tick is 40 or lower
As mentioned in #7483:

> This is a high-level design flaw rather than a bug (...).
> The problem is that traits like WithMuzzleFlash unconditionally call Animation.Tick from their own tick method, and this gets called once before the first render frame.
> The way to fix that will be to add if (!firstTick) checks, which will require some additional state tracking in some cases.

atlimit8:
Closes #9243.
Someone better at noticing the glitches will have to check.

### DIFF
--- a/OpenRA.Game/Graphics/Animation.cs
+++ b/OpenRA.Game/Graphics/Animation.cs
@@ -26,6 +26,7 @@ namespace OpenRA.Graphics
 
 		int frame = 0;
 		bool backwards = false;
+		bool rendered = false;
 
 		string name;
 
@@ -54,6 +55,7 @@ namespace OpenRA.Graphics
 
 		public IEnumerable<IRenderable> Render(WPos pos, WVec offset, int zOffset, PaletteReference palette, float scale)
 		{
+			rendered = true;
 			var imageRenderable = new SpriteRenderable(Image, pos, offset, CurrentSequence.ZOffset + zOffset, palette, scale, IsDecoration);
 
 			if (CurrentSequence.ShadowStart >= 0)
@@ -80,6 +82,7 @@ namespace OpenRA.Graphics
 		{
 			backwards = false;
 			tickAlways = false;
+			rendered = false;
 			CurrentSequence = sequenceProvider.GetSequence(name, sequenceName);
 			timeUntilNextFrame = CurrentSequence != null ? CurrentSequence.Tick : defaultTick;
 
@@ -101,6 +104,7 @@ namespace OpenRA.Graphics
 			var tick = CurrentSequence != null ? CurrentSequence.Tick : defaultTick;
 			timeUntilNextFrame = Math.Min(tick, timeUntilNextFrame);
 			frame %= CurrentSequence.Length;
+			rendered = false;
 			return true;
 		}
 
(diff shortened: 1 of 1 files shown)


reaperrr:
I did a quick, preliminary test a few days ago and this seems to solve it. I'll wait with a +1 until I've found time to review this properly, though.

OpenRA.Game/Graphics/Animation.cs
```diff
@@ -63,6 +64,7 @@ public IEnumerable<IRenderable> Render(WPos pos, WVec offset, int zOffset, Palet
 				return new IRenderable[] { shadowRenderable, imageRenderable };
 			}
 
+			rendered = true;
```
RoosterDragon: You need to move this to the top of the function. If there is a shadow, the return above will mean you don't set this in time.

atlimit8: Yea, I should have seen that.

atlimit8:
Fixed.

RoosterDragon:
It feels wrong to change the behaviour of Tick depending on whether Render has been called yet. Relying on the world to be rendered at least once per tick is bad, in my opinion.

atlimit8:
@RoosterDragon that is well beyond the scope of this PR.

atlimit8:
Checks would have to be added to *every* trait that uses `Tick` that way; this prevents that from being a problem (i.e. more robust).

atlimit8:
@RoosterDragon, are you sure the world needs to be rendered every `Tick` in terms of animation. It looks like render and world ticks can be different rates, but world tick updates animation tick keeping them in sync.

RoosterDragon:
We currently force a render for every tick during regular gameplay, see [Game.cs L653](https://github.com/OpenRA/OpenRA/blob/6bd3195c3d5fd7c7f6a46657c50103ad6a7ef92d/OpenRA.Game/Game.cs#L653-L655)

Which brings me to my point, let's say we're a replay and we're not forcing one render per tick. If I tick the world for a second (or more) before rendering, does this PR now mean that animation will be one second behind visually (if that was its first render)?

atlimit8:
What I could do is replace `rendered` with `isFirstTick` and have `Animation.Tick([int t])` return (`bool`) whether is first tick. Then, anything that calls `Animation.Tick([int t])` can decide whether to delay with it, and animations should never fall behind or only by one frame if an after `Action` is not used.

RoosterDragon:
On the other hand, I could be wrong. A fair number of things seem to Tick animations from inside their Render/Draw functions...

pchote:
> It feels wrong to change the behaviour of Tick depending on whether Render has been called yet. Relying on the world to be rendered at least once per tick is bad, in my opinion.

More than just wrong: bogus.  The simulation state *can not* depend on what the rendering is doing.  This is a fundamental design feature of the engine.

atlimit8:
@pchote it is strange that you would mention *simulation state* since that quote is about `Animation.Tick`; if anything `Tick` should be completely divorced from simulation and have `World` keep them in sync.

atlimit8:
> if anything Tick should be completely divorced from simulation and have World keep them in sync.

This would mean removing calls to Tick from all traits, effects, etc. In the long run it might make it easier to have different animation and simulation tick rates, but bugs in such replacement code might cause desyncs.

pchote:
Animations are actually (very limited) state machines that can and do have simulation logic attached in many important places.  They also just happen to have artwork associated with them, which can be fetched by calling their Render method.

We do have different animation and simulation rates (in a rather rudimentary and buggy way) - see the Tick parameter of sequences.

pchote:

PR #9905, opened 2015-11-06
Limit airstrike on cnc64gdi01.

### LINKED ISSUE #9903: cnc64gdi01 air-strike support power
ION cannon has been disabled (probably to eliminate the easy one shot kill on the bio lab objective) but the air-strike support power is granted upon building a comms centre which is still a one shot kill on the bio lab. 

Might be same in the original but the oRA balance of the air strike makes this mission easy to complete with a couple orcas and a comms centre.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28022906-cnc64gdi01-air-strike-support-power?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ghost:
Remove airstrike support power from comms centre to match removal of ION cannon.

Resolves #9903.

EDIT: Actually looking at #9118 there seem to be varying views on the airstrike. Again, I'll let everyone discuss.

### DIFF
--- a/mods/cnc/maps/cnc64gdi01/map.yaml
+++ b/mods/cnc/maps/cnc64gdi01/map.yaml
@@ -809,19 +809,17 @@ Rules:
 		AirstrikePower:
 			Icon: airstrike
 			StartFullyCharged: True
-			ChargeTime: 120
 			SquadSize: 3
 			QuantizedFacings: 8
 			Description: Air Strike
 			LongDesc: Deploy an aerial napalm strike.\nBurns buildings and infantry along a line.
+			OneShot: True
 			EndChargeSound: airredy1.aud
 			SelectTargetSound: select1.aud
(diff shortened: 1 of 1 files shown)


clemty:
I'm all for limiting the airstrike, but the commit looks as if it'd be permanently disabled after the first attack (I cannot test it at the moment), and that is inconsistent with how the tech levels were set up in the original TD: with that change you can still build advanced units like Mammoth Tanks, but the fairly basic airstrike is not available even after the SAMs are destroyed

r34ch:
Closing for now.

PR #9909, opened 2015-11-06
Change engineVersion for several upgrade rules to 20150920

### LINKED ISSUE #9894: RenderBuilding and Targetable upgrade rules have the wrong dates
From http://ppmforums.com/viewtopic.php?p=537686

>  Doing an upgrade from 20150919 resulted in Targetable stuff not being fixed (because the code reads if (engineVersion < 20150902) to change that stuff). Had to fool the upgrader by telling it the engine version was 20150901 
>
>  I also had to do RenderBuilding related traits manually. 

Both of these changes have engine dates < 20150919 but weren't shipped in the 20150919 release.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28001397-renderbuilding-and-targetable-upgrade-rules-have-the-wrong-dates?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Those changes were merged after the feature-freeze on 20150808, but predate the last stable release (20150919). Changing the engine version to the day after release makes it easier for modders to upgrade without using fake engine dates.

Also fixes upgrade rule for WithRotor rename to account for @ separator.

Closes #9894.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -1795,7 +1795,7 @@ namespace OpenRA.Mods.Common.UtilityCommands
 
 				// SpawnViceroid was replaced by SpawnActorOnDeath
 				// And LeavesHusk was renamed to SpawnActorOnDeath
-				if (engineVersion < 20150809)
+				if (engineVersion < 20150920)
 				{
 					if (node.Key == "SpawnViceroid")
 					{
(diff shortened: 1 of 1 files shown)


obrakmann:
:+1: 
I think that's the most sensible option for now. In the long run we need a more clever solution to that problem, though.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/de537f8b821aa7b46eda1233dcdaaf308608e7fe)

PR #9541, opened 2015-10-04
Rename WithTurret, WithBarrel and WithReloadingTurret to include 'Sprite' in trait name

reaperrr:
To make 100% clear to any modder that these are render traits (not to be confused with `Turreted`, for example) as well as for more consistent naming.

Additionally pulled out `ZOffsetFromCenter` from former `WithTurret` into a new `RenderUtils`, since there's quite a number of traits using it by now.

PR #9910, opened 2015-11-06
Switch WithSpriteRotorOverlay to WithRotorSpriteOverlay

reaperrr:
In accordance with other render traits and suggestion in #9541.

Note: While not strictly necessary for this release, it would spare modders from another rename with Next+1 if this went into prep-1512.


GraionDilach:
Upgrade rule?

pchote:
I think it may be better to wait so that mods written for these playtests remain compatible with the final release.

pchote:
@Phrohdoh raised an interesting point in IRC: This trait is redundant now that we have upgrades for flying vs landed aircraft.

reaperrr:
Oh right, we could use `WithIdleOverlay`. I'll close this for now, might revive it later to remove the trait instead.

PR #9913, opened 2015-11-07
Added D2k 7000 start cash option

OmegaBolt:
Added the 7k cash option from the original game (it was the max amount).

Could do a 14k / 15k option too but I don't know if there's much point.

### DIFF
--- a/mods/d2k/rules/player.yaml
+++ b/mods/d2k/rules/player.yaml
@@ -54,6 +54,7 @@ Player:
 		AdviceInterval: 650
 	AllyRepair:
 	PlayerResources:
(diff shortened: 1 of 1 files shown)


penev92:
:+1: 

Original went from 2500 to 7000. Kind of makes you wonder if there is a point in having the higher values, as the current 5000 we use is enough anyway.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/85dd57f2a829d15e7194c675adfc49b2b2c134ca...c2ee2db37faae4e5a1848c41ab50c14646a1c723)

PR #9918, opened 2015-11-07
Fix an oversight regarding building unlocking

### LINKED ISSUE #6683: AI doesn't check bounds before deploying
mcvroom

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5102796-ai-doesn-t-check-bounds-before-deploying?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #9902: deploy glitch
some

1. select mcv
2. rightclick on a cell near a cliff (or whatever)
3. while the mcv is moving to this position ... shift-rightclick on the mcv and wait for it to deploy

some2

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28016626-deploy-glitch?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

penev92:
Fix an oversight regarding building unlocking

### DIFF
--- a/OpenRA.Mods.Common/Traits/Buildings/Building.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/Building.cs
@@ -199,7 +199,7 @@ namespace OpenRA.Mods.Common.Traits
 				return;
 
 			BuildComplete = true;
-			Locked = false;
+			Unlock();
 
 			foreach (var notify in self.TraitsImplementing<INotifyBuildComplete>())
 				notify.BuildingComplete(self);


OpenRA.Mods.Common/Traits/Transforms.cs
```diff
@@ -127,11 +128,21 @@ public void DeployTransform(bool queued)
 				Faction = faction
 			};
 
+			var checkAndTransform = new CallFunc(() =>
+			{
+				// We locked it at the start of this method, so we need to unlock it now.
+				if (building != null)
+					building.Unlock();
```
pchote: This leaves the actor unlocked for a tick, which opens us up to a bunch of edge cases (e.g. selling / capturing the conyard immediately before it turns into an MCV).

You don't need to unlock the actor because it gets discarded when the transform succeeds.  I would set this up like:

```
queue(turn)
queue(lock)
queue(play unmake animation)
queue(check footprint, and transform if allowed)
queue(make animation)
queue(unlock)
```

If the transform succeeds then the activities after that won't run.  If the transform fails then it will restore itself to a valid state.

penev92:
Updated.

OpenRA.Mods.Common/Traits/Transforms.cs
```diff
@@ -127,11 +132,23 @@ public void DeployTransform(bool queued)
...
+					self.QueueActivity(transform);
+				else
+				{
+					// Pass null for the second parameter so we don't issue pointless BuildingComplete notifications.
+					if (makeAnimation != null)
+						makeAnimation.Forward(self, null);
+				}
+			});
+
 			if (makeAnimation != null)
```
pchote: As mentioned in the earlier discussion you also want to check this *before* running the make animation to filter out the cases that will always fail (e.g. trying to deploy next to invalid cells).

penev92: That only happens for unmake animations, and only if you unmake from a building to an actor that is bigger than that building. Make animations aren't played by  `Transforms`, just unmake animations.

OpenRA.Mods.Common/Traits/Render/WithMakeAnimation.cs
```diff
@@ -39,12 +39,18 @@ public WithMakeAnimation(ActorInitializer init, WithMakeAnimationInfo info)
 
 			var building = self.TraitOrDefault<Building>();
 			if (building != null && !building.SkipMakeAnimation)
+				Forward(self, building);
+		}
+
+		// TODO: All the Building stuff that uses make animations should go through this class,
```
pchote: Through this method?

OpenRA.Mods.Common/Traits/Transforms.cs
```diff
@@ -96,7 +100,7 @@ public Order IssueOrder(Actor self, IOrderTargeter order, Target target, bool qu
 		public void DeployTransform(bool queued)
 		{
 			var building = self.TraitOrDefault<Building>();
-			if (!CanDeploy() || (building != null && !building.Lock()))
+			if (!CanDeploy(true) || (building != null && !building.Lock()))
```
pchote: This check should only be done if `queued == false`.

pchote: Actually, this requires a bigger change to handle the building locking... so maybe not a good idea for the prep branch.  This trait needs a total rewrite in the near future.

OpenRA.Mods.Common/Traits/Transforms.cs
pchote: We should have a lint test that ensures that assumption remains true (but not necessary for this pr).

obrakmann:
Seems to work. :+1: 

I'm feeling iffy about taking this into stable, though. This code has regressed in a million subtle ways before, and I fear we'll just be trading one bug for another.

pchote:
I have mentioned in IRC and the earlier discussion here that there is a simpler short-term fix that would be more appropriate to merge to prep: keep everything else the same, but include a check for "can deploy right now" before starting the unmake animation / actual transform.

Phrohdoh:
Adding `[PR: Changes requested]` at least for https://github.com/OpenRA/OpenRA/pull/9918#discussion_r44338007, I'm not sure if the above comment is just a statement or asking for the mentioned route to be taken.

Mailaender:
I don't know if this was completely superseded by https://github.com/OpenRA/OpenRA/issues/9987 or is still improving things after a rebase.

penev92:
The main idea behind this was done with the other PR. I still plan to salvage some things from this though.

penev92:
Although the issue this PR was originally intended to fix was already fixed by @pchote, I decided to repurpose it and keep the little fix for something I stumbled on while debugging the original issues.
Removed the second commit, which is now irrelevant and updated the PR's name and description.

abcdefg30:
Well, that doesn't change anything*, but won't hurt either. :+1: 

*`Unlock()` just does `Locked = false;`

penev92:
That's the idea. This is accessing something that it shouldn't.

obrakmann:
:+1:

Phrohdoh:
If this should not be accessing `Locked` it should not be public (even though that wouldn't fix this case).

penev92:
Chances are its setter should indeed be private to avoid such mistakes in the future.

PR #9922, opened 2015-11-07
Validate player color when changing slots in the lobby

### LINKED ISSUE #9830: Color exploit.
#8380 works pretty good in the bleed. But it doesnt fix the spectator exploit. 

You can pick colors that aren't supposed to be allowed (Like the color I always roll with). To choose these colors you can edit the config file and then jump in a game. All you got to do is switch slots and you get your color. Not sure if there is a fix for this but you can watch all of my games and see how I have been doing.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27860563-color-exploit?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

penev92:
Fixes #9830.

### DIFF
--- a/OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
+++ b/OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
@@ -135,6 +135,10 @@ namespace OpenRA.Mods.Common.Server
 
 						client.Slot = s;
 						S.SyncClientToPlayerReference(client, server.MapPlayers.Players[s]);
+
+						var validatedColor = ColorValidator.ValidatePlayerColorAndGetAlternative(server, client.Color, client.Index, conn);
+						client.PreferredColor = client.Color = validatedColor;
+
 						server.SyncLobbyClients();
 						CheckAutoStart(server);
 


obrakmann:
Yup, that fixes it :+1:

OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
```diff
@@ -135,6 +135,10 @@ public bool InterpretCommand(S server, Connection conn, Session.Client client, s
 
 						client.Slot = s;
 						S.SyncClientToPlayerReference(client, server.MapPlayers.Players[s]);
+
+						var validatedColor = ColorValidator.ValidatePlayerColorAndGetAlternative(server, client.Color, client.Index, conn);
+						client.PreferredColor = client.Color = validatedColor;
```
pchote: I don't think you should be changing the PreferredColor?  The purpose of that field is to store the color that the player *would* use if it wasn't rejected by the validator.

penev92: If I don't set it up the color is invalid every time you change slots, so everyone in the lobby get a message that some color is invalid, without even being able to see what and why. Also you get a new color every time you change slots.
Setting PreferredColor means that notification occurs only the first time you change slots.

pchote: Sounds like somebody has broken how this feature is supposed to work, but I guess I can't penalize you for that.

pchote:
Looks legit, and @obrakmann has confirmed the fix, so :+1: from me too.

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/18c5bacc137bb8e57bb4858b580b04047318c948

Issue #9866, opened 2015-11-02
spice blooms appear too fast (and loud)


MicroJOo:
they pop up way too often (right at the start of the game) although they are meant to appear when the spice is gone ... beside that , they are kinda loud ... and you hear em constantly exploding ... i cant remember hearing this in the original all the time 

in the original : 
>so you have a patch of spice ...and once all of it is gone .... those spice blooms appeared ... and they did indeed explode at some time , but the explosions werent that loud ... and you had (atleast for a good amount of time) the chance to destroy them by yourself

i hope this isnt another unfinished feature ...

pchote:
Has this been resolved now?

MicroJOo:
@pchote : its better now

PR #9926, opened 2015-11-07
Further improve D2k spiceblooms

reaperrr:
This implements the last missing aspects of spiceblooms, namely only growing/'spawning' on spice-free sand (unless pre-placed in map editor) and displaying the animation of blowing spice into the air when spawned.

Additionally, spice particles now properly target cells free of spice when possible.

Closes https://github.com/OpenRA/OpenRA/issues/9866.

### DIFF
--- a/OpenRA.Mods.D2k/Traits/SpiceBloom.cs
+++ b/OpenRA.Mods.D2k/Traits/SpiceBloom.cs
@@ -23,6 +23,9 @@ namespace OpenRA.Mods.D2k.Traits
 	[Desc("Seeds resources by explosive eruptions after accumulation times.")]
 	public class SpiceBloomInfo : ITraitInfo, Requires<RenderSpritesInfo>
 	{
+		[ActorReference]
+		public readonly string SpawnActor = "spicebloom.spawnpoint";
+
 		[SequenceReference]
 		public readonly string[] GrowthSequences = { "grow1", "grow2", "grow3" };
 
@@ -159,7 +162,7 @@ namespace OpenRA.Mods.D2k.Traits
 					new FactionInit(self.Owner.Faction.InternalName),
 					new SkipMakeAnimsInit()
 				};
-				self.World.CreateActor(self.Info.Name, td);
+				self.World.CreateActor(info.SpawnActor, td);
 			})));
 		}
 	}
--- a/OpenRA.Mods.D2k/Traits/SpiceBloom.cs
+++ b/OpenRA.Mods.D2k/Traits/SpiceBloom.cs
@@ -39,7 +39,7 @@ namespace OpenRA.Mods.D2k.Traits
 
 		[Desc("The weapon to use for spice creation.")]
 		[WeaponReference]
-		public readonly string Weapon = "SpiceExplosion";
+		public readonly string Weapon = null;
 
 		[Desc("The amount of spice to expel.")]
 		public readonly int[] Pieces = { 2, 12 };
@@ -98,18 +98,19 @@ namespace OpenRA.Mods.D2k.Traits
 			}
 		}
 
-		public void Killed(Actor self, AttackInfo e)
+		void SeedResources(Actor self)
 		{
 			var pieces = self.World.SharedRandom.Next(info.Pieces[0], info.Pieces[1]) * ticks / growTicks;
 			if (pieces < info.Pieces[0])
 				pieces = info.Pieces[0];
 
-			for (var i = 0; pieces > i; i++)
+			var cells = self.World.Map.FindTilesInAnnulus(self.Location, 1, info.Range);
+
+			for (var i = 0; i < pieces; i++)
 			{
-				var cells = OpenRA.Traits.Util.RandomWalk(self.Location, self.World.SharedRandom);
-				var cell = cells.Take(info.Range).SkipWhile(p => resLayer.GetResource(p) == resType && resLayer.IsFull(p)).Cast<CPos?>().RandomOrDefault(self.World.SharedRandom);
+				var cell = cells.SkipWhile(p => resLayer.GetResource(p) == resType && resLayer.IsFull(p)).Cast<CPos?>().RandomOrDefault(self.World.SharedRandom);
 				if (cell == null)
-					cell = cells.Take(info.Range).Random(self.World.SharedRandom);
+					cell = cells.Random(self.World.SharedRandom);
 
 				var args = new ProjectileArgs
 				{
@@ -140,6 +141,12 @@ namespace OpenRA.Mods.D2k.Traits
 					}
 				});
 			}
+		}
+
+		public void Killed(Actor self, AttackInfo e)
+		{
+			if (!string.IsNullOrEmpty(info.Weapon))
+				SeedResources(self);
 
 			self.World.AddFrameEndTask(t => t.Add(new DelayedAction(respawnTicks, () =>
 			{
--- a/OpenRA.Mods.D2k/Traits/SpiceBloom.cs
+++ b/OpenRA.Mods.D2k/Traits/SpiceBloom.cs
@@ -34,6 +34,9 @@ namespace OpenRA.Mods.D2k.Traits
 
 		public readonly string ResourceType = "Spice";
 
+		[Desc("Spice blooms only grow on these terrain types.")]
+		public readonly HashSet<string> GrowthTerrainTypes = new HashSet<string>();
+
 		[Desc("The weapon to use for spice creation.")]
 		[WeaponReference]
 		public readonly string Weapon = "SpiceExplosion";
@@ -78,6 +81,12 @@ namespace OpenRA.Mods.D2k.Traits
 
 		public void Tick(Actor self)
 		{
+			if (!self.World.Map.Contains(self.Location))
+				return;
+
+			if (info.GrowthTerrainTypes.Count > 0 && !info.GrowthTerrainTypes.Contains(self.World.Map.GetTerrainInfo(self.Location).Type))
+				return;
+
 			ticks++;
 
 			if (ticks >= growTicks)
--- a/mods/d2k/weapons.yaml
+++ b/mods/d2k/weapons.yaml
@@ -820,3 +820,12 @@ BloomExplosion:
 		DamageTypes: Prone50Percent, TriggerProne, ExplosionDeath
 		AffectsParent: true
 
+BloomSpawn:
+	Range: 0c1
+	Projectile: Bullet
+		Speed: 1c0
+		Blockable: false
+		Image: null
+	Warhead@1Eff: CreateEffect
+		Explosion: bloomspawn
+
--- a/mods/d2k/rules/arrakis.yaml
+++ b/mods/d2k/rules/arrakis.yaml
@@ -1,3 +1,25 @@
+spicebloom.spawnpoint:
+	HiddenUnderShroud:
+		Type: CenterPosition
+	BodyOrientation:
+		QuantizedFacings: 1
+	RenderSprites:
+		Image: spicebloom
+	SpiceBloom:
+		GrowthTerrainTypes: Sand
+		SpawnActor: spicebloom
+		GrowthSequences: grow0
+		GrowthDelay: 250, 750
+		RespawnDelay: 1, 2
+	Explodes:
+		Weapon: BloomSpawn
+		EmptyWeapon: BloomSpawn
+	Health:
+		HP: 9999
+		Radius: 1
+	Immobile:
+		OccupiesSpace: false
+
 spicebloom:
 	HiddenUnderShroud:
 	BodyOrientation:
--- a/mods/d2k/sequences/misc.yaml
+++ b/mods/d2k/sequences/misc.yaml
@@ -46,6 +46,11 @@ explosion:
 		BlendMode: Alpha
 		Offset: 12, -10
 		Tick: 120
+	bloomspawn: DATA.R8
+		Start: 3980
+		Length: 8
+		Tick: 120
+		Offset: 0, -16
 	corpse: DATA.R8
 		Start: 430
 		Length: 12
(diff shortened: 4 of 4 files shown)


reaperrr:
I intentionally kept this out from my previous PR because this leaves the 'fix' area and goes more into 'feature' territory and might need some polish because it's a little hacky.

Additionally, `var cells` in `SpiceBloom` probably still returns `null` all the time, but I don't know how to fix that.

pchote:
Do you really even want that random walk?  That behaviour is used in the other mods because it nicely simulates random growth from a central tree... in this case you're throwing the resources through the air, so you only really care that it will land in an empty cell within some reasonable circular range.

reaperrr:
Updated.

Mailaender:
Very nice. Great attention to detail. This scales far better than infinite growth. :+1:

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -74,10 +81,18 @@ public SpiceBloom(ActorInitializer init, SpiceBloomInfo info)
 			respawnTicks = self.World.SharedRandom.Next(info.RespawnDelay[0], info.RespawnDelay[1]);
 			growTicks = self.World.SharedRandom.Next(info.GrowthDelay[0], info.GrowthDelay[1]);
 			anim.Animation.Play(info.GrowthSequences[0]);
+
+			spawnCell = self.World.Map.CellContaining(self.CenterPosition);
 		}
 
 		public void Tick(Actor self)
 		{
+			if (!self.World.Map.Contains(spawnCell))
```
pchote: It would be better to use `self.Location` instead of caching `self.World.Map.CellContaining(self.CenterPosition)`.  The latter will break if the actor moves for whatever reason.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -97,10 +112,10 @@ public void Killed(Actor self, AttackInfo e)
 
 			for (var i = 0; pieces > i; i++)
 			{
-				var cells = OpenRA.Traits.Util.RandomWalk(self.Location, self.World.SharedRandom);
-				var cell = cells.Take(info.Range).SkipWhile(p => resLayer.GetResource(p) == resType && resLayer.IsFull(p)).Cast<CPos?>().RandomOrDefault(self.World.SharedRandom);
+				var cells = self.World.Map.FindTilesInAnnulus(self.World.Map.CellContaining(self.CenterPosition), 1, info.Range);
```
pchote: Why change from `self.Location`?

pchote:
We'll want this for prep, right?

penev92:
I guess so.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -34,6 +37,9 @@ public class SpiceBloomInfo : ITraitInfo, Requires<RenderSpritesInfo>
 
 		public readonly string ResourceType = "Spice";
 
+		[Desc("Spice blooms only grow on this terrain type.")]
+		public readonly string GrowthTerrainType = "Sand";
```
penev92: Since we have 1 actor that uses the default and 1 that removes the default, I'd rather we default to "" (or `null`, whichever is appropriate according to our standards) and put "Sand" on the actor that needs it.

mods/d2k/weapons.yaml
```diff
@@ -820,3 +820,12 @@ BloomExplosion:
 		DamageTypes: Prone50Percent, TriggerProne, ExplosionDeath
 		AffectsParent: true
 
+BloomSpawn:
+	Range: 0c1
+	Projectile: Bullet
+		Speed: 1c0
+		Blockable: false
+		Image: null
+	Warhead@3Eff: CreateEffect
```
penev92: This isn't the 3rd warhead on this weapon though.

penev92:
Looks good ingame.
:+1:  with the two nits above.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
reaperrr: Done.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
reaperrr: Probably just misread it as self.CenterPosition originally. Anyway, fixed.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
reaperrr: Done.

mods/d2k/weapons.yaml
reaperrr: Fixed.

reaperrr:
Updated.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -34,6 +37,9 @@ public class SpiceBloomInfo : ITraitInfo, Requires<RenderSpritesInfo>
 
 		public readonly string ResourceType = "Spice";
 
+		[Desc("Spice blooms only grow on this terrain type.")]
+		public readonly string GrowthTerrainType = null;
```
pchote: Can we make this an array instead, for flexibility?

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -97,10 +109,10 @@ public void Killed(Actor self, AttackInfo e)
 
 			for (var i = 0; pieces > i; i++)
 			{
-				var cells = OpenRA.Traits.Util.RandomWalk(self.Location, self.World.SharedRandom);
-				var cell = cells.Take(info.Range).SkipWhile(p => resLayer.GetResource(p) == resType && resLayer.IsFull(p)).Cast<CPos?>().RandomOrDefault(self.World.SharedRandom);
+				var cells = self.World.Map.FindTilesInAnnulus(self.Location, 1, info.Range);
```
pchote: You can pull this and the `SkipWhile` outside the loop to save some perf.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -34,6 +37,9 @@ public class SpiceBloomInfo : ITraitInfo, Requires<RenderSpritesInfo>
 
 		public readonly string ResourceType = "Spice";
 
+		[Desc("Spice blooms only grow on this terrain type.")]
+		public readonly string GrowthTerrainType = null;
+
 		[Desc("The weapon to use for spice creation.")]
 		[WeaponReference]
 		public readonly string Weapon = "SpiceExplosion";
```
pchote: Can we default this to null so that you don't need to override it with an empty entry in the spawnpoint?

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -78,6 +84,12 @@ public SpiceBloom(ActorInitializer init, SpiceBloomInfo info)
 
 		public void Tick(Actor self)
 		{
+			if (!self.World.Map.Contains(self.Location))
+				return;
+
+			if (info.GrowthTerrainTypes.Any() && info.GrowthTerrainTypes.Contains(self.World.Map.GetTerrainInfo(self.Location).Type))
```
RoosterDragon: Checking for `Any` here is redundant - did you mean to check something else? If not you can remove it.

pchote: I think this should be `if (any && !contains)`, right?
We want to exit early if the current terrain type isn't in the set, but only if the set isn't empty.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
atlimit8: `HashSet<string>`

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -91,45 +103,50 @@ public void Tick(Actor self)
 
 		public void Killed(Actor self, AttackInfo e)
 		{
-			var pieces = self.World.SharedRandom.Next(info.Pieces[0], info.Pieces[1]) * ticks / growTicks;
-			if (pieces < info.Pieces[0])
-				pieces = info.Pieces[0];
-
-			for (var i = 0; pieces > i; i++)
+			if (!string.IsNullOrEmpty(info.Weapon))
```
pchote: `if (isnulllorempty) return;` would be cleaner.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -78,6 +84,12 @@ public SpiceBloom(ActorInitializer init, SpiceBloomInfo info)
 
 		public void Tick(Actor self)
 		{
+			if (!self.World.Map.Contains(self.Location))
+				return;
+
+			if (info.GrowthTerrainTypes.Any() && !info.GrowthTerrainTypes.Contains(self.World.Map.GetTerrainInfo(self.Location).Type))
```
RoosterDragon: Minor perf nit: Prefer `.Count > 0` since `Any` needs to allocate and this is called every tick.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -91,45 +103,50 @@ public void Tick(Actor self)
...
-				var cell = cells.Take(info.Range).SkipWhile(p => resLayer.GetResource(p) == resType && resLayer.IsFull(p)).Cast<CPos?>().RandomOrDefault(self.World.SharedRandom);
-				if (cell == null)
-					cell = cells.Take(info.Range).Random(self.World.SharedRandom);
+				var pieces = self.World.SharedRandom.Next(info.Pieces[0], info.Pieces[1]) * ticks / growTicks;
+				if (pieces < info.Pieces[0])
+					pieces = info.Pieces[0];
 
-				var args = new ProjectileArgs
+				var cells = self.World.Map.FindTilesInAnnulus(self.Location, 1, info.Range)
+					.SkipWhile(p => resLayer.GetResource(p) == resType && resLayer.IsFull(p));
```
pchote: This may need to be `ToList`ed in order to get the benefits of it.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
reaperrr: I think you're forgetting the spawning that starts at line 152 and is unrelated to the weapon.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -91,45 +103,50 @@ public void Tick(Actor self)
...
+
+				for (var i = 0; pieces > i; i++)
```
pchote: Style nit: `i < pieces`

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
pchote: Ah, I missed that.  Still, this is a sign that it may want to be pulled out into a SeedResources method?

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
reaperrr: Basically `.Where(p => resLayer.GetResource(p) != resType || !resLayer.IsFull(p)).ToList()`?

GraionDilach: Also see #10062.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
```diff
@@ -91,45 +103,50 @@ public void Tick(Actor self)
...
 				{
-					Weapon = self.World.Map.Rules.Weapons[info.Weapon.ToLowerInvariant()],
-					Facing = 0,
+					var cell = cells.Cast<CPos?>().RandomOrDefault(self.World.SharedRandom);
+					if (cell == null)
+						cell = cells.Random(self.World.SharedRandom);
```
RoosterDragon: What are these three lines trying to do exactly? `RandomOfDefault` will return null only on an empty collection, so repeating it with `Random` will just throw which seems odd.

As written, you could simplify the current logic down to `cell = cells.Random(self.World.SharedRandom)` for the same effect.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
RoosterDragon: `self.World.Map.FindTilesInAnnulus` will only find cells in the world by default, so you don't need the same check as #10062.

GraionDilach: Ah, sorry then.

OpenRA.Mods.D2k/Traits/SpiceBloom.cs
reaperrr: Tbh I didn't (and still don't) fully understand this piece of code from #8987, so I just avoided touching it too much.

RoosterDragon: Looks like the old code was attempting to take a random cell in range but skipping over anything already full of resource. If there were no such cells (everything was already maxed) it just used any random cell in range instead.

Try moving the `SkipWhile` currently on the `cells` variable back onto the first random search line - that ought to make it work like the original code again as best I can tell.

obrakmann:
Looks fine to me. :+1:

Can you please squash those last four commits into the earlier ones before merging? Thanks!

reaperrr:
Squashed and reordered commits.

Note: There were some unavoidable yaml rebase conflicts while squashing. I'm quite sure i caught them all, but I still recommend giving this a last quick look-over before merging.

obrakmann:
Nope, still looks all good, in-game too.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/958c08cff8820a3c01d0c3d0d26cf7b1fa4ab83a)

PR #9927, opened 2015-11-08
Easier actor selection in map editor by actor bounds center

atlimit8:
Selects actor under cursor by how close its bounds center is to the cursor and then by how high its elevation is. This allows for selection (and tooltip information) of actors masked by the "Scrin Ship" in the "Defend Crash Site" shellmap of the ts mod.

### DIFF
--- a/OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
+++ b/OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
@@ -39,6 +39,7 @@ namespace OpenRA.Mods.Common.Widgets
 		readonly EditorViewportControllerWidget editorWidget;
 		readonly EditorActorLayer editorLayer;
 		readonly Dictionary<int, ResourceType> resources;
+		int2 worldPixel;
 
 		public EditorDefaultBrush(EditorViewportControllerWidget editorWidget, WorldRenderer wr)
 		{
@@ -51,6 +52,18 @@ namespace OpenRA.Mods.Common.Widgets
 				.ToDictionary(r => r.Info.ResourceType, r => r);
 		}
 
+		long CalculateActorSelectionPriority(EditorActorPreview actor)
+		{
+			var centerPixel = new int2(actor.Bounds.X, actor.Bounds.Y);
+			var pixelDistance = (centerPixel - worldPixel).Length;
+
+			// If 2+ actors have the same pixel position, then the highest appears on top.
+			var worldZPosition = actor.CenterPosition.Z;
+
+			// Sort by pixel distance then in world z position.
+			return ((long)pixelDistance << 32) + worldZPosition;
+		}
+
 		public bool HandleMouseInput(MouseInput mi)
 		{
 			// Exclusively uses mouse wheel and right mouse buttons, but nothing else
@@ -59,10 +72,10 @@ namespace OpenRA.Mods.Common.Widgets
 				mi.Event == MouseInputEvent.Down)
 				return false;
 
+			worldPixel = worldRenderer.Viewport.ViewToWorldPx(mi.Location);
 			var cell = worldRenderer.Viewport.ViewToWorld(mi.Location);
 
-			var underCursor = editorLayer.PreviewsAt(worldRenderer.Viewport.ViewToWorldPx(mi.Location))
-				.FirstOrDefault();
+			var underCursor = editorLayer.PreviewsAt(worldPixel).MinByOrDefault(CalculateActorSelectionPriority);
 
 			var mapResources = world.Map.MapResources.Value;
 			ResourceType type;


Mailaender:

OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
```diff
@@ -39,6 +39,7 @@ public class EditorDefaultBrush : IEditorBrush
 		readonly EditorViewportControllerWidget editorWidget;
 		readonly EditorActorLayer editorLayer;
 		readonly Dictionary<int, ResourceType> resources;
+		int2 worldLocation;
```
Phrohdoh: This is a screen coordinate.

atlimit8: `worldLocation = worldRenderer.Viewport.ViewToWorldPx(mi.Location)` makes it world view or projection coordinates.
I could rename it `worldViewLocation`.

atlimit8: Renamed `worldPixel for clarity`

OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
```diff
@@ -51,6 +52,15 @@ public EditorDefaultBrush(EditorViewportControllerWidget editorWidget, WorldRend
 				.ToDictionary(r => r.Info.ResourceType, r => r);
 		}
 
+		long CalculateActorSelectionPriority(EditorActorPreview actor)
+		{
+			var center = new int2(actor.Bounds.X, actor.Bounds.Y);
+			long priority = (center - worldPixel).Length;
```
Phrohdoh: Is `var` not usable here?

OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
```diff
@@ -51,6 +52,15 @@ public EditorDefaultBrush(EditorViewportControllerWidget editorWidget, WorldRend
...
+			priority <<= 32;
```
Phrohdoh: This seems entirely arbitrary.

OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
penev92: I believe he wants to enforce long over int because of the shifting by 32 bits that is done on the next line.
Interesting approach though. I think this method can use a comment or two.

OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
atlimit8: Well all it needs to be is enough pixel closeness to dominate world height, `int`s are 32 bits, `long`s are 64 bits, and shifting 32 bits is the standard approach to stuffing two 32-bit words in a 64-bit word.

pchote: I agree that this is standard, however a lot of people are scared of low level bit-wrangling.  I think @penev92's suggestion of adding a short comment is good.

pchote: You could also make it more explicit that you're not doing anything magic by combining it into `return (long)((center - worldPixel).Length) << 32 + actor.CenterPosition.Z;` instead of successively modifying a temporary var.

atlimit8:
Added temporary variables and comments with the return statement having the merging by sorting dominance;

OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
```diff
@@ -59,10 +72,11 @@ public bool HandleMouseInput(MouseInput mi)
 				mi.Event == MouseInputEvent.Down)
 				return false;
 
+			worldPixel = worldRenderer.Viewport.ViewToWorldPx(mi.Location);
 			var cell = worldRenderer.Viewport.ViewToWorld(mi.Location);
 
 			var underCursor = editorLayer.PreviewsAt(worldRenderer.Viewport.ViewToWorldPx(mi.Location))
```
RoosterDragon: You could now use `worldPixel` here, right?

atlimit8: Yup.

atlimit8:
Minor change.

OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
```diff
@@ -51,6 +52,18 @@ public EditorDefaultBrush(EditorViewportControllerWidget editorWidget, WorldRend
 				.ToDictionary(r => r.Info.ResourceType, r => r);
 		}
 
+		long CalculateActorSelectionPriority(EditorActorPreview actor)
+		{
+			var centerPixel = new int2(actor.Bounds.X, actor.Bounds.Y);
+			var pixelDistance = (centerPixel - worldPixel).Length;
+
+			// If 2+ actors have the same pixel position, then the highest appears on top.
+			var worldZPPosition = actor.CenterPosition.Z;
```
RoosterDragon: Possible typo? (Two \`P\`s)

atlimit8: Yup - Oops.

atlimit8:
Fixed typo.

OpenRA.Mods.Common/EditorBrushes/EditorDefaultBrush.cs
```diff
@@ -51,6 +52,18 @@ public EditorDefaultBrush(EditorViewportControllerWidget editorWidget, WorldRend
 				.ToDictionary(r => r.Info.ResourceType, r => r);
 		}
 
+		long CalculateActorSelectionPriority(EditorActorPreview actor)
```
RoosterDragon: I think you should pass in the world pixel as a parameter, like you're doing in #10011 (maybe rename it too, for consistency).

Will also be worth using the shared method you've created in that PR for calculating the priority.

atlimit8: I could do that, but that would:
1) make this depend on #10011.
2) replace visual ordering with in game selection.
Not that these are undesirable.

RoosterDragon: Derp, they're different - my bad. Don't worry about that then.

obrakmann:
Nice, this also helps with infantry that's hiding behind trees in the regular 2d terrain. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/66e2c0458fc7f31a79a30303ff530e12774ad238)

PR #9928, opened 2015-11-08
Fix nod05 airstrike Lua crash and update map rules

ghost:
Don't call an airstrike while the player has no ground units.

### DIFF
--- a/mods/cnc/maps/nod05/map.yaml
+++ b/mods/cnc/maps/nod05/map.yaml
@@ -48,6 +48,11 @@ Players:
 		OwnsWorld: True
 		NonCombatant: True
 		Faction: gdi
+	PlayerReference@Civilians:
+		Name: Civilians
+		NonCombatant: True
+		Faction: gdi
+		Enemies: Nod
 	PlayerReference@Nod:
 		Name: Nod
 		Playable: True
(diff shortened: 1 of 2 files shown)


Phrohdoh:
You could just bail out of the function instead of wrapping its entire purpose in an if.

Biofreak1987:
Updated

obrakmann:
:+1:

Biofreak1987:
Actually... I just notice that nod05 also shows the owner row on husks. Should I fix that with this pr as well or create another one?

obrakmann:
You can do it here

Biofreak1987:
K, added

Phrohdoh:
The commit message no longer covers everything changed in this single commit.

Biofreak1987:
Sorry, fixed that.

Biofreak1987:
By replaying the original mission I've noticed that there are a few other things missing (you can build e3 units and sand bags in the original)... also nod can't use airstrikes in their campaign, should that be then also the case for the ports?

Biofreak1987:
I've ended up adding those changes so that the map ends up being closer to the original version. Details about it are in the commit message.

obrakmann:
FWIW, my :+1: still stands after the additional changes.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/6eeb35be7b58c73b84fd3db5cda2fc75c469bfd2...00ad73a6128e3477dea65aac93c4104b3a55ec8a)

Issue #9430, opened 2015-09-25
UnauthorizedAccessException when saving maps in system folders


airetaM:
my mapeditor crashes when i want to create new map

i will use the old editors meanwhile

here is the exception.log :

Tiberian Dawn Mod at Version release-20150919
on map b0f941e404b8113179ed8c358982f99d3b1021c5 (shellmap by Chris Forbes).
Operating System: Linux (Unix 3.13.0.61)
Runtime Version: Mono 3.2.8 (Debian 3.2.8+dfsg-4ubuntu1.1) CLR 4.0.30319.17020
Exception of type `System.UnauthorizedAccessException`: Access to the path "/usr/lib/openra/mods/cnc/maps/Rainy Forest.oramap" is denied.
  at System.IO.FileStream..ctor (System.String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, Boolean anonymous, FileOptions options) [0x00000] in <filename unknown>:0 
  at System.IO.FileStream..ctor (System.String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize) [0x00000] in <filename unknown>:0 
  at (wrapper remoting-invoke-with-check) System.IO.FileStream:.ctor (string,System.IO.FileMode,System.IO.FileAccess,System.IO.FileShare,int)
  at System.IO.File.Create (System.String path, Int32 bufferSize) [0x00000] in <filename unknown>:0 
  at System.IO.File.Create (System.String path) [0x00000] in <filename unknown>:0 
  at ICSharpCode.SharpZipLib.Zip.ZipFile.Create (System.String fileName) [0x00000] in <filename unknown>:0 
  at OpenRA.FileSystem.ZipFile..ctor (System.String filename, Int32 priority, System.Collections.Generic.Dictionary`2 contents) [0x00000] in <filename unknown>:0 
  at OpenRA.FileSystem.GlobalFileSystem.CreatePackage (System.String filename, Int32 order, System.Collections.Generic.Dictionary`2 content) [0x00000] in <filename unknown>:0 
  at OpenRA.Map.Save (System.String toPath) [0x00000] in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.Logic.SaveMapLogic+<SaveMapLogic>c__AnonStorey0.<>m__2 () [0x00000] in <filename unknown>:0 
[...]

PR #9929, opened 2015-11-08
Set the default map location to the support dir.

pchote:
Fixes the worst part of #9430 so that it does not block the next release.

That ticket will not be properly resolved until we catch other errors during map save, but that is out of scope for this PR.


obrakmann:
:+1:

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/ca74aef5b56819d01cf5e78501a46a20bb6f641e)

PR #9933, opened 2015-11-08
Prevent frozen actors leaking through shroud.

### LINKED ISSUE #9930: Attack cursor can be visible for a frozen actor under the shroud
Reported by @penev92 in IRC.

Explained in diff form:
```diff
diff --git a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
index 260f502..f4ed2f6 100644
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -97,6 +97,10 @@ void UpdateVisibility()
                                }
                        }
 
+                       // TODO: If Visible == true but the whole footprint is covered by shroud
+                       // (e.g. from a gap generator, hide-map crate, or debug cheat) we need to set Visible = false.
+                       // If we don't do this the player will see an attack cursor through the shroud.
+
                        if (Visible && !wasVisible)
                                NeedRenderables = true;
                }
```

It isn't immediately obvious to me how to fix this without hurting perf.  The brute-force fix (which we may have to go for) is to loop over the footprint a second time and check against `shroud.IsExplored`.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28070445-attack-cursor-can-be-visible-for-a-frozen-actor-under-the-shroud?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #9930.

### DIFF
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -41,6 +41,7 @@ namespace OpenRA.Traits
 		public DamageState DamageState;
 
 		public bool Visible = true;
+		public bool Shrouded { get; private set; }
 		public bool NeedRenderables { get; private set; }
 		public bool IsRendering { get; private set; }
 
@@ -84,6 +85,7 @@ namespace OpenRA.Traits
 		void UpdateVisibility()
 		{
 			var wasVisible = Visible;
+			Shrouded = true;
 
 			// We are doing the following LINQ manually for performance since this is a hot path.
 			// Visible = !Footprint.Any(shroud.IsVisible);
@@ -93,8 +95,12 @@ namespace OpenRA.Traits
 				if (shroud.IsVisible(puv))
 				{
 					Visible = false;
+					Shrouded = false;
 					break;
 				}
+
+				if (Shrouded && shroud.IsExplored(puv))
+					Shrouded = false;
 			}
 
 			if (Visible && !wasVisible)
@@ -119,6 +125,9 @@ namespace OpenRA.Traits
 				}
 			}
 
+			if (Shrouded)
+				return NoRenderables;
+
 			if (flashTicks > 0 && flashTicks % 2 == 0)
 			{
 				var highlight = wr.Palette("highlight");
@@ -129,7 +138,7 @@ namespace OpenRA.Traits
 			return renderables;
 		}
 
-		public bool HasRenderables { get { return renderables.Any(); } }
+		public bool HasRenderables { get { return !Shrouded && renderables.Any(); } }
 
 		public bool ShouldBeRemoved(Player owner)
 		{
@@ -175,6 +184,8 @@ namespace OpenRA.Traits
 			VisibilityHash = 0;
 			FrozenHash = 0;
 
+			// TODO: Track shroud updates using Shroud.CellsChanged
+			// and then only tick FrozenActors that might have changed
 			foreach (var kvp in frozen)
 			{
 				var hash = (int)kvp.Key;


obrakmann:
Works :+1:

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -175,6 +184,8 @@ public void Tick(Actor self)
 			VisibilityHash = 0;
 			FrozenHash = 0;
 
+			// TODO: Track shroud updates using Shroud.CellsChanged
+			// and then only tick FrozenActors that might have changed
```
RoosterDragon: We have issues for this already, this comment is redundant.

penev92:
:+1:

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -119,6 +125,9 @@ public IEnumerable<IRenderable> Render(WorldRenderer wr)
 				}
 			}
 
+			if (Shrouded)
+				return NoRenderables;
```
RoosterDragon: `FrozenActorLayer.Render` currently deals with filtering out invisible actors, I think you should move this check there instead.

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -129,7 +138,7 @@ public IEnumerable<IRenderable> Render(WorldRenderer wr)
 			return renderables;
 		}
 
-		public bool HasRenderables { get { return renderables.Any(); } }
+		public bool HasRenderables { get { return !Shrouded && renderables.Any(); } }
```
RoosterDragon: This check shouldn't be here. It should be up to the caller. i.e. UnitOrderGenerator should make the check (since it's already checking both Visible and HasRenderables anyway).

RoosterDragon:
`HasRenderables` is also used by the GpsDot effect. I'm not sure if that could potentially cause any side effects there but that seems strictly beyond the scope of the fix for #9930 so I thought I'd mention it.

RoosterDragon:
GPS seems fine, so the rest of my comments are basically are just style issues - nothing major ultimately.

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
pchote: This is actually a workaround for `FrozenUnderFog`, which will otherwise treat the actor as completely visible(!).

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
pchote: Likewise here - if we don't do this the actor will be treated the same as if it were not shrouded or fogged.

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
RoosterDragon: Unless I'm going crazy, the only method that calls this is `FrozenActorLayer`, which is where I think the check should be moved to.

Where does `FrozenUnderFog` call this?

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
RoosterDragon: I'm suggesting the check should be moved to the caller, not that it shouldn't happen.

PR #9934, opened 2015-11-08
D2k super weapon timers

OmegaBolt:
Adjusted the SW timers that makes them correctly relative to one another.

I simply timed them with a stopwatch, then multiplied them to hopefully match our game speed. They shouldn't be wildly off, and are at least more accurate than what is currently in place.

### DIFF
--- a/mods/d2k/rules/structures.yaml
+++ b/mods/d2k/rules/structures.yaml
@@ -721,7 +721,7 @@ high_tech_factory:
 		Icon: ornistrike
 		Description: Air Strike
 		Prerequisites: ~techlevel.superweapons, upgrade.hightech
-		ChargeTime: 180
+		ChargeTime: 300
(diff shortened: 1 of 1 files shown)


MustaphaTR:
IIRC, Saboteur builds quicker than fremen but i'm not sure.

OmegaBolt:
If they do Mustapha, it's within seconds because they came up as the same time, give or take the speed of my finger to stop the clock. It seemed to me that Fremen & Saboteur are the same and Ornithopters & Death Hand are the same.

obrakmann:
OK, I guess. :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/5fc595725c9e55dca00401f4e0f03d3fd784416b)

PR #9937, opened 2015-11-09
Changed ServerSettings cloning to a more reliable approach

hacklex:
Using your own copying constructor in classes without proper maps (e.g. ObjectToObject in BLToolkit) is a dangerous practice. As shown in this commit, one might easily forget to update the copying constructor with recently added fields (for example, TimestampFormat was not copied before this commit).

The more general solution (for simple cases like this one) is a call to a protected System.Object method MemberwiseClone(), which does exactly what its name says.

Actually, you may consider using a struct for cases like this. Structs are always copied on assignment; it takes a lot of trickery to actually modify (not replace) someone else's struct, and it is virtually impossible if it is published via a readonly property. But, since structs are all ValueTypes (which implies they are allocated on stack), it is not recommended to use them without fully understanding their nature.

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -358,7 +358,7 @@ namespace OpenRA
 			{
 				Settings.Server.Map = WidgetUtils.ChooseInitialMap(Settings.Server.Map);
 				Settings.Save();
-				CreateServer(new ServerSettings(Settings.Server));
+				CreateServer(Settings.Server.Clone());
 
 				while (true)
 				{
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -74,27 +74,9 @@ namespace OpenRA
 
 		public string TimestampFormat = "s";
 
-		public ServerSettings() { }
-
-		public ServerSettings(ServerSettings other)
+		public ServerSettings Clone()
 		{
-			Name = other.Name;
-			ListenPort = other.ListenPort;
-			ExternalPort = other.ExternalPort;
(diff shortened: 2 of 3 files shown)


atlimit8:
Code format error - https://travis-ci.org/OpenRA/OpenRA/builds/90010925#L1050:
> OpenRA.Game/Settings.cs:L81: [SA1009] Invalid spacing around the closing parenthesis.

It should be `return (ServerSettings) MemberwiseClone();`.

hacklex:
Fixed code style. Sorry guys, never committed here before.

atlimit8:
No problem. One more note, it would be preferable for the code style fix commit to be squashed so that it does not go into the main branch history. If you don't know how to squash commits, just ask. Thanks for contributing.

hacklex:
No problems with squashing. Done.

RoosterDragon:
You should remove the copying constructor - it's dead code now you've added the `Clone` method.

Don't forget to add yourself to the [AUTHORS](https://github.com/OpenRA/OpenRA/blob/bleed/AUTHORS) file, if you like.

hacklex:
The details on the above checks are strange... Can these checks be run again? I believe in the PR nothing is wrong...

penev92:
Don't worry about that. Travis is apparently experiencing technical difficulties today. It's not you.

RoosterDragon:
:+1:

abcdefg30:
> OpenRA.Game/Settings.cs:L76: [SP2000] Invalid spacing at the end of the line.

abcdefg30:
:+1: / :white_check_mark: with stylecop fixed.

hacklex:
Done

PR #9942, opened 2015-11-09
Fix the documentation of Explodes

abcdefg30:
The `DeathType` defines which weapon(s) trigger an explosion.
Also renamed it to plural, as we use it elsewhere (it was already a `HashSet<string>`, too).


penev92:
Seems legit :+1:

Phrohdoh:
@abcdefg30 Could you clarify the commit messages a bit so they have some context?
After that :+1:

abcdefg30:
Updated. Hope it's ok now.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9783f904b8e3d21a3f39f8fa0d3ce170430fda39)

PR #9945, opened 2015-11-09
Adjust orca sight range and prevent aircraft from flying into shroud

teees:


mods/ts/rules/aircraft.yaml
```diff
@@ -78,12 +78,13 @@ ORCA:
...
 		ROT: 5
 		Speed: 186
+		MoveIntoShroud: false
 	Health:
 		HP: 200
 	Armor:
 		Type: Light
 	RevealsShroud:
-		Range: 10c0
+		Range: 2c0
```
pchote: I expect that this will make it possible to creep them into the shroud.  If we're going to have `MoveIntoShroud: false` to match the original game, then we should remove the sight range to also match the original game.

teees: iirc it was possible to creep/micro them into shroud, just a bit tricky

teees: see https://www.youtube.com/watch?v=djhoinPs5f0&t=8m35s
I took that sight range from the original rules.ini

RoosterDragon:
Needs a rebase.

teees:
rebased

mods/ts/rules/aircraft.yaml
```diff
@@ -238,6 +241,7 @@ SCRIN:
 		ROT: 3
 		Speed: 168
 		AirborneUpgrades: airborne
+		MoveIntoShroud: false
```
pchote: Did the original banshee have the FlyCircle behaviour that ours does?  That negates the effectiveness of MoveIntoShroud a fair bit.

pchote: Likewise for the orca bomber.

pchote:
Looks fine to me :+1:.
You may also want to adjust the orca transport for consistency, even though that is only buildable with cheats.

mods/ts/rules/aircraft.yaml
penev92: They just landed on the ground upon becoming idle.

teees: iirc all air units handled pretty much the same except for attack behavior. They all landed when idle. I also noticed that with vision that low and fog of war the banshees attack move is nearly useless, since he can't turn fast enough to actually attack any unit after noticing it

teees:
I adjusted the drop pod sight range to 0, all other aircraft should math the original rules.ini

GraionDilach:
Beware! TS `Sight=2` does not equal to our `2c0`! Using the same values will result with a √2* bigger ranges (this is something I want to recalculate for a while now) due to the fact that OpenRA 1c0 is equal to a cell's diagonal, compared to WW cell measurement based on sides.

See http://i.imgur.com/qoiPx2f.png?1 for a visual showcase.

pchote:
@GraionDilach: note that this is a problem for *all* cell ranges, so I would be very hesitant to start changing arbitrary things by sqrt(2).  We will build a general solution instead.

GraionDilach:
I thought this is by design.

pchote:
It was, but we later realized that it had a rather sad consequence for the mod rules.  My current thought / plan is to have a new syntax (e.g. `3d512`) to specify diagonal distances, which are then treated as I mentioned just above when parsing the WDists from yaml.

UberWaffe:
After a quick playtest, :-1:
Aircraft seem to struggle to attack targets on their own.
Also, due to how the sight range works for aircraft, you can 'crawl' up, left and right, but not down.
And most of all, why do we want this? This *feels* broken.

GraionDilach:
We want this because this is part of the TS gameplay and until TS ain't recreated entirely, no balance changes should be made towards altering it. Even if the result might look broken.

UberWaffe:
Allright, I understand.
The problem with sight for aircraft not being equal in all directions is an issue with such short sight ranges. I'd say this is dependent on that being fixed first.

Though I still feel that while we maintain 'Fog of War' being enabled by default, matching sight ranges exactly to the original is wasted effort.

Mailaender:
Indeed a terrible mechanic from the original, but we can keep it as a test case here as long as this isn't released.

PR #9946, opened 2015-11-09
D2k Battle For Dune map update

OmegaBolt:
This just fixes a few mistakes I made, where I used cliffs that looked like rock-sand transitions instead of the correct tiles, and so were awkwardly unpassable.


obrakmann:
Very minor differences indeed. :+1:

obrakmann:
This map on this branch is already on map format 8. Since it's a binary file, I won't backport this to prep. The changes are minimal and not game-breaking, so I guess we'll all live.

PR #9948, opened 2015-11-09
D2k thumper worm attraction balance change

OmegaBolt:
Originally suggested by pchote, this gives the Thumper an inner ring where the worm is allowed to move freely around the Thumper. Makes them much more useful, as you can essentially "trap" a worm in an area but still allow them to attack other objects like enemy Harvesters etc.

By default it AttractWorm should have a Spread of 3072, making the inner circle 9 cells in total. (?)

The exact range of this effect could be argued about, but for now this is a big improvement.

### DIFF
--- a/mods/d2k/rules/infantry.yaml
+++ b/mods/d2k/rules/infantry.yaml
@@ -116,6 +116,7 @@ thumper:
 		UpgradeMinEnabledLevel: 1
 	AttractsWorms:
 		Intensity: 1000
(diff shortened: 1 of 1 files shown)


obrakmann:
Seems ok. :+1:

penev92:
Worms seem as retarded as desired, I guess :+1: (when there is a thumper in the vicinity).

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/e2690d194df26c43f861b27f689a408a25b93ddf)

PR #9951, opened 2015-11-10
Add a comment explaining some of the sequence parsing code.

phrohdoh:
I'm fine with closing this (or better, rewording!) if no one feels this is necessary, but it has confused me multiple times.

### DIFF
--- a/OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
+++ b/OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
@@ -37,6 +37,15 @@ namespace OpenRA.Mods.Common.Graphics
 				nodes.Remove("Defaults");
 				nodes = nodes.ToDictionary(kv => kv.Key, kv => MiniYaml.MergeStrict(kv.Value, defaults));
 
+				// Merge 'Defaults' animation image value. An example follows.
+				//
+				// - Before -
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
```diff
@@ -37,6 +37,20 @@ public virtual ISpriteSequence CreateSequence(ModData modData, TileSet tileSet,
 				nodes.Remove("Defaults");
 				nodes = nodes.ToDictionary(kv => kv.Key, kv => MiniYaml.MergeStrict(kv.Value, defaults));
 
+				// Merge default animation image value. An example follows.
+				// Note: This example isn't entirely true as we remove the 'Defaults' section above.
```
pchote: I would prefer it if we could use an example that was entirely true.  Could you fix the after case to reflect what it actually does?

Phrohdoh: Sure, I'll have to move it above the `MiniYaml.MergeStrict`.

Phrohdoh:
Updated the comment.

abcdefg30:
Well, simple enough. :+1:

obrakmann:
:+1:

PR #9959, opened 2015-11-10
Ra balance

ghost:
Erred on the side of caution with these changes but aimed to make each unit more unique. If they need buffed further / completely changed, feel free. The Tesla tank still needs its double zap.

General reasoning;
-Chronotank is now more viable at repetitive hit and run.
-Phase transport is better for flanking and deploying troops. LoS out-ranges a turret but not a coil.
-Tesla tank can now (soon to be burst) attack targets via drive-by and whilst retreating. 

Chronotank 
-charge time reduced from 20s to 12s 

Phase tank
-passengers increased from 3 to 4
-speed increased from 113 to 142 (matches APC)
-LoS increased from 6 to 7 cells (matches Tesla tank)

Tesla tank
-no longer needs to turn and face the enemy to fire
(Needs double zap)

### DIFF
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -719,10 +719,10 @@ STNK:
 	Armor:
 		Type: Light
 	Mobile:
-		Speed: 113
+		Speed: 142
 		Crushes: wall, mine, crate, infantry
 	RevealsShroud:
-		Range: 6c0
+		Range: 7c0
 	AutoTarget:
 		InitialStance: HoldFire
 		InitialStanceAI: ReturnFire
@@ -735,8 +735,8 @@ STNK:
 	WithTurret:
 	Cargo:
 		Types: Infantry
-		MaxWeight: 3
-		PipCount: 3
+		MaxWeight: 4
+		PipCount: 4
 	Cloak:
 		InitialDelay: 125
 		CloakDelay: 250
@@ -744,7 +744,7 @@ STNK:
 		UncloakSound: appear1.aud
 		IsPlayerPalette: true
 	DetectCloaked:
-		Range: 6c0
+		Range: 7c0
 	-MustBeDestroyed:
 	ProducibleWithLevel:
 		Prerequisites: vehicles.upgraded
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -672,6 +672,7 @@ CTNK:
 		LocalYaw: -100
 	AttackFrontal:
 	PortableChrono:
+		ChargeDelay: 300
 	ProducibleWithLevel:
 		Prerequisites: vehicles.upgraded
 
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -563,7 +563,7 @@ TTNK:
 	Armament:
 		Weapon: TTankZap
 		LocalOffset: 0,0,213
-	AttackFrontal:
+	AttackOmni:
 	WithIdleOverlay@SPINNER:
 		Sequence: spinner
 	SelectionDecorations:


cjshmyr:
There were discussions with active players on IRC that influenced these; I'm in favor of the changes. I think it would be nice to also lower the cost on some of them, but I might be a minority.

r34ch:
My thinking was to try and turn these units from being completely unviable or one trick ponies into powerful but weak units that reward intelligent play. These buffs should _hopefully_ increase their survivability along with their effectiveness, which is why I wanted to  hold off on the price reduction until we get player feedback on these changes.

If testing proves this false; feel free to make changes including dropping the price. As a player though, I'd rather pay a few extra $$$ to gain a unit that opens up otherwise non existent strategies.

r34ch:
Finally figured out a non bogus upgrade rule. 

As to why it has been implemented this way; it was suggested to me on IRC. If this is still the incorrect implementation (or other upgrade rules are needed) please advise.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2377,6 +2377,17 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
 				}
 
+				// Set ChargeDelay to 300 by default
+				if (engineVersion < 20151111)
+				{
+					if (depth == 1 && node.Key == "PortableChrono")
+					{
+						var delay = node.Value.Nodes.FirstOrDefault(n => n.Key == "ChargeDelay");
+						if (delay == null)
+							node.Value.Nodes.Add(new MiniYamlNode("ChargeDelay", "300"));
```
Phrohdoh: But now there is a difference between the engine fallback and the value you're inserting into yaml.

OpenRA.Mods.RA/Traits/PortableChrono.cs
```diff
@@ -21,7 +21,7 @@ namespace OpenRA.Mods.RA.Traits
 	class PortableChronoInfo : ITraitInfo
 	{
 		[Desc("Cooldown in ticks until the unit can teleport.")]
-		public readonly int ChargeDelay = 500;
+		public readonly int ChargeDelay = 0;
```
cjshmyr: Set this to 300, don't rely on the upgrade rule to be ran.

r34ch: I guess I completely misunderstood the initial discussion on all this. Fixing.

r34ch:
Fixed and updated. Thanks all for the input.

abcdefg30:
If you don't change the default value in C#, you don't need an upgrade rule.

r34ch:
Upgrade rule removed, thanks for the info.

penev92:
Sure, why not :+1:

MicroJOo:
so far we didnt receive any response to #9668
but i tell you what i think about these balance-changes ...

buffing the "phase-transporter" is not needed ... the idea behind it , isnt deploying "troops" ... its "sneaking in special units" ... the phase-transporter is also not a "scout" unit ... (so i dont really understand the LoS-Buff) ... the speed also doesnt have to match the soviets APC , cause he is still "invisible" ... 
if anything , we should work out a better solution "detecting" him ... cause all the defense-strucures and radars makes it not really interesting to give it a try (on certain maps) , also ... maybe a buff for the weapon somewhere in the future

the crono and teslatank buff sounds good but ... 

we just recently "nerfed" the shocktroopers ... because people (you) moaned about them being too good vs Ukraine ... now you want to buff TeslaTanks ... they are also a russian - faction special ...

r34ch:
@MicroJOo I think its a fair point you make about existing balance feedback and of Ukraine. I have left my thoughts on your PR. As for Ukraine, I wouldn't mind the demo truck speed slightly increased so that when it is killed, it'll die a cell or two closer to the intended target - thus being a little more cost effective. That said if you think buffing the TT is too soon after nerfing the other russian faction specific unit I have no issues removing that change until we got feedback. 

As for the phase however, I have to strongly disagree. If anything I feel the changes I propose strengthen its role of 'sneaking in special units'. The speed buffs allow a player to reduce the potential time of being spotted and to exploit temporary gaps in the enemies defences; you know how fast defences are built. The LoS buff allows you to avoid minor base defences much easier -and as such, detection- and the capacity increase allows you to sneak more special units into a base.

The reason I didn't want to buff the phase's weapon as you hint at is because then it would have too similar a role to the chronotank - hit and run. As you say, they are hard to detect on the battlefield so a good weapon would be OP in the right hands. Their cargo has the ability to cause a lot of damage if used right, but are visible and fragile - so easily countered.


Really though all this theory-crafting means nothing unless people go and test this stuff. I was however under the impression that a feature freeze did not cover balance so long as a playtest was due before the next release - so that these changes could be tested and if need be, reverted before release. But from what I have read on IRC it appears I am mistaken. I think you had a good idea making a couple 'balance changes maps'.

cjshmyr:
As for the tesla tanks > ukraine bit, I'm cool with a dtrk speed increase to help Ukraine out (or heck, if we are really generous we increase the speed of the parabomb badgers for breaking light forces quicker).

GrayHatter:
> (or heck, if we are really generous we increase the speed of the parabomb badgers for breaking light forces quicker).

I might actually play Ukraine with that change!

mods/ra/rules/vehicles.yaml
```diff
@@ -563,7 +563,7 @@ TTNK:
 	Armament:
 		Weapon: TTankZap
 		LocalOffset: 0,0,213
-	AttackFrontal:
+	AttackOmni:
```
obrakmann: This doesn't work at all for mobile units. Try ordering the tesla tank to a unit that's further away than its weapon range: it won't move towards the target at all.

pchote: Could instead give it AttackTurreted with an invisible ROT: 255 turret.

obrakmann: Yes, that works. Will make the ttnks be able to also shoot on the move, though.

Just add this, ROT is 255 already by default:
```
Turreted:
AttackTurreted:
```

obrakmann:
Superseded by #9992.

mods/ra/rules/vehicles.yaml
r34ch: Thanks for fixing this; I also had the same issue with AttackCharge (which implements AttackOmni)

PR #9891, opened 2015-11-05
Added sounds for orca takeoff and landing (TS)

teees:

PR #9963, opened 2015-11-11
Play Takeoff/Landing sounds for Planes as well as Helicopters

phrohdoh:
Written to fix https://github.com/OpenRA/OpenRA/pull/9891#issuecomment-154857048.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Air/TakeOff.cs
+++ b/OpenRA.Mods.Common/Activities/Air/TakeOff.cs
@@ -18,6 +18,7 @@ namespace OpenRA.Mods.Common.Activities
 	{
 		readonly Aircraft aircraft;
 		readonly IMove move;
+		bool playedSound;
 
 		public TakeOff(Actor self)
 		{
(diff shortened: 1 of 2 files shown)


obrakmann:
Thanks. Any reason why you added it `Fly` instead of `TakeOff`, though?

Phrohdoh:
Oh, only because the code I copied existed in `HeliFly`, I can change it if that is more correct.

penev92:
I agree it would fit better in `TakeOff`.

Phrohdoh:
Moved into `TakeOff`.

abcdefg30:
This doesn't build at the moment.

OpenRA.Mods.Common/Activities/Air/TakeOff.cs
```diff
@@ -30,6 +31,12 @@ public override Activity Tick(Actor self)
 			if (NextActivity == null)
 				self.CancelActivity();
 
+			if (!playedSound && plane.Info.TakeoffSound != null && self.IsAtGroundLevel())
```
abcdefg30: I think it's `aircraft` instead of `plane`.

Phrohdoh: Correct. I didn't even test this. (shame!)

Phrohdoh:
Fixed.

abcdefg30:
I still somehow don't get a takeoff sound.

Phrohdoh:
Neither do I, now.
I will investigate this more tomorrow.

PR #9970, opened 2015-11-11
Fixed the minimap orders for left click control

### LINKED ISSUE #9965: Minimap orders don't work in the left click control scheme
In the original C&C, you could select an unit, or a group of them and you could pilot them directly on the radar window, or order them to attack. Having such feature could nice, but in case, having even a "zoom" button that can let the user zoom in a particular area and giving order, avoiding the problem that the points are to small to locate, would help its usage.

So if you have to give an attack or patrol command, you can do that, without moving the main camera, and lose sight of what is going on there.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28156868-minimap-orders-don-t-work-in-the-left-click-control-scheme?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Reizoukodesu:
Use the correct actions when orders are issued over the minimap.
The `Game.Settings.Game.MouseButtonPreference` settings are referenced.

This commit fixes #9965

### DIFF
--- a/OpenRA.Mods.Common/Widgets/RadarWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/RadarWidget.cs
@@ -222,7 +222,7 @@ namespace OpenRA.Mods.Common.Widgets
 			var mi = new MouseInput
 			{
 				Location = location,
-				Button = MouseButton.Right,
+				Button = Game.Settings.Game.MouseButtonPreference.Action,
 				Modifiers = Game.GetModifierKeys()
 			};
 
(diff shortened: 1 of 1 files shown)


penev92:
Hey, thanks for the quick fix.
There seems to be an unrelated map file in the commit. Also this needs to be rebased on top of current bleed.

Sapena:
Well now I see it too... sorry for that, going to rebase and removed the unnecessary map file

Sapena:
The codebase should be up to date now, please let me know if theres something more to be done

pchote:
Our project conventions are to use rebasing when updating a pr rather than merging.  This keeps the project history cleaner.  https://github.com/OpenRA/OpenRA/wiki/Contributing#keeping-up-to-date-rebasing has some instructions on how to rebase if you're not familiar with how to do this.

abcdefg30:
Also squash the commits, please.

Sapena:
The commits have been squashed. Let me know if there's something more to do.

OpenRA.Mods.Common/Widgets/RadarWidget.cs
```diff
@@ -243,17 +243,20 @@ public override bool HandleMouseInput(MouseInput mi)
 
 			var cell = MinimapPixelToCell(mi.Location);
 			var pos = world.Map.CenterOfCell(cell);
-			if ((mi.Event == MouseInputEvent.Down || mi.Event == MouseInputEvent.Move) && mi.Button == MouseButton.Left)
+			if ((mi.Event == MouseInputEvent.Down || mi.Event == MouseInputEvent.Move)
+				&& mi.Button == Game.Settings.Game.MouseButtonPreference.Cancel)
+			{
 				worldRenderer.Viewport.Center(pos);
+			}
```
RoosterDragon: Our style convention is to not add braces for single line statements like this.

RoosterDragon:
:+1: Looks good to me (style fix would be a bonus but is not required)

OpenRA.Mods.Common/Widgets/RadarWidget.cs
Mailaender: I think in this case the braces really help with readability as the indention alone doesn't really make things clear.

Mailaender:
Thanks!

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3fc21eb763ed3c17839242689f5b8e20ed808a58...0f63d1ecb63ddd83d76b2e808c8a1e5215dcc5ec)

PR #9974, opened 2015-11-12
Fix spice bloom editor visibility

### LINKED ISSUE #9957: D2k spice blooms invisible in editor
As above, don't see no spice blooms in the editor.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28125827-d2k-spice-blooms-invisible-in-editor?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Closes #9957.

### DIFF
--- a/OpenRA.Mods.D2k/Traits/SpiceBloom.cs
+++ b/OpenRA.Mods.D2k/Traits/SpiceBloom.cs
@@ -14,6 +14,7 @@ using System.Linq;
 using OpenRA.Effects;
 using OpenRA.GameRules;
 using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
 using OpenRA.Mods.Common.Traits;
 using OpenRA.Primitives;
 using OpenRA.Traits;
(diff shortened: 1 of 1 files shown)


pchote:
Why is this trait diying it's rendering? It could use our existing traits with upgrades to switch.

GraionDilach:
On long-term I'd prefer an upgrade-based method to substitute `RenderSpritesEditorOnly` (since such an upgrade-based method would render that one trait moot), but on short-term, I guess this can do.

RoosterDragon:
:+1: Seems like an sensible short term fix.

pchote:
As a short term fix this is fine, so :+1: from me too.

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/0427378afaec38d90938ecbdbbb82293a64d9247

PR #9977, opened 2015-11-12
Ensure LuaValues are disposed

RoosterDragon:
Adding in these missing calls prevents these instances from having to be finalized.

[Diff ignoring whitespace](https://github.com/OpenRA/OpenRA/pull/9977/files?w=1)

### DIFF
--- a/OpenRA.Mods.RA/Scripting/Properties/ChronosphereProperties.cs
+++ b/OpenRA.Mods.RA/Scripting/Properties/ChronosphereProperties.cs
@@ -28,8 +28,13 @@ namespace OpenRA.Mods.RA.Scripting
 			{
 				Actor actor;
 				CPos cell;
-				if (!kv.Key.TryGetClrValue<Actor>(out actor) || !kv.Value.TryGetClrValue<CPos>(out cell))
-					throw new LuaException("Chronoshift requires a table of Actor,CPos pairs. Received {0},{1}".F(kv.Key.WrappedClrType().Name, kv.Value.WrappedClrType().Name));
+				using (kv.Key)
+				using (kv.Value)
+				{
+					if (!kv.Key.TryGetClrValue(out actor) || !kv.Value.TryGetClrValue(out cell))
+						throw new LuaException("Chronoshift requires a table of Actor,CPos pairs. Received {0},{1}".F(
+							kv.Key.WrappedClrType().Name, kv.Value.WrappedClrType().Name));
+				}
 
 				var cs = actor.TraitOrDefault<Chronoshiftable>();
 				if (cs != null && cs.CanChronoshiftTo(actor, cell))
--- a/OpenRA.Game/Scripting/ScriptContext.cs
+++ b/OpenRA.Game/Scripting/ScriptContext.cs
@@ -64,7 +64,19 @@ namespace OpenRA.Scripting
 		}
 	}
 
-	// For global-level bindings
+	/// <summary>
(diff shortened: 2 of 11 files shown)


OpenRA.Mods.RA/Scripting/Properties/ChronosphereProperties.cs
```diff
@@ -28,8 +28,16 @@ public void Chronoshift(LuaTable unitLocationPairs, int duration = 0, bool killC
 			{
 				Actor actor;
 				CPos cell;
-				if (!kv.Key.TryGetClrValue<Actor>(out actor) || !kv.Value.TryGetClrValue<CPos>(out cell))
-					throw new LuaException("Chronoshift requires a table of Actor,CPos pairs. Received {0},{1}".F(kv.Key.WrappedClrType().Name, kv.Value.WrappedClrType().Name));
+				try
+				{
+					if (!kv.Key.TryGetClrValue(out actor) || !kv.Value.TryGetClrValue(out cell))
+						throw new LuaException("Chronoshift requires a table of Actor,CPos pairs. Received {0},{1}".F(kv.Key.WrappedClrType().Name, kv.Value.WrappedClrType().Name));
```
GraionDilach: It's just a single character too long line for StyleCop!

pchote:
I'm not sure I like the idea of silently disposing the arguments to a method from deep inside it.  Would it be possible to use a `using` block or explicitly dispose in the methods that 'own' the objects instead?

RoosterDragon:
Which methods are you worried about specifically? Bear in mind most of these are callbacks from Lua land - the methods I'm disposing in \*are\* the owners of these objects. Eluant creates these specifically to pass as parameters for the CLR method - we own them the instant our callback is invoked.

RoosterDragon:
Trying to explain it reminds me of how this melted my brain a bit when I was fixing it in the first place.

I think I'll throw in some code comments to de-mystify it somewhat.

RoosterDragon:
So in the process of adding comments, I noticed I'd missing a bunch of things so I've had to throw in a bunch more cleanup.

I've also had to leave myself some TODOs. :(

I'm going to have to investigate more closely exactly where we need some of these calls to `Dispose` as previously I was just sticking in calls so the RA shellmap would have less to finalize. But since the RA shellmap doesn't exercise all the possible methods I don't know if I've done things comprehensively yet or not. Some of the stuff that goes into deferred actions is even more mind-bending. Do we really need a copy of the Lua functions/values for these? Which ones do we need to dispose if we do and when? Blugh.

RoosterDragon:
I have updated my code and solved the remaining TODOs.

`ScriptMemberWrapper.Invoke` now ensures any `LuaValues` it passed in as parameters are disposed automatically, which solves most of the problem without requiring major surgery in all the methods. 

The rest of the cleanup generally involves just making sure we call dispose or use usings in a few places. Some common patterns we've missed are when enumerating a `LuaTable` (the key and value are created on demand, and must be disposed unless they are saved elsewhere) and when populating a newly created `LuaTable` (the keys and values we add need to be disposed after we're done adding them).

Mailaender:
Can't spot negative side effects in scripted maps. :white_check_mark:

obrakmann:
Looks ok to me, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7cf33f2842adb549e6ed9cd755509a0ac3346e35)

PR #9659, opened 2015-10-17
Add atreides03a

abcdefg30:

PR #9979, opened 2015-11-12

RoosterDragon:
Allow VS to reorder entries in solution file.

Quick followup to #9659, see also #9696.


abcdefg30:
:+1:

PR #9980, opened 2015-11-13
Allow loading of ChromeLogic properties from YAML

penev92:
As opposed to having to add new properties to the widget and loading the values into those, then later having the `ChromeLogic` query the widget to get those.

Also allow mods to override the asset browser's supported file formats least so they can add their own file extensions.

### DIFF
--- a/OpenRA.Game/Widgets/WidgetLoader.cs
+++ b/OpenRA.Game/Widgets/WidgetLoader.cs
@@ -67,7 +67,14 @@ namespace OpenRA
 					foreach (var c in child.Value.Nodes)
 						LoadWidget(args, widget, c);
 
+			var logicNode = node.Value.Nodes.FirstOrDefault(n => n.Key == "Logic");
+			var logic = logicNode == null ? null : logicNode.Value.ToDictionary();
+			args.Add("logicArgs", logic);
+
 			widget.PostInit(args);
+
+			args.Remove("logicArgs");
+
 			return widget;
 		}
 
--- a/OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
@@ -22,7 +22,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 {
 	public class AssetBrowserLogic : ChromeLogic
 	{
-		static readonly string[] AllowedExtensions = { ".shp", ".r8", "tmp", ".tem", ".des", ".sno", ".int", ".jun", ".vqa" };
+		static string[] allowedExtensions;
 
 		readonly World world;
 
@@ -45,7 +45,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 		int currentFrame;
 
 		[ObjectCreator.UseCtor]
-		public AssetBrowserLogic(Widget widget, Action onExit, World world)
+		public AssetBrowserLogic(Widget widget, Action onExit, World world, Dictionary<string, MiniYaml> logicArgs)
 		{
 			this.world = world;
 
@@ -212,6 +212,11 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				prevButton.IsVisible = () => !isVideoLoaded;
 			}
 
+			if (logicArgs.ContainsKey("SupportedFormats"))
+				allowedExtensions = FieldLoader.GetValue<string[]>("SupportedFormats", logicArgs["SupportedFormats"].Value);
+			else
+				allowedExtensions = new string[0];
+
 			assetList = panel.Get<ScrollPanelWidget>("ASSET_LIST");
 			template = panel.Get<ScrollItemWidget>("ASSET_TEMPLATE");
 			PopulateAssetList();
@@ -359,7 +364,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			var files = assetSource.AllFileNames().OrderBy(s => s);
 			foreach (var file in files)
 			{
-				if (AllowedExtensions.Any(ext => file.EndsWith(ext, true, CultureInfo.InvariantCulture)))
+				if (allowedExtensions.Any(ext => file.EndsWith(ext, true, CultureInfo.InvariantCulture)))
 				{
 					AddAsset(assetList, file, template);
 					availableShps.Add(file);
--- a/mods/cnc/chrome/assetbrowser.yaml
+++ b/mods/cnc/chrome/assetbrowser.yaml
@@ -1,5 +1,6 @@
 Container@ASSETBROWSER_PANEL:
 	Logic: AssetBrowserLogic
+		SupportedFormats: .shp, .tem, .des, .sno, .jun, .vqa
 	X: (WINDOW_RIGHT - WIDTH)/2
 	Y: (WINDOW_BOTTOM - HEIGHT)/2
 	Width: 695
--- a/mods/ra/chrome/assetbrowser.yaml
+++ b/mods/ra/chrome/assetbrowser.yaml
@@ -1,5 +1,6 @@
 Background@ASSETBROWSER_PANEL:
 	Logic: AssetBrowserLogic
+		SupportedFormats: .shp, .r8, .tmp, .tem, .des, .sno, .int, .vqa
 	X: (WINDOW_RIGHT - WIDTH)/2
 	Y: (WINDOW_BOTTOM - HEIGHT)/2
 	Width: 700


mods/ra/chrome/assetbrowser.yaml
```diff
@@ -1,5 +1,6 @@
 Background@ASSETBROWSER_PANEL:
 	Logic: AssetBrowserLogic
+		SupportedFormats: .shp, .r8, tmp, .tem, .des, .sno, .int, .jun, .vqa
```
Phrohdoh: Why redefine these here if they are defined in C#?

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -22,7 +22,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 {
 	public class AssetBrowserLogic : ChromeLogic
 	{
-		static readonly string[] AllowedExtensions = { ".shp", ".r8", "tmp", ".tem", ".des", ".sno", ".int", ".jun", ".vqa" };
+		static string[] allowedExtensions = { ".shp", ".r8", "tmp", ".tem", ".des", ".sno", ".int", ".jun", ".vqa" };
```
Phrohdoh: It looks like these values are overridden no matter what.
Why have them if that is the case?

GraionDilach: A mod might add special variations to this - the first example I can think of would be `mouse.sha` from YR which is just an SHP with a different extension to not mix with the RA2 asset.

mods/ra/chrome/assetbrowser.yaml
GraionDilach: Dammit, I mixed the two lines in the diff view. >_> This is where I intended to mention that thirdparty mods might want to use their own extensions for the Asset Browser too. >_>

OpenRA.Game/Widgets/WidgetLoader.cs
```diff
@@ -48,6 +48,15 @@ public Widget LoadWidget(WidgetArgs args, Widget parent, MiniYamlNode node)
...
 				args = new WidgetArgs(args) { { "modRules", modData.DefaultRules } };
 
+			object parentLogic = null;
+			if (args.ContainsKey("logic"))
+			{
+				parentLogic = args["logic"];
+				args.Remove("logic");
+			}
+
+			args.Add("logic", node.Value.Nodes.FirstOrDefault(n => n.Key == "Logic"));
```
pchote: I think it would be a lot nice to ToDictionary the child nodes, and then store it under a `LogicArgs` key.  You then don't need to stash and restore `Logic`, and the code that uses these becomes simpler.

pchote:
This is something that we've needed for a long time. Thanks for taking it on!

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -212,6 +212,15 @@ public AssetBrowserLogic(Widget widget, Action onExit, World world)
...
 			}
 
+			var node = logic.Value.Nodes.FirstOrDefault(x => x.Key == "SupportedFormats");
+			if (node != null)
+			{
+				var value = node.Value.Value.Replace(" ", "");
+				allowedExtensions = value.Split(',');
+			}
+			else
+				allowedExtensions = new string[0];
```
penev92: NOTE: It was very late last night. In the end I opted for keeping the default value, so I should have removed the `else`. It is here only because I was going with no default value at one point.

OpenRA.Game/Widgets/WidgetLoader.cs
penev92: That still leaves me vulnerable to duplicate key exceptions though, inside the dictionary.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
pchote: I'd prefer that we defaulted to an empty list, and had everything defined in the yaml.  All the values are highly mod-specific, and this is now the only place that they are hardcoded (everything else goes via mod.yaml).

OpenRA.Game/Widgets/WidgetLoader.cs
pchote: In this case we should take only the last value.  You can either do it manually here or push it into an overload of ToDictionary - this is something that will be useful in other places too.

pchote: Actually, we have exactly the same problem with the auto FieldLoader dictionary loading.  I think it would be best to fix and then use that here instead.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
penev92: OK, that was my initial thinking as well. I just figured people will be against that. I'll remove the fallback value then.

mods/ra/chrome/assetbrowser.yaml
penev92: OK, those are no longer defined in YAML.

penev92:
Fixed.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -45,7 +45,7 @@ public class AssetBrowserLogic : ChromeLogic
 		int currentFrame;
 
 		[ObjectCreator.UseCtor]
-		public AssetBrowserLogic(Widget widget, Action onExit, World world)
+		public AssetBrowserLogic(Widget widget, Action onExit, World world, Dictionary<string, MiniYaml> logicArgs)
```
pchote: `Dictionary<string, string>` please.  The main reason for my earlier request was so that the chrome logic didn't need to deal with MiniYaml.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -212,6 +212,14 @@ public AssetBrowserLogic(Widget widget, Action onExit, World world)
 				prevButton.IsVisible = () => !isVideoLoaded;
 			}
 
+			if (logicArgs.ContainsKey("SupportedFormats"))
```
pchote: Use FieldLoader to convert to `string[]`.  This is basically reinventing the logic that already exists there.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
penev92: What if someone wants to pass a dictionary as one of the logic class' parameters?
Don't forget that the code that handles the initializations is very generic and handles the initialization of *all* widget logic classes.

pchote: Ah yes, fair enough.

mods/ra/chrome/assetbrowser.yaml
pchote: RA doesn't use `.r8` or `.jun`, so they can be dropped.  Maybe also `.tmp`?

mods/cnc/chrome/assetbrowser.yaml
```diff
@@ -1,5 +1,6 @@
 Container@ASSETBROWSER_PANEL:
 	Logic: AssetBrowserLogic
+		SupportedFormats: .shp, .r8, tmp, .tem, .des, .sno, .int, .jun, .vqa
```
pchote: TD doesn't use `.r8`, `tmp` (erm, no dot?), or `.int`.

mods/ra/chrome/assetbrowser.yaml
GraionDilach: What does `tmp` mean here? Terrain tiles are in TMP (RA) format and not in SHP format, but still having the terrain-specific extension.

pchote: literally `.tmp` as a file extension.

penev92:
Fixed.

pchote:
Code looks good to me know, but I haven't tested it. :+0.5:

pchote:
Works as advertised, so :+1:

penev92:
@ABrandau , @DoGyAUT , @GraionDilach you may want to see if your Asset Browsers need tweaking once this is merged.

Mailaender:
@Phrohdoh You will need to add `urb` to the list for Red Alert 2. http://modenc.renegadeprojects.com/Theatres

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/536b207603224899ef18045d3d6b4e761710cd52...0e196bf397dc6be759456e1a15245a56d3127159)

PR #7888, opened 2015-04-06
Added gates

Mailaender:
Took me quite long to actually get this working. It is not final quality, both from architectural related things (should probably inherit from the Building trait, but I could not get that to work due to trait inter-dependencies) and feature wise (the pathfinder still ignores friendly closed gates), but you need to start somewhere. You can use http://resource.openra.net/maps/2738 for testing.

Depends on the cleanups from https://github.com/OpenRA/OpenRA/pull/7887.

PR #9981, opened 2015-11-13
Added gates to Tiberian Sun

### LINKED ISSUE #3196: Gates [$25]
There are already custom made sprites in mods/ra/bits for both soviets/allies as well as horizontal/vertical (agate.shp, sgate.shp, vgate.shp) with complete open/close plus destruction animations. Currently unused because of missing code. Depends on #2248.

<bountysource-plugin>
---
There is a **[$25 open bounty](https://www.bountysource.com/issues/305907-gates?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

teees:
- Reuses some code from #7888
- Closes #3196
###### Works by
- Letting the pathfinder think there is no obstruction (for friendly and open gates)
- Blocking movement for closed gates and opens via OnNotifyBlockingMove
###### Improvements for the future:
- [fix zoffset for gates](https://github.com/OpenRA/OpenRA/pull/9981#issuecomment-174310942)
- placing gates on existing walls
- [fix the hitbox](https://github.com/OpenRA/OpenRA/pull/9981#issuecomment-170360964)

### DIFF
--- a/OpenRA.Game/FieldLoader.cs
+++ b/OpenRA.Game/FieldLoader.cs
@@ -373,6 +373,28 @@ namespace OpenRA
 
 				return InvalidValueAction(value, fieldType, fieldName);
 			}
+			else if (fieldType == typeof(CVec[]))
+			{
+				if (value != null)
+				{
+					var parts = value.Split(',');
+
+					if (parts.Length % 2 != 0)
+						return InvalidValueAction(value, fieldType, fieldName);
+
+					var vecs = new CVec[parts.Length / 2];
+					for (var i = 0; i < vecs.Length; i++)
+					{
+						int rx, ry;
+						if (int.TryParse(parts[2 * i], out rx) && int.TryParse(parts[2 * i + 1], out ry))
+							vecs[i] = new CVec(rx, ry);
+					}
+
+					return vecs;
+				}
+
+				return InvalidValueAction(value, fieldType, fieldName);
+			}
 			else if (fieldType.IsEnum)
 			{
 				try
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -284,6 +284,13 @@ namespace OpenRA.Traits
 		bool CanEnterTargetNow(Actor self, Target target);
 	}
 
+	[RequireExplicitImplementation]
+	public interface ITemporaryBlocker
+	{
+		bool CanRemoveBlockage(Actor self, Actor blocking);
+		bool IsBlocking(Actor self, CPos cell);
+	}
+
 	public interface INotifyBlockingMove { void OnNotifyBlockingMove(Actor self, Actor blocking); }
 
 	public interface IFacing
--- a/OpenRA.Game/WorldUtils.cs
+++ b/OpenRA.Game/WorldUtils.cs
@@ -37,6 +37,23 @@ namespace OpenRA
 				a => (a.CenterPosition - origin).HorizontalLengthSquared <= r.LengthSquared);
 		}
 
+		public static bool ContainsTemporaryBlocker(this World world, CPos cell, Actor ignoreActor = null)
+		{
+			var temporaryBlockers = world.ActorMap.GetActorsAt(cell);
+			foreach (var temporaryBlocker in temporaryBlockers)
+			{
+				if (temporaryBlocker == ignoreActor)
+					continue;
+
+				var temporaryBlockerTraits = temporaryBlocker.TraitsImplementing<ITemporaryBlocker>();
+				foreach (var temporaryBlockerTrait in temporaryBlockerTraits)
+					if (temporaryBlockerTrait.IsBlocking(temporaryBlocker, cell))
+						return true;
+			}
+
+			return false;
+		}
+
 		public static void DoTimed<T>(this IEnumerable<T> e, Action<T> a, string text)
 		{
 			// PERF: This is a hot path and must run with minimal added overhead.
--- a/OpenRA.Mods.Common/Activities/Move/Move.cs
+++ b/OpenRA.Mods.Common/Activities/Move/Move.cs
@@ -220,12 +220,14 @@ namespace OpenRA.Mods.Common.Activities
 
 			var nextCell = path[path.Count - 1];
 
+			var containsTemporaryBlocker = WorldUtils.ContainsTemporaryBlocker(self.World, nextCell, self);
+
 			// Next cell in the move is blocked by another actor
-			if (!mobile.CanMoveFreelyInto(nextCell, ignoredActor, true))
+			if (containsTemporaryBlocker || !mobile.CanMoveFreelyInto(nextCell, ignoredActor, true))
 			{
 				// Are we close enough?
 				var cellRange = nearEnough.Length / 1024;
-				if ((mobile.ToCell - destination.Value).LengthSquared <= cellRange * cellRange)
+				if (!containsTemporaryBlocker && (mobile.ToCell - destination.Value).LengthSquared <= cellRange * cellRange)
 				{
 					path.Clear();
 					return null;
--- a/OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
+++ b/OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
@@ -9,9 +9,16 @@
 #endregion
 
 using System.Linq;
+using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
+	[RequireExplicitImplementation]
+	public interface IBlocksProjectiles
+	{
+		WDist BlockingHeight { get; }
+	}
+
 	[Desc("This actor blocks bullets and missiles with 'Blockable' property.")]
 	public class BlocksProjectilesInfo : UpgradableTraitInfo
 	{
@@ -20,17 +27,20 @@ namespace OpenRA.Mods.Common.Traits
 		public override object Create(ActorInitializer init) { return new BlocksProjectiles(init.Self, this); }
 	}
 
-	public class BlocksProjectiles : UpgradableTrait<BlocksProjectilesInfo>
+	public class BlocksProjectiles : UpgradableTrait<BlocksProjectilesInfo>, IBlocksProjectiles
 	{
 		public BlocksProjectiles(Actor self, BlocksProjectilesInfo info)
 			: base(info) { }
 
+		WDist IBlocksProjectiles.BlockingHeight { get { return Info.Height; } }
+
 		public static bool AnyBlockingActorAt(World world, WPos pos)
 		{
 			var dat = world.Map.DistanceAboveTerrain(pos);
+
 			return world.ActorMap.GetActorsAt(world.Map.CellContaining(pos))
-				.Any(a => a.TraitsImplementing<BlocksProjectiles>()
-					.Where(t => t.Info.Height.Length >= dat.Length)
+				.Any(a => a.TraitsImplementing<IBlocksProjectiles>()
+					.Where(t => t.BlockingHeight >= dat)
 					.Any(Exts.IsTraitEnabled));
 		}
 
@@ -41,7 +51,7 @@ namespace OpenRA.Mods.Common.Traits
 
 			foreach (var a in actors)
 			{
-				var blockers = a.TraitsImplementing<BlocksProjectiles>()
+				var blockers = a.TraitsImplementing<IBlocksProjectiles>()
 					.Where(Exts.IsTraitEnabled).ToList();
 
 				if (!blockers.Any())
@@ -49,7 +59,7 @@ namespace OpenRA.Mods.Common.Traits
 
 				var hitPos = WorldExtensions.MinimumPointLineProjection(start, end, a.CenterPosition);
 				var dat = world.Map.DistanceAboveTerrain(hitPos);
-				if ((hitPos - start).Length < length && blockers.Any(t => t.Info.Height.Length >= dat.Length))
+				if ((hitPos - start).Length < length && blockers.Any(t => t.BlockingHeight >= dat))
 				{
 					hit = hitPos;
 					return true;
--- a/OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
+++ b/OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
@@ -40,7 +40,7 @@ namespace OpenRA.Mods.Common.Traits
 
 			return world.ActorMap.GetActorsAt(world.Map.CellContaining(pos))
 				.Any(a => a.TraitsImplementing<IBlocksProjectiles>()
-					.Where(t => t.BlockingHeight >= dat)
+					.Where(t => t.BlockingHeight > dat)
 					.Any(Exts.IsTraitEnabled));
 		}
 
@@ -59,7 +59,7 @@ namespace OpenRA.Mods.Common.Traits
 
 				var hitPos = WorldExtensions.MinimumPointLineProjection(start, end, a.CenterPosition);
 				var dat = world.Map.DistanceAboveTerrain(hitPos);
-				if ((hitPos - start).Length < length && blockers.Any(t => t.BlockingHeight >= dat))
+				if ((hitPos - start).Length < length && blockers.Any(t => t.BlockingHeight > dat))
 				{
 					hit = hitPos;
 					return true;
--- a/OpenRA.Mods.Common/Traits/Buildings/BuildingUtils.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/BuildingUtils.cs
@@ -71,12 +71,11 @@ namespace OpenRA.Mods.Common.Traits
 						continue; // Cell is empty; continue search
 
 					// Cell contains an actor. Is it the type we want?
-					if (world.ActorsHavingTrait<LineBuildNode>()
-						.Any(a => a.Location == cell
-							&& a.Info.TraitInfo<LineBuildNodeInfo>().Types.Overlaps(lbi.NodeTypes)))
-						dirs[d] = i; // Cell contains actor of correct type
-					else
-						dirs[d] = -1; // Cell is blocked by another actor type
+					var hasConnector = world.ActorMap.GetActorsAt(cell)
+						.Any(a => a.Info.TraitInfos<LineBuildNodeInfo>()
+							.Any(info => info.Types.Overlaps(lbi.NodeTypes) && info.Connections.Contains(vecs[d])));
+
+					dirs[d] = hasConnector ? i : -1;
 				}
 
 				// Place intermediate-line sections
--- a/OpenRA.Mods.Common/Traits/Buildings/Gate.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/Gate.cs
@@ -1,0 +1,135 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Will open and be passable for actors that appear friendly when there are no enemies in range.")]
+	public class GateInfo : BuildingInfo
+	{
+		public readonly string OpeningSound = null;
+		public readonly string ClosingSound = null;
+
+		[Desc("Ticks until the gate closes.")]
+		public readonly int CloseDelay = 150;
+
+		[Desc("Ticks until the gate is considered open.")]
+		public readonly int TransitionDelay = 33;
+
+		[Desc("Blocks bullets scaled to open value.")]
+		public readonly int BlocksProjectilesHeight = 640;
+
+		public override object Create(ActorInitializer init) { return new Gate(init, this); }
+	}
+
+	public class Gate : Building, ITick, ITemporaryBlocker, IBlocksProjectiles, INotifyBlockingMove, ISync
+	{
+		readonly GateInfo info;
+		readonly Actor self;
+		IEnumerable<CPos> blockedPositions;
+
+		public readonly int OpenPosition;
+		[Sync] public int Position { get; private set; }
+		int desiredPosition;
+		int remainingOpenTime;
+
+		public Gate(ActorInitializer init, GateInfo info)
+			: base(init, info)
+		{
+			this.info = info;
+			self = init.Self;
+			OpenPosition = info.TransitionDelay;
+		}
+
+		void ITick.Tick(Actor self)
+		{
+			if (self.IsDisabled() || Locked || !BuildComplete)
+				return;
+
+			if (desiredPosition < Position)
+			{
+				// Gate was fully open
+				if (Position == OpenPosition)
+				{
+					Game.Sound.Play(info.ClosingSound, self.CenterPosition);
+					self.World.ActorMap.AddInfluence(self, this);
+				}
+
+				Position--;
+			}
+			else if (desiredPosition > Position)
+			{
+				// Gate was fully closed
+				if (Position == 0)
+					Game.Sound.Play(info.OpeningSound, self.CenterPosition);
+
+				Position++;
+
+				// Gate is now fully open
+				if (Position == OpenPosition)
+				{
+					self.World.ActorMap.RemoveInfluence(self, this);
+					remainingOpenTime = info.CloseDelay;
+				}
+			}
+
+			if (Position == OpenPosition)
+			{
+				if (IsBlocked())
+					remainingOpenTime = info.CloseDelay;
+				else if (--remainingOpenTime <= 0)
+					desiredPosition = 0;
+			}
+		}
+
+		bool ITemporaryBlocker.IsBlocking(Actor self, CPos cell)
+		{
+			return Position != OpenPosition && blockedPositions.Contains(cell);
+		}
+
+		bool ITemporaryBlocker.CanRemoveBlockage(Actor self, Actor blocking)
+		{
+			return CanRemoveBlockage(self, blocking);
+		}
+
+		void INotifyBlockingMove.OnNotifyBlockingMove(Actor self, Actor blocking)
+		{
+			if (Position != OpenPosition && CanRemoveBlockage(self, blocking))
+				desiredPosition = OpenPosition;
+		}
+
+		bool CanRemoveBlockage(Actor self, Actor blocking)
+		{
+			return !self.IsDisabled() && BuildComplete && blocking.AppearsFriendlyTo(self);
+		}
+
+		public override void AddedToWorld(Actor self)
+		{
+			base.AddedToWorld(self);
+			blockedPositions = FootprintUtils.Tiles(self);
+		}
+
+		bool IsBlocked()
+		{
+			return blockedPositions.Any(loc => self.World.ActorMap.GetActorsAt(loc).Any(a => a != self));
+		}
+
+		WDist IBlocksProjectiles.BlockingHeight
+		{
+			get
+			{
+				return new WDist(info.BlocksProjectilesHeight * (OpenPosition - Position) / OpenPosition);
+			}
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
@@ -9,7 +9,9 @@
 #endregion
 
 using System;
+using System.Collections.Generic;
 using System.Drawing;
+using System.Linq;
 using OpenRA.Graphics;
 using OpenRA.Mods.Common.Effects;
 using OpenRA.Traits;
@@ -22,24 +24,28 @@ namespace OpenRA.Mods.Common.Traits
 		public object Create(ActorInitializer init) { return new CombatDebugOverlay(init.Self); }
 	}
 
-	public class CombatDebugOverlay : IPostRender, INotifyDamage
+	public class CombatDebugOverlay : IPostRender, INotifyDamage, INotifyCreated
 	{
 		readonly DeveloperMode devMode;
 
 		readonly HealthInfo healthInfo;
-		readonly BlocksProjectilesInfo blockInfo;
+		IBlocksProjectiles[] allBlockers;
 		Lazy<AttackBase> attack;
 		Lazy<BodyOrientation> coords;
 
 		public CombatDebugOverlay(Actor self)
 		{
 			healthInfo = self.Info.TraitInfoOrDefault<HealthInfo>();
-			blockInfo = self.Info.TraitInfoOrDefault<BlocksProjectilesInfo>();
 			attack = Exts.Lazy(() => self.TraitOrDefault<AttackBase>());
 			coords = Exts.Lazy(() => self.Trait<BodyOrientation>());
 
 			var localPlayer = self.World.LocalPlayer;
 			devMode = localPlayer != null ? localPlayer.PlayerActor.Trait<DeveloperMode>() : null;
+		}
+
+		public void Created(Actor self)
+		{
+			allBlockers = self.TraitsImplementing<IBlocksProjectiles>().ToArray();
 		}
 
 		public void RenderAfterWorld(WorldRenderer wr, Actor self)
@@ -53,10 +59,11 @@ namespace OpenRA.Mods.Common.Traits
 			if (healthInfo != null)
 				healthInfo.Shape.DrawCombatOverlay(wr, wcr, self);
 
-			if (blockInfo != null)
+			var blockers = allBlockers.Where(Exts.IsTraitEnabled).ToList();
+			if (blockers.Count > 0)
 			{
 				var hc = Color.Orange;
-				var height = new WVec(0, 0, blockInfo.Height.Length);
+				var height = new WVec(0, 0, blockers.Max(b => b.BlockingHeight.Length));
 				var ha = wr.ScreenPosition(self.CenterPosition);
 				var hb = wr.ScreenPosition(self.CenterPosition + height);
 				wcr.DrawLine(ha, hb, iz, hc);
--- a/OpenRA.Mods.Common/Traits/Mobile.cs
+++ b/OpenRA.Mods.Common/Traits/Mobile.cs
@@ -240,6 +240,11 @@ namespace OpenRA.Mods.Common.Traits
 				IsMovingInMyDirection(self, otherActor))
 				return false;
 
+			// If there is a temporary blocker in our path, but we can remove it, we are not blocked.
+			var temporaryBlocker = otherActor.TraitOrDefault<ITemporaryBlocker>();
+			if (temporaryBlocker != null && temporaryBlocker.CanRemoveBlockage(otherActor, self))
+				return false;
+
 			// If we cannot crush the other actor in our way, we are blocked.
 			if (self == null || Crushes == null || Crushes.Count == 0)
 				return true;
--- a/OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
@@ -73,8 +73,8 @@ namespace OpenRA.Mods.Common.Traits
 			var adjacentActorTraits = adjacent.SelectMany(self.World.ActorMap.GetActorsAt)
 				.SelectMany(a => a.TraitsImplementing<IWallConnector>());
 
-			foreach (var aat in adjacentActorTraits)
-				aat.SetDirty();
+			foreach (var rb in adjacentActorTraits)
+				rb.SetDirty();
 		}
 
 		public void RemovedFromWorld(Actor self)
--- a/OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
@@ -1,0 +1,93 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	class WithGateSpriteBodyInfo : WithSpriteBodyInfo, Requires<GateInfo>
+	{
+		[Desc("Cells (outside the gate footprint) that contain wall cells that can connect to the gate")]
+		public readonly CVec[] WallConnections = { };
+
+		[Desc("Wall type for connections")]
+		public readonly string Type = "wall";
+
+		public override object Create(ActorInitializer init) { return new WithGateSpriteBody(init, this); }
+
+		public override IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer init, RenderSpritesInfo rs, string image, int facings, PaletteReference p)
+		{
+			var anim = new Animation(init.World, image);
+			anim.PlayFetchIndex(RenderSprites.NormalizeSequence(anim, init.GetDamageState(), Sequence), () => 0);
+
+			yield return new SpriteActorPreview(anim, WVec.Zero, 0, p, rs.Scale);
+		}
+	}
+
+	class WithGateSpriteBody : WithSpriteBody, INotifyRemovedFromWorld, INotifyBuildComplete, IWallConnector
+	{
+		readonly WithGateSpriteBodyInfo gateInfo;
+		readonly Gate gate;
+
+		public WithGateSpriteBody(ActorInitializer init, WithGateSpriteBodyInfo info)
+			: base(init, info, () => 0)
+		{
+			gateInfo = info;
+			gate = init.Self.Trait<Gate>();
+		}
+
+		int GetGateFrame()
+		{
+			return int2.Lerp(0, DefaultAnimation.CurrentSequence.Length - 1, gate.Position, gate.OpenPosition);
+		}
+
+		public override void DamageStateChanged(Actor self, AttackInfo e)
+		{
+			DefaultAnimation.PlayFetchIndex(NormalizeSequence(self, Info.Sequence), GetGateFrame);
+		}
+
+		public override void BuildingComplete(Actor self)
+		{
+			DefaultAnimation.PlayFetchIndex(NormalizeSequence(self, Info.Sequence), GetGateFrame);
+			UpdateNeighbours(self);
+		}
+
+		void UpdateNeighbours(Actor self)
+		{
+			var footprint = FootprintUtils.Tiles(self).ToArray();
+			var adjacent = Util.ExpandFootprint(footprint, true).Except(footprint)
+				.Where(self.World.Map.Contains).ToList();
+
+			var adjacentActorTraits = adjacent.SelectMany(self.World.ActorMap.GetActorsAt)
+				.SelectMany(a => a.TraitsImplementing<IWallConnector>());
+
+			foreach (var aat in adjacentActorTraits)
+				aat.SetDirty();
+		}
+
+		public void RemovedFromWorld(Actor self)
+		{
+			UpdateNeighbours(self);
+		}
+
+		bool IWallConnector.AdjacentWallCanConnect(Actor self, CPos wallLocation, string wallType, out CVec facing)
+		{
+			facing = wallLocation - self.Location;
+			return wallType == gateInfo.Type && gateInfo.WallConnections.Contains(facing);
+		}
+
+		void IWallConnector.SetDirty() { }
+	}
+}
--- a/OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
@@ -17,7 +17,7 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
-	class WithGateSpriteBodyInfo : WithSpriteBodyInfo, Requires<GateInfo>
+	class WithGateSpriteBodyInfo : WithSpriteBodyInfo, IWallConnectorInfo, Requires<GateInfo>
 	{
 		[Desc("Cells (outside the gate footprint) that contain wall cells that can connect to the gate")]
 		public readonly CVec[] WallConnections = { };
@@ -33,6 +33,11 @@ namespace OpenRA.Mods.Common.Traits
 			anim.PlayFetchIndex(RenderSprites.NormalizeSequence(anim, init.GetDamageState(), Sequence), () => 0);
 
 			yield return new SpriteActorPreview(anim, WVec.Zero, 0, p, rs.Scale);
+		}
+
+		string IWallConnectorInfo.GetWallConnectionType()
+		{
+			return Type;
 		}
 	}
 
--- a/OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
@@ -8,6 +8,7 @@
  */
 #endregion
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Graphics;
@@ -74,9 +75,10 @@ namespace OpenRA.Mods.Common.Traits
 		int adjacent = 0;
 		bool dirty = true;
 
-		bool IWallConnector.AdjacentWallCanConnect(Actor self, CPos wallLocation, string wallType)
+		bool IWallConnector.AdjacentWallCanConnect(Actor self, CPos wallLocation, string wallType, out CVec facing)
 		{
-			return wallInfo.Type == wallType;
+			facing = wallLocation - self.Location;
+			return wallInfo.Type == wallType && Math.Abs(facing.X) + Math.Abs(facing.Y) == 1;
 		}
 
 		void IWallConnector.SetDirty() { dirty = true; }
@@ -104,20 +106,18 @@ namespace OpenRA.Mods.Common.Traits
 			adjacent = 0;
 			foreach (var a in adjacentActors)
 			{
-				var rb = a.TraitOrDefault<IWallConnector>();
-				if (rb == null || !rb.AdjacentWallCanConnect(self, self.Location, wallInfo.Type))
+				CVec facing;
+				var wc = a.TraitOrDefault<IWallConnector>();
+				if (wc == null || !wc.AdjacentWallCanConnect(a, self.Location, wallInfo.Type, out facing))
 					continue;
 
-				var location = self.Location;
-				var otherLocation = a.Location;
-
-				if (otherLocation == location + new CVec(0, -1))
+				if (facing.Y > 0)
 					adjacent |= 1;
-				else if (otherLocation == location + new CVec(+1, 0))
+				else if (facing.X < 0)
 					adjacent |= 2;
-				else if (otherLocation == location + new CVec(0, +1))
+				else if (facing.Y < 0)
 					adjacent |= 4;
-				else if (otherLocation == location + new CVec(-1, 0))
+				else if (facing.X > 0)
 					adjacent |= 8;
 			}
 
--- a/OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
@@ -68,11 +68,18 @@ namespace OpenRA.Mods.Common.Traits
 		}
 	}
 
-	class WithWallSpriteBody : WithSpriteBody, INotifyRemovedFromWorld, ITick
+	class WithWallSpriteBody : WithSpriteBody, INotifyRemovedFromWorld, IWallConnector, ITick
 	{
 		readonly WithWallSpriteBodyInfo wallInfo;
 		int adjacent = 0;
 		bool dirty = true;
+
+		bool IWallConnector.AdjacentWallCanConnect(Actor self, CPos wallLocation, string wallType)
+		{
+			return wallInfo.Type == wallType;
+		}
+
+		void IWallConnector.SetDirty() { dirty = true; }
 
 		public WithWallSpriteBody(ActorInitializer init, WithWallSpriteBodyInfo info)
 			: base(init, info, () => 0)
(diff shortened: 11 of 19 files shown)


OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+	public class GateInfo : BuildingInfo, ITraitInfo, Requires<WithSpriteBodyInfo>
+	{
+		public readonly string OpenSequence = "open";
+		public readonly string ClosingSequence = "closing";
+		public readonly string ClosedSequence = "closed";
+
+		public readonly string OpeningSound = "gateup1.aud";
+		public readonly string ClosingSound = "gatedwn1.aud";
+
+		[Desc("How many ticks until the gate will close after being opened.")]
```
penev92: "How many ticks until the gate starts closing after being opened?"

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+		public readonly int TicksToStayOpen = 50;
+
+		[Desc("How many ticks until the gate is open.")]
```
penev92: Is this the time it takes for the open/close animation to play? I don't understand the purpose from the name + description, so it probably needs a better description.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+ * see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public interface ITemporaryBlocker
```
penev92: Put this in `TraitInterfaces.cs`.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+	{
+		bool CanLiftBlockade(Actor self, Actor blocking);
```
penev92: `CanRemoveBlocker()` ?

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	public class Gate : Building, ITick, ITemporaryBlocker, INotifyBlockingMove
+	{
+		readonly GateInfo info;
+		readonly Actor self;
+		readonly WithSpriteBody wsb;
+
+		bool isOpen;
+		bool opening;
```
penev92: `isOpening`

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+
+		int remainingTime;
+
+		public Gate(ActorInitializer init, GateInfo info) : base(init, info)
```
penev92: Put the base constructor on the next line to be consistent with the rest of the codebase please.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+		{
+			this.info = info;
+			this.self = init.Self;
+
+			wsb = self.Trait<WithSpriteBody>();
+		}
+
+		void ITick.Tick(Actor self)
+		{
+			if (self.IsDisabled() || !BuildComplete)
```
penev92: Should probably consider `Locked` here as well, in case someone decides to capture them.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+			{
+				opening = true;
+				remainingTime = info.TicksToOpen;
+
+				Game.Sound.Play(info.OpeningSound, self.CenterPosition);
+				wsb.PlayCustomAnimationBackwards(self, info.ClosingSequence, () => wsb.PlayCustomAnimationRepeating(self, info.OpenSequence));
+			}
+		}
+
+		void Opened()
```
penev92: `OnOpened()` (or maybe `OnOpen()` ? ) please.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+		void ITick.Tick(Actor self)
+		{
+			if (self.IsDisabled() || !BuildComplete)
+				return;
+
+			if (opening && remainingTime-- <= 0)
+				Opened();
+			else if (isOpen)
+			{
+				if (remainingTime-- <= 0)
```
penev92: Perhaps you should extract a single `remainingTime--` before the first check to avoid both confusion and bugs.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+				{
+					if (!IsBlocked())
+						Close();
+					else
+						remainingTime = info.TicksToStayOpen;
+				}
+				else if (remainingTime % 10 == 0 && IsBlocked())
```
penev92: What's with the `%10` here?
This entire `if` seems redundant.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		bool ITemporaryBlocker.Blocks()
+		{
+			return !isOpen;
+		}
+
+		bool ITemporaryBlocker.CanLiftBlockade(Actor self, Actor blocking)
+		{
+			return !self.IsDisabled() && BuildComplete && blocking.AppearsFriendlyTo(self);
```
penev92: Again, you should check if `Locked` needs to be added here as well.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		void INotifyBlockingMove.OnNotifyBlockingMove(Actor self, Actor blocking)
+		{
+			if (!self.IsDisabled() && BuildComplete && !isOpen && (this as ITemporaryBlocker).CanLiftBlockade(self, blocking))
+				Open();
```
penev92: Same here regarding `Locked`.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		bool ITemporaryBlocker.CanLiftBlockade(Actor self, Actor blocking)
+		{
+			return !self.IsDisabled() && BuildComplete && blocking.AppearsFriendlyTo(self);
+		}
+
+		void INotifyBlockingMove.OnNotifyBlockingMove(Actor self, Actor blocking)
+		{
+			if (!self.IsDisabled() && BuildComplete && !isOpen && (this as ITemporaryBlocker).CanLiftBlockade(self, blocking))
```
penev92: Is it me, or are you basically doing `this.CanLiftBlocade(self, self)` here at the end?

pchote:
Also don't forget about the animation stuff we talked about in IRC!

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
Phrohdoh: Not all of our interfaces are there.
For example `IPreventsEjectOnDeath`.

It just depends on if it is specific to a trait, do we have another use for this currently?

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+				Open();
+		}
+
+		bool IsBlocked()
+		{
+			var eligibleLocations = FootprintUtils.Tiles(self).ToList();
+			foreach (var loc in eligibleLocations)
+			{
+				var blockers = self.World.ActorMap.GetActorsAt(loc).Where(a => a != self);
+				if (blockers != null && blockers.Any())
```
penev92: This can't be `null`.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+
+		void INotifyBlockingMove.OnNotifyBlockingMove(Actor self, Actor blocking)
+		{
+			if (!self.IsDisabled() && BuildComplete && !isOpen && (this as ITemporaryBlocker).CanLiftBlockade(self, blocking))
+				Open();
+		}
+
+		bool IsBlocked()
+		{
+			var eligibleLocations = FootprintUtils.Tiles(self).ToList();
```
penev92: The `.ToList()` here should be unnecessary. You only enumerate it once in the `foreach`.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
penev92: Or perhaps change it to `isInTransition` and set it to true both while opening and closing?

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,144 @@
...
+			isOpen = true;
+			opening = false;
+			remainingTime = info.TicksToStayOpen;
+
+			self.World.ActorMap.RemoveInfluence(self, this);
+		}
+
+		void Close()
+		{
+			if (isOpen)
```
penev92: You probably want to make sure it's not currently opening/closing here.

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -571,6 +575,15 @@ public bool CanEnterCell(CPos cell, Actor ignoreActor = null, bool checkTransien
 			return Info.CanEnterCell(self.World, self, cell, ignoreActor, checkTransientActors ? CellConditions.All : CellConditions.BlockedByMovers);
 		}
 
+		public bool ContainsTemporaryBlocker(CPos cell)
```
penev92: This may fit better in `WorldUtils`. Just needs a `Actor ignoreActor = null` parameter like `CanEnterCell()`.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
pchote: All these ugly flags and state is going to go away - it will be changed to work like the production doors instead

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
teees: will change it, those are the ticks it takes to open the gate.
I made it independent of the actual animation, since ts has different animation lengths for nod/gdi. Gates are considered closed in transitions, that's why it's only used for opening

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
teees: I had it there before, when it was still called IGate but moved it since it's mod specific and isn't used in `OpenRA.Game`

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
teees: it checks for blocking units every 10 ticks. Checking every tick seemed unecessary to me

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
penev92: `OpenRA.Mods.Common` has its own `TraitInterfaces.cs`.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
teees: thanks, changed

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
teees: thanks, changed

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
penev92: OK, but so does the first `if` - it checks for blockers every `remainingTime` ticks and even does the same when there are blockers.

pchote: Just so we're clear, the idea here is to predict ahead of time that something is approaching so that it can start opening earlier, right?  The OnNotifyBlockingMove implementation will cover the case where it is actively blocking a unit.

teees: No, currently the door will only open when "nudged" from OnNotifyBlockingMove. It is always considered to be in the way, since it's invisible to the pathfinder (for units that can actually open it). There is no prediction happening, I don't remember this being done in tiberian sun either. `IsBlocked()` checks if there is an actor standing on an opened gate, which will then wait for another `info.TicksToStayOpen` to close.

Enemies can still move into your base when they sneak on an opened gate as its influenced is removed from `ActorMap` while open

mods/ts/rules/defaults.yaml
```diff
@@ -721,3 +721,28 @@
...
 
+^Gate:
+	Inherits: ^Building
+	Valued:
+		Cost: 250
+	Health:
+		HP: 350
+	Armor:
+		Type: Heavy
+	LineBuildNode:
```
pchote: This doesn't seem to work.  Walls are not visually connecting to the gates like they do for component towers.

pchote:
I've pushed a prototype of what I was talking about earlier to e6f06e2fa21fc5dea26ba461658982fc06af714f.  This separates the world logic from rendering, removes most of the places where edge cases can hide in the state logic, and gives some hints on how to fix the wall connection issues.

abcdefg30:
> OpenRA.Mods.Common/Traits/Buildings/Gate.cs:L15: [SA1505] An opening curly bracket must not be followed by a blank line.

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -271,6 +271,12 @@ public interface IMove
 		bool CanEnterTargetNow(Actor self, Target target);
 	}
 
+	public interface ITemporaryBlocker
+	{
+		bool CanRemoveBlocker(Actor self, Actor blocking);
+		bool Blocks();
```
pchote: `bool Blocks(Actor self, CPos cell)` please.  Certain actor types may want to block only certain cells.

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -271,6 +271,12 @@ public interface IMove
 		bool CanEnterTargetNow(Actor self, Target target);
 	}
 
+	public interface ITemporaryBlocker
+	{
+		bool CanRemoveBlocker(Actor self, Actor blocking);
```
pchote: Maybe `CanRemoveBlockage`?

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,15 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
 			}
 		}
 
+		public static bool ContainsTemporaryBlocker(this World world, CPos cell, Actor ignoreActor = null)
+		{
+			var temporaryBlocker = world.ActorMap.GetActorsAt(cell).FirstOrDefault();
```
pchote: This should loop over all the actors in the cell.

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,15 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
...
+			if (temporaryBlocker == null || temporaryBlocker == ignoreActor)
+				return false;
+			var temporaryBlockerTrait = temporaryBlocker.TraitOrDefault<ITemporaryBlocker>();
```
pchote: This should use `TraitsImplementing` and loop over all results.

teees:
@pchote I incorporated your changes, thank you

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,24 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
 			}
 		}
 
+		public static bool ContainsTemporaryBlocker(this World world, CPos cell, Actor ignoreActor = null)
+		{
+			var temporaryBlockers = world.ActorMap.GetActorsAt(cell);
+			if (temporaryBlockers == null)
+				return false;
+			foreach (var temporaryBlocker in temporaryBlockers)
```
penev92: Empty line after the `return`, please.

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,24 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
 			}
 		}
 
+		public static bool ContainsTemporaryBlocker(this World world, CPos cell, Actor ignoreActor = null)
```
penev92: You don't seem to be using `ignoreActor` at all.

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,24 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
...
+		{
+			var temporaryBlockers = world.ActorMap.GetActorsAt(cell);
+			if (temporaryBlockers == null)
+				return false;
+			foreach (var temporaryBlocker in temporaryBlockers)
+			{
+				var temporaryBlockerTraits = temporaryBlocker.TraitsImplementing<ITemporaryBlocker>();
+				if (temporaryBlockerTraits == null)
+					return false;
+				foreach (var temporaryBlockerTrait in temporaryBlockerTraits)
```
penev92: Empty line after the `return`, please.

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -271,6 +271,12 @@ public interface IMove
 		bool CanEnterTargetNow(Actor self, Target target);
 	}
 
+	public interface ITemporaryBlocker
+	{
+		bool CanRemoveBlockage(Actor self, Actor blocking);
+		bool Blocks(Actor self, CPos cell);
```
penev92: What happened to `IsBlocking()` ?

teees: Might have been reverted when I merged it with @pchote 's branch, good catch

OpenRA.Game/WorldUtils.cs
teees: fixed, thanks

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,29 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
...
 
+		public static bool ContainsTemporaryBlocker(this World world, CPos cell, Actor ignoreActor = null)
+		{
+			var temporaryBlockers = world.ActorMap.GetActorsAt(cell);
+			if (temporaryBlockers == null)
+				return false;
+
+			foreach (var temporaryBlocker in temporaryBlockers)
+			{
+				if (temporaryBlocker == ignoreActor)
```
pchote: `ignoreActor` is used here.

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,29 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
...
+					return false;
```
pchote: This needs to be a continue, not a return.  There may be another blocking actor in the cell!

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,29 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
...
+
+				var temporaryBlockerTraits = temporaryBlocker.TraitsImplementing<ITemporaryBlocker>();
+				if (temporaryBlockerTraits == null)
+					return false;
```
pchote: `continue` here too.

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,29 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
...
+
+				foreach (var temporaryBlockerTrait in temporaryBlockerTraits)
+					if (temporaryBlockerTrait.IsBlocking(temporaryBlocker, cell))
+						return true;
```
pchote: This wants to stay as `return true;` since you know know that there definitely is a blocker.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,117 @@
...
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Will open and be passable for actors that appear friendly when there are no enemies in range.")]
+	public class GateInfo : BuildingInfo, ITraitInfo
+	{
+		public readonly string OpeningSound = "gateup1.aud";
+		public readonly string ClosingSound = "gatedwn1.aud";
```
pchote: Can we default these to `null` and then define them in yaml?

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,117 @@
...
+
+				// Gate is now fully open
+				if (Position == OpenPosition)
+				{
+					self.World.ActorMap.RemoveInfluence(self, this);
+					remainingOpenTime = info.TicksToStayOpen;
+				}
+			}
+
+			if (Position == OpenPosition)
```
pchote: braces around the body here

OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
```diff
@@ -0,0 +1,91 @@
...
+		public override IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer init, RenderSpritesInfo rs, string image, int facings, PaletteReference p)
+		{
+			var anim = new Animation(init.World, image);
+			anim.PlayFetchIndex(RenderSprites.NormalizeSequence(anim, init.GetDamageState(), Sequence), () => 0);
+
+			yield return new SpriteActorPreview(anim, WVec.Zero, 0, p, rs.Scale);
+		}
+	}
+
+	// TODO: Add a new interface type for "things that walls can connect to.  It will have a method for AdjacentWallCanConnect
```
pchote: This was left as an exercise for you to finish :)

OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
```diff
@@ -0,0 +1,91 @@
...
+
+		public override void BuildingComplete(Actor self)
+		{
+			DefaultAnimation.PlayFetchIndex(NormalizeSequence(self, Info.Sequence), GetGateFrame);
+			UpdateNeighbours(self);
+		}
+
+		static void UpdateNeighbours(Actor self)
+		{
+			// This code should stay basically the same, but requires an extra change in WithWallSpriteBody
```
pchote: this too

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
penev92: What's the standart? `null` or "" ? I know we're probably not very consistent with those, but still.

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,29 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
 			}
 		}
 
+		public static bool ContainsTemporaryBlocker(this World world, CPos cell, Actor ignoreActor = null)
+		{
+			var temporaryBlockers = world.ActorMap.GetActorsAt(cell);
+			if (temporaryBlockers == null)
```
RoosterDragon: `GetActorsAt` will never return null, no need to check for it.

OpenRA.Game/WorldUtils.cs
```diff
@@ -41,6 +41,29 @@ public static IEnumerable<Actor> FindActorsInCircle(this World world, WPos origi
...
+				return false;
+
+			foreach (var temporaryBlocker in temporaryBlockers)
+			{
+				if (temporaryBlocker == ignoreActor)
+					continue;
+
+				var temporaryBlockerTraits = temporaryBlocker.TraitsImplementing<ITemporaryBlocker>();
```
RoosterDragon: `TraitsImplementing` will never return null, no need to check for it.

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -251,6 +251,10 @@ bool IsBlockedBy(Actor self, Actor otherActor, Actor ignoreActor, CellConditions
 					return true;
 			}
 
+			var temporaryBlocker = otherActor.TraitOrDefault<ITemporaryBlocker>();
+			if (temporaryBlocker != null && temporaryBlocker.CanRemoveBlockage(otherActor, self))
+				return false;
```
RoosterDragon: This seems like it should go above all the crush checks to me.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,119 @@
...
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Will open and be passable for actors that appear friendly when there are no enemies in range.")]
+	public class GateInfo : BuildingInfo, ITraitInfo
```
pchote: This extra `ITraitInfo` isn't needed (this is inherited from `BuildingInfo`).

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,119 @@
...
+		public readonly int OpenPosition;
+		[Sync] public int Position { get; private set; }
+		int desiredPosition;
+		int remainingOpenTime;
+
+		public Gate(ActorInitializer init, GateInfo info)
+			: base(init, info)
+		{
+			this.info = info;
+			this.self = init.Self;
```
pchote: You don't need the `this` here.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,119 @@
...
+	public class GateInfo : BuildingInfo, ITraitInfo
+	{
+		public readonly string OpeningSound = null;
+		public readonly string ClosingSound = null;
+
+		[Desc("Ticks until the gate closes.")]
+		public readonly int TicksToStayOpen = 50;
+
+		[Desc("Ticks until the gate is considered open.")]
+		public readonly int TicksUntilOpen = 10;
```
pchote: Can we name this something like `TransitionDelay`?  It works in both directions, after all.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,119 @@
...
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Will open and be passable for actors that appear friendly when there are no enemies in range.")]
+	public class GateInfo : BuildingInfo, ITraitInfo
+	{
+		public readonly string OpeningSound = null;
+		public readonly string ClosingSound = null;
+
+		[Desc("Ticks until the gate closes.")]
+		public readonly int TicksToStayOpen = 50;
```
pchote: This should also have `Delay` somewhere in the name (that's our convention for things that are waits measured in ticks)

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,119 @@
...
+				if (IsBlocked())
+					remainingOpenTime = info.TicksToStayOpen;
+				else if (--remainingOpenTime <= 0)
+					desiredPosition = 0;
+			}
+		}
+
+		bool ITemporaryBlocker.IsBlocking(Actor self, CPos cell)
+		{
+			return Position != OpenPosition && FootprintUtils.Tiles(self).Contains(cell);
```
pchote: It would be worth implementing `INotifyAddedToWorld` and then caching `FootprintUtils.Tiles(self)`, because that does a bunch of work and this gets called a lot.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,119 @@
...
+
+		void INotifyBlockingMove.OnNotifyBlockingMove(Actor self, Actor blocking)
+		{
+			if (Position != OpenPosition && (this as ITemporaryBlocker).CanRemoveBlockage(self, blocking))
+				desiredPosition = OpenPosition;
+		}
+
+		bool IsBlocked()
+		{
+			var eligibleLocations = FootprintUtils.Tiles(self);
```
pchote: Likewise here.

OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
```diff
@@ -97,8 +104,8 @@ public void Tick(Actor self)
 			adjacent = 0;
 			foreach (var a in adjacentActors)
 			{
-				var rb = a.TraitOrDefault<WithWallSpriteBody>();
-				if (rb == null || rb.wallInfo.Type != wallInfo.Type)
+				var rb = a.TraitOrDefault<IWallConnector>();
```
pchote: Style nit: can we rename this to `wc` to match the new type?

OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
```diff
@@ -97,8 +104,8 @@ public void Tick(Actor self)
...
+				if (rb == null || !rb.AdjacentWallCanConnect(self, self.Location, wallInfo.Type))
```
pchote: This first arg needs to be `a` (since you're talking to a trait on the gate actor, not the wall).

pchote:
Here's some more commits to make this work better:
* Support loading `CVec[]` from yaml: 6708c4641d230b5ef03092ea61026d964ab70d3e
* Fix wall adjacency wrt multi-cell footprints: c783c1013c74e9138f58ff7b00fb1fc9f8d370ad
* Fix linebuild adjacency wrt multi-cell footprints: 98236b56366f10537d727b420b88d461f0bdd33a
* Fix gate adjacency wrt multi-cell footprints: a36dacb0b6d9a23175c29f78f596d71f22745545
* Yaml changes to set up the proper wall connections: 5d83ee1d990296bbdfc54ba12932d323af9e926f

These fix the last of the major issues I can find.

OpenRA.Mods.Common/Traits/Mobile.cs
teees: thanks, fixed

OpenRA.Game/WorldUtils.cs
teees: thanks, fixed

teees:
Will merge/rebase that fixup as soon as everything is sorted out

OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
```diff
@@ -97,20 +106,18 @@ public void Tick(Actor self)
 			adjacent = 0;
 			foreach (var a in adjacentActors)
 			{
-				var rb = a.TraitOrDefault<WithWallSpriteBody>();
-				if (rb == null || rb.wallInfo.Type != wallInfo.Type)
+				CVec facing;
+				var wc = a.TraitsImplementing<IWallConnector>().FirstOrDefault(b => b.IsTraitEnabled());
```
pchote: I think this should be `.FirstOrDefault(Exts.IsTraitEnabled);`?

mods/ts/sequences/structures.yaml
```diff
@@ -524,6 +524,84 @@ gawall:
 		Offset: 0, 0
 		UseTilesetCode: false
 
+gagate_a:
+	Defaults:
+		Offset: -24, -24
+		UseTilesetCode: true
+	idle:
+		Length: 10
```
Mailaender: You can add `		ShadowStart: 21` here

mods/ts/sequences/structures.yaml
```diff
@@ -524,6 +524,84 @@ gawall:
...
+	damaged-idle:
+		Start: 10
+		Length: 10
```
Mailaender: You can add `		ShadowStart: 21` here as well.

mods/ts/sequences/structures.yaml
GraionDilach: `20`, not `21`.

teees: 21 should be ok I think, [0-9] is idle, [10-19] is damaged-idle, 20 is dead, [21-30] should then be idle shadow etc

OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
atlimit8: This is an unnecessary lambda expression and the code usually uses `Where/FirstOrDefault/etc(Exts.IsTraitEnabled)`.
When I add the `Trait` base class, I plan on adding `public static bool IsEnabled` so that `Exts.IsTraitEnabled` is replaced by `Trait.IsEnabled`.

teees:
I added an IBlocksProjectiles interface to allow linear scaling of blocksprojectiles height as well as disabling the trait (for firestorm gen)

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -237,6 +237,11 @@ bool IsBlockedBy(Actor self, Actor otherActor, Actor ignoreActor, CellConditions
 				IsMovingInMyDirection(self, otherActor))
 				return false;
 
+			// If there is a closed gate / temporary blocker in our way and we can't open it, we are blocked.
```
RoosterDragon: Could this be changed to something like `We are not blocked by a temporary blocker that we can remove.` so you don't have to mentally negate the comment to match the code?

(I'm basically just asking this to be consistent with the other code & comments - parts that return false use "we are not blocked" whilst parts that return true use "we are blocked")

teees: changed it, thanks

teees:
rebased

teees:
rebased once again

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -282,6 +282,12 @@ public interface IMove
 		bool CanEnterTargetNow(Actor self, Target target);
 	}
 
+	public interface ITemporaryBlocker
+	{
+		bool CanRemoveBlockage(Actor self, Actor blocking);
```
Phrohdoh: This interface suggest the implementing trait (actor) is a blocker.
If so, why would they want to remove other blockage?

OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
```diff
@@ -13,6 +13,11 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
+	public interface IBlocksProjectiles
+	{
+		WDist GetHeight(Actor self);
```
Phrohdoh: I think this could be named better, but this is probably in line with the current projectile code.

OpenRA.Game/Traits/TraitsInterfaces.cs
teees: I guess you are referencing the `Actor self` argument? This was done to conform with the ongoing convention to have `self` as first argument. Not really any other reason.

OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
teees: Traits would get the height directly from the traits info before, that's why it's just simply a getter now. Suggestions on a better name?

pchote: Could become `WDist BlockingHeight { get; }`?

OpenRA.Game/Traits/TraitsInterfaces.cs
Phrohdoh: I mean having this method at all.

Phrohdoh:
screen shot 2015-12-16 at 7 52 35 pm

OpenRA.Game/Traits/TraitsInterfaces.cs
teees: The interface suggests the actor is a _temporary_ blocker, `CanRemoveBlockage` will return different results for different `blocking` actors or when powered down etc. This is essentially how gates work, the pathfinder will use this function to see if the unit can potentially remove the blockage and if yes will just "bump" into it.

teees:
iirc @pchote mentioned that this will be resolved with proper z-buffer sorting

OpenRA.Game/Traits/TraitsInterfaces.cs
Phrohdoh: This still makes no sense to me, but clearly I'm the only one, so ok.

pchote: @Phrohdoh: `blocking` refers the actor that is blocked (i.e. the unit that wants to go through the gate), not another blocking actor actor that `self` may be able to remove.

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
```diff
@@ -28,14 +30,14 @@ public class CombatDebugOverlay : IPostRender, INotifyDamage
...
-		readonly BlocksProjectilesInfo blockInfo;
+		readonly IBlocksProjectiles[] blocker;
 		Lazy<AttackBase> attack;
 		Lazy<BodyOrientation> coords;
 
 		public CombatDebugOverlay(Actor self)
 		{
 			healthInfo = self.Info.TraitInfoOrDefault<HealthInfo>();
-			blockInfo = self.Info.TraitInfoOrDefault<BlocksProjectilesInfo>();
+			blocker = self.TraitsImplementing<IBlocksProjectiles>().ToArray();
```
RoosterDragon: Any traits implementing `IBlocksProjectiles` are not guaranteed to have been created by the time you perform this lookup. Either wrap it in a `Lazy<T>` like the others do, or implement `INotifyCreated` and do your trait lookups in there (the traits will be ready by then).

teees:
rebased and adapted to @RoosterDragon's requested changes

teees:
and fixed my merge conflicts :/

mods/ts/rules/defaults.yaml
Mailaender: This seems to be resolved by now:

image

Mailaender:
You can fix the sprite overlapping for the open mode
```
		ZOffset: -1c511
```
however that will conflict with the closed mode.

You may need a dynamic Z offset that adapts to the closed and open mode.

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
```diff
@@ -53,10 +59,11 @@ public void RenderAfterWorld(WorldRenderer wr, Actor self)
 			if (healthInfo != null)
 				healthInfo.Shape.DrawCombatOverlay(wr, wcr, self);
 
-			if (blockInfo != null)
+			var block = blocker.Where(Exts.IsTraitEnabled).ToList();
+			if (block.Any())
```
RoosterDragon: Maybe `blocker` -> `allBlockers` and `block` -> `blockers` for consistency with elsewhere in your PR where `blockers` are the enabled `IBlocksProjectiles` traits.

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -241,6 +241,11 @@ bool IsBlockedBy(Actor self, Actor otherActor, Actor ignoreActor, CellConditions
 				IsMovingInMyDirection(self, otherActor))
 				return false;
 
+			// If there is a temporary blocker in our path, but we can remove it, we are not blocked
```
RoosterDragon: Needs a full stop.

pchote:
> You may need a dynamic Z offset that adapts to the closed and open mode.

This has been discussed multiple times now, and

> iirc @pchote mentioned that this will be resolved with proper z-buffer sorting

teees:
rebased and fixed

mods/ts/rules/defaults.yaml
pchote: Yes.  I fixed that in e58ef101839f1d4ed4bccd3001ca7a6d011e7adc.

mods/ts/rules/gdi-support.yaml
```diff
@@ -25,6 +25,26 @@ GAWALL:
...
 
+GAGATE_A:
+	Inherits: ^Gate_A
+	Buildable:
+		Queue: Defense
+		BuildPaletteOrder: 100
+		Prerequisites: gapile, ~structures.gdi
+	Tooltip:
+		Name: GDI Gate
+		Description: Stops infantry and blocks enemy fire.\nCan NOT be crushed by tanks.
```
pchote: I don't think anybody would expect gates to be crushable by tanks?  Seems like redundant info to me.

teees: Oh you got me - I copy pasted that, probably from walls.

Phrohdoh: Are walls even crushable in vanilla TS?

MustaphaTR: Sandbags are but walls aren't.

pchote: Kind of hard to come up with a description for something as simple as a gate... I suggest changing this to something like "Allows friendly units to pass through walls".

teees: I changed it to "Automated barrier that opens up for allied units."

Mailaender:
It does block bullets shot diagonally when gates are open and shouldn't block anything:

image

Mailaender:
Looks like bullets can pass closed gates when shot through a straight line:

image

MustaphaTR:
Bullets and other insta-hit projectiles have problems with blocking even on walls. It is still not fixed iirc.

pchote:
@MustaphaTR: that was fixed in #10243.

pchote:
Projectiles will be able to pass through the sides of the gate because we do not yet have a way to specify a proper hitbox for the gate (see #10349).  This doesn't explain why bullets are blocked by open gates, which is probably a bug.

teees:
Open gates shouldn't block anymore, the other issue is outside of the scope of this pr, as it is an already existing issue (see any line damage projectile missing (big) buildings if it doesn't traverse over the hitbox https://streamable.com/yoma )

Mailaender:
Works as promised. :white_check_mark: I consider the path-finding and wall integration very well done. Great work both @teees and @pchote :+1:

GraionDilach:
Can't the rectangle hitbox solve the footprint issue?

pchote:
Nope. The rectangle is in screen space.

pchote:
This already has my (so far implicit) :+1:, and nobody else seems to want to review it so I'll merge this tomorrow if no new review comments are made.

OpenRA.Game/FieldLoader.cs
```diff
@@ -373,6 +373,28 @@ public static object GetValue(string fieldName, Type fieldType, MiniYaml yaml, M
...
+			{
+				if (value != null)
+				{
+					var parts = value.Split(',');
+
+					if (parts.Length % 2 != 0)
+						return InvalidValueAction(value, fieldType, fieldName);
+
+					var vecs = new CVec[parts.Length / 2];
+					for (var i = 0; i < vecs.Length; ++i)
```
abcdefg30: `i++`, please.

> (15:51:25) abcdefg30 [...] just wanted to know if we have a code style rule about that
(15:51:38) pchote nothing written down, but i++ is used more
(15:51:42) pchote so we should use that

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -220,12 +220,14 @@ void SanityCheckPath(Mobile mobile)
 
 			var nextCell = path[path.Count - 1];
 
+			var containsTemporaryBlocker = WorldUtils.ContainsTemporaryBlocker(self.World, nextCell, self);
+
 			// Next cell in the move is blocked by another actor
-			if (!mobile.CanMoveFreelyInto(nextCell, ignoredActor, true))
+			if (!mobile.CanMoveFreelyInto(nextCell, ignoredActor, true) || containsTemporaryBlocker)
```
abcdefg30: Would the order of the checks matter here? Like if `containsTemporaryBlockers` is already `true`, we'd save us the `CanMoveFreelyInto` check with this setup:
`if (containsTemporaryBlocker || !mobile.CanMoveFreelyInto(nextCell, ignoredActor, true))`?

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -220,12 +220,14 @@ void SanityCheckPath(Mobile mobile)
...
 			{
 				// Are we close enough?
 				var cellRange = nearEnough.Length / 1024;
-				if ((mobile.ToCell - destination.Value).LengthSquared <= cellRange * cellRange)
+				if ((mobile.ToCell - destination.Value).LengthSquared <= cellRange * cellRange && !containsTemporaryBlocker)
```
abcdefg30: Similar question here.

OpenRA.Mods.Common/Traits/Buildings/BuildingUtils.cs
```diff
@@ -72,12 +72,11 @@ public static IEnumerable<CPos> GetLineBuildCells(World world, CPos location, st
...
 					// Cell contains an actor. Is it the type we want?
-					if (world.ActorsHavingTrait<LineBuildNode>()
-						.Any(a => a.Location == cell
-							&& a.Info.TraitInfo<LineBuildNodeInfo>().Types.Overlaps(lbi.NodeTypes)))
-						dirs[d] = i; // Cell contains actor of correct type
-					else
-						dirs[d] = -1; // Cell is blocked by another actor type
+					var hasConnector = world.ActorMap.GetActorsAt(cell)
+						.Any(a => a.Info.TraitInfos<LineBuildNodeInfo>()
+							.Any(info => info.Types.Overlaps(lbi.NodeTypes) && info.Connections.Contains(vecs[d])));
```
abcdefg30: I don't think we want another level of indentation here, do we?

pchote: It is nested inside the first `Any`, so the indentation is wanted.

abcdefg30:
What is with https://github.com/OpenRA/OpenRA/pull/9981#discussion_r44780501?

teees:
It still derives from the Building class, so all the regular building locks should work as usual I'd guess. It would probably only matter if you want different behaviour for capturing closed and opened gates.

IMO gates count as base defense and shouldn't even be capturable - You are of course free to provide a patch for that in another pr.

TS gates weren't capturable https://github.com/LipkeGu/TSIniFiles/blob/master/rules.ini#L4895

OpenRA.Mods.Common/Activities/Move/Move.cs
teees: The chance that the cell actually contains a temporary blocker is quite slim. Even in that case it wouldn't save you anything that much.

RoosterDragon:
If another mod wishes to have capturable gates, we should make sure that works correctly (or at least make sure it doesn't make bad things happen and leave a TODO somewhere).

I'm not familiar enough with `Locked` to know what needs to happen (if anything) - @penev92 can you clarify?

teees:
Will add blocking then, probably tomorrow. Testing it however is tricky, see http://logs.openra.net/index.php?year=2016&month=01&day=17#13:35:12

teees:
rebased and added check for `Locked`

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		bool ITemporaryBlocker.CanRemoveBlockage(Actor self, Actor blocking)
+		{
+			return !self.IsDisabled() && BuildComplete && blocking.AppearsFriendlyTo(self);
+		}
+
+		void INotifyBlockingMove.OnNotifyBlockingMove(Actor self, Actor blocking)
+		{
+			if (Position != OpenPosition && (this as ITemporaryBlocker).CanRemoveBlockage(self, blocking))
```
RoosterDragon: `((ITemporaryBlocker)this)` please.

Side comment: Needing this cast is icky - a sad side effect of the explicit interfaces. (Don't change the interfaces or anything - I'm just mulling aloud here)

pchote: IMO a better pattern would be to have a private implementation that is called from both places.

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
```diff
@@ -0,0 +1,137 @@
...
+		bool IsBlocked()
+		{
+			foreach (var loc in blockedPositions)
+			{
+				var blockers = self.World.ActorMap.GetActorsAt(loc).Where(a => a != self);
+				if (blockers.Any())
+					return true;
+			}
+
+			return false;
```
RoosterDragon: There's an odd blend of imperative and functional style in this function.

I would choose one of these and stick with it:
```c#
return blockedPositions.Any(loc => self.World.ActorMap.GetActorsAt(loc).Any(a => a != self));
```

```c#
foreach (var loc in blockedPositions)
    foreach (var blocker in self.World.ActorMap.GetActorsAt(loc))
        if (blocker != self)
            return true;

return false;
```

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
```diff
@@ -53,10 +59,11 @@ public void RenderAfterWorld(WorldRenderer wr, Actor self)
 			if (healthInfo != null)
 				healthInfo.Shape.DrawCombatOverlay(wr, wcr, self);
 
-			if (blockInfo != null)
+			var blockers = allBlockers.Where(Exts.IsTraitEnabled).ToList();
+			if (blockers.Any())
```
RoosterDragon: Prefer `.Count > 0` to `.Any()`.

OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
```diff
@@ -0,0 +1,93 @@
...
+			UpdateNeighbours(self);
+		}
+
+		void UpdateNeighbours(Actor self)
+		{
+			var footprint = FootprintUtils.Tiles(self).ToArray();
+			var adjacent = OpenRA.Traits.Util.ExpandFootprint(footprint, true).Except(footprint)
+				.Where(self.World.Map.Contains).ToList();
+
+			var adjacentActorTraits = adjacent.SelectMany(c => self.World.ActorMap.GetActorsAt(c))
```
RoosterDragon: `.SelectMany(self.World.ActorMap.GetActorsAt)`

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
RoosterDragon: Works for me.

RoosterDragon:
There's a few loose ends of polish that have been picked up that we can clean up in the future (placement over walls, z-offset, blocking projectiles more correctly) but this is a very cool feature and the core mechanic of blocking and allowing movement seems to work just fine as far as I can see. I have a few last points of pedantry over the code style.

:+1: "Neat"

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
teees: fixed

OpenRA.Mods.Common/Traits/Buildings/Gate.cs
teees: fixed

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
teees: replaced

OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
teees: fixed

teees:
changed and rebased :)

RoosterDragon:
Needs another rebase.

Mailaender:
You forgot to account for the recent changes in https://github.com/OpenRA/OpenRA/pull/10537.

teees:
should be fixed again

OpenRA.Mods.Common/Traits/Render/WithWallSpriteBody.cs
```diff
@@ -39,7 +40,7 @@ public override IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInit
 					var haveNeighbour = false;
 					foreach (var n in kv.Value)
 					{
-						var rb = init.World.Map.Rules.Actors[n].TraitInfoOrDefault<WithWallSpriteBodyInfo>();
+						var rb = init.World.Map.Rules.Actors[n].TraitInfos<WithWallSpriteBodyInfo>().FirstOrDefault(Exts.IsTraitEnabled);
```
pchote: Apologies for throwing up more roadblocks for getting this merged, but we're going to need an `IWallConnectorInfo` interface here to make sure that walls connect to gates in the map editor.

I'd also be happy to defer this to a followup PR.

OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -116,4 +116,10 @@ public interface IPreventsAutoTarget
 	{
 		bool PreventsAutoTarget(Actor self, Actor attacker);
 	}
+
+	interface IWallConnector
```
pchote: Could we get a `[RequireExplicitImplementation]` here?

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -284,6 +284,12 @@ public interface IMove
 		bool CanEnterTargetNow(Actor self, Target target);
 	}
 
+	public interface ITemporaryBlocker
```
pchote: Could we get a `[RequireExplicitImplementation]` here too?  Now that we have that interface we may as well use it.

OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
```diff
@@ -12,6 +12,11 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
+	public interface IBlocksProjectiles
```
pchote: and a `[RequireExplicitImplementation]` here too if that doesn't require too many followup changes.

obrakmann:
Gates still have some z-ordering issues:
screenshot from 2016-01-24 16 44 49

However, I'm fine with merging this as-is and fixing the remaining issues in follow-ups. At least those should be merge-able faster than this one.

teees:
Incorporated the requested changes

teees:
And rebased

Mailaender:
Still works, even after all those code reviews and rebases. Thanks for your endurance!

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/e623f0c014c33a384963396af475c96a3d0322a7...e1345029f195eddef5cc13df720becdcf77e3765)

PR #9988, opened 2015-11-14
Fix CellRegion.BoundingRegion behaviour.

pchote:
CellRegions use MPos internally, and so the bounding area must be calculated in these coordinates in order for it to behave properly in the isometric mods.  This fixes the wonky copy/paste selection area in the TS map editor.


RoosterDragon:
:+1:

Mailaender:
Still a bit wonky as the pase size changes when moving the selection one cell left/right:

image

image

obrakmann:
The shape-changing is due to the weird-ass way the coordinate system in ts works. This otherwise works as expected. :+1: 

Copy-pasting tiles with different heights still leaves gaps in the map, but that isn't due to this code.

PR #9993, opened 2015-11-14
Switch to a (hopefully) working NSProcess url.

pchote:
References the file uploaded by @Mailaender at https://build.opensuse.org/request/show/289794


obrakmann:
Wget works for me on that URL. :+1:

Issue #2598, opened 2013-01-17
Visualize waypoints set by the player


Mailaender:
You can currently use Shift + Click to set waypoints for your units. There should be a flag (suggesting the one from the rally point) to visualize this.

x-a-n-a-x:
+1 for all these ideas.

r34ch:
I didn't think way-points were implemented in the game! This will ease microing significantly...

I assume the waypoint logic could be extended to make nicer visual scripting waypoints in the ingame map editor?

huwpascoe:
I'm now taking on this issue.

Turupawn:
Hey guys! I'm new here and I just started working on this issue.
So far I managed to visualize the waypoint lines.
I made a [video showing my progress](https://youtu.be/jFSLQFePDqA).

I'm currently working on [this branch](https://github.com/Turupawn/OpenRA/tree/queued_lines) but I haven't made a Pull Request yet because the code is **very** messy. But I'll do so soon. Also, I'm waiting to get my other PR approved so I can understand better the workflow, see #11781 

Is this thing still alive? What do you think about my progress?

Cheers!

Mailaender:
I also tried my luck with it, but failed to achieve anything usable. Your progress looks good. :+1:

Turupawn:
I'll keep working on it.
Thanks! <3

pchote:
I strongly suggest using the existing `Activity.GetTargets` code instead of rolling your own.

Turupawn:
@pchote Sure! I'll do so :ok_hand:

Turupawn:
BTW, Thanks for adding me to @OpenRA/bug-assignees. I don't know exactly what that means but I really apreciate it. :blossom:

Mailaender:
You can now show others that you want to pick up a task.

GraionDilach:
If you mean "letting issues assigned to yourself", then sorry to tell, but that never worked. Would have done that sometimes already by now if it would have.

Turupawn:
@Mailaender  Cool!

Turupawn:
Hello again guys!
A quick update about my progress:

* I just managed to make this work without rolling my own tech at #11794
* I just made a [short video](https://youtu.be/A2HpUPjA030) that shows how they look now
* @pchote: I didn't use `Activity.GetTargets` but `GetDestination` which is working just fine, what do you think about it?
[...]

PR #9994, opened 2015-11-14
TargetLine for Rally Points

huwpascoe:
Also enabled allied players to see waypoints.
rally
Part of https://github.com/OpenRA/OpenRA/issues/2598

### DIFF
--- a/OpenRA.Mods.Common/Effects/RallyPointIndicator.cs
+++ b/OpenRA.Mods.Common/Effects/RallyPointIndicator.cs
@@ -22,11 +22,16 @@ namespace OpenRA.Mods.Common.Effects
 		readonly RallyPoint rp;
 		readonly Animation flag;
 		readonly Animation circles;
+		readonly ExitInfo[] exits;
 
-		public RallyPointIndicator(Actor building, RallyPoint rp)
+		readonly WPos[] targetLine = new WPos[2];
+		CPos cachedLocation;
+
+		public RallyPointIndicator(Actor building, RallyPoint rp, ExitInfo[] exits)
 		{
 			this.building = building;
 			this.rp = rp;
+			this.exits = exits;
 
 			flag = new Animation(building.World, rp.Info.Image);
 			flag.PlayRepeating(rp.Info.FlagSequence);
@@ -35,7 +40,6 @@ namespace OpenRA.Mods.Common.Effects
 			circles.Play(rp.Info.CirclesSequence);
 		}
 
-		CPos cachedLocation;
 		public void Tick(World world)
 		{
 			flag.Tick();
@@ -44,6 +48,26 @@ namespace OpenRA.Mods.Common.Effects
 			if (cachedLocation != rp.Location)
 			{
 				cachedLocation = rp.Location;
+
+				var rallyPos = world.Map.CenterOfCell(cachedLocation);
+				var exitPos = building.CenterPosition;
+
+				// Find closest exit
+				var dist = int.MaxValue;
+				foreach (var exit in exits)
+				{
+					var ep = building.CenterPosition + exit.SpawnOffset;
+					var len = (rallyPos - ep).Length;
+					if (len < dist)
+					{
+						dist = len;
+						exitPos = ep;
+					}
+				}
+
+				targetLine[0] = exitPos;
+				targetLine[1] = rallyPos;
+
 				circles.Play(rp.Info.CirclesSequence);
 			}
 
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Effects/RallyPointIndicator.cs
```diff
@@ -53,15 +53,23 @@ public void Tick(World world)
...
-				return SpriteRenderable.None;
+			if (!building.World.Selection.Actors.Contains(building))
+				yield break;
 
 			var pos = wr.World.Map.CenterOfCell(cachedLocation);
 			var palette = wr.Palette(rp.PaletteName);
-			return circles.Render(pos, palette).Concat(flag.Render(pos, palette));
+
+			yield return new TargetLineRenderable(
+				new[] { building.CenterPosition, pos }, building.Owner.Color.RGB);
```
Phrohdoh: Would it be more correct to start the line at an `Exit`?

huwpascoe:
Didn't know there was an exit property!

huwpascoe:
Okay now points from exit

atlimit8:
Perhaps there should be lines from each exit if a line is drawn from one exit.

huwpascoe:
Won't that look messy?

penev92:
It probably will if someone decides they want 6-8-10 exits on their structure.

pchote:
We already have production structures with 4 exits (naval yard / sub pen) and 3 exits (war factory).

atlimit8:
Perhaps the closest exit to the rally point rather than the first listed then.

huwpascoe:
Does the building actually default to the nearest point to rally when producing?

penev92:
No, it uses a random available one, iirc.

huwpascoe:
Okay I can set to closest one. All agreed?

GraionDilach:
Don't change the exit point usage. Apparently blocking exits is a common multiplayer tactic and it was just on this weekend to make the Allied Barracks' two exits less vulnerable against this tactic by not sharing their exitcells.

If you change that to just use the closest exit for the rally point, that's asking for trouble on MP level.

penev92:
Perhaps we should just go back to the building's center and save ourselves the troubles?

pchote:
People who block exits already know what cells to block, so I don't see that as being a good justification for not using the exits here.

GraionDilach:
Oh, for some reason I thought @huwpascoe also means to change the exit usage at producing as well.

huwpascoe:
now the closest exit to the rally point is used

OpenRA.Mods.Common/Effects/RallyPointIndicator.cs
```diff
@@ -22,11 +22,16 @@ class RallyPointIndicator : IEffect
 		readonly RallyPoint rp;
 		readonly Animation flag;
 		readonly Animation circles;
+		readonly IEnumerable<ExitInfo> exits;
 
-		public RallyPointIndicator(Actor building, RallyPoint rp)
+		CPos cachedLocation;
+		WPos[] targetLine;
+
+		public RallyPointIndicator(Actor building, RallyPoint rp, IEnumerable<ExitInfo> exits)
```
RoosterDragon: Can you either accept an array of exits, or call `ToArray` in the constructor please?

OpenRA.Mods.Common/Effects/RallyPointIndicator.cs
```diff
@@ -44,6 +48,24 @@ public void Tick(World world)
...
+					var ep = building.CenterPosition + exit.SpawnOffset;
+					var len = (rallyPos - ep).Length;
+					if (len < dist)
+					{
+						dist = len;
+						exitPos = ep;
+					}
+				}
+
+				targetLine = new[] { exitPos, rallyPos };
```
RoosterDragon: Please initialize this (`WPos[] targetLine = new WPos[2];`) and just update the existing array rather than creating a new one each time.

OpenRA.Mods.Common/Effects/RallyPointIndicator.cs
```diff
@@ -53,15 +75,21 @@ public void Tick(World world)
...
 		{
-			if (building.Owner != building.World.LocalPlayer)
-				return SpriteRenderable.None;
+			if (!building.IsInWorld || !building.Owner.IsAlliedWith(building.World.LocalPlayer))
+				yield break;
 
-			if (!building.IsInWorld || !building.World.Selection.Actors.Contains(building))
-				return SpriteRenderable.None;
+			if (!building.World.Selection.Actors.Contains(building))
+				yield break;
```
RoosterDragon: Please continue to use `return SpriteRenderable.None` here - it's more efficient. The actual rendering code can continue to use `yield` if you create another method for it.

RoosterDragon:
I like this.

huwpascoe:
Applied rooster's requests

RoosterDragon:
:+1:

Mailaender:
This does not yet adhere to `Game.Settings.Game.DrawTargetLine`.

huwpascoe:
added settings

Mailaender:
Well done!

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/0e196bf397dc6be759456e1a15245a56d3127159...3acd6bc6f843acae6a4ef834e4efa13546435e19)

huwpascoe:
yay

PR #10007, opened 2015-11-15
Don't attack actors outside vision range

### LINKED ISSUE #9999: AA-gun keeps shooting at units that are out of its vision, but in its range
1. Run 2 clients
2. Place an AA-gun (client 1)
3. Build a Mig and try to direct it into the AA-gun (client 2)
4. Once you're in its eyesight, try to evade it (hide behind the shroud again)
5. Evidence your Mig dying, because AA-gun was shooting into the shroud.

penev92:
Fixes  #9999.
We may (or may not) want this for prep.

Note: better test with and without GPS active!

### DIFF
--- a/OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
@@ -89,6 +89,9 @@ namespace OpenRA.Mods.Common.Traits
 			if (self.IsDisabled())
 				return false;
 
+			if (target.Type == TargetType.Actor && !self.Owner.CanTargetActor(target.Actor))
+				return false;
+
 			return true;
 		}
 


pi99y:
Does not work with GPS. Units are still targeted.
Replay: http://www.megafileupload.com/dSX8/GAPGPSattack.orarep
Map: http://www.megafileupload.com/dSX9/gapsy.oramap

(If you want to test on your own, just start the map with debug menu and click enable instant charges. Once the GPS reveals the map the AA will start shooting at Hind under the GAP)

UberWaffe:
Should we perhaps just define the expected behavior in the different cases to ensure everyone is on the same page?

Within Range + Can see = Fire at enemy

Outside Range = Cannot fire at enemy
Within Range + Cannot see = Cannot fire at enemy.

Can see = [ Within sight range OR ( Outside of sight + GPS icon is visible ) ] + target not stealthed + target not disguised.
GPS icon is visible = GPS available for shooter + target not under GAP shroud.

penev92:
OK after a discussion on IRC: the GPS has its own set of issues (well there's a surprise!).
I think this fix is still valid, and fixing the (semi-related) GPS issues should enable this to cover all cases.

pchote:
The GPS issues are unrelated, so I think they're best covered in their own issue/PR so they don't hold up this fix.

penev92:
Yes that's what I was suggesting.

pchote:
Code looks simple enough, and confirmed that this prevents AA guns from targeting things they can no longer see. :+1:

pchote:
Adding has dependencies so that this isn't accidentally merged with the testcase commit attached.

obrakmann:
Looks good to me. Tested in conjunction with #10023 and #9874.

You can remove the [test case](../commit/b63a980b0afeab8af548da7393700e0f1c992d46) now.

penev92:
Test case removed.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/52ee7259d729f178f4562cec74584c668326dade)

PR #10008, opened 2015-11-15
Add a null check to CachedTransform.Update()

### LINKED ISSUE #10006: Crash in multiplayer lobby/global chat using game filter
(Playtest 20151114) Win 8.1
To recreate click/go:
* Multiplayer
* Connect to global IRC
* select (single click) one of the servers (so that you can see the map preview)
* open filter and deselect and reselect the "empty" filter
(crash)

Ther might be other ways to replicate this, but this is easy enough. I tested nad crashed 4 times in a row.

exception.log
http://www.megafileupload.com/i4mo/exception.log

* If for some reason you don't crash select another server and do a bit of filter switching. "Works" for me :)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28263375-crash-in-multiplayer-lobby-global-chat-using-game-filter?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

penev92:
Fixes #10006.

### DIFF
--- a/OpenRA.Game/Widgets/WidgetUtils.cs
+++ b/OpenRA.Game/Widgets/WidgetUtils.cs
@@ -273,7 +273,7 @@ namespace OpenRA.Widgets
 
 		public U Update(T input)
 		{
-			if (initialized && ((input == null && lastInput == null) || input.Equals(lastInput)))
+			if (initialized && ((input == null && lastInput == null) || (input != null && input.Equals(lastInput))))
 				return lastOutput;
 
 			lastInput = input;


pchote:
This looks like it should fix the symptom, but not the underlying cause.  Could you also add a commit that addresses my comment in #10006?

penev92:
OK, I think we now came to a conclusion on IRC that this is still the proper fix?

pchote:
We have.  The problem case is the server title, which has special handling of the null case instead of hiding the label.  That means that this equality check is the only problem.

obrakmann:
I was unable to reproduce the crash on this branch. :+1:

pchote:
Fix confirmed, so here's my other :+0.5: to make :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/901b394574eb47030546f5e7e321367822128147

Issue #9242, opened 2015-09-03
DeployToUpgrade multiple revocations.


Phrohdoh:
```
Exception of type `System.InvalidOperationException`: Object <OpenRA.Mods.Common.Traits.DeployToUpgrade> revoked more levels of upgrade deployed than it granted for lpst 251.
  at OpenRA.Mods.Common.Traits.UpgradeManager.RevokeUpgrade (OpenRA.Actor self, System.String upgrade, System.Object source) [0x00060] in /Users/thill/projects/openra/OpenRA.Mods.Common//Traits/Upgrades/U
pgradeManager.cs:164
  at OpenRA.Mods.Common.Traits.DeployToUpgrade.RevokeUpgrades () [0x00027] in /Users/thill/projects/openra/OpenRA.Mods.Common//Traits/Upgrades/DeployToUpgrade.cs:168
  at OpenRA.Mods.Common.Traits.WithSpriteBody+<PlayCustomAnimationBackwards>c__AnonStorey3.<>m__0 () [0x00043] in /Users/thill/projects/openra/OpenRA.Mods.Common//Traits/Render/WithSpriteBody.cs:101
  at OpenRA.Graphics.Animation+<PlayThen>c__AnonStorey0.<>m__0 () [0x0008b] in /Users/thill/projects/openra/OpenRA.Game//Graphics/Animation.cs:122
  at OpenRA.Graphics.Animation.Tick (Int32 t) [0x00037] in /Users/thill/projects/openra/OpenRA.Game//Graphics/Animation.cs:182
  at OpenRA.Graphics.Animation.Tick () [0x0001f] in /Users/thill/projects/openra/OpenRA.Game//Graphics/Animation.cs:168
  at OpenRA.Mods.Common.Traits.RenderSprites.Tick (OpenRA.Actor self) [0x00026] in /Users/thill/projects/openra/OpenRA.Mods.Common//Traits/Render/RenderSprites.cs:178
  at OpenRA.World.<Tick>m__3 (TraitPair`1 x) [0x0000e] in /Users/thill/projects/openra/OpenRA.Game//World.cs:286
  at (wrapper delegate-invoke) System.Action`1<OpenRA.TraitPair`1<OpenRA.Traits.ITick>>:invoke_void_T (OpenRA.TraitPair`1<OpenRA.Traits.ITick>)
[...]

PR #10010, opened 2015-11-15
Fix UpgradeActorsNear revoking upgrades after the actor dies

penev92:
Very hard bug to reproduce: if you can power down a stealth generator just as it is dieing you can hit a race condition that leads to ActorExited() being called to revoke the upgrades after they have already been revoked once.

It helps if you have 100000 actors so the performance is bad enough for the race condition to be more likely to happen. If people want I can upload a couple of replays where I eventually managed to reproduce the crash.

Related to #9242 - exception is the same, with just the source being `UpgradeActorsNear` instead of `DeployToUpgrade`.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeActorsNear.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeActorsNear.cs
@@ -88,7 +88,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		void ActorEntered(Actor a)
 		{
-			if (a.Disposed)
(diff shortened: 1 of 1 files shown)


pchote:
Haven't tested, but this certainly can't hurt so :+1:.

obrakmann:
Makes sense. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/ede68a5d72ad0bcb9b722a9407430d8ca8e5f3e6)

Issue #4617, opened 2014-02-10
Per-cell infantry attack orders


pizzaoverhead:
Currently, if a cell contains one or more enemy infantry units, when attempting to attack, the player can mis-click on some of the empty space in the cell, resulting in a move command instead of an attack command. Higher resolutions with small sprite sizes exacerbate this problem. It is especially problematic for helicopters when attacking rocket soldiers.

This behaviour could be improved if the attack cursor was shown when over any part of a cell containing enemy infantry. It the cell contains more than one infantry unit, the attack order should be aimed at the infantry unit nearest the mouse position.
##

PR #10011, opened 2015-11-15
Better actor selection in game by actor bounds center

### LINKED ISSUE #9998: Husks aren't targetable on refinery pads
Caused by husks having no `Selectable:` `Priority:`. If an actor isn't selectable, it'll default to having the lowest selection priority possible.

Unit targeting code checks for all actors under cursor, and picks the actor with highest priority: [UnitOrderGenerator.cs:L24](https://github.com/OpenRA/OpenRA/blob/206f64494965c08d89db618cdea2f519c355b091/OpenRA.Game/Orders/UnitOrderGenerator.cs#L24)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28256378-husks-aren-t-targetable-on-refinery-pads?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

atlimit8:
Sibling to #9927.
Fixes #9998.

### DIFF
--- a/OpenRA.Game/Orders/UnitOrderGenerator.cs
+++ b/OpenRA.Game/Orders/UnitOrderGenerator.cs
@@ -17,31 +17,31 @@ namespace OpenRA.Orders
 {
 	public class UnitOrderGenerator : IOrderGenerator
 	{
-		static Target TargetForInput(World world, CPos xy, MouseInput mi)
+		static Target TargetForInput(World world, CPos cell, int2 worldPixel, MouseInput mi)
 		{
 			var actor = world.ScreenMap.ActorsAt(mi)
 				.Where(a => !world.FogObscures(a) && a.Info.HasTraitInfo<ITargetableInfo>())
-				.WithHighestSelectionPriority();
+				.WithHighestSelectionPriority(worldPixel);
 
 			if (actor != null)
 				return Target.FromActor(actor);
 
 			var frozen = world.ScreenMap.FrozenActorsAt(world.RenderPlayer, mi)
 				.Where(a => a.Info.HasTraitInfo<ITargetableInfo>() && a.Visible && a.HasRenderables)
-				.WithHighestSelectionPriority();
+				.WithHighestSelectionPriority(worldPixel);
 
 			if (frozen != null)
 				return Target.FromFrozenActor(frozen);
 
-			return Target.FromCell(world, xy);
+			return Target.FromCell(world, cell);
 		}
 
-		public virtual IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
+		public virtual IEnumerable<Order> Order(World world, CPos cell, int2 worldPixel, MouseInput mi)
 		{
-			var target = TargetForInput(world, xy, mi);
-			var actorsAt = world.ActorMap.GetActorsAt(xy).ToList();
+			var target = TargetForInput(world, cell, worldPixel, mi);
+			var actorsAt = world.ActorMap.GetActorsAt(cell).ToList();
 			var orders = world.Selection.Actors
-				.Select(a => OrderForUnit(a, target, actorsAt, xy, mi))
+				.Select(a => OrderForUnit(a, target, actorsAt, cell, mi))
 				.Where(o => o != null)
 				.ToList();
 
@@ -62,18 +62,18 @@ namespace OpenRA.Orders
 		public virtual IEnumerable<IRenderable> Render(WorldRenderer wr, World world) { yield break; }
 		public virtual IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world) { yield break; }
 
-		public virtual string GetCursor(World world, CPos xy, MouseInput mi)
+		public virtual string GetCursor(World world, CPos cell, int2 worldPixel, MouseInput mi)
 		{
 			var useSelect = false;
-			var target = TargetForInput(world, xy, mi);
-			var actorsAt = world.ActorMap.GetActorsAt(xy).ToList();
+			var target = TargetForInput(world, cell, worldPixel, mi);
+			var actorsAt = world.ActorMap.GetActorsAt(cell).ToList();
 
 			if (target.Type == TargetType.Actor && target.Actor.Info.HasTraitInfo<SelectableInfo>() &&
 					(mi.Modifiers.HasModifier(Modifiers.Shift) || !world.Selection.Actors.Any()))
 				useSelect = true;
 
 			var ordersWithCursor = world.Selection.Actors
-				.Select(a => OrderForUnit(a, target, actorsAt, xy, mi))
+				.Select(a => OrderForUnit(a, target, actorsAt, cell, mi))
 				.Where(o => o != null && o.Cursor != null);
 
 			var cursorOrder = ordersWithCursor.MaxByOrDefault(o => o.Order.OrderPriority);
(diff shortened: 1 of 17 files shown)


OpenRA.Game/Orders/UnitOrderGenerator.cs
```diff
@@ -19,18 +19,19 @@ class UnitOrderGenerator : IOrderGenerator
 	{
 		public IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
 		{
+			int2 worldPixel = world.ScreenMap.Viewport.ViewToWorldPx(mi.Location);
```
pchote: `var` here and below please.

atlimit8: Oops, carried over coding convention.

atlimit8:
Changed.

pchote:
This simple fix is a big deal because it fixes both the husk-refinery exploit and hiding missile silos / turrets / etc behind power plants / refs.

On the one hand I think it might be worth merging this on to the release branch because those are lame "strategies" that I can't wait to kill, but on the other I think that we shouldn't because this is a non-trivial gameplay change late in the playtest cycle.

deniz1a:
Weren't those fixed in #8388?

reaperrr:
> Weren't those fixed in #8388?

#8388 pretty much fixed the missile silo behind refinery etc., but strictly speaking that was a work-around and there are still corner-cases (literally, in some cases) where the bounds of actors overlap, and #8388's approach also wouldn't work well in TS due to isometry.

Besides, this would allow us to increase infantry bounds, making them easier to select.

pchote:
@reaperrr: are we able to increase the targetable bounds (using CustomSelectionSize) while still keeping the same smaller selection box for infantry?  If yes, then this effectively solves #4617.

reaperrr:
@pchote: if you mean the visual selection box, then yes (though `CustomSelectionSize` isn't needed, `Selectable.Bounds` would handle that).

If you mean different bounds for selecting and attacking respectively, then no.

OpenRA.Game/SelectableExts.cs
```diff
@@ -43,14 +43,30 @@ public static int SelectionPriority(this Actor a)
...
+
+			return ((long)-pixelDistance << 32) + actor.Info.SelectionPriority();
+		}
+
+		static long CalculateActorSelectionPriority(FrozenActor actor, int2 selectionPixel)
+		{
+			var centerPixel = new int2(actor.Bounds.X, actor.Bounds.Y);
+			var pixelDistance = (centerPixel - selectionPixel).Length;
+
+			return ((long)-pixelDistance << 32) + actor.Info.SelectionPriority();
```
RoosterDragon: Could you factor out a common method for the priority distance bits and pass in the specifics from the Actor/FrozenActor as parameters instead please?

OpenRA.Game/Traits/World/ScreenMap.cs
```diff
@@ -33,6 +33,7 @@ public class ScreenMap : IWorldLoaded
 		readonly Cache<Player, SpatiallyPartitioned<FrozenActor>> partitionedFrozenActors;
 		readonly SpatiallyPartitioned<Actor> partitionedActors;
 		WorldRenderer worldRenderer;
+		public Viewport Viewport { get; private set; }
```
RoosterDragon: Why not just a getter that returns `worldRenderer.Viewport?`

atlimit8: This avoids explicitly creating an instance member variable. However, if you prefer:
```c#
public Viewport Viewport { get { return viewport; } }
Viewport viewport;
```
I could easily change it.

RoosterDragon: I mean: `public Viewport Viewport { get { return worldRenderer.Viewport; } }`

atlimit8:
Changed.

OpenRA.Game/Traits/World/ScreenMap.cs
```diff
@@ -33,6 +33,7 @@ public class ScreenMap : IWorldLoaded
 		readonly Cache<Player, SpatiallyPartitioned<FrozenActor>> partitionedFrozenActors;
 		readonly SpatiallyPartitioned<Actor> partitionedActors;
 		WorldRenderer worldRenderer;
+		public Viewport Viewport { get { return worldRenderer.Viewport; } }
```
pchote: Requiring a WorldRenderer to access the Viewport (along with other render-only data) was a deliberate decision to force people to stop mixing simulation and rendering code is unmaintainable ways.  This makes the viewport accessible to anyone who has a world, so :-1:

OpenRA.Game/Orders/UnitOrderGenerator.cs
```diff
@@ -57,9 +58,10 @@ public IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
 		public string GetCursor(World world, CPos xy, MouseInput mi)
 		{
 			var useSelect = false;
-			var underCursor = world.ScreenMap.ActorsAt(mi)
+			var worldPixel = world.ScreenMap.Viewport.ViewToWorldPx(mi.Location);
```
pchote: It would be more appropriate to change the `GetCursor` signature to pass in the viewport or the worldrenderer.

atlimit8:
Added `int2 worldPixel` to `IOrderGenerator.Order` & `IOrderGenerator.GetCursor`:
```diff
diff --git a/OpenRA.Game/Orders/IOrderGenerator.cs b/OpenRA.Game/Orders/IOrderGenerator.cs
index 5d751ef..6384d4a 100644
--- a/OpenRA.Game/Orders/IOrderGenerator.cs
+++ b/OpenRA.Game/Orders/IOrderGenerator.cs
@@ -15,10 +15,10 @@ namespace OpenRA
 {
 	public interface IOrderGenerator
 	{
-		IEnumerable<Order> Order(World world, CPos xy, MouseInput mi);
+		IEnumerable<Order> Order(World world, CPos cell, int2 worldPixel, MouseInput mi);
 		void Tick(World world);
 		IEnumerable<IRenderable> Render(WorldRenderer wr, World world);
 		IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world);
-		string GetCursor(World world, CPos xy, MouseInput mi);
+		string GetCursor(World world, CPos cell, int2 worldPixel, MouseInput mi);
 	}
 }
```

RoosterDragon:
:+1: Needs a rebase though.

atlimit8:
Rebased.

Mailaender:
:white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/09bda78376fa481a35d222d60eed52660cd5f93f...2c59843814354cf00ebf3c1dc5af9be9b17ffb8e)

Mailaender:
> This simple fix is a big deal because it fixes both the husk-refinery exploit and hiding missile silos / turrets / etc behind power plants / refs.

Well done! :+1:

Micr0Bit:
@atlimit8 : im gonna be that cheeky and say "Thank you for this Fix" in the name of the whole open-redalert-playerbase ...

Issue #10005, opened 2015-11-15
Crash on "Bomber John" when trying to iron curtain minelayer. (current release)


epicelite:
```
Red Alert Mod at Version release-20150919
on map 8bc40b22e045f5561208cdb65cd182dc607efdc6 (Bomber John by Holloweye).
Operating System: Windows (Microsoft Windows NT 6.1.7601 Service Pack 1)
Runtime Version: .NET CLR 4.0.30319.34209
Exception of type `System.InvalidOperationException`: Actor does not have trait of type `OpenRA.Mods.Common.Traits.RenderBuilding`
   at OpenRA.TraitDictionary.TraitContainer`1.Get(UInt32 actor)
   at OpenRA.TraitDictionary.Get[T](Actor actor)
   at OpenRA.Actor.Trait[T]()
   at OpenRA.Mods.Common.Traits.GrantUpgradePower.Activate(Actor self, Order order, SupportPowerManager manager)
   at OpenRA.Mods.Common.Traits.SupportPowerInstance.Activate(Order order)
   at OpenRA.Mods.Common.Traits.SupportPowerManager.ResolveOrder(Actor self, Order order)
   at OpenRA.Network.UnitOrders.ProcessOrder(OrderManager orderManager, World world, Int32 clientId, Order order)
   at OpenRA.Network.OrderManager.Tick()
   at OpenRA.Game.InnerLogicTick(OrderManager orderManager)
   at OpenRA.Game.LogicTick()
   at OpenRA.Game.Loop()
   at OpenRA.Game.Run()
   at OpenRA.Program.Run(String[] args)
   at OpenRA.Program.Main(String[] args)
```

penev92:
Can you give us a replay?

pchote:
This is pretty clear cut, so we don't need a replay.  This map gives a tree actor the Iron Curtain power.  Trees don't use `RenderBuilding`, but `GrantUpgradePower` has a hardcoded `Trait<RenderBuilding>()` without a corresponding `Requires<RenderBuildingInfo>`.

pchote:
On bleed/prep the hardcoded `RenderBuilding` is changed to a hardcoded `WithSpriteBody`, so this will no longer crash.  That doesn't mean this bug is invalid, though: this trait shouldn't hardcode that lookup or at the very least it should declare the correct `Requires<>`.

pchote:
This is a simple fix for something that map/modders have already hit.  I think this qualifies it for a next release fixup.

[...]

PR #10013, opened 2015-11-15
Fix a crash when using the chronoshift power on the map bomber john

abcdefg30:
Fixes the issue discovered by @obrakmann here: https://github.com/OpenRA/OpenRA/issues/10005#issuecomment-156841805.

~~Just a workaround at the moment, as `Teleport` has an [hardcoded sequence](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.RA/Activities/Teleport.cs#L97).~~

### DIFF
--- a/OpenRA.Mods.RA/Activities/Teleport.cs
+++ b/OpenRA.Mods.RA/Activities/Teleport.cs
@@ -93,7 +93,7 @@ namespace OpenRA.Mods.RA.Activities
 			if (teleporter != null && self != teleporter && !teleporter.Disposed)
 			{
 				var building = teleporter.TraitOrDefault<WithSpriteBody>();
-				if (building != null)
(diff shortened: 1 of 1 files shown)


pchote:
How about adding a HasSequence check instead?

abcdefg30:
Changed.

pchote:
These support power animations will eventually be replaced with an `INotifySupportPowerStatus` interface, so we don't need to do anything more complicated than this in the short term.

obrakmann:
Doesn't crash anymore. :+1:

pchote:
Fix confirmed.  Here's my other :+0.5: to make :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/06e819762e2e0055ea6a47ecd0e580d773f12704

Issue #9986, opened 2015-11-14
Map editor tooltip margins are incorrect


pchote:
The margins of the tooltip type introduced by #9925 have at least three problem cases:

1. The top/bottom margins are wrong for all mods.
2. The margin between lines are wrong for all mods.
3. Leading / trailing newlines are not stripped, allowing empty rows.

![screen shot 2015-11-11 at 23 11 08](https://cloud.githubusercontent.com/assets/167819/11105835/957d186e-88c9-11e5-8385-9ebae6c9cb8b.png)
![screen shot 2015-11-11 at 23 10 43](https://cloud.githubusercontent.com/assets/167819/11105839/974c0830-88c9-11e5-8dc1-5c2150ababf8.png)
![screen shot 2015-11-11 at 23 14 36](https://cloud.githubusercontent.com/assets/167819/11105945/31f7fb1e-88ca-11e5-9557-63bf221dcb1d.png)
![screen shot 2015-11-11 at 23 13 25](https://cloud.githubusercontent.com/assets/167819/11105951/3981ba1e-88ca-11e5-8602-e9fbadbaf66b.png)

pchote:
Note: this doesn't need to block todays playtest, but *should* be fixed before the final release.

PR #10015, opened 2015-11-16
Add map editor only tooltips

atlimit8:
Polishes map editor with tooltips for actors that are not to show tooltips in game that should help those using the map editor (they don't have to figure out that the player starting points are `mpspawn`).
This is not meant as a partial fix for #9986.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Tooltip.cs
+++ b/OpenRA.Mods.Common/Traits/Tooltip.cs
@@ -12,12 +12,23 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
-	[Desc("Shown in the build palette widget.")]
-	public class TooltipInfo : ITraitInfo, ITooltipInfo
+	public abstract class TooltipInfoBase : ITraitInfo
 	{
 		[Translate] public readonly string Description = "";
 		[Translate] public readonly string Name = "";
 
+		public abstract object Create(ActorInitializer init);
+	}
+
+	[Desc("Shown in map editor.")]
+	public class EditorOnlyTooltipInfo : TooltipInfoBase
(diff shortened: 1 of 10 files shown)


pchote:
Nice idea!  I wonder if this might fit as an `EditorDescription` field on Tooltip instead of needing its own trait?

atlimit8:
My concern with adding something like `EditorDescription` to `TooltipInfo` is that there would need to be checks everywhere `TooltipInfo` is used to show a tooltip.

pchote:
Why is that?  The logic for building the editor tooltip is separate from the world tooltips, so there should be no overlap afaik.

atlimit8:
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs#L132
`.Where(a => !world.FogObscures(a) && a.Info.HasTraitInfo<ITooltipInfo>())`
to something like
`.Where(a => !world.FogObscures(a) && a.Info.HasTraitInfo<ITooltipInfo>() && !string.IsNullOrEmpty(a.Info.TraitInfo<ITooltipInfo>().TooltipForPlayerStance(a.Owner.Stances[world.RenderPlayer])))`

Some actors should not have tooltips in game.

I have not found any other places in my quick search.

pchote:
Ah, I see.

atlimit8:
Rebased.

abcdefg30:

pchote:
Code and behaviour look good to me too. :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/8a51e3828fce7153ddbc518212500b460003d5cc

PR #10017, opened 2015-11-16
Fix AA-only missiles not damaging landed aircraft

### LINKED ISSUE #9911: SAM missiles do not damage landed chinooks
Noticed this on gdi07 in the cnc mod

http://expirebox.com/download/828a6e39674ed88c2a44e4cd8e26af6b.html

On bleed (git-e609abf)

reaperrr:
This adresses the edge case where an AA-only missile is fired while the aircraft is in the air, but hits after the aircraft landed and then deals no damage, because the target type is no longer `Air`.

Fixes #9911.

### DIFF
--- a/mods/cnc/rules/defaults.yaml
+++ b/mods/cnc/rules/defaults.yaml
@@ -121,11 +121,11 @@
 	AppearsOnRadar:
 		UseLocation: yes
 	Targetable@GROUND:
-		TargetTypes: Ground, Vehicle
+		TargetTypes: Ground, Vehicle, Aircraft
 		UpgradeTypes: airborne
 		UpgradeMaxEnabledLevel: 0
 	Targetable@AIRBORNE:
-		TargetTypes: Air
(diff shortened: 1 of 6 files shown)


pchote:
This may open up a new edge case where splash damage leaks to aircraft on the ground below an explosion.

GraionDilach:
@reaperrr Why you didn't went with the way I said in the issue?

reaperrr:
@GraionDilach Because then AA missiles hitting a landed helicopter would deal splash damage vs. nearby actors. AA warheads aren't balanced for that, they often have 100% efficiency vs. none or wood, for example.

GraionDilach:
Then duplicate the warhead to damage Ground with say 1/5 damage as a start. I'd expect that splash, although I can see why you say it'd be imbalanced with the previous values.

reaperrr:
I don't have the time and "head-bandwidth" for this at the moment, so I'm concentrating on other PRs I consider more important.

PR #10018, opened 2015-11-16
Restore targeting cursor when mousing over a disguised enemy spy.

### LINKED ISSUE #10012: Preserve the original behaviour of spies (don't force players to force fire them)
Spies on `bleed` have been "fixed". So, in order to detect them you have to use either dogs (only Soviets) or try to select them and if that doesn't work, you're sure it's a spy, which is super confusing (and then you have to force-fire them). 

I'm strongly against that new behaviour. Firstly, nobody complained about spies being detectable with cursors. Usually there's a lot of infantry on the map, so it takes a while to scan all of them with your cursor, which already is punishing (but totally fine). Secondly, it's already pretty hard to detect spies. They can disguise as anyone, so there are no common patterns like "oh, I see a rifler, I know spies can only be riflers, so it's easy for me to select all riflers with the "W" key and check if it's my unit or a spy". Thirdly, new players might not know about the ability to force fire.

Please, reconsider this behaviour. Ideally, I would like them not to be "fixed".

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28268695-preserve-the-original-behaviour-of-spies-don-t-force-players-to-force-fire-them?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #10012.

This unfortunately requires a hack. A much cleaner fix (which I tried originally) would have been to make the manual targeting check against the original owner, ignoring EffectiveOwner.  Unfortunately the EffectiveOwner check is enforced for both manual and automatic orders on the ResolveOrder side, so these manual orders are ignored.  The only way to make this work without rewriting the attack logic again is to force a force attack.

On the plus side, at least this approach is easy to review and doesn't risk regressions...

### DIFF
--- a/OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
@@ -279,7 +279,7 @@ namespace OpenRA.Mods.Common.Traits
 			public int OrderPriority { get; private set; }
 			public bool OverrideSelection { get { return true; } }
 
-			bool CanTargetActor(Actor self, Target target, TargetModifiers modifiers, ref string cursor)
+			bool CanTargetActor(Actor self, Target target, ref TargetModifiers modifiers, ref string cursor)
 			{
 				IsQueued = modifiers.HasModifier(TargetModifiers.ForceQueue);
 
@@ -326,13 +326,13 @@ namespace OpenRA.Mods.Common.Traits
 				return true;
 			}
 
-			public bool CanTarget(Actor self, Target target, List<Actor> othersAtTarget, TargetModifiers modifiers, ref string cursor)
+			public bool CanTarget(Actor self, Target target, List<Actor> othersAtTarget, ref TargetModifiers modifiers, ref string cursor)
 			{
 				switch (target.Type)
 				{
 					case TargetType.Actor:
 					case TargetType.FrozenActor:
-						return CanTargetActor(self, target, modifiers, ref cursor);
+						return CanTargetActor(self, target, ref modifiers, ref cursor);
 					case TargetType.Terrain:
 						return CanTargetLocation(self, self.World.Map.CellContaining(target.CenterPosition), othersAtTarget, modifiers, ref cursor);
 					default:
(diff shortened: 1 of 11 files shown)


pchote:
Seems to work well except for those two queueing issues and the comment nit.

penev92:
#HackHackHack

On a more serious note, this is ugly as *. I am personally very much agains this, but seeing as the most likely alternative is blocking the next playtest (and ultimately the release)... :+1:
At least it shouldn't be breaking anything (yet).

OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
```diff
@@ -279,13 +279,21 @@ public AttackOrderTargeter(AttackBase ab, int priority, bool negativeDamage)
...
 				IsQueued = modifiers.HasModifier(TargetModifiers.ForceQueue);
 
 				if (modifiers.HasModifier(TargetModifiers.ForceMove))
 					return false;
 
+				// Disguised actors are revealed by the attack cursor
+				// HACK: works around limitations in the targeting code that force the
+				// targeting and attacking logic (which should be logically separate)
+				// to use the same code
+				if (target.Actor.EffectiveOwner != null && target.Actor.EffectiveOwner.Disguised &&
```
RoosterDragon: Isn't this (`target.Actor`) a null-reference risk if the target is a frozen actor?

pchote:
Fixed.

obrakmann:
Restores the behaviour and doesn't look like it breaks anything else. :+1: despite the hack.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/3a74f1e86c5eb2bc7aa20c6b615379bcd6779b78)

PR #10019, opened 2015-11-16
Add a couple community maps to the pool.

ghost:
Branched from playtest-20151114

I wanted to add a few competitive 1v1 / 2v2 maps from authors in the community that currently don't have any content in the game yet.

Tabula Rasa - FiveAces (I removed two cacti not sitting quite right on the terrain)
http://resource.openra.net/maps/5470/

Sidestep - SoScared
http://resource.openra.net/maps/6421/

I also wanted to add Siberian Road, but kyrylo is still working on a new version which needs testing
http://resource.openra.net/maps/6441/

Feel free to suggest others.


pchote:
You'll need to make sure that your version of Tabula Rasa is available on the resource center for our http://www.openra.net/games/ list to work properly.

r34ch:
@pchote if it is just two minor cosmetic changes, would it be better to simply use the original?

pchote:
Up to you, really.

r34ch:
Updated with FiveAces' original non edited map so that he keeps the credit on the resource centre.

pchote:
Both look good to me :+1:

obrakmann:
Are the pillboxes on Sidestep supposed to shoot at the player's units? Because with the current rules they wouldn't, which makes them kind of useless unless they're just meant to be captured.

The barrel placement on Tabula Rasa could be improved a bit, but I guess we can live with that.

r34ch:
@obrakmann the way I've seen them used is to be captured along with the derrick as defence.

obrakmann:
well, :+1: then.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/9130b9581a79c7cfc40867c38fc8b7caac94f96d)

PR #10021, opened 2015-11-17
added nawast

teees:
Added nawast, made harvester target only proc and weeder target only nawast.

EDIT:
- [x] fixed bounds
- [x] Made use of ntwast_b
- [x] use last 40 frames of ntwast_a correctly

### DIFF
--- a/mods/ts/rules/nod-vehicles.yaml
+++ b/mods/ts/rules/nod-vehicles.yaml
@@ -164,7 +164,7 @@ REPAIR:
 		Voice: Attack
 
 WEED:
-	Inherits: ^VoxelVehicle
+	Inherits: ^VoxelTank
 	Valued:
 		Cost: 1400
 	Tooltip:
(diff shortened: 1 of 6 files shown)


mods/ts/rules/nod-vehicles.yaml
```diff
@@ -183,6 +193,8 @@ WEED:
 		Type: Heavy
 	RevealsShroud:
 		Range: 4c0
+	-WithVoxelBody
```
reaperrr: Missing `:`

reaperrr:
To fix the Travis error, you have to add `ProvidesPrerequisite@buildingname:` to `NAMISL`.

mods/ts/rules/nod-vehicles.yaml
teees: thanks, fixed

teees:
@reaperrr thanks, added

pchote:
> use last 40 frames of ntwast_a correctly

This should be as simple as removing the `ShadowStart` lines and then setting the proper `Length`.

pchote:
:+1: after fixing the animation issue.

mods/ts/rules/world.yaml
```diff
@@ -54,6 +54,7 @@
 		PipColor: Red
 		AllowedTerrainTypes: Clear
 		AllowUnderActors: true
+		AllowUnderBuildings: true
```
pchote: Are you sure about this?

teees: In the original veins grew under buildings and destroyed them quite rapidly, so I guess we will probably want this as soon as veins are functional

pchote: Oh right, I forgot about that.

pchote: although note that this will also let them grow under walls, which we don't want.  It might be better to leave this disabled for now, and then explicitly enable it once we can support the behaviour and building damage properly.

teees: ok, will change it :)

teees:
Actually the current animation should be fine, 20-39 is definitely for damaged, after that should be shadow

Mailaender:

Mailaender:
Thanks! 

We could also think of splashing the veins on the grounds again upon destruction like we do with Tiberium refinery and harvesters.

pchote:
Veins come from a living creature, so having them spread on the ground doesn't really fit.

PR #10023, opened 2015-11-17
Fix gps dot

penev92:
More hacks to fix GPS bogosity.
Should no longer enable auto targeting of actors under shroud.

### DIFF
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -189,7 +189,12 @@ namespace OpenRA.Traits
 	public interface IDefaultVisibilityInfo : ITraitInfo { }
 	public interface IDefaultVisibility { bool IsVisible(Actor self, Player byPlayer); }
 	public interface IVisibilityModifier { bool IsVisible(Actor self, Player byPlayer); }
-	public interface IFogVisibilityModifier { bool HasFogVisibility(Player byPlayer); }
+
+	public interface IFogVisibilityModifier
+	{
+		bool IsVisible(Actor actor);
+		bool HasFogVisibility(Player byPlayer);
+	}
 
 	public interface IRadarColorModifier { Color RadarColorOverride(Actor self); }
 
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -193,7 +193,7 @@ namespace OpenRA.Traits
 	public interface IFogVisibilityModifier
 	{
 		bool IsVisible(Actor actor);
-		bool HasFogVisibility(Player byPlayer);
+		bool HasFogVisibility();
 	}
 
 	public interface IRadarColorModifier { Color RadarColorOverride(Actor self); }
--- a/OpenRA.Mods.RA/Effects/GpsDot.cs
+++ b/OpenRA.Mods.RA/Effects/GpsDot.cs
@@ -10,6 +10,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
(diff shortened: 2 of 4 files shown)


OpenRA.Mods.RA/Effects/GpsDot.cs
```diff
@@ -92,17 +100,21 @@ public void Tick(World world)
...
-			show = false;
-			if (!self.IsInWorld || self.IsDead || self.World.RenderPlayer == null)
+			if (!self.IsInWorld || self.IsDead)
 				return;
 
-			var gps = watcher[self.World.RenderPlayer];
-			show = (gps.Granted || gps.GrantedAllies) && ShouldShowIndicator();
+			foreach (var player in self.World.Players)
+			{
+				showToPlayer[player] = false;
```
pchote: You set this again unconditionally below, so this line is redundant.

pchote:
This fixes a bug that makes gap generators basically useless in the late game, so this should be merged to prep.

penev92:
Updated. I also figured the test commit from #10007 can also help out with the testing of this one.

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -189,7 +189,12 @@ public interface IRadarSignature
 	public interface IDefaultVisibilityInfo : ITraitInfo { }
 	public interface IDefaultVisibility { bool IsVisible(Actor self, Player byPlayer); }
 	public interface IVisibilityModifier { bool IsVisible(Actor self, Player byPlayer); }
-	public interface IFogVisibilityModifier { bool HasFogVisibility(Player byPlayer); }
+
+	public interface IFogVisibilityModifier
+	{
+		bool IsVisible(Actor actor);
```
pchote: This may want to include `Player byPlayer` for consistency.

pchote: Or maybe we should remove it from `HasFogVisibility` instead.

pchote:
I tested this in a multiplayer game, and AA guns were still able to attack migs that were hidden under a gap generator :(

penev92:
Updated.
FWIW this PR seems to be doing its job well enough. And that is disabling auto targeting under shroud (gap generator) when there is a GPS available.
I believe you are refering to #9999, which is fixed by #10007 and is totally different - it's about firing at an already acquired target that moves into shroud.
Basically both of these PRs each fix half of the problem. For best results you need them both, but for the moment you can test each separately.

pchote:
Code changes look good to me, and confirmed on a branch that also included #10007 that AA guns vs aircraft (finally!) work as expected both with and without GPS. :+1:

pchote:
Adding has dependencies so that this isn't accidentally merged with the testcase commit attached.

mods/ra/rules/aircraft.yaml
```diff
@@ -88,7 +88,7 @@ MIG:
 		Name: Mig Attack Plane
 		Description: Fast Ground-Attack Plane.\n  Strong vs Buildings, Tanks\n  Weak vs Infantry, Light armor, Aircraft
 	Health:
-		HP: 70
+		HP: 700
```
obrakmann: This seems rather unrelated.

obrakmann:
Looks good to me, :+1: . Tested in conjunction with #10007 and #9874.

You can remove the [test case](../commit/ff0cdfb49baab882deb69f9d9047628bf6306121) now.

penev92:
Test case removed.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/53e691963547224977d85559fe5f969770454c88)

PR #10025, opened 2015-11-18
D2k, upgrades without conyard

OmegaBolt:
This enables upgrades to be purchased without the presence of a Con Yard, as is possible in vanilla.

There was some discussion on the IRC that apparently this was tried earlier with issues, although I'm not sure if the whole thing was actually PR'd separately from #9106. I'd like to get this in as it is fairly annoying that you can't upgrade without a Con Yard, even if you have the upgradable building, but this should be properly tested to see if any issues do exist (no one apparently remembers what the issues were...). From my testing I don't see anything immediately noticable.

Also check the missions to make sure all the available tech is correct, I tried updating them to keep it the same.

The reason the player is also given the building queue is because if it isn't the case then the upgrade tab gets shown to you when deploying your first MCV. This is because the queue appears to activate once the buildup is finished, while the prerequisite is activated immediately. This simply means all buildings should have the construction_yard as a prerequisite, which was already made the case. Just a note for anyone doing any mapping.

### DIFF
--- a/mods/d2k/maps/atreides-02a/map.yaml
+++ b/mods/d2k/maps/atreides-02a/map.yaml
@@ -205,9 +205,9 @@ Rules:
 	carryall.reinforce:
 		Cargo:
 			MaxWeight: 10
-	construction_yard:
-		Production:
-			Produces: Building
+	upgrade.conyard:
+		Buildable:
+			Prerequisites: ~disabled
 	concreteb:
 		Buildable:
 			Prerequisites: ~disabled
--- a/mods/d2k/rules/player.yaml
+++ b/mods/d2k/rules/player.yaml
@@ -1,6 +1,8 @@
 Player:
 	AlwaysVisible:
 	TechTree:
+	Production:
+		Produces: Building, Upgrade
 	ClassicProductionQueue@Building:
 		Type: Building
 		BuildSpeed: 1.0
@@ -16,7 +18,6 @@ Player:
 		QueuedAudio: Upgrading
 		ReadyAudio: NewOptions
 		BlockedAudio: NoRoom
-		SpeedUp: true
 	ClassicProductionQueue@Infantry:
 		Type: Infantry
 		BuildSpeed: 1.0
--- a/mods/d2k/maps/atreides-01a/map.yaml
+++ b/mods/d2k/maps/atreides-01a/map.yaml
@@ -145,9 +145,9 @@ Rules:
 		WormManager:
 			Minimum: 1
 			Maximum: 1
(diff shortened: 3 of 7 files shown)


mods/d2k/rules/player.yaml
```diff
@@ -1,6 +1,8 @@
 Player:
 	AlwaysVisible:
 	TechTree:
+	Production:
+		Produces: Building, Upgrade
```
abcdefg30: Why did you add `Building` here?
It seems to work fine (even better) without it.

mods/d2k/maps/atreides-02a/map.yaml
```diff
@@ -205,9 +205,9 @@ Rules:
 	carryall.reinforce:
 		Cargo:
 			MaxWeight: 10
-	construction_yard:
-		Production:
-			Produces: Building
+	upgrade.conyard:
+		Buildable:
+			Prerequisites: ~disabled
```
abcdefg30: This won't give us the desired results, as you can still buy the barracks upgrade which has no effect.
You probably want to use
```
Player:
	-Production:
```
instead (on the other missions except 03a too).

mods/d2k/rules/player.yaml
OmegaBolt: Read the initial description.

pchote:
> The reason the player is also given the building queue is because if it isn't the case then the upgrade tab gets shown to you when deploying your first MCV. 

This would be better fixed by adding a flag to the upgrade queue or MCV so that it doesn't get automatically selected.

OmegaBolt:
Not sure exactly what you mean (adding a new flag or something else), but the deeper issue appears to be that production is only granted once buildup is complete whereas prerequisites are activated immediately. This is why, if you give only the upgrade queue to the player, it is selected first because the upgrades prerequisites are granted immediately thereby filling up the upgrade tab.

mods/d2k/rules/player.yaml
abcdefg30: Ops, sorry forgot about that.

pchote:
The fundamental problem is that the upgrade queue is automatically selected by the game, which we don't want to happen because the upgrade queue is special in that it exists to the side of the regular production / tech / build palette setup.

I'm suggesting that we should just *tell* the game this instead of relying on a fragile (its not an expected use, so it might break in the future) workaround to trick it into doing what we want.

pchote:
Unfortunately this seems to have died in its tracks, so closing to clean up the PR queue.

PR #10026, opened 2015-11-18
Line spacing support in font engine & multi-line tooltip fix-ups

atlimit8:
Adds supports for line spacing to the font engine and fixes vertical spacing as requested in #9986, but does not remove empty lines.

Use the `label`'s `Y` property (& the tooltip's `Height` property) to adjust the top padding, `label`'s `LineSpacing` property to adjust the line spacing, and the tooltip's `Height` property to adjust the bottom padding.

### DIFF
--- a/OpenRA.Game/Graphics/SpriteFont.cs
+++ b/OpenRA.Game/Graphics/SpriteFont.cs
@@ -21,7 +21,7 @@ namespace OpenRA.Graphics
 	{
 		static readonly Library Library = new Library();
 
-		readonly int size;
+		public readonly int Size;
 		readonly SheetBuilder builder;
 		readonly Func<string, float> lineWidth;
 		readonly Face face;
@@ -32,7 +32,7 @@ namespace OpenRA.Graphics
 			if (builder.Type != SheetType.BGRA)
 				throw new ArgumentException("The sheet builder must create BGRA sheets.", "builder");
 
-			this.size = size;
+			this.Size = size;
 			this.builder = builder;
 
 			face = new Face(Library, name);
@@ -50,7 +50,7 @@ namespace OpenRA.Graphics
 
 		void PrecacheColor(Color c, string name)
 		{
-			using (new PerfTimer("PrecacheColor {0} {1}px {2}".F(name, size, c.Name)))
+			using (new PerfTimer("PrecacheColor {0} {1}px {2}".F(name, Size, c.Name)))
 				for (var n = (char)0x20; n < (char)0x7f; n++)
 					if (glyphs[Pair.New(n, c)] == null)
 						throw new InvalidOperationException();
@@ -58,14 +58,14 @@ namespace OpenRA.Graphics
 
 		public void DrawText(string text, float2 location, Color c, int lineSpacing = 0)
 		{
-			location.Y += size;	// baseline vs top
+			location.Y += Size;	// baseline vs top
 
 			var p = location;
 			foreach (var s in text)
 			{
 				if (s == '\n')
 				{
-					location.Y += size + lineSpacing;
+					location.Y += Size + lineSpacing;
 					p = location;
 					continue;
 				}
@@ -95,10 +95,10 @@ namespace OpenRA.Graphics
 		public int2 Measure(string text, int lineSpacing = 0)
 		{
 			if (string.IsNullOrEmpty(text))
-				return new int2(0, size);
+				return new int2(0, Size);
 
 			var lines = text.Split('\n');
-			return new int2((int)Math.Ceiling(lines.Max(lineWidth)), lines.Length * (size + lineSpacing) - lineSpacing);
+			return new int2((int)Math.Ceiling(lines.Max(lineWidth)), lines.Length * (Size + lineSpacing) - lineSpacing);
 		}
 
 		GlyphInfo CreateGlyph(Pair<char, Color> c)
--- a/OpenRA.Game/Graphics/SpriteFont.cs
+++ b/OpenRA.Game/Graphics/SpriteFont.cs
@@ -56,7 +56,7 @@ namespace OpenRA.Graphics
 						throw new InvalidOperationException();
 		}
 
-		public void DrawText(string text, float2 location, Color c)
+		public void DrawText(string text, float2 location, Color c, int lineSpacing = 0)
 		{
 			location.Y += size;	// baseline vs top
 
@@ -65,7 +65,7 @@ namespace OpenRA.Graphics
 			{
 				if (s == '\n')
 				{
-					location.Y += size;
+					location.Y += size + lineSpacing;
 					p = location;
 					continue;
 				}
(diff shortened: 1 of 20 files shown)


OpenRA.Game/Graphics/SpriteFont.cs
```diff
@@ -49,22 +49,22 @@ public SpriteFont(string name, int size, SheetBuilder builder)
...
-			location.Y += size;	// baseline vs top
+			location.Y += Size;	// baseline vs top
 
 			var p = location;
 			foreach (var s in text)
 			{
 				if (s == '\n')
 				{
-					location.Y += size;
+					location.Y += Size + lineSpacing;
```
pchote: It might be better to have the linespacing either inclusive of the height, or as a multiple of the default line spacing (like text editors etc do).

pchote:
Unfortunately this regresses most of the tooltip types in TD (and probably the other mods too):

Lobby factions: bottom padding
screen shot 2015-11-19 at 18 22 22

Support powers: bottom padding
screen shot 2015-11-19 at 18 23 37

Production: description line spacing and bottom padding
screen shot 2015-11-19 at 18 23 09

screen shot 2015-11-19 at 18 23 18

The line spacing change is actually a feature, but we should dial it down a bit for these smaller fonts.

atlimit8:
Added line-wise vertical alignment for adding line spacing to line height, proportional line spacing that defaults 2/5ths font height, and patched ProductionTooltipLogic sizing logic.

Probably could use fresh eyes for checking spacing.

atlimit8:
I don't want to look at the border spacing myself for a while. Use the label's `Y` property (& the tooltip's `Height` property) to adjust the top spacing, label's `LineSpacing` property to adjust the inter-line spacing, and the tooltip's `Height` property to adjust the bottom spacing.

atlimit8:
No one seemed to want to check spacing with "PR: Help wanted" attached.

atlimit8:
Rebased.

Mailaender:
Looks like the padding issues have been resolved. :white_check_mark:

OpenRA.Mods.Common/Widgets/LabelWidget.cs
```diff
@@ -28,15 +29,40 @@ public class LabelWidget : Widget
 		public bool Contrast = ChromeMetrics.Get<bool>("TextContrast");
 		public Color ContrastColor = ChromeMetrics.Get<Color>("TextContrastColor");
 		public bool WordWrap = false;
+		public int LineSpacing = 2;
+
+		[Desc("line spacing = LineSpacing * font size / LineSpacingDivisor or LineSpacing if this is 0.")]
```
pchote: What is the benefit of this complexity vs having a single `[Desc("Line spacing as a multiple of the line height")]float LineSpacing = 1.2; // or whatever`?

atlimit8: I could replace these two with:
```c#
[Desc("Space between lines as a portion of line height or fixed pixel amount (rounded down).")]
public float LineSpacing = 0.2; // or 1.2 if preferred

[Desc("Whether line spacing is 'Proportional' to line height or 'Fixed' pixel amount.")]
public LineSpacingType LineSpacingType = LineSpacingType.Proportional;
```
The purpose is to keep an option for fixed pixel spacing

atlimit8:
Switched to having `Percentage` and `Fixed` `LineSpacingType`s for `LineSpacing`.

OpenRA.Mods.Common/Widgets/LabelWidget.cs
```diff
@@ -52,20 +81,49 @@ protected LabelWidget(LabelWidget other)
...
+		{
+			return WidgetUtils.WrapText(text, Bounds.Width, font.Value);
+		}
+
+		public string TruncateText(string text)
+		{
+			return WidgetUtils.TruncateText(text, Bounds.Width, font.Value);
+		}
+
+		public int2 AutoSize(string text)
```
pchote: This method name isn't very descriptive.  Please change it to something like `SizeToText`.

OpenRA.Mods.Common/Widgets/LabelWidget.cs
```diff
@@ -28,15 +30,42 @@ public class LabelWidget : Widget
 		public bool Contrast = ChromeMetrics.Get<bool>("TextContrast");
 		public Color ContrastColor = ChromeMetrics.Get<Color>("TextContrastColor");
 		public bool WordWrap = false;
+
+		[Desc("Space between lines as a percentage (default) of line height or fixed pixel amount (rounded down).")]
+		public int LineSpacing = 40;
```
pchote: Convention for line spacing (see e.g. CSS, Latex, MS Word, etc) is to specify the *full* line height, so please change this to `public float LineSpacing = 1.4f;`. This remains compatible with `LineSpacingType.Fixed`.

pchote:
Bottom padding of the faction tooltip is wrong if there is no faction description:
screen shot 2016-01-01 at 13 32 27

The size of the button tooltips is inconsistent with other single line tooltips:
screen shot 2016-01-01 at 13 39 48

OpenRA.Mods.Common/Widgets/LabelWidget.cs
atlimit8: Would `public int LineSpacing = 140;` be OK since percentages are used elsewhere instead of floats?

pchote: This is not synced code, so floats aren't a problem.

pchote: an integer percentage is probably okay, if you really want that.

Phrohdoh: My opinion is that we should use integer percentage everywhere possible for consistency.

atlimit8:
Removed leftover extra vertical padding addition from refactor and added support to FactionTooltipLogic.

pchote:
Ok, it feels like we're converging towards a nice fix here.

More issues from testing (its kind of annoying that i'm the only person looking for these...):
* The vertical padding on the single line tooltips in RA/D2K/TS is off – the text sits too high relative to the background.
* The padding around the faction tooltip in the lobby is still off – the "Any" faction tooltip should match the other single line tooltips.
* The padding at the bottom of the extra info (e.g. sell refund value) in the world actor tooltip is too large in all mods.

OpenRA.Mods.Common/Widgets/LabelWidget.cs
```diff
@@ -28,15 +30,53 @@ public class LabelWidget : Widget
 		public bool Contrast = ChromeMetrics.Get<bool>("TextContrast");
 		public Color ContrastColor = ChromeMetrics.Get<Color>("TextContrastColor");
 		public bool WordWrap = false;
+
+		[Desc("Space between lines as a percentage (default) of line height or fixed pixel amount.")]
```
pchote: We don't generate docs for widgets, so these don't do anything.

OpenRA.Mods.Common/Widgets/LabelWidget.cs
```diff
@@ -28,15 +30,53 @@ public class LabelWidget : Widget
...
+		public int LineSpacing = 140;
+
+		[Desc("Percentage: line height = LineSpacing% * font size.",
+			"FixedMargin: line height = font size + LineSpacing.",
+			"FixedHeight: line height = LineSpacing.")]
+		public LineSpacingType LineSpacingType = LineSpacingType.Percentage;
+		public LineVAlign LineVAlign = LineVAlign.Middle;
```
pchote: I'm not convinced we need this.  Top, Middle, Bottom can all be achieved by manipulating the widget position and VAlign, and Collapsed can be achieved by setting the LineSpacing to 100.

OpenRA.Mods.Common/Widgets/LabelWidget.cs
atlimit8: Should they be left for when that does happen, converted to comments, or removed?

pchote: Good question, I don't have a strong opinion on this.  It would be nice to document all of the widgets like this.

Phrohdoh: Keep them as desc imo.

OpenRA.Mods.Common/Widgets/LabelWidget.cs
atlimit8: Both LibreOffice Writer and Calligra Words put the text at the top of the line (don't move first line text based on line spacing).

pchote:
I think this is the best long-term fix, but we can't afford to block a release any longer on it so I've put #10913 together.  Moving this to the Future milestone for when someone has the time to resurrect it.

PR #10027, opened 2015-11-18
Remove editor actor selection tooltip blank line

atlimit8:
Removes the blank line for actors lacking tooltips as mentioned in #9986. With #10026, this provides a solution to #9986.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Editor/ActorSelectorLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Editor/ActorSelectorLogic.cs
@@ -137,7 +137,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 					item.IsVisible = () => true;
 
 					var tooltip = actor.TraitInfoOrDefault<TooltipInfo>();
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Widgets/Logic/Editor/ActorSelectorLogic.cs
```diff
@@ -137,7 +137,7 @@ void IntializeActorPreviews()
 					item.IsVisible = () => true;
 
 					var tooltip = actor.TraitInfoOrDefault<TooltipInfo>();
-					item.GetTooltipText = () => (tooltip == null ? "\nType: " : tooltip.Name + "\nType: ") + actor.Name;
+					item.GetTooltipText = () => (tooltip == null ? " < " + actor.Name + " > " : tooltip.Name) + "\nType: " + actor.Name;
```
pchote: `actor.Name` is included in the next line, so I think it would be better to just omit this first line completely if the tooltip is null.

atlimit8:
Removes instead now.

pchote:
Looks good to me now :+1:

abcdefg30:
:+1:

Issue #8593, opened 2015-06-28
Sound engine should support streaming


reaperrr:
Simply uncompressing AUD/IMA ADPCM into memory before playing them is not exactly efficient. In case of IMA 22khz/stereo (let alone 44khz/st) it can even cause a noticable lag spike when a large music file is loaded.

RoosterDragon:
Now music is no longer cached (#13461) - we pay the loading penalty every time, and this would assist in making it less noticeable.

aydjay:
It looks like there is a project which allows streaming of various file formats: [https://github.com/naudio/NAudio](url)

Unfortunately it is for windows only but they do support ADPCM.

Maybe it is possible to see what they have done and see if it is possible to implement a solution which would be suitable for OpenRA.

reaperrr:
Since we have a PR for it now, might as well move to Next release milestone.

Mailaender:
You can collect your bounty at https://www.bountysource.com/issues/23256122-sound-engine-should-support-streaming

PR #10032, opened 2015-11-18
Added VocLoader

teees:
Needed for .voc files in dune 2
As soon as Sound file loaders are dynamically loaded, this can be moved to d2 if not needed anywhere else

EDIT:
test case for d2k mod: https://gist.github.com/teees/9688e93ff300dc82d612

### DIFF
--- a/OpenRA.Game/FileFormats/VocLoader.cs
+++ b/OpenRA.Game/FileFormats/VocLoader.cs
@@ -1,0 +1,372 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace OpenRA.FileFormats
+{
+	public class VocLoader : ISoundLoader
+	{
+		bool ISoundLoader.TryParseSound(Stream stream, string fileName, out byte[] rawData, out int channels, out int sampleBits, out int sampleRate)
+		{
+			try
+			{
+				var vocStream = new VocStream(stream);
+				channels = vocStream.Channels;
+				sampleBits = vocStream.BitsPerSample;
+				sampleRate = vocStream.SampleRate;
+				rawData = vocStream.ReadAllBytes();
+			}
+			catch
+			{
+				rawData = null;
+				channels = sampleBits = sampleRate = 0;
+				return false;
+			}
+
+			return true;
+		}
+	}
+
+	public class VocStream : Stream
+	{
+		public int BitsPerSample { get { return 8; } }
+		public int Channels { get { return 1; } }
+		public int SampleRate { get; private set; }
+
+		int totalSamples = 0;
+		int samplePosition = 0;
+
+		Stream stream;
+		List<VocBlock> blocks = new List<VocBlock>();
+		IEnumerator<VocBlock> currentBlock;
+		int samplesLeftInBlock = 0;
+		byte[] buffer = new byte[4096];
+
+		struct VocFileHeader
+		{
+			public string Description;
+			public int DatablockOffset;
+			public int Version;
+			public int ID;
+
+			public static VocFileHeader Read(Stream s)
+			{
+				VocFileHeader vfh;
+				vfh.Description = s.ReadASCII(20);
+				vfh.DatablockOffset = s.ReadUInt16();
+				vfh.Version = s.ReadUInt16();
+				vfh.ID = s.ReadUInt16();
+				return vfh;
+			}
+		}
+
+		struct VocBlock
+		{
+			public int Code;
+			public int Length;
+			public VocSampleBlock SampleBlock;
+			public VocLoopBlock LoopBlock;
+		}
+
+		struct VocSampleBlock
+		{
+			public int Rate;
+			public int Samples;
+			public long Offset;
+		}
+
+		struct VocLoopBlock
+		{
+			public int Count;
+		}
+
+		public VocStream(Stream stream)
+		{
+			this.stream = stream;
+			CheckVocHeader();
+			Preload();
+		}
+
+		void CheckVocHeader()
+		{
+			var vfh = VocFileHeader.Read(stream);
+
+			if (!vfh.Description.StartsWith("Creative Voice File"))
+				throw new InvalidDataException("Voc header description not recognized");
+			if (vfh.DatablockOffset != 26)
+				throw new InvalidDataException("Voc header offset is wrong");
+			if (vfh.Version != 0x010A)
+				throw new InvalidDataException("Voc header version not recognized");
+			if (vfh.ID != ~vfh.Version + 0x1234)
+				throw new InvalidDataException("Voc header id is bogus - expected: " +
+					(~vfh.Version + 0x1234).ToString("X") + " but value is : " + vfh.ID.ToString("X"));
+		}
+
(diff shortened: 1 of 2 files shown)


OpenRA.Game/FileFormats/VocLoader.cs
```diff
@@ -0,0 +1,348 @@
...
+		}
+
+		public override void SetLength(long value)
+		{
+			throw new NotImplementedException();
+		}
+
+		public override int Read(byte[] buffer, int offset, int count)
+		{
+			int bytesWritten = 0;
```
pchote: `var`?

OpenRA.Game/FileFormats/VocLoader.cs
```diff
@@ -0,0 +1,348 @@
...
+			var samplesLeft = Math.Min(count, buffer.Length - offset);
+			while (samplesLeft > 0)
+			{
+				int len = FillBuffer(samplesLeft);
```
pchote: `var`?

OpenRA.Game/FileFormats/VocLoader.cs
```diff
@@ -0,0 +1,348 @@
...
+					samplesLeftInBlock = currentBlock.Current.SampleBlock.Samples;
+					return;
+				}
+			}
+		}
+
+		public byte[] ReadAllBytes()
+		{
+			Rewind();
+			byte[] buffer = new byte[totalSamples];
```
pchote: `var`?

pchote:
Code looks good aside from the `var` style nits.  Can't test it, though.

teees:
Added a testcase commit, you need `VOC.PAK`, `ATRE.PAK`, `HARK.PAK` and `ORDOS.PAK` from some dune2 install and copy it to the d2k content folder

evgeniysergeev:
+1 works for me :+1:

OpenRA.Game/FileFormats/VocLoader.cs
```diff
@@ -0,0 +1,348 @@
...
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace OpenRA.FileFormats
+{
+	public class VocLoader : Stream
```
pchote: Can we drop the stream stuff for now?  It's going to be a while before we can support that properly, and these sound loaders are soon going to be hidden behind an interface where nobody can see this.

pchote: Definitely keep this on a separate branch for the future, but for now it would be better to merge a simple parser that isn't overloaded with a bunch of functionality that it doesn't yet need.

pchote:
I can confirm that this works, but I think we should resolve my request (possibly by convincing me why its not reasonable) before merging.

OpenRA.Game/FileFormats/VocLoader.cs
pchote: Note that streaming is covered by #8593.

Mailaender: See also https://github.com/OpenRA/OpenRA/pull/10069.

pchote: When we do eventually move to streaming I think it would be better to have a method on the reader that returns a stream instead of making the reader itself a stream.  This will cut down on boilerplate for the cases where we already have the raw data as a stream (e.g. loading aud from a mix file) and can return that directly.

Thoughts?

pchote: Actually, I may be wrong there.  This approach may actually be the best way.

penev92:
As discussed on IRC, this should wait for the sound loaders PR and then rebase on that.

OpenRA.Game/Sound/Sound.cs
```diff
@@ -58,6 +58,10 @@ ISoundSource LoadSound(string filename)
 				using (var s = GlobalFileSystem.Open(filename))
 					return LoadWave(new WavLoader(s));
 
+			if (filename.ToLowerInvariant().EndsWith("voc"))
+				using (var s = GlobalFileSystem.Open(filename))
```
obrakmann: This should be `Game.ModData.ModFiles.Open(filename)` now that #9853 is merged.

Mailaender:
https://github.com/OpenRA/OpenRA/pull/10069 has been merged.

teees:
rebased and made compatible with #10069

OpenRA.Game/FileFormats/VocLoader.cs
```diff
@@ -0,0 +1,371 @@
...
+			var vfh = VocFileHeader.Read(stream);
+
+			if (!vfh.Description.StartsWith("Creative Voice File"))
+				throw new InvalidDataException("Voc header description not recognized");
+			if (vfh.DatablockOffset != 26)
+				throw new InvalidDataException("Voc header offset is wrong");
+			if (vfh.Version != 0x010A)
+				throw new InvalidDataException("Voc header version not recognized");
+			if (vfh.ID != ~vfh.Version + 0x1234)
+				throw new InvalidDataException("Voc header id is bogus");
```
Phrohdoh: With all of these exceptions could you include what the bogus value is and what is expected?

OpenRA.Game/FileFormats/VocLoader.cs
```diff
@@ -0,0 +1,371 @@
...
+				throw new InvalidDataException("Invalid frequency divisor 256 in voc file");
+			if (vocSampleRate == 0xa5 || vocSampleRate == 0xa6)
+				return 11025;
+			else if (vocSampleRate == 0xd2 || vocSampleRate == 0xd3)
+				return 22050;
+			else
+				return (int)(1000000L / (256L - vocSampleRate));
+		}
+
+		private void Preload()
```
Mailaender: Explicit `private` is against the https://github.com/OpenRA/OpenRA/wiki/Coding-Standard

OpenRA.Game/FileFormats/VocLoader.cs
teees: Those are mostly magic numbers documented in the Creative Voice file format standard. http://wiki.multimedia.cx/index.php?title=Creative_Voice

Only a subset of this standard was used by westwood in dune 2 and is therefore supported here. Specifically `0x01` major and `0x0A` minor.

Bogus in this case means that the version is invalid / fails the validity check which is defined as `id == (~version + 0x1234)`.

Phrohdoh: I understand that, but I'm asking you put them in the message so users can know what is wrong instead of having to come to us and then we redirect them to that page or another with documentation.

teees: Will do. I'm just a bit upset since this apparently wasn't expected from any other loader.

https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/FileFormats/AudLoader.cs#L41
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/FileFormats/ImaAdpcmLoader.cs#L27
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/FileFormats/PngLoader.cs#L38

pchote: I think its fine as-is.  Like you mention, we don't expect this from the other loaders.

Mailaender:
Works as promised. Rebased my https://github.com/Mailaender/OpenRA/tree/dune2 test branch on top of this and I hear the Dune II voices and sounds. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/69879e0bd637a93dca2c63acf06f1e8aaa180776...14d5faeebc80a5dc439acd4b63c52fe8fd427871)

PR #10035, opened 2015-11-19
Increased default range of angles that missiles can launch at

### LINKED ISSUE #10024: Orcas have trouble hitting close targets
Regression from #8717. RA Longbows might be affected as well.

Tweaking the vertical ROT and maybe minimum range should be enough.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28337364-orcas-have-trouble-hitting-close-targets?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
This changes the default values so missiles without custom angles can launch at up to 90° downwards (enough for Orcas, Longbows and so on) and 180° upwards (basically straight up, for SAM sites and Adv. Guard Towers).

Closes #10024.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -35,10 +35,10 @@ namespace OpenRA.Mods.Common.Effects
 		public readonly bool Shadow = false;
 
 		[Desc("Minimum vertical launch angle (pitch).")]
-		public readonly WAngle MinimumLaunchAngle = WAngle.Zero;
+		public readonly WAngle MinimumLaunchAngle = new WAngle(-64);
 
 		[Desc("Maximum vertical launch angle (pitch).")]
-		public readonly WAngle MaximumLaunchAngle = new WAngle(64);
+		public readonly WAngle MaximumLaunchAngle = new WAngle(128);
 
 		[Desc("Minimum launch speed in WDist / tick")]
 		public readonly WDist MinimumLaunchSpeed = new WDist(75);


OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -35,10 +35,10 @@ public class MissileInfo : IProjectileInfo
 		public readonly bool Shadow = false;
 
 		[Desc("Minimum vertical launch angle (pitch).")]
-		public readonly WAngle MinimumLaunchAngle = WAngle.Zero;
+		public readonly WAngle MinimumLaunchAngle = new WAngle(-64);
 
 		[Desc("Maximum vertical launch angle (pitch).")]
-		public readonly WAngle MaximumLaunchAngle = new WAngle(64);
+		public readonly WAngle MaximumLaunchAngle = new WAngle(128);
```
matija-hustic: This still seems to be insufficient for the Advanced Guard Tower to target Orcas smoothly. 128 corresponds to a 45° pitch w.r.t. the horizontal. To enable targeting directly upwards you need a 90°, ie, a value of 256

penev92: Do we want to enable them to seemingly shoot straight up, though? (Or down, for that matter)
The artwork doesn't suggest it should.

matija-hustic: 144 suffices then

reaperrr: I'd rather leave the default at 128 here and set it to something larger for the AGT's missile yaml rules specifically.

reaperrr:
Updated with yaml changes.

pchote:
We'll need to be careful about the balance implications of this.  In the last release missiles would have to fire out and then spiral around in order to hit targets that are overhead.  This therefore ends up buffing the AGT, which is already considered OP ([reference](http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=17430&start=45)).

reaperrr:
Removed the yaml changes.

pchote:
Looks good to me ingame :+1:

penev92:
Looks good :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/6a909f6d367a07022be756904a3f870eb1ed1a9e)

Issue #8317, opened 2015-05-31
"Sand Shores"


Unit158:
Currently in D2k, the worms can attack on tiles that are slightly covered by rock (transitioning from sand to rock). The animation not only looks really bad, but as there is rock, my first intuition is that the worm should not be able to attack there. Perhaps these tiles should be changed so that they are neither buildable nor shai-hulud territory?

penev92:
I just checked the sandworm definition. It does not allow for movement on such tiles, which means those tiles have wrong terrain definitions.

Mailaender:
Probably related to https://github.com/OpenRA/OpenRA/issues/4243 which also be addressed by tileset tweaks.

PR #10037, opened 2015-11-19
D2k spice no longer erases map details & terrain fixes

OmegaBolt:
Fun one for y'all to test!

Added a new terrain type called SpiceSand (could change the name) which is only on the clear sand tiles and allows spice to grow there. The original Sand now disallows spice growth and is used for all sand detail tiles, which would previously be erased from the map by spice.

There are also a number of other fixes, such as preventing units from walking over large, visibly unpassable rock details, over the destroyed & civilian buildings etc

Lastly I also moved a couple of rock-cliff tiles that were very clearly in the wrong categories.

### DIFF
--- a/mods/d2k/rules/arrakis.yaml
+++ b/mods/d2k/rules/arrakis.yaml
@@ -43,6 +43,7 @@ sandworm:
 		TerrainSpeeds:
 			Sand: 100
 			Dune: 100
+			SpiceSand: 100
 			Spice: 100
 	Targetable:
 		TargetTypes: Ground
--- a/mods/d2k/rules/defaults.yaml
+++ b/mods/d2k/rules/defaults.yaml
@@ -64,6 +64,7 @@
 			Rock: 100
 			Transition: 100
 			Concrete: 100
+			SpiceSand: 100
 			Spice: 100
 			SpiceBlobs: 100
 			Dune: 50
@@ -169,6 +170,7 @@
 			Rock: 100
 			Transition: 100
 			Concrete: 100
+			SpiceSand: 100
 			Spice: 100
 			SpiceBlobs: 100
 			Dune: 80
--- a/mods/d2k/rules/misc.yaml
+++ b/mods/d2k/rules/misc.yaml
@@ -4,7 +4,7 @@ crate:
 		Name: Crate
 	Crate:
 		Lifetime: 120
-		TerrainTypes: Sand, Dune, Rock
+		TerrainTypes: Sand, Dune, Rock, SpiceSand
 	GiveCashCrateAction@1:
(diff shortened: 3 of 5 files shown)


pchote:
I thought that the `Transition` terrain type was supposed to be used for the tiles that transition to rock etc where the details are being lost?

OmegaBolt:
That's not where the details are being lost. It's largely stuff in the sand details category, or basically any tile that has terrain type sand that isn't the clear sand tile.

Transition is used for the sand-to-rock edges that has a speed modifier, and a nice radar colour that blends rock to sand, so I don't think it would be very suitable to use it for this.

For instance this large detail would get removed from the map.
detail

Mailaender:
> There are also a number of other fixes, such as preventing units from walking over large, visibly unpassable rock details, over the destroyed & civilian buildings etc.

Those buildings buried by sand are meant to be spots for infantry to hide from getting run over. We even implemented the special cursor. https://github.com/OpenRA/OpenRA/issues/2253

OmegaBolt:
But that's meant for the infantry rocks surely. That's what infantry hide on from vehicles. Running over buildings makes no sense visually, it looks ugly and breaks any illusion of perspective, not to mention it wasn't possible in the original game.

Mailaender:
Did you try to fix https://github.com/OpenRA/OpenRA/issues/4243 and https://github.com/OpenRA/OpenRA/issues/8317 or look the template definitions up somewhere in the original?

obrakmann:
With the spicebloom tweaks, this has become a problem for the release. It can become evident even on the shellmap, so I guess this is another addition to the milestone.

reaperrr:
There are too many changes to guarantee that I didn't overlook any errors, but what I can say with absolute certainty is that this is a huge improvement over bleed.
:+1:

Mailaender:
This is a bit hard to test now that the give resources developer cheat does nothing, but I also can't spot any regressions. :white_check_mark:

Mailaender:
Thanks!

OmegaBolt:
> Did you try to fix #4243 and #8317 or look the template definitions up somewhere in the original?

Those two issues will be fixed because of the new terrain type, but whether or not those specific tiles have been changed I'm not sure. I fixed the transition one, not sure about all those cliffs.

What I could do now is make a map that contains all the edited tiles. Was going to do it earlier but didn't have time.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/958c08cff8820a3c01d0c3d0d26cf7b1fa4ab83a...c24d6757f99307f320b75cfe8df597e286655122)

Issue #3768, opened 2013-08-31
EjectOnDeath should take cargo in account


Mailaender:
This applies to chinook and maybe badgers, too. They should eject a configurable percentage of their cargo when they got hit.

Mailaender:
Added in https://github.com/OpenRA/OpenRA/pull/10041.

Issue #7874, opened 2015-04-05
Add TS features required for a full remake


reaperrr:
Features that need to be added for a (nearly) complete remake:
- [x] Z-Buffer support for the renderer #7520
- [x] Split shadow sprites #3571
- [x] Installation from CD #3883 and asset download #3888 including music #8161 
- [x] Installation of Firestorm assets from CD and download. #12429
- [x] Turrets attaching to walls #4970
- [x] Radar dishes that stop spinning on low power #5138
- [x] Remappable resources #5155
- [x] Idle overlay sprite support #3681
- [x] Overlays that only animate with active production #5134
- [x] Construction yard crane overlay #5124
- [x] Voxel harvester unload animation #5797
- [x] Harvesting sprite animations #3458, #7204
- [x] Building plugins (components) #7812
- [x] Resource splattering for destroyed units containing Tiberium #2346
- [x] Hovering #7033
- [x] EMP
  - [x] stun actors #6653
  - [x] overlay animation #8581
  - [x] but as a super weapon #7349
- [x] AlphaImage support #7853
- [ ] Ion storms
  - [x] Lighting [#7863]
  - [ ] Disabling actors
  - [x] Lightning #7872
- [x] AmbientSound #7871
- [x] Individual weapon factory door overlays #7862
- [ ] Support for per-height-level ambient lighting [#3605]
- [x] Gates #9981
- [ ] Subterranean movement (and detection of it) [#7777]
- [x] Make actor height customisable [#7717]
- [x] draw stealthed actors semi-translucent instead of in shadow color #8996
- [x] Carryalls to manually pick up and release units #8954
- [ ] Laser Fence logic #6072, #10790
- [ ] Firestorm generator logic and effects #10789
- [x] Voxel image changing when moving over water terrain type (for amphibious APC) #7981
- [ ] Particle effect system (damage/wielding sparks, railgun, gas clouds, smoke)
- [x] WithChargeOverlay (for Obelisk) #7982
- [x] Tunnel system [#6410]
- [ ] High Bridges [#8917]
- [ ] Refactor missiles [#7509]
  - [x] Airburst for missiles
  - [ ] Weapon spawning from airburst
  - [x] Projectile (missile) acceleration
  - [x] Customisable cruise altitude for missiles (and support for arcing flight curve) [#3736]
- [ ] Add support for voxel projectiles [#7673]
- [ ] Add support for voxel debris
[...]

PR #10041, opened 2015-11-20
Unload passengers / cargo on death

teees:
Bit hacky approach, works by nudging passengers on exit so they don't bunch up.

The changes in EnterTransport will cancel any of the passengers activity, since mobile will discard nudges if not idle.

Kills any cargo it can't unload (when in water etc)

Edit:
No changes in EnterTransport, added a check in Enter to move on to the next activity and don't wait for the next tick. Also use Mobile.Nudge instead of INotifyBlockingMove.OnNotifyBlockingMove

http://gfycat.com/ImpishImaginativeFlea

https://gfycat.com/BoilingLimpAfricanjacana

### DIFF
--- a/OpenRA.Mods.Common/Traits/Cargo.cs
+++ b/OpenRA.Mods.Common/Traits/Cargo.cs
@@ -36,6 +36,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("When this actor is sold should all of its passengers be unloaded?")]
 		public readonly bool EjectOnSell = true;
 
+		[Desc("When this actor dies should all of its passengers be unloaded?")]
+		public readonly bool EjectOnDeath = false;
+
 		[Desc("Terrain types that this actor is allowed to eject actors onto. Leave empty for all terrain types.")]
 		public readonly HashSet<string> UnloadTerrainTypes = new HashSet<string>();
 
@@ -317,6 +320,26 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void Killed(Actor self, AttackInfo e)
 		{
+			if (Info.EjectOnDeath)
+				while (!IsEmpty(self) && CanUnload())
+				{
+					var passenger = Unload(self);
+					var cp = self.CenterPosition;
+					var inAir = self.World.Map.DistanceAboveTerrain(cp).Length != 0;
+					var positionable = passenger.Trait<IPositionable>();
+					positionable.SetPosition(passenger, self.Location);
+
+					if (!inAir && positionable.CanEnterCell(self.Location, self, false))
+					{
+						self.World.AddFrameEndTask(w => w.Add(passenger));
+						var nbm = passenger.TraitOrDefault<INotifyBlockingMove>();
+						if (nbm != null)
+							nbm.OnNotifyBlockingMove(passenger, passenger);
+					}
+					else
+						passenger.Kill(e.Attacker);
+				}
+
 			foreach (var c in cargo)
 				c.Kill(e.Attacker);
 
--- a/mods/ts/rules/aircraft.yaml
+++ b/mods/ts/rules/aircraft.yaml
@@ -22,6 +22,7 @@ DPOD:
 		MaxWeight: 1
 		PipCount: 1
 		UnloadVoice: Move
+		EjectOnDeath: true
 	Armament:
 		Weapon: Vulcan2
 	AttackHeli:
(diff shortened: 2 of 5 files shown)


mods/ts/rules/aircraft.yaml
```diff
@@ -22,6 +22,7 @@ DPOD:
 		MaxWeight: 1
 		PipCount: 1
 		UnloadVoice: Move
+		EjectOnDie: true
```
GraionDilach: You refer this as `EjectOnDeath` in code. I prefer the code's form though.

teees: Just noticed that, refactored it before commit and forgot about it

OpenRA.Mods.Common/Traits/Cargo.cs
```diff
@@ -300,6 +303,12 @@ public void Load(Actor self, Actor a)
 
 		public void Killed(Actor self, AttackInfo e)
 		{
+			if (Info.EjectOnDeath)
+			{
```
pchote: Style nit: these braces can go.

OpenRA.Mods.Common/Traits/Cargo.cs
```diff
@@ -331,7 +340,9 @@ void SpawnPassenger(Actor passenger)
 				w.Add(passenger);
 				passenger.Trait<IPositionable>().SetPosition(passenger, self.Location);
 
-				// TODO: this won't work well for >1 actor as they should move towards the next enterable (sub) cell instead
+				var mobile = passenger.TraitOrDefault<Mobile>();
```
pchote: Can you use `OnNotifyBlockingMove` instead?  Hardcoding mobile references through unrelated code kind of sucks.

OpenRA.Mods.Common/Traits/Cargo.cs
```diff
@@ -36,6 +36,9 @@ public class CargoInfo : ITraitInfo, Requires<IOccupySpaceInfo>
 		[Desc("When this actor is sold should all of its passengers be unloaded?")]
 		public readonly bool EjectOnSell = true;
 
+		[Desc("When this actor dies should all of its passengers be unloaded?")]
+		public readonly bool EjectOnDeath = false;
```
pchote: It would be nice to have an option to eject the infantry with a health hit, but that makes this a bit more complicated so isn't required for my :+1:

pchote:
Works well ingame and matches the original TS behaviour. :+1: after addressing the nits above.

pchote:
Closes #3768.

teees:
Changed and rebased

OpenRA.Mods.Common/Traits/Cargo.cs
```diff
@@ -317,6 +320,10 @@ public void Load(Actor self, Actor a)
 
 		public void Killed(Actor self, AttackInfo e)
 		{
+			if (Info.EjectOnDeath)
+				while (!IsEmpty(self) && CanUnload())
```
obrakmann: Is there a reason why you don't check for `DistanceAboveTerrain` here already? Then you wouldn't even need to run `SpawnPassengers`, including all the trait lookups and whatnot, in that situation.

OpenRA.Mods.Common/Traits/Cargo.cs
```diff
@@ -343,13 +350,20 @@ public void Sold(Actor self)
...
-			});
+			if (positionable.CanEnterCell(self.Location, self, false) && !inAir)
+			{
+				self.World.AddFrameEndTask(w => w.Add(passenger));
+				var nbm = passenger.TraitOrDefault<INotifyBlockingMove>();
+				if (nbm != null)
+					nbm.OnNotifyBlockingMove(passenger, passenger);
+			}
+			else
+				passenger.Dispose(); // TODO: this may need to be a Kill() for statistics
```
obrakmann: Yes, I think `Kill()` would be better.

OpenRA.Mods.Common/Traits/Cargo.cs
teees: The way it is now not really anymore. I initially wanted to support parachutes as well (see EjectOnDeath). Will change it

Mailaender:
Also needs a rebase.

teees:
rebased and adapted to the requested changes

Mailaender:
Sorry, needs a rebase again.

teees:
rebased

Mailaender:
Both cases work fine. :white_check_mark:

Mailaender:
Nicely done. Another tick on our list: https://github.com/OpenRA/OpenRA/issues/7874

PR #10045, opened 2015-11-20
Rename Chan to Scientist.

ghost:
These infantry are used as scientists in ra missions.

I kept the 'CHAN' easter egg reference to C&C TD.

PR #10048, opened 2015-11-21
Add CHAN to EmitInfantryOnSell on certain structures.

ghost:
Currently when selling buildings there is a certain randomness involved in what infantry are returned. Selling a msilo could net you an engineer and a rifle or it could net you two unarmed civilians.

Due to the RNG, selling these high-tech structures in MP doesn't really have any influence on a match. These structures are rarely sold for any reason other than imminent destruction.

This PR adds the possibility to receive the scientist when selling certain high-tech structures. It is purely an 'easter egg' polish linking multi-player to single-player. This should have no effect in single-player missions.

Depends on #10045

### DIFF
--- a/mods/ra/rules/defaults.yaml
+++ b/mods/ra/rules/defaults.yaml
@@ -442,6 +442,11 @@
 	Sellable:
 		SellSounds: cashturn.aud
 
+^ScienceBuilding:
+	Inherits: ^Building
+	EmitInfantryOnSell:
+		ActorTypes: e1,e1,e1,e1,c1,c2,c3,c4,c5,c6,c7,c8,c9,c10,e6,e6,e6,e6,e6,chan,chan,chan,chan
+
 ^Defense:
 	Inherits: ^Building
 	Targetable:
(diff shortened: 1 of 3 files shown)


mods/ra/rules/defaults.yaml
```diff
@@ -442,6 +442,11 @@
 	Sellable:
 		SellSounds: cashturn.aud
 
+^ScienceBuilding
```
Mailaender: Missing `:` at the end.

cjshmyr:
Also removes the chance of `e1` coming from these buildings, which I think would be nice if there were still a few shares of it. Moreso thinking about the cases you sell in time against a Tanya

r34ch:
@cjshmyr Good point, they are military structures after all. 

Updated.

Mailaender:
:+1: Nice gimmick.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7631fc5a05ebbea95cf812497efcf20c5570cf77)

Issue #9795, opened 2015-10-28
Gems patches with a "greener" gem


StexoO:
If you watch this replay from RA95 https://www.youtube.com/watch?v=AFb-xgDDf7Q you can see one or two kind(s) of gem patches with a "greener" (diagonal) gem on it.
I don't see this patch in the OpenRA Map Editor and I don't remember seeing this in RA95 (I don't have the game right here to check), so is this patch missing from OpenRA or is this a modded RA95 ?
I also found this picture in the "MONEY MANAGEMENT" section of this page http://redalert1.com/skill.php :
![im1](https://cloud.githubusercontent.com/assets/8930382/10804199/8e46aa0a-7dc6-11e5-9397-31bd4d35f000.jpg)

MicroJOo:
indeed , there is also a more cyan looking gem , a really shiny one .. just like the green one ... 

the reason that we dont have it , is because our gem-growing doesnt support some of the gem-sprites or a similar reason  ... @pchote knows why

Mailaender:
Investigated this a bit and filed https://github.com/OpenRA/OpenRA/issues/10066. It was more difficult to debug this in the past, because we didn't have a map editor with proper resource rendering.

PR #10066, opened 2015-11-22
Fixed highest density sprite frame never being rendered

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/9795. In fact this changes the whole rendering. Not sure if my math is correct, but we now see the last gem frame and I believe single ore patches are also rendered in a thin manner.

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
+++ b/OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
@@ -163,7 +163,7 @@ namespace OpenRA.Mods.Common.Traits
 			NetWorth += t.Density * type.Info.ValuePerUnit;
 
 			var sprites = type.Variants[t.Variant];
-			var frame = int2.Lerp(0, sprites.Length - 1, t.Density - 1, type.Info.MaxDensity);
+			var frame = int2.Lerp(0, sprites.Length - 1, t.Density, type.Info.MaxDensity);
 			t.Sprite = sprites[frame];
 
 			return t;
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Traits/World/EditorResourceLayer.cs
```diff
@@ -163,8 +163,8 @@ public virtual CellContents UpdateDirtyTile(CPos c)
 			NetWorth += t.Density * type.Info.ValuePerUnit;
 
 			var sprites = type.Variants[t.Variant];
-			var frame = int2.Lerp(0, sprites.Length - 1, t.Density - 1, type.Info.MaxDensity);
-			t.Sprite = sprites[frame];
+			var frame = int2.Lerp(0, sprites.Length, t.Density, type.Info.MaxDensity);
+			t.Sprite = sprites[frame - 1];
```
pchote: This math doesn't look correct: You'll crash if `t.Density` is low (&rarr; `frame == 0` &rarr; indexing with `-1`).

pchote: Math-wise, the correct code should be:
```csharp
var frame = int2.Lerp(0, sprites.Length - 1, t.Density, type.Info.MaxDensity);
t.Sprite = sprites[frame];
```
If this doesn't work, then I expect the problem is actually in the values that `t.Density` are set to.

Mailaender: I think it can't crash anymore, because the Density = 0 bug has already been fixed in https://github.com/OpenRA/OpenRA/pull/5571.

pchote: It will happen for non-zero densities if sprites.Length < MaxDensity.

Mailaender: Updated with your math.

abcdefg30:

pchote:
Works as advertised :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7d578e297a947293f61054dd738e789e58a1975d

Issue #9671, opened 2015-10-18
Game Crash: When hovering over "Latency Ind. Icon" (Own Server) #crash


theLobster:
After Creating own Online Game (Con. with Masterserver established) the game crashed during hovering to the Lat. Ind. Icon:

Exception.log:
```
Red Alert Mod at Version release-20150919
Operating System: Windows (Microsoft Windows NT 6.1.7601 Service Pack 1)
Runtime Version: .NET CLR 4.0.30319.18444
Exception of type `System.Reflection.TargetInvocationException`: Ein Aufrufziel hat einen Ausnahmefehler verursacht.
Inner Exception of type `Mono.Nat.MappingException`: Error 401: Invalid Action
     bei Mono.Nat.Upnp.UpnpNatDevice.EndGetExternalIP(IAsyncResult result)
   bei Mono.Nat.AbstractNatDevice.GetExternalIP()
   bei OpenRA.Mods.Common.Widgets.Logic.ClientTooltipLogic..ctor(Widget widget, TooltipContainerWidget tooltipContainer, OrderManager orderManager, Int32 clientIndex)
   bei System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   bei System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   bei OpenRA.ObjectCreator.CreateUsingArgs(ConstructorInfo ctor, Dictionary`2 args)
   bei OpenRA.ObjectCreator.CreateObject[T](String className, Dictionary`2 args)
   bei OpenRA.Widgets.Widget.<PostInit>c__AnonStorey0.<>m__0(String l)
   bei System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   bei System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   bei System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   bei OpenRA.Widgets.Widget.PostInit(WidgetArgs args)
   bei OpenRA.WidgetLoader.LoadWidget(WidgetArgs args, Widget parent, MiniYamlNode node)
   bei OpenRA.WidgetLoader.LoadWidget(WidgetArgs args, Widget parent, String w)
   bei OpenRA.Widgets.Ui.LoadWidget(String id, Widget parent, WidgetArgs args)
[...]

Issue #9405, opened 2015-09-22
game crash on mouse over latency with my own 'server' 


Zjordy:
hey since the new release, whenever i use automatic port forwardning (upnp) to make a server everything works, but when i mouse over the latency red/orange/green light, my game will crash.
seems similar like a problem i had before in the older release and i think got it fixed then when i redownloaded red alert

Zjordy:
similar to my post: #9112

pchote:
Maybe a duplicate of #8787?

penev92:
Just had a similar report on bleed, commit 0e9257abcf39e2118b7b53df5c9f13fad429b828, happening in the skirmish lobby:

```
Crystallized Doom Mod at Version 0.3 Closed Dev Beta - 2015-26-09
Operating System: Windows (Microsoft Windows NT 6.1.7600.0)
Runtime Version: .NET CLR 4.0.30319.17929
Exception of type `System.Reflection.TargetInvocationException`: Ein Aufrufziel hat einen Ausnahmefehler verursacht.
Inner Exception of type `Mono.Nat.MappingException`: Error 8: The underlying connection was closed: The connection was lost unexpectedly ..
     bei Mono.Nat.Upnp.UpnpNatDevice.EndGetExternalIP(IAsyncResult result)
   bei Mono.Nat.AbstractNatDevice.GetExternalIP()
   bei OpenRA.Mods.Common.Widgets.Logic.LobbyUtils.GetExternalIP(Int32 clientIndex, OrderManager orderManager)
   bei OpenRA.Mods.Common.Widgets.Logic.ClientTooltipLogic..ctor(Widget widget, TooltipContainerWidget tooltipContainer, OrderManager orderManager, Int32 clientIndex)
```

PR #10068, opened 2015-11-22
Fixed crashes when NatDevice.GetExternalIP() fails

Mailaender:
Rule of thumb when working with Mono.NAT: try/catch everything.

Fixes https://github.com/OpenRA/OpenRA/issues/9405.
Fixes https://github.com/OpenRA/OpenRA/issues/9671.

Compare https://github.com/OpenRA/OpenRA/pull/2993.

It isn't obvious, but somehow even simply getting an external IP address can cause MappingExceptions?!

### DIFF
--- a/OpenRA.Game/Network/UPnP.cs
+++ b/OpenRA.Game/Network/UPnP.cs
@@ -9,7 +9,7 @@
 #endregion
 
 using System;
-
+using System.Net;
 using Mono.Nat;
 
 namespace OpenRA.Network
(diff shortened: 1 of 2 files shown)


OpenRA.Game/Network/UPnP.cs
```diff
@@ -116,5 +116,22 @@ public static void RemovePortforward()
...
+
+		public static IPAddress GetExternalIP()
+		{
+			if (natDevice == null)
+				return null;
+
+			try
+			{
+				var ip = natDevice.GetExternalIP();
+				return ip;
```
RoosterDragon: Maybe `return natDevice.GetExternalIP();`?

RoosterDragon:
:+1: Seems sensible.

OpenRA.Game/Network/UPnP.cs
Mailaender: Indeed.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/33a9ee760ffb4035fae438470e4c144443a5efe5...4bd64db81c04649e74fbaf78415937c10fed346b)

PR #10069, opened 2015-11-22
Add plumbing for mod-defined sound loaders

penev92:
Use cases are #10032 and moving `AudLoader` and `WavLoader` out of the engine and into Common.
This brings sound loaders on par with sprite loaders and their `ISpriteLoader` and enables every mod to define it's own loaders.

Note: this is for sound effects/voices only. Music is still [hardcoded](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/GameRules/MusicInfo.cs#L43-L48). Fixing that both is out of scope here and will conflict with the unglobaling of the file system.

### DIFF
--- a/OpenRA.Game/FileFormats/AudLoader.cs
+++ b/OpenRA.Game/FileFormats/AudLoader.cs
@@ -175,8 +175,22 @@ namespace OpenRA.FileFormats
 		{
 			channels = sampleBits = sampleRate = 0;
 
-			if (!LoadSound(stream, out rawData))
+			try
+			{
+				if (!LoadSound(stream, out rawData))
+					return false;
+			}
+			catch (Exception e)
+			{
+				// LoadSound() will check if the stream is in a format that this parser supports.
+				// If not, it will simply return false so we know we can't use it. If it is, it will start
+				// parsing the data without any further failsafes, which means that it will crash on corrupted files
+				// (that end prematurely or otherwise don't conform to the specifications despite the headers being OK).
+				Log.Write("debug", "Failed to parse AUD file {0}. Error message:".F(fileName));
+				Log.Write("debug", e.ToString());
+				rawData = null;
 				return false;
+			}
 
 			channels = 1;
 			sampleBits = 16;
--- a/OpenRA.Game/FileFormats/AudLoader.cs
+++ b/OpenRA.Game/FileFormats/AudLoader.cs
@@ -36,16 +36,18 @@ namespace OpenRA.FileFormats
 			Chunk c;
 			c.CompressedSize = s.ReadUInt16();
 			c.OutputSize = s.ReadUInt16();
+
 			if (s.ReadUInt32() != 0xdeaf)
 				throw new InvalidDataException("Chunk header is bogus");
 			return c;
 		}
 	}
 
-	public static class AudLoader
+	public class AudLoader : ISoundLoader
 	{
-		static int[] indexAdjust = { -1, -1, -1, -1, 2, 4, 6, 8 };
-		static int[] stepTable =
+		static readonly int ExpectedSampleRate = 22050;
+		static readonly int[] IndexAdjust = { -1, -1, -1, -1, 2, 4, 6, 8 };
+		static readonly int[] StepTable =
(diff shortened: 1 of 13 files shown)


abcdefg30:
> OpenRA.Game/FileFormats/AudLoader.cs:L49: [SA1311] Static readonly fields must start with an upper-case letter: indexAdjust.
OpenRA.Game/FileFormats/AudLoader.cs:L50: [SA1311] Static readonly fields must start with an upper-case letter: stepTable.

penev92:
Fixed.

penev92:
As discussed today on IRC this should be the first PR of the two to be merged so the other can adapt to the interface.

OpenRA.Game/ModData.cs
```diff
@@ -53,6 +54,18 @@ public ModData(string mod, bool useLoadScreen = false)
...
+			foreach (var format in Manifest.SoundFormats)
+			{
+				var loader = ObjectCreator.FindType(format + "Loader");
+				if (loader == null || !loader.GetInterfaces().Contains(typeof(ISoundLoader)))
+					throw new InvalidOperationException("Unable to find a sound loader for type '{0}'.".F(format));
+
+				soundLoaders.Add((ISoundLoader)ObjectCreator.CreateBasic(loader));
+			}
+
+			SoundLoaders = soundLoaders.ToArray();
```
RoosterDragon: Feels like you could extract a common method for the sound and sprite loaders.

RoosterDragon:
May be worth considering adding a try-catch around the internals of the sound parsing so you can return false for `TryParseSound` even when an exception occurs.

penev92:
The internal parsing code was intentionally left volatile so people couldn't get away with bogus files (or get the less helpful "No suitable parsers found" exception).

RoosterDragon:
You can still throw in the normal case.

penev92:
Added two new commits with the requested changes.

GraionDilach:
That last commit sounds like a workaround for #10124.

OpenRA.Game/FileFormats/AudLoader.cs
```diff
@@ -153,7 +166,33 @@ public static byte[] LoadSound(Stream s)
...
+			channels = sampleBits = sampleRate = 0;
+
+			try
+			{
+				if (!LoadSound(stream, out rawData))
+					return false;
+			}
+			catch (Exception e)
+			{
+				Log.Write("debug", "Failed to parse AUD file {0}. Error message:".F(fileName));
```
pchote: You're using TryParseSound for format detection as well as just loading.  This means you will spam the debug logs with "failed to load aud" errors for each valid wav if you define aud as the first loader in the list.

pchote: If you want to keep the current interface/api you will need to keep the try/catch, but remove the logging.

penev92: Actually no.
The inner parsing method (`LoadSound()` in this case) returns `true`/`false` depending on whether the file is in a format this parser accepts. So the `AudLoader` will just return `false` for WAV files and vice-versa instead of throwing.
The exceptions happen when the file is corrupt or otherwise bogus - ends prematurely, the metadata at the start doesn't correspond to the actual data (like in #10124) or other such situations. So only actual exceptions will be logged.

Perhaps I should add a comment to make that clear by just glancing at the code.

pchote: Ok, makes sense.  Yes, please do add a comment.

penev92:
Added the comments.

abcdefg30:
Needs a rebase.

penev92:
Rebased.

RoosterDragon:
:+1: Looks alright to me and sounds still work.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/40c248d9126f89c303cbf2696d4274f60df7984a...bd1daadb65b51ca711b3b3f2db2c790f4893cbd3)

PR #10072, opened 2015-11-22
Monster tank madness polish

### LINKED ISSUE #10049: Dr Dmitri doesn't have a Russian accent
In the mission 'Monster Tank Madness', you have to rescue a Soviet defector. 

Currently he uses the generic male civilian voice which I feel sounds out of place. Stavros' voiceset doesn't sound right here either.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28492643-dr-dmitri-doesn-t-have-a-russian-accent?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ghost:
Few bit of polish for this mission.
1. Dr Demitri now uses a Russian accent - Fixes #10049
2. Prison / tech centre owner tooltip hidden
3. Add some mission fail checks for the ability to land on the beach

Remaining issues

-When LST is killed after the spen is destroyed and forces haven't landed at the beach, the mission does not fail - (need an efficient lua method of checking if a produced unit of type is killed - please recommend)
-3rd e1 reinforcement at start of mission stands still on the bridge unless the light tanks are ordered to keep moving (don't stop at rally point)

### DIFF
--- a/mods/ra/maps/monster-tank-madness/monster-tank-madness.lua
+++ b/mods/ra/maps/monster-tank-madness/monster-tank-madness.lua
@@ -107,6 +107,7 @@ SendAlliedUnits = function()
 
 			Utils.Do(units, function(unit)
 				if unit.Type == "e6" then
+					Engineer = unit
 					Trigger.OnKilled(unit, LandingPossible)
 				end
 			end)
@@ -413,7 +414,17 @@ InitTriggers = function()
 		end
 	end)
 
+	LstProduced = 0
 	Trigger.OnKilled(USSRSpen, LandingPossible)
+	Trigger.OnProduction(USSRSpen, function(self, produced)
+		if produced.Type == "lst" then
+			LstProduced = LstProduced + 1
+			Trigger.OnKilled(produced, function()
+				LstProduced = LstProduced - 1
+				LandingPossible()
+			end)
+		end
+	end)
 end
 
 WorldLoaded = function()
--- a/mods/ra/maps/monster-tank-madness/monster-tank-madness.lua
+++ b/mods/ra/maps/monster-tank-madness/monster-tank-madness.lua
@@ -103,11 +103,23 @@ SendAlliedUnits = function()
 	Media.PlaySpeechNotification(player, "ReinforcementsArrived")
 	Utils.Do(AlliedUnits, function(table)
 		Trigger.AfterDelay(table.delay, function()
-			Reinforcements.Reinforce(player, table.types, { StartEntryPoint.Location, StartMovePoint.Location }, 18)
+			local units = Reinforcements.Reinforce(player, table.types, { StartEntryPoint.Location, StartMovePoint.Location }, 18)
+
+			Utils.Do(units, function(unit)
+				if unit.Type == "e6" then
+					Trigger.OnKilled(unit, LandingPossible)
+				end
+			end)
 		end)
 	end)
 
 	Trigger.AfterDelay(DateTime.Seconds(1), function() InitialUnitsArrived = true end)
+end
+
+LandingPossible = function()
+	if not beachReached and (USSRSpen.IsDead or Engineer.IsDead) and LstProduced < 1 then
+		 player.MarkFailedObjective(CrossRiver)
+	end
 end
 
 SuperTankDomeInfiltrated = function()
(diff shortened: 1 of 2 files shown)


mods/ra/maps/monster-tank-madness/monster-tank-madness.lua
```diff
@@ -103,13 +103,35 @@ SendAlliedUnits = function()
...
 	Trigger.AfterDelay(DateTime.Seconds(1), function() InitialUnitsArrived = true end)
 end
 
+LandingPossible = function()
+	if not beachReached then
+		fail = true
+		Utils.Do(Map.ActorsInBox(Map.TopLeft, Map.BottomRight), function(a)
+			if a.Owner == player then
+				if a.Type == "spen" or a.Type == "lst" then
+					fail = false
```
kyrylo: Can't you just avoid the `fail` variable (which is apparently global, btw)  and replace this line with `player.MarkFailedObjective(CrossRiver)` (and possibly immediately `return`)?

r34ch: ...that's a good point. Remnants of a re-factor.

mods/ra/maps/monster-tank-madness/monster-tank-madness.lua
```diff
@@ -103,13 +103,32 @@ SendAlliedUnits = function()
...
 
 	Trigger.AfterDelay(DateTime.Seconds(1), function() InitialUnitsArrived = true end)
 end
 
+LandingPossible = function()
+	if not beachReached then
+		Utils.Do(Map.ActorsInBox(Map.TopLeft, Map.BottomRight), function(a)
+			if a.Owner == player then
+				if a.Type == "spen" or a.Type == "lst" then
+					player.MarkFailedObjective(CrossRiver)
```
abcdefg30: Wait a moment, doesn't that mean that the mission fails now as soon as the player owns a sub pen or transport?

Also, you're calling this from the `OnKilled` trigger of the subpen, so it should be already dead by then and you don't need to check `a.Type == "spen"`.

So, all in all that should like:
```
if a.Owner == player and a.Type == "lst" then
	return
end
player.MarkFailedObjective(CrossRiver)

Comment (r34ch) [inline mods/ra/maps/monster-tank-madness/monster-tank-madness.lua]: Yes it does - I was too hasty pushing. 

The reason for the spen check is that the same code is called  when the engineer dies and this was the most DRY option I could think of.

**Comment by Mailaender** [comment]
I think this needs some kind of notification instead of just mission failed. Let me play the casual gamer here. "Just attack moved into the enemy base. Why did the mission suddenly fail?"

**Comment by r34ch** [comment]
@Mailaender that's a fair point. How about a DisplayMessage; "Unable to transport our forces to the mainland" with a 3 second delay prior to failing the objective?

**Comment by Mailaender** [comment]
That should do it.

**Comment by r34ch** [comment]
Added failure message followed by a 3 second delay prior to failing the mission.

**Comment by abcdefg30** [comment]
I filed a small pr with some suggestions on how to avoid this ugly `Utils.Any(Map.ActorsInBox...` call: https://github.com/r34ch/OpenRA/pull/1

:+1: / :white_check_mark: with that.

**Comment by r34ch** [comment]
Thanks @abcdefg30; that's a cleaner and more logical way of dealing with that a situation. 

Actually a little embarrassed I didn't think along those lines haha

**Comment by abcdefg30** [comment]
> Actually a little embarrassed I didn't think along those lines haha

No big deal, I have a little bit more experience. :)
Anyway, could you please rebase this PR onto latest bleed so we get rid of the merge-commit?

**Comment by r34ch** [comment]
Rebased.

File: mods/ra/maps/monster-tank-madness/monster-tank-madness.lua [hunk 3]
```diff
@@ -103,13 +103,27 @@ SendAlliedUnits = function()
...
+			end)
 		end)
 	end)
 
 	Trigger.AfterDelay(DateTime.Seconds(1), function() InitialUnitsArrived = true end)
 end
 
+LandingPossible = function()
+	if not beachReached and (USSRSpen.IsDead or Engineer.IsDead) and LstProduced < 1 then
+		Media.DisplayMessage("Unable to transport our forces to the mainland.")
```
obrakmann: FWIW, my opinion is that the objectives setup is wrong if you have to take such measures to make the player understand why a mission failed.

pchote: I agree.  This could be easily covered by changing "Find a way to transport your forces to the mainland." to "Secure transport to the mainland" and then failing that.

obrakmann:
:+1: after removing the message and the delay.

r34ch:
Removed the message and delay as well as rewording the objective as per @pchote's suggestion.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c08c7182e42f1486b7cb2ad9261ec3d21b702a5a)

Issue #9872, opened 2015-11-03
Implement multiple resource queues


scshunt:
Currently, there are only two types of resources: "cash" and "resources". This should not be hardcoded and works badly with TS (which has veins), RA2 (which doesn't have resources), and AoE (which has several different primary types of resources), to name a few.

The obvious part of this work is to implement separate traits for each different type of resource, replacing `PlayerResources`, which may have different properties (to begin with, some resources must have capacities; some must not).

The other part of this is to implement RA/TD/D2K's behaviour of using resources to spend for cash. This should be implemented using a separate trait to indicate the two. There are two different proposed designs (note that the naming is tentative in both designs):

The first is to have an `ResourceAlias` trait which would implement the same abstract interface as a regular resource, but be implemented differently:

    ResourceAlias: ore-or-cash
        Aliases: ore, cash

This would simply be a resource which would link back into the others. Spending this resource would go in order of the `Aliases` array, so in this case, ore first, then cash.

The second is to have a `SubstituteResource` trait which would allow spending one resource in place of another:

    SubstituteResource:
        Use: ore
        As: cash
        Priority: 1

Resources would be used in decreasing order of priority, so a positive priority indicates that ore would be used first; a negative priority would be used to spend ore only after cash was exhausted.

Feedback on which design to use is appreciated.

pchote:
The main advantage to the `ResourceAlias` is that it comes for free - no additional plumbing or coupling between traits is required.  The `SubstituteResource` will need an explicit check from either the resource being queried or the things trying to use it.

scshunt:
The more I think about it, the more I think that `ResourceAlias` is the more flexible design, since it provides more control to the client of the resource, and it allows for lots of possible permutations, possibly at the same time. On the flip side, it means that every client needs to be aware of all the possible aliases, and I doubt that flexibility would be needed in actual, reasonable module designs.

atlimit8:
I agree with the preference for `ResourceAlias` concept with it implementing `IResourceType`.
Except I would name it differently:
```yaml
ResourceConversion@Funds:
    Name: Funds
    From: Ore, Gems, Cash
```

GraionDilach:
Could someone elaborate to me why allowing to have multiple `PlayerResources` ain't a way to go?

pchote:
@GraionDilach: It is.  It has just been renamed.

Phrohdoh:
```
[13:09:56]	<Phrohdoh> hi all
[...]

PR #10075, opened 2015-11-22
Multiple resource queues: YAML & Trait refactors

ghost:
This represents the first work towards refactoring to permit multiple resource queues. The first commit requires explicit specification of resource types on various traits which affected cash or resources previously, and the second commit lets a unit's value span across multiple resources.

One thing that's missing here is handling of text. Currently, for all text purposes, all resources are still just summed up and used.

I'm also unsure of the map upgrades. In particular, I'm unsure how the oramaps are affected and whether the upgrade code works correctly. The upgrade code seemed to fail on upgrading `^Templates`, but worked fine on regular units. Any comments/feedback would be appreciated!


atlimit8:
Needs to be rebased to remove merged commits. Also, MiniYaml comments should be restored after applying rule upgrades.

pchote:
This keeps a single `PlayerResources` trait and introduces resource type strings to the methods (with an internal hardcoded alias for cash -> resources), which goes against what we planned in #9872.  From discussion in IRC @scshunt plans on this as a stepping stone to splitting the traits.

I don't like this approach because it means changing our interface to `PlayerResources` twice.  It is only a small extra amount of effort to the split now, which saves a lot of churn and keeps the roadmap clearer.

pchote:
If I were doing this myself, here's the steps I would follow for a smooth review and integration process:

1. Split out the "Silos needed" checks and audio to its own trait (pr of its own)
2. Split cash/resources (pr of its own) by:
    * Introducing a `Type` field on `PlayerResourcesInfo`
    * Adding two `PlayerResources` definitions to the player yaml, with `Type: cash` and `Type: resources`
    * Removing all the `*Cash` methods from `PlayerResources`
    * Replacing all the current uses of `playerActor.Trait<PlayerResources>().GiveCash()` with `playerActor.TraitsImplementing<PlayerResources>().FirstOrDefault(t => t.Info.Type == "cash")`, and similar for resources.
3. Add support for resource aliases to remove the hardcoded cash/resource prioritization
4. Add Type fields to traits that interact with PlayerResources in order to support arbitrary resources.

PR #10077, opened 2015-11-22
Fixes passability/buildability of several TD cliff tiles

reaperrr:
These definitely could not be moved/built on in the original (in fact, templates 29, 30, 31 and 43 have already been fixed before on desert tileset).

Note: There are a few more debatable cases that were of "Clear" terrain type in the original, but probably shouldn't have been (going by visuals/consistency), but I left those out for now.

### DIFF
--- a/mods/cnc/tilesets/desert.yaml
+++ b/mods/cnc/tilesets/desert.yaml
@@ -264,9 +264,9 @@ Templates:
 		Size: 3,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
+			0: Rock
 			1: Rock
-			2: Clear
+			2: Rock
 			3: Rock
 			4: Rock
 			5: Clear
@@ -362,7 +362,7 @@ Templates:
 		Size: 3,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
+			0: Rough
 			1: Rock
 			2: Rock
 			3: Rock
--- a/mods/cnc/tilesets/jungle.yaml
+++ b/mods/cnc/tilesets/jungle.yaml
@@ -408,9 +408,9 @@ Templates:
 		Size: 3,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
+			0: Rock
 			1: Rock
-			2: Clear
+			2: Rock
 			3: Rock
 			4: Rock
 			5: Clear
@@ -449,8 +449,8 @@ Templates:
 		Size: 2,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
-			1: Clear
+			0: Rock
+			1: Rock
 			2: Rock
 			3: Rock
 	Template@30:
@@ -459,8 +459,8 @@ Templates:
 		Size: 2,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
-			1: Clear
+			0: Rock
+			1: Rock
 			2: Rock
 			3: Rock
 	Template@31:
@@ -469,8 +469,8 @@ Templates:
 		Size: 2,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
-			1: Clear
+			0: Rock
+			1: Rock
 			2: Rock
 			3: Rock
 	Template@32:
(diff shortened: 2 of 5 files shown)


chrisforbes:
Have you verified that no existing maps we care about rely on this?

If so, :+1:

reaperrr:
I don't know which maps we care about (singleplayer missions would actually be more true to the original with this PR, since WW designed them with these passabilities) and there are too many to play each of them, but I looked at their previews and none of our shipping maps look like this could play a role.

pchote:
Do we need a utility to check our passabilities against those baked into the tiles?  That shouldn't be hard to build.

mods/cnc/tilesets/desert.yaml
```diff
@@ -264,9 +264,9 @@ Templates:
 		Size: 3,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
+			0: Rock
```
abcdefg30: Are you sure about this change?
tdcliff
looks like it should be passable.

mods/cnc/tilesets/jungle.yaml
```diff
@@ -408,9 +408,9 @@ Templates:
 		Size: 3,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
+			0: Rock
```
abcdefg30: Same here.

mods/cnc/tilesets/temperat.yaml
```diff
@@ -408,9 +408,9 @@ Templates:
 		Size: 3,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
+			0: Rock
```
abcdefg30: And here.

mods/cnc/tilesets/winter.yaml
```diff
@@ -408,9 +408,9 @@ Templates:
 		Size: 3,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
+			0: Rock
```
abcdefg30: And here.

mods/cnc/tilesets/jungle.yaml
```diff
@@ -449,8 +449,8 @@ Templates:
 		Size: 2,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
-			1: Clear
+			0: Rock
```
abcdefg30: Doesn't this want to be `Rough` here and below?
tdcliff2
(As you made a similiar looking spot `Rough` on tile number 35.)

mods/cnc/tilesets/temperat.yaml
```diff
@@ -449,8 +449,8 @@ Templates:
 		Size: 2,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
-			1: Clear
+			0: Rock
```
abcdefg30: I think this (and below) should remain passable.
tdcliff3

mods/cnc/tilesets/temperat.yaml
```diff
@@ -589,8 +589,8 @@ Templates:
...
```
abcdefg30: Same here.
tdcliff4

mods/cnc/tilesets/winter.yaml
```diff
@@ -449,8 +449,8 @@ Templates:
 		Size: 2,2
 		Category: Cliffs
 		Tiles:
-			0: Clear
-			1: Clear
+			0: Rock
```
abcdefg30: Here on this three tiles again:
tdcliff5

mods/cnc/tilesets/winter.yaml
```diff
@@ -589,8 +589,8 @@ Templates:
...
```
abcdefg30: And here:
tdcliff6

reaperrr:
@abcdefg30: I didn't go by how it looks, but rather how it was in the original. Additionally, it might look ok with infantry and small vehicles, but not with mammoth tanks or mcv (in my opinion, at least).

pchote:
@reaperrr how did you verify the config in the original?

reaperrr:
By testing (it's only a few tiles, afterall).

reaperrr:
Since this is a relatively insignificant fix and not everyone seems to agree, I'll close this for now.

Issue #8670, opened 2015-07-06
Chronoshift did not chrono all units back


Basti504:
Just encountered an interesting bug, when I chronod my tanks "german"-style into the enemy base the game surprisingly chronod all units back except one at the end of the chrono action.

![Permanent Chrono](http://i.imgur.com/MklvNAu.gif)

Replay:
https://www.dropbox.com/s/3qfqkdbp9ck3toi/OpenRA-2015-07-06T183755Z_chrono.orarep?dl=0
(Around 17:30)
##

Basti504:
After some playing with the source code, I see the reason it does not teleport that unit back. 
In the same game, the same breakpoint is hit when propably another chronoshift was busy.

Hope this helps :) 

![Alt code](http://i.imgur.com/A2KhZlD.png)

epicelite:
Happens sometimes to me too, dont know why.

penev92:
Thanks for the thorough report!
Also it's nice to see people debugging their own issues. This would save us some time. :)

WolfGaming:
Seems like the return teleport is not finding any valid cells to return the units to.
Are the original cells of the units occupied by other units?

Basti504:
Here the chrono origin of the Gif in the first post.
Not quite blocked it seems, apart from the one tank.

![Alt text](http://i.imgur.com/kj5xfio.gif)

Townes27:
Here's a replay where zero units are chronod back.
OpenRA release-20150614
Map: Squirrels Tournament
Around minute 39:00 Peddaa chronos 5 Mammoth Tanks of LizardLipz into dent's base.
They don't return even though the original cells aren't being occupied.
http://s000.tinyupload.com/index.php?file_id=17891347556467610837

Basti504:
After some more digging around in the source code I found it a nice challenge to track down what happens with that one tank not teleporting back as it is consequently reproducible using the replay of that game. The debugging I did in the posts above were not the actual problem as they were targeted to other actors. I did not think about how to solve this issue as I am not that familiar with the codebase, only track it down.

So here comes… 

1) First to track the tank itself some code was added to see the ActorId on the screen itself, so the tank is identifiable.

2) Using the ActorId from that one tank it was possible to follow the code that is executed for that tank using conditional breakpoints. It is actually quite nice to see the game in a sort of “Step mode” using the conditional breakpoints.

Below the start of chronoshift, the current activity gets canceled and the teleport queued.
[...]

PR #10078, opened 2015-11-22
Add an actor name overlay debug option

penev92:
Inspired by, and very similar to the screenshots on #8670.
Displays the actor type and ID over selected actors when enabled.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/RenderDebugState.cs
+++ b/OpenRA.Mods.Common/Traits/Render/RenderDebugState.cs
@@ -1,0 +1,101 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.AI;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Displays the actor's type and ID above the actor.")]
+	class RenderDebugStateInfo : ITraitInfo
+	{
+		public readonly string Font = "TinyBold";
+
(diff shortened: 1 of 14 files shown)


penev92:
Added current activity and AI squad information.

OpenRA.Mods.Common/Traits/Render/RenderDebugState.cs
```diff
@@ -0,0 +1,100 @@
...
+			{
+				var activityName = activity.GetType().ToString().Split('.').Last();
+				yield return new TextRenderable(font, self.CenterPosition, 0, color, activityName);
+			}
+
+			// Get the AI squad that this actor belongs to.
+			if (!self.Owner.IsBot)
+				yield break;
+
+			var ai = self.Owner.PlayerActor.TraitsImplementing<HackyAI>().FirstOrDefault(x => x.IsEnabled);
```
pchote: Its probably a good idea to cache this.

OpenRA.Mods.Common/Traits/Render/RenderDebugState.cs
```diff
@@ -0,0 +1,100 @@
...
+			if (ai == null)
+				yield break;
+
+			var squads = ai.Squads;
+			var squad = squads.FirstOrDefault(x => x.Units.Contains(self));
```
pchote: I don't think units can change squads (yet...)? If this is true then you could cache this too.

penev92: I'm going to skip on this one because it may or may not break stuff, and the performance gain is minimal - this code is only ran for currently selected AI-controlled actors **if** the debug option is enabled.

Mailaender:
Works as promised :white_check_mark: and an excellent idea. :+1: Good useability as you can control the amount of information by selecting actors.

UberWaffe:
:+1: once caching of `ai` trait is added.
Did a playtest in D2K.
Very useful for debugging.

penev92:
Cached the trait and rebased.

UberWaffe:
:+1:
Ran a retest in RA. Still good.

RoosterDragon:
Neat!

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/10b17db169ff5a14d55a97d9149fb3ffc2b935a0)

PR #9982, opened 2015-11-14
Unhardcode and document things in Chronoshiftable

abcdefg30:

PR #10079, opened 2015-11-22
Fix the description of ReturnToOrigin in Chronoshiftable.cs

abcdefg30:
Fixes what I missed in https://github.com/OpenRA/OpenRA/pull/9982#discussion_r45465710.


GraionDilach:
:+1:

PR #10082, opened 2015-11-22
D2k Deviator balance

OmegaBolt:
This aims to make it more like the original, where it did some decent damage and could attack structures and infantry too. The deviation effect should only work on vehicles.

### DIFF
--- a/mods/d2k/weapons.yaml
+++ b/mods/d2k/weapons.yaml
@@ -455,24 +455,22 @@ DeviatorMissile:
 	ReloadDelay: 160
 	Range: 5c0
 	Report: MISSLE1.WAV
-	InvalidTargets: Infantry, Structure
 	Projectile: Missile
 		MaximumLaunchSpeed: 281
 		RangeLimit: 40
+		HorizontalRateOfTurn: 3
 		Blockable: false
 		Shadow: yes
-		RateofTurn: 1
-		Inaccuracy: 256
+		Inaccuracy: 768
 		Image: MISSILE
 		TrailImage: deviator_trail
 		TrailPalette: deviatorgas
 		TrailUsePlayerPalette: true
 		TrailInterval: 1
 	Warhead@1Dam: SpreadDamage
-		InvalidTargets: Infantry, Structure
 		Spread: 256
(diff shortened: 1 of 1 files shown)


mods/d2k/weapons.yaml
```diff
@@ -455,24 +455,21 @@ DeviatorMissile:
 	ReloadDelay: 160
 	Range: 5c0
 	Report: MISSLE1.WAV
-	InvalidTargets: Infantry, Structure
 	Projectile: Missile
 		MaximumLaunchSpeed: 281
-		RangeLimit: 40
```
reaperrr: Removing this means it might fly infinitely (until it leaves the map) when it loses its target. I agree 40 might be too short, but something in the 50-75 range (=2-3 seconds at normal gamespeed) should be enough to prevent it from exploding too early.

penev92: That explains the missiles I've been seeing fly infinitely. Both from missile tanks and quads, I believe.

pchote: Well, that's fun.  Do we need another issue to track / fix that?

OmegaBolt: Yeah the Missile Tank doesnt have it which is why I removed it. Quads also miss ridiculously often but that's another issue.

reaperrr: > That explains the missiles I've been seeing fly infinitely. Both from missile tanks and quads, I believe.

Quad rockets have a RangeLimit, although it might be too lenient.

reaperrr: Filed #10085 for stable and proposed #10086 (presumably for Next+1).

Mailaender:
This requires more micromanagement and makes the AI a bit weaker in that regard, but that shouldn't stop us here. :+1:

penev92:
Changes seem reasonable :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/8945d3b33467797b9931e64133b93d0e9f7a29cc)

PR #10085, opened 2015-11-22
Give Missile Tank weapon a range limit 

reaperrr:
...and reduce Quad rocket range limit.

This should avoid/reduce missiles flying across the map.

For Next+1 we should consider converting `RangeLimit` from ticks to `WDist`, that would be more intuitive and easier to get right.


abcdefg30:
:+1:

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/7a8005a266efd2404ebf1378c877032b834147e1)

PR #10089, opened 2015-11-23
Try to win a small amount of performance when the shellmap is disabled

### LINKED ISSUE #8668: Game has full cpu utilization in main menu when shellmap is turned off
And music is turned off too, so it's doing nothing but task manager shows full cpu utilization...

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23806889-game-has-full-cpu-utilization-in-main-menu-when-shellmap-is-turned-off?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

abcdefg30:
Fixes #8668.

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -483,7 +483,11 @@ namespace OpenRA
 					Sound.Tick();
 					Sync.CheckSyncUnchanged(world, orderManager.TickImmediate);
 
-					if (world != null)
+					if (world == null)
+						return;
+
+					// Don't tick when the shellmap is disabled
+					if (world.ShouldTick)
 					{
 						var isNetTick = LocalTick % NetTickScale == 0;
 
@@ -506,12 +510,13 @@ namespace OpenRA
 
 							PerfHistory.Tick();
 						}
-						else
-							if (orderManager.NetFrameNumber == 0)
-								orderManager.LastTickTime = RunTime;
+						else if (orderManager.NetFrameNumber == 0)
+							orderManager.LastTickTime = RunTime;
 
 						Sync.CheckSyncUnchanged(world, () => world.TickRender(worldRenderer));
 					}
+					else
+						PerfHistory.Tick();
 				}
 			}
 		}
--- a/OpenRA.Mods.Common/Widgets/VqaPlayerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/VqaPlayerWidget.cs
@@ -37,14 +37,6 @@ namespace OpenRA.Mods.Common.Widgets
 
 		Action onComplete;
 
-		readonly World world;
-
-		[ObjectCreator.UseCtor]
-		public VqaPlayerWidget(World world)
-		{
-			this.world = world;
-		}
-
 		public void Load(string filename)
 		{
 			if (filename == cachedVideo)
--- a/OpenRA.Mods.Common/Widgets/VqaPlayerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/VqaPlayerWidget.cs
@@ -201,7 +201,7 @@ namespace OpenRA.Mods.Common.Widgets
 			Game.Sound.StopVideo();
 			video.Reset();
 			videoSprite.Sheet.GetTexture().SetData(video.FrameData);
-			world.AddFrameEndTask(_ => onComplete());
+			Game.RunAfterTick(onComplete);
 		}
 
 		public void CloseVideo()
--- a/OpenRA.Game/World.cs
+++ b/OpenRA.Game/World.cs
@@ -260,6 +260,7 @@ namespace OpenRA
 		public event Action<Actor> ActorAdded = _ => { };
 		public event Action<Actor> ActorRemoved = _ => { };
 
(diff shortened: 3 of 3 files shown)


Mailaender:
This doesn't seem to have side effects in the settings, asset browser and lobby. However start a mission and the game will freeze after the intro video.

abcdefg30:
Should be fixed now, good find!

pchote:
I think this is a bad idea: we're trading off clean design (by hacking out parts of the code, but not the bits that break!) for some perf that we could win back in other ways.  Can we please fix the cases that rely on `world.Tick` and then cleanly disable everything?

abcdefg30:
Updated. Now disabling the shellmap only allows the perf graph to continue ticking.
Also the VQA-Player doesn't use frame end tasks if `ShouldTick` is `false` anymore.

OpenRA.Game/Game.cs
```diff
@@ -481,7 +481,14 @@ static void InnerLogicTick(OrderManager orderManager)
...
 					Sync.CheckSyncUnchanged(world, orderManager.TickImmediate);
 
-					if (world != null)
+					if (world == null)
+						return;
+
+					// Don't tick when the shellmap is disabled
+					// We want the perf graph to continue though
+					if (!world.ShouldTick)
+						PerfHistory.Tick();
```
pchote: This is ticked inside the `if` and also inside the `else`.  you can move it outside the conditional to avoid the duplication.

OpenRA.Mods.Common/Widgets/VqaPlayerWidget.cs
```diff
@@ -202,7 +202,14 @@ public void Stop()
...
 			videoSprite.Sheet.GetTexture().SetData(video.FrameData);
-			world.AddFrameEndTask(_ => onComplete());
+
+			// If the world doesn't tick,
+			// frame end tasks aren't resolved
+			// So we need to call it directly here
+			if (world.ShouldTick)
+				world.AddFrameEndTask(_ => onComplete());
+			else
+				onComplete();
```
pchote: You could use `Game.RunAfterTick` for a bit of added safety.

abcdefg30:
Updated and resolved the nits.

Mailaender:
Can't spot regressions. :white_check_mark: The CPU performance impact is huge.

obrakmann:
The tick counter now starts looks strange, or at the very least different from bleed. I'm not sure if it's wrong now or if it exposes something funky that was hidden before. That said, I didn't notice any breakage.

On bleed, after you've visited the lobby:
bleed-lobby
this branch, before entering the lobby:
10089_before_lobby
this branch, after entering the lobby:
10089_after_lobby

pchote:
I expect that is caused by the change to the `world == null` check.  Once you join lobby there are now *two* worlds - one associated with the OrderManager, and one associated with the WorldRenderer (i.e. the shellmap).

abcdefg30:
That's because we agreed to move `PerfHistory.Tick()`.

But as it seems, that was a bad idea. Reverting it.

abcdefg30:
Actually scratch that, I still have the `if (world == null) return;`...

abcdefg30:
Well, *actually* scratch that I told you to scratch something.
It seems like it is correct as is at the moment.

pchote:
Removing needs +2, as the first review has been invalidated by the bugs and changes.

OpenRA.Mods.Common/Widgets/VqaPlayerWidget.cs
```diff
@@ -202,7 +202,14 @@ public void Stop()
...
 			videoSprite.Sheet.GetTexture().SetData(video.FrameData);
-			world.AddFrameEndTask(_ => onComplete());
+
+			// If the world doesn't tick,
+			// frame end tasks aren't resolved
+			// So we need to call it directly here
+			if (world.ShouldTick)
+				world.AddFrameEndTask(_ => onComplete());
+			else
+				Game.RunAfterTick(onComplete);
```
RoosterDragon: Why not just call `Game.RunAfterTick` in both cases?

abcdefg30: Seems like this works too.

abcdefg30:
Updated.

pchote:
Appveyor fails.  Maybe try rebasing?

abcdefg30:
Actually that fail came from a rebase I did yesterday...
Anyway, rebased and updated again.

RoosterDragon:
:+1: Looks OK to me and I can't spot any weirdness on the perf graph when enabling/disabling the shellmap and going into and out of lobbies. I can't see any other obvious uses of `AddFrameEndTask` that would need changing either.

Mailaender:
Also can't spot any regressions. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/628789a8521ff4c040e8b25c098d91f1d8a05ecd...c51c61539cc849f6cf7c63d31d1056b51a3e55f4)

PR #10092, opened 2015-11-23
Implement suicide bomb for D2K Saboteur

pchote:
Another aspect of the original d2k balance (see [video](https://youtu.be/dmflM70vtZI?t=232)).
This introduces a yaml option for specifying what should happen when the demolisher enters its target:
- Temporary: Unit leaves again once it is done (existing behaviour).
- Suicide: Unit kills itself, increasing the kill and death counters of the owner (used for saboteur).
- Dispose: Unit is disposed without adjusting the kill/death count (engineer/spy behaviour).

I think it would be useful to take this on to prep since it is part of the overall d2k balance, so I have limited this pr to only change C4Demolition.  I have a [followup branch](https://github.com/pchote/OpenRA/commits/other-enter-traits) for bleed that adds the same flexibility to Infiltrates and the repairing traits.

### DIFF
--- a/OpenRA.Mods.Common/Traits/C4Demolition.cs
+++ b/OpenRA.Mods.Common/Traits/C4Demolition.cs
@@ -35,6 +35,10 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Duration of each flash")]
 		public readonly int FlashDuration = 3;
 
+		[Desc("Behaviour when entering the structure.",
+			"Possible values are Exit, Suicide, Dispose.")]
+		public readonly EnterBehaviour EnterBehaviour = EnterBehaviour.Exit;
+
 		[Desc("Voice string when planting explosive charges.")]
 		[VoiceReference] public readonly string Voice = "Action";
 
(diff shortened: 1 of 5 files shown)


OpenRA.Mods.Common/Traits/C4Demolition.cs
```diff
@@ -23,18 +23,22 @@ class C4DemolitionInfo : ITraitInfo
...
+		[Desc("Interval between each flash.")]
 		public readonly int FlashInterval = 4;
 
-		[Desc("Duration of each flash")]
+		[Desc("Duration of each flash.")]
 		public readonly int FlashDuration = 3;
 
+		[Desc("Behaviour when entering the structure.",
+			"Possible values are Temporary, Suicide, Dispose.")]
+		public readonly EnterBehaviour EnterBehaviour = EnterBehaviour.Temporary;
```
Mailaender: "Temporary" is hard to understand for me without any explanation. How about "Escape", "Evacuate", "Exit", "Leave", "Withdraw" or other synonyms?

Mailaender:
Works as promised. :white_check_mark: They can still bomb vehicles despite the tooltip by the way.

OpenRA.Mods.Common/Traits/C4Demolition.cs
pchote: Thanks, I didn't like this either.  I'll go with `Exit`.

UberWaffe:
:+1: Tested D2K saboteurs. Work as advertised, and I see no problems with the code.

For thoroughness sake, I also tested other uses of the Enter class.
Did so by testing Mutant Hijackers, Engineers and Ghosts in TS. Did not see any unexpected behaviour.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/3136e2cb74254d5660196cfc0603d48193d8a165)

PR #10093, opened 2015-11-23
Made the HarvesterDockSequence class abstract

teees:
Will prevent silly developers like myself from instancing an unfinished base class


atlimit8:
:+1:

UberWaffe:
:+1: Tested, saw no regressions.
Code looks fine as well.

PR #9967, opened 2015-11-11
Made both barracks' doors exit to different cells

cjshmyr:
Small polish, but also prevents a single unit from blocking both doors.

Issue #7663, opened 2015-03-14
kennel placement is blocking dogs from spawning


TheRaffy:
if you aren't careful while you are placing your kennel building you might end up placing it 1 cell north of another building or unwalkable terrain. The result is that your dogs can't get spawned in front of the kennel.

Suggestion: 
- make a 2 celled building, add a 1 cell bib infront the kennel

WolfGaming:
Or make it so the dogs can go out the front or right side of the building.

x-a-n-a-x:
I confirm the bug. So for the moment we can't sing: https://www.youtube.com/watch?v=He82NBjJqf8 (cause dogs are blocked in the kennel).

![screenshot002](https://cloud.githubusercontent.com/assets/1732773/7429740/5f2bde4a-f007-11e4-9ee5-d289817c08e5.png)

Phrohdoh:
This is just a case of the kennel only having a single `Exit`, which is being blocked.

TheRaffy:
Does other buildings have more than one exit?

Phrohdoh:
Most do, but there are a few that were overlooked.

TheRaffy:
Found the structure.yaml:
Okey, If I'd like to add more exits, what values would make sense?

...
KENN:
	.....
	RallyPoint:
		RallyPoint: 0,2
	Exit:
		SpawnOffset: -280,400,0
		ExitCell: 0,1
...
TENT:
	Bib:
	RallyPoint:
	Exit@1:
		SpawnOffset: -42,810,0
		ExitCell: 0,2
	Exit@2:
		SpawnOffset: -725,640,0
		ExitCell: 0,2

GraionDilach:
I don't think another exit would make sense considering that the kennel has only only one visual exit (barracks have two). I'd moreso consider making the cell in front of the kennel to a bib.

TheRaffy:
I don't know maybe the kennel has a second opening in the back :D^^ No, you'r right!

Unit158:
There should be an exit that goes to the side imo. It looks like the dog should be able to get out of the kennel via the front then going sideways.

penev92:
Bottom line is if you want to block it - you can, and there is nothing the game (or developers) can do about it.

TheRaffy:
[...]

PR #10095, opened 2015-11-24
Add second exit to kennel

ghost:
Adds a second exit to the kennel. The AI loves to place these directly above other buildings completely blocking its infantry production.

The kennel exit and shp face south-west so it makes more sense that they can exit west as well as south. From a MP PoV, the logic is similar to the barracks change: #9967 (harder to block dog production with a single vehicle)

### DIFF
--- a/mods/ra/rules/structures.yaml
+++ b/mods/ra/rules/structures.yaml
@@ -1362,9 +1362,12 @@ KENN:
 		HasMinibib: True
 	RallyPoint:
 		Offset: 0,2
-	Exit:
+	Exit@1:
 		SpawnOffset: -280,400,0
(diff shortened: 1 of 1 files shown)


GraionDilach:
I kinda preferred the "let's give it a bib" method from #7663, but that would still allow blocking via other units, so I guess this is also required.

penev92:
Seems harmless enough to go into prep, and will help out the AI a bit.

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/a3aa96c1f043d76b1c28eae78f3447b449bd54a3)

PR #10104, opened 2015-11-24
Fix the D2K building selection decorations

pchote:
tags

This adds new "Primary Building" tag artwork for D2K (the original game repurposed the control group display, which doesn't work for us since we support grouped buildings), and implements the upgraded building decoration that the green star is supposed to be used for.

This is a simple pr that fixes a long standing polish annoyance, so i'd like to see this merged to prep.

The `WithDecoration:SelectionDecoration` option will be used in a future PR to generalize our horrible hardcoded tag and control group rendering.  This also provides our first shipped example of the global upgrades code.


OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
```diff
@@ -87,6 +93,16 @@ public void PlaySingleFrame(int frame)
 
 		public IEnumerable<IRenderable> Render(Actor self, WorldRenderer wr)
 		{
+			return !info.SelectionDecoration ? RenderInner(self, wr, self.Bounds) : Enumerable.Empty<IRenderable>();
+		}
+
+		public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr)
+		{
+			return info.SelectionDecoration ? RenderInner(self, wr, self.VisualBounds) : Enumerable.Empty<IRenderable>();
```
Phrohdoh: This makes me thinkg `SelectionDecoration` is really a `should this be rendered at all`, not `only when selected`.

mods/d2k/rules/structures.yaml
```diff
@@ -84,6 +84,17 @@ construction_yard:
...
+	GlobalUpgradable:
+		Prerequisites: upgrade.conyard
+		Upgrades: stardecoration
+	WithDecoration@upgraded:
+		SelectionDecoration: true
+		Image: pips
+		Sequence: tag-upgraded
+		ReferencePoint: Top, Right
+		ZOffset: 256
+		UpgradeTypes: stardecoration
```
Phrohdoh: Why not `star-decoration`?

Mailaender:
Works as promised. :white_check_mark: Indeed this was just a placeholder. Should be safe.

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/f9719a50d60d2e6580f2dd64fed1823d060f9acf)

PR #10109, opened 2015-11-25
drop target which was created from autotarget if targeted actor now p…

Arcitectus:
…revents autotarget. This is an improvement on the changes made with PR #9854 to resolve issue #6170.

This addresses the scenarios described by @r34ch 4 days ago:

> 1. If a unit is attacking a building and then a friendly engineer starts capturing it, that unit will not stop attacking until ordered to move; with the exception of
> 2. turreted units will continue to attack the building being captured even when moved, including moving out of range and back into range.

### DIFF
--- a/OpenRA.Game/Traits/Target.cs
+++ b/OpenRA.Game/Traits/Target.cs
@@ -16,12 +16,14 @@ using System.Linq;
 namespace OpenRA.Traits
 {
 	public enum TargetType { Invalid, Actor, Terrain, FrozenActor }
+	public enum TargetMethod { Manual, Automatic }
 	public struct Target
 	{
 		public static readonly Target[] None = { };
 		public static readonly Target Invalid = new Target { type = TargetType.Invalid };
 
 		TargetType type;
+		public TargetMethod TargetMethod { get; private set; }
 		Actor actor;
 		FrozenActor frozen;
 		WPos pos;
@@ -40,7 +42,7 @@ namespace OpenRA.Traits
 				: FromCell(w, o.TargetLocation);
 		}
 
-		public static Target FromActor(Actor a)
+		public static Target FromActor(Actor a, TargetMethod targetMethod = TargetMethod.Manual)
 		{
 			if (a == null)
 				return Invalid;
@@ -50,6 +52,7 @@ namespace OpenRA.Traits
 				actor = a,
 				type = TargetType.Actor,
 				generation = a.Generation,
+				TargetMethod = targetMethod,
 			};
 		}
 
@@ -77,7 +80,7 @@ namespace OpenRA.Traits
 			}
 		}
 
-		public bool IsValidFor(Actor targeter)
+		public bool IsValidForIgnoringMethod(Actor targeter)
 		{
 			if (targeter == null || Type == TargetType.Invalid)
 				return false;
--- a/OpenRA.Mods.Common/ActorExts.cs
+++ b/OpenRA.Mods.Common/ActorExts.cs
@@ -145,5 +145,14 @@ namespace OpenRA.Mods.Common
 		{
 			return cells.MinByOrDefault(c => (self.Location - c).LengthSquared);
 		}
+
+		public static bool PreventsAutoTarget(this Actor target, Actor attacker)
+		{
+			foreach (var pat in target.TraitsImplementing<IPreventsAutoTarget>())
+				if (pat.PreventsAutoTarget(target, attacker))
+					return true;
+
+			return false;
+		}
 	}
 }
--- a/OpenRA.Mods.Common/Traits/AutoTarget.cs
+++ b/OpenRA.Mods.Common/Traits/AutoTarget.cs
@@ -164,7 +164,7 @@ namespace OpenRA.Mods.Common.Traits
 		void Attack(Actor self, Actor targetActor, bool allowMove)
 		{
 			TargetedActor = targetActor;
-			var target = Target.FromActor(targetActor);
+			var target = Target.FromActor(targetActor, TargetMethod.Automatic);
 			self.SetTargetLine(target, Color.Red, false);
 			attack.AttackTarget(target, false, allowMove);
 		}
@@ -182,7 +182,7 @@ namespace OpenRA.Mods.Common.Traits
 				if (attackStances == OpenRA.Traits.Stance.Enemy && !actor.AppearsHostileTo(self))
 					continue;
 
-				if (PreventsAutoTarget(self, actor) || !self.Owner.CanTargetActor(actor))
+				if (self.PreventsAutoTarget(actor) || !self.Owner.CanTargetActor(actor))
 					continue;
 
 				// Select only the first compatible armament for each actor: if this actor is selected
@@ -220,14 +220,18 @@ namespace OpenRA.Mods.Common.Traits
 
 			return null;
 		}
+	}
 
-		bool PreventsAutoTarget(Actor attacker, Actor target)
+	public static class TargetExts
+	{
+		public static bool MethodIsValidFor(this Target target, Actor targeter)
 		{
-			foreach (var pat in target.TraitsImplementing<IPreventsAutoTarget>())
-				if (pat.PreventsAutoTarget(target, attacker))
-					return true;
+			return !(target.Actor != null && target.TargetMethod == TargetMethod.Automatic && target.Actor.PreventsAutoTarget(targeter));
+		}
 
-			return false;
+		public static bool IsValidFor(this Target target, Actor targeter)
+		{
+			return target.IsValidForIgnoringMethod(targeter) && target.MethodIsValidFor(targeter);
 		}
 	}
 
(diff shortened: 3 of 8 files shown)


pchote:
Unfortunately we cant really take solutions that break the mod/engine separation.  Can we find another way of doing this that doesnt rely on directly hooking into our core targeting code?

Arcitectus:
I will look into how the mod/engine separation works and which part of the PR would break it.

OpenRA.Game/ActorExts.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA
+{
+	public static class ActorExts
+	{
+		public static bool PreventsAutoTarget(this Actor self, Actor attacker)
+		{
+			return self.TraitsImplementing<IPreventsAutoTarget>().Any(t => t.PreventsAutoTarget(self, attacker));
```
GraionDilach: This breaks it. Try to keep referencing traits in Mods.Common level.

Arcitectus:
I have changed the PR to bring no changes to the 'OpenRA.Game' directory. Is that sufficient to not break the mod/engine separation?
The current version does not anymore address turreted units.

penev92:
I would suggest you take a look at what #10007 did for a sort of a similar situation and see if you can do something like that.
(Off the top of my head) You should be aiming for something like
```
if (target.Type == TargetType.Actor && target.TargetMethod == TargetMethod.AutoTarget && target.Actor.PreventsAutoTarget(a, self))
	return false;
```
which then gives you a much cleaner solution, that even covers all Attack* traits. You just need to ad a `TargetMethod` enum to that, with `Manual` and `AutoTarget` values.

Arcitectus:
penev92: I have looked at pull #10007 and it does not introduce additional state on the target. I am not yet sure if that would classify as directly hooking into the core targeting code.

@pchote: would adding additional state to `OpenRA.Traits.Target` break the mod/engine separation?

pchote:
@penev92's suggestion of adding a `TargetMethod` field is good - this is a general purpose field that will be useful for a few other cases.  I would call the enum values `Automatic` and `Manual` rather than `AutoTarget`, though.

penev92:
@Arcitectus the additional state comes from the `TargetMethod` property that you will add to `Target` ;)
For example that may want to default to `Manual`, to safe you changes all over the place, and then you set it to `Automatic` from inside `AutoTarget.cs`.
The whole thing should be +3 lines in `AttackBase.cs`, +5 lines in whatever file contains the `Target` class and +2-3 lines in `AutoTarget.cs`.

Arcitectus:
Good, I will move the state into the engine. This allows for a simpler implementation.
However, I noticed that a check in `CanAttack` would not suffice as it is not called in all the described cases. An example is the rifle Infantry in RA.

penev92:
I'm not sure I understand that.

Arcitectus:
I kept the added state in the engine and moved the check into Mods.Common.

OpenRA.Game/Traits/Target.cs
```diff
@@ -15,12 +15,14 @@
 namespace OpenRA.Traits
 {
 	public enum TargetType { Invalid, Actor, Terrain, FrozenActor }
-	public struct Target
+	public enum TargetMethod { Manual, Automatic }
+    public struct Target
```
abcdefg30: Use tabs instead of spaces, please.

OpenRA.Mods.Common/ActorExts.cs
```diff
@@ -144,5 +144,13 @@ public static CPos ClosestCell(this Actor self, IEnumerable<CPos> cells)
 		{
 			return cells.MinByOrDefault(c => (self.Location - c).LengthSquared);
 		}
+
+		public static bool PreventsAutoTarget(this Actor self, Actor targeter)
+		{
+			if (self == null)
```
penev92: You have `this Actor self`, so this can't be `null` anyway.

penev92:
At a glance it looks as I expect it to, but it may be a while before I can test it.

OpenRA.Mods.Common/ActorExts.cs
RoosterDragon: It can be null actually, but I wouldn't bother checking for it since if somebody tried it's probably a bug and an exception would be better anyway.

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -160,7 +160,7 @@ public void ScanAndAttack(Actor self, bool allowMove)
 		void Attack(Actor self, Actor targetActor, bool allowMove)
 		{
 			TargetedActor = targetActor;
-			var target = Target.FromActor(targetActor);
+			var target = Target.FromActor(targetActor, TargetMethod.Automatic);
```
penev92: Food for thought for reviewers: this is called both from `ScanAndAttack()` (in which case `Automatic` is perfectly fine) and `Damaged()`, in which case ...
1) If we set it as Automatic, that means the actor may end up in a situation where it doesn't return fire.
2) If we set it as Manual ... well that may lead to confusion and potentially bad handling by future code, because the player did *not* order the attack, it was an automated response.
3) We could introduce a third value for `TargetMethod`, but that may end up being completely redundant (same as `Manual`, just with a more appropriate name for this one case).

abcdefg30:
Needs a rebase.

Arcitectus:
Just rebased this.

There is one more thing that might need attention: Unlike the original implementation of this PR the current change will not anymore result in dropping the target but instead ignoring it (I edited the PR caption to reflect this). Since the attacker keeps the target, it will not pick up others which enter its range.

RoosterDragon:
I feel like we ought to be dropping the target, particularly as #9874 will solve that for targets out of range.

abcdefg30:
Sorry, needs a rebase again. I'm also adding the `fixup needed` label for @RoosterDragon's request.

Arcitectus:
I changed it to drop the target.

abcdefg30:

Except that they still show a red line although they won't start attacking.

RoosterDragon:
Why do we need to maintain a distinction between `target.IsValidFor` and `target.AndMethodIsValidFor`?  I'm not familiar with the targeting mechanics but it seems odd that some areas will now react to an automatically chosen target by dropping it and others would continue anyway. (I am probably just unaware with the motivation for it).

RoosterDragon:
I'm still not convinced that having two different definitions of validity works here. It seems like we need it to be consistent throughout unless somebody can explain to me otherwise.

Also, needs a rebase.

Arcitectus:
Sorry, I rushed it.
I agree it was inconsistent. I replaced the references except the ones located in the engine. I have not yet looked into whether leaving them as is would cause inconsistencies at runtime.

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -220,14 +220,18 @@ Actor ChooseTarget(Actor self, Stance attackStances, WDist range, bool allowMove
...
+		public static bool MethodIsValidFor(this Target target, Actor targeter)
 		{
-			foreach (var pat in target.TraitsImplementing<IPreventsAutoTarget>())
-				if (pat.PreventsAutoTarget(target, attacker))
-					return true;
+			return !(target.Actor != null && target.TargetMethod == TargetMethod.Automatic && target.Actor.PreventsAutoTarget(targeter));
+		}
 
-			return false;
+		public static bool AndMethodIsValidFor(this Target target, Actor targeter)
```
pchote: This name doesn't make sense to me.

Arcitectus: This name derives from the names referred to from the method: It conjugates the results of the method `IsValidFor` and the method `MethodIsValidFor`.
Another option that came to mind is changing the name of the existing method `Target.IsValidFor` to something which signifies that the `TargetMethod` introduced with this change is ignored and then use the name `IsValidFor` for the method you referred to.

Arcitectus:
I addressed the naming issue as described by pchote and consider this PR ready for review.

reaperrr:
Since there have been some changes since the +1, I'm removing the label so it won't get merged prematurely.

@abcdefg30 simply re-add it if you retested this.

abcdefg30:
Looks still good to me. 👍

RoosterDragon:
This design doesn't sit well with me - we're adding baggage to Target (which lives at engine level) with an extra field *it doesn't use* so that AutoTarget (at the mod level) can define the actual way targets are deemed valid or not.

I feel a cleaner design, if we want to extend the behaviour of `Target`, would be to make it a class and make `IsTargetValid` virtual. Then a subclass at mod level can provide the auto-targeting override behaviour without having to having to muck at the base class in the engine.

I'm open to opinions on the matter, though.

pchote:
I agree with @RoosterDragon.

reaperrr:
I'm closing this for now due to lack of progress. @Arcitectus if you decide to pick this up again and adress the above concerns, just reopen it.

Issue #6011, opened 2014-07-21
Power shown on selected power plants.


MustaphaTR:
Like RA2.
![image](https://cloud.githubusercontent.com/assets/7933210/3640614/fc07ebec-10a5-11e4-961f-45dca765bda7.png)
Maybe as
![image](https://cloud.githubusercontent.com/assets/7933210/3640694/75dc1e60-10a7-11e4-9226-4015d6e8b6ac.png)

MustaphaTR:
I think this is no longer needed. We have power as numbers on UI. Only CnC mod has power bar.

Phrohdoh:
This is needed.

sinf:
Clicking a big building is easy and fast. Waiting for tooltips to appear or squinting eyes to see the tiny power meter next to radar isn't as convenient. MustaphaTR's sketch looks alright but I'd rather have a slightly bigger font.
Something like this:
![screenshot](https://cloud.githubusercontent.com/assets/9487531/7437306/1737c9de-f05f-11e4-963e-4c3346c2cf98.png)

PR #10112, opened 2015-11-25
Power tooltip

DnAp:
I added information about using power in tooltip.

I don't know it is good idea or not.
It is my first experience in OpenRa.

### DIFF
--- a/OpenRA.Mods.Common/Traits/PowerTooltip.cs
+++ b/OpenRA.Mods.Common/Traits/PowerTooltip.cs
@@ -1,0 +1,53 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Shown power info on the build palette widget.")]
+	public class PowerTooltipInfo : ITraitInfo
+	{
+		public object Create(ActorInitializer init) { return new PowerTooltip(init.Self); }
+	}
+
+	public class PowerTooltip : IProvideTooltipInfo, INotifyOwnerChanged
+	{
+		readonly Actor self;
+		PowerManager powerManager;
+		DeveloperMode developerMode;
+
+		public PowerTooltip(Actor self)
+		{
+			this.self = self;
+			powerManager = self.Owner.PlayerActor.Trait<PowerManager>();
+			developerMode = self.Owner.PlayerActor.Trait<DeveloperMode>();
+		}
+
+		public bool IsTooltipVisible(Player forPlayer)
+		{
+			return forPlayer == self.Owner;
+		}
+
+		public string TooltipText
+		{
+			get
+			{
+				return "Power Usage: {0}{1}".F(powerManager.PowerDrained, developerMode.UnlimitedPower ? "" : "/" + powerManager.PowerProvided);
+			}
+		}
+
+		public void OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
+		{
+			powerManager = newOwner.PlayerActor.Trait<PowerManager>();
+			developerMode = newOwner.PlayerActor.Trait<DeveloperMode>();
+		}
+	}
+}
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -293,6 +293,7 @@
     <Compile Include="Traits\Burns.cs" />
     <Compile Include="Traits\C4Demolition.cs" />
     <Compile Include="Traits\Health.cs" />
+    <Compile Include="Traits\PowerTooltip.cs" />
     <Compile Include="Traits\VeteranProductionIconOverlay.cs" />
     <Compile Include="Traits\Capturable.cs" />
     <Compile Include="Traits\CaptureNotification.cs" />
--- a/mods/ts/rules/civilian-structures.yaml
+++ b/mods/ts/rules/civilian-structures.yaml
@@ -1092,6 +1092,7 @@ CTDAM:
 		Dimensions: 2, 5
 	Power:
 		Amount: 200
+	PowerTooltip:
 	Armor:
 		Type: heavy
 	Health:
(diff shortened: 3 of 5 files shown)


MustaphaTR:
In fact this is something already wanted as you can se from issue #6011. Not tested your code tho. A screenshot would be good.

OpenRA.Mods.Common/Traits/PowerTooltip.cs
```diff
@@ -0,0 +1,50 @@
...
+	[Desc("Shown power info on the build palette widget.")]
+	public class PowerTooltipInfo : UpgradableTraitInfo
+	{
+		public override object Create(ActorInitializer init) { return new PowerTooltip(init.Self, this); }
+	}
+
+	public class PowerTooltip : IProvideTooltipInfo
+	{
+		readonly Actor self;
+		readonly PowerTooltipInfo info;
```
penev92: This is never used, and that makes the CIS fail the build.

OpenRA.Mods.Common/Traits/PowerTooltip.cs
```diff
@@ -0,0 +1,50 @@
...
+		readonly PowerManager powerManager;
+
+		public PowerTooltip(Actor self, PowerTooltipInfo info)
+		{
+			this.self = self;
+			this.info = info;
+			powerManager = self.Owner.PlayerActor.Trait<PowerManager>();
+		}
+
+
```
penev92: Only one empty line, please.

OpenRA.Mods.Common/Traits/PowerTooltip.cs
```diff
@@ -0,0 +1,50 @@
...
+		public bool IsTooltipVisible(Player forPlayer)
+		{
+			return forPlayer == self.Owner;
+		}
+
+		public string TooltipText
+		{
+			get { return "Power Usage: " + powerManager.PowerDrained.ToString() + (powerManager.PowerProvided != 1000000 ? "/" + powerManager.PowerProvided.ToString() : ""); }
```
penev92: You should use "Power usage: {0}{1}...".F(...)

teees:
Looks like this:
http://gfycat.com/CoarseThornyIbis

Phrohdoh:
I like it, but in RA2 (as reference) each power provider displayed the amount of power they added to the overall power amount.

Currently this is just duplicate data of the already-available overall power usage.

MustaphaTR:
> but in RA2 (as reference) each power provider displayed the amount of power they added to the overall power amount.

No, they doesn't they shows Power/Drain only too.

atlimit8:
If this extra information is attached to power sources, then drain makes more sense than usage since usage implies that it is using rather than producing power which is obviously incorrect.
"Power Drain: 0/400"

This could be useful for mods without power information in the panel. However, I would prefer a "Power: x" where x is the current power output calculated by enabled `Power` traits and `ScalePowerWithHealth`; Although, if one is familiar with the rules, they could guess power output by upgrade overlays and the health bar.

Perhaps, a trait property such as `Total: true` or `TotalDrain: true` could be used for mods without power information in the panel, and the default could be used to tell players a structure's current power output.

obrakmann:
Does this need to be a new trait? I think it would work just as well if you just made `Power` implement `IProvideTooltipInfo`. Modders would then control it using an enum in `PowerInfo` like this:

```yaml
Building:
    Power:
        ShowTooltipInfo: Always, Never, ConsumerOnly or ProducerOnly
```
Alternatively, just a boolean would probably be enough.

What do others think?

atlimit8:
@obrakmann, actually, one actor can have multiple `Power` traits ("turbines"); if `Power` implemented `IProvideTooltipInfo`, then that info could be added multiple times.

obrakmann:
True, hadn't thought of that. Carry on, then.

OpenRA.Mods.Common/Traits/PowerTooltip.cs
```diff
@@ -0,0 +1,49 @@
...
+		}
+
+		public bool IsTooltipVisible(Player forPlayer)
+		{
+			return forPlayer == self.Owner;
+		}
+
+		public string TooltipText
+		{
+			get {
```
abcdefg30: Please put the brace on a new line:
```
get
{
	return ...;
}

**Comment by abcdefg30** [comment]

Could you squash the commits together, please?

**Comment by DnAp** [comment]
Done

File: OpenRA.Mods.Common/Traits/PowerTooltip.cs [hunk 5]
```diff
@@ -0,0 +1,50 @@
...
+		public bool IsTooltipVisible(Player forPlayer)
+		{
+			return forPlayer == self.Owner;
+		}
+
+		public string TooltipText
+		{
+			get
+			{
+				return "Power Usage: {0}{1}".F(powerManager.PowerDrained.ToString(), powerManager.PowerProvided != 1000000 ? "/" + powerManager.PowerProvided.ToString() : "");
```
pchote: These `.ToString()`s shouldn't be needed?

OpenRA.Mods.Common/Traits/PowerTooltip.cs
```diff
@@ -0,0 +1,50 @@
...
+ * see COPYING.
+ */
+#endregion
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Shown power info on the build palette widget.")]
+	public class PowerTooltipInfo : UpgradableTraitInfo
```
atlimit8: This inherits from `UpgradableTraitInfo`, but `PowerTooltip` does not inherit from `UpgradableTrait`. The simple solution is to replace `UpgradableTraitInfo` with `ITraitInfo` and remove `override` from `Create`.

OpenRA.Mods.Common/Traits/PowerTooltip.cs
```diff
@@ -0,0 +1,50 @@
...
+		public bool IsTooltipVisible(Player forPlayer)
+		{
+			return forPlayer == self.Owner;
+		}
+
+		public string TooltipText
+		{
+			get
+			{
+				return "Power Usage: {0}{1}".F(powerManager.PowerDrained, powerManager.PowerProvided != 1000000 ? "/" + powerManager.PowerProvided : "");
```
ScottNZ: I don't understand this `!= 1000000` bit. What are you trying to do?

atlimit8: It is/was a magic value for the unlimited power cheat. It should be replaced with something like `self.Owner.PlayerActor.Trait<DeveloperMode>().UnlimitedPower`. Actually, `self.Owner.PlayerActor.Trait<DeveloperMode>()` should be cached. See OpenRA.Mods.Common/Traits/Power/Player/PowerManager.cs

Mailaender:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c82346ddedc96113571f4954df3c7c7ac2255f17...cb83922a1c24f1a5c68f2cef30bac91f136af3e4)

PR #10117, opened 2015-11-26
Try to speed the frozen actor layer checks up

abcdefg30:
by not creating frozen actors for players that don't need them.

Not sure if it gives us that much, though.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
+++ b/OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
@@ -27,7 +27,7 @@ namespace OpenRA.Mods.Common.Traits
 		public object Create(ActorInitializer init) { return new FrozenUnderFog(init, this); }
 	}
 
-	public class FrozenUnderFog : IRenderModifier, IDefaultVisibility, ITick, ITickRender, ISync, INotifyCreated
+	public class FrozenUnderFog : IRenderModifier, IDefaultVisibility, ITick, ITickRender, ISync, INotifyCreated, INotifyOwnerChanged
 	{
 		[Sync] public int VisibilityHash;
 
@@ -60,16 +60,50 @@ namespace OpenRA.Mods.Common.Traits
 			footprint = footprintCells.SelectMany(c => map.ProjectedCellsCovering(c.ToMPos(map))).ToArray();
 		}
 
-		public void Created(Actor self)
+		void INotifyCreated.Created(Actor self)
 		{
 			frozenStates = new PlayerDictionary<FrozenState>(self.World, (player, playerIndex) =>
 			{
+				if (IsInvalidPlayer(self, player))
+					return null;
+
 				var frozenActor = new FrozenActor(self, footprint, player.Shroud, startsRevealed);
 				if (startsRevealed)
 					UpdateFrozenActor(self, frozenActor, playerIndex);
+
 				player.PlayerActor.Trait<FrozenActorLayer>().Add(frozenActor);
 				return new FrozenState(frozenActor) { IsVisible = startsRevealed };
 			});
+		}
+
+		void INotifyOwnerChanged.OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
+		{
+			for (var i = 0; i < frozenStates.Count; i++)
+			{
+				var player = self.World.Players[i];
+				var state = frozenStates[i];
+				if (state != null)
+				{
+					// If we became an invalid player, remove our frozen actor
+					if (IsInvalidPlayer(self, player))
+					{
+						player.PlayerActor.Trait<FrozenActorLayer>().Remove(state.FrozenActor);
+						state = null;
+					}
+
+					continue;
+				}
+
+				if (IsInvalidPlayer(self, player))
+					continue;
+
+				var frozenActor = new FrozenActor(self, footprint, player.Shroud, startsRevealed);
+				if (startsRevealed)
+					UpdateFrozenActor(self, frozenActor, i);
+
+				player.PlayerActor.Trait<FrozenActorLayer>().Add(frozenActor);
+				state = new FrozenState(frozenActor) { IsVisible = startsRevealed };
+			}
 		}
 
 		void UpdateFrozenActor(Actor self, FrozenActor frozenActor, int playerIndex)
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
```diff
@@ -75,13 +75,23 @@ bool IsVisibleInner(Actor self, Player byPlayer)
...
 
-			var stance = self.Owner.Stances[byPlayer];
-			return info.AlwaysVisibleStances.HasStance(stance) || IsVisibleInner(self, byPlayer);
+			var stance = self.Owner.Stances[player];
+			return info.AlwaysVisibleStances.HasStance(stance);
+		}
+
+		public bool IsVisible(Actor self, Player byPlayer)
+		{
+			return byPlayer == null || IsInvalidPlayer(self, byPlayer) || IsVisibleInner(self, byPlayer);
```
Phrohdoh: This frozen actor should be visible if `byPlayer` is an invalid player?

abcdefg30: I just reused code, before it was
```
if (byPlayer == null)
	return true;
```

RoosterDragon:
This disables frozen actors for all 5 players on the RA shellmap, which is certainly a speedup!

However, won't this cause targeting changes? Any "Invalid" players now effectively sees everything all the time, right? Is that bad?

pchote:
> Any "Invalid" players now effectively sees everything all the time, right? Is that bad?

Potentially, yes.  In theory maps shouldn't use NonCombatant players for anything that impacts combat, but in practice people ignore theory.

abcdefg30:
This is indeed a problem. And I don't have an good idea on how to solve it.

Should I just remove the `player.NonCombatant || !player.Playable` check?

RoosterDragon:
That would reduce our speedup to just removing frozen actors between perma-allies, right? Could still be a decent speedup for larger team games.

pchote:
This is yet another point in favor of removing the broken fragile alliances feature.

pchote:
The context of my comment (since it was discussed / questioned in IRC) is that team members are not perma-allies if that feature is enabled, negating the perf benefits.

RoosterDragon:
Some profiling reveals the stance lookup is almost as expensive as the state lookup - so unfortunately on average the method runs twice as slow since it often must do both lookups :(

This may be a dead-end, sorry.

RoosterDragon:
Thought: Could you move the check inside the `!initialized` branch and adjust the other branch to skip over players which lack a frozen actor (either `TryGetValue` on the dictionary itself, or maybe insert a null state and check for that, whichever makes life easier).

That might allow you to avoid the stance lookup cost on every tick.

abcdefg30:
Updated.

RoosterDragon:
This successfully prevents frozen actors being created for your own stuff - which on the face on it seems pretty good!

Unfortunately, that leads to a crash if an engineer captures your building and that then must become frozen under fog - there's no frozen actor for it and then bad times happen. :/

abcdefg30:
Rebased and updated.

Unfortunately, the newly introduced `PlayerDictionary frozenStates` is readonly, so I can't really create a new `FrozenActor` for the now-enemies (former allies) in case the actor was captured.
If noone has a good idea on how to solve that (without reverting the whole playerindex stuff), I'll close here.

What I could do is just not render when the player is an ally etc.
@RoosterDragon would it be worth the perf to do that?

RoosterDragon:
You could try making the `FrozenActor` field inside `FrozenState` writable and nulling/unnulling it as needed, or maybe when the owner changes you can build a new `PlayerDictionary`, copy things across, and replace the old one.

pchote:
Are the perf wins worth the increased complexity that would require, now that we spatially partition frozen actors updates?

RoosterDragon:
On the RA shellmap on bleed, ticking frozen actors takes 0.85% of total CPU, this PR in its current state brings that down to 0.61% which is pretty respectable.

I think it'd be worthwhile. When the owner changes you need to handle the following:
- Keep existing frozen actors in their current state.
- Add a new frozen actor if there isn't one yet (e.g. our building was captured).
- Remove the frozen actor if there was one but now we don't need it (e.g. we captured a building).

Should be doable.

pchote:
That sounds like a bunch of complexity for a 0.24% (0.02 ms, assuming a 10ms tick) performance improvement.

abcdefg30:
Updated and rebased. Should be fine now that fragile alliances are removed.

Mailaender:
Crashes on Allies 05 when putting the SPY in the TRUK which also removes the CAMERAs:

```
Exception of type `System.ArgumentException`: bounds must be non-empty.
Parameter name: bounds
  at OpenRA.Primitives.SpatiallyPartitioned`1[T].ValidateBounds (Rectangle bounds) [0x0002b] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Primitives/SpatiallyPartitioned.cs:37 
  at OpenRA.Primitives.SpatiallyPartitioned`1[T].Add (OpenRA.Primitives.T item, Rectangle bounds) [0x00003] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Primitives/SpatiallyPartitioned.cs:41 
  at OpenRA.Traits.ScreenMap.Add (OpenRA.Actor a) [0x0000f] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Traits/World/ScreenMap.cs:77 
  at OpenRA.Mods.Common.Traits.Mobile.AddedToWorld (OpenRA.Actor self) <0x41607740 + 0x00063> in <filename unknown>:0 
  at OpenRA.World.Add (OpenRA.Actor a) [0x00041] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/World.cs:245 
  at OpenRA.Mods.Common.Scripting.ActorGlobal+<Create>c__AnonStorey0.<>m__0 (OpenRA.World w) <0x416bacb0 + 0x00023> in <filename unknown>:0 
  at OpenRA.World.Tick () [0x00155] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/World.cs:309 
  at OpenRA.Game.InnerLogicTick (OpenRA.Network.OrderManager orderManager) [0x00226] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:557 
  at OpenRA.Game.LogicTick () [0x00055] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:582 
  at OpenRA.Game.Loop () [0x000d4] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:709 
  at OpenRA.Game.Run () [0x00042] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:749 
  at OpenRA.Program.Run (System.String[] args) [0x00011] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Support/Program.cs:117 
  at OpenRA.Program.Main (System.String[] args) [0x00050] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Support/Program.cs:39 
```

abcdefg30:
I don't think that crash is caused by the pr here: #10766

GraionDilach:
That's an issue.

On the other hand I'm not certain if we should encourage breaking the underlying game just because "hey, we don't have fragile alliances, who cares?".

Mailaender:
Can't spot negative side effects. :white_check_mark:

obrakmann:
Replays will crash when viewing a player whose structure got captured. I'm guessing `frozenStates[byPlayer]` will be null, anything else doesn't make sense.

```
  at OpenRA.Mods.Common.Traits.FrozenUnderFog.IsVisibleInner (OpenRA.Actor self, OpenRA.Player byPlayer) <0x41177c40 + 0x0008c> in <filename unknown>:0 
```

reaperrr:
@abcdefg30 Any progress? Otherwise I'd say we close this for now and you reopen it when you have time to fix and update it.

abcdefg30:
Yeah, sounds like a good idea. Sorry for letting this rot that long.

Issue #9252, opened 2015-09-03
Badger planes interfere with move order


deniz1a:
When a badger plane is clicked with the move cursor, unit moves to badger's location and not to the cell under the cursor. This can be tested with Crate Test map: http://resource.openra.net/maps/4446/
##

penev92:
Reproducible with any aircraft. Clicking on an airborne actor issues a move order for the cell that actor is flying over, not the one that it's visually at (which makes sense, but some might consider it a bug).

matjaeck:
This was fixed.

Issue #8917, opened 2015-08-03
Support high bridges in TS


pchote:
These are trickier than ground-level bridges because they exist in their own movement layer and allow units to go both over and under them (but not drive off the sides).

They share some similarities with tunnels [#6410], but are more complicated because the bridge must be treated as a proper layer of cells instead of being a wormhole connecting two points in the map.

If the pathfinder explicitly tracked the connections between cells then the bridges can insert connections into their own separate cell layer for pathfinding and the ActorMap.  We should be able to get away with a single layer for all high bridges - supporting multiple overlapping high bridges would add considerable extra complexity.

The low-level input code will also need to change to support this: it currently resolves the clicked cell at the Viewport level, which means we either need to teach that about high bridges (not very nice), or we move the click location -> cell resolution somewhere closer to mod code (also not very nice).

pchote:
The basic support for these was added in #12577. There is more work to be done (making them destroyable, allowing move orders onto them) but that PR should be good enough to cover the bounty and prevent bountysource from taking it.

PR #10118, opened 2015-11-26
Rework order targeting crazyness.

### LINKED ISSUE #7546: Unable to force fire on 3/4 cells on a bridge
If i force fire on the one cell that gives me an attack order icon the bridge gets destroyed but the units continue shooting after the bridge is destroyed. If I force fire at any of the other tree cells it will only show a move order icon and will only result in a move order unless there is a unit occupying that cell. 
The working cell:
116f1fa4-bceb-11e4-80f8-c55c6c46a78e
The none working cells:
22e8cce4-bceb-11e4-9d47-c9b897cd4ae2

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8908698-unable-to-force-fire-on-3-4-cells-on-a-bridge?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #10070: Cursors don't work correctly over bridges.
<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28512890-cursors-don-t-work-correctly-over-bridges?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Reduces some duplication in UnitOrderGenerator and fixes #10070.  As a side effect, this adds the other half of the plumbing we need to fix #7546 (#9800 adds the first half), and gets us a small step closer to #8917.
#10070 was introduced by #9327 which gave them a finite selection size.  This meant that the ScreenMap query for actors under the cursor now returned the bridge actor instead of null, and meant that the move cursor and order queries checked against the actor's center position instead of the moused over cell.

This fix comes in two parts:
1. Adding a second pass to `UnitOrderGenerator.OrderForUnit` that checks for orders against the cell if there are no valid orders against the actor.
2. Making move orders invalid for any non-cell targets.

This fixes the cursor, and allows us to again order actors to any cell in the bridge footprint.  It also fixes the inconsistency where you got a move cursor when mousing over trees, but not over barrels.

This may have unexpected consequences, so it shouldn't be merged to prep:
- If there are any cases that relied on move orders against actors being valid when the moused-over cell was invalid but the actor's center position was valid (I don't think we have any code _this_ bogus... although it might affect crushing)
- If there are any other where we want to prevent orders against cells when there is an un-targetable actor on top.

### DIFF
--- a/OpenRA.Game/Orders/UnitOrderGenerator.cs
+++ b/OpenRA.Game/Orders/UnitOrderGenerator.cs
@@ -17,34 +17,41 @@ namespace OpenRA.Orders
 {
 	class UnitOrderGenerator : IOrderGenerator
 	{
-		public IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
+		static Target TargetForInput(World world, CPos xy, MouseInput mi)
 		{
-			var underCursor = world.ScreenMap.ActorsAt(mi)
+			var actor = world.ScreenMap.ActorsAt(mi)
 				.Where(a => !world.FogObscures(a) && a.Info.HasTraitInfo<ITargetableInfo>())
 				.WithHighestSelectionPriority();
 
-			Target target;
-			if (underCursor != null)
-				target = Target.FromActor(underCursor);
-			else
-			{
-				var frozen = world.ScreenMap.FrozenActorsAt(world.RenderPlayer, mi)
-					.Where(a => a.Info.HasTraitInfo<ITargetableInfo>() && a.Visible && a.HasRenderables)
-					.WithHighestSelectionPriority();
-				target = frozen != null ? Target.FromFrozenActor(frozen) : Target.FromCell(world, xy);
-			}
+			if (actor != null)
+				return Target.FromActor(actor);
 
+			var frozen = world.ScreenMap.FrozenActorsAt(world.RenderPlayer, mi)
+				.Where(a => a.Info.HasTraitInfo<ITargetableInfo>() && a.Visible && a.HasRenderables)
+				.WithHighestSelectionPriority();
+
+			if (frozen != null)
+				return Target.FromFrozenActor(frozen);
+
+			return Target.FromCell(world, xy);
+		}
+
+		public IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
+		{
+			var target = TargetForInput(world, xy, mi);
 			var orders = world.Selection.Actors
 				.Select(a => OrderForUnit(a, target, mi))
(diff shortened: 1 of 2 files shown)


pchote:
Also fixes #9252.

OpenRA.Game/Orders/UnitOrderGenerator.cs
```diff
@@ -112,27 +119,30 @@ static UnitOrderResult OrderForUnit(Actor self, Target target, MouseInput mi)
...
-					var modifiers = TargetModifiers.None;
-					if (mi.Modifiers.HasModifier(Modifiers.Ctrl))
-						modifiers |= TargetModifiers.ForceAttack;
-					if (mi.Modifiers.HasModifier(Modifiers.Shift))
-						modifiers |= TargetModifiers.ForceQueue;
-					if (mi.Modifiers.HasModifier(Modifiers.Alt))
-						modifiers |= TargetModifiers.ForceMove;
+			var orders = self.TraitsImplementing<IIssueOrder>()
+				.SelectMany(trait => trait.Orders.Select(x => new { Trait = trait, Order = x }))
+				.OrderByDescending(x => x.Order.OrderPriority);
```
RoosterDragon: This may want a `ToList` to avoid being enumerated twice.

pchote: I'm not convinced: in the common case we only enumerate this once, and in the worst case we enumerate it twice.

RoosterDragon: I can agree with that.

chrisforbes:
:+1: from reading only, not tested.

RoosterDragon:
:+1: Code looks good. Bridges and badgers are indeed fixed and a cursory test reveals crushing is still possible.

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/ca997d14e6d6f60fa0fab5fbbef74c2c272de30f) - `"Improved UI behaviour when issuing orders to units."`

PR #10119, opened 2015-11-26
Interleave renderables for effects and actors

RoosterDragon:
Previously, renderables from actors were z-ordered, but effects were excluded and not ordered. Now we order them all as a whole.

For reference, the RA shellmap uses ~450 actors and ~330 effects at a given moment. Sorting just the actors alone costs 1.4% CPU, whilst sorting both actors and effects raises it to 1.8%. So the performance impact isn't too bad.

According to https://github.com/OpenRA/OpenRA/issues/9443#issue-108459573, this will hopefully fix some polish issues.

### DIFF
--- a/OpenRA.Game/Graphics/WorldRenderer.cs
+++ b/OpenRA.Game/Graphics/WorldRenderer.cs
@@ -94,10 +94,7 @@ namespace OpenRA.Graphics
 
 		List<IFinalizedRenderable> GenerateRenderables()
 		{
-			var actors = World.ScreenMap.ActorsInBox(Viewport.TopLeft, Viewport.BottomRight)
-				.Append(World.WorldActor);
-
-			// Include player actor for the rendered player
+			var actors = World.ScreenMap.ActorsInBox(Viewport.TopLeft, Viewport.BottomRight).Append(World.WorldActor);
 			if (World.RenderPlayer != null)
 				actors = actors.Append(World.RenderPlayer.PlayerActor);
 
@@ -105,19 +102,14 @@ namespace OpenRA.Graphics
 			if (World.OrderGenerator != null)
 				worldRenderables = worldRenderables.Concat(World.OrderGenerator.Render(this, World));
 
+			worldRenderables = worldRenderables.Concat(World.Effects.SelectMany(e => e.Render(this)));
 			worldRenderables = worldRenderables.OrderBy(RenderableScreenZPositionComparisonKey);
 
-			// Effects are drawn on top of all actors
-			// HACK: Effects aren't interleaved with actors.
-			var effectRenderables = World.Effects
-				.SelectMany(e => e.Render(this));
-
 			if (World.OrderGenerator != null)
-				effectRenderables = effectRenderables.Concat(World.OrderGenerator.RenderAfterWorld(this, World));
+				worldRenderables = worldRenderables.Concat(World.OrderGenerator.RenderAfterWorld(this, World));
 
 			Game.Renderer.WorldVoxelRenderer.BeginFrame();
-			var renderables = worldRenderables.Concat(effectRenderables)
-				.Select(r => r.PrepareRender(this)).ToList();
+			var renderables = worldRenderables.Select(r => r.PrepareRender(this)).ToList();
 			Game.Renderer.WorldVoxelRenderer.EndFrame();
 
 			return renderables;
--- a/mods/ts/sequences/misc.yaml
+++ b/mods/ts/sequences/misc.yaml
@@ -47,6 +47,7 @@ crate:
 crate-effects:
 	Defaults:
 		Length: *
+		ZOffset: 2047
 	dollar: money
 	reveal-map: reveal
 	hide-map: shroudx
--- a/mods/ts/sequences/misc.yaml
+++ b/mods/ts/sequences/misc.yaml
@@ -23,6 +23,7 @@ allyrepair:
 	repair: wrench
 		Length: *
 		Tick: 160
+		ZOffset: 2047
 
 rallypoint:
 	flag: mouse
--- a/mods/ts/sequences/misc.yaml
+++ b/mods/ts/sequences/misc.yaml
@@ -266,6 +266,7 @@ wake:
 	idle: wake2
 		Length: *
 		Tick: 180
+		ZOffset: 2047
 
 resources:
 	Defaults:
--- a/mods/ts/sequences/misc.yaml
+++ b/mods/ts/sequences/misc.yaml
@@ -30,15 +30,18 @@ rallypoint:
 	flag: mouse
 		Start: 221
 		Length: 8
+		ZOffset: 3072
 	circles: null
 
 beacon:
 	arrow: mouse
 		Start: 6
 		Offset: 1,-12
+		ZOffset: 2535
 	circles: ring
 		Length: 12
 		BlendMode: Additive
+		ZOffset: 2047
 
 crate:
 	idle:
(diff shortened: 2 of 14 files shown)


OpenRA.Game/Graphics/WorldRenderer.cs
```diff
@@ -98,10 +98,9 @@ public void ReplacePalette(string name, IPalette pal)
 
 		List<IFinalizedRenderable> GenerateRenderables()
 		{
-			var actors = World.ScreenMap.ActorsInBox(Viewport.TopLeft, Viewport.BottomRight)
+			var actors =
+				World.ScreenMap.ActorsInBox(Viewport.TopLeft, Viewport.BottomRight)
```
pchote: Was this formatting change intentional?

OpenRA.Game/Graphics/WorldRenderer.cs
```diff
@@ -109,19 +108,14 @@ List<IFinalizedRenderable> GenerateRenderables()
...
 			worldRenderables = worldRenderables.OrderBy(RenderableScreenZPositionComparisonKey);
 
-			// Effects are drawn on top of all actors
-			// HACK: Effects aren't interleaved with actors.
-			var effectRenderables = World.Effects
-				.SelectMany(e => e.Render(this));
-
 			if (World.OrderGenerator != null)
-				effectRenderables = effectRenderables.Concat(World.OrderGenerator.RenderAfterWorld(this, World));
+				worldRenderables = worldRenderables.Concat(World.OrderGenerator.RenderAfterWorld(this, World));
```
pchote: This is best left to its own PR.  It's already going to take a lot of work to verify the effect ordering, and order generators do weird things with sprites.

pchote:
Code changes look good to me.  We're now going to need to go through all the mods and verify that effects are ordered correctly, inserting `ZOffset` in the sequences where required to fix them.  Order generators are going to be fiddly, so I suggest leaving for a followup PR.

I have created https://github.com/pchote/OpenRA/commits/improve-trails as a testcase for this.

pchote:
For finding and fixing the effect offsets the sensible approach will be to go through the effect implementations class by class, finding where they are used, and then testing/fixing each in turn.  Playing the game normally and hoping that you notice the broken things is a bad idea that will almost guarantee regressions.

pchote:
Note that doing this pr properly is a prerequisite to doing #7520 properly, so this is a big deal for me.

OpenRA.Game/Graphics/WorldRenderer.cs
RoosterDragon: Yes (though I'm not picky if anybody wishes me to revert it)

pchote: Seems a bit redundant, considering that there are several even longer lines in this method.

reaperrr:
I made a list, this should make it a little easier to keep track of what needs changes.

`Game`:
[ ] `AsyncAction`: not checked/tested
[ ] `DelayedAction`: not checked/tested
[ ] `FlashTarget`: not checked/tested

`Mods.Common`:
[x] `Corpse`s are sorted properly in all mods, nothing to do here
[ ] `Bullet`: checked/tested
[ ] `Missile`: not checked/tested, but we should make torpedoes go below ships
[ ] `GravityBomb`: not checked/tested
[ ] `Beacon`: not checked/tested
[ ] `ContrailFader`: not checked/tested
[ ] `CrateEffect`: not checked/tested
[ ] `Explosion`: not checked/tested
[ ] `FloatingText`: not checked/tested
[ ] `LaserZap`: not checked/tested
[ ] `NukeLaunch`: not checked/tested
[ ] `Smoke`: not checked/tested
[ ] `PowerdownIndicator`: not checked/tested
[ ] `RallyPointIndicator`: not checked/tested
[ ] `RepairIndicator`: not checked/tested

[x] `LeavesTrails`: see @pchote's testcase
[ ] `SpawnActorPower`: not checked/tested

`Mods.RA`:
[ ] `GpsDot`: not checked/tested
[ ] `GpsSatellite`: not checked/tested
[ ] `SatelliteLaunch`: not checked/tested
[ ] `TeslaZap`: not checked/tested

`Mods.Cnc`:
[ ] `IonCannon`: not checked/tested


Feel free to contribute

reaperrr:
I updated my branch above and submitted a patch for this PR.

RoosterDragon:
Merged and rebased. I had a few conflicts which I resolved, but do let me know if I didn't merge in those changes 100% correctly.

RoosterDragon:
Updated the list with the following changes. `AsyncAction` and `DelayedAction` have no renderables. `GpsDot`, `GpsSatellite` and `GpsDot` seem fine to me.

OpenRA.Game/Graphics/WorldRenderer.cs
RoosterDragon: Trying everything on one line instead.

reaperrr:
Another bunch of updates (last 5 commits): https://github.com/reaperrr/OpenRA/tree/more-zoffsets1

Takes care of `Beacon`, the three `*Indicator`s and `CrateEffect`.

Note: ZOffsets might need to be increased even more in some cases (high buildings, or structures with overlays), testing this against every structure would have been a pain, so I only tested some presumable worst-cases, no 100% guarantee I didn't overlook anything.

reaperrr:
This one deals with beams (AreaBeam and LaserZap):
https://github.com/reaperrr/OpenRA/tree/more-zoffsets2

And this with the flash when issuing move orders (also covers `SpawnActorPower`, because the only case where that uses a visible effect is the RA sonar, which uses the move flash):
https://github.com/reaperrr/OpenRA/tree/more-zoffsets3

reaperrr:
> [19:20] <reaperrr> pchote: btw, can we make FloatingText use some very high hardcoded value for now, and leave customizability for later?
> [19:20] <+pchote> probably

I suggest changing `0` to `4096` [here.](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Effects/FloatingText.cs#L52)

reaperrr:
I filed one last PR against this branch that should take care of the remaining loose ends (contrails, mostly).

There are cases where it's difficult to find a perfect ZOffset because neither a lower nor a higher value is completely free of sorting issues, for example the end of contrails still sometimes disappears below larger sprites even with a high Z offset. I don't think we can fix everything 100% perfectly.

reaperrr:
Not sure how much it counts since I submitted most of the yaml commits, but :+1: from me once my last 3 submitted commits are merged into this.

pchote:
> There are cases where it's difficult to find a perfect ZOffset because neither a lower nor a higher value is completely free of sorting issues, for example the end of contrails still sometimes disappears below larger sprites even with a high Z offset. I don't think we can fix everything 100% perfectly.

This will have to wait for the per-pixel z-sorting stuff.

RoosterDragon:
Thanks so much for the help reaperrr! All your fixups are now added.

mods/ts/sequences/misc.yaml
```diff
@@ -241,12 +260,14 @@ moveflsh:
...
 		Length: *
 		Tick: 30
+		ZOffset: 2047
 
 wake:
 	idle: wake2
 		Length: *
 		Tick: 180
 
+		ZOffset: 2047
```
abcdefg30: The new line above this doesn't look correct. We probably want a new line below instead.

abcdefg30:
Looks good to me aside this small nit. :+1: / :white_check_mark:

Mailaender:
Very well done. This adds a great amount of visual polish. :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c51c61539cc849f6cf7c63d31d1056b51a3e55f4...a2fa25109d1936b7ddb54c673cae7a5d3978137c)

PR #7734, opened 2015-03-24
Lazily generate frozen actor renderables.

RoosterDragon:
Previously actors that could be frozen under fog but were currently visible would be rendered by the frozen under fog system constantly in order to keep a copy of the renderables ready to go for the frozen counterpart when the actor became invisible. Instead, we now delay this extra rendering until the actor actually becomes invisible. This eliminates the wasted rendering to generate renderables that were never used.

A great example is the RA shellmap, which previously spent 7.0% of the total CPU time generating these renderables just in case. Since the viewer can never see a frozen actor on the shellmap, this was all wasted. With this PR, it now spends 0% of its time doing this.

PR #10120, opened 2015-11-26
Fix broken GPS and frozen actor behaviour

### LINKED ISSUE #10155: Desync on playtest-20151129 related to Turreted
[Here are the sync logs](https://gist.github.com/abcdefg30/17c1084886ef0d6a558e/revisions).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28719879-desync-on-playtest-20151129-related-to-turreted?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RoosterDragon:
## Merge dictionaries in GpsDot

This speeds up GpsDot.Tick as it requires only one dictionary lookup per player rather than two. This change also provides useful infrastructure for later commits.

On the RA shellmap, this reduces total CPU time spent in GpsDot.Tick from 1.3% to 0.6%.
## Ensure frozen actors are rendered on the first tick they become visible

Fixes a regression from #7734. If we don't render the snapshot for the frozen actor straight away, the actor can continue ticking and we'll take a much newer than intended snapshot and display that instead. This doesn't tend to have bad effects in practise but nonetheless we revert back to using TickRender to get a snapshot at the correct time.

We still manage to keep the perf benefit of #7734, by only rendering the snapshot once and not again until we need to.
## Fix GPS dot relying on frozen actor render state

Fixes #10155. See the linked issue for more details. Ultimately I have seperated the simulation and render state of the GPS dot from the frozen actor rendering. I have attempted to maintain other behaviour, such as not rendering a GPS dot if there is already a visible frozen actor for it.
## Fix GPS refreshing for incorrect player

This was just noticed during the debugging for #10155 as clearly broken behaviour (relying on the render player in simulation code). I'm including a fix here, although we haven't seen this manifest any issues of its own as yet.
## Please review carefully

So far, every PR I have ever submitted or reviewed dealing with the frozen actor code has contained major regressions which I did not catch. Anything I've written here has a high chance of being super broken and you should treat it with due scepticism.

### DIFF
--- a/OpenRA.Mods.RA/Effects/GpsDot.cs
+++ b/OpenRA.Mods.RA/Effects/GpsDot.cs
@@ -47,7 +47,8 @@ namespace OpenRA.Mods.RA.Effects
 		class DotState
 		{
 			public readonly GpsWatcher Gps;
-			public bool IsVisible;
+			public bool IsTargetable;
+			public bool ShouldRender;
 			public DotState(GpsWatcher gps)
 			{
 				Gps = gps;
@@ -75,11 +76,13 @@ namespace OpenRA.Mods.RA.Effects
 
 		public bool IsDotVisible(Player toPlayer)
 		{
-			return stateByPlayer[toPlayer].IsVisible;
+			return stateByPlayer[toPlayer].IsTargetable;
 		}
 
-		bool ShouldShowIndicator(Player toPlayer)
+		bool IsTargetableBy(Player toPlayer, out bool shouldRenderIndicator)
 		{
+			shouldRenderIndicator = false;
+
 			if (cloak.Value != null && cloak.Value.Cloaked)
 				return false;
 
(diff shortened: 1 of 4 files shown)


abcdefg30:
For some reason I don't see the GPS-Icons from harvesters (maybe vehicles in general?) anymore.

RoosterDragon:
Yup. I wasn't setting `shouldRender` on the path in `ShouldShowIndicator` where it bails out early with `true` for a unit hidden under fog. Fixed now.

UberWaffe:
Did a quick playtest in RA, and also re-watched the replay from another's perspective.
Didn't see any regressions.

I haven't had time to do a code review, and this probably also needs testing in a network game.

chrisforbes:
:+1: based on reading only, not tested.

obrakmann:
:+1: looks ok.

OpenRA.Mods.RA/Effects/GpsDot.cs
```diff
@@ -60,19 +69,20 @@ public GpsDot(Actor self, GpsDotInfo info)
...
 		}
 
 		public bool IsDotVisible(Player toPlayer)
 		{
-			return showToPlayer[toPlayer];
+			return stateByPlayer[toPlayer].IsVisible;
 		}
 
-		bool ShouldShowIndicator(Player toPlayer)
+		bool ShouldShowIndicator(Player toPlayer, out bool shouldRenderIndicator)
```
pchote: What is the difference between `ShoudShowIndicator` (i.e. the return value) and `shouldRenderIndicator`?

pchote: Looks like this method should be renamed to something like `IsTargetable`.

pchote:
The changes make sense to me, but I don't have time to test this.

OpenRA.Mods.RA/Effects/GpsDot.cs
RoosterDragon: `ShouldShowIndicator` affects the `IsDotVisible` method which ties into `IFogVisibilityModifier` on the parent class. `ShouldShowIndicator` means "Am I visible under the fog for targeting purposes?"

`shouldRenderIndicator` means "Should I actually draw an indicator for this" which is currently "yes if they are targetable under the fog, unless there's already a frozen actor for them".

I could rename this to `IsTargetableBy` and change `IsVisible` on `DotState` to `IsTargetable` or something like that. What do you think?

pchote: Sounds good to me.

RoosterDragon:
Updated naming.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/579c3bed2d11584a6a414cbeef031d7f5bf7ca1c)

Issue #9142, opened 2015-08-23
Poor projectile impact performance


reaperrr:
I noticed arbitrary lag spikes caused by projectiles.

On my notebook with i5-4210M (and presumably some throttling even in 'performance' mode) the `Bullet` effect (sometimes Missile, too, but less pronounced) occasionally eats up to 80 ms on RA's shellmap in performance mode (and much more, sometimes over 300(!)ms in balanced/power-saving mode).

And while far less of an issue there, even on my i5-2500 main rig the projectiles cause occasional 10-20ms spikes.
Their appearance seems to be quite random, so I wonder whether it's only some specific part of the code that causes this.

A while ago I tried to find the exact cause by disabling the ShouldExplode checks and some effects (contrail, trail) one by one, but none of those improved the situation significantly (if at all).

Those hiccups can be quite annoying on slower systems, so it would be nice if someone managed to track down and get rid of the cause.

Edit: It appears it's not the projectile itself, but the stuff happening on impact (warheads).

Mailaender:
I added https://github.com/OpenRA/OpenRA/pull/3741 for desync hunting which from what I learned recently is very expensive. Try removing some `ISync` interfaces and test again to narrow it down.

Mailaender:
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/World.cs#L333 or somehow manage to skip decorational effects.

RoosterDragon:
If the perf numbers are from perf.log, they won't include sync time so it's probably not that.

The recorded time is from Bullet.Tick or Missile.Tick, so we need to look there.

I can't recreate any super-slow Bullet or Missile times on my system on bleed, so I can't suggest much. I've noticed ```Explode``` can sometimes be a little slow but not drastically so.

Maybe try sticking some perf timers inside various bits of the tick methods and see if you can narrow it down?

reaperrr:
> Maybe try sticking some perf timers inside various bits of the tick methods and see if you can narrow it down?

Ok, after some more testing and setting PerfTimers, I can say with certainty that
- the main problem is indeed power-throttling of my notebook
- the projectile effect is not the problem, not even the 'shouldExplode' checks, but rather the actual impact.

So it appears the projectiles perform just fine (better than I expected), but all the stuff happening at impact (DamageWarhead, EffectWarhead etc.) seems to add up (probably when many projectiles explode at the same tick).

RoosterDragon:
[...]

PR #10125, opened 2015-11-27
Updated to current version from Eluant fork

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/9142. See https://github.com/OpenRA/Eluant/pull/1 and https://github.com/OpenRA/OpenRA/pull/9978.


obrakmann:
Missions still work with the new .dll, didn't notice any perf issues, either. :+1:

pchote:
Confirmed that at least some missions still work, and that perf doesn't significantly change :+1:

PR #10126, opened 2015-11-27
Speed up FrozenUnderFog.Tick

RoosterDragon:
Two speedups here:
- Avoid the perf impact of accessing Lazy<T>.Value for the traits.
- Store states for each player in an array rather than a dictionary for much faster lookups. This relies on the players in the world never changing (they don't once the world is created) - however I have applied a small refactoring to firm this up a bit.

On the RA shellmap, this reduces total CPU time spent in FrozenUnderFog.Tick from 1.40% to 0.51%.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
+++ b/OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
@@ -35,11 +35,10 @@ namespace OpenRA.Mods.Common.Traits
 		readonly bool startsRevealed;
 		readonly PPos[] footprint;
 
-		readonly Lazy<ITooltip> tooltip;
-		readonly Lazy<Health> health;
-
 		readonly Dictionary<Player, FrozenState> stateByPlayer = new Dictionary<Player, FrozenState>();
 
+		ITooltip tooltip;
+		Health health;
 		bool initialized;
 
 		class FrozenState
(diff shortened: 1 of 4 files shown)


OpenRA.Mods.Common/Traits/World/EditorActorLayer.cs
```diff
@@ -164,10 +164,7 @@ void SyncMultiplayerCount()
 				var index = int.Parse(name.Substring(5));
 
 				if (index >= newCount)
-				{
 					Players.Players.Remove(name);
-					worldRenderer.World.Players.RemoveAll(pp => pp.InternalName == name);
```
RoosterDragon: @Mailaender: When opening the editor, `worldRenderer.World.Players` is always an empty collection and thus this line was a no-op. So I removed this line as part of my other changes. Just wanted to bring it to your attention in case this indicates a possible bug or anything.

Mailaender: I assume this is a workaround @pchote forgot to remove when rewriting the editor with his editor world concept.

OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
```diff
@@ -90,42 +88,55 @@ public void Tick(Actor self)
...
-					var state = stateByPlayer[player];
+					var state = stateByPlayerIndex[i];
 					frozenActor = state.FrozenActor;
 					isVisible = !frozenActor.Visible;
 					state.IsVisible = isVisible;
 				}
 
 				if (isVisible)
-					VisibilityHash += player.ClientIndex;
+					VisibilityHash |= 1 << (i % 32);
```
pchote: Might this be problematic if we have more than 32 players?

RoosterDragon: It's just a hash. Besides, it's better than adding the raw player indexes which generates far more collisions and leaves the upper bits unused.

Mailaender:

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/66e2c0458fc7f31a79a30303ff530e12774ad238...3075d119b42e906c3376e342ac8dacf4735f6445)

PR #9973, opened 2015-11-12
Allow integer range notation in yaml

phrohdoh:
@teees this should make gate sequences a bit cleaner.

PR #10128, opened 2015-11-28
Upgrade Checks Changes

atlimit8:


penev92:
Nice, I like the idea.

OpenRA.Mods.Common/Lint/LintExts.cs
```diff
@@ -19,17 +21,44 @@ public class LintExts
...
+				var getValues = type.GetProperty("Values").GetGetMethod();
+				type = type.GetGenericArguments()[1];
+				var errorFree = true;
+				Action<string> emitCustomError = message =>
+				{
+					emitError("Bad type for reference values on {0}.{1}. {2}".F(ruleInfo.GetType().Name, fieldInfo.Name, message));
+					errorFree = false;
+				};
+
+				foreach (var v in getValues.Invoke(value, new object[0]) as IEnumerable)
```
RoosterDragon: `null` will suffice instead of `new object[0]`.

OpenRA.Mods.Common/Lint/CheckUpgrades.cs
```diff
@@ -109,20 +109,6 @@ static IEnumerable<string> GetAllGrantedUpgrades(Action<string> emitError, Rules
...
-
-			foreach (var upgrade in gainsExperience)
-				yield return upgrade;
-
-			// TODO: HACK because Pluggable grants upgrades differently to most other sources.
-			var pluggable = rules.Actors.SelectMany(x => x.Value.TraitInfos<PluggableInfo>()
-				.SelectMany(y => y.Upgrades.SelectMany(z => z.Value)));
-
-			foreach (var upgrade in pluggable)
-				yield return upgrade;
```
RoosterDragon: Can you also remove the similar hack in `GetAllUsedUpgrades` now, or does something else need to happen to kill that one off?

penev92: Actually `GainsStatUpgrades` was removed a while back, so that is (supposed to be) dead code.
Will need to be confirmed, of course.

OpenRA.Mods.Common/Traits/GainsExperience.cs
```diff
@@ -19,7 +19,7 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("This actor's experience increases when it has killed a GivesExperience actor.")]
 	public class GainsExperienceInfo : ITraitInfo, Requires<ValuedInfo>, Requires<UpgradeManagerInfo>
 	{
-		[FieldLoader.Require]
+		[FieldLoader.Require, UpgradeGrantedReference, UpgradeUsedReference] // HACK: Treat upgrades as used
```
atlimit8: Added `UpgradeUsedReference` as a hack to silence:
`OpenRA.Utility(1,1): Warning: Actor type `<insert name here>` grants upgrade `eliteweapon` that is not used by anything!`

Another option would be to add an `IgnoreUsedUpgrades` trait that pretends to use upgrades to silence warning about them being unused.

atlimit8:
Changed - ignore lousy AppVeyor failure.

OpenRA.Mods.Common/Traits/GainsExperience.cs
pchote: > Another option would be to add an IgnoreUsedUpgrades trait that pretends to use upgrades to silence warning about them being unused.

Yes, please.  I've been asking for this from the start of this upgrade linting saga.

Phrohdoh:
What does OpenRA gain from this?

pchote:
@Phrohdoh: It appears to remove three `// HACK`s from the code, which is usually a good thing (haven't looked at the content of this pr in detail, so can't comment specifically).

OpenRA.Mods.Common/Traits/GainsExperience.cs
penev92: Why would that be needed? Sounds like it's designed to cover up a bug.

pchote: @penev92 we define upgrades in the actor defaults which are not necessarily implemented by the traits on every actor.

atlimit8:
Merged `UpgradesCache` into `UpgradeManager` (`Upgrades` => `Scriptable`) adding `IgnoredAsUnused` and `IgnoredAsNotGranted`.

Also refactored code for aggregating values  of actor trait properties / warheads with specified `Attribute`. I'm willing to remove this refactoring from this PR and save it for when adding another lint pass.

abcdefg30:
Needs a rebase.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -21,7 +21,19 @@ namespace OpenRA.Mods.Common.Traits
...
 	public class UpgradeManagerInfo : ITraitInfo, Requires<IUpgradableInfo>
 	{
-		public object Create(ActorInitializer init) { return new UpgradeManager(init); }
+		[UpgradeGrantedReference]
+		[Desc("Upgrades that can be granted/revoked from Lua scripts.")]
+		public readonly HashSet<string> Scriptable = new HashSet<string>();
+
+		[UpgradeUsedReference]
+		[Desc("Upgrades not to check whether used.")]
+		public readonly HashSet<string> IgnoredAsUnused = new HashSet<string>();
```
penev92: I don't know whether to call this a hack or a straight-up bug, but it is bad. (I actually still haven't heard a reason to have it)

atlimit8: Maybe a kind of hack, since it is for sticky situations like `eliteweapon` from `GainsExperience` where granted upgrades are inherited but not used and the only other ways to handle it are 1) redefine an entire MiniYaml entry or 2) redefine how `HashSet<string>` can be stored in MiniYaml to allow for removing elements.

Phrohdoh: By definition it cannot be a bug if it has the intended effects.

penev92: So it's "add a nasty hack that goes against the point of the lint check so we don't have to define veterancy upgrades per actor type".
-1 to that...

atlimit8:
Rebased.

atlimit8:
I plan changing the MiniYaml parser to support inline assign list values:
```yaml
a: [1, 7, 2, 3, 4]

a:
    5
    -7
```
=>
```yaml
a:
    1
    2
    3
    4
    5
```
and
a:
    6
    7


a: [1, 2, 3, 4]
```
=>
```yaml
a:
    1
    2
    3
    4
```

---
That should make this possible:
```yaml
^HasEliteWeapon
    GainsExperience:
        Upgrades:
            1600:
                eliteweapon
...
Actor:
    ...
    Inherits@ge0: ^GainsExperience
    Inherits@ge1: ^HasEliteWeapon
    ...
...
```

pchote:
Be careful that you dont break negative numbers in lists

OpenRA.Mods.Common/Lint/LintExts.cs
```diff
@@ -19,17 +22,78 @@ public class LintExts
...
+				return (HashSet<string>)value;
 
-			emitError("Bad type for reference on {0}.{1}. Supported types: string, string[], HashSet<string>"
-				.F(ruleInfo.GetType().Name, fieldInfo.Name));
+			emitError("Supported types: string, string[], HashSet<string>");
+			return Enumerable.Empty<string>();
+		}
 
-			return new string[] { };
+		public static IEnumerable<string> GetAllActorTraitValuesHavingAttribute<T>(Action<string> emitError, Ruleset rules) where T : Attribute
```
Phrohdoh: `Actor` is redundant here IMO as they are the only things to have traits.
But if others want this for clarity that's fine by me.

atlimit8:
Un-escaping strings in the core parser before passing should avoid special character issues.

pchote:
I really don't like the proposed alternative list syntax.  It breaks orthogonality and I can only imagine that it requires some ugly special case logic in the miniyaml parser, which already sufficiently complicated that it has never worked correctly in its entire history.

The implementation in this PR is a *much* cleaner solution, and I do not find @penev92's "I dislike this pragmatic solution, so :-1: to linting these" to be a compelling argument.

pchote:
Following up on the above (at the risk of sidetracking the pr further), I would prefer an inline syntax like `Foo: +[new, elements]` / `Foo: -[remove, existing, elements]` for modifying list entries (inspired by #9973).

atlimit8:
Should `[startNumber..endNumber]` syntax be included then?

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
pchote: @penev92: Why do you prefer "don't lint half our upgrades at all because the tools can't cope"? (i.e. the current situation)

pchote:
`+/-[start..end]` should also be supported if the other pr is merged, yes.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
penev92: You sure that's not the other way around?

pchote: Yes. Right now we aren't linting the upgrades granted by veterancy and probably other traits, which prevents us from being able to trust the lint rule.  I've complained about this before.

penev92: No, right now we lint-check everything. Just in a DYI way for `GainsExperience` and `Pluggable`.
I like the part of this PR about streamlining the checking by adding dictionary support to `LintExts.GetFieldValues()`.
None of that has any relation to the hashset above though, and what it's there to do.

penev92:
Well I was overruled on IRC, so... `¯\_(ツ)_/¯`    Go nuts on this.

*P.S.: Wow I see now that even my -1 was overruled (not just ignored).*

pchote:
You weren't, and it wasn't.

pchote:
This has sat idle for quite some time, so moving to the future milestone for when/if somebody wants to push this through to a merge.

Issue #8673, opened 2015-07-07
Benchmarking


Mailaender:
We already measure CPU and rendering performance and display it in-game when activated under *Settings* - *Advanced*. It is also possible to run replays via double-click or the command line as well as start semi-deterministic games using Lua scripts as seen in our shellmaps.

For @phoronix-test-suite integration we need some kind of machine readable output format for those. Either dump it to the console, into a log text file or raw values into CSV. I don't know what is preferable here.

pchote:
See also http://logs.openra.net/index.php?year=2015&month=07&day=07#03:52:21

Mailaender:
and http://www.phoronix.com/forums/forum/phoronix/phoronix-test-suite/809405-openra-test-suite

Mailaender:
@chrisforbes Is https://github.com/OpenRA/OpenRA/pull/10133 something that would work for Intel as well?

PR #10133, opened 2015-11-28
Added a benchmark mode

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/8673.

If you start the game with ~~`Debug.Benchmark=true`~~  `Launch.Benchmark=true` this adds two new files into the Log folder:

image

which record CPU and rendering performance over time in character separated values with `;` as separator to avoid confusion with localized floating point separators. 

You can create graphs, calculate average FPS like values or maybe even do live data analysis with tools like https://kst-plot.kde.org/ This shouldn't limit anyone's freedom.

The Red Alert desert shellmap is a good performance benchmark in itself, but you probably get the most reproducible results by starting a replay directly either from the command line or by double clicking on it.

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -45,6 +45,8 @@ namespace OpenRA
 		public static Sound Sound;
 		public static bool HasInputFocus = false;
 
+		public static bool BenchmarkMode = false;
+
 		public static GlobalChat GlobalChat;
 
 		public static OrderManager JoinServer(string host, int port, string password, bool recordReplay = true)
@@ -493,6 +495,9 @@ namespace OpenRA
 
 							Log.Write("debug", "--Tick: {0} ({1})", LocalTick, isNetTick ? "net" : "local");
 
+							if (BenchmarkMode)
+								Log.Write("cpu", "{0};{1}".F(LocalTick, PerfHistory.Items["tick_time"].LastValue));
+
 							if (isNetTick)
 								orderManager.Tick();
 
@@ -578,6 +583,9 @@ namespace OpenRA
 			PerfHistory.Items["batches"].Tick();
 			PerfHistory.Items["render_widgets"].Tick();
 			PerfHistory.Items["render_flip"].Tick();
+
+			if (BenchmarkMode)
+				Log.Write("render", "{0};{1}".F(RenderFrame, PerfHistory.Items["render"].LastValue));
 		}
 
 		static void Loop()
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -96,13 +96,25 @@ namespace OpenRA
 	public class GraphicSettings
 	{
 		public string Renderer = "Default";
+
+		[Desc("This can be set to Windowed, Fullscreen or PseudoFullscreen.")]
 		public WindowMode Mode = WindowMode.PseudoFullscreen;
+
+		[Desc("Screen resolution in fullscreen mode.")]
 		public int2 FullscreenSize = new int2(0, 0);
+
+		[Desc("Screen resolution in windowed mode.")]
 		public int2 WindowedSize = new int2(1024, 768);
+
 		public bool HardwareCursors = true;
+
 		public bool PixelDouble = false;
 		public bool CursorDouble = false;
+
+		[Desc("Add a frame rate limiter. It is recommended to not disable this.")]
 		public bool CapFramerate = true;
+
+		[Desc("At which frames per second to cap the framerate.")]
 		public int MaxFramerate = 60;
 
 		public int BatchSize = 8192;
--- a/OpenRA.Game/Support/LaunchArguments.cs
+++ b/OpenRA.Game/Support/LaunchArguments.cs
@@ -21,6 +21,9 @@ namespace OpenRA
 		[Desc("Automatically start playing the given replay file.")]
 		public string Replay;
 
+		[Desc("Dump performance data into cpu.csv and render.csv in the logs folder.")]
+		public bool Benchmark;
+
 		public LaunchArguments(Arguments args)
 		{
 			if (args == null)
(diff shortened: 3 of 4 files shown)


OpenRA.Game/Settings.cs
```diff
@@ -90,18 +90,32 @@ public class DebugSettings
...
 		public WindowMode Mode = WindowMode.PseudoFullscreen;
+
+		[Desc("Screen resolution in fullscreen mode.")]
 		public int2 FullscreenSize = new int2(0, 0);
+
+		[Desc("Screen resolution in windowed mode.")]
 		public int2 WindowedSize = new int2(1024, 768);
+
 		public bool HardwareCursors = true;
 		public bool PixelDouble = false;
```
atlimit8: This can be considered a form of an image quality option.
`[Desc("Use pixel doubling - in world graphics are rendered at 2x.")]`

OpenRA.Game/Settings.cs
```diff
@@ -90,18 +90,32 @@ public class DebugSettings
...
 		public bool CursorDouble = false;
```
atlimit8: If documenting `PixelDouble`, then here:
`[Desc("Use pixel doubling on cursor. Cursors are not scaled by PixelDouble.")]`

OpenRA.Game/Settings.cs
penev92: "in-world" ;)

Mailaender: Zooming doesn't seem to have a performance impact, but documenting it surely doesn't hurt. This is probably a bug. I guess the ScreenMap doesn't adhere to the zooming. Rendering half as many objects should somehow speedup rendering.

pchote: For the number of actors in an average screen you'll be halving a negligible number to make another negligible number.  A lot of the rendering cost is near constant (ui, terrain/resources/shroud, syncing with the gpu, etc) given the small number of vertices we are actually pushing.

Mailaender: I propose https://github.com/OpenRA/OpenRA/issues/10149 for something with more impact. This isn't just interesting for performance comparisons. I saw game critics actively search for such settings to judge how developed an engine is and complain if they are missing.

OpenRA.Game/Settings.cs
```diff
@@ -90,18 +90,36 @@ public class DebugSettings
 		public bool SanityCheckUnsyncedCode = false;
 		public int Samples = 25;
 		public bool IgnoreVersionMismatch = false;
+
+		[Desc("Dump performance data into cpu.csv and render.csv accessible via ~/.openra/Logs.")]
```
RoosterDragon: Could you cover paths for all platforms here?

OpenRA.Game/Settings.cs
```diff
@@ -90,18 +90,36 @@ public class DebugSettings
...
+		public bool Benchmark = false;
```
pchote: I don't think it is appropriate to include this as a settings flag because it doesn't have any UI and it has its own performance impact.  Please implement this as a launch argument instead.

OpenRA.Game/Settings.cs
Mailaender: This will be read in the Linux man page only. I will add a console text which will tell the real location when activated.

OpenRA.Game/Settings.cs
Mailaender: Good idea, I forgot about those. Done.

OpenRA.Game/Settings.cs
Phrohdoh: I'm with @RoosterDragon in that you could say `via your content log directory` or something to that effect.

Mailaender: Already did that as I promised https://github.com/OpenRA/OpenRA/pull/10133/files#diff-be79af2a24e3cc8e6022f6b41958d725R24 if you had checked the diff before posting.

Mailaender: Already did that as I promised https://github.com/OpenRA/OpenRA/pull/10133/files#diff-be79af2a24e3cc8e6022f6b41958d725R24

OpenRA.Game/Game.cs
```diff
@@ -491,6 +493,9 @@ static void InnerLogicTick(OrderManager orderManager)
 
 							Log.Write("debug", "--Tick: {0} ({1})", LocalTick, isNetTick ? "net" : "local");
 
+							if (BenchmarkMode)
+								Log.Write("cpu", "{0};{1}".F(LocalTick, PerfHistory.Items["tick_time"].Average(Settings.Debug.Samples)));
```
RoosterDragon: We'd do better to dump the raw values, rather than averaging it out.

Mailaender: Okay.

abcdefg30:
>  LoadScreens\BlankLoadScreen.cs(50,5): error CS0103: The name 'Console' does not exist in the current context

Mailaender:
Ugh, fixed the error from rebase hell.

x-a-n-a-x:
I have tried **OpenRA.exe Launch.Benchmark=true** but it doesn't create these .csv files in C:\Users\<username>\Documents\OpenRA.

Same report for two other Windows users (Lokke and Leodew) on IRC. 

When I type the command, OpenRA is launched but the cmd.exe prompt just goes to another line, without any error message.

RoosterDragon:
:+1: Works for me.

@x-a-n-a-x: Are you sure you checked out the PR and built everything OK before trying it? I'm not sure why else it wouldn't be working.

x-a-n-a-x:
Oops. You know what? I guess I was tired when I tried, last time. I just compiled bleed and expected this PR to work.. facepalm

I have retried but this time, using Mailaender's branch, and it DOES work. :+1:

abcdefg30:
:+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/73831742b328f3857dc7441b91032a4918b36d8e...349ef44f49580951edbc855c4827104149b11c6d)

PR #10134, opened 2015-11-28
Try to polish the upgrade rules

abcdefg30:
I'd like to get this into prep, as it makes the upgrade from `release-20150919` to the next one easier
(or at least it should).

I'm not offended if this PR is completely ignored in order to focus on more important things, though.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -1679,7 +1679,15 @@ namespace OpenRA.Mods.Common.UtilityCommands
 
 						var otherNodes = nodes;
 						var inherits = new Func<string, bool>(traitName => node.Value.Nodes.Where(n => n.Key.StartsWith("Inherits"))
-								.Any(inh => otherNodes.First(n => n.Key.StartsWith(inh.Value.Value)).Value.Nodes.Any(n => n.Key.StartsWith(traitName))));
+							.Any(inh =>
+							{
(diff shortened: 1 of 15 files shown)


OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2265,6 +2285,28 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 						node.Key = "Aircraft";
 						node.Value.Nodes.Add(new MiniYamlNode("CanHover", "True"));
 					}
+
+					var mplane = node.Value.Nodes.FirstOrDefault(n => n.Key == "-Plane");
+					if (mplane != null)
+					{
+						// Check if a Helicopter trait was renamed to Aircraft
+						// In that case, we don't want to straight negate it with -Aircraft again
```
obrakmann: I'm not sure what problem you're trying to solve here. An actor that simultaneously defines both `Helicopter` and `-Helicopter`? That doesn't make much sense. Also the comment refers to helicopters while it's checking for `Plane` :p. Additionally, this code will only be able to run when `depth` is 0, otherwise `mplane`/`mheli` will be null, so this code will always run before the other two checks (which only run at depth 1), so a renaming cannot have happened at this point.

Why not just add the two rules from the original PR, just with the traits inverted?

obrakmann:
Looks fine :+1:

penev92:
> [20:19] <penev> abcdefg30: ok put a "Warning this will end badly if you inherit Plane, remove it and add Helicopter or vice-versa"

abcdefg30:
Updated.

PR #10141, opened 2015-11-28
Fix actors not going back to idle animation after custom animation

reaperrr:
Before this fix, actors which had an idle animation (not an overlay, but a looping `idle` sequence with more than 1 frame) as well as an `active` animation of some kind would get stuck at the last frame of the active animation once it played, because the latter was basically never cancelled.
A few cases have already been fixed in the past (`WithBuildingPlacedAnimation`, `WithAttackAnimation`), but not all of them.

This has gone unnoticed because there are only very few actors in the shipping mods with more than 1 idle frame (RA airfields are the only case I can think of), while TS uses overlays.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithRepairAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithRepairAnimation.cs
@@ -38,7 +38,7 @@ namespace OpenRA.Mods.Common.Traits
 		{
 			var spriteBody = host.TraitOrDefault<WithSpriteBody>();
 			if (spriteBody != null && !(info.PauseOnLowPower && self.IsDisabled()))
-				spriteBody.PlayCustomAnimation(host, info.Sequence);
+				spriteBody.PlayCustomAnimation(host, info.Sequence, () => spriteBody.CancelCustomAnimation(self));
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Activities/Rearm.cs
+++ b/OpenRA.Mods.Common/Activities/Rearm.cs
@@ -59,7 +59,7 @@ namespace OpenRA.Mods.Common.Activities
 
 				var wsb = hostBuilding.Trait<WithSpriteBody>();
 				if (wsb.DefaultAnimation.HasSequence("active"))
-					wsb.PlayCustomAnimation(hostBuilding, "active");
+					wsb.PlayCustomAnimation(hostBuilding, "active", () => wsb.CancelCustomAnimation(self));
 
 				var sound = pool.Info.RearmSound;
 				if (sound != null)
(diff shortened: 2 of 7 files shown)


Mailaender:
This doesn't seem to regress anything, but I am also not sure what it really fixes. :white_check_mark:

reaperrr:
> This doesn't seem to regress anything, but I am also not sure what it really fixes.

The only noticable case in the shipping mods is the idle animation (that flag thing) of the RA airfield, which stopped after its active animation ran once.
We don't have a `WithIdleAnimation` trait, the sprite body itself loops its `Sequence`. Some traits like `WithActiveAnimation` cancelled that looping, but did not restart it after finishing, getting stuck at the last active frame instead. This means that before this fix, idle animations on tesla coils/obelisks, helipads, service depots, nuke silos etc. would get broken as soon as their special action was triggered and never recover.

This fixes it by always triggering the sprite bodies' default animation after the custom active animation has finished (see `CancelCustomAnimation` in `WithSpriteBody` for reference).
Basically, this applies the same fix we already implemented earlier on some other traits like `WithBuildingPlacedAnimation` and `WithAttackAnimation`.

Phrohdoh:
Oh awesome!
I'll test this in RA2. It is very noticeable there.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3075d119b42e906c3376e342ac8dacf4735f6445...bc448d004c32a1cf8c94986191f4f3249fec738e)

Mailaender:
Looks like we need to revert this. https://github.com/OpenRA/OpenRA/issues/10209

penev92:
Or just fix the issue?
It's a simple one-line fix.
`wsb.PlayCustomAnimation(hostBuilding, "active", () => wsb.CancelCustomAnimation(self));`
to
`wsb.PlayCustomAnimation(hostBuilding, "active", () => wsb.CancelCustomAnimation(hostBuilding));`
in `Rearm`. Yeah, copy-paste is dangerous. ;)

OpenRA.Mods.Common/Traits/Render/WithRepairAnimation.cs
```diff
@@ -38,7 +38,7 @@ public void Repairing(Actor self, Actor host)
 		{
 			var spriteBody = host.TraitOrDefault<WithSpriteBody>();
 			if (spriteBody != null && !(info.PauseOnLowPower && self.IsDisabled()))
-				spriteBody.PlayCustomAnimation(host, info.Sequence);
+				spriteBody.PlayCustomAnimation(host, info.Sequence, () => spriteBody.CancelCustomAnimation(self));
```
abcdefg30: I just noticed: Same as in `Rearm.cs` here.

Issue #10129, opened 2015-11-28
Missiles need some fine adjusting


cjshmyr:
The `Missile` enhancements that have gone in since the past release have been awesome and needed, but at the same time they beefed up missiles in all mods. I'm not proposing any code changes to the missiles, only yaml tweaks to get them back/similar to their release version states. **I'm also thinking TS is fine as-is, and this ticket applies only for RA, TD, and D2K**. We had received negative feedback in RA on this during a playtest, can't speak for other mods yet or if it's a concern there.

I have an automated RA test map here https://github.com/cjshmyr/OpenRA/commit/f941a0ed7f58f4fd184258f02ad38dce6893b5f2 that runs on both bleed and release. It should be useful when comparing the two, also when applying any balance changes back to bleed.

I've observed the following for RA:
- E3 anti-air is better
- Mammoth anti-air is better
- Destroyers all-around are better
- E3 and Chronotank land targeting seems unaffected

Not yet checked SAM sites.

cjshmyr:
Seems that setting `Acceleration: 0` (currently `5`) gets us very close to the release's behavior, since acceleration is a new thing.

Excuse the quality, here's one snippet of the test map ran on three different builds
![dd-fix-comparison](https://cloud.githubusercontent.com/assets/566742/11454030/63eec0c4-95dd-11e5-8663-94238d98c0ee.png)

Also SAM sites seem fine.

pchote:
Looks good to me. Lets set the default in the missile class to 0 (so it applies to all mods) and :ship:it

cjshmyr:
Explaining some of [Missile.cs](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Effects/Missile.cs#L24) for context:

`MaximumLaunchSpeed` - Missiles are launched at this speed. In heightmap enabled mods this varies depending on a height difference, which may use `MinimumLaunchSpeed`.
`MaximumSpeed` - Highest speed the missile can attain. The default is `384` and all missiles currently have this value.
`Acceleration` - Under certain conditions the missile will [change speed](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Effects/Missile.cs#L398) to reach its target. The default acceleration results in changes of `+/-5`.

[...]

PR #10142, opened 2015-11-28
Set default missile acceleration to 0

cjshmyr:
As discussed on IRC and proven out in #10129. If merged, we should make a ticket for `next+1` on the topic of re-adding acceleration, since we're short on time.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -50,7 +50,7 @@ namespace OpenRA.Mods.Common.Effects
 		public readonly WDist MaximumSpeed = new WDist(384);
 
 		[Desc("Projectile acceleration when propulsion activated.")]
-		public readonly WDist Acceleration = new WDist(5);
+		public readonly WDist Acceleration = new WDist(0);
 
 		[Desc("How many ticks before this missile is armed and can explode.")]
 		public readonly int Arm = 0;


GraionDilach:
This is stupid. Disabling `Acceleration` but keeping the `MaximumSpeed` > `MaximumLaunchSpeed` > `MinimumLaunchSpeed` defaults are just bogus and definitely far from elegant.

GraionDilach:
I did a quick glampse in the missile yamls -__ this is really bad__. Most missiles does not override the `LaunchSpeeds` so they will stay varied, which means missiles might end up being launched with a low speed without catching up to their old speed.

Set the _MaximumSpeed_ default to _MaximumLaunchSpeed_ instead of disabling Acceleration for the effect.

penev92:
>This is stupid.

No, this is changing a dice-roll-generated default value to a sensible one. It's not disabling `Acceleration`, it's setting a proper default to avoid side-effects like those that caused the issue to be opened in the first place. From there any weapon can set its own acceleration if it needs it.

But you are right, there are a lot more apparently random generated default values in that class. In any case, this is probably good enough to fix the issue for the release. We can do more fine tuning (and hopefully fixing those ridiculous defaults) and reintroducing acceleration to RA missiles, because it looks pretty good, for the next release.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -50,7 +50,7 @@ public class MissileInfo : IProjectileInfo
 		public readonly WDist MaximumSpeed = new WDist(384);
 
 		[Desc("Projectile acceleration when propulsion activated.")]
-		public readonly WDist Acceleration = new WDist(5);
+		public readonly WDist Acceleration = new WDist(0);
```
pchote: `WVec.Zero` please

cjshmyr:
> I did a quick glampse in the missile yamls -__ this is really bad__. Most missiles does not override the LaunchSpeeds so they will stay varied, which means missiles might end up being launched with a low speed without catching up to their old speed.

Just the TS/RA2/heightmap mods would have an issue with perma-slow missiles due to no acceleration. The minimum speed is only ever used with regards to heightmaps.

Personally I would want all the defaults to be for an accelerated missile, we'd need less duplication in yaml, and that's I think the end goal. Point of this patch was just to be a band-aid for prep (TS is not shipped there), so I'll need to retarget the merge branch with a new PR. We'd be following up on bleed with a more proper fix in the future.

PR #10143, opened 2015-11-29
Implement a utility command to check for explicit interface implementation violations.

phrohdoh:
cc @OpenRA/engine-hackers

`[PR: Help wanted]` is because I don't know how to add this as a required-passing step to Travis.

It would be of great help if all PRs now and forthcoming implemented their interfaces explicitly, at least until the above-noted Travis step is completed and we can check for it in each PR.

I could add this to `make check` but that is already incredibly slow.
Would that be the proper place to run this?

Currently this produces:

``` bash
tmba:openra thill\ $ orautil all --check-explicit-interfaces

Actor.OnScriptBind (from IScriptNotifyBind) is not explicit!
AsyncAction.Tick (from IEffect) is not explicit!
AsyncAction.Render (from IEffect) is not explicit!
DelayedAction.Tick (from IEffect) is not explicit!
DelayedAction.Render (from IEffect) is not explicit!
FlashTarget.Tick (from IEffect) is not explicit!
FlashTarget.Render (from IEffect) is not explicit!
SpriteEffect.Tick (from IEffect) is not explicit!
SpriteEffect.Render (from IEffect) is not explicit!
BagFile.GetContent (from IFolder) is not explicit!
BagFile.Exists (from IFolder) is not explicit!
BagFile.Write (from IFolder) is not explicit!
BagFile.Priority (from IFolder) is not explicit!
BagFile.Name (from IFolder) is not explicit!
[...snip...]
WithPermanentInjury.SequencePrefix (from IRenderInfantrySequenceModifier) is not explicit!
WithPermanentInjury.Damaged (from INotifyDamage) is not explicit!
WithVoxelUnloadBody.SelectionSize (from IAutoSelectionSize) is not explicit!
WithVoxelWalkerBody.Tick (from ITick) is not explicit!
WithVoxelWalkerBody.SelectionSize (from IAutoSelectionSize) is not explicit!
WithVoxelWaterBody.SelectionSize (from IAutoSelectionSize) is not explicit!
TSShroudPalette.PaletteNames (from IProvidesAssetBrowserPalettes) is not explicit!
TSShroudPalette.LoadPalettes (from ILoadsPalettes) is not explicit!
ImportLegacyTilesetCommand.ValidateArguments (from IUtilityCommand) is not explicit!
ImportLegacyTilesetCommand.Run (from IUtilityCommand) is not explicit!
ImportLegacyTilesetCommand.Name (from IUtilityCommand) is not explicit!

Explicit interface violations: 1983
```

To be less ambiguous I could add method params to the output but this gets ugly when generic types/params are introduced.

**Edit**:
This _still_ doesn't actually work:

``` bash
tmba:openra thill\ $ orautil all --check-explicit-interfaces | head

BigFile.GetContent (from IFolder) is not explicit!
BigFile.Exists (from IFolder) is not explicit!
BigFile.Write (from IFolder) is not explicit!
BigFile.Priority (from IFolder) is not explicit!
BigFile.Name (from IFolder) is not explicit!
D2kSoundResources.GetContent (from IFolder) is not explicit!
D2kSoundResources.Exists (from IFolder) is not explicit!
D2kSoundResources.Write (from IFolder) is not explicit!
D2kSoundResources.Priority (from IFolder) is not explicit!
D2kSoundResources.Name (from IFolder) is not explicit!

tmba:openra thill\ $ make
make: Nothing to be done for 'default'.
```

``` diff
@@ -135,13 +135,13 @@ public Stream GetContent(uint hash)
                        return null;
                }

-               public Stream GetContent(string filename)
+               Stream IFolder.GetContent(string filename)
                {
                        var hash = FindMatchingHash(filename);
                        return hash.HasValue ? GetContent(hash.Value) : null;
                }

-               public bool Exists(string filename)
+               bool IFolder.Exists(string filename)
                {
                        return FindMatchingHash(filename).HasValue;
                }
```

``` csharp
    public interface IFolder : IDisposable
    {
        Stream GetContent(string filename);
        bool Exists(string filename);
        IEnumerable<uint> ClassicHashes();
        IEnumerable<uint> CrcHashes();
        IEnumerable<string> AllFileNames();
        void Write(Dictionary<string, byte[]> contents);
        int Priority { get; }
        string Name { get; }
    }
```

I don't know why the code still sees IFolder as implicit in the above example/diff.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
@@ -1,0 +1,122 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+using System.Reflection;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.UtilityCommands
+{
+	public class CheckExplicitInterfacesCommand : IUtilityCommand
+	{
+		string IUtilityCommand.Name { get { return "--check-explicit-interfaces"; } }
+
+		bool IUtilityCommand.ValidateArguments(string[] args)
+		{
+			return args.Length == 1;
+		}
+
+		int violationCount;
+
+		[Desc("Check for explicit interface implementation violations in all assemblies referenced by the specified mod.")]
+		void IUtilityCommand.Run(ModData modData, string[] args)
+		{
+			var types = modData.ObjectCreator.GetTypes();
+
+			foreach (var implementingType in types.Where(t => !t.IsInterface))
+			{
+				if (implementingType.IsEnum)
+					continue;
+
+				var interfaces = implementingType.GetInterfaces();
+				foreach (var interfaceType in interfaces)
+				{
+					if (!interfaceType.HasAttribute<RequireExplicitImplementationAttribute>())
+						continue;
+
+					var interfaceMembers = interfaceType.GetMembers();
+					foreach (var interfaceMember in interfaceMembers)
+					{
+						if (interfaceMember.Name.StartsWith("get_") || interfaceMember.Name.StartsWith("set_"))
+							continue;
+
+						var interfaceMethod = interfaceMember as MethodInfo;
+						if (interfaceMethod != null)
+						{
+							var interfaceMethodParams = interfaceMethod.GetParameters();
+							foreach (var implementingMethod in implementingType.GetMethods())
+							{
+								if (implementingMethod.Name != interfaceMethod.Name
+									|| implementingMethod.ReturnType != interfaceMethod.ReturnType)
+									continue;
+
+								var implementingMethodParams = implementingMethod.GetParameters();
+								var lenImpl = implementingMethodParams.Length;
+								if (lenImpl != interfaceMethodParams.Length)
+									continue;
+
+								var allMatch = true;
+								for (var i = 0; i < lenImpl; i++)
+								{
+								    var implementingParam = implementingMethodParams[i];
+								    var interfaceParam = interfaceMethodParams[i];
+								    if (implementingParam.ParameterType != interfaceParam.ParameterType
+										|| implementingParam.Name != interfaceParam.Name
+										|| implementingParam.IsOut != interfaceParam.IsOut)
+								    {
+										allMatch = false;
+										break;
+								    }
+								}
+
+								// Explicitly implemented methods are never public in C#.
+								if (allMatch && implementingMethod.IsPublic)
+									OnViolation(implementingType, interfaceType, implementingMethod);
+							}
+						}
+
+						var interfaceProperty = interfaceMember as PropertyInfo;
+						if (interfaceProperty != null)
+						{
+							var implementingProperties = implementingType.GetProperties();
+							foreach (var implementingProperty in implementingProperties)
+							{
+								if (implementingProperty.PropertyType != interfaceProperty.PropertyType
+									|| implementingProperty.Name != interfaceProperty.Name)
+									continue;
+
+								if (!IsExplicitInterfaceProperty(implementingProperty))
+									OnViolation(implementingType, interfaceType, implementingProperty);
+							}
+						}
+					}
+				}
+			}
+
+			if (violationCount > 0)
+			{
+				Console.WriteLine("Explicit interface violations: {0}", violationCount);
+				Environment.Exit(1);
+			}
+		}
+
+		static bool IsExplicitInterfaceProperty(PropertyInfo pi)
+		{
+			return pi.Name.Contains(".");
+		}
+
+		void OnViolation(Type implementor, Type interfaceType, MemberInfo violator)
+		{
+			Console.WriteLine("{0} must explicitly implement the interface member {1}.{2}", implementor.Name, interfaceType.Name, violator.Name);
+			violationCount++;
+		}
+	}
+}
--- a/Makefile
+++ b/Makefile
@@ -227,6 +227,9 @@ check: utility mods
 	@echo
 	@echo "Checking for code style violations in OpenRA.Test..."
 	@mono --debug OpenRA.Utility.exe ra --check-code-style OpenRA.Test
+	@echo
+	@echo "Checking for explicit interface violations..."
+	@mono --debug OpenRA.Utility.exe all --check-explicit-interfaces
 
 NUNIT_CONSOLE := $(shell test -f thirdparty/download/nunit3-console.exe && echo mono thirdparty/download/nunit3-console.exe || \
 	which nunit3-console 2>/dev/null || which nunit2-console 2>/dev/null || which nunit-console 2>/dev/null)
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -19,6 +19,8 @@ using OpenRA.Primitives;
 
 namespace OpenRA.Traits
 {
+	public sealed class RequireExplicitImplementationAttribute : Attribute { }
+
 	public enum DamageState { Undamaged, Light, Medium, Heavy, Critical, Dead }
 
 	public interface IHealth
@@ -249,6 +251,8 @@ namespace OpenRA.Traits
 	public interface ILoadsPlayerPalettes { void LoadPlayerPalettes(WorldRenderer wr, string playerName, HSLColor playerColor, bool replaceExisting); }
 	public interface IPaletteModifier { void AdjustPalette(IReadOnlyDictionary<string, MutablePalette> b); }
 	public interface IPips { IEnumerable<PipType> GetPips(Actor self); }
+
+	[RequireExplicitImplementation]
 	public interface ISelectionBar { float GetValue(); Color GetColor(); }
 
 	public interface IPositionableInfo : ITraitInfoInterface { }
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -726,6 +726,7 @@
     <Compile Include="FileFormats\RLEZerosCompression.cs" />
     <Compile Include="FileFormats\IniFile.cs" />
     <Compile Include="Orders\GuardOrderGenerator.cs" />
+    <Compile Include="UtilityCommands\CheckExplicitInterfacesCommand.cs" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
--- a/OpenRA.Mods.Common/Traits/Buildings/BaseProvider.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/BaseProvider.cs
@@ -78,8 +78,7 @@ namespace OpenRA.Mods.Common.Traits
 				Color.FromArgb(96, Color.Black));
 		}
 
-		// Selection bar
-		public float GetValue()
+		float ISelectionBar.GetValue()
 		{
 			// Visible to player and allies
 			if (!ValidRenderPlayer())
@@ -92,6 +91,6 @@ namespace OpenRA.Mods.Common.Traits
 			return (float)progress / total;
 		}
 
-		public Color GetColor() { return Color.Purple; }
+		Color ISelectionBar.GetColor() { return Color.Purple; }
 	}
 }
--- a/OpenRA.Mods.Common/Traits/ExternalCapturableBar.cs
+++ b/OpenRA.Mods.Common/Traits/ExternalCapturableBar.cs
@@ -28,7 +28,7 @@ namespace OpenRA.Mods.Common.Traits
 			capturable = self.Trait<ExternalCapturable>();
 		}
 
-		public float GetValue()
+		float ISelectionBar.GetValue()
 		{
 			// only show when building is being captured
 			if (!capturable.CaptureInProgress)
@@ -37,6 +37,6 @@ namespace OpenRA.Mods.Common.Traits
 			return (float)capturable.CaptureProgressTime / (capturable.Info.CaptureCompleteTime * 25);
 		}
 
-		public Color GetColor() { return Color.Orange; }
+		Color ISelectionBar.GetColor() { return Color.Orange; }
 	}
 }
--- a/OpenRA.Mods.Common/Traits/Power/AffectedByPowerOutage.cs
+++ b/OpenRA.Mods.Common/Traits/Power/AffectedByPowerOutage.cs
@@ -28,7 +28,7 @@ namespace OpenRA.Mods.Common.Traits
 			playerPower = self.Owner.PlayerActor.Trait<PowerManager>();
 		}
 
-		public float GetValue()
+		float ISelectionBar.GetValue()
 		{
 			if (playerPower.PowerOutageRemainingTicks <= 0)
 				return 0;
@@ -36,7 +36,7 @@ namespace OpenRA.Mods.Common.Traits
 			return (float)playerPower.PowerOutageRemainingTicks / playerPower.PowerOutageTotalTicks;
 		}
 
-		public Color GetColor()
+		Color ISelectionBar.GetColor()
 		{
 			return Color.Yellow;
 		}
--- a/OpenRA.Mods.Common/Traits/Render/ProductionBar.cs
+++ b/OpenRA.Mods.Common/Traits/Render/ProductionBar.cs
@@ -70,7 +70,7 @@ namespace OpenRA.Mods.Common.Traits
 			value = current != null ? 1 - (float)current.RemainingCost / current.TotalCost : 0;
 		}
 
-		public float GetValue()
+		float ISelectionBar.GetValue()
 		{
 			// only people we like should see our production status.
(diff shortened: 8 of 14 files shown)


OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
```diff
@@ -0,0 +1,121 @@
...
+
+									allMatch = true;
+								}
+
+								if (allMatch && implementingMethod.IsPublic)
+									OnViolation(implementingType, implementingMethod, interfaceType);
+							}
+
+							// This member is not a property so we can skip the following instructions.
+							continue;
```
Phrohdoh: This only skips the cast, should it be removed?

pchote:
I think the best approach, to start with at least, is to keep a Type[] array whitelisting the specific interfaces we want to check. We can convert them over a few prs, and then after the dust settles change it to a blacklist of the few interfaces we dont want to enforce it on.

Phrohdoh:
I did not take that approach because:
1) White-listed interfaces must be accessible to this namespace at compilation time.
2) Who decides what should be white-listed? Are we going to bicker back and forth about which interfaces we should check and eventually get nowhere with this PR?

But I can agree that the proposed approach would move things along more smoothly.

I had the idea of using attributes for this instead of a `Type[]` such as `[RequireExplicitImplementation]`, but this may not be the best way forward.

Either way the logic here doesn't work properly and I do not know why.

pchote:
If you can make the attribute work, then go for that.  In any case we can't take this if it causes the tests to fail, and changing all the interfaces in this pr would be way out of scope, so you need *some* way to define what interfaces should be checked.

penev92:
It would probably be quite a while before I can take a look at this, but I'm all for the idea, so don't give up on it.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
```diff
@@ -0,0 +1,118 @@
...
+		{
+			var types = AppDomain.CurrentDomain.GetAssemblies()
+				.Where(asm => asm.FullName.Contains("OpenRA"))
+				.SelectMany(asm => asm.GetTypes());
+
+			foreach (var implementingType in types.Where(t => !t.IsInterface))
+			{
+				if (implementingType.IsGenericType ||
+					implementingType.IsEnum ||
+					implementingType.Name.EndsWith("Info"))
```
RoosterDragon: Why are we skipping:
- Generic types?
- The "Info" types?

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
```diff
@@ -0,0 +1,118 @@
...
+				foreach (var interfaceType in interfaces)
+				{
+					if (!interfaceType.Namespace.Contains("OpenRA"))
+						continue;
+
+					var interfaceMembers = interfaceType.GetMembers();
+					foreach (var interfaceMember in interfaceMembers)
+					{
+						if (interfaceMember.Name.Contains("_"))
+							continue;
```
RoosterDragon: Why are we skipping stuff with underscores in?

RoosterDragon:
Consider using [`GetInterfaceMap`](https://msdn.microsoft.com/en-us/library/system.type.getinterfacemap.aspx) to make converting between the interface definitions and the implementing methods easier and more robust.

I'm also unsure of a portable way to check for explicitly implemented methods across different languages. For example VB makes all interface methods explicit and public and you don't get a choice in the matter.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
Phrohdoh: We don't care about `get_` and `set_`.

Phrohdoh:
We don't care about VB at this time, nor do I think we'll ever.
If the time comes this can be reworked then.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
Phrohdoh: I don't remember the exact reasons, but I know I had some issues finding the correct types for these two cases.

Phrohdoh: Thinking about this a bit more we definitely don't want to skip `*Info` classes.
I may have skipped generics because I don't think it is worth my (or anyone else's) time to check N generic arguments that may be arranged in M ways.

But if this sits long enough I may get bored and do it.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
Phrohdoh: In fact, they cannot be included because they will always be shown as not explicit.

RoosterDragon: Cool, can we get a comment explaining that?

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
RoosterDragon: I'm a bit lost on the generics comment. If we're going to check `Trait : IFoo` why shouldn't we check `Trait<T>: IFoo`? Where's the extra work?

Phrohdoh: You have to remember I have to print these in a friendly way as well as checking the implementations.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
Phrohdoh: Sure, I don't see this being merged though.

RoosterDragon:
Well TIL `GetInterfaceMap` has some serious shortcomings. Pretty lame.

I'm not against this for trait interfaces for reasons that have been pointed out before - but I'm not a fan of blindly applying this to all our interfaces. Making the useful methods on objects unavailable unless you happen to remember to cast things seems counterproductive - moreso if you're not aware that's even an option.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
RoosterDragon: The default name of ```Trait `1``` will probably be fine, I think.

Phrohdoh: It is more about the generic return type and arguments, not so much the implementing type.
I'll try to get to this sometime over the coming break.

penev92:
Could not reproduce the bug you mention in the PR description.
Adding some 'IFolder.` to the methods inside `BigFile` produced this:
>ZipFile.Name (from IFolder) is not explicit!
BigFile.Write (from IFolder) is not explicit!
BigFile.Priority (from IFolder) is not explicit!
BigFile.Name (from IFolder) is not explicit!
TraitContainer`1.Add (from ITraitContainer) is not explicit!

The rest of them are apparently explicit now.

Phrohdoh:
At this point I'd very much like this to not be merged.
The code is not something I'm proud of and if we are updating to 4.5 I could write an analyzer for this that would be much more accurate.

If you all still want this implementation now instead of something else later I'll clean up my local commits and push because I now have a working `RequiresExplicitImplementationAttribute`.

Let me know.

pchote:
Attribute and clean up please.  Doing something new based on 4.5 is going to have its own set of issues to block it, so let's go with the solution which works now and we can potentially replace it later.

Phrohdoh:
This has been updated.

Phrohdoh:
Updated.
With the test commit you should get:

```
tmba:openra thill $ orautil all --check-explicit-interfaces

Aircraft.ROT (from IFacing) is not explicit!
Aircraft.Facing (from IFacing) is not explicit!
Husk.ROT (from IFacing) is not explicit!
Husk.Facing (from IFacing) is not explicit!
Mobile.ROT (from IFacing) is not explicit!
Mobile.Facing (from IFacing) is not explicit!
Explicit interface violations: 6
```

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
```diff
@@ -0,0 +1,120 @@
...
+		public string Name { get { return "--check-explicit-interfaces"; } }
+
+		bool IUtilityCommand.ValidateArguments(string[] args) { return true; }
+
+		int violationCount;
+
+		[Desc("Check for explicit interface implementation violations in all OpenRA assemblies referenced by the specified mod.")]
+		void IUtilityCommand.Run(ModData modData, string[] args)
+		{
+			var types = AppDomain.CurrentDomain.GetAssemblies()
```
pchote: `modData.ObjectCreator.GetTypes()` please.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
```diff
@@ -0,0 +1,120 @@
...
+				if (implementingType.IsEnum)
+					continue;
+
+				var interfaces = implementingType.GetInterfaces();
+				foreach (var interfaceType in interfaces)
+				{
+					if (!interfaceType.HasAttribute<RequireExplicitImplementationAttribute>())
+						continue;
+
+					if (!interfaceType.Namespace.Contains("OpenRA"))
```
pchote: Types from other namespaces won't have `RequireExplicitImplementationAttribute`, so you can probably drop this.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
```diff
@@ -0,0 +1,120 @@
...
+						continue;
+
+					var interfaceMembers = interfaceType.GetMembers();
+					foreach (var interfaceMember in interfaceMembers)
+					{
+						// Ignore generated get_ and set_ methods.
+						if (interfaceMember.Name.Contains("_"))
```
pchote: Will this also silently filter *everything* from interfaces which are (poorly) named with underscores?
Can you explicitly check against `.get_` and `.set_` instead?

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
Phrohdoh: Good point.
This existed because before the attrib was implemented Eluant wasn't playing nice.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
Phrohdoh: Yes it will, and yes I can.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
Phrohdoh: Does this make the output depend on the mod that was passed?
If so, is that desired?

pchote:
Works as advertised. Awesome!

:+1: after addressing the three fixups above and removing the test commit.

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
pchote: It already does, this just makes it explicit and consistent with the rest of our code.

Phrohdoh: Ok, will do.

pchote:
Found a problem with a bit more testing: this isn't picking up the implicit implementation of `ISelectionBar` by `OpenRA.OpenRA.Mods.RA.Traits.Chronoshiftable` (and probably other traits).

OpenRA.Mods.Common/UtilityCommands/CheckExplicitInterfacesCommand.cs
```diff
@@ -0,0 +1,120 @@
...
+							{
+								if (implementingMethod.Name != interfaceMethod.Name)
+									continue;
+
+								var implementingMethodParams = implementingMethod.GetParameters();
+								var lenImpl = implementingMethodParams.Length;
+								if (lenImpl != interfaceMethodParams.Length)
+									continue;
+
+								var allMatch = false;
```
pchote: This needs to be as follows:

```csharp
								var allMatch = true;
								for (var i = 0; i < lenImpl; i++)
								{
									var parameter = implementingMethodParams[i];
									if (parameter.ParameterType != interfaceMethodParams[i].ParameterType
										|| parameter.Name != interfaceMethodParams[i].Name)
									{
										allMatch = false;
										break;
									}
								}
```

Phrohdoh: Thanks!
Will swap these out and update the labels.

Phrohdoh:
This has been fixed.

```
tmba:openra thill $ orautil all --check-explicit-interfaces
Aircraft.ROT (from IFacing) is not explicit!
Aircraft.Facing (from IFacing) is not explicit!
BaseProvider.GetValue (from ISelectionBar) is not explicit!
BaseProvider.GetColor (from ISelectionBar) is not explicit!
ExternalCapturableBar.GetValue (from ISelectionBar) is not explicit!
ExternalCapturableBar.GetColor (from ISelectionBar) is not explicit!
Husk.ROT (from IFacing) is not explicit!
Husk.Facing (from IFacing) is not explicit!
Mobile.ROT (from IFacing) is not explicit!
Mobile.Facing (from IFacing) is not explicit!
AffectedByPowerOutage.GetValue (from ISelectionBar) is not explicit!
AffectedByPowerOutage.GetColor (from ISelectionBar) is not explicit!
ProductionBar.GetValue (from ISelectionBar) is not explicit!
ProductionBar.GetColor (from ISelectionBar) is not explicit!
SupportPowerChargeBar.GetValue (from ISelectionBar) is not explicit!
SupportPowerChargeBar.GetColor (from ISelectionBar) is not explicit!
TimedUpgradeBar.GetValue (from ISelectionBar) is not explicit!
TimedUpgradeBar.GetColor (from ISelectionBar) is not explicit!
Chronoshiftable.GetValue (from ISelectionBar) is not explicit!
Chronoshiftable.GetColor (from ISelectionBar) is not explicit!
PortableChrono.GetValue (from ISelectionBar) is not explicit!
PortableChrono.GetColor (from ISelectionBar) is not explicit!
TemporaryOwnerManager.GetValue (from ISelectionBar) is not explicit!
TemporaryOwnerManager.GetColor (from ISelectionBar) is not explicit!
Explicit interface violations: 24
```

ScottNZ:
I suggest a reword:
`A.B does not explicitly implement IA.B`

Phrohdoh:
How would you feel about `A does not explicitly implement IA.B`?
I don't see any gain from duplicating the member name.

pchote:
I'd prefer it if we could use wording that matched the other stylecop rules: "&lt;A&gt; must explicitly implement the interface I&lt;A&gt;.&lt;B&gt;"

Phrohdoh:
Ok, done.

RoosterDragon:
Food for thought: One drawback of doing this is that these interface methods become uncallable and unoverridable by derived classes.unless you manually provide workarounds for it.

See [here](https://msdn.microsoft.com/en-us/library/aa664593%28v=vs.71%29.aspx) and [here](https://msdn.microsoft.com/en-us/library/ms182153.aspx).

pchote:
@RoosterDragon is this a problem for us in practice?

RoosterDragon:
Probably not a major issue, but it would be odd in that we'd up end having to write explicit implementations that just defer to a public equivalent anyway. In fact we already have an example of this in `Warhead.cs` which does:
```c#
public readonly int Delay = 0;
int IWarhead.Delay { get { return Delay; } }
```

The explicit implementation here is entirely redundant:
- It does exactly what the default, implicit, implementation would.
- Since the `Delay` is already public, we're not protecting it from misuse by hiding it away in an explicit implementation.

Essentially, we're painting ourselves into a slightly dumb anti-pattern for those few traits that are intended to be inherited and modified.

pchote:
My understanding was that specific example was because we need a property for it to satisfy the warhead interface, and the explicit implementation is only there to resolve the naming conflict.  The same pattern is needed even for implicit implementations.

RoosterDragon:
I'm just saying that this sort of thing might up end cropping up in a few places if we need to satisfy this linter.

I'm not saying we shouldn't go ahead with this either, I'm just mentioning it.

pchote:
Ok, fair enough.  If this does become a problem then we can work around it by either keeping the interface implicit or by adding a new attribute to white-list specific members.

Needs a rebase, and then :+1: from me.

penev92:
Why drop the test commit? Fix it and make it not-test.
Otherwise this is adding dead weight.

pchote:
This should be ready for merge.  Can we get a rebase and fixup here?

Phrohdoh:
I count a single :+1:.

Phrohdoh:
Rebased and updated the test case to be a passing case.

obrakmann:
Looks fine to me. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/fb9d4c0aa4bd0ad8c9d14e1050a365f380cce09e)

Phrohdoh:
Cool! Thanks everyone.

PR #10144, opened 2015-11-29
Set missile default acceleration to 0 (prep only)

cjshmyr:
Needed to remake, this patch specifically targets the prep branch only. Previous discussion on #10142.

As discussed on IRC and proven out in #10129. We have to make another ticket regardless for `Next+1` to fix this properly in bleed. This is probably the least-invasive band-aid opposed to setting yaml values.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -50,7 +50,7 @@ namespace OpenRA.Mods.Common.Effects
 		public readonly WDist MaximumSpeed = new WDist(384);
 
 		[Desc("Projectile acceleration when propulsion activated.")]
-		public readonly WDist Acceleration = new WDist(5);
+		public readonly WDist Acceleration = WDist.Zero;
 
 		[Desc("How many ticks before this missile is armed and can explode.")]
 		public readonly int Arm = 0;


cjshmyr:
There is a flaw raised in the previous ticket I will carry over.

For the heightmap mods, the `MinimumLaunchSpeed` is only used when shooting over cliffs. By turning off acceleration the missiles don't behave quite like they should in these cases. Since this patch specifically targets prep/release, and we don't support TS in here yet, this shouldn't be too big an issue.

An alternate option is to keep acceleration on, and set the `MaximumSpeed` equal to `MaximumLaunchSpeed` in yaml for all missiles in the classic mods.

GraionDilach:
>> Set the MaximumSpeed default to MaximumLaunchSpeed instead of disabling Acceleration for the effect.
>
> If we go this route we'd make yaml additions instead of changing the info class default. I'm more in favor of this but opting for least effort if we can.

I'm still not sold. Using -1 on MaximumSpeed should disable acceleration on all missiles without requiring yaml changes either.

GraionDilach:
Meh, I reconsidered. It only affects prep and prolly one release cycle with lousy missile defaults ain't a big deal. :+1:

obrakmann:
:+1: alright, let's got with this.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/17730acc6e789cb12ef1c44e7e9d9a11a9854e8b)

PR #10151, opened 2015-11-30
Fixed FallToEarth groundlevel check

teees:
Fixes this activity for varrying height levels

PR #10152, opened 2015-11-30
Added aircraft husks (Tiberian Sun)

teees:
Needs #10151 to look acceptable. `Spins` and `Moves` are not sufficient for Tiberian Sun.

http://gfycat.com/SplendidLankyCoyote

### DIFF
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -517,6 +517,31 @@
 		CruiseAltitude: 2560
 	ReturnOnIdle:
 
+^AircraftHusk:
+	RenderVoxels:
+	RenderSprites:
+	WithVoxelBody:
+	BodyOrientation:
+		QuantizedFacings: 0
+		CameraPitch: 90
+	Aircraft:
+		AirborneUpgrades: airborne
+		CruisingUpgrades: cruising
+	Health:
+		HP: 280
+	Armor:
+		Type: Heavy
+	HiddenUnderFog:
+		Type: CenterPosition
+	AutoTargetIgnore:
+	ScriptTriggers:
+	Tooltip:
+		GenericName: Destroyed Aircraft
+	FallsToEarth:
+		Spins: true
+		Moves: true
+		Velocity: 86
+
 ^Viceroid:
 	Inherits@1: ^ExistsInWorld
 	Inherits@2: ^SpriteActor
--- a/mods/ts/rules/husks.yaml
+++ b/mods/ts/rules/husks.yaml
@@ -1,0 +1,85 @@
+DSHP.Husk:
+	Inherits: ^AircraftHusk
+	Tooltip:
+		Name: Dropship
+	Aircraft:
+		ROT: 5
+		Speed: 168
+	RenderSprites:
+		Image: dshp
+	RenderVoxels:
+		Image: dshp
+
+ORCA.Husk:
+	Inherits: ^AircraftHusk
+	Tooltip:
+		Name: Orca Fighter
+	Aircraft:
+		ROT: 5
+		Speed: 186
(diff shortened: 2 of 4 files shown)


mods/ts/rules/defaults.yaml
```diff
@@ -508,6 +508,30 @@
 		AirborneUpgrades: airborne
 	ReturnOnIdle:
 
+^AircraftHusk:
+	RenderVoxels:
```
penev92: You redefine this on each of the actors, so it's not necessary here.

epicelite:
They're supposed to crash different, but I'll take it.

teees:
@epicelite that is sadly not possible without modifying the trait

Mailaender:
Works as promised. :white_check_mark: You might want to rebase now that #10151 is merged.

mods/ts/rules/defaults.yaml
```diff
@@ -508,6 +508,30 @@
...
+	Armor:
+		Type: Heavy
+	HiddenUnderFog:
+		Type: CenterPosition
+	AutoTargetIgnore:
+	ScriptTriggers:
+	Tooltip:
+		GenericName: Destroyed Aircraft
+	FallsToEarth:
+		Spins: True
```
Phrohdoh: Purely for consistency, please use lowercase.

penev92: I know we go over this every couple of months, but I can't remember. Wasn't the standard (which, admittedly, isn't really enforced) to use "True" and "False"?

Phrohdoh:
```
tmba:openra thill\ $ ag "true" --yaml | wc -l
    1894
tmba:openra thill\ $ ag "True" --yaml | wc -l
     969
tmba:openra thill\ $ ag "false" --yaml | wc -l
    1577
tmba:openra thill\ $ ag "False" --yaml | wc -l
     531
```

So I'd say no.

teees: Still confusing since most values are upper case. Might want to check for this automatically and enforce it globally?

mods/ts/rules/defaults.yaml
pchote: Don't forget `RenderSprites` too!
All actors (even voxel ones) need this to display smoke effects, etc.

obrakmann:
:+1: once pchote's nit has been addressed.

Mailaender:
Needs a rebase.

teees:
rebased and added `RenderSprites:`

mods/ts/rules/husks.yaml
```diff
@@ -0,0 +1,73 @@
...
+	Inherits: ^AircraftHusk
+	Tooltip:
+		Name: Harpy
+	Aircraft:
+		ROT: 5
+		Speed: 130
+	WithSpriteRotorOverlay:
+		Offset: 85,0,384
+	RenderSprites:
+		Image: apache
```
abcdefg30: Why does this husk here have an `RenderSprites` override, but the others don't?

teees: Iirc it's currently only needed by WithSpriteRotorOverlay

pchote: Can you define it for the others too? This will make life easier when we add smoke and other effects.

atlimit8: The others have `RenderVoxels` which does the same thing as `RenderSprites` except for voxels instead of sprites.

pchote: Smoke and other effects are sprites.

teees:
rebased and adapted to requested changes

abcdefg30:
:+1: / :white_check_mark: now.

PR #10154, opened 2015-11-30
enable banshee death sound

teees:


mods/ts/rules/aircraft.yaml
```diff
@@ -260,6 +260,8 @@ SCRIN:
 		PipTypeEmpty: AmmoEmpty
 	AutoTarget:
 	RenderSprites:
+	DeathSounds:
+		DeathTypes: EnergyDeath, SmallExplosionDeath, ExplosionDeath, BulletDeath, HeadshotDeath, FireDeath
```
abcdefg30: This doesn't look correct. If no `DeathTypes` given, the sound should always play (regardless of death type).

I filed #10156 to fix this.

teees: Yeah I was asking in irc specifically because of this. Will adapt the pr

abcdefg30:
The sound is really hard to hear. Are we sure that this is the right sound and can we play it louder?

teees:
The problem here is that cabal is just too loud and the "ooops"/banshee death sound gets drowned by it.

abcdefg30:
Anyway, the sound is played. :+1:

obrakmann:
Sorry, but I don't hear anything, even when I remove the `ActorLostNotification` from `^Aircraft`.

obrakmann:
What's missing for this to work is this:
```diff
diff --git a/mods/ts/rules/aircraft.yaml b/mods/ts/rules/aircraft.yaml
index 23e4412..3db76e5 100644
--- a/mods/ts/rules/aircraft.yaml
+++ b/mods/ts/rules/aircraft.yaml
@@ -261,6 +261,7 @@ SCRIN:
        AutoTarget:
        RenderSprites:
        DeathSounds:
+               DeathTypes: Explosion, SmallExplosionDeath
 
 APACHE:
        Inherits: ^Helicopter
```

PR #10162, opened 2015-12-02
fix hovers trait to use heightmap

teees:
Uses the heightmap to get the correct altitude.
This was already worked around in Tiberian Sun via upgrades (#9716)

Since aircraft don't hover until reaching cruise altitude (at least in ts), I'd suggest reusing those upgrades in #10163

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/Hovers.cs
+++ b/OpenRA.Mods.Common/Traits/Render/Hovers.cs
@@ -23,19 +23,19 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Amount of Z axis changes in world units.")]
 		public readonly int OffsetModifier = -43;
 
+		public readonly int MinHoveringAltitude = 0;
+
 		public override object Create(ActorInitializer init) { return new Hovers(this, init.Self); }
 	}
 
 	public class Hovers : UpgradableTrait<HoversInfo>, IRenderModifier
 	{
 		readonly HoversInfo info;
-		readonly bool aircraft;
 
 		public Hovers(HoversInfo info, Actor self)
 			: base(info)
 		{
 			this.info = info;
-			aircraft = self.Info.HasTraitInfo<AircraftInfo>();
 		}
 
 		public IEnumerable<IRenderable> ModifyRender(Actor self, WorldRenderer wr, IEnumerable<IRenderable> r)
@@ -43,7 +43,9 @@ namespace OpenRA.Mods.Common.Traits
 			if (self.World.Paused || IsTraitDisabled)
 				return r;
 
-			var visualOffset = !aircraft || self.CenterPosition.Z > 0 ? (int)Math.Abs((self.ActorID + Game.LocalTick) / 5 % 4 - 1) - 1 : 0;
+			var visualOffset = self.World.Map.DistanceAboveTerrain(self.CenterPosition).Length >= info.MinHoveringAltitude
+				? (int)Math.Abs((self.ActorID + Game.LocalTick) / 5 % 4 - 1) - 1
+				: 0;
 			var worldVisualOffset = new WVec(0, 0, info.OffsetModifier * visualOffset);
 
 			return r.Select(a => a.OffsetBy(worldVisualOffset));


OpenRA.Mods.Common/Traits/Render/Hovers.cs
```diff
@@ -43,7 +43,7 @@ public IEnumerable<IRenderable> ModifyRender(Actor self, WorldRenderer wr, IEnum
 			if (self.World.Paused || IsTraitDisabled)
 				return r;
 
-			var visualOffset = !aircraft || self.CenterPosition.Z > 0 ? (int)Math.Abs((self.ActorID + Game.LocalTick) / 5 % 4 - 1) - 1 : 0;
+			var visualOffset = !aircraft || self.World.Map.DistanceAboveTerrain(self.CenterPosition).Length > 0 ? (int)Math.Abs((self.ActorID + Game.LocalTick) / 5 % 4 - 1) - 1 : 0;
```
Mailaender: > [SP2100] Code line must be no longer than 180 characters (now 181).

This one is a bit stupid, but you can easily solve this with 

```csharp
var height = self.World.Map.DistanceAboveTerrain(self.CenterPosition).Length;
```

GraionDilach: Or just a newline in front of the ?, then another in front of : for additional readability.

teees: Thanks, I missed that. Thought I would not need to make check my oneline changes :P

OpenRA.Mods.Common/Traits/Render/Hovers.cs
```diff
@@ -43,7 +43,9 @@ public IEnumerable<IRenderable> ModifyRender(Actor self, WorldRenderer wr, IEnum
 			if (self.World.Paused || IsTraitDisabled)
 				return r;
 
-			var visualOffset = !aircraft || self.CenterPosition.Z > 0 ? (int)Math.Abs((self.ActorID + Game.LocalTick) / 5 % 4 - 1) - 1 : 0;
+			var visualOffset = !aircraft || self.World.Map.DistanceAboveTerrain(self.CenterPosition).Length > 0
```
atlimit8: This would have to be `var visualOffset = aircraft == null || self.World.Map.DistanceAboveTerrain(self.CenterPosition).Length >= aircraft.MinAirborneAltitude`

penev92: Actually this trait shouldn't care about aircraft at all. Especially with the PR that only enables it at cruise altitude.

I'd lobby for keeping it `> 0`, but the TS MLRS is will need to hover only when above the ground (not when EMP-ed) (granted, that should still be possible now). So for the sake of both flexibility and clean architecture I suggest adding a `(Min)HoveringAltitude` or similar and removing the dependency on `Aircraft` entirely.

atlimit8: Actually, I agree with @penev92 here.

teees: While I agree with you both here, this pr should have only been about fixing a small oversight, not changing the way hovers works. Adding a `MinHoveringAltitude` will require yaml modifications.

teees: There could also be a third way: Since hover mrls in ts will spin to the ground when disabled via emp or ion storm, how about giving hovering vehicles a real altitude when hovering as well (probably in a second trait) and extend the FallsToEarth trait in a way where the actor can survive the crash on a fitting terrain (e.g. get killed on water). So checking for >= ground would still work.

teees:
Now depends on #10163 which disables the `Hovers:` trait when not cruising

OpenRA.Mods.Common/Traits/Render/Hovers.cs
Phrohdoh: If you do that take a look at the targeting setup to be sure the correct actors can still target it.

Mailaender:
https://github.com/OpenRA/OpenRA/pull/10163 has been merged.

teees:
rebased

Mailaender:
:+1:

penev92:
:+1:  Thanks.

PR #10163, opened 2015-12-02
Added cruising upgrade to aircraft

### LINKED ISSUE #9721: Actors with `Hovers` should not hover while changing altitude
Currently aircraft "hover" while landing/taking off.
We can fix that inside `Hovers` once `Aircraft` gets its long-needed overhaul.
Also the Hover MLRS shouldn't hover while going up/down slopes (unless that requires too much of a hack to be justifiable).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27608185-actors-with-hovers-should-not-hover-while-changing-altitude?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

teees:
Allows mods to toggle traits when reaching / leaving cruise altitude (e.g. hovering)

http://gfycat.com/FaithfulUnimportantAssassinbug

Fixes #9721 for all mods, makes the aircraft check in `Hovers.cs` redundant (see #10162)

### DIFF
--- a/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
+++ b/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
@@ -55,6 +55,10 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("The upgrades to grant to self while airborne.")]
 		public readonly string[] AirborneUpgrades = { };
 
+		[UpgradeGrantedReference]
+		[Desc("The upgrades to grant to self while at cruise altitude.")]
+		public readonly string[] CruisingUpgrades = { };
+
 		[Desc("Can the actor hover in place mid-air? If not, then the actor will have to remain in motion (circle around).")]
 		public readonly bool CanHover = false;
 
@@ -100,29 +104,7 @@ namespace OpenRA.Mods.Common.Traits
 		public int ROT { get { return Info.ROT; } }
 
 		bool airborne;
-		bool IsAirborne
-		{
-			get
-			{
-				return airborne;
-			}
-
-			set
-			{
-				if (airborne == value)
-					return;
-				airborne = value;
-				if (um != null)
-				{
-					if (airborne)
-						foreach (var u in Info.AirborneUpgrades)
-							um.GrantUpgrade(self, u, this);
-					else
-						foreach (var u in Info.AirborneUpgrades)
-							um.RevokeUpgrade(self, u, this);
-				}
-			}
-		}
+		bool cruising;
 
 		public Aircraft(ActorInitializer init, AircraftInfo info)
 		{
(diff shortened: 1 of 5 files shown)


OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -604,9 +599,54 @@ public void RemovedFromWorld(Actor self)
 			self.World.ActorMap.RemoveInfluence(self, this);
 			self.World.ActorMap.RemovePosition(self, this);
 			self.World.ScreenMap.Remove(self);
-			IsAirborne = false;
+			OnAirborneAltitudeLeft();
+			OnCruisingAltitudeLeft();
```
atlimit8: nit: These two should be flipped in case future `OnCruisingAltitudeLeft()` behavior assumes airborne.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -604,9 +599,54 @@ public void RemovedFromWorld(Actor self)
...
+		}
+
+		#region Airborne upgrades
+
+		void OnAirborneAltitudeReached()
+		{
+			if (airborne || um == null)
```
atlimit8: Either replace ` || um == null` with `if (um != null)` above `foreach (var u in Info.AirborneUpgrades)` or rename `airborne` to something like `airborneUpgraded` since `airborne` implies a generic state rather than an upgrade state.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -604,9 +599,54 @@ public void RemovedFromWorld(Actor self)
...
+				return;
+			airborne = true;
+			foreach (var u in Info.AirborneUpgrades)
+				um.GrantUpgrade(self, u, this);
+		}
+
+		void OnAirborneAltitudeLeft()
+		{
+			if (!airborne || um == null)
```
atlimit8: Same as above.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -604,9 +599,54 @@ public void RemovedFromWorld(Actor self)
...
+				um.RevokeUpgrade(self, u, this);
+		}
+
+		#endregion
+
+		#region Cruising upgrades
+
+		void OnCruisingAltitudeReached()
+		{
+			if (cruising || um == null)
```
atlimit8: Same as above.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -604,9 +599,54 @@ public void RemovedFromWorld(Actor self)
...
+				return;
+			cruising = true;
+			foreach (var u in Info.CruisingUpgrades)
+				um.GrantUpgrade(self, u, this);
 		}
 
+		void OnCruisingAltitudeLeft()
+		{
+			if (!cruising || um == null)
```
atlimit8: Same as above.

teees:
changed and rebased

Mailaender:
Seems to work without side-effects. :white_check_mark:

obrakmann:
Seems fine :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c2c3a90ac9fbfbcbd601ce8a1dc73caf439973bb)

Issue #10166, opened 2015-12-03
RA build queue isn't user friendly. 


GrayHatter:
I can not count the number of games I've lost because I've queued what I thought was a generous number of units. Only to run out of them, have too much cash, and see my build queue empty, just in time to get steamrolled. 

Two options that would fix this: 

Either play a warning sound when the queue is empty (similar to silo's needed).

Or, allow for setting an auto train unit. I.e. if I set my auto train unit to artillery, when the queue is empty, it'll auto build arty. But when queue any unit, it'll auto cancel the arty, and build that unit. Then once empty again, it'll automagically start building arty again.
##

GraionDilach:
Or just watch out your queues. You only have 6 afterall.

penev92:
"Just make the game play itself already."

jaZzKCS:
I have the feeling that this will again divide ppl into hardened blox of Opinionville.

If I understand it correctly, then you can set any unit on auto-queue but that would result in that production facility only producing that unit until the auto is stopped. Also true for any other unit from that same structure that is set to auto?

Because we want to make sure that you cannot auto-queue unit compositions, becasue then I really have to agree with penev here ("Hjelp! My game is playing itself" - a.k.a.: "Look Ma, no hands!")

TL;DR - It would be overkill if you could auto-queue more than one unit (compositions). Also keep in mind that this would make auto-queueing unit compositions instantly possible in TD since multiple construcion facilities give you multiple queues.

I have the feeling that this would have a far greater impact on TD than on RA in bigger games.

jaZzKCS:
referring to: https://github.com/OpenRA/OpenRA/pull/10169

Aaron-Lloyd:
The only time I feel auto-queuing or infinite queuing would ever actually be useful is maybe in the late part of a long TD match where you could potentially have 3 MCVs, 4 war factories, 4 helipads, and a couple of barracks and the number of queues becomes (a bit) less manageable. it would maybe make sense to be able to dedicate specific production facilities to certain units in that case with an autoqueue.

RoosterDragon:
At the risk of stating the obvious: It seems like a more reasonable solution to "I didn't notice the build queue was empty" would be to improve the UI so empty build queues were more obvious to the player.

r34ch:
@RoosterDragon +1

[...]

PR #10169, opened 2015-12-03
Added autoqueue feature.

irungentoo:
CTRL-~~Right~~ Left click on the queue to enable/disable auto queue.

Autoqueue makes the queue repeat itself infinitely until it's disabled or the items are removed which is very useful as it means you don't need to requeue units manually. Age of mythology has it.

I made the text color change to green to indicate when it is enabled.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
+++ b/OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
@@ -90,6 +90,7 @@ namespace OpenRA.Mods.Common.Traits
 		[Sync] public bool Enabled { get; private set; }
 
 		public string Faction { get; private set; }
+		public bool AutoQueue { get; private set; }
 
 		public ProductionQueue(ActorInitializer init, Actor playerActor, ProductionQueueInfo info)
 		{
@@ -98,6 +99,7 @@ namespace OpenRA.Mods.Common.Traits
 			playerResources = playerActor.Trait<PlayerResources>();
 			playerPower = playerActor.Trait<PowerManager>();
 			developerMode = playerActor.Trait<DeveloperMode>();
+			AutoQueue = false;
 
 			Faction = init.Contains<FactionInit>() ? init.Get<FactionInit, string>() : self.Owner.Faction.InternalName;
 			Enabled = !info.Factions.Any() || info.Factions.Contains(Faction);
@@ -230,10 +232,14 @@ namespace OpenRA.Mods.Common.Traits
 
 		public virtual void Tick(Actor self)
 		{
+			// Disable AutoQueue when nothing is building
+			if (queue.Count == 0)
+				AutoQueue = false;
+
 			while (queue.Count > 0 && BuildableItems().All(b => b.Name != queue[0].Item))
 			{
 				playerResources.GiveCash(queue[0].TotalCost - queue[0].RemainingCost); // refund what's been paid so far.
-				FinishProduction();
+				RemoveProduction();
 			}
 
 			if (queue.Count > 0)
@@ -245,54 +251,11 @@ namespace OpenRA.Mods.Common.Traits
 			if (!Enabled)
 				return;
 
-			var rules = self.World.Map.Rules;
 			switch (order.OrderString)
 			{
 				case "StartProduction":
 					{
-						var unit = rules.Actors[order.TargetString];
-						var bi = unit.TraitInfo<BuildableInfo>();
-						if (!bi.Queue.Contains(Info.Type))
-							return; /* Not built by this queue */
-
-						var cost = unit.HasTraitInfo<ValuedInfo>() ? unit.TraitInfo<ValuedInfo>().Cost : 0;
-						var time = GetBuildTime(order.TargetString);
-
-						if (BuildableItems().All(b => b.Name != order.TargetString))
-							return;	/* you can't build that!! */
-
-						// Check if the player is trying to build more units that they are allowed
-						var fromLimit = int.MaxValue;
-						if (!developerMode.AllTech && bi.BuildLimit > 0)
-						{
-							var inQueue = queue.Count(pi => pi.Item == order.TargetString);
-							var owned = self.Owner.World.ActorsHavingTrait<Buildable>().Count(a => a.Info.Name == order.TargetString && a.Owner == self.Owner);
-							fromLimit = bi.BuildLimit - (inQueue + owned);
-
-							if (fromLimit <= 0)
-								return;
-						}
-
-						var amountToBuild = Math.Min(fromLimit, order.ExtraData);
-						for (var n = 0; n < amountToBuild; n++)
-						{
-							var hasPlayedSound = false;
-							BeginProduction(new ProductionItem(this, order.TargetString, cost, playerPower, () => self.World.AddFrameEndTask(_ =>
-							{
-								var isBuilding = unit.HasTraitInfo<BuildingInfo>();
-
-								if (isBuilding && !hasPlayedSound)
-									hasPlayedSound = Game.Sound.PlayNotification(rules, self.Owner, "Speech", Info.ReadyAudio, self.Owner.Faction.InternalName);
-								else if (!isBuilding)
-								{
-									if (BuildUnit(order.TargetString))
-										Game.Sound.PlayNotification(rules, self.Owner, "Speech", Info.ReadyAudio, self.Owner.Faction.InternalName);
-									else if (!hasPlayedSound && time > 0)
-										hasPlayedSound = Game.Sound.PlayNotification(rules, self.Owner, "Speech", Info.BlockedAudio, self.Owner.Faction.InternalName);
-								}
-							})));
-						}
-
+						StartProduction(order.TargetString, order.ExtraData);
 						break;
 					}
 
@@ -307,6 +270,16 @@ namespace OpenRA.Mods.Common.Traits
 				case "CancelProduction":
 					{
 						CancelProduction(order.TargetString, order.ExtraData);
+						break;
+					}
+
+				case "ToggleAutoQueue":
+					{
+						// Toggle it for right queue only
+						if (BuildableItems().All(b => b.Name != order.TargetString))
+							return;
+
+						AutoQueue = order.ExtraData == 1;
 						break;
 					}
 			}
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -309,6 +272,15 @@ public void ResolveOrder(Actor self, Order order)
 						CancelProduction(order.TargetString, order.ExtraData);
 						break;
 					}
+				case "ToggleAutoQueue":
+					{
+						/* Toggle it for right queue only, */
```
abcdefg30: Our convention is to use `//`. Also please remove the comma at the end.

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -309,6 +272,15 @@ public void ResolveOrder(Actor self, Order order)
...
+						if (BuildableItems().All(b => b.Name != order.TargetString))
+							return;
+
+						autoqueue = order.ExtraData == 1 ? true : false;
```
abcdefg30: `autoqueue = order.ExtraData == 1;` should be enough.

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -90,6 +90,7 @@ public class ProductionQueue : IResolveOrder, ITick, ITechTreeElement, INotifyOw
 		[Sync] public bool Enabled { get; private set; }
 
 		public string Faction { get; private set; }
+		public bool autoqueue { get; private set; }
```
GraionDilach: This will trigger StyleCop for sure. Public members should be capitalized (as in `AutoQueue`) and in this case probably even synced as well.

pchote:
This is a controversial feature because there is a strong feeling by a lot of players that "the game shouldn't play itself".  Managing your production is one of the key features of the game, and is certainly one of the biggest deciders between an average/poor and a good player.

Do any other games have this feature?  Unless there is a precedence in other well known games (and therefore a reasonable expectation for this feature to exist), then I don't think we should include this feature.

GraionDilach:
IIRC Warzone 2100 also had this but it was quite a while ago I last played it. A boolean parameter in the Infos with the default being false could be an option.

x-a-n-a-x:
I would love to see a mod featuring this "autoqueue". 
It remembers me of SupremeCommander: https://github.com/OpenRA/OpenRA/issues/2997#issuecomment-22486399

irungentoo:
You still have to manage your queue with this, it just means less button pressing.

jaZzKCS:
I have the feeling that this will again divide ppl into hardened blox of Opinionville.

If I understand it correctly, then you can set any unit on auto-queue but that would result in that production facility only producing that unit until the auto is stopped. Also true for any other unit from that same structure that is set to auto?

Because we want to make sure that you _cannot_ auto-queue unit compositions, becasue then I really have to agree with penev here ("Hjelp! My game is playing itself" - a.k.a.: "Look Ma, no hands!")

pchote:
Auto-queing only one unit is redundant to spam-shift-clicking to queue up 500 units of that type.

irungentoo:
You can already queue/dequeue enough units using the shift key in seconds to last the entire game so I don't see how this makes the game play itself.

jaZzKCS:
I guess you want this as an alternative to those situations where you are frantically removing the 56 queues of that certain unit type you issued just 2 mins ago, becasue the in-game situation has changed and you need to build something else.

WolfGaming:
Imo if thats the reason it would be better to just add a cancel all button, wasnt there a PR for that recently?

atlimit8:
All this does is remove mindless click spamming to reach a number and not any significant thought. However, with #9595 for directly canceling and CTRL + SHIFT + left_click used to queue say 100 (or enough to last at least a long game), this can be approximated.

This is incompatible with #9595 since that uses CTRL + right_click for something else. Also, right clicking is associated with canceling and left clicking with queuing. Therefore, this should use CTRL + left_click.

Placing "PR: Changes Requested" for switching to the more orthogonal use of CTRL + left_click instead of CTRL + right_click.

GrayHatter:
I reject the arguments that thin makes the game play it self. For the exact reasons mentioned above, shift clicking to reach >100 is already common. I could use autohotkey to get this same effect. 

Lets be honest here, this only solves one problem, having to micro your build queue on broken maps, or broken units. The two units (I only play RA) that will see this, is Arty and SHOK, maybe a few others but not to the levels seen by these two. 

Why did you set units to defend by default? Why do miners do searching for gems/ore? why are there rally points for buildings. These are all the game "playing itself" except not really... if you set a bad rally point, kiss your units good bye, auto queue the wrong unit, good buy cash hello steamroller.

And just to make sure I hit every point, if you really want to go with the whole " separates better players" argument, why does every one who plays sovi spam shok inf & flak, and every one who plays allies spam arty and hinds?? Auto queue is a solution to one unit being the only one that anyone uses.

irungentoo:
As requested, CTRL-Left click is now used to enable autoqueue.

pchote:
@GrayHatter: Those points sound to me like an argument *against* this feature, and that the balance of the RA mod should be fixed instead.

GrayHatter:
@pchote, it is, but if I come out and say that arty and SHOK are both OP, I'd get lynched. No one want's you to nerf their favorite strategy. But that's still no reason that this shouldn't exist. Even if balance issues are fixed, the only thing you're doing is reducing the number of clicks.

pchote:
I assume you're talking about the current (old) release: SHOK have been heavily nerfed in the playtests/next release.

GrayHatter:
@pchote I play release more because of the performance, but even in the playtest, If I'm playing russia, I'm still spamming Shok and flak maybe some scuds MAYBE

penev92:
>You can already queue/dequeue enough units using the shift key in seconds to last the entire game so I don't see how this makes the game play itself.

And you don't see that as a problem? With the mod's/map's economy, gameplay or production, unit balance?

>Lets be honest here, this only solves one problem

The thing is it doesn't solve *any* problem (sorry @irungentoo ).
And with the argument made (by the supporters of this, no less!) that you can queue 150 units in 2 seconds anyway, I really don't see the point.
:-1: 

I could probably go with it if it is configurable (it's not even configurable!), so thirdparty mods can enable it per production queue if they want, but I'm guessing that's now what you want here, but rather want it in RA.

pchote:
If we do decide to take this feature then are some UI issues that will first need to be resolved:
* Ctrl-clicking on an item that currently isn't in the queue will do nothing.
* If you ctrl-click on an item in the queue while you have multiple units queued then it will repeat the entire queue (e.g. 5 harvesters, then 5 bikes, then 5 buggies in a loop), but the UI doesn't indicate what is going on: the queued numbers stay frozen at the original numbers.  This is confusing: expected behaviour would be to have the queued number count down so you know where in the queue you are.

pchote:
The best course of action would be to discuss this in IRC and work out exactly how the interaction for this should work before rewriting it (over and over again) to save a lot of work and hassle.

irungentoo:
Point 2 should be fixed, not sure what ctrl-clicking an item that isn't in the queue should do.

pchote:
Presumably the same thing as ctrl-clicking an item that isn't in the queue.

atlimit8:
The style fixes should be squashed away.

irungentoo:
I think everything is good now.

GrayHatter:
Just played it, no issues. worked as I expected it to!

irungentoo:
Just tested this on in real multiplayer game with 3 other people each using the autoqueue and it worked perfectly.

If anyone against this change hasn't tried it out, please do and you might change your mind.

pchote:
The real problem, which I guess I didn't make clear enough above, is that ctrl-clicking on an icon doesn't do anything related to that icon, which is not intuitive.  Ctrl-clicking on an icon should auto-queue *that* unit, and not just toggle the queue repeat.  Proper UX for this would be to have a separate "toggle queue" button.

I believe that @r34ch's proposal in #10166 to indicate queue status on the queue icon is a much better solution to the problem, so on balance I am now against taking this feature.

atlimit8:
> I believe that @r34ch's proposal in #10166 to indicate queue status on the queue icon is a much better solution to the problem, ...

atlimit8:
@pchote, @r34ch, etc. Should the tab have some kind of overlay, 4 states ([not] selected, [not] auto-queued), or something else?

obrakmann:
IMO, the fact alone that this is a problem *at all*, as opposed to, say, wondering whether you would even be able to *afford* all those units, is just an indicator to me that something is already horribly wrong with either our gameplay, or the maps that people play. :-1:

abcdefg30:
Needs a rebase.

GrayHatter:
@obrakmann IMO the problem with your game play is your (read: the common) view on it. You either think of it like an RTS, where spamming units is completely acceptable. Or a warfare game, in which spamming really shouldn't be a thing. 

In reality, any tank that one shots another, has already won, that tank is destroyed. So, if you know you're units are just going to go down to the first shot, you're a lot more tactical, and a lot more careful with them. 

If you want to stop the whole idea of spamming one unit, just crank up the damage by type to 11. If I know sending a wave of tanks into your army of tanks would just result in them dying before they can fire, I'd send a wave of rocket inf instead, hoping that you don't already have hind's or flame towers, or rifle inf already waiting. And by the same token, If I know you have a wall of shok troops, I might train something long range instead. 

The problem of spamming 1 unit, or one combo (flak and shok) over and over, is that there's no punishment for it. There needs to be a units for each that can't lose. Think rock paper scissors (without scissors), if you can just throw rock over and over, why would you do anything else?

WolfGaming:
However with that attitude I should just get 1 or 2 of every unit and stay
on the defence until my enemy runs out of money or my 5 nukes charge up.
Since attack move is delayed and the defender can get the first shot most
of the time.

klaas1:
I agree with Obrakman here, the problem lies in some of the maptypes people play (think "gods great devide"), where money is practically infinite, and the team that manages build ques the best wins.

On maps where money is not infinite, I think this option will be detrimental to the game. Instead of actively chosing which unit to build, moderate players will just spam the "best" unit, making the game more predictable and boring.

pchote:
It looks like the discussion here has run its course, and that the overall feeling is to not take this.  Closing as part of a wider clean up of the PR queue.

PR #10176, opened 2015-12-04
If PipCount < 0, then use max / -PipCount

atlimit8:
Applies to:
- `AmmoPool` : `Ammo / -PipCount`
- `Cargo` : `MaxWeight / -PipCount`
- `Harvester` : `Capacity / -PipCount`
- `StoresResources` : `Capacity / -PipCount`

### DIFF
--- a/OpenRA.Mods.Common/Traits/Cargo.cs
+++ b/OpenRA.Mods.Common/Traits/Cargo.cs
@@ -24,7 +24,8 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("The maximum sum of Passenger.Weight that this actor can support.")]
 		public readonly int MaxWeight = 0;
 
-		[Desc("Number of pips to display when this actor is selected.", "0 means no visible pips.", "-1 means same amount as MaxWeight.")]
+		[Desc("Number of pips to display when this actor is selected.", "0 means no visible pips.", "-1 means same amount as MaxWeight.",
+			"< -1 means Capacity / -PipCount pips.")]
 		public readonly int PipCount = 0;
 
 		[Desc("`Passenger.CargoType`s that can be loaded into this actor.")]
@@ -256,7 +257,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		public IEnumerable<PipType> GetPips(Actor self)
 		{
-			var numPips = Info.PipCount >= 0 ? Info.PipCount : Info.MaxWeight;
+			var numPips = Info.PipCount >= 0 ? Info.PipCount : Info.MaxWeight / -Info.PipCount;
 
 			for (var i = 0; i < numPips; i++)
 				yield return GetPipAt(i);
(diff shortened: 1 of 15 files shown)


GraionDilach:
RA2 had PipWrap: http://modenc.renegadeprojects.com/PipWrap

OpenRA.Mods.Common/Traits/Cargo.cs
```diff
@@ -264,7 +265,7 @@ public IEnumerable<PipType> GetPips(Actor self)
 
 		PipType GetPipAt(int i)
 		{
-			var n = i * Info.MaxWeight / Info.PipCount;
+			var n = Info.PipCount >= 0 ? i * Info.MaxWeight / Info.PipCount : i / -Info.PipCount;
```
pchote: This behaves counter-intuitively ingame.  If you have a half-full transport with `PipCount: -2` then I would expect that the pip row would be half full, not full.

OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -420,7 +421,7 @@ public void OnNotifyResourceClaimLost(Actor self, ResourceClaim claim, Actor cla
 
 		PipType GetPipAt(int i)
 		{
-			var n = i * Info.Capacity / Info.PipCount;
+			var n = Info.PipCount >= 0 ? i * Info.Capacity / Info.PipCount : i / -Info.PipCount;
```
pchote: Likewise here.  Having a harvester that continues to harvest after its pip bar is full is confusing.

pchote:
I agree with @penev92 that this is a bit confusing, especially with the behaviour I noted above.  What's the motivation behind doing this?

atlimit8:
`-PipCount` units per pip so that the number of pips scales with `Ammo`, `MaxWeight`, or `Capacity`.
If that is too confusing, I can roll this back to `-1` means same amount as `Ammo`, `MaxWeight`, or `Capacity`.

Phrohdoh:
Would it make sense to turn the construct of pips entirely into an interface that has a `string Type` so implementors could do the math and always return a positive (or `0`) value?

atlimit8:
I have a branch started with:
```c#
	public struct PipType
	{
		public readonly string Sequence;
		public readonly float Fill = 0;
	}
```
where `Sequence` is the pip sequence to use and `(int)((sequence_frame_count - 1) * Fill)` is the sequence frame to allow partial fill pips.

atlimit8:
Perhaps, this should be reduced to supporting just `-1`. Maybe another property should be added such as `PipCountType` from something like `enum PipCountType { Total, Weight[, Frames] }` where `Total` is current/default behavior, `Weight` means each full pip is `PipCount` units[, and `Frames` means each full pip is sequence frame count - 1 (for empty) units (in other branch)].

chrisforbes:
Big nak to magic values, magic handling of the sign value, etc.

Use another field as a magic behavior flag / mode selector.

PR #10182, opened 2015-12-05
Fix AttackMove not working properly under certain conditions

### LINKED ISSUE #8023: Rubber band selection does not work while AttackMove is active
Hello,

Since 24042015 release there is an annoying bug which is causing selection of units difficult when using hotkeys. Say you are selecting some units and move them using attack command then try to select another group by dragging on screen.. it doesn't work as it should.

In other words.. when your cursor is green you can select and re-select any group of units by dragging on screen... if you cursor is red you can't .

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12079825-rubber-band-selection-does-not-work-while-attackmove-is-active?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #9851: Attack-move works unreliably in new playtest
I was playing the playtest online and was quite frustrated that the attack-move function seemed to only work about 60% of the time. Another player in the game I played complained of the same thing.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27873861-attack-move-works-unreliably-in-new-playtest?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
This fixes the issue where AttackMove (and possibly other order modes) would not work properly when people were clicking their mouse buttons at the same time. A move order (the default order mode) would be issued instead. See [here](https://github.com/OpenRA/OpenRA/issues/9851#issuecomment-162229788).

Fixes #9851.
Fixes #8023.

### DIFF
--- a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
+++ b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
@@ -120,12 +120,25 @@ namespace OpenRA.Widgets
 							}
 						}
 					}
-					else if (dragStart.HasValue)
+					else
 					{
-						// Select actors in the dragbox
-						var newSelection = SelectActorsInBoxWithDeadzone(World, dragStart.Value, xy);
-						World.Selection.Combine(World, newSelection, mi.Modifiers.HasModifier(Modifiers.Shift), dragStart == xy);
+						/* The block below does three things:
+						// 1. Allows actor selection using a selection box regardless of input mode.
+						// 2. Allows actor deselection with a single click in the default input mode (UnitOrderGenerator).
+						// 3. Prevents units from getting deselected when exiting input modes (eg. AttackMove or Guard).
+						//
+						// We cannot check for UnitOrderGenerator here since it's the default order generator that gets activated in
+						// World.CancelInputMode. If we did check it, actor de-selection would not be possible by just clicking somewhere,
+						// only by dragging an empty selection box.
+						*/
+						if (dragStart.HasValue && (!(World.OrderGenerator is GenericSelectTarget) || hasBox))
+						{
+							var newSelection = SelectActorsInBoxWithDeadzone(World, dragStart.Value, xy);
+							World.Selection.Combine(World, newSelection, mi.Modifiers.HasModifier(Modifiers.Shift), dragStart == xy);
+						}
 					}
+
+					World.CancelInputMode();
 				}
 
 				dragStart = dragEnd = null;
--- a/OpenRA.Game/Orders/GenericSelectTarget.cs
+++ b/OpenRA.Game/Orders/GenericSelectTarget.cs
@@ -13,19 +13,19 @@ using OpenRA.Graphics;
 
 namespace OpenRA.Orders
 {
-	public class GenericSelectTarget : IOrderGenerator
+	public class GenericSelectTarget : UnitOrderGenerator
 	{
-		readonly IEnumerable<Actor> subjects;
-		readonly string order;
-		readonly string cursor;
-		readonly MouseButton expectedButton;
+		protected readonly IEnumerable<Actor> Subjects;
+		protected readonly string OrderName;
+		protected readonly string Cursor;
+		protected readonly MouseButton ExpectedButton;
 
 		public GenericSelectTarget(IEnumerable<Actor> subjects, string order, string cursor, MouseButton button)
 		{
-			this.subjects = subjects;
-			this.order = order;
-			this.cursor = cursor;
-			expectedButton = button;
+			Subjects = subjects;
+			OrderName = order;
+			Cursor = cursor;
+			ExpectedButton = button;
 		}
 
 		public GenericSelectTarget(IEnumerable<Actor> subjects, string order, string cursor)
@@ -37,26 +37,23 @@ namespace OpenRA.Orders
 		public GenericSelectTarget(Actor subject, string order, string cursor, MouseButton button)
 			: this(new Actor[] { subject }, order, cursor, button) { }
 
-		public IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
+		public override IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
 		{
-			if (mi.Button != expectedButton)
+			if (mi.Button != ExpectedButton)
 				world.CancelInputMode();
 			return OrderInner(world, xy, mi);
 		}
 
-		IEnumerable<Order> OrderInner(World world, CPos xy, MouseInput mi)
+		protected virtual IEnumerable<Order> OrderInner(World world, CPos xy, MouseInput mi)
 		{
-			if (mi.Button == expectedButton && world.Map.Contains(xy))
+			if (mi.Button == ExpectedButton && world.Map.Contains(xy))
 			{
 				world.CancelInputMode();
-				foreach (var subject in subjects)
-					yield return new Order(order, subject, false) { TargetLocation = xy };
+				foreach (var subject in Subjects)
+					yield return new Order(OrderName, subject, false) { TargetLocation = xy };
 			}
 		}
 
-		public virtual void Tick(World world) { }
-		public IEnumerable<IRenderable> Render(WorldRenderer wr, World world) { yield break; }
-		public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world) { yield break; }
-		public string GetCursor(World world, CPos xy, MouseInput mi) { return world.Map.Contains(xy) ? cursor : "generic-blocked"; }
+		public override string GetCursor(World world, CPos xy, MouseInput mi) { return world.Map.Contains(xy) ? Cursor : "generic-blocked"; }
 	}
 }
--- a/OpenRA.Game/Orders/UnitOrderGenerator.cs
+++ b/OpenRA.Game/Orders/UnitOrderGenerator.cs
@@ -15,7 +15,7 @@ using OpenRA.Traits;
 
 namespace OpenRA.Orders
 {
-	class UnitOrderGenerator : IOrderGenerator
+	public class UnitOrderGenerator : IOrderGenerator
 	{
 		static Target TargetForInput(World world, CPos xy, MouseInput mi)
 		{
@@ -36,7 +36,7 @@ namespace OpenRA.Orders
 			return Target.FromCell(world, xy);
 		}
 
-		public IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
+		public virtual IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
 		{
 			var target = TargetForInput(world, xy, mi);
 			var actorsAt = world.ActorMap.GetActorsAt(xy).ToList();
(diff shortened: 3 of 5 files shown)


obrakmann:
Sorry, Penev, I know you're dead-set against this, but I think this change is worth it, for several reasons:

1. For once, we have received feedback on our playtests, and it would be a really bad message if we weren't to act on it.
2. We've had *multiple* people that have encountered this. Imagine the number of people that would complain once the release is out with this bug.
2. It's not like it's disabling the feature altogether, it just uses a different button. I'm aware that the middle button is harder to use than the right button, but if you really can't use the middle button at all, I suggest you switch to left-click orders where the scrolling is still on the right button.

pchote:
I support the motivation behind this, so :+0.5: (haven't tested it).  We can always move it back to the RMB (or, ideally, make the button configurable) in a future release cycle where we aren't constrained for time.

penev92:
>[18:18:08]	< penev|m> We're not breaking the best thing that happened to OpenRA since the voxel parsing
[18:18:13]	< penev|m> To fix a non-issue
[18:18:33]	< penev|m> That's how I feel

So that is a BIG :-1: from me.

penev92:
Also for the record:

>[18:19:36]	< antares79> and moving to the middle mouse button would be a simple four-line fix
[18:19:56]	< penev|m> Yes, but would totally ruin usability

and 

> We can always move it back to the RMB

<Insert sarcastic remark about that happening here>

pchote:
@penev92: Feedback from our playtesters is that we *have* broken the best thing that happened to OpenRA since before voxel parsing (i.e. Attack Move).  This temporarily fixes the issue so that we can issue a release that won't "force" our incredibly fickle player base to abandon us.

If you can fix the input handling to make RMB scrolling compatible with our original attack move behavior, then I would support merging that over this.

penev92:
Analysis of that feedback clearly showed the problem is caused by user error.

But [oh, well...!](https://xkcd.com/1172/)

fgdzfggsfgsfgsf:
Thanks everyone, and sorry Penev :(

penev92:
```diff
diff --git a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
index 363aaa3..4f383dc 100644
--- a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
+++ b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
@@ -68,7 +68,7 @@ public override bool HandleMouseInput(MouseInput mi)
 				dragStart = xy;
 
 				// Place buildings, use support powers, and other non-unit things
-				if (!(World.OrderGenerator is UnitOrderGenerator))
+				if (!(World.OrderGenerator is UnitOrderGenerator || World.OrderGenerator is GenericSelectTarget))
 				{
 					ApplyOrders(World, mi);
 					dragStart = dragEnd = null;
@@ -128,6 +128,9 @@ public override bool HandleMouseInput(MouseInput mi)
 					}
 				}
 
+				if (World.OrderGenerator is GenericSelectTarget)
+					World.CancelInputMode();
+
 				dragStart = dragEnd = null;
 				YieldMouseFocus(mi);
 			}

```

Doesn't "break" order issuing even when given crap user input, and doesn't break joystick scrolling.

pchote:
The problem there is that only works around the problem for a-move specifically (which uses `GenericSelectTarget`.  It won't solve the problem for Guard or the other order generators that aren't explicitly whitelisted.

penev92:
Take two.

```diff
diff --git a/OpenRA.Game/Orders/GenericSelectTarget.cs b/OpenRA.Game/Orders/GenericSelectTarget.cs
index 3f987be..a84fb99 100644
--- a/OpenRA.Game/Orders/GenericSelectTarget.cs
+++ b/OpenRA.Game/Orders/GenericSelectTarget.cs
@@ -13,7 +13,7 @@
 
 namespace OpenRA.Orders
 {
-	public class GenericSelectTarget : IOrderGenerator
+	public class GenericSelectTarget : UnitOrderGenerator
 	{
 		readonly IEnumerable<Actor> subjects;
 		readonly string order;
@@ -37,7 +37,7 @@ public GenericSelectTarget(Actor subject, string order, string cursor)
 		public GenericSelectTarget(Actor subject, string order, string cursor, MouseButton button)
 			: this(new Actor[] { subject }, order, cursor, button) { }
 
-		public IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
+		public override IEnumerable<Order> Order(World world, CPos xy, MouseInput mi)
 		{
 			if (mi.Button != expectedButton)
 				world.CancelInputMode();
@@ -54,9 +54,6 @@ IEnumerable<Order> OrderInner(World world, CPos xy, MouseInput mi)
 			}
 		}
 
-		public virtual void Tick(World world) { }
-		public IEnumerable<IRenderable> Render(WorldRenderer wr, World world) { yield break; }
-		public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr, World world) { yield break; }
-		public string GetCursor(World world, CPos xy, MouseInput mi) { return world.Map.Contains(xy) ? cursor : "generic-blocked"; }
+		public override string GetCursor(World world, CPos xy, MouseInput mi) { return world.Map.Contains(xy) ? cursor : "generic-blocked"; }
 	}
 }
```

Do the same for `GuardOrderGenerator` and whoever else might want to join the club.
This way the existing checks do the work without needing to touch `WorldInteractionControllerWidget`. Results are same as with the previous diff.

pchote:
Assuming that @penev92's take two works (I haven't tested it) then can we go with that?

Phrohdoh:
If this goes in how do I "grab" the map and fling it about? (the current behaviour for middle mouse)

I'm for this completely as long as the above moves to the right mouse button and isn't trashed.

obrakmann:
> Assuming that @penev92's take two works (I haven't tested it) then can we go with that?

Sure. I'd prefer it if the scrolling could stay on the RMB as well. I'm going to check it out now.

obrakmann:
Updated with penev's suggestion, which indeed seems to do the trick. Didn't notice any misbehaviour during a quick skirmish and a game of allies05.

RoosterDragon:
:+1: I was never able to recreate the raw race-condition style issue from #9851, but the overlapping clicks case (https://github.com/OpenRA/OpenRA/issues/9851#issuecomment-162229788) is indeed fixed. Code appears fine to me.

PS: I would avoid rebasing onto bleed again until after we wrangle this onto prep.

penev92:
As a side-effect you can't get of the attack (or guard) cursor.
You'll need to add 

```diff
diff --git a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
index 363aaa3..887096f 100644
--- a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
+++ b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
@@ -126,6 +126,8 @@ public override bool HandleMouseInput(MouseInput mi)
 						var newSelection = SelectActorsInBoxWithDeadzone(World, dragStart.Value, xy);
 						World.Selection.Combine(World, newSelection, mi.Modifiers.HasModifier(Modifiers.Shift), dragStart == xy);
 					}
+
+					World.CancelInputMode();
 				}
 
 				dragStart = dragEnd = null;
```

obrakmann:
Updated. I haven't rebased it yet (which will be quite a bitch, what the hell happened?). I confirmed that the current state still applies cleanly to prep, though.

RoosterDragon:
#10118 happened - it may be easier to reapply the diff conceptually using the same "make a base class and override these few things" since that will be less conflicty than trying to just merge at a text level.

obrakmann:
> As a side-effect you can't get of the attack (or guard) cursor. You'll need to add [...]

That does a little bit too much now, it deselects the units, too.

obrakmann:
Updated. I rebased now. I have a second branch based on prep for which I will open a separate PR (with the identical changes, just without the conflicts).

obrakmann:
Fixed the issue where exiting guard mode would clear the selection.

OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
```diff
@@ -120,12 +120,14 @@ public override bool HandleMouseInput(MouseInput mi)
 							}
 						}
 					}
-					else if (dragStart.HasValue)
+					else if (dragStart.HasValue && (!(World.OrderGenerator is GenericSelectTarget) || hasBox))
```
pchote: Why can't you use `UnitOrderGenerator` here, like above?

obrakmann: Because then you can't deselect units anymore.

pchote: Ok, I figured it would be something like that.  Could you please add a comment explaining this?

obrakmann: Comment added

obrakmann: Whoops, I just noticed StyleCop complains about the comment's location. The way I did it now is the only way it wouldn't complain about the comment. Should we disable that rule (SA1108), to make comments like below possible?

```csharp
if (x)
{
    foo();
}
// Comment
else if (y)
{
    bar();
}

Comment (penev92) [inline OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs]: Fwiw the `dragStart.HasValue && ` part is completely pointless there.

**Comment by penev92** [comment]
Works for all order targeters I could think of and I don't see anything wrong with the code, so :+1: 
I count +2, but I won't merge this just yet because this has been rather controversial, so I'll wait and see if people have complaints/ideas/suggestions until tomorrow.

**Comment by abcdefg30** [comment]
Looks good to me from a quick look/test.

File: OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs (hunk 1 above)
Comment (obrakmann) [inline OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs]: I removed the `dragStart.HasValue` part now, it indeed didn't make a difference. Also found a way to place the comment in such way that it pleases StyleCop.

File: OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs [hunk 2]
```diff
@@ -120,12 +120,23 @@ public override bool HandleMouseInput(MouseInput mi)
 							}
 						}
 					}
-					else if (dragStart.HasValue)
-					{
-						// Select actors in the dragbox
-						var newSelection = SelectActorsInBoxWithDeadzone(World, dragStart.Value, xy);
-						World.Selection.Combine(World, newSelection, mi.Modifiers.HasModifier(Modifiers.Shift), dragStart == xy);
-					}
+					else
```
pchote: style nit: can we get some braces around this to avoid the hanging block?

obrakmann: Good idea, done.

abcdefg30:

abcdefg30:
The changelog will come with the pr against prep.

PR #10185, opened 2015-12-06
Fix showing unread messages count on active lobby chat button

obrakmann:
The lobby chat button text would change to include the unread messages count for a split second, even when the lobby chat was the current panel, leading to the button text sort of 'blinking' in an ugly way.

This is simple enough and provides better polish for a new feature that I think it's warranted to be added to the milestone.


abcdefg30:
:+1:

RoosterDragon:
:+1:

PR #10189, opened 2015-12-07
AttackBase and Armament optimizations

reaperrr:
- Removes redundant ITraitInfo from `AttackBase`.
- Moves `self.IsDisabled()` check in `AttackBase.CanAttack` up to first check, so later checks are skipped for disabled actors.
- Caches the assigned turret and `AmmoPool` as well als `BodyOrientation` and `rangeModifier` directly in `Armament` via `INotifyCreated` instead of using Lazy.

Note: Please review at least the 3rd commit very carefully, I'm not sure the performance impact is worth it and I'm not sure whether this might cause any regressions.
The approach of the last commit was simply copied from one of @RoosterDragon's PRs.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Armament.cs
+++ b/OpenRA.Mods.Common/Traits/Armament.cs
@@ -87,17 +87,17 @@ namespace OpenRA.Mods.Common.Traits
 		}
 	}
 
-	public class Armament : UpgradableTrait<ArmamentInfo>, ITick, IExplodeModifier
+	public class Armament : UpgradableTrait<ArmamentInfo>, INotifyCreated, ITick, IExplodeModifier
 	{
 		public readonly WeaponInfo Weapon;
 		public readonly Barrel[] Barrels;
 
 		readonly Actor self;
-		Lazy<Turreted> turret;
-		Lazy<BodyOrientation> coords;
-		Lazy<AmmoPool> ammoPool;
+		Turreted turret;
+		AmmoPool ammoPool;
+		BodyOrientation coords;
+		IEnumerable<int> rangeModifiers;
 		List<Pair<int, Action>> delayedActions = new List<Pair<int, Action>>();
-		Lazy<IEnumerable<int>> rangeModifiers;
 
 		public WDist Recoil;
 		public int FireDelay { get; private set; }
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Traits/Armament.cs
```diff
@@ -140,6 +139,13 @@ public WDist MaxRange()
 
 		public void Tick(Actor self)
 		{
+			if (!initialized)
```
penev92: How about moving this into `INotifyCreated`.

OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
```diff
@@ -70,7 +70,7 @@ public AttackBase(Actor self, AttackBaseInfo info)
 
 		protected virtual bool CanAttack(Actor self, Target target)
 		{
-			if (!self.IsInWorld || IsTraitDisabled)
+			if (!self.IsInWorld || self.IsDisabled() || IsTraitDisabled)
```
penev92: `if (!self.IsInWorld || IsTraitDisabled || self.IsDisabled())`  for bonus cookie points.

OpenRA.Mods.Common/Traits/Armament.cs
```diff
@@ -93,15 +93,16 @@ public class Armament : UpgradableTrait<ArmamentInfo>, ITick, IExplodeModifier
 		public readonly Barrel[] Barrels;
 
 		readonly Actor self;
-		Lazy<Turreted> turret;
+		Turreted turret;
+		AmmoPool ammoPool;
 		Lazy<BodyOrientation> coords;
-		Lazy<AmmoPool> ammoPool;
 		List<Pair<int, Action>> delayedActions = new List<Pair<int, Action>>();
 		Lazy<IEnumerable<int>> rangeModifiers;
```
RoosterDragon: Why not un-lazy `coords` and `rangeModifiers` too?

OpenRA.Mods.Common/Traits/Armament.cs
chrisforbes: Yes please!

OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
RoosterDragon: Whilst you're here - moving the `!target.IsValidFor(self)` check before the `!HasAnyValidWeapons(target)` check would also be worthwhile as the first is cheaper than the second.

OpenRA.Mods.Common/Traits/Armament.cs
reaperrr: I wasn't sure whether that could cause any regressions. If you say it won't, I'll unlazy them, too.

RoosterDragon: It won't.

The reason they're lazy is because the traits they're looking up may not exist in the constructor (trait initialization order and whatnot) - so by making them lazy we can avoid doing the lookup until later when they actually exist.

As long as you do the lookup in `INotifyCreated` then you'll be fine - as all the traits will have been fully constructed by that point in time.

reaperrr:
Updated.

RoosterDragon:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ff860ccfbc09b81ff27474339329c93b2e5e3a54...40c248d9126f89c303cbf2696d4274f60df7984a)

PR #10191, opened 2015-12-07
Add support for a FallbackSequence in WithDeathAnimation

abcdefg30:
Used when the actor is killed by non-standard means, like suicide.
Split from #9350.

### DIFF
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -301,6 +301,7 @@ namespace OpenRA.Traits
 
 	public interface IFacingInfo : ITraitInfoInterface { int GetInitialFacing(); }
 
+	[RequireExplicitImplementation]
 	public interface ICrushable
 	{
 		bool CrushableBy(HashSet<string> crushClasses, Player owner);
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -303,9 +303,14 @@ namespace OpenRA.Traits
 
 	public interface ICrushable
 	{
-		void OnCrush(Actor crusher);
-		void WarnCrush(Actor crusher);
 		bool CrushableBy(HashSet<string> crushClasses, Player owner);
+	}
+
+	[RequireExplicitImplementation]
+	public interface INotifyCrushed
+	{
+		void OnCrush(Actor self, Actor crusher, HashSet<string> crushClasses);
+		void WarnCrush(Actor self, Actor crusher, HashSet<string> crushClasses);
 	}
 
 	public interface ITraitInfoInterface { }
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -304,7 +304,7 @@ namespace OpenRA.Traits
 	[RequireExplicitImplementation]
 	public interface ICrushable
 	{
-		bool CrushableBy(HashSet<string> crushClasses, Player owner);
+		bool CrushableBy(Actor self, Actor crusher, HashSet<string> crushClasses);
 	}
 
 	[RequireExplicitImplementation]
--- a/OpenRA.Mods.Common/Traits/Crushable.cs
+++ b/OpenRA.Mods.Common/Traits/Crushable.cs
@@ -29,7 +29,7 @@ namespace OpenRA.Mods.Common.Traits
 		public object Create(ActorInitializer init) { return new Crushable(init.Self, this); }
 	}
 
-	class Crushable : ICrushable
+	class Crushable : ICrushable, INotifyCrushed
 	{
 		readonly Actor self;
 		readonly CrushableInfo info;
@@ -40,16 +40,23 @@ namespace OpenRA.Mods.Common.Traits
 			this.info = info;
 		}
 
-		public void WarnCrush(Actor crusher)
+		void INotifyCrushed.WarnCrush(Actor self, Actor crusher, HashSet<string> crushClasses)
 		{
+			if (!CrushableBy(crushClasses, crusher.Owner))
+				return;
+
 			var mobile = self.TraitOrDefault<Mobile>();
 			if (mobile != null && self.World.SharedRandom.Next(100) <= info.WarnProbability)
 				mobile.Nudge(self, crusher, true);
 		}
 
-		public void OnCrush(Actor crusher)
+		void INotifyCrushed.OnCrush(Actor self, Actor crusher, HashSet<string> crushClasses)
 		{
+			if (!CrushableBy(crushClasses, crusher.Owner))
+				return;
+
 			Game.Sound.Play(info.CrushSound, crusher.CenterPosition);
+
 			var wda = self.TraitsImplementing<WithDeathAnimation>()
 				.FirstOrDefault(s => s.Info.CrushedSequence != null);
 			if (wda != null)
--- a/OpenRA.Mods.Common/Traits/Crushable.cs
+++ b/OpenRA.Mods.Common/Traits/Crushable.cs
@@ -71,9 +71,9 @@ namespace OpenRA.Mods.Common.Traits
 			self.Kill(crusher);
 		}
 
-		bool ICrushable.CrushableBy(HashSet<string> crushClasses, Player crushOwner)
+		bool ICrushable.CrushableBy(Actor self, Actor crusher, HashSet<string> crushClasses)
 		{
-			return CrushableInner(crushClasses, crushOwner);
+			return CrushableInner(crushClasses, crusher.Owner);
 		}
 
 		bool CrushableInner(HashSet<string> crushClasses, Player crushOwner)
--- a/OpenRA.Mods.Common/Traits/Crushable.cs
+++ b/OpenRA.Mods.Common/Traits/Crushable.cs
@@ -42,7 +42,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		void INotifyCrushed.WarnCrush(Actor self, Actor crusher, HashSet<string> crushClasses)
 		{
-			if (!CrushableBy(crushClasses, crusher.Owner))
+			if (!CrushableInner(crushClasses, crusher.Owner))
 				return;
 
 			var mobile = self.TraitOrDefault<Mobile>();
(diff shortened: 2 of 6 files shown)


OpenRA.Mods.Common/Traits/Crushable.cs
```diff
@@ -34,6 +34,8 @@ class Crushable : ICrushable
 		readonly Actor self;
 		readonly CrushableInfo info;
 
+		public bool Crushed;
```
RoosterDragon: `{ get; private set; }` would probably be sensible.

OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
```diff
@@ -61,20 +64,34 @@ public static object LoadDeathTypes(MiniYaml yaml)
...
 	{
 		public readonly WithDeathAnimationInfo Info;
+		readonly Crushable crushable;
 		readonly RenderSprites rs;
 
 		public WithDeathAnimation(Actor self, WithDeathAnimationInfo info)
 		{
 			Info = info;
 			rs = self.Trait<RenderSprites>();
+			crushable = self.TraitOrDefault<Crushable>();
```
RoosterDragon: You should do this lookup in the `Killed` method, since there is no guarantee we will initialize the crushable trait before this one.

OpenRA.Mods.Common/Traits/Crushable.cs
RoosterDragon: Also, should this be added to the `ICrushable` interface?

abcdefg30:
Updated.

OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
```diff
@@ -71,10 +74,23 @@ public WithDeathAnimation(Actor self, WithDeathAnimationInfo info)
 
 		public void Killed(Actor self, AttackInfo e)
 		{
-			// Killed by some non-standard means. This includes being crushed
-			// by a vehicle (Actors with Crushable trait will spawn CrushedSequence instead).
+			// Actors with Crushable trait will spawn CrushedSequence.
+			var crushable = self.TraitOrDefault<Crushable>();
+			if (crushable != null && crushable.Crushed)
```
Mailaender: The new boolean seems redundant. Why not just implement `ICrushable.OnCrush` here? Also saves the trait lookup.

pchote:
Is anything in particular blocking this from being fixed?

OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
abcdefg30: Unfortunately this (nice in theory) solution won't work out:
If I use this approach, I'd need to have `ICrushable.CrushableBy` `return true` here,
as otherwise it would block out every crush attempt. But when I `return true` and `WithDeathAnimation` is the only trait that implements `ICrushable` on the actor, the actor can be "crushed" by everything suddenly

abcdefg30:
https://github.com/OpenRA/OpenRA/pull/10191#discussion_r49999157.

OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
pchote: You can fix this by splitting `INotifyCrushed` into its own interface.

abcdefg30:
Updated.

reaperrr:
> OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs:L87: [SA1513] Statements or elements wrapped in curly brackets must be followed by a blank line.

abcdefg30:
Fixed.

Mailaender:
Crushing still works :white_check_mark: and doing the checks in `OnWarn` individually in each trait instead of not firing the interface is also a nice cleanup. :+1:

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -296,11 +296,15 @@ public interface IFacingInfo : ITraitInfoInterface { int GetInitialFacing(); }
 
 	public interface ICrushable
 	{
-		void OnCrush(Actor crusher);
-		void WarnCrush(Actor crusher);
 		bool CrushableBy(HashSet<string> crushClasses, Player owner);
```
pchote: It might be worth changing this to `bool CrushableBy(Actor self, Actor crusher, HashSet<string> crushClasses);` to match `INotifyCrushed`.

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -296,11 +296,15 @@ public interface IFacingInfo : ITraitInfoInterface { int GetInitialFacing(); }
...
 	}
 
+	public interface INotifyCrushed
+	{
+		void OnCrush(Actor crusher, HashSet<string> crushClasses);
```
pchote: Can we get an `Actor self` parameter for consistency with our other interfaces?

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -296,11 +296,15 @@ public interface IFacingInfo : ITraitInfoInterface { int GetInitialFacing(); }
 
 	public interface ICrushable
```
pchote: You've already made all the implementations explicit, so can you please add `[RequireExplicitImplementation]` so they stay that way?

OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
```diff
@@ -70,10 +74,33 @@ public WithDeathAnimation(Actor self, WithDeathAnimationInfo info)
 
 		public void Killed(Actor self, AttackInfo e)
 		{
-			// Killed by some non-standard means. This includes being crushed
-			// by a vehicle (Actors with Crushable trait will spawn CrushedSequence instead).
+			// Actors with Crushable trait will spawn CrushedSequence.
+			if (crushed)
+			{
+				if (Info.CrushedSequence != null)
```
pchote: You could move this into the OnCrush handler, and then just have `if (crushed) return;`

OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
```diff
@@ -70,10 +74,33 @@ public WithDeathAnimation(Actor self, WithDeathAnimationInfo info)
...
+				}
+
+				return;
+			}
+
+			var palette = Info.DeathSequencePalette;
+			if (Info.DeathPaletteIsPlayerPalette)
+				palette += self.Owner.InternalName;
+
+			// Killed by some non-standard means. This includes being crushed by a vehicle.
```
pchote: This no longer includes being crushed by a vehicle.

OpenRA.Mods.RA/Traits/Mine.cs
```diff
@@ -48,7 +51,12 @@ public void OnCrush(Actor crusher)
 			self.Kill(crusher);
 		}
 
-		public bool CrushableBy(HashSet<string> crushClasses, Player owner)
+		bool ICrushable.CrushableBy(HashSet<string> crushClasses, Player owner)
+		{
+			return CrushableInner(crushClasses);
+		}
+
+		bool CrushableInner(HashSet<string> crushClasses)
```
pchote: I'm not convinced you gain anything by pulling this out into its own function.

abcdefg30:
Updated.

obrakmann:
The results are quite gruesome: =)

openra-2016-01-28t205154z

OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
```diff
@@ -101,5 +113,21 @@ public void SpawnDeathAnimation(Actor self, string sequence, string palette)
 					w.Add(new Corpse(w, self.CenterPosition, rs.GetImage(self), sequence, palette));
 			});
 		}
+
+		void INotifyCrushed.OnCrush(Actor self, Actor crusher, HashSet<string> crushClasses)
+		{
+			crushed = true;
+
+			if (Info.CrushedSequence == null)
+				return;
```
obrakmann: The problem seems to be that 

1. `Mobile.FinishedMoving` doesn't take actors with `SharedCells` into account when calling `OnCrush` (not sure if it should).
2. there is no check equivalent to `Crushable.CrushableBy` happening here.

Also, *if* `Mobile.FinishedMoving` actually calls `OnCrush` that often (ie. every time infantry units share cell), just setting `crushed` here will cause the wrong death animation to play if the unit is later shot, for example.

abcdefg30:
Should be fixed now.

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -590,10 +590,19 @@ public void EnteringCell(Actor self)
 			if (self.CenterPosition.Z != 0)
 				return;
 
-			var crushables = self.World.ActorMap.GetActorsAt(ToCell).Where(a => a != self)
-				.SelectMany(a => a.TraitsImplementing<ICrushable>().Where(b => b.CrushableBy(Info.Crushes, self.Owner)));
-			foreach (var crushable in crushables)
-				crushable.WarnCrush(self);
+			var actors = self.World.ActorMap.GetActorsAt(ToCell).Where(a => a != self);
```
RoosterDragon: You'll want to `ToList` this to avoid enumerating it twice,

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -590,10 +590,19 @@ public void EnteringCell(Actor self)
...
+			if (crushables.Count == 0)
+				return;
+
+			foreach (var crushes in crushables)
+				if (!crushes.Trait.CrushableBy(crushes.Actor, self, Info.Crushes))
+					return;
+
+			var notifiers = actors.SelectMany(a => a.TraitsImplementing<INotifyCrushed>().Select(t => new TraitPair<INotifyCrushed>(a, t)));
+			foreach (var notifyCrushed in notifiers)
+				notifyCrushed.Trait.WarnCrush(notifyCrushed.Actor, self, Info.Crushes);
```
RoosterDragon: You can extract common functions since most of the code in this method is shared with `FinishedMoving`.

abcdefg30:
Fixed.

obrakmann:
This works as advertised now, :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/62a21f349f42f08bd7c64cdd812270e3786e931c...659c4874761cc01b2650451107d5bced550156b6)

PR #10193, opened 2015-12-08
TraitInfoInterface base for interfaces implemented by trait infos

atlimit8:
Neither @pchote nor myself liked using `ITraitInfo` for both trait infos and interfaces (for queries) implemented by trait infos.

### DIFF
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -130,7 +130,7 @@ namespace OpenRA.Traits
 
 	public interface ISeedableResource { void Seed(Actor self); }
 
-	public interface ISelectionDecorationsInfo : ITraitInfo
+	public interface ISelectionDecorationsInfo : ITraitInfoInterface
 	{
 		int[] SelectionBoxBounds { get; }
 	}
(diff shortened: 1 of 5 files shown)


OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -294,12 +294,13 @@ public interface ICrushable
 		bool CrushableBy(HashSet<string> crushClasses, Player owner);
 	}
 
-	public interface ITraitInfo { object Create(ActorInitializer init); }
+	public interface ITraitInfoInterface { }
+	public interface ITraitInfo : ITraitInfoInterface { object Create(ActorInitializer init); }
```
penev92: This inheritance here seems to defeat the whole purpose of this excercise.

pchote:
Looks good to me, thanks :+1:

Phrohdoh:
I'm with @penev92 and would like to see why / the usage for this.

pchote:
It clearly distinguishes ITraitInfo (i.e. the trait info implementations) vs the IFooInfo interfaces that they can implement while not breaking everything involved with type constraints.

Phrohdoh:
:+1: since this is for clearing up intent of code.

PR #10201, opened 2015-12-11
Optionally discard resources when silo capacity has been reached

teees:


Mailaender:
Both cases work as promised. :white_check_mark:

abcdefg30:
:+1: / :white_check_mark: 

(Although think this behaviour is not fitting, but we want to stick to the original...)

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/4a8bbe5d5730b62d3f7adfb1d758988bf87de5aa...f2e8763a2318c476c0a3eccd0591afabe2f4cbd5)

pchote:
Balancing can come later.

Mailaender:
It serves as a good reference/test case for now.

PR #10208, opened 2015-12-12
Added some performance comments

RoosterDragon:
I've added some comments to help explain some parts of the code in performance sensitive areas - as well as to call out areas where the code needs to be fast to avoid people accidentally changing loops into LINQ and such in hot paths. I've prefixed such comments with `PERF:` for searchability.

The PR also includes a few commits with some minor cleanup.

### DIFF
--- a/OpenRA.Game/Actor.cs
+++ b/OpenRA.Game/Actor.cs
@@ -97,6 +97,9 @@ namespace OpenRA
 				}
 			}
 
+			// PERF: Cache all these traits as soon as the actor is created. This is a fairly cheap one-off cost per
+			// actor that allows us to provide some fast implementations of commonly used methods that are relied on by
+			// performance-sensitive parts of the core game engine, such as pathfinding, visibility and rendering.
 			Bounds = DetermineBounds();
 			VisualBounds = DetermineVisualBounds();
 			EffectiveOwner = TraitOrDefault<IEffectiveOwner>();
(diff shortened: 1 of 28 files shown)


Mailaender:
:+1:

OpenRA.Game/Actor.cs
```diff
@@ -206,6 +217,7 @@ public bool Equals(Actor other)
 
 		public override string ToString()
 		{
+			// PERF: Avoid format strings.
```
Phrohdoh: Question: is `string.Concat` any faster than `+` concatenation?

OpenRA.Game/Renderer.cs
```diff
@@ -111,6 +111,7 @@ public void BeginFrame(int2 scroll, float zoom)
 
 		public void SetViewportParams(int2 scroll, float zoom)
 		{
+			// PERF: SetViewportParams is slow. Only call it when things change.
```
Phrohdoh: This should go on the line above the method declaration.

RoosterDragon: It applies to the individual `*Render.SetViewportParams` calls, not the method itself

OpenRA.Game/Actor.cs
RoosterDragon: The compiler turns `+` into calls to `string.Concat` for you, handily.

OpenRA.Game/Renderer.cs
Phrohdoh: I see.
Yes please, though I do not know how to do so.

RoosterDragon: Updated a bit.

Phrohdoh:
:+1: looks good to me.

PR #10210, opened 2015-12-13
Add enter behaviour field to Infiltrates/EngineerRepair/RepairsBridges.

pchote:
Followup to #10092.

### DIFF
--- a/OpenRA.Mods.Common/Activities/RepairBridge.cs
+++ b/OpenRA.Mods.Common/Activities/RepairBridge.cs
@@ -18,7 +18,7 @@ namespace OpenRA.Mods.Common.Activities
 		readonly BridgeHut hut;
 
 		public RepairBridge(Actor self, Actor target)
-			: base(self, target)
+			: base(self, target, EnterBehaviour.Dispose)
 		{
 			hut = target.Trait<BridgeHut>();
 		}
--- a/OpenRA.Mods.Common/Activities/RepairBridge.cs
+++ b/OpenRA.Mods.Common/Activities/RepairBridge.cs
@@ -17,8 +17,8 @@ namespace OpenRA.Mods.Common.Activities
 	{
 		readonly BridgeHut hut;
 
-		public RepairBridge(Actor self, Actor target)
-			: base(self, target, EnterBehaviour.Dispose)
+		public RepairBridge(Actor self, Actor target, EnterBehaviour enterBehaviour)
+			: base(self, target, enterBehaviour)
 		{
 			hut = target.Trait<BridgeHut>();
 		}
@@ -32,8 +32,8 @@ namespace OpenRA.Mods.Common.Activities
 		{
 			if (hut.BridgeDamageState == DamageState.Undamaged || hut.Repairing || hut.Bridge.GetHut(0) == null || hut.Bridge.GetHut(1) == null)
 				return;
+
 			hut.Repair(self);
-			self.Dispose();
 		}
 	}
 }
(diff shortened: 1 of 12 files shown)


OpenRA.Mods.RA/Activities/Infiltrate.cs
```diff
@@ -48,10 +48,14 @@ protected override void OnInside(Actor self)
...
 
-			self.Dispose();
+			if (enterBehaviour == EnterBehaviour.Suicide)
+				self.Kill(self);
+			else if (enterBehaviour == EnterBehaviour.Dispose)
+				self.Dispose();
 
 			if (target.Info.HasTraitInfo<BuildingInfo>())
-				Game.Sound.PlayToPlayer(self.Owner, "bldginf1.aud");
+				Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Speech",
```
penev92: This is no longer tied to the building infiltration file, so perhaps this check should be moved to the infiltration trait, and if a notification shouldn't be played, pass null instead of the file name to the activity's constructor.

pchote:
Updated.

abcdefg30:
Needs a rebase.

pchote:
Rebased.

OpenRA.Mods.Common/Traits/RepairsBridges.cs
```diff
@@ -21,6 +21,10 @@ class RepairsBridgesInfo : ITraitInfo
 	{
 		[VoiceReference] public readonly string Voice = "Action";
 
+		[Desc("Behaviour when entering the structure.",
+			"Possible values are Exit, Suicide, Dispose.")]
+		public readonly EnterBehaviour EnterBehaviour = EnterBehaviour.Exit;
```
Mailaender: This wants to be `Dispose` to retain current behavior.

OpenRA.Mods.Common/Traits/EngineerRepair.cs
```diff
@@ -21,6 +21,10 @@ class EngineerRepairInfo : ITraitInfo
 	{
 		[VoiceReference] public readonly string Voice = "Action";
 
+		[Desc("Behaviour when entering the structure.",
+			"Possible values are Exit, Suicide, Dispose.")]
+		public readonly EnterBehaviour EnterBehaviour = EnterBehaviour.Exit;
```
Mailaender: This wants to be `Dispose` to retain current behavior.

Mailaender:
Otherwise okay. :+1:

pchote:
Fixed.

OpenRA.Mods.Common/Activities/RepairBridge.cs
```diff
@@ -32,8 +34,13 @@ protected override void OnInside(Actor self)
...
 			if (hut.BridgeDamageState == DamageState.Undamaged || hut.Repairing || hut.Bridge.GetHut(0) == null || hut.Bridge.GetHut(1) == null)
 				return;
+
 			hut.Repair(self);
-			self.Dispose();
+
+			if (enterBehaviour == EnterBehaviour.Suicide)
+				self.Kill(self);
+			else if (enterBehaviour == EnterBehaviour.Dispose)
+				self.Dispose();
```
obrakmann: Would it make sense to move that block into a method in the `Enter` activity? It's copied quite a few times, and afaics all those other activities inherit from `Enter`.

obrakmann:
:+1: after the duplication has been removed.

pchote:
Updated.

RoosterDragon:
:+1:

Mailaender:
Nice cleanup. Also well spotted on that hard-coded `bldginf1.aud` notification. :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/43c5c603689917b18357eddb12e4638e7f4fe602...8d2fdaf90a52a8df052432a53e3608c9a876b838)

PR #10215, opened 2015-12-13
Change AllowInvalidPlacement rules

RoosterDragon:
Instead of allowing placement of a building literally anywhere (including miles off the map area), we now ensure at least one cell of the footprint is on the map.

Currently, only concrete in Dune makes use of `AllowInvalidPlacement` and is the only thing that will be affected in the default mods. This change means you can no longer accidentally place concrete off the map where it has no use.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Buildings/BuildingUtils.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/BuildingUtils.cs
@@ -39,13 +39,14 @@ namespace OpenRA.Mods.Common.Traits
 
 		public static bool CanPlaceBuilding(this World world, string name, BuildingInfo building, CPos topLeft, Actor toIgnore)
 		{
+			var footprint = FootprintUtils.Tiles(world.Map.Rules, name, building, topLeft);
 			if (building.AllowInvalidPlacement)
-				return true;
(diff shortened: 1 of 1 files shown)


penev92:
This is about not being able to place concrete outside the playable area, a.k.a. in the map cordons, correct?
I expect that to not work well with scripted map bound changes. Or am I missing something?

RoosterDragon:
I am suggesting we make it so you can't place stuff outside the bounds even if it has `AllowInvalidPlacement `, yes. If there's some downside to that, then we can leave things as-is. At the moment you can place such things even outside the cordon area.

penev92:
Placing actors in the map cordons is used by some "advanced" maps to give "neutral" factions power that can't be shut down, for example. Or support powers, etc.
Also once we have scripted map bounds changes (which we eventually will want to have), the initial playable area for a lot of missions would be a very small part of the map.

pchote:
In theory this change should only affect `PlaceBuildingOrderGenerator`, but can you check and make that explicit?

RoosterDragon:
I only really wanted to prevent placement outside of even the cordon area - so I think my current code needs to check something else for that.

RoosterDragon:
Even restricting this to any cell on the map doesn't have a great impact, so closing.

PR #10216, opened 2015-12-13
Reimplement Primary and Fake tags using WithDecoration.

pchote:
Followup to #10104.  This removes the legacy `ITags` interface, and instead uses the `WithDecoration` trait for rendering the primary and fake building labels.

I had to first rework `WithDecoration` and the TS EMP implementation to follow our rendering conventions, which makes this a bit more complicated than it otherwise would have been.

This also fixes the UI decorations (rank / spy / upgrade star) pixel doubling with the world.


Mailaender:
This causes several lint errors.

pchote:
Added the missing `[UpgradeGrantedReference]` and fixed typo.

GraionDilach:
I terribly dislike the idea of repurposing `WithDecoration` into a strictly-UI trait and can't see that as a fix.

pchote:
`WithDecoration` *is* a strictly-UI trait - that is literally its only reason for existing.

GraionDilach:
On a second thought, I guess my problem would be better solved with a `WithIdleOverlay` with unhardcoded image instead of `WithDecoration`.

pchote:
Rebased.  Please leave a comment when attaching tags, otherwise nobody can see the update.

Mailaender:
Needs a rebase.

pchote:
Rebased again.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2702,6 +2703,67 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 						node.Value.Nodes.Add(new MiniYamlNode("ParticleSize", "1, 1"));
 				}
 
+				// Overhauled the actor decorations traits
+				if (engineVersion < 20151226)
+				{
+					if (depth == 1 && (node.Key.StartsWith("WithDecoration") || node.Key.StartsWith("WithRankDecoration")))
+					{
+						node.Value.Nodes.RemoveAll(n => n.Key == "Scale");
+						node.Value.Nodes.RemoveAll(n => n.Key == "Offset");
```
Phrohdoh: Is there a reason you aren't `||`ing these together?

pchote: Readability and consistency with the other upgrade rules.

Mailaender:
EMP overlay and primary/fake tags still work. :white_check_mark:

obrakmann:
Looks good to me, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4c4ac5dd153360a69911a50c51314835e5118896)

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2702,6 +2703,67 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+					{
+						var decNodes = new List<MiniYamlNode>();
+						decNodes.Add(new MiniYamlNode("RequiresSelection", "true"));
+						decNodes.Add(new MiniYamlNode("Image", "pips"));
+						decNodes.Add(new MiniYamlNode("Sequence", "tag-primary"));
+						decNodes.Add(new MiniYamlNode("ReferencePoint", "Top"));
+						decNodes.Add(new MiniYamlNode("ZOffset", "256"));
+						decNodes.Add(new MiniYamlNode("UpgradeTypes", "primary"));
+						decNodes.Add(new MiniYamlNode("UpgradeMinEnabledLevel", "1"));
+						node.Value.Nodes.Add(new MiniYamlNode("WithDecoration@primary", new MiniYaml("", decNodes)));
```
Mailaender: Sadly this doesn't seem to fire. Had to manually add it in https://github.com/OpenRA/ra2/commit/37d5707f60c886706a29aab7cfd92131fb246c94

pchote: RA2 was already based on a bleed version newer than 20151226, afaik?  That means the rule would have never run.

Mailaender: Only @abcdefg30 knows about that. The complete git commit history is gone.

abcdefg30: Seems like the last upgrade rule ran only updated the mod to `20151205`: https://github.com/Phrohdoh/OpenRA/commit/d7ded32318c796bf502a354cf7de4b51903bd1f4

PR #10232, opened 2015-12-15
Implement mod loading from the support directory

### LINKED ISSUE #8992: Support mods loaded from the support directory

penev92:
Allows loading of mods from both the game and the support directory (`\<user>\Documents\OpenRA\mods\` in this case)

Also unhardcodes mod logo and preview image paths. (This is a double-edged change because 1) it gives modders more flexibility with these; 2) it gives modders too much flexibility with these. Just saying.)

Also applies the new format to D2k. A follow-up PR will apply it to the rest of the official mods (for consistency).

Closes #8992.

### DIFF
--- a/OpenRA.Game/Manifest.cs
+++ b/OpenRA.Game/Manifest.cs
@@ -154,8 +154,10 @@ namespace OpenRA
 			if (!yaml.ContainsKey(key))
 				return new string[] { };
 
-			var list = yaml[key].ToDictionary().Keys.ToArray();
-			return parsePaths ? list.Select(Platform.ResolvePath).ToArray() : list;
+			if (parsePaths)
+				return yaml[key].Nodes.Select(node => Platform.ResolvePath(node.Key, node.Value.Value ?? string.Empty)).ToArray();
+
+			return yaml[key].ToDictionary().Keys.ToArray();
 		}
 
 		static IReadOnlyDictionary<string, string> YamlDictionary(Dictionary<string, MiniYaml> yaml, string key, bool parsePaths = false)
@@ -166,15 +168,19 @@ namespace OpenRA
 			var inner = new Dictionary<string, string>();
 			foreach (var node in yaml[key].Nodes)
 			{
+				var line = node.Key;
+				if (node.Value.Value != null)
+					line += ":" + node.Value.Value;
+
 				// '@' may be used in mod.yaml to indicate extra information (similar to trait @ tags).
 				// Applies to MapFolders (to indicate System and User directories) and Packages (to indicate package annotation).
-				if (node.Key.Contains('@'))
+				if (line.Contains('@'))
 				{
-					var split = node.Key.Split('@');
-					inner.Add(split[0], split[1]);
+					var split = line.Split('@');
+					inner.Add(parsePaths ? Platform.ResolvePath(split[0]) : split[0], split[1]);
 				}
 				else
-					inner.Add(node.Key, null);
+					inner.Add(line, null);
 			}
 
 			return new ReadOnlyDictionary<string, string>(inner);
--- a/OpenRA.Game/Manifest.cs
+++ b/OpenRA.Game/Manifest.cs
@@ -61,13 +61,16 @@ namespace OpenRA
 		readonly TypeDictionary modules = new TypeDictionary();
 		readonly Dictionary<string, MiniYaml> yaml;
 
-		public Manifest(string mod)
+		public Manifest(string modId, string modPath = null)
 		{
-			var path = Platform.ResolvePath(".", "mods", mod, "mod.yaml");
+			if (modPath == null)
+				modPath = ModMetadata.CandidateModPaths[modId];
+
+			var path = Path.Combine(modPath, "mod.yaml");
 			yaml = new MiniYaml(null, MiniYaml.FromFile(path)).ToDictionary();
 
 			Mod = FieldLoader.Load<ModMetadata>(yaml["Metadata"]);
-			Mod.Id = mod;
+			Mod.Id = modId;
 
 			// TODO: Use fieldloader
 			Folders = YamlList(yaml, "Folders", true);
@@ -202,21 +205,16 @@ namespace OpenRA
 
 		static Dictionary<string, Manifest> LoadMods()
 		{
-			// Get mods that are in the game folder.
-			var basePath = Platform.ResolvePath(".", "mods");
-			var mods = Directory.GetDirectories(basePath)
-				.Select(x => x.Substring(basePath.Length + 1));
-
 			var ret = new Dictionary<string, Manifest>();
-			foreach (var mod in mods)
+			foreach (var mod in ModMetadata.CandidateModPaths)
 			{
-				if (!File.Exists(Platform.ResolvePath(".", "mods", mod, "mod.yaml")))
+				if (!File.Exists(Path.Combine(mod.Value, "mod.yaml")))
 					continue;
 
 				try
 				{
-					var manifest = new Manifest(mod);
-					ret.Add(mod, manifest);
+					var manifest = new Manifest(mod.Key, mod.Value);
+					ret.Add(mod.Key, manifest);
 				}
 				catch (Exception ex)
 				{
--- a/OpenRA.Game/Manifest.cs
+++ b/OpenRA.Game/Manifest.cs
@@ -29,7 +29,7 @@ namespace OpenRA
 		}
 	}
 
-	// Describes what is to be loaded in order to run a mod
+	/// <summary> Describes what is to be loaded in order to run a mod. </summary>
 	public class Manifest
 	{
 		public static readonly Dictionary<string, Manifest> AllMods = LoadMods();
@@ -106,12 +106,10 @@ namespace OpenRA
 			RequiresMods = yaml["RequiresMods"].ToDictionary(my => my.Value);
 
 			// Allow inherited mods to import parent maps.
-			var compat = new List<string>();
-			compat.Add(mod);
+			var compat = new List<string> { Mod.Id };
 
 			if (yaml.ContainsKey("SupportsMapsFrom"))
-				foreach (var c in yaml["SupportsMapsFrom"].Value.Split(','))
-					compat.Add(c.Trim());
+				compat.AddRange(yaml["SupportsMapsFrom"].Value.Split(',').Select(c => c.Trim()));
 
 			MapCompatibility = compat.ToArray();
 
@@ -198,6 +196,7 @@ namespace OpenRA
 
 		static Dictionary<string, Manifest> LoadMods()
 		{
+			// Get mods that are in the game folder.
 			var basePath = Platform.ResolvePath(".", "mods");
 			var mods = Directory.GetDirectories(basePath)
 				.Select(x => x.Substring(basePath.Length + 1));
--- a/OpenRA.Game/Platform.cs
+++ b/OpenRA.Game/Platform.cs
@@ -95,10 +95,18 @@ namespace OpenRA
 
 		public static string GameDir { get { return AppDomain.CurrentDomain.BaseDirectory; } }
 
-		/// <summary>Replace special character prefixes with full paths</summary>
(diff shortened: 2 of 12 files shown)


OpenRA.Game/Platform.cs
```diff
@@ -108,6 +108,13 @@ public static string ResolvePath(string path)
 			if (path.StartsWith("./") || path.StartsWith(".\\"))
 				path = GameDir + path.Substring(2);
 
+			if (path.Contains("@"))
```
pchote: This should use `:` to denote `mod:path` for consistency with our package-specific filenames and our other yaml definitions.  See my explanation in IRC for how to implement this.

pchote:
> Also unhardcodes mod logo and preview image paths. (This is a double-edged change because 1) it gives modders more flexibility with these; 2) it gives modders too much flexibility with these. Just saying.)

Wouldn't it be better to keep the current restriction and just make the mod browser look in the mod's root folder?

atlimit8:
My [comment](https://github.com/OpenRA/OpenRA/issues/8992#issuecomment-143401249) from #8992:
> `:` suffix looks like a key-value pair.
> Perhaps something like `$`*mod-id*`/`*path* would be more distinguishable.
> The (UNIX) shell environment variable feel is intentional.

These require something other than a `:` separator to avoid MiniYaml parsing issues:
 * https://github.com/penev92/OpenRA/blob/e0dfa7eaddcb32cd3f7e3bdef3a47dc86b7b0d56/mods/all/mod.yaml#L20
 * https://github.com/penev92/OpenRA/blob/e0dfa7eaddcb32cd3f7e3bdef3a47dc86b7b0d56/mods/d2k/mod.yaml
   * Especially https://github.com/penev92/OpenRA/blob/e0dfa7eaddcb32cd3f7e3bdef3a47dc86b7b0d56/mods/d2k/mod.yaml#L25 which has a `:` for MiniYaml key-value parsing.

OpenRA.Game/Platform.cs
```diff
@@ -108,6 +108,13 @@ public static string ResolvePath(string path)
...
+			{
+				var mod = path.Substring(0, path.IndexOf('@'));
+				var file = path.Contains("@ ") ? path.Substring(path.IndexOf("@ ") + 2) : string.Empty;
+				path = Path.Combine(ModMetadata.CandidateModPaths[mod], file);
+			}
```
atlimit8: This would be:
```c#
			if (path.StartsWith("$"))
			{
				var mod = path.Substring(1, path.IndexOf('/'));
				var file = path.Contains("/") ? path.Substring(path.IndexOf("/") + 1) : string.Empty;
				path = Path.Combine(ModMetadata.CandidateModPaths[mod], file);
			}
```

Phrohdoh:
I very much disagree with the $ prefix as these are not sh variables and the last thing we should do, IMO, is to introduce more symbols in the yaml parser.

atlimit8:
The point is that it is not part of the MiniYaml parser. I would not be opposed to using:
```c#
if (!path.StartsWith(".") && !path.StartsWith("^"))
			if (!path.StartsWith(".") && !path.StartsWith("^"))
			{
				var mod = path.Substring(0, path.IndexOf('/'));
				var file = path.Contains("/") ? path.Substring(path.IndexOf("/") + 1) : string.Empty;
				path = Path.Combine(ModMetadata.CandidateModPaths[mod], file);
			}
```
Then, there would be no prefix or separator required for mods.

pchote:
> These require something other than a : separator to avoid MiniYaml parsing issues:

These have already been discussed: @ can be used for the map dir type and package hash type suffixes.

Like I mentioned above, we already support and use this syntax (`package-to-load-from:path/within/the.package`) in the filesystem code, so I don't buy these arguments for inventing a new and inconsistent syntax that is only used here.

atlimit8:
Okay on not using `$`.
However, that is also an argument against `package-to-load-from:path/within/the.package`; supporting a no prefix solution (with `package-to-load-from:path/within/the.package` deprecated).
 * `.` - Game dir
 * `^` - Support dir
 * *no prefix* - `mod-id/path`

I concede that a prefix or separator other than `:` is not needed since it can be done without either.

atlimit8:
More importantly, `:` can not be used in MiniYaml keys.

penev92:
This doesn't seem to be going anywhere.
I don't agree with the reasoning behind "use ':'", I don't like how it looks in the YAML, and I definitely don't like the engine hacks required to make it work (and yes, I did try; I lost a few hours and a lot of nerves already).
My thoughs from IRC, to sum it up:
>[04:17] < penev> this whole : thing is complete idiocity
[04:18] < penev> "it has a nice ring to it, so let't abandon all aesthetic and syntax YAML conventions and spam the code with hacks"
...basically the fact that you're loading it as YAML doesn't play well with the fact that you want the ':' as a special character

The last line is pretty much what @atlimit8 said in his last comment.
Also:

>Like I mentioned above, we already support and use this syntax (`package-to-load-from:path/within/the.package`) in the filesystem code

Not really (on the "support" part). That is a big hack right there. One that I'm not willing to duplicate.


So unless someone has actually constructive feedback (how to implement it using ':' without godawful engine hacks **OR** to say this is good and why), I'm going to close this.

And just for the record, because it'd be almost impossible to find in the IRC logs: the two big arguments for ':' have been 1. "we already use ':' for stuff, let's use it for more stuff" and 2."we already use ':' for packages so it's the same thing".
1. Using it for even more (unrelated to the previous usages) stuff is *not* a good idea in my opinion.
2. No, it is not the same. Mods being packages is an implementation detail (one we don't even support yet). 

All that said, I'd be interested to see a suggestion on how to implement it without godawful engine hacks (or to see an implementation without such).

Phrohdoh:
A tale as old as time (suddenly comes back in this PR):

```
Available mods:
        all: All mods ({DEV_VERSION})
        cnc: Tiberian Dawn ({DEV_VERSION})
        d2k: Dune 2000 ({DEV_VERSION})
        modchooser: Mod Chooser ({DEV_VERSION})
        ra: Red Alert ({DEV_VERSION})
        ts: Tiberian Sun ({DEV_VERSION})
        ra2: Red Alert 2 ({DEV_VERSION})
Loading mod: ra2
Exception of type `System.InvalidOperationException`: Unit `shroud` does not have a sequence named `shroud`
  at OpenRA.Graphics.SequenceProvider.GetSequence (System.String unitName, System.String sequenceName) [0x0006b] in /Users/thill/projects/openra/OpenRA.Game//Graphics/SequenceProvider.cs:68
  at OpenRA.Mods.Common.Traits.ShroudRenderer..ctor (OpenRA.World world, OpenRA.Mods.Common.Traits.ShroudRendererInfo info) [0x00144] in /Users/thill/projects/openra/OpenRA.Mods.Common//Traits/World/ShroudRenderer.cs:128
  at OpenRA.Mods.Common.Traits.ShroudRendererInfo.Create (OpenRA.ActorInitializer init) [0x00008] in /Users/thill/projects/openra/OpenRA.Mods.Common//Traits/World/ShroudRenderer.cs:48
  at OpenRA.Actor..ctor (OpenRA.World world, System.String name, OpenRA.Primitives.TypeDictionary initDict) [0x000b2] in /Users/thill/projects/openra/OpenRA.Game//Actor.cs:91
  at OpenRA.World.CreateActor (Boolean addToWorld, System.String name, OpenRA.Primitives.TypeDictionary initDict) [0x00004] in /Users/thill/projects/openra/OpenRA.Game//World.cs:226
  at OpenRA.World.CreateActor (System.String name, OpenRA.Primitives.TypeDictionary initDict) [0x00005] in /Users/thill/projects/openra/OpenRA.Game//World.cs:221
  at OpenRA.World..ctor (OpenRA.Map map, OpenRA.Network.OrderManager orderManager, WorldType type) [0x00146] in /Users/thill/projects/openra/OpenRA.Game//World.cs:164
  at OpenRA.Game.StartGame (System.String mapUID, WorldType type) [0x00074] in /Users/thill/projects/openra/OpenRA.Game//Game.cs:153
  at OpenRA.Game.LoadShellMap () [0x00019] in /Users/thill/projects/openra/OpenRA.Game//Game.cs:397
  at OpenRA.Mods.Common.LoadScreens.BlankLoadScreen.StartGame (OpenRA.Arguments args) [0x0012c] in /Users/thill/projects/openra/OpenRA.Mods.Common//LoadScreens/BlankLoadScreen.cs:75
  at OpenRA.Game.InitializeMod (System.String mod, OpenRA.Arguments args) [0x00374] in /Users/thill/projects/openra/OpenRA.Game//Game.cs:353
  at OpenRA.Game.Initialize (OpenRA.Arguments args) [0x0026d] in /Users/thill/projects/openra/OpenRA.Game//Game.cs:248
  at OpenRA.Program.Run (System.String[] args) [0x00007] in /Users/thill/projects/openra/OpenRA.Game//Support/Program.cs:115
  at OpenRA.Program.Main (System.String[] args) [0x00050] in /Users/thill/projects/openra/OpenRA.Game//Support/Program.cs:39
```

penev92:
Can't help it if it's not in the official mods.

Will fix travis later.

Phrohdoh:
This was just a case of cleaning and rebuilding, odd.
TS2 (copied mods/ts/ to ^/mods/ts) worked, so :+1:.

OpenRA.Game/Manifest.cs
```diff
@@ -163,15 +166,19 @@ static string[] YamlList(Dictionary<string, MiniYaml> yaml, string key, bool par
 			var inner = new Dictionary<string, string>();
 			foreach (var node in yaml[key].Nodes)
 			{
+				var line = node.Key;
+				if (node.Value.Value != null)
+					line += ":" + node.Value.Value;
```
Phrohdoh: Why are you having to add : here and combining the key and value?

penev92:
Fixed Travis and rebased.

Mailaender:
Tested this with Red Alert 2: https://github.com/OpenRA/ra2/pull/28. Even `mono --debug OpenRA.Utility.exe ra2 --upgrade-mod 20160101` worked flawlessly. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/b4384aa5d000b87650fb1bdec2b14bc64a987ad1...fa6f8a6676982a8a8ccd8669617cefcc30c70628)

PR #10235, opened 2015-12-15
Replace INotifyBuildComplete with upgrade system (part 1)

reaperrr:
Provided the general approach gets approved, this PR will be the first of two or more to get rid of the (hacky) `INotifyBuildComplete` interface, disabling actors/traits during make animation via upgrades instead.

This PR gives `WithMakeAnimation` the ability to provide upgrades, migrates `WithBuildingPlacedAnimation` (playing when TD and RA construction yards place a structure) and `WithSiloAnimation` from `INotifyBuildComplete` to upgrades, and makes the `DisabledOverlay` trait upgradable itself, because we don't want it to be enabled while the make animation runs.

I intentionally kept the number of migrated traits small in this PR, for easier reviewing of the general approach.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Modifiers/DisabledOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Modifiers/DisabledOverlay.cs
@@ -15,13 +15,19 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Use together with CanPowerDown/RequiresPower on buildings or Husk for vehicles.")]
-	public class DisabledOverlayInfo : TraitInfo<DisabledOverlay> { }
-
-	public class DisabledOverlay : IRenderModifier
+	public class DisabledOverlayInfo : UpgradableTraitInfo
 	{
+		public override object Create(ActorInitializer init) { return new DisabledOverlay(init, this); }
+	}
+
+	public class DisabledOverlay : UpgradableTrait<DisabledOverlayInfo>, IRenderModifier
+	{
+		public DisabledOverlay(ActorInitializer init, DisabledOverlayInfo info)
+			: base(info) { }
+
 		public IEnumerable<IRenderable> ModifyRender(Actor self, WorldRenderer wr, IEnumerable<IRenderable> r)
 		{
-			if (!self.IsDisabled())
+			if (IsTraitDisabled || !self.IsDisabled())
 				return r;
 
 			return ModifiedRender(self, wr, r);
--- a/OpenRA.Mods.Common/Traits/Render/WithMakeAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithMakeAnimation.cs
@@ -18,39 +18,70 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Replaces the sprite during construction.")]
-	public class WithMakeAnimationInfo : ITraitInfo, Requires<WithSpriteBodyInfo>
+	public class WithMakeAnimationInfo : ITraitInfo, Requires<WithSpriteBodyInfo>, Requires<UpgradeManagerInfo>
 	{
-		[Desc("Sequence name to use")]
+		[Desc("Sequence name to use.")]
 		[SequenceReference] public readonly string Sequence = "make";
+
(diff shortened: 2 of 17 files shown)


OpenRA.Mods.Common/Traits/Render/WithMakeAnimation.cs
```diff
@@ -18,39 +18,70 @@
...
+			}
+			else if (building == null)
+			{
+				foreach (var up in info.MakeUpgrades)
+					manager.GrantUpgrade(self, up, this);
+
+				wsb.PlayCustomAnimation(self, info.Sequence, () =>
+				{
+					foreach (var up in info.MakeUpgrades)
+						manager.RevokeUpgrade(self, up, this);
```
Phrohdoh: Since these two lines are written multiple times could you extract them out into appropriately named methods?

mods/cnc/rules/defaults.yaml
```diff
@@ -480,10 +480,13 @@
 		SellSounds: cashturn.aud
 	Capturable:
 	WithMakeAnimation:
+		MakeUpgrades: underConstruction
+	DisableOnUpgrade:
+		UpgradeTypes: underConstruction
```
Phrohdoh: I don't think we use camel case like this for any other upgrades.

OpenRA.Mods.Common/Traits/Modifiers/DisabledOverlay.cs
```diff
@@ -15,13 +15,19 @@
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Use together with CanPowerDown/RequiresPower on buildings or Husk for vehicles.")]
-	public class DisabledOverlayInfo : TraitInfo<DisabledOverlay> { }
+	public class DisabledOverlayInfo : UpgradableTraitInfo
```
pchote: We already have `UpgradeOverlay` trait for doing this with upgrades.  This trait is a legacy/deprecated leftover for working with the old `IsDisabled` code (which IMO may need to be cleaned up before `INotifyBuildComplete`).

OpenRA.Mods.Common/Traits/Render/WithMakeAnimation.cs
pchote: See DeployToUpgrade for an example of this.

mods/cnc/rules/defaults.yaml
penev92: Only because all others are single words.

ScottNZ: We have yaml strings which are fully lowercase, even for multiple words. I haven't seen any example of camelcase usage in the repository, but I haven't been reading it much lately. I'd rather we didn't introduce yet another capitalisation scheme to the yaml.

penev92: Going for "underconstructionupgradegrantedwhilebuildingisnotready" would be even worse though.

reaperrr: > We have yaml strings which are fully lowercase, even for multiple words.

I know, but as penev pointed out, it can get hard to read with all lower-case, though I guess using an underscore instead might be the better solution here.

penev92: camelCase is much better than lowercase_but_with_underscores. Just saying.

reaperrr:
> (...) the old `IsDisabled` code (which IMO should be cleaned up before we start on `INotifyBuildComplete`).

After discussing this on IRC, I'll tackle `IsDisabled` first and get back to this afterwards.

mods/cnc/rules/defaults.yaml
ScottNZ: I think it's much worse when it involves adding yet another inconsistency to already horribly inconsistent files.

penev92: In that case
>we should *fix* what already exists if it's not good enough to reuse 

Anyway, at this point does it really hurt to add *yet another one*?

Phrohdoh: Yes, that has been our whole argument.
Adding more and more inconsistency is bad, no matter how slowly.

reaperrr:
Closing this for now as it'll probably take a while until all dependencies are resolved.

PR #10236, opened 2015-12-16
Set ideal aircraft separation to half of the default value

teees:
before: https://streamable.com/8byl
after: https://streamable.com/z41e


RoosterDragon:
This makes the other, larger, aircraft uncomfortably close. Maybe you should refine this to just affect the smaller ones?

Mailaender:
Also needs a rebase.

teees:
rebased and changed

Mailaender:
This matches their visual sizes way better. :+1:

RoosterDragon:
:+1:

PR #10238, opened 2015-12-16
Fix DamagedWithoutFoundation causing more damage than it should

abcdefg30:
to its damaged animation.

~~Unfortunately I didn't find a better way to fix the issue than two `RunAfterTick`s.
If we don't want to use that, I can remove the first commit, as the second one will hide the issue.~~

EDIT: Outdated description.

### DIFF
--- a/OpenRA.Mods.D2k/Traits/Buildings/DamagedWithoutFoundation.cs
+++ b/OpenRA.Mods.D2k/Traits/Buildings/DamagedWithoutFoundation.cs
@@ -64,6 +64,8 @@ namespace OpenRA.Mods.D2k.Traits
 			var delta = health.HP - damageThreshold;
 			if (delta > 0)
 				health.InflictDamage(self, self.World.WorldActor, delta, null, false);
+
+			damageTicks = info.WeaponInfo.ReloadDelay;
 		}
 
 		public void Tick(Actor self)
--- a/OpenRA.Mods.D2k/Traits/Buildings/DamagedWithoutFoundation.cs
+++ b/OpenRA.Mods.D2k/Traits/Buildings/DamagedWithoutFoundation.cs
@@ -19,9 +19,13 @@ namespace OpenRA.Mods.D2k.Traits
 	[Desc("Reduces health points over time when the actor is placed on unsafe terrain.")]
 	class DamagedWithoutFoundationInfo : ITraitInfo, IRulesetLoaded, Requires<HealthInfo>
 	{
-		[WeaponReference]
+		[WeaponReference, Desc("The weapon to use for causing damage.")]
 		public readonly string Weapon = "weathering";
+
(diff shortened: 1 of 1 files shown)


pchote:
It looks like the nested RunAfterTicks is simply working around some other code which uses only one RunAfterTick that otherwise overrides the animation state. Can you work out what the other code is, and fix that instead?

abcdefg30:
Updated.
After some more debugging I found the real issue (and fixed it):
`Tick` already decreased the health in addition to the decrement in `AddedToWorld`.

I still wonder why the animation wasn't changed to "damaged-" in the first place, though.

pchote:
Confirmed that this fixes the initial flicker when repairing a newly placed structure.

There is another issue however: the damage warhead always applies a fixed amount of damage, so it can still push the building below the damage threshold if it is repaired slightly and then allowed to decay again.

OpenRA.Mods.Common/Traits/Buildings/RepairableBuilding.cs
```diff
@@ -125,7 +125,6 @@ public void Tick(Actor self)
 				}
 
 				// Bonus is applied after finding players who can pay
-
```
pchote: This doesn't appear to be related to this pr at all.

OpenRA.Mods.D2k/Traits/Buildings/DamagedWithoutFoundation.cs
```diff
@@ -19,9 +19,13 @@ namespace OpenRA.Mods.D2k.Traits
 	[Desc("Reduces health points over time when the actor is placed on unsafe terrain.")]
 	class DamagedWithoutFoundationInfo : ITraitInfo, IRulesetLoaded, Requires<HealthInfo>
 	{
-		[WeaponReference]
+		[WeaponReference, Desc("The weapon to use for causing damage.")]
 		public readonly string Weapon = "weathering";
+
+		[Desc("Terrain types on which no damage is caused.")]
```
pchote: Please keep documentation updates in their own commit :)

obrakmann:
Well, it works. :+1: after pchote's comments have been addressed.

pchote:
> There is another issue however: the damage warhead always applies a fixed amount of damage, so it can still push the building below the damage threshold if it is repaired slightly and then allowed to decay again.

Note that i'm not asking for this to be fixed now.  That would up the complexity of this pr to the point where we shouldn't take it for the release (but i'd also be okay with too, if you did want to do everything together).

abcdefg30:
Updated.

penev92:
:+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/c7db59d06358df35e399773be30827046b74e742)

PR #10242, opened 2015-12-18
Fix the StartsRevealed option of FrozenUnderFog being ignored

### LINKED ISSUE #10228: Neutral frozen actors are not revealed when starting with explored map
Noticed during stream of playtest-20151213.  May be a duplicate.  Affects tech structures, walls, etc.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29123058-neutral-frozen-actors-are-not-revealed-when-starting-with-explored-map?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
Fixes #10228.

In `FrozenActor`'s ctor, the `NeedsRenderables` property is effectively being set to the value of the actor's `FrozenUnderFogInfo.StartsRevealed` field. Later in the ctor, the `UpdateVisibility` method is called, which ignores and overwrites the previous value. This change ensures that `NeedsRenderables` won't be unset by `UpdateVisibility` if it was already set. It will be reset to false by [this line in `FrozenUnderFog.TickRender`](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs#L167) eventually.

### DIFF
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -100,7 +100,7 @@ namespace OpenRA.Traits
 					Shrouded = false;
 			}
 
-			NeedRenderables = Visible && !wasVisible;
+			NeedRenderables |= Visible && !wasVisible;
 		}
 
 		public void Flash()


pchote:
Looks good to me :+1:

abcdefg30:
:+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/c7db59d06358df35e399773be30827046b74e742...d7ca75fa6d5ac3530bef8235d9f6a86eaf053c68)

PR #10246, opened 2015-12-20
Allow tile templates to define their own palette

### LINKED ISSUE #8885: Add multi palette support for tilesets.
I request the ability to use custom pallets within each tileset, with something like this for example.

	Templates:
		Template@0:
			Palette: isotem.pal

		Template@1:
			Palette: isosno.pal 

This would greatly expand the possibilities of custom terrain for 3rd party mods.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25275125-add-multi-palette-support-for-tilesets?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

penev92:
From the commit message:

> Keep the tileset's palette as default, defined on the tileset, but override it for any tile templates that may want to do so with a palette defined on the template.

Basically does what #8649 did for the resources, as pchote suggested (https://github.com/OpenRA/OpenRA/issues/8885#issuecomment-127011935).

Closes  #8885.

### DIFF
--- a/OpenRA.Game/Graphics/TerrainRenderer.cs
+++ b/OpenRA.Game/Graphics/TerrainRenderer.cs
@@ -9,25 +9,34 @@
 #endregion
 
 using System;
+using System.Collections.Generic;
 using OpenRA.Traits;
 
 namespace OpenRA.Graphics
 {
 	sealed class TerrainRenderer : IDisposable
 	{
-		readonly TerrainSpriteLayer terrain;
+		const string TerrainPalette = "terrain";
+
+		readonly World world;
+		readonly Dictionary<string, TerrainSpriteLayer> spriteLayers = new Dictionary<string, TerrainSpriteLayer>();
 		readonly Theater theater;
 		readonly CellLayer<TerrainTile> mapTiles;
 		readonly CellLayer<byte> mapHeight;
 
 		public TerrainRenderer(World world, WorldRenderer wr)
 		{
+			this.world = world;
 			theater = wr.Theater;
 			mapTiles = world.Map.MapTiles.Value;
 			mapHeight = world.Map.MapHeight.Value;
 
-			terrain = new TerrainSpriteLayer(world, wr, theater.Sheet, BlendMode.Alpha,
-				wr.Palette("terrain"), wr.World.Type != WorldType.Editor);
+			foreach (var template in world.TileSet.Templates)
+			{
+				var palette = template.Value.Palette ?? TerrainPalette;
+				spriteLayers.GetOrAdd(palette, pal =>
+					new TerrainSpriteLayer(world, wr, theater.Sheet, BlendMode.Alpha, wr.Palette(palette), wr.World.Type != WorldType.Editor));
+			}
 
 			foreach (var cell in world.Map.AllCells)
 				UpdateCell(cell);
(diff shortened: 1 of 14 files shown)


OpenRA.Game/Graphics/TerrainRenderer.cs
```diff
@@ -9,25 +9,34 @@
...
 using System;
+using System.Collections.Generic;
 using OpenRA.Traits;
 
 namespace OpenRA.Graphics
 {
 	sealed class TerrainRenderer : IDisposable
 	{
-		readonly TerrainSpriteLayer terrain;
+		const string TerrainPalette = "terrain";
```
pchote: Can we define this in the tileset header instead?  That would keep the default and overrides within the same class.

penev92: Will do when I can. Don't let that stop you from reviewing, people ;)

OpenRA.Game/Graphics/TerrainRenderer.cs
```diff
@@ -38,12 +47,24 @@ public TerrainRenderer(World world, WorldRenderer wr)
...
+			var palette = world.TileSet.Templates[tile.Type].Palette ?? TerrainPalette;
+			var sprite = theater.TileSprite(tile);
+			foreach (var kv in spriteLayers)
+			{
+				if (palette != kv.Key)
+					continue;
+
+				kv.Value.Update(cell, sprite);
+				break;
+			}
```
RoosterDragon: Maybe this is copied code, but why is this a loop? It appears to be doing a `TryGetValue` on `spriteLayers`?

pchote: The original code this was copied from operated over all the layers (clearing stale artwork).  It needs to be doing the same here, too.

OpenRA.Game/Graphics/TerrainRenderer.cs
Mailaender: It would also make it lint-able.

penev92:
Updated.

pchote:
You'll need to adjust the `TerrainTemplatePreviewWidget`s in the editor to use the per-template palettes too.

penev92:
Fixed it.

OpenRA.Mods.Common/Widgets/TerrainTemplatePreviewWidget.cs
```diff
@@ -93,7 +93,7 @@ public override void Draw()
 					var u = gridType == MapGridType.Rectangular ? x : (x - y) / 2f;
 					var v = gridType == MapGridType.Rectangular ? y : (x + y) / 2f;
 					var pos = origin + scale * (new float2(u * ts.Width, (v - 0.5f * tileInfo.Height) * ts.Height) - 0.5f * sprite.Size);
-					Game.Renderer.SpriteRenderer.DrawSprite(sprite, pos, worldRenderer.Palette(Palette), size);
+					Game.Renderer.SpriteRenderer.DrawSprite(sprite, pos, worldRenderer.Palette(Template.Palette ?? Palette), size);
```
pchote: This will force-override any palette overrides that the external code or yaml might set.  Could you set this up with a `GetPalette = t => t.Palette ?? Palette` function instead?

penev92:
Fixed up.

pchote:
Looks good to me now. :+1: after removing the testcase.

penev92:
Removed the test commit.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/fc170a12800b6229637217867fef3355f25aa6ba)

PR #10203, opened 2015-12-12
Rewrite line renderer: Part 2 (remove LineRenderer plus other cleanups)

pchote:
Followup to and depends on #10148 (which appears as the first 10 commits in this pr).

The main visual change here is the ability to draw connected line strips, which fixes the holes that can appear at the corners of rectangles and the beacon triangle.  It also fixes the position of the voxel shadow bounding box in the rendering geometry visualization.

PR #10249, opened 2015-12-20
Replace deprecated OpenGL features with more modern replacements

pchote:
Depends on #10203 (which appears as the first 22 commits in this pr).

After #10203 (which this depends on, and appears as the first 22 commits in this pr) we are left with two legacy GL features that are deprecated and removed from more modern GL specs and GL ES.  This PR replaces them with their more modern equivalents:
- Quads &rarr; Triangles.
- Implicit / magic attribute names &rarr; explicit attribute bindings.

### DIFF
--- a/OpenRA.Platforms.Default/Shader.cs
+++ b/OpenRA.Platforms.Default/Shader.cs
@@ -18,6 +18,9 @@ namespace OpenRA.Platforms.Default
 {
 	class Shader : ThreadAffine, IShader
 	{
+		public const int VertexPosAttributeIndex = 0;
+		public const int TexCoordAttributeIndex = 1;
+
 		readonly Dictionary<string, int> samplers = new Dictionary<string, int>();
 		readonly Dictionary<int, ITexture> textures = new Dictionary<int, ITexture>();
 		readonly int program;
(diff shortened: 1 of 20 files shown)


pchote:
Rebased.

Phrohdoh:
All line types I could find/recall work the exact same.
:+1:

pchote:
Note that this affects all rendering, and not just line types.

OpenRA.Platforms.Default/Shader.cs
```diff
@@ -62,6 +65,12 @@ public Shader(string name)
 			// Assemble program
 			program = GL.CreateProgram();
 			ErrorHandler.CheckGlError();
+
+			GL.BindAttribLocation(program, VertexPosAttributeIndex, "aVertexPosition");
+			ErrorHandler.CheckGlError();
+			GL.BindAttribLocation(program, TexCoordAttributeIndex, "aVertexTexcoord");
```
RoosterDragon: Minor quibble: Should this (and the others) be cased as `aVertexTexCoord`?

RoosterDragon:
:+1: Rendering seems fine to me and I don't notice any severe performance impact on my machine despite all the extra vertices flowing around.

pchote:
Rebased and fixuped.

pchote:
Added a new commit to remove more unused code.

Mailaender:
Also renders fine on Linux. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c08c7182e42f1486b7cb2ad9261ec3d21b702a5a...22d511f739045566aa9ae8e6104510d150c687ea)

PR #10251, opened 2015-12-20
Remove ISpriteBody

abcdefg30:
> 16:19:51 (abcdefg30) why do we even have ISpriteBody when it's only implemented by WithSpriteBody ?
> 16:20:08 (pchote) it was there while we were transitioning to WithSpriteBody
> 16:20:18 (pchote) the old Render*'s used to implement it too
> 16:20:23 (abcdefg30) ah, so basically i can remove it?
> 16:20:31 (pchote) should be able to


penev92:
By the way, I *think* I had bigger plans for that than just a placeholder during the transition, but I can't remember at the moment, so (*shrug*).

Mailaender:
:+1:

atlimit8:
This will reduce unnecessary entries in the TraitDictionary; although, it is just one redundant "container".

PR #10252, opened 2015-12-20
Remove obsolete TransformOnPassenger trait

penev92:
Partially to avoid future confusion like https://github.com/Phrohdoh/OpenRA/pull/84.


Phrohdoh:
Outright removing it may very well be a breaking change.
Does anyone have any ideas for backwards compatibility / making it (visibly) obsolete?

Mailaender:
You will need to write an upgrade rule which does the actor transformation and owner change. Probably not very easy.

pchote:
Looks good to me, and doesn't break anything ingame (because it isn't used) :+1:

As mentioned above, this will need an upgrade rule to notify that this has been removed ("TransformOnPassenger has been removed.  Use the upgrades system to apply modifiers to the transport actor instead") like we do for other changes that cannot be automatically converted.

penev92:
Added an upgrade rule and rebased.

Mailaender:
But the upgrade system can't do a transform yet afaik, so it isn't really an alternative.

pchote:
That's kind of the point – transforming is an anti-feature that we do not want to support.  This trait has only ever been used as a hacky garrisoning logic, and was completely superceded for our purposes by the combination of `AttackGarrisoned` and upgrades.  Do you have a specific different usecase in mind that this PR breaks?

GraionDilach:
The sole reason I can see `TransformOnPassenger` being viable is that the "passengers convert an actor to their side via entering" option doesn't exist yet AFAIK.

I do acknowledge TOP is a workaround but I'd say the above does still justify it's existence for now.

Mailaender:
We already have https://github.com/OpenRA/OpenRA/wiki/Traits#transformoncapture for that.

I could think of an engineer getting into a vehicle and it transforms into a construction yard. Transforming is still a major part of this engine. It is used on every skirmish game start multiple times. It is also way easier to use and understand than loosely connected "upgrades".

GraionDilach:
`TransformOnCapture` does not returns the actor to the owner when all passengers exit. RA2/Generals garrisons/Ares's Squatter's Rights -  http://ares-developers.github.io/Ares-docs/new/buildings/urbancombattrenches.html#squatters-rights - come into my mind.

pchote:
A `InheritFirstPassengerOwner` flag on Cargo combined with allowing neutral actors to enter would be sufficient for that, and has been loosely planned for quite some time.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c77e365820c8408bf5874c37548ccb538e0457b5...c232ea66cfa78134a7380b2346c72e3db266dd40)

PR #10256, opened 2015-12-21
Spatially partition frozen actors

### LINKED ISSUE #9614: Use spatial partitioning to speed up frozen actors
One of our biggest perf hogs is FrozenActorLayer, which checks every frozen actor for visibility every tick.  We could save a lot of perf by tracking dirty flags for each bin in a spatial partition, and only ticking the frozen actors that are in dirty bins.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27333596-use-spatial-partitioning-to-speed-up-frozen-actors?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RoosterDragon:
Like other parts of the engine, this PR now tracks change in the shroud so it only had to update frozen actors on the map in areas where the shroud may have changed, rather than needing to update them all.

On the RA shellmap, this reduces CPU time spent in FrozenActorLayer.Tick from 2.32% to 1.51%. (For comparison - this is our most expensive trait to tick at the moment, the next worst is RenderSprites at 1.46%)

Closes #9614.

### DIFF
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -13,13 +13,17 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
 using OpenRA.Graphics;
+using OpenRA.Primitives;
 
 namespace OpenRA.Traits
 {
 	[Desc("Required for FrozenUnderFog to work. Attach this to the player actor.")]
-	public class FrozenActorLayerInfo : ITraitInfo
+	public class FrozenActorLayerInfo : Requires<ShroudInfo>, ITraitInfo
 	{
-		public object Create(ActorInitializer init) { return new FrozenActorLayer(init.Self); }
+		[Desc("Size of partition bins (cells)")]
+		public readonly int BinSize = 10;
+
+		public object Create(ActorInitializer init) { return new FrozenActorLayer(init.Self, this); }
 	}
 
 	public class FrozenActor
@@ -74,13 +78,11 @@ namespace OpenRA.Traits
 
 		public void Tick()
 		{
-			UpdateVisibility();
-
 			if (flashTicks > 0)
 				flashTicks--;
 		}
 
-		void UpdateVisibility()
+		public void UpdateVisibility()
 		{
 			var wasVisible = Visible;
 			Shrouded = true;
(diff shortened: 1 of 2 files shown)


OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -146,54 +148,123 @@ public class FrozenActorLayer : IRender, ITick, ISync
...
 
 		public void Add(FrozenActor fa)
 		{
-			frozen.Add(fa.ID, fa);
+			frozenActorsById.Add(fa.ID, fa);
 			world.ScreenMap.Add(owner, fa);
+			partitionedFrozenActorIds.Add(fa.ID, FootprintBounds(fa));
+		}
+
+		Rectangle FootprintBounds(FrozenActor fa)
```
RoosterDragon: @pchote: This appears to work in TS - but can you confirm if this is a valid way to get the bounds in isometric space?

pchote: Haven't tested it, but the underlying logic at least is sound.

Mailaender:
Can't spot anything wrong in-game. :white_check_mark:

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -146,54 +148,123 @@ public class FrozenActorLayer : IRender, ITick, ISync
...
+			partitionedFrozenActorIds = new SpatiallyPartitioned<uint>(
+				world.Map.MapSize.X, world.Map.MapSize.Y, binSize);
+			dirtyBins = new bool[world.Map.MapSize.X / binSize + 1, world.Map.MapSize.Y / binSize + 1];
+			self.Trait<Shroud>().CellsChanged += cells =>
+			{
+				foreach (var cell in cells)
+				{
+					var x = cell.U / binSize;
+					var y = cell.V / binSize;
+					dirtyBins[x, y] = true;
```
pchote: You can save a bit more perf (unnecessary bounds checking) by making this a 1d array indexed on `y * world.Map.MapSize.X + x`.

RoosterDragon: Added, but with the correct dimensions.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/61ee467173d19078eb221d21c9a15ba167f9ad89...1eee8270bd3d9989efa85c9d42883bddc056ad3d)

PR #10258, opened 2015-12-22
DeployToUpgrade: Deploy Animations processed before starting deploy

Iamgoofball:
When trying to implement the Allied G.I.'s Deploy function at this PR:
https://github.com/OpenRA/ra2/pull/3
I noticed that DeployToUpgrade wasn't wanting to render the DeployAnimation.

This fixes that issue, by causing the DeployAnimation to render before the rest of the effects of the Deploy/Undeploy.

The only other unit in the entire project to use DeployToUpgrade is the Dune 2000 Thumper unit, and they do not have a DeployAnimation, and are therefore unaffected.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
@@ -64,7 +64,7 @@ namespace OpenRA.Mods.Common.Traits
 		readonly UpgradeManager manager;
 		readonly bool checkTerrainType;
 		readonly bool canTurn;
-		readonly Lazy<ISpriteBody> body;
+		readonly Lazy<WithSpriteBody> body;
 
 		DeployState deployState;
 
@@ -75,7 +75,7 @@ namespace OpenRA.Mods.Common.Traits
 			manager = self.Trait<UpgradeManager>();
 			checkTerrainType = info.AllowedTerrainTypes.Count > 0;
 			canTurn = self.Info.HasTraitInfo<IFacingInfo>();
-			body = Exts.Lazy(self.TraitOrDefault<ISpriteBody>);
+			body = Exts.Lazy(self.TraitOrDefault<WithSpriteBody>);
 			if (init.Contains<DeployStateInit>())
 				deployState = init.Get<DeployStateInit, DeployState>();
 		}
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -194,17 +197,15 @@ void Deploy(bool init)
...
 			if (!init)
 				OnDeployStarted();
 
 			// If there is no animation to play just grant the upgrades that are used while deployed.
 			// Alternatively, play the deploy animation and then grant the upgrades.
-			if (string.IsNullOrEmpty(info.DeployAnimation) || body.Value == null)
-				OnDeployCompleted();
-			else
-				body.Value.PlayCustomAnimation(self, info.DeployAnimation, OnDeployCompleted);
+			OnDeployCompleted();
```
pchote: This makes OnDeployCommand to run immediately, before the animation has completed.  That's not what we want to happen.

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -217,16 +218,14 @@ void Undeploy(bool init)
...
 			if (!init)
 				OnUndeployStarted();
 
 			// If there is no animation to play just grant the upgrades that are used while undeployed.
 			// Alternatively, play the undeploy animation and then grant the upgrades.
-			if (string.IsNullOrEmpty(info.DeployAnimation) || body.Value == null)
-				OnUndeployCompleted();
-			else
-				body.Value.PlayCustomAnimationBackwards(self, info.DeployAnimation, OnUndeployCompleted);
+			OnUndeployCompleted();
```
pchote: Likewise here.

Iamgoofball:
As I stated in IRC, locking the effects to occur after the animation is what caused the animations to not play, and when I tried to do this without the broken animations it would cause crashes due to undeploying while undeployed somehow.

OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -194,17 +197,15 @@ void Deploy(bool init)
 
 			if (!string.IsNullOrEmpty(info.DeploySound))
 				Game.Sound.Play(info.DeploySound, self.CenterPosition);
-
+			if (!string.IsNullOrEmpty(info.DeployAnimation))
+				body.Value.PlayCustomAnimation(self, info.DeployAnimation);
```
penev92: This will break the Mobile Sensor Array/Tick Tank/the rest of the deploying TS vehicles.

Their `SpriteBody` is enabled by the upgrades granted below, because they are voxel actors while mobile, and sprites while deployed. That's why the upgrades have to come first.

penev92:
Left a comment on https://github.com/OpenRA/ra2/pull/3 explaining how to fix your animation issue with YAML changes only. :)

reaperrr:
> @OpenRA/engine-hackers I noticed that for some reason WithInfantryBody doesn't inherit from WithSpriteBody. Is it me, or is that an oversight on our part?

Not migrating it was intentional on my end due to all the infantry-specific complexity, but I guess ultimately we should at least try to migrate it as well.

penev92:
This has been resolved downstream.

PR #10260, opened 2015-12-22
Factor silos needed warning to a new trait.

ghost:
This was requested in #10075

### DIFF
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -382,6 +382,7 @@
     <Compile Include="Traits\Player\ProductionQueue.cs" />
     <Compile Include="Traits\Player\ProvidesPrerequisite.cs" />
     <Compile Include="Traits\Player\ProvidesTechPrerequisite.cs" />
+    <Compile Include="Traits\Player\ResourceStorageWarning.cs" />
     <Compile Include="Traits\Player\StrategicVictoryConditions.cs" />
     <Compile Include="Traits\Player\TechTree.cs" />
     <Compile Include="Traits\Power\AffectedByPowerOutage.cs" />
--- a/OpenRA.Game/Traits/Player/PlayerResources.cs
+++ b/OpenRA.Game/Traits/Player/PlayerResources.cs
@@ -17,7 +17,6 @@ namespace OpenRA.Traits
 	{
 		public readonly int[] SelectableCash = { 2500, 5000, 10000, 20000 };
 		public readonly int DefaultCash = 5000;
-		public readonly int AdviceInterval = 250;
 
 		public object Create(ActorInitializer init) { return new PlayerResources(init.Self, this); }
 	}
@@ -27,14 +26,12 @@ namespace OpenRA.Traits
 		const float DisplayCashFracPerFrame = .07f;
 		const int DisplayCashDeltaPerFrame = 37;
 		readonly Player owner;
-		int adviceInterval;
 
 		public PlayerResources(Actor self, PlayerResourcesInfo info)
 		{
 			owner = self.Owner;
 
 			Cash = self.World.LobbyInfo.GlobalSettings.StartingCash;
-			adviceInterval = info.AdviceInterval;
 		}
 
 		[Sync] public int Cash;
@@ -44,7 +41,6 @@ namespace OpenRA.Traits
 
 		public int DisplayCash;
 		public int DisplayResources;
-		public bool AlertSilo;
 
 		public int Earned;
 		public int Spent;
@@ -61,8 +57,6 @@ namespace OpenRA.Traits
 
 			if (Resources > ResourceCapacity)
 			{
-				nextSiloAdviceTime = 0;
-
 				Earned -= Resources - ResourceCapacity;
 				Resources = ResourceCapacity;
 			}
@@ -126,7 +120,6 @@ namespace OpenRA.Traits
 			return true;
 		}
 
-		int nextSiloAdviceTime = 0;
 		int nextCashTickTime = 0;
 
 		public void Tick(Actor self)
@@ -140,19 +133,6 @@ namespace OpenRA.Traits
 
 			if (Resources > ResourceCapacity)
 				Resources = ResourceCapacity;
-
-			if (--nextSiloAdviceTime <= 0)
-			{
-				if (Resources > 0.8 * ResourceCapacity)
-				{
-					Game.Sound.PlayNotification(self.World.Map.Rules, owner, "Speech", "SilosNeeded", owner.Faction.InternalName);
-					AlertSilo = true;
-				}
-				else
-					AlertSilo = false;
-
-				nextSiloAdviceTime = adviceInterval;
-			}
 
 			var diff = Math.Abs(Cash - DisplayCash);
 			var move = Math.Min(Math.Max((int)(diff * DisplayCashFracPerFrame), DisplayCashDeltaPerFrame), diff);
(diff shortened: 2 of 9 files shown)


OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
```diff
@@ -0,0 +1,60 @@
...
+		public readonly int AdviceInterval = 250;
+
+		[Desc("The threshold (as a ratio of resources to capacity) above which a warning is played.")]
+		public readonly float Threshold = 0.8f;
+
+		[Desc("The sound type of the warning.")]
+		public readonly string SoundType = "Speech";
+
+		[Desc("Sound to play for the warning.")]
+		public readonly string SoundName = "SilosNeeded";
```
pchote: Please make this `Notification` and then remove the `SoundType` field above.  See the comments in #9605 for an explanation as to why a toggle between speech and notifications is bad.

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Provides the player with an audible warning when their storage is nearing full.")]
+	public class StorageWarningInfo : ITraitInfo, Requires<PlayerResourcesInfo>
+	{
+		[Desc("Interval at which to check if more storage is needed.")]
+		public readonly int AdviceInterval = 250;
+
+		[Desc("The threshold (as a ratio of resources to capacity) above which a warning is played.")]
+		public readonly float Threshold = 0.8f;
```
pchote: Can you make this an integer percentage?

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
```diff
@@ -0,0 +1,60 @@
...
+		public readonly string SoundName = "SilosNeeded";
+
+		public object Create(ActorInitializer init) { return new StorageWarning(init.Self, this); }
+	}
+
+	public class StorageWarning : ITick
+	{
+		int nextSiloAdviceTime = 0;
+
+		public readonly StorageWarningInfo Info;
```
pchote: This isn't used anywhere else, so you can drop the `public` and rename it to `info`.

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
```diff
@@ -0,0 +1,60 @@
...
+		public StorageWarning(Actor self, StorageWarningInfo info)
+		{
+			Info = info;
+		}
+
+		public void Tick(Actor self)
+		{
+			if (--nextSiloAdviceTime <= 0)
+			{
+				var resources = self.Trait<PlayerResources>();
```
pchote: Please cache this in the constructor.

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
```diff
@@ -0,0 +1,60 @@
...
+				var owner = self.Owner;
```
pchote: This looks like an unnecessary variable – you only use it once.

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
```diff
@@ -0,0 +1,60 @@
...
+
+				if (resources.Resources > Info.Threshold * resources.ResourceCapacity)
+				{
```
pchote: OpenRA code style is to omit braces around single line blocks.

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Provides the player with an audible warning when their storage is nearing full.")]
+	public class StorageWarningInfo : ITraitInfo, Requires<PlayerResourcesInfo>
+	{
+		[Desc("Interval at which to check if more storage is needed.")]
+		public readonly int AdviceInterval = 250;
```
pchote: The naming and use of this variable implies that it should be specified in seconds instead of ticks.  In the code you can convert it to ticks by multiplying the value (in seconds) by 1000 and then dividing by the current game speed's Timestep.

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
scshunt: I didn't do that because it didn't look safe to do. What if this trait is constructed first?

GraionDilach: `INotifyCreated`?

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
scshunt: Hmm... my goal was to preserve the existing code. Could I just change the documentation to indicate that it's in ticks?

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
scshunt: Ah, thanks.

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
scshunt: It's used twice on the same line.

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
pchote: `INotifyCreated` is not necessary.  The `Requires<PlayerResourcesInfo>` guarantees that the trait will be available in the constructor.

scshunt:
Updated with new fixes.

OpenRA.Mods.Common/Traits/Player/StorageWarning.cs
```diff
@@ -0,0 +1,55 @@
...
+		public readonly string Notification = "SilosNeeded";
+
+		public object Create(ActorInitializer init) { return new StorageWarning(init.Self, this); }
+	}
+
+	public class StorageWarning : ITick
+	{
+		int nextSiloAdviceTime = 0;
+		readonly StorageWarningInfo info;
+		PlayerResources resources;
```
pchote: This can be made readonly.

OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
```diff
@@ -382,6 +382,7 @@
     <Compile Include="Traits\Player\ProductionQueue.cs" />
     <Compile Include="Traits\Player\ProvidesPrerequisite.cs" />
     <Compile Include="Traits\Player\ProvidesTechPrerequisite.cs" />
+    <Compile Include="Traits\Player\StorageWarning.cs" />
```
penev92: Can we slap a "Resource" at the beginning of that name?

scshunt:
Fixups done, including some messes of renamed files. Should be good to go now.

Mailaender:
Nice cleanup. :+1: Works aso promised. :white_check_mark:

obrakmann:
ok :+1: 

Needs a rebase (probably just a conflict in the upgrade rules) before merging. Please update the `engineVersion`, too, then.

scshunt:
Done.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/debd28325c8724f0ee944d551da0b6b85d2ff8b5)

PR #7411, opened 2015-01-31
Added restart in singleplayer for cnc, ra, d2k

Smith00101010:
We implemented restart buttons for the singleplayer of cnc, ra, d2k

Adresses #4874

Issue #4874, opened 2014-03-15
Implement a restart mission button


deniz1a:
This would be useful for single player but also for multi player game. It would restart the game with the same settings. In multiplayer, there could be a logic similar to the ready tick box in game lobby. If every player agrees, the game restarts.

Mailaender:
https://github.com/OpenRA/OpenRA/issues/10264 has been proposed to add it.

abcdefg30:
#10264 was merged.

PR #10264, opened 2015-12-23
Restart button for missions

whinis:
This Addresses #4874 and follows on Pull Request #7411
Adds a restart button to the abort menu and end game menu to allow restarting of the mission
Fixes bug in Pull Request #7411 that prevented new replay files from being created for restarted missions and crashes in the resulting single file.

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -177,8 +177,39 @@ namespace OpenRA
 
 		public static void RestartGame()
 		{
-			OrderManager.World.EndGame();
-			StartGame(OrderManager.World.Map.Uid, WorldType.Regular);
+			var replay = OrderManager.Connection as ReplayConnection;
+			var replayName = replay != null ? replay.Filename : null;
+			var uid = OrderManager.World.Map.Uid;
+			var globalSettings = OrderManager.LobbyInfo.GlobalSettings;
+
+			// Disconnect from the current game
+			Disconnect();
+			Ui.ResetAll();
+
+			// Restart the game with the same replay/mission
+			if (replay != null)
+				JoinReplay(replayName);
+			else
+				StartMission(uid, globalSettings.GameSpeedType, globalSettings.Difficulty);
+		}
+
+		public static void StartMission(string mapUID, string gameSpeed, string difficulty, Action onStart = null)
+		{
+			OrderManager om = null;
+
+			Action lobbyReady = null;
+			lobbyReady = () =>
+			{
+				LobbyInfoChanged -= lobbyReady;
+				om.IssueOrder(Order.Command("gamespeed {0}".F(gameSpeed)));
+				om.IssueOrder(Order.Command("difficulty {0}".F(difficulty)));
+				om.IssueOrder(Order.Command("state {0}".F(Session.ClientState.Ready)));
+				if (onStart != null)
+					onStart();
+			};
+			LobbyInfoChanged += lobbyReady;
+
+			om = JoinServer(IPAddress.Loopback.ToString(), CreateLocalServer(mapUID), "");
 		}
 
 		public static bool IsHost
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -175,6 +175,12 @@ namespace OpenRA
 			GC.Collect();
 		}
 
+		public static void RestartGame()
+		{
+			OrderManager.World.EndGame();
+			StartGame(OrderManager.World.Map.Uid, WorldType.Regular);
+		}
+
 		public static bool IsHost
 		{
 			get
--- a/OpenRA.Game/Network/ReplayRecorder.cs
+++ b/OpenRA.Game/Network/ReplayRecorder.cs
@@ -69,6 +69,9 @@ namespace OpenRA.Network
 
 		public void Receive(int clientID, byte[] data)
 		{
+			if (disposed) // TODO: This can be removed once NetworkConnection is fixed to dispose properly.
+				return;
+
 			if (preStartBuffer != null && IsGameStart(data))
 			{
 				writer.Flush();
--- a/OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
+++ b/OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
@@ -55,6 +55,39 @@ namespace OpenRA.Mods.Common.Widgets
 					onCancel();
 			};
 		}
+		/**
+		 * open confirmation dialog for mission / game restart
+		 */
+		public static void PromptAbortMission(World world, string title, string text, Action onAbort, Action onCancel = null, Action closeMenu = null)
+		{
+			var isMultiplayer = !world.LobbyInfo.IsSinglePlayer && !world.IsReplay;
+			var prompt = Ui.OpenWindow("ABORT_MISSION_PROMPT");
+			prompt.Get<LabelWidget>("PROMPT_TITLE").GetText = () => title;
+			prompt.Get<LabelWidget>("PROMPT_TEXT").GetText = () => text;
+			prompt.Get<ButtonWidget>("ABORT_BUTTON").OnClick = () =>
+			{
+				Ui.CloseWindow();
+				onAbort();
+			};
+
+			var restartButton = prompt.Get<ButtonWidget>("RESTART_BUTTON");
+			restartButton.IsVisible = () => !isMultiplayer;
+			restartButton.OnClick = () =>
+			{
+				if (closeMenu != null)
+					closeMenu();
+
+				Ui.CloseWindow();
+				Game.RestartGame();
+			};
+
+			prompt.Get<ButtonWidget>("CANCEL_BUTTON").OnClick = () =>
+			{
+				Ui.CloseWindow();
+				if (onCancel != null)
+					onCancel();
+			};
+		}
 
 		public static void TextInputPrompt(
 			string title, string prompt, string initialText,
--- a/OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
+++ b/OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
@@ -15,28 +15,60 @@ namespace OpenRA.Mods.Common.Widgets
 {
 	public static class ConfirmationDialogs
 	{
-		public static void PromptConfirmAction(string title, string text, Action onConfirm, Action onCancel = null, string confirmText = null, string cancelText = null)
+		public static void PromptConfirmAction(
+			string title,
+			string text,
+			Action onConfirm,
+			Action onCancel = null,
+			Action onOther = null,
+			string confirmText = null,
+			string cancelText = null,
+			string otherText = null)
 		{
 			var prompt = Ui.OpenWindow("CONFIRM_PROMPT");
+			var confirmButton = prompt.Get<ButtonWidget>("CONFIRM_BUTTON");
+			var cancelButton = prompt.GetOrNull<ButtonWidget>("CANCEL_BUTTON");
+			var otherButton = prompt.GetOrNull<ButtonWidget>("OTHER_BUTTON");
+
 			prompt.Get<LabelWidget>("PROMPT_TITLE").GetText = () => title;
 			prompt.Get<LabelWidget>("PROMPT_TEXT").GetText = () => text;
 			if (!string.IsNullOrEmpty(confirmText))
-				prompt.Get<ButtonWidget>("CONFIRM_BUTTON").GetText = () => confirmText;
-			if (!string.IsNullOrEmpty(cancelText))
-				prompt.Get<ButtonWidget>("CANCEL_BUTTON").GetText = () => cancelText;
+				confirmButton.GetText = () => confirmText;
+			if (!string.IsNullOrEmpty(otherText) && otherButton != null)
+				otherButton.GetText = () => otherText;
+			if (!string.IsNullOrEmpty(cancelText) && cancelButton != null)
+				cancelButton.GetText = () => cancelText;
 
-			prompt.Get<ButtonWidget>("CONFIRM_BUTTON").OnClick = () =>
+			confirmButton.OnClick = () =>
 			{
 				Ui.CloseWindow();
 				onConfirm();
 			};
 
-			prompt.Get<ButtonWidget>("CANCEL_BUTTON").OnClick = () =>
+			if (onCancel != null && cancelButton != null)
 			{
-				Ui.CloseWindow();
-				if (onCancel != null)
-					onCancel();
-			};
+				cancelButton.IsVisible = () => true;
+				cancelButton.OnClick = () =>
+				{
+					Ui.CloseWindow();
+					if (onCancel != null)
+						onCancel();
+				};
+			}
+			else if (cancelButton != null)
+				cancelButton.IsVisible = () => false;
+
+			if (onOther != null && otherButton != null)
+			{
+				otherButton.IsVisible = () => true;
+				otherButton.OnClick = () =>
+				{
+					if (onOther != null)
+						onOther();
+                };
+			}
+			else if (otherButton != null)
+				otherButton.IsVisible = () => false;
 		}
 
 		public static void CancelPrompt(string title, string text, Action onCancel = null, string cancelText = null)
@@ -47,39 +79,6 @@ namespace OpenRA.Mods.Common.Widgets
 
 			if (!string.IsNullOrEmpty(cancelText))
 				prompt.Get<ButtonWidget>("CANCEL_BUTTON").GetText = () => cancelText;
-
-			prompt.Get<ButtonWidget>("CANCEL_BUTTON").OnClick = () =>
-			{
-				Ui.CloseWindow();
-				if (onCancel != null)
-					onCancel();
-			};
-		}
-		/**
-		 * open confirmation dialog for mission / game restart
-		 */
-		public static void PromptAbortMission(World world, string title, string text, Action onAbort, Action onCancel = null, Action closeMenu = null)
-		{
-			var isMultiplayer = !world.LobbyInfo.IsSinglePlayer && !world.IsReplay;
-			var prompt = Ui.OpenWindow("ABORT_MISSION_PROMPT");
-			prompt.Get<LabelWidget>("PROMPT_TITLE").GetText = () => title;
-			prompt.Get<LabelWidget>("PROMPT_TEXT").GetText = () => text;
-			prompt.Get<ButtonWidget>("ABORT_BUTTON").OnClick = () =>
-			{
-				Ui.CloseWindow();
-				onAbort();
-			};
-
-			var restartButton = prompt.Get<ButtonWidget>("RESTART_BUTTON");
-			restartButton.IsVisible = () => !isMultiplayer;
-			restartButton.OnClick = () =>
-			{
-				if (closeMenu != null)
-					closeMenu();
-
-				Ui.CloseWindow();
-				Game.RestartGame();
-			};
 
 			prompt.Get<ButtonWidget>("CANCEL_BUTTON").OnClick = () =>
 			{
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
@@ -95,7 +95,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				}
 
 				hideMenu = true;
-				ConfirmationDialogs.PromptConfirmAction("Abort Mission", "Leave this game and return to the menu?", onQuit, showMenu);
+				ConfirmationDialogs.PromptAbortMission(world, "Abort Mission", "Leave this game and return to the menu?", onQuit, showMenu, closeMenu);
 			};
 
 			var exitEditorButton = menu.Get<ButtonWidget>("EXIT_EDITOR");
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
@@ -88,14 +88,42 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 			abortMissionButton.OnClick = () =>
 			{
-				if (world.IsGameOver)
-				{
-					onQuit();
-					return;
-				}
-
 				hideMenu = true;
-				ConfirmationDialogs.PromptAbortMission(world, "Abort Mission", "Leave this game and return to the menu?", onQuit, showMenu, closeMenu);
+
+				if (world.LocalPlayer == null || (world.LocalPlayer.WinState != WinState.Won &&
+					(!world.IsGameOver || world.Map.Visibility == MapVisibility.MissionSelector)))
+				{
+					Action restartAction = null;
+					if (world.IsReplay || world.Map.Visibility == MapVisibility.MissionSelector)
+					{
+						var iop = world.WorldActor.TraitsImplementing<IObjectivesPanel>().FirstOrDefault();
+						var exitDelay = iop != null ? iop.ExitDelay : 0;
+						restartAction = () =>
+						{
+							Ui.CloseWindow();
+							if (mpe != null)
+							{
+								if (Game.IsCurrentWorld(world))
+									mpe.Fade(MenuPaletteEffect.EffectType.Black);
+								exitDelay += 40 * mpe.Info.FadeLength;
+							}
+
+							Game.RunAfterDelay(exitDelay, Game.RestartGame);
+						};
+					}
+
+					ConfirmationDialogs.PromptConfirmAction(
+						title: "Leave Mission",
+						text: "Leave this game and return to the menu?",
+						onConfirm: onQuit,
+						onCancel: showMenu,
+						confirmText: "Leave",
+						cancelText: "Stay",
+						otherText: "Restart",
+						onOther: restartAction);
+				}
+				else
+					onQuit();
 			};
 
 			var exitEditorButton = menu.Get<ButtonWidget>("EXIT_EDITOR");
@@ -103,7 +131,11 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			exitEditorButton.OnClick = () =>
 			{
 				hideMenu = true;
-				ConfirmationDialogs.PromptConfirmAction("Exit Map Editor", "Exit and lose all unsaved changes?", onQuit, showMenu);
+				ConfirmationDialogs.PromptConfirmAction(
+					title: "Exit Map Editor",
+					text: "Exit and lose all unsaved changes?",
+					onConfirm: onQuit,
+					onCancel: showMenu);
 			};
 
 			Action onSurrender = () =>
@@ -117,7 +149,11 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			surrenderButton.OnClick = () =>
 			{
 				hideMenu = true;
-				ConfirmationDialogs.PromptConfirmAction("Surrender", "Are you sure you want to surrender?", onSurrender, showMenu);
+				ConfirmationDialogs.PromptConfirmAction(
+					title: "Surrender",
+					text: "Are you sure you want to surrender?",
+					onConfirm: onSurrender,
+					onCancel: showMenu);
 			};
 
 			var saveMapButton = menu.Get<ButtonWidget>("SAVE_MAP");
--- a/OpenRA.Mods.Common/Widgets/Logic/ReplayBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/ReplayBrowserLogic.cs
@@ -403,16 +403,15 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			Action<ReplayMetadata, Action> onDeleteReplay = (r, after) =>
 			{
 				ConfirmationDialogs.PromptConfirmAction(
-					"Delete selected replay?",
-					"Delete replay '{0}'?".F(Path.GetFileNameWithoutExtension(r.FilePath)),
-					() =>
+					title: "Delete selected replay?",
+					text: "Delete replay '{0}'?".F(Path.GetFileNameWithoutExtension(r.FilePath)),
+					onConfirm: () =>
 					{
 						DeleteReplay(r);
 						if (after != null)
 							after.Invoke();
 					},
-					null,
-					"Delete");
+					confirmText: "Delete");
 			};
 
 			{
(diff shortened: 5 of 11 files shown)


OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
```diff
@@ -56,6 +56,37 @@ public static void CancelPrompt(string title, string text, Action onCancel = nul
 			};
 		}
 
+		public static void PromptAbortMission(World world, string title, string text, Action onAbort, Action onCancel = null, Action closeMenu = null)
```
pchote: Do you know why the original PR introduced this duplicate method?  It would be better to generalize `PromptConfirmAction` to cover this use case instead of duplicating code and yaml.

OpenRA.Game/Network/ReplayRecorderConnection.cs
```diff
@@ -89,10 +89,13 @@ public void Receive(Action<int, byte[]> packetFn)
 						StartSavingReplay(preStartData);
 					}
 
-					writer.Write(client);
-					writer.Write(data.Length);
-					writer.Write(data);
-					packetFn(client, data);
+					if (!disposed)
```
pchote: I don't think its valid to do anything in here if its been disposed... how about adding an `if (disposed)\nreturn;` at the top instead?

OpenRA.Game/Game.cs
```diff
@@ -172,7 +172,32 @@ internal static void StartGame(string mapUID, WorldType type)
...
 		}
 
-		public static bool IsHost
+		public static void RestartGame()
+		{
+			string uid = OrderManager.World.Map.Uid;
+			bool record = false;
+			string difficulty = OrderManager.LobbyInfo.GlobalSettings.Difficulty;
+			string gameSpeed = OrderManager.LobbyInfo.GlobalSettings.GameSpeedType;
+			if (OrderManager.Connection is ReplayRecorderConnection)
```
pchote: you can simplify this down to `var record = OrderManager.Connection is ReplayRecorderConnection`.  Please also move it to just before the `JoinServer` call where it is used.

OpenRA.Game/Game.cs
```diff
@@ -172,7 +172,32 @@ internal static void StartGame(string mapUID, WorldType type)
 			GC.Collect();
 		}
 
-		public static bool IsHost
+		public static void RestartGame()
+		{
+			string uid = OrderManager.World.Map.Uid;
```
pchote: OpenRA code style is to use `var` instead of explicitly typing local variables.

OpenRA.Game/Game.cs
```diff
@@ -172,7 +172,32 @@ internal static void StartGame(string mapUID, WorldType type)
...
+			Disconnect();
+			OrderManager om = null;
+			Action lobbyReady = null;
+			lobbyReady = () =>
+			{
+				om.IssueOrder(Order.Command("gamespeed {0}".F(gameSpeed)));
+				om.IssueOrder(Order.Command("difficulty {0}".F(difficulty)));
+				LobbyInfoChanged -= lobbyReady;
+				om.IssueOrder(Order.Command("state {0}".F(Session.ClientState.Ready)));
+			};
```
pchote: Style nit: newline after this please.

pchote:
The general approach here looks good to me, although I haven't tested it yet.  A few code nits as noted.

OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
```diff
@@ -15,13 +15,25 @@ namespace OpenRA.Mods.Common.Widgets
...
-		public static void PromptConfirmAction(string title, string text, Action onConfirm, Action onCancel = null, string confirmText = null, string cancelText = null)
+		public static void PromptConfirmAction(
+			string title,
+			string text,
+			Action onConfirm,
+			Action onCancel = null,
+			string confirmText = null,
+			string cancelText = null,
+			World world = null,
+			Action onRestart = null,
```
pchote: Please put this with the other Actions.

OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
```diff
@@ -15,13 +15,25 @@ namespace OpenRA.Mods.Common.Widgets
...
+			string restartText = null)
```
pchote: Please put this with the other labels.

OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
```diff
@@ -37,6 +49,23 @@ public static void PromptConfirmAction(string title, string text, Action onConfi
 				if (onCancel != null)
 					onCancel();
 			};
+
+			var restartButton = prompt.Get<ButtonWidget>("RESTART_BUTTON");
+			if (world != null)
+			{
+				var isVisible = world.IsReplay || world.Map.Visibility == MapVisibility.MissionSelector; // only show for missions and replays
```
pchote: This check should be done by the code calling `PromptConfirmAction` instead of hardcoding it here.

pchote: You should then be able to remove the `world` argument.

OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
```diff
@@ -37,6 +49,23 @@ public static void PromptConfirmAction(string title, string text, Action onConfi
...
+				restartButton.IsVisible = () => isVisible;
+				restartButton.OnClick = () =>
+				{
+					if (onRestart != null)
+						onRestart();
+
+					Ui.CloseWindow();
+					Game.RunAfterTick(() => Game.RestartGame());
```
pchote: `Game.RunAfterTick(Game.RestartGame)`.

OpenRA.Game/Game.cs
```diff
@@ -172,6 +172,35 @@ internal static void StartGame(string mapUID, WorldType type)
 			GC.Collect();
 		}
 
+		public static void RestartGame()
+		{
+			if (OrderManager.Connection is ReplayConnection)
```
pchote: The disconnect special casing doesn't sit well with me - we don't do this in the normal abort case, so it shouldn't be needed here.  How about something like:

```csharp
var replay = OrderManager.Connection as ReplayConnection;
var replayName = replay != null ? replay.Filename : null;

// Disconnect from the current game
Disconnect();

// Restart the game with the same replay/mission
if (replay != null)
	JoinReplay(i.Filename);
else
{
	// the rest of the code
}
```

OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
whinis: If I change the action and strings around all current calls to PromptConfirmAction will require a null before the labels because they do not use restart.

whinis:
So I attempted to check if this code would work with skirmish missions but there is currently no easy way of syncing the settings or bots again. However as currently pushed it does work for replays (allowing you to start at the beginning of the replay) and for all missions tested so far in mod RA,CNC, and D2k

atlimit8:
Style fix commits should be squashed away.

whinis:
Im not really sure how to better squash as with talks to pchote in IRC he said I should split my commits due to changes to PromptConfirmAction. Also some of my work relies on previous work but after these rewrites im not sure how much.

OpenRA.Game/Game.cs
```diff
@@ -172,6 +172,41 @@ internal static void StartGame(string mapUID, WorldType type)
...
+			var uid = OrderManager.World.Map.Uid;
+			var globalSettings = OrderManager.LobbyInfo.GlobalSettings;
+			var record = OrderManager.Connection is ReplayRecorderConnection;
+
+			// Disconnect from the current game
+			Disconnect();
+
+			// Restart the game with the same replay/mission
+			if (replay != null)
+			{
```
Phrohdoh: Unnecessary brackets.

OpenRA.Game/Network/ReplayRecorderConnection.cs
```diff
@@ -79,21 +79,23 @@ void StartSavingReplay(byte[] initialContent)
 
 		public void Receive(Action<int, byte[]> packetFn)
 		{
+			if (disposed) // if has been disposed do no continue as writer nolonger exist
```
Phrohdoh: This comment will need to be fixed.
It should go above the `if` line and grammatically corrected, if we are going to keep it.

OpenRA.Game/Network/ReplayRecorderConnection.cs
```diff
@@ -79,21 +79,23 @@ void StartSavingReplay(byte[] initialContent)
...
+				return;
```
Phrohdoh: Please add a newline after this.

OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
```diff
@@ -37,6 +48,22 @@ public static void PromptConfirmAction(string title, string text, Action onConfi
 				if (onCancel != null)
 					onCancel();
 			};
+
+			var restartButton = prompt.Get<ButtonWidget>("RESTART_BUTTON");
```
Phrohdoh: Please use `GetOrDefault` followed by a null check.
Mods should not be required to add this button.

OpenRA.Mods.Common/Widgets/Logic/ReplayBrowserLogic.cs
```diff
@@ -444,16 +443,15 @@ void SetupManagement()
 					}
 
 					ConfirmationDialogs.PromptConfirmAction(
-						"Delete all selected replays?",
-						"Delete {0} replays?".F(list.Count),
-						() =>
+						title: "Delete all selected replays?",
+						text: "Delete {0} replays?".F(list.Count),
```
Phrohdoh: `Delete 1 replays?` will look a bit unpolished.

mods/cnc/chrome/dialogs.yaml
```diff
@@ -199,7 +206,7 @@ Container@CANCEL_PROMPT:
 			Width: 140
 			Height: 35
 			Text: Cancel
-
+
```
Phrohdoh: Spurious whitespace.

atlimit8:
Sorry for confusing you. I saw that the commit messages say "style fix". Generic "style fix" short log descriptions make it harder to review history, so I will mark them for requested changes to keep the short log understandable. Please replace the short log descriptions with clearer descriptions of what the commits do.

OpenRA.Game/Network/ReplayRecorderConnection.cs
abcdefg30: It should also start with a captial.

whinis:
atlimit8 What would be the best way to handle that? I am not the person who made that commit nor do I have contact with them.

atlimit8:
Do you know how to use `git rebase -i`?

atlimit8:
The commits can be rewritten with a rebase; no need to contact the original author(s).

whinis:
That might be better, I squashed the style changes one

whinis:
Any update ?

whinis:
Further Squashed

Mailaender:

OpenRA.Game/Game.cs
```diff
@@ -172,6 +172,40 @@ internal static void StartGame(string mapUID, WorldType type)
...
+			Ui.ResetAll();
+
+			// Restart the game with the same replay/mission
+			if (replay != null)
+				JoinReplay(replayName);
+			else
+			{
+				OrderManager om = null;
+				Action lobbyReady = null;
+				var difficulty = globalSettings.Difficulty;
```
pchote: Style nit: these can be moved inline with the `Order.Command` lines below

OpenRA.Game/Network/ReplayRecorderConnection.cs
```diff
@@ -79,21 +79,25 @@ void StartSavingReplay(byte[] initialContent)
 
 		public void Receive(Action<int, byte[]> packetFn)
 		{
+			// If has been disposed do not continue as writer no longer exist
```
pchote: Grammar nit: "no longer _exists_"

OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
```diff
@@ -15,13 +15,25 @@ namespace OpenRA.Mods.Common.Widgets
...
+			string text,
+			Action onConfirm,
+			Action onCancel = null,
+			Action onRestart = null,
+			string confirmText = null,
+			string cancelText = null,
+			string restartText = null)
 		{
 			var prompt = Ui.OpenWindow("CONFIRM_PROMPT");
+			var restartButton = prompt.GetOrNull<ButtonWidget>("RESTART_BUTTON");
```
pchote: Could you pull out variables for `confirmButton` and `cancelButton` while you're here too?  That saves two unnecessary widget lookups.

pchote: This is a general purpose third button, so I suggest that we name it `otherButton` and `"OTHER_BUTTON"`

OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
```diff
@@ -37,6 +49,21 @@ public static void PromptConfirmAction(string title, string text, Action onConfi
...
+
+			if (!string.IsNullOrEmpty(restartText) && restartButton != null)
+			{
+				restartButton.IsVisible = () => true;
+				restartButton.OnClick = () =>
+				{
+					if (onRestart != null)
+						onRestart();
+
+					Ui.CloseWindow();
```
pchote: These should be pulled outside this function so that this can be kept as a general purpose button.

OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
```diff
@@ -37,6 +49,21 @@ public static void PromptConfirmAction(string title, string text, Action onConfi
 				if (onCancel != null)
 					onCancel();
 			};
+
+			if (!string.IsNullOrEmpty(restartText) && restartButton != null)
+			{
+				restartButton.IsVisible = () => true;
```
pchote: It would make more sense to toggle the visibility (for all three buttons) based on whether the action is null or not.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -88,14 +88,20 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
...
-				{
-					onQuit();
-					return;
-				}
-
 				hideMenu = true;
-				ConfirmationDialogs.PromptConfirmAction("Abort Mission", "Leave this game and return to the menu?", onQuit, showMenu);
+				var restartText = "";
+				if (world.IsReplay || world.Map.Visibility == MapVisibility.MissionSelector)
+					restartText = "Restart";
```
pchote: Once you've extracted the restart-specific code out of the dialog code and toggle the button visibility based on the action, you can then set the restartAction to be non-null here (basically just assign the block that you pull out of the helper).

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -88,14 +88,20 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
...
+
+				ConfirmationDialogs.PromptConfirmAction(
+					title: "Abort Mission",
+					text: "Leave this game and return to the menu?",
+					onConfirm: onQuit,
+					onCancel: showMenu,
+					onRestart: closeMenu,
+					confirmText: "Abort",
+					cancelText: "Cancel",
+					restartText: restartText);
```
pchote: You can then define `restartText: "Restart"` here.

OpenRA.Game/Game.cs
```diff
@@ -172,6 +172,38 @@ internal static void StartGame(string mapUID, WorldType type)
...
+				JoinReplay(replayName);
+			else
+			{
+				OrderManager om = null;
+				Action lobbyReady = null;
+				lobbyReady = () =>
+				{
+					om.IssueOrder(Order.Command("gamespeed {0}".F(globalSettings.GameSpeedType)));
+					om.IssueOrder(Order.Command("difficulty {0}".F(globalSettings.Difficulty)));
+					LobbyInfoChanged -= lobbyReady;
```
pchote: Style nit: this should go after the IssueOrders, not in the middle(!).

OpenRA.Mods.Common/Widgets/ConfirmationDialogs.cs
```diff
@@ -15,28 +15,60 @@ namespace OpenRA.Mods.Common.Widgets
...
-			if (!string.IsNullOrEmpty(cancelText))
-				prompt.Get<ButtonWidget>("CANCEL_BUTTON").GetText = () => cancelText;
+				confirmButton.GetText = () => confirmText;
+			if (!string.IsNullOrEmpty(otherText) && otherButton != null)
+				otherButton.GetText = () => otherText;
+			if (!string.IsNullOrEmpty(cancelText) && cancelButton != null)
+				cancelButton.GetText = () => cancelText;
 
-			prompt.Get<ButtonWidget>("CONFIRM_BUTTON").OnClick = () =>
+			confirmButton.OnClick = () =>
```
pchote: You'll now need null checks before using these variables (since `GetOrNull` can return null – `Get` would immediately throw an exception).

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -88,14 +88,25 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
...
-				{
-					onQuit();
-					return;
-				}
-
 				hideMenu = true;
-				ConfirmationDialogs.PromptConfirmAction("Abort Mission", "Leave this game and return to the menu?", onQuit, showMenu);
+				Action restartAction = null;
+				if (world.IsReplay || world.Map.Visibility == MapVisibility.MissionSelector)
+					restartAction = () =>
```
pchote: You've got a hanging block here, so please add braces around it.

whinis:
Found a crash actually while testing, if a mission implements https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Scripting/Global/UserInterfaceGlobal.cs then it will crash whenever you attempt to restart the replay. However restarting the actual missions itself is fine.

OpenRA.Game/Game.cs
```diff
@@ -172,6 +172,38 @@ internal static void StartGame(string mapUID, WorldType type)
 			GC.Collect();
 		}
 
+		public static void RestartGame()
```
ScottNZ: I think it's weird to have this Game.RestartGame() in the same file as Game.StartGame() which works completely differently to it, and StartMission() (the real corresponding method to RestartGame()) way out in MissionBrowserLogic.

I'd like to see RestartGame() and StartMission() moved together if possible.

whinis: The plan is for RestartGame to possibly do more in the future but right now it restarts missions and replays. As for moving them together, I'm not exactly sure how I would do that.

ScottNZ: Maybe move https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Widgets/Logic/MissionBrowserLogic.cs#L311-L326 into this file, add a map uid parameter to it, rename it to something more appropriate, and call it from MissionBrowserLogic.

whinis: Actually let me change that and push it

whinis:
Problem with GetOrNull has been fixed, a future PR will make the PromptConfirmAction function more generic to allow one message with 1-3 buttons rather than 2 functions.

whinis:
Previously mentions crash no longer happens after #10304 was merged

mods/cnc/chrome/dialogs.yaml
```diff
@@ -156,19 +156,26 @@ Container@CONFIRM_PROMPT:
...
 			Width: 140
 			Height: 35
-			Text: Cancel
-		Button@CONFIRM_BUTTON:
-			Key: return
-			X: 230
+			Text: Abort
+		Button@OTHER_BUTTON:
+			Key: r
+			X: 180
```
pchote: The UI standard for TD is to group related buttons with 10px clear space between them, with 'forward' moving actions to the right, and 'back' moving actions to the left.  Please move this 30px to the right to fit in with that.

whinis:
Is this what you meant pchote? it looks a little odd now

pchote:
Yes, that is what I mean.  Dialogs now look fine to me.

One (hopefully) last issue: pressing restart should do the "battle control terminated" notification and fade out instead of immediately jumping to the load screen.

pchote:
It would also be good if you could rebase this so that we don't need to clear and re-fetch dependencies when testing.

whinis:
Alright, I rebased for sake of testing and made it so the mission will fade before restarting. I noticed an odd bug while doing this in that if missiles are fired while its fading their trails stay on top of the fade.

OpenRA.Game/Network/ReplayRecorder.cs
```diff
@@ -69,6 +69,9 @@ void StartSavingReplay(byte[] initialContent)
 
 		public void Receive(int clientID, byte[] data)
 		{
+			if (disposed)
+				return;
```
RoosterDragon: Is this needed to prevent a crash? I feel we should fix the caller to not call methods on a disposed object if they're not going to work any more, rather than hiding the problem here.

whinis: So this is to prevent a crash but there isn't a way to check for disposed currently (other than making a public variable we check). Currently this is the only disposed object that crashes and that's because of the writer (once disposed accessing it causes an exception) and my understanding is that currently no other disposed object is checked because of this.

RoosterDragon: What I'm getting at is: we should fix it so nobody calls `Receive` after we've already disposed `ReplayRecorder`. We *should* crash if `ReplayRecorder` is getting new stuff after being disposed, because that likely indicates a bug.

whinis: As we talked about in IRC, I wouldn't be comfortable removing this piece of code until an upstream fix is implemented because otherwise it will crash every restart. Which would mean a broken PR

RoosterDragon: Sorry for not getting to that yet - that's pretty lame of me.

Could you add this comment:  `TODO: This can be removed once NetworkConnection is fixed to dispose properly.`

That'll remind me to get to it eventually, and we can take your PR in the meantime.

whinis:
Todo Added and fade code finally implemented. I forgot to push it it seems

whinis:
Rebased to be safe

obrakmann:
Do you want to update the author on the last commit before merging?

whinis:
Will do, should be home soon

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -88,14 +88,38 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
 
 			abortMissionButton.OnClick = () =>
 			{
-				if (world.IsGameOver)
```
obrakmann: I'd prefer it if we could keep this check (in spirit). It would probably need to be a bit smarter and only show when the player has not won, ie. something like
```csharp
if (world.LocalPlayer != null && world.LocalPlayer.WinState != WinState.Won)
   // do the prompt thing
else
{
    onQuit();
    return;
}
```

If there has already been discussion about this, I apologize, I missed it then.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -88,14 +88,38 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
...
 				}
 
-				hideMenu = true;
-				ConfirmationDialogs.PromptConfirmAction("Abort Mission", "Leave this game and return to the menu?", onQuit, showMenu);
+				ConfirmationDialogs.PromptConfirmAction(
+					title: "Abort Mission",
+					text: "Leave this game and return to the menu?",
+					onConfirm: onQuit,
+					onCancel: showMenu,
+					confirmText: "Abort",
```
obrakmann: Otherwise, if you *don't* add back the check mentioned above, I'd like you to change all instances of "Abort" to "Leave". Otherwise you complete a mission and the dialog will ask you if you want to abort, which would be strange.

obrakmann:
This looks good otherwise, well done. I'll take a final look after you've done the changes.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
Phrohdoh: I agree about the button text changes, "cancel" doesn't tell you what exactly you're canceling.
Please read http://ux.stackexchange.com/a/9960 if you have the time.

whinis: I was just copying over what was mostly there but I do see the value in showing a restart button even if you won the game. What if you wanted to restart the mission to complete a secondary objective you missed? So what would you rather happen since it appears the old behavior was after you win you have the option of staying "cancel" or leaving "abort".

whinis:
obrakmann Is that what you wanted?

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -88,14 +88,48 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
...
+					};
 				}
 
-				hideMenu = true;
-				ConfirmationDialogs.PromptConfirmAction("Abort Mission", "Leave this game and return to the menu?", onQuit, showMenu);
+				var confirmText = "Abort";
+				var cancelText = "Cancel";
+				var otherText = "Restart";
+
+				if (world.LocalPlayer == null || world.LocalPlayer.WinState == WinState.Won)
```
Phrohdoh: What if you have lost?

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -88,14 +88,48 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
...
+					var iop = world.WorldActor.TraitsImplementing<IObjectivesPanel>().FirstOrDefault();
+					var exitDelay = iop != null ? iop.ExitDelay : 0;
+					restartAction = () =>
+					{
+						Ui.CloseWindow();
+						if (mpe != null)
+						{
+							if (Game.IsCurrentWorld(world))
+								mpe.Fade(MenuPaletteEffect.EffectType.Black);
+							exitDelay += 40 * mpe.Info.FadeLength;
```
Phrohdoh: If this `40` is ticks it should probably use the current gamespeed.

whinis: Im using the same code as https://github.com/whinis/OpenRA/blob/bleed/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs#L50, I'm not sure if its suppose to be 40 ticks but does produce a fading effect that pchote requested.

Phrohdoh: It is definitely ticks and should use gamespeed, but that can be left for another PR because your reasoning for not changing it is valid.

[Reference](https://github.com/OpenRA/OpenRA/blob/da4f04acfb2983038d07baa52d1cc42fcfafcd31/OpenRA.Mods.Common/Traits/PaletteEffects/MenuPaletteEffect.cs#L21-L22).

whinis:
That should be a little more clear then, im not an expert on that area but tried to make it a little more context sensitive.

obrakmann:
> obrakmann Is that what you wanted?

Not quite. I'd like `abortMissionButton.OnClick` to look like this:

```csharp
			abortMissionButton.OnClick = () =>
			{
				hideMenu = true;
				Action restartAction = null;
				if (world.IsReplay || world.Map.Visibility == MapVisibility.MissionSelector)
				{
					var iop = world.WorldActor.TraitsImplementing<IObjectivesPanel>().FirstOrDefault();
					var exitDelay = iop != null ? iop.ExitDelay : 0;
					restartAction = () =>
					{
						Ui.CloseWindow();
						if (mpe != null)
						{
							if (Game.IsCurrentWorld(world))
								mpe.Fade(MenuPaletteEffect.EffectType.Black);
							exitDelay += 40 * mpe.Info.FadeLength;
						}

						Game.RunAfterDelay(exitDelay, () =>
						{
							Game.RestartGame();
						});
					};
				}

				if (world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Won &&
					(!world.IsGameOver || world.Map.Visibility == MapVisibility.MissionSelector))
				{
					ConfirmationDialogs.PromptConfirmAction(
						title: "Leave Mission",
						text: "Leave this game and return to the menu?",
						onConfirm: onQuit,
						onCancel: showMenu,
						confirmText: "Leave",
						cancelText: "Stay",
						otherText: "Restart",
						onOther: restartAction);
				}
				else
					onQuit();
			};
```

That will result in the following behaviour:

* Any game, in progress => Show prompt
* Skirmish game, when game is over => Don't show prompt
* Mission, when game is over
    * if the player won => Don't show prompt
    * if the player lost => Show prompt

pchote:
Not that `Game.RunAfterDelay(exitDelay, Game.RestartGame)` should work in the above.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -88,14 +88,42 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
 
 			abortMissionButton.OnClick = () =>
 			{
-				if (world.IsGameOver)
+				hideMenu = true;
+
+				if ((world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Won) &&
+								   (!world.IsGameOver || world.Map.Visibility == MapVisibility.MissionSelector))
```
obrakmann: `if (world.LocalPlayer == null || (world.LocalPlayer.WinState != WinState.Won && (!world.IsGameOver || world.Map.Visibility == MapVisibility.MissionSelector)))`

Also a few tabs too much on the second line, it seems.

obrakmann:
Thanks, looks good now. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/35181146a279dc1ee15c2f4ebcf5f7b9a67439b8)

PR #10266, opened 2015-12-23
Paradrop Production

Nelax:
I've been working on this for a little bit. It's a Production trait in the vein of ProductionAirdrop that paradrops the unit being produced. I think it still needs audio but seems to work well otherwise. I'll do some more independent testing but I figured I'd open up a pull request anyhow and get some input.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Parachute.cs
+++ b/OpenRA.Mods.Common/Activities/Parachute.cs
@@ -20,15 +20,17 @@ namespace OpenRA.Mods.Common.Activities
 		readonly IPositionable pos;
 		readonly ParachutableInfo para;
 		readonly WVec fallVector;
+		readonly Actor ignore;
 
 		WPos dropPosition;
 		WPos currentPosition;
 		bool triggered = false;
 
-		public Parachute(Actor self, WPos dropPosition)
+		public Parachute(Actor self, WPos dropPosition, Actor ignoreActor = null)
 		{
 			um = self.TraitOrDefault<UpgradeManager>();
 			pos = self.TraitOrDefault<IPositionable>();
+			ignore = ignoreActor;
 
 			// Parachutable trait is a prerequisite for running this activity
 			para = self.Info.TraitInfo<ParachutableInfo>();
@@ -60,7 +62,7 @@ namespace OpenRA.Mods.Common.Activities
 					um.RevokeUpgrade(self, u, this);
 
 			foreach (var npl in self.TraitsImplementing<INotifyParachuteLanded>())
-				npl.OnLanded();
+				npl.OnLanded(ignore);
 
 			return NextActivity;
 		}
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -396,6 +396,7 @@
     <Compile Include="Traits\Power\RequiresPower.cs" />
     <Compile Include="Traits\Power\ScalePowerWithHealth.cs" />
     <Compile Include="Traits\Production.cs" />
+    <Compile Include="Traits\ProductionParadrop.cs" />
     <Compile Include="Traits\ProductionQueueFromSelection.cs" />
     <Compile Include="Traits\ProvidesRadar.cs" />
     <Compile Include="Traits\ProximityCaptor.cs" />
--- a/OpenRA.Mods.Common/Traits/Parachutable.cs
+++ b/OpenRA.Mods.Common/Traits/Parachutable.cs
@@ -8,6 +8,7 @@
  */
 #endregion
 
+using System.Linq;
 using OpenRA.Mods.Common.Effects;
 using OpenRA.Traits;
 
@@ -53,12 +54,15 @@ namespace OpenRA.Mods.Common.Traits
 			positionable = self.TraitOrDefault<IPositionable>();
 		}
 
-		public void OnLanded()
+		public void OnLanded(Actor ignore)
 		{
 			if (!info.KilledOnImpassableTerrain)
 				return;
 
 			if (positionable.CanEnterCell(self.Location, self))
+				return;
+
+			if (ignore != null && self.World.ActorMap.GetActorsAt(self.Location).Any(a => a != ignore))
 				return;
 
 			var terrain = self.World.Map.GetTerrainInfo(self.Location);
--- a/OpenRA.Mods.Common/Traits/ProductionParadrop.cs
+++ b/OpenRA.Mods.Common/Traits/ProductionParadrop.cs
@@ -1,0 +1,172 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Drawing;
+using System.Linq;
+using OpenRA.Activities;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Primitives;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Deliver the unit in production via paradrop.")]
+	public class ProductionParadropInfo : ProductionInfo, Requires<ExitInfo>
+	{
+		[Desc("Cargo aircraft used. Must have Aircraft trait.")]
+		[ActorReference(typeof(AircraftInfo))] public readonly string ActorType = "badr";
+
+		[Desc("Sound to play when dropping the unit.")]
+		public readonly string ChuteSound = "chute1.aud";
+
+		[Desc("Notification to play when dropping the unit.")]
+		public readonly string ReadyAudio = null;
+
+		public override object Create(ActorInitializer init) { return new ProductionParadrop(init, this); }
+	}
+
+	class ProductionParadrop : Production
+	{
+		readonly Lazy<RallyPoint> rp;
+
+		public ProductionParadrop(ActorInitializer init, ProductionParadropInfo info)
+			: base(init, info)
+		{
+			rp = Exts.Lazy(() => init.Self.IsDead ? null : init.Self.TraitOrDefault<RallyPoint>());
+		}
+
+		public override bool Produce(Actor self, ActorInfo producee, string factionVariant)
+		{
+			var owner = self.Owner;
+
+			// Start a fixed distance away: the width of the map.
+			// This makes the production timing independent of spawnpoint
+			var startPos = self.Location + new CVec(owner.World.Map.Bounds.Width, 0);
+			var endPos = new CPos(owner.World.Map.Bounds.Left - 5, self.Location.Y);
+
+			// Assume a single exit point for simplicity
+			var exit = self.Info.TraitInfos<ExitInfo>().First();
+
+			foreach (var notify in self.TraitsImplementing<INotifyDelivery>())
+				notify.IncomingDelivery(self);
+
+			var info = (ProductionParadropInfo)Info;
+			var actorType = info.ActorType;
+
+			owner.World.AddFrameEndTask(w =>
+			{
+				if (!self.IsInWorld || self.IsDead)
+					return;
+
+				var altitude = self.World.Map.Rules.Actors[actorType].TraitInfo<AircraftInfo>().CruiseAltitude;
+				var actor = w.CreateActor(actorType, new TypeDictionary
+				{
+					new CenterPositionInit(w.Map.CenterOfCell(startPos) + new WVec(WDist.Zero, WDist.Zero, altitude)),
+					new OwnerInit(owner),
+					new FacingInit(64)
+				});
+
+				actor.QueueActivity(new Fly(actor, Target.FromCell(w, self.Location)));
+				actor.QueueActivity(new CallFunc(() =>
+				{
+					if (!self.IsInWorld || self.IsDead)
+						return;
+
+					foreach (var cargo in self.TraitsImplementing<INotifyDelivery>())
+						cargo.Delivered(self);
+
+					self.World.AddFrameEndTask(ww => DoProduction(self, producee, exit, factionVariant));
+					Game.Sound.Play(info.ChuteSound, self.CenterPosition);
+					Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Speech", info.ReadyAudio, self.Owner.Faction.InternalName);
+				}));
+
+				actor.QueueActivity(new Fly(actor, Target.FromCell(w, endPos)));
+				actor.QueueActivity(new RemoveSelf());
+			});
+
+			return true;
+		}
+
+		public override void DoProduction(Actor self, ActorInfo producee, ExitInfo exitinfo, string factionVariant)
+		{
+			var exit = CPos.Zero;
+			var exitLocation = CPos.Zero;
+			var target = Target.Invalid;
+
+			var info = (ProductionParadropInfo)Info;
+			var actorType = info.ActorType;
+
+			var bi = producee.TraitInfoOrDefault<BuildableInfo>();
+			if (bi != null && bi.ForceFaction != null)
+				factionVariant = bi.ForceFaction;
+
+			var altitude = self.World.Map.Rules.Actors[actorType].TraitInfo<AircraftInfo>().CruiseAltitude;
+			var td = new TypeDictionary
+			{
+				new OwnerInit(self.Owner),
+			};
+
+			if (self.OccupiesSpace != null)
+			{
+				exit = self.Location + exitinfo.ExitCell;
+				var spawn = self.CenterPosition + exitinfo.SpawnOffset + new WVec(WDist.Zero, WDist.Zero, altitude);
+				var to = self.World.Map.CenterOfCell(exit);
+
+				var fi = producee.TraitInfoOrDefault<IFacingInfo>();
+				var initialFacing = exitinfo.Facing < 0 ? Util.GetFacing(to - spawn, fi == null ? 0 : fi.GetInitialFacing()) : exitinfo.Facing;
+
+				exitLocation = rp.Value != null ? rp.Value.Location : exit;
+				target = Target.FromCell(self.World, exitLocation);
+
+				td.Add(new LocationInit(exit));
+				td.Add(new CenterPositionInit(spawn));
+				td.Add(new FacingInit(initialFacing));
+			}
+
+			self.World.AddFrameEndTask(w =>
+			{
+				if (factionVariant != null)
+					td.Add(new FactionInit(factionVariant));
+
+				var newUnit = self.World.CreateActor(producee.Name, td);
+
+				newUnit.QueueActivity(new Parachute(newUnit, newUnit.CenterPosition, self));
+
+				var move = newUnit.TraitOrDefault<IMove>();
+				if (move != null)
+				{
+					if (exitinfo.MoveIntoWorld)
+					{
+						if (exitinfo.ExitDelay > 0)
+							newUnit.QueueActivity(new Wait(exitinfo.ExitDelay, false));
+
+							newUnit.QueueActivity(move.MoveIntoWorld(newUnit, exit));
+							newUnit.QueueActivity(new AttackMoveActivity(
+								newUnit, move.MoveTo(exitLocation, 1)));
+					}
+				}
+
+				newUnit.SetTargetLine(target, rp.Value != null ? Color.Red : Color.Green, false);
+
+				if (!self.IsDead)
+					foreach (var t in self.TraitsImplementing<INotifyProduction>())
+						t.UnitProduced(self, newUnit, exit);
+
+				var notifyOthers = self.World.ActorsWithTrait<INotifyOtherProduction>();
+				foreach (var notify in notifyOthers)
+					notify.Trait.UnitProducedByOther(notify.Actor, self, newUnit);
+
+				foreach (var t in newUnit.TraitsImplementing<INotifyBuildComplete>())
+					t.BuildingComplete(newUnit);
+			});
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Traits/Crates/Crate.cs
+++ b/OpenRA.Mods.Common/Traits/Crates/Crate.cs
@@ -88,7 +88,7 @@ namespace OpenRA.Mods.Common.Traits
 			}
 		}
 
-		public void OnLanded()
+		public void OnLanded(Actor ignore)
 		{
 			// Check whether the crate landed on anything
 			var landedOn = self.World.ActorMap.GetActorsAt(self.Location)
--- a/OpenRA.Mods.Common/Traits/Production.cs
+++ b/OpenRA.Mods.Common/Traits/Production.cs
@@ -42,7 +42,7 @@ namespace OpenRA.Mods.Common.Traits
 			Faction = init.Contains<FactionInit>() ? init.Get<FactionInit, string>() : init.Self.Owner.Faction.InternalName;
 		}
 
-		public void DoProduction(Actor self, ActorInfo producee, ExitInfo exitinfo, string factionVariant)
+		public virtual void DoProduction(Actor self, ActorInfo producee, ExitInfo exitinfo, string factionVariant)
 		{
 			var exit = CPos.Zero;
 			var exitLocation = CPos.Zero;
--- a/OpenRA.Mods.Common/Traits/Render/WithCrateBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithCrateBody.cs
@@ -57,7 +57,7 @@ namespace OpenRA.Mods.Common.Traits
 			rs.Add(anim);
 		}
 
-		public void OnLanded()
+		public void OnLanded(Actor ignore)
 		{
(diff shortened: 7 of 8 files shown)


OpenRA.Mods.Common/Activities/Parachute.cs
```diff
@@ -20,15 +20,17 @@ public class Parachute : Activity
...
 		readonly ParachutableInfo para;
 		readonly WVec fallVector;
+		readonly Actor ignored;
 
 		WPos dropPosition;
 		WPos currentPosition;
 		bool triggered = false;
 
-		public Parachute(Actor self, WPos dropPosition)
+		public Parachute(Actor self, WPos dropPosition, Actor ignore)
```
pchote: I suggest renaming this to `ignoreActor` for consistency with the other traits, and you can default it to null so that you don't need to modfy EjectOnDeath or Paradrop.

OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
```diff
@@ -391,6 +391,7 @@
     <Compile Include="Traits\Power\RequiresPower.cs" />
     <Compile Include="Traits\Power\ScalePowerWithHealth.cs" />
     <Compile Include="Traits\Production.cs" />
+	<Compile Include="Traits\ProductionParadrop.cs" />
```
pchote: tabs -> spaces here.

OpenRA.Mods.Common/Traits/Parachutable.cs
```diff
@@ -71,7 +71,8 @@ public void OnLanded()
 			if (sequence != null && palette != null)
 				self.World.AddFrameEndTask(w => w.Add(new Explosion(w, self.OccupiesSpace.CenterPosition, sequence, palette)));
 
-			self.Kill(self);
+			if(self.World.ActorMap.GetActorsAt(self.Location).Equals(ignore))
```
pchote: `GetActorsAt` returns a collection, so comparing it with `ignore` (an actor) will always be false.  You want something like `self.World.ActorMap.GetActorsAt(self.Location).Any(a => a != ignore)`.  Also space between `if (` please.

mods/ra/rules/aircraft.yaml
```diff
@@ -355,4 +355,3 @@ U2:
 		Offset: -1c43,0,0
 		Interval: 2
 	RejectsOrders:
-
```
pchote: Please drop these newline removals from the pr.

OpenRA.Mods.Common/Traits/Parachutable.cs
```diff
@@ -71,7 +72,8 @@ public void OnLanded()
 			if (sequence != null && palette != null)
 				self.World.AddFrameEndTask(w => w.Add(new Explosion(w, self.OccupiesSpace.CenterPosition, sequence, palette)));
 
-			self.Kill(self);
+			if (self.World.ActorMap.GetActorsAt(self.Location).Any(a => a != ignore))
```
GraionDilach: What's the point of this?

You have blown the parachuted actor up by this point - even playing it's explosion.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+
+					newUnit.SetTargetLine(target, rp.Value != null ? Color.Red : Color.Green, false);
+
+					if (!self.IsDead)
+						foreach (var t in self.TraitsImplementing<INotifyProduction>())
+							t.UnitProduced(self, newUnit, exit);
+
+					var notifyOthers = self.World.ActorsWithTrait<INotifyOtherProduction>();
+					foreach (var notify in notifyOthers)
+						notify.Trait.UnitProducedByOther(notify.Actor, self, newUnit);
```
GraionDilach: Is this needed here? I'm not certain about what other traits use this interface but `UpgradeActorsNear` expects the unit being knocked out by then and not just flying in to the map.

GraionDilach:
There are a bunch of formatting errors as well. See `make check` and the Travis log.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
Phrohdoh: It isn't needed, and I don't think we have a use for it but that doesn't mean there can't be an Info flag regarding this and only notifying if that flag is true.

But if you do keep this code in any form you need to be sure you don't notify this production structure.
Look at `ClonesProducedUnits` for more information.

pchote: @Phrohdoh: This trait doesn't implement `INotifyOtherProduction` so there is no risk of it notifying itself. The code here is the standard pattern for production structures.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+				exitLocation = rp.Value != null ? rp.Value.Location : exit;
+				target = Target.FromCell(self.World, exitLocation);
+
+				td.Add(new LocationInit(exit));
+				td.Add(new CenterPositionInit(spawn));
+				td.Add(new FacingInit(initialFacing));
+			}
+
+			self.World.AddFrameEndTask(w =>
+				{
```
pchote: This block should be unindented by one tab to line up with the rest of the outer block.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+			var exit = self.Info.TraitInfos<ExitInfo>().First();
+
+			foreach (var tower in self.TraitsImplementing<INotifyDelivery>())
+				tower.IncomingDelivery(self);
+
+			var info = (ProductionParadropInfo)Info;
+			var actorType = info.ActorType;
+
+			owner.World.AddFrameEndTask(w =>
+				{
```
pchote: This block should be unindented by one tab to line up with the rest of the outer block.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
Phrohdoh: This trait doesn't but other traits given to an actor using this trait might.

pchote: The other production traits do not do anything special with regards to `INotifyOtherProduction`.  What makes this one different?

Phrohdoh: You're right, as usual.
The check I was talking about should be done on the receiving end.

GraionDilach: My concerns are coming from the case that this trait spawns the actor in a completely different place compared to the other production tabs and that might make a difference for example in the case of UAN.

OpenRA.Mods.Common/Traits/Parachutable.cs
Bynnar18: You're right, I put this in rather quickly and somehow it ended up working before but I just retested it. I'll need to fix it.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+
+namespace OpenRA.Mods.Common
+{
+
+	[Desc("Deliver the unit in production via paradrop.")]
+	public class ProductionParadropInfo : ProductionInfo
+	{
+		public readonly string ReadyAudio = "ReinforcementsArrived";
+		[Desc("Cargo aircraft used.")]
+		[ActorReference] public readonly string ActorType = "badr";
```
atlimit8: `Transport` or `TransportType`
`ActorType` is an ambiguous name.

Bynnar18: ActorType is what ProductionAirdrop calls the transport. It really doesn't matter since it doesn't need to be a transport technically.

atlimit8: Then that probably should also be changed. Even if it is not `TransportType`, `ActorType` seems too generic.
 Both could be done in another PR when something better comes to someone's mind.

atlimit8:
Fix-ups for indentation as mentioned above.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
Phrohdoh: `TransportActorType` or `TransportActor` are perfectly valid.

GraionDilach:
Changes requested also fits here for the reversion of the change in `Parachutable`.

OpenRA.Mods.Common/Traits/Parachutable.cs
GraionDilach: What are you trying to achieve with this? That would probably help. Right now I would say the interface change isn't even needed.

Bynnar18:
Ok, it still isn't completely functional but I did some fixup. I need to figure out why the unit just stays on the landing pad without moving though and why the unit is being dropped off the default location. The ladder probably just needs changes to the code in ProductionParadrop, the former is a little more difficult.

OpenRA.Mods.Common/Traits/Parachutable.cs
```diff
@@ -61,17 +62,20 @@ public void OnLanded()
...
 
 			var sound = terrain.IsWater ? info.WaterImpactSound : info.GroundImpactSound;
-			Game.Sound.Play(sound, self.CenterPosition);
+			Game.Sound.Play (sound, self.CenterPosition);
 
 			var sequence = terrain.IsWater ? info.WaterCorpseSequence : info.GroundCorpseSequence;
 			var palette = terrain.IsWater ? info.WaterCorpsePalette : info.GroundCorpsePalette;
 			if (sequence != null && palette != null)
-				self.World.AddFrameEndTask(w => w.Add(new Explosion(w, self.OccupiesSpace.CenterPosition, sequence, palette)));
+				self.World.AddFrameEndTask (w => w.Add (new Explosion (w, self.OccupiesSpace.CenterPosition, sequence, palette)));
```
Phrohdoh: These spaces are not desired.

OpenRA.Mods.Common/Traits/Parachutable.cs
```diff
@@ -61,17 +62,20 @@ public void OnLanded()
...
 
-			self.Kill(self);
+			self.Kill (self);
```
Phrohdoh: Nor is this one.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common
+{
+
+	[Desc("Deliver the unit in production via paradrop.")]
+	public class ProductionParadropInfo : ProductionInfo
+	{
+		public readonly string ReadyAudio = "ReinforcementsArrived";
+		[Desc("Cargo aircraft used.")]
```
Phrohdoh: Please add a blank line between the desc and the previous line.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+
+			// Start a fixed distance away: the width of the map.
+			// This makes the production timing independent of spawnpoint
+			var startPos = self.Location + new CVec(owner.World.Map.Bounds.Width, 0);
+			var endPos = new CPos(owner.World.Map.Bounds.Left - 5, self.Location.Y);
+
+			// Assume a single exit point for simplicity
+			var exit = self.Info.TraitInfos<ExitInfo>().First();
+
+			foreach (var tower in self.TraitsImplementing<INotifyDelivery>())
```
Phrohdoh: This is a bad variable name because it has no real merit here (many things could be notified, it isn't always control), could you please rename it?

I'd prefer `notify`.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+
+namespace OpenRA.Mods.Common
+{
+
+	[Desc("Deliver the unit in production via paradrop.")]
+	public class ProductionParadropInfo : ProductionInfo
+	{
+		public readonly string ReadyAudio = "ReinforcementsArrived";
+		[Desc("Cargo aircraft used.")]
+		[ActorReference] public readonly string ActorType = "c17";
```
Phrohdoh: It should be obvious, but I don't like to leave 'obvious' things lying around.
Since this actor must fulfill the `Fly` activity please note in a description that the actor must have the `Aircraft` trait.

Source:
```csharp
		public Fly(Actor self, Target t)
		{
			plane = self.Trait<Aircraft>();
			target = t;
		}
```

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+				td.Add(new CenterPositionInit(spawn));
+				td.Add(new FacingInit(initialFacing));
+			}
+
+			self.World.AddFrameEndTask(w =>
+			{
+				if (factionVariant != null)
+					td.Add(new FactionInit(factionVariant));
+				
+				var newUnit = self.World.CreateActor(producee.Name, td);
```
Phrohdoh: For safety you could check this `newUnit` for null.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+				if (move != null)
+				{
+					if (exitinfo.MoveIntoWorld)
+					{
+						if (exitinfo.ExitDelay > 0)
+							newUnit.QueueActivity(new Wait(exitinfo.ExitDelay, false));
+
+							newUnit.QueueActivity(move.MoveIntoWorld(newUnit, exit));
+							newUnit.QueueActivity(new AttackMoveActivity(
+							newUnit, move.MoveTo(exitLocation, 1)));
```
Phrohdoh: This needs to be indented if not included on the above line.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+				var notifyOthers = self.World.ActorsWithTrait<INotifyOtherProduction>();
+				foreach (var notify in notifyOthers)
+					notify.Trait.UnitProducedByOther(notify.Actor, self, newUnit);
+				
+				foreach (var t in newUnit.TraitsImplementing<INotifyBuildComplete>())
+					t.BuildingComplete(newUnit);
+			});
+		}
+	}
+
```
Phrohdoh: Extra line here and at the end.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
Bynnar18: Most of this code, this included, is basically copied from Production. I just needed to add Parachute to the new unit and move it to spawning in the air.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
Bynnar18: Again, it's from Production. I can add a check but I think it should be a separate PR if we're doing fixes on Production as well.

Bynnar18:
Ok fixed more formatting errors since those were easy enough to fix. I still got to fix the above issues.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
Phrohdoh: I understand that, but that doesn't mean you cannot improve on it.

Bynnar18:
Ok, the issues above have been fixed as well as adding a null check for ignore with parachutable. The exiting issue happened to be with my test case and the spawning issue to be with the queued parachute activity. I also squashed the commits.

atlimit8:
What about https://github.com/OpenRA/OpenRA/pull/10266#discussion_r48386178?

Bynnar18:
Wow, I thought I'd fixed that. It should be fixed now (unless my squash screwed something up).

Bynnar18:
Ok, I rebased and removed that useless merge commit.

Mailaender:
Can you create a test map where this is used and upload it to http://resource.openra.net?

Bynnar18:
How exactly would I do that? Edit an existing map or create a new one with a building using this trait?

GraionDilach:
Whichever you prefer - as long as it has a use-case inside (either preplaced either buildable) it'll do.

Bynnar18:
Ok, here's one: http://resource.openra.net/maps/8274/

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,167 @@
...
+using System;
+using System.Drawing;
+using System.Linq;
+using OpenRA.Activities;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Primitives;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common
```
Mailaender: This should be `OpenRA.Mods.Common.Traits`

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,167 @@
...
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Drawing;
+using System.Linq;
+using OpenRA.Activities;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Mods.Common.Traits;
```
Mailaender: You can then remove this line.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,167 @@
...
+				actor.QueueActivity(new CallFunc(() =>
+					{
+						if (!self.IsInWorld || self.IsDead)
+							return;
+
+						foreach (var cargo in self.TraitsImplementing<INotifyDelivery>())
+							cargo.Delivered(self);
+
+						self.World.AddFrameEndTask(ww => DoProduction(self, producee, exit, factionVariant));
+						Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Speech", info.ReadyAudio, self.Owner.Faction.InternalName);
```
Mailaender: This is a bit weird. You get `Unit ready` although the airplane hasn't arrived yet.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,167 @@
...
+using OpenRA.Activities;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Primitives;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common
+{
+	[Desc("Deliver the unit in production via paradrop.")]
+	public class ProductionParadropInfo : ProductionInfo
```
Mailaender: Needs a `Requires<ExitInfo>` as you have `var exit = self.Info.TraitInfos<ExitInfo>().First();` below.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,167 @@
...
+			var actorType = info.ActorType;
+
+			owner.World.AddFrameEndTask(w =>
+			{
+				if (!self.IsInWorld || self.IsDead)
+					return;
+
+				var altitude = self.World.Map.Rules.Actors[actorType].TraitInfo<AircraftInfo>().CruiseAltitude;
+				var actor = w.CreateActor(actorType, new TypeDictionary
+					{
```
Mailaender: These extra indentions are very confusing.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,167 @@
...
+						new CenterPositionInit(w.Map.CenterOfCell(startPos) + new WVec(WDist.Zero, WDist.Zero, altitude)),
+						new OwnerInit(owner),
+						new FacingInit(64)
```
Mailaender: For bonus points you could move this magic value into the TraitInfo to make it configurable.

Mailaender:
Some small changes required. Otherwise works as promised. :white_check_mark:

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
penev92: Yes please.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
Bynnar18: This is in line with what ProductionAirdrop does. I can still remove it but I don't think it needs to be removed.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
Bynnar18: I'll leave that for a separate PR since it will require reworking where the plane is spawned, not just adding a facing variable.

Bynnar18:
Ok, I've fixed those things. The only thing really that could be improved is the facing but that'd require reworking the plane spawning so I'll leave that for a future PR.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Deliver the unit in production via paradrop.")]
+	public class ProductionParadropInfo : ProductionInfo, Requires<ExitInfo>
+	{
+		[Desc("Cargo aircraft used. Must have Aircraft trait.")]
+		[ActorReference] public readonly string ActorType = "badr";
+		[Desc("Sound to play when dropping the unit.")]
```
Phrohdoh: Please include empty lines between these pairs now since you're here.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,170 @@
...
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Primitives;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Deliver the unit in production via paradrop.")]
+	public class ProductionParadropInfo : ProductionInfo, Requires<ExitInfo>
+	{
+		[Desc("Cargo aircraft used. Must have Aircraft trait.")]
```
Phrohdoh: You no longer need to have the second sentence (but can keep it since this feature isn't documented).
Instead you can give `ActorReference` a type that is required.
`[ActorReference(typeof(AircraftInfo))]`

Bynnar18: Is there an example of where this is used in another trait? I get a compile error when adding that saying ActorReferenceAttribute does not contain a constructor that takes 1 argument.

Phrohdoh: Have you rebased within the past week?
#10302

Bynnar18: Ah ok, I'll rebase and fix that then.

Bynnar18:
I implemented the fixes Phrohdoh mentioned but I couldn't get it to build for some reason on my end. It didn't seem like my repository rebased properly.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,172 @@
...
+
+				var move = newUnit.TraitOrDefault<IMove>();
+				if (move != null)
+				{
+					if (exitinfo.MoveIntoWorld)
+					{
+						if (exitinfo.ExitDelay > 0)
+							newUnit.QueueActivity(new Wait(exitinfo.ExitDelay, false));
+
+							newUnit.QueueActivity(move.MoveIntoWorld(newUnit, exit));
```
abcdefg30: The indentation here and below is one level too much.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -0,0 +1,172 @@
...
+				td.Add(new FacingInit(initialFacing));
+			}
+
+			self.World.AddFrameEndTask(w =>
+			{
+				if (factionVariant != null)
+					td.Add(new FactionInit(factionVariant));
+
+				var newUnit = self.World.CreateActor(producee.Name, td);
+
```
abcdefg30: Can you drop this new line, please?

obrakmann:
This looks fine to me, :+1: 

@Bynnar18, can we get a rebase and a fixup for the two whitespace issues, please?

Then this can get merged.

Bynnar18:
Sorry, I'll return to this sometime, rebase it and get it merged but right now I'm have real life stuff to attend to and don't really have the time even for the simple fixes so I'm closing this for now. Someone can fork it and fix it up if they wish.

obrakmann:
Rebased and fixed up: https://github.com/OpenRA/OpenRA/compare/bleed...obrakmann:pr10266_paradrop-production
I'll pr this probably some time this weekend. It would be a shame to leave this hanging just short of the finish line.

Issue #10226, opened 2015-12-14
Spy gets detected by enemy soldiers and flame towers


dkessel:
I am using playtest 20151213
When playing the RA single player mission "Tanya's Tale", the spy gets detected by the soldiers and the flame turrets.
This makes completing the mission very hard, if not impossible.

I believe in the original game, the soldiers did not detect a spy. I am unsure about the flame towers, though.

obrakmann:
Stupid question, but did you disguise the spy?

dkessel:
nevermind :) it's really been a while

abcdefg30:
I filed #10267 to make it clearer that you have to disguise your spy.

PR #10267, opened 2015-12-23
Adjust the description of allies05a

### LINKED ISSUE #10265: Spy gets shot
OpenRA.exe ver 12/13/2015. Mission 5a, Tanya's Tale: spy gets noticed and shot by soldiers.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29349204-spy-gets-shot?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

abcdefg30:
To prevent issues like #10226 and #10265.
-> Fixes #10265.

In addition we could disguise the spy automatically at mission start, if you feel like this change in the description is not enough.

### DIFF
--- a/mods/ra/maps/allies-05a/allies05a.lua
+++ b/mods/ra/maps/allies-05a/allies05a.lua
@@ -213,6 +213,10 @@ SendSpy = function()
 		SpyCameraA = Actor.Create("camera", true, { Owner = greece, Location = SpyCamera1.Location })
 		SpyCameraB = Actor.Create("camera", true, { Owner = greece, Location = SpyCamera2.Location })
 	end
+
+	Trigger.AfterDelay(DateTime.Seconds(3), function()
+		Media.DisplayMessage("Commander! You have to disguise me in order to get through the enemy patrols.", "Spy")
+	end)
 end
 
 ActivatePatrols = function()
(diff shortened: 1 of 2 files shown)


GraionDilach:
It's still kinda confusing, as the tooltip would still imply that the spy would pass all enemy units while only the disguised enemy player's allies would not shoot him.

But I guess people are just such stupid.

ScottNZ:
An actual tutorial popup would help the most with this I think

abcdefg30:
Added a small message to the beginning of the mission, as discussed on IRC.

obrakmann:
:+1:

Bynnar18:
While I do agree this is good, I think the message could use some rewording. Something like "Commander! I'll need a disguise to get past enemy patrols." Since technically the commander is suppose to just be giving the orders to get a disguise. Right now it's a little too direct, but maybe that's just me.

atlimit8:
Please add "fixes #10226" and "fixes #10265" or equivalent to the description so that they are closed when this is merged.

Mailaender:
Good idea. Works as promised. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/6a1a64209e1c1229ca5025d8cf5da6417d2f1f7d...a888a1fcf9db91b1924696e23c4ea01f11cc9c0f)

PR #10269, opened 2015-12-23
Load shaders relative to the game directory, rather than the CWD

RoosterDragon:
If OpenRA is started with a current working directory that does not match the path to the game, then loading shaders will fail. By using a path relative to the game directory, we can successfully start the game with different working directories.

(I'm mostly scratching my own itch here - the memory profiler I use changes the working directory and thus gives me exceptions like `System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\Users\Roost\CLRProfiler\32\glsl\shp.vert'.`)

### DIFF
--- a/OpenRA.Platforms.Default/Shader.cs
+++ b/OpenRA.Platforms.Default/Shader.cs
@@ -25,7 +25,7 @@ namespace OpenRA.Platforms.Default
 		protected int CompileShaderObject(ShaderType type, string name)
 		{
 			var ext = type == ShaderType.VertexShader ? "vert" : "frag";
-			var filename = "glsl{0}{1}.{2}".F(Path.DirectorySeparatorChar, name, ext);
+			var filename = Path.Combine(Platform.GameDir, "glsl", name + "." + ext);
 			var code = File.ReadAllText(filename);
 
 			var shader = GL.CreateShader(type);


OpenRA.Platforms.Default/Shader.cs
```diff
@@ -25,7 +26,7 @@ class Shader : ThreadAffine, IShader
 		protected int CompileShaderObject(ShaderType type, string name)
 		{
 			var ext = type == ShaderType.VertexShader ? "vert" : "frag";
-			var filename = "glsl{0}{1}.{2}".F(Path.DirectorySeparatorChar, name, ext);
+			var filename = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "glsl", name + "." + ext);
```
pchote: Can you use `Platform.GameDir` for consistency?

RoosterDragon: Ah, sure.

atlimit8:
:+1: There (currently) is no reason for the current working directory to be used by the game for anything.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ec7ecf178936d4b22e4f4bba0f2478b9924369c0...ff860ccfbc09b81ff27474339329c93b2e5e3a54)

PR #10271, opened 2015-12-23
Support sending sequences of points when drawing lines

RoosterDragon:
This allows the graph widgets to avoid having to create temporary arrays just to draw these lines, which reduces GC pressure.

On the RA shellmap with the perf graph active, this reduces ongoing allocations by about 9%.

### DIFF
--- a/OpenRA.Game/Graphics/RgbaColorRenderer.cs
+++ b/OpenRA.Game/Graphics/RgbaColorRenderer.cs
@@ -11,6 +11,7 @@
 using System;
 using System.Collections.Generic;
 using System.Drawing;
+using System.Linq;
 
 namespace OpenRA.Graphics
 {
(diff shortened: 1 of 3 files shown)


OpenRA.Game/Graphics/RgbaColorRenderer.cs
```diff
@@ -178,18 +196,12 @@ void DrawConnectedLine(float2[] points, float width, Color color, bool closed)
...
-				if (points.Length < 2)
-					return;
-
-				for (var i = 1; i < points.Length; i++)
-					DrawLine(points[i - 1], points[i], width, color);
-			}
+				DrawDisconnectedLine(points, width, color);
 			else
-				DrawConnectedLine(points, width, color, false);
+				DrawConnectedLine(points as float2[] ?? points.ToArray(), width, color, false);
```
Phrohdoh: What is the reasoning behind expecting `as float2[]` to fail and produce `null`, if we are able to call `.ToArray()` on it?

RoosterDragon: Performance optimization if somebody sent an array in the first place. If they did, we just cast rather than having to create a second copy of the array.

Phrohdoh: That sounds like trying to reach really far for perf, but okay.

RoosterDragon: Casting is constant time, creating an array is linear time. It's not that much of a stretch.

Mailaender:
:white_check_mark:

pchote:
:+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/557ccfc9ef3fc6e9f239bf3fad502e35397fd3b2

PR #10272, opened 2015-12-23
Speed up dictionaries keyed on Player

RoosterDragon:
Adds a specialized collection, `PlayerDictionary`, that also stores state in an array accessible by player index. By relying on the fact the players in the world never change, in some performance sensitive parts of the code we can exchange slower dictionary access for speedier array accesses instead. This change was already made in #10126 - I am now expanding it into a few other areas, so I've created a dedicated object for it.

This change reduces CPU used in the RA shellmap from 1.75% to 0.74% in the affected methods (`FrozenUnderFog.Tick`, `FrozenUnderFog.TickRender` and `GpsDot.Tick`)

### DIFF
--- a/OpenRA.Game/Primitives/PlayerDictionary.cs
+++ b/OpenRA.Game/Primitives/PlayerDictionary.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace OpenRA.Primitives
+{
+	/// <summary>
+	/// Provides a mapping of players to values, as well as fast lookup by player index.
+	/// </summary>
+	public struct PlayerDictionary<T> : IReadOnlyList<T>, IReadOnlyDictionary<Player, T> where T : class
+	{
+		readonly T[] valueByPlayerIndex;
+		readonly Dictionary<Player, T> valueByPlayer;
+
+		public PlayerDictionary(World world, Func<Player, T> valueFactory)
+		{
+			var players = world.Players;
+			if (players.Length == 0)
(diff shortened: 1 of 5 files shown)


OpenRA.Game/Primitives/PlayerDictionary.cs
```diff
@@ -0,0 +1,49 @@
...
+		{
+			var players = world.Players;
+			if (players.Length == 0)
+				throw new InvalidOperationException("The players in the world have not yet been set.");
+
+			// The world players never change, so we can safely maintain an array of values.
+			// We need to enforce T is a class, so if it changes that change is available in both collections.
+			valueByPlayerIndex = new T[players.Length];
+			valueByPlayer = new Dictionary<Player, T>(players.Length);
+			for (var i = 0; i < players.Length; i++)
```
Phrohdoh: If we're going for performance in places like this shouldn't you pull out the upper limit?
`int i = 0, max = players.Length; ...`

RoosterDragon: The slow part of the loops are the memory accesses in the body. Checking the loop condition simply isn't slow.

Mailaender:
I build a GPS satellite, seems to work fine. Sold it and discovered that enemy buildings frozen under fog will now disappear upon destruction. Is this a regression?

Mailaender:
Apparently not as this also happens in the release.

Mailaender:
Can't spot problems with GPS and frozen under fog. :white_check_mark:

PR #10279, opened 2015-12-25
Updated to .NET 4.5

Mailaender:
On August 7, 2014, Microsoft announced that support will end for .NET Framework 4, 4.5, and 4.5.1 on January 12, 2016. https://support.microsoft.com/en-gb/lifecycle?C2=548

### DIFF
--- a/packaging/windows/OpenRA.nsi
+++ b/packaging/windows/OpenRA.nsi
@@ -148,14 +148,14 @@ Section "-DotNet" DotNet
 	IfErrors error 0
 	IntCmp $0 1 done error done
 	error:
-		MessageBox MB_YESNO ".NET Framework v4.0 or later is required to run OpenRA. $\n \
+		MessageBox MB_YESNO ".NET Framework v4.5 or later is required to run OpenRA. $\n \
 		Do you wish for the installer to launch your web browser in order to download and install it?" \
 		IDYES download IDNO error2
 	download:
-		ExecShell "open" "http://www.microsoft.com/en-us/download/details.aspx?id=17113"
+		ExecShell "open" "http://www.microsoft.com/en-us/download/details.aspx?id=30653"
 		Goto done
 	error2:
-		MessageBox MB_OK "Installation will continue, but be aware that OpenRA will not run unless .NET v4.0 \
+		MessageBox MB_OK "Installation will continue, but be aware that OpenRA will not run unless .NET v4.5 \
 		or later is installed."
 	done:
 SectionEnd
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@
 
 ############################## TOOLCHAIN ###############################
 #
-CSC         = dmcs
+CSC         = mcs -sdk:4.5
 CSFLAGS     = -nologo -warn:4 -codepage:utf8 -unsafe -warnaserror
 DEFINE      = TRACE
 COMMON_LIBS = System.dll System.Core.dll System.Data.dll System.Data.DataSetExtensions.dll System.Drawing.dll System.Xml.dll thirdparty/download/ICSharpCode.SharpZipLib.dll thirdparty/download/FuzzyLogicLibrary.dll thirdparty/download/Mono.Nat.dll thirdparty/download/MaxMind.Db.dll thirdparty/download/MaxMind.GeoIP2.dll thirdparty/download/Eluant.dll thirdparty/download/SmarIrc4net.dll
--- a/OpenRA.Game/OpenRA.Game.csproj
+++ b/OpenRA.Game/OpenRA.Game.csproj
@@ -31,6 +31,7 @@
     <ApplicationVersion>1.0.0.%2a</ApplicationVersion>
     <UseApplicationTrust>false</UseApplicationTrust>
     <BootstrapperEnabled>true</BootstrapperEnabled>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x86'">
     <DebugSymbols>true</DebugSymbols>
--- a/OpenRA.GameMonitor/OpenRA.GameMonitor.csproj
+++ b/OpenRA.GameMonitor/OpenRA.GameMonitor.csproj
@@ -7,6 +7,7 @@
     <OutputType>WinExe</OutputType>
     <RootNamespace>OpenRA</RootNamespace>
     <AssemblyName>OpenRA</AssemblyName>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x86'">
     <DebugSymbols>true</DebugSymbols>
--- a/OpenRA.Mods.Cnc/OpenRA.Mods.Cnc.csproj
+++ b/OpenRA.Mods.Cnc/OpenRA.Mods.Cnc.csproj
@@ -30,6 +30,7 @@
     <IsWebBootstrapper>false</IsWebBootstrapper>
     <UseApplicationTrust>false</UseApplicationTrust>
     <BootstrapperEnabled>true</BootstrapperEnabled>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x86'">
     <DebugSymbols>true</DebugSymbols>
@@ -66,9 +67,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
-    <Reference Include="System.Core">
-      <RequiredTargetFramework>4.0</RequiredTargetFramework>
-    </Reference>
+    <Reference Include="System.Core" />
     <Reference Include="System.Drawing" />
     <Reference Include="Eluant">
       <HintPath>..\thirdparty\download\Eluant.dll</HintPath>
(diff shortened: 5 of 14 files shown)


pchote:
This removes compatibility with Windows XP, which means that @normannorman and other players will be forced out of the community.  My opinion has always been that increasing the minimum requirements should only ever be done as a last resort when it provides strong benefits that can't be achieved in other ways.

Mailaender:
He could try to use http://www.mono-project.com/download/#download-win on Windows XP.

packaging/windows/OpenRA.nsi
```diff
@@ -147,14 +147,14 @@ Section "-DotNet" DotNet
...
+		MessageBox MB_YESNO ".NET Framework v4.5.2 or later is required to run OpenRA. $\n \
 		Do you wish for the installer to launch your web browser in order to download and install it?" \
 		IDYES download IDNO error2
 	download:
-		ExecShell "open" "http://www.microsoft.com/en-us/download/details.aspx?id=17113"
+		ExecShell "open" "http://www.microsoft.com/en-us/download/details.aspx?id=42643"
 		Goto done
 	error2:
-		MessageBox MB_OK "Installation will continue, but be aware that OpenRA will not run unless .NET v4.0 \
+		MessageBox MB_OK "Installation will continue, but be aware that OpenRA will not run unless .NET v4.5.2 \
```
pchote: Can we make this a proper sentence while you're at it?

Mailaender: It line breaks into a proper sentence: `or later is installed."`

penev92:
- [ ] This seems to be changing us to .NET 4.5, not 4.5.2.
 - [ ] If the "mono on windows" thing works, I'd be fine with this. Will test on XP when I get my hands on one. If someone else has Win XP handy, please go ahead and test.

pchote:
While I don't like this, on the balance of things I think that this + #10280, #10281 give enough benefits to make this worthwhile as long as XP and mono 2.10 are the only things we kill support for.

In particular, we need to check whether mono 3.2 still works, which is our current minreq for OSX and debian-based linux.

pchote:
Confirmed that this still works (with both #10280 and #10281) on mono 3.2.0 and 3.2.8, so this does not affect our minimum requirements for OSX or debian-based linuxes.

pchote:
Adding fixup requested for the 4.5 vs 4.5.2 thing.

pchote:
Adding has dependencies for testing mono on windows xp and checking the new minimum requirements for VS and monodevelop.

Mailaender:
https://github.com/OpenRA/OpenRA/pull/10280 will do/need the update to 4.5.2 then.

pchote:
That's a bad idea - you're introducing a rebase conflict and doubling the amount of work that we have to do to test the runtime.  Please go straight to 4.5.2 here, or lets close this pr in favor of #10280.

pchote:
Ok, I take the above back.  Mono doesn't explicitly install 4.5.2 support, so specifying that in the csproj breaks monodevelop.

pchote:
Please update the Debian `control` file to reflect the new minimum version requirement (>= 3.2, and possibly the named support packages).

Mailaender:
Updated the control file. The support packages stay *4.0-cil if I am not mistaken.

huwpascoe:
> We should make sure we don't drop support for Visual Studio 2012! I'm fine with dropping support for VS 2010 (if we accidentally still support it). I will test when I get my hands on one. If someone else has VS 2012 handy, please go ahead and test.

The .net version supported by 2012 is 4.5.0

VS 2015 is the current version and supports Windows 7 SP1, I use the community edition which does away with the separate "express" model of an IDE for each language in favor of the full thing which is so much easier. With that in mind, is it still worth supporting VS2012?

Phrohdoh:
What is the oldest Windows (and SP) that 2012 supports?
I think that is the deciding factor, not the newest version of the software.

Mailaender:
> The .Net version supported by 2012 is 4.5.0

Then we will still support it even after this is merged.

penev92:
>With that in mind, is it still worth supporting VS2012?

Yes, it is. If I want tto stick to my VS 2012 Ultimate instead of switching to 2015 Community, that's a reason.

pchote:
VS2012 requires Win7 or greater, so there's no problem there.

RoosterDragon:
If we're going to require mono for XP, can we get the installer to detect this and help out somehow?

Mailaender:
I am just testing Mono on Windows XP and it isn't really looking good.

http://download.mono-project.com/archive/3.2.3/windows-installer/ runs, but doesn't start the game:

```
Operating System: Windows (Microsoft Windows NT 5.1.2600.196608 Service Pack 3)
Runtime Version: Mono 3.3.0 (master/6cd4ddc) CLR 4.0.30319.17020
Exception of type `System.TypeInitializationException`: An exception was thrown by the type initializer for OpenRA.Support.PerfHistory
Inner Exception of type `System.TypeInitializationException`: An exception was thrown by the type initializer for System.Drawing.KnownColors
  Inner Exception of type `System.TypeInitializationException`: An exception was thrown by the type initializer for System.Drawing.GDIPlus
    Inner Exception of type `System.DllNotFoundException`: /tmp/install/lib/libgdiplus.so
      TypeName=``
        at (wrapper managed-to-native) System.Drawing.GDIPlus:GdiplusStartup (ulong&,System.Drawing.GdiplusStartupInput&,System.Drawing.GdiplusStartupOutput&)
  at System.Drawing.GDIPlus..cctor () [0x00000] in <filename unknown>:0 
      at System.Drawing.KnownColors..cctor () [0x00000] in <filename unknown>:0 
    at System.Drawing.Color.get_Red () [0x00000] in <filename unknown>:0 
  at OpenRA.Support.PerfHistory..cctor () [0x00000] in <filename unknown>:0 
  at OpenRA.Game.Initialize (OpenRA.Arguments args) [0x00000] in <filename unknown>:0 
  at OpenRA.Program.Run (System.String[] args) [0x00000] in <filename unknown>:0 
  at OpenRA.Program.Main (System.String[] args) [0x00000] in <filename unknown>:0 
```

and Mono 3.12.X upwards won't work with Windows XP anymore and no one will invest time to fix it on @Xamarin side: http://mono.1490590.n4.nabble.com/Can-t-run-mono-3-12-X-on-Windows-XP-td4665759.html

huwpascoe:
> Yes, it is. If I want tto stick to my VS 2012 Ultimate instead of switching to 2015 Community, that's a reason.

:+1: for vs 2012 support in that case.

@Mailaender Seems you're going to have to make a choice between supporting the retired system and using new technology, it's going to be increasingly difficult to bridge that gap as software advances.

Mailaender:
See https://bugzilla.xamarin.com/show_bug.cgi?id=8309 for the upstream bug. It is as simple as fixing a configuration file, but @Xamarin never released Mono 3.4.0 for Windows and left Mono 3.2.8 broken.

Mailaender:
It runs on Windows XP using the Mono 3.4.0 binaries from http://www.codeproject.com/Articles/769292/How-to-build-Mono-on-Windows

mono-3 4 0-windows-xp

GraionDilach:
This will probably need some docs for WinXP users though - I guess `make dependencies && make all`on WinXP wouldn't trigger `mcs` but `msbuild`, which prolly would be helpful for XP users who want to check out bleed for a bug or a fix.

Mailaender:
I updated https://github.com/OpenRA/OpenRA/wiki/FAQ#unable-to-find-a-version-of-the-runtime-to-run-this-application and added https://github.com/OpenRA/OpenRA/wiki/Windows-XP

pchote:
I'd like to go ahead with this, but we really should provide a trusted (compiled and hosted ourselves) version of mono, and have the windows installer detect XP and download the mono installer (in the same way that it used to download the game assets) and adjust the installed shortcuts to use it.  I expect that the installer tweaks should be simple, because it's just copying and modifying (previously-)existing script.

Mailaender:
I asked @Xamarin to release a build at https://bugzilla.xamarin.com/show_bug.cgi?id=8309#c9 They already have the infrastructure in place to make them.

penev92:
Couldn't get it to run on XP using the instructions on the wiki. :/

>The assembly mscorlib.dll was not found or could not be loaded.
It should have been installed in the `C:/cygwin/usr/local/lib\mono\4.5\mscorlib.dll' directory.

Mailaender:
It must have worked with https://github.com/OpenRA/OpenRA/pull/10299, but not this pull request then. I probably got confused as AppVeyor only produces OpenRA-.exe files. Sorry.

Okay, I guess we are effectively removing Windows XP support with this. I also checked: there is literally no way of getting .NET 4.5 to run on Windows XP. They changed something in their debugging tool chain. Even force installing .NET 4.5 will just end up in a broken .NET on XP. Officially http://blogs.msdn.com/b/dotnet/p/dotnet45xp.aspx and unofficially this won't work.

I don't really want to play the numbers game. Also @x-a-n-a-x and @chrisforbes payed for Facebook ads that ran in developing countries. It seems a bit unfair to tease them and then stop supporting them.

Mailaender:
> I probably got confused as AppVeyor only produces OpenRA-.exe files. Sorry.

Decided to fix this in https://github.com/OpenRA/OpenRA/pull/10603.

rapgro:
> It runs on Windows XP using the Mono 3.4.0 binaries from http://www.codeproject.com/Articles/769292/How-to-build-Mono-on-Windows

Is that a joke? Come on, we have 2016!

Mailaender:
@pchote might allow to drop Windows XP support, but only after gathering stats on how many players would be affected https://github.com/OpenRA/OpenRA/issues/10662 which is effectively another barrier we have to take after my terrible trip of getting modern Mono to work on XP. This will probably take much longer, sorry.

Mailaender:
As we gathered a lot of statistics and the Windows XP user number seems to be dropping further, I made a new attempt: https://github.com/OpenRA/OpenRA/pull/11284

Issue #8264, opened 2015-05-26
Fedora packages or COPR repo


rapgro:
Fedora has strange issues to provide an usable Mono environment, at least for Mono versions 3 and 4.
Request is to provide all OpenRA packages incl. unbundling at COPR via a clean repository. SUSE OpenBuild Service (OBS) can not deal with partly COPR based packages. So let's move the whole stuff, at least for Fedora concerns, to COPR.

See also:
https://github.com/OpenRA/OpenRA/issues/7140
https://github.com/OpenRA/OpenRA/issues/6863
##

rapgro:
There are semi-official builds at Copr for both Mono 3 and 4. Maybe use those repos to provide official Fedora assemblies from OBS?
Mono4 https://copr.fedoraproject.org/coprs/tpokorra/mono/
Mono3 https://copr.fedoraproject.org/coprs/elsupergomez/mono/

Mailaender:
:+1: 

I currently maintain it at https://build.opensuse.org/project/show/games:openra but as I don't run Fedora myself those packages are completely untested. Started to unbundle some dependencies, but I haven't finished the quest yet. I also know few about the Fedora packaging specialties so I am glad an expert can take over and relieve me of it. Once you are done, we may want to advertise your repositories at https://github.com/OpenRA/OpenRAWeb.

You will run into compilation errors from time to time as we don't test our builds against Mono 2.10 anymore and got a WONTFIX from @Xamarin on https://bugzilla.xamarin.com/show_bug.cgi?id=30256 However the patches are usually simple and we will try to fix those compilation errors due to oversights in our code ASAP and can help you there.

I was told that Mono 4 is currently work in progress and @tpokorra is currently updating everything at https://github.com/tpokorra/lbs-mono-fedora It was even noteworthy https://www.phoronix.com/scan.php?page=news_item&px=Mono-4-Fedora-23

rapgro:
Thank for your support. IMHO it's not realistic to expect Mono 4 in a good state and for Fedora 23, because of all the needed bootstrapping from discontinued Mono 2 binaries. Maybe it's a good decision to build OpenQA packages for Mono 3 at the moment.
https://fedoraproject.org/wiki/Changes/Mono_4

rapgro:
https://github.com/OpenRA/OpenRA/issues/9254

rapgro:
https://fedoraproject.org/wiki/Changes/Mono_4
[...]

PR #10280, opened 2015-12-25
Updated to MaxMind.GeoIP2 2.5.0 and .NET 4.5.2

Mailaender:
Depends on https://github.com/OpenRA/OpenRA/pull/10279.

This has several benefits for proper Linux packaging #8264:
- Library gets a strong name so it can be installed into GAC.
- Removes the dependency on RestSharp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -403,7 +403,6 @@ install-core: default
 	@$(INSTALL_PROGRAM) MaxMind.Db.dll "$(DATA_INSTALL_DIR)"
 	@$(INSTALL_PROGRAM) MaxMind.GeoIP2.dll "$(DATA_INSTALL_DIR)"
 	@$(INSTALL_PROGRAM) Newtonsoft.Json.dll "$(DATA_INSTALL_DIR)"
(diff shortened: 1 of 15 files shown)


abcdefg30:
Appveyor fails.

Mailaender:
Any idea why?

Mailaender:
I need to update to .NET 4.5.2 which my MonoDevelop complains about. Not yet sure how to fix that.

pchote:
Mono does not ship an explicit 4.5.2 framework, so we can't target that.  4.5 works fine there, but as we found the first time around that breaks msbuild.

Mailaender:
We will have to wait for https://github.com/lontivero/Open.NAT/pull/31 and bundle the necessary async framework DLLs ourselves then. That won't be much fun, but it is the only way to keep .NET 4.0 support.

rapgro:
Open.NAT Version 2.0.11 (commit a5c834d) is successfully packaged for Fedora.

Mailaender:
.NET 4.5.2 is not required anymore, so I am making a new attempt: https://github.com/OpenRA/OpenRA/pull/11285

Issue #8794, opened 2015-07-26
UPNP XML parsing crash.


pchote:
Reported at https://www.reddit.com/r/openra/comments/3emk67/new_error_openragameexe_has_stopped_working/

```
Red Alert Mod at Version release-20150614
Operating System: Windows (Microsoft Windows NT 6.2.9200.0)
Runtime Version: .NET CLR 4.0.30319.34209
Exception of type System.Xml.XmlException: An error occurred while parsing EntityName. Line 1, position 459.
at System.Xml.XmlTextReaderImpl.Throw(Exception e)
at System.Xml.XmlTextReaderImpl.Throw(String res, String arg)
at System.Xml.XmlTextReaderImpl.ParseEntityName()
at System.Xml.XmlTextReaderImpl.ParseEntityReference()
at System.Xml.XmlTextReaderImpl.Read()
at System.Xml.XmlLoader.LoadNode(Boolean skipOverWhitespace)
at System.Xml.XmlLoader.LoadDocSequence(XmlDocument parentDoc)
at System.Xml.XmlLoader.Load(XmlDocument doc, XmlReader reader, Boolean preserveWhitespace)
at System.Xml.XmlDocument.Load(XmlReader reader)
at System.Xml.XmlDocument.LoadXml(String xml)
at Mono.Nat.Upnp.MessageBase.Decode(UpnpNatDevice device, String message)
at Mono.Nat.Upnp.UpnpNatDevice.DecodeMessageFromResponse(Stream s, Int64 length)
at Mono.Nat.Upnp.UpnpNatDevice.EndMessageInternal(IAsyncResult result)
at Mono.Nat.Upnp.UpnpNatDevice.EndGetAllMappingsInternal(IAsyncResult result)
at System.Net.LazyAsyncResult.Complete(IntPtr userToken)
at System.Net.ContextAwareResult.CaptureOrComplete(ExecutionContext& cachedContext, Boolean returnContext)
at System.Net.ContextAwareResult.FinishPostingAsyncOp()
at System.Net.HttpWebRequest.BeginGetResponse(AsyncCallback callback, Object state)
at Mono.Nat.Upnp.UpnpNatDevice.EndGetAllMappingsInternal(IAsyncResult result)
at System.Net.LazyAsyncResult.Complete(IntPtr userToken)
at System.Net.ContextAwareResult.CaptureOrComplete(ExecutionContext& cachedContext, Boolean returnContext)
[...]

Issue #7140, opened 2014-12-16
Migrate to Open.NAT and replace Mono.Nat


rapgro:
The project Mono.Nat is nearly dead. 

Since mono-nat (respectively as the old name) is retired in Fedora and can not be wake up easily, I would like to know how far the plan is to migrate to Open.NAT as promised already in some other issue (number unkonwn?).

rapgro:
The nterry fork has a build problem cause of automake failure.

Can you please provide a patch to disable / remove the anyhow b0rken mono-nat feature? As it turns out, this dependency is mostly needed to support XP users. Fedora should find another way to handle NAT issues, IMHO.

Mailaender:
You can still use older Mono.Nat versions with OpenRA.

pchote:
We discussed this again in IRC, and I suggested that we could modify OpenRA to (gracefully) disable the UPNP feature if the mono.nat dll isn't found.  This would be a straightforward way to work around Fedora's strict packaging restrictions that are blocking us from being packaged upstream.

Mailaender:
Sure, but Mono.Nat is very easy to package: https://build.opensuse.org/package/show/openSUSE:Factory/mono-nat

pchote:
Yep, but according to the earlier discussions the whole problem here is that Fedora won't accept dead/unmaintained libraries into their repositories.

Mailaender:
FuzzyLogicLibrary is also "unmaintained". Eluant and SDL2-CS are both forked and the latter one is rolling release. Mono.Nat is the least problem with the actively developed https://github.com/nterry/Mono.Nat fork.

rapgro:
No. This bug is about Mono.NAT / Open.NAT only.


From downstream (Fedora): https://bugzilla.redhat.com/show_bug.cgi?id=1166897#c7

As it turns out, the github tarball luckily includes LICENSE and AUTHORS. Unfortunately, this tarball does not provide any Makefile nor Makefile.in files, but Makefile.am only. I tried to tweak that somehow with autoreconf or automake, but without any success so far.

Please help to use automake properly. Otherwise, I would suggest to leave mono-nat as dead as it is ("to not wake up any bad zombie").

For reference, I'll upload the spec file without any claim to fix FTBFS.

https://raphgro.fedorapeople.org/review/mono/mono-nat/mono-nat.spec


Executing(%build): /bin/sh -e /var/tmp/rpm-tmp.JMcPjo
+ umask 022
[...]

PR #10281, opened 2015-12-25
Dump Mono.NAT in favor of Open.NAT 2

Mailaender:
https://github.com/lontivero/Open.NAT/wiki/Why-Open.NAT

Closes https://github.com/OpenRA/OpenRA/issues/7140.
Fixes https://github.com/OpenRA/OpenRA/issues/8794.
Depends on https://github.com/OpenRA/OpenRA/pull/10279.

It also deprecates the `Settings.Server.NatDeviceAvailable` helper variable and always logs into `nat.log` (so we can get rid of `Settings.Server.VerboseNatDiscovery` as well) as even in verbose mode it is now pretty readable as Open.NAT only searches for NAT devices, not every other UPnP enabled device in the network.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@
 
 ############################## TOOLCHAIN ###############################
 #
-CSC         = dmcs
+CSC         = mcs -sdk:4.5
 CSFLAGS     = -nologo -warn:4 -codepage:utf8 -unsafe -warnaserror
 DEFINE      = TRACE
 COMMON_LIBS = System.dll System.Core.dll System.Data.dll System.Data.DataSetExtensions.dll System.Drawing.dll System.Xml.dll thirdparty/download/ICSharpCode.SharpZipLib.dll thirdparty/download/FuzzyLogicLibrary.dll thirdparty/download/Mono.Nat.dll thirdparty/download/MaxMind.Db.dll thirdparty/download/MaxMind.GeoIP2.dll thirdparty/download/Eluant.dll thirdparty/download/SmarIrc4net.dll
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@
 CSC         = mcs -sdk:4.5
 CSFLAGS     = -nologo -warn:4 -codepage:utf8 -unsafe -warnaserror
 DEFINE      = TRACE
-COMMON_LIBS = System.dll System.Core.dll System.Data.dll System.Data.DataSetExtensions.dll System.Drawing.dll System.Xml.dll thirdparty/download/ICSharpCode.SharpZipLib.dll thirdparty/download/FuzzyLogicLibrary.dll thirdparty/download/Mono.Nat.dll thirdparty/download/MaxMind.Db.dll thirdparty/download/MaxMind.GeoIP2.dll thirdparty/download/Eluant.dll thirdparty/download/SmarIrc4net.dll
+COMMON_LIBS = System.dll System.Core.dll System.Data.dll System.Data.DataSetExtensions.dll System.Drawing.dll System.Xml.dll thirdparty/download/ICSharpCode.SharpZipLib.dll thirdparty/download/FuzzyLogicLibrary.dll thirdparty/download/MaxMind.Db.dll thirdparty/download/MaxMind.GeoIP2.dll thirdparty/download/Eluant.dll thirdparty/download/SmarIrc4net.dll
 NUNIT_LIBS_PATH :=
 NUNIT_LIBS  := $(NUNIT_LIBS_PATH)nunit.framework.dll
 
@@ -107,7 +107,7 @@ endif
 game_SRCS := $(shell find OpenRA.Game/ -iname '*.cs')
 game_TARGET = OpenRA.Game.exe
 game_KIND = winexe
-game_LIBS = $(COMMON_LIBS) $(game_DEPS) thirdparty/download/SharpFont.dll
+game_LIBS = $(COMMON_LIBS) $(game_DEPS) thirdparty/download/SharpFont.dll thirdparty/download/Open.Nat.dll
 game_FLAGS = -win32icon:OpenRA.Game/OpenRA.ico
 PROGRAMS += game
 game: $(game_TARGET)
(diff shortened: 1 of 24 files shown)


abcdefg30:
> OpenRA.Game/Network/UPnP.cs:L41: [SA1000] The spacing around the keyword 'catch' is invalid.

Mailaender:
Fixed.

pchote:
This doesn't work for my router. Mono.NAT works fine:

```
Open.NAT Information: 0 : Start Discovery
Open.NAT Information: 0 : Searching for: UpnpSearcher
Open.NAT Error: 0 : Error searching UpnpSearcher - Details:
Open.NAT Error: 0 : System.Net.Sockets.SocketException: The requested address is not valid in this context
  at System.Net.Sockets.Socket.SendTo_nochecks (System.Byte[] buffer, Int32 offset, Int32 size, SocketFlags flags, System.Net.EndPoint remote_end) [0x0003f] in /private/tmp/source-mono-mac-4.2.0-branch/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.2/mcs/class/System/System.Net.Sockets/Socket.cs:2466 
  at System.Net.Sockets.Socket.SendTo (System.Byte[] buffer, Int32 offset, Int32 size, SocketFlags flags, System.Net.EndPoint remote_end) [0x00028] in /private/tmp/source-mono-mac-4.2.0-branch/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.2/mcs/class/System/System.Net.Sockets/Socket.cs:2454 
  at System.Net.Sockets.UdpClient.DoSend (System.Byte[] dgram, Int32 bytes, System.Net.IPEndPoint endPoint) [0x0001b] in /private/tmp/source-mono-mac-4.2.0-branch/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.2/mcs/class/System/System.Net.Sockets/UdpClient.cs:320 
Open.NAT Information: 0 : UPnP Response: Router advertised a 'urn:schemas-upnp-org:service:WANIPConnection:1' service!!!
Open.NAT Information: 0 : Found device at: http://192.168.1.254:58461/rootDesc.xml
Open.NAT Information: 0 : 192.168.1.254:58461: Fetching service list
Open.NAT Information: 0 : Found device at: http://192.168.1.254:58461/rootDesc.xml
Open.NAT Information: 0 : 192.168.1.254:58461: Parsed services list
Open.NAT Information: 0 : 192.168.1.254:58461: Found service: urn:schemas-upnp-org:service:WANIPConnection:1
Open.NAT Information: 0 : 192.168.1.254:58461: Found upnp service at: /ctl/IPConn
Open.NAT Information: 0 : 192.168.1.254:58461: Handshake Complete
Open.NAT Information: 0 : 192.168.1.254:58461: Absolute URI detected. Host address is now: 192.168.1.254:58461
Open.NAT Information: 0 : 192.168.1.254:58461: New control url: /ctl/IPConn
Open.NAT Information: 0 : Stop Discovery
Open.NAT Information: 0 : GetExternalIPAsync - Getting external IP address
Open.NAT Verbose: 0 : SOAPACTION: **GetExternalIPAddress** url:http://192.168.1.254:58461/ctl/IPConn
  at System.Threading.Tasks.Task.ThrowIfExceptional (Boolean includeTaskCanceledExceptions) [0x00014] in /private/tmp/source-mono-mac-4.2.0-branch/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.2/external/referencesource/mscorlib/system/threading/Tasks/Task.cs:2168 
  at System.Threading.Tasks.Task`1[TResult].GetResultCore (Boolean waitCompletionNotification) <0x1a18960 + 0x0008a> in <filename unknown>:0 
  at System.Threading.Tasks.Task`1[TResult].get_Result () <0x1a188e0 + 0x00038> in <filename unknown>:0 
  at OpenRA.Network.UPnP+<TryNatDiscovery>c__async0.MoveNext () [0x0010c] in /Users/paul/src/OpenRA/OpenRA.Game//Network/UPnP.cs:39 
```

pchote:
I suspect I am suffering from https://github.com/lontivero/Open.NAT/issues/26 (I have both `WANIPConnection` and `WANPPPConnection`), but the log files in that issue have expired so I can't be certain.

lontivero:
The opened bug in lontivero/Open.NAT#26 is because a HTTP 400 - Bad Request. Is this one because of the same problem?

Btw, Mono.Nat raises an event for each device it finds and that means when two are found we our event handler is invoked twice, one per device. That means that in many cases it fails with one but it succeeds with the other. Is this one of those cases?

pchote:
Looks like it is failing on `WANIPConnection:1` and succeeding on `WANPPPConnection:1`.  

@Mailaender's integration with OpenRA catches the exception that is generated by the first failure and globally disables NAT, so we don't see the results from the second device.  I can hack around the problem locally by removing the `WANIPConnection` services from `Open.Nat.UpnpSearcher`.

Mailaender:
So which exceptions exactly should I catch and which not?

Mailaender:
Shouldn't I try/catch in general here?

pchote:
@lontivero: Would you be able to advise on the best way to query the result of GetExternalIPAsync() while ignoring the first error?

lontivero:
Umm i think there was a misunderstanding here. I was talking about Mono.Nat. @pchote said Mono.Nat works okay and for that reason i detailed how it works (it raises an event for each device it finds).

Open.NAT doesn't raise events, it returns all the devices it finds before timeout and then you should try with them in a foreach loop.

Mailaender:
Updated to make use of DiscoverDevice**s**Async now.

pchote:
You'll still need a try/catch inside the loops to skip over the errors from the bogus device.

pchote:
Even with the try/catches I recieve the following, and ports are not forwarded:
```
Open.Nat.MappingException: Error 718: ConflictInMappingEntry
```
The port that I am trying to forward (tried both 1234 and 1238) are *not* already forwarded, despite what the error claims.

pchote:
Crash when loading the settings menu:
```
Inner Exception of type `System.InvalidOperationException`: ServerSettings does not contain a preference type VerboseNatDiscovery
```

Mailaender:
Updated.

Mailaender:
Not sure if this is async enough, but neither `Start()` nor `RunSynchronously()` will work here.

pchote:
Not async enough, sorry.  Put it in a background thread if you have to, or like I mentioned above at least move it later in the init so it doesn't delay the window creation and load screen display.

Mailaender:
Rebased and moved it below. I am a bit inexperienced with this async concept.

Mailaender:
Made a new attempt https://github.com/OpenRA/OpenRA/pull/11286 based on this pull request.

PR #10288, opened 2015-12-26
Replace MiniTK.Graphics with custom OpenGL shim.

pchote:
This replaces the ~30k line monster that's hidden in our custom sdl2-cs.dll with a minimal shim that only exposes the functionality that we actually need and uses the proper c-style conventions.  This removes the relatively long delay (at least on my machine) between window creation and first render, and gets us a step closer to being able to use the "official" sdl2-cs library.

Depends on #10249 which currently makes up the first 6 commits in this pr (don't review these here!).

### DIFF
--- a/OpenRA.Platforms.Default/OpenGL.cs
+++ b/OpenRA.Platforms.Default/OpenGL.cs
@@ -1,0 +1,413 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+using System.Text;
+using SDL2;
+
+namespace OpenRA.Platforms.Default
+{
+	[SuppressMessage("Microsoft.StyleCop.CSharp.NamingRules", "SA1300:ElementMustBeginWithUpperCaseLetter",
+		Justification = "C-style naming is kept for consistency with the underlying native API.")]
+	[SuppressMessage("Microsoft.StyleCop.CSharp.NamingRules", "SA1310:FieldNamesMustNotContainUnderscore",
+		Justification = "C-style naming is kept for consistency with the underlying native API.")]
+	internal static class OpenGL
+	{
+		public const int GL_FALSE = 0;
+
+		// ClearBufferMask
+		public const int GL_COLOR_BUFFER_BIT = 0x4000;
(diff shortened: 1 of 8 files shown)


pchote:
This may also fix #7622 (which was closed prematurely).

Mailaender:
I don't see any faster startup, but the more light weight dependencies are still very much appreciated. We should get rid of the unnecessary parts from SDL2-CS and OpenTK.OpenAL as well. :+1:

pchote:
Rebased.

pchote:
Updated to drop the OpenRA.Constants enum. The constants are now defined as constants on the class for consistency with OpenAL-CS, which will be introduced in the next pr.

OpenRA.Platforms.Default/OpenGL.cs
```diff
@@ -0,0 +1,413 @@
...
+	[SuppressMessage("Microsoft.StyleCop.CSharp.NamingRules", "SA1310:FieldNamesMustNotContainUnderscore",
+		Justification = "C-style naming is kept for consistency with the underlying native API.")]
+	internal static class OpenGL
+	{
+		public const int GL_FALSE = 0;
+
+		// ClearBufferMask
+		public const int GL_COLOR_BUFFER_BIT = 0x4000;
+		public const int GL_DEPTH_BUFFER_BIT = 0x0100;
+		public const int GL_STENCIL_BUFFER_BIT =	0x0400;
```
Phrohdoh: Spurious whitespace after `=`.

OpenRA.Platforms.Default/OpenGL.cs
```diff
@@ -0,0 +1,413 @@
...
+
+		// GL_ARB_imaging
+		public const int GL_FUNC_ADD = 0x8006;
+		public const int GL_FUNC_SUBTRACT = 0x800A;
+		public const int GL_FUNC_REVERSE_SUBTRACT = 0x800B;
+		public const int GL_BLEND_COLOR = 0x8005;
+
+		// OpenGL 1.1 - 1.5
+		public const int GL_CLIENT_PIXEL_STORE_BIT = 0x0001;
+		public const int GL_BGRA	= 0x80E1;
```
Phrohdoh: And here.

Phrohdoh:
The code looks good to me and everything renders ingame. :+1:

pchote:
Fixed.

Phrohdoh:
Merging with this comment from @reaperrr:

```
09:23 < reaperrr> [16:02] <+Phrohdoh> Hm, it may be good to get a windows +1 before merging.
09:23 < reaperrr> no problems on Win7x64
```

Phrohdoh:
[Changelog entry](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/090e4083480384f9fa5beff12b9e0fbe891aa645...1c45fff1d5392bf53a42841acef60d043b18e115).

PR #10299, opened 2015-12-27
Replace MiniTK.Audio with OpenAL-CS and update to latest SDL2-CS.

pchote:
This removes our last hard dependency on the legacy OpenTK code.

I have forked https://github.com/OpenRA/OpenAL-CS because according to @Mailaender the upstream author does not want the library to be included in nuget.


pchote:
Added a commit that switches us to the latest master branch of SDL2-CS.  The updated minitk-free dll is about 2.7MB smaller.

Phrohdoh:
Code is good and sound works fine for me. :+1:
cc @Mailaender for Linux
cc @reaperrr for Windows.

pchote:
Note to testers: be sure to do a make clean and delete anything in the thirdparty/download dir before doing a full rebuild.

Mailaender:
Sound still works for me as well. :white_check_mark:

AUTHORS
```diff
@@ -141,11 +141,8 @@ distributed under the CC BY-SA 3.0 license.
 Using SharpFont created by Robert Rouhani and
 distributed under the MIT license.
 
-Using the Open Toolkit distributed under the
-MIT license.
-
-Using SDL2# created by Ethan Lee and released
-under the zlib license.
+Using SDL2-CS and OpenAL-CS created by Ethan
+Lee and released under the zlib license.
```
penev92: Can you break the above line *after* his name?

pchote: No, not without breaking the rendering ingame.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7252081c1ed78e444571fbf29d075524dcf5c763)

PR #10301, opened 2015-12-27
Singles rebalance

ghost:
In relation to #10298 

map on resource centre http://resource.openra.net/maps/8270/ (parsed on bleed...)

I started with the map I felt needed most changes. Reasoning and diagrams in next comment. Most other maps should have fairly obvious changes so shouldn't be so tl;dr I hope.

Original
orig

Terrain and ore adjustment

Terrain, ore, trees and debris

This PR currently contains two commits;
-The terrain and ore adjusted
-The additional trees and islands (and some more minor adjustments)

I think the additional trees and debris added (sparingly) make a huge difference, but awaiting feedback before I squash.

PR #10303, opened 2015-12-27
Rebalance doubles.oramap in RA mod

ghost:
Since it was a simple update after redoing singles, here is doubles.

map on resource centre http://resource.openra.net/maps/8276/ (parsed on bleed...)

For those that wondered, apart from more spawns, the original (and this re-imagined) version of doubles has slightly more ore than singles.


r34ch:
Updated with fixed map size, decorations and improved spawn placements.

Mailaender:
Map borders are sane now, no tiling errors spotted and the additional details adds polishing. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7165d4bd0d23938c52a1315d86cf5d514827a154)

SoScared:
Following the same argument as shared on #10301, I made a version reverting some of the aesthetic add-on to allow for more build space. Also, altered the ore patches a bit:

http://resource.openra.net/maps/9487

doubles compare piz

doubles compare piz

PR #10304, opened 2015-12-27
Reworking ReplayRecorderConnection to be not be a Connection

whinis:
I made it so that the ReplayRecorder is no longer a connection itself but a component of certain connections. This allows one to keep a connection but create a new Replay by disposing of the old one and then starting a new one. 

This also removes a large amount of complexity from the recorder which was mostly forwarding to an actual connection and modifying some functions

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -49,9 +49,9 @@ namespace OpenRA
 
 		public static OrderManager JoinServer(string host, int port, string password, bool recordReplay = true)
 		{
-			IConnection connection = new NetworkConnection(host, port);
+			var connection = new NetworkConnection(host, port);
 			if (recordReplay)
-				connection = new ReplayRecorderConnection(connection, TimestampedFilename);
+				connection.StartRecording(TimestampedFilename);
 
 			var om = new OrderManager(host, port, password, connection);
 			JoinInner(om);
--- a/OpenRA.Game/Network/Connection.cs
+++ b/OpenRA.Game/Network/Connection.cs
@@ -44,6 +44,7 @@ namespace OpenRA.Network
 		}
 
 		protected List<ReceivedPacket> receivedPackets = new List<ReceivedPacket>();
+		public ReplayRecorder Recorder { get; private set; }
 
 		public virtual int LocalClientId
 		{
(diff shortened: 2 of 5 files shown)


OpenRA.Game/World.cs
```diff
@@ -211,7 +211,9 @@ public void LoadComplete(WorldRenderer wr)
 			foreach (var player in Players)
 				gameInfo.AddPlayer(player, OrderManager.LobbyInfo);
 
-			var rc = OrderManager.Connection as ReplayRecorderConnection;
+			var rc = (OrderManager.Connection is EchoConnection) ?
+				(OrderManager.Connection as EchoConnection).Recorder : null;
```
RoosterDragon:
```c#
var echo = OrderManager.Connection as EchoConnection;
var rc = echo != null ? echo.Recorder : null;
```

OpenRA.Game/Network/Connection.cs
```diff
@@ -107,7 +121,9 @@ public virtual void Receive(Action<int, byte[]> packetFn)
 		public void Dispose()
 		{
 			Dispose(true);
-			GC.SuppressFinalize(this);
+			if (Recorder != null)
+				Recorder.Dispose();
+            GC.SuppressFinalize(this);
```
RoosterDragon: You'll want to place this in the other method instead.
```c#
protected virtual void Dispose(bool disposing)
{
    if (disposing && Recorder != null)
        Recorder.Dispose();
}
```

OpenRA.Game/Network/Connection.cs
```diff
@@ -33,6 +33,7 @@ public interface IConnection : IDisposable
 		void SendImmediate(List<byte[]> orders);
 		void SendSync(int frame, byte[] syncData);
 		void Receive(Action<int, byte[]> packetFn);
+		void StartRecording(Func<string> chooseFilename);
```
RoosterDragon: I would remove this from the interface and from `ReplayConnection`.

OpenRA.Game/Game.cs
```diff
@@ -51,7 +51,7 @@ public static OrderManager JoinServer(string host, int port, string password, bo
 		{
 			IConnection connection = new NetworkConnection(host, port);
```
RoosterDragon: If you change this to `var connection = `... then you don't need to rely on defining `StartRecording` on the interface any more.

OpenRA.Game/Network/Connection.cs
whinis: I was trying to stay as close to the old one as possible in that it could be used as any connection type. The old recorder was part of the interface so to allow any interface to record I placed it there

OpenRA.Game/Network/Connection.cs
```diff
@@ -99,10 +100,26 @@ public virtual void Receive(Action<int, byte[]> packetFn)
...
 				packetFn(p.FromClient, p.Data);
+				if (Recorder != null)
+					Recorder.Receive(p.FromClient, p.Data);
+			}
 		}
 
-		protected virtual void Dispose(bool disposing) { }
+		public void StartRecording(Func<string> chooseFilename)
+		{
+			// If we have a previous recording then save/dipose it and start a new one
```
RoosterDragon: `dipose` -> `dispose`. End with a full stop.

RoosterDragon:
:+1: Looks OK to me and recording still works fine.

whinis:
Problem with comment fixed

Mailaender:
Doesn't seem to cause replay metadata parsing regressions: http://resource.openra.net/replays/10/ :white_check_mark:

Mailaender:
I assume this is just a refactor for something upcoming. Won't file a changelog entry.

PR #10309, opened 2015-12-28
Update the default NUnit from 2.6.4 to 3.0.1

PiDelport:
The older version of NUnit seems to have trouble running, at least on Ubuntu 15.10 and recent versions of Mono.

With this update, `make dependencies` followed by `make nunit` works out of the box for me.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -103,6 +103,7 @@ Also thanks to:
     * Paul Dovydaitis (pdovy)
     * Pavlos Touboulidis (pav)
     * Pizzaoverhead
+    * Piët Delport (pjdelport)
     * Psydev
     * Raymond Bedrossian (Squiggles211)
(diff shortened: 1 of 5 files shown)


Mailaender:
The colored console output is indeed nicer. :+1:

obrakmann:
Can you please add the `--noresult` option when running `nunit3-console.exe`? It will prevent the `TestResult.xml` file from being generated, which we don't use anyway and which otherwise just spams the openra source dir.

:+1: aside from that small addition, works for me.

pjdelport:
Sure thing!

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9b34c598fa74f78feede5b1866c2e6f32948097e)

PR #10316, opened 2015-12-28
Swap the Disable- and EnablePower sounds of D2K

abcdefg30:
Although the naming suggests something different, the sounds are more fitting the other way round imho.
(Or my assets are screwed.)

### DIFF
--- a/mods/d2k/audio/notifications.yaml
+++ b/mods/d2k/audio/notifications.yaml
@@ -44,8 +44,8 @@ Sounds:
 	Notifications:
 		RadarUp: MULTI1
 		RadarDown:
-		DisablePower: POWRDN1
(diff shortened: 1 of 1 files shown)


Mailaender:
I can confirm that this is not a regression from https://github.com/OpenRA/OpenRA/pull/9605, but a fuckup from the original Dune 2000 game which was never patched at least not by renaming the .AUD files so this should be safe in every regard. :+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4a8bbe5d5730b62d3f7adfb1d758988bf87de5aa)

PR #10322, opened 2015-12-29
Store triggers in ScriptTriggers in an array.

RoosterDragon:
We can use the enum to index the array directly, in order to give faster lookups compared to a dictionary.

Provides a modest speedup to `ScriptTriggers.TickIdle` on the RA shellmap, reducing CPU time from 0.41% to 0.34%.


obrakmann:
It seems that missions are still sufficiently triggered. :+1:

abcdefg30:
Mission still work for me, too. :+1: / :white_check_mark:

PR #10323, opened 2015-12-29
Ensure some structs implement IEquatable<T>.

RoosterDragon:
I have moved around some of the code to make it consistent with existing overrides in other vectors such as `CPos`, `WVec`, etc.

Implementating `IEquatable<T>` is important for structs as this is used as the default equality comparer (when available) for inbuilt collections that need a definition of equality (like `Dictionary`). Implementing it allows them to avoid boxing costs if such a struct is the key.


abcdefg30:
:+1:

PR #10324, opened 2015-12-29
GPS updates frozen actors that change owner

RoosterDragon:
There are two main changes in this PR:
- A new trait, `FrozenUnderFogUpdatedByGps`, replaces `GpsRemoveFrozenActor`. It supports removing frozen actors that are sold or die but also now supports refreshing the frozen actor when ownership changes. This brings it in line with the GPS dot which changes colour as required.
- Remove the `IRemoveFrozenActor` interface. This caused `FrozenActorLayer.Tick` to be quite slow. The new trait only needs to run when the GPS state changes, rather than every tick. This is faster.

I would test the following:
- When a GPS is launched, that frozen actors are updated as follows:
  - A frozen actor for a sold building is removed.
  - A frozen actor for a destroyed building is removed.
  - A frozen actor for a building that was captured is updated with new colours.
- With an active GPS, check that frozen actors are updated live for the above three scenarios.

---

On the RA shellmap, reduces CPU time spent in `FrozenActorLayer.Tick` from 1.39% to 0.83%.

Suggested changelog: GPS updates frozen actors that change owner


OpenRA.Mods.RA/Traits/GpsRemoveFrozenActor.cs
```diff
@@ -35,21 +34,31 @@ public GpsRemoveFrozenActor(Actor self, GpsRemoveFrozenActorInfo info)
...
+		void RemoveFrozenActor(Actor self)
+		{
 			foreach (var w in watchers)
 			{
-				if (w.Owner != owner && !(info.GrantAllies && w.Owner.IsAlliedWith(owner)))
+				if (!w.Granted || !(info.GrantAllies ? w.Owner.IsAlliedWith(self.Owner) : w.Owner == self.Owner))
 					continue;
 
-				if (w.Launched)
-					return true;
```
RoosterDragon: Attention: I have changed this check from `Launched` to `Granted` which I believe is more correct and matches the description of the trait (an *active* GPS power).

OpenRA.Mods.RA/Traits/GpsRemoveFrozenActor.cs
```diff
@@ -35,21 +34,31 @@ public GpsRemoveFrozenActor(Actor self, GpsRemoveFrozenActorInfo info)
 			watchers = self.World.ActorsWithTrait<GpsWatcher>().Select(w => w.Trait).ToArray();
 		}
 
-		public bool RemoveActor(Actor self, Player owner)
+		public void Tick(Actor self)
```
pchote: You can't guarantee this to be called after an actor is killed.  You should use INotifyRemovedFromWorld or similar instead.

RoosterDragon: I am already implementing `INotifyKilled` which seems to pick up sold and destroyed buildings just fine.

pchote: Should be able to remove this one then.

RoosterDragon: We need this one to erase any frozen actors that are already dead when the GPS first activates.

RoosterDragon: Nope I'm being dumb here. I'll probably need to wire up a way for the GPS to signal when the GPS grant status changes so this trait can re-evaluate stuff.

pchote: `INotifyGPSStateChanged`?

RoosterDragon:
This has grown in complexity somewhat. The trait needing to do things after the actor dies makes it more difficult than usual to notify it about things.

On the plus side, this change means I can support the GPS updating a frozen actor that changes ownership, which is nice.

I have updated the opening post.

Mailaender:
Needs a rebase now.

RoosterDragon:
Rebased.

abcdefg30:
Needs a rebase again. :/

RoosterDragon:
Rebased.

Mailaender:
> A frozen actor for a destroyed building is removed.

Works and I can't spot any related regressions. :white_check_mark:

obrakmann:
Looks good to me, :+1: 

Please rebase and update the engine version in the upgrade rule so we can get this merged.

RoosterDragon:
Rebased and date amended to today.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/ee752baeb16de71a29ed395c617c0f9272795c68)

PR #10329, opened 2015-12-29
power interfaces

ghost:
Asking for help with this.

Trying to implement INotifyPower Interfaces for use with the GPS power. Wondering if I am remotely on the right track.

Currently these only trigger on ATEK construction. I can use `self.Owner.PlayerActor.Trait<PowerManager>();` to decide what power lever the player has but I don't know where I should put this logic (I assume Power.cs)

What is the correct approach to trigger these on power change?


pchote:
This seems like a logical place to use an upgrade, which is triggered by low power.  We are already planning on replacing the current IDisable logic (which low power uses) with upgrades for flexibility.

r34ch:
Closing this to clean up and attempt upgraded instead

PR #10327, opened 2015-12-29
Add "make nunit" step to Travis

PiDelport:
Historically, this wasn't working, but it should be more stable now.

See discussion in #10325.

PR #10333, opened 2015-12-29
Fix-up NUnit 2 & 3 use in Unix Makefile

atlimit8:
Cleans up NUint 2 & 3 after #10309 and #10327.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@
 #
 # to check unit tests (requires NUnit version >= 2.6), run:
 #  make nunit [NUNIT_CONSOLE=<path-to/nunit[2]-console>] [NUNIT_LIBS_PATH=<path-to-libs-dir>] [NUNIT_LIBS=<nunit-libs>]
-#      Use NUNIT_CONSOLE if nunit[2|3]-console was not downloaded by `make dependencies` nor is it not in bin search paths
+#      Use NUNIT_CONSOLE if nunit[3|2]-console was not downloaded by `make dependencies` nor is it in bin search paths
 #      Use NUNIT_LIBS_PATH if NUnit libs are not in search paths. Include trailing /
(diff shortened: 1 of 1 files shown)


obrakmann:
I think it was the point to keep the possibility of still using nunit 2.

Makefile
```diff
@@ -229,18 +229,20 @@ check: utility mods
 	@mono --debug OpenRA.Utility.exe ra --check-code-style OpenRA.Test
 
 NUNIT_CONSOLE := $(shell test -f thirdparty/download/nunit3-console.exe && echo mono thirdparty/download/nunit3-console.exe || \
-	which nunit2-console 2>/dev/null || which nunit3-console 2>/dev/null || which nunit-console 2>/dev/null)
```
atlimit8: Even if NUnit 2.6 support is kept, these two need to be swapped:
```
which nunit3-console 2>/dev/null || which nunit2-console 2>/dev/null || which nunit-console 2>/dev/null)
```

atlimit8:
Also if NUnit 2.6 support is kept, then the changelog should say something like: "Changed the preferred NUnit version to 3.0.1 (from 2.6.4)" instead of "Changed the used NUnit version to 3.0.1 (from 2.6.4)"

atlimit8:
Changed to "Fix-up NUint 2 & 3 use in Unix Makefile" instead.

Phrohdoh:
LGTM :+1:

atlimit8:
Fixed a couple of typos as well.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d15a54ddb88b9119dc4fa16f3ce7ab0dc205b9e6...549a6e1719c582994ae4a79eef6109053142a371)

PR #10336, opened 2015-12-30
Fix parameters sent to DetectionCircleRenderable.

RoosterDragon:
The parameters had been accidentally reversed in the builder methods.

Discovered by Coverity.


atlimit8:
:+1: Matches the constructor almost right above.

PR #10337, opened 2015-12-30
Fix UpgradeChromeMetrics.

RoosterDragon:
This needs to handle the null value passed in for the parent parameter from it's only (non-recursive) callsite.

Discovered by Coverity.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -3410,13 +3410,15 @@ namespace OpenRA.Mods.Common.UtilityCommands
 
 		internal static void UpgradeChromeLayout(int engineVersion, ref List<MiniYamlNode> nodes, MiniYamlNode parent, int depth)
 		{
+			var parentKey = parent != null ? parent.Key.Split('@').First() : null;
+
 			foreach (var node in nodes)
 			{
 				if (engineVersion < 20151102)
 				{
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -3410,13 +3410,15 @@ internal static void UpgradeChromeMetrics(int engineVersion, ref List<MiniYamlNo
...
+			var parentKey = parent != null ? parent.Key.Split('@').First() : null;
+
 			foreach (var node in nodes)
 			{
 				if (engineVersion < 20151102)
 				{
 					if (node.Key == "Color" || node.Key == "ReadyTextAltColor" || node.Key == "TextColor" || node.Key == "TextColorDisabled")
 					{
-						if (parent.Key.StartsWith("MapPreview@"))
+						if (parentKey == "MapPreview")
```
Mailaender: Doesn't it need a null check here or above as well?

pchote: Not as far as I can see. Comparing null against a string is fine.

atlimit8:
:+1: OK

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8674b542df2e0b3bde25a9666e1cd7e03dcaab9e...aabdc5c9b95bca254f3e548ea3a90bca96155be1)

PR #10345, opened 2015-12-30
Suppress warning about Is.StringContaining on NUnit 3

RoosterDragon:
A side effect of moving to NUnit 3, I believe.

### DIFF
--- a/OpenRA.Test/OpenRA.Game/ActorInfoTest.cs
+++ b/OpenRA.Test/OpenRA.Game/ActorInfoTest.cs
@@ -64,11 +64,14 @@ namespace OpenRA.Test
 			}
 			catch (Exception e)
 			{
+				// Is.StringContaining is deprecated in NUnit 3, but we need to support NUnit 2 so we ignore the warning.
+				#pragma warning disable CS0618
 				Assert.That(e.Message, Is.StringContaining("MockA"));
 				Assert.That(e.Message, Is.StringContaining("MockB"));
 				Assert.That(e.Message, Is.StringContaining("MockC"));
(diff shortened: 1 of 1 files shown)


RoosterDragon:
Hmm, is Travis still running 2?

pchote:
Preference is still for 2 if there is a system install.  It will only use nunit 3 if you fetch a local copy.

atlimit8:
Actually, it tries in this order:
 1. NUnit 3 from NuGet
 2. System's `nunit3-console`
 3. System's `nunit2-console`
 4. System's `nunit-console`
--
So, NUnit 2 is still supported.

RoosterDragon:
Well if we need to be able to run on 2, I can't switch to these methods since they don't exist!

Should I just disable the warnings in this section instead?

RoosterDragon:
Changed to suppress the warnings.

OpenRA.Test/OpenRA.Game/ActorInfoTest.cs
```diff
@@ -64,11 +64,14 @@ public void TraitsInConstructOrderB()
 			}
 			catch (Exception e)
 			{
+				// Is.StringContaining is deprecated in NUnit3, but we need to support NUnit2 so we ignore the warning.
```
atlimit8: NUnit 3
NUnit 2

Mailaender:
:+1:

OpenRA.Test/OpenRA.Game/ActorInfoTest.cs
Phrohdoh: What are you suggesting or pointing out here?

PR #10346, opened 2015-12-30
Dispose fixes + misc touchups

RoosterDragon:
- Fix some Dispose implementations, and add some missing calls to Dispose.
- Fix serialization of custom exceptions.
- Rethrow exceptions correctly.

### DIFF
--- a/OpenRA.Mods.Common/LoadScreens/LogoStripeLoadScreen.cs
+++ b/OpenRA.Mods.Common/LoadScreens/LogoStripeLoadScreen.cs
@@ -79,12 +79,12 @@ namespace OpenRA.Mods.Common.LoadScreens
 			r.EndFrame(new NullInputHandler());
 		}
 
-		public override void Dispose()
+		protected override void Dispose(bool disposing)
 		{
-			if (sheet != null)
+			if (disposing && sheet != null)
 				sheet.Dispose();
 
-			base.Dispose();
+			base.Dispose(disposing);
 		}
 	}
 }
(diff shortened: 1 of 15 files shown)


OpenRA.Mods.Common/LoadScreens/LogoStripeLoadScreen.cs
```diff
@@ -79,12 +79,15 @@ public override void Display()
 			r.EndFrame(new NullInputHandler());
 		}
 
-		public override void Dispose()
+		protected override void Dispose(bool disposing)
 		{
-			if (sheet != null)
-				sheet.Dispose();
+			if (disposing)
```
Phrohdoh: Just combine the two ifs?
`if (disposing && sheet != null)`

RoosterDragon: I tend to format it like this so future people can just dunk any extra stuff into the section without redoing the whole block, e.g.:
```
if (disposing)
{
    if (a != null)
        a.Dispose();
    b.Dispose();
}

Phrohdoh: Well there is no `a` vs `b` to worry about here, and this is inconsistent with the rest of our formatting.

RoosterDragon: OK, changed.

Mailaender:
Something isn't right with the voxel rendering here:

image

RoosterDragon:
So `VoxelRenderer` is actually reusing sheets and disposing them too. The extra dispose I added would mean I'd kill off textures which were then re-used next frame and hence missing. It's hacky but I can see why it's doing it that way.

I've simply removed my changes to disposing in `VoxelRenderer` instead.

Mailaender:
:white_check_mark: No further regressions spotted.

pchote:
:+1:

Issue #10298, opened 2015-12-27
Balanced 1v1 maps in the offical pool


r34ch:
In the RA1 mod particularly, we ship many 1v1 maps that I feel are not balanced. Although competitive maps are trickling into the resource centre and are slowly being added to the official pool, we have had a number of competitive tournaments where official maps have been used.

I think we should update some of the more unbalanced maps to match our current drive of unit balancing. At the same time, I think we should also increase the aesthetics of the much older maps.

I am _not_ suggesting that maps should be redone to be perfectly symmetrical or such. I do feel however that oRA's gameplay has evolved over time to provide some unique gameplay mechanics and quirks that now cause issues on some maps.

-Base walking to initial ore patches within CY build radius
-Ore mine distance to REFs, accounting for northern REF disadvantage
-Distances to expansions, attack range distances and tech buildings

I will be creating some PR's to show off some changes I am proposing to remedy this. This is more of an announcement to get peoples thoughts on the idea. 

As always suggestions are welcome. And testing.

- [x] Add various community maps #10320
- [x] Singles #10301
- [x] Man to Man #10312 
- [x] Kotg 2 #10347

klaas1:
I agree, in SoScared tournament we've seen some really nice maps, and I'm glad that two of them made it to the official map pool. But also to me the whole official map pool feels a bit cluttered. I'd say that the official maps should be:

1) Balanced (although this does not mean symetric per se)
2) Aestehic
3) Showcasing all the tilesets
4) Include both small and big maps, and some maps with naval warfare
5) Not include near duplicates, nor have to many maps

r34ch:
I would like to add a couple more 1v1 community maps over time (especially some naval ones) but for now we should have enough competitive official maps in the 1v1 pool for variation in future tourneys.

PR #10347, opened 2015-12-30
Rebalance keep-off-the-grass-2.oramap for the RA mod

ghost:
In relation to #10298

map on resource centre http://resource.openra.net/maps/8264/ (parsed on bleed...)

Original
k1

Updated
k2

-Ore patches in spawns and expansions have been adjusted to remove the northern REF disadvantage.
-Ore mines at corner expansions have been reduced from three to two; terrain made less chokey
-Both ore mines in gem patch have been replaced by a single gem mine
-Trees, debris and detailing


cjshmyr:
:+1:

MicroJOo:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/0b3a94a3a480364f12e63b7d7702644dd04e1d85...628789a8521ff4c040e8b25c098d91f1d8a05ecd)

SoScared:
Just a heads-up, two separate ore patches at the starting position will pull builds orders towards getting your 2nd ref there as opposed to moving early for an expansion.

r34ch:
@SoScared Please feel free to make changes as you see fit, you make a valid point. 

Fwiw my reasoning was such; The large ore patch hinders base placement for those who stayed in spawn. Those who moved out at game start could get good eco with a single large base (3 mines in each corner plus the two in the middle if you didn’t mind redeploying your initial MCV a couple times). It almost made little sense to stay in spawn.

By splitting the ore in spawn it gave people in spawn a little more room for defensive placement of tech structures and by removing 1 ore spawn from each corner, lessened the advantage of taking and entrenching these areas so early.

That said, I didn't think of the changing 2 REF start. And it was only theory crafting.

SoScared:
Here's the upload of mine (see screenshot): http://resource.openra.net/maps/8905/

The only changes was the original patch and I re-positioned one of the three ore mines at the corner expansion to free up more space for more equal ref placements there. The three ore mines at the corner expansion are almost void of ore so it cannot support a build-up of ore harvesters alone. If a player wants to move the MCV from the start he may take that chance but that's gonna require the player successfully take full control of the middle ore as well as the expansion to benefit the move.

I suggest setting the ore patches this way to preserve the map's playstyle.

SoScared:
I'm inspired to go back and decorate the other maps more after seeing your detailing.

SoScared:
Ok so here's your map with some reverses in ore placements: http://resource.openra.net/maps/8990/

keep off the grass 2 rew2

PR #10352, opened 2015-12-31
Speed up GlobalLightingPaletteEffect.

RoosterDragon:
Perform everything in integer arithmetic using some bit twiddling to avoid the performance cost of converting to and from floating point values.

On the TS shellmap, reduces CPU time spent in `GlobalLightingPaletteEffect.AdjustPalette` from 9.66% to 1.16%.

### DIFF
--- a/OpenRA.Mods.Common/Traits/PaletteEffects/GlobalLightingPaletteEffect.cs
+++ b/OpenRA.Mods.Common/Traits/PaletteEffects/GlobalLightingPaletteEffect.cs
@@ -9,7 +9,6 @@
 #endregion
 
 using System.Collections.Generic;
-using System.Drawing;
 using System.Linq;
 using OpenRA.Graphics;
 using OpenRA.Traits;
@@ -54,6 +53,12 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void AdjustPalette(IReadOnlyDictionary<string, MutablePalette> palettes)
 		{
+			// Calculate ambient color multipliers as integers for speed. To handle fractional ambiance, we'll increase
+			// the magnitude of the result by 8 bits.
+			var ar = (uint)((1 << 8) * Ambient * Red);
+			var ag = (uint)((1 << 8) * Ambient * Green);
+			var ab = (uint)((1 << 8) * Ambient * Blue);
+
 			foreach (var kvp in palettes)
 			{
 				if (info.ExcludePalettes.Contains(kvp.Key))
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/PaletteEffects/GlobalLightingPaletteEffect.cs
```diff
@@ -66,11 +71,37 @@ public void AdjustPalette(IReadOnlyDictionary<string, MutablePalette> palettes)
...
+					// If the multiply overflowed we clamp the value, otherwise we mask out the fractional bits.
+					// 3: Finally, we shift the color component back to its correct place. We're already 8 bits higher
+					// than expected due to the multiply, so we don't have to shift as far to get back.
+					var r1 = ((from & 0x00FF0000) >> 16) * ar;
+					var r2 = r1 >= 0x0000FF00 ? 0x0000FF00 : r1 & 0x0000FF00;
+					var r3 = r2 << 8;
+
+					var g1 = ((from & 0x0000FF00) >> 8) * ag;
+					var g2 = g1 >= 0x0000FF00 ? 0x0000FF00 : g1 & 0x0000FF00;
+					var g3 = g2 << 0;
```
atlimit8: Should this line even be here.
It could be commented out like this:
```c#
					// var g3 = g2 << 0; // Already in the right position.
```

RoosterDragon: I find it helpful to leave in to make the pattern clearer, as a personal preference. If you do all the constant folding manually a lot of the structure tends to disappear and it gets more confusing to look at.

atlimit8: OK

OpenRA.Mods.Common/Traits/PaletteEffects/GlobalLightingPaletteEffect.cs
```diff
@@ -54,6 +53,12 @@ public GlobalLightingPaletteEffect(GlobalLightingPaletteEffectInfo info)
 
 		public void AdjustPalette(IReadOnlyDictionary<string, MutablePalette> palettes)
 		{
+			// Calculate ambient color multipliers as integers for speed. To handle fractional ambiance, we'll increase
+			// the magnitude of the result by 8 bits.
+			var ar = (uint)((1 << 8) * Ambient * Red);
+			var ag = (uint)((1 << 8) * Ambient * Green);
+			var ab = (uint)((1 << 8) * Ambient * Blue);
```
atlimit8: `(1 << 8)` should be 255 since 1f <=> 255 for color components.

atlimit8:
:+1:

Mailaender:
Results in the same coloring. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/bab4da04555db1a769d70d2d33d166ad812ba5c7...0b3a94a3a480364f12e63b7d7702644dd04e1d85)

Mailaender:
Thanks! Never thought of this being so wasteful. My original code was inspired by the other palette effect traits so maybe the nuke flash is similarly expensive.

OpenRA.Mods.Common/Traits/PaletteEffects/GlobalLightingPaletteEffect.cs
Phrohdoh: I don't see the need for a `<< 0`.

But my comment is too late. :(

PR #10354, opened 2015-12-31
Remove some old 1v1 maps from the RA mod

ghost:
With the recent addition of three community maps, I felt these maps are played so little and would require such an overhaul that the original feeling of the maps would be lost.

Instead I feel we should remove them and wait for more competitive maps from the community (esp some naval 1v1 maps) and include them instead when the time comes.


MicroJOo:
i see no reason to remove these maps ... they are OpenRA-Classics and some of them are as old as OpenRA itself 
theres still other reason to play maps ... other then having a 1v1 tournament match 
ive watched noobs playing burlesca and really enjoing the shit out of it ... we used to play forest-path on our first 1v1 tournys (nukem made it just for the tourny)

these maps might not be enough for tourny-matches anymore ... but its enough to play against a bot ... or a fun match against friends 

i highly suggest respecting these old contributions and not just remove it ... 

they are still playable !

r34ch:
@MicroJOo I think that's a valid point; not everyone plays the game the same. If the other changes are accepted I think there is enough variety to keep these.

Issue #10248, opened 2015-12-20
Fedora: SDK path could not be resolved


liquidat:
Dear team,
I try to build playtest 2015-12-13 on Fedora 23 but it fails during the make all command with the following error:

    $ make all
    + download_dir=./thirdparty/download
    + mkdir -p ./thirdparty/download
    + cd ./thirdparty/download
    + '[' '!' ']'
    + which nuget
    + '[' '!' -f StyleCopPlus.dll ']'
    + '[' '!' -f StyleCop.dll ']'
    + '[' '!' -f ICSharpCode.SharpZipLib.dll ']'
    + '[' '!' -f MaxMind.GeoIP2.dll ']'
    + '[' '!' -f SharpFont.dll ']'
    + '[' '!' -f nunit.framework.dll ']'
    + '[' '!' -f Mono.Nat.dll ']'
    + '[' '!' -f FuzzyLogicLibrary.dll ']'
    + '[' '!' -f SDL2-CS.dll ']'
    + '[' '!' -f Eluant.dll ']'
    + '[' '!' -f SmarIrc4net.dll ']'
    CSC fixheader.exe
    error CS8001: Warning as Error: SDK path could not be resolved
    Compilation failed: 1 error(s), 0 warnings
    Makefile:251: recipe for target 'fixheader.exe' failed
    make: *** [fixheader.exe] Error 1

I wonder what the given SDK could be. I installed the following mono and sdl packages:

    mono-addins-1.1-2.fc23.x86_64
    mono-core-4.0.5-1.fc23.x86_64
    mono-data-4.0.5-1.fc23.x86_64
    mono-data-sqlite-4.0.5-1.fc23.x86_64
    mono-devel-4.0.5-1.fc23.x86_64
    monodevelop-5.9.7-1.fc23.x86_64
    monodoc-4.0.5-1.fc23.x86_64
    mono-extras-4.0.5-1.fc23.x86_64
    mono-locale-extras-4.0.5-1.fc23.x86_64
    mono-mvc-4.0.5-1.fc23.x86_64
    mono-mvc-devel-4.0.5-1.fc23.x86_64
    mono-wcf-4.0.5-1.fc23.x86_64
    mono-web-4.0.5-1.fc23.x86_64
    mono-winforms-4.0.5-1.fc23.x86_64
    mono-winfx-4.0.5-1.fc23.x86_64
    SDL-1.2.15-20.fc23.x86_64
[...]

PR #10358, opened 2015-12-31
Experience added to Player Actor, Statistics

Phoib:


mods/ra/rules/aircraft.yaml
```diff
@@ -33,6 +33,8 @@ BADR:
 		Interval: 2
 	-EjectOnDeath:
 	RejectsOrders:
+    GivesExperience:
```
reaperrr: Our style convention is to use tabs for indentation.

Phoib: *grumble* wasn't there a lint checker for this in the Makefile?

pchote: I don't think we have a specific lint test for this, but thats a really good idea.

Phoib: A lint test should ignore the LongDesc, as d2k has a few with spaces, not tabs.

But that might be another ticket.

Phrohdoh: This won't be a standard lint test, just more of a file reader.

```csharp
foreach (var line in File.ReadAllLines(file))
    if (line.Trim().Length != 0 && !line.StartsWith("\t"))
        emitError();
```

OpenRA.Mods.Common/Traits/Player/PlayerStatistics.cs
```diff
@@ -118,12 +120,23 @@ public void Killed(Actor self, AttackInfo e)
...
 			{
-				var cost = self.Info.TraitInfo<ValuedInfo>().Cost;
+				cost = self.Info.TraitInfo<ValuedInfo>().Cost;
 				attackerStats.KillsCost += cost;
 				defenderStats.DeathsCost += cost;
 			}
+
+			var experienceInfo = self.Info.TraitInfo<GivesExperienceInfo>();
+			if (experienceInfo != null && experienceInfo.Experience > 0)
+            {
```
pchote: These lines appear to be mixing space and tab indentation.

mods/cnc/rules/player.yaml
```diff
@@ -30,4 +30,4 @@ Player:
 		Prerequisites: techlevel.low, techlevel.medium, techlevel.high, techlevel.superweapons
 	GlobalUpgradeManager:
 	ResourceStorageWarning:
-
```
pchote: Please keep the newlines at the end of the file.

mods/d2k/rules/player.yaml
```diff
@@ -86,4 +86,4 @@ Player:
 	GlobalUpgradeManager:
 	ResourceStorageWarning:
 		AdviceInterval: 26
-
```
pchote: Please keep the newlines at the end of the file.

pchote:
Code looks fine, and appears to behave as advertised ingame.  A few smallish nits:

* Can you please split the style fixes that aren't directly related to the feature into their own commit?
* Please also add this to the observer stats.
* The TS ingame menu overflows:
screen shot 2016-01-01 at 14 34 23
* I suspect that the experience number may overflow the available space in the late game.  You will probably need to truncate the number when it gets too long.

Phoib:
Split off the style fixes in #10366

I cannot test the observer stats or the TS currently, as #10248 prevents me from building the code.

pchote:
My understanding (I haven't been following the discussion closely, and the ticket hasn't been updated with the details) is that #10248 is a bug in the fedora-packaged mono.  @Mailaender has packaged his own version of mono at https://build.opensuse.org/package/show/games:openra/mono which I believe fixes that issue.

pchote:
Discussion (not a specific request to change... yet): it might make more sense to present the player experience as a generic 'Score', and then introduce additional methods of increasing player experience (e.g. harvesting, collecting crates).  We could introduce a divisor into the value so that the numbers don't grow quite so quickly.

Phoib:
For viewing purposes, score would work as well (did anybody figure out how the old RA score was calculated?)

For me, this is a mechanism to introduce upgrades in the form of Generals Points, triggered by Experience. 

I am not sure if harvesting would merit experience points, or score. The player is already rewarded with cash. If a mechanism like Generals Points was introduced, this would double the reward for harvesting.

pchote:
Certainly.  I'm thinking mainly of how we expose this in our default mods, which is orthogonal to using this for generals points (since you can omit the trait that converts resources to exp, and change/remove the UI for the score display).

Phoib:
Rebased on bleed, fixed TS overflow, added to observer stats.

Unit158:
What's the application of this?

Phoib:
Experience collecting for a player, as a base to facilitate player upgrades like the Generals Points from the game Generals.

OpenRA.Game/Traits/Player/PlayerExperience.cs
```diff
@@ -0,0 +1,35 @@
...
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+
+namespace OpenRA.Traits
+{
+	public class PlayerExperienceInfo : ITraitInfo
```
Mailaender: Can you add a `[Desc("")]` here for https://github.com/OpenRA/OpenRA/wiki/Traits documentation?

OpenRA.Mods.Common/Traits/GivesExperience.cs
```diff
@@ -41,14 +41,16 @@ public void Killed(Actor self, AttackInfo e)
 
 			var valued = self.Info.TraitInfoOrDefault<ValuedInfo>();
 
-			// Default experience is 100 times our value
 			var exp = info.Experience >= 0
 				? info.Experience
-				: valued != null ? valued.Cost * 100 : 0;
```
Mailaender: This will need an upgrade rule in OpenRA.Utility

Phoib: Where exactly would I do this?

OpenRA.Game/Traits/Player/PlayerExperience.cs
Phoib: Done

OpenRA.Mods.Common/Traits/GivesExperience.cs
Mailaender: Above https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs#L2879.

Phoib: Started the upgrade rule, but must check if 
a) I am doing this the right way
b) How to replace rules in a list.

Could you look at the rule and shout if I started in the wrong way?

Mailaender: I am also not very good at writing those. Some untested code ahead:

```csharp
if (depth == 2 && parentKey.StartsWith("GivesExperience"))
{
	if (node.Key == "Upgrades")
	{
		foreach (var upgrades in node.Value.Nodes)
		{
			var before = int.Parse(upgrades.Key);
			upgrades.Key = (before / 100).ToString();
		};
	}
}
```

Mailaender: On a second thought this probably wants to use something like `DivideBy(upgrades.Key, 100)`

```csharp
internal static void DivideBy(ref string input, int factor)
{
   var before = Exts.ParseIntegerInvariant(input);
   input = (before / factor).ToString()
}
```

mods/ra/chrome.yaml
```diff
@@ -575,7 +575,7 @@ button-disabled: dialog.png
 	corner-tr: 639,256,1,1
 	corner-bl: 512,383,1,1
 	corner-br: 639,383,1,1  
-    
+
```
abcdefg30: Instead of adding a tab here, could you just empty the line completely, please?
(Same below.)

mods/ts/chrome.yaml
```diff
@@ -498,7 +498,7 @@ newsbutton-highlighted: dialog.png
 	corner-tr: 895,0,1,1
 	corner-bl: 768,127,1,1
 	corner-br: 895,127,1,1
-    
+
```
abcdefg30: Here too.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2876,6 +2876,22 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 					}
 				}
 
+				if (engineVersion < 20160109)
+				{
+					//Divide levels by 100
+					if (depth == 1 && (node.Key == "GainsExperience" || node.Key.StartsWith("GainsExperience@"))
+							&& node.Value.Nodes.FirstOrDefault(n => n.Key == "Upgrades") != null)
+					{
+						foreach (var kv in node.Value.Nodes.Where(n => n.Key == "Upgrades")) {
```
abcdefg30: Please move the brace on a new line:
```
foreach (var kv in node.Value.Nodes.Where(n => n.Key == "Upgrades"))
{

File: OpenRA.Mods.Common/Traits/Player/PlayerStatistics.cs [hunk 10]
```diff
@@ -118,12 +120,23 @@ public void Killed(Actor self, AttackInfo e)
...
 			{
-				var cost = self.Info.TraitInfo<ValuedInfo>().Cost;
+				cost = self.Info.TraitInfo<ValuedInfo>().Cost;
 				attackerStats.KillsCost += cost;
 				defenderStats.DeathsCost += cost;
 			}
+
+			var experienceInfo = self.Info.TraitInfo<GivesExperienceInfo>();
+			if (experienceInfo != null && experienceInfo.Experience > 0)
+			{
```
abcdefg30: You should omit the braces around single-line statements:
```
if (experienceInfo != null && experienceInfo.Experience > 0)
	attackerStats.ExperienceGained += experienceInfo.Experience;
else if (experienceInfo != null && cost > 0)
	attackerStats.ExperienceGained += cost;

File: OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs [hunk 11]
```diff
@@ -2876,6 +2876,22 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+							var key = float.Parse(kv.Key) / 100;							
+							// t is a MiniYamlNode that looks like ("200", "firepower, damage, speed, reload, inaccuracy, rank")
```
abcdefg30: Comments should go before the line they refer to.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2876,6 +2876,22 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+							// The key of this needs to be divided by 100 and saved.
+							var newKv = new MiniYamlNode(key.ToString(), kv.Value);
```
abcdefg30: Doesn't `kv.Key = key.ToString();` work here?

Mailaender: Probably not.

abcdefg30: We do it in other places though, for example [here](https://github.com/OpenRA/OpenRA/blob/7729646f45ea254d7f7e36666893d9b00868120c/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs#L433-L435) or [here](https://github.com/OpenRA/OpenRA/blob/7729646f45ea254d7f7e36666893d9b00868120c/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs#L1577-L1578), or am I missing something again?

mods/ra/chrome.yaml
Phoib: got in there with rebasing, I suspect. Removed

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
Phoib: Comments are for own benefit here, the code doesn't work yet. When I get it to work, the comments will be removed, and the commit rebased into the original one.

OpenRA.Mods.Common/Traits/GivesExperience.cs
Phoib: Thanks. There is no testing mechanism for this, right?

Phoib:
Ok, this should work.

OpenRA.Mods.Common/Traits/GivesExperience.cs
pchote: You can (and should!) test this by removing your manual yaml changes and running the updater over the default mods.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2876,6 +2882,19 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
 
+				if (engineVersion < 20160110)
+				{
+					// Divide levels by 100
+					if (depth == 2 && parentKey.StartsWith("GivesExperience"))
+					{
+						if (node.Key == "Upgrades")
+						{
+        						foreach (var upgrades in node.Value.Nodes)
+ 								DivideBy(ref upgrades.Key, 100);
```
abcdefg30: This needs one level of indentation more.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2876,6 +2882,19 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+    						}
```
abcdefg30: And this here one level less.

mods/ra/chrome.yaml
```diff
@@ -657,7 +657,7 @@ newsbutton-highlighted: dialog.png
 	corner-tr: 895,0,1,1
 	corner-bl: 768,127,1,1
 	corner-br: 895,127,1,1
-    
+
```
abcdefg30: This tab here should be removed, too.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
Mailaender: I also don't know how else to solve it. There is only one way to be sure. Run the upgrade code on the pull request.

abcdefg30:
Needs a rebase now, too.

pchote:
Any progress here @Phoib?  We rely on PR authors working through the review process, otherwise PRs become stale and die.

Phoib:
Rebased it and pushed it.

I never got the upgrade rule to work correctly, I'll have another go next Thursday.

Phoib:
Added a console output to change the experience levels. I never got it to work automatically.

GraionDilach:
@Phoib do you need help with the upgrade rule? I guess I can write a proper one.

Phoib:
Yes please.
I'll do a rebase this weekend.

GraionDilach:
Filed https://github.com/Phoib/OpenRA/pull/1.

OpenRA.Mods.Common/Traits/GivesExperience.cs
GraionDilach: Actually, now that I think of it, having that upgrade rule is a terrible solution here. That change sacrifices the possibility to have levelup requirements like 2.5 times the cost.

A better option would be to change https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/GainsExperience.cs#L61 to `kv.Key * value / 100` without touching the yamls.

reaperrr: For what it's worth, I agree with @GraionDilach.

OpenRA.Mods.Common/Traits/Player/PlayerStatistics.cs
Phrohdoh: No need to check `experienceInfo != null` twice.

obrakmann:
Superseded by #11474.

PR #10344, opened 2015-12-30
Allow only valid file types when saving maps in the editor

obrakmann:
It will also allow saving to files and directories which have a dot in their name, which was the ultimate cause for the crash in #10339.

Fixes #10339.

PR #10362, opened 2016-01-01
Fixed a crash on new map creation

Mailaender:
A regression from https://github.com/OpenRA/OpenRA/pull/10344.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Editor/SaveMapLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Editor/SaveMapLogic.cs
@@ -93,8 +93,14 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 					directoryDropdown.ShowDropDown("LABEL_DROPDOWN_TEMPLATE", 210, mapDirectories.Keys, setupItem);
 			}
 
-			var attr = File.GetAttributes(map.Path);
-			var mapIsUnpacked = attr.HasFlag(FileAttributes.Directory);
+			var mapIsUnpacked = false;
+
+			if (map.Path != null)
+			{
+				var attr = File.GetAttributes(map.Path);
+				mapIsUnpacked = attr.HasFlag(FileAttributes.Directory);
+			}
+
 			var filename = widget.Get<TextFieldWidget>("FILENAME");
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Widgets/Logic/Editor/SaveMapLogic.cs
```diff
@@ -93,10 +93,16 @@ public SaveMapLogic(Widget widget, Action<string> onSave, Action onExit, Map map
...
+
+			if (map.Path != null)
+			{
+				var attr = File.GetAttributes(map.Path);
+				mapIsUnpacked = attr.HasFlag(FileAttributes.Directory);
+			}
+
 			var filename = widget.Get<TextFieldWidget>("FILENAME");
-			filename.Text = mapIsUnpacked ? Path.GetFileName(map.Path) : Path.GetFileNameWithoutExtension(map.Path);
+			filename.Text = mapIsUnpacked && map.Path != null ? Path.GetFileName(map.Path) : Path.GetFileNameWithoutExtension(map.Path);
```
obrakmann: This isn't needed. If `mapIsUnpacked` is true, `map.Path` cannot be null.

obrakmann:
Other than that, :+1:

OpenRA.Mods.Common/Widgets/Logic/Editor/SaveMapLogic.cs
Mailaender: I know, it is an extra safeguard for future changes.

obrakmann: It won't safeguard anything, `map.Path` is used on both sides of the `:`. Also both `Path.GetFilename` and `Path.GetFileNameWithoutExtension` work fine when you pass them a `null`. Please remove it.

Mailaender: Removed.

RoosterDragon:
:+1:

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c77e365820c8408bf5874c37548ccb538e0457b5)

PR #10363, opened 2016-01-01
Added hospital healing to Tiberian Sun

Mailaender:
~~Compare https://github.com/OpenRA/OpenRA/pull/2230.~~

### DIFF
--- a/mods/ts/rules/civilian-structures.yaml
+++ b/mods/ts/rules/civilian-structures.yaml
@@ -719,9 +719,10 @@ CACRSH05:
 		HP: 400
 
 CAHOSP:
-	Inherits: ^CivBuilding
+	Inherits: ^BasicBuilding
 	Tooltip:
 		Name: Civilian Hospital
+		Description: Gives friendly units a medkit to heal themselves.
 	Building:
 		Footprint: xxx xxx xxx xxx
 		Dimensions: 3, 4
@@ -731,6 +732,9 @@ CAHOSP:
 		HP: 800
 	RenderSprites:
 		Palette: player
+	Capturable:
+	CaptureNotification:
+	ProvidesPrerequisite@BuildingName:
 
 CAPYR01:
 	Inherits: ^CivBuilding
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -315,6 +315,23 @@
 	DeathSounds@ZAPPED:
 		Voice: Zapped
 		DeathTypes: EnergyDeath
+	SelfHealing@HOSPITAL:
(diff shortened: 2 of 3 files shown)


pchote:
The original hospital logic (send an infantry inside to heal) was never used in TS ([reference](http://cnc.wikia.com/wiki/Civilian_Hospital)), probably because it was lame.  Because this is a custom addition (so we are not constrained by "we must make it act like the original") it would be better if we made this use the same global selfheal that we use in our RA/TD, and which was used in RA2YR.

GraionDilach:
It was apparently enabled. https://github.com/LipkeGu/TSIniFiles/blob/master/rules.ini#L6555

pchote:
I meant that it was not enabled in a way that affected gameplay - hospitals were not capturable.

GraionDilach:
Ah, good point, I totally forgot Capturable's default value.

Mailaender:
This is definitely not the final balancing. It just makes it a bit more interesting for now. Otherwise all mods share the same gameplay mechanic.

abcdefg30:
:+1: / :white_check_mark:

pchote:
I still think this isn't a great idea.

Mailaender:
Rebased. This doesn't really effect game play that much as most maps don't feature a hospital. I think it is the best solution for now as you can see it work. If you capture a hospital in the other mods, you don't get any visual feedback whatsoever about what is going on. It is way too early to argue here. I don't see us releasing this anytime soon. Especially considering the pace at which small fixups are integrated.

Phrohdoh:
Regarding the visual feedback we have everything in place to add the red-on-white cross symbol commonly associated with hospitals (RA2-style), I don't see any of that as an argument to add this just because we can.

I think we should work on getting TS off the ground (as a vanilla mod) then balance/tweak it.
It seems like every week we have that discussion.

Mailaender:
In fact RA2 did have the TS-style hospital healing as well. The feature and required artwork is YR-exclusive.

pchote:
I guess my main objection here is just that it looks horribly unpolished to have infantry stand on top of the building while they are being healed.  If the units entered the building properly to heal then I could live with it.

Mailaender:
I redid this after finding the appropriate red cross icon. This now works exactly like the TD mod. https://github.com/OpenRA/OpenRA/pull/10680 You can use http://resource.openra.net/maps/8994/ as a test case which features a civilian hospital in the center of the map.

abcdefg30:
:+1: / :white_check_mark:

pchote:
The cross icon looks weird placed at their feet, but that can be tweaked later.  :+1:

PR #7998, opened 2015-04-20
Added the Tiberian Sun docking overlays

Mailaender:
will need some tweaking afterwards.

PR #10364, opened 2016-01-01
Added the Tiberian Sun harvester unload overlay

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/7998. @gangster-ppm also asked for this.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -2876,6 +2876,13 @@ namespace OpenRA.Mods.Common.UtilityCommands
 					}
 				}
 
+				// Rename WithDockingOverlay to WithDockedOverlay
+				if (engineVersion < 20160116)
+				{
+					if (node.Key.StartsWith("WithDockingOverlay"))
+						node.Key = "WithDockedOverlay" + node.Key.Substring(18);
+				}
+
 				UpgradeActorRules(engineVersion, ref node.Value.Nodes, node, depth + 1);
 			}
 		}
--- a/OpenRA.Mods.TS/Activities/VoxelHarvesterDockSequence.cs
+++ b/OpenRA.Mods.TS/Activities/VoxelHarvesterDockSequence.cs
@@ -10,6 +10,7 @@
 
 using OpenRA.Activities;
 using OpenRA.Mods.Common.Activities;
+using OpenRA.Mods.Common.Traits;
 using OpenRA.Mods.TS.Traits;
 
 namespace OpenRA.Mods.TS.Activities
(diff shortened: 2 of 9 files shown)


pchote:
This doesn't sync with the docking sequence code, and instead appears to rely on manual animation timings.  Please implement this following the example of `WithDockingAnimation` instead.

pchote:
A simple testcase to illustrate the incorrect behaviour is to remove `DiscardExcessResources: true` from the TS refinery and then wait for it to run out of storage space.

Mailaender:
Implemented this in a more robust way.

OpenRA.Mods.TS/Activities/VoxelHarvesterDockSequence.cs
```diff
@@ -17,24 +18,52 @@ namespace OpenRA.Mods.TS.Activities
 	public class VoxelHarvesterDockSequence : HarvesterDockSequence
 	{
 		readonly WithVoxelUnloadBody body;
+		readonly WithHarvesterUnloadOverlay spriteOverlay;
```
pchote: Please call this `WithDockingOverlayInfo` to match the existing `WithDockingAnimationInfo`.

Mailaender: That will conflict with the already existing https://github.com/OpenRA/OpenRA/wiki/Traits#withdockingoverlay

pchote: That trait should be renamed to `WithDockedOverlay`. It is named wrong because responds to harvester *docked* events, not *docking* events.

Mailaender: Renamed.

pchote:
Needs a rebase already, sorry.

Mailaender:
Rebased and resorted the upgrade rule location. It's the only thing that keeps conflicting constantly.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2807,6 +2807,13 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 					}
 				}
 
+				// rename WithDockingOverlay to WithDockedOverlay
+				if (engineVersion < 20160101)
+				{
+					if (node.Key.StartsWith("WithDockingOverlay"))
+						node.Key = "WithDockedOverlay";
```
pchote: You can append `node.Key.Substring(18)` to retain any `@suffix` that the original had (such as `smoke` in the rules below).

Mailaender: Good idea. Done.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2807,6 +2807,13 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 					}
 				}
 
+				// rename WithDockingOverlay to WithDockedOverlay
```
pchote: Please capitalize this when updating the date below.

pchote:
Looks fine to me now, but can you please update the upgrade rules to the current date and fix the comment?

Mailaender:
Capitalized, but avoided changing the date as that will just cause rebase problems.

pchote:
> Capitalized, but avoided changing the date as that will just cause rebase problems.

This will break all mods that have upgraded to a version of bleed between 20160101 and now because they will never rule the rule.

Mailaender:
That is true. I forgot about https://github.com/OpenRA/ra2/commit/699fa6ada5223512677c45c3c9723c57a350f1b3. Upgrade rules are pretty toxic for quick pull request resolutions though. Reminds me of the times where we had rebase hell during Changelog conflicts.

Mailaender:
Updated.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f90932636408faf83e540e267937ecb66565ecb8)

I only changelogged the rename since we don't add TS stuff to the changelog yet.

PR #10365, opened 2016-01-01
Misc Missile and CreateEffectWarhead tweaks

### LINKED ISSUE #10350: Missiles no longer trigger splash reliably when hitting water
Observed on RA destroyer and rocket soldier when shooting at empty water. More often than not, an explosion is shown instead.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29495037-missiles-no-longer-trigger-splash-reliably-when-hitting-water?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
- Fixes #10350 by using `airMargin` in CEWH water check
- unhardcodes `Missile` freefall gravity
- Make `CreateEffectWarhead` check for `HitShape.DistanceFromEdge` instead of `.OuterRadius` (to match `SpreadDamageWarhead`)

### DIFF
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -76,6 +76,9 @@ namespace OpenRA.Mods.Common.Effects
 		[Desc("Vertical rate of turn.")]
 		public readonly int VerticalRateOfTurn = 6;
 
+		[Desc("Gravity applied while in free fall.")]
+		public readonly int Gravity = 10;
+
 		[Desc("Run out of fuel after being activated this many ticks. Zero for unlimited fuel.")]
 		public readonly int RangeLimit = 0;
 
@@ -152,8 +155,7 @@ namespace OpenRA.Mods.Common.Effects
 		readonly ProjectileArgs args;
 		readonly Animation anim;
 
-		// NOTE: Might be desirable to unhardcode the number -10
-		readonly WVec gravity = new WVec(0, 0, -10);
+		readonly WVec gravity;
 
 		int ticks;
 
@@ -191,6 +193,7 @@ namespace OpenRA.Mods.Common.Effects
 
 			pos = args.Source;
 			hFacing = args.Facing;
+			gravity = new WVec(0, 0, -info.Gravity);
 			targetPosition = args.PassiveTarget;
 
 			var world = args.SourceActor.World;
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -76,6 +76,9 @@ public class MissileInfo : IProjectileInfo
 		[Desc("Vertical rate of turn.")]
 		public readonly int VerticalRateOfTurn = 6;
 
+		[Desc("Gravity applied while in free fall.")]
```
Phrohdoh: Could you please add that the negative of this value is used internally?
Or better yet make the default `-10` and use whatever value the modder actually put in yaml.

reaperrr: I felt that negative values are unintuitive for most 'end-users'/modders (negative gravity means "falling" towards the sky in my book).

Phrohdoh: Okay, that is a fair point.
I'd rather we come up with a better name (something implying a force acting on the missile) and use the values given to us.

But if you're against that and others are, no problem, just please update the description.

pchote: I agree with @reaperrr.  This claims to be gravity, not an acceleration vector (which would indeed need to be negative).

reaperrr: > Could you please add that the negative of this value is used internally?

Personally, I think modders don't need to know how this works internally. In fact I'm afraid adding that to the description might just cause unnecessary confusion.

GraionDilach: Siding with @reaperrr here. Gravity was falling in TS/RA2 as well. The description is fine.

RoosterDragon:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/048fac368fdb63b65615fd0dae2de772398e5a61...34b61a0a82735428140a6a447dac782569fdb24e)

PR #10366, opened 2016-01-01
Eliminated superfluous linendings

Phoib:
Eliminated linendings from player files


pchote:
Looks good to me, thanks. :+1:

penev92:
Fwliw the utility automatically adds those every time you run an upgrade.

Mailaender:
Yeah, thanks. This only adds even more noise when you run it now.

Phoib:
But why does the tool add it, if it is against the styling convention?

penev92:
Because it needs some love, but nobody has been interested in giving it.

Mailaender:
Fixed in https://github.com/OpenRA/OpenRA/pull/10384.

PR #10368, opened 2016-01-01
Added the isometric legacy map importer

### LINKED ISSUE #9335: Legacy isometric map import
Work in progress :construction: at

* https://github.com/pchote/OpenRA/tree/ts-map-importer
* ~~https://github.com/Mailaender/OpenRA/tree/ts-map-import~~

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26652688-legacy-isometric-map-import?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mailaender:
This is a rebased and slightly improved version of the one published by @pchote. It adds argument validation, fixed the file not found errors, uncommented decorational actors in the translation table and made sure it will also work with Red Alert 2 maps. I am submitting this now, because it is quite hard to maintain outside this repository.

Closes https://github.com/OpenRA/OpenRA/issues/9335.

### DIFF
--- a/OpenRA.Game/FileFormats/Format80.cs
+++ b/OpenRA.Game/FileFormats/Format80.cs
@@ -60,6 +60,70 @@ namespace OpenRA.FileFormats
 			}
 		}
 
+		public static unsafe uint DecodeInto(byte* src, byte* dest) {
+			byte* pdest = dest;
+			byte* readp = src;
+			byte* writep = dest;
+
+			while (true) {
+				byte code = *readp++;
+				byte* copyp;
+				int count;
+				if ((~code & 0x80) != 0) {
+					//bit 7 = 0
+					//command 0 (0cccpppp p): copy
+					count = (code >> 4) + 3;
+					copyp = writep - (((code & 0xf) << 8) + *readp++);
+					while (count-- != 0)
+						*writep++ = *copyp++;
+				}
+				else {
+					//bit 7 = 1
+					count = code & 0x3f;
+					if ((~code & 0x40) != 0) {
+						//bit 6 = 0
+						if (count == 0)
+							//end of image
+							break;
+						//command 1 (10cccccc): copy
+						while (count-- != 0)
+							*writep++ = *readp++;
+					}
+					else {
+						//bit 6 = 1
+						if (count < 0x3e) {
+							//command 2 (11cccccc p p): copy
+							count += 3;
+							copyp = &pdest[*(ushort*) readp];
+
+							readp += 2;
+							while (count-- != 0)
+								*writep++ = *copyp++;
+						}
+						else if (count == 0x3e) {
+							//command 3 (11111110 c c v): fill
+							count = *(ushort*) readp;
+							readp += 2;
+							code = *readp++;
+							while (count-- != 0)
+								*writep++ = code;
+						}
+						else {
+							//command 4 (copy 11111111 c c p p): copy
+							count = *(ushort*) readp;
+							readp += 2;
+							copyp = &pdest[*(ushort*) readp];
+							readp += 2;
+							while (count-- != 0)
+								*writep++ = *copyp++;
+						}
+					}
+				}
+			}
+
+			return (uint) (dest - pdest);
+		}
+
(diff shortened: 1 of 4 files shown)


pchote:
My version of the importer has issues with the map border calculation, and with actor positions being wrong by a couple of cells (possibly due to differences in the corner that TS used as the reference point).  I did not want to merge this until after those issues (especially the second one) had been fixed, because they generate bogus maps.

OpenRA.Game/FileFormats/Format80.cs
```diff
@@ -60,6 +60,70 @@ static void ReplicatePrevious(byte[] dest, int destIndex, int srcIndex, int coun
 			}
 		}
 
+		public static unsafe uint DecodeInto(byte* src, byte* dest) {
```
pchote: This method does not follow our coding conventions.

OpenRA.Mods.TS/UtilityCommands/ImportIsometricLegacyMapCommand.cs
```diff
@@ -0,0 +1,476 @@
...
+using System.IO;
+using System.Linq;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.FileFormats;
+using OpenRA.FileSystem;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.TS.UtilityCommands
+{
+	public class Format5
```
pchote: This class does not follow our coding conventions, and probably doesn't want to live here.

OpenRA.Mods.TS/UtilityCommands/ImportIsometricLegacyMapCommand.cs
```diff
@@ -0,0 +1,476 @@
...
+		}
+
+		void ReadActors(Map map, IniFile file, string type)
+		{
+			var structuresSection = file.GetSection(type, true);
+			foreach (var kv in structuresSection)
+			{
+				var entries = kv.Value.Split(',');
+
+				//var owner = entries[0];
```
pchote: dead code

OpenRA.Mods.TS/UtilityCommands/ImportIsometricLegacyMapCommand.cs
```diff
@@ -0,0 +1,476 @@
...
+					float.TryParse(kv.Value, NumberStyles.Float, NumberFormatInfo.InvariantInfo, out ambient);
+					break;
+				default:
+					Console.WriteLine("Ignoring unknown lighting type: `{0}`".F(kv.Key));
+					break;
+				}
+			}
+
+			var node = new MiniYamlNode("World", new MiniYaml("", new List<MiniYamlNode>()
+				{
```
pchote: This should be unindented by one level.

Mailaender:
Actor position looks fine when I compare houses to streets and bridge huts to bridges:

image

OpenRA.Mods.TS/UtilityCommands/MiniLZO.cs
```diff
@@ -0,0 +1,446 @@
+
+#region Copyright notice
```
pchote: This file does not conform to our project style, and should be rewritten.

pchote:
Basically, this was a proof of concept / prototype that lifted code from other projects without regard to integrating it with our own code base, proper documentation, licenses, or attribution.  It should not be merged until the low level bits have been properly integrated with our own format / parsing code.

Phrohdoh:
@pchote So are we going to consider merging this at all if that is the case?

pchote:
There is a bit of work to address my complaints, but it's by no means impossible.  The only tricky bit is rewriting the MiniLZO implementation, but we are only using one (~150 line) function from it so even that isn't too bad.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/10397/ which includes StyleCop cleanups, map bound fixes and doesn't add redundant global lighting variables.

Issue #9067, opened 2015-08-14
TD: rename "v37" civilian building?


clemty:
According to http://nyerguds.arsaneus-design.com/N64Project/editing/cut_stuff.txt
>Civilian Building "The Studio" [V37]
>Is unused in PC C&C; represents the Westwood Studios building

According to http://www.giantbomb.com/westwood-studios-inc/3010-261/ Westwood Studios was located at 5333 South Arville, Suite 104, Las Vegas (which may be a PO box?), but google maps/earth shows a similar building there: https://www.google.com/maps/place/36%C2%B005'34.9%22N+115%C2%B011'59.2%22W/@36.093025,-115.199777,228m/data=!3m1!1e3!4m2!3m1!1s0x0:0x0

edit:
currently it is called:
- Civilian Building

but maybe this would be more accurate:

- Westwood Studios Building

Mailaender:
We already support generic tooltip names https://github.com/OpenRA/OpenRA/pull/6613 so we can have both:

```yaml
Tooltip:
		Name: Westwood Studios
		GenericName: Civilian Building
```

In fact all those civilian buildings had names attached: http://nyerguds.arsaneus-design.com/cnc95upd/inirules/structs.ini

PR #10372, opened 2016-01-01
Added tooltips for civilian buildings to Tiberian Dawn

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/9067.

### DIFF
--- a/mods/cnc/rules/defaults.yaml
+++ b/mods/cnc/rules/defaults.yaml
@@ -508,8 +508,9 @@
 	Health:
 		HP: 400
 	Tooltip:
-		Name: Civilian Building
-		GenericVisibility: None
+		GenericName: Civilian Building
+		GenericStancePrefix: false
+		ShowOwnerRow: false
 	FrozenUnderFog:
 		StartsRevealed: true
 
@@ -521,8 +522,9 @@
 		Footprint: x
 	WithSpriteBody:
 	Tooltip:
-		Name: Civilian Building (Destroyed)
-		GenericVisibility: None
+		GenericName: Civilian Building (Destroyed)
+		GenericStancePrefix: false
+		ShowOwnerRow: false
 	FrozenUnderFog:
 		StartsRevealed: true
 	ScriptTriggers:
@@ -543,7 +545,7 @@
 	Inherits: ^CivBuilding
 	-Selectable:
 	Tooltip:
-		Name: Field
+		GenericName: Field
 	-WithBuildingExplosion:
 	-Targetable:
 	-Demolishable:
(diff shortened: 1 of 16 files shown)


pchote:
Looks good, but you'll need to set up the generic tooltips for in the missions (they should stay as Civilian Building there).

atlimit8:
Needs rebase.

Mailaender:
Rebased. Not sure if the generic tooltip stances are working at all or I just don't understand if they work. Also not sure why they should be generic in single-player.

mods/cnc/rules/defaults.yaml
```diff
@@ -491,7 +491,7 @@
 	Health:
 		HP: 400
 	Tooltip:
-		Name: Civilian Building
+		GenericName: Civilian Building
 		GenericVisibility: None
```
pchote: These overrides explicitly disable the generic visibility everywhere...

Mailaender: Fixed, but I think I am suffering from yaml merge problems elsewhere. Also tooltips a la "Enemy Civilian Building" isn't really worth it.

pchote: We explicitly don't want the "Enemy" / "Allied" for these.  You can turn them off by setting `GenericStancePrefix: false`.

Mailaender: Okay, also set ShowOwnerRow to false as well.

pchote:
Still doesn't work in the campaigns:
screen shot 2016-01-03 at 18 17 12

You need to enable the generic visibility in the maps:
```
	^Building:
		Tooltip:
			GenericVisibility: Enemy, Neutral
			ShowOwnerRow: false
```

Mailaender:
Fixed. Note: Don't test with the shroud disable cheats.

RoosterDragon:
In nod01, all the buildings start out as "Civilian Building" but after being destroyed take on one of the new terms like "Pablo's Wicked Pub (Destroyed)". I assume this is not intentional.

Mailaender:
Should be fixed now. It is utter non-sense to me that those should not be displayed in missions and that this exception has to be configured per mission map.yaml.

RoosterDragon:
This still doesn't seem to have fixed the issue.

Mailaender:
Works fine here:

image

You may have tested this with developer cheats on.

Mailaender:
Oh you meant civilian building husks. Fixed that.

RoosterDragon:
:+1:

obrakmann:
You'll need to touch the following maps as well:

* gdi03
* nod03a
* nod03b
* nod05
* nod06a
* nod06b
* nod06c

Other than that, :+1:

Mailaender:
Updated for more missions.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/9b225288af95e3082a83d876487e809c74b482f3...62a21f349f42f08bd7c64cdd812270e3786e931c)

PR #10373, opened 2016-01-01
Return a Target.FrozenActor in Actor.ResolveFrozenActorOrder

### LINKED ISSUE #10361: Can't shoot at units frozen under fog
image

Found out about it while testing https://github.com/OpenRA/OpenRA/pull/9452, but it is already in `bleed` so it didn't regress there, but before.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29516475-can-t-shoot-at-units-frozen-under-fog?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
...instead of a Target.Actor, which would make code dealing with the order work with false assumptions.

Fixes #10361.

Here's a list of things that need fixing before this can be merged:
- [ ] `ExternalCaptures` + `ExternalCaptureActivity`.
- [ ] `MoveAdjacentTo`.
- [ ] `Attack*` (non-turreted).
- [ ] `Enter` and friends.

...

### DIFF
--- a/OpenRA.Mods.Common/ActorExts.cs
+++ b/OpenRA.Mods.Common/ActorExts.cs
@@ -81,7 +81,7 @@ namespace OpenRA.Mods.Common
 
 			// Target is still alive - resolve the real order
 			if (frozen.Actor != null && frozen.Actor.IsInWorld)
-				return Target.FromActor(frozen.Actor);
+				return Target.FromFrozenActor(frozen);
 
 			if (!order.Queued)
 				self.CancelActivity();
--- a/OpenRA.Mods.Common/ActorExts.cs
+++ b/OpenRA.Mods.Common/ActorExts.cs
@@ -80,21 +80,8 @@ namespace OpenRA.Mods.Common
 			self.SetTargetLine(frozen, targetLine, true);
 
 			// Target is still alive - resolve the real order
-			if (frozen.Actor != null && frozen.Actor.IsInWorld)
+			if (frozen.Actor != null)
 				return Target.FromFrozenActor(frozen);
-
-			if (!order.Queued)
-				self.CancelActivity();
-
-			var move = self.TraitOrDefault<IMove>();
-			if (move != null)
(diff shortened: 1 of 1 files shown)


pchote:
This workaround was there so that we didn't need to make every ResolveOrder trait aware of frozen actors.  I believe that this is the right change to make, but we're now going to need to go through all of our traits and check that they don't break in their own special ways when they receive an order against a frozen actor.

OpenRA.Mods.Common/ActorExts.cs
```diff
@@ -81,7 +81,7 @@ public static Target ResolveFrozenActorOrder(this Actor self, Order order, Color
 
 			// Target is still alive - resolve the real order
 			if (frozen.Actor != null && frozen.Actor.IsInWorld)
-				return Target.FromActor(frozen.Actor);
+				return Target.FromFrozenActor(frozen);
```
pchote: We probably want to do this if the target is dead, too.  We can then drop the workaround below, which causes the "artillery drive up to a dead frozen actor instead of shooting them" bug.

obrakmann: Yup, added. Do you know if we have a ticket for that? I couldn't find anything in a quick search.

pchote:
Adding has dependencies tag so this isn't merged before we check for downstream order resolving breakage.

obrakmann:
Ok, fixing all the ingrained assumptions around this workaround proves to be infeasible. I vote for reverting #10007 instead. Closing here.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/10420.

PR #10376, opened 2016-01-02
Speed up Aircraft.Repulse.

RoosterDragon:
Avoid LINQ in the repulsion force calculation, and cache a trait lookup used when determining movement speed.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
+++ b/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
@@ -97,6 +97,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		UpgradeManager um;
 		IDisposable reservation;
+		IEnumerable<int> speedModifiers;
 
 		[Sync] public int Facing { get; set; }
 		[Sync] public WPos CenterPosition { get; private set; }
@@ -124,7 +125,11 @@ namespace OpenRA.Mods.Common.Traits
 			IsPlane = !info.CanHover;
 		}
 
-		public void Created(Actor self) { um = self.TraitOrDefault<UpgradeManager>(); }
+		public void Created(Actor self)
+		{
+			um = self.TraitOrDefault<UpgradeManager>();
+			speedModifiers = self.TraitsImplementing<ISpeedModifier>().ToArray().Select(sm => sm.GetSpeedModifier());
+		}
 
 		public void AddedToWorld(Actor self)
 		{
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -124,7 +125,11 @@ public Aircraft(ActorInitializer init, AircraftInfo info)
 			IsPlane = !info.CanHover;
 		}
 
-		public void Created(Actor self) { um = self.TraitOrDefault<UpgradeManager>(); }
+		public void Created(Actor self)
+		{
+			um = self.TraitOrDefault<UpgradeManager>();
+			speedModifiers = self.TraitsImplementing<ISpeedModifier>().ToArray().Select(sm => sm.GetSpeedModifier());
```
Phrohdoh: Won't this make it impossible for the speed modifiers to change over the course of this actor's life?

RoosterDragon: Remember it's an enumerable, the `Select` will be re-evaluated every time.

Phrohdoh: Oh right, carry on.

RoosterDragon: The enumerable is passed directly into `ApplyPercentageModifiers` which does the heavy lifting.

So all we're doing is passing a field directly to a method - it doesn't come cheaper than that.

pchote:
Looks good to me :+1:

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
penev92: I'm curious - is the .ToList() in the middle there really helping?

atlimit8: Yes, the `IEnumerable<int>` created by `.Select(sm => sm.GetSpeedModifier())` holds a reference to the unnamed `ISpeedModifier[]` created by `.ToArray()`.

penev92: Neat, thanks.

atlimit8:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/453a36bfa96816ac224f45a6b3fae2fa3ec1eaae...81efcab0d7a701c6fe52c53fc7f87a4c0aeef542)

PR #10377, opened 2016-01-02
Added TS map Tiberium Garden Redux

Mailaender:
Another test for https://github.com/OpenRA/OpenRA/pull/10368. This is a map that doesn't highlight the deficiencies of our TS mod (missing bridges) and should be already playable.

Source: http://ra2maps.zzattack.org/Tiberian%20Sun/index.php?dir=Official%20Multiplayer%2FTiberium%20Garden%20Redux

Description inspired by http://cnc.wikia.com/wiki/Tiberium_Garden_Redux.


pchote:
The bottom border of the map is truncated, and needs to be extended by reducing the south cordon.

Mailaender:
Corrected the bounds. Verified against FinalSun. It also makes sense as the bottom left decorational old war factory and bottom right Tiberium field is now included.

pchote:
:+1:

abcdefg30:

PR #10383, opened 2016-01-02
Made attackable positions customizable

### LINKED ISSUE #7717: Make actor height (and Z height for targeting) customisable
Should affect blockable projectiles and possibly flying actors.
Needed for TS (and my MechCommander RTS mod).
- [x] customisable height for blocking actors
- [ ] custom target point ZOffset for projectiles
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9892042-make-actor-height-and-z-height-for-targeting-customisable?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
This allows modders to define custom offsets relative to `Actor.CenterPosition` that will be targeted instead of `CenterPosition`, which is extremely useful for tall units like mechs.

~~Also adds custom HitShapes to all TS units (not buildings for now, no point in that until we have footprint/custom shape support).~~ split to #11169.

Closes #7717.

### DIFF
--- a/OpenRA.Game/Traits/Target.cs
+++ b/OpenRA.Game/Traits/Target.cs
@@ -165,6 +165,35 @@ namespace OpenRA.Traits
 			}
 		}
 
+		public IEnumerable<WPos> AttackablePositions
+		{
+			get
+			{
+				switch (Type)
+				{
+					case TargetType.Actor:
+						if (!actor.Targetables.Any(Exts.IsTraitEnabled))
+							return new[] { actor.CenterPosition };
+
+						var attackablePositions = actor.TraitsImplementing<IAttackablePositions>();
+						if (attackablePositions.Any())
+						{
+							var target = this;
+							return attackablePositions.SelectMany(ap => ap.AttackablePositions(target.actor));
+						}
+
+						return new[] { actor.CenterPosition };
+					case TargetType.FrozenActor:
+						return new[] { frozen.CenterPosition };
+					case TargetType.Terrain:
+						return new[] { pos };
+					default:
+					case TargetType.Invalid:
+						return NoPositions;
+				}
+			}
+		}
+
 		public bool IsInRange(WPos origin, WDist range)
 		{
 			if (Type == TargetType.Invalid)
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -780,7 +780,7 @@ namespace OpenRA.Mods.Common.Effects
 			// Check if target position should be updated (actor visible & locked on)
 			var newTarPos = targetPosition;
 			if (args.GuidedTarget.IsValidFor(args.SourceActor) && lockOn)
-				newTarPos = args.GuidedTarget.CenterPosition
+				newTarPos = args.GuidedTarget.AttackablePositions.PositionClosestTo(args.Source)
 					+ new WVec(WDist.Zero, WDist.Zero, info.AirburstAltitude);
 
 			// Compute target's predicted velocity vector (assuming uniform circular motion)
--- a/OpenRA.Mods.Common/HitShapes/Capsule.cs
+++ b/OpenRA.Mods.Common/HitShapes/Capsule.cs
@@ -85,6 +85,12 @@ namespace OpenRA.Mods.Common.HitShapes
 		public WDist DistanceFromEdge(WPos pos, Actor actor)
 		{
 			var actorPos = actor.CenterPosition;
+			var attackablePositions = actor.TraitsImplementing<IAttackablePositions>();
+			if (attackablePositions.Any())
+			{
+				var positions = attackablePositions.SelectMany(ap => ap.AttackablePositions(actor));
+				actorPos = positions.PositionClosestTo(actorPos);
+			}
 
 			if (pos.Z > actorPos.Z + VerticalTopOffset)
 				return DistanceFromEdge((pos - (actorPos + new WVec(0, 0, VerticalTopOffset))).Rotate(-actor.Orientation));
@@ -98,6 +104,12 @@ namespace OpenRA.Mods.Common.HitShapes
 		public void DrawCombatOverlay(WorldRenderer wr, RgbaColorRenderer wcr, Actor actor)
 		{
 			var actorPos = actor.CenterPosition;
+			var attackablePositions = actor.TraitsImplementing<IAttackablePositions>();
+			if (attackablePositions.Any())
+			{
+				var positions = attackablePositions.SelectMany(ap => ap.AttackablePositions(actor));
+				actorPos = positions.PositionClosestTo(actorPos);
+			}
 
 			var a = actorPos + new WVec(PointA.X, PointA.Y, VerticalTopOffset).Rotate(actor.Orientation);
 			var b = actorPos + new WVec(PointB.X, PointB.Y, VerticalTopOffset).Rotate(actor.Orientation);
--- a/OpenRA.Mods.Common/HitShapes/Circle.cs
+++ b/OpenRA.Mods.Common/HitShapes/Circle.cs
@@ -49,6 +49,12 @@ namespace OpenRA.Mods.Common.HitShapes
 		public WDist DistanceFromEdge(WPos pos, Actor actor)
 		{
 			var actorPos = actor.CenterPosition;
+			var attackablePositions = actor.TraitsImplementing<IAttackablePositions>();
+			if (attackablePositions.Any())
+			{
+				var positions = attackablePositions.SelectMany(ap => ap.AttackablePositions(actor));
+				actorPos = positions.PositionClosestTo(actorPos);
+			}
 
 			if (pos.Z > actorPos.Z + VerticalTopOffset)
 				return DistanceFromEdge(pos - (actorPos + new WVec(0, 0, VerticalTopOffset)));
@@ -62,6 +68,12 @@ namespace OpenRA.Mods.Common.HitShapes
 		public void DrawCombatOverlay(WorldRenderer wr, RgbaColorRenderer wcr, Actor actor)
 		{
 			var actorPos = actor.CenterPosition;
+			var attackablePositions = actor.TraitsImplementing<IAttackablePositions>();
+			if (attackablePositions.Any())
+			{
+				var positions = attackablePositions.SelectMany(ap => ap.AttackablePositions(actor));
+				actorPos = positions.PositionClosestTo(actorPos);
+			}
 
 			RangeCircleRenderable.DrawRangeCircle(
 				wr, actorPos + new WVec(0, 0, VerticalTopOffset), Radius, 1, Color.Yellow, 0, Color.Yellow);
--- a/OpenRA.Mods.Common/HitShapes/Rectangle.cs
+++ b/OpenRA.Mods.Common/HitShapes/Rectangle.cs
@@ -89,6 +89,12 @@ namespace OpenRA.Mods.Common.HitShapes
 		public WDist DistanceFromEdge(WPos pos, Actor actor)
 		{
 			var actorPos = actor.CenterPosition;
+			var attackablePositions = actor.TraitsImplementing<IAttackablePositions>();
+			if (attackablePositions.Any())
+			{
+				var positions = attackablePositions.SelectMany(ap => ap.AttackablePositions(actor));
+				actorPos = positions.PositionClosestTo(actorPos);
+			}
 
 			if (pos.Z > actorPos.Z + VerticalTopOffset)
 				return DistanceFromEdge((pos - (actorPos + new WVec(0, 0, VerticalTopOffset))).Rotate(-actor.Orientation));
@@ -102,6 +108,12 @@ namespace OpenRA.Mods.Common.HitShapes
 		public void DrawCombatOverlay(WorldRenderer wr, RgbaColorRenderer wcr, Actor actor)
 		{
 			var actorPos = actor.CenterPosition;
+			var attackablePositions = actor.TraitsImplementing<IAttackablePositions>();
+			if (attackablePositions.Any())
+			{
+				var positions = attackablePositions.SelectMany(ap => ap.AttackablePositions(actor));
+				actorPos = positions.PositionClosestTo(actorPos);
+			}
 
 			var vertsTop = combatOverlayVertsTop.Select(v => wr.ScreenPosition(actorPos + v.Rotate(actor.Orientation)));
 			var vertsBottom = combatOverlayVertsBottom.Select(v => wr.ScreenPosition(actorPos + v.Rotate(actor.Orientation)));
--- a/OpenRA.Mods.Common/Traits/Armament.cs
+++ b/OpenRA.Mods.Common/Traits/Armament.cs
@@ -219,7 +219,7 @@ namespace OpenRA.Mods.Common.Traits
 				Source = muzzlePosition(),
 				CurrentSource = muzzlePosition,
 				SourceActor = self,
-				PassiveTarget = target.CenterPosition,
+				PassiveTarget = target.AttackablePositions.First(),
 				GuidedTarget = target
 			};
 
--- a/OpenRA.Mods.Common/Traits/Armament.cs
+++ b/OpenRA.Mods.Common/Traits/Armament.cs
@@ -33,6 +33,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Has to be defined in weapons.yaml as well.")]
 		public readonly string Weapon = null;
 
+		[Desc("Ignore any custom attackable positions and always aim at the targets' CenterPosition.")]
+		public readonly bool IgnoreAttackablePositionOffset = false;
+
 		[Desc("Which limited ammo pool (if present) should this armament be assigned to.")]
 		public readonly string AmmoPoolName = "primary";
 
@@ -219,7 +222,7 @@ namespace OpenRA.Mods.Common.Traits
 				Source = muzzlePosition(),
 				CurrentSource = muzzlePosition,
 				SourceActor = self,
-				PassiveTarget = target.AttackablePositions.First(),
+				PassiveTarget = Info.IgnoreAttackablePositionOffset ? target.CenterPosition : target.AttackablePositions.PositionClosestTo(muzzlePosition()),
 				GuidedTarget = target
 			};
 
--- a/OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
@@ -58,6 +58,25 @@ namespace OpenRA.Mods.Common.Traits
 			if (healthInfo != null)
 				healthInfo.Shape.DrawCombatOverlay(wr, wcr, self);
 
+			var attackablePositions = self.TraitsImplementing<IAttackablePositions>();
+			foreach (var ap in attackablePositions)
+			{
+				var attackPos = self.CenterPosition;
+				var positions = ap.AttackablePositions(self);
+				if (positions.Any())
+					attackPos = positions.First();
+
+				var attackPosColor = Color.Magenta;
+				var apx = new WVec(96, 0, 0);
+				var apy = new WVec(0, 96, 0);
+				var apa = wr.ScreenPosition(attackPos + apx);
+				var apb = wr.ScreenPosition(attackPos - apx);
+				var apc = wr.ScreenPosition(attackPos + apy);
+				var apd = wr.ScreenPosition(attackPos - apy);
+				wcr.DrawLine(apa, apb, iz, attackPosColor);
+				wcr.DrawLine(apc, apd, iz, attackPosColor);
+			}
+
 			var blockers = allBlockers.Where(Exts.IsTraitEnabled).ToList();
 			if (blockers.Count > 0)
 			{
--- a/OpenRA.Mods.Common/Traits/Health.cs
+++ b/OpenRA.Mods.Common/Traits/Health.cs
@@ -9,6 +9,7 @@
  */
 #endregion
 
+using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Mods.Common.HitShapes;
 using OpenRA.Traits;
@@ -19,8 +20,12 @@ namespace OpenRA.Mods.Common.Traits
 	{
 		[Desc("HitPoints")]
 		public readonly int HP = 0;
+
 		[Desc("Trigger interfaces such as AnnounceOnKill?")]
 		public readonly bool NotifyAppliedDamage = true;
+
+		[Desc("Modifies offset of targeting position for targeting by weapons.")]
+		public readonly WVec AttackablePositionOffset = WVec.Zero;
 
 		[FieldLoader.LoadUsing("LoadShape")]
 		public readonly IHitShape Shape;
@@ -55,9 +60,10 @@ namespace OpenRA.Mods.Common.Traits
 		public virtual object Create(ActorInitializer init) { return new Health(init, this); }
 	}
 
-	public class Health : IHealth, ISync, ITick
+	public class Health : IHealth, ISync, ITick, IAttackablePositions, INotifyCreated
 	{
 		public readonly HealthInfo Info;
+		BodyOrientation body;
 
 		[Sync] int hp;
 
@@ -73,11 +79,26 @@ namespace OpenRA.Mods.Common.Traits
 			DisplayHP = hp;
 		}
 
+		public void Created(Actor self)
+		{
+			body = self.TraitOrDefault<BodyOrientation>();
+		}
+
 		public int HP { get { return hp; } }
 		public int MaxHP { get; private set; }
 
 		public bool IsDead { get { return hp <= 0; } }
 		public bool RemoveOnDeath = true;
+
+		public IEnumerable<WPos> AttackablePositions(Actor self)
+		{
+			var offset = body != null ? Info.AttackablePositionOffset.Rotate(body.QuantizeOrientation(self, self.Orientation))
+				: Info.AttackablePositionOffset;
+
+			var pos = self.CenterPosition + (body != null ? body.LocalToWorld(offset) : offset);
+
+			return new WPos[] { pos };
+		}
 
 		public DamageState DamageState
 		{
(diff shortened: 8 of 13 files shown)


OpenRA.Mods.Common/Traits/Armament.cs
```diff
@@ -65,6 +65,9 @@ public class ArmamentInfo : UpgradableTraitInfo, IRulesetLoaded, Requires<Attack
 		public WeaponInfo WeaponInfo { get; private set; }
 		public WDist ModifiedRange { get; private set; }
 
+		[Desc("Should the targets' custom TargetingPointZOffset be ignored.")]
```
Phrohdoh: `target's` is proper here.

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
```diff
@@ -51,8 +51,20 @@ public void RenderAfterWorld(WorldRenderer wr, Actor self)
 			var iz = 1 / wr.Viewport.Zoom;
 
 			if (healthInfo != null)
+			{
 				healthInfo.Shape.DrawCombatOverlay(wr, wcr, self);
 
+				var targetZc = Color.Magenta;
+				var tx = new WVec(96, 0, 0);
```
Phrohdoh: What is this 96 value and why 96?
Can this logically be customizable?

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -19,9 +19,13 @@ public class HealthInfo : ITraitInfo, UsesInit<HealthInit>
 	{
 		[Desc("HitPoints")]
 		public readonly int HP = 0;
+
 		[Desc("Trigger interfaces such as AnnounceOnKill?")]
 		public readonly bool NotifyAppliedDamage = true;
 
+		[Desc("Distance above CenterPosition for targeting by weapons.")]
```
Phrohdoh: I don't think this makes much sense (doesn't "scale").
Consider a east-to-west facing ship.
Would this make more sense as a worldly vector offset? (consider turret offset)

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
reaperrr: > What is this 96 value and why 96?

Offset relative to TargetingCenterPosition, used to define the start and end points of the lines for drawing the cross overlay.

> Can this logically be customizable?

Yes. But nearly all other debug overlays we have aren't customizable, either.

OpenRA.Mods.Common/Traits/Health.cs
reaperrr: The value is passed to a WVec's Z value later, so technically it already is.

reaperrr: I was considering to make this directly WVec, but felt it would complicate matters (and increase risk of bugs in damage calculation etc.) if X and Y were customizable as well.

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
Phrohdoh: Okay, these look to be width and height values.

>Yes. But nearly all other debug overlays we have aren't customizable, either.

I find that to be a poor reason not to allow someone to customize something, but ok for consistency.

OpenRA.Mods.Common/Traits/Health.cs
Phrohdoh: I tend to agree that it may, but it also allows this feature to be used with actors whose graphics aren't perfectly vertical.

Imagine a Titan or similar with a weak spot that is offset from the centerpoint's Z axis.

GraionDilach: Doesn't worth the hassle. If directional armors get in, a combo of that would work better.

Phrohdoh: Directional armor doesn't at all relate to the projectile's initial target position.
The "hassle" is literally changing this to a `WVec` then changing the referencing code to use this directly instead of creating a `WVec` from components.

But ok, this can be done later when someone requests it.

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
reaperrr: I don't mean that I'm generally against making it customizable, I just didn't bother while working on this PR.

OpenRA.Mods.Common/Traits/Health.cs
reaperrr: On the other hand, `Capsule` and `Rectangle` shapes can already be positioned along the Y axis, so they're not necessarily centered around the `Actor.CenterPosition`. So admittedly, allowing to fully customize the offset and not just Z would be more thorough and consistent.

I guess I'll give it a try.

reaperrr:
Updated. Fixed description and offset is now a full WVec, allowing custom X/Y position as well (updated TS Titan as showcase).

OpenRA.Game/Actor.cs
```diff
@@ -52,6 +52,18 @@ public sealed class Actor : IScriptBindable, IScriptNotifyBind, ILuaTableBinding
...
 
+		public WPos TargetingCenterPosition
+		{
+			get
+			{
+				var offset = WVec.Zero;
+				if (health != null)
+					offset = health.Offset;
+
+				return CenterPosition + offset;
```
Phrohdoh: `return CenterPosition +  health == null ? WVec.Zero : health.Offset;`

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -757,8 +757,10 @@ public void Tick(World world)
 			// Check if target position should be updated (actor visible & locked on)
 			var newTarPos = targetPosition;
 			if (args.GuidedTarget.IsValidFor(args.SourceActor) && lockOn)
-				newTarPos = args.GuidedTarget.CenterPosition
-					+ new WVec(WDist.Zero, WDist.Zero, info.AirburstAltitude);
+			{
+				var airburstVec = new WVec(WDist.Zero, WDist.Zero, info.AirburstAltitude);
+				newTarPos = (!args.IgnoreTargetPointOffset ? args.GuidedTarget.TargetingCenterPosition : args.GuidedTarget.CenterPosition) + airburstVec;
```
Phrohdoh: You can simplify that ternary.
`args.IgnoreTargetPointOffset ? args.GuidedTarget.CenterPosition : args.GuidedTarget.TargetingCenterPosition`

OpenRA.Mods.Common/HitShapes/Capsule.cs
```diff
@@ -27,6 +27,12 @@ public class CapsuleShape : IHitShape
 
 		public readonly WDist Radius = new WDist(426);
 
+		[Desc("Defines the top offset relative to the actors' target point.")]
+		public readonly int VerticalTopOffset = 0;
+
+		[Desc("Defines the bottom offset relative to the actors' target point.")]
```
Phrohdoh: `actor's` for both descriptions.

OpenRA.Mods.Common/HitShapes/Circle.cs
```diff
@@ -22,11 +22,21 @@ public class CircleShape : IHitShape
 		[FieldLoader.Require]
 		public readonly WDist Radius = new WDist(426);
 
+		[Desc("Defines the top offset relative to the actors' target point.")]
```
Phrohdoh: `actor's` for both descriptions.

OpenRA.Mods.Common/HitShapes/Rectangle.cs
```diff
@@ -25,10 +25,17 @@ public class RectangleShape : IHitShape
 		[FieldLoader.Require]
 		public readonly int2 BottomRight;
 
+		[Desc("Defines the top offset relative to the actors' target point.")]
```
Phrohdoh: `actor's` again

OpenRA.Mods.Common/HitShapes/Circle.cs
```diff
@@ -35,12 +45,24 @@ public WDist DistanceFromEdge(WVec v)
 
 		public WDist DistanceFromEdge(WPos pos, Actor actor)
 		{
-			return DistanceFromEdge(pos - actor.CenterPosition);
+			var actorTargPos = actor.TargetingCenterPosition;
+
+			if (pos.Z <= actorTargPos.Z + VerticalTopOffset && pos.Z >= actorTargPos.Z + VerticalBottomOffset)
+				return DistanceFromEdge(pos - new WPos(actorTargPos.X, actorTargPos.Y, pos.Z));
```
Phrohdoh: Newlines please.

OpenRA.Mods.Common/HitShapes/Rectangle.cs
```diff
@@ -68,13 +86,24 @@ public WDist DistanceFromEdge(WVec v)
 
 		public WDist DistanceFromEdge(WPos pos, Actor actor)
 		{
-			return DistanceFromEdge((pos - actor.CenterPosition).Rotate(-actor.Orientation));
+			var actorTargPos = actor.TargetingCenterPosition;
+
+			if (pos.Z <= actorTargPos.Z + VerticalTopOffset && pos.Z >= actorTargPos.Z + VerticalBottomOffset)
+				return DistanceFromEdge((pos - new WPos(actorTargPos.X, actorTargPos.Y, pos.Z)).Rotate(-actor.Orientation));
```
Phrohdoh: Newlines, please.

OpenRA.Mods.Common/HitShapes/Rectangle.cs
```diff
@@ -68,13 +86,24 @@ public WDist DistanceFromEdge(WVec v)
...
+				return DistanceFromEdge((pos - (actorTargPos + new WVec(0, 0, VerticalBottomOffset))).Rotate(-actor.Orientation));
+
+			return DistanceFromEdge((pos - actorTargPos).Rotate(-actor.Orientation));
 		}
 
 		public void DrawCombatOverlay(WorldRenderer wr, RgbaColorRenderer wcr, Actor actor)
 		{
-			var verts = combatOverlayVerts.Select(v => wr.ScreenPosition(actor.CenterPosition + v.Rotate(actor.Orientation)));
-			wcr.DrawPolygon(verts.ToArray(), 1, Color.Yellow);
+			var vertsTo = combatOverlayVertsTop.Select(v => wr.ScreenPosition(actor.TargetingCenterPosition + v.Rotate(actor.Orientation)));
```
Phrohdoh: One more letter makes this a bit easier to understand at a glance.
Might as well change the below to `vertsBelow`, also.

OpenRA.Mods.Common/HitShapes/Circle.cs
reaperrr: Done.

reaperrr:
Updated.

pchote:
I'm really not a fan of having separate `CenterPosition` and `TargetCenterPosition`s.  CenterPosition being at the bottom of the actor is a bug, and should be fixed properly for all cases.

reaperrr:
> I'm really not a fan of having separate CenterPosition and TargetCenterPositions. CenterPosition being at the bottom of the actor is a bug, and should be fixed properly for all cases.

I see your point, but wouldn't that cause all kinds of regressions with pathfinding, attack distance and so on (at least in TS)? Problems like #5841 could become more common and I'm not sure I'd manage to find and fix them all (not on my own, in any case).

reaperrr:
I made a quick test locally, and among other things, `AnimationWithOffset` is positioned relative to `Actor.CenterPosition`. Generally speaking, In my opinion using `Actor.CenterPosition` for everything is just asking for trouble and regressions.

But more importantly, after some thinking I'd like to point out this:
If someone (maybe myself) where to add SW:Empire at War / Universe at War-style hit zones, then splitting targeting points from CenterPosition would become mandatory anyway.

My idea in that regard is to refactor the `Health` trait into a `HitPoint` of which an actor can have multiple ones, and only when health/HP of all (or at least all required) HitPoints has been reduced to 0 is the actor considered 'killed'.
I'd allow modders to assign `Armor` types to a specific `HitPoint`, give `HitPoint` a customizable `Priority` property (HitPoints with higher priority would be attacked first, if multiple HitPoints have the same priority, the attacker switches between them randomly), and a `RequiredForKill` property together with adding a property to warheads to ignore certain `HitPoint`s.
That could be used to implement StarCraft-style shields as well, among other nice things. 
Could come in quite handy for my MechCommander RTS mod as well. I guess I'll make a ticket.

pchote:
> But more importantly, after some thinking I'd like to point out this:
If someone (maybe myself) where to add SW:Empire at War / Universe at War-style hit zones, then splitting targeting points from CenterPosition would become mandatory anyway.

We already have a system in place for this: the `ITargetablePositions` interface and `Target.Positions`.

reaperrr:
Just wanted to note that I'm still working on this, using `ITargetablePositions` as suggested. It's just not ready yet.

reaperrr:
Updated.

No idea how the commit order got messed up (commit 3 should be commit 5), but that shouldn't have any effect as long as you don't test the first 5 commits individually.

reaperrr:
Updated.

`TargetPositionOffset` is now actually relative to body orientation. 
Used to fix the TS weed harvester's target position to match the voxel.

GraionDilach:
>OpenRA.Mods.Common/Traits/Health.cs:L108: [SP2000] Invalid spacing at the end of the line.

reaperrr:
Fixed style issue.

abcdefg30:
Needs a rebase.

reaperrr:
Rebased.

pchote:
Can we add a height / offset for infantry too? You should be able to set it once on the default with something reasonable.

pchote:
Having actors not shoot at eachother's feet / wheels is a subtle but very nice polish improvement.

reaperrr:
@pchote: Fixed the factory offsets.

About infantry, I only gave them a 'ceiling' but left the target position at ground for now.
Two reasons:
1) TS' machine gun impact and anti-tank warhead explosions are visually designed for hitting flat surfaces, so while it looks good enough for vehicles and buildings, it looks rather weird on infantry.
2) Infantry prone sequences show them flat on the ground and - just like tick tanks - we'll need upgradable HitShapes + a tie to the prone logic to cover this properly, and I don't want to open that can of worms in this PR.

Mailaender:
I got:

```
Exception of type `System.InvalidOperationException`: Sequence contains no elements
  at System.Linq.Enumerable.First[TSource] (IEnumerable`1 source) [0x00067] in /home/abuild/rpmbuild/BUILD/mono-4.2.2/external/referencesource/System.Core/System/Linq/Enumerable.cs:930 
  at OpenRA.Mods.Common.Warheads.CreateEffectWarhead.GetDirectHit (OpenRA.World world, CPos cell, WPos pos, OpenRA.Actor firedBy, Boolean checkTargetType) <0x422b2680 + 0x0010f> in <filename unknown>:0 
  at OpenRA.Mods.Common.Warheads.CreateEffectWarhead.GetImpactType (OpenRA.World world, CPos cell, WPos pos, OpenRA.Actor firedBy) <0x422b24b0 + 0x000d7> in <filename unknown>:0 
  at OpenRA.Mods.Common.Warheads.CreateEffectWarhead.IsValidImpact (WPos pos, OpenRA.Actor firedBy) <0x422b23b0 + 0x00097> in <filename unknown>:0 
  at OpenRA.Mods.Common.Warheads.CreateEffectWarhead.DoImpact (Target target, OpenRA.Actor firedBy, IEnumerable`1 damageModifiers) <0x422b2050 + 0x000cf> in <filename unknown>:0 
  at OpenRA.GameRules.WeaponInfo+<Impact>c__AnonStorey1.<>m__0 () [0x00027] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/GameRules/WeaponInfo.cs:161 
  at OpenRA.GameRules.WeaponInfo.Impact (Target target, OpenRA.Actor firedBy, IEnumerable`1 damageModifiers) [0x00095] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/GameRules/WeaponInfo.cs:165 
  at OpenRA.Mods.Common.Activities.FallToEarth.Tick (OpenRA.Actor self) <0x42311290 + 0x0013b> in <filename unknown>:0 
  at OpenRA.Traits.ActivityUtils.RunActivity (OpenRA.Actor self, OpenRA.Activities.Activity act) [0x00024] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Traits/ActivityUtils.cs:37 
  at OpenRA.Actor.Tick () [0x00010] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Actor.cs:164 
  at OpenRA.World.Tick () [0x000c3] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/World.cs:331 
  at OpenRA.Game.InnerLogicTick (OpenRA.Network.OrderManager orderManager) [0x00226] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:556 
  at OpenRA.Game.LogicTick () [0x00055] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:581 
  at OpenRA.Game.Loop () [0x000d4] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:708 
  at OpenRA.Game.Run () [0x00042] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Game.cs:748 
  at OpenRA.Program.Run (System.String[] args) [0x00011] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Support/Program.cs:118 
  at OpenRA.Program.Main (System.String[] args) [0x00050] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Support/Program.cs:40
```

when shooting down an Orca Bomber with a SAM site. Not sure if it is related though.

reaperrr:
`CreateEffectWarhead` doesn't use Linq on bleed, so it must be caused by cb3a760.

I'm not sure under which circumstances the Sequence can contain no elements there, I couldn't reproduce the crash so far.

reaperrr:
Rebased and updated. I suspect that `var healthInfo = unit.Info.TraitInfos<HealthInfo>();` never returned `null`, so the following null check never triggered.

Mailaender:
Seems to work fine now. :white_check_mark:

reaperrr:
As requested by @pchote I have split this up into less intimidating, more reviewable chunks.

#11168 first, then this, then #11169.

reaperrr:
Updated.

GraionDilach:
Do you intend to use/have a usecase for the Footprint targetable position? I can understand why did you made the restriction, however, it seems really useless with it and all the cases where I'd even think about using it could be resolved with a bit of math/trial-and-error on CenterPosition offset even.

With all in mind, I probably would be more okay with dropping/removing it atm instead of merging it in it's unfinished state.

reaperrr:
> Do you intend to use/have a usecase for the Footprint targetable position?

Not really, I just preserved the old `TargetablePositions` from `Building` that way. Personally I agree that we probably don't need it, but maybe we're overlooking something. Could you chime in on this @pchote ?

GraionDilach:
Oh wait, now I remember what that was supposed to do which might even break on this PR: if a unit has lower range than the building width, the unit might be unable to reach the targeting position and thereby attack the building (say, melee units).

reaperrr:
You're right, and unfortunately it really breaks on this PR.
Setting the type to Footprint fixes the targeting, but then weapons aim at the first (top-left) cell of the footprint. This will need to be fixed, obviously.

reaperrr:
Updated.

The underlying issue of what @GraionDilach pointed out is that on bleed, targeting and actual attacking aim at different position types. Using `TargetablePositions` for everything would have broken that separation, and using something hacky like first pos = targeting, 2nd pos attacking felt wrong, so in oder to keep the regression risk low and the separation intact, I decided to introduce an additional `IAttackablePositions` interface that basically works like `TargetablePositions`, but allows to leave the latter as-is and not cause any trouble there.

In addition to that, things like `Armament` no longer just aim at the first `AttackablePosition`, but rather the closest one to the source by default (unless this behavior is disabled). Later, when we have a Footprint HitShape, we can use this to reproduce D2k-like shooting at the closest cell of a building rather than the center.

OpenRA.Game/Traits/Target.cs
```diff
@@ -165,6 +165,35 @@ public IEnumerable<WPos> Positions
...
+				{
+					case TargetType.Actor:
+						if (!actor.Targetables.Any(Exts.IsTraitEnabled))
+							return new[] { actor.CenterPosition };
+
+						var attackablePositions = actor.TraitsImplementing<IAttackablePositions>();
+						if (attackablePositions.Any())
+						{
+							// TODO: Add the AttackablePositions of all traits implementing IAttackablePositions
+							return attackablePositions.First().AttackablePositions(actor);
```
obrakmann: This compiles, and seems to work in RA, at least.
```csharp
if (attackablePositions.Any())
{
	// TODO: Add the AttackablePositions of all traits implementing IAttackablePositions
	var target = this;
	return attackablePositions.SelectMany(ap => ap.AttackablePositions(target.actor));
}
```

That workaround is required because `Target` is a struct, not a class.

OpenRA.Mods.Common/Traits/CombatDebugOverlay.cs
```diff
@@ -58,6 +58,28 @@ public void RenderAfterWorld(WorldRenderer wr, Actor self)
 			if (healthInfo != null)
 				healthInfo.Shape.DrawCombatOverlay(wr, wcr, self);
 
+			var attackablePositions = self.TraitsImplementing<IAttackablePositions>();
+			if (attackablePositions.Any())
```
obrakmann: You don't need this, `foreach` will handle the case just fine.

OpenRA.Mods.Common/HitShapes/Capsule.cs
```diff
@@ -85,6 +85,13 @@ public WDist DistanceFromEdge(WVec v)
 		public WDist DistanceFromEdge(WPos pos, Actor actor)
 		{
 			var actorPos = actor.CenterPosition;
+			var attackablePositions = actor.TraitsImplementing<IAttackablePositions>();
+			if (attackablePositions.Any())
+			{
+				var positions = attackablePositions.First().AttackablePositions(actor);
+				if (positions.Any())
+					actorPos = positions.First();
+			}
```
obrakmann: Use `SelectMany` here as well, then use `PositionoClosestTo(actorPos)`? Using `First()` - twice! - here feels a bit dodgy.

Same thing in the hunk below.

reaperrr: > Using First() - twice! - here feels a bit dodgy

Yeah, I wrote this before I came up with `PositionClosestTo`.

reaperrr:
Updated.

obrakmann:
ok, looks fine so far. Three more things to note, though:

1. Could you mention a use case for `Armament.IgnoreAttackablePositionOffset` in the commit message of the second-to-last commit, please?
2. Could you please describe (in its commit message) which problem the last commit is supposed to solve? Also, does that depend on the equivalent setting on `Armament` to be set as well? I imagine that otherwise the barrel of a turret could look one way while it would shoot the other way.
3. I'm not too sure if I like this `AttackablePositions` thing living on `Health`. Would `Targetable` work instead, for example?

GraionDilach:
Targetable doesn't care about the physical properties of the unit (hitshapes and stuff). The best solution could prolly be to decouple hitshapes and all related from Health into a new trait (especially if multiple health traits arrive which I guess @reaperrr do wants).

reaperrr:
> I'm not too sure if I like this AttackablePositions thing living on Health. Would Targetable work instead, for example?

@obrakmann I see your point, but it's like @GraionDilach said. This PR will be followed by a PR that allows multiple Health / HitZone traits per actor (splitting off things like `HitShapes` to a separate trait sounds like a good idea), and I don't want to mess with `Targetable`'s pure "boolean" nature.

I definitely want to complete the entire feature before the next release, so letting `AttackablePositions` live on `Health` should be acceptable as temporary solution.

pchote:
I agree with @obrakmann, putting this on health doesnt really sit right with me, even in the short term.  If you can wait a few days for me to recover from my work mission then i can look at this and possibly give some suggestions.

obrakmann:
Alright, so do I understand correctly that in the end it would live on a new trait that would combine both the hit shapes as well as the attackable positions? If so, :+1: 

(I would really prefer that "new" trait to be `Targetable`, it just makes sense context-wise, unless I'm missing something.)

GraionDilach:
I am really not a fan of expanding `Targetable` with this, because that would overcomplicate multiple active `Targetable` traits on actors without any actual benefit. The same reason I'd go with having HitShapes and positions on a third trait (multiple Health traits prolly also would complicate that usecase).

pchote:
We already have an interface and code in place for target positions, which very deliberately avoid touching Health.  Health is concerned with hitshapes because those determine how damage (from e.g. Spread) is applied.  This is only weakly related to the positions that other actors should target.

pchote:
Having separate positions for targeting vs attacking is rather crap - conceptually they are supposed to be the same thing. Can we fix that, instead?

reaperrr:
> Having separate positions for targeting vs attacking is rather crap - conceptually they are supposed to be the same thing. Can we fix that, instead?

If you have a different solution for the issue @GraionDilach pointed out a few days ago (melee units not being able to attack large buildings if their footprint doesn't allow the melee units to move close enough to the targetable position), then I don't mind.

pchote:
Buildings are supposed to expose a targetable position for each cell in their footprint, which is exposed via `Target.Positions`.  The targeting and attack code are both supposed to pick one of these points to attack.  If melee units are picking positions that are outside their range then that's a bug that needs to be fixed.

My different solution is to work out what broke in the logic above and then fix that :)

reaperrr:
The sheer amount of comments and changes this has gone through made it harder to review and a few dependencies still need to be implemented first, so I'll probably open a new, 'fresh' PR when I come back to this.

PR #10384, opened 2016-01-02
Fix duplicated end of file newline being saved to upgraded yaml.

pchote:
This cuts down the upgrade diff noise.


Mailaender:
:+1:

RoosterDragon:
:+1:

PR #10388, opened 2016-01-03
Speed up AutoTarget.ChooseTarget among groups of allied units.

RoosterDragon:
Most auto target scans will be conducted among groups of allied units unable to target each other because they are allied and their weapons only target enemies. Since they cannot target each other, the scan will be repeated constantly. Realising this, we can significantly reduce the performance impact of auto target scanning by bailing early in this scenario to avoid carrying out expensive targeting checks on friendly units which we know will fail anyway.

On the RA shellmap, this reduces total CPU usage of `ChooseTarget` from 2.47% to 1.54%. I watched a replay from release-20151224 earlier that had large standing armies for a fair bit that clocked in at like 10%. That will probably get significantly reduced too.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
@@ -287,6 +287,17 @@ namespace OpenRA.Mods.Common.Traits
 				&& (target.IsInRange(self.CenterPosition, GetMaximumRange()) || (allowMove && self.Info.HasTraitInfo<IMoveInfo>()));
 		}
 
+		public Stance UnforcedAttackTargetStances()
+		{
+			// PERF: Avoid LINQ.
+			var stances = Stance.None;
+			foreach (var armament in Armaments)
+				if (!armament.IsTraitDisabled)
+					stances |= armament.Info.TargetStances;
+
+			return stances;
+		}
+
 		class AttackOrderTargeter : IOrderTargeter
 		{
 			readonly AttackBase ab;
--- a/OpenRA.Mods.Common/Traits/AutoTarget.cs
+++ b/OpenRA.Mods.Common/Traits/AutoTarget.cs
@@ -140,8 +140,14 @@ namespace OpenRA.Mods.Common.Traits
 			if (nextScanTime <= 0)
 			{
 				nextScanTime = self.World.SharedRandom.Next(info.MinimumScanTimeInterval, info.MaximumScanTimeInterval);
-				var range = info.ScanRadius > 0 ? WDist.FromCells(info.ScanRadius) : attack.GetMaximumRange();
-				return ChooseTarget(self, range, allowMove);
+
+				// If we can't attack right now, there's no need to try and find a target.
+				var attackStances = attack.UnforcedAttackTargetStances();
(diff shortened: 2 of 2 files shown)


OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -168,6 +168,13 @@ Actor ChooseTarget(Actor self, WDist range, bool allowMove)
 			var actorsInRange = self.World.FindActorsInCircle(self.CenterPosition, range);
 			foreach (var actor in actorsInRange)
 			{
+				// PERF: Most units can only attack enemy units. If this is the case but the target is not an enemy, we
+				// can bail early and avoid the more expensive targeting checks and armament selection. For groups of
+				// allied units, this helps significantly reduce the cost of auto target scans. This is important as
+				// these groups will continuously rescan their allies until an enemy finally comes into range.
+				if (!actor.AppearsHostileTo(self) && attack.UnforcedAttackCanOnlyTargetEnemies())
```
pchote: `attack.UnforcedAttackCanOnlyTargetEnemies()` can be hoisted outside the loop for even more perf.

Mailaender:
My guess is this will improve performance drastically. We played a match on http://resource.openra.net/maps/7773/ yesterday and I hoped all our island related path finding early bail outs would help make this more playable and it still became super slow in the end game. Not due to path finding, but tick_time spiked in the graph and indeed auto targeting related traits showed up in perf.log as we also had large standing armies that couldn't eliminate each other, because manually transporting them to another island was much slower than producing new units. :+1: 

I tested an AI skirmish on fastest and saw no regressions. :white_check_mark:

OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
```diff
@@ -287,6 +287,16 @@ public bool IsReachableTarget(Target target, bool allowMove)
 				&& (target.IsInRange(self.CenterPosition, GetMaximumRange()) || (allowMove && self.Info.HasTraitInfo<IMoveInfo>()));
 		}
 
+		public bool UnforcedAttackCanOnlyTargetEnemies()
+		{
+			// PERF: Avoid LINQ.
+			foreach (var armament in Armaments)
+				if (!armament.IsTraitDisabled && armament.Info.TargetStances > Stance.Enemy)
```
atlimit8: Should be `if (!armament.IsTraitDisabled && armament.Info.TargetStances != Stance.Enemy)` since `armament.Info.TargetStances > Stance.Enemy` is unnecessarily dependent on the order in the `enum Stance` definition.

atlimit8:
Everything else looks good to me.

OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
RoosterDragon: You're right. However I did want to catch `None` here too. I've realised I can rework this a bit to achieve both.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/05a54ac04ec3d203f899548397313c3ce409cc3f...d234a24fb34bc00d4b564a292d63449b619cfd20)

PR #10391, opened 2016-01-03
New interface for resource display information

ghost:
This is part of the resource refactor.

Currently, the chrome bits pull there information directly from the `PlayerResources` trait. This will function poorly once we have multiple `PlayerResources` traits. In order to make sure that they will continue to work, I've created a new interface, `IResourceDisplay`, indicating that a trait provides display information about a resource. I've also created a trait `ResourceDisplay` which implements the interface provides all of the current UI logic for resources.

This will make it easier for mods to customize their resource displays while reusing the existing chrome bits, and reduce the coupling of the UI code on the structure of resources.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Player/ResourceDisplay.cs
+++ b/OpenRA.Mods.Common/Traits/Player/ResourceDisplay.cs
@@ -1,0 +1,78 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public class ResourceDisplayInfo : ITraitInfo, Requires<PlayerResourcesInfo>
+	{
+		public object Create(ActorInitializer init) { return new ResourceDisplay(init.Self, this); }
+	}
+
+	public class ResourceDisplay : ITick, IResourceDisplay
+	{
+		const float DisplayFracPerFrame = .07f;
+		const int DisplayDeltaPerFrame = 37;
+
+		readonly PlayerResources resources;
+		int ActualAmount { get { return resources.Cash + resources.Resources; } }
+		int nextCashTickTime = 0;
+
+		public int Amount { get; private set; }
+		public int CappedAmount { get { return resources.Resources; } }
+		public int Capacity { get { return resources.ResourceCapacity; } }
+		public int Earned { get { return resources.Earned; } }
+		public int Spent { get { return resources.Spent; } }
+
+		public ResourceDisplay(Actor self, ResourceDisplayInfo info)
+		{
+			resources = self.Trait<PlayerResources>();
+			Amount = ActualAmount;
+		}
+
+		public void Tick(Actor self)
+		{
+			if (nextCashTickTime > 0)
+				nextCashTickTime--;
+
+			var diff = Math.Abs(ActualAmount - Amount);
+			var move = Math.Min(Math.Max((int)(diff * DisplayFracPerFrame), DisplayDeltaPerFrame), diff);
+
+			if (Amount < ActualAmount)
+			{
+				Amount += move;
+				PlayTickUp(self);
+			}
+			else if (Amount > ActualAmount)
+			{
+				Amount -= move;
+				PlayTickDown(self);
+			}
+		}
+
+		public void PlayTickUp(Actor self)
+		{
+			if (Game.Settings.Sound.CashTicks)
+				Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Sounds", "CashTickUp", self.Owner.Faction.InternalName);
+		}
+
+		public void PlayTickDown(Actor self)
+		{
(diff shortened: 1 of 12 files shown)


abcdefg30:
Needs a rebase already. :/

scshunt:
Rebased.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameCashCounterLogic.cs
```diff
@@ -18,12 +19,12 @@ public class IngameCashCounterLogic : ChromeLogic
 		[ObjectCreator.UseCtor]
 		public IngameCashCounterLogic(Widget widget, World world)
 		{
-			var playerResources = world.LocalPlayer.PlayerActor.Trait<PlayerResources>();
+			var resourceDisplay = world.LocalPlayer.PlayerActor.TraitsImplementing<IResourceDisplay>().Single();
```
RoosterDragon: `Trait<T>` will do the same thing as `TraitsImplementing<T>().Single()` but with more specific errors, so I would stick with that.

scshunt:
Rebased again.

abcdefg30:

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -395,4 +395,13 @@ public interface IRemoveFrozenActor
 
 	public interface IRulesetLoaded<TInfo> { void RulesetLoaded(Ruleset rules, TInfo info); }
 	public interface IRulesetLoaded : IRulesetLoaded<ActorInfo>, ITraitInfoInterface { }
+
+    public interface IResource
+    {
+        string ResourceName { get; }
```
pchote: If we're going to start explicitly implementing interfaces everywhere then these names are more verbose than they need to be.

OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -117,4 +117,16 @@ public interface IPreventsAutoTarget
 	{
 		bool PreventsAutoTarget(Actor self, Actor attacker);
 	}
+
+	// This provides resource infromation to the chrome bits. Note that some
```
pchote: infromation

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2876,6 +2876,13 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 					}
 				}
 
+				// Add ResourceDisplay for each PlayerResource
+				if (engineVersion < 20160104)
+				{
+					if (node.Value.Nodes.Any(x => x.Key == "PlayerResources"))
```
pchote: You may also want to check that `ResourceDisplay` hasn't already been added.

OpenRA.Mods.Common/Widgets/Logic/Ingame/ProductionTooltipLogic.cs
```diff
@@ -24,7 +24,7 @@ public ProductionTooltipLogic(Widget widget, TooltipContainerWidget tooltipConta
 		{
 			var mapRules = palette.World.Map.Rules;
 			var pm = palette.World.LocalPlayer.PlayerActor.Trait<PowerManager>();
-			var pr = palette.World.LocalPlayer.PlayerActor.Trait<PlayerResources>();
+			var res = palette.World.LocalPlayer.PlayerActor.TraitsImplementing<IResourceDisplay>().Single();
```
pchote: `.Trait<IResourceDisplay>()`.

OpenRA.Game/Traits/TraitsInterfaces.cs
scshunt: This wasn't supposed to be in this patch, but I was thinking they weren't because they might collide. I don't really care though. I've pulled this out of this patch though.

OpenRA.Mods.Common/Widgets/Logic/Ingame/ProductionTooltipLogic.cs
scshunt: Thought I got all these.

scshunt:
Fixed issues.

OpenRA.Mods.Common/Traits/Player/ResourceDisplay.cs
```diff
@@ -0,0 +1,78 @@
...
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public class ResourceDisplayInfo : ITraitInfo, Requires<PlayerResourcesInfo>
+	{
+		public object Create(ActorInitializer init) { return new ResourceDisplay(init.Self, this); }
+	}
+
+	public class ResourceDisplay : ITick, IResourceDisplay
```
pchote: It occurs to me that this doesn't even need to be a trait.  This is all logic for the chrome, so it would better live in a ChromeLogic class (e.g. `IngameCashCounterLogic`).

pchote: Note that this doesn't go against the description in the pr: chrome logic classes are just as pluggable as traits, and you can also manage multiple backing PlayerResources from inside them.

scshunt: Ok! That makes more sense. I'll take a look at chrome logic, which I haven't studied before. Does it use the same Logic<> type thing, or do I need to get more creative (feel free to move to IRC)?

pchote:
This has been idle for quite a while, so closing to clean up the PR queue.

PR #10393, opened 2016-01-03
Polished the Dune 2000 low power graphics

Mailaender:
~~The dish not spinning should be enough.~~ Our disabled overlay clashes with other overlays.

image

I also fixed the Starport lights not showing up at all. Probably a regression from https://github.com/OpenRA/OpenRA/pull/10119.
Made them not blink when on low power. It is a bit unlogical that the starport is powered down and non-operational, but the lights are still there. Our `WithProductionOverlay` was missing that switch.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
@@ -36,6 +36,9 @@ namespace OpenRA.Mods.Common.Traits
 
 		public readonly bool PauseOnLowPower = false;
 
+		[Desc("Optional decoration element which doesn't cast shadows and won't get overlayed.")]
+		public readonly bool IsDecoration = false;
+
 		public override object Create(ActorInitializer init) { return new WithIdleOverlay(init.Self, this); }
 
 		public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer init, RenderSpritesInfo rs, string image, int facings, PaletteReference p)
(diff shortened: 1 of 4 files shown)


pchote:
Manual powering down isn't even implemented properly in d2k We should either fix it properly or remove it completely.

Only four arbitrary buildings can be powered down: outpost and palace (but none of the other tech structures), starport (but none of the other production structures), and the missile turret (but not the gun turret).

Mailaender:
Updated.

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
```diff
@@ -36,6 +36,9 @@ public class WithIdleOverlayInfo : UpgradableTraitInfo, IRenderActorPreviewSprit
 
 		public readonly bool PauseOnLowPower = false;
 
+		[Desc("Optional decoration element which doesn't cast shadows, shouldn't get overlayed and won't change palettes on upgrade.")]
```
pchote: Won't change palettes on upgrade?

Mailaender: https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/Modifiers/UpgradeOverlay.cs#L48

pchote: That line marks the color overlay generated by the trait as a decoration.  It's exactly the same as the low power shadow overlay, but generic.

Mailaender: Fixed.

OpenRA.Mods.Common/Traits/Render/WithProductionOverlay.cs
```diff
@@ -31,6 +31,8 @@ public class WithProductionOverlayInfo : ITraitInfo, Requires<RenderSpritesInfo>
 		[Desc("Custom palette is a player palette BaseName")]
 		public readonly bool IsPlayerPalette = false;
 
+		public readonly bool PauseOnLowPower = false;
```
pchote: Please add the [Desc()]` from `WithIdleAnimationInfo`:
```csharp
[Desc("Pause when the actor is disabled.  Deprecated.  Use upgrades instead.")]
```

pchote: Or, even better, make this use upgrades now so that we don't need to worry about porting things back to upgrades later.

Mailaender: If neither the `UpgradeWhenDisabled` nor the `LowPowerUpgrade` already exist than this isn't deprecated yet. Make a ticket for your future plans instead.

abcdefg30:
Looks good to me, but somehow I still get those delivery light while being low on power.

Mailaender:
Looks like this isn't required anymore. I can't reproduce the issue this is fixing.

PR #10404, opened 2016-01-03
Organized legacy map import into respective mod DLLs

Mailaender:
This organizes the legacy map import the way the utility commands are structured ~~now with a unified layout: `OpenRA.Utility.exe $mod --import-map FILENAME`~~. This sacrifices guessing ra/cnc for cleaner design: common methods are in Mods.Common while the mod specific stuff is in the correct DLL.

Also added support for parsing the `[Briefing]` section.

Removed the conversion tables that didn't convert anything and added the out-commented different crate types to the mod rules instead so that these won't be omitted.


OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
```diff
@@ -0,0 +1,209 @@
...
+		public void Run(ModData modData, string[] args)
+		{
+			// HACK: The engine code assumes that Game.modData is set.
+			Game.ModData = modData;
+			Game.ModData.MountFiles();
+
+			var rules = Game.ModData.RulesetCache.Load();
+			var filename = args[1];
+
+			using (var stream = Game.ModData.ModFiles.Open(filename))
```
pchote: This probably wants to use `File.Open` so people aren't forced to have the legacy maps in the openra dir.

Mailaender: `mono --debug OpenRA.Utility.exe ra --import-map /home/matthias/Downloads/scg01ea.ini` works just fine for me.

OpenRA.Utility/Program.cs
```diff
@@ -62,7 +62,8 @@ static void Main(string[] args)
 			{
 				var command = (IUtilityCommand)Activator.CreateInstance(commandType);
 				var kvp = new KeyValuePair<Action<ModData, string[]>, Func<string[], bool>>(command.Run, command.ValidateArguments);
-				actions.Add(command.Name, kvp);
+				if (!actions.ContainsKey(command.Name))
+					actions.Add(command.Name, kvp);
```
Mailaender: Note: this may be to fragile as it now requires the mod DLLs to be sorted so that the correct one is preferred.

pchote: It is probably a better idea to give these commands their own names as well as separating them by dll.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
```diff
@@ -0,0 +1,209 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using OpenRA.FileFormats;
+using OpenRA.Mods.Common.UtilityCommands;
+using OpenRA.Primitives;
+
+namespace OpenRA.Mods.Cnc.UtilityCommands
+{
+	class ImportTiberianDawnLegacyMapCommand : IUtilityCommand
```
pchote: These should subclass the common implementation instead of making the internals of that public.

OpenRA.Utility/Program.cs
penev92: Or keep a `--import-map` command, and automatically choose an importer based on the passed mod (as all utility commands require a mod).

pchote: The trick is doing that in a way that doesn't require the mod choices to be hardcoded (because that then defeats the purpose).

abcdefg30:
Also needs a rebase.

Mailaender:
Rebased and improved `NewINIFormat` detection so it doesn't produce too many warnings. For some reason most TD single-player maps don't set it correctly.

Mailaender:
Had to rebase again and redid it so there will be no conflicts, but in fact improved auto-detection. http://resource.openra.net/upload/map/ will now also swallow Dune 2000 maps.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
Mailaender: Sounds good in theory, but won't work here in practice as key points in between are not very similar. I also don't want to change too much to avoid breaking things.

OpenRA.Utility/Program.cs
Mailaender: I added a stub `mono --debug OpenRA.Utility.exe $mod --map-import` that will do auto-detection and re-run the utility with the best guess. This way we retain compatibility and robustness.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
pchote: You have changed it much by opening the internals of the class to the outside world (which is bad code design).

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -22,20 +27,59 @@ public bool ValidateArguments(string[] args)
...
 			Game.ModData.MountFiles();
 
-			var rules = Game.ModData.RulesetCache.Load();
-			var map = LegacyMapImporter.Import(args[1], modData.Manifest.Mod.Id, rules, Console.WriteLine);
+			var filename = args[1];
+			if (filename.ToLowerInvariant().EndsWith(".map"))
+			{
+				Console.WriteLine("Detected a Dune 2000 map.");
+				args[0] = "d2k --import-d2k-map";
+				RunCommand(args);
```
pchote: Chaining RunCommand with hardcoded commandline parameters is really bad for maintainability.  Please don't do this.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -22,20 +27,59 @@ public bool ValidateArguments(string[] args)
...
 		public void Run(ModData modData, string[] args)
 		{
 			// HACK: The engine code assumes that Game.modData is set.
 			Game.ModData = modData;
 			Game.ModData.MountFiles();
 
-			var rules = Game.ModData.RulesetCache.Load();
-			var map = LegacyMapImporter.Import(args[1], modData.Manifest.Mod.Id, rules, Console.WriteLine);
+			var filename = args[1];
+			if (filename.ToLowerInvariant().EndsWith(".map"))
```
pchote: The problem with hardcoded autodetection is that this then locks out any other map importers, because you need to edit this upstream file to make them work.  Can we please just stick with `--import-<mod>-map`?

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
Mailaender: I don't see an alternative at the moment.

pchote: Not adding autodetection, and having users run `--import-<mod>-map` directly.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
Mailaender: Agreed, but I don't think anyone outside this repository will write a map importer. If we stick with only `import-<mod>-map` then @ihptru has to update http://ressource.openra.net

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -22,20 +27,59 @@ public bool ValidateArguments(string[] args)
...
 
-			var fileName = Path.GetFileNameWithoutExtension(args[1]);
-			var dest = fileName + ".oramap";
-			map.Save(dest);
-			Console.WriteLine(dest + " saved.");
+				RunCommand(args);
+			}
+		}
+
+		void RunCommand(string[] args)
```
pchote: I'm sorry, but this code is horrible. :-1: while this remains.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
Mailaender: Do you have a good code pattern where I can look for inspiration. Afaik we don't have utility commands that sub-class a common one which also gets hidden.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
Mailaender: Please try to give constructive criticism.

pchote: I've given suggestions for alternative approaches in the line comments above.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
pchote: Updating the resource center is a better fix for this problem.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
pchote: Make the common base class abstract and not implement `IUtilityCommand`.  The RA and TD importers then inherit that and implement `IUtilityCommand`.

Mailaender:
Updated. This now sub-classes a common implementation. Avoids even more redundancy.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
Mailaender: I guess that is okay as the user should know which mod it is for and trial & error will take not more than 2 steps.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -9,33 +9,344 @@
...
+using System.Text;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.FileFormats;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Primitives;
 
 namespace OpenRA.Mods.Common.UtilityCommands
 {
-	class ImportLegacyMapCommand : IUtilityCommand
+	public class ImportLegacyMapCommand : IUtilityCommand
```
pchote: Please make this abstract instead of declaring virtual methods with bogus/dummy bodies.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -9,33 +9,344 @@
...
+			}
+
+			// Overwrite default player definitions if needed
+			if (!mapPlayers.Players.ContainsKey(section))
+				mapPlayers.Players.Add(section, pr);
+			else
+				mapPlayers.Players[section] = pr;
+		}
+
+		public static string[] OverlayActors = new string[]
```
pchote: These are (at least in principle) mod specific, so should be hoisted out to the RA/TD importers.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -9,33 +9,344 @@
...
+					if (section == "INFANTRY")
+						actor.Add(new SubCellInit(Exts.ParseIntegerInvariant(parts[4])));
+
+					var actorCount = map.ActorDefinitions.Count;
+
+					if (!rules.Actors.ContainsKey(parts[1].ToLowerInvariant()))
+						Console.WriteLine("Ignoring unknown actor type: `{0}`".F(parts[1].ToLowerInvariant()));
+					else
+						map.ActorDefinitions.Add(new MiniYamlNode("Actor" + actorCount++, actor.Save()));
+				} catch (Exception)
```
pchote: Style nit: needs a newline.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
```diff
@@ -0,0 +1,161 @@
...
+						new OwnerInit("Neutral")
+					};
+
+					var actorCount = Map.ActorDefinitions.Count;
+					Map.ActorDefinitions.Add(new MiniYamlNode("Actor" + actorCount++, ar.Save()));
+				}
+			}
+		}
+
+		void ReadTrees(IniFile file)
```
pchote: You can avoid the duplication between this and RA by having a common ReadTrees method that you pass a function that maps the ini name &rarr; OpenRA name.

pchote: You can probably do the same for the other sections too.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
```diff
@@ -0,0 +1,161 @@
...
+	{
+		public override string Name { get { return "--import-td-map"; } }
+
+		[Desc("FILENAME", "Convert a legacy Tiberian Dawn INI/MPR map to the OpenRA format.")]
+		public override void Run(ModData modData, string[] args)
+		{
+			base.Run(modData, args);
+		}
+
+		public override int MapSize
```
pchote: You could pass this into the base class via the constructor.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
```diff
@@ -0,0 +1,161 @@
...
+			string color;
+			string faction;
+			switch (section)
+			{
+			case "GoodGuy":
+				color = "gold";
+				faction = "gdi";
+				break;
+			case "BadGuy":
+				color = "red"; // TODO: grey unit color theme
```
pchote: This TODO can be removed.  We now support this through rule overrides.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
```diff
@@ -0,0 +1,161 @@
...
+					new LocationInit(new CPos(loc % MapSize, loc / MapSize)),
+					new OwnerInit("Neutral")
+				};
+
+				var actorCount = Map.ActorDefinitions.Count;
+				Map.ActorDefinitions.Add(new MiniYamlNode("Actor" + actorCount++, ar.Save()));
+			}
+		}
+
+		public override void LoadPlayer(IniFile file, string section)
```
pchote: You could drop duplication here by defining mapping dictionaries that are referred to from the base class.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
pchote: Instead of passing the function you can declare abstract methods for doing the transformations, which these subclasses then implement.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
Mailaender: I mean it is not added automatically by the map import rules.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
Mailaender: Not sure how to do that. I get an error upon execution when making this `abstract`:

> System.MissingMethodException: Default constructor not found

Mailaender:
Added back some redundancy with `overlayActors` and removed some with `ReadTrees()`.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
Mailaender: and when I do https://github.com/Mailaender/OpenRA/commit/3b1290a409ea5741322e1aaaf36dd5dfa234f41d I get this:
> Cannot create an abstract class

Mailaender:
Using `abstract` methods now.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
```diff
@@ -0,0 +1,149 @@
...
+namespace OpenRA.Mods.Cnc.UtilityCommands
+{
+	class ImportTiberianDawnLegacyMapCommand : ImportLegacyMapCommand, IUtilityCommand
+	{
+		public string Name { get { return "--import-td-map"; } }
+
+		[Desc("FILENAME", "Convert a legacy Tiberian Dawn INI/MPR map to the OpenRA format.")]
+		public override void Run(ModData modData, string[] args)
+		{
+			MapSize = 64; // NOTE: original engine limitation
```
pchote: It would be better to set this in the constructor: define `public ImportTiberianDawnLegacyMapCommand() : base(64) { }` and then `public ImportLegacyMapCommand(int mapSize) { /* set constant */ }` in the base class.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
```diff
@@ -0,0 +1,149 @@
...
+			base.Run(modData, args);
+		}
+
+		public override void ValidateMapFormat(int format)
+		{
+			if (format > 1)
```
pchote: This could similarly be passed in the constructor as an array of allowed format types.

Mailaender: I am doing `> 1` vs `< 2` so that doesn't work too well.

OpenRA.Mods.Cnc/ImportTiberianDawnLegacyMapCommand.cs
Mailaender: Done.

abcdefg30:
Needs a rebase again.

abcdefg30:
I think you can now also give those new crate actors their respective image: #10571

Mailaender:
Rebased and gave the RA wooden crate it's real image.

abcdefg30:
Looks good to me and importing still works. :+1: / :white_check_mark:

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -23,19 +40,329 @@ public bool ValidateArguments(string[] args)
...
+		{
+			var briefingSection = file.GetSection("Briefing", true);
+			if (briefingSection == null)
+				return string.Empty;
+
+			var briefing = new StringBuilder();
+			foreach (var s in briefingSection)
+				briefing.AppendLine(s.Value);
+
+			return briefing.Replace("\n", " ").ToString();
```
obrakmann: It seems that RA used `@` as a new line placeholder. Not sure how C&C handled it since all my maps use mission.ini for their briefing text.

obrakmann:
Looks fine to me otherwise. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/0f7a7daf105803232d37512c65e828711b895ca6)

Mailaender:
@ihptru You will need to adapt the http://resource.openra.net/upload/map/ now.

ihptru:
https://github.com/OpenRA/OpenRA-Resources/issues/283

PR #10407, opened 2016-01-03
Fix the TS resource minimap colors.

pchote:
Colors extracted from the minimap previews used by the original game.


Mailaender:
:+1: also way easier to the eyes.

atlimit8:
No changelog entry since the ts mode has not been released.

PR #10410, opened 2016-01-03
Fix FrozenActorLayer.FootprintBounds.

### LINKED ISSUE #10403: Invisible and non-targetable flame turret on Monster Tank Madness
The turret is in the middle of the three tanks, and you can see the fireball that it has just shot at the tank on the left:

screen shot 2016-01-03 at 17 47 01

It cannot be seen or manually targeted, but units will autotarget it.

Replay: https://www.dropbox.com/s/9c5sc9aehk3b10x/OpenRA-2016-01-03T174451Z.orarep?dl=0

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29554910-invisible-and-non-targetable-flame-turret-on-monster-tank-madness?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RoosterDragon:
The size of the bounds was too small (as rectangle has exclusive edges on the right and bottom). This meant some intersection tests would fail and thus the frozen actor was not found when searching the partition bins.

Fixes #10403.

### DIFF
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -209,7 +209,7 @@ namespace OpenRA.Traits
 					maxV = p.V;
 			}
 
-			return Rectangle.FromLTRB(minU, minV, maxU, maxV);
+			return Rectangle.FromLTRB(minU, minV, maxU + 1, maxV + 1);
 		}
 
 		public void Tick(Actor self)


abcdefg30:
Looks good to me, walls and flame turrets render again. :+1: / :white_check_mark:

Mailaender:
Thanks for the quick fix!

PR #10411, opened 2016-01-04
ra: decrease RevealsShroud for artillery

kyrylo:
Artillery is meant to be a support unit, so usually something else
(Hinds) provides vision to it. With reduced RevealsShroud riflers can
now at least have a slight chance approach artillery without being
killed.

### DIFF
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -218,7 +218,7 @@ ARTY:
 		ROT: 2
 		Speed: 85
 	RevealsShroud:
-		Range: 5c0
+		Range: 3c0
 	Armament:
 		Weapon: 155mm
 		LocalOffset: 624,0,208


cjshmyr:
I agree with the sentiment but `3c0` is pretty low (old demo truck vision), i'd be more in favor of `4c0`

kyrylo:
I wish it was possible to set it to `3.5c0` :P

atlimit8:
Actually, you can; however, it is `3c512` rather than `3.5c0`. For  *x*`c`*y*, there are 1024 *y* units in 1 *x* (cell).

kyrylo:
Thanks for the info. Let me know if you agree with the change and what exact value you would set.

MustaphaTR:
I'm not sure about 3c512, 4c0 sounds better. By the way, what's reveal range of V2.

abcdefg30:
The V2 has a range of [`5c0`](https://github.com/OpenRA/OpenRA/blob/55d074b25d1aed705ceda27b326eb8920c72e140/mods/ra/rules/vehicles.yaml#L18-L19).

cjshmyr:
Had a chance to play around with it. `3c0` is simply too small, and `4c0` might just be annoying. I agree that arty needs a change but in different ways (firepower, range, splash, cost, etc).

kyrylo:
Not sure why `3c0` is too small to be honest. With `5c0` you can use artillery on its own and avoid low tier infantry, because you see them and they don't see you. The idea is to make artillery to have its own role rather than the solution to all problems. It is *way too* easy to use this unit.

I would also increase [MinRange](https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/weapons/largecaliber.yaml#L142) by 1.

kyrylo:
I just watched [this](http://www.gamereplays.org/openra/replays.php?game=92&show=details&id=308764) game and noticed this:

output

Notice how the last arty is able to shoot at infantry before they see it. This PR tries to eliminate that. Arty shouldn't be able to do that unless you force fire.

cjshmyr:
Oh I agree artillery is too effective at dispatching infantry, I'd just rather tweak the weapon rather than alter the edge case where it's all alone. Artillery typically has has accompanied air/land/gps vision or it's near buildings.

abcdefg30:
So, what is the consensus here? I don't see much discussion going on so it doesn't look like this is really a (big) issue.

epicelite:
Well late game you usually have a GPS anyway but I'm all for reducing the view range.

SoScared:
The stated problem scenario above is played out on Sidestep where the artillery is effectively wiped out of the meta. The nerf in view range would almost exclusively affect situations where you want to stay active out on the map. Why make the artillery more of a liability.

Situations where you move out with infantry and artillery are being slammed and forcing players to mix in rangers, or wait for additional tech to move out sounds bad to me.

 I don't really buy into this 'artillery is blind' theory either. It's expensive (1800, Radar tech and 800 per unit) and it's a _tank_. A tank made of bubblewrap, but a tank nevertheless.

Perhaps I'm missing something here but the effect, no matter how small, looks to be exclusively negative for the game balance.

:-1:

Mailaender:
No consensus.

kyrylo:
Almost everyone was interested in trying out a less aggressive value, but you closed this. Seems a bit unfair, but I won't insist on this change.

PR #10412, opened 2016-01-04
ra: buff minelayer's HP

kyrylo:
Currently 1 rocket soldier can kill a minelayer with 2 shots. With this
change a rocket soldier needs 5 rockets. This behaviour mimics the
light tank behaviour where a rocket soldier needs 5 rockets to kill it.
I think this buff is needed because light tank is a paper tank and
costs $700, but minelayer costs $800 and it's even less sturdy.

### DIFF
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -405,7 +405,7 @@ MNLY.AP:
 		Name: Minelayer
 		Description: Lays mines to destroy\nunwary enemy units.\nCan detect mines.\n  Unarmed
 	Health:
-		HP: 100
+		HP: 150
 	Armor:
 		Type: Heavy
 	Mobile:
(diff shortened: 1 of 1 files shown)


atlimit8:
Minelayers are support vehicles with 2 special abilities (laying and spotting mines); thus, at least part of the expense should be for the special abilities. A more modest increase of HP to 150 would be more acceptable.

kyrylo:
Amended to `150` and rebased.

MustaphaTR:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d234a24fb34bc00d4b564a292d63449b619cfd20...69879e0bd637a93dca2c63acf06f1e8aaa180776)

PR #10417, opened 2016-01-04
Make Log thread-safe.

RoosterDragon:
Since logging is exposed via static methods, it need to be thread-safe or multiple threads trying to log at the same time will cause bad things to happen.

[Diff ignoring whitespace](https://github.com/OpenRA/OpenRA/pull/10417/files?w=1)

### DIFF
--- a/OpenRA.Game/Network/UPnP.cs
+++ b/OpenRA.Game/Network/UPnP.cs
@@ -22,7 +22,7 @@ namespace OpenRA.Network
 		{
 			try
 			{
-				NatUtility.Logger = Log.Channels["server"].Writer;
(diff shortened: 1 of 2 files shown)


abcdefg30:
Logging still works. :+1: / :white_check_mark:

Mailaender:
Benchmark logs as well. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/2e22eb83fcbf8aefc2d2d40c327d4bc693cc98be...e2380aa8ad6dabf9bf9f1172b84c833f13a0b01a)

PR #10418, opened 2016-01-04
Rename WW compression formats and move into Mods.Common.

pchote:
Also moves VqaReader and IniFile.

### DIFF
--- a/OpenRA.Mods.Common/FileFormats/IniFile.cs
+++ b/OpenRA.Mods.Common/FileFormats/IniFile.cs
@@ -14,7 +14,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Text.RegularExpressions;
 
-namespace OpenRA.FileFormats
+namespace OpenRA.Mods.Common.FileFormats
 {
 	public class IniFile
 	{
--- a/OpenRA.Mods.Common/SpriteLoaders/ShpD2Loader.cs
+++ b/OpenRA.Mods.Common/SpriteLoaders/ShpD2Loader.cs
@@ -21,7 +21,7 @@ namespace OpenRA.Mods.Common.SpriteLoaders
 		[Flags] enum FormatFlags : int
 		{
 			PaletteTable = 1,
-			SkipFormat80 = 2,
+			NotLCWCompressed = 2,
 			VariableLengthTable = 4
 		}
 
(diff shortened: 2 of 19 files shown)


pchote:
Erp... something went wrong in the last minute rebase.  Will fix.

pchote:
Fixed, and also moved IniFile.

OpenRA.Mods.Common/SpriteLoaders/ShpD2Loader.cs
```diff
@@ -21,7 +21,7 @@ public class ShpD2Loader : ISpriteLoader
 		[Flags] enum FormatFlags : int
 		{
 			PaletteTable = 1,
-			SkipFormat80 = 2,
+			SkipLCW = 2,
```
penev92: `SkipLCWCompression` please, in the hopes that it would be slightly less confusing to everyone if a few months.

OpenRA.Mods.Common/SpriteLoaders/ShpTDLoader.cs
```diff
@@ -71,7 +71,7 @@ public bool TryParseSprite(Stream s, out ISpriteFrame[] frames)
 
 	public class ShpTDSprite
 	{
-		enum Format { Format20 = 0x20, Format40 = 0x40, Format80 = 0x80 }
+		enum Format { XORPrev = 0x20, XORLCW = 0x40, LCW = 0x80 }
```
penev92: Where did `XORLCW` come from?

pchote: It's XORing against a given LCW frame (as opposed to the previous frame).

penev92:
:+1:

OpenRA.Mods.Common/FileFormats/IniFile.cs
```diff
@@ -14,7 +14,7 @@
 using System.IO;
 using System.Text.RegularExpressions;
 
-namespace OpenRA.FileFormats
+namespace OpenRA.Mods.Common.FileFormats
```
Mailaender: This isn't really mod-specific.

pchote: That is why it is in Mods.Common.

pchote:
Renamed `SkipLCW` &rarr; `NotLCWCompressed` and rebased.

Mailaender:
:+1: on the descriptive comments on the algorithms.

PR #10421, opened 2016-01-05
Unhardcode music file types [Superseded]

penev92:
Use the code from  #10069 to unhardcode music file types like that PR unhardcoded sound (effects, speech, and such) file types.
#10032 would have to take this into account as well, sorry @teees . At least it looks like it should be pretty simple.

### DIFF
--- a/OpenRA.Game/GameRules/MusicInfo.cs
+++ b/OpenRA.Game/GameRules/MusicInfo.cs
@@ -8,6 +8,8 @@
  */
 #endregion
 
+using System;
+using System.IO;
 using OpenRA.FileFormats;
 
 namespace OpenRA.GameRules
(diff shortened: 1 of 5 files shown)


OpenRA.Game/GameRules/MusicInfo.cs
```diff
@@ -39,12 +41,18 @@ public void Load()
...
 			Exists = true;
-			using (var s = Game.ModData.ModFiles.Open(Filename))
+
+			using (var stream = Game.ModData.ModFiles.Open(Filename))
 			{
-				if (Filename.ToLowerInvariant().EndsWith("wav"))
-					Length = (int)WavLoader.WaveLength(s);
-				else
-					Length = (int)AudLoader.SoundLength(s);
+				foreach (var loader in Game.ModData.SoundLoaders)
```
pchote: It would be much cleaner to identify the file type first, and then call GetLength for that type.

penev92: Changed it to expose a `CanParse(Stream)` method via the interface.
The alternative would be to do filename checks, with each parser providing a file extension, but that's not very reliable as it can easily fail on bogus files. (Would be much faster though, so may very well be worth it).

penev92:
Needs to rebase on top of #10032 now...

teees:
referencing commit is meant to make it work with the merged vocloader.
Any other commits in that branch are to rework the `ISoundLoader` interface and make it similar to the other file format loaders.

penev92:
Rebased and squashed the commit by @teees into the first commit. Thanks @teees !
But I cannot verify that Dune II still works.

pchote:
I still think that it would be cleaner to split out @teees's `ISoundFormat` changes from his music streaming branch, otherwise we will end up having to revert these changes again in the medium term.

penev92:
Superceded by #10676.

pchote:
Now properly superseded by #10912.

PR #10422, opened 2016-01-05
Add support for multiple VoiceSets on an actor [Depends on sound setup rewrite]

penev92:
As wished by @DoGyAUT and company.
If multiple VoiceSets are provided, one will be chosen at random when the actor is created, and will be used for the entire lifetime of that actor.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Voiced.cs
+++ b/OpenRA.Mods.Common/Traits/Voiced.cs
@@ -8,6 +8,7 @@
  */
 #endregion
 
+using System.Collections.Generic;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
@@ -16,8 +17,8 @@ namespace OpenRA.Mods.Common.Traits
 	public class VoicedInfo : ITraitInfo
 	{
 		[FieldLoader.Require]
-		[Desc("Which voice set to use.")]
-		[VoiceSetReference] public readonly string VoiceSet = null;
+		[Desc("A collection of voice sets that can be used. One will be selected at random.")]
+		[VoiceSetReference] public readonly string[] VoiceSets = new string[0];
 
 		[Desc("Multiply volume with this factor.")]
 		public readonly float Volume = 1f;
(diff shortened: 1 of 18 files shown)


OpenRA.Mods.Common/Traits/Voiced.cs
```diff
@@ -32,22 +30,23 @@ public class Voiced : IVoiced
 	{
 		public readonly VoicedInfo Info;
 
+		public string VoiceSet { get; }
```
Phrohdoh: I'm pretty sure `{ get; }` is a getter-only auto-property which was introduced in C# 6 (which OpenRA doesn't support yet).

atlimit8: Why not `public readonly string VoiceSet;`?

pchote:
It would be nice if we could use upgrades to turn certain sets on and/or off.

GraionDilach:
This could work nicely with factions as well - solving the RA2 Engineer voice problem.

RoosterDragon:
Needs a rebase.

OpenRA.Mods.Common/Traits/Voiced.cs
penev92: Because it implements it from the interface, and I don't want to mess with that more than I need to here.

penev92:
Rebased. The upgrade ideas do have potential, but are out of scope here, and also need to have the methodology cleared out (check if enabled in `PlaySound()`, or when picking a voice set, or what).

abcdefg30:
Needs a rebase again.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -2833,7 +2833,7 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 				if (engineVersion < 20160107 && depth == 1 && node.Key.StartsWith("Cloak"))
 				{
 					var defaultCloakType = Traits.UncloakType.Attack
-						| Traits.UncloakType.Unload | Traits.UncloakType.Infiltrate | Traits.UncloakType.Demolish;
+					                       | Traits.UncloakType.Unload | Traits.UncloakType.Infiltrate | Traits.UncloakType.Demolish;
```
abcdefg30: This change doesn't look correct to me, please revert it.

penev92:
Rebased.

penev92:
As was pointed out on IRC today ([link](http://logs.openra.net/?year=2016&month=02&day=06#15:55:09)), this can be (kinda) achieved with the voice set's `Variants`.
It is a hacky and unpleasant approach IMO, but will have to do for now until we rewrite the audio setup
>[18:03] <+pchote> the alternative would be to remove the legacy implementation
[18:03] <+pchote> and transition the current mods to the new setup
[18:04] <+pchote> penev: it would make your pr the first part of the sound rewrite
...
[18:09] <+penev> anyway, if we're going to transition from the current glued-together-with-duck-tape-and-gum audio setup to sound sequences
...
[18:18] <+penev> ok so I guess that would want to be closed until the audio re-setup

PR #10423, opened 2016-01-05
MAD tank balance changes/overhall

Ethal-A:
the Mad tank has been changed so that it is more useful as a support infantry unit that still causes devastation
video: https://vid.me/CXgI

### DIFF
--- a/mods/ra/weapons/other.yaml
+++ b/mods/ra/weapons/other.yaml
@@ -263,10 +263,69 @@ Mandible:
 MADTankThump:
 	InvalidTargets: MADTank, Infantry
 	Warhead@1Dam: HealthPercentageDamage
-		Spread: 7c0
-		Damage: 1
+		Spread: 1c0
+		damage: 2
+			Delay: 8
 		InvalidTargets: MADTank, Infantry
-
+	Warhead@26mu: LeaveSmudge
+		Delay: 8
+		SmudgeType: Crater
+		Size: 1
+	Warhead@2Dam: HealthPercentageDamage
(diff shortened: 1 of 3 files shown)


cjshmyr:
I think I'd be willing to accept that with a few tweaks:
- Less crater intervals, so they don't appear so thick by the end in the full circle.
- Not sure about the damage yet, haven't tested it.

mods/ra/weapons/other.yaml
```diff
@@ -263,10 +263,69 @@ Mandible:
 MADTankThump:
 	InvalidTargets: MADTank, Infantry
 	Warhead@1Dam: HealthPercentageDamage
+		Spread: 1c0
+		damage: 2
```
GraionDilach: Damage is with a capital D.

mods/ra/weapons/other.yaml
```diff
@@ -263,10 +263,69 @@ Mandible:
...
+			Delay: 8
```
GraionDilach: The indentation is wrong here - notsure if this'll even work with that.

Truth99A:
lets forget this happened ok?
(i have no idea how to use github lmao and in my attempts to remove the commits I messed up a tone of things to the point of no return and may need to completely delete my repository)
as for now go on the other one pull request with the exact same name

mods/ra/weapons/other.yaml
Phrohdoh: It will not work with incorrect indentation.

PR #10428, opened 2016-01-05
Ra mad tank balance/overhall

Ethal-A:
the Mad tank has been changed so that it is more useful as a support infantry unit that still causes devastation
video: https://vid.me/CXgI

p.s forget about the other PR of this.... it was a grave mistake 
on the up side I spell checked it! now Damage is spelt with a capital :dancer:

### DIFF
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -695,7 +695,7 @@ QTNK:
 		BuildPaletteOrder: 200
 		Prerequisites: fix, stek, ~vehicles.soviet, ~techlevel.unrestricted
 	Valued:
-		Cost: 2000
+		Cost: 2500
 	Tooltip:
 		Name: MAD Tank
 		Description: Deals seismic damage to nearby vehicles\nand structures.\n  Strong vs Vehicles, Buildings\n  Weak vs Infantry, Aircraft
@@ -714,6 +714,8 @@ QTNK:
 	-EjectOnDeath:
 	Targetable:
 		TargetTypes: Ground, MADTank, Repair
+	-Chronoshiftable:
+	-DamageMultiplier@IRONCURTAIN:
 
 STNK:
 	Inherits: ^Vehicle
--- a/mods/ra/weapons/other.yaml
+++ b/mods/ra/weapons/other.yaml
@@ -265,6 +265,8 @@ MADTankThump:
 	Warhead@1Dam: HealthPercentageDamage
 		Spread: 1c0
 		Damage: 3
+		Versus:
+			wood: 1
 		Delay: 8
 		InvalidTargets: MADTank, Infantry
 	Warhead@26mu: LeaveSmudge
@@ -274,11 +276,15 @@ MADTankThump:
 	Warhead@2Dam: HealthPercentageDamage
 		Spread: 2c0
 		Damage: 3
+		Versus:
+			wood: 1
 		Delay: 16
 		InvalidTargets: MADTank, Infantry
 	Warhead@3Dam: HealthPercentageDamage
 		Spread: 3c0
 		Damage: 3
+		Versus:
+			wood: 1
 		Delay: 24
 		InvalidTargets: MADTank, Infantry
 	Warhead@28mu: LeaveSmudge
@@ -288,11 +294,15 @@ MADTankThump:
 	Warhead@4Dam: HealthPercentageDamage
 		Spread: 4c0
 		Damage: 3
+		Versus:
+			wood: 1
 		Delay: 32
 		InvalidTargets: MADTank, Infantry
 	Warhead@5Dam: HealthPercentageDamage
 		Spread: 5c0
 		Damage: 3
+		Versus:
+			wood: 1
 		Delay: 40
 		InvalidTargets: MADTank, Infantry
 	Warhead@30mu: LeaveSmudge
@@ -302,11 +312,15 @@ MADTankThump:
 	Warhead@6Dam: HealthPercentageDamage
 		Spread: 6c0
 		Damage: 3
+		Versus:
+			wood: 1
 		Delay: 48
 		InvalidTargets: MADTank, Infantry
 	Warhead@7Dam: HealthPercentageDamage
 		Spread: 7c0
 		Damage: 3
+		Versus:
+			wood: 1
 		Delay: 56
 		InvalidTargets: MADTank, Infantry
 	Warhead@32mu: LeaveSmudge
--- a/mods/ra/weapons/other.yaml
+++ b/mods/ra/weapons/other.yaml
@@ -264,9 +264,7 @@ MADTankThump:
 	InvalidTargets: MADTank, Infantry
 	Warhead@1Dam: HealthPercentageDamage
 		Spread: 1c0
-		Damage: 3
-		Versus:
-			wood: 1
+		Damage: 1
 		Delay: 8
 		InvalidTargets: MADTank, Infantry
 	Warhead@26mu: LeaveSmudge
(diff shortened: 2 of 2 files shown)


pchote:
The excessive shaking and cratering hurts polish, IMO.  Can you turn those down?

I'd like to see the final detonation removed and the individual thump damage decreased, so that this becomes an AOE damage dealer.  This would distinguish it from the demo truck and hopefully give it its own niche.

r34ch:
+1 AoE weapon

Truth99A:
alright, i changed it so that there isn't any detonation damage and fixed the delay problem as for the craters i don't know how to properly manage them so that they are limited in depth/size. look at the most recent commit for detail

epicelite:
Maybe it should deal directional damage instead of radial? I mean by the time you get this slow ass thing into a group of enemy tanks or their base it would have been better to just use heavy tanks.

pchote:
I actually meant removing the detonation completely, not just the damage from it.

Unit158:
How is this against buildings? 3 mad tanks to destroy a conyard is still reasonable, imo.

cjshmyr:
Agreeing with @pchote and @r34ch. I'll add this makes it even more similar to the demo truck now with removed chronoable/curtainable traits, and extra damage.

MicroJOo:
well , i dont like this ... 

first of all , the craters on the ground dont make sense ... the mad-tank is more like a "bio-weapon" but for "metal-objects" ... meaning , its basicaly a shockwave and therefor shouldnt leave "craters" on the ground ... 

and second ... buffing the price and damage equal to demotrucks (yes demo is 2500) is just a really boring idea ... 

faction dont need to have the same firepower , range , costs and so on ... this just makes the game boring ...

Truth99A:
MADTank Update-v.3
Video of newest version (with fixed delay&3%damage per thump) (i had the tesla attack it, take note of damage done to Madtank when considering strength):
https://sendvid.com/tau7iu74

Note: changed damage vs wood to 1% due to extreme damage against buildings

@pchote Trying to make it not destroy itself, attempting to add deploy trait can be seen here: http://pastebin.com/PTgRB2Eb however it is not working correctly (not really sure how to implement it correctly) also the detonation trait requires to be changed/turned into something that can be disabled however i have no idea how to config it correctly or how to config at all on desktop as it seems to not exist in /openra/ra, seen here: https://github.com/OpenRA/OpenRA/blob/71478aa2b8299ffa141cd036483a633d70b8e4b0/OpenRA.Mods.RA/Traits/MadTank.cs#L41

@epicelite then its not a balanced mad tank anymore XD

@Unit158 you can check the video, and atm it can destroy an mcv if it is close to it, as for if this is balanced. idk but it would be very hard to get it near the mcv with how slow it.

@MicroJOo the craters are a visual representation of the damage and how far it has reached, for example: after 8 ticks the damage reaches 1c0 and at the same time the craters appear as a visual represenation of that damage. if they didn't exist then you wouldn't be able to see how far the damage has gone.
however:
I do see the depth of the craters as a problem which i've tried to fix by trying to make a SmallCrater squence seen here: http://pastebin.com/05uS8A9i however this isn't working :( and the only way to inforce small craters atm that i know of would be to change crater length to 1 without creating a new smudgetype however this makes it so that all craters created by anything are limited to length 1.
.
As for it being like the demo truck, it is nothing like the demo truck in gameplay (slower, more health, doesn't effect infantry [can be used as inf support] etc.).

penev92:
OK I'm jumping in the middle of the discussion, but ... how is this different from a demo truck?

Unit158:
the entire point is that you use mad tanks with chrono and ic. they are different in that way, which is why 3 were needed to destroy a conyard.

On 2016-01-07, at 3:23, Pavel Penev <notifications@github.com<mailto:notifications@github.com>> wrote:


OK I'm jumping in the middle of the discussion, but ... how is this different from a demo truck?

r34ch:
Apart from the IC Chrono one trick pony strat, I still see almost no other viable gameplay use for the MAD in its current or proposed (v3) state. Personally I think this unit requires an entire overhaul to make it fit in the game properly.

As a start however, I think the AoE radius should be much larger but with much less DPS. That would at least give it a slightly more unique role to existing artillary and demo truck units.

And imho the screen shake needs to go. It gives me a headache and microing units for anyone looking at a deployed MAD tank is nearly impossible.

Unit158:
@r34ch you don't consider distracting from micro a potential use?

pchote:
The screen shake intensity can (and should IMO) at least be turned down, without removing it completely.

r34ch:
@pchote Yea that's fair, buildings shake the screen briefly when they explode so its only logical its keeps the screen shake. I just would like it to shake the screen less than a building explosion for example, especially if it is a prolonged shake. When used right there would be building destruction screen shakes anyway.

In fact I might just make a new issue about screen shakes.

@Unit158 yes to an extent, but it's currently more violent than a nuke explosion.

Truth99A:
v4.
@r34ch i've given your idea a try and made the range of the damage go from 7 to 9 and lowered the damage to 1%, however when tested the damage done to vehicles was way too little imo (p.s i found out that "versus" doesn't actually work with the MADtank through my testing) so to counter how little damage was done to vechicles I put back the Madtank detonate damage and made it 40% and made it so that it only effects vehicles not structures!
Mad Tank without detonate damage: https://sendvid.com/y67eu4u0
Mad Tank with detonate damage: https://sendvid.com/nu3phi2t

@pchote Sceen shaking is written in madtank.cs, found here; https://github.com/OpenRA/OpenRA/blob/71478aa2b8299ffa141cd036483a633d70b8e4b0/OpenRA.Mods.RA/Traits/MadTank.cs#L67
not sure how i can change that without creating a new trait (still don't know how to create a new trait X_X but i've gotten a lot of help from a lot of people which i do appreciate a lot)

@penev92 hmmm, i was gonna try and say that its different in many ways but after testing v3 in player vs player its not too different :(

p.s hey guys do you actually want the mad tank to be undeployable? (like mcv deploy and undeploy)

Mailaender:
You can change `ThumpShakeIntensity`, `ThumpShakeMultiplier` and `ThumpShakeTime` without touching the code. See https://github.com/OpenRA/OpenRA/pull/8410 for the future of the MAD tank trait.

UberWaffe:
@Truth99A 
> @pchote Trying to make it not destroy itself,
> Truth99A

You can add the 'AffectsParent' tag to the warheads to make the madtank not affect itself.

Example: https://github.com/OpenRA/OpenRA/blob/55d074b25d1aed705ceda27b326eb8920c72e140/mods/ts/weapons/energyweapons.yaml#L18

r34ch:
@Truth99A Well I have my own thoughts on balance of this unit. I wasn't going to say anything since I really want to see how the suggestions in the other thread  #10437 play out first, but I'll throw this out there anyway since people are talking about it.

I would like the MAD thump/quake to out-range all artillery units, have a verrry low DPS and no detonation. The MAD tank could have very high HP with the ability to undeploy, while taking its own quake damage fairly slowly. That would give it a very unique role in that it would be an almost continual but slow source of low damage. This would force infantry and air counters but allow the time to prepare for that. 

Not only would it be a real area of denial weapon, but you could use a group of them to tank damage from defences rather than lay siege from a distance. If you did choose to deploy them, to increase the very weak DPS you could deploy a group of them - but then they would damage themselves much faster - so it would be a balancing act. Do you use a single one over a prolonged period of time to harass an enemy and force their hand or do you just deploy a bunch an cause higher damage over much less time?

But in all honesty I would like it if the Allies got the (buffed) MAD and the Soviet mammoth a got a buff to make it viable again. I'd rather see both units have _very_ high HP and be dedicated damage soakers with extremely poor DPS. I think @cjshmyr's idea of buffing the med/heavy hp slightly is definitely needed, but it would only further relegate the currents mammoths role.

At the moment, apart from support weapons and fragile V2s/artys, we do not have any units that properly tank. And I mean tank. Mammoths literally get shredded by base defences and RSs without an IC in seconds.

It's a fairly controversial idea for the purists, but it could open up some interesting strategies on chokey maps, esp in combination with slow and fragile infantry as well as air units.

Truth99A:
@Mailaender 
tried AffectsParent: false
no luck

Mailaender:
Needs a rebase and squashing of the commits which don't pass the CI. If you need assistance, ask in IRC for help.

pchote:
The motivation to rebalance the MAD tank seems to have died, but it seems to me that the way forward will involve more than just yaml changes.  I'll close this for now to help tidy our PR queue.

Issue #10426, opened 2016-01-05
Balance : Nerfing Tesla Tank ? 


ShadowA:
Hello :)

Yesterday I've been testing the new Tesla Tank 
(http://s000.tinyupload.com/index.php?file_id=67625660725952331414)
- watching SoScared stream 
  (http://www.twitch.tv/soscared/v/33626018 ,  usage 51st min)

While Tesla tanks being now really fun and a good hit'n'run unit, it might have become a bit too strong in the this new release, because it has all the advantages : 
- High speed
- Good range
- High damage
- Quite good armor 
  (against tanks for instance, although 1 full Yak / Hind can destroy it
  EDIT : after second try, it's super fragile against all gun types ... but still)
- Crushing ability
- Turreted attack

It's almost like a "highspeed-hybrid-long-range-tank-unit" :) which could cause some balance issues.

Only disadvantages IMO :
- Anti-air vulnerability (but mixing with anti-air units (flaks especially) reduces this disadvantage)
- Fragile against gun-type (air, pillboxes, ...)
- Price / construction time maybe ? Not so high though.

Maybe one of these traits could be nerfed ? Not sure which one though 
(Speed ? As it is a tank. But that would reduce the hit'n'run ability, which is a nice trait of this unit.
Long range ? Why should they have bigger range than other tanks ?
Armor ?)

Thoughts ? :)
##

Townes27:
You forgot one thing. They can be curtained. If shockies were OP, then this is über-UP.

r34ch:
:+1: I agree they need a nerf. I would propose they get a substantial health nerf and to loose their crushing ability. They are far too tanky right now and the crushing ability is silly considering their speed and the IC support power. 

That would keep them as light support units and allow air and infantry to be far more effective against them.

klaas1:
I think we should watch out calling higher tier units overpowered. I mean, they should be better or more versatile than the heavy tank/ medium tank, because otherwise they are obsolete and nobody will ever build them.

That said, it is a bit easy to crush infantry ATM with testla tanks, and even more so with chrono tanks, balance wise this could be an issue.

ShadowA:
Yes you're right @klaas1 
They're powerful but still counterable and got their weak points, so I'd not say OP

But maybe a slight tweak somehow of one the traits could be considered

MurtoTheRay:
[...]

PR #10429, opened 2016-01-05
Remove RA Tesla Tanks ability to crush infantry and walls

ghost:
In relation to #10426

This PR removes the TT's ability to crush infantry and walls. Its speed combined with the IC makes this too effective against massed infantry.

### DIFF
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -556,14 +556,14 @@ TTNK:
 		Cost: 1350
 	Tooltip:
 		Name: Tesla Tank
-		Description: Tank with mounted tesla coil.\n  Strong vs Infantry, Vehicles, Buildings\n  Weak vs Aircraft
+		Description: Tank with mounted tesla coil.\n  Strong vs Vehicles, Buildings\n  Weak vs Aircraft, Infantry
 	Health:
 		HP: 400
 	Armor:
 		Type: Light
 	Mobile:
 		Speed: 113
-		Crushes: wall, mine, crate, infantry
+		Crushes: mine, crate
 	RevealsShroud:
 		Range: 7c0
 	Armament:


pchote:
The tesla tank is one of the biggest in the game, I think it would be counterintuitive for it to not crush at least infantry.

r34ch:
Actually if you ever see the renders of it it is a half track with pathetically small wheels (which is why I kept it as ^tank and not ^vehicle). And the original I don't think crushed infantry.

http://vignette3.wikia.nocookie.net/cnc/images/8/8a/RAAF_Tesla_Tank_SG_Render.png/revision/latest?cb=20091221210129

From a balance perspective it just makes infantry more viable against them. Otherwise with IC, flaks and TTs you can overcome any obstacle, which was the issue with IC, flaks and shockies last patch.

ScottNZ:
Infantry aren't supposed to be viable against them though.

kyrylo:
@r34ch On that render I can see a minigun. I guess we shouldn't be fooled by renders.

The only counter to mass tesla tanks is air units. If you mix in some flak trucks, then there are no counters. What if we made the flak truck available only for Ukraine? It means that tesla tanks would be an easy target without protection. The only viable mobile protection would be rocket soldiers, and they're slow (so get an APC). So you can go for a ride with them, but it would be pretty risky (currently it's not risky because flak trucks are as fast).

I'm not a fan of nerfing tesla tank at the moment simply because it's a high tech unit. You have to spend a lot of time and money to be able to build it. I wouldn't want to have a cheap Chinese zap-zap toy.

In terms of crushing abilities it's similar to the light tank from TD.

mods/ra/rules/vehicles.yaml
```diff
@@ -556,14 +556,14 @@ TTNK:
 		Cost: 1350
 	Tooltip:
 		Name: Tesla Tank
-		Description: Tank with mounted tesla coil.\n  Strong vs Infantry, Vehicles, Buildings\n  Weak vs Aircraft
+		Description: Tank with mounted tesla coil.\n  Strong vs Vehicles, Buildings\n  Weak vs Aircraft, Infantry
```
Mailaender: It instantly zaps infantry. I wouldn't consider that weak.

r34ch:
Closing in favour of #10437

PR #10430, opened 2016-01-05
Give gap tech exclusively to France

ghost:
Although France's fake buildings got a buff, overall it is still pretty weak in comparison to the other factions. This PR proposes that France should exclusively build both the gap gen and gap truck.

### DIFF
--- a/mods/ra/rules/structures.yaml
+++ b/mods/ra/rules/structures.yaml
@@ -60,7 +60,7 @@ GAP:
 	Buildable:
 		Queue: Defense
 		BuildPaletteOrder: 110
-		Prerequisites: atek, ~structures.allies, ~techlevel.unrestricted
+		Prerequisites: atek, ~structures.france, ~techlevel.unrestricted
 	Building:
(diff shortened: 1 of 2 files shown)


kyrylo:
I like this. :)

Truth99A:
makes sense

MustaphaTR:
Sorry but i want to give a :-1: to this. All factions currently have 2 special things aren't they? Giving 3 to France doesn't sound good to me. Buffing the Gap Truck sounds better to me. Maybe by Gap Range, Sight, HP or speed.

ScottNZ:
It would be best if we could make it actually work as it did in the original, which would actually remove an enemy actor's ability to reveal shroud more than one cell away while it was inside the area of affect (and create shroud as it currently does).

pchote:
That should in theory be simple.  Just reverse the orders that we resolve "cells that actors reveal" and "cells that actors hide" so that the generated shroud takes precedence.

r34ch:
Closing in favour of #10437

PR #10328, opened 2015-12-29
Buff light tank in the RA mod

ghost:
Closes #10250

PR #10432, opened 2016-01-06
changed the light tank's cost from $700 to $600

Ethal-A:
this was done because the medium tank (far better than the light) costs only $150 more than the light tank which basicly gives people no reason to build the light tank after they get out a service depo

at the same time the hum-v costs $500 so you wouldn't want people to have a better alternative
but then again the hum-v counters infantry so it has its own purpose.

### DIFF
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -38,7 +38,7 @@ V2RL:
 		BuildPaletteOrder: 50
 		Prerequisites: ~vehicles.allies, ~techlevel.low
 	Valued:
-		Cost: 700
+		Cost: 600
 	Tooltip:
 		Name: Light Tank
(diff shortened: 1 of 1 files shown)


pchote:
I'm concerned that all these individual ("pot-shot") balance prs are going to hurt more than they help.  Balance is about more than just individual a vs b stats – everything needs to be considered at the same time.   It feels to me that this is being overlooked with the current collection of parallel / independent balance prs.  I don't mean to pick on you / this ticket specifically, but I feel that something needs to be said.

Mailaender:
> gives people no reason to build the light tank after they get out a service depo

It's speed and high DPS already make it a good choice in late game for rush tactics, but indeed that is the reason to tech up: you get better stuff. Sorry, but your reasoning isn't sound. :-1:

cjshmyr:
#10328 also just happened, I'd like to see how that goes

MustaphaTR:
Agreed with @Mailaender . So :-1: from me too.

MicroJOo:
:-1:

Issue #9810, opened 2015-10-30
Replace MiniYaml Ticks with Delay/Duration


atlimit8:
`Delay` has become the standard way of denoting an interval in ticks. There are 10 fields that use `Ticks`, 42 that use `Delay`, and 2 more that use `DelayInTicks`:

```
OpenRA.Game/Traits/DrawLineToTarget.cs:19:              public readonly int Ticks = 60;
OpenRA.Mods.Common/Traits/AmmoPool.cs:45:               public readonly int ReloadTicks = 25 * 2;
OpenRA.Mods.Common/Traits/AmmoPool.cs:51:               public readonly int SelfReloadTicks = 25 * 2;
OpenRA.Mods.Common/Traits/Harvester.cs:35:              public readonly int LoadTicksPerBale = 4;
OpenRA.Mods.Common/Traits/Harvester.cs:38:              public readonly int UnloadTicksPerBale = 4;
OpenRA.Mods.Common/Traits/Player/StrategicVictoryConditions.cs:25:              public readonly int TicksToHold = 25 * 60 * 5;
OpenRA.Mods.Common/Traits/Render/WithInfantryBody.cs:21:                public readonly int MinIdleWaitTicks = 30;
OpenRA.Mods.Common/Traits/Render/WithInfantryBody.cs:22:                public readonly int MaxIdleWaitTicks = 110;
OpenRA.Mods.Common/Traits/SelfHealing.cs:21:            public readonly int Ticks = 5;
OpenRA.Mods.Common/Traits/Wanders.cs:22:                public readonly int TicksToWaitBeforeReducingMoveRadius = 5;
```

```
OpenRA.Game/GameRules/WeaponInfo.cs:43:         public readonly int ReloadDelay = 1;
OpenRA.Game/GameRules/WeaponInfo.cs:57:         public readonly int BurstDelay = 5;
OpenRA.Mods.Cnc/Traits/SupportPowers/IonCannonPower.cs:27:              public readonly int CameraRemoveDelay = 25;
OpenRA.Mods.Cnc/Traits/SupportPowers/IonCannonPower.cs:39:              public readonly int WeaponDelay = 7;
OpenRA.Mods.Common/AI/HackyAI.cs:51:            public readonly int MinimumAttackForceDelay = 0;
OpenRA.Mods.Common/AI/HackyAI.cs:61:            public readonly int StructureProductionInactiveDelay = 125;
[...]

PR #10444, opened 2016-01-07
Replaced Ticks with Delay/Duration

colonelpopcorn:
Per issue #9810: refactored ticks to delay in all the specified files pointed out by @atlimit8.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -75,6 +75,7 @@ Also thanks to:
     * Jeff Harris (jeff_1amstudios)
     * Jes
     * Joakim Lindberg (booom3)
+    * Jonathan Ling (colonelpopcorn)
     * Joppy Furr
     * Kanar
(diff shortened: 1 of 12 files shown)


GraionDilach:
You have to write your own upgrade rule, not modifying an existing one.

penev92:
Thanks for the refactor.
These renamings will need upgrade rules, which you may want to see us on IRC about.

colonelpopcorn:
Will do, I'll see about writing an upgrade rule tomorrow.

colonelpopcorn:
Is there any documentation on upgrade rules? It looks look it's a program built to output a proper .yaml file for maps and mods based on version #, but I can't make heads or tales of it. In lieu of documentation could I get an output .yaml file?

pchote:
Superseded by #10711.

PR #10457, opened 2016-01-07
Don't update player list when lobby doesn't have our client yet

### LINKED ISSUE #10450: Rare Crash when entering Lobby
This crash can only happen if you click Skirmish button fast enough. I wish I could say more, but that's how I get this crash. You can reliably recreate this if you can click on the Skirmish button fast enough. (I'd say about 4 or 5 times a second)

Here is a paste of the crash log: http://pastebin.com/6tAKpdaK

Safe to say that someone with patience will never encounter this issue, but if you're like me and like to spam the mouse button it'll happen from time to time.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29665219-rare-crash-when-entering-lobby?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
Fixes #10450.

I was actually only able to reproduce this once (on the first try, too), never again after that. But with the crash happening on [this line](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs#L895), it's pretty clear what's going on. Since quite a few more things are referencing that, I decided to move the check up to the top of the method. It's probably fair to assume that both the loops running before the check in line 895 would have iterated over empty collections anyway, so they wouldn't have done anything useful, so we might just as well bail out early.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
@@ -776,6 +776,9 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 		void UpdatePlayerList()
 		{
+			if (orderManager.LocalClient == null)
+				return;
+
 			var idx = 0;
 			foreach (var kv in orderManager.LobbyInfo.Slots)
 			{


abcdefg30:
Looks good to me. Unfortunately I wasn't able to reproduce the crash in the first place. :+1:

Mailaender:
I don't think anything could go wrong here. Thanks for the fix of this heisenbug.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3869468b2506f4768b5fa496845a739d844448ff...0527433c47fc3a66f7eed00ded18a1d290c2fbce)

PR #10462, opened 2016-01-09
Prevent items without size from being added to SpatiallyPartitioned.

RoosterDragon:
Items with no size act unexpectedly as they can fail to be returned when querying partition bins as they do not intersect along the top or left edges of the bin. We prevent such items being added in the first place to avoid this scenario.

Example:

``` c#
new Rectangle(0, 0, 0, 0).IntersectsWith(new Rectangle(0, 0, 2, 2));
false

new Rectangle(0, 0, 1, 1).IntersectsWith(new Rectangle(0, 0, 2, 2));
true
```

As a side effect - we must now prevent any Immobile items that do not have size from being added to the screen map.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Immobile.cs
+++ b/OpenRA.Mods.Common/Traits/Immobile.cs
@@ -55,14 +55,18 @@ namespace OpenRA.Mods.Common.Traits
 		{
 			self.World.ActorMap.AddInfluence(self, this);
 			self.World.ActorMap.AddPosition(self, this);
-			self.World.ScreenMap.Add(self);
+
+			if (!self.Bounds.Size.IsEmpty)
+				self.World.ScreenMap.Add(self);
 		}
 
 		public void RemovedFromWorld(Actor self)
 		{
 			self.World.ActorMap.RemoveInfluence(self, this);
 			self.World.ActorMap.RemovePosition(self, this);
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Traits/Immobile.cs
```diff
@@ -53,6 +53,9 @@ public Immobile(ActorInitializer init, ImmobileInfo info)
 
 		public void AddedToWorld(Actor self)
 		{
+			if (occupied == IOccupySpaceExts.NoOccupiedCells)
+				return;
+
 			self.World.ActorMap.AddInfluence(self, this);
 			self.World.ActorMap.AddPosition(self, this);
 			self.World.ScreenMap.Add(self);
```
pchote: `ScreenMap.Add` uses the actor bounds rather than the footprint.  The size of the bounds is completely disconnected from the footprint defined in this trait.

RoosterDragon: Changed.

abcdefg30:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/8300874c4c615da49dbaa39145bb4c00a7c61643)

PR #10468, opened 2016-01-09
Fix shroud not updating when disabled

RoosterDragon:
If you disable visibility checks using the debug options, the affected areas of the shroud are not refreshed.

This can be most easily seen during a replay if you spectate a player who used this option, as your view of the shroud remains outdated until units move through it.

Suggested changelog: Fix shroud not updating when "Disable Visibility Checks" is used.

### DIFF
--- a/OpenRA.Game/Traits/World/Shroud.cs
+++ b/OpenRA.Game/Traits/World/Shroud.cs
@@ -22,8 +22,6 @@ namespace OpenRA.Traits
 
 	public class Shroud : ISync
 	{
-		[Sync] public bool Disabled;
-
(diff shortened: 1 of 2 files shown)


Mailaender:
Note to reviewers: you can't reproduce this with `/all` only `/visibility`. This indeed fixes it. :white_check_mark:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7b0aba18fe3172425fa449fab526dbc32cd1ab24)

Issue #10387, opened 2016-01-02
D2k map importer has wrong actor names for Deviator and Devastator


penev92:
Those were apparently forgotten during the recent mod overhaul.

PR #10471, opened 2016-01-10
Fixed the Dune 2000 legacy map import

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/10387 and also adds a lint test so it won't happen again. Spotted another one with it: `palacec`.

### DIFF
--- a/OpenRA.Mods.D2k/Lint/CheckImportActors.cs
+++ b/OpenRA.Mods.D2k/Lint/CheckImportActors.cs
@@ -1,0 +1,28 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using OpenRA.Mods.D2k.UtilityCommands;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.D2k.Lint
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.D2k/Lint/CheckImportActors.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace OpenRA.Mods.D2k.Lint
+{
+	public class CheckImportActors : ILintRulesPass
+	{
+		public void Run(Action<string> emitError, Action<string> emitWarning, Ruleset rules)
+		{
+			foreach (var actorData in D2kMapImporter.ActorDataByActorCode.Values)
+			{
+				if (!rules.Actors.Keys.Contains(actorData.First))
```
RoosterDragon: `Keys.Contains` -> `ContainsKey`

RoosterDragon:
:+1:

OpenRA.Mods.D2k/Lint/CheckImportActors.cs
Mailaender: I assume this is faster?

RoosterDragon: Yes - the first is doing the lookup using the dictionary, the second requires a linear search.

penev92:
Thanks :+1:

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/b4384aa5d000b87650fb1bdec2b14bc64a987ad1)

PR #10478, opened 2016-01-10
Sync replay support timers to original game time.

### LINKED ISSUE #10466: Watching replays on different speeds affects superweapon timers
On release, RA mod.

If you toggle a replay between the different speeds the timer in the top corner is affected.

timer
Here it says 0:19 seconds remaining when the actual time is closer to 7 minutes when viewing a replay on max speed.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29734547-watching-replays-on-different-speeds-affects-superweapon-timers?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #10466.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ObserverSupportPowerIconsWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ObserverSupportPowerIconsWidget.cs
@@ -20,11 +20,11 @@ namespace OpenRA.Mods.Common.Widgets
 {
 	public class ObserverSupportPowerIconsWidget : Widget
 	{
-		public Func<Player> GetPlayer;
-		Animation icon;
-		World world;
-		WorldRenderer worldRenderer;
-		Dictionary<string, Animation> clocks;
+		readonly Animation icon;
+		readonly World world;
+		readonly WorldRenderer worldRenderer;
+		readonly Dictionary<string, Animation> clocks;
+		readonly int timestep;
 
 		public int IconWidth = 32;
 		public int IconHeight = 24;
@@ -33,6 +33,7 @@ namespace OpenRA.Mods.Common.Widgets
 		public string ClockAnimation = "clock";
 		public string ClockSequence = "idle";
 		public string ClockPalette = "chrome";
+		public Func<Player> GetPlayer;
 
 		[ObjectCreator.UseCtor]
 		public ObserverSupportPowerIconsWidget(World world, WorldRenderer worldRenderer)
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Widgets/ObserverSupportPowerIconsWidget.cs
```diff
@@ -41,6 +42,14 @@ public ObserverSupportPowerIconsWidget(World world, WorldRenderer worldRenderer)
 			this.worldRenderer = worldRenderer;
 			clocks = new Dictionary<string, Animation>();
 			icon = new Animation(world, "icon");
+
+			// Timers should be synced to the effective game time, not the playback time.
+			GameSpeed speed;
+			var gameSpeeds = Game.ModData.Manifest.Get<GameSpeeds>();
```
Phrohdoh: Can we `GetOrDefault` so we don't force gamespeeds?
Or are we going to force a single gamespeed at minimum?

Mailaender: This is already enforced everywhere else. You can't define a mod without it. https://github.com/OpenRA/OpenRA/pull/9255

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -23,25 +23,34 @@ public class SupportPowerTimerWidget : Widget
...
+
+			// Timers in replays should be synced to the effective game time, not the playback time.
+			timestep = world.Timestep;
+			if (world.IsReplay)
+			{
+				GameSpeed speed;
+				var gameSpeeds = Game.ModData.Manifest.Get<GameSpeeds>();
+				if (gameSpeeds.Speeds.TryGetValue(world.LobbyInfo.GlobalSettings.GameSpeedType, out speed))
+					timestep = speed.Timestep;
+			}
```
Mailaender: How about moving a `public static int GetTimestep(World world)` method in WidgetUtils or a similar collection to avoid copy pasta?

pchote: These cases are subtly (and importantly) different, so there is no meaningful amount of code that can be reused.

Mailaender:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f5c1b12e6cbf8cf7eb768404c12a947257fef0d8)

PR #10482, opened 2016-01-11
Fixed crashes with Red Alert 2 wav files

Mailaender:
The recent changes made the https://github.com/OpenRA/ra2 mod unplayable. Even clicking a button crashes the whole game. This reverts parts of https://github.com/OpenRA/OpenRA/commit/5063e56786c7aa47dda640df6a8acf3d6a941c3a to aid debugging. I stumbled upon some `if (Format != "WAVE")` cases with `Format = "fmt"` so this rejected valid *.wav files without reason. ~~It won't fix it completely though. Something else is still bogus.~~ The generic exceptions "not a valid sound file" are far from useful so I put it into the appropriate log files.

### DIFF
--- a/OpenRA.Game/FileFormats/AudLoader.cs
+++ b/OpenRA.Game/FileFormats/AudLoader.cs
@@ -170,6 +170,7 @@ namespace OpenRA.FileFormats
 			out int sampleRate)
 		{
 			channels = sampleBits = sampleRate = 0;
+			var position = stream.Position;
 
 			try
 			{
@@ -186,6 +187,10 @@ namespace OpenRA.FileFormats
 				Log.Write("sound", e.ToString());
 				rawData = null;
 				return false;
+			}
+			finally
+			{
+				stream.Position = position;
 			}
 
 			channels = 1;
--- a/OpenRA.Game/FileFormats/AudLoader.cs
+++ b/OpenRA.Game/FileFormats/AudLoader.cs
@@ -182,8 +182,8 @@ namespace OpenRA.FileFormats
 				// If not, it will simply return false so we know we can't use it. If it is, it will start
 				// parsing the data without any further failsafes, which means that it will crash on corrupted files
 				// (that end prematurely or otherwise don't conform to the specifications despite the headers being OK).
-				Log.Write("debug", "Failed to parse AUD file {0}. Error message:".F(fileName));
-				Log.Write("debug", e.ToString());
+				Log.Write("sound", "Failed to parse AUD file {0}. Error message:".F(fileName));
+				Log.Write("sound", e.ToString());
 				rawData = null;
 				return false;
 			}
--- a/OpenRA.Game/FileFormats/AudLoader.cs
+++ b/OpenRA.Game/FileFormats/AudLoader.cs
@@ -125,12 +125,12 @@ namespace OpenRA.FileFormats
 			sampleRate = s.ReadUInt16();
 			var dataSize = s.ReadInt32();
 			var outputSize = s.ReadInt32();
-			var readFlag = s.ReadByte();
-			var readFormat = s.ReadByte();
 
+			var readFlag = s.ReadByte();
 			if (!Enum.IsDefined(typeof(SoundFlags), readFlag))
 				return false;
 
+			var readFormat = s.ReadByte();
 			if (!Enum.IsDefined(typeof(SoundFormat), readFormat))
 				return false;
 
--- a/OpenRA.Game/FileFormats/VocLoader.cs
+++ b/OpenRA.Game/FileFormats/VocLoader.cs
@@ -19,6 +19,8 @@ namespace OpenRA.FileFormats
 	{
 		bool ISoundLoader.TryParseSound(Stream stream, string fileName, out byte[] rawData, out int channels, out int sampleBits, out int sampleRate)
 		{
+			var position = stream.Position;
+
 			try
 			{
 				var vocStream = new VocStream(stream);
@@ -32,6 +34,10 @@ namespace OpenRA.FileFormats
 				rawData = null;
 				channels = sampleBits = sampleRate = 0;
 				return false;
+			}
+			finally
+			{
+				stream.Position = position;
 			}
 
 			return true;
--- a/OpenRA.Game/FileFormats/WavLoader.cs
+++ b/OpenRA.Game/FileFormats/WavLoader.cs
@@ -181,6 +181,7 @@ namespace OpenRA.FileFormats
 		{
 			rawData = null;
 			channels = sampleBits = sampleRate = 0;
+			var position = stream.Position;
 
 			try
 			{
@@ -196,6 +197,10 @@ namespace OpenRA.FileFormats
 				Log.Write("sound", "Failed to parse WAV file {0}. Error message:".F(fileName));
 				Log.Write("sound", e.ToString());
 				return false;
+			}
+			finally
+			{
+				stream.Position = position;
 			}
 
 			rawData = RawOutput;
--- a/OpenRA.Game/FileFormats/WavLoader.cs
+++ b/OpenRA.Game/FileFormats/WavLoader.cs
@@ -68,24 +68,17 @@ namespace OpenRA.FileFormats
 						s.ReadBytes(FmtChunkSize - 16);
 						break;
 					case "fact":
-						{
-							var chunkSize = s.ReadInt32();
-							UncompressedSize = s.ReadInt32();
-							s.ReadBytes(chunkSize - 4);
-						}
-
+						var chunkSize = s.ReadInt32();
+						UncompressedSize = s.ReadInt32();
+						s.ReadBytes(chunkSize - 4);
 						break;
 					case "data":
 						DataSize = s.ReadInt32();
 						RawOutput = s.ReadBytes(DataSize);
 						break;
 					default:
-						// Ignore unknown chunks
-						{
-							var chunkSize = s.ReadInt32();
-							s.ReadBytes(chunkSize);
-						}
-
+						var unknownChunkSize = s.ReadInt32();
+						s.ReadBytes(unknownChunkSize);
 						break;
 				}
 			}
--- a/OpenRA.Game/FileFormats/WavLoader.cs
+++ b/OpenRA.Game/FileFormats/WavLoader.cs
@@ -201,8 +201,8 @@ namespace OpenRA.FileFormats
 				// If not, it will simply return false so we know we can't use it. If it is, it will start
 				// parsing the data without any further failsafes, which means that it will crash on corrupted files
 				// (that end prematurely or otherwise don't conform to the specifications despite the headers being OK).
-				Log.Write("debug", "Failed to parse WAV file {0}. Error message:".F(fileName));
-				Log.Write("debug", e.ToString());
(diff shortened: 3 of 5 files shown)


OpenRA.Game/FileFormats/WavLoader.cs
```diff
@@ -35,23 +35,20 @@ public enum WaveType { Pcm = 0x1, ImaAdpcm = 0x11 }
 
 		bool LoadSound(Stream s)
 		{
-			var type = s.ReadASCII(4);
```
pchote: Please augment, don't just remove.  Removing these checks means that the parser will unconditionally treat the file as a wave, and treat bytes 4..8 as a byte count which is then read.  This will then probably crash, as this bogus size can easily be larger than the actual file size.

Mailaender: Why is the auto-detection using the same function as the loading? No wonder this is broken.

pchote: Using the same function avoids having to parse the file twice.  These `return false`s indicate that it is not a wave file and that the public `TryParseSound` should also return false.

Mailaender: Oh, I think I got it quite wrong on how this works. Will drop my changes for now.

OpenRA.Game/Sound/Sound.cs
```diff
@@ -69,7 +69,8 @@ ISoundSource LoadSound(string filename)
 						return soundEngine.AddSoundSourceFromMemory(rawData, channels, sampleBits, sampleRate);
 				}
 
-				throw new InvalidDataException(filename + " is not a valid sound file!");
```
pchote: Crashing here was a deliberate design choice.

Mailaender: Apparently not in https://github.com/OpenRA/OpenRA/pull/10482/files#diff-798c8d09024b0e9c0ff1bdb93edd2abcR56 above. It is stupid to constantly crash the RA2 mod for weeks now, just because button click sounds are not parsed correctly.

pchote: That doesn't contradict what I said above...?

penev92: The design pattern with all our `*Loader`s is the same - sprites, sound, and soon packages - throw on invalid files (yes, the message should be something like "Could not find a suitable SoundLoader for file X").

Also
>[23:01] <+penev> ]later Mailaender I guess the RA2 wavs just need a slight tweak in the "is this a real wav file" checks

pchote: I explained in IRC earlier today what the problem was and how to fix it, if somebody else wants to take a stab at that.

Mailaender: Can you give the timestamp, because I can't really find it.

Mailaender:
Finally fixed it. We had offset problems that manifested when first parsing a WAV file in a BAG container as AUD, then as WAV. Also pushed some surrounding cleanups. Note to reviewers: the RA2 mod is currently broken. You need https://github.com/OpenRA/ra2/pull/28 to test this.

Mailaender:
CC @abcdefg30

pchote:
As mentioned in IRC there are two fundamental problems here, and while this may hide the crash for RA2 it does not solve either problem:

1. The `TryParseSound` implementations have a fairly major side-effect of modifying the stream position, which is then hidden by Sound resetting the stream position to 0 in the loop that calls them.  This is conceptually bogus, and will break as soon as we implement any performance optimizations that use views into a larger stream.  This can be easily fixed by storing the stream position at the start of `TryParseSound` and then resetting that position before returning.

2. `MergedStream` does not implement `Position.set`.  We use `Position.set` through all of our file parsing code, and so hacking Position &rarr; Seek in the one place that covers up your specific crash is just horrible.  This can be easily fixed by implementing `Position.set` to call `Seek` to set the proper positions in the underlying streams.

Mailaender:
I tried my best to solve those problems.

OpenRA.Game/FileFormats/AudLoader.cs
```diff
@@ -170,11 +170,15 @@ public static bool LoadSound(Stream s, out byte[] rawData, out int sampleRate)
 			out int sampleRate)
 		{
 			channels = sampleBits = sampleRate = 0;
+			var position = stream.Position;
 
 			try
 			{
 				if (!LoadSound(stream, out rawData, out sampleRate))
+				{
+					stream.Seek(position, SeekOrigin.Begin);
```
pchote: Please use `stream.Position = position` instead of `stream.Seek` everywhere for consistency with our other file parsers.

OpenRA.Game/FileFormats/VocLoader.cs
```diff
@@ -31,6 +33,7 @@ bool ISoundLoader.TryParseSound(Stream stream, string fileName, out byte[] rawDa
 			{
 				rawData = null;
 				channels = sampleBits = sampleRate = 0;
+				stream.Seek(position, SeekOrigin.Begin);
 				return false;
```
pchote: You can use a `finally` block to unconditionally reset the stream position when returning.

OpenRA.Game/Primitives/MergedStream.cs
```diff
@@ -35,7 +36,7 @@ public override void Flush()
 
 		public override long Seek(long offset, SeekOrigin origin)
 		{
-			var position = Position;
+			position = Position;
```
pchote: This line will be redundant now.

OpenRA.Game/Sound/Sound.cs
```diff
@@ -62,12 +62,10 @@ ISoundSource LoadSound(string filename)
 				int channels;
 				int sampleBits;
 				int sampleRate;
+				stream.Seek(0, SeekOrigin.Begin);
```
pchote: This line looks redundant and can be removed.

Mailaender:
Updated.

OpenRA.Game/FileFormats/WavLoader.cs
```diff
@@ -201,10 +196,15 @@ public byte[] DecodeImaAdpcmData()
 				// If not, it will simply return false so we know we can't use it. If it is, it will start
 				// parsing the data without any further failsafes, which means that it will crash on corrupted files
 				// (that end prematurely or otherwise don't conform to the specifications despite the headers being OK).
-				Log.Write("debug", "Failed to parse WAV file {0}. Error message:".F(fileName));
-				Log.Write("debug", e.ToString());
+				Log.Write("sound", "Failed to parse WAV file {0}. Error message:".F(fileName));
+				Log.Write("sound", e.ToString());
+				stream.Seek(position, SeekOrigin.Begin);
```
pchote: This line is now unnecessary and should be removed.

OpenRA.Game/FileFormats/WavLoader.cs
```diff
@@ -33,32 +33,33 @@ public class WavLoader : ISoundLoader
 		public enum WaveType { Pcm = 0x1, ImaAdpcm = 0x11 }
 		public static WaveType Type { get; private set; }
 
+		bool wavDetected;
+
 		bool LoadSound(Stream s)
 		{
-			var type = s.ReadASCII(4);
```
pchote: This reintroduces the original problem:
> Please augment, don't just remove. Removing these checks means that the parser will unconditionally treat the file as a wave, and treat bytes 4..8 as a byte count which is then read. This will then probably crash, as this bogus size can easily be larger than the actual file size.

Sure the crash gets caught, but what if it doesn't crash?  You may still accidentally detect an aud or other audio file as a wav full of bogus chunks that are ignored then plays nothing.

The wav file format dictates that the file starts with `RIFF` followed by `WAVE`, so that is what we check to determine whether it is a wav.

Mailaender: Nope. In this case I am ahead already. You can test this yourself by reordering to `Aud, Wav` at https://github.com/OpenRA/ra2/pull/28/files#diff-1ea1b12ca5110238b7484e0a950b8f62R234 to make sure detection works in both cases. The `bool wavDetected` stays false if it isn't a WAVE file although I don't stop at byte 4, but force it to parse the whole stream.

OpenRA.Game/FileFormats/WavLoader.cs
Mailaender: Fixed. Sorry for the sloppiness.

OpenRA.Game/FileFormats/WavLoader.cs
pchote: "It works" is not an excuse for code design complaints.

Mailaender: I looked it up. http://cgit.freedesktop.org/xdg/shared-mime-info/tree/freedesktop.org.xml.in#n4139 is file ending *.wav and just looking for WAVE or WAV.

pchote: Yes, just like this block of code used to do before it was removed.

Mailaender: Removed the change again.

penev92:
Please also fix the commit messages. Half of them don't mean anything without the context of this PR, and at least one doesn't even *with* the context.

abcdefg30:
Looks good to me. :+1:

OpenRA.Game/FileFormats/WavLoader.cs
```diff
@@ -57,8 +56,8 @@ bool LoadSound(Stream s)
 						AudioFormat = s.ReadInt16();
 						Type = (WaveType)AudioFormat;
 
-						if (Type != WaveType.Pcm && Type != WaveType.ImaAdpcm)
-							throw new NotSupportedException("Compression type is not supported.");
+						if (!Enum.IsDefined(typeof(WaveType), Type))
+							throw new NotSupportedException("Compression type {0} is not supported.".F(AudioFormat));
```
pchote: `AudioFormat` is *not* the compression type, so this message does not make sense.

pchote:
Code looks fine by me except for the one issue above and the earlier comments about using sensible commit messages.

Mailaender:
Beautified the commit messages.

OpenRA.Game/FileFormats/WavLoader.cs
```diff
@@ -69,24 +68,17 @@ bool LoadSound(Stream s)
...
 						break;
 					default:
-						// Ignore unknown chunks
-						{
-							var chunkSize = s.ReadInt32();
-							s.ReadBytes(chunkSize);
-						}
-
+						var unknownChunkSize = s.ReadInt32();
+						s.ReadBytes(unknownChunkSize);
```
Phrohdoh: You can inline this since you are not doing any checking on the int.

Mailaender: I use it to document what this is about. See the comment removal a few lines before.

pchote:
Thanks for the fixes :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/762f8eca37c4ff1d45e4c2cb3c21e112d6fc13ea

Mailaender:
Thanks for your help as well!

PR #10486, opened 2016-01-11
Hardcode mod logo and preview image file names

penev92:
Fixes a oversight on my part. I had said I'd remove those, but forgot.
See https://github.com/OpenRA/OpenRA/pull/10232#commitcomment-15368566 (some of the discussion is probably on IRC though :/ )

### DIFF
--- a/OpenRA.Game/ModMetadata.cs
+++ b/OpenRA.Game/ModMetadata.cs
@@ -25,8 +25,6 @@ namespace OpenRA
 		public string Description;
 		public string Version;
 		public string Author;
-		public string LogoImagePath;
-		public string PreviewImagePath;
 		public bool Hidden;
 		public ContentInstaller Content;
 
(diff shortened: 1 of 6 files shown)


pchote:
The reasoning behind hardcoding these is that we need to guarantee that they are present (for both the mod chooser and external tools like the resource site or OS shell integration), and not referencing a file from another mod package.  This is the same strategy that we use for mod/map.yaml, map.png, etc.

Phrohdoh:
Code looks good and I agree that some paths absolutely need to be consistent/hardcoded. :+1:

pchote:
Confirmed that the images still work ingame :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9db100230f2db9b4e5105505748bb4366f1d1b80

PR #10487, opened 2016-01-11
Convert mod file/folder paths to use package syntax [Superseded]

penev92:
A follow-up of #10232.

### DIFF
--- a/mods/ra/mod.yaml
+++ b/mods/ra/mod.yaml
@@ -9,14 +9,14 @@ RequiresMods:
 
 Folders:
 	.
-	./mods/ra
-	./mods/ra/bits
-	./mods/ra/bits/desert
-	./mods/ra/uibits
+	ra:
+	ra:bits
+	ra:bits/desert
+	ra:uibits
 	~^Content/ra
 
 MapFolders:
-	./mods/ra/maps@System
+	ra:maps@System
 	~^maps/ra@User
 
 Packages:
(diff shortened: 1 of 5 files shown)


mods/ra/mod.yaml
```diff
@@ -37,113 +37,113 @@ Packages:
...
 Cursors:
-	./mods/ra/cursors.yaml
+	ra:cursors.yaml
 
 Chrome:
-	./mods/ra/chrome.yaml
+	ra:chrome.yaml
 
 Assemblies:
 	./mods/common/OpenRA.Mods.Common.dll
```
pchote: Why doesn't common:OpenRA.Mods.Common.dll work?

penev92: I haven't tried it, but I assumed since that is pretty much an engine thing it doesn't need to be moved around anyway, so I just left it with the full path.

GraionDilach: Still looks odd that everything else has the nnew fiormat excluding that one.

mods/ra/mod.yaml
```diff
@@ -187,7 +187,7 @@ Fonts:
 		Font:./mods/common/FreeSansBold.ttf
 		Size:14
 	Title:
-		Font:./mods/ra/ZoodRangmah.ttf
+		Font:ra:ZoodRangmah.ttf
 		Size:48
 	MediumBold:
 		Font:./mods/common/FreeSansBold.ttf
```
Mailaender: I assume this wants to be
```diff
-Font:./mods/common/FreeSansBold.ttf
+Font:common:FreeSansBold.ttf
```
for consistency?

penev92: Same argument as above.
Common (and `all`) is basically a system (engine) thing. The mods are packages that may or may not be there, depending on whether the user wants them.
So I'm not convinced about either requested fixup.

Does everyone disagree with me?

GraionDilach: I prefer consistency, having a mixed setup would confuse lower quality (iow: dumb) modders.

Mailaender: > Common (and all) is basically a system (engine) thing. The mods are packages that may or may not be there, depending on whether the user wants them.

That applies to d2k, ra and cnc as well.

Mailaender: > I prefer consistency, having a mixed setup would confuse

or it would document the alternative way to set this up so it may stay as a test case and so we don't forget.

Mailaender:
All 3 mods still work. :white_check_mark:

mods/ra/mod.yaml
pchote: I'd also prefer consistency.

obrakmann:
Count me in on the "consistency ftw" team. :+1: after that.

pchote:
Also note that by the time i'm finished with the filesystem rework the two syntaxes will use a different mechanism for loading files (mod syntax will be an explicit package request, and may load from a zip or other package type).

penev92:
"Other package" still includes directory.
Also I think your requests for "consistency" here are based on the flwed premises that this is the same thing, and that we're getting rid of the old syntax.

pchote:
At this point it may be easier to close this pr, and wait for them to be fixed in my filesystem PRs (the next one removes the '.' mount from the default mods as a performance optimization).  We have to resolve a huge rebase conflict otherwise.

penev92:
>[20:58] <+pchote> penev: i'll have a pr superseding #10487 probably before the end of the night

Issue #10415, opened 2016-01-04
Lobby / Global tabs mouseover highlight does not work when the global chat tab is activated in the lobby.


pchote:
##

rob-v:
Is this still open issue? Could you please clarify it ideally with a screenshot? I see the Refresh, Browse, .. buttons highlighted.

kyrreso:
Related #13184

PR #10498, opened 2016-01-12
Add a debug setting to draw the RenderBounds of Widgets.

phrohdoh:
screen shot 2016-01-12 at 1 09 27 pm

screen shot 2016-01-12 at 1 22 48 pm

Useful for debugging position and size of widgets.
Hopefully more to come.

### DIFF
--- a/OpenRA.Game/Widgets/Widget.cs
+++ b/OpenRA.Game/Widgets/Widget.cs
@@ -74,7 +74,18 @@ namespace OpenRA.Widgets
 
 		public static void PrepareRenderables() { Root.PrepareRenderablesOuter(); }
 
-		public static void Draw() { Root.DrawOuter(); }
+		public static void Draw()
+		{
+			Root.DrawOuter();
+
+			if (Game.Settings.Debug.WidgetRenderBoundsLines)
+			{
+				Root.DrawDebugRenderBounds();
+
+				if (MouseOverWidget != null)
+					MouseOverWidget.DrawDebugRenderBounds(true, false);
+			}
+		}
 
 		public static bool HandleInput(MouseInput mi)
 		{
@@ -162,6 +173,10 @@ namespace OpenRA.Widgets
 		public bool Visible = true;
 		public bool IgnoreMouseOver;
 		public bool IgnoreChildMouseOver;
+		public Color DebugBoundsColor = Color.Red;
+		public float DebugBoundsWidth = 1;
+		public Color DebugMouseOverBoundsColor = Color.Green;
+		public float DebugMouseOverBoundsWidth = 2;
 
 		// Calculated internally
 		public Rectangle Bounds;
@@ -185,6 +200,11 @@ namespace OpenRA.Widgets
 			IsVisible = widget.IsVisible;
 			IgnoreChildMouseOver = widget.IgnoreChildMouseOver;
 			IgnoreMouseOver = widget.IgnoreMouseOver;
+
+			DebugBoundsColor = widget.DebugBoundsColor;
+			DebugBoundsWidth = widget.DebugBoundsWidth;
+			DebugMouseOverBoundsColor = widget.DebugMouseOverBoundsColor;
+			DebugMouseOverBoundsWidth = widget.DebugMouseOverBoundsWidth;
 
 			foreach (var child in widget.Children)
 				AddChild(child.Clone());
@@ -418,6 +438,23 @@ namespace OpenRA.Widgets
 		}
 
 		public virtual void Draw() { }
+
+		public void DrawDebugRenderBounds(bool drawEvenIfInvisible = false, bool recurseDownAllChildren = true)
+		{
+			if (drawEvenIfInvisible || IsVisible())
+			{
+				var isMouseOverWidget = this == Ui.MouseOverWidget;
+				Game.Renderer.RgbaColorRenderer.DrawRect(
+					new float2(RenderBounds.Left, RenderBounds.Top),
+					new float2(RenderBounds.Right, RenderBounds.Bottom),
+					isMouseOverWidget ? DebugMouseOverBoundsWidth : DebugBoundsWidth,
+					isMouseOverWidget ? DebugMouseOverBoundsColor : DebugBoundsColor);
+
+				if (recurseDownAllChildren)
+					foreach (var child in Children.OfType<Widget>().Reverse())
+						child.DrawDebugRenderBounds();
+			}
+		}
 
 		public virtual void DrawOuter()
 		{
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -91,6 +91,7 @@ namespace OpenRA
 		public bool SanityCheckUnsyncedCode = false;
 		public int Samples = 25;
 		public bool IgnoreVersionMismatch = false;
+		public bool WidgetRenderBoundsLines = false;
 	}
 
 	public class GraphicSettings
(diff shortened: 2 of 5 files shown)


Phrohdoh:
While fixing Travis I noticed that if a widget overrides `DrawOuter` without calling the base (all the way up the chain) these lines won't be drawn.
I do not know how to guarantee these are drawn if the setting is checked.

Mailaender:
You forgot about the `cnc` mod settings chrome:

```
Loading mod: cnc
Inner Exception of type `System.InvalidOperationException`: Widget ADVANCED_PANEL has no child WIDGETDEBUG_CHECKBOX of type CheckboxWidget
```

Mailaender:
Works as promised :white_check_mark: and is indeed a useful tool :+1: although I expect it will cause UI perfectionists to do even harder reviews now. =)

OpenRA.Game/Widgets/Widget.cs
```diff
@@ -419,11 +421,25 @@ public virtual void PrepareRenderablesOuter()
...
+				DebugBoundsWidth,
+				DebugBoundsColor);
+		}
+
 		public virtual void DrawOuter()
 		{
 			if (IsVisible())
 			{
 				Draw();
+				DrawDebugRenderBounds();
```
pchote: You can save an unnecessary message call overhead in the general case by moving the settings check here.

pchote:
It would be really useful (e.g. for debugging #10415) if this could call `Ui.MouseOverWidget.DrawDebugRenderBounds()` with a different color and/or width after rendering all the other bounds (so that it is rendered on top).  Could you please add this?

Phrohdoh:
@Mailaender Oh, I knew I left something out, CnC it is, will fix.
@pchote Sure, where would you like the other color/width defined?
Currently those properties are per-widget instance.

I also need to add these new properties to the `Widget(Widget other)` ctor.

Mailaender:
> @Mailaender Oh, I knew I left something out, CnC it is, will fix.

No need to ping me when you are not even done yet.

Phrohdoh:
CnC has been fixed and `MouseOverWidget` can be given a special color and width combination.

screen shot 2016-01-13 at 8 12 30 pm

This leaves us with a bit of ugliness when the mouse isn't moved after opening a new window.

OpenRA.Game/Widgets/Widget.cs
```diff
@@ -162,6 +173,10 @@ public abstract class Widget
 		public bool Visible = true;
 		public bool IgnoreMouseOver;
 		public bool IgnoreChildMouseOver;
+		public Color DebugBoundsColor = Color.Red;
+		public float DebugBoundsWidth = 1;
+		public Color DebugMouseOverBoundsColor = Color.Green;
+		public float DebugMouseOverBoundsWidth = 2;
```
pchote: What's the logic behind having this per-widget?  This makes it a real pain to adjust mod-wide.  I'd like to see these at least load their defaults from the chrome metrics, but I unless you have a specific reason for wanting chrome.yaml-level overrides I think ti would be better to *only* load it from the metrics (no per-widget overrides).

Phrohdoh: I'm not sure how to do this, and see this as a change request not a fixup.

pchote: `public Color DebugBoundsColor = ChromeMetrics.Get<Color>("DebugBoundsColor")` etc.

Phrohdoh: I got to this point but realized that I'll be highlighting all children unless I pass two floats and two colors, which is the route we could take.

```diff
diff --git a/OpenRA.Game/Widgets/RootWidget.cs b/OpenRA.Game/Widgets/RootWidget.cs
index dad5b21..13177d3 100644
--- a/OpenRA.Game/Widgets/RootWidget.cs
+++ b/OpenRA.Game/Widgets/RootWidget.cs
@@ -8,15 +8,37 @@
  */
 #endregion
 
+using System.Drawing;
 using OpenRA.Graphics;
 
 namespace OpenRA.Widgets
 {
 	public class RootWidget : ContainerWidget
 	{
+		public Color DebugBoundsColor = Color.Red;
+		public float DebugBoundsWidth = 1;
+		public Color DebugMouseOverBoundsColor = Color.Green;
+		public float DebugMouseOverBoundsWidth = 2;
+
 		public RootWidget()
 		{
 			IgnoreMouseOver = true;
+
+			Color debugBoundsColor;
+			if (ChromeMetrics.TryGet<Color>("DebugBoundsColor", out debugBoundsColor))
+				DebugBoundsColor = debugBoundsColor;
+
+			float debugBoundsWidth;
+			if (ChromeMetrics.TryGet<float>("DebugBoundsWidth", out debugBoundsWidth))
+				DebugBoundsWidth = debugBoundsWidth;
+
+			Color debugMouseOverBoundsColor;
+			if (ChromeMetrics.TryGet<Color>("DebugMouseOverBoundsColor", out debugMouseOverBoundsColor))
+				DebugMouseOverBoundsColor = debugMouseOverBoundsColor;
+
+			float debugMouseOverBoundsWidth;
+			if (ChromeMetrics.TryGet<float>("DebugMouseOverBoundsWidth", out debugMouseOverBoundsWidth))
+				DebugMouseOverBoundsWidth = debugMouseOverBoundsWidth;
 		}
 
 		public override bool HandleKeyPress(KeyInput e)
diff --git a/OpenRA.Game/Widgets/Widget.cs b/OpenRA.Game/Widgets/Widget.cs
index cbad4e7..62d74e9 100644
--- a/OpenRA.Game/Widgets/Widget.cs
+++ b/OpenRA.Game/Widgets/Widget.cs
@@ -173,10 +173,6 @@ public abstract class Widget
 		public bool Visible = true;
 		public bool IgnoreMouseOver;
 		public bool IgnoreChildMouseOver;
-		public Color DebugBoundsColor = Color.Red;
-		public float DebugBoundsWidth = 1;
-		public Color DebugMouseOverBoundsColor = Color.Green;
-		public float DebugMouseOverBoundsWidth = 2;
 
 		// Calculated internally
 		public Rectangle Bounds;
@@ -201,11 +197,6 @@ public Widget(Widget widget)
 			IgnoreChildMouseOver = widget.IgnoreChildMouseOver;
 			IgnoreMouseOver = widget.IgnoreMouseOver;
 
-			DebugBoundsColor = widget.DebugBoundsColor;
-			DebugBoundsWidth = widget.DebugBoundsWidth;
-			DebugMouseOverBoundsColor = widget.DebugMouseOverBoundsColor;
-			DebugMouseOverBoundsWidth = widget.DebugMouseOverBoundsWidth;
-
 			foreach (var child in widget.Children)
 				AddChild(child.Clone());
 		}
@@ -439,7 +430,7 @@ public virtual void PrepareRenderablesOuter()
 
 		public virtual void Draw() { }
 
-		public void DrawDebugRenderBounds(bool drawEvenIfInvisible = false, bool recurseDownAllChildren = true)
+		public void DrawDebugRenderBounds(float width, Color color, bool drawEvenIfInvisible = false, bool recurseDownAllChildren = true)
 		{
 			if (drawEvenIfInvisible || IsVisible())
 			{
@@ -447,12 +438,12 @@ public void DrawDebugRenderBounds(bool drawEvenIfInvisible = false, bool recurse
 				Game.Renderer.RgbaColorRenderer.DrawRect(
 					new float2(RenderBounds.Left, RenderBounds.Top),
 					new float2(RenderBounds.Right, RenderBounds.Bottom),
-					isMouseOverWidget ? DebugMouseOverBoundsWidth : DebugBoundsWidth,
-					isMouseOverWidget ? DebugMouseOverBoundsColor : DebugBoundsColor);
+					width,
+					color);
 
 				if (recurseDownAllChildren)
 					foreach (var child in Children.OfType<Widget>().Reverse())
-						child.DrawDebugRenderBounds();
+						child.DrawDebugRenderBounds(/* what do I pass here? */);
 			}
 		}
```

Phrohdoh:
I've spent a lot of time trying to find the right place to store these new values and where to load them and cannot find a solution that works.

I'm closing this.
If someone else wants to open this up with their fix/solution be my guest.

PR #8059, opened 2015-04-28
Added option to replace disconnected clients with bots

mtarranger:
PR for [issue #3001](https://github.com/OpenRA/OpenRA/issues/3001): replacing disconnected clients with bots.

**Logic:**

Server drops player and send command "ActivateBot"
Game admin activates bot, notification in chat about this showed for everybody

**Detailed:**

(Option is enabled)
1. A player got dropped while a game is in progress.
2. Server doesn't send "player quit" packet (0xBF).
3. But it sends "Disconnected" instant message as usually - clients show "player disconnected" message.
4. Just after that server sends "ActivateBot" instant message.
- All clients (including admin) change their configuration so disconnected client is treated in the same way like it was a bot activated on the admin on start.
- It doesn't meter if dropped client had or had not sent packages for recent frames - clients will not check for them after this point (also look at explanation of third commit lower).
- All clients show "bot activated for <player name>" message
- Admin activates bot

Currently nothing will show that a player was real client and then was replaced by a bot except server chat message. Statistics will include combined player and bot progress after game end.

Please check spelling, especially for strings, variable names, code style, caveats.

I introduced:
"ActivateBot" command from server to clients
"Autopilot" string parameter in GlobalSettings which is a bot name if feature is enabled, null otherwise
"autopilot <bot name>" command from the admin to the lobby to enable option
"disable_autopilot" to disable

Please check implementation of the options gui. Do you have any cautions, will it work with specific maps (like scripted), foresights of bugs while player interact with the option in the lobby?

**Commits explanations (also look at the diffs and messages):**

Commit [1 of 4](https://github.com/OpenRA/OpenRA/commit/91491277ba9c3c21a0cb1a9b1d2bee18e2e5f00d):
- I didn't found any pattern for entries order for game options fields in lobby yaml, so I reordered them to follow left to right, line after line. That helped me in future changes.
- Some source and resource files had had execute bit on. It doesn't mean anything on windows, but that doesn't look pretty under *nix, so fixed some. On *nix you can check for that with `find -executable -type f` or `find -executable -type f -not -name '\*.sh' -not -path './.git/\*' -not -path './configure'` or auto fix with `find -executable -type f -name '\*.cs' | xargs chmod a-x` 

Commit [2 of 4](https://github.com/OpenRA/OpenRA/commit/626493676dbc47c272d56055d895e78c5b6f1b42):
[...]

PR #10502, opened 2016-01-13
Added an option to replace disconnected players with AI bots

Mailaender:
A simplified version of https://github.com/OpenRA/OpenRA/pull/8059. Closes https://github.com/OpenRA/OpenRA/issues/3001. Depends on https://github.com/OpenRA/OpenRA/pull/10787.

### DIFF
--- a/OpenRA.Game/Network/UnitOrders.cs
+++ b/OpenRA.Game/Network/UnitOrders.cs
@@ -101,6 +101,35 @@ namespace OpenRA.Network
 						break;
 					}
 
+				case "ReplacePlayerWithBot":
+					{
+						var client = orderManager.LobbyInfo.ClientWithIndex(clientId);
+						if (client != null && world != null)
+						{
+							var player = world.FindPlayerByClient(client);
+							if (player == null || player.IsBot)
+								break;
+
+							client.State = Session.ClientState.Ready;
+
+							var controller = orderManager.LobbyInfo.Clients.FirstOrDefault(c => c.Index == order.ExtraData);
+							if (controller == null)
+								break;
+
+							client.BotControllerClientIndex = controller.Index;
+							client.Bot = order.TargetString;
+							client.IpAddress = null;
+
+							orderManager.StopWaitingForClient(client);
+
+							player.ActivateBot(client.Bot);
+
+							Game.Debug("Replaced player {0} with bot {1} controlled by {2}".F(client.Name, client.Bot, controller.Name));
+						}
+
+						break;
+					}
+
 				case "TeamChat":
 					{
 						var client = orderManager.LobbyInfo.ClientWithIndex(clientId);
--- a/OpenRA.Game/Network/UnitOrders.cs
+++ b/OpenRA.Game/Network/UnitOrders.cs
@@ -67,6 +67,37 @@ namespace OpenRA.Network
 						var client = orderManager.LobbyInfo.ClientWithIndex(clientId);
 						if (client != null)
 							client.State = Session.ClientState.Disconnected;
+
+						break;
+					}
+
+				case "NewAdmin": // Find a new admin and hand bots over mid-game.
+					{
+						var client = orderManager.LobbyInfo.ClientWithIndex(clientId);
+						if (client == null || !client.IsAdmin)
+							break;
+
+						var nextAdmin = orderManager.LobbyInfo.Clients.FirstOrDefault(c => c.Index == order.ExtraData);
+						if (nextAdmin != null)
+						{
+							nextAdmin.IsAdmin = true;
+							Game.AddChatLine(Color.White, "Server", "{0} is now the admin.".F(nextAdmin.Name));
+
+							var bots = orderManager.LobbyInfo.Clients.Where(c => c.Bot != null && c.BotControllerClientIndex == client.Index);
+							foreach (var bot in bots)
+							{
+								bot.BotControllerClientIndex = nextAdmin.Index;
+								var botPlayer = orderManager.World.FindPlayerByClient(bot);
+								botPlayer.ActivateBot(bot.Bot);
+								Log.Write("server", "Bot {0} is now controlled by {1}.".F(bot.Name, nextAdmin.Name));
+							}
+						}
+						else
+						{
+							Log.Write("server", "Suggested new game admin {0} not found.".F(order.TargetString));
+							orderManager.LobbyInfo.Clients.RemoveAll(c => c.Bot != null && c.BotControllerClientIndex == client.Index);
+						}
+
 						break;
 					}
 
--- a/OpenRA.Game/Server/Server.cs
+++ b/OpenRA.Game/Server/Server.cs
@@ -594,8 +594,11 @@ namespace OpenRA.Server
 				var replaceWithBot = Settings.BotsForDisconnectedClients && State == ServerState.GameStarted;
 				if (replaceWithBot)
 				{
-					var bot = Map.Rules.Actors["player"].TraitInfos<IBotInfo>().Random(Random);
-					DispatchOrdersToClients(toDrop, 0, Order.ReplacePlayerWithBot(bot.Type, botController.Index).Serialize());
+					Session.LobbyOptionState dropoutReplacement;
+					var dropoutBotAvailable = LobbyInfo.GlobalSettings.LobbyOptions.TryGetValue("dropoutbotreplacment", out dropoutReplacement);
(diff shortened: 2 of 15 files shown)


OpenRA.Game/Settings.cs
```diff
@@ -73,6 +73,9 @@ public class ServerSettings
 		[Desc("Disallow AI bots.")]
 		public bool LockBots = false;
 
+		[Desc("If a player drops, the AI should take over.")]
+		public bool BotsForDisconnectedClients = true;
```
pchote: This should be a lobby/match option, not a global OpenRA setting.

OpenRA.Game/Server/Server.cs
```diff
@@ -579,7 +580,19 @@ public void DropClient(Connection toDrop, int frame)
...
+					var bot = Map.Rules.Actors["player"].TraitInfos<IBotInfo>().Random(Random);
+					DispatchOrdersToClients(toDrop, 0, new ServerOrder("ReplacePlayerWithBot", bot.Name).Serialize());
+
+					dropClient.IpAddress = null;
+					var admin = LobbyInfo.Clients.First(c => c.IsAdmin);
+					dropClient.BotControllerClientIndex = admin.Index;
+					dropClient.Bot = bot.Name;
+				}
+				else
+					LobbyInfo.Clients.RemoveAll(c => c.Index == toDrop.PlayerIndex);
```
Phrohdoh: Is there a reason you aren't using `RemoveAt(toDrop.PlayerIndex)`?

OpenRA.Game/Network/UnitOrders.cs
```diff
@@ -67,6 +67,35 @@ public static void ProcessOrder(OrderManager orderManager, World world, int clie
...
+						var client = orderManager.LobbyInfo.ClientWithIndex(clientId);
+						if (client != null && world != null)
+						{
+							var player = world.FindPlayerByClient(client);
+							if (player == null || player.IsBot)
+								break;
+
+							client.State = Session.ClientState.Ready;
+
+							var admin = orderManager.LobbyInfo.Clients.FirstOrDefault(c => c.IsAdmin);
```
pchote: What if it was the admin who dropped (on a dedicated server)?

Phrohdoh: Would it be possible to sort clients by ping and pick the best result to be the next admin, or is the idea of ping irrelevant if there is no admin/host?

OpenRA.Game/Server/Server.cs
Mailaender: This line hasn't changed, just moved.

OpenRA.Game/Network/UnitOrders.cs
Mailaender: @pchote Then this won't fire. See https://github.com/OpenRA/OpenRA/pull/10502#discussion-diff-49606608R583

@Phrohdoh See https://github.com/OpenRA/OpenRA/pull/10502/files#diff-e34ac0a837f43698d67024b60cd1cf12R598 but I doubt "ping" would be a good measurement.

pchote: That is my point, yes.  Introducing a half-baked feature that works for some clients but not others is asking for trouble.

pchote: Also, if the admin drops then there may also be bots that are now left without control.  Control for all of these should be switched to another player.

whinis:
Played with Nolt and Mailaender, when Nolt left bot took over and started playing as expected. No Desync Observed

OpenRA.Game/Network/UnitOrders.cs
Mailaender: Don't try to expand the scope here

> Added an option to replace disconnected players with AI bots

I didn't promise to fix bots on dedicated servers, nor transfer adminship and bot controlling. Those are existing limitations, but I don't think they should be tackled here.

pchote: The pr does not work as advertised:

> Added an option to replace disconnected players with AI bots

Players will not be replaced with an AI bot if they are the server admin.

pchote:
As mentioned above in the inline comments (which may be lost if the pr updates), I feel that this pr has fundamental flaws that cause more problems than they fix.  I have to give a :-1: for that reason.

OpenRA.Game/Settings.cs
Mailaender: Dedicated server admins will want to disable this. See the setting above.

pchote: Not all server admins.

OpenRA.Game/Network/UnitOrders.cs
whinis: Is there any particular reason the bot controller has to be admin ?
If not couldn't you select a play at random, assign the bots to them and if the dropped player had any bots remake them for different random players ?

Micr0Bit:
- "What if it was the admin who dropped (on a dedicated server)?"
- "Then this won't fire."
- "That is my point, yes. Introducing a half-baked feature that works for some clients but not others is asking for trouble."

this pretty much got me to the point for a :-1:

Mailaender:
Can we please move back to constructive criticism instead of throwing people to the lions with blunt negativity? You are creating a very hostile environment for contributors who give their best to improve the game.

pchote:
Unfortunately that is what happens when the initial review comments pointing out fundamental issues are dismissed with responses like like "Don't try to expand the scope too much" and "I went for KISS here. Don't change half the project, change the GUI for a small feature. Better get it to work now and extend it later. Don't always pull the whole arm, when someone gives you a hand.".   If you're going to shut down the review dialog with a blanket "no, i'm not going to fix that", then there is little choice but to become increasingly blunt in order to get the point across.

Mailaender:
I made a ticket for your request https://github.com/OpenRA/OpenRA/issues/10515. You should have noted that this is a pre-requisite already in https://github.com/OpenRA/OpenRA/pull/8059. Now it feels like an arbitrary request. I also don't have a solution for it and it probably should be implemented, tested and reviewed in it's own pull request.

OpenRA.Game/Settings.cs
Mailaender: Is there a reason for a player to not want to have this? The original https://github.com/OpenRA/OpenRA/pull/8059 already had it in so it is easy to copy it, but it felt like a needless switch in the already crowded lobby.

pchote:
The prerequisite #10787 was stalled by desync issues that will probably also be a problem here.  These problems aren't likely to be fixed in the foreseeable future, so unfortunately we can't do much more here either.

xp65:
It would be very cool if this was in the next release, would save many failed games due to a player leaving due to connection timing out later in the game removing all the fun :) Killing an idle player is quote stupid. Adding this feature would be worth a release on its own for me.

Micr0Bit:
@xp65 : i feel you , it seems that the recent increase on new players sadly also increasing the games with people just disconnecting or surrendering too early ... noticeable in ra - teamgames

Mailaender:
If you want this badly, you should first go and verify that https://github.com/OpenRA/OpenRA/pull/10787 works as it should. I did some tests locally with multiple clients and dedicated servers, but that is far from a real field test.

pchote:
Considering the current status of our @OpenRA/engine-hackers it seems unlikely that we will be able to get this into a merge worthy state in any reasonable time frame.  Moving back to the future milestone.

Mailaender:
I fixed this up so that it even works, when the admin disconnects on a dedicated server.

pchote:
See http://www.sleipnirstuff.com/forum/viewtopic.php?t=20232 for related discussion. I expect that enabling this by default will not be recieved well by some players, and considering how toxic that part of our community has become, i’d prefer not to anywhere near this.

Mailaender:
While I don't think you should reward toxic behavior with special attention, the request to disable this in the lobby is valid as it might not be for everyone and players/maps may want to pre-define the AI type.

KOYK:
@pchote "I expect that enabling this by default will not be recieved well by some players"

Yes because their fingers will get tired by clicking on it...I personally find this a good feature and any one who doesn't like it is because they never got a disconnected team mate who is about to fire a nuke in the last minute. By the way, adding staff is good specially if the user can turn them off and on, removing staff in the other hand is a bad thing to do, so toxic or not it doesn't really matter.

As long as it is a lobby option it doesn't matter if you even add chickens instead of tanks to the game.

pchote:
Moving back to the Future milestone alongside #10787.

Issue #10503, opened 2016-01-13
Make WithParachute respect player remap colors.


Phrohdoh:
Currently it does not remap as expected.
[It only has a palette property, not a matching player flag](https://github.com/OpenRA/OpenRA/blob/8a755e673c20ede44cab5d82e2a1b21e9b98fa97/OpenRA.Mods.Common/Traits/Render/WithParachute.cs#L34-L35).

![](https://cloud.githubusercontent.com/assets/756669/12301729/e1e0cb04-ba21-11e5-9a01-b34efab82a72.png)

pchote:
It also appears to be missing the lint attribute.

PR #10505, opened 2016-01-13
Fixed parachute palette configuration for Red Alert 2

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/10503.

image

~~To test you will have to merge this into https://github.com/OpenRA/OpenRA/pull/10482 and https://github.com/OpenRA/ra2/pull/10 into https://github.com/OpenRA/ra2/pull/28.~~

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithParachute.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithParachute.cs
@@ -32,7 +32,8 @@ namespace OpenRA.Mods.Common.Traits
 		[SequenceReference("Image")] public readonly string ClosingSequence = null;
 
 		[Desc("Palette used to render the parachute.")]
-		public readonly string Palette = "player";
+		[PaletteReference("IsPlayerPalette")] public readonly string Palette = "player";
+		public readonly bool IsPlayerPalette = true;
 
 		[Desc("Parachute position relative to the paradropped unit.")]
 		public readonly WVec Offset = new WVec(0, 0, 384);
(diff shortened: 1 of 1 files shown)


abcdefg30:
Looks good to me. :+1: 
Parachutes look correct in both ra and ra2. :white_check_mark:

OpenRA.Mods.Common/Traits/Render/WithParachute.cs
```diff
@@ -32,7 +32,8 @@ public class WithParachuteInfo : UpgradableTraitInfo, ITraitInfo, IRenderActorPr
 		[SequenceReference("Image")] public readonly string ClosingSequence = null;
 
 		[Desc("Palette used to render the parachute.")]
-		public readonly string Palette = "player";
+		[PaletteReference("IsPlayerPalette")] public readonly string Palette = "player";
```
Phrohdoh: Could you add a newline after this string definition (if you insist on putting the attribute on the same line).

This could confuse someone in the future because typically attributes are placed on the line before the object they apply to.

pchote:
This works as advertised, so :+1:

At the risk of creating drama, i'm taking this despite @Phrohdoh's nit because our attribute positioning is a mess project-wide, and that can/should be cleaned up everywhere as its own task.

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/658db8214e5ba02f81451aa1b25cca3f69a2f619

Mailaender:
Yeah, better change this project wide, because I was just doing it like we always did. In fact I had no idea what @Phrohdoh wanted to see changed there.

PR #10516, opened 2016-01-14
Recompressed a number of shp(td) files with shapeset

reaperrr:
There were still a number of shps with far from optimal compression. I recompressed all larger shps that showed signifcant gains (either in terms of percentage, absolute gains or both) with Blade's shapeset command line tool, which applies optimal compression to each frame.

I intentionally didn't check most files smaller than 4KB (too much hassle for low absolute gains), and also intentionally didn't update files where the gains were negligible, to keep the number of updated files in check.

This shrinks the installed size of OpenRA by ~700-800KB and the installer by ~1/10th of that.


pchote:
Looks fine to me :+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/fa8c0a2be75dd1c2aa85bd1763853c15324e0b9c)

Issue #7274, opened 2015-01-04
Clean up D2k YAML files.


penev92:
Example:
Combat tanks for all 3 Houses (`COMBATA`, `COMBATH`, `COMBATO`) have some traits attached to them that are also attached to `^COMBAT`, which they all inherit from. (with no changes to the traits' Info) This is not much of an issue, but I think it would be nice polishing.
I'm mainly dropping this here because I don't currently have the time to do it. If anyone feels like doing it - go ahead. If not - I hope I'll get around to it eventually.

penev92:
Unfortunatelly #7477 won't fix this completely. There are things still left like `carryall.infantry:` still using the default aircraft `CruiseAltitude`, which is almost half of that defined on the other carryalls.
Also `carryall.scripted:` is a hack that should be removed when harvester insurance gets implemented.
There are probably other things as well. :/

pchote:
How much of this is still relevant after the rules rewrite?

reaperrr:
I'm sure I fixed most of these during the migration to original balance. I'll try to check later.

penev92:
I believe we can close this now. Thanks!

PR #10517, opened 2016-01-14
Some D2k clean-ups

reaperrr:
Adresses most remaining issues from  #7274.

### DIFF
--- a/mods/d2k/rules/misc.yaml
+++ b/mods/d2k/rules/misc.yaml
@@ -223,6 +223,8 @@ upgrade.barracks:
 		FactionImages:
 			atreides: barracks.atreides
 			ordos: barracks.ordos
+			mercenary: barracks.ordos
+			smuggler: barracks.ordos
 	ProvidesPrerequisite@upgradename:
 
 upgrade.light:
@@ -244,6 +246,8 @@ upgrade.light:
 		FactionImages:
 			atreides: light.atreides
 			ordos: light.ordos
+			mercenary: light.ordos
+			smuggler: light.ordos
 	ProvidesPrerequisite@upgradename:
 
 upgrade.heavy:
@@ -266,6 +270,7 @@ upgrade.heavy:
 			atreides: heavy.atreides
 			ordos: heavy.ordos
 			mercenary: heavy.mercenary
+			smuggler: heavy.ordos
 	ProvidesPrerequisite@upgradename:
 
 upgrade.hightech:
--- a/mods/d2k/rules/misc.yaml
+++ b/mods/d2k/rules/misc.yaml
@@ -265,7 +265,7 @@ upgrade.heavy:
 		FactionImages:
 			atreides: heavy.atreides
 			ordos: heavy.ordos
-			corrino: heavy.corrino
+			mercenary: heavy.mercenary
 	ProvidesPrerequisite@upgradename:
 
 upgrade.hightech:
--- a/mods/d2k/rules/structures.yaml
+++ b/mods/d2k/rules/structures.yaml
@@ -43,6 +43,7 @@ concreteb:
 
 construction_yard:
 	Inherits: ^Building
+	-DamagedWithoutFoundation:
 	Building:
 		Footprint: xxx xxx
 		Dimensions: 3,2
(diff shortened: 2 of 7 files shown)


reaperrr:
Hm, yeah. I think Smugglers and Mercs were using Ordos graphics for everything else, right?

Mailaender:
I think the roof for the heavy factory is different, but otherwise all graphics are identical. Look at https://d2kplus.com/map-and-mission-editor-by-klofkac/ to be sure.

reaperrr:
Updated the merc/smuggler commit.

They now use ordos graphics for everything except their exclusive faction images (starport for smugglers, heavy factory for mercs).

Additionally, removed bogus, potentially crashing `corrino: conyard.corrino` entry (`conyard.corrino` sequence doesn't exist, corrino uses harkonnen gfx anyway).

mods/d2k/rules/structures.yaml
```diff
@@ -79,7 +80,8 @@ construction_yard:
 		FactionImages:
 			atreides: conyard.atreides
 			ordos: conyard.ordos
-			corrino: conyard.corrino
+			smuggler: conyard.ordos
+			mercenaries: conyard.ordos
```
pchote: You could make these definitions more efficient by making the default `Image: conyard.ordos` (and so on for other buildings), and then you only need to override `harkonnen: conyard.harkonnen` instead of overriding ordos, smugger, and mercinary.

mods/d2k/rules/world.yaml
```diff
@@ -35,6 +35,14 @@
 		Name: Corrino
 		InternalName: corrino
 		Selectable: false
+	Faction@Mercenaries:
+		Name: Mercenaries
+		InternalName: mercenaries
+		Selectable: false
+	Faction@Smugglers:
+		Name: Smugglers
+		InternalName: smuggler
```
pchote: Internal name `smuggler` (singular) vs `mercinaries` (plural) seems a bit odd.  Is this correct?

MustaphaTR: No, it should be `smugglers`.

mods/d2k/rules/structures.yaml
reaperrr: I noticed, but 
a) I'd have to override corrino as well, so we'd save only 1 line per structure and not 2, and 
b) it would increase the diff and changing the default increases the risk of regressions.

I'll do it if you insist, but I felt it wasn't worth the additional work.

pchote: Nope, not insisting.

mods/d2k/rules/world.yaml
reaperrr: In the original game the internal name for smugglers was `smuggler`, so I used that, but I couldn't find out the internal name for the mercs.

I now looked at the d2k exe a moment ago and the internal name for mercs is indeed singular `mercenary` as well.

Does it make a difference to the map importer? If yes, I'd change `mercenaries` to `mercenary`, if no, I'd change `smuggler` to `smugglers`.

pchote: Lets go with `mercenary`.

MustaphaTR: Plural ones sound better but if orginal game used singular ones go with singulars.

reaperrr:
Updated, changed internal name from `mercenaries` to `mercenary` to match original.

reaperrr:
Updated, moved the weapons.yaml split to its own PR.

mods/d2k/rules/misc.yaml
```diff
@@ -265,7 +265,7 @@ upgrade.heavy:
 		FactionImages:
 			atreides: heavy.atreides
 			ordos: heavy.ordos
-			corrino: heavy.corrino
+			mercenary: heavy.mercenary
```
pchote: This probably wants a `smugger` definition too for consistency.

pchote:
:+1: after that small addition.

reaperrr:
Updated, fixed upgraded merc/smuggler barracks and light factory images as well.

obrakmann:
Looks ok to me :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9a139aa86b4087337317354f733d3a5ef20fbcaa)

Issue #4823, opened 2014-03-07
Unit facings not in line with original games


reaperrr:
I know this is technically Westwoods fault for using such weird facing, but it's still rather noticeable in some cases.

Basically, for example for units with 32 facings, WW used the 4th frame when moving NW, 14th for SW, 20th for SE, 30th for NE, instead of 5,13,21,29 like OpenRA.

The problem is that Westwoods SHPs were made with their odd approach in mind, not the technically correct approach OpenRA is using. I believe some older versions of OpenRA actually took this into consideration, but some of the refactoring seems to have removed that.

This is most noticeable when watching harvesters harvest ore/tiberium, because their facing and harvest animation are not fully aligned visually.

PR #10521, opened 2016-01-15
Remove Util.GetFacing.

pchote:
This introduces WVec.Yaw and WAngle.Facing primitives, and uses them instead of Util.GetFacing.
This is another small step towards fixing our horrible facing code (which is required for proper handling of TS slopes), and is an immediate blocker to fixing #4823.

### DIFF
--- a/OpenRA.Game/WAngle.cs
+++ b/OpenRA.Game/WAngle.cs
@@ -41,6 +41,8 @@ namespace OpenRA
 		public bool Equals(WAngle other) { return other == this; }
 		public override bool Equals(object obj) { return obj is WAngle && Equals((WAngle)obj); }
 
+		public int Facing { get { return Angle / 4; } }
+
 		public int Sin() { return new WAngle(Angle - 256).Cos(); }
 
 		public int Cos()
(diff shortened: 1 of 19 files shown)


OpenRA.Game/WAngle.cs
```diff
@@ -41,6 +41,8 @@ public WAngle(int a)
 		public bool Equals(WAngle other) { return other == this; }
 		public override bool Equals(object obj) { return obj is WAngle && Equals((WAngle)obj); }
 
+		public int Facing { get { return Angle / 4; } }
```
Phrohdoh: Would this make more sense as `AsFacing`?

pchote: These are essentially fundamental properties of the owning type (like `X`, `Y`, `Angle`, etc), and not a conversion to another complex object.

reaperrr:
The code is a bit over my head, but I didn't spot anything that looks wrong.
Also spotted no in-game regressions, so I'll risk a :+1:

pchote:
Rebased.

PR #10509, opened 2016-01-13
Cancel custom input modes when opening the menu.

pchote:
Fixes #3277.

Testcase: Enter the building placement mode and then press escape to open the menu.

PR #10523, opened 2016-01-15
Fix a crash if an order generator is active when a game ends.

pchote:
The UI code expects to be run from an unsynced context, and as of #10509 it asserts this requirement with an exception (via the world.OrderGenerator setter).  This leads to a crash if you have PBOG or another order generator active when the GameOver event fires and the options button click handler cancels the OG.

Repro case: Destroy your last base building while having the building placement order generator active.


RoosterDragon:
:+1:

abcdefg30:
:+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3aac27d5e142d9a09cedfba908d340b028ec1954...84f92e37793f1811f6f07aae6c0efd50a20d225e)

pchote:
Fixed changelog: https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4055120f664d7c6a0f821b5eaff2189cdde32f0a...4d0a86930be3c26fd89dc8d117c81de818cca43d

(this was a fix for a bleed-only regression)

PR #10525, opened 2016-01-15
Split IPackage into IReadOnlyPackage and IReadWritePackage

pchote:
Another small step towards mod-defined package parsers.  This also removes the ability to package maps as a mix file (which was the only place where we used the mix file writer).

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -157,7 +157,7 @@ namespace OpenRA
 		[FieldLoader.Ignore] public readonly WVec[] SubCellOffsets;
 		public readonly SubCell DefaultSubCell;
 		public readonly SubCell LastSubCell;
-		[FieldLoader.Ignore] public IPackage Container;
+		[FieldLoader.Ignore] public IReadWritePackage Container;
 		public string Path { get; private set; }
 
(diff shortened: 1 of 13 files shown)


OpenRA.Game/Map/Map.cs
```diff
@@ -317,7 +317,9 @@ public Map(TileSet tileset, int width, int height)
 		public Map(string path)
 		{
 			Path = path;
-			Container = Game.ModData.ModFiles.OpenPackage(path, null, int.MaxValue);
+			Container = Game.ModData.ModFiles.OpenPackage(path, null, int.MaxValue) as IReadWritePackage;
+			if (Container == null)
+				throw new InvalidDataException("Unsupported map package type");
```
RoosterDragon: I would move this try-upcast-and-throw logic into a method such as `OpenPackageAsWritable` or such as I can imagine that would be a common pattern for anybody else who needs a writeable package.

RoosterDragon:
:+1: LGTM modulo my request for the helper method.

pchote:
Fixed.

Mailaender:
Indeed, no one uses our tools to create packages. This just removes a collection of `NotImplementedException` stub methods.

PR #10526, opened 2016-01-15
Improve ProximityCapturable

obrakmann:
The following changes are made:
- Replaced the custom actor search with ProximityTriggers.
- Added a `Sticky` flag, which causes the ownership to not revert back to
  the original owner when the captor leaves.
- Added an option to show a range circle for the actor, either constantly
  or on selection, drawn in the colour of the current owner.
  This will be especially useful for invisible ProximityCapturables.
- Added trait documentation.

### DIFF
--- a/OpenRA.Mods.Common/Traits/ProximityCapturable.cs
+++ b/OpenRA.Mods.Common/Traits/ProximityCapturable.cs
@@ -9,8 +9,11 @@
 #endregion
 
 using System.Collections.Generic;
+using System.Drawing;
 using System.Linq;
 using OpenRA.Effects;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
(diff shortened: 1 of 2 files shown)


abcdefg30:
Looks good to me, and ingame worked fine, too. :+1: / :white_check_mark:

OpenRA.Mods.Common/Traits/ProximityCapturable.cs
```diff
@@ -9,24 +9,44 @@
...
+using System.Drawing;
 using System.Linq;
 using OpenRA.Effects;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
+	public enum RangeCircleVisibility { Never, Always, WhenSelected }
```
Mailaender: How about using `WithRangeCircle` instead or `RangeCircleRenderable.RenderDebugGeometry`?

obrakmann: Fixing `WithRangeCircle` makes sense. I never quite liked its limitations. Updated.

@abcdefg30: I guess I'll have to remove your :+1: for that, sorry.

Mailaender:
Can you rebase this please? https://github.com/OpenRA/OpenRA/issues/10548#issuecomment-172661671

obrakmann:
Rebased. Added a testcase as well, with which I found and fixed a few bugs.

The test case is Hope's Anchor. Harvester can be proximity-captured in this version (both `Sticky` and `MustBeClear` are on), and also the three strategic points have different characteristics:

* The left-most one: acts like a regular strategic point, but has a range circle when selected.
* The middle one: has the sticky behaviour, and `MustBeClear` is off. Also has the range circle on selection.
* The right-most one: Can be captured only once, and always shows the range circle.

Mailaender:

abcdefg30:
Looks good to me. :+1:  Remove the testcase, please.

obrakmann:
Test case removed. Let the record show that it can be found as commit b3794ff.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/deedeb7c0f2d5fd49f76a596b281644b17046283...609de6edce5808ad4110c1be3862ddbe26892b47)

Issue #10529, opened 2016-01-16
Installing from CD does not work


pchote:
The installer creates the /Content directory, but not the Content/mod subdir.  It then fails because it can't find the dir to copy the files to.

pchote:
We should also truncate the error message to fit the dialog.

![screen shot 2016-01-16 at 11 40 53](https://cloud.githubusercontent.com/assets/167819/12372180/20cf2bfa-bc46-11e5-9305-9ec8a3c78802.png)

yaaaaa:
maybe it halp to metod finde te path on disk?


[startool.zip](https://github.com/OpenRA/OpenRA/files/92871/startool.zip)

Mailaender:
This was already fixed for the TFD installation in https://github.com/OpenRA/OpenRA/pull/9622.

PR #10530, opened 2016-01-16
Index filesystem contents by filename instead of hash.

pchote:
This removes the requirement for package formats to expose classic/crc hashes to the filesystem code.  These hashes were included as a workaround for the mix format not storing filenames.  We now instead require that all mix files have corresponding mix database entries (true for all default mixes as well as any custom mixes with valid local mix database.  This will break any mods that intentionally write corrupted mix databases as "copy protection").

Tested that TD/RA/TS still run, and that they can still be installed from CD (however note #10529).


Mailaender:
Does this give any benefit as in performance wise?

pchote:
In principle, yes, but only after the rest of the filesystem fixes (I have more, but they are blocked on currently open prs) are merged.

pchote:
Rebased.

Note that this introduces an important change:  If code asks for files with the wrong case then they will miss the cache and take the slow path.  #10531 fixes the only place I could find this to be an issue.

pchote:
I have pushed some additional cleanup commits that don't make a lot of sense to review separately.  I have tested the InstallShield and D2kSoundResources changes, but am not able to test the changes to BagFile (which is a RA2-specific format).

pchote:
Note that this this makes InstallShieldPackage / D2kSoundResources / BagFile case sensitive.

abcdefg30:
All mods (and RA2) still load and work. :+0.75:

mods/cnc/mod.yaml
```diff
@@ -146,13 +146,11 @@ LoadScreen: CncLoadScreen
...
 ContentInstaller:
 	TestFiles: ^Content/cnc/conquer.mix, ^Content/cnc/desert.mix, ^Content/cnc/sounds.mix, ^Content/cnc/speech.mix, ^Content/cnc/temperat.mix, ^Content/cnc/tempicnh.mix, ^Content/cnc/winter.mix
-	FilesToCopy: CONQUER.MIX, DESERT.MIX, SCORES.MIX, SOUNDS.MIX, TEMPERAT.MIX, WINTER.MIX
-	FilesToExtract: speech.mix, tempicnh.mix, transit.mix
 	PackageMirrorList: http://www.openra.net/packages/cnc-mirrors.txt
 	DiskTestFiles: conquer.mix, desert.mix, install/setup.z
 	PackageToExtractFromCD: install/setup.z
 	ExtractFilesFromCD:
-		.: speech.mix, tempicnh.mix, transit.mix
+		.: SPEECH.MIX, TEMPICNH.MIX, TRANSIT.MIX
```
Mailaender: This looks like it could be dangerous for case-sensitive file systems. Don't have a CD drive to test on Linux at hand, sorry.

pchote: These files are inside setup.z, which lists them in upper case.  The install code .ToLowerInvariants the output filename, so there is no issue for case sensitive file systems.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -107,16 +106,9 @@ void MountInner(IReadOnlyPackage package)
 		{
 			MountedPackages.Add(package);
 
-			foreach (var hash in package.ClassicHashes())
+			foreach (var hash in package.AllFileNames())
```
penev92: Nit: This is no longer a hash.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -127,6 +119,9 @@ public bool Unmount(IReadOnlyPackage mount)
 			if (MountedPackages.Contains(mount))
 				mount.Dispose();
 
+			foreach (var f in fileIndex)
+				f.Value.RemoveAll(p => p == mount);
```
penev92: Shouldn't this go *before* the `.Dispose()` above?

pchote:
Fixed.

penev92:
D2k CD asset installation crashes with a duplicate key exception.
I traced it to a second "LAUNCHER.BMP" being read from `setup.z` and added to `InstallShieldPackage.index`. You will need to read the directory names inside the package as well and attach those to the file names, I guess (they are currently read and ignored).

pchote:
Fixed D2K installation from CD – both my own issues and preexisting problems on case sensitive file systems.

abcdefg30:
> OpenRA.Mods.Common/Widgets/Logic/Installation/InstallFromCDLogic.cs:L194: [SP2100] Code line must be no longer than 180 characters (now 181).

pchote:
Fixed.

pchote:
Rebased.

Mailaender:
Dune 2000 package installation works so far except videos:

```yaml
Videos:
	Briefing: a_br01_e.vqa
```

vs.

image

Mailaender:
RA2 also still works. :white_check_mark:

pchote:
Fixed.

pchote:
Rebased again.

obrakmann:
Aside from that one comment, looks fine to me. Confirmed that current installations continue to run, and that installation from both the freeware CDs as well as downloaded content still work. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9ab4b27f52967b594aa9891ad176e310cff6227d)

PR #10531, opened 2016-01-16
Fix case mismatches in TS tileset importer.

pchote:
As of #10530 the fast path for loading files requires a case sensitive match for filenames.  The filenames listed in the xcc mix databases are all in lower case.  This fixes the TS tileset definitions to use the correct case.


Mailaender:
Using your newly generated files will also fix the case-sensitivity problems I have with @gangster-ppm's Rewire builds. :+1:

obrakmann:
:+1:

PR #10532, opened 2016-01-16
Apply some general polish to the code/yaml

abcdefg30:
You probably want to review the 3rd commit with `?w=1`: https://github.com/OpenRA/OpenRA/commit/ac70d1ed097c360c499dfc70ffe89d18c7fd40c1?w=1

### DIFF
--- a/OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
+++ b/OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
@@ -249,66 +249,61 @@ namespace OpenRA.Mods.Common.Traits
 			switch (order.OrderString)
 			{
 				case "StartProduction":
+					var unit = rules.Actors[order.TargetString];
+					var bi = unit.TraitInfo<BuildableInfo>();
+
+					// Not built by this queue
+					if (!bi.Queue.Contains(Info.Type))
+						return;
+
+					var cost = unit.HasTraitInfo<ValuedInfo>() ? unit.TraitInfo<ValuedInfo>().Cost : 0;
+					var time = GetBuildTime(order.TargetString);
+
+					// You can't build that
+					if (BuildableItems().All(b => b.Name != order.TargetString))
+						return;
+
+					// Check if the player is trying to build more units that they are allowed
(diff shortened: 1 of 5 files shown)


abcdefg30:
Updated.

abcdefg30:
Rebased.

Mailaender:
:+1:

mods/ts/rules/defaults.yaml
```diff
@@ -77,9 +77,21 @@
 		UpgradeTypes: notmobile
 		UpgradeMaxEnabledLevel: 0
 
+^CloakableUnit:
```
penev92: Nit: "CloakableActor", or just "Cloakable" please.

abcdefg30: Changed it to `Cloakable`.

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -249,80 +249,83 @@ public void ResolveOrder(Actor self, Order order)
...
+			if (self.World.AllowDevCommands && self.Owner.PlayerActor.Trait<DeveloperMode>().FastBuild)
+				return 0;
+
 			var unit = self.World.Map.Rules.Actors[unitString];
-			if (unit == null || !unit.HasTraitInfo<BuildableInfo>())
+			if (unit == null)
 				return 0;
 
+			var time = unit.GetBuildTime() * Info.BuildSpeed;
+			return (int)time;
```
obrakmann: You could just `return GetBuildTime(unit)` here. Move the null check up and remove the cheat check.

obrakmann:
:+1: to everything except the last commit. You would need to write a `ClassicProductionQueue.GetBuildTime(ActorInfo unit)` as well. Currently this causes the production tooltip in RA to show the wrong build time when you have multiple factories.

abcdefg30:
Dropped the last commit, as it wasn't polish anymore and got complex. I'll file a new pr for that.

abcdefg30:
Here is the other pr:  #10557

obrakmann:
:+1:

PR #10535, opened 2016-01-16
Auto-detect mix hash type.

pchote:
This rewrites the mix file parser to be guess whether it is uses classic or CRC style hashes, and removes the last direct workarounds for mix file uglyness from the rest of the engine.

This can impact/regress mods in two ways:
1. A valid local and/or global mix database containing the filenames is now required to load files from the package.  This shouldn't be a problem unless somebody went out of their way to make their mix files unusable by XCC mixer.
2.  Filenames inside mixes are now case sensitive, and must be queried using the exact case written in the local or global mix database (usually all lower case).

I've tested that TD/RA/TS still successfully install from CD (except for #10529) and can launch a skirmish.

### DIFF
--- a/OpenRA.Game/FileSystem/MixFile.cs
+++ b/OpenRA.Game/FileSystem/MixFile.cs
@@ -20,19 +20,18 @@ namespace OpenRA.FileSystem
 {
 	public sealed class MixFile : IReadOnlyPackage
 	{
-		readonly Dictionary<uint, PackageEntry> index;
+		public string Name { get; private set; }
+
+		readonly Dictionary<string, PackageEntry> index;
 		readonly long dataStart;
 		readonly Stream s;
 		readonly int priority;
-		readonly string filename;
 		readonly FileSystem context;
-		readonly PackageHashType type;
 
-		public MixFile(FileSystem context, string filename, PackageHashType type, int priority)
+		public MixFile(FileSystem context, string filename, int priority)
 		{
-			this.filename = filename;
+			Name = filename;
 			this.priority = priority;
-			this.type = type;
 			this.context = context;
 
 			s = context.Open(filename);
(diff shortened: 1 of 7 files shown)


GraionDilach:
Most `Report`s in the TD/RA/TS mods are spelled with capital letters. This needs a commit fixing them even.

pchote:
See #10538.

GraionDilach:
TS has capital filenames in
 - voices.yaml
 - weapon reports.

pchote:
Prereqs have now been merged, so rebased.  Testing should be similar to #10530.

OpenRA.Game/FileSystem/MixFile.cs
```diff
@@ -66,6 +64,58 @@ public MixFile(FileSystem context, string filename, PackageHashType type, int pr
...
+					var db = new XccLocalDatabase(GetContent(kv.Value));
+					foreach (var e in db.Entries)
+						allPossibleFilenames.Add(e);
+
+					break;
+				}
+			}
+
+			// Load the global mix database
+			// TODO: This should be passed to the mix file ctor
```
pchote: Note: This TODO will be addressed in the pr that introduces the concept of a package loader (like our existing sprite/sequence/audio loaders)

reaperrr:
I'm getting `FileSystem/MixFile.cs(114,84): error CS0135: 'filename' conflicts with the declaration 'OpenRA.FileSystem.MixFile.filename'` when trying to compile this branch.

pchote:
@reaperrr: I backported a change from the next pr that will probably fix that.  Can you try again?

reaperrr:
That issue is fixed now.

RA/TD/TS still load mod + skirmish fine with our download packages, didn't notice any regressions there. 
Didn't test CD installations yet, so :+0.5: for now.

pchote:
Adding to next release milestone so that modders can fix their package definitions and fix any case sensitivity issues in one go, instead of dragging it over multiple releases.

Mailaender:
This causes a not so nice regression when the mod is unprepared where this creates a lot of folders (will probably crash due to not being writeable in a non-dev environment).

image

Can we get a `mod.yaml` upgrade rule for this? Might be faster than writing documentation and pointing people at it.

pchote:
> Can we get a mod.yaml upgrade rule for this? Might be faster than writing documentation and pointing people at it.

This isn't possible because the upgrader relies on a working mod.yaml to run.

Mailaender:
Also works for Red Alert 2. :white_check_mark:

abcdefg30:
:+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a683e29df13bd5d46d3e19ac3e592155f7560835) - Merged the file system PR entries together as requested, too.

PR #10536, opened 2016-01-16
Added a warning when bots have been disabled

Mailaender:
If people click skirmish and choose an island map, they get a wrong illusion that we don't support AI. See https://www.facebook.com/openra/posts/993082560747961?comment_id=993237267399157&comment_tracking=%7B%22tn%22%3A%22R0%22%7D


pchote:
These should be disabled (and possibly show a different message when first joining) if the server disables bots globally.

Mailaender:
Good idea. Done.

chrisforbes:
With that change, :+1:

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f48fbe23826ab1bd6244e12115c942e6f3f5cb0c...e623f0c014c33a384963396af475c96a3d0322a7)

PR #10538, opened 2016-01-17
Fix filename case on weapon reports.

pchote:
Fixes the RA and TD mods in preparation for #10535.

### DIFF
--- a/mods/cnc/weapons/largecaliber.yaml
+++ b/mods/cnc/weapons/largecaliber.yaml
@@ -1,7 +1,7 @@
 70mm:
 	ReloadDelay: 30
 	Range: 4c0
-	Report: TNKFIRE3.AUD
+	Report: tnkfire3.aud
 	Projectile: Bullet
 		Image: 120MM
 		Speed: 853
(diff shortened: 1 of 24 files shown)


abcdefg30:
Could we get an upgrade rule for maps?

pchote:
I don't think we can, unless you want to risk the rule breaking other things.  For example, D2K's sounds are legitimately upper case.

GraionDilach:
TS isn't fixed though, on the other hand, even unit voices are capitalized in TS.

pchote:
@GraionDilach: could you list all the case inconsistencies you know of in #10535?  This one is reasonably self-contained, so I'll do another pr for the remaining cleanups.

GraionDilach:
Oh, I overlooked for some reason I haven't seen the TS fixes in this.

pchote:
Added a commit for TS voices, since that is easy enough to test while testing the weapon reports in TS.

Mailaender:
I can still hear the sounds. :white_check_mark: It would be nice to get a pull request for https://github.com/OpenRA/ra2/ or at least get the command you used for this conversion.

pchote:
I used the global find and replace in textmate with a search string of `Report: (\w+)` and a replacement string of `Report: \L$1.aud`.

abcdefg30:
Sounds still work ingame. :white_check_mark:

PR #10542, opened 2016-01-17
General Cleanup

RoosterDragon:
Run automated fixes from VS to simplify names, remove unused usings and remove redundant casts.

I have made inline comments against minor tweaks where I intervened - fixes are otherwise automated so should be pretty safe to take despite the diff size.

### DIFF
--- a/OpenRA.Game/Scripting/ScriptContext.cs
+++ b/OpenRA.Game/Scripting/ScriptContext.cs
@@ -86,8 +86,8 @@ namespace OpenRA.Scripting
 		public ScriptGlobal(ScriptContext context)
 			: base(context)
 		{
-			// The 'this.' resolves the actual (subclass) type
-			var type = this.GetType();
+			// GetType resolves the actual (subclass) type
+			var type = GetType();
 			var names = type.GetCustomAttributes<ScriptGlobalAttribute>(true);
 			if (names.Length != 1)
 				throw new InvalidOperationException("[ScriptGlobal] attribute not found for global table '{0}'".F(type));
--- a/OpenRA.Game/WVec.cs
+++ b/OpenRA.Game/WVec.cs
@@ -82,7 +82,7 @@ namespace OpenRA
 
 			// Add an additional quadratic variation to height
 			// Uses fp to avoid integer overflow
-			var offset = (int)((float)((float)(b - a).Length * pitch.Tan() * mul * (div - mul)) / (float)(1024 * div * div));
+			var offset = (int)((float)(b - a).Length * pitch.Tan() * mul * (div - mul) / (1024 * div * div));
 			return new WVec(ret.X, ret.Y, ret.Z + offset);
 		}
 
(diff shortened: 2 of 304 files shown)


OpenRA.Game/Scripting/ScriptContext.cs
```diff
@@ -86,8 +86,8 @@ public abstract class ScriptGlobal : ScriptObjectWrapper
 		public ScriptGlobal(ScriptContext context)
 			: base(context)
 		{
-			// The 'this.' resolves the actual (subclass) type
-			var type = this.GetType();
+			// GetType resolves the actual (subclass) type
+			var type = GetType();
```
RoosterDragon: Fixed the comment here.

OpenRA.Platforms.Default/Sdl2Input.cs
```diff
@@ -17,7 +17,7 @@ namespace OpenRA.Platforms.Default
 {
 	class Sdl2Input
 	{
-		MouseButton lastButtonBits = (MouseButton)0;
+		MouseButton lastButtonBits = MouseButton.None;
```
RoosterDragon: This was initially simplified to `0` - I have instead tweaked it to the equivalent `MouseButton.None`.

Mailaender:
:+1:

OpenRA.Game/Scripting/ScriptContext.cs
pchote: This may break things for people who aren't using the same compiler/runtime as you

pchote: Specifically: IIRC, when I wrote this, the `this.` was needed to make this not return `ScriptGlobal` for everything.

RoosterDragon: `GetType()` doesn't work that way.

pchote: Are you confident that is true across the various older mono releases?
This was definitely an issue at the time, otherwise I wouldn't have left a comment specifically mentioning it.

RoosterDragon: Are you sure you're not getting mixed up with `typeof(ScriptGlobal)` versus `this.GetType()`?

pchote: I can still remember the giant WTF moment when I found that adding `this.` fixed the issues that I was having, so i'm pretty sure, yeah.

RoosterDragon: Well we're already doing an unqualified call in the abstract [SupportPowerInfo](https://github.com/OpenRA/OpenRA/blob/9b6dba02a89f93879e151e0c6cbc5c8885e0e3dd/OpenRA.Mods.Common/Traits/SupportPowers/SupportPower.cs#L58) class - so I suspect this is no longer an issue (if it ever was) or support powers would be quite broken.

pchote: Fair enough.  Lets take this, and we'll know where to look if something breaks.

OpenRA.Game/WVec.cs
```diff
@@ -82,7 +82,7 @@ public static WVec LerpQuadratic(WVec a, WVec b, WAngle pitch, int mul, int div)
 
 			// Add an additional quadratic variation to height
 			// Uses fp to avoid integer overflow
-			var offset = (int)((float)((float)(b - a).Length * pitch.Tan() * mul * (div - mul)) / (float)(1024 * div * div));
```
pchote: Maybe convert this to `decimal` while we're here?  Trade a bit of perf for a guarantee to not desync.

pchote: Not in this pr though, lets keep this one automatic fixes.

Phrohdoh: How does a `decimal` type guarantee no desync?

pchote: `decimal` math is implemented in software, and doesn't use the FPU (with all the trouble that brings).

abcdefg30:
:+1:

Issue #10533, opened 2016-01-16
RFE: Persistent Skirmish/multiplayer game options and ability to restart with said options


MagicFab:
I'd like to be able to restart a game when in Skirmish or multiplayer modes while kepeing the same game options.

When initially setting up such a game, one may alter game speed, money available, player slots (including AI, location, teams, etc), map, etc. Once the game is finished the only choice is to completely exit and having to set again all options manually to restart the same type of game.

It would be convenient to be able to restart the same game with the same options quickly (add a "Play again" option?), while keeping the option to reset completely and start over.
##

whinis:
Now that #10264 has been merged I am starting work on just this, skirmish is likely to happen before multiplayer.

abcdefg30:
Closing as fixed/duplicate of #4122.

PR #10543, opened 2016-01-18
Restart skirmish matches

whinis:
This allows one to return to the skirmish lobby with the same settings and bots as one had in a previous match. Tested while in the middle of a match with another bot, could go back to lobby and then hit start again with no problems and all settings and bots stayed the same

Should handle the skirmish part of #10533

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -179,8 +179,11 @@ namespace OpenRA
 		{
 			var replay = OrderManager.Connection as ReplayConnection;
 			var replayName = replay != null ? replay.Filename : null;
-			var uid = OrderManager.World.Map.Uid;
-			var globalSettings = OrderManager.LobbyInfo.GlobalSettings;
+			var lobbyInfo = OrderManager.LobbyInfo;
+			var orders = new[] {
+					Order.Command("sync_lobby {0}".F(lobbyInfo.Serialize())),
+					Order.Command("startgame")
+			};
 
 			// Disconnect from the current game
 			Disconnect();
@@ -190,10 +193,10 @@ namespace OpenRA
 			if (replay != null)
 				JoinReplay(replayName);
 			else
-				StartMission(uid, globalSettings.GameSpeedType, globalSettings.Difficulty);
+				CreateAndStartLocalServer(lobbyInfo.GlobalSettings.Map, orders);
 		}
 
-		public static void StartMission(string mapUID, string gameSpeed, string difficulty, Action onStart = null)
+		public static void CreateAndStartLocalServer(string mapUID, IEnumerable<Order> setupOrders, Action onStart = null)
 		{
 			OrderManager om = null;
 
@@ -201,9 +204,9 @@ namespace OpenRA
 			lobbyReady = () =>
 			{
 				LobbyInfoChanged -= lobbyReady;
-				om.IssueOrder(Order.Command("gamespeed {0}".F(gameSpeed)));
-				om.IssueOrder(Order.Command("difficulty {0}".F(difficulty)));
-				om.IssueOrder(Order.Command("state {0}".F(Session.ClientState.Ready)));
+				foreach (var o in setupOrders)
+					om.IssueOrder(o);
+
 				if (onStart != null)
 					onStart();
 			};
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
@@ -90,14 +90,14 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			{
 				hideMenu = true;
 
-				if (world.LocalPlayer == null || (world.LocalPlayer.WinState != WinState.Won &&
-					(!world.IsGameOver || world.Map.Visibility == MapVisibility.MissionSelector)))
+				if (world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Won)
 				{
 					Action restartAction = null;
-					if (world.IsReplay || world.Map.Visibility == MapVisibility.MissionSelector)
+					var iop = world.WorldActor.TraitsImplementing<IObjectivesPanel>().FirstOrDefault();
+					var exitDelay = iop != null ? iop.ExitDelay : 0;
+
+					if (world.LobbyInfo.IsSinglePlayer)
 					{
-						var iop = world.WorldActor.TraitsImplementing<IObjectivesPanel>().FirstOrDefault();
-						var exitDelay = iop != null ? iop.ExitDelay : 0;
 						restartAction = () =>
 						{
 							Ui.CloseWindow();
@@ -145,7 +145,9 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			};
 			var surrenderButton = menu.Get<ButtonWidget>("SURRENDER");
 			surrenderButton.IsVisible = () => world.Type == WorldType.Regular;
-			surrenderButton.IsDisabled = () => (world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined) || hasError;
+			surrenderButton.IsDisabled = () =>
+				world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined ||
+				world.Map.Visibility.HasFlag(MapVisibility.MissionSelector) || hasError;
 			surrenderButton.OnClick = () =>
 			{
 				hideMenu = true;
@@ -153,7 +155,9 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 					title: "Surrender",
 					text: "Are you sure you want to surrender?",
 					onConfirm: onSurrender,
-					onCancel: showMenu);
+					onCancel: showMenu,
+					confirmText: "Surrender",
+					cancelText: "Stay");
 			};
 
 			var saveMapButton = menu.Get<ButtonWidget>("SAVE_MAP");
--- a/OpenRA.Mods.Common/Widgets/Logic/MissionBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/MissionBrowserLogic.cs
@@ -14,6 +14,7 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using OpenRA.Graphics;
+using OpenRA.Network;
 using OpenRA.Primitives;
 using OpenRA.Widgets;
 
@@ -301,6 +302,12 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				return;
 
 			var gameStartVideo = selectedMap.Videos.GameStart;
+			var orders = new[] {
+				Order.Command("gamespeed {0}".F(gameSpeed)),
+				Order.Command("difficulty {0}".F(difficulty)),
+				Order.Command("state {0}".F(Session.ClientState.Ready))
+			};
+
 			if (gameStartVideo != null && Game.ModData.ModFiles.Exists(gameStartVideo))
 			{
 				var fsPlayer = fullscreenVideoPlayer.Get<VqaPlayerWidget>("PLAYER");
@@ -308,11 +315,11 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				PlayVideo(fsPlayer, gameStartVideo, PlayingVideo.GameStart, () =>
 				{
 					StopVideo(fsPlayer);
-					Game.StartMission(selectedMapPreview.Uid, gameSpeed, difficulty, onStart);
+					Game.CreateAndStartLocalServer(selectedMapPreview.Uid, orders, onStart);
 				});
 			}
 			else
-				Game.StartMission(selectedMapPreview.Uid, gameSpeed, difficulty, onStart);
+				Game.CreateAndStartLocalServer(selectedMapPreview.Uid, orders, onStart);
 		}
 
 		class DropDownOption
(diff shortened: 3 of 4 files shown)


OpenRA.Game/Game.cs
```diff
@@ -189,8 +193,47 @@ public static void RestartGame()
...
+						}
+					},
+					{ "onStart",
+						() => { }
+					},
+					{ "skirmishMode", true }
+				});
+
+				// Needs to run after delay to overwrite skirmish default bot addition
+				RunAfterDelay(200, () =>
```
whinis: I could change this to 
`LobbyInfoChanged += WidgetUtils.Once(() =>` if I set skirmishMode to false however that changes the back button to leave instead of back

pchote:
This relies on directly manipulating the internal state of the server thread from the client, which is a bad idea for several reasons.  All communication with the server needs to go via orders.

whinis:
I am essentially taking the old LobbyInfo from the server (its a local server) and then using the SyncInfo command to push it to the client. The only other way I can see to do this is to loop through every setting, every bot, and every option of the bot and set them individually. This would not only make some nasty code, it would spam the chat log every time you hit restart as all the settings are changed. 

Do you have another idea ?

whinis:
Updated to to use a client -> server command

Mailaender:

OpenRA.Game/Game.cs
```diff
@@ -189,8 +192,20 @@ public static void RestartGame()
 			// Restart the game with the same replay/mission
 			if (replay != null)
 				JoinReplay(replayName);
-			else
+			else if (mission)
 				StartMission(uid, globalSettings.GameSpeedType, globalSettings.Difficulty);
+			else
+			{
+				JoinServer(IPAddress.Loopback.ToString(), CreateLocalServer(map), "");
+				LobbyInfoChanged += WidgetUtils.Once(() =>
```
pchote: Can you combine this with `StartMission` (which was only introduced in the last PR, IIRC) to avoid duplication?

pchote: I suggest splitting this into a method that takes the game speed and difficulty and returns a LobbyInfo, and then a common method that takes a lobbyinfo and sets up and starts a game.

whinis:
Reverted the changes I made in #10264 for https://github.com/whinis/OpenRA/blob/9059e3e2c81df5b79fa2bc2461004a0194d49040/OpenRA.Mods.Common/Widgets/Logic/MissionBrowserLogic.cs as there is no sane way to implement that in this code. This simplified the restart code as the same method can be used for both

OpenRA.Game/Game.cs
```diff
@@ -190,26 +189,18 @@ public static void RestartGame()
...
 			{
-				LobbyInfoChanged -= lobbyReady;
-				om.IssueOrder(Order.Command("gamespeed {0}".F(gameSpeed)));
-				om.IssueOrder(Order.Command("difficulty {0}".F(difficulty)));
-				om.IssueOrder(Order.Command("state {0}".F(Session.ClientState.Ready)));
-				if (onStart != null)
-					onStart();
-			};
-			LobbyInfoChanged += lobbyReady;
+				LobbyInfoChanged += WidgetUtils.Once(() =>
```
pchote: Please use the same pattern as `StartMission` here: Bind to LobbyInfo and then remove the binding from inside the callback before it returns.

OpenRA.Game/Game.cs
```diff
@@ -190,26 +189,18 @@ public static void RestartGame()
...
+				{
+					OrderManager.IssueOrder(Order.Command("sync_lobby {0}".F(lobbyInfo.Serialize())));
+					RunAfterTick(() =>
+					{
+						OrderManager.IssueOrder(Order.Command("startgame"));
```
pchote: Why does this need to be different from what `StartMission` does?

OpenRA.Mods.Common/Widgets/Logic/MissionBrowserLogic.cs
```diff
@@ -300,14 +303,32 @@ void StartMissionClicked()
...
-		}
+				StartMission();
+        }
+
+		void StartMission()
+		{
+			OrderManager om = null;
+
+			Action lobbyReady = null;
+			lobbyReady = () =>
```
pchote: You should be able to generalize this for both cases by having it take a list of orders to issue as arguments (with gamespeed/difficulty for missions and sync_lobby for skirmish)

whinis: The reason its difficult is that the server sets up a large section of the lobby info for me including map,slots, and clients. So while its easy to make that work for restarting as there is already a lobby info around at first start there is not. My understanding of lambda functions is that whats inside of them is from local scope whenever they are created. This means I can't get the current lobby info inside of them or atleast last night I couldn't think of a nice way of doing it.

pchote: I'm suggesting something like the following:
```csharp
public static void CreateAndStartLocalServer(string mapUID, IEnumerable<Order> setupOrders, Action onStart = null)
{
	OrderManager om = null;

	Action lobbyReady = null;
	lobbyReady = () =>
	{
		LobbyInfoChanged -= lobbyReady;
		foreach (var o in setupOrders)
			om.IssueOrder(o);

		if (onStart != null)
			onStart();

		om.IssueOrder(Order.Command("startgame"));
	};
	LobbyInfoChanged += lobbyReady;

	om = JoinServer(IPAddress.Loopback.ToString(), CreateLocalServer(mapUID), "");
}
```

Which is then called like:
```csharp
var orders = new[] { Order.Command("gamespeed {0}".F(gameSpeed)), Order.Command("difficulty {0}".F(difficulty)) };
Game.StartLocalGame(selectedMapPreview.Uid, orders, onStart);
```
or 
```csharp
Game.StartLocalGame(selectedMapPreview.Uid, new[] { Order.Command("sync_lobby {0}".F(lobbyInfo.Serialize()) }, onStart);
```

whinis:
Rebased and changes made to reduce code usage

OpenRA.Mods.Common/Widgets/Logic/MissionBrowserLogic.cs
```diff
@@ -300,14 +307,14 @@ void StartMissionClicked()
...
+					Game.CreateAndStartLocalServer(selectedMapPreview.Uid, orders, onStart);
 				});
 			}
 			else
-				Game.StartMission(selectedMapPreview.Uid, gameSpeed, difficulty, onStart);
+				Game.CreateAndStartLocalServer(selectedMapPreview.Uid, orders, onStart);
 		}
 
-		class DropDownOption
+class DropDownOption
```
pchote: Style nit: whitespace derp.

OpenRA.Game/Game.cs
```diff
@@ -179,8 +179,10 @@ public static void RestartGame()
 		{
 			var replay = OrderManager.Connection as ReplayConnection;
 			var replayName = replay != null ? replay.Filename : null;
-			var uid = OrderManager.World.Map.Uid;
-			var globalSettings = OrderManager.LobbyInfo.GlobalSettings;
+			var lobbyInfo = OrderManager.LobbyInfo;
+			var orders = new[] {
+					Order.Command("sync_lobby {0}".F(lobbyInfo.Serialize())),
+					Order.Command("state {0}".F(Session.ClientState.Ready)) };
```
pchote: Style nit: `};` on its own line please.

pchote:
The game will halt (not freeze, but end up in a state where you can't do anything) if you spectate an all-bot match and then try to restart.

pchote:
Two more nits that I noticed while testing, not regressions but related:

* The surrender dialog shows an "Abort" button instead of surrender.
* Offering a surrender option in singleplayer doesn't even make sense - can we disable it?

pchote:
You can't restart a skirmish game after you've lost.  Is this intentional?

whinis:
I have pushed changes to fix the spectating a bot match and the surrender text still saying abort. However as far as restarting after you have lost the skirmish match I can see that it seems to have been intentional according to https://github.com/OpenRA/OpenRA/pull/10264#issuecomment-172065515. I am not sure if it was because skirmish restarting had not yet been completed or an oversight but I can change it if that is desired.

obrakmann:
@whinis: that comment was made when restart for skirmishes didn't exist yet. Now that it does, the same mechanic as in missions should be used.

whinis:
Alright, it should now be set so that if your in a skirmish or mission and have not yet won you can restart. If you are in a replay you can restart at any time.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -144,8 +144,9 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
 				closeMenu();
 			};
 			var surrenderButton = menu.Get<ButtonWidget>("SURRENDER");
+			var surrenderVisCheck = world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined || world.Map.Visibility == MapVisibility.MissionSelector;
 			surrenderButton.IsVisible = () => world.Type == WorldType.Regular;
-			surrenderButton.IsDisabled = () => (world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined) || hasError;
+			surrenderButton.IsDisabled = () => surrenderVisCheck || hasError;
```
RoosterDragon: Moving this code out into `surrenderVisCheck` means it will be evaluated once when the menu opens but not after that. Therefore if you win/lose whilst the menu is open the button won't become disabled now. Is that intentional?

whinis:
Pushed commit to fix surrender button checking visibility only once

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -145,7 +145,11 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
 			};
 			var surrenderButton = menu.Get<ButtonWidget>("SURRENDER");
 			surrenderButton.IsVisible = () => world.Type == WorldType.Regular;
-			surrenderButton.IsDisabled = () => (world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined) || hasError;
+			surrenderButton.IsDisabled = () =>
+			{
+				var surrenderVisCheck = world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined || world.Map.Visibility == MapVisibility.MissionSelector;
```
RoosterDragon: Where did `world.Map.Visibility == MapVisibility.MissionSelector` come from?

whinis: It was talked about on IRC I believe, the surrender button shouldn't be active for missions

RoosterDragon: Ah right - that makes sense.

This might need to be `world.Map.Visibility.HasFlag(MapVisibility.MissionSelector)` - most of the other checks do it that way.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -145,7 +145,11 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
 			};
 			var surrenderButton = menu.Get<ButtonWidget>("SURRENDER");
 			surrenderButton.IsVisible = () => world.Type == WorldType.Regular;
-			surrenderButton.IsDisabled = () => (world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined) || hasError;
+			surrenderButton.IsDisabled = () =>
+			{
+				var surrenderVisCheck = world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined || world.Map.Visibility.HasFlag(MapVisibility.MissionSelector);
+				return surrenderVisCheck || hasError;
+			};
```
RoosterDragon: I think the line got too long again. Maybe try:
```c#
surrenderButton.IsDisabled = () =>
    world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined ||
    world.Map.Visibility.HasFlag(MapVisibility.MissionSelector) || hasError;
```

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameMenuLogic.cs
```diff
@@ -145,7 +145,9 @@ public IngameMenuLogic(Widget widget, World world, Action onExit, WorldRenderer
 			};
 			var surrenderButton = menu.Get<ButtonWidget>("SURRENDER");
 			surrenderButton.IsVisible = () => world.Type == WorldType.Regular;
-			surrenderButton.IsDisabled = () => (world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined) || hasError;
+			surrenderButton.IsDisabled = () =>
+				(world.LocalPlayer == null || world.LocalPlayer.WinState != WinState.Undefined ||
+				world.Map.Visibility.HasFlag(MapVisibility.MissionSelector)) || hasError;
```
RoosterDragon: Small nit: The extra brackets are not needed. `(A || B) || C` is equivalent to `A || B || C`.

RoosterDragon:
:+1:

obrakmann:
Nicely done, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/fb6087d97e29e46a8e11e530a616e953a56e2fcf)

PR #10551, opened 2016-01-19
Change CONNECTIONFAILED_PANEL title if server requires password

### LINKED ISSUE #7160: Password Protected IngameMatch-Message has wrong title
when i join a lan game which is password protected, then i got a message with title like "Connection error" the Message title should be "Password required" or "Authentification required" but not "Connection Error" like style :(

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7164409-password-protected-ingamematch-message-has-wrong-title?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Xinayder:
I've also fixed the alignment of the Retry button on the `CONNECTIONFAILED_PANEL`.

This is how the panel looks now:

Fixes #7160

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/ConnectionLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/ConnectionLogic.cs
@@ -106,6 +106,9 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			var connectionError = widget.Get<LabelWidget>("CONNECTION_ERROR");
 			connectionError.GetText = () => orderManager.ServerError;
 
+			var panelTitle = widget.Get<LabelWidget>("TITLE");
+			panelTitle.GetText = () => orderManager.AuthenticationFailed ? "Password Required" : "Connection Failed";
+
 			passwordField = panel.GetOrNull<PasswordFieldWidget>("PASSWORD");
 			if (passwordField != null)
 			{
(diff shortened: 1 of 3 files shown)


Mailaender:
If you can't connect to a server for other reasons, it still displays "password required":

image

RockyTV:
Issue is fixed.

OpenRA.Mods.Common/Widgets/Logic/ConnectionLogic.cs
```diff
@@ -106,6 +106,9 @@ public ConnectionFailedLogic(Widget widget, OrderManager orderManager, Action on
 			var connectionError = widget.Get<LabelWidget>("CONNECTION_ERROR");
 			connectionError.GetText = () => orderManager.ServerError;
 
+			var panelTitle = widget.Get<LabelWidget>("TITLE");
+			panelTitle.GetText = () => orderManager.AuthenticationFailed ? "Password Required" : "Connection Failed";
```
pchote: Can you please remove the text from the yaml file, so its only defined once here?

Phrohdoh: Wouldn't it be more sane and modder friendly to have two strings defined in YAML and referenced here?

pchote:
:+1: Looks good to me now, thanks.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7660db9f8c05d7cef0400206c2eb2a957528c94c)

PR #10554, opened 2016-01-20
Remove IReadOnlyPackage.Priority

### LINKED ISSUE #10529: Installing from CD does not work

pchote:
This removes the last of the legacy crap from the `IReadOnlyPackage` interface by replacing the hardcoded package priorities with mod-defined priority via the package order in mod.yaml.

### DIFF
--- a/OpenRA.Game/FileSystem/FileSystem.cs
+++ b/OpenRA.Game/FileSystem/FileSystem.cs
@@ -23,52 +23,50 @@ namespace OpenRA.FileSystem
 		readonly Dictionary<IReadOnlyPackage, int> mountedPackages = new Dictionary<IReadOnlyPackage, int>();
 
 		static readonly Dictionary<string, Assembly> AssemblyCache = new Dictionary<string, Assembly>();
-
-		int order;
 		Cache<string, List<IReadOnlyPackage>> fileIndex = new Cache<string, List<IReadOnlyPackage>>(_ => new List<IReadOnlyPackage>());
 
-		public IReadWritePackage CreatePackage(string filename, int order, Dictionary<string, byte[]> content)
+		public IReadWritePackage CreatePackage(string filename, Dictionary<string, byte[]> content)
 		{
 			if (filename.EndsWith(".zip", StringComparison.InvariantCultureIgnoreCase))
-				return new ZipFile(this, filename, order, content);
+				return new ZipFile(this, filename, content);
 			if (filename.EndsWith(".oramap", StringComparison.InvariantCultureIgnoreCase))
-				return new ZipFile(this, filename, order, content);
+				return new ZipFile(this, filename, content);
 
-			return new Folder(filename, order, content);
+			return new Folder(filename, content);
 		}
 
-		public IReadOnlyPackage OpenPackage(string filename, int order)
+		public IReadOnlyPackage OpenPackage(string filename)
 		{
 			if (filename.EndsWith(".mix", StringComparison.InvariantCultureIgnoreCase))
-				return new MixFile(this, filename, order);
+				return new MixFile(this, filename);
 			if (filename.EndsWith(".zip", StringComparison.InvariantCultureIgnoreCase))
-				return new ZipFile(this, filename, order);
+				return new ZipFile(this, filename);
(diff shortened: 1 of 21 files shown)


OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -103,63 +85,63 @@ public void Mount(string name, string annotation = null)
...
-			MountedPackages.Add(package);
-
-			foreach (var hash in package.ClassicHashes())
+			var mountCount = 0;
+			if (mountedPackages.TryGetValue(package, out mountCount))
 			{
-				var packageList = classicHashIndex[hash];
-				if (!packageList.Contains(package))
-					packageList.Add(package);
+				mountedPackages[package] = mountCount + 1;
```
pchote: This will also need to enumerate the package contents and bump up the per-file pacakge priority if it is not already at the end of the list.

pchote:
Adding to next release milestone so that modders can fix their package definitions and fix any case sensitivity issues in one go, instead of dragging it over multiple releases.

pchote:
Prereqs have been merged.  Rebased.

mods/cnc/mod.yaml
```diff
@@ -29,13 +18,22 @@ Packages:
...
 	desert.mix
-	snow.mix
 	~movies-gdi.mix
 	~movies-nod.mix
 	~movies.mix
 	~scores.mix
 	~scores2.mix
 	~transit.mix
+	./mods/cnc/bits
+	snow.mix
```
abcdefg30: Is there a reason behind this reordering?

abcdefg30:
Looks good to me, otherwise. :+1: 
All mods (including ra2 after a quick fix) work, too. :white_check_mark:

mods/cnc/mod.yaml
pchote: Packages listed later override packages defined earlier.  The contents of the bits directory override the contents of the content mix files, so it is listed further down the list.  `snow.mix` is shipped in the bits directory, so must be referenced after bits (otherwise it wouldn't be found).

reaperrr: This will possibly break my improved snow palette (which is in bits).
We should consider extracting the content from snow.mix into a subfolder to avoid such issues.

pchote:
Changed snow.mix to be mounted with lower priority than bits to avoid breaking the newer snow palette.

mods/cnc/mod.yaml
```diff
@@ -29,13 +18,22 @@ Packages:
...
 	winter.mix
 	desert.mix
-	snow.mix
 	~movies-gdi.mix
 	~movies-nod.mix
 	~movies.mix
 	~scores.mix
 	~scores2.mix
 	~transit.mix
+	./mods/cnc/bits/snow.mix
```
Mailaender: This is still mounted before `./mods/cnc/bits` and probably shouldn't have this prefix?

GraionDilach: That's the intention. The prefix is required otherwise the game cannot find it. See https://github.com/OpenRA/OpenRA/pull/10554#discussion_r51493376

Mailaender: You are right. I even added it myself. https://github.com/OpenRA/OpenRA/commit/a20d1c95c8c6dca8ad431992a5b2b891c97ae4dc

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a683e29df13bd5d46d3e19ac3e592155f7560835...44961affd542fa9c173a0f2272170c319e234a54)

PR #10555, opened 2016-01-20
Fix TD/RA/TS installation from CD.

### LINKED ISSUE #10529: Installing from CD does not work

pchote:
Fixes #10529 to aid with testing #10530, #10535, #10554.

### DIFF
--- a/OpenRA.Mods.Common/InstallUtils.cs
+++ b/OpenRA.Mods.Common/InstallUtils.cs
@@ -41,8 +41,7 @@ namespace OpenRA.Mods.Common
 		public static bool ExtractFromPackage(string srcPath, string package, string annotation, Dictionary<string, string[]> filesByDirectory,
 			string destPath, bool overwrite, Action<string> onProgress, Action<string> onError)
 		{
-			if (!Directory.Exists(destPath))
-				Directory.CreateDirectory(destPath);
+			Directory.CreateDirectory(destPath);
 
 			Log.Write("debug", "Mounting {0}".F(srcPath));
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/InstallUtils.cs
```diff
@@ -86,6 +86,9 @@ public static string GetMountedDisk(Func<string, bool> isValidDisk)
 		public static bool CopyFiles(string srcPath, Dictionary<string, string[]> files, string destPath,
 			bool overwrite, Action<string> onProgress, Action<string> onError)
 		{
+			if (!Directory.Exists(destPath))
```
Phrohdoh: No need to test.
[The docs say the directory will only be created if it doesn't already exist.](https://msdn.microsoft.com/en-us/library/54a0at6s%28v=vs.110%29.aspx)

pchote: Fixed.

chrisforbes:
:+1:

PR #10562, opened 2016-01-21
Implement ClassicFacingFudge for fudging classic facings.

### LINKED ISSUE #4823: Unit facings not in line with original games

pchote:
Fixes #4823.

This is a relatively invasive change, so please check carefully for regressions or any actors that I've missed fudges for.  Be sure to check both the editor and ingame.

### DIFF
--- a/OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
+++ b/OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
@@ -75,7 +75,7 @@ namespace OpenRA.Mods.Common.Graphics
 	{
 		static readonly WDist DefaultShadowSpriteZOffset = new WDist(-5);
 		readonly Sprite[] sprites;
-		readonly bool reverseFacings, transpose;
+		readonly bool reverseFacings, transpose, useClassicFacingFudge;
 
 		protected readonly ISpriteSequenceLoader Loader;
 
@@ -129,6 +129,8 @@ namespace OpenRA.Mods.Common.Graphics
 				Tick = LoadField(d, "Tick", 40);
 				transpose = LoadField(d, "Transpose", false);
 				Frames = LoadField<int[]>(d, "Frames", null);
+				useClassicFacingFudge = LoadField(d, "UseClassicFacingFudge", false);
+
 				var flipX = LoadField(d, "FlipX", false);
 				var flipY = LoadField(d, "FlipY", false);
 
@@ -138,6 +140,11 @@ namespace OpenRA.Mods.Common.Graphics
 					reverseFacings = true;
 					Facings = -Facings;
 				}
+
+				if (useClassicFacingFudge && Facings != 32)
+					throw new InvalidOperationException(
+						"{0}: Sequence {1}.{2}: UseClassicFacingFudge is only valid for 32 facings"
+						.F(info.Nodes[0].Location, sequence, animation));
 
 				var offset = LoadField(d, "Offset", float2.Zero);
 				var blendMode = LoadField(d, "BlendMode", BlendMode.Alpha);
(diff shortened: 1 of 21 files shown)


reaperrr:
You did not enable it on the RA flak truck's sequences, but it definitely looks better with it so it should be added.

Additionally, it should be enabled on RA helicopters, they have 32 facings.

Finally, `UseClassicFacingFudge: True` is not set on the rules defaults of vehicle + helicopter husks in either RA or TD, is that intentional? If not, that needs to be added as well.

As for regressions, the helicopter facing when landed on helipad is strange/wrong, but I haven't checked yet whether that's caused by this PR or on bleed already.

reaperrr:
Update on the heli landing facing:

The `InitialFacing` now needs to be `32` in RA/TD to make up for the facing fudge.
I want to point out that the originals used facing 224, though (north-east), but maybe that was intentional by whoever set that 20 in our mods due to that remapped block on the helipad's NE-edge.

pchote:
Updated.

GraionDilach:
What's the point of having this both as sequence and both as a trait flag? It kinda confuses me at the moment.

Mailaender:
It is hard for me to spot the difference, but I can't find any problems in TS/D2k and RA/TD. :white_check_mark:

reaperrr:
> What's the point of having this both as sequence and both as a trait flag? It kinda confuses me at the moment.

It's needed on sequences to make the sequence loader compile and load sequences accordingly.
And I think it's better to be able to switch it on/off per-actor as well, if only because only 2 mods need it and it's essentially just a hack to work around WW's weird rendering perspective, so I'd feel uncomfortable to have it tick away without use in D2k/TS/3rd-party mods. 
@pchote might have some other reason(s) as well.

pchote:
The trait flag is needed because there is a bunch of logic that needs to know the rendered facing of an actor (e.g. the combat code: you don't want muzzle flashes to appear in the gaps between artwork facings), but the simulation code can't explicitly depend on sequences.

GraionDilach:
Then please add a "Requires `UseClassicFacingFudge` being set on the actor's sequences as well." note to the `BodyOrientation` flag's documentation to make sure modders get it right (since sequence docs ain't generated at the moment).

pchote:
The issue is that no such requirement exists, and I can imagine situations where mods may want the tag on the BodyOrientation but not in the sequences (e.g. a FTUR-like structure that has a single fixed frame, but needs its muzzles to match up with the perspective of the artwork).

reaperrr:
`InitialFacing` of GUN turret in RA and TD needs to be increased from 50 to 64 to match the make anim.

:+1: after that.

pchote:
Changed gun turret facings to 56 (64 was too far in the other direction).

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/6e458842ac27f62eae884c7187c4cb6d4892d1e3...8c7f09d79f592c19c4a2d7180fd5ca24c33c05d1)

PR #10563, opened 2016-01-22
Fixes Obelisk not attacking again after a forced attack command

### LINKED ISSUE #10165: TS obelisk only shoots once when force-firing
When ordering an obelisk (in TS) to shoot at an allied actor via force-fire, it will stop after 1 shot. Does not happen with TD obelisk, so it's probably a yaml rule issue.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28796563-ts-obelisk-only-shoots-once-when-force-firing?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

whinis:
This PR should fix #10165,
Reload delay was set to 120 so the obelisk would wait that long before charging again. Another attack move would overwrite this request explaining the behavior

### DIFF
--- a/mods/ts/rules/nod-support.yaml
+++ b/mods/ts/rules/nod-support.yaml
@@ -95,8 +95,7 @@ NAOBEL:
 		LocalOffset: 1400,210,800
 	AttackCharge:
 		ChargeAudio: obelpowr.aud
-		ReloadTime: 50
-		InitialChargeDelay: 50
+		InitialChargeDelay: 120
 	WithChargeOverlay:
 		Sequence: active
 		Palette: player
--- a/mods/ts/weapons/energyweapons.yaml
+++ b/mods/ts/weapons/energyweapons.yaml
@@ -171,7 +171,7 @@ Proton:
 		ValidImpactTypes: Water
 
 ObeliskLaserFire:
-	ReloadDelay: 120
+	ReloadDelay: 1
 	Range: 10c512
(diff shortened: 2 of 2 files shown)


Mailaender:
:white_check_mark:

abcdefg30:
I think you should also increase the [charge delays](https://github.com/OpenRA/OpenRA/blob/77b0bd60671ac6f84cff52e29e313f2e04c2388d/mods/ts/rules/nod-support.yaml#L98) then, as the original had `ROF=120`.

whinis:
Both have been changed, however now the animation seems funky on it

mods/ts/rules/nod-support.yaml
```diff
@@ -95,8 +95,8 @@ NAOBEL:
 		LocalOffset: 1400,210,800
 	AttackCharge:
 		ChargeAudio: obelpowr.aud
-		ReloadTime: 50
-		InitialChargeDelay: 50
+		ReloadTime: 120
```
abcdefg30: This is already the default value.

abcdefg30:
Looks good to me now, too. :+1: / :white_check_mark:

PR #10564, opened 2016-01-22
Added support for playing a "Our ally is under attack" notification

abcdefg30:
Also tried to improve performance by moving a few checks around.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Player/BaseAttackNotifier.cs
+++ b/OpenRA.Mods.Common/Traits/Player/BaseAttackNotifier.cs
@@ -47,10 +47,6 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void Damaged(Actor self, AttackInfo e)
 		{
-			// only track last hit against our base
-			if (!self.Info.HasTraitInfo<BuildingInfo>())
-				return;
-
 			if (e.Attacker == null)
 				return;
 
@@ -58,6 +54,10 @@ namespace OpenRA.Mods.Common.Traits
 				return;
 
 			if (e.Attacker == self.World.WorldActor)
+				return;
+
+			// Only track last hit against our base
+			if (!self.Info.HasTraitInfo<BuildingInfo>())
 				return;
 
 			if (e.Attacker.Owner.IsAlliedWith(self.Owner) && e.Damage <= 0)
--- a/OpenRA.Mods.Common/Traits/Player/BaseAttackNotifier.cs
+++ b/OpenRA.Mods.Common/Traits/Player/BaseAttackNotifier.cs
@@ -28,6 +28,10 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("The audio notification type to play.")]
 		public string Notification = "BaseAttack";
 
+		[Desc("The audio notification to play to allies when under attack.",
+			"Won't play a notification to allies if this is null.")]
+		public string AllyNotification = null;
+
 		public object Create(ActorInitializer init) { return new BaseAttackNotifier(init.Self, this); }
 	}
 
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.Common/Traits/Player/BaseAttackNotifier.cs
```diff
@@ -60,12 +60,27 @@ public void Damaged(Actor self, AttackInfo e)
...
 			if (e.Attacker.Owner.IsAlliedWith(self.Owner) && e.Damage <= 0)
 				return;
 
 			if (self.World.WorldTick - lastAttackTime > info.NotifyInterval * 25)
 			{
-				Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Speech", info.Notification, self.Owner.Faction.InternalName);
+				var rules = self.World.Map.Rules;
+				Game.Sound.PlayNotification(rules, self.Owner, "Speech", info.Notification, self.Owner.Faction.InternalName);
+
+				if (info.AllyNotification != null)
```
Phrohdoh: This should probably be `string.IsNullOrWhitespace`.

Mailaender: I sometimes use is `string.IsNullOrEmpty`, but I don't think it is required here. Whitespaces are stripped from the fieldloader so that shouldn't cause a problem.

OpenRA.Mods.Common/Traits/Player/HarvesterAttackNotifier.cs
```diff
@@ -47,12 +47,16 @@ public HarvesterAttackNotifier(Actor self, HarvesterAttackNotifierInfo info)
...
-			// only track last hit against our harvesters
-			if (!self.Info.HasTraitInfo<HarvesterInfo>())
+			// Don't track self-damage
+			if (e.Attacker != null && e.Attacker.Owner == self.Owner)
 				return;
 
-			// don't track self-damage
-			if (e.Attacker != null && e.Attacker.Owner == self.Owner)
+			// This player won't hear us anyway
+			if (!self.Owner.Playable)
```
Mailaender: This will disable radar pings when watching replays of bot matches.

OpenRA.Mods.Common/Traits/Player/BaseAttackNotifier.cs
```diff
@@ -60,12 +60,27 @@ public void Damaged(Actor self, AttackInfo e)
...
+					foreach (Player p in self.World.Players)
```
Mailaender: You need to exclude `self.Owner` here, otherwise it will also overlap the `BaseAttack` notification above.

OpenRA.Mods.Common/Traits/Player/HarvesterAttackNotifier.cs
abcdefg30: I think I'll just drop this, there are no real benefits and this just causes such trouble.

OpenRA.Mods.Common/Traits/Player/BaseAttackNotifier.cs
abcdefg30: Done.

abcdefg30:
Updated.

Mailaender:

OpenRA.Mods.Common/Traits/Player/BaseAttackNotifier.cs
```diff
@@ -60,12 +60,22 @@ public void Damaged(Actor self, AttackInfo e)
...
+						if (p != self.Owner && p.IsAlliedWith(self.Owner))
```
obrakmann: Do we want to play the notification to the player that is attacking, too (in case of friendly fire, which might either be deliberate or not)?

If yes, then :+1:, otherwise you'd need to compare `p` with `e.Attacker.Owner` here as well.

I haven't quite formed an opinion yet on what I'd prefer.

abcdefg30: Added the check now.

abcdefg30:
Updated.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/3eb9b4a2a04f54c11b7f66412b982e6912702b53)

PR #10570, opened 2016-01-23
Add support for playing an InsufficientFunds warning

abcdefg30:
Also enables those warnings in ts and ra.

### DIFF
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -299,6 +299,7 @@
     <Compile Include="Traits\Burns.cs" />
     <Compile Include="Traits\C4Demolition.cs" />
     <Compile Include="Traits\Health.cs" />
+    <Compile Include="Traits\Player\InsufficientFundsWarning.cs" />
     <Compile Include="Traits\PowerTooltip.cs" />
     <Compile Include="Traits\VeteranProductionIconOverlay.cs" />
     <Compile Include="Traits\Capturable.cs" />
--- a/OpenRA.Mods.Common/Traits/Player/InsufficientFundsWarning.cs
+++ b/OpenRA.Mods.Common/Traits/Player/InsufficientFundsWarning.cs
@@ -1,0 +1,53 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Provides the player with an audible warning when they run out of money while producing.")]
+	public class InsufficientFundsWarningInfo : ITraitInfo, Requires<PlayerResourcesInfo>
+	{
+		[Desc("The speech to play for the warning.")]
+		public readonly string Notification = "InsufficientFunds";
+
+		public object Create(ActorInitializer init) { return new InsufficientFundsWarning(this); }
+	}
+
+	public class InsufficientFundsWarning : INotifyInsufficientFunds
+	{
+		readonly InsufficientFundsWarningInfo info;
+
+		bool played;
+
+		public InsufficientFundsWarning(InsufficientFundsWarningInfo info)
+		{
+			this.info = info;
+		}
+
+		void INotifyInsufficientFunds.InsufficientFunds(Actor self)
+		{
+			Game.RunAfterTick(() =>
+			{
+				if (played)
+					return;
+
+				played = true;
+				var owner = self.Owner;
+				Game.Sound.PlayNotification(self.World.Map.Rules, owner, "Speech", info.Notification, owner.Faction.InternalName);
+			});
+		}
+
+		void INotifyInsufficientFunds.SufficientFunds(Actor self)
+		{
+			played = false;
+		}
+	}
+}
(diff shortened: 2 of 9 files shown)


OpenRA.Mods.Common/Traits/Player/InsufficientFundsWarning.cs
```diff
@@ -0,0 +1,57 @@
...
+
+		public object Create(ActorInitializer init) { return new InsufficientFundsWarning(init.Self, this); }
+	}
+
+	public class InsufficientFundsWarning : ITick
+	{
+		readonly InsufficientFundsWarningInfo info;
+		readonly PlayerResources resources;
+
+		int nextWarnTime = 0;
```
Phrohdoh: This needs to be set to `info.WarnInterval` in the ctor otherwise the condition checking this will be true immediately.

pchote:
I'm pretty sure that the original games didn't set this up as a repeating warning, which is going to be quite annoying for people who play the game efficiently (running with a very low float).

MustaphaTR:
In orginal game InsufficientFunds plays at two times, when you go out of credits, and when you try to build something when you are out of credits.  They doesn't repeat it unless you gain some credits. and lost it again iirc.

abcdefg30:
Updated. It plays now when you reach zero money while trying to produce something.

OpenRA.Mods.Common/Traits/Player/InsufficientFundsWarning.cs
```diff
@@ -0,0 +1,56 @@
...
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Provides the player with an audible warning when their storage is nearing full.")]
```
Mailaender: Copy pasta error.

abcdefg30: Ops, will fix.

abcdefg30:
Fixed the typo.

OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -82,6 +82,11 @@ public interface IProductionIconOverlay
 		bool IsOverlayActive(ActorInfo ai);
 	}
 
+	public interface INotifyInsufficientFunds
```
pchote: Please add `[RequireExplicitImplementation]`

OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -82,6 +82,11 @@ public interface IProductionIconOverlay
...
+	{
+		void InsufficientFunds();
```
pchote: This should accept an `Actor self` parameter for consistency with other interfaces.

abcdefg30:
Updated.

OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
```diff
@@ -745,4 +746,4 @@ cd "$(SolutionDir)"</PostBuildEvent>
   <Target Name="AfterBuild">
   </Target>
   -->
-</Project>
+</Project>
```
pchote: Files should end with newlines.  Please drop this change.

abcdefg30:
Fixed again.

Unit158:
This in d2k as well.

Mailaender:
> I'm pretty sure that the original games didn't set this up as a repeating warning, which is going to be quite annoying for people who play the game efficiently (running with a very low float).

This still overlaps with "Building". I also think it should replace it with "Insufficient Funds" and don't repeat the message.

Mailaender:
That is more like the original now and less annoying. :white_check_mark:

obrakmann:
I see two problems with this:

1. If you're building from multiple queues at the same time, you get a notification from each queue.
2. If you queue other items while production is ongoing, you no longer get the queued notification.

abcdefg30:
Sorry guys, this wasn't actually completed yet (which is why I didn't leave an "Updated" comment).

I already had a small talk with @pchote on IRC to fix the issue @obrakmann pointed out.
Seems like I have to get back to interfaces for it, sorry @Mailaender,

abcdefg30:
Updated and rebased.

> 2. If you queue other items while production is ongoing, you no longer get the queued notification.

You'll get the notification now when it actually starts buildings. This is a side effect from fixing "This still overlaps with "Building!"" (https://github.com/OpenRA/OpenRA/pull/10570#issuecomment-174271390). Not sure if we want to call that an improvement. If it is unwanted, I'll just drop the change entirely and then we got the overlapping (except someone comes up with a good idea).

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -458,7 +471,23 @@ public void Tick(PlayerResources pr)
...
+				}
+
 				return;
+			}
+
+			if (insufficientFundsPlayed)
+			{
+				insufficientFundsPlayed = false;
+				foreach (var funds in insufficientFunds)
+					funds.SufficientFunds(owner.PlayerActor);
```
obrakmann: You need to move the `foreach` out of this block. Otherwise you won't get a warning in all circumstances. Ie. if you have $300 left and try to build a helipad, you'll get a warning. If you then build a E1 the build succeeds, which should reset the warning. However, if you try to build an E3 after the E1, the warning will not appear. Moving the `foreach` will fix that.

abcdefg30:
Updated.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/16ccc0d5e736b44aec0e83a4eb1a131c1c2f8a68)

richlieb:
InsufficientFunds warning can be triggered too rapidly in the current playtest version of Dune (OpenRA-playtest-20160403).

For example your funds are at 100, your harvester is restocking your funds and you produce a building, a unit and a vehicle. The harvester raises your funds above 100 and in the next moment your funds get reduced below 100 due to your production and every time your funds drop below 100 the notifications gets triggered

As a result the notification sound overlaps, creates a kind of unpleasent acoustic canon.

Issue #9678, opened 2015-10-19
Weird visual 'drop' when vehicles exit factory in TS


reaperrr:
Only noticable in isometric perspective. They suddenly 'drop' by a few pixels when they move past a certain point after exiting the factory. Applies to voxel and sprite actors alike, so probably not just a visual issue.

teees:
Looks like this: http://gfycat.com/PresentConsiderateGreendarnerdragonfly

teees:
Probably related: actors exiting wf clipping through other actors

http://gfycat.com/InfiniteCreepyHagfish

Issue #10565, opened 2016-01-23
Stuck AI harvesters can be a problem


Mailaender:
This happens in TS/RA2 where the AI stacks building too close together or builds a refinery which exits to a cliff as the bibs are solely decoration. Harvesters failing to path constantly is expensive.

![image](https://cloud.githubusercontent.com/assets/756669/12529587/ce325cb0-c1bc-11e5-9a30-05043d5ddf80.png)
##

pchote:
The exit/docking cells aren't set up properly for basically all the buildings in TS.  See also #9678 and the stream where fiveaces complains about barracks being blocked by cliffs.

reaperrr:
Additionally, our current AI refinery placement logic places the ref in the direction of resources if possible, but as close to the base center (c. yard) as possible, as visible in the above screenshot. Would be nice if we could add some minimum adjacency from base center to this.

PR #10040, opened 2015-11-20
Move weapons factory exit and waypoint closer to the building

teees:
Imgur

Fixes this glitch: http://gfycat.com/InfiniteCreepyHagfish

The visual drop in #9678 still persists though

PR #10573, opened 2016-01-23
Adjusted the Tiberian Sun harvester spawn point

Mailaender:
so it becomes created at the docking position.

### DIFF
--- a/mods/ts/rules/shared-structures.yaml
+++ b/mods/ts/rules/shared-structures.yaml
@@ -85,7 +85,7 @@ PROC:
 		Value: 600
 	FreeActor:
 		Actor: HARV
-		SpawnOffset: 3,1
+		SpawnOffset: 2,1
 		Facing: 160
 	WithIdleOverlay@REDLIGHTS:
 		Sequence: idle-redlights


abcdefg30:
Looks good to me. Nice polish. :+1: / :white_check_mark:

mods/ts/rules/gdi-structures.yaml
```diff
@@ -142,12 +142,12 @@ GAWEAP:
 	RevealsShroud:
 		Range: 4c0
 	RallyPoint:
-		Offset: 4,1
+		Offset: 3,1
```
pchote: These were already set to the correct value in #10040.  We tested and showed that this does not fix #9678.

abcdefg30: Somehow I didn't notice #9678 on this pr.

Mailaender:
Removed https://github.com/Mailaender/OpenRA/commit/4fa7ac15fe0f3d85326d91c66978003712eeb6c9 and only kept the harvester spawn.

mods/ts/rules/gdi-structures.yaml
abcdefg30: Ah never mind, it is still there but more hidden.

Mailaender: You were giving misleading instructions at https://github.com/OpenRA/OpenRA/issues/10565#issuecomment-174167375 then, @pchote.

obrakmann:
yea, that looks better. :+1:

PR #10578, opened 2016-01-24
Directional armor

### LINKED ISSUE #9268: Implement support for directional armor
A possible, relatively simple approach would be to compare the projectile facing right before explosion with the facing of directly hit actors (explosion happened inside their health radius / hit box / whatever).

If the projectile facing exactly matched or diverted by no more than X degrees, it must have hit the weak rear armor, giving a damage bonus.
If the facing was the opposite, it must have hit the front, reducing the damage.
In between it must have hit the sides, so apply normal damage.

This is just a possibility. If we had a raycaster and rectangular hit boxes, it would be possible to check more accurately what side was hit.

Unit158:
Closes #9268

### DIFF
--- a/OpenRA.Mods.Common/Traits/DirectionalArmor.cs
+++ b/OpenRA.Mods.Common/Traits/DirectionalArmor.cs
@@ -1,0 +1,71 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Used to define weapon efficiency modifiers with different percentages per Type.")]
+	public class DirectionalArmorInfo : UpgradableTraitInfo, Requires<HealthInfo>
+	{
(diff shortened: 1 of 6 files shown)


OpenRA.Mods.Common/Traits/DirectionalArmor.cs
```diff
@@ -0,0 +1,71 @@
...
+
+		public string GetArmor(Actor self, Actor firedBy, WPos impact)
+		{
+			var healthInfo = self.Info.TraitInfo<HealthInfo>();
+
+			var vec = impact - self.CenterPosition;
+
+			// If the impact position is within any actor's HitShape, we have a direct hit.
+			if ((vec).LengthSquared <= healthInfo.Shape.DistanceFromEdge(impact, self).LengthSquared)
+				vec = firedBy.CenterPosition - self.CenterPosition;
```
pchote: If the projectile was a missile then this may not be at all related to the angle that it hits from.

Unit158: This (and most other dumb cases) should be solvable if I can get the original target info, so I'll see if I can just pass that along and play with it instead in that case.

pchote: Lets fix that first (in its own prs), so that we can do this part properly.

Unit158:
This will require some modifications to various projectiles, so I'll close this until then.

Leo512bit:
Can this be reworked on? I want to see this in, considering that it is 2022 has the projectile issues been solved?

PR #10579, opened 2016-01-24
Added support for the Kate text editor

Mailaender:
See http://kate-editor.org/2012/11/02/using-the-projects-plugin-in-kate/ for reference. I should have set this up way earlier. It supports project wide search (and replace) which is now quite necessary after the rules have been split up into many different files and adds other convenient features such as build commands. Not sure if this is helpful for anyone else. I may be the only user, but on the other hand this JSON file is hidden, small and won't need any maintenance.

image

### DIFF
--- a/.kateproject
+++ b/.kateproject
@@ -1,0 +1,9 @@
+{
+  "name": "OpenRA"
+, "files": [ { "git": 1 } ]
+, "build": {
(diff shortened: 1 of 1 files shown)


obrakmann:
Taking your word for it that it works. :+1:

PR #10580, opened 2016-01-24
Add a "no superweapons" tech level to RA.

pchote:
This addresses one of the biggest remaining complaints/requests that we see from casual players scattered over our various feedback channels.

The new "No Superweapons" options disables the three superweapon structures in the game: Missile Silo, Iron Curtain, Chronosphere.  It does _not_ remove any of the support powers (GPS, parabombs, etc), nor does it attempt to address the balance of the chrono tank (which is implicitly removed) or the demo truck (which isn't).

### DIFF
--- a/mods/ra/chrome/lobby-options.yaml
+++ b/mods/ra/chrome/lobby-options.yaml
@@ -54,16 +54,16 @@ Background@LOBBY_OPTIONS_BIN:
 					Height: 20
 					Text: Debug Menu
 				Label@DIFFICULTY_DESC:
-					X: PARENT_RIGHT - WIDTH - 145
+					X: PARENT_RIGHT - WIDTH - 165
 					Y: 70
-					Width: 120
+					Width: 160
 					Height: 25
 					Text: Mission Difficulty:
 					Align: Right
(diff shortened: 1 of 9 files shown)


pchote:
Please note (if the emphasis above wasn't enough) that this is specifically for our casual players who usually just want to turtle up against an AI without getting a nuke in the face.  I've intentionally restricted the scope of this pr to try and avoid following the same derailment over balance, GPS, etc that always happens whenever this topic is discussed.

obrakmann:
I expect we will receive bug reports over the balance, but we already do for both the balance and the missing techlevel option, so I'm fine with it. :+1:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/e1345029f195eddef5cc13df720becdcf77e3765...786e546b3ed119e1f11df606606c8f7e8daefe73)

pi99y:
Would it be perhaps just a bit better to have the Chronosphere building still buildable - granting access just to Chrono tanks (=less change) but not the Chronosphere superpower? Or is that much harder to do and therefore not worth it since it's aimed just towards the casual players?

PR #10458, opened 2016-01-09
Added Tiberian Sun map Cliffs of Insanity

Mailaender:
Classic map that already works as it doesn't have any bridges. Test case for https://github.com/OpenRA/OpenRA/pull/10397.

http://resource.openra.net/maps/8377

PR #10583, opened 2016-01-24
Added original TS maps Tactical and Terraces

Mailaender:
Both maps are very open which is good for gameplay. They also don't have bridges.
- http://resource.openra.net/maps/8526/
- http://resource.openra.net/maps/8525/


reaperrr:
> Both maps are very open which is good for gameplay. They also don't have bridges.

Should be good for performance too, wrt path-finding and AI.

Mailaender:
Indeed. https://github.com/OpenRA/OpenRA/pull/10458 for example is very choked by cliffs and trees.

obrakmann:
I'm not sure if the tiberium distribution on either map is balanced fairly, but I guess Westwood knew what they were doing. :+1:

abcdefg30:
:+1:

PR #10590, opened 2016-01-24
Add ReloadAmmo trait and remove Rearm activity

reaperrr:
This is the first step towards getting rid of `Rearm`/`Repair` activities in favor of upgrade-based solutions for increased flexibility.

Allows actors to be rearmed and repaired at the same time.

This also moves the `RearmBuildings` property from `Aircraft` and `Minelayer` to `ReloadAmmo`.

Supercedes #9606.


abcdefg30:
Needs a rebase already?

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -450,12 +450,15 @@ bool HasAdequateAirUnitReloadBuildings(ActorInfo actorInfo)
 			if (aircraftInfo == null)
 				return true;
 
-			var ammoPoolsInfo = actorInfo.TraitInfos<AmmoPoolInfo>();
+			// We assume that UpgradeMinEnabledLevel = 0 means the actor reloads ammo by itself.
+			// To have ReloadAmmo also implies having at least 1 AmmoPool.
+			var buildingReload = actorInfo.TraitInfos<ReloadAmmoInfo>()
+				.Where(x => x.UpgradeMinEnabledLevel > 0).ToList();
```
RoosterDragon: You could use `.FirstOrDefault()` instead of `ToList()` since you only care about the first result.

reaperrr:
Updated.

Now depends on #10630 to prevent helicopters from reloading while still in the air.

reaperrr:
Updated.

reaperrr:
By now I'm thinking about waiting for (or implementing it myself) some UpgradeDockedActor trait instead of using UpgradeActorsNear, so I'm closing this for now to clear the queue.

Issue #10460, opened 2016-01-09
IndexOutOfRangeException in FootprintBounds


steadyfield:
```
Tiberian Sun Mod at Version {DEV_VERSION}
on map f302316e85c754d108fc1c3d79993e90a0faf4e1 (Tiberium Garden Redux by Westwood Studios).
Operating System: Windows (Microsoft Windows NT 6.2.9200.0)
Runtime Version: .NET CLR 4.0.30319.34209
Exception of type `System.IndexOutOfRangeException`:
   at OpenRA.Traits.FrozenActorLayer.FootprintBounds(FrozenActor fa) in d:\git\OpenRA-bleed-1-9\OpenRA.Game\Traits\Player\FrozenActorLayer.cs:Line 194
   at OpenRA.Traits.FrozenActorLayer.Add(FrozenActor fa) in d:\git\OpenRA-bleed-1-9\OpenRA.Game\Traits\Player\FrozenActorLayer.cs:Line 189
   at OpenRA.Mods.Common.Traits.FrozenUnderFog.<>c__DisplayClass5.<Created>b__4(Player player, Int32 playerIndex)
   at OpenRA.Primitives.PlayerDictionary`1..ctor(World world, Func`3 valueFactory) in d:\git\OpenRA-bleed-1-9\OpenRA.Game\Primitives\PlayerDictionary.cs:Line 27
   at OpenRA.Mods.Common.Traits.FrozenUnderFog.Created(Actor self)
   at OpenRA.World.CreateActor(Boolean addToWorld, String name, TypeDictionary initDict) in d:\git\OpenRA-bleed-1-9\OpenRA.Game\World.cs:Line 232
   at OpenRA.World.CreateActor(String name, TypeDictionary initDict) in d:\git\OpenRA-bleed-1-9\OpenRA.Game\World.cs:Line 225
   at OpenRA.Mods.Common.Traits.PlaceBuilding.<>c__DisplayClass6.<ResolveOrder>b__2(World w)
   at OpenRA.World.Tick() in d:\git\OpenRA-bleed-1-9\OpenRA.Game\World.cs:Line 309
   at OpenRA.Game.InnerLogicTick(OrderManager orderManager) in d:\git\OpenRA-bleed-1-9\OpenRA.Game\Game.cs:Line 520
   at OpenRA.Game.LogicTick() in d:\git\OpenRA-bleed-1-9\OpenRA.Game\Game.cs:Line 545
[...]

PR #10592, opened 2016-01-24
Debug frozen actors without footprints

RoosterDragon:
Intended to provide more information the next time #10460 occurs.

### DIFF
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -64,6 +64,19 @@ namespace OpenRA.Traits
 				.Where(m => shroud.Contains(m))
 				.ToArray();
 
+			if (Footprint.Length == 0)
+				throw new ArgumentException(("This frozen actor has no footprint! Please report a bug!\n" +
+					"Actor Name: {0}\n" +
+					"Actor Location: {1}\n" +
+					"Input footprint: [{2}]\n" +
+					"Input footprint (shroud.Contains): [{3}]\n" +
+					"Input footprint (map.Contains): [{4}]")
+					.F(actor.Info.Name,
+					actor.Location.ToString(),
+					footprint.Select(p => p.ToString()).JoinWith("|"),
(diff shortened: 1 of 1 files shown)


abcdefg30:
Looks simple enough. :+1:

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -64,6 +64,18 @@ public FrozenActor(Actor self, PPos[] footprint, Shroud shroud, bool startsRevea
 				.Where(m => shroud.Contains(m))
 				.ToArray();
 
+			if (Footprint.Length == 0)
+				throw new ArgumentException(("This frozen actor has no footprint! Please report a bug!\n" +
+					"Map: {0}\nActor Location: {1}\n" +
+					"Input footprint: [{2}]\n" +
+					"Input footprint (shroud.Contains): [{3}]\n" +
+					"Input footprint (map.Contains): [{4}]")
+					.F(actor.World.Map.Title,
```
Mailaender: This is probably redundant as we add it to every stacktrace, but the actor name would be interesting. I assume this could be bogus actor definition which we then could catch with lint checks.

RoosterDragon:
Updated per @Mailaender's suggestion.

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -64,6 +64,18 @@ public FrozenActor(Actor self, PPos[] footprint, Shroud shroud, bool startsRevea
 				.Where(m => shroud.Contains(m))
 				.ToArray();
 
+			if (Footprint.Length == 0)
+				throw new ArgumentException(("This frozen actor has no footprint! Please report a bug!\n" +
+					"Actor Name: {0}\nActor Location: {1}\n" +
```
Phrohdoh: Consistency nit-pick: The location string should go on its own line as another concat.

obrakmann:
:+1:

PR #10599, opened 2016-01-25
Updated to SDL 2.0.4 for Windows

Mailaender:
Changelog: https://forums.libsdl.org/viewtopic.php?t=11711

The PowerShell script hasn't been tested so this probably won't work on first try.

### DIFF
--- a/thirdparty/fetch-thirdparty-deps-windows.sh
+++ b/thirdparty/fetch-thirdparty-deps-windows.sh
@@ -18,10 +18,10 @@ function get()
 }
 
 if [ ! -f SDL2.dll ]; then
-	echo "Fetching SDL2 from NuGet"
-	get sdl2.redist 2.0.3
-	cp ./sdl2.redist/build/native/bin/Win32/dynamic/SDL2.dll .
-	rm -rf sdl2.redist
+	echo "Fetching SDL2 from libsdl.org"
+	wget https://www.libsdl.org/release/SDL2-2.0.4-win32-x86.zip
+	unzip SDL2-2.0.4-win32-x86.zip SDL2.dll
+	rm SDL2-2.0.4-win32-x86.zip
 fi
 
 if [ ! -f freetype6.dll ]; then
--- a/thirdparty/fetch-thirdparty-deps.ps1
+++ b/thirdparty/fetch-thirdparty-deps.ps1
@@ -69,11 +69,13 @@ if (!(Test-Path "nunit.framework.dll"))
 
 if (!(Test-Path "windows/SDL2.dll"))
 {
-	echo "Fetching SDL2 from NuGet."
-	./nuget.exe install sdl2 -Version 2.0.3 -ExcludeVersion
-	cp sdl2.redist/build/native/bin/Win32/dynamic/SDL2.dll ./windows/
-	rmdir sdl2 -Recurse
-	rmdir sdl2.redist -Recurse
+	echo "Fetching SDL2 from libsdl.org"
+	$target = Join-Path $pwd.ToString() "SDL2-2.0.4-win32-x86.zip"
+	(New-Object System.Net.WebClient).DownloadFile("https://www.libsdl.org/release/SDL2-2.0.4-win32-x86.zip", $target)
+	$destination = Join-Path $pwd.ToString() "windows"
+	Add-Type -assembly "system.io.compression.filesystem"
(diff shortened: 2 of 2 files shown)


thirdparty/fetch-thirdparty-deps-windows.sh
```diff
@@ -19,9 +19,9 @@ function get()
 
 if [ ! -f SDL2.dll ]; then
 	echo "Fetching SDL2 from NuGet"
-	get sdl2.redist 2.0.3
-	cp ./sdl2.redist/build/native/bin/Win32/dynamic/SDL2.dll .
-	rm -rf sdl2.redist
+	wget https://www.libsdl.org/release/SDL2-2.0.4-win32-x86.zip
```
Phrohdoh: This is no longer pulling from a NuGet feed.

pchote: It's pulling from the official website instead, which is *much* better than a random unverified nuget upload.

pchote:
Can you please modify the commit message to specify that this only affects Windows? OSX and Linux are not updated to 2.0.4 by this pr.

thirdparty/fetch-thirdparty-deps-windows.sh
Phrohdoh: `echo "Fetching SDL2 from NuGet"` is no longer true.

Mailaender:
Updated.

thirdparty/fetch-thirdparty-deps.ps1
```diff
@@ -69,11 +69,11 @@ if (!(Test-Path "nunit.framework.dll"))
 
 if (!(Test-Path "windows/SDL2.dll"))
 {
-	echo "Fetching SDL2 from NuGet."
-	./nuget.exe install sdl2 -Version 2.0.3 -ExcludeVersion
-	cp sdl2.redist/build/native/bin/Win32/dynamic/SDL2.dll ./windows/
-	rmdir sdl2 -Recurse
-	rmdir sdl2.redist -Recurse
+	echo "Fetching SDL2 from libsdl.org"
+	$target = Join-Path $pwd.ToString() "SDL2-2.0.4-win32-x86.zip"
```
Phrohdoh: `$target = [IO.Path]::Combine($PWD, "SDL2-2.0.4-win32-x86.zip")` does what you actually want here.

Mailaender: Are you sure this is failing it, because the download works fine and this is just what we do below.

Phrohdoh: Odd. You are correct.

When I tried this vs what I suggested earlier today this didn't work (it resulted in just the value of `$PWD`), but now it works just fine.

What is failing is:
```
File: "C:\projects\openra\thirdparty\download\windows\SDL2.dll" -> no files found.
Usage: File [/nonfatal] [/a] ([/r] [/x filespec [...]] filespec [...] |
   /oname=outfile one_file_only)
Error in script "packaging/windows/OpenRA.nsi" on line 112 -- aborting creation process
Command exited with code 1
```

Should you be using `7z e SDL2-2.0.4-win32-x86.zip -o .\windows SDL2.dll` instead of `7z e SDL2-2.0.4-win32-x86.zip -o.\windows SDL2.dll`?

The difference being a space after `-o`.

Mailaender: It won't work with a space. The syntax is extremely weird. https://sevenzip.osdn.jp/chm/cmdline/switches/output_dir.htm

Mailaender:
Finally fixed the Windows build. Now you won't need 7-zip installed. It uses pure PowerShell syntax. blogs.technet.com/b/heyscriptingguy/archive/2015/03/11/use-powershell-to-extract-zipped-files.aspx

pchote:
Confirmed that the fetch-thirdparty-deps-windows.sh script still works, but I can't test the rest of this.

abcdefg30:
Looks good to me. :+1: 
(Although Windowed mode with 0,0 as size renders the same as fullscreen now.)

RoosterDragon:
:+1:

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/97d1a4300af42442595b7e42fd0f4f67da2373ad)

PR #10602, opened 2016-01-25
Fix TD aircraft being visible through fog.

pchote:
The fog represents the lack of visibility, so seeing things move under it does not make sense (this is the same logic we use for the Carryalls in D2K).  This also looks glitchy, as the contrail rendering is disabled under the fog.

### DIFF
--- a/mods/cnc/rules/defaults.yaml
+++ b/mods/cnc/rules/defaults.yaml
@@ -398,6 +398,7 @@
 		UseLocation: yes
 	HiddenUnderFog:
 		Type: CenterPosition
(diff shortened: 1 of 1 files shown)


reaperrr:
:+1:

obrakmann:
Fine with me, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9ade0dfdf2676eb16b16ed582e6ac18ce40837d0)

PR #10603, opened 2016-01-25
Fixed AppVeyor pull request artifact suffixes

Mailaender:
`OpenRA-.exe` to `OpenRA-11455c8.exe` if everything goes right.


pchote:
> OpenRA-8c197826b18f16137a3cc135829f243592be84db.exe

Can we trim this down to the first 7 characters?

pchote:
Actually, this doesn't match the commit hash at all?

obrakmann:
That hash is not the PR's commit, but the resulting merge commit from merging this PR into the then-HEAD of bleed: https://github.com/OpenRA/OpenRA/commit/8c197826b18f16137a3cc135829f243592be84db

I'm fine with keeping the full hash in the name, too, we rarely ever use that installer anyway, and shortening the name would break it on tagged builds. :+1:

Mailaender:
Updated. I can simply make it just trim the commit hash. The regular tags will be untouched.

pchote:
Looks fine here now. :+1:

PR #10608, opened 2016-01-26
Enable Chrono Tank at "No Superweapons"

reaperrr:
Enables Germany's Chrono Tank at "No Superweapons" by removing Chronosphere from prerequisites and lowering tech-level requirement to 'high'.

### DIFF
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -656,7 +656,7 @@ CTNK:
 	Buildable:
 		Queue: Vehicle
 		BuildPaletteOrder: 210
-		Prerequisites: atek, pdox, ~vehicles.germany, ~techlevel.unrestricted
+		Prerequisites: atek, ~vehicles.germany, ~techlevel.high
 	Valued:
 		Cost: 1350
 	Tooltip:


MustaphaTR:
> personally I'd vote for keeping parabombs and disabling demo trucks.

Agreed. Keep parabombs, disable Demo Trucks. As chrono tank requires Chronoshphere, thry already disabled in No SW, at least removing Demo Truck balances it.

pchote:
> giving it a potentially huge advantage over other factions at this TL when it comes to dealing damage.

I don't agree here: Russia and Ukraine are restricted in the same way, and disabling 2/3 of Ukraines faction specials (vs none of Russia's) puts them at a significant disadvantage.

obrakmann:
:+1: For disabling demo trucks. For the parabombs, maybe a compromise could be to decrease the squad size to 1 or 2 (I'd probably go with 2) aircraft instead of keeping all three.

reaperrr:
> For the parabombs, maybe a compromise could be to decrease the squad size to 1 or 2 (I'd probably go with 2) aircraft instead of keeping all three.

If we go down that route, we'll have to introduce a "high-only" tech level that is not included in 'unrestricted', right?

In that case, we could also make a drastically nerfed (but also cheaper) "no nuke, just normal explosives" demo truck available exclusively at that tech level.

reaperrr:
Adding 'Has dependencies' so this doesn't get accidentally considered 'ready for merge' until we've decided on something.

pchote:
My second preference would be to remove demo trucks (which makes some sense, considering they have a nuke warhead) but keep the parabombs untouched.

cjshmyr:
I think removing both of these is presumptuous and I'd rather not introduce tech-level unit value tweaks unless absolutely having to.

But since Germany is currently neutered by this tech level, I won't object if we disabled the demo truck. Please keep the parabombs.

reaperrr:
Alternatively, we could keep demo trucks and introduce a "No Superweapons"-exclusive variant of the Chrono Tank with `pdox` stripped from prerequisites. If that makes it too powerful, we could increase price or decrease chrono range a little to counter the earlier availability.

That way we'd un-neuter Germany instead of neutering Ukraine as well.

pchote:
We should move CTNK to ATEK on all tech levels.  DTRK doesn't depend on MSLO, after all.

reaperrr:
Done.

cjshmyr:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/991bd25de1d966995016ce9b4da83684813119bf)

PR #10611, opened 2016-01-26
Make FlipX/FlipY flip respective embedded offsets as well

reaperrr:
When a sequence is flipped along an axis, flip embedded offsets for that axis as well.

### DIFF
--- a/OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
+++ b/OpenRA.Mods.Common/Graphics/DefaultSpriteSequence.cs
@@ -150,7 +150,7 @@ namespace OpenRA.Mods.Common.Graphics
 					{
 						var sd = sub.Value.ToDictionary();
 
-						// Allow per-sprite offset, start, and length
+						// Allow per-sprite offset, flipping, start, and length
 						var subStart = LoadField(sd, "Start", 0);
 						var subOffset = LoadField(sd, "Offset", float2.Zero);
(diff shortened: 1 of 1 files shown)


reaperrr:
This should have zero effect on our maintained mods. It is a requirement for my MechCommander mod. The only Westwood-related mod where this might play a role is Dune 2, pinging @evgeniysergeev whether this causes any regressions there.

Phrohdoh:
Do you happen to have a set of gifs comparing before and after this commit so we can see a difference?

pchote:
Does this really need to be optional?  Offsets are there to fudge the sprites, so I don't see a situation where you *wouldn't* want to flip the offsets.  Using offsets to fake in-world offsets is a hack.

reaperrr:
@Phrohdoh screenshots should be enough:
without flipping offsets:
no-offset-flip

with flipped offsets:
offset-flip

@pchote: I just wanted to play it safe, but personally I wouldn't mind at all if this was non-optional.

reaperrr:
Updated.

pchote:
Looks fine to me :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/991bd25de1d966995016ce9b4da83684813119bf...8c16969a90785a50b1d7e98423e5ec619dbdbbdb)

Issue #7913, opened 2015-04-10
Wiki does not auto populate non-trait [Desc] tags


UberWaffe:
After noticing #7659 and #7174 I realized that I had entirely neglected the documentation side of the A.I. support power decision addition.

I've noticed that the [Desc] tags in https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/AI/SupportPowerDecision.cs is not being automatically populated into the wiki, presumably because there is no category for non-trait type auto documentation.

I've updated https://github.com/OpenRA/OpenRA/wiki/Defining-custom-AI in order to explain how the AI decisions work, though it might be useful to make the auto documenting for the wiki also check the rest of the code (or at least sections of it) for [Desc] tags and adding them to some or the other wiki category.

PR #10613, opened 2016-01-26
Added Gates to the Red Alert mod

Mailaender:
Only took us 4 years. https://github.com/OpenRA/OpenRA/commit/11b253e4c3c71dbea2abcda4d899aea1b1dd196e In contrast to Tiberian Sun this won't suffer from overlaps. https://github.com/OpenRA/OpenRA/pull/9981

### DIFF
--- a/mods/ra/rules/defaults.yaml
+++ b/mods/ra/rules/defaults.yaml
@@ -511,6 +511,32 @@
 			TopLeft: -512, -512
 			BottomRight: 512, 512
 
+^Gate:
+	Inherits: ^BasicBuilding
+	Valued:
+		Cost: 250
+	Health:
+		HP: 350
+	Armor:
+		Type: Heavy
+	LineBuildNode:
+		Types: concrete, gate
+	-Building:
+	-WithSpriteBody:
+	WithGateSpriteBody:
+	Power:
+	CanPowerDown:
+	Tooltip:
+		Name: Gate
+		Description: Automated barrier that opens for allied units.
+	Gate:
+		Adjacent: 4
+		BlocksProjectilesHeight: 0
+		BuildSounds: place2.aud
+		OpeningSound: cashturn.aud
+		ClosingSound: cashturn.aud
+		TerrainTypes: Clear, Road
+
 ^TechBuilding:
 	Inherits: ^BasicBuilding
 	Health:
--- a/mods/ra/rules/structures.yaml
+++ b/mods/ra/rules/structures.yaml
@@ -1666,6 +1666,40 @@ BRIK:
 	WithWallSpriteBody:
 		Type: concrete
 
+VGATE:
+	Inherits: ^Gate
+	Gate:
+		Footprint: x x x
+		Dimensions: 1,3
+	WithGateSpriteBody:
+		WallConnections: 0,-1, 0,3
+	LineBuildNode:
+		Connections: 0,-1, 0,1
+	Tooltip:
+		Name: Vertical Gate
+
+HGATE:
+	Inherits: ^Gate
+	Gate:
+		Footprint: xxx
+		Dimensions: 3,1
+	WithGateSpriteBody:
+		WallConnections: -1,0, 3,0
+	LineBuildNode:
+		Connections: -1,0, 1,0
+	Tooltip:
+		Name: Horizontal Gate
+	RenderSprites:
+		Image: agate
+		FactionImages:
+			allies: agate
+			england: agate
+			france: agate
+			germany: agate
+			soviet: sgate
+			russia: sgate
+			ukraine: sgate
+
 CYCL:
 	Inherits: ^Wall
 	Tooltip:
--- a/mods/ra/sequences/structures.yaml
+++ b/mods/ra/sequences/structures.yaml
@@ -598,6 +598,33 @@ brik:
 		Length: 16
 	icon: brikicon
 
+vgate:
+	make:
+		Length: 7
+	idle:
+		Frames: 6, 5, 4, 3, 2, 1, 0
+		Length: 7
+		ZOffset: -1c511
+	damaged-idle:
+		Frames: 13, 12, 11, 10, 9, 8, 7
+		Length: 7
+		ZOffset: -1c511
+
+agate:
+	make:
+		Length: 7
+	idle:
+		Frames: 6, 5, 4, 3, 2, 1, 0
+		Length: 7
+		ZOffset: -1c511
+	damaged-idle:
+		Frames: 13, 12, 11, 10, 9, 8, 7
+		Length: 7
+		ZOffset: -1c511
+
+sgate:
+	Inherits: agate
+
 sbag:
 	idle:
 		Length: 16


pchote:
I'm going to play the voice of unpopular opinion: do we really need these in RA?  What gameplay gap in the RA mod do they fill?  It took quite some time to recover from the "add ALL the things (and deal with the balance implications later)" phase of the RA mod.

Mailaender:
They aren't buildable. Not just because they have no icon, but also because they don't really play a role in practice. I plan to use this in a Lua port of @ScottNZ's spy co-op mission where you have to get past a checkpoint. https://github.com/Mailaender/OpenRA/tree/infiltration-mission

pchote:
It may be better to include them as map specific rules.  The last time we added a new actor type for a planned mission was the zombie, and that instead sat around for years without strong justification for keeping it.

Mailaender:
Give it time. Last Sunday someone finally delivered: http://resource.openra.net/maps/8514/

pchote:
Two years and 6 months (plus a cameo in the shellmap appearance once a year that could be replaced by ants) to be enough time by more than an order of magnitude.  They could quite easily live as map-specific rules in that map.

Mailaender:
Why do you come up with removing these just when we finished our job and why the derailing of this pull request with zombie sprite comparisons. This isn't adding a new unit, but more like a decorational element. If we don't set up the rules in the base mod, it will stay invisible in the map editor with zero discoverability.

pchote:
I have a long history of objecting to adding new third-party content that doesn't have a clear use-case.  Setting these up as a hidden prop is IMO much less useful than integrating them properly with the game (after making sure that they have a place in the balance).

Mailaender:
Added the props to a map so we get a showcase.

GraionDilach:
The more map props, the merrier. Technically they aren't hidden this way anyway since the map editor shows them and not every mapper is a YAML modder as well to enable them themselves.

The zombie is clearly a mission unit, due to having an active role. Map props like, say Nooze's lighthouse or gates are a complete different thingy.

mods/ra/rules/structures.yaml
```diff
@@ -1666,6 +1666,42 @@ BRIK:
...
+	WithGateSpriteBody:
+		WallConnections: -1,0, 3,0
+	LineBuildNode:
+		Connections: -1,0, 1,0
+	Tooltip:
+		Name: Horizontal Gate
+	RenderSprites:
+		FactionImages:
+			Random: agate
+			RandomAllies: agate
```
pchote: Random factions aren't playable, so shouldn't be listed.

Mailaender: This is required for the map editor. Try it on Keep of the Grass 2.

pchote: Compare with d2k, which doesn't list the random faction.

Micr0Bit:
yeah , no ... we dont need this in RA1 ... the next release will feature proper wall-defending (as they are finally fixed) ... so this will change the gameplay allready (most likely)

i know its a good that we can finally make those gates ... but theres no reason to add them to RA ... (other then showing off) 

:-1:

pchote:
As a compromise, could we add these as map-specific assets/rules to Fort Lonestar?

Micr0Bit:
>As a compromise, could we add these as map-specific assets/rules to Fort Lonestar?

thats okay for me

pchote:
Note that ^ is assuming that it wouldn't hurt the gameplay of Lonestar.  Ping @abcdefg30.

Micr0Bit:
yeah , i though about this for a minute ... but it probably "help" the gameplay more ... as you can block the enemys better ... prevent them from quickly destroying the important buildings (barracks , derrick)

either way , its gonna be playtested anyway ... before we would fully shipit

abcdefg30:
Sorry, but I don't think that this would be a good idea, as gates would delay your units from moving out of the base and instead of adding further protection, make you more vulnerable to flame artillery.

We can of course try it though, just ping me on IRC if you want a testgame.

Anyway, I think the original idea was to use this for an interior mission?

Mailaender:
The original idea is to keep this in the map editor available for everyone. It will probably be used in single-player missions as a decorational gimmick. This won't break multiplayer games as it isn't available there and this doesn't need to be added to each map individually. The whole discussion sparked here is mostly uninformed nonsense.

Phrohdoh:
I see no reason to blanket-add this feature to RA.
Mappers are capable of implementing it themselves if they really want it, and say after some months we notice most maps do have them perhaps we could revisit the discussion with actual figures rather than "probably/maybe".

@Micr0Bit made an excellent point about gates being a special TS-only feature.

Just because you don't agree with the discussion doesn't mean it is uninformed nonsense.

GraionDilach:
One should not expect mappers to understand the underlying game for additional assets to be utilized. That's a modder's work.

A mapper won't add something to his/her map without the knowledge of said thing exists which is why it should be added to the map editor and thereby the game first. Your counterargument is stupid and relies on the expectation that all mappers are modders.

Micr0Bit:
we could still add this to a different (new?) custom map , doesnt have to be lonestar

Micr0Bit:
@GraionDilach : the thing though is ... stuff that is easy available for "mappers" also gets heavily used in maps ... zombies and ants as best example (biolab , right?)

imagine all the mappers combined together as a big child ... the devs give them a couple of red and blue bricks ... now the child builds various structures with those bricks ... 

what you can do now is ... give the child a special yellow brick , which it can use after a bit of effort to set "accents" on his buildings ...

or 

just give him the yellow brick without any effort , and then have hundreds of stuctures , everything contains a mix of red-blue-yellow bricks

GraionDilach:
What is the difference between gates and say, lighthouse or windmill? Ah yeah, nothing.

Both are just map props, yet one gets added without any argument/usecase and one is booed to hell. That's everything I think about this.

Micr0Bit:
isnt't the gate a "state-changing" feature which could potentially effect the game-outcome ?

pchote:
The lighthouse and windmill are generic map props that could be reasonably expected to be used across many maps.  Gates are very specific, and nobody has yet presented even one specific usecase for them in RA maps (except for Lonestar, which was shot down for a valid reason).

MustaphaTR:
> The lighthouse and windmill are generic map props that could be reasonably expected to be used across many maps. Gates are very specific, and nobody has yet presented even one specific usecase for them in RA maps (except for Lonestar, which was shot down for a valid reason).

I should say i disagree to you. Gates are also usable in most maps, in my case they are more usable bacause almost all of my maps includes an army base. Props are props, using them is left to mapper if we add them mappers can find a usage don't worry about it.

Mailaender:
> nobody has yet presented even one specific usecase for them in RA maps

Like I wrote last month https://github.com/OpenRA/OpenRA/pull/10613#issuecomment-175245358 I actually had a specific use case in mind. As a workaround @ScottNZ built a barrier with sand bags and infantry moving aside. But if getting even such simple dependencies merged, I doubt I will want to go through the pain of contributing a complex Lua scripted remake here. The `CHANGES REQUESTED` button has also become an instrument to enforce a very strict way how things have to be done, which limits the ideas and creative freedom of individual contributors.

Everybody seems to have an opinion on this, but I expect that very few people have actually downloaded this branch, tried it out and gave some hints on how to improve it, which is the only thing I ask for when submitting something for review. I don't care for meta discussions about how functional a prop can be and how possibly game-breaking or irrelevant gates were in the TS. Please leave that to the chat room or the discussion forums.

pchote:
> Everybody seems to have an opinion on this, but I expect that very few people have actually downloaded this branch, tried it out and gave some hints on how to improve it, which is the only thing I ask for when submitting something for review. I don't care for meta discussions about how functional a prop can be and how possibly game-breaking or irrelevant gates were in the TS. Please leave that to the chat room or the discussion forums.

The primary purpose of a PR is to evaluate a set of changes for inclusion.  The very first point that needs to be established is "do we want this": discussion about how the feature integrates with the rest of the project is *essential* for this.  Most technical issues with a feature can be fixed if the feature is wanted, but I don't see much point spending time commenting on technical issues if it doesn't pass the initial smell test.

I find it difficult to read the dismissal of the integration discussion as anything other than "if it works then you should merge it – I don't care how this affects the game, merge it now and *you* can sort out the details later".

Mailaender:
My point is: the sprites are already included https://github.com/OpenRA/OpenRA/tree/bleed/mods/ra/bits and shipped so that decision was made 4 years ago by @chrisforbes et al.

Mailaender:
The point is: enough talking. You need to come to a conclusion now. `rm -r mods/ra/gate*.shp` and upload a prefab map to http://resource.openra.net or get this merged.

pchote:
I've made my opinion clear on why I think this is a bad idea, but I also don't think gates are a big enough deal to override the review process and veto the pr.  My biggest issue here has been with attitude.  If the other reviewers want gates then you'll still need to fix the FactionImages issue (I even gave some hints on how to improve it!) before merging.

GraionDilach:
@Micr0Bit: if they would be buildable by the player, then yes. However, this PR ain't make them as such, so yeah, they are still merely map props.

penev92:
My two cents, since I see I haven't yet commented on this:
 - I don't see why RA needs gates.
 - Since we're talking about adding a map prop only (like any tree or rock), the drama over this is unnecessary. If you wanted to add them as buildable structures, then there would need to be a serious gameplay discussion.

I usually prefer to stick to the "authors need to justify why the PR needs to be merged" viewpoint, but with all the arguing this is caused (for a map prop!) I'd be OK if we compromised on taking them as nothing more than unbuildable, uncontrollable map props.

mods/ra/rules/structures.yaml
Mailaender: I checked d2k which defaults everything to Harkonnen imagery. That also has it's downsides, but I changed it nonetheless.

Mailaender:
> you'll still need to fix the FactionImages issue (I even gave some hints on how to improve it!) before merging.

I used the same hack as you did for Dune 2000 now by defaulting to a hard-coded image https://github.com/OpenRA/OpenRA/blob/bleed/mods/d2k/rules/structures.yaml#L79

mods/ra/sequences/structures.yaml
```diff
@@ -598,6 +598,40 @@ brik:
 		Length: 16
 	icon: brikicon
 
+vgate:
+	make:
+		Start: 0
```
reaperrr: This line is redundant (0 is the default).

mods/ra/sequences/structures.yaml
```diff
@@ -598,6 +598,40 @@ brik:
...
+		Length: 7
+		ZOffset: -1c511
+	damaged-idle:
+		Frames: 13, 12, 11, 10, 9, 8, 7
+		Length: 7
+		ZOffset: -1c511
+
+agate:
+	make:
+		Start: 0
```
reaperrr: This line is redundant (0 is the default).

mods/ra/sequences/structures.yaml
```diff
@@ -598,6 +598,40 @@ brik:
...
+		Length: 7
+		ZOffset: -1c511
+	damaged-idle:
+		Frames: 13, 12, 11, 10, 9, 8, 7
+		Length: 7
+		ZOffset: -1c511
+
+sgate:
+	make:
+		Start: 0
```
reaperrr: This line is redundant (0 is the default).

mods/ra/sequences/structures.yaml
```diff
@@ -598,6 +598,40 @@ brik:
...
+		Length: 6
+	idle:
+		Start: 6
+		Length: -6
+
```
reaperrr: The whole sgate sequence definition looks quite unfinished (Length is wrong, no damage stage, no lower ZOffset), I recommend simply using `Inherits: agate`.

reaperrr:
As map prop this is ok, once the sequences have been cleaned/fixed.

reaperrr:
Forgot something: Please change the hitshape to a rectangle covering all occupied cells for both vgate and hgate, otherwise only the middle blocks projectiles.

Mailaender:
Rebased. Fixed the sequences and added a rectangular hit shape. Having to look that up myself, I found our https://github.com/OpenRA/OpenRA/wiki/Traits#health documentation lacking and added at least a small stub.

OpenRA.Mods.Common/Traits/Health.cs
```diff
@@ -23,6 +23,7 @@ public class HealthInfo : ITraitInfo, UsesInit<HealthInit>
 		public readonly bool NotifyAppliedDamage = true;
 
 		[FieldLoader.LoadUsing("LoadShape")]
+		[Desc("Capsule, Circle and Rectangle are supported.")]
```
pchote: The properties of the three shapes aren't documented (#7913), so I don't see any benefit from this.

pchote: Modders can also add their own implementations, so hardcoding only these three is a bad idea.

GraionDilach: `By default, Capsule...` then imo. This is still useful for modders to know what to look up in the yamls.

pchote: Modders can already look up `Shape` and find exactly the same information.

reaperrr:
The easiest fix for the Travis/Lint errors is to increase the `TargetExtraSearchRadius` default values of `Bullet`, `Missile` and `SpreadDamageWarhead` to `2048`.

pchote:
Changing the default settings across all mods definitely takes this outside of the "meh, why not?" category.

reaperrr:
> Changing the default settings across all mods (or even just RA) definitely takes this outside of the "meh, why not?" category.

Hm yeah, and setting custom, larger radi on all RA bullet/missile projectiles and spread dmg warheads just for this would be absolute overkill (and quite some work, even with an upgrade rule).

Mailaender:
Removed 33b2a9e again and made gates non-targetable by default.

reaperrr:
> Removed 33b2a9e again and made gates non-targetable by default.

That makes them invulnerable, I don't think that's a good idea.

Since this is just a map prop, I'd rather set `BlocksProjectilesHeight` to `0`, so it only blocks enemy movement but not projectiles. Then the default circle hit shape isn't a problem either, and we can leave it `Targetable` and destructible.

Mailaender:
Hmm, okay that is indeed a less visible limitation. It doesn't really matter to me, because you can tweak the rules accordingly when using it in a custom map.

reaperrr:
:+1:

mods/ra/rules/defaults.yaml
```diff
@@ -511,6 +511,32 @@
...
+	Health:
+		HP: 350
+	Armor:
+		Type: Heavy
+	LineBuildNode:
+		Types: concrete, gate
+	-Building:
+	-WithSpriteBody:
+	WithGateSpriteBody:
+	Power:
```
abcdefg30: Does this even make sense without an amount specified?

Mailaender: https://github.com/OpenRA/OpenRA/wiki/Traits#canpowerdown requires it. This is needed to be able to manually turn off gates.

abcdefg30: Ah well, I guessed so, but couldn't really imagine why someone would want to turn gates off by powering them down...

abcdefg30:

mods/ra/rules/defaults.yaml
pchote: Why are you allowing people to power down gates, especially if they don't draw power?

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a329dbc9d8da7d5b1a4556c5a1cef0a8ecba2938...f7d4b9d46201377fedba38dff9d62cb3096e09fb)

Issue #5030, opened 2014-04-05
Tech building ideas


dan9550:
Some ideas:

* When you have captured a tech hospital your inf should have a little icon in the selection box to indicate they are being healed by the hospital (i swear this was in ra2 couldn't find a picture)

* Capturable Biolab makes inf immune to tiberium perhaps? not even sure what the biolab does at the moment couldn't find it in the YAML either.

Mailaender:
> Capturable Biolab makes inf immune to tiberium perhaps? not even sure what the biolab does at the moment couldn't find it in the YAML either.

Draw a hazmat suit icon and create a new upgrade that can be purchased from there. Code is in place since we merged https://github.com/OpenRA/OpenRA/pull/8114.

GraionDilach:
How would it work? Neither Armor, neither TargetTypes are upgradable (this is a good example why I want TargetType upgrading would look like most others, not via #8339's UpgradedTargetTypes).

I do agree something like
```
^Infantry
TargetableUnit:
		TargetTypes: TiberiumDamage
		UpgradeTypes: biolab
		UpgradeMaxEnabledLevel: 0
```

would be nice but we cannot have such.

Mailaender:
https://github.com/OpenRA/OpenRA/wiki/Traits-(playtest)#poisonedbytiberium is upgradeable.

GraionDilach:
Oh, I overlooked that it has a separate trait. -_-

pchote:
:+1: to removing the buildable dinos and adding a tib-heal upgrade.

Issue #3788, opened 2013-09-02
Encyclopedia


HSFOutcast:
Something that might take a while to add but would be totaly worth it imo.

In Dune 2 you could check an encyclopedia with all units in that game.

We could use something like that in openRA (it wont be much use when you play multiplayer, but it will be a nice touch imo)

A good example here is the game "Battle for Wesnoth".

x-a-n-a-x:
@AlexaDeWit : any news about this issue and your branch ?

AlexaDeWit:
@x-a-n-a-x Sorry man, life has taken and adventurous series of events and my time has been very limited. Work has essentially been abandoned for the foreseeable future. Hopefully someone with more time and ambition will take up the cause :3

PR #10616, opened 2016-01-26
Added HazMat suites to Tiberian Dawn

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/5030.


pchote:
Great!  This is something that has bugged me for ages.

mods/cnc/rules/defaults.yaml
```diff
@@ -204,6 +204,11 @@
 		CargoType: Infantry
 	HiddenUnderFog:
 	PoisonedByTiberium:
+		UpgradeTypes: hazmatsuites
```
cjshmyr: typo, `suits` or `suit`

Mailaender: Hehe, fixed.

pchote:
Works as advertised, so :+1:

A followup pr should add some kind of selection decoration to infantry to indicate that the upgrade is active.

GraionDilach:
I made a pip.

https://copy.com/Ty6WXfxYzxT21nSS

obrakmann:
Thanks @GraionDilach, I added them in https://github.com/Mailaender/OpenRA/pull/56

Mailaender:
Awesome, thanks! Can you also make those for hospital healing?

pchote:
I believe that RA has a medic icon we can use for that.

Phrohdoh:
This is looking pretty sharp!

GraionDilach:
>I believe that RA has a medic icon we can use for that.

Indeed it had one. I took it out, converted to TD palette and cropped the canvas.
https://copy.com/MDuTLsTOpJoRuTxo

Mailaender:
Thanks. Submitted it in https://github.com/OpenRA/OpenRA/pull/10680.

GraionDilach:
Come to think of it, this will be wrong, they will overlap with each other/ranks.

Mailaender:
They work together with ranks although with those tiny infantry it is already quite edgy. I added a `Description` which can be used in future tooltips or #3788.

pchote:
You could set it up with upgrades to render the bio glyph in a different position if the hospital upgrade is already applied.

GraionDilach:
My suggestion would be to shift hazmat and hospital into topright corner - that worked well in YR (for hosp and machine shop) and with the disguise pip in our RA/TS mod.

obrakmann:
Let's do the fine-tuning in the spin-off PR. For now, this works as advertised. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/63626aadca490f8425cfaf74be010f391239265f)

PR #10618, opened 2016-01-27
Set the correct PlatformTarget for OpenRA.Test

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/9114 or at least makes it consistent with other libraries.

### DIFF
--- a/OpenRA.Test/OpenRA.Test.csproj
+++ b/OpenRA.Test/OpenRA.Test.csproj
@@ -3,7 +3,7 @@
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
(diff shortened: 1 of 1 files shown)


RoosterDragon:
I don't think this would fix it (the other assemblies that are still x86 presumably would suffer the same problem). But maybe somebody suffering from this issue can try it and see if I'm wrong.

Mailaender:
For some reason it seems to fix the annoying warnings I get all the time:
image

RoosterDragon:
Those are totally different warnings from the linked issue and should be suppressed by #10345, besides.

Mailaender:
For some reason they aren't on MonoDevelop.

RoosterDragon:
Well you should file a compiler bug/feature request so it respects that pragma then.

We can't just change the Test dll - it has 32-bit dependencies for one thing:

> There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "C:\...\OpenRA\OpenRA.Game.exe", "x86". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.

(Same again for the Mods.Common dll)

PR #10620, opened 2016-01-27
Fix units warping from an exit subcell to cellcenter when produced

### LINKED ISSUE #9678: Weird visual 'drop' when vehicles exit factory in TS

teees:
Fixes #9678

Before:
https://streamable.com/jhed
After:
https://streamable.com/p5wg

### DIFF
--- a/OpenRA.Mods.Common/Traits/Mobile.cs
+++ b/OpenRA.Mods.Common/Traits/Mobile.cs
@@ -756,7 +756,7 @@ namespace OpenRA.Mods.Common.Traits
 			var pos = self.CenterPosition;
 
 			if (subCell == SubCell.Any)
-				subCell = self.World.ActorMap.FreeSubCell(cell, subCell);
+				subCell = Info.SharesCell ? self.World.ActorMap.FreeSubCell(cell, subCell) : SubCell.FullCell;
 
 			// TODO: solve/reduce cell is full problem
 			if (subCell == SubCell.Invalid)


pchote:
Oops, well that was silly!  How does this behave for infantry, which should ideally be making use of subcells?

teees:
Hmm I checked that shortly and thought it was alright, turns out it isn't. Infantry clips through each other. Probably better to fix it in Mobile.MoveIntoWorld and check if the unit can actually use subcells?

teees:
This should now be the right behavior for `MoveIntoWorld`, which also fixes the production bug

Mailaender:
:+1:

abcdefg30:
:+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3eb9b4a2a04f54c11b7f66412b982e6912702b53...c4ef2f40894524c00434ed2c7f65768747699a31)

PR #10624, opened 2016-01-27
Fix color validator issues and increase color picker palette size.

### LINKED ISSUE #10217: Color validator is rejecting most colors on the ~left half of the color chooser. [$15]
This is silly.  If we really do mean to be rejecting this much of the color space, then the picker should be adjusted to not show this area of rejected colors.

<bountysource-plugin>
---
There is a **[$15 open bounty](https://www.bountysource.com/issues/29090014-color-validator-is-rejecting-most-colors-on-the-left-half-of-the-color-chooser?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #10217 by changing the following:
- Default color match threshold decreased from 0x70 to 0x50 (and can now be overridden in mod.yaml).
- Removed matching against all terrain types except clear/water/resources.
- Added additional matching in HSV space so that colors outside the picker range are disallowed.

Also does some code cleanup and increases the size of the color picker to make it more useable.

You can test using 87979d92385d07936d4a1929ff35748d583f33a9 to visualize the blocked colors.


Mailaender:
:+1: A lot less annoying to use. I assume we don't need the workaround for glitchy overbright colors anymore since we switched to the HTML color notation format.

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/6b6c9bf0a3e8a4341bf33a393da02359cead10dd...e38d5a7c217467857631dd3bd00833b85bf1b461)

PR #10627, opened 2016-01-27
Added Attack*.StopToShoot

Mailaender:
~~Closes https://github.com/OpenRA/OpenRA/issues/2327.~~ See http://modenc.renegadeprojects.com/NoMovingFire for reference.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
@@ -35,10 +35,13 @@ namespace OpenRA.Mods.Common.Traits
 
 		[VoiceReference] public readonly string Voice = "Action";
 
+		[Desc("Unable to fire while in motion.")]
+		public readonly bool NoMovingFire = false;
+
 		public override abstract object Create(ActorInitializer init);
 	}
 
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
```diff
@@ -35,10 +35,13 @@ public abstract class AttackBaseInfo : UpgradableTraitInfo
 
 		[VoiceReference] public readonly string Voice = "Action";
 
+		[Desc("Unable to fire while in motion.")]
+		public readonly bool StopToShoot = false;
```
pchote: The name implies that it should actively stop before shooting, but this just prevents it from firing while moving.  This introduces a situation where the unit may not fire at all, which isn't really what #2327 wanted.

Mailaender: Renamed it.

pchote:
Note that even with the name change that this doesn't close #2327, which is asking for the targeting/turret behavior of the original units.

pchote:
To be able to solve #2327 this should stay named `StopToShoot`, and also have a check in the turreted attack code to stop moving if the targeted unit is in range.  I don't see a usecase for the feature in its current state, so i'd be hesitant to take it as-is.

PR #10628, opened 2016-01-28
Don't grant upgrades to newly built units while disabled

teees:
before:
https://streamable.com/2rl4
after:
https://streamable.com/ygw5

(you need sound to hear it)

What happened is that UpgradeActorsNear would grant upgrades on newly built actors and revoke it in the next tick when disabled.


abcdefg30:
:+1: / :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c4ef2f40894524c00434ed2c7f65768747699a31...152c367eee3255f695d747bd408966dcac8a0726)

PR #10630, opened 2016-01-28
Make UpgradeActorsNear vertical range customizable

reaperrr:
Includes plumbing to do the same for other ProximityTrigger-based features, but the focus of this PR is UpgradeActorsNear.

Makes it more flexible and a requirement to implement #10590 properly (otherwise aircraft would start reloading while still in the air).

### DIFF
--- a/OpenRA.Game/Traits/World/ActorMap.cs
+++ b/OpenRA.Game/Traits/World/ActorMap.cs
@@ -98,22 +98,26 @@ namespace OpenRA.Traits
 
 			WPos position;
 			WDist range;
+			WDist vRange;
+
 			IEnumerable<Actor> currentActors = Enumerable.Empty<Actor>();
 
-			public ProximityTrigger(WPos pos, WDist range, Action<Actor> onActorEntered, Action<Actor> onActorExited)
+			public ProximityTrigger(WPos pos, WDist range, WDist vRange, Action<Actor> onActorEntered, Action<Actor> onActorExited)
 			{
 				this.onActorEntered = onActorEntered;
 				this.onActorExited = onActorExited;
 
-				Update(pos, range);
+				Update(pos, range, vRange);
 			}
 
-			public void Update(WPos newPos, WDist newRange)
+			public void Update(WPos newPos, WDist newRange, WDist newVRange)
 			{
 				position = newPos;
 				range = newRange;
+				vRange = newVRange;
 
-				var offset = new WVec(newRange, newRange, WDist.Zero);
+				var offset = new WVec(newRange, newRange, newVRange);
+
 				TopLeft = newPos - offset;
 				BottomRight = newPos + offset;
 
@@ -128,7 +132,8 @@ namespace OpenRA.Traits
 				var oldActors = currentActors;
 				var delta = new WVec(range, range, WDist.Zero);
 				currentActors = am.ActorsInBox(position - delta, position + delta)
-					.Where(a => (a.CenterPosition - position).HorizontalLengthSquared < range.LengthSquared)
+					.Where(a => (a.CenterPosition - position).HorizontalLengthSquared < range.LengthSquared
+						&& (vRange.Length == 0 || (a.World.Map.DistanceAboveTerrain(a.CenterPosition).LengthSquared <= vRange.LengthSquared)))
 					.ToList();
 
 				var entered = currentActors.Except(oldActors);
@@ -439,10 +444,10 @@ namespace OpenRA.Traits
 			}
 		}
 
-		public int AddProximityTrigger(WPos pos, WDist range, Action<Actor> onEntry, Action<Actor> onExit)
+		public int AddProximityTrigger(WPos pos, WDist range, WDist vRange, Action<Actor> onEntry, Action<Actor> onExit)
 		{
 			var id = nextTriggerId++;
-			var t = new ProximityTrigger(pos, range, onEntry, onExit);
+			var t = new ProximityTrigger(pos, range, vRange, onEntry, onExit);
 			proximityTriggers.Add(id, t);
 
 			foreach (var bin in BinsInBox(t.TopLeft, t.BottomRight))
@@ -463,7 +468,7 @@ namespace OpenRA.Traits
 			t.Dispose();
 		}
 
-		public void UpdateProximityTrigger(int id, WPos newPos, WDist newRange)
+		public void UpdateProximityTrigger(int id, WPos newPos, WDist newRange, WDist newVRange)
 		{
 			ProximityTrigger t;
 			if (!proximityTriggers.TryGetValue(id, out t))
@@ -472,7 +477,7 @@ namespace OpenRA.Traits
 			foreach (var bin in BinsInBox(t.TopLeft, t.BottomRight))
 				bin.ProximityTriggers.Remove(t);
 
-			t.Update(newPos, newRange);
+			t.Update(newPos, newRange, newVRange);
 
 			foreach (var bin in BinsInBox(t.TopLeft, t.BottomRight))
 				bin.ProximityTriggers.Add(t);
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeActorsNear.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeActorsNear.cs
@@ -57,7 +57,7 @@ namespace OpenRA.Mods.Common.Traits
 		public void AddedToWorld(Actor self)
 		{
 			cachedPosition = self.CenterPosition;
-			proximityTrigger = self.World.ActorMap.AddProximityTrigger(cachedPosition, cachedRange, ActorEntered, ActorExited);
+			proximityTrigger = self.World.ActorMap.AddProximityTrigger(cachedPosition, cachedRange, WDist.Zero, ActorEntered, ActorExited);
 		}
 
 		public void RemovedFromWorld(Actor self)
@@ -80,7 +80,7 @@ namespace OpenRA.Mods.Common.Traits
 			{
 				cachedPosition = self.CenterPosition;
 				cachedRange = desiredRange;
-				self.World.ActorMap.UpdateProximityTrigger(proximityTrigger, cachedPosition, cachedRange);
+				self.World.ActorMap.UpdateProximityTrigger(proximityTrigger, cachedPosition, cachedRange, WDist.Zero);
 			}
 		}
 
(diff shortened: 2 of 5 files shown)


reaperrr:
#10590 serves as testcase.

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -94,25 +94,34 @@ class ProximityTrigger : IDisposable
...
+			public void Update(WPos newPos, WDist newRange, WDist newVRange)
 			{
 				position = newPos;
 				range = newRange;
+				vRange = newVRange;
 
 				var offset = new WVec(newRange, newRange, WDist.Zero);
+
+				if (considerVRange)
+					offset = new WVec(newRange, newRange, newVRange);
```
pchote: You can combine these into a single `var offset = new WVec(newRange, newRange, considerVRange ? newVRange : WDist.Zero);`

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -126,9 +135,16 @@ public void Tick(ActorMap am)
 
 				var oldActors = currentActors;
 				var delta = new WVec(range, range, WDist.Zero);
-				currentActors = am.ActorsInBox(position - delta, position + delta)
-					.Where(a => (a.CenterPosition - position).HorizontalLengthSquared < range.LengthSquared)
-					.ToList();
+
+				if (considerVRange)
+					currentActors = am.ActorsInBox(position - delta, position + delta)
```
pchote: You can move these shared lines above the if to avoid duplication.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeActorsNear.cs
```diff
@@ -22,6 +22,12 @@ public class UpgradeActorsNearInfo : ITraitInfo
 		[Desc("The range to search for actors to upgrade.")]
 		public readonly WDist Range = WDist.FromCells(3);
 
+		[Desc("Should the vertical range be considered.")]
+		public readonly bool ConsiderVerticalRange = false;
+
+		[Desc("The vertical range to search for actors to upgrade.")]
+		public readonly WDist VerticalRange = WDist.Zero;
```
pchote: Maybe use a range of `-1` instead of `ConsiderVerticalRange == false`?

reaperrr:
Updated.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeActorsNear.cs
```diff
@@ -22,6 +22,9 @@ public class UpgradeActorsNearInfo : ITraitInfo
 		[Desc("The range to search for actors to upgrade.")]
 		public readonly WDist Range = WDist.FromCells(3);
 
+		[Desc("The vertical range to search for actors to upgrade. -1 means this is ignored (actors are upgraded regarless of altitude).")]
```
Phrohdoh: `regardless`

pchote:
This feels like a bit of a hack IMO, what you really want for #10590 is a docking trait that applies an upgrade instead of a specific area.

GraionDilach:
This is hacky for that usecase I agree but IMO very well required longterm for stuff like damaging auras emitted from an infantry not affecting air etc.

reaperrr:
> This feels like a bit of a hack IMO, what you really want for #10590 is a docking trait that applies an upgrade to a specific actor instead of a specific area.

Generally speaking I'm aware of that, but I felt making `UpgradeActorsNear` (and `ProximityTrigger`s in general) more flexible would be a solid interim solution which could be beneficial for other features as well in the long run.
For reloading/repairing specifically, it's not meant to be the final, perfect solution. I'm aware that the docking approach would be better for static actors like buildings and should ultimately be used there, but 3rd-party mods might want auras (whether healing, damaging or boosting doesn't matter) or map triggers with a finite vertical range, like @GraionDilach mentioned.

As for implementing docking, to my knowledge @penev92 was working on that some time ago. I haven't even begun looking into that, and I don't feel like waiting for that to be finished nor taking it over right now just for this, when `UpgradeActorsNear` can be used as well.

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -94,25 +94,31 @@ class ProximityTrigger : IDisposable
 
 			readonly Action<Actor> onActorEntered;
 			readonly Action<Actor> onActorExited;
+			readonly bool considerVRange;
 
 			WPos position;
 			WDist range;
+			WDist vRange;
```
pchote: Can we merge `considerVRange` into this (special case `vRange == 0` to mean any height, since a range of 0 is not meaningful on its own)?

pchote: Note: this supersedes my previous suggestion of `-1` (although that would also be fine)

reaperrr:
Updated (#10590 as well, for testing).

abcdefg30:
Looks good to me (this is easily testable with the TS stealth generator and aircraft). :+1: / :white_check_mark: 

I wonder though if we can't somehow also get this for a negative WDist (like for subterrainian stuff).

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -128,6 +132,7 @@ public void Tick(ActorMap am)
 				var delta = new WVec(range, range, WDist.Zero);
 				currentActors = am.ActorsInBox(position - delta, position + delta)
 					.Where(a => (a.CenterPosition - position).HorizontalLengthSquared < range.LengthSquared)
+					.Where(a => (vRange.Length < 1) || Math.Abs(a.CenterPosition.Z - position.Z) <= vRange.Length)
```
RoosterDragon: Minor quibbling follows:

Brackets around `vRange.Length < 1` are not needed.

Also,  you could instead do (but you don't have to):

```c#
var newCurrentActors = am.ActorsInBox...
if (vRange.Length >= 1)
    newCurrentActors = newCurrentActors.Where(Math.Abs(a.CenterPosition.Z - position.Z) <= vRange.Length);
currentActors  = newCurrentActors.ToList();
```
Also maybe a `VerticalLength` and `VerticalLengthSquared` would be good to introduce? (If you do, check against the squared distance.)

reaperrr:
Updated.

obrakmann:
Wouldn't a strategically placed `actor.IsAtGroundLevel` check be enough for your use case?

As it stands, this will break on heightmapped terrain.

reaperrr:
> Wouldn't a strategically placed `actor.IsAtGroundLevel` check be enough for your use case?

For the shipping mods that might be enough, but as soon as a 3rd-party mod has a landing position above ground it fails (and I don't think that's all that unlikely to happen).

> As it stands, this will break on heightmapped terrain.

I assume you mean when an actor is at lower height level than the upgrading actor. I'd rather adress that instead of the ground level check.

reaperrr:
@obrakmann 
> As it stands, this will break on heightmapped terrain.

Actually, doesn't the 4th commit cover that?

obrakmann:
No.  The point is that your logic seems wrong to me. You want to avoid that aircraft get reloaded in flight. Comparing their altitude in relation to the upgrading actor doesn't work then, you need to compare the upgraded actor's altitude against the ground level of its location. Imagine the refineries in the screenshot below were aircraft. They're both clearly at ground level, but wouldn't get reloaded. To do it right, you'd need to compare `map.DistanceAboveTerrain(actor.CenterPosition)` against the `VerticalRange`, IMO. That would solve the problem with non-zero landing positions as well. (I'd call it an offset instead of range, though.)

```diff
diff --git a/mods/ts/rules/nod-structures.yaml b/mods/ts/rules/nod-structures.ya
index 262357a..02f1fb8 100644
--- a/mods/ts/rules/nod-structures.yaml
+++ b/mods/ts/rules/nod-structures.yaml
@@ -389,6 +389,7 @@ NASTLH:
        UpgradeActorsNear:
                Upgrades: cloakgenerator
                Range: 12c0
+               VerticalRange: 0c256
                EnableSound: cloak5.aud
                DisableSound: cloak5.aud
                AffectsParent: true
```

screenshot from 2016-03-25 13 25 17

reaperrr:
Updated.

reaperrr:
Since we're extremely low on review bandwidth, nobody really seems to care and other PRs are more important to me, I'm closing this for now and will maybe bring it back when I feel the time is right.

reaperrr:
Reopened.

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -98,22 +98,26 @@ class ProximityTrigger : IDisposable
...
 			}
 
-			public void Update(WPos newPos, WDist newRange)
+			public void Update(WPos newPos, WDist newRange, WDist newVRange)
 			{
 				position = newPos;
 				range = newRange;
+				vRange = newVRange;
+
+				var offset = new WVec(newRange, newRange, newVRange.Length != 0 ? newVRange : WDist.Zero);
```
abcdefg30: If the length is 0, isn't `newVRange` not already the same as `WDist.Zero`?
So this can be simplified to `var offset = new WVec(newRange, newRange, newVRange);`

reaperrr:
Updated. Used the opportunity to rebase onto latest bleed and squash the `ActorMap` fixups into the first commit.

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -129,6 +133,7 @@ public void Tick(ActorMap am)
 				var delta = new WVec(range, range, WDist.Zero);
 				currentActors = am.ActorsInBox(position - delta, position + delta)
 					.Where(a => (a.CenterPosition - position).HorizontalLengthSquared < range.LengthSquared)
+					.Where(a => vRange.Length == 0 || (a.World.Map.DistanceAboveTerrain(a.CenterPosition).LengthSquared <= vRange.LengthSquared))
```
obrakmann: I'm not sure how smart Linq is about these expressions, but just to be safe I think it would be better to use only a single `.Where` and combine both conditions, lest the list is walked twice.

obrakmann:
Aside from that small nit it looks good to me now, :+1:

reaperrr:
Adressed nit.

OpenRA.Game/Traits/World/ActorMap.cs
RoosterDragon: For future reference, it would only walk the list once either way. It is marginally more efficient to combine the conditions though.

RoosterDragon:
:+1:

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/249caa5a4015baae954a5975699a6a5082f13090)

PR #10632, opened 2016-01-29
Improve Harvester-related code

reaperrr:
This removes some duplication and should improve AI performance by reducing frequency of the more expensive AI resource search.

Issues fixed by this PR:
- `Harvester.UnblockRefinery` was not just unblocking, but also performing a resource claim and queueing a `FindResources` activity (which would then perform another resource claim). 
- The `ContinueHarvesting` method would queue another `FindResources` after `UnblockRefinery`, resulting in potentially queueing `FindResources` twice.
- In `FindResources`, if no harvestable position was found, it would perform `harv.UnblockRefinery` and its own unblock at the same tick due to a bogus if check.
- the AI was making the assumption that a `Wait`ing harvester is idle. That is plain wrong, as `Wait` activities are intentionally queued by all harvester-related activities. This resulted in far too many expensive AI resource searches.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Harvester.cs
+++ b/OpenRA.Mods.Common/Traits/Harvester.cs
@@ -57,6 +57,12 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Search radius (in cells) from the last harvest order location to find more resources.")]
 		public readonly int SearchFromOrderRadius = 12;
 
+		[Desc("Maximum duration of being idle before queueing a Wait activity.")]
+		public readonly int MaxIdleDuration = 25;
+
+		[Desc("Duration to wait before becoming idle again.")]
+		public readonly int WaitDuration = 25;
+
 		[VoiceReference] public readonly string HarvestVoice = "Action";
 		[VoiceReference] public readonly string DeliverVoice = "Action";
 
@@ -239,6 +245,7 @@ namespace OpenRA.Mods.Common.Traits
 			}
 		}
 
+		int idleDuration;
 		public void TickIdle(Actor self)
 		{
 			// Should we be intelligent while idle?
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -263,9 +253,16 @@ public void TickIdle(Actor self)
 			}
 
 			UnblockRefinery(self);
+			idleDuration = idleDuration + 1;
```
Phrohdoh: `+= 1`

OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -263,9 +253,16 @@ public void TickIdle(Actor self)
...
+
+			// Wait a bit before queueing Wait activity
+			if (idleDuration > 25)
+			{
+				idleDuration = 0;
 
-			// Wait for a bit before becoming idle again:
-			self.QueueActivity(new Wait(10));
+				// Wait for a bit before becoming idle again:
+				self.QueueActivity(new Wait(25));
```
Phrohdoh: If you're going to change this number you may as well expose it to yaml.

reaperrr:
Updated.

Mailaender:
:+1: Can't spot any negative side effects in a d2k test game as well. :white_check_mark:

RoosterDragon:
I noticed a pair of harvesters become stuck during an all AI test game. They were both partially full and harvesting the same ore patch. They stopping harvesting/moving and just sat there. Hard to know if it's related to these changes or an existing issue though.

RoosterDragon:
I saw a similar harvester freeze whilst testing - but not using this PR. Therefore it must be a pre-existing problem.

reaperrr:
@RoosterDragon That's part of the reason for this PR, to make the code at least a little cleaner, easier to read and easier to debug.

obrakmann:
Looks good to me, didn't notice any misbehaviour, either. :+1:

PR #10552, opened 2016-01-19
TD Balance Changes v1192016

AoAGeneral:
Balance changes are as followed:

Increase power of AGT to 50 from 40.

MSAM faster reload to 45 from 100.

Increase APC damage vs light target to 105 from 100.

MLRS attack speed increase to 100 from 140.

Rocket Infantry damage increase to 35 from 30.

MCV Price increase to 4000 from 2000. Build time increase from 0:48 to
1:36.

MCV unit HP increase to 950 from 750. armor type Heavy from Light.

MCV now requires both Command Center and Repair Pad to produce.

Construction yard HP increase to 2000 from 1400.

Repair pad HP increase to 600 from 400.

Link discussion at:
http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&p=294421#294421

PR #10636, opened 2016-01-29
Refactor naming of Missile's speed-related properties

reaperrr:
`MaximumSpeed` renamed to just `Speed`.

`MinimumLaunchSpeed` and `MaximumLaunchSpeed` now default to `Speed`.

Note: To solve #10129, additional yaml adjustments might be necessary.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -41,14 +41,14 @@ namespace OpenRA.Mods.Common.Effects
 		[Desc("Maximum vertical launch angle (pitch).")]
 		public readonly WAngle MaximumLaunchAngle = new WAngle(128);
 
-		[Desc("Minimum launch speed in WDist / tick")]
-		public readonly WDist MinimumLaunchSpeed = new WDist(75);
+		[Desc("Minimum launch speed in WDist / tick. Defaults to Speed if -1.")]
+		public readonly WDist MinimumLaunchSpeed = new WDist(-1);
 
-		[Desc("Maximum launch speed in WDist / tick")]
-		public readonly WDist MaximumLaunchSpeed = new WDist(200);
+		[Desc("Maximum launch speed in WDist / tick. Defaults to Speed if -1.")]
+		public readonly WDist MaximumLaunchSpeed = new WDist(-1);
 
 		[Desc("Maximum projectile speed in WDist / tick")]
-		public readonly WDist MaximumSpeed = new WDist(384);
+		public readonly WDist Speed = new WDist(384);
 
 		[Desc("Projectile acceleration when propulsion activated.")]
 		public readonly WDist Acceleration = new WDist(5);
@@ -243,7 +243,9 @@ namespace OpenRA.Mods.Common.Effects
 		void DetermineLaunchSpeedAndAngleForIncline(int predClfDist, int diffClfMslHgt, int relTarHorDist,
 			out int speed, out int vFacing)
 		{
-			speed = info.MaximumLaunchSpeed.Length;
+			var minLaunchSpeed = info.MinimumLaunchSpeed.Length > -1 ? info.MinimumLaunchSpeed.Length : info.Speed.Length;
+			var maxLaunchSpeed = info.MaximumLaunchSpeed.Length > -1 ? info.MaximumLaunchSpeed.Length : info.Speed.Length;
+			speed = info.MaximumLaunchSpeed.Length > -1 ? info.MaximumLaunchSpeed.Length : info.Speed.Length;
 
 			// Find smallest vertical facing, for which the missile will be able to climb terrAltDiff w-units
 			// within hHeightChange w-units all the while ending the ascent with vertical facing 0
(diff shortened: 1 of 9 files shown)


Unit158:
What's the difference between this and acceleration?

GraionDilach:
This is supposed to be a maximum cap. I agree the tag name might not be the best.

reaperrr:
I know, but I couldn't come up with anything more precise (without making it too long, at least).

Maybe I should rename `Acceleration` -> `AccelerationStep` and `MaximumSpeedBonus` -> `MaximumAcceleration`.

Unit158:
Maximum velocity is probably a better name

pchote:
What if we went back to a `Speed` property that defines the maximum speed of the projectile, with an optional `MinimumLaunchSpeed`/`MaximumLaunchSpeed` (default to `Speed` if they aren't explicitly set).  We would then migrate the current `MaximumLaunchSpeed` values to `Speed` for the non-ts mods, and have TS calculate its appropriate minimum and maximum values.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -397,7 +397,7 @@ bool JammedBy(TraitPair<JamsMissiles> tp)
 
 		void ChangeSpeed(int sign = 1)
 		{
-			speed = (speed + sign * info.Acceleration.Length).Clamp(0, info.MaximumSpeed.Length);
+			speed = (speed + sign * info.Acceleration.Length).Clamp(0, info.MaximumLaunchSpeed.Length + info.MaximumSpeedBonus.Length);
```
pchote: For clarity: My suggestion above is basically regarding this line, which implicitly defines the cruising speed of the missile as the sum of two other speeds.  I'm suggesting that we use a single `info.Speed.Length` here, and then adjust the other speeds to make it work out.

reaperrr:
I'm done with implementing that suggestion, with one exception:

I'd like to see #10637 reviewed and merged first, to avoid rebase errors/regressions.

reaperrr:
Updated so you can review it, but I haven't upgraded the d2k weapon rules yet due to #10637.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -42,13 +42,13 @@ public class MissileInfo : IProjectileInfo
 		public readonly WAngle MaximumLaunchAngle = new WAngle(128);
 
 		[Desc("Minimum launch speed in WDist / tick")]
```
pchote: Could you please add a note that -1 means "use Speed"?

reaperrr: Good point, will do.

reaperrr:
Updated.

cjshmyr:
Tested all mods and re-ran the ra test map, :+1:

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -3432,6 +3432,25 @@ internal static void UpgradeWeaponRules(int engineVersion, ref List<MiniYamlNode
 					node.Value.Nodes.RemoveAll(x => x.Key == "Charges");
 				}
 
+				if (engineVersion < 20160129)
+				{
+					var mod = Game.ModData.Manifest.Mod.Id;
+					if (mod == "cnc" || mod == "ra" || mod == "d2k")
```
obrakmann: Can you please change this so it'll work with non-default mods?

pchote: This could be moved to an else for the ts if just below.

penev92:
Pchote's description above sounds like we'd need to document this.

reaperrr:
> Pchote's description above sounds like we'd need to document this.

I'm not sure where to document this and what to write.

reaperrr:
Updated upgrade rule to work for 3rd-party mods and rebased.

abcdefg30:
Needs a rebase again. :/

reaperrr:
Rebased.

cjshmyr:
@AoAGeneral [likes bleed's behavior in TD](https://github.com/OpenRA/OpenRA/pull/10552#issuecomment-172982004). Might be a misconception that the changed speeds were a feature and not a bug.

My vote is we took this PR as-is, restoring the engine to release's behavior -- but @AoAGeneral will have to do more balance work in TD. Other option is we treat TD like TS in this pull, though the current missile config in bleed seems very bogus to me.

I am fine either way.

pchote:
Lets focus on fixing the bug first, and then @AoAGeneral can iterate as required later.

AoAGeneral:
The missile bug was something I thought to be a feature yes. People have been wanting this added though because it buffs a lot of ballistics missiles. Recon Bikes have issues firing at air units and other unit types as they run. Orcas are left unaffect however which is fine, and E3 are now useful anti air units instead of easily dodged. The E3 also land their shots completely instead of a half land which creates half the damage done to the unit. (IE: Light tanks movement speeds)

Samsites and AGTs are unaffected by this due to their extremely long travel times as it is. Stealth Tanks is a slight buff but for their anti air capabilities is still dodgable. The speed of the missile can also be changed to fix this individual (IE: MSAM trajectory change).

in the forums http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=17430&postdays=0&postorder=asc&start=75 near the top of page 6 Jazz and Psydev also replied for a missile buff to the recon bikes.

cjshmyr:
The balance work is excellent. I'm in favor of fixing the bug across all mods, then letting you apply missile changes overtop of it to get the behavior back. It would be fairly simple to restore, but also makes sure the yaml is configured properly (right now it's not consistent with some missiles reaching ~384 speed, some not, etc)

AoAGeneral:
Yea just chatted with you on IRC. Since it is a bug it does make sense to get it fixed. (I was also wondering why it didnt affect Orca missiles either). Once this merge goes through then I will confer with you on how to set it back up. But I will check the bleed before making any adjustments and see how it looks too.

abcdefg30:
:+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/bfd4dc37658fb384985f4805a5042ac95713b342...6e0a4fb6e21cac82cb73b6e23873f97ff4621fc8)

PR #10642, opened 2016-01-29
Recovered flood fill functionality from the old editor

Reizoukodesu:
This commit recovers the flood fill functionality using a "Shift" Modifier. I couldn't see any code related to this, hope this recovery can be used.

### DIFF
--- a/OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
+++ b/OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
@@ -136,21 +136,22 @@ namespace OpenRA.Mods.Common.Widgets
 				return;
 
 			var queue = new Queue<CPos>();
-			var touched = new bool[map.MapSize.X, map.MapSize.Y];
+			var touched = new CellLayer<bool>(map);
 
 			var rules = map.Rules;
 			var tileset = rules.TileSets[map.Tileset];
 			var template = tileset.Templates[Template];
 
-			Action<CPos> maybeEnqueue = (newCell) =>
+			Action<CPos> maybeEnqueue = newCell =>
 			{
-				if (map.Contains(cell) && !touched[newCell.X, newCell.Y])
+				if (map.Contains(cell) && !touched[newCell])
 				{
 					queue.Enqueue(newCell);
-					touched[newCell.X, newCell.Y] = true;
+					touched[newCell] = true;
 				}
 			};
-			Func<CPos, bool> shouldPaint = (cellToCheck) =>
+
+			Func<CPos, bool> shouldPaint = cellToCheck =>
 			{
 				for (var y = 0; y < template.Size.Y; y++)
 				{
--- a/OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
+++ b/OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
@@ -8,6 +8,8 @@
  */
 #endregion
 
+using System;
+using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
 using OpenRA.Graphics;
@@ -74,20 +76,36 @@ namespace OpenRA.Mods.Common.Widgets
 			if (!painting)
 				return true;
 
+			if (mi.Event != MouseInputEvent.Down && mi.Event != MouseInputEvent.Move)
+				return true;
+
+			var cell = worldRenderer.Viewport.ViewToWorld(mi.Location);
+			var isMoving = mi.Event == MouseInputEvent.Move;
+
+			if (mi.Modifiers.HasModifier(Modifiers.Shift))
+			{
+				FloodFillWithBrush(cell, isMoving);
+				painting = false;
+			}
+			else
+				PaintCell(cell, isMoving);
+
+			return true;
+		}
+
+		void PaintCell(CPos cell, bool isMoving)
+		{
 			var map = world.Map;
 			var mapTiles = map.MapTiles.Value;
 			var mapHeight = map.MapHeight.Value;
-			var cell = worldRenderer.Viewport.ViewToWorld(mi.Location);
-
-			if (mi.Event != MouseInputEvent.Down && mi.Event != MouseInputEvent.Move)
-				return true;
 
 			var rules = map.Rules;
 			var tileset = rules.TileSets[map.Tileset];
 			var template = tileset.Templates[Template];
 			var baseHeight = mapHeight.Contains(cell) ? mapHeight[cell] : (byte)0;
-			if (mi.Event == MouseInputEvent.Move && PlacementOverlapsSameTemplate(template, cell))
-				return true;
+
+			if (isMoving && PlacementOverlapsSameTemplate(template, cell))
+				return;
 
 			var i = 0;
 			for (var y = 0; y < template.Size.Y; y++)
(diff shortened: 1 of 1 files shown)


pchote:
This doesn't appear to tile properly when you have a 2x2 or greater tile.  Can you make this use the same logic from the drag-fill, which does?

Sapena:
The tiling has been fixed

pchote:
Looks better, but now we have the opposite problem: flood filling with a NxN tile will do complete copies of the filled tile, even if they overlap tiles that shouldn't be replaced (most easily seen if you fill the large bridge tiles over the basic grass tiles - it will also replace cliffs and other decorations).

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
```diff
@@ -106,8 +123,66 @@ public bool HandleMouseInput(MouseInput mi)
...
+			{
+				var queuedCell = queue.Dequeue();
+				if (mapTiles[queuedCell].Type != replace.Type)
+					continue;
+
+				var previousCell = FindEdge(map, queuedCell, new CVec(-1 * template.Size.X, 0), replace);
+				var nextCell = FindEdge(map, queuedCell, new CVec(1 * template.Size.X, 0), replace);
+
+				for (var x = previousCell.X; x <= nextCell.X; x += template.Size.X)
+				{
```
pchote: You'll need to do a type check in this loop to fix the overfill issue.

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
```diff
@@ -74,20 +76,35 @@ public bool HandleMouseInput(MouseInput mi)
...
 				return true;
 
+			if (mi.Event != MouseInputEvent.Down && mi.Event != MouseInputEvent.Move)
+				return true;
+
+			var cell = worldRenderer.Viewport.ViewToWorld(mi.Location);
+			bool isMoving = mi.Event == MouseInputEvent.Move;
+
+			if (mi.Modifiers.HasModifier(Modifiers.Shift))
+			{
```
pchote: OpenRA code style is to omit braces around single line blocks.

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
```diff
@@ -74,20 +76,35 @@ public bool HandleMouseInput(MouseInput mi)
 			if (!painting)
 				return true;
 
+			if (mi.Event != MouseInputEvent.Down && mi.Event != MouseInputEvent.Move)
+				return true;
+
+			var cell = worldRenderer.Viewport.ViewToWorld(mi.Location);
+			bool isMoving = mi.Event == MouseInputEvent.Move;
```
pchote: OpenRA code style is to use `var` instead of explicit types.

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
```diff
@@ -106,8 +123,66 @@ public bool HandleMouseInput(MouseInput mi)
...
+				if (mapTiles[queuedCell].Type != replace.Type)
+					continue;
+
+				var previousCell = FindEdge(map, queuedCell, new CVec(-1 * template.Size.X, 0), replace);
+				var nextCell = FindEdge(map, queuedCell, new CVec(1 * template.Size.X, 0), replace);
+
+				for (var x = previousCell.X; x <= nextCell.X; x += template.Size.X)
+				{
+					PaintCell(new CPos(x, queuedCell.Y), isMoving);
+					var upperCell = new CPos(x, queuedCell.Y - (1 * template.Size.Y));
```
pchote: This looks a bit iffy to me.  Why aren't you just looping over all the cells in the template directly?

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
```diff
@@ -106,8 +123,66 @@ public bool HandleMouseInput(MouseInput mi)
...
+				}
+			}
+		}
+
+		CPos FindEdge(Map map, CPos cell, CVec direction, TerrainTile replace)
+		{
+			for (;;)
+			{
+				var newCell = cell + direction;
+				if (!map.Contains(newCell)) return cell;
```
pchote: Returns on their own line, please.

pchote:
Looks good overall, but there's still a few things to sort out.  Can you please squash the commits together once you're done with the fixups?

Sapena:
The filling does not overwrite the tiles on the edges anymore.

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
```diff
@@ -106,8 +124,80 @@ public bool HandleMouseInput(MouseInput mi)
...
+			var template = tileset.Templates[Template];
+
+			Action<CPos> maybeEnqueue = (newCell) =>
+			{
+				if (map.Contains(cell) && !touched[newCell.X, newCell.Y])
+				{
+					queue.Enqueue(newCell);
+					touched[newCell.X, newCell.Y] = true;
+				}
+			};
```
pchote: Style nit: newline after ending a block, please.

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
```diff
@@ -106,8 +124,80 @@ public bool HandleMouseInput(MouseInput mi)
...
+			Func<CPos, bool> shouldPaint = (cellToCheck) =>
```
pchote: Style nit: omit the parens when you have a single argument to an anonymous function

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
```diff
@@ -106,8 +124,80 @@ public bool HandleMouseInput(MouseInput mi)
...
+				return;
+
+			var queue = new Queue<CPos>();
+			var touched = new bool[map.MapSize.X, map.MapSize.Y];
+
+			var rules = map.Rules;
+			var tileset = rules.TileSets[map.Tileset];
+			var template = tileset.Templates[Template];
+
+			Action<CPos> maybeEnqueue = (newCell) =>
```
pchote: Style nit: omit the parens when you have a single argument to an anonymous function

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
```diff
@@ -106,8 +124,80 @@ public bool HandleMouseInput(MouseInput mi)
...
+			{
+				if (map.Contains(cell) && !touched[newCell.X, newCell.Y])
```
pchote: Indexing a 2D array with cell coordinates unfortunately isn't a valid thing to do.  This will crash the TS mod.  Use a `CellLayer<bool>` instead.

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
```diff
@@ -106,8 +124,80 @@ public bool HandleMouseInput(MouseInput mi)
...
+				if (!shouldPaint(queuedCell))
+					continue;
+
+				var previousCell = findEdge(queuedCell, new CVec(-1 * template.Size.X, 0));
+				var nextCell = findEdge(queuedCell, new CVec(1 * template.Size.X, 0));
+
+				for (var x = previousCell.X; x <= nextCell.X; x += template.Size.X)
+				{
+					PaintCell(new CPos(x, queuedCell.Y), isMoving);
+					var upperCell = new CPos(x, queuedCell.Y - (1 * template.Size.Y));
```
pchote: Likewise here, you need to be using MPos and associated functions rather than CPos, otherwise this will crash in TS and RA2.

pchote:
This works well now in the classic mods (good job!), but breaks in TS.  See the inline comments for some suggestions on how to fix that.

Sapena:
The flood fill seems to work now on every mod. Anything else to adapt?

OpenRA.Mods.Common/EditorBrushes/EditorTileBrush.cs
Sapena: I've tested the map editor with the TS mod and it seems to work now, should I change them nevertheless?

pchote: I guess this is probably fine.  It just means you're expanding diagonally instead of along the map axis.

pchote:
Confirmed that this now works in TS.  Code looks fine overall, so :+1:

Mailaender:
Nice! Works well. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8c16969a90785a50b1d7e98423e5ec619dbdbbdb...c5cdcd3c03eb29232f3c642a1e50b559c6c25d63)

Issue #10207, opened 2015-12-12
YAML merging issues


obrakmann:
Since #9710, yaml rules that used to be fine in our missions and mini-games don't work anymore (see [irc logs](http://logs.openra.net/?year=2015&month=12&day=12#16:10:29), for example).

abcdefg30:
See also https://github.com/OpenRA/OpenRA/pull/10206#issuecomment-164196266 or [this irc log](http://logs.openra.net/?year=2015&month=12&day=08#16:22:33).

pchote:
So much for unit tests catching future breaking :)

Can we get a testcase showing this working before #9710 and not working now?  I'll then look into what went wrong.

pchote:
I expect this is happening because of the split between resolving merges/removals (which happens in the lowest level miniyaml code) and resolving inheritance (which happens in the actor loading code).  Before #9710 we resolved all removals before considering inheritance, which meant that you couldn't remove nodes from inherited parents.  After #9710 we resolve all removals at the end, which I expect means that it *also* removes from children, as well as parents (which is equally bogus).  The fix for this will be to change the actor inheritance to use head recursion and fully resolve the removals at each step on the way out.

pchote:
Calling `MiniYaml.ApplyRemovals` inside `MergeWithParents` may be sufficient, but we need some testcases to verify this first.

RoosterDragon:
Try adding these to `ActorInfoTest.cs`

```c#
class MockStringInfo : MockTraitInfo { public string AString = null; }
```

```c#
		[TestCase(TestName = "Trait can be removed and later overridden")]
		public void TraitCanBeRemovedAndLaterOverridden()
		{
			var baseYaml = @"
^BaseA:
	MockStringInfo:
		Foo: ""Base""
Actor:
	Inherits: ^BaseA
	-MockStringInfo:
";
			var overrideYaml = @"
Actor:
	MockStringInfo:
		Foo: ""Override""
";

			var actorInfo = CreateActorInfoFromYaml("Actor", null, baseYaml, overrideYaml);
			Assert.IsTrue(actorInfo.HasTraitInfo<MockStringInfo>(), "Actor should have the MockStringInfo trait, but does not.");
			Assert.IsTrue(actorInfo.TraitInfo<MockStringInfo>().AString == "Override",
				"MockStringInfo trait has not been set with the correct override value for AString.");
		}
```

I didn't test this pre-#9710, but it fails on the current bleed with:
```
[...]

PR #10646, opened 2016-01-30
Implement special-case actor yaml merging.

pchote:
Proposed fix for #10207.  It passes @RoosterDragon's test case (after fixing a couple of errors in the test), and fixes fort lonestar in RA.  More testing is required in all mods: ~~there are side-effects (e.g. the turbocharged hovercraft in TD) which could eithery be a bug in new new parsing or old rules that are being parsed "correctly" for the first time.~~

As I explained in #10207 all these bugs are caused because the code tries to treat yaml merging and inheritance as separable steps, which is bogus.  This implements line-by-line rule parsing that immediately resolves inherits and removals.

### DIFF
--- a/OpenRA.Game/GameRules/RulesetCache.cs
+++ b/OpenRA.Game/GameRules/RulesetCache.cs
@@ -102,7 +102,7 @@ namespace OpenRA
 			var partial = files
 				.Select(s => MiniYaml.FromFile(s))
 				.Append(nodes)
-				.Aggregate(MiniYaml.MergePartial);
+				.Aggregate(MiniYaml.MergePartial2);
 
 			Func<MiniYamlNode, Dictionary<string, MiniYaml>, T> wrap = (wkv, wyy) =>
 			{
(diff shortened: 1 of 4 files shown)


OpenRA.Game/GameRules/RulesetCache.cs
```diff
@@ -101,7 +101,8 @@ public Ruleset Load(Map map = null)
 
 			var partial = files
 				.Select(s => MiniYaml.FromFile(s))
-				.Aggregate(nodes, MiniYaml.MergePartial);
+				.Append(nodes)
+				.Aggregate(MiniYaml.MergePartial);
```
pchote: The remaining bogosity is hiding somewhere in here.  We need the map rules to be aggregated last in order for the removals to be processed properly, but this then causes the wrong sub-child value to be taken.

A better solution may be to move the combined merge+inherit logic up into MiniYaml, and then throw out this MergePartial junk.

pchote:
I believe this should now behave correctly, but I need help to (dis)prove it.  After we know whether or not this works I will clean up the code.

pchote:
I'm going to take a different approach.

PR #10650, opened 2016-01-30
Remove fragile diplomacy.

### LINKED ISSUE #6254: Diplomacy change does not change teamtalk
We teamed up in the game, but could not communicate. He got my beacons, but did the others also get them?? Chat could not be turned into team-chat mode..

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3732550-diplomacy-change-does-not-change-teamtalk?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
This feature has never worked properly, and is implicated in many bugs and polish issues.  It is rarely used ingame, and the consensus on IRC is that it should be removed.

This removes the UI and lobby plumbing, but the actual stance changing logic is used by the monster tank madness mission (but nothing else).  We will need to think about how we handle missions before changing or removing the remaining code.

Closes #6254.


obrakmann:
I *think* Monster Tank Madness could be fixed by having an at-first unused player and then changing the owner of the mad tanks instead of using `SetStance`. Then `Player.SetStance` could be removed as well.

Could you make the lobby check box placement between TD and the other mods consistent, please? In TD, Build Off Allied ConYards, Short Game and Crates are switched compared the other mods (I think I'd prefer TD's layout: 1st column visibility, 2nd column balancing, 3rd column misc.). My muscle memory will thanks you :)

pchote:
If someone could throw together a patch that could try that for MTM I'd be happy to include it here and then remove the remaining code.

pchote:
Updated.

obrakmann:
https://github.com/pchote/OpenRA/pull/12

pchote:
Updated.

obrakmann:
:+1:

x-a-n-a-x:
While I can understand that this feature needs to be removed (since, according to you, it's buggy), I still think this Diplomacy option could be useful in third-party mods like this one: http://www.sleipnirstuff.com/forum/viewtopic.php?f=83&p=290395

Example: Player A (who manages policemen) wants to switch teams with Player B (who plays Firefighters), while Player C wants an AI bot to take over his Ambulance faction. What I mean is that this option makes sense for some mods (but maybe not for our 3/4 main mods, I can agree with that).

GraionDilach:
This is a really stupid decision. I'm not gonna rant about it - I did it enough times already.

Mailaender:
Note: some people apparently like it http://logs.openra.net/?year=2015&month=02&day=13#19:45:42 and the polish issues are rather minor and solely UI related. I could understand the removal if this causes desyncs or something similar serious. Also the Lua API breakage is probably not a good idea.

abcdefg30:
Tbh, I don't really like this removal too, at least the " Lua API breakage". But I won't insist on keeping it.

pchote:
Issues like http://logs.openra.net/?year=2016&month=01&day=28#21:54:27 are more than just minor UI polish.

obrakmann:
Diplomacy the way it's implemented now is broken beyond repair. It leads to asymmetric alliances, which causes all sorts of issues (chat, scoreboard, victory conditions etc.). While some interesting effects can be created using asymmetric alliances (Monster Tank Madness is one example), you just *cannot* give players the tools to create such situations themselves.

It's also highly annoying when players switch allegiance mid-game.

Doing it properly would involve adding a challenge-response mechanic. People would not be able to choose stances against individual players, but rather select a team and click on a "join" button. Then the remaining members of the parting team would need to authorize the player leaving their team, and all of the new team members would need to accept the request.

The only situation that I can come with where such a feature would help would be, for example, a 3 vs 2 match where one player from the smaller team disconnected for whatever reason. then one player of the 3 player team could switch to turn it into a 2v2. I don't think that this situation arises often enough to warrant the effort required to write this feature, when the players could just as easily create a new game.

GraionDilach:
2v2 random start pos, left vs right. A lot of Mental Omega games started like that when I actively played it online.

pchote:
@GraionDilach: That use case could be covered by a "Assign random spawns" option in the slot admin dropdown.

chrisforbes:
Good riddance to this nonsense.

chrisforbes:
:+1:

Micr0Bit:
:+1:  for removing diplomacy for now

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/095aebeed1a7456ef94ea280b7ca195ec3390310)

Mailaender:
There seems to be critical voices about a) the removal and b) the poorly communicated reasons like what specific problems were faced as those were not visible to players:

* http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19534
* http://www.openra.net/news/playtest-20160403/#comment-2605791696

Micr0Bit:
critical "voices" ? ... thats only one and the same guy ... "glanzor"

GraionDilach:
I was and am still critical as well, but I was aware noone else will care.

Micr0Bit:
@GraionDilach : i knew you gonna answering that :D 

well its really bad implemented ... i mean in a way that make it really "fucked up" ... as far as i remember you can join someone to be in his team ... aka "get his vision" and all (even use his buildradius) and then  just leave the team whenever you want (which is kinda stupid ... isnt it? ... cause all the revealed shroud just doesnt get reset and you could just destroy bases from inside) ... so its pretty much what paul said -> poorly implemented 

it comes all down to this " [ ] Troll / [x] Troll " using it in a multiplayer-match

abcdefg30:
IMHO that's not really a valid argument. If you don't want to use it in a public match, then don't.
I don't see a problem if people want to use it in there local (LAN) games like described in [this Sleipnir's Stuff thread](http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19534).

Micr0Bit:
wouldn't it make sense to reimplement it "properly" ...and have it working as it should ... instead of having it half broken in ?
i mean ive played "thousand's" of games ... and only rarely had any "diplomacy" games (for a good reason , thats why almost everyone is avoiding this option "except" new players who arent familiar with it being broken)
you want to keep a feature because otherwise 1 guys "heart would be broken" ? ( i still refer to glanzor only ... cause i would find it hard to believe that @GraionDilach has a heart , scnr :D )

abcdefg30:
> @abcdefg30 : wouldn't it make sense to reimplement it "properly" ...and have it working as it should ... instead of keeping it half broken ? dont you think that could probably even encourage someone to do it finally ? 

Well, I didn't say that I'm against doing it properly and thus also didn't 👎 this pr. ;)

chrisforbes:
:+1: Good riddance to this. Accumulating rarely used things "as an option" isn't harmless -- they all have a maintenance & testing overhead.

PR #10651, opened 2016-01-30
Add xmldoc to Widget.HandleMouseInput

phrohdoh:
cc @pchote, @RoosterDragon

### DIFF
--- a/OpenRA.Game/Widgets/Widget.cs
+++ b/OpenRA.Game/Widgets/Widget.cs
@@ -108,6 +108,9 @@ namespace OpenRA.Widgets
 			return handled;
 		}
 
+		/// <summary>Possibly handle keyboard input (if this widget has keyboard focus)</summary>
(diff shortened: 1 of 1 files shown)


RoosterDragon:
Sure why not :+1:

obrakmann:
I don't quite see the point of documenting just that one method. At least do `HandleKeyPress` as well.

Phrohdoh:
Are you going to require all of these too?

```csharp
		public static bool HandleInput(MouseInput mi)
		public static bool HandleTextInput(string text)
		public bool HandleMouseInputOuter(MouseInput mi)
		public virtual bool HandleKeyPress(KeyInput e) { return false; }
		public virtual bool HandleKeyPressOuter(KeyInput e)
		public virtual bool HandleTextInput(string text) { return false; }
		public virtual bool HandleTextInputOuter(string text)
		public override bool HandleKeyPress(KeyInput e) { return OnKeyPress(e); }
		public override bool HandleMouseInput(MouseInput mi)
```

obrakmann:
no, just `HandleKeyPress`, since it's essentially the same text. :+1:

PR #10653, opened 2016-01-30
Moved shared metrics to mods/commons.

Mailaender:
Avoids redundancy and bit rot in mods which are not maintained in this repository as missing metrics definitions always result in crashes in the startup sequence. https://github.com/OpenRA/ra2/issues/96


obrakmann:
Makes sense :+1: 

BTW: what did we agree on to use as path spec for the common mod? `./mods/common/whatever` or `common:whatever`? We should clean all those up eventually in one go.

pchote:
@obrakmann: I cover that on my filesystem branches.

abcdefg30:
:+1: / :white_check_mark:

PR #10654, opened 2016-01-30
Add 'bool instantiateLogicObjects' to widget loading methods

phrohdoh:
This allows one to open a widget (possibly root) without instantiating their logic objects which may change the UI's layout and interaction with the user.

I use this in my WIP designer branch.


pchote:
My preference would be to include this as part of the PR that first uses this option.  Otherwise it just sits around doing nothing.

Phrohdoh:
```
[01:53:06]	<Phrohdoh> Points against both of my PRs were fair and valid so I've closed them.
[01:53:18]	<Phrohdoh> I'll try to get a prototype I feel I am able to show soonish to validate my approach. 
```

PR #10657, opened 2016-01-31
Improve sync performance

RoosterDragon:
Previously, generating the sync hash for each syncable trait required a lookup to find the hash function every tick, as well as lookups for the syncable traits.

Instead, we now cache these lookups on actor creation as they do not change. This means we avoid repeated lookups for the traits and the hash functions.

On the RA shellmap, this reduces time spent syncing the world state from 9.65% of total CPU to 6.88%.

### DIFF
--- a/OpenRA.Game/Sync.cs
+++ b/OpenRA.Game/Sync.cs
@@ -38,12 +38,11 @@ namespace OpenRA
 			{ typeof(int2), ((Func<int2, int>)HashInt2).Method },
 			{ typeof(CPos), ((Func<CPos, int>)HashCPos).Method },
 			{ typeof(CVec), ((Func<CVec, int>)HashCVec).Method },
-			{ typeof(WDist), ((Func<WDist, int>)Hash).Method },
-			{ typeof(WPos), ((Func<WPos, int>)Hash).Method },
-			{ typeof(WVec), ((Func<WVec, int>)Hash).Method },
-			{ typeof(WAngle), ((Func<WAngle, int>)Hash).Method },
-			{ typeof(WRot), ((Func<WRot, int>)Hash).Method },
-			{ typeof(TypeDictionary), ((Func<TypeDictionary, int>)HashTDict).Method },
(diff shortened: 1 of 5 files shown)


OpenRA.Game/Sync.cs
```diff
@@ -62,15 +67,13 @@ static void EmitSyncOpcodes(Type type, ILGenerator il)
 				il.Emit(OpCodes.Ldc_I4, 0x555);
 				il.MarkLabel(l);
 			}
-			else if (type.HasAttribute<SyncAttribute>())
```
abcdefg30: If we don't use this, could you also remove it from all other places?

abcdefg30:
:+1: / :white_check_mark: otherwise. Looks fine ingame and the game desyncs correctly.

OpenRA.Game/Sync.cs
Mailaender: The trait lookup isn't gone, just cached at https://github.com/OpenRA/OpenRA/pull/10657/files#diff-25e3f19d154f93bcb747e9f12c252028R126.

Mailaender: Nevermind, I also got confused, but `SyncAttribute` is required for lint rules and used in `SyncReport` .

RoosterDragon: These particular lines are dead code - we only apply the sync attribute to fields and properties (this is enforced by the attribute usage). In this context `type` will be either a class or struct definition. Therefore, this condition is always false.

I'm not removing the sync attribute or changing how it works - I'm just removing this useless code.

Mailaender:
Jackpot!

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c5cdcd3c03eb29232f3c642a1e50b559c6c25d63...3d4a2e2597d270145655642d7c67c56ba289b12f)

PR #10664, opened 2016-01-31
Move MiniYaml into its own OpenRA.Markup project

phrohdoh:
This will allow other projects/tools to depend on miniyaml structures and parsing without needing the entirety of the engine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,9 +104,19 @@ endif
 #
 # Core binaries
 
+# OpenRA.Markup
+markup_SRCS := $(shell find OpenRA.Markup/ -iname '*.cs')
+markup_TARGET = OpenRA.Markup.dll
+markup_KIND = library
+#markup_DEPS = none
+markup_LIBS = $(COMMON_LIBS)
+PROGRAMS += markup
+markup: $(markup_TARGET)
+
 game_SRCS := $(shell find OpenRA.Game/ -iname '*.cs')
 game_TARGET = OpenRA.Game.exe
 game_KIND = winexe
+game_DEPS = $(markup_TARGET)
 game_LIBS = $(COMMON_LIBS) $(game_DEPS) thirdparty/download/SharpFont.dll
 game_FLAGS = -win32icon:OpenRA.Game/OpenRA.ico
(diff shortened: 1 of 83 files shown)


Phrohdoh:
Rebased but now I need to remove `OpenRA.Mods.Common/UtilityCommands/LegacyMapImporter.cs` because the original removal must have been lost in the rebase.

Phrohdoh:
Removed and ready for review.

obrakmann:
Similar to [this comment](https://github.com/OpenRA/OpenRA/pull/10654#issuecomment-177563890), I think it would be better to let this be part of a PR that actually has code that uses it. Otherwise it's just needless churn in preparation for code that may or may not ever turn up.

penev92:
My two cents (so this doesn't get merged because "it works" without at least this):
>[17:16:26]	< penev> I guess my question is "why should this change be made upstream" 
...
[17:34:47]	< penev> `¯\_(ツ)_/¯` if you ask me it's best where it is 
[17:35:24]	< penev> having a DLL with 1 class that is used by 1 other project just for the sake of a downstream experiment...
[17:35:31]	< penev> doesn't justify the upstream wtf IMO 

And just to be thorough - regarding the "to move it out of the engine":
>[17:18:23]	< penev> if this really needs to move (which I don't see), the (possibly not-yet-existing) FileFormats project would probably be the place for it 
[17:32:14]	< pchote> penev: there is a difference between a mod-level fileformats dll and an engine-level fileformats dll 
[17:32:27]	< pchote> mod-level is good, because its all implicit
[17:32:32]	< pchote> nothing actually cares about the things in it
[17:32:37]	< pchote> because its all via interfaces 

This one here doesn't actually abstract away from the MiniYaml code specifics.

Phrohdoh:
```
[01:53:06]	<Phrohdoh> Points against both of my PRs were fair and valid so I've closed them.
[01:53:18]	<Phrohdoh> I'll try to get a prototype I feel I am able to show soonish to validate my approach. 
```

chrisforbes:
Nak for calling something 'markup' that isn't.

Issue #5017, opened 2014-04-02
Ugly kerning in new mod switcher.


Unit158:
##

RoosterDragon:
Is this still an issue? It looks OK to me these days.

![](http://i.imgur.com/nLpKgKM.png)

pchote:
I expect it is.  We still aren't using the kerning hints that are built into the font files.

Mailaender:
The mod chooser was removed. https://github.com/OpenRA/OpenRA/issues/12554

PR #10666, opened 2016-01-31
Updated NuGet dependencies

Mailaender:
- https://www.nuget.org/packages/Mono.Nat/
- https://www.nuget.org/packages/SharpFont/

should both be safe as they don't come with API or Framework changes.

### DIFF
--- a/thirdparty/fetch-thirdparty-deps.ps1
+++ b/thirdparty/fetch-thirdparty-deps.ps1
@@ -52,7 +52,7 @@ if (!(Test-Path "MaxMind.GeoIP2.dll"))
 if (!(Test-Path "SharpFont.dll"))
 {
 	echo "Fetching SharpFont from NuGet."
-	./nuget.exe install SharpFont -Version 3.0.0 -ExcludeVersion
+	./nuget.exe install SharpFont -Version 3.1.0 -ExcludeVersion
 	cp SharpFont/lib/net20/SharpFont* .
 	cp SharpFont/config/SharpFont.dll.config .
 	rmdir SharpFont -Recurse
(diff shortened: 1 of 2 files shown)


pchote:
What changes does this bring?

Mailaender:
See the NuGet release notes.

penev92:
Sooo... do the new versions give us anything?

Mailaender:
I need the SharpFont Face.GetKerning() bugfix for https://github.com/OpenRA/OpenRA/issues/5017 and I hope using the latest Mono.Nat version solves some problems with our UPnP based automatic port forwarding.

abcdefg30:
Game still works. :white_check_mark:

obrakmann:
It does indeed. :+1: 

I'd still really prefer if we dropped UPNP support, btw. I hate that we're encouraging users to enable that freaking security nightmare on their routers. That shit belongs to be dead and buried six feet underground.

That's not even mentioning that it only works for a minority of users anyway. It causes way more trouble than it's worth.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/a5f910dab12d63d2c18d0b4d7c91ce055c0bd549)

GraionDilach:
Manual port forwarding is also as user-friendly as the Chinese language for Average Joe, @obrakmann.

Mailaender:
Disabling it by default seems to be a good compromise.

GraionDilach:
It's disabled by default already.

pchote:
It's not just disabled by default: it's disabled by default, with a second checkbox (*also* disabled by default) hidden in the advanced settings menu that you must change before the game will let you enable it.

This is the usability equivalent of putting it on display at the bottom of a locked filing cabinet stuck in a disused lavatory with a sign on the door saying _beware of the leopard_.

PR #10667, opened 2016-01-31
Updated to use the unified Mono compiler suite

Mailaender:
> http://www.mono-project.com/docs/about-mono/languages/csharp/:
> Starting with Mono version 2.11 a new unified compiler mcs is available.

Time to update our `Makefile` as well. We are a bit late actually. Updated the build documentation, too. The only dependency from http://packages.ubuntu.com/trusty/cli-common-dev we need is mono-devel so I reference that directly.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -34,7 +34,6 @@ Type `sudo make install-all` for system wide installation. Run `make install-lin
 Debian/Ubuntu
 -------------
 
-* mono-mcs
 * libmono-system-windows-forms4.0-cil
 * nuget
 * mono-devel
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -34,7 +34,6 @@ Type `sudo make install-all` for system wide installation. Run `make install-lin
 Debian/Ubuntu
 -------------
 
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -34,10 +34,10 @@ Type `sudo make install-all` for system wide installation. Run `make install-lin
 Debian/Ubuntu
 -------------
 
-* mono-dmcs
+* mono-mcs
 * libmono-system-windows-forms4.0-cil
 * nuget
-* cli-common-dev (>= 2.10)
+* mono-devel
```
pchote: shouldn't this include mono-mcs?

Mailaender: Indeed. http://packages.ubuntu.com/trusty/mono-devel Cleaned it up further.

pchote:
Looks reasonable to me.  Breaking default support for 2.10 is a feature at this point.

pchote:
Still works on OSX, so :+1:

obrakmann:
Packaging on linux works as well. :+1:

PR #10669, opened 2016-01-31
Update VS version in solution file

RoosterDragon:
VS 2015 Update 1 changed the version number slightly. Committing this stops people on an up-to-date Visual Studio from having to reset this change whenever VS paves over the solution file with changes.


abcdefg30:
:+1:

PR #10670, opened 2016-02-01
Disable 1 human versus bots games

### LINKED ISSUE #10660: Replace Server.LockBots option with Server.DisableSinglePlayer
The LockBots option was originally added in response to players using dedicated servers to play 1vbot games (back in the days when only TD had the skirmish mode).  Disabling bots even for proper multiplayer games is overkill, and rather unfair.

This has become topical again with the recent flood of players who have been asking / complaining about port forwarding.  Our best response "You don't need to - we provide dedicated servers for this" falls flat when one of the main game options for casual players is explicitly disallowed on them.

This is easy to fix and gives a big usability win, so i'm suggesting that we prioritise this for the next release. We already have code in place to disable the game starting if specific lobby slots are not full, and this can be extended with an additional check for > 1 human players.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30384699-replace-server-lockbots-option-with-server-disablesingleplayer?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Xinayder:
Changed the Server.LockBots setting to Server.DisableSinglePlayer.

If the setting is enabled, and there is only one client in the server,
the game won't start.

Fixes #10660

### DIFF
--- a/OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
+++ b/OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
@@ -67,6 +67,10 @@ namespace OpenRA.Mods.Common.Server
 			if (server.LobbyInfo.Slots.Any(sl => sl.Value.Required && server.LobbyInfo.ClientInSlot(sl.Key) == null))
 				return;
 
+			// Does server have only one player?
+			if (server.Settings.DisableSinglePlayer && playerClients.Count() == 1)
+				return;
+
 			server.StartGame();
 		}
 
@@ -112,6 +116,13 @@ namespace OpenRA.Mods.Common.Server
 							server.LobbyInfo.ClientInSlot(sl.Key) == null))
 						{
 							server.SendOrderTo(conn, "Message", "Unable to start the game until required slots are full.");
+							return true;
+						}
+
+						if (server.Settings.DisableSinglePlayer &&
+							server.LobbyInfo.Clients.Where(c => c.Bot == null && c.Slot != null).Count() == 1)
+						{
+							server.SendOrderTo(conn, "Message", "Unable to start the game until another player joins.");
 							return true;
 						}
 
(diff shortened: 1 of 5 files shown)


OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
```diff
@@ -67,6 +67,10 @@ static void CheckAutoStart(S server)
 			if (server.LobbyInfo.Slots.Any(sl => sl.Value.Required && server.LobbyInfo.ClientInSlot(sl.Key) == null))
 				return;
 
+			// Does server have only one player?
+			if (playerClients.Count() == 1 && server.Settings.Dedicated && server.Settings.DisableSingePlayer)
```
pchote: You'll need a similar check in the "startgame" order handler too.

pchote: Do we need the explicit dedicated check?

RockyTV:
Update: removed explicit dedicated check, and added the singleplayer check to the "startgame" order handler.

OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
```diff
@@ -115,6 +119,13 @@ public bool InterpretCommand(S server, Connection conn, Session.Client client, s
 							return true;
 						}
 
+						if (server.Settings.DisableSinglePlayer &&
+							server.LobbyInfo.Clients.Where(c => c.Bot == null && c.Slot != null).Count() == 1)
+						{
+							server.SendOrderTo(conn, "Message", "Unable to start the game because singleplayer games are disabled on this server.");
```
pchote: "Unable to start the game until another player joins" would be a bit more concise and matches the message above.

Mailaender:
It tells me that single-player games aren't allowed, but offers the start game button:

image

Pressing start then disables the UI:

image

This is a bit weird from a usability standpoint.

RockyTV:
@Mailaender I think it will disable the button now if there is only one player in the server and the server does not allow 1vbot games.

Mailaender:
Good idea. :+1:

abcdefg30:
Um,why is "Change Map" suddenly disabled (second image)?

obrakmann:
@abcdefg30: I think that happened in the previous iteration where the start button wasn't disabled yet and clicking it would set the `gameStarting` flag, which disables the change map button as a side effect. It doesn't happen in the current iteration.

obrakmann:
Also :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/1e0ec9e90baae0426aed0f815e725d8f9170d4ca)

PR #10678, opened 2016-02-02
Fixed aircraft doing an unnecessary wait loop

### LINKED ISSUE #10029: Airplanes given "return to base" command do an unnecessary loop (if already flying towards base)
(playtest: 20151114, and probably in previous versions too)
If an airplane (Yak, Mig... but not helicopters) is given a "return to base command" (default hotkey = "F") it sometimes does an unnecessary loop before returning home.
It turns out that it happens (only) when the airplane is already flying in the "right" direction (towards own base). 
 
This is quite annoying if it happens when the aircraft is in danger (over enemy troops) and since it is unexpected.
 
Here's a short replay. Watch the top, green player. When the Yak flies away from base (first right, then left) and is given the "return to home" command it works nicely. But it "fails" the third time (flying up, just before replay ends) since the Yak was already facing the base.
http://www.megafileupload.com/5v52/yakTurn.orarep

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28357578-airplanes-given-return-to-base-command-do-an-unnecessary-loop-if-already-flying-towards-base?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bateramos:
The problem that i found was that the airplanes when facing the airport try to move to very close position, making a strange loop. I resolve this by verifying if the position was too close to the airplane, it will ignore.


Mailaender:
Can you squash the fixup commits together please?

bateramos:
sorry for the confusion, but can you cancel this pull request? I create a new branch with only this bug fix and clean up the commit messages. https://github.com/OpenRA/OpenRA/pull/10695

Issue #10643, opened 2016-01-30
Possibility to switch assets quickly


Sapena:
Using the editor I've noticed that it is somehow time expensive to switch between the assets, for example when placing some beach tiles and then some water cliffs and back again.

My idea is to be able to switch between the assets by right clicking an existing asset on the map. Another option would be to add hotkeys, if it doesn't exist yet, couldn't find any reference to this in the code.

I created an example behavior, see [switch_assets_fast](https://github.com/Sapena/OpenRA/tree/switch_assets_fast) 913d24491333757972f9700ee44bf09a47ac2c63

pchote:
Do you mean like an eye-dropper tool? (based on the code, I assume yes).  This is a great idea, but I think it would be better to have an explicit button for entering this mode like we have for copy/paste, instead of making it a modifier of the default cursor type.

Sapena:
I like the idea with the explicit button, but I would add a hotkey too. So the user can toggle the eye-dropper tool quickly without moving the mouse too much.

Narg6:
Wow! This is a very good idea indeed.

Sapena:
![quick_asset_selection](https://cloud.githubusercontent.com/assets/2463962/12699242/9a56cbe4-c7b4-11e5-9406-8121fbf4e8a9.gif)
Just to clarify what I meant.

Issue #10652, opened 2016-01-30
Random tileset terrain painting


Sapena:
A friend of mine [topaxi](https://github.com/topaxi) had the idea to paint cliffs randomly of the same type. I couldn't find any functionality or an issue to something similar, so I've created a prototype function:

https://github.com/Sapena/OpenRA/tree/similar_terrain_tilesets b86ea5a9d0a422c9da83da76a617939f88688155

The idea is to be able to drag a line of cliffs and the editor switches automatically the templates, but only using similar tiles. The prototype for now only works with the vertical and horizontal cliffs (Excluding the diagonal cliffs):
* Horizontal: 135-141, 149-155
* Vertical: 142-148, 156-162

The functionality could be applied to other kind of tiles, like water cliffs, roads, etc. Another extended and nice functionality would be, that the editor is able to handle displacements, like moving upwards while dragging a horizontal line.

Is this idea usable? Or just kind of overkill?

MustaphaTR:
There is a cliff drawer in Final Alert 2 even i don't use that much. That would be good to have one in OpenRA's Editor.

Sapena:
Forgot to mention that the prototype is only roughly tested for the RA mod.

Sapena:
![similar_tiles](https://cloud.githubusercontent.com/assets/2463962/12699170/814e2ba8-c7b2-11e5-9690-e920d1d91809.gif)
Here a little demonstration of the idea I tried to describe

pchote:
I left a comment in #10655 about how we could implement this in a cross-mod compatible way.

Narg6:
Another good idea for map makers !

pchote:
If we do this as its own brush type (as per #10655) I suggest that we render the preview as a 2x2 block of the white selection overlay to indicate that you're not placing any specific cliff type.

Sapena:
Good input, so it is easier to stay within the line and it is less confusing.

LipkeGu:
#8436

PunkPun:
This has been implemented into the dev version

PR #10681, opened 2016-02-02
Switch assets fast

Reizoukodesu:
Added a eyedropper tool toggle button for the editor which allows to switch quickly between tileset categories and select the clicked tile template as current brush. See #10643


pchote:
Could you please try and make this work as an `IEditorBrush` implementation, instead of adding the logic to the viewport widget?

Sapena:
I've extracted the eyedropper brush logic to a new brush type.
I also added a visual feedback for the tool, I've called it EditorHighlightLayer. This could be used for other functions as well, as mentioned in [#10652](https://github.com/OpenRA/OpenRA/issues/10652#issuecomment-177339989). Should I extract the EditorHighlightLayer functionality to a different pull request?

OpenRA.Mods.Common/Traits/World/EditorHighlightLayer.cs
```diff
@@ -0,0 +1,81 @@
...
+		public readonly string Image = "editor-overlay";
+
+		[SequenceReference("Image")]
+		[Desc("Sequence to use for the highlight overlay.")]
+		public readonly string HighlightSequence = "highlight";
+
+		public virtual object Create(ActorInitializer init) { return new EditorHighlightLayer(init.Self, this); }
+	}
+
+	public class EditorHighlightLayer : IWorldLoaded, IPostRender
```
pchote: There's a lot of duplication between this and the existing `EditorSelectionLayer`. Can you merge the two together?

OpenRA.Mods.Common/Widgets/EditorViewportControllerWidget.cs
```diff
@@ -35,7 +36,28 @@ public EditorViewportControllerWidget(World world, WorldRenderer worldRenderer)
 			CurrentBrush = defaultBrush = new EditorDefaultBrush(this, worldRenderer);
 		}
 
+		public void ToggleEyedropping()
```
pchote: The modifications in this file aren't needed.  Just set the brush type directly from the `MapEditorLogic` (see how the copy/paste brush does it).

OpenRA.Mods.Common/Widgets/Logic/Editor/TileSelectorLogic.cs
```diff
@@ -50,6 +52,13 @@ public TileSelectorLogic(Widget widget, WorldRenderer worldRenderer, Ruleset mod
 			IntializeTilePreview(widget, worldRenderer, tileset, categories.First());
 		}
 
+		public void SwitchCategory(WorldRenderer worldRenderer, TileSet tileset, string category)
+		{
+			var tileCategorySelector = tileSelectorContainer.Get<DropDownButtonWidget>("TILE_CATEGORY");
```
pchote: This can be cached from the constructor.

mods/ra/chrome/editor.yaml
```diff
@@ -394,3 +394,9 @@ Container@EDITOR_WORLD_ROOT:
 			Align: Left
 			Font: Bold
 			Contrast: true
+		Button@EYEDROPPER_BUTTON:
+			X: 665
+			Width: 120
+			Height: 25
+			Text: Eyedropper tool
```
pchote: This label is too long, and overlaps the minimap.  I suggest renaming this to "Select terrain" and moving the button.

mods/ra/rules/world.yaml
```diff
@@ -174,3 +174,4 @@ EditorWorld:
 	EditorActorLayer:
 	EditorResourceLayer:
 	EditorSelectionLayer:
+	EditorHighlightLayer:
```
pchote: Please add the feature to all the default mods, not just RA.

OpenRA.Mods.Common/Traits/World/EditorHighlightLayer.cs
Sapena: I couldn't find a good way to merge them together, so I extracted the shared logic to a parent class, is this ok?

Sapena:
Hope all issues are fixed. Let me know if there's something more to do.

Mailaender:
Can you squash the fixup commits using `git rebase`?

OpenRA.Mods.Common/Traits/World/EditorCellRegionLayer.cs
```diff
@@ -0,0 +1,60 @@
...
+				return;
+
+			palette = wr.Palette(Info.Palette);
+		}
+
+		public abstract void Clear();
+
+		public virtual void RenderAfterWorld(WorldRenderer wr, Actor self)
+		{
+			if (wr.World.Type != WorldType.Editor)
```
pchote: You immediately return if this is false, too, so this check is a no-op.

pchote:
Any progress here @Sapena?

reaperrr:
There hasn't been any activity in two months, so I'm closing this for now to clear the PR queue.
@Sapena Feel free to reopen it when you have the time to rebase and adress the requests.

PR #10687, opened 2016-02-03
Load mod yaml via the virtual filesystem.

pchote:
This changes our rule/sequence/chrome/etc loaders to use the virtual filesystem code instead of .net filesystem code.  This is the next step towards adding support for oramod packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ sudo: false
 cache:
   directories:
   - thirdparty/download
+  - Support/Content
 
 addons:
   apt:
@@ -26,12 +27,14 @@ env:
 
 # Fetch dependencies
 # Run the build script
+# Fetch mod assets
 # Check source code with StyleCop
 # call OpenRA to check for YAML errors
 # Run the NUnit tests
 script:
  - travis_retry make all-dependencies
  - make all
+ - travis_retry ./thirdparty/fetch-travis-assets.sh
  - make check
  - make check-scripts
  - make test
--- a/OpenRA.Game/Scripting/ScriptContext.cs
+++ b/OpenRA.Game/Scripting/ScriptContext.cs
@@ -11,6 +11,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using Eluant;
@@ -158,7 +159,7 @@ namespace OpenRA.Scripting
 				.ToArray();
 
 			runtime.Globals["GameDir"] = Platform.GameDir;
-			runtime.DoBuffer(Game.ModData.ModFiles.Open(Platform.ResolvePath(".", "lua", "scriptwrapper.lua")).ReadAllText(), "scriptwrapper.lua").Dispose();
+			runtime.DoBuffer(File.Open(Platform.ResolvePath(".", "lua", "scriptwrapper.lua"), FileMode.Open).ReadAllText(), "scriptwrapper.lua").Dispose();
 			tick = (LuaFunction)runtime.Globals["Tick"];
 
 			// Register globals
--- a/OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
@@ -44,7 +44,6 @@ namespace OpenRA.Mods.Common.UtilityCommands
 		{
 			// HACK: The engine code assumes that Game.modData is set.
 			Game.ModData = modData;
-			Game.ModData.MountFiles();
 
 			Rules = Game.ModData.RulesetCache.Load();
 
--- a/Makefile
+++ b/Makefile
@@ -254,15 +254,19 @@ test: utility mods
 	@echo
 	@echo "Testing Tiberian Sun mod MiniYAML..."
 	@mono --debug OpenRA.Utility.exe ts --check-yaml
+	@mono --debug OpenRA.Utility.exe ts --check-sequence-sprites
 	@echo
 	@echo "Testing Dune 2000 mod MiniYAML..."
 	@mono --debug OpenRA.Utility.exe d2k --check-yaml
+	@mono --debug OpenRA.Utility.exe d2k --check-sequence-sprites
 	@echo
 	@echo "Testing Tiberian Dawn mod MiniYAML..."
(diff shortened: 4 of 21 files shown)


pchote:
It turns out that this change (and the plumbing we need for oramods in general) is fundamentally incompatible with the way our travis checks work.  Currently travis relies on not mounting anything in the filesystem as a workaround for not having a copy of the game assets.

We have three options:

1. Mark the downloaded content packages optional (using `~`) in the packages list, and rely on the load screen content checks to stop the game from loading without them.  This is a bit shit, but is the least invasive change.
2. Make travis download the mod content.  We can have travis cache the content files, which means this will have negligible effect on our infrastructure.  This would also let us remove hacks from the sequence parsers (which needs to defer any missing artwork errors to work around travis), and lets us add additional lint checks for sequence/sound references to missing assets. This is my preferred option because it makes our lint checks a lot more useful.  The downside is that mods that rely on proprietary assets that they can't distribute will not be able to use travis for lint checking.
3. Keep our current inconsistent mix of direct filesystem and vfs calls, and abandon our oramod plans (not realistic, but included for completeness).

Phrohdoh:
I think the obvious choice here for the engine is 2.
Thirdparty mods that want lint checking (or CI in general) are probably advanced enough to build their own system or find a solution that does work for them (local git hooks?).

This will bite us if/when RA2 is moved into OpenRA/OpenRA.

pchote:
>  The downside is that mods that rely on proprietary assets that they can't distribute will not be able to use travis for lint checking.

A solution for this would be for these mods to generate a dummy content package: files which match the structure and size of the proper files, but garbage content (e.g. fully transparent pixels for sprites, silence for sound files).

GraionDilach:
I guess I'll be participating in doing such assets for RA2, so size as frame size or bytesize.

pchote:
@GraionDilach best to have that discussion later in its own ticket if we decide to go with option 2.

pchote:
Added asset downloading and caching for Travis.

OpenRA.Game/Scripting/ScriptContext.cs
```diff
@@ -158,7 +159,7 @@ public sealed class ScriptContext : IDisposable
 				.ToArray();
 
 			runtime.Globals["GameDir"] = Platform.GameDir;
-			runtime.DoBuffer(Game.ModData.ModFiles.Open(Platform.ResolvePath(".", "lua", "scriptwrapper.lua")).ReadAllText(), "scriptwrapper.lua").Dispose();
+			runtime.DoBuffer(File.Open(Platform.ResolvePath(".", "lua", "scriptwrapper.lua"), FileMode.Open).ReadAllText(), "scriptwrapper.lua").Dispose();
```
penev92: I thought we were moving away from the .NET/Mono file system in order to use our own?

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -44,7 +44,6 @@ public virtual void Run(ModData modData, string[] args)
 		{
 			// HACK: The engine code assumes that Game.modData is set.
```
penev92: So when do we remove those?

OpenRA.Game/Scripting/ScriptContext.cs
pchote: This is a core engine file (like the glsl shaders), so isn't referenced from the mod yaml.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
pchote: When the engine code stops assuming Game.modData is set :(

pchote: There's about 200 more uses to remove first.

penev92:
Fine :+1:

Mailaender:
You can enable `--check-sequence-sprites` checks now.

pchote:
Removed the two levels of laziness that used to hide sequence-asset mismatches from travis.  The third level of lazyness (at the Ruleset leve) remains, so this shouldn't affect the ingame perf.  Also removed the now-unnecessary `--check-sequence-sprites` utility command: sequence errors are now caught by the existing map rules lint pass.

Mailaender:
Looks like you forgot about mounting map assets like https://github.com/OpenRA/OpenRA/tree/bleed/mods/ra/maps/bomber-john

Mailaender:
Depends on https://github.com/OpenRA/ra2/pull/108 which currently is waiting for Travis admins to whitelist 7z for me.

pchote:
> Looks like you forgot about mounting map assets like https://github.com/OpenRA/OpenRA/tree/bleed/mods/ra/maps/bomber-john

Indeed I did.  Fixing this is non-trivial (we need to be able to mount the selected map in the lobby into its own isolated filesystem before we can validate its rules) and well outside the scope of this PR, so I have removed the commit that enables sequence linting and introduces the problem.

penev92:
This PR *can't* possibly depend on a downstream fix for a random thirdparty mod. That's not how this works.

.travis.yml
```diff
@@ -10,6 +10,7 @@ sudo: false
 cache:
   directories:
   - thirdparty/download
+  - Support
```
Mailaender: You probably should only cache `Support/Content`.

pchote:
Changed travis to only cache Support/Content.

Mailaender:
I think you can enable sprite sequence checks again. Without the refactoring commit `mono --debug OpenRA.Utility.exe ra --check-sequence-sprites` works fine on my machine using this pull request.

Mailaender:
Already came up with a solution for the Red Alert 2 CI checks. https://github.com/OpenRA/ra2/pull/108

pchote:
Done.

obrakmann:
Custom maps don't seem to get loaded on this branch.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/b286f1fd6d7b18e6508ba182134a1f5a5fceaa29)

pchote:
Turns out that this also broke dedicated servers and the resource site.  We probably want to go with option (1) in the short term so that we don't need a copy of the assets *everywhere*.

PR #10688, opened 2016-02-03
Added Factions to Voiced.

Mailaender:
Required for Red Alert 2 which has complicated rules for faction specific voices.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Voiced.cs
+++ b/OpenRA.Mods.Common/Traits/Voiced.cs
@@ -8,6 +8,7 @@
  */
 #endregion
 
+using System.Linq;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
@@ -18,6 +19,9 @@ namespace OpenRA.Mods.Common.Traits
 		[FieldLoader.Require]
 		[Desc("Which voice set to use.")]
 		[VoiceSetReference] public readonly string VoiceSet = null;
+
+		[Desc("Which faction voice to use.")]
+		public readonly string[] Factions = null;
 
 		[Desc("Multiply volume with this factor.")]
 		public readonly float Volume = 1f;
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Voiced.cs
```diff
@@ -19,6 +20,9 @@ public class VoicedInfo : ITraitInfo
 		[Desc("Which voice set to use.")]
 		[VoiceSetReference] public readonly string VoiceSet = null;
 
+		[Desc("Which faction voice to use.")]
+		public readonly string[] Factions = null;
```
Phrohdoh: Arrays need to be empty, not null, by default.

pchote:
This conflicts with the plan to extract all the ugly mod-specific sound quirks into "sound sequences".  Are you able to live with an ra2-mod specific `RA2Voiced` trait subclass until this gets solved more generally upstream?

pchote:
You should be able to solve your engineer voices using our existing faction-specific voices support.  Look at how d2k does it.

Mailaender:
Been there, tried that. https://github.com/OpenRA/ra2/pull/49

GraionDilach:
:+1: for the `RA2Voiced` trait. This is still overlapping with the faction prefixes in the voice.yaml and such hackish approaches should be downstream.

abcdefg30:
Why is that hacky? To me this Looks valid and ok.

Iamgoofball:
so whats the replacement for this PR because we kind of need this for RA2

Mailaender:
We can either implement a `FactionVoiced` trait in the OpenRA.Mods.RA2.dll or go with something like https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Graphics/TilesetSpecificSpriteSequence.cs#L70 but for voices. For the RA2 soviet engineer we will just need to make sure that it retries another one if it can't find a variant.

PR #10692, opened 2016-02-05
Move CellCorners to MapGrid.

pchote:
We no longer need to store this per-map.  Moving it to MapGrid saves a small amount of memory and performance per map.


Mailaender:
:+1: You just need to fix the StyleCop error.

pchote:
Fixed.

penev92:
:+1:

PR #10693, opened 2016-02-05
Mines will no longer kill players own units.

estuans:
In current bleed, Mines will kill owner units of any kind - regardless of type.
I've added a simple or condition to the trigger that checks if the crushing unit's owner is the same as it's own and returns with no action if this passes.

Otherwise, blow up as normal :)

This can be repeated in current head by simply laying a field of mines in skirmish and making any unit cross them.

### DIFF
--- a/OpenRA.Mods.RA/Traits/Mine.cs
+++ b/OpenRA.Mods.RA/Traits/Mine.cs
@@ -38,7 +38,7 @@ namespace OpenRA.Mods.RA.Traits
 
 		public void OnCrush(Actor crusher)
 		{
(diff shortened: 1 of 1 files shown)


ScottNZ:
I think it would be better to just change AvoidFriendly to true (or remove the line so it uses the default value of true in the .cs file), here: https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/rules/defaults.yaml#L744

estuans:
Aww shucks. Well that makes more sense, and is simpler.

Do we want that behaviour to be configurable? I'm happy to update the YAML and push that instead.

GraionDilach:
Yes, we do.

estuans:
OK, closing pull request and will update YAML.

penev92:
Actually making this use Stances would be best because of consistency with other traits (see `WithDecoration` and `Infiltrates`).

PR #10695, opened 2016-02-05
Bug fix #10029 Plane resupply

### LINKED ISSUE #10029: Airplanes given "return to base" command do an unnecessary loop (if already flying towards base)

bateramos:
This bug was happening if the next move target was too close to the plane.
The fix is see if the move target is too close, if is, just don't create a Activity to move to this position.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
+++ b/OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
@@ -9,6 +9,7 @@
 #endregion
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Activities;
 using OpenRA.Mods.Common.Traits;
@@ -62,7 +63,7 @@ namespace OpenRA.Mods.Common.Activities
 
 			// Add 10% to the turning radius to ensure we have enough room
 			var speed = plane.MovementSpeed * 32 / 35;
-			var turnRadius = (int)(141 * speed / planeInfo.ROT / (float)Math.PI);
+			var turnRadius = CalculateTurnRadius(speed);
 
 			// Find the center of the turning circles for clockwise and counterclockwise turns
 			var angle = WAngle.FromFacing(plane.Facing);
@@ -111,12 +112,24 @@ namespace OpenRA.Mods.Common.Activities
 					return new FlyCircle(self);
 			}
 
-			return ActivityUtils.SequenceActivities(
-				new Fly(self, Target.FromPos(w1)),
-				new Fly(self, Target.FromPos(w2)),
-				new Fly(self, Target.FromPos(w3)),
-				new Land(self, Target.FromActor(dest)),
-				NextActivity);
+			List<Activity> landingProcedures = new List<Activity>();
+
+			var turnRadius = CalculateTurnRadius(planeInfo.Speed);
+
+			landingProcedures.Add(new Fly(self, Target.FromPos(w1), WDist.Zero, new WDist(turnRadius * 3)));
+			landingProcedures.Add(new Fly(self, Target.FromPos(w2)));
+
+			// Fix a problem when the airplane is send to resupply near the airport
+			landingProcedures.Add(new Fly(self, Target.FromPos(w3), WDist.Zero, new WDist(turnRadius / 2)));
+			landingProcedures.Add(new Land(self, Target.FromActor(dest)));
+			landingProcedures.Add(NextActivity);
+
+			return ActivityUtils.SequenceActivities(landingProcedures.ToArray());
+		}
+
+		int CalculateTurnRadius(int speed)
+		{
+			return (int)(141 * speed / planeInfo.ROT / (float)Math.PI);
 		}
 	}
 }


OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -110,13 +113,21 @@ public override Activity Tick(Actor self)
 				else
 					return new FlyCircle(self);
 			}
+			//// ignore destination generated by Calculate() if they are too close to self
+			List<Activity> landingProcedures = new List<Activity>();
+
+			var w1DistanceFromSelf = (w1 - self.CenterPosition).HorizontalLength;
+
+			if (w1DistanceFromSelf > TooCloseDestinationThreshold) {
+				landingProcedures.Add(new Fly(self, Target.FromPos(w1)));
```
abcdefg30: You should drop braces around single line statements.

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -110,13 +113,21 @@ public override Activity Tick(Actor self)
 				else
 					return new FlyCircle(self);
 			}
+			//// ignore destination generated by Calculate() if they are too close to self
```
abcdefg30: Please add a new line in front of the comment and start it with only two slashes and a capital:
`// Ignore destintation...`

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -18,6 +19,8 @@ namespace OpenRA.Mods.Common.Activities
 {
 	public class ReturnToBase : Activity
 	{
+		const int TooCloseDestinationThreshold = 5000;
```
abcdefg30: Couldn't this get a field in the `AircraftInfo` class?

bateramos: I think that this constant is to resolve this class specific problem. It do not affect other classes or is the main indicator that the destination is too close. On my opinion we can send it to the Info class when we know a real case for the parametrization of the number.

pchote: The required value is going to depend on the aircraft speed.  It should be calculated from that instead of being specified anywhere.

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -111,12 +112,23 @@ public override Activity Tick(Actor self)
...
+
+			// Fix a problem when the airplane is send to resupply near the airport
+			landingProcedures.Add(new Fly(self, Target.FromPos(w3), new WDist(0), new WDist((int)(turnRadius / 2))));
+			landingProcedures.Add(new Land(self, Target.FromActor(dest)));
+			landingProcedures.Add(NextActivity);
+
+			return ActivityUtils.SequenceActivities(landingProcedures.ToArray());
+		}
+
+		int CalculateTurnRadius(int speed) {
```
abcdefg30: Please put the brace on a new line:
```
int CalculateTurnRadius(int speed)
{
	return ...;
}

File: OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs [hunk 5]
```diff
@@ -111,12 +112,24 @@ public override Activity Tick(Actor self)
...
-				new Fly(self, Target.FromPos(w1)),
-				new Fly(self, Target.FromPos(w2)),
-				new Fly(self, Target.FromPos(w3)),
-				new Land(self, Target.FromActor(dest)),
-				NextActivity);
+			List<Activity> landingProcedures = new List<Activity>();
+
+			var turnRadius = CalculateTurnRadius(planeInfo.Speed);
+
+			landingProcedures.Add(new Fly(self, Target.FromPos(w1), new WDist(0), new WDist(turnRadius * 3)));
```
reaperrr: Nit: `WDist.Zero` instead of `new WDist(0)`.

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -111,12 +112,24 @@ public override Activity Tick(Actor self)
...
+			landingProcedures.Add(new Fly(self, Target.FromPos(w2)));
+
+			// Fix a problem when the airplane is send to resupply near the airport
```
reaperrr: Nit: `sent` instead of `send`.

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -111,12 +112,24 @@ public override Activity Tick(Actor self)
...
+			landingProcedures.Add(new Fly(self, Target.FromPos(w3), new WDist(0), new WDist((int)(turnRadius / 2))));
```
reaperrr: Nit: `WDist.Zero` instead of `new WDist(0)`.

Also, since `turnRadius` already is an `int`, this `(int)` should be unnecessary.

reaperrr:
Apart from the minor nits, this looks fine to me. :+1:

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/bf31caad8ad1958f1485a491988969392302be87...af5aa26ad0a70a4aafbdf84820e772e30f589dde)

PR #10696, opened 2016-02-05
Remove null checks against self

RoosterDragon:
Remove checks for self being null in Mobile.cs.

Picked up by a Coverity check which detected that self was dereferenced inside a call to `CanRemoveBlockage` without a null check - but the other code around it has such null checks.

In OpenRA, 'self' should never be null by convention (if it is - a NRE is probably the least of your problems).

### DIFF
--- a/OpenRA.Mods.Common/Traits/Mobile.cs
+++ b/OpenRA.Mods.Common/Traits/Mobile.cs
@@ -179,7 +179,6 @@ namespace OpenRA.Mods.Common.Traits
 		static bool IsMovingInMyDirection(Actor self, Actor other)
 		{
 			if (!other.IsMoving()) return false;
-			if (self == null) return true;
 
 			var selfMobile = self.TraitOrDefault<Mobile>();
 			if (selfMobile == null) return false;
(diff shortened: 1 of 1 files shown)


Mailaender:
:+1:

obrakmann:
> In OpenRA, 'self' should never be null by convention (if it is - a NRE is probably the least of your problems).

This is a static method in the Info class, though. `self` just might be null, you never know what's passed into it.

RoosterDragon:
If somebody passes null for self, an exception is better than hiding the problem.

obrakmann:
true dat. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/61b144005b4642a34879b0e1d3b72a5af8c06e42)

RoosterDragon:
This breaks [SpawnMPUnits](https://github.com/OpenRA/OpenRA/blob/870867760a15825f536a3cdd031589e0b61d2f91/OpenRA.Mods.Common/Traits/World/SpawnMPUnits.cs#L59) - but I think fixing the method signature would be better than just reverting this - so give me a bit.

PR #10697, opened 2016-02-05
Remove Map references from MapPreview.

### LINKED ISSUE #10433: MapPreview should not hold a live reference to Map
Keeping a copy of every map in memory is becoming a problem for dedicated servers that sync with the resource center: we are now looking at 600MB+ memory usage per server instance just for the maps.

Maps should be loaded on demand to get the info that is required, and then disposed.

> [04:40:57]	&lt;xnaquada&gt; The RA servers initial load is 340-380mb without any runs (e.g. noone has played on them since restart). After they get some plays, the ram can approach somewhere between 500-600mb
[04:41:18]	&lt;xnaquada&gt; I can actually tell what servers are what by ram usage
[04:41:51]	&lt;xnaquada&gt; RA =~350mb CNC ~160mb D2K ~33mb (all inital load after a fresh restart)
[04:42:14]	&lt;xnaquada&gt; kyrylo, what O/S?
[04:42:48]	&lt;kyrylo&gt; ubuntu
[04:43:41]	&lt;xnaquada&gt; (ram use is all per instance/openra server btw, just to be clear for clarity's sake).
[04:44:40]	&lt;xnaquada&gt; I've observed 500-600mb is where I end up eventually for the RA mod after some plays.
[04:47:04]	&lt;kyrylo&gt; i just have 1 server with 1 instance
[04:47:11]	&lt;kyrylo&gt; I end up with 850
[04:47:21]	&lt;kyrylo&gt; starting from 600
[04:47:31]	&lt;kyrylo&gt; so those 250 megs are never freed 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29630886-mappreview-should-not-hold-a-live-reference-to-map?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Closes #10433.

This removes the ram-hogging Map references from MapPreview.  The trade-off is that it takes (significantly) longer to generate the previews in the map chooser.  A future PR will reduce this delay by introducing a custom map parsing object that only loads the minimum amount of data required to generate a map preview.

This also removes the visible janks when selecting a map with custom rules in the lobby and missions in the mission browser.  It also gets us another step closer to oramod packages.

I have not tested that this actually reduces the memory usage of dedicated servers, but these changes will stand on their own merit even if that does not pan out (which would be very surprising).

### DIFF
--- a/OpenRA.Game/Map/MapCache.cs
+++ b/OpenRA.Game/Map/MapCache.cs
@@ -174,7 +174,8 @@ namespace OpenRA
 					if (bitmap == null)
 					{
 						createdPreview = true;
-						bitmap = Minimap.RenderMapPreview(modData.DefaultRules.TileSets[p.Map.Tileset], p.Map, modData.DefaultRules, true);
+						var map = new Map(p.Path);
+						bitmap = Minimap.RenderMapPreview(modData.DefaultRules.TileSets[map.Tileset], map, modData.DefaultRules, true);
 					}
 
 					Game.RunAfterTick(() =>
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
@@ -26,7 +26,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 	{
 		static readonly Action DoNothing = () => { };
 
-		public MapPreview Map = MapCache.UnknownMap;
+		public MapPreview MapPreview = MapCache.UnknownMap;
 
 		readonly Action onStart;
 		readonly Action onExit;
@@ -168,7 +168,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 					var onSelect = new Action<string>(uid =>
 					{
 						// Don't select the same map again
-						if (uid == Map.Uid)
+						if (uid == MapPreview.Uid)
 							return;
 
 						orderManager.IssueOrder(Order.Command("map " + uid));
@@ -178,7 +178,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 					Ui.OpenWindow("MAPCHOOSER_PANEL", new WidgetArgs()
 					{
-						{ "initialMap", Map.Uid },
+						{ "initialMap", MapPreview.Uid },
 						{ "initialTab", MapClassification.System },
 						{ "onExit", DoNothing },
 						{ "onSelect", Game.IsHost ? onSelect : null },
@@ -190,7 +190,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			var slotsButton = lobby.GetOrNull<DropDownButtonWidget>("SLOTS_DROPDOWNBUTTON");
 			if (slotsButton != null)
 			{
-				slotsButton.IsDisabled = () => configurationDisabled() || panel != PanelType.Players ||	Map.RuleStatus != MapRuleStatus.Cached ||
+				slotsButton.IsDisabled = () => configurationDisabled() || panel != PanelType.Players ||	MapPreview.RuleStatus != MapRuleStatus.Cached ||
 					(orderManager.LobbyInfo.Slots.Values.All(s => !s.AllowBots) &&
 					orderManager.LobbyInfo.Slots.Count(s => !s.Value.LockTeam && orderManager.LobbyInfo.ClientInSlot(s.Key) != null) == 0);
 
@@ -294,7 +294,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 			var optionsTab = lobby.Get<ButtonWidget>("OPTIONS_TAB");
 			optionsTab.IsHighlighted = () => panel == PanelType.Options;
-			optionsTab.IsDisabled = () => Map.RuleStatus != MapRuleStatus.Cached || panel == PanelType.Kick || panel == PanelType.ForceStart;
+			optionsTab.IsDisabled = () => MapPreview.RuleStatus != MapRuleStatus.Cached || panel == PanelType.Kick || panel == PanelType.ForceStart;
 			optionsTab.OnClick = () => panel = PanelType.Options;
 
 			var playersTab = lobby.Get<ButtonWidget>("PLAYERS_TAB");
@@ -317,9 +317,10 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			var startGameButton = lobby.GetOrNull<ButtonWidget>("START_GAME_BUTTON");
 			if (startGameButton != null)
 			{
-				startGameButton.IsDisabled = () => configurationDisabled() || Map.RuleStatus != MapRuleStatus.Cached ||
+				startGameButton.IsDisabled = () => configurationDisabled() || MapPreview.RuleStatus != MapRuleStatus.Cached ||
 					orderManager.LobbyInfo.Slots.Any(sl => sl.Value.Required && orderManager.LobbyInfo.ClientInSlot(sl.Key) == null) ||
 					(orderManager.LobbyInfo.GlobalSettings.DisableSingleplayer && orderManager.LobbyInfo.IsSinglePlayer);
+
 				startGameButton.OnClick = () =>
 				{
 					// Bots and admins don't count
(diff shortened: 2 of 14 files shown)


pchote:
Confirmed that this reduces the memory usage by ~700MB after loading the RA mod (and generating the map preview) with the resource center maps synced.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -26,7 +26,8 @@ public class LobbyLogic : ChromeLogic
 	{
 		static readonly Action DoNothing = () => { };
 
-		public MapPreview Map = MapCache.UnknownMap;
+		public MapPreview MapPreview = MapCache.UnknownMap;
+		public Map Map = null;
```
RoosterDragon: You're missing null checks win a few places when accessing this field in `LobbyMapPreviewLogic.cs` and `LobbyLogic.cs` that cause crashes if the preview is slow to generate. This also isn't assigned in a thread-safe manner from the worker thread.

OpenRA.Game/Map/MapCache.cs
```diff
@@ -174,7 +174,8 @@ void LoadAsyncInternal()
 					if (bitmap == null)
 					{
 						createdPreview = true;
-						bitmap = Minimap.RenderMapPreview(modData.DefaultRules.TileSets[p.Map.Tileset], p.Map, modData.DefaultRules, true);
+						var map = new Map(p.Path);
```
RoosterDragon: This is super not thread-safe. There's way too much shared state being shared here.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -741,31 +742,33 @@ bool SwitchTeamChat()
...
+			MapPreview = Game.ModData.MapCache[uid];
+			if (MapPreview.Status == MapStatus.Available)
 			{
 				// Maps need to be validated and pre-loaded before they can be accessed
 				new Thread(_ =>
 				{
-					var map = Map;
-					map.CacheRules();
+					Map = new Map(MapPreview.Path);
+					Map.PreloadRules();
```
RoosterDragon: If this thread and the map caching thread happen to be loading the same map at the same time this will likely explode - they'll be seeking around in the same streams when loading the files from the mod container and messing with each other's offsets - super bad times.

OpenRA.Game/Map/MapCache.cs
RoosterDragon: You're generally only getting away with this because this thread loads maps one at a time. If it attempted to parallelize the loading you'd probably get way more random crashes.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
pchote: Are you sure about this? It shouldn't be any different to what we were already doing - i've just moved the reference from inside the MapPreview to here.

OpenRA.Game/Map/MapCache.cs
pchote: This is going to be replaced with a much lighterweight parser in a future pr.  Is this bogus enough to have to expand the scope of the pr with these extra changes now?

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
pchote: How would they be sharing streams?

pchote: Also note that the existing code was already doing this (CacheRules was a wrapper for PreloadRules).

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
RoosterDragon: Yes - you can more easily repro these crashes by simply returning from the worker threads without actually generating the previews.

Here's a few I can see offhand:
[LobbyMapPreviewLogic.cs#L28](https://github.com/pchote/OpenRA/blob/557fcbdedd012741498e10f910f3612d23f372fa/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyMapPreviewLogic.cs#L28)
[LobbyMapPreviewLogic.cs#L59](https://github.com/pchote/OpenRA/blob/557fcbdedd012741498e10f910f3612d23f372fa/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyMapPreviewLogic.cs#L59)
[LobbyLogic.cs#L193](https://github.com/pchote/OpenRA/blob/557fcbdedd012741498e10f910f3612d23f372fa/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs#L193)
[LobbyLogic.cs#L320](https://github.com/pchote/OpenRA/blob/557fcbdedd012741498e10f910f3612d23f372fa/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs#L320)

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
RoosterDragon: `Container.GetStream("map.yaml")` in the map ctor which will call into a lower level `IReadOnlyPackage.GetStream` call. Some implementations have a shared stream to resource (e.g. the file) and copy the portion they need to a `MemoryStream` before returning that. This breaks down in a multi-threaded scenario.

pchote: The SegmentStream is only used for mix files.  Maps are only supported from zip files and directories.

OpenRA.Game/Map/MapCache.cs
RoosterDragon: The Map constructor simply isn't thread-safe - so this change is bogus by default unless you fix that.

The fact this thread loads one at a time mostly saves you - but I can contrive some scenarios where it still explodes - such as switching mods whilst this thread is still generating previews and causing the global `Game.ModData` to change.

You could so something terribly hacky be reworking the Map constructor to do something like:
```
Map(string path)
{
    lock(StaticSyncObjectWhichIsLameButHeyItsBetterThanRandomThreadBugs)
    {
        // Load here
    }
}

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
RoosterDragon: That's a lot of relying on implementation details in those methods.

RoosterDragon:
Although the thread-safety worries me in general - if the follow up PR comes along fairly promptly and fixes things by introducing a new parser that doesn't have these issues - then I'm otherwise happy to take this.

pchote:
Fixed the LobbyLogic thread safety issues (you can verify this by adding a `Thread.Sleep` at the start of the thread body at ~L754).  As discussed on IRC the other thread safety issues already exist in the current code.

OpenRA.Game/Map/MapCache.cs
pchote: The main issue is reading from the filesystem, so a better approach may be to:

1.  Use locks around the innards of the various Filesystem code to avoid race conditions between e.g. mounting packages and querying contents
2.  Make sure that the stream objects returned by the package implementations do not implicitly share state.

abcdefg30:
Needs a rebase.

penev92:
:+1:

pchote:
Rebased.

obrakmann:
Looks good to me. Memory usage for me on the ra main menu with disabled shellmap  is down about 20 MB compared to bleed (and about 70/80 MB compared to release). CPU usage compared to release is down by about 10%, btw. Nice. :+1:

obrakmann:
Added  the [Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/79158c1bcb801f2590d1193e6dc58a11a6f46f1c) entry to the rest of the filesystem stuff.

PR #10701, opened 2016-02-06
Remove automatic map format upgrades and add versioned user map dirs

pchote:
This removes the "feature" where we would automatically update maps to the latest format when they are loaded, which has caused problems with the resource center and people switching between releases and playtests.  It also gets us a step closer to immutable maps, which will open up some new perf improvements.  Map format upgrades are now done when running the mod/map upgrade rules.

To avoid the game trying and failing to load a bunch of outdated versions on startup I have added per-version map folders.  If somebody wants to copy their self-made map to a newer version of the game they will need to either (a) manually copy it to the new version directory, and then possibly run the upgrader, or (b) let the resource center automatically upgrade it, and then redownload it when joining a server on the new version.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckMapMetadata.cs
+++ b/OpenRA.Mods.Common/Lint/CheckMapMetadata.cs
@@ -17,9 +17,9 @@ namespace OpenRA.Mods.Common.Lint
 	{
 		public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 		{
-			if (map.MapFormat < Map.MinimumSupportedMapFormat)
-				emitError("Map format {0} is older than the minimum supported version {1}."
-					.F(map.MapFormat, Map.MinimumSupportedMapFormat));
+			if (map.MapFormat != Map.SupportedMapFormat)
+				emitError("Map format {0} does not match the supported version {1}."
+					.F(map.MapFormat, Map.SupportedMapFormat));
 
 			if (map.Author == null)
 				emitError("Map does not define a valid author.");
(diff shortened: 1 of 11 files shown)


pchote:
Note: this will break existing scripts to rsync dedicated servers with the resource center.

pchote:
Rebased.

obrakmann:
Looks good to me. :+1:

OpenRA.Mods.Common/Lint/CheckMapMetadata.cs
```diff
@@ -17,9 +17,9 @@ public class CheckMapMetadata : ILintMapPass
 	{
 		public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 		{
-			if (map.MapFormat < Map.MinimumSupportedMapFormat)
-				emitError("Map format {0} is older than the minimum supported version {1}."
-					.F(map.MapFormat, Map.MinimumSupportedMapFormat));
+			if (map.MapFormat != Map.SupportedMapFormat)
+				emitError("Map format {0} not the supported version {1}."
```
Mailaender: Missing  a verb: "_is_ not the supported version"

pchote:
Fixed and rebased.

Mailaender:
Really good idea that will fix a lot of annoyance from both developers and players. :+1: Works as promised. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/b9a1bcd38e77afcc5a6952fe0b59ea1293f9b886...4081a6ddbcc57a96e48baf7fb2528fcab9c06ddc)

PR #10703, opened 2016-02-06
Unify filesystem package syntax.

pchote:
This unifies our handling of mod/package specific files with common code in the virtual file system.

There are three new/changed syntax elements for use in yaml:
1. `explicit-name|file` specifies that `file` should be loaded from the package that has been explicitly named `explicit-name`.
2. `package: explicit-name` in the `Packages:` section of `mod.yaml` defines an explicitly named package that can be referenced using the `|` syntax.
3. `$mod` in the `Packages:` section of `mod.yaml` defines that the mod named `mod` should be mounted as a package.  Mods may be loaded from either the system or support dirs, and will eventually support both unpacked and oramods.

As a compromise based on the earlier discussions/arguments I have changed the explicit package operator from `:` to `|` and reverted the change for `MapFolders`.

This has the side-effect of making all our default mods portable between the system and support directories.

Supersedes #10487.

### DIFF
--- a/OpenRA.Game/FileSystem/FileSystem.cs
+++ b/OpenRA.Game/FileSystem/FileSystem.cs
@@ -21,6 +21,8 @@ namespace OpenRA.FileSystem
 	{
 		public IEnumerable<IReadOnlyPackage> MountedPackages { get { return mountedPackages.Keys; } }
 		readonly Dictionary<IReadOnlyPackage, int> mountedPackages = new Dictionary<IReadOnlyPackage, int>();
+		readonly Dictionary<string, IReadOnlyPackage> explicitMounts = new Dictionary<string, IReadOnlyPackage>();
+
 		Cache<string, List<IReadOnlyPackage>> fileIndex = new Cache<string, List<IReadOnlyPackage>>(_ => new List<IReadOnlyPackage>());
 
 		public IReadWritePackage CreatePackage(string filename, Dictionary<string, byte[]> content)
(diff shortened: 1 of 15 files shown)


pchote:
Dependencies have been merged.

OpenRA.Game/FileSystem/FileSystem.cs
```diff
@@ -164,58 +182,67 @@ public Stream Open(string filename)
...
-			// Used for faction specific packages; rule out false positive on Windows C:\ drive notation
-			var explicitPackage = name.Contains(':') && !Directory.Exists(Path.GetDirectoryName(name));
-			if (explicitPackage)
+			var explicitSplit = path.IndexOf('|');
+			if (explicitSplit > 0)
 			{
-				var divide = name.Split(':');
-				packageName = divide.First();
-				filename = divide.Last();
+				if (explicitMounts.TryGetValue(path.Substring(0, explicitSplit), out package))
```
abcdefg30: You could combine those two checks: `if (explicitSplit > 0 && explicitMounts...)`

pchote:
Fixed and rebased.

obrakmann:
The map editor saves in a strange place when you tell it to save in the mod dir: 
```
Saved current map at Support/maps/cnc/test1.oramap
Saved current map at cnc|maps/test2.oramap
```

abcdefg30:
I can't even access/see the maps in my support directory.
This looks good to me otherwise. :+1: / :white_check_mark:

pchote:
Fixed.

penev92:
:+1:

pchote:
Fixed the support dir map enumeration.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a5f910dab12d63d2c18d0b4d7c91ce055c0bd549...0cb1071567a9f240829e2fed666adda665806b4a)

penev92:
This probably needs another changelog entry aside from the generic one to note that the syntax used has changed?

PR #10707, opened 2016-02-06
Fix mobile checks

RoosterDragon:
Follow up to #10696 to allow us to do mobility checks without a real actor. SpawnMPUnits needs this so we do need the check after all (otherwise the game crashes) - this is the cleanest way I can see to do it.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Mobile.cs
+++ b/OpenRA.Mods.Common/Traits/Mobile.cs
@@ -232,6 +232,12 @@ namespace OpenRA.Mods.Common.Traits
 			if (otherActor == ignoreActor)
 				return false;
 
+			// If self is null, we don't have a real actor - we're just checking what would happen theoretically.
+			// In such a scenario - we'll just assume any other actor in the cell will block us by default.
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -232,6 +232,12 @@ bool IsBlockedBy(Actor self, Actor otherActor, Actor ignoreActor, CellConditions
 			if (otherActor == ignoreActor)
 				return false;
 
+			// If self is null, we don't have a real actor - we're just checking what would happen theoretically.
+			// In such a scenario - we'll just assume any other actor in the cell will block us by default.
```
pchote: What about infantry and other sub-cell stuff?

RoosterDragon: See `CanMoveFreelyInto` - if there's a free sub cell you're fine. Only after that do blockers get checked.

pchote: Ok, cool.

chrisforbes:
:+1:

obrakmann:
:+1:

PR #10706, opened 2016-02-06
Loads mod assemblies via the virtual filesystem.

pchote:
Prerequisite to #10703 and oramods.

Unfortunately this delays the time to first loadscreen render, but I don't see any alternatives.

PR #10704, opened 2016-02-06
Loads fonts via the virtual filesystem.

pchote:
Prerequisite to #10703 and oramods.

PR #10713, opened 2016-02-07
Unify static mod metadata collections.

pchote:
This merges the static Manifest collection into the static ModMetadata collection and replaces the path reference with a Package reference.

This is one of the last pieces of plumbing for supporting oramod packages: once the rest of the engine is using the virtual filesystem properly (#10704, #10706, #10703) it's as simple as adding an `else` case to `ModMetadata.ValidateMods` to return the oramod container instead of a folder.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
@@ -82,17 +82,19 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			{
 				try
 				{
-					using (var preview = new Bitmap(Platform.ResolvePath(ModMetadata.CandidateModPaths[mod.Id], "preview.png")))
-						if (preview.Width == 296 && preview.Height == 196)
-							previews.Add(mod.Id, sheetBuilder.Add(preview));
+					using (var stream = ModMetadata.AllMods[mod.Id].Package.GetStream("preview.png"))
+						using (var preview = new Bitmap(stream))
+							if (preview.Width == 296 && preview.Height == 196)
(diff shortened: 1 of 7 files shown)


OpenRA.Mods.Common/Widgets/Logic/ModBrowserLogic.cs
```diff
@@ -82,17 +82,19 @@ public ModBrowserLogic(Widget widget)
 			{
 				try
 				{
-					using (var preview = new Bitmap(Platform.ResolvePath(ModMetadata.CandidateModPaths[mod.Id], "preview.png")))
-						if (preview.Width == 296 && preview.Height == 196)
-							previews.Add(mod.Id, sheetBuilder.Add(preview));
+					using (var stream = ModMetadata.AllMods[mod.Id].Package.GetStream("preview.png"))
```
pchote: Note that this will temporarily break the ability to view these files for mods in the support dir, but, like the other cases, this will be restored in #10703.

penev92: Actually it doesn't change anything in that regard. The mod runs (this can't break it even if it wanted to), and the mod's logo and preview images are still loaded (test case - d2k, as I already set it up to be ran from the Support dir on bleed).
So we're good here.

penev92:
:+1:

obrakmann:
Looks ok, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/cc179ce237eb6cd79446afc5807e38fdb453bd29)

PR #10714, opened 2016-02-07
Added basic RA2/TS mod rules importers

Mailaender:
Useful when setting up a large amount of decorational props.

### DIFF
--- a/OpenRA.Mods.TS/UtilityCommands/LegacyRulesImporter.cs
+++ b/OpenRA.Mods.TS/UtilityCommands/LegacyRulesImporter.cs
@@ -47,10 +47,10 @@ namespace OpenRA.Mods.TS.UtilityCommands
 			var terrainObjects = rulesIni.GetSection("TerrainTypes").Select(b => b.Value).Distinct();
 			Console.WriteLine("# Terrain Objects");
 			Console.WriteLine();
-			ImportStructures(terrainObjects);
+			ImportStructures(terrainObjects, true);
 		}
 
-		void ImportStructures(IEnumerable<string> structures)
+		void ImportStructures(IEnumerable<string> structures, bool useTerrainPalette = false)
 		{
 			foreach (var building in structures)
 			{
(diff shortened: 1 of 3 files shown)


abcdefg30:
Looks good to me. :+1:

OpenRA.Mods.TS/UtilityCommands/LegacyRulesImporter.cs
```diff
@@ -0,0 +1,216 @@
...
+							Console.Write("\t\tFootprint:");
+							var width = 0;
+							int.TryParse(dimensions.First(), out width);
+							var height = 0;
+							int.TryParse(dimensions.Last(), out height);
+							for (var y = 0; y < height; y++)
+							{
+								Console.Write(" ");
+								for (var x = 0; x < width; x++)
+									Console.Write("x");
```
GraionDilach: In TS/RA2 the whole last row was the bib if the building had the `Bib=yes` tag. Worths a try adding that as part of this PR.

OpenRA.Mods.TS/UtilityCommands/LegacyRulesImporter.cs
```diff
@@ -0,0 +1,216 @@
...
+					var buildup = artSection.GetValue("Buildup", string.Empty);
+					if (!string.IsNullOrEmpty(buildup) && buildup != "none")
+						Console.WriteLine("\tWithMakeAnimation:");
+
+					var terrainPalette = artSection.GetValue("TerrainPalette", string.Empty);
+					if (!string.IsNullOrEmpty(terrainPalette))
+						bool.TryParse(terrainPalette, out useTerrainPalette);
+
+					var remapable = artSection.GetValue("Remapable", string.Empty);
+					if (!string.IsNullOrEmpty(remapable) && remapable == "yes")
```
GraionDilach: `Remapable` doesn't even work from what I know, the buildings always ended up being remapped. The tag doesn't do anything.

GraionDilach: Assuming it would still work, this would still mean `Palette: player` and not terrain though, but the terrain palette overrode the player palette, in which case remapping became automatically disabled.

OpenRA.Mods.TS/UtilityCommands/LegacyRulesImporter.cs
```diff
@@ -0,0 +1,216 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
```
obrakmann: I can't comment on the functionality, but you should use the updated copyright header here. It's also 2016 :)

pchote:
Any progress here?

OpenRA.Mods.TS/UtilityCommands/LegacyRulesImporter.cs
Mailaender: I came up with this conversion rule when automatically setting up the map props for DTA at http://www.ppmforums.com/viewtopic.php?t=40639 where this worked well as an exception. I don't need to explicitly set the player palette as that is already the default: https://github.com/OpenRA/OpenRA/wiki/Traits#rendersprites

OpenRA.Mods.TS/UtilityCommands/LegacyRulesImporter.cs
Mailaender: Do I need to add an extra line or just convert the last line to `_`? I haven't tested this with anything other than civilian buildings or trees to be fair.

Mailaender:
Added the 2016 copyright header.

OpenRA.Mods.TS/UtilityCommands/LegacyRulesImporter.cs
GraionDilach: Alright.

OpenRA.Mods.TS/UtilityCommands/LegacyRulesImporter.cs
GraionDilach: Convert the last one.

reaperrr:
Needs a rebase again.

Mailaender:
Rebased.

reaperrr:
Would be nice if this gained the ability to import weapons, vehicles and infantry later on, but I think this already saves a lot of work and is a good starting point, so :+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/98266b431efd26656591760a57ae5af40fa98cfd...2eb0276da0bf1c49f3f1c9fd78c424787b9f9c2d)

PR #10715, opened 2016-02-07
Added a TS/RA2 sprite sequence importer for infantry and buildings

Mailaender:
These are traditionally time consuming to do manually.

### DIFF
--- a/OpenRA.Mods.TS/UtilityCommands/LegacySequenceImporter.cs
+++ b/OpenRA.Mods.TS/UtilityCommands/LegacySequenceImporter.cs
@@ -1,0 +1,259 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using OpenRA.Mods.Common.FileFormats;
+
+namespace OpenRA.Mods.TS.UtilityCommands
+{
+	class ImportLegacySequenceCommand : IUtilityCommand
+	{
+		public bool ValidateArguments(string[] args)
+		{
+			return args.Length >= 2;
+		}
+
+		public string Name { get { return "--sequence-import"; } }
+
+		IniFile file;
+
+		[Desc("FILENAME", "Convert ART.INI to the OpenRA sequence definition format.")]
+		public void Run(ModData modData, string[] args)
+		{
+			// HACK: The engine code assumes that Game.modData is set.
+			Game.ModData = modData;
+
+			file = new IniFile(File.Open(args[1], FileMode.Open));
+
+			foreach (var section in file.Sections)
+			{
+				var sequence = section.GetValue("Sequence", string.Empty);
+				if (!string.IsNullOrEmpty(sequence))
+					ImportInfantrySequences(section, sequence);
+
+				var foundation = section.GetValue("Foundation", string.Empty);
+				if (!string.IsNullOrEmpty(foundation))
+					ImportBuildingSequences(section);
+			}
+		}
+
+		void ImportBuildingSequences(IniSection section)
+		{
+			Console.WriteLine(section.Name + ":");
+
+			Console.WriteLine("\tDefaults:");
+
+			var theater = section.GetValue("Theater", string.Empty);
+			if (!string.IsNullOrEmpty(theater) && theater == "yes")
+				Console.WriteLine("\t\tUseTilesetExtension: true");
+			else
+			{
+				var toOverlay = section.GetValue("ToOverlay", string.Empty);
+				if (!string.IsNullOrEmpty(toOverlay))
+				{
+					var overlaySection = file.GetSection(toOverlay);
+					var overlayTheater = overlaySection.GetValue("Theater", string.Empty);
+					if (!string.IsNullOrEmpty(overlayTheater) && overlayTheater == "yes")
+						Console.WriteLine("\t\tUseTilesetExtension: true");
+				}
+			}
+
+			var newTheater = section.GetValue("NewTheater", string.Empty);
+			if (!string.IsNullOrEmpty(newTheater) && newTheater == "yes")
+				Console.WriteLine("\t\tUseTilesetCode: true");
+
+			var cameo = section.GetValue("Cameo", string.Empty);
+			if (!string.IsNullOrEmpty(cameo))
+			{
+				Console.WriteLine("\ticon: " + cameo.ToLowerInvariant());
+				Console.WriteLine("\t\tOffset: 0,0");
+				Console.WriteLine("\t\tUseTilesetExtension: false");
+				Console.WriteLine("\t\tUseTilesetCode: false");
+			}
+
+			Console.WriteLine("\tidle: ");
+
+			var buildup = section.GetValue("Buildup", string.Empty);
+			if (!string.IsNullOrEmpty(buildup) && buildup != "none")
+			{
+				Console.WriteLine("\tmake: " + buildup.ToLowerInvariant());
+			}
+
+			var bibShape = section.GetValue("BibShape", string.Empty);
+			if (!string.IsNullOrEmpty(bibShape))
+			{
+				Console.WriteLine("\tbib: " + bibShape.ToLowerInvariant());
+				Console.WriteLine("\t\tLength: 1");
+				Console.WriteLine("\t\tZOffset: -1024");
+			}
+
+			Console.WriteLine();
+		}
+
+		void ImportInfantrySequences(IniSection section, string sequence)
+		{
+			Console.WriteLine(section.Name + ":");
+
+			var image = section.GetValue("Image", string.Empty);
+			if (!string.IsNullOrEmpty(image))
+			{
+				Console.WriteLine("\tDefaults: " + image.ToLowerInvariant());
+				Console.WriteLine("\t\tTick: 80");
+			}
+
+			var cameo = section.GetValue("Cameo", string.Empty);
+			if (!string.IsNullOrEmpty(cameo))
+			{
+				Console.WriteLine("\ticon: " + cameo.ToLowerInvariant());
+				Console.WriteLine("\t\tOffset: 0,0");
+				Console.WriteLine("\t\tUseTilesetExtension: false");
+				Console.WriteLine("\t\tUseTilesetCode: false");
+			}
+
+			if (file.Sections.Any(s => s.Name == sequence.ToLowerInvariant()))
+			{
+				var sequenceSection = file.GetSection(sequence);
+				var guard = sequenceSection.GetValue("Guard", string.Empty);
+				if (!string.IsNullOrEmpty(guard))
+				{
+					Console.WriteLine("\tstand:");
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.TS/UtilityCommands/LegacySequenceImporter.cs
```diff
@@ -0,0 +1,197 @@
...
+				if (!string.IsNullOrEmpty(foundation))
+					ImportBuildingSequences(section);
+			}
+		}
+
+		void ImportBuildingSequences(IniSection section)
+		{
+			Console.WriteLine(section.Name + ":");
+
+			Console.WriteLine("\tDefaults:");
```
abcdefg30: Won't this create an "empty" `Defaults:` "section" in the yaml, if the imported structure doesn't use tileset extensions?

But anyway, I guess even if, this shouldn't hurt.

abcdefg30:
Hm, seems like this is generation `Facings: 0` values, which is not correct (e.g. import the sequences of the rocketeer and look at its idle definitions) and on other sequences it used `Facings: 6` where it should be 8 (probably not your fault, as the value is really 6, but the assets browser shows me 8 facings...).

GJ otherwise.

Mailaender:
It indeed has some quirks. For example the `stand` sequence always needs to be corrected to:

```diff
	stand:
-		Start: 0
-		Length: 1
-		Facings: 1
	stand:
+		Start: 0
+		Length: 6
+		Facings: 8
```

Those definitions seem to have different meanings for each type although the enumeration looks identical.

OpenRA.Mods.TS/UtilityCommands/LegacySequenceImporter.cs
Mailaender: No tileset codes/extensions is quite the exception for TS/RA2 mods. These auto-generated rules need a cleanup afterwards anyway.

OpenRA.Mods.TS/UtilityCommands/LegacySequenceImporter.cs
```diff
@@ -0,0 +1,197 @@
...
+		}
+
+		void ConvertStartLengthFacings(string input)
+		{
+			var splitting = input.Split(',');
+			if (splitting.Length >= 3)
+			{
+				Console.WriteLine("\t\tStart: " + splitting[0]);
+				Console.WriteLine("\t\tLength: " + splitting[1]);
+				Console.WriteLine("\t\tFacings: " + splitting[2]);
```
GraionDilach: `splitting.Length == 4 ? Console.WriteLine("\t\tFacings: 1") : Console.WriteLine("\t\tFacings: 8")` please.

If I get it right, `splitting[2] - splitting[1]` would be OpenRA `Stride:` but I never seen infantry which either used this logic either used the `Stride` tag in OpenRA.

GraionDilach:
That 0 1 1 stand looks legit though, TS/RA2 couldn't use more frames for idling at all, jumpjet infantry are a giant hack itself, your 0 6 8 fix sounds wrong to me instead tbh. `Length` shouldn't be increased.

Mailaender:
Thanks, added your suggested fixes and some more infantry sequences such as die, panic and cheer.

abcdefg30:
:+1: Looks good to me now.

OpenRA.Mods.TS/UtilityCommands/LegacySequenceImporter.cs
```diff
@@ -0,0 +1,253 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
```
obrakmann: I can't comment on the functionality, but you should use the updated copyright header here. It's also 2016 :)

pchote:
Any progress here?

GraionDilach:
Technically this doesn't need a progress (excluding the header). I think this is done but haven't found time to check the output to give my thumb.

pchote:
The header is what i'm referring to.

Mailaender:
Updated with the 2016 copyright header.

GraionDilach:
You missed `LegacySequenceImporter`.

Mailaender:
Fixed.

pchote:
You have access to the footprint information, so really should generate the appropriate artwork offsets for buildings (or at least something that is approximately correct).  Ignoring them here and then doing them all by hand (https://github.com/OpenRA/ra2/issues/159) is a giant waste of effort and potential.

Mailaender:
Really, I thought that information just wasn't available and all the TS offsets were corrected manually?

pchote:
It's not available to the OpenRA sprite sequence parsers, because it depends on the building footprints in the rules.  You have access to the building foundation here.

GraionDilach:
For reference:

In TS/RA2 the center of the SHP was the building's top-right (in isometric perspective, the top corner) cells top pixel. OpenRA assumes the center of the SHP is the building's middle.

Besides the footprints, it is based on the actual cell's size, and the offset math is also quite complex at non-square footprints (y offset will always be cellheight*y/2 but the x I'm not certain about, at square footprints, it's 0, but notsure at the rest, might be cellwidth*(x-y)/2 but I'm just guessing).

Including this here might be more complicated than the rest which are universal and shared between TS and RA2 which is why I felt that's being out-of-scope here.

GraionDilach:
Attaching a picture from SHP Builder with TS cellsize to point this out further - although it's known SHP Builder mispositions this grid some pixels away, probably due to rounding errors.

shpbuildertsgrid

reaperrr:
Adding support for automatic building offset generation would be nice.
Needs a rebase in any case.

Mailaender:
Rebased and added the offset guessing, which is indeed an excellent idea and your suggestions work rather good in practice https://github.com/OpenRA/ra2/issues/159 as most houses have a square footprint.

pchote:
The offsets are imported incorrectly for all the rectangular footprints that I checked in TS.

Mailaender:
I don't see any pattern to be honest:

Dimensions: 4,3

```patch
gaweap:
	Defaults:
-		Offset: 0,-36
+		Offset: -12, -42
```

Dimensions: 3,2

```patch
nahand:
	Defaults:
-		Offset: 0,-24
+		Offset: -6, -30
```
	
Dimensions: 2,3
	
```patch
naapwr:
	Defaults:
-		Offset: -24,-36
+		Offset: 12, -30
```

pchote:
Your best test cases will be the TS gates, which are currently imported with an offset that is wrong by multiple cells.

Mailaender:
Dimensions: 3,1

```patch
gagate_a:
	Defaults:
-		Offset: 48,-12
+		Offset: -24, -24
```

Dimensions: 1,3

```patch
gagate_b:
	Defaults:
-		Offset: -48,-36
+		Offset: 24, -24
```

Mailaender:
Removed https://github.com/OpenRA/OpenRA/pull/10715/commits/718107bbc50003398141957655ed6bb1f4e45641, because while this is a good idea, it is a large enough feature for its own pull request and this one already in development for 4 months.

pchote:
From above:
> You have access to the footprint information, so really should generate the appropriate artwork offsets > for buildings (or at least something that is approximately correct). Ignoring them here and then doing
> them all by hand (OpenRA/ra2#159) is a giant waste of effort and potential for this tool.

I don't think it is fair to ask us to merge an unfinished tool upstream just because you don't want to work through the equations to finish it.  That *forces* the burden of effort onto someone else, compared with *asking* for help and receiving it on their timescale.

Mailaender:
In fact I removed the commit to not submit unfinished code, but only proven one.

pchote:
Do you have a branch or commit sha1 for the version with the incorrect offset calculation?

pchote:
`FootprintUtils.CenterOffset` gives you the offset between the center of the top-left cell and the center of the actor.  Just add 512 to Y and then multiply by the tile pixel size / 1024.

Mailaender:
Updated with an improved version of the building offset converter thanks to your help. I just changed the sign and divided the X offset by 2. Updated https://github.com/OpenRA/ra2/pull/224 where the map "Powder Keg" is a good test case, because it features a lot of buildings that are aligned to the street tiles.

OpenRA.Mods.TS/UtilityCommands/LegacySequenceImporter.cs
```diff
@@ -0,0 +1,275 @@
...
+		}
+
+		void ImportBuildingSequences(IniSection section)
+		{
+			Console.WriteLine(section.Name + ":");
+
+			Console.WriteLine("\tDefaults:");
+
+			var foundation = section.GetValue("Foundation", string.Empty);
+			if (!string.IsNullOrEmpty(foundation))
```
pchote: You can simplify this much more:
```csharp
			if (!string.IsNullOrEmpty(foundation))
			{
				var size = foundation.Split('x');
				if (size.Length == 2)
				{
					var x = int.Parse(size[0]);
					var y = int.Parse(size[1]);
	
					var xOffset = (x - y) * grid.TileSize.Width / 4;
					var yOffset = (x + y) * grid.TileSize.Height / 4;
					Console.WriteLine("\t\tOffset: {0},{1}", -xOffset, -yOffset);
				}
			}
```

OpenRA.Mods.TS/UtilityCommands/LegacySequenceImporter.cs
```diff
@@ -0,0 +1,275 @@
...
+
+		void ConvertStartLengthFacings(string input)
+		{
+			var splitting = input.Split(',');
+			if (splitting.Length >= 3)
+			{
+				Console.WriteLine("\t\tStart: " + splitting[0]);
+				Console.WriteLine("\t\tLength: " + splitting[1]);
+
+				// TODO splitting[2] - splitting[1] may be Stride:
```
pchote: Do you have an example of where splitting[2] is used?

GraionDilach: https://github.com/OpenRA/OpenRA/pull/10715#discussion_r53581402

GraionDilach: Oh, apparently there are cases in RA2 where this is used (crawl's first frames are reused as prone in RA2). https://github.com/Phrohdoh/oramod-ra2/blob/master/extra-files/art.ini#L9637

pchote: Ok, then we should resolve this easy TODO before merging.

Mailaender: I have no idea if I understand what this strides is, but I added it.

OpenRA.Mods.TS/UtilityCommands/LegacySequenceImporter.cs
Mailaender: Added. This won't hide as you commented on a line that isn't affected by your suggestions.

pchote:
This now spits out `Stride: 0` for everything that doesn't need it, and all the non-zero values are negative.

pchote:
For definitions with 3 entries splitting[2] appears to be the stride, and should only be written if it is different to splitting[1].

Mailaender:
Updated the stride calculation with your recent findings.

pchote:
I think i've done enough nit-picking now.  The generated sequences look sensible, so :+1:.

pchote:
This feature isn't shipped, so omitting changelog (and removing the entry for the rules importer).

PR #10716, opened 2016-02-07
Re-enabled All Packages which seems to be fast enough now.

Mailaender:
It pops up instantly even for Red Alert 2. Not sure what kind of performance problems we were suffering in the past, they seem to be gone now. This is way more practical if you are setting up sequences with sprites scattered around in tileset specific MIX files.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
@@ -22,7 +22,8 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 {
 	public class AssetBrowserLogic : ChromeLogic
 	{
-		static string[] allowedExtensions;
+		readonly string[] allowedExtensions;
+		readonly IEnumerable<IReadOnlyPackage> acceptablePackages;
 
 		readonly World world;
 
@@ -216,6 +217,9 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			else
 				allowedExtensions = new string[0];
 
+			acceptablePackages = Game.ModData.ModFiles.MountedPackages.Where(p =>
+				p.Contents.Any(c => allowedExtensions.Contains(Path.GetExtension(c).ToLowerInvariant())));
+
 			assetList = panel.Get<ScrollPanelWidget>("ASSET_LIST");
 			template = panel.Get<ScrollItemWidget>("ASSET_TEMPLATE");
 			PopulateAssetList();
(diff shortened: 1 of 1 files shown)


pchote:
The performance issues would have been fixed by #10535.

pchote:
Could you please make the selection default to "all packages" instead of the first package?  :+1: after that change.

pchote:
Another simple usability improvement would be to only show packages that have viewable content in the dropdown.  You can do this by counting the contents of the packages that match the supported filetypes.

Mailaender:
Updated.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -217,6 +218,16 @@ public AssetBrowserLogic(Widget widget, Action onExit, World world, Dictionary<s
 			else
 				allowedExtensions = new string[0];
 
+			foreach (var package in Game.ModData.ModFiles.MountedPackages)
```
pchote: You may be able to simplify this to something like `var acceptablePackages = Game.ModData.ModFiles.MountedPackages.Where(p => p.Contents.Any(c => allowedExtensions.Contains(Path.GetExtension(c))));`

Mailaender: Simplified.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -355,14 +356,8 @@ void PopulateAssetList()
...
 
-			// TODO: This is too slow to run in the main thread
-			// var files = AssetSource != null ? AssetSource.AllFileNames() :
-			// GlobalFileSystem.MountedFolders.SelectMany(f => f.AllFileNames());
-			if (assetSource == null)
-				return;
-
-			var files = assetSource.Contents.OrderBy(s => s);
-			foreach (var file in files)
+			var files = assetSource != null ? assetSource.Contents : Game.ModData.ModFiles.MountedPackages.SelectMany(f => f.Contents);
```
obrakmann: Can you tack a `.Distinct()` onto the end here, please? Otherwise you get duplicate entries in the `all packages` view for stuff where we have replacements in the bits directory.

obrakmann:
Otherwise fine, :+1:

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
Mailaender: Sure. Thanks for the hint.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/430fd4f7065a4806dfd92f32dd61ddc5d74f445f)

PR #10720, opened 2016-02-07
Do not crash during importing a TS map if a waypoint ID isn't an integer.

GraionDilach:
While FinalAlert itself does not insert waypoints with IDs having alpabetical characters, mappers could manually place waypoints where the IDs could contain such.

### DIFF
--- a/OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
+++ b/OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
@@ -347,7 +347,8 @@ namespace OpenRA.Mods.TS.UtilityCommands
 				var dy = rx + ry - fullSize.X - 1;
 				var cell = new MPos(dx / 2, dy).ToCPos(map);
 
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
```diff
@@ -347,7 +347,8 @@ void ReadWaypoints(Map map, IniFile file)
 				var dy = rx + ry - fullSize.X - 1;
 				var cell = new MPos(dx / 2, dy).ToCPos(map);
 
-				var ar = new ActorReference(int.Parse(kv.Key) <= 7 ? "mpspawn" : "waypoint");
+				int wpindex;
+				var ar = new ActorReference((int.TryParse(kv.Key, out wpindex) == false || wpindex > 7) ? "waypoint" : "mpspawn");
```
Phrohdoh: `!int.TryParse(..`

GraionDilach: Ah, true. I'm rusty.

penev92:
Looks OK :+1:

PR #10721, opened 2016-02-07
Fix mod assembly loading on Windows.

pchote:
This fixes a regression from #10706 that would cause multiple copies of the mod assemblies to be loaded into the domain and subsequent bogus crashes.


pchote:
Note that this hack isn't fundamentally different from what we were already doing: i've just moved it from FileSystem to ObjectCreator, cleaned it up a bit, and changed it to key on data hash instead of filename (which is no longer reliable across mods).

abcdefg30:
Mods load for me again. :+1: / :white_check_mark:

Mailaender:
Mod switching still works fine on Linux. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/430fd4f7065a4806dfd92f32dd61ddc5d74f445f...b9a1bcd38e77afcc5a6952fe0b59ea1293f9b886)

PR #10722, opened 2016-02-08
Remove asset requirement from dedicated servers and the resource site.

pchote:
Partially reverts #10687 so that we don't force the asset requirement on all dedicated servers and the resource site.  The ContentInstaller tests when launching the game itself ensures that the content is installed before running the game.

This is intended as a shortish term fix for until the server code is adapted to handle itself properly.


obrakmann:
This just made me realize just how subtle the difference between packages without the optional flag and the files listed in the `ContentInstaller`'s `TestFiles` list really is.

Anyway, tested a dedicated server, which worked again. :+1:

Mailaender:
The reverts of https://github.com/pchote/OpenRA/commit/ee7403d2ae9b3461af3f1c0a5920a83f663de388 https://github.com/pchote/OpenRA/commit/3b58de770c215c3f226a597fdd729d4d9bc9c50f aren't necessary. Can you put those in a separate pull request? I would like to decide on those in isolation.

pchote:
If you'd made it clear that your request was a blocker at the time we could have cleared this up much earlier.  I consider them to be an integral part of this PR, which switches from solution (2) mentioned in #10687 to solution (1).

The general consensus on IRC (from at least myself, @penev92, and @Phrohdoh) was that downloading the asset files to travis was a horrible idea, but that the benefits outweighed the horribleness.  It turns out that the benefits did not.

obrakmann:
FWIW, I only agreed to the assets on travis thing out of necessity, not because I thought it's a grand idea. I'm all for removing it again.

Mailaender:
What is the downside of having the assets cached on Travis? We check for bogus sprite references which will catch crash bugs. I consider those vital for stable releases.

GraionDilach:
The assets are really static, we don't have a mod which is still under heavy implementation in artwork. I don't see why local sequence checks wouldn't be enough without Travis's sequence checks.

Mailaender:
Seems like I am out-voted pretty much here. The only positive I see here is that it speeds up the Travis CI checks a tiny bit. It has grown quite a lot compared to the fast AppVeyor builds.

Issue #10661, opened 2016-01-31
Drop Windows XP and .NET 4.0 support.


pchote:
Maintaining Windows XP support means that we are stuck with .NET 4.0, which is holding us back in a number of ways (which I expect to grow as time goes on):

1.  We can't fix our broken UPNP support because OpenNAT has only ever supported .NET >= 4.5 (#7140).  This then leads to confusion from people who don't know why we have a port forwarding option that can't obviously be enabled.
2.  The GeoIP library has recently dropped support for .NET < 4.5, so we are now stuck on an older version that won't receive bugfixes and makes life more difficult for linux packagers (#10280).
3.  The Roslyn compiler will give us access to much more powerful analysis tools for linting and documentation generation, but can only target the .NET >= 4.5 runtimes.
4.  Xamarin now only maintain partial support the 4.0 runtime by shipping pre-compiled assemblies (it is no longer possible to compile them from source, AFAIK).   Linux distros such as Fedora and Debian testing will not distribute these, so we can't support OpenRA on these platforms unless the player compiles/installs their own version of mono.
5. Microsoft have dropped all support for .NET 4.0 (https://support.microsoft.com/en-gb/lifecycle?C2=548), and Windows > XP support 4.5.2 or greater.

Depends on #10662.  Pulling the rug out from anyone who does use XP isn't very polite, so we need to know how big of an effect this will have.

AoAGeneral:
According to https://www.netmarketshare.com/operating-system-market-share.aspx?qprid=10&qpcustomd=0 it is showing that 52% is under Windows 7.

According to http://store.steampowered.com/hwsurvey within the first month of January 32% show using Windows 7.

Going by http://www.w3schools.com/browsers/browsers_os.asp It shows a more detailed usage of the OS in category and in type.

News feed from softpedia at http://news.softpedia.com/news/windows-10-becomes-the-third-most-used-desktop-os-in-the-world-496203.shtml states that Windows 7 of novemeber last year is the most used OS. Windows 10 being the third most used.

Back when Vista made its release no one wished to make the switch due to how buggy the system was. Particularly in the time that 64 bit was still a new technology and running 4gb of ram on windows XP was still enough.

It started to get to a point where people wanted more then 4gb due to advancements of video editing and other software designs which released windows XP 64 editions. However, Windows 7 was close on the rise so there were people who waited. (Estimated number whom waited I am unsure of)

[...]

PR #10726, opened 2016-02-09
add troubleshooting info for SDK path error

tmsbrg:
Tell me if this is stupid and should be fixed another way. But this would have helped me find out what this error meant and fix it for myself faster(Google wasn't too helpful this time).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -69,3 +69,8 @@ Gentoo
 * dev-lang/lua-5.1.5
 * x11-misc/xdg-utils
 * gnome-extra/zenity
+
+Troubleshooting
+-------------
+
+* __getting `error CS8001: Warning as Error: SDK path could not be resolved'__: This means you don't have the right version of the Mono SDK installed. Some systems, such as Debian testing, install 4.5 by default instead of 4.0. It's possible to edit the Makefile by changing -sdk:4.0 to -sdk:4.5 to still compile.


pchote:
Related to #10661.

obrakmann:
We could also add that to the `Compiling` wiki page (why don't we auto-generate `INSTALL.md` from that?).

At least we point people to the `Compiling` page more often than we tell them to read `INSTALL.md`.

Mailaender:
This is worded like a https://github.com/OpenRA/OpenRA/wiki/FAQ entry so it might be placed there. https://github.com/OpenRA/OpenRA/blob/bleed/INSTALL.md has very bad discover-ability. We should probably consolidate our documentation a bit.

Mailaender:
> (why don't we auto-generate `INSTALL.md` from that?).

The tarball published at https://github.com/OpenRA/OpenRA/releases won't contain it.

pchote:
Why don't we remove the compiling page and point people to INSTALL.md instead?

Mailaender:
The wiki has the benefit that it is quicker to update by anyone.

tmsbrg:
I didn't even know about the wiki page, so maybe the line in INSTALL.md should instead link to the wiki, and the wiki should contain this troubleshooting info. I'm used to just cloning the repository and checking INSTALL.md and README.md for instructions.

tmsbrg:
I've updated the [Compiling](https://github.com/OpenRA/OpenRA/wiki/Compiling) page on the wiki. So this is no longer needed to be in INSTALL.md. Probably still need a link to the wiki in INSTALL.md. But I'll close this pull request as it's no longer relevant. If you see any errors I made updating the wiki page, feel free to fix them.

PR #10740, opened 2016-02-13
Fix Missile projectile crashing with NRE if it has no Image

### LINKED ISSUE #10739: Missiles crash with NRE if projectile Image is null
Unlike `Bullet`, `Missile` doesn't `yield break` in `IEnumerable<IRenderable> Render` when `anim = null`.
This currently makes invisible tracking projectiles (or tracking projectiles where only the trail and/or contrail is visible) impossible unless you work around it with an empty shp, so I consider this a bug or at least an unnecessary limitation.

I'll submit a fix shortly.

reaperrr:
Closes #10739.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -841,6 +841,9 @@ namespace OpenRA.Mods.Common.Effects
 			if (info.ContrailLength > 0)
 				yield return contrail;
 
+			if (anim == null)
+				yield break;
+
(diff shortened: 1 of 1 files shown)


Mailaender:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7c1eded2a56a35d66997aa4b0b741df3a0b02d8c)

PR #10742, opened 2016-02-13
Fix max view distance check.

mazarf:
A range of 51 wasn't stopped by the original statement.  It triggered an out of bounds exception from TilesByDistance a few lines later.

This is my first pull request.  Let me know if I did something wrong.

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -1151,7 +1151,7 @@ namespace OpenRA
 			if (maxRange < minRange)
 				throw new ArgumentOutOfRangeException("maxRange", "Maximum range is less than the minimum range.");
 
-			if (maxRange > TilesByDistance.Length)
+			if (maxRange >= TilesByDistance.Length)
 				throw new ArgumentOutOfRangeException("maxRange", "The requested range ({0}) exceeds the maximum allowed ({1})".F(maxRange, MaxTilesInCircleRange));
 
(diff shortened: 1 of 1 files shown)


RoosterDragon:
:+1: Nice catch.

If you like, you're welcome to add yourself to the [AUTHORS](https://github.com/OpenRA/OpenRA/blob/bleed/AUTHORS) file.

obrakmann:
Actually, even 50 triggers the OOB exception. 49 works (on bleed). With 51 I get the exception shown below, suggesting that somewhere along the line the requested range is incremented by one.

```
Exception of type `System.ArgumentOutOfRangeException`: The requested range (52) exceeds the maximum allowed (50)
```

mazarf:
Yeah I noticed that.  It's caused by line 85 in Shroud.cs.  The extra 512 bumps it up to 51.

```
// Account for potential extra half-cell from odd-height terrain
var r = (range.Length + 1023 + 512) / 1024;
```

obrakmann:
Ah, I see. Although that sucks, that means that `FindTilesInCircle` still does the correct thing. I think we don't have that 50 cell maximum documented anywhere, at least not in the trait documentation, so we're in the clear there. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/420e400338df4fc1c880bcf7a003958562da127f)

PR #10751, opened 2016-02-13
Improve Util.Shuffle

RoosterDragon:
Reduce algorithmic complexity from O(n^2) to O(n).

### DIFF
--- a/OpenRA.Mods.Common/Util.cs
+++ b/OpenRA.Mods.Common/Util.cs
@@ -77,16 +77,21 @@ namespace OpenRA.Mods.Common
 			return WPos.Lerp(w.Map.CenterOfCell(from), w.Map.CenterOfCell(to), 1, 2);
 		}
 
-		/* pretty crap */
 		public static IEnumerable<T> Shuffle<T>(this IEnumerable<T> ts, MersenneTwister random)
 		{
-			var items = ts.ToList();
-			while (items.Count > 0)
+			// Fisher-Yates
+			var items = ts.ToArray();
+			for (var i = 0; i < items.Length - 1; i++)
 			{
-				var t = items.Random(random);
-				yield return t;
-				items.Remove(t);
+				var j = random.Next(items.Length - i);
+				var item = items[i + j];
+				items[i + j] = items[i];
+				items[i] = item;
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Util.cs
```diff
@@ -77,16 +77,21 @@ public static WPos BetweenCells(World w, CPos from, CPos to)
 			return WPos.Lerp(w.Map.CenterOfCell(from), w.Map.CenterOfCell(to), 1, 2);
 		}
 
-		/* pretty crap */
 		public static IEnumerable<T> Shuffle<T>(this IEnumerable<T> ts, MersenneTwister random)
 		{
-			var items = ts.ToList();
-			while (items.Count > 0)
+			// Fisher�Yates
```
Mailaender: @MonoDevelop does not like the special char as it seems:

image

Mailaender:
Still works :white_check_mark:  and I even understood the algorithm: https://en.wikipedia.org/wiki/Fisher-Yates_shuffle :+1:

OpenRA.Mods.Common/Util.cs
RoosterDragon: I guess I copy pasted in some darned fancy dash or something. Amended.

abcdefg30:
One thing I still don't get is that we only shuffeled the items and returned each of them before, but now we also allow duplicates or not returning one of the items at all, correct?

RoosterDragon:
No - once we choose an item at random we swap it to the front of the array. When we choose the next item at random we only select from the items towards the end of the array that haven't been chosen yet.

This means we cannot select an item more than once, and that we must select each at least once. i.e. The items remain the same - only the order changes.

abcdefg30:
Ah derp, should have seen that myself.

Anyway, this works ingame, too. :+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1c17e3e7bfc1bfa396082e6fbd17ca092859a38c...bf31caad8ad1958f1485a491988969392302be87)

PR #10756, opened 2016-02-14
Added the Tiberian Sun beacon

Mailaender:
While cleaning up the existing `PlaceBeacon` I noticed that it is over-engineered for the artwork used in RA2/TS so I created a new simplified trait.

### DIFF
--- a/OpenRA.Mods.TS/Effects/AnimatedBeacon.cs
+++ b/OpenRA.Mods.TS/Effects/AnimatedBeacon.cs
@@ -1,0 +1,64 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Effects;
+using OpenRA.Graphics;
+using OpenRA.Scripting;
+
+namespace OpenRA.Mods.TS.Effects
+{
+	public class AnimatedBeacon : IEffect
+	{
+		readonly Player owner;
+		readonly WPos position;
+		readonly string beaconPalette;
+		readonly bool isPlayerPalette;
+		readonly Animation beacon;
+
+		public AnimatedBeacon(Player owner, WPos position, int duration, string beaconPalette, bool isPlayerPalette, string beaconImage, string beaconSequence)
+		{
+			this.owner = owner;
+			this.position = position;
+			this.beaconPalette = beaconPalette;
+			this.isPlayerPalette = isPlayerPalette;
+
+			if (!string.IsNullOrEmpty(beaconSequence))
+			{
+				beacon = new Animation(owner.World, beaconImage);
+				beacon.PlayRepeating(beaconSequence);
+			}
+
+			if (duration > 0)
+				owner.World.Add(new DelayedAction(duration, () => owner.World.Remove(this)));
+		}
+
+		public void Tick(World world)
+		{
+			if (beacon != null)
+				beacon.Tick();
+		}
+
+		public IEnumerable<IRenderable> Render(WorldRenderer r)
+		{
+			if (beacon == null)
+				return Enumerable.Empty<IRenderable>();
+
+			if (!owner.IsAlliedWith(owner.World.RenderPlayer))
+				return Enumerable.Empty<IRenderable>();
+
+			var palette = r.Palette(isPlayerPalette ? beaconPalette + owner.InternalName : beaconPalette);
+			return beacon.Render(position, palette);
+		}
+	}
+}
--- a/OpenRA.Mods.TS/Player/PlaceSimpleBeacon.cs
+++ b/OpenRA.Mods.TS/Player/PlaceSimpleBeacon.cs
@@ -1,0 +1,81 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Mods.TS.Effects;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.TS.Traits
+{
+	[Desc("A beacon that consists of a single sprite that can be animated.")]
+	public class PlaceSimpleBeaconInfo : ITraitInfo
+	{
+		public readonly int Duration = 30 * 25;
+		public readonly string NotificationType = "Sounds";
+		public readonly string Notification = "Beacon";
+
+		public readonly bool IsPlayerPalette = false;
+		[PaletteReference("IsPlayerPalette")] public readonly string Palette = "effect";
+
+		public readonly string BeaconImage = "beacon";
+		[SequenceReference("BeaconImage")] public readonly string BeaconSequence = "idle";
+
+		public object Create(ActorInitializer init) { return new PlaceSimpleBeacon(init.Self, this); }
+	}
+
+	public class PlaceSimpleBeacon : IResolveOrder
+	{
+		readonly PlaceSimpleBeaconInfo info;
+		readonly RadarPings radarPings;
+
+		AnimatedBeacon playerBeacon;
+		RadarPing playerRadarPing;
+
+		public PlaceSimpleBeacon(Actor self, PlaceSimpleBeaconInfo info)
+		{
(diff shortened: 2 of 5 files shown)


OpenRA.Mods.TS/Effects/AnimatedBeaconSprite.cs
```diff
@@ -0,0 +1,60 @@
...
+				beacon.Tick();
+		}
+
+		public IEnumerable<IRenderable> Render(WorldRenderer r)
+		{
+			if (!owner.IsAlliedWith(owner.World.RenderPlayer))
+				yield break;
+
+			var palette = r.Palette(isPlayerPalette ? beaconPalette + owner.InternalName : beaconPalette);
+			foreach (var b in beacon.Render(position, palette))
```
abcdefg30: This will crash when `beacon == null`, I guess?

abcdefg30:

OpenRA.Mods.TS/Effects/AnimatedBeaconSprite.cs
Mailaender: Fixed.

OpenRA.Mods.TS/Effects/AnimatedBeaconSprite.cs
```diff
@@ -0,0 +1,63 @@
...
+			if (beacon != null)
+				beacon.Tick();
+		}
+
+		public IEnumerable<IRenderable> Render(WorldRenderer r)
+		{
+			if (!owner.IsAlliedWith(owner.World.RenderPlayer))
+				yield break;
+
+			if (beacon == null)
```
abcdefg30: Sorry to interrupt again, but I think we could move this check above the `IsAlliedWith` check to save some perf. ^^

Mailaender: Resorted.

OpenRA.Mods.Common/Scripting/Global/BeaconGlobal.cs
```diff
@@ -26,9 +26,10 @@ public BeaconGlobal(ScriptContext context) : base(context)
 
 		[Desc("Creates a new beacon that stays for the specified time at the specified WPos. " +
 			"Does not remove player set beacons, nor gets removed by placing them.")]
-		public Beacon New(Player owner, WPos position, int duration = 30 * 25, bool showRadarPings = true, string palettePrefix = "player")
+		public Beacon New(Player owner, WPos position, int duration = 30 * 25, bool showRadarPings = true,
+			string palette = "player", bool isPlayerPalette = true, string beaconCollection = "beacon", string arrowSprite = "arrow", string circleSprite = "circle")
```
pchote: I don't like the idea of exposing sequence / animation names to lua.

mods/ts/rules/nod-support.yaml
```diff
@@ -296,6 +296,9 @@ NAMISL:
 		DisplayBeacon: False
 		DisplayRadarPing: True
 		BeaconPoster:
+		ClockSprite:
```
pchote: Nulling out properties in yaml isn't very nice.  The default should be empty, and the other mods override it with their additions.

mods/ts/rules/nod-support.yaml
```diff
@@ -296,6 +296,9 @@ NAMISL:
 		DisplayBeacon: False
```
pchote: This doesn't show a beacon, so the overrides below shouldn't even be needed...?

OpenRA.Mods.TS/Effects/AnimatedBeaconSprite.cs
```diff
@@ -0,0 +1,63 @@
...
+
+using System;
+using System.Collections.Generic;
+using OpenRA.Effects;
+using OpenRA.Graphics;
+using OpenRA.Scripting;
+
+namespace OpenRA.Mods.TS.Effects
+{
+	public class AnimatedBeacon : IEffect
```
pchote: Why did you add all those extra properties to the other beacon type, if you are then making a new one for TS?

mods/ts/rules/nod-support.yaml
Mailaender: The linter doesn't care for that boolean and will otherwise complain.

OpenRA.Mods.TS/Effects/AnimatedBeaconSprite.cs
Mailaender: The initial plan was to use the existing beacon code, but it is so hard-coded to our CTF + remapped rings workaround that I created a new simplified one. I kept the un-hardcoding for documentation and lint testing.

OpenRA.Mods.Common/Scripting/Global/BeaconGlobal.cs
Mailaender: Any idea how else to solve it here?

OpenRA.Mods.TS/Effects/AnimatedBeaconSprite.cs
pchote: This has made a bit of a confusing mess.

OpenRA.Mods.TS/Effects/AnimatedBeaconSprite.cs
```diff
@@ -0,0 +1,63 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
```
obrakmann: I can't comment on the functionality, but you should use the updated copyright header here. It's also 2016 :)

OpenRA.Mods.TS/Effects/AnimatedBeaconSprite.cs
Mailaender: Moved that into https://github.com/Mailaender/OpenRA/tree/lint-beacon now.

Mailaender:
Updated.

OpenRA.Mods.TS/Player/PlaceSimpleBeacon.cs
```diff
@@ -0,0 +1,81 @@
...
+	public class PlaceSimpleBeaconInfo : ITraitInfo
+	{
+		public readonly int Duration = 30 * 25;
+		public readonly string NotificationType = "Sounds";
+		public readonly string Notification = "Beacon";
+
+		public readonly bool IsPlayerPalette = false;
+		[PaletteReference("IsPlayerPalette")] public readonly string Palette = "effect";
+
+		public readonly string BeaconCollection = "beacon";
```
reaperrr: @pchote recently mentioned that the term `Collection` should only be used for UI.

I think `BeaconImage` would be better here and more consistent with other traits.

OpenRA.Mods.TS/Player/PlaceSimpleBeacon.cs
```diff
@@ -0,0 +1,81 @@
...
+		[SequenceReference("BeaconCollection")] public readonly string BeaconSprite = "idle";
```
reaperrr: I'd suggest `BeaconSequence` here for similar reasons.

OpenRA.Mods.TS/Effects/AnimatedBeacon.cs
```diff
@@ -0,0 +1,64 @@
...
+		readonly Animation beacon;
+
+		public AnimatedBeacon(Player owner, WPos position, int duration, string beaconPalette, bool isPlayerPalette, string beaconCollection, string beaconSprite)
+		{
+			this.owner = owner;
+			this.position = position;
+			this.beaconPalette = beaconPalette;
+			this.isPlayerPalette = isPlayerPalette;
+
+			if (!string.IsNullOrEmpty(beaconSprite))
```
reaperrr: `beaconSequence`

OpenRA.Mods.TS/Effects/AnimatedBeacon.cs
```diff
@@ -0,0 +1,64 @@
...
+			{
+				beacon = new Animation(owner.World, beaconCollection);
```
reaperrr: `beaconImage`

reaperrr:
:+1: after that.

Mailaender:
Renamed for consistency.

OpenRA.Mods.TS/Effects/AnimatedBeacon.cs
```diff
@@ -0,0 +1,64 @@
...
+		{
+			if (beacon == null)
+				yield break;
+
+			if (!owner.IsAlliedWith(owner.World.RenderPlayer))
+				yield break;
+
+			var palette = r.Palette(isPlayerPalette ? beaconPalette + owner.InternalName : beaconPalette);
+			foreach (var b in beacon.Render(position, palette))
+				yield return b;
```
RoosterDragon: Minor quibble: Can you make these `yield break;` into `return Enumerable.Empty<IRenderable>();`? I believe you can then return `beacon.Render(...)` directly as well.

Mailaender: Yes, indeed. Updated.

obrakmann:
:+1:

PR #10757, opened 2016-02-14
Added kerning support to SpriteFont.DrawText

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/5017. Depends on #10666.


OpenRA.Game/Graphics/SpriteFont.cs
```diff
@@ -71,11 +75,16 @@ public void DrawText(string text, float2 location, Color c)
...
-						(int)Math.Round(p.X + g.Offset.X, 0),
-						p.Y + g.Offset.Y));
+				var x = (int)Math.Round(p.X + g.Offset.X, 0);
+				var y = p.Y + g.Offset.Y;
+				Game.Renderer.RgbaSpriteRenderer.DrawSprite(g.Sprite, new float2(x ,y));
 				p.X += g.Advance;
+
+				if (face != null && face.HasKerning && i < text.Length - 1)
+				{
+					var kerning = face.GetKerning(face.GetCharIndex(s), face.GetCharIndex(text[i + 1]), KerningMode.Default);
```
pchote: You can make this cleaner by tracking the current and last face index each time through the loop and then updating only one each time around.

Mailaender: But I need current and next face here.

Mailaender:
Updated.

RoosterDragon:
The kerning values for our fonts are all coming back as zero for me - meaning this new code doesn't actually change the rendered output at all (at least on my system).

The code looks fine, and `face.HasKerning` reports true - so I'm not sure what's up. Does anybody else get properly kerned output on their system? Do our fonts actually include some kerning hints?

Mailaender:
I replaced the GNU FreeFont *.ttf with the successor format *.otf which now gives a visible change.

pchote:
Confirmed that kerning is working, but the reduced spacing seriously hurts the readability of the mod chooser description on my system:

screen shot 2016-02-23 at 22 50 49

RoosterDragon:
On my system, some of the glyphs seem to be off and contain chunks of the next glyph - not sure if that's an artefact of the font file or the code - but it sure looks odd.
openrafontkerning

Mailaender:
I can't reproduce the problem on Linux and I don't know enough about fonts to improve it for small fonts.

pchote:
I don't think there's much more we can do here until we can find a new set of fonts that both have the kerning metadata, and render properly under all our supported OS/Freetype versions. Moving to the Future milestone, unless somebody has an idea on how to solve this now(?).

Issue #8918, opened 2015-08-03
Add support for TS ice tiles


pchote:
This should be relatively straightforward:

* Add a new `IceLayer` trait that lives on the world.
   * Contains a `CellLayer<int>` representing the ice "strength" at each tile.
   * Scans the map during load and populates the initial state based on the tiles on the map.  Tiles that contain ice are switched with the special ice-growable water tiles.
   * Tiles that contain ice are replaced with the ice-growable water tile.
   * Artwork for each cell is picked based on the current cell strength and the neighbouring state (very similar to `D2KResourceLayer`).
   * Include a method for reducing the strength at a given cell.  If strength falls below 0 then remove the custom terrain type from the cell (allowing it to revert to "water").
   * Grow the ice using similar logic to our resource spawners: Every <configurable interval> increase the strength of <one or more> ice cell(s) that are either adjacent to another ice tile or adjacent to the shore.
* Add a new `DamagesIce` trait that lives on actors.
   * Each tick, check if the actor is on an ice tile.  If it is, apply damage.  If the the actor is no longer allowed in the cell after the damage is applied (i.e. it has broken the tile, but decoupled using the existing plumbing provided by terrain types and `Mobile`) then kill it with a special warhead that includes a damage type that triggers the spash death effect.

* Add a new `DamagesIce` warhead type.
   * This reduces strength of any ice tiles in the damage area.

Phrohdoh:
I did a bit more work on this today.
I have nothing to show yet, but you can follow along [here](https://github.com/Phrohdoh/OpenRA/tree/feat%23ice) if you are so inclined.

Phrohdoh:
I've updated the linked branch again with all I can do.
Removing my assignment.

Mailaender:
I now have `IceLayer` which scans the map for initial state to assign strength values and `RenderIceState` which displays those as a debug overlay:

![image](https://cloud.githubusercontent.com/assets/756669/13035664/b22d329c-d355-11e5-88f8-75db38ef5c13.png)

fruestueck:
Currently it's possible to build apon ice.
![unbenannt](https://user-images.githubusercontent.com/25386322/36567419-a46e8724-1826-11e8-8805-142007afc577.JPG)

teinarss:
> 
>     * Add a new `DamagesIce` trait that lives on actors.
>       
[...]

Issue #10730, opened 2016-02-10
Refactor Domain Index possible performance improvements.


Codes4Fun:
Just some observations regarding the domain indices, and not from actual testing for performance issues (my computer is probably too fast to notice performance issues).

1) domain indices are built after a map is loaded, it seems as though it is something that could be done before or when saving the map and avoided when loading the map. It will increase the map size but it is easily compressed, maybe simple run length encoding will be enough.
2) If the map starts with a destroyed bridge, you end up with separate domains which is good for performance (early rejection before heavier path finding). When the bridge is repaired it creates a permanent connection between the domains to not reject path finding across those domains but when it is destroyed again those connections are not removed so you lose that benefit and will do heavier path finding each time after.
3) If the map starts with a bridge not destroyed it creates a single domain between both sides of the bridge, it loses out on the early rejection from #2 when the bridge is destroyed. It will do a heavier path search if there are no connection between both sides of the bridge.

So for #2 and #3, it would make sense that bridges are somehow ignored when building a domain index, separating both sides of a bridge into separate domains (unless there is a non-bridge connection between both sides of course). Bridge tiles getting their own domain value would probably work best for that and will also benefit in early rejecting trying to path to a destroyed bridge. And 'living' bridges would reference connections between domains, if a bridge is destroyed it dereferences the connection, and the connection is removed when there are zero references (zero bridges between domains).

I'm thinking on working on this when I get some time, maybe do some profiling first, but I wanted to throw it on here first.
##

GraionDilach:
This is actually interesting, because TS ice logic might also get affected by this - where if units stood too much time on the "bridge", the tiles simply broke.

pchote:
For the purposes of pathfinding there's no difference between bridges and ice.

Codes4Fun:
For (1) I looked at perf.log after loading lots of maps in the 4 mods and found that in the worst case scenarios DomainIndex.WorldLoad was only 7% of load time somewhere between 20 ms and 170 ms. Making changes DomainIndex generation does not seem like it would save a lot of time and probably not worth the effort or complexity it would add.

I did notice that for all mods except D2k, most of the time was spent in PrepareMap, and most of that time (more than 1 second) is not currently profiled. I would say the average seconds for TS is 1.8, RA 1.5, TD 1. The peak I saw was 2.7 seconds for the largest map in TS.

pchote:
[...]

PR #10760, opened 2016-02-14
Fixed Tiberian Sun units walking over icy water

Mailaender:
First steps towards https://github.com/OpenRA/OpenRA/issues/8918. This sets the initial state correctly. Also adds a `/debugice` chat command.

image

### DIFF
--- a/OpenRA.Mods.TS/OpenRA.Mods.TS.csproj
+++ b/OpenRA.Mods.TS/OpenRA.Mods.TS.csproj
@@ -74,9 +74,9 @@
     <Compile Include="Traits\Render\WithVoxelWalkerBody.cs" />
     <Compile Include="Traits\Render\WithVoxelUnloadBody.cs" />
     <Compile Include="Traits\Render\WithVoxelWaterBody.cs" />
-    <Compile Include="UtilityCommands\LegacyTilesetImporter.cs" />
     <Compile Include="Traits\World\TSShroudPalette.cs" />
     <Compile Include="UtilityCommands\ImportTSMapCommand.cs" />
+    <Compile Include="UtilityCommands\LegacyTilesetImporter.cs" />
   </ItemGroup>
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.
--- a/OpenRA.Mods.TS/OpenRA.Mods.TS.csproj
+++ b/OpenRA.Mods.TS/OpenRA.Mods.TS.csproj
@@ -69,11 +69,13 @@
     <Compile Include="Activities\VoxelHarvesterDockSequence.cs" />
     <Compile Include="SpriteLoaders\TmpTSLoader.cs" />
     <Compile Include="Traits\Buildings\TiberianSunRefinery.cs" />
+    <Compile Include="Traits\Render\RenderIceState.cs" />
     <Compile Include="Traits\Render\WithDockingOverlay.cs" />
     <Compile Include="Traits\Render\WithPermanentInjury.cs" />
     <Compile Include="Traits\Render\WithVoxelWalkerBody.cs" />
     <Compile Include="Traits\Render\WithVoxelUnloadBody.cs" />
     <Compile Include="Traits\Render\WithVoxelWaterBody.cs" />
+    <Compile Include="Traits\World\IceLayer.cs" />
     <Compile Include="Traits\World\TSShroudPalette.cs" />
     <Compile Include="UtilityCommands\ImportTSMapCommand.cs" />
     <Compile Include="UtilityCommands\LegacyTilesetImporter.cs" />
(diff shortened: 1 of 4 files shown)


OpenRA.Mods.TS/OpenRA.Mods.TS.csproj
```diff
@@ -69,14 +69,16 @@
     <Compile Include="Activities\VoxelHarvesterDockSequence.cs" />
     <Compile Include="SpriteLoaders\TmpTSLoader.cs" />
     <Compile Include="Traits\Buildings\TiberianSunRefinery.cs" />
+    <Compile Include="Traits\Render\RenderIceState.cs" />
     <Compile Include="Traits\Render\WithDockingOverlay.cs" />
     <Compile Include="Traits\Render\WithPermanentInjury.cs" />
     <Compile Include="Traits\Render\WithVoxelWalkerBody.cs" />
     <Compile Include="Traits\Render\WithVoxelUnloadBody.cs" />
     <Compile Include="Traits\Render\WithVoxelWaterBody.cs" />
-    <Compile Include="UtilityCommands\LegacyTilesetImporter.cs" />
```
GraionDilach: Whoops?

Mailaender: No, just resorting https://github.com/OpenRA/OpenRA/pull/10760/files#diff-c9b9ea950ff2d3777d20d61f3d92fbf9R81.

GraionDilach: Oh, ignore me then.

Mailaender: Put it into a separate commit now to clarify.

abcdefg30:
Look good to me. :+1: / :white_check_mark:

OpenRA.Mods.TS/Traits/Render/RenderIceState.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Drawing;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Commands;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.TS.Traits
+{
+	[Desc("Displays ice strength above the tile.")]
+	class RenderIceStateInfo : ITraitInfo
```
pchote: Both traits live on the world actor, so it would be better to `Requires<IceLayer>` here.

OpenRA.Mods.TS/Traits/World/IceLayer.cs
```diff
@@ -0,0 +1,125 @@
...
+		};
+
+		public object Create(ActorInitializer init) { return new IceLayer(init.Self, this); }
+	}
+
+	class IceLayer : IWorldLoaded
+	{
+		readonly IceLayerInfo info;
+
+		public CellLayer<int> Strength;
```
pchote: You can initialize this in the constructor and then make it readonly.

Mailaender:
Updated.

OpenRA.Mods.TS/Traits/World/IceLayer.cs
```diff
@@ -0,0 +1,125 @@
...
+				var template = w.TileSet.Templates[tile.Type];
+				if (info.StrengthPerTile.ContainsKey(template.Id))
+				{
+					var strength = info.StrengthPerTile[template.Id];
+					Strength[cell] = strength;
+
+					if (strength < 4)
+						w.Map.CustomTerrain[cell] = w.TileSet.GetTerrainIndex(info.ImpassableTerrainType);
+
+					// Saves some resources by only updating domain indices for previous water tiles.
```
pchote: What happens for strength = 2 or 3?

Mailaender: The initial state for all those tiles is `Clear` so they stay pathable.

pchote: In the line just above you explicitly set them to be impassable.

Mailaender: I am making the assumption that only the 100% water state (ice strength = 1) is relevant to separate larger land masses here. This saves us from checking all the decorational ice tiles until https://github.com/OpenRA/OpenRA/issues/10730 is solved and performance is not an issue anymore.

pchote: States 2 and 3 (what about 0?) separate the landmasses too, so this choice appears to leave you with inconsistent domains.

Mailaender: `0` is the cell layer's default value (no ice possible) which will be already taken care of at the initial domain index pass. I will set state 2 and 3 to passable for now.

pchote:
As discussed in IRC, I believe the real issue here is with the assignment of strengths to tiles.  This will need to be rethought a bit.

pchote:
IRC link for context: http://logs.openra.net/?year=2016&month=02&day=20#21:12:44

reaperrr:
There hasn't been any progress in over two months, so I'm closing it for now to clear the PR queue.

Mailaender:
I have no idea what to change here. Could simplify the assignments of strengths to

```csharp
enum Strength
{
	IcyWater = 1,
	CrackedIce = 2,
	Ice = 3
}
```

but that would then lose the information about neighbors and edges which I used to avoid glitches.

PR #10761, opened 2016-02-14
TD Balance Changes 22142016.

AoAGeneral:
TD Balance Changes 22142016.

Grenadier damage vs light increase to 80 from 75.

Increase APC HP to 210 from 200.

Guard Tower damage vs heavy reduced to 30 from 35.

Guard Tower gains armor type Wood.

SAM damage increase to 35 from 30.

Light Tank movement speed reduced to 110 from 113.

Light Tank Cost increased to 700 from 600. (Build time increase to 17
seconds from 15)

Light Tank HP Decreased to 340 from 350.

Rocket Infantry damage vs none decreased to 30 from 50.

Increase Apache damage vs wood to 80 from 50.

Stealth Tank damage vs wood increased to 100 from 75

AGT Damage vs wood increased to 100 from 25.

Obelisk damage vs wood increased to 100 from 50.

Gun Turret damage vs wood increased to 100 from 25.

The changes made to E3 is something I forgot about when testing. They
won't kill infantry like they were in testing.

The light tanks have been balanced to prevent en massing and adding a
higher price makes them more valuable instead of suicide tanks.

The armor type wood added to the guard tower made me realize how much of
a damage problem it was vs weapon types. (Gun turrets did only 25
damage?!) It has been calibrated and fixed.

An updated forum post of all the information will be here:
http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=17430&postdays=0&postorder=asc&start=75

### DIFF
--- a/mods/cnc/rules/structures.yaml
+++ b/mods/cnc/rules/structures.yaml
@@ -778,6 +778,8 @@ GTWR:
 	Building:
 	Health:
 		HP: 400
+	Armor:
+		Type: Wood
 	RevealsShroud:
 		Range: 7c0
 	Bib:
--- a/mods/cnc/rules/vehicles.yaml
+++ b/mods/cnc/rules/vehicles.yaml
@@ -94,7 +94,7 @@ APC:
 		ROT: 8
 		Speed: 128
 	Health:
-		HP: 200
+		HP: 210
 	Armor:
 		Type: Heavy
 	RevealsShroud:
@@ -301,7 +301,7 @@ JEEP:
 LTNK:
 	Inherits: ^Tank
 	Valued:
-		Cost: 600
+		Cost: 700
 	Tooltip:
 		Name: Light Tank
 		Description: Fast, light tank.\n  Strong vs Vehicles, Tanks\n  Weak vs Infantry
@@ -311,9 +311,9 @@ LTNK:
 		Queue: Vehicle.Nod
 	Mobile:
 		ROT: 7
-		Speed: 113
+		Speed: 110
 	Health:
-		HP: 350
+		HP: 340
 	Armor:
 		Type: Heavy
 	RevealsShroud:
--- a/mods/cnc/weapons/largecaliber.yaml
+++ b/mods/cnc/weapons/largecaliber.yaml
@@ -103,7 +103,7 @@ TurretGun:
 		Damage: 40
 		Versus:
 			None: 20
-			Wood: 25
+			Wood: 100
 			Light: 100
 			Heavy: 100
 		DamageTypes: Prone50Percent, TriggerProne, ExplosionDeath
--- a/mods/cnc/weapons/missiles.yaml
+++ b/mods/cnc/weapons/missiles.yaml
@@ -19,7 +19,7 @@ Rockets:
 		Damage: 35
 		ValidTargets: Ground, Air
 		Versus:
-			None: 50
+			None: 30
 			Wood: 85
 			Light: 100
 			Heavy: 100
@@ -227,7 +227,7 @@ MammothMissiles:
 		ValidTargets: Ground, Air
 		Versus:
 			None: 25
-			Wood: 75
+			Wood: 100
 			Light: 100
 			Heavy: 90
 		DamageTypes: Prone50Percent, TriggerProne, ExplosionDeath
@@ -294,7 +294,7 @@ TowerMissle:
 		ValidTargets: Ground
 		Versus:
 			None: 50
-			Wood: 25
+			Wood: 100
 			Light: 100
 			Heavy: 100
 		DamageTypes: Prone50Percent, TriggerProne, SmallExplosionDeath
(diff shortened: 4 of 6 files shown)


abcdefg30:
Let's see how this turns out in the next playtest. :+1:

reaperrr:
> The armor type wood added to the guard tower made me realize how much of a damage problem it was vs weapon types. (Gun turrets did only 25 damage?!) It has been calibrated and fixed.

I think the idea was that defenses are supposed to be weak vs. building armor types (wood and concrete in most cases) to make base crawling less effective (defenses outperform most units in terms of dps per $).

AoAGeneral:
@reaperrr 

Thats the point lol. Gun turrets do decent damage to structures EXCEPT Guard towers cause their damage type vs wood is only 25. I put it up to 100. The same issue was on AGTs as well as they did 100 damage type vs heavy and armored but only 25 to wood. Which means these defense structures did next to no damage at all. (IE: the guard tower can out repair gun turrets shots and the guard tower can kill the turret). This makes it so guard towers are killed by the correct structures and units.

GraionDilach:
Please do not refer the GT change as "gaining" or "adding" armor. OpenRA supports multiple layers of armor on a single actor and their protection would accumulate. I'd read that above as GT having two armors whatever it had and wood. You switched it's armor.

AoAGeneral:
@GraionDilach 

I added the text line Armor - Type: Wood under it because it had no armor context in the file. Otherwise what armor type did it have? O.o

GraionDilach:
Wait a second: https://github.com/OpenRA/OpenRA/blob/bleed/mods/cnc/rules/defaults.yaml#L456

All buildings have wood. That gets inherited through `^BaseBuilding` and `^Defense`. Your GT armor change should have no effect.

AoAGeneral:
@GraionDilach 

It does have an effect. Tested it with the change vs turrets and AGTs. Otherwise are all buildings utilize some basic armor type that is unlisted? Otherwise it has no category listed because if its damage type vs none then the damage modifiers do not match.

GraionDilach:
You changed turrets, Obeliskes and AGTs as well to affect wood. What I can think about is that the YAML inheritance bug which got fixed last week affects TD balance. `Buildings` start out as `Wood` in the template, and no template overrides it. Since you also use no suffixes, the GT ends up with a single `Wood` armor instance - you're overriding itself with itself.

You could also recheck bleed vs release with the GT armor override removed, but maintaining the rest. Although yes, you just made mid-caliber and high-caliber defenses cool for base destruction.

AoAGeneral:
@GraionDilach 

Then the problem persists. Because base destruction will cause affect due to GTs able to take out Turrets due to their disadvantage to the wood issue. Since GTs use no armor affect they count as wood and turrets, AGT, and obli make them soaky and tanky for fire. (Let alone the GT can kill gun turrets and even AGTs which should not be possible) So this would actually be a major problem with structures all basing under wood.

GraionDilach:
>AGT Damage vs wood increased to 100 from 25.
Obelisk damage vs wood increased to 100 from 50.
Gun Turret damage vs wood increased to 100 from 25.

This is from your own changelog which sounds like you're trying to deal with this problem, with ignoring the default `Wood` armor on all buildings!

If you want to make defenses better against GT but only the GT, your best bet is to do something like
```
GTWR:
	[...]
	Armor@GuardTower:
		Type: GuardTower

TowerMissile:
	[...]
	Warhead@1Dam: SpreadDamage
		[...]
		Versus:
			Wood: 25
			GuardTower: 400
```
where the armor suffix allows GTs to gain a secondary armor besides `Wood` and this `GuardTower` armor can multiply `Wood` back to your intended values (25%*400%  = 100%). 100% is the default value so you can omit that from warheads and thereby no DPS should be affected unless you explicitly list a versus for taht specific armor. The explicit 100 versus values can also be removed.

AoAGeneral:
@GraionDilach 

That is a ton of work and testing needed to make sure this all works. Specially if it has to be applied to all units (And possible future balance issues).

Might just have to drop the GT armor issue and hope no one abuses it until a better solution comes around. Because these actors and modifiers need testing and I want the light tank issues pushed through first.

GraionDilach:
Then revert that three lines I quoted from your changelog as well and remove the GT armor override. If you make defenses awesome against buildings just because of the GT, you're replacing an issue with a bigger one. I presented you a solution only affecting the GT and you decided to not take it because of... noidea.

I can't advise better than this.

AoAGeneral:
@GraionDilach 

"I presented you a solution only affecting the GT and you decided to not take it because of... no idea"

AoAGenera: "That is a ton of work and testing needed to make sure this all works. Specially if it has to be applied to all units (And possible future balance issues)."

That is the reason why. Im not good with this coding business so that will require a lot of trial and error to get done. While the solution was lobbied about half a year ago may have worked it created a loophole issue vs GT of defense structures and created a bottleneck issue. As for reverting I have no idea on how to do this and will close the PR and recreate.

PR #10764, opened 2016-02-15
Added static Tiberian Sun low bridges.

Mailaender:
This is a prototype towards https://github.com/OpenRA/OpenRA/issues/8916. In Tiberian Sun the individual bridge parts are overlay actors. They are not yet connected to the `BridgeLayer`, that is pretty much tied to converting terrain tiles to actors, which is not required here. For now they are indestructible and not repairable.


OpenRA.Mods.TS/Traits/Buildings/LowBridge.cs
```diff
@@ -0,0 +1,47 @@
...
+		public void Created(Actor self)
+		{
+			var buildingInfo = self.Info.TraitInfo<BuildingInfo>();
+			var cells = FootprintUtils.PathableTiles(self.Info.Name, buildingInfo, self.Location);
+			foreach (var cell in cells)
+				self.World.Map.CustomTerrain[cell] = self.World.TileSet.GetTerrainIndex(info.TerrainType);
+
+			var domainIndex = self.World.WorldActor.TraitOrDefault<DomainIndex>();
+			if (domainIndex != null)
+				domainIndex.UpdateCells(self.World, cells);
```
pchote: Having buildings update the domain index is setting off warning bells in my head.  This will need to be considered in more detail.

Mailaender: Shall we remove it from https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/Buildings/Bridge.cs#L267 as well?

pchote: Eventually, yes.  Your recent comment in #10730 about performance issues confirmed my heebie jeebies.

Mailaender:
Rebased and made sure this also works for destroyed bridges:

image

Added a new `/debugcustomterrain` command to verify.

mods/ts/sequences/bridges.yaml
```diff
@@ -0,0 +1,167 @@
+lobrdg01:
```
pchote: You can use yaml inheritance to avoid all this duplication.

mods/ts/rules/bridges.yaml
```diff
@@ -0,0 +1,212 @@
...
+	Inherits: ^LowBridge
+	Building:
+		Footprint: ___
+		Dimensions: 3, 1
+	LowBridge:
+
+LOBRDG19:
+	Inherits: ^LowBridge
+	Tooltip:
+		Name: Bridge Ramp
```
pchote: Is it really necessary to distinguish these as ramps (as opposed to just "Bridge")?

mods/ts/rules/bridges.yaml
```diff
@@ -0,0 +1,212 @@
...
+LOBRDG10:
+	Inherits: ^LowBridge
+	Building:
+		Footprint: ___
+		Dimensions: 3, 1
+	LowBridge:
+
+LOBRDG11:
+	Inherits: ^LowBridge
+	Building:
```
pchote: Can you pull out all these duplicated Building and LowBridge definitions into templates?

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
```diff
@@ -108,11 +110,14 @@ public bool ValidateArguments(string[] args)
 			{ 0x63, "lobrdg26" },
 			{ 0x64, "lobrdg27" },
 			{ 0x65, "lobrdg28" },
+
+			/*
 			{ 0x7A, "lobrdg1" },
```
pchote: Why are these four still disabled?

OpenRA.Mods.Common/Traits/World/EditorActorLayer.cs
```diff
@@ -104,7 +104,9 @@ EditorActorPreview Add(string id, ActorReference reference, bool initialSetup =
 
 			var preview = new EditorActorPreview(worldRenderer, id, reference, owner);
 			previews.Add(id, preview);
-			screenMap.Add(preview, preview.Bounds);
+
+			if (preview.Bounds.Width > 0 && preview.Bounds.Height > 0)
```
pchote: What is the reasoning behind this?  It looks like a workaround for bogus ActorPreviews, which should be fixed in another way.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
Mailaender: These are actually high bridge elements. #8917

OpenRA.Mods.Common/Traits/World/EditorActorLayer.cs
Mailaender: Dead low bridge actors come with invisible sprites. I will give them render editor only sprites.

mods/ts/sequences/bridges.yaml
Mailaender: That only works if the sprite image is identical. For example:

```yaml
tracks01:
	idle:
		UseTilesetExtension: true
		ZOffset: -1c0

tracks02:
	Inherits: tracks01
```

etc. won't get rendered.

OpenRA.Mods.Common/Traits/World/EditorActorLayer.cs
Mailaender: Will remove it although it is strange to keep this crash in the map editor exclusively.

mods/ts/sequences/bridges.yaml
Mailaender: It works for bridges and tunnel tops though, which makes it even more weird.

mods/ts/rules/bridges.yaml
```diff
@@ -0,0 +1,107 @@
+^NorthWestSouthEastBridge:
+	Inherits: ^LowBridge
+	Building:
+		Footprint: _ _ _
+		Dimensions: 1, 3
+	LowBridge:
```
pchote: Why not move these to ^LowBridge?

Mailaender: It will then complain `Building:` is missing.

pchote: Ok.

Mailaender:
> Please replace the broken bridges in A River Runs Near it with intact ones. The map isn't really playable chopped in half.

That would remove the test case for broken bridges. Tread Lightly isn't playable, either. If you want to ship only playable maps, we should remove them here and store them at the resource site instead.

pchote:
Importing each overlay type as its own actor type isn't going to work in the long term.  Can you please set this up with one bridge type for each direction/ramp with damage states, and placeholder sequences for a future PR (which I believe @penev92 has already written) that enables randomized selection of the three variants?

Mailaender:
Added a checkbox to the cheats menu.

Mailaender:
> Importing each overlay type as its own actor type isn't going to work in the long term. Can you please set this up with one bridge type for each direction/ramp with damage states, and placeholder sequences for a future PR (which I believe @penev92 has already written) that enables randomized selection of the four variants?

No, because it isn't that simple:
a) the random sprite selection isn't in place and I am not convinced mappers want it
b) you can't place damaged actors in the map editor, yet
c) the damaged bridge ends are directional, so picking the right variant will be tricky
d) legacy map import will break

and I don't want to jump ahead. This is just adding the bridges in a static state. I would like to keep it simple and step by step.

Mailaender:
Rebased.

pchote:
This issue is that this is not compatible with an "openra-like" final implementation, which would have bridges as normal actors with damaged and destroyed artwork states.  This adds additional work towards this end-goal, where the person who does the next steps is forced to redo the actor definitions from scratch and then write an upgrade rule to convert any existing bridges on maps.

>  d) legacy map import will break

You keep repeating this, but I don't understand your point.  There is nothing to break, because OpenRA doesn't support TS-style overlays.  The importer converts one concept from the original engine to a different concept in the OpenRA engine.  If something breaks, then that is an issue with the implementation that can be fixed.

Mailaender:
Don't worry as I will be the one finalizing the low bridges and additional work will only be running the updated importer again. Making bridges destructible (and repairable) by combining bridge actors into one is the next step, but I am happy that this already works and would like to submit this milestone.

pchote:
The temporary changes don't need to be merged into the main repository to act as a milestone.  You can develop on top of this locally, and then we can review and merge the next step with merged actors.

Mailaender:
> The temporary changes don't need to be merged into the main repository to act as a milestone.

This will however as is enable a lot of more possible maps and act as vanilla TS/RA2 skirmish with "indestructible bridges" already.

pchote:
There is nothing stopping you from implementing indestructible bridges with damage states, and having the importer set the initial health to use them.  I'm not asking you to implement bridge destruction, only to not import a bunch of unwanted new actor types that will later need to be thrown out.

pchote:
Reinforcing the point further, we actually have all the logic you need to implement bridge destruction.  The only bits we're still missing for complete low bridges are the random art selection (this could be done with a random upgrade trait switching different WithSpriteBody traits), and demolition/repair via bridge huts.

You can implement destruction by setting up two "normal" actors (one for each bridge orientation), two "dead" husk actors, and the four ramps (which I believe were indestructable in the original).  The normal actors have regular and damaged art states, and spawn the "dead" actors on death.  The "dead" actors use WithWallSpriteBody to render the damaged-end overhang when they are next to or between undamaged bridge segments.  The artwork for the different adjacency cases are in different files, which you assemble into a single sequence for the wall trait using an aggregate sequence.

Mailaender:
The problem is that I then can't reuse building and footprint, but generate it dynamically. I started to try using HealthInit and FacingInit, but that is a whole new approach throwing all of this into the trash bin.

PR #10519, opened 2016-01-15
Decrease Engineer and Medic priority

Xinayder:
Fixes issue #10113.

PR #10767, opened 2016-02-15
Undo lowering selection priority of medics and mechanics

obrakmann:
This undoes parts of #10519.

That PR made using medics and mechanics really cumbersome to use because you always have to manually select them after selecting the rest of your attack force if you want to issue attack move commands.
#8768 mostly fixed the problem of medics and mechanics running in front of the other units they are escorting, so it's safe to tell them to attack-move along with others. They will stop to repair/heal damaged units and not run head-on into the enemy.

The change from #10519 does make sense for engineers, though, so they keep the lower selection priority.

### DIFF
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -322,8 +322,6 @@ MEDI:
 		AttackSequence: heal
 	Voiced:
 		VoiceSet: MedicVoice
-	Selectable:
-		Priority: 5
 
 MECH:
 	Inherits: ^Soldier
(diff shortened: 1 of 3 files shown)


GraionDilach:
Reasons sound more than enough for me. :+1:

reaperrr:
[modified changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/6b044db6c076ad9378e033f405808d5502dc2d7f...8ad6d3d3778e2e8617a332a3b7a599164a1fa4db)

PR #10765, opened 2016-02-15
Don't reenumerate FileSystem when switching maps.

pchote:
This changes the filesystem to be mounted once at mod-load, and then left alone thereafter.  This then lets us drop the re-enumeration on map load, and cuts a couple of seconds off our load time.

This also fixes a bunch of `Game.modData` references, which gets us closer to removing the `// HACK: The engine code assumes that Game.modData is set.`.

This may break one or more of the utility commands.

Depends on #10703.

Issue #10717, opened 2016-02-07
Clean up the filesystem code


pchote:
The virtual filesystem (VFS) code is/was a giant ball of technical debt, and cleaning it up will win us a lot of perf and feature improvements.  This ticket tracks the previous and future progress on the issue.

- [x] Un-global the FileSystem class. [#9853]
- [x] Streamline the package interface, moving the various mix-format workarounds from engine code into the mix file parser. [#10525, #10530, #10535, #10554]
- [x] Replace direct-filesystem lookups with VFS lookups. [#10687, #10703, #10704, #10705, #10706, #10713]
- [x] Reference maps via their packages instead of paths [#10786]
- [x] Restore "All packages" listing in the asset browser [#10716]
- [ ] Separate Map and MapPreview, and only load Map when absolutely necessary:
  - [x] Remove Map references from MapPreview [#10697]
  - [x] Remove automatic map format updating [#10701]
  - [x] Load map metadata without initializing throwaway maps. [#10768]
  - [x] Load map previews without initializing throwaway maps.
  - [x] Use MapPreviews in the server code.
    - [ ] Support querying the resource API directly, removing the need for dedicated servers to sync maps. [#10717]
  - [x] Use MapPreviews in the lobby code.
- [x] Implement `IReadOnlyFileSystem` interface, and replace use of `Game.ModData.ModFiles` with a passed `IReadOnlyFileSystem` object. [#10765]
- [x] Split sequence initialization out of RuleSet, and store the SequenceProvider on the individual Map and defaults in a `DefaultSequences` dictionary (keyed by tileset) on ModData.
- [x] Remove the lazy loading in Map.
- [x] Add support for "oramod" packages. [[pchote/oramod](https://github.com/pchote/OpenRA/commits/oramod)]

pchote:
I have split #10993 into its own ticket because that turned out to be relatively separate from the other cleanups.

PR #10768, opened 2016-02-15
Avoid creating throwaway map objects at startup.

pchote:
This parses MapPreview objects instead of going to the effort of constructing a full map and then throwing it away.  Saves ~30% on map load time and is another step towards #10717.

Depends on #10765.  Only the last two commits are part of this PR.

### DIFF
--- a/OpenRA.Game/Manifest.cs
+++ b/OpenRA.Game/Manifest.cs
@@ -169,6 +169,11 @@ namespace OpenRA
 			return new ReadOnlyDictionary<string, string>(inner);
 		}
 
+		public bool Contains<T>() where T : IGlobalModData
+		{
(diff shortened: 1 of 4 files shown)


Phrohdoh:
s/initialising/initializing

obrakmann:
This looks fine as well, :+1:

Mailaender:
Red Alert mod map load time improvements for me

```diff
-577 ms
+286 ms
```

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ae9b02e501767d512d703923b138778f33f25cc5...1c17e3e7bfc1bfa396082e6fbd17ca092859a38c)

PR #10769, opened 2016-02-15
Fix the spacing at the end of the AUTHORS file

abcdefg30:
Polishes the ingame look.
Before:
endcredits
After:
endcreditsfixed

### DIFF
--- a/mods/cnc/chrome/credits.yaml
+++ b/mods/cnc/chrome/credits.yaml
@@ -22,6 +22,7 @@ Container@CREDITS_PANEL:
 					Y: 15
 					Width: PARENT_RIGHT - 30
 					Height: PARENT_BOTTOM - 30
+					TopBottomSpacing: 18
 					Children:
 						Label@CREDITS_TEMPLATE:
 							X: 8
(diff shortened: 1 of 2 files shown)


pchote:
People keep on breaking this by helpfully "fixing" the AUTHORS file, and I'm sure it will happen again.  How about we fix this in the game instead?  I believe we have a specific ScrollPanel property for the end of list margin.

obrakmann:
This will do it:
```diff
diff --git a/mods/cnc/chrome/credits.yaml b/mods/cnc/chrome/credits.yaml
index 3a806c8..1d90905 100644
--- a/mods/cnc/chrome/credits.yaml
+++ b/mods/cnc/chrome/credits.yaml
@@ -22,6 +22,7 @@ Container@CREDITS_PANEL:
                                        Y: 15
                                        Width: PARENT_RIGHT - 30
                                        Height: PARENT_BOTTOM - 30
+                                       TopBottomSpacing: 4
                                        Children:
                                                Label@CREDITS_TEMPLATE:
                                                        X: 8
```

abcdefg30:
Thanks @obrakmann. Updated. I went for a spacing of `18` though to make it look like in the opening post:
endcreditsfixedtd
(The only downside here might be that we also get additional spacing at the top,
but I think that still looks ok.)

obrakmann:
:+1:

Mailaender:
I am leaving this without a changelog as I can't reproduce the problem on the christmas release.

PR #10777, opened 2016-02-19
Fix missiles not moving to CruiseAltitude if launched above it

### LINKED ISSUE #10741: Missiles seem to ignore CruiseAltitude and MaximumLaunchAngle
They appear to launch and stay at a slightly upwards course until close enough to target when they start to descent (assuming it's a ground target).

Best observed with a missile projectile with high range, low speed, shadow and on flat terrain.

reaperrr:
Closes #10741.

Missiles can now descend towards CruiseAltitude as well.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -664,6 +664,11 @@ namespace OpenRA.Mods.Common.Effects
 					// of vertical facing bound by the maximum vertical rate of turn
 					var vDist = new WVec(-diffClfMslHgt - info.CruiseAltitude.Length, -speed, 0);
 					desiredVFacing = (sbyte)vDist.HorizontalLengthSquared != 0 ? vDist.Yaw.Facing : vFacing;
+
+					// If the missile is launched above CruiseAltitude, it has to descend instead of climbing
+					if (-diffClfMslHgt > info.CruiseAltitude.Length)
+						desiredVFacing = -desiredVFacing;
+
 					desiredVFacing = desiredVFacing.Clamp(-info.VerticalRateOfTurn, info.VerticalRateOfTurn);
 
 					ChangeSpeed();
@@ -675,6 +680,11 @@ namespace OpenRA.Mods.Common.Effects
 				// of vertical facing bound by the maximum vertical rate of turn
 				var vDist = new WVec(-diffClfMslHgt - info.CruiseAltitude.Length, -speed, 0);
 				desiredVFacing = (sbyte)vDist.HorizontalLengthSquared != 0 ? vDist.Yaw.Facing : vFacing;
+
+				// If the missile is launched above CruiseAltitude, it has to descend instead of climbing
+				if (-diffClfMslHgt > info.CruiseAltitude.Length)
+					desiredVFacing = -desiredVFacing;
+
 				desiredVFacing = desiredVFacing.Clamp(-info.VerticalRateOfTurn, info.VerticalRateOfTurn);
 
 				ChangeSpeed();
--- a/mods/ra/weapons/missiles.yaml
+++ b/mods/ra/weapons/missiles.yaml
@@ -14,6 +14,7 @@ Maverick:
 		Inaccuracy: 512
 		Image: DRAGON
 		HorizontalRateOfTurn: 5
+		CruiseAltitude: 2c0
 		RangeLimit: 60
 	Warhead@1Dam: SpreadDamage
 		Spread: 128


Mailaender:
This might be easier to review with a test case, e.g. a map with custom rules on the resource site.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -659,8 +659,13 @@ int IncreaseAltitude(int predClfDist, int diffClfMslHgt, int relTarHorDist, int
 				{
 					// Aim to attain cruise altitude as soon as possible while having the absolute value
 					// of vertical facing bound by the maximum vertical rate of turn
+					var dat = args.SourceActor.World.Map.DistanceAboveTerrain(pos);
 					var vDist = new WVec(-diffClfMslHgt - info.CruiseAltitude.Length, -speed, 0);
 					desiredVFacing = (sbyte)vDist.HorizontalLengthSquared != 0 ? vDist.Yaw.Facing : vFacing;
+
+					if (dat.Length > info.CruiseAltitude.Length)
+						desiredVFacing = (sbyte)vDist.HorizontalLengthSquared != 0 ? -vDist.Yaw.Facing : -vFacing;
```
RoosterDragon: Why not `desiredVFacing = -desiredVFacing;` ?

reaperrr: I just didn't think of that, thanks for the hint.

reaperrr:
Updated.

testcase on bleed: https://github.com/reaperrr/OpenRA/tree/missile-cruise-testcase-bleed
testcase with fix: https://github.com/reaperrr/OpenRA/tree/missile-cruise-testcase-fixed

In both cases, build a Mig and fire on a ground cell further away (15-20 cells recommended).
On bleed, even though the `CruiseAltitude` is 1024 WDist vs. the Mig's 2560 WDist, the missiles move *up* until they're very close to the target cell and ignore `CruiseAltitude`.

On the fixed testcase, the missiles will go down towards `CruiseAltitude` correctly and stay there until close enough to target.

The best unmodified test for potential regressions is probably the allied destroyer, since it launches missiles at 128 WDIst above ground and the missiles are relatively slow and long-range, making it a little easier to see the flight behavior. You should notice zero difference between bleed and this PR wrt the destroyer's missiles.

abcdefg30:
Looks good to me. :+1:

GraionDilach:
I honestly liked the behaviour on my helicopters but that can be remade via setting a `CruiseAltitude` of their `AirborneAltitude`, so :+1:

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -662,8 +662,13 @@ int IncreaseAltitude(int predClfDist, int diffClfMslHgt, int relTarHorDist, int
 				{
 					// Aim to attain cruise altitude as soon as possible while having the absolute value
 					// of vertical facing bound by the maximum vertical rate of turn
+					var dat = args.SourceActor.World.Map.DistanceAboveTerrain(pos);
 					var vDist = new WVec(-diffClfMslHgt - info.CruiseAltitude.Length, -speed, 0);
 					desiredVFacing = (sbyte)vDist.HorizontalLengthSquared != 0 ? vDist.Yaw.Facing : vFacing;
+
+					if (dat.Length > info.CruiseAltitude.Length)
```
pchote: Shouldn't this be checking against `diffClfMslHgt` instead of `dat`?

reaperrr: Possibly, I'll take a closer look later.

reaperrr: @pchote I'm not 100% sure given the missile code as a whole is quite a bit above me, but checking against `-diffClfMslHgt` instead of `dat.Length` indeed fixes the bug just as well on flat terrain and probably better on TS/RA2 cliffs (though I didn't notice any practical problems with TS cliffs while using `dat.Length`, either).

reaperrr:
Updated.
According to my own testing it still works, but I'm removing the Needs+2 since the code changed.

reaperrr:
Updated, gave the MiG's Maverick a higher `CruiseAltitude` to keep flight curve similar to bleed (looks better visually).

reaperrr:
@OpenRA/engine-hackers 
It would be nice if this made it into the upcoming stable release, otherwise I'll have to include a custom Missile projectile in TDX dll since the SSM Launcher is affected by this bug in my mod.

It's a simple fix that shouldn't cause any regressions and might benefit more 3rd-party mods than just TDX.

Phrohdoh:
This may be the issue that caused this headache:

ra2 missiles

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -664,6 +664,11 @@ int IncreaseAltitude(int predClfDist, int diffClfMslHgt, int relTarHorDist, int
 					// of vertical facing bound by the maximum vertical rate of turn
 					var vDist = new WVec(-diffClfMslHgt - info.CruiseAltitude.Length, -speed, 0);
 					desiredVFacing = (sbyte)vDist.HorizontalLengthSquared != 0 ? vDist.Yaw.Facing : vFacing;
+
+					// If the missile is launched above CruiseAltitude, it has to descent instead of climbing
```
pchote: descent &rarr; descend.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -675,6 +680,11 @@ int IncreaseAltitude(int predClfDist, int diffClfMslHgt, int relTarHorDist, int
 				// of vertical facing bound by the maximum vertical rate of turn
 				var vDist = new WVec(-diffClfMslHgt - info.CruiseAltitude.Length, -speed, 0);
 				desiredVFacing = (sbyte)vDist.HorizontalLengthSquared != 0 ? vDist.Yaw.Facing : vFacing;
+
+				// If the missile is launched above CruiseAltitude, it has to descent instead of climbing
```
pchote: descent &rarr; descend.

pchote:
Can you please rebase this? It's based on a two month old version of bleed that crashes on startup for me.

pchote:
I didn't notice any regressions in RA or TS, and it seems to fix the issue, so :+1:

reaperrr:
Updated.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/3b41864812c3b8711c1e1e510ab86ad37caffa1e)

PR #10779, opened 2016-02-19
Added friendly capturing with force modifier

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/5691.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Capturable.cs
+++ b/OpenRA.Mods.Common/Traits/Capturable.cs
@@ -18,9 +18,10 @@ namespace OpenRA.Mods.Common.Traits
 	{
 		[Desc("Type listed under Types in Captures: trait of actors that can capture this).")]
 		public readonly string Type = "building";
-		public readonly bool AllowAllies = false;
-		public readonly bool AllowNeutral = true;
-		public readonly bool AllowEnemies = true;
+
+		public readonly Stance TargetStances = Stance.Enemy | Stance.Neutral;
+		public readonly Stance ForceTargetStances = Stance.Enemy | Stance.Neutral;
+
 		[Desc("Health percentage the target must be at (or below) before it can be captured.")]
 		public readonly int CaptureThreshold = 50;
 		public readonly bool CancelActivity = false;
@@ -33,17 +34,11 @@ namespace OpenRA.Mods.Common.Traits
 			if (c == null)
 				return false;
 
-			var playerRelationship = owner.Stances[captor.Owner];
-			if (playerRelationship == Stance.Ally && !AllowAllies)
-				return false;
-
-			if (playerRelationship == Stance.Enemy && !AllowEnemies)
-				return false;
-
-			if (playerRelationship == Stance.Neutral && !AllowNeutral)
-				return false;
-
 			if (!c.CaptureTypes.Contains(Type))
+				return false;
+
+			var stances = ForceTargetStances | TargetStances;
+			if (!stances.HasStance(captor.Owner.Stances[owner]))
 				return false;
 
 			return true;
--- a/OpenRA.Mods.Common/Traits/EngineerRepair.cs
+++ b/OpenRA.Mods.Common/Traits/EngineerRepair.cs
@@ -106,6 +106,9 @@ namespace OpenRA.Mods.Common.Traits
 
 			public override bool CanTargetActor(Actor self, Actor target, TargetModifiers modifiers, ref string cursor)
 			{
+				if (modifiers != TargetModifiers.None)
+					return false;
+
 				if (!target.Info.HasTraitInfo<EngineerRepairableInfo>())
 					return false;
 
--- a/OpenRA.Mods.Common/Traits/EngineerRepair.cs
+++ b/OpenRA.Mods.Common/Traits/EngineerRepair.cs
@@ -26,6 +26,9 @@ namespace OpenRA.Mods.Common.Traits
 			"Possible values are Exit, Suicide, Dispose.")]
 		public readonly EnterBehaviour EnterBehaviour = EnterBehaviour.Dispose;
 
+		public readonly Stance TargetStances = Stance.Ally;
+		public readonly Stance ForceTargetStances = Stance.Enemy | Stance.Neutral | Stance.Ally;
+
 		public object Create(ActorInitializer init) { return new EngineerRepair(init, this); }
 	}
 
@@ -112,7 +115,9 @@ namespace OpenRA.Mods.Common.Traits
 				if (!target.Info.HasTraitInfo<EngineerRepairableInfo>())
 					return false;
 
-				if (self.Owner.Stances[target.Owner] != Stance.Ally)
+				var info = self.Info.TraitInfo<EngineerRepairInfo>();
+				var stances = modifiers == TargetModifiers.ForceAttack ? info.ForceTargetStances : info.TargetStances;
+				if (!stances.HasStance(self.Owner.Stances[target.Owner]))
 					return false;
 
 				if (target.GetDamageState() == DamageState.Undamaged)
@@ -126,7 +131,9 @@ namespace OpenRA.Mods.Common.Traits
 				if (!target.Info.HasTraitInfo<EngineerRepairableInfo>())
 					return false;
 
-				if (self.Owner.Stances[target.Owner] != Stance.Ally)
+				var info = self.Info.TraitInfo<EngineerRepairInfo>();
+				var stances = modifiers == TargetModifiers.ForceAttack ? info.ForceTargetStances : info.TargetStances;
+				if (!stances.HasStance(self.Owner.Stances[target.Owner]))
 					return false;
 
 				if (target.DamageState == DamageState.Undamaged)
--- a/OpenRA.Mods.Common/Traits/ExternalCapturable.cs
+++ b/OpenRA.Mods.Common/Traits/ExternalCapturable.cs
@@ -18,9 +18,10 @@ namespace OpenRA.Mods.Common.Traits
 	{
 		[Desc("Type of actor (the ExternalCaptures: trait defines what Types it can capture).")]
 		public readonly string Type = "building";
-		public readonly bool AllowAllies = false;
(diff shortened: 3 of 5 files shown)


GraionDilach:
Doesn't this break husks so they would require to use forcecapture?

Micr0Bit:
ew ... i dont like this tbh ... that smells like allowing mapmodders to "sneak in" a option for trolling ... capture your allies buildings without them getting notified about this rule-change

chrisforbes:
Either enable by default in some mod, or don't do it.

Mailaender:
> Doesn't this break husks so they would require to use forcecapture?

No.

Mailaender:
> Doesn't this break husks so they would require to use forcecapture?

Fixed.

Mailaender:
As I already mentioned this is nothing new, but already supported. https://github.com/OpenRA/OpenRA/issues/5691#issuecomment-48113533

At the moment you just need

```yaml
Rules:
	^Building:
		ExternalCapturable:
			AllowAllies: true
		-EngineerRepair:
```

for it to work. This fixes it.

Micr0Bit:
by supported i mean , "the idea of it getting obvious positive reactions" ... which isnt the case

abcdefg30:
Wow, why so offensive against this change guys?

@Micr0Bit I think there are already enough ways you can troll with custom maps... And you already get the "This map contains custom rules. Game experience may change." warning when someone selects such a map, so if you don't want to risk getting trolled, don't play on those maps... I could easily create a map that lets rifle man explode (nuke explosion) when they stay on the same cell or add a script that gives players named `abcdefg30` infinite mammoth paradrops, just as example that trolling is already possible in many ways.

> Either enable by default in some mod, or don't do it.

@chrisforbes what do you mean by that? If I didn't missunderstood it: Since we do we require explicit usecases for fixing bugs/adding features?

---------------------------------------------------------------------------------------------------------------------------------------

So, somehow I don't get your points, could you please elaborate a bit more?

Micr0Bit:
this notification ... "This map contains custom rules." doesnt work 100%

abcdefg30:
Then file a bug report, please, so that this gets fixed.

Micr0Bit:
still this doesnt look like it would close #5691 ... as xanax (and myself in the origin of the idea #5739 ) were looking for something to (atleast) turn on/off ingame 
something that extends the core-gameplay (worse enough now that im re-thinking it ) ... not adding another map-modder option

Mailaender:
@pchote already rejected the lobby option variant. https://github.com/OpenRA/OpenRA/issues/5691#issuecomment-51006790

Micr0Bit:
okay thats a point , then im indifferent on this

GraionDilach:
I'd prefer if this would have a notation like what `Armament` got, by having a `TargetModifier` and a `ForceTargetModifier`. This way seems arbitrary atm.

pchote:
> I'd prefer if this would have a notation like what Armament got, by having a TargetModifier and a ForceTargetModifier. This way seems arbitrary atm.

This sounds like the best approach to me.  We can also set up infiltration/repair/etc the same way.

MustaphaTR:
> the notification ... "This map contains custom rules." doesnt work 100%

I think this notification should have a clickable link to resource center, so people can easily see what the map changes.

Micr0Bit:
@MustaphaTR : people want to "play games" ... and not every single player to study the rules before they decide to actually stay and play on a server ... "contains custom rules" is really easy and good enough ...

Mailaender:
> > I'd prefer if this would have a notation like what Armament got, by having a TargetModifier and a ForceTargetModifier. This way seems arbitrary atm.

> I agree (except for naming: `TargetStances` and `ForceTargetStances` like Armament would be better).  We can also set up infiltration/repair/etc the same way.

Updated as suggested.

OpenRA.Mods.Common/Traits/Capturable.cs
```diff
@@ -32,16 +33,6 @@ public bool CanBeTargetedBy(Actor captor, Player owner)
 			if (c == null)
 				return false;
 
-			var playerRelationship = owner.Stances[captor.Owner];
```
pchote: You need to check the stances here too (and likewise for the other traits).  The order targeters are client side code, so the checks there could be modified or removed by somebody malicious.

OpenRA.Mods.Common/Traits/Capturable.cs
```diff
@@ -17,9 +17,10 @@ public class CapturableInfo : ITraitInfo
 	{
 		[Desc("Type listed under Types in Captures: trait of actors that can capture this).")]
 		public readonly string Type = "building";
-		public readonly bool AllowAllies = false;
-		public readonly bool AllowNeutral = true;
-		public readonly bool AllowEnemies = true;
+
+		public readonly Stance TargetStances = Stance.Enemy | Stance.Neutral;
+		public readonly Stance ForceTargetStances = Stance.Enemy | Stance.Neutral | Stance.Ally;
```
pchote: This sneaks in enabling for allies by default.

OpenRA.Mods.Common/Traits/ExternalCapturable.cs
```diff
@@ -17,9 +17,10 @@ public class ExternalCapturableInfo : ITraitInfo
 	{
 		[Desc("Type of actor (the ExternalCaptures: trait defines what Types it can capture).")]
 		public readonly string Type = "building";
-		public readonly bool AllowAllies = false;
-		public readonly bool AllowNeutral = true;
-		public readonly bool AllowEnemies = true;
+
+		public readonly Stance TargetStances = Stance.Enemy | Stance.Neutral;
+		public readonly Stance ForceTargetStances = Stance.Enemy | Stance.Neutral | Stance.Ally;
```
pchote: Likewise here.

OpenRA.Mods.Common/Traits/Capturable.cs
Mailaender: I added the checks back (below).

Mailaender:
Updated.

OpenRA.Mods.Common/Traits/Capturable.cs
```diff
@@ -33,17 +34,11 @@ public bool CanBeTargetedBy(Actor captor, Player owner)
...
-
-			if (playerRelationship == Stance.Enemy && !AllowEnemies)
-				return false;
-
-			if (playerRelationship == Stance.Neutral && !AllowNeutral)
+			if (!c.CaptureTypes.Contains(Type))
 				return false;
 
-			if (!c.CaptureTypes.Contains(Type))
+			var stances = ForceTargetStances | TargetStances;
```
GraionDilach: This worries me a bit - in teamgame, this allow a player to capture a freshly-captured-by-his-ally actor just by arriving late.

OpenRA.Mods.Common/Traits/ExternalCapturable.cs
```diff
@@ -33,17 +34,11 @@ public bool CanBeTargetedBy(Actor captor, Player owner)
...
-
-			if (playerRelationship == Stance.Enemy && !AllowEnemies)
-				return false;
-
-			if (playerRelationship == Stance.Neutral && !AllowNeutral)
+			if (!c.CaptureTypes.Contains(Type))
 				return false;
 
-			if (!c.CaptureTypes.Contains(Type))
+			var stances = ForceTargetStances | TargetStances;
```
GraionDilach: Same here.

GraionDilach:
Could these be changed via including the force modifier passed an argument then checking for it explicitly?

:+1: otherwise.

Mailaender:
I don't have access to the force modifier in those function, otherwise I would have mirrored the checks.

pchote:
You can/should do what the attack code does.

Mailaender:
I already checked. It is called in other contexts as well where those modifiers simply aren't available.

pchote:
The two contexts I am aware of are the order targeting and ResolveOrder.  The modifiers are available in both.

GraionDilach:
The `CaptureActor` activity also uses it inside the `CanReserve` call which sounds like the denying context - although I'm not certain what that supposed to check for there.

pchote:
That activity will be called via ResolveOrder, so there should be no problem.

pchote:
However it might pay to try that in a separate commit - if the plumbing ends up being horrible then the current solution might be acceptable if we change Attack to use similar logic.

OpenRA.Mods.Common/Traits/Capturable.cs
pchote: The target generation increment that happens on owner change should stop that.

Mailaender:
> That activity will be called via ResolveOrder, so there should be no problem.

Sorry, I can't find any `TargetModifier` near `ResolveOrder` in `Captures`.

pchote:
While i'm no longer convinced that this is the right approach (see my last comment), my original comment on the topic said:

> You can/should do what the attack code does, which gives you access to the force modifier in those functions.

Please don't remove the changes requested tag until the discussion prompting it is resolved.  It was added in this case because of the inconsistency this introduces compared with AttackBase.

OpenRA.Mods.Common/Traits/EngineerRepair.cs
```diff
@@ -98,40 +101,6 @@ public void ResolveOrder(Actor self, Order order)
 			self.SetTargetLine(target, Color.Yellow);
 			self.QueueActivity(new RepairBuilding(self, target.Actor, info.EnterBehaviour));
 		}
-
-		class EngineerRepairOrderTargeter : UnitOrderTargeter
```
pchote: Like discussed in other PRs please keep these as internal child classes.

Mailaender:
Updated and experimented a bit. Exposing `TargetModifiers` into activities won't work and doesn't seem like something we want. I have no idea what to change in `AttackBase`.

abcdefg30:
So, what is the status here?

Mailaender:
I don't know how to fulfill the requested changes. :(

reaperrr:
Closing for now due to lack of progress, to clean up the PR queue.

PR #10781, opened 2016-02-19
Fix filesystem references under Windows.

pchote:
Fixes scripted maps, and the TS mod crashes under Windows.

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -1178,7 +1178,8 @@ namespace OpenRA
 		// Placeholders for future implementation
 		public Stream Open(string filename)
 		{
-			if (Container.Contains(filename))
+			// Explicit package paths never refer to a map
+			if (!filename.Contains("|") && Container.Contains(filename))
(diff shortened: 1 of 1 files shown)


reaperrr:
Fixes shellmaps and I didn't experience any crash in TS, either (Win7x64).

:+1:

OpenRA.Game/Map/Map.cs
```diff
@@ -1178,7 +1178,8 @@ public IEnumerable<CPos> FindTilesInCircle(CPos center, int maxRange, bool allow
 		// Placeholders for future implementation
 		public Stream Open(string filename)
 		{
-			if (Container.Contains(filename))
+			// Explicit paths never refer to a map
+			if (filename.IndexOf('|') == -1 && Container.Contains(filename))
```
Mailaender: `filename.IndexOf('|') == -1` translates to `!filename.Contains('I')`?

pchote: Yeah, I guess that would be clearer.

pchote:
Updated.

PR #10786, opened 2016-02-20
Remove internal use of map paths.

pchote:
Some important refactoring in preparation for the next round of filesystem/map cleanups.  This replaces the string file paths everywhere with a reference to a map package (which are stored/owned by a MapPreview, and exist for the lifetime of the ModData).

This is the last major set of plumbing work to support oramods.  The remaining steps are either localized cleanups (in the map enumeration and editor code), or pure additions (adding support for mounting sub-directories inside a zip file).

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -302,10 +302,6 @@ namespace OpenRA
 			PostInit();
 		}
 
-		/// <summary>Initializes a map loaded from disk.</summary>
-		public Map(string path)
-			: this(Game.ModData, Game.ModData.ModFiles.OpenPackage(path)) { }
-
 		public Map(ModData modData, IReadOnlyPackage package)
 		{
 			this.modData = modData;
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -120,7 +120,6 @@ namespace OpenRA
 		public readonly SubCell LastSubCell;
 
 		public IReadOnlyPackage Package { get; private set; }
-		public string Path { get { return Package.Name; } }
 
 		// Yaml map data
 		public string Uid { get; private set; }
@@ -492,11 +491,6 @@ namespace OpenRA
 		public Ruleset PreloadRules()
 		{
 			return rules.Value;
-		}
-
-		public void Save(string toPath)
-		{
-			Save(modData.ModFiles.CreatePackage(toPath, new Dictionary<string, byte[]>()));
 		}
 
 		public void Save(IReadWritePackage toContainer)
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -357,12 +357,6 @@ namespace OpenRA
 
 			PostInit();
 
-			// The Uid is calculated from the data on-disk, so
-			// format changes must be flushed to disk.
-			// TODO: this isn't very nice
-			if (MapFormat < 8)
-				Save(path);
-
 			Uid = ComputeUID(Container);
 		}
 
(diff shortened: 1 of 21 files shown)


OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -60,7 +60,7 @@ public virtual void Run(ModData modData, string[] args)
 				ValidateMapFormat(format);
 
 				var tileset = GetTileset(mapSection);
-				Map = new Map(Rules.TileSets[tileset], MapSize, MapSize)
+				Map = new Map(Game.ModData, Rules.TileSets[tileset], MapSize, MapSize)
```
Mailaender: Why not `ModData` or `modData` here?

Mailaender:
I noticed that saving a map and playing it in skirmish right away leads to "map can't be found on the resource server", but that is resolved by re-selecting it from the change map menu.

Otherwise I can't spot regressions when saving maps in the editor and resizing maps from the command line or loading custom maps and mini games which add their own artwork. :white_check_mark:

pchote:
> I noticed that saving a map and playing it in skirmish right away leads to "map can't be found on the resource server", but that is resolved by re-selecting it from the change map menu.

I can't trigger this either when creating a new map or when modifying an existing one.  Can you provide step by step instructions to reproduce?

pchote:
Ok, reproduced.  This happens when you edit and save over the map that was previously selected in skirmish.  This is not a new regression: bleed behaves the same, but oddly not release.  I suggest filing an issue.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
pchote: Because that got lost in rebase gymnastics :)  Fixed.

abcdefg30:
```
OpenRA.Game/Map/Map.cs:L116: [SA1307] Public and internal fields must start with an upper-case letter: modData.
OpenRA.Game/Map/Map.cs:L116: [SA1304] Readonly variables that are not declared private must start with an upper-case letter: modData.

**Comment by pchote** [comment]
Updated.

File: OpenRA.Game/Map/Map.cs [hunk 2]
```diff
@@ -494,7 +489,7 @@ public Ruleset PreloadRules()
 			return rules.Value;
 		}
 
-		public void Save(string toPath)
+		public void Save(IReadWritePackage toContainer)
 		{
 			MapFormat = 8;
```
obrakmann: Not related to this PR at all, but could we make this `MapFormat = SupportedMapFormat`, please?

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -3601,94 +3602,91 @@ internal static void UpgradeActors(int engineVersion, ref List<MiniYamlNode> nod
...
+			entries.Add("map.yaml", Encoding.UTF8.GetBytes(yaml.Nodes.WriteToString()));
+			foreach (var file in package.Contents)
+			{
+				if (file == "map.yaml")
+					continue;
 
-					entries.Add(file, package.GetStream(file).ReadAllBytes());
-				}
+				entries.Add(file, package.GetStream(file).ReadAllBytes());
 			}
```
obrakmann: I think there's a `package.Write(entries)` missing here as well.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c72403dfcee9a3e220625e1091c6f4686f35caa1)

Issue #10515, opened 2016-01-14
Reassign admin for game in progress upon host disconnect


Mailaender:
At the moment the whole game will just freeze. Also needs to make sure that bots get handed over. https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Server/Server.cs#L585

pchote:
The game freezes when the admin of the local server disconnects because they take the server down with them.  This doesn't happen on dedicated servers (I just tested it), where the admin is just a normal client that has an extra bit set in their client data.

Mailaender:
I wonder how we will reassign admin for game in progress via an order if the server doesn't exist anymore.

pchote:
On dedicated servers the admin is just a normal client, and the server hasn't gone anywhere.

PR #10787, opened 2016-02-20
Fixed game end when admin leaves on dedicated server

Mailaender:
Instead a new admin is found and all bots are handed over. You can use `launch-dedicated.sh` and `launch-game.sh` to quickly setup a test environment with 1 dedicated server and 2 clients.

Closes https://github.com/OpenRA/OpenRA/issues/10515.

### DIFF
--- a/OpenRA.Game/Network/Order.cs
+++ b/OpenRA.Game/Network/Order.cs
@@ -167,6 +167,11 @@ namespace OpenRA
 			return new Order("CancelProduction", subject, false) { ExtraData = (uint)count, TargetString = item };
 		}
 
+		public static Order NewAdmin(int nextAdminIndex)
+		{
+			return new Order("NewAdmin", null, false) { ExtraData = (uint)nextAdminIndex };
+		}
+
 		// For scripting special powers
 		public Order()
 			: this(null, null, null, CPos.Zero, null, false, CPos.Zero, 0) { }
--- a/OpenRA.Game/Network/UnitOrders.cs
+++ b/OpenRA.Game/Network/UnitOrders.cs
@@ -67,6 +67,37 @@ namespace OpenRA.Network
 						var client = orderManager.LobbyInfo.ClientWithIndex(clientId);
 						if (client != null)
 							client.State = Session.ClientState.Disconnected;
+
+						break;
+					}
+
+				case "NewAdmin": // Find a new admin and hand bots over mid-game.
+					{
+						var client = orderManager.LobbyInfo.ClientWithIndex(clientId);
+						if (client == null || !client.IsAdmin)
+							break;
+
+						var nextAdmin = orderManager.LobbyInfo.Clients.FirstOrDefault(c => c.Index == order.ExtraData);
+						if (nextAdmin != null)
+						{
+							nextAdmin.IsAdmin = true;
+							Game.AddChatLine(Color.White, "Server", "{0} is now the admin.".F(nextAdmin.Name));
+
+							var bots = orderManager.LobbyInfo.Clients.Where(c => c.Bot != null && c.BotControllerClientIndex == client.Index);
+							foreach (var bot in bots)
+							{
+								bot.BotControllerClientIndex = nextAdmin.Index;
+								var botPlayer = orderManager.World.FindPlayerByClient(bot);
+								botPlayer.ActivateBot();
+								Log.Write("server", "Bot {0} is now controlled by {1}.".F(bot.Name, nextAdmin.Name));
+							}
+						}
+						else
+						{
+							Log.Write("server", "Suggested new game admin {0} not found.".F(order.TargetString));
+							orderManager.LobbyInfo.Clients.RemoveAll(c => c.Bot != null && c.BotControllerClientIndex == client.Index);
+						}
+
 						break;
 					}
 
--- a/OpenRA.Game/Player.cs
+++ b/OpenRA.Game/Player.cs
@@ -138,19 +138,28 @@ namespace OpenRA
 
 			// Enable the bot logic on the host
 			IsBot = BotType != null;
-			if (IsBot && Game.IsHost)
-			{
-				var logic = PlayerActor.TraitsImplementing<IBot>().FirstOrDefault(b => b.Info.Type == BotType);
-				if (logic == null)
-					Log.Write("debug", "Invalid bot type: {0}", BotType);
-				else
-					logic.Activate(this);
-			}
+			if (IsBot)
+				ActivateBot();
 
 			stanceColors.Self = ChromeMetrics.Get<Color>("PlayerStanceColorSelf");
 			stanceColors.Allies = ChromeMetrics.Get<Color>("PlayerStanceColorAllies");
 			stanceColors.Enemies = ChromeMetrics.Get<Color>("PlayerStanceColorEnemies");
 			stanceColors.Neutrals = ChromeMetrics.Get<Color>("PlayerStanceColorNeutrals");
+		}
+
+		public void ActivateBot()
+		{
+			if (string.IsNullOrEmpty(BotType) || !Game.IsHost)
+				return;
+
+			var logic = PlayerActor.TraitsImplementing<IBot>().FirstOrDefault(b => b.Info.Type == BotType);
+			if (logic == null)
+				Log.Write("debug", "Invalid bot type: {0}", BotType);
+			else
+			{
+				logic.Activate(this);
+				Log.Write("debug", "Activated bot {0} on player {1}".F(BotType, InternalName));
+			}
 		}
 
 		public override string ToString()
--- a/OpenRA.Game/Server/Server.cs
+++ b/OpenRA.Game/Server/Server.cs
@@ -559,24 +559,32 @@ namespace OpenRA.Server
 				// Send disconnected order, even if still in the lobby
 				DispatchOrdersToClients(toDrop, 0, new ServerOrder("Disconnected", "").Serialize());
 
-				LobbyInfo.Clients.RemoveAll(c => c.Index == toDrop.PlayerIndex);
-				LobbyInfo.ClientPings.RemoveAll(p => p.Index == toDrop.PlayerIndex);
-
-				// Client was the server admin
-				// TODO: Reassign admin for game in progress via an order
-				if (Dedicated && dropClient.IsAdmin && State == ServerState.WaitingPlayers)
+				// Client was the server admin, but the game isn't over yet.
+				if (Dedicated && dropClient.IsAdmin)
 				{
-					// Remove any bots controlled by the admin
-					LobbyInfo.Clients.RemoveAll(c => c.Bot != null && c.BotControllerClientIndex == toDrop.PlayerIndex);
+					// Pick the human player who joined right after the previous admin.
+					var nextAdmin = LobbyInfo.Clients.Where(c => c.Bot == null && c != dropClient).MinByOrDefault(c => c.Index);
 
-					var nextAdmin = LobbyInfo.Clients.Where(c1 => c1.Bot == null)
-						.MinByOrDefault(c => c.Index);
-
-					if (nextAdmin != null)
+					// Still in the lobby so handle things directly.
+					if (State == ServerState.WaitingPlayers)
 					{
-						nextAdmin.IsAdmin = true;
-						SendMessage("{0} is now the admin.".F(nextAdmin.Name));
+						// Remove the previous admin.
+						LobbyInfo.Clients.RemoveAll(c => c.Index == toDrop.PlayerIndex);
+
+						// Stop pinging when a player drops from the server.
+						LobbyInfo.ClientPings.RemoveAll(p => p.Index == toDrop.PlayerIndex);
+
+						// Remove any bots controlled by the previous admin.
+						LobbyInfo.Clients.RemoveAll(c => c.Bot != null && c.BotControllerClientIndex == toDrop.PlayerIndex);
+
+						if (nextAdmin != null)
+						{
+							nextAdmin.IsAdmin = true;
+							SendMessage("{0} is now the admin.".F(nextAdmin.Name));
+						}
 					}
+					else if (nextAdmin != null)
+						DispatchOrdersToClients(toDrop, 0, Order.NewAdmin(nextAdmin.Index).Serialize());
 				}
 
 				DispatchOrders(toDrop, toDrop.MostRecentFrame, new byte[] { 0xbf });


Mailaender:
Rebased.

obrakmann:
This seems to work fine in-game. One thing I noticed is that the replay is crashing the game. Not sure if that is caused by any of these changes, though.

OpenRA.Game/Player.cs
```diff
@@ -130,20 +130,27 @@ public Player(World world, Session.Client client, PlayerReference pr)
...
-			if (IsBot && Game.IsHost)
+			ActivateBotOnHost(botType);
+
+			stanceColors.Self = ChromeMetrics.Get<Color>("PlayerStanceColorSelf");
+			stanceColors.Allies = ChromeMetrics.Get<Color>("PlayerStanceColorAllies");
+			stanceColors.Enemies = ChromeMetrics.Get<Color>("PlayerStanceColorEnemies");
+			stanceColors.Neutrals = ChromeMetrics.Get<Color>("PlayerStanceColorNeutrals");
+		}
+
+		public void ActivateBotOnHost(string botType)
```
pchote: The `OnHost` in the name here is a bit confusing – based on the name I would expect you to be passing the host as a parameter.  `ActivateBot` is fine.

OpenRA.Game/Player.cs
```diff
@@ -130,20 +130,27 @@ public Player(World world, Session.Client client, PlayerReference pr)
...
+		{
+			if (!string.IsNullOrEmpty(botType) && Game.IsHost)
 			{
 				var logic = PlayerActor.TraitsImplementing<IBot>().FirstOrDefault(b => b.Info.Name == botType);
 				if (logic == null)
 					Log.Write("debug", "Invalid bot type: {0}", botType);
 				else
+				{
 					logic.Activate(this);
+					Game.Debug("Activated bot {0}".F(botType));
```
pchote: This looks like leftover debug code.

OpenRA.Game/Server/Server.cs
```diff
@@ -586,23 +586,28 @@ public void DropClient(Connection toDrop, int frame)
...
-					var nextAdmin = LobbyInfo.Clients.Where(c1 => c1.Bot == null)
-						.MinByOrDefault(c => c.Index);
+						// Remove any bots controlled by the previous admin.
+						LobbyInfo.Clients.RemoveAll(c => c.Bot != null && c.BotControllerClientIndex == toDrop.PlayerIndex);
 
-					if (nextAdmin != null)
-					{
-						nextAdmin.IsAdmin = true;
-						SendMessage("{0} is now the admin.".F(nextAdmin.Name));
+						// Find a new human client for adminship.
```
pchote: `// Find a new human admin`.

pchote:
Confirmed the crash when trying to load a replay where bots have been switched:
```
Exception of type `System.IO.EndOfStreamException`: Attempted to read past the end of the stream.
  at OpenRA.StreamExts.ReadBytes (System.IO.Stream s, System.Byte[] buffer, Int32 offset, Int32 count) <0xa1b6298 + 0x000bf> in <filename unknown>:0 
  at OpenRA.StreamExts.ReadBytes (System.IO.Stream s, Int32 count) <0xa1b61e0 + 0x00043> in <filename unknown>:0 
  at OpenRA.Network.ReplayConnection..ctor (System.String replayFilename) <0x1cada098 + 0x002cf> in <filename unknown>:0 
  at OpenRA.Game.JoinReplay (System.String replayFile) <0x1cada000 + 0x0002f> in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.Logic.ReplayBrowserLogic.<WatchReplay>m__19 () <0x1cad9fc0 + 0x0001f> in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.Logic.ReplayBrowserLogic.WatchReplay () <0x1cad9d28 + 0x00098> in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.Logic.ReplayBrowserLogic+<ReplayBrowserLogic>c__AnonStorey0.<>m__3 () <0x1cad9cf8 + 0x00013> in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.ButtonWidget.<ButtonWidget>m__A (MouseInput _) <0x19a38b28 + 0x00017> in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.ButtonWidget.HandleMouseInput (MouseInput mi) <0x19133120 + 0x00293> in <filename unknown>:0 
  at OpenRA.Widgets.Widget.HandleMouseInputOuter (MouseInput mi) <0x19132820 + 0x00264> in <filename unknown>:0 
  at OpenRA.Widgets.Ui.HandleInput (MouseInput mi) <0x19132668 + 0x0009f> in <filename unknown>:0 
  at OpenRA.DefaultInputHandler+<OnMouseInput>c__AnonStorey2.<>m__0 () <0x191325e8 + 0x0006b> in <filename unknown>:0 
  at OpenRA.Sync.CheckSyncUnchanged[T] (OpenRA.World world, System.Func`1 fn) <0xc0b30d0 + 0x00074> in <filename unknown>:0 
  at OpenRA.DefaultInputHandler.OnMouseInput (MouseInput input) <0x19132508 + 0x000cf> in <filename unknown>:0 
  at OpenRA.Platforms.Default.Sdl2Input.PumpInput (IInputHandler inputHandler) <0xaa65ab0 + 0x00886> in <filename unknown>:0 
  at OpenRA.Platforms.Default.Sdl2GraphicsDevice.PumpInput (IInputHandler inputHandler) <0xaa65a70 + 0x0002b> in <filename unknown>:0 
  at OpenRA.Renderer.EndFrame (IInputHandler inputHandler) <0xaa65a00 + 0x0002d> in <filename unknown>:0 
  at OpenRA.Game.RenderTick () <0xc0f37d0 + 0x0031f> in <filename unknown>:0 
  at OpenRA.Game.Loop () <0xc0b13e8 + 0x0016f> in <filename unknown>:0 
  at OpenRA.Game.Run () <0xc0b12a8 + 0x0006b> in <filename unknown>:0 
  at OpenRA.Program.Run (System.String[] args) <0x760908 + 0x00053> in <filename unknown>:0 
  at OpenRA.Program.Main (System.String[] args) <0x75fee0 + 0x000d3> in <filename unknown>:0 
```

pchote:
I am also seeing sporadic desyncs when the admin drops. I've so far only reproduced it when one of the clients is a spectator, if that helps.  I don't have a sync diff because the sync reports from multiple clients on one machine overwrite eachother.

OpenRA.Game/Player.cs
Mailaender: The player should be informed about this.

OpenRA.Game/Server/Server.cs
Mailaender: That removes almost all vital information from the comment, but it is redundant anyway.

Mailaender:
`EndOfStreamException` sounds like your two local clients are both writing on the same replay file when one is doing abrupt exits. Not sure if it is related to this change. Looking at `ReplayConnection` I also have no idea how to fix it.

pchote:
#10981 will be useful for debugging that and the desyncs.

Adding changes requested back until the root cause is identified and either fixed or excluded as an unrelated bug

OpenRA.Game/Player.cs
```diff
@@ -130,20 +130,27 @@ public Player(World world, Session.Client client, PlayerReference pr)
...
+		{
+			if (!string.IsNullOrEmpty(botType) && Game.IsHost)
 			{
 				var logic = PlayerActor.TraitsImplementing<IBot>().FirstOrDefault(b => b.Info.Name == botType);
 				if (logic == null)
 					Log.Write("debug", "Invalid bot type: {0}", botType);
 				else
+				{
 					logic.Activate(this);
+					Log.Write("debug", "Activated bot {0}".F(botType));
```
pchote: This causes the player chat to be spammed at the start of the game.  We've had bots for years, and nobody has ever said that they wanted or needed to know this (since it has always been obvious).

Players are already informed by other log messages in the admin-drop case.

obrakmann:
It seems that the replay bug was indeed caused by two clients writing to the same file. I couldn't reproduce it with clients on separate computers. Remove changes requested, adding fixup requested for the bot messages.

OpenRA.Game/Network/UnitOrders.cs
```diff
@@ -65,9 +65,40 @@ public static void ProcessOrder(OrderManager orderManager, World world, int clie
...
+								nextAdmin.IsAdmin = true;
+								Game.AddChatLine(Color.White, "Server", "{0} is now the admin.".F(nextAdmin.Name));
+
+								var bots = orderManager.LobbyInfo.Clients.Where(c => c.Bot != null && c.BotControllerClientIndex == client.Index);
+								foreach (var bot in bots)
+								{
+									bot.BotControllerClientIndex = nextAdmin.Index;
+									var botPlayer = orderManager.World.FindPlayerByClient(bot);
+									botPlayer.ActivateBot(bot.Bot);
+									Game.Debug("Bot {0} is now controlled by {1}.".F(bot.Name, nextAdmin.Name));
```
obrakmann: I'd move this to debug.log as well. Players probably don't care about these implementation details.

obrakmann:
Oh, also my :+1:

OpenRA.Game/Player.cs
Mailaender: Log messages spam the player chat?

pchote: Ah, nevermind.  Your comment on the last thread relating to this made me think that you didn't want to change it, so I didn't catch that you moved from `Game.Debug` to `Log.Write`.

Mailaender:
Removed the debug spam.

Mailaender:
Rebased.

OpenRA.Game/Server/Server.cs
```diff
@@ -584,23 +584,28 @@ public void DropClient(Connection toDrop, int frame)
...
+						// Find a new admin.
+						var nextAdmin = LobbyInfo.Clients.Where(c1 => c1.Bot == null).MinByOrDefault(c => c.Index);
+						if (nextAdmin != null)
+						{
+							nextAdmin.IsAdmin = true;
+							SendMessage("{0} is now the admin.".F(nextAdmin.Name));
+						}
 					}
+					else
+						DispatchOrdersToClients(toDrop, 0, new ServerOrder("NewAdmin", "").Serialize());
```
pchote: Sending this with a `frame` of 0 isn't going to work, and may be the cause of the desync that I reported earlier.  I don't have the time or the attention span to dig any deeper into this today.

pchote: To elaborate: Orders with `frame == 0` are interpreted as immediate orders, which bypass our usual world sync mechanism and run immediately.  If two clients are simulating different ticks when the order arrives (due to network or game lag) then the bots will be seen to switch on different ticks, and the game state will desync.

pchote: `ServerOrder`s are also always processed as an immediate order (they Serialize with a signature byte 0xfe, which is then deserialized to an order with `IsImmediate = true` in `Order.Deserialize`.

Mailaender: Whenever I send this with a frame (frame + 1, frame - 1, frame + 2) whatever doesn't matter the game will crash, because the frame already contains an order (probably the drop/disconnect one).

image

 If I issue an order through `Game.OrderManager` the dedicated server will crash, as OrderManager is probably null there.

I have no idea how to solve this.

WolfGaming:
Does this crash a dedicated server if all human players leave or does the game just end?

Mailaender:
I tried to address https://github.com/OpenRA/OpenRA/pull/10787#discussion_r59982236 and it made it worse, so yes at the moment it will crash.

pchote:
This does not appear to be fixable in the foreseeable future, so closing as part of our much-needed PR queue cleanup.

Mailaender:
I rebased this after https://github.com/OpenRA/OpenRA/issues/12677 promised that the problems will be gone and also fixed some oversights as the latest revision where I wildly experimented with different send order strategies would pick the original admin as new admin. This now works stable in a local test session in a dedicated server scenario with and without bots.

OpenRA.Game/Network/UnitOrders.cs
```diff
@@ -65,9 +65,43 @@ public static void ProcessOrder(OrderManager orderManager, World world, int clie
 						var client = orderManager.LobbyInfo.ClientWithIndex(clientId);
 						if (client != null)
 							client.State = Session.ClientState.Disconnected;
+
 						break;
 					}
 
+				case "NewAdmin": /* Find a new admin and hand bots over mid-game. */
```
abcdefg30: Comments on the line above, please. (Also `//`)

OpenRA.Game/Network/UnitOrders.cs
```diff
@@ -65,9 +65,43 @@ public static void ProcessOrder(OrderManager orderManager, World world, int clie
...
+					{
+						var client = orderManager.LobbyInfo.ClientWithIndex(clientId);
+						if (client == null)
+							break;
+
+						if (client.IsAdmin)
```
abcdefg30: You could move this up into the `null` check above, return early and save one level of indentation:
```
if (client == null || !client.IsAdmin)
	break;
```

OpenRA.Game/Server/Server.cs
Mailaender: Both
```csharp
DispatchOrders(toDrop, toDrop.MostRecentFrame, Order.NewAdmin(nextAdmin.Name).Serialize());
```

```csharp
DispatchOrdersToClients(toDrop, toDrop.MostRecentFrame, Order.NewAdmin(nextAdmin.Name).Serialize());
```

cause
```
Exception of type `System.ArgumentException`: An item with the same key has already been added.
  at System.ThrowHelper.ThrowArgumentException (System.ExceptionResource resource) [0x0000b] in <8f2c484307284b51944a1a13a14c0266>:0 
  at System.Collections.Generic.Dictionary`2[TKey,TValue].Insert (TKey key, TValue value, System.Boolean add) [0x0008e] in <8f2c484307284b51944a1a13a14c0266>:0 
  at System.Collections.Generic.Dictionary`2[TKey,TValue].Add (TKey key, TValue value) [0x00000] in <8f2c484307284b51944a1a13a14c0266>:0 
  at OpenRA.Network.FrameData.AddFrameOrders (System.Int32 clientId, System.Int32 frame, System.Byte[] orders) [0x0000e] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Network.OrderManager+<TickImmediate>c__AnonStorey0.<>m__0 (System.Int32 clientId, System.Byte[] packet) [0x00077] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Network.EchoConnection.Receive (System.Action`2[T1,T2] packetFn) [0x00047] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Network.OrderManager.TickImmediate () [0x000b4] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Sync+<CheckSyncUnchanged>c__AnonStorey0.<>m__0 () [0x00001] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Sync.CheckSyncUnchanged[T] (OpenRA.World world, System.Func`1[TResult] fn) [0x00043] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Sync.CheckSyncUnchanged (OpenRA.World world, System.Action fn) [0x0000e] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Game.InnerLogicTick (OpenRA.Network.OrderManager orderManager) [0x0011a] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Game.LogicTick () [0x00050] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Game.Loop () [0x000d6] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Game.Run () [0x00042] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Program.Run (System.String[] args) [0x00011] in <67c5022a3d434b6a913576734b344516>:0 
  at OpenRA.Program.Main (System.String[] args) [0x0004f] in <67c5022a3d434b6a913576734b344516>:0 
```

on the remaining clients. This doesn't happen when dispatching with frame `0`.

OpenRA.Game/Network/UnitOrders.cs
Mailaender: Updated accordingly.

pchote:
Unfortunately there doesn't seem to be much interest in anyone else picking this up, so i'm moving back to the Future milestone.

Mailaender:
This was closed prematurely. It should work as is with no one requesting more changes. Rebased.

RoosterDragon (dismissed):
Tested this:
- Dropping in the lobby - admins are switched fine.
- Dropping in a 2 player game with bots, admins switch fine, no desyncs, 2 replays are generated of the correct lengths, both play back without desyncs.
- Dropping in a 3 player game with bots, 1 of which is spec, then dropping spec later, admins switched fine, no desyncs, 3 replays are generated of the correct lengths and play back without desyncs.

Looks like all the previous review comments were addressed and this appears to be functional to me, so :+1:.

OpenRA.Game/Network/Order.cs
```diff
@@ -167,6 +167,11 @@ public static Order CancelProduction(Actor subject, string item, int count)
 			return new Order("CancelProduction", subject, false) { ExtraData = (uint)count, TargetString = item };
 		}
 
+		public static Order NewAdmin(string admin)
+		{
+			return new Order("NewAdmin", null, false) { IsImmediate = false, TargetString = admin };
```
pchote: Two easy to fix issues here:
* IsImmediate defaults to false, so doesn't need to be specified
* The admin client name isn't robust (it might change in the lobby, and is not unique), so please use the client ID instead.

Mailaender: Fixed.

RoosterDragon (dismissed):

abcdefg30 (dismissed):
Looks good to me otherwise. 👍

@RoosterDragon is your 👍 still valid?

OpenRA.Game/Server/Server.cs
```diff
@@ -559,24 +559,32 @@ public void DropClient(Connection toDrop)
...
+						LobbyInfo.Clients.RemoveAll(c => c.Bot != null && c.BotControllerClientIndex == toDrop.PlayerIndex);
+
+						if (nextAdmin != null)
+						{
+							nextAdmin.IsAdmin = true;
+							SendMessage("{0} is now the admin.".F(nextAdmin.Name));
+						}
 					}
+					else
+						DispatchOrdersToClients(toDrop, 0, Order.NewAdmin(nextAdmin.Index).Serialize());
```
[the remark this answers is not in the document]
Mailaender: Added the `null` check.

OpenRA.Game/Server/Server.cs
```diff
@@ -559,24 +559,32 @@ public void DropClient(Connection toDrop)
...
+						LobbyInfo.Clients.RemoveAll(c => c.Bot != null && c.BotControllerClientIndex == toDrop.PlayerIndex);
+
+						if (nextAdmin != null)
+						{
+							nextAdmin.IsAdmin = true;
+							SendMessage("{0} is now the admin.".F(nextAdmin.Name));
+						}
 					}
+					else if (nextAdmin != null)
+						DispatchOrdersToClients(toDrop, 0, Order.NewAdmin(nextAdmin.Index).Serialize());
```
RoosterDragon: Could you clarify something for me? Is the `orderManager.LobbyInfo.Clients.RemoveAll(c => c.Bot != null && c.BotControllerClientIndex == client.Index);` line that runs if no admin can be found when resolving the order important?

If so, should we not send the order (with some dummy index) anyway to cause it to run?
If not, why do we need to run it if the server found an admin and sent the order, but the client fails to resolve the index?

Just trying to straighten out the logic in my head here.

pchote: I believe that the only case where `nextAdmin` can be null is when there are no players left in the server.  In that case it doesn't matter what we do, the server will be shut down and recycled immediately after this.

pchote: Sending this order at frame 0 (i.e. immediate) is asking for desyncs.  All players need to process the frame at the same world tick!  `toDrop.MostRecentFrame` should give the correct result here I think.

pchote (changes requested):

penev92:
Couldn't get this to desync, but that might be because I'm using a local dedicated server + local clients.
One issue though: the bot transfer works only once. If you run 3 clients and leave with the admin, the second client does take over and the bots carry on. But if you leave with that one, the third client does not take over for the bots.

pchote:
Progress has stalled again here, and we don't have any immediate volunteers to take over.  Moving back to the future milestone, and hoping that it will only require one more cycle in the future to get it merged.

OpenRA.Game/Server/Server.cs
Mailaender: It crashes the game:

```
Exception of type `System.ArgumentException`: An item with the same key has already been added.
  at System.ThrowHelper.ThrowArgumentException (System.ExceptionResource resource) [0x0000b] in <dca3b561b8ad4f9fb10141d81b39ff45>:0 
  at System.Collections.Generic.Dictionary`2[TKey,TValue].Insert (TKey key, TValue value, System.Boolean add) [0x0008e] in <dca3b561b8ad4f9fb10141d81b39ff45>:0 
  at System.Collections.Generic.Dictionary`2[TKey,TValue].Add (TKey key, TValue value) [0x00000] in <dca3b561b8ad4f9fb10141d81b39ff45>:0 
  at OpenRA.Network.FrameData.AddFrameOrders (System.Int32 clientId, System.Int32 frame, System.Byte[] orders) [0x0000e] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Network.OrderManager+<TickImmediate>c__AnonStorey0.<>m__0 (System.Int32 clientId, System.Byte[] packet) [0x00077] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Network.EchoConnection.Receive (System.Action`2[T1,T2] packetFn) [0x00056] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Network.OrderManager.TickImmediate () [0x000b4] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Sync+<CheckSyncUnchanged>c__AnonStorey0.<>m__0 () [0x00001] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Sync.CheckSyncUnchanged[T] (OpenRA.World world, System.Func`1[TResult] fn) [0x00043] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Sync.CheckSyncUnchanged (OpenRA.World world, System.Action fn) [0x0000e] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Game.InnerLogicTick (OpenRA.Network.OrderManager orderManager) [0x0010a] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Game.LogicTick () [0x00050] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Game.Loop () [0x000d6] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Game.Run () [0x00042] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Program.Run (System.String[] args) [0x00011] in <f2be23e29023499d8549d9e0cebc473c>:0 
  at OpenRA.Program.Main (System.String[] args) [0x0004f] in <f2be23e29023499d8549d9e0cebc473c>:0
```

OpenRA.Game/Player.cs
```diff
@@ -138,21 +138,30 @@ public Player(World world, Session.Client client, PlayerReference pr)
...
+			if (IsBot)
+				ActivateBot(BotType);
 
 			stanceColors.Self = ChromeMetrics.Get<Color>("PlayerStanceColorSelf");
 			stanceColors.Allies = ChromeMetrics.Get<Color>("PlayerStanceColorAllies");
 			stanceColors.Enemies = ChromeMetrics.Get<Color>("PlayerStanceColorEnemies");
 			stanceColors.Neutrals = ChromeMetrics.Get<Color>("PlayerStanceColorNeutrals");
 		}
 
+		public void ActivateBot(string botType)
```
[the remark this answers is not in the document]
Mailaender: Yes.

pchote:
The probable desync issue above has not been addressed.

Mailaender:
I can't reproduce the probably desync on my end. Can you be more specific with your suggested fix?

pchote:
Not without investing a lot of time that I don't have digging into the networking code. Refer to #12677 which fixed the similar (and similarly hard to reproduce) desyncs when players dropped from a game.

pchote:
This is now at 8 months with no activity, so moving back to the Future milestone to clean up the PR queue.

Issue #10766, opened 2016-02-15
Red Alert 1 - Allies 05a crash on latest build


Biofreak1987:
Game crashes when the spy infiltrates the Soviet warfactory.

> Red Alert Mod at Version {DEV_VERSION}
on map 75b7c2d7819cccf06fb3dd857d84feddfd3ffa1f (05a: Tanya's Tale by Westwood Studios).
Operating System: Windows (Microsoft Windows NT 6.2.9200.0)
Runtime Version: .NET CLR 4.0.30319.42000
Exception of type `System.ArgumentException`: bounds must be non-empty.
Parameter name: bounds
   at OpenRA.Primitives.SpatiallyPartitioned`1.ValidateBounds(Rectangle bounds) in d:\Games\OpenRA\OpenRA.Game\Primitives\SpatiallyPartitioned.cs:line 36
   at OpenRA.Primitives.SpatiallyPartitioned`1.Add(T item, Rectangle bounds) in d:\Games\OpenRA\OpenRA.Game\Primitives\SpatiallyPartitioned.cs:line 41
   at OpenRA.Traits.ScreenMap.Add(Actor a) in d:\Games\OpenRA\OpenRA.Game\Traits\World\ScreenMap.cs:line 77
   at OpenRA.Mods.Common.Traits.Mobile.AddedToWorld(Actor self)
   at OpenRA.World.Add(Actor a) in d:\Games\OpenRA\OpenRA.Game\World.cs:line 245
   at OpenRA.Mods.Common.Scripting.ActorGlobal.<>c__DisplayClass3.<Create>b__1(World w)
   at OpenRA.World.Tick() in d:\Games\OpenRA\OpenRA.Game\World.cs:line 309
   at OpenRA.Game.InnerLogicTick(OrderManager orderManager) in d:\Games\OpenRA\OpenRA.Game\Game.cs:line 557
   at OpenRA.Game.LogicTick() in d:\Games\OpenRA\OpenRA.Game\Game.cs:line 582
   at OpenRA.Game.Loop() in d:\Games\OpenRA\OpenRA.Game\Game.cs:line 709
   at OpenRA.Game.Run() in d:\Games\OpenRA\OpenRA.Game\Game.cs:line 749
   at OpenRA.Program.Run(String[] args) in d:\Games\OpenRA\OpenRA.Game\Support\Program.cs:line 117
[...]

PR #10795, opened 2016-02-21
Fixed screen map updates for invisible mobile actors

Mailaender:
Some traits already kept that in mind, but not all of them. Fixes https://github.com/OpenRA/OpenRA/issues/10766 and related crashes.

### DIFF
--- a/OpenRA.Game/World.cs
+++ b/OpenRA.Game/World.cs
@@ -193,6 +193,35 @@ namespace OpenRA
 					player.Shroud.ExploreAll(this);
 		}
 
+		public void AddToMaps(Actor self, IOccupySpace ios)
+		{
+			ActorMap.AddInfluence(self, ios);
+			ActorMap.AddPosition(self, ios);
+
+			if (!self.Bounds.Size.IsEmpty)
+				ScreenMap.Add(self);
+		}
+
+		public void UpdateMaps(Actor self, IOccupySpace ios)
+		{
+			if (!self.IsInWorld)
+				return;
+
+			if (!self.Bounds.Size.IsEmpty)
+				ScreenMap.Update(self);
+
+			ActorMap.UpdatePosition(self, ios);
+		}
+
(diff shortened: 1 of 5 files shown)


RoosterDragon:
My bad for missing these originally - perhaps we should factor out some helpers methods for adds, updates and removes to fix this more fully.

Mailaender:
You mean we should do the boundary checks in https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Traits/World/ScreenMap.cs#L78?

RoosterDragon:
I was thinking of methods just to reduce the duplication of:
```c#
```

but moving the checks to ScreenMap is also a good idea and we should probably do that. Why not both?

Mailaender:
I summarized them in a new function. This will now also check everything against `IsInWorld` before doing actor influence changes.

OpenRA.Game/World.cs
```diff
@@ -193,6 +193,35 @@ internal World(Map map, OrderManager orderManager, WorldType type)
...
+			ActorMap.UpdatePosition(self, ios);
+		}
+
+		public void RemoveFromMaps(Actor self, IOccupySpace ios)
+		{
+			ActorMap.RemoveInfluence(self, ios);
+			ActorMap.RemovePosition(self, ios);
+
+			if (!self.Bounds.Size.IsEmpty)
+				self.World.ScreenMap.Remove(self);
```
RoosterDragon: Maybe just `ScreenMap.Remove(self);` for consistency.

RoosterDragon:
:+1:

OpenRA.Game/World.cs
Mailaender: Woops. Fixed.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/3aabb4e3b210d335773e1d26056c212c3bcf60c8)

PR #10801, opened 2016-02-21
Update license header for 2016 and GPLv3+.

### LINKED ISSUE #9626: Licensing: specify GPL version in headers and README.md
The licensing info in the README.md and in headers currently states: "It is made available to you under the terms of the GNU General Public License as published by the Free Software Foundation. For more information, see COPYING."

It does not specify which version of the GPL applies. When checking COPYING, one can see it is the GPL version 3, but there is no mention as to whether the "or any later version" clause applies. I would suggest to make it clearer at least in the README.md, and ideally also in the code header. If you go for GPLv3+ (3 or any later version), a typical license declaration would read:

```
# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27381353-licensing-specify-gpl-version-in-headers-and-readme-md?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #9626.

### DIFF
--- a/OpenRA.Game/Activities/Activity.cs
+++ b/OpenRA.Game/Activities/Activity.cs
@@ -3,8 +3,9 @@
  * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
  * This file is part of OpenRA, which is free software. It is made
  * available to you under the terms of the GNU General Public License
- * as published by the Free Software Foundation. For more information,
- * see COPYING.
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
  */
 #endregion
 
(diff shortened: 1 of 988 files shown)


Mailaender:
:+1:

pchote:
I also wouldn't be opposed to moving back to the standard header blob, even though that is a fair bit longer than our current one.

Mailaender:
Can you update README.md as well?

Phrohdoh:
I'm not a fan of the über verbose GPL header so I'd say stick with what we have as of this PR. :+1:

pchote:
Updated readme.

obrakmann:
The readme still states "Copyright 2007-2015 The OpenRA Developers". Either fix or remove the dates, please.

pchote:
Fixed.

PR #10802, opened 2016-02-21
Added exit codes to CheckSequenceSprites

Mailaender:
otherwise it will always exit with 0 indicating CI services that everything is fine.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/CheckSequenceSprites.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/CheckSequenceSprites.cs
@@ -28,7 +28,9 @@ namespace OpenRA.Mods.Common.UtilityCommands
 		{
 			// HACK: The engine code assumes that Game.modData is set.
 			Game.ModData = modData;
-			modData.SpriteSequenceLoader.OnMissingSpriteError = s => Console.WriteLine("\t" + s);
+
+			var failed = false;
+			modData.SpriteSequenceLoader.OnMissingSpriteError = s => { Console.WriteLine("\t" + s); failed = true; };
 
 			foreach (var t in modData.Manifest.TileSets)
 			{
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/UtilityCommands/CheckSequenceSprites.cs
```diff
@@ -23,12 +23,14 @@ public bool ValidateArguments(string[] args)
 			return true;
 		}
 
+		int exit;
```
pchote: This can be a local `var failed = false;`, and then choose the exit code at the very end.

Mailaender: What is the benefit of that?

pchote: A combination of minimizing scope and saying what you mean instead of using magic values.

Mailaender:
Updated.

RoosterDragon:
:+1:

OpenRA.Mods.Common/UtilityCommands/CheckSequenceSprites.cs
```diff
@@ -39,6 +41,8 @@ public void Run(ModData modData, string[] args)
 				foreach (var n in nodes)
 					modData.SpriteSequenceLoader.ParseSequences(modData, ts, sc, n);
 			}
+
+			Environment.Exit(failed ? 1 : 0);
```
Phrohdoh: `0` is the default exit when a program closes so isn't necessary here.
If you encounter an error just exit with `1`.

Mailaender: Updated.

pchote:
Works as advertised :+1:

PR #10803, opened 2016-02-21
Add ability to collect system information from users' systems

### LINKED ISSUE #10662: Use the battlefield news query to send system stats to the master server
Before we can fix #10661 we need an idea of how many players this will affect.
The battlefield news query runs once per game start, so it gives us a simple way to track active player statistics (unless the player explicitly opts out).

Implementation:

1. Add a `http://master.openra.net/gamenews` page, which can initially just do a http redirect to http://openra.net/gamenews.
2.  Change the `NewsUrl` to this new value.
2. Add an advanced setting for sending system statistics (enabled by default).
3. If the setting is enabled, send the current mod, `Platform.CurrentPlatform`, `Environment.OSVersion`, and `Platform.RuntimeVersion` (these are the same things we include in exception.log) as either GET or POST data along with the news query.
4. (can be left until after release) Change the gamenews page to start logging this information so that we look at aggregate OS/usage stats.

If we want to do anything about #10661 in the next year we really need to get this logging in the next release, so adding to the milestone.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30385358-use-the-battlefield-news-query-to-send-system-stats-to-the-master-server?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
Closes #10662.

Edit: Just to be sure: DON'T MERGE BEFORE THE REDIRECT ON master.openra.net IS SET UP!

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
@@ -220,7 +220,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				if (newsButton != null)
 				{
 					if (!fetchedNews)
-						new Download(Game.Settings.Game.NewsUrl, cacheFile, e => { },
+						new Download(Game.Settings.Game.NewsUrl + SysInfoQuery(), cacheFile, e => { },
 							(e, c) => NewsDownloadComplete(e, cacheFile, currentNews,
 							() => newsButton.AttachPanel(newsPanel)));
 
@@ -250,6 +250,22 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				"",
 				() => { Game.LoadEditor(uid); },
 				() => { Game.CloseServer(); SwitchMenu(MenuType.MapEditor); });
+		}
+
+		string SysInfoQuery()
+		{
+			if (!Game.Settings.Debug.SendSystemInformation)
+				return null;
+
+			return "?id={0}&platform={1}&os={2}&runtime={3}&lang={4}&version={5}&mod={6}&modversion={7}".F(
+				Uri.EscapeUriString(Game.Settings.Debug.UUID),
+				Uri.EscapeUriString(Platform.CurrentPlatform.ToString()),
+				Uri.EscapeUriString(Environment.OSVersion.ToString()),
+				Uri.EscapeUriString(Platform.RuntimeVersion),
+				Uri.EscapeUriString(System.Globalization.CultureInfo.InstalledUICulture.TwoLetterISOLanguageName),
+				Uri.EscapeUriString(ModMetadata.AllMods["modchooser"].Version),
+				Uri.EscapeUriString(Game.ModData.Manifest.Mod.Id),
+				Uri.EscapeUriString(Game.ModData.Manifest.Mod.Version));
 		}
 
 		void SetNewsStatus(string message)
--- a/OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
@@ -257,11 +257,12 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			if (!Game.Settings.Debug.SendSystemInformation)
 				return null;
 
-			return "?id={0}&platform={1}&os={2}&runtime={3}&lang={4}&version={5}&mod={6}&modversion={7}".F(
+			return "?id={0}&platform={1}&os={2}&runtime={3}&gl={4}&lang={5}&version={6}&mod={7}&modversion={8}".F(
 				Uri.EscapeUriString(Game.Settings.Debug.UUID),
 				Uri.EscapeUriString(Platform.CurrentPlatform.ToString()),
 				Uri.EscapeUriString(Environment.OSVersion.ToString()),
 				Uri.EscapeUriString(Platform.RuntimeVersion),
+				Uri.EscapeUriString(Game.Renderer.GLVersion),
 				Uri.EscapeUriString(System.Globalization.CultureInfo.InstalledUICulture.TwoLetterISOLanguageName),
 				Uri.EscapeUriString(ModMetadata.AllMods["modchooser"].Version),
 				Uri.EscapeUriString(Game.ModData.Manifest.Mod.Id),
--- a/mods/cnc/chrome/settings.yaml
+++ b/mods/cnc/chrome/settings.yaml
@@ -523,6 +523,13 @@ Container@SETTINGS_PANEL:
 							Height: 20
 							Font: Regular
 							Text: Show Performance Text
+						Checkbox@PERFGRAPH_CHECKBOX:
+							X: 15
+							Y: 100
+							Width: 300
+							Height: 20
+							Font: Regular
+							Text: Show Performance Graph
 						Checkbox@FETCH_NEWS_CHECKBOX:
 							X: 310
 							Y: 40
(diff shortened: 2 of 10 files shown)


obrakmann:
Added version of the 'all' mod to the query.

OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
```diff
@@ -251,6 +251,21 @@ void LoadMapIntoEditor(string uid)
...
+			if (!Game.Settings.Game.PlayerCensusParticipation)
+				return null;
+
+			return "?id={0}&platform={1}&os={2}&runtime={3}&lang={4}&version={5}&mod={6}".F(
+				Uri.EscapeUriString(Game.Settings.Game.UUID),
+				Uri.EscapeUriString(Platform.CurrentPlatform.ToString()),
+				Uri.EscapeUriString(Environment.OSVersion.ToString()),
+				Uri.EscapeUriString(Platform.RuntimeVersion),
+				Uri.EscapeUriString(System.Globalization.CultureInfo.InstalledUICulture.TwoLetterISOLanguageName),
+				Uri.EscapeUriString(ModMetadata.AllMods["all"].Version),
```
pchote: Can we log the current mod's ID and version instead?  IMO this is more useful than the "all" version and the mod title.

pchote: or we could keep this, but replace the mod title with ^

obrakmann: No, that's fine with me. I was actually looking for the id, but totally missed that field. Updated.

pchote:
As a starting point (so this can be merged) the redirect can be done by dropping a `gamenews` file on the master server with the following contents (may contain typos - it's been a long time since I last wrote php):
```php
<?php
header('Location: http://www.openra.net/gamenews');
?>
```

obrakmann:
Seperated game version and mod version.

mods/cnc/chrome/settings.yaml
```diff
@@ -508,43 +515,43 @@ Container@SETTINGS_PANEL:
...
 							Text: Fetch Community News
-						Checkbox@PERFGRAPH_CHECKBOX:
+						Checkbox@CENSUS_CHECKBOX:
 							X: 310
 							Y: 70
 							Width: 300
 							Height: 20
 							Font: Regular
-							Text: Show Performance Graph
+							Text: Participate in Player Census
```
Phrohdoh: Could we be a bit more straightforward and call this (and other necessary places)`Send System Information` instead?

obrakmann:
Changed the wording

pchote:
Can we pretty please also include `glGetString(GL_VERSION)` in this data?  We will be able to dramatically improve our rendering code if/when we can enforce GL >= 3

obrakmann:
Sure. You'll need to tell me how I should massacre the code to allow me to get at that, though. The OpenGL class is `internal static` and OpenRA.Mods.Common doesn't currently have access to the Platforms assemblies.

pchote:
I suggest exposing a `string OpenGLVersion` property on Game.Renderer.

obrakmann:
Sorry, but that doesn't help. I think it would be quicker if you sent a patch.

pchote:
ok

LipkeGu:
We should also give the User the ability to disable the collection, and inform them.

pchote:
I've set up the redirect, and will try and send you a patch for the GL version tonight.

mods/cnc/chrome/settings.yaml
```diff
@@ -508,43 +515,43 @@ Container@SETTINGS_PANEL:
...
 							Text: Fetch Community News
-						Checkbox@PERFGRAPH_CHECKBOX:
+						Checkbox@SENDSYSINFO_CHECKBOX:
 							X: 310
 							Y: 70
 							Width: 300
 							Height: 20
 							Font: Regular
-							Text: Show Performance Graph
+							Text: Send system information
```
pchote: System and information should be capitalized to match the other options.

mods/ra/chrome/settings.yaml
```diff
@@ -509,43 +516,43 @@ Background@SETTINGS_PANEL:
...
 					Text: Fetch Community News
-				Checkbox@PERFGRAPH_CHECKBOX:
+				Checkbox@SENDSYSINFO_CHECKBOX:
 					X: 310
 					Y: 70
 					Width: 300
 					Height: 20
 					Font: Regular
-					Text: Show Performance Graph
+					Text: Send system information
```
pchote: Here too

mods/cnc/chrome/settings.yaml
pchote: Can you put, in the same `Tiny` font that we use for the video desc label (probably needs to be split over two lines) something like: "Your OS, OpenGL version, and and .NET runtime version, and language settings are sent with an anonymous ID to help priorise future development"?

pchote:
https://github.com/obrakmann/OpenRA/pull/5

pchote:
Code looks good, and only a couple of minor UI suggestions.  This gets a deserved :+1: after fixing those and merging the GL patch.

Sample query URL: http://master.openra.net/gamenews?id=3fd3b08c-7933-47e9-a5b2-52e6bbceeb5b&platform=OSX&os=Unix%2015.3.0.0&runtime=Mono%204.2.2%20(explicit/996df3c%20Wed%20Jan%2020%2000:19:48%20EST%202016)%20CLR%204.0.30319.17020&gl=2.1%20INTEL-10.12.13lang=en&version=%7BDEV_VERSION%7D&mod=cnc&modversion=%7BDEV_VERSION%7D

which decodes to:
```
http://master.openra.net/gamenews
   ?id=3fd3b08c-7933-47e9-a5b2-52e6bbceeb5b&platform=OSX
   &os=Unix 15.3.0.0
   &runtime=Mono 4.2.2 (explicit/996df3c Wed Jan 20 00:19:48 EST 2016) CLR 4.0.30319.17020
   &gl=2.1 INTEL-10.12.13
   &lang=en
   &version={DEV_VERSION}
   &mod=cnc
   &modversion={DEV_VERSION}
```

obrakmann:
Updated.
screenshot from 2016-03-02 21 27 00

Phrohdoh:
Off topic but this just showcases our horrible font rendering. :(

abcdefg30:
:+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ad9f6bc6b72594194c9511b172c004d5457e2900)
I put it under `Engine / Modding` now (although @penev92 suggested to put it under `Spyware`).

Might be worth to mention that somewhere in the newspost, so that people who don't want to give their data don't rage later on (hence the `Spyware` above ^^)?

pchote:
I was already planning to mention it in the news, yeah.

mods/cnc/chrome/settings.yaml
```diff
@@ -530,43 +537,51 @@ Container@SETTINGS_PANEL:
...
-							Text: Show Performance Graph
+							Text: Send System Information
+						Label@SENDSYSINFO_DESC:
+							X: 310
+							Y: 85
+							Width: 250
+							Height: 30
+							Font: Tiny
+							WordWrap: True
+							Text: Your Operating System, OpenGL and .NET runtime versions, and language settings will be sent along with an anonymous ID to help priorise future development.
```
ScottNZ: `priorise` should have been `prioritize`

PR #10804, opened 2016-02-21
Fix broken dev commands

### LINKED ISSUE #10683: Tab completion isn't working for dev commands
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30500406-tab-completion-isn-t-working-for-dev-commands?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obrakmann:
Fixes #10683.

Not sure if that's really the way we should be going to solve this, but it does the job with minimal impact. A proper solution would likely require heavy surgery to the way we instantiate the widget tree.

I explained the problem itself in https://github.com/OpenRA/OpenRA/issues/10683#issuecomment-179478285:

> So all the *Commands do their thing in WorldLoaded. IngameChatLogic gathers all the commands in its constructor.
> 
> As it turns out, the widgets themselves are created during LoadWidgetAtGameStart's WorldLoaded method, which, as LoadWidgetAtGameStart is declared as one of the first traits in world.yaml, runs before the WorldLoaded method of the *Commands traits.

### DIFF
--- a/mods/cnc/rules/world.yaml
+++ b/mods/cnc/rules/world.yaml
@@ -47,8 +47,6 @@
 		PipColor: Blue
 		AllowedTerrainTypes: Clear,Road
 		AllowUnderActors: true
-	LoadWidgetAtGameStart:
-		ShellmapRoot: MENU_BACKGROUND
 
 World:
 	Inherits: ^BaseWorld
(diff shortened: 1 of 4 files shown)


Mailaender:
I am okay with this minimally invasive trait resorting. :+1:

abcdefg30:
Tab completion works again. :+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/659c4874761cc01b2650451107d5bced550156b6...b06d92ff221b1b64f5ce42e4304468d9a97e5e86)

PR #10805, opened 2016-02-21
Added WithExitOverlay rendered when the actor is leaving production

Mailaender:
Red Alert 2 has a different approach to war factory rendering. Instead of door opening/closing animations it simply draws overlays which a) hide the produced vehicle with a roof and b) add open door overlays. This was a bit weird to understand, but it is quite controllable when playing around with Z offsets. In followup pull requests this will differentiate per queue to allow the roof opening overlays for helicopters and airships.


pchote:
Considering that this isn't used in our default mods, it might be better to hold on this until you can implement the exit differentiation.  It could live in the RA2 dll until then?

Mailaender:
I believe it is generic enough to stay in the Mods.Common namespace.

pchote:
My concern is that this is unfinished, rather than not generic enough.

Mailaender:
Moved it completely to https://github.com/OpenRA/ra2/pull/145 now that the OpenRA.Mods.RA2.dll is actually buildable by Travis. https://github.com/OpenRA/ra2/pull/155

PR #10371, opened 2016-01-01
Generalized the DemoTruck trait for Tiberian Sun

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/9589.
One step closer to https://github.com/OpenRA/OpenRA/issues/7526.

PR #10808, opened 2016-02-22
Added the Devastator self destruction

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/4926. This also splits MadTank into more reusable parts: `AttackSeismic` for the screen shaking and chronoshift prevention and `AttackDetonates` for the charging and explosion as well as using the existing `WithChargeAnimation`/`Overlay`.


OpenRA.Mods.Common/Traits/Attack/AttackDetonates.cs
```diff
@@ -0,0 +1,140 @@
...
+	[Desc("Fires it's armament after a chargeup.")]
+	public class AttackDetonatesInfo : ITraitInfo
+	{
+		[Desc("Measured in ticks.")]
+		public readonly int ChargeDelay = 0;
+		public readonly string ChargeSound = null;
+
+		[Desc("Measured in ticks.")]
+		public readonly int DetonationDelay = 0;
+		[WeaponReference] public readonly string DetonationWeapon = null;
```
abcdefg30: I think this will want a `[FieldLoader.Require]`. (Or you need to check for `null` later on, line 67).

OpenRA.Mods.Common/Traits/Attack/AttackDetonates.cs
```diff
@@ -0,0 +1,140 @@
...
+
+		[VoiceReference] public readonly string Voice = "Action";
+
+		[Desc("Kills itself in the end.")]
+		public readonly bool Suicides = true;
+
+		public object Create(ActorInitializer init) { return new AttackDetonates(init.Self, this); }
+	}
+
+	public class AttackDetonates : IIssueOrder, IResolveOrder, IOrderVoice
```
pchote: This appears to duplicate a significant amount of `AttackSuicides`, which you introduced in #10371.

mods/d2k/rules/vehicles.yaml
```diff
@@ -337,6 +337,11 @@ devastator:
 		Step: 5
 		Ticks: 3
 		HealIfBelow: 50%
+	AttackDetonates:
```
pchote: My understanding is that this should only be available if the unit is at critical health.

abcdefg30:
FWIW, the behaviour ingame looks good. (You can count this as :+1: if above doesn't "trigger" "changes requested".)

OpenRA.Mods.Common/Traits/Attack/AttackDetonates.cs
Mailaender: Starting a detonation sequence, firing a weapon and then self-killing is different to immediate self-killing so I don't see how I can cut down any duplication by inheriting `AttackSuicides` and expanding `AttackSuicides` with charging, detonation will require a lot of conditionals and implement unused interfaces.

pchote: It sounds like you want a timed countdown that applies/remove upgrade, similar to the saboteur's cloaking ability.

mods/d2k/rules/vehicles.yaml
Mailaender: Upgradeable attack traits could solve this, but that feels like a big change in itself.

OpenRA.Mods.Common/Traits/Attack/AttackDetonates.cs
Mailaender: Doing this in unconnected traits might be hard to get the timing right as these are queued activities.

Mailaender:
Rebased.

OpenRA.Mods.Common/Traits/Attack/AttackDetonates.cs
Mailaender: Added the `[FieldLoader.Require]`.

Micr0Bit:
@Mailaender : is this the way it works in the original ?

mods/d2k/rules/vehicles.yaml
pchote: Detonation is a manually issued deploy order, so it doesn't need/want to be an attack trait at all.

MustaphaTR:
Devastator has self heal up to 50% in orginal D2K.

Micr0Bit:
MustaphaTR : any chance you can compare the "Self Destruction" here ... with the original from d2000 ?

MustaphaTR:
I will take a look to this when i back home. I'm at school now.

MustaphaTR:
I will take a look to this when i back home. I'm at school now.

Micr0Bit:
MustaphaTR : i have a feeling that you could still "move" the vehicle , while the detonation was allready triggered ... i mean , you could move the vehicle while the countdown for the detonation has started ...

MustaphaTR:
No you can't. At least in orginal. Triggering detonation stops the vehicle.

MustaphaTR:
I did some checks and found some problems. Damage looks too less in orginal only one Devastator can destroy can destroy a Power Plant/Barracks/Repair Deport. At OpenRA it doesn't even decrase its health to yellow.

Explosion anim also has some elctricity things on it, not just normal explosion. 

Not sure is it a problem but also current self destruct anim is translucent and makes it hard to see.

Also about AI with this, In orginal game if a Devastator changes its side with Deviator, or backs. AI usually but not always makes them self-destruct.

pchote:
My inline comment may be missed, so i'll repeat again here: this feature is manually triggered, and does not use the targeting cursor, so it is not appropriate to implement this as an `Attack*` subclass.  Ideally it should use the same deploy-to-timed-upgrade logic as the saboteur cloak (with the timed upgrade stopping the vehicle and showing the animation, and then an upgrade at the end of the sequence triggering a KillsSelf trait).

GraionDilach:
I have to agree with @Mailaender though, while your way would be the best, it would look complicatzed in yaml and also would overcomplicate the PR as well. (Although I have to acknowledge it would also bring RA2 Crazy Ivans a step closer.) I'd even suggest wait with such until the Upgrade->Condition rename and conditional upgrades, iirc planned by @atlimit8.

A random idea - unsure on feasibility - that the deploy would fire a weapon instead which has the unit artwork via one warhead and have the explosion in another with the delay set in `SpreadDamageWarhead` with a custom `DeathType`, which could only kill the Devastator, thereby triggering an `Explodes` trait specific to this `DeathType`. Although I'm not sure if this is how D2k worked, asin deploying is only one-way and the unit cannot be moved thenafter.

pchote:
Note that my comment is intentionally split into two parts.

The first is that `AttackDetonates` isn't appropriate for this feature, and comes with the changes requested tag (must be fixed to pass review).

The second is giving a constructive suggestion on how this could be done instead.  It doesn't *have* to be done that way, but it is going to be the one of the few potential implementations that won't need to be replaced again in the future.

penev92:
Oooor, we could look at the previous N discussions on the subject where I laid out an implementation plan that was *not* "complicatzed in yaml and also would overcomplicate the PR"  :|

(Spoiler alert: I haven't read all of ^ or the PR details, but I felt I should mention there was already a plan, since this seems pretty controversial).

Micr0Bit:
- only be available when the unit is at critical health
- correct damage-output (TibEd would have details on this values)
- correct explosion animation (3947-3964)
- fixing the less visible translucent self-destruct animation

and you get a triple-A-thumbs-up from me

mods/d2k/rules/vehicles.yaml
Mailaender: For the MAD tank it is both a deploy and an attack order. For a unit with another `Attack*` trait like the Devastator the `AttackFrontal` takes precedence.

MustaphaTR:
> * only be available when the unit is at critical health

What? Devastator self-destruct is avays available.

Micr0Bit:
thats what pchote said " It should also only be available when the unit is at critical health." ... did sound reasonable to me ...

MustaphaTR:
Aren't we still a bit early to do balance changes. I think we should add support of it but leave it as is.

Micr0Bit:
good , then without this restriction

Mailaender:
> TibEd would have details on this values

Can you provide those? I don't have access to that program.

Micr0Bit:
@Mailaender :

> i just switched to linux 1-2 weeks ago , and i did not get into that "wine-thing" so far

http://www.tibed.net/tibed1/

Mailaender:
I am also running Linux. It doesn't have to be you.

MustaphaTR:
I'm looking now. Wait a sec.

Micr0Bit:
@MustaphaTR :

we probably need more then that ... 

check -> Explosions -> Ex_Devastator 

place buildings (windtraps for example) around a Devastator ... look how far the damage spreads , and with what percentage of health buildings survive ... we should allready have the same hp as in the original (hopefully) ... 

also @OmegaBolt or @reaperrr should be able to help here

MustaphaTR:
I don't think we need that much testing. All other armor, versus, damage vaules macthes with orginal. We now know the damage we need. We just need to find a good fallout: vaule for explosion that looks like orginal.

Micr0Bit:
oh , yeah sry ... i did'nt check the screenshot correctly ... so the damage is 5000  ... and the range is 96 ...air - carryalls , arent effected

but why a "fallout-value" ?

MustaphaTR:
I meant `Falloff:` How the damage decrases by range may be different with this and orginal and this is what this tag controls. This is possibly hardcoded tho, so we can't understand easily.

Micr0Bit:
>I meant Falloff: How the damage decrases by range may be different with this and orginal and this is what this tag controls. This is possibly hardcoded tho, so we can't understand easily.

oh i didnt know that this is called falloff :) ... well looks like we are getting closer now

OmegaBolt:
In the original game the damage isn't dealt like a normal warhead detonation using falloff, it uses a YR style shrapnel system IIRC. When the Devastator explodes it "infects" a certain number of random targets around the vehicle that take the damage.

reaperrr:
Closing this for now to clear the PR queue.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/13884.

Issue #10091, opened 2015-11-23
SpreadDamageWarhead impacts spread use ambiguity


UberWaffe:
I started looking into warhead impacts because of [Comment](https://github.com/OpenRA/OpenRA/pull/9357#issuecomment-159034056).
It appears that nearly all uses of warhead impacts uses the impact at position, with only ```PoisonedByTiberium``` and ```DamagedWithoutFoundation``` applying directly to actors (as well as the pending ```AreaBeam```).

Long story short, in the cases that ```SpreadDamageWarhead``` is applied directly to actors, its ```Spread``` and ```Range``` is effectively ignored entirely.

The suggestion is that ```SpreadDamageWarhead``` should always use its ```Spread```, and that a new warhead ```DirectDamageWarhead``` should be defined for cases where no spread is needed.

Hopefully this will help avoid confusion where it appears that spread is not being applied.

Search done in project for uses of warhead impact uses.
```
...OpenRA.Game\GameRules\WeaponInfo.cs - (153, 15) : public void Impact(Target target, Actor firedBy, IEnumerable<int> damageModifiers)
...OpenRA.Mods.Cnc\Effects\IonCannon.cs - (47, 12) : weapon.Impact(target, firedBy.PlayerActor, Enumerable.Empty<int>());
...OpenRA.Mods.Cnc\Traits\PoisonedByTiberium.cs - (51, 20) : Info.WeaponInfo.Impact(Target.FromActor(self), self.World.WorldActor, Enumerable.Empty<int>());
...OpenRA.Mods.Common\Activities\Air\FallToEarth.cs - (41, 27) : info.ExplosionWeapon.Impact(Target.FromPos(self.CenterPosition), self, Enumerable.Empty<int>());
...OpenRA.Mods.Common\Effects\AreaBeam.cs - (181, 18) : args.Weapon.Impact(Target.FromActor(a), args.SourceActor, adjustedModifiers);
...OpenRA.Mods.Common\Effects\Bullet.cs - (209, 16) : args.Weapon.Impact(Target.FromPos(pos), args.SourceActor, args.DamageModifiers);
...OpenRA.Mods.Common\Effects\GravityBomb.cs - (78, 17) : args.Weapon.Impact(Target.FromPos(pos), args.SourceActor, args.DamageModifiers);
[...]

PR #10813, opened 2016-02-23
Added damaging through Veins.

Mailaender:
`-PoisonedByTiberium:` is currently the http://modenc.renegadeprojects.com/ImmuneToVeins equivalent. This is the easy part of getting veins to work.

### DIFF
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -257,7 +257,9 @@
 			DirtRoad: 100
 			Rough: 80
 			Tiberium: 90
+				PathingCost: 90
 			BlueTiberium: 90
+				PathingCost: 90
 			Veins: 50
 	SelectionDecorations:
 		Palette: pips
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -398,6 +398,7 @@
 			Tiberium: 50
 			BlueTiberium: 50
 			Veins: 50
+				PathingCost: 90
 		TurnSpeed: 5
 		Voice: Move
 	Selectable:
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -437,6 +437,9 @@
 		Weapons: SmallDebris
 		Pieces: 3, 7
 		Range: 2c0, 5c0
+	PoisonedByTiberium:
+		Weapon: Veins
+		Resources: Veins
 
 ^Tank:
 	Inherits: ^Vehicle
--- a/mods/ts/weapons/otherweapons.yaml
+++ b/mods/ts/weapons/otherweapons.yaml
@@ -66,3 +66,13 @@ Tiberium:
 		Spread: 42
 		Damage: 2
 		DamageTypes: EnergyDeath # TODO: FIX ME!
+
+Veins:
+	ReloadDelay: 16
+	Warhead@Damage: SpreadDamage
+		Spread: 42
+		Damage: 5
+		DamageTypes: EnergyDeath # TODO: FIX ME!
+	Warhead@Effect: CreateEffect
+		Explosions: veins
+		ExplosionPalette: player
--- a/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
@@ -91,6 +91,9 @@ namespace OpenRA.Mods.Common.Warheads
 
 		public override void DoImpact(Target target, Actor firedBy, IEnumerable<int> damageModifiers)
 		{
+			if (!target.IsValidFor(firedBy))
+				return;
+
 			var pos = target.CenterPosition;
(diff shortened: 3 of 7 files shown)


mods/ts/rules/defaults.yaml
```diff
@@ -417,6 +417,8 @@
 		Weapons: SmallDebris
 		Pieces: 3, 7
 		Range: 2c0, 5c0
+	PoisonedByTiberium:
+		Resources: Veins
```
pchote: This will damage vehicles with the "Tiberium" weapon, which probably doesn't make sense.

pchote: Rules.ini defines `VeinDamage=5`.

Mailaender: I have no idea what `VeinDamage=5` has to be translated to. The default damage makes sense. It damages vehicles slowly.

GraionDilach: Sounds like 5 damage per tick to me, `VeinDamage` is.

pchote: The question is per *whose* tick?  (per vein grab, per second, or per TS game tick?)

pchote: You can/should also set up the vein grab animation as a weapon effect.

GraionDilach: I haven't tested it but I'd assume TS game tick.

Mailaender:
Adjusted the vein damage and setup the vein tentacle sprites.

reaperrr:
- The Nod bike needs to be made immune to them as well
- the tentacle animation is only displayed very rarely, but it should loop constantly while it does damage

LipkeGu:
@reaperrr  Nod bikes arent veine immune! keep this in mind ... :)

Mailaender:
I am not sure why the tentacle animation is only displayed when the vehicle moves, because damage is dealt constantly. Added support for infantry avoiding Tiberium patches and vehicles avoiding vein pits.

OpenRA.Mods.Cnc/Traits/PoisonedByTiberium.cs
```diff
@@ -49,7 +49,7 @@ public void Tick(Actor self)
 			if (r == null || !Info.Resources.Contains(r.Info.Name))
 				return;
 
-			Info.WeaponInfo.Impact(Target.FromActor(self), self.World.WorldActor, Enumerable.Empty<int>());
+			Info.WeaponInfo.Impact(Target.FromPos(self.CenterPosition), self.World.WorldActor, Enumerable.Empty<int>());
```
pchote: Why change this? The old code is more logically correct, so if that doesn't work for some reason then something else will need to be fixed.

Mailaender: It crashes with the exceptions I mentioned in https://github.com/OpenRA/OpenRA/pull/10813/commits/7b80114e9c6e3cfcbe71a3c4dd1df5026e40f975.

pchote: That will only happen if the actor is already dead, so you can check and bail out inside DoImpact instead.

mods/ts/weapons/otherweapons.yaml
```diff
@@ -66,3 +66,14 @@ Tiberium:
...
+Veins:
+	ReloadDelay: 16
+	Warhead@Damage: SpreadDamage
+		Spread: 42
+		Damage: 5
+		DamageTypes: EnergyDeath # TODO: FIX ME!
+	Warhead@Effect: CreateEffect
+		Explosions: veins
+		ExplosionPalette: player
+		ValidImpactTypes: Ground
```
reaperrr: This makes it display ONLY if the target is 'missed'. You need to add `GroundHit` as well.

GraionDilach: Or remove it.

Mailaender:
Thanks for the help. Should be fixed now.

reaperrr:
:+1: 

@OpenRA/engine-hackers 
I'm wondering whether the last commit that checks for invalid targets in `CreateEffectWarhead.DoImpact` should actually go into prep/stable once this is merged? It fixes a potential crash, afterall.

Mailaender:
You can just cherry-pick https://github.com/OpenRA/OpenRA/pull/10813/commits/133a0f54a3673a84ebfff671fc967421d81af3f2 if you want it in prep/stable.

reaperrr:
Yes, assuming someone else reviews and merges this quickly enough.

GraionDilach:
This looks good as-is. :+1:

I'm not sure about the cherry-pick though. It's a partial fix to the issue #10091 brings up and I'm unsure how to treat it.

GraionDilach:
Meh, I reconsidered, I'm :+1:ing the cherry-pick idea as well. The other warheads are already aware of the edge-case #10091 brings up (this includes `CreateEffectWarhead` as well) so indeed it makes sense.

PR #10814, opened 2016-02-23
Fixed selectable crash sites

Mailaender:
These are decorational actors like rocks which shouldn't have a health bar. They also still contained internal developer tooltips which aren't really atmospheric.


abcdefg30:
:+1: / :white_check_mark:

RoosterDragon:
:+1:

Issue #8233, opened 2015-05-24
Starting a game on a map that has just been downloaded/installed kicks the player


obrakmann:
Downloading, leaving the lobby, entering again and then starting the game works.

Looks like some state isn't updated properly.

The bug already exists in at least release-20150424. Haven't investigated further.

pchote:
This ruins the usability of the map installation feature, so added to the milestone.

PR #10816, opened 2016-02-23
Fixed lobby self-kicking when clicking start game quickly

Mailaender:
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs#L742 sets the map to null so I can easily check against that. The problems appears when you click skirmish and immediately the start game button. If a debugger is attached and the map check is slowed down this will frequently happen. The game will even kick you in skirmish games which are technically a local server with one client.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
@@ -320,7 +320,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			var startGameButton = lobby.GetOrNull<ButtonWidget>("START_GAME_BUTTON");
 			if (startGameButton != null)
 			{
-				startGameButton.IsDisabled = () => configurationDisabled() ||
+				startGameButton.IsDisabled = () => configurationDisabled() || Map == null ||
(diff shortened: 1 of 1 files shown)


pchote:
Nice find!

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -320,7 +320,7 @@ void CloseWindow()
 			var startGameButton = lobby.GetOrNull<ButtonWidget>("START_GAME_BUTTON");
 			if (startGameButton != null)
 			{
-				startGameButton.IsDisabled = () => configurationDisabled() ||
+				startGameButton.IsDisabled = () => configurationDisabled() || Map == null ||
```
pchote: Actually, this doesn't make sense.... `Map == null` is already checked inside `configurationDisabled`!

Mailaender: Strange that it doesn't seem to work with my extra check.

Mailaender: Strange that it doesn't seem to work without my extra check.

pchote:
I don't think there's much more we can do here... I suspect that the self-kicking issue will be solved when we fix #8233.  Closing to tidy the PR queue.

PR #10820, opened 2016-02-24
Adjust the bounds of structures in TD

abcdefg30:
My brother was so kind to do those little adjustments, he also made images for you (reviewers)! ;)
(Unfortunately he used different ingame colors, but whatever.)

bounds1 bounds2 bounds3
bounds4 bounds5
(He also tweaked the bounds of `PYLE` a bit, but it's hardly noticeable anyway, as he just "cut" the bounds so that they don't overlap with the bib too much.)

### DIFF
--- a/mods/cnc/rules/structures.yaml
+++ b/mods/cnc/rules/structures.yaml
@@ -266,6 +266,8 @@ PYLE:
 	Power:
 		Amount: -20
 	ProvidesPrerequisite@buildingname:
+	Selectable:
+		Bounds: 48,42,0,-5
 
 HAND:
 	Inherits: ^BaseBuilding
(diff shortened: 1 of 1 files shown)


Phrohdoh:
This all looks pretty good to me. :+1:

Get him on IRC sometime so he can do more things that we're too ~~lazy~~ busy to do. :)

Mailaender:
Nice attention to details.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d2a3c411d2b4d0fbe11bc380ed024d9144af2059...e09612cdccff8bc90b00dc7b7f90bc978c29876d)

abcdefg30:
> I have one small concern, if you place a guard tower just behind/above another guard tower can you select the backmost one?

You still can (just tested it too), as actually just the visual bounds changed.

pchote:
We fixed that a while back: it will now select the actor with the closest CenterPosition, instead of the one with larger Y.

PR #10775, opened 2016-02-18
Carryall traits for tiberian sun carryall behavior

teees:
`AutoCarryall` and `AutoCarryable` behave like dune2k carryalls and carryables ~~and can be used there once this gets moved to `Common`~~

Replaces Dune2k traits and activites, moved everything to common

Fixes #8954
##### Dependencies:
- [x] #10824
###### major:
- Units using upgrades to deploy can be picked up while deployed
- ~~Loaded Carryalls are able to move into shroud even when MoveIntoShroud is false~~
- ~~Carryalls can be queue-ordered to pickup more than one unit, only the last unit gets dropped~~
###### cosmetic:
- Units using upgrades to deploy (mobile sensors array, tick tank) will use the wrong preview when being carried (shown as deployed)
- Cancelling a loaded carryall (pressing S) will land the carryall below the height of the carryable
- ~~You can send a loaded carryall to rearm on the heli pad, which causes it to land there (same as above)~~
- Cancelling a unloaded carryall while descending to a carryable will let it stay at that position in-air
- ~~Facings don't match for sprite actors~~
- Shadow offset is incorrect for sprite actors (drawn in-air)

PR #10824, opened 2016-02-24
Add DynamicFacingInit to change the facing of previews dynamically

teees:
Needed for carryalls (#10775)

### DIFF
--- a/OpenRA.Game/Map/ActorInitializer.cs
+++ b/OpenRA.Game/Map/ActorInitializer.cs
@@ -11,7 +11,6 @@
 
 using System.Linq;
 using OpenRA.Primitives;
-using OpenRA.Traits;
 
 namespace OpenRA
 {
@@ -48,39 +47,6 @@ namespace OpenRA
 		T Value(World world);
 	}
 
-	public class FacingInit : IActorInit<int>
-	{
-		[FieldFromYamlKey] readonly int value = 128;
-		public FacingInit() { }
-		public FacingInit(int init) { value = init; }
-		public int Value(World world) { return value; }
-	}
-
-	public class LocationInit : IActorInit<CPos>
-	{
-		[FieldFromYamlKey] readonly CPos value = CPos.Zero;
-		public LocationInit() { }
-		public LocationInit(CPos init) { value = init; }
-		public CPos Value(World world) { return value; }
-	}
-
-	public class SubCellInit : IActorInit<SubCell>
-	{
-		[FieldFromYamlKey] readonly int value = (int)SubCell.FullCell;
-		public SubCellInit() { }
-		public SubCellInit(int init) { value = init; }
-		public SubCellInit(SubCell init) { value = (int)init; }
-		public SubCell Value(World world) { return (SubCell)value; }
-	}
-
-	public class CenterPositionInit : IActorInit<WPos>
-	{
-		[FieldFromYamlKey] readonly WPos value = WPos.Zero;
-		public CenterPositionInit() { }
-		public CenterPositionInit(WPos init) { value = init; }
-		public WPos Value(World world) { return value; }
-	}
-
 	public class OwnerInit : IActorInit<Player>
 	{
 		[FieldFromYamlKey] public readonly string PlayerName = "Neutral";
@@ -104,13 +70,11 @@ namespace OpenRA
 		}
 	}
 
-	// Allows maps / transformations to specify the faction variant of an actor.
-	public class FactionInit : IActorInit<string>
+	public class LocationInit : IActorInit<CPos>
 	{
-		[FieldFromYamlKey] public readonly string Faction;
-
-		public FactionInit() { }
-		public FactionInit(string faction) { Faction = faction; }
-		public string Value(World world) { return Faction; }
+		[FieldFromYamlKey] readonly CPos value = CPos.Zero;
+		public LocationInit() { }
+		public LocationInit(CPos init) { value = init; }
+		public CPos Value(World world) { return value; }
 	}
 }
--- a/OpenRA.Mods.Common/ActorInitializer.cs
+++ b/OpenRA.Mods.Common/ActorInitializer.cs
@@ -1,0 +1,55 @@
+﻿using System;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common
+{
+	public class FacingInit : IActorInit<int>
+	{
+		[FieldFromYamlKey] readonly int value = 128;
+		public FacingInit() { }
+		public FacingInit(int init) { value = init; }
+		public int Value(World world) { return value; }
+	}
+
+	public class DynamicFacingInit : IActorInit<Func<int>>
+	{
+		readonly Func<int> func;
+		public DynamicFacingInit(Func<int> func) { this.func = func; }
+		public Func<int> Value(World world) { return func; }
+	}
+
+	public class BodyAnimationFrameInit : IActorInit<uint>
+	{
+		[FieldFromYamlKey] readonly uint value = 0;
+		public BodyAnimationFrameInit() { }
+		public BodyAnimationFrameInit(uint init) { value = init; }
+		public uint Value(World world) { return value; }
+	}
+
+	public class SubCellInit : IActorInit<SubCell>
+	{
+		[FieldFromYamlKey] readonly int value = (int)SubCell.FullCell;
+		public SubCellInit() { }
+		public SubCellInit(int init) { value = init; }
+		public SubCellInit(SubCell init) { value = (int)init; }
+		public SubCell Value(World world) { return (SubCell)value; }
+	}
+
+	public class CenterPositionInit : IActorInit<WPos>
+	{
+		[FieldFromYamlKey] readonly WPos value = WPos.Zero;
+		public CenterPositionInit() { }
+		public CenterPositionInit(WPos init) { value = init; }
+		public WPos Value(World world) { return value; }
+	}
+
+	// Allows maps / transformations to specify the faction variant of an actor.
+	public class FactionInit : IActorInit<string>
+	{
+		[FieldFromYamlKey] public readonly string Faction;
+
+		public FactionInit() { }
+		public FactionInit(string faction) { Faction = faction; }
+		public string Value(World world) { return Faction; }
+	}
+}
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -195,6 +195,7 @@
     <Compile Include="Lint\LintBuildablePrerequisites.cs" />
     <Compile Include="Lint\LintExts.cs" />
     <Compile Include="LoadScreens\ModChooserLoadScreen.cs" />
+    <Compile Include="ActorInitializer.cs" />
     <Compile Include="ShroudExts.cs" />
     <Compile Include="Orders\BeaconOrderGenerator.cs" />
     <Compile Include="Orders\DeployOrderTargeter.cs" />
--- a/OpenRA.Mods.Common/Traits/Render/RenderVoxels.cs
+++ b/OpenRA.Mods.Common/Traits/Render/RenderVoxels.cs
@@ -9,6 +9,7 @@
  */
 #endregion
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Graphics;
@@ -20,7 +21,7 @@ namespace OpenRA.Mods.Common.Traits.Render
 	public interface IRenderActorPreviewVoxelsInfo : ITraitInfo
 	{
 		IEnumerable<VoxelAnimation> RenderPreviewVoxels(
-			ActorPreviewInitializer init, RenderVoxelsInfo rv, string image, WRot orientation, int facings, PaletteReference p);
+			ActorPreviewInitializer init, RenderVoxelsInfo rv, string image, Func<WRot> orientation, int facings, PaletteReference p);
 	}
 
 	public class RenderVoxelsInfo : ITraitInfo, IRenderActorPreviewInfo, Requires<BodyOrientationInfo>
@@ -58,11 +59,8 @@ namespace OpenRA.Mods.Common.Traits.Render
 				body.QuantizedFacings;
 			var palette = init.WorldRenderer.Palette(Palette ?? PlayerPalette + ownerName);
 
-			var ifacing = init.Actor.TraitInfoOrDefault<IFacingInfo>();
-			var facing = ifacing != null ? init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : ifacing.GetInitialFacing() : 0;
-			var orientation = WRot.FromFacing(facing);
 			var components = init.Actor.TraitInfos<IRenderActorPreviewVoxelsInfo>()
-				.SelectMany(rvpi => rvpi.RenderPreviewVoxels(init, this, image, orientation, facings, palette))
+				.SelectMany(rvpi => rvpi.RenderPreviewVoxels(init, this, image, init.GetOrientation(), facings, palette))
 				.ToArray();
 
 			yield return new VoxelPreview(components, WVec.Zero, 0, Scale, LightPitch,
--- a/OpenRA.Mods.Common/Traits/Render/WithFacingSpriteBody.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithFacingSpriteBody.cs
@@ -9,6 +9,7 @@
  */
 #endregion
 
+using System;
 using System.Collections.Generic;
 using OpenRA.Graphics;
 using OpenRA.Mods.Common.Graphics;
@@ -22,13 +23,10 @@ namespace OpenRA.Mods.Common.Traits.Render
 
 		public override IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer init, RenderSpritesInfo rs, string image, int facings, PaletteReference p)
 		{
-			var ifacing = init.Actor.TraitInfoOrDefault<IFacingInfo>();
-			var facing = ifacing != null ? init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : ifacing.GetInitialFacing() : 0;
-
-			var anim = new Animation(init.World, image, () => facing);
+			var anim = new Animation(init.World, image, init.GetFacing());
 			anim.PlayRepeating(RenderSprites.NormalizeSequence(anim, init.GetDamageState(), Sequence));
 
-			yield return new SpriteActorPreview(anim, WVec.Zero, 0, p, rs.Scale);
+			yield return new SpriteActorPreview(anim, () => WVec.Zero, () => 0, p, rs.Scale);
 		}
 	}
 
--- a/OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
@@ -9,6 +9,7 @@
  */
 #endregion
 
+using System;
 using System.Collections.Generic;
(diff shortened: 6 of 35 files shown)


OpenRA.Game/Map/ActorInitializer.cs
```diff
@@ -53,7 +54,23 @@ public class FacingInit : IActorInit<int>
...
+	}
+
+	public class DynamicFacingInit : FacingInit
+	{
+		readonly Func<int> func;
+		public DynamicFacingInit(Func<int> func) { this.func = func; }
+		public override int Value(World world) { return func(); }
+	}
+
+	public class BodyAnimationFrameInit : IActorInit<uint>
```
pchote: It looks like all of these `*Init`s except `OwnerInit` can be moved into mod code.  Can you create an `ActorInitializers.cs` file in Mods.Common for them?

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
```diff
@@ -47,13 +48,13 @@ public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer i
...
-			var facing = init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : 0;
-			var anim = new Animation(init.World, image, () => facing);
+			Func<int> facing = () => init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : 0;
+			var anim = new Animation(init.World, image, facing);
 			anim.PlayRepeating(RenderSprites.NormalizeSequence(anim, init.GetDamageState(), Sequence));
 
-			var orientation = body.QuantizeOrientation(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(facing)), facings);
-			var offset = body.LocalToWorld(Offset.Rotate(orientation));
-			yield return new SpriteActorPreview(anim, offset, offset.Y + offset.Z + 1, p, rs.Scale);
+			Func<WRot> orientation = () => body.QuantizeOrientation(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(facing())), facings);
```
pchote: You can simplify this down to `Func<WRot> orientation = () => body.QuantizeOrientation(WRot.FromFacing(facing()), facings);`

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
```diff
@@ -47,13 +48,13 @@ public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer i
...
+			Func<WVec> offset = () => body.LocalToWorld(Offset.Rotate(orientation()));
+			yield return new SpriteActorPreview(anim, offset, () => offset().Y + offset().Z + 1, p, rs.Scale);
```
pchote: Kind of sucks to evaluate `offset` twice.  Could you pull out an `Action<int>` that calculates it once and then takes the `.Y` and `.Z` from the result?

OpenRA.Mods.Common/Traits/Render/WithInfantryBody.cs
```diff
@@ -31,14 +31,10 @@ public class WithInfantryBodyInfo : UpgradableTraitInfo, IRenderActorPreviewSpri
 
 		public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer init, RenderSpritesInfo rs, string image, int facings, PaletteReference p)
 		{
-			var facing = 0;
 			var ifacing = init.Actor.TraitInfoOrDefault<IFacingInfo>();
-			if (ifacing != null)
-				facing = init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : ifacing.GetInitialFacing();
-
-			var anim = new Animation(init.World, image, () => facing);
+			var anim = new Animation(init.World, image, () => ifacing != null ? init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : ifacing.GetInitialFacing() : 0);
```
pchote: The `init` and `GetInitialFacing` aren't going to change, so moving them here hurts perf without giving us anything.  Can we keep this as it was?

OpenRA.Mods.Common/Traits/Render/WithParachute.cs
```diff
@@ -73,9 +74,9 @@ public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer i
 
 			var body = init.Actor.TraitInfo<BodyOrientationInfo>();
 			var facing = init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : 0;
```
pchote: This is going to miss the dynamic facing changes.

OpenRA.Mods.Common/Traits/Render/WithInfantryBody.cs
pchote: Actually this comments apply to most of the uses of `init.Contains<FacingInit>()`.  Hiding the dynamic-ness of the facing inside `DynamicFacingInit.Value` makes it easy to mess up the use of facings.  I think it would be much better to pull out a helper method that takes an init dict, a fallback facing (from InitialFacing or elsewhere), and returns a `Func<int>`.  Internally it would explicitly check for `DynamicFacingInit` vs `FacingInit`.

OpenRA.Mods.Common/Traits/Render/WithSpriteBarrel.cs
```diff
@@ -46,10 +47,10 @@ public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer i
 			var anim = new Animation(init.World, image, () => t.InitialFacing);
 			anim.Play(RenderSprites.NormalizeSequence(anim, init.GetDamageState(), Sequence));
 
-			var turretOrientation = body.QuantizeOrientation(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(t.InitialFacing)), facings);
-			var turretOffset = body.LocalToWorld(t.Offset.Rotate(turretOrientation));
+			Func<WRot> turretOrientation = () => body.QuantizeOrientation(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(t.InitialFacing)), facings);
```
pchote: Can use `WRot.FromFacing` here too.

pchote: The turret facing doesn't change, so this can stay as an `int` that you wrap in a `Func` in the ctor below.

OpenRA.Mods.Common/Traits/Render/WithSpriteRotorOverlay.cs
```diff
@@ -33,13 +34,13 @@ public class WithSpriteRotorOverlayInfo : ITraitInfo, IRenderActorPreviewSprites
...
-			var facing = init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : 0;
-			var anim = new Animation(init.World, image, () => facing);
+			Func<int> facing = () => init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : 0;
+			var anim = new Animation(init.World, image, facing);
 			anim.PlayRepeating(RenderSprites.NormalizeSequence(anim, init.GetDamageState(), Sequence));
 
-			var orientation = body.QuantizeOrientation(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(facing)), facings);
-			var offset = body.LocalToWorld(Offset.Rotate(orientation));
-			yield return new SpriteActorPreview(anim, offset, offset.Y + offset.Z + 1, p, rs.Scale);
+			Func<WRot> orientation = () => body.QuantizeOrientation(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(facing())), facings);
```
pchote: `WRot.FromFacing` (note that this helper was added after all this code was originally written, and I apparently forgot to backport it).

OpenRA.Mods.Common/Traits/Render/WithSpriteRotorOverlay.cs
```diff
@@ -33,13 +34,13 @@ public class WithSpriteRotorOverlayInfo : ITraitInfo, IRenderActorPreviewSprites
...
+			Func<WVec> offset = () => body.LocalToWorld(Offset.Rotate(orientation()));
+			yield return new SpriteActorPreview(anim, offset, () => offset().Y + offset().Z + 1, p, rs.Scale);
```
pchote: We want to avoid evaluating `offset()` twice here, too.

OpenRA.Mods.Common/Traits/Render/WithSpriteTurret.cs
```diff
@@ -45,15 +46,15 @@ public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer i
...
 
 			var anim = new Animation(init.World, image, () => turretFacing);
 			anim.Play(RenderSprites.NormalizeSequence(anim, init.GetDamageState(), Sequence));
 
-			var orientation = body.QuantizeOrientation(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(bodyFacing)), facings);
-			var offset = body.LocalToWorld(t.Offset.Rotate(orientation));
-			yield return new SpriteActorPreview(anim, offset, -(offset.Y + offset.Z) + 1, p, rs.Scale);
+			Func<WRot> orientation = () => body.QuantizeOrientation(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(bodyFacing())), facings);
+			Func<WVec> offset = () => body.LocalToWorld(t.Offset.Rotate(orientation()));
+			yield return new SpriteActorPreview(anim, offset, () => -(offset().Y + offset().Z) + 1, p, rs.Scale);
```
pchote: Here too.

OpenRA.Mods.Common/Traits/Render/WithVoxelTurret.cs
```diff
@@ -37,12 +39,12 @@ public IEnumerable<VoxelAnimation> RenderPreviewVoxels(ActorPreviewInitializer i
...
 
 			var voxel = VoxelProvider.GetVoxel(image, Sequence);
-			var turretOffset = body.LocalToWorld(t.Offset.Rotate(orientation));
+			Func<WVec> turretOffset = () => body.LocalToWorld(t.Offset.Rotate(orientation()));
 
 			var turretFacing = Turreted.GetInitialTurretFacing(init, t.InitialFacing, Turret);
-			var turretBodyOrientation = new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(turretFacing) - orientation.Yaw);
-			var turretOrientation = new[] { turretBodyOrientation, body.QuantizeOrientation(orientation, facings) };
-			yield return new VoxelAnimation(voxel, () => turretOffset, () => turretOrientation, () => false, () => 0);
+			Func<WRot> turretBodyOrientation = () => new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(turretFacing) - orientation().Yaw);
```
pchote: Can use `WRot.FromYaw` here (`FromFacing` won't work).

OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -124,4 +124,9 @@ interface IWallConnector
 		bool AdjacentWallCanConnect(Actor self, CPos wallLocation, string wallType, out CVec facing);
 		void SetDirty();
 	}
+
+	public interface IBodyAnimation
+	{
+		uint GetBodyAnimationFrame();
```
pchote: Might be cleaner as a property instead?

Phrohdoh: Shouldn't this take a `self` param and as such remain a method for consistency?

OpenRA.Game/Map/ActorInitializer.cs
teees: Moved all of them to Common, except OwnerInit and LocationInit

OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
```diff
@@ -753,4 +754,4 @@ cd "$(SolutionDir)"</PostBuildEvent>
   <Target Name="AfterBuild">
   </Target>
   -->
-</Project>
+</Project>
```
pchote: Please keep the EOF newline here (just drop this line change).

OpenRA.Mods.Common/Traits/Render/RenderVoxels.cs
```diff
@@ -59,8 +60,18 @@ public virtual IEnumerable<IActorPreview> RenderPreview(ActorPreviewInitializer
 			var palette = init.WorldRenderer.Palette(Palette ?? PlayerPalette + ownerName);
 
 			var ifacing = init.Actor.TraitInfoOrDefault<IFacingInfo>();
-			var facing = ifacing != null ? init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : ifacing.GetInitialFacing() : 0;
-			var orientation = WRot.FromFacing(facing);
+			Func<WRot> orientation = () => WRot.Zero;
+			if (ifacing != null)
```
pchote: This block is duplicated across a few places.  Can you pull it out into a helper method?

OpenRA.Mods.Common/Traits/Render/WithFacingSpriteBody.cs
```diff
@@ -23,12 +24,22 @@ public class WithFacingSpriteBodyInfo : WithSpriteBodyInfo, Requires<BodyOrienta
 		public override IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer init, RenderSpritesInfo rs, string image, int facings, PaletteReference p)
 		{
 			var ifacing = init.Actor.TraitInfoOrDefault<IFacingInfo>();
-			var facing = ifacing != null ? init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : ifacing.GetInitialFacing() : 0;
+			Func<int> facing = () => 0;
+			if (ifacing != null)
```
pchote: Likewise for this (for facing instead of WRot).

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
```diff
@@ -47,13 +48,19 @@ public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer i
 				p = init.WorldRenderer.Palette(Palette);
 
 			var body = init.Actor.TraitInfo<BodyOrientationInfo>();
-			var facing = init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : 0;
-			var anim = new Animation(init.World, image, () => facing);
+			Func<int> facing = init.Contains<FacingInit>() ? (Func<int>)(() => init.Get<FacingInit, int>()) : () => 0;
```
pchote: Do you really need the inner `(Func<int>)` cast here?

OpenRA.Mods.Common/Traits/Render/WithVoxelBody.cs
```diff
@@ -25,13 +25,13 @@ public class WithVoxelBodyInfo : UpgradableTraitInfo, IRenderActorPreviewVoxelsI
...
-		public IEnumerable<VoxelAnimation> RenderPreviewVoxels(ActorPreviewInitializer init, RenderVoxelsInfo rv, string image, WRot orientation, int facings, PaletteReference p)
+		public IEnumerable<VoxelAnimation> RenderPreviewVoxels(
+			ActorPreviewInitializer init, RenderVoxelsInfo rv, string image, Func<WRot> orientation, int facings, PaletteReference p)
 		{
 			var body = init.Actor.TraitInfo<BodyOrientationInfo>();
 			var voxel = VoxelProvider.GetVoxel(image, "idle");
-			var bodyOrientation = new[] { body.QuantizeOrientation(orientation, facings) };
 			yield return new VoxelAnimation(voxel, () => WVec.Zero,
-				() => bodyOrientation,
+				() => new[] { body.QuantizeOrientation(orientation(), facings) },
```
pchote: @RoosterDragon would we win anything noticable by reusing the array (updating its value) instead of creating new objects in cases like this?

RoosterDragon: I wouldn't worry about it much unless it showed up in profiling as expensive.

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
teees: I could put it in yet another helper function - it's a bit tricky to assign a lambda function using the ternary operator. (see http://stackoverflow.com/questions/11309557/assign-a-lambda-expression-using-the-conditional-ternary-operator)

teees:
Rebased

OpenRA.Mods.Common/ActorInitializer.cs
```diff
@@ -0,0 +1,60 @@
...
+	public class FacingInit : IActorInit<int>
+	{
+		[FieldFromYamlKey]
+		readonly int value = 128;
+		public FacingInit() { }
+		public FacingInit(int init) { value = init; }
+		public virtual int Value(World world) { return value; }
+	}
+
+	public class DynamicFacingInit : FacingInit
```
pchote: I mentioned this earlier, but it may have been lost in the scrum of other comments: can we please make this not subclass `FacingInit`?  My suggestions for the `GetFacing` and `GetOrientation` methods were specifically to enable this separation :smile:

The issue is that ActorInits are supposed to be immutable, and while this technically doesn't change that it means that in practice you can query something that you *think* is an immutable `FacingInit`, but is actually an immutable `DynamicFacingInit` that gives you different facing values each time you query it – violating expectations for a regular `FacingInit`.

teees:
@pchote I separated those two cases as far as possible, there is some code duplication, but it should probably save some resources. Not sure if I could get those into some helper function.

Also rebased, and rebased #10775 on top of it

teees:
Read your comment again and made the dynamicfacing actor initializer return a `Func<int>` as requested

teees:
rebased

pchote:
When testing this with #10775 actor turrets don't track the dynamic facing.  This is most visible with the Titan.

pchote:
Code looks good, and didn't spot any other issues aside from that.

OpenRA.Mods.Common/Traits/Render/WithIdleOverlay.cs
```diff
@@ -47,13 +48,34 @@ public IEnumerable<IActorPreview> RenderPreviewSprites(ActorPreviewInitializer i
...
+			{
+				var facing = init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : 0;
+				var anim = new Animation(init.World, image, () => facing);
+				anim.PlayRepeating(RenderSprites.NormalizeSequence(anim, init.GetDamageState(), Sequence));
+
+				var orientation = body.QuantizeOrientation(WRot.FromFacing(facing), facings);
+				var offset = body.LocalToWorld(Offset.Rotate(orientation));
+				var zOffset = offset.Y + offset.Z + 1;
+				yield return new SpriteActorPreview(anim, () => offset, () => zOffset, p, rs.Scale);
+			}
```
RoosterDragon: Is there a way to extract this code into a common method somewhere? It's far from trivial and used in several places.

pchote:
@teees: would you mind if someone else took this over to do the final fixes so we can get it merged?

teees:
@pchote of course no, feel free to take over anything needed :)

pchote:
I'll take this over and file a new PR when ready.

pchote:
While looking into the turret facing issue I found a problem that is going to keep on causing trouble if we don't fix it now.  I'll fix that and include my old rework of cargo rendering (which is finally feasible with this PR) as a testcase.  This will take a bit longer than I first expected.

PR #10835, opened 2016-02-26
Fix undisposed package when caching maps

RoosterDragon:
Discovered by Coverity.


pchote:
`package` is stored in the map preview, and then used extensively through the lifetime of the mod.

I'm in the process of pulling out a proper fix into its own pr from my ball of remaining oramod changes.

RoosterDragon:
I should have read the error more closely - it was merely complaining about the exceptional path (which does indeed need adjusting).

If you're working on some changes anyway, I'll let you apply those first.

PR #10837, opened 2016-02-26
Add tests for PriorityQueue

RoosterDragon:
- Added some tests to verify the queue does the right things.
- Fixed `Peek` to throw when the queue is empty.
- Adjusted the exception message to be consistent with the mesage from the BCL `Queue` & `Stack` when peeking/popping empty collections.
- Changed `if (level == 0 && index == 0)` to `if (level <= 0 && index <= 0)` to hopefully fix a false positive from static analysis via Coverity (this should provide a hint to the analysis that `level` and `index` cannot be negative in normal code paths)

### DIFF
--- a/OpenRA.Test/OpenRA.Game/PriorityQueueTest.cs
+++ b/OpenRA.Test/OpenRA.Game/PriorityQueueTest.cs
@@ -1,0 +1,48 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System;
+using NUnit.Framework;
+using OpenRA.Primitives;
+
+namespace OpenRA.Test
+{
+	[TestFixture]
+	class PriorityQueueTest
+	{
+		[TestCase(TestName = "PriorityQueue maintains invariants when adding and removing items.")]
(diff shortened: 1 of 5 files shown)


abcdefg30:
This needs a rebase for some reason, too?

obrakmann:
Yeah, all the PRs do. Either github broke something or our repo is fscked.

OpenRAWeb is fine, though.

pchote:
Somebody turned on https://help.github.com/articles/enabling-required-status-checks/ without talking about it first.  This force-disables merges unless the pr is based on the latest bleed commit (even if it doesn't conflict!), and is going to be a giant PITA because of that.

Mailaender:
I set https://github.com/OpenRA/OpenRA/settings/branches/bleed as it said
> Disables force-pushes to this branch and prevents it from being deleted.

I also enabled `Require status checks to pass before merging` which seemed to only include CI checks and is the rule anyway. Feel free to disable it. I didn't mean to disrupt activities. Just read about the Linux Mint incident and panicked.

Mailaender:
PS: I don't see the pull request to require a rebase? Everything looks green.

Mailaender:
Removed https://help.github.com/articles/enabling-required-status-checks/ but kept https://help.github.com/articles/configuring-protected-branches/ to resolve this. Sorry again.

OpenRA.Test/OpenRA.Game/PriorityQueueTest.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
```
pchote: Please don't forget our standard file header!

pchote:
Looks sensible, so :+1: after fixing the file header.

RoosterDragon:
Sorry for the delay, fixed.

abcdefg30:
:+1:

PR #10838, opened 2016-02-27
Rework map loading and saving.

pchote:
- Fixes bogosity with `MapPreview` disposing
- Cleans the last nastyness from the `IReadWritePackage` interface.
- Prevents saving maps to non-writable directories (Fixes #9430).
- Final prerequisites for `.oramod` support.

PR #10839, opened 2016-02-27
Extract map rules to external files (and map format changes).

### LINKED ISSUE #4251: MapFormat increment checklist
- [ ] Fix up the map smudge yaml layout. `cr1 60,79 0:` is inconsistent in that it doesn't follow the idiomatic yaml tree structure that we use for everything else.
- [ ] Let all files in map archives contribute to the map uid, instead of just map.yaml and map.bin. This lets maps with edited Lua code have different uids.
- [ ] Allow rules to be in their own files in addition to map.yaml.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1350490-mapformat-increment-checklist?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
This upgrades our map format to version 9, with the following changes:
- Removed `Smudges` block.  Initial map smudges can now be defined using the `InitialSmudges` property on `SmudgeLayer`.
- Map UIDs are now calculated by hashing all yaml, lua, and bin file in the map package.
- Custom rules/sequences/weapons/etc are now defined in their own files referenced from map.yaml.  Maps can also files in the mod dir (e.g. `cnc|rules/campaign-tooltips.yaml, cnc|rules/campaign-palettes.yaml`), which a future PR will use to reduce duplication in the TD campaign.

This also changes the mod upgrade command to _not_ upgrade custom maps, which was a common frustration.

Fixes #4251.

Depends on #10838


pchote:
I would like to include two more changes into this format bump:

Make map.png mandatory, and have the editor generate it at map save time.
* This is the most efficient way to solve the long delays generating the preview when loading the map chooser
* This makes the previews available to external tools, like the quick look generator i'd like to eventually write for OSX.

Move the `Options` block (and the matching mod.yaml `LobbyDefaults` block) parameters on to their associated traits.
* There's no reason why these couldn't or shouldn't be trait properties.
* This is the first step towards unhardcoding them from the engine.
* It simplifies the planned dedicated server – resource center queries (we only need to fetch the map rules).

If there are no objections, then I will incorporate these before we merge this.

GraionDilach:
Yes yes yes for map pngs, man that would be so nice, that would also allow modders to actually hide layout for say cooperative missions.

GraionDilach:
On a second thought... do this or should this update the map importers as well?

pchote:
Now depends on #10857, which does the MapOptions and MapVideos changes.  Doing that as a standalone PR first simplifies the process because we don't need to deal with N rules files.

pchote:
> Yes yes yes for map pngs, man that would be so nice, that would also allow modders to actually hide layout for say cooperative missions.

This has been possible for a couple of years (since #4878).  I'm suggesting that we change it from an optional extra to a required part of the map package as a performance optimization.

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
```diff
@@ -36,6 +42,33 @@ public class SmudgeLayerInfo : ITraitInfo
...
+		public static object LoadInitialSmudges(MiniYaml yaml)
+		{
+			MiniYaml smudgeYaml;
+			var nd = yaml.ToDictionary();
+			var smudges = new Dictionary<CPos, MapSmudge>();
+			if (nd.TryGetValue("InitialSmudges", out smudgeYaml))
+			{
+				try
+				{
+					foreach (var node in smudgeYaml.Nodes)
```
obrakmann: Wouldn't it make more sense to move the try/catch into the loop?

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -783,9 +779,256 @@ internal static void UpgradeMapFormat(ModData modData, IReadWritePackage package
...
+				if (worldNode.Value.Nodes.Any() && !rules.Value.Nodes.Contains(worldNode))
+					rules.Value.Nodes.Add(worldNode);
+
+				if (playerNode.Value.Nodes.Any() && !rules.Value.Nodes.Contains(playerNode))
+					rules.Value.Nodes.Add(playerNode);
+			}
+
+			// Format 9 -> 10 extracted map rules, sequences, voxelsequences, weapons, voices, music, notifications,
+			// and translations to external files, moved smudges to SmudgeLayer, and uses map.png for all maps
+			if (mapFormat < 10)
```
obrakmann: You didn't actually apply those changes in this PR.

pchote: Indeed.  There's not much point until #10857 is merged.

pchote:
Rebased.

abcdefg30:
:+1: / :white_check_mark:

obrakmann:
:+1:

obrakmann:
[Changeog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/3cf3d244478caf0a8fb0ff4c2c4822ee9211f0d8)

Pinging @ihptru: the resource center will eventually need to be updated to fetch advanced rules from rules.yaml (or whatever the author called it) instead of map.yaml

PR #10851, opened 2016-02-28
Fix ProductionFromMapEdge to work properly with ground units

obrakmann:
`ProductionFromMapEdge` has several limitations when it comes to working with ground units:
1. It doesn't respect rally points, and units are always ordered to the factory's location.
2. `Mobile.MoveIntoWorld` ends up using the `VisualMove` activity, which basically just moves a sprite across the screen but does not do any pathfinding or collision detection at all. Units will happily drive through other units and across rivers, cliffs etc.
3. As a consequence of both (1) and (2) units end up driving right into the middle of their production building.

This PR fixes all of that.

Testcase included :)

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -250,6 +250,7 @@ namespace OpenRA
 
 		[FieldLoader.Ignore] public ProjectedCellRegion ProjectedCellBounds;
 		[FieldLoader.Ignore] public CellRegion AllCells;
+		public List<CPos> AllEdgeCells { get; private set; }
 
 		void AssertExists(string filename)
 		{
@@ -392,6 +393,8 @@ namespace OpenRA
 			CustomTerrain = new CellLayer<byte>(this);
 			foreach (var uv in AllCells.MapCoords)
 				CustomTerrain[uv] = byte.MaxValue;
+
+			AllEdgeCells = UpdateEdgeCells();
 		}
 
 		void InitializeCellProjection()
@@ -1056,20 +1059,44 @@ namespace OpenRA
 			return edge.V == Bounds.Bottom ? unProjected.MaxBy(x => x.V) : unProjected.MinBy(x => x.V);
 		}
 
+		public CPos ChooseClosestMatchingEdgeCell(CPos cell, Func<CPos, bool> match)
+		{
+			return AllEdgeCells.OrderBy(c => (cell - c).Length).FirstOrDefault(c => match(c));
+		}
+
+		List<CPos> UpdateEdgeCells()
+		{
+			var edgeCells = new List<CPos>();
+			var unProjected = new List<MPos>();
+			var bottom = Bounds.Bottom - 1;
+			for (var u = Bounds.Left; u < Bounds.Right; u++)
+			{
+				unProjected = Unproject(new PPos(u, Bounds.Top));
+				if (unProjected.Any())
+					edgeCells.Add(unProjected.MinBy(x => x.V).ToCPos(Grid.Type));
+
+				unProjected = Unproject(new PPos(u, bottom));
+				if (unProjected.Any())
+					edgeCells.Add(unProjected.MaxBy(x => x.V).ToCPos(Grid.Type));
+			}
+
+			for (var v = Bounds.Top; v < Bounds.Bottom; v++)
+			{
+				unProjected = Unproject(new PPos(Bounds.Left, v));
+				if (unProjected.Any())
+					edgeCells.Add((v == bottom ? unProjected.MaxBy(x => x.V) : unProjected.MinBy(x => x.V)).ToCPos(Grid.Type));
+
+				unProjected = Unproject(new PPos(Bounds.Right - 1, v));
+				if (unProjected.Any())
+					edgeCells.Add((v == bottom ? unProjected.MaxBy(x => x.V) : unProjected.MinBy(x => x.V)).ToCPos(Grid.Type));
+			}
+
+			return edgeCells;
+		}
+
 		public CPos ChooseRandomEdgeCell(MersenneTwister rand)
 		{
-			List<MPos> cells;
-			do
-			{
-				var isU = rand.Next(2) == 0;
-				var edge = rand.Next(2) == 0;
-				var u = isU ? rand.Next(Bounds.Left, Bounds.Right) : (edge ? Bounds.Left : Bounds.Right);
-				var v = !isU ? rand.Next(Bounds.Top, Bounds.Bottom) : (edge ? Bounds.Top : Bounds.Bottom);
-
-				cells = Unproject(new PPos(u, v));
-			} while (!cells.Any());
-
-			return cells.Random(rand).ToCPos(Grid.Type);
+			return AllEdgeCells.Random(rand);
 		}
 
 		public WDist DistanceToEdge(WPos pos, WVec dir)
--- a/OpenRA.Mods.D2k/Traits/Buildings/ProductionFromMapEdge.cs
+++ b/OpenRA.Mods.D2k/Traits/Buildings/ProductionFromMapEdge.cs
@@ -22,10 +22,17 @@ namespace OpenRA.Mods.D2k.Traits
 		public override object Create(ActorInitializer init) { return new ProductionFromMapEdge(init, this); }
 	}
 
-	class ProductionFromMapEdge : Production
+	class ProductionFromMapEdge : Production, INotifyCreated
 	{
+		RallyPoint rp;
+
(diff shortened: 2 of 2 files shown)


OpenRA.Game/Map/Map.cs
```diff
@@ -1068,20 +1071,37 @@ public MPos ChooseClosestEdgeCell(MPos uv)
...
+		public CPos ChooseClosestMatchingEdgeCell(CPos cell, Func<CPos, bool> match)
 		{
-			List<MPos> cells;
-			do
+			return AllEdgeCells.OrderBy(c => (cell - c).Abs().Length).FirstOrDefault(c => match(c));
+		}
+
+		List<CPos> UpdateEdgeCells()
+		{
+			List<CPos> edgeCells = new List<CPos>();
```
pchote: `var`

OpenRA.Game/Map/Map.cs
```diff
@@ -1068,20 +1071,37 @@ public MPos ChooseClosestEdgeCell(MPos uv)
...
-				var u = isU ? rand.Next(Bounds.Left, Bounds.Right) : (edge ? Bounds.Left : Bounds.Right);
-				var v = !isU ? rand.Next(Bounds.Top, Bounds.Bottom) : (edge ? Bounds.Top : Bounds.Bottom);
+				edgeCells.Add(Unproject(new PPos(u, Bounds.Top)).MinBy(x => x.V).ToCPos(Grid.Type));
+				edgeCells.Add(Unproject(new PPos(u, bottom)).MaxBy(x => x.V).ToCPos(Grid.Type));
+			}
 
-				cells = Unproject(new PPos(u, v));
-			} while (!cells.Any());
+			var unProjected = new List<MPos>();
+			for(var v = Bounds.Top; v < Bounds.Bottom; v++)
```
pchote: `for (`

OpenRA.Mods.D2k/Traits/Buildings/ProductionFromMapEdge.cs
```diff
@@ -17,27 +18,47 @@
...
+			var mobileInfo = producee.TraitInfoOrDefault<MobileInfo>();
+
+			var location = spawnLocation;
+			if (location == CPos.Zero)
+			{
+				if (aircraftInfo != null)
+					location = self.World.Map.ChooseClosestEdgeCell(self.Location);
+
+				if (mobileInfo != null)
+					location = self.World.Map.ChooseClosestMatchingEdgeCell(self.Location, c => mobileInfo.CanEnterCell(self.World, null, c));
```
pchote: You should prioritize cells that have the same DomainIndex as the destination.

OpenRA.Mods.D2k/Traits/Buildings/ProductionFromMapEdge.cs
```diff
@@ -17,27 +18,47 @@
...
 				pos += new WVec(0, 0, aircraftInfo.CruiseAltitude.Length);
 
-			var initialFacing = self.World.Map.FacingBetween(location, self.Location, 0);
+			var rp = self.TraitOrDefault<RallyPoint>();
+			var destination = rp != null ? rp.Location : self.Location;
+
+			var initialFacing = self.World.Map.FacingBetween(location, destination, 0);
 
 			self.World.AddFrameEndTask(w =>
 				{
```
pchote: Looks like this whole block can be unindented by one tab.

obrakmann:
Updated. Good call on the DomainIndex.

OpenRA.Game/Map/Map.cs
```diff
@@ -1068,20 +1071,44 @@ public MPos ChooseClosestEdgeCell(MPos uv)
 			return edge.V == Bounds.Bottom ? unProjected.MaxBy(x => x.V) : unProjected.MinBy(x => x.V);
 		}
 
-		public CPos ChooseRandomEdgeCell(MersenneTwister rand)
+		public CPos ChooseClosestMatchingEdgeCell(CPos cell, Func<CPos, bool> match)
 		{
-			List<MPos> cells;
-			do
+			return AllEdgeCells.OrderBy(c => (cell - c).Abs().Length).FirstOrDefault(c => match(c));
```
pchote: The `Abs()` isn't needed because the length calculation doesn't care about the sign of the components.

pchote:
This feels like something that could really use a unit test, but that may take an unreasonable amount of work.  Any thoughts?

pchote:
I don't have time to test this properly yet, but if someone wants to beat me to it I suggest temporarily modifying `TerrainGeometryOverlay` to draw the `Map.AllEdgeCells` and verify that they look sensible across different maps (especially in TS).

OpenRA.Mods.D2k/Traits/Buildings/ProductionFromMapEdge.cs
```diff
@@ -17,62 +18,100 @@
...
 		{
-			var location = self.World.Map.ChooseClosestEdgeCell(self.Location);
+			var aircraftInfo = producee.TraitInfoOrDefault<AircraftInfo>();
+			var mobileInfo = producee.TraitInfoOrDefault<MobileInfo>();
+
+			var passable = mobileInfo != null ? (uint)mobileInfo.GetMovementClass(self.World.TileSet) : 0;
+			var destination = rp != null ? rp.Location : self.Location;
+
+			var location = spawnLocation;
+			if (location == CPos.Zero)
```
pchote: Relying on `CPos.Zero` as a magic value has caused lots of problems in the past.  Can we make spawnLocation a `CPos?` and check for `HasValue` instead?

OpenRA.Mods.D2k/Traits/Buildings/ProductionFromMapEdge.cs
```diff
@@ -17,62 +18,100 @@
...
 	{
+		readonly CPos spawnLocation;
+		readonly RallyPoint rp;
+		readonly DomainIndex domainIndex;
+
 		public ProductionFromMapEdge(ActorInitializer init, ProductionInfo info)
-			: base(init, info) { }
+			: base(init, info)
+		{
+			rp = init.Self.TraitOrDefault<RallyPoint>();
```
pchote: This will introduce a yaml-ordering dependency.  Please `ProductionFromMapEdgeInfo: INotifyCreated` and push this into the `Created()` method.

obrakmann:
Updated. Added drawing all the edge cells in a solid red color in the terrain debug overlay to the test case commit.

obrakmann:
Here's a diff to compare the edge cells on TS between this branch and bleed. As can be seen, on bleed `ChooseRandomEdgeCell` only returns cells from the left or top edge. On this branch, all edges work. The gaps and half-cell differences in the edge cells appear on both branches, so I'm assuming it's due to the way un-projecting cells works.

```
diff --git a/OpenRA.Mods.Common/Traits/World/TerrainGeometryOverlay.cs b/OpenRA.Mods.Common/Traits/World/TerrainGeometryOverlay.cs
index 4b304ad..51558d2 100644
--- a/OpenRA.Mods.Common/Traits/World/TerrainGeometryOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/World/TerrainGeometryOverlay.cs
@@ -91,6 +91,17 @@ public void RenderAfterWorld(WorldRenderer wr, Actor self)
                                        wcr.DrawLine(screen[i], screen[j], 3 / wr.Viewport.Zoom, Color.Navy);
                                }
                        }
+
+                       // Draw some edge cells
+                       var qr = Game.Renderer.WorldRgbaColorRenderer;
+                       for(var i=0; i<250; i++)
+                       {
+                               var cell = map.ChooseRandomEdgeCell(wr.World.SharedRandom);
+                               var pos = wr.World.Map.CenterOfCell(cell);
+                               var tl = wr.ScreenPxPosition(pos - new WVec(512, 512, 0));
+                               var br = wr.ScreenPxPosition(pos + new WVec(511, 511, 0));
+                               qr.FillRect(tl, br, Color.Red);
+                       }
                }
        }
 }
```

pchote:
Thanks for the fixes.  Looks good and works well ingame :+1:

pchote:
Adding fixup tag so we don't forget to remove the test case.

pchote:
Can we get another review here?  Would be good to get this merged.

abcdefg30:
You can remove the testcommit now.

obrakmann:
Test case removed.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/78fa3f32ebe24fea97215b8908c8ce0bdbf36d2e...785755cd6a9d2820911ef09bac7e1e061c315a00)

PR #10857, opened 2016-03-02
Remove hardcoded map options and videos from the Map format.

pchote:
Another step towards streamlining the map format (reducing the chances that it will have to change again in the future), and the first step towards supporting mod-defined lobby options.


pchote:
Question: The `Description` field has never been used for anything useful aside from mission briefings.  Do we want to add a `BriefingText` field to `MapVideos` (and then rename it to e.g. `MissionData`) and drop that cruft too?

ScottNZ:
I think It would be nice to show the description in normal skirmish/MP lobbies

pchote:
I can see that discussion being controversial, so lets move it to #10858 instead.

pchote:
Now fixes #10858.

abcdefg30:
Random question (seems like I missed the discussion on IRC): Why change the map format to 9?
The release has still 7 so this would "pass unnoticed" (as we're already changing to 8).
(Or do we want to use the "Look how much we worked!" approach? :P)

pchote:
These changes break any non-upgraded version 8 maps, of which there are at least a few out there.

OpenRA.Mods.Common/UtilityCommands/UpgradeMapCommand.cs
```diff
@@ -23,13 +25,25 @@ public bool ValidateArguments(string[] args)
 			return args.Length >= 3;
 		}
 
-		[Desc("MAP", "CURRENTENGINE", "Upgrade map rules to the latest engine version.")]
-		public void Run(ModData modData, string[] args)
+		delegate void UpgradeAction(int engineVersion, ref List<MiniYamlNode> nodes, MiniYamlNode parent, int depth);
+
+		static void ProcessYaml(Map map, IEnumerable<string> files, int engineDate, UpgradeAction processFile)
```
obrakmann: What is this used by? It looks like dead code. `UpgradeAction`, too.

OpenRA.Game/Server/Server.cs
```diff
@@ -154,16 +154,14 @@ public Server(IPEndPoint endpoint, ServerSettings settings, ModData modData)
...
-			foreach (var t in serverTraits.WithInterface<INotifyServerStart>())
-				t.ServerStarted(this);
-
-			Log.Write("server", "Initial mod: {0}", ModData.Manifest.Mod.Id);
-			Log.Write("server", "Initial map: {0}", LobbyInfo.GlobalSettings.Map);
-
 			new Thread(_ =>
 			{
+				foreach (var t in serverTraits.WithInterface<INotifyServerStart>())
+					t.ServerStarted(this);
```
obrakmann: This ends up running `LobbyCommands.LoadMap()`. Is it safe do to that off the main thread?

OpenRA.Game/Traits/Player/DeveloperMode.cs
```diff
@@ -14,6 +14,12 @@ namespace OpenRA.Traits
 	[Desc("Attach this to the player actor.")]
 	public class DeveloperModeInfo : ITraitInfo
 	{
+		[Desc("Default value of the developer mode checkbox in the lobby.")]
+		public bool Enabled = false;
+
+		[Desc("Prevent the enabled state from being changed in the lobby.")]
```
obrakmann: The enabled state of what?

OpenRA.Mods.Common/Traits/World/CrateSpawner.cs
```diff
@@ -20,6 +20,12 @@ namespace OpenRA.Mods.Common.Traits
 {
 	public class CrateSpawnerInfo : ITraitInfo
 	{
+		[Desc("Default value of the crates checkbox in the lobby.")]
+		public readonly bool Enabled = true;
+
+		[Desc("Prevent the enabled state from being changed in the lobby.")]
```
obrakmann: The enabled state of what?

OpenRA.Game/Traits/Player/DeveloperMode.cs
pchote: The field two lines above.

OpenRA.Mods.Common/Traits/World/CrateSpawner.cs
pchote: The field two lines above.

OpenRA.Mods.Common/Traits/World/MapCreeps.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Controls the 'Creeps' checkbox in the lobby options.")]
+	public class MapCreepsInfo : TraitInfo<MapCreeps>
+	{
+		[Desc("Default value of the creeps checkbox in the lobby.")]
+		public readonly bool Enabled = true;
+
+		[Desc("Prevent the enabled state from being changed in the lobby.")]
```
obrakmann: The enabled state of what?

OpenRA.Game/Server/Server.cs
pchote: That was actually the entire purpose of this change: our horrible load screen code will otherwise flash the load screen when opening the skirmish lobby (because it is running on the main thread).  Fixing the jank when starting a skirmish was just a nice side effect.

As far as I can tell this should be safe: we already load maps off the main thread in the lobby and mission browser, and the server does not start listening for clients until later in the thread.

OpenRA.Mods.Common/UtilityCommands/UpgradeMapCommand.cs
pchote: This will be introduced and used by #10839.  For now it's just a rebase failure.

OpenRA.Mods.Common/Traits/World/CrateSpawner.cs
obrakmann: That's not obvious from the context, especially when a TraitInfo has more than just these two fields. Please make it more specific, as you already did with many of the other moved options.

OpenRA.Mods.Common/Traits/World/MapOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+	public class MapOptionsInfo : TraitInfo<MapOptions>
+	{
+		[Desc("Default value of the short game checkbox in the lobby.")]
+		public readonly bool ShortGameEnabled = true;
+
+		[Desc("Prevent the short game enabled state from being changed in the lobby.")]
+		public readonly bool ShortGameLocked = false;
+
+		[Desc("Default tech level.")]
+		public readonly string TechLevel = "none";
```
obrakmann: Needs to be "Unrestricted"

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -482,19 +520,15 @@ void CloseWindow()
...
-				techLevel.IsVisible = () => techTraits.Count > 0;
-
-				var techLevelDescription = optionsBin.GetOrNull<LabelWidget>("TECHLEVEL_DESC");
-				if (techLevelDescription != null)
-					techLevelDescription.IsVisible = () => techTraits.Count > 0;
+				var mapOptions = new CachedTransform<Map, MapOptionsInfo>(
+					map => map.Rules.Actors["world"].TraitInfo<MapOptionsInfo>());
 
-				techLevel.IsDisabled = () => configurationDisabled() || Map.Options.TechLevel != null || techTraits.Count <= 1;
+				techLevel.IsDisabled = () => configurationDisabled() || mapOptions.Update(Map).TechLevelLocked;
```
obrakmann: This doesn't disable the dropdown when there are no techlevels defined (as it is the case in TS atm). Of course you should argue that the dropdown shouldn't even be there then.

obrakmann:
I get this when I join a server with a standard map and default lobby options:
screenshot from 2016-03-03 21 13 25

OpenRA.Game/Traits/World/Shroud.cs
```diff
@@ -18,6 +18,18 @@ namespace OpenRA.Traits
...
 	public class ShroudInfo : ITraitInfo
 	{
+		[Desc("Default value of the fog checkbox in the lobby.")]
+		public bool FogEnabled = true;
+
+		[Desc("Prevent the fog enabled state from being changed in the lobby.")]
+		public bool FogLocked = false;
+
+		[Desc("Default value of the explore map checkbox in the lobby.")]
+		public bool ExploredMapEnabled = true;
```
obrakmann: Wait, you're sneaking something in here :p

pchote:
Fixed.

obrakmann:
Crash when changing the map to Fort Lonestar in the lobby:
```
  at OpenRA.Graphics.SpriteFont.DrawText (System.String text, float2 location, Color c) [0x000c8] in /home/oliver/devel/openra/git/OpenRA.Game/Graphics/SpriteFont.cs:65 
  at OpenRA.Mods.Common.Widgets.ButtonWidget.Draw () <0x4173a7c0 + 0x00567> in <filename unknown>:0 
  at OpenRA.Mods.Common.Widgets.DropDownButtonWidget.Draw () <0x4173b0f0 + 0x00057> in <filename unknown>:0
```

pchote:
Fixed.

obrakmann:
I just had this crash, but don't remember exactly what I did so can't reproduce it. I was switching from one map to another (Fort Lonestar to Arctic Triangle Affair, I think, but that doesn't crash):
```
  at OpenRA.Mods.Common.Widgets.Logic.LobbyLogic.<LobbyLogic>m__B (OpenRA.Map map) <0x40c2d470 + 0x0000b> in <filename unknown>:0 
  at OpenRA.Widgets.CachedTransform`2[T,U].Update (OpenRA.Widgets.T input) [0x0006a] in /home/oliver/devel/openra/git/OpenRA.Game/Widgets/WidgetUtils.cs:286 
  at OpenRA.Mods.Common.Widgets.Logic.LobbyLogic+<LobbyLogic>c__AnonStorey16.<>m__0 () <0x40c2d410 + 0x0002b> in <filename unknown>:0 
```

OpenRA.Mods.Common/Widgets/Logic/MissionBrowserLogic.cs
```diff
@@ -177,15 +178,42 @@ void SelectMap(MapPreview preview)
...
+			var infoVideo = "";
+			var infoVideoVisible = false;
+			var infoVideoDisabled = true;
+
+			new Thread(() =>
+			{
+				selectedMap.PreloadRules();
+				var mapOptions = selectedMap.Rules.Actors["world"].TraitInfo<MapOptionsInfo>();
+
+				difficulty = mapOptions.Difficulty;
```
obrakmann: It seems that maps have this set to `null` by default, even when they have difficulties defined. So this should probably be `difficulty = mapOptions.Difficulty ?? mapOptions.Difficulties.FirstOrDefault();`

pchote: Yup.  Fixed.

pchote:
> I just had this crash, but don't remember exactly what I did so can't reproduce it. 

Reproduced and fixed.  I forgot a `Map == null` check when I added the techlevel hiding with no tech levels.

obrakmann:
OK, I think it's fine now, can't see any more obvious breakage. :+1:

abcdefg30:
Something is wrong with the mission descriptions:
missionbriefings
Allies01 lacks a scroll button, so you can't see the entire text, while
Allies02 has a scroll button, just to scroll into nothing.

abcdefg30:
Wait, actually there is more bogus stuff going on here, while switching between Allies03a and Allies03b, I sometimes get a scroll bar, but other times not.

pchote:
Rewrote the mission browser part of 01a14d9ae5ffe5272537f4cc2257267d1314acf9 to fix that.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/861e135cfb795d2a93087837056d56ca4333a132...7d07481348187412c91e602a4bd2a5faa2b95d13)

PR #10862, opened 2016-03-02
Adjust TS Actor stats to match vanilla TS and FS stats.

ABrandau:
So far me and dutchygamer have only checked the basic stats of the vanilla actors, hp, cost, armor types, etc. We can't say if the other things like unit speed and the weapons match the vanilla TS and FS data.

### DIFF
--- a/mods/ts/rules/gdi-infantry.yaml
+++ b/mods/ts/rules/gdi-infantry.yaml
@@ -37,6 +37,8 @@ MEDIC:
 		Prerequisites: ~gapile
 	Voiced:
 		VoiceSet: Medic
+	RevealsShroud:
+		Range: 6c0		
 	Mobile:
 		Speed: 56
 	Health:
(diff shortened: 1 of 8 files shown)


pchote:
A note about world-unit ranges: `1c0` sometimes does *not* match 1 cell in ts-terms (but sometimes, if there's an internal conversion from WDist to cells like for sight ranges, it does!).  In OpenRA `1c0` means the horizontal width, which is actually `sqrt(2) = 1.1414` of a TS cell.  This means that a TS six cell range actually needs to become `4c248` in OpenRA.

Mailaender:
> Adjust TS Actor stats to match vanilla TS and FS stats. 

We took TS (not FS) for reference. I assume you migrated the rules to FS now.

ABrandau:
The rule we took was to match the FS rules if aplicable, and if there are none, then use the TS rules, but there are not that many stat balances on FS, so no.

Mailaender:
:+1: Thanks for the effort. As most rules were added manually copy & paste errors are not a surprise. Also the focus was on getting functionality and rendering correctly during initial rule setups.

abcdefg30:
:+1:

PR #10863, opened 2016-03-02
Renamed ROT to TurnSpeed

reaperrr:
...on `Mobile`, `Aircraft`, `Turreted` and `ThrowsParticle`.

It was mentioned long time ago that we should get rid of abbreviations.

For the record: `TurnSpeed` got the most agreement on IRC (because it matches `Speed`), so I will not rename it again.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Turreted.cs
+++ b/OpenRA.Mods.Common/Traits/Turreted.cs
@@ -18,8 +18,8 @@ namespace OpenRA.Mods.Common.Traits
 	public class TurretedInfo : ITraitInfo, UsesInit<TurretFacingInit>, Requires<BodyOrientationInfo>
 	{
 		public readonly string Turret = "primary";
-		[Desc("Rate of Turning")]
-		public readonly int ROT = 255;
+		[Desc("Speed at which the turret turns.")]
+		public readonly int TurnSpeed = 255;
 		public readonly int InitialFacing = 128;
 
 		[Desc("Number of ticks before turret is realigned. (-1 turns off realignment)")]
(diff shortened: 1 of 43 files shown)


OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -599,6 +599,15 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 					}
 				}
 
+				if (engineVersion < 20160301)
+				{
+					// Renamed ROT -> TurnSpeed
+					if (node.Key == "ROT")
+					{
```
Phrohdoh: Unnecessary brackets.

obrakmann:
:+1:

OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -19,7 +19,7 @@ public class TurretedInfo : ITraitInfo, UsesInit<TurretFacingInit>, Requires<Bod
 	{
 		public readonly string Turret = "primary";
 		[Desc("Rate of Turning")]
```
GraionDilach: Shouldn't this be updated then as well?

reaperrr: Yeah, well spotted.

GraionDilach: There is another instance of this though. So yeah, all descs should be unified as part of this PR imo.

reaperrr:
Updated descriptions.

Also changed `Turreted`'s `InitialFacing` to 0, to match `Mobile`.

abcdefg30:
:+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c4e1ab43e44df23bc425ace06b3952b3f7b457fc...a24659a4bef92ad3a28350400825ca3a038add16)

PR #10875, opened 2016-03-04
Close some packages

RoosterDragon:
When validating mods, close packages we don't eventually return.

Discovered by Coverity.

### DIFF
--- a/OpenRA.Game/ModMetadata.cs
+++ b/OpenRA.Game/ModMetadata.cs
@@ -38,9 +38,9 @@ namespace OpenRA
 			var ret = new Dictionary<string, ModMetadata>();
 			foreach (var pair in GetCandidateMods())
 			{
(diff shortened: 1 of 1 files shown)


pchote:
The package may also be left open if the yaml parsing throws an exception.  It would pay to add a check dispose in the catch block too.

RoosterDragon:
Nice catch, added.

obrakmann:
:+1:

Mailaender:
Not sure if this is a recent regression from https://github.com/OpenRA/OpenRA/issues/10717 or something that now has been fixed compared to the previous release which might justify a changelog entry.

pchote:
Recent regression. [Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4b43b1f74a632aa8753045b5bdc0a99c7b645577).

PR #10876, opened 2016-03-04
Fix a theoretical null reference in UpgradeWeaponRules.

RoosterDragon:
Parent is null at the root level, so we should ensure we don't explode on that.

Discovered by Coverity.


abcdefg30:
:+1:

pchote:
:+1:

PR #10877, opened 2016-03-04
Ensure we do not leak bitmaps when parsing invalid PNG files

RoosterDragon:
If a bitmap contained another header, we would leak the first bitmap and keep going rather than yelling about the invalid file.

Also, ensures we clean up the bitmap if there is an exception some time before we get to return it.

[DIff ignoring whitespace](https://github.com/OpenRA/OpenRA/pull/10877/files?w=1)

Discovered by Coverity.

### DIFF
--- a/OpenRA.Game/FileFormats/PngLoader.cs
+++ b/OpenRA.Game/FileFormats/PngLoader.cs
@@ -42,114 +42,126 @@ namespace OpenRA.FileFormats
 				Color[] palette = null;
 				var data = new List<byte>();
 
-				for (;;)
+				try
 				{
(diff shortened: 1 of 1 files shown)


pchote:
Looks fine, and code still works :+1:

Note that this is only used by the utility `ConvertPngToShpCommand`.  Our ingame png loading uses the standard .NET types.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/13ffeb7e9926f2e4177020921fc1a7492b6db646...fee7e5f5ea03e6eca4b5fe337ae07f44baff34aa)

Issue #10374, opened 2016-01-01
Adhere to XDG Base Directory Specification


Mailaender:
http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

pchote:
How do you propose our directory structure look to adhere to the specification?

Mailaender:
Maybe:

* `$XDG_DATA_HOME/openra/maps`
* `$XDG_DATA_HOME/openra/mods`
* `$XDG_DATA_HOME/openra/replays`
* `$XDG_DATA_HOME/openra/content`

* `$XDG_CONFIG_HOME/openra/settings.yaml`
* `$XDG_CONFIG_HOME/openra/motd.txt`

* `$XDG_CACHE_HOME/openra/logs`

pchote:
According to https://github.com/mono/mono/blob/master/mcs/class/corlib/System/Environment.cs#L619 we want to use`SpecialFolder.ApplicationData` and `SpecialFolder.LocalApplicationData` (if we want split config/data) on Linux.

We should also change to `SpecialFolder.CommonApplicationData` on Windows – this maps to `C:/ProgramData` on recent versions, and is shared for all users.  It maps to `/usr/share` on Linux, so it isn't usable there.

Mailaender:
Actually if you do this right, you can use `SpecialFolder.CommonApplicationData` and even properly adhere to FHS. We just have to remove the hardcoded shipped `./mods` folder which needs to be located in `SpecialFolder.CommonApplicationData` then, but nothing else.

pchote:
Adding to the release milestone because this fits so well with the other asset installation changes that it would be silly to not include it.

pchote:
We need root privileges to create the shared data directory (in `/Library/Application Support`) on OSX, which means completely redoing our packaging there.  That's too big a job for the next release, so removing from the milestone.

pchote:
> Actually if you do this right, you can use `SpecialFolder.CommonApplicationData` and even properly adhere to FHS. We just have to remove the hardcoded shipped `./mods` folder which needs to be located in `SpecialFolder.CommonApplicationData` then, but nothing else.
> - `/usr/share/openra/mods/ra` etc.
> - `/usr/lib/openra/OpenRA.Game.exe` etc.

The `Engine.ModSearchPaths` parameter added in #12638 solves this.

Mailaender:
[...]

PR #10882, opened 2016-03-05
Test the main game directory's mod subdirectory for write permissions

penev92:
And add the result to the system information query.

To (eventually) settle the argument that took place on IRC tonight.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
@@ -257,7 +257,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			if (!Game.Settings.Debug.SendSystemInformation)
 				return null;
 
-			return "?id={0}&platform={1}&os={2}&runtime={3}&gl={4}&lang={5}&version={6}&mod={7}&modversion={8}".F(
+			return "?id={0}&platform={1}&os={2}&runtime={3}&gl={4}&lang={5}&version={6}&mod={7}&modversion={8}&haswritepermission={9}".F(
 				Uri.EscapeUriString(Game.Settings.Debug.UUID),
 				Uri.EscapeUriString(Platform.CurrentPlatform.ToString()),
 				Uri.EscapeUriString(Environment.OSVersion.ToString()),
@@ -266,7 +266,21 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				Uri.EscapeUriString(System.Globalization.CultureInfo.InstalledUICulture.TwoLetterISOLanguageName),
 				Uri.EscapeUriString(ModMetadata.AllMods["modchooser"].Version),
 				Uri.EscapeUriString(Game.ModData.Manifest.Mod.Id),
-				Uri.EscapeUriString(Game.ModData.Manifest.Mod.Version));
+				Uri.EscapeUriString(Game.ModData.Manifest.Mod.Version),
+				Uri.EscapeUriString(TestWritePermissions().ToString()));
+		}
+
+		byte TestWritePermissions()
+		{
+			try
+			{
+				using (var fs = File.Create(Path.Combine(".", "mods", ".testwritable"), 1, FileOptions.DeleteOnClose))
+					return 1;
+			}
+			catch
+			{
+				return 0;
+			}
 		}
 
 		void SetNewsStatus(string message)


OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
```diff
@@ -266,7 +266,21 @@ string SysInfoQuery()
...
-				Uri.EscapeUriString(Game.ModData.Manifest.Mod.Version));
+				Uri.EscapeUriString(Game.ModData.Manifest.Mod.Version),
+				Uri.EscapeUriString(TestWritePermissions().ToString()));
+		}
+
+		byte TestWritePermissions()
+		{
+			try
+			{
+				using (var fs = File.Create(Path.Combine(".", "mods", ".testwritable"), 1, FileOptions.DeleteOnClose))
```
pchote: To settle the argument you will also need to show that Windows isn't redirecting the write to the VirtualStore directory.

GraionDilach:
:+1:

This will prove me valid in the long term anyway.

pchote:
Note to others: The context of this argument is that @penev92 and @GraionDilach strongly believe that we should expose UI allowing users to install content (mods/maps) into the OpenRA install directory because there is a subset of users who either run the game with admin privileges, or who don't know / are happy to rely on Windows's VirtualStore write redirection.

GraionDilach:
My point is that most players on Windows will install their game to another partition and not to the default Program Files because it's a game and not an application.

penev92:
Let's stay on topic. This is strictly for collecting information **on the basis of which** we can later argue.

Also allowing that might be terrible idea, but no worse than what we do know.

OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
Mailaender: Why not use `Directory.GetAccessControl`?

Mailaender:
I assume the Windows VirtualStore write redirection is a workaround for legacy applications that don't have the ability to put configuration files at proper user directories. It is not quite clear to me what this is trying to audit and why we need it. However a check and a user friendly message like "don't run OpenRA with administrator rights" or "C:\ProgramFiles\OpenRA\Support\ isn't writable" might help us have less obvious and repetitive support cases where people confuse this with a 1995 era game. It seems what Windows users try when things don't work is reinstall, run as admin or compatibility mode when instead they should check their graphics drivers or other common problems from the FAQ.

obrakmann:
I think the discussion here is missing the point a bit. Graion's intention is to not install mods onto his (smaller, presumably) C: partition, where the user's support directory would be located by default. His solution is to allow installing mods into the game directory, whose location the user can customize during installation. Of course, as mentioned by pchote, that solution is horrible. Users shouldn't mess with the main installation.

The proper solution, in my eyes, would be to allow the user to specify the location of the user's support directory as well. This is currently not possible, afaik. Then people would be able to move their stuff from the C: drive to whatever other drive they wish.

pchote:
I'd be all for replacing the hardcoded game-dir support dir override with a more general commandline option to specify the support dir.  People on windows who want a portable or custom install can then edit their game shortcut (or we could even have the game installer write custom shortcuts) to use it.

obrakmann:
After thinking some more about it, an even better solution would be to split the support dir into a "Settings" and a "Content" directory. The former would contain settings.yaml, news.yaml, motd.txt and the Logs directory and be located in the same place where the Support dir currently is. The content directory would contain the Content, mods, maps and Replays directories. Its default location could be the same place as the settings directory, or somewhere else, TBD.

The advantage would be that we can just use a normal game settings property that gets saved to settings.yaml to change the support directory and won't have to do any gymnastics with command-line parameters and editing shortcuts. Also something similar was already planned for Linux: #10374.

GraionDilach:
>I'd be all for replacing the hardcoded game-dir support dir override with a more general commandline option to specify the support dir. People on windows who want a portable or custom install can then edit their game shortcut (or we could even have the game installer write custom shortcuts) to use it.

This makes the modchooser redundant and switching mods on the MP screen a mess.

However specifying additional such custom Content folders and then listing them in the settings.yaml sounds like a solution. The game could drop the invalidated ones during launch then.

pchote:
The ability to specify the location of the support directory is completely unrelated to the mod chooser.

GraionDilach:
It isn't. Specifying a custom Content folder for mods with shortcut B while having all the other mod contents in default location under shortcut A will crash the game if you want to switch from one affected mod of shortcut A to one under shortcut B, and will also mess up the remote MP launching as well, since you cannot rely on command-line arguments on that.

obrakmann:
There would only be one Content folder, not additional ones.

pchote:
The support directory is a single location that is used for *all* user-created / user-installed content (settings, log files, maps, replays, mods).  It doesn't make sense to talk about different directories for different mods.

pchote:
> mess up the remote MP launching as well, since you cannot rely on command-line arguments on that.

Under windows the support dir location chosen during installation could be stored in the registry, and then the command-line argument would be provided by the windows-specific launcher.

Phrohdoh:
If this is merged are the people for it going to be ok with its removal after significant data is gathered?
I see no reason to keep this around in the long-term.

LipkeGu:
We could implement an option in the Installer for specifying the Location of the Content directory :) point it by default to the Documents Directory ...

HenrytheSlav:
@GraionDilach 
> My point is that most players on Windows will install their game to another partition and not to the > default Program Files because it's a game and not an application.

Sorry, but almost all of users I met install everything into Program Files because they either believe that this is the correct folder to contain, well, program files, or don't have another partition (notebooks). If we are talking common users here of course.

reaperrr:
Apart from some discussion there hasn't been any progress in quite a while, so I'm closing this for now to clear the PR queue. @penev92 just reopen when you get back to this.

PR #10886, opened 2016-03-06
Fix GravityBomb crash when Image is null

### LINKED ISSUE #10883: GravityBomb cannot lack an image.
It crashes when firing the default value of null. The common projectiles (Missile and Bullet) do not.

reaperrr:
Additionally,
- changed impact check to use `DistanceAboveTerrain` instead of `.Z`
- made shadow palette configurable

Closes  #10883.

### DIFF
--- a/OpenRA.Mods.Common/Effects/GravityBomb.cs
+++ b/OpenRA.Mods.Common/Effects/GravityBomb.cs
@@ -31,6 +31,8 @@ namespace OpenRA.Mods.Common.Effects
 
 		public readonly bool Shadow = false;
 
+		[PaletteReference] public readonly string ShadowPalette = "shadow";
+
 		public readonly WDist Velocity = WDist.Zero;
 
 		[Desc("Value added to velocity every tick.")]
@@ -95,7 +97,7 @@ namespace OpenRA.Mods.Common.Effects
 				{
 					var dat = world.Map.DistanceAboveTerrain(pos);
 					var shadowPos = pos - new WVec(0, 0, dat.Length);
-					foreach (var r in anim.Render(shadowPos, wr.Palette("shadow")))
+					foreach (var r in anim.Render(shadowPos, wr.Palette(info.ShadowPalette)))
 						yield return r;
 				}
 
--- a/OpenRA.Mods.Common/Effects/GravityBomb.cs
+++ b/OpenRA.Mods.Common/Effects/GravityBomb.cs
@@ -56,10 +56,10 @@ namespace OpenRA.Mods.Common.Effects
 			velocity = new WVec(WDist.Zero, WDist.Zero, -info.Velocity);
 			acceleration = new WVec(WDist.Zero, WDist.Zero, info.Acceleration);
 
-			anim = new Animation(args.SourceActor.World, info.Image);
-
 			if (!string.IsNullOrEmpty(info.Image))
 			{
+				anim = new Animation(args.SourceActor.World, info.Image);
+
 				if (!string.IsNullOrEmpty(info.OpenSequence))
 					anim.PlayThen(info.OpenSequence, () => anim.PlayRepeating(info.Sequence));
 				else
@@ -85,6 +85,9 @@ namespace OpenRA.Mods.Common.Effects
 
 		public IEnumerable<IRenderable> Render(WorldRenderer wr)
 		{
+			if (anim == null)
+				yield break;
+
 			var world = args.SourceActor.World;
 			if (!world.FogObscures(pos))
 			{


pchote:
Minor scope creep if you feel like it: we had a request in IRC about allowing the bombs to fall "forwards" instead of always vertical.  We could support this by changing Velocity to a vector in local world coordinates, rotate it by the facing on creation, and then apply this 3D vector each tick.

obrakmann:
Just a note: putting forward velocity on bombs will require setting the appropriate range on the weapon because the actor will need to release the bombs earlier to actually hit its target. Even then there might be situations where the actor will still miss its target on height-enabled maps. Ideally the attack code would compute the perfect release point from the distance to the target, height of the actor and the velocity of the bombs. But I don't think we even have a proper interface to do that atm.

reaperrr:
I might give it a try, but I'd prefer to do that in a separate PR. I'm inclined against delaying fixes due to feature scope-creep, I just threw in the shadow palette commit because it was so easy to do.

OpenRA.Mods.Common/Effects/GravityBomb.cs
```diff
@@ -92,7 +95,7 @@ public IEnumerable<IRenderable> Render(WorldRenderer wr)
 				{
 					var dat = world.Map.DistanceAboveTerrain(pos);
 					var shadowPos = pos - new WVec(0, 0, dat.Length);
-					foreach (var r in anim.Render(shadowPos, wr.Palette("shadow")))
+					foreach (var r in anim.Render(shadowPos, wr.Palette(info.ShadowPalette)))
```
pchote: Looks like you need a null-check here too, otherwise you're just shifting the anim NRE later.

reaperrr:
Updated, added `yield break` when `anim == null`.

OpenRA.Mods.Common/Effects/GravityBomb.cs
```diff
@@ -72,9 +74,10 @@ public void Tick(World world)
 			velocity -= acceleration;
 			pos += velocity;
 
-			if (pos.Z <= args.PassiveTarget.Z)
+			var dat = world.Map.DistanceAboveTerrain(pos);
+			if (dat.Length <= 0)
```
abcdefg30: So we are assuming `args.PassiveTarget.Z == 0`?
Wouldn't we want `if (dat.Length <= world.Map.DistanceAboveTerrain(args.PassiveTarget))` here?
E.g. what happens when we try to hit an air unit (with a lower cruise altitude) with an gravity bomb?

pchote: Actually, I think the original code was already correct here (assuming the PassiveTarget is not underground).

abcdefg30: > DAT is not needed assuming the PassiveTarget is not underground

Well, then we should fix that as we're already here, IMHO.

reaperrr: So, if I understand correctly I should revert/remove my 2nd commit?

pchote: The best option, as @abcdefg30 pointed out, would be to check both `pos.Z <= args.PassiveTarget.Z` *and* `pos.Z <= args.PassiveTarget.Z`

Phrohdoh: @pchote Not sure what you mean, those are the exact same checks.

pchote: Not if the PassiveTarget is below the ground (subterranean unit).

reaperrr: @pchote But your examples are literally exactly the same...?

pchote: err... look again.

reaperrr:
Updated.
I removed the Z check changes for now, I decided to split that to an upcoming PR to get the image fix in asap.

Reasoning: The PassiveTarget.Z check means that if the target was an aircraft, the bombs would explode mid-air even if the target actor moved away in the meantime. That's crap, in my opinion. The correct behavior should be to always fall to the ground, unless blocked by an actor, so in my opinion DAT check + blockability (and making aircraft blockers) should be the way to go. But that's out of the scope of the original "quick fix" nature of this PR.

pchote:
:+1: Looks good to me now, and works ingame.  Thanks.

GraionDilach:
Yep, looks good. :+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/485ff5850bcf02cdd24895e15c53b06d349f562a...78fa3f32ebe24fea97215b8908c8ce0bdbf36d2e)

PR #10889, opened 2016-03-07
Add SoScared to AUTHORS

Micr0Bit:
should not have to explain why ... thank you @SoScared for :

https://www.youtube.com/watch?v=ueosN-JFqG0
http://www.twitch.tv/soscared
https://www.youtube.com/user/SoScared101
http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19414
http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19354


SoScared:
Thanks! Really appreciate that. Also cool to see my full name on that list :)

PR #10893, opened 2016-03-07
Fix an exploit were players were able to repair at enemy service depots

abcdefg30:
Just ping me in IRC if you want to know how to trigger the exploit (I'm not sure if I should post that in public, although it's not too game breaking).


obrakmann:
Yup, fixed the bug. :+1:

pchote:
I'm not sure how to trigger the exploit, but the code change looks obviously correct (and if it breaks anything, then the broken code was already bogus) so :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/485ff5850bcf02cdd24895e15c53b06d349f562a

PR #10895, opened 2016-03-07
Make mammoth tanks capable of crushing concrete walls

abcdefg30:
Like the do in TD.

CC @SoScared, @kyrylo (as this was suggested on stream, too).

### DIFF
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -165,7 +165,7 @@ V2RL:
 		Type: Heavy
 	Mobile:
 		Speed: 42
-		Crushes: wall, mine, crate, infantry
+		Crushes: wall, mine, crate, infantry, heavywall
 	RevealsShroud:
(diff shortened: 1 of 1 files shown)


SoScared:
Awesome! This is gonna make Mammoth tank a really fun unit to play. Especially on late-game team matches. :+1:

kyrylo:
This idea has been floating around for ages (or months) :)
This is a big improvement for mammoths, but I am not totally sure it's enough.

MustaphaTR:
:+1: That would be better if we make them crush light vehicles too as in my maps, but this is a start.

obrakmann:
Well, neither walls nor mammoths are used much at all from what I can see, so I doubt this change would make any big impact.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/861e135cfb795d2a93087837056d56ca4333a132)

Issue #10718, opened 2016-02-07
[RA] The official map pool. Altering existing maps.


SoScared:
I encourage you to give specific feedback/contribututions to: 
**The Official Map Pool - collaboration thread** 
http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&p=294875#294875

I've gone through the existing RA maps and noted specific changes that IMO needs to be done. The changes are partially aesthetic and some balance related. This may or may not give the map new life but in any case they will look much better in the map chooser and in-game. If we are to hold on to the classic maps, they should at least be up to par with competing maps on aesthetics and balance. I've done my best to preserve as much of the maps' original character with the recommended changes.

Separate issues will eventually be submitted for each map for a final review and chance for additional discussion. All maps have been thoroughly playtested and altered after feedback from players before being submitted individually.

Map and author names should stay unchanged unless the map is altered a lot beyond it's original structure.

**Map list:**

Screenshots and comparisons are found on the altered maps' pages.

[**Forest Path**](http://resource.openra.net/maps/7295/) (2 players)
Move player spawns one cell away horizontally from the ore patches.
Change layout of actors around secondary ore patches to equalize access for both spawns.
Switch out actors (trees) to vary the fauna.
[Altered](http://resource.openra.net/maps/9138)
**Playtested, submitted at #10928**

[**Keep Off The Grass 2**](http://resource.openra.net/maps/7242/) (2 players)
Move the first ore patch for both spawns to balance ore access.
[Altered](http://resource.openra.net/maps/9008/)
also see #10347 
**Playtested, submitted at #10954**

[**Man to Man**](http://resource.openra.net/maps/7196/) (2 players)
Remove oil derricks -or- access them by bridge.
[Altered](http://resource.openra.net/maps/9230/)
also see #10312
**Playtested, submitted at #10953**

[**Sidestep**](http://resource.openra.net/maps/8176/) (2 players)
Add edge ridges for more distinction.
Add more roads, actors and debris to make the map less blank.
[Altered](http://resource.openra.net/maps/9353/)
**Playtested, submitted at #10927**

[**Singles**](http://resource.openra.net/maps/7353/) (2 players)
Map structure overhaul. see #10301
Work on balance and aesthetics.
[Altered](http://resource.openra.net/maps/9487)
**Playtested, submitted at #10987**

[**Snowy Ridge**](http://resource.openra.net/maps/7996/) (3 players)
Move the first ore patch for all spawns to balance ore access.
[...]

Issue #9684, opened 2015-10-19
Allow maps to define custom tileset additions


pchote:
http://resource.openra.net/maps/6132 has some impressive custom terrain, but relies on a really horrible hack (see http://www.sleipnirstuff.com/forum/viewtopic.php?f=83&t=18951) to do it.  It would be great if we could support this properly, but this is going to be fiddy to do without breaking our existing caching.

MicroJOo:
![something](https://cloud.githubusercontent.com/assets/9139965/10594259/80586f2a-76cb-11e5-9b4c-648b28882973.png)

MicroJOo:
this hack doesnt work anymore .. and broke on of my maps ...
i used this to get a black floor , but the ZOffset doesnt work the way it did before 

![berz](https://cloud.githubusercontent.com/assets/9139965/12161588/9d91052e-b4f7-11e5-9ded-a64ec51f2030.png)

pchote:
The work i'm doing on the filesystem and map init code will have the side effect (once finished) of making this easy to implement.  I plan on wrapping up those changes by fixing this.

pchote:
I'm tentatively adding this to the release milestone for the same reasons as #4251, but i'm skeptical that I will be able to get all the prerequisites done and reviewed in time for the first playtest.

pchote:
My skepticism was right.  Bumping to Next + 1.

PR #10898, opened 2016-03-08
Add support for multiple map categories.

pchote:
This is the final backwards-incompatible change that I'd like to propose for the next release.  I've split it into its own PR because it depends on resource center changes that may not happen immediately (it needs to be able to parse and expose the categories via the map API).

This replaces the map `Type` field with a list of `Categories`, where the first category is treated as the main map type, and remaining categories are used for filtering the map chooser.  As a test case I have updated and added categories for the TS maps; this will currently break the other mods (updating those now would just end up with rebase conflicts while we wait for the resource center changes).

Depends on #10839. Only the last four commits are new.


GraionDilach:
Map importers won't spit out format 11 maps.

obrakmann:
```
OpenRA.Game/Map/MapPreview.cs:L247: [SA1026] There should be no space between the new keyword and the opening square bracket in an implicitly typed array allocation.

OpenRA.Game/Map/MapPreview.cs:L247: [SA1010] Invalid spacing around the opening square bracket.
```

pchote:
Adding the dependencies tag back, since this requires changes to the resource center API.

ihptru:
Resource Center implemented support at https://github.com/ihptru/OpenRA-Resources/commit/d52495495e0a1675044a83f948d586677270fe78

It's in production now.

pchote:
Updated.

OpenRA.Mods.Common/Widgets/Logic/ReplayBrowserLogic.cs
```diff
@@ -83,7 +83,10 @@ public ReplayBrowserLogic(Widget widget, ModData modData, Action onExit, Action
 
 			var type = panel.GetOrNull<LabelWidget>("MAP_TYPE");
 			if (type != null)
-				type.GetText = () => selectedReplay.GameInfo.MapPreview.Type;
+			{
+				var mapType = selectedReplay.GameInfo.MapPreview.Categories.FirstOrDefault() ?? "";
```
obrakmann: This crashes with an NRE.

pchote:
Fixed and rebased.

obrakmann:
Looks fine, :+1: ,but already needs another rebase.

pchote:
Rebased and changed LobbyMapPreviewLogic a bit.

pchote:
This pr is blocking fixes for #10718 and #9684.  We need to merge it before we can make more progress towards a playtest...

penev92:
:+1:

GraionDilach:
eh... eh.

I ended up being ignored here. The map importers were untouched.

pchote:
The map importers don't need to be touched...?

GraionDilach:
Oh, those are set in the map constructors and not in the actual map importers, sorry... dammit, I shouldn't review/catch-up stuff when I'm sleepy.

penev92:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c1fb7ea215fbc91ea4afaeed16a9e982cb60e8f4)

PR #10900, opened 2016-03-09
Allow multiple AttackTurreted traits.

GraionDilach:
This is something really important for a faction of mine (+ various random other units).

Public testcase will come soon in the form of the RA2 IFV.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
@@ -69,7 +69,7 @@ namespace OpenRA.Mods.Common.Traits
 
 			public AttackActivity(Actor self, Target target, bool allowMove, bool forceAttack)
 			{
-				attack = self.Trait<AttackFollow>();
+				attack = self.TraitsImplementing<AttackFollow>().Where(af => !af.IsTraitDisabled).First();
 				move = allowMove ? self.TraitOrDefault<IMove>() : null;
 
 				// HACK: Mobile.OnRails is horrible. Blergh.
--- a/OpenRA.Mods.Common/Traits/Attack/AttackTurreted.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackTurreted.cs
@@ -20,12 +20,14 @@ namespace OpenRA.Mods.Common.Traits
 		public override object Create(ActorInitializer init) { return new AttackTurreted(init.Self, this); }
 	}
 
-	public class AttackTurreted : AttackFollow
+	public class AttackTurreted : AttackFollow, INotifyCreated
 	{
 		protected Turreted[] turrets;
 
 		public AttackTurreted(Actor self, AttackTurretedInfo info)
-			: base(self, info)
+			: base(self, info) { }
+
+		public void Created(Actor self)
 		{
 			turrets = self.TraitsImplementing<Turreted>().ToArray();
 		}
@@ -35,11 +37,13 @@ namespace OpenRA.Mods.Common.Traits
 			if (!base.CanAttack(self, target))
 				return false;
 
+			// Rotate all turrets to make sure upgrading a turret does not change the turret's facing.
+			var faceTarget = false;
 			foreach (var t in turrets)
-				if (t.FaceTarget(self, target))
-					return true;
+				if (t.FaceTarget(self, target) && !t.IsTraitDisabled)
+					faceTarget = true;
 
-			return false;
+			return faceTarget;
 		}
 	}
 }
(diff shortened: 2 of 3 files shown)


OpenRA.Mods.Common/Traits/Attack/AttackTurreted.cs
```diff
@@ -20,12 +20,14 @@ public class AttackTurretedInfo : AttackFollowInfo, Requires<TurretedInfo>
...
-	public class AttackTurreted : AttackFollow
+	public class AttackTurreted : AttackFollow, INotifyCreated
 	{
 		protected Turreted[] turrets;
 
 		public AttackTurreted(Actor self, AttackTurretedInfo info)
-			: base(self, info)
+			: base(self, info) { }
+
+		public void Created(Actor self)
```
pchote: The `Requires<TurretedInfo>` ensures that the `Turreted` traits have already been initialized before this one, so there should be no problem with keeping this query in the constructor.

GraionDilach: As I explained it over IRC - it shouldn't be but it does. `Turreted` traits defined after `AttackTurreted` (if there's a `Turreted` trait before the `AttackTurreted` as well) will don't get their facing updated later when disabled.

pchote: Yes.  I'm mentioning it again here because if you are correct, then we have may have a serious regression in our actor creation code that will need to be investigated and fixed.

OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -85,7 +87,7 @@ public virtual void Tick(Actor self)
 			// NOTE: FaceTarget is called in AttackTurreted.CanAttack if the turret has a target.
 			if (attack != null)
```
obrakmann: `attack` won't ever be null anymore, so this can go.

OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -85,7 +87,7 @@ public virtual void Tick(Actor self)
...
 			{
-				if (!attack.IsAttacking)
+				if (attack.Where(a => a.IsAttacking && !a.IsTraitDisabled).Count() == 0)
```
obrakmann: `if (!attack.Any(a => a.IsAttacking && !a.IsTraitDisabled))`

OpenRA.Mods.Common/Traits/Attack/AttackTurreted.cs
```diff
@@ -35,11 +37,13 @@ protected override bool CanAttack(Actor self, Target target)
 			if (!base.CanAttack(self, target))
 				return false;
 
+			// Rotate all turrets to make sure upgrading a turret does not change the turret's facing.
+			var faceTarget = false;
 			foreach (var t in turrets)
-				if (t.FaceTarget(self, target))
-					return true;
+				if (t.FaceTarget(self, target) && !t.IsTraitDisabled)
```
obrakmann: Be aware of `FaceTarget`'s side effects. Do you want the turret to turn towards the target even when that turret is disabled? Otherwise the `!t.IsTraitDisabled` needs to be the first check in the condition.

OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
```diff
@@ -69,7 +69,7 @@ class AttackActivity : Activity
 
 			public AttackActivity(Actor self, Target target, bool allowMove, bool forceAttack)
 			{
-				attack = self.Trait<AttackFollow>();
+				attack = self.TraitsImplementing<AttackFollow>().Where(af => !af.IsTraitDisabled).First();
```
obrakmann: `attack = self.TraitsImplementing<AttackFollow>().First(af => !af.IsTraitDisabled)`

Also, this will crash if they're all disabled. Intentional?

OpenRA.Mods.Common/Traits/Attack/AttackTurreted.cs
GraionDilach: That's the point. Any other way and turretchanging would look odd (say IFV targets something, passenger exits and turret suddenly looks somewhere else).

It's also a bit workaroundish because turrets and `AttackTurret` is only paired at `Armament` level which `FaceTarget` doesn't check for, allowing the facings to go out of sync. (Although I have fixed that since, notsure.)

OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
GraionDilach: Definitely not.

GraionDilach:
Closed because of missing `AutoTarget` support (which is a practical prerequisite for using this), and #10907.

PR #10902, opened 2016-03-09
Restore the ability for maps to include inline rules.

pchote:
Removing support for inline yaml rules etc was a little premature: we need this feature for the server-resource center integration.  It therefore makes sense to bring this back for regular maps too.

This mostly reverts the changes to the importers and format upgrader to keep the custom rules in map.yaml.  The TS lighting definitions have been moved back as a testcase because it was a little silly to have a 3-4 line single-trait override in its own file.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckSequences.cs
+++ b/OpenRA.Mods.Common/Lint/CheckSequences.cs
@@ -26,18 +26,17 @@ namespace OpenRA.Mods.Common.Lint
 
 		public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 		{
-			if (map != null && !map.SequenceDefinitions.Any())
+			if (map.SequenceDefinitions == null)
 				return;
 
 			var modData = Game.ModData;
 			this.emitError = emitError;
 
-			var mapSequences = map != null ? map.SequenceDefinitions : new string[0];
-			sequenceDefinitions = MiniYaml.Merge(modData.Manifest.Sequences.Append(mapSequences).Select(s => MiniYaml.FromStream(map.Open(s))));
+			sequenceDefinitions = MiniYaml.Load(map, modData.Manifest.Sequences, map.SequenceDefinitions);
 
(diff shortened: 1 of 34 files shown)


OpenRA.Mods.Common/Lint/CheckSequences.cs
```diff
@@ -26,14 +26,13 @@ class CheckSequences : ILintMapPass
 
 		public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 		{
-			if (map != null && !map.SequenceDefinitions.Any())
+			if (map != null || map.SequenceDefinitions == null)
```
obrakmann: `map == null || map.SequenceDefinitions == null`

Caught an NRE before Coverity for once! \o/

GraionDilach:
Does this means additional yamls are optional or impossible?

pchote:
Optional, and recommended if you have anything more than trivial overrides.

obrakmann:
OK, lgtm, aside from that one inverted check, :+1:

pchote:
Fixed.

mods/ts/maps/t_garden/map.yaml
```diff
@@ -830,4 +830,10 @@ Actors:
 		Location: 99,74
 		Owner: Neutral
 
-Rules: rules.yaml
+Rules:
+	World:
+		GlobalLightingPaletteEffect:
+			Red: 1
```
GraionDilach: Could these be deleted as part of this PR? 1 is default.

pchote:
Done.

abcdefg30:
Looks good to me, too. :+1:  / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3cf3d244478caf0a8fb0ff4c2c4822ee9211f0d8...bf3d3b6fc5aff3e9809a3b60f89c0e24d610be77)

Issue #10878, opened 2016-03-04
Theoretical NullReferenceException in GameInfoStatsLogic


RoosterDragon:
Discovered by Coverity.

There are null checks in `GameInfoStatsLogic` to guard against `client` being null. Previous feature, bug and fix for context: #6360 #6444, #6450.

As of #10067, `client` is now directly dereferenced [here](https://github.com/OpenRA/OpenRA/blob/aa78dc32e391acdf95de879a26d15509429a8a3b/OpenRA.Mods.Common/Widgets/Logic/Ingame/GameInfoStatsLogic.cs#L55) (and the called method also unconditionally deferences it).

I can't grok enough of it to decide what the fix is. Either we can drop the null checks because they're redundant now - or null is still possible in which case the line I called out needs to handle that somehow (but I have no idea how).

pchote:
The main difference is that it only gets called once during setup there, and client shouldn't be null.  We can probably get away with squashing the error by adding an `if (client != null)` check before calling that method: if something is bogus enough for client to be null then they simply won't get a tooltip.

RoosterDragon:
Fixed in #11079.

PR #10905, opened 2016-03-10
Fix coverity issues

obrakmann:


OpenRA.Mods.Common/Widgets/Logic/Ingame/GameInfoStatsLogic.cs
```diff
@@ -85,7 +85,7 @@ public GameInfoStatsLogic(Widget widget, World world, OrderManager orderManager)
 				}
 
 				var team = item.Get<LabelWidget>("TEAM");
-				var teamNumber = pp.PlayerReference.Playable ? ((client == null) ? 0 : client.Team) : pp.PlayerReference.Team;
+				var teamNumber = pp.PlayerReference.Playable ? client.Team : pp.PlayerReference.Team;
```
obrakmann: `client` already gets de-referenced [earlier](https://github.com/OpenRA/OpenRA/pull/10905/files#diff-d371faa3132142c9bb8b07d783a5ea38R53), so if it was null, we'd never get here anyway.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -652,7 +652,7 @@ int IncreaseAltitude(int predClfDist, int diffClfMslHgt, int relTarHorDist, int
 					else
 					{
 						// Aim for the target
-						var vDist = new WVec(-relTarHgt, -relTarHorDist * (targetPassedBy ? -1 : 1), 0);
+						var vDist = new WVec(-relTarHgt, -relTarHorDist, 0);
```
abcdefg30: Wouldn't that mean that `targetPassedBy` is always `true` when we reach this code? In that case you need to keep the `* -1`, i.e. just remove the `-` infront of `relTarHorDist`.

OpenRA.Game/Map/MapPreview.cs
```diff
@@ -323,7 +323,8 @@ public void Install()
 						Game.RunAfterTick(() =>
 						{
 							var package = modData.ModFiles.OpenPackage(mapFilename, mapInstallPackage);
-							UpdateFromMap(package, mapInstallPackage, MapClassification.User, null, GridType);
+							if (package != null)
+								UpdateFromMap(package, mapInstallPackage, MapClassification.User, null, GridType);
```
RoosterDragon: Under what circumstances might the package possibly be null after we downloaded the map? If it is null, should we not report some sort of error?

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -652,7 +652,7 @@ int IncreaseAltitude(int predClfDist, int diffClfMslHgt, int relTarHorDist, int
 					else
 					{
 						// Aim for the target
-						var vDist = new WVec(-relTarHgt, -relTarHorDist * (targetPassedBy ? -1 : 1), 0);
+						var vDist = new WVec(-relTarHgt, relTarHorDist, 0);
 						desiredVFacing = (sbyte)vDist.HorizontalLengthSquared != 0 ? vDist.Yaw.Facing : vFacing;
 						if (desiredVFacing < 0 && info.VerticalRateOfTurn < (sbyte)vFacing)
 							desiredVFacing = 0;
```
RoosterDragon: This duplication would be better tackled by extracting a common method.

OpenRA.Mods.Common/Widgets/Logic/Ingame/GameInfoStatsLogic.cs
RoosterDragon: See #10878 - it's possible this indicates a missing null check earlier (I don't know enough to say either way).

RoosterDragon: There's also some other checks for a null client you would need to remove, if you do believe the null check is redundant.

OpenRA.Game/FileFormats/PngLoader.cs
```diff
@@ -111,6 +111,9 @@ public static Bitmap Load(Stream s)
 
 								case "IEND":
 									{
+										if (bitmap == null)
+											throw new InvalidDataException("Image header not found.");
```
RoosterDragon: There is a check [a bit further up](https://github.com/OpenRA/OpenRA/pull/10905/files#diff-d982fdbe9f03fe9111a8f46963a257abR55) for this. Either I've messed it up and we should fix that check, or the static analysis isn't clever enough to realise it can't get to here with a null bitmap. We should try and fix that check rather than inserting redundant code.

OpenRA.Mods.Common/Widgets/Logic/Ingame/GameInfoStatsLogic.cs
obrakmann: My assumption was that `client` can't be null once we're in a game, but #6444 seems to prove me wrong. I'll add null check before calling `SetupClientWidget` instead.

OpenRA.Game/FileFormats/PngLoader.cs
obrakmann: Looks like it's not clever enough. I'll drop this and mark it as a false positive.

OpenRA.Game/Map/MapPreview.cs
obrakmann: Possibly. Either way, `OpenPackage` returns null as a last resort. I'm fine with saying 'this will never happen' and ignoring the issue.

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -673,16 +666,29 @@ int IncreaseAltitude(int predClfDist, int diffClfMslHgt, int relTarHorDist, int
...
-				desiredVFacing = desiredVFacing.Clamp(-info.VerticalRateOfTurn, info.VerticalRateOfTurn);
-
+				desiredVFacing = CalculateDesiredVFacingForAscent(vFacing, diffClfMslHgt);
 				ChangeSpeed();
 			}
 
 			return desiredVFacing;
 		}
 
+		int CalculateDesiredVFacingForDescent(WVec vDist, int vFacing)
```
obrakmann: Those method names are most likely wrong. If anyone has better suggestions, have at it.

OpenRA.Mods.Common/Widgets/Logic/Ingame/GameInfoStatsLogic.cs
```diff
@@ -52,7 +52,8 @@ public GameInfoStatsLogic(Widget widget, World world, OrderManager orderManager)
 				var pp = p;
 				var client = world.LobbyInfo.ClientWithIndex(pp.ClientIndex);
 				var item = playerTemplate.Clone();
-				LobbyUtils.SetupClientWidget(item, client, orderManager, client.Bot == null);
+				if (client != null)
+					LobbyUtils.SetupClientWidget(item, client, orderManager, client.Bot == null);
```
RoosterDragon: `SetupClientWidget` sets the visibility of some widgets. Will these all be visible by default if we skip this call because the client is null? It doesn't look like that would be correct behaviour.

obrakmann:
ah, you know what, it isn't worth it to waste time on this for issues that are largely theoretical.

PR #10910, opened 2016-03-11
Turrets actively track targets

### LINKED ISSUE #5203: Turreted units wait until they are in range of target before aiming turrets
It would make more sense if a turreted unit would rotate its turret around to face the target it is chasing as soon as it gets the order to attack, as opposed to waiting until it drives into range, and only then beginning to slowly rotate its turret to aim. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1863799-turreted-units-wait-until-they-are-in-range-of-target-before-aiming-turrets?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RoosterDragon:
Whilst a turreted unit has a valid target, all turrets will be made to track the target even if it is not ready to attack yet. This means once it becomes ready (e.g. enters range, reloads, etc) it will already be lined up.

Fixes #5203.

### DIFF
--- a/OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
+++ b/OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
@@ -79,7 +79,7 @@ namespace OpenRA.Mods.Cnc.Traits
 				return false;
 			}
 
-			return turret.FaceTarget(self, target);
+			return true;
 		}
 
 		public void TickIdle(Actor self)
--- a/OpenRA.Mods.Common/Traits/Armament.cs
+++ b/OpenRA.Mods.Common/Traits/Armament.cs
@@ -186,6 +186,9 @@ namespace OpenRA.Mods.Common.Traits
 			if (ammoPool != null && !ammoPool.HasAmmo())
 				return null;
 
+			if (turret != null && !turret.HasAchievedDesiredFacing)
+				return null;
+
 			if (!target.IsInRange(self.CenterPosition, MaxRange()))
 				return null;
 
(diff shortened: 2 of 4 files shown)


RoosterDragon:
@reaperrr: Does this also want an `IsAttacking` check per your comment (https://github.com/OpenRA/OpenRA/issues/5203#issuecomment-195404111)?

reaperrr:
Doesn't look like it, this already works quite well. Nice :+1:

pchote:
This isn't quite right: the unit will fire from all turrets, even those that aren't facing the target.  A simple test case is to give the cruiser one fast turret and one slow one.

GraionDilach:
Oh damn, don't tell me #10900 and this should be treated along...

RoosterDragon:
Added another commit to ensure turrets have the correct facing before being allowed to fire.

Mailaender:
Tanks kiting the enemy now behave way better! :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a40d7554b4c030d0c254383ba199bf1a473c5598...0c4999cbd91a63c8312e38c4e2610f30f8b8f306)

obrakmann:
It appears that this broke infantry targeting in a big way.

pchote:
Note that infantry have turrets by way of the `TakeCover` trait.

Mailaender:
That was unexpected for me. Do they need `Turreted`?

PR #10911, opened 2016-03-12
Fix Smoke facings to not track spawner's facing

### LINKED ISSUE #10894: Trails with facings still 'listen' to parent actor/projectile after being spawned
I didn't notice this before, but this is a potential show-stopper.

Basically, if the parent actor/projectile changes facing, not only newly spawned smoke, but also any remaining smoke effects that were spawned eariler instantly change to the same facing.

That's bogus, but I have yet to figure out how to fix it.

reaperrr:
Closes #10894.


GraionDilach:
Yep, it does. :+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/a3400c734aae41264e6062481a60dd509ecb34c3)

PR #10676, opened 2016-02-02
SoundLoader changes to allow streaming in the near future

teees:
This is an intermediate pull request, it's fully functional but not that optimized to avoid too many changes in one pr.

Follow-up PRs will be referenced here, which will add the actual streaming capabilities to `OpenALSoundEngine`, `AudLoader` and `WavLoader`.

`VocLoader` is already fully capable of being streamed.

PR #10912, opened 2016-03-12
Introduce ISoundFormat for parsing sound files.

pchote:
This takes #10676 and changes `MusicInfo` to use the new `LengthInSeconds` property.

The original SoundFormat implementations would load the entire file up-front, which is overkill when we are only interested in the length.  I've worked around that by splitting out proper format detection methods, and then lazily loading the raw bytes only when needed.  This can later be changed to use proper streaming.

### DIFF
--- a/OpenRA.Game/FileFormats/AudLoader.cs
+++ b/OpenRA.Game/FileFormats/AudLoader.cs
@@ -46,12 +46,29 @@ namespace OpenRA.FileFormats
 
 	public class AudLoader : ISoundLoader
 	{
+		bool IsAud(Stream s)
+		{
+			var start = s.Position;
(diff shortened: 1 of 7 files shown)


Mailaender:
No regressions on all mods including RA2. :white_check_mark:

abcdefg30:
Didn't spot anything wrong, too. :+1: / :white_check_mark:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d8f2d77b2a5f55ed38979beb1a30b54d1a77665e...48efba167de7752dcc774b8c3e9a280f41d94a8a)

PR #10919, opened 2016-03-13
Fix some actors not being available for the neutral player in the map editor

### LINKED ISSUE #10780: mpspawn not shown in editor for "Neutral" until switching to different faction and back
Testing latest bleed(just updated).
OS: Debian testing
Mono 4.2.1

What I did:

1) Start the map editor. Create a new map or load it
2) Go to actors
3) Note that the default faction is "Neutral"

What I expected:
mpspawn is one of the actors I can place

What I saw:
mpspawn and some other special actors are not present. If I switch to "Creeps" and then switch back, I do see mpspawn and the other actors present for "Neutral".

The result of this was that at first I thought I could only place mpspawn as Creeps, and that that's what I was supposed to do. Seems clearly a bug if others can reproduce it.

obrakmann:
This is basically a follow-up from #10804, since both issues have the same cause, which just wasn't realized until now. For some reason I didn't put `LoadWidgetAtGameStart` at the end of the list for the `EditorWorld` in that PR, which would have already fixed the issue a while ago.

Fixes #10780.

### DIFF
--- a/mods/cnc/rules/world.yaml
+++ b/mods/cnc/rules/world.yaml
@@ -156,7 +156,7 @@ World:
 
 EditorWorld:
 	Inherits: ^BaseWorld
-	LoadWidgetAtGameStart:
 	EditorActorLayer:
 	EditorResourceLayer:
 	EditorSelectionLayer:
+	LoadWidgetAtGameStart:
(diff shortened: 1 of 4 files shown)


pchote:
:+1: Fix confirmed.  #10804 happened because the player palettes are set up in the `EditorActorLayer`.  If the widgets are loaded first, then the ActorPreview rendering fails for everything that doesn't use the terrain palette, and those actors are dropped from the selector.

Mailaender:
Thanks for investigating and fixing this.

PR #10924, opened 2016-03-14
Fixed AppVeyor after_test depending on a Unix cp command

Mailaender:
Apparently without it the `cp` command relies on a Unix `cp.exe` available from an arbitrary depdendency we were not aware of and it vanishing was the reason for yesterdays failed builds.  Source: http://help.appveyor.com/discussions/problems/4291-cp-fails-on-every-build


abcdefg30:
Well, appveyor passes again. :white_check_mark:

PR #10931, opened 2016-03-15
Fix and overhaul Linux appdata

pchote:
Our appdata.xml contained bogus unicode characters that prevented recent versions of Gnome Software Center from parsing it.  While debugging this issue I discovered the appdata linter, which pointed out a bunch more issues.  This removes all the invalid data, adds proper screenshots, and expands the description text to make the page look a bit more attractive.  This also adds the metadata to our deb packages in preparation for Ubuntu 16.04, and changes the deb maintainer back to me.

screen shot 2016-03-15 at 22 20 27

For some reason I couldn't get OpenRA to display in the software center in openSUSE.  They ship older versions of the software center and appdata tools, but the file passes the lint tests there.  I suspect this may be a side effect of some other problem that prevents the software center from detecting that it is installed.

### DIFF
--- a/packaging/linux/openra.appdata.xml
+++ b/packaging/linux/openra.appdata.xml
@@ -1,48 +1,60 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop">
   <id>openra.desktop</id>
-  <pkgname>openra</pkgname>
-  <name>OpenRA</name>
+  <metadata_license>CC0-1.0</metadata_license>
   <project_license>GPL-3.0</project_license>
+  <name>OpenRA</name>
   <summary>Reimagining of early Westwood real-time strategy games</summary>
-  <developer_name>The OpenRA developers</developer_name>
   <description>
     <p>
-      OpenRA is a Libre/Free Real Time Strategy project
-      that recreates the classic Command &amp; Conquer titles.
+      OpenRA is a Libre/Free Real Time Strategy project that recreates and reimagines
+      the classic Command &amp; Conquer titles.  The primary focus is cross-platform
+      multiplayer between Windows, OS X, and Linux; however, a growing number of the
+      original campaign missions are included, and there is support for skirmish games
+      against AI bots.
     </p>
     <p>
-      It includes recreations of Tiberian Dawn, Red Alert,
-      and Dune 2000. These are not intended to be perfect
-      copies, but instead combine the classic gameplay of
-      the originals with modern improvements such as unit
-      veterancy and the fog of war.
+      In a world where Hitler was assassinated and the Third Reich never existed,
+      the Soviet Union seeks power over all of Europe. Allied against this Evil Empire,
+      the free world faces a Cold War turned hot. The Red Alert mod fuses the quick
+      and fun gameplay of the original C&amp;C: Red Alert, with balance improvements
+      and new gameplay features inspired by modern RTS games.
+    </p>
+    <p>
+      Join the Global Defense Initiative or the Brotherhood of Nod in our recreation of
+      the classic game that started it all! The Tiberian Dawn mod modernizes the original
+      Command &amp; Conquer gameplay by introducing features from later games, including
+      per-factory production queues, unit veterancy, and capturable tech structures.
+    </p>
+    <p>
+      Three great houses fight for the precious spice, melange. He who controls the spice
+      controls the universe! In the Dune 2000 mod you must establish a foothold on the
+      desert planet Arrakis, surviving its harsh environmental conditions while protecting
+      your harvesting operations from giant sandworms and ruthless enemy factions.
     </p>
   </description>
-  <url type="bugtracker">http://bugs.openra.net</url>
-  <url type="donation">https://www.bountysource.com/teams/openra</url>
-  <url type="faq">http://wiki.openra.net/FAQ</url>
-  <url type="help">http://wiki.openra</url>
-  <url type="homepage">http://www.openra</url>
-  <icon type="cached">openra.png</icon>
-  <icon type="remote" width="128" height="138">http://www.openra.net/images/soviet-logo-fallback.png</icon>
-  <icon type="local" width="128" height="128">/usr/share/icons/hicolor/128x128/apps/openra.png</icon>
-  <categories>
-​    <category>Game</category>
-​    <category>StrategyGame</category>
-​  </categories>
   <screenshots>
     <screenshot type="default">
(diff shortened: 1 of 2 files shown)


packaging/linux/openra.appdata.xml
```diff
@@ -1,48 +1,60 @@
...
+      desert planet Arrakis, surviving its harsh environmental conditions while protecting
+      your harvesting operations from giant sandworms and ruthless enemy factions.
     </p>
   </description>
-  <url type="bugtracker">http://bugs.openra.net</url>
-  <url type="donation">https://www.bountysource.com/teams/openra</url>
-  <url type="faq">http://wiki.openra.net/FAQ</url>
-  <url type="help">http://wiki.openra</url>
-  <url type="homepage">http://www.openra</url>
-  <icon type="cached">openra.png</icon>
```
pchote: Before somebody asks: these removed tags, while valid appstream metadata, are not included in the subset of supported appdata tags.  They are either explicitly flagged as errors by the linter, or accepted but do nothing.

packaging/linux/openra.appdata.xml
```diff
@@ -1,48 +1,60 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop">
   <id>openra.desktop</id>
-  <pkgname>openra</pkgname>
-  <name>OpenRA</name>
+  <metadata_license>CC0-1.0</metadata_license>
```
pchote: `CC0-1.0` seems to be the default choice for these metadata files.  Are you okay with this choice @Mailaender?

packaging/linux/buildpackage.sh
```diff
@@ -30,6 +30,10 @@ make install-man-page prefix="/usr" DESTDIR="$PWD/packaging/linux/$ROOTDIR"
 mkdir -p $PWD/packaging/linux/$ROOTDIR/usr/share/doc/openra/
 cp *.html $PWD/packaging/linux/$ROOTDIR/usr/share/doc/openra/
 
+# Appdata
+mkdir -p $PWD/packaging/linux/$ROOTDIR/usr/share/appdata/
+cp $PWD/packaging/linux/openra.appdata.xml $PWD/packaging/linux/$ROOTDIR/usr/share/appdata/
```
obrakmann: This is already handled by the Makefile.

obrakmann:
Doesn't make any difference on 14.04's software center, even after I replaced /usr/share/appdata/openra.appdata.xml with the version from this PR, it still only shows the deb metadata.

packaging/linux/openra.appdata.xml
Mailaender: Technically a software screenshot is licensed under the terms the software is released. https://commons.wikimedia.org/wiki/Template:Free_screenshot As the Westwood / EA owned assets are making up most of the screenshots, those terms apply. If there is nothing for proprietary freeware, then CC0 although incorrect might be the best fit.

packaging/linux/openra.appdata.xml
Mailaender: You mean the icons? What about the URLs? Are they also flagged by Canonical's linter? I guess it doesn't matter as the main website is still linked: https://github.com/OpenRA/OpenRA/pull/10931/files#diff-b92d0fe2ae3f3e132ba0b1d1d4fa643aR58 and that points to everything else.

pchote:
> Doesn't make any difference on 14.04's software center

Indeed.  Canonical are throwing that out for 16.04 in favor of the gnome software center.

packaging/linux/openra.appdata.xml
pchote: The linter is part of the appstream tools, and has nothing to do with canonical.  The additional url types are not flagged but they are not exposed in the UI, and are not defined by any of the other appdata files that I have seen.  This makes me fairly confident that they are not part of the appdata spec.

pchote: I was referring to `<developer_name>`, `<icon>`, and `<categories>` for tags flagged by the linter, and the additional `<url>` types and `<pkgname>` for tags that are not flagged but still don't appear to be part of the spec.  The linter also raised errors about the screenshot sizes not matching the artwork, and the screenshots being too small.

pchote:
Dropped the buildpackage.sh script change.  That was left over from before I discovered the syntax errors.

packaging/linux/openra.appdata.xml
Mailaender: The URL metadata you removed is part of the spec. The Ubuntu/GNOME software center just doesn't implement it. See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url The rest is a part of Debian extensions. https://www.freedesktop.org/software/appstream/docs/sect-AppStream-DEP11.html I don't know where they are used.

pchote: appstream != appdata.  See https://people.freedesktop.org/~hughsient/appdata/.

Mailaender:
```
matthias@mPad-X230:~/Downloads>  appstream-util validate openra.appdata.xml 
openra.appdata.xml: OK
```

:white_check_mark:

packaging/linux/openra.appdata.xml
```diff
@@ -1,48 +1,60 @@
...
-  <url type="donation">https://www.bountysource.com/teams/openra</url>
-  <url type="faq">http://wiki.openra.net/FAQ</url>
-  <url type="help">http://wiki.openra</url>
-  <url type="homepage">http://www.openra</url>
-  <icon type="cached">openra.png</icon>
-  <icon type="remote" width="128" height="138">http://www.openra.net/images/soviet-logo-fallback.png</icon>
-  <icon type="local" width="128" height="128">/usr/share/icons/hicolor/128x128/apps/openra.png</icon>
-  <categories>
-​    <category>Game</category>
-​    <category>StrategyGame</category>
```
Mailaender: I can confirm that my file was causing trouble here although the validator error is cryptic:

```
appstream-util validate /usr/share/appdata/openra.appdata.xml 
/usr/share/appdata/openra.appdata.xml: failed to parse /usr/share/appdata/openra.appdata.xml: Error on line 32 char 9: <categories> already set '​' and tried to replace with '
    '
```

obrakmann:
I can't test this, but it looks fine, judging by that screenshot. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9cff43665bd15d7fc373364c23f103b7df895fb4)

Issue #8972, opened 2015-08-06
Replay hotkeys for pause/play/speed


SoScared:
More convenience to an already damn good replay viewer. I'm looking through tons of replays searching for that special shot. A simple set for hotkeys (preferably customizable) would make the experience that much better. I think one hotkey for play/pause and one for increase/decrease speed would be the most convenient.

pchote:
Good idea, and this should be relatively easy.

StexoO:
I just hit [Space] to pause a replay, like in any video player, but nothing happened. :disappointed_relieved: 
So :+1:

SoScared:
#10933 looks good.

Issue #10682, opened 2016-02-03
Key-support for the replay-section


Micr0Bit:
would like to skim through my replay with the "up" and "down" key ... instead of clicking every single one ... also , could use "left" and right" to skip a whole stack 

![openra-2016-02-03t101950z](https://cloud.githubusercontent.com/assets/9139965/12779641/5543eeee-ca69-11e5-903c-6a2b3857f328.png)

PR #10933, opened 2016-03-16
Replay hotkeys for pause/play/change of speed

Hanzik:
Related to #8972. I also added hint to button tooltips via edditing {mod}/chrome/ingame-observer.yaml or in case of cnc ingame.yaml. You can see the result here: http://imgur.com/a/kk64J

I added the hotkeys to OpenRA.Game/Settings.cs and added this functionality to OpenRA.Mods.Common/Widgets/Logic/Ingame/ReplayControlBarLogic.cs. Hopefully that is the correct place for that.

Keys are mapped to:
Pause/Play: Spacebar
Slow speed: F5
Regular speed: F6
Fast speed: F7
Max speed: F8

### DIFF
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -269,6 +269,11 @@ namespace OpenRA
 		public Hotkey SupportPower05Key = new Hotkey(Keycode.UNKNOWN, Modifiers.None);
 		public Hotkey SupportPower06Key = new Hotkey(Keycode.UNKNOWN, Modifiers.None);
 
+		public Hotkey ReplaySpeedSlowKey = new Hotkey(Keycode.F5, Modifiers.None);
+		public Hotkey ReplaySpeedRegularKey = new Hotkey(Keycode.F6, Modifiers.None);
+		public Hotkey ReplaySpeedFastKey = new Hotkey(Keycode.F7, Modifiers.None);
+		public Hotkey ReplaySpeedMaxKey = new Hotkey(Keycode.F8, Modifiers.None);
+
 		static readonly Func<KeySettings, Hotkey>[] ProductionKeys = GetKeys(24, "Production");
 		static readonly Func<KeySettings, Hotkey>[] SupportPowerKeys = GetKeys(6, "SupportPower");
 
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/ReplayControlBarLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/ReplayControlBarLogic.cs
@@ -46,15 +46,18 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				var originalTimestep = world.Timestep;
 
 				var pauseButton = widget.Get<ButtonWidget>("BUTTON_PAUSE");
+				pauseButton.GetKey = _ => Game.Settings.Keys.PauseKey;
 				pauseButton.IsVisible = () => world.Timestep != 0 && orderManager.NetFrameNumber < replayNetTicks;
 				pauseButton.OnClick = () => world.Timestep = 0;
 
 				var playButton = widget.Get<ButtonWidget>("BUTTON_PLAY");
+				playButton.GetKey = _ => Game.Settings.Keys.PauseKey;
 				playButton.IsVisible = () => world.Timestep == 0 || orderManager.NetFrameNumber >= replayNetTicks;
 				playButton.OnClick = () => world.Timestep = (int)Math.Ceiling(originalTimestep * multipliers[speed]);
 				playButton.IsDisabled = () => orderManager.NetFrameNumber >= replayNetTicks;
 
 				var slowButton = widget.Get<ButtonWidget>("BUTTON_SLOW");
+				slowButton.GetKey = _ => Game.Settings.Keys.ReplaySpeedSlowKey;
 				slowButton.IsHighlighted = () => speed == PlaybackSpeed.Slow;
 				slowButton.IsDisabled = () => orderManager.NetFrameNumber >= replayNetTicks;
 				slowButton.OnClick = () =>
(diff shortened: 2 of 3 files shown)


OpenRA.Mods.Common/Widgets/Logic/Ingame/ReplayControlBarLogic.cs
```diff
@@ -51,49 +56,88 @@ public ReplayControlBarLogic(Widget widget, World world, OrderManager orderManag
 
 				var playButton = widget.Get<ButtonWidget>("BUTTON_PLAY");
 				playButton.IsVisible = () => world.Timestep == 0 || orderManager.NetFrameNumber >= replayNetTicks;
-				playButton.OnClick = () => world.Timestep = (int)Math.Ceiling(originalTimestep * multipliers[speed]);
+				playButton.OnClick = () => world.Timestep = (int)Math.Ceiling(originalTimestep * multipliers[currentSpeed]);
 				playButton.IsDisabled = () => orderManager.NetFrameNumber >= replayNetTicks;
 
 				var slowButton = widget.Get<ButtonWidget>("BUTTON_SLOW");
-				slowButton.IsHighlighted = () => speed == PlaybackSpeed.Slow;
+				slowButton.IsHighlighted = () => currentSpeed == PlaybackSpeed.Slow;
```
pchote: You can simplify things by binding `slowButton.GetKey = () => Game.Settings.Keys.ReplaySpeedSlowKey` (and similar for the other buttons).  Most of the other changes in this file can then go away.

mods/ra/chrome/ingame-observer.yaml
```diff
@@ -124,13 +124,14 @@ Container@OBSERVER_WIDGETS:
...
 			Children:
+				LogicKeyListener@REPLAY_KEYHANDLER:
 				Button@BUTTON_PAUSE:
 					X: 9
 					Y: 5
 					Width: 28
 					Height: 28
 					Background: sidebar-button-observer
-					TooltipText: Pause
+					TooltipText: Pause (Spacebar)
```
pchote: The tooltip will automatically show the hotkeys (in a different color) after you set up the button key bindings, so these changes will be able to go too.

OpenRA.Game/Settings.cs
```diff
@@ -269,6 +269,12 @@ public class KeySettings
 		public Hotkey SupportPower05Key = new Hotkey(Keycode.UNKNOWN, Modifiers.None);
 		public Hotkey SupportPower06Key = new Hotkey(Keycode.UNKNOWN, Modifiers.None);
 
+		public Hotkey PauseAltKey = new Hotkey(Keycode.SPACE, Modifiers.None);
```
pchote: You should also add these to the observer hotkeys dictionary in `SettingsLogic` so that they can be configured through the GUI.

obrakmann: Space is already used for going to the location of the last event.

Hanzik: Alright, I will tweak it around a littlebit. Spacebar is usually related to play/pause, but I can change it to something else, no problem with that.

pchote: I suggest just using the existing pause key, since that doesn't do anything in replays anyway.

Hanzik:
Alright, I made changes as you suggested and I added these values to SettingsLogic... you can see the result on this screenshot, too. http://imgur.com/7qmHBK8

Micr0Bit:
is #10682 related to this ?

Hanzik:
No. But I can have a look at that too at some point...

Phrohdoh:
Question for @pchote mainly: Would it be sensible to dynamically build up these settings? Not all mods are guaranteed to have the same steps that we have.

Perhaps not in this PR, just food for thought.

pchote:
@Phrohdoh: yup, that was always the plan.  But, as you well know, "short-term" hacks have a strange habit of lingering for the long-term if/when they are merged.

Phrohdoh:
Then this looks solid to me. :+1: without a test, so take that as you will.

SoScared:
Nice, thanks! I've been looking forward to this^^

OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
```diff
@@ -467,7 +467,12 @@ Action InitInputPanel(Widget panel)
 				var hotkeys = new Dictionary<string, string>()
 				{
 					{ "ObserverCombinedView", "All Players" },
-					{ "ObserverWorldView", "Disable Shroud" }
+					{ "ObserverWorldView", "Disable Shroud" },
+					{ "PauseKey", "Pause/Play" },
```
obrakmann: Really minor nitpick, you don't have to do this unless you feel like it: I'd put the pause key at the end of the list, so that the four speed modifier keys are next to each other in two lines, just for symmetry's sake.

obrakmann:
Works fine, :+1: 

I'll wait with merging to give you a chance to associate those commits with your account.

OpenRA.Mods.Common/Widgets/Logic/SettingsLogic.cs
Hanzik: I was wondering about the same thing, but then I thought Pause/Play key has more priority. I might change that, will see...

obrakmann:
@Hanzik: in case you aren't aware, keeping the email addresses in github and your git config in sync doesn't automatically update existing commits. You need to amend those commits with the `--author` parameter, like `git commit --amend --author=yourmail@example.com`.

Hanzik:
Yeh, I was wondering if the standard command updates existing commits... didn't know about this command, thanks!

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/950e0f7a30e3e3871f5007981a5170a11611795b)

PR #10935, opened 2016-03-16
Remove sdk flag from default makefile.

pchote:
Linux distros are slowly migrating to mono >= 4, and most of them don't seem to be including support for the 4.0 SDK (probably because, as far as I understand, this relies on shipping binaries that can no longer be compiled from source).  This then leads to issues like #10248 and #10726.

This removes the forced SDK from the default makefile, but keeps it for our travis builds (which use the Xamarin mono builds that _do_ have 4.0 support) to retain compatibility with .NET 4 on windows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,8 @@
 
 ############################## TOOLCHAIN ###############################
 #
-CSC         = mcs -sdk:4.0
+SDK         ?=
+CSC         = mcs $(SDK)
 CSFLAGS     = -nologo -warn:4 -codepage:utf8 -unsafe -warnaserror
 DEFINE      = TRACE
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -42,7 +42,8 @@
 
 ############################## TOOLCHAIN ###############################
 #
-CSC         = mcs -sdk:4.0
+SDK        :=
```
Phrohdoh: Missing a space here?

Phrohdoh:
:+1:

Makefile
pchote: A space where?

obrakmann: he means that you need one more space of indentation in front of the :=

obrakmann:
Phro's nit aside, looks fine :+1:

Makefile
Phrohdoh: Before the `:`.
Unless you were aligning on the  `=` which is inconsistent.

Mailaender: I was told that `?=` should be used for parametrizable variables, compare https://github.com/OpenRA/OpenRA/pull/10935/files#diff-b67911656ef5d18c4ae36cb6741b7965R66 and following, but honestly I don't know why.

pchote:
Updated.

Makefile
obrakmann: @Mailaender: see https://www.gnu.org/software/make/manual/html_node/Setting.html#Setting. I never had an issue with overriding a make variable from the command line, even when ?= wasn't used, though.

PR #10938, opened 2016-03-18
Unhardcode Explosion "Image"

reaperrr:
Adds `ExplosionCollection` property to `CreateEffectWarhead`.

Additionally - although I didn't test it - going by the code I believe that before my fix, `Explodes`, `Parachutable`, `WithBuildingExplosion` would only pass lint tests but NOT work in-game with anything other than "explosion", because that image string was still hardcoded on the effect itself.

Should make life easier for 3rd-party mods.

### DIFF
--- a/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
@@ -18,8 +18,11 @@ namespace OpenRA.Mods.Common.Warheads
 {
 	public class CreateEffectWarhead : Warhead
 	{
-		[Desc("List of explosion effects that can be used.")]
-		public readonly string[] Explosions = new string[0];
+		[Desc("List of explosion sequences that can be used.")]
+		[SequenceReference("Image")] public readonly string[] Explosions = new string[0];
+
+		[Desc("Image containing explosion effect sequence.")]
+		public readonly string Image = "explosion";
 
 		[Desc("Palette to use for explosion effect."), PaletteReference("UsePlayerPalette")]
 		public readonly string ExplosionPalette = "effect";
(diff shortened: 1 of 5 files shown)


OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
```diff
@@ -19,7 +19,10 @@ namespace OpenRA.Mods.Common.Warheads
 	public class CreateEffectWarhead : Warhead
 	{
 		[Desc("List of explosion effects that can be used.")]
-		public readonly string[] Explosions = new string[0];
+		[SequenceReference("ExplosionCollection")] public readonly string[] Explosions = new string[0];
+
+		[Desc("Explosion collection containing effect.")]
+		public readonly string ExplosionCollection = "explosion";
```
Phrohdoh: Do we use `Collection` anywhere else?
This doesn't sound familiar to me at all.

reaperrr: Well, Parachutable and WithBuildingExplosion already did.

The main reason is that `Image` would be somewhat unfitting given how the explosion sequences are set up.

pchote: Those two traits will need fixing too.  The `Collection` term is used for UI artwork, not sequences.

reaperrr:
Updated.

Did away with the prefixes and went for just `Image` for the sake of consistency.
Added upgrade rule as a bonus, although that's probably unnecessary since everything else than "explosion" would currently crash on bleed, so I suspect no 3rd-party mod has tried to use custom values anyway.

OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
```diff
@@ -101,8 +104,8 @@ public override void DoImpact(Target target, Actor firedBy, IEnumerable<int> dam
 				palette += firedBy.Owner.InternalName;
 
 			var explosion = Explosions.RandomOrDefault(Game.CosmeticRandom);
-			if (explosion != null)
-				world.AddFrameEndTask(w => w.Add(new Explosion(w, pos, explosion, palette)));
+			if (Image != null && explosion != null)
```
Phrohdoh: You'll probably want to check `string.IsNullOrEmpty` instead.

Mailaender: I can't point to the line in `FieldLoader` where it is decided, but my experience is that it returns `null` instead of empty strings in all relevant cases.

Mailaender:
Exposing this to https://github.com/OpenRA/OpenRA/wiki/Traits documentation is always a good idea. :+1:

obrakmann:
Looks good to me, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/80b652d9a135b644e88830e887f41aaf7dc60430)

Issue #10852, opened 2016-02-28
AI without Refinery but not in low power will stop production.


GraionDilach:
Compared to that, AI in low power and without Conyard will still train new stuff from the remaining factories.

obrakmann:
Are you sure about that? I just tested it and the AI kept on producing units just fine:
![screenshot from 2016-03-08 19 18 02](https://cloud.githubusercontent.com/assets/4331210/13611405/b1a63e08-e562-11e5-92ad-69dec0662cb3.png)

The AI *will*, however, stop producing things when it is out of refineries. Could that be what you saw?

GraionDilach:
It finishes building what it started but does not queue any more. I'm watching AI matches all day, it's definitely not the refinery.

obrakmann:
Hmmno, I couldn't reproduce that. One AI was building hordes of infantry after I destroyed its conyard.

abcdefg30:
Sorry, can't reproduce this on my end in RA either. Could it be your AI definition?

GraionDilach:
It might be the refinery as well though. TBH I lost my ideas too.

abcdefg30:
Yep, it is the refinery.

MustaphaTR:
Are you sure it is not because AI loses al the doposited ore and goes of money, so it can't build anything.

abcdefg30:
I just gave the AI some map placed buildings. With a refinery it happily started to produce blobs of units. Without a refinery (but still with the starting money) it just did nothing.

pchote:
This appears to be by design: HackyAI checks for this this specific case and disables production: see https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/AI/HackyAI.cs#L1017 and https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/AI/HackyAI.cs#L432

pchote:
`HasAdequateProc` should probably include a check for the construction yard to avoid a permanent stall if you destroy its refineries and structure production.

GraionDilach:
The autocloser ended up bogus.

PR #10940, opened 2016-03-18
Fix AI stalling after losing its conyard and ref.

pchote:
Speculative fix for #10852.

I haven't tested it, but the code speaks for itself.


obrakmann:
Seems to work as expected. The AI halts production of units until after a refinery could be built, but continues building if the conyard is gone as well. :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c2ab348c9fe9d4e0d6b8d3fa6d12ff58820b5faa...32c933322f0664401a7f21afb2d6e67c070bc9a0)

PR #10942, opened 2016-03-18
Remove map lazyness.

pchote:
`Map` objects are now only initialized when we need access to the full map object.  This removes a layer of indirection that is no longer necessary, cleaning up the public interface and making the performance and memory usage more predictable.


obrakmann:
Looks ok, :+1:

Mailaender:
Map loading also after mod switching and spawn lint checks still work. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5629b65077366b285721f2163bf23ddaaa6a5b20...6b75e9d7c23f1c61579dabcde2ea2c0adad746c9)

PR #10943, opened 2016-03-18
Move Map.TilesByDistance to MapGrid and un-static it.

pchote:
As a bonus, mods can now configure the 50 cell limit in their mod.yaml.

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -70,7 +70,6 @@ namespace OpenRA
 	{
 		public const int SupportedMapFormat = 10;
 
-		public const int MaxTilesInCircleRange = 50;
 		public readonly MapGrid Grid;
 		readonly ModData modData;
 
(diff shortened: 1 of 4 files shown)


obrakmann:
looks ok, :+1:

OpenRA.Game/Map/Map.cs
```diff
@@ -1152,8 +1110,8 @@ public IEnumerable<CPos> FindTilesInAnnulus(CPos center, int minRange, int maxRa
 			if (maxRange < minRange)
 				throw new ArgumentOutOfRangeException("maxRange", "Maximum range is less than the minimum range.");
 
-			if (maxRange >= TilesByDistance.Length)
-				throw new ArgumentOutOfRangeException("maxRange", "The requested range ({0}) exceeds the maximum allowed ({1})".F(maxRange, MaxTilesInCircleRange));
+			if (maxRange >= Grid.TilesByDistance.Length)
+				throw new ArgumentOutOfRangeException("maxRange", "The requested range ({0}) exceeds the maximum allowed ({1})".F(maxRange, Grid.MaximumTileSearchRange));
```
obrakmann: Actually, now that this is configurable, you might want to mention the setting in the message, something like "The requested range ({0}) cannot exceed the value of MaximumTileSearchRange ({1})."

(Similar thing in `Teleport.cs` below.

pchote:
Fixed.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/32c933322f0664401a7f21afb2d6e67c070bc9a0...5629b65077366b285721f2163bf23ddaaa6a5b20)

PR #10957, opened 2016-03-20
Fix D2k deviator not taking effect

### LINKED ISSUE #10278: Deviator tanks don't seem to work on release-20151224
I played several games against Ordos, with a constant stream of Deviators coming my way, but I didn't see a single one of my units have it's owner changed.
I suspect the (somewhat) recently added inaccuracy doesn't play well with their small damage spread, and the targets don't get affected at all by the gas.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29389798-deviator-tanks-don-t-seem-to-work-on-release-20151224?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
The warhead's scan range was lower than both the missile inaccuracy as well as the Missile CloseEnough value of 298, so it almost never "hit" the target.

Closes #10278.

### DIFF
--- a/mods/d2k/weapons/missiles.yaml
+++ b/mods/d2k/weapons/missiles.yaml
@@ -151,7 +151,7 @@ DeviatorMissile:
 		HorizontalRateOfTurn: 3
 		Blockable: false
 		Shadow: yes
-		Inaccuracy: 768
+		Inaccuracy: 384
 		Image: MISSILE
 		TrailImage: deviator_trail
 		TrailPalette: deviatorgas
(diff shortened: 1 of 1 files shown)


obrakmann:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/aa82fc9102a8bf2d24f62e1a03e06baf5fb256bf...c2ab348c9fe9d4e0d6b8d3fa6d12ff58820b5faa)

PR #10959, opened 2016-03-20
Expose passenger enter/exit notifications to Lua

phrohdoh:
Example usage:

``` lua
    local trans = Actor.Create(...)

    Trigger.OnPassengerExited(trans, function(_, passenger)
        passenger.AttackMove(somewhere.Location)

        if not trans.HasPassengers then
            trans.Move(somewhere_else.Location)
        end
    end)

...

Trigger.OnPassengerExited(myTransport, DoStuff)

DoStuff = function(trans, pass)
    if pass.Type == "dog" then
        trans.Destroy()
    end
end
```

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Global/TriggerGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/TriggerGlobal.cs
@@ -51,6 +51,20 @@ namespace OpenRA.Mods.Common.Scripting
 			Context.World.AddFrameEndTask(w => w.Add(new DelayedAction(delay, doCall)));
 		}
 
+		[Desc("Call a function for each passenger when it enters a transport. " +
+		      "The callback function will be called as func(Actor transport, Actor passenger).")]
+		public void OnPassengerEntered(Actor a, LuaFunction func)
+		{
+			GetScriptTriggers(a).RegisterCallback(Trigger.OnPassengerEntered, func, Context);
+		}
+
+		[Desc("Call a function for each passenger when it exits a transport. " +
+			"The callback function will be called as func(Actor transport, Actor passenger).")]
+		public void OnPassengerExited(Actor a, LuaFunction func)
+		{
+			GetScriptTriggers(a).RegisterCallback(Trigger.OnPassengerExited, func, Context);
+		}
+
 		[Desc("Call a function each tick that the actor is idle. " +
 			"The callback function will be called as func(Actor self).")]
 		public void OnIdle(Actor a, LuaFunction func)
--- a/OpenRA.Mods.Common/Scripting/ScriptTriggers.cs
+++ b/OpenRA.Mods.Common/Scripting/ScriptTriggers.cs
@@ -12,6 +12,7 @@
 using System;
 using System.Collections.Generic;
 using Eluant;
+using OpenRA.Mods.Common.Traits;
 using OpenRA.Scripting;
 using OpenRA.Traits;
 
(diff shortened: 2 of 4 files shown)


OpenRA.Mods.Common/Scripting/ScriptTriggers.cs
```diff
@@ -421,6 +422,46 @@ public void OnDiscovered(Actor self, Player discoverer, bool playNotification)
...
+		public void OnPassengerEntered(Actor self, Actor passenger)
+		{
+			if (world.Disposing)
+				return;
+
+			foreach (var f in Triggerables(Trigger.OnPassengerEntered))
+			{
+				try
+				{
+					using(var pass = passenger.ToLuaValue(f.Context))
```
pchote: Space before the `(` please (here and below)

Phrohdoh:
Fixed style.

Mailaender:
https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/maps/desert-shellmap/desert-shellmap.lua#L57 provides a real world test case for this.

Phrohdoh:
Not really.
All that does is unloads the passengers.
There is no task for them to being afterwards, unless you're wanting me to add `a.Hunt()`.

Mailaender:
Yeah, something like that might make them act less stupid.

OpenRA.Mods.Common/Scripting/Global/TriggerGlobal.cs
```diff
@@ -51,6 +51,20 @@ public void AfterDelay(int delay, LuaFunction func)
 			Context.World.AddFrameEndTask(w => w.Add(new DelayedAction(delay, doCall)));
 		}
 
+		[Desc("Call a function for each passenger when it enters a transport. " +
+		      "The callback function will be called as func(Actor passenger).")]
```
obrakmann: It might be useful to get the transport in the callback as well (e.g., make the transport move somewhere once an actor has entered). Same with `OnPassengerExited`.

obrakmann:
Aside from that, looks good, :+1:

Phrohdoh:
Added the transport to the callbacks.

OpenRA.Mods.Common/Traits/Attack/AttackGarrisoned.cs
```diff
@@ -93,7 +93,7 @@ public AttackGarrisoned(Actor self, AttackGarrisonedInfo info)
 			getArmaments = () => armaments;
 		}
 
-		public void PassengerEntered(Actor self, Actor passenger)
+		public void OnPassengerEntered(Actor self, Actor passenger)
```
GraionDilach: Since you're renaming this, should this be now explicitly implemented according to your recent craze?

pchote: I agree. Please do this.

Phrohdoh: Sure thing.

Phrohdoh:
Explicitly implemented interfaces on AttackGarrisoned.

OpenRA.Mods.Common/Scripting/ScriptTriggers.cs
```diff
@@ -421,6 +422,48 @@ public void OnDiscovered(Actor self, Player discoverer, bool playNotification)
 			}
 		}
```
pchote: Here too please, and then add the attribute that requires an explicit implementation. Sorry, I assumed that was obvious from the original discussion, but I guess not.

Phrohdoh: You cannot explicitly implement interface members here.
These members are access via a `ScriptTriggers` instance, not an interface.

pchote: The reason that the compiler complains is that you've forgotten to implement `INotifyPassengerEntered` and `INotifyPassengerExited` on `ScriptTriggers`.  This means that your triggers will never actually fire.

Phrohdoh:
Updated and split this into two commits.

OpenRA.Mods.Common/Scripting/ScriptTriggers.cs
Phrohdoh: I was mistaken.
You cannot explicitly implement interface members on `TriggerGlobal`.

abcdefg30:

Btw, your testcase doesn't work. :p
I had to declare `DoStuff` before using it in the `Trigger.OnPassengerEntered`.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/504026b67097b030830080c4ab6f8954c0797fb5...2bf7be9b1b8e1775be82f9d5d0abc0a62b15fc19)

PR #10963, opened 2016-03-21
Rework map data load/save.

pchote:
This reorganizes the map fields into a semblance of order, and, where possible, makes them readonly or `{ private set; }`.  Also tidies up the TS map importer a bit.

### DIFF
--- a/OpenRA.Game/Map/Map.cs
+++ b/OpenRA.Game/Map/Map.cs
@@ -15,6 +15,7 @@ using System.Drawing;
 using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Security.Cryptography;
 using System.Text;
 using OpenRA.FileSystem;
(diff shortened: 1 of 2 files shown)


OpenRA.Game/Map/Map.cs
```diff
@@ -66,28 +67,185 @@ public enum MapVisibility
...
+			new MapField("MapFormat"),
+			new MapField("RequiresMod"),
+			new MapField("Title"),
+			new MapField("Author"),
+			new MapField("Tileset"),
+			new MapField("MapSize"),
+			new MapField("Bounds"),
+			new MapField("Visibility"),
+			new MapField("Type"),
+			new MapField("LockPreview", required: false, ignoreIfValue: "False"),
```
Phrohdoh: Does this subtly enforce case with `True`/`False`?

pchote: The yaml saver already does that.

Phrohdoh:
:+1:

obrakmann:
Looks fine here as well, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/7b6845c062dce19a38d2f955bccb0d12badfd1f2)

PR #10965, opened 2016-03-22
Strengthen trait ordering rules

### LINKED ISSUE #10907: Requires<> is bogus.
As my recent modding experiments uncovered it, `Requires<>` turned out to be bogus. If the required trait has multiple instances on the actual actor and the trait which is required is placed between them, `Requires<>` will not force the followup instances to be loaded before the trait which requires them.

Testcase in bleed TS: add `AutoTarget` between the two `Attack*` traits to the Tick Tank (`ttnk`). The game won't crash. Add it to the end of the unit and the game will crash due to `AutoTarget` is still incapable of handling more `Attack*` traits. `AutoTarget` requires `AttackBase`.

RoosterDragon:
Fixes #10907.

Currently, if you have multiple dependencies for a trait, as long as one of them is created first then the ordering is satisfied. This means the others might be created after the thing that requires them!

Now, we strengthen the rules to require all of the dependencies to be created up front. This means the dependent can now guarantee all its dependencies are ready, rather than just at least one.

I have adjusted the tests to verify this.

This created some circular dependency chains on `UpgradeManager` in the default mods which I have resolved by doing the trait lookups after the actor is created. Without this RA and D2k fail to load.

### DIFF
--- a/OpenRA.Game/GameRules/ActorInfo.cs
+++ b/OpenRA.Game/GameRules/ActorInfo.cs
@@ -100,9 +100,15 @@ namespace OpenRA
 			var unresolved = source.Except(resolved);
 
 			var testResolve = new Func<Type, Type, bool>((a, b) => a == b || a.IsAssignableFrom(b));
-			var more = unresolved.Where(u => u.Dependencies.All(d => resolved.Exists(r => testResolve(d, r.Type)) && !unresolved.Any(u1 => testResolve(d, u1.Type))));
 
-			// Re-evaluate the vars above until sorted
+			// This query detects which unresolved traits can be immediately resolved as all their direct dependencies are met.
+			var more = unresolved.Where(u =>
+				u.Dependencies.All(d => // To be resolvable, all dependencies must be satisfied according to the following conditions:
+					resolved.Exists(r => testResolve(d, r.Type)) && // There must exist a resolved trait that meets the dependency.
+					!unresolved.Any(u1 => testResolve(d, u1.Type)))); // All matching traits that meet this dependency must be resolved first.
+
+			// Continue resolving traits as long as possible.
+			// Each time we resolve some traits, this means dependencies for other traits may then be possible to satisfy in the next pass.
 			while (more.Any())
 				resolved.AddRange(more);
 
--- a/OpenRA.Game/GameRules/ActorInfo.cs
+++ b/OpenRA.Game/GameRules/ActorInfo.cs
@@ -100,7 +100,7 @@ namespace OpenRA
 			var unresolved = source.Except(resolved);
 
 			var testResolve = new Func<Type, Type, bool>((a, b) => a == b || a.IsAssignableFrom(b));
-			var more = unresolved.Where(u => u.Dependencies.All(d => resolved.Exists(r => testResolve(d, r.Type))));
+			var more = unresolved.Where(u => u.Dependencies.All(d => resolved.Exists(r => testResolve(d, r.Type)) && !unresolved.Any(u1 => testResolve(d, u1.Type))));
 
 			// Re-evaluate the vars above until sorted
 			while (more.Any())
@@ -122,14 +122,14 @@ namespace OpenRA
 					exceptionString += u.Type + ": { " + string.Join(", ", deps) + " }\r\n";
 				}
 
-				throw new Exception(exceptionString);
+				throw new YamlException(exceptionString);
 			}
 
 			constructOrderCache = resolved.Select(r => r.Trait).ToList();
 			return constructOrderCache;
 		}
 
-		static IEnumerable<Type> PrerequisitesOf(ITraitInfo info)
+		public static IEnumerable<Type> PrerequisitesOf(ITraitInfo info)
 		{
 			return info
 				.GetType()
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
@@ -17,12 +17,18 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Attach this to a unit to enable dynamic upgrades by warheads, experience, crates, support powers, etc.")]
-	public class UpgradeManagerInfo : ITraitInfo, Requires<IUpgradableInfo>
+	public class UpgradeManagerInfo : TraitInfo<UpgradeManager>, IRulesetLoaded
 	{
-		public object Create(ActorInitializer init) { return new UpgradeManager(init); }
+		public void RulesetLoaded(Ruleset rules, ActorInfo info)
+		{
+			if (!info.Name.StartsWith("^") && !info.TraitInfos<IUpgradableInfo>().Any())
+				throw new YamlException(
+					"There are no upgrades to be managed for actor '{0}'. You are either missing some upgradeable traits, or this UpgradeManager trait is not required.".F(
+						info.Name));
+		}
 	}
 
-	public class UpgradeManager : ITick
+	public class UpgradeManager : INotifyCreated, ITick
 	{
 		class TimedUpgrade
 		{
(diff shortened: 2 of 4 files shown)


OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -67,20 +67,27 @@ class UpgradeState
...
+				hasUpgradeables = true;
+				foreach (var t in up.UpgradeTypes)
+					upgrades.GetOrAdd(t).Traits.Add(up);
+			}
 
-				return ret;
-			});
+			if (!hasUpgradeables)
+				throw new Exception(
+					"There are no upgrades to be managed for actor '{0}'. You are either missing some upgradeable traits, or this UpgradeManager trait is not required.".F(
```
pchote: Could we check for this via `IRulesetLoaded` in the trait info?

OpenRA.Game/GameRules/ActorInfo.cs
```diff
@@ -100,7 +100,7 @@ public IEnumerable<ITraitInfo> TraitsInConstructOrder()
 			var unresolved = source.Except(resolved);
 
 			var testResolve = new Func<Type, Type, bool>((a, b) => a == b || a.IsAssignableFrom(b));
-			var more = unresolved.Where(u => u.Dependencies.All(d => resolved.Exists(r => testResolve(d, r.Type))));
+			var more = unresolved.Where(u => u.Dependencies.All(d => resolved.Exists(r => testResolve(d, r.Type)) && !unresolved.Any(u1 => testResolve(d, u1.Type))));
```
pchote: Vagely related scope creep (not a blocker): Using LINQ here might be compact, but it makes what is otherwise a simple algorithm really difficult to parse.  How much would I have to bribe to you rewrite this block of 6 lines in a clearer way or to add comments?

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -17,12 +17,19 @@
...
 {
 	[Desc("Attach this to a unit to enable dynamic upgrades by warheads, experience, crates, support powers, etc.")]
-	public class UpgradeManagerInfo : ITraitInfo, Requires<IUpgradableInfo>
+	public class UpgradeManagerInfo : ITraitInfo, IRulesetLoaded
 	{
 		public object Create(ActorInitializer init) { return new UpgradeManager(init); }
+		public void RulesetLoaded(Ruleset rules, ActorInfo info)
+		{
+			if (!info.Name.StartsWith("^") && !info.TraitInfos<IUpgradableInfo>().Any())
+				throw new Exception(
```
pchote: I believe the convention is to use `YamlException` here (?)

pchote:
This looks correct to me, but I haven't tested it so :+0.5: for now.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
Phrohdoh: Yes, please do that.

RoosterDragon: Shall I make the same change in `TraitsInConstructOrder`? (Currently also throws Exception)

RoosterDragon:
Actually this may cause issues for `Cloak` and `TimedUpgradeBar` which use the upgrade manager in their created callbacks. There's no guarantee that will work since the upgrade manager may not have run its creation callback yet,

I will have to think that over some more.

RoosterDragon:
As pointed out to me on IRC, Cloak can be fixed with a frame end task and TimedUpgradeBar has a requires so that solves that.

I have tweaked the exception types and added some extra comments.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -67,20 +74,17 @@ class UpgradeState
...
-			{
-				var ret = new Dictionary<string, UpgradeState>();
-				foreach (var up in init.Self.TraitsImplementing<IUpgradable>())
-					foreach (var t in up.UpgradeTypes)
-						ret.GetOrAdd(t).Traits.Add(up);
+		public UpgradeManager(ActorInitializer init) { }
 
-				return ret;
-			});
+		public void Created(Actor self)
```
Phrohdoh: Since you are introducing this here and don't need to call this yourself, please explicitly implement this.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -67,20 +74,17 @@ class UpgradeState
 		}
 
 		readonly List<TimedUpgrade> timedUpgrades = new List<TimedUpgrade>();
-		readonly Lazy<Dictionary<string, UpgradeState>> upgrades;
 		readonly Dictionary<IUpgradable, int> levels = new Dictionary<IUpgradable, int>();
+		Dictionary<string, UpgradeState> upgrades;
```
pchote: Why not `readonly Dictionary<string, UpgradeState> upgrades = new Dictionary<string, UpgradeState>();`?

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
```diff
@@ -67,20 +74,17 @@ class UpgradeState
...
 
-		public UpgradeManager(ActorInitializer init)
-		{
-			upgrades = Exts.Lazy(() =>
-			{
-				var ret = new Dictionary<string, UpgradeState>();
-				foreach (var up in init.Self.TraitsImplementing<IUpgradable>())
-					foreach (var t in up.UpgradeTypes)
-						ret.GetOrAdd(t).Traits.Add(up);
+		public UpgradeManager(ActorInitializer init) { }
```
pchote: You could make `UpgradeManagerInfo : TraitInfo<UpgradeManager>, IRulesetLoaded` and then remove this empty ctor definition.

OpenRA.Test/OpenRA.Game/ActorInfoTest.cs
```diff
@@ -28,89 +27,55 @@ class MockDInfo : MockTraitInfo, Requires<MockEInfo> { }
...
-		ActorInfo CreateActorInfoFromYaml(string name, string mapYaml, params string[] yamls)
-		{
-			var nodes = mapYaml == null ? new List<MiniYamlNode>() : MiniYaml.FromString(mapYaml);
-			var sources = yamls.ToList();
-			if (mapYaml != null)
-				sources.Add(mapYaml);
+			var count = (
+				new Regex("MockD").Matches(ex.Message).Count +
+				new Regex("MockE").Matches(ex.Message).Count +
+				new Regex("MockF").Matches(ex.Message).Count) / 3.0;
```
pchote: Why not just compare the counts directly?

pchote:
Looks good otherwise :+1:

OpenRA.Mods.Common/Traits/Upgrades/UpgradeManager.cs
RoosterDragon: I wanted this to throw if the methods are accessed before `Created` was called. I've added an explicit exception with a message since this will be clearer in the code and for consumers.

OpenRA.Test/OpenRA.Game/ActorInfoTest.cs
RoosterDragon: Just didn't refactor it quite that much. I've changed the asserts.

RoosterDragon:
Made a few minor but non-trivial changes, so removing the +1.

OpenRA.Game/GameRules/ActorInfo.cs
```diff
@@ -100,9 +100,15 @@ public IEnumerable<ITraitInfo> TraitsInConstructOrder()
...
+			// This query detects which unresolved traits can be immediately resolved as all their direct dependencies are met.
+			var more = unresolved.Where(u =>
+				u.Dependencies.All(d => // To be resolvable, all dependencies must be satisfied according to the following conditions:
+					resolved.Exists(r => testResolve(d, r.Type)) && // There must exist a resolved trait that meets the dependency.
+					!unresolved.Any(u1 => testResolve(d, u1.Type)))); // All matching traits that meet this dependency must be resolved first.
+
+			// Continue resolving traits as long as possible.
+			// Each time we resolve some traits, this means dependencies for other traits may then be possible to satisfy in the next pass.
 			while (more.Any())
 				resolved.AddRange(more);
```
obrakmann: Why isn't this an infinite loop? `more` isn't being modified in the loop, afaics, so why should the result from `more.Any()` ever change?

pchote: `more` is an enumerable rather than a discrete collection.

obrakmann:
I guess this is fine then, :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/9608ab5fa8174518d46d0a07e10baa3d8fd333f7...7360a1b115e14cc5c11ad854125718061f8ff2a5)

Issue #10437, opened 2016-01-06
RA balance


ghost:
Making a single issue for everyone to discuss RA balance for the next release due to the feedback in another thread https://github.com/OpenRA/OpenRA/pull/10432#issuecomment-169298999

The last big rebalance mainly happened due to a chance conversation between lots of people on IRC, so this may be a good place to collect everyone's views and thoughts. I'll let others start as I know some people have interesting ideas for certain issues which I am keen to hear.
##

anjew175:
A lot people I have asked have agreed there is something off with the Allies artillery. I personally think it has to do with the range as you can have a lot of units/structures in the way of your artillery make them impossible to take care of in some situations

cjshmyr:
Agreed wrt Artillery - Lowering its weapon range slightly from 14 (to 12?) and lowering `155mm`'s damage versus `None` very slightly could help approaches, also put the unit in more vulnerable positions. I'm assuming `None` is infantry only.

Medium & Heavy tanks - A small HP buff to these would help them "tank" better, eating an extra base defense shot or rocket. I think they should be a bit more durable and push into bases better.

MAD re-work - Differentiate it from the demo truck by removing the self destruct and letting it un-deploy. Not sure if it should become a Ukraine unit

British Spy - Get rid of this unit completely and move the normal Spy to England. I think the bonuses it grants are pretty decent. Can keep its cheap cost of $300.

Radar Dome stealth - Remove stealth detection from the radar dome. Helps phase transport and hijacker.

ScottNZ:
:+1: Especially a tank buff

MicroJOo:
:+1: Especially the british Spy

Townes27:
:-1:  regarding the spies. There'd be no incentive to build dogs against Germany or France then.

anjew175:
my suggestion for spies would be make the regular spies have a capped stealing value and dont allow them vet units, keep all other traits. This would make the British Spy a better alternative to normal spies and they are so cheaper.

kyrylo:
Agreed wrt Artillery - :+1: but my preferred solution is https://github.com/OpenRA/OpenRA/pull/10411
Medium & Heavy tanks - :+1:
MAD re-work - dunno, this is a silly unit. Maybe make it work against structures only.
British Spy - :+1:
Radar Dome stealth - :+1: but infantry should be able to detect them (unless it already works like that)

ShadowA:
[...]

PR #10968, opened 2016-03-23
RA - Added Fake Power Plants for France

MustaphaTR:
As discussed in #10437.


Micr0Bit:
i honestly see no reason to add those ... it might even have a bad effect on basewalking  (disturbing defense to shoot on even cheaper fakes instead of the enemy defense or atleast the real "money-wasting version of powerplants) 

sry , but i was involved at the time on introducing the fakes together with scott ... and i think it is fine as it is now ... could find more interesting ways to buff france (if its even really necessary)

:-1:

MustaphaTR:
> bad effect on basewalking 
They don't give buildable area, at least they aren't supposed to do.

As @ShadowA said their primary purpose is being against spies. That would be really bad for Spy owner to waste its spy even worse failing to power down enemy base may cause big bad effects on attacker.

pchote:
Fake power plants seem like a reasonable addition to me
* It obfuscates your actual power usage vs generation, which a decent player can otherwise estimate by looking at your base.
* It disrupts the tactic of destroying your power plants to put you into low power.
* It disrupts spies.

obrakmann:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/4323ff4bd5e6198931e57f710474537c59b53325...3042ea078f707eba41d50645cf4bf6790e931634)

PR #10971, opened 2016-03-23
Fix bad bridge art for TD snow template

### LINKED ISSUE #6412: No smooth transition on Tiberian Dawn SNOW bridge tile edges
image

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4229432-no-smooth-transition-on-tiberian-dawn-snow-bridge-tile-edges?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

HenrytheSlav:
Fixes #6412.

You can see comparsion of the two cases here.

bridge_reivison_2

Of course this is custom art and just a cosmetic change, thus open to discussion/change.


reaperrr:
Looks good enough to me (and clearly better than the unmodified art).

GraionDilach:
What is the difference, may I ask? I can't find it that obvious.

reaperrr:
The right border of the original art is cut off quite abruptly, as if there was a part missing (the darker area).

GraionDilach:
Now that I noticed it... did anyone checked if the RA1 snow bridge is also affected? Might be good to be fixed here as well (I guess it is since this is a converted RA1 tile afaik.)

reaperrr:
> Now that I noticed it... did anyone checked if the RA1 snow bridge is also affected? Might be good to be fixed here as well (I guess it is since this is a converted RA1 tile afaik.)

Nope, the RA1 bridge looks fine.

The TD snow bridge is just the TD winter art painted onto snow tiles (the winter version doesn't look as bad due to the darker ground color).

HenrytheSlav:
Not shown is also the destroyed version, made the same changes as here because the tile had same abrupt cut,

Phrohdoh:
Do we really need to put the entire `snow.mix` in `/bits` and therefor source control, or would it be feasible to just extract that single file?

pchote:
`snow.mix` is already in `/bits`.  It could be extracted to a directory like we do in RA, however.

Phrohdoh:
Ah well then that sounds like a good idea for another PR.

:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/3e7167eec3a326222ae0b1c782ddd9f1440eedb1...c3f0467b65df8911cfa13b16789c7841784371c8)

Issue #10845, opened 2016-02-27
Remove remaining floating point calculations from simulation code.


pchote:
The Shattered Paradise stream had a couple of desyncs, which I suspect may be caused by fp disagreements that we don't see in the default mods because of our default values.  For consistency and safety we should work to remove the remaining fp calculations from the simulation code.

The ones I know (there may be more) are:
- [x] `WVec.LerpQuadratic` (can use Decimal or long)
- [x] `ReturnToBase.CalculateTurnRadius` (can use int constants)
- [x] `ProductionQueueInfo.BuildSpeed` (can use an int percentage)
- [x] `CapturableInfo.CaptureThreshold` (can use an int percentage)
- [x] `CapturesInfo.SabotageHPRemoval` (can use an int percentage)
- [x] `EmitInfantryOnSellInfo.ValuePercent` (can use an int percentage)
- [x] `EmitInfantryOnSellInfo.MinHpPercent` (can use an int percentage)
- [x] `StrategicVictoryConditionsInfo.RatioRequired` (can use an int percentage)
- [x] `SelfHealingInfo.HealIfBelow` (can use an int percentage)
- [x] `MinelayerInfo.MinefieldDepth` (can use WDist)

reaperrr:
Is it just me, or does nothing actually use `EmitInfantryOnSellInfo.MinHpPercent`? Couldn't find any logic making use of it...

PR #10972, opened 2016-03-24
Migrated some remaining floats in simulation to integers (and decimals)

reaperrr:
Partially adresses #10845.

### DIFF
--- a/OpenRA.Game/WPos.cs
+++ b/OpenRA.Game/WPos.cs
@@ -64,8 +64,8 @@ namespace OpenRA
 				return ret;
 
 			// Add an additional quadratic variation to height
-			// Attempts to avoid integer overflow by keeping the intermediate variables reasonably sized
-			var offset = (int)(((((((long)(b - a).Length * mul) / div) * (div - mul)) / div) * pitch.Tan()) / 1024);
+			// Uses decimal to avoid integer overflow
+			var offset = (int)((decimal)(b - a).Length * pitch.Tan() * mul * (div - mul) / (1024 * div * div));
 			return new WPos(ret.X, ret.Y, ret.Z + offset);
 		}
 
--- a/OpenRA.Game/WVec.cs
+++ b/OpenRA.Game/WVec.cs
@@ -82,8 +82,8 @@ namespace OpenRA
 				return ret;
 
 			// Add an additional quadratic variation to height
-			// Uses fp to avoid integer overflow
-			var offset = (int)((float)(b - a).Length * pitch.Tan() * mul * (div - mul) / (1024 * div * div));
+			// Uses decimal to avoid integer overflow
+			var offset = (int)((decimal)(b - a).Length * pitch.Tan() * mul * (div - mul) / (1024 * div * div));
 			return new WVec(ret.X, ret.Y, ret.Z + offset);
 		}
 
--- a/OpenRA.Mods.Common/Traits/EmitInfantryOnSell.cs
+++ b/OpenRA.Mods.Common/Traits/EmitInfantryOnSell.cs
@@ -19,8 +19,8 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Spawn new actors when sold.")]
 	public class EmitInfantryOnSellInfo : ITraitInfo
 	{
-		public readonly float ValuePercent = 40;
-		public readonly float MinHpPercent = 30;
+		public readonly int ValuePercent = 40;
+		public readonly int MinHpPercent = 30;
 
 		[ActorReference]
 		[Desc("Be sure to use lowercase. Default value is \"e1\".")]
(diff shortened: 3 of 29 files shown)


OpenRA.Game/WVec.cs
```diff
@@ -82,8 +82,8 @@ public static WVec LerpQuadratic(WVec a, WVec b, WAngle pitch, int mul, int div)
 				return ret;
 
 			// Add an additional quadratic variation to height
-			// Uses fp to avoid integer overflow
-			var offset = (int)((float)(b - a).Length * pitch.Tan() * mul * (div - mul) / (1024 * div * div));
+			// Uses long to avoid integer overflow
```
pchote: This isn't a hot path, so we could probably get away with using `decimal` here (there's another similar line somewhere using long that could also change) and remove the risk of overflow completely.

reaperrr: I think I saw WPos use long in a similar case, should that change to decimal as well?

pchote: Yeah, I think that is the line I was referring to.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -644,6 +644,71 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
...
+				if (engineVersion < 20160323)
+				{
+					if (node.Key == "SelfHealing")
+					{
+						var healIfBelowNode = node.Value.Nodes.FirstOrDefault(x => x.Key == "HealIfBelow");
+						if (healIfBelowNode != null)
+						{
+							// The HealIfBelow value is now an int percentage, so multiply the float with 100.
+							var oldValue = FieldLoader.GetValue<float>("HealIfBelow", healIfBelowNode.Value.Value);
+							healIfBelowNode.Value.Value = (oldValue * 100).ToString();
```
pchote: You should convert this to an int to truncate any fractional part before turning it to a string.

pchote: Likewise for the other conversions.

mods/cnc/maps/gdi06/rules.yaml
```diff
@@ -163,7 +163,7 @@ RMBO.easy:
 		HP: 300
 	SelfHealing:
 		Delay: 10
-		HealIfBelow: 50%
```
pchote: Is this the last use of the `%` syntax for loading floats?  We should drop that from FieldLoader if it is.

OpenRA.Mods.Common/Traits/SelfHealing.cs
```diff
@@ -18,7 +18,8 @@ class SelfHealingInfo : UpgradableTraitInfo, Requires<HealthInfo>
 	{
 		public readonly int Step = 5;
 		public readonly int Delay = 5;
-		public readonly float HealIfBelow = .5f;
+		[Desc("Heal if current health is below this percentage of full health.")]
+		public readonly int HealIfBelow = 50;
```
pchote: Do we want to rename these fields to include `Percent` in their name?  (I know we do this for some fields, but I'm not sure how widely that convention is used).

mods/cnc/maps/gdi06/rules.yaml
reaperrr: AI still uses `%` for actor build decisions.

OpenRA.Mods.Common/Traits/SelfHealing.cs
reaperrr: I'd say it's almost a tie between `Percent` and `Percentage` currently, plus a couple of cases like this with neither.

reaperrr:
Updated.

OpenRA.Game/WPos.cs
```diff
@@ -65,7 +65,7 @@ public static WPos LerpQuadratic(WPos a, WPos b, WAngle pitch, int mul, int div)
 
 			// Add an additional quadratic variation to height
 			// Attempts to avoid integer overflow by keeping the intermediate variables reasonably sized
-			var offset = (int)(((((((long)(b - a).Length * mul) / div) * (div - mul)) / div) * pitch.Tan()) / 1024);
+			var offset = (int)(((((((decimal)(b - a).Length * mul) / div) * (div - mul)) / div) * pitch.Tan()) / 1024);
```
pchote: You can now drop most of the nested parenthesis, and the comment above will need to be updated.

pchote: Compare with the calculation in `WVec`.

OpenRA.Mods.Common/Traits/EmitInfantryOnSell.cs
```diff
@@ -59,8 +59,14 @@ void Emit(Actor self)
 			var health = self.TraitOrDefault<Health>();
 			var dudesValue = info.ValuePercent * cost;
 			if (health != null)
-				dudesValue = dudesValue * health.HP / health.MaxHP;
-			dudesValue /= 100;
+			{
+				var healthPercentage = 100 * health.HP / health.MaxHP;
```
pchote: You can avoid rounding issues near the threshold by doing:
```csharp
if (100 * health.HP >= info.MinHpPercentage * health.MaxHP)
    dudesValue = health.HP * dudesValue / health.MaxHP;
else
    dudesValue = 0;
```

pchote:
Looks good otherwise :+1:

reaperrr:
Updated.

Mailaender:
Needs a rebase.

reaperrr:
Rebased.
Also added one last commit that moves all upgrade rules from both PRs to the same `engineVersion`.

Mailaender:
Thanks!

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3042ea078f707eba41d50645cf4bf6790e931634...1859227b12a757093693056d5012051c1950e8eb)

PR #10973, opened 2016-03-24
Getting rid of float in simulation code Part 2

reaperrr:
Covers the rest of examples listed in #10845 + `ThrowsParticle`.

### DIFF
--- a/OpenRA.Mods.Common/Traits/ThrowsParticle.cs
+++ b/OpenRA.Mods.Common/Traits/ThrowsParticle.cs
@@ -38,7 +38,7 @@ namespace OpenRA.Mods.Common.Traits
 		public readonly int Velocity = 75;
 
 		[Desc("Speed at which the particle turns.")]
-		public readonly float TurnSpeed = 15;
+		public readonly int TurnSpeed = 15;
 
 		public object Create(ActorInitializer init) { return new ThrowsParticle(init, this); }
 	}
@@ -53,8 +53,8 @@ namespace OpenRA.Mods.Common.Traits
 		int tick = 0;
 		int length;
 
-		float facing;
-		float rotation;
+		WAngle facing;
+		WAngle rotation;
 
 		public ThrowsParticle(ActorInitializer init, ThrowsParticleInfo info)
 		{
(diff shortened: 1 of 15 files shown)


OpenRA.Mods.Common/Traits/ThrowsParticle.cs
```diff
@@ -53,8 +53,8 @@ class ThrowsParticle : ITick
 		int tick = 0;
 		int length;
 
-		float facing;
-		float rotation;
+		int facing;
```
pchote: Could you make these `WAngle`, and then do `() => facing.Angle / 4` in the Animation ctor?  That makes it clearer what these are for, and has a side-effect of giving you a bit more precision.

pchote:
:+1: after that.

reaperrr:
@pchote Updated, but that required a bit more code changes than I had hoped, and I'm not confident in the code (though it seems to work fine in-game), so I kept it as separate commit for now.

OpenRA.Mods.Common/Traits/ThrowsParticle.cs
```diff
@@ -64,7 +64,7 @@ public ThrowsParticle(ActorInitializer init, ThrowsParticleInfo info)
 
 			// TODO: Carry orientation over from the parent instead of just facing
 			var bodyFacing = init.Contains<FacingInit>() ? init.Get<FacingInit, int>() : 0;
-			facing = Turreted.GetInitialTurretFacing(init, 0);
+			facing = new WAngle(Turreted.GetInitialTurretFacing(init, 0));
```
pchote: Use WAngle.FromFacing here.

OpenRA.Mods.Common/Traits/ThrowsParticle.cs
```diff
@@ -76,9 +76,9 @@ public ThrowsParticle(ActorInitializer init, ThrowsParticleInfo info)
 			length = (finalPos - initialPos).Length / info.Velocity;
 
 			// Facing rotation
-			rotation = WDist.FromPDF(Game.CosmeticRandom, 2).Length * info.TurnSpeed / 1024;
+			rotation = new WAngle(WDist.FromPDF(Game.CosmeticRandom, 2).Length * info.TurnSpeed / 1024);
```
pchote: FromFacing

reaperrr:
Updated.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3042ea078f707eba41d50645cf4bf6790e931634...1859227b12a757093693056d5012051c1950e8eb)

PR #10956, opened 2016-03-20
Add NamedActorNames to MapGlobal

phrohdoh:
This allows a script author to do something like:

``` lua
    for name, actor in pairs(Map.NameToActorKeyValues) do
        -- do something with `name` and `actor`
    end
```

Real-world example:
I am assigning owners based on the actor's name in yaml (which corresponds to player spawn points).

``` lua
spawn2alpha = { }
spawn2alpha[1] = 'a'
spawn2alpha[2] = 'b'
spawn2alpha[3] = 'c'
spawn2alpha[4] = 'd'

SetupStartingForces = function(player)
    for name, actor in pairs(Map.NameToActorKeyValues) do
        local prefix = name:sub(0, 1):lower()
        if prefix == spawn2alpha[player.Spawn] then
            actor.Owner = player
        end
    end
end
```

Where the `Actors` section of the `map.yaml` contains entries like this:

``` yaml
    a_caoild_1: caoild
        Location: 49,12
        Owner: Neutral
    a_caoild_2: caoild
        Location: 50,15
        Owner: Neutral
    b_caoild_1: caoild
        Location: 85,-45
        Owner: Neutral
    b_caoild_2: caoild
        Location: 83,-42
        Owner: Neutral
    c_caoild_1: caoild
        Owner: Neutral
        Location: 142,-1
    c_caoild_2: caoild
        Owner: Neutral
        Location: 144,2
    d_gapile_1: gapile
        Owner: Neutral
        Location: 102,43
    d_e1_1: e1
        Owner: Neutral
        Location: 81,44
    d_e1_2: e1
        Owner: Neutral
        Location: 100,53
```

Issue #10464, opened 2016-01-09
Add a way to tag actors via the Lua API


kyrylo:
Scenario: I spawn a bunch of crates and do not store them in a variable. Then I use `Map.ActorsInBox` to destroy some actors including these spawned crates. My simplified function:

```lua
function ClearMap()
  local unitsToDestroy =
    Map.ActorsInBox(Map.TopLeft,Map.BottomRight, function(actor)
      return IsHusk(actor) or actor.Owner ~= neutral or actor.HasProperty('Crate')
    end)

  Utils.Do(unitsToDestroy, function(actor)
   -- destroy...
  end)
end
```

As you can see the `actor.HasProperty('Crate')` part is not going to work because there's no such property. It would be rather handy to tag actors, so I could query them directly, without storing them in variables.

kyrylo:
Now I'm not sure it's needed, because I [re]discovered `actor.Type` :(

Unit158:
Set them in the global table.

Phrohdoh:
This seems like a workaround for improper programming (not storing them in variables, but want to access them later).

On the other hand engines such as Unity 3D have a very similar functionality.

pchote:
A simpler method would be to have a `ScriptTags` trait (like our existing `ScriptTriggers`) and a `ScriptTagsProperties : : ScriptActorProperties, Requires<ScriptTagsInfo>` that exposes methods for `AddTag`, `HasTag`, `RemoveTag`.

PR #10975, opened 2016-03-25
Add ScriptTags trait

phrohdoh:
This allows actors to be tagged with arbitrary strings.
Also includes `ScriptTagsInit` and `MapGlobal.ActorsWithTag`
  which expose this functionality to map scripters.

I am using this like such:

``` lua
_paradrop_wps = {
    ['a'] = Map.ActorsWithTag('paradrop_a'),
    ['b'] = Map.ActorsWithTag('paradrop_b'),
    ['c'] = Map.ActorsWithTag('paradrop_c'),
    ['d'] = Map.ActorsWithTag('paradrop_d')
}

...

SetupStartingForces = function(player)
    local spawn2alpha = { 'a', 'b', 'c', 'd' }
    for _, actor in ipairs(Map.NamedActors) do
        local tag = 'owner_' .. spawn2alpha[player.Spawn]
        if actor.HasTag(tag) then
            actor.Owner = player
        end
    end
end
```

ref #10956

### DIFF
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -736,6 +736,7 @@
     <Compile Include="Traits\World\MapOptions.cs" />
     <Compile Include="Traits\World\MissionData.cs" />
     <Compile Include="UtilityCommands\ExtractMapRules.cs" />
+    <Compile Include="Traits\ScriptTags.cs" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
--- a/OpenRA.Mods.Common/Scripting/Global/MapGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/MapGlobal.cs
@@ -9,6 +9,7 @@
  */
 #endregion
 
+using System.Collections.Generic;
 using System.Linq;
 using Eluant;
 using OpenRA.Mods.Common.Traits;
@@ -112,6 +113,12 @@ namespace OpenRA.Mods.Common.Scripting
 		public bool IsNamedActor(Actor actor)
 		{
 			return actor.ActorID <= sma.LastMapActorID && actor.ActorID > sma.LastMapActorID - sma.Actors.Count;
+		}
+
+		[Desc("Returns a table of all actors tagged with the given string.")]
+		public Actor[] ActorsWithTag(string tag)
+		{
+			return Context.World.ActorsHavingTrait<ScriptTags>(t => t.HasTag(tag)).ToArray();
 		}
 	}
 }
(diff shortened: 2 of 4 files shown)


OpenRA.Mods.Common/Traits/Taggable.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Allows this actor to be 'tagged' with arbitrary strings. Tags must be unique or they will be rejected.")]
+	public class TaggableInfo : UsesInit<TagsInit>
+	{
+		object ITraitInfo.Create(ActorInitializer init) { return new Taggable(init, this); }
+	}
+
+	public class Taggable
```
pchote: How about we name this `ScriptTags` to match e.g. `ScriptTriggers` and `ScriptUpgradesCache`?

OpenRA.Mods.Common/Traits/Taggable.cs
```diff
@@ -0,0 +1,57 @@
...
+	{
+		public readonly HashSet<string> Tags;
+
+		public Taggable(ActorInitializer init, TaggableInfo info)
+		{
+			var tags = init.Contains<TagsInit>() ? init.Get<TagsInit, string[]>() : null;
```
pchote: You're effectively using `null` as a magic value here.  I would write this as the clearer:
```csharp
		public readonly HashSet<string> Tags = new HashSet<string>();

		public Taggable(ActorInitializer init, TaggableInfo info)
		{
			if (init.Contains<TagsInit>())
				foreach (var tag in init.Get<TagsInit, string[]>())
					Tags.Add(tag);
		}
```

OpenRA.Mods.Common/Traits/Taggable.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+	[Desc("Allows this actor to be 'tagged' with arbitrary strings. Tags must be unique or they will be rejected.")]
+	public class TaggableInfo : UsesInit<TagsInit>
+	{
+		object ITraitInfo.Create(ActorInitializer init) { return new Taggable(init, this); }
+	}
+
+	public class Taggable
+	{
+		public readonly HashSet<string> Tags;
```
pchote: You have exposed methods to check tags, so this doesn't need to be public.

pchote:
Looks fine otherwise, although I haven't tested it.

OpenRA.Mods.Common/Traits/Taggable.cs
Phrohdoh: This isn't exclusive to scripting in any way other than my current usage, so I don't see this as a good idea.

pchote: Do you have any use-cases in mind that aren't related to scripted maps?  I can't think of any.

Phrohdoh: No, that is a fair point.
I will change this to `ScriptTags`.

Phrohdoh:
Rebased and renamed the trait.

pchote:
This implements the request at #10464.

obrakmann:
Looks fine to me, :+1:

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/72505d7fcbf7c71ab1ff19382c5a484929c0bd7a...2ec245ee0e51f17e7995d6d0ac03f27fd10b6f78)

PR #10981, opened 2016-03-27
Support custom SupportDir overrides.

pchote:
This adds support for a `SupportDir=FooBar` commandline argument to force a specified support directory.  The main motivation is to simplify testing changes relating to map installation on dedicated servers, but this could also be used by mods that want to manage a self-contained engine distribution that is automatically versioned and upgraded.

### DIFF
--- a/OpenRA.Game/Platform.cs
+++ b/OpenRA.Game/Platform.cs
@@ -63,11 +63,26 @@ namespace OpenRA
 			}
 		}
 
-		public static string SupportDir { get { return supportDir.Value; } }
-		static Lazy<string> supportDir = Exts.Lazy(GetSupportDir);
+		public static string SupportDir { get; private set; }
 
-		static string GetSupportDir()
+		public static void Initialize(Arguments args)
 		{
+			SupportDir = FindSupportDir(args);
+		}
+
+		static string FindSupportDir(Arguments args)
+		{
+			// Check for support directory overrides
+			if (args.Contains("SupportDir"))
+			{
+				var path = args.GetValue("SupportDir", "");
+				var separator = Path.DirectorySeparatorChar.ToString();
+				if (!path.EndsWith(separator))
+					path += separator;
+
+				return path;
+			}
+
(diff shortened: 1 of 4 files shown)


OpenRA.Game/Platform.cs
```diff
@@ -63,11 +63,26 @@ public static string RuntimeVersion
...
-		static string GetSupportDir()
+		public static void Initialize(Arguments args)
 		{
+			SupportDir = FindSupportDir(args);
+		}
+
+		static string FindSupportDir(Arguments args)
+		{
+			// Check for support directory overrides
+			if (args.Contains("SupportDir"))
```
obrakmann: This will need some way to be picked up by the utility so that it can be added to the man page.

pchote:
I don't see an simple solution for exposing this to the man page (the `LaunchArguments` object isn't created until *much* later in the init), nor am I really interested in adjusting our man page stuff so that I can install and test it under OSX.

This is supposed to be a development helper, so people can cherry pick this commit instead.

GraionDilach:
Requesting to reopen and merge this as-is as the followup to the discussion taking place in #11301, please.

Mailaender:
Just add it manually to `CreateManPage` and we can get it merged.

GraionDilach:
Okay, I'm testing this commit atm.

It only works with the / character being the input path separator, even on Windows (where one would assume \ could also be used).
It doesn't see the mods in the support folder.

image

I dunno what I'm doing wrong.

GraionDilach:
These arguments should also be passable to Utility and Test, otherwise mods in the custom support folders cannot be linted.

Phrohdoh:
I had difficulty linting thirdparty mods on OS X previously ([commit here](https://github.com/Phrohdoh/OpenRA/commit/acbd959f5798bc645869af17244a03a5003aece6)).
This may be something to consider.

pchote:
@GraionDilach that's a good point which I hadn't considered.  We'll need to think about how to solve this more generally.

PR #10992, opened 2016-03-28
Allow Dedicated Servers to query map info from the Resource Center.

### LINKED ISSUE #3558: Maps not available on the server should not be selectable
but greyed out in the dedicated host scenario.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/459903-maps-not-available-on-the-server-should-not-be-selectable?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #10690: Re-read map files on dedicated servers without restarting it
I'm getting a few questions regarding the map sync on my servers if they are working or not, because when people join my servers and try to pick a map they've uploaded to resource center for greater than 5 minutes ago, they get "Map not found on server" - Even though it's there.

This seems to be because the maps are only loaded on server start (Same with client).

Is there any chance to put in a looping check of map files after the mod has been loaded? So the servers check the <mod>\maps every X minutes? (At least for dedicated servers, as these are usually unattended)

This way the servers don't need  to be restarted / play a game before they become available.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30515982-re-read-map-files-on-dedicated-servers-without-restarting-it?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #10717: Clean up the filesystem code

pchote:
Dedicated servers will now query the map API in the same way that clients do.
This removes the need to manually rsync maps.

Closes #3558.
Closes #10690.
Closes #10717.

### DIFF
--- a/OpenRA.Game/Map/MapPreview.cs
+++ b/OpenRA.Game/Map/MapPreview.cs
@@ -17,6 +17,7 @@ using System.Drawing;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Text;
 using System.Threading;
 using OpenRA.FileSystem;
 using OpenRA.Graphics;
@@ -51,6 +52,9 @@ namespace OpenRA
 		public readonly MapGridType map_grid_type;
 		public readonly string minimap;
 		public readonly bool downloading;
+		public readonly string tileset;
+		public readonly string rules;
+		public readonly string players_block;
 	}
 
 	public class MapPreview : IDisposable, IReadOnlyFileSystem
@@ -334,7 +338,7 @@ namespace OpenRA
 			return true;
 		}
 
-		public void UpdateRemoteSearch(MapStatus status, MiniYaml yaml)
+		public void UpdateRemoteSearch(MapStatus status, MiniYaml yaml, Action<MapPreview> parseMetadata = null)
 		{
 			var newData = innerData.Clone();
 			newData.Status = status;
@@ -358,14 +362,31 @@ namespace OpenRA
 					newData.Author = r.author;
 					newData.PlayerCount = r.players;
 					newData.Bounds = r.bounds;
+					newData.TileSet = r.tileset;
 
 					var spawns = new CPos[r.spawnpoints.Length / 2];
 					for (var j = 0; j < r.spawnpoints.Length; j += 2)
 						spawns[j / 2] = new CPos(r.spawnpoints[j], r.spawnpoints[j + 1]);
 					newData.SpawnPoints = spawns;
 					newData.GridType = r.map_grid_type;
-
 					newData.Preview = new Bitmap(new MemoryStream(Convert.FromBase64String(r.minimap)));
+
+					var playersString = Encoding.UTF8.GetString(Convert.FromBase64String(r.players_block));
+					newData.Players = new MapPlayers(MiniYaml.FromString(playersString));
+
+					newData.SetRulesetGenerator(modData, () =>
+					{
+						var rulesString = Encoding.UTF8.GetString(Convert.FromBase64String(r.rules));
+						var rulesYaml = new MiniYaml("", MiniYaml.FromString(rulesString)).ToDictionary();
+						var ruleDefinitions = LoadRuleSection(rulesYaml, "Rules");
+						var weaponDefinitions = LoadRuleSection(rulesYaml, "Weapons");
+						var voiceDefinitions = LoadRuleSection(rulesYaml, "Voices");
+						var musicDefinitions = LoadRuleSection(rulesYaml, "Music");
+						var notificationDefinitions = LoadRuleSection(rulesYaml, "Notifications");
+						var sequenceDefinitions = LoadRuleSection(rulesYaml, "Sequences");
+						return Ruleset.Load(modData, this, TileSet, ruleDefinitions, weaponDefinitions,
+							voiceDefinitions, notificationDefinitions, musicDefinitions, sequenceDefinitions);
+					});
 				}
 				catch (Exception) { }
 
@@ -374,6 +395,9 @@ namespace OpenRA
 
 				if (innerData.Preview != null)
 					cache.CacheMinimap(this);
+
+				if (parseMetadata != null)
+					parseMetadata(this);
 			}
 
 			// Update the status and class unconditionally
--- a/OpenRA.Game/Map/MapPreview.cs
+++ b/OpenRA.Game/Map/MapPreview.cs
@@ -55,6 +55,65 @@ namespace OpenRA
 
 	public class MapPreview : IDisposable, IReadOnlyFileSystem
 	{
+		/// <summary>Wrapper that enables map data to be replaced in an atomic fashion</summary>
+		class InnerData
+		{
+			public string Title;
+			public string[] Categories;
+			public string Author;
+			public string TileSet;
+			public MapPlayers Players;
+			public int PlayerCount;
+			public CPos[] SpawnPoints;
+			public MapGridType GridType;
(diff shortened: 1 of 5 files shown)


pchote:
I recommend cherry-picking #10981 to make testing this less painful on a single machine.

pchote:
http://resource.openra.net/maps/9515/ and http://resource.openra.net/maps/9514/ are available on the resource center to test with (delete them locally from the server and/or client first).

OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
```diff
@@ -338,64 +339,87 @@ public bool InterpretCommand(S server, Connection conn, Session.Client client, s
...
 							}
-							else if (c.Bot != null)
-								server.LobbyInfo.Clients.Remove(c);
-						}
 
-						// Validate if color is allowed and get an alternative it isn't
-						foreach (var c in server.LobbyInfo.Clients)
-							if (c.Slot == null || (c.Slot != null && !server.LobbyInfo.Slots[c.Slot].LockColor))
-								c.Color = c.PreferredColor = SanitizePlayerColor(server, c.Color, c.Index, conn);
+							// Validate if color is allowed and get an alternative it isn't
```
Phrohdoh: `if it isn't`?

OpenRA.Game/Map/MapPreview.cs
```diff
@@ -305,11 +310,43 @@ public void UpdateRemoteSearch(MapStatus status, MiniYaml yaml)
 					GridType = r.map_grid_type;
 
 					Preview = new Bitmap(new MemoryStream(Convert.FromBase64String(r.minimap)));
+
+					var playersString = Encoding.UTF8.GetString(Convert.FromBase64String(r.players_block));
+					Players = new MapPlayers(MiniYaml.FromString(playersString));
```
obrakmann: Wouldn't it make sense to have this in a try/catch block as well?

obrakmann:
Seems to work fine, :+1:

OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
pchote: I don't understand your comment here...?  In any case, this is just moving existing code.

pchote:
Fixed.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -814,8 +823,12 @@ void UpdateCurrentMap()
 					});
 				}).Start();
 			}
+			else if (Map.Status == MapStatus.DownloadAvailable)
+				LoadMapPreviewRules(Map);
 			else if (Game.Settings.Game.AllowDownloading)
-				modData.MapCache.QueryRemoteMapDetails(new[] { uid });
+			{
```
abcdefg30: Braces around a single-line statement?

pchote: These were left over from an earlier implementation that had multiple lines here.  Fixed.

OpenRA.Game/Map/MapPreview.cs
```diff
@@ -272,47 +276,81 @@ bool EvaluateUserFriendliness(Dictionary<string, PlayerReference> players)
...
-			// Update on the main thread to ensure consistency
-			Game.RunAfterTick(() =>
+			if (status == MapStatus.DownloadAvailable)
 			{
-				if (status == MapStatus.DownloadAvailable)
+				// HACK/PERF: The following is very much not thread safe for code (mainly UI) that uses the
+				// properties. We can get away without locking because the outside code must always check the
+				// Status field, and knows not to access it unless it is Available or DownloadAvailable.
+				// This means that no other threads should be accessing these info fields until *after* the
+				// Status field is set to MapStatus.DownloadAvailable at the end of the method.
```
RoosterDragon: This is no good - without some sort of synchronisation the UI thread may well never pick up these changes. One option might be to abstract all the data into a single volatile field and replace that field with a new reference you build here.

pchote: The UI thread picks up the changes from the `Status` field changing.  That's what I was trying to explain with the comment.

pchote: Using a single field with references into it sounds like a decent solution – certainly better than having to write a pile of getters with explicit locks.

pchote: Done.

pchote:
Updated and removed the original :+1: because the update changed a fundamental part of the PR.

obrakmann:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/4877914d6f211c77d1a6c6ccc642bc3a2c2903d8...9608ab5fa8174518d46d0a07e10baa3d8fd333f7)

PR #10995, opened 2016-03-28
Allow empty harvesters to be ordered to refineries

### LINKED ISSUE #161: Empty harvesters can not teleport to a refinery
We are using `INotifyHarvesterAction.MovingToRefinery` at the moment which doesn't support this.

ghost:
The primary reason is to fix OpenRA/ra2#161. Rather than make the
harvester trait more complicated by making this behaviour configurable,
this commit just applies to all harvesters.

The primary impact on gameplay is, I believe, that ordering a harvester out of a combat zone now has another option that requires a little bit less micro since the harvester will go back to harvesting once it's gone back. The behaviour here matches that of the original RA2: the harvester will dock and unload even though it is empty.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Harvester.cs
+++ b/OpenRA.Mods.Common/Traits/Harvester.cs
@@ -302,7 +302,7 @@ namespace OpenRA.Mods.Common.Traits
 			{
 				yield return new EnterAlliedActorTargeter<IAcceptResourcesInfo>("Deliver", 5,
 					proc => IsAcceptableProcType(proc),
-					proc => !IsEmpty && proc.Trait<IAcceptResources>().AllowDocking);
+					proc => proc.Trait<IAcceptResources>().AllowDocking);
 				yield return new HarvestOrderTargeter();
 			}
 		}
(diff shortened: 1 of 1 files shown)


Mailaender:

Phrohdoh:
I don't think we should consider this a good thing.
>the harvester will dock and unload even though it is empty.

Teleporting back is fine but there is no reason to dock if the harv is empty.

But this works and is sensible, the above isn't really a blocker. :+1:

scshunt:
I'd be open to changing that as part of another PR. It wouldn't be hard; just change the activity to check for emptiness before beginning the docking procedure.

Mailaender:
>  The behaviour here matches that of the original RA2

This is good enough. I agree that it can be tweaked later.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/cad34ff14e796ac322f851ddf840b41085cd15d1...64249a8731952e823fccebcab594a2a91cc800b6)

PR #11002, opened 2016-03-29
Pillbox/CamPillbox blind spot fix

### LINKED ISSUE #9529: Pillbox does not shoot infantry on a specific spot
force fire gets rejected ... and it does not get triggered when its a enemy unit 

pillbox fire

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27145347-pillbox-does-not-shoot-infantry-on-a-specific-spot?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

HenrytheSlav:
Fixes #9529. It is just lame hack, but it works. If you feel like this could break something then tell me and I'll throw it away/change values. Increasing by 1 was not enough so I arbitrarily chose 100.

### DIFF
--- a/mods/ra/rules/structures.yaml
+++ b/mods/ra/rules/structures.yaml
@@ -568,7 +568,7 @@ PBOX:
 		Armaments: garrisoned
 		PortOffsets: 384,0,128, 224,-341,128, -224,-341,128, -384,0,128, -224,341,128, 224,341,128
 		PortYaws: 0, 176, 341, 512, 682, 853
-		PortCones: 86, 86, 86, 86, 86, 86
+		PortCones: 88, 88, 88, 88, 88, 88
 	RenderRangeCircle:
 		FallbackRange: 6c0
 	Power:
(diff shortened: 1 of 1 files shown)


mods/ra/rules/structures.yaml
```diff
@@ -568,7 +568,7 @@ PBOX:
 		Armaments: garrisoned
 		PortOffsets: 384,0,128, 224,-341,128, -224,-341,128, -384,0,128, -224,341,128, 224,341,128
 		PortYaws: 0, 176, 341, 512, 682, 853
-		PortCones: 86, 86, 86, 86, 86, 86
+		PortCones: 100, 100, 100, 100, 100, 100
```
pchote: IIRC from my own earlier tests increasing these to 87 should be sufficient and will avoid graphical glitches (firing out the wrong hole).

GraionDilach: I'd prefer 88 though, 87 might have rounding issues.

HenrytheSlav: As I said, 87 didn't work.Might try use lowest working. I have map with tons of unmoving engineers in every cell around pbox to test with.

HenrytheSlav:
Updated to 88, that works and should cause no issues.

GraionDilach:
:+1:

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/b7d86f4cf5ca8acc0876677b193d8fb66488d2e1...e9cea306e1e0e3ef579975143057ce4f8a0588d3)

PR #11005, opened 2016-03-30
Extract dedicated server interface to OpenRA.Server.exe

### LINKED ISSUE #7252: Create a true headless OpenRA.Server.exe
instead of always checking against `Game.Settings.Server.Dedicated` and deciding individually whether it should be loaded.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7440668-create-a-true-headless-openra-server-exe?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #9065: settings.yaml is conflicting when running multiple servers on same users
I'm having some issues with running multiple servers on same users (Linux verified).

It seems on start the server writes the settings to settings.yaml in .openra/ directory.

Earlier I've noticed a problem where the server name could actually be affected by this and duplicate, if the servers started up too fast (Sorted with adding a sleep in the script) - But I've experienced another "worse" problem now.

Even though I'm starting my servers first, without any Server.Password, if I then start other servers, on the same user, but with Server.Password parameter, the first servers too, get locked (I guess it re-reads settings.yaml every now and then, and finds the Password-parameter).

Is there any way to create a settings.yaml per server? Maybe settings-$PID.yaml if it's easy to get the PID on all OS'es.

It's very "time consuming/waste" to create own user for every server you might want to change the settings on.

Hope there's a solution to a more multi-setup friendly environment :)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25757690-settings-yaml-is-conflicting-when-running-multiple-servers-on-same-users?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #10690: Re-read map files on dedicated servers without restarting it

pchote:
The dedicated server mode has always been bolted on to the side of the game – this extracts it into its own executable where it belongs.  This and #10992 take us a big step towards our long-planned dedicated server overhaul.

Fixes #7252.
Fixes #9065.
Closes #10690.

Partially reverts b2df8d3, c419127.

Ping @baxxxster: this will probably break your bleed server auto-updating script.


OpenRA.Game/Game.cs
```diff
@@ -279,7 +279,7 @@ internal static void Initialize(Arguments args)
 				}
 			}
 
-			Sound = new Sound(Settings.Server.Dedicated ? "Null" : Settings.Sound.Engine);
+			Sound = new Sound(Settings.Sound.Engine);
```
obrakmann: Do we even need `OpenRA.Platforms.Null` anymore for anything after this, or can we finally get rid of it?

obrakmann:
1. You'll need to add `OpenRA.Server.exe` to the `FILES` array in n`packaging/package-all.sh`.
2. You'll need to add it to `OpenRA.nsi` as well.

OpenRA.Server/Program.cs
```diff
@@ -0,0 +1,64 @@
...
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
```
Mailaender: Are all these `using`s required? I don't see a LINQ call for example.

OpenRA.Server/Program.cs
```diff
@@ -0,0 +1,64 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
```
Mailaender: 2016 and GPLv3 etc.

Mailaender:
Great job. Bonus points if you add an `openra-server` man page as well.

pchote:
Fixed everything except the man page, which I'd prefer to steer clear of.

OpenRA.Game/Game.cs
pchote: Removed.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/9dc8d2ed8adce84e4891dafc76777729bb5f9ad7...cad34ff14e796ac322f851ddf840b41085cd15d1)

PR #11014, opened 2016-04-01
Added CanUndeploy property

ghost:


OpenRA.Mods.Common/Traits/Upgrades/DeployToUpgrade.cs
```diff
@@ -53,6 +53,9 @@ public class DeployToUpgradeInfo : ITraitInfo, Requires<UpgradeManagerInfo>
 		[Desc("Sound to play when undeploying.")]
 		public readonly string UndeploySound = null;
 
+		[Desc("Is the initial deployment of this actor permanent?")]
+		public readonly bool IsDeploymentPermanent = false;
```
pchote: `Permanent` would be sufficient as a name. `Deployment` is redundant, and inconsistent with our usual `Deploy` terminology.

dtluna:
@pchote Will CanUndeploy be better?
P.S. I originally named it that way, but mr. Phrohdoh convinced me to change it to the current variant.

pchote:
CanUndeploy is fine by me.

GraionDilach:
Uhm, what is the point of this? Could you explain the usecase?

dtluna:
@GraionDilach I need the Reckoner APC in my mod to be not able to undeploy.

GraionDilach:
Because? I asked for an explanation, not an example, because all the examples I would think about - notundeployable mobile buildings á lá RA3 Empire of the Rising Sun - are more fitting for `Transforms`. A random unit name from a mod I'm not familiar with ain't explain this to me.

dtluna:
@GraionDilach Because this is how this unit from C&C3:KW works. It is a pretty complex unit and using `Transforms` for this will be not very convinient.

GraionDilach:
Okay, so, now that I looked at the C&C Wikia...

"Because this unit is a mobile bunker which can be deployed to lose mobility in favor of more strength and `Transforms` on this would be messy."

`CanUndeploy` is better, yes.

GraionDilach:
:+1:

abcdefg30:
You probably also want to disable the cursor when you disable undeploying.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/e0039c82813f746304524d0de31708304011f51e...3bd5d502a3b2568874b0e3ab9e0704d601baa872)

PR #11015, opened 2016-04-01
Fix paratroopers on heightmapped terrain

abcdefg30:
You can't see the effect in the default mods though (only ra2 will show one).
Anyway, all things that use `Mobile.SetPosition`\* seem to still work fine.

\* e.g. DeliverUnit, SwallowActor, Teleport and Cargo (EjectOnDeath there).

### DIFF
--- a/OpenRA.Mods.Common/Activities/Parachute.cs
+++ b/OpenRA.Mods.Common/Activities/Parachute.cs
@@ -54,7 +54,7 @@ namespace OpenRA.Mods.Common.Activities
 
 		Activity LastTick(Actor self)
 		{
-			pos.SetPosition(self, currentPosition - new WVec(0, 0, currentPosition.Z));
+			pos.SetPosition(self, currentPosition);
 
 			if (um != null)
 				foreach (var u in para.ParachuteUpgrade)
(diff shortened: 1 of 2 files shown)


Phrohdoh:
Checked `SendParatroopers` and other general `Mobile` actors.
:+1: once the commit message is corrected from `highmapped` to `heightmapped`.

Mailaender:
Still off ~1 cell, but definitely better.

abcdefg30:
> once the commit message is corrected from  highmapped  to  heightmapped .

Oops, my english gud.

abcdefg30:
> Still off ~1 cell, but definitely better.

Hm, are you sure? For me the paratroopers landed on the correct spot (before they visually landed somewhere below and then "teleported up" to their real position).

Mailaender:
Must have tested this on the wrong branch, sorry. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/64249a8731952e823fccebcab594a2a91cc800b6...fa1fc930514c40a7b205b637a55aa195bd146c9b)

abcdefg30:
Seems like this regressed crate collection in ra somehow.

OpenRA.Mods.Common/Activities/Parachute.cs
```diff
@@ -54,7 +54,7 @@ Activity FirstTick(Actor self)
 
 		Activity LastTick(Actor self)
 		{
-			pos.SetPosition(self, currentPosition - new WVec(0, 0, currentPosition.Z));
```
Mailaender: Is it maybe this change which should rather be
```csharp
pos.SetPosition(self, currentPosition - new WVec(0, 0, self.World.Map.DistanceAboveTerrain(currentPosition).Length))
```
?

PR #11017, opened 2016-04-02
Grouped the radar traits together

Mailaender:
Just moving code that belongs together.


pchote:
If you moved these into their own namespace then they would appear grouped on https://github.com/OpenRA/OpenRA/wiki/Traits too.

Mailaender:
Good idea. Done.

abcdefg30:
Well, looks simple enough. :+1: / :white_check_mark:

GraionDilach:
TBH, I thought @pchote already thought about removing `JamsRadar` in favor of an `UpgradesActorsNear`-disable radar upgrade combination so I'm not certain if this worths the effort (especially the separate namespace aspect).

Mailaender:
I doubt anyone is working on that at the moment.

Issue #9962, opened 2015-11-11
Suggestion: Hide GPS icons when player has low power [$10]


r34ch:
The GPS launch count down is paused when the Allied tech centre has low power but once launched, GPS dots can be seen even when the player has low power.

This means that the only way to sneak attack allied players after this point is to destroy the tech centre. This reduces the effectiveness of attacking a players power supply.

pi99y:
I'm all for limiting the GPS!
- GPS diminishes importance of scouting
- like mentioned, this would allow sneak attacks
- the change is logical and since the Allied are supposed to be the "tech" faction this is the downside (more dependent on power)

GraionDilach:
@penev92 S_i_ggistion?

penev92:
:+1: 
As a one-up I'd like to suggest "Hide GPS icons when player has low power or the radar is offline/non-existent".

pi99y:
:+1:  to @penev92 's radar suggestion

klaas1:
Agree with all of the above, GPS is too strong and relatively easy to maintain once up.

Issue #7605, opened 2015-03-07
More benefits for keeping the radar online


Snurre86:
Possible handicaps for off-lining the radar:
1. Not reveal the fog of war around it if offline.
2. Not detect cloaked units if offline.
3. Still use -20 power when offline.
4. Same health as fake radar when offline.
5. Radar tech tree is disabled if radar is offline.

Possible benefits for on-lining the radar:
1. Make the radar screen more attractive (details not coming).
2. Make units that need radar to be built shot 1 cell longer if radar is online.
3. Make units that need radar to be built more accurate if the radar is online.
4. Make units that need radar to be built see 1 cell longer if radar is online.

Other:
I think there should be an arrow on the screen showing where a beacon is placed before a radar has been possible to build or on lower tech levels.

![openra game_2015_03_10_09_56_12_884](https://cloud.githubusercontent.com/assets/7798635/6571726/cc6606fe-c70b-11e4-81f7-4eea39f66b2b.png)
##

deniz1a:
Same health as fake radar is not good.

Mailaender:
See also https://github.com/OpenRA/OpenRA/issues/3769.

abcdefg30:
> 2) Not detect cloaked units if offline.
> 4) Don't detect cloaked units.

What's the difference here?

Snurre86:
denisuana you can't just add a statement with no arguments.

ScottNZ:
I proposed a while ago to have the radar dome (and other buildings that require power) not provide prerequisites if they are powered down. So you'll need a powered radar dome to build things that require a radar dome etc

Snurre86:
ScottNZ: Well it would help some but you would still just power on and off as needed?

ScottNZ:
Some important units require the radar dome (hind, arty are two notable Allied examples), so I can see people leaving it up permanently.

Snurre86:
ScottNZ: what happens to the hind, arty when radar is offline you suggest?

ScottNZ:
Nothing, but since most players build them constantly it should be fine to do this.

Snurre86:
Some good players don't tech up at all I find, I don't want to encourage that I think.

ScottNZ:
Not really convinced that this is a bad idea tbh

Snurre86:
Added to possible handicaps, I don't know if it alone is enough, that's all.

MicroJOo:
[...]

PR #11018, opened 2016-04-02
Fixed GPS power not taking tech center disablement into account

Mailaender:
and as a bonus make it depend on an active radar to punish power down there as well, see also https://github.com/OpenRA/OpenRA/issues/7605. Closes https://github.com/OpenRA/OpenRA/issues/9962.

### DIFF
--- a/OpenRA.Mods.RA/Traits/SupportPowers/GpsPower.cs
+++ b/OpenRA.Mods.RA/Traits/SupportPowers/GpsPower.cs
@@ -34,7 +34,7 @@ namespace OpenRA.Mods.RA.Traits
 		public override object Create(ActorInitializer init) { return new GpsPower(init.Self, this); }
 	}
 
-	class GpsPower : SupportPower, INotifyKilled, INotifySold, INotifyOwnerChanged
+	class GpsPower : SupportPower, INotifyKilled, INotifySold, INotifyOwnerChanged, ITick
 	{
 		readonly GpsPowerInfo info;
 		GpsWatcher owner;
@@ -82,6 +82,22 @@ namespace OpenRA.Mods.RA.Traits
 			RemoveGps(self);
 			owner = newOwner.PlayerActor.Trait<GpsWatcher>();
 			owner.GpsAdd(self);
+		}
+
+		bool wasDisabled;
+
+		public void Tick(Actor self)
+		{
+			if (!wasDisabled && self.IsDisabled())
+			{
+				wasDisabled = true;
+				RemoveGps(self);
+			}
+			else if (wasDisabled && !self.IsDisabled())
+			{
+				wasDisabled = false;
+				owner.GpsAdd(self);
+			}
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.RA/Traits/SupportPowers/GpsPower.cs
```diff
@@ -31,17 +31,22 @@ class GpsPowerInfo : SupportPowerInfo
 		[SequenceReference("SatelliteImage")] public readonly string SatelliteSequence = "idle";
 		[PaletteReference] public readonly string SatellitePalette = "effect";
 
+		[Desc("Requires an actor with `ProvidesRadar` to show GPS dots.")]
+		public readonly bool RequiresRadar = true;
```
pchote: We already have a system in place for declaring prerequisites!  Please make this check for prereqs in the tech tree instead of looking for a hardcoded trait type.

Mailaender: You are misunderstanding what this is doing. I should probably call it `RequiresActiveRadar`.

Mailaender:
Updated.

abcdefg30:

I think we should try to get this in the next playtest to get quick feedback.
What do you think @Micr0Bit , @SoScared ?

mods/ra/rules/structures.yaml
```diff
@@ -776,13 +776,16 @@ ATEK:
 		OneShot: yes
 		ChargeTime: 480
 		Description: GPS Satellite
-		LongDesc: Reveals map terrain and provides tactical\ninformation.
+		LongDesc: Reveals map terrain and provides tactical\ninformation. Requires power and radar.
```
kyrylo: s/radar/active radar/, perhaps?

kyrylo:
Sounds reasonable :+1:

SoScared:
Sounds great, including the radar dome dependency. Logical. With the change the Allied tech tree is a bit more vulnerable - the Allied Tech Center could perhaps use a small health buff to counteract cheap parabomb- and double-YAK snipes.

mods/ra/rules/structures.yaml
Mailaender: Yes, good idea. Done.

Micr0Bit:
im dumping this here from the irc :

no matter what ... you would shift the "crono" game 12 seconds ahead ... which means soviets have no chance reacting to that with a curtain (when you allow manually powering ATEK down )

only allies getting punished here for not having the radar-dome online ... that would be unfair

Mailaender:
Removed https://github.com/OpenRA/OpenRA/pull/11018/commits/f8c433707d68d6478fbf345b48f843d28e20c7e6 again to avoid the imbalance.

pchote:
I suspect we want this for the next playtest since it has strong balance implications?

Mailaender:
I don't want to risk destabilizing it. This can be tested in the next release cycle.

Phrohdoh:
Or it could be reverted if it is found to be a negative change.
That is the purpose of the playtests.

Mailaender:
If you make lot's of exceptions for feature freeze, then you end up with an endlessly postponed release.

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/249faef3c39530cb537eb61d6b26831e4a260443...c84a35981245013dd50c6cc9124ded4bd93b457f)

PR #11021, opened 2016-04-02
Added more support for multiple Mobile

Mailaender:
Actor upgrades are in theory a good idea, but they completely fail with simple things like replacing mobile properties. The engine doesn't like multiple `Mobile` on one actor. Fixed it as far as I could. It will now crash later with "Actor has multiple traits of type `OpenRA.Traits.IOccupySpace`", but I am not able to resolve that.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Production.cs
+++ b/OpenRA.Mods.Common/Traits/Production.cs
@@ -140,7 +140,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		static bool CanUseExit(Actor self, ActorInfo producee, ExitInfo s)
 		{
-			var mobileInfo = producee.TraitInfoOrDefault<MobileInfo>();
+			var mobileInfo = producee.TraitInfos<MobileInfo>().FirstOrDefault(Exts.IsTraitEnabled);
 
 			self.NotifyBlocker(self.Location + s.ExitCell);
 
--- a/OpenRA.Mods.Common/Traits/World/DomainIndex.cs
+++ b/OpenRA.Mods.Common/Traits/World/DomainIndex.cs
@@ -29,9 +29,9 @@ namespace OpenRA.Mods.Common.Traits
 		{
 			domainIndexes = new Dictionary<uint, MovementClassDomainIndex>();
 			var tileSet = world.Map.Rules.TileSet;
-			var movementClasses =
-				world.Map.Rules.Actors.Where(ai => ai.Value.HasTraitInfo<MobileInfo>())
-					.Select(ai => (uint)ai.Value.TraitInfo<MobileInfo>().GetMovementClass(tileSet)).Distinct();
+
+			var movementClasses = world.Map.Rules.Actors.Values.SelectMany(a => a.TraitInfos<MobileInfo>())
+				.Select(m => (uint)m.GetMovementClass(tileSet)).Distinct();
 
 			foreach (var mc in movementClasses)
 				domainIndexes[mc] = new MovementClassDomainIndex(world, mc);


pchote:
> but they completely fail with simple things like replacing mobile properties

We have very different ideas of simple! IMO `Mobile` is by far our most complicated piece of gameplay code.

OpenRA.Mods.Common/Traits/World/DomainIndex.cs
```diff
@@ -29,9 +29,16 @@ public void WorldLoaded(World world, WorldRenderer wr)
...
+
+			var movementClasses = new List<uint>();
+			foreach (var actor in world.Map.Rules.Actors.Values)
+			{
+				var mobileInfos = actor.TraitInfos<MobileInfo>();
+				foreach (var mobileInfo in mobileInfos)
+					movementClasses.Add((uint)mobileInfo.GetMovementClass(tileSet));
+			}
+
+			movementClasses.Distinct();
```
pchote: `Distinct` returns a new collection, so this does nothing.

Mailaender: Fixed.

OpenRA.Mods.Common/Traits/World/DomainIndex.cs
```diff
@@ -29,11 +29,16 @@ public void WorldLoaded(World world, WorldRenderer wr)
 		{
 			domainIndexes = new Dictionary<uint, MovementClassDomainIndex>();
 			var tileSet = world.Map.Rules.TileSet;
-			var movementClasses =
-				world.Map.Rules.Actors.Where(ai => ai.Value.HasTraitInfo<MobileInfo>())
-					.Select(ai => (uint)ai.Value.TraitInfo<MobileInfo>().GetMovementClass(tileSet)).Distinct();
```
pchote: Use SelectMany and TraitInfos.  You don't need to roll your own list.

OpenRA.Mods.Common/Traits/Production.cs
```diff
@@ -140,7 +140,7 @@ public virtual bool Produce(Actor self, ActorInfo producee, string factionVarian
 
 		static bool CanUseExit(Actor self, ActorInfo producee, ExitInfo s)
 		{
-			var mobileInfo = producee.TraitInfoOrDefault<MobileInfo>();
+			var mobileInfo = producee.TraitInfos<MobileInfo>().FirstOrDefault(t => t.IsTraitEnabled());
```
pchote: `.FirstOrDefault(Exts.IsTraitEnabled)`

OpenRA.Mods.Common/Traits/World/DomainIndex.cs
Mailaender: Updated.

OpenRA.Mods.Common/Traits/World/DomainIndex.cs
```diff
@@ -29,11 +29,11 @@ public void WorldLoaded(World world, WorldRenderer wr)
...
 
-			foreach (var mc in movementClasses)
+			var movementClasses = world.Map.Rules.Actors.Values.SelectMany(a => a.TraitInfos<MobileInfo>())
+				.Select(m => (uint)m.GetMovementClass(tileSet)).Distinct();
+
+			foreach (var mc in movementClasses.Distinct())
```
pchote: You don't need a second Distinct here.

Mailaender: Indeed. Fixed.

Phrohdoh:
> It will now crash later with "Actor has multiple traits of type OpenRA.Traits.IOccupySpace", but I am not able to resolve that.

Based off of that alone I don't see why we'd take this.

Mailaender:
Because it fixes 2 known crashes.

reaperrr:
> It will now crash later with "Actor has multiple traits of type `OpenRA.Traits.IOccupySpace`", but I am not able to resolve that.

That particular one is rather easy to resolve (replace the handful of info lookups with `TraitInfos<IOccupySpaceInfo>().FirstOrDefault(Exts.IsTraitEnabled)` and the trait lookup in `Actor` with `TraitsImplementing<IOccupySpace>().First()` for now), the problem is that it's only the beginning. `IMove` and `IFacing` are where this gets messy.

reaperrr:
Sorry, but this isn't really a step forward on its own. It's not even decided (or rather, we probably won't) whether we'll make Mobile properly upgradable the way it is now, so I'm closing this to clear the PR queue.

PR #11028, opened 2016-04-03
Fix parachutes in non-heightmapped mods.

pchote:
#11015 put the actors at the wrong height, which stops them from being crushable.  This broke crates in RA.


abcdefg30:
:+1: / :white_check_mark: Thanks.

Mailaender:
I think this regresses RA2 again (the off by 1 cell problem on landing), but that isn't a priority now.

PR #11029, opened 2016-04-03
Fix Turreted units not dropping cancelled targets.

### LINKED ISSUE #11027: "Stop" is invalid for defensive structures
Cannot stop any defensive structure from force attacking. I tried pressing "S", but it didn't work. It only works for infantries and vehicles.

pchote:
Fixes #11027.

I don't see a cleaner way to work around this without doing a major rewrite.
The attack code is so horrible :disappointed:

### DIFF
--- a/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
@@ -66,6 +66,7 @@ namespace OpenRA.Mods.Common.Traits
 			readonly Target target;
 			readonly bool forceAttack;
 			readonly bool onRailsHack;
+			bool hasTicked;
 
 			public AttackActivity(Actor self, Target target, bool allowMove, bool forceAttack)
 			{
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
```diff
@@ -103,7 +104,13 @@ public override Activity Tick(Actor self)
 					var maxRange = targetIsMobile ? new WDist(Math.Max(weapon.Weapon.MinRange.Length, modifiedRange.Length - 1024))
 						: modifiedRange;
 
+					// Check that AttackFollow hasn't cancelled the target by modifying attack.Target
+					// Having both this and AttackFollow modify that field is a horrible hack.
```
Mailaender: > this and AttackFollow

Isn't this also `AttackFollow`?

pchote: No, this is `AttackActivity.AttackActivity`.

Mailaender:
Fixes the problem as advertised. :white_check_mark:

obrakmann:
:+1: works and doesn't impact the original bug fix. The gunboat in gdi01 continues to work as well.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/abfdd5a44d3aa17631772323bfb7290eacd7f5f5)

PR #10771, opened 2016-02-16
Add a utility command to output the miniyaml tree of an actor type

phrohdoh:
When reviewing please ignore comments in between this description and [this comment](https://github.com/OpenRA/OpenRA/pull/10771#issuecomment-184809619).

`$ orautil ra --miniyaml-tree-actor E1`

``` yaml
E1:
    AppearsOnRadar: 
    CombatDebugOverlay: 
    GivesExperience: 
    ScriptTriggers: 
    UpgradeManager: 
    Huntable: 
    RenderDebugState: 
    GainsExperience: 
        Upgrades: 
            200: firepower, damage, speed, reload, inaccuracy, rank
            400: firepower, damage, speed, reload, inaccuracy, rank
            800: firepower, damage, speed, reload, inaccuracy, rank
            1600: firepower, damage, speed, reload, inaccuracy, rank, eliteweapon, selfheal
    DamageMultiplier@EXPERIENCE: 
        UpgradeTypes: damage
        Modifier: 95, 90, 85, 75
    FirepowerMultiplier@EXPERIENCE: 
        UpgradeTypes: firepower
        Modifier: 105, 110, 120, 130
    SpeedMultiplier@EXPERIENCE: 
        UpgradeTypes: speed
        Modifier: 105, 110, 120, 140
    ReloadDelayMultiplier@EXPERIENCE: 
        UpgradeTypes: reload
        Modifier: 95, 90, 85, 75
    InaccuracyMultiplier@EXPERIENCE: 
        UpgradeTypes: inaccuracy
        Modifier: 90, 80, 70, 50
    SelfHealing@ELITE: 
        Step: 2
        Ticks: 100
        HealIfBelow: 1
        DamageCooldown: 125
        UpgradeTypes: selfheal
        UpgradeMinEnabledLevel: 1
    WithRankDecoration: 
        Image: rank
        Sequence: rank
        Palette: effect
        ReferencePoint: Bottom, Right
        UpgradeTypes: rank
        ZOffset: 256
        UpgradeMinEnabledLevel: 1
        UpgradeMaxAcceptedLevel: 4
    BodyOrientation: 
    QuantizeFacingsFromSequence: 
        Sequence: stand
    AutoSelectionSize: 
    RenderSprites: 
    DrawLineToTarget: 
    Health: 
        HP: 50
        Shape: Circle
            Radius: 128
    Armor: 
        Type: None
    RevealsShroud: 
        Range: 4c0
    Mobile: 
        Speed: 56
[...]

PR #11031, opened 2016-04-03
Add a utility command to output an actor's MiniYaml tree

phrohdoh:
Possibly taking in a map path (to a .zip/.oramap/directory).

Usage: `orautil ts --actor-yaml World mods/ts/maps/arivruns/`

If you exclude the map path just the default ruleset will be used (files in the `Rules` section of the manifest).

GitHub won't let me reopen #10771.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/OutputActorMiniYamlCommand.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/OutputActorMiniYamlCommand.cs
@@ -1,0 +1,51 @@
+using System;
+using System.IO;
+using System.Linq;
+
+namespace OpenRA.Mods.Common.UtilityCommands
+{
+	public class OutputActorMiniYamlCommand : IUtilityCommand
+	{
+		string IUtilityCommand.Name { get { return "--actor-yaml"; } }
+
+		bool IUtilityCommand.ValidateArguments(string[] args)
+		{
+			return args.Length == 2 || args.Length == 3;
+		}
+
+		[Desc("ACTOR-TYPE [PATH/TO/MAP]", "Display the finalized, merged MiniYaml tree for the given actor type. Input values are case-sensitive.")]
+		void IUtilityCommand.Run(ModData modData, string[] args)
+		{
+			// HACK: The engine code assumes that Game.modData is set.
+			Game.ModData = modData;
+
+			var actorType = args[1];
+			string mapPath = null;
+
+			Map map = null;
+			if (args.Length == 3)
+				try
+				{
+					mapPath = args[2];
+					map = new Map(modData, modData.ModFiles.OpenPackage(mapPath));
+				}
+				catch (InvalidDataException)
+				{
+					Console.WriteLine("Could not load map '{0}'.", mapPath);
+					Environment.Exit(2);
+				}
+
+			var fs = map ?? modData.DefaultFileSystem;
+			var topLevelNodes = MiniYaml.Load(fs, modData.Manifest.Rules, map == null ? null : map.RuleDefinitions);
+
+			var result = topLevelNodes.FirstOrDefault(n => n.Key == actorType);
+			if (result == null)
+			{
+				Console.WriteLine("Could not find actor '{0}' (name is case-sensitive).", actorType);
+				Environment.Exit(1);
+			}
+
+			Console.WriteLine(result.Value.Nodes.WriteToString());
+		}
+	}
+}
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -755,6 +755,7 @@
     <Compile Include="UtilityCommands\ListMixContentsCommand.cs" />
     <Compile Include="UtilityCommands\ListMSCabContentsCommand.cs" />
     <Compile Include="FileFormats\MSCabCompression.cs" />
+    <Compile Include="UtilityCommands\OutputActorMiniYamlCommand.cs" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>


OpenRA.Mods.Common/UtilityCommands/OutputActorMiniYamlCommand.cs
```diff
@@ -0,0 +1,59 @@
...
+				Console.WriteLine("Could not find actor '{0}'.", actorType);
+				Console.WriteLine("Found keys: " + tree.Keys.JoinWith(","));
+				Environment.Exit(1);
+			}
+
+			Console.WriteLine(args[1] + ":");
+			RecursivePrintMiniYamlNodes(actorMiniYaml.Nodes, 1);
+		}
+
+		void RecursivePrintMiniYamlNodes(List<MiniYamlNode> nodes, int indent)
```
pchote: I'm pretty sure we already have code for this - we use it for saving miniyaml in e.g. the rule updater.

OpenRA.Mods.Common/UtilityCommands/OutputActorMiniYamlCommand.cs
```diff
@@ -0,0 +1,59 @@
...
+		{
+			foreach (var n in nodes)
+			{
+				Console.WriteLine(new string('\t', indent) + "{0}: {1}", n.Key, n.Value.Value);
+				RecursivePrintMiniYamlNodes(n.Value.Nodes, indent + 1);
+			}
+		}
+
+		Dictionary<string, MiniYaml> LoadYamlRules(IReadOnlyFileSystem fs, Manifest manifest, MiniYaml additionalMiniYaml)
```
pchote: Why extract a function for a standard one-liner?

pchote: You don't really need a dictionary either - just change the TryGetValue with a FirstOrDefault(n => n.Key ==)

Phrohdoh: I prefer the dictionary approach.
There is nothing that prevents a property from sharing a name with an actor which will only lead to confusion if that happens.

But if you (or someone else) insists on this being done with LINQ that is a two minute change I can do.

Phrohdoh:
Updated with most of @pchote's suggestions but I [kept the dictionary](https://github.com/OpenRA/OpenRA/pull/11031#discussion_r58310084).

This now outputs something similar to `# Rules for World (merged with map A River Runs Near It/042af00bc74f7a86536cb96cc89bc843a796c1c1):` as the first line of output before the actual rules.

OpenRA.Mods.Common/UtilityCommands/OutputActorMiniYamlCommand.cs
pchote: The query only applies to top-level nodes, so property names vs actor names doesn't come into this.

OpenRA.Mods.Common/UtilityCommands/OutputActorMiniYamlCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+
+			var actorType = args[1];
+			var mapPath = args[2];
+
+			Map map = null;
+			if (args.Length == 3)
+				map = new Map(modData, modData.ModFiles.OpenPackage(mapPath));
+
+			var fs = map ?? modData.DefaultFileSystem;
+			var tree = MiniYaml.Load(fs, modData.Manifest.Rules, map == null ? null : map.RuleDefinitions);
```
Phrohdoh: I'll rename this to `topLevelNodes` for clarity.

OpenRA.Mods.Common/UtilityCommands/OutputActorMiniYamlCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		[Desc("ACTOR-TYPE [PATH/TO/MAP]", "Display the merged MiniYaml tree for the given actor type. Input values are case-sensitive.")]
+		void IUtilityCommand.Run(ModData modData, string[] args)
+		{
+			// HACK: The engine code assumes that Game.modData is set.
+			Game.ModData = modData;
+
+			var actorType = args[1];
+			var mapPath = args[2];
```
Phrohdoh: I think this is bogus.

If `Length == 2`:
[0] will be the command
[1] the actor type
[2] **exception**

if `Length == 3`:
[0] will be the command
[1] the actor type
[2] the map path

Will fix.

Phrohdoh:
Rebased and updated.

OpenRA.Mods.Common/UtilityCommands/OutputActorMiniYamlCommand.cs
```diff
@@ -0,0 +1,40 @@
...
+		void IUtilityCommand.Run(ModData modData, string[] args)
+		{
+			// HACK: The engine code assumes that Game.modData is set.
+			Game.ModData = modData;
+
+			var actorType = args[1];
+
+			Map map = null;
+			if (args.Length == 3)
+				map = new Map(modData, modData.ModFiles.OpenPackage(args[2]));
```
RoosterDragon: You may wish to try-catch an invalid map path and exit gracefully.

OpenRA.Mods.Common/UtilityCommands/OutputActorMiniYamlCommand.cs
```diff
@@ -0,0 +1,40 @@
...
+
+			var fs = map ?? modData.DefaultFileSystem;
+			var topLevelNodes = MiniYaml.Load(fs, modData.Manifest.Rules, map == null ? null : map.RuleDefinitions);
+
+			var result = topLevelNodes.FirstOrDefault(n => n.Key == actorType);
+			if (result == null)
+			{
+				Console.WriteLine("Could not find actor '{0}'.", actorType);
```
RoosterDragon: Maybe include "(name is case-sensitive)" in the message?

RoosterDragon:
:+1:

obrakmann:
:+1: Looks good to me, other than the nits Rooster mentioned. Can you update this so it can be merged, please?

Phrohdoh:
Rebased and addressed both requests.

> tmba:openra thill $ orautil ts --actor-yaml E1 something/fake
Could not load map 'something/fake'.

Exited `2`.

> tmba:openra thill $ orautil ts --actor-yaml e1
Could not find actor 'e1' (name is case-sensitive).

Exited `1`

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5350a8f968587773a39c9dcf646ee46e8ca0269a)

PR #11058, opened 2016-04-06
Use vertex depths for rendering terrain depth preview.

pchote:
This implements the second and one of the most fiddly steps towards #7520.

Initial support for defining 3d positions is added, and `TerrainSpriteLayer` is set up to give each tile a z coordinate evaluated at the center of the tile.  The projection matrix is expanded to transform the raw z coordinates into normalized device coordinates... this was tricky to solve and is the most likely place to find bugs because any vertex mapped outside the range -1...1 it will be discarded, leaving a hole in the map.  Finally, the depth preview is adjusted to combine the vertex depth with the depth sprite to produce the final depth image that is drawn into the color buffer.  It will eventually be written into the depth buffer, but we don't need that yet so it's not included here.

The feature must be explicitly enabled by defining `EnableDepthBuffer` on the MapGrid so it shouldn't affect the classic mods (but this should be tested) or ts once the test commit is dropped.  Because actors etc don't write sensible vertex z coordinates they end up being discarded for most map scroll positions (reason explained above).  My [depthbuffer-part-two](https://github.com/pchote/OpenRA/tree/depthbuffer-part-two) branch prototypes that support, which shows off the almost-proper depth sorting that this pr is working towards.

Things to look out for / be aware of when testing:
- If the projection is wrong then holes will appear in the map, probably at the top or bottom of the screen or where the terrain is very high.
- The depth image should look consistent with the map everywhere (no sudden jumps in shade that aren't explained by the map tiles).
- I've included the test map that I uses for tweaking and debugging the depth transformations.
- Loading the "Town of Karasjok" map into the editor is a good way of getting a feeling for this.  You can enable depth rendering in the editor by starting a skirmish, enabling depth rendering, returning to the shellmap, then starting the editor.

Good luck!

### DIFF
--- a/OpenRA.Game/FieldLoader.cs
+++ b/OpenRA.Game/FieldLoader.cs
@@ -508,6 +508,26 @@ namespace OpenRA
 
 				return InvalidValueAction(value, fieldType, fieldName);
 			}
+			else if (fieldType == typeof(float3))
+			{
+				if (value != null)
+				{
+					var parts = value.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+					float x = 0;
+					float y = 0;
+					float z = 0;
+					float.TryParse(parts[0], NumberStyles.Float, NumberFormatInfo.InvariantInfo, out x);
+					float.TryParse(parts[1], NumberStyles.Float, NumberFormatInfo.InvariantInfo, out y);
+
+					// z component is optional for compatibility with older float2 definitions
+					if (parts.Length > 2)
+						float.TryParse(parts[2], NumberStyles.Float, NumberFormatInfo.InvariantInfo, out z);
+
+					return new float3(x, y, z);
+				}
+
+				return InvalidValueAction(value, fieldType, fieldName);
+			}
 			else if (fieldType == typeof(Rectangle))
 			{
 				if (value != null)
--- a/OpenRA.Game/Primitives/float3.cs
+++ b/OpenRA.Game/Primitives/float3.cs
@@ -1,0 +1,54 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Drawing;
+using System.Runtime.InteropServices;
+
+namespace OpenRA
+{
+	[SuppressMessage("StyleCop.CSharp.NamingRules", "SA1300:ElementMustBeginWithUpperCaseLetter", Justification = "Mimic a built-in type alias.")]
+	[StructLayout(LayoutKind.Sequential)]
+	public struct float3
+	{
+		public readonly float X, Y, Z;
+		public float2 XY { get { return new float2(X, Y); } }
+
+		public float3(float x, float y, float z) { X = x; Y = y; Z = z; }
+		public float3(float2 xy, float z) { X = xy.X; Y = xy.Y; Z = z; }
+
+		public static implicit operator float3(int2 src) { return new float3(src.X, src.Y, 0); }
+		public static implicit operator float3(float2 src) { return new float3(src.X, src.Y, 0); }
+
+		public static float3 operator +(float3 a, float3 b) { return new float3(a.X + b.X, a.Y + b.Y, a.Z + b.Z); }
+		public static float3 operator -(float3 a, float3 b) { return new float3(a.X - b.X, a.Y - b.Y, a.Z - b.Z); }
+		public static float3 operator -(float3 a) { return new float3(-a.X, -a.Y, -a.Z); }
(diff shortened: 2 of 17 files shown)


OpenRA.Game/FieldLoader.cs
```diff
@@ -508,6 +508,26 @@ public static object GetValue(string fieldName, Type fieldType, MiniYaml yaml, M
 
 				return InvalidValueAction(value, fieldType, fieldName);
 			}
+			else if (fieldType == typeof(float3))
+			{
+				if (value != null)
+				{
+					var parts = value.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+					float xx = 0;
```
Phrohdoh: Why can these not be `x, y, z`?

OpenRA.Game/FieldLoader.cs
```diff
@@ -508,6 +508,26 @@ public static object GetValue(string fieldName, Type fieldType, MiniYaml yaml, M
...
+					float yy = 0;
+					float zz = 0;
+					float res;
+					if (float.TryParse(parts[0].Replace("%", ""), NumberStyles.Float, NumberFormatInfo.InvariantInfo, out res))
+						xx = res * (parts[0].Contains('%') ? 0.01f : 1f);
+					if (float.TryParse(parts[1].Replace("%", ""), NumberStyles.Float, NumberFormatInfo.InvariantInfo, out res))
+						yy = res * (parts[1].Contains('%') ? 0.01f : 1f);
+					if (parts.Length > 2 && float.TryParse(parts[2].Replace("%", ""), NumberStyles.Float, NumberFormatInfo.InvariantInfo, out res))
```
Phrohdoh: Why is the `z` component not required?

Phrohdoh:
59a189f4248e9a6cca2fc91f29b90a44fe718d5b introduces `float3` but 4c57c70d1a5a021f00abc81cfc0e8e807b312043 (which comes before 59a189) makes use of them in voxel code, how is this possible?

OpenRA.Game/FieldLoader.cs
pchote: Backward compatibility with `float2`.

OpenRA.Game/FieldLoader.cs
pchote: Dunno, this is just copying what `float2` does.

pchote:
That's just github's usual commit ordering glitch.  In the actual git history 59a189f comes before 4c57c70.

OpenRA.Game/Primitives/float3.cs
```diff
@@ -0,0 +1,53 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
```
GraionDilach: Old header.

pchote:
Updated `float3` to fix the issues noted by @Phrohdoh and @GraionDilach.

pchote:
Note that I have another ~1.5 weeks of holiday that I can (partially) spend on pushing through the followup PRs.  These will significantly advance our progress towards releasable TS and RA2 previews, so it would be a shame to lose this opportunity because of lack of review interest.

reaperrr:
All actors and shroud/fog borders disappear (visually) from A River Runs Near It once you move the camera below the upper third of the map.
On Town of Karasjok, if visibility checks are NOT disabled, the whole screen goes black approximately in that upper third of the map where otherwise actors would become visible.

pchote:
I explained that in the pr description:

>  Because actors etc don't write sensible vertex z coordinates they end up being discarded for most map scroll positions (reason explained above). My depthbuffer-part-two branch prototypes that support, which shows off the almost-proper depth sorting that this pr is working towards.

Mailaender:
No regressions in `d2k` and `ra2`. :white_check_mark:

abcdefg30:

Didn't spot any regressions on a look through all mods (including `ra2` and `ts` without the testcase).
@reaperrr yeah, the effects of the testcase are... quite impressing :p
weirdRendering

@pchote I think you can remove the testcase now, although my 👍 probably doesn't even count as much as a :+0.25:.

reaperrr:
> I explained that in the pr description:

Sorry, I missed that part about scrolling causing it.
And I wasn't sure whether the black screen fell under known issues.

Anyway, other than that I didn't spot any issues, so :+1: once the test commit is removed.

pchote:
Removed test commit.

OpenRA.Game/Primitives/float3.cs
```diff
@@ -0,0 +1,53 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
```
GraionDilach: This is still outdated header. (pre specifying GPLv3 explicitly).

pchote:
Fixed.

GraionDilach:
The math seems legit to me and no noticeable performance regressions are visible even on my toaster. I guess :tada:/:+1: for the first step?

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/e6682db8d57ef1276d4d3743a0da49e1a337c3ab...211bbe61b097ead3fd0532c88219ceae1b7fdfa7)

PR #11065, opened 2016-04-07
Add a lua method for querying all actors that are currently InWorld

abcdefg30:
As requested to fight the abuse of `Map.ActorsInBox`. Closes https://github.com/OpenRA/OpenRA/issues/8549.

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs
@@ -9,6 +9,7 @@
  */
 #endregion
 
+using System.Collections;
 using System.Linq;
 using Eluant;
 using OpenRA.Scripting;
@@ -52,6 +53,20 @@ namespace OpenRA.Mods.Common.Scripting
 						return false;
 
 			return true;
+		}
+
+		[Desc("Returns the original collection filtered with the func.")]
+		public LuaTable Where(LuaValue[] collection, LuaFunction func)
+		{
+			var t = Context.CreateTable();
+
+			foreach (var c in collection)
+				using (var ret = func.Call(c))
+				using (var result = ret.FirstOrDefault())
+					if (result != null && result.ToBoolean())
+						t.Add(t.Count + 1, c);
+
+			return t;
 		}
 
 		[Desc("Returns the first n values from a collection.")]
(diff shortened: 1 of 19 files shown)


abcdefg30:
I'll updated all our maps to use this instead of `Map.ActorsInBox` tomorrow, to justify the `For Stable`.

abcdefg30:
Updaded. I can squash the `Replace ...` commits if wished.

obrakmann:
Looks good to me, :+1:

OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs
```diff
@@ -54,6 +55,20 @@ public bool All(LuaValue[] collection, LuaFunction func)
 			return true;
 		}
 
+		[Desc("Returns the original collection filtered with the func.")]
+		public LuaTable Where(LuaValue[] collection, LuaFunction func)
+		{
+			var t = Context.CreateTable();
+
+			foreach (var c in collection)
+				using (var ret = func.Call(c))
```
Phrohdoh: This call should be disposed.

```diff
diff --git a/OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs b/OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs
index 50ca3e5..1172b6a 100644
--- a/OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/UtilsGlobal.cs
@@ -37,9 +37,13 @@ public bool Any(LuaValue[] collection, LuaFunction func)
 			foreach (var c in collection)
 				using (var ret = func.Call(c))
 				using (var result = ret.FirstOrDefault())
+				{
 					if (result != null && result.ToBoolean())
 						return true;
 
+					ret.Dispose();
+				}
+
 			return false;
 		}

```

abcdefg30: Um, that doesn't look like the `Where` function, but rather like `Any`...
Anyway, I could do that here in theory, but should I? I mean we don't do this anywhere here.

Mailaender: > The using statement ensures that Dispose is called even if an exception occurs while you are calling methods on the object.

https://msdn.microsoft.com/en-us/library/yh598w02.aspx?f=255&MSPPError=-2147217396

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/211bbe61b097ead3fd0532c88219ceae1b7fdfa7...261708038bc091af20c43928628338e3b54098fd)

PR #11076, opened 2016-04-08
Fix issues with the insufficient funds warning

### LINKED ISSUE #11052: Audio feedback when queuing production
(Playtest 20160403)
Previous versions of the game played an immediate "building" or "training" sound when you queued multiple units or buildings for production. 

Example: Clicking on rifle infantry and immediatly on the rocket infantry (queuing it) would play "training" sound twice. This gave an audio confirmation of your action.

Now the "training" sound of the queued unit plays only when the unit actually starts producing (so much later).

While this could be intentional it seems less usefull to me.

### LINKED ISSUE #11073: TD SAM site and advanced guard tower don't play building notifications
See title

### LINKED ISSUE #11075: TD SAM Site and Advanced Guard Tower delay before building cost applies
When building these, the building process starts but money isn't taken from the player until construction is about 1/8th done. Production cannot be put on hold during this time, instead the building is cancelled, which made me think this isn't working as intended.

Possibly related to [#11073](https://github.com/OpenRA/OpenRA/issues/11073)?

obrakmann:
This fixes several issues with the insufficient funds warning, as well as an unrelated polish issue with the production code.
1. Fixes several warnings being played at once by moving the logic that ultimately causes the sound to be played from from the production queues to `PlayerResources`.
2. Fixes warnings being played shortly after one another by adding a customizable delay.
3. Adds notifications being triggered by repairs.
4. Fixes #11052.
5. Fixes #11073.

Another difference between this implementation and the old one is that the new one will keep repeating the warning (at the interval of the delay) until funds are available again.


OpenRA.Game/Traits/Player/PlayerResources.cs
```diff
@@ -111,9 +121,19 @@ public void GiveCash(int num)
...
+		public bool TakeCash(int num, bool notifyLowFunds = false)
 		{
-			if (Cash + Resources < num) return false;
+			if (Cash + Resources < num)
+			{
+				if (notifyLowFunds && !string.IsNullOrEmpty(info.InsufficientFundsNotification) &&
+					owner.World.WorldTick - lastNotificationTick >= info.InsufficientFundsNotificationDelay)
+				{
+					lastNotificationTick = owner.World.WorldTick;
+					Game.RunAfterTick(() =>	Game.Sound.PlayNotification(owner.World.Map.Rules, owner, "Speech", info.InsufficientFundsNotification, owner.Faction.InternalName));
```
abcdefg30: Tab after `=>`.

abcdefg30:

The last comment adds really nice polish, too!

OpenRA.Game/Traits/Player/PlayerResources.cs
```diff
@@ -111,9 +121,19 @@ public void GiveCash(int num)
...
+		public bool TakeCash(int num, bool notifyLowFunds = false)
 		{
-			if (Cash + Resources < num) return false;
+			if (Cash + Resources < num)
+			{
+				if (notifyLowFunds && !string.IsNullOrEmpty(info.InsufficientFundsNotification) &&
+					owner.World.WorldTick - lastNotificationTick >= info.InsufficientFundsNotificationDelay)
+				{
+					lastNotificationTick = owner.World.WorldTick;
+					Game.RunAfterTick(() => Game.Sound.PlayNotification(owner.World.Map.Rules, owner, "Speech", info.InsufficientFundsNotification, owner.Faction.InternalName));
```
pchote: What's the purpose of the RunAfterTick here?

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -433,7 +436,14 @@ public void Tick(PlayerResources pr)
 			{
 				var time = Queue.GetBuildTime(Item);
 				if (time > 0)
+				{
 					RemainingTime = TotalTime = time;
+					framesToSkip = TotalTime - TotalCost;
```
pchote: Can you add a comment similar to the commit description explaining what's going on here?  At first glance it looks like a bug to be relating TotalTime with TotalCost (and tbh i'm still not convinced whether this will work for all custom build time / queue multiplier values).

OpenRA.Game/Traits/Player/PlayerResources.cs
obrakmann: No clue, I was wondering as well. I just took it over from the original code. I'll remove it.

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
obrakmann: How did you manage to comment on an outdated diff three hours ago when I last pushed more than two days ago?

Anyway, yeah, I can add a comment.

And yes, this will break again (ie. the original bug will reappear) if the difference between `TotalCost` and `TotalTime` gets large (if `TotalTime` ends up being more than twice the value of `TotalCost`). With the way our queues are set up atm, that will happen with a `CustomBuildTimeValue` that is five times that of `Valued.Cost`.

The problem is that integer division gets very imprecise when the two values are close to each other, so the result ends up being 1, which means that every frame would be skipped until they are all accounted for. Only then the regular cash ticks would occur, just like in the original bug.

Not sure how to do it any other way, though. We can't use proper floating-point division, since it's simulation code, and it also wouldn't make sense, as half a tick doesn't quite work ;)

pchote: Another github bug I guess...

obrakmann:
Dropped the last commit as discussed in IRC.

pchote:
:+1: code looks fine, and the improved behaviour works well (as far as I can tell) ingame.

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/a52c7acc76a14c3e7b51cb22374bce625e71bc2e and cherry-picked to prep-1604.

PR #11086, opened 2016-04-10
Add Marn's Shellmap "Fields of Green" to TS

phrohdoh:
screen shot 2016-04-09 at 10 02 06 pm


mods/ts/maps/fields-of-green/map.lua
```diff
@@ -0,0 +1,97 @@
...
+WorldLoaded = function()
+    nod = Player.GetPlayer("Nod")
+    gdi = Player.GetPlayer("GDI")
+	
+	SetupFactories()
+	Utils.Do(ProducedUnitTypes, ProduceUnits)
+	SendNodInfantry(NodW.Location, NForce, NForceInterval)
+	Trigger.AfterDelay(DateTime.Seconds(50), SendNodVehicles, South.Location, VNForce, VNForceInterval)
+	SendGDIInfantry(GDIW.Location, GForce, GForceInterval)
+	Trigger.AfterDelay(DateTime.Seconds(70), SendGDIVehicles, North.Location, VGForce, VGForceInterval)
```
abcdefg30: Um, this line and the three above don't seem to be correct at all.
`SendNodInfantry` and `SendGDIInfantry` don't accept parameters, they're already hardcoded anyway.
Also, it's `Trigger.AfterDelay(delay, function)`. You're calling it with 5 arguments instead?
Those are anyway unnecessary again as they're hardcoded in the respective functions.

I think it would be correct that way:
```
SendNodInfantry()
Trigger.AfterDelay(DateTime.Seconds(50), SendNodVehicles)
SendGDIInfantry()
Trigger.AfterDelay(DateTime.Seconds(70), SendGDIVehicles)
```

mods/ts/maps/fields-of-green/map.yaml
```diff
@@ -0,0 +1,1459 @@
...
+	HARV:
+		-Targetable:
+	GALITE:
+		-Targetable:
+	APC:
+		Cargo:
+			InitialUnits: e1, e1, e2, e2, medic
+	GACTWR:
+		Health:
+			HP: 99999999
```
abcdefg30: Couldn't we just make those buildings invulnerable using damage modifiers (like the ra shellmap does)?

mods/ts/maps/fields-of-green/map.yaml
```diff
@@ -0,0 +1,1459 @@
...
+Rules:
+	Player:
+		-ConquestVictoryConditions:
+	World:
+		-CrateSpawner:
+		-StartGameNotification:
+		-SpawnMPUnits:
+		-MPStartLocations:
+		LuaScript:
+			Scripts: map.lua
```
abcdefg30: I think you might want to rename this file and add it to `OpenRA.sln` like we do with the other lua files.

mods/ts/maps/fields-of-green/map.yaml
```diff
@@ -0,0 +1,1459 @@
...
+		ResourceType@Tiberium:
+			ResourceType: 1
+			ValuePerUnit: 0
+		MusicPlaylist:
+			BackgroundMusic: intro
+		GlobalLightingPaletteEffect:
+			Red: 1.0
+			Green: 1.0
```
abcdefg30: Here and the line above: Aren't those already the default values?

pchote:
Can you please remove the old shellmap too?

mods/ts/maps/fields-of-green/map.yaml
```diff
@@ -0,0 +1,1459 @@
...
+	Actor1: tibtre03
+		Location: 66,-1
+		Owner: Neutral
+	Actor2: tibtre02
+		Location: 60,11
+		Owner: Neutral
+	Actor3: tibtre01
+		Location: 58,16
+		Owner: Neutral
+	Actor428: mpspawn
```
pchote: These don't seem to be used/needed?

mods/ts/maps/fields-of-green/map.yaml
```diff
@@ -0,0 +1,1459 @@
+MapFormat: 11
+
+RequiresMod: ts
+
+Title: Fields of Green
```
pchote: Can we call this "Shellmap" or similar so we can find it in the map chooser?

Phrohdoh:
Comments addressed.

mods/ts/maps/fields-of-green/map.yaml
Phrohdoh: The `Category` already covers this.

pchote: Ok, fair enough.

reaperrr:
:+1:

mods/ts/maps/fields-of-green/fields-of-green.lua
```diff
@@ -0,0 +1,106 @@
...
+end
+
+SetupFactories = function()
+    Utils.Do(ProducedUnitTypes, function(pair)
+        Trigger.OnProduction(pair[1], function(_, a) BindActorTriggers(a) end)
+    end)
+end
+
+SetupInvulnerability = function()
+   for i, actor in ipairs(Map.NamedActors) do
```
abcdefg30: It doesn't look like you're using `i` here. So please use `Utils.Do`.

Phrohdoh: Good catch, will do.

abcdefg30:

mods/ts/maps/fields-of-green/fields-of-green.lua
```diff
@@ -0,0 +1,106 @@
...
+end
+
+SetupFactories = function()
+    Utils.Do(ProducedUnitTypes, function(pair)
+        Trigger.OnProduction(pair[1], function(_, a) BindActorTriggers(a) end)
+    end)
+end
+
+SetupInvulnerability = function()
+   Utils.Do(Map.NamedActors, function()
```
Mailaender: You need to add `function(actor)` here. 

```
/home/matthias/Entwicklung/OpenRA//lua/sandbox.lua:150: [string "BindingSupport.lua"]:30: Uncaught CLR exception at Lua->CLR boundary: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> Eluant.LuaException: [string "fields-of-green.lua"]:41: attempt to index global 'actor' (a nil value)
```

Phrohdoh:
Added the missing param to the anonymous function in `SetupInvulnerability`.

PR #11089, opened 2016-04-10
Fix TS debris angles.

GraionDilach:
I r lazy. I'm aware the lint rules ain't written yet but I've noticed this bug pops up too many times now to be ignored.


Phrohdoh:
👍

obrakmann:
:+1:

PR #11090, opened 2016-04-10
Added Tiberian Dawn mission GDI 07

Mailaender:
See http://cnc.wikia.com/wiki/Finish_The_Nod_Base for reference. I left the original atk\* trigger names intact, so you can compare easier with the original mission configuration.


HenrytheSlav:
Got fatal error regarding lua script during first minute.
http://pastebin.com/7qrHiKwk

Mailaender:
Fixed.

HenrytheSlav:
I would prefer if there was at least basic production scripted (On 2000 credits, produce 5 inf + 2 rockets) like we do in my for example Soviet5. This way the map is kinda bland without any attacks later. Also it gives at least illusion of AI actually defending.
I would also like to adress the MCV landing on same tile as medium tank. In RA, there is "repulsion" of units from transport, in cnc it seems isn't (bug/issue?) but anyway, this way if you don't move your tank at start MCV ends up in same tile.

As for functionality there were no more issues encountered, however I keep my +1 until these issues are adressed/explained. :)

Mailaender:
Addressed the MCV issue. Will wait for https://github.com/OpenRA/OpenRA/issues/11187 instead of copy pasting the soviet 5 script around.

HenrytheSlav:
Please, this #11187 will take a lot of time to be put up and it will still need to include reworks of all missions, so feel free to merge in current status after +2. NOD missions are being merged too.

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/2ef51fda08cb582b4ad424763177e1348858068b...45174300564572f8782008ef4c83b6cc452e53fb)

PR #11091, opened 2016-04-10
Add depth buffer support for sprites.

pchote:
The next step towards #7520.

This enables the depth buffer properly for mods that define `EnableDepthBuffer` in their mod.yaml, and writes sensible data into the sprite z coordinates.  This fixes most of the "scroll the map and X disappears" issues from #11058 (but not for voxels or color effects), and gives the desired clipping behaviour for sprite based actors behind cliffs.

Large sprite actors (buildings) are partly clipped below the terrain: this is expected, and will be fixed in a future PR that adds support for per-sprite depth maps.  The clip level is not actually ground level because of the way the terrain depth is defined.  That will be fixed in the next PR.  It also disables the old per-sprite depth preview (with `EnableDepthBuffer` disabled), because that is no longer useful and it would be a pain to maintain support for it.

Github has also buggered up the commit order in their preview, as usual.  The "TEST COMMIT" is the head commit, and should be removed before merging.

### DIFF
--- a/OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
+++ b/OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
@@ -213,7 +213,7 @@ namespace OpenRA.Platforms.Default
 			VerifyThreadAffinity();
 			OpenGL.glClearColor(0, 0, 0, 1);
 			OpenGL.CheckGLError();
-			OpenGL.glClear(OpenGL.GL_COLOR_BUFFER_BIT);
+			OpenGL.glClear(OpenGL.GL_COLOR_BUFFER_BIT | OpenGL.GL_DEPTH_BUFFER_BIT);
 			OpenGL.CheckGLError();
 		}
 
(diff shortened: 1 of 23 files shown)


obrakmann:
:+1:

OpenRA.Platforms.Default/Sdl2GraphicsDevice.cs
```diff
@@ -224,13 +224,17 @@ public void EnableDepthBuffer()
...
+			OpenGL.glDepthFunc(OpenGL.GL_LEQUAL);
+			OpenGL.CheckGLError();
 		}
 
 		public void DisableDepthBuffer()
 		{
 			VerifyThreadAffinity();
 			OpenGL.glDisable(OpenGL.GL_DEPTH_TEST);
 			OpenGL.CheckGLError();
+			OpenGL.glDepthFunc(OpenGL.GL_ALWAYS);
```
chrisforbes: Why?

pchote: Detritus from early experiments... removed.

reaperrr:
:+1: 

Test commit can be removed.

pchote:
Removed test commit.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/74c1e89890220a2c4c81b711eb27fad5a24ffe58...4a7400dc0971d6c272286d320ad3761201b46d16)

Issue #10892, opened 2016-03-07
WithDecorations specified only for Neutral and Enemy stances are visible to spectators


GraionDilach:
This is a regression, but uncertain if caused by fragile alliances removal or the WithDecoration refactor.

abcdefg30:
I guess you mean when the spectator uses the view of a player?

GraionDilach:
No.

This is from the default NoShroud view.

Example case in below. 

```
WithRankDecoration:
		Image: herorank
		Sequence: rank
		Palette: ra
		ReferencePoint: Bottom, Right
		Offset: 2, 2
		UpgradeTypes: rank
		ZOffset: 256
		UpgradeMinEnabledLevel: 1
		UpgradeMaxAcceptedLevel: 3
	WithDecoration@defrank:
		Image: herorank
		Sequence: star
		ReferencePoint: Bottom, Right
		ZOffset: 256
		Offset: 2, 2
		Palette: ra
		UpgradeTypes: rank
		UpgradeMaxEnabledLevel: 0
		UpgradeMaxAcceptedLevel: 3
	WithDecoration@startoenemy:
		Image: herorank
		Sequence: star
		ReferencePoint: Bottom, Right
		Offset: 2, 2
		Stances: Neutral, Enemy
		ZOffset: 256
		Palette: ra
```

GraionDilach:
Previously when a hero gained a level, the NoShroud view only seen the the proper rank, but now both defrank and startoenemy are visible which is a mess.

pchote:
The convention that has been applied everywhere so far is that *everything* is visible to players with `RenderPlayer == null`.  That quite obviously breaks down in cases like this.

pchote:
I expect the proper fix will be to introduce a `Stance.Observer` value.

PR #11093, opened 2016-04-10
Add an option to AnnounceOnBuild/AnnounceOnKill to only announce their activity to it's owner.

GraionDilach:
Requested by @ABrandau and SchismNavigator.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Sound/AnnounceOnBuild.cs
+++ b/OpenRA.Mods.Common/Traits/Sound/AnnounceOnBuild.cs
@@ -19,6 +19,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Voice to use when built/trained.")]
 		[VoiceReference] public readonly string Voice = "Build";
 
+		[Desc("Should the voice be played for the owner alone?")]
+		public readonly bool OnlyToOwner = false;
+
 		public object Create(ActorInitializer init) { return new AnnounceOnBuild(init.Self, this); }
 	}
 
@@ -33,6 +36,9 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void BuildingComplete(Actor self)
 		{
+			if (info.OnlyToOwner && self.Owner != self.World.RenderPlayer)
+				return;
+
 			self.PlayVoice(info.Voice);
 		}
 	}
--- a/OpenRA.Mods.Common/Traits/Sound/AnnounceOnKill.cs
+++ b/OpenRA.Mods.Common/Traits/Sound/AnnounceOnKill.cs
@@ -22,6 +22,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Voice to use when killing something.")]
 		[VoiceReference] public readonly string Voice = "Kill";
 
+		[Desc("Should the voice be played for the owner alone?")]
+		public readonly bool OnlyToOwner = false;
+
 		public object Create(ActorInitializer init) { return new AnnounceOnKill(init.Self, this); }
 	}
 
@@ -42,6 +45,9 @@ namespace OpenRA.Mods.Common.Traits
 			// Don't notify suicides
 			if (e.DamageState == DamageState.Dead && damaged != e.Attacker)
 			{
+				if (info.OnlyToOwner && self.Owner != self.World.RenderPlayer)
+					return;
+
 				if (self.World.WorldTick - lastAnnounce > info.Interval * 25)
 					self.PlayVoice(info.Voice);
 


abcdefg30:
Um, I know you'll probably hate me for that scope creep, but couldn't we get some kind of stances flag instead? Like have an option for `only the owner`, `allies`(, `neutrals`) and `enemies`.

GraionDilach:
Yes. I will. I haven't started looking into the stances setup and usecases as part of the #10892 fix either. But if that's what everyone wants, then I'll look into it (honestly, I wanted to mod today and only did this as a 5 minute break).

OpenRA.Mods.Common/Traits/Sound/AnnounceOnBuild.cs
```diff
@@ -19,6 +19,9 @@ public class AnnounceOnBuildInfo : ITraitInfo
 		[Desc("Voice to use when built/trained.")]
 		[VoiceReference] public readonly string Voice = "Build";
 
+		[Desc("Should the voice be played for the owner alone?")]
+		public readonly bool AnnounceOnlyToOwner = false;
```
Mailaender: `OnlyToOwner` instead maybe? It is otherwise a bit redundant.

GraionDilach: Considering we have some English-pickers in the form of @Phrohdoh I thought it's better to have it redundant than nonEnglish.

Mailaender: I can't follow as I read this as `AnnounceOnBuild.AnnounceOnlyToOwner`.

GraionDilach: Then do not follow.

Mailaender:
:+1:

pchote:
Please make this use stances. It is silly to rewrite this later with upgrade rules when we can do it properly the first time around.

GraionDilach:
Due to some circumstances, I definitely won't find time for that in the next week. Closing this then.

GraionDilach:
It is also silly to scopecreep a supposed-to-be-5-minute-polish to ___replace every usage of Stance.Ally in the code with Stance.Owner and Stance.StrictAlly___ but after the CompletelyMissingThePoint PR, I got pissed off enough much that I actually started to work on that instead of some actual RL stuff I should care about.

Phrohdoh:
I don't think asking a PR author to implement something in a consistent manner is scopecreep.

SchismNavigator:
The whole point of this change is to enable the player to hear when their units are being built, and when they are fighting/winning.

I am not against it being expanded upon. But can we just get this basic functionality in first? It'll go a long way to adding more personality to units and personalising each faction.

obrakmann:
@SchismNavigator: I don't see why those voice-overs should not be heard by the allies as well. The personality would still be limited to that faction. (Yeah, I know, that's what Generals & co. did, but idgaf.)

GraionDilach:
Generals, Warcraft, Starcraft and co. uses the creation voiceovers as a replacement of the _Unit completed_ EVA message, to give a better oversight to the player which unit is done (since all these have individual queues and thereby one generic response is too vague). If a modder decided to use `AnnounceOnBuild` for this purpose, then allies shouldn't hear each other's units.

I don't understand `AnnounceOnKill` myself either, but I just imagine the same logic was applied to it ˙by them as well.

Mailaender:
Being stubborn and blocking the PR caused an unnecessary fork: https://github.com/GraionDilach/OpenRA.Mods.AS/commit/0f4b0fe3c6ce8e8201762eee664e33251c2e35ee

GraionDilach:
That's just a workaround. I'm ongoing with expanding the Stances flag, just a bit slow with updating and looking up all usecases and some RL issues which probably won't resolve themselves until early June as it stands atm.

pchote:
> Being stubborn and blocking the PR caused an unnecessary fork: GraionDilach/OpenRA.Mods.AS@0f4b0fe

Downstream mods implementing their own custom logic in their own dll is a *good thing*.

Mailaender:
It isn't own logic nor a good thing. That is copy paste of existing code with minor additions. Requiring this is an indicator of total failure.

Mailaender:
Back to the topic. Do we really want a stance `Owner`?

abcdefg30:
> Back to the topic. Do we really want a stance `Owner`?

Quite the late response... I'd say yes. ^^
Also: `[01:19:13] <penev> I'd say "yes" to that question`

pchote:
I also agree that `Stance.Ally` is far to general to cover all our use cases.  Separating `Ally` and `Owner` is a good start, but may not be enough (or maybe just not the right name?) if we want to replace the `AffectsParent` flags used in warheads and external conditions.

GraionDilach:
I actually had a working code back during those days where I kept Stance.Ally as-is and added Stance.StrictAlly and Stance.Owner (due to being lazy to write upgrade rules) and I only had regressions with shroud on first sight since shroud wasn't based on stances back those days.

AffectsParent should be kept unrelated to the stance setups, if the warhead is used to generate a wave explosion (missile with a rangelimit of 1 in bleed, InstantExplode projectile in AS) it would be expected that the firer wouldn't damage itself, but would still damage it's allied units owned by it's owner even - see the Firestorm Mobile EMP as an example, where friendly MEMPS do EMP each other down, but not themselves)

pchote:
Ok

GraionDilach:
I was wrong - Firestorm didn't have this logic (it used TypeImmune which means MEMPs are immune to themselves unrelated to owners), YR had it under the DamageSelf tag (used for the Nuke Silo, to prevent own silo damage but still damage enemy ones).

PR #11098, opened 2016-04-11
Replace SheetType.DualIndexed with secondary texture coordinates on Vertex.

pchote:
In the current renderer we have two cases where we care about _two_ indices per pixel:
- Voxels (color + normal)
- Terrain (color + depth)

Both of these have a one-to-one relationship between the two components: every pixel has its own unique normal/depth value.  This was supported with a kind-of-hacky-but-works setup where we reserved every second channel in the sprite sheet for this extra data, and assumed/enforced that the secondary data for a given pixel was at the same x/y coord and channel + 1.

Supporting depth data for regular sprites is more complicated: there is a single depth map (buildngz.shp) that is shared for all building sprites.  Keeping the old implementation would be extremely wasteful of texture space because we would need to duplicate the same depth sprite for every regular sprite.

This PR reworks the "secondary texture data" concept by adding two extra components to the vertex format, which are interpreted as a secondary texture coordinate.  The secondary channel index is merged in to the channel component and extracted in the vertex shader.  This introduces a little waste for the classic mods (they will always be pushing two unused floats to the GPU for each vertex), but this dramatically simplifies life for the gen2 mods who can now share the same depth sprite for all sprites on the same sheet.

Voxels and terrain are reworked using the new code, and the old DualIndexed code is dropped.

### DIFF
--- a/OpenRA.Game/Graphics/SheetBuilder.cs
+++ b/OpenRA.Game/Graphics/SheetBuilder.cs
@@ -25,8 +25,7 @@ namespace OpenRA.Graphics
 	public enum SheetType
 	{
 		Indexed = 1,
-		DualIndexed = 2,
-		BGRA = 4,
+		BGRA = 2,
 	}
 
 	public sealed class SheetBuilder : IDisposable
--- a/OpenRA.Game/Graphics/RgbaColorRenderer.cs
+++ b/OpenRA.Game/Graphics/RgbaColorRenderer.cs
@@ -69,12 +69,12 @@ namespace OpenRA.Graphics
 			var eb = endColor.B / 255.0f;
 			var ea = endColor.A / 255.0f;
 
-			vertices[nv++] = new Vertex(start - corner + Offset, sr, sg, sb, sa);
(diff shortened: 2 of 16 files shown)


reaperrr:
Didn't spot any regressions :+1:

reaperrr:
Needs rebase.

pchote:
Rebased.

Mailaender:
I can confirm that this doesn't add a noticeable render tick impact even on integrated Intel HD 4000 on a sub-notebook. :white_check_mark:

OpenRA.Game/Graphics/SheetBuilder.cs
```diff
@@ -25,8 +25,7 @@ public SheetOverflowException(string message)
 	public enum SheetType
 	{
 		Indexed = 1,
-		DualIndexed = 2,
-		BGRA = 4,
+		BGRA = 2,
```
Mailaender: Now I see. This was subtle. https://github.com/OpenRA/OpenRA/pull/12696

PR #11102, opened 2016-04-11
Add stance awareness to AnnounceOnBuild/Kill

obrakmann:
Resurrected #11093.

cc: @GraionDilach.


GraionDilach:
That doesn't solve anything at all itself. The original request was to make them work like in Generals and beyond C&Cs, to have an option for these to be only heard by their owner __alone__ which you guys decided should be prerequisited with the whole Stance enum being expanded beforehand.

PR #11108, opened 2016-04-13
Emp cannon support power

### LINKED ISSUE #7349: EMP Cannon should act as SupportPower (EMPSpecial)
We introduced the EMP logic for TS to get this work but we should transform them to an EMPSpecial as like in vanilla TS.

In the Moment the EMP is a game killer because:
- When i build of a Wall of EMP Cannons the Game will lag
- The EMP Cannon is firering every second
- Buildlimit should be limited to 1

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7844571-emp-cannon-should-act-as-superweapon-empspecial?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

teees:
Adds a new `AttackOrderPower` which orders the unit to attack the selected location and stops after one shot.

Fixes #7349

https://streamable.com/ekh8
https://streamable.com/2sv6
https://streamable.com/qcyd

With `AllowMultiple: true`: https://streamable.com/iex9
##### Possible issues mentioned in comments:
- No beacon or camera support
- No burst weapon support (cannon stops after the first shot) - Solved with #11158
- `self.World.IssueOrder(new Order("Stop", self, false));` has to be removed (!)
##### Outstanding polishing issues not covered by this pr:
- ~~Targeter works only for one EMP cannon and doesn't check for other cannons in the area~~
- ~~Reset emp range to original value~~
- ~~Set proper Armament local offset and yaw~~
- No charge animation
- Charging end sound is the same for all factions
- Select target sound is the same for all factions

### DIFF
--- a/OpenRA.Mods.TS/Traits/SupportPowers/AttackOrderPower.cs
+++ b/OpenRA.Mods.TS/Traits/SupportPowers/AttackOrderPower.cs
@@ -1,0 +1,148 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.TS.Traits
+{
+	class AttackOrderPowerInfo : SupportPowerInfo
+	{
+		public override object Create(ActorInitializer init) { return new AttackOrderPower(init.Self, this); }
+	}
+
+	class AttackOrderPower : SupportPower, INotifyAddedToWorld, INotifyAttack
+	{
+		readonly AttackOrderPowerInfo info;
+		AttackBase attack;
+
+		public AttackOrderPower(Actor self, AttackOrderPowerInfo info)
+			: base(self, info)
+		{
+			this.info = info;
+		}
+
+		public override void SelectTarget(Actor self, string order, SupportPowerManager manager)
+		{
+			Game.Sound.PlayToPlayer(manager.Self.Owner, Info.SelectTargetSound);
+			self.World.OrderGenerator = new SelectAttackPowerTarget(self, order, manager, info.Cursor, MouseButton.Left, attack);
+		}
+
+		public override void Activate(Actor self, Order order, SupportPowerManager manager)
+		{
+			base.Activate(self, order, manager);
+			attack.AttackTarget(Target.FromCell(self.World, order.TargetLocation), false, false, true);
+		}
+
+		void INotifyAddedToWorld.AddedToWorld(Actor self)
+		{
+			attack = self.Trait<AttackBase>();
+		}
+
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			self.World.IssueOrder(new Order("Stop", self, false));
+		}
+	}
+
+	public class SelectAttackPowerTarget : IOrderGenerator
+	{
+		readonly SupportPowerManager manager;
+		readonly SupportPowerInstance instance;
+		readonly string order;
+		readonly string cursor;
+		readonly string cursorBlocked;
+		readonly MouseButton expectedButton;
+		readonly AttackBase attack;
+
+		public SelectAttackPowerTarget(Actor self, string order, SupportPowerManager manager, string cursor, MouseButton button, AttackBase attack)
+		{
+			// Clear selection if using Left-Click Orders
+			if (Game.Settings.Game.UseClassicMouseStyle)
+				manager.Self.World.Selection.Clear();
+
+			instance = manager.GetPowersForActor(self).FirstOrDefault();
+			this.manager = manager;
+			this.order = order;
+			this.cursor = cursor;
+			expectedButton = button;
+			this.attack = attack;
+			cursorBlocked = cursor + "-blocked";
+		}
+
+		Actor GetFiringActor(World world, CPos cell)
+		{
+			var pos = world.Map.CenterOfCell(cell);
+			var range = attack.GetMaximumRange().LengthSquared;
+
+			return instance.Instances.Where(i => !i.Self.IsDisabled()).MinByOrDefault(a => (a.Self.CenterPosition - pos).HorizontalLengthSquared).Self;
+		}
+
+		bool IsValidTarget(World world, CPos cell)
+		{
+			var pos = world.Map.CenterOfCell(cell);
+			var range = attack.GetMaximumRange().LengthSquared;
+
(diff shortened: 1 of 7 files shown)


mods/ts/rules/shared-support.yaml
```diff
@@ -9,6 +9,7 @@ NAPULS:
 		Queue: Defense
 		BuildPaletteOrder: 90
 		Prerequisites: radar
+		BuildLimit: 1
```
MustaphaTR: I'm not sure if this is like that in TS. Did u added that because it doesn't work yet.

Original ones only gives range support doesn't give new sw.

teees: Partly because it doesn't work with more than one emp cannon, but also because that's what was requested in #7349

MustaphaTR: I disagree original behaviour is much better. As i said more EMP Cannon doesn't mean more SW. As EMP Cannon has a range limit, others just provides ability to attack other places.

OpenRA.Mods.TS/Traits/SupportPowers/AttackOrderPower.cs
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		void INotifyAddedToWorld.AddedToWorld(Actor self)
+		{
+			attack = self.Trait<AttackBase>();
+		}
+
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			self.World.IssueOrder(new Order("Stop", self, false));
```
GraionDilach: Won't this break if the weapon has `Burst` or `ChargeDelay`? In TS, the EMP had the latter.

mods/ts/rules/shared-support.yaml
GraionDilach: This behaviour is fine. The shared-countdown-to-a-single-firing was stupid.

OpenRA.Mods.TS/Traits/SupportPowers/AttackOrderPower.cs
GraionDilach: Ah, I can see you shifted the Delay to the SW, but `Burst should be supported here IMO.

GraionDilach:
On a second thought, gameplay terms this looks worse than my try at #9827, because mine actually shown the player the armament ranges when activating and had burst support. You also ignored/missed my sources of frustrations, namely camera and beacon support.

mods/ts/rules/shared-support.yaml
teees: https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/SupportPowers/SupportPowerManager.cs#L230

It chooses one of those support powers to get the order targeter. I'd have to search for all other emp cannons somewhere. Which I haven't done in this pr (yet). Which means that you can't fire at targets in range of any of your emp cannons, but of one of those.

teees:
I haven't implemented any other support powers yet, so this is something new for me. I didn't even think about adding any camera or beacon support yet, but I'll consider it once I've time for that.

Burst support wasn't considered either, since the ts emp cannon doesn't need it.

GraionDilach:
The `INotifyFiredSalvo` interface would take care of both charge and burst. TBH, I'm fine with lacking camera/beacons atm, but rendering the armament ranges during activation and the interface addition are something I'd strongly request there, because these should be relatively easy compared to beacon/camera which still makes me unghhh, due to missiles cannot be estimated properly. I do favor your method of implementation over mine, but I will still request to reach the same level of polish where I've got with it. :P

pchote:
I agree with some of graion's points: it would make the most sense to set this up now with support for AllowMultiple: true behavior as independent powers restricted to their own range, and AllowMultiple: false taking the closest available tower to the target point.

pchote:
Moving the support power charging to the armament would make for ugly logic, so i don't support that idea.

GraionDilach:
Uhm, I don't support that either. What I wanted to do with that `INotifyFiredSalvo` to use it for stopping the weapon instead of this ugly `IsAttacking` hack it has at the moment, which practically breaks immediately if the weapon has a `FireDelay`.

teees:
`IAttacking` is used all over the place, it's not more hacky than e.g. `AttackCharge` - which also doesn't support any salvos

teees:
So while this is useful it should imo be added in another pr, adding support for other activities as well (?)

teees:
https://streamable.com/2sv6

Added range circles, added support for multiple emp cannons

Mailaender:
Works as promised. :white_check_mark: Some easy to fix polish is the missing description/name in the support power tooltip and the firing offset.

OpenRA.Mods.TS/Traits/SupportPowers/AttackOrderPower.cs
```diff
@@ -0,0 +1,151 @@
...
+
+		public override void SelectTarget(Actor self, string order, SupportPowerManager manager)
+		{
+			Game.Sound.PlayToPlayer(manager.Self.Owner, Info.SelectTargetSound);
+			self.World.OrderGenerator = new SelectAttackPowerTarget(self, order, manager, info.Cursor, MouseButton.Left, attack);
+		}
+
+		public override void Charged(Actor self, string key)
+		{
+		}
```
abcdefg30: I think we want `public override void Charged(Actor self, string key) { }` in one line here.

OpenRA.Mods.TS/Traits/SupportPowers/AttackOrderPower.cs
```diff
@@ -0,0 +1,151 @@
...
+			expectedButton = button;
+			this.attack = attack;
+			cursorBlocked = cursor + "-blocked";
+		}
+
+		public IEnumerable<Order> Order(World world, CPos cell, int2 worldPixel, MouseInput mi)
+		{
+			world.CancelInputMode();
+			if (mi.Button == expectedButton && IsValidTarget(world, cell))
+				yield return new Order(order, manager.Self, false) {
```
abcdefg30: Please put the brace on a new line.

teees:
Changed range to vanilla tib sun range, added charging end sound, added local offset and yaw:
https://streamable.com/qcyd

OpenRA.Mods.TS/Traits/SupportPowers/AttackOrderPower.cs
```diff
@@ -0,0 +1,148 @@
...
+			if (mi.Button == expectedButton && IsValidTarget(world, cell))
+				yield return new Order(order, manager.Self, false)
+				{
+					TargetActor = GetFiringActor(world, cell),
+					TargetLocation = cell,
+					SuppressVisualFeedback = true
+				};
+		}
+
+		public virtual void Tick(World world)
```
GraionDilach: Why is this virtual?

GraionDilach:
And I repeat again: The interface I suggested would resolve the issues related to charging.

teees:
I created another pr to add that interface so it can be used in other mods and traits. Once merged I'll create a followup pr to use it in the emp cannon (or add it if it get's merged first)

OpenRA.Mods.TS/Traits/SupportPowers/AttackOrderPower.cs
pchote: This will interrupt any other activities that the actor might be doing (e.g. moving).  Can you please make this interact directly with the trait?

abcdefg30:
Superseded by #11445.

PR #11111, opened 2016-04-13
Remove custom TD & D2k engineer selection priority

### LINKED ISSUE #11067: Remove lower selection priority for engineers in TD
I said on one of the original post that making a lower selection priority for engineers in TD would be counter-productive since they are used differently to everything in RA except the spy.
Could you please remove this from TD

reaperrr:
Closes #11067.

### DIFF
--- a/mods/cnc/rules/infantry.yaml
+++ b/mods/cnc/rules/infantry.yaml
@@ -148,8 +148,6 @@ E6:
 		CaptureTypes: building, husk
 	-AutoTarget:
 	-GainsExperience:
-	Selectable:
-		Priority: 5
 
 RMBO:
 	Inherits: ^Soldier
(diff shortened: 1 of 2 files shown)


obrakmann:
TBH, I'd do it for all the mods. It would suck if it worked one way in cnc and another way in the other mods. But that's just my opinion, maybe the players prefer it in RA.

pchote:
I wasn't a fan of the original change, so I'd also support reverting the remaining parts of #10519.

pchote:
I guess the RA engineers could be argued either way (since they work differently to other units), but could we revert the d2k engineers here for consistency?

reaperrr:
Updated, reverted d2k engineers as well.

Mailaender:
The back and forth is a bit ridiculous, but anyway this should  be safe.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/04d270cec2f5df7524a83654568b801b0153596f...b67f0e84ec192194534fa9a746446a70f2d4ddbf)

PR #11112, opened 2016-04-13
Add facings support and other controls to LeavesTrails

### LINKED ISSUE #3188: Tire tracks as in Dune II [$10]
https://github.com/OpenRA/OpenRA/issues/7974 but with support for facings.

<bountysource-plugin>
---
There is a **[$10 open bounty](https://www.bountysource.com/issues/293155-tire-tracks-as-in-dune-ii?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Closes #3188.

### DIFF
--- a/OpenRA.Game/Effects/SpriteEffect.cs
+++ b/OpenRA.Game/Effects/SpriteEffect.cs
@@ -23,7 +23,7 @@ namespace OpenRA.Effects
 		readonly bool visibleThroughFog;
 		readonly bool scaleSizeWithZoom;
 
-		public SpriteEffect(WPos pos, World world, string image, string sequence, string palette, int facing = 0, bool visibleThroughFog = false, bool scaleSizeWithZoom = false)
+		public SpriteEffect(WPos pos, World world, string image, string sequence, string palette, bool visibleThroughFog = false, bool scaleSizeWithZoom = false, int facing = 0)
 		{
 			this.world = world;
 			this.pos = pos;
--- a/OpenRA.Game/Effects/SpriteEffect.cs
+++ b/OpenRA.Game/Effects/SpriteEffect.cs
@@ -23,14 +23,14 @@ namespace OpenRA.Effects
 		readonly bool visibleThroughFog;
 		readonly bool scaleSizeWithZoom;
 
-		public SpriteEffect(WPos pos, World world, string image, string sequence, string palette, bool visibleThroughFog = false, bool scaleSizeWithZoom = false)
+		public SpriteEffect(WPos pos, World world, string image, string sequence, string palette, int facing = 0, bool visibleThroughFog = false, bool scaleSizeWithZoom = false)
 		{
 			this.world = world;
 			this.pos = pos;
 			this.palette = palette;
 			this.scaleSizeWithZoom = scaleSizeWithZoom;
 			this.visibleThroughFog = visibleThroughFog;
-			anim = new Animation(world, image);
+			anim = new Animation(world, image, () => facing);
 			anim.PlayThen(sequence, () => world.AddFrameEndTask(w => w.Remove(this)));
 		}
 
--- a/OpenRA.Mods.Common/Traits/Render/LeavesTrails.cs
+++ b/OpenRA.Mods.Common/Traits/Render/LeavesTrails.cs
@@ -80,7 +80,7 @@ namespace OpenRA.Mods.Common.Traits.Render
 
 				if (Info.TerrainTypes.Contains(type) && !string.IsNullOrEmpty(Info.Image))
 					self.World.AddFrameEndTask(w => w.Add(new SpriteEffect(pos, self.World, Info.Image,
-						Info.Sequences.Random(Game.CosmeticRandom), Info.Palette, Info.VisibleThroughFog)));
+						Info.Sequences.Random(Game.CosmeticRandom), Info.Palette, 0, Info.VisibleThroughFog)));
 
 				cachedPosition = self.CenterPosition;
 				ticks = 0;
--- a/OpenRA.Mods.Common/Traits/Render/LeavesTrails.cs
+++ b/OpenRA.Mods.Common/Traits/Render/LeavesTrails.cs
@@ -52,13 +52,10 @@ namespace OpenRA.Mods.Common.Traits.Render
 		[Desc("Delay before first trail.")]
 		public readonly int StartDelay = 0;
 
-		[Desc("Position relative to body.")]
-		public readonly WVec Offset = WVec.Zero;
+		[Desc("Trail spawn positions relative to actor position. (forward, right, up) triples")]
+		public readonly WVec[] Offsets = { WVec.Zero };
 
-		[Desc("Use opposite offset for every second spawned trail.")]
-		public readonly bool AlternateOffset = false;
-
-		[Desc("Should the trail spawn at the last position or at current position?")]
+		[Desc("Should the trail spawn relative to last position or current position?")]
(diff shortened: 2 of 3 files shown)


OpenRA.Mods.Common/Traits/Render/LeavesTrails.cs
```diff
@@ -49,41 +49,89 @@ public class LeavesTrailsInfo : UpgradableTraitInfo
...
 				if (Info.TerrainTypes.Contains(type) && !string.IsNullOrEmpty(Info.Image))
 					self.World.AddFrameEndTask(w => w.Add(new SpriteEffect(pos, self.World, Info.Image,
-						Info.Sequences.Random(Game.CosmeticRandom), Info.Palette, Info.VisibleThroughFog)));
+						Info.Sequences.Random(Game.CosmeticRandom), Info.Palette, spawnFacing, Info.VisibleThroughFog)));
 
 				cachedPosition = self.CenterPosition;
+				cachedFacing = facing != null ? facing.Facing : 0;
 				ticks = 0;
+
+				if (!evenNumber)
```
pchote: evenNumber ^= true;

OpenRA.Mods.Common/Traits/Render/LeavesTrails.cs
```diff
@@ -49,41 +49,89 @@ public class LeavesTrailsInfo : UpgradableTraitInfo
...
 		public readonly int MovingInterval = 0;
 
+		[Desc("Delay before first trail.")]
+		public readonly int StartDelay = 0;
+
+		[Desc("Position relative to body.")]
+		public readonly WVec Offset = WVec.Zero;
+
+		[Desc("Use opposite offset for every second spawned trail.")]
+		public readonly bool AlternateOffset = false;
```
pchote: A list of offsets that are looped might be cleaner and more general.

reaperrr:
I get an IndexOutOfRangeException with my attempt at looping through offsets, not sure what I need to change to fix that.

OpenRA.Mods.Common/Traits/Render/LeavesTrails.cs
```diff
@@ -109,11 +106,15 @@ public void Tick(Actor self)
 				var spawnCell = Info.SpawnAtLastPosition ? self.World.Map.CellContaining(cachedPosition) : self.World.Map.CellContaining(self.CenterPosition);
 				var type = self.World.Map.GetTerrainInfo(spawnCell).Type;
 
-				var offsetRotation = Info.Offset.Rotate(body.QuantizeOrientation(self, self.Orientation));
-				var spawnOffset = body.LocalToWorld(Info.AlternateOffset && evenNumber ? -offsetRotation : offsetRotation);
+				if (offset < Info.Offsets.Length)
```
pchote: If `offset == Info.Offsets.Length - 1` it will pass this check and be incremented to `Info.Offsets.Length`, which then gives your OutOfBoundsException.

You can simplify this if/else block to:

```csharp
if (++offset >= Info.Offsets.Length)
    offset = 0;
```

pchote:
Looks like Github is somehow screwing up commit-level comments, thinking that they're outdated.  See my comment just above for a fix.

reaperrr:
@pchote I'm still getting
```
Exception of type `System.IndexOutOfRangeException`: Der Index war außerhalb des Arraybereichs.
   bei OpenRA.Mods.Common.Traits.LeavesTrails.Tick(Actor self)
   bei OpenRA.World.<Tick>b__7(TraitPair`1 x) in e:\ORA_work_coding\OpenRA.Game\World.cs:Zeile 329.
   bei OpenRA.WorldUtils.DoTimed[T](IEnumerable`1 e, Action`1 a, String text) in e:\ORA_work_coding\OpenRA.Game\WorldUtils.cs:Zeile 70.
   bei OpenRA.World.Tick() in e:\ORA_work_coding\OpenRA.Game\World.cs:Zeile 0.
   bei OpenRA.Game.InnerLogicTick(OrderManager orderManager) in e:\ORA_work_coding\OpenRA.Game\Game.cs:Zeile 522.
   bei OpenRA.Game.LogicTick() in e:\ORA_work_coding\OpenRA.Game\Game.cs:Zeile 547.
   bei OpenRA.Game.Loop() in e:\ORA_work_coding\OpenRA.Game\Game.cs:Zeile 674.
   bei OpenRA.Game.Run() in e:\ORA_work_coding\OpenRA.Game\Game.cs:Zeile 714.
   bei OpenRA.Program.Run(String[] args) in e:\ORA_work_coding\OpenRA.Game\Support\Program.cs:Zeile 118.
   bei OpenRA.Program.Main(String[] args) in e:\ORA_work_coding\OpenRA.Game\Support\Program.cs:Zeile 40.
```

with your fix.

reaperrr:
Nevermind, silly me had a typo in the test case and no default value in `Offsets`.

reaperrr:
Updated, works now.

Mailaender:
Confirmed, works as promised. :white_check_mark:

reaperrr:
On a side-note, the facings support added to `SpriteEffect` will allow us to get rid of both `Smoke` and `Explosion` effects since they feature nothing exclusive and are effectively just more limited reverse-duplicates of `SpriteEffect`.
That would also have the positive effect of a consistent parameter order, which is currently inconsistent between all 3, and allow directional explosion effects down the line.

abcdefg30:
Needs a rebase.

reaperrr:
Rebased. Since the test case was responsible for the need to rebase, I moved it to a separate branch:
https://github.com/reaperrr/OpenRA/tree/improve-LeavesTrails-test

OpenRA.Game/Effects/SpriteEffect.cs
```diff
@@ -23,14 +23,14 @@ public class SpriteEffect : IEffect
 		readonly bool visibleThroughFog;
 		readonly bool scaleSizeWithZoom;
 
-		public SpriteEffect(WPos pos, World world, string image, string sequence, string palette, bool visibleThroughFog = false, bool scaleSizeWithZoom = false)
+		public SpriteEffect(WPos pos, World world, string image, string sequence, string palette, int facing = 0, bool visibleThroughFog = false, bool scaleSizeWithZoom = false)
```
obrakmann: Is there a special reason that you're adding the `facing` parameter smack in the middle? If you'd put it on the end, people (well, those few with 3rd party dlls) wouldn't need to adjust their code.

obrakmann:
Looks ok otherwise I'm really looking forward to the eye candy this will bring. :+1:

OpenRA.Game/Effects/SpriteEffect.cs
reaperrr: Well, one thing I had in mind was to later ditch inferior special case effects like `Explosion`, `Smoke` and so on and use `SpriteEffect` for those as well, and I wanted to avoid having to explicitly set `false` for the other two optional parameters. But I guess you're right, downstream compatibility is a bit more important.

reaperrr:
Updated.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/4da65db625de26ba0629bab8f8166c9dcb989e5b)

PR #11113, opened 2016-04-14
Fix "Team colors" affecting (superweapon) countdown timers

### LINKED ISSUE #11048: "Team colors" affecting (superweapon) countdown timers
(Playtest 20160403)
When "team colors" are enabled the coutdown timers of super weapons (like GPS satelite) gets colored in team colors as well. So you could see **multiple** GPS **countdown timers** all **in the same** red **color.**

I say this is not helping since you usually want to know specifically (in order to prevent it) who's building it. Disabling team colors during game is an option (but an additional step) and many (new) players may not know how to do it (shortcut).

It would be usefull if **both pieces of information** would be represented: 
- TEAM "membership" (color) and 
- PLAYER COLOR.

Example:
- Ally: GPS countdown 3:00
- Enemy: GPS countdown: 2:15
  (both in their colors)

Additionaly words **"Ally" and "Enemy" could be team colored** (to distinguish multiple teams).
... or "Enemy (Team 2)", "Enemy (Team 3)" ...

Best regards.

abcdefg30:
Fixes #11048.

This just drops the entire `PlayerStanceColor` thing from the `SupportPowerTimerWidget`.
If we still want to display some information, we could add some kind of `(Enemy)`/`(Ally)` pre-/suffixes?

### DIFF
--- a/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
@@ -54,12 +54,7 @@ namespace OpenRA.Mods.Common.Widgets
 				var time = WidgetUtils.FormatTime(p.RemainingTime, false, timestep);
 				var text = Format.F(p.Info.Description, time);
 				var self = p.Instances[0].Self;
-				var playerColor = self.Owner.Color.RGB;
-
-				if (Game.Settings.Game.UsePlayerStanceColors)
-					playerColor = self.Owner.PlayerStanceColor(self);
-
-				var color = !p.Ready || Game.LocalTick % 50 < 25 ? playerColor : Color.White;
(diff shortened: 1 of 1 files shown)


abcdefg30:
Btw, you can test this easily with my old [`Timer Testmap`](http://resource.openra.net/maps/11518/).

pchote:
I'm not convinced that #11048 is a bug that should be fixed. While the argument there is reasonable, i think it sucks to have the timers as an inconsistent exception to the team color rule. Team colors must be explicitly enabled, after all.

obrakmann:
I think I agree that making a special exception for the timers is the wrong way to go, and that #11048 should be closed with "works as designed".

reaperrr:
I'm also not convinced. At the very least I think this shouldn't be rushed into this release, as it's neither a bug nor something overly important either way.

reaperrr:
We have 3 people inclined against it, so rather than letting it die slowly due to lack of attention and clog up the PR queue, I guess it's better to close it.

Issue #6955, opened 2014-11-14
Silo's needed alert goes of way too often


chriseilander:
I don't know at which point the originals started nagging about building silo's, but I think that due the increased gameplay speed of openra, the "silo's needed" alert goes of way too often.

I did see some nice ideas from https://github.com/OpenRA/OpenRA/issues/3816 that wouldn't have the need for this alert, and changing the mechanic of the silo's completely. I can live with that.

colonelpopcorn:
Is this something we should surface in the options menu? This sounds more like a preference that should be able to be tweaked by each individual user.

PR #11114, opened 2016-04-14
Added a new checkbox to settings, where you can turn off the silo war…

SzabKel:
Added a new checkbox to settings, where you can turn off the silo warnings, also changed the default silo warning time to 30s
Did the following request: #6955

### DIFF
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -150,6 +150,7 @@ namespace OpenRA
 		public string Device = null;
 
 		public bool CashTicks = true;
+		public bool SiloWarnings = true;
 		public bool Mute = false;
 	}
 
--- a/OpenRA.Mods.Common/Traits/Player/ResourceStorageWarning.cs
+++ b/OpenRA.Mods.Common/Traits/Player/ResourceStorageWarning.cs
@@ -17,7 +17,7 @@ namespace OpenRA.Mods.Common.Traits
 	public class ResourceStorageWarningInfo : ITraitInfo, Requires<PlayerResourcesInfo>
 	{
 		[Desc("Interval, in seconds, at which to check if more storage is needed.")]
-		public readonly int AdviceInterval = 10;
+		public readonly int AdviceInterval = 30;
 
 		[Desc("The percentage threshold above which a warning is played.")]
 		public readonly int Threshold = 80;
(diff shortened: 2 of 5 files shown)


pchote:
I don't like the idea of adding a checkbox for this.  We should instead fix the warnings to be less annoying, which will improve the game for everyone.

Phrohdoh:
I'm behind the idea of not forcing the notification but to compromise perhaps a value `< 0` could be introduced as a "never play the sound" option.

SzabKel:
I thought a checkbox would be a good idea, because the Cash Ticks option is basically the same.

SzabKel:
What do you think about the following: If the notification goes off, the first time it does we add +15s for the next notification, if the notification does not go off, we reset the timer. Something like this: Warning 1st time(+15 to the next notif), after 15 sec the 2nd warning goes off, after 30 sec the next one...etc.

obrakmann:
@SzabKel: I like your last suggestion. Sounds like a good solution.

pchote:
IMO the simplest and best fix is to set the notification up the same way as the new insufficient funds notification, triggered on the harv unloading being blocked.

LipkeGu:
@pchote :+1:

LipkeGu:
I compared the TS silo logic and the one from OpenRA and indeed the Logic from openra is different as the original silo logic from TS. Fixing the OpenRA Logic should avoid this silo spam

1. Refinery is empty ... -> http://imgur.com/Dl7e5Ve
2. Refinery is get filled up -> http://imgur.com/SdV0PFH
3. but silo is empty -> http://imgur.com/IwPZXA8

MustaphaTR:
I find this shared storage logic more. This is also like that in D2K. Making them load to the Refinery first has an annoying problem that if u lose it u lose every money u have. But in current OpenRA logic if you lose Refinery u lose only money that is refinery which leaves some money is Silos.

I'm not sure if this is really working like that tho. Do we lose money when we lose Ref in OpenRA? This does happen in TS and RA2(Ares) and quite annoying with their Silo Logic.

abcdefg30:
@SzabKel what's the status here?

SzabKel:
@abcdefg30 I think this should be closed, the code here is not needed and I don't think I will work on it anymore.

PR #11117, opened 2016-04-14
NotifyActorsBelow trait for aircraft

teees:
Notify actors implementing `INotifyActorAbove`.

Prerequisite for firestorm generator

### DIFF
--- a/OpenRA.Mods.TS/Traits/NotifyActorsBelow.cs
+++ b/OpenRA.Mods.TS/Traits/NotifyActorsBelow.cs
@@ -1,0 +1,53 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.TS.Traits
+{
+	[Desc("Actor notifies other (ground) actors when positions overlap.")]
+	public interface INotifyActorAbove
+	{
+		string NotifyActorType { get; }
+
+		void NotifyActorAbove(Actor self, Actor above);
+	}
+
+	[Desc("Actor notifies other (ground) actors when positions overlap.")]
+	public class NotifyActorsBelowInfo : ITraitInfo
+	{
+		[Desc("Which actor types to notify, e.g. building")]
+		public readonly HashSet<string> NotifyActorTypes = new HashSet<string>();
+
+		public virtual object Create(ActorInitializer init) { return new NotifyActorsBelow(init, this); }
+	}
+
+	public class NotifyActorsBelow : ITick
+	{
+		readonly NotifyActorsBelowInfo info;
+
+		public NotifyActorsBelow(ActorInitializer init, NotifyActorsBelowInfo info)
+		{
+			this.info = info;
+		}
+
+		public void Tick(Actor self)
+		{
+			self.World.ActorMap.GetActorsAt(self.Location)
+				.Where(a => a != self)
+				.Do(a => a.TraitsImplementing<INotifyActorAbove>()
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.TS/Traits/NotifyActorsBelow.cs
```diff
@@ -0,0 +1,39 @@
...
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.TS.Traits
+{
+	public interface INotifyActorAbove
+	{
+		void NotifyActorAbove(Actor self, Actor above);
+	}
+
+	public class NotifyActorsBelowInfo : TraitInfo<NotifyActorsBelow> { }
```
pchote: Can we get a HashSet<string> of notification types here to avoid overlaps if something else wants to use this in the future?

OpenRA.Mods.TS/Traits/NotifyActorsBelow.cs
```diff
@@ -0,0 +1,52 @@
...
+	{
+		[Desc("Which actor types to notify, e.g. building")]
+		public readonly HashSet<string> NotifyActorTypes = new HashSet<string>();
+
+		public virtual object Create(ActorInitializer init) { return new NotifyActorsBelow(init, this); }
+	}
+
+	public class NotifyActorsBelow : ITick
+	{
+		NotifyActorsBelowInfo info;
```
Mailaender: readonly

OpenRA.Mods.TS/Traits/NotifyActorsBelow.cs
```diff
@@ -0,0 +1,52 @@
...
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.Traits;
+using System.Collections.Generic;
```
Mailaender: Ordering, see also StyleCop output.

Phrohdoh:
Could you add a trait description? Currently this doesn't offer any insight as to what this trait is supposed to be used for.

teees:
Added a description as well. It informs actors at the same position. This is mainly useful so that aircraft informs firestorm tiles of them being above.

Mailaender:
I have no idea how to test this in isolation, but it will probably work. :+1:

OpenRA.Mods.TS/Traits/NotifyActorsBelow.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace OpenRA.Mods.TS.Traits
+{
+	public interface INotifyActorAbove
+	{
+		string NotifyActorType { get; }
+
+		void NotifyActorAbove(Actor self, Actor above);
+	}
+
+	[Desc("Actor informs other actors of intersecting positions")]
```
GraionDilach: This looks quite nonEnglish to me.

GraionDilach:
Considering that the interface is unused atm, I guess it will work, but noidea on the how aspect myself either. I mean we've got a hashset to set on this trait but since the interface only has a getter, there is no way to specify what actors the strings in the hashset would imply.

teees:
This was intended as pretty one shot trait, used for looking up any aircraft flying over firestorm defense walls in a cheap way. (other than let every single wall patch search for any aircraft above). If you have any suggestions / ideas on how to make this more usable for other units, please let me know.

teees:
Rebased

OpenRA.Mods.TS/Traits/NotifyActorsBelow.cs
```diff
@@ -0,0 +1,53 @@
...
+ */
+#endregion
+
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.TS.Traits
+{
+	[Desc("Actor notifies other (ground) actors when positions overlap.")]
```
GraionDilach: You don't need to Desc an interface.

Phrohdoh: xmldoc would be best.

teees:
Closing this, I'll revert this to a firestorm/energywall only trait

PR #11120, opened 2016-04-15
Fix barrel recoil being set immediately, not when actually firing

teees:
Before: https://streamable.com/5g73
After: https://streamable.com/5lkz


GraionDilach:
:+1:

teees:
ammended the date to fix that appveyor timeout

pchote:
Code looks reasonable, but I haven't tested it.  Need to make sure that this doesn't change anything balance related in the gen1 mods.

Mailaender:
Doesn't seem to effect turreted units in RA.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/4a7400dc0971d6c272286d320ad3761201b46d16...e0039c82813f746304524d0de31708304011f51e)

PR #11122, opened 2016-04-15
Move Render traits to Traits.Render namespace

reaperrr:
Sort of follow-up to #11097 and #11017.

While most render traits follow a relatively consistent naming scheme, this still makes it easier to tell them apart from other trait categories in the wiki (and presumably some dev GUIs).

### DIFF
--- a/OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
+++ b/OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
@@ -11,6 +11,7 @@
 
 using System.Linq;
 using OpenRA.Mods.Common.Traits;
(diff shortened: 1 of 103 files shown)


Mailaender:
Good idea. :+1:

Phrohdoh:
Along this vein I'd also like to see the debug overlays grouped together.
Dumping traits into the root of `Traits` hasn't served me well.

reaperrr:
About the debug overlays, I'm not sure what to do about `CombatDebugOverlay`.
The other debug overlays are world traits and would be better located in a `Traits.World` namespace in my opinion (maybe with `.Debug` added), but CDO is a trait for individual actors and would better fit into `Traits.Render`.

obrakmann:
:+1:

PR #11124, opened 2016-04-15
Add tileset depth metadata support.

pchote:
This adds support for per-tile Z-ramp and Z-offsets to the tileset code and gen2 tileset importer.  The TS tilesets are reimported with proper values.

The defaults are set up so that tiles without any extra depth metadata will be "flat" at ground level – the test commit enables this in RA for illustration.  The TS tilesets are set up so that tile tile depths are correctly zeroed at the middle of the tile.

The expected behaviour is (for mods with `EnableDepth: true`) for sprite data to be cut off for all y < CenterPosition.Y, because for flat sprites with Z = CenterPosition.Z anything lower is rendered behind the terrain.  A future PR that adds support for sprite depth maps will fix that.  This puts the ground work in place by setting the correct terrain depth.

### DIFF
--- a/OpenRA.Game/Graphics/SheetBuilder.cs
+++ b/OpenRA.Game/Graphics/SheetBuilder.cs
@@ -59,15 +59,15 @@ namespace OpenRA.Graphics
 			this.allocateSheet = allocateSheet;
 		}
 
-		public Sprite Add(ISpriteFrame frame) { return Add(frame.Data, frame.Size, frame.Offset); }
-		public Sprite Add(byte[] src, Size size) { return Add(src, size, float2.Zero); }
-		public Sprite Add(byte[] src, Size size, float2 spriteOffset)
+		public Sprite Add(ISpriteFrame frame) { return Add(frame.Data, frame.Size, 0, frame.Offset); }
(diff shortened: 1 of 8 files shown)


reaperrr:
Needs rebase.

pchote:
Rebased.

Mailaender:
Units correctly vanish behind cliffs. Sprites are cut-off systematically, which I assume then is exactly the wanted behavior. It is a bit strange that veins are not rendered, while Tiberium is regardless of the height level, but that is probably also to be expected. No negative side effects on ra2 and d2k. :white_check_mark:

LipkeGu:
So far i can tell works good ... and fiends are clipped behind the cliffs perfectly fast :) i know @pchote mentored a few PR before that the rendering will fixd in a next Pr (Buildings) and yeah... i can live a few days with half cutted buildings :) :+1:

reaperrr:
:+1: 

Test commit can be removed.

pchote:
Removed test commit.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/0bc033f9d9da111c43ffa036db40b7933f3dd753...bfdd5cf59a078b899c253c637a837e8196b049b1)

PR #11129, opened 2016-04-16
You can now add and remove orders from RejectsOrders

### LINKED ISSUE #10978: Make RejectsOrders upgradable
Requested by @ABrandau.

SzabKel:
Closes #10978

### DIFF
--- a/OpenRA.Game/Traits/RejectsOrders.cs
+++ b/OpenRA.Game/Traits/RejectsOrders.cs
@@ -41,5 +41,17 @@ namespace OpenRA.Traits
 			var r = self.TraitOrDefault<RejectsOrders>();
 			return r == null || r.Except.Contains(orderString);
 		}
+
+		public static void AddOrder(this Actor self, string orderString)
+		{
+			var r = self.TraitOrDefault<RejectsOrders>();
+			r.Except.Add(orderString);
+		}
+
+		public static void RemoveOrder(this Actor self, string orderString)
+		{
+			var r = self.TraitOrDefault<RejectsOrders>();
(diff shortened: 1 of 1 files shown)


OpenRA.Game/Traits/RejectsOrders.cs
```diff
@@ -41,5 +41,17 @@ public static bool AcceptsOrder(this Actor self, string orderString)
 			var r = self.TraitOrDefault<RejectsOrders>();
 			return r == null || r.Except.Contains(orderString);
 		}
+
+		public static void AddOrder(this Actor self, string orderString)
+		{
+			var r = self.TraitOrDefault<RejectsOrders>();
+			r.Except.Add(orderString);
```
pchote: These will throw NullReferenceExceptions if the actor doesn't have RejectsOrders.

pchote:
This doesn't appear to implement the upgrade support requested in #10978.

SzabKel:
What does upgrading a reject order means?

pchote:
It means implementing the `IUpgradable` / `UpgradableTrait` logic so that it can be toggled using the [actor upgrades](https://github.com/OpenRA/OpenRA/wiki/Actor-Upgrades) system.

GraionDilach:
`RejectsOrders` already had a YAMl tag of filtering out orders even. I don't understand what this PR is supposed to do at the current stage.

reaperrr:
@SzabKel Sorry, we can't take this as it is now so I'm closing this to clear the PR queue. If you re-implement it they way it was meant in #10978, feel free to update and reopen it.

PR #11135, opened 2016-04-16
Fix PixelDoubling not working properly after zooming with mousewheel

obrakmann:
I decided to write it this way (instead of setting `Game.Settings.Graphics.PixelDouble` in `Viewport.Zoom` appropriately) because I see the double pixel setting as the user's preference that the game uses at the start of a match and which should not be changed only because the user decided to zoom in a bit.

### DIFF
--- a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
+++ b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
@@ -345,8 +345,16 @@ namespace OpenRA.Widgets
 
 		bool TogglePixelDouble()
 		{
-			Game.Settings.Graphics.PixelDouble ^= true;
-			worldRenderer.Viewport.Zoom = Game.Settings.Graphics.PixelDouble ? 2 : 1;
+			if (worldRenderer.Viewport.Zoom == 1f)
+				worldRenderer.Viewport.Zoom = 2f;
+			else
+			{
+				// Reset zoom to regular view if it was anything else before
+				// (like a zoom level only reachable by using the scroll wheel).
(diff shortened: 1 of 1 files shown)


reaperrr:
I understand your reasoning, but I think you should at least add a comment explaining that to the code, because otherwise the line
`Game.Settings.Graphics.PixelDouble = worldRenderer.Viewport.Zoom == 2f;`
might look like weird to someone who doesn't know the reason.

reaperrr:
Looks good to me :+1:

obrakmann:
I added a comment, but I'm not sure if that is what you were looking for. To be honest, I don't quite get what's hard to understand about that line. It's pretty obvious in the context of the whole method.

reaperrr:
> I added a comment, but I'm not sure if that is what you were looking for.

Not entirely, but that's my fault for misreading the line I mentioned above. I think the comment can stay anyway.

pchote:
This feels a little hacky to me.  Wouldn't it be make more sense to have the hotkey cycle zoom levels (like we do in the editor) instead?  If the saved level is outside the allowed range it can be changed to the closest matching on map load.

obrakmann:
Personally, if it was up to me, I'd change it so that the hotkey doesn't modify the default at all, then make it cycle through all the zoom levels.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/8fc80038a0008becd45cda7e8596056e2bf52ac2...0ba24d092681070f0435f8d5a1f75e3085d998a9)

Issue #3231, opened 2013-05-06
Change in game music tracks with next and previous media keys


dan9550:
I thought it would be pretty cool idea if you were able to control the in game music player with the universal media keys on keyboards and such.

PR #11136, opened 2016-04-16
Added music controlling hotkeys.

SzabKel:
You can change them in options, by default it is set to audio keys.
Resolves?: #3231

### DIFF
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -284,6 +284,11 @@ namespace OpenRA
 		public Hotkey ReplaySpeedFastKey = new Hotkey(Keycode.F7, Modifiers.None);
 		public Hotkey ReplaySpeedMaxKey = new Hotkey(Keycode.F8, Modifiers.None);
 
+		public Hotkey NextTrack = new Hotkey(Keycode.AUDIONEXT, Modifiers.None);
+		public Hotkey PreviousTrack = new Hotkey(Keycode.AUDIOPREV, Modifiers.None);
+		public Hotkey StopMusic = new Hotkey(Keycode.AUDIOSTOP, Modifiers.None);
+		public Hotkey PauseMusic = new Hotkey(Keycode.AUDIOPLAY, Modifiers.None);
+
 		static readonly Func<KeySettings, Hotkey>[] ProductionKeys = GetKeys(24, "Production");
 		static readonly Func<KeySettings, Hotkey>[] SupportPowerKeys = GetKeys(6, "SupportPower");
 
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -541,6 +541,7 @@
     <Compile Include="UtilityCommands\ExtractLanguageStringsCommand.cs" />
     <Compile Include="UtilityCommands\ExtractLuaDocsCommand.cs" />
     <Compile Include="UtilityCommands\ExtractTraitDocsCommand.cs" />
+    <Compile Include="Widgets\Logic\Ingame\MusicControllerLogic.cs" />
     <Compile Include="WorldExtensions.cs" />
     <Compile Include="UtilityCommands\GetMapHashCommand.cs" />
     <Compile Include="UtilityCommands\Glob.cs" />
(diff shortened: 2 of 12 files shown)


OpenRA.Game/Settings.cs
```diff
@@ -284,6 +284,11 @@ public class KeySettings
 		public Hotkey ReplaySpeedFastKey = new Hotkey(Keycode.F7, Modifiers.None);
 		public Hotkey ReplaySpeedMaxKey = new Hotkey(Keycode.F8, Modifiers.None);
 
+		public Hotkey NextTrack = new Hotkey(Keycode.AUDIONEXT, Modifiers.None);
```
pchote: These don't seem to work on OSX.  Do they work on Linux or Windows?

OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
```diff
@@ -291,11 +292,79 @@ public override bool HandleKeyPress(KeyInput e)
 					return ToggleMute();
 				else if (key == Game.Settings.Keys.TogglePlayerStanceColorsKey)
 					return TogglePlayerStanceColors();
+				else if (key == Game.Settings.Keys.NextTrack)
```
pchote: Implementing these in `WorldInteractionControllerWidget` means that they won't work in any of the menus.  It would be better if these hotkeys could work anywhere where music is playing.

You'll also need to be careful to not break the special case background music on the shellmap, on game completion, and ambient effects (rain etc).

pchote: You should be use the methods exposed on MusicPlaylist, which know how to handle these special cases.  See `MusicPlayerLogic` for an example of how to do this.

OpenRA.Game/Settings.cs
SzabKel: I tested it on Windows

SzabKel: Also, I changed the default keys to numpad keys, because my keyboard doesn't have multimedia keys.

OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
SzabKel: Okay I get it. 
I started using `WorldInteractionControllerWidget ` because it had `HandleKeyPress(KeyInput e)` and MusicPlayerLogic don't and couldn't find any music player widget with handle method. So where should I handle the keypress?

OpenRA.Game/Settings.cs
SzabKel: I saw the AUDIO* keys in the Keycode class, I assumed they are working.

pchote: This may be an OSX specific limitation, which is why I asked about the other platforms.

OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
pchote: You could add a LogicKeyListener widget to the yaml and then attach logic to its OnKeyPress handler.

OpenRA.Game/Settings.cs
Mailaender: My laptop only works by using a `Fn` button combo for media keys and that `Fn` key is not understood:

> The key you just pressed is not recognized by SDL. To help get this fixed, please report this to the SDL mailing list <sdl@libsdl.org> X11 KeyCode 151 (143), X11 KeySym 0x1008FF2B (XF86WakeUp).

but that is clearly an SDL upstream problem.

Mailaender:

HenrytheSlav:
OK, I just tested it. Unfortunately I do not have the audio keys, so I had to use KeyTweak to remap them. Weird thing is, in Windows Media Player remaped keys work fine, but in ORA they don't, if I try to assign my remaped key to the Next Audio, it assign letter instead. Then both the letter and my remaped key work okay.
That said, the function itself works as promised, so :+1: for me.

Mailaender:
Your first merged pull request for this project if I am not mistaken! :tada:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8a4c6eb999e718a921335ded2706349fe009165f...746f30e89724085addfb391bcf2a47dda631475b)

PR #11137, opened 2016-04-17
Removed unnecessary null checks in CheckTargetHealthRadius

Mailaender:
Coverity complained that this leaves logically dead code while assigning `targetExtraSearchRadius`. It will never hit the `WDist.Zero` case.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckTargetHealthRadius.cs
+++ b/OpenRA.Mods.Common/Lint/CheckTargetHealthRadius.cs
@@ -58,9 +58,16 @@ namespace OpenRA.Mods.Common.Lint
 					if (bullet == null && missile == null && areabeam == null)
 						continue;
 
-					var targetExtraSearchRadius = bullet != null ? bullet.TargetExtraSearchRadius :
-						missile != null ? missile.TargetExtraSearchRadius :
-						areabeam != null ? areabeam.TargetExtraSearchRadius : WDist.Zero;
+					var targetExtraSearchRadius = WDist.Zero;
+
+					if (bullet != null)
+						targetExtraSearchRadius = bullet.TargetExtraSearchRadius;
+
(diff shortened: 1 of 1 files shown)


GraionDilach:
This broke the lint check.

Mailaender:
Should be fixed now.

abcdefg30:
:+1:

OpenRA.Mods.Common/Lint/CheckTargetHealthRadius.cs
```diff
@@ -55,13 +55,13 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Ruleset ru
...
 					var areabeam = weaponInfo.Value.Projectile as AreaBeamInfo;
 
-					if (bullet == null && missile == null && areabeam == null)
-						continue;
-
 					var targetExtraSearchRadius = bullet != null ? bullet.TargetExtraSearchRadius :
 						missile != null ? missile.TargetExtraSearchRadius :
 						areabeam != null ? areabeam.TargetExtraSearchRadius : WDist.Zero;
 
+					if (targetExtraSearchRadius == WDist.Zero)
```
pchote: This will bypass the checks if one of the projectiles actually does set this to 0 (which is a legal value to set via yaml, but invalid to use ingame).

Mailaender: Solved it differently now.

abcdefg30:
My :+1: still stands.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/415f6a9f406a9ba1afb466650acc8a227db9b214...0bc033f9d9da111c43ffa036db40b7933f3dd753)

PR #11140, opened 2016-04-17
Fixed NullReferenceExceptions if package opening fails.

Mailaender:
`OpenPackage` explicitly returns null if the file suffixes don't match .zip or .oramap, which I assume doesn't happen very often in practice, but Coverity is right, that this should not crash here.

### DIFF
--- a/OpenRA.Game/Map/MapPreview.cs
+++ b/OpenRA.Game/Map/MapPreview.cs
@@ -486,8 +486,13 @@ namespace OpenRA
 						Game.RunAfterTick(() =>
 						{
 							var package = modData.ModFiles.OpenPackage(mapFilename, mapInstallPackage);
-							UpdateFromMap(package, mapInstallPackage, MapClassification.User, null, GridType);
-							onSuccess();
+							if (package == null)
+								innerData.Status = MapStatus.DownloadError;
+							else
+							{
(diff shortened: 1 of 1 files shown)


abcdefg30:
:+1:

OpenRA.Game/Map/MapPreview.cs
```diff
@@ -486,8 +486,11 @@ public void Install(Action onSuccess)
 						Game.RunAfterTick(() =>
 						{
 							var package = modData.ModFiles.OpenPackage(mapFilename, mapInstallPackage);
-							UpdateFromMap(package, mapInstallPackage, MapClassification.User, null, GridType);
-							onSuccess();
+							if (package != null)
```
pchote: This will leave the lobby UI in an inconsistent state (probably forever showing the "downloading" bar).

pchote: You can add an else branch that sets MapStatus.DownloadError.

Mailaender: Done.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/bf50abdc9792eb36db9128c764ffbb7b20d3ff4e)

PR #11141, opened 2016-04-17
Removed redundant targetPassedBy check.

Mailaender:
This is probably a copy & paste mistake as you can see the identical block a few lines below in another conditional. In line 599 the whole block is already checking against `targetPassedBy` so Coverity is right that it can never be true and therefore only introduces dead code.


abcdefg30:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/415f6a9f406a9ba1afb466650acc8a227db9b214...0bc033f9d9da111c43ffa036db40b7933f3dd753)

PR #11144, opened 2016-04-17
Generate frozen actor renderables for all map-placed actors.

### LINKED ISSUE #11080: Pre-built walls and fences (neutral) are invisible with 'explored map'.
Relevant issue for both players and spectators. Many official maps have fences, some around oil derricks, e.g. it's a bit awkward revealing fences when you approach oil derricks. Also maps such as Keep of The Grass 2 has walls defending the middle ore/gems.

pchote:
Fixes #11080.  This is a rather obvious polish bug in the current playtest, so if there are no objections I would like to see this included in the next playtest.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
+++ b/OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
@@ -20,8 +20,6 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("This actor will remain visible (but not updated visually) under fog, once discovered.")]
 	public class FrozenUnderFogInfo : ITraitInfo, Requires<BuildingInfo>, IDefaultVisibilityInfo
 	{
-		public readonly bool StartsRevealed = false;
-
 		[Desc("Players with these stances can always see the actor.")]
 		public readonly Stance AlwaysVisibleStances = Stance.Ally;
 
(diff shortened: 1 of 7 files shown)


OpenRA.Mods.Common/Traits/Modifiers/FrozenUnderFog.cs
```diff
@@ -20,8 +20,6 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("This actor will remain visible (but not updated visually) under fog, once discovered.")]
 	public class FrozenUnderFogInfo : ITraitInfo, Requires<BuildingInfo>, IDefaultVisibilityInfo
 	{
-		public readonly bool StartsRevealed = false;
```
Mailaender: This will need an upgrade rule for clean up.

Mailaender:
This doesn't just fix the issue which is probably only updating the fence rules, but rather changing behavior completely. Now you can't differentiate between map props which should be revealed and let's say mission objectives the player should search on a fogged map.

pchote:
The old behaviour doesn't make logical sense: it would make specific actor types always show under the fog when they are first spawned, without giving any way to say that individual actors of a type should be revealed.

pchote:
Maps that want to have a building hidden under the fog can spawn it after game start through Lua.

Phrohdoh:
That last point goes entirely against the point of placing actors in the editor and shoving them into the global table, I don't think it is fair to say that and expect people to use the tools in that way.

GraionDilach:
Why are mission objectives a problem here? The shroud will hide them, regardless their FoW status.

Generals FoW AFAIK works the same way this PR does.

Mailaender:
I am not saying that going this way is a bad move, just that it changes default behavior and existing missions and mini-games (outside of this repository) might rely on how the engine worked. Also there is no upgrade path for them to retain legacy behavior as the switch is also removed.

pchote:
The legacy property behaviour is effectively equivalent to removing the FrozenUnderFog trait from the actor.  I'll add an ActorInit that lets mappers hide individual actors, but that's more of a new feature request than a behaviour fixup.

pchote:
Updated: added upgrade rule and `HiddenUnderFogInit`.

obrakmann:
This breaks the `OnDiscovered` trigger in the Lua API (see Monster Tank Madness or the GDI05s).

pchote:
Fixed so that the previews are only explored if the "Explored Map" map option is enabled.

obrakmann:
:+1:

Mailaender:
I see no negative side effects. You can also keep the legacy behavior (if it is vital to anyone) by removing the shroud after the game has started (tested via dev cheats).

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/390fae4c2069ecac0092801ed8fed77913bf823c...6a3850d81aa44fbc3ee4939be12e7a30bf6d6eab)

PR #11147, opened 2016-04-18
Enable group-unit-selection for spectators.

### LINKED ISSUE #9492: Replay viewer and observer should enable selecting unit type by double clicking units
This was possible with earlier releases, and seem to have got lost in later development. It was so nice to  double click units (especially key units, e.g. harvesters, aircrafts...) as an observer and quickly see the amount of them. Same thing with structures. I'm hoping this is something that can easily come back to us.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27060847-replay-viewer-and-observer-should-enable-selecting-unit-type-by-double-clicking-units?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
This should, in theory, fix #9492.  The spectator and selection code is a bit horrible, so I'm not confident that this covers all the edge cases that we care about and/or is the best approach.

### DIFF
--- a/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
+++ b/OpenRA.Game/Widgets/WorldInteractionControllerWidget.cs
@@ -120,7 +120,8 @@ namespace OpenRA.Widgets
 						var unit = World.ScreenMap.ActorsAt(mousePos)
 							.WithHighestSelectionPriority(mousePos);
 
-						if (unit != null && unit.Owner == (World.RenderPlayer ?? World.LocalPlayer))
+						var p = World.RenderPlayer ?? World.LocalPlayer;
+						if (unit != null && (unit.Owner == p || p == null || p.Spectating))
 						{
 							var s = unit.TraitOrDefault<Selectable>();
 							if (s != null)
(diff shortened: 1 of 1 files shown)


Micr0Bit:
this will come in really handy for specs ... as you can (for example) have all AA's or SAM's activated by doubleclick on one ... and this reveals quickly when theres a gap in someones air-defense

chrisforbes:
Why aren't we just comparing to `unit.Owner` ?

pchote:
> Why aren't we just comparing to `unit.Owner`?

Because spectators are special snowflakes 😦

chrisforbes:
To clarify, I mean the directly clicked unit by `unit` -- the local player or render player shouldn't have any bearing on this. Behavior should be uniform for real players, specs, replay viewers, etc.

obrakmann:
This doesn't quite do the trick. It only works with double-clicking, not with the hotkey, and stops working in player views.

abcdefg30:
Needs a rebase as well.

pchote:
Unfortunately I don't have the time or interest to work on this further, so closing.

PR #11151, opened 2016-04-19
Fix the length of the idle2 sequence of tanya

abcdefg30:
Had a wrong frame in the end (tanya held her hands up 🙌 ):
tanyaidle

### DIFF
--- a/mods/ra/sequences/infantry.yaml
+++ b/mods/ra/sequences/infantry.yaml
@@ -725,7 +725,7 @@ e7:
 		Tick: 120
 	idle2:
 		Start: 248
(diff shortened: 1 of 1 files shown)


jaZzKCS:
"...and wave 'em like you just don't care!"

Mailaender:
Well spotted. :+1: It runs so fast that no one really noticed the glitch all those years.

HenrytheSlav:
:+1: Nice catch.

abcdefg30:
Btw, the credits for spotting this go to my brother. ;)

PR #11152, opened 2016-04-19
Fix reselecting a dropdown option in the lobby causing a server message

abcdefg30:
On bleed: Just reselect an already selected option and you get a "x changed y to z" notification
(although nothing changed).

I tried to put the checks as early as possible without changing other code (as we want to bail out early).


Mailaender:
:+1:

GraionDilach:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/415f6a9f406a9ba1afb466650acc8a227db9b214)

PR #11153, opened 2016-04-19
Use Contains method on Selection when possible.

RoosterDragon:
This allows the set to be utilized for a fast check, rather than degrading to a linear search via LINQ.


Mailaender:
:+1:

abcdefg30:

PR #11157, opened 2016-04-20
Fix the debris spawned when walls are destroyed in d2k

abcdefg30:
Looks like this on bleed:
https://www.dropbox.com/s/790kwftd2x4mfcv/d2kwalls.gif?dl=0 (File too big to upload it on github).

The debris configuration I added is taken from `^Defense`.

(Discovered by my brother, btw. ;))

### DIFF
--- a/mods/d2k/rules/structures.yaml
+++ b/mods/d2k/rules/structures.yaml
@@ -652,9 +652,9 @@ wall:
 		SellSounds: CHUNG.WAV
 	Guardable:
 	ThrowsShrapnel:
-		Weapons: Debris, Debris2, Debris3, Debris4
-		Pieces: 3, 7
-		Range: 2c0, 5c0
+		Weapons: Debris, Debris3
+		Pieces: 2, 2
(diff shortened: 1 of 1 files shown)


MustaphaTR:
Not, sure. IIRC that happens in original game too. Someone needs to check this.

reaperrr:
I think this is good enough as fix for stable :+1: 

For the next release cycle we should probably take another look at debris count in general, because I think it scaled with value or occupied cells in D2k.

GraionDilach:
:+1:

pchote:
I second @reaperrr's suggestion to revisit debris for the next release cycle.  There is more to it than just count: the debris in the original also had a probability of bouncing like the TS disk thrower disks.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/67e8bd52e6f3873b4bdcdbb899fd83a554f4ecb6)

PR #11158, opened 2016-04-21
Added INotifyFiredSalvo trait interface

teees:
Notifies once a full salvo has been fired.

Possible usecase is #11108 

Patch to test this on top of #11108 :
https://gist.github.com/teees/26e38296a44c58ffcf3ae9519579e818

### DIFF
--- a/OpenRA.Mods.Common/Traits/Armament.cs
+++ b/OpenRA.Mods.Common/Traits/Armament.cs
@@ -249,6 +249,9 @@ namespace OpenRA.Mods.Common.Traits
 					.Select(m => m.GetReloadModifier());
 				FireDelay = Util.ApplyPercentageModifiers(Weapon.ReloadDelay, modifiers);
 				Burst = Weapon.Burst;
+
+				foreach (var nfs in self.TraitsImplementing<INotifyFiredSalvo>())
+					nfs.FiredSalvo(self, target, this);
 			}
 
 			return barrel;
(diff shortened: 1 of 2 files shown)


Phrohdoh:
Shouldn't this be a part of #11108 then?
Why make a separate PR for just the interface?

If this is merged and not the other then we have unused code.

teees:
This is actually intended as a followup pr to #11108, which does what it should regarding the ts emp cannon. Once #11108 gets merged I can push https://gist.github.com/teees/26e38296a44c58ffcf3ae9519579e818 to this pr. (or the other way round, depending on which pr gets merged first).

Phrohdoh:
I've labeled this PR `Has dependencies` so it isn't merged before #11108.

pchote:
Can we normalize the naming between this and `Burst`?  It's not otherwise obvious that they're connected.  My gut feeling is to rename `Burst`, `BurstDelay` &rarr; `SalvoSize`, `SalvoDelay`

pchote:
A less invasive change would be to rename this to `INotifyBurstComplete`.

GraionDilach:
Go with `INotifyBurstComplete` then. :+1: also.

reaperrr:
Removing dependency label since it would be better to merge this first and then use it to do #11108 properly, in my opinion.

This still needs to be renamed to `INotifyBurstComplete`, though.

obrakmann:
:+1: 

@teees: can you do the rename, please? This can otherwise be merged and you can go on with #11108.

obrakmann:
Superseded by #11398

PR #11154, opened 2016-04-19
Fix looking up un-required trait in Targetable ctor

obrakmann:
Fixes #11055.

PR #11162, opened 2016-04-21
Enable Targetable to handle multiple Cloak traits

### LINKED ISSUE #11159: Actors can't have multiple Cloakable traits
Looks like #11154 was poorly tested by us reviewers.

While it fixed #11055, now actors like the TS stealth tank crash the game because they have more than one cloakable trait.

The fix for that would probably be to change

`cloak = self.TraitOrDefault<Cloak>();`

to

`cloak = self.TraitsImplementing<Cloak>().FirstOrDefault(Exts.IsTraitEnabled);`

reaperrr:
Closes #11159.

Make sure to not just review the code but also test this, including TS.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Targetable.cs
+++ b/OpenRA.Mods.Common/Traits/Targetable.cs
@@ -10,6 +10,7 @@
 #endregion
 
 using System.Collections.Generic;
+using System.Linq;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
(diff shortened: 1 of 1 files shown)


obrakmann:
I'm not really comfortable with taking this change for stable. Things like this have bitten us before and caused subtle, but annoying bugs.

Phrohdoh:
> Make sure to not just review the code but also test this, including TS.

Is this not always the case?

OpenRA.Mods.Common/Traits/Targetable.cs
```diff
@@ -29,24 +30,23 @@ public class TargetableInfo : UpgradableTraitInfo, ITargetableInfo
 	public class Targetable : UpgradableTrait<TargetableInfo>, ITargetable, INotifyCreated
 	{
 		protected static readonly string[] None = new string[] { };
-		protected Cloak cloak;
+		protected Cloak[] cloak;
```
Phrohdoh: This variable should be `cloaks` now, or something similar.

reaperrr:
> Is this not always the case?

Obviously not, otherwise #11159 should have been caught while reviewing #11154 (when testing TS stealth tank).

> I'm not really comfortable with taking this change for stable. Things like this have bitten us before and caused subtle, but annoying bugs.

@obrakmann I see your point. Personally I don't mind either way, some 3rd-party modders might complain when they run into #11159, though.

GraionDilach:
I did tested TD and RA but only those because it was for stable and I forgot D2k has the Stealth Trike fyi.

OpenRA.Mods.Common/Traits/Targetable.cs
```diff
@@ -29,24 +30,23 @@ public class TargetableInfo : UpgradableTraitInfo, ITargetableInfo
...
 		{
-			cloak = self.TraitOrDefault<Cloak>();
+			cloak = self.TraitsImplementing<Cloak>().ToArray();
 		}
 
 		public virtual bool TargetableBy(Actor self, Actor viewer)
 		{
 			if (IsTraitDisabled)
 				return false;
-			if (cloak == null || (!viewer.IsDead && viewer.Info.HasTraitInfo<IgnoresCloakInfo>()))
```
pchote: For safety (.IsVisible may do other trait lookups) I would keep this check here, but replace the null check with the .Any() check.

pchote:
With my requested change this will have identical behaviour in the common case, and the case where things may regress changes from a crash to *probably* works.  That's good enough for stable by me.

reaperrr:
Updated.

RoosterDragon:
:+1: Seems fine and reasonable for stable.

obrakmann:
This looks ok to me now. The code is pretty much equivalent to the previous version, and I didn't notice any regression. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/a138632931e65944459e3671297dec413e6a6d07)

PR #11165, opened 2016-04-23
Fix missile facing check.

### LINKED ISSUE #10305: Missiles often crash prematurely
When shooting north (facing 0), missiles quite often make a sharp curve to the right and explode, sometimes several cells before the target. Other directions don't seem to be affected.

Introduced by #8717 or one of its follow-ups (#9646, #9746).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29424547-missiles-often-crash-prematurely-when-flying-north?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #10305, which happens when the desired / current horizontal facing jitters between 255 and 0.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Missile.cs
+++ b/OpenRA.Mods.Common/Effects/Missile.cs
@@ -9,6 +9,7 @@
  */
 #endregion
 
+using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
@@ -701,9 +702,9 @@ namespace OpenRA.Mods.Common.Effects
 			var velVec = tarDistVec + predVel;
 			var desiredHFacing = velVec.HorizontalLengthSquared != 0 ? velVec.Yaw.Facing : hFacing;
 
-			if (allowPassBy && System.Math.Abs(desiredHFacing - hFacing) >= System.Math.Abs(desiredHFacing + 128 - hFacing))
+			if (allowPassBy && Math.Abs((desiredHFacing - hFacing) & 0xFF) >= Math.Abs((desiredHFacing + 128 - hFacing) & 0xFF))
 			{
-				desiredHFacing += 128;
+				desiredHFacing = (desiredHFacing + 128) & 0xFF;
 				targetPassedBy = true;
 			}
 			else


obrakmann:
Confirmed, :+1:

OpenRA.Mods.Common/Effects/Missile.cs
```diff
@@ -701,7 +702,7 @@ WVec HomingTick(World world, WVec tarDistVec, int relTarHorDist)
 			var velVec = tarDistVec + predVel;
 			var desiredHFacing = velVec.HorizontalLengthSquared != 0 ? velVec.Yaw.Facing : hFacing;
 
-			if (allowPassBy && System.Math.Abs(desiredHFacing - hFacing) >= System.Math.Abs(desiredHFacing + 128 - hFacing))
+			if (allowPassBy && Math.Abs((desiredHFacing - hFacing) & 0xFF) >= Math.Abs((desiredHFacing + 128 - hFacing) & 0xFF))
 			{
 				desiredHFacing += 128;
```
obrakmann: Although, wouldn't this need a similar fix?

pchote:
Updated.

reaperrr:
@pchote After letting two bazookas shoot north for 2 minutes, I can say it's not 100% fixed, but happens an order of magnitude less often.

What I noticed was that in the common case you fixed, the missile would sharply turn right and blow up which no longer happens, while in those rare remaining cases they sharply turn left instead.

In case you have any idea how to fix it entirely I'm not merging this yet, otherwise I think this is good enough for stable, so :+1:

pchote:
I suspect that there are other similar wraparound issues elsewhere in the code.  I really don't want to have to parse and potentially rewrite most of that logic (at the very least fixing the variable names as part of understanding what it does) – that is best left to the planned rewrite.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/3b41864812c3b8711c1e1e510ab86ad37caffa1e...c86633c858744a50e29f9da06f98a7ad35321265)

reaperrr:
Unfortunately this regressed things rather badly for other directions. Watching the RA shellmap for a while makes that fairly obvious.

PR #11169, opened 2016-04-23
Set individual shapes and shape offsets for non-building TS actors

reaperrr:
Split from #10383 to make it less intimidating and serves as its test case.

This PR really only consists of the last 6 commits.


reaperrr:
I guess people won't review this until the dependencies are merged, so I'm closing this for now.

reaperrr:
Removing from Future milestone since this is now only a matter of time.

PR #7361, opened 2015-01-19
Try to fix RA2

abcdefg30:
Worst PR ever.

PR #11177, opened 2016-04-24
Make SupportPowerChargeBar and the SupportPowerTimer customizable with stances

abcdefg30:
~~As a workaround for #11048.~~

~~If we show the timer, we now also show the charge bar (both is customizable with a stances field now).~~

~~(If this is merged before the next release, I'd like to get it into `stable`. It's not compulsory though.)~~

EDIT:

> This now only adds support to switch the charge bars and timers on/off based on stances.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
+++ b/OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
@@ -26,22 +26,22 @@ namespace OpenRA.Mods.Common.Traits.Render
 		public object Create(ActorInitializer init) { return new SupportPowerChargeBar(init.Self, this); }
 	}
 
-	class SupportPowerChargeBar : ISelectionBar
+	class SupportPowerChargeBar : ISelectionBar, INotifyOwnerChanged
 	{
 		readonly Actor self;
 		readonly SupportPowerChargeBarInfo info;
+		SupportPowerManager spm;
 
 		public SupportPowerChargeBar(Actor self, SupportPowerChargeBarInfo info)
 		{
 			this.self = self;
 			this.info = info;
+			spm = self.Owner.PlayerActor.Trait<SupportPowerManager>();
 		}
 
 		float ISelectionBar.GetValue()
 		{
-			var spm = self.Owner.PlayerActor.Trait<SupportPowerManager>();
 			var power = spm.GetPowersForActor(self).FirstOrDefault(sp => !sp.Disabled);
-
 			if (power == null)
 				return 0;
 
@@ -53,5 +53,10 @@ namespace OpenRA.Mods.Common.Traits.Render
 		}
 
 		Color ISelectionBar.GetColor() { return info.Color; }
+
+		void INotifyOwnerChanged.OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
+		{
+			spm = newOwner.PlayerActor.Trait<SupportPowerManager>();
+		}
 	}
 }
--- a/OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
+++ b/OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
@@ -15,9 +15,12 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits.Render
 {
-	[Desc("Display the time remaining until the super weapon attached to the actor is ready to the player and his allies.")]
+	[Desc("Display the time remaining until the super weapon attached to the actor is ready.")]
 	class SupportPowerChargeBarInfo : ITraitInfo
 	{
+		[Desc("Defines to which players the bar is to be shown.")]
+		public readonly Stance DisplayStances = Stance.Ally;
+
 		public readonly Color Color = Color.Magenta;
 
 		public object Create(ActorInitializer init) { return new SupportPowerChargeBar(init.Self, this); }
@@ -36,13 +39,15 @@ namespace OpenRA.Mods.Common.Traits.Render
 
 		float ISelectionBar.GetValue()
 		{
-			if (!self.Owner.IsAlliedWith(self.World.RenderPlayer))
-				return 0;
-
 			var spm = self.Owner.PlayerActor.Trait<SupportPowerManager>();
 			var power = spm.GetPowersForActor(self).FirstOrDefault(sp => !sp.Disabled);
 
-			if (power == null) return 0;
+			if (power == null)
+				return 0;
+
+			var viewer = self.World.RenderPlayer ?? self.World.LocalPlayer;
+			if (viewer != null && info.DisplayStances.HasStance(self.Owner.Stances[viewer]))
+				return 0;
 
 			return 1 - (float)power.RemainingTime / power.TotalTime;
 		}
--- a/OpenRA.Mods.Common/Traits/SupportPowers/SupportPower.cs
+++ b/OpenRA.Mods.Common/Traits/SupportPowers/SupportPower.cs
@@ -44,7 +44,8 @@ namespace OpenRA.Mods.Common.Traits
 		public readonly string IncomingSound = null;
 		public readonly string IncomingSpeechNotification = null;
 
-		public readonly bool DisplayTimer = false;
+		[Desc("Defines to which players the timer is shown.")]
+		public readonly Stance DisplayTimerStances = Stance.None;
 
 		[Desc("Palette used for the icon.")]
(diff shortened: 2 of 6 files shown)


GraionDilach:
>Checking for code style violations in OpenRA.Mods.Common...
OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs:L16: [SA1210] Using directives must be sorted alphabetically by the namespaces.

Worst Travis error ever?

OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
```diff
@@ -36,13 +37,20 @@ public SupportPowerChargeBar(Actor self, SupportPowerChargeBarInfo info)
 
 		float ISelectionBar.GetValue()
 		{
-			if (!self.Owner.IsAlliedWith(self.World.RenderPlayer))
-				return 0;
-
 			var spm = self.Owner.PlayerActor.Trait<SupportPowerManager>();
```
reaperrr: Out of curiosity, is there a reason why this can't be cached in ctor/on creation?

abcdefg30: Saw that too, but didn't want to touch that in this pr. Can do it ofc and will probably,
now that someone explicitly asked.

reaperrr: Ok. If you change your mind, I could add it to #11179.

abcdefg30:
> Worst Travis error ever?

Well yeah, especially as `atom` refused to let me interactively rebase...
Anyway, not sure if that can compare to #7361. :p

Also added one commit to cache the `SupportPowerManager` trait in `SupportPowerChargeBar.cs`.

OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
obrakmann: #11179 has already been merged, so please change it here. Don't forget to implement `INotifyOwnerChanged`.

OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
```diff
@@ -27,22 +28,29 @@ class SupportPowerChargeBar : ISelectionBar
 	{
 		readonly Actor self;
 		readonly SupportPowerChargeBarInfo info;
+		readonly SupportPowerManager spm;
 
 		public SupportPowerChargeBar(Actor self, SupportPowerChargeBarInfo info)
 		{
 			this.self = self;
 			this.info = info;
+			spm = self.Owner.PlayerActor.Trait<SupportPowerManager>();
```
obrakmann: You'll need to update this when the owner changes.

obrakmann:
:+1: besides that.

OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
abcdefg30: Ops, fixed. Thanks.

abcdefg30:
Updated.

OpenRA.Mods.Common/Traits/SupportPowers/SupportPower.cs
```diff
@@ -40,6 +40,9 @@ public abstract class SupportPowerInfo : UpgradableTraitInfo
 
 		public readonly bool DisplayTimer = false;
 
+		[Desc("Defines to which players the timer is shown when 'DisplayTimer' is true.")]
+		public readonly Stance DisplayTimerStances = Stance.Ally | Stance.Neutral | Stance.Enemy;
```
pchote: This makes `DisplayTimer` redundant, so please remove that field and default this to none.

OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
```diff
@@ -15,38 +15,51 @@
...
-			var spm = self.Owner.PlayerActor.Trait<SupportPowerManager>();
 			var power = spm.GetPowersForActor(self).FirstOrDefault(sp => !sp.Disabled);
+			if (power == null)
+				return 0;
 
-			if (power == null) return 0;
+			if (power.Info.DisplayTimer)
+			{
+				var viewer = self.World.RenderPlayer ?? self.World.LocalPlayer;
+				if (viewer != null && !power.Info.DisplayTimerStances.HasStance(self.Owner.Stances[viewer]))
```
pchote: There are cases where mods might want to show the bar but not the timer, so it would be better to have a separate `DisplayBarStances` field for this.

abcdefg30:
Hm, I noticed that this would break fake nukes (they don't have a timer) and will need extra logic to fix that.

pchote:
Lets drop the [for stable] and think about this in more detail for the next release.

GraionDilach:
Couldn't we add a completely fake Nuke - it falls down, but delivers no damage - as a solution?

OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
```diff
@@ -15,38 +15,51 @@
...
+					return 0;
+			}
+			else if (!self.Owner.IsAlliedWith(self.World.RenderPlayer))
+				return 0;
```
Phrohdoh: What is the point of showing the bar if you're going to lie about the value?

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -49,7 +50,14 @@ public SupportPowerTimerWidget(World world)
 
 		public override void Tick()
 		{
-			texts = powers.Select(p =>
+			var displayedPowers = powers.Where(p =>
+			{
+				var owner = p.Instances[0].Self.Owner;
```
Phrohdoh: I'm not familiar with this code, is at least one instance guaranteed?

OpenRA.Mods.Common/Traits/Render/SupportPowerChargeBar.cs
abcdefg30: Why are we lying?

Anyway, this code is going to change, just didn't push it yet.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
abcdefg30: https://github.com/OpenRA/OpenRA/pull/11177/files#diff-f65c5dc3568e124821d5e68818c9c218R64 it should be.

abcdefg30:
Updated. This now only adds support to switch the charge bars and timers on/off based on stances.

GraionDilach:
Needs a rebase and the upgrade rule be applied to the TS mod as well. Otherwise :+1:

abcdefg30:
Rebased, bumped the upgrade rule date and also run the rule on TS.

GraionDilach:
My :+1: still stands.

Mailaender:
Needs a rebase.

abcdefg30:
Rebased.

OpenRA.Mods.Common/Widgets/SupportPowerTimerWidget.cs
```diff
@@ -34,7 +35,7 @@ public SupportPowerTimerWidget(World world)
 			powers = world.ActorsWithTrait<SupportPowerManager>()
 				.Where(p => !p.Actor.IsDead && !p.Actor.Owner.NonCombatant)
 				.SelectMany(s => s.Trait.Powers.Values)
-				.Where(p => p.Instances.Any() && p.Info.DisplayTimer && !p.Disabled);
+				.Where(p => p.Instances.Any() && p.Info.DisplayTimerStances > 0 && !p.Disabled);
```
RoosterDragon: Maybe ` != Stance.None` rather than `> 0`.

RoosterDragon:
:+1:

GraionDilach:
Also needs the upgrade rule date bumped again. :+1: after it.

reaperrr:
:+1: from me too, once the nits pointed out by @RoosterDragon and @GraionDilach have been adressed.

abcdefg30:
Rebased and updated.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a881944ff7afef33ac243c2d4d1f8377f079e840...cc629c8bf99e9c53b21a00e5c537eac10207fb17)

PR #11179, opened 2016-04-24
Minor trait caching tweaks

reaperrr:
I couldn't find any noteworthy low-hanging fruits, though the `ResourceLayer` caching in `PoisonedByTiberium` might be worthwhile with lots of actors on the map.

### DIFF
--- a/OpenRA.Mods.Cnc/Traits/PoisonedByTiberium.cs
+++ b/OpenRA.Mods.Cnc/Traits/PoisonedByTiberium.cs
@@ -24,16 +24,20 @@ namespace OpenRA.Mods.Cnc.Traits
 
 		public WeaponInfo WeaponInfo { get; private set; }
 
-		public override object Create(ActorInitializer init) { return new PoisonedByTiberium(this); }
+		public override object Create(ActorInitializer init) { return new PoisonedByTiberium(init, this); }
 		public void RulesetLoaded(Ruleset rules, ActorInfo ai) { WeaponInfo = rules.Weapons[Weapon.ToLowerInvariant()]; }
 	}
 
 	class PoisonedByTiberium : UpgradableTrait<PoisonedByTiberiumInfo>, ITick, ISync
 	{
+		readonly ResourceLayer rl;
 		[Sync] int poisonTicks;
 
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.Common/Traits/Repairable.cs
```diff
@@ -33,12 +33,14 @@ class Repairable : IIssueOrder, IResolveOrder, IOrderVoice
...
 		readonly RepairableInfo info;
 		readonly Health health;
+		readonly IMove movement;
 		readonly AmmoPool[] ammoPools;
 
 		public Repairable(Actor self, RepairableInfo info)
 		{
 			this.info = info;
 			health = self.Trait<Health>();
+			movement = self.Trait<IMove>();
```
pchote: `Requires<IMoveInfo>` or `INotifyCreated` to avoid ordering issues.

OpenRA.Mods.Cnc/Traits/PoisonedByTiberium.cs
```diff
@@ -28,13 +28,19 @@ class PoisonedByTiberiumInfo : UpgradableTraitInfo, IRulesetLoaded
...
 	{
+		ResourceLayer rl;
 		[Sync] int poisonTicks;
 
 		public PoisonedByTiberium(PoisonedByTiberiumInfo info)
 			: base(info) { }
 
+		public void Created(Actor self)
+		{
+			rl = self.World.WorldActor.Trait<ResourceLayer>();
```
pchote: This is referencing the world actor, so you could do this in the ctor and make it readonly.

OpenRA.Mods.Common/Traits/Repairable.cs
reaperrr: I'll go with `Requires<IMoveInfo>` then, without a trait implementing `IMove` it would later crash anyway.

reaperrr:
Updated.

abcdefg30:

obrakmann:
:+1:

chrisforbes:
Perf work should come with perf numbers. :+1: though.

Issue #10879, opened 2016-03-05
Crash after removing earphones


sprice:
OpenRA crashed immediately after I removed the earphones from my Macbook's earphone jack.

OSX 10.11.3

exception.log
```
Red Alert Mod at Version release-20150919
on map a55c86fa5b11a125a9b0f800aafc76c30bc64893 (Alaska Anarchy Redux by Buddha).
Operating System: OSX (Unix 15.0.0.0)
Runtime Version: Mono 4.0.1 ((detached/ed1d3ec) CLR 4.0.30319.17020
Exception of type `System.InvalidOperationException`: Out of sync in frame 670.
 Compare syncreport.log with other players.
  at OpenRA.Network.OrderManager.OutOfSync (Int32 frame) [0x00000] in <filename unknown>:0 
  at OpenRA.Network.OrderManager.CheckSync (System.Byte[] packet) [0x00000] in <filename unknown>:0 
  at OpenRA.Network.OrderManager+<TickImmediate>c__AnonStorey0.<>m__0 (Int32 clientId, System.Byte[] packet) [0x00000] in <filename unknown>:0 
  at OpenRA.Network.ReplayRecorderConnection+<Receive>c__AnonStorey0.<>m__0 (Int32 client, System.Byte[] data) [0x00000] in <filename unknown>:0 
  at OpenRA.Network.EchoConnection.Receive (System.Action`2 packetFn) [0x00000] in <filename unknown>:0 
  at OpenRA.Network.ReplayRecorderConnection.Receive (System.Action`2 packetFn) [0x00000] in <filename unknown>:0 
  at OpenRA.Network.OrderManager.TickImmediate () [0x00000] in <filename unknown>:0 
  at OpenRA.Sync+<CheckSyncUnchanged>c__AnonStorey0.<>m__0 () [0x00000] in <filename unknown>:0 
  at OpenRA.Sync.CheckSyncUnchanged[Boolean] (OpenRA.World world, System.Func`1 fn) [0x00000] in <filename unknown>:0 
  at OpenRA.Sync.CheckSyncUnchanged (OpenRA.World world, System.Action fn) [0x00000] in <filename unknown>:0 
  at OpenRA.Game.InnerLogicTick (OpenRA.Network.OrderManager orderManager) [0x00000] in <filename unknown>:0 
  at OpenRA.Game.LogicTick () [0x00000] in <filename unknown>:0 
[...]

PR #11183, opened 2016-04-24
Fixed a NullReferenceException in OpenAL QueryDevices

### LINKED ISSUE #11181: The game crashes when open "settings" from main menu [Linux, playtest-20160424]
Hi! When I'm clicking on "settings" button in the main game menu, the game immediately crashes with System.NullReferenceException . I believe it's a sound related problem if judge by the game output. Please find log files in the attachment. I tested it only on my Linux machine.

[exception.txt](https://github.com/OpenRA/OpenRA/files/233581/exception.txt)
[sound.txt](https://github.com/OpenRA/OpenRA/files/233582/sound.txt)

Mailaender:
According to https://msdn.microsoft.com/en-us/library/7b620dhe(v=vs.110).aspx Marshal.PtrToStringAnsi can return null so I assume this fixes https://github.com/OpenRA/OpenRA/issues/11181.

### DIFF
--- a/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
+++ b/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
@@ -64,7 +64,7 @@ namespace OpenRA.Platforms.Default
 
 			do
 			{
-				var str = Marshal.PtrToStringAnsi(next);
+				var str = Marshal.PtrToStringAuto(next);
 				next += str.Length + 1;
 				devices.Add(str);
 			} while (Marshal.ReadByte(next) != 0);
--- a/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
+++ b/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
@@ -13,6 +13,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text;
 using OpenAL;
 
 namespace OpenRA.Platforms.Default
@@ -65,7 +66,7 @@ namespace OpenRA.Platforms.Default
 			do
 			{
 				var str = Marshal.PtrToStringAuto(next);
-				next += str.Length + 1;
+				next += UnicodeEncoding.Default.GetByteCount(str) + 1;
 				devices.Add(str);
 			} while (Marshal.ReadByte(next) != 0);
 
--- a/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
+++ b/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
@@ -56,18 +56,18 @@ namespace OpenRA.Platforms.Default
 
 			var devices = new List<string>();
 			var next = ALC10.alcGetString(IntPtr.Zero, type);
+			if (next == IntPtr.Zero || AL10.alGetError() != AL10.AL_NO_ERROR)
+			{
+				Log.Write("sound", "Failed to query OpenAL device list using {0}", label);
+				return new string[] { };
+			}
+
 			do
 			{
 				var str = Marshal.PtrToStringAnsi(next);
 				next += str.Length + 1;
 				devices.Add(str);
 			} while (Marshal.ReadByte(next) != 0);
-
-			if (AL10.alGetError() != AL10.AL_NO_ERROR)
-			{
-				Log.Write("sound", "Failed to query OpenAL device list using {0}", label);
-				return new string[] { };
-			}
 
 			return devices.ToArray();
 		}


obrakmann:
I think this would result in an infinite loop since the `next` pointer wouldn't advance.

I haven't found any documentation on what `ALC10.alcGetString()` might return, but I guess the `next` pointer might also end up being null, in which case the following would fix it:

```diff
diff --git a/OpenRA.Platforms.Default/OpenAlSoundEngine.cs b/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
index 65f56f9..878be70 100644
--- a/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
+++ b/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
@@ -56,14 +56,17 @@ static string[] QueryDevices(string label, int type)
 
                        var devices = new List<string>();
                        var next = ALC10.alcGetString(IntPtr.Zero, type);
-                       do
+                       if (next != IntPtr.Zero)
                        {
-                               var str = Marshal.PtrToStringAnsi(next);
-                               next += str.Length + 1;
-                               devices.Add(str);
-                       } while (Marshal.ReadByte(next) != 0);
+                               do
+                               {
+                                       var str = Marshal.PtrToStringAnsi(next);
+                                       next += str.Length + 1;
+                                       devices.Add(str);
+                               } while (Marshal.ReadByte(next) != 0);
+                       }
 
-                       if (AL10.alGetError() != AL10.AL_NO_ERROR)
+                       if (next == IntPtr.Zero || AL10.alGetError() != AL10.AL_NO_ERROR)
                        {
                                Log.Write("sound", "Failed to query OpenAL device list using {0}", label);
                                return new string[] { };

```

Mailaender:
You are right. This can and should be catched earlier. I wasn't sure if checking points for null is valid and possible, but it is. Updated accordingly.

LipkeGu:
@Mailaender  this could also fix #10879 :)

Mailaender:
I don't think so. This is just related to setting up the dropdown box in the audio settings. I filed https://github.com/OpenRA/OpenRA/pull/11185 as the "ALC Device not closed" message is still there when exiting by closing the game window with `x` buttons.

OpenRA.Platforms.Default/OpenAlSoundEngine.cs
```diff
@@ -56,14 +56,17 @@ static string[] QueryDevices(string label, int type)
...
+				do
+				{
+					var str = Marshal.PtrToStringAnsi(next);
+					next += str.Length + 1;
+					devices.Add(str);
+				} while (Marshal.ReadByte(next) != 0);
+			}
 
-			if (AL10.alGetError() != AL10.AL_NO_ERROR)
+			if (next == IntPtr.Zero || AL10.alGetError() != AL10.AL_NO_ERROR)
```
pchote: It looks like you could move this above the dowhile and avoid the new IntPtr.Zero check.

Mailaender: I think you are right. Updated.

RoosterDragon:
:+1: Looks good to me.

OpenRA.Platforms.Default/OpenAlSoundEngine.cs
```diff
@@ -56,19 +56,25 @@ static string[] QueryDevices(string label, int type)
...
+			{
+				Log.Write("sound", "Failed to query OpenAL device list using {0}", label);
+				return new string[] { };
+			}
+
 			do
 			{
 				var str = Marshal.PtrToStringAnsi(next);
+				if (string.IsNullOrEmpty(str))
```
pchote: This condition will never be true.  As mentioned earlier `PtrToStringAnsi` will only return null if `next` is null, and that case is covered by the check above.  `str` can also never be an empty string, because we are explicitly checking for that as our loop-termination condition.

Mailaender: Have you read https://github.com/OpenRA/OpenRA/issues/11181#issuecomment-217708570?

pchote: I have, yes.

pchote: referencesource.microsoft.com/#mscorlib/system/runtime/interopservices/marshal.cs,124 confirms the behaviour of `PtrToStringAnsi`.  Recent versions of mono use the same source.

Mailaender: I also know about that, which is why my first attempt was to just catch there, but apparently it isn't enough. What changes do you request in particular?

pchote: Identifying the steps that cause `str` to be null here, and then adding the check / workaround at the source of the problem.

Mailaender: I will most likely not be able to achieve that.

Mailaender:
Updated last week.

KevinYeti:
@Mailaender Hi, I'm new here and I'm a fans of RA from China. Maybe I can do the translation work of Chinese.  
But now the code seems not working on my laptop.The OS is Fedora 23.
After try to debug the code, I found if my language is Chinese, the code will not work. 
> next += str.Length + 1;

I've change the code to this on my laptop.

> next += System.Text.UnicodeEncoding.Default.GetByteCount(str) + 1; 

This code will work. And of course the characters of Chinese is not right showed. Maybe we should do the multi-language support at a higher priority level ?

Mailaender:
Fixed the byte count for Chinese.

Mailaender:
Confirmed that it fixes the issue at https://github.com/OpenRA/OpenRA/issues/11181#issuecomment-220829433 :white_check_mark:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/c100fc8cc031620ca4301d530b22e478970ed13d)

PR #9395, opened 2015-09-20
Fixed the OpenAL device not being properly closed

Mailaender:
I always get reminded in the command line when closing the game.

> AL lib: (EE) alc_cleanup: 1 device not closed

PR #9394, opened 2015-09-20
Unstatic the Sound class.

pchote:
Followup to #9274.  This unstatics the `Sound` class, and moves it to a static field in Game to make it consistent with how we access the renderer.

PR #11185, opened 2016-04-25
Fixed the OpenAL device not being properly closed

Mailaender:
Revival of https://github.com/OpenRA/OpenRA/pull/9395 as https://github.com/OpenRA/OpenRA/pull/9394 did not solve it completely.

### DIFF
--- a/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
+++ b/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
@@ -102,6 +102,8 @@ namespace OpenRA.Platforms.Default
 					throw new InvalidOperationException("Can't create OpenAL device");
 			}
 
+			Game.OnQuit += () => { ALC10.alcCloseDevice(device); };
+
 			var ctx = ALC10.alcCreateContext(device, null);
 			if (ctx == IntPtr.Zero)
 				throw new InvalidOperationException("Can't create OpenAL context");


OpenRA.Platforms.Default/OpenAlSoundEngine.cs
```diff
@@ -102,6 +102,8 @@ public OpenAlSoundEngine()
 					throw new InvalidOperationException("Can't create OpenAL device");
 			}
 
+			Game.OnQuit += () => { ALC10.alcCloseDevice(device); };
```
pchote: This is already done when disposing the sound engine.  If the sound engine isn't being disposed then that needs to be fixed – don't just force close it again!

Mailaender: Shall I change this to Dispose OnQuit?

pchote: This shouldn't be using OnQuit at all. That's a legacy hack from before we started treating object lifetimes properly.

Mailaender:
Closing as this is not wanted and no constructive alternatives are provided. The bug https://github.com/OpenRA/OpenRA/issues/11193 persists.

pchote:
I gave the alternative in my first comment: you need to identify the code path where the game quits without disposing the SoundDevice, and then add the code to dispose it.

Mailaender:
No, you said:

> This is already done when disposing the sound engine.

LipkeGu:
> If the sound engine isn't being disposed then that needs to be fixed – don't just force close it again!

Mailaender:
Probably a misunderstanding, but either way I don't know why the game doesn't dispose properly when the window manager quits the game or ALT + F4 is hit.

LipkeGu:
maybe it isnt a OpenRA related thing... mayb it is SDL2 related... we should forward this to the SDL devs :/

LipkeGu:
i used google.... it is OpenAL related....
> 

The AL lib is the "audio library" used by Libgdx (one of the OpenAL variants).

I believe this message just means that the audio library is cleaning up some (in your case just one) audio streams/handles for you. If you see this at exit, its harmless as all resources will be cleaned up by the OS.

If you internally cleanup your audio before exiting, the message should go away.

For more details, look for alc_cleanup in here: http://repo.or.cz/w/openal-soft.git/blob/HEAD:/Alc/ALc.c


and also this info: AL.destroy(); must be called before the application shuts down. this call is normally included in a finally block, but if the application calls System.exit(0), finally blocks are no longer executed.

PR #11192, opened 2016-04-26
Fix direct hit check in CreateEffectWarhead

### LINKED ISSUE #11184: Fix CreateEffectWarhead's GetDirectHit
Currently, it sometimes doesn't consider impacts within a `HitShape` as direct hits (particularly in TS).

One problem might be that `GetDirectHit` only checks for actors on the impact cell, but the target's `HitShape` might very well have been touched on a cell adjacent to the cell housing the actor's `CenterPosition`, so we should extend this to a search radius similar to what `SpreadDamageWarhead` has.

Additionally,
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs#L85
looks bogus to me. If I read that line correctly, when `DistanceFromEdge` is 0 (= inside `HitShape`), impact distance to target `CenterPosition` must be 0 as well to return `true`. But unless the projectile hits the target smack in the center, the target's `CenterPosition` will obviously be at least a bit further away than the `HitShape`, so this will almost never be true for slightly inaccurate weapons like missiles.

reaperrr:
Closes #11184.

Originally, this was comparing distance beween `pos` and `unit.CenterPosition` with `HitShape`'s `OuterRadius`. However, the `OuterRadius` can exceed the shape for `Capsule` and `Rectangular` shapes, so I tried to adress that a few months ago by using the `DistanceFromEdge` check instead. The approach was bogus, though. `DistanceFromEdge` just calculates the distance of a position to the edge, so by comparing it with the distance between `pos` and `victim.CenterPosition` in combination with using `LengthSquared`, it was entirely possible the explosion would be within the `HitShape`, but closer to the edge than the `victim.CenterPosition` and the check would return `false`, despite the explosion happening inside the shape.

Now we simply check if `DistanceFromEdge` is `0` or negative.

### DIFF
--- a/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
@@ -30,6 +30,9 @@ namespace OpenRA.Mods.Common.Warheads
 		[Desc("Remap explosion effect to player color, if art supports it.")]
 		public readonly bool UsePlayerPalette = false;
 
+		[Desc("Search radius around impact for 'direct hit' check.")]
+		public readonly WDist TargetSearchRadius = new WDist(2048);
+
 		[Desc("List of sounds that can be played on impact.")]
 		public readonly string[] ImpactSounds = new string[0];
 
@@ -72,7 +75,7 @@ namespace OpenRA.Mods.Common.Warheads
 
 		public bool GetDirectHit(World world, CPos cell, WPos pos, Actor firedBy, bool checkTargetType = false)
 		{
-			foreach (var unit in world.ActorMap.GetActorsAt(cell))
+			foreach (var unit in world.FindActorsInCircle(pos, TargetSearchRadius))
 			{
 				if (checkTargetType && !IsValidAgainst(unit, firedBy))
 					continue;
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Warheads/CreateEffectWarhead.cs
```diff
@@ -30,6 +30,9 @@ public class CreateEffectWarhead : Warhead
 		[Desc("Remap explosion effect to player color, if art supports it.")]
 		public readonly bool UsePlayerPalette = false;
 
+		[Desc("Search radius around impact for 'direct hit' check.")]
+		public readonly WDist TargetSearchRadius = new WDist(3072);
```
pchote: IIRC we have a lint check for testing these radii.  You'll need to add this there.

reaperrr:
Updated.

GraionDilach:
Looks good to me. :+1:

reaperrr:
Actually, the bug this fixes is even visible in the release (RA Destroyer's Stinger missiles always display water splash when hitting a water target), so I'm promoting this to Next, just in case.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/8f3546613d13955750a52359ecb9c124309d448c)

PR #11194, opened 2016-04-26
Fix ProvidesPrerequisite not refreshing the tech tree (on owner change)

abcdefg30:
Not sure if this is the correct approach. Context (see also https://github.com/OpenRA/ra2/pull/191):

> 21:44:08 (abcdefg30) ok, so I might need some advice here
> 21:44:22 (abcdefg30) it works how I want it too, but not sure if that's the right thing to do
> 21:44:38 (abcdefg30) case: we have a neutral airport, when you capture it you get faction specific paradrops
> 21:45:03 (abcdefg30) unfortunately, we need different prerequisites for the different faction specific paradrops
> 21:45:15 (abcdefg30) so I just let the building itself provide a faction specific prereq
> 21:45:42 (abcdefg30) now the thing: the building just grants the prereq of the first owner (which would be neutral with its faction)
> 21:46:17 (abcdefg30) when I capture the neutral building then, I still somehow get the old factions prereq / paratroopers power :s

I'll investigate a bit more tomorrow, if noone has a good idea.

### DIFF
--- a/OpenRA.Mods.Common/Traits/ProximityCapturable.cs
+++ b/OpenRA.Mods.Common/Traits/ProximityCapturable.cs
@@ -189,7 +189,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		void INotifyOwnerChanged.OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
 		{
-			skipTriggerUpdate = false;
+			Game.RunAfterTick(() => skipTriggerUpdate = false);
 		}
 	}
 }
--- a/OpenRA.Game/Actor.cs
+++ b/OpenRA.Game/Actor.cs
@@ -301,11 +301,11 @@ namespace OpenRA
 				Owner = newOwner;
 				Generation++;
 
-				if (wasInWorld)
-					w.Add(this);
-
 				foreach (var t in TraitsImplementing<INotifyOwnerChanged>())
 					t.OnOwnerChanged(this, oldOwner, newOwner);
+
+				if (wasInWorld)
+					w.Add(this);
 			});
 		}
 


OpenRA.Mods.Common/Traits/Player/ProvidesPrerequisite.cs
```diff
@@ -46,6 +47,8 @@ public ProvidesPrerequisite(ActorInitializer init, ProvidesPrerequisiteInfo info
 			if (string.IsNullOrEmpty(prerequisite))
 				prerequisite = init.Self.Info.Name;
 
+			techTree = init.Self.Owner.PlayerActor.Trait<TechTree>();
```
pchote: This is used at most once, so there is no benefit in querying and caching this here. Look it up when you need it in the capture callback.

OpenRA.Mods.Common/Traits/Player/ProvidesPrerequisite.cs
```diff
@@ -64,8 +67,15 @@ public IEnumerable<string> ProvidesPrerequisites
...
-				Update(newOwner, newOwner.Faction.InternalName);
+			if (!info.ResetOnOwnerChange)
+				return;
+
+			// The TechTree of the oldOwner gets updated in the Update method
+			Update(newOwner, newOwner.Faction.InternalName);
+
+			// Now refresh the TechTree of the new owner
+			techTree = newOwner.PlayerActor.Trait<TechTree>();
+			techTree.Update();
```
pchote: These shouldnt be needed - the tech tree code should already be detecting the prereq change and updating itself. Find what is broken there and dlfix that instead.

OpenRA.Mods.Common/Traits/Player/ProvidesPrerequisite.cs
```diff
@@ -76,7 +86,9 @@ void Update(Player owner, string faction)
 				enabled = info.Factions.Contains(faction);
 
 			if (info.RequiresPrerequisites.Any() && enabled)
-				enabled = owner.PlayerActor.Trait<TechTree>().HasPrerequisites(info.RequiresPrerequisites);
+				enabled = techTree.HasPrerequisites(info.RequiresPrerequisites);
```
pchote: Querying the state of the tech tree beforeupdating the tech tree doesnt look right at all. I expect this is the real cause of your bug. If you drop everything else and run this at the end of tick then your problem might go away.

MustaphaTR:
I don't think this is the true way to go. It may break more things than it brings. I think there should be a new tag like `UpdateOnOwnerChange:` under `ProvidesPrerequisite:`.

abcdefg30:
> I don't think this is the true way to go. It may break more things than it brings. I think there should be a new tag like  UpdateOnOwnerChange:  under  ProvidesPrerequisite: .

We already have this and it's called `ResetOnOwnerChange`. The main problem here is probably that the `ParatroopersPower`s don't get "notified" about the now different prerequisites due to the different factions (same also happens for other things using prerequisites).

OpenRA.Mods.Common/Traits/Player/ProvidesPrerequisite.cs
abcdefg30: Actually, I don't even use `RequiresPrerequisites` in my testcase, so I doubt this will be the cause.

OpenRA.Mods.Common/Traits/Player/ProvidesPrerequisite.cs
abcdefg30: Ok, my fault. ^^ I think it even worked without this, but it seemed a bit suspicius to me when I only update one Tech Tree.

abcdefg30:
Updated.

OpenRA.Mods.Common/Traits/Player/ProvidesPrerequisite.cs
```diff
@@ -75,8 +75,11 @@ void Update(Player owner, string faction)
 			if (info.Factions.Any())
 				enabled = info.Factions.Contains(faction);
 
+			var techTree = owner.PlayerActor.Trait<TechTree>();
 			if (info.RequiresPrerequisites.Any() && enabled)
-				enabled = owner.PlayerActor.Trait<TechTree>().HasPrerequisites(info.RequiresPrerequisites);
+				enabled = techTree.HasPrerequisites(info.RequiresPrerequisites);
+
+			techTree.Update();
```
pchote: You may be able to save a bit of perf by only calling this if `enabled` actually changes.

abcdefg30:
Updated. Now I moved adding the captured actor back to world to after the `INotifyOwnerChanged` callbacks have run as suggested by @pchote.

OpenRA.Mods.Common/Traits/ProximityCapturable.cs
```diff
@@ -189,7 +189,7 @@ void ChangeOwnership(Actor self, Actor captor)
 
 		void INotifyOwnerChanged.OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
 		{
-			skipTriggerUpdate = false;
+			Game.RunAfterTick(() => skipTriggerUpdate = false);
```
abcdefg30: This now needed to prevent the `AddedToWorld` method from running here.

yaaaaa:
maybe...
If it captured by soviet engineer -> it give a soviet paradroopers
If it captured by allied engineer -> it give a allied paradroopers
???

abcdefg30:
> maybe...
 If it captured by soviet engineer -> it give a soviet paradroopers
 If it captured by allied engineer -> it give a allied paradroopers
 ???

RA2 doesn't differentiate between allied and soviet engineer. Anyway, going the way you suggested would need more code changes and Imho it's better to fix the existing logic than creating new.

pchote:
Have you looked over all of the `OnOwnerChanged` implementations to make sure that there aren't any other regressions hiding in them?

MustaphaTR:
> Have you looked over all of the OnOwnerChanged implementations to make sure that there aren't any other regressions hiding in them?

That is what i'm afraid of. For example RA1 WFac/Cyards. They are same building with side specific prerequisites. So capturing them from other side should give their Vehicles/Buildings, not our own. I think i need to check.

pchote:
That is unrelated to my question.  I suggest you read through the `ProvidesPrerequisite` so you can see how race-specific tech works.

MustaphaTR:
Whatever, this looks like working fine. I think i shouldn't mess with code changes.

abcdefg30:
> Have you looked over all of the `OnOwnerChanged` implementations to make sure that there aren't any other regressions hiding in them?

I should have covered all. (But you can never be sure, can you? Especially as I'm derping all the time. :p)

> What i understand from these sentences is it doesn't work as supposed, so fix it.

Yep, that's what we're doing here.
To summarize it again (for others that don't want to read through all comments):
- When capturing an actor two things that interest us are happening (in this order): a) The actor gets removed from world and then readded. Here the `TechTree` refreshes. b) The `INotifyOwnerChanged` callback runs.
- Now, when the callback runs, the prerequiste granted by `ProvidesPrerequisite` is updated. The `TechTree` is *not* refreshed again.
- To "fix" that, @pchote suggested to move adding the actor back to world to *after* the callback ran. So now the `TechTree` gets refreshed (during `AddedToWorld`) after `ProvidesPrerequisite` updated its prerequisite.

MustaphaTR:
Got it. I think we need changes over RA2 mod yamls too. This doesn't sound like fix it alone to me.

Also, can `Player:` actor provide prerequisite? If yes, i think using it instead of `structure.xxx` may also fix the issue. But this may be a hacky aproach. Just asking, current one is fine.

pchote:
Looks good to me :+1:

I went through the `INotifyOwnerChanged` implementations and didn't find any other cases that I expect to break with this change.

I did notice, however, that capturing a transport will rewrite Owner directly instead of using ChangeOwner (which will then break any traits that rely on the owner change callback).  That's a different bug for a different PR.

OpenRA.Mods.Common/Traits/ProximityCapturable.cs
obrakmann: Can you change the `self.Owner` in line 186 to `captor.Owner` while you're at it, please? The owner only changes after another frame end task queued by the actor itself, not during the one queued here, so the wrong player ends up getting the notification. Thanks!

obrakmann:
Looks ok, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/25f95c135d7ed9287b6452b6ff3c0c0c36235178)

PR #11196, opened 2016-04-27
Halven CloseEnough on Stinger.

### LINKED ISSUE #10186: Destroyer Max Range Damage Bug
Basically when the destroyers are at max range they do 0 or effectively 0 damage.  FiveAces has mentioned this a lot but I didn't see anything in issues other than this [erroneously?] closed issue.
https://github.com/OpenRA/OpenRA/issues/8476

GraionDilach:
With 170 Speed, 298 CloseEnough, and 128 Spread, it will always act as a proximity weapon against the targets. Other missiles aren't affected because all the other missiles have a fair margin to satisfy the optimal Speed+Spread > CloseEnough condition.

(With this change, there is a ~26% chance (1-(128/149)²) to deliver 37% damage and all the other cases full damage due to the missile characteristics and since the Speed is 170, a unit must move at 128 speed to outrun the missile and trigger it turning back to hit it).

Fixes #10186.

### DIFF
--- a/mods/ra/weapons/missiles.yaml
+++ b/mods/ra/weapons/missiles.yaml
@@ -295,6 +295,7 @@ Stinger:
 		HorizontalRateOfTurn: 20
 		RangeLimit: 50
 		Speed: 170
+		CloseEnough: 149
 	Warhead@1Dam: SpreadDamage
 		Spread: 128
 		Damage: 30
--- a/mods/ra/weapons/missiles.yaml
+++ b/mods/ra/weapons/missiles.yaml
@@ -293,7 +293,7 @@ Stinger:
 		ContrailLength: 10
 		Image: DRAGON
 		HorizontalRateOfTurn: 20
-		RangeLimit: 50
+		RangeLimit: 65
 		Speed: 170
 		CloseEnough: 149
 	Warhead@1Dam: SpreadDamage


obrakmann:
Your conclusion sounds plausible. Unfortunately, I can still reproduce this on this pr:
openra-2016-04-27t181247z

I assume @reaperrr's hitshape PR would eventually fix this.

GraionDilach:
No, another property was off as well. Thanks for the screenshot, RangeLimit was also too small (see the commit message).

obrakmann:
Ah, I see. Where does the 1.5 in your calculation come from, though? If you reverse the first calculation, I'd expect a minimum `RangeLimit` of 55 (9216 / 170 = 54.21). And indeed, setting it to 55 fixes it as well for me.

Setting it to 85 is a huge buff to the destroyer (14 cell range under the right circumstances!). How about 60?

GraionDilach:
Tend to play safe.

On a second thought, this doesn't really make a difference, since it uses a separate weapon against aircraft and among the ground vehicles, only the Ranger, the Phase Transport and the Flak Truck can be faster than 100 speed, or should those be capable of evading this missile?

obrakmann:
In my testing, it was firing at a harvester that was just inside the weapon range, but driving away. It still got hit when it was four or five cells away, which is exactly the situation where the increased `RangeLimit` makes a difference. The much longer flight time was clearly noticeable as well. (and that's me saying that, who never notices shit like that :p)

GraionDilach:
I'll go with 65 then. 67 would be 125% on the RangeLimit/Range and IMO the 20-25% runaway range is fair'n'square.

obrakmann:
:+1:

reaperrr:
For Next+1, we should investigate whether we even need CloseEnough anymore, originally it was just a hack to work around an inadequate shouldExplode check, but @pchote fixed that a while ago.

GraionDilach:
The current missile code needs CloseEnough from whatIi remember. Also, the TS Splits logic is just a combination of CloseEnough + RA2-style shrapnel IMO.

reaperrr:
OK. For this PR, :+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/4ad573f5615b1cdaebdd0d8c5a4c09bcb042798a...d6646f01b62ac897ca4852184fdd82c98cbad4f7)

PR #11197, opened 2016-04-27
Add sysinfo opt out prompt

pchote:
This displays an opt-out dialog the first time the game is run (after each change to the data collection):

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
@@ -12,16 +12,20 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Net;
+using OpenRA.Primitives;
 using OpenRA.Widgets;
 
 namespace OpenRA.Mods.Common.Widgets.Logic
 {
+	[SuppressMessage("StyleCop.CSharp.OrderingRules", "SA1203:ConstantsMustAppearBeforeFields",
+		Justification = "SystemInformation version should be defined next to the dictionary it refers to.")]
 	public class MainMenuLogic : ChromeLogic
 	{
-		protected enum MenuType { Main, Singleplayer, Extras, MapEditor, None }
+		protected enum MenuType { Main, Singleplayer, Extras, MapEditor, SystemInfoPrompt, None }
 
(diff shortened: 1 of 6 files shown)


obrakmann:
Looks ok to me, :+1: 

Will need the server-side stuff, though, before we release it into the wild.

abcdefg30:
The scrollbar still looks a bit weird to me, why do we have it if we can't scroll and don't use line breaks either (one line even extends the horizontal width)? Just minor visual polish though.

Phrohdoh:
On the topic of nits:
`s/.NET Runtime/Runtime/` is a bit more true to Mono and may cut down on the length of that line enough to prevent clipping.

pchote:
> Will need the server-side stuff, though, before we release it into the wild.

This is compatible with the current server-side stuff.  The changes I talked about in IRC are in addition to what it does now.

pchote:
Mono *is* a .NET runtime, so I don't think it should be changed.

OpenRA.Mods.Common/Widgets/Logic/MainMenuLogic.cs
```diff
@@ -32,6 +36,22 @@ protected enum MenuType { Main, Singleplayer, Extras, MapEditor, None }
...
+		const int SystemInformationVersion = 1;
+		Dictionary<string, Pair<string, string>> GetSystemInformation()
+		{
+			var lang = System.Globalization.CultureInfo.InstalledUICulture.TwoLetterISOLanguageName;
+			return new Dictionary<string, Pair<string, string>>()
+			{
+				{ "id", Pair.New("Anonymous ID", Game.Settings.Debug.UUID) },
+				{ "platform", Pair.New("OS Type", Platform.CurrentPlatform.ToString()) },
+				{ "os", Pair.New("OS Version", Environment.OSVersion.ToString()) },
+				{ "runtime", Pair.New(".NET Runtime", Platform.RuntimeVersion) },
```
Mailaender: `.NET Runtime: .NET CLR 4.0.30319.42000` will be redundant though.

pchote: The .NET Runtime used on windows is the .NET CLR 4.0.x.  There's no redundancy here.

obrakmann:
I already added the changelog entry to [the Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog)

reaperrr:
Looks good to me :+1:

PR #11199, opened 2016-04-29
Fix potential NRE in Parachutable

reaperrr:
Not sure how likely this is to occur, but it did happen to me on RA's shellmap while working on something else.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Parachutable.cs
+++ b/OpenRA.Mods.Common/Traits/Parachutable.cs
@@ -16,7 +16,7 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Can be paradropped by a ParaDrop actor.")]
-	class ParachutableInfo : ITraitInfo
+	class ParachutableInfo : ITraitInfo, Requires<IPositionableInfo>
 	{
 		[Desc("If we land on invalid terrain for my actor type should we be killed?")]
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Parachutable.cs
```diff
@@ -51,8 +51,11 @@ public Parachutable(ActorInitializer init, ParachutableInfo info)
 		{
 			self = init.Self;
 			this.info = info;
+		}
 
-			positionable = self.TraitOrDefault<IPositionable>();
+		void INotifyCreated.Created(Actor self)
+		{
+			positionable = self.Trait<IPositionable>();
```
Phrohdoh: AFAICT this isn't being required.

pchote:
`Requires<IPositionableInfo>` on the ParachutableInfo and assigning in the ctor would be better, because it doesn't make sense to parachute a non-positionable object.

reaperrr:
Updated.
Do we possibly want this in stable?

Mailaender:
:+1: Yeah, makes sense as this stabilizes the feature.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/6796a609c688f8cb77b1a66a9a87a0bf0bd3eb15)

PR #11205, opened 2016-04-30
Refactored Missile.RangeLimit to be a WDist value

reaperrr:
Tick-based `RangeLimit` is tedious to get right (you need to divide weapon Range by projectile speed, and that doesn't even cover any kind of acceleration), and many values throughout the shipping mods + TS are rough or even arbitrary guesses.

This refactor gives much more accurate control over maximum flight distance, especially in mods like TS which use acceleration.

All yaml updates were done automatically via the provided upgrade rules.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -826,6 +826,34 @@ namespace OpenRA.Mods.Common.UtilityCommands
 					}
 				}
 
+				// Refactor Missile RangeLimit from ticks to WDist
+				if (engineVersion < 20160501)
+				{
+					var weapRange = node.Value.Nodes.FirstOrDefault(n => n.Key == "Range");
+					var projectile = node.Value.Nodes.FirstOrDefault(n => n.Key == "Projectile");
+
+					if (projectile != null && weapRange != null && projectile.Value.Value == "Missile")
+					{
+						var oldWDist = FieldLoader.GetValue<WDist>("Range", weapRange.Value.Value);
+						var rangeLimitNode = projectile.Value.Nodes.FirstOrDefault(x => x.Key == "RangeLimit");
+
+						// RangeLimit is now a WDist value, so for the conversion, we take weapon range and add 20% on top.
+						// Overly complicated calculations using Range, Speed and the old RangeLimit value would be rather pointless,
+						// because currently most mods have somewhat arbitrary, usually too high and in a few cases too low RangeLimits anyway.
+						var newValue = oldWDist.Length * 120 / 100;
+						var newCells = newValue / 1024;
+						var newCellPart = newValue % 1024;
+
+						if (rangeLimitNode != null)
+							rangeLimitNode.Value.Value = newCells.ToString() + "c" + newCellPart.ToString();
+						else
+						{
+							// Since the old default was 'unlimited', we're using weapon range * 1.2 for missiles not defining a custom RangeLimit as well
+							projectile.Value.Nodes.Add(new MiniYamlNode("RangeLimit", newCells.ToString() + "c" + newCellPart.ToString()));
+						}
+					}
+				}
+
 				UpgradeWeaponRules(engineVersion, ref node.Value.Nodes, node, depth + 1);
 			}
 		}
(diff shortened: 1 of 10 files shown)


reaperrr:
Now that I think about it, a lint check that warns about RangeLimit < weapon range would probably be useful as well.

GraionDilach:
I really don't like changing the default value here - that just nerfs all missiles which would use that default value. Could the upgrade rule instead create RangeLimits for these missiles as well instead?

reaperrr:
> I really don't like changing the default value here - that just nerfs all missiles which would use that default value.

I wonder how many missiles use that default value, and out of those, how many use it intentionally?
In the shipping mods the answer is 'zero', in TS the only case is the `CyCannon`, and that certainly wasn't intentional.

But I'll update the upgrade rule as you suggested, just in case.

reaperrr:
Updated.
Travis passed, AppVeyor just failed to install something.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -826,6 +826,34 @@ internal static void UpgradeWeaponRules(int engineVersion, ref List<MiniYamlNode
...
+						// because currently most mods have somewhat arbitrary, usually too high and in a few cases too low RangeLimits anyway.
+						var newValue = oldWDist.Length * 120 / 100;
+						var newCells = newValue / 1024;
+						var newCellPart = newValue - (newCells * 1024);
+
+						if (rangeLimitNode != null)
+							rangeLimitNode.Value.Value = newCells.ToString() + "c" + newCellPart.ToString();
+						else
+						{
+							// Since the old default was 'unlimited', we're using weapon range * 1.5 for missiles not defining a custom RangeLimit
```
GraionDilach: * 1.2

reaperrr:
Updated.

GraionDilach:
Looks good to me. :+1:

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -826,6 +826,34 @@ internal static void UpgradeWeaponRules(int engineVersion, ref List<MiniYamlNode
...
+					{
+						var oldWDist = FieldLoader.GetValue<WDist>("Range", weapRange.Value.Value);
+						var rangeLimitNode = projectile.Value.Nodes.FirstOrDefault(x => x.Key == "RangeLimit");
+
+						// RangeLimit is now a WDist value, so for the conversion, we take weapon range and add 20% on top.
+						// Overly complicated calculations using Range, Speed and the old RangeLimit value would be rather pointless,
+						// because currently most mods have somewhat arbitrary, usually too high and in a few cases too low RangeLimits anyway.
+						var newValue = oldWDist.Length * 120 / 100;
+						var newCells = newValue / 1024;
+						var newCellPart = newValue - (newCells * 1024);
```
abcdefg30: Wait, I don't get this here... Above we say `newCells = newValue / 1024` so this here gets `newCellPart = newValue - ((newValue / 1024) * 1024) = newValue - newValue = 0`?

GraionDilach: No, this is the equvivalent of `newCellPart = newValue % 1024` just a bit messy. `newValue` might not be fully dividable with 1024 afterall.

reaperrr: Let's say `oldWDist` was `2c0 / 2048`.

`newValue` is the full new distance value, in our example `2457` (`2c409`).
`newCells` divides that by 1024, the result would be ~2.4 in fp, but is truncated to 2 for integer div, so this is the number of full cells (in other words, the number before the `c)`.
`newCellPart` then takes the original full WDist (2457), and substracts the number of full cells * 1024 (2 * 1024 = 2048 in this example), so the result here will be 2457 - 2048 = 409. So we have the WDist to put after the `c` as well now.

abcdefg30: I think we should use `newValue % 1024` then.

reaperrr:
Updated.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/12f43aecdbeb52ec5b291b0b7f8f3b675e1f5769...52879969195c4dd29b673f81ec7261d8e4796720)

Issue #4791, opened 2014-03-05
Craters should be removed when placing a building or deploying an MCV on top of it.


VVheat:
Right now, after the building is sold, or the MCV unpacked, the crater is still there.

This would let players undo the damage they did to the map.

evgeniysergeev:
Any news on this issue?
This video illustrate issue: https://streamable.com/t945

yaaaaa:
Евгений, мод dune2? где можно последить за прогрессом?

evgeniysergeev:
#11207 will helps to clean the ground.

PR #11207, opened 2016-04-30
add RemoveSmudge function

evgeniysergeev:
Allow not only Add smudges in SmudgeLayer but remove them too. This need to clean craters then put new concrete on rocks with craters or on concrete with craters.

Closes https://github.com/OpenRA/OpenRA/issues/4791.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Buildings/Building.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/Building.cs
@@ -34,6 +34,12 @@ namespace OpenRA.Mods.Common.Traits
 		public readonly CVec Dimensions = new CVec(1, 1);
 		public readonly bool RequiresBaseProvider = false;
 		public readonly bool AllowInvalidPlacement = false;
+		[Desc("Clear smudges from underneath the building footprint.")]
+		public readonly bool RemoveSmudgesOnBuild = true;
+		[Desc("Clear smudges from underneath the building footprint on sell.")]
+		public readonly bool RemoveSmudgesOnSell = true;
+		[Desc("Clear smudges from underneath the building footprint on transform.")]
+		public readonly bool RemoveSmudgesOnTransform = true;
 
 		public readonly string[] BuildSounds = { "placbldg.aud", "build5.aud" };
 		public readonly string[] UndeploySounds = { "cashturn.aud" };
@@ -185,6 +191,9 @@ namespace OpenRA.Mods.Common.Traits
 
 		public virtual void AddedToWorld(Actor self)
 		{
+			if (Info.RemoveSmudgesOnBuild)
+				RemoveSmudges();
+
 			self.World.ActorMap.AddInfluence(self, this);
 			self.World.ActorMap.AddPosition(self, this);
 
@@ -215,6 +224,9 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void Selling(Actor self)
 		{
+			if (Info.RemoveSmudgesOnSell)
+				RemoveSmudges();
+
 			BuildComplete = false;
 		}
 
@@ -222,11 +234,23 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void BeforeTransform(Actor self)
 		{
+			if (Info.RemoveSmudgesOnTransform)
+				RemoveSmudges();
+
 			foreach (var s in Info.UndeploySounds)
 				Game.Sound.PlayToPlayer(self.Owner, s, self.CenterPosition);
 		}
 
 		public void OnTransform(Actor self) { }
 		public void AfterTransform(Actor self) { }
+
+		public void RemoveSmudges()
+		{
+			var smudgeLayers = self.World.WorldActor.TraitsImplementing<SmudgeLayer>();
+
+			foreach (var smudgeLayer in smudgeLayers)
+				foreach (var footprintTile in FootprintUtils.Tiles(self))
+					smudgeLayer.RemoveSmudge(footprintTile);
+		}
 	}
 }
--- a/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
+++ b/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
@@ -142,7 +142,7 @@ namespace OpenRA.Mods.Common.Traits
 			if (Game.CosmeticRandom.Next(0, 100) <= Info.SmokePercentage)
 				world.AddFrameEndTask(w => w.Add(new SpriteEffect(world.Map.CenterOfCell(loc), w, Info.SmokeType, Info.SmokeSequence, Info.SmokePalette)));
 
-			if (!dirty.ContainsKey(loc) && !tiles.ContainsKey(loc))
+			if ((!dirty.ContainsKey(loc) || dirty[loc].Sprite == null) && !tiles.ContainsKey(loc))
 			{
 				// No smudge; create a new one
 				var st = smudges.Keys.Random(world.SharedRandom);
@@ -163,6 +163,22 @@ namespace OpenRA.Mods.Common.Traits
 			}
 		}
 
+		public void RemoveSmudge(CPos loc)
+		{
+			if (dirty.ContainsKey(loc))
+			{
+				var tile = dirty[loc];
+				tile.Depth = 0;
+				tile.Sprite = null;
+				dirty[loc] = tile;
+			}
+			else
+			{
+				var st = smudges.Keys.Random(world.SharedRandom);
+				dirty[loc] = new Smudge { Type = st, Depth = 0, Sprite = null };
+			}
+		}
+
 		public void TickRender(WorldRenderer wr, Actor self)
 		{
 			var remove = new List<CPos>();
@@ -170,7 +186,10 @@ namespace OpenRA.Mods.Common.Traits
 			{
 				if (!self.World.FogObscures(kv.Key))
 				{
-					tiles[kv.Key] = kv.Value;
+					if (kv.Value.Sprite == null)
+						tiles.Remove(kv.Key);
+					else
+						tiles[kv.Key] = kv.Value;
 					render.Update(kv.Key, kv.Value.Sprite);
 
 					remove.Add(kv.Key);
--- a/mods/d2k/tilesets/arrakis.yaml
+++ b/mods/d2k/tilesets/arrakis.yaml
@@ -19,6 +19,7 @@ Terrain:
 	TerrainType@Concrete:
 		Type: Concrete
 		TargetTypes: Ground
+		AcceptsSmudgeType: RockCrater
 		Color: E8C498
 	TerrainType@Dune:
 		Type: Dune


OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
```diff
@@ -163,18 +163,36 @@ public void AddSmudge(CPos loc)
...
+		public void RemoveSmudge(CPos loc)
+		{
+			if (dirty.ContainsKey(loc))
+			{
+				var tile = dirty[loc];
+				tile.Sprite = null;
+				dirty[loc] = tile;
+			} else {
+				var st = smudges.Keys.Random(world.SharedRandom);
+				dirty[loc] = new Smudge { Type = st, Depth = 0, Sprite = null };
```
pchote: Why is this creating a random empty smudge?

evgeniysergeev: using null sprite as indicator to update render in this point and remove this null sprite in TickRender function

GraionDilach:
Could this be used at D2k concrete? I assume this only adds the function as-is, but does not trigger it.

evgeniysergeev:
Usage example: 
[D2ClearTerrain.cs.zip](https://github.com/OpenRA/OpenRA/files/243969/D2ClearTerrain.cs.zip)

evgeniysergeev:
added RemoveSmudges trait in common mod and use it in d2k for concrete slabs.

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
```diff
@@ -163,14 +163,29 @@ public void AddSmudge(CPos loc)
...
 		}
 
+		public void RemoveSmudge(CPos loc)
+		{
+			if (dirty.ContainsKey(loc))
+			{
+				var tile = dirty[loc];
+				tile.Sprite = null;
+				dirty[loc] = tile;
+			} else {
```
abcdefg30: Please move the `else` onto a new line and drop the braces:
```
}
else
	dirty[loc] = new Smudge { Type = "", Depth = 0, Sprite = null };
```

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
```diff
@@ -163,14 +163,29 @@ public void AddSmudge(CPos loc)
...
+				dirty[loc] = new Smudge { Type = "", Depth = 0, Sprite = null };
```
pchote: I think you will need to set a valid Type here (so the original random would have been fine), otherwise you will cause a crash if something else calls AddSmudge on the same cell before the dirty cells are resolved.

OpenRA.Mods.Common/Traits/Buildings/RemoveSmudges.cs
```diff
@@ -0,0 +1,53 @@
...
+#endregion
+
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public class RemoveSmudgesInfo : ITraitInfo, Requires<BuildingInfo>
```
pchote: I suspect this would be better done via a flag on `BuildingInfo`, so that you can reuse the footprint defined there.

pchote: You could then enable the flag by default to fix #4791.

mods/d2k/rules/world.yaml
```diff
@@ -83,6 +83,10 @@ World:
 		Type: RockCrater
 		Sequence: rockcraters
 		SmokePercentage: 0
+	SmudgeLayer@Concrete:
+		Type: ConcreteCrater
+		Sequence: rockcraters
```
pchote: This uses the same artwork as the rock craters, so why does it need a new layer?

evgeniysergeev: fixed

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
evgeniysergeev: restored

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
evgeniysergeev: restored

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
penev92: Please move then to separate lines:
```
}
else
{
```
like we do everywhere else.

OpenRA.Mods.Common/Traits/Buildings/Building.cs
```diff
@@ -34,6 +34,10 @@ public class BuildingInfo : ITraitInfo, IOccupySpaceInfo, IPlaceBuildingDecorati
 		public readonly CVec Dimensions = new CVec(1, 1);
 		public readonly bool RequiresBaseProvider = false;
 		public readonly bool AllowInvalidPlacement = false;
+		[Desc("clear smudges before add building")]
+		public readonly bool RemoveSmudgesOnBuild = true;
+		[Desc("clear smudges when selling")]
+		public readonly bool RemoveSmudgesOnSell = true;
```
pchote: What's the use case for this?

If we keep it, then it should apply on Selling instead of Sold (otherwise you'll see the smudges pop away), and there should be a similar flag for transforming.

OpenRA.Mods.Common/Traits/Buildings/Building.cs
```diff
@@ -34,6 +34,10 @@ public class BuildingInfo : ITraitInfo, IOccupySpaceInfo, IPlaceBuildingDecorati
 		public readonly CVec Dimensions = new CVec(1, 1);
 		public readonly bool RequiresBaseProvider = false;
 		public readonly bool AllowInvalidPlacement = false;
+		[Desc("clear smudges before add building")]
```
pchote: These are sentences, so should start with a capital letter and end with a full stop.  It would also be useful to mention that they apply to the building footprint.

e.g. "Clear smudges from underneath the building footprint."

OpenRA.Mods.Common/Traits/Buildings/Building.cs
```diff
@@ -226,5 +237,19 @@ public void BeforeTransform(Actor self)
 
 		public void OnTransform(Actor self) { }
 		public void AfterTransform(Actor self) { }
+
+		public void RemoveSmudges()
+		{
+			var smudgeLayers = self.World.WorldActor.TraitsImplementing<SmudgeLayer>().ToDictionary(x => x.Info.Type);
```
pchote: You aren't using this as a dictionary, so please keep it as an IEnumerable.

OpenRA.Mods.Common/Traits/Buildings/Building.cs
evgeniysergeev: OnSell added becase of #4791. Not sure is it very useful, but adding that flag is easy so why not

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
abcdefg30: Well, we don't need the last brace, as it's a one-line-statement.

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
```diff
@@ -163,14 +163,32 @@ public void AddSmudge(CPos loc)
...
+			if (dirty.ContainsKey(loc))
+			{
+				var tile = dirty[loc];
+				tile.Sprite = null;
+				dirty[loc] = tile;
+			}
+			else
+			{
+				var st = smudges.Keys.Random(world.SharedRandom);
+				dirty[loc] = new Smudge { Type = st, Depth = 0, Sprite = null };
```
Phrohdoh: If there isn't a smudge here already why add an invisible one?

evgeniysergeev: to update later this location in TickRender

RoosterDragon: Since you want to clear the smudge either way, this entire method can be simplified to `dirty[loc] = new Smudge();`.

pchote: That would break the case where someone calls `AddSmudge` before the dirty cells are committed.

RoosterDragon: Good point, that kind of sucks.

OpenRA.Mods.Common/Traits/Buildings/Building.cs
```diff
@@ -213,18 +222,35 @@ public void NotifyBuildingComplete(Actor self)
...
+
+		public void RemoveSmudges()
+		{
+			var smudgeLayers = self.World.WorldActor.TraitsImplementing<SmudgeLayer>();
+
+			foreach (var occupiedCell in OccupiedCells())
+			{
+				foreach (var smudgeLayer in smudgeLayers)
+					smudgeLayer.RemoveSmudge(occupiedCell.First);
+			}
```
RoosterDragon: You may wish to switch the two loops here so you only have to enumerate the smudge layers once.

```c#
foreach (var smudgeLayer in smudgeLayers)
    foreach (var occupiedCell in OccupiedCells())
        ...
```

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
RoosterDragon: In that case, if a previous smudge exists, should we not blanket overwrite it? (i.e. reset the depth and randomize the type again) otherwise adding a new one will reuse the same type at whatever the old depth was, right?

evgeniysergeev: AddSmudge now generate smudge again if existing smudge have null Sprite

OpenRA.Mods.Common/Traits/Buildings/Building.cs
evgeniysergeev: loops switched. smudgeLayer now topmost loop

Mailaender:

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
```diff
@@ -142,7 +142,7 @@ public void AddSmudge(CPos loc)
 			if (Game.CosmeticRandom.Next(0, 100) <= Info.SmokePercentage)
 				world.AddFrameEndTask(w => w.Add(new Smoke(w, world.Map.CenterOfCell(loc), Info.SmokeType, Info.SmokePalette, Info.SmokeSequence)));
 
-			if (!dirty.ContainsKey(loc) && !tiles.ContainsKey(loc))
+			if ((!dirty.ContainsKey(loc) || (dirty[loc].Sprite == null)) && !tiles.ContainsKey(loc))
```
pchote: The `()` around `dirty[loc].Sprite == null` aren't needed here.

OpenRA.Mods.Common/Traits/Buildings/Building.cs
```diff
@@ -213,18 +222,35 @@ public void NotifyBuildingComplete(Actor self)
...
 
 		public void OnTransform(Actor self) { }
 		public void AfterTransform(Actor self) { }
+
+		public void RemoveSmudges()
+		{
+			var smudgeLayers = self.World.WorldActor.TraitsImplementing<SmudgeLayer>();
+
+			foreach (var smudgeLayer in smudgeLayers)
+			{
```
pchote: Minor style nit: these braces aren't needed here.

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
pchote: You should be able to simplify this to @RoosterDragon's suggestion now then, right?

abcdefg30:
Isn't https://github.com/OpenRA/OpenRA/pull/11207/files#r63291380 still an open point?

evgeniysergeev:
I think all requested changes made. but tell me if it is not.

GraionDilach:
You should squash the two commits and cleanup the commit messages then.

abcdefg30:
> I think all requested changes made. but tell me if it is not.

You should be able to just use `dirty[loc] = new Smudge();` now [in line 178](https://github.com/OpenRA/OpenRA/pull/11207/files#diff-288b404a7eb1501ffc0c6919729361a3R178) as suggested in
https://github.com/OpenRA/OpenRA/pull/11207#discussion_r62227568.

evgeniysergeev:
> You should be able to just use dirty[loc] = new Smudge(); now in line 178 as suggested in
> #11207 (comment).

What if dirty contain null sprite, but tiles contain some Sprite at location loc. Then ‘if’ in line 178 goes to ‘else’ statement but dirty[loc].st at this point is zero. I think dirty[loc].st should be always valid value, even if Sprite is null

evgeniysergeev:
> should squash the two commits and cleanup the commit messages then.

first commit related to ability removing smudges, second uses this ability. Are you sure about squash this different commits ?

RoosterDragon:
I think that's acceptable.

evgeniysergeev:
finally, do I need to modify anything in this pr?

OpenRA.Mods.Common/Traits/Buildings/Building.cs
```diff
@@ -213,18 +222,33 @@ public void NotifyBuildingComplete(Actor self)
...
 
 		public void OnTransform(Actor self) { }
 		public void AfterTransform(Actor self) { }
+
+		public void RemoveSmudges()
+		{
+			var smudgeLayers = self.World.WorldActor.TraitsImplementing<SmudgeLayer>();
+
+			foreach (var smudgeLayer in smudgeLayers)
+				foreach (var occupiedCell in OccupiedCells())
```
pchote: I don't think `OccupiedCells()` is the right thing here.  This excludes building bibs and cells that are part of the building but passable by other units (e.g. the refinery unloading bay).

pchote: You probably want to use `FootprintUtils.Tiles` instead of `FootprintUtils.UnpathableTiles` (indirectly via `OccupiedCells`).

pchote:
Sorry for the delay.  I can't speak for the others, but I've had almost no time spare for OpenRA recently.

OpenRA.Mods.Common/Traits/Buildings/Building.cs
evgeniysergeev: OccupiedCells replaced with FootprintUtils.Tiles

abcdefg30:
https://github.com/OpenRA/OpenRA/pull/11207#discussion_r63291370 still needs to be resolved.

abcdefg30:

yaaaaa:
=)

evgeniysergeev:
done

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/0096f1a3d0c2e9a521754bf769ed6180a3d0afc8)

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
```diff
@@ -163,14 +163,33 @@ public void AddSmudge(CPos loc)
...
 			var remove = new List<CPos>();
 			foreach (var kv in dirty)
 			{
 				if (!self.World.FogObscures(kv.Key))
 				{
-					tiles[kv.Key] = kv.Value;
+					if (kv.Value.Sprite == null)
+						tiles.Remove(kv.Key);
+					else
+						tiles[kv.Key] = kv.Value;
```
Mailaender: Isn't this redundant to the code commented above meaning that `render.Update` will already take care as it was dirtied?

Could this be the reason for #11660 because it changes game logic in a render tick and relies on `RenderPlayer` via `FogObscures`?

PR #11211, opened 2016-05-01
Added a CustomTerrainDebugOverlay

Mailaender:
A custom terrain visualizing debug overlay, see https://github.com/OpenRA/OpenRA/pull/10764#issuecomment-186585903

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/CustomTerrainDebugOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/CustomTerrainDebugOverlay.cs
@@ -1,0 +1,79 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using System.Drawing;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Commands;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Displays custom terrain types.")]
+	class CustomTerrainDebugOverlayInfo : ITraitInfo
+	{
+		public readonly string Font = "TinyBold";
+
+		public object Create(ActorInitializer init) { return new CustomTerrainDebugOverlay(init.Self, this); }
+	}
+
+	class CustomTerrainDebugOverlay : IWorldLoaded, IChatCommand, IRender
+	{
+		const string CommandName = "debugcustomterrain";
+		const string CommandDesc = "Toggles the custom terrain debug overlay.";
+
+		public bool Enabled;
+
+		readonly SpriteFont font;
+
+		public CustomTerrainDebugOverlay(Actor self, CustomTerrainDebugOverlayInfo info)
+		{
+			font = Game.Renderer.Fonts[info.Font];
+		}
+
+		void IWorldLoaded.WorldLoaded(World w, WorldRenderer wr)
+		{
+			var console = w.WorldActor.TraitOrDefault<ChatCommands>();
+			var help = w.WorldActor.TraitOrDefault<HelpCommand>();
+
+			if (console == null || help == null)
+				return;
+
+			console.RegisterCommand(CommandName, this);
+			help.RegisterHelp(CommandName, CommandDesc);
+		}
+
+		void IChatCommand.InvokeCommand(string name, string arg)
+		{
+			if (name == CommandName)
+				Enabled ^= true;
+		}
+
+		IEnumerable<IRenderable> IRender.Render(Actor self, WorldRenderer wr)
+		{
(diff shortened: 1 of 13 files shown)


OpenRA.Mods.Common/Traits/Render/CustomTerrainDebugOverlay.cs
```diff
@@ -0,0 +1,85 @@
...
+			if (name == CommandName && devMode != null)
+				devMode.ShowCustomTerrain ^= true;
+
+			if (arg.Contains("allcells"))
+				allCells ^= true;
+		}
+
+		public IEnumerable<IRenderable> Render(Actor self, WorldRenderer wr)
+		{
+			if (devMode != null && !devMode.ShowCustomTerrain)
```
pchote: The `devMode != null` makes this unconditionally render the overlays in shellmaps and for observers, which is pretty rotten for polish.

pchote: You should follow the example of TerrainGeometryOverlay instead.

Mailaender:
Updated. This make the trait also properly optional.

OpenRA.Mods.Common/Traits/Render/CustomTerrainDebugOverlay.cs
```diff
@@ -0,0 +1,84 @@
...
+	{
+		public readonly string Font = "TinyBold";
+
+		public object Create(ActorInitializer init) { return new CustomTerrainDebugOverlay(init.Self, this); }
+	}
+
+	class CustomTerrainDebugOverlay : IWorldLoaded, IChatCommand, IRender
+	{
+		const string CommandName = "debugcustomterrain";
+		const string CommandDesc = "Toggles the custom terrain debug overlay. Optional parameter: 'allcells'";
```
pchote: This optional parameter doesn't appear to do anything.  Based on the name I would expect it to show the terrain type for all cells, not just the ones with custom types?

pchote: How odd.  When I tried this again it worked, I must have typed it wrong 5 or 6 times...

However, this just shows 255 colored in the color of the actual terrain type.  If you're not going to show the actual terrain type, then what is the use case of this?

pchote: Adding the changes requested tag to either remove this or explain the use case that justifies it.

Mailaender: It is a left over from a previous implementation. We don't need it. Removed.

pchote:
The terrain type numbers are generated by the engine and aren't exposed anywhere that a modder could find them.  This works a bit better (especially when zoomed) with the following change:
```diff
diff --git a/OpenRA.Mods.Common/Traits/Render/CustomTerrainDebugOverlay.cs b/OpenRA.Mods.Common/Traits/Render/CustomTerrainDebugOverlay.cs
index abc3d8e..71425fe 100644
--- a/OpenRA.Mods.Common/Traits/Render/CustomTerrainDebugOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/CustomTerrainDebugOverlay.cs
@@ -71,8 +71,8 @@ public IEnumerable<IRenderable> Render(Actor self, WorldRenderer wr)
                                if (terrainType == byte.MaxValue)
                                        continue;
 
-                               var color = wr.World.Map.GetTerrainInfo(cell).Color;
-                               yield return new TextRenderable(font, center, 0, color, terrainType.ToString());
+                               var info = wr.World.Map.GetTerrainInfo(cell);
+                               yield return new TextRenderable(font, center, 0, info.Color, info.Type);
                        }
                }
        }
```

screen shot 2016-06-05 at 00 01 00

Mailaender:
Updated with friendly terrain names and explicit interfaces.

OpenRA.Mods.Common/Widgets/Logic/Ingame/DebugMenuLogic.cs
```diff
@@ -134,6 +134,18 @@ public DebugMenuLogic(Widget widget, World world)
 				showActorTagsCheckbox.IsChecked = () => devTrait.ShowActorTags;
 				showActorTagsCheckbox.OnClick = () => devTrait.ShowActorTags ^= true;
 			}
+
+			var showCustomTerrainCheckbox = widget.GetOrNull<CheckboxWidget>("SHOW_CUSTOMTERRAIN_OVERLAY");
+			if (showCustomTerrainCheckbox != null)
+			{
+				var customTerrainDebugTrait = world.WorldActor.Trait<CustomTerrainDebugOverlay>();
+				showCustomTerrainCheckbox.Disabled = customTerrainDebugTrait == null;
```
Phrohdoh: Checking for null indicates that you want `TraitOrDefault`.

Mailaender: Yes, indeed. Thank you.

GraionDilach:
:+1:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/038d2174c52c285788fa91b99d13bcb4e71614f5...95406a831475da885087de2a4f52b301437b53da)

PR #11215, opened 2016-05-02
Extract common campaign rules for d2k and ra

obrakmann:


HenrytheSlav:
Good to see this trend on, Placeholder and remider for me to test it when I get home.

HenrytheSlav:
OK, I am not sure if I done everything right, but I get (same) crashes for example in sov1, 3 and 4a:
Exception of type `System.InvalidOperationException`: :0: Sequence v04.damaged-active uses frames [3..3], but only 0..2 actually exist

It seems some civ building (v04) has badly set frames?

obrakmann:
This PR doesn't touch any sequences at all. On bleed, v04 does indeed use frame 3 for the damaged-active sequence, but the frame actually exists in the shp. Check the asset browser to see how many frames your v04.sno file has. Maybe you've got a corrupted install or something.

HenrytheSlav:
That was an error on my side, tested many maps in both RA and D2k, no issues encountered. :+1:

obrakmann:
Added `Demolishable` to tanks and vehicles by default.

abcdefg30:

PR #11216, opened 2016-05-02
Make the circle and arrow sequences of Beacon optional

abcdefg30:
You might only want to use one of them, especially for support powers, where you can't use `AnimatedBeacon` or just want a `beaconposter` without any other animations.

### DIFF
--- a/OpenRA.Mods.Common/Effects/Beacon.cs
+++ b/OpenRA.Mods.Common/Effects/Beacon.cs
@@ -42,11 +42,17 @@ namespace OpenRA.Mods.Common.Effects
 			this.beaconPalette = beaconPalette;
 			this.isPlayerPalette = isPlayerPalette;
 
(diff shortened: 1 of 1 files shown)


Phrohdoh:
The lint passes should probably guarantee that at least one is used.

abcdefg30:
Are we sure about that? As I said, you might only want a `beaconposter`.

Phrohdoh:
Oh, sorry I had forgotten that was a part of the beacon setup.
That's what I get for making quick comments.

obrakmann:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3094568fe01ea99609fbc035fa81501cc6a0051f...8143e51c8e3187beb18e5d9815ce081d1383061e)

Issue #8544, opened 2015-06-22
Utility gets rotations inverted by Y-axis


HenrytheSlav:
Hi guys, while converting maps from original soviet RA campaign I noticed a lot of units facing walls and stuff. The problem is apparently in the utility, which converts the rotations badly. In two screenshots, there is original map shown in RAED with correct rotations, and then screenshot of unchanged converted map. You can see that for example harvester remains same, facing down, while the units near the wall completely change direction left to right, and those facing SW now face SE and reverse.

![openra 2015-06-22 19-14-39-64](https://cloud.githubusercontent.com/assets/12610736/8288055/f22dae96-1913-11e5-836b-0df5bdaf8ac0.png)
![bez nazvu](https://cloud.githubusercontent.com/assets/12610736/8288056/f2301848-1913-11e5-9be9-8f20371ef249.png)

clemty:
I suspect the utility does it right, and OpenRA itself displays the units incorrectly.

from http://cnc.wikia.com/wiki/Tiberian_Dawn_Mission_Manual
>direction which it is facing. 0=north, 32=north-east, 64=east,
>96=south-east, 128=south, 160=south-west 192=west, 224=north-west.

I converted a TD mission (scg30ea), when run in C&C95 1.06, the gunship faces left; after importing it into OpenRA it faces right (yaml contains "Facing: 192" for the boat). (After manually changing it to 64 it faces east)

Unit158:
Facings should work in the direction of bearings if they do not do so already.

HenrytheSlav:
Closed in #11227.

PR #11227, opened 2016-05-05
Fixes Utility import facings mirror-flip

HenrytheSlav:
Solves #8544.
Tested on several maps and it seems to work.
I warn you that I know next to nothing about C# or how it uses integers like that, but the scope of that should still be 0-255.
If you think there is some issue/harm being done with crude hack like this, we can rework it.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
@@ -390,7 +390,7 @@ namespace OpenRA.Mods.Common.UtilityCommands
 					if (health != 100)
 						initDict.Add(new HealthInit(health));
 					if (facing != 0)
-						initDict.Add(new FacingInit(facing));
+						initDict.Add(new FacingInit(255 - facing));
 
 					if (section == "INFANTRY")
 						actor.Add(new SubCellInit(Exts.ParseIntegerInvariant(parts[4])));


OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -390,7 +390,7 @@ public static void LoadActors(IniFile file, string section, List<string> players
 					if (health != 100)
 						initDict.Add(new HealthInit(health));
 					if (facing != 0)
-						initDict.Add(new FacingInit(facing));
+						initDict.Add(new FacingInit(255 - facing));
```
Lukemtesta: Instead of hard coding this, it would be better to define a configurable default argument to use 8-bit int max here

GraionDilach: I can see you're new here... there's no reason to configure this, because this code only deals with Westwood Tiberian Dawn/Red Alert maps which had only one way to define facings.

8bit int mix looks nicer though, but the hardcoding is absolutely justified.

GraionDilach:
:+1:

obrakmann:
:+1:

obrakmann:
Decided to put this into stable. It's simple enough.

[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/92a44f57977084d628062c458e60120f15e619cb).

HenrytheSlav:
Thank @obrakmann, that will save trouble to people importing maps in next devcycle :hamster:

PR #11232, opened 2016-05-06
Fix possible instant repairs by spamming Repair activities

abcdefg30:
Noticeable when spam-clicking on a naval yard/sub pen.


obrakmann:
The downside of this is that the unit lags a bit when you give it another order while it's repairing, but I guess that's justifiable. :+1:

HenrytheSlav:
Just tested, after I found out how to reproduce, and can confirm this fixes the issue. :+1:

pchote:
:+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/3376f681f954039be706418a80c39fca844c2c62

PR #11235, opened 2016-05-07
TD Campaign: Add nod07b - "Sick And Dying (b)

ghost:
Add nod07b - "Sick And Dying (b)"
http://cnc.wikia.com/wiki/Sick_And_Dying


HenrytheSlav:
Tested, one minor nitpick is that Oil Derricks belong to the Civilians, who are enemy of NOD which means your army will autotarget them, but you might have interest in preserving them. Otherwise issues encountered :+1:

Biofreak1987:
Fixed all issues listed above except for the harvester one ... I've got idea why it does that.

HenrytheSlav:
That will be probably general pathfinding bug. BTW, I noticed that this mission has village as enemy (Civilians) while the (a) doesn't so no autotargetting there. It is weird to have it different, but I guess it is for you to decide which you want to have.

I have no idea why appveyor failed though.

Otherwise :+1:

Biofreak1987:
I've changed it on (a) as you are forced to pass through the village in order to reach your base. It would be way too annoying if your units autotarget those buildings.

mods/cnc/maps/nod07b/nod07b.lua
```diff
@@ -0,0 +1,265 @@
...
+SetAttackWaypoints = function(actor, waypoints)
+	if not actor.IsDead then
+		Utils.Do(waypoints, function(waypoint)
+			actor.AttackMove(waypoint.Location)
+		end)
+	end
+end
+
+SendGDIAirstrike = function(hq, delay)
+	if not hq.IsDead and hq.Owner == enemy then
```
Mailaender:
```lua
and not player.IsObjectiveCompleted(NodObjective2)
```

otherwise you still get harassed no matter the SAM sites.

Biofreak1987: Wouldn't that prevent airstrikes on the entire map? I thought the point of that code was to secure your base from airstrikes.

Mailaender: Ah okay. That was a misunderstanding from my side then.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/52879969195c4dd29b673f81ec7261d8e4796720...05bcc17cd6a6244b11ade0383ae1c914098f2b17)

HenrytheSlav:
Good job at pushing the campaign to completion @Biofreak1987!

PR #11237, opened 2016-05-07
Prevent projectiles from overshooting blocking actors

reaperrr:
One of the prerequisites for directional armor and explosion effects.

While the issue with blockable projectiles passing through walls was fixed a while ago, they still quite often explode on the 'wrong side' of the blocking actor.

This PR calculates the direction the projectile came from and enforces the explosion to happen on the side facing the source (or at least previous position of the projectile).

The concrete walls in our shipping mods already provide a good test case for this.

### DIFF
--- a/OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
+++ b/OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
@@ -58,7 +58,12 @@ namespace OpenRA.Mods.Common.Traits
 				if (!blockers.Any())
 					continue;
 
-				var hitPos = WorldExtensions.MinimumPointLineProjection(start, end, a.CenterPosition);
+				// FindActorsOnLine only finds actors with health trait, so this actor is guaranteed to have it
+				var healthInfo = a.Info.TraitInfo<HealthInfo>();
+				var vector = (a.CenterPosition - source) * healthInfo.Shape.InnerRadius.Length / (a.CenterPosition - source).Length;
+				var hitShapeEdgePos = a.CenterPosition - vector;
+				var hitPos = WorldExtensions.MinimumPointLineProjection(source, a.CenterPosition, hitShapeEdgePos);
+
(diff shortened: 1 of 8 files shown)


OpenRA.Mods.Common/Traits/BlocksProjectiles.cs
```diff
@@ -58,7 +58,12 @@ public static bool AnyBlockingActorsBetween(World world, WPos start, WPos end, W
 				if (!blockers.Any())
 					continue;
 
-				var hitPos = WorldExtensions.MinimumPointLineProjection(start, end, a.CenterPosition);
+				// FindActorsOnLine only finds actors with health trait, so this actor is guaranteed to have it
+				var healthInfo = a.Info.TraitInfo<HealthInfo>();
+				var vector = (a.CenterPosition - source) * healthInfo.Shape.InnerRadius.Length / (a.CenterPosition - source).Length;
```
pchote: Why do you need to introduce `source` here? Isn't `start` a more accurate choice?

pchote: Likewise for `InnerRadius`.  You care about the direction rather than the length, so you shouldn't need to scale this at all.

pchote: The overall goal is to find the "first" intersection of the line defined by (end - start) with the hitbox shape (do you really want the edge? or do you want the middle?).  You can't do this properly because the `IHitShape` interface doesn't expose the geometry info you need – `InnerRadius` is still a rather crude workaround for this that i'm not really happy about.

A better way to do this will be to define a `bool IHitShape.TryGetIntersection(WPos start, WPos end, out WPos intersection)` so that each shape can do its own logic to find the appropriate intersection point (e.g. circle vs line test for circles, line vs line tests for rectangle).

reaperrr:
I'll probably need some time to adress this properly, so I'm closing it for now to not clog up the PR queue.

Issue #8115, opened 2015-05-08
old westwood-maps causing lobby-crash


Micr0Bit:
some of those old westwood maps that didnt get fixed by any players ... do allow more players then actually spots are available on the map ... this is causing the lobby to crash after a couple of seconds 

the map "stalingrad" is a example http://resource.openra.net/maps/2540/... it has 5 player-slots ... but only 4 map-starting-points

______________________________________________________________________
Red Alert Mod at Version release-20150424
Operating System: Windows (Microsoft Windows NT 6.1.7601 Service Pack 1)
Runtime Version: .NET CLR 4.0.30319.18408
Exception of type `System.ArgumentException`: Ein Element mit dem gleichen Schlüssel wurde bereits hinzugefügt.
   bei System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   bei System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   bei System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   bei System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector)
   bei OpenRA.Mods.Common.Widgets.Logic.LobbyUtils.GetSpawnOccupants(Session lobbyInfo, MapPreview preview)
   bei OpenRA.Mods.Common.Widgets.Logic.LobbyMapPreviewLogic.<>c__DisplayClass47.<.ctor>b__3()
   bei OpenRA.Mods.Common.Widgets.MapPreviewWidget.Draw()
   bei OpenRA.Widgets.Widget.DrawOuter()
   bei OpenRA.Widgets.Widget.DrawOuter()
   bei OpenRA.Widgets.Widget.DrawOuter()
   bei OpenRA.Widgets.Widget.DrawOuter()
   bei OpenRA.Widgets.Widget.DrawOuter()
   bei OpenRA.Widgets.Widget.DrawOuter()
   bei OpenRA.Widgets.Widget.DrawOuter()
   bei OpenRA.Widgets.Ui.Draw()
   bei OpenRA.Game.RenderTick()
   bei OpenRA.Game.Loop()
   bei OpenRA.Game.Run()
   bei OpenRA.Program.Run(String[] args)
   bei OpenRA.Program.Main(String[] args)
[...]

PR #11240, opened 2016-05-08
Added a lint test for actor locations

Mailaender:
I wanted to write a lint check to detect flaws like https://github.com/OpenRA/OpenRA/issues/8115 and https://github.com/OpenRA/OpenRA/issues/8467 and was surprised how often actors share a location in the maps we ship. Not sure how to deal with that.


pchote:
I'm not sure this is smart enough to behave well in the cases we care about.  This doesn't account for building footprints or actors which are allowed to overlap (e.g. those that don't reserve cells, or things like service depots or refineries).

Mailaender:
Indeed, removed the warning completely. Checking building footprints needs a way more sophisticated check and it will probably also detect lot's of non-problematic cases.

Issue #10101, opened 2015-11-24
Game crash when replay is selected in the wrong mod


StexoO:
I wanted to watch the replay named BTT2015-51.orarep (RA replay) so I copied it, went into my Documents\OpenRA\Replays folder, spam-clicked the first line in the Windows explorer (so the first alphabetical folder was cnc), pasted the replay there, started OpenRA (CNC mod running), Extras, Replays, and when I selected this replay the game crashed with
```
Tiberian Dawn Mod at Version release-20150919
on map b0f941e404b8113179ed8c358982f99d3b1021c5 (shellmap by Chris Forbes).
Operating System: Windows (Microsoft Windows NT 6.2.9200.0)
Runtime Version: .NET CLR 4.0.30319.42000
Exception of type `System.ArgumentException`: Sprite flags/ukraine was not found.
   à OpenRA.Mods.Common.Widgets.ImageWidget.Draw()
   à OpenRA.Widgets.Widget.DrawOuter()
   à OpenRA.Widgets.Widget.DrawOuter()
   à OpenRA.Mods.Common.Widgets.ScrollPanelWidget.DrawOuter()
   à OpenRA.Widgets.Widget.DrawOuter()
   à OpenRA.Widgets.Widget.DrawOuter()
   à OpenRA.Widgets.Widget.DrawOuter()
   à OpenRA.Widgets.Widget.DrawOuter()
   à OpenRA.Game.RenderTick()
   à OpenRA.Game.Loop()
   à OpenRA.Game.Run()
   à OpenRA.Program.Run(String[] args)
   à OpenRA.Program.Main(String[] args)
```
This seems logical but the game could display a dialog box or load the corresponding mod instead.

In the other hand, if you select the CNC mod in OpenRA before exiting OpenRA, and if you load this replay by double clicking the orarep file then OpenRA will start, load the RA mod and play the replay. :+1: 
Finally, pasting the replay in the directory with the right mod but not the right version (playtest-20151114 instead of release-20150919) leads to an "Incompatible Replay" dialog box. :+1:

PR #11241, opened 2016-05-08
Fixed a crash when replay is selected in the wrong mod

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/10101.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/ReplayBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/ReplayBrowserLogic.cs
@@ -18,6 +18,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using OpenRA.FileFormats;
 using OpenRA.Primitives;
+using OpenRA.Traits;
 using OpenRA.Widgets;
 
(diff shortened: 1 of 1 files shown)


RoosterDragon:
Perhaps we should just remove the flag widget if the faction doesn't exist, in case a mod doesn't provide a "Random" flag?

:+1: Works though, so happy to take as-is.

Mailaender:
I also thought about it, but the faction "Random" is hard-coded in other parts of the engine as well.

obrakmann:
I'm a bit torn on this. On the one hand, I think it's wrong that the replay browser even considers invalid replay files at all, since #10101 is clearly a user error (I just came up with [this](https://github.com/OpenRA/OpenRA/compare/bleed...obrakmann:fix10101_discard-invalid-replays-in-browser) to fix that), on the other hand, there'd be no way - at least not a nice one - to inform the user about the issue, which also sucks.

Ho-hum. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/e9f4161a484539117537f78b16c2435917f53dbf)

PR #11243, opened 2016-05-08
Polished the Tiberian Sun laser weaponry

Mailaender:
Marn's mod looks way more polished. He fixed the blocky laser artwork by making them less thick and adding transparency to the sonic beam which looks way more like in the original this way.


GraionDilach:
Could this look consistent? The values I'm fine with, but some having an unneeded 0c prefix and others lacking looks odd to me.

reaperrr:
I agree with @GraionDilach, :+1: after that.

Mailaender:
Removed `0c` everywhere.

PR #11244, opened 2016-05-08
Added a welding overlay to the Nod repair vehicle

Mailaender:
In Marn's Tiberian Sun mod the Nod repair tank does not only the welding sounds, but also has a nice visual overlay which makes it way easier to understand for the player what is going on.

### DIFF
--- a/mods/ts/rules/nod-vehicles.yaml
+++ b/mods/ts/rules/nod-vehicles.yaml
@@ -216,8 +216,11 @@ REPAIR:
 		OutsideRangeCursor: repair
 		TargetStances: Ally
 		ForceTargetStances: None
+		LocalOffset: 300,0,500
+		MuzzleSequence: emp-overlay
 	AttackFrontal:
 		Voice: Attack
+	WithMuzzleOverlay:
 
 WEED:
 	Inherits: ^VoxelTank


GraionDilach:
I'm not certain if the EMP overlay is a good idea. The original game thrown a lot of sparks from the welder, which can be simulated with a white 1x1px SHP inaccurate bullet weapon (although probably the repair vehicle should have 5-10 instance of this dummy weapon to not let them look like a stream but lots of sparks).

Mailaender:
So it was a hard-coded engine effect and not simply an SHP we just haven't discovered yet?

Marnador:
Yep it was an effect with randomly generated sparks that could even collide with the ground (from my memory) and it was also used on damaged vehicles but with a smaller amount of sparks.
I played around with GraionDilach idea and it looks pretty nice!
The only problem I found is if the dummy weapon has to have -1 damage to work(every spark heals when it hits the unit), and if you have it on burst it looses sync with the original repair weapon sometimes. It pretty much forced me to make it continuous.
But I am going to try and play with it regardless, see if anything breaks during a play-test.

GraionDilach:
@Mailaender: It used spark particles. See https://github.com/LipkeGu/TSIniFiles/blob/master/rules.ini#L8061 and https://github.com/LipkeGu/TSIniFiles/blob/master/rules.ini#L8191. WeldingPart was the part it was throwing (with randomized colors and velocities, the EC can be ignored because it was longer than the amount of frames they'd need to reach the ground and IIRC these particles just faded away imminently when reaching it) and WeldingSys controlled the effect (spawning 20 particles in each frame for 25 frames).

@Marnador D2k already has dummy weapons to properly render the trike's firinganims. See https://github.com/OpenRA/OpenRA/blob/bleed/mods/d2k/weapons/smallguns.yaml#L162 for such a dummy implementation - you do not even need to specify a value within the damagewarhead. I also stated that you should have a lot of dummyweapons (Armament@dummy1, Armament@dummy2 up to, say, 10) because yes, you cannot fire a weapon twice during the same frame.

GraionDilach:
@Marnador Okay, I had some thoughts on this. There are two issues with my method as-was (again, I still aim for the feelies):
1) The weldsparks cannot use Burst since that'll go out of sync - however using Armament's FireDelay will fix that, although that will mean even more dummyweapons, each having a single Burst (dummyweapon 1-10 having no FireDelay 11-20 has a frame of 1 or 2). This probably looks sloppy in the yaml for some though.
2) The TS sparks changed their color. Well, there are two options here, one can either use 1x1 animated projectiles or create a custom palette with crude but suitable gradients (IMO HSL based, with L varied from bright to dark, say 20% steps or more) and have an 1x1 SHP with the amount of frames of the HSL gradients, each having the gradient's white color and then use RotationPaletteEffect to switch colors instead of animating the SHP itself, setting up, so that each frame is a separate sequence of the same image. (I kinda prefer this latter myself tbh).

I'm all fine if you won't experiment with this, though but this is where I'd go regarding the welding sparks.

pchote:
It would be easier and cleaner to just add support for particle system animations.

PR #11249, opened 2016-05-09
Reduce RA Artillery range from 14 to 12 cells

cjshmyr:
Some consensus was reached on the [forums](http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19556).

### DIFF
--- a/mods/ra/weapons/largecaliber.yaml
+++ b/mods/ra/weapons/largecaliber.yaml
@@ -138,7 +138,7 @@ TurretGun:
 
 155mm:
 	ReloadDelay: 85
-	Range: 14c0
+	Range: 12c0
 	MinRange: 3c0
 	Report: tank5.aud
 	Projectile: Bullet


HenrytheSlav:
Since this seems like legit middle ground achieved between supporters of nerf, and conservatives, I am all for letting it into playtest.

abcdefg30:
Well, let's try this then. :+1:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ea7b0c9f7963a419ed72e4f98cecd85820c5108f...8c4b091296a959033cc41ad9564ec38c266a36d0)

airetaM:
This is a good idea but not without nerfing the flaktruck, wich is already a really good counter against artys

HenrytheSlav:
Unless you send arty without any protection flak trucks are useless unless they are under IC because of their weak armor/low health.

airetaM:
But you should not just nerf such an important allies unit with no compensate

HenrytheSlav:
1) this is probably smallest nerf we could make. It almost doesn't hurt arty at all, since every other but range remains same. 2) if you want to balance things, you have to get weak units better and strong a bit weaker. We nerfed arty. We are working on making Mamooth better. We are now discussing Chrono tanks, MiGs, Longbows and most important Shock Troopers on forums. Please, feel free to join discussion [there](http://www.sleipnirstuff.com/forum/viewforum.php?f=82). This is change was agreed on by community beforehand after long and tedious discussion.

PR #8674, opened 2015-07-07
Removed InitialActivity from FreeActor and Buildable

reaperrr:
For `Harvester`s, `FindResources` is now queued from `Created()`, provided `SearchOnCreation` is set to `true`.

PR #11250, opened 2016-05-10
Remove the last relicts of InitialActivity

abcdefg30:
Seems like this was forgotten in #8674. The upgrade rule looks fine though.
Reported by `The_madd` on IRC.


GraionDilach:
:+1:

HenrytheSlav:
Had a game, harvesters work as intended adter building ref/producing harv from refinery. :+1:

Issue #10865, opened 2016-03-03
[RA] Reduce Demo Truck damage against air units.


SoScared:
By my own hand I derailed the previous discussion on this on #10799 with a bad title and thus unintentionally argued for a general nerf of the Demo Truck. I should have been chased and lynched for that.

To keep it short, I'm arguing for a slight nerf on the Demo Truck's blast damage against air units. Currently it's a kill guarantee within about 10 cells' radius for the YAK/Hind and 8 cells for the MiG/Longbow. For comparison, here's the kill guarantee for the blast radius versus artillery and infantry:
![air inf](https://cloud.githubusercontent.com/assets/2715583/13485492/aa91f764-e108-11e5-8f11-1b2c80daa96d.png)

About 10 cells as well.

My comment from the previous issue:

Logically air units are more resilient against nuke blasts than ground targets.

Balance wise I think it's a good idea to not allow the nuke truck to completely wipe out a nearby (very expensive) air fleet 7-10 cells away. Scouting for nuke trucks should be rewarded. Sniping nuke trucks with lower tier air units already cost the attacker a loss ($1350 Hind/YAK).

Issue #10799, opened 2016-02-21
[RA] Nerf Demo Truck and A-Bomb damage against air units.


SoScared:
At ground 0 everything ought to be killed as the current state but IMO for air units, the range is too wide (aproximately 10 cells radius).

Hinds:
![hinds 1](https://cloud.githubusercontent.com/assets/2715583/13203028/56358b6c-d8ad-11e5-80e6-166699841d15.png)
![hinds 2](https://cloud.githubusercontent.com/assets/2715583/13203029/579987d8-d8ad-11e5-874c-e0f973fec5d7.png)

Yaks:
![yak 1](https://cloud.githubusercontent.com/assets/2715583/13203035/8d4340ae-d8ad-11e5-8324-faa4ce889d4f.png)
![yak 2](https://cloud.githubusercontent.com/assets/2715583/13203037/91fe8b3a-d8ad-11e5-8a39-a2b5b8f7d7da.png)

Logically air units are more resilient against nuke blasts than ground targets.

Balance wise I think it's a good idea to not allow the nuke truck to completely wipe out a nearby (very expensive) air fleet 7-10 cells away. Scouting for nuke trucks should be rewarded. Sniping nuke trucks with air units already guarantees the attacker a loss ($1350 Hind/YAK).

fgdzfggsfgsfgsf:
I agree, it feels a little large for killing infantry as well but just this is an improvement.

MustaphaTR:
Infantry is OK i think, but air should be nerfed.

kyrylo:
I agree, it's a bit large. I would make it smaller only by 1, though. Air units are mobile enough to escape from demo trucks.

klaas1:
Also, the demo truck is one of the most fun units in the game, and pretty risky to use. I don't mind it being high risk/ high reward. Please don't nerf it to the point people stop using them, because the game needs to be fun, not only perfectly balanced.

Micr0Bit:
i agree with @klaas1

SoScared:
@kyrylo Demo trucks can't be expected to be anticipated well tho - with exceptions they're spotted seconds before it's detonation and players in general don't really have the action points to move off their air units. I'd think a 7 cell guarantee-kill radius would be a good compromise.

[...]

PR #11253, opened 2016-05-10
Reduce effectiveness of demo truck against air targets

HenrytheSlav:
I did the nerf by removing the ValidTarget: Air from some of the warheads. Now it roughly seems that any unit in the scorched-earth area will be instantly killed, others are only damaged
openra-2016-05-10t184901z

This is VERY open to discussion, which already partly happened in #10799 but got sidetracked. This is response to both #10799 and #10865.

### DIFF
--- a/mods/ra/weapons/explosions.yaml
+++ b/mods/ra/weapons/explosions.yaml
@@ -104,7 +104,7 @@ MiniNuke:
 		Damage: 60
 		Falloff: 1000, 368, 135, 50, 18, 7, 0
 		Delay: 10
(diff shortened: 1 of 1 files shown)


reaperrr:
Makes sense, in my opinion :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/6ce87dc3ff99af6f828919bcd45bde6b15cc7bab...9a0cd3b97b5963cd8b281ebc80c8cb002de9983f)

Issue #11224, opened 2016-05-05
Harvester self-healing to 50%


HenrytheSlav:
After a quick discussion in IRC where there were mostly possitive responses to this idea, I create this ticket to track the disscusion about the possibility to implement original harvester self-healing.

There are several factors to consider, especially balance in early rushing tactics and the speed of heal, which was quite slow in original. Please, feel free to elaborate your opinon on this idea.

Micr0Bit:
if this is related to redalert ... we have repair-pads and mechanics ... you can even repair a harvester-husk to get a destroyed one back ... (the soviet counter-part to mechanics is the hijacker which "surprise surprise" is mostly good at stealing slow vehicles like harvesters , he still needs to get more polished though) 

D2k has this feature for harvesters ... good for D2k 

and this isnt balance-related at all ...since there is no issue with harvesters that could justify such a "buff"

HenrytheSlav:
The argument for is authenticity, the argument against would be balancing, at least that is how I see it. But yeah, this is just to gather opinions on this matter, since I can't find any mention of this being even considered.

SoScared:
If it's a slow heal then I don't see why not. It was so in the original game and as mentioned it won't affect balance too much. As a convenience I think casual players will appreciate not having to micro every low health harvester they've got.

AoAGeneral:
Is this for RA, TD, or both?

HenrytheSlav:
I meant RA, since I don't know if TD had this.
http://cnc.wikia.com/wiki/Harvester_%28Tiberian_Dawn%29 doesn't meniton it, while RA version has self-healing ability included.

Lukemtesta:
I agree with @SoScared - If it was by design in the original a slow heal is unlikely to unbalance the game - The heal should probably be significantly slower than the average rate a unit can chase and damage one continuously

Townes27:
The harvester in Red Alert 1 had the same selfheal property as the mammoth tank.

abcdefg30:
I'd be for a slow regeneration rate up to 50% health for harvesters in all mods.

AoAGeneral:
In regards to the regeneration effect is it possible to allow harvesters the regeneration if a silo is built?

HenrytheSlav:
[...]

PR #11254, opened 2016-05-10
Add slow selfealing to the harvester (RA mod)

HenrytheSlav:
Realization of #11224 due general acceptance, for RA mod atm, if people agree I can push commits for other mods.

While most people argued that heal should be at Mamooth Tank rate, I found that to be stupidly high (harv is attacked by group of infantry, goes back to ref and is healed almost to 50% when it gets back). I set it to quite slow, feeling that while it decreases unnecessary macro, it also motivates player to play safe and reapir it to full health fast (for cash).

### DIFF
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -275,6 +275,11 @@ HARV:
 	HarvesterHuskModifier:
 		FullActor: HARV.FullHusk
 		FullnessThreshold: 50
+	SelfHealing:
+		Step: 1
+		Delay: 25
+		HealIfBelow: 50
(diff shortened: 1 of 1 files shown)


reaperrr:
At first glance the repair rate looked a bit slow to me, but on the other hand the average repair rate is still twice that of elite units.

Reasoning makes sense, too. :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8c4b091296a959033cc41ad9564ec38c266a36d0...6ce87dc3ff99af6f828919bcd45bde6b15cc7bab)

Issue #10846, opened 2016-02-27
You can just undeploy your MCV to save it from C4.


epicelite:
It should still blow up anyway, obvious bug.

EDIT: or sell any building.

AMHOL:
I was thinking about this, even when you get the "flash" from Tanya C4 they can still un-deploy and save it, I think players should be rewarded for noticing and un-deploying, I was thinking maybe the MCV could explode with some burst damage, enough to seriously harm or kill the tanya and maybe cause a bit of damage to nearby structures, maybe getting a power-down or something.

PR #11257, opened 2016-05-10
Buildings get locked by Demolishable activity before their destruction

HenrytheSlav:
Solves #10846 ! Made mostly by @abcdefg30 so love (or hate) him, I only debugged the last part. :tophat:

### DIFF
--- a/OpenRA.Mods.Common/Activities/Demolish.cs
+++ b/OpenRA.Mods.Common/Activities/Demolish.cs
@@ -57,6 +57,10 @@ namespace OpenRA.Mods.Common.Activities
 				if (cloak != null && cloak.Info.UncloakOn.HasFlag(UncloakType.Demolish))
 					cloak.Uncloak();
 
+				var building = target.TraitOrDefault<Building>();
+				if (building != null)
+					building.Lock();
+
 				for (var f = 0; f < flashes; f++)
 					w.Add(new DelayedAction(flashesDelay + f * flashInterval, () =>
 						w.Add(new FlashTarget(target, ticks: flashDuration))));


OpenRA.Mods.Common/Activities/Demolish.cs
```diff
@@ -57,6 +58,14 @@ protected override void OnInside(Actor self)
 				if (cloak != null && cloak.Info.UncloakOn.HasFlag(UncloakType.Demolish))
 					cloak.Uncloak();
 
+				if (!chargePlanted)
+				{
+				chargePlanted = true;
+				var building = target.TraitOrDefault<Building>();
+				if (building != null)
+					building.Lock();
```
reaperrr: All four lines within the brackets need one additional tab indentation.

HenrytheSlav:
Sorry for that, updated with correct indentation.

obrakmann:
works as advertised, :+1:

OpenRA.Mods.Common/Activities/Demolish.cs
```diff
@@ -57,6 +58,14 @@ protected override void OnInside(Actor self)
 				if (cloak != null && cloak.Info.UncloakOn.HasFlag(UncloakType.Demolish))
 					cloak.Uncloak();
 
+				if (!chargePlanted)
+				{
+					chargePlanted = true;
```
Mailaender: Why is this required?

GraionDilach: There couléd be multiple actors entering to the same target (D2k Saboteurs came to my mind) in which case multiple `Lock`s might be errandeous.

pchote: This particular code only ever runs once, so the check does nothing.

epicelite:
I approve!

OpenRA.Mods.Common/Activities/Demolish.cs
HenrytheSlav: Sorry I thought this runs on every tick like external capturing which I took as template. I will fix that.

HenrytheSlav:
Rewritten, works, but idk if it is okay with coding style etc.

GraionDilach:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/11e324f36332bc3e801fda533ff130e2b6fa0e89...362a05194bfbccd3f6f4dc58ee82e3b80e061706)

PR #11276, opened 2016-05-14
Fix AutoTarget handling of multiple/upgradable Attack* traits

### LINKED ISSUE #11258: Tick Tank not attacking properly
Tick tank don't focus enemy units when they are in range and A-move just make him move.

reaperrr:
Fixes #11258.

### DIFF
--- a/OpenRA.Mods.Common/Traits/AutoTarget.cs
+++ b/OpenRA.Mods.Common/Traits/AutoTarget.cs
@@ -53,7 +53,7 @@ namespace OpenRA.Mods.Common.Traits
 	public class AutoTarget : UpgradableTrait<AutoTargetInfo>, INotifyIdle, INotifyDamage, ITick, IResolveOrder, ISync
 	{
 		readonly AttackBase attack;
-		readonly AttackFollow at;
+		readonly AttackFollow attackFollow;
 		[Sync] int nextScanTime = 0;
 
 		public UnitStance Stance;
@@ -75,7 +75,7 @@ namespace OpenRA.Mods.Common.Traits
 				Stance = self.Owner.IsBot || !self.Owner.Playable ? info.InitialStanceAI : info.InitialStance;
 
 			PredictedStance = Stance;
-			at = self.TraitOrDefault<AttackFollow>();
+			attackFollow = self.TraitsImplementing<AttackFollow>().FirstOrDefault(Exts.IsTraitEnabled);
 		}
 
 		public void ResolveOrder(Actor self, Order order)
@@ -117,8 +117,8 @@ namespace OpenRA.Mods.Common.Traits
 				return;
 
 			Aggressor = attacker;
-			if (at == null || !at.IsReachableTarget(at.Target, allowMove))
 			var allowMove = Info.AllowMovement && Stance != UnitStance.Defend;
+			if (attackFollow == null || !attackFollow.IsReachableTarget(attackFollow.Target, allowMove))
 				Attack(self, Aggressor, allowMove);
 		}
 
@@ -130,8 +130,8 @@ namespace OpenRA.Mods.Common.Traits
 			if (Stance < UnitStance.Defend || !Info.TargetWhenIdle)
 				return;
 
-			if (at == null || !at.IsReachableTarget(at.Target, allowMove))
 			var allowMove = Info.AllowMovement && Stance != UnitStance.Defend;
+			if (attackFollow == null || !attackFollow.IsReachableTarget(attackFollow.Target, allowMove))
 				ScanAndAttack(self, allowMove);
 		}
 
--- a/OpenRA.Mods.Common/Traits/AutoTarget.cs
+++ b/OpenRA.Mods.Common/Traits/AutoTarget.cs
@@ -53,7 +53,7 @@ namespace OpenRA.Mods.Common.Traits
 	public class AutoTarget : UpgradableTrait<AutoTargetInfo>, INotifyIdle, INotifyDamage, ITick, IResolveOrder, ISync
 	{
 		readonly AttackBase[] attackBases;
-		readonly AttackFollow attackFollow;
+		readonly AttackFollow[] attackFollows;
 		[Sync] int nextScanTime = 0;
 
 		public UnitStance Stance;
@@ -75,7 +75,7 @@ namespace OpenRA.Mods.Common.Traits
 				Stance = self.Owner.IsBot || !self.Owner.Playable ? info.InitialStanceAI : info.InitialStance;
 
 			PredictedStance = Stance;
-			attackFollow = self.TraitsImplementing<AttackFollow>().FirstOrDefault(Exts.IsTraitEnabled);
+			attackFollows = self.TraitsImplementing<AttackFollow>().ToArray();
 		}
 
 		public void ResolveOrder(Actor self, Order order)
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -103,7 +105,7 @@ public void Damaged(Actor self, AttackInfo e)
 			}
 
 			// not a lot we can do about things we can't hurt... although maybe we should automatically run away?
-			if (!attack.HasAnyValidWeapons(Target.FromActor(attacker)))
+			if (!attackBases.Any(Exts.IsTraitEnabled) || !attackBases.First(Exts.IsTraitEnabled).HasAnyValidWeapons(Target.FromActor(attacker)))
```
pchote: I don't like this recent trend of "upgrading" from "requires only one" to "supports many, but breaks in unexpected ways if you have multiple active at the same time".  Can we please check over all active traits and *never* use First/FirstOrDefault?

reaperrr: Yeah, good point.

reaperrr:
Updated, but I'm not entirely confident in how I did it so review carefully.

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -103,7 +105,7 @@ public void Damaged(Actor self, AttackInfo e)
 			}
 
 			// not a lot we can do about things we can't hurt... although maybe we should automatically run away?
-			if (!attack.HasAnyValidWeapons(Target.FromActor(attacker)))
+			if (!attackBases.Any(Exts.IsTraitEnabled) || !attackBases.Where(a => a.HasAnyValidWeapons(Target.FromActor(attacker))).Any(Exts.IsTraitEnabled))
```
pchote: This simplifies to `!attackbases.Any(a => !a.IsTraitDisabled && a.HasAnyValidWeapons(Target.FromActor(attacker)))`

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -115,39 +117,53 @@ public void Damaged(Actor self, AttackInfo e)
 				return;
 
 			Aggressor = attacker;
-			var allowMove = info.AllowMovement && Stance != UnitStance.Defend;
-			if (at == null || !at.IsReachableTarget(at.Target, allowMove))
+			var allowMove = Info.AllowMovement && Stance != UnitStance.Defend;
+			if (!attackFollows.Any() || !attackFollows.Any(Exts.IsTraitEnabled)
```
pchote: These three can be combined in the same way as the one above.

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -115,39 +117,53 @@ public void Damaged(Actor self, AttackInfo e)
...
+			if (IsTraitDisabled)
 				return;
 
-			var allowMove = info.AllowMovement && Stance != UnitStance.Defend;
-			if (at == null || !at.IsReachableTarget(at.Target, allowMove))
+			if (Stance < UnitStance.Defend || !Info.TargetWhenIdle)
+				return;
+
+			var allowMove = Info.AllowMovement && Stance != UnitStance.Defend;
+			if (!attackFollows.Any() || !attackFollows.Any(Exts.IsTraitEnabled)
```
pchote: This one too.

reaperrr:
Updated.

Mailaender:

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -103,7 +105,7 @@ public void Damaged(Actor self, AttackInfo e)
 			}
 
 			// not a lot we can do about things we can't hurt... although maybe we should automatically run away?
-			if (!attack.HasAnyValidWeapons(Target.FromActor(attacker)))
+			if (!attackBases.Any(a => !a.IsTraitDisabled && a.HasAnyValidWeapons(Target.FromActor(attacker))))
```
pchote: Minor style nit: these might be easier to read as `if (attackBases.All(a => a.IsTraitDisabled || !a.HasAnyValidWeapons(Target.FromActor(attacker))))` (removing the double negatives).

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -166,10 +180,13 @@ void Attack(Actor self, Actor targetActor, bool allowMove)
 			TargetedActor = targetActor;
 			var target = Target.FromActor(targetActor);
 			self.SetTargetLine(target, Color.Red, false);
-			attack.AttackTarget(target, false, allowMove);
+
+			foreach (var ab in attackBases)
+				if (!ab.IsTraitDisabled)
```
pchote: Minor style nit: above you use `attackBases.Where(Exts.IsTraitEnabled)` and here you check manually.  It would be nice if you could make these consistent.

pchote:
The code changes look fine aside from those very minor nits.  I haven't tested the changes, however.

reaperrr:
Adressed style nits.

Marnador:
Tested on the latest bleed TS, seem to get `System.NullReferenceException` after attempting to move basic infantry on a Tiberium field and attempting selecting it over and over, right before the cell where the infantry is supposed to stop.

reaperrr:
@Marnador do you still have the full exception? If yes, could you upload or post it somewhere?

Marnador:
@reaperrr Sure, here is the whole exception. https://www.dropbox.com/s/yvvjzbt8nif87kg/exception.log?dl=0

reaperrr:
Unless I'm missing something here, neither the steps to reproduce it nor the exception look like they're caused by this PR.
It rather looks like the `DrawLineToTarget` code can run into a NRE if the targeted actor dies on the same tick you clicked at it, but that would mean it's an issue on the bleed branch already.

Marnador:
@reaperrr I am not sure what the problem is, but with a naked bleed I tried over and over and failed to replicate the issue. Yet, when I implement the change I can replicate the issue pretty reliably.
What I do is build a single basic infantry and send it on the Tiberium field and right before it stops spam [Q]-(select all) as fast as possible(that happens while the soldier is still alive). Anyone is welcome to try, I am curious if anyone is able to replicate it as well.

reaperrr:
Actually you're right, I just managed to reproduce it. No idea about the cause yet, though.

OpenRA.Mods.Common/Traits/AutoTarget.cs
```diff
@@ -103,7 +105,7 @@ public void Damaged(Actor self, AttackInfo e)
 			}
 
 			// not a lot we can do about things we can't hurt... although maybe we should automatically run away?
-			if (!attack.HasAnyValidWeapons(Target.FromActor(attacker)))
+			if (attackBases.All(a => a.IsTraitDisabled && !a.HasAnyValidWeapons(Target.FromActor(attacker))))
```
reaperrr: Ouch. The `&&` needs to be a `||`, of course.

reaperrr:
Updated. @Marnador Could you re-test this?

Marnador:
@reaperrr Tested for quite a bit, failed to find any issues so far. Works pretty well with TickTanks, been looking forward to that fix for quite some time now :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/cf438bf3c2647ff617132dfb8424b79ed6783a10...00693c94d7b29afd998ed86875162239eec715e3)

Issue #7454, opened 2015-02-09
Combat caused by Lua actor command clears entire queue


Computerfreaked:
When doing for example:

```
unit.Hunt()
unit.AttackMove(waypoint7.Location)
unit.AttackMove(waypoint94.Location)
```

Or another example:

```
unit.AttackMove(waypoint7.Location)
unit.Hunt()
```

(where unit is a valid actor)

If the unit gets in a firefight but does not die it goes idle after the fight. It does not complete the action it was doing or any subsequent actions on the queue. It looks like the whole queue is emptied when a single bullet is fired upon the unit or the unit fires itself.
##

Bynnar18:
I've run into this scripting the TD shellmap. It's really annoying. I would fix it myself if I thought I had the experience to.

Computerfreaked:
> So something like this will cause the unit to stop even if it hasn't reached waypoint1:

>     Trigger.OnIdle(unit, function() unit.AttackMove(waypoint1.Location) end)

This is what I thought a solution was and I just tried it. It works :D

obrakmann:
Found the problem. We were only creating an `AttackMoveActivity`, but didn't set the `TargetLocation` in the `AttackMove` trait.

Bynnar18:
>This is what I thought a solution was and I just tried it. It works :D
>
>Nevertheless it is still a workaround if you ask me.

I may of been mistaken, or maybe it just takes some time for it to trigger but I didn't see it work when I tried it.

Unit158:
@obrakmann any progress on this?

obrakmann:
Not really. We can't just set the `TargetLocation` and run the activity, because that causes the `AttackMove` trait do wake up and go off to do its own thing if the unit ever goes idle (and then it would go *back* to the `TargetLocation`, which may be on the other side of the map at that point. We can't unset `TargetLocation` either at the end of the activity, because that would break `AttackMove`. You can't even unset it in the `Cancel` method, because that breaks `AttackMove` as well. And we also can't unset `TargetLocation` in a `CallFunc` or sth. similar, because that might never execute if the activity queue is canceled.

I've tried re-writing activities to allow that sort of thing, but that turned out to be fruitless in the end due to the way that the `Attack` activity in `AttackMoveActivity` is generated (ie. not directly, but three or four method calls deeper into the call chain, where you cannot distinguish between the different use-cases anymore).

[...]

PR #11283, opened 2016-05-15
Add nod08b

ghost:
Add nod08b - "Battle for Zaire (b)"
http://cnc.wikia.com/wiki/Battle_for_Zaire

**Known issue:**
Boats turn into idle actors and stop their patrol, each time they attack an unit.
Is there a way to avoid turning them into idle actors?


Mailaender:
I believe the gunboat stopping during attacks is related to https://github.com/OpenRA/OpenRA/issues/3288.

obrakmann:
The patrol thing is actually related to https://github.com/OpenRA/OpenRA/issues/7454#issuecomment-107056625. The way `AttackMove` works right now is absolutely moronic, and needs a complete re-write of `Activity` to be fixed properly.

Mailaender:
Actually for me the boat turned around when it was attacked, but I don't think you should try to fix it during this pull request. The only nit I have with the boats is that you can destroy them in this mission, but they don't define an explosion on death.

Biofreak1987:
it's turning around as it's wrapped inside a Trigger.OnIdle, which tells it to patrol (The first waypoint is located on the left). It would stop moving without that one.

HenrytheSlav:
Idk why GDI ships have no naming, I see "Enemy " (the blank space is intentional.
campaign-tooltips.yaml defines ships and Boat inherits ^ship. Is there something more to this?

I assume the camera/visibility of boats is intentional and taken from original. But maybe it would be cool to turn it on when you build Communications centre? It coould in theory be secondary tasks (Build commcentre to track enemy naval support). Nod player has almost no reason to build commcentre since he gets no airstrikes.

Another thing to consider is that you get flame troops to build at hand of nod, which allows you kinda sneakattack at start using chinnok (also in (a) ).

Otherwise seems verya well made.

Mailaender:
> Should I make them invincible or do you plan on adding an explosion for this actor type?

They are already quite sturdy. Making them invincible would be overkill. Just add an explosion type.

HenrytheSlav:
> Just add an explosion type.

Maybe even define that generally? In separate commit in this PR. I guess it doesn't hurt to have.

Biofreak1987:
Added an explosion to the boat.
Revealing GDI ships when you have a communications center sounds interesting... feel free to create a separate commit for that.

HenrytheSlav:
I actually meant to make the gunboat explosion general. (in original yaml)

Biofreak1987:
Reverted the last change seeing as boats get used in nod09, nod11a/b and nod13a/b/c as well. I rather change it with a different merge request by updating the ship.yaml and giving it a explosion similar to red alert ships.

HenrytheSlav:
Could you please also solve that they have wrong name? At least I saw tooltip with "Enemy ". That should be done together.

Mailaender:
We can do it in a separate pull request. That is true. The mission itself works fine. :white_check_mark:

HenrytheSlav:
Could you fix this? Preferably globally like the explosion? Otherwise :+1:
openra-2016-05-16t164742z

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/caf0f75a7345de7c7cf88fa12210aac7eb46bd36...4ca1c0b136ac7451632c67df95cbda4f43a85de5)

PR #11275, opened 2016-05-14
Add libmono-i18n4.0-all to the list of Debian package dependencies

obrakmann:
Fixes #11256 for real, hopefully.

PR #11284, opened 2016-05-15
Updated to .NET version 4.5

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/10661.

### DIFF
--- a/packaging/linux/deb/DEBIAN/control
+++ b/packaging/linux/deb/DEBIAN/control
@@ -3,7 +3,7 @@ Version: {VERSION}
 Architecture: all
 Maintainer: Paul Chote <paul@chote.net>
 Installed-Size: {SIZE}
-Depends: libopenal1, mono-runtime (>= 2.10), libmono-system-core4.0-cil, libmono-system-drawing4.0-cil, libmono-system-data4.0-cil, libmono-system-numerics4.0-cil, libmono-system-runtime-serialization4.0-cil, libmono-system-xml-linq4.0-cil, libmono-i18n4.0-all, libfreetype6, libc6, libasound2, libgl1-mesa-glx, libgl1-mesa-dri, xdg-utils, zenity, libsdl2 | libsdl2-2.0-0, liblua5.1-0
+Depends: libopenal1, mono-runtime (>= 3.2), libmono-system-core4.0-cil, libmono-system-drawing4.0-cil, libmono-system-data4.0-cil, libmono-system-numerics4.0-cil, libmono-system-runtime-serialization4.0-cil, libmono-system-xml-linq4.0-cil, libmono-i18n4.0-all, libfreetype6, libc6, libasound2, libgl1-mesa-glx, libgl1-mesa-dri, xdg-utils, zenity, libsdl2 | libsdl2-2.0-0, liblua5.1-0
 Section: games
 Priority: extra
 Homepage: http://www.openra.net/
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ env:
 # Run the NUnit tests
 script:
  - travis_retry make all-dependencies
- - make all SDK="-sdk:4.0"
+ - make all SDK="-sdk:4.5"
  - make check
  - make check-scripts
  - make test
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -7,7 +7,7 @@ Windows
 =======
 
 * [Windows PowerShell >= 4.0](http://microsoft.com/powershell)
-* [.NET Framework >= 4.0 (Client Profile)](http://www.microsoft.com/en-us/download/details.aspx?id=17113)
+* [.NET Framework >= 4.5 (Client Profile)](http://www.microsoft.com/en-us/download/details.aspx?id=30653)
 * [SDL 2](http://www.libsdl.org/download-2.0.php) (included)
 * [FreeType](http://gnuwin32.sourceforge.net/packages/freetype.htm) (included)
 * [zlib](http://gnuwin32.sourceforge.net/packages/zlib.htm) (included)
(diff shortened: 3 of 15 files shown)


packaging/linux/deb/DEBIAN/control
```diff
@@ -3,7 +3,7 @@ Version: {VERSION}
 Architecture: all
 Maintainer: Paul Chote <paul@chote.net>
 Installed-Size: {SIZE}
-Depends: libopenal1, mono-runtime (>= 2.10), libmono-system-core4.0-cil, libmono-system-drawing4.0-cil, libmono-system-data4.0-cil, libmono-system-numerics4.0-cil, libmono-system-runtime-serialization4.0-cil, libmono-system-xml-linq4.0-cil, libfreetype6, libc6, libasound2, libgl1-mesa-glx, libgl1-mesa-dri, xdg-utils, zenity, libsdl2 | libsdl2-2.0-0, liblua5.1-0
+Depends: libopenal1, mono-runtime (>= 3.2), libmono-system-core4.0-cil, libmono-system-drawing4.0-cil, libmono-system-data4.0-cil, libmono-system-numerics4.0-cil, libmono-system-runtime-serialization4.0-cil, libmono-system-xml-linq4.0-cil, libfreetype6, libc6, libasound2, libgl1-mesa-glx, libgl1-mesa-dri, xdg-utils, zenity, libsdl2 | libsdl2-2.0-0, liblua5.1-0
```
obrakmann: This is already in #11275. Shall I drop it from there? It makes more sense here, contextually.

Mailaender: Yes, indeed.

obrakmann: Done.

Phrohdoh:
The minimum version on [the installation page](https://github.com/OpenRA/OpenRA/wiki/Installation) will need to be updated if/once this is merged.

Mailaender:
This is currently waiting for https://github.com/OpenRA/OpenRA/pull/11275 to get merged first.

reaperrr:
Needs rebase.

Mailaender:
Rebased.

Phrohdoh:
I can build this via the Makefile (mono 4.2.3.4) and `xbuild` (v12.0).
The game runs fine (tested the map editor, and asset viewer in all default mods).

Some things were missed:
`make.ps1:3:     $msBuildVersions = @("4.0")`
`OpenRA.Server/OpenRA.Server.csproj:61:      <RequiredTargetFramework>4.0</RequiredTargetFramework>`

Do any of the `libmono-*4.0-cil` references need to be updated? (@obrakmann)

Other than those fixes, 👍.
We're on the fast track to newer tooling. :-)

Mailaender:
Updated.

make.ps1
```diff
@@ -1,6 +1,6 @@
 function FindMSBuild
 {
-	$msBuildVersions = @("4.0")
+	$msBuildVersions = @("4.5")
```
reaperrr: Nope, this needs to be reverted. Internally, 4.5 is still a sub-version of 4.0, Windows won't find MSBuild if you change this.

reaperrr:
At least on my system, I get lots of CPU mismatch warnings. MSBuild defaults to Framework64, but we target x86.

The compiled code seems to work fine anyhow, but these warnings at least slow down the compile quite a bit. Not sure how to fix that, though, so I won't block the PR over this.

Maybe @penev92, @GraionDilach or @RoosterDragon have an idea how to fix this?

GraionDilach:
All the compilations I've seen on x64 Windows had those warnings - I'm actually surprised this PR introduces it to you. I'm aware because I accidentally had `TreatWarningsAsErrors: true` in the AS project file and `make all` ended up failing at 90% of my testers with everyone reporting the compilation warnings as the reason.

Mind you, I'm still on x86 myself so I probably couldn't reproduce it.

RoosterDragon:
You're probably seeing the same issue as in #9114.

GraionDilach:
Correct me if I'm wrong but all our projects are compiled to be x86, aren't they? Do we even use nonx86 libraries at all?

I'm thinking about silencing the warnings with a 
```
	<propertyGroup>
		<ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
			None
		</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
	</propertyGroup>
```
configuration in the project files, but it might be a bad idea.

make.ps1
Mailaender: Reverted.

RoosterDragon:
See https://github.com/OpenRA/OpenRA/issues/9114#issuecomment-172318063. If you're seeing MSB3644 in the build logs, then it can't find the reference assemblies on your machine and it's just going to go hunting in the GAC for whatever. It'll probably find the 64-bit assemblies and use those, hence the platform mismatch.

It's a build problem local to your machines - rather than any problem with the actual build config. It seems the reference assemblies need to be installed - I'm not able to suggest how to do that easily as it's always worked on my machine (I assume VS installs it).

I don't get any build problems with this 4.5 config - it seems fine :)

reaperrr:
Considering @RoosterDragon's comment and the fact that - apart from the warnings - this hasn't caused any real problems for me or @GraionDilach so far, I guess I can :+1: this.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/8f3546613d13955750a52359ecb9c124309d448c...c454af8068dd2c735f7989c49f251e992379d87b)

reaperrr:
@GraionDilach I installed .NET 4.6.1 from the Windows Updater yesterday and the warnings are gone. If you haven't yet, updating might fix it for you, too.

PR #11286, opened 2016-05-15
Dump Mono.NAT in favor of Open.Nat 2.0.16

### LINKED ISSUE #7140: Migrate to Open.NAT and replace Mono.Nat

### LINKED ISSUE #8794: UPNP XML parsing crash.

Mailaender:
Closes #7140.
Fixes #8794.

Depends on #11284. I can try the .NET 3.5 variant, but I don't want to as we will then need to ship our own version of `System.Threading` DLLs, which will cause a lot of unnecessary bloat.

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -18,6 +18,7 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Threading;
+using System.Threading.Tasks;
 using OpenRA.Chat;
 using OpenRA.Graphics;
 using OpenRA.Network;
@@ -50,6 +51,8 @@ namespace OpenRA
 		public static bool BenchmarkMode = false;
 
 		public static GlobalChat GlobalChat;
+
+		static Task discoverNat;
 
 		public static OrderManager JoinServer(string host, int port, string password, bool recordReplay = true)
 		{
@@ -247,16 +250,7 @@ namespace OpenRA
 			Log.AddChannel("graphics", "graphics.log");
 			Log.AddChannel("geoip", "geoip.log");
 			Log.AddChannel("irc", "irc.log");
-
-			if (Settings.Server.DiscoverNatDevices)
-				UPnP.TryNatDiscovery();
-			else
-			{
-				Settings.Server.NatDeviceAvailable = false;
-				Settings.Server.AllowPortForward = false;
-			}
-
-			GeoIP.Initialize();
+			Log.AddChannel("nat", "nat.log");
 
 			var renderers = new[] { Settings.Graphics.Renderer, "Default", null };
 			foreach (var r in renderers)
@@ -277,6 +271,16 @@ namespace OpenRA
 				}
 			}
 
+			GeoIP.Initialize();
+
+			if (!Game.Settings.Server.DiscoverNatDevices)
+				Game.Settings.Server.AllowPortForward = false;
+			else
+			{
+				discoverNat = UPnP.DiscoverNatDevices(Settings.Server.NatDiscoveryTimeout);
+				Game.Settings.Server.AllowPortForward = true;
+			}
+
 			Sound = new Sound(Settings.Sound.Engine);
 
 			GlobalChat = new GlobalChat();
@@ -286,9 +290,6 @@ namespace OpenRA
 				Console.WriteLine("\t{0}: {1} ({2})", mod.Key, mod.Value.Title, mod.Value.Version);
 
 			InitializeMod(Settings.Game.Mod, args);
-
-			if (Settings.Server.DiscoverNatDevices)
-				RunAfterDelay(Settings.Server.NatDiscoveryTimeout, UPnP.StoppingNatDiscovery);
 		}
 
 		public static bool IsModInstalled(string modId)
@@ -391,6 +392,18 @@ namespace OpenRA
 			PerfHistory.Items["render_flip"].HasNormalTick = false;
 
 			JoinLocal();
+
+			try
+			{
+				if (discoverNat != null)
+					discoverNat.Wait();
+			}
+			catch (Exception e)
+			{
+				Console.WriteLine("NAT discovery failed: {0}", e.Message);
+				Log.Write("nat", e.ToString());
+				Game.Settings.Server.AllowPortForward = false;
+			}
 
 			ModData.LoadScreen.StartGame(args);
 		}
--- a/OpenRA.Game/Network/UPnP.cs
+++ b/OpenRA.Game/Network/UPnP.cs
@@ -10,127 +10,67 @@
 #endregion
 
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Net;
-using Mono.Nat;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Open.Nat;
 
 namespace OpenRA.Network
 {
-	public static class UPnP
+	public class UPnP
 	{
-		static INatDevice natDevice;
+		static NatDevice natDevice;
+		static Mapping mapping;
 
-		public static void TryNatDiscovery()
+		public static IPAddress ExternalIP { get; private set; }
+
+		public static async Task DiscoverNatDevices(int timeout)
 		{
+			NatDiscoverer.TraceSource.Switch.Level = SourceLevels.Verbose;
+			var logChannel = Log.Channel("nat");
+			NatDiscoverer.TraceSource.Listeners.Add(new TextWriterTraceListener(logChannel.Writer));
+
+			var natDiscoverer = new NatDiscoverer();
+			var token = new CancellationTokenSource(timeout);
+			natDevice = await natDiscoverer.DiscoverDeviceAsync(PortMapper.Upnp, token);
 			try
 			{
-				NatUtility.Logger = Log.Channel("server").Writer;
-				NatUtility.Verbose = Game.Settings.Server.VerboseNatDiscovery;
-				NatUtility.DeviceFound += DeviceFound;
-				Game.Settings.Server.NatDeviceAvailable = false;
-				NatUtility.StartDiscovery();
-				Log.Write("server", "NAT discovery started.");
+				ExternalIP = await natDevice.GetExternalIPAsync();
 			}
 			catch (Exception e)
 			{
-				Log.Write("server", "Can't discover UPnP-enabled device: {0}", e);
-				Game.Settings.Server.NatDeviceAvailable = false;
-				Game.Settings.Server.AllowPortForward = false;
+				Console.WriteLine("Getting the external IP from NAT device failed: {0}", e.Message);
+				Log.Write("nat", e.StackTrace);
 			}
 		}
 
-		public static void StoppingNatDiscovery()
+		public static async Task ForwardPort(int listen, int external)
 		{
-			Log.Write("server", "Stopping NAT discovery.");
-			NatUtility.StopDiscovery();
-
-			if (natDevice == null || natDevice.GetType() != typeof(Mono.Nat.Upnp.UpnpNatDevice))
-			{
-				Log.Write("server",
-					"No NAT devices with UPnP enabled found within {0} ms deadline. Disabling automatic port forwarding.".F(Game.Settings.Server.NatDiscoveryTimeout));
-				Game.Settings.Server.NatDeviceAvailable = false;
-				Game.Settings.Server.AllowPortForward = false;
-			}
-		}
-
-		public static void DeviceFound(object sender, DeviceEventArgs args)
-		{
-			Log.Write("server", "NAT device discovered.");
-
-			Game.Settings.Server.NatDeviceAvailable = true;
-			Game.Settings.Server.AllowPortForward = true;
-
+			mapping = new Mapping(Protocol.Tcp, listen, external, "OpenRA");
 			try
 			{
-				natDevice = args.Device;
-				Log.Write("server", "Type: {0}", natDevice.GetType());
-				Log.Write("server", "Your external IP is: {0}", natDevice.GetExternalIP());
-
-				foreach (var mp in natDevice.GetAllMappings())
-					Log.Write("server", "Existing port mapping: protocol={0}, public={1}, private={2}",
-						mp.Protocol, mp.PublicPort, mp.PrivatePort);
+				await natDevice.CreatePortMapAsync(mapping);
 			}
 			catch (Exception e)
 			{
-				Log.Write("server", "Can't fetch information from NAT device: {0}", e);
-
-				Game.Settings.Server.NatDeviceAvailable = false;
-				Game.Settings.Server.AllowPortForward = false;
+				Console.WriteLine("Port forwarding failed: {0}", e.Message);
+				Log.Write("nat", e.StackTrace);
 			}
 		}
 
-		public static void ForwardPort(int lifetime)
+		public static async Task RemovePortForward()
 		{
 			try
 			{
-				var mapping = new Mapping(Protocol.Tcp, Game.Settings.Server.ExternalPort, Game.Settings.Server.ListenPort, lifetime);
-				natDevice.CreatePortMap(mapping);
-				Log.Write("server", "Create port mapping: protocol = {0}, public = {1}, private = {2}, lifetime = {3} s",
-					mapping.Protocol, mapping.PublicPort, mapping.PrivatePort, mapping.Lifetime);
-			}
-			catch (MappingException e)
-			{
-				if (e.ErrorCode == 725 && lifetime != 0)
-				{
-					Log.Write("server", "NAT device answered with OnlyPermanentLeasesSupported. Retrying...");
-					ForwardPort(0);
-				}
-				else
-				{
-					Log.Write("server", "Can not forward ports via UPnP: {0}", e);
-					Game.Settings.Server.AllowPortForward = false;
-				}
-			}
-		}
-
-		public static void RemovePortforward()
-		{
-			try
-			{
-				var mapping = new Mapping(Protocol.Tcp, Game.Settings.Server.ExternalPort, Game.Settings.Server.ListenPort);
-				natDevice.DeletePortMap(mapping);
-				Log.Write("server", "Remove port mapping: protocol = {0}, public = {1}, private = {2}, expiration = {3}",
-					mapping.Protocol, mapping.PublicPort, mapping.PrivatePort, mapping.Expiration);
+				await natDevice.DeletePortMapAsync(mapping);
 			}
 			catch (Exception e)
 			{
-				Log.Write("server", "Can not remove UPnP portforwarding rules: {0}", e);
-				Game.Settings.Server.AllowPortForward = false;
-			}
-		}
-
-		public static IPAddress GetExternalIP()
-		{
-			if (natDevice == null)
-				return null;
-
-			try
-			{
-				return natDevice.GetExternalIP();
-			}
-			catch (Exception e)
-			{
-				Log.Write("server", "Failed to get the external IP from NAT device: {0}", e);
-				return null;
+				Console.WriteLine("Port removal failed: {0}", e.Message);
+				Log.Write("nat", e.StackTrace);
 			}
 		}
 	}
--- a/OpenRA.Game/Server/Server.cs
+++ b/OpenRA.Game/Server/Server.cs
@@ -137,7 +137,7 @@ namespace OpenRA.Server
 			randomSeed = (int)DateTime.Now.ToBinary();
 
 			if (Settings.AllowPortForward)
-				UPnP.ForwardPort(3600);
+				UPnP.ForwardPort(Settings.ListenPort, Settings.ExternalPort).Wait();
 
 			foreach (var trait in modData.Manifest.ServerTraits)
 				serverTraits.Add(modData.ObjectCreator.CreateObject<ServerTrait>(trait));
@@ -202,7 +202,8 @@ namespace OpenRA.Server
 					if (State == ServerState.ShuttingDown)
 					{
 						EndGame();
-						if (Settings.AllowPortForward) UPnP.RemovePortforward();
+						if (Settings.AllowPortForward)
+							UPnP.RemovePortForward().Wait();
 						break;
 					}
 				}
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ SDK         ?=
 CSC         = mcs $(SDK)
 CSFLAGS     = -nologo -warn:4 -codepage:utf8 -unsafe -warnaserror
 DEFINE      = TRACE
-COMMON_LIBS = System.dll System.Core.dll System.Data.dll System.Data.DataSetExtensions.dll System.Drawing.dll System.Xml.dll thirdparty/download/ICSharpCode.SharpZipLib.dll thirdparty/download/FuzzyLogicLibrary.dll thirdparty/download/Mono.Nat.dll thirdparty/download/MaxMind.Db.dll thirdparty/download/MaxMind.GeoIP2.dll thirdparty/download/Eluant.dll thirdparty/download/SmarIrc4net.dll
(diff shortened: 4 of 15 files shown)


OpenRA.Game/Network/UPnP.cs
```diff
@@ -10,127 +10,85 @@
...
+				Console.WriteLine("NAT discovery failed: {0}", e.Message);
+				Log.Write("nat", e.StackTrace);
 				Game.Settings.Server.AllowPortForward = false;
 			}
+
+			Game.Settings.Server.AllowPortForward = true;
 		}
 
-		public static void ForwardPort(int lifetime)
+		public static async void ForwardPort()
```
pchote: Please pass in the ports as arguments instead of reading them from `Game.Settings`.

Passing data around via global state is poor practice, and is why our server and utility classes are full of `// HACK: The engine code assumes that Game.Settings is set.` and `// HACK: The engine code assumes that Game.modData is set.`

OpenRA.Game/Network/UPnP.cs
```diff
@@ -10,127 +10,85 @@
...
-			{
-				Log.Write("server", "Can't discover UPnP-enabled device: {0}", e);
-				Game.Settings.Server.NatDeviceAvailable = false;
-				Game.Settings.Server.AllowPortForward = false;
-			}
-		}
+		public static IPAddress ExternalIP;
 
-		public static void StoppingNatDiscovery()
+		public static async Task TryNatDiscovery()
```
pchote: Likewise here.  Please move the `Game.Settings` references to the part of `Game` that calls this instead.

Mailaender:
Moved all setting lookups out of the UPnP class.

OpenRA.Game/Game.cs
```diff
@@ -250,16 +250,25 @@ internal static void Initialize(Arguments args)
 			Log.AddChannel("geoip", "geoip.log");
 			Log.AddChannel("irc", "irc.log");
 
-			if (Settings.Server.DiscoverNatDevices)
-				UPnP.TryNatDiscovery();
-			else
+			GeoIP.Initialize();
```
pchote: Please keep all this *after* the renderer initialization and first load screen paint.  Doing it here increases the delay between running the game and having the window and load screen first appear.

Mailaender: Okay. Done.

OpenRA.Game/Game.cs
```diff
@@ -279,6 +269,25 @@ internal static void Initialize(Arguments args)
...
 
+			GeoIP.Initialize();
+
+			try
+			{
+				if (Game.Settings.Server.DiscoverNatDevices == false)
+					Game.Settings.Server.AllowPortForward = false;
+				else
+				{
+					UPnP.DiscoverNatDevices(Settings.Server.NatDiscoveryTimeout).Wait();
```
pchote: Unfortunately this is still before the first load screen draw (which happens after ModData is created).
This whole block will either need to move later still, or you could keep it asynchronous, like the old code, to avoid stalling the init.

pchote: Just be be clear, this is more than a theoretical issue: `Wait()`ing on this here causes a 2-4 second pause with a black screen on my machine.

pchote:
Github marked the comment as outdated, so i'll repeat here for clarity: Calling `Wait()` forces the game to pause while the discovery runs, which takes a second or two on my machine.  The problem is that you currently do this after creating the window but before the load screen is initialized, so this causes a noticeable delay with a black screen.

The proper fix here is to start the discovery and then block on the result only at the end of the load, which is what the original implementation did.  This lets it run in parallel and hopefully complete without adding any additional delay to loading.

Mailaender:
The original implementation also used wait, but I tried to make this run truely async now. https://stackoverflow.com/questions/22864367/fire-and-forget-approach Can't really test as I don't see any noticeable delay, but I see multiple threads starting between the Nat logging events so I assume this works now.

reaperrr:
Needs rebase.

Mailaender:
Rebased.

reaperrr:
UI changes and in-game on Win7x64 seem fine, but the code is out of my league, so :+0.5:

OpenRA.Game/Game.cs
```diff
@@ -279,6 +269,25 @@ internal static void Initialize(Arguments args)
...
 
+			GeoIP.Initialize();
+
+			try
+			{
+				if (!Game.Settings.Server.DiscoverNatDevices)
+					Game.Settings.Server.AllowPortForward = false;
+				else
+				{
+					UPnP.DiscoverNatDevices(Settings.Server.NatDiscoveryTimeout).ConfigureAwait(false);
```
RoosterDragon: If we don't `await` this, then there's no guarantee this has completed! This means calls to forward ports later would fail since devices won't have been discovered yet.

You can kick the tyres on the task by doing
`var deviceDiscovery = UPnP.DiscoverNatDevices(Settings.Server.NatDiscoveryTimeout);` back up top - to get it going early. This means discovery will be running in the background whilst the rest of loading takes place.

You can then do `await deviceDiscovery;` here to ensure it has completed.

Your try-catch should be around the awaited task, as that's where any exceptions would be thrown.

OpenRA.Game/Network/UPnP.cs
```diff
@@ -10,127 +10,77 @@
...
-					Log.Write("server", "Can not forward ports via UPnP: {0}", e);
-					Game.Settings.Server.AllowPortForward = false;
+					Console.WriteLine("Getting the external IP from NAT device failed: {0}", e.Message);
+					Log.Write("nat", e.StackTrace);
 				}
 			}
 		}
 
-		public static void RemovePortforward()
+		public static async void ForwardPort(int listen, int external)
```
RoosterDragon: `async void` is an anti-pattern unless the method is an event handler. This should be `async Task`.

OpenRA.Game/Network/UPnP.cs
```diff
@@ -10,127 +10,77 @@
...
+			var token = new CancellationTokenSource(timeout);
+			natDevices = await natDiscoverer.DiscoverDevicesAsync(PortMapper.Upnp, token);
+			foreach (var natDevice in natDevices)
 			{
-				if (e.ErrorCode == 725 && lifetime != 0)
+				try
 				{
-					Log.Write("server", "NAT device answered with OnlyPermanentLeasesSupported. Retrying...");
-					ForwardPort(0);
+					ExternalIP = natDevice.GetExternalIPAsync().Result;
```
RoosterDragon: `await natDevice.GetExternalIPAsync();` would be better.

Also, should we break after we set the ExternalIP? It doesn't seem like we need to query every device if one of them responds.

OpenRA.Game/Network/UPnP.cs
```diff
@@ -10,127 +10,77 @@
...
+				catch (Exception e)
+				{
+					Console.WriteLine("Port forwarding failed: {0}", e.Message);
+					Log.Write("nat", e.StackTrace);
+				}
 			}
 		}
 
-		public static IPAddress GetExternalIP()
+		public static async void RemovePortforward()
```
RoosterDragon: `async Task`.

OpenRA.Game/Server/Server.cs
```diff
@@ -137,7 +137,7 @@ public Server(IPEndPoint endpoint, ServerSettings settings, ModData modData, boo
 			randomSeed = (int)DateTime.Now.ToBinary();
 
 			if (Settings.AllowPortForward)
-				UPnP.ForwardPort(3600);
+				UPnP.ForwardPort(Settings.ListenPort, Settings.ExternalPort);
```
RoosterDragon: You should wait on this task to complete with either `await` or `.Wait()`.

OpenRA.Game/Network/UPnP.cs
RoosterDragon: Also, you should make the caller wait on this task to complete with either `await` or `.Wait()`.

RoosterDragon: I think this may wish to be `RemovePortForward` as well, since we're here.

RoosterDragon:
Since our static fields are now likely being updated by other threads due to the tasks, I'm no longer sure they are thread-safe. The tasks will add the correct memory fences, but I'm not sure if that means the fields are then safe to be read by our main thread without some sort of synchronisation. I'm no expert in this stuff. :/

Playing it safe and throwing some locks or `volatile` on things may be better.

Mailaender:
I updated it according to the comments. Guess I can't `await` in Game.cs, because I would then have to make the whole game loop `async`. I also have no idea where to add `volatile`. I just went with the safe `Wait()` there. Hope this resolves it.

OpenRA.Game/Game.cs
```diff
@@ -279,6 +269,25 @@ internal static void Initialize(Arguments args)
...
 
+			GeoIP.Initialize();
+
+			try
+			{
+				if (!Game.Settings.Server.DiscoverNatDevices)
+					Game.Settings.Server.AllowPortForward = false;
+				else
+				{
+					UPnP.DiscoverNatDevices(Settings.Server.NatDiscoveryTimeout).Wait();
```
pchote: `DiscoverNatDevices` returns a `Task`, so why can't you assign that to a variable and then `Wait()` on it at the end of the loading?

Mailaender: Good idea. Done.

pchote:
This now crashes on launch with 
```
NAT discovery failed: Object reference not set to an instance of an object
Exception of type `System.Exception`: Tried logging to non-existent channel nat
  at OpenRA.Log.Channel (System.String channelName) [0x0002e] in /Users/paul/src/OpenRA/OpenRA.Game//Support/Log.cs:42 
  at OpenRA.Log.Write (System.String channel, System.String value) [0x00002] in /Users/paul/src/OpenRA/OpenRA.Game//Support/Log.cs:80 
  at OpenRA.Game.InitializeMod (System.String mod, OpenRA.Arguments args) [0x00391] in /Users/paul/src/OpenRA/OpenRA.Game//Game.cs:402 
  at OpenRA.Game.Initialize (OpenRA.Arguments args) [0x0023c] in /Users/paul/src/OpenRA/OpenRA.Game//Game.cs:293 
  at OpenRA.Program.Run (System.String[] args) [0x00007] in /Users/paul/src/OpenRA/OpenRA.Game//Support/Program.cs:116 
  at OpenRA.Program.Main (System.String[] args) [0x00050] in /Users/paul/src/OpenRA/OpenRA.Game//Support/Program.cs:40 
```

Mailaender:
Should be fixed now though I can't reproduce.

pchote:
It no longer crashes, but the nat log gives:
```
Open.NAT Error: 0 : System.Net.Sockets.SocketException: The requested address is not valid in this context
  at System.Net.Sockets.Socket.SendTo_nochecks (System.Byte[] buffer, Int32 offset, Int32 size, SocketFlags flags, System.Net.EndPoint remote_end) [0x0003f] in /private/tmp/source-mono-4.2.4/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.4/mcs/class/System/System.Net.Sockets/Socket.cs:2466 
  at System.Net.Sockets.Socket.SendTo (System.Byte[] buffer, Int32 offset, Int32 size, SocketFlags flags, System.Net.EndPoint remote_end) [0x00028] in /private/tmp/source-mono-4.2.4/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.4/mcs/class/System/System.Net.Sockets/Socket.cs:2454 
  at System.Net.Sockets.UdpClient.DoSend (System.Byte[] dgram, Int32 bytes, System.Net.IPEndPoint endPoint) [0x0001b] in /private/tmp/source-mono-4.2.4/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.4/mcs/class/System/System.Net.Sockets/UdpClient.cs:320 
```

Mailaender:
Sorry, that it doesn't work for you. I also didn't find anything related at https://github.com/lontivero/Open.NAT/issues

obrakmann:
just a hunch, as I haven't looked at Open.NAT at all: pchote, is your IP end point v4 or v6? How does Open.NAT work in v6 environments?

pchote:
It looks like a repeat of the same issue I brought up in #10281.

OpenRA.Game/Game.cs
```diff
@@ -392,6 +393,17 @@ public static void InitializeMod(string mod, Arguments args)
...
 			JoinLocal();
 
+			try
+			{
+				discoverNat.Wait();
+			}
+			catch (Exception e)
+			{
+				Console.WriteLine("NAT discovery failed: {0}", e.Message);
+				Log.Write("nat", e.StackTrace);
```
obrakmann: The stacktrace I got isn't particularly helpful:
```
  at OpenRA.Game.InitializeMod (System.String mod, OpenRA.Arguments args) [0x00366] in /home/oliver/devel/openra/git/OpenRA.Game/Game.cs:396 
```

Mailaender: Shall I log something else from that exception instead?

obrakmann: I guess the exception message would be better.

In my case that only contained `NAT discovery failed: Object reference not set to an instance of an object`, though.

Erm, added some more checking there. It seems `discoverNat` is null in the `try` block. No wonder it throws :p

Mailaender: Fixed the NRE for the NAT discovery disabled case and added the whole exception to the log.

OpenRA.Game/Game.cs
```diff
@@ -392,6 +393,18 @@ public static void InitializeMod(string mod, Arguments args)
 
 			JoinLocal();
 
+			try
+			{
+				if (Game.Settings.Server.DiscoverNatDevices)
```
pchote: Minor nit: it would be more correct and robust to check `if (discoverNat != null)`.

OpenRA.Game/Network/UPnP.cs
```diff
@@ -10,127 +10,76 @@
...
-				Log.Write("server", "NAT discovery started.");
-			}
-			catch (Exception e)
-			{
-				Log.Write("server", "Can't discover UPnP-enabled device: {0}", e);
-				Game.Settings.Server.NatDeviceAvailable = false;
-				Game.Settings.Server.AllowPortForward = false;
-			}
-		}
+		public static IPAddress ExternalIP;
```
pchote: Can you please turn this into a `{ get; private set; }` property?

OpenRA.Game/Network/UPnP.cs
```diff
@@ -10,127 +10,76 @@
...
-			{
-				var mapping = new Mapping(Protocol.Tcp, Game.Settings.Server.ExternalPort, Game.Settings.Server.ListenPort, lifetime);
-				natDevice.CreatePortMap(mapping);
-				Log.Write("server", "Create port mapping: protocol = {0}, public = {1}, private = {2}, lifetime = {3} s",
-					mapping.Protocol, mapping.PublicPort, mapping.PrivatePort, mapping.Lifetime);
-			}
-			catch (MappingException e)
+			var natDiscoverer = new NatDiscoverer();
+			var token = new CancellationTokenSource(timeout);
+			natDevices = await natDiscoverer.DiscoverDevicesAsync(PortMapper.Upnp, token);
```
pchote: All the upstream examples I can find use the single-device version.  Why did we go with this version and looping?

pchote:
The code looks fine as far as I can tell, except for the query about looping vs single device - trying to open/remove forwarding over multiple devices doesn't seem right to me.

I'm still suffering from https://github.com/lontivero/Open.NAT/issues/26, so there's not much more I can do here.

OpenRA.Game/Network/UPnP.cs
Mailaender: `await natDiscoverer.DiscoverDeviceAsync` causes null reference exceptions me.

Mailaender: No, wait I am stupid. Defining a local variable of course creates problems later on:

```csharp
var natDevices = await natDiscoverer.DiscoverDevicesAsync(PortMapper.Upnp, token);
```

Mailaender:
Removed support for multiple NAT devices.

obrakmann:
So, as it seems that noone wants to drop UPNP support, I guess I need to give this my :+1: 

Code looks ok as far as I can be any judge of that, and things still works on my end, even though my router doesn't support UPNP.

reaperrr:
Things still work on my end, I guess together with @pchote's code review this can be counted as 2nd :+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ff09d82e581286c147aa20f0f9b0b143ae46419a...638042778eff25fc7689f58fe2479e347ab4bf35)

PR #11290, opened 2016-05-15
Fixed a crash when loading the Dune 2000 shellmap in the map editor

Mailaender:
It will crash with `Bounds of actor concreteb Actor26 are empty.` now after I added some more useful information to the stacktraces. It happens, because concrete has `RenderSprites`, but only uses it to draw the icon. The concrete itself is a terrain modification using the `LaysTerrain` trait. So without anything that calculates the bounds of this invisible actor the editor crashes. Will probably happen with any Dune 2000 map that contains pre-placed buildings.


pchote:
The real bug here is that somebody placed the concrete actor in the map editor.  They are transient actors that should only be placed by the player and are then automatically removed from the world.

OpenRA.Mods.Common/Traits/World/EditorActorLayer.cs
```diff
@@ -106,7 +106,8 @@ EditorActorPreview Add(string id, ActorReference reference, bool initialSetup =
 
 			var preview = new EditorActorPreview(worldRenderer, id, reference, owner);
 			previews.Add(preview);
-			screenMap.Add(preview, preview.Bounds);
+			if (!preview.Bounds.IsEmpty)
```
pchote: Please throw a sensible exception here instead of silently ignoring it.

Mailaender: The exception thrown is already sensible thanks to the first 2 commits. It just shouldn't throw here. We already silently skip the actors from the editor actor selector.

pchote: We silently skip in the actor selector because we don't have a declarative way to say which actors should be there and which shouldn't.  We do have a declarative way of saying which actors should be in the map, and shouldn't tolerate cases where invalid actors are defined there.

Mailaender:
Fixed up the map instead and kept the potential crash.

obrakmann:
:+1:

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/894d740667ac6aa7594d4cf23e1d0a73a05e1dbc...9f51f3d0e8cec6b8a2aa4cd0a62d2315adadca82)

PR #11294, opened 2016-05-15
Renamed C4Demolition to Demolition

Mailaender:
A rather boring tidy-up refactoring submission.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -777,6 +777,17 @@ namespace OpenRA.Mods.Common.UtilityCommands
 						node.Value.Nodes.RemoveAll(x => x.Key == "StartsRevealed");
 				}
 
+				if (engineVersion < 20160515)
+				{
+					// Use generic naming for building demolition using explosives.
+					if (node.Key == "C4Demolition")
+						node.Key = "Demolition";
+
+					foreach (var n in node.Value.Nodes)
+						if (n.Key == "C4Delay")
+							n.Key = "DetonationDelay";
+				}
+
 				UpgradeActorRules(engineVersion, ref node.Value.Nodes, node, depth + 1);
 			}
 		}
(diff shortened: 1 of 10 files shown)


OpenRA.Mods.Common/Orders/DemolitionOrderTargeter.cs
```diff
@@ -0,0 +1,36 @@
...
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Orders
+{
+	class DemolitionOrderTargeter : UnitOrderTargeter
```
pchote: Why do you keep on trying to break these targeters out into the main namespace (this is at least the fifth or sixth case)?  They are encapsulated in the parent class because they have no use to outside code.

Mailaender: https://github.com/OpenRA/OpenRA/tree/bleed/OpenRA.Mods.Common/Orders It is not my idea, but how the code is already structured. This adds consistency.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -777,6 +777,17 @@ internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode>
 						node.Value.Nodes.RemoveAll(x => x.Key == "StartsRevealed");
 				}
 
+				if (engineVersion < 20160515)
+				{
+					// Removed FrozenUnderFog.StartsRevealed
```
GraionDilach: Needs a fixup.

Mailaender: Fixed.

abcdefg30:

GraionDilach:
Looks good to me as well, no issues, albeit I'm unsure if you two dealt on the ordertargeter shifts. :+0.95: uintil then.

OpenRA.Mods.Common/Orders/DemolitionOrderTargeter.cs
chrisforbes: If you end up being able to actually unify some more of the weird OTs, I could get behind this. If it's just taking an implementation detail of a trait and throwing it somewhere else, I don't think it's a good plan.

Mailaender:
Reverted the order targeter namespace shift.

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d47a15806453d452f08a7031d0e6f680c23a5265...4ae5fb89ef418cffa928969285f168674439dc01)

PR #11295, opened 2016-05-15
Added the sniper back into multiplayer for France

Mailaender:
Fits to the deception theme. Achieved broad consensus in the balancing discussion [forums](http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&p=292926#292926).

### DIFF
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -42,6 +42,7 @@ DOG:
 	IgnoresDisguise:
 	DetectCloaked:
 		CloakTypes: Cloak, Hijacker
+	RenderDetectionCircle:
 	Voiced:
 		VoiceSet: DogVoice
 	-TakeCover:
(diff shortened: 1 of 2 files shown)


HenrytheSlav:
I feel rather strongly opposed to this, as seen on irc, because

- allies are way stronger against infantry already. 
- this is not original unit, thus there should be no NEED to implement it again globally

Compare effectiveness against infantry, for example:

- Hind vs. yak, if we skip intentional crashing, hinds are far easier controllable, and can be used suicidally too in the end.
- Pillboxes/Flametower, that is nobrainer.
- Allies have Tanya, who is meatgrinder. Soviets have teslas, which are good, but latetech, Russia only and cost-inefficient.
- Ranger/ARTY vs. Flak Truck, APC, V2. APC is good at antiinf, flaktruck is goodish, while arty is meatgrinder again. V2 is comparable but has reloading.

I feel this is unneeded adition.

MustaphaTR:
Agrreed with @HenrytheSlav . :-1:

Mailaender:
I don't think the sniper excels at mass killing infantry, but as a stealth-ed scout.

Mailaender:
Guess I am the only one who liked the sniper unit around here.

PR #11296, opened 2016-05-15
Removed dead AI code

reaperrr:
Edit: Changed PR to remove unused code instead.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -192,8 +192,6 @@ namespace OpenRA.Mods.Common.AI
 		public object Create(ActorInitializer init) { return new HackyAI(this, init); }
 	}
 
-	public class Enemy { public int Aggro; }
-
 	public enum BuildingType { Building, Defense, Refinery }
 
 	public sealed class HackyAI : ITick, IBot, INotifyDamage
@@ -232,7 +230,6 @@ namespace OpenRA.Mods.Common.AI
 
 		BitArray resourceTypeIndices;
 
-		Cache<Player, Enemy> aggro = new Cache<Player, Enemy>(_ => new Enemy());
 		List<BaseBuilder> builders = new List<BaseBuilder>();
 
 		List<Actor> unitsHangingAroundTheBase = new List<Actor>();
@@ -1051,9 +1048,6 @@ namespace OpenRA.Mods.Common.AI
 
 			if (!e.Attacker.Info.HasTraitInfo<ITargetableInfo>())
 				return;
-
-			if (e.Damage > 0)
-				aggro[e.Attacker.Owner].Aggro += e.Damage;
 
 			// Protected harvesters or building
 			if ((self.Info.HasTraitInfo<HarvesterInfo>() || self.Info.HasTraitInfo<BuildingInfo>()) &&
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -559,45 +559,6 @@ namespace OpenRA.Mods.Common.AI
 				World.IssueOrder(orders.Dequeue());
 		}
 
-		internal Actor ChooseEnemyTarget()
-		{
-			if (Player.WinState != WinState.Undefined)
-				return null;
-
-			var liveEnemies = World.Players
-				.Where(p => Player != p && Player.Stances[p] == Stance.Enemy && p.WinState == WinState.Undefined);
-
-			if (!liveEnemies.Any())
-				return null;
-
-			var leastLikedEnemies = liveEnemies
-				.GroupBy(e => aggro[e].Aggro)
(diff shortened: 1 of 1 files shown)


reaperrr:
Note: It is hard to tell how much this effectively helps compared to bleed, so this will require some more extensive testing.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -98,6 +98,12 @@ public class BuildingCategories
 		[Desc("Delay (in ticks) until rechecking for new BaseProviders.")]
 		public readonly int CheckForNewBasesDelay = 1500;
 
+		[Desc("Delay (in ticks) before deciding which enemy to attack first.")]
+		public readonly int ChooseInitialEnemyDelay = 750;
```
abcdefg30: Can you elaborate a bit why you added this?
I mean, couldn't we just use `UpdateAggroInterva` only instead?

reaperrr: I just implemented the intitial delay first and added the interval later. You're right, it's better to keep just the interval.

reaperrr:
Updated.

reaperrr:
Updated, changed `*Interval` to `*Delay` to avoid an upgrade rule later. As @GraionDilach mentioned in #9810, we need to rename Interval to Delay for full consistency.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -559,6 +565,36 @@ public void Tick(Actor self)
...
+
+			var leastLikedEnemies = liveEnemies
+				.GroupBy(e => aggro[e].Aggro)
+				.MaxByOrDefault(g => g.Key);
+
+			var enemy = (leastLikedEnemies != null) ?
+				leastLikedEnemies.Random(Random) : liveEnemies.FirstOrDefault();
+
+			// Bump the aggro slightly to avoid changing our mind
+			if (leastLikedEnemies.Count() > 1)
```
abcdefg30: What happens when `leastLikedEnemies` is `null`?

reaperrr: I'm not sure whether there are any circumstances under which `leastLikedEnemies` can be `null` as long as there are still `liveEnemies`. I assumed there aren't any, but I'm not 100% sure how `MaxByOrDefault` works.

But yes, if it's possible for `leastLikedEnemies` to be null, that case needs to be covered, and if not, the `var enemy` can be simplified.

reaperrr:
Updated:
- now covers the case when `leastLikedEnemies` is `null` but `enemy` is not
- moved `var deadEnemies` behind the `liveEnemies.Any()` check so it's only performed if there are any live enemies remaining

RoosterDragon:
So a quick breeze over HackyAI shows that `Aggro` is only used within `ChooseEnemyTarget` and `ChooseEnemyTarget` is unused!

Everything else seems to find `FindClosestEnemy` to decide what to attack, which doesn't care about aggro.

Looks like some refactoring may be in order!

pchote:
Well that would explain why the AI suddenly turned dumb a couple of releases ago.

reaperrr:
So the AI sittling idle sometimes is due to the closest enemy being too far away...

RoosterDragon:
There is no upper limit to the search radius for finding an enemy most of the time, so at a glance I would expect the AI to attack things across the map if it had to.

This will probably need some deeper investigation to determine why the AI idles.

GraionDilach:
I've seen the AI going against worms in D2k immediately instead of attacking each other - during reproducing that activity-stack-overflow report - so I'd guess the AI ends up aggroing the wrong player. Couldn't check it yet how this PR helps on that though.

RoosterDragon:
First bad commit is e0b7242f1b0dac2c282aadbb2ecbca5fb6f3d1e1 (#2736, #2600), which would've affected release-20130514.

Given we've been without aggro for so long, it may be more prudent to remove the dead code and fix the existing code as needed.

reaperrr:
> So a quick breeze over HackyAI shows that Aggro is only used within ChooseEnemyTarget and ChooseEnemyTarget is unused!

It's used when an actor is damaged, Aggro against the attack's owner is buffed by the damage in that case (unless that's somehow not working either, but my impression is that it works, the AI usually responds to being attacked).

Anyway, AI is not exactly my strong point so I changed this PR to just remove the dead `ChooseEnemyTarget()` code for now.

RoosterDragon:
Sure, being attacked updates the aggro value. But nothing ever reads that value and uses it! So it has no effect in the end. You may as well get rid of it entirely.

(Also, you should remove the "fixes" in the OP lest merging this closes those now)

obrakmann:
:+1:

reaperrr:
Updated, removed the remaining unused aggro left-overs.

pchote:
It seems a shame to remove this instead of fixing it, but removing is better than leaving it broken so :+1:

pchote:
(no changelog because this has no player- or modder-visible changes)

PR #11297, opened 2016-05-15
Added WithResourceSeedingAnimation and WithResourceSeedingOverlay

Mailaender:
A more true to the original approach. From the discussion started at https://github.com/OpenRA/OpenRA/pull/9205#issuecomment-136151851.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithResourceSeedingAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithResourceSeedingAnimation.cs
@@ -1,0 +1,56 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits.Render
+{
+	[Desc("Plays an animation replacing the default body animation when resources are spawned.")]
+	public class WithResourceSeedingAnimationInfo : UpgradableTraitInfo, Requires<WithSpriteBodyInfo>
+	{
+		[SequenceReference, Desc("Sequence names to use.")]
+		public readonly string[] Sequences = { "active" };
+
+		[Desc("Time (in ms) to wait before playing.")]
+		public readonly int Delay = 0;
+
+		public override object Create(ActorInitializer init) { return new WithResourceSeedingAnimation(init.Self, this); }
+	}
+
+	public class WithResourceSeedingAnimation : UpgradableTrait<WithResourceSeedingAnimationInfo>, INotifyResourceSeeded, INotifyBuildComplete
+	{
+		readonly WithSpriteBody wsb;
+		bool buildComplete;
+
+		public WithResourceSeedingAnimation(Actor self, WithResourceSeedingAnimationInfo info)
+			: base(info)
+		{
+			wsb = self.Trait<WithSpriteBody>();
+			buildComplete = !self.Info.HasTraitInfo<BuildingInfo>(); // always render instantly for units
+		}
+
+		void INotifyResourceSeeded.OnResourceSeeded(Actor self)
+		{
+			if (!buildComplete || IsTraitDisabled)
+				return;
+
+			Game.RunAfterDelay(Info.Delay, () => {
+				wsb.PlayCustomAnimation(self, Info.Sequences.Random(Game.CosmeticRandom),
+					() => wsb.CancelCustomAnimation(self));
+			});
+		}
+
+		public void BuildingComplete(Actor self)
+		{
+			buildComplete = true;
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Traits/Render/WithResourceSeedingOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithResourceSeedingOverlay.cs
@@ -1,0 +1,73 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits.Render
+{
+	[Desc("Renders a decorative overlay animation when resources are spawned.")]
+	public class WithResourceSeedingOverlayInfo : UpgradableTraitInfo, Requires<RenderSpritesInfo>, Requires<BodyOrientationInfo>
+	{
+		[SequenceReference, Desc("Sequence names to use.")]
+		public readonly string[] Sequences = { "active-overlay" };
+
+		[Desc("Custom palette name")]
+		[PaletteReference("IsPlayerPalette")] public readonly string Palette = null;
+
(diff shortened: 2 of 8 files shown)


OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -38,6 +38,7 @@ public interface INotifyAttack { void Attacking(Actor self, Target target, Armam
 	public interface INotifyCharging { void Charging(Actor self, Target target); }
 	public interface INotifyChat { bool OnChat(string from, string message); }
 	public interface INotifyParachuteLanded { void OnLanded(Actor ignore); }
+	public interface INotifyResourceSeed { void OnSeed(Actor self); }
```
pchote: Minor nit: Can we call these `INotifyResourceSeeded` and `OnResourceSeeded`?

OpenRA.Mods.Common/Traits/Render/WithResourceSeedingOverlay.cs
```diff
@@ -0,0 +1,73 @@
...
+
+			var anim = new AnimationWithOffset(overlay,
+				() => body.LocalToWorld(info.Offset.Rotate(body.QuantizeOrientation(self, self.Orientation))),
+				() => IsTraitDisabled || played,
+				p => RenderUtils.ZOffsetFromCenter(self, p, 1));
+
+			rs.Add(anim, info.Palette, info.IsPlayerPalette);
+		}
+
+		public void OnSeed(Actor self)
```
pchote: Please use the explicit interface syntax: `void INotifyResourceSeed.OnSeed(Actor self)`.

reaperrr:
I think the delay functionality should be added to `WithResourceSeedingAnimation` trait as well.

Mailaender:
Updated.

OpenRA.Mods.Common/Traits/Render/WithResourceSeedingOverlay.cs
```diff
@@ -0,0 +1,73 @@
...
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits.Render
+{
+	[Desc("Renders a decorative overlay animation when resources are spawned.")]
+	public class WithResourceSeedingOverlayInfo : UpgradableTraitInfo, Requires<RenderSpritesInfo>, Requires<BodyOrientationInfo>
+	{
+		[Desc("Sequence name to use")]
+		[SequenceReference] public readonly string Sequence = "active-overlay";
```
reaperrr: It would be nice to have support for random sequences here as well.

OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -38,6 +38,7 @@ public interface INotifyAttack { void Attacking(Actor self, Target target, Armam
 	public interface INotifyCharging { void Charging(Actor self, Target target); }
 	public interface INotifyChat { bool OnChat(string from, string message); }
 	public interface INotifyParachuteLanded { void OnLanded(Actor ignore); }
+	public interface INotifyResourceSeeded { void OnResourceSeeded(Actor self); }
```
pchote: Please add the explicit interface attribute here too, so this can be linted.

OpenRA.Mods.Common/Traits/Render/WithResourceSeedingOverlay.cs
```diff
@@ -0,0 +1,73 @@
...
+		[PaletteReference("IsPlayerPalette")] public readonly string Palette = null;
+
+		[Desc("Custom palette is a player palette BaseName")]
+		public readonly bool IsPlayerPalette = false;
+
+		[Desc("Position relative to body")]
+		public readonly WVec Offset = WVec.Zero;
+
+		[Desc("Time (in ms) to wait before playing.")]
+		public readonly int Delay = 640;
```
pchote: @reaperrr clarified in IRC that he meant a delay the other way: to delay the seeding relative to the animation.  Delaying the animation relative to the seeding doesn't have a use case (that I can think of, at least).

reaperrr: What I had in mind was to play the animation/overlay only every 5th seed or so, but as @pchote pointed out on IRC that can be implemented with random sequences (for example: `Sequences: active, idle, idle, idle, idle`)

Mailaender:
Updated again.

OpenRA.Mods.Common/Traits/Render/WithResourceSeedingAnimation.cs
```diff
@@ -0,0 +1,56 @@
...
+			wsb = self.Trait<WithSpriteBody>();
+			buildComplete = !self.Info.HasTraitInfo<BuildingInfo>(); // always render instantly for units
+		}
+
+		void INotifyResourceSeeded.OnResourceSeeded(Actor self)
+		{
+			if (!buildComplete || IsTraitDisabled)
+				return;
+
+			Game.RunAfterDelay(Info.Delay, () => {
```
GraionDilach: RunAfterTick would be better IMO, RunAfterDelay is only used in SwallowActor within game logic.

Mailaender: `RunAfterTick` does not allow to specify the time and is only used for UI operations.

pchote: Neither `Game.RunAfterTick` nor `Game.RunAfterDelay` exist at a lower level / outside the world, and should not be used from gameplay code!

Like mentioned earlier, i'm not aware of a use case for delaying the animation further after the seeding, so this should just be removed.

Mailaender: Look at the TS example before making hastily claims. This is required for the spread overlay, which needs to spray after the tree has made his blow animation. Having both at once will look broken.

pchote: I could say the same thing!  Like we talked about in IRC yesterday this code is the exact opposite of what the TS overlay needs: adding a delay here makes the desync between the animation and spreading *worse*, not better.

I thought you were purposely ignoring that discussion, and didn't want this to turn into another PR where I need to continuously ride your back about getting changes done... so was going to wait for this to be merged without the delay and then fix that properly myself.

Mailaender: I wasn't aware of that discussion and thought your minor "please rename stuff" nit picks were your sign of approval for this approach. Looking at the logs you didn't even address me and I wasn't even in the room when you two chattered...

pchote: Or if you want to fix it now: Introduce an `OnWillSeedResource` method to the interface and change this to use that.  In the resource seeding trait implement a delayed action queue copying `Armament`. Trigger `OnWillSeedResource` when queing the delayed seeding action (delay length configured on the seeding trait) to start the animation here, and then trigger `OnResourceSeeded` when the seeding actually occurs N ticks later.

pchote: My minor "please rename stuff" *is* a sign of my approval for this approach.  The fix I outlined in the comment above is a natural extension to what you've done here (aside from the Delay stuff) that I would have been happy to take in a followup pr.

GraionDilach: TBH, I mixed `Armament`'s `ScheduledDelayAction` with `RunAfterTick` when I started this conversation.

pchote:
Adding the changes requested tag for removing the Delay or implementing it properly on the resource spawning side (your choice).

Mailaender:
The whole idea from https://github.com/OpenRA/OpenRA/pull/9205#issuecomment-136151851 turned into a bad result. Because we spawn resources way faster than the original the Tiberium trees look like factory chimneys constantly puffing spores out.

Creating dependencies between the resource seeding and rendering traits also sounds like a horrible plan.

pchote:
> Creating dependencies between the resource seeding and rendering traits also sounds like a horrible plan.

The idea of implementing it as an interface was to avoid that.

Issue #8993, opened 2015-08-08
Improve support for external mods


pchote:
- [x] Add engine and required-mod version checks. [#8991]
- [x] Support mods loaded from the support directory. [#8992]
- [ ] Support oramod version checks and upgrades from the mod chooser. [#9028]
- [ ] Support oramod installation via an `openra-mod://` URI. [#9029]
##

Phrohdoh:
Will we update via a direct link to a `.oramod` that is saved and unpacked into the support directory (or possibly kept packed and loaded from zip)?

As an extension idea related to bullet point 3 I'd like to see a place in the UI and in the mod directory for a changelog file that mod authors can include (not necessary!) that players can read and make the decision to upgrade or not based on the information provided.

Mailaender:
The current best practice for vanilla Tiberian Sun mods is to have a spawner which checks for updates and has a button to automatically download and install those. We were asked if such a content delivery feature (like we do for mods) is planned for OpenRA as well. http://www.ppmforums.com/viewtopic.php?p=533798#533798 You might want to take that into consideration.

reaperrr:
#11326 would also improve support for external mods a bit, in my opinion.

pchote:
The remaining issues have been closed, so this is now complete.

Issue #8850, opened 2015-07-31
Unite mod DLLs?


penev92:
This was suggested by @chrisforbes and is something that has crossed my mind a few times before.

Currently there is little code in each of the mod DLLs, and a ton of code in `Common`. The difference for `Common` would be negligible, but this would make the "Can this mod/map use trait X that is from mod Y?" issues a lot better.

> [23:23:58]< chrisf> things that turn out to be useful will move into common code and slowly get more general
> [23:24:17]  < Phrohdoh> certainly, but where do we draw the line?
> [23:24:40]  < chrisf> im not sold on there _being_ a line for things like traits.
> [23:24:50]  < Phrohdoh> Then we should just have a mods.dll
> [23:24:52]  < chrisf> if you have some weird ui junk, sure, keep that in your own dll 

I know we've talked about splitting `Common` to "graphics stuff" vs "logic", or "sprite stuff" vs "voxel stuff" or whatever. Maybe we can dump all the stuff in `Common` first. Then have `Activities.dll`, `Traits.dll`, `Traits.Render.dll`, etc. or whatever we name them.

At this point I am not sure how much point there is to have per-mod DLLs (custom mods of course will add their own logic in their own DLLs) for the official mods, considering how little code they actually contain.

Opening this for discussion. I particularly want to hear what @reaperrr and @GraionDilach think about this, as people with their own mods, because I think this will affect them. Of course we need opinions from the other people interested/involved/capable on commenting on the project's architecture.

(I'm not going to label this issue for now because it's a `Question` that I hope will quickly move to an `Idea` about a `Refactor`.)
##

Phrohdoh:
I've suggested this quite a few times but iirc @pchote has his reasons for us not doing this, so if someone digs those up from the logs they'd be a good addition to this ticket.

GraionDilach:
I am a scavenger and adopter by nature. I personally see no difference in either case.

The sole thing I can think of that if all traits be moved to Common, rejecting anyone's own traits even though out of scope - like RA2 traits or written-for-mod-only stuff - is less justified than the current "shared traits in Common, mod-specific logic for the mod" direction which at least imply that traits without a usecase in the shipped mods are less welcome.

Mailaender:
[...]

PR #11301, opened 2016-05-16
Added OpenRA.Mods.RA2.dll

Mailaender:
This just moves the existing source code files to ease repository setup and reuse. The mod rules and configuration stays at https://github.com/OpenRA/ra2/


obrakmann:
No. :-1: 

For two reasons:

1. I said from the start that I was only ok with the existence of https://github.com/OpenRA/ra2  as long as it doesn't cause work for us (which you very well know, so I really find this PR offensive). This change means that pull requests for ra2 will end up on our main repo, which will definitely cause work and even worse, distractions from our more immediate goals. We already have a hard enough time reviewing all PRs on this repo, this will just make it worse.
2. I don't see why ra2 has to be the special snow flake that needs to live upstream while a number of other third-party mods do just fine without.

It is now clear to me that allowing https://github.com/OpenRA/ra2 to be created was completely wrong, sent the wrong signal and should have been prevented. The same is probably true for https://github.com/OpenRA/d2. I'm sorry now for taking such a laissez-faire approach and not a more active stance against it when the issue came up half a year ago.

reaperrr:
I'm neutral on 2), but I agree on 1).

reaperrr:
I'm closing this, I don't think we should waste time on repeating discussions that will have the same outcome as before anyway.

Mailaender:
I did this out of necessity not to put burden onto anyone. The layout in which OpenRA is organized doesn't allow a real modular approach. This gives two options
* a) fork OpenRA as OpenRA2 to focus on RA2 and TS in a combined repository
* b) step down from working on RA2 as it is impossible with the current restrains

GraionDilach:
@obrakmann: I wouldn't go that harsh with words over there, calling a probably good step in the right direction a bad one. https://github.com/OpenRA/ra2/issues/208 explains the reasons behind this PR (albeit I acknowledge and agree with declining it). The sole reason why thirdparty mods atm work is because all thirdparty modders deal with the release issues on their own and probably not try to automate a release (best example: @reaperrr compiled/uploaded a TDX release manually ffrom what I saw to GitHub).

The true solution to this problem would be (notsure how feasible this is, because I lack the required knowledge of .gitignore and makefiles) something like to make gitignoring a lot more explicit (ignore any mod folders which are not OpenRA-supported both in C# - since thirdparty libraries already follw the OpenRA naming convention with OpenRA.Mods.RA2/D2/AS/TDX - and within the mods, with only tracking cnc, d2k, modchooser, ra, and ts folder there) and somehow rewriting the makefile to be scripted to notice all additional all mod code projects without the c'n'p tricks it has to rely on atm.

If commits like https://github.com/GraionDilach/OpenRA/commit/97b0f450d4c82ddf0e933ce77c5e5139f2502af5 wouldn't be required to ensure automated compilation (NB: I'm not asking for the solution expansion nor the code check aspect) then the issues this PR meant to solve can be solved without upstream interference.

pchote:
My 2 cents, aimed mostly in response to @GraionDilach's comment:

This is a problem that we need to solve for all modders, and myself and @penev92 have done our best towards it (adding support for mod dependencies, oramods, my rejected patch to allow arbitrary support dirs via the commandline), but have received very little engagement from the modder side in return.  The RA2 mod has adopted an "upstream or bust" philosophy, and other mods (like AS, CD, SP) try to work within the current limitations without giving feedback on the issues or how to improve things.

The IRC logs have multiple discussion fragments where I have tried to kickstart automatic mod compilation (using the engine as a dependency), but nobody (okay... really just the RA2 mod) is interested.

Mailaender:
I am going to try something in https://github.com/OpenRA2 which might either result in a fork where TS and RA2 are the focus and RA/TD/D2k are ignored/removed or everything (mods, engine, tools maybe even dependencies) is stored in it's own repository glued together with git sub-modules.

pchote:
You can do whatever you want, of course, but if you go ahead with this and keep the "OpenRA 2" name without consensus agreement then you risk generating a huge amount of bad will.

mods/ra/rules/defaults.yaml
```diff
@@ -243,8 +243,6 @@
 		GroundImpactSound: squishy2.aud
 		WaterImpactSound: splash9.aud
 		WaterCorpseSequence: small_splash
-	Cloneable:
-		Types: Infantry
```
Mailaender: Do you really want to keep the unused trait?

Mailaender:
I am always open to suggestions. The blue icon might make it easier to distinguish. Hopefully this won't be permanent.

I came up with the best of both worlds. A minimally invasive https://github.com/OpenRA2/OpenRA fork that just adds some project file adjustments and git submodules.

mods/ra/rules/defaults.yaml
Mailaender: I wonder if this really wants to stay as it seems to be an unused trait.

pchote:
This is a good example of https://en.wikipedia.org/wiki/Cart_before_the_horse.  It is not feasible or sensible to make the core engine project depend on everybody's third party mods.

Mailaender:
What do you suggest instead?

pchote:
The same thing i've been suggesting for months.  Here's two irc excerpts, but there are at least 3 or 4 more.

```
[15:08:08]	<pchote> Mailaender: you'd make life much easier for yourselves if you used the version dependency stuff
[15:08:18]	<pchote> and then had weekly (or whenever you feel like it) update prs
[15:08:40]	<pchote> none of this random breakage and high priority prs to fix it
[15:08:44]	<Mailaender> to work on OpenRA I would then probably need to have 2 repositories
[15:09:38]	<pchote> when working on engine features for ra2 you can just change your ra2-local-engine-fetch script to point to your own repo branch
[15:10:32]	<pchote> in other words: yes, you would, but why is that a problem?
[15:11:43]	<Mailaender> not a problem I assume just a lot of working setting that up again so I probably won't do it
---
[15:12:54]	<Mailaender> feel free to send a pull request 
```

```
[14:35:44]	<pchote> also, theres an obvious technical solution to this
[14:36:02]	<Phrohdoh> Mailaender: You can change the version manually in the mod.yaml, not everything needs to be done by a script.
---
[14:36:36]	<pchote> install the ra2 mod into the support dir, and then have makefile commands to clone and update a local engine version *inside* the ra2 mod dir
---
[14:37:09]	<pchote> if someone ever fixed the ability to specify the support dir on the commandline then it could go anywhere
[14:37:13]	<abcdefg30> submodule's again
[14:37:23]	<pchote> abcdefg30: no, not submodules
[14:37:28]	<pchote> although you could do that too I guess
```

Mailaender:
It turned out the problem you were trying to solve with keeping dependencies in mod.yaml isn't such a big issue. The main problem was building the OpenRA.Mod.RA2.dll without OpenRA. The copied or linked dependencies worked, but aren't redistributable. Linking static DLLs causes `TypeLoadExceptions` in the OpenRA engine, see https://github.com/OpenRA/ra2/issues/208 and many people have problems setting up their development environment so git submodules seems to be the easiest way to automate that. I agree that the monolithic structure of the OpenRA project is far from optimal. It has to change at some point.

pchote:
I reference that issue in both of the above quotes.  The engine code is cloned to a directory or submodule inside your project dir, and your project file references that.

Mailaender:
Something like `$HOME/.openra/mods/ra2/OpenRA/OpenRA.Game.exe` would pretty much violate all sane standards as it puts a duplicate binary inside a user data folder as well as the placement structures this project adheres to. https://github.com/Phrohdoh/oramod-ra2 went that route. It was abandoned in 2015. Not sure if that simplifies repository setup and is at the same time easy to understand for more and more people who want to try out the development version of the RA2 mod.

pchote:
If you're developing a mod against an unstable engine then you need to keep a copy of the specific engine version that you need *somewhere*.  Requiring people to manually manage that engine version is what has lead to your current clusterfuck.

You have also missed the part of the quote above where I suggest passing the support / mod dir directly to the engine so that it doesn't need to live in `$HOME/.openra/mods/`.

Mailaender:
The idea is to develop against OpenRA bleed so you have just one repository to work with. A throwaway OpenRA version or a second one would limit the possibilities or take extra maintenance work. To solve the problem of incompatible version, we need stable releases of the OpenRA mod that work with the current or upcoming releases, which also needs this change to produce a OpenRA.Mods.RA2.dll that actually works outside of a development environment.

How is that supposed to work with the portable support directory? With `ra2/OpenRA/Support/mods/ra2` that would require a recursive backlink to itself. Also have fun writing a tutorial for that setup.

GraionDilach:
Since I am partially responsible in this PR and it's aftermath:

 - I am not supporting this engine fork (and sorry, @Mailaender, I am not accepting that invitation either). Complete forks are IMO the worst overhead possible and the RA2 mod already had that path as well in Phrohdoh/OpenRA.
 - I amn't certain what would be the ideal solution nor I am the best modder to provide feedback in terms of mod support, since I am too much of an adjuster by nature. I just know that right now my environment setup (which I converted SP as well weeks ago) works well for me, and only has the overhead of running separate repositories in the same folder, with only rebasing a single commit all the time against upstream - something I am completely fine myself. What I can tell at this point that so far to me, the new mod supporting ideas - like .oramod - are helpful when releasing a mod, but doesn't seem much useful when developing one.
 - In any cases, I am planning to return to IRC to somehow give modding support feedback, however lately, I've grinded my PC to a halt (due to some RL reasons) and while I already ordered a new PC (mostly because I had fed up with my GPU performance and the lag resulted from it, which was the main reason why the IRC feedback considered that one AS stream catastrophic back then), it's already two weeks late in deliver and even today I have no idea when I actually will get my hands on it and I don't intend to return to active OpenRA discussions beyond PR reviewing until that computer arrives.

pchote:
Please try to read and most importantly understand the things I have written above, because i'm getting a bit tired of repeating myself verbatim.

Developing a mod against a moving target is always going to be difficult, you can't avoid that, and the idea of moving this difficulty upstream by making the core engine project reference downstream mod repositories is completely unacceptable.

The proposal I have suggested covers your moving-target use case.  For releasing mods you would obviously not use this, and instead you should be packaging an oramod that is built against a release-tagged engine version.

GraionDilach:
Okay, I should've looked at the fork. The name must indeed change, but otherwise it follows my setup from what I see and it's not following Phrohdoh/OpenRA. Should have checked it out beforehand I go against it. :S 

@pchote: Up til today, the RA2 mod actually followed the system your second IRC snippet intended albeit crudely (without automating the things). I have a rough guess how to interpret your project setup:

```
/RA2/ (this would be the "support" dir, which has it's own solution, makefile and whatnot, to compile everything in it's /OpenRA/ subfolder, also supplying bats/whatnot to launch /OpenRA.Game.exe with the support dir override)
/RA2/OpenRA (this would be the OpenRA dir which would be forced to take .. as support)
```

Is this really how would it look like though? By shifting all edited coming-from-upstream files a folder above with the structure and then taking them from a relative path during compiling?

pchote:
The location doesn't really matter, it could be `/RA2/OpenRA`, `/RA2/dependencies/engine`, `/OpenRA-RA2`, `/var/temp/<random string>`, or anything else you could dream of.  The key points are that (1) it is a directory that is understood to be owned and managed by the mod's makefile and treated as a thirdparty dependency and (2) that it is a known path so that you can reference the binaries from your ra2 project files.

You then build your makefile commands to automatically manage that self-contained engine version based on metadata that you define for the RA2 mod.  I've been suggesting that you use the mod.yaml dependency line so that you only need to set the upstream version in one place, but that appears to have caused at least some confusion.

You then modify the upstream engine code to define some way to run your self-contained engine copy with an arbitrary mod path, either by setting the support directory, overloading the Game.Mod parsing to accept a path as well as an ID, or any other solution that works for you.

GraionDilach:
Okay, I think I got it. This means that the mod.yaml dependency line would refer to an exact commit, which would be checked out by the mod's make script, which can just simply compile the upstream dependencies, engine, and version them thenafter.

I have to acknowledge that such a setup would actually make the OpenRA/RA2 repository the most efficient one.

pchote:
Exactly.  You specify the `git-<shorthash>` version that is automatically generated by `make version`, and then trim the leading `git-` to get the commit hash to check out.

GraionDilach:
Okay, if nothing else comes in, I will try to get this setup implemented in an OpenRA/RA2 branch tomorrow (gotta abuse the push rights if I have them :P). However to make this system working seamlessly, #10981 needs to be merged beforehand.

Mailaender:
I would rather prefer something like https://git.eclipse.org/c/platform/eclipse.platform.releng.aggregator.git/tree/, but git doesn't support nested sub-modules which we would need for a twisted OpenRA directory layout.

pchote:
As above, solutions that rely on an upstream repository referencing downstream mods is not viable.  It is a selfish solution that effectively says that we only support specially nominated mods and that we don't care about anyone else.

GraionDilach:
Not just that, this current proposal has the potential to end the - now habitual - "revamp the RA2 mod setup every fourth month" with bringing in an automated and as-user-friendly-as-upstream setup introductions with an acceptable overhead (have a second OpenRA upstream downloaded) a lot of mod maintainers already have, without going overly complicated.

I guess I'm sold on it now. I do have to demand for #10981 to get it done, though.

Iamgoofball:
lets be honest here

the most played and well known WW game is probably RA2

the majority of people will probably want to play RA2

if you launched RA2 as an official part of OpenRA, people would be all over that shit like a group of children in a candy store

the reality is that if RA2 gets released, it will probably be the most played mod on OpenRA

so why not capitalize on that and just ship OpenRA with RA2 by default like tibsun is being shipped

pchote:
Tibsun isn't being shipped.

Iamgoofball:
why not? is development on tibsun also going to be pushed to another repo?

pchote:
Neither tibsun or ra2 are anywhere near done enough to ship to our tens of thousands of players.

Iamgoofball:
did you really think I was talking about shipping it now immediately 100%

GraionDilach:
@Iamgoofball Please do not meddle into discussions you don't understand. Especially with such a tone. There aren't enough manpower to fully finish TS and D2k at the moment either, broadening the main project at this point is beyond the capacity. The RA2 project should stay third-party to decrease the wandering and let upstream concentrate to the upstream aims.

Besides that, RA2 is pledged with enough legal issues as-is already.

TLDR: You have no idea what you're talking about so it's better if you'd just drop the topic, thank you.

Iamgoofball:
okay, that makes sense

you should look into a locked down irc channel/contributor discussion locked github issue to continue discussion so this PR isn't getting flooded with people like me who are confused

Mailaender:
The legal issues of RA2 are nearly identical to all other C&C derived based if you put down the fanboy freeware release myth, which didn't solve the IP nor the redistribution problems. The idea of keeping the most anticipated mod out of this repository so maintenance of the other ones is less work, is selfish, short-sighted and inefficient if I look back at all the past troubles, wasted time to maintain it outside of this repositories, but it is a majority decision that gets repeated religiously.

Everybody just working on their pet projects without getting disturbed would be no problem if OpenRA was a modular project. Sadly the way OpenRA is structured with a huge shared and fluctuating common traits library #8850 and hard-coded locations to search for mods, libraries and dependencies is a big problem. However this does not only apply to the RA2 mod, but to every thirdparty one, so improving thirdparty mod support #8993 is now on the agenda.

penev92:
Just throwing it out there: it has been on the agenda for more than half a year now.

pchote:
> However this does not only apply to the RA2 mod, but to every thirdparty one, so improving thirdparty mod support #8993 is now on the agenda.

This is what I have been saying all along, and is why I have been insisting that we *fix these issues* instead of merging the RA2 mod upstream.  I am glad that you have finally seen the light on this point, maybe now we can get all these arguments behind us.

Mailaender:
I still see all these arguments used as an excuse and simply to annoy contributors and players alike. The whole thirdparty mod infrastructure isn't thought through and has delayed things for months. That missed the chance to hop on the band wagon of the time limited RA2 Freeware promo on Origin. The impractical and currently broken approach at https://github.com/OpenRA/ra2/issues/212 is still a complete disaster. If we never intend to ship and merge the ra2 mod into the regular OpenRA release, then I won't waste my time on this project anymore.

GraionDilach:
Then "don't waste your time on this project anymore".

Seriously. I give up. You know what was a disaster IMO? Phrohdoh/OpenRA. And what was a catastrophe? The current OpenRA/ra2 setup. And those setups were the reason why I only reviewed for months, I was too pissed on the complication level of the setups or how well they make sense. I just knew that noone would actually care about what I thought in the matter, but this time, I'm gonna stand up for once and state it outright that that setup is probably the best, due to being longterm manageable, well-automated, and release-friendly.

I respect what you're doing, @Mailaender, but seriously, when you get feedback like @obrakmann's "It is now clear to me that allowing https://github.com/OpenRA/ra2 to be created was completely wrong, sent the wrong signal and should have been prevented." you should stop pushing for immediate upstream merges and aim for a compromise which doesn't get anyone against your throat. My own solution was bad. You guys stated that, almost immediately. I got it.

If your only concern is that it cannot be merged upstream as it will look like, then I think when the time comes, one can set up another branch where it gets reverted to an OpenRA/OpenRA-appropriate dictionary structure, merge that (via adding OpenRA/OpenRA as a remote, then merge/rebase the whole on top of it) and then PR the whole repo as it is. That's not impossible either. We just end up with three additional commits with the shift. The history still ends up intact.

That PR isn't finished and I had no time to actually debug the mod loader to get it finished, because it's the supportdir-PR which prevents me right now to progress with it further. I intend to finish it. Even better, I intend to shift D2 into that format after RA2 is done. I, again, just haven't found time so far to debug the supportdir-PR to fix why it loads everything but mods from the specified folder. Sorry for not fulfilling your "imminent solutions required" desire.

reaperrr:
>  That missed the chance to hop on the band wagon of the time limited RA2 Freeware promo on Origin.

I still maintain my position that it would have been shallow publicity of little value anyway. Most attention generated through that would have been gone after a few months once they realised how many details and features are still missing and how immature some vital basics of the main OpenRA (sound engine, pathfinding and a lot of smaller things) still are and how slow the progress on some of these aspects is.
And some of the attention remaining might have been counter-productive as well. I still think we didn't lose anything there, because we wouldn't have been able to meet the expectations anytime soon.

>  If we never intend to ship and merge the ra2 mod into the regular OpenRA release, then I won't waste my time on this project anymore.

Who said we don't intend that?
As far as I know, @pchote said some time ago it *will* be merged into the regular OpenRA release, once the legal and mod support prerequisites are met and the mod is closer to completion and in a mergeable state at that point. I think you're just way, *way* to impatient about this.

Mailaender:
Of course I am frustrated. When I read stuff like

> It is now clear to me that allowing https://github.com/OpenRA/ra2 to be created was completely wrong, sent the wrong signal and should have been prevented.

which effectively means thanks for nothing, please throw work of several months in the garbage can and @pchote's overly complicated approaches, which solve no problems (this pull request is here to solve linker errors, nothing else) and just add new ones still feeling overly self-confident quoting his past sayings all over again ignoring everything but his own opinion, I am just annoyed as hell.

As we will stay in this "it will be released when it is done" for a long time, having just a few traits living in this repository won't hurt anyone. The outcome of this idiocy is @GraionDilach copy pasting my code https://github.com/GraionDilach/OpenRA.Mods.AS/commit/c7efe5e293ee40eb85a9ad6a6bf3e21e3f229c06 (at least some credit would be nice) to make it available to thirdparty modders and himself. When I intend to go for the same repository layout approach with https://github.com/OpenRA2/Mods.RA2 to aid the community I get repressed by "branding" issues. That is narcissism at a level that is hard to bear for everyone who just wants to get stuff done.

pchote:
The problems with your proposal and how to fix them have been explained to you multiple times now.  I suggest that you re-evaluate your perspective instead of insulting me and the others who disagree with you.

GraionDilach:
I think I can actually drop the RA2 c'n'p traits after all these go through because I think it will be possible to employ a level of automation on that even.

PR #11306, opened 2016-05-16
Add nod09

ghost:
Add nod09 - "Reinforce Egypt"
http://cnc.wikia.com/wiki/Reinforce_Egypt


mods/cnc/maps/nod09/nod09-AI.lua
```diff
@@ -0,0 +1,220 @@
...
+BaseNuke4 = { type = "nuke", pos = CPos.New(59, 10), cost = 500, exists = true }
+InfantryProduction = { type = "pyle", pos = CPos.New(49, 4), cost = 500, exists = true }
+VehicleProduction = { type = "weap", pos = CPos.New(51, 6), cost = 2000, exists = true }
+
+BaseBuildings = { BaseProc, BaseNuke1, BaseNuke2, BaseNuke3, BaseNuke4, InfantryProduction, VehicleProduction }
+
+AutoGuard = function(guards)
+	Utils.Do(guards, function(guard)
+		Trigger.OnDamaged(guard, function(guard)
+		IdleHunt(guard)
```
abcdefg30: Looks like you're missing one level of indentation here.

mods/cnc/maps/nod09/nod09.lua
```diff
@@ -0,0 +1,275 @@
...
+		player.MarkCompletedObjective(NodObjective2)
+
+		Trigger.AfterDelay(DateTime.Seconds(5), function()
+			Media.PlaySpeechNotification(player, "Reinforce")
+			Reinforcements.ReinforceWithTransport(player, 'tran.in', EngineerReinforcements, { HelicopterEntryEngineer.Location, HelicopterGoalEngineer.Location }, { HelicopterEntryEngineer.Location }, nil, nil)
+		end)
+
+		Trigger.AfterDelay(DateTime.Minutes(1), function()
+			FlareEngineerCamera1.Kill()
+			FlareEngineerCamera2.Kill()
```
abcdefg30: I think it would be better to just `Destroy` the cameras instead.

abcdefg30:
Looks quite good to me, good work. :+1: / :white_check_mark:
Just some minor polish questions:
- Could we let the commando blow up vehicles like we let tanya in ra?
- What about disabling attack move and autotarget for the commando on higher difficulties (like we do in ra with tanya again)? This will prevent the mission from being too easy (aka a-move the commando around the map killing basically all infantry without any difficulty).
- As usually I derped at first and ran into the wrong direction, discovering the abandoned base first (without engineers and thus cash) and consequently failing the mission. Maybe we could use beacons to signalise on where the player is ought to go first? I don't know, maybe that is too much.

Biofreak1987:
There are already beacons that tell you where you are supposed to go? o.o
Also, there shouldn't be a problem if you discover the abandoned base before you meet the engineers. The AI will first start attacking if you capture the refinery, which requires to get the engineers.

abcdefg30:
> There are already beacons that tell you where you are supposed to go? o.o

You mean the flares, right. Somehow forgot about them, they should be fine I think.

> Also, there shouldn't be a problem if you discover the abandoned base before you meet the engineers. The AI will first start attacking if you capture the refinery, which requires to get the engineers.

Yeah well, I somehow aggro'd the medium tank and then (or even before that? I don't remember) sold my construction yard (after building a power plant and barracks, I had exactly enough money for that) like I had to do in the previous missions (although I already figured that I'm probably not supposed to sell it in this mission). With the new cash I trained the engineers. I guess that there is not really much we can do to prevent (stupid) people (like me) from doing that, except someone has a good idea...?

> Not sure about the commando being able to destroy vehicles, I thought there was a reason why he can't do it in td.

We also only allow that in missions for tanya. And you're right, C4ing vehicles is still a little bit glitchy (as are hijackers). I'd vote for doing it in td as well, would be good to hear what @Mailaender and @obrakmann (maybe @pchote as well - I mean, he has an opinion on everything :p) think about it though.

Biofreak1987:
Maybe we could let the player start with 0 money and add the original 1000 when he captures the refinery?

abcdefg30:
That sounds good.

Biofreak1987:
Changed it to that system

Mailaender:
C4 on vehicles wasn't added to multiplayer due to balancing https://github.com/OpenRA/OpenRA/pull/4524#issuecomment-33781190 but that doesn't apply to isolated single player missions where it is an interesting gimmick.

Biofreak1987:
Looking back on that mission, giving the commando the abillity to use c4 on vehicles makes going for the first flare to receive the rocket infantry pretty much pointless.

obrakmann:
My opinion is
:+1: to disabling AttackMove for the commando, and
:-1: to enabling C4 demolition for vehicles. Let's not make Tanya and the commando completely the same.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3335c028ca798aea033736cbe72248bf14fad3fd...d47a15806453d452f08a7031d0e6f680c23a5265)

PR #11308, opened 2016-05-16
Fix IsMoving returning false when activity is a MovePart-derivative

### LINKED ISSUE #11304: IsMoving is returning false for actors who are performing a MoveFirstHalf or MoveSecondHalf
> [21:28] pizzaoverhead Hi, I think I've spotted a bug with ActorExtensionsForMove.IsMoving
[21:29] pizzaoverhead An AI without a conyard is driving an MCV from place to place without deploying, repeatedly checking whether IsMoving before changing its planned build location.
[21:30] pizzaoverhead IsMoving is returning false for actors who are performing a "MoveFirstHalf" or "MoveSecondHalf"

Might also be the underlying issue of #11112 leaving strange "gaps" when turning, same for actors "WithMoveAnimation" sometimes restarting the animation while moving.

reaperrr:
If it works as intended, it should fix #11304.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -825,7 +825,8 @@ namespace OpenRA.Mods.Common.AI
 
 			foreach (var mcv in mcvs)
 			{
-				if (mcv.IsMoving())
+				var mover = mcv.TraitOrDefault<IMove>();
+				if (mover != null && mover.IsMoving)
 					continue;
 
 				var factType = mcv.Info.TraitInfo<TransformsInfo>().IntoActor;
--- a/OpenRA.Mods.Common/Activities/Move/Move.cs
+++ b/OpenRA.Mods.Common/Activities/Move/Move.cs
@@ -432,22 +432,4 @@ namespace OpenRA.Mods.Common.Activities
 			}
 		}
 	}
-
-	public static class ActorExtensionsForMove
-	{
-		public static bool IsMoving(this Actor self)
-		{
-			var a = self.GetCurrentActivity();
-			if (a == null)
-				return false;
-
-			// HACK: Dirty, but it suffices until we do something better:
-			if (a.GetType() == typeof(Move)) return true;
-			if (a.GetType() == typeof(MoveAdjacentTo)) return true;
-			if (a.GetType() == typeof(AttackMoveActivity)) return true;
-
-			// Not a move:
-			return false;
-		}
-	}
 }
--- a/OpenRA.Mods.Common/Traits/Mobile.cs
+++ b/OpenRA.Mods.Common/Traits/Mobile.cs
@@ -168,7 +168,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		public int CalculateTilesetMovementClass(TileSet tileset)
 		{
-			/* collect our ability to cross *all* terraintypes, in a bitvector */
+			// collect our ability to cross *all* terraintypes, in a bitvector
 			return TilesetTerrainInfo[tileset].Select(ti => ti.Cost < int.MaxValue).ToBits();
 		}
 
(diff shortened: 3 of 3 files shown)


RoosterDragon:
Should these two parts of code be doing the same thing? i.e. Should MovePart.Tick also consider MoveAdjacentTo and AttackMoveActivity as "moving"?

What's the difference and why?

obrakmann:
Makes sense, :+1: 

@RoosterDragon: the check in the extension method tests the actor's `CurrentActivity`. With both `MoveAdjacentTo` and `AttackMove`, the `Move` activities are inner activities that never get promoted to be the `CurrentActivity`. The check in the `Tick` method however sets the `IsMoving` flag only when *its* inner method returns a `Move`-related activity. Since this will also happen when the `Move` is actually an inner activity itself, it doesn't need to check for `MoveAdjacentTo` or `AttackMove`, it already works implicitly.

pchote:
Could you please (if you haven't already) investigate whether we can fix this by removing `ActorExtensionsForMove` and using `IMove.IsMoving` instead?  This would be a big win for code sanity.

reaperrr:
@pchote For some reason it seems to break WithInfantryBody (probably other render traits related to Mobile as well), they never stop being considered moving, it seems.

However, other than that there wasn't much blocking this, so if anyone can help me figure out why this breaks and how to fix it, we might as well remove them.

Pushed that commit so you guys can take a look at it.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -867,7 +867,7 @@ void FindAndDeployBackupMcv(Actor self)
 
 			foreach (var mcv in mcvs)
 			{
-				if (mcv.IsMoving())
+				if (mcv.TraitsImplementing<IMove>().Any(m => m.IsMoving))
```
pchote: All the other code assumes at most one IMove, so this could simplify down to a TraitOrDefault followed by a `!= null && .IsMoving`

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -291,7 +291,8 @@ public override IEnumerable<Target> GetTargets(Actor self)
 			return Target.None;
 		}
 
-		abstract class MovePart : Activity
+		// HACK: This is only public to make the ActorExtensionsForMove.IsMoving hack work.
+		public abstract class MovePart : Activity
 		{
```
pchote: This hack can be removed too.

pchote:
@pchote For some reason it seems to break WithInfantryBody (probably other render traits related to Mobile as well), they never stop being considered moving, it seems.

Looks like this was broken by d92537371e91b5a53f2a16a6adb01cbbd4254c9e, not the ActorExtensionsForMove removal.

reaperrr:
Updated.

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -179,14 +179,14 @@ public int GetMovementClass(TileSet tileset)
...
 		{
-			if (!other.IsMoving()) return false;
-
 			var selfMobile = self.TraitOrDefault<Mobile>();
 			if (selfMobile == null) return false;
 
 			var otherMobile = other.TraitOrDefault<Mobile>();
 			if (otherMobile == null) return false;
 
+			if (!otherMobile.IsMoving) return false;
```
abcdefg30: While we're here, could we get the codestyle right?
```
if (!otherMobile.IsMoving)
	return false;
```
or even (although not codestyle)
```
var otherMobile = other.TraitOrDefault<Mobile>();
if (otherMobile == null || !otherMobile.IsMoving)
	return false;
```

obrakmann:
Still looks fine to me, :+1:

pchote:
The `FindAndDeployBackupMcv` check will be removed by #11463, and I just realized that `IsMovingInMyDirection` has been broken for years and can never return true.

That means there is zero risk of this introducing regressions, so :+1: for the code removal.

pchote:
This doesn't change anything user visible, so no changelog needed.

PR #11311, opened 2016-05-17
Revamp custom build times.

GraionDilach:
Removed the `CustomBuildTimeValue` trait, shifted it's properties to `Buildable`, renamed `ProductionQueue.BuildSpeed` to `BuildDurationMultiplier` for consistency.

IMO it's more-modderfriendly to specify a multiplier on the cost-based value than to specify arbitrary, non-scalable/inheritable constants.


OpenRA.Mods.Common/Traits/CustomBuildTime.cs
```diff
@@ -13,27 +13,36 @@
...
+			var csv = a.TraitInfoOrDefault<CustomBuildTimeInfo>();
+			var cost = a.HasTraitInfo<ValuedInfo>() ? a.TraitInfo<ValuedInfo>().Cost : 0;
+
 			if (csv != null)
-				return csv.Value;
+			{
+				if (csv.Value > 0)
+					return csv.Value;
+
+				cost *= csv.Multiplier / 100;
```
Phrohdoh: You'll probably want to guard against a `0` numerator.

GraionDilach: Why? That would result with an instant building. Sounds like intended.

pchote:
The build time code has always been a horrible mess.  The parameters for calculating the actual build time are spread over four(!) different traits and an `Actor` extension method.

I think you have the right idea, but propose taking it further.  We should remove the `CustomBuildTimeValue` trait, the `CustomBuildTimeExts` class, `ProductionQueueInfo.BuildSpeed`, and define the following on `Buildable`:

```
[Desc("Base build time (-1 indicates use the unit's Value)")
public readonly int BuildDuration = -1;

[Desc("Percentage modifier to apply to the build duration")]
public readonly int BuildDurationModifier = 40;
```

This gives you the same configuration, but with all the important things (except the actor value) in one place.

GraionDilach:
I don't understand why should this deprecate `ProductionQueueInfo.BuildSpeed` tbh. I'll look into the rest tomorrow, but these two aren't interchangeable to me - there might be cases where an actor might want to be trained in two queues and one might be a slower queue than the other (individual-queue-based satrport vs war factory f. e.).

pchote:
Starport vs war factory use two different actor types with a special hack for group select.

GraionDilach:
I'm aware. Which is why I said individual-queue-based.

What I meant is that there might be cases, when a modder might want to have a limited-availability-but-faster queue (buildimited/expensive queue provider) and a slower-but-common queue (cheaper/nonbuildlimited queue provider) running along. without cloning each unit between all the queues - something the D2k starport does atm.

pchote:
Sure, but that use case is rare / edge-casey enough that it is cleaner for that specific case to define new inherited actors instead having an extra redundant parameter for all actors in all mods.

GraionDilach:
Actually, we already do this in the d2k starport. https://github.com/OpenRA/OpenRA/blob/bleed/mods/d2k/rules/player.yaml#L42

pchote:
Indeed, but it could just as easily be defined as an additional override on the already-overriden `Buildable` trait for each starport actor.

GraionDilach:
I assume you want the modifier adjust the fixed value as well then? Because the D2K vehicles also rely on fixed buildtimes as well.

pchote:
Yes – the idea is to reduce the current configuration mess down to two parameters (with one having a magic value to say "use this third parameter") that work consistently across all configuration setups.

GraionDilach:
I'm still against this, though, because YR also had the queue-based buildspeed multiplier and the unit-based-buildspeed multiplier this PR intends to introduce - the WW games before that only had queue-based buildspeed.

And honestly, I can't see what's messy in that setup.

pchote:
Right now we have this situation (assuming the non-classic production queue – yet another factor is included for the production speedup on the classic version):

```
No custom build time:
duration = 1500 * Valued.Cost * ProductionQueue.BuildSpeed / 100000;

Custom build time, multiplier > 0
duration =  CustomBuildTime.Value * ProductionQueue.BuildSpeed / 100;

Custom build time, multiplier <= 0
duration = 1500 * Valued.Cost * CustomBuildTime.Multiplier * ProductionQueue.BuildSpeed / 100000;
```

This is needlessly complicated and does not solve any problems that can't be covered by:

```
duration = (CustomBuildTime.Value OR Valued.Cost) * CustomBuildTime.Multiplier / 100
```

plus additional actor types (which already exist for our one mod that uses this feature) if you want to have actors with different build times on different queues.

pchote:
Note for people using email notifications: my original comment had a bunch of code typos which have been corrected.

GraionDilach:
I understand and I intend to simplify that constant. However, I will not shoot myself in the leg by removing a feature I rely on in my own PR.

Right now, I'm heavily relying on `ProductionQueueInfo.BuildSpeed` on infantry, because I rely on really cheap and slower trained infantry, to downplay zergspam tactics. (I'm also using it to speed up the vehicle training a bit for this same reason.) However, I'm running into an issue that since high-tier units require more money, they end up being trained too slow so that lowtier stuff can easily outspam them. Since this is a global issue, I thought that introducing a tier-based buildspeed bonus would counter that issue of mine.

With your solution, I cannot fulfill my original aim of expanding my already-established techlevel template with the buildspeed bonuses from the techlevel, and would force me to introduce another 16 templates for my actors to get their buildspeed from (since atm I have 4 distinct unit queues and 4 techlevels and I plan to have more queues even) which I'm not a fan of, personally.

pchote:
I don't want to kill what is otherwise a good cleanup over a philosophical point, so i'll back off on the `BuildSpeed` removal.

Can you please implement the other changes, so that we have a consistent calculation of
```
duration = (Buildable.BuildDuration OR Valued.Cost) * Buildable.BuildDurationModifier * ProductionQueue.BuildSpeed / 10000;
```
?

GraionDilach:
Sure, I'll do that tomorrow.

GraionDilach:
Updated.

OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
```diff
@@ -71,7 +71,19 @@ public int BuildTime(string type)
...
-			return ai.GetBuildTime();
+			var bi = ai.TraitInfoOrDefault<BuildableInfo>();
+
+			if (bi != null)
+			{
+				var time = bi.BuildDuration != -1
+					? (ai.HasTraitInfo<ValuedInfo>() ? ai.TraitInfo<ValuedInfo>().Cost : 0)
+					: bi.BuildDuration;
+
+				time *= bi.BuildDurationMultiplier / 100;
```
pchote: This is integer division, so you need to do it *after* the multiplication: `time = time * bi.BuildDurationMultiplier / 100`.

GraionDilach: Thought *= and * are on the same priority in terms of operation resolution. >_>

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -35,7 +35,7 @@ public class ProductionQueueInfo : ITraitInfo
 		public readonly bool Sticky = true;
 
 		[Desc("This percentage value is multiplied with actor cost to translate into build time (lower means faster).")]
-		public readonly int BuildSpeed = 40;
+		public readonly int BuildDurationMultiplier = 40;
```
pchote: Please set this to 100 and then merge the 40 into Buildable's modifier.

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -311,15 +311,19 @@ public void ResolveOrder(Actor self, Order order)
...
 		}
 
-		public virtual int GetBuildTime(ActorInfo unit, BuildableInfo bi = null)
+		public virtual int GetBuildTime(ActorInfo unit, BuildableInfo bi)
 		{
 			if (developerMode.FastBuild)
 				return 0;
 
-			var time = unit.GetBuildTime() * Info.BuildSpeed / 100;
+			var time = bi.BuildDuration != -1
```
pchote: Looks like you have this check inverted.

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -311,15 +311,19 @@ public void ResolveOrder(Actor self, Order order)
...
+				? (unit.HasTraitInfo<ValuedInfo>() ? unit.TraitInfo<ValuedInfo>().Cost : 0)
+				: bi.BuildDuration;
+
+			time *= bi.BuildDurationMultiplier * Info.BuildDurationMultiplier / 10000;
```
pchote: Likewise here for `*=` vs `/`

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
GraionDilach: That doesn't sound auto-upgradable, though.

GraionDilach:
Updated.

Right now, the upgrade rules shouldn't change the status of the mods. I don't think there is a way to completely look up the proper queue speed multiplier from the unit in `UpgradeRules`, but if you think outputting a warning is enough, I'll do it.

OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
```diff
@@ -71,7 +71,19 @@ public int BuildTime(string type)
 			if (!Context.World.Map.Rules.Actors.TryGetValue(type, out ai))
 				throw new LuaException("Unknown actor type '{0}'".F(type));
 
-			return ai.GetBuildTime();
+			var bi = ai.TraitInfoOrDefault<BuildableInfo>();
+
+			if (bi != null)
```
pchote: Minor style nit: the convention we use in cases like this is to check for null and return zero here, and then have the main logic unindented

OpenRA.Mods.Common/Traits/Buildable.cs
```diff
@@ -37,6 +37,12 @@ public class BuildableInfo : TraitInfo<Buildable>
 		[Desc("Palette used for the production icon.")]
 		[PaletteReference] public readonly string IconPalette = "chrome";
 
+		[Desc("Base build time (-1 indicates to use the unit's Value)")]
+		public readonly int BuildDuration = -1;
+
+		[Desc("Percentage modifier to apply to the build duration")]
+		public readonly int BuildDurationMultiplier = 150;
```
pchote: Can you double check the naming conventions we use in other cases? IIRC these should be Modifier not Multiplier (because they dont multiply)

pchote:
Yes, i agree that printing a warning and doing a manual fixup is the best approach.

OpenRA.Mods.Common/Traits/Buildable.cs
GraionDilach: Apparently even the `*Multiplier` traits call their actions Modifiers internally, so I guess I do have to switch it to Modifier..

OpenRA.Mods.Common/Traits/Player/ClassicProductionQueue.cs
```diff
@@ -120,15 +120,15 @@ protected override bool BuildUnit(ActorInfo unit)
 
 		public override int GetBuildTime(string unitString)
 		{
-			return GetBuildTime(self.World.Map.Rules.Actors[unitString]);
+			return GetBuildTime(self.World.Map.Rules.Actors[unitString], self.World.Map.Rules.Actors[unitString].TraitInfo<BuildableInfo>());
 		}
 
-		public override int GetBuildTime(ActorInfo unit, BuildableInfo bi = null)
+		public override int GetBuildTime(ActorInfo unit, BuildableInfo bi)
```
abcdefg30: You'll have to fix https://github.com/OpenRA/OpenRA/pull/11311/files#diff-910e0462c780643261207e2829e4041dR135 then as well (we don't need to check for `bi` being `null` anymore).

OpenRA.Mods.Common/Traits/Player/ClassicProductionQueue.cs
```diff
@@ -120,15 +120,15 @@ protected override bool BuildUnit(ActorInfo unit)
 
 		public override int GetBuildTime(string unitString)
 		{
-			return GetBuildTime(self.World.Map.Rules.Actors[unitString]);
+			return GetBuildTime(self.World.Map.Rules.Actors[unitString], self.World.Map.Rules.Actors[unitString].TraitInfo<BuildableInfo>());
```
abcdefg30: This looks bad (doing the lookup twice). What about:
```
var actorInfo = self.World.Map.Rules.Actors[unitString];
return GetBuildTime(actorInfo, actorInfo.TraitInfo<BuildableInfo>());

File: OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs [hunk 10]
```diff
@@ -311,15 +311,19 @@ public void ResolveOrder(Actor self, Order order)
 
 		public virtual int GetBuildTime(string unitString)
 		{
-			return GetBuildTime(self.World.Map.Rules.Actors[unitString]);
+			return GetBuildTime(self.World.Map.Rules.Actors[unitString], self.World.Map.Rules.Actors[unitString].TraitInfo<BuildableInfo>());
```
abcdefg30: Same here.

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -311,15 +311,19 @@ public void ResolveOrder(Actor self, Order order)
...
 
-		public virtual int GetBuildTime(ActorInfo unit, BuildableInfo bi = null)
+		public virtual int GetBuildTime(ActorInfo unit, BuildableInfo bi)
 		{
 			if (developerMode.FastBuild)
 				return 0;
 
-			var time = unit.GetBuildTime() * Info.BuildSpeed / 100;
+			var time = bi.BuildDuration == -1
+				? (unit.HasTraitInfo<ValuedInfo>() ? unit.TraitInfo<ValuedInfo>().Cost : 0)
```
abcdefg30: Isn't it the convention to use `TraitOrDefault` instead?
```
var time = bi.BuildDuration;
if (time == -1)
{
	var valued = unit.TraitOrDefault<ValuedInfo>();
	if (valued != null)
		time = valued.Cost;
}

Comment (GraionDilach) [inline OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs]: I copied this over and probably the segment predates `TraitOrDefault`.

**Comment by GraionDilach** [comment]
Updated.

**Comment by RoosterDragon** [comment]
Needs a rebase.

**Comment by GraionDilach** [comment]
Rebased.

**Comment by GraionDilach** [comment]
Rebased again.

File: OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs [hunk 12]
```diff
@@ -71,7 +71,20 @@ public int BuildTime(string type)
...
+				return 0;
+
+			var time = bi.BuildDuration;
+			if (time == -1)
+			{
+				var valued = ai.TraitInfoOrDefault<ValuedInfo>();
+				time = valued != null ? valued.Cost : 0;
+			}
+
+			time = time * bi.BuildDurationModifier / 100;
```
abcdefg30: This ignores `BuildDurationModifier` of `ProductionQueue`. I don't think that is a good idea.
If you fetch the `ProductionQueue` trait anyway, you can also get rid of the code duplication.

GraionDilach: It doesn't fetch the prodqueue.

abcdefg30: Yeah, that is the point. Shouldn't you do that (to get `BuildDurationModifier` from it)?

GraionDilach: Where should it know which queue it needs to fetch? I'm following a position here that the actor can be produced from multiple queues which have different queue multipliers here.

pchote: This is a pretty strong argument favour of my original "remove the per-queue build modifier" request.

abcdefg30: Hm right, that's a problem. At the moment this makes the method inaccurate (doesn't it?), we might to include the building from which the actor is produced as parameter as well then?

GraionDilach: TBH, `ProductionQueue.BuildDurationModifier` is ignored here even in bleed.

@pchote: If you try to force me to get it removed again, then I'm gonna close this PR. I'm not gonna work against myself. The per-queue modifier existed in all the WW games while the per-unit one arrived in YR.

pchote: It's not really acceptable to have a Lua API that returns demonstrably wrong values when using traits in a supported way, so this will beed to be resolved in *some* way.  You could look up the default queue type and include an optional param to override if there are multiple.

GraionDilach:
Rebased and updated.

OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
```diff
@@ -64,14 +64,52 @@ public Actor Create(string type, bool addToWorld, LuaTable initTable)
...
+			var bi = ai.TraitInfoOrDefault<BuildableInfo>();
+
+			if (bi == null)
+				return 0;
+
+			var time = bi.BuildDuration;
+			if (time == -1)
+			{
+				var valued = ai.TraitInfoOrDefault<ValuedInfo>();
+				time = valued != null ? valued.Cost : 0;
```
abcdefg30: We should already `return 0` here if `valued == null` (or `valued.Cost == 0`).
That'll save us all the lookups from below.

abcdefg30:
Needs a rebase again.

OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
```diff
@@ -64,14 +64,52 @@ public Actor Create(string type, bool addToWorld, LuaTable initTable)
...
+			}
+
+			int pbi;
+			if (producer != null)
+			{
+				ActorInfo prodi;
+				if (!Context.World.Map.Rules.Actors.TryGetValue(producer, out prodi))
+					throw new LuaException("Unknown actor type '{0}'".F(type));
+
+				var pqueue = prodi.TraitInfos<ProductionQueueInfo>().Where(x => bi.Queue.Contains(x.Type)).FirstOrDefault();
```
abcdefg30: You can simplify this to `var pqueue = prodi.TraitInfos<ProductionQueueInfo>().FirstOrDefault(x => bi.Queue.Contains(x.Type))`.

abcdefg30: `x.Type` is also not correct: `Eluant.LuaException: Actor 'tent' cannot produce actor 'mech'!`.
You would need to check `x`'s `Produces` trait. For example, `tent` has
```
	Production:
		Produces: Infantry, Soldier
```
and `mech`
```
	Buildable:
		Queue: Infantry
		BuildAtProductionType: Soldier
```

abcdefg30:
Sorry, I know it's nasty...
The lua method doesn't seem to work for TD.

OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
GraionDilach: In RA the player actor is the producer, not the buildings.

GraionDilach:
How did you tested this on TD?

```diff
--- a/mods/cnc/rules/structures.yaml
+++ b/mods/cnc/rules/structures.yaml
@@ -28,6 +28,7 @@ FACT:
 		LowPowerSlowdown: 2
 		QueuedAudio: Building
 		ReadyAudio: ConstructionComplete
+		BuildDurationModifier: 200 #default 100
 	ProductionQueue@NodBuilding:
 		Type: Building.Nod
 		Factions: nod
@@ -35,6 +36,7 @@ FACT:
 		LowPowerSlowdown: 2
 		QueuedAudio: Building
 		ReadyAudio: ConstructionComplete
+		BuildDurationModifier: 200 #default 100
 	ProductionQueue@GDIDefense:
 		Type: Defence.GDI
 		Factions: gdi
@@ -42,6 +44,7 @@ FACT:
 		LowPowerSlowdown: 3
 		QueuedAudio: Building
 		ReadyAudio: ConstructionComplete
+		BuildDurationModifier: 200 #default 100
 	ProductionQueue@NodDefense:
 		Type: Defence.Nod
 		Factions: nod
@@ -49,6 +52,7 @@ FACT:
 		LowPowerSlowdown: 3
 		QueuedAudio: Building
 		ReadyAudio: ConstructionComplete
+		BuildDurationModifier: 200 #default 100
 	BaseBuilding:
 	ProductionBar@BuildingGDI:
 		ProductionType: Building.GDI
@@ -103,6 +107,7 @@ NUKE:
 		BuildPaletteOrder: 10
 		Prerequisites: fact
 		Queue: Building.GDI, Building.Nod
+		BuildDurationModifier: 40 #default 60
 	Building:
 		Footprint: xx xx
 		Dimensions: 2,2
diff --git a/OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs b/OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
index 8ebdf81..d404ca2 100644
--- a/OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
@@ -109,6 +112,9 @@ public int BuildTime(string type, string producer = null)
			}
			time = time * bi.BuildDurationModifier * pbi / 10000;
+
+			Game.Debug("{0} has a build time of {1}.", type, time);
+
			return time;
		}
```

returns me 400 for the power plant with the yaml changes and 300 without in Nod07a.

GraionDilach:
Rebased, updated.

OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
GraionDilach: Reworded the parameter description to explicitly mention this needs the actor with the producing queue and not the visual producer.

abcdefg30:
I get `Eluant.LuaException: No actors can produce actor 'e1'!` when trying `Media.DisplayMessage("e " .. Actor.BuildTime("e1"))` on gdi01.

GraionDilach:
Fixed.

GraionDilach:
Rebased.

OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
```diff
@@ -64,14 +64,55 @@ public Actor Create(string type, bool addToWorld, LuaTable initTable)
 			return a;
 		}
 
-		[Desc("Returns the build time (in ticks) of the requested unit type.")]
-		public int BuildTime(string type)
+		[Desc("Returns the build time (in ticks) of the requested unit type."
+			+ "An optional second value can be used to exactly specify the actor with the producing queue.")]
+		public int BuildTime(string type, string producer = null)
```
pchote: This should accept the queue name instead of an actor.  There's no reason why an actor couldn't produce the same actor from two different queues with their own time modifiers (e.g. a combined war-factory / starport).

OpenRA.Mods.Common/Traits/Player/ClassicProductionQueue.cs
```diff
@@ -120,19 +120,20 @@ protected override bool BuildUnit(ActorInfo unit)
 
 		public override int GetBuildTime(string unitString)
 		{
-			return GetBuildTime(self.World.Map.Rules.Actors[unitString]);
+			var actorInfo = self.World.Map.Rules.Actors[unitString];
```
pchote: This overload of `GetBuildTime` is only called from one place, so you might as well inline this lookup there and remove the method.

OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
GraionDilach: ProductionQueue->Type or ->Group?

GraionDilach:
Updated.

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
```diff
@@ -425,13 +428,14 @@ public ProductionItem(ProductionQueue queue, string item, int cost, PowerManager
...
 			this.pm = pm;
+			ai = Queue.Actor.World.Map.Rules.Actors[Item];
 		}
 
 		public void Tick(PlayerResources pr)
 		{
 			if (!Started)
 			{
-				var time = Queue.GetBuildTime(Item);
+				var time = Queue.GetBuildTime(ai, ai.TraitInfo<BuildableInfo>());
```
obrakmann: Please cache this as well in the constructor.

OpenRA.Mods.Common/Scripting/Global/ActorGlobal.cs
```diff
@@ -64,14 +64,53 @@ public Actor Create(string type, bool addToWorld, LuaTable initTable)
 			return a;
 		}
 
-		[Desc("Returns the build time (in ticks) of the requested unit type.")]
-		public int BuildTime(string type)
+		[Desc("Returns the build time (in ticks) of the requested unit type."
+			+ "An optional second value can be used to exactly specify the producing queue type.")]
```
obrakmann: When you split lines with a `+`, you need to leave a space either at the end of the preceding or the beginning of the following string. When you use a `,` to split strings, you don't.

OpenRA.Mods.Common/Traits/Buildable.cs
```diff
@@ -37,6 +37,12 @@ public class BuildableInfo : TraitInfo<Buildable>
 		[Desc("Palette used for the production icon.")]
 		[PaletteReference] public readonly string IconPalette = "chrome";
 
+		[Desc("Base build time (-1 indicates to use the unit's Value).")]
```
obrakmann: Please specify the unit of measurement.

obrakmann:
Looks fine to me, aside from those minor nits. :+1:

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
GraionDilach: Can't do this one, I think. Caching this in the constructor would mean that classic queue speedups would be applied at the moment of queuing the actor instead of when the building starts.

GraionDilach:
Updated.

OpenRA.Mods.Common/Traits/Player/ProductionQueue.cs
obrakmann: I meant caching the `ai.TraitInfo<BuildableInfo>()`, not the time.

GraionDilach: Oh.

GraionDilach:
Updated again.

reaperrr:
Looks fine to me, however I'm not that familiar with Lua in general so I'd like to see some feedback from @abcdefg30 on that specific part.
:+1: for the rest after that.

GraionDilach:
NTM: This will need a rebase before merging to fix the commit typo in https://github.com/OpenRA/OpenRA/pull/11311/commits/a0f44d10d551f2e8585a74f568e3a4ae058315ea

GraionDilach:
Rebased and fixed the commit typo.

GraionDilach:
Rebased again.

obrakmann:
Sorry Graion, you'll need to rebase again. I'll merge this right after you do.

GraionDilach:
Done. During rebase, I also fixed the outdated-by-more-than-a-month upgrade rule date of #11177.

abcdefg30:
The Lua part looks fine to me.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/68609d247c2d9d9d8bb40cecd31251fcb63e4ec7)

PR #11316, opened 2016-05-18
Add Scott's Allies02 multiplayer mission as "Evacuation"

obrakmann:
Converted from the [C# mission script found in release-20131223](https://github.com/OpenRA/OpenRA/blob/release-20131223/OpenRA.Mods.RA/Missions/Allies02Script.cs).

It follows the original script pretty closely, there are only a few minor deviations. That also means that the difficulty is quite high, and I'll probably end up introducing a new "Easy" skill level and shifting the other two up accordingly. I'll need some feedback on that, however.

// cc: @ScottNZ


mods/ra/maps/evacuation/evacuation.lua
```diff
@@ -0,0 +1,365 @@
...
+	if allies2.IsLocalPlayer and DateTime.GameTime <= ReinforcementsDelay then
+		UserInterface.SetMissionText("Allied reinforcements arrive in " .. Utils.FormatTime(ReinforcementsDelay - DateTime.GameTime))
+	end
+end
+
+SetupSoviets = function()
+	soviets.Cash = 1000
+
+	if Map.Difficulty == "Easy" then
+		for i, sam in ipairs(Sams) do
```
abcdefg30: Why didn't you use `Utils.Do` here?

mods/ra/maps/evacuation/evacuation.lua
```diff
@@ -0,0 +1,365 @@
...
+
+	if Map.Difficulty ~= "Easy" and allies1.IsLocalPlayer then
+		Trigger.AfterDelay(DateTime.Seconds(2), function()
+			Media.DisplayMessage("According to the rules of engagement I need your explicit orders to fire, Commander!", "Tanya")
+		end)
+	end
+end
+
+ReassignActors = function(actors, from, to)
+	for i, a in ipairs(actors) do
```
abcdefg30: Same here: `Utils.Do(actors, function(a)...`

mods/ra/maps/evacuation/evacuation.lua
```diff
@@ -0,0 +1,365 @@
...
+
+	-- Allies1 is the player starting on the right, controlling Tanya
+	allies1 = Player.GetPlayer("Allies1")
+
+	-- Allies2 is the player starting on the left, defending the base
+	allies2 = Player.GetPlayer("Allies2")
+
+	soviets = Player.GetPlayer("Soviets")
+
+	for i, player in ipairs({ allies1, allies2 }) do
```
abcdefg30: And same question here.

OpenRA.Mods.Common/Traits/Player/MissionObjectives.cs
```diff
@@ -215,7 +215,10 @@ public void OnPlayerLost(Player player)
 
 					if (Info.EarlyGameOver)
 						foreach (var p in enemies)
-							p.PlayerActor.Trait<MissionObjectives>().ForceDefeat(p);
+						{
```
abcdefg30: Doesn't our code style also require braces around the `if` now?

mods/ra/maps/evacuation/evacuation.lua
```diff
@@ -0,0 +1,365 @@
...
+	TownUnit01, TownUnit02, TownUnit03, TownUnit04, TownUnit05, TownUnit06, TownUnit07,
+	TownUnit08, TownUnit09, TownUnit10, TownUnit11, TownUnit12, TownUnit13, TownUnit14
+}
+
+ParabombDelay = DateTime.Seconds(30) 
+ParatroopersDelay = DateTime.Minutes(5)
+Paratroopers =
+{
+	{
+		-- `proxy` cannot be assigned in global scope since a player is needed to create the actor
```
abcdefg30: Why do we need to have the actor inside this array anyway?
I think using the name would be better:
```
Paratroopers =
{
	{
		proxy = "powerproxy.paras1",
		entry = BadgerEntryPoint1.Location,
		drop  = BadgerDropPoint1.Location,
	},
[...]
```
And then in `SendParatroopers` (as we use this only once anyway):
```
Utils.Do(Paratroopers, function(para)
	local units = Actor.Create(para.proxy, false, { Owner = soviets }).SendParatroopersFrom(para.entry, para.drop)
	Utils.Do(units, function(unit)
[...]

File: mods/ra/maps/evacuation/evacuation.lua [hunk 6]
```diff
@@ -0,0 +1,365 @@
...
+	{
+		proxy = nil,
+		entry = BadgerEntryPoint1.Location + CVec.New(3, 0),
+		drop  = BadgerDropPoint2.Location,
+	},
+	{
+		proxy = nil,
+		entry = BadgerEntryPoint1.Location + CVec.New(6, 0),
+		drop  = BadgerDropPoint3.Location,
+	},
```
abcdefg30: I'm not sure what our convention is here. I never add a comma behind the last element of an array. That is up to you though, I won't really complain about it (much). I just wondered as this is even inconsistently done in this single script.

mods/ra/maps/evacuation/evacuation.lua
```diff
@@ -0,0 +1,365 @@
...
+		return a.Owner == allies1 and not a.IsDead and a ~= Einstein and a ~= Tanya and a ~= Engineer
+	end)
+
+	if (#targets > 0) then
+		yak.Attack(Utils.Random(targets))
+	end
+	yak.Move(Map.ClosestEdgeCell(yak.Location))
+	yak.Destroy()
+	Trigger.OnRemovedFromWorld(Yak, function()
+		Yak = nil
```
abcdefg30: Hm, I don't think we really need this, we could just add an `Yak.IsInWorld` check at the `OnEnteredProximityTrigger` instead: `if Yak.IsInWorld or actor.Owner ~= allies1 `

mods/ra/maps/evacuation/evacuation.lua
abcdefg30: Ah wait, we might want to destroy the proxy actor as we don't use it again anyway, so I'd suggest following instead:
```
Utils.Do(Paratroopers, function(para)
    local proxy = Actor.Create(para.proxy, false, { Owner = soviets })
    local units = proxy.SendParatroopersFrom(para.entry, para.drop)
    proxy.Destroy()
    Utils.Do(units, function(unit)
[...]

File: mods/ra/maps/evacuation/evacuation.lua [hunk 8]
```diff
@@ -0,0 +1,365 @@
...
+		unit.AttackMove(TownPoint.Location)
+		Trigger.OnIdle(unit, unit.Hunt)
+	end)
+end
+
+SendParabombs = function()
+	local proxy = Actor.Create("powerproxy.parabombs", false, { Owner = soviets })
+	proxy.SendAirstrikeFrom(BadgerEntryPoint2.Location, ParabombPoint1.Location)
+	proxy.SendAirstrikeFrom(BadgerEntryPoint2.Location + CVec.New(0, 3), ParabombPoint2.Location)
+end
```
abcdefg30: Similiar thing here: Just destroy the proxy actor when we don't need it anymore.

mods/ra/maps/evacuation/evacuation.lua
```diff
@@ -0,0 +1,365 @@
...
+
+SendAttackGroup = function()
+	if #AttackGroup < AttackGroupSize then
+		return
+	end
+
+	Utils.Do(AttackGroup, function(unit)
+		if not unit.IsDead then
+			unit.Hunt()
```
abcdefg30: I guess you *could* remove this if you feel like, as you assign an "IdleHunt" trigger to the unit the line below anyway.

mods/ra/maps/evacuation/evacuation.lua
```diff
@@ -0,0 +1,365 @@
...
+	AttackGroup = { }
+end
+
+ProduceInfantry = function()
+	if SovietBarracks.IsDead or SovietBarracks.Owner ~= soviets then
+		return
+	end
+
+	soviets.Build({ Utils.Random(SovietInfantry) }, function(units)
+		unit = units[1]
```
abcdefg30: `local unit` please, but I'd drop that entirely and just use `table.insert(AttackGroup, units[1])` instead below.

mods/ra/maps/evacuation/evacuation.lua
```diff
@@ -0,0 +1,365 @@
...
+	end)
+end
+
+ProduceVehicles = function()
+	if SovietWarFactory.IsDead or SovietWarFactory.Owner ~= soviets then
+		return
+	end
+
+	soviets.Build({ Utils.Random(SovietVehicles[SovietVehicleType]) }, function(units)
+		unit = units[1]
```
abcdefg30: Same here.

mods/ra/maps/evacuation/evacuation.lua
```diff
@@ -0,0 +1,365 @@
...
+				Media.PlaySpeechNotification(player, "MissionAccomplished")
+			end)
+
+			Trigger.OnPlayerLost(player, function()
+				Media.PlaySpeechNotification(player, "MissionFailed")
+			end)
+		end
+	end
+
+	if allies2 == nil or allies2.IsLocalPlayer then
```
abcdefg30: We usually don't explicitly check for `nil`. I'd just use `if not allies2 or allies2.IsLocalPlayer then`.

mods/ra/maps/evacuation/evacuation.lua
```diff
@@ -0,0 +1,365 @@
...
+		Camera.Position = Allies2BasePoint.CenterPosition
+	else
+		Camera.Position = ChinookHusk.CenterPosition
+	end
+
+	if allies2 == nil then
```
abcdefg30: Same here. (`if not allies2 then`)

mods/ra/maps/evacuation/rules.yaml
```diff
@@ -0,0 +1,182 @@
+Player:
+	MissionObjectives:
+		Cooperative: True
+	PlayerResources:
+		DefaultCash: 5000
+
+World:
+	MissionData:
+		Briefing: One of our outposts is under heavy attack by a nearby Soviet base. Einstein, Tanya and many of our forces have been evacuated via helicopters, but those were shot down on the other side of the river. Fortunately they survived the crash, but are now in enemy territory. Tanya and the rest of the surviving forces need to fight their way out of there. The air defenses must be eliminated before another Search & Rescue helicopter can arrive.\n\nMeanwhile, the forces remaining at the base need to fend off the Soviet attacks.
```
abcdefg30: Could you add another `\n` at the end, please? This increases readability:
evacuationdesca
vs
evacuationdescb

abcdefg30:
Some issues I noticed during my first playthrough:
- You can see following units but will never be able to get their prerequisites: Gap Generator, Chronosphere, Iron Curtain, Missile Silo, M.A.D Tank, MiG, Longbow, Missile Sub
- An exception from above is the Flamethrower. You can - in theory - get the required prerequisite by capturing a flame tower... I think here we should just make flame towers buildable.
- Just minor: It feels a bit weird that Heavy Tanks require a service depot and Medium Tanks don't.
- Einstein has the same selection priority as normal soldiers while the engineer has not, this is somehow confusing imho.
- If you play alone you get "Mission Accomplished", but the game just continues.
- Just for bonus points: I had to search a bit before I found the flare and the helicopter I need to send Einstein to. What about placing a beacon there?
- To "That also means that the difficulty is quite high [...]": Well, not too, imho. It was quite easy to snipe ever soldier with tanya as they don't come for you like they do in the allies03 missions, you could add that for extra bonus points. :D

mods/ra/maps/evacuation/evacuation.lua
obrakmann: I'll keep it this way since it's easier to code. `Yak` would still be nil when the trigger first fires, so I need to check for it anyway.

obrakmann:
Updated.

Re: the points in your last comments:

1. Fixed.
2. Made FTUR buildable.
3. Yeah. Do you think I should add a FIX to the base?
4. I would very much like to revert the selection priority change for the engineer, it's totally annoying. Shouldn't be part of this PR, though.
5. That was due to the dummy 'Allies' actor not being marked as a non-combatant, I'm pretty sure.
6. Beacon added.
7. Yeah, no. You're right, on easy it's quite doable.

abcdefg30:
> 3. Yeah. Do you think I should add a FIX to the base?

Hm, I'm not really convinced. Would that work good together with power management? And it would make it easier to not loose 100 units. So it' rather say don't add a FIX.

> 7. Yeah, no. You're right, on easy it's quite doable.

I talked about normal, but ok. :p

obrakmann:
Forgot to lock the tech level and debug lobby options previously, fixed now.

cjshmyr:
Haven't checked the code, but map played well :+1:

Mailaender:
Nicely done. Works well (singleplayer) :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/62f1cfcba470dca435f3928bf6a73f6c6d32d086...d752175cacec624e46ca4dbb35c118b1713c21c7)

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/62f1cfcba470dca435f3928bf6a73f6c6d32d086...d752175cacec624e46ca4dbb35c118b1713c21c7)

mods/ra/maps/evacuation/map.yaml
```diff
@@ -0,0 +1,1757 @@
...
+
+Author: Scott_NZ
+
+Tileset: SNOW
+
+MapSize: 128,128
+
+Bounds: 16,16,96,96
+
+Visibility: MissionSelector, Lobby
```
pchote: Maps should be one or the other, not both.

Mailaender: These are Co-Op mission which can be played single-player as well so this entry is valid. Our UI is just not capable enough. There is even a TODO in the code where it crashes.

pchote: Having a thing appear in two different places that otherwise display completely disjoint content is terrible UX.

Mailaender: For me the way to solve https://github.com/OpenRA/OpenRA/issues/11692 is to
* add a multi-selection dropdown
* disable UI support for the setting
* make it a simple text edit widget

I wouldn't recommend just avoid using values and I don't see how having this on both the skirmish lobby and the mission selector contributes to negative UX. It just makes it easier to discover.

obrakmann: But it doesn't make sense to duplicate the whole map just with different Visibility entries, either. And the map *is* both single- and multi-player capable. I was very careful to allow for that especially.

Either way, the fix is easy: http://stackoverflow.com/a/8317808

mods/ra/maps/evacuation/rules.yaml
```diff
@@ -0,0 +1,217 @@
...
+		Prerequisites: ~disabled
+
+MSUB:
+	Buildable:
+		Prerequisites: ~disabled
+
+CAMERA.Large:
+	Inherits: CAMERA
+	RevealsShroud:
+		Range: 1000c0
```
Mailaender: This would trigger:

```
Exception of type `System.ArgumentOutOfRangeException`: The requested range (101) cannot exceed the value of MaximumTileSearchRange (50)
Parameter name: maxRange
```

if actually used in-game.

obrakmann: It's not used, it's just [a left-over from the original map](https://github.com/OpenRA/OpenRA/blob/release-20131223/mods/ra/maps/allies-02/map.yaml#L1932-L1935). Also, this PR is merged, no need to review ;)

Mailaender: I just left a note for me and others as I am using your rule port and scripts as templates. https://github.com/OpenRA/OpenRA/pull/11679

Phrohdoh: If it isn't used it should have been removed.

Mailaender: Sorry, the time travel DeLorean is out of fuel.

PR #11327, opened 2016-05-20
Make TS not load unnecessary mix files

reaperrr:
- gmenu.mix only contains the menu orverride from Patch 2.0/Firestorm
- maps0*.mix only contain singleplayer maps 
- multi.mix only contains multiplayer maps
- patch.mix only contains ini files.

None of these are used or even usable directly by OpenRA as-is, so it's pointless to load them.


GraionDilach:
:+1:

obrakmann:
I'll take your word for it. Either way, those aren't included in our asset download anyway. :+1:

PR #11337, opened 2016-05-22
[RA] Buff mig and longbow AG weapons

ghost:
So from [various](http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19498&start=75&postdays=0&postorder=asc&highlight=) RA balance [threads](#10437) there seems to be a consensus that the MiG and Longbow could do with some buffs. This PR does NOT introduce mig AA weaponry. I think @SoScared made [some very good points](https://github.com/OpenRA/OpenRA/issues/10437#issuecomment-171100185) about the possible balance affects of giving Migs AA.

I am instead suggesting to balance these units by fixing some of the issues I have discovered through testing, which in turn should give them a better defined role in the game. With these changes, they should be pretty good investments for sniping mobile armour while matching their cheaper counterparts vs buildings.

**Issue 1.** In the current release the longbow anti ground missile has one cell less range than the hind chaingun. Now both are equal. 

**Issue 2.** Missiles for both the MiG and the longbow would explode just after their max range. This made damaging moving targets nearly impossible without somewhat fiddly micro. This has now been fixed. Rangers are the only land unit fast enough to evade rockets fairly reliably while APCs can travel the furthest before being hit.

**Issue 3.** A single hind and yak in the current release does more damage to _nearly every building in the game_ compared to a single Mig or LB. Now Migs and LBs do slightly more damage overall to most buildings compared to their cheaper counterparts and twice the damage to concrete walls as before.

**Balance 1.** The longbow now fires in a volley of 4 rockets instead of 2, buts its cool-down between volleys has been increased by 40%. This is still an overall buff and makes the LB more useful for hit and run. This will hopefully make Allied late game a little more interesting.

### DIFF
--- a/mods/ra/weapons/missiles.yaml
+++ b/mods/ra/weapons/missiles.yaml
@@ -112,7 +112,7 @@ HellfireAG:
 
 HellfireAA:
 	ReloadDelay: 60
-	Range: 4c0
+	Range: 5c0
 	MinRange: 1c256
 	Report: missile6.aud
 	Burst: 2
@@ -126,7 +126,7 @@ HellfireAA:
 		Inaccuracy: 128
 		Image: DRAGON
 		HorizontalRateOfTurn: 10
-		RangeLimit: 4c819
+		RangeLimit: 7c0
 	Warhead@1Dam: SpreadDamage
 		Spread: 128
 		Damage: 40
--- a/mods/ra/weapons/missiles.yaml
+++ b/mods/ra/weapons/missiles.yaml
@@ -15,16 +15,15 @@ Maverick:
 		Image: DRAGON
 		HorizontalRateOfTurn: 5
 		CruiseAltitude: 2c0
-		RangeLimit: 10c819
+		RangeLimit: 14c410
 	Warhead@1Dam: SpreadDamage
 		Spread: 128
 		Damage: 70
 		ValidTargets: Ground, Water
 		Versus:
 			None: 30
-			Wood: 75
-			Light: 75
-			Concrete: 50
+			Wood: 90
+			Light: 90
 		DamageTypes: Prone50Percent, TriggerProne, ExplosionDeath
 	Warhead@2Smu: LeaveSmudge
 		SmudgeType: Crater
@@ -76,11 +75,11 @@ Dragon:
 
 HellfireAG:
 	ReloadDelay: 60
-	Range: 4c0
+	Range: 5c0
 	MinRange: 1c256
 	Report: missile6.aud
-	Burst: 2
-	BurstDelay: 10
+	Burst: 4
+	BurstDelay: 14
 	ValidTargets: Ground, Water
 	Projectile: Missile
 		Speed: 256
@@ -90,16 +89,15 @@ HellfireAG:
 		Inaccuracy: 128
 		Image: DRAGON
 		HorizontalRateOfTurn: 10
-		RangeLimit: 4c819
+		RangeLimit: 8c512
 	Warhead@1Dam: SpreadDamage
 		Spread: 128
 		Damage: 60
 		ValidTargets: Ground, Water
 		Versus:
 			None: 30
-			Wood: 75
-			Light: 75
-			Concrete: 50
+			Wood: 90
+			Light: 90
 		DamageTypes: Prone50Percent, TriggerProne, ExplosionDeath
 	Warhead@2Smu: LeaveSmudge
 		SmudgeType: Crater


obrakmann:
Regarding issue 3: has anyone already suggested nerfing the hind and yak against buildings? Seems sensible to me, they're pretty strong in that regard.

HenrytheSlav:
It was sugested on forums, but people seem against it. It looks like buff of MiG/Longbow is more supported than nerf of 1st tier aircraft.

As for the balance point, can we also make MiG release more rockets against buildings or something? Idk, their hit and run damage is not that great atm.

r34ch:
Here is a map to test these changes
http://resource.openra.net/maps/14235/

I would strongly advise testing these changes for yourselves before suggesting more buffs. I think these changes make these units fairly viable. 

That said, If more buffs are required then as per usual, people are free to make the changes. I just feel a playtest would be welcome before more adjustments are made.

As to why I didn't buff the Mig @HenrytheSlav, with good micro you can avoid damage and greatly increase DPS even in its current release state due to the Migs range and turn rate. The current release longbow is just a sitting duck, which is why I buffed it. I tried just reducing the delay between bursts alone but I didn't think it would really help with survivability in comparison to the Mig.

With a higher burst salvo, you get some interesting micro options and it makes the Mig and Longbow more contrasting units, which I feel fits the overall balance theme of oRA. The LB becomes a somewhat high alpha strike unit that has to tank briefly while the Mig can use distance and deal more continuous damage over time.

HenrytheSlav:
@r34ch At last monday stream with @SoScared there was some testing of balance, I assume your changes, since it was related to aircraft, and it was found that if longbow has two weapons with different ranges, f.e. AA 5 cells and ground 6 cells, then it will fail to shoot/move closer when at 6 cells distance against aircraft because it tries to use wrong weapon.

r34ch:
Would anyone be against a extra cell increase to the AA distance then?

r34ch:
Buffed the range of the AA weapon by 1 cell to match the range of the AG weapon. Buffed the max range of the AA weapon by 2 cells so it should be a little easier to hit retreating planes.

mods/ra/weapons/missiles.yaml
```diff
@@ -90,16 +89,15 @@ HellfireAG:
 		Inaccuracy: 128
 		Image: DRAGON
 		HorizontalRateOfTurn: 10
-		RangeLimit: 4c819
+		RangeLimit: 8c819
```
GraionDilach: 6c0.

Right now, all RangeLimits are 1.2*Range and not arbitrary numbers even if they seem as such at first sight.

GraionDilach: Okay, i reread the PR and understand issue 2 - 20% additional range might not be enough. Please consider picking a relative percentage instead of an arbitrary value regardless.

mods/ra/weapons/missiles.yaml
```diff
@@ -15,16 +15,15 @@ Maverick:
 		Image: DRAGON
 		HorizontalRateOfTurn: 5
 		CruiseAltitude: 2c0
-		RangeLimit: 10c819
+		RangeLimit: 14c819
```
GraionDilach: Same here.

mods/ra/weapons/missiles.yaml
```diff
@@ -128,7 +126,7 @@ HellfireAA:
 		Inaccuracy: 128
 		Image: DRAGON
 		HorizontalRateOfTurn: 10
-		RangeLimit: 4c819
+		RangeLimit: 6c819
```
GraionDilach: And here.

mods/ra/weapons/missiles.yaml
r34ch: @GraionDilach Ah ok that makes much more sense. That said, 4 x 1.2 = 4.8 exactly. What maths am I missing to get the 4c8**19** part. (I assume c simply stands for cells)

Also I assume I can mix up the percentages if need be? I.e. longbow max range is 1.7 normal range while mig max range is 1.4 normal range (example percentages)

GraionDilach: 0c1024 = 1c0. Feel free, as long as you document the percentages at least in the PR or the commit message.

r34ch:
Updated the max range values

 - Maverick 1.2 -> 1.6
 - HellfireAG 1.2 -> 1.7
 - HellfireAA 1.2 -> 1.4

obrakmann:
Let's see how this plays out. Would be nice to see more migs again. :+1:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/9b53ef7be1227bba6d665ff2654b27fdea459c60...d3a7cc98c5fc01bc7b9fc3ff03f9a4d8d31e0eee)

PR #11347, opened 2016-05-23
Fix the bogus getAirstrikeTarget methods of the nod08 missions

abcdefg30:
They always searched 6 times even after finding a target.
Now they also account for SAM sites (like in the previous missions).

And yes, you can indeed get SAM sites in those missions, it isn't too easy though and thus I'd like the player to be rewarded for getting them.

### DIFF
--- a/mods/cnc/maps/nod08a/nod08a.lua
+++ b/mods/cnc/maps/nod08a/nod08a.lua
@@ -50,12 +50,18 @@ getAirstrikeTarget = function()
 
 	local target = list[DateTime.GameTime % #list + 1].CenterPosition
 
-	if searches < 6 then
+	local sams = Map.ActorsInCircle(target, WDist.New(8 * 1024), function(actor)
+		return actor.Type == "sam" end)
+
+	if #sams == 0 then
+		searches = 0
+		return target
+	elseif searches < 6 then
 		searches = searches + 1
 		return getAirstrikeTarget()
 	else
 		searches = 0
-		return target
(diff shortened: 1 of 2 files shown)


obrakmann:
:+1:

pchote:
This doesn't seem to work.  The AI will still destroy my base in the corner when it is protected by SAMs.

abcdefg30:
> This doesn't seem to work. The AI will still destroy my base in the corner when it is protected by SAMs. (built using cheats)

That one airstrike is entirely scripted to destroy the base, iirc.

Biofreak1987:
I don't mind if you rework that scripted airstrike to take sams into account, I've scripted the mission back then without the knowledge that you can actually build sams.

pchote:
I don't think it's a big deal.  I don't know whether it's even possible to get a sam built in time to stop that first strike.

abcdefg30:
> That one airstrike is entirely scripted to destroy the base, iirc.

https://github.com/OpenRA/OpenRA/blob/bleed/mods/cnc/maps/nod08a/nod08a.lua#L138 "hardcodes" this airstrike to hit only the base when you captured the outpost (you can't build SAMs before you captured it).

pchote:
Appears to work as advertised, so :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1089e9fc7310ca19100f4babee06302540558ceb

PR #11353, opened 2016-05-24
Add liblua.so path for raspberry pi.

hoall:
At least raspbian uses a different path for liblua. Adding this
path allows further development to run OpenRA on Raspbian.

It compiles without errors. It does not start though.

### DIFF
--- a/thirdparty/configure-native-deps.sh
+++ b/thirdparty/configure-native-deps.sh
@@ -6,7 +6,7 @@
 # Copy-paste the entire script into http://shellcheck.net to check.
 ####
 
-locations="/lib /lib64 /usr/lib /usr/lib64 /usr/lib/i386-linux-gnu /usr/lib/x86_64-linux-gnu /usr/local/lib /opt/lib"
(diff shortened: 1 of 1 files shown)


Hoall:
In addition I want to say that, in my opinion, it doesn't run because of the experimental OpenGL support. This should change in the next months.

[graphics.log.txt](https://github.com/OpenRA/OpenRA/files/280216/graphics.log.txt)

obrakmann:
It doesn't run because the raspi currently only supports OpenGL ES, while we require full OpenGL. There's an experimental branch with GL ES support, but the raspi [only manages about two FPS](http://logs.openra.net/?year=2016&month=02&day=03#15:43:26).

For this reason, I don't think we should take this, there would be no use, until there's a more powerful raspi3, perhaps.

pchote:
@obrakmann: there is a more powerful raspi3, but I very much doubt it will significantly improve the FPS.

obrakmann:
Oh, I didn't know that. But yeah, seems like it's been out for three months now.

Anyway, closing here. Sorry, @Hoall, but thanks for your interest!

PR #11348, opened 2016-05-23
Remove redundant Effects

reaperrr:
`SpriteEffect` can (now) do everything that `Explosion`, `Corpse` and `Smoke` could do, so I scrapped those three to get rid of some redundancy and duplication.

Additionally, the only thing in `Game` that used `SpriteEffect` was the `WorldInteractionControllerWidget`, which itself is not required by anything in `Game`, so I moved both up to `Mods.Common`.

PR #11358, opened 2016-05-26
Moved some more things from Game to Mods.Common

reaperrr:
Follow-up to #11348.

Suggestions about what else can/should be moved from engine to mod-level are welcome.


GraionDilach:
RejectsOrders? :D

reaperrr:
Updated.

GraionDilach:
Looks good. :+1:

PR #11361, opened 2016-05-26
Add new lua functions and fix the air attacks in allies05a

abcdefg30:
Following functions are added:

_AircraftProperties_
- void Land(Actor landOn)
- void Resupply()
- ~~Actor GetActorBelow()~~

_AmmoPoolProperties_
- int AmmoCount(string poolName = "primary")
- ~~bool HasFullAmmo(string poolName = "primary")~~
- ~~bool HasAmmo(string poolName = "primary")~~
- void Reload(string poolName = "primary", int amount = 1)
- int MaximumAmmoCount(string poolName = "primary") 

I also made the `Attack` lua method available for non-mobile actors.
As bonus: I fixed the air attacks in `allies05a` and this is the prerequisite for `allies04`.

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Properties/AircraftProperties.cs
+++ b/OpenRA.Mods.Common/Scripting/Properties/AircraftProperties.cs
@@ -46,5 +46,19 @@ namespace OpenRA.Mods.Common.Scripting
 			else
 				Self.QueueActivity(new HeliReturnToBase(Self));
 		}
+
+		[ScriptActorPropertyActivity]
+		[Desc("Queues a landing activity on the specififed actor.")]
+		public void Land(Actor landOn)
+		{
+			Self.QueueActivity(new Land(Self, Target.FromActor(landOn)));
+		}
+
+		[ScriptActorPropertyActivity]
+		[Desc("Starts the resupplying activity when being on a host building.")]
+		public void Resupply()
+		{
+			Self.QueueActivity(new ResupplyAircraft(Self));
+		}
 	}
 }
--- a/OpenRA.Mods.Common/Scripting/Properties/AmmoPoolProperties.cs
+++ b/OpenRA.Mods.Common/Scripting/Properties/AmmoPoolProperties.cs
@@ -1,0 +1,75 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using Eluant;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Scripting;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Scripting
+{
+	[ScriptPropertyGroup("AmmoPool")]
+	public class AmmoPoolProperties : ScriptActorProperties, Requires<AmmoPoolInfo>
+	{
+		readonly Actor self;
+		readonly AmmoPool[] ammoPools;
+
+		public AmmoPoolProperties(ScriptContext context, Actor self)
+			: base(context, self)
+		{
+			this.self = self;
+			ammoPools = self.TraitsImplementing<AmmoPool>().ToArray();
+		}
+
+		[Desc("Returns the count of the actor's specified ammopool.")]
+		public int AmmoCount(string poolName = "primary")
+		{
+			var pool = ammoPools.FirstOrDefault(a => a.Info.Name == poolName);
+			if (pool == null)
+				throw new LuaException("Invalid ammopool name {0} queried on actor {1}.".F(poolName, self));
+
+			return pool.GetAmmoCount();
+		}
+
+		[Desc("Returns the maximum count of ammo the actor can load.")]
+		public int MaximumAmmoCount(string poolName = "primary")
+		{
+			var pool = ammoPools.FirstOrDefault(a => a.Info.Name == poolName);
+			if (pool == null)
+				throw new LuaException("Invalid ammopool name {0} queried on actor {1}.".F(poolName, self));
+
+			return pool.Info.Ammo;
+		}
+
+		[Desc("Adds the specified amount of ammo to the specified ammopool.",
+			"(Use a negative amount to remove ammo.)")]
+		public void Reload(string poolName = "primary", int amount = 1)
+		{
+			var pool = ammoPools.FirstOrDefault(a => a.Info.Name == poolName);
+			if (pool == null)
+				throw new LuaException("Invalid ammopool name {0} queried on actor {1}.".F(poolName, self));
+
+			if (amount > 0)
+			{
+				while (amount-- > 0)
+					if (!pool.GiveAmmo())
+						return;
+			}
+			else
+			{
+				while (amount++ < 0)
+					if (!pool.TakeAmmo())
+						return;
+			}
+		}
+	}
+}
--- a/mods/ra/maps/allies-05a/allies05a-AI.lua
+++ b/mods/ra/maps/allies-05a/allies05a-AI.lua
@@ -243,25 +243,31 @@ ProduceAircraft = function()
 	end
 
 	ussr.Build(SovietAircraftType, function(units)
-		Yaks[#Yaks + 1] = units[1]
+		local yak = units[1]
+		Yaks[#Yaks + 1] = yak
 
-		if #Yaks == 2 then
-			Trigger.OnKilled(units[1], ProduceAircraft)
-		else
+		Trigger.OnKilled(yak, ProduceAircraft)
+		if #Yaks == 1 then
 			Trigger.AfterDelay(DateTime.Minutes(1), ProduceAircraft)
 		end
 
 		local target = nil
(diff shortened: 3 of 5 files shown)


pchote:
Please don't expose GetActorBelow to lua. It is a horrible hack that needs to die, and exposing it to lua makes that much harder.  If you are using it to get actors placed in the editor then you can use actor tags instead.

abcdefg30:
Okay, removed it.

OpenRA.Mods.Common/Scripting/Properties/AmmoPoolProperties.cs
```diff
@@ -0,0 +1,86 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
```
obrakmann: Header is outdated.

mods/ra/maps/allies-05a/allies05a-AI.lua
```diff
@@ -243,25 +243,32 @@ ProduceAircraft = function()
 	end
 
 	ussr.Build(SovietAircraftType, function(units)
-		Yaks[#Yaks + 1] = units[1]
+		local yak = units[1]
+		Yaks[#Yaks + 1] = yak
 
 		if #Yaks == 2 then
-			Trigger.OnKilled(units[1], ProduceAircraft)
+			Trigger.OnKilled(yak, ProduceAircraft)
```
obrakmann: This will only put the trigger on the second yak, not the first one. I can't imagine that's what you had in mind.

mods/ra/maps/allies-05a/allies05a-AI.lua
```diff
@@ -243,25 +243,32 @@ ProduceAircraft = function()
...
 				local enemies = Utils.Where(Map.ActorsInWorld, function(self) return self.Owner == greece and self.HasProperty("Health") end)
 				if #enemies > 0 then
 					target = Utils.Random(enemies)
-					units[1].Attack(target)
+				else
+					yak.Wait(DateTime.Seconds(5))
 				end
+			end
+
+			if yak.HasAmmo() then
```
obrakmann: I guess this should be `if target and yak.HasAmmo() then`, just to be safe. `target` is guaranteed to be set here (although it probably will be).

obrakmann:
Aside from those few nits, looks fine to me, :+1:

mods/ra/maps/allies-05a/allies05a-AI.lua
abcdefg30: Well, derp. Thanks.

abcdefg30:
Sorry for the late update, but here we go (better late than never, eh? :p).
Note: Test commit is still in to make reviewing for the 2nd reviewer easier (but don't merge straight!).

OpenRA.Mods.Common/Scripting/Properties/AmmoPoolProperties.cs
```diff
@@ -0,0 +1,87 @@
...
+
+		public AmmoPoolProperties(ScriptContext context, Actor self)
+			: base(context, self)
+		{
+			this.self = self;
+			ammoPools = self.TraitsImplementing<AmmoPool>().ToArray();
+		}
+
+		[Desc("Returns the count of the actor's specified ammopool.")]
+		public int GetAmmoCount(string poolName = "primary")
```
pchote: Can we call this `AmmoCount`? We don't use Get* in our lua API function names.

OpenRA.Mods.Common/Scripting/Properties/AmmoPoolProperties.cs
```diff
@@ -0,0 +1,87 @@
...
+		{
+			var pool = ammoPools.FirstOrDefault(a => a.Info.Name == poolName);
+			if (pool == null)
+				throw new LuaException("Invalid ammopool name {0} queried on actor {1}.".F(poolName, self));
+
+			return pool.GetAmmoCount();
+		}
+
+		[Desc("Returns true if the actor's specified ammopool is full.")]
+		public bool FullAmmo(string poolName = "primary")
```
pchote: `CanReload` (returning the negative) or `HasFullAmmo` would be more in line with our API conventions.

OpenRA.Mods.Common/Scripting/Properties/AmmoPoolProperties.cs
```diff
@@ -0,0 +1,87 @@
...
+		{
+			var pool = ammoPools.FirstOrDefault(a => a.Info.Name == poolName);
+			if (pool == null)
+				throw new LuaException("Invalid ammopool name {0} queried on actor {1}.".F(poolName, self));
+
+			return pool.HasAmmo();
+		}
+
+		[Desc("Adds the specified amount of ammo to the specified ammopool.")]
+		public void GiveAmmo(string poolName = "primary", int amount = 1)
```
pchote: This doesn't really fit either - how about `Reload`?

OpenRA.Mods.Common/Scripting/Properties/AmmoPoolProperties.cs
```diff
@@ -0,0 +1,87 @@
...
+			if (pool == null)
+				throw new LuaException("Invalid ammopool name {0} queried on actor {1}.".F(poolName, self));
+
+			while (amount-- > 0)
+				if (!pool.GiveAmmo())
+					return;
+		}
+
+		[Desc("Removes the specified amount of ammo from to specified ammopool.")]
+		public void TakeAmmo(string poolName = "primary", int amount = 1)
```
pchote: Can we Reload a negative amount instead?

abcdefg30:
Updated.

OpenRA.Mods.Common/Scripting/Properties/AircraftProperties.cs
```diff
@@ -46,5 +46,19 @@ public void ReturnToBase(Actor airfield = null)
 			else
 				Self.QueueActivity(new HeliReturnToBase(Self));
 		}
+
+		[ScriptActorPropertyActivity]
+		[Desc("Queues a landing activity on the specififed actor.")]
+		public void Land(Actor landOn)
+		{
+			Self.QueueActivity(new Land(Self, Target.FromActor(landOn)));
```
Phrohdoh: Shouldn't you check `landOn` for validity? (InWorld, !Dead, etc)

mods/ra/maps/survival02/survival02.lua
```diff
@@ -397,4 +397,9 @@ WorldLoaded = function()
 	InitObjectives()
 	InitMission()
 	SetupSoviets()
+
+	Trigger.AfterDelay(0, function()
+		-- The turret most to the right will attack an AA gun
```
Phrohdoh: `the rightmost turret`

mods/ra/maps/allies-03b/allies03b.lua
```diff
@@ -366,4 +366,59 @@ WorldLoaded = function()
...
 	InitTriggers()
 	SetupAlliedUnits()
+
+	AmmoPoolTest()
+end
+
+AmmoPoolTest = function()
+	local yak = Actor.Create("yak", true, { Owner = player, Location = Tanya.Location + CVec.New(1, 1) })
+
+	Media.DisplayMessage("Ammo count: " .. yak.AmmoCount())
```
Phrohdoh: It would be good if you could also display the max value so the full ammo checks are easier to make sense of.

mods/ra/maps/survival02/survival02.lua
abcdefg30: This is just inside the test commit, and doesn't matter thus.
Thanks though. :)

Phrohdoh: Oh, right.

OpenRA.Mods.Common/Scripting/Properties/AircraftProperties.cs
abcdefg30: Isn't that already done with `target.IsValidFor(self)` at https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Activities/Air/Land.cs#L31

abcdefg30:
Actually, I just noticed that `HasFullAmmo` and `HasAmmo` are kind of redundant. I could/should just expose a `MaximumAmmoCount` and then use `AmmoCount == MaximumAmmoCount` or `AmmoCount > 0`. (`MaximumAmmoCount` makes it also easier to calculate the count for `Reload` [when you want e.g. only 50% reloaded].)

Opinions?

mods/ra/maps/allies-03b/allies03b.lua
abcdefg30: Added.

abcdefg30:
Updated (the testcommit as well).

Removed `HasFullAmmo` and `HasAmmo`, added `MaximumAmmoCount`.

Phrohdoh:
The test works as expected and the code looks fine.
I'd personally rename `Reload` to `AddAmmo`, but that isn't necessary. 👍

abcdefg30:
Removed the testcase. This should be ready for merge now.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/0490b9752049cf3a68f9f27d8142f632e6b5d730)

PR #11363, opened 2016-05-27

RoosterDragon:
Allowing VS to format this file as it likes means users of this IDE don't need to reset it each time VS decides to streamroll over it with arbitrary changes.

Updates version number to VS2015 Update 2. (See also #10669)

Sorts files in its preferred order (See also #9979, #9696)


GraionDilach:
:+1:

Issue #4122, opened 2013-11-10
Remember lobby option checkbox states


xan2622:
I think it could be interesting if Map Options (shroud, fog of war, crates appear...) could be remembered after modificating them.

This way, a player could create new games according to his preferences (and without always having to re-enable them each time a new game is created).

Same for the Debug (Developer Mode), checkboxes could be saved too.

abcdefg30:
Just in case someone wants to work on this:

> 21:19:01  (abcdefg30) you could fix #4122
21:19:14  (pchote) please don't
21:19:17  (abcdefg30) ?
21:19:23  (pchote) that issue is full of subtlety
21:20:16  (pchote) its bad enough when switching to a map with custom rules dumps out the default bots
21:20:58  (abcdefg30) and it doesn't change e.g. your faction if you change from a map with custom rules to a normal one...
21:21:11  (abcdefg30) well ok so fixing ^ is required before fixing #4122?
21:21:37  (pchote) the problem is when you switch to a map that (for example) force disables shroud, and then back to a normal map
21:21:46  (pchote) if you remembered the last state, then your new map will start with shroud disabled
21:21:53  (pchote) or starting units configured, or whatever
21:22:08  (pchote) which is something that the host had manually set
21:22:22  (pchote) and is asking for endless bug reports from people who are confused about why their settings changed
21:22:52  (pchote) *which is not something that was manually set
21:23:46  (hamb) +1
21:24:07  (abcdefg30) well, we shouldn't remember the settings of custom maps anyway
21:24:11  (abcdefg30) as i said ^
21:24:11  (Scott\`) that's lame
21:24:21  (Scott\`) we already solved that issue for player colours
21:24:23  (abcdefg30) you can get the "allied" faction from custom maps on a normal map 
21:24:39  (abcdefg30) hm, Scott\` what was the solution again? i don't remember
21:24:57  (Scott\`) have 2 colours, a colour preference and the actual colour
21:25:14  (Scott\`) map chooses the actual colour or lets you use your preference
21:25:23  (abcdefg30) ok, so same to settings?
21:25:26  (Scott\`) using the colour picker sets the colour preference
21:25:35  (abcdefg30) sounds good to me
21:26:56  (Scott\`) i even implemented ^ for the same reason
21:27:06  (Scott\`) i got annoyed that my colour wasnt persisting after switching to a mission map
21:27:08  (Scott\`) :P

pchote:
[...]

PR #11176, opened 2016-04-24
Query lobby checkbox state via their owning traits.

pchote:
This continues the work started by #10857 to migrate from hardcoded lobby options to a more flexible trait-driven system.

The direct checks against the boolean `World.LobbyInfo.GlobalSettings.*` values are replaced with checks against the respective checkbox-owning traits.  A future PR (see [pchote:lobby-trait-options](https://github.com/pchote/OpenRA/commits/lobby-trait-options) for a WIP preview) will then be able to swap out the `INotifyCreated.Created` implementations without having to deal with all of these outside references.

PR #11364, opened 2016-05-27
Unhardcode the lobby options backend.

pchote:
This follows on from #11176 and #10857, taking us another step towards mod/map defined lobby options.

All of the options except for map difficulty and game speed are migrated to a new mod-level options order, and the traits are set up to use these.  Difficulty and game speed require nontrivial changes to integrate with the new system, so these will be done with their own followup PRs.  After those are complete we can implmement the dynamically generated options bin UI.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Player/ProvidesTechPrerequisite.cs
+++ b/OpenRA.Mods.Common/Traits/Player/ProvidesTechPrerequisite.cs
@@ -48,7 +48,8 @@ namespace OpenRA.Mods.Common.Traits
 		public ProvidesTechPrerequisite(ProvidesTechPrerequisiteInfo info, ActorInitializer init)
 		{
 			this.info = info;
-			enabled = info.Name == init.World.LobbyInfo.GlobalSettings.TechLevel;
+			var mapOptions = init.World.WorldActor.TraitOrDefault<MapOptions>();
+			enabled = mapOptions != null && mapOptions.TechLevel == info.Id;
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Traits/Player/ProvidesTechPrerequisite.cs
+++ b/OpenRA.Mods.Common/Traits/Player/ProvidesTechPrerequisite.cs
@@ -15,7 +15,14 @@ namespace OpenRA.Mods.Common.Traits
 {
 	public class ProvidesTechPrerequisiteInfo : ITechTreePrerequisiteInfo
 	{
+		[Desc("Internal id for this tech level.")]
+		public readonly string Id;
+
+		[Translate]
+		[Desc("Name shown in the lobby options.")]
 		public readonly string Name;
+
+		[Desc("Prerequisites to grant when this tech level is active.")]
 		public readonly string[] Prerequisites = { };
 
 		public object Create(ActorInitializer init) { return new ProvidesTechPrerequisite(this, init); }
@@ -23,8 +30,9 @@ namespace OpenRA.Mods.Common.Traits
 
 	public class ProvidesTechPrerequisite : ITechTreePrerequisite
 	{
-		ProvidesTechPrerequisiteInfo info;
+		readonly ProvidesTechPrerequisiteInfo info;
 		bool enabled;
+
 		static readonly string[] NoPrerequisites = new string[0];
 
 		public string Name { get { return info.Name; } }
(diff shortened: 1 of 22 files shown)


pchote:
This PR half-fixes #4122: options are remembered across map changes, but forgotten if you leave / re-enter the lobby.

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -347,73 +347,31 @@ void CloseWindow()
...
+			var optionCheckboxes = new Dictionary<string, string>()
+			{
+				{ "EXPLORED_MAP_CHECKBOX", "explored" },
+				{ "CRATES_CHECKBOX", "crates" },
+				{ "SHORTGAME_CHECKBOX", "shortgame" },
+				{ "FOG_CHECKBOX", "fog" },
+				{ "ALLYBUILDRADIUS_CHECKBOX", "allybuild" },
+				{ "ALLOWCHEATS_CHECKBOX", "cheats" },
+				{ "CREEPS_CHECKBOX", "creeps" },
+			};
```
obrakmann: Doesn't this kind of defeat the purpose of unhardcording the options?

Before reading this PR, my expectation was that each trait that provided a lobby option would contain fields for type (checkbox or dropdown) and precedence (for ordering in the UI), and the UI would then built from template definitions.

Is that going to happen in the future, or is this the final state of the solution?

OpenRA.Mods.Common/Traits/Player/ProvidesTechPrerequisite.cs
```diff
@@ -40,7 +48,8 @@ public IEnumerable<string> ProvidesPrerequisites
 		public ProvidesTechPrerequisite(ProvidesTechPrerequisiteInfo info, ActorInitializer init)
 		{
 			this.info = info;
-			enabled = info.Name == init.World.LobbyInfo.GlobalSettings.TechLevel;
+			var mapOptions = init.World.WorldActor.TraitOrDefault<MapOptions>();
+			enabled = mapOptions != null && mapOptions.TechLevel == info.Name;
```
obrakmann: `info.Id` (sometimes I think you always introduce one stupid error in your PRs just to see if people read them :p)

obrakmann:
Looks fine otherwise, :+1:

pchote:
Fixed and rebased.

OpenRA.Mods.Common/Traits/World/MapCreeps.cs
```diff
@@ -15,22 +15,36 @@
...
 	{
 		[Desc("Default value of the creeps checkbox in the lobby.")]
 		public readonly bool Enabled = true;
 
 		[Desc("Prevent the creeps state from being changed in the lobby.")]
 		public readonly bool Locked = false;
+
+		IEnumerable<LobbyOption> ILobbyOptions.LobbyOptions(Ruleset rules)
+		{
+			yield return new LobbyBooleanOption("creeps", "Worms", Enabled, Locked);
```
Mailaender: How about exposing "Worms" to the trait info so this becomes configurable and more general?

Mailaender: Doing it everywhere has the extra benefit that the field can become `[Translateable]` as well.

Mailaender:
This looks very sane architectural wise :+1: and I can't spot regressions in-game. :white_check_mark:

OpenRA.Mods.Common/Traits/World/MapCreeps.cs
pchote: That's the plan, but a bunch of other changes need to happen for the dynamic UI (e.g. the `bool Locked` needs to change to an enum for `Configurable, Locked, Hidden` for all traits) so this is left for a future PR.

reaperrr:
I consider @obrakmann's +1 still valid since only the fixup and rebase changed since then, so this already has +2. I couldn't find anything wrong either, so I think this can be merged.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5a025816b473eeccf823b0e806343b6cde24930e...e95b192f856205878dec5cded43309c5d08b1527)

PR #11334, opened 2016-05-22
Fixed various offsets in TS

reaperrr:
And enabled muzzle flash on Titan.

PR #11367, opened 2016-05-28
Fixed various offsets in TS

reaperrr:
Successor to #11334.

No changes, I just messed up #11334 while cleaning my repo.

### DIFF
--- a/mods/ts/rules/gdi-support.yaml
+++ b/mods/ts/rules/gdi-support.yaml
@@ -101,7 +101,7 @@ GACTWR:
 		UpgradeTypes: tower.vulcan
 		UpgradeMinEnabledLevel: 1
 		Weapon: VulcanTower
-		LocalOffset: 768,85,512
+		LocalOffset: 416,85,960
 		MuzzleSequence: muzzle
 		MuzzleSplitFacings: 8
 	Armament@VULCSECONDARY:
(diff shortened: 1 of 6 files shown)


mods/ts/rules/gdi-support.yaml
```diff
@@ -101,27 +101,27 @@ GACTWR:
...
 		UpgradeMinEnabledLevel: 1
 		Weapon: RPGTower
-		LocalOffset: 512,-128,512
+		LocalOffset: 192,-65,1056
 	Armament@SAM:
 		UpgradeTypes: tower.sam
 		UpgradeMinEnabledLevel: 1
 		Weapon: RedEye2
-		LocalOffset: 512,0,512
+		LocalOffset: 192,0,1152
```
pchote: This looks wrong ingame to me - it comes from the middle of the turret instead of the front.

mods/ts/rules/gdi-vehicles.yaml
```diff
@@ -158,8 +158,11 @@ MMCH:
 		Recoils: no
 	Armament:
 		Weapon: 120mm
+		MuzzleSequence: muzzle
 		Recoil: 127
 		RecoilRecovery: 26
+		LocalOffset: 512, 224, 784
```
pchote: This also looks quite far offset:
screen shot 2016-05-28 at 18 33 38

pchote: You should check whether the facing fudge has been disabled.

reaperrr:
Updated:
- fixed Titan offset, `ClassicFacingFudge` was indeed not disabled
- fixed GDI SAM turret firing offset

Additionally,
- fixed Nod SAM Site offsets (sequences and armament)
- fixed `InitialFacing` of all defense to face north-east (was considered "north" in original TS)

reaperrr:
I just realised the InitialFacings need to be 224, not 192. Will fix that tomorrow/later today.

reaperrr:
Updated.

reaperrr:
@Mailaender Since I added the InitialFacing fixes after your review, I guess you should check that commit to validate the :+1:

pchote:
Looks fine to me now.  If you enable the combat overlay the firing positions for the sam site aren't right, but that cancels out the effect of the otherwise badly positioned smoke trail so i'm not going to block the pr over that.

reaperrr:
>  If you enable the combat overlay the firing positions for the sam site aren't right, but that cancels out the effect of the otherwise badly positioned smoke trail so i'm not going to block the pr over that.

Yeah, that's intentional.

Mailaender:
Thanks. Looking good. :white_check_mark:

Phrohdoh:
An incorrect overlay is intentional?

reaperrr:
@Phrohdoh No, the larger-than-necessary distance of the firing offsets from the turret is intentional, because it looks better in practice.

PR #11368, opened 2016-05-28
Add nod07c

ghost:
Add nod07c - "ORCA Theft"
http://cnc.wikia.com/wiki/ORCA_Theft


Mailaender:
I got a crash after attacking the lone grenadier. He came after me, ran over Tiberium and that causes it to be damaged by a null actor location:

```
Fatal Lua Error: [string "BindingSupport.lua"]:30: Uncaught CLR exception at Lua->CLR boundary: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an object
  at OpenRA.Actor.get_Location () [0x00007] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Actor.cs:60 
  at OpenRA.Mods.Common.Scripting.GeneralProperties.get_Location () <0x41592890 + 0x0001b> in <filename unknown>:0 
  at (wrapper delegate-invoke) <Module>:invoke_callvirt_CPos_GeneralProperties (OpenRA.Mods.Common.Scripting.GeneralProperties)
  at System.Reflection.MonoProperty.GetterAdapterFrame[T,R] (System.Reflection.Getter`2 getter, System.Object obj) [0x00000] in /home/abuild/rpmbuild/BUILD/mono-4.2.3/mcs/class/corlib/System.Reflection/MonoProperty.cs:329 
  at System.Reflection.MonoProperty.GetValue (System.Object obj, System.Object[] index) [0x0008e] in /home/abuild/rpmbuild/BUILD/mono-4.2.3/mcs/class/corlib/System.Reflection/MonoProperty.cs:397 
  --- End of inner exception stack trace ---
  at System.Reflection.MonoProperty.GetValue (System.Object obj, System.Object[] index) [0x000a1] in /home/abuild/rpmbuild/BUILD/mono-4.2.3/mcs/class/corlib/System.Reflection/MonoProperty.cs:399 
  at OpenRA.Scripting.ScriptMemberWrapper.Get (Eluant.LuaRuntime runtime) [0x00041] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Scripting/ScriptMemberWrapper.cs:104 
  at OpenRA.Scripting.ScriptObjectWrapper.get_Item (Eluant.LuaRuntime runtime, Eluant.LuaValue keyValue) [0x0002a] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Scripting/ScriptObjectWrapper.cs:58 
  at OpenRA.Actor.get_Item (Eluant.LuaRuntime runtime, Eluant.LuaValue keyValue) [0x0000e] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Actor.cs:393 
  at Eluant.LuaRuntime.<IndexCallback>m__D (IList`1 toDispose) <0x41590440 + 0x00132> in <filename unknown>:0 
  at Eluant.LuaRuntime.LuaToClrBoundary (IntPtr state, Eluant.LuaToClrBoundaryCallback callback) <0x4158fe00 + 0x00159> in <filename unknown>:0 
  at OpenRA.Scripting.ScriptContext.FatalError (System.String message) [0x00001] in <filename unknown>:0 
  at OpenRA.Mods.Common.Scripting.ScriptTriggers.Damaged (OpenRA.Actor self, OpenRA.Traits.AttackInfo e) [0x00000] in <filename unknown>:0 
  at OpenRA.Mods.Common.Traits.Health.InflictDamage (OpenRA.Actor self, OpenRA.Actor attacker, Int32 damage, IWarhead warhead, Boolean ignoreModifiers) [0x00000] in <filename unknown>:0 
  at OpenRA.Actor.InflictDamage (OpenRA.Actor attacker, Int32 damage, IWarhead warhead) [0x00027] in <filename unknown>:0 
  at OpenRA.Mods.Common.Warheads.DamageWarhead.DoImpact (OpenRA.Actor victim, OpenRA.Actor firedBy, IEnumerable`1 damageModifiers) [0x00000] in <filename unknown>:0 
  at OpenRA.Mods.Common.Warheads.DamageWarhead.DoImpact (Target target, OpenRA.Actor firedBy, IEnumerable`1 damageModifiers) [0x00000] in <filename unknown>:0 
  at OpenRA.GameRules.WeaponInfo+<Impact>c__AnonStorey1.<>m__0 () [0x00027] in <filename unknown>:0 
  at OpenRA.GameRules.WeaponInfo.Impact (Target target, OpenRA.Actor firedBy, IEnumerable`1 damageModifiers) [0x00095] in <filename unknown>:0 
  at OpenRA.Mods.Cnc.Traits.PoisonedByTiberium.Tick (OpenRA.Actor self) [0x00000] in <filename unknown>:0 
  at OpenRA.World.<Tick>m__3 (TraitPair`1 x) [0x0000e] in <filename unknown>:0 
  at OpenRA.WorldUtils.DoTimed[T] (IEnumerable`1 e, System.Action`1 a, System.String text) [0x00022] in <filename unknown>:0 
  at OpenRA.World.Tick () [0x00122] in <filename unknown>:0 
  at OpenRA.Game.InnerLogicTick (OpenRA.Network.OrderManager orderManager) [0x00226] in <filename unknown>:0 
  at OpenRA.Game.LogicTick () [0x00055] in <filename unknown>:0 
  at OpenRA.Game.Loop () [0x000d4] in <filename unknown>:0 
  at OpenRA.Game.Run () [0x00042] in <filename unknown>:0 
  at OpenRA.Program.Run (System.String[] args) [0x00011] in <filename unknown>:0 
  at OpenRA.Program.Main (System.String[] args) [0x00050] in <filename unknown>:0 
```

mods/cnc/maps/nod07c/rules.yaml
```diff
@@ -0,0 +1,139 @@
...
+	LuaScript:
+		Scripts: nod07c.lua
+	MusicPlaylist:
+		StartingMusic: justdoit
+		VictoryMusic: nod_win1
+	MissionData:
+		Briefing: GDI has left themselves open to the deadliest weapon of all, information.\n\nOur spies tell us the GDI base in your area contains an advanced attack helicopter, the Orca.\n\nSteal this weapon, then fly to the nearby village and destroy it.\n\nThe attack will be blamed on GDI.
+		BriefingVideo: nod7b.vqa
+		StartVideo: bcanyon.vqa
+		WinVideo: x.vqa
```
obrakmann: x.vqa doesn't exist. x is the placeholder for 'no vqa' in td .inis, iirc, though. Our map import used to do that correctly...

mods/cnc/maps/nod07c/rules.yaml
```diff
@@ -0,0 +1,139 @@
...
+		InitialSmudges:
+			cr1 40,22 2
+
+Player:
+	PlayerResources:
+		DefaultCash: 0
+
+^TechBuilding:
+	Capturable:
+		Type: ~disabled
```
obrakmann: Why?

mods/cnc/maps/nod07c/rules.yaml
```diff
@@ -0,0 +1,139 @@
...
+	AirstrikePower:
+		Prerequisites: gdi
+
+BOAT:
+	Health:
+		HP: 1500
+	AutoTarget:
+		InitialStance: AttackAnything
+	RejectsOrders:
+		Except: Attack
```
obrakmann: Strictly speaking you don't need to change either `AutoTarget` or `RejectsOrders`. You don't control the unit anyway, so `RejectsOrders` is pointless, and AttackAnything is already the default for AI players (and it should be `InitialStanceAI` in that case, anyway).

But it doesn't hurt any, either.

mods/cnc/maps/nod07c/rules.yaml
Biofreak1987: So that new players don't end up wasting 3 of their 4 engineers at the oil derricks, as they are restricted to capturing the hpad only anyway.

mods/cnc/maps/nod07c/nod07c.lua
```diff
@@ -0,0 +1,186 @@
...
+				actor.Patrol(team.waypoints, true, DateTime.Seconds(25))
+				count = count + 1
+			end
+		end)
+	end
+end
+
+SendDefenders = function(team)
+	defendersActive = true
+	player.MarkCompletedObjective(NodObjective3)
```
obrakmann: It's a bit weird that you still get this granted even when you attack those units only after you're already in the base.

mods/cnc/maps/nod07c/nod07c.lua
```diff
@@ -0,0 +1,186 @@
...
+Utils.Do(Targets, function(actor)
+	Trigger.OnKilled(actor, function()
+		targetsKilled = targetsKilled + 1
+	end)
+end)
+
+Trigger.OnCapture(GDIHpad, function()
+	hpadCaptured = true
+	player.MarkCompletedObjective(NodObjective1)
+	local Orca = Actor.Create("orca", true, { Owner = player, Location = GDIHpad.Location })
```
obrakmann: Can't we make an orca variant that doesn't automatically lift off instead? Not sure if we have a control for that, but if we don't, it should be added.

This way looks rather weird with the orac popping up like that.

mods/cnc/maps/nod07c/nod07c.lua
Biofreak1987: Good point, should I add a check to disable it when you destroy the guard tower with your infantry?

mods/cnc/maps/nod07c/rules.yaml
obrakmann: Well, they only need a fourth ;)

I'd remove the cash trickler then. It's otherwise frustrating to see that source of income and not having a way to get at it.

obrakmann:
One more issue: since there's no production going on, you should cap the GDI's resources every tick so that the harvester keeps harvesting.

And yes, please remove the x.vqa. It's only there due to an oversight.

mods/cnc/maps/nod07c/nod07c.lua
Biofreak1987: I was just thinking about to preplace an orca which is owned by a neutral faction... but then I realized that I can't place it at the center of the hpad... so, no Idea how to do it.

mods/cnc/maps/nod07c/nod07c.lua
obrakmann: Or when the guard tower gets sight of your units. Something like that, yes.

GraionDilach:
Shouldn't GDI and Target be allied with each other?

Biofreak1987:
**Changelog:**
- Added a Trigger.OnDiscovered to the SideEntrance guard tower which executes a fail on the secondanry objective, if not completed
- Allied target with GDI
- Removed x.vqa Video
- Engineers can now capture the oil derricks
- GDI will lose ressources so that the harvester keeps harvesting

I've still got no solution for the Orca, I've thought about preplacing one orca actor on the hpad... but that's not possible since the center is between the cells

hpad

I would be thankful if someone could help me out with that one.

pchote:
>Still got no solution for the Orca. I've thought about preplacing one orca actor onto the hpad... but that's not possible since the center is between the cells

Try placing it on the top-left cell of the helipad in the editor, and then edit map.yaml by hand to change `Location: <cpos>` to `CenterPosition: <cpos.x> * 1024 + 512, <cpos.y> * 1024 + 768`.   That should move it to the proper dock position.

Biofreak1987:
Thx, that seemed to work though. I also had to modify the footprint of the hpad as otherwise the orca would fly away from it.

pchote:
The mission didn't fail when I lost all my engineers.  Is this intentional?

mods/cnc/maps/nod07c/nod07c.lua
```diff
@@ -0,0 +1,204 @@
+NodUnitsVehicles1 = { 'bggy', 'bggy', 'bike', 'bike' }
+NodUnitsVehicles2 = { 'ltnk', 'ltnk' }
+NodUnitsEngineers = { 'e6', 'e6', 'e6', 'e6' }
+NodUnitsRockets = { 'e3', 'e3', 'e3', 'e3' }
+NodUnitsGunners = { 'e1', 'e1', 'e1', 'e1' }
+NodUnitsFlamers = { 'e4', 'e4', 'e4', 'e4' }
+
+Targets = { Actor76, Actor77, Actor78, Actor79, Actor80, Actor81, Actor82, Actor83, Actor84, Actor86, Actor87, Actor88, Actor89, Actor93, Actor94, Actor95, Actor96, Actor97, Actor104, Actor105, Actor106, Actor107, Actor108, Actor157, Actor158, Actor159, Actor160 }
```
pchote: This would be a good usecase for the newish ScriptTags feature.  You could add a tag to all of the targets in map.yaml (eventually the map editor) and fetch these by the API.

Biofreak1987:
Should be fixed now. I've changed it from OnAllKilled to OnAllRemovedFromWorld.

Biofreak1987:
Updated to use the ScriptTags: system.

pchote:
IMO the changes with the engineers and oil derricks break the theme of the mission.  From the wikia link:
> When Seth was satisfied that sufficient damage had been done to the village, the Nod forces fled the vicinity of the base making it appear that GDI forces had actually successfully repelled them. 

Starting the mission with four engineers and having the derricks there as an income stream encourages the player to capture the whole base instead of just the Orca and then sneaking back out.

mods/cnc/maps/nod07c/nod07c.lua
```diff
@@ -0,0 +1,202 @@
...
+		end
+	end)
+end
+
+SendGuards = function(team)
+	for type, amount in pairs(team.units) do
+		local count = 0
+		local actors = enemy.GetActorsByType(type)
+		Utils.Do(actors, function(actor)
+			if actor.IsIdle and count < amount then
```
abcdefg30: Instead of using those local `count` variables here and above you could just use `Utils.Take`.
For example
```
SendGuards = function(team)
	for type, amount in pairs(team.units) do
		local actors = Utils.Take(amount, enemy.GetActorsByType(type))
		Utils.Do(actors, function(actor)
			if actor.IsIdle then
				actor.Patrol(team.waypoints, true, DateTime.Seconds(25))
			end
		end)
	end
end
```
`Utils.Take` will even work when there are less elements in the table than you want to take.

Biofreak1987:
I'm still for disabling building captures on everything except the hpad but I would rather wait for the input of others.

obrakmann:
I can follow the argument that being able to take over the whole base distracts from the main objective, so I'm fine with disabling building captures.

mods/cnc/maps/nod07c/map.yaml
```diff
@@ -0,0 +1,772 @@
+MapFormat: 11
+
+RequiresMod: cnc
+
+Title: ORCA Theft (c)
```
obrakmann: BTW, we're adding the mission numbers now in #11399, so you can already change this to `07c: Orca Theft`

Biofreak1987:
- Updated mission name
- Added Utils.Take
- Disabled building capture for everything except the hpad
- Disabled selling the hpad and building trans
- Disabled hpad power consumption to get rid of the "low power" message
- Cleaned up some of the building health for users with enabled healthbars

abcdefg30:
Ah, good, I already wondered how to incorporate capturing the base. I think not allowing the player to capture other buildings is a good idea, but might lead to confusion (e.g. bugreports by people who wonder why capturing is "broken"). What about renaming the objective to something like `Only capture the hpad`?

mods/cnc/maps/nod07c/nod07c.lua
```diff
@@ -0,0 +1,199 @@
...
+Tick = function()
+	if not Gunboat.IsDead then
+		GunboatCamera.Teleport(Gunboat.Location)
+	end
+
+	if DateTime.GameTime > 2 and player.HasNoRequiredUnits() then
+		enemy.MarkCompletedObjective(GDIObjective)
+	end
+
+	if targetsKilled >= 10 then
```
abcdefg30: I think this is a bit weird, and leads to situations like: (mission is "Accomplished", but most of the town still stands)
openra-2016-06-05t110640z
(although building transports is now disabled, you could still achieve that with only the orca).

mods/cnc/maps/nod07c/nod07c.lua
```diff
@@ -0,0 +1,199 @@
...
+
+GunboatPatrolPath = { GunboatLeft.Location, GunboatRight.Location }
+
+targetsKilled = 0
+
+AutoGuard = function(guards)
+	Utils.Do(guards, function(guard)
+		Trigger.OnDamaged(guard, function(guard, attacker)
+			if not guard.IsDead and not attacker.IsDead then
+				guard.AttackMove(attacker.Location)
```
abcdefg30: This still crashes the game for me (#11383).
I'd suggest only going for `.Hunt()` and leaving the `attacker` part out:
```
if not guard.IsDead then
	guard.Hunt()
end
```

pchote:
> I think not allowing the player to capture other buildings is a good idea, but might lead to confusion

What if we had an explicit objective about being covert, which is failed if you capture or destroy any of the main base structures (base defenses wouldn't count)?  We could also have an objective at the end about escaping the area.

This is basically bike-shedding on my part now, so feel free to ignore these suggestions.

abcdefg30:
I think that is even a better idea.

mods/cnc/maps/nod07c/nod07c.lua
Biofreak1987: .Hunt() is in this situation a bad idea, as it will make the enemy go for the rest of your squad, after the vehicles are destroyed, rendering the distraction useless.

abcdefg30: Oh, interesting. I always just drew my units back and the GDI forces did nothing but waiting in front of the main entrance.

mods/cnc/maps/nod07c/nod07c.lua
```diff
@@ -0,0 +1,199 @@
...
+				if not defendersActive then
+					defendersActive = true
+					player.MarkFailedObjective(NodObjective3)
+				end
+			end)
+		end)
+	end)
+
+	Trigger.OnAllRemovedFromWorld(Engineers, function()
+		if not player.IsObjectiveCompleted(NodObjective1) then
```
abcdefg30: Trailing space after `then`.

mods/cnc/maps/nod07c/nod07c.lua
```diff
@@ -0,0 +1,199 @@
...
+	Reinforcements.Reinforce(player, { 'ltnk'}, { ReinforcementsTopSpawn.Location, ReinforcementsTankRally.Location }, 1)
+	local Engineers = Reinforcements.Reinforce(player, NodUnitsEngineers, { ReinforcementsTopSpawn.Location, ReinforcementsEngineersRally.Location }, 10)
+	Reinforcements.Reinforce(player, NodUnitsRockets, { ReinforcementsBottomSpawn.Location, ReinforcementsRocketsRally.Location }, 10)
+	Trigger.AfterDelay(DateTime.Seconds(3), function()
+		Reinforcements.Reinforce(player, NodUnitsGunners, { ReinforcementsBottomSpawn.Location, ReinforcementsGunnersRally.Location }, 10)
+		Reinforcements.Reinforce(player, NodUnitsFlamers, { ReinforcementsTopSpawn.Location, ReinforcementsFlamersRally.Location }, 10)
+	end)
+	Trigger.AfterDelay(DateTime.Seconds(5), function()
+		local unitsA = Reinforcements.ReinforceWithTransport(player, 'tran.in', NodUnitsVehicles1, { GunboatRight.Location, ReinforcementsHelicopter1Rally.Location }, { GunboatRight.Location })[2]
+
```
abcdefg30: Trailing tab.

mods/cnc/maps/nod07c/nod07c.lua
```diff
@@ -0,0 +1,199 @@
...
+Trigger.OnKilled(GDIOrca, function()
+	if not player.IsObjectiveCompleted(NodObjective2) then
+		player.MarkFailedObjective(NodObjective2)
+	end
+end)
+
+Trigger.OnDamaged(GDIBuilding11, function()
+	SendGuards(MTANK)
+end)
+
```
abcdefg30: Trailing space here as well.

Biofreak1987:
- Fixed those trailing tab / space issues
- Enabled capturing for tech and normal buildings
- added an objective that fails the mission if a GDI main building gets captured or destroyed (excluding the helipad and guard towers)
- replaced attackMove with hunt
- Update the village objective description and raised the kill counter from 10 to 15

obrakmann:
looks fine now, thanks! :+1:

mods/cnc/maps/nod07c/nod07c.lua
```diff
@@ -0,0 +1,207 @@
...
+			if not guard.IsDead then
+				guard.Hunt()
+			end
+		end)
+	end)
+end
+
+InsertNodUnits = function()
+	Media.PlaySpeechNotification(player, "Reinforce")
+	Reinforcements.Reinforce(player, { 'ltnk'}, { ReinforcementsTopSpawn.Location, ReinforcementsTankRally.Location }, 1)
```
abcdefg30: Please add a space after `'ltnk'`.

Mailaender:
Nice. This mission is much more enjoying with the OpenRA balancing (mid-air rocket reloading) and oil derricks to generate funds for more Orcas. :white_check_mark:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/db3f012595a3a3eee9aa03383ec46c4f8b480165...038d2174c52c285788fa91b99d13bcb4e71614f5)

PR #11372, opened 2016-05-29
[RA] Add operation-goldmine.oramap

ghost:
http://resource.openra.net/maps/14344/ (updated)

An author I've never heard of before with a really nice 4 player map. Strategically it has a good number of expansions and ore mines, nice placement of tech buildings and a nice balance between choke-points and open areas.

Aesthetically I think its very tasteful. Props are used sparingly and fit the theater. The attention to detail is great and is up there with the likes of PizzaAtomica's maps.

It seems the bar has been raised with 1v1 maps but the competitive 4 player maps in the official pool is still seriously lacking. This map fits nicely with the current trend of maps and I would hope to see this in tourneys.


obrakmann:
Didn't spot any tiling errors. Still a couple of things that need fixing:

1. There's a zombie by the church at the top of the map, owned by the neutral player. Please replace that with a regular civilian.
2. The last five or so actors in the map (two ore mines and some trees) are owned by the Creeps player. That needs to be changed to Neutral.

Otherwise ok, :+1: 

(too bad that the minimap looks like a mandelbrot set, though)

r34ch:
Replaced the zombie, changed the actors to neutral and fixed a single tiling error.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/26ec1c74cd02dba13ff80a675e268da6f7d25563...571db0910534e8c49ea49996428b59ceca09c745)

PR #11180, opened 2016-04-24
Added burning trees to the Tiberian Dawn mod

Mailaender:
:evergreen_tree: can catch :fire: but only with flame inducing weaponry. They leave a scorched charcoaled tree, but can't be chopped down completely. Artwork for destructible fauna is available, but I fear the balancing shit storm as maps have been designed with trees as indestructible barriers in mind. Closes https://github.com/OpenRA/OpenRA/issues/6403. I can set up RA :deciduous_tree: as well if this gets positive feedback.

PS: Added an additional lint test to catch my own copy & paste mistakes which will crash later on: https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/Buildings/FootprintUtils.cs#L51

PR #11373, opened 2016-05-29
Fixed footprints of several RA and TD base structures

reaperrr:
There were several cases of base buildings in RA and TD with the top row of the footprint only consisting of `_` cells, namely

RA:
- Advanced Power Plant
- Fake Advanced Power Plant
- Tesla Coil
- Gap Generator
- Flak Cannon

TD:
- Weapons Factory
- Temple of Nod
- Hand of Nod
- Obelisk
- Advanced Guard Tower

In the case of defenses, this meant that the `CenterPosition` was located between the empty cell and the occupied cell, rather than centered on the occupied cell, which - while consistent with TD - was not consistent with original RA and made placing, attacking or C4'ing these structures look a bit dodgy.

To make placement, attacking, C4'ing and later on HitShape setup easier and more consistent, I've removed all those empty top rows and adjusted the art offsets instead.

_Note_: Due to the sheer amount of actors that would be affected, I've refrained from doing the same for trees and rocks (fortunately civilian structures are not affected), since they are indestructible in RA/TD and therefore wouldn't really benefit from similar fixes. And, of course, the diff and risk of regressions would go up considerably.

_Note2_: The upgrade rule did work on two maps, but then I ran into a bug in our map upgrading so I had to use a slightly modified version of my previous MapFormat upgrade rule to apply it to all maps.
Additionally, I ran into a zlib error when trying to upgrade the `.oramap` packages, so I could only fix the unzipped maps (though I'm not sure we ship any packaged maps which have these specific actors pre-placed.)

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
@@ -365,7 +365,12 @@ namespace OpenRA.Mods.Common.UtilityCommands
 				mapPlayers.Players[section] = pr;
 		}
 
-		public static void LoadActors(IniFile file, string section, List<string> players, int mapSize, Map map)
+		public virtual CPos ParseActorLocation(string input, int loc)
+		{
+			return new CPos(loc % MapSize, loc / MapSize);
+		}
+
+		public void LoadActors(IniFile file, string section, List<string> players, int mapSize, Map map)
 		{
 			foreach (var s in file.GetSection(section, true))
 			{
@@ -385,8 +390,10 @@ namespace OpenRA.Mods.Common.UtilityCommands
 					var health = Exts.ParseIntegerInvariant(parts[2]) * 100 / 256;
 					var facing = (section == "INFANTRY") ? Exts.ParseIntegerInvariant(parts[6]) : Exts.ParseIntegerInvariant(parts[4]);
 
-					var actor = new ActorReference(parts[1].ToLowerInvariant()) {
-						new LocationInit(new CPos(loc % mapSize, loc / mapSize)),
+					var actorType = parts[1].ToLowerInvariant();
+
+					var actor = new ActorReference(actorType) {
+						new LocationInit(ParseActorLocation(actorType, loc)),
 						new OwnerInit(parts[0]),
 					};
 
@@ -424,9 +431,11 @@ namespace OpenRA.Mods.Common.UtilityCommands
 			foreach (var kv in terrain)
 			{
 				var loc = Exts.ParseIntegerInvariant(kv.Key);
-				var ar = new ActorReference(ParseTreeActor(kv.Value))
+				var treeActor = ParseTreeActor(kv.Value);
+
+				var ar = new ActorReference(treeActor)
 				{
-					new LocationInit(new CPos(loc % MapSize, loc / MapSize)),
+					new LocationInit(ParseActorLocation(treeActor, loc)),
 					new OwnerInit("Neutral")
 				};
 
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -92,7 +92,7 @@ namespace OpenRA.Mods.Common.UtilityCommands
 			catch { }
 		}
 
-		internal static void UpgradeActorRules(int engineVersion, ref List<MiniYamlNode> nodes, MiniYamlNode parent, int depth)
+		internal static void UpgradeActorRules(ModData modData, int engineVersion, ref List<MiniYamlNode> nodes, MiniYamlNode parent, int depth)
 		{
 			var addNodes = new List<MiniYamlNode>();
 
@@ -185,14 +185,14 @@ namespace OpenRA.Mods.Common.UtilityCommands
 					}
 				}
 
-				UpgradeActorRules(engineVersion, ref node.Value.Nodes, node, depth + 1);
+				UpgradeActorRules(modData, engineVersion, ref node.Value.Nodes, node, depth + 1);
 			}
 
 			foreach (var a in addNodes)
 				nodes.Add(a);
 		}
 
-		internal static void UpgradeWeaponRules(int engineVersion, ref List<MiniYamlNode> nodes, MiniYamlNode parent, int depth)
+		internal static void UpgradeWeaponRules(ModData modData, int engineVersion, ref List<MiniYamlNode> nodes, MiniYamlNode parent, int depth)
 		{
 			foreach (var node in nodes)
 			{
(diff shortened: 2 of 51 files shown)


pchote:
This will cause issues for the map importer, right? (IIRC we already have that problem for tib trees and maybe other things).

Can you add something similar to the upgrade rules to those too?

Biofreak1987:
A few lua files need to be updated as well as they include coordinates for rebuilding lost structures... though I might do that with a different pr as I've planned to rework some of the AI code.

reaperrr:
Rebased, added "upgrade rule" to importer (for TD tib trees as well), and updated AI build positions for TD weapons factory (no other cases in TD or RA, as far as I could tell).

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -934,7 +934,144 @@ internal static void UpgradeActors(int engineVersion, ref List<MiniYamlNode> nod
 		{
 			foreach (var node in nodes)
 			{
-				// Add rules here
+				// Fix RA building footprints to not use _ when it's not necessary
+				if (engineVersion < 20160530 && Game.ModData.Manifest.Mod.Id == "ra")
+				{
+					// Tesla Coil
+					if (node.Value.Value == "tsla")
```
obrakmann: I don't quite see the point of these upgrade rules. With hard-coding things like this, it'll only work on the official ra mod, which - when you have the utility that contains these rules - already has been adapted. I think you can safely remove these, unless I'm missing something not obvious.

reaperrr: This is for upgrading maps from the resource center that don't ship with OpenRA, mostly.
@pchote requested them (iirc).

pchote:
> Note: Due to the sheer amount of actors that would be affected, I've refrained from doing the same for trees and rocks (fortunately civilian structures are not affected), since they are indestructible in RA/TD and therefore wouldn't really benefit from similar fixes. And, of course, the diff and risk of regressions would go up considerably.

Do #11180 and #11377 change this?

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
pchote: This is indeed for the resource center and other third party maps.  It sucks to have these checks hardcoded in common code, but we can't fix that until we redo the upgrade rules plumbing.

OpenRA.Mods.Common/UtilityCommands/ImportLegacyMapCommand.cs
```diff
@@ -415,6 +417,10 @@ public static void LoadActors(IniFile file, string section, List<string> players
 
 		public abstract string ParseTreeActor(string input);
 
+		public virtual int ModifyLocationX(string input) { return 0; }
```
pchote: Can we please change this to a `public virtual CPos ParseActorLocation(int loc)` and include the `mapSize` calculation here?

pchote: Alternatively you could pull out a `ParseActor` (like `ParseTreeActor` below) and then move the `LocationInit` stuff into both of these.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -934,7 +934,144 @@ internal static void UpgradeActors(int engineVersion, ref List<MiniYamlNode> nod
 		{
 			foreach (var node in nodes)
 			{
-				// Add rules here
+				// Fix RA building footprints to not use _ when it's not necessary
+				if (engineVersion < 20160530 && Game.ModData.Manifest.Mod.Id == "ra")
```
pchote: Can you please pass `ModData modData` in as the first argument to these methods?  Referencing `Game.ModData` causes a bunch of issues, and i'm slowly trying to eliminate all these references.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -934,7 +934,144 @@ internal static void UpgradeActors(int engineVersion, ref List<MiniYamlNode> nod
...
+				{
+					// Tesla Coil
+					if (node.Value.Value == "tsla")
+					{
+						var locationNode = node.Value.Nodes.First(n => n.Key == "Location");
```
pchote: Lots of code duplication here.  Can you pull out a common method to do this?

pchote: I think the cleanest solution would be to define a `Dictionary<string, CVec> locationOffsets` for each mod and then do the manipulation if it `ContainsKey(node.Value.Value)`.

reaperrr:
> Do #11180 and #11377 change this?

Not really. Burning trees are just a cosmetical feature, and all flame weapons have enough spread that we don't need footprint-shaped HitShapes for trees to make sure they get hit by them.

reaperrr:
Updated.

I didn't entirely do as requested for the upgrade rule, but mostly because I wasn't sure how exactly to implement the suggestions so I opted for a compromise that is hopefully sufficient.

pchote:
50fbe66f05ef3fd3a03cf9e82acf9c537c37fae3 looks good to me.

Mailaender:
This needs to take https://github.com/OpenRA/OpenRA/pull/11368 into account as well now.

Biofreak1987:
gdi07 is missing as well

reaperrr:
Rebased and adapted Nod7c + GDI7.

Mailaender:

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -182,61 +182,94 @@ internal static void UpgradeWeaponRules(int engineVersion, ref List<MiniYamlNode
...
-				// Add rules here
-				UpgradeActors(engineVersion, ref node.Value.Nodes, node, depth + 1);
+				// Fix RA building footprints to not use _ when it's not necessary
+				if (engineVersion < 20160530 && modData.Manifest.Mod.Id == "ra")
+				{
+					var buildings = new List<string>() { "tsla", "gap", "agun", "apwr", "fapw" };
+					if (buildings.Contains(node.Value.Value))
+					{
+						var locationNode = node.Value.Nodes.First(n => n.Key == "Location");
+						var parts = locationNode.Value.Value.Split(',');
```
pchote: You should be able to use `FieldLoader.GetValue<CPos>("Location", locatioinNode.Value.Value)` instead of rolling your own parser.

pchote: It might be a good idea to extract out a method that takes a ref string and a CVec offset and does the manipulation in place.  You can then use it here and below.

pchote: See https://github.com/OpenRA/OpenRA/blob/0ac7b73d37edd18d13efb9e8a9d5ecc9be7e711f/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs#L69

reaperrr:
Rebased and updated.

I can't test the refactored upgrade rule, so review the code carefully. I'll squash the two commits together if the last commit is considered fine.

For the record, the Travis and AppVeyor failures are both not the the fault of this PR.

reaperrr:
GH messed up the commit order again. https://github.com/OpenRA/OpenRA/pull/11373/commits/31e008e4ec540224b852bdf49e98b70469994c80 is actually the 2nd-to-last commit, right before the last one.

reaperrr:
Updated.

mods/ra/rules/structures.yaml
```diff
@@ -408,13 +408,13 @@ TSLA:
 		Name: Tesla Coil
 		Description: Advanced base defense.\nRequires power to operate.\nCan detect cloaked units.\n  Strong vs Vehicles, Infantry\n  Weak vs Aircraft
 	Building:
-		Footprint: _ x
-		Dimensions: 1,2
+		Footprint: x
```
pchote: This is the default footprint, so you can remove it completely (here and the other defenses).

mods/ra/sequences/structures.yaml
```diff
@@ -131,12 +131,16 @@ powr:
 
 apwr:
 	idle:
+		Offset: 0,-10
```
pchote: Shouldn't these be -12 (half a cell)?

pchote: You mention in one of the other commits that you are also adjusting for future footprint positions, so I guess this may be intentional.

mods/cnc/maps/nod07a/nod07a-AI.lua
```diff
@@ -16,7 +16,7 @@ BaseNuke1 = { type = "nuke", pos = CPos.New(16, 56), cost = 500, exists = true }
 BaseNuke2 = { type = "nuke", pos = CPos.New(18, 57), cost = 500, exists = true }
 BaseNuke3 = { type = "nuke", pos = CPos.New(27, 51), cost = 500, exists = true }
 InfantryProduction = { type = "pyle", pos = CPos.New(18, 54), cost = 500, exists = true }
-VehicleProduction = { type = "weap", pos = CPos.New(27, 54), cost = 2000, exists = true }
+VehicleProduction = { type = "weap", pos = CPos.New(27, 55), cost = 2000, exists = true }
```
pchote: Not for this PR, but it would be nice if we could place these in the map editor and query this info at setup time.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -198,61 +198,83 @@ internal static void UpgradeWeaponRules(int engineVersion, ref List<MiniYamlNode
...
+		}
+
+		internal static void UpgradeActors(ModData modData, int engineVersion, ref List<MiniYamlNode> nodes, MiniYamlNode parent, int depth)
 		{
 			foreach (var node in nodes)
 			{
-				// Add rules here
-				UpgradeActors(engineVersion, ref node.Value.Nodes, node, depth + 1);
+				// Fix RA building footprints to not use _ when it's not necessary
+				if (engineVersion < 20160530 && modData.Manifest.Mod.Id == "ra" && depth == 2)
```
pchote: Can you please bump this date up to current?

mods/cnc/rules/structures.yaml
```diff
@@ -821,12 +822,12 @@ ATWR:
 		Prerequisites: anyhq, ~techlevel.medium
 		Queue: Defence.GDI
 	Building:
-		Footprint: _ x
-		Dimensions: 1,2
+		Footprint: x
+		Dimensions: 1,1
```
pchote: Don't forget about the firing offsets!

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
pchote: These need to be `depth == 1` if you want the rule to run :smile:

pchote:
You've missed RA's evacuation map (merged after you started on this?).

Aside from that and the bugs above, the fixed upgrade rules match the changes you've done here.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
reaperrr: > These need to be depth == 1 if you want the rule to run

Hm... are you sure? I thought using `parent` for the actor node (which should be depth 1) and `node` for the `Location` (should be depth 2) would be correct.

mods/ra/sequences/structures.yaml
reaperrr: Yeah, this change is basically similar to the airfield and barracks offset changes, it's supposed to make the visuals match the occupied cells a little better.

mods/cnc/rules/structures.yaml
reaperrr: In case of the ATWR, I changed the 'turret' offset instead.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
pchote: When this runs the actor node is depth 0 and `Location` etc are depth 1.

reaperrr: Alright.

mods/cnc/rules/structures.yaml
pchote: When I tested this the missiles came from tabove the roof instead of from the windows or the launchers below the windows.

reaperrr:
Updated.

- removed redundant default footprints from yaml
- removed redundant default spawn offset from RA airfield
- lowered TD `ATWR` firing offset
- fixed date and depth of upgrade rule
- adapted RA Evacuation

pchote:
Unfortunately this has regressed the selection rectangles on at least the weapons factory, temple, and hand.  I haven't checked the other structures:

screen shot 2016-06-19 at 18 20 33

reaperrr:
Updated, fixed selection boxes of those 3. I did take a look at the other structures as well, but the defenses look fine to me and the selection box of the apwr/fake apwr looked just slightly too small, which I adressed just in case.

reaperrr:
td-select
ra-select

reaperrr:
As requested on IRC, tweaked Gap Gen and Tesla selection boxes:
gap-tsla--fixed

reaperrr:
Updated. I hope this is it...

abcdefg30:

Not merging, as I'm not sure if @Mailaender's ✅ is still valid.

pchote:
Needs a rebase now.

reaperrr:
Rebased. Travis failed at something unrelated.

Mailaender:
I just checked infantry entering buildings as I thought this was the point of the pull request. It looks like @pchote checked this more thoroughly and we should wait for this final. :+1:

obrakmann:
hm, you didn't touch the Iron Curtain, did you? It gets imported with a wrong offset as well:
screenshot from 2016-06-21 21 58 35. It should look like this: 
allies14

OpenRA.Mods.RA/ImportRedAlertLegacyMapCommand.cs
```diff
@@ -126,6 +126,17 @@ public override string ParseTreeActor(string input)
 			return input.ToLowerInvariant();
 		}
 
+		public override CPos ParseActorLocation(string input, int loc)
+		{
+			var newLoc = new CPos(loc % MapSize, loc / MapSize);
+			var vectorDown = new CVec(0, 1);
+
+			if (input == "tsla" || input == "agun" || input == "gap" || input == "apwr")
```
obrakmann: OK, so it looks like this should just contain `iron` as well. No need to add it to the upgrade rules since its rules haven't changed.

obrakmann:
:+1: after that has been addressed. Haven't checked all the maps, but those that I took at look at were fine.

reaperrr:
Updated. Added Iron Curtain to RA map importer location modifier.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a07d045378564146371aea5252c59783e2fb21a6)

PR #11374, opened 2016-05-29
Fix calling OnOwnerChange with wrong player in ProximityCapturable

obrakmann:
The actor's owner only gets changed in an end frame task queued by `Actor.ChangeOwner`. So during this frame end task, `self.Owner` still points to the old owner, whereas we want the new one, represented by `captor.Owner`.

This bug lead to the capture notification being played when _losing_ a building, but not when actually capturing it.


Mailaender:
:+1:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/198c0481cc76958b3ff2ad574b7fac4750cf3e27...9b53ef7be1227bba6d665ff2654b27fdea459c60)

Issue #7115, opened 2014-12-10
Split Burns into separate render and damage traits


reaperrr:
And replace some of the hard-coding with customisation.
##

PR #11377, opened 2016-05-29
Added burning trees to the Red Alert mod

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/11180. Also found a missing :cactus: which was not set up.

### DIFF
--- a/mods/ra/rules/decoration.yaml
+++ b/mods/ra/rules/decoration.yaml
@@ -5,6 +5,8 @@ T01:
 		Dimensions: 2,2
 	EditorTilesetFilter:
 		ExcludeTilesets: DESERT, INTERIOR
+	SpawnActorOnDeath:
+		Actor: T01.Husk
 
 T02:
 	Inherits: ^Tree
@@ -13,6 +15,8 @@ T02:
 		Dimensions: 2,2
 	EditorTilesetFilter:
 		ExcludeTilesets: DESERT, INTERIOR
+	SpawnActorOnDeath:
+		Actor: T02.Husk
 
 T03:
 	Inherits: ^Tree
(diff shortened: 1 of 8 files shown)


r34ch:
I've been looking forward to this feature for _ages_, nice job Mailaender. Some polish suggestions;

- Is it possible that once trees catch fire, they continue to burn down to a husk? Trees are very sparse in maps and only a few weapons set them alight, so this polish won't be seen that frequently. It also doesn't make a ton of sense that they self extinguish before they transition to a burnt husk.

- Can fire animations be transitioned to larger fire animations over time or more fire animations added? I remember the trees in RA1 had a large fire animation covering most of them before they burnt out. Failing that, a larger fire animation would suffice.

- When a tree transforms into a husk, the husk should continue to burn. Currently when a tree transitions to a husk it is very noticeable since there is only the smoke overlay. The larger fire animation would also help mask this transition.

- Once set alight, trees should burn for a noticeably long time. Weapons that do set trees alight tend to be very destructive weapons. Trees and tree husks burning for a long time would really emphasize the destructive power of these weapons and add a lot of atmosphere to the aftermath of a skirmish. For comparison, tank husks in RA burn for ~40s and are much more numerous and frequently seen. Perhaps healthy trees should burn for ~2mins and the husks for about ~3mins.

- The time trees burn for should be randomized. When a nuke hits all the trees self extinguish at the same time.

- Once the fire extinguishes, the smoke overlay should play for much longer too.

Also when targeting the trees, they all seem to be offset to the left. I guess this is what @reaperrr was referring to in the first note in #11373?

Mailaender:
I agree that this is currently a very simple model. It plays the flame overlay and the more incendiary damage is done, the larger the flame gets. Having flame based weapons deal damage over time with appropriate animations is a pull request of it's own. See https://github.com/OpenRA/OpenRA/issues/7115.

obrakmann:
:+1:

reaperrr:
I think this is good enough as a start.
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/d3a7cc98c5fc01bc7b9fc3ff03f9a4d8d31e0eee...0acda573219942e2fcb8ba7fd13d9231da16b8cb)

PR #10759, opened 2016-02-14
Do not render trails under fog by default.

GraionDilach:
Exposed a boolean for rendering a `SpriteEffect` under fog. Exposed a setting in `LeavesTrails` for this boolean. Sandworm trails in D2k remains visible under fog thanks to this.

There's a usecase of `SpriteEffect` in `SpawnActorPower` as well, but I'll assume there that the spawned actor will reveal the fog for the animation if intended to be seen.

PR #11380, opened 2016-05-30
Added UpgradeOnTerrain

Mailaender:
In the example the jumpjet infantry will be able to cross water without getting wet feed and jumpjetting over cliffs. This is inspired by https://github.com/OpenRA/ra2/pull/172, but using a more generic approach with actor upgrades. It works surprisingly well with this simple setup already.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTerrain.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTerrain.cs
@@ -1,0 +1,69 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public class UpgradeOnTerrainInfo : ITraitInfo, Requires<UpgradeManagerInfo>
+	{
+		[UpgradeGrantedReference]
+		public readonly string[] Upgrades = { "terrain" };
+
+		[Desc("Terrain names to trigger the upgrade.")]
+		public readonly string[] TerrainTypes = { };
+
+		public object Create(ActorInitializer init) { return new UpgradeOnTerrain(init, this); }
+	}
+
+	public class UpgradeOnTerrain : ITick
+	{
+		readonly Actor self;
+		readonly UpgradeOnTerrainInfo info;
+		readonly UpgradeManager manager;
+
+		bool granted;
+		string previousTerrain;
+
+		public UpgradeOnTerrain(ActorInitializer init, UpgradeOnTerrainInfo info)
+		{
+			self = init.Self;
+			this.info = info;
+			manager = self.Trait<UpgradeManager>();
+		}
+
+		public void Tick(Actor self)
+		{
+			var currentTerrain = self.World.Map.GetTerrainInfo(self.Location).Type;
+			var wantsGranted = info.TerrainTypes.Contains(currentTerrain);
+			if (currentTerrain != previousTerrain)
+			{
+				if (wantsGranted && !granted)
+				{
+					foreach (var up in info.Upgrades)
+						manager.GrantUpgrade(self, up, this);
+
+					granted = true;
+				}
+				else if (!wantsGranted && granted)
+				{
+					foreach (var up in info.Upgrades)
+						manager.RevokeUpgrade(self, up, this);
+
+					granted = false;
+				}
+			}
+
+			previousTerrain = currentTerrain;
+		}
+	}
+}
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -498,6 +498,7 @@
     <Compile Include="Traits\Upgrades\UpgradableTrait.cs" />
     <Compile Include="Traits\Upgrades\UpgradeActorsNear.cs" />
     <Compile Include="Traits\Upgrades\UpgradeOnDamage.cs" />
+    <Compile Include="Traits\Upgrades\UpgradeOnTerrain.cs" />
     <Compile Include="Traits\Upgrades\UpgradeManager.cs" />
     <Compile Include="Traits\Valued.cs" />
     <Compile Include="Traits\Voiced.cs" />


OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTerrain.cs
```diff
@@ -0,0 +1,67 @@
...
+						manager.RevokeUpgrade(self, up, this);
+
+					granted = false;
+				}
+
+				previousTerrain = currentTerrain;
+			}
+		}
+	}
+}
```
GraionDilach: Isn't there should be a line-ending here?

GraionDilach:
This is actually a wonderful idea, to be honest and quite certain it's closer to what WW had in mind with this unit than what they coded in ultimately.

mods/ts/rules/gdi-infantry.yaml
```diff
@@ -74,6 +74,9 @@ JUMPJET:
 		VoiceSet: JumpJet
 	Mobile:
 		Speed: 56
+		TerrainSpeeds:
+			Water: 100
+			Cliff: 100
```
GraionDilach: I'd prolly give this 75 tbh. 100 seems a bit fast.

Phrohdoh:
A concern of mine is targeting checking position and not upgrades.
A jumpjet could appear to be flying over a stream but could be targeted by anti-ground weapons.

On the other hand, that commit is just an example and isn't meant to be merged I assume.

DoGyAUT:
This works perfect, gonna use it for Crystallized Doom for sure :+1:

reaperrr:
> On the other hand, that commit is just an example and isn't meant to be merged I assume.

It doesn't look like a test commit to me.

Fixing targetability is easy, but as long as the JJ infantry flies almost at ground level, it should be treated like a hover unit and be targetable by anti-ground weapons all the time.

GraionDilach:
It works as a hover unit in the cases the PR's effect kicks in and not as a jumpjet. The sole time it technically flies higher than ground is when it's inbetween passing cliffs but it's practically still hovering level.

Mailaender:
I thought about having jumpjet only fly on water and then make `Aircraft` upgradeable for cliffs, but it turned out that it already works reasonably well with just `Hovers` and a `WithInfantryBody` upgrade. As they don't really fly, a `Targetable` upgrade might be weird. I am not sure if the `WithInfantryBody@Ground` is disabled properly on upgrade though. `WithInfantryBody@Airborne` seems to be drawn on top. Not sure what I am missing there or if this is a bug in the actor upgrade system.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTerrain.cs
Mailaender: :question:

mods/ts/rules/gdi-infantry.yaml
```diff
@@ -87,8 +90,23 @@ JUMPJET:
 	-Crushable:
 	AttackFrontal:
 		Voice: Attack
-	WithInfantryBody:
+	WithInfantryBody@Ground:
```
reaperrr: `WithInfantryBody` is inherited from the default, so either remove the `@Ground` so the inherited one is overridden, or add `-WithInfantryBody` above to disable it.

pchote:
The trait looks reasonable, but I'm not a big fan of the TS jumpjet behaviour.  If we're going to do any quick placeholders (which I dislike on principle), then it would be better to have them properly flying all the time, like the RA2 rocketeers.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTerrain.cs
GraionDilach: https://github.com/OpenRA/OpenRA/pull/10759#discussion_r52845580

Mailaender: I added a newline.

mods/ts/rules/gdi-infantry.yaml
Mailaender: Thanks. That fixed it.

pchote:
@DoGyAUT: is that a formal review :+1:, or just an "I like this"?

DoGyAUT:
@Mailaender reaplace WithInfantryBody@Ground with WithInfantryBody or add -WithInfantryBody then the sequence should be working fine.

Otherwise this works pretty well in CD so far - I used them so far as RA2 type rocketeers units.

Mailaender:
@DoGyAUT Thanks. Already fixed it as @reaperrr also pointed it out to me.

reaperrr:
I've tested this now, and the trait on its own definitely gets my :+1: 

But to be honest, I'm not that happy with the jumpjet change, either. It still looks (and is) rather hacky. I'm definitely against using the RA2 rocketeer approach, I just tried that and it looks and works even worse due to our current helicopter climb code (they ascend/descend straight vertically on every height level change).
But all in all, while I wouldn't block the PR over it, I'd prefer to only merged the trait now (it's useful enough on its own) and fix jumpjets properly in one go once our mobile/locomotor code is up for it.

Mailaender:
Removed https://github.com/OpenRA/OpenRA/pull/11380/commits/469faa4128e55f4adadc702b12ec8fd245c0b15f under protest.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTerrain.cs
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		bool granted;
+		string previousTerrain;
+		public void Tick(Actor self)
+		{
+			var currentTerrain = self.World.Map.GetTerrainInfo(self.Location).Type;
+			if (currentTerrain != previousTerrain)
+			{
+				if (info.TerrainTypes.Contains(currentTerrain))
```
pchote: You need to check against `granted` here too, otherwise you will end up granting a second time when moving from `info.TerrainTypes.Contains(previousTerrain)` to `info.TerrainTypes.Contains(currentTerrain)` (and then end up with unbalanced levels when you move to a non-granting terrain type).

Mailaender: Added.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTerrain.cs
```diff
@@ -0,0 +1,67 @@
...
+
+		public UpgradeOnTerrain(ActorInitializer init, UpgradeOnTerrainInfo info)
+		{
+			self = init.Self;
+			this.info = info;
+			manager = self.Trait<UpgradeManager>();
+		}
+
+		bool granted;
+		string previousTerrain;
```
abcdefg30: Don't we usually declare variables above the `ctor` (but below the `readonly` variables)?

Mailaender: This variable isn't declared in the constructor, but only used internally in the following Tick method.

Phrohdoh: Yes instance members are typically declared above the ctor.

reaperrr: I wasn't aware of this either, but I guess that should be changed then.

Mailaender: I doubt it is the default code style for this game, but I also don't want to argue about it.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTerrain.cs
```diff
@@ -0,0 +1,68 @@
...
+			if (currentTerrain != previousTerrain)
+			{
+				if (info.TerrainTypes.Contains(currentTerrain) && !granted)
+				{
+					foreach (var up in info.Upgrades)
+						manager.GrantUpgrade(self, up, this);
+
+					granted = true;
+				}
+				else if (granted)
```
obrakmann: The conditional is written here will cause the upgrade to be lost when you move from one supported terrain type to another. `granted` would be true in that case, so you wouldn't enter the `if` branch but the `else` branch, where only `granted` is checked. You need to check the terrain type again here.

pchote: It might be nicer to have an inner conditional instead of repeating the same Contains.

reaperrr: @Mailaender While you're at it, you might want to move the `granted`/``!granted`` checks before the terrain checks, as the former are presumably a little cheaper.

obrakmann:
Looks otherwise fine, :+1:

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTerrain.cs
pchote: The cleanest fix would be to pull out a `var wantsGranted = info.TerrainTypes.Contains(currentTerrain)`, and then check `if (wantsGranted && !granted)` and `else if (!wantsGranted && granted)`.

Mailaender: Updated.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/edd7a75895b76bee337b324d58cefd743d1a45bb)

PR #11381, opened 2016-05-31
Cleaned up upgrade rules

reaperrr:
Considering there's been about 5 months of upgrade rules included in the latest release, I thought we might as well make another clear cut here. From my own experience, it's generally safer to upgrade 3rd-party mods at least every release anyway.


pchote:
The last rule cleanup was forced by some engine changes that would have otherwise required a lot of upgrade rule rewrites.  I'm fine with this (so :+1:), but it would be much better if somebody could take on the task of fixing upgrade rules properly (something i've talked about a lot on IRC) instead.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/f3f62af420375bade9458edb6d29c65b2b155d43)

PR #11382, opened 2016-05-31
Moved projectiles to their own namespace and streamlined property names

reaperrr:
Even though they are (currently) technically effects, I think it's a bit un-intuitive to have them located along more regular effects.

Additionally, there were several inconsistencies between the sprite projectiles, which made switching between those projectiles a larger hassle than necessary.

I also found some descriptions to be a little confusing.

Finally, `LaserZap` and `TeslaZap` classes were not `public` yet.

The upgrade rules were only put together into the last commit to avoid rebase conflicts as much as possible, all yaml changes were performed by them, so they've been tested and seem to work fine.


OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -865,6 +865,30 @@ internal static void UpgradeWeaponRules(int engineVersion, ref List<MiniYamlNode
 					}
 				}
 
+				// Streamline some projectile property names and functionality
+				if (engineVersion < 20160601)
+				{
+					if (node.Key == "Sequence")
+					{
```
pchote: Minor style nit: can we drop the unnecessary braces?

pchote:
Nice cleanup, thanks :)  :+1: after that minor fix.

reaperrr:
Updated.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/367ea93d0937224ab224cee39d38fb450bd29011)

I didn't mention the description updates since they aren't exposed anywhere yet.

PR #11386, opened 2016-06-01
Reduce gliding effect on smooth turns while moving

reaperrr:
The worst offender is the turn around a single cell obstacle, because the turning actor will just shift straight from one cell corner to the other. Therefore, until we have a better turning logic, we disable smooth turning for these specific turns.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Move/Move.cs
+++ b/OpenRA.Mods.Common/Activities/Move/Move.cs
@@ -387,7 +387,11 @@ namespace OpenRA.Mods.Common.Activities
 				var nextCell = parent.PopPath(self);
 				if (nextCell != null)
 				{
-					if (IsTurn(mobile, nextCell.Value.First))
+					// HACK: small U-turns currently look crap, because the actor will just shift on the edge between cells.
+					// Therefore we disable smooth turning for them until we have a better turning logic.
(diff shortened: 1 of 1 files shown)


pchote:
I don't like the idea of just turning this off.

A not-much-harder workaround that would look much better would be to fake the turn using spline interpolation:

https://en.wikipedia.org/wiki/Centripetal_Catmull%E2%80%93Rom_spline

https://github.com/pchote/star-simulator/blob/master/cloudgen.c#L85

`p1` and `p2` are the start and end points of the move, and `p0` and `p3` can be found by adding a `WVec` to `p1` and `p2` in the direction of the movement into and out of the move.  The only tricky bit is to work out how long those vectors should be: For now you could assume a constant value (maybe half or a full cell), but for a more realistic behaviour we want to scale the length based on the inverse of the ROT.

reaperrr:
I somewhat understand the general logic, but I have no idea how that would have to look like in the `MovePart` code.

pchote:
The `WPos.Lerp` inside `UpdateCenterLocation` would be replaced with a new function that interpolates along the spline defined by the four points (`From - <initial vector>`, `From`, `To`, `To + <final vector>`) using the equations from the links above.

reaperrr:
Closing this for now, mostly since I don't know when I'll get back to this. When/If I manage to implement that suggestion successfully, I'll reopen it.

PR #11390, opened 2016-06-02
Add Dune 2000 map Lynch Pass (2 player)

OmegaBolt:
lynchpass
A 2 player map that forces vehicles through the middle S chokepoint, in the centre of which sand worms are able to spawn. Infantry, however, can access the cliff tops in the centre, allowing them unique freedom to move through the map and strike the enemy base and spice expansion. Expansion is quite necessary too, as as the game progresses players will need to claim further fields and there are two large rock areas for expansion directly opposite each starting position.


MustaphaTR:
Looks like a nice map :+1:

MustaphaTR:
Looks like a nice map :+1:

obrakmann:
Worms don't attack units because the `Creeps` player doesn't have any `Enemies:` defined. Also the AI players have a pretty hard time finding paths on that map. Not sure what can be done about that one, though.

:+1: when the worms have been fixed.

OmegaBolt:
Might be because I upgraded this from a version in December. What's the best way to fix that?

obrakmann:
Add `Enemies: Multi0, Multi1` to the `Creeps` player in map.yaml.

OmegaBolt:
Yeah, but how do I get that back into .oramap so it doesnt have to be unpacked?

obrakmann:
err... any decent archive program should allow you to open the map.yaml in an editor and ask you whether the archive should be updated when you save it. The .oramap is just a .zip file, btw.

pchote:
You can also tell the map editor to save an an unpacked/packed map.

OmegaBolt:
I didn't know it was just a zip, thanks. Will the editor repack an unpacked map though if you change the setting?

Updated it, should be fixed.

pchote:
> Will the editor repack an unpacked map though if you change the setting?

It will, yes.

pchote:
Looks fine to me :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/198c0481cc76958b3ff2ad574b7fac4750cf3e27

PR #11391, opened 2016-06-02
Added the FLAMEGUY death animation

Mailaender:
A revival of https://github.com/OpenRA/OpenRA/pull/9350.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckDeathTypes.cs
+++ b/OpenRA.Mods.Common/Lint/CheckDeathTypes.cs
@@ -28,7 +28,9 @@ namespace OpenRA.Mods.Common.Lint
 				if (!animations.Any())
 					continue;
 
-				var deathTypes = animations.SelectMany(x => x.DeathTypes.Select(y => y.Key)).ToList();
+				var deathAnimationDeathtypes = animations.SelectMany(x => x.DeathTypes.Select(y => y.Key)).ToList();
+				var spawnActorDeathtypes = actorInfo.Value.TraitInfos<SpawnActorOnDeathInfo>().Where(s => !string.IsNullOrEmpty(s.DeathType)).Select(a => a.DeathType);
+				var deathTypes = deathAnimationDeathtypes.Concat(spawnActorDeathtypes).Distinct();
 				if (!deathTypes.Any())
 					continue;
 
@@ -51,7 +53,7 @@ namespace OpenRA.Mods.Common.Lint
 							continue;
 
 						if (!warhead.DamageTypes.Overlaps(deathTypes))
-							emitError("Actor type `{0}` does not define a death animation for weapon `{1}`!"
+							emitError("Actor type {0} doesn't define a death animation or spawn an actor on death for weapon {1}!"
 								.F(actorInfo.Key, weaponInfo.Key));
 					}
 				}
--- a/mods/ts/sequences/infantry.yaml
+++ b/mods/ts/sequences/infantry.yaml
@@ -960,10 +960,6 @@ cyborg:
 		Length: 6
 		Facings: 8
 		ShadowStart: 582
-	die5: # TODO: unused running frame 322 and following
-		Start: 292
-		Length: 14
-		ShadowStart: 662
 	die6: electro
 		Length: *
 	emp-overlay: emp_fx01
--- a/mods/ts/sequences/infantry.yaml
+++ b/mods/ts/sequences/infantry.yaml
@@ -60,10 +60,6 @@ e1.gdi:
 		Length: 2
 		Facings: 8
 		ShadowStart: 552
-	die5: flameguy # TODO: walking animation unused
-		Start: 42
-		Length: 104
-		ShadowStart: 190
 	die6: electro
 		Length: *
 	icon: sidebar-gdi|e1icon
@@ -130,10 +126,6 @@ e1.nod:
 		Length: 2
 		Facings: 8
 		ShadowStart: 552
-	die5: flameguy # TODO: walking animation unused
-		Start: 42
-		Length: 104
-		ShadowStart: 190
 	die6: electro
 		Length: *
 	icon: sidebar-nod|e1icon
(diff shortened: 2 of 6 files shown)


OpenRA.Mods.Common/Lint/CheckDeathTypes.cs
```diff
@@ -51,7 +55,7 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Ruleset ru
 							continue;
 
 						if (!warhead.DamageTypes.Overlaps(deathTypes))
-							emitError("Actor type `{0}` does not define a death animation for weapon `{1}`!"
+							emitError("Actor type `{0}` does neither define a death animation nor spawns an actor on death for weapon `{1}`!"
```
pchote: Weird english here.  "Actor type `{0}` doesn't define a death animation or spawn an actor on death for weapon `{1}`!" would be better.

pchote:
The running speed feels comically fast to me.  Can you slow it down?

Looks good aside from those nits, so :+1: after they are fixed.

Mailaender:
Slowed it down.

OpenRA.Mods.Common/Lint/CheckDeathTypes.cs
```diff
@@ -36,6 +36,10 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Ruleset ru
 				if (!targetable.Any())
 					continue;
 
+				var sa = actorInfo.Value.TraitInfos<SpawnActorOnDeathInfo>();
+				if (sa.Any())
+					deathTypes.AddRange(sa.Where(t => t.DeathType != null).Select(dt => dt.DeathType).ToList());
```
pchote: I believe the `ToList` here is redundant.

pchote: This whole block can simplify down to something like:
`deathTypes.AddRange(actorInfo.Value.TraitInfos<SpawnActorOnDeathInfo>().SelectMany(s => s.DeathType).Where(dt => dt != null).Distinct()))`

pchote: This code won't run at all if the actor doesn't have any death type animations.  You will need to adjust the checks above to account for `SpawnActorOnDeath`.

pchote:
Removing my original :+1: because I didn't properly test the lint rules.

Marnador:
From my test the flameguy shadow sequence seems a bit off. 
My suggestion is to put  **ShadowStart: 148** under **Defaults:** and remove **ShadowStart: 48** under **run:**. That way it doesn't flickers on and off when the actor is standing still for a few milliseconds, and it will render the right shadows when the actor is running.
But otherwise works pretty well! was actually looking in to it not too long ago :)

Mailaender:
Rewrote the lint check and fixed the shadow sequence.

mods/ts/sequences/infantry.yaml
```diff
@@ -1706,9 +1622,21 @@ civ3:
...
-	die5: flameguy # TODO: walking animation unused
-		Start: 42
-		Length: 104
-		ShadowStart: 190
 	die6: electro
 		Length: *
+
+flameguy:
+	Defaults:
+		Facings: 7
```
reaperrr: This looks highly suspicious. Does the art really cover only 7 facings?

Mailaender: No, it is wrong: `8` like all the others.

reaperrr: Nevermind, it actually does.

reaperrr: * sigh* scratch that, I somehow skipped the frames of the 8th facing.

pchote:
One last glitch: I don't think this should apply to cyborgs.  It looks rather strange for a broken robot to suddenly grow legs and run around on fire.

Mailaender:
This should now also take cyborgs into account.

pchote:
Cyborgs still look quite glitchy, instantly growing new legs before falling over again on fire.

reaperrr:
In my opinion cyborgs should just explode on death (regardless of death type) like they did in the original, the WithInjury logic makes anything else look glitchy without custom art.

Mailaender:
Explosible cyborgs seems unrelated to this pull request (it doesn't regress it) and I want to keep the proper setup in the commit history.

pchote:
The tag is there for reverting the specific flame death additions for cyborgs, so that they remain as they currently are on bleed.

Mailaender:
They already remain as they currently are on bleed.

pchote:
ab415d874a835bd97289ecd91f14c2c769ba306d adds a `FLAMECYBORG` actor and sequences.

Mailaender:
It isn't used at the moment, but that doesn't mean the sequence setup should be thrown away.

pchote:
Do you have a usecase for it?

reaperrr:
> # TODO: unused running frame 322 and following

That comment was simply wrong (at least in the sense that it implied they should be used), `die5` needs to be a duplicate of `die4` for cyborgs. Actually, the whole Cyborg death setup is a bit bogus currently. Fixing that is outside the scope of this PR of course, but that's just another reason to only concentrate on avoiding changes/regressions over bleed.

> The obvious use case for FLAMECYBORG is using it for effects in maps and missions or triggering the cyborg crippling only for certain weapons or at random.

That goes against our "make TS as close to the original first before we make custom changes" policy. And mappers can still do that via custom map rules, if they want to.

Mailaender:
Added the explosion back in.

reaperrr:
@pchote's (and mine) fixup request for not adding unused sequences etc. still stands.

Mailaender:
Removed the `FLAMECYBORG`. At least the past commit is referenced so not all is wasted.

pchote:
The cyborg explosion is now drawn in the wrong palette.  I suspect you are missing a palette override on the `WithDeathAnimation`.

Mailaender:
Updated again. I would have wished if all this unrelated crap could have stayed out of this pull request.

reaperrr:
Looks good to me now :+1:

pchote:
:+1:

abcdefg30:
Thanks for reviving this, @Mailaender!

PR #11393, opened 2016-06-02
Fixed air unit resupply

### LINKED ISSUE #5966: Implement queued reloading for aircraft [$69]
Since aircraft need to be reloaded, it is very time consuming and boring to reload 20 helicopters at one helipad for example. But if there was a feature that reloaded each aircraft one after another when a group of them are ordered to reload, it would make this process much easier. The same logic could also be applied to repairs at the service depot.

<bountysource-plugin>
---
There is a **[$69 open bounty](https://www.bountysource.com/issues/3148314-implement-queued-reloading-for-aircraft?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bateramos:
Correction to the resupply queue for airplanes and helicopters.

https://www.youtube.com/watch?v=VSiwaujBaPA

Fixes  #5966.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
+++ b/OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
@@ -41,7 +41,7 @@ namespace OpenRA.Mods.Common.Activities
 			// Move to the next activity only if all ammo pools are depleted and none reload automatically
 			// TODO: This should check whether there is ammo left that is actually suitable for the target
 			if (ammoPools.All(x => !x.Info.SelfReloads && !x.HasAmmo()))
-				return NextActivity;
+				return ActivityUtils.SequenceActivities(new ReturnToBase(self), NextActivity);
 
 			if (attackPlane != null)
 				attackPlane.DoAttack(self, target);
--- a/OpenRA.Mods.Common/Activities/Air/FlyCircleTimed.cs
+++ b/OpenRA.Mods.Common/Activities/Air/FlyCircleTimed.cs
@@ -1,0 +1,36 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using OpenRA.Activities;
+using OpenRA.Mods.Common.Traits;
+
+namespace OpenRA.Mods.Common.Activities
+{
+	public class FlyCircleTimed : FlyCircle
+	{
+		int remainingTicks;
+
+		public FlyCircleTimed(int ticks, Actor self) : base(self)
+		{
+			remainingTicks = ticks;
+		}
+
+		public override Activity Tick(Actor self)
+		{
+			if (IsCanceled || remainingTicks-- == 0)
+				return NextActivity;
+
+			base.Tick(self);
+
+			return this;
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
+++ b/OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
@@ -44,13 +44,28 @@ namespace OpenRA.Mods.Common.Activities
 			{
 				var rearmBuildings = heli.Info.RearmBuildings;
 				var nearestHpad = self.World.ActorsHavingTrait<Reservable>()
-									.Where(a => a.Owner == self.Owner && rearmBuildings.Contains(a.Info.Name))
-									.ClosestTo(self);
+					.Where(a => a.Owner == self.Owner && rearmBuildings.Contains(a.Info.Name))
+					.ClosestTo(self);
 
 				if (nearestHpad == null)
 					return ActivityUtils.SequenceActivities(new Turn(self, initialFacing), new HeliLand(self, true), NextActivity);
 				else
-					return ActivityUtils.SequenceActivities(new HeliFly(self, Target.FromActor(nearestHpad)));
+				{
+					var distanceFromHelipad = (nearestHpad.CenterPosition - self.CenterPosition).HorizontalLength;
+					var distanceLength = heli.Info.WaitDistanceFromResupplyBase.Length;
+
+					// If no pad is available, move near one and wait
+					if (distanceFromHelipad > distanceLength)
+					{
+						var randomPosition = WVec.FromPDF(self.World.SharedRandom, 2) * distanceLength / 1024;
+
+						var target = Target.FromPos(nearestHpad.CenterPosition + randomPosition);
+
+						return ActivityUtils.SequenceActivities(new HeliFly(self, target, WDist.Zero, heli.Info.WaitDistanceFromResupplyBase), this);
+					}
+
+					return this;
+				}
 			}
 
 			heli.MakeReservation(dest);
@@ -62,7 +77,8 @@ namespace OpenRA.Mods.Common.Activities
 				new HeliFly(self, Target.FromPos(dest.CenterPosition + offset)),
 				new Turn(self, initialFacing),
 				new HeliLand(self, false),
-				new ResupplyAircraft(self));
+				new ResupplyAircraft(self),
+				NextActivity);
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
+++ b/OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
@@ -32,8 +32,17 @@ namespace OpenRA.Mods.Common.Activities
 			if (host == null)
 				return NextActivity;
 
+			if (aircraft.IsPlane)
+				return ActivityUtils.SequenceActivities(
+					aircraft.GetResupplyActivities(host)
+					.Append(new CallFunc(() => aircraft.UnReserve()))
+					.Append(new WaitFor(() => NextActivity != null || Reservable.IsReserved(host)))
+					.Append(new TakeOff(self))
+					.Append(NextActivity).ToArray());
+
+			// If is helicopter move away as soon as the resupply ends
 			return ActivityUtils.SequenceActivities(
-				aircraft.GetResupplyActivities(host).Append(NextActivity).ToArray());
+				aircraft.GetResupplyActivities(host).Append(new TakeOff(self)).Append(NextActivity).ToArray());
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
+++ b/OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
@@ -106,11 +106,13 @@ namespace OpenRA.Mods.Common.Activities
 			{
 				var nearestAfld = ChooseAirfield(self, false);
 
-				self.CancelActivity();
 				if (nearestAfld != null)
-					return ActivityUtils.SequenceActivities(new Fly(self, Target.FromActor(nearestAfld)), new FlyCircle(self));
+					return ActivityUtils.SequenceActivities(
+						new Fly(self, Target.FromActor(nearestAfld), WDist.Zero, plane.Info.WaitDistanceFromResupplyBase),
+						new FlyCircleTimed(plane.Info.NumberOfTicksToVerifyAvailableAirport, self),
+						this);
 				else
-					return new FlyCircle(self);
+					return NextActivity;
 			}
 
 			List<Activity> landingProcedures = new List<Activity>();
@@ -123,6 +125,7 @@ namespace OpenRA.Mods.Common.Activities
 			// Fix a problem when the airplane is send to resupply near the airport
 			landingProcedures.Add(new Fly(self, Target.FromPos(w3), WDist.Zero, new WDist(turnRadius / 2)));
 			landingProcedures.Add(new Land(self, Target.FromActor(dest)));
+			landingProcedures.Add(new ResupplyAircraft(self));
 			landingProcedures.Add(NextActivity);
 
 			return ActivityUtils.SequenceActivities(landingProcedures.ToArray());
--- a/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
+++ b/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
@@ -83,6 +83,12 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Sound to play when the actor is landing.")]
 		public readonly string LandingSound = null;
 
+		[Desc("The distance of the resupply base that the aircraft will wait for its turn.")]
+		public readonly WDist WaitDistanceFromResupplyBase = new WDist(3072);
+
+		[Desc("The number of ticks that a airplane will wait to make a new search for an available airport.")]
+		public readonly int NumberOfTicksToVerifyAvailableAirport = 150;
+
 		public IReadOnlyDictionary<CPos, SubCell> OccupiedCells(ActorInfo info, CPos location, SubCell subCell = SubCell.Any) { return new ReadOnlyDictionary<CPos, SubCell>(); }
 		bool IOccupySpaceInfo.SharesCell { get { return false; } }
 	}
@@ -98,6 +104,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		UpgradeManager um;
 		IDisposable reservation;
+		Actor reservedActor;
 		IEnumerable<int> speedModifiers;
 
 		[Sync] public int Facing { get; set; }
@@ -181,6 +188,13 @@ namespace OpenRA.Mods.Common.Traits
 			if (!Info.Repulsable)
 				return WVec.Zero;
 
+			if (reservation != null)
+			{
+				var distanceFromReservationActor = (reservedActor.CenterPosition - self.CenterPosition).HorizontalLength;
+				if (distanceFromReservationActor < Info.WaitDistanceFromResupplyBase.Length)
+					return WVec.Zero;
+			}
+
 			// Repulsion only applies when we're flying!
 			var altitude = self.World.Map.DistanceAboveTerrain(CenterPosition).Length;
 			if (altitude != Info.CruiseAltitude.Length)
@@ -261,7 +275,10 @@ namespace OpenRA.Mods.Common.Traits
 			UnReserve();
 			var reservable = target.TraitOrDefault<Reservable>();
 			if (reservable != null)
+			{
 				reservation = reservable.Reserve(target, self, this);
+				reservedActor = target;
+			}
 		}
 
 		public void UnReserve()
@@ -271,6 +288,8 @@ namespace OpenRA.Mods.Common.Traits
 
 			reservation.Dispose();
 			reservation = null;
+			reservedActor = null;
+
 			if (self.World.Map.DistanceAboveTerrain(CenterPosition).Length <= Info.LandAltitude.Length)
 				self.QueueActivity(new TakeOff(self));
 		}
--- a/AUTHORS
+++ b/AUTHORS
@@ -105,6 +105,7 @@ Also thanks to:
     * Paolo Chiodi (paolochiodi)
     * Paul Dovydaitis (pdovy)
     * Pavlos Touboulidis (pav)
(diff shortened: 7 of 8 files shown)


OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
```diff
@@ -36,12 +36,12 @@ public FlyAttack(Actor self, Target target)
 		public override Activity Tick(Actor self)
 		{
 			if (!target.IsValidFor(self))
-				return NextActivity;
+				return new ReturnToBase(self);
```
chrisforbes: Why?

OpenRA.Mods.Common/Activities/Execute.cs
```diff
@@ -0,0 +1,34 @@
...
+#endregion
+
+using System;
+using OpenRA.Activities;
+
+namespace OpenRA.Mods.Common.Activities
+{
+	public class Execute : Activity
+	{
+		Action f;
```
chrisforbes: We used to have an activity like this, called 'CallFunction'. It was nuked from orbit.

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -20,6 +20,8 @@ namespace OpenRA.Mods.Common.Activities
 {
 	public class ReturnToBase : Activity
 	{
+		static int ticksForNextValidation = 50;
```
chrisforbes: This should come from traitinfo data.

OpenRA.Mods.Common/Activities/Execute.cs
bateramos: In this case, I should create a Activity that wait for the airport to receive a reservation call?

OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
bateramos: This is to make the aircraft try to return to the resupply base even if its already reserved after destroying a target. It should be better to preserve the NextActivity? Maybe but it on a Activity Sequence that the ReturnToBase is the last?
I know for sure that when there is no more commands, NextActivity is null. Maybe only return to base is this case?

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
bateramos: ok!

OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
pchote: These will break queued activity chains (where players can queue orders using the shift key).

OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
```diff
@@ -44,16 +44,33 @@ public override Activity Tick(Actor self)
...
 				else
-					return ActivityUtils.SequenceActivities(new HeliFly(self, Target.FromActor(nearestHpad)));
+				{
+					var distanceFromHelipad = (nearestHpad.CenterPosition - self.CenterPosition).HorizontalLength;
+					var distanceLength = heli.Info.WaitDistanceFromResupplyBase.Length;
+
+					// If no pad is available, move near one and wait
+					if (distanceFromHelipad > distanceLength)
+					{
+						var randomXPosition = self.World.SharedRandom.Next(-distanceLength, distanceLength);
```
pchote: You can use `WVec.FromPDF` to get a random 2D offset (then multiply by `distanceLength / 1024`) instead of DIYing your own.

OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
```diff
@@ -44,16 +44,33 @@ public override Activity Tick(Actor self)
...
+
+						return ActivityUtils.SequenceActivities(new HeliFly(self, target, WDist.Zero, heli.Info.WaitDistanceFromResupplyBase), this);
+					}
+
+					return this;
+				}
 			}
 
 			heli.MakeReservation(dest);
+			heli.DisableRepulsing();
```
pchote: This looks dangerous.  What do we gain by manually disabling and enabling repulsation?

OpenRA.Mods.Common/Activities/Execute.cs
pchote: This would be best handled by the long-discussed and long-procrastinated docking rewrite.  CC @reaperrr and @penev92.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -176,6 +189,18 @@ public void Repulse()
 			SetPosition(self, CenterPosition + FlyStep(speed, repulsionForce.Yaw.Facing));
 		}
 
+		public void DisableRepulsing()
```
pchote: This looks dangerous.  I would strongly prefer it if we didn't do this – we have wasted so much time over the years debugging issues with manually managed state like this.

OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
bateramos: But if I call SequenceActivities with ReturnToBase as the last activity will fix this problem?

OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
bateramos: Ok, I will change that!

OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
pchote: It should work if you use `QueueActivity` instead.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
bateramos: The other alternative is creating a manager that control the queue and the positions of the helicopters to resupply. I could not find any other solution for this. The repulsion from others helicopters sometime knocks the helicopter that was descending away from the supply pad.
I see as two options. Create this controller that handle all positioning of the helicopters or disable the repulsion (maybe there should be a better way of disabling this).
But i really all hear to what do you thing is better. In the case of the controller, think I need a example class that I can base my work or just a north on how this should be done.

OpenRA.Mods.Common/Activities/Execute.cs
bateramos: Maybe I can help on this too. There is a forum thread or a issue on this?

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
pchote: There is already a check in GetRepulsionForce that disables the repulsion if the actor isn't at its cruising altitude.  I suspect the problem is that it gets repulsed during the one tick where it is at cruising height and preparing to land.  You could change that check to additionally disable repulsion if the actor is within X WRange of the reserved docking point.

bateramos: I made some changes in the last commit. I think your approach solved this issue.

OpenRA.Mods.Common/Activities/Execute.cs
obrakmann: Actually, the `CallFunc` activity still exists.

bateramos: But I can use this Class? Its used in 13 different places.

obrakmann: Yes, you can.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
```diff
@@ -32,8 +32,17 @@ public override Activity Tick(Actor self)
...
 
+			if (aircraft.IsPlane)
+				return ActivityUtils.SequenceActivities(
+					aircraft.GetResupplyActivities(host)
+					.Append(new CallFunc(() => aircraft.UnReserve()))
+					.Append(new WaitFor(() => Reservable.IsReserved(host)))
+					.Append(new TakeOff(self))
+					.Append(NextActivity).ToArray());
+
+			// If is helicopter move alway as soon as the resupply ends
```
Mailaender: Typo:

```patch
-alway
+away
```

Mailaender:

OpenRA.Mods.Common/Activities/Air/FlyCircleTimed.cs
```diff
@@ -0,0 +1,35 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2014 The OpenRA Developers (see AUTHORS)
```
Mailaender: 2016 and GPLv3 or later. https://github.com/OpenRA/OpenRA/pull/10801

Mailaender: You may also want to add yourself to `AUTHORS`.

OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
```diff
@@ -36,12 +36,15 @@ public FlyAttack(Actor self, Target target)
...
+			{
+				self.QueueActivity(new ReturnToBase(self));
 				return NextActivity;
+			}
 
 			// Move to the next activity only if all ammo pools are depleted and none reload automatically
 			// TODO: This should check whether there is ammo left that is actually suitable for the target
 			if (ammoPools.All(x => !x.Info.SelfReloads && !x.HasAmmo()))
-				return NextActivity;
+				return new ReturnToBase(self);
```
pchote: This is going to erase the order queue.  You need to sequence the RTB before continuing the NextActivity with something like `return ActivityUtils.SequenceActivities(new ReturnToBase(self), NextActivity)`.

OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
```diff
@@ -36,12 +36,15 @@ public FlyAttack(Actor self, Target target)
 		public override Activity Tick(Actor self)
 		{
 			if (!target.IsValidFor(self))
+			{
+				self.QueueActivity(new ReturnToBase(self));
```
pchote: This still doesn't look right, unfortunately  This appends the RTB at the end of the activity queue, which in theory should mean it is redundant vs the `ReturnOnIdle` trait.  If another activity is blocking then neither is going to run.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
```diff
@@ -32,8 +32,17 @@ public override Activity Tick(Actor self)
 			if (host == null)
 				return NextActivity;
 
+			if (aircraft.IsPlane)
+				return ActivityUtils.SequenceActivities(
+					aircraft.GetResupplyActivities(host)
+					.Append(new CallFunc(() => aircraft.UnReserve()))
+					.Append(new WaitFor(() => Reservable.IsReserved(host)))
+					.Append(new TakeOff(self))
+					.Append(NextActivity).ToArray());
```
pchote: SequenceActivities accepts a list as parameters, so building your own array shouldn't be needed.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -83,6 +83,12 @@ public class AircraftInfo : ITraitInfo, IPositionableInfo, IFacingInfo, IOccupyS
 		[Desc("Sound to play when the actor is landing.")]
 		public readonly string LandingSound = null;
 
+		[Desc("The distance of the resupply base that the aircraft will wait for its turn.")]
+		public readonly WDist WaitDistanceFromResupplyBase = new WDist(3072);
+
+		[Desc("The number of ticks that a airplane will wait to make a new search for a available airport.")]
+		public readonly int NumberOfTicksToVerifyAvailableAirport = 50;
```
pchote: This is only two seconds at normal game speed.  Isn't that a bit fast?

reaperrr:
Also needs a rebase.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
Phrohdoh: `an available`

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
bateramos: I used this way to reutilize the aircraft GetSupplyActivities, that return a IEnumerable<Activity>

bateramos:
sorry @reaperrr why i need to rebase? Is the commit message?

OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
```diff
@@ -44,13 +44,28 @@ public override Activity Tick(Actor self)
...
 				if (nearestHpad == null)
 					return ActivityUtils.SequenceActivities(new Turn(self, initialFacing), new HeliLand(self, true), NextActivity);
 				else
-					return ActivityUtils.SequenceActivities(new HeliFly(self, Target.FromActor(nearestHpad)));
+				{
+					var distanceFromHelipad = (nearestHpad.CenterPosition - self.CenterPosition).HorizontalLength;
+					var distanceLength = heli.Info.WaitDistanceFromResupplyBase.Length;
+
+					// If no pad is available, move near one and wait
+					if (distanceFromHelipad > distanceLength)
```
obrakmann: Shouldn't helicopters also move *away* from helipads when they're too close to make room for others?

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -181,6 +188,13 @@ public virtual WVec GetRepulsionForce()
 			if (!Info.Repulsable)
 				return WVec.Zero;
 
+			if (reservation != null)
+			{
+				var distanceFromReservationActor = (reservedActor.CenterPosition - self.CenterPosition).HorizontalLength;
+				if (distanceFromReservationActor < Info.WaitDistanceFromResupplyBase.Length)
+					return WVec.Zero;
+			}
```
obrakmann: I'm not sure if I like this (mind you, I haven't playtested it yet, so it might be alright). My first gut instinct would be to separate waiting helicopters vertically instead of disabling repulsion. Like a holding pattern, kinda.

OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
```diff
@@ -44,13 +44,28 @@ public override Activity Tick(Actor self)
 			{
 				var rearmBuildings = heli.Info.RearmBuildings;
 				var nearestHpad = self.World.ActorsHavingTrait<Reservable>()
-									.Where(a => a.Owner == self.Owner && rearmBuildings.Contains(a.Info.Name))
-									.ClosestTo(self);
+					.Where(a => a.Owner == self.Owner && rearmBuildings.Contains(a.Info.Name))
+					.ClosestTo(self);
```
obrakmann: Eventually we might to add the behaviour the harvesters use for refineries now (ie, three harvs per refinery, if it's full, move on to the next one).

obrakmann:
In `HeliReturnToBase` (after line 80), you need to add `NextActivity` behind `ResupplyAircraft`.

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -108,7 +108,10 @@ public override Activity Tick(Actor self)
 
 				self.CancelActivity();
 				if (nearestAfld != null)
-					return ActivityUtils.SequenceActivities(new Fly(self, Target.FromActor(nearestAfld)), new FlyCircle(self));
+					return ActivityUtils.SequenceActivities(
+						new Fly(self, Target.FromActor(nearestAfld), WDist.Zero, plane.Info.WaitDistanceFromResupplyBase),
+						new FlyCircleTimed(plane.Info.NumberOfTicksToVerifyAvailableAirport, self),
+						new ReturnToBase(self));
```
obrakmann: Use `this` instead of creating a new instance. I had better results with that change. The activity queue probably gets lost otherwise.

obrakmann:
This looks very nice indeed. This just needs a couple of changes to make sure the activity queue doesn't get lost, then it's good to go. :+1: 

I'll still need to check how this impacts scripted missions, though.

Rebasing this won't be very hard, git manages to auto-merge the changes, so you don't actually have to do anything besides running the rebase command.

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -108,7 +108,10 @@ public override Activity Tick(Actor self)
 
 				self.CancelActivity();
```
obrakmann: This isn't actually needed.

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -108,7 +108,10 @@ public override Activity Tick(Actor self)
...
 				if (nearestAfld != null)
-					return ActivityUtils.SequenceActivities(new Fly(self, Target.FromActor(nearestAfld)), new FlyCircle(self));
+					return ActivityUtils.SequenceActivities(
+						new Fly(self, Target.FromActor(nearestAfld), WDist.Zero, plane.Info.WaitDistanceFromResupplyBase),
+						new FlyCircleTimed(plane.Info.NumberOfTicksToVerifyAvailableAirport, self),
+						new ReturnToBase(self));
 				else
 					return new FlyCircle(self);
```
obrakmann: Please change this to `return NextActivity`. Otherwise (some of) the mission scripts break, and 2) if `NextActivity` is null, aircraft will end up circling by themselves anyway, thanks to `FlyAndContinueWithCirclesWhenIdle`.

OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
bateramos: Hmmmm I don't think this is necessary. In my tests, the problem occurs when the helicopters were approaching to resupply. When are fully supplied, they already form a line in the helipad exit waypoint.

bateramos: Hmmmm I don't think this is necessary. In my tests, the problem occurs when the helicopters were approaching to resupply. When are fully supplied, they already form a line in the helipad exit waypoint.

obrakmann: I was thinking of helicopters that are already there for other reasons, because the player ordered them there for example, not because they were sent there for rearming. But I agree that it's a corner case that might not actually matter in the end.

bateramos: I just made a test with 10 helicopters on top of the helipad, indeed the heli that just resupplied get "stuck" with the others, but this does not prevent others heli from resupply, as on approach the repulsion is disabled. Its not "elegant" but it is funcional.

bateramos:
Sorry to disturbe again with this merge messages, but how I can remove this from my branch?

obrakmann:
Just run a rebase once, it will remove the merge commit.

Looks good to me now, :+1:

obrakmann:
This unfortunately needs a rebase again. I'd hold off from doing that until this gets the second :+1:, though. I won't add the tag either to keep people from intentionally skipping this pr.

Mailaender:
Still works as promised. :white_check_mark:

bateramos:
the rebase is now ok?
Checksum error.. What! Why? How?

Mailaender:
Our Windows CI https://ci.appveyor.com/project/OpenRA/openra/history is unstable for some reason. Don't worry. The build went fine. Just the setup.exe creation failed, because https://chocolatey.org/packages/nsis didn't install properly.

Mailaender:
I hope https://github.com/OpenRA/OpenRA/pull/11644 will help in the future.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/bdc6e4397091d77a8b7e7989313d0a4f5c521eab)

PR #11394, opened 2016-06-04
Restricted effect rendering to the viewport

Mailaender:
Another take on https://github.com/OpenRA/OpenRA/issues/9443 after https://github.com/OpenRA/OpenRA/issues/9699 grew way too complex and error prone.

### DIFF
--- a/OpenRA.Game/Graphics/Viewport.cs
+++ b/OpenRA.Game/Graphics/Viewport.cs
@@ -310,5 +310,10 @@ namespace OpenRA.Graphics
 				return allCells;
 			}
 		}
+
+		public Rectangle Bounds
+		{
+			get { return new Rectangle(TopLeft.X, TopLeft.Y, viewportSize.X, viewportSize.Y); }
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


pchote:
Unfortunately, like I mentioned in IRC earlier, I do not think this a good approach:

* It ignores the existing conventions that we use for actor rendering, and instead invents its own.
* We still pay the cost of generating the renderables, and then pay an additional cost of a duplicate `PrepareRender` on each.

I suspect that these tradeoffs may completely negate the benefits that we want from #9443.

reaperrr:
> We still pay the cost of generating all the renderables (which is a significant chunk of the rendering!), and then pay an additional cost of a duplicate PrepareRender on each.

Just a thought, would it be possible/feasible to simply create some "is WPos in player's viewport" boolean (similar to the `FogObscures` check, just hopefully much cheaper in terms of performance), and then only generate the renderable if that is `true`?

pchote:
Yes (except in screen space instead of world).  This is essentially what actors and #9699 do.

Mailaender:
I tried it in https://github.com/OpenRA/OpenRA/issues/9443 and it doesn't work for text and line renderables, because they don't define a position and don't have a valid bound ready in the constructor, because that is calculated just before getting rendered.

reaperrr:
@Mailaender isn't the 2nd commit possibly the fix for the issue you had in #9699?

Mailaender:
Sadly screen bounds for actors are offset + bounds and effects are location + bounds. Plus all effects will crash the game as they have empty screen bounds in constructor and the first iterations of Tick.

pchote:
Thosw are implementation details/issues with those specific effects (which can be fixed).  Adding engine level workarounds for bad code just cements the technical debt and makes it harder for the next person.

Issue #3476, opened 2013-06-27
Have enemy units reveal themselves when they shoot at your units


ScottNZ:
This should go some ways into fixing the artillery balance problems.

GraionDilach:
For the record, TS/RA2 worked like this.

MustaphaTR:
In TS/RA2 there was a Weapon tag as `RevealOnFire=` which is default to yes. But some weapons like Sniper's `[AWP]` has changed to `no`.

reaperrr:
This has been decided as "won't fix" for RA/TD, and has been adressed for D2k & TS. Engine support was added with https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Effects/RevealShroudEffect.cs (enabled via `RevealOnFire` trait).

PR #11396, opened 2016-06-04
Speed up sync by only syncing projectiles

Mailaender:
This improves CPU performance on a Thinkpad X230 from an average 12,4 ms at commit 910d04f (red) to 11,3 ms with this pull request (green) measured over 1000 ticks at the Red Alert shellmap with #10133.

image

When I wrote https://github.com/OpenRA/OpenRA/pull/3741 we were hunting weapon desyncs and syncing everything including decorations was just a side effect, because the engine didn't differentiate properly.


pchote:
We need an `IProjectile`-like interface to fix continuous-fire weapons and support power beacons, so this PR is worthwhile even if the perf changes are negligible.

OpenRA.Game/Network/OrderManager.cs
```diff
@@ -52,7 +52,7 @@ public sealed class OrderManager : IDisposable
 
 		bool disposed;
 
-		void OutOfSync(int frame)
+		public void OutOfSync(int frame)
```
pchote: This doesn't appear to be required as part of this PR?

Mailaender: Yes, this was a left-over. I trigger an `OutOfSync` exception after 500 frames to test this.

Mailaender:
I wouldn't say ~10% performance improvements are completely negligible.

pchote:
This is an aside that doesn't impact the reviewing, but you did bring it up :smile:

If you did a straight mean (no sigma clipping) then I don't at all believe the 10% improvement figure.  You can see by eye that the difference is dominated by the large humps in the red at ~150, 500, 550 ticks, which aren't likely to be related to these changes (you'd need a massive spike in non-projectile effects to see any significant change in the effect syncing time).

Mailaender:
My assumption is that a huge number of non-projectile sprite effects like infantry corpses can have a significant impact in large scale battles and that is what happened here, but I agree that the graph shows that the Red Alert shellmap may not be reproducible enough. I did the plotting and averaging with http://www.gnumeric.org/ How do you calculdate the straight mean (no sigma clipping) with spreadsheets?

OpenRA.Game/World.cs
```diff
@@ -366,9 +371,9 @@ public int SyncHash()
 						ret += n++ * (int)(1 + actor.ActorID) * syncHash.Hash;
 
 				// TODO: don't go over all effects
```
reaperrr: Nit: This comment can be removed with this PR.

reaperrr:
Aside from that nit, this looks good to me. :+1: 

> you'd need a massive spike in non-projectile effects to see that big of a change in the effect hashing time

Well, lots of weapons shooting also means lots of contrails, sprite trails, explosions and corpses, so heated battles should be where this makes the biggest difference.

I agree that a 10% average sounds far too high, though, and around Tick 150 there isn't any battling on the RA shellmap yet.

OpenRA.Game/World.cs
Mailaender: Removed. I actually had this in mind as a main goal. Weird that I forgot to erase it.

obrakmann:
We may eventually want to sync effects other than projectiles (for example #5061 and #3476 will probably be solved by an IEffect that reveals a piece of shroud). In that case, limiting ourselves to IProjectile would be bad.

I propose to have this hierarchy:

`IEffect`
`IEffectSynced : IEffect, ISync`
`IProjectile : IEffectSynced`

`World` would then keep lists for both `IEffect` and `IEffectSynced`, but not for `IProjectile`.

pchote:
We could avoid adding an extra `IEffectSynced` by having the world check if the projectile implements `ISync`.  Please lets keep `IProjectile` for the reasons I stated above, even though it would no longer be necessary for the original purpose of the PR.

Mailaender:
Rebased and added the new interface.

GraionDilach:
IonCannon.cs and NukeLaunch.cs should be shifted to the Projectiles folder from the Effects folder then IMO.

pchote:
Removing the original :+1:, since this has had a major change.

pchote:
I'm not sure where you had trouble, this is all you need to avoid the extra interface:
```diff
diff --git a/OpenRA.Game/GameRules/WeaponInfo.cs b/OpenRA.Game/GameRules/WeaponInfo.cs
index cb151ee..973ee5b 100644
--- a/OpenRA.Game/GameRules/WeaponInfo.cs
+++ b/OpenRA.Game/GameRules/WeaponInfo.cs
@@ -31,8 +31,7 @@ public class ProjectileArgs
 		public Target GuidedTarget;
 	}
 
-	public interface IEffectSynced :  IEffect, ISync { }
-	public interface IProjectile : IEffectSynced { }
+	public interface IProjectile : IEffect { }
 	public interface IProjectileInfo { IProjectile Create(ProjectileArgs args); }
 
 	public sealed class WeaponInfo
diff --git a/OpenRA.Game/World.cs b/OpenRA.Game/World.cs
index fadab7f..e43672c 100644
--- a/OpenRA.Game/World.cs
+++ b/OpenRA.Game/World.cs
@@ -31,7 +31,7 @@ public sealed class World : IDisposable
 		internal readonly TraitDictionary TraitDict = new TraitDictionary();
 		readonly SortedDictionary<uint, Actor> actors = new SortedDictionary<uint, Actor>();
 		readonly List<IEffect> effects = new List<IEffect>();
-		readonly List<IEffectSynced> syncedEffects = new List<IEffectSynced>();
+		readonly List<IEffect> syncedEffects = new List<IEffect>();
 		readonly Queue<Action<World>> frameEndActions = new Queue<Action<World>>();
 
 		public int Timestep;
@@ -274,10 +274,8 @@ public void Remove(Actor a)
 				t.RemovedFromWorld(a);
 		}
 
-		public void Add(IEffect b) { effects.Add(b); }
-		public void Add(IEffectSynced b) { effects.Add(b); syncedEffects.Add(b); }
-		public void Remove(IEffect b) { effects.Remove(b); }
-		public void Remove(IEffectSynced b) { effects.Remove(b); syncedEffects.Remove(b); }
+		public void Add(IEffect b) { effects.Add(b); if (b is ISync) syncedEffects.Add(b); }
+		public void Remove(IEffect b) { effects.Remove(b); if (b is ISync) syncedEffects.Remove(b); }
 		public void RemoveAll(Predicate<IEffect> predicate) { effects.RemoveAll(predicate); syncedEffects.RemoveAll(predicate); }
 
 		public void AddFrameEndTask(Action<World> a) { frameEndActions.Enqueue(a); }
@@ -338,7 +336,7 @@ public void TickRender(WorldRenderer wr)
 
 		public IEnumerable<Actor> Actors { get { return actors.Values; } }
 		public IEnumerable<IEffect> Effects { get { return effects; } }
-		public IEnumerable<IEffectSynced> SyncedEffects { get { return syncedEffects; } }
+		public IEnumerable<IEffect> SyncedEffects { get { return syncedEffects; } }
 
 		public Actor GetActorById(uint actorId)
 		{
```

Mailaender:
Updated.

OpenRA.Game/World.cs
```diff
@@ -30,6 +31,8 @@ public sealed class World : IDisposable
 		internal readonly TraitDictionary TraitDict = new TraitDictionary();
 		readonly SortedDictionary<uint, Actor> actors = new SortedDictionary<uint, Actor>();
 		readonly List<IEffect> effects = new List<IEffect>();
+		readonly List<IEffect> syncedEffects = new List<IEffect>();
```
RoosterDragon: A `List<ISync>` would make more sense - all the current loops care about the sync interface, rather than the effect interface.

Mailaender: That is true. Updated.

Mailaender:
I assume this will have a greater effect on mods like http://www.moddb.com/mods/cc-tiberian-sun-project-nemesis/ that have lot's of decorational effects. Especially syncing the dust cloud movement trails can be saved this way.

OpenRA.Game/Network/SyncReport.cs
```diff
@@ -77,9 +77,8 @@ void GenerateSyncReport(Report report)
 							NamesValues = DumpSyncTrait(syncHash.Trait)
 						});
 
-			foreach (var e in orderManager.World.Effects)
+			foreach (var sync in orderManager.World.SyncedEffects)
 			{
-				var sync = e as ISync;
 				if (sync != null)
```
abcdefg30: Do we still need this null check now?

OpenRA.Game/World.cs
```diff
@@ -365,10 +369,9 @@ public int SyncHash()
 					foreach (var syncHash in actor.SyncHashes)
 						ret += n++ * (int)(1 + actor.ActorID) * syncHash.Hash;
 
-				// TODO: don't go over all effects
-				foreach (var e in Effects)
+				// hash game state relevant effects such as projectiles
+				foreach (var sync in SyncedEffects)
 				{
-					var sync = e as ISync;
 					if (sync != null)
```
abcdefg30: Same here.

OpenRA.Game/World.cs
```diff
@@ -365,10 +369,9 @@ public int SyncHash()
 					foreach (var syncHash in actor.SyncHashes)
 						ret += n++ * (int)(1 + actor.ActorID) * syncHash.Hash;
 
-				// TODO: don't go over all effects
-				foreach (var e in Effects)
+				// hash game state relevant effects such as projectiles
```
abcdefg30: Comments start with a capital letter, please.

OpenRA.Game/World.cs
```diff
@@ -272,8 +275,8 @@ public void Remove(Actor a)
 				t.RemovedFromWorld(a);
 		}
 
-		public void Add(IEffect b) { effects.Add(b); }
-		public void Remove(IEffect b) { effects.Remove(b); }
+		public void Add(IEffect b) { effects.Add(b); if (b is ISync) syncedEffects.Add(b as ISync); }
+		public void Remove(IEffect b) { effects.Remove(b); if (b is ISync) syncedEffects.Remove(b as ISync); }
 		public void RemoveAll(Predicate<IEffect> predicate) { effects.RemoveAll(predicate); }
```
abcdefg30: Won't this also need to remove `predicate` from `syncedEffects`?

OpenRA.Game/World.cs
Mailaender: Not in this file though.

OpenRA.Game/World.cs
Mailaender: No synced effect uses this predicate and I can't just add it as that won't compile.

OpenRA.Game/Network/SyncReport.cs
Mailaender: I am not sure. We don't check for null at earlier stages so I would like to keep it.

OpenRA.Game/World.cs
RoosterDragon: `syncedEffects.RemoveAll(e => predicate((IEffect)e))`

OpenRA.Game/Network/SyncReport.cs
pchote: You check before adding them to SyncedEffects, so this can be removed.

OpenRA.Game/World.cs
```diff
@@ -272,8 +275,8 @@ public void Remove(Actor a)
 				t.RemovedFromWorld(a);
 		}
 
-		public void Add(IEffect b) { effects.Add(b); }
-		public void Remove(IEffect b) { effects.Remove(b); }
+		public void Add(IEffect b) { effects.Add(b); if (b is ISync) syncedEffects.Add(b as ISync); }
```
pchote: `var sb = b as ISync; if (sb != null) syncedEffects.Add(sb);`

abcdefg30: Could we then stretch this over multiple lines, while we're at it?

Mailaender:
Updated for even more PERF :rocket: and readability.

OpenRA.Game/World.cs
Mailaender: Thanks!

OpenRA.Game/World.cs
```diff
@@ -356,24 +378,20 @@ public int SyncHash()
 				var n = 0;
 				var ret = 0;
 
-				// hash all the actors
+				// Hash all the actors.
 				foreach (var a in Actors)
 					ret += n++ * (int)(1 + a.ActorID) * Sync.HashActor(a);
 
-				// hash all the traits that tick
+				// Hash all the traits that tick.
```
pchote: Since you've already touched this, can you change it to be correct or just remove it?
It has been a very long time since this has acted on the traits that tick.

pchote:
Looks fine now otherwise :+1:

OpenRA.Game/World.cs
Mailaender: Updated.

reaperrr:
This is not a fixup request (yet), just a question for now: Do we have a specific reason why `LaserZap` and `TeslaZap` don't implement `ISync`?

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/bf9b8afe04ca0b181c325df7e273003ca6f07290...92285c894e789552c93fabd0bd90af71c4236ca7)

PR #11405, opened 2016-06-05
Added burning trees to Tiberian Sun

Mailaender:
I went a different direction compared to Red Alert and Tiberian Dawn, because Tiberian Sun doesn't seem to have tree husks. So here nothing is really damaged or killed. Instead a decorational overlay is enabled via the upgrade system that wears off after some time. The benefit is that this allows per tree offsets and sequences.

It turned out to be very fiddly and I was very annoyed that it didn't work. I finally found out that `UpgradeManager` is missing for trees when I added a `TimedUpgradeBar` for debugging purposes which requires `UpgradeManager` and added a violent, but helpful crash. Added something similar to all optionally upgradeable traits.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradableTrait.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradableTrait.cs
@@ -41,7 +41,7 @@ namespace OpenRA.Mods.Common.Traits
 	/// Requires basing *Info on UpgradableTraitInfo and using base(info) constructor.
 	/// Note that EnabledByUpgrade is not called at creation even if this starts as enabled.
 	/// </summary>
-	public abstract class UpgradableTrait<InfoType> : IUpgradable, IDisabledTrait, ISync where InfoType : UpgradableTraitInfo
+	public abstract class UpgradableTrait<InfoType> : IUpgradable, INotifyCreated, IDisabledTrait, ISync where InfoType : UpgradableTraitInfo
 	{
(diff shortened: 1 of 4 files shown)


pchote:
I'm not familiar with how the original game handled burning trees.  Is this consistent with the original behaviour?

GraionDilach:
The original game killed trees when the flames went off. Might as well worthy to introduce that before we end up with dozens of OpenRA-TS maps where trees are placed with the intention of invulnerable.

Mailaender:
I didn't find sprites for killed trees. Do they just disappear?

pchote:
> Might as well worthy to introduce that before we end up with dozens of OpenRA-TS maps where trees are placed with the intention of invulnerable.

I agree completely.

Mailaender:
But how is tree destruction actually done in the original Tiberian Sun?

Biofreak1987:
Quickly checked it, they just disappear when you burn them. No burned tree sprites except for the flame overlay.

Mailaender:
Thanks. I just watched some videos on YouTube with Devils Tongue's and there I just saw flames on trees. They also looked different so I am not sure if I even identified the right overlay sequences.

GraionDilach:
https://github.com/LipkeGu/TSIniFiles/blob/master/rules.ini#L354
https://github.com/LipkeGu/TSIniFiles/blob/master/art.ini#L3670

In our language, Fire2 should have the combined sequence of 2 and 1.

PR #11406, opened 2016-06-05
Added an error message when setting up upgrades without UpgradeManager

Mailaender:
A major downside from actor upgrades is that they can be setup in ways where nothing happens and you don't get any feedback why. This should help reduce the annoyance. Split from https://github.com/OpenRA/OpenRA/pull/11405.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckUpgrades.cs
+++ b/OpenRA.Mods.Common/Lint/CheckUpgrades.cs
@@ -31,6 +31,9 @@ namespace OpenRA.Mods.Common.Lint
 
 			foreach (var actorInfo in rules.Actors)
 			{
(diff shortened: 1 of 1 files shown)


pchote:
Can you please fix the `CheckUpgrades` lint test instead of adding runtime-only code for this?

Mailaender:
Good idea. That also gives more specific error messages.

obrakmann:
:+1:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/338916fa730db81da97a43de422d5781a20d08d9...f009259e41a5b4d97314fcea9dcc8981160c6fc6)

PR #11407, opened 2016-06-05
Various TS infantry fixes

reaperrr:
- fixed zapped animation colors
- cut down a lot of redundancy in sequences (I recommend looking at the full file, not just the diff)
- implemented TS-style Cyborg death explosions
- fixed crippled cyborgs returning to full speed once regaining health

The last remaining bit to make cyborgs work like in TS would be to play an explosion on the normal -> crippled transition, but I need more time to come up with something non-hacky.


pchote:
> The last remaining bit to make cyborgs work like in TS would be to play an explosion on the normal -> crippled transition, but I need more time to come up with something non-hacky.

Just write a trait that implements `INotifyDamageStateChanged` and fires a weapon.  The upgrades stuff didn't *completely* supersede the ability to write special-case traits :wink:

mods/ts/sequences/infantry.yaml
```diff
@@ -45,110 +47,263 @@ e1.gdi:
...
+		Start: 110
+		Length: 9
+		Facings: 8
+		ShadowStart: 418
+	crippled-stand:
+		Start: 110
+		Facings: 8
+		Stride: 9
+		ShadowStart: 418
 	die1:
```
GraionDilach: What's the point in keeping/editing these? You removed WithDeathAnimation entirely from cyborgs with the new template.

GraionDilach:
Regarding infdie.shp and it's removal from HeadshotDeath: I'm fine with that, but IMO purge HeadshotDeath entirely then, IIRC it's only used at Umagon - remember http://logs.openra.net/?year=2015&month=07&day=18#22:17:12 - This could also be an opportunity to restore die3 into ExplosionDeath properly (s_bang34) although I dunno if a weapon in TS actually used that deathtype.

mods/ts/sequences/infantry.yaml
reaperrr: I just forgot to remove them, tbh.

mods/ts/sequences/infantry.yaml
```diff
@@ -1,5 +1,5 @@
-e1.gdi:
-	Defaults: e1
+^e1:
```
pchote: Can we call this `^BasicInfantry` or something similar?

mods/ts/sequences/infantry.yaml
```diff
@@ -45,110 +47,263 @@ e1.gdi:
...
-	shoot:
+	standup-0:
+		Start: 260
+		Length: 2
+		Facings: 8
+		ShadowStart: 552
+	die6: electro
+		Length: *
+
+^e1attack:
```
pchote: Likewise here.

mods/ts/rules/defaults.yaml
```diff
@@ -339,8 +304,50 @@
...
+		Voice: Zapped
+		DeathTypes: EnergyDeath
+	Crushable:
+		CrushSound: squish6.aud
+	SpawnActorOnDeath:
+		Actor: visc_sml
+		Probability: 10
+		OwnerType: InternalName
+		InternalOwner: Creeps
+		DeathType: EnergyDeath # TODO: FIX ME! (Tiberium currently uses the wrong damage type!)
```
pchote: Wouldn't this be a great time to fix that?

mods/ts/rules/defaults.yaml
```diff
@@ -339,8 +304,50 @@
 		UpgradeTypes: hospitalheal
 		UpgradeMinEnabledLevel: 1
 
+^RegularInfantryDeath:
+	WithDeathAnimation@normal:
+		DeathTypes:
+			BulletDeath: 1
```
pchote: Could we change these to use descriptive death suffixes instead of integers? I'm assuming that you only need to change a handful of places now that you've deduplicated both the rules and sequence setup.

reaperrr:
Updated.

reaperrr:
Updated again.

mods/ts/rules/defaults.yaml
```diff
@@ -339,8 +304,52 @@
...
+		DeathSequencePalette: ra
+		DeathPaletteIsPlayerPalette: False
+		DeathTypes:
+			EnergyDeath: melting
+	WithDeathAnimation:
+		CrushedSequence: die-crushed
+		CrushedSequencePalette: player
+		CrushedPaletteIsPlayerPalette: true
+	DeathSounds@NORMAL:
+		DeathTypes: BulletDeath, SmallExplosionDeath, ExplosionDeath
```
GraionDilach: You sure about this? Shouldn't ExplosionDeath have one of the expnew auds as it's sound?

reaperrr: It should have both, actually. I just played the original and there was a death scream in addition to the explosion. I'll leave this entry as-is and add a `DeathSounds@EXPLOSION` that only triggers on `ExplosionDeath`.

reaperrr:
Updated.

GraionDilach:
:+1:

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamage.cs
```diff
@@ -29,26 +29,32 @@ public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>
 		[Desc("Levels of damage at which to grant upgrades.")]
 		public readonly DamageState ValidDamageStates = DamageState.Heavy | DamageState.Critical;
 
+		[Desc("Are upgrades irrevokable once the conditions have been met?")]
```
abcdefg30: `irrevokable` -> `irrevocable`

abcdefg30:

Mailaender:
debug.log now spams:

```
File not found: ^BasicInfantry.shp
File not found: ^BasicInfantry.shp
File not found: ^BasicInfantry.shp
File not found: ^BasicInfantry.shp
File not found: ^BasicInfantry.shp
File not found: ^BasicInfantry.shp
File not found: ^BasicInfantry.shp
File not found: ^BasicInfantry.shp
File not found: ^BasicInfantry.shp
File not found: ^BasicInfantry.shp
File not found: ^BasicInfantryAttack.shp
File not found: ^BasicInfantryAttack.shp
```

reaperrr:
I might be mistaken, but to me it looks like this is rather our engine's fault for not recognizing those as abstract sequences, rather than this PR's.

Mailaender:
That wasn't meant as a reproach. It just needs to be fixed one way or the other.

PR #11410, opened 2016-06-06
Fix possible NREs in RenderAsync in VoxelRenderer.cs

abcdefg30:
Detected by Coverity.

[`MatrixInverse`](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Graphics/Util.cs#L318) might return `null` for `invCameraTransform` which will then crash in [`MatrixVectorMultiply`](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Graphics/Util.cs#L194).


pchote:
If the inverse camera transform is null then we can't do anything.  Continuing the rendering without applying that transform is bogus.

There are only two sensible ways to handle this case:
* throw an exception
* give up and don't render anything

abcdefg30:
Updated.

obrakmann:
:+1:

Mailaender:
Thanks. No changelog as these are theoretical problems.

Issue #4561, opened 2014-02-04
AI should make use of oil derricks


pizzaoverhead:
Currently AI players ignore capturable oil derricks. Capturing these should be prioritised from the start. When oil derricks are available and ore is limited or unavailable, AI players should not build multiple refineries. Small teams of infantry and light vehicles could be assembled to escort the engineers or their transports, and then guard the captured derricks. An AI-owned derrick under attack should cause the AI to send support units and repair it if damaged.

MustaphaTR:
They now capture the Oil Derricks after AI Engineer logic implemented. Also Normal and Turtle use them with my PR. I think this may be closed.

pchote:
@MustaphaTR: add "Fixes #4561" to the description of your PR, and github will automatically close this when it is merged.  We do not manually close tickets before a PR is merged.

obrakmann:
This is already in the playtest, if only for one AI, so closing is fine, IMO.

pchote:
See #12206.

PR #11417, opened 2016-06-07
Add naive capturing to HackyAI

phrohdoh:
Currently only RA's Rush AI is setup in a testable manner.

This PR adds these fields to `HackyAIInfo`:

``` csharp
[Desc("Actor types that can capture other actors (via `Captures` or `ExternalCaptures`).",
    "Leave this empty to disable capturing.")]
public HashSet<string> CapturingActorTypes = new HashSet<string>();

[Desc("Actor types that can be targeted for capturing.",
      "Leave this empty to include all actors.")]
public HashSet<string> CapturableActorTypes = new HashSet<string>();

[Desc("Minimum delay (in ticks) between trying to capture with CapturingActorTypes.")]
public readonly int MinimumCaptureDelay = 50;

[Desc("Maximum number of options to consider for capturing.")]
public readonly int MaximumCaptureTargetOptions = 5;

[Desc("Should visibility (Shroud, Fog, Cloak, etc) be considered when searching for capturable targets?")]
public readonly bool CheckCaptureTargetsForVisibility = true;
```

`CheckCaptureTargetsForVisibility` defaulting to `true` means exploration must happen before any of this new code has any noticeable effect, should `false` be a better default value I have no objections to changing it.

@obrakmann made the statement:

>  instead of the *CaptureTargetOptions, I'd have a list of actor types that are suitable for capturing

This is fair, I can implement this once this PR gets some initial review going.

Here is a series of images from a test game where the orange AI captured a conyard and built off of it before the original owner captured it back:

screen shot 2016-06-07 at 1 35 53 pm
screen shot 2016-06-07 at 1 35 58 pm
screen shot 2016-06-07 at 1 36 11 pm

I am leaving the commits as they are for the initial reviews, they will be squashed and fixed up as necessary if this goes anywhere.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -16,12 +16,28 @@ using System.Linq;
 using OpenRA.Mods.Common.Activities;
 using OpenRA.Mods.Common.Pathfinder;
 using OpenRA.Mods.Common.Traits;
-using OpenRA.Primitives;
 using OpenRA.Support;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.AI
 {
+	class CaptureTarget<TInfoType> where TInfoType : class, ITraitInfoInterface
+	{
+		internal readonly Actor Actor;
+		internal readonly TInfoType Info;
+
+		/// <summary>The order string given to the capturer so they can capture this actor.</summary>
+		/// <example>ExternalCaptureActor</example>
+		internal readonly string OrderString;
+
+		internal CaptureTarget(Actor actor, string orderString)
+		{
+			Actor = actor;
+			Info = actor.Info.TraitInfoOrDefault<TInfoType>();
+			OrderString = orderString;
+		}
+	}
+
 	public sealed class HackyAIInfo : IBotInfo, ITraitInfo
 	{
 		public class UnitCategories
@@ -174,6 +190,27 @@ namespace OpenRA.Mods.Common.AI
 		[FieldLoader.LoadUsing("LoadDecisions")]
 		public readonly List<SupportPowerDecision> PowerDecisions = new List<SupportPowerDecision>();
 
+		[Desc("Actor types that can capture other actors (via `Captures` or `ExternalCaptures`).",
+			"Leave this empty to disable capturing.")]
+		public HashSet<string> CapturingActorTypes = new HashSet<string>();
+
+		[Desc("Actor types that can be targeted for capturing.",
+			"Leave this empty to include all actors.")]
+		public HashSet<string> CapturableActorTypes = new HashSet<string>();
+
+		[Desc("Minimum delay (in ticks) between trying to capture with CapturingActorTypes.")]
+		public readonly int MinimumCaptureDelay = 375;
+
+		[Desc("Maximum number of options to consider for capturing.",
+			"If a value less than 1 is given 1 will be used instead.")]
+		public readonly int MaximumCaptureTargetOptions = 10;
+
+		[Desc("Should visibility (Shroud, Fog, Cloak, etc) be considered when searching for capturable targets?")]
+		public readonly bool CheckCaptureTargetsForVisibility = true;
+
+		[Desc("Player stances that capturers should attempt to target.")]
+		public readonly Stance CapturableStances = Stance.Enemy | Stance.Neutral;
+
 		static object LoadUnitCategories(MiniYaml yaml)
 		{
 			var categories = yaml.Nodes.First(n => n.Key == "UnitsCommonNames");
@@ -254,6 +291,8 @@ namespace OpenRA.Mods.Common.AI
 		int assignRolesTicks;
 		int attackForceTicks;
 		int minAttackForceDelayTicks;
+		int minCaptureDelayTicks;
+		readonly int maximumCaptureTargetOptions;
 
 		readonly Queue<Order> orders = new Queue<Order>();
 
@@ -279,6 +318,8 @@ namespace OpenRA.Mods.Common.AI
 
 			foreach (var decision in info.PowerDecisions)
 				powerDecisions.Add(decision.OrderName, decision);
+
+			maximumCaptureTargetOptions = Math.Max(1, Info.MaximumCaptureTargetOptions);
 		}
 
 		public static void BotDebug(string s, params object[] args)
@@ -311,6 +352,7 @@ namespace OpenRA.Mods.Common.AI
 			assignRolesTicks = Random.Next(0, Info.AssignRolesInterval);
 			attackForceTicks = Random.Next(0, Info.AttackForceInterval);
 			minAttackForceDelayTicks = Random.Next(0, Info.MinimumAttackForceDelay);
+			minCaptureDelayTicks = Random.Next(0, Info.MinimumCaptureDelay);
 
 			var tileset = World.Map.Rules.TileSet;
 			resourceTypeIndices = new BitArray(tileset.TerrainInfo.Length); // Big enough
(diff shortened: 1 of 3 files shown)


cjshmyr:
I'd like to see @obrakmann's suggestion, otherwise everything looks okay.  I think this will teach new players about the danger of engineers in your base.

Phrohdoh:
Updated with `HashSet<string> CapturableActorTypes`.

MustaphaTR:
Why just RA?

Phrohdoh:
There is no reason to setup the other mods if this is still in flux.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -626,6 +668,97 @@ void AssignRolesToIdleUnits(Actor self)
...
+		{
+			if (Info.CapturingActorTypes.Count == 0 || Player.WinState != WinState.Undefined)
+				return;
+
+			var capturers = unitsHangingAroundTheBase.Where(a => a.IsIdle && Info.CapturingActorTypes.Contains(a.Info.Name)).ToArray();
+			if (capturers.Length == 0)
+				return;
+
+			if (combatantPlayers == null)
+				combatantPlayers = World.Players.Where(p => p.InternalName != "Everyone").ToArray();
```
GraionDilach: Don't you need to check for Spectating and/or alliance here?

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -626,6 +668,97 @@ void AssignRolesToIdleUnits(Actor self)
...
+			var randPlayer = combatantPlayers.Random(Random);
+
+			var targetOptions = Info.CheckCaptureTargetsForVisibility
+				? GetVisibleActorsBelongingToPlayer(randPlayer)
+				: GetActorsThatCanBeOrderedByPlayer(randPlayer);
+
+			var capturableTargetOptions = targetOptions
+				.Select(a => new CapturerTarget<CapturableInfo>(a, "CaptureActor"))
+				.Where(target => target.Info != null && capturers.Any(capturer => target.Info.CanBeTargetedBy(capturer, target.Actor.Owner)))
+				.OrderByDescending(target => target.Actor.GetSellValue())
```
GraionDilach: A better option IMO was to capture the closest capturable actor or a mixture of both. This will make the AI blindly aiming for conyards with engineers, trying to march through the base from midgame onwards.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -22,6 +22,23 @@
 
 namespace OpenRA.Mods.Common.AI
 {
+	class CapturerTarget<TInfoType> where TInfoType : class, ITraitInfoInterface
```
pchote: Why not `CaptureTarget`?

OpenRA.Mods.Common/AI/HackyAI.cs
pchote: This is already done below.

pchote:
The code looks reasonable, but I haven't been able to test this yet.

Can you also hook up one of the TD AIs so we can test the internal capturing logic?

OpenRA.Mods.Common/AI/HackyAI.cs
Phrohdoh: I didn't want to imply `Capture` vs `ExternalCapture` or anything else.
I can of course change this.

OpenRA.Mods.Common/AI/HackyAI.cs
Phrohdoh: I had not thought of this, good catch.

OpenRA.Mods.Common/AI/HackyAI.cs
Phrohdoh: If desired I could swap the order.
Closest then sell value.

pchote: I think that would give more natural looking behavior, so please do.

OpenRA.Mods.Common/AI/HackyAI.cs
pchote: IMO you wouldn't lose any generality by renaming it.  The current name feels rather awkward, which then implies (to me at least) that you might be using it in an awkward way when you're actually not.

OpenRA.Mods.Common/AI/HackyAI.cs
Phrohdoh: I recall why I did not do this now.

I don't have access to an individual capturer yet meaning this logic becomes much more difficult to read.

Phrohdoh:
Renamed `CapturerTarget` to `CaptureTarget` and wired up CNC.
screen shot 2016-06-19 at 3 14 36 pm

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -167,6 +183,24 @@ public class BuildingCategories
...
+		[Desc("Actor types that can capture other actors (via `Captures` or `ExternalCaptures`).",
+			"Leave this empty to disable capturing.")]
+		public HashSet<string> CapturingActorTypes = new HashSet<string>();
+
+		[Desc("Actor types that can be targeted for capturing.",
+			"Leave this empty to include all actors.")]
+		public HashSet<string> CapturableActorTypes = new HashSet<string>();
+
+		[Desc("Minimum delay (in ticks) between trying to capture with CapturingActorTypes.")]
+		public readonly int MinimumCaptureDelay = 50;
```
obrakmann: I'd probably set this a bit higher, to 375 (15 seconds) maybe.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -626,6 +667,98 @@ void AssignRolesToIdleUnits(Actor self)
...
+				.Select(a => new CaptureTarget<CapturableInfo>(a, "CaptureActor"))
+				.Where(target => target.Info != null && capturers.Any(capturer => target.Info.CanBeTargetedBy(capturer, target.Actor.Owner)))
+				.OrderByDescending(target => target.Actor.GetSellValue())
+				.Take(maximumCaptureTargetOptions);
+
+			var externalCapturableTargetOptions = targetOptions
+				.Select(a => new CaptureTarget<ExternalCapturableInfo>(a, "ExternalCaptureActor"))
+				.Where(target => target.Info != null && capturers.Any(capturer => target.Info.CanBeTargetedBy(capturer, target.Actor.Owner)))
+				.OrderByDescending(target => target.Actor.GetSellValue())
+				.Take(maximumCaptureTargetOptions);
```
obrakmann: I feel that this could be parameterized somehow. Maybe a dictionary in the info, with the order name as the key and a string pair consisting of the names of the trait pair that make up the behaviour. Then this would even be able to support spy infiltration (I think). Just a suggestion if you feel like giving it a try, though, not required.

obrakmann:
Looks overall fine to me, :+1:

OpenRA.Mods.Common/AI/HackyAI.cs
Phrohdoh: Sure thing.

OpenRA.Mods.Common/AI/HackyAI.cs
Phrohdoh: I could give this a shot but this will require a rework as generics don't play well with reflection.

Phrohdoh: Sorry, this isn't going to happen in this PR.

Phrohdoh:
Squashed.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -630,6 +671,98 @@ void AssignRolesToIdleUnits(Actor self)
...
+		{
+			if (!Info.CapturingActorTypes.Any() || Player.WinState != WinState.Undefined)
+				return;
+
+			var capturers = unitsHangingAroundTheBase.Where(a => a.IsIdle && Info.CapturingActorTypes.Contains(a.Info.Name)).ToArray();
+			if (capturers.Length == 0)
+				return;
+
+			if (combatantPlayers == null)
+				combatantPlayers = World.Players.Where(p => p.InternalName != "Everyone" &&
```
Mailaender:
```csharp
World.Players.Where(p => !p.NonCombatant)
```

mods/ra/rules/ai.yaml
```diff
@@ -42,7 +42,12 @@ Player:
 			stek: 1%
 			fix: 0.1%
 			dome: 10%
+		CapturingActorTypes: e6
+		CapturableActorTypes: mslo, gap, iron, pdox, atek, weap, fact, proc, silo, hpad, afld, powr, apwr, stek, barr, kenn, tent, fix
```
Mailaender: You could add `oilb` here and maybe have it prioritize neutral unguarded oil derricks.

mods/cnc/rules/ai.yaml
```diff
@@ -46,7 +46,12 @@ Player:
 			silo: 0%
 			fix: 1%
 			hpad: 2%
+		CapturingActorTypes: e6
+		CapturableActorTypes: fact, fact.gdi, fact.nod, nuke, nuk2, proc, pyle, hand, afld, weap, hpad, hq, fix, eye, tmpl, obli, cycl, brik
```
Mailaender: Adding `v19` here probably doesn't hurt.

OpenRA.Mods.Common/AI/HackyAI.cs
Phrohdoh: Not quite. I don't want `Everyone` nor do I want allies.

mods/cnc/rules/ai.yaml
GraionDilach: Remove brik as well. Since when Engineers capture walls?

Mailaender: Oh well spotted: `cycl` as well then.

GraionDilach:
This can fix #4561 with the nits addressed.

Phrohdoh:
This has been updated.

abcdefg30:
It seems like you made all AIs in ra build `e6`s, although only `Rush AI` utilizes them?

OpenRA.Mods.Common/AI/HackyAI.cs
abcdefg30: This still needs to be resolved.

Phrohdoh: There is nothing to resolve here as far as I see.

obrakmann: There is. This should look like this:
```csharp
if (combatantPlayers == null)
	combatantPlayers = World.Players.Where(p => !p.Spectating &&
		(p.Stances[Player] == Stance.Enemy || p.Stances[Player] == Stance.Neutral)).ToArray();
```

Phrohdoh: Why should they explicitly avoid specs?

obrakmann: 1. Because you won't have to hardcode the "Everyone" user, which is an implementation detail this code shouldn't need to know about.
2. Why do you even want the AI to target specs?

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -630,6 +671,98 @@ void AssignRolesToIdleUnits(Actor self)
...
+
+			QueueOrder(new Order(target.OrderString, capturer, true) { TargetActor = target.Actor });
+			BotDebug("AI ({0}): Ordered {1} to capture {2}", Player.ClientIndex, capturer, target.Actor);
+			activeUnits.Remove(capturer);
+		}
+
+		CaptureTarget<TTargetType> GetCapturerTargetClosestToOrDefault<TTargetType>(Actor capturer, IEnumerable<CaptureTarget<TTargetType>> targets)
+			where TTargetType : class, ITraitInfoInterface
+		{
+			return targets.OrderByDescending(target => (target.Actor.CenterPosition - capturer.CenterPosition).LengthSquared).LastOrDefault();
```
RoosterDragon: `.OrderByDescending(...).LastOrDefault()` -> `.MinByOrDefault(...)`

Phrohdoh:
Updated:
* Player filtering removed
* Only add new properties to RA's Rush AI

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -631,6 +670,94 @@ void AssignRolesToIdleUnits(Actor self)
 			}
 
 			FindAndDeployBackupMcv(self);
+
+			if (--minCaptureDelayTicks <= 0)
+			{
+				minCaptureDelayTicks = Info.MinimumCaptureDelay;
+				QueueCaptureOrders();
+			}
```
RoosterDragon: Just noticed a long-standing bug here, a bit higher up we do:
```c#
if (--assignRolesTicks > 0)
    return;
```
which means we will only decrement `minAttackForceDelayTicks` and now `minCaptureDelayTicks` much less frequently than we expect!

We should address this either here, or in another PR that you'll need to rebase onto.

Phrohdoh: Let's have another PR then that isn't related to capturing.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -631,6 +670,94 @@ void AssignRolesToIdleUnits(Actor self)
...
+		void QueueCaptureOrders()
+		{
+			if (!Info.CapturingActorTypes.Any() || Player.WinState != WinState.Undefined)
+				return;
+
+			var capturers = unitsHangingAroundTheBase.Where(a => a.IsIdle && Info.CapturingActorTypes.Contains(a.Info.Name)).ToArray();
+			if (capturers.Length == 0)
+				return;
+
+			var randPlayer = World.Players.Random(Random);
```
Phrohdoh: I didn't build out an agro system for capturing (because that isn't logical IMO) so this will cause enemies and neutral  (and allies, but the capturing code may filter that out) to be targeted without bias.

GraionDilach: Except that I don't see allies being filtered out. I'd guess if it picks an ally the whole attempt gets wasted.

Phrohdoh: I am not filtering out allies, the capturing code would do that if desired (as I stated above).

obrakmann: Which would mean that you'd throw away the AI's capturing attempt and have to wait until the timeout expires again. This is an especially bad decision if you consider that in a two-sided match, less than 50% of players would be eligible for capturing. No wonder it's not working properly anymore in the current revision.

Pleas just implement the change that I requested in https://github.com/OpenRA/OpenRA/pull/11417#discussion_r71756442. I'll remove my :+1: until that happens.

Phrohdoh: @obrakmann Have you actually reread the code?
That portion doesn't exist anymore.

obrakmann: Yes, I have. The same way you did it in a previous iteration, only that you fix the query the way I suggested.

GraionDilach: ` World.Players.Where(p => !p.Spectating && (p.Stances[Player] == Stance.Enemy || p.Stances[Player] == Stance.Neutral)).Random(Random)` incase you're cba to look it back.

Phrohdoh:
This has been updated.

Target player selection is now done like so:
```csharp
    var randPlayer = World.Players.Where(p => !p.Spectating
        && Info.CapturableStances.HasStance(Player.Stances[p])).Random(Random);
```

obrakmann:
Thanks, :+1:

GraionDilach:
Looks good to me as well. :+1:

WolfGaming:
:+1: Looks good to me except sometimes when there is impassable terrain between the engineers and their targets the ai just stops trying to capture and leaves the engineers standing.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/dee5d6aa90e922cf471ab6d2648ee4761d311aec)

Phrohdoh:
🎊

PR #11421, opened 2016-06-10
Fixed the AppVeyor build

Mailaender:
Looks like the https://chocolatey.org/packages/pandoc changed towards a more sane install.

```
pandoc has been installed.
pandoc installed to 'C:\Program Files (x86)\Pandoc'
Adding pandoc to the PATH if needed
```

Adjusting our build scripts to fix the systematically failing builds.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -54,13 +54,13 @@ test_script:
 
 after_test:
   - choco install pandoc -y
-  - '%LOCALAPPDATA%\Pandoc\pandoc.exe -o README.html README.md'
-  - '%LOCALAPPDATA%\Pandoc\pandoc.exe -o CONTRIBUTING.html CONTRIBUTING.md'
+  - '"%ProgramFiles(x86)%\Pandoc\pandoc.exe" -o README.html README.md'
+  - '"%ProgramFiles(x86)%\Pandoc\pandoc.exe" -o CONTRIBUTING.html CONTRIBUTING.md'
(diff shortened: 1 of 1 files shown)


GraionDilach:
Can't we just add an `ALLUSERS=0` to line 56? From the linked report you mention, this might be AppVeyor preventing the path be updated.

GraionDilach:
Also, see my comment in the upstream issue.

Mailaender:
> Can't we just add an `ALLUSERS=0` to line 56? From the linked report you mention, this might be AppVeyor preventing the path be updated.

Sadly not.

Mailaender:
Finally fixed the build.

GraionDilach:
:+1:

Let's hope the next update won't break it again.

appveyor.yml
```diff
@@ -54,13 +54,13 @@ test_script:
 
 after_test:
   - choco install pandoc -y
-  - '%LOCALAPPDATA%\Pandoc\pandoc.exe -o README.html README.md'
-  - '%LOCALAPPDATA%\Pandoc\pandoc.exe -o CONTRIBUTING.html CONTRIBUTING.md'
+  - '"%ProgramFiles(x86)%\Pandoc\pandoc.exe" -o README.html README.md'
```
pchote: The outer `'` look unnecessary now.

Mailaender: Let's give it a try without them.

Mailaender: > Error parsing appveyor.yml: (Line: 57, Col: 5, Idx: 1635) - (Line: 57, Col: 32, Idx: 1662): While parsing a quoted scalar, find unknown escape character.

pchote:
👍

PR #11422, opened 2016-06-10
Expose TeslaZap duration to yaml.

GraionDilach:


reaperrr:
An idea (not necessarily for this PR, but maybe a follow-up):
Replace the `doneDamage` boolean with a `DamageDuration` property, so modders can make the projectile deal damage over multiple ticks as well.

GraionDilach:
IIRC LaserZap doesn't have that feature either so a followup where both gets it at the same time would make more sense.

obrakmann:
:+1:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/b0a5f1c580e404e157053cf548d7541acef5b41a...5a025816b473eeccf823b0e806343b6cde24930e)

PR #11423, opened 2016-06-10
Add a maximum limit to InfiltrateForCash.

GraionDilach:
**This does not affect the default mods.**

My reason behind this would be "because Ares (http://ares-developers.github.io/Ares-docs/new/buildings/spybehavior.html#money)" and because modders might not want to allow spies to have a chance of granting more than, say ten times the spy's own value even from an unlucky hoarder as a balance choice.

### DIFF
--- a/OpenRA.Mods.RA/Traits/Infiltration/InfiltrateForCash.cs
+++ b/OpenRA.Mods.RA/Traits/Infiltration/InfiltrateForCash.cs
@@ -26,6 +26,9 @@ namespace OpenRA.Mods.RA.Traits
 			"When negative, the production price of the infiltrating actor will be used instead.")]
 		public readonly int Minimum = -1;
 
+		[Desc("Maximum amount of funds which will be stolen.")]
+		public readonly int Maximum = int.MaxValue;
+
 		[Desc("Sound the victim will hear when they get robbed.")]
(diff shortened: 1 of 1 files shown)


obrakmann:
:+1:

OpenRA.Mods.RA/Traits/Infiltration/InfiltrateForCash.cs
```diff
@@ -26,6 +26,9 @@ class InfiltrateForCashInfo : ITraitInfo
 			"When negative, the production price of the infiltrating actor will be used instead.")]
 		public readonly int Minimum = -1;
 
+		[Desc("Amount of maximum funds which will be stolen.")]
```
Phrohdoh: `Maximum amount of funds` reads better.

GraionDilach: Updated.

Phrohdoh:
👍

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f009259e41a5b4d97314fcea9dcc8981160c6fc6...b0a5f1c580e404e157053cf548d7541acef5b41a)

PR #11428, opened 2016-06-10
Allow attack anims/overlays to start before weapons fire

### LINKED ISSUE #11331: FireDelay and animation
In RA, despite the grenadier having a value for `FireDelay`, the start of the firing animation begins when the projectile is launched. This makes the grenadier throwing animation play a noticeable amount of frames after the projectile is already in the air. 

Also, moving immediately after the attack command is given the unit can be in a different position to where the weapon spawns. This is more noticeable as the value is increased.

On bleed.

reaperrr:
This allows attack animations/overlays to ignore `Armament.FireDelay` and start right away, which in turn allows fixing Grenadier throw anim timing.

Closes #11331.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Infantry/ScaredyCat.cs
+++ b/OpenRA.Mods.Common/Traits/Infantry/ScaredyCat.cs
@@ -82,11 +82,13 @@ namespace OpenRA.Mods.Common.Traits
 				Panic();
 		}
 
-		public void Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
 		{
 			if (self.World.SharedRandom.Next(100 / info.AttackPanicChance) == 0)
 				Panic();
 		}
+
+		void INotifyAttack.PreparingAttack(Actor self, Target target, Armament a, Barrel barrel) { }
 
 		public int GetSpeedModifier()
 		{
--- a/OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
@@ -47,13 +47,15 @@ namespace OpenRA.Mods.Common.Traits.Render
 			wsb = init.Self.Trait<WithSpriteBody>();
 		}
 
-		public void Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel) { }
+
+		void INotifyAttack.PreparingAttack(Actor self, Target target, Armament a, Barrel barrel)
 		{
 			if (!string.IsNullOrEmpty(info.AttackSequence))
 				wsb.PlayCustomAnimation(self, info.AttackSequence, () => wsb.CancelCustomAnimation(self));
 		}
 
-		public void Tick(Actor self)
+		void ITick.Tick(Actor self)
 		{
 			if (string.IsNullOrEmpty(info.AimSequence) && string.IsNullOrEmpty(info.ReloadPrefix))
 				return;
--- a/OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
@@ -28,6 +28,12 @@ namespace OpenRA.Mods.Common.Traits.Render
 		[Desc("Shown while reloading.")]
 		[SequenceReference(null, true)] public readonly string ReloadPrefix = null;
 
+		[Desc("Delay in ticks before animation starts, either relative to attack preparation or attack.")]
+		public readonly int Delay = 0;
+
+		[Desc("Should the animation be delayed relative to preparation or actual attack?")]
+		public readonly AttackDelayType DelayRelativeTo = AttackDelayType.Preparation;
+
 		public object Create(ActorInitializer init) { return new WithAttackAnimation(init, this); }
 	}
 
@@ -38,6 +44,8 @@ namespace OpenRA.Mods.Common.Traits.Render
 		readonly Armament armament;
 		readonly WithSpriteBody wsb;
 
+		int tick;
+
 		public WithAttackAnimation(ActorInitializer init, WithAttackAnimationInfo info)
 		{
 			this.info = info;
@@ -47,16 +55,39 @@ namespace OpenRA.Mods.Common.Traits.Render
 			wsb = init.Self.Trait<WithSpriteBody>();
 		}
 
-		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel) { }
-
-		void INotifyAttack.PreparingAttack(Actor self, Target target, Armament a, Barrel barrel)
+		void PlayAttackAnimation(Actor self)
 		{
 			if (!string.IsNullOrEmpty(info.AttackSequence))
 				wsb.PlayCustomAnimation(self, info.AttackSequence, () => wsb.CancelCustomAnimation(self));
 		}
 
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			if (info.DelayRelativeTo == AttackDelayType.Attack)
+			{
+				if (info.Delay > 0)
+					tick = info.Delay;
+				else
+					PlayAttackAnimation(self);
+			}
+		}
+
+		void INotifyAttack.PreparingAttack(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			if (info.DelayRelativeTo == AttackDelayType.Preparation)
+			{
+				if (info.Delay > 0)
+					tick = info.Delay;
+				else
+					PlayAttackAnimation(self);
+			}
+		}
+
 		void ITick.Tick(Actor self)
 		{
+			if (info.Delay > 0 && --tick == 0)
+				PlayAttackAnimation(self);
+
 			if (string.IsNullOrEmpty(info.AimSequence) && string.IsNullOrEmpty(info.ReloadPrefix))
 				return;
 
--- a/OpenRA.Mods.Common/Traits/Render/WithAttackOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithAttackOverlay.cs
@@ -27,16 +27,23 @@ namespace OpenRA.Mods.Common.Traits.Render
 		[Desc("Custom palette is a player palette BaseName")]
 		public readonly bool IsPlayerPalette = false;
 
+		[Desc("Delay in ticks before overlay starts, either relative to attack preparation or attack.")]
+		public readonly int Delay = 0;
+
+		[Desc("Should the overlay be delayed relative to preparation or actual attack?")]
+		public readonly AttackDelayType DelayRelativeTo = AttackDelayType.Preparation;
+
 		public object Create(ActorInitializer init) { return new WithAttackOverlay(init, this); }
 	}
 
-	public class WithAttackOverlay : INotifyAttack
+	public class WithAttackOverlay : INotifyAttack, ITick
 	{
 		readonly Animation overlay;
 		readonly RenderSprites renderSprites;
 		readonly WithAttackOverlayInfo info;
 
 		bool attacking;
+		int tick;
 
 		public WithAttackOverlay(ActorInitializer init, WithAttackOverlayInfo info)
 		{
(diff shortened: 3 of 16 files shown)


OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -35,6 +35,7 @@ public interface IPlaceBuildingDecorationInfo : ITraitInfo
 	}
 
 	public interface INotifyAttack { void Attacking(Actor self, Target target, Armament a, Barrel barrel); }
+	public interface INotifyAttackAnimation { void AttackingAnimation(Actor self, Target target, Armament a, Barrel barrel); }
```
pchote: Can we get a `[RequireExplicitImplementation]` here?

reaperrr: Done. That caught the implicit implementation in `WithInfantryBody` as well.

reaperrr:
Updated.

OpenRA.Mods.Common/Traits/Sound/WithAttackAnimationSounds.cs
```diff
@@ -0,0 +1,42 @@
...
+using OpenRA.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits.Render
+{
+	[Desc("Played together with an attack animation or overlay.")]
+	public class WithAttackAnimationSoundsInfo : ITraitInfo
+	{
+		[Desc("Play a randomly selected sound from this list along with any attack animations or overlays.")]
+		public readonly string[] Sounds = null;
```
Phrohdoh: Make it empty, not null.

reaperrr:
Updated.

reaperrr:
Updated.

Replaced the `AttackAnimsIgnoreFireDelay` boolean with a more straightforward and flexible `AttackAnimationDelay` that defaults to `FireDelay`.
Although it's unlikely to see much use, you may never know when some 3rd-party mod wants to delay both, but not to the same tick, and this way we can avoid having to write upgrade rules and upgrade our mods later on.

pchote:
Rewording my thoughts from IRC to be a bit clearer:

I'm thinking we could make the abstraction a bit cleaner by merging `INotifyAttackAnimation` into `INotifyAttack` by defining a `INotifyAttack.PreparingAttack` method that is called immediately when preparaing the armament.  The `WithAttackAnimation` trait can then have its own properties for defining the delay and an enum to say whether that delay is relative to preparing or the actual attack.  Likewise for the attack sounds

This brings two advantages:
1) Armaments don't have to concern themselves with animation-specific code.
2) Different downstream traits (e.g. animations vs sounds) can use different delays.

reaperrr:
Updated, done as suggested.

OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
```diff
@@ -47,13 +54,25 @@ public WithAttackAnimation(ActorInitializer init, WithAttackAnimationInfo info)
 			wsb = init.Self.Trait<WithSpriteBody>();
 		}
 
-		public void Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			if (info.DelayRelativeTo == AttackDelayType.Attack && !string.IsNullOrEmpty(info.AttackSequence))
+				self.World.AddFrameEndTask(w => w.Add(new DelayedAction(info.Delay, () =>
+					{
```
abcdefg30: One level of indentation too much.

reaperrr: Hm, are you sure? the brackets belong to the `DelayedAction`, not the `AddFrameEndTask`.

pchote: Cases like this are why we have a(n often ignored) rule to avoid hanging blocks.  When you have an inner block with braces all outer blocks should also have braces, so please add them here.

reaperrr:
Updated, added brackets for better readability.

OpenRA.Mods.Common/Traits/Sound/PreparingAttackSounds.cs
```diff
@@ -0,0 +1,45 @@
...
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits.Render
```
obrakmann: That doesn't look like the right place.

reaperrr: Ouch, well spotted.

obrakmann:
This removes the burst-y firing behaviour from the obelisk. Are people cool with that?

obrakmann:
Nevermind, seems that I misremembered the obelisk firing in two-shot bursts. :+1: then, after the namespace fixup.

reaperrr:
Fixed.

OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
```diff
@@ -57,19 +57,29 @@ public WithAttackAnimation(ActorInitializer init, WithAttackAnimationInfo info)
 		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
 		{
 			if (info.DelayRelativeTo == AttackDelayType.Attack && !string.IsNullOrEmpty(info.AttackSequence))
-				self.World.AddFrameEndTask(w => w.Add(new DelayedAction(info.Delay, () =>
+			{
+				self.World.AddFrameEndTask(w =>
+				{
+					w.Add(new DelayedAction(info.Delay, () =>
 					{
```
Phrohdoh: These brackets actually aren't necessary, this is a single instruction.

OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
```diff
@@ -57,19 +57,29 @@ public WithAttackAnimation(ActorInitializer init, WithAttackAnimationInfo info)
...
 
 		void INotifyAttack.PreparingAttack(Actor self, Target target, Armament a, Barrel barrel)
 		{
 			if (info.DelayRelativeTo == AttackDelayType.Preparation && !string.IsNullOrEmpty(info.AttackSequence))
-				self.World.AddFrameEndTask(w => w.Add(new DelayedAction(info.Delay, () =>
+			{
+				self.World.AddFrameEndTask(w =>
+				{
+					w.Add(new DelayedAction(info.Delay, () =>
 					{
```
Phrohdoh: Same here.

OpenRA.Mods.Common/Traits/Sound/PreparingAttackSounds.cs
```diff
@@ -0,0 +1,45 @@
...
+			: base(info)
+		{
+			this.info = info;
+		}
+
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel) { }
+
+		void INotifyAttack.PreparingAttack(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			if (info.Sounds != null && info.Sounds.Any())
```
Phrohdoh: `info.Sounds` won't be null.

reaperrr:
Updated.

mods/cnc/rules/structures.yaml
```diff
@@ -757,15 +757,15 @@ OBLI:
...
+		Sounds: obelpowr.aud
 	Armament:
 		Weapon: Laser
 		LocalOffset: 0,-85,1280
-		FireDelay: 0
-	AttackCharge:
-		ChargeAudio: obelpowr.aud
-		ReloadDelay: 40
-		InitialChargeDelay: 50
+		FireDelay: 40
```
pchote: @GraionDilach pointed out in IRC that this makes the actor ignore the maximum range check, and fire unconditionally on the target.  `AttackCharge` (actually `ChargeFire`) has a specific extra range check that avoids that.

pchote: Fast units can move quite a way (4-5 cells) past the max range during the charge animation, so this is a pretty bad regression.

reaperrr: > so this is a pretty bad regression.

I agree, but I'd like to point out that the underlying problem is (or was, as I have a fix for it) that `Armament` schedules the delayed action almost unconditionally, a larger `FireDelay` for other weapons would have the same issue, ignoring the minimum range as well.

pchote: Indeed, but in most (all?) other existing cases the `FireDelay`s are small enough for this to be much less of an issue.

reaperrr:
Updated.

Projectiles are now only launched if the target didn't move out of max range or inside min range during the `FireDelay`.

pchote:
I'm not convinced that's a good approach.  We've already committed to firing the armament by this point, so I think it would be better to change the target position to the ground at the edge of the max range.  Obelisk is a special case that is still best handled by its own targeting trait (ref the recent discussions in IRC).

Nelax:
I know at least what I plan to use this logic for, a turreted charging laser weapon, that would not be preferable but I could see it being the preferred way of firing for any normal projectile weapon.

pchote:
Can we, for now, drop f6bca6c and the obelisk and tesla changes?

This is otherwise a straightforward and uncontroversial plumbing improvement that doesn't need to be blocked by these other issues.

reaperrr:
> Can we, for now, drop f6bca6c and the obelisk and tesla changes?

Okay, done.

Phrohdoh:
59816017730d49697dac5d4c7123db07d4be8fd9 should be squashed into whichever commit introduces that code.

OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
```diff
@@ -47,13 +54,35 @@ public WithAttackAnimation(ActorInitializer init, WithAttackAnimationInfo info)
 			wsb = init.Self.Trait<WithSpriteBody>();
 		}
 
-		public void Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			if (info.DelayRelativeTo == AttackDelayType.Attack && !string.IsNullOrEmpty(info.AttackSequence))
+			{
+				self.World.AddFrameEndTask(w =>
```
pchote: Can you please pull out a `void PlayAnimation` method to avoid this duplication?

OpenRA.Mods.Common/Traits/Render/WithAttackOverlay.cs
```diff
@@ -52,10 +57,34 @@ public WithAttackOverlay(ActorInitializer init, WithAttackOverlayInfo info)
 				info.Palette, info.IsPlayerPalette);
 		}
 
-		public void Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			if (info.DelayRelativeTo == AttackDelayType.Attack)
+			{
+				self.World.AddFrameEndTask(w =>
```
pchote: Same here.

OpenRA.Mods.Common/Traits/Sound/PreparingAttackSounds.cs
```diff
@@ -0,0 +1,71 @@
...
+			: base(info)
+		{
+			this.info = info;
+		}
+
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			if (info.DelayRelativeTo == AttackDelayType.Attack && info.Sounds.Any())
+			{
+				self.World.AddFrameEndTask(w =>
```
pchote: Same here.

OpenRA.Mods.Common/TraitsInterfaces.cs
```diff
@@ -34,7 +36,12 @@ public interface IPlaceBuildingDecorationInfo : ITraitInfo
 		IEnumerable<IRenderable> Render(WorldRenderer wr, World w, ActorInfo ai, WPos centerPosition);
 	}
 
-	public interface INotifyAttack { void Attacking(Actor self, Target target, Armament a, Barrel barrel); }
+	public interface INotifyAttack
```
pchote: Can you please add `[RequireExplicitImplementation]` to this?

OpenRA.Mods.Common/Traits/Attack/AttackOmni.cs
```diff
@@ -29,6 +29,14 @@ public override Activity GetAttackActivity(Actor self, Target newTarget, bool al
 			return new SetTarget(this, newTarget);
 		}
 
+		public override void ResolveOrder(Actor self, Order order)
```
pchote: What's the motivation for moving this from `AttackCharge` to `AttackOmni` (why not `AttackBase`)?

OpenRA.Mods.Common/Traits/Infantry/ScaredyCat.cs
```diff
@@ -82,12 +82,14 @@ public void Damaged(Actor self, AttackInfo e)
 				Panic();
 		}
 
-		public void Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
```
pchote: Does anyone know what the purpose of this is?  It was introduced in 11c45e43884f57de6daa2a38543eb7670a0c99e4 but doesn't appear to be used at all.

OpenRA.Mods.Common/Traits/Attack/AttackOmni.cs
reaperrr: I just wasn't sure if there's a reason why this was not part of `AttackBase` and didn't want to take any risks, but I guess it can't really hurt to move it to `AttackBase`.

OpenRA.Mods.Common/Traits/Infantry/ScaredyCat.cs
reaperrr: The idea is (or was) that a scared actor would have a certain probability of panicking when attacking, I guess (in the original games, civilians that were shot at would run around in panic between attacks even while attacking a target). Not sure if that still works, but I consider that out of scope for this PR either way.

reaperrr:
Updated.

Everything should work now, but I guess this changed too much since @obrakmann reviewed it so I'm removing the Needs +2.

obrakmann:
still looks fine to me, :+1:

reaperrr:
Rebased so reviewers won't have to switch dependencies between this and newer PRs.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/579de74a8ace61b9dbe2fcabcfadc62349f5df7f...36162753843e86ccd04d9faebe93e46b05c9ef2d)

Issue #11429, opened 2016-06-10
[RA] Rocket soldiers damage own pillbox


r34ch:
Placing a rocket soldier in a pillbox and alternating fire either side of the pill box will cause the rockets to damage the pillbox.

On bleed.

FrameLimiter:
This can be fixed by giving the E3 a garrisoned weapon that has a horizontal rate of turn of 255:

E3:
+Armament@GARRISONED:
+Name: garrisoned
+Weapon: Dragon.garrisoned

Dragon.garrisoned
+Inherits: Dragon
+Projectile: Missile
++HorizontalRateOfTurn: 255

GraionDilach:
This is just another bug of AttackGarrisoned and it's inability of interleaving the garrisoned actor's properties with the garrison's.

PR #11431, opened 2016-06-11
[RA] Rebalance camo pillbox as country specific building

ghost:
Another balance suggestion from the [community](http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&p=296162#296162) that was well received.

Test map available here - http://resource.openra.net/maps/14515/

This PR gives the camo pillbox to France.
- It costs the same as a normal pillbox
- It is available at the same tech level as the normal pillbox
- It has an extra 200hp (same health as current camo PB)
- It has cloaking obviously
- It has been moved in the sidebar so that defense hotkeys are consistent across allied factions

France's early play-style is rather deceptive and defensive due to their fake buildings. Static defences are still a big investment early game (and can be avoided) so I think camo PBs should be fairly strong early game. Currently players can hold fire their camo pillboxes to lure early infantry rushes closer for maximum kill chance. I wanted to build on this with the very under utilized infantry garrison mechanism.

$400 camo PB (with free $100 rifle) + $300 RS = $600 turret + $100 rifle. 

With the cheaper price and lower tech level I hope to see some interesting RS garrison plays (once #11429 is fixed) to counter early LT and APC blocking strats safely. Although camo PBs still have 200hp more than turrets, turrets fire much faster and have better range so they won't be replaced by camo RS pillboxes.

~~The original pillbox for France still exists unchanged as I felt any changes would make them more likely to be spammed. I left it in to give French players more options if your opponent doesn't know you are playing France. A live decoy as it were.~~

### DIFF
--- a/mods/ra/rules/structures.yaml
+++ b/mods/ra/rules/structures.yaml
@@ -586,8 +586,8 @@ HBOX:
 	Building:
 	Buildable:
 		Queue: Defense
-		BuildPaletteOrder: 50
-		Prerequisites: tent, ~structures.allies, ~techlevel.medium
+		BuildPaletteOrder: 150
+		Prerequisites: tent, ~structures.france, ~techlevel.low
 	Valued:
 		Cost: 600
 	Health:
--- a/mods/ra/rules/world.yaml
+++ b/mods/ra/rules/world.yaml
@@ -25,7 +25,7 @@
 		Name: France
 		InternalName: france
 		Side: Allies
-		Description: France: Deception\nSpecial Ability: Can build fake structures\nSpecial Unit: Mobile Gap Generator
+		Description: France: Deception\nSpecial Ability: Can build fake structures\nSpecial Defence: Camo Pillbox\nSpecial Unit: Mobile Gap Generator
 	Faction@3:
 		Name: Germany
 		InternalName: germany


cjshmyr:
> The original pillbox for France still exists unchanged as I felt any changes would make them more likely to be spammed. I left it in to give French players more options if your opponent doesn't know you are playing France. A live decoy as it were.

I don't think this is valid, if someone wants to see the country they can look at the tooltip flag. Should just remove it from France since the camo pbox is a direct upgrade.

pchote:
I agree with @cjshmyr: the regular pillbox is inferior in every way, so it doesn't make a lot of sense to keep it as an option.

pchote:
Can you please update the country descriptions in the lobby to include this?

r34ch:
@cjshmyr My mistake, I will remove it. I tried `-~structures.france` as well as `~structures.germany, ~structures.england` but neither worked. Any suggestions? `~structures.france.germany` doesn't work either

@pchote How does "Special Defence: Camo Pillbox" sound?

cjshmyr:
`Prerequisites: tent, ~structures.allies, ~!structures.france, ~techlevel.low` should work on pillbox. The `~!` removes if structures.france is met.

I like that description you suggested as well.

r34ch:
Thanks hamb, updated.

cjshmyr:
With singleplayer missions if both are available, we run into the same case where the regular isn't worth building. I'm not sure what to suggest at this point, other than bumping the price back up. Thoughts?

Micr0Bit:
:+1: for this , and i also agree with @cjshmyr removing the regular pillbox from france

r34ch:
@cjshmyr Very valid point about single player. I was going to suggest a range or RoF adjustment, but I don't think its currently possible to modify a garrisoned units weapon characteristics depending on the structure garrisoned?

I can bump the price up to $600 again but then its no longer a direct upgrade and I think players would be annoyed at not having the option to choose the cheaper option if the original pillbox stayed removed.

cjshmyr:
I agree, it's simplest to re-add the regular pillbox back to France, and keeping the camo exclusive to them.

GraionDilach:
@r34ch: It is possible, albeit only through weapon cloning atm. Thanks for reminding to get that one working, it's on my tasklist.

r34ch:
Ok I've bumped up the price for now. 

When the feature of changing garrisoned weapon characteristics depending on the structure is implemented, I would like to revisit this and implement the camo pillbox more as a side-grade so that France can do without the original pillbox.

obrakmann:
> With singleplayer missions if both are available, we run into the same case where the regular isn't worth building. I'm not sure what to suggest at this point, other than bumping the price back up. Thoughts?

Just add the old rules to the `campaign-rules.yaml` file.

cjshmyr:
I like the simpler move to france vs campaign rules, since a $400 camo pillbox sounds too strong? Some players already prefer the $600 camo pillbox

r34ch:
Yea that's fair, it would be better to play test this change along with the fake SD first.

My thinking was that France has no offensive special weaponry. A $400 pillbox counters all early game threats just as well as a $600 camo pillbox except for blobs. If you invest in fakes to counter early scouting thats less money for defense in general.

Personally I find them more useful late game to lure tanya and paradrops but a visible pillbox is just as good as a deterrent. By mid and late game you have to deal with armor so I feel their usefulness becomes reduced at that point. The meta is also slowly shifting to mobile armies on wider open maps which further reduces the effectiveness of static deferences. I thought a reduction in price might see them scattered about bases and expansions more which would help slow pushes against someone playing France.

A different change I toyed with which seemed ok was buffing the garrisoned RS RoF from 50 -> 40. That gives you an option to use camo pillboxes for effective anti armour as well. It would buff pillboxes too but their lower HP doesn't begin to rival a turret.

SoScared:
I hate to say this _now_ but I've had second thoughts on the France exclusive c.pillbox for some time now. It doesn't feel right to alter the options for all Allies simply in order to help France diversify. There was a big surge of enthusiasm on the idea in midst of misc. playtesting that went all over the place (not to say it was bad). I understand how the c.pillbox ties into France as a deception faction but I think the gameplay would benefit more from giving the Allies' factions a real option between the two pillboxes instead, altering the price and HP of the camo. pillbox so that the regular doesn't become redundant. The c.pillbox is crucial for Allies on defending its' tech and supports creative tactics that fit's moments during gameplay, _not_ as an overall pre-game strategy.

When we tested out the exclusive French camo pillbox the price was reduced to $500 and HP down to 450 so that France wouldn't be stuck with the $600 price tag and to adress the alledgedly OP 600 HP. In the scenario where you have both options, I think $600 for 450HP could be enough to have players decide between the camo. and the regular throughout a game.

r34ch:
It would seem there is no agreement on how to balance this properly for both multiplayer and singleplayer then?

@SoScared I have to completely disagree. I don't think a camo pillbox is crucial for defending late game tech any more than properly used normal pillboxes. I will however concede the point I made about focusing on early game which is why the current state of the pull request is simply keeping the camo pillbox unchanged, apart from it being Faction specific.

Currently choosing a faction as Allied is gimmicky compared to the Soviets. Anti infantry for Ukraine is completely different to the shock balls of Russia. The Soviet factions are suited to players with slightly different playstyles in the same way as the Allied and Soviet sides are suited for players with very different playstyles.

Allies have so many underused and overlapping units that I never feel that any one Allied faction actually lends itself to **any** particular playstyle. I think it would be good to try and actively encourage that like has been done with the Soviet factions.

If players prefer to aggressively expand and use defenses for expansions then France _should_ be a good choice. If players prefer more microing and harassment then England _should_ be a good choice. If players like aggressively attacking then Germany _should_ be a good choice.

SoScared:
Is it fair to say there's a consensus on the value of c.pillbox vs regular being disproportionate? 

The idea of a Faction specific c.pillbox spawned out of the discussion of the regular pillbox being redundant. At the same time there was a discussion on about the c.pillbox being a little bit too beefy with it's 600HP together with it's cloak ability. Nerfing the c.pillbox HP (e.g. from 600 to 450) could solve a couple of issues right off the bat. E.g. if you're cheap, broke or in need of some emergency pillboxes then the regular pillbox would stand out as the obvious choice.

r34ch:
@SoScared would you mind clarifying a little bit? this part threw me

>  Nerfing the c.pillbox HP (e.g. from 600 to 450) could solve a couple of issues right off the bat. E.g. if you're cheap, broke or in need of some emergency pillboxes then the regular pillbox would stand out as the obvious choice.

As it stands the current pillbox is already a better choice if you are low on cash (and build time). I think the HP decrease is a fair point but I don't see how the HP change would influence peoples decisions if money was the issue?

SoScared:
Pillbox - $400 (10s) - 400HP
C.Pillbox - $600 (15s) - 600HP (current)

C.Pillbox - $600 (15s) - 450HP (suggested)


Currently we're getting exactly the same $ value in terms of durability, c.pillbox vs regular pillbox. On top of that the c.pillbox is cloaked. This is the main reason the regular pillbox is getting phased out. Even when players are broke they favor the c.pillbox.

With a 450HP c.pillbox the choice would no longer be obvious. You'd be deciding between cloak or durability for your buck.

r34ch:
@SoScared ah I understand you now, I think thats a fair proposal and makes sense. 

Although I get the argument that the camo + hp combo is currently too attractive for the price, it does seem a shame to allow the Allies to keep a variation of the same minor defense (even rebalanced) since it does seem to fit nicely with France's theme.

PR #11434, opened 2016-06-11
Replace WithVoxelWaterBody with UpgradeOnTerrain.

GraionDilach:
Didn't write an upgrade rule because the TS DLL isn't shipped, though.


reaperrr:
Nit: If you switch the commit order, the first commit won't make TS crash. Though I guess it doesn't really matter.

GraionDilach:
Nit addressed.

abcdefg30:

PR #11441, opened 2016-06-12
Added static Tiberian Sun low bridges

Mailaender:
A slightly improved and rebased version of https://github.com/OpenRA/OpenRA/pull/10764 which now imports dead bridges as dead actors that are rendered transparent in the map editor.


pchote:
This doesn't even try to address the main issue that makes #10764 unacceptable. I've gone to great lengths to explain why that approach is no good, and you even appeared to understand it.  I don't understand why you've reopened virtually the same thing again.

:-1: because, as discussed earlier, all of these changes would need to be reverted and rewritten from scratch by the person who wants to implement bridge damage and repair.

Mailaender:
>  all of these changes would need to be reverted and rewritten from scratch by the person who wants to implement bridge damage and repair.

Actually not if this get written with bridge damage and repair done via actor replacement.

pchote:
I have already explained why that is a horrible solution that I consider to be completely unacceptable.

pchote:
Please replace the individual overlay types (remember that TS did not use actors for bridges) with bridge actors that define proper damage states.

OpenRA.Mods.TS/Traits/Buildings/LowBridge.cs
```diff
@@ -0,0 +1,55 @@
...
+		void INotifyCreated.Created(Actor self)
+		{
+			if (self.IsDead)
+				return;
+
+			var buildingInfo = self.Info.TraitInfo<BuildingInfo>();
+			var cells = FootprintUtils.PathableTiles(self.Info.Name, buildingInfo, self.Location);
+			var tileSet = self.World.Map.Rules.TileSet;
+			foreach (var cell in cells)
+				self.World.Map.CustomTerrain[cell] = tileSet.GetTerrainIndex(info.TerrainType);
```
abcdefg30: Couldn't you move the `tileSet.GetTerrainIndex(info.TerrainType);` out of the loop?
E.g.
```
var terrainIndex = tileSet.GetTerrainIndex(info.TerrainType);
foreach (var cell in cells)
	self.World.Map.CustomTerrain[cell] = terrainIndex;
```

Mailaender:
I already tried that https://github.com/Mailaender/OpenRA/commit/d21988f4fc102dc5d2f1f88f2ce7da754441e632. It ends in a rubbish solution.

mods/ts/sequences/bridges.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+lobrdg25:
+	Inherits: lobrdg01
+
+lobrdg26:
+	Inherits: lobrdg01
+
+lobrdg27:
+	idle: lobrdg01
+		Start: 1
```
abcdefg30: Won't inheriting `lobrdg01` and just overwriting the file used for the idle sequence (like already done above) work as well?

Mailaender: Can you give me an example how it should work?

abcdefg30: Sure.
```
lobrdg27:
	Inherits: lobrdg01
	idle: lobrdg01
```
(Or maybe `Defaults: lobrdg01` instead of `idle: lobrdg01`, didn't try it myself yet.)

OpenRA.Mods.TS/Traits/Buildings/LowBridge.cs
Mailaender: Yes. Done.

mods/ts/sequences/bridges.yaml
Mailaender: This is obsolete now with multiple bridge actors combined.

Mailaender:
Rebased so this can be conveniently tested with https://github.com/OpenRA/OpenRA/pull/11211. Replaced individual bridge actors with combined ones importing their supposed damage state. This looks a bit bland as we don't support sprite image randomization, but it isn't terrible.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
```diff
@@ -72,40 +72,43 @@ class ImportTSMapCommand : IUtilityCommand
...
+			{ 0x5B, "lobrdg_a_ne" }, // lobrdg18
+			{ 0x5C, "lobrdg_r_se" }, // lobrdg19
+			{ 0x5D, "lobrdg_r_se" }, // lobrdg20
+			{ 0x5E, "lobrdg_r_nw" }, // lobrdg21
+			{ 0x5F, "lobrdg_r_nw" }, // lobrdg22
+			{ 0x60, "lobrdg_r_ne" }, // lobrdg23
+			{ 0x61, "lobrdg_r_ne" }, // lobrdg24
+			{ 0x62, "lobrdg_r_sw" }, // lobrdg25
+			{ 0x63, "lobrdg_r_sw" }, // lobrdg26
+			{ 0x64, "lobrdg_b" }, // lobrdg27
```
pchote: My spec from #8916 was to have these two dead versions as their own actor types.  This (1) lets us handle the special case adjacency rendering that only the dead state needs, and (2) will let us eventually remove the hacky zombie / resurrection logic that is now only used for bridges.

Mailaender: Added the dead husk actors back.

Mailaender:
This is now looking pretty slick with the transparent bridge husks in the editor:

image

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
```diff
@@ -121,6 +124,51 @@ class ImportTSMapCommand : IUtilityCommand
...
+			{ 0x5D, DamageState.Heavy },
+			{ 0x5E, DamageState.Undamaged },
+			{ 0x5F, DamageState.Heavy },
+			{ 0x60, DamageState.Undamaged },
+			{ 0x61, DamageState.Heavy },
+			{ 0x62, DamageState.Undamaged },
+			{ 0x63, DamageState.Heavy },
+
+			// placeholders for resurrection
+			{ 0x64, DamageState.Dead },
```
pchote: These should be imported as Undamaged now.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
```diff
@@ -303,7 +351,26 @@ static void ReadOverlay(Map map, IniFile file, int2 fullSize)
...
+							else if (damageState == DamageState.Critical)
+								health = 25;
+							else if (damageState == DamageState.Heavy)
+								health = 50;
+							else if (damageState == DamageState.Medium)
+								health = 75;
+							else if (damageState == DamageState.Undamaged)
+								health = 100;
+
+							ar.Add(new HealthInit(health, true));
```
pchote: This should be skipped if `health == 100`.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
Mailaender: Removed the zombie bridges.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
Mailaender: Done.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
```diff
@@ -72,40 +72,43 @@ class ImportTSMapCommand : IUtilityCommand
...
-			{ 0x65, "lobrdg28" },
+
+			{ 0x4A, "lobrdg_b" }, // lobrdg01
+			{ 0x4B, "lobrdg_b" }, // lobrdg02
+			{ 0x4C, "lobrdg_b" }, // lobrdg03
+			{ 0x4D, "lobrdg_b" }, // lobrdg04
+			{ 0x4E, "lobrdg_b_nw" }, // lobrdg05
+			{ 0x4F, "lobrdg_b_se" }, // lobrdg06
+			{ 0x50, "lobrdg_b" }, // lobrdg07
+			{ 0x51, "lobrdg_b_nw" }, // lobrdg08
```
pchote: I'm pretty certain that these two are the art states for the dead bridge (your `lobrdg_b_d`) when there is an intact bridge neighbour (because they are not a full cell, so cannot be walkable).

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
```diff
@@ -72,40 +72,43 @@ class ImportTSMapCommand : IUtilityCommand
...
+			{ 0x52, "lobrdg_b_se" }, // lobrdg09
+			{ 0x53, "lobrdg_a" }, // lobrdg10
+			{ 0x54, "lobrdg_a" }, // lobrdg11
+			{ 0x55, "lobrdg_a" }, // lobrdg12
+			{ 0x56, "lobrdg_a" }, // lobrdg13
+			{ 0x57, "lobrdg_a_sw" }, // lobrdg14
+			{ 0x58, "lobrdg_a_ne" }, // lobrdg15
+			{ 0x59, "lobrdg_a" }, // lobrdg16
+			{ 0x5A, "lobrdg_a_sw" }, // lobrdg17
```
pchote: And likewise these two are the alternative art states for `lobrdg_b_d` when there are intact neighbours.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
```diff
@@ -121,6 +124,51 @@ class ImportTSMapCommand : IUtilityCommand
...
+			{ 0x4B, DamageState.Undamaged },
+			{ 0x4C, DamageState.Undamaged },
+			{ 0x4D, DamageState.Undamaged },
+
+			{ 0x4E, DamageState.Heavy },
+			{ 0x4F, DamageState.Heavy },
+
+			{ 0x50, DamageState.Heavy },
+
+			{ 0x51, DamageState.Critical },
```
pchote: And then these should go back to DamageState.Undamaged to match.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
```diff
@@ -121,6 +124,51 @@ class ImportTSMapCommand : IUtilityCommand
...
+			{ 0x54, DamageState.Undamaged },
+			{ 0x55, DamageState.Undamaged },
+			{ 0x56, DamageState.Undamaged },
+
+			{ 0x57, DamageState.Heavy },
+			{ 0x58, DamageState.Heavy },
+
+			{ 0x59, DamageState.Heavy },
+
+			{ 0x5A, DamageState.Critical },
```
pchote: and these two too.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
```diff
@@ -72,40 +72,43 @@ class ImportTSMapCommand : IUtilityCommand
...
+			{ 0x5E, "lobrdg_r_nw" }, // lobrdg21
+			{ 0x5F, "lobrdg_r_nw" }, // lobrdg22
+			{ 0x60, "lobrdg_r_ne" }, // lobrdg23
+			{ 0x61, "lobrdg_r_ne" }, // lobrdg24
+			{ 0x62, "lobrdg_r_sw" }, // lobrdg25
+			{ 0x63, "lobrdg_r_sw" }, // lobrdg26
+			{ 0x64, "lobrdg_b_d" }, // lobrdg27
+			{ 0x65, "lobrdg_a_d" }, // lobrdg28
+
+			/* High bridge segments
```
pchote: The art frames of these appear to be exact duplicates of the other undamaged ramps.  They definitely aren't for high bridges.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
pchote: Scratch that, after inspecting the original game I see that these should actually be converted to `lobrdg_b`.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
pchote: And these to `lobrdg_a`.

pchote:
I tested the original game in detail and see now that it selects the neighbour-based artwork *on the living* actors rather than the dead.  This is the opposite of how the earlier games did it, but the same rendering principles apply on our side.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
Mailaender: They are not duplicates. If you stitch the tiles together with an undamaged one in the editor using https://github.com/OpenRA/OpenRA/pull/10764 you can see that they leave a gap at the northern or southern side, which is why I have to differentiate them. It may be easier to see with the `lobrdg_a`  variant, because it has a larger gap.

pchote: See my comment in the main thread for more context on how this works.

pchote:
The real issue here is matching the art frames to defined (I suspect pairs of own, neighbour) damage state.  We can't define oddballs like `lobrdg_b_se` because that would mean that those artwork states couldn't be used by the standard bridge types when they or their neighbours are damaged.

OpenRA.Mods.TS/UtilityCommands/ImportTSMapCommand.cs
Mailaender: Shall I set them up as different facings of the same actor?

pchote: Facings isn't a good fit.  This will want to use directional checks like walls.

Mailaender:
Updated this with your recent findings.

abcdefg30:

pchote:
Needs a rebase.

Mailaender:
Rebased.

OpenRA.Mods.TS/Traits/Buildings/LowBridge.cs
```diff
@@ -0,0 +1,50 @@
...
+	class LowBridge : INotifyCreated
+	{
+		readonly LowBridgeInfo info;
+
+		public LowBridge(Actor self, LowBridgeInfo info)
+		{
+			this.info = info;
+		}
+
+		void INotifyCreated.Created(Actor self)
```
pchote: This will need to be done on AddedToWorld, and then a matching RemovedFromWorld should remove the custom terrain.

I'm not sure why I missed this earlier, sorry.

Mailaender: Not yet, the bridges are indestructible for the moment.

pchote: Sure, but that's no excuse to leave this logic half-unfinished in the meantime.

Mailaender: It adds code with no test case, which will probably be broken then.

pchote: The code to avoid bogus state on removal is not complicated.  It can be tested once by making the bridges destroyable and checking with your terrain debug layers.

Mailaender: You are forcing me to jump ahead far from what I wanted to do in this initial pull request. I added destructible bridges and added a test case that we can keep as it won't be game breaking on that map.

OpenRA.Mods.TS/Traits/Buildings/LowBridge.cs
```diff
@@ -24,27 +25,39 @@ class LowBridgeInfo : ITraitInfo, Requires<BuildingInfo>
...
 		}
+
+		void INotifyAddedToWorld.AddedToWorld(Actor self)
+		{
+			var tileSet = self.World.Map.Rules.TileSet;
+			var terrainIndex = tileSet.GetTerrainIndex(info.TerrainType);
+			UpdateTerrain(self, terrainIndex);
+		}
+
+		void INotifyRemovedFromWorld.RemovedFromWorld(Actor self)
```
pchote: You'll need to trigger a check (probably in a FrameEndTask) to kill any actors that are no longer allowed in the cell.  Otherwise you end up with orphaned units where they don't belong:

screen shot 2016-08-26 at 16 59 58

pchote: Note that my last request was to tidy up a specific code issue with the trait so that we could merge the code at a logical and self-contained stopping point.  The extra changes that you added on top of this exposes a new problem which you could alternatively avoid here by removing them for a future PR.  So please don't go off on me about scope-creep, I'm just trying to avoid merging things that are obviously half-baked.

Mailaender: This is a fairly simple and reasonable request that is well inside my skill scope so I don't mind.

Mailaender:
Also updated the map importer to avoid adding 3x more bridges than required. It works fine except for some `lobrdg_a` actors. The reason is probably me still not understanding the isometric coordinate system and how to traverse it in an orderly manner.

pchote:
Please reset this onto https://github.com/pchote/OpenRA/commits/lowbridges and then we can get this merged.

Mailaender:
Thanks for helping me out with the importer. It now imports bridges without the need for manual tweaking. :+1:

abcdefg30:
Looks still good to me. :+1: / :white_check_mark:

pchote:
:tada:

Mailaender:
:bridge_at_night:

PR #11448, opened 2016-06-12
Updated to Mono 4.4

Mailaender:
See http://www.mono-project.com/docs/about-mono/releases/4.4.0/


Mailaender:
It isn't available via http://download.mono-project.com/repo/debian/dists/wheezy/snapshots/

PR #11456, opened 2016-06-15
Disable single-player games by default on dedicated servers

obrakmann:
Rationale: it's stupid to allow single players to hog someone else's server by themselves by default. This change would force them to use the single-player lobby instead.

### DIFF
--- a/OpenRA.Game/Server/Server.cs
+++ b/OpenRA.Game/Server/Server.cs
@@ -149,7 +149,7 @@ namespace OpenRA.Server
 					RandomSeed = randomSeed,
 					Map = settings.Map,
 					ServerName = settings.Name,
-					DisableSingleplayer = settings.DisableSinglePlayer,
+					EnableSingleplayer = settings.EnableSingleplayer || !dedicated,
 				}
 			};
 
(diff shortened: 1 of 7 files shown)


OpenRA.Game/Server/Server.cs
```diff
@@ -394,7 +394,7 @@ void ValidateClient(Connection newConn, string data)
 				if (!LobbyInfo.IsSinglePlayer && Map.DefinesUnsafeCustomRules)
 					SendOrderTo(newConn, "Message", "This map contains custom rules. Game experience may change.");
 
-				if (Settings.DisableSinglePlayer)
+				if (!LobbyInfo.GlobalSettings.EnableSingleplayer)
 					SendOrderTo(newConn, "Message", "Singleplayer games have been disabled on this server.");
```
pchote: Minor scope creep: could you change this message to something like "Two or more human players are required by this server."?  IMO "Singleplayer" has always been a bit too ambiguous.

pchote:
Code looks good, and works as advertised :+1:

obrakmann:
Done.

Phrohdoh:
Since we've changed the message (wrt 'singleplayer') should the member names change too?

obrakmann:
I wouldn't mind changing the name of the setting. I have trouble coming up with a name that is both succinct and descriptive enough, though. Any suggestions?

abcdefg30:
I think the name is still quite descriptive as the setting (dis)allowes if only a single player can play...

We just have inconsistencies on the case. You sometimes used `Singeplayer` and other times `SinglePlayer`. I'd say we should only use one of them (to be consistent)?

abcdefg30:
Looks good to me otherwise. 👍

obrakmann:
Changed the names to `EnableSingleplayer` everywhere.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a07d045378564146371aea5252c59783e2fb21a6...298bfd7e67263e4e4d457edfce26f05bdb5f82bc)

Issue #9266, opened 2015-09-05
Not all Armaments ranges are taken into account in reach target checks


Holloweye:
If you have multiple Armaments only the first one will be counted for the unit range. And if you for example disable (with upgrades) the first armament (which might have a longer range then the second Armament) then the unit will just stand still and try to shot but cant reach the target. As the range for the units is not updated upon disabling the first Armament.

obrakmann:
Sounds like a duplicate of #8184.

pchote:
Maybe fixed by #8768?

obrakmann:
No, unfortunately it doesn't.

matija-hustic:
Could you check if #8768 solved this?

abcdefg30:
This should be fixed now (if not by #8768 then by #11459).

PR #11459, opened 2016-06-16
Fix weapon min/max range and validity checks

### LINKED ISSUE #7499: Aircraft targeting actors that are too close
I don't know what happened here:

The AI built a ranger, I ordered my Mig to attack. (It fired once and missed).
The ranger started moving back and forth like a viceroid and the Mig just kept circling above it. They kept doing that for over 5 minutes.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8698852-aircraft-targeting-actors-that-are-too-close?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #11430: Air units won't close the distance to attack other air units if ground weapon range is different
As was mentioned by @HenrytheSlav in https://github.com/OpenRA/OpenRA/pull/11337#issuecomment-221490963 

> it was found that if longbow has two weapons with different ranges, f.e. AA 5 cells and ground 6 cells, then it will fail to shoot/move closer when at 6 cells distance against aircraft because it tries to use wrong weapon.

On bleed

reaperrr:
By only considering weapons valid versus the selected target.

Additionally adressed helicopter-related TODOs regarding weapons that are out of ammo and need reloading (they are now considered invalid weapons until their ammo is replenished).

Fixes #11430.
Probably fixes #7499.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Air/HeliAttack.cs
+++ b/OpenRA.Mods.Common/Activities/Air/HeliAttack.cs
@@ -65,9 +65,8 @@ namespace OpenRA.Mods.Common.Activities
 				return ActivityUtils.SequenceActivities(new HeliFly(self, newTarget));
 			}
 
-			// If all ammo pools are depleted and none reload automatically, return to helipad to reload and then move to next activity
-			// TODO: This should check whether there is ammo left that is actually suitable for the target
-			if (ammoPools.All(x => !x.Info.SelfReloads && !x.HasAmmo()))
+			// If any AmmoPool is depleted and no weapon is valid against target, return to helipad to reload and then move to next activity
+			if (ammoPools.Any(x => !x.Info.SelfReloads && !x.HasAmmo()) && !attackHeli.HasAnyValidWeapons(target))
 				return ActivityUtils.SequenceActivities(new HeliReturnToBase(self), NextActivity);
 
 			var dist = target.CenterPosition - self.CenterPosition;
--- a/OpenRA.Mods.Common/Activities/Air/HeliAttack.cs
+++ b/OpenRA.Mods.Common/Activities/Air/HeliAttack.cs
@@ -80,13 +80,10 @@ namespace OpenRA.Mods.Common.Activities
 				return this;
 
 			// Fly towards the target
-			// TODO: Fix that the helicopter won't do anything if it has multiple weapons with different ranges
-			// and the weapon with the longest range is out of ammo
 			if (!target.IsInRange(self.CenterPosition, attackHeli.GetMaximumRangeVersusTarget(target)))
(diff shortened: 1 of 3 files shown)


reaperrr:
Rebased and moved testcase [here](https://github.com/reaperrr/OpenRA/tree/fix-range-check-test).

GraionDilach:
Could this resolve the TODOs in HeliAttack or is that too much of a nit?

reaperrr:
> Could this resolve the TODOs in HeliAttack or is that too much of a nit?

It's on my agenda, but I'd like to leave that for a follow-up PR.

Main reason, apart from making this PR larger, is that it would be better if I completed my `UpgradeWhileDocked`-based reloading refactor for `AmmoPool` first, otherwise I'd have to rewrite a good part of the fix later, and I don't like wasting work just because I did things in the wrong order.

pchote:
I may be missing something, but why can't you just skip weapons without ammo when calculating the range?

reaperrr:
Well, turned out to be a relatively simple change, so yeah.

Updated.

pchote:
This behaves very nicely ingame, with the exception of one regression (using your test commit): you lose the ability to force-fire on aircraft when you run out of AA ammo.

reaperrr:
Context/reminder:
```
[22:52] <+pchote> if the behaviour is consistent then you would lose the attack cursor completely if the unit only has one ammo type
[22:52] <+pchote> which is most certainly wrong
[22:53] <+pchote> and if you don't, then that's probably a strong hint as to where to fix it
[22:53] <+pchote> i'd start at the order targeter and work down until you find which of the things you modified is called
[22:53] <+pchote> you can probably just get away with removing one of your sets of ammo checks
```

pchote:
Does this fix #9266?

reaperrr:
> Does this fix #9266?

Yes.

reaperrr:
Updated. That issue was indeed caused by one ammo check too much (in `ChooseArmamentsForTarget`).

Additionally, I added a commit that makes the target cursor take into consideration which of the valid armaments has the highest range and whether ammo is available, rather than unconditionally taking the first valid armament.

reaperrr:
I'd like to see this make it into the next release.

obrakmann:
Looks fine to me, and didn't notice any misbehaviour in-game. :+1:

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/0ab9baa3283eb012251999796faac0316e092277...d8ca7063b49310fb1496a93145ebeba9642e4cb3)

PR #11460, opened 2016-06-16
Fix damage of RA/TD minigunners and RA Jeep vs. Heavy armor

### LINKED ISSUE #11425: Veterancy on Heavy Armour makes them invincible to riflemen
Found this bug in RA and TD
Heavy Armour vehicles receive no damage from rifleman

http://logs.openra.net/index.php?year=2016&month=06&day=10#16:08:39

reaperrr:
Fixes #11425.

### DIFF
--- a/mods/cnc/weapons/smallcaliber.yaml
+++ b/mods/cnc/weapons/smallcaliber.yaml
@@ -113,7 +113,7 @@ M16:
 			None: 100
 			Wood: 25
 			Light: 30
-			Heavy: 10
+			Heavy: 15
 		DamageTypes: Prone50Percent, TriggerProne, BulletDeath
 	Warhead@2Eff: CreateEffect
 		Explosions: piff
--- a/mods/ra/weapons/smallcaliber.yaml
+++ b/mods/ra/weapons/smallcaliber.yaml
@@ -265,7 +265,7 @@ M60mg:
 		Versus:
 			Wood: 10
 			Light: 30
-			Heavy: 10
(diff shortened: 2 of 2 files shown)


Mailaender:
:+1:

pchote:
The balance implications should be negligible in practice, so this looks completely reasonable to me.  :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a76ee1959155c1624bf2c669a5754c528e453847

MurtoTheRay:
This balance change makes tanks melt to riflemen. It was also based on a false claim, riflemen can damage heavy tanks: https://snag.gy/8OGMJt.jpg The top right one has been fired on by a single rifleman, the top right by two and the bottom middle by around 10.

AoAGeneral:
This actually didn't fix the issue. As @MurtoTheRay mentions it is now killing tanks like crazy. The issue that Anjew was pointing out that tanks with veterancy was immune to minigunner fire. But only tanks with veterancy.

SoScared:
Compared to the release the riflemen deals twice the damage against Heavy Tanks with this change (280 shots to kill the Heavy tank in the bleed vs 550 in the realease). Tanks cannot support infantry with this fix.

Riflemen doing little to no damage vs heavy armor was never an issue.

SoScared:
Alternatively #11719 could fix this issue rather than give riflemen more damage.

PR #11462, opened 2016-06-16
Extract installer data using direct stream-stream copies

pchote:
The old implementation would read all the data into a memory buffer and then dump that buffer to disk as two separate options.  This is bad for the UI (no way to measure progress) and for memory (because we need to load entire files into memory – potentially twice).

This removes the intermediate memory buffers and adds (XX%) indicators to the installation progress UI.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Installation/InstallFromDiscLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Installation/InstallFromDiscLogic.cs
@@ -172,10 +172,18 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 									}
 
 									Log.Write("install", "Copying {0} -> {1}".F(sourcePath, targetPath));
-									message = "Copying " + Path.GetFileName(sourcePath);
 									extracted.Add(targetPath);
 									Directory.CreateDirectory(Path.GetDirectoryName(targetPath));
-									File.Copy(sourcePath, targetPath);
+
+									using (var source = File.OpenRead(sourcePath))
+									using (var target = File.OpenWrite(targetPath))
+									{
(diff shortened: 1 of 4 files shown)


OpenRA.Mods.Common/Widgets/Logic/Installation/InstallFromDiscLogic.cs
```diff
@@ -226,9 +234,25 @@ void InstallFromDisc(string path, ModContent.ModDisc disc)
...
 		}
 
+		static void CopyStream(Stream input, Stream output, long length, Action<long> onProgress = null)
+		{
+			var buffer = new byte[4096];
+			var copied = 0L;
+			while (copied < length)
+			{
+				var n = (int)Math.Min(buffer.Length, length - copied);
+				input.Read(buffer, 0, n);
```
RoosterDragon: Need to check the actual bytes read here.

pchote: Good catch.  Fixed.

pchote:
Added a new commit that adds a 25 MB size threshold for the % indicator.

Phrohdoh:
Why 25MB?

pchote:
It is a nice number that is larger than all the small files that we don't want the indicator on, but smaller than the large files that we do want it on. I wouldn't want to go much bigger for the sake of people stuck with spinning disks, or much smaller for the sake of people with SSDs.

Mailaender:

obrakmann:
Looks fine to me, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5b236ab2ed4e93d51334fddba6d0771444cee0a4)

Issue #11305, opened 2016-05-16
AI doesn't deploy replacement MCV


pizzaoverhead:
After destroying the AI's conyard in bleed d2k, its remaining two MCVs continued to path across the map from one buildable area to the next, without ever deploying either MCV. The cause of this appears to be that ActorExtensionsForMove.IsMoving doesn't detect MovePart moves (MoveFirstHalf and MoveSecondHalf), resulting in new location orders regularly being appended to the units' queues. While debugging, there were just under 2000 Move orders queued for one of the MCVs.

GraionDilach:
This sounds like the reason I also tend to see that it takes 4 minutes at least for my AI to deploy it's mobile war factories.

obrakmann:
Thanks for the report, but reaperrr was quicker :)

Duplicate of #11304.

reaperrr:
Reopening this because accoding to my tests, fixing the `IsMove` check alone won't be enough, `HackyAI.FindAndDeployBackupMcv` is simply crap.

pchote:
Closing as a reverse-duplicate of #12159.

PR #11463, opened 2016-06-16
Improve AI MCV deployment

reaperrr:
While #11308 has merit of its own, it doesn't really solve #11305.

Instead, I took a closer look at the AI MCV deployment code. This PR fixes the two more problematic out of three issues I could find:
- The first is that the AI was spamming orders even to MCVs that were already on the move.
- The second is that the AI often chose locations near the enemy base, usually resulting in the MCV's destruction before it's even deployed.

The third is that `ChooseBuildLocation` works rather poorly in practice, sometimes choosing locations where some obstacle doesn't allow deployment, but fixing that will be harder and is therefore left for some other time.

The commits should be descriptive enough, so I won't repeat the details here.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -825,8 +825,7 @@ namespace OpenRA.Mods.Common.AI
 
 			foreach (var mcv in mcvs)
 			{
-				var mover = mcv.TraitOrDefault<IMove>();
-				if (mover != null && mover.IsMoving)
+				if (!mcv.IsIdle)
 					continue;
 
 				var factType = mcv.Info.TraitInfo<TransformsInfo>().IntoActor;
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -125,6 +125,9 @@ namespace OpenRA.Mods.Common.AI
 		[Desc("Radius in cells around the center of the base to expand.")]
 		public readonly int MaxBaseRadius = 20;
 
+		[Desc("Should deployment of additional MCVs be restricted to MaxBaseRadius if explicit deploy locations are missing or occupied?")]
+		public readonly bool RestrictMCVDeploymentFallbackToBase = true;
+
 		[Desc("Radius in cells around each building with ProvideBuildableArea",
 			"to check for a 3x3 area of water where naval structures can be built.",
 			"Should match maximum adjacency of naval structures.")]
@@ -817,7 +820,7 @@ namespace OpenRA.Mods.Common.AI
 		}
 
 		// Find any newly constructed MCVs and deploy them at a sensible
-		// backup location within the main base.
+		// backup location.
 		void FindAndDeployBackupMcv(Actor self)
 		{
 			var mcvs = self.World.Actors.Where(a => a.Owner == Player &&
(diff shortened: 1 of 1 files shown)


abcdefg30:
Won't this make the AI only deploy mcvs in their build radius?
Not sure if that is really beneficial (as the build radii overlap).

reaperrr:
> Won't this make the AI only deploy mcvs in their build radius?

Yes. In my opinion overlapping of build radi is a far lesser evil than never getting to deploy the MCV at all or lose the deployed yard too quickly, at least for the AI.

Deploying it within the build radius will still increase the maximum size of the base, increase build speed and make it easier to defend the new CYard compared to deploying it somewhere random and getting it destroyed/captured on its way there or shortly after.

If all player/AI slots are filled, the area around the own starting location is usually the sanest spot to deploy a second MCV anyway.


We can revisit this aspect once we come up with a smarter deploy location decision-making for the AI, but that is worth its own PR as it will likely require more work.

pchote:
> We can revisit this aspect once we come up with a smarter deploy location decision-making for the AI, but that is worth its own PR as it will likely require more work.

Sensible games allow map authors to add AI metadata for things like secondary base locations using waypoints or regions.  I've been saying for a long time that we should do this for OpenRA as part of our much needed AI overhaul.

reaperrr:
What we could do as a first step in that direction would be to make the AI check the spawnpoints on the map first, and if there are any with no enemies within a certain perimeter around that spawnpoint, send the MCV there, and if none are available, deploy it inside the own spawnpoint perimeter.

GraionDilach:
I might be alone here but I like the wandering mobile war factory mode (abusing the AI Mcvs list for MWFs), with considering only the "find-location-go-elsewhere-only-deploy-on-the-fifth-fitting-location" aspect as a bug, but such an additional waypoint system sounds good too.

yaaaaa:
I think that AI should build an MCV, even if it has a CONYARD.
And then deploy an MCV to another CONYARD next to the spawn resources area (oge mine or gem mine).
AI must do so prior to the moment when all the ore mines and/or gem mines areas will not be busy itself conyards.

pchote:
Needs a rebase now.

reaperrr:
Rebased.
I'm a bit low on head-bandwidth at the moment and according to https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/AI/HackyAI.cs#L820 commit 2 is indeed a bugfix, so I'll stick to this for now and leave spawnpoint consideration or AI metadata for dedicated follow-up(s).

Nelax:
Much better logic. The AI tries to keep base integrity with deploying units. Maybe in the future the AI will make a calculated decision on whether to deploy the MCV near its base or not.
[img]http://imgur.com/xlW3dal.png[/img]

Phrohdoh:
I consider this a sane base expansion strategy. 👍 

```diff
diff --git a/mods/ra/rules/ai.yaml b/mods/ra/rules/ai.yaml
index bafed36..f322f25 100644
--- a/mods/ra/rules/ai.yaml
+++ b/mods/ra/rules/ai.yaml
@@ -43,24 +43,7 @@ Player:
 			fix: 0.1%
 			dome: 10%
 		UnitsToBuild:
-			e1: 50%
-			e2: 20%
-			e3: 10%
-			e4: 2%
-			dog: 2%
-			shok: 2%
-			harv: 10%
-			apc: 30%
-			jeep: 40%
-			arty: 15%
-			v2rl: 40%
-			ftrk: 50%
-			1tnk: 70%
-			2tnk: 25%
-			3tnk: 50%
-			4tnk: 10%
-			ttnk: 10%
-			stnk: 5%
+			mcv: 100%
 		UnitLimits:
 			dog: 4
 			harv: 8
```

screen shot 2016-06-24 at 11 14 14 am

reaperrr:
Updated. Made the restriction to base radius optional.

reaperrr:
I count a total of +3 so far and Nolt confimed on IRC that this fixes his problems with flying MCVs in his mod, so I'll merge it now.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8963319f838e8f4efbe662ca82a062456ea1fe2f...7eb9f03c390e3c3843e8de9618c8282a46363545)

PR #11469, opened 2016-06-18
Fix initial server map randomization.

pchote:
This was previously trying to use values that had not yet been initialised, which meant that _all_ maps were treated as unsuitable.  This would cause the mod to always fall back to the first installed map.

Moving this to MapCache means avoids that subtlety, and fixes the issue.

### DIFF
--- a/OpenRA.Game/Map/MapCache.cs
+++ b/OpenRA.Game/Map/MapCache.cs
@@ -243,11 +243,31 @@ namespace OpenRA
 				});
 		}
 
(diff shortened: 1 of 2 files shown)


abcdefg30:
👍

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a76ee1959155c1624bf2c669a5754c528e453847...b20c9970d129be5c43dd99366900fe23eeab43ee)

PR #11470, opened 2016-06-18
Fix Mobile.IsMovingInMyDirection check.

pchote:
The old code would always return false (because `selfMobile.ToCell == self.Location` by definition), so check has effectively been a no-op for years.

My tests indicate a huge improvement to unit behavior across bridges as long as you issue move orders far enough past the bridge that you are outside the 8-cell `nearEnough` radius that will cause units to give up and stop on/in front of the bridge.


reaperrr:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1b1e07ae1959faabdda1e5bc755c978bd609022b)

PR #11473, opened 2016-06-18
Disable hide map crate if explored map is enabled.

pchote:
Suggested by lucassss in http://www.sleipnirstuff.com/forum/viewtopic.php?t=19622.

While it could certainly be argued that the hide map penalty remains valid, I think it is reasonable to treat the player's choice to explore the map as a request to disable the shroud completely.  This is also consistent with our original disabled-shroud behavior.


obrakmann:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/92285c894e789552c93fabd0bd90af71c4236ca7...b7ced950406170fb5d2a95cb5081590884b8620e)

PR #11474, opened 2016-06-18
Add foundation for player experience tracking

obrakmann:
Adds a `PlayerExperience` trait, first and foremost. The experience is added to by killing units with the `GivesExperience` trait. That behaviour can be removed by setting the `GivesExperienceInfo.PlayerExperienceModifier` to 0 (the default). The trait can otherwise be used to simply track a score, for example in Lua scripts.

The original PR enabled the feature in the existing mods, both in the rules and the UI. I'm not sure if we should actually do that. I added the rules changes as a test case, and split off the UI changes into a separate commit so that they can be easily removed, if we decide that way.

Supersedes #10358.

//cc: @Phoib

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Properties/PlayerExperienceProperties.cs
+++ b/OpenRA.Mods.Common/Scripting/Properties/PlayerExperienceProperties.cs
@@ -1,0 +1,43 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using Eluant;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Scripting;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Scripting
+{
+	[ScriptPropertyGroup("Player")]
+	public class PlayerExperienceProperties : ScriptPlayerProperties, Requires<PlayerExperienceInfo>
+	{
+		readonly PlayerExperience exp;
+
+		public PlayerExperienceProperties(ScriptContext context, Player player)
+			: base(context, player)
+		{
+			exp = player.PlayerActor.Trait<PlayerExperience>();
+		}
+
+		public int Experience
+		{
+			get
+			{
+				return exp.Experience;
+			}
+
+			set
+			{
+				exp.GiveExperience(value - exp.Experience);
+			}
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Traits/GivesExperience.cs
+++ b/OpenRA.Mods.Common/Traits/GivesExperience.cs
@@ -22,6 +22,12 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Stance the attacking player needs to receive the experience.")]
 		public readonly Stance ValidStances = Stance.Neutral | Stance.Enemy;
 
+		[Desc("Percentage of the `Experience` value that is being granted to the killing actor.")]
+		public readonly int ActorExperienceModifier = 10000;
+
+		[Desc("Percentage of the `Experience` value that is being granted to the player owning the killing actor.")]
+		public readonly int PlayerExperienceModifier = 0;
+
 		public object Create(ActorInitializer init) { return new GivesExperience(init.Self, this); }
 	}
 
@@ -44,14 +50,17 @@ namespace OpenRA.Mods.Common.Traits
 
 			var valued = self.Info.TraitInfoOrDefault<ValuedInfo>();
 
-			// Default experience is 100 times our value
 			var exp = info.Experience >= 0
 				? info.Experience
-				: valued != null ? valued.Cost * 100 : 0;
+				: valued != null ? valued.Cost : 0;
 
 			var killer = e.Attacker.TraitOrDefault<GainsExperience>();
 			if (killer != null)
-				killer.GiveExperience(exp);
+				killer.GiveExperience(Util.ApplyPercentageModifiers(exp, new[] { info.ActorExperienceModifier }));
+
+			var attackerExp = e.Attacker.Owner.PlayerActor.TraitOrDefault<PlayerExperience>();
+			if (attackerExp != null)
+				attackerExp.GiveExperience(Util.ApplyPercentageModifiers(exp, new[] { info.PlayerExperienceModifier }));
 		}
 	}
 }
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -245,6 +245,7 @@
     <Compile Include="Scripting\Properties\MissionObjectiveProperties.cs" />
     <Compile Include="Scripting\Properties\MobileProperties.cs" />
     <Compile Include="Scripting\Properties\DemolitionProperties.cs" />
+    <Compile Include="Scripting\Properties\PlayerExperienceProperties.cs" />
     <Compile Include="Scripting\Properties\PlayerProperties.cs" />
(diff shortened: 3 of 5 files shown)


OpenRA.Mods.Common/Traits/GivesExperience.cs
```diff
@@ -22,6 +22,14 @@ class GivesExperienceInfo : ITraitInfo
 		[Desc("Grant experience for team-kills.")]
 		public readonly bool FriendlyFire = false;
 
+		[Desc("Modifier with which the `Experience` value will be multiplied (in percent)",
+			"before being granted to an actor.")]
+		public readonly int ActorExperienceModifier = 10000;
+
+		[Desc("Modifier with which the `Experience` value will be multiplied (in percent)",
+			"before being granted to a player.")]
+		public readonly int PlayerExperienceModifier = 1;
```
obrakmann: Note that the default value here is 0, the 1 is from the test case commit.

mods/cnc/chrome/ingame-info.yaml
```diff
@@ -2,6 +2,7 @@ Container@GAME_INFO_PANEL:
 	X: (WINDOW_RIGHT - WIDTH) / 2
 	Y: (WINDOW_BOTTOM - HEIGHT) / 2
 	Width: 512
+	Width: 582
```
pchote: Looks like you accidentally dropped removing the old value here.

OpenRA.Mods.Common/Scripting/Properties/PlayerExperienceProperties.cs
```diff
@@ -0,0 +1,48 @@
...
+		public int Experience
+		{
+			get
+			{
+				return exp.Experience;
+			}
+
+			set
+			{
+				exp.GiveExperience(exp.Experience - value);
```
abcdefg30: This doesn't look quite correct. Wouldn't it need to be `-exp.Experience + value` (or `value - exp.Experience`)?

OpenRA.Mods.Common/Traits/GivesExperience.cs
```diff
@@ -22,6 +22,14 @@ class GivesExperienceInfo : ITraitInfo
 		[Desc("Grant experience for team-kills.")]
 		public readonly bool FriendlyFire = false;
 
+		[Desc("Modifier (in percent) with which the `Experience` value will be multiplied",
+			"before being granted to an actor.")]
+		public readonly int ActorExperienceModifier = 10000;
```
abcdefg30: Why do we inflate this by a factor of 100 (i.e. making it a percentage)? Wouldn't just `100` work as well?

OpenRA.Mods.Common/Scripting/Properties/PlayerExperienceProperties.cs
```diff
@@ -0,0 +1,48 @@
...
+			}
+		}
+
+		public void GiveExperience(int num)
+		{
+			exp.GiveExperience(num);
```
abcdefg30: As discussed on IRC: This seems to be quite redundant to the setter above.

OpenRA.Mods.Common/Scripting/Properties/PlayerExperienceProperties.cs
obrakmann: Yup, that's reversed.

OpenRA.Mods.Common/Traits/GivesExperience.cs
obrakmann: It *is* a percentage. Actors really do get 100x the value of the unit they killed counted towards their experience. If we didn't inflate the value, you couldn't increase experience by less than the killed actor's value (ie. anything below 100%).

OpenRA.Mods.Common/Scripting/Properties/PlayerExperienceProperties.cs
obrakmann: I don't quite agree, since it's semantically different, but I'll remove it.

obrakmann:
Updated.

OpenRA.Mods.Common/Scripting/Properties/PlayerExperienceProperties.cs
pchote: We can always add it back later if we need it.  It would be much harder to remove, so i'm in favour of being conservative.

OpenRA.Mods.Common/Traits/GivesExperience.cs
abcdefg30: I know that this is a percentage. My question was if we really need this high value, `100 * 100`.
As we divide by [100](https://github.com/OpenRA/OpenRA/pull/11474/commits/70a58e82c2938222fb719bd5a44048e7d3119437#diff-73755936b9d513f9b2e7ad3a3f11bbd2R65) later anyway it wouldn't matter to just make this a "normal" number (no percentage), i.e. just `100`.

abcdefg30: Not a valid issue (of this PR), as discussed on IRC.

OpenRA.Mods.Common/Traits/GivesExperience.cs
```diff
@@ -22,6 +22,14 @@ class GivesExperienceInfo : ITraitInfo
 		[Desc("Stance the attacking player needs to receive the experience.")]
 		public readonly Stance ValidStances = Stance.Neutral | Stance.Enemy;
 
+		[Desc("Modifier (in percent) with which the `Experience` value will be multiplied",
```
pchote: After all the recent misunderstandings I suggest we blacklist any use of the word "multiply" referring to modifiers.

Can we change this to "Percentage of the Experience value that is granted to the killer"?  This is more in line with our descriptions for other modifiers?

pchote:
:+1: for taking this without the testcase.

obrakmann:
Updated:
* Changed the wording in the `Desc`
* Used `Util.ApplyPercentageModifiers` instead of dividing by 100

GraionDilach:
:+1:  and I'd keep a testcase otherwise it just shows up as a visible 0 in the stats screens.

obrakmann:
I have a follow-up that adds granting player experience to `DonateSupplies`, `Captures`, `ExternalCaptures`, `Infiltrates`, `RepairBuildings` and `RepairsUnits`. I can re-do the changes from the test case there.

pchote:
Maybe carry over the gui commit to that PR too then?

obrakmann:
Done

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/e5c139a0cc5aaed95c04aa8aa0e59f5fcc11cec8...a09994761eebce06846591c14bb95e9957ba5f2b)

OpenRA.Mods.Common/Scripting/Properties/PlayerExperienceProperties.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using Eluant;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Scripting;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Scripting
+{
+	[ScriptPropertyGroup("Player")]
+	public class PlayerExperienceProperties : ScriptPlayerProperties, Requires<PlayerExperienceInfo>
```
abcdefg30: `Requires<PlayerExperienceInfo>`! This breaks all scripts for me.

We can either wait for #11627 to add the trait to the default mod rules, or take something like https://github.com/OpenRA/OpenRA/compare/bleed...abcdefg30:reqExp.

pchote: As discussed in IRC this looks like a bug in the scripting.  Working around it in the player experience code would be wrong.

pchote: This happens because https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Scripting/ScriptContext.cs#L158 does not filter on Requires, like actors do.

PR #11476, opened 2016-06-19
Remove spurious space in TS' mod.yaml

phrohdoh:
Introduced in 27699ce3d732ff448a15dde771c5aab5c38542b3 for seemingly no reason.

As far as I am aware we shouldn't be parsing this properly, but apparently we are.


Mailaender:
:+1:

pchote:
Oops.

PR #11479, opened 2016-06-19
Improve WithShadow

reaperrr:
Makes the trait public, upgradable and the offset customizable.

Uses the customizable offset to tweak shadow position of TD and RA aircraft, which looks a little better especially for landed aircraft (they no longer completely hide the shadow).

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithShadow.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithShadow.cs
@@ -21,6 +21,12 @@ namespace OpenRA.Mods.Common.Traits.Render
 	{
 		[PaletteReference] public readonly string Palette = "shadow";
 
+		[Desc("Shadow position offset relative to actor position (ground level).")]
+		public readonly WVec Offset = WVec.Zero;
+
+		[Desc("Shadow Z offset relative to actor sprite.")]
+		public readonly int ZOffset = -5;
+
 		public override object Create(ActorInitializer init) { return new WithShadow(this); }
 	}
 
@@ -46,8 +52,8 @@ namespace OpenRA.Mods.Common.Traits.Render
 			var height = self.World.Map.DistanceAboveTerrain(self.CenterPosition).Length;
 			var shadowSprites = r.Where(s => !s.IsDecoration)
 				.Select(a => a.WithPalette(wr.Palette(info.Palette))
-					.OffsetBy(new WVec(0, 0, -height))
-					.WithZOffset(a.ZOffset + height)
+					.OffsetBy(info.Offset - new WVec(0, 0, height))
+					.WithZOffset(a.ZOffset + (height + info.ZOffset))
 					.AsDecoration());
 
 			return shadowSprites.Concat(r);
--- a/OpenRA.Mods.Common/Traits/Render/WithShadow.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithShadow.cs
@@ -16,25 +16,32 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits.Render
 {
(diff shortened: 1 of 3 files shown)


Mailaender:
The improvement seems to be very subtle. :white_check_mark:

reaperrr:
It's a little more visible with pixel-doubling.

OpenRA.Mods.Common/Traits/Render/WithShadow.cs
```diff
@@ -16,30 +16,40 @@
...
+
+			if (self.IsDead || !self.IsInWorld)
+				return Enumerable.Empty<IRenderable>();
+
 			// Contrails shouldn't cast shadows
 			var height = self.World.Map.DistanceAboveTerrain(self.CenterPosition).Length;
 			var shadowSprites = r.Where(s => !s.IsDecoration)
 				.Select(a => a.WithPalette(wr.Palette(info.Palette))
-					.OffsetBy(new WVec(0, 0, -height))
+					.OffsetBy(new WVec(info.Offset.X, info.Offset.Y, info.Offset.Z - height))
```
pchote: You can simplify this to `info.Offset - new WVec(0, 0, height)`.

mods/cnc/rules/defaults.yaml
```diff
@@ -153,6 +153,7 @@
 		GenericName: Helicopter
 	WithFacingSpriteBody:
 	WithShadow:
+		Offset: 43, 43, -85
```
pchote: Putting the shadows below ground level isn't great and may cause rendering glitches as we progress further with zbuffer support.  Can you move that offset into Y instead?

reaperrr:
Don't merge this yet, one more fixup coming in.

reaperrr:
Updated. ZOffset is now customziable as well.

That's necessary, because positive Y values otherwise make the shadow overlap the body sprite.

GraionDilach:
Could WithShadow get an option that instead of the actor, it'd generate the sahdow from a custom overlay? RA2 used a "blob" shadow from jjshadow.shp instead of this method.

reaperrr:
I don't feel like doing that via this PR.

Besides, you can already do that by adding a `WithSpriteBody@shadow` with `Sequence: shadow` to your actor (that's what I do in my MechCommander mod).

OpenRA.Mods.Common/Traits/Render/WithShadow.cs
```diff
@@ -16,31 +16,44 @@
...
+	public class WithShadowInfo : UpgradableTraitInfo
 	{
 		[PaletteReference] public readonly string Palette = "shadow";
 
-		public object Create(ActorInitializer init) { return new WithShadow(this); }
+		[Desc("Shadow position offset relative to actor position (ground level).")]
+		public readonly WVec Offset = WVec.Zero;
+
+		[Desc("Shadow Z offset relative to actor sprite.")]
+		public readonly int ZOffset = -5;
```
abcdefg30: Hm, couldn't we just include this in the normal `Offset`?

pchote: No, they are very different things.

reaperrr: > Hm, couldn't we just include this in the normal Offset?

`Offset` is a world position offset, while `ZOffset` could also be called `ZSortingOffset`, but we already call it just `ZOffset` everywhere else.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/b7ced950406170fb5d2a95cb5081590884b8620e...686afa9caf1a5b05407d104fd56bf039e299d21b)

PR #11484, opened 2016-06-19
Added burning trees to the Tiberian Sun mod

Mailaender:
A new attempt after the feedback from https://github.com/OpenRA/OpenRA/pull/11405.

Closes #11484.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Burns.cs
+++ b/OpenRA.Mods.Common/Traits/Burns.cs
@@ -9,6 +9,7 @@
  */
 #endregion
 
+using System.Collections.Generic;
 using OpenRA.Graphics;
 using OpenRA.Mods.Common.Traits.Render;
 using OpenRA.Traits;
@@ -16,7 +17,7 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("This actor will play a fire overlay animation over its body and take damage over time.")]
-	class BurnsInfo : ITraitInfo, Requires<RenderSpritesInfo>
+	class BurnsInfo : PausableConditionalTraitInfo, Requires<RenderSpritesInfo>
 	{
 		public readonly string Image = "fire";
 		[SequenceReference("Image")] public readonly string Anim = "1";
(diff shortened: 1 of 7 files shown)


mods/ts/rules/defaults.yaml
```diff
@@ -668,6 +668,31 @@
...
+		DamageType: Incendiary
+		Image: burn-s
+		MinimumDamageState: Light
+		MaximumDamageState: Dead
+		Palette: effect
+	WithDamageOverlay@MediumBurn:
+		DamageType: Incendiary
+		Image: burn-m
+		MinimumDamageState: Medium
+		MaximumDamageState: Dead
```
reaperrr: This makes all three fire overlays play at once when the tree reaches `Heavy` damage state.

The Maximum should be `Light` for `SmallBurn` and `Medium` here.

mods/ts/rules/defaults.yaml
```diff
@@ -668,6 +668,31 @@
...
+		Palette: effect
+	WithDamageOverlay@LargeBurn:
+		DamageType: Incendiary
+		Image: burn-l
+		MinimumDamageState: Heavy
+		MaximumDamageState: Dead
```
reaperrr: These are the defaults anyway, so you can remove line 692 and 693.

reaperrr:
There are still two major aspects missing compared to original TS:

- Burning trees would continously take damage as long as the fire animation was running. So I actually doubt using `WithDamageOverlay` is the right approach, it doesn't loop long enough either. You might be better off using `Burns`, though in its current state that trait is too limited, it will need to be made upgradable and the animation part split off to a render trait and unhardcoded at least.
- Trees would incinerate nearby trees when the flames grew larger. This can be done by using `ExplosionOnDamageTransition` to fire a weapon at `Critical` (or maybe at `Heavy`) that does nothing but incinerate (needs a large enough spread of course, but make sure it doesn't exceed the 8 adjacent cells around the tree).

GraionDilach:
Regarding @reaperrr's point 1, I'd say grant a negative SelfHeal upgrade instead (Burns can completely be replaced with a negative SelfHeal and a WithIdleOverlay afterall as-is).

reaperrr:
> Regarding @reaperrr's point 1, I'd say grant a negative SelfHeal upgrade instead (Burns can completely be replaced with a negative SelfHeal and a WithIdleOverlay afterall as-is).

Good idea. This approach can be used to increase damage along with flame size as well.

mods/ts/rules/defaults.yaml
Mailaender: This is partly on purpose. See https://github.com/OpenRA/OpenRA/pull/11405#issuecomment-223821259.

reaperrr: In TS it was possible to 'chain' animations. The fire anims didn't play at the same time.

https://github.com/LipkeGu/TSIniFiles/blob/master/rules.ini#L354 is just a randomizer for the initial flame anim.

Mailaender: Oh, I thought they were overlayed. Then I probably chose the wrong SHPs as well.

MustaphaTR:
Needs Rebase.

Mailaender:
Rebased and vastly improved according to the suggestions made.

OpenRA.Mods.Common/Traits/Burns.cs
```diff
@@ -9,50 +9,62 @@
...
+			: base(info)
 		{
 			this.info = info;
 
-			var anim = new Animation(self.World, info.Image, () => 0);
+			var anim = new Animation(self.World, info.Image, () => 0, () => IsTraitPaused);
 			anim.IsDecoration = true;
+			var animWithOffset = new AnimationWithOffset(anim, () => WVec.Zero, () => IsTraitDisabled, p => RenderUtils.ZOffsetFromCenter(self, p, 1));
+			self.Trait<RenderSprites>().Add(animWithOffset, info.Palette, info.IsPlayerPalette);
 			anim.PlayRepeating(info.Anim);
```
reaperrr: Shouldn't this be `animWithOffset` then? (note: I haven't checked, I'm just asking)

mods/ts/rules/defaults.yaml
```diff
@@ -951,6 +951,9 @@
 		Palette: effect
 		RequiresCondition: burning
 		DamageTypes: Incendiary
+	ExplosionOnDamageTransition
```
reaperrr: Missing `:` at the end.

reaperrr:
Trees are killed way too fast (you can probably up the HP to 2000 to get more original-like duration), flames are displayed way too low (they should be centered on the tree), need more transparency (I'd guess 50%) and fixed ZRamp.

OpenRA.Mods.Common/Traits/Burns.cs
Mailaender: I believe this also confused me, but I checked against existing code on how to use this.

Mailaender:
I updated this a bit, but I didn't got far with offsets. I don't know what ZRamp really is.

pchote:
Moving to the future milestone as discussed in IRC.

Mailaender:
Superseded by https://github.com/OpenRA/OpenRA/pull/18730 and https://github.com/OpenRA/OpenRA/pull/17987.

PR #11486, opened 2016-06-19
Remove LobbyDefaults from reservedModuleNames in Manifest.cs

abcdefg30:
It seems like LobbyDefaults was removed as part of #10839.


Phrohdoh:
👍

PR #11496, opened 2016-06-20
ProvidesPrerequisiteFromAllies

phrohdoh:
Inspired by [0AD](http://www.play0ad.com)'s [planned feature](https://github.com/0ad/0ad/blob/92bb12a56552429476bd6e2955cc6cf53e4cb51c/binaries/data/mods/public/simulation/data/civs/spart.json#L81-L85) (which will go by the name `Peloponnesian League` for their Spartan faction's "team bonus").

This allows providing a prereq to a player based of the faction of allied players.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Player/ProvidesPrerequisiteFromAllies.cs
+++ b/OpenRA.Mods.Common/Traits/Player/ProvidesPrerequisiteFromAllies.cs
@@ -1,0 +1,71 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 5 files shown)


OpenRA.Mods.Common/Traits/Player/ProvidesPrerequisiteFromAllies.cs
```diff
@@ -0,0 +1,70 @@
...
+				if (!enabled)
+					yield break;
+
+				yield return prerequisite;
+			}
+		}
+
+		void INotifyCreated.Created(Actor self)
+		{
+			self.World.AddFrameEndTask(world => {
```
pchote: Brace on its own line, please.

Phrohdoh:
Fixed the prereq lint.

obrakmann:
Looks overall fine to me, and it's small and self-contained enough for my taste to take it even without an immediate use-case.

However, I have one question: what should happen when the ally that provides the prerequisite is defeated/disconnects/whatever? At the moment, the availability is only checked right at the start of the game and persists forever.

pchote:
IMO it would make sense to disable it in that case.  This would be a good usecase for an `INotifyPlayerDefeated` interface.

Phrohdoh:
I can buy that, but I'll make it configurable.
Will do.

Phrohdoh:
I can't get this working the way I'd like and I'm quite busy with life right now, so I'm closing this.

Issue #11310, opened 2016-05-17
Stack overflow: IP: 0x41bcebe4, fault addr: (nil) crash


ika2:
release-20160508
CentOS Linux release 7.2.1511 (Core)
Application crashes when high count of forces (vehicles or infantry).

Loading mod: d2k
[2016-05-17T16:29:23] Game started
Stack overflow: IP: 0x40fc1c0f, fault addr: (nil)
Stacktrace:
  at OpenRA.Activities.Activity.Queue (OpenRA.Activities.Activity) <IL 0x00013, 0x00031>
  <...>
  at OpenRA.Actor.QueueActivity (OpenRA.Activities.Activity) <IL 0x0001f, 0x0005d>
  at OpenRA.Mods.D2k.Traits.Carryable.Dropped () <IL 0x0005b, 0x000e7>
  at OpenRA.Mods.D2k.Activities.DeliverUnit.Release () <IL 0x00061, 0x0014f>
  at OpenRA.Mods.D2k.Activities.DeliverUnit.Tick (OpenRA.Actor) <IL 0x0011a, 0x00237>
  at OpenRA.Traits.ActivityUtils.RunActivity (OpenRA.Actor,OpenRA.Activities.Activity) <IL 0x00024, 0x0005e>
  at OpenRA.Actor.Tick () <IL 0x00010, 0x0003f>
  at OpenRA.World.Tick () <IL 0x000c3, 0x00227>
  at OpenRA.Game.InnerLogicTick (OpenRA.Network.OrderManager) <IL 0x00226, 0x00673>
  at OpenRA.Game.LogicTick () <IL 0x00055, 0x000b3>
  at OpenRA.Game.Loop () <IL 0x000d4, 0x0014b>
  at OpenRA.Game.Run () <IL 0x00042, 0x0006f>
  at OpenRA.Program.Run (string[]) <IL 0x00011, 0x00053>
  at OpenRA.Program.Main (string[]) <IL 0x00050, 0x000db>
  at (wrapper runtime-invoke) <Module>.runtime_invoke_int_object (object,intptr,intptr,intptr) <IL 0x00064, 0xffffffff>
[x@ntbk ~]$ date
Tue May 17 16:53:34 CEST 2016

pchote:
What version of mono are you running? (`mono --version`, or reported in the exception.log)

ika2:
The exception.log file doesn't exist.
mono --version
[...]

PR #11501, opened 2016-06-21
Add Carryall/Carryable traits, merging ts and d2k carryall behavior

### LINKED ISSUE #11345: Skirmish battles crash after 30-45 minutes of gameplay
(follow up on #11333)

Good evening OpenRA developers,

As per obrakmann's request, I am submitting a new bug report with additional information about this problem.

To reiterate, when playing a skirmish battle in Dune 2000, the game crashes to desktop after about 30-45 minutes of gameplay.  

As it turns out, the game does not write any information to exception.log when this occurs.  Rather, the crash seems to be due to a .NET error, as two Application error logs are written into the Event Viewer after the crash occurs.

This instance of the crash occurred on a desktop computer running Windows XP Professional x64 with a GeForce GTX 280 (Driver Version 6.14.13.681, Driver Date 9/23/2012), but I have also encountered this same problem on a laptop computer running Windows 7 Professional x64 with a GeForce 9800M GTS (Driver Version 9.18.13.623, Driver Date 8/30/2012).

I have attached screenshots of the battle setup I used for this particular battle and the .NET Application errors I found in my Event Viewer.

Once again, thank you in advance for any help you can provide!
openra d2k game setup
event viewer error 1
event viewer error 2

pchote:
#10775 rebased and with some small variable naming tweaks.

Fixes #11345.

### DIFF
--- a/OpenRA.Mods.Common/Activities/DeliverUnit.cs
+++ b/OpenRA.Mods.Common/Activities/DeliverUnit.cs
@@ -187,6 +187,13 @@ namespace OpenRA.Mods.Common.Activities
 			var targetPosition = self.CenterPosition + body.LocalToWorld(localOffset);
 			var targetLocation = self.World.Map.CellContaining(targetPosition);
 			positionable.SetPosition(carryall.Carryable, targetLocation, SubCell.FullCell);
+
+			// HACK: directly manipulate the turret facings to match the new orientation
+			// This can eventually go away, when we make turret facings relative to the body
+			var facingDelta = carryallFacing.Facing - carryableFacing.Facing;
+			foreach (var t in carryall.Carryable.TraitsImplementing<Turreted>())
+				t.TurretFacing += facingDelta;
+
 			carryableFacing.Facing = carryallFacing.Facing;
 
 			// Put back into world
--- a/OpenRA.Mods.Common/Activities/DeliverUnit.cs
+++ b/OpenRA.Mods.Common/Activities/DeliverUnit.cs
@@ -9,6 +9,7 @@
  */
 #endregion
 
+using System.Linq;
 using OpenRA.Activities;
 using OpenRA.Mods.Common.Traits;
 using OpenRA.Traits;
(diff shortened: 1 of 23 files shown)


mods/d2k/rules/defaults.yaml
```diff
@@ -93,7 +93,7 @@
 	MustBeDestroyed:
 	Voiced:
 		VoiceSet: VehicleVoice
-	Carryable:
+	AutoCarryable:
```
pchote: This will want an upgrade rule.

OpenRA.Mods.Common/Traits/Carryall.cs
```diff
@@ -10,206 +10,314 @@
...
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Automatically transports harvesters with the Carryable trait between resource fields and refineries.")]
-	public class CarryallInfo : ITraitInfo, Requires<BodyOrientationInfo>
+	public class CarryallInfo : ITraitInfo, Requires<BodyOrientationInfo>, Requires<AircraftInfo>
 	{
-		[Desc("Set to false when the carryall should not automatically get new jobs.")]
-		public readonly bool Automatic = true;
+		public readonly int LoadingDelay = 0;
+		public readonly int UnLoadingDelay = 0;
```
pchote: This should change to `Unloading` for consistency.

pchote:
```
OpenRA.Mods.Common/Traits/Carryall.cs:L150: [SP2000] Invalid spacing at the end of the line.
OpenRA.Mods.Common/Traits/Carryall.cs:L159: [SP2000] Invalid spacing at the end of the line.
```

OpenRA.Mods.Common/Activities/DeliverUnit.cs
```diff
@@ -74,65 +78,104 @@ bool CanDropHere()
...
+					{
+						var altitude = self.World.Map.DistanceAboveTerrain(self.CenterPosition);
+						if ((altitude - carryable.CarryableHeight).Length < aircraft.Info.MinAirborneAltitude)
+							Game.Sound.Play(aircraft.Info.TakeoffSound);
+						return ActivityUtils.SequenceActivities(movement.MoveTo(targetl, 0), this);
+					}
 
 				case State.Land:
 					if (!CanDropHere())
+						return ChangeState(State.Transport);
```
pchote: It may be possible to sequence in a `Land` order here to reduce duplication.

OpenRA.Mods.Common/Activities/DeliverUnit.cs
```diff
@@ -74,65 +78,104 @@ bool CanDropHere()
...
 
 				case State.Release:
 					if (!CanDropHere())
+						return ChangeState(State.Transport);
+
+					Release();
+					return ChangeState(State.TakeOff);
+
+				case State.TakeOff:
+					if (HeliFly.AdjustAltitude(self, aircraft, aircraft.Info.CruiseAltitude))
```
pchote: Likewise here for taking off.

OpenRA.Mods.Common/Activities/PickupUnit.cs
```diff
@@ -19,87 +19,131 @@ public class PickupUnit : Activity
...
 						return this;
 					}
 
-					return ActivityUtils.SequenceActivities(movement.MoveTo(cargo.Location, 0), this);
+					return ChangeState(State.Land);
+				case State.Land: // Land on the carryable
+					if (self.Location != cargo.Location || carryallFacing.Facing != carryableFacing.Facing)
+						return ChangeState(State.MoveToCarryable);
+
+					if (HeliFly.AdjustAltitude(self, aircraft, aircraft.Info.LandAltitude + carryable.CarryableHeight))
```
pchote: Same here – move this to an innerActivity.

OpenRA.Mods.Common/Traits/AutoCarryable.cs
```diff
@@ -0,0 +1,159 @@
...
+#endregion
+
+using System.Linq;
+using OpenRA.Activities;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Can be carried by units with the trait `Carryall`.")]
```
pchote: `[Desc("The actor will automatically transport harvesters and service depot transport requests.")]` to match the description for `AutoTarget`.

OpenRA.Mods.Common/Traits/AutoCarryable.cs
```diff
@@ -0,0 +1,159 @@
...
+			if (state == State.Locked)
+				return;
+
+			WantsTransport = false;
+			afterLandActivity = null;
+
+			// TODO: We could implement something like a carrier.Trait<Carryall>().CancelTransportNotify(self) and call it here
+		}
+
+		long CalculateDistanceSquared(CPos destination)
```
pchote: TODO: Investigate this.  It looks a bit suspicious.

OpenRA.Mods.Common/Traits/AutoCarryable.cs
```diff
@@ -0,0 +1,159 @@
...
+	[Desc("Can be carried by units with the trait `Carryall`.")]
+	public class AutoCarryableInfo : CarryableInfo
+	{
+		[Desc("Required distance away from destination before requesting a pickup. Default is 6 cells.")]
+		public readonly WDist MinDistance = WDist.FromCells(6);
+
+		public override object Create(ActorInitializer init) { return new AutoCarryable(init.Self, this); }
+	}
+
+	public class AutoCarryable : Carryable, INotifyHarvesterAction, ICallForTransport
```
pchote: This can be renamed to something a bit more descriptive like `RequestsAutoCarryall`.

OpenRA.Mods.Common/Traits/AutoCarryall.cs
```diff
@@ -0,0 +1,121 @@
...
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Automatically transports harvesters with the Carryable trait between resource fields and refineries.")]
+	public class AutoCarryallInfo : CarryallInfo
+	{
+		[Desc("Set to false when the carryall should not automatically get new jobs.")]
+		public readonly bool Automatic = true;
```
pchote: This looks potentially redundant.

OpenRA.Mods.Common/Traits/AutoCarryall.cs
```diff
@@ -0,0 +1,121 @@
...
+						return false;
+
+					if (actor.IsIdle)
+						return false;
+
+					return true;
+				})
+				.OrderBy(p => (self.Location - p.Actor.Location).LengthSquared);
+
+			foreach (var p in carryables)
```
pchote: Why aren't we doing this on the carryables themselves?

OpenRA.Mods.Common/Traits/Carryable.cs
```diff
@@ -19,181 +17,89 @@ namespace OpenRA.Mods.Common.Traits
...
 	{
-		[Desc("Required distance away from destination before requesting a pickup. Default is 6 cells.")]
-		public readonly WDist MinDistance = WDist.FromCells(6);
-
 		[UpgradeGrantedReference]
 		[Desc("The upgrades to grant to self while waiting or being carried.")]
 		public readonly string[] CarryableUpgrades = { };
 
-		public object Create(ActorInitializer init) { return new Carryable(init.Self, this); }
+		public readonly WDist CarryableHeight = new WDist(200);
```
pchote: This probably wants to change to a more general `WVec CarryallOffset`

OpenRA.Mods.Common/Traits/Carryable.cs
```diff
@@ -19,181 +17,89 @@ namespace OpenRA.Mods.Common.Traits
...
+		protected enum State { Free, Reserved, Locked }
+		protected State state;
+		protected bool attached = false;
 
 		public Carryable(Actor self, CarryableInfo info)
 		{
 			this.info = info;
 			this.self = self;
 			upgradeManager = self.Trait<UpgradeManager>();
+			Destination = CPos.Zero;
```
pchote: Using `CPos.Zero` as a magic value has caused many problems in the past.  It may be better to change this to a `CPos?` and initialize it to null.

OpenRA.Mods.Common/Traits/Carryall.cs
```diff
@@ -10,206 +10,314 @@
...
 
-		public object Create(ActorInitializer init) { return new Carryall(init.Self, this); }
+		[VoiceReference]
+		public readonly string Voice = "Move";
```
pchote: `Action` might be a better default.

OpenRA.Mods.Common/Traits/Carryall.cs
```diff
@@ -10,206 +10,314 @@
...
-				return true;
+				if (Carryable.IsInWorld && !Carryable.IsDead)
+					Carryable.Kill(e.Attacker);
+				Carryable = null;
 			}
 
+			UnreserveCarryable();
+		}
+
+		public virtual bool RequestTransportNotify(Actor carryable)
```
pchote: I'm not sure it makes sense to define this here - it probably wants to stay inside `AutoCarryall`.

pchote:
Issuing a stop order while the carryall is landing to pick up another actor will make it stop mid-air.

OpenRA.Mods.Common/Activities/DeliverUnit.cs
```diff
@@ -69,35 +72,82 @@ CPos GetLocationToDrop(CPos requestedPosition)
...
+
+			if ((carryall.State == Carryall.CarryallState.Idle || carryall.Carryable.IsDead) && state != State.TakeOff)
+				state = State.Aborted;
+
 			switch (state)
 			{
 				case State.Transport:
-					var targetl = GetLocationToDrop(carryable.Destination);
+				{
+					var targetLocation = GetLocationToDrop(location);
```
pchote: `GetLocationToDrop` needs to be rewritten too.

mods/d2k/rules/defaults.yaml
pchote: On second thought an upgrade rule isn't helpful.  This trait was only available to d2k in the last release, and both uses of the traits there are exceptions to what the automated rule could do.

pchote:
This should also support force-move to move without dropping the cargo.

pchote:
Updated.  I've overhauled this to fix my comments above plus a whole collection of other issues that carried over from the old carryall code.  This should now be a lot more robust, so this may fix #11310 (but without a repro case I can't guarantee it).

pchote:
Rebased and added the confirmed fix for #11345.

obrakmann:
There's an infinite loop in here somewhere. I got it repeatedly during playtesting.

pchote:
Can you add a `writer.Flush()` to the end of `ReplayRecorder.Receive` and then upload a replay with the looping?

The replay won't be replayable until a metadata block is manually tacked on the end, but I can do that.

obrakmann:
It's looping in `RunActivity` with `FindResources` as the current activity.

OpenRA.Mods.Common/Activities/FindResources.cs
```diff
@@ -66,39 +66,43 @@ public override Activity Tick(Actor self)
 			if (!closestHarvestablePosition.HasValue)
 			{
 				if (!harv.IsEmpty)
-					return deliver;
+					return ActivityUtils.SequenceActivities(deliver, this);
```
pchote: This change is the cause of @obrakmann's infinite loop.  DeliverResources has a check at the start of its tick to do nothing and return the next activity is that is non-null, so this will always loop.

pchote:
Fixed by reverting the change that I commented on inline.

OpenRA.Mods.Common/Traits/Carryall.cs
```diff
@@ -10,206 +10,336 @@
...
 using OpenRA.Mods.Common.Activities;
-using OpenRA.Mods.Common.Traits.Render;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Mods.Common.Orders;
+using OpenRA.Primitives;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Automatically transports harvesters with the Carryable trait between resource fields and refineries.")]
```
GraionDilach: Description update required.

obrakmann:
I've seen a bug where harvesters keep getting stuck in the `Wait` activity after a carryall drops them off in front of the refinery. They stay there until another harvester comes along to nudge them out of their way, then they deliver the resources after the other harvester is gone.

pchote:
@obrakmann isn't that an existing problem?

obrakmann:
Yeah, it seems it is.

obrakmann:
OK, looks fine then, :+1: 

What I noticed is that you can pick up a deployed tick tank, but that can be fixed with some upgrade magic in the yaml rules.

pchote:
Is the `Fixup requested` for the tick tank and sensor array?

obrakmann:
That and the `Desc` that Graion mentioned

pchote:
"some upgrade magic in the yaml rules." &rarr; adding upgrade support to `Carryable` and `DeployToUpgrade`.  This isn't trivial, so fixup &rarr; changes requested.

pchote:
Rebased and fixed trait descriptions.  As discussed in IRC we'll deal with (not-) transporting deployed actors in a future PR to avoid massive scope creep here.

pchote:
Rebased again.  Lets not forget that this is a release blocker!

pchote:
Fixed.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/9dce860e3395263f3cfb801827efc49a5951e9a6...7c368955001728b432f79a539f2a50a79b14b15c)

Issue #11503, opened 2016-06-22
ProductionParadrop issues


Nelax:
I just tried using this trait with a moving unit and even with a deployable version there would be issues. Primarily, the unit spawns where the producer currently is instead of where the play flies over which is where ever the producer was when it finished production. If I made it into a deployable version, basically where it deploys into the factory, you run into the issue of having a crash (most likely, I didn't actually test this) when the trait tries to access where the factory when it's been undeployed. It would be referencing a null actor.

PR #11504, opened 2016-06-22
Improved ProductionParadrop plane targeting

Nelax:
Fixes issues with ProductionParadrop by making the producer actor the target instead of it's location when production is finished. This allows it to handle mobile actors far better. Only trading off somewhat questionable logic involving aligning the plane with the exit location.

Fixes the primary issue raised in #11503.

EDIT: As pchote pointed out. The exit really needs to be taken into proper account instead of how it's handled here.

### DIFF
--- a/OpenRA.Mods.Common/Traits/ProductionParadrop.cs
+++ b/OpenRA.Mods.Common/Traits/ProductionParadrop.cs
@@ -53,9 +53,8 @@ namespace OpenRA.Mods.Common.Traits
 
 			// Start a fixed distance away: the width of the map.
 			// This makes the production timing independent of spawnpoint
-			var dropPos = self.Location + exit.ExitCell;
-			var startPos = dropPos + new CVec(owner.World.Map.Bounds.Width, 0);
-			var endPos = new CPos(owner.World.Map.Bounds.Left - 5, dropPos.Y);
+			var startPos = self.Location + new CVec(owner.World.Map.Bounds.Width, 0);
+			var endPos = new CPos(owner.World.Map.Bounds.Left - 5, self.Location.Y);
 
 			foreach (var notify in self.TraitsImplementing<INotifyDelivery>())
 				notify.IncomingDelivery(self);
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -53,9 +53,8 @@ public override bool Produce(Actor self, ActorInfo producee, string factionVaria
 
 			// Start a fixed distance away: the width of the map.
 			// This makes the production timing independent of spawnpoint
-			var dropPos = self.Location + exit.ExitCell;
-			var startPos = dropPos + new CVec(owner.World.Map.Bounds.Width, 0);
-			var endPos = new CPos(owner.World.Map.Bounds.Left - 5, dropPos.Y);
+			var startPos = self.Location + new CVec(owner.World.Map.Bounds.Width, 0);
```
pchote: :-1: to making this ignore the exit (which is a standard part of the production code) and always targeting the top-left cell of the actor.  You should be able to resolve the exit location when you need it, instead.

Nelax:
I'll work this out sometime in the future. Agreeing with Paul here, the exit is probably more important to remain intact than a proper drop zone. Most cases will be structures of large sizes which will need a variable offset. My case is very specific. I'll revisit this when my mod is further along in production.

OpenRA.Mods.Common/Traits/ProductionParadrop.cs
```diff
@@ -76,7 +75,7 @@ public override bool Produce(Actor self, ActorInfo producee, string factionVaria
 					new FacingInit(64)
 				});
 
-				actor.QueueActivity(new Fly(actor, Target.FromCell(w, dropPos)));
+				actor.QueueActivity(new Fly(actor, Target.FromActor(self)));
```
GraionDilach: You coulda try Target.FromPos with the exit offset and the actor location.

Nelax: I'm not quite sure if there is all that much difference between FromCell and FromPos. The problem is that by doing anything based on the actor's location, it's going to be outdated almost instantly if the actor is moving. Really what needs to be done is pass the plane somehow to the doProduction method.

PR #11510, opened 2016-06-23
Move audio loaders from engine to mod level

reaperrr:
There wasn't anything significant blocking this move, so audio loaders now follow sprite loaders.

Bag loading of wav/ima in RA2 needs to be tested, just in case. 
cc @GraionDilach @abcdefg30 @Mailaender


Mailaender:
This doesn't regress anything for RA2. :white_check_mark:

Mailaender:
You should change the namespace to `OpenRA.Mods.Common.FileFormats`.

reaperrr:
Updated.

However, in order to make things more consistent, I moved the *Loaders to `AudioLoaders` namespace while splitting the readers to `FileFormats`.

Not entirely sure what to do with `VocLoader`, though, so I left it as-is for now.

OpenRA.Mods.Common/AudioLoaders/AudLoader.cs
```diff
@@ -0,0 +1,88 @@
...
+			this.stream = stream;
+
+			var position = stream.Position;
+			rawData = Exts.Lazy(() =>
+			{
+				try
+				{
+					byte[] data;
+					if (!AudReader.LoadSound(stream, out data, out sampleRate))
+						throw new InvalidDataException();
```
Phrohdoh: An empty exception isn't helpful.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ff09d82e581286c147aa20f0f9b0b143ae46419a)

PR #11511, opened 2016-06-23
D2k: Add Cliffmaze 1v1 map by S.D.

OmegaBolt:
cliffmaze

This is a small 1v1 map with a maze-like cliff layout between bases. Lots of sand dunes make infantry very useful, and the short distance between bases should make for a very intense tug of war.

I merely fixed some tile issues, so if there are any then let me know. SD gave permission for this to be included in ORA.


Mailaender:
Comes with a good amount of detail and no tiling errors. :white_check_mark: Nice find!

obrakmann:
Looking good, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c67fbe7ffa1fffe29a9038425762415ab90e3774)

PR #11512, opened 2016-06-24
Added Squad Exclusion to HackyAI

Nelax:
Thanks to GraionDilach for pointing this out to me but HackyAI will exclude MCVs and harvesters from squads but otherwise will add any available units to squads including possibility unarmed utility units like the Manticore faction in my mod, Phoenix Chronicles, has. So I have an AI that will send power-generating unarmed trucks and it's HQ among other base-units into battle. This gives a modder a way to make units with the sole purpose of being around the base. Could be useful for the mobile repair vehicle for Nod in TS.

No test case included, but I did test it before pushing by adding like 90% of the unit roster to the exclusion list. Nothing included appeared to be put into a squad.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -27,6 +27,7 @@ namespace OpenRA.Mods.Common.AI
 		public class UnitCategories
 		{
 			public readonly HashSet<string> Mcv = new HashSet<string>();
+			public readonly HashSet<string> ExcludeFromSquads = new HashSet<string>();
 		}
 
 		public class BuildingCategories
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -27,7 +27,8 @@ public sealed class HackyAIInfo : IBotInfo, ITraitInfo
 		public class UnitCategories
 		{
 			public readonly HashSet<string> Mcv = new HashSet<string>();
-		}
+            public readonly HashSet<string> ExcludeFromSquads = new HashSet<string>();
+        }
```
abcdefg30: Please use tabs instead of spaces.

OpenRA.Mods.Common/AI/HackyAI.cs
```diff
@@ -678,7 +679,8 @@ void GiveOrdersToIdleHarvesters()
 		void FindNewUnits(Actor self)
 		{
 			var newUnits = self.World.ActorsHavingTrait<IPositionable>()
-				.Where(a => a.Owner == Player && !Info.UnitsCommonNames.Mcv.Contains(a.Info.Name) && !activeUnits.Contains(a));
+				.Where(a => a.Owner == Player && !Info.UnitsCommonNames.Mcv.Contains(a.Info.Name) &&
+				!Info.UnitsCommonNames.ExcludeFromSquads.Contains(a.Info.Name) && !activeUnits.Contains(a));
```
Phrohdoh: This line needs another level of indentation.

GraionDilach:
:+1:

reaperrr:
I think this good enough until we get around to refactoring squads. :+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/920d054fe55ffac01a89e08aab7b4aef8c4c692d...6226a8550e1acc6691c5afc29eedc4ac41b3029e)

Issue #11383, opened 2016-05-31
PoisonedByTiberium can crash ScriptTriggers.Damaged


Mailaender:
When testing https://github.com/OpenRA/OpenRA/pull/11368 I got a crash after attacking the lone grenadier. He came after me, ran over Tiberium and that causes it to be damaged by a null actor location:

```
Fatal Lua Error: [string "BindingSupport.lua"]:30: Uncaught CLR exception at Lua->CLR boundary: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an object
  at OpenRA.Actor.get_Location () [0x00007] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Actor.cs:60 
  at OpenRA.Mods.Common.Scripting.GeneralProperties.get_Location () <0x41592890 + 0x0001b> in <filename unknown>:0 
  at (wrapper delegate-invoke) <Module>:invoke_callvirt_CPos_GeneralProperties (OpenRA.Mods.Common.Scripting.GeneralProperties)
  at System.Reflection.MonoProperty.GetterAdapterFrame[T,R] (System.Reflection.Getter`2 getter, System.Object obj) [0x00000] in /home/abuild/rpmbuild/BUILD/mono-4.2.3/mcs/class/corlib/System.Reflection/MonoProperty.cs:329 
  at System.Reflection.MonoProperty.GetValue (System.Object obj, System.Object[] index) [0x0008e] in /home/abuild/rpmbuild/BUILD/mono-4.2.3/mcs/class/corlib/System.Reflection/MonoProperty.cs:397 
  --- End of inner exception stack trace ---
  at System.Reflection.MonoProperty.GetValue (System.Object obj, System.Object[] index) [0x000a1] in /home/abuild/rpmbuild/BUILD/mono-4.2.3/mcs/class/corlib/System.Reflection/MonoProperty.cs:399 
  at OpenRA.Scripting.ScriptMemberWrapper.Get (Eluant.LuaRuntime runtime) [0x00041] in /home/matthias/Entwicklung/OpenRA/OpenRA.Game/Scripting/ScriptMemberWrapper.cs:104 
[...]

PR #11513, opened 2016-06-24
Remove passing the warheads from DamageWarhead to AttackInfo.

GraionDilach:
Added an IDamage wrapper interface and a DamageObject implementation to pass damage(types) instead.
This removes a great amount of overhead and longterm opens possibilities to have damagetypes without warheads.

### DIFF
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -41,7 +41,7 @@ namespace OpenRA.Traits
 		int DisplayHP { get; }
 		bool IsDead { get; }
 
-		void InflictDamage(Actor self, Actor attacker, int damage, IWarhead warhead, bool ignoreModifiers);
+		void InflictDamage(Actor self, Actor attacker, Damage damage, bool ignoreModifiers);
 		void Kill(Actor self, Actor attacker);
 	}
 
@@ -82,11 +82,28 @@ namespace OpenRA.Traits
 
 	public class AttackInfo
 	{
-		public int Damage;
+		public Damage Damage;
 		public Actor Attacker;
-		public IWarhead Warhead;
 		public DamageState DamageState;
 		public DamageState PreviousDamageState;
+	}
+
+	public class Damage
+	{
+		public readonly int Value;
+		public readonly HashSet<string> DamageTypes;
+
+		public Damage(int damage, HashSet<string> damageTypes)
+		{
+			Value = damage;
+			DamageTypes = damageTypes;
+		}
+
+		public Damage(int damage)
+		{
+			Value = damage;
+			DamageTypes = new HashSet<string>();
+		}
 	}
 
 	public interface ITick { void Tick(Actor self); }
@@ -249,7 +266,7 @@ namespace OpenRA.Traits
 	}
 
 	public interface IRenderModifier { IEnumerable<IRenderable> ModifyRender(Actor self, WorldRenderer wr, IEnumerable<IRenderable> r); }
-	public interface IDamageModifier { int GetDamageModifier(Actor attacker, IWarhead warhead); }
+	public interface IDamageModifier { int GetDamageModifier(Actor attacker, Damage damage); }
 	public interface ISpeedModifier { int GetSpeedModifier(); }
 	public interface IFirepowerModifier { int GetFirepowerModifier(); }
 	public interface IReloadModifier { int GetReloadModifier(); }
--- a/OpenRA.Mods.Common/Scripting/Properties/HealthProperties.cs
+++ b/OpenRA.Mods.Common/Scripting/Properties/HealthProperties.cs
@@ -29,7 +29,7 @@ namespace OpenRA.Mods.Common.Scripting
 		public int Health
 		{
 			get { return health.HP; }
-			set { health.InflictDamage(Self, Self, health.HP - value, null, true); }
+			set { health.InflictDamage(Self, Self, new Damage(health.HP - value), true); }
 		}
 
 		[Desc("Maximum health of the actor.")]
@@ -38,7 +38,7 @@ namespace OpenRA.Mods.Common.Scripting
 		[Desc("Kill the actor.")]
 		public void Kill()
 		{
-			health.InflictDamage(Self, Self, health.MaxHP, null, true);
+			health.InflictDamage(Self, Self, new Damage(health.MaxHP), true);
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Traits/Infantry/TakeCover.cs
+++ b/OpenRA.Mods.Common/Traits/Infantry/TakeCover.cs
@@ -56,8 +56,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void Damaged(Actor self, AttackInfo e)
 		{
-			var warhead = e.Warhead as DamageWarhead;
-			if (e.Damage <= 0 || warhead == null || !warhead.DamageTypes.Overlaps(info.DamageTriggers))
+			if (e.Damage.Value <= 0 || !e.Damage.DamageTypes.Overlaps(info.DamageTriggers))
 				return;
 
 			if (!IsProne)
@@ -79,16 +78,15 @@ namespace OpenRA.Mods.Common.Traits
 			get { return true; }
 		}
 
-		public int GetDamageModifier(Actor attacker, IWarhead warhead)
+		public int GetDamageModifier(Actor attacker, Damage damage)
 		{
 			if (!IsProne)
 				return 100;
 
-			var damageWh = warhead as DamageWarhead;
-			if (damageWh == null)
+			if (damage.DamageTypes.Count == 0)
 				return 100;
 
-			var modifierPercentages = info.DamageModifiers.Where(x => damageWh.DamageTypes.Contains(x.Key)).Select(x => x.Value);
+			var modifierPercentages = info.DamageModifiers.Where(x => damage.DamageTypes.Contains(x.Key)).Select(x => x.Value);
 			return Util.ApplyPercentageModifiers(100, modifierPercentages);
 		}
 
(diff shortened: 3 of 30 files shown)


OpenRA.Mods.Common/Traits/Infantry/TakeCover.cs
```diff
@@ -56,8 +56,7 @@ public TakeCover(ActorInitializer init, TakeCoverInfo info)
 
 		public void Damaged(Actor self, AttackInfo e)
 		{
-			var warhead = e.Warhead as DamageWarhead;
-			if (e.Damage <= 0 || warhead == null || !warhead.DamageTypes.Overlaps(info.DamageTriggers))
+			if (e.Damage.Damage <= 0 || e.Damage.DamageTypes == null || !e.Damage.DamageTypes.Overlaps(info.DamageTriggers))
```
abcdefg30: Redundant null check once you replaced the default.

OpenRA.Mods.Common/Traits/Infantry/TerrainModifiesDamage.cs
```diff
@@ -41,10 +41,9 @@ public TerrainModifiesDamage(Actor self, TerrainModifiesDamageInfo info)
 			this.self = self;
 		}
 
-		public int GetDamageModifier(Actor attacker, IWarhead warhead)
+		public int GetDamageModifier(Actor attacker, IDamage damage)
 		{
-			var damageWh = warhead as DamageWarhead;
-			if (attacker.Owner.IsAlliedWith(self.Owner) && (damageWh != null && damageWh.Damage < 0) && !Info.ModifyHealing)
+			if (attacker.Owner.IsAlliedWith(self.Owner) && (damage.DamageTypes != null && damage.Damage < 0) && !Info.ModifyHealing)
```
abcdefg30: You'll need to modify this check as well then.

OpenRA.Mods.Common/Traits/Render/WithDamageOverlay.cs
```diff
@@ -55,12 +55,11 @@ public WithDamageOverlay(Actor self, WithDamageOverlayInfo info)
 
 		public void Damaged(Actor self, AttackInfo e)
 		{
-			var warhead = e.Warhead as DamageWarhead;
-			if (info.DamageTypes.Count > 0 && (warhead != null && !warhead.DamageTypes.Overlaps(info.DamageTypes)))
+			if (info.DamageTypes.Count > 0 && (e.Damage.DamageTypes != null && !e.Damage.DamageTypes.Overlaps(info.DamageTypes)))
```
abcdefg30: Same here.

OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs
```diff
@@ -84,7 +84,7 @@ public void Killed(Actor self, AttackInfo e)
 				palette += self.Owner.InternalName;
 
 			// Killed by some non-standard means
-			if (e.Warhead == null || !(e.Warhead is DamageWarhead))
+			if (e.Damage.DamageTypes == null)
```
abcdefg30: And here.

OpenRA.Mods.Common/Traits/Sound/DeathSounds.cs
```diff
@@ -39,11 +39,8 @@ public class DeathSounds : INotifyKilled
 
 		public void Killed(Actor self, AttackInfo e)
 		{
-			var warhead = e.Warhead as DamageWarhead;
-
-			// If the warhead is null, the actor was killed by some non-standard means
-			if (info.DeathTypes.Count == 0 || (warhead != null && warhead.DamageTypes.Overlaps(info.DeathTypes)))
+			if (info.DeathTypes.Count == 0 || (e.Damage.DamageTypes != null && e.Damage.DamageTypes.Overlaps(info.DeathTypes)))
```
abcdefg30: And here.

pchote:
I don't have the time or head-space to review the design aspects of this in the immediate future.

Please don't merge this solely on whether or not the patch works until *somebody* considers how this fits into the wider engine design.  This doesn't immediately sit well with me, but I can't commit the time to explore the problems or their possible soutions.

GraionDilach:
I think I am actually avoiding a duplication. Overloading DamageWarhead with all this could be another option but that just maintains the overhead which is the main reason why PoisonedByTiberium is messy as well (something which longterm should be replaced with SelfHealing + UpgradeOnTerrain).

pchote:
Your `IDamage` interface is exactly what `IWarhead` is *supposed* to do.  We have a bunch of leftover null-warheads because of crufty legacy code that nobody has bothered to clean up.

GraionDilach:
`IWarhead` cannot do this. We have only 2(+1 abstract) warheads which delivers damage and 5 (+2 in my AS DLL and both should be merged upstream in the longterm) which does not. Nevertheless, all usecases of this at the receiving end only care about damagetypes (with only one case looking at the damage value, again a DamageWarhead-derivate) which means all the other cases are redundant.

pchote:
My point is that `IWarhead` probably needs to be cleaned up so that it *can*.

GraionDilach:
IWarhead cannot do this. All the IWarhead properties apply to all the warheads (delay, etc) and has nothing to do with actually damaging actors.

pchote:
I don't buy that at all, but as mentioned above won't be able to argue about it for the next few days.  Someone else who we trust with engine plumbing design will need to continue with the design review.

GraionDilach:
IWarhead is used for warhead target validation, timing and general activation through DoImpact. All cases affect more than damagewarheads (GrantUpgradeWarhead is a great example which needs these as well) while any trait code interacting with a warhead activation only cares about the damaging ones, and selfheal, tibdamage etc will also only need to duplicate that aspect of their behaviour, not the entire warhead activation procedure.

GraionDilach:
Updated.

pchote:
For the sake of continuing the discussion / review lets turn around the above discussion.  Please sell us on why we should take this specific implementation (elaborate on "This removes a great amount of overhead and longterm opens possibilities to have damagetypes without warheads.").

GraionDilach:
DamageTypes without warheads eliminates the need of having an entire weapon on `PoisonedByTiberium`/#11499's `DamagedByTerrain`.

GraionDilach:
That and I have a followup PR for selfhealing with damagetypes.

GraionDilach:
This pretty much would be the first step towards a solution you describe in https://github.com/OpenRA/OpenRA/issues/11383#issuecomment-222801775 even. Also see #10091.

obrakmann:
The difference between the solution that Paul envisions and the one that you've written is the level they're working at. Your solution basically decouples `Health` from dealing with warheads, while what Paul really wants (as far as I understood it) would decouple - or split, more correctly - warheads and weapons. Ie. the only warheads that would remain after would be `DamageWarhead` and its children, while the rest would evolve to become some kind of effect that could be spawned by weapons, warheads or anything else, really. @pchote, correct me if I stated this wrongly in any way.

Now, the question is whether this PR stands in the way of such a future rewrite. The code added in this PR would certainly not survive it, but I haven't really made up my mind yet (since I'm not sure what that rewrite would look like exactly) whether it would block it.

pchote:
@obrakmann that's about right.  I think this solves more than it hurts, so on balance I think we're better taking this than not (but the eventual total rewrite is going to really suck).  Lets rename `DamageObject` to something more sensible first, though, please.

GraionDilach:
That was the most sensible name option I could come up with.

obrakmann:
Alright, :+1: then. I don't have any good suggestion for another name, either. Maybe just `Damage`?

pchote:
I think the root of the problem is that `IDamage` is too specific as-is to warrant multiple implementations.  If `IDamage.Damage` changed to take arguments then you could have e.g. `ConstantDamage` vs `PercentDamage`.  Otherwise this as-is would be better done with a `Damage` struct.

GraionDilach:
If I go with the Damage struct, can it be placed to Actor(Exts).cs? I don't know where would such intermediate objects be placed in the source tree.

pchote:
It could live in TraitsInterfaces next to `AttackInfo`.

GraionDilach:
Updated.

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -82,13 +82,30 @@ public enum ImpactType
...
+		public Damage Damage;
 		public Actor Attacker;
-		public IWarhead Warhead;
 		public DamageState DamageState;
 		public DamageState PreviousDamageState;
 	}
 
+	public class Damage
+	{
+		public int Value;
```
pchote: Please make these `readonly`.

GraionDilach: https://github.com/OpenRA/OpenRA/pull/11513/files#diff-e107107188001ae1afb408ff1ad5821cR148

I can make the types readonly though.

GraionDilach: On a second thought, if you're fine remaking the object there then I think I can make both of 'em readonly.

pchote: Yes, please do.

GraionDilach: Done.

OpenRA.Mods.Common/Scripting/Properties/HealthProperties.cs
```diff
@@ -29,7 +29,7 @@ public HealthProperties(ScriptContext context, Actor self)
 		public int Health
 		{
 			get { return health.HP; }
-			set { health.InflictDamage(Self, Self, health.HP - value, null, true); }
+			set	{ health.InflictDamage(Self, Self, new Damage(health.HP - value), true); }
```
abcdefg30: The space to tab change (after `set`) looks wrong.

GraionDilach: Thanks, updated.

Mailaender:
No changelog as this is an internal API change.

PR #11515, opened 2016-06-24
Fix Wanders always triggering move at first game tick

reaperrr:
Currently, all actors with `AttackWander` or some other `Wanders`-derivative wander at the first tick because `countdown` has not yet been set to a random value when the first `TickIdle` runs.

I added a `firstTick` check so `TickIdle` won't do anything before the 2nd tick, ensuring `countdown` has been set properly by that time.

Additionally made `TickIdle` virtual so traits like `AttackWander` can override it, if necessary.


abcdefg30:

obrakmann:
More accurately, `OnBecomingIdle` isn't called when a unit starts idle.

:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8e6e960b00c53807da26ef2bffd2fd6918f8c8f8)

PR #11517, opened 2016-06-24
Add a note in the tooltip of build-limited TS structures

### LINKED ISSUE #7349: EMP Cannon should act as SupportPower (EMPSpecial)

abcdefg30:
~~Also updates the tooltip(s) to reflect that.~~

~~Closes #7349.~~

EDIT:

> Now I only adjust the tooltip of buildings with a buildlimit.

### DIFF
--- a/mods/ts/rules/nod-structures.yaml
+++ b/mods/ts/rules/nod-structures.yaml
@@ -405,7 +405,7 @@ NAWAST:
 		Cost: 1600
 	Tooltip:
 		Name: Waste Refinery
-		Description: Processes Veins\ninto useable resources.
+		Description: Processes Veins\ninto useable resources.\nMaximum 1 can be built.
 	Buildable:
 		BuildPaletteOrder: 110
 		Prerequisites: namisl, ~structures.nod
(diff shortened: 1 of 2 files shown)


MustaphaTR:
:-1: This is not how Vanilla E.M.Pulse cannon works. I was always against the build limit and still is. More than one Cannon should be buildable but only give range support not extra SW to launch.

Mailaender:
This seems to be a misunderstanding. https://github.com/OpenRA/OpenRA/issues/7349 probably only suggested the build limit to workaround the fact that it wasn't a support power back then.

abcdefg30:
Updated. Now I only adjust the tooltip of buildings with a buildlimit.

Mailaender:
Okay. :+1:

GraionDilach:
:+1:

PR #11518, opened 2016-06-24
Convert HackyAI BuildingFractions and UnitsToBuild to int percentages

reaperrr:
As far as I can tell, this is the last simulation-related place in our engine still using `float`.
For consistency and accuracy, we should use integer percentages here, too.

All mods except RA were upgraded with the provided upgrade rule.


Mailaender:
This can safely still use real percentages as the AI calculation is always done on one bot controlling client so there is no danger of desyncs.

reaperrr:
Ok, in that case there's no point in giving up the more fine-grained float/% values.

pchote:
> As far as I can tell, this is the last simulation-related place in our engine still using float.

Health.DamageState still uses floats.

PR #7222, opened 2014-12-29
Add Wanders trait and make AttackWander inherit that

penev92:
The idea is to make wandering more generic. Now we can have both AttackMove-wandering and simple Move-wandering.
I'm aware that currently only `AttackWander` inherits the new `Wanders` trait, but I have another use for it on another branch that will wait for this to be merged.

PR #11520, opened 2016-06-25
Make civilians wander around when idle

reaperrr:
Via new `IdleWander` trait.


Mailaender:

OpenRA.Mods.Common/Traits/AttackWander.cs
```diff
@@ -27,7 +27,7 @@ class AttackWander : Wanders
 		public AttackWander(Actor self, AttackWanderInfo info)
 			: base(self, info)
 		{
-			attackMove = self.TraitOrDefault<AttackMove>();
+			attackMove = self.Trait<AttackMove>();
```
Mailaender: :+1:

Mailaender:
Do `CHAN` and `DELPHI` in the `ra` mod also need the disablement?

reaperrr:
> Do CHAN and DELPHI in the ra mod also need the disablement?

`DELPHI` in RA actually has it disabled.

I don't remember any RA missions with `CHAN`, and he was wandering in original TD as well.
Also, mappers can still disable the wandering in map rules, if really necessary.

OpenRA.Mods.Common/Traits/IdleWander.cs
```diff
@@ -0,0 +1,38 @@
...
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Will Move to a random location within MoveRadius when idle.",
```
pchote: Why does this need its own trait? This is what `Wanders` is supposed to do!

Mailaender: I agree that the broken by design trait https://github.com/OpenRA/OpenRA/pull/7222/files#diff-e2c08dd74a071ebf738fa2aaa9ba81dcR86 is terrible. This might be an opportunity to fix it.

reaperrr:
> Why does this need its own trait? This is what `Wanders` is supposed to do!

Updated accordingly.

While I was at it, I also made the trait upgradable and support both `Mobile` and `Aircraft` actors.

OpenRA.Mods.Common/Traits/Wanders.cs
```diff
@@ -28,36 +28,49 @@ public abstract class WandersInfo : ITraitInfo
...
+		public override object Create(ActorInitializer init) { return new Wanders(init.Self, this); }
 	}
 
-	public class Wanders : INotifyIdle, INotifyBecomingIdle
+	public class Wanders : UpgradableTrait<WandersInfo>, INotifyCreated, INotifyIdle, INotifyBecomingIdle
 	{
 		readonly Actor self;
 		readonly WandersInfo info;
+		Mobile mobile;
+		Aircraft aircraft;
```
pchote: You may be able to store `IResolveOrder move = self.TraitOrDefault<IMove> as IResolveOrder` to avoid the explicit mobile/aircraft.

reaperrr:
Updated.

Mailaender:
Even better now. :+1:

OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
```diff
@@ -277,7 +277,7 @@
     <Compile Include="Traits\Attack\AttackOmni.cs" />
     <Compile Include="Traits\Attack\AttackSuicides.cs" />
     <Compile Include="Traits\Attack\AttackTurreted.cs" />
-    <Compile Include="Traits\Attack\AttackWander.cs" />
+    <Compile Include="Traits\AttackWander.cs" />
```
Phrohdoh: Why was this moved?

OpenRA.Mods.Common/Traits/Wanders.cs
```diff
@@ -28,36 +28,47 @@ public abstract class WandersInfo : ITraitInfo
...
-		public abstract object Create(ActorInitializer init);
+		public override object Create(ActorInitializer init) { return new Wanders(init.Self, this); }
 	}
 
-	public class Wanders : INotifyIdle, INotifyBecomingIdle
+	public class Wanders : UpgradableTrait<WandersInfo>, INotifyCreated, INotifyIdle, INotifyBecomingIdle
 	{
 		readonly Actor self;
 		readonly WandersInfo info;
+		IResolveOrder move;
```
Phrohdoh: This is... odd.
Why do you need `IMove` as an `IResolveOrder`, and can you guarantee every `IMove` is an `IResolveOrder`?

OpenRA.Mods.Common/Traits/Wanders.cs
```diff
@@ -92,7 +103,7 @@ CPos PickTargetLocation()
 
 		public virtual void DoAction(Actor self, CPos targetCell)
 		{
-			throw new NotImplementedException("Base class Wanders does not implement method DoAction!");
+			move.ResolveOrder(self, new Order("Move", self, false) { TargetLocation = targetCell });
```
Phrohdoh: Shouldn't you issue this order instead of directly resolving it?

pchote: This is synced code (runs on every client), so issuing an order won't work.

OpenRA.Mods.Common/Traits/Wanders.cs
reaperrr: This approach was suggested by @pchote to avoid having to explicitly look up both `Mobile` and `Aircraft`.

pchote: See the earlier comments for context.

OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
reaperrr: Because the trait name is misleading, this is not an `AttackBase` derivative and not strictly an attack trait by itself, either.

obrakmann:
Looks fine to me, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/b9dab0b6c8a8c9a026aa1a128ad81b2779fd3681)

Issue #2344, opened 2012-07-28
Support Dune 2000 patch 1.06 gamefiles


Mailaender:
The http://dune2k.com/Download/30 DATA.R8 from Dune 2000 version 1.06 patch has more frames and currently fails to extract. Using code from http://code.google.com/p/dune2000plusone/source/browse/trunk/Dune2k1/ResourceEditor/ResourceEditor/Form1.cs with explicit permission by mvi and a very permissive license might help. It also adds new terrain tilesets with white houses (it says 2, but I only see 1 namely BLOXXMAS) and new units: grenade thrower with icon, stealth raider icon. Most work might be indexing all the frames manually again as the numbers change.

pchote:
We're going to need this if we want to support asset installation from digital sources (like the d2k gruntmods edition).  The best approach will be to split our installation source into "original disk files" and "1.06 patch files", and offer our own download for the patch files for people who want to install from an original disk.

pchote:
WIP at https://github.com/pchote/OpenRA/tree/d2k-106.

PR #11522, opened 2016-06-25
Support content installation from "C&C The Ultimate Collection" (Origin versions).

pchote:
This adds support for querying the Windows registry to find an install location that is used as an asset source, and then tweaks the content and installation UI to distinguish between external and internal media.

TD, RA, and TS have been updated to support installation fro TUC.  This required a few minor tweaks to the file layout, which we will be able to clean up as part of the future file-migrations PR.

I have [another branch](https://github.com/pchote/OpenRA/tree/gruntmods) that adds support for installing D2K assets from the gruntmods edition, but that's blocked on #2344.

### DIFF
--- a/mods/ra/mod.yaml
+++ b/mods/ra/mod.yaml
@@ -229,24 +229,24 @@ Migrations:
 ModContent:
 	InstallPromptMessage: Red Alert requires artwork and audio from the original game.\n\nQuick Install will automatically download this content (without music\nor videos) from a mirror of the 2008 Red Alert freeware release.\n\nAdvanced Install includes options for downloading the music and for\ncopying the videos and other content from an original game disc.
 	QuickDownload: basefiles
-	HeaderMessage: The original game content may be copied from an original game disc,\nor downloaded from an online mirror of the 2008 freeware release.
+	HeaderMessage: Game content may be extracted from the original game discs or an\nexisting digital install. OpenRA can also download the base game\nfiles from an online mirror of the 2008 freeware release of RA.
 	Packages:
 		base: Base Game Files
 			TestFiles: ^Content/ra/allies.mix, ^Content/ra/conquer.mix, ^Content/ra/interior.mix, ^Content/ra/redalert.mix, ^Content/ra/russian.mix, ^Content/ra/snow.mix, ^Content/ra/sounds.mix, ^Content/ra/temperat.mix
-			Sources: allied, allied-linux, soviet, soviet-linux
+			Sources: allied, allied-linux, soviet, soviet-linux, origin
 			Required: true
 			Download: basefiles
 		music: Base Game Music
 			TestFiles: ^Content/ra/scores.mix
-			Sources: allied, allied-linux, soviet, soviet-linux
+			Sources: allied, allied-linux, soviet, soviet-linux, origin
 			Download: music
 		movies-allied: Allied Campaign Briefings
 			TestFiles: ^Content/ra/movies1.mix
-			Sources: allied, allied-linux
+			Sources: allied, allied-linux, origin
 		movies-soviet: Soviet Campaign Briefings
 			TestFiles: ^Content/ra/movies2.mix
-			Sources: soviet, soviet-linux
-		music-covertops: Counterstrike Music
+			Sources: soviet, soviet-linux, origin
+		music-counterstrike: Counterstrike Music
 			TestFiles: ^Content/ra/scores-counterstrike.mix
 			Sources: counterstrike, counterstrike-linux
 	Downloads:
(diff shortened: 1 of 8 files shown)


pchote:
Updated.  This now also fixes #11524.

obrakmann:
Code looks good. I couldn't actually test this due to lack of Windows and TUC, though.

Nelax:
Any chance of adding RA2 to this as well?

pchote:
Not in this pr, no (ra2 is a different repo).

Mailaender:
Aftermath and Counterstrike music installs under Linux. :white_check_mark:

mods/ra/mod.yaml
```diff
@@ -466,6 +466,72 @@ ModContent:
 					^Content/ra/music/vr2.aud:
 						Offset: 229462230
 						Length: 2920396
+		aftermath: Aftermath Expansion Disc (English)
+			IDFiles:
+				README.TXT: 9902fb74c019df1b76ff5634e68f0371d790b5e0
+				SETUP/INSTALL/PATCH.RTP: 5bce93f834f9322ddaa7233242e5b6c7fea0bf17
+			Install:
+				extract-raw: MAIN.MIX
+					^Content/ra/music/await.aud:
```
reaperrr: I could swear that `await.aud` is already in vanilla RA's scores.mix.

mods/ra/mod.yaml
```diff
@@ -403,24 +406,66 @@ ModContent:
...
+						Length: 2991979
+					^Content/ra/music/twinmix1.aud:
+						Offset: 224112470
+						Length: 2536972
+					^Content/ra/music/under3.aud:
+						Offset: 226649442
+						Length: 2812788
+					^Content/ra/music/vr2.aud:
+						Offset: 229462230
+						Length: 2920396
```
reaperrr: The original version of "Dense" from vanilla RA is cut off at 3:23, while CS has a fixed, full-length version. I'd recommend adding an entry for CS' `dense_r.aud` while you're at it.

mods/ra/mod.yaml
pchote: IIRC that is `await_r.aud`, which is set up as a separate definition in our music yaml.

mods/ra/mod.yaml
pchote: I guess we could extract this as `dense.aud` to override the broken original.

mods/ts/mod.yaml
```diff
@@ -396,3 +396,56 @@ ModContent:
 					^Content/ts/temperat.mix:
 						Offset: 73056940
 						Length: 2037606
+		origin: C&C The Ultimate Collection (Origin version)
```
Mailaender: You hard-coded this to the English version which you might want to label as you did above.

mods/ts/mod.yaml
```diff
@@ -396,3 +396,56 @@ ModContent:
...
+			Type: Install
+			RegistryKey: HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\EA Games\Command and Conquer Tiberian Sun
+			RegistryValue: Install Dir
+			IDFiles:
+				TSLauncher.exe: 0b3e8dfd7cb701868eccef3ba3717a1ced802707
+				GDFBinary_en_US.dll: 4bb56a449bd0003e7ae67625d90a11ae169319d6
```
Mailaender: If this is the only difference and the archives are otherwise the same, you can remove it and get multi-language support for free.

mods/ts/mod.yaml
pchote: Do non-english versions exist?

mods/ra/mod.yaml
Mailaender: They only have the same label in music.yaml and exact same song length which is a bit irritating.

mods/ts/mod.yaml
pchote: If there are multiple languages then by definition the archives won't otherwise be the same.

mods/ra/mod.yaml
```diff
@@ -26,7 +27,6 @@ Packages:
 	~snow.mix
 	~interior.mix
 	~scores.mix
-	~scores-counterstrike.mix
```
Mailaender: This probably requires a migration rule as well. https://github.com/OpenRA/OpenRA/pull/11528

pchote: We haven't shipped a release with this, so there is no need.

mods/ts/mod.yaml
Mailaender: Origin seems to automatically pick the localized version at least during the on the house promotion and I am not sure it allows you to download original locales like Steam does:
image

mods/ts/mod.yaml
```diff
@@ -396,3 +396,56 @@ ModContent:
...
+			Install:
+				copy: .
+					^Content/ts/scores.mix: SCORES.MIX
+				extract-raw: TIBSUN.MIX
```
Mailaender: Why are you extracting TIBSUN.MIX instead of copying it?

pchote: Because we don't want tibsun.mix.  We only want the contents below.

Mailaender: You can make this work with the localized versions as well if you don't rely on the file sizes of the .mix files so it might be smart to just copy the TIBSUN.MIX here.

pchote: We have already dropped support for loading TS with content inside tibsun.mix, and allowing this to copy non-english versions would go against one of the core principles of this rework (having a definitive set of metadata describing exactly what is being installed).

Non-english content can be described with their own set of disk definitions in the future after the low level translation support is completed.

Mailaender:
I can confirm that this works with Red Alert 2 :white_check_mark:  https://github.com/OpenRA/ra2/pull/246

mods/ra/mod.yaml
pchote: In any case this is an issue with our music setup rather than our installer, so if something needs to be changed it can be done in another PR.

pchote:
Updated.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/001d139675d80cec78ce88136caefcdb2776bc3b...c11e7afa2db09681b4106a670fe5ee0a0fc72ac9)

PR #11375, opened 2016-05-29
New mod content installer

pchote:
The original content installer has been hacked together over the years, and had a whole bunch of limitations.  This PR completely rewrites it, bringing the following benefits:
- Now understands the difference between different content sources, making it easy to add new ones in the future (Origin / cncnet / GOG etc, as well as translated versions).
- Can now install disk-specific assets (FMVs, expansion music)
- Improved discovery for music installation
- Includes mod-specific description text, making it clear that our asset downloads come from the freeware TD/RA/TS releases
- Removes the modchooser dependency on game-specific package parsers (Installshield, Mix, etc)

There are currently two minor regressions which I would like to leave to followup PRs (the PR is big enough as it is, and these are separate to the main changes):
- We lose TFD support until somebody who has access to the disk can set up the new metadata for it
- The music installation prompt is not shown for players who have the hacked scores.mix from the original assets download.  A future PR will add metadata that tells the game to delete that file as a migration step

Closes #7412, Closes #11344, Closes #10556
Related to #9479, #5970, #8993, #4361, #9082

screen shot 2016-05-29 at 19 55 37
screen shot 2016-05-29 at 19 56 03
screen shot 2016-05-29 at 19 56 07
screen shot 2016-05-29 at 19 56 32

PR #11528, opened 2016-06-25
Delete legacy downloaded scores.mix on first run

### LINKED ISSUE #11525: Add a "Migrations" metadata block for deleting old files when loading a mod.
This should run before the content installation check.

pchote:
The asset download ships a mangled scores.mix that only has Hell March.  This causes problems for the music player (#11375 ripped out the special place hacks it used to have) and content manager.

This implements a new "Migrations" block that, for now, lets us match files based on name and length and delete them.  I have implemented this in a general enough way that we can expand it later as needs arise for moving content, migrating maps to newer releases, cleaning up older content (replays, maps – with a confirmation prompt), etc.

Closes #11525.

### DIFF
--- a/OpenRA.Game/Game.cs
+++ b/OpenRA.Game/Game.cs
@@ -344,6 +344,13 @@ namespace OpenRA
 			using (new PerfTimer("LoadMaps"))
 				ModData.MapCache.LoadMaps();
 
+			if (ModData.Manifest.Contains<Migrations>())
+			{
+				var reload = ModData.Manifest.Get<Migrations>().Run();
+				if (reload)
+					InitializeMod(mod, args);
+			}
+
 			var content = ModData.Manifest.Get<ModContent>();
 			var isModContentInstalled = content.Packages
 				.Where(p => p.Value.Required)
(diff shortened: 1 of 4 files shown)


obrakmann:
Code looks good and works well enough for this use case, :+1: 

If this does get expanded with more use cases in the future, I feel that there should probably be some kind of state-tracking. Running more elaborate rules on every mod load would feel a bit wrong. Also, if we  eventually do add the cleanup option, and the user declines, he'd be forced into the dialog every time the mod loads.

I don't really have a good idea where to keep that state, though. Maybe just storing a cookie with the mod's version in settings.yaml might be enough. Then skip migrations if that cookie matches the current version, otherwise run them. Would only work for upgrades, though, not downgrades.

I could see migrations ending up needing identifiers and a flag indicating whether they are mandatory as well.

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/638042778eff25fc7689f58fe2479e347ab4bf35...aea552a4f7f20aa00a1cfc9264b8d4a6038f6406)

Issue #11279, opened 2016-05-14
Wall rendering broken when Explored Map is enabled


obrakmann:
![openra-2016-05-14t144749z](https://cloud.githubusercontent.com/assets/4331210/15269176/bc0e9024-19f7-11e6-9a86-af5a8ae8e227.png)

abcdefg30:
This is turning out to be very weird:
- We could only reproduce this in multiplayer
- But not every player and even spectator does see the bug

(For example: I had two spectating clients, one saw it, the other not).

pchote:
I've been able to show that this happens because/when Game calls RenderTick before the regular Tick... which shouldn't be possible, but is happening anyway.  I don't have a reliable way to reproduce the glitch, so I haven't yet been able to debug this any further.

pchote:
Got it.  #12148.

Issue #5908, opened 2014-07-09
Minimap shows harvested resources, even if not discovered


abcdefg30:
On playtest-20140709:
![minimapbuga](https://cloud.githubusercontent.com/assets/7704140/3529761/435efb50-079d-11e4-94a8-16b8f78c0ccb.png)
![minimapbugb](https://cloud.githubusercontent.com/assets/7704140/3529762/476687b8-079d-11e4-9428-e0585255291c.png)
![minimapbugc](https://cloud.githubusercontent.com/assets/7704140/3529862/6d760a72-079e-11e4-8a0e-1633df048ff9.png)
Appears when the fog of war is disabled.
(In the pics I used the spysatellite.)
##

pchote:
The "All players" view is a union of all the player's sight.  What you're suggesting is an intersection, which would be equivalent (when applied to shroud) to having the whole map hidden except for the parts that *everyone* has explored.  That doesn't make a lot of sense.

pi99y:
Please also take look at this related issue #9155 which does almost the opposite - doesn't show the discovered buildings under the fog of war on the mini-map. Thanks.

Mailaender:
I fiddled around with it and this is harder to fix than I thought. Not even sure what is the best approach here. Create a new interface/celllayer for the radar to query or keep the special logic: "resources hidden under fog, but not during first init as we don't if the map started with fog explored" inside RadarWidget.

pchote:
That data already exists, and is what we render in the world.  The radar just needs to use it, instead of the true resource state.

Mailaender:
Not sure why I overcomplicated it so much. Indeed the `FogObscures` check is already done `RenderContent` for easy re-use: https://github.com/Mailaender/OpenRA/blob/hide-minimap-resources/OpenRA.Mods.Common/Traits/World/ResourceLayer.cs#L168

SoScared:
@Mailaender I'm not sure how much people have nagged you guys about this issue in the past but it's a problem being mentioned very frequently among players the past few years. A fix would definitely be an improvement competitive wise.

Mailaender:
I tried to fix it, but essentially failed in the end. https://github.com/OpenRA/OpenRA/pull/11672

SoScared:
Ah ok. Looks like the path continued with #12093 . Here's to hoping^^

pchote:
[...]

PR #11531, opened 2016-06-26
Set explored map as the default lobby option for all mods

ghost:
This seems to be how the majority of players in TD and recently [RA](http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19611) wish to play on-line. Aside from the points raised about helping new players, I have found this actually removes _a lot_ of frustration out of playing early game in general, so I just set it for all mods.
- Eliminates units refusing to move if you click on impassable terrain under the shroud. Also eliminates having to shift click spam a black screen to get around this. 
- If playing random spawns, you no longer have to pan the screen in four different directions to figure out where the map border is (and hence your spawn position)
- No more memorizing a map for REF placement, tech buildings, debris / ore in the spawn blocking building placement etc. No more shift click spamming scouts just to uncover the map - now you can focus on scouting the enemy and countering their scouts.
- No longer at a disadvantage on new maps so you can focus on just playing. Much less frustrating to try different tactics like changing your initial MCV deployment and rushes.

### DIFF
--- a/OpenRA.Game/Traits/World/Shroud.cs
+++ b/OpenRA.Game/Traits/World/Shroud.cs
@@ -26,7 +26,7 @@ namespace OpenRA.Traits
 		public bool FogLocked = false;
 
 		[Desc("Default value of the explore map checkbox in the lobby.")]
-		public bool ExploredMapEnabled = false;
+		public bool ExploredMapEnabled = true;
 
 		[Desc("Prevent the explore map enabled state from being changed in the lobby.")]
 		public bool ExploredMapLocked = false;


pchote:
:+1: to the idea, but we should make fixing #5908 a priority (I'm tempted to say a prerequisite) if we take this.

HenrytheSlav:
^Basically what Paul says, the minimap really bothers me. Recently I was able to expect ambush from behind only because I saw ore being eaten elsewhere than expected. Otherwise probably good idea, I am jsut worried that over time people will forget about using shroud altogether, and that would be sad imho. It still is fun thing to have, combined with GPS loss etc.

Micr0Bit:
:-1: 

oh and btw. entering a RadarDome with a Spy will have no effect ... theres also the "exploration crates" settiing back shroud , or revealing it , that might need a different effect instead ...so you guys might consider fixing that too ... since we are talking about changing something to default those things should still have effect even when shroud is off

i have nothing more to say here , other then 

@pchote : the only applause you will earn here , comes from a temporary community and 2014-16 rookies ... and changing rules by having a crowd of newbies (in terms of invested openra-time) that add momentum to your ideas , although you personally are hardly effected by them ... won't make you happy in the future

pchote:
@Micr0Bit the community vote at http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19611 is overwhelmingly in favour of enabling it by default.  If the "hardcore" player base wants their voice heard on this matter then they need to actually participate in the discussion.

Micr0Bit:
@pchote : most of the "overwhelmingly" votes for "yes please" could be also just multi-accounts ... 

and beside that , the poll had really bad voting-options at the beginning (before you changed it) ... and earned a good amount of "yes-please" voters that way ...

Micr0Bit:
you should consider fixing your forum before thinking about influencing open-source decisions with votes from it ...

Mailaender:
Also :+1: to this default setting. However indeed we should fix the known problems first to avoid a bad first impression. Guess we need to add #11279 to that list.

pchote:
If we are going to impose all of
- [ ] #3654
- [x] #5908
- [x] #11279

(which seems a bit overkill, IMO) then we probably want to close this for the near (and probably medium-to-long term) future until someone wants to tackle those.

SoScared:
:-1: I believe the majority vote was somewhat skewed on Sleipnir. Regardless I think this change would lessen the experience of the mods. The role of the shroud is discovery and casual players are coming in for the fun. I don't think there's an experience of frustration among the vast majority of players when it comes to the shroud. It's something that is there that rewards you for exploring. It's satisfying to wipe away the dark with units. Defaulting explored map would take this away from a lot of players that don't consider tactics or strategy.

pchote:
The obvious next step is to poll the wider community instead of relying on the relatively closed group of active forum and IRC members.  We can have a news post asking for player feedback and syndicate that to social media and the rest of our usual community places.

pchote:
It's clear that this is going nowhere fast, so I'll close this for now to clear up the PR queue.

SoScared:
Just wanna nudge in a change of heart from my earlier statement on this. Despite the good intentions there's no denying that new players in the MP lobby quickly starts favoring explored map once they get used to the UI. Seems then counter-productive to impose a 'better experience' for newcomers if they don't care for it.

Would be nice to have https://github.com/OpenRA/OpenRA/pull/11531#issuecomment-231556362 fixed as a part of such a change tho.

Micr0Bit:
>Just wanna nudge in a change of heart from my earlier statement on this. I'm seeing new players in the MP lobby quickly starts favoring explored map once they get used to the UI. Seems then counter-productive to impose a 'better experience' for newcomers if they don't care for it.

after being forced many times to play public team-games where the shroud gets turned off "manually" ... 80-90% of the games im playing,  without a sign of this trend finding an end ... i have to second this

@pchote : im sry again for the troubles i have caused on sleipnir about this issue ... though i would assume , that the constant stream of changes on the mods towards gameplay and/or balance (not the engine itself) can get depressing for some of the deep-rooted regular players

PR #11532, opened 2016-06-26
[RA] Rebalance russia faction / tesla weaponry

ghost:
It seems there is some [agreement](http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19579&postdays=0&postorder=asc&start=0) that shockies are still a little OP for their price. There have been a number of varying suggestions from the community (accreditations might be off or a little inaccurate, I apologize) on how to fix them so I tried a bunch of them together (tweaked) and the results seem to gel pretty good.
- Damage vs infantry reduced (Suggested by AoAGeneral)
- Damage vs everything else reduced (Suggested by klaas, SoScared) by ~20%
- HP reduced (Suggested by Murto, SoScared) by 20%
- Slight fire delay

These stats sound extreme but in reality it means
- Medics, tanyas and mechs are no longer one shotted
- Shockies now pause before firing for about the same time as a flametrooper
- They are a little more fragile and a little less melty

To make up for that, I buffed the TT a little
- Now has a 2 shot burst (suggested by FiveAces)
- Slight fire delay

Which means
- Now takes slightly longer than a shock trooper to fire
- Can now finish off two very low health targets per burst or two full health infantry

Finally all tesla weaponry damage vs walls has been reduced to 50% (suggested by SoScared)

### DIFF
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -479,7 +479,7 @@ SHOK:
 		Name: Shock Trooper
 		Description: Elite infantry with portable tesla coils.\n  Strong vs Infantry, Vehicles\n  Weak vs Aircraft
 	Health:
-		HP: 60
+		HP: 50
 	Mobile:
 		Voice: Move
 	RevealsShroud:
@@ -487,9 +487,11 @@ SHOK:
 	Armament@PRIMARY:
 		Weapon: PortaTesla
 		LocalOffset: 427,0,341
+		FireDelay: 12
 	Armament@GARRISONED:
 		Name: garrisoned
 		Weapon: PortaTesla
+		FireDelay: 12
 	AttackFrontal:
 		Voice: Attack
 	AttackMove:
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -573,8 +573,10 @@ TTNK:
 	Armament:
 		Weapon: TTankZap
 		LocalOffset: 0,0,213
-	AttackTurreted:
-	Turreted:
+	AttackCharge:
+		InitialChargeDelay: 16
+		MaxCharges: 2
+		ReloadDelay: 120
 	WithIdleOverlay@SPINNER:
 		Sequence: spinner
 	SelectionDecorations:
--- a/mods/ra/weapons/other.yaml
+++ b/mods/ra/weapons/other.yaml
@@ -152,8 +152,9 @@ TeslaZap:
 		Spread: 42
 		Damage: 100
 		Versus:
-			None: 1000
+			None: 200
 			Wood: 60
+			Concrete: 50
 		DamageTypes: Prone50Percent, TriggerProne, ElectricityDeath
 
 PortaTesla:
@@ -163,21 +164,23 @@ PortaTesla:
 	Projectile: TeslaZap
 	Warhead@1Dam: SpreadDamage
 		Spread: 42
-		Damage: 45
+		Damage: 35
 		Versus:
-			None: 1000
+			None: 200
+			Concrete: 50
 		DamageTypes: Prone50Percent, TriggerProne, ElectricityDeath
 
 TTankZap:
-	ReloadDelay: 120
+	ReloadDelay: 2
 	Range: 7c0
 	Report: tesla1.aud
 	Projectile: TeslaZap
 	Warhead@1Dam: SpreadDamage
 		Spread: 42
-		Damage: 100
+		Damage: 50
 		Versus:
-			None: 1000
+			None: 200
+			Concrete: 50
 		DamageTypes: Prone50Percent, TriggerProne, ElectricityDeath
 
 DogJaw:


pchote:
Doesn't this ruin/remove the distinction between flame troopers and shockies?

jaZzKCS:
I am all for what AoA suggested. Adjusting the damage versus infantry to a value much more close to original RA96 (two-to-three hits per death for inf, instead of insta-kill) would make a huge difference in their over all performance, which seems to be a tad too high on all fronts atm.

HenrytheSlav:
pchote: I was always taking flamers as 1) antibuilding assault units and 2) anti mass-infantry. In lower numbers/spread infantry they do low damage for their cost. On the other hand, shockies do damage always same, and are good against vehicles too.

pchote:
That's what I mean – shockies were explicitly balanced as a sniper-like unit to (a) replace the sniper and (b) keep them distinct from flamers.  These changes seem to be undoing all of those distinctions in one extreme swoop, and compensates by buffing *a different* unit.

HenrytheSlav:
I see your pount, and I didn't say I agree to the changes. Most sensible for me would be to keep the damage (it is very high tech unit) and lower health by 25% (making it high risk high reward) and possibly uping a price by 100, but that might be too much. Nerf damage vs. everything is always bullshit, this is faction specific and very late game tech. Don't want to be steamrolled, snipe the techcentre.

r34ch:
@pchote @jaZzKCS  shockies still snipe **all** infantry in the game except the 3 I mention, in which case two shockies are needed to snipe a full health medic, mech or tanya. In reality it means a player has to take out a tanya by other means first before moving shockies in, which I think is fair considering she is a single limit hero unit.

HenrytheSlav:
@r34ch But if they are sniper-tier unit, then that is intented imho. I don't really feel like damage nerf is okay here. Just make them more vulnerable since players can't afford to lose many. Make V2/Arty/Yak/Hind the hard counters.

r34ch:
@HenrytheSlav Must a sniper unit one shot _every_ infantry in the game? I think one shotting all infantry in the game except for three of them is still fair? One of those units has a build limit of 1 and the other 2 are completely defenseless.

The damage nerf is just to slow them down a little once they get going. No other units in the game are quite so 'melty'. I mean 20% literally means you have to build one more shock trooper for every 4 you currently build.

kyrylo:
I don't like anything presented in the PR. I like that shock troopers are better than all other infantry in the game. It's an expensive unit and it must be good. Their main problem, in my opinion is that they destroy **all** vehicles in seconds (and maybe even outrange tanks). My main complaint about them is the range. My humble suggestion is not to make any drastic changes anymore, but make their range/vision equal to the rifler's vision. Another suggestion: `range = riflerVision + 1cell`, so they can be used as a support unit.

What makes them OP is that it's just too easy to a-move them, without microing. With the range change you would need to care about their positioning. I think it's a good balance change to try, without breaking the current balance (which I believe this PR will do). I also don't see how the proposed change is related to tesla tanks at all. It just doesn't make any sense, these units are not logically dependant on each other in any way.

Summarising, I think reducing their range is the way to go (but leaving everything else the same). This leaves a chance for dogs to eat them, tanks to reach them, defence structures to live a little bit longer. In 1v1 versus any infantry, a shock trooper will still win, which is how it should be. As of now, because of the range they feel like mini-inf-arties.

HenrytheSlav:
I see @kyrylo argument, how about range and health nerf? I stand behind health since it makes people more protective of them - no more just a move - while weaker against hind/artry/whatever. Their price/tech is very matching their damage to everything.

r34ch:
Ok @kyrylo, @HenrytheSlav I think you've both made valid points about the ease of a moving and given the confusion about the tesla tank I think I will create new separate PRs with the ideas and to reduce confusion.

r34ch:
Closing to try and get a consensus rather than trying to please everyone. [Please discuss on the forums](http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&p=296702#296702)

MurtoTheRay:
Hello everyone, i thought i would pitch in here. I have played hundreds of games using Russia in the past year on a variety of maps which has given me a lot of experience using shock troopers. In my (unpopular) opinion shock troopers shouldn't be nerfed but i am willing to push this belief to one side and accept the health nerf due to the amount of people who are unhappy with shock troopers. I will try to be non-biased and since i will be playing Ukraine mainly from now on i don't really have a reason to protect Shock Troopers. 

Shock Troopers are actually a really hard unit to get. They cant be rushed against someone who scouting properly and are expensive to produce. Also, the current meta is large armies and open maps which means that rushing shock troopers results in you being trampled by the opponents army as you have sacrificed production for tech which wont pay off until you have built shock troopers.

But on some maps (such as Behind the Veil or Winter Storm) it is much more possible to rush them as the map has more available chokes and longer rush distances. But even then it is still possible  to counter shock troopers. Shock Troopers require a lot of income to build constantly and  so mass expanding an opponent who is tech rushing is also a possible counter, although less intuitive and is the least utilized counter measure i have seen vs shockies. So there is still learning to do for quite a  few people as most are too focused on winning by attacking instead of winning through attrition.

Back to the meta. As i have already mentioned, the meta is focused around large armies: infantry and tank armies to be exact. Opening with a tank/infantry composition is so common that you will see it in pretty much every game with moderately skilled players. This is such easy prey for shock troopers if the opponent keeps this up into the late game. Artillery is also common due to the infantry masses that are created but the flak/shocky combo is a hard counter to this. Harassment based tactics such as chrono shifting enemy harvesters and utilizing longbows(this will be much better if they get buffed any time soon) are far better vs shockies than the norm atm.

Well that's why they do so well in the meta. Now onto the proposed changes:

Range - Back when shock troopers were truly OP their range was they only thing changed and that's all it took to make an OP unit. RA is very dependent on range as it takes so long for units to get into range if they are being out-ranged. Changing it down, even slightly, could break the unit and make it useless again. Test it though.

Damage - Well you really need to have a lot of  shockies to do a better job than the cheaper infantry. Rocket troopers kill vehicles faster in small numbers, same with flamers and buildings, grens and buildings, rockets and defences although am unsure about rifles and infantry. Though rifles change their role to meat shields to break up shock troopers and increase their survivability.

Slight fire delay - maybe but i don't think both this and the health nerf should be applied. I could  see this working out well on it's own though and it would make the range not as overpowering. It would also make shock troopers an aggressive unit as a player with the intuitive vs defensive shock troopers could exploit the charge time to do real damage.

PR #11534, opened 2016-06-26
Fixed a minor mistake in Red Alert's ModContent 

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/11375.


pchote:
This is already fixed in #11522.

Issue #10903, opened 2016-03-10
UPnP has bad discoverability


Alex34567889:
Hello.Sorry a silly question. Why can not check the box the Automatic port forwarding in the Create Server dialog. Prompt me please.
![2016-03-10_105058](https://cloud.githubusercontent.com/assets/17353812/13662956/4ba45ca4-e6af-11e5-8bcb-27324ec7ef91.jpg)

jaZzKCS:
It means that UPnP port forwarding is either disabled or unavailable on your system.
In this case - if you are behind a router - you would have to log-on into your router and forward the neccessray ports manually to the internal LAN address of the computer you want the server to run on / want to host the game.

ScottNZ:
I think it would be good to have a tooltip explaining this when you hover over the checkbox

Alex34567889:
UPnP enabled. It remains a problem in other.
![2016-03-10_162422](https://cloud.githubusercontent.com/assets/17353812/13670778/e3c7578e-e6dc-11e5-8719-df41fcb68758.jpg)
That is without this command, the server does not create?

pchote:
The current UPnP library is quite broken and relatively unmaintained, and our preferred alternative requires us to upgrade our minimum depedencies to .NET 4.5.2 (which means dropping support for Windows XP and certain older but still supported Linux distributions).

We're making steps to hopefully make this switch in the second release after 20151224.

KevinYeti:
@pchote hi, I'm new here and I just want to know the UPnp library has been upgraded for now ? because the Upnp function seems not work fine for me.

pchote:
Not yet, but soon (for the next release). #11284 finally removed support for windows xp and the older versions of mono that were holding us back, and #11286 is currently open to switch to the new UPnP library and enable it by default.

fadyanwar:
Is UPnP still and issue? I'm unable to host a server on my machine. Release 20190314
  Operating System: Ubuntu 19.04
  Kernel: Linux 5.0.0-16-generic
  Architecture: x86-64

PR #11535, opened 2016-06-26
Improved the UPnP settings discoverability

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/10903.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs
@@ -94,6 +94,10 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			if (labelUPnP != null)
 				labelUPnP.IsVisible = () => !Game.Settings.Server.DiscoverNatDevices;
 
+			var labelUPnPUnsupported = panel.GetOrNull<LabelWidget>("UPNP_UNSUPPORTED_NOTICE");
+			if (labelUPnPUnsupported != null)
+				labelUPnPUnsupported.IsVisible = () => Game.Settings.Server.DiscoverNatDevices && !Game.Settings.Server.AllowPortForward;
+
 			var passwordField = panel.GetOrNull<PasswordFieldWidget>("PASSWORD");
 			if (passwordField != null)
 				passwordField.Text = Game.Settings.Server.Password;
--- a/OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs
@@ -90,6 +90,10 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			checkboxUPnP.OnClick = () => allowPortForward ^= true;
 			checkboxUPnP.IsDisabled = () => !Game.Settings.Server.AllowPortForward;
 
+			var labelUPnP = panel.GetOrNull<LabelWidget>("UPNP_NOTICE");
+			if (labelUPnP != null)
+				labelUPnP.IsVisible = () => !Game.Settings.Server.DiscoverNatDevices;
+
 			var passwordField = panel.GetOrNull<PasswordFieldWidget>("PASSWORD");
 			if (passwordField != null)
 				passwordField.Text = Game.Settings.Server.Password;
--- a/mods/cnc/chrome/multiplayer-createserver.yaml
+++ b/mods/cnc/chrome/multiplayer-createserver.yaml
@@ -101,6 +101,15 @@ Container@MULTIPLAYER_CREATESERVER_PANEL:
 							Align: Left
 							WordWrap: True
 							Text: Requires Network Discovery (UPnP). Enable in Settings - Advanced.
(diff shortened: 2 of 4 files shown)


mods/cnc/chrome/multiplayer-createserver.yaml
```diff
@@ -92,6 +92,15 @@ Container@MULTIPLAYER_CREATESERVER_PANEL:
...
 							Text: Automatic port forwarding
+						Label@UPNP_NOTICE:
+							X: 130
+							Y: 180
+							Width: 200
+							Height: 25
+							Font: Tiny
+							Align: Left
+							WordWrap: True
+							Text: Requires Network Discovery (UPnP). Enable in Settings - Advanced.
```
pchote: We only disable UPnP if there's an error, so i'm not sure we should be casually telling people how to reenable it.  That will cause confusion when the game immediately disables it again.

Mailaender: We disable UPnP (discovery) by default, because @chrisforbes and @obrakmann don't think it is polite to add holes to network security without asking first. You are confusing this with the `AllowPortForward` and `NatDeviceAvailable` settings.

OpenRA.Mods.Common/Widgets/Logic/ServerCreationLogic.cs
```diff
@@ -90,6 +90,10 @@ public ServerCreationLogic(Widget widget, ModData modData, Action onExit, Action
 			checkboxUPnP.OnClick = () => allowPortForward ^= true;
 			checkboxUPnP.IsDisabled = () => !Game.Settings.Server.AllowPortForward;
 
+			var labelUPnP = panel.GetOrNull<LabelWidget>("UPNP_NOTICE");
+			if (labelUPnP != null)
+				labelUPnP.IsVisible = () => !Game.Settings.Server.DiscoverNatDevices;
```
pchote: Elaborating on my last comment after checking our settings code, I don't think this is helpful.  `DiscoverNatDevices` is the manual override that very few people will ever touch.  The confusion from users comes from `AllowPortForward` being automatically disabled by the game.

pchote: On that note, `Game.Settings.Server.AllowPortForward` really needs to be moved from Settings to UPNP, because it has nothing to do with settings.

Mailaender: `Game.Settings.Server.AllowPortForward` is a setting. It decides whether to do automatic port forwarding if a NAT device is available and let's the user decide whether he wants to or not. You may be confusing this with `NatDeviceAvailable` which indeed could be moved to UPnP or replaced by a `NatDevice == null` check.

pchote: Well this just reinforces my point about this: we should not need three distinct settings to decide whether or not to enable port forwarding.

Mailaender: I already removed it in https://github.com/OpenRA/OpenRA/pull/11286/files#diff-0871ba7519d8b084807b23cec86099d8 Forgot about that.

obrakmann:
How about we hide the checkbox instead when UPNP isn't available? Then players wouldn't be confused why the box is greyed out.

abcdefg30:
On the other hand, they might wonder where the UPNP feature went at all (and open bug reports or search the whole settings). Just a random thought.

abcdefg30:

GraionDilach:
This looks fine by me as well, but I'm not certain if https://github.com/OpenRA/OpenRA/pull/11535#discussion_r68504319 has reached a consensus. :+0.9:?

obrakmann:
> On the other hand, they might wonder where the UPNP feature went at all (and open bug reports or search the whole settings). Just a random thought.

And if we add this text as is, we'll get bug reports from users whose router doesn't support UPNP because then the box still stays grey even though the setting is checked and seemingly doesn't work.

How about this solution then:

1. UPNP is disabled in the settings:
 Show the disabled checkbox and the label from this PR.

2. UPNP is enabled in the settings, but disabled because the router doesn't support it:
 Hide both the checkbox and the label, or
 Show the disabled checkbox and another label that explains the missing UPNP support in the router.

3. UPNP is enabled in the settings and the router supports it:
 Only show the enabled checkbox, hide the label.

pchote:
That solution sounds good to me.

obrakmann:
https://github.com/Mailaender/OpenRA/pull/57

mods/cnc/chrome/multiplayer-createserver.yaml
pchote: I still really don't like this "Enable in Settings - Advanced." label.  It is bad for polish.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5ae91346de22745455e2c98df90ad2a68d403a31)

abioneperhobby:
upnp not working even the router has upnp on and upnp is working on every other application (emule, torrent, etc).

PR #11543, opened 2016-06-29
Allow overriding previously registered chatcommands

phrohdoh:
This bit me a few times when writing #11544.

### DIFF
--- a/OpenRA.Mods.Common/Commands/ChatCommands.cs
+++ b/OpenRA.Mods.Common/Commands/ChatCommands.cs
@@ -48,7 +48,8 @@ namespace OpenRA.Mods.Common.Commands
 
 		public void RegisterCommand(string name, IChatCommand command)
 		{
-			Commands.Add(name.ToLowerInvariant(), command);
+			// Override possible duplicates instead of crashing.
+			Commands[name.ToLowerInvariant()] = command;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


abcdefg30:
Looks simple enough. 👍

OpenRA.Mods.Common/Commands/ChatCommands.cs
```diff
@@ -48,7 +48,10 @@ public bool OnChat(string playername, string message)
 
 		public void RegisterCommand(string name, IChatCommand command)
 		{
-			Commands.Add(name.ToLowerInvariant(), command);
+			name = name.ToLowerInvariant();
```
pchote: Wouldnt it make more sense for the later command to take priority?  You could then simplify it down to one line Commands[name] = ...

Phrohdoh:
Updated.

OpenRA.Mods.Common/Commands/ChatCommands.cs
```diff
@@ -48,7 +48,7 @@ public bool OnChat(string playername, string message)
 
 		public void RegisterCommand(string name, IChatCommand command)
 		{
-			Commands.Add(name.ToLowerInvariant(), command);
+			Commands[name.ToLowerInvariant()] = command;
```
pchote: Maybe worth a comment explicitly stating that we choose to ignore earlier duplicates instead of crashing?

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/4445ba5352a97836eee06aad2d6355036914bf74...579de74a8ace61b9dbe2fcabcfadc62349f5df7f)

PR #11544, opened 2016-06-29
Add ExitsDebugOverlay

phrohdoh:
screen shot 2016-07-02 at 5 45 05 pm
screen shot 2016-07-02 at 5 45 11 pm
screen shot 2016-07-02 at 5 45 21 pm
screen shot 2016-07-02 at 6 01 31 pm

### DIFF
--- a/OpenRA.Mods.Common/Traits/ExitsDebugOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/ExitsDebugOverlay.cs
@@ -1,0 +1,105 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Drawing;
+using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Displays `Exit` data for factories.")]
+	public class ExitsDebugOverlayInfo : ITraitInfo, Requires<ExitInfo>
+	{
+		[Desc("Should cell vectors be drawn for each perimeter cell?")]
+		public readonly bool DrawPerimiterCellVectors = true;
+
+		[Desc("Should cell vectors be drawn for each exit cell?")]
+		public readonly bool DrawExitCellVectors = true;
+
+		[Desc("Should lines be drawn for each exit (from spawn offset to the center of the exit cell)?")]
+		public readonly bool DrawSpawnOffsetLines = true;
+
+		object ITraitInfo.Create(ActorInitializer init) { return new ExitsDebugOverlay(init.Self, this); }
+	}
+
+	public class ExitsDebugOverlay : IPostRender
+	{
+		readonly ExitsDebugOverlayManager manager;
+		readonly ExitsDebugOverlayInfo info;
+		readonly RgbaColorRenderer rgbaRenderer;
+		readonly ExitInfo[] exits;
+
+		CPos[] exitCells;
+		WPos[] spawnPositions;
+		CPos[] perimeterCells;
+
+		public ExitsDebugOverlay(Actor self, ExitsDebugOverlayInfo info)
+		{
+			this.info = info;
+			manager = self.World.WorldActor.TraitOrDefault<ExitsDebugOverlayManager>();
+			rgbaRenderer = Game.Renderer.WorldRgbaColorRenderer;
+			exits = self.Info.TraitInfos<ExitInfo>().ToArray();
+		}
+
+		void IPostRender.RenderAfterWorld(WorldRenderer wr, Actor self)
+		{
+			if (manager == null || !manager.Enabled)
+				return;
+
+			exitCells = exits.Select(e => self.Location + e.ExitCell).ToArray();
+
+			if (info.DrawExitCellVectors)
+			{
+				foreach (var exitCell in exitCells)
+				{
+					var color = self.Owner.Color.RGB;
+					var vec = exitCell - self.Location;
+					var center = wr.World.Map.CenterOfCell(exitCell);
+					new TextRenderable(manager.Font, center, 0, color, vec.ToString()).Render(wr);
+				}
+			}
+
+			if (info.DrawPerimiterCellVectors)
+			{
+				var occupiedCells = self.OccupiesSpace.OccupiedCells().Select(p => p.First).ToArray();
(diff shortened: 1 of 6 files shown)


reaperrr:
> OpenRA.Mods.Common/Traits/ExitsDebugOverlay.cs:L53: [SP2000] Invalid spacing at the end of the line.

Phrohdoh:
I rewrote this a bit and added some uses to RA.

OpenRA.Mods.Common/Traits/ExitsDebugOverlay.cs
```diff
@@ -0,0 +1,120 @@
...
+			var exits = self.Info.TraitInfos<ExitInfo>();
+
+			exitCells = exits.Select(e => self.Location + e.ExitCell).ToArray();
+			spawnPositions = exits.Select(e => self.CenterPosition + e.SpawnOffset).ToArray();
+
+			var occupiedCells = self.Trait<Building>().OccupiedCells().Select(p => p.First);
+			perimeterCells = Util.ExpandFootprint(occupiedCells, true).Except(occupiedCells).ToArray();
+
+			var console = self.World.WorldActor.TraitOrDefault<ChatCommands>();
+			var help = self.World.WorldActor.TraitOrDefault<HelpCommand>();
```
Mailaender: This should live in `WorldLoaded`.

OpenRA.Mods.Common/Traits/ExitsDebugOverlay.cs
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		void INotifyCreated.Created(Actor self)
+		{
+			var exits = self.Info.TraitInfos<ExitInfo>();
+
+			exitCells = exits.Select(e => self.Location + e.ExitCell).ToArray();
+			spawnPositions = exits.Select(e => self.CenterPosition + e.SpawnOffset).ToArray();
+
+			var occupiedCells = self.Trait<Building>().OccupiedCells().Select(p => p.First);
```
Mailaender: You can move the `Building` (and `ExitInfo`) lookup into the constructor as the `Requires` will make sure everything is set up in the right order.

OpenRA.Mods.Common/Traits/ExitsDebugOverlay.cs
```diff
@@ -0,0 +1,120 @@
...
+
+			rgbaRenderer = Game.Renderer.WorldRgbaColorRenderer;
+		}
+
+		void INotifyCreated.Created(Actor self)
+		{
+			var exits = self.Info.TraitInfos<ExitInfo>();
+
+			exitCells = exits.Select(e => self.Location + e.ExitCell).ToArray();
+			spawnPositions = exits.Select(e => self.CenterPosition + e.SpawnOffset).ToArray();
```
pchote: Caching these at creation time won't work for actors that can move.

Mailaender: I would like to see these lookups and calculations done when the command is actually invoked if possible. Right now it will be done for every player, for every production building during every regular game.

OpenRA.Mods.Common/Traits/ExitsDebugOverlay.cs
Phrohdoh: It should not.
`IWorldLoaded` is only queried for traits that go on the World actor.

OpenRA.Mods.Common/Traits/ExitsDebugOverlay.cs
Phrohdoh: @pchote Do we have any mobile actors with Building and Exit?

pchote: Why should this be limited to only building actors?

Phrohdoh: I suppose I could calculate perimeter cells based on the existence of Building (or if it doesn't exist I can just expand `self.Location`).

pchote: `IOccupySpace.OccupiedCells()` should work for all actors.

pchote: Or alternatively only display the exits, and not the other perimeter cells.

Phrohdoh: I'd like to keep the perimeter cells so I'll go with what you mentioned above, I did not realize that existed.

Phrohdoh:
I'm going to rewrite this with a manager-like trait that lives on the world actor that individual actor traits will check.

OpenRA.Mods.Common/Traits/World/ExitsDebugOverlayManager.cs
```diff
@@ -0,0 +1,54 @@
+﻿using OpenRA.Graphics;
```
obrakmann: Header is missing

obrakmann:
Looks okay to me. :+1: 

Any reason why you only added it to actors with multiple exits? Not complaining, just curious. We don't really need it in the default mods anyway, so it'll just serve as an example, I guess.

Phrohdoh:
I only added this to factories with multiple exits (except for that TS shot) because that is when I found the overlay most helpful.

Would you all be opposed to not using this in the shipped mods?
As you stated we don't have any reason to use them and they're just more objects sitting around not contributing to gameplay.

Phrohdoh:
I've removed the uses from RA/TD/D2k in favor of putting the traits on all of the producers in TS.

screen shot 2016-07-07 at 9 01 57 am
screen shot 2016-07-07 at 9 05 12 am

GraionDilach:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f5b80a7101fdbe965f7d7d5d08671550f37ccb91)

OpenRA.Mods.Common/Traits/ExitsDebugOverlay.cs
```diff
@@ -0,0 +1,105 @@
...
+
+using System.Drawing;
+using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Displays `Exit` data for factories.")]
```
Mailaender: You should mention that this only works in combination with `ExitsDebugOverlayManager` now.

OpenRA.Mods.Common/Traits/World/ExitsDebugOverlayManager.cs
```diff
@@ -0,0 +1,65 @@
...
+ */
+#endregion
+
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Commands;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public class ExitsDebugOverlayManagerInfo : ITraitInfo
```
Mailaender: `[Desc("To be attached on to the world actor")]` would be helpful here.

PR #11536, opened 2016-06-26
Add The First Decade metadata to the new content installer.

pchote:
Closes #11523.

PR #11545, opened 2016-06-29
Overhaul RA installed content.

### LINKED ISSUE #11537: Remove RA:Aftermath assets from repo and installer.
#11375 finally gave us the framework we need to manage expansion asset management and installation.

We can now:
* Remove these files from our repo (and installers).
* Add a new required "Aftermath assets" package in the content metadata and set up the disc install metadata.
* Host a new zip on our mirrors with these assets.
* Host a new zip on our mirrors with the base ra + aftermath content for the quick install option.
* While we're at it, host new packages for the base assets and music to replace our current hacked versions.

pchote:
This overhauls the RA asset extraction and downloads with a few goals in mind:
- Aftermath and C&C Desert assets are now managed by the content installer system.  This removes a bunch of binary assets from the repo and cuts almost 1MB from the windows installer download size (according to Appveyor).
- Downloaded asset packages provide identical content to the data installed from disc.
- lores.mix gives us a full set of low resolution icons for beacons and the observer interface.

Depends on #11536.
Closes #11537.


pchote:
backwards compatibility (new playtest <-> last release and new bleed <-> older bleed) is going to be a problem here, and the migrations code won't be able to help.  I think our only viable option will be to namespace the assets by putting them in a different directory.

obrakmann:
In Ruby On Rails, migrations used to have an `up()` and a `down()` method, for up- and downgrading respectively, obviously. Would that help here? We would need a way to store the versions of mods that were last loaded/installed, though. But I think that would be helpful in the long run, anyway, once we allow in-game mod installation where people can jump back and forth between mod versions with just a couple of mouse clicks.

pchote:
I don't think that's worth the effort, and honestly would probably give up on finishing this rework if that became a requirement.  The main reason this sucks so badly is that our old package setup is full of years of hacks and technical debt.  Once everyone migrates to a consistent and non-mangled set of packages then future changes should be limited to simple additions (if we need assets in mix files that aren't included).

obrakmann:
No, it's cool, I'm fine with an "upgrades only"-policy. The mod browser should enforce that then, though, otherwise we'd probably get flooded with bug reports.

pchote:
Yup, that's why I added the changes requested tag.  I'm going to wait for #11536 before I touch this again.

pchote:
Updated.

This now fixes #11665 by completely removing the now-redundant migrations feature.


I have split #11741 to its own ticket because addressing it here would add several hundred more lines to mod.yaml

mods/ra/mod.yaml
```diff
@@ -240,10 +240,12 @@ ModContent:
 			TestFiles: ^Content/ra/v2/expand/expand2.mix, ^Content/ra/v2/expand/hires1.mix, ^Content/ra/v2/expand/lores1.mix, ^Content/ra/v2/expand/chrotnk1.aud, ^Content/ra/v2/expand/fixit1.aud, ^Content/ra/v2/expand/jburn1.aud, ^Content/ra/v2/expand/jchrge1.aud, ^Content/ra/v2/expand/jcrisp1.aud, ^Content/ra/v2/expand/jdance1.aud, ^Content/ra/v2/expand/jjuice1.aud, ^Content/ra/v2/expand/jjump1.aud, ^Content/ra/v2/expand/jlight1.aud, ^Content/ra/v2/expand/jpower1.aud, ^Content/ra/v2/expand/jshock1.aud, ^Content/ra/v2/expand/jyes1.aud, ^Content/ra/v2/expand/madchrg2.aud, ^Content/ra/v2/expand/madexplo.aud, ^Content/ra/v2/expand/mboss1.aud, ^Content/ra/v2/expand/mhear1.aud, ^Content/ra/v2/expand/mhotdig1.aud, ^Content/ra/v2/expand/mhowdy1.aud, ^Content/ra/v2/expand/mhuh1.aud, ^Content/ra/v2/expand/mlaff1.aud, ^Content/ra/v2/expand/mrise1.aud, ^Content/ra/v2/expand/mwrench1.aud, ^Content/ra/v2/expand/myeehaw1.aud, ^Content/ra/v2/expand/myes1.aud
 			Sources: aftermath, aftermath-linux, tfd, ra-origin
 			Required: true
+			Download: aftermath
```
obrakmann: Aftermath wasn't released as freeware, was it?

obrakmann:
Looks fine, :+1:

mods/ra/mod.yaml
pchote: Not explicitly, no.  Nobody seems to care about that fact though and it is far too late for us to remove our dependency on its assets.

obrakmann: Well, we're shipping it already, this is just changing the mode of delivery, so to speak. Guess we'll have to live with it.

mods/ra/mod.yaml
```diff
@@ -8,15 +8,16 @@ RequiresMods:
...
-	~^Content/ra/music
+	~^Content/ra/v2/
+	~^Content/ra/v2/expand
+	~^Content/ra/v2/cnc
 	.
 	$ra: ra
 	$cnc: cnc
 	./mods/common: common
 	~main.mix
-	~redalert.mix
```
Mailaender: This makes the game not detect manually copied installations and attempt to download redundant files.

Mailaender: Guess this is a trade-off for better performance. https://github.com/OpenRA/OpenRA/issues/11741

pchote: one of the pr goals is to guarantee a standard set of installed assets.  Calling this a feature might be a bit far, but it's certainly not an unwanted side effect.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8b66ea95d7e16c270ccbf9b23fee73fed56e52b6...68a508305aff53f9e01854d6566b8327480b7803)

reaperrr:
This regressed the RA shellmap slightly, there's now a tile error on th river near the right soviet base.

Probably a missing file that somehow didn't get included in the RA desert content download.

PR #11547, opened 2016-06-30
Fix debris P03 template in TD jungle and snow terrain

### LINKED ISSUE #11539: template 69 in cnc's snow.yaml and jungle.yaml is of size 1,1 but contains 3 tiles
I bumped into this while working on the map generator. Take a look at the template in jungle.yaml:
```yaml
	Template@69:
		Id: 69
		Images: p03.jun
		Size: 1,1
		PickAny: True
		Category: Debris
		Tiles:
			0: Wall
			1: Clear
			2: Clear
```

Size 1,1 but has 3 tiles? Seems wrong. Other tilesets whichs use the 69 template like temperat.yaml and desert.yaml each only define one tile. This caused a weird bug while working on my map generator. Placing these tiles with the generator caused completely black tiles to be added to the map. My current workaround is to just not place these tiles while adding debris.

reaperrr:
Closes #11539.

### DIFF
--- a/mods/cnc/tilesets/jungle.yaml
+++ b/mods/cnc/tilesets/jungle.yaml
@@ -715,12 +715,9 @@ Templates:
 		Id: 69
 		Images: p03.jun
 		Size: 1,1
-		PickAny: True
 		Category: Debris
 		Tiles:
 			0: Wall
-			1: Clear
-			2: Clear
 	Template@70:
 		Id: 70
 		Images: p04.jun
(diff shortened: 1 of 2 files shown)


Mailaender:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/0ab9baa3283eb012251999796faac0316e092277)

PR #11549, opened 2016-06-30
Fix RA Grenadier damage versus Heavy armor

### LINKED ISSUE #11103: Fix grenadier damage vs heavy and concrete
Currently the grenadier does 5 damage per hit on ALL defense structures because it uses the heavy armor class (the only structures that do)
The grenadier also does its full damage to concrete walls because concrete isn't defined in the weapons

reaperrr:
By increasing it by a factor of 5 (from a measly 5% to still-rather-low 25%).

Closes #11103.

### DIFF
--- a/mods/ra/weapons/other.yaml
+++ b/mods/ra/weapons/other.yaml
@@ -101,6 +101,7 @@ Grenade:
 			Wood: 100
 			Light: 25
 			Heavy: 25
+			Concrete: 100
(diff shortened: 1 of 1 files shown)


Mailaender:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3082b3945d3ddc276ed76c31ad13e31d5f6b0175)

PR #11550, opened 2016-06-30
TD Balance Changes 06302016

AoAGeneral:
These are the following changes for TD:

APC AA Gun projectile speed increase from 1c6 to 2c0.

Flame Tank damage vs light increase from 67 to 100.

Flame Tank explosion damage increase from 100 to 115.

Flame Tank damage spread increase from 341 to 400.

Repair Pad HP increase from 600 to 700.

Artillery Turn Speed increased from 2 to 4.

MCV HP increase from 950 to 1200.

### DIFF
--- a/mods/cnc/weapons/smallcaliber.yaml
+++ b/mods/cnc/weapons/smallcaliber.yaml
@@ -185,7 +185,7 @@ APCGun.AA:
 	Report: gun20.aud
 	ValidTargets: Air
 	Projectile: Bullet
-		Speed: 1c682
+		Speed: 2c0
 		Blockable: false
 	Warhead@1Dam: SpreadDamage
 		Spread: 128
--- a/mods/cnc/rules/structures.yaml
+++ b/mods/cnc/rules/structures.yaml
@@ -524,7 +524,7 @@ FIX:
 	SelectionDecorations:
 		VisualBounds: 72,48
 	Health:
-		HP: 600
+		HP: 700
 	RevealsShroud:
 		Range: 5c0
 	Bib:
--- a/mods/cnc/rules/vehicles.yaml
+++ b/mods/cnc/rules/vehicles.yaml
@@ -15,7 +15,7 @@ MCV:
 		Speed: 71
 		Crushes: crate, infantry
 	Health:
-		HP: 950
+		HP: 1200
 	Armor:
 		Type: Heavy
 	RevealsShroud:
@@ -137,7 +137,7 @@ ARTY:
 		Prerequisites: anyhq, ~techlevel.medium
 		Queue: Vehicle.Nod
 	Mobile:
-		TurnSpeed: 2
+		TurnSpeed: 4
 		Speed: 85
 	Health:
 		HP: 75
--- a/mods/cnc/weapons/explosions.yaml
+++ b/mods/cnc/weapons/explosions.yaml
@@ -1,7 +1,7 @@
 FlametankExplode:
 	Warhead@1Dam: SpreadDamage
 		Spread: 1c0
-		Damage: 100
+		Damage: 115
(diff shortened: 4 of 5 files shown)


mods/cnc/weapons/smallcaliber.yaml
```diff
@@ -185,7 +185,7 @@ APCGun.AA:
 	Report: gun20.aud
 	ValidTargets: Air
 	Projectile: Bullet
-		Speed: 1c682
+		Speed: 2
```
GraionDilach: What?!

Unless you actually aimed this to be evaded literally by anything moving faster than a snail, this is wrong.

GraionDilach: Going from the changelog, I assume you wanted `2c0` here, because that would be an increase.

AoAGeneral:
Actually its faster. Doesn't c = Cell location? I tested with this Anjew on the 2 and it hit point target more accurately then 1c682

mods/cnc/rules/tech.yaml
```diff
@@ -1,6 +1,23 @@
+^TechBuildingHusk:
+	Inherits: ^CivBuildingHusk
+	Health:
+		HP: 1
+	DamageMultiplier:
+		Modifier: 0
+	Targetable:
+		RequiresForceFire: yes
+	Capturable:
```
pchote: From a game design perspective, I think this is a bad idea.  Making dead tech structures repairable removes the incentive and requirement to defend them, and removes the strategy of killing enemy tech structures.

mods/cnc/rules/tech.yaml
```diff
@@ -19,7 +38,7 @@ V19.Husk:
 		Footprint: x
 		Dimensions: 1,1
 	Tooltip:
-		Name: Oil Derrick (Destroyed)
+		Name: Oil Derrick (Recapturable)
```
pchote: In any case, please do not change the tooltips.  They are descriptive, and it *is* a destroyed oil derrick.

AoAGeneral:
While I can agree with the defending and keeping your tech structures alive, It also costs an engineer to capture the tech structures. Oil Derricks do not refund the 500 so you make the strategy of "Do I take it early and risk unable to defend it? Or do I wait to capture it knowing I can defend it?" It still involves that strategy and also enables late game capture and reuse. Which can also mean holding ground in a designated spot is important rather then keeping to regular expansion trees.

GraionDilach:
I'm all up for rebuildable tech structures.

Denying through demolition still works fine, since from what I seen the tech buildings are usually at arbitrary places which are hard to defend, and engineers are quite expensive as well. What I'd suggest though, is to transfer these building husks to be owned by the Creeps player instead of their original owner with the following:
```
SpawnActorOnDeath:
		OwnerType: InternalName
		InternalOwner: Creeps
```
Not doing this will mean that building husks will get destroyed when the player dies.

AoAGeneral:
Tech Building changes reverted due to some disagreements.

GraionDilach:
This needs to be squashed then.

AoAGeneral:
Whats the command for it? I don't see it in the Git Bash help file or on the contribute page @_@

GraionDilach:
https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

GraionDilach:
I'm no balance expert but it does what it states, so I guess it's okay to be given my :+1:.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/dd30e3189d8a66bb9beb1fd65530b8aa58799b3a...8b48fb8cb9a1680e61b748fa69b99c69d33e9b13)

PR #11551, opened 2016-06-30
Lower the case of the TechLevel in all rules.yaml files

abcdefg30:
Since #11364 the case of the `TechLevel`s is lower-case.
This prohibited building in several missions.

Also cleaned a bit up: Removed the `MapOptions` trait where it's not needed.


Phrohdoh:
👍

reaperrr:
:+1:

reaperrr:
No changelog since this error wasn't shipped.

Issue #11087, opened 2016-04-10
Incomplete map upgrade


Mailaender:
While converting format 6 to 11 the following did not work:
* Removal of `UseAsShellmap: False`
* Removal of `Selectable: True`
* Removal of `Options: FragileAlliances: False`
* Translation of `Players: PlayerReference: Race:` to `Players: PlayerReference: Faction:`
* Conversion of actor health from float to int.

pchote:
> Removal of Options: FragileAlliances: False 

The options block doesn't exist at all in format 11.  Are you saying that it somehow persisted after saving?

Mailaender:
Yes, the whole one:

```yaml
Options:
	Crates: False
	Fog: True
	Shroud: True
	FragileAlliances: False
	AllyBuildRadius: False
	StartingCash: 5000
	ConfigurableStartingUnits: False
```

pchote:
Can you please link to the map and the command you used to do the upgrade?

Mailaender:
https://scm.thm.de/openra/openra-develop/source/93d00386f5b73d66c6d26407492be5840a50d0e9:mods/cnc/maps/gdi07 using
```shell
mono --debug OpenRA.Utility.exe cnc --upgrade-map mods/cnc/maps/gdi07 20150101
```

pchote:
For me, the command dies in the middle with "Unsupported engine version. Use the release-20151224 utility to update to that version, and then try again".  The map is left in a half upgraded / inconsistent state (the block removal is done at the end by running the map through a load/save cycle).

Mailaender:
It would be nice if it could still do the map format upgrade though. Most maps don't come with custom rules. This seems to be the downside of combining map and rule upgrades.

pchote:
That could be fixed by only throwing the "unsupported engine version" error if the map actually defines rule overrides.

pchote:
My suggestion was rejected in #11554.

The original issue seems to be resolved, so closing.

PR #10797, opened 2016-02-21
Fix mod updater and clear out legacy rules.

pchote:
Fixes #10793 and applies some desperately needed dejunking by removing all upgrade rules earlier than the last release.

PR #11554, opened 2016-07-01
Fixed the map upgrade needlessly not saving format changes

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/11087.


pchote:
After looking at this in a bit more detail I don't think #11087 is valid.  We also care about upgrading the actor and player blocks, which had rules removed by #10797.

obrakmann:
I don't think this is a good idea. It might work now, but once someone adds (critical) rules to `UpgradeRules.UpgradeActors` or `UpgradeRules.UpgradePlayers`, it will just silently stop working, and I can almost guarantee that no-one will catch it before we release it. Let's just leave the unsupported map formats properly unsupported.

pchote:
I don't think there's anything more to do here, so closing.

PR #11370, opened 2016-05-28
Fix shellmap UI disappearing prematurely when starting a mission or replay.

pchote:
The mission and replay browsers now bind their UI cleanup to the `BeforeGameStart` event, matching what the Lobby UI has always done.  This fixes their UI being removed a tick too early.

PR #11555, opened 2016-07-01
Use BeforeGameStart to remove shellmap UI on editor start.

pchote:
This extends #11370 to also fix the menu -> editor transition.


abcdefg30:
👍

Issue #10416, opened 2016-01-04
Give the admin a 'Ready' checkbox even when in a spectator slot


pchote:
From http://www.sleipnirstuff.com/forum/viewtopic.php?t=19306:

> At present an admin who is speccing has no way to keep the players from starting the game themselves, possibly before everything is properly setup (eg crates turned off, right map).
##

MustaphaTR:
I think this is already implemented and can be closed.

PR #11559, opened 2016-07-02
Added ready checkbox for spectating admins

MUzzell:
Feature #10416

An admin is now able to spectate a game and is provided with a ready checkbox.
ra

Game does not start until the spectating admin is ready.
spec_admin_not_ready

An admin spectating still has the start game option.
two_spectators

cnc

### DIFF
--- a/OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
+++ b/OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
@@ -60,7 +60,8 @@ namespace OpenRA.Mods.Common.Server
 
 		static void CheckAutoStart(S server)
 		{
-			var playerClients = server.LobbyInfo.Clients.Where(c => c.Bot == null && c.Slot != null);
+			// A spectating admin is included for checking these rules
+			var playerClients = server.LobbyInfo.Clients.Where(c => (c.Bot == null && c.Slot != null) || c.IsAdmin);
 
 			// Are all players ready?
 			if (!playerClients.Any() || playerClients.Any(c => c.State != Session.ClientState.Ready))
@@ -188,6 +189,7 @@ namespace OpenRA.Mods.Common.Server
 							client.SpawnPoint = 0;
 							client.Color = HSLColor.FromRGB(255, 255, 255);
 							server.SyncLobbyClients();
+							CheckAutoStart(server);
 							return true;
 						}
 						else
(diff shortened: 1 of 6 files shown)


OpenRA.Mods.Common/ServerTraits/LobbyCommands.cs
```diff
@@ -60,7 +60,7 @@ public static bool ValidateCommand(S server, Connection conn, Session.Client cli
 
 		static void CheckAutoStart(S server)
 		{
-			var playerClients = server.LobbyInfo.Clients.Where(c => c.Bot == null && c.Slot != null);
+			var playerClients = server.LobbyInfo.Clients.Where(c => (c.Bot == null && c.Slot != null) || c.IsAdmin);
```
pchote: Can you please add a comment here explaining that spectating admins are an exception to the normal rules?

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -315,7 +315,8 @@ void ConnectionStateChanged(OrderManager om)
 			var startGameButton = lobby.GetOrNull<ButtonWidget>("START_GAME_BUTTON");
 			if (startGameButton != null)
 			{
-				startGameButton.IsDisabled = () => configurationDisabled() || Map.Status != MapStatus.Available ||
+				startGameButton.IsDisabled = () => (configurationDisabled() && (orderManager.LocalClient != null && !orderManager.LocalClient.IsAdmin)) ||
```
pchote: Most of the checks in `configurationDisabled` are there for technical reasons that shouldn't be ignored (map has bogus rules, admin has the kick dialog open, the start-game order is currently in-flight across the network).  Why do you need this?

OpenRA.Mods.Common/Widgets/Logic/Lobby/LobbyLogic.cs
```diff
@@ -791,6 +792,9 @@ void UpdatePlayerList()
 						template = editableSpectatorTemplate.Clone();
 
 					LobbyUtils.SetupEditableNameWidget(template, null, c, orderManager);
+
+					if (client.IsAdmin)
+						LobbyUtils.SetupEditableReadyWidget(template, null, client, orderManager, Map, !Map.RulesLoaded || Map.InvalidCustomRules);
```
pchote: There is no guarantee that `Map.RulesLoaded` and `Map.InvalidCustomRules` won't change (from false to true) after this widget has been set up, so there's a high chance that you'll end with UI out of sync with the logic.

obrakmann:
There's one change you need to do to ensure consistent behaviour: in `LobbyCommands.cs`, please add a call to `CheckAutostart(server)` in the section for the `spectate` lobby command (in line 192, after the call to `server.SyncLobbyClients`). Without that change, the game wouldn't autostart when the last client with an unticked ready box chooses to be a spectator.

Please also squash all your commits into a single one, thanks.

This otherwise looks very good and will be a very welcome change for tournament matches and the like, :+1:

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/6fc36b6b30ca30c3e5d5a6e86269e6d4610233d2...9205bbe00c42b8f23ec5c4a4a78b311f5925dc9b)

PR #11482, opened 2016-06-19
Rework zip updating

pchote:
Fixes #11203.

PR #11561, opened 2016-07-02
Replace GivesExperience.FriendlyFire and GivesBounty.Stances with ValidStances

### LINKED ISSUE #11546: Replace GivesExperience.FriendlyFire boolean with Stances
`GivesBounty` already works that way.

Would make things more consistent and allow to disable giving experience for neutral actors, for example.

phrohdoh:
~~Depends on~~ #11482  has been merged.
Closes #11546

### DIFF
--- a/OpenRA.Mods.Common/Traits/GivesExperience.cs
+++ b/OpenRA.Mods.Common/Traits/GivesExperience.cs
@@ -19,8 +19,8 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("If -1, use the value of the unit cost.")]
 		public readonly int Experience = -1;
 
-		[Desc("Grant experience for team-kills.")]
-		public readonly bool FriendlyFire = false;
+		[Desc("Stance the attacking player needs to receive the experience.")]
+		public readonly Stance ValidStances = Stance.Neutral | Stance.Enemy;
 
 		public object Create(ActorInitializer init) { return new GivesExperience(init.Self, this); }
 	}
@@ -36,8 +36,10 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void Killed(Actor self, AttackInfo e)
 		{
-			// Prevent TK from giving exp
-			if (e.Attacker == null || e.Attacker.Disposed || (!info.FriendlyFire && e.Attacker.Owner.Stances[self.Owner] == Stance.Ally))
+			if (e.Attacker == null || e.Attacker.Disposed)
+				return;
+
+			if (!info.ValidStances.HasStance(e.Attacker.Owner.Stances[self.Owner]))
 				return;
 
 			var valued = self.Info.TraitInfoOrDefault<ValuedInfo>();
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -185,6 +185,33 @@ namespace OpenRA.Mods.Common.UtilityCommands
 					}
 				}
 
+				if (engineVersion < 20160702)
+				{
+					if (node.Key.StartsWith("GivesExperience"))
+					{
+						var ff = "FriendlyFire";
+						var ffNode = node.Value.Nodes.FirstOrDefault(n => n.Key == ff);
+						if (ffNode != null)
+						{
+							var newStanceStr = "";
+							if (FieldLoader.GetValue<bool>(ff, ffNode.Value.Value))
+								newStanceStr = "Neutral, Enemy, Ally";
+							else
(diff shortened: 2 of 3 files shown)


OpenRA.Mods.Common/Traits/GivesExperience.cs
```diff
@@ -19,8 +20,8 @@ class GivesExperienceInfo : ITraitInfo
 		[Desc("If -1, use the value of the unit cost.")]
 		public readonly int Experience = -1;
 
-		[Desc("Grant experience for team-kills.")]
-		public readonly bool FriendlyFire = false;
+		[Desc("Stance the attacking player needs to receive the experience.")]
+		public readonly Stance[] Stances = { Stance.Neutral, Stance.Enemy };
```
abcdefg30: Shouldn't `Stance ValidStances = Stance.Neutral | Stance.Enemy;` work as well (as we do elsewhere)?

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -185,6 +185,23 @@ internal static void UpgradeActorRules(ModData modData, int engineVersion, ref L
...
 				}
 
+				// GivesExperience was changed from having a FriendlyFire bool to a Stances array.
+				if (engineVersion < 20160702 && node.Key.StartsWith("GivesExperience"))
+				{
+					var ff = "FriendlyFire";
+					var ffNode = node.Value.Nodes.FirstOrDefault(n => n.Key == ff);
+					if (ffNode != null)
+					{
+						if (FieldLoader.GetValue<bool>(ff, ffNode.Value.Value))
```
abcdefg30: It's quite unlikely to happen, but what if an actor has this set to `false` in order to overwrite an inherited `true`?

Phrohdoh: Then the node is removed and that is the end of the upgrade rule, leaving them with `Neutral` and `Enemy`.

OpenRA.Mods.Common/Traits/GivesExperience.cs
Phrohdoh: Yes that also works, I just took this directly from `GivesBounty`.
I can change it but I'd like us to decide on a definitive way forward so this doesn't come up again.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
abcdefg30: Wouldn't the actor we inherit from (who had `true` before) get `Neutral, Enemy, Ally`?
And we inherit that then (although we only want `false`/`Neutral, Enemy`).

OpenRA.Mods.Common/Traits/GivesExperience.cs
abcdefg30: I'd vote for what I suggested.

Phrohdoh:
Updated.

reaperrr:
> OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs:L209: [SA1108] A comment may not be placed within the bracketed statement.

Phrohdoh:
Updated.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -185,6 +185,33 @@ internal static void UpgradeActorRules(ModData modData, int engineVersion, ref L
...
+						var ffNode = node.Value.Nodes.FirstOrDefault(n => n.Key == ff);
+						if (ffNode != null)
+						{
+							var newStanceStr = "";
+							if (FieldLoader.GetValue<bool>(ff, ffNode.Value.Value))
+								newStanceStr = "Neutral, Enemy, Ally";
+							else
+								newStanceStr = "Neutral, Enemy";
+
+							node.Value.Nodes.Add(new MiniYamlNode("Stances", newStanceStr));
```
reaperrr: This needs to be "ValidStances" now.

Phrohdoh: Right-o, will fix.

reaperrr:
Looks good to me :+1:

abcdefg30:

Remove the testcases now, please.

Phrohdoh:
Ready for merge.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-%28bleed%29/_compare/ae7f86975193fb27fca3a5b21f6542f3b01f2c21...4445ba5352a97836eee06aad2d6355036914bf74)

Issue #10475, opened 2016-01-10
Translation support


pchote:
Prerequisites for initial translations:
- [ ] Engine support:
  - [x] Translation support for yaml-defined strings. [#3929]
  - [ ] Translation support for code-defined strings.
  - [ ] Change server messages to send a string ID and arguments to translate client-side.
- [ ] Tooling:
  - [x] Automated chrome string extraction. [#5152]
  - [ ] Automated format string sanity checks (verify that strings with formatting parameters define the same number across all translations).
- [ ] Extract hardcoded strings into english translation.
- [ ] Extract yaml strings into english translation.
- [ ] Include language setting in crash logs.

Later improvements:
- [ ] Support for detecting/downloading/installing foreign content bundles (sound/art).  Related to the ongoing mod browser / installation work.
- [ ] Support for switching art/sounds based on translation.

pchote:
#11375 and followups check off support for detecting/downloading/installing foreign content bundles.

Mailaender:
I believe the whole approach taken is a dead end. Instead of building our own incompatible translation system we should try to hook up something standardized like https://github.com/neris/NGettext where tooling already exists.

Mailaender:
I played a bit with https://crowdin.com/project/openra and yml is supported although we don't adhere to the Ruby translation system. We will need tooling that converts our `@PLACEHOLDERS@` to English, generate English to Language files for the web interface and converter Language to `@PLACEHOLDERS@` back.

abcdefg30:
Is adding translation support to the auto generated documentation also considered as part of this ticket?

pchote:
It is not, and I would argue against doing that (but not in this ticket).

forcecore:
There's font problem too, the default fonts don't support CJK characters. There should be some way of providing additional fonts for the translations.

Sav87:
Why not do a translation like this: there is a file "ra|chrome/ingame-player.yaml" and need a translation for "XXX".
So try to download it from ".\mods\ra\languages\XXX\chrome\ingame-player.yaml", if there is no file try to download ".\mods\ra\chrome\ingame-player.yaml" then the load error is mod.

reaperrr:
In principle I'd be willing to, but I'll need some help getting my head around the "how to".
That's something for "later this year" at best, though.

forcecore:
True, I took a look in the past but I have no idea how.

Sav87:
[...]

PR #11424, opened 2016-06-10
Polish the TS PrimaryBuilding decoration.

GraionDilach:
openra-2016-06-10t142336z

Resolves #10996.

PR #11565, opened 2016-07-03
Added WithTextDecoration

Mailaender:
All games from Dune 2000 onward seem to render these dynamically. This is required for https://github.com/OpenRA/ra2/issues/181. It will also ease localisation of the game https://github.com/OpenRA/OpenRA/issues/10475.


abcdefg30:
👍 / ✅

OpenRA.Mods.Common/Traits/Render/WithTextDecoration.cs
```diff
@@ -0,0 +1,120 @@
...
+namespace OpenRA.Mods.Common.Traits.Render
+{
+	[Desc("Displays a text overlay relative to the selection box.")]
+	public class WithTextDecorationInfo : UpgradableTraitInfo
+	{
+		[FieldLoader.Require] [Translate] public readonly string Text = null;
+
+		public readonly string Font = "TinyBold";
+
+		public readonly Color Color = Color.White;
```
Phrohdoh: Please document and allow usage of player color.

OpenRA.Mods.Common/Traits/Render/WithTextDecoration.cs
```diff
@@ -0,0 +1,120 @@
...
+
+		[Desc("Point in the actor's selection box used as reference for offsetting the decoration image. " +
+			"Possible values are combinations of Center, Top, Bottom, Left, Right.")]
+		public readonly ReferencePoints ReferencePoint = ReferencePoints.Top | ReferencePoints.Left;
+
+		[Desc("The Z offset to apply when rendering this decoration.")]
+		public readonly int ZOffset = 1;
+
+		[Desc("Player stances who can view the decoration.")]
+		public readonly Stance Stances = Stance.Ally;
```
Phrohdoh: `ValidStances` please.

Mailaender: I would like to keep it compatible to `WithDecoration`.

Phrohdoh: I recently fixed `GivesBounty` to use `ValidStances` so  now we have:
```
tmba:openra thill $ ag "Stance Stances" --csharp | wc -l
       1
tmba:openra thill $ ag "Stance\[\] Stances" --csharp | wc -l
       1
tmba:openra thill $ ag "Stance ValidStances" --csharp | wc -l
       4
```

So I think it'd only do us injustice to introduce another `Stances` instead of fixing them both.

reaperrr: Agreed.

Mailaender: Updated according to https://github.com/OpenRA/OpenRA/pull/11561/.

OpenRA.Mods.Common/Traits/Render/WithTextDecoration.cs
Mailaender: I can't come up with a use case, but it is an easy addition so I did it.

reaperrr:
Needs rebase.

Mailaender:
Rebased.

reaperrr:
I just had a look at original TS and noticed two things:
- TS primary tag indeed used player color.
- TS primary tag was not centered, but rather positioned on the "top" of the non-bib area, in other words you need to move it upwards on the helipads and GDI barracks, and up + a bit to the left for weapon factories and Hand of Nod.

Additionally, I noticed you put the PlayerColor changes into 0f8d8a5, I think it would be better to split them into their own commit or squash them into 3c99a3f instead.

Mailaender:
Ah nice, then we also got a test case for that. Reverted https://github.com/OpenRA/OpenRA/pull/11424 and updated accordingly.

reaperrr:
Looking good now and everything works, just one last thing: I'm sorry, but I messed up while testing TS, turns out Westwood just happened to use a gold shade very similar to my player color, but actually didn't use player color.

If you replace the `UsePlayerColor: true` with `Color: E0D048`, this can be merged afterwards.

Phrohdoh:
But do we want gold text when we could use player color?
Was this gold used for Nod too, or was it a shade of red then?

reaperrr:
> Was this gold used for Nod too, or was it a shade of red then?

It was gold for Nod, too.
My other concern about using player color - apart from that the original didn't - is that dark colors might make it hard to read.

Phrohdoh:
Alright, gold works for me then.

Mailaender:
Okay, golden it is now.

Phrohdoh:
Looking good. 👍
screen shot 2016-07-03 at 6 19 09 pm

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3082b3945d3ddc276ed76c31ad13e31d5f6b0175...058348626035889d137ec1a64cafabfcb9fbc305)

Mailaender:
If we want the font to match the original I guess we need https://github.com/OpenRA/OpenRA/issues/11566 or use the TTF conversions from https://forums.cncnet.org/index.php?topic=126.msg397#msg397

pchote:
Where did d2k keep its original font files, and are they compatible with the "proper" C&C ones?

Mailaender:
http://old.d2kplus.com/wiki/index.php?title=.FNT_And_.FPL seems to be a bit different regarding the file format, but the same pixel array in principle.

pchote:
D2K did not render this dynamically.  It used the "1" symbol from DATA.R8 in the bottom-left corner of the selection rectangle.  That format documentation also appears to be incorrect.

PR #11567, opened 2016-07-03
Documented and exposed the control group sprites to lint testing

Mailaender:
Not setting them up will cause problems to modders.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
+++ b/OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
@@ -44,7 +44,7 @@ namespace OpenRA.Mods.Common.Traits.Render
 		public int[] SelectionBoxBounds { get { return VisualBounds; } }
 	}
 
-	public class SelectionDecorations : IPostRenderSelection
+	public class SelectionDecorations : IPostRenderSelection, ITick
 	{
 		// depends on the order of pips in TraitsInterfaces.cs!
 		static readonly string[] PipStrings = { "pip-empty", "pip-green", "pip-yellow", "pip-red", "pip-gray", "pip-blue", "pip-ammo", "pip-ammoempty" };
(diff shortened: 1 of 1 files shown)


abcdefg30:
👍

OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
```diff
@@ -104,9 +109,9 @@ IEnumerable<IRenderable> DrawControlGroup(WorldRenderer wr, Actor self, int2 bas
 			if (group == null)
 				yield break;
 
-			var pipImages = new Animation(self.World, "pips");
+			var pipImages = new Animation(self.World, Info.Image);
```
reaperrr: This should be cached in the constructor.

OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
```diff
@@ -104,9 +109,9 @@ IEnumerable<IRenderable> DrawControlGroup(WorldRenderer wr, Actor self, int2 bas
...
 			var pal = wr.Palette(Info.Palette);
```
reaperrr: Same here, while you're at it.

OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
```diff
@@ -119,7 +124,7 @@ IEnumerable<IRenderable> DrawPips(WorldRenderer wr, Actor self, int2 basePositio
 			if (!pipSources.Any())
 				yield break;
 
-			var pipImages = new Animation(self.World, "pips");
+			var pipImages = new Animation(self.World, Info.Image);
```
reaperrr: Same here.

reaperrr: and same for [line 133](https://github.com/Mailaender/OpenRA/blob/3a7558c273bdebd2e7466173331f1100d9ece1e8/OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs#L133)

reaperrr:
I think this might be a good opportunity to unhardcode [PipStrings](https://github.com/Mailaender/OpenRA/blob/3a7558c273bdebd2e7466173331f1100d9ece1e8/OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs#L50) as `PipSequences` with the comment added to the property description.

OpenRA.Mods.Common/Traits/Render/SelectionDecorations.cs
Mailaender: I don't have access to `WorldRenderer` in the constructor.

pchote:
In the long term we should move the pip definitions out of this trait completely (to the traits that want to show the pips).

reaperrr:
> I will stay clear of PipStrings for now as it will require all pips to be setup.

Fair enough.

PR #11572, opened 2016-07-03
Add building depth imposters for TS base structures.

pchote:
This takes us another small step towards enabling the new depth rendering in TS.

When testing you can quickly toggle between normal and depth rendering using the <kbd>`</kbd> key.


pchote:
Note that this sets `EnableDepthBuffer: True` in mod.yaml, which breaks rendering of many things (including voxels, lasers, sprites being drawn partially under the terrain).  This PR is fixing one corner of that problem space.

OpenRA.Game/Widgets/RootWidget.cs
```diff
@@ -47,6 +48,10 @@ public override bool HandleKeyPress(KeyInput e)
 
 					return true;
 				}
+
+				var world = Game.OrderManager.World;
+				if (world != null && world.LocalPlayer != null && hk == new Hotkey(Keycode.BACKQUOTE, Modifiers.None))
```
Mailaender: People with non-US keyboards can use `new Hotkey(Keycode.D, Modifiers.Ctrl)` instead.

OpenRA.Mods.Common/Traits/Render/WithGateSpriteBody.cs
```diff
@@ -26,6 +26,9 @@ class WithGateSpriteBodyInfo : WithSpriteBodyInfo, IWallConnectorInfo, Requires<
 		[Desc("Wall type for connections")]
 		public readonly string Type = "wall";
 
+		[Desc("Override sequence to use when fully open (if non-null).")]
```
Mailaender: The `(if non-null)` isn't obvious when looking at just the traits documentation. Just skip it.

Mailaender:
This does something. :white_check_mark:  I have no idea if it is correct though.

Phrohdoh:
c038c36f901e269999c440cb95147af70fb9cc0a does more than what the commit message says, though that is one of the uses.

These look good to me. The east and west-most corners of structures still have some issues but I think those would need custom depth sprites to look any better. 👍

I couldn't convince you to do the same for RA2, could I?

GraionDilach:
That depth offset looks like something which can be added to the building art importer to me, atleast as a rough estimate by duplicating the second offset.

Why does it need a new sprite though and why can't buildngz.shp be used?

Also why did the Wolverine got additional z-offset as well, but the Titan/deployed MSA/tick tanks didn't?

pchote:
* Depth importing will be part of a later pr that sets up the civilian actors.
* TS manages its depth buffer differently, and its buildngz.shp doesn't work for us.
* The wolverine is there in the testcase commit to help test depth sorting only.  This PR isn't about vehicles (which will remain completely broken until voxel rendering is fixed)

obrakmann:
:+1:, I guess. Please remove the test-case.

PR #11573, opened 2016-07-03
Expose the WeatherOverlay to lua

### LINKED ISSUE #11538: Lua Scripted WeatherOverlay Commands
It would be nice if you can control the WeatherOverlay trait through Lua. You can already control the Lighting with Lua.

abcdefg30:
Closes #11538.

A testcase can be found at my [lwTest branch](https://github.com/abcdefg30/OpenRA/tree/lwTest).

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Global/WeatherGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/WeatherGlobal.cs
@@ -1,0 +1,170 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using System.Drawing;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Scripting;
+
+namespace OpenRA.Mods.Common.Scripting.Global
+{
+	[ScriptGlobal("Weather")]
+	public class WeatherGlobal : ScriptGlobal
+	{
+		readonly WeatherOverlay overlay;
+		readonly bool hasAnOverlay;
+
+		public WeatherGlobal(ScriptContext context)
+			: base(context)
+		{
+			overlay = context.World.WorldActor.TraitOrDefault<WeatherOverlay>();
+			hasAnOverlay = overlay != null;
+		}
+
+		public bool Active
+		{
+			get { return hasAnOverlay && overlay.Active; }
+			set { if (hasAnOverlay) overlay.Active = (bool)value; }
+		}
+
+		[Desc("Average number of particles per 1000x1000 px square.")]
+		public int ParticleDensityFactor
+		{
+			get
+			{
+				return hasAnOverlay ? (int)overlay.ParticleDensityFactor : 0;
+			}
+
+			set
+			{
+				if (!hasAnOverlay)
+					return;
+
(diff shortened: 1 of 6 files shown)


OpenRA.Mods.Common/Scripting/Global/WeatherGlobal.cs
```diff
@@ -0,0 +1,165 @@
...
+				if (!hasAnOverlay)
+					return;
+
+				overlay.ParticleDensityFactor = (int)value;
+				overlay.CalculateMaxParticleCount();
+			}
+		}
+
+		[Desc("Should the level of the wind change over time, or just stick to the first value of WindLevels?")]
+		public bool ChangingWindLevel
```
pchote: Do we really need to expose all this to lua?  Scripts can modify things themselves if they want it to evolve.

pchote:
Could we use upgrades to switch between different instances of the weather trait defined in map.yaml?

OpenRA.Mods.Common/Scripting/Global/WeatherGlobal.cs
pchote: Directly mapping the trait's properties into lua will make it much more difficult if we want to change the way the trait works in the future (which we know that we need to, e.g. wind).

abcdefg30:
> Could we use upgrades to switch between different instances of the weather trait defined in map.yaml?

This won't give us the accurancy one might want, or would force people into having hundreds of `WeatherOverlay` definition for a soft change from e.g. heavy rain to almost no rain, wouldn't it?

pchote:
The weather trait can implement upgrade enabled/disabled handlers to fade the effect in/out to hide the transition between states.

pchote:
Repeating/paraphrasing here since the logs aren't available for context:

My main objection is that this goes against our long-standing rule that Lua APIs should be task oriented and not directly bridge trait properties.  This is especially true for the weather trait because (a) its properties are low level and a bit horrible, and (b) we already know that these need to change (extracting wind out so that it can also affect gas projectiles in TS), and baking the current trait setup into our nominally-stable Lua API makes that much harder.

abcdefg30:
Alright, so the next thing to get this through would be extracting wind to a different trait, and then query this new trait inside the `WeatherOverlay`?

I would extract following fields:
- `bool ChangingWindLevel`
- `int[] WindLevels`
- `int[] WindTick`
- `bool InstantWindChanges`

Mailaender:
I assume https://github.com/OpenRA/OpenRA/issues/11641 was the dependency.

Mailaender:
I have to agree with @pchote that keeping properties at trait level is easier to use and can be automatically updated when the weather code gets refactored. It would be better to have the WeatherOverlay upgradeable instead.

abcdefg30:
> I assume #11641 was the dependency.

Actually this is the depedency:
> Alright, so the next thing to get this through would be extracting wind to a different trait, and then query this new trait inside the `WeatherOverlay`?

Mailaender:
Even after a potential refactor for Tiberian Sun we don't really want this directly exposed to Lua and turn on/off traits functionality already exists with upgrades so just make the weather trait upgradeable. For extra polish you can implement the transitioning during upgrades.

Mailaender:
That would also a first step towards https://github.com/OpenRA/ra2/issues/149.

GraionDilach:
Uhm, I don't see how the Weather Control Device or ion storms (on-map effects) are related to our current weather overlay (whole-screen-covering visual). In fact, RA2 WCD shouldn't even use WeatherOverlay at all.

Mailaender:
I thought about adding not just a palette effect upgrade, but also a stormy weather upgrade triggered by the Weather Control Device which would give it more polish.

abcdefg30:
Closing this for now, as I won't have time for it in the next week.

PR #11574, opened 2016-07-03
Prevent non-combatant players from receiving units from crate pickups

obrakmann:


pchote:
Looks reasonable to me, but not tested :+0.5:

Phrohdoh:
Is there any reason the same shouldn't be done to `DuplicateUnitCrateAction`?

obrakmann:
A couple:

1. I forgot about it ;)
2. It wouldn't cause civilians to get units that they don't have access to. They just get another civ unit, which is ok by me.

GraionDilach:
I'm not certain if I'd like visceroids getting duplicated.

obrakmann:
Good point, updated.

GraionDilach:
:+1:

Mailaender:
No more random neutral MCVs appearing on the battlefield. ;)

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/4c7bee14330cd2530d78d9a57191549779e4e8e2...e5c139a0cc5aaed95c04aa8aa0e59f5fcc11cec8)

PR #11578, opened 2016-07-04
Don't add duplicate facing entries when modifying the DeathActorInit in Turreted.cs

abcdefg30:
Causes a crash when having multiple `SpawnActorOnDeath` traits on a single turreted actor.


Mailaender:
:+1: what could go wrong.

GraionDilach:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/32551f2174399fc8b5d84c2d0c06b263ee5c524f)

PR #11586, opened 2016-07-06
Expand SelfHealing.

GraionDilach:
Adds a percentage healing option and damagetypes.
Depends on #11513 (also can be used as a justification of that PR).

Includes testcase.

### DIFF
--- a/OpenRA.Mods.Common/Traits/SelfHealing.cs
+++ b/OpenRA.Mods.Common/Traits/SelfHealing.cs
@@ -16,10 +16,18 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Attach this to actors which should be able to regenerate their health points.")]
 	class SelfHealingInfo : UpgradableTraitInfo, Requires<HealthInfo>
 	{
+		[Desc("Absolute amount of health points added in each step.")]
 		public readonly int Step = 5;
+
+		[Desc("Relative percentages of health added in each step.",
+			"When both values are defined, their summary will be applied.")]
+		public readonly int PercentageStep = 0;
+
 		public readonly int Delay = 5;
+
 		[Desc("Heal if current health is below this percentage of full health.")]
 		public readonly int HealIfBelow = 50;
+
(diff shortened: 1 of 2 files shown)


GraionDilach:
Rebased, no longer needs dependency.

GraionDilach:
Would it make sense if I'd replace the Dune2000 elite selfhealing with a 4% percentage heal? Considering that even the Rifle Infantry has 600 HP there, the current +2 HP per 125 frames doesn't really make a difference.

OpenRA.Mods.Common/Traits/SelfHealing.cs
```diff
@@ -9,19 +9,30 @@
...
 
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Attach this to actors which should be able to regenerate their health points.")]
 	class SelfHealingInfo : UpgradableTraitInfo, Requires<HealthInfo>
 	{
+		[Desc("Absolute amount of health points added in each step.")]
 		public readonly int Step = 5;
+
+		[Desc("Relative percentages of health added in each step.")]
```
obrakmann: In the description, can you mention the fact that the sum of `Step` and `PercentageStep` gets applied when both are defined, please?

obrakmann:
:+1: 

Yeah, go for it. With the current self-healing, some infantry units need almost 20 minutes to heal back to full health from near death (if I didn't screw up my maths). With the 4% option, it would be about two minutes for all units.

OpenRA.Mods.Common/Traits/SelfHealing.cs
GraionDilach: Added.

GraionDilach:
Replaced the earlier testcommit with the D2k selfhealing change.

pchote:
Looks fine to me :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d1b3edd232e89143477bf2710f43c6f6e46cf924

Issue #11566, opened 2016-07-03
FNT font format support


Mailaender:
The original games include their own fonts. See http://xhp.xwis.net/documents/ccfonts.txt for the file format. There are even editors for them: http://forums.cncnz.com/topic/8223-ra2yr-font-editor-released/

Mailaender:
The [Westwood Font Editor](http://www.ppmforums.com/viewtopic.php?t=42926) might be a good source of inspiration.

Nyerguds:
Well, I got c# code for converting the fonts to 8-bit paletted .Net images. Though they don't include Y-offset; the editor does that in UI. But the font class also contains code for expanding the image and shifting image data in all directions, so it'd just be a couple of lines of code to add it.

And it's released under WTFPL, so, go nuts with it.

Nyerguds:
The editor has gotten a true preview function, meaning the font class actually has a function to generate text as image now.

chrisforbes:
What do we intend to use this for in the game? It seems a poor replacement for anywhere player names or chat may show up, as those need a much wider range of characters.

pchote:
TS and D2K (to a lesser extent) used these fonts for ingame decorations ("Primary" building label, production icon labels, etc).  The argument is that we should generate these using the original assets instead of making and shipping our own.

Nyerguds:
Do note, the original D2K fonts have been found to contain some errors. In "FONTCOL.FNT", the double quote, hash sign and underscore symbols have errors, as well as the lowercase letter 'q'.

Mailaender:
> What do we intend to use this for in the game?

Not just us. Nowadays we build OpenRA for the modding community and they are used to their tools and file formats.

I also thought about simply using http://www.dafont.com/de/c-c-red-alert-inet.font but it is again legal trouble as it was copied from the original copyrighted FNT and then again with another copyright added on top of that (for the retrace) from @N3tRunn3r so essentially worseless for us.

This is related to https://github.com/OpenRA/OpenRA/issues/8264 but not really a blocker. There are more severe cases of IP trouble in this repository. See also https://bugzilla.redhat.com/show_bug.cgi?id=1159091#c16

Nyerguds:
Some notes on Dune 2000:

[...]

PR #11587, opened 2016-07-06
Overhaul D2K installed content.

### LINKED ISSUE #2344: Support Dune 2000 patch 1.06 gamefiles

### LINKED ISSUE #11351: D2k Map Editor: Grenadier bounds aren't cropped
Compared to everything else in every mod AFAIK.

openra-2016-05-24t143142z

pchote:
This switches D2K to the 1.06 content, which lets us remove a bunch of files from the bits directory, and lets us install content from an existing gruntmods d2k install.

The new download pages are temporarily hosted on dropbox until we can get some mirrors online (cc @baxxxster).

Fixes #2344.
Fixes #11351.


GraionDilach:
I assume this fixes #11351 then?

pchote:
It does, yes.

pchote:
Note that this also installs the three font-related files from the original game.  I'm expecting we will use one or more of these to improve on #11565 before we ship the next release.  We can then update the mirrored content to remove the ones we don't need.

pchote:
Updated.

pchote:
Since nobody seems to be in a rush to get a new playtest out i'm going to make this depend on #11566 so that we can sort out which fonts we extract sooner rather than later.

pchote:
After some research I can say that we only really need the .FNT and FON files.  The FPL colours don't match the white and black/grey that we will want to use ingame, so we'll need to use a custom palette there.  That is a small file, so there's no harm in including it.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -281,6 +281,36 @@ internal static void UpgradeWeaponRules(ModData modData, int engineVersion, ref
...
+			if (frame < 4036)
+				return frame + 261;
+			return frame + 264;
+		}
+
+		internal static void UpgradeSequences(ModData modData, int engineVersion, ref List<MiniYamlNode> nodes, MiniYamlNode parent, int depth)
+		{
+			foreach (var node in nodes)
+			{
+				if (engineVersion < 20160705 && modData.Manifest.Mod.Id == "d2k" && depth == 2)
```
obrakmann: the date needs to be updated before merging.

obrakmann:
Tested with the downloadable content, looks fine to me, :+1:

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
pchote: I'll wait until a second :+1: to avoid having to repeat this again.

mods/d2k/mod.yaml
```diff
@@ -1930,3 +1931,296 @@ ModContent:
 					^Content/d2k/v2/SOUND.RS:
 						Offset: 20559431
 						Length: 1929247
+		gruntmods: Dune 2000: GruntMods Edition
+			Type: Install
+			RegistryKey: HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Dune 2000: Gruntmods Edition
+			IDFiles:
+				Dune 2000/DUNE2000.EXE: 24ac183eb6fd17f63859f4beb26ad505cc8213ed
+				Dune 2000/Launch Dune 2000.exe: 4877ad15e059ad8c0ba9a19bb08de6ba9858b0ae
```
Mailaender: I assume these will change with every new binary patch or @cncnet client update.

Mailaender:
My comment https://github.com/OpenRA/OpenRA/pull/11587#discussion_r71293815 is hidden and I can't comment on the `mod.yaml` as the diff is so large, but I wouldn't recommend ID-ing the https://gruntmods.com/dune_2000_gruntmods_edition/ on volatile .EXE files we don't even care about. Just make sure the DATA.R8 etc. isn't corrupted.

Otherwise looking good. :+1:

pchote:
Updated.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/75c5c0ce6ddc4ae242f6991346eaabfcc19fe208)

PR #11592, opened 2016-07-07
Fix smudges spawning/growing on cells occupied by structures or vehicles

### LINKED ISSUE #6464: Craters/Scorches are spawned even if something was hit
That might be ok if the target is a small infantry unit, but a crater forming below a mammoth tank or building looks weird.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4364633-craters-scorches-are-spawned-even-if-something-was-hit?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
With this, the `LeaveSmudge` warhead checks the target type(s) of actors on affected cells, rather than only caring about the terrain.

This can be used to prevent craters or scorches spawning and growing under buildings and vehicles, which matches the behavior in the originals and looks more realistic anyway.

Also, but not related, the explosion sound of TD artillery weapon was not playing because it was listed as uppercase, this is fixed as well.

Fixes #6464.

### DIFF
--- a/OpenRA.Mods.Common/Warheads/LeaveSmudgeWarhead.cs
+++ b/OpenRA.Mods.Common/Warheads/LeaveSmudgeWarhead.cs
@@ -38,7 +38,12 @@ namespace OpenRA.Mods.Common.Warheads
 			foreach (var sc in allCells)
 			{
 				var smudgeType = world.Map.GetTerrainInfo(sc).AcceptsSmudgeType.FirstOrDefault(SmudgeType.Contains);
-				if (smudgeType == null) continue;
+				if (smudgeType == null)
+					continue;
+
+				var cellActors = world.ActorMap.GetActorsAt(sc);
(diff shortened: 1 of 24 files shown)


reaperrr:
Note: Would be nice to get this in before the playtest, I think it's some nice polish with little regression risk.

GraionDilach:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d94f6540823b3a6aef767bb1477ddb1fa23139d6...e5c3a00c6563d7e851cb646cb184997b56a12dfd)

PR #11596, opened 2016-07-07
Fix a wrong check in LeavesTrails.

### LINKED ISSUE #11595: LeavesTrails is bugged.
tfdfguf

When a unit starts moving, it skips the first cell. TS mod.

GraionDilach:
dfj

Fixes #11595.

Checking for WasStationary there was redundant.

The bug is caused by the WasStationary check - if StartDelay is 0 then it will imminently spawn a new trail in the starting position and will load the stationary delay for moving actors.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/LeavesTrails.cs
+++ b/OpenRA.Mods.Common/Traits/Render/LeavesTrails.cs
@@ -98,8 +98,11 @@ namespace OpenRA.Mods.Common.Traits.Render
 			if ((isMoving && !Info.TrailWhileMoving) || (!isMoving && !Info.TrailWhileStationary))
 				return;
 
-			if (isMoving && wasStationary)
+			if (isMoving == wasStationary)
+			{
 				cachedInterval = Info.StartDelay;
+				ticks = 0;
+			}
 
 			if (++ticks >= cachedInterval)
 			{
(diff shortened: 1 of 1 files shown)


reaperrr:
:+1:

OpenRA.Mods.Common/Traits/Render/LeavesTrails.cs
```diff
@@ -125,7 +125,7 @@ public void Tick(Actor self)
 				cachedFacing = facing != null ? facing.Facing : 0;
 				ticks = 0;
 
-				cachedInterval = isMoving && !wasStationary ? Info.MovingInterval : Info.StationaryInterval;
+				cachedInterval = isMoving ? Info.MovingInterval : Info.StationaryInterval;
```
pchote: This unconditionally overrides the `cachedInterval = Info.StartDelay` above, breaking that feature.

GraionDilach: It doesn't, this still happens within the if (++ticks >= cachedInterval) block, which only runs if StartDelay is smaller than ticks.

pchote: Ah yes, I see.  Wouldn't it make sense to reset ticks to 0 when transitioning from stationary to moving?

GraionDilach: Yes. Yes, it does. Also when transitioning from moving to stationary as well, so I updated that check accordingly to that.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/24f71638e24b866f075599af52e4dcd9a4b7cba1)

PR #11600, opened 2016-07-08
Allow multiple AttackTurreted traits.

GraionDilach:
Basically #10900 reopened - should have reopened the PR before I pushed though.

### DIFF
--- a/OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
+++ b/OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
@@ -39,25 +39,30 @@ namespace OpenRA.Mods.Cnc.Traits
 		public override object Create(ActorInitializer init) { return new AttackPopupTurreted(init, this); }
 	}
 
-	class AttackPopupTurreted : AttackTurreted, INotifyBuildComplete, INotifyIdle, IDamageModifier
+	class AttackPopupTurreted : AttackTurreted, INotifyBuildComplete, INotifyIdle, IDamageModifier, INotifyCreated
 	{
 		enum PopupState { Open, Rotating, Transitioning, Closed }
 
 		readonly AttackPopupTurretedInfo info;
 		readonly WithSpriteBody wsb;
-		readonly Turreted turret;
 
 		int idleTicks = 0;
 		PopupState state = PopupState.Open;
 		bool skippedMakeAnimation;
+		Turreted turret;
 
 		public AttackPopupTurreted(ActorInitializer init, AttackPopupTurretedInfo info)
 			: base(init.Self, info)
 		{
 			this.info = info;
-			turret = turrets.FirstOrDefault();
 			wsb = init.Self.Trait<WithSpriteBody>();
 			skippedMakeAnimation = init.Contains<SkipMakeAnimsInit>();
+		}
+
+		void INotifyCreated.Created(Actor self)
+		{
+			SetupTurrets(self);
+			turret = turrets.Single();
 		}
 
 		protected override bool CanAttack(Actor self, Target target)
--- a/OpenRA.Mods.Cnc/Traits/Render/WithReloadingSpriteTurret.cs
+++ b/OpenRA.Mods.Cnc/Traits/Render/WithReloadingSpriteTurret.cs
@@ -60,7 +60,7 @@ namespace OpenRA.Mods.Cnc.Traits.Render
 		public override void Tick(Actor self)
 		{
 			if (Info.AimSequence != null)
-				sequence = Attack.IsAttacking ? Info.AimSequence : Info.Sequence;
+				sequence = Attack.Any(a => a.IsAttacking && !a.IsTraitDisabled) ? Info.AimSequence : Info.Sequence;
 
 			var currentAmmo = ammoPool.GetAmmoCount();
 			if (reloadStages < 0)
--- a/OpenRA.Mods.Common/Activities/Hunt.cs
+++ b/OpenRA.Mods.Common/Activities/Hunt.cs
@@ -23,10 +23,11 @@ namespace OpenRA.Mods.Common.Activities
 
 		public Hunt(Actor self)
 		{
-			var attack = self.Trait<AttackBase>();
+			var attack = self.TraitsImplementing<AttackBase>();
 			targets = self.World.ActorsHavingTrait<Huntable>().Where(
 				a => self != a && !a.IsDead && a.IsInWorld && a.AppearsHostileTo(self)
-				&& a.IsTargetableBy(self) && attack.HasAnyValidWeapons(Target.FromActor(a)));
+				&& a.IsTargetableBy(self)
+				&& attack.Any(att => !att.IsTraitDisabled && att.HasAnyValidWeapons(Target.FromActor(a))));
 		}
 
 		public override Activity Tick(Actor self)
--- a/OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
+++ b/OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
@@ -78,12 +78,12 @@ namespace OpenRA.Mods.Common.Scripting
 	[ScriptPropertyGroup("Combat")]
 	public class GeneralCombatProperties : ScriptActorProperties, Requires<AttackBaseInfo>
 	{
-		readonly AttackBase attackBase;
+		readonly AttackBase[] attackBase;
 
 		public GeneralCombatProperties(ScriptContext context, Actor self)
 			: base(context, self)
 		{
-			attackBase = self.Trait<AttackBase>();
+			attackBase = self.TraitsImplementing<AttackBase>().ToArray();
 		}
 
 		[Desc("Attack the target actor. The target actor needs to be visible.")]
@@ -96,7 +96,12 @@ namespace OpenRA.Mods.Common.Scripting
 			if (!targetActor.Info.HasTraitInfo<FrozenUnderFogInfo>() && !Self.Owner.CanTargetActor(targetActor))
 				Log.Write("lua", "{1} is not revealed for player {0}!", Self.Owner, targetActor);
 
-			attackBase.AttackTarget(target, true, allowMove, forceAttack);
+			var attack = attackBase.Where(x => !x.IsTraitDisabled && x.HasAnyValidWeapons(target));
+			if (!attack.Any())
+				Log.Write("lua", "Actor {0} was ordered to attack {1} but has no valid Attack* trait to do so!.".F(Self, targetActor));
+
+			foreach (var a in attack)
+				a.AttackTarget(target, true, allowMove, forceAttack);
 		}
 	}
 }
--- a/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
@@ -10,6 +10,7 @@
 #endregion
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Activities;
 using OpenRA.Traits;
@@ -61,7 +62,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		class AttackActivity : Activity
 		{
-			readonly AttackFollow attack;
+			readonly AttackFollow[] attackArray;
 			readonly IMove move;
 			readonly Target target;
 			readonly bool forceAttack;
@@ -70,7 +71,7 @@ namespace OpenRA.Mods.Common.Traits
 
 			public AttackActivity(Actor self, Target target, bool allowMove, bool forceAttack)
 			{
-				attack = self.Trait<AttackFollow>();
+				attackArray = self.TraitsImplementing<AttackFollow>().ToArray();
 				move = allowMove ? self.TraitOrDefault<IMove>() : null;
 
 				// HACK: Mobile.OnRails is horrible. Blergh.
@@ -93,9 +94,10 @@ namespace OpenRA.Mods.Common.Traits
 				if (self.IsDisabled())
 					return this;
 
-				var weapon = attack.ChooseArmamentsForTarget(target, forceAttack).FirstOrDefault();
-				if (weapon != null)
+				var attack = SelectActiveAttack(target, forceAttack).FirstOrDefault();
+				if (attack.Key != null)
 				{
+					var weapon = attack.Value;
 					var targetIsMobile = (target.Type == TargetType.Actor && target.Actor.Info.HasTraitInfo<IMoveInfo>())
 						|| (target.Type == TargetType.FrozenActor && target.FrozenActor.Info.HasTraitInfo<IMoveInfo>());
 
@@ -106,10 +108,10 @@ namespace OpenRA.Mods.Common.Traits
 
 					// Check that AttackFollow hasn't cancelled the target by modifying attack.Target
 					// Having both this and AttackFollow modify that field is a horrible hack.
-					if (hasTicked && attack.Target.Type == TargetType.Invalid)
+					if (hasTicked && attack.Key.Target.Type == TargetType.Invalid)
 						return NextActivity;
 
-					attack.Target = target;
+					attack.Key.Target = target;
 					hasTicked = true;
 
 					if (move != null)
@@ -121,9 +123,19 @@ namespace OpenRA.Mods.Common.Traits
 				}
 
 				if (!onRailsHack)
-					attack.Target = Target.Invalid;
+					attack.Key.Target = Target.Invalid;
 
 				return NextActivity;
+			}
+
+			IEnumerable<KeyValuePair<AttackFollow, Armament>> SelectActiveAttack(Target t, bool forceAttack)
+			{
+				foreach (var a in attackArray)
+					if (!a.IsTraitDisabled)
+						foreach (var w in a.ChooseArmamentsForTarget(t, forceAttack))
+							yield return new KeyValuePair<AttackFollow, Armament>(a, w);
+
+				yield break;
 			}
 		}
 	}
--- a/OpenRA.Mods.Common/Traits/Attack/AttackTurreted.cs
+++ b/OpenRA.Mods.Common/Traits/Attack/AttackTurreted.cs
@@ -20,14 +20,23 @@ namespace OpenRA.Mods.Common.Traits
 		public override object Create(ActorInitializer init) { return new AttackTurreted(init.Self, this); }
 	}
 
-	public class AttackTurreted : AttackFollow
+	public class AttackTurreted : AttackFollow, INotifyCreated
 	{
 		protected Turreted[] turrets;
 
 		public AttackTurreted(Actor self, AttackTurretedInfo info)
-			: base(self, info)
+			: base(self, info) { }
+
+		protected void SetupTurrets(Actor self)
 		{
-			turrets = self.TraitsImplementing<Turreted>().ToArray();
+			var armamentturrets = getArmaments().Select(x => x.Info.Turret).ToHashSet();
+
+			turrets = self.TraitsImplementing<Turreted>().Where(x => armamentturrets.Contains(x.Name)).ToArray();
+		}
+
+		void INotifyCreated.Created(Actor self)
+		{
+			SetupTurrets(self);
 		}
 
 		protected override bool CanAttack(Actor self, Target target)
@@ -38,7 +47,7 @@ namespace OpenRA.Mods.Common.Traits
 			// Don't break early from this loop - we want to bring all turrets to bear!
 			var turretReady = false;
 			foreach (var t in turrets)
-				if (t.FaceTarget(self, target))
+				if (t.FaceTarget(self, target, this) && !t.IsTraitDisabled)
 					turretReady = true;
 
 			return turretReady && base.CanAttack(self, target);
--- a/OpenRA.Mods.Common/Traits/Buildings/Building.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/Building.cs
@@ -186,7 +186,7 @@ namespace OpenRA.Mods.Common.Traits
 		public void Created(Actor self)
 		{
 			if (SkipMakeAnimation || !self.Info.HasTraitInfo<WithMakeAnimationInfo>())
-				NotifyBuildingComplete(self);
+				self.World.AddFrameEndTask(w => NotifyBuildingComplete(self));
 		}
 
 		public virtual void AddedToWorld(Actor self)
--- a/OpenRA.Mods.Common/Traits/Render/WithSpriteTurret.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithSpriteTurret.cs
@@ -65,7 +65,7 @@ namespace OpenRA.Mods.Common.Traits.Render
 	public class WithSpriteTurret : UpgradableTrait<WithSpriteTurretInfo>, INotifyBuildComplete, INotifySold, INotifyTransform, ITick, INotifyDamageStateChanged
 	{
 		public readonly Animation DefaultAnimation;
-		protected readonly AttackBase Attack;
+		protected readonly AttackBase[] Attack;
 		readonly RenderSprites rs;
 		readonly BodyOrientation body;
 		readonly Turreted t;
@@ -79,7 +79,7 @@ namespace OpenRA.Mods.Common.Traits.Render
 		{
 			rs = self.Trait<RenderSprites>();
 			body = self.Trait<BodyOrientation>();
-			Attack = self.TraitOrDefault<AttackBase>();
+			Attack = self.TraitsImplementing<AttackBase>().ToArray();
 			t = self.TraitsImplementing<Turreted>()
 				.First(tt => tt.Name == info.Turret);
 			arms = self.TraitsImplementing<Armament>()
@@ -124,7 +124,7 @@ namespace OpenRA.Mods.Common.Traits.Render
 			if (Info.AimSequence == null)
 				return;
 
-			var sequence = Attack.IsAttacking ? Info.AimSequence : Info.Sequence;
+			var sequence = Attack.Any(a => a.IsAttacking && !a.IsTraitDisabled) ? Info.AimSequence : Info.Sequence;
 			DefaultAnimation.ReplaceAnim(sequence);
 		}
 
--- a/OpenRA.Mods.Common/Traits/Turreted.cs
+++ b/OpenRA.Mods.Common/Traits/Turreted.cs
@@ -11,12 +11,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using OpenRA.Primitives;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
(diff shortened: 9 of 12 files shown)


OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -85,32 +86,30 @@ public static Func<int> TurretFacingFromInit(IActorInitializer init, int def, st
...
+			: base(info)
 		{
 			this.info = info;
 			TurretFacing = TurretFacingFromInit(init, info.InitialFacing, info.Turret)();
 		}
 
 		public void Created(Actor self)
 		{
-			attack = self.TraitOrDefault<AttackTurreted>();
+			attack = self.TraitsImplementing<AttackTurreted>().ToArray();
```
pchote: It would make sense for AttackTurreted to list the turrets that it controls.

GraionDilach: Sure, since this is only used in Turreted.Tick() which could be shifted to AttackTurreted, I guess.

GraionDilach: On a second thought, Turreted.Tick() cannot be shifted, because there can be cases where more AttackTurreted traits control the same turret.

abcdefg30:
Needs a rebase.

OpenRA.Mods.Cnc/Traits/Render/WithReloadingSpriteTurret.cs
```diff
@@ -60,7 +60,7 @@ public WithReloadingSpriteTurret(Actor self, WithReloadingSpriteTurretInfo info)
 		public override void Tick(Actor self)
 		{
 			if (Info.AimSequence != null)
-				sequence = Attack.IsAttacking ? Info.AimSequence : Info.Sequence;
+				sequence = Attack.Any(a => a.IsAttacking && !a.IsTraitDisabled) ? Info.Sequence : Info.AimSequence;
```
abcdefg30: Is the swapping of the sequences here intentional?

OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -85,32 +86,30 @@ public static Func<int> TurretFacingFromInit(IActorInitializer init, int def, st
...
 			facing = self.TraitOrDefault<IFacing>();
 			body = self.Trait<BodyOrientation>();
 		}
 
 		public virtual void Tick(Actor self)
 		{
 			// NOTE: FaceTarget is called in AttackTurreted.CanAttack if the turret has a target.
-			if (attack != null)
+			if (!attack.Any(a => a.IsAttacking && !a.IsTraitDisabled))
```
abcdefg30: If we don't have any `AttackTurreted` traits this condition will be true (although it shouldn't).

OpenRA.Mods.Cnc/Traits/Render/WithReloadingSpriteTurret.cs
GraionDilach: No, I misinterpreted how this supposed to work.

OpenRA.Mods.Common/Traits/Turreted.cs
GraionDilach: Replaced this with a `.FirstOrDefault() == null` check.

GraionDilach:
Rebased.

OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
```diff
@@ -96,7 +96,7 @@ public void Attack(Actor targetActor, bool allowMove = true, bool forceAttack =
 			if (!targetActor.Info.HasTraitInfo<FrozenUnderFogInfo>() && !Self.Owner.CanTargetActor(targetActor))
 				Log.Write("lua", "{1} is not revealed for player {0}!", Self.Owner, targetActor);
 
-			attackBase.AttackTarget(target, true, allowMove, forceAttack);
+			attackBase.First(x => !x.IsTraitDisabled && x.HasAnyValidWeapons(target)).AttackTarget(target, true, allowMove, forceAttack);
```
pchote: Please use FirstOrDefault and then throw a LuaException if it's null.  We don't want to leak cryptic C# exceptions into the scripting environment

OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
```diff
@@ -93,7 +93,8 @@ public override Activity Tick(Actor self)
 				if (self.IsDisabled())
 					return this;
 
-				var weapon = attack.ChooseArmamentsForTarget(target, forceAttack).FirstOrDefault();
+				Armament weapon = null;
+				var attack = attackArray.FirstOrDefault(af => !af.IsTraitDisabled && (weapon = af.ChooseArmamentsForTarget(target, forceAttack).FirstOrDefault()) != null);
```
pchote: Assigning `weapon` from inside the LINQ selector is really really really ugly C# style.  Please expand this out into a more expressive block.

OpenRA.Mods.Common/Traits/Turreted.cs
pchote: My motivation here is to eventually allow multiple `Attack*` traits each controlling their own set of armaments (like the Generals overlord tank, or special case auto-crush logic at short range).  You shouldn't need to move any logic around, just add filtering at appropriate places.

OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
GraionDilach: That was for perf. Everything else I'd do would involve looking up something twice.

OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
GraionDilach: Done.

OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
pchote: You can pull this out to a method that enumerates using a `foreach` and returns either a `Pair` of trait and weapon, returns one and uses an out param for the other, or uses out params for both and returns a bool (false if there are no matches).

OpenRA.Mods.Common/Traits/Turreted.cs
GraionDilach: Currently we support N-to-N AttackTurreted-Turreted linkings. As long as that is kept, Turreted.Tick() needs to stay as-is and so does this line.

pchote: These two lines may stay the same, but there are others that need to change.  There is currently no way to define the relationship between AttackTurreted and turrets.  Each AttackTurreted will try to control all available turrets.

GraionDilach: Okay, now I see what you mean.

AttackTurreted should filter out to the turrets which are appropriate for it's respective armaments. I'd say right now that means AttackTurreted.Turrets would be set up in RulesetLoaded, because Requires would introduce a cyclic dependency and longterm to shift the turret handling from armaments to the attack traits as well (which will probably be a mess on it's own).

OpenRA.Mods.Common/Traits/Render/WithSpriteTurret.cs
```diff
@@ -124,7 +124,7 @@ public virtual void Tick(Actor self)
 			if (Info.AimSequence == null)
 				return;
 
-			var sequence = Attack.IsAttacking ? Info.AimSequence : Info.Sequence;
+			var sequence = Attack.Any(a => a.IsAttacking && !a.IsTraitDisabled) ? Info.Sequence : Info.AimSequence;
```
abcdefg30: https://github.com/OpenRA/OpenRA/pull/11600#discussion_r70274265
> Is the swapping of the sequences here intentional?

Here (not) as well?

OpenRA.Mods.Common/Traits/Turreted.cs
GraionDilach: I'm uncertain if I can get that worki9ng in this PR, because if I do it seemingly right, the unit cannot turn any turret which got enabled on the world with an upgrade but was disabled when launch, and I've put 2 days into it without finding the cause of that issue.

GraionDilach:
Rebased, updated. Fixed all issues, and `AttackTurreted` now only turns the turrets it's supposed to.

OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
```diff
@@ -39,27 +39,33 @@ class AttackPopupTurretedInfo : AttackTurretedInfo, Requires<BuildingInfo>, Requ
...
 		public AttackPopupTurreted(ActorInitializer init, AttackPopupTurretedInfo info)
 			: base(init.Self, info)
 		{
 			this.info = info;
-			turret = turrets.FirstOrDefault();
 			wsb = init.Self.Trait<WithSpriteBody>();
 			skippedMakeAnimation = init.Contains<SkipMakeAnimsInit>();
 		}
 
+		void INotifyCreated.Created(Actor self)
```
WolfGaming: Concerning the hack below in BuildingComplete, won't this run twice if the hack runs?
Unless Building.Created makes the Code skip INotifyCreated in which case its a bug.

OpenRA.Mods.Common/Activities/Hunt.cs
```diff
@@ -44,4 +45,4 @@ public override Activity Tick(Actor self)
 				this);
 		}
 	}
-}
\ No newline at end of file
+}
```
WolfGaming: extra Line ending, intentional?

OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
```diff
@@ -96,7 +96,11 @@ public void Attack(Actor targetActor, bool allowMove = true, bool forceAttack =
 			if (!targetActor.Info.HasTraitInfo<FrozenUnderFogInfo>() && !Self.Owner.CanTargetActor(targetActor))
 				Log.Write("lua", "{1} is not revealed for player {0}!", Self.Owner, targetActor);
 
-			attackBase.AttackTarget(target, true, allowMove, forceAttack);
+			var attack = attackBase.FirstOrDefault(x => !x.IsTraitDisabled && x.HasAnyValidWeapons(target));
+			if (attack == null)
+				throw new LuaException("Actor {0} was ordered to attack {1} but has no active Attack* trait to do so!.".F(Self, targetActor));
```
WolfGaming: What if it has an active attack trait but its not valid against the actor for example rifleman vs mig?

OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
```diff
@@ -93,7 +94,8 @@ public override Activity Tick(Actor self)
 				if (self.IsDisabled())
 					return this;
 
-				var weapon = attack.ChooseArmamentsForTarget(target, forceAttack).FirstOrDefault();
+				var attack = SelectActiveAttack(target, forceAttack).FirstOrDefault();
+				var weapon = attack.Value;
```
WolfGaming: You use FirstOrDefault above and then assume attack can never be null.

OpenRA.Mods.Common/Traits/Attack/AttackTurreted.cs
```diff
@@ -38,7 +42,7 @@ protected override bool CanAttack(Actor self, Target target)
 			// Don't break early from this loop - we want to bring all turrets to bear!
 			var turretReady = false;
 			foreach (var t in turrets)
-				if (t.FaceTarget(self, target))
+				if (t.FaceTarget(self, target, this) && !t.IsTraitDisabled)
```
WolfGaming: Do we still want to call t.FaceTarget even if t.IsTraitDisabled is true?

OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -102,19 +103,24 @@ public virtual void Tick(Actor self)
...
-					if (realignTick < info.RealignDelay)
-						realignTick++;
-					else if (info.RealignDelay > -1)
-						DesiredFacing = null;
-
-					MoveTurret();
+					attack = null;
 				}
 			}
 			else
```
WolfGaming: Can't this be else if (DesiredFacing != null) or do we need to call MoveTurret with the DesiredFacing as null?

OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -102,19 +103,24 @@ public virtual void Tick(Actor self)
...
+				if (!attack.IsAttacking || attack.IsTraitDisabled)
 				{
-					if (realignTick < info.RealignDelay)
-						realignTick++;
-					else if (info.RealignDelay > -1)
-						DesiredFacing = null;
-
-					MoveTurret();
+					attack = null;
 				}
```
WolfGaming: Convention was to have no curly brackets around if statements if they are a single line and don't contain an if statement inside and else outside.

OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -102,19 +103,24 @@ public virtual void Tick(Actor self)
...
+				{
+					if (info.RealignDelay > -1)
+					{
+						DesiredFacing = null;
+						realignTick = 0;
+					}
+				}
+
 				MoveTurret();
 			}
```
WolfGaming:
```
if (attack != null)
{
	if (!attack.IsAttacking || attack.IsTraitDisabled)
		attack = null;
}
else
{
	realignTick = 0;
	if (realignTick < info.RealignDelay)
		realignTick++;
	else
	{
		if (info.RealignDelay > -1)
		{
			DesiredFacing = null;
			realignTick = 0;
		}
	}
	MoveTurret();
}

```
Can be refactored to

```
if (attack == null)
{
	if (realignTick < info.RealignDelay)
		realignTick++;
	else if (info.RealignDelay > -1)
	{
		DesiredFacing = null;
		realignTick = 0;
	}
	MoveTurret();
	return;
}

if (!attack.IsAttacking || attack.IsTraitDisabled)
	attack = null;
```

OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
GraionDilach: Yes, it will run twice. Not gonna cause any harm though.

OpenRA.Mods.Common/Traits/Attack/AttackTurreted.cs
GraionDilach: Yes, see the comment three lines above why.

OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
WolfGaming: We can add a if (turret != null) return; then, no reason to do all those lookups twice.
Unless we want to try and fix the hack by introducing priorities to the INotifies or something.

OpenRA.Mods.Common/Activities/Hunt.cs
GraionDilach: Not intentional, but desired. https://github.com/OpenRA/OpenRA/pull/10759#discussion_r52845580

OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
GraionDilach: Should this silently fail in that case? Trying to attack in such a scenario also sounds bogus.

OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
GraionDilach: Another way to fix the hack would be to trigger NotifyComplete in the frame-end task.

OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
WolfGaming: I don't know, it probably could be done in some for loop that attacks units produced by Lua code or something in which case we want it to silently fail or have some way to ignore the error.

It depends on what we want the method to do so it probably needs some discussion on IRC or you get pchote to chip in.

OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
WolfGaming: That's probably what we want to do then unless Building.Created somehow relies on NotifyComplete to have ran.

Which would be bogus imo.

OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
pchote: I haven't looked at this pr properly, but at least this case seems to be written to only care about the "switch from one Attack* type to another" case.  We want to be able to support multiple active Attack*'s, so this should enumerate over all valid ones instead of picking the first, and not throw an exception if there are none (you could return a bool instead).

OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
GraionDilach: `if (attack != null)` throws me a syntax error. Should I check for the key instead of the value? (Because there is a nullcheck in the next line.)

WolfGaming: Turns out KeyValuePair can never be null so it will return a default value, if the default values for the key and value is null then you can ignore this comment.

Though you could check if IEnumerable.Any() returns True before doing First but I guess thats unnecessary.

GraionDilach: Yes, it returns a null, null pair.

OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
GraionDilach: Resolved.

OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
GraionDilach: Resolved. Now it sets all suitable `Attack*` traits to attack - á lá `AutoTarget` - and writes a warning to `lua.log` if none found.

GraionDilach:
Updated.

WolfGaming:
Sorry my refactoring had a style error, missing new line between
} and MoveTurret();

GraionDilach:
Fixed.

GraionDilach:
Could someone remove the Rebase me label?

OpenRA.Mods.Common/Activities/Hunt.cs
```diff
@@ -23,10 +23,11 @@ public class Hunt : Activity
 
 		public Hunt(Actor self)
 		{
-			var attack = self.Trait<AttackBase>();
+			var attack = self.TraitsImplementing<AttackBase>();
 			targets = self.World.ActorsHavingTrait<Huntable>().Where(
 				a => self != a && !a.IsDead && a.IsInWorld && a.AppearsHostileTo(self)
```
WolfGaming: a.AppearsHostileTo(self) will this ever be true if self != a is true?

OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
```diff
@@ -96,7 +96,13 @@ public void Attack(Actor targetActor, bool allowMove = true, bool forceAttack =
...
 				Log.Write("lua", "{1} is not revealed for player {0}!", Self.Owner, targetActor);
 
-			attackBase.AttackTarget(target, true, allowMove, forceAttack);
+			var attack = attackBase.Where(x => !x.IsTraitDisabled && x.HasAnyValidWeapons(target));
+			if (!attack.Any())
+				Log.Write("lua", "Actor {0} was ordered to attack {1} but has no valid Attack* trait to do so!.".F(Self, targetActor));
+			foreach (var a in attack)
+			{
+				a.AttackTarget(target, true, allowMove, forceAttack);
+			}
```
WolfGaming: Convention not to have the Curly brackets here.

OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
```diff
@@ -121,10 +123,26 @@ public override Activity Tick(Actor self)
...
+				foreach (var a in attackArray)
+				{
+					if (!a.IsTraitDisabled)
+					{
+						foreach (var w in a.ChooseArmamentsForTarget(t, forceAttack))
+						{
+							yield return new KeyValuePair<AttackFollow, Armament>(a, w);
+						}
+					}
+				}
```
WolfGaming: Pretty sure all of those curly brackets aren't supposed to be here, correct me if I'm wrong, bit fuzzy on inner loops with an if in the convention.

WolfGaming:
Otherwise :+1: still haven't tested this, will do that later today.

OpenRA.Mods.Common/Activities/Hunt.cs
GraionDilach: Well, it has to, for units owned by `self`s enemies.

WolfGaming: No I mean why do we need to check self != a if we later check if self is hostile to a?
Can an actor ever be hostile to himself?

GraionDilach: TS berserk logic can assume such.

WolfGaming: Well fine, it will have to be there then :)

WolfGaming:
Tested and works as expected :+1: from me when those nits are fixed.

OpenRA.Mods.Common/Scripting/Properties/CombatProperties.cs
GraionDilach: Would you prove it, please? I'm only aware of that rule with if blocks, not foreach ones.

WolfGaming: UpgradeManager.cs line 216
Game.cs line 217 and 320

GraionDilach:
Updated.

OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
```diff
@@ -39,27 +39,33 @@ class AttackPopupTurretedInfo : AttackTurretedInfo, Requires<BuildingInfo>, Requ
...
+		{
+			var armamentturrets = getArmaments().Select(x => x.Info.Turret).ToHashSet();
+
+			turret = self.TraitsImplementing<Turreted>().Where(x => armamentturrets.Contains(x.Name)).First();
```
pchote: As a general rule if you're requiring a single value you should prefer `.Single` over `.First`.

OpenRA.Mods.Common/Traits/Turreted.cs
```diff
@@ -125,11 +125,13 @@ void MoveTurret()
 			TurretFacing = Util.TickFacing(TurretFacing, df, info.TurnSpeed);
 		}
 
-		public bool FaceTarget(Actor self, Target target)
+		public bool FaceTarget(Actor self, Target target, AttackTurreted at)
 		{
 			if (self.IsDisabled())
 				return false;
 
+			attack = at;
```
pchote: This sets off warning bells in my head.  What's going on here?

GraionDilach: This caches the current active AttackTurreted trait for FaceTarget (instead of constantly looking up all available AttackTurreted traits which has an armament controlling this turret). I'm assuming that only one AttackTurreted trait will be active at a turret at a time.

OpenRA.Mods.Cnc/Traits/AttackPopupTurreted.cs
abcdefg30: Won't you also be able to use `turrets` (from [`AttackTurreted#L25](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/Attack/AttackTurreted.cs#L25)) instead doing the trait lookup again?

abcdefg30: You can drop the `Where` as well then and use `.Single(x => ...)`.

GraionDilach: Guess I can. I remember I had issues here due to something I messed up elsewhere at first. Don't ask me what those were.

abcdefg30: I just noticed that you're touching `turrets` in this PR anyway. So the whole lookup here is redundant? So you can just use `turrets.Single()`?

OpenRA.Mods.Common/Traits/Attack/AttackFollow.cs
```diff
@@ -121,10 +123,20 @@ public override Activity Tick(Actor self)
 				}
 
 				if (!onRailsHack)
-					attack.Target = Target.Invalid;
+					attack.Key.Target = Target.Invalid;
 
 				return NextActivity;
 			}
+
+			IEnumerable<KeyValuePair<AttackFollow, Armament>> SelectActiveAttack(Target t, bool forceAttack)
```
abcdefg30: Do you plan on using this method in any other places than the current one (and still requiring a whole enumerable)?
Otherwise you could just return the first fitting pair and bail out early.

GraionDilach: I don't plan per se, but @pchote's longterm plan explained in https://github.com/OpenRA/OpenRA/pull/11600#discussion_r70411452 would involve the attack activity handling all the pairs returned.

However it's way out of scope for this PR IMO.

GraionDilach:
Updated. Had to rip out the turret initializer code to a helper function otherwise I have no way to call it in `AttackPopupTurreted`.

OpenRA.Mods.Common/Traits/Buildings/Building.cs
```diff
@@ -186,7 +186,7 @@ public IEnumerable<WPos> TargetablePositions(Actor self)
 		public void Created(Actor self)
 		{
 			if (SkipMakeAnimation || !self.Info.HasTraitInfo<WithMakeAnimationInfo>())
-				NotifyBuildingComplete(self);
+				self.World.AddFrameEndTask(w => NotifyBuildingComplete(self));
```
abcdefg30: This looks rather dangerous (i.e. could easily introduce regressions).
Wouldn't it be easier to add a simple `turret != null` check in `BuildComplete` (re https://github.com/OpenRA/OpenRA/pull/11600#discussion_r74230299)?
(And possibly set `DesiredFacing` to `null` in `Created`?)

GraionDilach: I've tested this in TD and RA. Your danger does not exist.

GraionDilach:
seems this ain't getting anywhere and I'm not sure if I'll have time for this from next week onwards so closing.

PR #11601, opened 2016-07-08
Replace reworked singles and doubles maps with originals

ghost:
Replacing the reworked singles and doubles maps with their original counterparts. These were the only two 'heavily' reworked maps that made it into a release and they don't seem to sit well with some of the more nostalgic players.


pchote:
From what I saw on the forum only one person doesn't like the new maps (materia).  The poll at http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19681 suffers small number statistics, but currently shows more like the new than the old.

Mailaender:
No consensus so closing. The old maps are saved as revisions on http://resource.openra.net/ so this shouldn't be a big problem.

PR #11621, opened 2016-07-09
Added With*ControlGroup

Mailaender:
This splits the control group rendering from `SelectionDecorations` which gives a choice to render this with fonts (Tiberian Sun and above) or sprites or not at all. A followup of https://github.com/OpenRA/OpenRA/pull/11565 and https://github.com/OpenRA/OpenRA/pull/11567.


abcdefg30:
Needs a rebase.

abcdefg30:
The control group isn't rendered anymore when you use the `visibilty` command.

abcdefg30:

Mailaender:
I replaced

```csharp
			if (!self.Owner.IsAlliedWith(wr.World.RenderPlayer))
				yield break;
```

with

```csharp
			if (self.Owner != wr.World.RenderPlayer)
				yield break;
```

intentionally as I think control groups are something that just concern the player and not their allies.

abcdefg30:
Isn't `RenderPlayer` `null` when the visibility is disabled? So you could add a `null` check.
That way spectators could also see control groups (why shouldn't they).

pchote:
Control groups are client-side only.  Their state is not shared over the network, so there is no way to tell anything about units that aren't your own.

abcdefg30:
Ah, well then scratch that with spectators.

reaperrr:
Needs rebase.

Mailaender:
Rebased.

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroup.cs
```diff
@@ -0,0 +1,86 @@
...
+			if (group == null)
+				yield break;
+
+			pipImages.PlayFetchIndex(Info.GroupSequence, () => (int)group);
+
+			var pos = basePosition - (0.5f * pipImages.Image.Size.XY).ToInt2() + Info.ScreenOffset;
+			yield return new UISpriteRenderable(pipImages.Image, self.CenterPosition, pos, 0, palette, 1f);
+		}
+
+		void ITick.Tick(Actor self)
```
obrakmann: `Tick`, or `RenderTick`?

obrakmann:
Looks fine overall.

The positioning in TS looks weird and needs more work, though:
screenshot from 2016-07-27 21 12 57

GraionDilach:
IIRC bottom-left was the reference point in TS for this and top-left for RA2.

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroup.cs
Mailaender: Not sure, but I thought `Tick` would be better for performance.

Mailaender:
Updated the offsets with better defaults.

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroup.cs
pchote: `TickRender` will tick when the game is paused or lagging. `Tick` won't.

pchote: Why do you have this at all?

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroup.cs
```diff
@@ -0,0 +1,86 @@
...
+	{
+		[PaletteReference] public readonly string Palette = "chrome";
+
+		public readonly string Image = "pips";
+
+		[Desc("Sprite sequence used to render the control group 0-9 numbers.")]
+		[SequenceReference("Image")] public readonly string GroupSequence = "groups";
+
+		[Desc("Manual offset in screen pixel.")]
+		public readonly int2 ScreenOffset = new int2(9, 5);
```
pchote: Requiring every actor to set this manually is a bit horrible.

Please implement `ReferencePoint` like we have on `With*Decoration`.

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroup.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Drawing;
+using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits.Render
+{
+	[Desc("Renders Ctrl groups using pixel art.")]
+	public class WithSpriteControlGroupInfo : ITraitInfo
```
pchote: This and the text trait are both decorations, so should follow the naming, property, and style conventions of `WithDecoration` and `WithTextDecoration`.

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroup.cs
Mailaender: This is in practice unified for all actors. It depends more on the sprite size of the font and the thickness of the selection decorations. Actually no change here. This is exactly how the legacy traits implemented it. I just moved the hard-coded value to the trait info.

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroup.cs
Mailaender: I guess I saw it in another trait and thought it was required or best practices.

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroup.cs
Mailaender: Renamed it.

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroup.cs
Phrohdoh: Just because something was implemented poorly previously doesn't mean it shouldn't be proper now.

Mailaender: I don't disagree on that. Just it isn't my code and it isn't the intention of this pull request.

pchote: You should be able to directly copy what WithSpriteDecoration does.

Mailaender: Replaced the manual offsets.

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroupDecoration.cs
```diff
@@ -0,0 +1,109 @@
...
+
+			pipImages = new Animation(self.World, Info.Image);
+		}
+
+		IEnumerable<IRenderable> IPostRenderSelection.RenderAfterWorld(WorldRenderer wr)
+		{
+			if (self.World.FogObscures(self))
+				yield break;
+
+			if (self.Owner != wr.World.RenderPlayer)
```
pchote: This should be `wr.World.LocalPlayer`.  The underlying logic uses `LocalPlayer` to filter which actors can be assigned to groups.

OpenRA.Mods.Common/Traits/Render/WithSpriteControlGroupDecoration.cs
```diff
@@ -0,0 +1,109 @@
...
+			var b = self.VisualBounds;
+			var pos = wr.ScreenPxPosition(self.CenterPosition);
+			var tl = wr.Viewport.WorldToViewPx(pos + new int2(b.Left, b.Top));
+			var pal = wr.Palette(Info.Palette);
+
+			foreach (var r in DrawControlGroup(wr, self, tl, pal))
+				yield return r;
+		}
+
+		IEnumerable<IRenderable> DrawControlGroup(WorldRenderer wr, Actor self, int2 basePosition, PaletteReference palette)
```
pchote: `basePosition` is now unused, so this and the calculations above should be removed.

pchote: Now that we have a single-purpose trait you may as well inline this in `RenderAfterWorld`.

Mailaender:
Updated.

GraionDilach:
Needs a rebase.

Mailaender:
Rebased.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -270,6 +270,16 @@ internal static void UpgradeActorRules(ModData modData, int engineVersion, ref L
 					}
 				}
 
+				if (engineVersion < 20160717)
```
GraionDilach: Should have bumped this during the rebase, sorry for not pointing it out yesterday. :S

GraionDilach:
:+1: otherwise.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
Mailaender: It is still a few days above the `20160713` entry. Let's please not time travel even when reviews take forever.

reaperrr:
One last nit:
In my opinion, `ReferencePoint: Top` looks better on RA and TD infantry, because it effectively moves the sprite into the top-left corner of the visual bounds (I guess this is due to the wide canvas of `pips.shp`).
:+1: once the RA and TD infantry defaults have been changed accordingly.

Mailaender:
Updated.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d8c61eea9eb5ebe4e8df894962d3cd5091521962...d024b56d0212b6bc21840d7986a981834ad503ec)

PR #11622, opened 2016-07-09
Fixed parent package from ZipFile not getting disposed.

Mailaender:
This tries to fix another defect from #11482, which was detected by Coverity.

### DIFF
--- a/OpenRA.Game/FileSystem/ZipFile.cs
+++ b/OpenRA.Game/FileSystem/ZipFile.cs
@@ -57,7 +57,8 @@ namespace OpenRA.FileSystem
 			// SharpZipLib breaks when asked to update archives loaded from outside streams or files
 			// We can work around this by creating a clean in-memory-only file, cutting all outside references
 			pkgStream = new MemoryStream();
-			p.GetStream(name).CopyTo(pkgStream);
+			using (var sourceStream = p.GetStream(name))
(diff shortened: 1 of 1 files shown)


pchote:
Please don't blindly trust everything Coverity says.  `p` is a reference to an object owned by the filesystem, and this code has no business disposing it.  This will break things.

abcdefg30:
So, what happens with this PR now? Wontfix?

RoosterDragon:
Coverity is trying to indicate the stream returned by `GetStream(name)` is not disposed - which is correct.

The fix we need is actually:
```c#
pkgStream = new MemoryStream();
using (var sourceStream = p.GetStream(name))
	sourceStream.CopyTo(pkgStream);
pkgStream.Position = 0;
```

Mailaender:
Thanks. Updated accordingly. I tend to misunderstand the Coverity Scan output quite often.

RoosterDragon:
:+1:

GraionDilach:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5b25a08507868fbf9b07f7ad3dfb13dff469a204...5e0f632a37b07611ffda829739942c1507393c58)

PR #9770, opened 2015-10-25
flip sprites in combined sequences

evgeniysergeev:
allow to use frames: in Combined: sequences plus allow flip or mirror sprites in sequences.
needed for dune2 mod

Issue #7673, opened 2015-03-16
Add support for voxel projectiles (TS/RA2)


reaperrr:
- missiles
- bullets
- gravity bombs

PR #11624, opened 2016-07-09
Implement the TS Cluster Missile.

### LINKED ISSUE #7773: Implement TS cluster missile
It is currently implemented as a placeholder.

GraionDilach:
Resolves #7773.


OpenRA.Mods.Common/Warheads/FireClusterWarhead.cs
```diff
@@ -0,0 +1,96 @@
...
+	{
+		[WeaponReference, FieldLoader.Require]
+		[Desc("Has to be defined in weapons.yaml as well.")]
+		public readonly string Weapon = null;
+
+		[Desc("Ignore the weapon's target settings and always fire this weapon?",
+			"Enabling this allows legacy WW behaviour.")]
+		public readonly bool ForceFire = false;
+
+		[Desc("The range of the cells where the weapon should be fired."
```
Mailaender: Use a `,` or add a manual space ` ` when keeping the `+`.

GraionDilach: Ah yeah, right.

Mailaender:

OpenRA.Mods.Common/Traits/SupportPowers/NukePower.cs
```diff
@@ -108,7 +111,7 @@ public override void Activate(Actor self, Order order, SupportPowerManager manag
 				wsb.PlayCustomAnimation(self, info.ActivationSequence, () => wsb.CancelCustomAnimation(self));
 			}
 
-			var targetPosition = self.World.Map.CenterOfCell(order.TargetLocation);
+			var targetPosition = self.World.Map.CenterOfCell(order.TargetLocation) + new WVec(WDist.Zero, WDist.Zero, info.AirburstAltitude);
```
abcdefg30: `new WVec(0, 0, info.AirburstAltitude)` should be fine as well.

OpenRA.Mods.Common/Warheads/FireClusterWarhead.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+	public class FireClusterWarhead : Warhead, IRulesetLoaded<WeaponInfo>
+	{
+		[WeaponReference, FieldLoader.Require]
+		[Desc("Has to be defined in weapons.yaml as well.")]
+		public readonly string Weapon = null;
+
+		[Desc("Ignore the weapon's target settings and always fire this weapon?",
+			"Enabling this allows legacy WW behaviour.")]
+		public readonly bool ForceFire = false;
```
abcdefg30: This can go away as discussed on IRC.

OpenRA.Mods.Common/Traits/SupportPowers/NukePower.cs
GraionDilach: AirburstAltitude is a `WDist` and this was copied from the missile projectile from consistency.

GraionDilach:
Forgot to announce I'"ve updated this half a day ago.

OpenRA.Mods.Common/Traits/SupportPowers/NukePower.cs
GraionDilach: WVec only has a WDist WDist WDist and an int int int constructor which matters in this case. Falling back to the int one would mean the WDist valuefrom the yaml would need to be converted and this is more readable IMO.

reaperrr:
@GraionDilach I currently don't have access to my dev system so I can't check what the file name was, but there's an shp in TS' `conquer.mix` showing the cluster missile pointing diagonally downwards which might have a suitable frame (might need vertical flipping, not sure). If it looks good enough, it would be a better temporary solution than using the RA1 nuke, in my opinion.

pchote:
Please set this up using `AttackOrderPower` and a missile projectile instead of abusing the nuke.  You'll still need to use temporary artwork (one of the normal missiles) until we add voxel support for missiles, but this at least gets us 90% of the way to a proper implementation.

GraionDilach:
There is no suitable projectile for this in the mixes. What @reaperrr remembers is the crate animation.

Also `AttackOrderPower` doesn't support multiple weapons, so right now you can't have chem missiles and multi missiles at the same time. The normal missiles are tiny at best.

pchote:
We don't support anything related to chem missiles right now, and `AttackOrderPower` will in any case need to be fixed if that is true.  That should not block setting this up in a future proof way now.

GraionDilach:
Another point of merging this as-is is because the D2k Death Hand missile will also need airburst support.

pchote:
It would probably be better to set that up here instead of further hacking the TS powers in the short term.

GraionDilach:
That uses a different warhead (one which throws debris/shrapnel around instead of distributing projectiles equally to an area), one I don't have the code of atm.

GraionDilach:
In any case, the code commits are already futureproof (the first and the third is required by an accurate TS Cluster Missile, the second by an accurate Death Hand), and only the current yaml case isn't.

OpenRA.Game/Map/Map.cs
```diff
@@ -1121,6 +1121,26 @@ public IEnumerable<CPos> FindTilesInCircle(CPos center, int maxRange, bool allow
...
+			var x = center.X - (width - 1) / 2;
+			var y = center.Y - (height - 1) / 2;
+
+			Func<CPos, bool> valid = Contains;
+			if (allowOutsideBounds)
+				valid = Tiles.Contains;
+
+			for (var i = 0; i < width; i++)
+			{
+				for (int j = 0; j < height; j++)
```
pchote: `var`

GraionDilach: Fixed.

Mailaender:
> There is no suitable projectile for this in the mixes. What @reaperrr remembers is the crate animation.

As it is a placeholder anyway until we have https://github.com/OpenRA/OpenRA/issues/7673 you can use any artwork. We have support for `FlipX` and `FlipY` since https://github.com/OpenRA/OpenRA/pull/9770. As we don't own the assets, SHP based workarounds are something to avoid.

GraionDilach:
Updated artwork with a converted-to-SHP version of the TS cluster missile voxel. Had to add an additional commit to make it remapable.

Mailaender:
:+1:

OpenRA.Mods.Common/Projectiles/Missile.cs
```diff
@@ -876,8 +884,7 @@ public IEnumerable<IRenderable> Render(WorldRenderer wr)
 						yield return r;
 				}
 
-				var palette = wr.Palette(info.Palette);
-				foreach (var r in anim.Render(pos, palette))
+				foreach (var r in anim.Render(pos, wr.Palette(palette)))
 					yield return r;
```
pchote: Looking up the palette each render is wasteful because we know it won't change.  Please cache this on first render.

OpenRA.Mods.Common/Warheads/FireClusterWarhead.cs
```diff
@@ -0,0 +1,92 @@
...
+
+		[Desc("The range of the cells where the weapon should be fired.",
+			"A single value means radius, two values mean rectangle.")]
+		public readonly int[] Range = new int[] { 1 };
+
+		WeaponInfo weapon;
+
+		public void RulesetLoaded(Ruleset rules, WeaponInfo info)
+		{
+			weapon = rules.Weapons[Weapon.ToLowerInvariant()];
```
pchote: Please use `TryGetValue` and throw a friendlier exception if the weapon doesn't exist (cf. https://github.com/OpenRA/OpenRA/commit/4f798ede3c34cc249e6ab26aafa452aebfdec6f1).

pchote:
The cluster behaviour doesn't match, and IMO looks a lot worse than the original game.  Looks like you need to increase the spread range and randomize the cells that it picks within that.

Video for reference: https://www.dropbox.com/s/9zb2y8n55f6oqys/tibsun.mov?dl=0

GraionDilach:
I was quite certain it wasn't using the splits-logic there but apparently it already does. Sounds like while this logic itself also existed in a WW game, it was only in RA2 as a bugged out variant of the above.

Guess I need to drop commit https://github.com/OpenRA/OpenRA/pull/11624/commits/643eaf9d6088f27d8f61a80766de5bd72391dd02 altogether. However if I'd drop that one, then commit https://github.com/OpenRA/OpenRA/pull/11624/commits/c5e5bece2aa544c598ab5bf3a70160abea4e0560 also loses it's point here and need to be sent in in another PR to update the SWs RA2 carried over from RA.

Great.

GraionDilach:
I have all the code now to fulfill the change requested in https://github.com/GraionDilach/OpenRA.Mods.AS/commit/7a0b70593fb4f9c200fbdbfc968ee40b08b6457d, however...

In order to resolve more issues than cluster/death hand missiles - for example, Firestorm Cyborg Reaper or RA2 Prism Tank -, I had to add `TargetType` support via https://github.com/GraionDilach/OpenRA.Mods.AS/blob/master/Traits/Warheads/WarheadAS.cs (with even including a hardcoded `TargetType` for air without an actor), based on the `CreateEffectWarhead code @reaperrr didn't updated in #11638. I don't know what are his plans with it, so I'm not certain if I should knock into that - considering that this PR as-is already required 2(+1) polish commits as well.

I don't know what to do with this PR tbh.

pchote:
IMO you're making life needlessly difficult by trying to add more and more behaviour to our existing projectile types.  There's no reason why we can't have a `ClusterMissile` projectile type that is designed to fly above a target and spawn new projectiles in the area around it.

GraionDilach:
I don't add anything to projectiles (besides airburst to nukes for Death Hand support). Everything's strictly warhead code. I'm not interested to maintain a `ClusterBullet`, `ClusterTesla`, `ClusterLaser` or whatever a modder would like to have - and as a modder, I definitely will have some of these.

The TS cluster missile can and should use a missile projectile, but spawning the clusters should be a warhead function due to scalability and generalization.

pchote:
Well... you've already outlined the complications with that approach, and I've offered a simple solution for progressing forward.

pchote:
Another complication to consider is premature detonation caused by running out of fuel or hitting a (firestorm) wall.  You won't want to fire the cluster behaviour in that case.  Using different warheads / target types makes that easy.

GraionDilach:
Again, I have implemented `TargetTypes` support already, it's just that I base it upon a streamlined `ImpactTypes` code and I don't know how @reaperrr wants to refactor `ImpactTypes`.

pchote:
`ImpactTypes` won't help with the fuel / firestorm case, which is why I mentioned them.

GraionDilach:
Firestorm needs to provide it's own `TargetType` - required to play the hit effects on FS wall - and there's `ExplodeWhenEmpty` on missile for running out of fuel.

reaperrr:
@GraionDilach 
My plan is/was to remove all ImpactTypes except the `Target*` ones and `None`, and then make use of `ValidTargets`/`InvalidTargets`.
What I got stuck on is that I couldn't figure out how to avoid regressions in certain edge cases (submerged subs in RA, hover/amphibious units on water in TS).

GraionDilach:
@reaperrr Well, I have a solution in the form of https://github.com/GraionDilach/OpenRA.Mods.AS/blob/master/Traits/Warheads/WarheadAS.cs but it has a hardcoded assumption of the Air TargetType which might be unfeasible upstream.

The real problem is that you cannot remove `ImpactTypes` properly without resolving the empty air case - I don't see the regressions in the cases you listed, those sounds like something resolvable with an additional targettype on them and listed as valid for ground and invalid for water hits.

GraionDilach:
In any case, I am not touching this PR until the `ImpactTypes`-issue has been resolved.

The AS codebase already has all the issues resolved here (although not in `FireClusterWarhead`, but `FireShrapnelWarhead`), with additional functionality which @ABrandau and I depend on. I'm more okay with closing this PR as-is than dumbing down my code (with axing the `ImpactTypes`-based `TargetTypes` support) to get it merged upstream then renaming my own trait to resolve the name conflicts and fix the issues in the mods coming from it.

GraionDilach:
Seems this ain't getting anywhere and the dependency it needs has an uncertain future, so closing.

Issue #11452, opened 2016-06-13
Teamplay - statistic


Micr0Bit:
i thought we could track teamplay-actions (or support-actions) and reward players for that at the end of a game. ideas for that : 

* placing defense-stuctures (pillbox, coils, turrets, flame-towers) inside a team-members buildradius (not overlapped by your own buildradius)
* killing enemy-units from within a team-members buildradius (not overlapped by your buildradius)
* repair bridges
* sending supply-trucks to a team-member
* unloading harvester on team-members refinery
* reloading team-members units
* transporting team-members unit x-amount of cells
* repairing team-members building with an engineer or repair-button

Faction-Specific
* heal/repair team-members units with medic/mechanic
* use spyplane/paratroops

HenrytheSlav:
Big warning - easily abusable with force fire. We hardly have any point score system in place. I think we should first establish some.

Micr0Bit:
you can also get a lot of kills on purpose ... if you play in a certain way ... so you can abuse that too ... or getting deaths , apm ...

Big Warning , thats not the point here ... 

im just asking to reward teamplay-actions with points ... if the players abuse that for whatever reason it doesnt matter ...

HenrytheSlav:
Fair point, but it's true that we lack any sensible scoring system. So we need to establish that first (possibly including these sugestions ofc).

pchote:
#10358 has been sitting there for a long time as a first step towards this.

abcdefg30:
From what I heard @GraionDilach wanted to take that over. Not sure if he still has time and will though.

PR #11627, opened 2016-07-10
Enable tracking of player experience in all mods

obrakmann:
This PR adds granting player experience via the `SupplyTruck`, `Captures`, `ExternalCaptures`, `Infiltrates`, `RepairableBuilding` and `RepairsUnits` traits.

Except for capturing and infiltrating, experience is only granted when those operations are carried out on allied actors. Experience through capturing is only granted when capturing enemy structures, not neutral ones.

This will cross a few things off of the list in #11452.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Repair.cs
+++ b/OpenRA.Mods.Common/Activities/Repair.cs
@@ -49,6 +49,13 @@ namespace OpenRA.Mods.Common.Activities
 
 			if (health.DamageState == DamageState.Undamaged)
 			{
+				if (host.Owner != self.Owner)
+				{
+					var exp = host.Owner.PlayerActor.TraitOrDefault<PlayerExperience>();
+					if (exp != null)
+						exp.GiveExperience(repairsUnits.PlayerExperience);
+				}
+
 				Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Speech", repairsUnits.FinishRepairingNotification, self.Owner.Faction.InternalName);
 				return NextActivity;
 			}
--- a/OpenRA.Mods.Common/Traits/Buildings/RepairableBuilding.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/RepairableBuilding.cs
@@ -35,6 +35,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Suffixed by the internal repairing player name.")]
 		public readonly string IndicatorPalettePrefix = "player";
 
+		[Desc("Experience gained by a player for repairing structures of allied players.")]
+		public readonly int PlayerExperience = 0;
+
 		public override object Create(ActorInitializer init) { return new RepairableBuilding(init.Self, this); }
 	}
 
@@ -133,6 +136,16 @@ namespace OpenRA.Mods.Common.Traits
 
 				if (health.DamageState == DamageState.Undamaged)
 				{
+					Repairers.Do(r =>
+					{
+						if (r == self.Owner)
+							return;
+
+						var exp = r.PlayerActor.TraitOrDefault<PlayerExperience>();
+						if (exp != null)
+							exp.GiveExperience(Info.PlayerExperience);
+					});
+
 					Repairers.Clear();
 					RepairActive = false;
 					return;
--- a/OpenRA.Mods.Common/Traits/Captures.cs
+++ b/OpenRA.Mods.Common/Traits/Captures.cs
@@ -22,10 +22,18 @@ namespace OpenRA.Mods.Common.Traits
 	{
 		[Desc("Types of actors that it can capture, as long as the type also exists in the Capturable Type: trait.")]
 		public readonly HashSet<string> CaptureTypes = new HashSet<string> { "building" };
+
 		[Desc("Unit will do damage to the actor instead of capturing it. Unit is destroyed when sabotaging.")]
 		public readonly bool Sabotage = true;
+
 		[Desc("Only used if Sabotage=true. Sabotage damage expressed as a percentage of enemy health removed.")]
 		public readonly int SabotageHPRemoval = 50;
+
+		[Desc("Experience granted to the capturing player.")]
+		public readonly int PlayerExperience = 0;
+
+		[Desc("Stance that the structure's previous owner needs to have for the capturing player to receive Experience.")]
+		public readonly Stance PlayerExperienceStances = Stance.Enemy;
 
 		[VoiceReference] public readonly string Voice = "Action";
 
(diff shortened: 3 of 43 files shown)


OpenRA.Mods.Common/Activities/Repair.cs
```diff
@@ -49,6 +49,13 @@ public override Activity Tick(Actor self)
 
 			if (health.DamageState == DamageState.Undamaged)
 			{
+				if (host.Owner != self.Owner)
```
pchote: Why not use stances here too?

OpenRA.Mods.Common/Traits/Buildings/RepairableBuilding.cs
```diff
@@ -133,6 +136,16 @@ public void Tick(Actor self)
 
 				if (health.DamageState == DamageState.Undamaged)
 				{
+					Repairers.Do(r =>
+					{
+						if (r == self.Owner)
```
pchote: and here.

OpenRA.Mods.Common/Traits/Captures.cs
```diff
@@ -22,11 +22,19 @@ public class CapturesInfo : ITraitInfo
...
 		public readonly bool Sabotage = true;
+
 		[Desc("Only used if Sabotage=true. Sabotage damage expressed as a percentage of enemy health removed.")]
 		public readonly int SabotageHPRemoval = 50;
 
+		[Desc("Experience granted to the capturing player.")]
+		public readonly int PlayerExperience = 0;
+
+		[Desc("Stance that the structure's previous owner needs to have for the capturing player to receive Experience.")]
+		public readonly Stance ValidStances = Stance.Enemy;
```
pchote: These should be `PlayerExperienceStances` to avoid confusion (the logical assumption is that this would be the stances that can be captured)

OpenRA.Mods.Common/Traits/ExternalCaptures.cs
```diff
@@ -26,6 +26,12 @@ class ExternalCapturesInfo : ITraitInfo
 		[Desc("Destroy the unit after capturing.")]
 		public readonly bool ConsumeActor = false;
 
+		[Desc("Experience granted to the capturing player.")]
+		public readonly int PlayerExperience = 0;
+
+		[Desc("Stance that the structure's previous owner needs to have for the capturing player to receive Experience.")]
+		public readonly Stance ValidStances = Stance.Enemy;
```
pchote: Same here.

OpenRA.Mods.RA/Activities/Infiltrate.cs
```diff
@@ -46,6 +48,10 @@ protected override void OnInside(Actor self)
 			foreach (var t in target.TraitsImplementing<INotifyInfiltrated>())
 				t.Infiltrated(target, self);
 
+			var exp = self.Owner.PlayerActor.TraitOrDefault<PlayerExperience>();
+			if (exp != null && target.Owner != self.Owner)
```
pchote: IMO this should either be unconditional (i.e. based on the valid stances for infiltration), or have its own set of stances.

OpenRA.Mods.Common/Activities/Repair.cs
obrakmann: Because afaik, `self.Owner.Stances[self.Owner] == Stance.Ally`, which means we can't filter on ownership.

GraionDilach: You rejected #11093 for this very reason afterall.

pchote:
@Micr0Bit suggested in IRC that we remove the separate kill and death columns in the menu UI.  I think that's a good idea, since we then IIRC match the original game's stats (the "world" icon under the radar) and don't leaking as much world data.

GraionDilach:
Will I think removing them is a really bad idea. The K/D values and ratio are quite useful as basic balance debugger tools as well.

pchote:
They would remain available in the stats UI together with the other fields that would be equally useful for balance.

GraionDilach:
Alrighto then.

OpenRA.Mods.Common/Traits/Captures.cs
GraionDilach: TBH I'd say it would be nice if Captures/ExternalCaptures would even grant experience to the capturing actor as well.

pchote:
We should definitely aim to include this in the next release, so that milestone was appropriate IMO.

obrakmann:
Like this?
screenshot from 2016-07-11 15 32 37

pchote:
Yeah, although I think it would look a bit better if we renamed Experience &rarr; Score and made the column narrower.

It would be really awesome if the names were also sorted by score, we can probably cheat by evaluating it once when opening the menu and assuming it won't change (actually, do we even update the exp and old kills/deaths dynamically?)

obrakmann:
I don't think that's a good idea. It could lead to the wrong impression that that 'score' would influence the outcome of the game.

Micr0Bit:
its not a big deal ... but i would enjoy having names sorted by the score ... i mean at the end of every game there will be a "you lost" ... or "you are victorious" ... no matter what your score or killcount is ...

obrakmann:
Alright, how does this look? (I'll remove the team column, it's redundant now. Any idea how to use the free space?)

screenshot from 2016-07-12 22 32 40

pchote:
The team headers look a bit odd.  Could you try them with TinyBold (probably looks bad) or like we do them in the observer stats window?

obrakmann:
TinyBold does look bad. Here's a variant that does it like the stats windows:

screenshot from 2016-07-12 23 36 30

obrakmann:
Another edit, with smaller bars, and bold font. I think I like this one best so far.

screenshot from 2016-07-13 00 55 52

obrakmann:
Updated.

Mailaender:
This works and looks really well. :white_check_mark:

pchote:
Could we use "Combat Points" instead of "Rating"?

Phrohdoh:
Not if repairing your own units is going to count towards this number.

pchote:
Why... repairing your units is part of combat?

Micr0Bit:
i also dont like "rating" ... i would be fine just with "Score"

obrakmann:
Updated/rebased.

Micr0Bit:
:+1: tested on red alert

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/9d5ff92ed70af549bbb28333de7585bd89f9e36e...16daaa55b4d2203aec5231645f5597c97eb9d023)

PR #11631, opened 2016-07-10
Add curl as dependency for linux

tmsbrg:
`make dependencies` fails if you don't have curl installed:
`./thirdparty/fetch-thirdparty-deps.sh: 105: ./thirdparty/fetch-thirdparty-deps.sh: curl: not found`

So I guess curl is a dependency. I already added it to the list on [the Compiling page on the wiki.](https://github.com/OpenRA/OpenRA/wiki/Compiling). This adds the package names to INSTALL.md as well.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -42,6 +42,7 @@ Debian/Ubuntu
 * libsdl2-2.0-0
 * xdg-utils
 * zenity
+* curl
 
 openSUSE
 --------
@@ -54,6 +55,7 @@ openSUSE
 * lua51
 * xdg-utils
 * zenity
+* curl
 
 Gentoo
 ------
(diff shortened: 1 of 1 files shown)


Mailaender:
Yes, indeed. Not everything is downloaded via `nuget`. :+1:

obrakmann:
We might as well add the zip/unzip package while we're at it.

pchote:
This is a bit of a slippery slope... we also depend on `make`, `awk`, `sed`, and `cp`, but listing those would be silly.

obrakmann:
Yeah, but those are pretty much expected to be installed, while zip isn't. Oh well. I added it to the wiki instead. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1ccb816dd6e9ceeed489c91484d154d44141ec5e)

PR #11633, opened 2016-07-11
Add an Image option to WithIdleOverlay.

GraionDilach:


GraionDilach:
I have no idea why the lint rule went bogus on oldlst. It makes absolutely no sense to me.

pchote:
Does the lint rule understand the "if null fall back to default" logic?

GraionDilach:
Not quite, true that - https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Lint/CheckSequences.cs#L81 should be a string.IsNullOrEmpty - but there is absolutely nothing which could even trigger that there, oldlst doesn't use any WIO-derivate.

Mailaender:
I guess it crashes inside the `FirstOrDefault` so adding a null check there might help (untested):

```csharp
var imageField = fields.FirstOrDefault(f => f.Name != null && f.Name == sequenceReference.ImageReference);
```

GraionDilach:
Tested that but couldn't resolve taht issue.

I'm closing this for now to not clogger the queue with it. Will return it when I figure out what went wrong.

PR #11638, opened 2016-07-11
Deprecate terrain IsWater check - Part 1

reaperrr:
This removes all remaining `IsWater` checks except for `CreateEffectWarhead`. The latter is better done in its own PR, because it would make sense to refactor `ImpactTypes` along with it, which might require some rather large upgrade rule and amount of yaml changes.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Parachutable.cs
+++ b/OpenRA.Mods.Common/Traits/Parachutable.cs
@@ -9,6 +9,7 @@
  */
 #endregion
 
+using System.Collections.Generic;
 using System.Linq;
 using OpenRA.Mods.Common.Effects;
 using OpenRA.Traits;
@@ -31,6 +32,9 @@ namespace OpenRA.Mods.Common.Traits
 		public readonly string WaterImpactSound = null;
 		[SequenceReference("Image")] public readonly string WaterCorpseSequence = null;
 		[PaletteReference] public readonly string WaterCorpsePalette = "effect";
+
+		[Desc("Terrain types on which to display WaterCorpseSequence.")]
+		public readonly HashSet<string> WaterTerrainTypes = new HashSet<string> { "Water" };
 
 		public readonly int FallRate = 13;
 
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.Common/Traits/Render/WithCrateBody.cs
```diff
@@ -74,7 +79,9 @@ void INotifyParachuteLanded.OnLanded(Actor ignore)
 
 		void PlaySequence()
 		{
-			var sequence = self.World.Map.GetTerrainInfo(self.Location).IsWater ? info.WaterSequence : info.LandSequence;
+			var cell = self.Location;
+			onWater = self.World.Map.Contains(cell) && info.WaterTerrainTypes.Contains(self.World.Map.GetTerrainInfo(cell).Type);
```
Mailaender: Just curious: Why did you add `self.World.Map.Contains(cell)` here?

Mailaender:
:+1:

pchote:
Yay!  Thanks for making some progress on this.

OpenRA.Mods.Common/Traits/Render/WithCrateBody.cs
reaperrr: Originally I wanted to make sure the sequence is only played when the cell is in the visible map area. But to be honest, I'm not sure such a safeguard is really needed, and I see now that this would be the wrong place to put it anyway, as it would just switch to LandSequence instead. I guess I'll just remove it.

abcdefg30:
Is there a specific reason why you used `bool onWater` and not just `var onWater`?

Phrohdoh:
`var` cannot be used outside of a method body.

abcdefg30:
> `var` cannot be used outside of a method body.

I meant to implicitly ask why we need this variable outside of the method body.

OpenRA.Mods.Common/Traits/Parachutable.cs
```diff
@@ -47,6 +51,8 @@ class Parachutable : INotifyParachuteLanded
 		readonly ParachutableInfo info;
 		readonly IPositionable positionable;
 
+		bool onWater;
```
obrakmann: abcdefg30 is right here, this doesn't need to be global. Same in the other file below.

obrakmann:
Looks fine otherwise, :+1:

reaperrr:
Updated.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8b66ea95d7e16c270ccbf9b23fee73fed56e52b6)

PR #11641, opened 2016-07-12
Convert ParticleDensityFactor from a float to an int

abcdefg30:
Split from #11573.

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/WeatherOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/World/WeatherOverlay.cs
@@ -19,8 +19,8 @@ namespace OpenRA.Mods.Common.Traits
 	[Desc("Adds a particle-based overlay.")]
 	public class WeatherOverlayInfo : ITraitInfo, ILobbyCustomRulesIgnore
 	{
-		[Desc("Factor for particle density. As higher as more particles will get spawned.")]
-		public readonly float ParticleDensityFactor = 0.0007625f;
+		[Desc("Average number of particles per 100x100 px square.")]
+		public readonly int ParticleDensityFactor = 8;
 
 		[Desc("Should the level of the wind change over time, or just stick to the first value of WindLevels?")]
 		public readonly bool ChangingWindLevel = true;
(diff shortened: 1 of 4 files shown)


Mailaender:
Can you give a small summary why this is beneficial or required?

abcdefg30:
I converted it because our Lua API doesn't understand floats. [IRC logs](http://logs.openra.net/?year=2016&month=02&day=01#14:58:59):
> 14:58:59  (abcdefg30) ok, this is stupid..., why do we have a ParticleDensity of 0.0007625f and not just 7625 and then scale internally down
14:59:16  (abcdefg30) will be fun with lua now, as we don't have floats there
15:00:53  (pchote) abcdefg30: what does that number represent?
15:01:02  (abcdefg30) Desc("Factor for particle density. As higher as more particles will get spawned.")
15:01:18  (abcdefg30) ok, maybe a big int isn't too nice
15:01:23  (abcdefg30) but 0.000something?
15:01:36  (pchote) abcdefg30: what does the factor actually represent?
15:01:48  (pchote) if its just an arbitrary number then I agree that it should be bigger
15:01:57  (pchote) is it particles per pixel squared?
15:02:29  (pchote) because you could redefine it as particles per 100x100px, so that it is 10000 times bigger
15:02:35  (pchote) which is then a reasonable number
15:03:10  (abcdefg30) it does "particleCount = Game.Renderer.Resolution.Width * Game.Renderer.Resolution.Height * factor"
15:03:37  (pchote) yeah, so it is in particles per pixels squared
15:03:46  (pchote) I suggest ^
15:04:02  (pchote) you could then just default it to 7
15:04:03  (pchote) or 8
15:05:04  (abcdefg30) ok, then ^ would kinda go "particleCount = particleCount / 10000"?
15:05:28  (abcdefg30) fun, I used 10000 as temporary scale factor to test the lua stuff
15:05:35  (pchote) particleCount = Game.Renderer.Resolution.Width * Game.Renderer.Resolution.Height * factor / 10000;
15:05:42  (abcdefg30) k thx
15:06:27  (pchote) and then Desc("Factor for particle density in units of average particles per 100x100 px square.")
15:06:33  (pchote) or maybe simpler:
15:06:46  (pchote) Desc("Average number of particles per 100x100 px square.")

Unfortunately I forgot that @pchote suggest a number just 10000 bigger, so I made it 1000000 times bigger. Can easily change that though if wished.

Mailaender:
TL;DR: Lua doesn't know about floats. :+1:

abcdefg30:
Reverted back to a 10000 times bigger value.

GraionDilach:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/907347ca33424f475ad2a6c9d378857f583a7d9a)

PR #11499, opened 2016-06-21
Merged PoisonedByTiberium and DamagedWithoutFoundations into DamagedByTerrain

Mailaender:
This allows veins to damage buildings. Closes #9690.

PR #11653, opened 2016-07-15
Fixed a division through zero in DamagedByTerrain.

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/11499 after Coverity complained about it.

### DIFF
--- a/OpenRA.Mods.Common/Traits/DamagedByTerrain.cs
+++ b/OpenRA.Mods.Common/Traits/DamagedByTerrain.cs
@@ -62,6 +62,9 @@ namespace OpenRA.Mods.Common.Traits
 					safeTiles++;
 			}
 
+			if (totalTiles == 0)
+				return;
+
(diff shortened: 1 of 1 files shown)


obrakmann:
:+1:

OpenRA.Mods.Common/Traits/DamagedByTerrain.cs
```diff
@@ -62,6 +62,9 @@ public void AddedToWorld(Actor self)
 					safeTiles++;
 			}
 
+			if (totalTiles < 1)
```
Phrohdoh: I think this would be better expressed as `== 0`, because we aren't ever worried about a negative value.

Mailaender: Okay.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/9205bbe00c42b8f23ec5c4a4a78b311f5925dc9b...d7ef2165915e984791a1acab72f351662ddc0f94)

PR #11644, opened 2016-07-13
Cached the NSIS installation on AppVeyor CI

Mailaender:
Following the advice of https://www.appveyor.com/docs/build-cache#caching-chocolatey-packages

PR #11656, opened 2016-07-15
Fixed AppVeyor failing at the MarkDown to HTML step

Mailaender:
It looks like I accidentally regressed pandoc when caching NSIS in https://github.com/OpenRA/OpenRA/pull/11644. Let's see if this fixes the build.


abcdefg30:
Appveyor passes, so 👍 .

GraionDilach:
:+1:

PR #11657, opened 2016-07-16
Replaced Chocolatey with more reliable alternatives

Mailaender:
TIL that NSIS is already pre-installed http://www.appveyor.com/docs/installed-software#tools and there is an API to render GFM using https://developer.github.com/v3/markdown/ getting rid of @chocolatey so we will hopefully get a more reliable Windows CI post compile.

### DIFF
--- a/ConvertFrom-Markdown.ps1
+++ b/ConvertFrom-Markdown.ps1
@@ -1,0 +1,65 @@
+<#
+.SYNOPSIS
+   Converts Markdown formatted text to HTML.
+.DESCRIPTION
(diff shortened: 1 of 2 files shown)


abcdefg30:
👍 once `ConvertFrom-Markdown.ps1` is (visually) cleaned up.

Mailaender:
"visually" cleaned?

abcdefg30:
Ah well, never mind.

Mailaender:
Updated with 2365fbdabe1f72d83b22919ebfbecfe80b7a64cf.

obrakmann:
:+1:

PR #11658, opened 2016-07-16
Add allies04

abcdefg30:
As we don't support taking the map from the end of allies02 yet, I just added a "placeholder", trying to suggest that the base was attacked meanwhile and partly destroyed.

I also added soviet convoys that pass the area from time to time, to get some more action going. I thought it's a good idea as the whole thing is about securing the area for convoys.


Mailaender:
A nice modernization. :white_check_mark:

r34ch:
I can't comment on the lua, but the convoys are a great touch.

A couple things;
- The AI doesn't attack concrete walls. In fact concrete walls make the mission very easy on hard.
- Is it possible to move the secondary mission trigger cells outside the map border? I found that if a convoy truck reaches a cell within in the map boarder on the right hand side you fail the secondary mission even though I think its still possible to kill it.

side

abcdefg30:
Hm, maybe I should just remove concrete walls from this mission entirely.

MustaphaTR:
I don't know how you made that. But in PS Edition, you start the mission with CYard and a PP also some units, iirc. I need to check but i think we need to go with this.

MustaphaTR:
OK, here are the screenshots, it indeed reset in PS Edition.

This was how i left mission 2:
screenshot_2016-07-24-17-37-18

This is the Mission 4. Just the minelayer comes from north barracks, cyard and pp. No other units:
screenshot_2016-07-24-17-53-51

abcdefg30:
Made concrete walls unbuildable for now.

mods/ra/maps/allies-04/allies04.lua
```diff
@@ -0,0 +1,170 @@
...
+WorldLoaded = function()
+	player = Player.GetPlayer("Greece")
+	ussr = Player.GetPlayer("USSR")
+
+	Camera.Position = AlliedConyard.CenterPosition
+
+	InitObjectives()
+
+	local difficulty = Map.LobbyOption("difficulty")
+	CovoyDelay = CovoyDelays[difficulty]
```
obrakmann: Convoy?

mods/ra/maps/allies-04/allies04-AI.lua
```diff
@@ -0,0 +1,264 @@
...
+				else
+					yak.Wait(DateTime.Seconds(5))
+				end
+			end
+
+			if target and yak.AmmoCount() > 0 then
+				yak.Attack(target)
+			else
+				yak.ReturnToBase()
+				yak.Resupply()
```
obrakmann: The Yak gets stuck in a `WaitFor` activity here. I'm not quite sure, which one. Probably it's [this one](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs#L39), but simply adding a `yak.Wait(1)` should have fixed that then, which it didn't.

obrakmann:
This looks fine, aside from the aircraft issue.

mods/ra/maps/allies-04/allies04.lua
abcdefg30: Oops, fixed. Thanks.

abcdefg30:
Updated.
- Made the advanced power plant and supply truck unbuildable
- Fixed the yaks here and in `allies-05a`

> Is it possible to move the secondary mission trigger cells outside the map border? I found that if a convoy truck reaches a cell within the map boarder on the right hand side you fail the secondary mission even though I think its still possible to kill it.

@r34ch sorry, but it isn't possible. The convoy units can't drive outside the map.

obrakmann:
Alright, seems to work ok now, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1bee32d04a775223189311c4a32fd0227433bd0f)

Issue #11660, opened 2016-07-16
Replays on bleed crash with an out-of-sync exception


obrakmann:
some time during playback, not immediately at the start or end.

Edit: commit 5b61206
[This replay](https://seacloud.cc/d/5d1adbebce/files/?p=/ora_bug11660_replay_desync.orarep) desyncs in frame 4476.

GraionDilach:
I already get that replay crashed on https://github.com/OpenRA/OpenRA/commit/8c8fa2d258f45d797547a732478efab7ad912447 and reading from the logs where you were talking about SharedRandom usage, I'd blame https://github.com/OpenRA/OpenRA/pull/10684/files#diff-90995d9756122b52b0ff4cedc81af1b8R59.

Holloweye:
Crash when settings building in multiplayer or capturing a oil derrick

Mailaender:
See https://github.com/OpenRA/OpenRA/issues/11721 for a suggested fix.

PR #11669, opened 2016-07-19
Fixed inappropriate usage of SharedRandom

Mailaender:
A broad-spectrum anti-desync against https://github.com/OpenRA/OpenRA/issues/11660.

### DIFF
--- a/OpenRA.Mods.Common/Projectiles/NukeLaunch.cs
+++ b/OpenRA.Mods.Common/Projectiles/NukeLaunch.cs
@@ -60,7 +60,7 @@ namespace OpenRA.Mods.Common.Effects
 
 			pos = launchPos;
 			if (weapon.Report != null && weapon.Report.Any())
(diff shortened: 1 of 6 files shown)


pchote:
The desync in #11660 isn't caused by SharedRandom.  SharedRandom is simply the first piece of `[Sync]`ed world state that the diverged game flow touches.  Removing these without understanding the actual desync will just push the bug to an even harder to reproduce corner of the engine.  This would be very unwise.

Mailaender:
These are all safe imho as they have decorational or otherwise game state irrelevant pieces not use the expensive shared and synced random.

pchote:
@obrakmann and @abcdefg30's testing appears to have isolated the desync to the scorch removals.  This is not an irrelevant piece, and there is no perf difference between the shared and synced randoms.  The scorch code was intentionally using synced random and sync-safe code so that we can have cross-client consistency when somebody finally decides to implement multicell craters.

Mailaender:
Is the weapon sound randomization also shared on purpose?

PR #11670, opened 2016-07-19
Check that noiseDirection is not a zero

evgeniysergeev:
fix unexpected freezes that can happen because of infinity loop in RescanForTargets

### DIFF
--- a/OpenRA.Mods.D2k/Traits/Sandworm.cs
+++ b/OpenRA.Mods.D2k/Traits/Sandworm.cs
@@ -111,6 +111,13 @@ namespace OpenRA.Mods.D2k.Traits
 
 			while (!self.World.Map.Contains(moveTo) || !mobile.CanEnterCell(moveTo, null, false))
 			{
+				// without this check, this while can be infinity loop
+				if (moveTo == self.Location)
+				{
+					self.CancelActivity();
+					return;
+				}
+
 				noiseDirection /= 2;
 				moveTo = self.World.Map.CellContaining(self.CenterPosition + noiseDirection);
 			}


OpenRA.Mods.D2k/Traits/Sandworm.cs
```diff
@@ -111,6 +111,10 @@ void RescanForTargets(Actor self)
 
 			while (!self.World.Map.Contains(moveTo) || !mobile.CanEnterCell(moveTo, null, false))
 			{
+				// without this check, this while statement can be infinity loop
```
pchote: It would be better to move the `(moveTo == self.Location)` check below inside this loop instead.  That will catch the error case faster, and cancel the activity before returning.

evgeniysergeev: >  move the (moveTo == self.Location)

not moved, but copied this check into while statement and leave additional check after while statement.

OpenRA.Mods.D2k/Traits/Sandworm.cs
```diff
@@ -111,6 +111,13 @@ void RescanForTargets(Actor self)
 
 			while (!self.World.Map.Contains(moveTo) || !mobile.CanEnterCell(moveTo, null, false))
 			{
+				// without this check, this while can be infinity loop
+				if (moveTo == self.Location)
```
obrakmann: How can the sandworm ever end up at a location that is either not in the map, or that it cannot enter?

pchote: Spawner outside the cordon, or on a cell with another actor?

obrakmann:
Fair enough. :+1:

GraionDilach:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/6e9a46d935c699ce3b8c20d2db2d66e443be7e54)

PR #11671, opened 2016-07-19
Delay growing back the shroud on actor killed

huwpascoe:
For the bounty #7916

I don't have an opinion on how long the delay should be so feedback please!

### DIFF
--- a/OpenRA.Game/Traits/World/Shroud.cs
+++ b/OpenRA.Game/Traits/World/Shroud.cs
@@ -40,7 +40,7 @@ namespace OpenRA.Traits
 		public object Create(ActorInitializer init) { return new Shroud(init.Self, this); }
 	}
 
-	public class Shroud : ISync, INotifyCreated
+	public class Shroud : ISync, INotifyCreated, ITick
 	{
 		public event Action<IEnumerable<PPos>> CellsChanged;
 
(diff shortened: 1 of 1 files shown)


OpenRA.Game/Traits/World/Shroud.cs
```diff
@@ -92,7 +94,7 @@ public Shroud(Actor self, ShroudInfo info)
 			explored = new CellLayer<bool>(map);
 		}
 
-		void INotifyCreated.Created(Actor self)
+		public void Created(Actor self)
```
Mailaender: Don't revert this although it is used inconsistently in the code base. https://github.com/OpenRA/OpenRA/pull/10143

huwpascoe: Should I explicitly declare ITick.Tick?

pchote:
We've already gone most of the way to implementing #7916 by generalizing the shroud code to support more than just actors (as described in the IRC quote there).  I think it would be a bad idea to take this as-is because it is much more limited than the alternative, has extra perf overheads, and puts that earlier work to waste.

Can you instead please look into replacing the `Actor` arguments to Add and Remove with a more general type, and then have actors spawn a new IEffect class on death that gives visibility for a set amount of time.

huwpascoe:
Sorry, it's too vague an issue at that point. I'm gonna have to decline.

Issue #3923, opened 2013-10-10
Filter repeated actions


Mailaender:
> In AoE, clicking repeatedly when the users were excitedly attacking (clik-lik-lik-lik-lik -- go go go ) was causing huge spikes in the number of commands issued per second -- and if they were pathing a large group of units -- huge spikes in the network demand as well. A simple filter to discard repeat commands at the same location drastically reduced the impact of this behavior.

http://www.gamasutra.com/view/feature/131503/1500_archers_on_a_288_network_.php?page=2 
##

abcdefg30:
https://github.com/OpenRA/OpenRA/pull/11673#issuecomment-619400522 rejected the basic idea here. It is indeed not as easy to do this while issuing orders since some orders might behave different if multiple come through. #19397 offers an idea to solve this on the `ResolveOrder` side of things which should increase usability.

PR #11673, opened 2016-07-19
Added a filter for repeated move orders

Mailaender:
as suggested in https://github.com/OpenRA/OpenRA/issues/3923#issuecomment-34561991. You can see the benefit of units not halting for every click as this avoids https://github.com/OpenRA/OpenRA/issues/3288 at least when you click at the exact same tile.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
+++ b/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
@@ -115,6 +115,8 @@ namespace OpenRA.Mods.Common.Traits
 		bool airborne;
 		bool cruising;
 
+		CPos currentDestination;
+
 		public Aircraft(ActorInitializer init, AircraftInfo info)
 		{
 			Info = info;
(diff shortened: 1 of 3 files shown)


pchote:
I don't think this is going to work properly in all the cases we care about.  There are lots of places where other code orders the actor to move somewhere, and they will cause the `currentDestination` to get out of sync with the actual world state.

Mailaender:
It was based on your suggestion https://github.com/OpenRA/OpenRA/issues/3923#issuecomment-34561991 back then.

Mailaender:
Would 

```csharp
static Order previousOrder;

static void ResolveOrder(Order order)
{
	if (order == previousOrder)
		System.Console.WriteLine("repeated order detected");
```

be a more suitable location? It seems that `Order` also might need a custom

```csharp
public override bool Equals(object obj)
```

pchote:
No. Issuing an order has various side effects (e.g. forcing a repath) which would cause noticable gameplay regressions if skipped. This is just a bad idea in general when considered in the context of how OpenRA works.

PR #11676, opened 2016-07-19
Added ScreenMap support for effects

Mailaender:
now with laser :red_circle: and tesla :zap: support. Closes https://github.com/OpenRA/OpenRA/issues/9443.

### DIFF
--- a/OpenRA.Game/Traits/World/ScreenMap.cs
+++ b/OpenRA.Game/Traits/World/ScreenMap.cs
@@ -13,6 +13,7 @@ using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
+using OpenRA.Effects;
 using OpenRA.Graphics;
 using OpenRA.Primitives;
 
@@ -33,6 +34,7 @@ namespace OpenRA.Traits
 		readonly Func<Actor, bool> actorIsInWorld = a => a.IsInWorld;
 		readonly Cache<Player, SpatiallyPartitioned<FrozenActor>> partitionedFrozenActors;
 		readonly SpatiallyPartitioned<Actor> partitionedActors;
+		readonly SpatiallyPartitioned<IEffect> partitionedEffects;
 		WorldRenderer worldRenderer;
 
 		public ScreenMap(World world, ScreenMapInfo info)
@@ -43,6 +45,7 @@ namespace OpenRA.Traits
 			partitionedFrozenActors = new Cache<Player, SpatiallyPartitioned<FrozenActor>>(
 				_ => new SpatiallyPartitioned<FrozenActor>(width, height, info.BinSize));
 			partitionedActors = new SpatiallyPartitioned<Actor>(width, height, info.BinSize);
+			partitionedEffects = new SpatiallyPartitioned<IEffect>(width, height, info.BinSize);
 		}
 
 		public void WorldLoaded(World w, WorldRenderer wr) { worldRenderer = wr; }
@@ -68,6 +71,13 @@ namespace OpenRA.Traits
 			partitionedFrozenActors[viewer].Add(fa, FrozenActorBounds(fa));
 		}
 
+		public void Add(IEffect effect, WPos position, Rectangle bounds)
+		{
+			var screenPos = worldRenderer.ScreenPxPosition(position);
+			var screenBounds = new Rectangle(screenPos.X - bounds.Width / 2, screenPos.Y - bounds.Height / 2, bounds.Width, bounds.Height);
+			partitionedEffects.Add(effect, screenBounds);
+		}
+
 		public void Remove(Player viewer, FrozenActor fa)
 		{
 			partitionedFrozenActors[viewer].Remove(fa);
@@ -83,9 +93,25 @@ namespace OpenRA.Traits
 			partitionedActors.Update(a, ActorBounds(a));
 		}
 
+		public void Update(IEffect effect, WPos position, Rectangle bounds)
+		{
+			if (!partitionedEffects.Contains(effect))
+				return;
+
+			var screenPos = worldRenderer.ScreenPxPosition(position);
+			var screenBounds = new Rectangle(screenPos.X - bounds.Width / 2, screenPos.Y - bounds.Height / 2, bounds.Width, bounds.Height);
+			partitionedEffects.Update(effect, screenBounds);
+		}
+
 		public void Remove(Actor a)
 		{
 			partitionedActors.Remove(a);
+		}
+
+		public void Remove(IEffect e)
+		{
+			if (e != null && partitionedEffects.Contains(e))
+				partitionedEffects.Remove(e);
 		}
 
 		public IEnumerable<FrozenActor> FrozenActorsAt(Player viewer, int2 worldPx)
@@ -120,6 +146,11 @@ namespace OpenRA.Traits
 			return ActorsInBox(RectWithCorners(a, b));
 		}
 
+		public IEnumerable<IEffect> EffectsInBox(int2 a, int2 b)
+		{
+			return partitionedEffects.InBox(RectWithCorners(a, b));
+		}
+
 		public IEnumerable<Actor> ActorsInBox(Rectangle r)
 		{
 			return partitionedActors.InBox(r).Where(actorIsInWorld);
@@ -128,6 +159,11 @@ namespace OpenRA.Traits
 		public IEnumerable<FrozenActor> FrozenActorsInBox(Player p, int2 a, int2 b)
 		{
 			return FrozenActorsInBox(p, RectWithCorners(a, b));
+		}
+
+		public IEnumerable<IEffect> EffectsInBox(Rectangle r)
+		{
+			return partitionedEffects.InBox(r);
 		}
 
 		public IEnumerable<FrozenActor> FrozenActorsInBox(Player p, Rectangle r)
--- a/OpenRA.Mods.Cnc/Effects/AnimatedBeacon.cs
+++ b/OpenRA.Mods.Cnc/Effects/AnimatedBeacon.cs
@@ -39,11 +39,17 @@ namespace OpenRA.Mods.Cnc.Effects
 			this.duration = duration;
 			this.delay = delay;
 
+			var world = owner.World;
+
 			if (!string.IsNullOrEmpty(beaconSequence))
 			{
-				beacon = new Animation(owner.World, beaconImage);
+				beacon = new Animation(world, beaconImage);
 				beacon.PlayRepeating(beaconSequence);
+				world.ScreenMap.Add(this, position, beacon.Image.Bounds);
 			}
+
+			if (duration > 0)
+				world.Add(new DelayedAction(duration, () => { world.Remove(this); world.ScreenMap.Remove(this); }));
 		}
 
 		void IEffect.Tick(World world)
--- a/OpenRA.Mods.Cnc/Graphics/TeslaZapRenderable.cs
+++ b/OpenRA.Mods.Cnc/Graphics/TeslaZapRenderable.cs
@@ -88,7 +88,10 @@ namespace OpenRA.Mods.Cnc.Graphics
 			cache.Do(c => c.Render(wr));
 		}
 
-		public Rectangle ScreenBounds(WorldRenderer wr) { return Rectangle.Empty; }
+		public Rectangle ScreenBounds(WorldRenderer wr)
+		{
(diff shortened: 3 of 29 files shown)


abcdefg30:
Needs a rebase.

OpenRA.Mods.Common/Effects/Beacon.cs
```diff
@@ -42,20 +43,25 @@ public Beacon(Player owner, WPos position, int duration, string beaconPalette, b
...
 			}
 
 			if (!string.IsNullOrEmpty(circleSprite))
 			{
-				circles = new Animation(owner.World, beaconCollection);
+				circles = new Animation(world, beaconCollection);
 				circles.Play(circleSprite);
 			}
 
+			var bounds = Rectangle.Union(arrow.Image.Bounds, circles.Image.Bounds);
```
abcdefg30: Can't `arrow` or `circles` be `null` here?

Mailaender: Absolutely right. Fixed.

OpenRA.Mods.Common/Projectiles/Missile.cs
```diff
@@ -826,6 +827,8 @@ public void Tick(World world)
 				ticksToNextSmoke = info.TrailInterval;
 			}
 
+			world.ScreenMap.Update(this, pos, anim.Image.Bounds);
```
reaperrr: `anim` can be `null`, so you should check for that, like in `Bullet` and `GravityBomb`.

reaperrr:
Makes a noticable difference with lots of off-screen contrails and smoke trails, so I'd like this to make it into Next+1.

reaperrr:
> Exception of type `System.ArgumentException`: Bounds of actor OpenRA.Mods.Common.Projectiles.LaserZap are empty.

I reproduced it multiple times with a large battle of laser-armed actors.

Changing line 77 in `BeamRenderable` to
> return new Rectangle(0, 0, length.HorizontalLength *+ 1*, length.VerticalLength *+ 1*);

fixed that particular crash, however I'm not sure whether that is the proper fix or just hides the real bug.

pchote:
> just hides the real bug.

reaperrr:
I suspect it happens when the beam shoots straight up,down,right or left, because beam width is not taken into account, so either X or Y will we 0 in that case. I don't know how to fix that at code level, though, other than with the ^ work-around.

pchote:
If that's true then the proper code-level fix will be to take the beam width into account when calculating the bounds.  This would also be needed to avoid issues with wide beams like the TS disruptor.

pchote:

Mailaender:
Rebased, changed the BeamRenderable rectangle calculation, added a null check for the missile anim and adapted the new Railgun code as well.

reaperrr:
Nice 👍

OpenRA.Mods.Cnc/Effects/AnimatedBeacon.cs
```diff
@@ -39,11 +39,17 @@ public AnimatedBeacon(Player owner, WPos position, int duration, string beaconPa
...
 			if (!string.IsNullOrEmpty(beaconSequence))
 			{
-				beacon = new Animation(owner.World, beaconImage);
+				beacon = new Animation(world, beaconImage);
 				beacon.PlayRepeating(beaconSequence);
+				world.ScreenMap.Add(this, position, beacon.Image.Bounds);
 			}
+
+			if (duration > 0)
+				world.Add(new DelayedAction(duration, () => { world.Remove(this); world.ScreenMap.Remove(this); }));
```
pchote: Rebase error?  The effect is now removed in `IEffect.Tick`

OpenRA.Mods.Cnc/Projectiles/TeslaZap.cs
```diff
@@ -55,6 +56,12 @@ public TeslaZap(TeslaZapInfo info, ProjectileArgs args)
 			ticksUntilRemove = info.Duration;
 			damageDuration = info.DamageDuration > info.Duration ? info.Duration : info.DamageDuration;
 			target = args.PassiveTarget;
+
+			var pos = args.GuidedTarget.IsValidFor(args.SourceActor) ? args.GuidedTarget.CenterPosition : args.PassiveTarget;
+			zap = new TeslaZapRenderable(args.Source, 0, pos - args.Source,
+				info.Image, info.BrightSequence, info.BrightZaps, info.DimSequence, info.DimZaps, info.Palette);
+			var world = args.SourceActor.World;
+			world.ScreenMap.Add(this, pos, zap.ScreenBounds(null));
```
pchote: Zaps jump about randomly, so I don't think we can trust the bounds from a single zap?  We would need to make a sufficiently larger box to cover all the places it could wander.

OpenRA.Mods.Common/Effects/Beacon.cs
```diff
@@ -48,17 +49,28 @@ public class Beacon : IEffect, IScriptBindable, IEffectAboveShroud
 			this.duration = duration;
 			this.delay = delay;
 
+			var world = owner.World;
+			var bounds = new Rectangle();
```
pchote: Style nit: `Rectangle bounds;`

OpenRA.Mods.Common/Effects/Beacon.cs
```diff
@@ -48,17 +49,28 @@ public class Beacon : IEffect, IScriptBindable, IEffectAboveShroud
...
+
 			if (!string.IsNullOrEmpty(arrowSprite))
 			{
-				arrow = new Animation(owner.World, beaconCollection);
+				arrow = new Animation(world, beaconCollection);
 				arrow.Play(arrowSprite);
+				bounds = Rectangle.Union(bounds, arrow.Image.Bounds);
```
pchote: I believe that the default Rectangle has a position of 0, so unioning this with that will give a giant box that extends to the corner of the map.

OpenRA.Mods.Common/Effects/ContrailFader.cs
```diff
@@ -21,19 +22,26 @@ public class ContrailFader : IEffect
...
 		{
+			if (ticks == 0)
+				world.ScreenMap.Add(this, pos, bounds);
+
 			if (ticks++ == trail.Length)
-				world.AddFrameEndTask(w => w.Remove(this));
+				world.AddFrameEndTask(w => { w.Remove(this); w.ScreenMap.Remove(this); });
 
 			trail.Update(pos);
+			world.ScreenMap.Update(this, pos, bounds);
```
pchote: `bounds` are never updated as the contrails move about the map?

OpenRA.Mods.Common/Traits/SupportPowers/ParatroopersPower.cs
```diff
@@ -108,6 +108,7 @@ public Actor[] SendParatroopers(Actor self, WPos target, bool randomize = true,
 					self.World.AddFrameEndTask(w =>
 					{
 						w.Remove(beacon);
+						w.ScreenMap.Remove(beacon);
```
pchote: Where are these added to the ScreenMap?

OpenRA.Mods.Common/Traits/SupportPowers/AirstrikePower.cs
```diff
@@ -92,6 +92,7 @@ public void SendAirstrike(Actor self, WPos target, bool randomize = true, int at
 					self.World.AddFrameEndTask(w =>
 					{
 						w.Remove(beacon);
+						w.ScreenMap.Remove(beacon);
```
pchote: Where are these added to the ScreenMap?

OpenRA.Mods.Common/Projectiles/Railgun.cs
```diff
@@ -145,17 +145,21 @@ public Railgun(ProjectileArgs args, RailgunInfo info, Color beamColor, Color hel
...
+				hitanim = new Animation(world, info.HitAnim);
+				world.ScreenMap.Add(this, target, hitanim.Image.Bounds);
+			}
 
-			CalculateVectors();
+			CalculateVectors(world);
 		}
 
-		void CalculateVectors()
+		void CalculateVectors(World world)
```
pchote: `world` is already available from `args.SourceActor.World`, so why pass it in again here?

OpenRA.Mods.Common/Projectiles/NukeLaunch.cs
```diff
@@ -64,12 +64,16 @@ public class NukeLaunch : IProjectile
...
 				ticks = turn;
+
+			firedBy.World.ScreenMap.Add(this, pos, anim.Image.Bounds);
 		}
 
 		public void Tick(World world)
 		{
 			anim.Tick();
 
+			world.ScreenMap.Update(this, pos, anim.Image.Bounds);
```
pchote: This needs to be updated *after* updating `pos`.

OpenRA.Mods.Cnc/Graphics/TeslaZapRenderable.cs
```diff
@@ -88,7 +88,10 @@ public void Render(WorldRenderer wr)
 			cache.Do(c => c.Render(wr));
 		}
 
-		public Rectangle ScreenBounds(WorldRenderer wr) { return Rectangle.Empty; }
+		public Rectangle ScreenBounds(WorldRenderer wr)
+		{
+			return new Rectangle(0, 0, length.HorizontalLength, length.VerticalLength);
```
pchote: This needs to account for the actual position `pos` of the renderable.

OpenRA.Mods.Cnc/Projectiles/TeslaZap.cs
pchote: Creating a `TeslaZapRenderable` here is also generally bogus.  Instead, define a rectangle from the source to the target with a sufficient margin to cover the wandering.

OpenRA.Mods.Common/Effects/Beacon.cs
pchote: Define `bounds` as a `Rectangle?`, and then initialize it unconditionally here.

OpenRA.Mods.Common/Effects/Beacon.cs
```diff
@@ -48,17 +49,28 @@ public class Beacon : IEffect, IScriptBindable, IEffectAboveShroud
...
 			}
 
 			if (!string.IsNullOrEmpty(circleSprite))
 			{
-				circles = new Animation(owner.World, beaconCollection);
+				circles = new Animation(world, beaconCollection);
 				circles.Play(circleSprite);
+				bounds = Rectangle.Union(bounds, circles.Image.Bounds);
```
pchote: And then if `bounds` is null here initialize it, otherwise union with it.

OpenRA.Mods.Common/Effects/ContrailFader.cs
```diff
@@ -21,19 +22,26 @@ public class ContrailFader : IEffect
...
 		ContrailRenderable trail;
 		int ticks;
+		Rectangle bounds;
 
 		public ContrailFader(WPos pos, ContrailRenderable trail)
 		{
 			this.pos = pos;
 			this.trail = trail;
+
+			bounds = new Rectangle(0, 0, trail.Length, trail.Length);
```
pchote: `trail.Length` is the number of items in an array, not a measurement of size!

OpenRA.Game/Traits/World/ScreenMap.cs
```diff
@@ -68,6 +71,13 @@ public void Add(Player viewer, FrozenActor fa)
 			partitionedFrozenActors[viewer].Add(fa, FrozenActorBounds(fa));
 		}
 
+		public void Add(IEffect effect, WPos position, Rectangle bounds)
```
pchote: Several of my earlier comments raise issues about passing in `Rectangles` that have a bogus origin.
I just noticed here that you aren't even using the origin here or in `Update`, so `Rectangle` is the wrong thing to use.

Please replace these with two overloads that take a `Sprite` or a `Size`, solving the need to define dummy positions that can cause confusion.

pchote (changes requested):
Some more issues running until I discovered what was going on with `ScreenMap.Add`/`.Update`.  There may be still other issues with the effects.

Mailaender:
I made several crucial errors and have problems fixing them as well reliably testing if the calculations are correct so I better leave this to someone who has a better understanding.

PR #8004, opened 2015-04-22
Remove deprecated Buildable.Owner field.

pchote:

PR #11679, opened 2016-07-20
Added Scott's Allies03 co-op multiplayer mission as "Exodus"

Mailaender:
@ScottNZ's original mission converted from the [C# mission code found in release-20131223](https://github.com/OpenRA/OpenRA/blob/release-20131223/OpenRA.Mods.RA/Missions/Allies03Script.cs). This port is very faithful to the original. However I changed a lot of checks that were done every n ticks to use triggers instead, which didn't exist back then. As a result this should run pretty smooth.


OpenRA.Mods.Common/Scripting/Properties/TransportProperties.cs
```diff
@@ -31,6 +31,9 @@ public TransportProperties(ScriptContext context, Actor self)
 		[Desc("Specifies whether transport has any passengers.")]
 		public bool HasPassengers { get { return cargo.Passengers.Any(); } }
 
+		[Desc("Specifies the amount of passengers.")]
+		public int Passengers { get { return cargo.Passengers.Count(); } }
```
Phrohdoh: This should be `PassengerCount`, just `Passengers` would make me think this is a table of actors.

Mailaender: I somehow realized that as well when writing the commit message. Fixed.

r34ch:
Again can't comment on the code but the mission on hard is very easy. Perhaps the friendly reinforcements should be at intervals of units evacuated? The AI air units mostly circled their bases and AI attacks in general were mostly comprised of single units.

Mailaender:
I fixed the yak air raids. There were several problems I solved:
* `Attack` doesn't work without visibility so I added back the `CAMERA.Large`.
* `FlyCircle` is an activity that doesn't count as idle so the re-targeting never triggered.
* The yaks picked invalid targets. I restrained them to buildings and units now.

The constant yak harassment should increase difficulty.

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,394 @@
...
+	UserInterface.SetMissionText(UnitsEvacuated .. "/" .. unitsEvacuatedThreshold .. " units evacuated.")
+	Utils.Do(humans, function(player)
+		if player then
+			evacuateUnits = player.AddPrimaryObjective("Evacuate " .. unitsEvacuatedThreshold .. " units.")
+			destroyAirbases = player.AddSecondaryObjective("Destroy the nearby Soviet airbases.")
+			evacuateMgg = player.AddSecondaryObjective("Evacuate at least one mobile gap generator.")
+		end
+	end)
+
+	Trigger.OnAllKilled(SovietAirfields, function()
```
abcdefg30: This wants to be `OnAllKilledOrCaptured`, I think.

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,394 @@
...
+	Reinforcements.Reinforce(allies1, units, { Allies1EntryPoint.Location, Allies1MovePoint.Location })
+
+	if allies2 then
+		Reinforcements.Reinforce(allies2, units, { Allies2EntryPoint.Location, Allies2MovePoint.Location })
+	end
+
+	Utils.Do(humans, function(player)
+		if player and player.IsLocalPlayer then
+			Trigger.AfterDelay(DateTime.Seconds(3), function()
+				Media.PlaySpeechNotification(allies1, "AlliedReinforcementsArrived")
```
abcdefg30: `allies1` -> `player` I guess?

abcdefg30: I also found this a bit confusing, as I expected more reinforcements came in from another direction and I just didn't see them. Not sure how to fix that though, maybe reduce the delay to 2 seconds?

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,394 @@
...
+	hard = 500,
+	normal = 500,
+	easy = 600
+}
+
+MinAttackAtFrame =
+{
+	hard = 200,
+	normal = 100,
+	easy = 50
```
abcdefg30: This seems to increase the delay with difficulty. Intentional or swapped?

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,394 @@
...
+SendSovietParadrop = function()
+	local sovietParadrops = SovietParadrops[Difficulty]
+
+	if (SovietParadrop > sovietParadrops) then
+		return
+	end
+
+	SovietParadrop = SovietParadrop + 1
+
+	Media.PlaySpeechNotification(allies1, "SovietForcesApproaching")
```
abcdefg30: We'll want this for `allies2` as well, when he/she/it exists.

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,394 @@
...
+
+	local scatteredUnits = Utils.Where(soviets.GetGroundAttackers(), function(unit)
+		return not Map.IsNamedActor(unit) and unit.IsIdle
+	end)
+	Utils.Do(scatteredUnits, IdleHunt)
+end
+
+AircraftTargets = function()
+	local targets = Utils.Where(Map.ActorsInWorld, function(a)
+		return a.Owner == allies1 and
```
abcdefg30: `(a.Owner == allies1 or a.Owner == allies2) and ...`?

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,394 @@
...
+			(a.HasProperty("StartBuildingRepairs") or a.HasProperty("Move"))
```
abcdefg30: `a.HasProperty("Health")` would make more sense, imho. (What were those "invalid targets" you were talking about anyway?)

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,394 @@
...
+			end)
+
+			Trigger.OnPlayerLost(player, function()
+				Media.PlaySpeechNotification(player, "MissionFailed")
+			end)
+		end
+	end)
+
+	unitsEvacuatedThreshold = UnitsEvacuatedThreshold[Difficulty]
+	UserInterface.SetMissionText(UnitsEvacuated .. "/" .. unitsEvacuatedThreshold .. " units evacuated.")
```
abcdefg30: I think it would look better if we use the player color from the allied campaign here. Just visual and just me though.

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,394 @@
...
+		end)
+		if #grounded > 0 then
+			Utils.Do(grounded, YakAttack)
+		end
+	end
+end
+
+UnitsEvacuated = 0
+EvacuateAlliedUnit = function(unit)
+	if (unit.Owner == allies1 or unit.Owner == allies2) and unit.IsMobile then
```
abcdefg30: This also wants a `unit..HasProperty("Move")` check.

mods/ra/maps/exodus/exodus.lua
abcdefg30: Ah, I see. Yaks can't attack camo pillboxes. Those aren't excluded though, so maybe we want to check if the unit is cloaked?

mods/ra/maps/exodus/exodus.lua
Mailaender: Changed it to "Allied reinforcements have arrived from the North." to avoid confusion.

mods/ra/maps/exodus/exodus.lua
Mailaender: Copy paste mistake: https://github.com/OpenRA/OpenRA/blob/release-20131223/OpenRA.Mods.RA/Missions/Allies03Script.cs#L408 has the values I wanted.

Mailaender:
Thanks for the extensive review. Updated accordingly. Also removed a
```yaml
Buildable:
		Owner: allies, soviets
```
leftover from when we didn't have tech capturing. https://github.com/OpenRA/OpenRA/pull/8004

mods/ra/maps/exodus/exodus.lua
Mailaender: I just removed `HBOX` as that is easier and we are in a low tech mode anyway.

obrakmann:
I got this exception some time into the game:
```
Actor 'camera.paradrop' does not define a property 'Hunt'
```

OpenRA.Mods.Common/Scripting/Properties/AircraftProperties.cs
```diff
@@ -60,5 +60,12 @@ public void Resupply()
 		{
 			Self.QueueActivity(new ResupplyAircraft(Self));
 		}
+
+		[ScriptActorPropertyActivity]
+		[Desc("Returns true if the aircraft circling over it's home base.")]
+		public bool IsOnHoldingPattern
```
obrakmann: Just call it `IsCircling`. Also the description is misleading (the "over its home base" part).

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,398 @@
...
+
+IdleHunt = function(unit)
+	Trigger.OnIdle(unit, unit.Hunt)
+end
+
+ManageSovietUnits = function()
+	Utils.Do(SovietRallyPoints, function(rallyPoint)
+		local radius = WDist.FromCells(8)
+		local units = Map.ActorsInCircle(rallyPoint.CenterPosition, radius, function(actor)
+			return actor.Owner == soviets
```
obrakmann: I suppose adding `and actor.HasProperty("Hunt")` here would fix the crash I saw.

mods/ra/maps/exodus/rules.yaml
```diff
@@ -0,0 +1,185 @@
+Player:
+	MissionObjectives:
+	PlayerResources:
+		DefaultCash: 5000
```
obrakmann: The whole `Player` block is redundant, it already has `MissionObjectives` and 5000 already is the `DefaultCash` default value.

obrakmann:
Looks otherwise good to me. It's nice to have these back!

mods/ra/maps/exodus/rules.yaml
obrakmann: Err, slight correction. Because of [https://github.com/OpenRA/OpenRA/pull/11679/files#diff-c170037b524787407d1f07a0151d1d71R312](this check), you actually do want to set the `Cooperative` flag on `MissionObjectives`.

Background: You can leave the flag unset if you're sure that you will always change both players objectives the same way at the same time, but in the linked case, only one player's objective would be changed.

Mailaender:
Updated accordingly. Yeah, I got a bit nostalgic when looking at http://www.moddb.com/games/openra/images/evacuation#imagebox

mods/ra/maps/exodus/map.yaml
```diff
@@ -0,0 +1,1429 @@
...
+		Owner: Soviets
+	Actor213: tsla
+		Location: 71,59
+		Owner: Soviets
+	Actor295: brik
+		Location: 73,59
+		Owner: Soviets
+	Actor180: v01
+		Location: 100,29
+		Owner: Soviets
```
obrakmann: Is there a reason this church is owned by the Soviets?

OpenRA.Mods.Common/Scripting/Properties/AircraftProperties.cs
```diff
@@ -60,5 +60,12 @@ public void Resupply()
 		{
 			Self.QueueActivity(new ResupplyAircraft(Self));
 		}
+
+		[ScriptActorPropertyActivity]
+		[Desc("Returns true if the aircraft is flying circles.")]
```
obrakmann: Can we add that this is a hacky workaround for our sucky activities and that people shouldn't rely on it existing in the long term, please?

mods/ra/maps/exodus/rules.yaml
```diff
@@ -0,0 +1,180 @@
...
+		Scripts: exodus.lua
+	ScriptLobbyDropdown@difficulty:
+		ID: difficulty
+		Label: Difficulty
+		Values:
+			easy: Easy
+			normal: Normal
+			hard: Hard
+		Default: normal
+
```
obrakmann: The debug menu can still be enabled.

mods/ra/maps/exodus/rules.yaml
obrakmann: Actually, the default cash thing was wrong, with the campaign rules, the default is 0. I'm afraid it needs to be there after all.

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,398 @@
...
+		local circling = Utils.Where(sovietYaks, function(yak) return yak.IsCircling end)
+		if #circling > 0 then
+			Utils.Do(circling, YakAttack)
+		end
+
+		local grounded = Utils.Where(sovietYaks, function(yak)
+			return yak.CenterPosition.Z < 1 and yak.AmmoCount() == yak.MaximumAmmoCount()
+		end)
+		if #grounded > 0 then
+			Utils.Do(grounded, YakAttack)
```
obrakmann: This doesn't work because of [some weird code in `ResupplyAircraft`](https://github.com/OpenRA/OpenRA/blob/8233857be1e4684d0b7519760148404fecc0c8e8/OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs#L39). My suggestion would be to simply insert a `yak.Stop()` before `yak.Attack(target)` in `YakAttack`.

mods/ra/maps/exodus/rules.yaml
Mailaender: Why should we disallow it?

mods/ra/maps/exodus/exodus.lua
Mailaender: Added the `yak.Stop()`.

OpenRA.Mods.Common/Scripting/Properties/AircraftProperties.cs
Mailaender: Sounds a bit harsh, but sure.

mods/ra/maps/exodus/map.yaml
Mailaender: I assume it is a mistake in the original map design, a rather ironic one as well. Fixed.

mods/ra/maps/exodus/rules.yaml
obrakmann: I guess to prevent people from cheating on missions by being able to see the whole map. But of course they can now also just look at them in editor. I guess there's really no point to it then.

Could I ask you to add a commit that removes [these two lines](https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/maps/evacuation/rules.yaml#L6-L7) then, please?

OpenRA.Mods.Common/Scripting/Properties/AircraftProperties.cs
```diff
@@ -8,7 +8,7 @@
  * information, see COPYING.
  */
 #endregion
-
+
```
obrakmann: OpenRA.Mods.Common/Scripting/Properties/AircraftProperties.cs:L11: [SP2000] Invalid spacing at the end of the line.

obrakmann:
:+1:

mods/ra/maps/exodus/rules.yaml
Mailaender: Done.

Mailaender:
Fixed.

mods/ra/maps/exodus/exodus.lua
WolfGaming: yeah this didn't crash before since every actor had the hunt property.

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,399 @@
...
+	local scatteredUnits = Utils.Where(soviets.GetGroundAttackers(), function(unit)
+		return not Map.IsNamedActor(unit) and unit.IsIdle and unit.HasProperty("Hunt")
+	end)
+	Utils.Do(scatteredUnits, IdleHunt)
+end
+
+AircraftTargets = function()
+	local targets = Utils.Where(Map.ActorsInWorld, function(a)
+		return (a.Owner == allies1 or a.Owner == allies2)
+			and (a.HasProperty("Move") or a.HasProperty("StartBuildingRepairs"))
```
abcdefg30: Is there a reason why you don't just use `a.HasProperty("Health")` instead?

mods/ra/maps/exodus/exodus.lua
```diff
@@ -0,0 +1,399 @@
...
+	else
+		yak.Wait(DateTime.Seconds(5))
+	end
+
+	if yak.AmmoCount() > 0 then
+		yak.Stop()
+		yak.Attack(target)
+	else
+		yak.ReturnToBase()
+		yak.Resupply()
```
abcdefg30: This shouldn't be needed as pointed out by @obrakmann in https://github.com/OpenRA/OpenRA/pull/11658#discussion_r73580912.

mods/ra/maps/exodus/exodus.lua
Mailaender: I want them to snipe units first, then take on the buildings.

abcdefg30: I'm really unsure if this works that way.

obrakmann: It doesn't. You'd have to sort the list returned by `Utils.Where` first.

Mailaender:
Fixed the naive code.

abcdefg30:
This has my 👍 when https://github.com/Mailaender/OpenRA/pull/58 is accepted.

One final note though: Currently you are able to build in the evacuation area and thus are able to produce units *behind* the cell trigger. We could either expand the cell trigger or somehow try to prevent people from building there. But that's not a real blocker.

Mailaender:
Thanks for the code cleanup. I squashed them in. Removed https://github.com/Mailaender/OpenRA/pull/58/commits/c10b8c31a249200dcab0285b4670744aaed1f166 as that is decided by the script and I ditched https://github.com/Mailaender/OpenRA/pull/58/commits/74423648b8aedc16d82cbd235ea5ede43c968a60 and https://github.com/Mailaender/OpenRA/pull/58/commits/7af519b81e4066ee51a0437146c2ce7c8b312854 as that caused breakages in my tests.

abcdefg30:
Bleh. I guess I won't block this PR (and release blocker) over this and file a follow-up PR.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ac1c9c61e8fc470bddbb95b89d7868c097fe5982...463f1f2cd6f00b2963b22f2865c95f04936bc6c7)

PR #11680, opened 2016-07-20
Removed the unused jeep unload sequence

Mailaender:
We don't use the unload animations (anymore) so this workaround isn't required. Redundant to idle anyway so that sequence can be referenced when it makes a return. Might save a small amount of space on the sprite map. PS: This applies to a few other occasions as well. Updated accordingly.

### DIFF
--- a/mods/cnc/sequences/campaign.yaml
+++ b/mods/cnc/sequences/campaign.yaml
@@ -2,9 +2,6 @@ lst:
 	idle: lstnew
 		UseClassicFacingFudge: True
 		Facings: 32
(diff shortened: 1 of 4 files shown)


GraionDilach:
:+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/9d5ff92ed70af549bbb28333de7585bd89f9e36e)

Issue #9086, opened 2015-08-16
Investigate using MoonSharp instead of a custom Eluant


Mailaender:
I haven't done any experiments, but at first glance there seem to be some benefits from using @xanathar's http://www.moonsharp.org/ vs. our https://github.com/OpenRA/Eluant fork:

* Proper stacktraces with not just line numbers in the script wrapper, but the map script code.
* It is well maintained with Nuget packages available.
* We don't need to worry about system Lua anymore as it is written in C# like KopiLua.
* Seems to have been designed for sandboxing and games in the first place.

Unit158:
So, I've looked at this today and notice this so far:
- Memory management is done entirely in C#, meaning limiting the memory use will require some hacks.
- Our setup to add things to the interpreter would be massively simplified, but this would require nearly a complete re-write of how we do it now.

Unit158:
From IRC: `tl;dr: We can just replace Eluant (essentially) piece for piece with MoonSharp, We can replace our current interfaces with MoonSharp's interfaces which provide the same functionality or we can remove all references to the actual interpreters from mod and engine code` 

Which is preferred?

Phrohdoh:
I still haven't heard a reason as to *why* MoonSharp may be better other than "it lets mappers do what they want" which doesn't explain or justify anything.

LavenderMoon:
It's still under active development, static analysis support is on it's roadmap, and it removes our dependency on an external install of Lua. It also uses Unicode text, which as far as I know is what the CLR uses, so that might mean we have less marshaling cost. In fact, on it's about page, it says "The entire purpose of MoonSharp is to sacrifice Lua raw execution speed in order to gain performance on Lua/C# cross calls." Since we're not doing the heavy lifting with Lua, that sounds like the right choice to make for OpenRA. :+1:

Unit158:
Aye, the above is true, and if you search for MoonSharp in the irc logs, Matt and I have had at least one discussion why MoonSharp is preferable to Eluant.

Unit158:
Although I have a fair bit of work done, I won't be able to work on this until at least christmas, if not next summer.

Mailaender:
It is also used in professionally: https://www.patreon.com/posts/owlboy-linux-7892249?cid=3863989

Mailaender:
Rebased @Unit158's attempt https://github.com/Mailaender/OpenRA/tree/moonsharp but didn't get too far.

PunkPun:
[...]

PR #11686, opened 2016-07-22
Added API and auto-complete description generator for ZeroBrane Studio

Mailaender:
https://studio.zerobrane.com is an Open Source cross-platform Lua IDE with code completion support. This adds a generator for the API definition.

I submitted a snapshot of the last OpenRA release to  https://github.com/pkulchenko/ZeroBraneStudio/pull/670. You don't have to wait for upstream inclusion. Just copy the API definition and interpreter files to your local installation and keep going.

This is how it looks in the IDE. You get tooltips and auto-completion:
image

I am not sure if it is technically possible with our current Lua bridge, but the IDE also allows you to extend interpreters for static code analysis and live debugging.

PS: Now available as a standalone plugin that needs to be placed in `ZBS/packages/` or `~/.zbstudio/packages` https://github.com/pkulchenko/ZeroBranePackage/blob/master/openra.lua

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/ExtractZeroBraneStudioLuaAPI.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/ExtractZeroBraneStudioLuaAPI.cs
@@ -1,0 +1,152 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+using System.Reflection;
+using OpenRA.Scripting;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.UtilityCommands
+{
+	// See https://studio.zerobrane.com/doc-api-auto-complete for reference
+	class ExtractZeroBraneStudioLuaAPI : IUtilityCommand
+	{
+		string IUtilityCommand.Name { get { return "--zbstudio-lua-api"; } }
+
+		bool IUtilityCommand.ValidateArguments(string[] args)
+		{
+			return true;
+		}
+
+		[Desc("Generate ZeroBrane Studio Lua API and auto-complete descriptions.")]
+		void IUtilityCommand.Run(Utility utility, string[] args)
+		{
+			// HACK: The engine code assumes that Game.modData is set.
+			Game.ModData = utility.ModData;
+
+			Console.WriteLine("local interpreter = {");
+			Console.WriteLine("  name = \"OpenRA\",");
+			Console.WriteLine("  description = \"OpenRA map scripting Lua API\",");
+			Console.WriteLine("  api = {\"baselib\", \"openra\"},");
+			Console.WriteLine("  hasdebugger = false,");
+			Console.WriteLine("  skipcompile = true,");
+			Console.WriteLine("}");
+			Console.WriteLine();
+
+			Console.WriteLine("-- This is an automatically generated Lua API definition generated for release-{0} of OpenRA.", Game.ModData.Manifest.Metadata.Version);
+			Console.WriteLine("-- https://github.com/OpenRA/OpenRA/wiki/Utility was used with the --zbstudio-lua-api parameter.");
+			Console.WriteLine("-- See https://github.com/OpenRA/OpenRA/wiki/Lua-API for human readable documentation.");
+			Console.WriteLine();
+			Console.WriteLine("local api = {");
+
+			var tables = Game.ModData.ObjectCreator.GetTypesImplementing<ScriptGlobal>().OrderBy(t => t.Name);
+			foreach (var t in tables)
+			{
+				var name = t.GetCustomAttributes<ScriptGlobalAttribute>(true).First().Name;
+				Console.WriteLine("  " + name + " = {");
+				Console.WriteLine("    type = \"class\",");
+				Console.WriteLine("    childs = {");
+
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/UtilityCommands/ExtractZeroBraneStudioLuaAPI.cs
```diff
@@ -0,0 +1,126 @@
...
+	class ExtractZeroBraneStudioLuaAPI : IUtilityCommand
+	{
+		public string Name { get { return "--zbstudio-lua-api"; } }
+
+		public bool ValidateArguments(string[] args)
+		{
+			return true;
+		}
+
+		[Desc("Generate a ZeroBrane Studio Lua API and auto-complete description.")]
```
Phrohdoh: Seeing as you are outputting to the stdout the user must redirect the data, it would be polite to state that it should be written to a lua file.

Currently this is a guessing game (or the person has to stumble across https://github.com/pkulchenko/ZeroBranePackage/blob/226dbc946624be2a933464d23b4767adb4815d50/openra.lua).

Even though this will probably only be used by us to generate docs for release it doesn't hurt to add helpful information.

Mailaender: I added https://github.com/OpenRA/OpenRA/wiki/Map-scripting#zerobrane-studio

Phrohdoh: I don't see that being helpful to someone who only has the source (possibly working offline).

pchote: I both agree and disagree with @Phrohdoh:
* We don't mention that `--docs` or `--lua-docs` should be redirected to md files, so I don't see why this should include a special mention here.
* The zerobrane section on the map scripting page is a good start for scripters, but we need documentation (or even better, automation) for updating https://github.com/pkulchenko/ZeroBranePackage/blob/master/openra.lua  somewhere that *we* are going to read.  I expect that integration will be forgotten and/or neglected by everyone except @Mailaender otherwise.

pchote: One option would be to add a step to our travis packaging that generates a `zerobrane-completion.zip` file with the generated lua and a README that explains where to put it.

Phrohdoh: I feel the proper fix to your first point is to also mention redirection in those two command descriptions.

Mailaender: The IDE has no package management https://github.com/pkulchenko/ZeroBraneStudio/issues/528 so it can't be properly automatized. I don't want to maintain this at https://github.com/pkulchenko/ZeroBranePackage and a separate `zerobrane-completion.zip`. Also I suggest not to over-engineer a release management for such a small addon only a very small percentage of our very advanced users will use. I left documentation at https://github.com/pkulchenko/ZeroBranePackage/blob/master/openra.lua#L9 to make it easy for everyone to update it.

pchote:
I don't plan on ever using this IDE or getting deeply into map scripting, so I'm not a good fit for reviewing this.

Mailaender:
While working on https://github.com/OpenRA/OpenRA/issues/5490 I can only recommend to use an IDE. Our scripting engine is so very fragile and starting the game only to get to a syntax error and retry again is very slow and painful. This doesn't solve in-game errors not saying a lot. That is something only https://github.com/OpenRA/OpenRA/issues/9086 can improve with proper stacktraces.

I tested quite a few IDEs and ZeroBrane Studio is the most feature complete one to date which is why I stopped working on https://github.com/monodevelop/LuaBinding which would also need some kind of API description file anyway. We can keep this pull request open forever and I just need to rebase it for every release. That would be idiotic for many reasons.

abcdefg30 (approved):
The utility command seems to miss the "header":

``` lua
local interpreter = {
  name = "OpenRA",
  description = "OpenRA map scripting Lua API",
  api = {"baselib", "openra"},
  hasdebugger = false,
  skipcompile = true,
}
```

Looks good to me otherwise. 👍 / ✅

Mailaender:
Yes, because at the time of writing that was in a separate file.  https://github.com/pkulchenko/ZeroBraneStudio/pull/670 I added it for completion sake although it is static and won't change.

Mailaender:
Updated https://github.com/pkulchenko/ZeroBranePackage/pull/40 and https://software.opensuse.org/package/zerobranestudio is now available to @openSUSE users via 1-Click-Install or directly via the main repository when using Tumbleweed rolling release.

pchote:
Are there any other @OpenRA/engine-hackers who want to review, and then take over reponsibility for maintaining this?

Mailaender:
This is really low maintenance. I will keep https://github.com/pkulchenko/ZeroBranePackage/ updated myself.

reaperrr:
@Mailaender is there some prerequisite to get these tooltips? I installed ZB and the package, but I haven't seen any Tooltips or auto-complete (tested with allies01 and soviet07).

Mailaender:
Be sure to set the Lua interpreter to OpenRA via - _Project_ - _Lua Interpreter_ - [x] OpenRA

image

reaperrr:
> Be sure to set the Lua interpreter to OpenRA via - Project - Lua Interpreter - [x] OpenRA

Ah, thanks.

I might use this at some point, so I'm willing to give this a chance. 👍

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/b05d01aeaf7c366c542e9f3bb9d185607e245c98...e1b591ebffa7a07736fe70a31f0cfd54f817ba01)

Issue #8276, opened 2015-05-27
Per mod settings


Mailaender:
Required for
* https://github.com/OpenRA/OpenRA/issues/4787
* https://github.com/OpenRA/OpenRA/issues/7680

etc.

dan9550:
This could also be used to save campaign information correct?
e.g. Mark all completed missions with a star or something

rob-v:
PR created https://github.com/OpenRA/OpenRA/pull/12990

Added support for per mod settings in settings._mod_.yaml file. Currently supported options:

Player - Name, Color, LastServer (MP Direct IP Address:port)
Skirmish - Faction, Team
Multiplayer - Faction, Team
Server - all options from settings.yaml (MP Create Game name, port, map,...)

Issue #4180, opened 2013-11-29
Move BuildPaletteOrder to ChromeMetrics


Mailaender:
##

pchote:
I don't think this is relevant anymore, so closing.

Mailaender:
I opened it up due to the following line, which is still part of the code base:

https://github.com/OpenRA/OpenRA/blob/bb536ee4fcf0e5d5afd6cf2537fd92b76db059cf/OpenRA.Mods.Common/Traits/Buildable.cs#L49-L50

pchote:
Removed in #15599.

PR #11687, opened 2016-07-22
Tiedied up d2k building build menu icon order.

CH4Code:
Changed BuildPalletteOrder values for all d2k buildings. Turrets last, defenses where they fit and all other buildings adjacent to, or in line with their prerequesites. Feels much better than before.

Also uploaded seven test maps using this modification to the resource center. Those are available on the current release build. They have "(+BetterBuildmenu)" after their name.

Also removed the construction_yard prerequisite from all buildings, which was just cluttering the tooltips.
ra mod has no conyard prerequiseites as well.

### DIFF
--- a/mods/d2k/rules/structures.yaml
+++ b/mods/d2k/rules/structures.yaml
@@ -16,8 +16,7 @@
 		RemoveInstead: true
 	Buildable:
 		Queue: Building
-		Prerequisites: construction_yard
-		BuildPaletteOrder: 10
+		BuildPaletteOrder: 10 
 
 concretea:
 	Inherits: ^concrete
@@ -42,7 +41,7 @@ concreteb:
 		Value: 81
 	Buildable:
 		BuildPaletteOrder: 40
-		Prerequisites: construction_yard, upgrade.conyard
+		Prerequisites: upgrade.conyard
 
 construction_yard:
 	Inherits: ^Building
@@ -112,7 +111,6 @@ wind_trap:
 	Inherits: ^Building
 	Buildable:
 		Queue: Building
-		Prerequisites: construction_yard
 		BuildPaletteOrder: 20
 	Selectable:
 		Bounds: 64,64
@@ -150,7 +148,7 @@ wind_trap:
 barracks:
 	Inherits: ^Building
 	Buildable:
-		Prerequisites: construction_yard, wind_trap
+		Prerequisites: wind_trap
 		Queue: Building
 		BuildPaletteOrder: 50
 	Selectable:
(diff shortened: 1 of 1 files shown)


Mailaender:
Before: 
image

After:
image

Mailaender:
I guess we want https://github.com/OpenRA/OpenRA/issues/4180 together with https://github.com/OpenRA/OpenRA/issues/8276 to avoid these constant preferences based changes.

pchote:
The palette ordering should ideally reflect the tech requirements of each actor and, if possible to group production and tech structures onto common rows. I don't know the d2k tech tree off the top of my head, so I can't comment specifically on the layout here currently.

It also makes some sense to have the two turrets and walls on the same row.

CH4Code:
Here is a picture of the Techtree:
techtree_d2k_20160508_and_bb

@Mailaender: What is the crane-thing in your picture, is this for next release? What prequesites does it have? Maybe I can come up with an even better order, if this is included.

@pchote: It does exactly this. prequisites are as in line as possible , earlygame buildings on F1-F9.  I could not put turrets+wall on the same row, because the last row was only 2 entries. Also, you have a silo, too...

GraionDilach:
The crane thingie is the conyard.

CH4Code:
@GraionDilach:  Huh? So it does not appear in-game in the build menu normally, right? Because you can't build a conyard from another conyard directly (without an MCV-vehicle). In release 20160508 it doesn't appear.
FYI: Currently this icon is also used for the conyard upgrade.

Mailaender:
When you enable developer cheats such as `/enabletech` you can see them. http://strategywiki.org/wiki/OpenRA/Cheats. However that is nothing the regular player will see.

mods/d2k/rules/structures.yaml
```diff
@@ -16,8 +16,7 @@
 		RemoveInstead: true
 	Buildable:
 		Queue: Building
-		Prerequisites: construction_yard
-		BuildPaletteOrder: 10
+		BuildPaletteOrder: 10
```
GraionDilach: You have messed something up here I guess (otherwise it wouldn't mark this line added). Line ending?

CH4Code: @GraionDilach: I got some warning about CLRF being converted to LF, so it would "keep original line endings". I ignored it because im not sure what it means^^ Is there a problem?
I think i also did enter an " " to be able to save the file, because my git (+ext) at one time did not show my changes in the git (+ext) editor. If you mark the second line with your cursor, you see the additional " ". thats why.

CH4Code:
Ok, thank you.
I wasn't a fan of the crane idea anyways. :D

abcdefg30:
Could you squash the last two commits together, please?
(I don't know how familiar you are with `git`, so here's an article about `squashing`: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html )

abcdefg30:
I took the freedom and submitted a rebased and squashed version of this in #11856, so that we can hopefully get the changes in for the next playtest/release. :)

PR #11697, opened 2016-07-23
Couple "Warning: Server ports are not forwarded." with the UPnP feature

Mailaender:
I found it a bit confusing this message is printed always.


pchote:
I might be missing something, but I thought that the whole point of this feature was to notify people when their ports aren't forwarded?  That is orthogonal to UPnP.

Mailaender:
Indeed the master server response code is very much tied to port forwarding it seems.

PR #7863, opened 2015-04-04
Added a global palette effect for day/night lighting

Mailaender:
Add the following to any `map.yaml` which defaults to night:

``` yaml
Rules:
    World:
        GlobalPaletteEffect:
```

and if you are not epileptic try

``` yaml
Rules:
    World:
        GlobalPaletteEffect:
            Flash: true
```

PR #11702, opened 2016-07-24
Added Scott's Allies04 co-op multiplayer mission as "Infiltration"

Mailaender:
@ScottNZ's original mission converted from the [C# mission code found in release-20131223](https://github.com/OpenRA/OpenRA/blob/release-20131223/OpenRA.Mods.RA/Missions/Allies04Script.cs). My favorite one as it tells a good story without relying on too much text and has a nice attention to detail. The intro sequence is very hard, but it works well in co-operative settings with one player already guiding the newbie. It is also exciting (and hilarious) for spectators to watch.

Regarding features it was way ahead of it's time, but relied a lot on non-reusable hacks and suffered from missing engine features. It even had civilians walking around randomly which we just introduced in https://github.com/OpenRA/OpenRA/pull/11520 for all maps and mods. I managed to replace all expensive checks with triggers and made use of the global palette effect from https://github.com/OpenRA/OpenRA/pull/7863 to create the dawn lighting.

This should be very faithful to the original. The only behavioral change I introduced is to let the spy "hotwire" (externally capture) the truck. Afterwards all spies can enter regularly. This avoids fragile script workarounds. Capturing would destroy the spy actor and `LoadPassenger` will duplicate it, leading to an inconsistent world state.


pchote:
We used to have logic in place that lets a spy infiltrate an actor without being destroyed.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -207,6 +207,13 @@ public void Debug(string text)
 			Game.Debug(text);
 		}
 
+		[Desc("Write a message into debug.log")]
+		public void DebugLog(string text)
```
abcdefg30: Isn't this kind of redundant to the normal `Debug` (see method above)?

Mailaender: No, that uses `Game.Debug` which displays it in-game. If you want to dump a large amount of data the chat log would get spammed and you can't search it.

pchote: Isnt this redundant to the print builtin that we overrode for this purpose?

abcdefg30:
Maybe we should mention that you need to capture a supply truck. Maybe in the description with just something like `The best way to get in for our spies is to hijack a supply truck.`?

abcdefg30:
Or even add an extra objective for it.

mods/ra/maps/infiltration/rules.yaml
```diff
@@ -0,0 +1,213 @@
...
+		Default: normal
+
+MISS:
+	AutoTargetIgnore:
+	Targetable:
+		TargetTypes: Ground, C4, DetonateAttack, Structure, MissionObjective
+
+LST.Unselectable.UnloadOnly:
+	Inherits: LST
+	Buildable:
```
abcdefg30: What is this for?

abcdefg30: Seems like you forgot to add `Prerequisites: ~disabled` below.

mods/ra/maps/infiltration/rules.yaml
```diff
@@ -0,0 +1,213 @@
...
+	WithLandingCraftAnimation:
```
abcdefg30: This is already defined on `LST`.

mods/ra/maps/infiltration/rules.yaml
```diff
@@ -0,0 +1,213 @@
...
+	-Selectable:
+	RenderSprites:
+		Image: LST
+	WithFacingSpriteBody:
```
abcdefg30: This on `^Ship`.

mods/ra/maps/infiltration/rules.yaml
```diff
@@ -0,0 +1,213 @@
...
+	AutoSelectionSize:
```
abcdefg30: And this on `^SpriteActor`.

mods/ra/maps/infiltration/rules.yaml
```diff
@@ -0,0 +1,213 @@
...
+		CaptureTypes: MissionObjective
+
+DOG.Patrol:
+	Inherits: DOG
+	Buildable:
+		Prerequisites: ~disabled
+	Mobile:
+		Speed: 56
+	WithInfantryBody:
+	WithDeathAnimation:
```
abcdefg30: This and the one above are also inherited and not necessary thus.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
pchote: As discussed in IRC, please remove this and use `print()` instead.

OpenRA.Mods.Common/Scripting/Global/TriggerGlobal.cs
```diff
@@ -291,6 +291,38 @@ public void OnAllKilledOrCaptured(Actor[] actors, LuaFunction func)
 			}
 		}
 
+		[Desc("Call a function when all of the actors in a group have been killed, captured or sold. " +
+			"The callback function will be called as func().")]
+		public void OnAllKilledCapturedOrSold(Actor[] actors, LuaFunction func)
```
pchote: This is starting to get silly.  Please just bind the individual triggers instead, or use a variation of `OnRemovedFromWorld`.

OpenRA.Mods.Common/Traits/Render/RenderDebugState.cs
```diff
@@ -60,7 +60,7 @@ public void AddedToWorld(Actor self)
 
 		public void OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
 		{
-			color = GetColor();
+			color = newOwner.Color.RGB;
```
pchote: This looks bogus: you're leaking state that could be considered an exploit.

Mailaender:
Fixed all the remarks.

OpenRA.Mods.Common/Scripting/Properties/GeneralProperties.cs
```diff
@@ -59,6 +59,9 @@ public Player Owner
 
 			set
 			{
+				if (value == null || !(value is Player))
```
pchote: Do you have a repro case of this being called with `!(value is Player)`?  That shouldn't be possible.

OpenRA.Mods.Common/Traits/Render/RenderDebugState.cs
```diff
@@ -65,7 +65,7 @@ public void OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
 
 		Color GetColor()
 		{
-			return self.EffectiveOwner != null ? self.EffectiveOwner.Owner.Color.RGB : self.Owner.Color.RGB;
+			return self.EffectiveOwner != null && self.IsInWorld ? self.EffectiveOwner.Owner.Color.RGB : self.Owner.Color.RGB;
```
pchote: What are you actually trying to avoid here?  `self.EffectiveOwner.Owner.Color.RGB` should have nothing to do with whether the actor is in the world or not.

OpenRA.Mods.Common/Scripting/Properties/GeneralProperties.cs
Mailaender: This is a wild guess to be honest. Even the null check. You suggested on IRC to add this. I can't really prove this fixes anything and I believe the Lua bridge will crash on a lower level part if you throw wrong parameters on it "can't convert stupid scripter typo without naming even the variable nil to LuaValue in line number 30 in sandbox.lua". I also tried to improve the stacktrace a bit in those situations as it is _really_ annyoing, but failed. Will remove this and hope for @Unit158 to finish #9086.

OpenRA.Mods.Common/Traits/Render/RenderDebugState.cs
Mailaender: This is a pretty tough one. The stacktrace doesn't tell much except that `OnOwnerChanged` triggers `GetColor()` not even with line numbers. Somehow querying `EffectiveOwner` for actors outside the world or newly added to the world crashes badly. Maybe caching is responsible https://github.com/OpenRA/Mailaender/OpenRA/blob/infiltration-mission/OpenRA.Game/Actor.cs#L116 Will do some experiments to solve this elsewhere, but it will probably get even messier.

Mailaender: Apparently nothing is null (`t`, `this`, `oldOwner`, `newOwner`) yet it crashes with a `NullReferenceException`:
image

Also handing over `EffectiveOwner` doesn't help. Nor does running the `INotifyOwnerChange after another `AddFrameEndTask`.

Mailaender: Will just remove the offending trait as without `RenderDebugState` this isn't triggered.

obrakmann:
Here you'll want to set the `Cooperative` flag on `MissionObjectives` as well, for the same reason as in Allies03.

Mailaender:
I had trouble with the `Cooperative` flag on https://github.com/OpenRA/OpenRA/pull/11679 as it doesn't seem to work well with shared/duplicated objectives, which is the case here as well.

obrakmann:
What problems exactly? It should just work.

OpenRA.Mods.Common/Scripting/Properties/SellableProperties.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+	[ScriptPropertyGroup("General")]
+	public class SellableProperties : ScriptActorProperties, Requires<SellableInfo>
+	{
+		readonly Sellable sellable;
+
+		public SellableProperties(ScriptContext context, Actor self)
+			: base(context, self)
+		{
+			sellable = self.Trait<Sellable>();
```
obrakmann: You don't really need to cache this, it will only be used once for every actor.

OpenRA.Mods.Common/Scripting/ScriptTriggers.cs
```diff
@@ -364,6 +364,46 @@ public void RemovedFromWorld(Actor self)
 			OnRemovedInternal(self);
 		}
 
+		public void Selling(Actor self)
```
obrakmann: This file would really benefit from using explicitly implemented interfaces.

obrakmann:
Looks otherwise fine so far. The mission is fucking annoying in the beginning, though, with all the dogs around and you having no idea where to go. I eventually looked at the map in the editor and then knew where I had to go, but still never got there due to the dogs. How about removing some on easier skill levels?

abcdefg30:
> How about removing some on easier skill levels?

We could also add cameras, like in `allies05a` on easier difficulties.

OpenRA.Mods.Common/Scripting/Properties/SellableProperties.cs
abcdefg30: What if we sell many actors at once? Wouldn't that create a tiny lag without caching?
What I actually want to says, now that we already cache it, do we really want to revert that back?

pchote: Arguing over micro-optimizations is a bit silly, but I think its fair to say that an actor will be sold on average zero times (or at least much closer to zero than one).  Caching it here means doing this lookup for *all* buildings, but doing it in Sell means doing it only on the buildings that you sell.

abcdefg30: Ah well. Then nvm my comment.

Mailaender:
Updated the Lua API.

> How about removing some on easier skill levels?

I won't oppose this. but I prefer to keep re-balancing separate from this.

mods/ra/maps/infiltration/infiltration.lua
```diff
@@ -0,0 +1,343 @@
...
+			end)
+
+			Trigger.OnPlayerLost(player, function()
+				Media.PlaySpeechNotification(player, "MissionFailed")
+			end)
+		end
+	end)
+
+	InsertSpies()
+	Utils.Do(TownAttackers, IdleHunt)
```
obrakmann: Ah, now I get it. These town attackers are the group that instantly jumps you when you get your MCV. The reason is that they're idle and constantly looking for enemy units, which they don't find as long your spy is disguised. Can you find some way around that, please? Clearing their idle trigger when the lab is infiltrated and having them do something else or something like that?

Mailaender: Yes, I forgot to implement something like https://github.com/OpenRA/OpenRA/blob/1d9681a1d2769a9c736b79ba612204b979bf9d13/OpenRA.Mods.RA/Missions/Allies04Script.cs#L160

> Also, are they meant to destroy the hospital?

In the original, yes.

Mailaender:
The town attackers will now stop the right once they succeeded.

obrakmann:
Thanks. Looks ok to me now. :+1:

mods/ra/maps/infiltration/infiltration.lua
```diff
@@ -0,0 +1,355 @@
...
+	Utils.Do(Patrol2, function(patrol3) NewPatrol(patrol3, PatrolPoints3[1], PatrolPoints3) end)
+	Utils.Do(Patrol2, function(patrol4) NewPatrol(patrol4, PatrolPoints4[1], PatrolPoints4) end)
+
+	if Difficulty == "hard" then
+		Utils.Do(Patrol3, function(patrol5) NewPatrol(patrol5, PatrolPoints2[1], PatrolPoints2) end)
+	end
+
+	local checkpoint = { BaseGuardTruckPos.Location }
+	Trigger.OnEnteredFootprint(checkpoint, function()
+		BaseGuard.ScriptedMove(BaseGuardMovePos.Location)
```
abcdefg30: You should probably remove this trigger once it fired and at least add an `IsDead` check for `BaseGuard`.

abcdefg30: I'll just prepare another fixup PR where I fix this and minor other things I noticed.

obrakmann:
BTW, you might also want to set the subs' stance to Defend or AttackAnything. Currently it's still the AI default of ReturnFire.

mods/ra/maps/infiltration/infiltration.lua
Mailaender: Is it even possible to exit the vehicle and shoot the guard? Dogs and tesla coils everywhere.

Mailaender:
Set it to `Defend` to match https://github.com/OpenRA/OpenRA/blob/release-20131223/OpenRA.Mods.RA/Missions/Allies04Script.cs#L352

mods/ra/maps/infiltration/infiltration.lua
abcdefg30: You can kill him later in the game (with tanks and other units), as the trigger isn't removed.

abcdefg30:
I left my remarks at https://github.com/Mailaender/OpenRA/pull/59.

Mailaender:
Thanks. Squashed them in.

abcdefg30:
Looks good to me now (after travis is fixed). 👍 / ✅

abcdefg30:
https://github.com/Mailaender/OpenRA/pull/60 fixes travis.

Mailaender:
And another crash fixed https://github.com/Mailaender/OpenRA/pull/61. Thanks.

mods/ra/maps/infiltration/infiltration.lua
```diff
@@ -0,0 +1,380 @@
...
+	-- This includes killed, captured (actor is temporarily removed) and sold.
+	Trigger.OnAllRemovedFromWorld(sovietbuildings, function()
+		Utils.Do(humans, function(player)
+			player.MarkCompletedObjective(destroyBase)
+		end)
+	end)
+
+	Utils.Do(sovietbuildings, function(sovietbuilding)
+		Trigger.OnDamaged(sovietbuilding, function(building)
+			if building.Health < building.MaxHealth * 3/4 then
```
obrakmann: This needs an owner check.

mods/ra/maps/infiltration/infiltration.lua
```diff
@@ -0,0 +1,380 @@
...
+	end)
+end
+
+labSecured = false
+CheckLabSecured = function()
+	if not reinforcementsHaveArrived or labSecured then
+		return
+	end
+
+	if player1.HasNoRequiredUnits() and player2 and player2.HasNoRequiredUnits() then
```
obrakmann: This will not work as intended in single-player mode.

mods/ra/maps/infiltration/infiltration.lua
```diff
@@ -0,0 +1,380 @@
...
+			end
+		end)
+	end
+end
+
+reinforcementsHaveArrived = false
+LabInfiltrated = function()
+	Utils.Do(humans, function(player)
+		if player then
+			secureLab = player.AddPrimaryObjective("Secure the laboratory before the soviets can react.")
```
obrakmann: This objective is confusing. I wasn't sure what I was meant to do (capture it?, or infiltrate it again?) until I read the script. How about "Destroy all units guarding the lab."?

obrakmann:
:+1: after those things have been addressed

mods/ra/maps/infiltration/infiltration.lua
Mailaender: What do you suggest instead?

obrakmann: Make it `if player1.HasNoRequiredUnits() or (player2 and player2.HasNoRequiredUnits()) then`

It won't happen anyway, since the AI doesn't run attacks by itself.

Mailaender: I think my intention was to keep the game running when only one player dies, but indeed that doesn't matter. Player are fighting the clock here.

Mailaender:
Updated.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f71c45600e838b4410b378c77aab13ab8b865b83)

OpenRA.Mods.Common/Traits/Render/RenderDebugState.cs
Mailaender: This heisenbug made a re-appearance in https://github.com/OpenRA/OpenRA/issues/11996 as well.

PR #11704, opened 2016-07-24
Remove CAMERA.Large from Evacuation and Survival01

abcdefg30:
Resolves https://github.com/OpenRA/OpenRA/pull/11316#discussion_r71973766.

### DIFF
--- a/mods/ra/maps/evacuation/rules.yaml
+++ b/mods/ra/maps/evacuation/rules.yaml
@@ -200,11 +200,6 @@ MSUB:
 	Buildable:
 		Prerequisites: ~disabled
 
-CAMERA.Large:
-	Inherits: CAMERA
-	RevealsShroud:
-		Range: 1000c0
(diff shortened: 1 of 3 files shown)


Mailaender:
Well spotted. :+1: `Range: 1000` is just one cell afaik and a leftover from an upgrade rule that didn't trigger possibly because an earlier upgrade rule already touched it or the engine used was too new. The time gap for resolving https://github.com/OpenRA/OpenRA/issues/5490 was quite huge for most of the maps. Also the currently used `Hunt` activity that is used instead of manual targeting doesn't require vision.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/6b266c4f751ae5b73a86a5a35847483e9380cf8b)

PR #11715, opened 2016-07-25
Improved the Red Alert dog behavior

### LINKED ISSUE #7506: Dog issues
I was looking into the dog issues, starting with https://github.com/OpenRA/OpenRA/issues/7077.

I implemented a simple algorithm to check wether the dog has a clear line of sight (meaning no obstructions). This worked but introduced another issue, that is ... the pathplanning is not optimal because the dog will run to the wall, stop and needs to reroute. I think that is not really wanted. So maybe we could introduce a new option that some weapons (such as the DogJaw) can only be fired with no obstructions with the target, modify the path planner to keep this in mind and build in a check on the leap to prevent jumping over blocked tiles. 

Any ideas about this? Other approach? There are also some other bugs with the dogs such as https://github.com/OpenRA/OpenRA/issues/6798 that may could be resolved by a solid approach.

screenshot from 2015-02-20 01 27 40

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/issues/7506. I redid this from scratch and tried to match RA95 behavior taking https://youtu.be/4ZyAlXOY0aM?t=2m for reference still avoiding legacy limitations.

Instead of leaping onto a targetable actor and killing everything in that cell, I separate leap movement and attack execution. This means the path finder now makes sure that an attack can be executed so jumping over cliffs and other glitches are gone.

The insta-kill is also not a hard-coded hack, but using the regular weapon system. This means we can also armor soldiers accordingly so they only take damage from a hit for a more fine grained balancing. For now the weapon is a high speed bullet like all the other melee attacks, but we can change this into a close combat "projectile" later on to avoid the overhead.

I also tried to adhere to our modular https://github.com/OpenRA/OpenRA/wiki/Actor-Upgrades system instead of putting everything into a monolithic one-purpose trait. The `UpgradeOnTarget` trait can be used to recreate http://cnc.wikia.com/wiki/Terrorist_(Generals) style units and I guess creative modders will find a lot of uses for `UpgradeOnAttack` I didn't even think of yet.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnAttack.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnAttack.cs
@@ -1,0 +1,59 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Trigger upgrades during Attack trait activities.")]
+	public class UpgradeOnAttackInfo : ITraitInfo, Requires<UpgradeManagerInfo>
+	{
+		[Desc("Lasts until the attack is executed.")]
+		[UpgradeGrantedReference] public readonly string[] PreparingAttackUpgrades = { };
+
+		[Desc("Timed upgrade for the duration of the fire delay.")]
+		[UpgradeGrantedReference] public readonly string[] AttackingUpgrades = { };
+
+		[Desc("How long the timed attacking upgrade should last.")]
+		public readonly int AttackingDelay = 0;
+
+		public object Create(ActorInitializer init) { return new UpgradeOnAttack(init, this); }
+	}
+
+	public class UpgradeOnAttack : INotifyAttack
+	{
+		readonly Actor self;
+		readonly UpgradeOnAttackInfo info;
+		readonly UpgradeManager manager;
+
+		public UpgradeOnAttack(ActorInitializer init, UpgradeOnAttackInfo info)
+		{
+			self = init.Self;
+			this.info = info;
+			manager = self.Trait<UpgradeManager>();
+		}
+
+		void INotifyAttack.PreparingAttack(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			foreach (var up in info.PreparingAttackUpgrades)
+				manager.GrantUpgrade(self, up, this);
+		}
+
+		void INotifyAttack.Attacking(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			foreach (var up in info.PreparingAttackUpgrades)
+				manager.RevokeUpgrade(self, up, this);
+
+			foreach (var up in info.AttackingUpgrades)
+				manager.GrantTimedUpgrade(self, up, info.AttackingDelay, this);
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTarget.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTarget.cs
@@ -1,0 +1,61 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Trigger upgrades when an actor is targeted.")]
+	public class UpgradeOnTargetInfo : ITraitInfo, Requires<UpgradeManagerInfo>
+	{
+		[UpgradeGrantedReference] public readonly string[] Upgrades = { };
+
+		public object Create(ActorInitializer init) { return new UpgradeOnTarget(init, this); }
+	}
+
+	public class UpgradeOnTarget : ITick
+	{
+		readonly Actor self;
+		readonly UpgradeOnTargetInfo info;
+		readonly UpgradeManager manager;
+
+		bool granted;
+
+		public UpgradeOnTarget(ActorInitializer init, UpgradeOnTargetInfo info)
+		{
+			self = init.Self;
+			this.info = info;
+			manager = self.Trait<UpgradeManager>();
+		}
+
+		public void Tick(Actor self)
+		{
+			var activity = self.GetCurrentActivity();
+			var wantsGranted = activity != null && activity.GetTargets(self).Any(t => t.Type == TargetType.Actor);
+			if (wantsGranted && !granted)
+			{
+				foreach (var up in info.Upgrades)
+					manager.GrantUpgrade(self, up, this);
+
+				granted = true;
+			}
+			else if (!wantsGranted && granted)
+			{
+				foreach (var up in info.Upgrades)
+					manager.RevokeUpgrade(self, up, this);
+
+				granted = false;
+			}
+		}
+	}
+}
--- a/OpenRA.Mods.RA/Activities/Leap.cs
+++ b/OpenRA.Mods.RA/Activities/Leap.cs
@@ -9,70 +9,54 @@
  */
 #endregion
 
-using System;
-using System.Linq;
+using System.Collections.Generic;
 using OpenRA.Activities;
-using OpenRA.GameRules;
 using OpenRA.Mods.Common.Traits;
-using OpenRA.Mods.Common.Traits.Render;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.RA.Activities
 {
-	class Leap : Activity
+	public class Leap : Activity
 	{
 		readonly Mobile mobile;
-		readonly WeaponInfo weapon;
+		readonly WPos origin, destination;
 		readonly int length;
 
-		WPos from;
-		WPos to;
-		int ticks;
-		WAngle angle;
+		int ticks = 0;
 
-		public Leap(Actor self, Actor target, WeaponInfo weapon, WDist speed, WAngle angle)
+		/// <summary> Visible move that changes the position in the world. </summary>
+		public Leap(Actor self, WPos origin, WPos destination, int length, Mobile mobile)
 		{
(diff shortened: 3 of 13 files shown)


OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnAttack.cs
```diff
@@ -0,0 +1,56 @@
...
+		{
+			self = init.Self;
+			this.info = info;
+			manager = self.Trait<UpgradeManager>();
+		}
+
+		void INotifyAttack.PreparingAttack(Actor self, Target target, Armament a, Barrel barrel)
+		{
+			foreach (var up in info.PreparingAttackUpgrades)
+				manager.GrantUpgrade(self, up, this);
```
pchote: Is it possible to call PreparingAttack without a corresponding Attacking? Need to avoid mismatched calls.

OpenRA.Mods.RA/Traits/Attack/AttackLeap.cs
```diff
@@ -9,49 +9,44 @@
...
+using OpenRA.Activities;
 using OpenRA.Mods.Common.Traits;
 using OpenRA.Mods.RA.Activities;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.RA.Traits
 {
-	[Desc("Dogs use this attack model.")]
-	class AttackLeapInfo : AttackFrontalInfo
+	[Desc("Move onto the attack then execute the attack.")]
```
Mailaender: Typo: `target`

mods/ra/weapons/other.yaml
```diff
@@ -185,9 +185,12 @@ TTankZap:
...
 	ValidTargets: Infantry
-	ReloadDelay: 10
-	Range: 3c0
+	ReloadDelay: 100
+	FireDelay: 200
 	Report: dogg5p.aud
+	Range: 1c512
+	Projectile: Bullet
+		Speed: 1c682
 	Warhead@1Dam: SpreadDamage
```
pchote: Dogs have small mouths, so spread isn't a great model here.  Someone needs to bite the bullet and add a proper direct-damage warhead (bypassing spread and positions altogether), would you like to volunteer?

Mailaender:
Small update with feedback from yesterdays IRC chat:
* It should now not crash when the dogs are not in the world.
* The dogs are immobile during the eating step.

MustaphaTR:
> It should now not crash when the dogs are not in the world.

No, it still does.

Also 1c512 as range is too short. They can't always attack from some ways. In original game they have 2.2 range.

Mailaender:
Can you post a stacktrace and a way to reproduce the crash?

MustaphaTR:
I killed an enemy dog with my dog.

Mailaender:
I updated this again:
* Fixed the crash when 2 dogs attacked each other simultaneously.
* Added a direct-damage warhead and a melee projectile.

GraionDilach:
It would be cool if `UpgradeOnAttack` would have some sort of `Armament` filtering, say with a `Name` field, but I can understand if that gets dismissed.

Mailaender:
Yes, that should be easy. Also a target type filter for `UpgradeOnTarget` would be possible, but I don't have a test case in this pull request as dogs only have one suitable target type and only one armament so this is out of scope and the pull request already touches a lot of things and will probably have a very slow review process. I suggest you create followup pull request based on this on your own.

mods/ra/weapons/other.yaml
```diff
@@ -186,10 +186,10 @@ TTankZap:
 DogJaw:
 	ValidTargets: Infantry
 	ReloadDelay: 10
-	Range: 3c0
+	Range: 2c256
```
MustaphaTR: That's `2.25` not `2.2` but i'm not sure if it really matters or not.

GraionDilach:
Alright, I have no changes requested. Code looks good otherwise, but I couldn't test this and I'm gone for camping for a week from tomorrow onwards, so :+0.5: from here.

mods/ra/weapons/other.yaml
Mailaender: I don't really know how the original `RULES.INI` values have to be converted to match the OpenRA distances. Based it on a rough visual comparison. Also we probably want to rebalance this properly. I never saw someone on @cncnet actually train dogs.

Biofreak1987:
the jumpattack and eating animations still seem to have some problems:
- jumpattack animations never play
- eating animations only play "sometimes" when attacking other dogs

pchote:
This doesn't work at all for me... the dogs will run up next to my infantry and then stop.

OpenRA.Mods.RA/Traits/Attack/AttackLeap.cs
```diff
@@ -9,49 +9,47 @@
...
 		{
-			if (target.Type != TargetType.Actor || !CanAttack(self, target))
-				return;
-
-			var a = ChooseArmamentsForTarget(target, true).FirstOrDefault();
-			if (a == null)
-				return;
-
-			if (!target.IsInRange(self.CenterPosition, a.MaxRange()))
+			if (IsAttacking || !mobile.CanMoveFreelyInto(self.World.Map.CellContaining(target.CenterPosition)))
```
pchote: This should be moving to the target infantry's sub-cell rather than the center of the cell.

OpenRA.Mods.RA/Activities/Leap.cs
```diff
@@ -9,70 +9,53 @@
...
+			if (++ticks >= length)
 			{
-				mobile.SetLocation(mobile.ToCell, mobile.ToSubCell, mobile.ToCell, mobile.ToSubCell);
-				mobile.FinishedMoving(self);
 				mobile.IsMoving = false;
-
-				self.World.ActorMap.GetActorsAt(mobile.ToCell, mobile.ToSubCell)
-					.Except(new[] { self }).Where(t => weapon.IsValidAgainst(t, self))
-					.Do(t => t.Kill(self));
+				mobile.SetPosition(self, pos);
```
pchote: This ignores the sub-cell positioning, which I expect will look wrong.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnAttack.cs
pchote: This isn't possible, except when the actor is killed in the middle.  But the actor is then dead, so we don't care.

mods/ra/weapons/other.yaml
WolfGaming: I can't believe we still don't have a direct-damage warhead...

GraionDilach: The problem is that only the `*Zap`s and `Missile` tracks the target, (and `Missile` will never hit it due to CloseEnough). Even with a direct-damage warhead present, it will be still useless because all projectiles will miss it - some doesn't even use `GuidedTarget` to begin with!

Mailaender: We now do have it. https://github.com/OpenRA/OpenRA/pull/11715/commits/979b328e6e931c22d06480ea3e73b887cd6ef231

Mailaender:
I updated this with sub-cell positioning in mind although I thought `WPos` already takes care about it even more precisely. Will remove https://github.com/OpenRA/OpenRA/pull/11715/commits/979b328e6e931c22d06480ea3e73b887cd6ef231 as scope creep if problems persists.

OpenRA.Mods.Common/Projectiles/MeleeAttack.cs
```diff
@@ -0,0 +1,50 @@
...
+#endregion
+
+using System.Collections.Generic;
+using OpenRA.GameRules;
+using OpenRA.Graphics;
+
+namespace OpenRA.Mods.Common.Projectiles
+{
+	[Desc("Simple invisible direct on target actor fake projectile.")]
+	public class MeleeAttackInfo : IProjectileInfo
```
GraionDilach: I'd rename this to InstantHitTarget or something along the line because that's what it is and can be used for other things besides melee.

Phrohdoh: 👍 Due to melee implying a very small distance.

Mailaender:
Gave the projectile a more generic name as suggested.

I also added a new leap upgrade and made the attack upgrade delay configurable at trait level. This makes balancing more independent from how the artwork works and resolves all regressions introduced during review compared to the initial submission.

abcdefg30:
Needs a rebase.

Mailaender:
Rebased.

pchote:
Infantry are killed long-before the dog reaches them: https://www.dropbox.com/s/m96wjtb6isznhb2/dog.mov?dl=0

It really puts me off reviewing PRs when, over multiple PR iterations, I find merge-blocking bugs within seconds of testing them.

Mailaender:
That video gives me error 500. It is a bit sad that I hadn't saved the initial revision of the pull request before the review requests for more features and complexity began.

GraionDilach:
The traits seem fine at first glance tbh, so I could try giving you a helping hand at the yaml setup, if you'd like to - prolly only on Wednesday though.

Mailaender:
Yes, that would be greatly appreciated. I probably don't have an eye for the details and the timing.

pchote:
I guess we don't want to match the RA95 behaviour *too* well:

https://www.reddit.com/r/IAmA/comments/19epje/i_am_joe_bostic_22_year_game_developer_11_years/c8o39zv

GraionDilach:
Haven't found time, but I guess the blinking frame comes in due to no upgrade is applied during PreparingAttack - granting rip should fix the issue I guess.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnTarget.cs
```diff
@@ -0,0 +1,61 @@
...
+		bool granted;
+
+		public UpgradeOnTarget(ActorInitializer init, UpgradeOnTargetInfo info)
+		{
+			self = init.Self;
+			this.info = info;
+			manager = self.Trait<UpgradeManager>();
+		}
+
+		public void Tick(Actor self)
```
reaperrr: `void ITick.Tick`, please.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnAttack.cs
```diff
@@ -0,0 +1,59 @@
...
+	public class UpgradeOnAttackInfo : ITraitInfo, Requires<UpgradeManagerInfo>
+	{
+		[Desc("Lasts until the attack is executed.")]
+		[UpgradeGrantedReference] public readonly string[] PreparingAttackUpgrades = { };
+
+		[Desc("Timed upgrade for the duration of the fire delay.")]
+		[UpgradeGrantedReference] public readonly string[] AttackingUpgrades = { };
+
+		[Desc("How long the timed attacking upgrade should last.")]
+		public readonly int AttackingDelay = 0;
```
reaperrr: I think in this case `AttackingDuration` would be better. `Delay` is misleading, in my opinion.

reaperrr:
Needs rebase.

I'll try to review this sometime next weekend.

pchote:

pchote:
This has been superseded by #13513, so removing from the milestone.

PR #11634, opened 2016-07-11
Fixup DamagedByTerrain.

GraionDilach:
Depends on #11513.

Resolves #10091 and probably fixes #11383.

Closes #11720 as well.

PR #11720, opened 2016-07-26
Silence unused field warning in DamagedByTerrainInfo

RoosterDragon:
This produces a warning in VS2015 Update 3. Explicit assignment silences the warning.


GraionDilach:
Another reason for #11634.

PR #11721, opened 2016-07-26
Fix desync when removing smudges

### LINKED ISSUE #11660: Replays on bleed crash with an out-of-sync exception

obrakmann:
The cause of the crash was that the `RemoveSmudge` method only invokes `SharedRandom` in the else-branch of an if condition that is only taken when the `RenderPlayer` has fog visibility on the location (locations the `RenderPlayer` can see get removed from `dirty` in `TickRender`).

To add insult to injury, the `Type` field does not even need to get set since [we are only using the `Sprite` field's null value as a marker to be picked up by the `TickRender` method](https://github.com/OpenRA/OpenRA/blob/e13fa2d474a963d4148ab1db7407fb84b0f90016/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs#L189). [The `Type` field is only ever used twice in `AddSmudge`](https://github.com/OpenRA/OpenRA/blob/e13fa2d474a963d4148ab1db7407fb84b0f90016/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs#L155-L159), in a branch of an if-condition [that will not be taken when the `Sprite` field is null](https://github.com/OpenRA/OpenRA/blob/e13fa2d474a963d4148ab1db7407fb84b0f90016/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs#L145), which we set explicitly. The same holds true for the `Depth` field.

Fixes #11660.

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
+++ b/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
@@ -165,18 +165,10 @@ namespace OpenRA.Mods.Common.Traits
 
 		public void RemoveSmudge(CPos loc)
 		{
-			if (dirty.ContainsKey(loc))
-			{
-				var tile = dirty[loc];
-				tile.Depth = 0;
-				tile.Sprite = null;
-				dirty[loc] = tile;
-			}
-			else
-			{
-				var st = smudges.Keys.Random(world.SharedRandom);
-				dirty[loc] = new Smudge { Type = st, Depth = 0, Sprite = null };
-			}
+			var tile = dirty.ContainsKey(loc) ? dirty[loc] : new Smudge();
+
+			tile.Sprite = null;
+			dirty[loc] = tile;
 		}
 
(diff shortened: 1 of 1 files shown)


Mailaender:
Code looks good and the explanation is sound. :+1:  Pinging @evgeniysergeev.

evgeniysergeev:
checked. works as expected

Mailaender:
Kudos to @obrakmann for investigating this critical desync!

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/12a7a48f6b89bf8916a6260719a80b48bde4bfb6...5b25a08507868fbf9b07f7ad3dfb13dff469a204)

PR #11571, opened 2016-07-03
Add command to generate settings docs

Xinayder:
Fixes #2858

PR #11722, opened 2016-07-26
Added the auto-generated settings documentation deployment

### LINKED ISSUE #2858: Automatically update default settings documentation
Do the same with https://github.com/OpenRA/OpenRA/wiki/Settings as #2808 does with https://github.com/OpenRA/OpenRA/wiki/Traits
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/136903-automatically-update-default-settings-documentation?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/11571 to finally fix https://github.com/OpenRA/OpenRA/issues/2858.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,9 +14,6 @@
 # to check the official mod dlls for StyleCop violations, run:
 #   make check
 #
-# to generate documentation aimed at modders, run:
-#   make docs
-#
 # to install, run:
 #   make [prefix=/foo] [bindir=/bar/bin] install
 #
@@ -494,9 +491,6 @@ help:
 	@echo
 	@echo 'to check the official mods for erroneous yaml files, run:'
 	@echo '  make test'
-	@echo
-	@echo 'to generate documentation aimed at modders, run:'
-	@echo '  make docs'
 	@echo
 	@echo 'to install, run:'
 	@echo '  make [prefix=/foo] [bindir=/bar/bin] install'
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,7 @@ OpenRA.Launcher.Mac/OpenRA.xcodeproj/*.mode1v3
 # auto-generated documentation
 DOCUMENTATION.md
 Lua-API.md
+Settings.md
 *.html
 openra.6
 
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -97,7 +97,10 @@ namespace OpenRA
 		public bool IgnoreVersionMismatch = false;
 
 		public bool StrictActivityChecking = false;
+
+		[Desc("Allow the collection of anonymous data such as Operating System, .NET runtime, OpenGL version and language settings.")]
 		public bool SendSystemInformation = true;
+
 		public int SystemInformationVersionPrompt = 0;
 		public string UUID = System.Guid.NewGuid().ToString();
 	}
--- a/OpenRA.Game/Settings.cs
+++ b/OpenRA.Game/Settings.cs
@@ -86,10 +86,16 @@ namespace OpenRA
 		public bool LuaDebug = false;
 		public bool PerfText = false;
 		public bool PerfGraph = false;
+
+		[Desc("Amount of time required for triggering perf.log output.")]
 		public float LongTickThresholdMs = 1;
+
 		public bool SanityCheckUnsyncedCode = false;
 		public int Samples = 25;
+
+		[Desc("Show incompatible games in server browser.")]
 		public bool IgnoreVersionMismatch = false;
+
 		public bool StrictActivityChecking = false;
 		public bool SendSystemInformation = true;
 		public int SystemInformationVersionPrompt = 0;
@@ -147,6 +153,7 @@ namespace OpenRA
 
 	public class PlayerSettings
 	{
+		[Desc("Sets the player nickname for in-game and IRC chat.")]
 		public string Name = "Newbie";
 		public HSLColor Color = new HSLColor(75, 255, 180);
 		public string LastServer = "localhost:1234";
--- a/OpenRA.Mods.Common/UtilityCommands/ExtractSettingsDocsCommand.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/ExtractSettingsDocsCommand.cs
@@ -74,7 +74,7 @@ namespace OpenRA.Mods.Common.UtilityCommands
 						Console.WriteLine("```yaml");
 						Console.WriteLine("{0}: ", section.Key);
 						Console.WriteLine("\t{0}: {1}", field.Name, value);
-						Console.WriteLine("```  ");
+						Console.WriteLine("```");
 					}
 					else
 						Console.WriteLine();
--- a/OpenRA.Mods.Common/UtilityCommands/ExtractSettingsDocsCommand.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/ExtractSettingsDocsCommand.cs
@@ -28,7 +28,7 @@ namespace OpenRA.Mods.Common.UtilityCommands
 		{
 			Game.ModData = utility.ModData;
 			Console.WriteLine(
-				"This documentation is aimed at server administrators. It displays all settings with default values and description. " +
+				"This documentation displays annotated settings with default values and description. " +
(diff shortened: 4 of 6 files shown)


packaging/windows/OpenRA.nsi
```diff
@@ -201,6 +202,7 @@ Function ${UN}Clean
 	Delete $INSTDIR\CHANGELOG.html
 	Delete $INSTDIR\CONTRIBUTING.html
 	Delete $INSTDIR\DOCUMENTATION.html
+	Delete $INSTDIR\Settings.html
```
obrakmann: While you're at it, could you add the Lua-API.html file here as well? It seems to be missing.

packaging/package-all.sh
```diff
@@ -29,12 +29,13 @@ markdown README.md > README.html
...
 markdown DOCUMENTATION.md > DOCUMENTATION.html
 markdown Lua-API.md > Lua-API.html
+markdown Settings.md > Settings.html
 
 # List of files that are packaged on all platforms
 FILES=('OpenRA.Game.exe' 'OpenRA.Game.exe.config' 'OpenRA.Utility.exe' 'OpenRA.Server.exe' 
 'OpenRA.Platforms.Default.dll' \
 'lua' 'glsl' 'mods/common' 'mods/ra' 'mods/cnc' 'mods/d2k' 'mods/modchooser' \
-'AUTHORS' 'COPYING' 'README.html' 'CONTRIBUTING.html' 'DOCUMENTATION.html' 'CHANGELOG.html' \
+'AUTHORS' 'COPYING' 'README.html' 'CONTRIBUTING.html' 'DOCUMENTATION.html' 'CHANGELOG.html' ' Settings.html' \
```
obrakmann: The Lua-API.html file is missing here as well.

obrakmann:
Looks otherwise ok, :+1:

Mailaender:
Updated.

packaging/windows/OpenRA.nsi
```diff
@@ -96,6 +96,8 @@ Section "Game" GAME
 	File "${SRCDIR}\CHANGELOG.html"
 	File "${SRCDIR}\CONTRIBUTING.html"
 	File "${SRCDIR}\DOCUMENTATION.html"
+	File "${SRCDIR}\Lua-API.html"
```
pchote: Please rename these to match the all-caps names of the other html files.

Makefile
```diff
@@ -370,6 +370,7 @@ version: mods/ra/mod.yaml mods/cnc/mod.yaml mods/d2k/mod.yaml mods/ts/mod.yaml m
 docs: utility mods version
 	@mono --debug OpenRA.Utility.exe all --docs > DOCUMENTATION.md
```
pchote: If we're going to be shipping multiple API files with the installer then this should be renamed to TRAIT-API.md.

pchote:
I personally think that the hassle of maintaining the continuously-breaking md -> html infrastructure far outweighs the benefits of shipping trait/lua documentation files that I doubt anybody has ever read.  I'd much prefer we remove the misnamed DOCUMENTATION.html from the installers instead of shipping even more files that people won't read.

Makefile
Mailaender: It is not really an application programming interface. I will rename it to match the wiki name instead.

Mailaender:
Your request of removing the MarkDown to HTML "infrastructure" due to volatility comes many months too late. All problems are solved and they were rather introduced by Windows package management being still in its infancy. Removing all offline documentation is definitely not going to happen.

Makefile
```diff
@@ -368,8 +368,9 @@ version: mods/ra/mod.yaml mods/cnc/mod.yaml mods/d2k/mod.yaml mods/ts/mod.yaml m
 	done
 
 docs: utility mods version
-	@mono --debug OpenRA.Utility.exe all --docs > DOCUMENTATION.md
-	@mono --debug OpenRA.Utility.exe all --lua-docs > Lua-API.md
+	@mono --debug OpenRA.Utility.exe all --docs > TRAITS.md
+	@mono --debug OpenRA.Utility.exe all --lua-docs > LUA-API.md
+	@mono --debug OpenRA.Utility.exe all --settings-docs > Settings.md
```
pchote: This is still inconsistent with the rest of the filenames.

pchote: Having all three pages in regular case would also be fine by me.  The important point is that they agree.

Mailaender: Renamed.

packaging/windows/OpenRA.nsi
```diff
@@ -200,7 +202,9 @@ Function ${UN}Clean
 	Delete $INSTDIR\README.html
 	Delete $INSTDIR\CHANGELOG.html
 	Delete $INSTDIR\CONTRIBUTING.html
-	Delete $INSTDIR\DOCUMENTATION.html
+	Delete $INSTDIR\TRAITS.html
+	Delete $INSTDIR\Lua-API.html
```
pchote: Filename case here needs fixing.

Mailaender: Right. Fixed.

pchote:
The output this generates includes a section on the launch parameters, which have nothing to do with the settings infrastructure.  This is quite confusing.

Mailaender:
I disagree and that request also has nothing to do with this pull request. You should have commented on https://github.com/OpenRA/OpenRA/pull/11571 and as it already had reached consensus you can submit your own patch and find voters for it.

pchote:
This PR is about shipping a file called SETTINGS.html with the implication that it is documenting the game's settings.yaml.  The  Launch block is right in the middle of this file (between the graphics and server blocks), with no indication that this has nothing to do with settings.yaml.  The first sentence of the page claims "This documentation is aimed at server administrators.", but those arguments aren't used for servers either.

We should not be shipping incorrect documentation, so we should not be merging this without clarification to either the purpose of the documentation, or its content. :-1:

Also shame on you for calling review feedback invalid because *I didn't catch it early enough*.  I had made it clear well in advance that I was going to be busy with work for all of July and most of August, and that I wouldn't be available for seriously reviewing PRs.  There wasn't any indication from the notification chatter that those docs included the Launch block.

Mailaender:
As I already said: be my guest. File a pull request to modify the settings extraction output yourself. You are not the only one busy with work.

pchote:
Moving to the future milestone as part of PR queue cleanup.

Mailaender:
>  I'd much prefer we remove the misnamed DOCUMENTATION.html from the installers instead of shipping even more files that people won't read.

I did just that now after https://github.com/OpenRA/OpenRA/pull/13215/ has been merged. Also fixed the other minor requested changes.

Preview: https://github.com/OpenRA/OpenRA/wiki/Settings

pchote:
I'd like to apologise for my comment above.  While I stand by the points I raised, the tone I used was uncalled for and unacceptable.  I was going through a stressful time then, and let frustration get the better of me.

Mailaender:
Nevermind. I had exactly the same feeling. Not sure how it escalated that quickly over such minor issues. Also added some field descriptors that we described manually before  https://github.com/OpenRA/OpenRA/wiki/Settings/_compare/7efefffd47c07b53dc384571edc9f4f686daecac...1034cb2b44b540429e6500407ad919a5ddbc05ec

pchote:
> This documentation is aimed at server administrators.

This is no longer really true, the server will ignore most of these settings and we have https://github.com/OpenRA/OpenRA/blob/bleed/launch-dedicated.sh to document the settings that it does expect.

Can you please update the header to reflect that this is documenting settings.yaml for the game / players?

Mailaender:
Yes, done.

pchote (approved):
It would be good if you could squash down some of the commits, but I won't block merging over that.

reaperrr (approved):

reaperrr:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/45ae80d3c029f87c33bda5c5b0d1631e6c67740f...0a50c65d209005d3d1fabc8d5a8be1efd12b1e0f

Issue #11077, opened 2016-04-09
Mammoth Tank Speed Buff


SoScared:
Increase Mammoth tank speed from 42 to 50

It seems like every attack done with Mammoth tank nowadays is a guaranteed suicide run. When target fired upon the tank simply can't get away in time. Pulling back damaged Mammoths in time for health regeneration isn't working. The ability of crushing walls in the playtest is great but only accounts for use in specific situations. 

For comparisons, the Soviet MAD tank moves at speed 54. Zombies at 42 :)

The current Mammoth Tank is a sleeping beast.

AoAGeneral:
Considering that in TD the mammoth speed is 56 and performs rather well, might want to look at testing the Mammoth tanks missile fire. See what its like if it fires cannons and missiles at all targets. This sort of change might actually make the mammoth a lot more dangerous then what it is now.

HenrytheSlav:
We can return to this now, sicne feature freeze is over. I feel like 50-56 seems okay. I could make a PR after this gets decided. Maybe introduce a poll on forum? It is a balance change after all.

SoScared:
I've been fiddling a bit with the 50 speed on a modded map. The reason I like this speed is that it's fast enough to allow for better damage pull-backs (including IC and wall crush play) and can follow infantry squads better but slow enough to not simply become a heavy tank upgrade.

MustaphaTR:
#12040 has merged, i think this can be closed.

PR #11731, opened 2016-07-29
[RA] Increase Mammoth Tank Speed 42->50

SoScared:
Discussed and playtested extensively. Pretty straightforward balance change in order to make the Mammoth Tank useful for it's own purposes.

http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&p=295320#295320

Opinions appear to be unanimous #11077, #10437

### DIFF
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -164,7 +164,7 @@ V2RL:
 	Armor:
 		Type: Heavy
 	Mobile:
-		Speed: 42
+		Speed: 50
 		Crushes: wall, mine, crate, infantry, heavywall
 	RevealsShroud:
 		Range: 6c0


r34ch:
:-1:

Although I kind of see the logic (more survivability via micro to use the health regen) i think this is a poor implementation. This moves the mammoths role _even closer_ to that of the heavy tank and without any changes to turn speed I can't see this actually having an impact in relation to survivability - its seems like an arbitrary buff for the sake of a buff.

Although it buffs the ability to get in position to crush concrete walls, where is the late game Allied option to do the same?

SoScared:
Not arbritary - read the discussions.  In conclusion, the only issue with the Mammoth tank was the inability to pull it back after using it as a breach unit. Even with the help of the Iron Curtain the unit becomes a liability and not worth the investment. Turret rate was never an issue.

The Mammoth Tank could potentially replace the Heavy tank on maps with no manouverability, such as Singles, Doubles and on team games with chokes. No issue imo.

If concrete walls is a problem for Allies then that should be solved in a seperate issue.

Micr0Bit:
im for putting this on hold ... since we already have enough balance changes for RA

MurtoTheRay:
Mammoth tanks CAN be powerful in their current state but require numbers to do so. I suggest removing the CustomBuildTimeValue trait from the mammoth tank or reduce it's custom build time value.

r34ch:
I'm not talking about the turret turn rate i'm taking about vehicle turn rate. If they are being used as breach units then they are taking most hits when either 
- pausing to fire (breaking through) 
- turning on the spot to retreat (inability to pull back)

When they are moving they don't take as much damage due to bullet travel time. You are deciding to improve upon this aspect when arguably it can be said this isn't the aspect that needs improved.

SoScared:
Sorry, misunderstood. 

However Mammoth, Heavy and Medium Tanks currently recieve the exact same turret damage when still or in motion. I'd love to see a change with that some time in the future tho.

airetaM:
You should think about that arty is already rangenerfed for next playtest.
That means mammoth will be attacked later by them and they can faster reach an enemys base wich is a big advantage. 
I think that doesnt fit to that unit. i would prefer a faster selfheal rate for them because they difinitly need a slight buff, but we will see in the playtest how the game is balanced then.

AoAGeneral:
I think the mammoth speed is alright in all honesty. Turning speed of the Mammoth being slow can be a weakness of the Mammoth in all cases. I would actually test having the Mammoth fire slightly faster and be able to fire missiles at all targets rather then just infantry only. This would make the Mammoth not only a damage soaker but a damage dealer as well. Right now mediums and heavy tanks out damage the Mammoth. Ive also had a comment from an RA player saying how the TD Mammoths are much stronger then the RA ones.

SoScared:
@AoAGeneral the aim of the PR is to help Mammoth Tanks to break stalemate frontlines, not introduce it as a late game main tank as in TD. As of now the Mammoth Tank does neither.

For the record the turning speed of Mammoth and Heavy Tanks are virtually identical.

The Mammoth Tanks' weakness is demonstrably the speed. As for Artillery, eating tons Artillery fire in and out of combat is a good example on how the Mammoth Tank remains dormant. Its' selfhealrate doesnt factor in if it's lost regardless if you commit or not.

AoAGeneral:
Im under the impression that the self heal rate does not tick at all until x amount of seconds out of combat damage? I believe this mechanic was in both old school RA and CNC

Micr0Bit:
@AoAGeneral : indeed

pchote:
We'll need to make a decision to either take, adjust, or close this before we can ship the next playtest.

Micr0Bit:
well they just recently got a buff in the last release so they cant get blocked by walls ... they just crush walls now ... 
in this release , artys get a nerf ... i dont see it justified for mamoths to get another buff now ... 
we already have enough balance changes for this release 

:-1:

SoScared:
Mammoth Tanks have been utterly void in this release and thus so is the argument that the wall crush ability is a sufficient buff. Even the Chrono Tank have seen more light this release, despite being looked upon as an underperforming unit. The fact is, just like with other "exotic" units in the RA mod, Mammoths are mostly built for spoofs and by new players. It's built when players sort of shop around the build menu looking for fun units to play. What the zombie-like speed is doing is literally punishing players looking to have fun with the unit.
 
The Mammoth Tank getting eaten by artillery fire is at the core of the problem for the unit. Attack or retreat the Mammoth is dead.

Even at speed 50 the Mammoth Tank  is _vastly_ outperformed by the Heavy Tank. Imo, with exception to spammy big team player choke maps the change isn't gonna get anywhere close to putting the balance at risk.

Get this through a playtest already :+1:

SoScared:
Worth to mention, any future buff to the Mammoth Tank will likely have to adopt a speed buff _regardless_. Balancing a 42-speed Mammoth tank in the future would likely be time consuming and problematic as it'd need to overcompensate health, damage or some other new excotic feature just to make it relevant in a game. Playtesting the Mammoth is on par with playtesting Cruisers, Chrono Tanks and other late tech units - it needs _months_, sometimes more, before the balance change settles in the game. 

As I see it we get this through or we live with a dud unit for more releases to come.

pchote:
The logic behind this doesn't seem unreasonable, and we can always revert it if the playtest shows this was a bad move.  http://www.sleipnirstuff.com/forum/viewtopic.php?t=19767 indicates that the wider community is more for the change than the comments above suggest.  :+1:

r34ch:
> It's a neccessary first step.

But its not. I cant understand why the issue of survivability is being tip toed around here by increasing **speed** when the HP is straight up the issue here? 

2 Mammoths - 1800hp - $4000 - 80s
3 Heavies - 1650hp - $3450 - 84s

For a negligible difference in HP (150hp) I can have 3 heavies in the same build time, much cheaper, much earlier in the game, that are faster and do more dps.

Buffing speed fixes none of these issues. 

People are complaining about underused units. Buffing the mammoths speed is making the mammoth more like the heavy rather than making it its own unique unit with a functional purpose.

SoScared:
>But its not. I cant understand why the issue of survivability is being tip toed around here by increasing speed when the HP is straight up the issue here?

It's a necessary first step if you can't buff the 42 speed Mammoth Tank without making it a dumb unit for the sake of uniqueness.  E.g. buffing its HP alone is gonna move it along the route of becoming a "boss" unit. A slow beast moving towards you, you'd have to try stopping it like you're player Murto's Tower defense.

No, whatever will happen with this unit in the future it's not going to circumvent the speed buff.

>Buffing speed fixes none of these issues.

These were never really issues and I'm confused - are you arguing against the speed buff because it won't help the Mammoth Tank become a main battle tank?

Micr0Bit:
i agree with @r34ch regarding the buff itself 

though the main-issue i have , is with the time we get this in ... the next release would be inappropriate imo ... as (like i already said) a main-unit of allies gets nerfed ... i found it wrong to throw in a soviet-buff of a main soviet-unit ... at the same time (while it already got buffed in the release before) 

so in theory , im not even against giving mammoths a better buff ... we just currently have enough for the playtests ...

SoScared:
@Micr0Bit I'm sorry but your measuring of balance with unit by unit nerf/buff is nonsensical, especially considering the lack of crossover dynamics between the Mammoth Tank and Artillery in context of the Heavy Tank comparison. You also ignored the fact that the wall crush buff the previous release did little to nothing to invite Mammoth Tanks into the games.

Murto:
There isn't really an issue here of whether increasing the speed of the mammoth would be a general improvement it it or whether it would make other units worthless or whatever. The real issue is that no one can agree on what the mammoth tank's role should be.

Should it be an all-purpose upgrade for the heavy? (generally better damage/health and decent speed)

Should it be a Goliath that needs a lot of pounding from units to bring it down (good health, decent damage, low speed)

These are the two main ideas being thrown around and neither is wrong. What is wrong is that there isn't an agreement of which it should be or if it should be something else.

There is no point arguing over these kinds of buffs until everyone decides on the role.

anjew175:
> im not even against giving mammoths a better buff ... we just currently have enough for the playtests

so there is only a finite amount of balances needed? What happens when you add too many? Does it get too balanced?

SoScared:
@Murto: I tend to agree but the speed buff is mainly to pull it out of a completely dysfunctional state. As I see it, speed 42 Mammoth Tanks doesn't have a future. As for designating a specific role to it I'm open to all sorts of ideas.

Micr0Bit:
@anjew175 :  if you throw in to much , you might risk that people (regular players) wander off when they see that you changed the whole gameplay in 1 release ... 

"balance-changes" have to get in "bite after bite" ... jamming it up the players throat is the wrong way

anjew175:
you're saying that this specific balance change is the straw that broke the camels back? After you just admitted that the change would do nothing? lol

Micr0Bit:
im saying that we have "enough" for next release 
we shouldnt add "useless buffs" but i wouldnt +1 a useful buff either "right now" ... because , we have enough

SoScared:
With the exception for the Artillery nerf and changes to the Concrete damage, the balance changes consists mostly of late-game air tech changes that will take _months_ to settle, way beyond the playtests. So no. We have currently an abysmal amount of changes for the next playtest.

I've done what I could with this, the playtest is around the corner. If this balance PR doesn't fill the requirement for testing, reasoning and consensus (first comment), I'm not sure exactly what will.

SoScared:
Closing in favor of #11927

PR #11744, opened 2016-07-30
Close OpenAL sound device correctly

### LINKED ISSUE #11193: OpenAL device not being properly closed
I always get reminded in the command line when closing the game window with `x` buttons.

> AL lib: (EE) alc_cleanup: 1 device not closed

RoosterDragon:
Fixes #11193, plus some additional code tiding.

### DIFF
--- a/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
+++ b/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
@@ -49,6 +49,7 @@ namespace OpenRA.Platforms.Default
 		readonly Dictionary<uint, PoolSlot> sourcePool = new Dictionary<uint, PoolSlot>();
 		float volume = 1f;
 		IntPtr device;
+		IntPtr context;
 
 		static string[] QueryDevices(string label, int type)
 		{
@@ -103,10 +104,10 @@ namespace OpenRA.Platforms.Default
 					throw new InvalidOperationException("Can't create OpenAL device");
 			}
 
-			var ctx = ALC10.alcCreateContext(device, null);
-			if (ctx == IntPtr.Zero)
+			context = ALC10.alcCreateContext(device, null);
+			if (context == IntPtr.Zero)
 				throw new InvalidOperationException("Can't create OpenAL context");
-			ALC10.alcMakeContextCurrent(ctx);
+			ALC10.alcMakeContextCurrent(context);
 
 			for (var i = 0; i < PoolSize; i++)
 			{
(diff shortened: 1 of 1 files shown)


OpenRA.Platforms.Default/OpenAlSoundEngine.cs
```diff
@@ -66,7 +66,7 @@ static string[] QueryDevices(string label, int type)
 			do
 			{
 				var str = Marshal.PtrToStringAuto(next);
-				next += UnicodeEncoding.Default.GetByteCount(str) + 1;
+				next += Encoding.Default.GetByteCount(str) + 1;
```
pchote: What's the difference between these encodings?

obrakmann: Wasn't that done explicitly to ensure it runs on Chinese systems etc.?

Mailaender:
:+1:

OpenRA.Platforms.Default/OpenAlSoundEngine.cs
pchote: That's my concern, yeah.

Mailaender: I might cause problems on systems that don't really work with Unicode properly. Saw some gibberish when running bleed on Windows lately. This might use the system one, which is a good compromise from https://github.com/OpenRA/OpenRA/pull/11183/commits/f22bc68d948531b90a7e9ce9dc6f93e021007836. This still should count the bytes properly.

RoosterDragon: `Encoding.Default` === `UnicodeEncoding.Default`. `Default` is a *static* member of the `Encoding` class. This is referring to the same encoding (the current system code page). `Encoding.Unicode` would be referring to a Unicode encoding.

This is a good reason to simply this name - it is inherently confusing.

obrakmann: Ok, so according to MSDN, `UnicodeEncoding` derives from `Encoding` and doesn't seem to overwrite the `Default` property, so `Encoding.Default` and `UnicodeEncoding.Default` are bound to be the same due to inheritance. This change therefore doesn't make any functional difference, it's still the same code in the end.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/93b91712dff3d65f716815a14f726c5e58e41f57)

PR #11752, opened 2016-07-31
Added SpawnDelay for SpawnActorOnDeath trait

Mistranger:
This new property will delay actor spawn on death (useful when you need to play unit's death animation first without overlapping dead actor and new actor)

### DIFF
--- a/OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs
+++ b/OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs
@@ -11,6 +11,8 @@
 
 using System.Linq;
 using OpenRA.Mods.Common.Warheads;
+using System;
+using OpenRA.Effects;
 using OpenRA.Primitives;
 using OpenRA.Traits;
 
@@ -45,6 +47,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Should an actor only be spawned when the 'Creeps' setting is true?")]
 		public readonly bool RequiresLobbyCreeps = false;
 
+		[Desc("Time (in frames) until spawn (0 means instant)")]
+		public readonly int SpawnDelay = 0;
+
 		public object Create(ActorInitializer init) { return new SpawnActorOnDeath(init, this); }
 	}
 
@@ -53,6 +58,7 @@ namespace OpenRA.Mods.Common.Traits
 		readonly SpawnActorOnDeathInfo info;
 		readonly string faction;
 		readonly bool enabled;
+		string huskActor;
 
 		public SpawnActorOnDeath(ActorInitializer init, SpawnActorOnDeathInfo info)
 		{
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs
```diff
@@ -75,38 +81,43 @@ public void Killed(Actor self, AttackInfo e)
...
-				};
-
-				if (info.OwnerType == OwnerType.Victim)
-					td.Add(new OwnerInit(self.Owner));
-				else if (info.OwnerType == OwnerType.Killer)
-					td.Add(new OwnerInit(e.Attacker.Owner));
-				else
-					td.Add(new OwnerInit(self.World.Players.First(p => p.InternalName == info.InternalOwner)));
+			// Actor has been disposed by something else before its death (for example `Enter`).
+			if (self.Disposed)
```
pchote: Moving the execution of all this code from the FrameEndTask to here may have unexpected side-effects.  I don't have the time to check this myself, but another reviewer definitely should.

OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs
```diff
@@ -75,38 +81,43 @@ public void Killed(Actor self, AttackInfo e)
...
+			if (info.OwnerType == OwnerType.Victim)
+				td.Add(new OwnerInit(self.Owner));
+			else if (info.OwnerType == OwnerType.Killer)
+				td.Add(new OwnerInit(e.Attacker.Owner));
+			else
+				td.Add(new OwnerInit(self.World.Players.First(p => p.InternalName == info.InternalOwner)));
 
-				foreach (var modifier in self.TraitsImplementing<IDeathActorInitModifier>())
-					modifier.ModifyDeathActorInit(self, td);
+			if (info.SkipMakeAnimations)
```
GraionDilach: This change seems wrong.

GraionDilach: Oh wait, this wasn't enabled previously, my mistake.

OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs
reaperrr: I don't really see the need to move all this code, either. In my opinion the code would be both safer and much easier to review if only the delay-relevant lines were added to the end of the FrameEndTask and the rest was left untouched as far as possible.

If there's a specific reason for the move, please explain, otherwise I'd request to move everything back into the FrameEndTask.

reaperrr (changes requested):

reaperrr:
> OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs:L113: [SA1507] The code must not contain multiple blank lines in a row.
OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs:L12: [SA1210] Using directives must be sorted alphabetically by the namespaces.
OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs:L14: [SA1208] System using directives must be placed before all other using directives.

Additionally, these style issues need to be adressed.

reaperrr:
Superseded by #12288.

PR #11753, opened 2016-07-31

RoosterDragon:
Allowing VS to format this file as it likes means users of this IDE don't need to reset it each time VS decides to streamroll over it with arbitrary changes.

Updates version number to VS2015 Update 3 and sorts files in its preferred order. (See also #11363)

### DIFF
--- a/OpenRA.sln
+++ b/OpenRA.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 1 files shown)


pchote:
Considering that most development is now (arguably, I guess) done with monodevelop, wouldn't it make more sense to standardize on it's preferred format?  Letting monodevelopisms creep in over time between VS resets just means that everyone has to steamroll changes.

RoosterDragon:
Does monodevelop like to arbitrarily reorganize the solution file too? Is it incompatible with VS's preferred format?

Mailaender:
Well, I assume that the Mac guys use Xamarin Studio 6 while Linux is stuck to MonoDevelop 5 I doubt we will get common grounds here. MonoDevelop tries to adhere to the Visual Studio format anyway so I am okay that current VS is the standard. Also not much to argue about such minor changes. :+1: 

What I would like to see in the future is automatic code formatting on file save instead of manually fiddling with the code to make StyleCop happy. I guess that information is also stored in the .csproj files. Hopefully we can get something that works everywhere here.

PR #11757, opened 2016-08-01
Remove the "Output Disabled" audio device and simplify platform initialization.

### LINKED ISSUE #11252: FileNotFoundException when initializing the game
Hi,
I just installed the game and fiddled with the settings, saved and restarted.
Then I had an invisible cursor, so I reset the configuration.

Now the game does not start anymore (even after reinstall) and fails with this being the only error:

**exception.log**
```
Runtime Version: .NET CLR 4.0.30319.18444
Exception of type `System.IO.FileNotFoundException`: Das System kann die angegebene Datei nicht finden. (Exception from HRESULT: 0x80070002)
   at System.Reflection.RuntimeAssembly.nLoadFile(String path, Evidence evidence)
   at System.Reflection.Assembly.LoadFile(String path)
   at OpenRA.Sound..ctor(String engineName)
   at OpenRA.Game.Initialize(Arguments args)

```

Is there a possible workaround for me? (:

pchote:
The sound engine parameter was added for three main reasons, which are all now invalid:
- To allow dedicated servers to run headless (no longer needed thanks to the overhaul in the last release)
- As an automatic fallback if OpenAL can't be initialized (broken somewhere between release-20150919 and release-20151224)
- As a way for people to completely mute audio (we now have a mute checkbox)

These plus #11252 (game crashes on launch if the "Output Disabled" option is selected in release-20160508) convinced me that this feature has outlived its usefulness and should be removed.

I also used this chance to clean up the platform initialization in general.

Fixes #11252.


Mailaender:
I can confirm this fixes #11252 :white_check_mark: and agree on removing the unused modular sound engine approach. :+1:

OpenRA.Game/Graphics/IGraphicsDevice.cs
```diff
@@ -15,23 +15,10 @@
...
-			Type = graphicsDeviceType;
-		}
-	}
-
-	public interface IDeviceFactory
+	public interface IPlatform
 	{
 		IGraphicsDevice CreateGraphics(Size size, WindowMode windowMode);
-		ISoundEngine CreateSound();
+		ISoundEngine CreateSound(string device);
```
Mailaender: I guess this file and it's directory location should change at one point as it contains more than `IGraphicsDevice` now.

pchote: Done.

pchote:
This doesn't remove much modularity in practice – you can still switch between platform dlls, and these can choose to implement their own sub-devices if they have a need to.

Mailaender:
Guess it makes sense if someone adds a DirectX platform that it will then contain Direct3D and DirectSound altogether.

OpenRA.Game/Renderer.cs
```diff
@@ -79,17 +77,6 @@ static Size GetResolution(GraphicSettings graphicsSettings)
 			return new Size(size.X, size.Y);
 		}
 
-		static IGraphicsDevice CreateDevice(Assembly platformDll, int width, int height, WindowMode window)
-		{
-			foreach (PlatformAttribute r in platformDll.GetCustomAttributes(typeof(PlatformAttribute), false))
```
obrakmann: You can remove the `System.Reflection` using from both `Renderer.cs` and `Sound.cs` now.

pchote:
Updated.

OpenRA.Game/Renderer.cs
pchote: Done.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/e5e098af7f61e615e8540c6f961389f61e7da752)

PR #11771, opened 2016-08-05
Rework mod enumeration and split content metadata into their own files.

### LINKED ISSUE #11740: Content installer metadata should be split into their own files

pchote:
This reworks the mod enumeration and caching code so that we are able to create and mount one mod's filesystem inside and isolated from another mod.

While I was there I moved the last of the mod content code out of the engine, and added support for loading mods from arbitrary locations by specifying `Game.Mod=/full/path/to/your.oramod` (something we talked about a lot, but which I couldn't find a ticket for).

Fixes #11740.


pchote:
The commits are split in a way that this could be reviewed as three separate PRs, but I figure that people are sick of my never-ending PR chains by now.

Mailaender:
:+1: Looking good.

I haven't tested the `Game.Mod=/full/path/to/your.oramod` nor if the content extractors still work.

OpenRA.Game/IUtilityCommand.cs
```diff
@@ -20,6 +30,6 @@ public interface IUtilityCommand
 
 		bool ValidateArguments(string[] args);
 
-		void Run(ModData modData, string[] args);
+		void Run(Utility utility, string[] args);
```
obrakmann: You might as well add the `[Traits.RequireExplicitImplementation]` attribute here when you already did all the work.

OpenRA.Game/InstalledMods.cs
```diff
@@ -0,0 +1,111 @@
...
+
+			return mods;
+		}
+
+		static Dictionary<string, Manifest> GetInstalledMods()
+		{
+			var ret = new Dictionary<string, Manifest>();
+			foreach (var pair in GetCandidateMods())
+			{
+				IReadOnlyPackage package = null;
```
obrakmann: If you feel like it, you could already factor out the body of this loop into a `LoadMod` method or so, which will be needed for mod installation down the line, anyway.

OpenRA.Utility/Program.cs
```diff
@@ -16,6 +16,7 @@
 
 namespace OpenRA
 {
+	using System.IO;
```
obrakmann: Why is this here and not with the other `using`s above?

obrakmann:
Aside from those minor nigglings, looks and works fine. :+1:

pchote:
Updated.

OpenRA.Game/IUtilityCommand.cs
pchote: Done.

reaperrr:
Looks and works fine, as far as I could test it (I don't have any CDs or TFD around, so I can't test their installations).
I also tested the `Game.Mod=/full/path/to/your.oramod` command and it seems to work fine on Win7x64.

:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/cf1c5a588dcbcd00250d9a809b3d845f772fc373...d8c61eea9eb5ebe4e8df894962d3cd5091521962)

PR #11774, opened 2016-08-06
Added dynamic railroad track terrain to Tiberian Sun

Mailaender:
Another small step towards https://github.com/OpenRA/OpenRA/issues/2652.

### DIFF
--- a/OpenRA.Mods.Common/Traits/ChangesTerrain.cs
+++ b/OpenRA.Mods.Common/Traits/ChangesTerrain.cs
@@ -1,0 +1,50 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Modifies the terrain type underneath the actors location.")]
+	class ChangesTerrainInfo : ITraitInfo, Requires<ImmobileInfo>
+	{
+		[FieldLoader.Require] public readonly string TerrainType = null;
+
+		public object Create(ActorInitializer init) { return new ChangesTerrain(this); }
+	}
+
+	class ChangesTerrain : INotifyAddedToWorld, INotifyRemovedFromWorld
+	{
+		readonly ChangesTerrainInfo info;
+		byte previousTerrain;
+
+		public ChangesTerrain(ChangesTerrainInfo info)
+		{
+			this.info = info;
+		}
+
+		void INotifyAddedToWorld.AddedToWorld(Actor self)
+		{
+			var cell = self.Location;
+			var map = self.World.Map;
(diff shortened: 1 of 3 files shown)


pchote:
I'm not a big fan of having three completely separate implementations for doing the same thing (`Bridge`/`BridgeLayer`, `LaysTerrain`/`BuildableTerrainLayer`, and now `ChangesTerrain`).

There's got to be a better way forward than this...

OpenRA.Mods.Common/Traits/ChangesTerrain.cs
```diff
@@ -0,0 +1,41 @@
...
+	class ChangesTerrain : INotifyAddedToWorld
+	{
+		readonly ChangesTerrainInfo info;
+
+		public ChangesTerrain(ChangesTerrainInfo info)
+		{
+			this.info = info;
+		}
+
+		void INotifyAddedToWorld.AddedToWorld(Actor self)
```
pchote: You will need to clean up after yourself on removal, and think about what happens if CustomTerrain is already non-null.

Mailaender: Not for indestructible railways at the moment, but yes indeed later on that will be required.

pchote: This is a general purpose trait that will be available to all modders, so i'm not really happy with the idea of merging this with bugs.

Mailaender: You are right. This is an unnecessary limitation. Fixed.

abcdefg30:
What about mobile units?
Also https://github.com/OpenRA/OpenRA/pull/11774#r73786035:
> [...] and think about what happens if CustomTerrain is already non-null.

pchote:
> What about mobile units?

The `Requires<ImmobileInfo>` addresses that.

OpenRA.Mods.Common/Traits/ChangesTerrain.cs
```diff
@@ -0,0 +1,50 @@
...
+			var map = self.World.Map;
+			var terrain = map.Rules.TileSet.GetTerrainIndex(info.TerrainType);
+			previousTerrain = map.CustomTerrain[cell];
+			map.CustomTerrain[cell] = terrain;
+		}
+
+		void INotifyRemovedFromWorld.RemovedFromWorld(Actor self)
+		{
+			var cell = self.Location;
+			var map = self.World.Map;
```
pchote: These are redundant, just use them inline.

LipkeGu:
>I'm not a big fan of having three completely separate implementations for doing the same thing >(Bridge/LowBridge/BridgeLayer, LaysTerrain/BuildableTerrainLayer, and now ChangesTerrain).
>
>There's got to be a better way forward than this...

ChangeTerrainLayer for all would be good for all three

OpenRA.Mods.Common/Traits/ChangesTerrain.cs
Mailaender: They make the code easier to read for me.

abcdefg30 (approved):

pchote (approved):

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/212324080959dc53a9f78990694911ee7e7ef73b

PR #11775, opened 2016-08-06
Added a TerrainDebugOverlay

Mailaender:
I added this to debug weird behavior of my trains near tunnels. Can also be used to investigate problems like https://github.com/OpenRA/ra2/issues/150. Complements https://github.com/OpenRA/OpenRA/pull/11211.


pchote:
It would be more efficient if this used the same trait as the custom terrain overlay.

PR #11781, opened 2016-08-07
Same unit command sounds wont overlap anymore

Turupawn:
When the player submits an order to a unit, it will now stop the previous sound before playing the new one.
This will prevent ventriloquists units.

I made a [short video](https://youtu.be/4uljtQ3zRjU) that shows the results.

Sorry for the graphics lag, I haven't got a proper screen recording setup right now.

### DIFF
--- a/OpenRA.Game/Sound/Sound.cs
+++ b/OpenRA.Game/Sound/Sound.cs
@@ -10,6 +10,7 @@
 #endregion
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using OpenRA.FileSystem;
 using OpenRA.GameRules;
@@ -39,6 +40,7 @@ namespace OpenRA
 		ISound music;
 		ISound video;
 		MusicInfo currentMusic;
+		Dictionary<uint, ISound> currentSounds = new Dictionary<uint, ISound>();
 
 		public Sound(IPlatform platform, SoundSettings soundSettings)
 		{
@@ -348,9 +350,18 @@ namespace OpenRA
 			var name = prefix + clip + suffix;
 
 			if (!string.IsNullOrEmpty(name) && (p == null || p == p.World.LocalPlayer))
-				soundEngine.Play2D(sounds[name],
+			{
+				var sound = soundEngine.Play2D(sounds[name],
 					false, relative, pos,
 					InternalSoundVolume * volumeModifier, attenuateVolume);
+				if (id != 0)
+				{
+					if (currentSounds.ContainsKey(id))
+						soundEngine.StopSound(currentSounds[id]);
+
+					currentSounds[id] = sound;
+				}
+			}
 
 			return true;
 		}


OpenRA.Game/Sound/Sound.cs
```diff
@@ -39,6 +40,7 @@ public sealed class Sound : IDisposable
 		ISound music;
 		ISound video;
 		MusicInfo currentMusic;
+		Hashtable currentSounds = new Hashtable();
```
GraionDilach: Wouldn't a HashSet suffice here?

GraionDilach: Oh wait, now I realized what's the point of using a table. Forget it.

pchote: Please make this a `Dictionary<int, ISound>` instead.

OpenRA.Game/Sound/Sound.cs
```diff
@@ -348,9 +350,27 @@ public float VideoSeekPosition
...
+					}
+
+					ISound sound = soundEngine.Play2D(sounds[name],
+						false, relative, pos,
+						InternalSoundVolume * volumeModifier, attenuateVolume);
+						currentSounds.Add(id, sound);
+				}
+				else
+				{
+					soundEngine.Play2D(sounds[name],
```
pchote: You can avoid duplication here by pulling out a single `var sound = soundEngine.Play2D(...)` before doing the `currentSounds` manipulation.

Turupawn:
Thanks guys
Looking better now?

OpenRA.Game/Sound/Sound.cs
```diff
@@ -348,9 +350,21 @@ public float VideoSeekPosition
 			var name = prefix + clip + suffix;
 
 			if (!string.IsNullOrEmpty(name) && (p == null || p == p.World.LocalPlayer))
-				soundEngine.Play2D(sounds[name],
+			{
+				ISound sound = soundEngine.Play2D(sounds[name],
```
pchote: OpenRA code style is to use `var` instead of explicit types in unambiguous cases like these.

OpenRA.Game/Sound/Sound.cs
```diff
@@ -348,9 +350,21 @@ public float VideoSeekPosition
...
 					InternalSoundVolume * volumeModifier, attenuateVolume);
+				if (id != 0)
+				{
+					if (currentSounds.ContainsKey(id))
+					{
+						soundEngine.StopSound(currentSounds[id]);
+						currentSounds.Remove(id);
+					}
+
+					currentSounds.Add(id, sound);
```
pchote: On second look you can simplify this a bit further by dropping the `.Remove(id)` above and changing this to `currentSounds[id] = sound`.

pchote:
Just a couple more small code style nits.  At a code-design level its a bit horrible to filter this at the sound engine level, but I can see that this is the only viable approach until we tackle our long-wanted sound engine refactoring.  `PlayPredefined` with `id != 0` is only ever called from the actor voice code, so this shouldn't regress any other cases.

OpenRA.Game/Sound/Sound.cs
GraionDilach: Wouldn't taht only work if the entry already exists in the dictionary?

Turupawn:
Cool!
What about now?

OpenRA.Game/Sound/Sound.cs
Turupawn: So far Remove() is working fine. Looks like it doesn't do anything if it the key doesn't exists.

OpenRA.Game/Sound/Sound.cs
```diff
@@ -348,9 +350,20 @@ public float VideoSeekPosition
...
+			{
+				var sound = soundEngine.Play2D(sounds[name],
 					false, relative, pos,
 					InternalSoundVolume * volumeModifier, attenuateVolume);
+				if (id != 0)
+				{
+					if (currentSounds.ContainsKey(id))
+						soundEngine.StopSound(currentSounds[id]);
+
+					currentSounds.Remove(id);
```
GraionDilach: Isn't this redundant (since the following line would overwrite this entry anyway)?

WolfGaming: This line is indeed not needed.

WolfGaming:
This looks good except for that one nit, :+1: from me when that's fixed.

Turupawn:
Fixed!
Thanks @GraionDilach @WolfGaming !

GraionDilach:
:+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a2c99ec612f27893a46e01c2b65ee2c4d4fa782b...81d2a6a2190cf99b5aa0f25362acdf1731dd79f0)

Mailaender:
Your first pull request merged here if I am not mistaken. :tada: You may want to add yourself to https://github.com/OpenRA/OpenRA/blob/bleed/AUTHORS

Issue #10382, opened 2016-01-02
Arbitrary zoom / custom resolution support


pchote:
This outlines my plan for fixing our issues relating to zoom, custom resolutions, and HiDPI support in one fell swoop.  I've referred to it several times in the past, so I figure it is time to document this properly.  From my perspective this supersedes all the previous tickets, but of course others are free to argue against this.

**User interface**

Two new dropdowns are included in the display settings: UI resolution and World resolution (like UT3 and i'm sure other games have done).  They contain values for "Native", a selection of popular defaults with a fixed width and a height calculated from the aspect ratio, and a "custom" option that exposes text fields for entering a manual width.  These set the (fixed) size of the UI, and the default size/zoom of the world.

Players can use ctrl + scroll to arbitrarily zoom the world (but not the UI) between the minimum and maximum limits (explained below).  The `.` key (currently used for pixel doubling) is used to reset the world to the default zoom configured in the settings.

**Implementation**

The world rendering (everything that scales with the world, so this excludes selection decorations, target lines, etc) is rendered into a framebuffer that is configured a viewport size calculated from the zoom and default resolution, and the result is rendered using a quad covering the window viewport.  The UI rendering (everything else from the world, and then the widgets) is similarly rendered first into a framebuffer (maybe the same as the first, depending on the cost of reconfiguring a buffer vs switching to a new one), and then scaled to fill the screen.

The default "Native" value is calculated by the window / display size divided by the native pixel scale (queried from SDL).  The pixel scale will be 2 for retina displays under OSX, or between 0.75 and 2 (I think?) for the different DPI options under Windows.  We can calculate the pixel scale by comparing the resolution that we asked for with the output from [SDL_GetRendererOutputSize](https://wiki.libsdl.org/SDL_GetRendererOutputSize).  The minimum supported resolution is defined in mod.yaml (with defaults for our mods of 1024 x 700), and the maximum values are queried from the GPU (the smaller of `GL_MAX_RENDERBUFFER_SIZE` / `GL_MAX_TEXTURE_SIZE`).

To enable HiDPI rendering we need to set `SDL_WINDOW_ALLOW_HIGHDPI` when [creating the window](https://wiki.libsdl.org/SDL_CreateWindow), and under OSX we also need to set `NSHighResolutionCapable` in Info.plist.  I have no idea how DPI scaling works across the various linux WMs or whether they are supported by SDL.

[...]

PR #11782, opened 2016-08-07
Removed unused SDL2-CS code paths

Mailaender:
The real change can be inspected at https://github.com/OpenRA/SDL2-CS/pull/1. It shouldn't cause any changes except us shipping a slightly smaller DLL.

### DIFF
--- a/thirdparty/fetch-thirdparty-deps.ps1
+++ b/thirdparty/fetch-thirdparty-deps.ps1
@@ -130,7 +130,7 @@ if (!(Test-Path "SDL2-CS.dll"))
 {
 	echo "Fetching SDL2-CS from GitHub."
 	$target = Join-Path $pwd.ToString() "SDL2-CS.dll"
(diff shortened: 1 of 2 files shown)


pchote:
We should use this chance to resync with upstream so that we get the SDL 2.0.4 bindings (wanted for #10382).

Mailaender:
Upstream https://github.com/flibitijibibo/SDL2-CS doesn't implement `GetDisplayDPI` yet. You have to contribute a patch for a DLLImport yourself.

pchote:
It does: https://github.com/flibitijibibo/SDL2-CS/blob/master/src/SDL2.cs#L1262

Mailaender:
Strange that GitHub code search didn't find it.

Mailaender:
I decided to go another route with this. https://github.com/OpenRA/OpenRA/pull/11793

PR #11788, opened 2016-08-07
Make turning-while-moving actors follow curved paths.

### LINKED ISSUE #11212: Turning looks too smooth
I've read criticism about how OpenRA's turning mechanism doesn't look good visually, and I'm inclined to agree. It's particularly noticable on half-circles around 1-cell obstacles and such, it somehow looks unnatural and too smooth.

pchote:
The code for moving units through cells independently lerps actor positions along disjoint straight-lines between cell edges while smoothly varying the facing.  This leads to subtly wrong-looking moves when units turn corners (actors move in a different direction to their facing).

The effect is small, but extremely irritating when you know what to look for... or even when you don't: this is the main cause of the often-repeated "OpenRA's turning is too smooth" complaint.

This PR replaces the linear position lerping with arcs.  It still ignores the unit's ROT because incorporating that would (a) be much more complicated and (b) break balance (fast units wouldn't be able to turn sharply enough to move without stopping).

Before and after videos:
- [before](https://www.dropbox.com/s/qp5u64egxgsdb5h/before.mov?dl=0)
- [after](https://www.dropbox.com/s/ziio8vbzldgwnn2/after.mov?dl=0)

Fixes #11212.

### DIFF
--- a/OpenRA.Game/WAngle.cs
+++ b/OpenRA.Game/WAngle.cs
@@ -64,6 +64,20 @@ namespace OpenRA
 			return new WAngle(Angle - 512).Tan();
 		}
 
+		public static WAngle Lerp(WAngle a, WAngle b, int mul, int div)
+		{
+			// Map 1024 <-> 0 wrapping into linear space
+			var aa = a.Angle;
+			var bb = b.Angle;
+			if (aa > bb && aa - bb > 512)
+				aa -= 1024;
+
+			if (bb > aa && bb - aa > 512)
+				bb -= 1024;
+
+			return new WAngle(aa + (bb - aa) * mul / div);
+		}
+
 		public static WAngle ArcTan(int y, int x) { return ArcTan(y, x, 1); }
 		public static WAngle ArcTan(int y, int x, int stride)
 		{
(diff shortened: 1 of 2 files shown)


reaperrr:
Subjectively, I feel that small U-turns

```
x
o  x
x
```

still look a bit more "glidingly" than in the originals.
To me it looks like the curve is a bit smaller and the visual facing changes a little faster in OpenRA than in WW's games.

But I guess replicating Westwood's movement with 100% accuracy would require far more trial-and-error and presumably more hacky code, and this is enough of an improvement in my eyes.
:+0.5: because the math is out of my league, so I only tested the PR.

obrakmann:
I fail to notice any difference at all, but it doesn't break anything and people obviously get their panties in a twist over this, so :+1:

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/243fb802a586cf25357adae21637fc31d8e3dddd...0cfb96d35e0f3c3ba661f8edf51b14856d70a7d5)

PR #11791, opened 2016-08-09
Fix voxel rendering with EnableDepthBuffer: True

pchote:
This gets us another step closer towards being able to enable the depth buffer rendering in the gen2 mods (nearly there!) by rendering voxels with an approximately correct z coordinate so that they are composited at (nearly) the right place in the world.

The "approximately" and "(nearly)" are because we throw away the per-pixel depth information associated with the model and for now can only render a flat billboard.  As commented in the code, this is because we have at most four colour channels to store 5 sets of data (RGB, alpha, and depth).  For now hacking the depth gives a good-enough result that we can improve further down the line when the voxel renderer is rewritten.

You'll notice when testing that the voxels clip beneath the ground when moving over slopes.  This isn't the renderer's fault: the heightmap-unaware movement code really is making the actors move partially underground, in addition to not being able to rotate them to match the slope.

### DIFF
--- a/OpenRA.Game/Graphics/VoxelRenderer.cs
+++ b/OpenRA.Game/Graphics/VoxelRenderer.cs
@@ -22,9 +22,9 @@ namespace OpenRA.Graphics
 		public readonly Sprite Sprite;
 		public readonly Sprite ShadowSprite;
 		public readonly float ShadowDirection;
-		public readonly float2[] ProjectedShadowBounds;
+		public readonly float3[] ProjectedShadowBounds;
 
-		public VoxelRenderProxy(Sprite sprite, Sprite shadowSprite, float2[] projectedShadowBounds, float shadowDirection)
+		public VoxelRenderProxy(Sprite sprite, Sprite shadowSprite, float3[] projectedShadowBounds, float shadowDirection)
 		{
 			Sprite = sprite;
 			ShadowSprite = shadowSprite;
@@ -145,7 +145,7 @@ namespace OpenRA.Graphics
 
 			var shadowScreenTransform = Util.MatrixMultiply(cameraTransform, invShadowTransform);
 			var shadowGroundNormal = Util.MatrixVectorMultiply(shadowTransform, groundNormal);
-			var screenCorners = new float2[4];
+			var screenCorners = new float3[4];
 			for (var j = 0; j < 4; j++)
 			{
 				// Project to ground plane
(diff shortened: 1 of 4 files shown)


pchote:
We have three options for enabling per-pixel voxel depth data in a future PR:

1. Rewrite the voxel renderer to directly draw into the screen buffer.
  * Pros: we keep the dynamic lighting calculations.
  * Cons: breaks support for the gen1 mods that don't use the depth buffer, and requires a major rewrite of our rendering code that may not ever happen.
2.  Store the depth data in the alpha channel.
   * Pros: we keep the dynamic lighting calculations and compatibility with the gen1 mods, requires only simple changes to the voxel renderer.
   * Cons: breaks any effects that rely on palette alpha (e.g. stealth), and shadows will need a special case hack.
3.  Replace the dynamic normal calculations with the VPL lookup table used by the original game.
   * Pros: keeps compatibility with palette effects and the gen1 mods, improves rendering performance and VRAM usage (because we can pack twice as many voxels to a framebuffer sheet).
   * Cons: our voxel rendering will match the original game, which will make the TS modding community very unhappy.

My preference is strongly for option 3, which is something I've been warning about for years now.

GraionDilach:
I'm fine with using a lookup table, if it's not the actual VPL, but a bunch of cloned palettes generated with the current light settings. The VPL was severely limited due to forcing all shades back to the palette, something the current renderer doesn't do. I'm even fine with manually creating the required 16/32 palettes to be used as normal tables.

pchote:
Letting the lookup table reference into other palettes takes us back into the "major engine rewrite" territory, so not likely to happen.  The vpl definitions will need to be per-palette, so nothing would stop you from using say 64 colours for the voxel and the other 192 for shaded versions.

GraionDilach:
Remap already clones palettes. I don't understand why voxels can't follow such a system, instead of the remap getting changed, the whole palette gets darkened/lightened. That would still work like the VPL but look dozens better than VPL ever will.

pchote:
The palette for a sprite is specified for the *whole* sprite in the vertex data.  Specifying a per-pixel palette is identical (from the code's perspective) to having a 16-bit indexed color (8 bits to select the palette, then 8 bits to select the index within that palette).

The renderer works by writing all the data to 32 bit textures, so sprites must choose between 8, 16, or 32 bits per pixel.  The color and depth channels take up 16 bits, adding a palette select channel bumps it to 24 which then doesn't pack cleanly.  It would have to waste the last 8 bits and kill the VRAM benefit... and thats without addressing the complexity of creating and defining the palettes for the up-to-256 normal facings (which will then easily blow our palette count limits).

Anyway, this is all an aside to the main purpose of this specific PR.

Mailaender:

reaperrr:
:+1: 
Test commit can be removed.

pchote:
Rebased and removed test commit.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a6bb03e3806d3c7b56c1fbbd1e480c2f52b713ad...9b693e9933c3b6742ebca9b69ed573ebbfb2226a)

PR #11793, opened 2016-08-09
Updated and included the SDL2 and OpenAL bindings

Mailaender:
We will never see releases for them, because they are meant to be included in commercial game ports. This gives us the benefit to adapt them to our needs (or rather strip what we don't need like #10288) and it is 2 fewer non-reusable dependencies for @rapgro and others to package which brings us closer to #8264.


pchote:
While this is less extreme than #8538 was, I still remain firmly against inlining these libraries for the same reasons that I argued in that PR.  Sorry, but :-1: from over here.

obrakmann:
The whole point of using upstream libraries is to not have to write and maintain them yourself. I certainly don't feel like doing either. :-1:

Mailaender:
What do you suggest instead? "Upstream" is
* https://github.com/OpenRA/SDL2-CS
* https://github.com/OpenRA/OpenAL-CS

Shall we include them with git submodules?

pchote:
My understanding of the fedora situation is that their package reviewers have decided that any library compiled from an outside repository (even ones from the same project) need to be packaged separately.  That is frankly stupid, and not at all our problem.  Taking the contents of that library/repo and importing them into OpenRA/OpenRA solves no practical problems and just makes our lives harder.

rapgro:
@pchote  Please read https://fedoraproject.org/wiki/Bundled_Libraries

sgallagher:
@pchote Please note that the bundled library link that @rapgro links to is the list of reasons why Fedora prefers not to have bundled libraries. However, it is also important to note https://fedoraproject.org/wiki/Bundled_Software_policy which states that bundling is no longer forbidden in Fedora and provides instruction on how to properly annotate the packages so that we know where bundling is occurring.

pchote:
@sgallagher good to know that the policy has changed.  I'm aware of the reasons for wanting libraries to be unbundled, but the issue that we had already discussed to death was that the upstream author of these otherwise excellent libraries is explicitly against unbundled use.

sgallagher:
@pchote Right, so at least within the context of Fedora, it would be okay to carry them bundled, provided that the package carries a comment in the spec pointing to upstream's reasoning and some other metadata that lets us figure out which packages need an update if a CVE happens.

pchote:
Indeed.  I'm glad to learn that there is a simpler way forward for fedora packaging!

PR #11794, opened 2016-08-09
Visualising waypoints

Turupawn:
Hello guys I think this solves this issue: #2598 in a very simplified and kind of presentable way. What do you think?

I'm woried about [this line](https://github.com/Turupawn/OpenRA/commit/d090ec758c9498143821b22edc6451526bb2a814#diff-d7a57eebc0a2b3d470e099590d38922eL97). I replaced it to make it work. Does it breakes another feature?


pchote:
To address #2598 properly it will need to work for all actor types (including aircraft) and activities.

I suggested using and/or modifying `GetTargets` in that ticket because all activities already have a basic implementation that you can build off.  `GetTargets` is currently used for a long-redundant debug overlay, so some activities (like `Move`) return targets for every cell in the path, but you can modify those to only return the end destination.

Mailaender:
My https://github.com/OpenRA/OpenRA/pull/11715 adds some more attack and move related `public override IEnumerable<Target> GetTargets(Actor self)` functions as those traits also make use of them.

Turupawn:
I think now I understand this better thanks to you and the guys that helped me at the IRC.
I just removed my List<Targets> and I'm now using GetActivityList() + GetTargets()
I had to deal with that NextActivity on the Tick() method in order to access the whole queue.
I think I'm close now, I know that I have to deal with that MoveFirstPart and MoveSecondPart activities, I'll look into that later.

pchote:
I'm sorry that I can't give more helpful (or more positive) feedback at this point, but I'm currently away on a business trip and can't commit the time to anything more in depth.

At a quick look the changes to Activity/Actor/Move break encapsulation by making things that should remain private public – we're not going to be able to take those.  The CPos.Null is also a problem because (-1, -1) is a valid coordinate that may be used various situations.

OpenRA.Game/Actor.cs
```diff
@@ -160,6 +160,23 @@ Rectangle DetermineVisualBounds()
 
 		public void Tick()
 		{
+			var activity_iterator = currentActivity;
+			List<Target> currentTargets = new List<Target>();
```
WolfGaming: Convention to use var instead of List<Target> here.

OpenRA.Game/Actor.cs
```diff
@@ -160,6 +160,23 @@ Rectangle DetermineVisualBounds()
...
+				}
+
+				activity_iterator = activity_iterator.NextActivity;
+			}
+
+			if (currentActivity != null)
+			{
+				currentActivity.DrawLines(this, currentTargets);
+			}
+
```
WolfGaming: This should probably not happen in Tick and not on the actor itself, for the first thing we probably want it in some render method, I can´t give you a suggestion since I am at work atm.

For the second this is because many of the Actors don't need this like buildings, trees etc so making this a trait that can be put on actors that require it is better.

OpenRA.Game/Activities/Activity.cs
```diff
@@ -40,6 +40,15 @@ public virtual IEnumerable<Target> GetTargets(Actor self)
 		{
 			yield break;
 		}
+
+		public virtual CPos GetDestination(Actor self)
+		{
+			return CPos.Null;
+		}
```
WolfGaming: Make this into a public abstract bool? Destination { get; } or public virtual bool? Destination { get { return null; } }, this allows you to drop the change to CPos and prevent a possible bug if some code uses -1, -1 as an actual destination (as in go outside the map) also allows you to simply do a Destination.HasValue for a null check.

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -179,7 +180,9 @@ public override Activity Tick(Actor self)
 			if (firstFacing != mobile.Facing)
 			{
 				path.Add(nextCell.Value.First);
-				return ActivityUtils.SequenceActivities(new Turn(self, firstFacing), this);
+				var turn_activity = new Turn(self, firstFacing);
+				turn_activity.NextActivity = this;
+				return turn_activity;
```
WolfGaming: The removed code and the added code should be doing the same thing so why the change?

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -196,6 +199,9 @@ public override Activity Tick(Actor self)
 					mobile.Facing,
 					0);
 
+				move.IsCanceled = this.IsCanceled;
```
WolfGaming: this.IsCanceled will always be false as we check if (IsCanceled) at the top of the method.

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -278,6 +284,16 @@ public override void Cancel(Actor self)
 			base.Cancel(self);
 		}
 
+		public override CPos GetDestination(Actor self)
+		{
+			return this.destination.Value;
+		}
+
+		public override void DrawLines(Actor self, List<Target> currentTargets)
```
WolfGaming: You probably want to use IEnumerable\<Target\> instead of List as that prevents the overridden methods from changing the List as Lists are passed by reference.

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -287,14 +303,24 @@ public override IEnumerable<Target> GetTargets(Actor self)
 			return Target.None;
 		}
 
-		abstract class MovePart : Activity
+		public abstract class MovePart : Activity
 		{
-			protected readonly Move Move;
+			public readonly Move Move;
```
WolfGaming: Why does this need to be public?

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -287,14 +303,24 @@ public override IEnumerable<Target> GetTargets(Actor self)
 			return Target.None;
 		}
 
-		abstract class MovePart : Activity
+		public abstract class MovePart : Activity
```
WolfGaming: This and all the other MoveSomething shouldn't need to be public.

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -326,6 +352,9 @@ public override Activity Tick(Actor self)
 					moveFraction = MoveFractionTotal;
 				UpdateCenterLocation(self, Move.mobile);
 
+				if (IsCanceled)
+					ret.NextActivity = Move.NextActivity;
+
```
WolfGaming: I am pretty sure canceling a MovePart is a bad idea, @pchote correct me if I am wrong.

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -394,6 +423,8 @@ protected override MovePart OnComplete(Actor self, Mobile mobile, Move parent)
 							Util.GetNearestFacing(mobile.Facing, self.World.Map.FacingBetween(mobile.ToCell, nextCell.Value.First, mobile.Facing)),
 							moveFraction - MoveFractionTotal);
 
+						ret.IsCanceled = this.Move.IsCanceled;
```
WolfGaming: Move FirstHalf shouldn't be canceled as it's a MovePart.

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -410,13 +441,15 @@ protected override MovePart OnComplete(Actor self, Mobile mobile, Move parent)
 					mobile.Facing,
 					moveFraction - MoveFractionTotal);
 
+				ret2.IsCanceled = this.Move.IsCanceled;
```
WolfGaming: Same here.

OpenRA.Mods.Common/Activities/Turn.cs
```diff
@@ -40,5 +43,10 @@ public override Activity Tick(Actor self)
 
 			return this;
 		}
+
+		public override void DrawLines(Actor self, List<Target> currentTargets)
+		{
+			self.SetTargetLines(currentTargets, Color.Red);
```
WolfGaming: Why is the color red and why do we draw a line in a turning activity that doesn't change the actors position?

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
```diff
@@ -72,13 +72,18 @@ public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr)
 			if (targets == null || targets.Count == 0)
 				yield break;
 
+			List<WPos> valid_targets = new List<WPos>();
```
WolfGaming: Convention to use var instead of List\<WPos\> and camel case for variables so change to var validTargets

OpenRA.Game/Actor.cs
```diff
@@ -160,6 +160,23 @@ Rectangle DetermineVisualBounds()
 
 		public void Tick()
 		{
+			var activity_iterator = currentActivity;
```
WolfGaming: Convention to use camel case aka activityIterator.

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -179,7 +180,9 @@ public override Activity Tick(Actor self)
 			if (firstFacing != mobile.Facing)
 			{
 				path.Add(nextCell.Value.First);
-				return ActivityUtils.SequenceActivities(new Turn(self, firstFacing), this);
+				var turn_activity = new Turn(self, firstFacing);
```
WolfGaming: Camelcase turnActivity.

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
```diff
@@ -72,13 +72,18 @@ public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr)
...
 			foreach (var target in targets)
 			{
 				if (target.Type == TargetType.Invalid)
 					continue;
 
-				yield return new TargetLineRenderable(new[] { self.CenterPosition, target.CenterPosition }, c);
+				valid_targets.Add(target.CenterPosition);
 			}
+
+			yield return new TargetLineRenderable(valid_targets, c);
```
WolfGaming: Why are we even using yield returns and yield breaks here if we aren't returning multiples inside the foreach loop?
Please use a normal return and return an empty array in cases where yield break is as yield is a pretty expensive keyword that shouldn't be used if there is no reason to.

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
```diff
@@ -94,7 +99,7 @@ public static void SetTargetLines(this Actor self, List<Target> targets, Color c
 		{
 			var line = self.TraitOrDefault<DrawLineToTarget>();
 			if (line != null)
-				self.World.AddFrameEndTask(w => line.SetTargets(self, targets, color, false));
+				line.SetTargets(self, targets, color, true);
```
WolfGaming: Why the change?
Maybe it's right I'm just interested in your reason for it :)

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -179,7 +180,9 @@ public override Activity Tick(Actor self)
 			if (firstFacing != mobile.Facing)
 			{
 				path.Add(nextCell.Value.First);
-				return ActivityUtils.SequenceActivities(new Turn(self, firstFacing), this);
+				var turnActivity = new Turn(self, firstFacing);
+				turnActivity.NextActivity = this;
+				return turnActivity;
```
WolfGaming: Why the change? For debugging?

OpenRA.Mods.Common/Activities/Turn.cs
```diff
@@ -9,7 +9,10 @@
  */
 #endregion
 
+using System.Collections.Generic;
+using System.Drawing;
 using OpenRA.Activities;
+using OpenRA.Mods.Common.Traits;
 using OpenRA.Traits;
```
WolfGaming: please revert the using changes to this class as you aren't doing anything here.

OpenRA.Mods.Common/Traits/Mobile.cs
```diff
@@ -500,21 +500,28 @@ void PerformMoveInner(Actor self, CPos targetLocation, bool queued)
...
+			PerformMoveInner(self, targetLocation, queued);
+
+			var activityIterator = self.GetCurrentActivity();
+			var currentTargets = new List<Target>();
+			while (activityIterator != null)
+			{
+				if (activityIterator.GetTargets(self).Count() > 0 && !activityIterator.IsCanceled)
+				{
+					currentTargets.Add(activityIterator.GetTargets(self).Last());
+				}
```
WolfGaming: Convention to have no curly braces if the body of the if is one line.

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
```diff
@@ -72,13 +72,18 @@ public IEnumerable<IRenderable> RenderAfterWorld(WorldRenderer wr)
...
 			foreach (var target in targets)
 			{
 				if (target.Type == TargetType.Invalid)
 					continue;
 
-				yield return new TargetLineRenderable(new[] { self.CenterPosition, target.CenterPosition }, c);
+				validTargets.Add(target.CenterPosition);
 			}
+
+			yield return new TargetLineRenderable(validTargets, c);
```
WolfGaming: Still using yield here, switch for returns, in
   `yield break;`
do
   `return new TargetLineRenderable[0];`
and instead of
   `yield return new TargetLineRenderable(validTargets, c);`
do
   `return new TargetLineRenderable[1] { new TargetLineRenderable(validTargets, c) };`

WolfGaming:
You are getting there, just a couple of fixups and some extensive testing on our part and you will be good to go :)

OpenRA.Mods.Common/Activities/Move/Move.cs
pchote: It is, yes.  I already said as much above.

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
Phrohdoh: Is there a reason you aren't suggesting `Enumerable.Empty<TargetLineRenderable>()`?

`return new TargetLineRenderable[1] { new TargetLineRenderable(validTargets, c) };`
could be:
`return new[] { new TargetLineRenderable(validTargets, c) };`

WolfGaming: No I didn't know Enumerable.Empty<TargetLineRenderable>() existed, I also didn't know that c sharp allowed the array to inherit they type in this case.

OpenRA.Mods.Common/Activities/Move/Move.cs
Turupawn: I did this beacause I needed to glue the Move activity after the Turn using:
`turnActivity.NextActivity = this;`

OpenRA.Mods.Common/Activities/Move/Move.cs
Turupawn: I'm canceling the MoveFirstHalf and MoveSecondHalf because in the render, I'm only drawing the [non canceled activities](https://github.com/Turupawn/OpenRA/blob/visualising_waypoints/OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs#L76).
What do you guys suggest? An extra flag maybe?

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
Turupawn: I don't know :dolphin: 
I just wasn't working until I did that, to make it work I copied this from `SetTargetLine()` I think.

Turupawn:
Thanks again guys!

Now these are the 2 two things that I need to think about (any help is welcome as always :octopus: ):

1. I removed the `List<Target> targets` from `class DrawLineToTarget` but now I'm concerned about the Harvester gold lines beacause it was also using that list. (I don't even know who the Harvester is! :rabbit: )
2. Find another way to know if the MoveFirstHalf and MoveSecondHalf's Move is canceled without using IsCanceled

pchote:
For (2) you shouldnt need to touch Move*Half at all: they can return no targets, then let Move (which is queued after them) return the destination cell.

OpenRA.Mods.Common/Activities/Move/Move.cs
pchote: SequenceActivities already does that. You shouldn't be manipulating NextActivity directly.

OpenRA.Mods.Common/Activities/Move/Move.cs
pchote: The move code is subtle and extremely fragile. Please try very hard to not change the way it queues and runs activities.

WolfGaming:
@Turupawn As pchote said you shouldn't care about the halves at all since the cannot be cahnged or canceled, just skip them.

OpenRA.Mods.Common/Activities/Move/Move.cs
Turupawn: I made this public again so it could be accessed from DrawLineToTarget

pchote: `DrawLineToTarget` doesn't want or need to care about this.  Just change `GetTargets` here to `yield break`.

Turupawn:
I think I just solved (2) by letting `GetTargets()` return all the targets on the `Activity` queue. What do you guys think?

Turupawn:
Sorry guys I wasn't able to keep up with the latest changes. Is it ok if we close this PR and continue on #11998 which is working just fine?

abcdefg30:
Superseeded by #11998 then I guess.

It would be better though if you just pushed to the current PR instead of creating new ones. It's not really severe though. If you need/want help with that feel free to join our IRC channel #openra (http://webchat.freenode.net/?channels=#openra).

Turupawn:
Thanks @abcdefg30!

PR #11800, opened 2016-08-10
Delete the Huntable trait for walls and gates.

### LINKED ISSUE #11760: Enemies try to hunt down walls in Survival01
image

This means the mission won't properly end when the player should loose.

MunWolf:
This closes #11760 by making the walls non huntable, we might want to make some other things non huntable, like trees.

### DIFF
--- a/mods/ra/rules/defaults.yaml
+++ b/mods/ra/rules/defaults.yaml
@@ -5,7 +5,6 @@
 		PlayerExperienceModifier: 1
 	ScriptTriggers:
 	UpgradeManager:
-	Huntable:
 	RenderDebugState:
 
 ^SpriteActor:
@@ -68,6 +67,7 @@
 	Inherits@2: ^GainsExperience
 	Inherits@3: ^IronCurtainable
 	Inherits@4: ^SpriteActor
+	Huntable:
 	DrawLineToTarget:
 	UpdatesPlayerStatistics:
 	Mobile:
@@ -166,6 +166,7 @@
 	Inherits@1: ^ExistsInWorld
 	Inherits@2: ^GainsExperience
 	Inherits@3: ^SpriteActor
+	Huntable:
 	DrawLineToTarget:
 	Health:
 		HP: 25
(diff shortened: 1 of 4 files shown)


WolfGaming:
To elaborate this is because nearly everything inherits from ^ExistsInWorld which defines Huntable, and I don't think huntable is mentioned anywhere else in the yaml definitions so everything is huntable.

Phrohdoh:
I updated the title (directly from the commit message) for searchability and clarity, hope you don't mind.

GraionDilach:
Well, neither defines `MustBeDestroyed`, so I don't think there is any harm in this PR. :+1:

WolfGaming:
I was wondering if we want to do the -Huntable on trees and other actors just to remove a redundant trait, any opinions?

pchote:
Wouldn't it make sense to remove `Huntable` from ExistsInWorld, and then add it only to the defaults that actually want it? (basically only player-owned stuff)

WolfGaming:
I would think so but I didn't want to do it until I got an ok from the peoples, and if someone wants huntable to be on trees/walls they can always add them back in the map.yaml that needs it.

WolfGaming:
PR updated.

mods/ra/rules/defaults.yaml
```diff
@@ -296,6 +297,7 @@
 		MaxMoveDelay: 750
 
 ^ArmedCivilian:
+	Huntable:
```
GraionDilach: You don't need this, `^ArmedCivilian` is a template extension of `CivInfantry`, and every user of this template will have the `Huntable` from `Soldier` already.

Mailaender:
:+1: This is less fragile. Compare https://github.com/OpenRA/OpenRA/pull/2855.

obrakmann:
Could I convince you to do the same for the other mods?

WolfGaming:
@obrakmann It is done.

obrakmann:
Thanks. Looks fine, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/b08e3a68023f9f8d395dcff7cc5f18a27e2f110a)

PR #11807, opened 2016-08-11
Clearup mod AIs.

### LINKED ISSUE #11797: AI gets bored and stops spending money
I have found that in most games the AI will eventually just stop spending money. I started an Infantry Only tech level skirmish to test a few things. The computer worked for a while and eventually stopped. There's 2 barracks, 4 ore refineries, lots of silos, and 12 infantry (two of them just outside the screencap area. 

infantry

This is a pitiful force which could be easily overcome with some infantry. The computer never explores, attacks, or continues building.

GraionDilach:
The percentages might sound lame but having the summary of the riflemen and rocket infantry above 100 will make sure no techlevel gets stuck - excluding the RA Naval AI, which doesn't even use infantry at all (notsure if it should be allowed to spam those two).

Fixes #11797.

### DIFF
--- a/mods/d2k/rules/ai.yaml
+++ b/mods/d2k/rules/ai.yaml
@@ -48,14 +48,11 @@ Player:
 			upgrade.hightech: 0.1%
 		UnitsToBuild:
 			carryall: 1%
-			light_inf: 6%
-			trooper: 5%
-			medic: 1%
-			fremen: 0.5%
-			sardaukar: 1.5%
-			saboteur: 0.5%
+			light_inf: 65%
+			trooper: 40%
+			sardaukar: 20%
 			harvester: 1%
-			grenadier: 1%
+			grenadier: 20%
 			trike.starport: 5%
 			quad.starport: 7.5%
 			siege_tank.starport: 5%
@@ -165,13 +162,10 @@ Player:
 			upgrade.hightech: 0.1%
 		UnitsToBuild:
 			carryall: 1%
-			light_inf: 2%
-			trooper: 2%
-			medic: 0.5%
-			fremen: 0.25%
-			sardaukar: 1%
-			saboteur: 0.5%
-			grenadier: 1%
+			light_inf: 65%
+			trooper: 40%
+			sardaukar: 20%
+			grenadier: 20%
 			harvester: 1%
 			trike.starport: 7.5%
 			quad.starport: 12.5%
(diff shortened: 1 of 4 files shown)


GraionDilach:
A question of mine - is there a reason why the AI doesn't build commandos in TD/RA? Woulda add it as part of this PR if there isn't.

obrakmann:
Probably the same reason why other "special" actors aren't listed: the AI doesn't know how to work with them. For Tanya and the Commando you could argue that they could work at least half-way. I don't think it would enhance the gameplay much, if any, though.

mods/d2k/rules/ai.yaml
```diff
@@ -48,14 +48,12 @@ Player:
...
 			carryall: 1%
-			light_inf: 6%
-			trooper: 5%
-			medic: 1%
-			fremen: 0.5%
-			sardaukar: 1.5%
-			saboteur: 0.5%
+			light_inf: 65%
+			trooper: 40%
+			thumper: 1%
```
obrakmann: I'd leave those out, the AI doesn't know what to do with them (ie., how and where to deploy them).

GraionDilach:
My main reason is that in RA the Allied AI only use the two basic infantry tbh. Guess I should remove thumpers as well then, since the AI cannot deploy those..

GraionDilach:
Dropped the medic-to-thumper rename commit and removed the now-defunct medic entries in the last commit.

obrakmann:
Well, I think medics and mechanics *could* work for the AI, since they just use a weapon now. It might be worth testing.

GraionDilach:
After a test, it appears #8768 and related made the AI enough effective at handling medics/mechanics to add them. (They always used a weapon.) Added them to Normal and Turtle AI.

Also unitlimited the engineerspam of the Rush AIs.

Rebased and squashed the commits to one per mod.

GraionDilach:
Forgot to update TS. >_>

WolfGaming:
This looks good, will test it later tonight.

WolfGaming:
This works, the ai no longer gets bored though they are still shitty so :+1: from me.

obrakmann:
Looks ok, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ca96c360473ec56c1e4557a68220ceeee6f02b02)

PR #11809, opened 2016-08-11
Fix Lua API returning wrong player info

### LINKED ISSUE #11808: Player field "Team" returns 0 no matter what team the player is on.
If you try to get the team a player is on using Lua you will have 0 returned no matter what team the player is actually on.

obrakmann:
The Lua API would return wrong information on player fields such as the team or faction when it was changed in the lobby. It referenced fields on the PlayerReference, which are pretty much read-only and don't get changed by the lobby.

screenshot from 2016-08-11 20 21 51

Fixes #11808.

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Properties/PlayerProperties.cs
+++ b/OpenRA.Mods.Common/Scripting/Properties/PlayerProperties.cs
@@ -39,18 +39,25 @@ namespace OpenRA.Mods.Common.Scripting
 			get
 			{
 				Game.Debug("The property `PlayerProperties.Race` is deprecated! Use `PlayerProperties.Faction` instead!");
-				return Player.PlayerReference.Faction;
+				return Player.Faction.InternalName;
 			}
 		}
 
 		[Desc("The player's faction.")]
-		public string Faction { get { return Player.PlayerReference.Faction; } }
+		public string Faction { get { return Player.Faction.InternalName; } }
 
 		[Desc("The player's spawnpoint ID.")]
 		public int Spawn { get { return Player.SpawnPoint; } }
 
 		[Desc("The player's team ID.")]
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Scripting/Properties/PlayerProperties.cs
```diff
@@ -39,18 +39,25 @@ public string Race
...
 			{
 				Game.Debug("The property `PlayerProperties.Race` is deprecated! Use `PlayerProperties.Faction` instead!");
-				return Player.PlayerReference.Faction;
+				return Player.Faction.InternalName;
 			}
 		}
 
 		[Desc("The player's faction.")]
-		public string Faction { get { return Player.PlayerReference.Faction; } }
+		public string Faction { get { return Player.Faction.InternalName; } }
```
obrakmann: I chose the internal name here because that's what's in `PlayerReference.Faction`. However, for scripting/display purposes, it might be nicer to have the real faction's name. Do we want to break backward compatibility here? Add a second property?

Currently, this property isn't used at all, at least by our shipped maps.

//cc @abcdefg30

Mailaender:
:+1: Looks like this was initially written with single-player missions in mind. Thanks for fixing this.

OpenRA.Mods.Common/Scripting/Properties/PlayerProperties.cs
pchote: "real name" will break if we ever introduce translations, so lets keep the internal name.

Phrohdoh:
👍

WolfGaming:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/201827560617981996641d39a839c7c9936173f9)

PR #11814, opened 2016-08-12
Ensure HackyAI tries to attack and capture as frequently as it should.

RoosterDragon:
Returning early in AssignRolesToIdleUnits would skip ticking down the counters that trigger new attack and capture attempts. This means they would be attempted far less often than intended.

Fixes https://github.com/OpenRA/OpenRA/pull/11417#discussion_r72907137.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -659,20 +659,19 @@ namespace OpenRA.Mods.Common.AI
 					s.Update();
 			}
 
-			if (--assignRolesTicks > 0)
-				return;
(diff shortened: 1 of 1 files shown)


WolfGaming:
This is also probably why Normal ai never seems to attack at all.

Phrohdoh:
👍

GraionDilach:
:+1:

Phrohdoh:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/6564cc57f9b7bad7adad343bb8f19d26aa87b36c...b420470f61af318ccc2bf34017180fca50a4d4ad)

PR #11815, opened 2016-08-12
Unit order viewer (WIP)

MunWolf:
This is a UI widget that allows units to issue special orders, traits are going to have to add those orders to the ResolveOrders method for it to do anything.

Todo list:
- [x] Trait that describes an Order with ui info
- [x] Widget that looks for OrderProvider traits and displays them.
- [ ] Finalize the Chrome
  - [ ] RA
  - [ ] CnC
  - [ ] D2K
  - [ ] TS
- [ ] Make Icons for all the orders
- [ ] Allow Mouse interaction
- [x] Allow Hotkey interaction
- [ ] Display Order interaction
- [ ] Change IOrderTargeters and Traits to fully support UIOrders.
- [ ] Add Order definitions to yaml where needed
- [ ] Show tooltip in the Order UI with name and description of what the order does and hotkey.
- [ ] Add support for instant orders that take effect as soon as you hit the button in the ui.


OpenRA.Game/Selection.cs
```diff
@@ -14,19 +14,29 @@
 using OpenRA.Graphics;
 using OpenRA.Primitives;
 using OpenRA.Traits;
+using System;
 
 namespace OpenRA
 {
 	public class Selection
 	{
+		readonly List<Action<Selection>> onSelectionChanged = new List<Action<Selection>>();
```
pchote: This should be a `public event Action` for consistency with our other event handlers.

OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -353,6 +353,7 @@ public interface Requires<T> where T : class, ITraitInfoInterface { }
 	public interface UsesInit<T> : ITraitInfo where T : IActorInit { }
 
 	public interface INotifySelected { void Selected(Actor self); }
+	public interface INotifyDeselected { void Deselected(Actor self); }
```
pchote: This could live as a method inside `INotifySelected`.  It doesn't make much sense for something to implement one but not the other.

OpenRA.Mods.Common/Traits/OrderProvider.cs
```diff
@@ -0,0 +1,51 @@
+﻿using OpenRA.Traits;
```
pchote: File header.

OpenRA.Mods.Common/Traits/OrderProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	// TODO: We don't need this interface if we could find multiples of traits on actors.
```
pchote: What's going on here?  This works - we do it in many times through the engine and common mod code.

OpenRA.Mods.Common/Traits/OrderProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+		public OrderProvider(Actor self, OrderProviderInfo info)
+		{
+			this.self = self;
+			Info = info;
+		}
+
+		public void IssueOrder()
+		{
+			foreach (var resolver in self.TraitsImplementing<IResolveOrder>())
+				resolver.ResolveOrder(self, new Order(Info.OrderCode, self, false));
```
pchote: This will desync the game in multiplayer.  You need to issue orders properly.

pchote:
You shouldn't be modifying the chrome png files directly - they are exported from source psd/svg files.

OpenRA.Mods.Common/Traits/OrderProvider.cs
WolfGaming: When I do it only with OrderProvider  and not IOrderProvider  I don't get any results.

WolfGaming:
This currently does not work btw, still working on issuing orders correctly.

WolfGaming:
We are now issuing orders correctly, so this works except for the items in the initial post.

Phrohdoh:
One nit you may want to try to knock out early:

This is causing our long-standing bug where viewport navigation (arrow keys) appear to be 'stuck'.
Typically this happens when a widget is opened while a key is down, but in this PR it happens whenever I use the arrow keys.

Looking awesome so far!

WolfGaming:
I might split the UIOrderManager code into a different PR, this pr is getting bigger and bigger.

WolfGaming:
@Phrohdoh That bug has now been fixed, nice catch.

WolfGaming:
If people could test this in multiplayer to check for sync errors and other bugs that would be very appreciated, I already did some local testing between two clients but the more testing that is done the better I'll feel :)

pchote:
Agreed.  Please open new smaller scoped PRs if you want to get back into this.

PR #11816, opened 2016-08-12
Fix UpgradeOnDamage to apply if initial DamageState is valid

reaperrr:
As pointed out by Nolt on IRC, on bleed upgrades don't work even if the initial `DamageState` is valid, until the actor's damage state changes.
However, modders might want to use the trait to apply upgrades while the actor is undamaged, and there might be maps with actors starting at valid damage states, too.

This PR makes the trait perform an additional check for a valid `DamageState` at creation.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamage.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamage.cs
@@ -14,7 +14,7 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("Applies an upgrade to the actor at specified damage states.")]
-	public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>
+	public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>, Requires<HealthInfo>
 	{
 		[UpgradeGrantedReference, FieldLoader.Require]
 		[Desc("The upgrades to grant.")]
@@ -35,16 +35,36 @@ namespace OpenRA.Mods.Common.Traits
 		public object Create(ActorInitializer init) { return new UpgradeOnDamage(init.Self, this); }
 	}
 
-	public class UpgradeOnDamage : INotifyDamageStateChanged
+	public class UpgradeOnDamage : INotifyDamageStateChanged, INotifyCreated
 	{
 		readonly UpgradeOnDamageInfo info;
 		readonly UpgradeManager um;
+		readonly Health health;
 		bool granted;
 
 		public UpgradeOnDamage(Actor self, UpgradeOnDamageInfo info)
 		{
 			this.info = info;
 			um = self.Trait<UpgradeManager>();
+			health = self.Trait<Health>();
+		}
+
+		void INotifyCreated.Created(Actor self)
+		{
+			GrantUpgradeOnValidDamageState(self, health.DamageState);
+		}
+
+		void GrantUpgradeOnValidDamageState(Actor self, DamageState state)
+		{
+			if (!info.ValidDamageStates.HasFlag(state))
+				return;
+
+			granted = true;
+			var rand = Game.CosmeticRandom;
+			var sound = info.EnabledSounds.RandomOrDefault(rand);
+			Game.Sound.Play(sound, self.CenterPosition);
+			foreach (var u in info.Upgrades)
+				um.GrantUpgrade(self, u, this);
 		}
 
 		void INotifyDamageStateChanged.DamageStateChanged(Actor self, AttackInfo e)
@@ -52,18 +72,12 @@ namespace OpenRA.Mods.Common.Traits
 			if (granted && info.GrantPermanently)
 				return;
 
-			var rand = Game.CosmeticRandom;
-			if (!granted && info.ValidDamageStates.HasFlag(e.DamageState) && !info.ValidDamageStates.HasFlag(e.PreviousDamageState))
-			{
-				granted = true;
-				var sound = info.EnabledSounds.RandomOrDefault(rand);
-				Game.Sound.Play(sound, self.CenterPosition);
-				foreach (var u in info.Upgrades)
-					um.GrantUpgrade(self, u, this);
-			}
+			if (!granted && !info.ValidDamageStates.HasFlag(e.PreviousDamageState))
+				GrantUpgradeOnValidDamageState(self, health.DamageState);
 			else if (granted && !info.ValidDamageStates.HasFlag(e.DamageState) && info.ValidDamageStates.HasFlag(e.PreviousDamageState))
 			{
 				granted = false;
+				var rand = Game.CosmeticRandom;
 				var sound = info.DisabledSounds.RandomOrDefault(rand);
 				Game.Sound.Play(sound, self.CenterPosition);
 				foreach (var u in info.Upgrades)
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamage.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamage.cs
@@ -44,14 +44,11 @@ namespace OpenRA.Mods.Common.Traits
 		public UpgradeOnDamage(Actor self, UpgradeOnDamageInfo info)
 		{
 			this.info = info;
-			um = self.TraitOrDefault<UpgradeManager>();
+			um = self.Trait<UpgradeManager>();
 		}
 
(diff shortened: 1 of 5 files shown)


OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamage.cs
```diff
@@ -35,18 +35,37 @@ public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
 
 		public UpgradeOnDamage(Actor self, UpgradeOnDamageInfo info)
 		{
 			this.info = info;
 			um = self.TraitOrDefault<UpgradeManager>();
 		}
 
+		void INotifyCreated.Created(Actor self)
+		{
+			health = self.TraitOrDefault<Health>();
```
pchote: Can we put this in the constructor and add `Requires<HealthInfo>`?  It makes zero sense for an actor without Health to implement this trait.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamage.cs
```diff
@@ -35,18 +35,37 @@ public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
+
+			if (health == null)
+				throw new YamlException("Actor type '" + self.Info.Name + "': UpgradeOnDamage requires a Health trait!");
```
pchote: You can and should then drop this.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamage.cs
```diff
@@ -35,18 +35,37 @@ public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
+
+			if (!granted && info.ValidDamageStates.HasFlag(health.DamageState))
```
pchote: This bit probably should stay inside `Created`.

reaperrr:
Updated.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamage.cs
```diff
@@ -35,23 +35,35 @@ public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
+			um = self.Trait<UpgradeManager>();
+			health = self.Trait<Health>();
 		}
 
-		void INotifyDamageStateChanged.DamageStateChanged(Actor self, AttackInfo e)
+		void INotifyCreated.Created(Actor self)
 		{
-			if (um == null)
-				return;
+			if (!granted && info.ValidDamageStates.HasFlag(health.DamageState))
```
pchote: It would be nice to avoid duplicating code here.  Can we get away with calling `DamageStateChanged` directly from here? (note: you'll need to pull out a common private method that both this and `INotifyDamageStateChanged.DamageStateChanged` call)

reaperrr: To be honest, I don't follow, at least not completely. Do you just mean to put the granting stuff into its own method and then just call that method here and in `INotifyDamageStateChanged.DamageStateChanged`, or should `Created` trigger a fake `DamageStateChanged` (pretending it changed)?

GraionDilach: Yeah, I guess he means the former one.

reaperrr:
Updated.
Additionally, renamed trait to `UpgradeOnDamage__State__`, since it supports `Undamaged` and no longer requires the damage state to change to be triggered.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamageState.cs
```diff
@@ -32,41 +32,55 @@ public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
-				return;
+			if (!granted)
+				GrantUpgradeOnValidDamageState(self, health.DamageState);
+		}
 
-			var rand = Game.CosmeticRandom;
-			if (!granted && info.ValidDamageStates.HasFlag(e.DamageState) && !info.ValidDamageStates.HasFlag(e.PreviousDamageState))
+		void GrantUpgradeOnValidDamageState(Actor self, DamageState state)
+		{
+			if (info.ValidDamageStates.HasFlag(state))
```
WolfGaming: Isn't the convention to do a not in front and return instead of putting it inside the if here.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamageState.cs
```diff
@@ -32,41 +32,55 @@ public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
+
+		void INotifyDamageStateChanged.DamageStateChanged(Actor self, AttackInfo e)
+		{
+			if (granted && info.GrantPermanently)
+				return;
+
+			if (!granted && !info.ValidDamageStates.HasFlag(e.PreviousDamageState))
+			{
+				GrantUpgradeOnValidDamageState(self, health.DamageState);
+			}
```
WolfGaming: Unneeded curly brackets.

reaperrr: I was a bit torn between convention and readability here (due to the else if below).

WolfGaming: Well I'll leave the final say to @pchote but I have never had these brackets even with the else if,
except when there is an if inside that if.

Phrohdoh: I don't see how these brackets improve readability.

reaperrr:
Updated.
I think modders working with releases should get the fix and new name "out of the box", so I'm promoting this to release milestone to make sure it gets merged in time.

WolfGaming:
I really feel like those if statements could be made simpler but I don't have any ideas atm.
0,5 :+1: from me until I'm done testing this tonight.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamageState.cs
```diff
@@ -32,41 +32,53 @@ public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
+	public class UpgradeOnDamageState : INotifyDamageStateChanged, INotifyCreated
 	{
-		readonly UpgradeOnDamageInfo info;
+		readonly UpgradeOnDamageStateInfo info;
 		readonly UpgradeManager um;
+		readonly Health health;
 		bool granted;
 
-		public UpgradeOnDamage(Actor self, UpgradeOnDamageInfo info)
+		public UpgradeOnDamageState(Actor self, UpgradeOnDamageStateInfo info)
```
obrakmann: Do we want an upgrade rule for this? This trait has never been released, but it seems that there are already people using it.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamageState.cs
```diff
@@ -32,41 +32,53 @@ public class UpgradeOnDamageInfo : ITraitInfo, Requires<UpgradeManagerInfo>
...
 			this.info = info;
-			um = self.TraitOrDefault<UpgradeManager>();
+			um = self.Trait<UpgradeManager>();
+			health = self.Trait<Health>();
 		}
 
-		void INotifyDamageStateChanged.DamageStateChanged(Actor self, AttackInfo e)
+		void INotifyCreated.Created(Actor self)
+		{
+			if (!granted)
```
obrakmann: Isn't this check redundant? I don't see a way how `granted` could become true before `Created` runs.

obrakmann:
Looks otherwise fine to me, :+1:

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnDamageState.cs
reaperrr: I guess it can't hurt.

reaperrr:
Updated. Ignore the messed-up commit order, that's just github acting funny again.

Mailaender:
:+1:

obrakmann:
The test-case can be removed now.

reaperrr:
Removed test-case.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ca70eb5043fb6a61ea094546e50dbaa4eab3608f)

PR #11821, opened 2016-08-15
Fix for System Overflow exception

UnknownProgrammer:
caused by casting a greater than int.MaxValue decimal to int. Fix limits offset to int.MaxValue respectively int.MinValue.

### DIFF
--- a/OpenRA.Game/WPos.cs
+++ b/OpenRA.Game/WPos.cs
@@ -63,10 +63,11 @@ namespace OpenRA
 			if (pitch.Angle == 0)
 				return ret;
 
-			// Add an additional quadratic variation to height
-			// Uses decimal to avoid integer overflow
-			var offset = (int)((decimal)(b - a).Length * pitch.Tan() * mul * (div - mul) / (1024 * div * div));
-			return new WPos(ret.X, ret.Y, ret.Z + offset);
+			var tempoffset = (decimal)(b - a).Length * pitch.Tan() * mul * (div - mul) / (1024 * div * div);
+
+			var offset = (int)Math.Min(Math.Max((decimal)int.MinValue, tempoffset + (decimal)ret.Z), (decimal)int.MaxValue);
+
+			return new WPos(ret.X, ret.Y, offset);
 		}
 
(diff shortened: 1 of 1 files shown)


WolfGaming:
This looks to be sane though we should really do Lint checking for this too so modders know that they aren't getting what they expect.

GraionDilach:
:+1:

pchote:
Looks fine, but please restore the comment and remove the blank lines between the variables.

OpenRA.Game/WPos.cs
```diff
@@ -63,10 +63,11 @@ public static WPos LerpQuadratic(WPos a, WPos b, WAngle pitch, int mul, int div)
 			if (pitch.Angle == 0)
 				return ret;
 
-			// Add an additional quadratic variation to height
-			// Uses decimal to avoid integer overflow
-			var offset = (int)((decimal)(b - a).Length * pitch.Tan() * mul * (div - mul) / (1024 * div * div));
-			return new WPos(ret.X, ret.Y, ret.Z + offset);
+			var tempoffset = (decimal)(b - a).Length * pitch.Tan() * mul * (div - mul) / (1024 * div * div);
```
pchote: Our code style is to use camel case for variable names, so `var tempOffset`

OpenRA.Game/WPos.cs
```diff
@@ -63,10 +63,11 @@ public static WPos LerpQuadratic(WPos a, WPos b, WAngle pitch, int mul, int div)
...
+
+			var offset = (int)Math.Min(Math.Max((decimal)int.MinValue, tempoffset + (decimal)ret.Z), (decimal)int.MaxValue);
```
pchote: You may be able to simplify this to `tempOffset.Clamp((decimal)int.MinValue, (decimal)int.MaxValue);` instead of hand-rolling the checks.

pchote:
@UnknownProgrammer: It is good practice to leave a comment when you push a fix, otherwise we don't know to re-review it!

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/b19974b30b46b56f21d484fda3af374083a1c19a...650bfddccc01bedf992d4e6953d0ad9458880b8b)

PR #11830, opened 2016-08-18
Make Armament as overridable as possible

reaperrr:
To allow downstream mods to ship customised `Armament` derivatives without breaking backwards-compatibility with `Common`.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Armament.cs
+++ b/OpenRA.Mods.Common/Traits/Armament.cs
@@ -107,8 +107,8 @@ namespace OpenRA.Mods.Common.Traits
 		List<Pair<int, Action>> delayedActions = new List<Pair<int, Action>>();
 
 		public WDist Recoil;
-		public int FireDelay { get; private set; }
-		public int Burst { get; private set; }
+		public int FireDelay { get; protected set; }
+		public int Burst { get; protected set; }
 
 		public Armament(Actor self, ArmamentInfo info)
 			: base(info)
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Armament.cs
```diff
@@ -264,7 +264,7 @@ public Barrel CheckFire(Actor self, IFacing facing, Target target)
 		public bool ShouldExplode(Actor self) { return !IsReloading; }
```
WolfGaming: You don't want this to be virtual too?

GraionDilach: That's coming from an interface tbh.

GraionDilach:
I'm fine with this, but may you link your use-case example, please?

reaperrr:
> I'm fine with this, but may you link your use-case example, please?

I can't link anything playable for now (I'd have to upload the content somewhere, which I want to avoid as long as possible), but I could push the code later.

Essentially, I'm using a custom `WeaponSlot` derived from `Armament` to do a few extra things,
- linking weapons to a specific `HeatPool` trait to which they contribute heat per burst
- implementing customisable `TargetPositions` in a way that is good enough for me, without having to wait months until that maybe makes its way into OpenRA

GraionDilach:
Was only interested in the code example. The first reasons sounds enough for me. :+1:

OpenRA.Mods.Common/Traits/Armament.cs
Phrohdoh: Could this be farmed off to a public virtual method that the (explicit) interface impl wraps?

reaperrr:
Updated.
- made `ShouldExplode` interface explicit
- made `IsReloading` and `OutOfAmmo` virtual
- added virtual `AllowExplode`, which `ShouldExplode` wraps

abcdefg30:

@GraionDilach your 👍 still stands after the recent changes?

GraionDilach:
Yeah, I think so.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1bee32d04a775223189311c4a32fd0227433bd0f...d35f606b0a72147035b497264e7ad693d93d6398)

PR #11832, opened 2016-08-18
Fix 'make dependencies' choking on Open.NAT directory name

obrakmann:
The directory name appears to have changed in the package for 2.1.0.

### DIFF
--- a/thirdparty/fetch-thirdparty-deps.sh
+++ b/thirdparty/fetch-thirdparty-deps.sh
@@ -85,9 +85,9 @@ fi
 
 if [ ! -f Open.Nat.dll ]; then
 	echo "Fetching Open.Nat from NuGet"
-	get Open.Nat 2.1.0
-	cp ./Open.Nat/lib/net45/Open.Nat.dll .
-	rm -rf Open.Nat
+	get Open.NAT 2.1.0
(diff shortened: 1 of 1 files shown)


Phrohdoh:
👍

Mailaender:
It is still `thirdparty/download/Open.Nat/lib/net45` when using the `noget.sh` so for me this change breaks things.

thirdparty/fetch-thirdparty-deps.sh
```diff
@@ -86,8 +86,8 @@ fi
 if [ ! -f Open.Nat.dll ]; then
 	echo "Fetching Open.Nat from NuGet"
 	get Open.Nat 2.1.0
```
Mailaender: You need to change this to `get Open.NAT 2.1.0` as well.

obrakmann:
Ah, I see. Updated.

Mailaender:
Thanks.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d024b56d0212b6bc21840d7986a981834ad503ec...943334b019fec31d9a95aca7b191325980739520)

PR #11836, opened 2016-08-19
Add upgrade support to CashTrickler.

### LINKED ISSUE #11835: Feature request: add upgrades to Cashtrickler
as the title.

GraionDilach:
Resolves #11835.

### DIFF
--- a/OpenRA.Mods.Common/Traits/GiveCashOnCapture.cs
+++ b/OpenRA.Mods.Common/Traits/GiveCashOnCapture.cs
@@ -1,4 +1,4 @@
-#region Copyright & License Information
+﻿#region Copyright & License Information
 /*
  * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
  * This file is part of OpenRA, which is free software. It is made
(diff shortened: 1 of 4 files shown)


abcdefg30:
👍 

Do we also want to make `CaptureAmount` optionally disable-able?

GraionDilach:
Oh, good point, that needs a fixup.

abcdefg30:
Just to clear out what I mean: IMHO we should add another boolean to let the modder decide if `CaptureAmount` gets payed out when the trait is disabled.

pchote:
Disabled should mean disabled, so that shouldn't need its own boolean.

If there *is* a usecase for that, then it would be better to split out a separate `GiveCashOnCapture` trait that can on its own be toggled using upgrades.

GraionDilach:
There is a usecase - the RA2 Oil Derrick only gave the initial amount to the first player which captured it, not every time.

pchote:
I'm not sure how that's related to upgrades/disabling?

GraionDilach:
Well, one can have a global upgrade defined on the player actor to upgrade the CashTrickler on Oil Derricks from 0 Amount/1000 CaptureAmount to x Amount/0 CaptureAmount as a solution to that problem.

pchote:
Right now this pr isn't suitable to merge because it doesn't completely disable the `CashTrickler`.  You can fix that by adding `IsTraitDisabled` to the `if (info.CaptureAmount > 0)` check.

You argue that you want to be able to control the initial bonus separately to the trickle behaviour.   This makes sense, because they are completely different behaviours (the only logic shared between the two parts is the `ShowTicks` parameter).  If you want to support that then the proper fix is to split that initial bonus out into its own `UpgradableTrait`.  The two traits are then self-contained, and can be individually toggled.

GraionDilach:
Okay, then replace fixup with change until I split the trait.

GraionDilach:
Updated. This probably revokes both labels.

OpenRA.Mods.Common/Traits/GiveCashOnCapture.cs
```diff
@@ -0,0 +1,56 @@
...
+ */
+#endregion
+
+using OpenRA.Mods.Common.Effects;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	[Desc("Lets the actor grant cash when captured.")]
+	class GiveCashOnCaptureInfo : UpgradableTraitInfo
```
pchote: This will want to be `Gives` for consistency with `GivesBounty` and `GivesExperience`.

OpenRA.Mods.Common/Traits/GiveCashOnCapture.cs
```diff
@@ -0,0 +1,56 @@
...
+	{
+		[Desc("Whether to show the cash tick indicators (+$15 rising from actor).")]
```
pchote: "+$15" is awfully arbitrary. How about "Whether to show the cash tick indicators rising from the actor." (in both traits).

OpenRA.Mods.Common/Traits/GiveCashOnCapture.cs
```diff
@@ -0,0 +1,56 @@
...
+		public readonly bool ShowTicks = true;
+		[Desc("Amount of money awarded for capturing the actor.")]
```
pchote: Style nit: newlines between fields please.

pchote (commented):
I haven't tested these changes, but this all looks correct.  See a few minor style nits below.

MustaphaTR:
Needs Rebase.

reaperrr:
@GraionDilach Any chance of a near-term update?

pchote:

reaperrr:
Superseded by #12358.

PR #11787, opened 2016-08-07
Added the Tiberian Sun: Firestorm assets

Mailaender:
Depends on:
- [ ] ~~https://github.com/OpenRA/OpenRAWeb/pull/307~~
- [x] https://github.com/OpenRA/OpenRA/pull/11771

PR #11839, opened 2016-08-20
Added the GDI Juggernaut to Tiberian Sun 

Mailaender:
Depends on / testcase for https://github.com/OpenRA/OpenRA/pull/11787.

### DIFF
--- a/mods/ts/installer/downloads.yaml
+++ b/mods/ts/installer/downloads.yaml
@@ -19,3 +19,7 @@ music: Freeware Music
 	MirrorList: http://www.openra.net/packages/ts-music-mirrors.txt
 	Extract:
 		^Content/ts/scores.mix: scores.mix
+expansion: Expansion Freeware Content
+	MirrorList: http://www.openra.net/packages/fs-mirrors.txt
+	Extract:
+		^Content/ts/expand01.mix: expand01.mix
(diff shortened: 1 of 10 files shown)


pchote:
Can we please make this use upgrades like the tick tank and mobile sensor array?

Mailaender:
See https://github.com/OpenRA/OpenRA/blob/bleed/mods/ts/rules/nod-vehicles.yaml#L192 after which this is modeled. I guess the Nod artillery not using upgrades has a reason.

GraionDilach:
The Nod arty setup predates the Tick Tank/MSA setup. There shouldn't be a reason actually, AttackTurreted can be upgraded to work only in deployed mode.

My question is, will we ship FS as TS or FS will be a techlevel or something akin? (Due to tileset updates and whatnot.)

Mailaender:
> There shouldn't be a reason actually, AttackTurreted can be upgraded to work only in deployed mode.

I am bit more worried about the rendering which needs to change completely.

reaperrr:
> I am bit more worried about the rendering which needs to change completely. Upgrades usually break and are very fiddly to work with in complicated setups.

You just need to replicate the Tick Tank setup. I haven't noticed any breakage or other problems with the tick tank so far.

reaperrr:
Except for the sequences this needs a major rewrite, so I'm taking over and will file an updated PR once #12429 is merged.

Mailaender:
Okay, thanks.

PR #11841, opened 2016-08-20
Added the Nod Cyborg Reaper

Mailaender:
Depends on https://github.com/OpenRA/OpenRA/pull/11787 and https://github.com/OpenRA/OpenRA/pull/11624 as well as something that fixes 

```
 Actor has multiple traits of type `OpenRA.Mods.Common.Traits.Render.RenderSprites
```

for webguy to work.

### DIFF
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -77,6 +77,25 @@
 		UpgradeTypes: notmobile
 		UpgradeMaxEnabledLevel: 0
 
+^WebDisable:
+	WithInfantryBody@WebDisable:
+		UpgradeTypes: WebDisable
+		UpgradeMinEnabledLevel: 1
+		MoveSequence: struggle
+		StandSequences: struggle
+	DisableOnUpgrade@WebDisable:
+		UpgradeTypes: WebDisable
+		UpgradeMinEnabledLevel: 1
+	TimedUpgradeBar@WebDisable:
+		Upgrade: WebDisable
+		Color: FFFFFF
+	Mobile:
+		UpgradeTypes: WebDisable
+		UpgradeMaxEnabledLevel: 0
+	WithInfantryBody:
+		UpgradeTypes: WebDisable
+		UpgradeMaxEnabledLevel: 0
+
 ^Cloakable:
 	Cloak@CLOAKGENERATOR:
 		UpgradeTypes: cloakgenerator
@@ -240,6 +259,7 @@
 	Inherits@2: ^ExistsInWorld
 	Inherits@3: ^SpriteActor
 	Inherits@4: ^Cloakable
+	Inherits@5: ^WebDisable
 	Huntable:
 	DrawLineToTarget:
 	Health:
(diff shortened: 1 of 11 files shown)


reaperrr:
> Actor has multiple traits of type `OpenRA.Mods.Common.Traits.Render.RenderSprites

Unless I'm overlooking something, I'd say it would be way smarter to add webguy as sequence to all infantry and then use upgrades to disable `WithInfantryBody` and enable a `WithSpriteBody` that displays the webguy sequence by default.

We don't need to make `RenderSprites` upgradable for this.

Mailaender:
You are right. It now works pretty cool. If you order your soldier to move, he will start to struggle.

mods/ts/rules/defaults.yaml
```diff
@@ -240,6 +256,7 @@
 	Inherits@2: ^ExistsInWorld
 	Inherits@3: ^SpriteActor
 	Inherits@4: ^Cloakable
+	Inherits@4: ^WebDisable
```
reaperrr: This needs to be `@5`!

reaperrr:
> You are right. It now works pretty nice. If you order your soldier to move, he will start to struggle.

No reason to get sarcastic when it's your fault.
I said With**Sprite**Body for a reason, not by mistake.

Also, you need to set
```
WithInfantryBody:
	UpgradeTypes: WebDisable
	UpgradeMaxEnabledLevel: 0
```
on `^Infantry`, not `^Soldier`, civilians and cyborgs were web-able as well in the original.

But more importantly:
The Cyborg Reaper was a *unit* in Firestorm, not an infantry, which makes sense since it's more mechanical than biological, and we should keep it that way.
That would also save you the hassle of disabling all kinds of inherited traits. Set it up the same way like the Wolverine/SMECH (except that you need to add the `WithDeathAnimation`).

Another thing: You need to add the `webguy` sequence to more than just the `^BasicInfantry`, unfortunately there were too many deviations for some of the special infantry, so medic, both cyborgs, jumpjet, and weedguy don't inherit it.

Mailaender:
> No reason to get sarcastic when it's your fault.
I wasn't being sarcastic. Rather happy for you good suggestion.

reaperrr:
> I wasn't being sarcastic. Rather happy for you good suggestion. Don't spoil it.

Sorry, I misunderstood. I thought you had already updated the branch and it still wasn't working.

mods/ts/weapons/missiles.yaml
```diff
@@ -294,3 +294,30 @@ RedEye2:
 	Warhead@2Eff: CreateEffect
 		Explosions: tiny_twlt
 		ImpactSounds: expnew05.aud
+
+QuadLauncher:
+	ReloadDelay: 180
+	Range: 7c0
+	Report: samshot1.aud
+	ValidTargets: Air
```
ABrandau: Cyborg reapers were also able to attack ground targets, see http://cnc.wikia.com/wiki/Reaper_(Firestorm) , the Ingame paragraph.

Mailaender:
Fixed this up a bit.

mods/ts/rules/defaults.yaml
Mailaender: Weird that it didn't crash.

reaperrr:
This will need a major rewrite, due to the upgrades -> conditions refactor and some rather large inconsistencies with the original FS counterpart, so I'm taking over and will file an updated PR once #12429 is merged.

Mailaender:
Thanks.

PR #11842, opened 2016-08-20
Fix desert tileset tiling issues in RA

### LINKED ISSUE #11829: RA desert tileset regression
There's a tile error on the shellmap even though the shellmaps' `map.bin` didn't change for ages, so it's likely a regression from [#11545](https://github.com/OpenRA/OpenRA/pull/11545/commits/68e176c531b703500e53b7fa32656ed86e126748).

obrakmann:
This is what it looks like on bleed:
screenshot from 2016-08-20 12 42 55

The template that's displayed there is template 80 (both on last release and on bleed). However, compared to last release, the templates 79 and 80 appear to have been swapped. Somehow `falls1.des` and `falls2.des` must have been mixed up when the desert tileset was added to RA.

In TD, `falls1.des` is 3x3 cells and `falls2.des` is 3x2 cells as well, so that fits, [see here](https://github.com/OpenRA/OpenRA/blob/bleed/mods/cnc/tilesets/desert.yaml#L1567-L1593).

The first commit fixes this, and restores the situation from last release.

This is what it used to look like:
screenshot from 2016-08-20 12 43 42

Notice however that that's actually a tiling error, the template is cut off on the right side (the bush on the lower bank is cut in half). The left part is template 80, the cave thing itself is template 76.

I corrected the tiling error like this:
screenshot from 2016-08-20 13 02 23

Fixes #11829.


pchote:
It kind of sucks to have the two tile sets be incompatible in this one specific way.  Could we use an upgrade rule to fix affected maps instead?

obrakmann:
Not sure how you'd fix the `map.bin` from an upgrade rule.

GraionDilach:
:+1:

pchote:
We'd need to implement a new rule type that parses, tweaks, then resaves the file.  Not worth blocking this fix over it, but i still think its the sensible thing to do in the long term.

pchote:
Looks fine to me :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a881944ff7afef33ac243c2d4d1f8377f079e840

PR #11852, opened 2016-08-21
Multiply hitpoints and damage by factor 10 in D2k and 100 in other mods

### LINKED ISSUE #11425: Veterancy on Heavy Armour makes them invincible to riflemen

### LINKED ISSUE #11719: Multiply all damage and health values by 100 or 1000 in the nonD2K mods.
One of the actual benefits of Dune 2000's arbitrarily huge health pools and damage values is to have more bits for the float-integer conversion during actual damage calculation, preventing issues like #11425 to happen.

I think it would be useful to add this D2k trait to the other mods as well (would give more freedom to balancers without triggering rounding-to-0 possibilities).

### LINKED ISSUE #11850: (Bleed) RA/TD Minigunners damage vs Heavy Armor fix threatens balance.
#11460 doubles minigunners' damage vs heavy armor in order to retain damage output against veteran heavy armor. In effect, heavy armor melts in the prescense of infantry blobs making tanks less relevant as infantry support and the whole range of heavy armor units is effectively nerfed. 

The issue came out of riflemen/minigunners doing 0 damage vs veteran heavy armor. #11425

reaperrr:
This should pretty much fix all rounding issues with damage percentages, versus values etc.

Closes #11719.
Fixes #11850.
Properly fixes #11425 (the TD Jeep + Buggy were still affected).

_Note:_ This _will_ somewhat affect balancing, as several weapons' damage versus specific armor types won't be rounded down nearly as much as before, in some cases increasing effective damage by as much as 50% against these armor types (effective damage of some RA/TD miniguns vs. Heavy goes up from 1 to 150 compared to pre-#11460 bleed, for example).

At least a certain amount of testing will have to go into this PR.

### DIFF
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -92,6 +92,12 @@ namespace OpenRA.Mods.Common.UtilityCommands
 			catch { }
 		}
 
+		internal static string MultiplyByFactor(int oldValue, int factor)
+		{
+			oldValue = oldValue * factor;
+			return oldValue.ToString();
+		}
+
 		internal static void UpgradeActorRules(ModData modData, int engineVersion, ref List<MiniYamlNode> nodes, MiniYamlNode parent, int depth)
 		{
 			var addNodes = new List<MiniYamlNode>();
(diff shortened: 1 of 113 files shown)


reaperrr:
Disclaimer: I know the upgrade rule could probably be de-duplicated a bit more, but I don't feel like investing more time into that.

mods/ts/rules/civilian-structures.yaml
```diff
@@ -1254,14 +1254,13 @@ GALITE:
...
 	Health:
-		HP: 600
+		HP: 60000
 	Armor:
 		Type: Wood
 	RevealsShroud:
 		Range: 2c0
 	Power:
 		Amount: 0
-# TODO: should use terrain lighting instead, depends on https://github.com/OpenRA/OpenRA/issues/3605
```
reaperrr: The diff was too large for my gui to show this file completely, so I couldn't discard this change. I'll restore it before this gets merged.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -92,6 +92,12 @@ internal static void TryUpdateHSLColor(ref string value)
 			catch { }
 		}
 
+		internal static string MultiplyByFactor(int oldValue, int factor)
```
pchote: Deciding to split only this part into a function seems like a really weird choice to me.

Please either inline this as one line (`= (oldValue * 10).ToString()`) or move the `var oldValue` and the mod id checks inside here to properly remove the duplication.

pchote:
Can you please also remove the weapon hacks (mainly in superweapons) where we multiply the Falloff by 10 and divide the damage by 10?

pchote:
On second thought... this has the unfortunate and not very nice side-effect that our integer-percentage damage modifiers are now nowhere near precise enough.  We had previously worked around this for superweapons and other high-damage weapons by multiplying the damage falloff by 10 and dividing the base damage to compensate, but this is now going to be a problem everywhere.

GraionDilach:
It's exactly the integer percentages being too precise which brought up the issues leading to this PR. "Not precise enough" shouldn't be an issue, with this PR damage calculation is still more precise than without.

pchote:
Not really, unless you want to sweep it in with

> Note: This will somewhat affect balancing,
> ...
> At least a certain amount of testing will have to go into this PR.

(which I guess is valid enough).

The key point is that this PR only solves part of the problem, so I would very much prefer that we revert the numbers back to the values they had in the last N releases and then address all the problems together in next+1.  Otherwise we're going to lose another month or two on our already far-too-delayed next release, risk more balance regressions, and then need to repeat this all over again in the next cycle anyway.

reaperrr:
Filed #11873 as short-term solution for next release until we've decided how to deal with the other issues.

reaperrr:
I'm closing this for now since I suspect the code-level prerequisites won't happen that soon.

PR #11853, opened 2016-08-21
Remove and document slope-specific tiberium artwork.

pchote:
http://www.ppmforums.com/viewtopic.php?p=549033 explains that tib13-20 are used for slopes, and this can be easily confirmed by checking the shapes with the asset browser.

This PR removes these tiles from the Variants list so that they aren't drawn on flat tiles (which gives gappy looking fields).  It also adds some comments for the future when we decide to add ramp support for resources.


abcdefg30:
Looks simple enough. :+1: / :white_check_mark:

GraionDilach:
:+1:

PR #11855, opened 2016-08-21
Add support for sending messages to certain teams or players with lua

abcdefg30:
This extends the normal `DisplayMessage` to include a team parameter (0 means to everyone).
And secondly adds a new method for sending private messages which accepts a player parameter.

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
@@ -188,14 +188,16 @@ namespace OpenRA.Mods.Common.Scripting
 			return true;
 		}
 
-		[Desc("Display a text message to the player.")]
-		public void DisplayMessage(string text, string prefix = "Mission", HSLColor? color = null)
+		[Desc("Display a text message to all players.")]
+		public void DisplayMessage(string text, string prefix = "Mission", HSLColor? color = null, int team = 0)
 		{
 			if (string.IsNullOrEmpty(text))
 				return;
 
 			Color c = color.HasValue ? HSLColor.RGBFromHSL(color.Value.H / 255f, color.Value.S / 255f, color.Value.L / 255f) : Color.White;
-			Game.AddChatLine(c, prefix, text);
+
+			if (team == 0 || (world.LocalPlayer != null && world.LobbyInfo.ClientWithIndex(world.LocalPlayer.ClientIndex).Team == team))
+				Game.AddChatLine(c, prefix, text);
 		}
 
 		[Desc("Displays a debug message to the player, if \"Show Map Debug Messages\" is checked in the settings.")]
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -188,14 +188,24 @@ public bool PlayMovieInRadar(string movie, LuaFunction playComplete = null)
...
+		public void DisplayMessage(string text, string prefix = "Mission", HSLColor? color = null, int team = 0)
 		{
 			if (string.IsNullOrEmpty(text))
 				return;
 
 			Color c = color.HasValue ? HSLColor.RGBFromHSL(color.Value.H / 255f, color.Value.S / 255f, color.Value.L / 255f) : Color.White;
-			Game.AddChatLine(c, prefix, text);
+			var world = Context.World;
+
+			if (team == 0 || world.LobbyInfo.ClientWithIndex(world.LocalPlayer.ClientIndex).Team == team)
```
obrakmann: `world.LocalPlayer` will be null during replays. If you used `RenderPlayer`, those messages could be seen in replays as well (with the correct player view selected). You'd still need a null check, though.

OpenRA.Mods.Common/Scripting/Global/MediaGlobal.cs
```diff
@@ -188,14 +188,24 @@ public bool PlayMovieInRadar(string movie, LuaFunction playComplete = null)
...
+				Game.AddChatLine(c, prefix, text);
+		}
+
+		[Desc("Display a text message only to one player.")]
+		public void PrivateMessage(Player player, string text, string prefix = "Mission [Private]", HSLColor? color = null)
+		{
+			if (Context.World.LocalPlayer == player)
```
obrakmann: `RenderPlayer` might be more nicer here as well.

obrakmann:
Just out of curiosity: what's your use case?

pchote:
RenderPlayer will break normal gameplay with cheats

abcdefg30:
Included a null check for LocalPlayer. Thanks, somehow I completely forgot about that.

> Just out of curiosity: what's your use case?

It was requested for a Coop-mission (so you don't have to spam everyone with only player-specific instructions).

obrakmann:
> RenderPlayer will break normal gameplay with cheats

FWIW, I care more about the game behaving as expected in replays than when the debug options are enabled.

abcdefg30:
We don't show team-messages in replays as well, so I'm not sure if we should here.

obrakmann:
> It was requested for a Coop-mission (so you don't have to spam everyone with only player-specific instructions).

You can already do that in Lua, though, by checking the `IsLocalPlayer` and `Team` properties. That's how the mission objective notifications in our new multiplayer missions work, for example.

pchote:
If we can already do this with existing code then we shouldn't add a new redundant API method.

abcdefg30:
I guess you can. Derp me.

PR #11856, opened 2016-08-21
Tiedied up d2k building build menu icon order

abcdefg30:
Supersedes #11687.

### DIFF
--- a/mods/d2k/rules/structures.yaml
+++ b/mods/d2k/rules/structures.yaml
@@ -16,7 +16,6 @@
 		RemoveInstead: true
 	Buildable:
 		Queue: Building
-		Prerequisites: construction_yard
 		BuildPaletteOrder: 10
 
 concretea:
@@ -40,7 +39,7 @@ concreteb:
 		Cost: 50
 	Buildable:
 		BuildPaletteOrder: 40
-		Prerequisites: construction_yard, upgrade.conyard
+		Prerequisites: upgrade.conyard
 		BuildDuration: 81
 		BuildDurationModifier: 40
 
@@ -112,7 +111,6 @@ wind_trap:
 	Inherits: ^Building
 	Buildable:
 		Queue: Building
-		Prerequisites: construction_yard
 		BuildPaletteOrder: 20
 		BuildDuration: 180
 		BuildDurationModifier: 40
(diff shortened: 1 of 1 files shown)


GraionDilach:
:+1:

pchote:
Style nit: Please put two line breaks between the two sentences in each of the commit messages, which is the easiest way to bring them in line with the standard git conventions (so they don't break in git log or the github ui)

pchote:
I'm not sure that this new order makes sense.  The layout doesn't match the order that tech is unlocked: e.g. the Palace sits immediately to the left of the IX Lab, which is its only prereq.

penev92:
My two cents:
```
[17:09:15] php_bg: turrets are way too far down to have any meaningful hotkey
[17:09:20] php_bg: palace should be further down
[17:09:38] php_bg: not sure I like the "starport before outpost" idea
[17:10:14] php_bg: IMO move palace & IX lab to the last row
[17:10:22] php_bg: and move the turrets up
[17:10:41] php_bg: also switch the palace and IX lab
[17:11:02] php_bg: as the palace is your "endgame" building, being the "superweapon" and all
```
Also possibly change row 3 to Outpost, Heavy Factory, Starport?

abcdefg30:
Changes:
- Switched Outpost and Starport
- Switched Palace and IX Lab
- Moved turrets above the High Tech Factory

d2kbuildmenu

reaperrr:
Both commits look fine to me.

I faintly remember that making the CYard an explicit prerequisite had to do with upgrades, so we need to make sure there aren't any regressions. 
I couldn't spot any, though, so maybe the original reason for listing it has already been taken care of.

CH4Code:
(My) The previous order made some sense, since the two linear tech paths
pp->barracks->outpost->high tech factory      (and rocket turret)
(pp)->refinery-> lt. factory / hvy. factory -> repair pad
were in line.
I see why turrets must be in front. But please switch starport and outpost back
(they made up a linear tech line too: starport <- outpost <+ hvy factory )
and replace row 4 with (gun-turret, rocket-turret, silo) so silo is in line with ref, and rocket-turret is in line with outpost.
Thank you.

reaperrr:
> I see why turrets must be in front. But please switch starport and outpost back
(they made up a linear tech line too: starport <- outpost <+ hvy factory )
and replace row 4 with (gun-turret, rocket-turret, silo) so silo is in line with ref, and rocket-turret is in line with outpost.

@abcdefg30: The above suggestions by @CH4Code make sense, and this PR has been sitting idle for quite long, would be nice to finally get this in.

reaperrr:
Superseded by #12534.

PR #11858, opened 2016-08-21
Yaml style clean-up

reaperrr:
It annoys me that these come up whenever I apply upgrade rules, I think it's time to actually adress them on bleed.


obrakmann:
:+1:

pchote:
Why have the map.pngs changed? If this is. Windows vs everyone else problem then it just makes things dirty for others.

reaperrr:
> Why have the map.pngs changed? If this is. Windows vs everyone else problem then it just makes things dirty for others.

I don't know.

Maybe someone not on Windows could run upgrade rules on this branch (with something like 20160823 as date, so no real upgrade rules get applied), if the pngs don't change again, they should be fine for all OSes.

FWIW, the PNGs that @abcdefg30 & Co. 'crushed' for singleplayer missions were not changed, so I suspect the PNGs that the upgrade rules touched were saved with some old engine revision or some other 3rd-party software.

pchote:
The singleplayer missions define `LockPreview: True`, which tells the game to never overwrite them.

obrakmann:
FWIW, the `map.png`s don't get flagged here when I run `./OpenRA.Utility.exe ra --upgrade-mod 20160823`, neither on this branch nor on bleed. So I guess that is indeed a windows-only thing

GraionDilach:
These map.pngs gets flagged on bleed on my Kubuntu 14.04.5 x64, but I haven't tested this branch yet.

pchote:
The `map.png`s don't get flagged for me, and the versions here are bigger than those in the repo.

reaperrr:
> The singleplayer missions define `LockPreview: True`, which tells the game to never overwrite them.

Do upgrade rules also 'listen' to `LockPreview: True`?

Anyway, I've removed that commit so we can get at least the style fixes merged quickly.

abcdefg30:
👍

Issue #11749, opened 2016-07-31
[RA] Add features to the Engineer.


SoScared:
It feels like there's more potential to the Engineer. As of now the unit is more or less discarded after the first minutes of a game. This could also serve as an extended olive branch to players wanting to have the unit used as a production queue killer as it gives the unit a purpose after it has finished capturing oil derrics: http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&p=297153#297153

Allow the Engineer to detect, walk over and disassemble mines: So what if you could a-move the Engineer, triggering it to disassemble mines along its' path, perhaps with a detection range of say, 3?

Allow the Engineer to gradually repair structures: CTRL+Right Click a structure, start repairing. Slowly.

r34ch:
+1 for the engie mine detection and disable.

I like the gradual (free) repair of structures idea but it would make more sense for the in-game logic to match the medic/mechanic in that any building in a nearby radius is repaired rather than modifying an existing core repair mechanic (i.e. forget to Ctrl, loose the engi completely..)

GraionDilach:
Medic and mechanic both uses a weapon to fix nearby damaged actors. (Although providing a healaura is also possible, it's not used in any shipped mods.)

pchote:
C&C3 (and RA3?) let engineers recover husks, and we copied this over to the engineers in our TD. Of course, RA has the mechanic, which messes up this otherwise obvious use for engineers there.

whi7ed3vil:
I'd just love for the ability to stop my units from destroying a building currently being captured by an engie. Only solution now is to move them out of range. Probably shoud split this to a new topic ...

Smittytron:
Allow the engineer to C4 a concrete wall, destroying a 3 or 5 cell radius of wall. Would it be used that often? No. Would it look awesome when someone uses it successfully? Heck yes.

SoScared:
@Smittytron Intriguing! Tho 5 cell radius sounds like a lot,

Smittytron:
I view the engineer as a utility unit as opposed to a support unit. Something that does things across the battlefield but has no role in pitched combat like a medic or mechanic.
Here is my wishlist of engineer abilities:
- Combat Engineer: Engineers can C4 concrete walls and sandbags, destroying a 3x3 radius. (using Tanya’s C4 cursor)
- Wire Cutters: Engineers can instantly remove one cell of adjacent wire fencing. (Play fence being run over sound)
[...]

PR #11801, opened 2016-08-10
Allow multiple minelayers in a single MinefieldOrderGenerator.

MunWolf:
This fixes #11763 by allowing multiple minelayers to be a part of the same minefield order.

PR #11859, opened 2016-08-21
Let engineers detect mines

abcdefg30:
The minor part of #11749. The ability to defuse mines will be added in a follow-up PR.

### DIFF
--- a/mods/ra/rules/infantry.yaml
+++ b/mods/ra/rules/infantry.yaml
@@ -185,6 +185,9 @@ E6:
 		VoiceSet: EngineerVoice
 	Selectable:
 		Priority: 5
+	DetectCloaked:
+		Range: 4c0
+		CloakTypes: Mine
 
 SPY:
 	Inherits: ^Soldier


SoScared:
:+1: One more reason to keep engineers around after caps.

Micr0Bit:
why would i want to defuse a mine ... rather then shooting on it and  even get a chain-reaction ?

pchote:
This doesn't seem like a great idea to me.  Do players ever use mines?  Why would we want to make those even less useful?

GraionDilach:
The playerbase reason behind this is to justify fixing the externalcapture-freeze bug(!) because the Engineer needs active roles beyond capturing(!).

Yes, I agree, the playerbase reasons are terrible at best here.

pchote:
Also note that both sides already *have* a unit for detecting enemy mines – the minelayer.

obrakmann:
While players might begin to use mines more after #11801 made the minelaying not as cumbersome*, I think that's a good thing and we shouldn't immediately add another counter to it.

Also engineers already have a double role. Let's just close #11749. I'm very much in agreement with Graion here.

*IF they even realize that it was fixed and even know about the minefield-laying command.

SoScared:
Minelayers' abscense is noticable on big team and choke maps since constant action makes mine laying redundant. On maps with larger areas per player Minelayers are actually starting being used to delay attacks and narrow down attack routes. #11801 will likely boost Minelayer popularity.

Shifting the role of the engineer from engi-cap-canceling to an utility unit is not really a bad idea. Using engineers as a mine detector would allow players to avoid building a Minelayer simply to detect mines after which the unit is disgarded or put to use laying its own mines by whim because, well, now you have a Minelayer.

Engineers are cheaper and more available, but as opposed to the Minelayer it's slow, fragile and will still die if walking over or standing next to an exploding mine field.

Micr0Bit:
another thing that concerns me , that its quite hard for "allies" factions ... to protect a fragile engineer from air-attacks ... allies dont have "mobile AA" (soviets do) ... all they have is rocket-troops ... and funny because those make it perfect for yaks to suicide on infantry ... on purpose 

no thanks ... 
then i would rather want to keep the mine-detecting job for minelayers ... instead of a 10-20hp single infantry-unit that costs 500 and even has less speed then a minelayer

r34ch:
> why would i want to defuse a mine ... rather then shooting on it and even get a chain-reaction ? 

I would assume defusing would be stealthy and give no audio queue

pchote:
@abcdefg30 asked me on IRC to be more explicit with my opinions on these balance topics, so here we go... this is a bit of a meta-complaint, but this pr seems the best place to put it.

Engineers are one of the most iconic C&C units, and i'm not comfortable with moving the RA engineers even further away from that formula, even if it does improve them from a purely gameplay perspective (keeping the external capturing after we fixed the technical issues with the "classic" behaviour is bad enough...)

Modernizing/tweaking the balance to suit modern expectations is one thing, but it shouldn't be a free ticket to let the holy-grail of perfect gameplay override all other concerns.   The C&C/Red Alert names come with expectations, and we have a responsibility to constrain our changes to respect that.

Of course, different people have different opinions on the best tradeoff between gameplay and faithfulness, so this argument is a giant messy can of worms.  Personally, I consider the large majority of our changes to RA and TD as solid improvements that broadly respect the gameplay mechanics and themes of the originals.  RA Engineers are one thing that I've long (and loudly) considered we got wrong, and I see this ticket as taking us even further in that wrong direction.


TL;DR: A sad :-1:

Micr0Bit:
actually , why not give it a test :+1:

TheRaffy:
hmm. I'm not a big fan of this idea.  Soscared said: ~mines are finally be used. 
I think making them more easy to detect will harm the useage of mines again. 
It's that thing in my opinion to force the other player to also build minelayers for that detection, that makes mines just that good. 

Also think about 800$+Service Depot + warfactroy vs. 500$ +barracks

reaperrr:
I'm inclined against this as well.

In my opinion, the ability to defuse mines makes much more sense for engies than the ability to detect them. I'd rather make minelayers a little cheaper instead.

reaperrr:
We have a total of 3 people against the PR in its current form, with @pchote and @TheRaffy making good arguments against this change, and there has been no progress on either the PR or discussion in over a month.

I'm closing this for now due to staleness.

PR #11862, opened 2016-08-22
Draw tactical overlays (beacons, rallypoints, text overlays, order generators) above the shroud.

### LINKED ISSUE #6106: Beacons should be shown ontop of shroud
beaconbuga
beaconbugb
There is no beacon shown, when you place a supportpower ontop of shroud.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3391424-beacons-should-be-shown-ontop-of-shroud?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
These things can be thought of like scribbles on top of a video feed, so drawing them behind the shroud doesn't make a lot of sense – it leads to inconsistencies in the classic mods, and downright bogus rendering in the heightmap-enabled mods when we turn on the z-buffer (the overlays can clip behind the terrain as shown in the screenshot below).
- Fixes #6106.
- Makes render circles / selection boxes from order generators render above the shroud like they do everywhere else (I don't think we have a ticket for this issue).
- Works towards #7520.
- Removes duplication/hacks from the health bar rendering.
- Fixes some code style issues.

line

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithTextDecoration.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithTextDecoration.cs
@@ -54,34 +54,28 @@ namespace OpenRA.Mods.Common.Traits.Render
 		}
 	}
 
-	public class WithTextDecoration : UpgradableTrait<WithTextDecorationInfo>, IRender, IPostRenderSelection, INotifyCapture
+	public class WithTextDecoration : UpgradableTrait<WithTextDecorationInfo>, IRender, IRenderAboveShroudWhenSelected, INotifyCapture
 	{
-		readonly Actor self;
 		readonly SpriteFont font;
-
 		Color color;
 
 		public WithTextDecoration(Actor self, WithTextDecorationInfo info)
 			: base(info)
 		{
-			this.self = self;
-
-			if (!Game.Renderer.Fonts.TryGetValue(info.Font, out font))
-				throw new YamlException("Font '{0}' is not listed in the mod.yaml's Fonts section".F(info.Font));
-
+			font = Game.Renderer.Fonts[info.Font];
 			color = Info.UsePlayerColor ? self.Owner.Color.RGB : Info.Color;
 		}
 
 		public virtual bool ShouldRender(Actor self) { return true; }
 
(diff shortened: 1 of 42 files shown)


OpenRA.Mods.Common/Traits/Render/WithTextDecoration.cs
```diff
@@ -54,34 +54,28 @@ void IRulesetLoaded<ActorInfo>.RulesetLoaded(Ruleset rules, ActorInfo info)
...
 
 		public WithTextDecoration(Actor self, WithTextDecorationInfo info)
 			: base(info)
 		{
-			this.self = self;
-
-			if (!Game.Renderer.Fonts.TryGetValue(info.Font, out font))
-				throw new YamlException("Font '{0}' is not listed in the mod.yaml's Fonts section".F(info.Font));
-
+			font = Game.Renderer.Fonts[info.Font];
```
pchote: Before anyone asks: we already check this in the *Info, so this was redundant.

pchote:
Also added floating text.

GraionDilach:
Floating text, like cash ticks shouldn't be rendered above shroud though.

pchote:
The text isn't rendered if the cell it is spawned from is shrouded.  The change is that it won't move under the shroud in the next cell as it floats up.

pchote:
Added repair and power down indicators so that they work with the depth buffer enabled and are more visible for structures at the edge of the map.

Before:


After:

Mailaender:
I can't spot any visual regressions :white_check_mark: and the code changes look fine. :+1:

reaperrr:
> I can't spot any visual regressions :white_check_mark: and the code changes look fine. :+1:

Same here.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3220d57e9f0e9135a37d9308321ff09efb400996...03f83addc8ccaffd5cd542b8345cd735b3480fd2)

PR #11863, opened 2016-08-22
Restore graphical look of the TS Ion Cannon.

ABrandau:
I defined the ion ring as the main effect of the weapon because that way it appears in the center of the targetted cell, and in order to make the beam I had to code it as many explosion effects, using many instances of it while also using the offset tag on the sequence. 
ioncannonpolish

see here https://www.youtube.com/watch?v=YnJlb-Zorz8 0:40.


GraionDilach:
Looks good to me. :+1:

abcdefg30:

PR #11867, opened 2016-08-23
Fix GPSDots being visible when the unit is not InWorld or dead

abcdefg30:
It was still visible when e.g. someone loaded a unit into a transport.
gpsdotbugagpsdotbugb
gpsdotbugc
gpsdotbugdgpsdotbuge

### DIFF
--- a/OpenRA.Mods.RA/Effects/GpsDot.cs
+++ b/OpenRA.Mods.RA/Effects/GpsDot.cs
@@ -124,16 +124,14 @@ namespace OpenRA.Mods.RA.Effects
 			if (self.Disposed)
 				world.AddFrameEndTask(w => w.Remove(this));
 
-			if (!self.IsInWorld || self.IsDead)
-				return;
-
 			for (var playerIndex = 0; playerIndex < dotStates.Count; playerIndex++)
 			{
 				var state = dotStates[playerIndex];
 				var shouldRender = false;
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.RA/Effects/GpsDot.cs
```diff
@@ -124,15 +124,18 @@ public void Tick(World world)
...
 				var state = dotStates[playerIndex];
 				var shouldRender = false;
-				var targetable = (state.Gps.Granted || state.Gps.GrantedAllies) && IsTargetableBy(world.Players[playerIndex], out shouldRender);
-				state.IsTargetable = targetable;
+				var targetable = false;
+
+				if (self.IsInWorld && !self.IsDead)
+				{
+					targetable = (state.Gps.Granted || state.Gps.GrantedAllies) && IsTargetableBy(world.Players[playerIndex], out shouldRender);
+					state.IsTargetable = targetable;
```
Phrohdoh: Since you are here already just assign this to the expression above (instead of using a temporary variable).

abcdefg30: I still need the variable three lines below.

abcdefg30:
Updated.

GraionDilach:
:+1:

pchote:
Fix confirmed :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/e1a26a96e7f01bb328b90a9899bb6b8ea5f0ce04

PR #11868, opened 2016-08-23
Fix the floating text for bounties of actors not being InWorld showing up

abcdefg30:
Before you could see the text at the location the unit was removed from world.

gpsbounty

### DIFF
--- a/OpenRA.Mods.Common/Traits/GivesBounty.cs
+++ b/OpenRA.Mods.Common/Traits/GivesBounty.cs
@@ -65,15 +65,15 @@ namespace OpenRA.Mods.Common.Traits
 			return (slevel > 0) ? slevel * info.LevelMod : 100;
 		}
 
-		int GetBountyValue(Actor self, int percentage)
+		int GetBountyValue(Actor self)
 		{
 			// Divide by 10000 because of GetMultiplier and info.Percentage.
-			return self.GetSellValue() * GetMultiplier() * percentage / 10000;
+			return self.GetSellValue() * GetMultiplier() * info.Percentage / 10000;
 		}
 
 		int GetDisplayedBountyValue(Actor self)
 		{
-			var bounty = GetBountyValue(self, info.Percentage);
+			var bounty = GetBountyValue(self);
 			if (cargo == null)
 				return bounty;
 
@@ -102,7 +102,7 @@ namespace OpenRA.Mods.Common.Traits
 			if (info.ShowBounty && self.IsInWorld && displayedBounty > 0 && e.Attacker.Owner.IsAlliedWith(self.World.RenderPlayer))
 				e.Attacker.World.AddFrameEndTask(w => w.Add(new FloatingText(self.CenterPosition, e.Attacker.Owner.Color.RGB, FloatingText.FormatCashTick(displayedBounty), 30)));
 
-			e.Attacker.Owner.PlayerActor.Trait<PlayerResources>().GiveCash(GetBountyValue(self, info.Percentage));
+			e.Attacker.Owner.PlayerActor.Trait<PlayerResources>().GiveCash(GetBountyValue(self));
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


pchote:
We probably want to aggregate the value of cargo in a transport and show a single cash tick for the whole value.

GraionDilach:
Sometimes it aggregates already. Really random though.

abcdefg30:
Included your suggestion:
gpsbounty2

OpenRA.Mods.Common/Traits/GivesBounty.cs
```diff
@@ -69,13 +82,24 @@ void INotifyKilled.Killed(Actor self, AttackInfo e)
...
-				e.Attacker.World.AddFrameEndTask(w => w.Add(new FloatingText(self.CenterPosition, e.Attacker.Owner.Color.RGB, FloatingText.FormatCashTick(bounty), 30)));
+			// Divide by 10000 because of GetMultiplier and info.Percentage.
+			var bounty = GetBountyValue(self, info.Percentage);
+
+			if (info.ShowBounty && self.IsInWorld && bounty > 0 && e.Attacker.Owner.IsAlliedWith(self.World.RenderPlayer))
+			{
+				var displayedBounty = bounty;
+				if (cargo != null)
+				{
+					foreach (var a in cargo.Passengers)
```
pchote: This loop should move inside GetBountyValue, and then the lookup below should use the trait instead of the info so that this behaves properly when there are recursive transports (infantry in an APC in a naval transport).

abcdefg30: Why should this move inside of `GetBountyValue`? This is purely about the visuals. Each actor still adds the actual cash to the player itself.

abcdefg30:
Updated again.

OpenRA.Mods.Common/Traits/GivesBounty.cs
```diff
@@ -41,23 +41,48 @@ class GivesBountyInfo : ITraitInfo
 	class GivesBounty : INotifyKilled
 	{
 		readonly GivesBountyInfo info;
+		readonly GainsExperience gainsExp;
+		readonly Cargo cargo;
 
 		public GivesBounty(Actor self, GivesBountyInfo info)
 		{
 			this.info = info;
+			gainsExp = self.TraitOrDefault<GainsExperience>();
```
pchote: These need to move into `INotifyCreated.Created`, otherwise you introduce yaml-order initialization bugs.

pchote: Note that this won't save you any perf because these lookups only ever happen once anyway.

OpenRA.Mods.Common/Traits/GivesBounty.cs
```diff
@@ -58,6 +58,29 @@ int GetMultiplier(Actor self)
 			return (slevel > 0) ? slevel * info.LevelMod : 100;
 		}
 
+		int GetBountyValue(Actor self, int percentage)
```
pchote: Passing the percentage in here is redundant.  You only ever call this with `info.Percentage`.

OpenRA.Mods.Common/Traits/GivesBounty.cs
abcdefg30: Actually not *only*. This might be used twice, `GetBountyValue` is called from `INotifyKilled.Killed` and from `GetDisplayedBountyValue`. But yeah, caching might be not worth it.

abcdefg30:
Updated.

pchote:
Thanks, that looks better. 👍

obrakmann:
Looks fine, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d223624d335424288b3c524f8543f5b849382a59)

OpenRA.Mods.Common/Traits/GivesBounty.cs
```diff
@@ -38,26 +38,55 @@ class GivesBountyInfo : ITraitInfo
...
+
+		int GetDisplayedBountyValue(Actor self)
+		{
+			var bounty = GetBountyValue(self);
+			if (cargo == null)
+				return bounty;
+
+			foreach (var a in cargo.Passengers)
+			{
+				var givesBounty = a.TraitOrDefault<GivesBounty>();
```
GraionDilach: This lookup is wrong. Since GivesBounty was updated with damagetypes, it's likely and valid that an actor will have more of this trait.

PR #11869, opened 2016-08-23
Make naval transports chronoshiftable

### LINKED ISSUE #8503: Cronoshiftable Unit still dies when Chronoed in a Transpot
To try add `Chronoshiftable:` to an infantry. It chronoes normally but dies when in a transport. Transport boats aren't chronoshiftable so this is not cause any problem for vanilla RA mod. But block me on my map with custom rules.

abcdefg30:
I don't see a reason why they should not be chronoshiftable.
(It felt more like a bug/oversight, as we allow all other transport/naval units to chronoshift.)

Fixes #8503.

### DIFF
--- a/OpenRA.Mods.RA/Activities/Teleport.cs
+++ b/OpenRA.Mods.RA/Activities/Teleport.cs
@@ -68,30 +68,7 @@ namespace OpenRA.Mods.RA.Activities
 			self.Generation++;
 
 			if (killCargo && teleporter != null)
-			{
-				var cargo = self.TraitOrDefault<Cargo>();
-				if (cargo != null)
-				{
-					var actorsToLoad = new List<Actor>();
-					while (!cargo.IsEmpty(self))
-					{
-						var a = cargo.Unload(self);
-						var chronoshiftable = a.TraitOrDefault<Chronoshiftable>();
-						if (chronoshiftable != null && !chronoshiftable.Info.ExplodeInstead)
-						{
-							actorsToLoad.Add(a);
-							continue;
-						}
-
-						// Kill all the units that are unloaded into the void
-						// Kill() handles kill and death statistics
-						a.Kill(teleporter);
-					}
-
-					foreach (var a in actorsToLoad)
-						cargo.Load(self, a);
-				}
-			}
+				KillCargo(self);
 
 			// Consume teleport charges if this wasn't triggered via chronosphere
 			if (teleporter == self && pc != null)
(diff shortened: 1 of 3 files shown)


pchote:
Probably consistency with the original game?

MustaphaTR:
This makes #8503 a problem in ra mod itself tho, do we really want tanks in transport boat die?

abcdefg30:
Updated. This also fixes #8503 now.

pchote:
This now has very strong balance implications so it would be a bad idea to try and sneak this in without any community discussion.

abcdefg30:
We don't have to take it straight before the release. I can also remove the first commit if wanted, but for now it is a good testcase at least.

Micr0Bit:
@pchote is right 

if we would allow to crono vehicles inside a transporter ... that would mean i could crono 5 transporter a` 5 vehicles ... that means 25 vehicles (with a standard crono) ... that allows me to teleport all those transporters directly on the enemy shore ... and unload ... further more ... what happens with the vehicles when they get unloaded ? do they stay on the position when they get unloaded ? ... or teleported back ?

this would cause a balance-issue in the favor of allies factions ... and could almost completely be uneffected by water-defense ... or better said , it would make water-dominance useless ...

MustaphaTR:
It is not like that currently in this PR, but they should come bacl and possibly die as they'll land into water when they chronp back.

Micr0Bit:
@MustaphaTR : then it sounds strange that i cant crono vehicles by my own hand into water ...

abcdefg30:
See http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19763.

pchote (commented):
If you remove the first commit (which is where all the balance changes come from) then the rest of this looks like a sensible bug/feature fix for downstream mods.

abcdefg30:
http://www.sleipnirstuff.com/forum/viewtopic.php?f=82&t=19763 has been long enough open for everyone to vote, I think.
The results are:
chronolstvote
Which is tight, but still favors the first commit.

Anyhow, in the worst case,

> I can also remove the first commit if wanted, but for now it is a good testcase at least.

still applies, so I don't see any blocker for this PR.

pchote:
> Probably consistency with the original game?

I have confirmed, mainly for my own satisfaction, that the original RA did not let you chronoshift naval transports.

reaperrr:
👍 for the code changes, but I agree that the first commit should be removed before this is merged, that should go to a separate PR where the community could concentrate on the balance implications.

obrakmann:
Ignoring the balance implications for now, just for the sake of discussion, should we maybe move the `KillCargo` flag from `ChronoshiftPower` to `Cargo`?

MustaphaTR:
`Chronoshiftable:` would be a better place to move it imo.

Maybe we should have tags about killing passengers on all of these 3 traits. I don't know how exactly tho.

pchote:
I agree that `Chronoshiftable` would be a better place.  Making our very-generic cargo code depend on our very-ra chronoshift code would be bad.

reaperrr:
> I agree that Chronoshiftable would be a better place. Making our very-generic cargo code depend on our very-ra chronoshift code would be bad.

Agreed.

Also needs a rebase.

pchote:
Closing stale PR.

PR #11872, opened 2016-08-24
Fix d2k grenadier sequences.

### LINKED ISSUE #11785: D2k grenadier's prone sequences are wrong.
They currently break-dance and then turn into holes in the ground when firing.

pchote:
Fixes #11785.

### DIFF
--- a/mods/d2k/rules/infantry.yaml
+++ b/mods/d2k/rules/infantry.yaml
@@ -189,7 +189,7 @@ grenadier:
 	AttackFrontal:
 	WithInfantryBody:
 		AttackSequence: throw
-		IdleSequences: idle
+		IdleSequences: idle, idle2
 	Explodes:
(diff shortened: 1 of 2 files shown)


obrakmann:
:+1:

reaperrr:
Am I right that this was a regression from the switch to original 1.06 assets, so it happened after last release and doesn't need a changelog entry?

pchote:
Best would be to add it as a second pr on the entry explaining the switch to the 1.06 assets.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/e4c7bd12e0feb5d7432c5a9cb9055aee3f91ea31...3220d57e9f0e9135a37d9308321ff09efb400996)

PR #11745, opened 2016-07-30
Fix cancellation of downloads.

RoosterDragon:
The Download class cancels asynchronously, which means callers must handle cancellation inside the completion event, and not after requesting cancellation.

Fixes #11739.

When a download is cancelled, the `Cancelled` flag is set, and `Error` contains an exception with a `RequestCanceled` Status. This means if you want to check for either cancellation or an error, simply checking for `Error != null` is sufficient. It also means if you want to handle cancellation differently from an error, you must handle cancellation first.

PR #11878, opened 2016-08-25
Allow download window to be closed on failed downloads.

### LINKED ISSUE #11843: 'Cancel' in Content download dialogue possibly broken
I noticed this on #11841, when trying to download the Firestorm assets (which failed due to lack of mirrors), only the 'Retry' button worked, the 'Cancel' button did nothing.

RoosterDragon:
Fixes #11843. Regression from #11745.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Installation/DownloadPackageLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Installation/DownloadPackageLogic.cs
@@ -113,6 +113,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 				progressBar.Percentage = 100;
 				getStatusText = () => "Error: " + s;
 				retryButton.IsVisible = () => true;
+				cancelButton.OnClick = Ui.CloseWindow;
 			});
 
 			Action<AsyncCompletedEventArgs> onDownloadComplete = i =>


OpenRA.Mods.Common/Widgets/Logic/Installation/DownloadPackageLogic.cs
```diff
@@ -208,6 +208,7 @@ void ShowDownloadDialog()
 					if (i.Error != null)
 					{
 						onError(Download.FormatErrorMessage(i.Error));
+						cancelButton.OnClick = Ui.CloseWindow;
```
obrakmann: Hmm... .Wouldn't it make more sense to add this line to the error handler action instead (somewhere around line 115)? Alternatively, you'd have to add this same line before line 131.

RoosterDragon: You're right - will move it to the error handler.

RoosterDragon: Done.

obrakmann:
:+1:

pchote:
Fix confirmed, thanks :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/eb5c2b18d8c881b079cd1075f2b5bb4cd0bb1702

PR #11879, opened 2016-08-25
Fix issuing orders when releasing RMB after joystick-scrolling

### LINKED ISSUE #11849: (Bleed) Screen-grab scrolling (Standard) with Right Mouse Button executes orders when units are selected.
Happens every time RMB is released when scrolling.

#11148 #9399

obrakmann:
Fixes #11849.

Please test this properly, there may be regressions lurking here which I haven't found yet or am unable to find since my use of the controls is different from how other people use theirs. (I never ran into the original bug in the first place, for example.)

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
@@ -215,7 +215,7 @@ namespace OpenRA.Mods.Common.Widgets
 
 			var blockedDirections = worldRenderer.Viewport.GetBlockedDirections();
 
-			if (IsJoystickScrolling)
+			if (IsJoystickScrolling || isStandardScrolling)
 			{
 				foreach (var dir in JoystickCursors)
 					if (blockedDirections.Includes(dir.Key))
--- a/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
@@ -35,6 +35,7 @@ namespace OpenRA.Mods.Common.Widgets
 		public int EdgeCornerScrollThreshold = 35;
 
 		int2? joystickScrollStart, joystickScrollEnd;
+		bool isStandardScrolling;
 
 		static readonly Dictionary<ScrollDirection, string> ScrollCursors = new Dictionary<ScrollDirection, string>
 		{
@@ -287,9 +288,18 @@ namespace OpenRA.Mods.Common.Widgets
 			{
 				if (mi.Event == MouseInputEvent.Move)
 				{
+					isStandardScrolling = true;
 					var d = scrollType == MouseScrollType.Inverted ? -1 : 1;
 					worldRenderer.Viewport.Scroll((Viewport.LastMousePos - mi.Location) * d, false);
 					return true;
+				}
+				else if (mi.Event == MouseInputEvent.Up)
+				{
+					var wasStandardScrolling = isStandardScrolling;
+					isStandardScrolling = false;
+
+					if (wasStandardScrolling)
+						return true;
 				}
 			}
 
--- a/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
@@ -305,7 +305,7 @@ namespace OpenRA.Mods.Common.Widgets
 
 				if (mi.Event == MouseInputEvent.Up)
 				{
-					var wasJoystickScrolling = IsJoystickScrolling;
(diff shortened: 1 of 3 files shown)


SoScared:
I was a bit unclear on #11849. I was originally referring to 'standard' RMB scrolling, not joystick.

Sininini:
This still fixes the potential polish issue that I described in  #11849 . Order cursor still appears but doesn't execute orders til RMB is released to exit scrolling mode.

obrakmann:
Ah, I see. Yeah, it's much worse with standard scrolling. Fixed as well now.

OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
```diff
@@ -35,6 +35,7 @@ public class ViewportControllerWidget : Widget
 		public int EdgeCornerScrollThreshold = 35;
 
 		int2? joystickScrollStart, joystickScrollEnd;
+		bool isStandardScrolling;
```
pchote: It would be good if we could suppress the order cursor when this is true.  This should fix

> Order cursor still appears but doesn't execute orders til RMB is released to exit scrolling mode.

pchote:
Looks good otherwise :+1:

obrakmann:
Updated

Micr0Bit:
@obrakmann : just gave it a test ... it did already work before your update 
:+1:

Sininini:
This also fixes RMB joystick scrolling in map editor, but not RMB standard scrolling - entities still get deleted if you start dragging the map with cursor on top of them. Could you perhaps take a look at this too?

obrakmann:
The map editor issue is a different problem. It seems that the mouse input handler of the editor runs before the one from the viewport controller.

Micr0Bit:
i can confirm what @Sininini said

obrakmann:
OK, fixed that as well, hopefully without introducing regressions in the editor.

The cause for that bug was that the `HandleMouseInput` method of the `EditorViewPortController` ran before the one from the regular `ViewPortController`, thus intercepting the button press ([see here](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Widgets/Widget.cs#L383)).

Switching the order of the two around in the editor's chrome yaml fixes that, but introduced another problem, namely that the tooltips stopped working. Switching the order meant that the `ViewPortController` laid "on top" of the `EditorViewPortController`, so the latter didn't receive the `MouseEntered` event anymore ([see here](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Widgets/Widget.cs#L119-L120)). Setting the `IgnoreMouseOver` flag in the `ViewPortController` fixes this.

Sininini:
Thanks! Fixed in all editors now.

abcdefg30:
👍 / ✅

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/224dc7644af57f2ec6d24e3e851e8fa139c71f6a...c03418c08207e57480b98cedbe7c1ba9f018729f)

obrakmann:
Although a bug fix, this does in practice lead to an awful game control experience. Many clicks just go unnoticed just because you moved your mouse slightly.

pchote:
The standard approach is to have a deadzone so that movements must be above a certain threshold to count.

pchote:
Has this introduced a regression that we need to fix?

Sininini:
Standard scrolling never had deadzone because it used to be on middle mouse button and that didn't do anything else. The issue wasn't noticed when ability to use RMB for scrolling was added because this bug was still there. Now that it's fixed the issue became noticeable. Standard scrolling needs a small deadzone now if used on RMB. 
Release blocker for sure, but shouldn't be too difficult to fix.

Joystick scrolling has no issues that I could notice.

pchote:
Joystick scrolling uses a deadzone, so it would be concerning if it did have issues!

OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
```diff
@@ -305,7 +315,7 @@ public override bool HandleMouseInput(MouseInput mi)
 
 				if (mi.Event == MouseInputEvent.Up)
 				{
-					var wasJoystickScrolling = IsJoystickScrolling;
+					var wasJoystickScrolling = joystickScrollStart.HasValue && joystickScrollEnd.HasValue;
```
obrakmann: This is the obvious cause for the breakage for joystick-scrolling. I simply changed the code to ignore the deadzone. Reverting the change to this line *will* re-introduce the bug from #11849 for joystick-scrolling.

Sininini:
Re-introducing that joystick scrolling bug seems fine to me because I seriously doubt anyone would use it in the way required for it to trigger.

PR #11881, opened 2016-08-25
Allow keeping empty selection bars visible

reaperrr:
External mods might want to display bars even when they are empty.

In my example, I want a bar displaying heat level stay visible even if the current heat level is zero.

### DIFF
--- a/OpenRA.Game/Traits/TraitsInterfaces.cs
+++ b/OpenRA.Game/Traits/TraitsInterfaces.cs
@@ -280,7 +280,7 @@ namespace OpenRA.Traits
 	public interface IPips { IEnumerable<PipType> GetPips(Actor self); }
 
 	[RequireExplicitImplementation]
-	public interface ISelectionBar { float GetValue(); Color GetColor(); }
+	public interface ISelectionBar { float GetValue(); Color GetColor(); bool DisplayWhenEmpty { get; } }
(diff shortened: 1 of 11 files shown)


OpenRA.Game/Traits/TraitsInterfaces.cs
```diff
@@ -280,7 +280,7 @@ public interface IPaletteModifier { void AdjustPalette(IReadOnlyDictionary<strin
 	public interface IPips { IEnumerable<PipType> GetPips(Actor self); }
 
 	[RequireExplicitImplementation]
-	public interface ISelectionBar { float GetValue(); Color GetColor(); }
+	public interface ISelectionBar { float GetValue(); Color GetColor(); bool KeepWhenEmpty(); }
```
pchote: Could we please make this a property, and call it DisplayWhenEmpty for clarity?

reaperrr:
Updated.

GraionDilach:
Yes, this is something I like and already have use for even. :+1:

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/88771d945e314ba4942f8931919b8876456f82e1)

Issue #3605, opened 2013-07-27
Support TS-style terrain lighting


pchote:
This should be straightforward to do by changing the GL blend mode or using a custom fragment shader to lighten fragments under the spotlight circle.

Phrohdoh:
This may come from #7811.

pchote:
It turns out that this is more fiddly than I first assumed.  TS/RA2 applied lighting to *all* objects, with a base level that depends on the ground height and then any additional modifiers blended in.  This affects both the terrain *and* (some) actor sprites.  My original approach (rendering overlay sprites with a blend mode) won't cut it.

Instead, we will probably need to:

* Add plumbing for a per-sprite color tint through the renderer, and then into the shader via `GL.ColorPointer` and `gl_color`.
* Apply the color tint in the fragment shader after all other effects have been applied (hopefully won't conflict with other things we do).
* Add a new `LightMap` engine trait (with a direct reference from World) that manages all the light sources in the world and exposes a `CellLayer<Color>` for the aggregated per-tile lighting.
* The TerrainRenderer will need to bind to the CellLayer so that it can update the vertex buffer whenever the lighting changes.
* Sprite/Voxel/Beam/etc Renderable will need to query `LightMap` for their specific tints from `PrepareRender`.

chrisforbes:
This sounds like a texture, not vertex data.

pchote:
Yeah it does, doesn't it?

I guess the simplest way would be to define a 1 x N texture with all the light values, and then index that with one (down from four) float per vertex.

Mailaender:
If I understand https://github.com/zzattack/ccmaps-net/blob/master/CNCMaps.Engine/Rendering/Palette.cs correctly @zzattack solved it with palette changes.

pchote:
I have a local proof of concept demonstrating classic lighting behaviour (both terrain height and IN*LMP) using vertex colours:

![Screenshot 2020-06-16 at 21 04 46 (2)](https://user-images.githubusercontent.com/167819/84825682-b5439e80-b019-11ea-856e-651e844b8d02.png)

This plumbing to support this enables lighting effects for RGBA sprites, and can be used in the future to add RGBA support to more things (e.g. target flashes, start/end game fading).

PR #11882, opened 2016-08-25
Remove legacy A* visualisation.

pchote:
This visualization looks fun, but hasn't been useful for as long as I can remember.  It needs a complete rewrite before it will properly with isometric or heightmap-enabled mods.

This PR puts the feature out of its misery so that it doesn't cause any more maintenance burden until it is next needed.  If that time ever comes it can then be re-implemented in a way that works for all mods (e.g. using color-tinted sprite overlays or whatever technique we end up using for #3605).


reaperrr:
:+1:

obrakmann:
Please add an upgrade rule. :+1: after that.

pchote:
Done.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1e9153a44ac9e2e9bfb2effd9341cac428b587e7...44e35626787ae660e54b77f1b3dcb8090d71d4d9)

PR #11886, opened 2016-08-26
Add UpgradeOnMovement trait

reaperrr:
As the name states, the trait provides upgrades while an actor is moving, this can be used to increase inaccuracy while moving, give the unit a more vulnerable armor type while moving and so on.

This nicely complements our set of upgrades-providing traits.

The trait is upgradable itself, which could be used to switch the provided upgrades manually in conjunction with `DeployToUpgrade`.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnMovement.cs
+++ b/OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnMovement.cs
@@ -1,0 +1,75 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Traits
+{
+	public class UpgradeOnMovementInfo : UpgradableTraitInfo, Requires<UpgradeManagerInfo>, Requires<IMoveInfo>
+	{
+		[UpgradeGrantedReference, FieldLoader.Require]
+		[Desc("The upgrades to grant.")]
+		public readonly string[] Upgrades = { };
+
+		[Desc("Apply upgrades on straight vertical movement as well.")]
+		public readonly bool ConsiderVerticalMovement = false;
+
+		public override object Create(ActorInitializer init) { return new UpgradeOnMovement(init.Self, this); }
+	}
+
+	public class UpgradeOnMovement : UpgradableTrait<UpgradeOnMovementInfo>, ITick
+	{
+		readonly UpgradeManager manager;
+		readonly IMove movement;
+
+		bool granted;
+
+		public UpgradeOnMovement(Actor self, UpgradeOnMovementInfo info)
+			: base(info)
+		{
+			manager = self.Trait<UpgradeManager>();
+			movement = self.Trait<IMove>();
+		}
+
+		void Revoke(Actor self)
+		{
+			if (!granted)
+				return;
+
+			foreach (var up in Info.Upgrades)
+				manager.RevokeUpgrade(self, up, this);
+
+			granted = false;
+		}
+
+		void ITick.Tick(Actor self)
+		{
+			if (IsTraitDisabled)
+			{
+				Revoke(self);
+				return;
+			}
+
+			var isMovingVertically = Info.ConsiderVerticalMovement ? movement.IsMovingVertically : false;
+			var isMoving = !self.IsDead && (movement.IsMoving || isMovingVertically);
+			if (isMoving && !granted)
+			{
+				foreach (var up in Info.Upgrades)
+					manager.GrantUpgrade(self, up, this);
+
+				granted = true;
+			}
+			else if (!isMoving)
+				Revoke(self);
+		}
+	}
+}
--- a/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
+++ b/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj
@@ -499,6 +499,7 @@
     <Compile Include="Traits\Upgrades\UpgradeActorsNear.cs" />
     <Compile Include="Traits\Upgrades\UpgradeOnDamageState.cs" />
     <Compile Include="Traits\Upgrades\UpgradeOnTerrain.cs" />
+    <Compile Include="Traits\Upgrades\UpgradeOnMovement.cs" />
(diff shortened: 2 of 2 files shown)


ABrandau:
Seems to work fine, I emped the unit to see if it had any issues and nope, same as when the unit attack moves.

WolfGaming:
Do we want to have a property that flips the functionality? Aka grants an upgrade when not moving?

pchote:
The upgrades system already lets you do that without making a new type.

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnMovement.cs
```diff
@@ -0,0 +1,68 @@
...
+		public override object Create(ActorInitializer init) { return new UpgradeOnMovement(init.Self, this); }
+	}
+
+	public class UpgradeOnMovement : UpgradableTrait<UpgradeOnMovementInfo>, ITick
+	{
+		readonly UpgradeManager manager;
+		readonly IMove movement;
+
+		bool granted;
+		bool isMoving;
```
WolfGaming: Why have this here and not defined in the tick like oldCachedPosition?

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnMovement.cs
```diff
@@ -0,0 +1,68 @@
...
+
+		void ITick.Tick(Actor self)
+		{
+			if (IsTraitDisabled)
+				return;
+
+			var oldCachedPosition = cachedPosition;
+			cachedPosition = self.CenterPosition;
+
+			isMoving = !self.IsDead && (movement.IsMoving || (oldCachedPosition - cachedPosition).HorizontalLengthSquared != 0);
```
WolfGaming: Do you do (oldCachedPosition - cachedPosition).HorizontalLengthSquared != 0 for actors inside transports or am I missing something?

reaperrr: During my tests there were some 'hiccups' when only using IMove.IsMoving (during turns, usually). If we managed to make that more robust instead this could go away, but until then this is needed to make sure the upgrades are applied all the time while moving.

reaperrr:
Updated:
- upgrades are now also revoked if `UpgradeOnMovement` is disabled
- `isMovement` is now a `var` inside Tick
- moved the testcase to https://github.com/reaperrr/OpenRA/tree/UpgradeOnMovement-test

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnMovement.cs
```diff
@@ -0,0 +1,76 @@
...
+			if (IsTraitDisabled)
+			{
+				Revoke(self);
+				return;
+			}
+
+			var oldCachedPosition = cachedPosition;
+			cachedPosition = self.CenterPosition;
+
+			var isMoving = !self.IsDead && (movement.IsMoving || (oldCachedPosition - cachedPosition).HorizontalLengthSquared != 0);
```
abcdefg30: This "breaks" with aircraft. They always have `movement.IsMoving = self.World.Map.DistanceAboveTerrain(CenterPosition).Length > 0`.

Question now: Do we want to have this disabled for hovering helicopters (I'd say yes).
This could also be left to another PR, as it seems to be a general "bug" with aircraft.

abcdefg30:

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnMovement.cs
reaperrr: So, should I change the `Aircraft.IsMoving` check to only return `true` if the aircraft is moving (horizontally or vertically)?

pchote: For at least the orca `WithMoveAnimation` case we'll only want to consider horizontal movements.  I believe that we've already ported all the other fiddly uses of `Aircraft.IsMoving` to upgrades so I *think* (but please check) we're otherwise safe to make that change.

reaperrr:
Updated and rebased.

I didn't update the testcase, instead the TD Orca now uses this instead of `WithMoveAnimation`.
Note: I have an upgrade rule ready to get rid of `WithMoveAnimation`, but we need to make `WithAttackAnimation` able to handle multiple sprite body traits before we can use this for actors that use both (like the TS Wolverine).

@abcdefg30 could you revalidate your +1?

pchote:
We need to complete rewrite/rethink the way `WithInfantryBody` etc work with respect to animations, the current approach just doesn't work (and changing WithAttackAnimation etc won't help).  The body trait needs to listen to these things itself so that it can handle the transitions between them (e.g. tiltrotor animations, or liedown/standup).

reaperrr:
Generally speaking I agree, but that sounds like another larger-scale refactor and those usually take time.

abcdefg30:
Looks still good to me. 👍 / ✅

OpenRA.Mods.Common/Traits/Upgrades/UpgradeOnMovement.cs
```diff
@@ -0,0 +1,75 @@
...
+namespace OpenRA.Mods.Common.Traits
+{
+	public class UpgradeOnMovementInfo : UpgradableTraitInfo, Requires<UpgradeManagerInfo>, Requires<IMoveInfo>
+	{
+		[UpgradeGrantedReference, FieldLoader.Require]
+		[Desc("The upgrades to grant.")]
+		public readonly string[] Upgrades = { };
+
+		[Desc("Apply upgrades on straight vertical movement as well.")]
+		public readonly bool ConsiderStraightVerticalMovement = false;
```
pchote: I find this naming confusing.  Lets simplify this to `ConsiderVerticalMovement`?

mods/cnc/rules/aircraft.yaml
```diff
@@ -158,8 +158,15 @@ ORCA:
 	Explodes:
 		Weapon: HeliExplode
 		EmptyWeapon: HeliExplode
-	WithMoveAnimation:
-		MoveSequence: move
+	UpgradeOnMovement:
+		Upgrades: movingforward
+	WithFacingSpriteBody:
```
pchote: Given our more recent discovery / discussion about animations and sprite bodies I think it might be a better idea to keep the Orca as it is, keeping this trait for downstream modders.

pchote (changes requested):
Code changes look fine, and works as advertised so :+1: after we resolve the following.

reaperrr:
Updated as requested.

pchote (approved):

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5e4566ddb90509af8d946ebcd1d028a8dd0a43bd

PR #11887, opened 2016-08-26
TS medic crawl shadow frames correction.

ABrandau:
Small mistake on the shadow frames of the GDI medic for the crawl sequences, it was using the shadow frames of the last idle animation.


pchote:
Fix confirmed, thanks :+1:

GraionDilach:
:+1:

PR #11876, opened 2016-08-24
Official Map Pool Overhaul

SoScared:
Includes:

1) Map alterations (balance, aesthetics) - 16 maps #11143

2) Map preview fixes for lobby viewing (order of spawn letters, visibility of ore mines) - 21 maps ##11698

Closes #11875

PR #11893, opened 2016-08-27
Official Map Pool - Ins and Outs

SoScared:
We ought to take advantage of the work done by the community to refresh the official map pool for the next playtest and release.

Removing:
- Snowy Ridge (3 players) http://resource.openra.net/maps/13064/ (swap with Icy Ridge)
- Equal Opportunity (4 players) http://resource.openra.net/maps/13666/
- First come, First served (4 players) http://resource.openra.net/maps/12328/
- Calm before the storm (4 players) http://resource.openra.net/maps/12596/ (swap with X-Lake)
- Marooned II (6 players) http://resource.openra.net/maps/12791/
- Suffrage (6 players) http://resource.openra.net/maps/12482/
- Great Sahara (8 players) http://resource.openra.net/maps/13088/ (swap with Great Sahara 2)

Adding:
- Barren Land (2 players) http://resource.openra.net/maps/15123/
- Desert Rats (2 players) http://resource.openra.net/maps/15904/
- Icy Ridge (3 players) http://resource.openra.net/maps/15907/ 
- Tandem (4 players) http://resource.openra.net/maps/15910/
- Crossfire (4 players) http://resource.openra.net/maps/15121/
- X-Lake (4 players) http://resource.openra.net/maps/15911/
- Pie of Animosity (6 players) http://resource.openra.net/maps/15117/
- Great Sahara 2 (8 players) http://resource.openra.net/maps/15905/
- Hypothermia (9 players) http://resource.openra.net/maps/15906/
- Nine Lives (9 players) http://resource.openra.net/maps/15161/
- Imminent Destruction (12 players) http://resource.openra.net/maps/15908/

Some of the maps I've reuploaded to accommodate standards of #11876


ChristianDK:
I think it’s extremely important that this bug gets fixed, mainly because you can’t always defend against it and it´s very game breaking too, some maps with oil are barely playable right now. It doesn’t add depth to the game but annoyance, randomness and luck.

As for defending against it:

Defense structures doesn´t necessarily cut it.

-	Flame towers are completely useless as their fireballs are too slow so it will only be possible to counter it if it’s really poorly executed.
-	Pillboxes: It can sometimes be countered with a pillbox but it depends on a few factors, if it’s perfectly timed then it can’t be countered if the engineer jumps out of the right angle (which is random) from the ranger/APC since the delay between each shot is too long. 

Since the engineer can reset the production queue for all production facilities you need to protect both your con yard and your war factory and that’s a big problem.

As for the Con yard: against a soviet player the only foolproof way to secure your con yard is to build walls around it. Spending $400 on walls is a significant expenditure and it will slow you down a lot on some maps. Take a map as Sidestep for example. There’s only 1 oil derrick on that map and if you need to spend $500 on an engineer as well as money on a few riflemen for scouting and protecting the engineer and want to have a defensive structure queued up, then spending $400 on walls is a lot. So it´s only preventable at a significant cost.

Now the main issue is that you can´t really secure your war factory in the same way as the con yard. Even if it was possible the expenditure would be so big that it wouldn´t be worth it.

Also the outcome is very random too and the impact varies tremendously as well (from losing nothing to +30 seconds production queue).

obrakmann:
:+1:

Mailaender:
You sir, picked some very high quality maps here. :tophat:

abcdefg30:
👍 

[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c03418c08207e57480b98cedbe7c1ba9f018729f...7b11ad070685f1133bf19a75a383f7fd9553b80b)

SoScared:
@Mailaender: Thanks! For the record, also thanks to NoobMapMaker for initiating the Map Contest this spring. 7 of the maps came out of that contest.

Do you follow a policy of always shipping custom rule maps unpacked (and with a rules.yaml)? I noticed that trend in my repository's map folder. Map Desert Rats has custom lighting and is merged as .oramap.

Mailaender:
We extracted all Lua or custom rules maps so we can see a diff when running upgrade rules. As running map format updates poses essentially the same problem, we could think about extracting them all. There may have been a discussion in IRC about that topic already.

PR #11899, opened 2016-08-27
Do not apply the cell's Z offset twice during crate positioning.

### LINKED ISSUE #11861: Using CrateSpawner's parachuting-crates option in isometric mods leads to the crate falling until reaching ground level instead of the terrain's actual height.
Might be with regular paradrops too,l haven't checked that yet.

GraionDilach:
Fixes #11861.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Crates/Crate.cs
+++ b/OpenRA.Mods.Common/Traits/Crates/Crate.cs
@@ -144,7 +144,7 @@ namespace OpenRA.Mods.Common.Traits
 		{
 			var cell = self.World.Map.CellContaining(pos);
 			SetLocation(self, cell);
-			SetVisualPosition(self, self.World.Map.CenterOfCell(cell) + new WVec(0, 0, pos.Z));
+			SetVisualPosition(self, self.World.Map.CenterOfCell(cell) + new WVec(WDist.Zero, WDist.Zero, self.World.Map.DistanceAboveTerrain(pos)));
 		}
 
 		// Sets the location (Location) and visual position (CenterPosition)


OpenRA.Mods.Common/Traits/Crates/Crate.cs
```diff
@@ -144,7 +144,7 @@ public void SetPosition(Actor self, WPos pos)
 		{
 			var cell = self.World.Map.CellContaining(pos);
 			SetLocation(self, cell);
-			SetVisualPosition(self, self.World.Map.CenterOfCell(cell) + new WVec(0, 0, pos.Z));
+			SetVisualPosition(self, self.World.Map.CenterOfCell(cell) + new WVec(WDist.Zero, WDist.Zero, self.World.Map.DistanceAboveTerrain(pos)));
```
pchote: The `CPos` version below has the same bug.  A better fix would be to replace the body of this method with `SetPosition(self, self.World.Map.CellContaining(pos))` and then move this change there.

pchote:
This is the same as what `Mobile` does, so looks fine otherwise.

OpenRA.Mods.Common/Traits/Crates/Crate.cs
GraionDilach: The CPos version has no way to offset from the cell's CenterPosition - CenterOfCell returns the cell's Z offset afterall, the bug doesn't exist there.

pchote: Ah right, my bad.

pchote:
:+1:

chrisforbes:
:+1: LGTM, can go ahead and merge this (and do the changelog)

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/e420e451d9ce8dfd14cd838116440994c569cff0

PR #11901, opened 2016-08-27
Fix GpsDots being displayed when their CenterPosition is shrouded

abcdefg30:
Prevents things like:
gpsshroud

### DIFF
--- a/OpenRA.Mods.RA/Effects/GpsDot.cs
+++ b/OpenRA.Mods.RA/Effects/GpsDot.cs
@@ -111,7 +111,7 @@ namespace OpenRA.Mods.RA.Effects
 
 			shouldRenderIndicator = !f2.HasRenderables;
 
-			return f2.Visible && !f2.Shrouded;
+			return f2.Visible && !f2.Shrouded && !toPlayer.World.ShroudObscures(self.CenterPosition);
 		}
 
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.RA/Effects/GpsDot.cs
```diff
@@ -111,7 +111,7 @@ bool IsTargetableBy(Player toPlayer, out bool shouldRenderIndicator)
 
 			shouldRenderIndicator = !f2.HasRenderables;
 
-			return f2.Visible && !f2.Shrouded;
+			return f2.Visible && !f2.Shrouded && toPlayer.Shroud.IsExplored(self.CenterPosition);
```
pchote: It would be better to use `!world.ShroudObscures()` to make sure that you are synced with the player's render state.

abcdefg30:
Updated.

pchote:
Looks fine to me now :+1:

Note that this happens only when you have a gap generator most but not all of a 3x3 building's footprint.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/a2c99ec612f27893a46e01c2b65ee2c4d4fa782b)

PR #11903, opened 2016-08-27
Fix a possible crash in the lua Player.Team property

abcdefg30:
Reproduction case:
1. Use a two player map, first put yourself in as spectator.
2. Remove all bots.
3. Add a bot in the first slot. You go into the second.
4. Execute `Utils.Do({ Player.GetPlayer("Multi0"), Player.GetPlayer("Multi1") }, function(p) Media.DisplayMessage("" .. p.Team) end)`.

### DIFF
--- a/OpenRA.Game/Server/Server.cs
+++ b/OpenRA.Game/Server/Server.cs
@@ -220,10 +220,6 @@ namespace OpenRA.Server
 			}) { IsBackground = true }.Start();
 		}
 
-		/* lobby rework TODO:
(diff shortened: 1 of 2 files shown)


pchote:
I haven't tested this, but it is indeed the valid way to query a client by index so :+1:

obrakmann:
Thanks, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/201827560617981996641d39a839c7c9936173f9)

PR #11904, opened 2016-08-27
Enable per-pixel depth calculations in TS

pchote:
This PR sets up the depth metadata for the remaining cases that I could find, and sets `EnableDepthBuffer: True` for good.

There are probably more cases that i've forgotten, but they will be easier to find, test, and fix with the new rendering method enabled.

Obligatory promo shot:
ts-teaser

### DIFF
--- a/mods/ts/mod.yaml
+++ b/mods/ts/mod.yaml
@@ -111,6 +111,7 @@ TileSets:
 
 MapGrid:
 	TileSize: 48,24
+	EnableDepthBuffer: True
 	Type: RectangularIsometric
 	MaximumTerrainHeight: 16
 	SubCellOffsets: 0,0,0, -256,128,0, 0,-128,0, 256,128,0
--- a/mods/ts/sequences/civilian.yaml
+++ b/mods/ts/sequences/civilian.yaml
@@ -301,6 +301,7 @@ cabhut:
 		DepthSprite: isodepth.shp
 		UseTilesetCode: true
 		Offset: 0, -12, 12
+		ShadowStart: 1
 
 ^cacrsh:
 	idle:
--- a/mods/ts/sequences/infantry.yaml
+++ b/mods/ts/sequences/infantry.yaml
@@ -1,6 +1,7 @@
 ^BasicInfantry:
 	Defaults:
 		Tick: 80
+		Offset: 0, 0, 16
 	stand:
 		Facings: 8
 		ShadowStart: 292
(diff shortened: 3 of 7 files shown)


Mailaender:
I already tested it when it was announced as a merger of all the previous half-baked pull requests and couldn't spot any rendering glitches. :white_check_mark:

GraionDilach:
Requesting the TS-sequence-importer calculating depth offsets.

pchote:
@GraionDilach that doesn't have much to do with this, so should be done in its own PR.

reaperrr:
Things I noticed:
- Laser beams are now often drawn below walls and such, which looks rather bad.
- If an explosion happens on a slope, it clips below adjacent higher cells.
- actors clip below adjacent higher cells, too (watch the shellmap for a while, it becomes more noticable when Nod vehicles come in from the south).

Mailaender:
@GraionDilach maybe this excerpt from IRC helps you: http://logs.openra.net/?year=2016&month=08&day=27#18:28:49

pchote:
> * If an explosion happens on a slope, it clips below adjacent higher cells.
* actors clip below adjacent higher cells, too (watch the shellmap for a while, it becomes more noticable when Nod vehicles come in from the south).

These two are expected - the terrain height calculations are incomplete, placing the simulation ground level on a flat plane at the height of the bottom of the slope.

Do you have a screenshot for the wall case?  They are supposed to collide with the walls?

pchote:
Re lasers: TS actually shows the same thing (notice how the central opaque line is broken by the wall)

tslaser

It looks like WW then draw an always-on-top wider and more translucent beam over it to help hide the issue... sneaker buggers.

Sininini:
Crates are sunk halfway into ground:
applicationframehost_2016-08-28_17-16-24

Tiberium Blossom Tree bottoms are  under ground:
applicationframehost_2016-08-28_17-16-53
How it's supposed to be:
applicationframehost_2016-08-28_17-17-01

Units driving through forests flicker to front and behind of trees.

pchote:
Fixed crates and tib trees.  I can't reproduce any issues with forests.  Bridge huts aren't touched by this PR at all.

pchote:
Added a fix for the bridge hut shadow too, although that's not at all related to this depth buffer stuff.

ROCKNROLLKID:
I don't know if this is specific to this PR, but do you think you can make the lasers more transparent, like they were in the original.

pchote:
I'm not aware of any remaining release-blockers, so it would be good if we could get this merged.

reaperrr:
I guess the remaining visual regressions on ramp tiles and with regard to lasers/other projectiles clipping through walls are tolerable for now, considering this is still an unpublished WIP mod. 
:+1:

PR #11907, opened 2016-08-27
Move resource tick logic into chrome classes.

alercah:
This is towards an end goal of being able to support generic resources by moving cash/resource-specific logic to the chrome.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameCashCounterLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameCashCounterLogic.cs
@@ -9,6 +9,9 @@
  */
 #endregion
 
+using System;
+using System.Linq;
+using OpenRA.Mods.Common.Traits;
 using OpenRA.Traits;
 using OpenRA.Widgets;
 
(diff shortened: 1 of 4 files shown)


OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameCashCounterLogic.cs
```diff
@@ -9,22 +9,68 @@
...
+				nextCashTickTime--;
+
+			var actual = playerResources.Cash + playerResources.Resources;
+
+			var diff = Math.Abs(actual - displayResources);
+			var move = Math.Min(Math.Max((int)(diff * DisplayFracPerFrame), DisplayDeltaPerFrame), diff);
+
+			if (displayResources < actual)
+			{
+				displayResources += move;
```
pchote: You can implement a minor perf / memory churn improvement here by caching `displayLabel = label.F(displayResources)` and binding `cash.GetText = () => displayLabel` in the constructor.

pchote:
Looks good to me, :+1: with our without that suggested fix.

Changing from the display cash to the actual cash in the observer stats was my suggestion: (1) the extra complexity to cover that case isn't worth it (2) it matches the instant updating of the Earned This Minute stat, and (3) if you're actively producing (cash actively ticking down) the change isn't noticeable.

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/62f1e77bad901bb951db3ec1a5cd3bacb7a243f1...8251edc9ded9749f22bdead2f78ac546fae1aa2e)

PR #11909, opened 2016-08-27
Use the MATLAB Jet colormap for displaying depth data.

pchote:
Increases contrast by varying hue and saturation instead of just luminosity.

Uses the algorithm from https://github.com/kbinani/glsl-colormap.

### DIFF
--- a/glsl/color.frag
+++ b/glsl/color.frag
@@ -1,6 +1,21 @@
 varying vec4 vColor;
 uniform bool EnableDepthPreview;
 
+float jet_r(float x)
+{
+	return x < 0.7 ? 4.0 * x - 1.5 : -4.0 * x + 4.5;
+}
+
+float jet_g(float x)
+{
+	return x < 0.5 ? 4.0 * x - 0.5 : -4.0 * x + 3.5;
+}
+
+float jet_b(float x)
+{
+	return x < 0.3 ? 4.0 * x + 0.5 : -4.0 * x + 2.5;
+}
+
 void main()
 {
 	float depth = gl_FragCoord.z;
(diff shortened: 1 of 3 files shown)


glsl/color.frag
```diff
@@ -1,6 +1,18 @@
 varying vec4 vColor;
 uniform bool EnableDepthPreview;
 
+float jet_r(float x) {
```
abcdefg30: Should braces go on a new line here as well (we do it in the rest of this file, too)?

Mailaender:
Doesn't work for me:

image

glsl/color.frag
```diff
@@ -10,8 +22,11 @@ void main()
 
 	if (EnableDepthPreview)
 	{
-		// Front of the depth buffer is at 0, but we want to render it as bright
-		gl_FragColor = vec4(vec3(1.0 - gl_FragDepth), 1.0);
+		float x = 1.0 - gl_FragDepth;
+		float r = clamp(jet_r(x), 0.0, 1.0);
+		float g = clamp(jet_g(x), 0.0, 1.0);
+		float b = clamp(jet_b(x), 0.0, 1.0);
```
WolfGaming: You sure you want to clamp this? And not do just a modulus?

pchote: Quite sure.

pchote:
Fixed braces.  This doesn't depend on #11904, but if you want the game to write non-zero depth data to the GPU then you do of course need to enable the depth buffer.

glsl/color.frag
chrisforbes: GL NDC convention has depth from -1 to 1, not 0 to 1. Are you sure you've got everything lined up how you want it, @pchote ?

pchote: Pretty sure all the numbers are right.  I convert from NDC to window coords in the line above where the diff starts reporting.

reaperrr:
Ingame:
colormap1

Makes it a bit less hard to tell things apart (though I still find myself using pixel-doubling when checking depth data).
:+1:

reaperrr:
@pchote I guess now that #11904 is merged, you could rebase this so the next reviewer doesn't have to enable depth manually, but that's up to you.

obrakmann:
fine with me, :+1:

PR #11912, opened 2016-08-28
Fixed transparency for LaserZap

Mailaender:
This now allows to set initial transparency values, but still does the fade-out effect. The documentation tag was out-dated. We may be better off explaining value types at a separate place instead of individually.

### DIFF
--- a/OpenRA.Mods.Common/Projectiles/LaserZap.cs
+++ b/OpenRA.Mods.Common/Projectiles/LaserZap.cs
@@ -36,7 +36,7 @@ namespace OpenRA.Mods.Common.Projectiles
 
 		public readonly bool UsePlayerColor = false;
 
-		[Desc("Laser color in (A,)R,G,B.")]
+		[Desc("Laser color.")]
 		public readonly Color Color = Color.Red;
 
 		[Desc("Impact animation.")]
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.Common/Projectiles/LaserZap.cs
```diff
@@ -36,7 +36,7 @@ public class LaserZapInfo : IProjectileInfo
 
 		public readonly bool UsePlayerColor = false;
 
-		[Desc("Laser color in (A,)R,G,B.")]
+		[Desc("Laser color.")]
```
pchote: Having the description for "Laser: Color" be "Laser color." doesn't tell us anything useful.  Just remove it.

Mailaender: @Phrohdoh will probably disagree.

GraionDilach:
:+1:

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c2e27ccbc68193f12d4a8186484585f4325345b8...1115b7424b3d838b49d0e360406f76778f16b995)

Mailaender:
> We may be better off explaining value types at a separate place instead of individually.

https://github.com/OpenRA/OpenRA/pull/11913 adds explanations for the color notation.

PR #11915, opened 2016-08-28
Disable modifiers on TS alpha palette.

pchote:
This fixes the black rectangle that appears around lightposts when the map fades in or out from black.


chrisforbes:
:+1:

PR #11918, opened 2016-08-29
Fix mcv husks losing their faction after being captured

### LINKED ISSUE #11491: After repaired from Husk MCV loses its old prerequisites.
As i said in tile. I was Allies and, enemy was Soviet. I killed his MCV repaired it with mechanic but it still provides Allied Structures for me.

I also tested hijacking them for comparasion it works fine. I still get MCVs old owner's building after hijacking.

abc013:
Fixes #11491.

### DIFF
--- a/OpenRA.Mods.Common/Traits/TransformOnCapture.cs
+++ b/OpenRA.Mods.Common/Traits/TransformOnCapture.cs
@@ -20,19 +20,24 @@ namespace OpenRA.Mods.Common.Traits
 		public readonly int ForceHealthPercentage = 0;
 		public readonly bool SkipMakeAnims = true;
 
-		public virtual object Create(ActorInitializer init) { return new TransformOnCapture(this); }
+		public virtual object Create(ActorInitializer init) { return new TransformOnCapture(init, this); }
 	}
 
 	public class TransformOnCapture : INotifyCapture
 	{
 		readonly TransformOnCaptureInfo info;
+		readonly string faction;
 
-		public TransformOnCapture(TransformOnCaptureInfo info) { this.info = info; }
+		public TransformOnCapture(ActorInitializer init, TransformOnCaptureInfo info)
+		{
+			this.info = info;
+			faction = init.Contains<FactionInit>() ? init.Get<FactionInit, string>() : init.Self.Owner.Faction.InternalName;
+		}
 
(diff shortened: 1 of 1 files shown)


pchote:
This just pushes the bug down a level: husks of mcvs that were captured from another player will still lose their faction.  You should store the original faction in the constructor and then use that instead.

pchote:
See https://github.com/OpenRA/OpenRA/blob/master/OpenRA.Mods.Common/Traits/Player/ProvidesPrerequisite.cs#L49 for an example of how to get the faction.

abc013:
Updated.

obrakmann:
Fix confirmed, :+1:

OpenRA.Mods.Common/Traits/TransformOnCapture.cs
```diff
@@ -20,19 +20,24 @@ public class TransformOnCaptureInfo : ITraitInfo
 		public readonly int ForceHealthPercentage = 0;
 		public readonly bool SkipMakeAnims = true;
 
-		public virtual object Create(ActorInitializer init) { return new TransformOnCapture(this); }
+		public virtual object Create(ActorInitializer init) { return new TransformOnCapture(this, init); }
```
Phrohdoh: Should be `init, this`

abc013:
@Phrohdoh changed.

OpenRA.Mods.Common/Traits/TransformOnCapture.cs
```diff
@@ -20,19 +20,24 @@ public class TransformOnCaptureInfo : ITraitInfo
...
+		public virtual object Create(ActorInitializer init) { return new TransformOnCapture(init, this); }
 	}
 
 	public class TransformOnCapture : INotifyCapture
 	{
 		readonly TransformOnCaptureInfo info;
+		readonly string faction;
 
-		public TransformOnCapture(TransformOnCaptureInfo info) { this.info = info; }
+		public TransformOnCapture(TransformOnCaptureInfo info, ActorInitializer init)
```
GraionDilach: You also have to change order here then.

abc013:
Thanks!

pchote:
:+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ce0e7ddd03dfe066e80f5bea279e4423f227ffcb

PR #11920, opened 2016-08-30
Remove the bogus Health property from CAMERA

abcdefg30:
It just doesn't make sense and does not appear to be used anyway.

Removing this just fixes bugs, like for example that yak's try to target them in our lua missions:
`camera 471 is an invalid target for yak 473!`
(This can cause the game to freeze when a player has only cameras left.)


obrakmann:
Cameras still seem to work in missions and with nukes, so :+1:

Mailaender:
This is a regression introduced during code review https://github.com/OpenRA/OpenRA/pull/11679#discussion_r75598410 so no changelog.

PR #11924, opened 2016-08-30
Fix NRE in SmudgeLayer.AddSmudge

### LINKED ISSUE #11922: System.ArgumentNullException in SmudgeLayer.AddSmudge
You can find the replay and map at https://github.com/abcdefg30/OpenRA/tree/crash-mission.
```
Red Alert Mod at Version {DEV_VERSION}
on map fc07e13e80de569874bfa533198258d126962762 (Infiltration by Scott_NZ).
Date: 2016-08-30 11:11:11Z
Exception of type `System.ArgumentNullException`: The value must not be NULL. (Der Wert darf nicht NULL sein.)
Parametername: key
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at OpenRA.Mods.Common.Traits.SmudgeLayer.AddSmudge(CPos loc)
   at OpenRA.Mods.Common.Warheads.LeaveSmudgeWarhead.DoImpact(Target target, Actor firedBy, IEnumerable`1 damageModifiers)
   at OpenRA.GameRules.WeaponInfo.<>c__DisplayClassa.<Impact>b__6() in \OpenRA\OpenRA.Game\GameRules\WeaponInfo.cs:Line 162.
   at OpenRA.Effects.DelayedAction.<Tick>b__0(World w) in \OpenRA\OpenRA.Game\Effects\DelayedAction.cs:Line 32.
   at OpenRA.World.Tick() in \OpenRA\OpenRA.Game\World.cs:Line 347.
   at OpenRA.Game.InnerLogicTick(OrderManager orderManager) in \OpenRA\OpenRA.Game\Game.cs:Line 561.
   at OpenRA.Game.LogicTick() in \OpenRA\OpenRA.Game\Game.cs:Line 586.
   at OpenRA.Game.Loop() in \OpenRA\OpenRA.Game\Game.cs:Line 713.
   at OpenRA.Game.Run() in \OpenRA\OpenRA.Game\Game.cs:Line 753.
   at OpenRA.Program.Run(String[] args) in \OpenRA\OpenRA.Game\Support\Program.cs:Line 119.
   at OpenRA.Program.Main(String[] args) in \OpenRA\OpenRA.Game\Support\Program.cs:Line 40.

### PR DESCRIPTION
**Pull request by obrakmann**
The cause for this issue was indeed a regression from #11721.

It turns out that there may be situations\* where the else-branch (the "there already is a smudge here"-case) in `AddSmudge` is taken even though there isn't a smudge (when the location in question is contained in both the `dirty` and `tiles` arrays, but `RemoveSmudges` ran previously for the same location).

In that case, the `Tile` field is null, but gets referenced in that else-branch, producing the NRE.

There are two possible ways to fix this. The first is to set the `Tile` field to a random entry in `RemoveSmudges`, the better way is to check the smudge for the "deleted" marker, which is a `Sprite` field set to `null`.

Fixes #11922.

*Didn't delve too deeply into it, but in my replay a building was sold while oil barrels around it were being blown up, so I assume it's some kind of race condition where `RemoveSmudge` and `AddSmudge` run during the same tick for the same location.

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
+++ b/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
@@ -142,6 +142,7 @@ namespace OpenRA.Mods.Common.Traits
 			if (Game.CosmeticRandom.Next(0, 100) <= Info.SmokePercentage)
 				world.AddFrameEndTask(w => w.Add(new SpriteEffect(world.Map.CenterOfCell(loc), w, Info.SmokeType, Info.SmokeSequence, Info.SmokePalette)));
 
+			// A null Sprite indicates a deleted smudge.
 			if ((!dirty.ContainsKey(loc) || dirty[loc].Sprite == null) && !tiles.ContainsKey(loc))
 			{
 				// No smudge; create a new one
@@ -151,6 +152,7 @@ namespace OpenRA.Mods.Common.Traits
 			else
 			{
 				// Existing smudge; make it deeper
+				// A null Sprite indicates a deleted smudge.
 				var tile = dirty.ContainsKey(loc) && dirty[loc].Sprite != null ? dirty[loc] : tiles[loc];
 				var maxDepth = smudges[tile.Type].Length;
 				if (tile.Depth < maxDepth - 1)
@@ -167,6 +169,7 @@ namespace OpenRA.Mods.Common.Traits
 		{
 			var tile = dirty.ContainsKey(loc) ? dirty[loc] : new Smudge();
 
+			// Setting Sprite to null to indicate a deleted smudge.
 			tile.Sprite = null;
 			dirty[loc] = tile;
 		}
@@ -178,6 +181,7 @@ namespace OpenRA.Mods.Common.Traits
 			{
 				if (!self.World.FogObscures(kv.Key))
 				{
+					// A null Sprite indicates a deleted smudge.
 					if (kv.Value.Sprite == null)
 						tiles.Remove(kv.Key);
 					else
--- a/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
+++ b/OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
@@ -151,7 +151,7 @@ namespace OpenRA.Mods.Common.Traits
 			else
 			{
 				// Existing smudge; make it deeper
-				var tile = dirty.ContainsKey(loc) ? dirty[loc] : tiles[loc];
+				var tile = dirty.ContainsKey(loc) && dirty[loc].Sprite != null ? dirty[loc] : tiles[loc];
 				var maxDepth = smudges[tile.Type].Length;
 				if (tile.Depth < maxDepth - 1)
 				{

### CODE REVIEW

File: OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs [hunk 1]
```diff
@@ -151,7 +151,7 @@ public void AddSmudge(CPos loc)
 			else
 			{
 				// Existing smudge; make it deeper
-				var tile = dirty.ContainsKey(loc) ? dirty[loc] : tiles[loc];
```
pchote: Can you please add a comment '// .Sprite will be null if the smude has just been removed'?

RoosterDragon:
@pchote raised that exact problem in the original PR: (read the whole comment chain) https://github.com/OpenRA/OpenRA/pull/11207#discussion_r61749736 when I suggested this same simplification that would've resulted in an error like this.

I think #11721 fell for the same trap whilst solving the desync issue.

If you call `RemoveSmudge` and then `AddSmudge` again before `TickRender` has a chance to process the dirty cells, you an end up with a "deleted" smudge that exists (in the dictionary, but a null `Sprite`) and I'm not sure we updated all the code to handle that correctly. 

This *appears* to be the right fix according to that train of thought, however, the stack trace in #11922 says that a null value was passed to a dictionary indexer, and I don't see how the current change of checking the for the null `Sprite` helps, as that is not used elsewhere in this method, let alone in an indexer. Hopefully I'm just missing something here and you can point out the flaw in my thinking :).

(I wish we had unit tests for this trait)

obrakmann:
@RoosterDragon: we use the `Sprite` field as a marker. If it's null, it means the smudge has been deleted, and all its other fields will be default values as well (null, 0, whatever. In the crash case, `Tile` was dereferenced in the line right below the one touched by this PR). So if `Sprite` is null, we go ahead and select a smudge from the `tiles` array instead, which only contains valid smudges.

OpenRA.Mods.Common/Traits/World/SmudgeLayer.cs
obrakmann: Done

RoosterDragon:
Ah I see, `Sprite` being null would indicate `Type` would be null too, if `RemoveSmudge` operated on a smudge not yet dirtied since that would do `new Smudge()`.

I feel the design is needlessly confusing overall, but I can see the fix is correct for that case now.

So :+1:

pchote:
> (I wish we had unit tests for this trait)

Integration tests would do just as well here, and would avoid the design-, perf-, and soul-destroying process of wrapping every piece of interesting logic with mockable interfaces :smile:

abcdefg30:
Fix confirmed. 👍 / ✅

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/2366cd39649ec58a2c0ca47404ea890490a0d908...f96c45c40e976d20e71f8f94e58f56bb9e4e896f)

PR #11928, opened 2016-08-31
Rename stand3 to liedown

### LINKED ISSUE #11703: Infantry sequence stand3 is unused
Not sure what it is for.

abcdefg30:
`liedown` appears to be what `stand3` is.
(I also moved `liedown` next to/right above `standup`.)

Closes #11703.

### DIFF
--- a/mods/ra/sequences/infantry.yaml
+++ b/mods/ra/sequences/infantry.yaml
@@ -30,7 +30,7 @@ e1:
 		Start: 128
 		Length: 2
 		Facings: 8
-	standup-0:
+	standup:
 		Start: 176
 		Length: 2
 		Facings: 8
@@ -868,7 +868,7 @@ gnrl:
 		Start: 104
 		Length: 4
 		Facings: 8
-	standup-0:
+	standup:
 		Start: 136
 		Length: 2
 		Facings: 8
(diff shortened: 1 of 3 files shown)


Mailaender:
:+1: to giving this a better name, but how does that resolve the issue?

abcdefg30:
This kind of makes #11703 a duplicate of #6330 (and thus I'd close #11703).

mods/ra/sequences/infantry.yaml
```diff
@@ -29,6 +26,9 @@ e1:
 		Length: 4
 		Facings: 8
 		Tick: 100
+	liedown:
+		Start: 128
+		Length: 16
```
pchote: These should all be Length: 2 and Facings: 8, like the standup animations.

mods/ra/sequences/infantry.yaml
```diff
@@ -29,6 +26,9 @@ e1:
...
 	standup-0:
```
pchote: I have no idea why I called this `standup-0` back in befe22e170b1a06143f59da9880cbda3a71f5757, but this has been randomly copy pasta-ed to a few more places.  Can you please rename them to `standup` for consistency while we're here.

pchote (changes requested):

abcdefg30:
Sorry for the delay. Updated.

pchote (approved):

reaperrr:
Does this need a changelog entry?

pchote:
I don't think so.

PR #11932, opened 2016-09-01
Updated SharpFont to version 4.0.1

Mailaender:
See https://www.nuget.org/packages/SharpFont/4.0.1 for a changelog.

### DIFF
--- a/thirdparty/fetch-thirdparty-deps.ps1
+++ b/thirdparty/fetch-thirdparty-deps.ps1
@@ -50,8 +50,8 @@ if (!(Test-Path "MaxMind.GeoIP2.dll"))
 if (!(Test-Path "SharpFont.dll"))
 {
 	echo "Fetching SharpFont from NuGet."
-	./nuget.exe install SharpFont -Version 3.1.0 -ExcludeVersion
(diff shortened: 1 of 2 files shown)


pchote:
This probably wants to wait for Next + 1.  Doing a major-version update on a dependency immediately before shipping a new release is a bad idea.  There doesn't seem to be a changelog for 4.0.0, so its hard to say what this actually gives us.

Mailaender:
4.0.1 was just a re-release https://github.com/Robmaister/SharpFont/compare/v4.0.0...v4.0.1

pchote:
Freetype 2.7 has just come out, do you know if we expect any SharpFont changes to go with that?

Mailaender:
No, I am not really in contact with the author. Just updated this, because I checked all dependencies systematically for old versions.

pchote:
Ok.

pchote (approved):
No objections here, and works fine on OSX. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1d315e8323162d3ec6f9c0b21e126773050eac71)

PR #11934, opened 2016-09-01
Updated NUnit to version 3.4.1

Mailaender:
See https://github.com/nunit/docs/wiki/Release-Notes for a changelog.

### DIFF
--- a/thirdparty/fetch-thirdparty-deps.sh
+++ b/thirdparty/fetch-thirdparty-deps.sh
@@ -70,8 +70,8 @@ fi
 
 if [ ! -f nunit.framework.dll ]; then
 	echo "Fetching NUnit from NuGet"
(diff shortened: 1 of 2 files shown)


thirdparty/fetch-thirdparty-deps.sh
```diff
@@ -70,8 +70,8 @@ fi
 
 if [ ! -f nunit.framework.dll ]; then
 	echo "Fetching NUnit from NuGet"
-	get NUnit 3.0.1
-	cp ./NUnit/lib/net40/nunit.framework* .
+	get NUnit 3.4.1
```
pchote: `NUnit.Console` has also been updated to 3.4.1, so I guess we'll want to keep these in sync.

pchote (changes requested):

PR #11937, opened 2016-09-01
Lobby now retains last selected Faction

Pritchy96:
The Lobby screen now auto selects the last faction you played as for any mod. Does not disturb missions or locked faction selections, as far as I can tell.


OpenRA.Game/Server/Server.cs
```diff
@@ -306,7 +306,7 @@ void ValidateClient(Connection newConn, string data)
 					Slot = LobbyInfo.FirstEmptySlot(),
 					PreferredColor = handshake.Client.Color,
 					Color = handshake.Client.Color,
-					Faction = "Random",
+					Faction = handshake.Client.Faction,
```
pchote: These options are hardcoded to avoid exploits where players could set state that is restricted by the map/mod settings or bogus state to crash other clients.  You'll need to validate that the faction is valid before assigning this here.

OpenRA.Game/Network/UnitOrders.cs
```diff
@@ -152,15 +157,20 @@ public static void ProcessOrder(OrderManager orderManager, World world, int clie
 						}
 
 						Game.Settings.Player.Name = Settings.SanitizedPlayerName(Game.Settings.Player.Name);
+
+						var faction = "Random";
+
+						if (Game.Settings.Game.LastFaction.ContainsKey(mod.Id))
```
pchote: You can use Dictionary.TryGetValue to check and assign in one operation.

OpenRA.Game/Settings.cs
```diff
@@ -159,6 +159,8 @@ public class GameSettings
 		public string Mod = "modchooser";
 		public string PreviousMod = "ra";
 
+		public Dictionary<string, string> LastFaction = new Dictionary<string, string>();
```
pchote: This will eventually want to move into mod-specific settings, but I don't have a big problem with putting this here for now.

OpenRA.Game/Network/UnitOrders.cs
```diff
@@ -105,6 +105,11 @@ public static void ProcessOrder(OrderManager orderManager, World world, int clie
 							break;
 						}
 
+						// Save most recently chosen Faction.
+						var client = orderManager.LobbyInfo.ClientWithIndex(clientId);
+						Game.Settings.Game.LastFaction[Game.ModData.Manifest.Id] = client.Faction;
```
pchote: This looks like the wrong place to be doing this.  You probably want to do this from the lobby code instead, and will also need to be careful about the preferred / server state (see how we handle player colors).

OpenRA.Game/Server/Server.cs
Pritchy96: I've been having a crack at this and am trying to wrap my head around the code base, and am having a little trouble. The way I was looking at doing it was checking to see whether the faction name is present in the array of factions within the world.map.rules ala as done in Player.cs's ChooseFaction(), as I see it appears to return selectable factions only. Am I on the right lines? I must apologise for asking what feels like blind questions, I'm just trying to wrap my head around the code and trying to get to a point to where I can contribute effectively.

pchote: The `SyncClientToPlayerReference` immediately below already handles the locked faction case, so we only need to worry about maliciously set state.

If you follow the code flow a bit further down still you'll see that it notifies any ServerTraits that implement `IClientJoined`.  One of these is `LobbyCommands`, which has a check to validate the player color inside its `ClientJoined` method.  The best place for the selectable factions check is in there.

You can get a list of all the valid factions with
```csharp
var factions = server.Map.Rules.Actors["world"].TraitInfos<FactionInfo>()
	.Where(f => f.Selectable).Select(f => f.InternalName);
```
and then do something like
```csharp
if (!factions.Contains(client.Faction))
	client.Faction = "Random";
```

PR #11939, opened 2016-09-01
RA - Some AI Improvements

MustaphaTR:
**Needs Testing, Values May Change!**

I tested a little bit to be sure if adding Chrono Tank, Radar Jammer and Gap Generator is a good idea and looked fine also fixed problems i see. I'm not sure about decreasing Airfield/Helipad Count for Rush AI or add Planes/Helis to them but i've gone with first.

### DIFF
--- a/mods/ra/rules/ai.yaml
+++ b/mods/ra/rules/ai.yaml
@@ -169,11 +169,16 @@ Player:
 			atek: 1%
 			stek: 1%
 			mslo: 1%
+		CapturingActorTypes: e6
+		CapturableActorTypes: oilb, hosp
+		CheckCaptureTargetsForVisibility: false
+		MaximumCaptureTargetOptions: 5
 		UnitsToBuild:
 			e1: 65%
 			e2: 25%
 			e3: 40%
 			e4: 15%
+			e6: 3%
 			medi: 3%
 			mech: 3%
 			dog: 15%
@@ -201,6 +206,7 @@ Player:
 			pt: 10%
 		UnitLimits:
 			dog: 4
+			e6: 1
 			medi: 3
 			mech: 3
 			harv: 8
@@ -303,11 +309,16 @@ Player:
 			atek: 1%
 			stek: 1%
 			mslo: 1%
+		CapturingActorTypes: e6
+		CapturableActorTypes: oilb, hosp
+		CheckCaptureTargetsForVisibility: false
+		MaximumCaptureTargetOptions: 10
 		UnitsToBuild:
 			e1: 65%
 			e2: 25%
 			e3: 40%
 			e4: 15%
+			e6: 3%
 			medi: 5%
 			mech: 5%
 			dog: 15%
@@ -335,6 +346,7 @@ Player:
 			pt: 10%
 		UnitLimits:
 			dog: 4
+			e6: 1
 			medi: 4
 			mech: 4
 			harv: 8
--- a/mods/ra/rules/ai.yaml
+++ b/mods/ra/rules/ai.yaml
@@ -43,7 +43,7 @@ Player:
 			fix: 0.1%
 			dome: 10%
 		CapturingActorTypes: e6
-		CapturableActorTypes: mslo, gap, iron, pdox, atek, weap, fact, proc, silo, hpad, afld, powr, apwr, stek, barr, kenn, tent, fix, oilb, hosp
+		CapturableActorTypes: mslo, gap, iron, pdox, atek, weap, fact, proc, silo, hpad, afld, powr, apwr, stek, barr, kenn, tent, dome, fix, oilb, hosp
 		CheckCaptureTargetsForVisibility: false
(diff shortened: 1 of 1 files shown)


mods/ra/rules/ai.yaml
```diff
@@ -190,6 +197,9 @@ Player:
 			4tnk: 15%
 			ttnk: 15%
 			stnk: 10%
+			ctnk: 2.5%
+			mrj: 0.25%
+			mgg: 0.1%
```
pchote: These low percentages are effectively meaningless.

MustaphaTR: They still build them. I don't think they shoud use them too often.

MustaphaTR:
When i'm in it, i want to ask, should i add Bio List to capture list too and add Zombies and Ant (Especially Ants they're pretty powerful) to unit list.

Also i think we should make all AIs build Engineers to capture Tech Structures not just Rush AI.

mods/ra/rules/ai.yaml
pchote: The AI will build anything as long as the percentage is less than the specified value.  0.1% is going to be no different to 1% 5% or even 100% thanks to the UnitLimits you specify above.

MustaphaTR: So how much should i use? 1%, 5%?

GraionDilach:
By that logic, AIs should build Hijackers as well, but then the AI-capturing code should be rewritten to have an actor - capturable options dictionary instead of the global list.

MustaphaTR:
> By that logic, AIs should build Hijackers as well, but then the AI-capturing code should be rewritten to 
> have an actor - capturable options dictionary instead of the global list.

Sorrry, i'm can't and not gonna touch there.

GraionDilach:
A second issue which came to my mind regarding https://github.com/OpenRA/OpenRA/pull/11939/commits/768858958bdc73c6085bf4a86145bec521f9f093 - on maps without tech buildings/no tech buildings present, the AI will still train engineers but they'll be sitting ducks.

MustaphaTR:
Just one engineer tho. Also this is more like another problem with the AI-capturing code. Sounds like a simple check can fix that. I can't do it tho. Can you send a pulll request for it either to my fork or as another pull request.

jaZzKCS:
Doesn't the AI keep building the units at the set rate (percentages)? So it will not be only one engineer.

MustaphaTR:
You can set some units upper limits. I gave 1 for Engineer of Normal and Turtle so they'll only build 1.

MustaphaTR:
Should i squash all these commits into one or just 3e3b6d8858fe7bab2d4011a5e3f7567d9c50d1d8 to cc7eaed8bd31765ccd87b024f1ff7f3b6ed72168 would be enough?

MustaphaTR:
Squashed that 2 commits.

MustaphaTR:
Updated.

abcdefg30:

MustaphaTR:
Updated as discussed in IRC.

MustaphaTR:
I know the MGG and MRJ is not working as good as supposed and i'm not sure about 75a963a as there are plans to remove Bio Lab's ability to give Ants and Zombies, i still want to see this merged before the next playtest.

pchote:
> I know the MGG and MRJ is not working as good as supposed and i'm not sure about 75a963a as there are plans to remove Bio Lab's ability to give Ants and Zombies

We're now in the final stretch of the last (hopefully) playtest, so we can't take this while it has unresolved issues and don't want to hold up the release over it.

MustaphaTR:
I can remove/split the MGG/MRJ commit if necessary. Other ones don't have much impact on balance/AI power much.

obrakmann:
I think it's too late now for this to be included in the upcoming release. The current playtest already hardly gets any plays, subsequent ones will get even less.

obrakmann:
I'm not a fan of 75a963ab17e883456ba1ff8a3e68ebb406ed6649. Also, what's the reasoning behind e9bd7f5c087e6f5fb5c88947300f75cc24d564cb? The rest is fine with me.

MustaphaTR:
About e9bd7f5 , it is here because Rush AI doesn't build air units. HPAD and AFLD only makes AI build unnecessary buildings and slow them down.

MustaphaTR:
Updated.

MustaphaTR:
Splited the PR to two. Ones that looks complately fine to me are moved to  #12216.

reaperrr:
@MustaphaTR could you split 57f84b2 to a separate PR as well? I suspect these units might have similar issues like medics/mechanics.

The capturing-related commits make sense, but I haven't tested this yet.

MustaphaTR:
I'll take a look when i find time but i probably won't in a while. I had an operation last week and i can't use the computer. I may do it on mobile tho, it doesn't need much stuff.

PR #11941, opened 2016-09-01
Fix mouse input breakage

### LINKED ISSUE #11940: Add a small deadzone for standard scrolling
https://github.com/OpenRA/OpenRA/pull/11879 seems to have exposed issue with lack of deadzone for standard scrolling when it's used on right mouse button. Ever so slight mouse movements while ordering units around count as standard scrolling making orders not go through.

obrakmann:
This re-introduces the bug from #11849 for joystick-scrolling and fixes #11940 for standard-scrolling.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
@@ -35,6 +35,7 @@ namespace OpenRA.Mods.Common.Widgets
 		public int EdgeCornerScrollThreshold = 35;
 
 		int2? joystickScrollStart, joystickScrollEnd;
+		int2? standardScrollStart;
 		bool isStandardScrolling;
 
 		static readonly Dictionary<ScrollDirection, string> ScrollCursors = new Dictionary<ScrollDirection, string>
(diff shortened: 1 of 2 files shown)


pchote:
This seems to regress the scroll-cursor while drag-scrolling too.  If you could show that while outside the deadzone it would make the two modes feel more consistent.

abcdefg30:

OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
```diff
@@ -286,7 +286,8 @@ public override bool HandleMouseInput(MouseInput mi)
 
 			if (scrollType == MouseScrollType.Standard || scrollType == MouseScrollType.Inverted)
 			{
-				if (mi.Event == MouseInputEvent.Move)
+				if (mi.Event == MouseInputEvent.Move &&
+					(isStandardScrolling || (Viewport.LastMousePos - mi.Location).Length > Game.Settings.Game.MouseScrollDeadzone))
```
pchote: This needs to be relative to the start of the move, not the last pos. If you move your mouse slow enough it will never trigger.

obrakmann:
Updated.

pchote:
:+1: and https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/64f86829256bcef6467dac57a1ffe20149eb9299.

PR #11946, opened 2016-09-02
Fix an inverted check in SupportPowerChargeBar

abcdefg30:
It showed the charge bar to enemies, but not allies.
You could also notice it missing when selecting the "All Players" view in replays.


obrakmann:
:+1:

pchote:
:+1: and https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/1aa7afdf2cb5005e2d0def61365c0c25d830caa6

PR #11949, opened 2016-09-03
Polished the Dune 2000 infantry sequences

Mailaender:
Actually contributed by [-Jes-](http://www.sleipnirstuff.com/forum/memberlist.php?mode=viewprofile&u=3081). Sent to me via private message on @pchote's broken forums in January. It is a combination of e-Mail notifications not working and the default theme also not displaying new private messages so sorry for the delay. The SubSilver skin seems to works though. See https://github.com/OpenRA/OpenRAWeb/issues/48 https://github.com/OpenRA/OpenRAWeb/issues/47.

### DIFF
--- a/mods/d2k/rules/defaults.yaml
+++ b/mods/d2k/rules/defaults.yaml
@@ -192,6 +192,7 @@
 	QuantizeFacingsFromSequence:
 		Sequence: stand
 	WithInfantryBody:
+		IdleSequences: idle1, idle2
 	TakeCover:
 		DamageModifiers:
 			Prone50Percent: 50
--- a/mods/d2k/sequences/infantry.yaml
+++ b/mods/d2k/sequences/infantry.yaml
@@ -43,24 +43,24 @@ light_inf:
 		Facings: -8
 		Transpose: true
 	die1: DATA.R8
-		Frames: 382, 389, 396, 403, 410, 417, 424, 431, 438, 439, 440, 441
+		Frames: 382, 389, 396, 403, 410, 417, 424, 431, 438, 443, 448, 453
 		Length: 12
 		Tick: 80
 	die2: DATA.R8
-		Frames: 383, 390, 397, 404, 411, 418, 425, 432
-		Length: 8
+		Frames: 383, 390, 397, 404, 411, 418, 425, 432, 439, 444, 449, 454
+		Length: 12
 		Tick: 80
 	die3: DATA.R8
-		Frames: 384, 391, 398, 405, 412, 419, 426, 433
-		Length: 8
+		Frames: 384, 391, 398, 405, 412, 419, 426, 433, 440, 445, 450, 455
+		Length: 12
 		Tick: 80
 	die4: DATA.R8
-		Frames: 385, 392, 399, 406, 413, 420, 427, 434
-		Length: 8
+		Frames: 385, 392, 399, 406, 413, 420, 427, 434, 441, 446, 451, 456
+		Length: 12
 		Tick: 80
 	die-crushed: DATA.R8
-		Frames: 386, 393, 400, 407, 414, 421, 428, 435
-		Length: 8
+		Frames: 386, 393, 400, 407, 414, 421, 428, 435, 442, 447, 452, 457
+		Length: 12
 		Tick: 800
 		ZOffset: -511
 	icon: DATA.R8
--- a/mods/d2k/sequences/infantry.yaml
+++ b/mods/d2k/sequences/infantry.yaml
@@ -3,6 +3,14 @@ light_inf:
 		Start: 206
 		Facings: -8
 		Transpose: true
+	idle1: DATA.R8
+		Frames: 387, 394, 401, 408, 415, 422, 429, 436
+		Length: 8
+		Tick: 80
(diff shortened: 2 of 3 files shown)


mods/d2k/sequences/infantry.yaml
```diff
@@ -35,24 +43,24 @@ light_inf:
 		Facings: -8
 		Transpose: true
 	die1: DATA.R8
-		Frames: 382, 389, 396, 403, 410, 417, 424, 431, 438, 439, 440, 441
+		Frames: 382, 389, 396, 403, 410, 417, 424, 431, 438, 443, 448, 453
```
pchote: This doesn't look right.  The frames in this sequence increment by 7, and it wouldn't make sense for the original game to change stride half way through the animation.  By that logic these should be using frames 445, 452, 459.

mods/d2k/sequences/infantry.yaml
```diff
@@ -35,24 +43,24 @@ light_inf:
...
 		Length: 12
 		Tick: 80
 	die2: DATA.R8
-		Frames: 383, 390, 397, 404, 411, 418, 425, 432
-		Length: 8
+		Frames: 383, 390, 397, 404, 411, 418, 425, 432, 439, 444, 449, 454
```
pchote: Likewise here: 439, 445, 453, 450.

mods/d2k/sequences/infantry.yaml
```diff
@@ -35,24 +43,24 @@ light_inf:
...
+		Length: 12
 		Tick: 80
 	die3: DATA.R8
-		Frames: 384, 391, 398, 405, 412, 419, 426, 433
-		Length: 8
+		Frames: 384, 391, 398, 405, 412, 419, 426, 433, 440, 445, 450, 455
```
pchote: ... and so on for the rest.

mods/d2k/rules/defaults.yaml
```diff
@@ -192,6 +192,7 @@
 	QuantizeFacingsFromSequence:
 		Sequence: stand
 	WithInfantryBody:
+		IdleSequences: idle1, idle2
```
pchote: Please also remove the now-redundant definition at infantry.yaml#L192.

pchote:
I can confirm that the new idle animations look correct, but see the other inline comments.

mods/d2k/sequences/infantry.yaml
Mailaender: You mean  439, 446, 453, 460?

mods/d2k/sequences/infantry.yaml
Mailaender: Have you checked your proposed change? It will cause glitches.

mods/d2k/sequences/infantry.yaml
pchote: Yes.

mods/d2k/sequences/infantry.yaml
pchote: No, i'm just pointing out that the original game couldn't have used the proposed frame ranges.  I'm relying on you to correct it.

Mailaender: In this case your suggestion was misleading. Apparently the corpse decaying animation uses 5 frame steps for whatever reason.

mods/d2k/rules/defaults.yaml
Mailaender: Removed.

Mailaender:
Updated.

mods/d2k/sequences/infantry.yaml
pchote: That means that those additional frames are not part of the light inf death animation, but a separate corpse sequence.

pchote: This then makes me strongly suspect that all infantry will need similar fixes, because it wouldn't make sense for light infantry to do something special.  Can you please check this?

Mailaender: There isn't really a pattern here. For example the rocket trooper death animation has a 1 step death decaying animation https://github.com/OpenRA/OpenRA/pull/11949/files#diff-6211bf63a321e3de7a22a5294c07c2e5R114 which we already set up.

pchote: Ok then.

pchote:
:+1:

obrakmann:
How are those animations supposed to look? With a facing of 0 (north), some units (rifle infantry, for example) make a 180° turn for some animations, which looks weird. Is that intentional?

pchote:
Do you mean for the idle animations? All the mods do that.

obrakmann:
Well, I guess you can tell how closely I look at those things. It's really weird though, when you concentrate on it. Anyway, :+1: then.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/06406e35fdde7c1b5b4dc5200bc151651085c55c)

Mailaender:
I believe Dune 2000 might already have directional idle animations or some precursor to that. It is most notable with the Grenadier juggling.

PR #11950, opened 2016-09-03
Fixed building placement on Tiberian Sun low bridges

Mailaender:
A followup of https://github.com/OpenRA/OpenRA/pull/11441 that fixes some oversights.

### DIFF
--- a/OpenRA.Mods.TS/Traits/Buildings/LowBridge.cs
+++ b/OpenRA.Mods.TS/Traits/Buildings/LowBridge.cs
@@ -20,7 +20,7 @@ namespace OpenRA.Mods.TS.Traits
 	[Desc("Bridge actor that can't be passed underneath.")]
 	class LowBridgeInfo : ITraitInfo, Requires<BuildingInfo>
 	{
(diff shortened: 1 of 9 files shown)


pchote:
:+1: fix confirmed.

pchote:
This is an easily reviewable PR. It would be good to get this merged.

OpenRA.Mods.TS/Traits/Buildings/LowBridge.cs
```diff
@@ -20,7 +20,7 @@ namespace OpenRA.Mods.TS.Traits
 	[Desc("Bridge actor that can't be passed underneath.")]
 	class LowBridgeInfo : ITraitInfo, Requires<BuildingInfo>
 	{
-		public readonly string TerrainType = "Road";
+		public readonly string TerrainType = "Bridge";
```
reaperrr: You need to define actor speeds for this new terrain type, currently this makes low bridges unpassable by all actors.

Mailaender: Fixed.

PR #11951, opened 2016-09-03
Rework external capture -> production interaction

### LINKED ISSUE #8122: Disabled production when a factory gets captured
Introduced in #6766. The problematic parts are [lines 56-58 in `ClassicProductionQueue.cs`]( https://github.com/OpenRA/OpenRA/blob/50d726ea317f643fd35fd706d5dc9584d3073d8a/OpenRA.Mods.Common/Traits/Player/ClassicProductionQueue.cs#L56-L58).

The *real* problem is that both `Transform` and `ExternalCapture` lock the building and so are indistinguishable to other code and aren't treated differently.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13930430-disabled-production-when-a-factory-gets-captured?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #8146: ProductionQueue-transform-freeze still alive
transformfreeze
(I deployed one conyard using hotkeys while placing the power plant.)

You need just two conyards.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/14280464-productionqueue-transform-freeze-still-alive?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #8704: MCV can't undeploy
[03:28] < pr37> the mcv bug still exists.  map: MT. Glorious V1.1; replay: http://s000.tinyupload.com/index.php?file_id=00602654922471369806
[03:28] < pr37> player dent, at around 8:50.
[03:30] < pr37> placed a helipad just as i was undeploying the mcv, which automatically redeployed and thereafter couldn't undeploy it any longer.
[03:31] < pr37> openra release-20150614

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24164942-mcv-can-t-undeploy?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #8122, which was a side-effect of an incorrect fix for a bug that triggered when a building was placed while a construction yard was being sold.  Also fixes #8146 and resolves #8704.

First off, this _properly_ fixes the sell-place-animation bug by adding proper checks to `WithBuildingPlacedAnimation`. It then changes `BaseProvider` to disable building placement around a locked construction yard, and `Production` to disable units from being produced from locked factories.  Finally, it removes the tech-disabling hack that causes #8122.

In practice this should give the following gameplay changes in RA:
- When capturing is triggered on a player's only construction yard: production of structures will continue but they cannot be placed around it (previously: construction was cancelled)
- When capturing is triggered on a construction yard, and the player has multiple: structures cannot be placed around the locked yard (previously: could be placed there)
- When capturing is triggered on a player's only factory: production of units will continue but they will stop at the "Ready" phase, like when the exit is blocked (previously: construction was cancelled)
- When capturing is triggered on a factory, and the player has multiple: units will not come out of the locked factory, even if it is the primary (previously: they would).

Mods that don't use external capturing shouldn't be affected, because locking is only otherwise used as part of things that permanently disable the actor (transform, sell, instant capture).

### DIFF
--- a/OpenRA.Mods.Common/Traits/Production.cs
+++ b/OpenRA.Mods.Common/Traits/Production.cs
@@ -28,18 +28,25 @@ namespace OpenRA.Mods.Common.Traits
 		public virtual object Create(ActorInitializer init) { return new Production(init, this); }
 	}
 
-	public class Production
+	public class Production : INotifyCreated
 	{
 		readonly Lazy<RallyPoint> rp;
 
 		public readonly ProductionInfo Info;
 		public string Faction { get; private set; }
 
+		Building building;
+
 		public Production(ActorInitializer init, ProductionInfo info)
 		{
 			Info = info;
 			rp = Exts.Lazy(() => init.Self.IsDead ? null : init.Self.TraitOrDefault<RallyPoint>());
 			Faction = init.Contains<FactionInit>() ? init.Get<FactionInit, string>() : init.Self.Owner.Faction.InternalName;
+		}
+
+		void INotifyCreated.Created(Actor self)
+		{
+			building = self.TraitOrDefault<Building>();
 		}
 
 		public virtual void DoProduction(Actor self, ActorInfo producee, ExitInfo exitinfo, string factionVariant)
@@ -122,7 +129,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		public virtual bool Produce(Actor self, ActorInfo producee, string factionVariant)
 		{
-			if (Reservable.IsReserved(self))
+			if (Reservable.IsReserved(self) || (building != null && building.Locked))
 				return false;
 
 			// Pick a spawn/exit point pair
--- a/OpenRA.Mods.Common/Traits/Render/WithBuildingPlacedAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithBuildingPlacedAnimation.cs
@@ -22,7 +22,7 @@ namespace OpenRA.Mods.Common.Traits.Render
 		public object Create(ActorInitializer init) { return new WithBuildingPlacedAnimation(init.Self, this); }
 	}
 
-	public class WithBuildingPlacedAnimation : INotifyBuildingPlaced, INotifyBuildComplete
+	public class WithBuildingPlacedAnimation : INotifyBuildingPlaced, INotifyBuildComplete, INotifySold, INotifyTransform
 	{
 		readonly WithBuildingPlacedAnimationInfo info;
 		readonly WithSpriteBody wsb;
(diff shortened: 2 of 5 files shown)


pchote:
If people are going to kick up a fuss about the minor balance changes then option (b) is to take only the first and last commit that fix the regression and restore the behaviour we had in all versions before `release-20140608`.

MustaphaTR:
I'm OK with the balance changes this PR does.

abcdefg30:
Does this also solve #8146?

pchote:
Yes, it will.

OpenRA.Mods.Common/Traits/Production.cs
```diff
@@ -122,7 +122,8 @@ public virtual void DoProduction(Actor self, ActorInfo producee, ExitInfo exitin
 
 		public virtual bool Produce(Actor self, ActorInfo producee, string factionVariant)
 		{
-			if (Reservable.IsReserved(self))
+			var b = self.TraitOrDefault<Building>();
```
abcdefg30: Could this be cached?

OpenRA.Mods.Common/Traits/Render/WithBuildingPlacedAnimation.cs
```diff
@@ -40,6 +40,20 @@ public void BuildingComplete(Actor self)
 			buildComplete = true;
 		}
 
+		public void Sold(Actor self) { }
+		public void Selling(Actor self)
```
abcdefg30: This should use explicit implementation, I guess (minor as well though)?

obrakmann:
Seems to work alright. Please address abc's comments, then :+1:

pchote:
Done.

Mailaender:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f71c45600e838b4410b378c77aab13ab8b865b83...1948e87698e384278349d71dbbcc44afd108f9ee) :three: strike bugfix!

Mailaender:
You need to claim your bounty: https://www.bountysource.com/issues/13930430-disabled-production-when-a-factory-gets-captured @pchote

TheRaffy:
There might be a bug releated to this. look at #13983

PR #11956, opened 2016-09-03
Fix aircraft reservation madness

### LINKED ISSUE #5671: C4 on airfield/hpad won't destroy landed airplanes
Before planting C4:
airplainbug1
After:
airplainbug2

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2749567-c4-on-airfield-hpad-won-t-destroy-landed-airplanes?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #5995: Planes don't follow rally point after empty ammo
When a yak comes back to airfield on its own after empty ammo, it doesn't follow the rally point, just stays on the airfield. They do follow it after pressing f to reload.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3202005-planes-don-t-follow-rally-point-after-empty-ammo?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #7481: Aircraft attack move issues
If a Yak or a Mig is given attack move order and there is a target on the way, it attacks the target and then comes back to airfield to rearm. After rearming it continues the attack move starting from the airfield. This happens once and after that it comes back to base and just circles above construction yard and doesn't rearm or go to the attack move location.

Helicopters have a different behaviour. After they attack the target and run out of ammo, they come back to helipad, rearm and continue the attack move. If they encounter another target (or the same target), they attack it and come back to helipad for the second time. But instead of landing and rearming, they immediately skip to the attack move once they arrive at helipad. Since they carry out the attack move without ammo they come back to helipad at first enemy encounter and this repeats as a loop. 

Aircraft attack move works if a second airfield/helipad is built. After that it does this rearming and attack moving action until no target remains along the path and it arrives at attack move location.

I think the reason is that the airfield/helipad is reserved for some reason after the first rearming and the aircraft can't land. It is also not possible to give a manual landing order. When a second airfield/helipad is built it works.

This happens in latest release and playtest.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8596577-aircraft-attack-move-issues?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #7545: Was able to recharge 5 helis on 3 helipads at the same time
Here is an image, if you look closely you can see two helicopters on the far right pad and one helicopter in between the pads. 
8317451c-bce8-11e4-9ef6-b0eb8e7e88f1
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8908083-was-able-to-recharge-5-helis-on-3-helipads-at-the-same-time?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #11953: Planes will not stay on their airfields
They immediately take off again.  This must be a relatively recent regression.

obrakmann:
Fixes #11953.
Fixes #5995.
Fixes #7545.
Fixes #7481.

The important part of this change is the `MayYieldReservation` flag on `Aircraft`. It is set after an aircraft finished resupplying, and will cause it to be kicked off the airfield by the `Reservable` trait if another aircraft tries to get a reservation.

One change wrt to bleed that I couldn't find a solution for is that producing a plane will block the newly-built plane as long as the other one is still resupplying, but will cause both to take off once its done.

Even with this change, there can still (rarely) arise situations where airfields will get locked by a reservation. From the way it looked in my tests, it seemed that the aircraft that took the reservation was unable to find the proper approach to land on the airfield. I didn't yet look deeper into it, as it seemed unrelated to the issue being dealt with by this PR.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
+++ b/OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
@@ -19,6 +19,7 @@ namespace OpenRA.Mods.Common.Activities
 	public class FlyAttack : Activity
 	{
 		readonly Target target;
+		readonly Aircraft aircraft;
 		readonly AttackPlane attackPlane;
 		readonly AmmoPool[] ammoPools;
 
@@ -28,6 +29,7 @@ namespace OpenRA.Mods.Common.Activities
 		public FlyAttack(Actor self, Target target)
 		{
 			this.target = target;
+			aircraft = self.Trait<Aircraft>();
 			attackPlane = self.TraitOrDefault<AttackPlane>();
 			ammoPools = self.TraitsImplementing<AmmoPool>().ToArray();
 			ticksUntilTurn = attackPlane.AttackPlaneInfo.AttackTurnDelay;
@@ -56,6 +58,10 @@ namespace OpenRA.Mods.Common.Activities
 					inner = ActivityUtils.SequenceActivities(new FlyTimed(ticksUntilTurn, self), new Fly(self, target), new FlyTimed(ticksUntilTurn, self));
 				else
 					inner = ActivityUtils.SequenceActivities(new Fly(self, target), new FlyTimed(ticksUntilTurn, self));
+
+				// HACK: This needs to be done in this round-about way because TakeOff doesn't behave as expected when it doesn't have a NextActivity.
+				if (self.World.Map.DistanceAboveTerrain(self.CenterPosition).Length < aircraft.Info.MinAirborneAltitude)
+					inner = ActivityUtils.SequenceActivities(new TakeOff(self), inner);
 			}
 
 			inner = ActivityUtils.RunActivity(self, inner);
--- a/OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
+++ b/OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
@@ -40,10 +40,9 @@ namespace OpenRA.Mods.Common.Activities
 			if (!target.IsValidFor(self))
 				return NextActivity;
 
-			// Move to the next activity only if all ammo pools are depleted and none reload automatically
 			// TODO: This should check whether there is ammo left that is actually suitable for the target
 			if (ammoPools.All(x => !x.Info.SelfReloads && !x.HasAmmo()))
-				return ActivityUtils.SequenceActivities(new ReturnToBase(self), NextActivity);
+				return ActivityUtils.SequenceActivities(new ReturnToBase(self), this);
 
 			if (attackPlane != null)
 				attackPlane.DoAttack(self, target);
--- a/OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
+++ b/OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
@@ -19,6 +19,7 @@ namespace OpenRA.Mods.Common.Activities
 	public class ResupplyAircraft : Activity
 	{
 		readonly Aircraft aircraft;
+		Activity inner;
 
 		public ResupplyAircraft(Actor self)
 		{
@@ -27,22 +28,45 @@ namespace OpenRA.Mods.Common.Activities
 
 		public override Activity Tick(Actor self)
 		{
-			var host = aircraft.GetActorBelow();
-
-			if (host == null)
+			if (IsCanceled)
 				return NextActivity;
 
-			if (aircraft.IsPlane)
-				return ActivityUtils.SequenceActivities(
-					aircraft.GetResupplyActivities(host)
-					.Append(new CallFunc(() => aircraft.UnReserve()))
-					.Append(new WaitFor(() => NextActivity != null || Reservable.IsReserved(host)))
-					.Append(new TakeOff(self))
-					.Append(NextActivity).ToArray());
+			if (inner == null)
+			{
+				var host = aircraft.GetActorBelow();
 
-			// If is helicopter move away as soon as the resupply ends
-			return ActivityUtils.SequenceActivities(
-				aircraft.GetResupplyActivities(host).Append(new TakeOff(self)).Append(NextActivity).ToArray());
+				if (host == null)
+					return NextActivity;
+
+				if (aircraft.IsPlane)
+				{
+					inner = ActivityUtils.SequenceActivities(
+						aircraft.GetResupplyActivities(host)
+						.Append(new CallFunc(() => aircraft.MayYieldReservation = true))
+						.Append(new WaitFor(() => NextActivity != null || aircraft.ReservedActor == null))
+						.ToArray());
+				}
+				else
+				{
+					// Helicopters should take off from their helipad immediately after resupplying.
+					// HACK: NextActivity needs to be appended here because otherwise TakeOff does stupid things.
+					inner = ActivityUtils.SequenceActivities(
+						aircraft.GetResupplyActivities(host).Append(new TakeOff(self)).Append(NextActivity).ToArray());
+				}
+			}
+			else
+				inner = ActivityUtils.RunActivity(self, inner);
+
+			// The inner == NextActivity check is needed here because of the TakeOff issue mentioned in the comment above.
+			return inner == null || inner == NextActivity ? NextActivity : this;
+		}
+
+		public override void Cancel(Actor self)
+		{
+			if (!IsCanceled && inner != null)
+				inner.Cancel(self);
+
+			base.Cancel(self);
 		}
 	}
 }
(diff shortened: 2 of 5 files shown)


obrakmann:
As per discussion on IRC, added a commit that disables auto-takeoff for aircraft.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
```diff
@@ -27,22 +28,29 @@ public ResupplyAircraft(Actor self)
...
-					aircraft.GetResupplyActivities(host)
-					.Append(new CallFunc(() => aircraft.UnReserve()))
-					.Append(new WaitFor(() => NextActivity != null || Reservable.IsReserved(host)))
-					.Append(new TakeOff(self))
-					.Append(NextActivity).ToArray());
-
-			// If is helicopter move away as soon as the resupply ends
-			return ActivityUtils.SequenceActivities(
-				aircraft.GetResupplyActivities(host).Append(new TakeOff(self)).Append(NextActivity).ToArray());
+			if (inner == null)
```
pchote: IMO this would be clearer as

```csharp
if (inner != null)
{
	inner = ActivityUtils.RunActivity(self, inner);
	return this;
}
// your if case with one less level of indentation
```

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
```diff
@@ -27,22 +28,29 @@ public ResupplyAircraft(Actor self)
...
+			{
+				var host = aircraft.GetActorBelow();
+
+				if (host == null)
+					return NextActivity;
+
+				inner = aircraft.IsPlane ?
```
pchote: This feels like an abuse of the terniary statement.  Can we have an if-else?

OpenRA.Mods.Common/Traits/Buildings/Reservable.cs
```diff
@@ -29,11 +29,18 @@ public void Tick(Actor self)
 				return;		/* nothing to do */
 
 			if (!Target.FromActor(reservedFor).IsValidFor(self))
-				reservedFor = null;		/* not likely to arrive now. */
+			{
+				/* not likely to arrive now. */
+				reservedFor = null;
+				reservedForAircraft = null;
```
pchote: I wonder if not calling `reservedForAircraft.UnReserve();` here might cause your reservation issue.

pchote:
Can we please have the `Production` trait try to boot off a parked plane if there are no free spots?

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
```diff
@@ -27,22 +28,29 @@ public ResupplyAircraft(Actor self)
...
+					ActivityUtils.SequenceActivities(
+						aircraft.GetResupplyActivities(host)
```
pchote: Shifting this to an inner activity seems to have regressed the ability to cancel mid-reload and fly away to do something else.

pchote:
If you have multiple circling empty aircraft then issue the stop command they will all land on the same empty airfield to reload.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
obrakmann: I don't like that for two reasons:

1. It makes the method harder to read. You'd have the stuff that happens in subsequent ticks at the top, and the thing that runs during the first tick at the bottom.
2. I'd need two return statements.

obrakmann:
> Can we please have the Production trait try to boot off a parked plane if there are no free spots?

It already does? Unless all aircraft have just been built and never did anything else, then it'll block the last produced plane. But once they've been airborne, they'll always be in the `ResupplyAircraft` activity, which eventually sets the flag that allows the plane to get booted off.

obrakmann:
> If you have multiple circling empty aircraft then issue the stop command they will all land on the same empty airfield to reload.

Can't reproduce, works fine here. What did you do exactly?

I broke helicopter-reloading, though.

pchote:
Here are my steps:

1. Start game on ra mod, a51d9e5688fae6b06e7ebadd5e6c290749dd7fdc
2. `/all` to enable cheats.
3. Build airfield then two migs (one second apart).  Observe that first mig does not take off the make room for the second.
4. Order it to take off, then drag select both migs.  Tell them to shoot stuff until they are half ammo, then move to a location and circle.
5.  Select both migs then press `s`.  Observe them both return to base and land on the same airfield.

pchote:
> Unless all aircraft have just been built and never did anything else, then it'll block the last produced plane.

Ah, that is what i'm encountering.  If we can't fix that we should restore the rallypoint move on production.

obrakmann:
> Here are my steps:

Interesting. I get exactly the opposite result: the airfield stays locked and both migs start circling. That seems to be another bug. It again only happens to newly-produced units. If you order it to attack something right from the airfield, it will not un-reserve. If you order it to move, however, it will.

obrakmann:
Updated.
* Addressed review comments.
* Removed the commit that disabled the auto-takeoff for planes again.
* Un-broke helicopter-reloading (the activity queue would get lost on the first iteration of this PR).

pchote:
I'm not sure if it's a regression from this PR or an earlier one, but if you target something a plane will use all its ammo, reload, and then go to the rallypoint instead of continuing its attack.

pchote:
Confirmed that planes are now working properly, thanks.

Helicopters will need a similar fix.

OpenRA.Mods.Common/Activities/Air/FlyAttack.cs
```diff
@@ -56,6 +57,10 @@ public override Activity Tick(Actor self)
 					inner = ActivityUtils.SequenceActivities(new FlyTimed(ticksUntilTurn, self), new Fly(self, target), new FlyTimed(ticksUntilTurn, self));
 				else
 					inner = ActivityUtils.SequenceActivities(new Fly(self, target), new FlyTimed(ticksUntilTurn, self));
+
+				// This needs to be done in this round-about way because TakeOff doesn't behave as expected when it doesn't have a NextActivity.
```
pchote: Can we prepend a `// HACK: ` to this?  We really should fix TakeOff in a future PR.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
pchote: Ok.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
```diff
@@ -27,22 +28,45 @@ public ResupplyAircraft(Actor self)
...
+					inner = ActivityUtils.SequenceActivities(
+						aircraft.GetResupplyActivities(host)
+						.Append(new CallFunc(() => aircraft.MayYieldReservation = true))
+						.Append(new WaitFor(() => NextActivity != null || aircraft.ReservedActor == null))
+						.ToArray());
+				}
+				else
+				{
+					// Helicopters should take off from their helipad immediately after resupplying.
+					// NextActivity needs to be appended here because otherwise TakeOff does stupid things.
```
pchote: Could do with `// HACK: `s here too.

OpenRA.Mods.Common/Traits/Buildings/Reservable.cs
```diff
@@ -29,11 +29,19 @@ public void Tick(Actor self)
 				return;		/* nothing to do */
 
 			if (!Target.FromActor(reservedFor).IsValidFor(self))
-				reservedFor = null;		/* not likely to arrive now. */
+			{
+				/* not likely to arrive now. */
```
pchote: Minor style nit: `// Not likely to arrive now` please.

pchote:
All looks good to me now, great work :+1:

Mailaender:
This does work better than before in my stress tests. :white_check_mark: However I don't understand most of the code interactions.

pchote:
Just a note that i've added #5995, #7545, #7481 to the list of issues fixed in the main description.

PR #11960, opened 2016-09-03
Remove invalid cell check from Parachutable.

### LINKED ISSUE #11959: Infantry parachuted from destroyed aircraft don't die on impathable terrain
screen shot 2016-09-03 at 19 02 43

screen shot 2016-09-03 at 19 05 07

Not sure when this broke, but it can't have been *too* long ago.

pchote:
I'm not quite sure what the logic was here, maybe this was meant to be inverted?  In any case, `positionable.CanEnterCell` already accounts for the map bounds so it isn't needed here.

Fixes #11959.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Parachutable.cs
+++ b/OpenRA.Mods.Common/Traits/Parachutable.cs
@@ -64,9 +64,6 @@ namespace OpenRA.Mods.Common.Traits
 				return;
 
 			var cell = self.Location;
-			if (self.World.Map.Contains(cell))
-				return;
-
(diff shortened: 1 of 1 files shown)


Mailaender:
As the `map.Contains` check is several nested functions away I assume this was not on purpose. :+1:

abcdefg30:
👍

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/cba69446c0f42344c6d19f17c0c95a49392c2362...c16c3b4b77cfe4dae7b863bd837aafa10e0556e8)

PR #11962, opened 2016-09-03
Clean up GPS code.

### LINKED ISSUE #11955: GPS satellite does not reveal map if the player doesn't have a radar when it launches
The map will be revealed when the player later rebuilds a radar, which is very confusing if you don't know what is going on.

Tying the GPS dot state to the radar is great, but the shroud should be cleared immediately regardless of whether the radar (and the dots) are active.

pchote:
This takes a few small steps towards making the GPS code comprehensible, and then fixes #11955.

The intentional logic changes are:
- Map explore now runs at most once per player, triggered the first time that they or one of their allies' GPSWatcher becomes enabled (previously it would happen every time any player enabled or disabled their gps dots by enabling or disabling a radar or tech center)
- GpsPower &rarr; GPSWatcher updates may be delayed by a tick on capture or building, in exchange for not getting horribly out of sync when capturing while having/not having a radar dome.
- Frozen actor preview removals no longer requires you _and_ an ally to have an active GPS to run (pretty sure this was an unreported bug)
- Activating a GPS will now only refresh your allies' state, not all players.  If this code were sensible this would have no impact on the game, but this code is not sensible.  Beware while testing.

I strongly recommend reviewing commit-by-commit to follow the changes.

### DIFF
--- a/OpenRA.Mods.RA/Traits/GpsWatcher.cs
+++ b/OpenRA.Mods.RA/Traits/GpsWatcher.cs
@@ -45,33 +45,30 @@ namespace OpenRA.Mods.RA.Traits
 		public void RemoveSource(Actor source)
 		{
 			sources.Remove(source);
-			RefreshGps(source);
+			Refresh();
 		}
 
 		public void AddSource(Actor source)
 		{
 			sources.Add(source);
-			RefreshGps(source);
+			Refresh();
 		}
 
-		public void Launch(Actor atek, GpsPowerInfo info)
+		public void Launch(Actor source, GpsPowerInfo info)
 		{
-			atek.World.Add(new DelayedAction(info.RevealDelay * 25, () =>
+			source.World.Add(new DelayedAction(info.RevealDelay * 25, () =>
 			{
 				Launched = true;
-				RefreshGps(atek);
+				Refresh();
 			}));
 		}
 
-		public void RefreshGps(Actor atek)
+		void Refresh()
 		{
 			RefreshGranted();
 
-			foreach (var i in atek.World.ActorsWithTrait<GpsWatcher>())
+			foreach (var i in owner.World.ActorsWithTrait<GpsWatcher>())
 				i.Trait.RefreshGranted();
-
-			if ((Granted || GrantedAllies) && atek.Owner.IsAlliedWith(owner))
-				atek.Owner.Shroud.ExploreAll();
 		}
 
 		void RefreshGranted()
--- a/OpenRA.Mods.RA/Traits/GpsWatcher.cs
+++ b/OpenRA.Mods.RA/Traits/GpsWatcher.cs
@@ -28,16 +28,22 @@ namespace OpenRA.Mods.RA.Traits
 
 	class GpsWatcher : ISync, IFogVisibilityModifier
 	{
-		[Sync] bool enabled;
-		[Sync] bool grantedAllies;
-		[Sync] bool granted;
-
-		public bool Active { get { return granted || grantedAllies; } }
+		public bool Active { get { return active || allyActive; } }
 
 		readonly Player owner;
 
 		readonly List<Actor> sources = new List<Actor>();
 		readonly HashSet<TraitPair<IOnGpsRefreshed>> notifyOnRefresh = new HashSet<TraitPair<IOnGpsRefreshed>>();
+
+		// Whether this watcher has been enabled (by launching a satellite)
+		[Sync] bool enabled;
+
+		// Whether this watcher has explored the terrain (by becoming enabled, or an ally becoming enabled)
+		[Sync] bool explored;
+
+		// Whether this or an ally's watcher is currently active (> 0 sources and enabled)
+		[Sync] bool active;
+		[Sync] bool allyActive;
 
 		public GpsWatcher(Player owner)
 		{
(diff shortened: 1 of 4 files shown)


OpenRA.Mods.RA/Traits/GpsWatcher.cs
```diff
@@ -34,34 +34,33 @@ class GpsWatcher : ISync, IFogVisibilityModifier
...
 
 		public void Launch(Actor atek, GpsPowerInfo info)
 		{
-			atek.World.Add(new DelayedAction(info.RevealDelay * 25,
-				() =>
-				{
-					Launched = true;
-					RefreshGps(atek);
-				}));
+			atek.World.Add(new DelayedAction(info.RevealDelay * 25, () =>
```
abcdefg30: Shouldn't that adapt to game speed?

pchote: I wanted to limit the scope of this pr to things that didn't require upgrade rules.

OpenRA.Mods.RA/Traits/SupportPowers/GpsPower.cs
```diff
@@ -48,19 +48,19 @@ class GpsPowerInfo : SupportPowerInfo
...
+		GpsWatcher watcher;
+		bool wasDisabled;
 
 		public GpsPower(Actor self, GpsPowerInfo info)
 			: base(self, info)
 		{
 			this.self = self;
 			this.info = info;
-			owner = self.Owner.PlayerActor.Trait<GpsWatcher>();
-			owner.GpsAdd(self);
```
pchote: `wasDisabled` needs to start `true` to compensate for this change.

OpenRA.Mods.RA/Traits/GpsWatcher.cs
```diff
@@ -77,12 +83,18 @@ void Refresh()
 		void RefreshInner()
 		{
 			var wasActive = Active;
+			var allyWatchers = owner.World.ActorsWithTrait<GpsWatcher>().Where(kv => kv.Actor.Owner.IsAlliedWith(owner));
```
abcdefg30: `GpsWatcher` is used on the player actors, so we should be able to cache this, shouldn't we?

pchote: now that we've ditched fragile alliances, yes, but again I wanted to minimize the changes for the sake of getting this this immediately before the playtest.  This is the least of the remaining problems in the GPS code.

OpenRA.Mods.RA/Traits/SupportPowers/GpsPower.cs
pchote: Fixed.

OpenRA.Mods.RA/Traits/GpsWatcher.cs
```diff
@@ -28,72 +28,82 @@ interface IOnGpsRefreshed { void OnGpsRefresh(Actor self, Player player); }
...
+			var allyEnabled = allyWatchers.Any(w => w.Trait.enabled);
+			if ((enabled || allyEnabled) && !explored)
+			{
+				explored = true;
 				owner.Shroud.ExploreAll();
+			}
 
-			if (wasGranted != Granted || wasGrantedAllies != GrantedAllies)
+			if (wasActive != Active)
 				foreach (var tp in notifyOnRefresh.ToList())
```
penev92: This .ToList() is really bugging me.

OpenRA.Mods.RA/Traits/GpsWatcher.cs
```diff
@@ -28,72 +28,82 @@ interface IOnGpsRefreshed { void OnGpsRefresh(Actor self, Player player); }
...
 		{
-			RefreshGranted();
-
-			foreach (var i in atek.World.ActorsWithTrait<GpsWatcher>())
-				i.Trait.RefreshGranted();
+			RefreshInner();
 
-			if ((Granted || GrantedAllies) && atek.Owner.IsAlliedWith(owner))
-				atek.Owner.Shroud.ExploreAll();
+			// Refresh the state of all allied players (including ourselves, again...)
```
penev92: Why refresh above if the foreach is going to refresh everyone?

pchote: Because the other updates depend on the state of this one being updated first.

OpenRA.Mods.RA/Traits/GpsWatcher.cs
pchote: Again, the GPS code is horrifically fragile so I really don't want to touch more than I have to in this release-blocking PR.

OpenRA.Mods.RA/Traits/SupportPowers/GpsPower.cs
```diff
@@ -84,33 +85,35 @@ public override void Activate(Actor self, Order order, SupportPowerManager manag
...
+			// Note: watcher registration is already handled by removal/addition from world
 			watcher = newOwner.PlayerActor.Trait<GpsWatcher>();
-			watcher.AddSource(self);
 		}
 
-		bool NoActiveRadar { get { return !self.World.ActorsHavingTrait<ProvidesRadar>(r => r.IsActive).Any(a => a.Owner == self.Owner); } }
-		bool wasDisabled;
+		void INotifyAddedToWorld.AddedToWorld(Actor self)
+		{
+			// Registration will happen in the next tick is needed
```
obrakmann: _if_ needed?

OpenRA.Mods.RA/Traits/GpsWatcher.cs
```diff
@@ -58,38 +60,38 @@ public void Launch(Actor source, GpsPowerInfo info)
...
+		void RefreshInner()
 		{
-			var wasGranted = Granted;
-			var wasGrantedAllies = GrantedAllies;
+			var wasActive = Active;
 
-			Granted = sources.Count > 0 && Launched;
-			GrantedAllies = owner.World.ActorsHavingTrait<GpsWatcher>(g => g.Granted).Any(p => p.Owner.IsAlliedWith(owner));
+			granted = sources.Count > 0 && enabled;
+			grantedAllies = owner.World.ActorsHavingTrait<GpsWatcher>(g => g.granted).Any(p => p.Owner.IsAlliedWith(owner));
```
obrakmann: Should this be set when only you have GPS?

pchote: This doesn't affect anything in practice, but should indeed be fixed as part of a larger cleanup.

OpenRA.Mods.RA/Traits/SupportPowers/GpsPower.cs
pchote: Fixed.

obrakmann:
Something's not quite right here. I just watched an AI game as a spectator where one AI had a tech center, the satellite was deployed (didn't see it myself, but the countdown was gone), but there were no GPS dots. Not sure what happened there. The other AIs' GPS was working fine. Watching the replay now.

pchote:
Remember that the GPS dots now require a working radar / sufficient power.

pchote:
As discussed on IRC even this small amount of sanity refactoring is too risky.  Lets go with #11968 instead.

PR #11964, opened 2016-09-03
fix crash in missions because cameras need .destroy, not .kill to be removed

### LINKED ISSUE #11963: nod07a: crash because script tries to kill camera
when reaching the bridge on the left side of the map, the Nod base is discovered
this triggers a flare at the Nod base, and creates a camera which shows the base

when reaching the Nod base from the left (not from the south!), reinforcements are triggered
this triggers a timer which kills the flare and the camera after a delay (2 minutes)
when the timer runs out, the mission crashes since cameras can't be killed:

```
Fatal Lua Error: [string "BindingSupport.lua"]:30: Actor 'camera' does not define a property 'Kill'
```
to reproduce more easily: 
in line 217 in nod07a.lua (Trigger.AfterDelay(DateTime.Minutes(2), function()) change "Minutes(2)" to "Seconds(2)", then just send the light tank to the area to the left of the Nod base to trigger the reinforcements

clemty:
fixes #11963


mods/cnc/maps/nod07a/nod07a.lua
```diff
@@ -216,7 +216,6 @@ Trigger.OnEnteredFootprint(ReinforcementsTrigger, function(a, id)
 
 		Trigger.AfterDelay(DateTime.Minutes(2), function()
 			Flare.Destroy()
-			FlareCamera.Kill()
```
pchote: Why not `.Destroy()` on this?

pchote:
We should check the other missions too.  This may not be the only use of this now-invalid pattern.

clemty:
I ran grep for "kill" and all other instances seem to point to variable names where the killed objects are actual units/buildings

obrakmann:
Looks good to me. These seem to be all instances:
```
$ rgrep -Eil 'camera.*Kill' mods/*/maps
mods/cnc/maps/nod08b/nod08b.lua
mods/cnc/maps/nod07a/nod07a.lua
mods/cnc/maps/nod08a/nod08a.lua
mods/ra/maps/soviet-02b/soviet02b.lua
mods/ra/maps/soviet-06a/soviet06a.lua
```

:+1:

abcdefg30:
👍

PR #11973, opened 2016-09-05
Fix AfterBuild tasks in csprojs

phrohdoh:
These should now work on every OS.

cc @alercah @abcdefg30

I would really like to keep the debug comments in for future improvements by myself or others.

### DIFF
--- a/OpenRA.Game/OpenRA.Game.csproj
+++ b/OpenRA.Game/OpenRA.Game.csproj
@@ -36,7 +36,7 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x86'">
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
-    <OutputPath>..\</OutputPath>
+    <OutputPath>bin\Debug</OutputPath>
     <DefineConstants>TRACE;DEBUG</DefineConstants>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <CodeAnalysisRuleSet>AllRules.ruleset</CodeAnalysisRuleSet>
@@ -45,7 +45,7 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|x86'">
     <DebugSymbols>true</DebugSymbols>
-    <OutputPath>..\</OutputPath>
+    <OutputPath>bin\Release</OutputPath>
     <DefineConstants>TRACE</DefineConstants>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <DebugType>pdbonly</DebugType>
--- a/OpenRA.Game/OpenRA.Game.csproj
+++ b/OpenRA.Game/OpenRA.Game.csproj
@@ -338,12 +338,28 @@
     <Content Include="OpenRA.ico" />
   </ItemGroup>
   <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
-  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
-       Other similar extension points exist, see Microsoft.Common.targets.
-  <Target Name="BeforeBuild">
-  </Target>
   <Target Name="AfterBuild">
+    <Copy
+      SourceFiles="$(TargetPath)"
+      DestinationFolder="$(SolutionDir)"/>
+    <!--
+        We need to copy OpenRA.Game.pdb (not `.dll.pdb`). This is only necessary on Windows.
+        Mono outputs a `.dll.mdb` so we can just append `.mdb` directly.
+    -->
+    <Copy
+      SourceFiles="$(TargetDir)$(TargetName).pdb"
+      DestinationFolder="$(SolutionDir)"
+      Condition="'$(OS)' == 'Windows_NT'"/>
+    <Copy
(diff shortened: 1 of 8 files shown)


Phrohdoh:
This is broken on Windows because `OpenRA.Game.pdb` doesn't exist.
Can someone on Windows tell me if this file should instead be `OpenRA.Game.exe.pdb`, or if it does not exist in any fashion?

RoosterDragon:
`GetFilenameWithoutExtension` removes the full path and returns literally just the filename. This means it's looking for the relative path `OpenRA.Game.pdb` and god knows what the current directory is.

Try `SourceFiles="$(TargetDir)$(TargetName).pdb"`.

RoosterDragon:
Also I might note this is not visible in VS's UI, short of loading the actual .csproj file. I'm not particularly against that, but thought it might be worth pointing out.

abcdefg30:
You also want to add the `.pdb` files `.gitignore`.

Phrohdoh:
I have addressed both @RoosterDragon and @abcdefg30's comments (sans visibility in VS, which I cannot change, nor do I believe will be an issue for us).

abcdefg30:
Seems to work fine now on Windows. ✅

abcdefg30:
You may want to remove the out-commented lines for debugging before merge.

Phrohdoh:
>I would really like to keep the debug comments in for future improvements by myself or others.

OpenRA.Game/OpenRA.Game.csproj
```diff
@@ -338,12 +338,29 @@
     <Content Include="OpenRA.ico" />
   </ItemGroup>
   <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
-  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
-       Other similar extension points exist, see Microsoft.Common.targets.
-  <Target Name="BeforeBuild">
-  </Target>
   <Target Name="AfterBuild">
+    <MakeDir Directories="$(SolutionDir)"/>
```
Phrohdoh: This line is unnecessary. I will remove it.

abcdefg30:
`make clean` also needs to remove the `.pdb` files.

Phrohdoh:
Updated: `pdb`s will now be cleaned up.

pchote:
My one concern here is that this may break older versions of monodevelop and/or xbuild, which had very spotty support for the project files.  If we can confirm that this still works fine with mono 3.2 and whatever version of md/xs comes with debian etc, as well as the latest versions, then i'm all for merging this.

pchote:
```
paul@Mjolnir:OpenRA+pr/11973$ mono --version
Mono JIT compiler version 3.2.0 ((no/7c7fcc7 Tue Jul 23 19:59:39 EDT 2013)
...
paul@Mjolnir:OpenRA+pr/11973$ xbuild
...
			Target DeployOutputFiles:
				Copying file from '/Users/paul/src/OpenRA/OpenRA.Mods.Common/obj/x86/Debug/OpenRA.Mods.Common.dll.mdb' to '/Users/paul/src/OpenRA/OpenRA.Mods.Common/bin/Debug/OpenRA.Mods.Common.dll.mdb'
				Copying file from '/Users/paul/src/OpenRA/OpenRA.Mods.Common/obj/x86/Debug/OpenRA.Mods.Common.dll' to '/Users/paul/src/OpenRA/OpenRA.Mods.Common/bin/Debug/OpenRA.Mods.Common.dll'
			Target AfterBuild:
				Copying file from '/Users/paul/src/OpenRA/OpenRA.Mods.Common/bin/Debug/OpenRA.Mods.Common.dll' to '/Users/paul/src/OpenRA/mods/common/OpenRA.Mods.Common.dll'
				Copying file from '/Users/paul/src/OpenRA/OpenRA.Mods.Common/bin/Debug/OpenRA.Mods.Common.dll.mdb' to '/Users/paul/src/OpenRA/mods/common/OpenRA.Mods.Common.dll.mdb'
		Done building project "/Users/paul/src/OpenRA/OpenRA.Mods.Common/OpenRA.Mods.Common.csproj".
...
```
This is good enough for me, considering the alternative is to let this rot in the PR queue.  :+1:

OpenRA.Game/OpenRA.Game.csproj
```diff
@@ -45,7 +45,7 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|x86'">
     <DebugSymbols>true</DebugSymbols>
-    <OutputPath>..\</OutputPath>
+    <OutputPath>bin\Release</OutputPath>
```
RoosterDragon: Was changing these paths intentional? This breaks at runtime since make no longer dumps dependencies into the right places.

Phrohdoh: It was intentional.
Sorry, I do not have time currently to fix the `make.ps1` script.

pchote: So much for @abcdefg30's ✅ :grin:

abcdefg30: This works just fine for me with both the makefile and my IDE (Sharp Develop).

pchote: http://logs.openra.net/index.php?year=2016&month=10&day=13#18:24:40

Issue #9613, opened 2015-10-11
Merge Rearm and Repair activities into a generalistic Resupply activity


reaperrr:
Multiple activities cannot run in parallel, so in order to support replenishing health and ammo at the same time, this is a necessary step.

reaperrr:
As #11982 and some other issues have shown, we should probably give this some priority to get it into Next+1.

PR #11982, opened 2016-09-05
Fix helicopter reservations

### LINKED ISSUE #11979: Hinds stack onto of another when rearmed via shortcut [playtest-20160904]
Reported by Frame_Limiter via IRC:
If you rearm your hinds using the shortcut key "F", the hinds will stack on top of one another and remain on the helipad.

obrakmann:
Fixes #11979

### DIFF
--- a/OpenRA.Mods.Common/Activities/Air/AllowYieldingReservation.cs
+++ b/OpenRA.Mods.Common/Activities/Air/AllowYieldingReservation.cs
@@ -1,0 +1,32 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using OpenRA.Activities;
+using OpenRA.Mods.Common.Traits;
+
+namespace OpenRA.Mods.Common.Activities
+{
+	public class AllowYieldingReservation : Activity
+	{
+		readonly Aircraft aircraft;
+
+		public AllowYieldingReservation(Actor self)
+		{
+			aircraft = self.Trait<Aircraft>();
+		}
+
+		public override Activity Tick(Actor self)
+		{
+			aircraft.AllowYieldingReservation();
+			return NextActivity;
+		}
+	}
+}
--- a/OpenRA.Mods.Common/Activities/Air/HeliAttack.cs
+++ b/OpenRA.Mods.Common/Activities/Air/HeliAttack.cs
@@ -60,9 +60,9 @@ namespace OpenRA.Mods.Common.Activities
 			{
 				var newTarget = Target.FromCell(self.World, self.World.Map.CellContaining(target.CenterPosition));
 
-				self.CancelActivity();
+				Cancel(self);
 				self.SetTargetLine(newTarget, Color.Green);
-				return ActivityUtils.SequenceActivities(new HeliFly(self, newTarget));
+				return new HeliFly(self, newTarget);
 			}
 
 			// If any AmmoPool is depleted and no weapon is valid against target, return to helipad to reload and then resume the activity
--- a/OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
+++ b/OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
@@ -42,7 +42,7 @@ namespace OpenRA.Mods.Common.Activities
 				{
 					inner = ActivityUtils.SequenceActivities(
 						aircraft.GetResupplyActivities(host)
-						.Append(new CallFunc(() => aircraft.MayYieldReservation = true))
+						.Append(new AllowYieldingReservation(self))
 						.Append(new WaitFor(() => NextActivity != null || aircraft.ReservedActor == null))
 						.ToArray());
 				}
@@ -52,7 +52,7 @@ namespace OpenRA.Mods.Common.Activities
 					// HACK: Append NextActivity to TakeOff to avoid moving to the Rallypoint (if NextActivity is non-null).
 					inner = ActivityUtils.SequenceActivities(
 						aircraft.GetResupplyActivities(host)
-						.Append(new CallFunc(() => aircraft.MayYieldReservation = true))
+						.Append(new AllowYieldingReservation(self))
 						.Append(new TakeOff(self)).Append(NextActivity).ToArray());
 				}
 			}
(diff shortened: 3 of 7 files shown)


obrakmann:
Added a fix for another bug that would cause an infinite loop when an airplane goes out of ammo and the player has no airfields anymore.

abcdefg30:

obrakmann:
Fixed another instance of an inner activity clobbering the queue. Haven't had time to test if this fixes all the issues yet, though.

What is definitely not fixed is that the target lines keep pointing to targeted units beneath the shroud. But that's happening even in the old release, so isn't a regressions and opens a whole different can of worms, so I'd like to keep that out of here.

obrakmann:
OK, I think this is ready for review again.

@abcdefg30, I'm removing your +1 because of the newly added commit. The earlier commits you already reviewed didn't change.

pchote:
screen shot 2016-09-14 at 22 29 05

I ended up with two hinds on the same pad after attack-moving a blob of enemy infantry who snuck into my base while testing aircraft...

pchote:
Happens with planes too:

screen shot 2016-09-14 at 22 35 02

pchote (changes requested):
Fixes the stacking issue mentioned in the original ticket, but testing shows additional stacking issues with both planes and helos.

If it isn't too complicated then we should aim to fix these before merging – otherwise, I think this is fine as long as our final behaviour is better than what we have on the current release.

pchote:
Heli stacking at 6:39: ﻿https://www.dropbox.com/s/kou5u233tdimr5c/ra-helo-stack.orarep?dl=0
Plane stacking at 1:36 and 2:29: https://www.dropbox.com/s/59opu0ea47w65od/ra-plane-stack.orarep?dl=0

obrakmann:
Found the problem for the aircraft one: it only happens for actors whose reserved building gets sold. `Reservable.Selling` calls `Aircraft.UnReserve`, which sets the `MayYieldReservation` flag to false, but then the queued `CallFunc` from `ResupplyAircraft` runs, which sets it to true again, where it stays and causes misbehaviour the next time the aircraft lands.

The helicopter case seems to be different as you didn't see the helipad there, unless I missed it during the fast-forward.

pchote:
We can guard against any other stacking issues by having the aircraft check their reservations on landing, and immediately taking off again if someone else holds it.

obrakmann:
I'd prefer to not do that. All the stacking issues so far have been legitimate bugs. I'd rather find and fix them than adding a workaround that would hide them.

pchote:
That's a good point that I frequently use myself, so ok.

pchote:
I think it's out of scope for this PR (best to have one of its own), but the new reservation system is a little wonky in TD, where helis repair but don't reload: using the return to base key should make the helis return to the vicinity of the helipad, but only those that actually need repairs should land.

pchote:
Found another issue, unfortunately: If you attack-move a group of empty aircraft to a location with enemies they will fly there and circle instead of automatically reloading.  If you attack-move a group of loaded aircraft then they will empty, reload, and return as expected.

obrakmann:
That has nothing to do with these changes, and happened the same way before these changes were merged.  The problem seems to be that `AttackPlane.CanAttack()` (or even `AttackBase.CanAttack()` doesn't check the state of the ammo pools. Either way, I'd argue it's out of scope, just as another `AttackMove`-related issue, namely that aircraft stop attack-moving once they reach the destination.

pchote:
Huh... I'm pretty sure that it used to work in earlier versions of the PR, but I may be mistaken.

pchote (approved):
Couldn't find any other new issues, so lets get this in.

reaperrr:
If I give a reloading aircraft a stop order, it stops reloading. This can be quite annoying if I just want to prevent it from continuing to attack a target after reloading.

Not sure if that is fixable without causing other side-effects elsewhere, though.

abcdefg30:
Isn't that wanted behaviour? (To e.g. have helicopters stay at the ground for repairs.)

pchote:
That's what repair depots are for.  A bug having potentially useful side-effects doesn't make it any less of a bug.

abcdefg30:
> That's what repair depots are for. A bug having potentially useful side-effects doesn't make it any less of a bug.

We explicitly allow mechanics to repair helicopters when they're landed, so I wouldn't call that "side-effects". (Same goes for planes, btw.)

pchote:
That came from a world where the helicopter-helipad interaction was hopelessly broken.  Using it as an excuse to not fix things properly is backwards.  Remember that RA isn't the only mod with aircraft.

reaperrr:
> Isn't that wanted behaviour? (To e.g. have helicopters stay at the ground for repairs.)

In my opinion the wanted behaviour should be that the aircraft continues reloading as long as it is on the helipad/airfield, and only other queued activities like taking off afterwards should be canceled. 
Having to manually order it to take off and land again to resume reloading just because you pressed 'S' to prevent it from doing something else is a bug to me.

obrakmann:
I guess this would be possible (going from memory of the code here), but rather ugly. The `Cancel()` method in the `ResupplyAircraft` activity would need to be overridden so that it doesn't cancel itself, but just sets `NextActivity` to null, and also sets the `NextActivity` of the inner activity that is the last returned from `GetResupplyActivities()`  (for aircraft) or the Takeoff activity (for helis) to null.

I think I'd be opposed to doing that for the ugly code involved alone.

A better solution might be to make the `ReturnToBase` order check whether we're already on the ground and then just queueing only the resupply activities (+ Takeoff for helis). That difference in behaviour is a major pain in the rear, btw.

obrakmann:
Actually, #9613 would be my favourite solution. Not going to happen for this release, though. I wouldn't say this issue is important enough to hold up the release over it anyway.

pchote:
I agree on all points.

reaperrr:
OK, I agree it's not a showstopper. Looks good to me otherwise, :+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/58b1d17c2db03ada148a6875875e6708a63fd641...55ff40ce45123c9f6d3d5e2897f394bf41ae1820)

e1d1s1:
Hinds are still stacking. Screenshots attached. Issued a F command and it returned to the first helipad that already had a hind on it.
openra-2016-09-24t202642563z
openra-2016-09-24t202642627z
openra-2016-09-24t202643026z

Also, mechanics cannot fully repair hinds when they are on the pad, since they do not IDLE on the pad. They are only IDLING as long as it takes to rearm. The slow-rotor animation should continue to play while helos are on the ground/pad. They should not be full throttle on the ground.

Helos should remain IDLE until given another order after rearming, unless the player has explicitly ordered a force attack/attack move, which already was the behavior.

Helos haven't chased a target after a reload in any version of command and conquer. This is not StarCraft where your flying units do not need to rearm. Now they are suicidal and nearly un-repairable. This entire set of changes is a regression.

pchote:
#12102.

PR #11984, opened 2016-09-05
Make Tooltip upgradable

reaperrr:
This allows the tooltip to change via upgrades, for example when an actor is deployed or reaches elite status.

### DIFF
--- a/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
+++ b/OpenRA.Game/Traits/Player/FrozenActorLayer.cs
@@ -40,7 +40,7 @@ namespace OpenRA.Traits
 
 		public ITooltipInfo TooltipInfo { get; private set; }
 		public Player TooltipOwner { get; private set; }
-		readonly ITooltip tooltip;
+		readonly ITooltip[] tooltips;
 
 		public int HP { get; private set; }
 		public DamageState DamageState { get; private set; }
@@ -80,7 +80,7 @@ namespace OpenRA.Traits
 			Bounds = self.Bounds;
 			TargetTypes = self.GetEnabledTargetTypes().ToHashSet();
 
-			tooltip = self.TraitsImplementing<ITooltip>().FirstOrDefault();
+			tooltips = self.TraitsImplementing<ITooltip>().ToArray();
 			health = self.TraitOrDefault<IHealth>();
 
 			UpdateVisibility();
@@ -101,6 +101,7 @@ namespace OpenRA.Traits
 				DamageState = health.DamageState;
 			}
 
+			var tooltip = tooltips.FirstOrDefault(Exts.IsTraitEnabled);
 			if (tooltip != null)
 			{
 				TooltipInfo = tooltip.TooltipInfo;
--- a/OpenRA.Mods.Common/Traits/Tooltip.cs
+++ b/OpenRA.Mods.Common/Traits/Tooltip.cs
@@ -13,11 +13,9 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits
 {
-	public abstract class TooltipInfoBase : ITraitInfo
+	public abstract class TooltipInfoBase : UpgradableTraitInfo
 	{
 		[Translate] public readonly string Name = "";
-
-		public abstract object Create(ActorInitializer init);
 	}
 
 	[Desc("Shown in map editor.")]
@@ -61,7 +59,7 @@ namespace OpenRA.Mods.Common.Traits
 		public bool IsOwnerRowVisible { get { return ShowOwnerRow; } }
 	}
 
-	public class Tooltip : ITooltip
+	public class Tooltip : UpgradableTrait<TooltipInfo>, ITooltip
 	{
 		readonly Actor self;
 		readonly TooltipInfo info;
@@ -70,6 +68,7 @@ namespace OpenRA.Mods.Common.Traits
 		public Player Owner { get { return self.Owner; } }
 
 		public Tooltip(Actor self, TooltipInfo info)
+			: base(info)
 		{
 			this.self = self;
 			this.info = info;
--- a/OpenRA.Mods.Common/Traits/World/EditorActorPreview.cs
+++ b/OpenRA.Mods.Common/Traits/World/EditorActorPreview.cs
@@ -70,7 +70,9 @@ namespace OpenRA.Mods.Common.Traits
 				Footprint = new ReadOnlyDictionary<CPos, SubCell>(footprint);
 			}
 
-			var tooltip = Info.TraitInfoOrDefault<EditorOnlyTooltipInfo>() as TooltipInfoBase ?? Info.TraitInfoOrDefault<TooltipInfo>();
+			var tooltip = Info.TraitInfos<EditorOnlyTooltipInfo>().FirstOrDefault(Exts.IsTraitEnabled) as TooltipInfoBase
+				?? Info.TraitInfos<TooltipInfo>().FirstOrDefault(Exts.IsTraitEnabled);
+
 			Tooltip = (tooltip == null ? " < " + Info.Name + " >" : tooltip.Name) + "\n" + owner.Name + " (" + owner.Faction + ")"
 				+ "\nID: " + ID + "\nType: " + Info.Name;
 
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/ProductionTooltipLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/ProductionTooltipLogic.cs
@@ -115,7 +115,7 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 		static string ActorName(Ruleset rules, string a)
 		{
 			ActorInfo ai;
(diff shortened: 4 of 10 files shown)


OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -80,7 +80,7 @@ public FrozenActor(Actor self, PPos[] footprint, Shroud shroud, bool startsRevea
 			Bounds = self.Bounds;
 			TargetTypes = self.GetEnabledTargetTypes().ToHashSet();
 
-			tooltip = self.TraitsImplementing<ITooltip>().FirstOrDefault();
+			tooltip = self.TraitsImplementing<ITooltip>().FirstOrDefault(Exts.IsTraitEnabled);
```
pchote: You'll need to store all of these here, and then pick the appropriate one in `RefreshState`.

mods/ts/rules/nod-vehicles.yaml
```diff
@@ -73,6 +73,13 @@ TTNK:
 	Tooltip:
 		Name: Tick Tank
 		Description: Nod's main battle tank.\nCan deploy to gain extra protection.\n  Strong vs Vehicles\n  Weak vs Infantry, Aircraft
+		UpgradeTypes: deployed
+		UpgradeMaxEnabledLevel: 0
+	Tooltip@DEPLOYED:
+		Name: Deployed Tick Tank
```
pchote: "Tick Tank (deployed)" please

MustaphaTR: Deployed Tick Tank is what vanilla TS uses. But i think Tick Tank (deployed)  is better too.

reaperrr:
Updated.

mods/ts/rules/nod-vehicles.yaml
```diff
@@ -73,6 +73,13 @@ TTNK:
 	Tooltip:
 		Name: Tick Tank
 		Description: Nod's main battle tank.\nCan deploy to gain extra protection.\n  Strong vs Vehicles\n  Weak vs Infantry, Aircraft
+		UpgradeTypes: deployed
+		UpgradeMaxEnabledLevel: 0
+	Tooltip@DEPLOYED:
+		Name: Tick Tank (deployed)
+		Description: Nod's main battle tank.\nDeployed to gain extra protection.\n  Strong vs Vehicles\n  Weak vs Infantry, Aircraft
```
pchote: This highlights the bogosity of having this on `Tooltip` instead of `Buildable`.  These alternate descriptions will never be used.  I won't ask you to do it here (it belongs in its own PR), but we really need to move this to the proper trait.

reaperrr: Filed #12045 so we don't forget.

reaperrr:
I think it's better to get #12141 merged first.

reaperrr:
Rebased.

OpenRA.Mods.Common/Widgets/ObserverProductionIconsWidget.cs
```diff
@@ -91,7 +91,7 @@ public override void Draw()
 
 				var rsi = actor.TraitInfo<RenderSpritesInfo>();
 				var icon = new Animation(world, rsi.GetImage(actor, world.Map.Rules.Sequences, faction));
-				icon.Play(actor.TraitInfo<TooltipInfo>().Icon);
+				icon.Play(actor.TraitInfos<TooltipInfo>().First(Exts.IsTraitEnabled).Icon);
```
reaperrr: I was just wondering, shouldn't we move `Icon` to `Buildable` as well?

reaperrr:
Filed #12258 which should be merged first, too.

OpenRA.Mods.Common/Traits/World/EditorActorPreview.cs
```diff
@@ -70,7 +70,9 @@ public EditorActorPreview(WorldRenderer worldRenderer, string id, ActorReference
 				Footprint = new ReadOnlyDictionary<CPos, SubCell>(footprint);
 			}
 
-			var tooltip = Info.TraitInfoOrDefault<EditorOnlyTooltipInfo>() as TooltipInfoBase ?? Info.TraitInfoOrDefault<TooltipInfo>();
+			var tooltip = Info.TraitInfos<EditorOnlyTooltipInfo>().FirstOrDefault() as TooltipInfoBase
+				?? Info.TraitInfos<TooltipInfo>().FirstOrDefault();
```
reaperrr: Note to self: This should probably be `.First(Exts.IsTraitEnabled)` too, in any case it should match `ActorSelectorLogic`.

abcdefg30:
The dependencies are merged now.

reaperrr:
Rebased and updated.

OpenRA.Game/Traits/Player/FrozenActorLayer.cs
```diff
@@ -101,8 +101,9 @@ public void RefreshState()
 				DamageState = health.DamageState;
 			}
 
-			if (tooltip != null)
+			if (tooltips.Any(Exts.IsTraitEnabled))
```
pchote: You can make this cleaner by pulling a `var tooltip = tooltips.FirstOrDefault(Exts.IsTraitEnabled)` outside the conditional and then keeping this as a null check.

OpenRA.Mods.Common/Traits/Tooltip.cs
```diff
@@ -13,11 +13,11 @@
 
 namespace OpenRA.Mods.Common.Traits
 {
-	public abstract class TooltipInfoBase : ITraitInfo
+	public abstract class TooltipInfoBase : UpgradableTraitInfo
 	{
 		[Translate] public readonly string Name = "";
 
-		public abstract object Create(ActorInitializer init);
+		public override object Create(ActorInitializer init) { return this; }
```
pchote: Returning `this` here is bogus (that is a traitinfo, not a trait).  You should be able to remove this line completely.

OpenRA.Mods.Common/Widgets/Logic/Ingame/ProductionTooltipLogic.cs
```diff
@@ -115,7 +115,7 @@ static string ActorName(Ruleset rules, string a)
 		{
 			ActorInfo ai;
 			if (rules.Actors.TryGetValue(a.ToLowerInvariant(), out ai) && ai.HasTraitInfo<TooltipInfo>())
-				return ai.TraitInfo<TooltipInfo>().Name;
+				return ai.TraitInfos<TooltipInfo>().First(Exts.IsTraitEnabled).Name;
```
pchote: It looks like this method is expected to return `a` if there isn't a valid tooltip, so better to use `FirstOrDefault` and only return if non-null here.

OpenRA.Mods.Common/Widgets/ViewportControllerWidget.cs
```diff
@@ -191,7 +191,7 @@ public void UpdateMouseover()
 
 			if (underCursor != null)
 			{
-				ActorTooltip = underCursor.TraitsImplementing<ITooltip>().First();
+				ActorTooltip = underCursor.TraitsImplementing<ITooltip>().First(Exts.IsTraitEnabled);
```
pchote: Crashing here if all tooltips are disabled isn't great (previously the check in `underCursor` would have prevented this.  It would be better to use `FirstOrDefault` and then only set `ActorTooltipExtra` / `TooltipType` if it is non-null.

pchote (changes requested):

reaperrr:
Updated accordingly.

In addition, I changed `Buildable` actors without `Tooltip` to not crash the game, but display the internal name on the build palette instead and throw a lint error.
I believe this is a bit more modder-friendly for mods in their early stages, as actors won't require `Tooltip` to be buildable.

GraionDilach:
Please add an upgraded tooltip to the Sensor Array as well.

reaperrr:
> Please add an upgraded tooltip to the Sensor Array as well.

Updated.

OpenRA.Mods.Common/Widgets/Logic/Ingame/ProductionTooltipLogic.cs
```diff
@@ -115,7 +115,11 @@ static string ActorName(Ruleset rules, string a)
 		{
 			ActorInfo ai;
 			if (rules.Actors.TryGetValue(a.ToLowerInvariant(), out ai) && ai.HasTraitInfo<TooltipInfo>())
```
abcdefg30: Can't we drop the `HasTraitInfo` check now?

abcdefg30:

OpenRA.Mods.Common/Widgets/Logic/Ingame/ProductionTooltipLogic.cs
reaperrr: Done.

reaperrr:
Updated.

obrakmann (approved):
👍

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/12729333016d5b6ef42830b9a8e135ce262fe830)

PR #11987, opened 2016-09-06
Expand Proximity/CellTriggers with an onOwnerChanged method.

### LINKED ISSUE #11707: ChangeOwnerWarhead + UpgradeActorsNear crash
Basically the game crashes if you use a ChangeOnwerWarhead on any actor that is under the effect of many UpgradeActorsNear auras. I've only tested this with a ChangeOnwerWarhead that has a limited time duration.

As a little explanation for the situation, the actor that provided the upgraded gave the mind control victim various upgrades, firepower, speed, and so on, once that unit is mind controlled, the game will crash when the unit goes back to it's original onwer.

I am not totally sure what caused the second crash, but it started to happen once I made the Mind Control unit unable to affect units upgraded by the conflicting buff.

The game stopped crashing once I disabled the mind control unit (tested three or four times).

```
Shattered Paradise Mod at Version {DEV_VERSION}
on map bcc00f3f544adef433c0b08d26a4f4852a385acf (Drawbridges by Westwood Studios).
Date: 2016-07-17 05:08:54Z
Exception of type `System.InvalidOperationException`: Object <OpenRA.Mods.Common.Traits.UpgradeActorsNear> revoked more levels of upgrade firepower than it granted for wolf 1295.
   en OpenRA.Mods.Common.Traits.UpgradeManager.RevokeUpgrade(Actor self, String upgrade, Object source)
   en OpenRA.Mods.Common.Traits.UpgradeActorsNear.ActorExited(Actor a)
   en OpenRA.Traits.ActorMap.ProximityTrigger.Dispose() en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Traits\World\ActorMap.cs:línea 157
   en OpenRA.Traits.ActorMap.RemoveProximityTrigger(Int32 id) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Traits\World\ActorMap.cs:línea 468
   en OpenRA.Mods.Common.Traits.UpgradeActorsNear.RemovedFromWorld(Actor self)
   en OpenRA.World.Remove(Actor a) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\World.cs:línea 275
   en OpenRA.Actor.<>c__DisplayClass11.<ChangeOwner>b__10(World w) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Actor.cs:línea 299
   en OpenRA.World.Tick() en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\World.cs:línea 347
   en OpenRA.Game.InnerLogicTick(OrderManager orderManager) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Game.cs:línea 542
   en OpenRA.Game.LogicTick() en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Game.cs:línea 567
   en OpenRA.Game.Loop() en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Game.cs:línea 694
   en OpenRA.Game.Run() en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Game.cs:línea 734
   en OpenRA.Program.Run(String[] args) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Support\Program.cs:línea 119
   en OpenRA.Program.Main(String[] args) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Support\Program.cs:línea 40
```

```
Shattered Paradise Mod at Version {DEV_VERSION}
on map bcc00f3f544adef433c0b08d26a4f4852a385acf (Drawbridges by Westwood Studios).
Date: 2016-07-21 18:25:49Z
Exception of type `System.InvalidOperationException`: Object <OpenRA.Mods.Common.Traits.UpgradeActorsNear> revoked more levels of upgrade speeddebuff than it granted for mutfiend 1781.
   en OpenRA.Mods.Common.Traits.UpgradeManager.RevokeUpgrade(Actor self, String upgrade, Object source)
   en OpenRA.Mods.Common.Traits.UpgradeActorsNear.ActorExited(Actor a)
   en OpenRA.Traits.ActorMap.ProximityTrigger.Tick(ActorMap am) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Traits\World\ActorMap.cs:línea 148
   en OpenRA.Traits.ActorMap.Tick(Actor self) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Traits\World\ActorMap.cs:línea 409
   en OpenRA.World.<Tick>b__a(TraitPair`1 x) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\World.cs:línea 341
   en OpenRA.WorldUtils.DoTimed[T](IEnumerable`1 e, Action`1 a, String text) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\WorldUtils.cs:línea 75
   en OpenRA.World.Tick() en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\World.cs:línea 0
   en OpenRA.Game.InnerLogicTick(OrderManager orderManager) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Game.cs:línea 542
   en OpenRA.Game.LogicTick() en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Game.cs:línea 567
   en OpenRA.Game.Loop() en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Game.cs:línea 694
   en OpenRA.Game.Run() en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Game.cs:línea 734
   en OpenRA.Program.Run(String[] args) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Support\Program.cs:línea 119
   en OpenRA.Program.Main(String[] args) en c:\Users\Alfred\Documents\OpenRA-Bleed\OpenRA.Game\Support\Program.cs:línea 40
```

GraionDilach:
If an actor changes ownership while under the active effect of a
proximity trigger affecting the two players differently, the trigger
misses out to update the actor in the moment of the owner-change
reflecting that and will try to revoke the new owner's states when the
actor later leaves the trigger while leaving the old owner's state
there.

This can crash the game if one of these proximity triggers is an
UpgradeActorsNear, which is supposed to only grant upgrades to the new
owner - it misses granting them but later tries to properly revoke them.

Credits to lucassss for figuring out the underlying issue.

Fixes #11707.

### DIFF
--- a/OpenRA.Game/Traits/World/ActorMap.cs
+++ b/OpenRA.Game/Traits/World/ActorMap.cs
@@ -49,15 +49,17 @@ namespace OpenRA.Traits
 
 			readonly Action<Actor> onActorEntered;
 			readonly Action<Actor> onActorExited;
+			readonly Action<Tuple<Actor, Player, Player>> onOwnerChanged;
 
 			IEnumerable<Actor> currentActors = Enumerable.Empty<Actor>();
 
-			public CellTrigger(CPos[] footprint, Action<Actor> onActorEntered, Action<Actor> onActorExited)
+			public CellTrigger(CPos[] footprint, Action<Actor> onActorEntered, Action<Actor> onActorExited, Action<Tuple<Actor, Player, Player>> onOwnerChanged)
 			{
 				Footprint = footprint;
 
 				this.onActorEntered = onActorEntered;
 				this.onActorExited = onActorExited;
+				this.onOwnerChanged = onOwnerChanged;
 
 				// Notify any actors that are initially inside the trigger zone
 				Dirty = true;
@@ -67,6 +69,18 @@ namespace OpenRA.Traits
 			{
 				if (!Dirty)
 					return;
+
+				if (actorMap.ActorsWithChangedOwner.Any())
+				{
+					var ownerschanged = currentActors.Intersect(actorMap.ActorsWithChangedOwner.Select(x => x.Item1));
+					foreach (var a in actorMap.ActorsWithChangedOwner)
+					{
+						if (ownerschanged.Contains(a.Item1))
+						{
+							onOwnerChanged(a);
+						}
+					}
+				}
 
 				var oldActors = currentActors;
 				currentActors = Footprint.SelectMany(actorMap.GetActorsAt).ToList();
@@ -95,6 +109,7 @@ namespace OpenRA.Traits
 
 			readonly Action<Actor> onActorEntered;
 			readonly Action<Actor> onActorExited;
+			readonly Action<Tuple<Actor, Player, Player>> onOwnerChanged;
 
 			WPos position;
 			WDist range;
@@ -102,10 +117,12 @@ namespace OpenRA.Traits
 
 			IEnumerable<Actor> currentActors = Enumerable.Empty<Actor>();
 
-			public ProximityTrigger(WPos pos, WDist range, WDist vRange, Action<Actor> onActorEntered, Action<Actor> onActorExited)
+			public ProximityTrigger(WPos pos, WDist range, WDist vRange,
+				Action<Actor> onActorEntered, Action<Actor> onActorExited, Action<Tuple<Actor, Player, Player>> onOwnerChanged)
 			{
 				this.onActorEntered = onActorEntered;
 				this.onActorExited = onActorExited;
+				this.onOwnerChanged = onOwnerChanged;
 
 				Update(pos, range, vRange);
 			}
@@ -128,6 +145,18 @@ namespace OpenRA.Traits
 			{
 				if (!Dirty)
 					return;
+
+				if (am.ActorsWithChangedOwner.Any())
+				{
+					var ownerschanged = currentActors.Intersect(am.ActorsWithChangedOwner.Select(x => x.Item1));
+					foreach (var a in am.ActorsWithChangedOwner)
+					{
+						if (ownerschanged.Contains(a.Item1))
+						{
+							onOwnerChanged(a);
+						}
+					}
+				}
 
 				var oldActors = currentActors;
 				var delta = new WVec(range, range, WDist.Zero);
@@ -175,6 +204,8 @@ namespace OpenRA.Traits
 		readonly HashSet<Actor> addActorPosition = new HashSet<Actor>();
 		readonly HashSet<Actor> removeActorPosition = new HashSet<Actor>();
 		readonly Predicate<Actor> actorShouldBeRemoved;
+
+		public HashSet<Tuple<Actor, Player, Player>> ActorsWithChangedOwner = new HashSet<Tuple<Actor, Player, Player>>();
 
 		public ActorMap(World world, ActorMapInfo info)
 		{
@@ -407,12 +438,14 @@ namespace OpenRA.Traits
 
 			foreach (var t in proximityTriggers)
 				t.Value.Tick(this);
+
+			ActorsWithChangedOwner.Clear();
 		}
 
-		public int AddCellTrigger(CPos[] cells, Action<Actor> onEntry, Action<Actor> onExit)
+		public int AddCellTrigger(CPos[] cells, Action<Actor> onEntry, Action<Actor> onExit, Action<Tuple<Actor, Player, Player>> onChangedOwner)
 		{
 			var id = nextTriggerId++;
-			var t = new CellTrigger(cells, onEntry, onExit);
+			var t = new CellTrigger(cells, onEntry, onExit, onChangedOwner);
 			cellTriggers.Add(id, t);
 
 			foreach (var c in cells)
@@ -444,10 +477,10 @@ namespace OpenRA.Traits
 			}
 		}
 
-		public int AddProximityTrigger(WPos pos, WDist range, WDist vRange, Action<Actor> onEntry, Action<Actor> onExit)
+		public int AddProximityTrigger(WPos pos, WDist range, WDist vRange, Action<Actor> onEntry, Action<Actor> onExit, Action<Tuple<Actor, Player, Player>> onChangedOwner)
 		{
 			var id = nextTriggerId++;
-			var t = new ProximityTrigger(pos, range, vRange, onEntry, onExit);
+			var t = new ProximityTrigger(pos, range, vRange, onEntry, onExit, onChangedOwner);
 			proximityTriggers.Add(id, t);
 
 			foreach (var bin in BinsInBox(t.TopLeft, t.BottomRight))
--- a/OpenRA.Game/Actor.cs
+++ b/OpenRA.Game/Actor.cs
@@ -301,6 +301,8 @@ namespace OpenRA
 				Owner = newOwner;
 				Generation++;
 
+				w.ActorMap.ActorsWithChangedOwner.Add(new Tuple<Actor, Player, Player>(this, oldOwner, newOwner));
+
 				foreach (var t in TraitsImplementing<INotifyOwnerChanged>())
 					t.OnOwnerChanged(this, oldOwner, newOwner);
 
--- a/OpenRA.Mods.Common/Scripting/Global/TriggerGlobal.cs
+++ b/OpenRA.Mods.Common/Scripting/Global/TriggerGlobal.cs
@@ -313,7 +313,7 @@ namespace OpenRA.Mods.Common.Scripting
 				}
 			};
 
-			triggerId = Context.World.ActorMap.AddCellTrigger(cells, invokeEntry, null);
+			triggerId = Context.World.ActorMap.AddCellTrigger(cells, invokeEntry, null, null);
 
 			return triggerId;
 		}
@@ -340,7 +340,36 @@ namespace OpenRA.Mods.Common.Scripting
 				}
 			};
 
-			triggerId = Context.World.ActorMap.AddCellTrigger(cells, null, invokeExit);
+			triggerId = Context.World.ActorMap.AddCellTrigger(cells, null, invokeExit, null);
+
+			return triggerId;
+		}
+
+		[Desc("Call a function when a ground-based actor changes it's owner within this cell footprint. " +
+			"Returns the trigger id for later removal using RemoveFootprintTrigger(int id). " +
+			"The callback function will be called as func(Actor a, Player oldOwner, Player newOwner, int id).")]
+		public int OnOwnerChangedFootprint(CPos[] cells, LuaFunction func)
+		{
+			// We can't easily dispose onOwnerChange, so we'll have to rely on finalization for it.
+			var onOwnerChange = (LuaFunction)func.CopyReference();
+			var triggerId = 0;
+			Action<Tuple<Actor, Player, Player>> invokeOwnerChange = a =>
+			{
+				try
+				{
+					using (var luaActor = a.Item1.ToLuaValue(Context))
+					using (var luaOldOwner = a.Item2.ToLuaValue(Context))
+					using (var luaNewOwner = a.Item3.ToLuaValue(Context))
+					using (var id = triggerId.ToLuaValue(Context))
+						onOwnerChange.Call(luaActor, luaOldOwner, luaNewOwner, id).Dispose();
+				}
+				catch (Exception e)
+				{
+					Context.FatalError(e.Message);
+				}
+			};
+
+			triggerId = Context.World.ActorMap.AddCellTrigger(cells, null, null, invokeOwnerChange);
 
 			return triggerId;
 		}
@@ -373,7 +402,7 @@ namespace OpenRA.Mods.Common.Scripting
 				}
 			};
 
-			triggerId = Context.World.ActorMap.AddProximityTrigger(pos, range, WDist.Zero, invokeEntry, null);
+			triggerId = Context.World.ActorMap.AddProximityTrigger(pos, range, WDist.Zero, invokeEntry, null, null);
 
 			return triggerId;
 		}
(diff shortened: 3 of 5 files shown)


OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -376,6 +376,11 @@ static void RemoveInfluenceInner(ref InfluenceNode influenceNode, Actor toRemove
 
 		public void Tick(Actor self)
 		{
+			UpdateTriggers();
+		}
+
+		public void UpdateTriggers()
```
pchote: Updating *all* triggers every time *one* actor changes is a bad idea for a couple of reasons: duplicated/redundant effort, and inconsistent run orders.

Instead, you should make this method set a dirty flag that is then checked during Tick, keeping the real update there.

GraionDilach: You're missing the crucial issue here: by the time the Tick runs, the old owner data is lost.

pchote: You'll need to expand the interface to the ActorMap to deal with that.  Brute-forcing the timer updates isn't a fix that we can merge.

GraionDilach:
Updated.

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -402,13 +402,43 @@ public void Tick(Actor self)
...
+		public void AddActor(Actor a)
+		{
+			var pos = a.CenterPosition;
+			var col = WorldCoordToBinIndex(pos.X).Clamp(0, cols - 1);
+			var row = WorldCoordToBinIndex(pos.Y).Clamp(0, rows - 1);
+			var bin = BinAt(row, col);
+
+			bin.Actors.Add(a);
+
+			UpdateTriggers();
```
pchote: These still have the same problem of ticking the triggers at multiple and arbitrary times.

GraionDilach:
Updated again.

pchote:
This still doesn't look safe to me...

A more robust approach may be to explicitly record owner-changed events in the trigger and then let the things that bind the trigger deal with these themselves.

GraionDilach:
Thing is: owner-changing runs in a frame-end task, and in order to maintain sync with proxtriggers, they also need to be updated in that frame-end task.

pchote:
This is why the fix needs to be carefully designed and integrated.  The timing/order of execution is subtle and important.  Brute-forcing the triggers to update at arbitrary times *will* break edge cases that we probably won't predict.

GraionDilach:
If timing/order of execution is important, then why you keep insisting to break the actor-proxtrigger synchronization by a tick (unless you want to introduce a race condition when the trigger provider gets changed in the followup frame of an owner change on an affected-by-trigger actor).

GraionDilach:
Also this PR as-is doesn't resolve the crash because the affected proxtriggfers on the actor doesn't get updated. The previous iteration could fix it, but this doesn't anymore.

pchote:
I'll be blunt because your comments are coming across as increasingly aggressive.

You want to make it possible to execute the triggers from an arbitrary point in a FrameEndTask.  This means that any logic run by the trigger (i'll point out here that you've forgotten about cell triggers in this pr) becomes sensitive to the order that FrameEndTasks are executed.

Consider a lua-bound trigger that makes a player lose if an enemy unit crosses a trigger region.  An enemy captures one of your units inside the region during the same tick that that is killed.  If the capture task runs before the kill task, then you lose and go WTF there was no enemy unit in my area!?!  If the kill task runs first, then everything is fine.  The order that the tasks run will typically depend on the order that the units were initialized by the game.

Consider a trigger that changes the owner of something within its range.  Congratulations, you've introduced a potential stack overflow.

There are a whole pile of subtle issues like this that the trigger system was specifically set up to avoid.  Running the triggers at a single known time each tick is the only way we can safely avoid them.

GraionDilach:
Updated.

OpenRA.Game/Traits/World/ActorMap.cs
```diff
@@ -47,12 +47,12 @@ class CellTrigger
 			public readonly CPos[] Footprint;
 			public bool Dirty;
 
-			readonly Action<Actor> onActorEntered;
-			readonly Action<Actor> onActorExited;
+			readonly Action<Tuple<Actor, Player>> onActorEntered;
+			readonly Action<Tuple<Actor, Player>> onActorExited;
```
pchote: I believe it would be cleaner (also somewhat more efficient) to maintain a `readonly Action<Tuple<Actor, Player, Player> changedOwner` instead of adding `Player`s to all of our existing collections.

This is what I was asking for in my earlier requests, but looking back at what I wrote that probably wasn't clear.

GraionDilach: So all Triggers would have that instead of this?

pchote: Yes, and each list would have on average zero elements inside them.

GraionDilach: Alright, but then let's agree on an order of execution as well before I run another round at here. So currently it's `OnEntered` then `OnExited`, what I'd change `OnExited` then `OnEntered` tbh. Make `OnActorChanged` the last?

pchote: Making OnExited run first just makes sense.  I may have had a specific bug-related reason for having it backwards, but lets hope not.

I don't see a strong usecase at this point for having the things that bind to triggers handle actor changes.  Run OnExited on the exiting actors, and then on any owner-changed actors that haven't left the region.  Then repeat the process for OnEntered.

chrisforbes:
Are the invalidation conditions the same or different to those used for Targets?

OpenRA.Game/Traits/World/ActorMap.cs
GraionDilach: I'm currently in the middle of an update of this and I realized - you have to start with `OnActorChanged`. While doing that does trigger in your "makes a player lose if an enemy unit crosses a trigger region. An enemy captures one of your units inside the region during the same tick that that is killed" example, having `OnExited` running before `OnActorChanged` would crash the game if a player hijacks an enemy tank which was just leaving his Cloak Generator's range in TS (since the Mutant Hijacker is also a Nod unit).

pchote: I think you missed an important detail here: if the unit leaves the trigger area then *you discard/ignore it from the owner changed list*.  `OnExited` should never be able to run twice for the same actor in the same tick.

GraionDilach:
Updated.

pchote:
Please see my comment above re owner / exit ordering.

GraionDilach:
No, you missed an important detail here: _the unit still leaves the trigger a tick later compared to leaving the area affected_. If I do what you say, it will just render #11707 to a heisencrash which will still happen in the situation I described above the post you responded.

Also, `OnExited` never runs twice. Please review the actual code.

pchote:
Please explain the code flow that you think causes the problem, and we can go from there.  The entire motivation for my suggestions above were to ensure that actor change events would be managed in the same tick as the movement triggers.

pchote:
> Also, OnExited never runs twice. Please review the actual code.

My comment above was with regards to

> having OnExited running before OnActorChanged would crash the game if a player hijacks an enemy tank which was just leaving his Cloak Generator's range in TS

Please explain what you mean by "crash" if you aren't referring to `OnExited` running twice and double-revoking UpgradeActorsNear.

GraionDilach:
If `OnOwnerChanged` in this code iteration is shifted to happen only after `OnExited`, then `OnExited` will crash the game by removing the cloakgenerator upgrade which the captured unit could only acquire from `OnOwnerChanged`.

GraionDilach:
Example situation (although it will be the vice versa problem, where no crash happens with assuming on current TS rules):

Unit A owned by Player X is under X's cloakgenerator and X ordered it to leave it.
It leaves the area in Tick 33, and in this same Tick 33, player Y successfully hijacks it.
Between Tick 33 and Tick 34, the capture function runs.
In Tick 34, ActorMap.Tick calls the trigger updates.
IF OnOwnerChanged runs before Exited/Entered
 - OnOwnerChanged revokes the cloakgen since X lost the unit and will not apply the upgrade back since Y isn't an ally of X.
 - OnExited does nothing, since Y isn't an ally of X.
A leaves the area with the cloak upgrade removed.
IF OnExited/Entered runs first
 - OnExited does nothing, since Y isn't an ally of X.
 - On OwnerChanged will be prevented to run, since the unit is no longer in the area of the cloakgen.
A leaves the are with the cloakgenerator upgrade still enabled.

This is the actual issue.

pchote:
All these problems come from `UpgradeActorsNear` having no sanity checks preventing it from granting upgrades more than once or removing more than it has applied.  It would be good practice, regardless of what happens outside the trait, for it to keep its own `HashSet<Actor>` and skip grant/revokes that would violate these rules.

We can then remove the stances check from the revoke case so that any actor that has been granted an upgrade will lose it when it leaves.

We then have, with `OnExit` running before `OnOwnerChanged`:
* At the start of the tick we have the unit owned by the "wrong" player and the cloak upgrade applied
* `OnExit` runs, and the upgrade is removed because it has left the trigger area (owner doesn't matter).
* `OnOwnerChanged` doesn't run, because `OnExit` has been run.
* At the end of the tick we have the enemy actor outside the trigger without the upgrade applied.

Or alternatively (actor captured enemy -> ally when leaving a trigger):
* At the start of the tick we have the unit owned by the "right" player, with no upgrade applied
* `OnExit` runs, and nothing happens because there is no upgrade to revoke.
* `OnOwnerChanged` doesn't run, because `OnExit` has been run.
* At the end of the tick we have the allied actor outside the trigger without the upgrade applied.

And on the reverse case, with `OnEnter` running before `OnOwnerChanged`:
* At the start of the tick we have the unit owned by the "wrong" player, with no upgrade applied.
* `OnEnter` runs, and the upgrade isn't applied because the owner is wrong.
* `OnOwnerChanged` doesn't run, because `OnEnter` has been run.
* At the end of the tick we have the enemy actor inside the trigger without the upgrade applied.

Or alternatively (actor captured enemy -> ally when entering a trigger):
* At the start of the tick we have the unit owned by the "right" player, with no upgrade applied.
* `OnEnter` runs, and the upgrade is applied because the owner is correct.
* `OnOwnerChanged` doesn't run, because `OnEnter` has been run.
* At the end of the tick we have the allied actor inside the trigger the the upgrade applied.

These are the behaviors we want for `UpgradeActorsNear`, without regressing the cases that I explained above.

pchote:
>  It would be good practice, regardless of what happens outside the trait, for it to keep its own `HashSet<Actor>` and skip grant/revokes that would violate these rules.

If we're smart about this then we don't even need a new `HashSet`.  The upgrades code already tracks which objects have granted levels, so we just need to expose a method to check whether `this` (the `UpgradeActorsNear` trait) has granted an upgrade level or not.

GraionDilach:
Sounds like a "downstream-workaround-for-an-upstream-issue" hack to me if you insist changing the PR's current contents as well.

pchote:
I've tried to be reasonable, and have explained why I consider the approach in this PR to be flawed.  I have also spent a lot more time than I consider worthwhile giving directions towards fixing these, but so far all you've done is be hostile towards these review comments.

Do you want to work with me towards getting this important fix merged, or shall we give up and close this PR?

GraionDilach:
I don't see why the trigger-sanity checks should be strictly relegated to the trigger usecases besides that at the moment there's no proximity trigger applying a continuous effect except UAN.

pchote:
UAN is a special case because it uses the trigger notifications to modify *other* actors.  This means that it needs to deal with the timing and state consistency of other actors, not just itself (which is handled by the triggers themselves).

GraionDilach:
I don't see what makes that so much of a special case that it needs safeguards relegated to the trigger user code - I haven't seen any comment where said effect is beyond the trigger's intended use (considering that the Lua trigger descriptions also imply them be used to affect all actors in the affected area).

pchote:
Are you really trying to argue that UAN is so simple that we should intentionally avoid adding sanity or safety checks?

GraionDilach:
No, I am trying to argue that UAN shouldn't require safety checkes in itself and the trigger code should provide the required checks for the complicated case of UAN as well.

pchote:
This is a notification ordering problem, which is something that we don't generally have control over.  Traits are expected to be robust against the order that notifications are called, which is why we have FrameEndTasks and Actor.IsDead checks everywhere.

This is a specific case where we *can* control it, but that is not a valid excuse to avoid our best practices.  Changing the engine code to depend on the implementation details of one specific trait in mod code is a terrible idea.  When the next person comes along and wants to change it back (because they don't see why their code should implement sanity checks when they can just change the trigger order) we end up back at square one with hard to reproduce crashes.

pchote:
I think this discussion has run its course.  I will end up solving some of these problems myself in the upgrades rewrite which is now underway.  Moving the the future milestone so that we can find and resurrect the remaining parts of this once that painful process is complete.

pchote:
#12449 implements the UAN rework that I outlined above.

PR #11992, opened 2016-09-06
 Remove bogus LZRange field and unnecessary ParaDrop definition

abcdefg30:
Looks like `LZRange` was removed from the code long ago.

### DIFF
--- a/mods/cnc/rules/aircraft.yaml
+++ b/mods/cnc/rules/aircraft.yaml
@@ -165,8 +165,6 @@ ORCA:
 
 C17:
 	Inherits: ^Plane
-	ParaDrop:
-		LZRange: 1
 	Tooltip:
 		Name: Supply Aircraft
 		Description: Drops vehicle reinforcements on Airstrips


obrakmann:
I didn't even know the C17 was set up to do that. :+1:

pchote:
I'd prefer it if we just removed the trait.  We don't have any of the other paradrop plumbing set up in TD, so it isn't at all useful.

abcdefg30:
Hm, that makes sense. But now that I think of it, we could break custom maps which enable crate dropping with `c17`s as `DeliveryAircraft` (like the `Vicemania` map, which defines the `ParaDrop` trait itself, though).

abcdefg30:
However, just see that as thinking out loud. The possibility that such a thing happens is low, thus updated.

abcdefg30:
@obrakmann sorry, seems like I invalidated your 👍 with that change.

reaperrr:
:+1:

obrakmann:
still ok, :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/4a0e86b81a79f8217769f572fba03cc5a64a2419)

PR #11994, opened 2016-09-06
gdi07 map yaml polish

clemty:
- map will now correctly show up in the map editor with the other campaign maps
- changes owner of chainlink fence to Nod (Nod tanks will no longer run their own fences over when you approach their base from the left)
- changes owner of brick walls to Nod (should not make a difference in game)

### DIFF
--- a/mods/cnc/maps/gdi07/map.yaml
+++ b/mods/cnc/maps/gdi07/map.yaml
@@ -14,7 +14,7 @@ Bounds: 4,1,58,50
 
 Visibility: MissionSelector
 
-Categories: Mission
+Categories: Campaign
 
 LockPreview: True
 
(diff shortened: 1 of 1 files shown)


Mailaender:
This should not cause problems since we have https://github.com/OpenRA/OpenRA/pull/11800 now. :+1:

clemty:
to clarify: Nod tanks didn't attack the chainlink fence, they just tried to move towards my GDI forces in the fastest way, crushing the chainlink fence (this would be expected behavior if the chainlink fence was in my base, but not in theirs). The same issue was present in gdi05 and was fixed last year, so I don't think that has anything to do with #11800

abcdefg30:
Thanks for the cleanup. 👍 / ✅

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/8107ecc01e42a792b4e230ac7f7fe67a765f4e69...61869ae999c04304c003316455e34af6769febde)
I also think this is suited for stable: 23f9998c7effbef5e91d1fb62ebe87e56ee408ec

PR #11998, opened 2016-09-07
Visualising waypoints2

Turupawn:
This solves #2598. It's a #11794 follow up which I had to do another branch because I wasn't able to merge the lastest changes (sorry about that). You guys work fast!

I'm just concerned about `LineTargetExts::SetTargetLine()` because it looks like `LineTargetExts::SetTargetLines()` can do all the trick. Should we remove all SetTargetLine and use SetTargetLines only?
see:
https://github.com/OpenRA/OpenRA/pull/11998/files#diff-d7a57eebc0a2b3d470e099590d38922eR95


OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
```diff
@@ -66,37 +65,40 @@ IEnumerable<IRenderable> IRenderAboveShroudWhenSelected.RenderAboveShroud(Actor
...
+				validTargets.Add(target.CenterPosition);
 			}
+
+			return new[] { (IRenderable)new TargetLineRenderable(validTargets, c) };
 		}
 
 		void INotifyBecomingIdle.OnBecomingIdle(Actor a)
 		{
-			if (a.IsIdle)
-				targets = null;
```
WolfGaming: Please remove this method and the interface it is implementing (INotifyBecomingIdle) since the class isn't using it anymore.

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
```diff
@@ -37,16 +38,14 @@ public DrawLineToTarget(Actor self, DrawLineToTargetInfo info)
...
 		{
-			targets = new List<Target> { target };
 			this.c = c;
 
 			if (display)
 				lifetime = info.Delay;
 		}
 
-		public void SetTargets(Actor self, List<Target> targets, Color c, bool display)
+		public void SetTargets(Actor self, Color c, bool display)
```
WolfGaming: This method now has nothing to do with setting targets and should be renamed.

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
```diff
@@ -37,16 +38,14 @@ public DrawLineToTarget(Actor self, DrawLineToTargetInfo info)
 
 		public void SetTarget(Actor self, Target target, Color c, bool display)
```
WolfGaming: Same as SetTargets.

OpenRA.Mods.Common/Activities/Move/Move.cs
```diff
@@ -280,11 +280,22 @@ public override void Cancel(Actor self)
...
-				return Enumerable.Reverse(path).Select(c => Target.FromCell(self.World, c));
-			if (destination != null)
-				return new Target[] { Target.FromCell(self.World, destination.Value) };
-			return Target.None;
+			var validTargets = new List<Target>();
+
+			var activityIterator = (Activity)this;
+			while (activityIterator != null)
+			{
+				if (!activityIterator.IsCanceled && activityIterator is Move)
```
pchote: Why is move still special cased here?

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
```diff
@@ -66,37 +65,40 @@ IEnumerable<IRenderable> IRenderAboveShroudWhenSelected.RenderAboveShroud(Actor
...
-				yield break;
+				return new IRenderable[0];
 
 			if (!(force || Game.Settings.Game.DrawTargetLine))
-				yield break;
+				return new IRenderable[0];
 
-			if (targets == null || targets.Count == 0)
-				yield break;
+			if (self.GetCurrentActivity() == null)
```
WolfGaming: you do self.GetCurrentActivity() 3 times, please put it inside a variable.

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
```diff
@@ -66,37 +65,40 @@ IEnumerable<IRenderable> IRenderAboveShroudWhenSelected.RenderAboveShroud(Actor
...
+				return new IRenderable[0];
 
-			foreach (var target in targets)
-			{
-				if (target.Type == TargetType.Invalid)
-					continue;
+			if (self.GetCurrentActivity().GetTargets(self).Count() <= 0)
```
WolfGaming: You also do GetTargets(self) 2 times,put that into a variable too as it can become expensive depending on the implementation of the activity.

OpenRA.Mods.Common/Activities/Move/Move.cs
pchote: Err, hang on... You really don't want to be walking the activity queue from *inside* an activity on the queue!

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
pchote: It would be much more robust to remove this completely, and have `GetTargets` return the color as well as the destination.

OpenRA.Mods.Common/Traits/Render/DrawLineToTarget.cs
```diff
@@ -66,37 +65,40 @@ IEnumerable<IRenderable> IRenderAboveShroudWhenSelected.RenderAboveShroud(Actor
...
+				return new IRenderable[0];
+
+			var validTargets = new List<WPos>();
+			validTargets.Add(self.CenterPosition);
 
-				yield return new TargetLineRenderable(new[] { self.CenterPosition, target.CenterPosition }, c);
+			foreach (var target in self.GetCurrentActivity().GetTargets(self))
```
pchote: You'll want to be enumerating over all the activities and collecting your target path to draw here.

Turupawn:
@pchote : I started working on the getTargets() refactoring and so far I changed this function:
https://github.com/OpenRA/OpenRA/pull/11998/commits/510c404a16ba562bb2fc3728ae35e96f1b4f6513#diff-27d71a7223c28efcd80ec32c4ca56ae4L39
There's still some glitches and stuff, but do you think this is heading in a good direction?

pchote:
Keep it as `IEnumerable<KeyValuePair<Target, Color>>`, otherwise you won't be able to deal with inner activities.

Turupawn:
Just wiped the LineTargetExts class with his SetTargetLine methods. So far the waypoints are still working fine. But now I have a couple of questions:
* How should we handle the lifetime of the lines now?
* How should we handle the color.
* Any idea why the Attack commands aren't rendering anymore?

LipkeGu:
Needs Rebase

reaperrr:
To clean up the PR queue, we're closing PRs with unadressed issues that haven't been updated for several weeks or more.

If you resume work on this, simply re-open it when you update your branch (or file a new PR).

PR #12001, opened 2016-09-09
Add New Shellmap to D2K

### LINKED ISSUE #12000: D2K - Adding a new Shellmap
D2K has old shellmap with small Harvesting Base for a long while, i think it needs one that has more action like RA and CnC ones. I started to make one but having issues for now. Just opening issue to say i'm on it.

MustaphaTR:
Fixes #12000.

~~I need to do something with harvesters first tho, either i'm gonna add check to make AI rebuild them or make them unkillable. Edit: I'll add Carryalls to give new Harvesters when old ones killed.~~ Edit: Added Carryall Reinforcements.

~~Also i'm not sure if(i checked it does) AI goes low on cash as Spice actually doesn't give anything here, i may make all units free too or remove `ResourceType@Spice:    ValuePerUnit: 0` from old map so they'll normally work.~~ Changed unit values to Zero ~~,but now they are insta-gaining 4th level Veterancy when they kill something.~~

I wanted to ask for this two first so haven't done anything about them yet.


mods/d2k/maps/shellmap/d2k-shellmap.lua
```diff
@@ -0,0 +1,534 @@
...
+		IdlingUnitsOfEmp[#IdlingUnitsOfEmp + 1] = unit[1]
+		Trigger.AfterDelay(delay, EmparorProduceTanks)
+
+		if #IdlingUnitsOfEmp >= (AttackGroupSize[1] * 2.5) then
+			SendAttackEmp()
+		end
+	end)
+end
+
+EmparorProduceStarport = function()
```
Mailaender: Typo: `Emperor`

mods/d2k/maps/shellmap/d2k-shellmap.lua
```diff
@@ -0,0 +1,534 @@
...
+		IdlingUnitsOfAtr[#IdlingUnitsOfAtr + 1] = unit[1]
+		Trigger.AfterDelay(delay, AtreidesProduceInfantry)
+
+		if #IdlingUnitsOfAtr >= (AttackGroupSize[1] * 2.5) then
+			SendAttackAtr()
+		end
+	end)
+end
+
+AtreidesProduceVehicles = function()
```
Mailaender: There is a lot of redundancy in your code. You can simplify it to

```lua
Produce(atreides, infantry)
```

with code like this

```lua
Produce = function(house, units)
	if HoldProduction[house] then
		Trigger.AfterDelay(DateTime.Minutes(1), units)
		return
	end

	local delay = Utils.RandomInteger(AttackDelay[1], AttackDelay[2])
	local toBuild = { Utils.Random(units) }
	atreides.Build(toBuild, function(unit)
		IdlingUnits[house][#IdlingUnits[house] + 1] = unit[1]
		Trigger.AfterDelay(delay, function() Produce(atreides, infantry) end)

		if #IdlingUnits[house] >= (AttackGroupSize[1] * 2.5) then
			SendAttack(house)
		end
	end)
end
```

Note: this isn't tested and needs changes elsewhere in the code as well. Hope you get the idea.

mods/d2k/maps/shellmap/d2k-shellmap.lua
MustaphaTR: Fixed locally. Will reupload soon.

Mailaender:
The map already looks and works great. :+1:

MustaphaTR:
Note that i won't be around for a week so i may not be able to finish this. Harvester reinforcement code is still not working. I actually managed to add a check but it gives lua eror when it happens so problem is with Reinforcement trigger.

mods/d2k/maps/shellmap/d2k-shellmap.lua
MustaphaTR: I'll take a look what i can do.

MustaphaTR:
Updated to make code available tfor you to see, it is currently not working, i need help.

`IdlingUnits` looks like having some problems right now. Harvesters also not working yet but i want to do production first.

MustaphaTR:
Ok hopefully two things left:

1. Harvesters after ejected from Carryall doesn't go to harvesting but stay.
2. Production triggers still crash for some reason about `IdlingUnits` with following:

```
Dune 2000 Mod at Version playtest-20160904
on map ab005862cdaace7925b9b963817ad03e321e56c4 (D2k Shellmap by Westwood Studios, MustaphaTR).
Date: 2016-09-16 11:42:56Z
Exception of type `Eluant.LuaException`: [string "d2k-shellmap.lua"]:10: attempt to get length of field '?' (a nil value)
   konum: Eluant.LuaRuntime.Call(IList`1 args)
   konum: Eluant.LuaRuntime.Call(LuaFunction fn, IList`1 args)
   konum: Eluant.LuaFunction.Call(IList`1 args)
   konum: OpenRA.Mods.Common.Scripting.ClassicProductionQueueProperties.<Build>c__AnonStorey1.<>m__0(Actor factory, Actor unit)
   konum: OpenRA.Mods.Common.Scripting.ClassicProductionQueueProperties.<ClassicProductionQueueProperties>c__AnonStorey0.<>m__0(Actor factory, Actor unit)
   konum: System.Action`2.Invoke(T1 arg1, T2 arg2)
   konum: OpenRA.Mods.Common.Scripting.ScriptTriggers.UnitProducedByOther(Actor self, Actor producee, Actor produced)
   konum: OpenRA.Mods.Common.Traits.Production.<DoProduction>c__AnonStorey1.<>m__0(World w)
   konum: OpenRA.World.Tick()
   konum: OpenRA.Game.InnerLogicTick(OrderManager orderManager)
   konum: OpenRA.Game.LogicTick()
   konum: OpenRA.Game.Loop()
   konum: OpenRA.Game.Run()
   konum: OpenRA.Program.Run(String[] args)
   konum: OpenRA.Program.Main(String[] args)
```

mods/d2k/maps/shellmap/d2k-shellmap.lua
```diff
@@ -0,0 +1,188 @@
...
+	har_cyard.Produce(Upgrades[3])
+	har_cyard.Produce(Upgrades[4])
+	ord_cyard.Produce(Upgrades[1])
+	ord_cyard.Produce(Upgrades[2])
+	ord_cyard.Produce(Upgrades[3])
+	ord_cyard.Produce(Upgrades[4])
+	emp_cyard.Produce(Upgrades[1])
+	emp_cyard.Produce(Upgrades[2])
+	emp_cyard.Produce(Upgrades[3])
+	emp_cyard.Produce(Upgrades[4])
```
abcdefg30: This can be simplified to something like:
```
Utils.Do(Utils.Take(4, Upgrades), function(upgrade)
	atr_cyard.Produce(upgrade)
	har_cyard.Produce(upgrade)
	ord_cyard.Produce(upgrade)
	emp_cyard.Produce(upgrade)
end)
atr_cyard.Produce(Upgrades[5])

File: mods/d2k/maps/shellmap/d2k-shellmap.lua [hunk 4]
```diff
@@ -0,0 +1,188 @@
...
+		end
+	end
+
+	return units
+end
+
+SendNewHarv = function(house, waypoint, count)
+	local harvesters = house.GetActorsByType("harvester")
+	if #harvesters < count then
+		Reinforcements.ReinforceWithTransport(house, "carryall.reinforce", Harvester , waypoint, { waypoint[1] })
```
abcdefg30: Trailing space after `Harvester`.

abcdefg30: Trailing space after `Harvester`.

abcdefg30: Trailing space after `Harvester`.

mods/d2k/maps/shellmap/d2k-shellmap.lua
abcdefg30: This can be simplified to something like:
```
Utils.Do(Utils.Take(4, Upgrades), function(upgrade)
	atr_cyard.Produce(upgrade)
	har_cyard.Produce(upgrade)
	ord_cyard.Produce(upgrade)
	emp_cyard.Produce(upgrade)
end)
atr_cyard.Produce(Upgrades[5])

File: mods/d2k/maps/shellmap/d2k-shellmap.lua (hunk 4 above)
Comment (MustaphaTR) [inline mods/d2k/maps/shellmap/d2k-shellmap.lua]: Fixed.

File: mods/d2k/maps/shellmap/d2k-shellmap.lua (hunk 3 above)
Comment (MustaphaTR) [inline mods/d2k/maps/shellmap/d2k-shellmap.lua]: Fixed locally.

**Comment by MustaphaTR** [comment]
Ok, i think everything works fine. At least it no longer crashes.

For my hacky LUA harvester reinforcements instead of using HarvesterInsurance, i don't wanna touch the code itself, if someone is interested in expanding, that hacky LUA code can be removed.

**Comment by reaperrr** [comment]
@MustaphaTR: just add `-GainsExperience:` to the `^Vehicle` and `^Infantry` defaults of your map rules. I think it's better to have no veterancy on the shellmap than these instant-elites.

Otherwise it looks good enough to me. :+1:

**Comment by Micr0Bit** [comment]
:+1:

File: mods/d2k/rules/player.yaml [hunk 5]
```diff
@@ -87,7 +87,6 @@ Player:
 		Prerequisites: techlevel.low, techlevel.medium, techlevel.high, techlevel.superweapons
 		Id: unrestricted
 	EnemyWatcher:
-	HarvesterInsurance:
```
reaperrr: Uh, this disables it for the whole mod!
You should move this to the shellmaps' rules.yaml.

reaperrr (changes requested):
In summary:
- disable GainsExperience in shellmap/rules.yaml
- disable HarvesterInsurance in in shellmap/rules.yaml, not in rules/player.yaml!

mods/d2k/rules/player.yaml
MustaphaTR: Why i did this is because original game didn't have such thing in Skirnish/MP, it is a Mission Logic and what HarvesterInsurance currently does is not same as Original. It needs to be rewrote.

MustaphaTR:
Disabling GainsExperience alone was causing a problem, not remember what exactly, i think i need to remove some other tags with it but not sure what. I said about HarvesterInsurance above.

MustaphaTR:
Updated and fixed:
> disable GainsExperience in shellmap/rules.yaml

I still think HarvesterInsurance must be complately removed, if not should be moved to `campaign-rules.yaml` so will only be in Missions.

reaperrr:
> I still think HarvesterInsurance must be complately removed, if not should be moved to campaign-rules.yaml so will only be in Missions.

Ok.

> Disabling GainsExperience alone was causing a problem, not remember what exactly

I don't see why that should cause any problems, but maybe it has to do with the new player experience system.
Instead of disabling the trait directly, you could set `PlayerExperienceModifier: 0` like in
`rules/campaign-disable-experience.yaml`, that's probably safer.

MustaphaTR:
> I don't see why that should cause any problems, but maybe it has to do with the new player experience system.
> Instead of disabling the trait directly, you could set PlayerExperienceModifier: 0 like in
rules/campaign-disable-experience.yaml, that's probably safer.

It worked this time with current code, for some reason it didn't and still doesn't work if i add that one by one to all units. I didn't think about disabling it from `^GainsExperience:` before.

reaperrr (approved):

reaperrr:
One last thing I noticed: is there a way to make the AIs send larger forces more quickly at each other? Currently, after some minutes a lot of units are clogging up the bases.

abcdefg30:
> I still think HarvesterInsurance must be complately removed, if not should be moved to campaign-rules.yaml so will only be in Missions.

I somewhat disagree. We should keep it for now, even in singleplayer. Otherwise worms get quite frustrating, especially to new players (who aren't used to looking after their harvesters) and AIs (in the past we had issues with the AIs stalling because all their harvesters were eaten).
I'm fine with expanding it for missions to replace all lost harvesters.

However, this change should be made in a new PR, as it is not really related to the new shellmap.

reaperrr:
Actually, I think @abcdefg30 is right.

MustaphaTR:
Ok, @abcdefg30 has a point, i'll temove that commit and add increase attach foce size when i find time.

reaperrr:
@MustaphaTR as you see, disabling experience this way results in a lint error, seems it would be better to use the `PlayerExperienceModifier: 0` approach afterall.

MustaphaTR:
`PlayerExperienceModifier: 0` doesn't work at all. Also i'm not sure how exactly i can fix:

> is there a way to make the AIs send larger forces more quickly at each other? Currently, after some minutes a lot of units are clogging up the bases.

reaperrr:
Ok, I think that's good enough for now, we can fine-tune it later.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5e4566ddb90509af8d946ebcd1d028a8dd0a43bd...9d67b2a9274db3432ad02c098ca841ea12155ee7)

PR #12004, opened 2016-09-09
More projectile property streamlining and cleanups

reaperrr:
Another small step towards more consistency in property names, also some negligible performance tweaks and Lint rules to warn modders about invalid launch angles.

All yaml changes were performed with the respective upgrade rules.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckAngle.cs
+++ b/OpenRA.Mods.Common/Lint/CheckAngle.cs
@@ -1,0 +1,70 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2016 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System;
+using System.Linq;
+using OpenRA.Mods.Common.Projectiles;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.Common.Lint
+{
+	class CheckAngle : ILintRulesPass
+	{
+		public void Run(Action<string> emitError, Action<string> emitWarning, Ruleset rules)
+		{
+			foreach (var weaponInfo in rules.Weapons)
+			{
+				var missile = weaponInfo.Value.Projectile as MissileInfo;
+				if (missile != null)
+				{
+					var minAngle = missile.MinimumLaunchAngle.Angle;
+					var maxAngle = missile.MaximumLaunchAngle.Angle;
+
+					// If both angles are identical, we only need to test one of them
+					var testMaxAngle = minAngle != maxAngle;
+					CheckLaunchAngles(weaponInfo.Key, minAngle, testMaxAngle, maxAngle, emitError);
+				}
+
+				var bullet = weaponInfo.Value.Projectile as BulletInfo;
+				if (bullet != null)
+				{
+					var minAngle = bullet.LaunchAngle[0].Angle;
+					var maxAngle = bullet.LaunchAngle.Length > 1 ? bullet.LaunchAngle[1].Angle : minAngle;
+
+					// If both angles are identical, we only need to test one of them
+					var testMaxAngle = minAngle != maxAngle;
+					CheckLaunchAngles(weaponInfo.Key, minAngle, testMaxAngle, maxAngle, emitError);
+				}
+			}
+		}
+
+		static bool InvalidAngle(int value)
+		{
+			return value > 255 && value < 769;
+		}
+
+		static void CheckLaunchAngles(string weaponInfo, int minAngle, bool testMaxAngle, int maxAngle, Action<string> emitError)
+		{
+			if (InvalidAngle(minAngle))
+				emitError("Weapon `{0}`: Projectile minimum LaunchAngle must not exceed (-)255!".F(weaponInfo));
+			if (testMaxAngle && InvalidAngle(maxAngle))
+				emitError("Weapon `{0}`: Projectile maximum LaunchAngle must not exceed (-)255!".F(weaponInfo));
+
+			if ((minAngle < 256) && (maxAngle < 256) && (minAngle > maxAngle))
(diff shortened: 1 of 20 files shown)


reaperrr:
> OpenRA.Utility(1,1): Error: OpenRA.Mods.Common.Lint.CheckAngle failed with exception: System.NullReferenceException: Object reference not set to an instance of an object

Does anyone have an idea what the hell is going on here?
Because this test *passes* on my local branch (I didn't make any changes since I pushed).

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
```diff
@@ -28,7 +28,7 @@ public class AreaBeamInfo : IProjectileInfo
 		public readonly WDist[] Speed = { new WDist(128) };
 
 		[Desc("The maximum duration (in ticks) of each beam burst.")]
-		public readonly int Duration = 10;
+		public readonly int BeamDuration = 10;
```
pchote: We already know that this is a beam, so adding it again here doesn't help.

OpenRA.Mods.Common/Projectiles/LaserZap.cs
```diff
@@ -23,6 +23,9 @@ namespace OpenRA.Mods.Common.Projectiles
 	[Desc("Not a sprite, but an engine effect.")]
 	public class LaserZapInfo : IProjectileInfo
 	{
+		[Desc("Should the beam be visually rendered? False = Beam is invisible.")]
+		public readonly bool RenderBeam = true;
```
pchote: A laser that isn't rendered isn't a laser at all (just a normal projectile), so isn't suited for this projectile type at all.

OpenRA.Mods.Common/Projectiles/LaserZap.cs
```diff
@@ -100,17 +103,18 @@ public void Tick(World world)
...
 		{
-			if (wr.World.FogObscures(target) &&
-				wr.World.FogObscures(args.Source))
+			if (!info.RenderBeam && hitanim == null)
 				yield break;
 
-			if (ticks < info.BeamDuration)
+			var targetObscured = wr.World.FogObscures(target);
+			var sourceObscured = wr.World.FogObscures(args.Source);
+			if (!(targetObscured && sourceObscured) && ticks < info.BeamDuration)
```
pchote: This will cause the whole beam to be invisible if you get shot by an obelisk hidden under the fog, which looks much worse than what we have right now (I had already tried this a long time ago).

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
reaperrr: I intend to add `DamageDuration` in my next PR and felt this would be better to separate them, but if you prefer to rename `BeamDuration` in `LaserZap` to just `Duration` instead, I guess I can do that.

reaperrr:
Updated.

However, I still need help with the lint check. For some reason it only works with positive values, negative angles always trigger the lint error (which is why I've disabled it for missiles for now).

pchote:
`WAngle` is only defined for numbers between 0 and 1024.  The `new WAngle(-64)` in Missile.cs is internally converted to 960, which then fails your > 255 check.

reaperrr:
> You check should be something like `return value > 255 && value < 768`;

Done, though I went for `value < 76*9*` because 768 should be the highest invalid value (1024 - 256 = 768).

OpenRA.Mods.Common/Projectiles/Missile.cs
```diff
@@ -37,20 +37,12 @@ public class MissileInfo : IProjectileInfo
...
-		public readonly WDist MinimumLaunchSpeed = new WDist(-1);
-
-		[Desc("Maximum launch speed in WDist / tick. Defaults to Speed if -1.")]
-		public readonly WDist MaximumLaunchSpeed = new WDist(-1);
-
-		[Desc("Maximum projectile speed in WDist / tick")]
-		public readonly WDist Speed = new WDist(384);
+		[Desc("Projectile speed in WDist / tick, two values indicate launch speed and maximum speed,",
+			"three values represent minimum launch speed, maximum launch speed and maximum speed (in that order).")]
+		public readonly WDist[] Speed = { new WDist(384) };
```
GraionDilach: Notcertain if I'm a fan of this, this forces unneeded duplication at inherited ones.

reaperrr:
Removed the first two laser-related commits, which are included in #12068 now.

abcdefg30:
Needs a rebase.

reaperrr:
Before I rebase this, any other opinions on the angle and speed field merging?

I see @GraionDilach's point, so I could live with dropping the commits in question.

pchote:
I agree with @GraionDilach, so lets drop those commits.

reaperrr:
Updated. 
Removed the commits in question and extended lint check to also emit errors when minimum launch angle exceeds maximum launch angle.

reaperrr:
Updated.

Removed some syncing from `Bullet`, as it seemed excessive compared to `Missile`.

pchote (commented):
Looks sensible overall, but I haven't looked in full detail or tested, so :+0.5: for now.

OpenRA.Mods.Common/Lint/CheckAngle.cs
```diff
@@ -0,0 +1,67 @@
...
+namespace OpenRA.Mods.Common.Lint
+{
+	class CheckAngle : ILintRulesPass
+	{
+		public void Run(Action<string> emitError, Action<string> emitWarning, Ruleset rules)
+		{
+			foreach (var weaponInfo in rules.Weapons)
+			{
+				var missile = weaponInfo.Value.Projectile as MissileInfo;
+				if (missile != null)
```
pchote: Can you please pull this out into a `void CheckLaunchAngles(WAngle min, WAngle max)` to avoid duplication between these cases?

pchote (dismissed):
Looks good after this fix

reaperrr:
Updated the lint check.

Also made it slightly smarter, `maxAngle` is now only tested if it's not identical to `minAngle` to avoid duplicate errors when that single/identical angle is invalid.

abcdefg30:

pchote (approved):

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/7f462506ec75487de697078778867d61862f5891

PR #12007, opened 2016-09-09
Allow MixFile to support all stream types

### LINKED ISSUE #11481: Mix files can't be loaded from inside oramods
https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/FileSystem/MixFile.cs#L195 assumes/requires that the mix file lives on disk, and crashes when it discovers that it was loaded from a `MemoryStream` instead of a `FileStream`.

RoosterDragon:
In #9627, mix files always came from disk and as an optimization we avoided copying each portion we cared about into memory as this created redundant copies.

Mix files can now come from in memory streams with the changes for mod support.

Therefore, we bring back the copying mechanism so we can handle any type of stream, but still keep the optimization for file streams when possible.

Fixes #11481.

(Disclaimer: Untested)

### DIFF
--- a/OpenRA.Game/FileSystem/MixFile.cs
+++ b/OpenRA.Game/FileSystem/MixFile.cs
@@ -191,9 +191,22 @@ namespace OpenRA.FileSystem
 		public Stream GetContent(PackageEntry entry)
 		{
 			Stream parentStream;
-			var offset = dataStart + entry.Offset + SegmentStream.GetOverallNestedOffset(s, out parentStream);
-			var path = ((FileStream)parentStream).Name;
-			return new SegmentStream(File.OpenRead(path), offset, entry.Length);
+			var baseOffset = dataStart + entry.Offset;
+			var nestedOffset = baseOffset + SegmentStream.GetOverallNestedOffset(s, out parentStream);
+
+			// Special case FileStream - instead of creating an in-memory copy,
+			// just reference the portion of the on-disk file that we need to save memory.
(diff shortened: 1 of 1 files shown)


pchote (approved):
Works as advertised :+1:

OpenRA.Game/FileSystem/MixFile.cs
```diff
@@ -191,9 +191,22 @@ static uint[] ReadBlocks(Stream s, long offset, int count)
...
+			// Special case FileStream - instead of creating an in-memory copy,
+			// just reference the portion of the on-disk file that we need to save memory.
+			// We use GetType instead of 'is' here since we can't handle any derived classes of FileStream.
+			if (parentStream.GetType() == typeof(FileStream))
+			{
+				var path = ((FileStream)parentStream).Name;
+				return new SegmentStream(File.OpenRead(path), nestedOffset, entry.Length);
+			}
+
+			// For all other streams, create an copy in memory.
```
obrakmann: Minor nit: "a copy"

obrakmann:
Looks otherwise ok, :+1:

OpenRA.Game/FileSystem/MixFile.cs
RoosterDragon: Fixed.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/d65d2cf0ca6c13e153d9524ed2c0dee8984a3358)

PR #12011, opened 2016-09-10
Start my glorious mod

Flamewh33l:


Phrohdoh:
This is going to have to live in its own repo.

PR #12015, opened 2016-09-11
Improve some exception types

RoosterDragon:
Throw more specific subclasses of `Exception`.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Multipliers/UpgradableMultiplierTrait.cs
+++ b/OpenRA.Mods.Common/Traits/Multipliers/UpgradableMultiplierTrait.cs
@@ -45,7 +45,7 @@ namespace OpenRA.Mods.Common.Traits
 		protected UpgradeMultiplierTrait(UpgradeMultiplierTraitInfo info, string modifierType, string actorType)
 		{
 			if (info.Modifier.Length == 0)
-				throw new Exception("No modifiers in " + modifierType + " for " + actorType);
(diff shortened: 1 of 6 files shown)


Phrohdoh:
I don't see why not. 👍

OpenRA.Mods.Common/Traits/Multipliers/UpgradableMultiplierTrait.cs
```diff
@@ -45,7 +45,7 @@ public abstract class UpgradeMultiplierTrait : IUpgradable, IDisabledTrait, ISyn
 		protected UpgradeMultiplierTrait(UpgradeMultiplierTraitInfo info, string modifierType, string actorType)
 		{
 			if (info.Modifier.Length == 0)
-				throw new Exception("No modifiers in " + modifierType + " for " + actorType);
+				throw new ArgumentException("No modifiers in " + modifierType + " for " + actorType);
```
pchote: Missing the argument name here?

RoosterDragon: Any of the three arguments could need changing here if you see this exception - I usually only specify an argument if it is the only one that could affect the check.

pchote: Ah right, I see.

pchote (approved):
These are all obviously more correct, so :+1:

PR #12026, opened 2016-09-14
Fix kicking players joining servers hosting a map with non-default rules

### LINKED ISSUE #11986: Players seem to have trouble joining servers where the current map is a mp mission

obrakmann:
Specifically, rules that remove traits which implement ILobbyOptions

Fixes #11986.

### DIFF
--- a/OpenRA.Mods.Common/ServerTraits/LobbySettingsNotification.cs
+++ b/OpenRA.Mods.Common/ServerTraits/LobbySettingsNotification.cs
@@ -35,8 +35,10 @@ namespace OpenRA.Mods.Common.Server
 			foreach (var kv in server.LobbyInfo.GlobalSettings.LobbyOptions)
 			{
 				Session.LobbyOptionState def;
+				LobbyOption option;
 				if (!defaults.LobbyOptions.TryGetValue(kv.Key, out def) || kv.Value.Value != def.Value)
-					server.SendOrderTo(conn, "Message", options[kv.Key].Name + ": " + kv.Value.Value);
+					if (options.TryGetValue(kv.Key, out option))
+						server.SendOrderTo(conn, "Message", option.Name + ": " + kv.Value.Value);
 			}
 		}
 	}


OpenRA.Mods.Common/ServerTraits/LobbySettingsNotification.cs
```diff
@@ -35,7 +35,7 @@ public void ClientJoined(OpenRA.Server.Server server, Connection conn)
 			foreach (var kv in server.LobbyInfo.GlobalSettings.LobbyOptions)
 			{
 				Session.LobbyOptionState def;
-				if (!defaults.LobbyOptions.TryGetValue(kv.Key, out def) || kv.Value.Value != def.Value)
+				if ((!defaults.LobbyOptions.TryGetValue(kv.Key, out def) || kv.Value.Value != def.Value) && options.ContainsKey(kv.Key))
```
pchote: Could possibly use another TryGetValue here?

obrakmann: I don't see it. What do you mean?

pchote: Something like
```csharp
foreach (var kv in server.LobbyInfo.GlobalSettings.LobbyOptions)
{
	Session.LobbyOptionState def;
	LobbyOption option;
	if ((!defaults.LobbyOptions.TryGetValue(kv.Key, out def) || kv.Value.Value != def.Value) && options.TryGetValue(kv.Key, out option))
		server.SendOrderTo(conn, "Message", option.Name + ": " + kv.Value.Value);
}
```

pchote: ... although that's a bit of a mouthful, so it might be worth splitting out an inner `if` for readability if we go this way.

obrakmann:
Updated

pchote (approved):
Fixes the problem, and code looks good :+1:

RoosterDragon (approved):

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/83db5dc564138510cb9457f8a529911d684a75c6)

Issue #12022, opened 2016-09-13
Utility --check-yaml fails in Bleed for any Mod


ihptru:
Example:
```
mono --debug OpenRA.Utility.exe ts --check-yaml /path/to/file.oramap
 
OpenRA.Utility(1,1): Error: Failed with exception: System.IO.FileNotFoundException: Unable to find parent package for /path/to/file.oramap
  at OpenRA.FileSystem.ZipFile..ctor (IReadOnlyFileSystem context, System.String filename) [0x0008a] in /home/servers/servers/tmp/bleed/OpenRA.Game/FileSystem/ZipFile.cs:65 
  at OpenRA.FileSystem.FileSystem.OpenPackage (System.String filename) [0x00050] in /home/servers/servers/tmp/bleed/OpenRA.Game/FileSystem/FileSystem.cs:52 
  at OpenRA.Mods.Common.UtilityCommands.CheckYaml.OpenRA.IUtilityCommand.Run (OpenRA.Utility utility, System.String[] args) [0x00000] in <filename unknown>:0

```

Example of uploaded maps:
http://resource.openra.net/maps/16189/
http://resource.openra.net/maps/15973/

abcdefg30:
I could not reproduce this on Windows.

pchote:
Reproduced on OSX.  It only happens for oramaps:

```
paul@Mjolnir:OpenRA+master$ mono --debug OpenRA.Utility.exe ra --check-yaml /Users/paul/src/OpenRA/mods/ra/maps/sunstroke.oramap 
OpenRA.Utility(1,1): Error: Failed with exception: System.IO.FileNotFoundException: Unable to find parent package for /Users/paul/src/OpenRA/mods/ra/maps/sunstroke.oramap
  at OpenRA.FileSystem.ZipFile..ctor (IReadOnlyFileSystem context, System.String filename) [0x0008a] in /Users/paul/src/OpenRA/OpenRA.Game//FileSystem/ZipFile.cs:65 
  at OpenRA.FileSystem.FileSystem.OpenPackage (System.String filename) [0x00050] in /Users/paul/src/OpenRA/OpenRA.Game//FileSystem/FileSystem.cs:52 
  at OpenRA.Mods.Common.UtilityCommands.CheckYaml.OpenRA.IUtilityCommand.Run (OpenRA.Utility utility, System.String[] args) <0x2d2d000 + 0x00867> in <filename unknown>:0 
[...]

PR #12027, opened 2016-09-14
Fix crash when launching from a different directory than the game binary.

pchote:
This regression will have been introduced during the FileSystem/oramod refactoring.

This also includes a few other minor fixes to Platform.cs

### DIFF
--- a/OpenRA.Game/Platform.cs
+++ b/OpenRA.Game/Platform.cs
@@ -102,14 +102,14 @@ namespace OpenRA
 			path = path.TrimEnd(new char[] { ' ', '\t' });
 
 			// paths starting with ^ are relative to the support dir
-			if (path.StartsWith("^"))
+			if (path.StartsWith("^", StringComparison.Ordinal))
 				path = SupportDir + path.Substring(1);
 
 			// paths starting with . are relative to the game dir
 			if (path == ".")
 				return GameDir;
 
-			if (path.StartsWith("./") || path.StartsWith(".\\"))
+			if (path.StartsWith("./", StringComparison.Ordinal) || path.StartsWith(".\\", StringComparison.Ordinal))
 				path = GameDir + path.Substring(2);
 
 			return path;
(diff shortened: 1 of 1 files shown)


OpenRA.Game/Platform.cs
```diff
@@ -102,14 +102,14 @@ public static string ResolvePath(string path)
 			path = path.TrimEnd(new char[] { ' ', '\t' });
 
 			// paths starting with ^ are relative to the support dir
-			if (path.StartsWith("^"))
+			if (path.StartsWith("^", StringComparison.InvariantCulture))
```
obrakmann: I just read up on what `StringComparison.InvariantCulture` does, and I think an ordinal comparison would be more appropriate here. Why did you go with this instead?

pchote: I went with this only because it's what we use everywhere else.  I didn't put any thought into the different options or whether the other places were doing it correctly.

pchote: I agree that ordinal is a better fit, and will change that here.

obrakmann:
Doesn't seem to help:

```
$ git rev-parse HEAD
a1307e646ef52e1cae3d26097d1259c96c1bf16a
$ make
make: Nothing to be done for `default'.
$ ./OpenRA.Utility.exe ra --check-yaml mods/ra/maps/allies-01/
Testing map: 01:   In the thick of it
$ ./OpenRA.Utility.exe ra --check-yaml mods/ra/maps/forest-path.oramap 
OpenRA.Utility(1,1): Error: Failed with exception: System.IO.FileNotFoundException: Unable to find parent package for mods/ra/maps/forest-path.oramap
  at OpenRA.FileSystem.ZipFile..ctor (IReadOnlyFileSystem context, System.String filename) <0x407a1760 + 0x00327> in <filename unknown>:0 
  at OpenRA.FileSystem.FileSystem.OpenPackage (System.String filename) <0x40779000 + 0x00147> in <filename unknown>:0 
  at OpenRA.Mods.Common.UtilityCommands.CheckYaml.OpenRA.IUtilityCommand.Run (OpenRA.Utility utility, System.String[] args) <0x407a0590 + 0x0095b> in <filename unknown>:0 
```

pchote:
Erm, this isn't supposed to fix #12022.  Someone else must have edited that in.

pchote:
Fixed the `StringComparison`.

obrakmann (approved):
:+1:

obrakmann (changes requested):
just testing

obrakmann (approved):

reaperrr:
Code looks good to me, but I'm not sure how to test this.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/cafbf01aee0cb5af5b9f8c3a6dfc66ceaba09f5c...3586f3829e06186b970f329cb0e3f52d5ffb41ae)

PR #12036, opened 2016-09-16
Fix not being able to close the chat window in replays

### LINKED ISSUE #12031: No longer possible to close chat dialog in replays with hotkeys
#11957 has the side-effect that you can no longer use enter to close the expanded chat overlay.  This is a minor but annoying regression.  We could assign enter as the cancel hotkey in that situation.

obrakmann:
Fixes #12031.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameChatLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameChatLogic.cs
@@ -131,6 +131,21 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 
 			CloseChat();
 			chatText.IsDisabled = () => world.IsReplay;
+
+			var keyListener = chatChrome.Get<LogicKeyListenerWidget>("KEY_LISTENER");
+			keyListener.OnKeyPress = e =>
+			{
+				if (e.Event == KeyInputEvent.Up || !chatText.IsDisabled())
+					return false;
+
+				if ((e.Key == Keycode.RETURN || e.Key == Keycode.KP_ENTER || e.Key == Keycode.ESCAPE) && e.Modifiers == Modifiers.None)
+				{
+					CloseChat();
+					return true;
+				}
+
+				return false;
+			};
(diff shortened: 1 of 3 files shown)


pchote:
hmm.... I guess both of those limitations come from the way our Button widget keypress handling works.  We could solve these by using a `LogicKeyListenerWidget` instead of the button.

obrakmann:
That works. Updated.

OpenRA.Mods.Common/Widgets/Logic/Ingame/IngameChatLogic.cs
```diff
@@ -131,6 +131,21 @@ public IngameChatLogic(Widget widget, OrderManager orderManager, World world, Mo
...
 			CloseChat();
 			chatText.IsDisabled = () => world.IsReplay;
+
+			var keyListener = chatChrome.Get<LogicKeyListenerWidget>("KEY_LISTENER");
+			keyListener.OnKeyPress = e =>
+			{
+				if (e.Event == KeyInputEvent.Up || !chatText.IsDisabled())
+					return false;
+
+				if ((e.Key == Keycode.RETURN || e.Key == Keycode.KP_ENTER) && e.Modifiers == Modifiers.None)
```
pchote: I just noticed that the regular chat input also binds escape to close.  Please add ` || e.Key == Keycode.ESCAPE` here, and this will be perfect.

pchote (changes requested):
:+1: after that one fix.

pchote (approved):

pchote (changes requested):
mis-click...

obrakmann:
Added ESC key.

pchote (approved):

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/afab2dadb3e38be9bcab5764266eec174077f03e...6b6eb3b973c57e5f42ce890a6625b9a72cb64906)

Issue #12013, opened 2016-09-11
Add wiki entry regarding balancing feature-complete mods.


SoScared:
https://github.com/OpenRA/OpenRA/issues/11995#issuecomment-246181198

Circumventing the idea of balance teams for the moment, one specific solution involves adding a wiki section on game balance PRs that presents some guidelines and/or prerequisites for a solid balance PR. 

Note that this wouldn't necessary _exclude_ balance PR's (including fixes affecting balance) that don't follow these guidelines, as done in the past, but rather include those that are prepared below the radar of frequent GitHub contributors. Also since different mods are in different development stages this entry could be added onto feature-complete mods:
 https://github.com/OpenRA/OpenRA/wiki/Development-Goals#tiberian-dawn-and-red-alert 

The wiki entry could allow PR authors, contributors and developers to be better equipped taking on balance PRs that's based more on activity outside the sphere of GitHub and Sleipnir's forum.

Guidelines/prerequisite standards ought to include:

1. Sourcing relevant opinions (fully excluding anonymous polls) and/or discussions.
2. Sourcing playtesting and evaluation, all done by a recognized group of players.
3. All stages sourced must have been open to the public and fully transparent during the process.

Contributors could then evaluate certain balance PRs on the quality of the authors sourced homework, thus forcing discussions topics on to elements not covered directly with the prerequisites. This would also unload a chunk of work upon the PR author and streamline the process in such a way that contributors/developers dont have to spend extra time researching claims presented in the PR.

In turn this could encourage the community to involve itself more in the development process, especially testing out new ideas and creative solutions knowing that if it proves to add value to the game it has a chance of making its way to a fair evaluation.

Issue #11995, opened 2016-09-07
[RA] Review balance change methodology and add more balance changes to the playtest.


SoScared:
Following the Mammoth Tank buff suggestion in #11731 I'd planned to bring along a series of balance PR's but was stalled due to the miscalculation on my part on how balance PR's was evaluated on GitHub. I honestly believed that this particular PR would be merged within hours due to the initial comment pointing to balance discussions among many of OpenRA's most invested players, the series of recorded map playtests and widely distributed map downloads (thousands!) over the course of 4 months with the particular balance change involved. Naturally I didn't expect everyone to agree on the PR itself but the sheer volume of support behind the balance change (compared to anything else noted in the changelog history) led me to believe that regardless of whatever gobbledygook it would meet in the comments in terms of impact on gameplay and balance it would be merged into the bleed like a magnet.

First, just to put this issue dead once and for all, despite mentioned in #11731 **5 times**, there's little-to-nothing in terms of balance changes in this playtest for the RA mod. The artillery nerf and late tech air buff effects has to some degree already been mapped out, demonstrated for months with the map playtests mentioned above including Frame Limiters own extensive map playtesting. With the exception of faster aircraft rearmament there's little else of effect on gameplay we should expect from the couple of weeks that the official playtest will be running. Ore harvester self-healing is a minor tweak and faction neutral. Transport accessibility to beach tiles has been improved, Nuke Truck damage vs air nerfed. GPS down with Tech Center and/or Radar Dome powerdown. Minor tweaks - improves gameplay with specific scenarios, not overall gameplay. 

Unfortunately I was a bit preoccupied and came up short in the comment section of the Mammoth PR which was the reason why I put up the forum poll thread, knowing it was gonna be a landslide. Personally I'm not a big fan of forum polls when it comes to balance discussions but with the PR being unreasonably sandbagged I just didn't care at that point.

[...]

PR #12040, opened 2016-09-16
Add RA balance changes to the next playtest/release.

SoScared:
Partially following #12013 and in spirit of #11995 I'm probing a couple of balance changes aimed for the next playtest or directly to the next release. 

Paraphrasing selection of the list provided in #11995:

**Mammoth Tank:**

Speed 50, up from 42

As pointed to in the forum discussion thread (_not_ the recent poll thread) linked in the PR the speed buff is exactly what is expressed as the most crucial element as to making this unit useful for whatever purpose you want it assigned to. 

Initially the speed boost will reward casual players looking to invest in Mammoth Tanks for its iconic value. Long term this represents a foundation of which additional changes to the Mammoth Tank can be added, given that the current speed **fundamentally cripples** the tank both as an "utility" unit as well as a front line army unit.

**Camo. Pillbox:** 

Health: 450HP down from 600HP

The camo pillbox having cloak and at the same giving the same health/price value as the regular pillbox makes it too beefy and at the same time makes the regular pillbox more or less dormant. This is a very old observation made by a great majority of the community players. The camo. pillbox was super close to being merged in the RA mod as a french exclusive structure #11431. The idea behind it was to further distinguish France as a desceptive faction but also for the regular pillbox to make its way back into the game as a standard defensive structure. As per map playtesting the 450HP c.pillbox accomplished the latter and overall made competetive games a better experience.

**AA Missile Sub:** (**UPDATE**: Removed AAMissileSub commit from the PR.)

Cost: $2000, down from $2400
Damage: 25, down from 30
Added air as a valid target
Visual: Adjusted missile speed and height.

This is an old one but it's really bad that this hasn't made it's way into the game yet. Hamb first exposed this on his twitch a year ago and pretty much got a round of applause. This balance change more or less drowned in the balance discussions raging on before and after the 20151224 holiday release and pretty much remained dormant as focus went more over the fast evolving land battle meta (and naval maps slipping off the radar). However it was brought back in map playtesting from May and despite having limited exposure due to being a Soviet late tech naval unit it did deliver as prmoised - picking off lazy Hinds and Longbows over or close to water and damaging clusters of idling YAKs with the sudden burst of its submissiles. Plays well and fun as hell. The damage and cost was somewhat reduced to gain the unit a little more accessibility and the missile speed/trajectory altered slightly in order for the AA trajectory to look proper vs air.

Closes #11731

### DIFF
--- a/mods/ra/rules/structures.yaml
+++ b/mods/ra/rules/structures.yaml
@@ -583,7 +583,7 @@ HBOX:
 	Valued:
 		Cost: 600
 	Health:
-		HP: 600
+		HP: 450
 	Armor:
 		Type: Heavy
 	RevealsShroud:
--- a/mods/ra/rules/vehicles.yaml
+++ b/mods/ra/rules/vehicles.yaml
@@ -164,7 +164,7 @@ V2RL:
 	Armor:
 		Type: Heavy
 	Mobile:
-		Speed: 42
+		Speed: 50
 		Crushes: wall, mine, crate, infantry, heavywall
 	RevealsShroud:
 		Range: 6c0


SoScared:
@kyrylo @ChristianDK @Micr0Bit @ghost  @anjew175 @Murto @abcdefg30 

Given your previous involvement on related issues and/or participating competetively I'd love to hear your input.

The predecessing balance issue inadvertently became part of a bigger picture concerning RA balancing. If GitHub contributors are inequipped to take on community input properly we may relatively soon end up with a dedicated RA blance team (as recently touch upon by the developers) in order to circumvent blocks on moving the RA gameplay forward. 

If we want to keep the dynamic open between the casual/competetive player and the active GitHub contributors then I believe that this balance PR is a step forward in the right direction and will further encourage the larger community to get involved in discussions, experimenting and playtesting. As I see it there is are **tons** of potential for making the RA mod's casual and competetive gameplay richer and with the limited playerbase we have today, so much to gain by moving the balancing forward.

pchote (approved):
I consider our current (lack of) RA balancing strategy to be untenable, so I completely support @SoScared's efforts even if these changes end up being a bad move and reverted.  They don't look unreasonable, so :+1: from me.

obrakmann:
Is the AA capability on the missile sub meant to be ballistic only? It doesn't track targets at all.

SoScared:
@obrakmann: Yes. Allies retain their AA and air superiority at sea but passive air units are punished for camping Soviet navy and/or idling for vision.

kyrylo:
Since I was pinged here, I am stating my opinion.

* **Mammoth Tank**: massive 👍  This is an obvious and much needed change.
* **Camo Pillbox**: I never had a chance to play those playtest games, but I agree that it needs some change, since normal pillboxes are never used in competitive games. The proposed change looks good to me, too. 👍 
* **AA Missile Sub**: might not be perfect, but it's a lot better than nothing. It's a good start, so 👍  here as well

Great work!

pchote (changes requested):
I just tried the AA missile subs ingame, and IMO the current behaviour (without any tracking) feels like an unintentional bug rather than a feature.

Can we please enable tracking for them, but set the tracking parameters to be relatively bad: the end goal should be the same, but it would feel much better for players.

SoScared:
Visually I imagned the missiles being programmed to explode at the xyz position of the aircraft so I've always been happy with the behaviour.

Tracking AA submissiles would break with the behaviour of the missiles hitting land targets with no tracking. It might be worth looking into other missile types exclusively with AA that can approximate the same behavior with a slow tracking.

pchote:
The main issue is that (I think, but may be wrong) all our AA missiles in all our mods have tracking, so having these ones work differently breaks expectations and so feels buggy.

SoScared:
http://resource.openra.net/maps/14612/

Frame Limiter has been using tracking AA missiles for his playtests, adding a weapon type SubMissileAA. His SubMissileAA is tweaked to function as more ordinary tracking AA but look really nice - it looks like approximating the PR's Submissile with his version should be easy.

SoScared:
Another option: https://www.youtube.com/watch?v=FJABUKtJSXA (unlisted video)

I was playing around with the properties (on top of the work done by @hamb). I lowered the Angle all the way down to 25 and put the range down to 9 cells. This way the missiles draw less of an arc but rather "lifts" up to the target. I set Inaccuracy to 1c938 (from 2c938) so as to guarantee kill targets 5 cells away. Targets beyond 9 cells are pretty meaningless.

This would require adding a secondary weapon. I don't know how to do that (yet) and would need help as to putting a tracking value to the missile.

SoScared:
Looks like an approximation with Frame Limiter's projectile:missile might be the way to go for this release cycle given that projectile:bullet can't properly demonstrate the height of impact when it doesn't hit the target. One can dream tho: #12087

SoScared:
If done correctly, I've removed the AA Missile Sub commit from this PR so as to not hold up the next playtest unneccesary.

SoScared:
Working on a new commit for the AA Missile Sub: https://github.com/OpenRA/OpenRA/commit/86ad7576cefa51e7fcb1042af546b703bed2b002

pchote (approved):

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/552f2dd2372a22bb8b90c8d8bdb08f56c9b57d2b...167c40fcae06ab73742d64d46fe757e9112a72f4)

PR #12041, opened 2016-09-16
Fix the music resetting from time to time

### LINKED ISSUE #12024: Music reset issue on playtest
The music will reset randomly in the game whether you are playing or spectating. It seems to be causing it when the airstrikes or other ion/nukes are being used which has those notifications missing as well.

abcdefg30:
Fixes #12024.

Reproduction case: Start a game with many AIs, heavy support and fastest game speed. Let it run a minute and then restart. Continue restarting until you notice the reset.

We didn't clear the `currentSounds` dictionary before, so the contents were carried over into a new world.

### DIFF
--- a/OpenRA.Game/Sound/Sound.cs
+++ b/OpenRA.Game/Sound/Sound.cs
@@ -76,6 +76,7 @@ namespace OpenRA
 		public void Initialize(ISoundLoader[] loaders, IReadOnlyFileSystem fileSystem)
 		{
 			sounds = new Cache<string, ISoundSource>(s => LoadSound(loaders, fileSystem, s));
+			currentSounds = new Dictionary<uint, ISound>();
 			music = null;
 			currentMusic = null;
 			video = null;


OpenRA.Game/Sound/Sound.cs
```diff
@@ -76,6 +76,7 @@ ISoundSource LoadSound(ISoundLoader[] loaders, IReadOnlyFileSystem fileSystem, s
 		public void Initialize(ISoundLoader[] loaders, IReadOnlyFileSystem fileSystem)
 		{
 			sounds = new Cache<string, ISoundSource>(s => LoadSound(loaders, fileSystem, s));
+			currentSounds =  new Dictionary<uint, ISound>();
```
pchote:
```
OpenRA.Game/Sound/Sound.cs:L79: [SA1025] The code contains multiple spaces in a row. Only one space is needed.
```

abcdefg30: Oops, fixed.

pchote (approved):
This should indeed fix the issue as reported :+1:

obrakmann (approved):
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/0e42f7d29c7af991e4e662ff598ae853c27de41a)

PR #12047, opened 2016-09-18
Added the Tiberium Fona from Firestorm

Mailaender:
which is some kind of vegetation. I would rather call it flora.

image

Depends on https://github.com/OpenRA/OpenRA/pull/11787 which now supports those sprites.


LipkeGu:
Superseed by #12113

PR #12048, opened 2016-09-18
Added the Firestorm tiles

Mailaender:
Depends on #11787. Screenshot from the original Cityscape map which makes use of them:

image


LipkeGu:
Superseeded by #12113

PR #12051, opened 2016-09-18
Fix locking an airfield that is rearming aircraft while being sold

### LINKED ISSUE #12035: Selling an airfield while a yak reloads locks the building

obrakmann:
Fixes #12035.

If this is too big for stable, I can file a separate fix that just checks `building.Locked`.

### DIFF
--- a/mods/ra/rules/structures.yaml
+++ b/mods/ra/rules/structures.yaml
@@ -1116,6 +1116,7 @@ HPAD:
 		ZOffset: 256
 		UpgradeTypes: primary
 		UpgradeMinEnabledLevel: 1
+	WithRearmAnimation:
 
 AFLD:
 	Inherits: ^Building
@@ -1232,6 +1233,7 @@ AFLD:
 		ZOffset: 256
 		UpgradeTypes: primary
 		UpgradeMinEnabledLevel: 1
+	WithRearmAnimation:
 
 POWR:
 	Inherits: ^Building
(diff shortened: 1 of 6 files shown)


mods/ra/rules/structures.yaml
```diff
@@ -1116,6 +1116,7 @@ HPAD:
 		ZOffset: 256
 		UpgradeTypes: primary
 		UpgradeMinEnabledLevel: 1
+	WithRearmAnimation:
```
pchote: We'll need these in TS too, right?

pchote:
The scope of this is fine by me for stable.

mods/ra/rules/structures.yaml
obrakmann: I thought so too, but it seems gahpad and nahpad don't have an `active` sequence.

pchote: They definitely do artwork-wise, but they will be overlays instead of animations, and therefore never hooked up.  Another PR can add `WithRearmOverlay` to fix that.

obrakmann:
Added an upgrade rule, kinda sorta.

pchote (approved):
Code looks good to me, and works as advertised.  No regressions to TD heli repairing.

reaperrr:
> Code looks good to me, and works as advertised. No regressions to TD heli repairing.

Confirmed.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/0e42f7d29c7af991e4e662ff598ae853c27de41a...3087552b3526308cb5bf68c293067ec4916d7d1d)

PR #12066, opened 2016-09-20
server: Fix server getting wedged after map query with bad minimap

chrisforbes:
If the mapserver returned an unusable minimap blob, we'd end up dropping
the resulting exceptions on the floor, and committing a completely broken
MapPreview, which would then blow up the tail end of the map change
process, and all future ValidateClient calls after players join (which
itself was handled by kicking the player and logging some noise of dubious
value).

Adjusts exception handling in a number of places to log the exception
rather than dropping it on the floor, and makes the mapserver response
parsing tolerant of bogus minimap blobs -- in this case, we'd rather just
have no minimap.

Candidate for stable, as it fixes a bug present in the current release and
the current playtest series.

Signed-off-by: Chris Forbes chrisf@ijw.co.nz

### DIFF
--- a/OpenRA.Game/Map/MapCache.cs
+++ b/OpenRA.Game/Map/MapCache.cs
@@ -150,9 +150,10 @@ namespace OpenRA
 					foreach (var kv in yaml)
 						maps[kv.Key].UpdateRemoteSearch(MapStatus.DownloadAvailable, kv.Value, mapDetailsReceived);
 				}
-				catch
+				catch (Exception e)
 				{
 					Log.Write("debug", "Can't parse remote map search data:\n{0}", data);
+					Log.Write("debug", "Exception: {0}", e);
 					if (queryFailed != null)
 						queryFailed();
 				}
--- a/OpenRA.Game/Map/MapPreview.cs
+++ b/OpenRA.Game/Map/MapPreview.cs
@@ -367,7 +367,15 @@ namespace OpenRA
 						spawns[j / 2] = new CPos(r.spawnpoints[j], r.spawnpoints[j + 1]);
 					newData.SpawnPoints = spawns;
 					newData.GridType = r.map_grid_type;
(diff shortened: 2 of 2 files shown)


OpenRA.Game/Map/MapCache.cs
```diff
@@ -150,9 +150,10 @@ public void QueryRemoteMapDetails(IEnumerable<string> uids, Action<MapPreview> m
 					foreach (var kv in yaml)
 						maps[kv.Key].UpdateRemoteSearch(MapStatus.DownloadAvailable, kv.Value, mapDetailsReceived);
 				}
-				catch
+				catch (Exception e)
 				{
 					Log.Write("debug", "Can't parse remote map search data:\n{0}", data);
+					Log.Write("debug", "Exception: {0}", e);
```
pchote: Not a job for this PR, but we should think about redirecting these messages to server.log on dedicated servers instead.

obrakmann:
Fix confirmed, :+1: 
```
Failed to load rules for `Vicemania 1.2` with error :map.yaml:331: There are no elements with key `WithSmoke` to remove
MapCache.LoadAsyncInternal started
Failed parsing mapserver minimap response: System.ArgumentNullException: Value cannot be null.
Parameter name: s
  at System.Convert.FromBase64String (System.String s) <0x7f99ffb60dc0 + 0x0004b> in <filename unknown>:0 
  at OpenRA.MapPreview.UpdateRemoteSearch (MapStatus status, OpenRA.MiniYaml yaml, System.Action`1 parseMetadata) [0x00149] in /home/oliver/devel/openra/git/OpenRA.Game/Map/MapPreview.cs:372 
```

pchote (approved):
Code looks good, and no problems reported after a day running (backported) on the 64h release servers :+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/7c1fac5cb3cb36f5f751e19229b6430d1d36ea6a

PR #12070, opened 2016-09-20
Render WithDecoration above shroud

### LINKED ISSUE #12058: Veterancy icons cannot be seen in the current Tiberian Sun  Bleed

abcdefg30:
Fixes #12058.

It also looked to me like using `SpriteRenderable.None` is better for performance than using `Enumerable.Empty<IRenderable>()`. (Is that correct?)

EDIT: Removed the `SpriteRenderable.None` change.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
@@ -55,7 +55,7 @@ namespace OpenRA.Mods.Common.Traits.Render
 		public override object Create(ActorInitializer init) { return new WithDecoration(init.Self, this); }
 	}
 
-	public class WithDecoration : UpgradableTrait<WithDecorationInfo>, ITick, IRender, IRenderAboveShroudWhenSelected
+	public class WithDecoration : UpgradableTrait<WithDecorationInfo>, ITick, IRenderAboveShroud, IRenderAboveShroudWhenSelected
 	{
 		protected readonly Animation Anim;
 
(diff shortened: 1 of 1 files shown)


RoosterDragon:
Both MS and mono back `Enumerable.Empty` with a static array - so it makes no differences in the end. I would probably leave off that change for simplicity.

Phrohdoh:
I think this should live behind a info field.
Coupled with #12086's `InfiltrateForDecoration` an enemy will always know an infiltrated actor's position.

pchote:
Actually not.  This only affects the render order for the actors that are visible.  It won't change anything for actors that are hidden by the shroud or fog (which aren't rendered at all).

Phrohdoh:
Well alright then, carry on!
This can always be changed later if a consumer actually cares.

abcdefg30:
Removed the `SpriteRenderable.None` change.

RoosterDragon (approved):

reaperrr:
This makes chevrons always visible, even if the actor is behind a cliff/in a tunnel/under a bridge, and not just to the owner, but enemies as well. Do we really want that?

pchote:
> This makes chevrons always visible, even if the actor is behind a cliff/in a tunnel/under a bridge

Yes.  IMO this makes sense from a user interface perspective, and it is the only technical choice we have (either we collide with the terrain, and have it not render, or we render it on top of all terrain).

> and not just to the owner, but enemies as well. Do we really want that?

This sounds like a bug.

reaperrr:
> This sounds like a bug.

Actually I was mistaken, chevrons are not visible to enemies. Watched too many AI-only games as spectator...

> Yes. IMO this makes sense from a user interface perspective, and it is the only technical choice we have (either we collide with the terrain, and have it not render, or we render it on top of all terrain).

Ok. In that case :+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/d770f0c7385364f2b67c75c1b524aa7befbfaca5...6ed72aea81b2a50e3d428e9e7437003b4eb973f1)

pchote:
#11862 and followups were merged after the release-split, so this shouldn't have been merged onto `prep-1609`,

PR #12074, opened 2016-09-20
Change Game.RunTime to a long to prevent overflow

### LINKED ISSUE #12073: Game.Runtime overflows impact dedicated servers.
The Game.Runtime counter overflows every ~25 days.  We use this on dedicated servers for at least the player pinger logic, and from this I can see several active servers that currently have negative ping times.

I suspect this may contribute towards our grey ping server issues, but I haven't yet found a direct link.

If we change Game.Runtime from an int to a long we push this problem from ~25 days out to ~300 million years, which should be more than enough for our purposes.

RoosterDragon:
Fixes #12073.

Just changed Game.RunTime to a long and kept going until the compiler stopped yelling.

### DIFF
--- a/OpenRA.Game/Primitives/ActionQueue.cs
+++ b/OpenRA.Game/Primitives/ActionQueue.cs
@@ -21,7 +21,7 @@ namespace OpenRA.Primitives
 	{
 		readonly List<DelayedAction> actions = new List<DelayedAction>();
 
-		public void Add(Action a, int desiredTime)
+		public void Add(Action a, long desiredTime)
 		{
 			if (a == null)
 				throw new ArgumentNullException("a");
@@ -34,7 +34,7 @@ namespace OpenRA.Primitives
 			}
 		}
 
-		public void PerformActions(int currentTime)
+		public void PerformActions(long currentTime)
 		{
 			DelayedAction[] pendingActions;
 			lock (actions)
@@ -67,10 +67,10 @@ namespace OpenRA.Primitives
 
 	struct DelayedAction : IComparable<DelayedAction>
 	{
-		public readonly int Time;
+		public readonly long Time;
 		public readonly Action Action;
 
-		public DelayedAction(Action action, int time)
+		public DelayedAction(Action action, long time)
 		{
 			Action = action;
 			Time = time;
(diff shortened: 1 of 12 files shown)


OpenRA.Game/Primitives/ActionQueue.cs
```diff
@@ -67,10 +67,10 @@ int Index(DelayedAction action)
 
 	struct DelayedAction : IComparable<DelayedAction>
```
obrakmann: Argh, we have a `struct DelayedAction` and a `class DelayedAction`.

OpenRA.Game/Primitives/ActionQueue.cs
```diff
@@ -21,7 +21,7 @@ public class ActionQueue
 	{
 		readonly List<DelayedAction> actions = new List<DelayedAction>();
 
-		public void Add(Action a, int desiredTime)
+		public void Add(Action a, long desiredTime)
```
obrakmann: We usually try to avoid using floating-point units in things that affect the world. This is quite heavily used for that. Do we feel that this is safe enough (looks to me like only comparisons and an addition (in `World.cs`) are happening to these numbers.

Phrohdoh: `long` isn't floating-point, so I do not understand the worry in this comment.

obrakmann: Right. Sorry, that was stupid. Ignore me.

pchote (approved):
Changes look good, and confirmed that things work properly when `int.MaxValue` is added to `Game.RunTime`. :+1:

OpenRA.Game/Primitives/ActionQueue.cs
RoosterDragon: We could make the struct an inner member of the `ActionQueue` which would reduce it's visibility (as well as possibly renaming it).

obrakmann:
:+1:

obrakmann:
[#12074](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/c4bced524f837a448dfcd1904ecb84b4c248eb53)

PR #12075, opened 2016-09-20
Read data from *all* connections to the server.

### LINKED ISSUE #10094: Unusable "grey ping" servers
These have been increasingly common over the last few releases.  We still don't know what triggers them, but we do know that restarting the server fixes them.

Clients can connect to the server and recieve the initial MOTD and map, but the ping then immediately switches to grey and players are kicked for "connection problems".

When connecting to one of these servers I see the following traffic: http://pastebin.com/raw.php?i=jyWF8cAF

There are two things to notice here:

1.  128 orphaned ClientPing entries in the `SyncInfo` order (only the last one, for my client id 496, is valid).
2.  No `SyncClientPings` or `SyncLobbyClients` orders, which should have followed immediately after this if the server was working properly.

I strongly suspect that this was introduced by #5115, but don't have the time to investigate further.
Hopefully this is a good enough start to identify the trigger and kill this issue for good.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28553066-unusable-grey-ping-servers?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pchote:
Fixes #10094 🎉 

This was a really dumb one: the server would ignore all of the "normal" connections if there were any connections waiting in the handshaking (PreConns) list.  If a connection died before completing the handshake, then that would permanently lock the server for everyone else.

This doesn't try to clean up any dead connections in the PreConns list: they will be thrown out when the game switches to the playing state.

Ask on IRC if you want steps to easily trigger the grey-ping issue.  I'm not going to announce them here because they could be used maliciously.

### DIFF
--- a/OpenRA.Game/Server/Server.cs
+++ b/OpenRA.Game/Server/Server.cs
@@ -185,17 +185,21 @@ namespace OpenRA.Server
 					foreach (var s in checkRead)
 					{
 						if (s == listener.Server)
+						{
 							AcceptConnection();
-						else if (PreConns.Count > 0)
-						{
-							var p = PreConns.SingleOrDefault(c => c.Socket == s);
-							if (p != null) p.ReadData(this);
+							continue;
 						}
-						else if (Conns.Count > 0)
+
+						var preConn = PreConns.SingleOrDefault(c => c.Socket == s);
+						if (preConn != null)
 						{
-							var conn = Conns.SingleOrDefault(c => c.Socket == s);
-							if (conn != null) conn.ReadData(this);
+							preConn.ReadData(this);
+							continue;
 						}
+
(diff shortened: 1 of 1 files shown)


chrisforbes (approved):
That would do it... looks good to me.

obrakmann:
:+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/686cfd575d22b33c24caf66b745559215ce20676)

PR #12076, opened 2016-09-20
Clean up ClientPings when a player drops from the server.

### LINKED ISSUE #12072: Client ping state gets out of sync when a player drops.
We aren't removing the ClientPing from `LobbyInfo.ClientPings` when a player is dropped from the server.  This is especially visible on the "grey ping" servers, where you get sent many hundreds of bogus ping objects (one for each player who has tried to connect to the server) on join.

This doesn't seem to be a direct cause of the grey ping bugs, but it certainly doesn't help matters.

pchote:
Fixes #12072.

The easiest way to test this is to add `System.Console.WriteLine(order.TargetString);` to the top of `UnitOrders.ProcessOrder` and then join/disconnect from a server a few times.  Before the fix you'll see all the orphaned ping objects be sent to your client.  After the fix you won't.

### DIFF
--- a/OpenRA.Game/Server/Server.cs
+++ b/OpenRA.Game/Server/Server.cs
@@ -563,6 +563,7 @@ namespace OpenRA.Server
 				DispatchOrdersToClients(toDrop, 0, new ServerOrder("Disconnected", "").Serialize());
 
 				LobbyInfo.Clients.RemoveAll(c => c.Index == toDrop.PlayerIndex);
+				LobbyInfo.ClientPings.RemoveAll(p => p.Index == toDrop.PlayerIndex);
 
 				// Client was the server admin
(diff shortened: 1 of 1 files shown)


chrisforbes (approved):
Looks good to me.

RoosterDragon:
There are several other places where we call `Remove` or `RemoveAll` on the `Clients` list without also tidying the `ClientPings` list - should these also be changed or do we have a valid reason to ignore those?

pchote:
They should all be for bots, which don't have pings associated with them.

obrakmann:
Those other instances do indeed all handle bots. :+1:

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/61c9c780c057e18c342c724e44597fdc5d1673e0)

PR #11985, opened 2016-09-05
Fixed game restart deleting the syncreport.log

Mailaender:
I have no idea why we didn't do that earlier.

PR #12077, opened 2016-09-21
Include dates in crash/sync log filenames.

pchote:
This builds on #11985/#12056 by giving the sync and exception logs unique filenames.  This makes it easier to identify and grab older logs, with the tradeoff of making the logs dir more cluttered for people like ourselves who regularly crash the game.

This also fixes a regression from #4138, which broke the GameUID written to the syncreports.  If the log renames are controversial then we should at least take this fix.

### DIFF
--- a/OpenRA.Game/Network/SyncReport.cs
+++ b/OpenRA.Game/Network/SyncReport.cs
@@ -18,6 +18,7 @@ using OpenRA.Primitives;
 
 namespace OpenRA.Network
 {
+	using System.Globalization;
 	using NamesValuesPair = Pair<string[], object[]>;
 
 	class SyncReport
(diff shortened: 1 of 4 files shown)


OpenRA.Game/Network/SyncReport.cs
```diff
@@ -92,7 +92,8 @@ void GenerateSyncReport(Report report)
 
 		internal void DumpSyncReport(int frame, IEnumerable<FrameData.ClientOrder> orders)
 		{
-			Log.AddChannel("sync", "syncreport.log");
+			var reportName = "syncreport-" + DateTime.UtcNow.ToString("yyyy-MM-ddTHHmmssZ") + ".log";
```
RoosterDragon: Consider using [`ToString("u")`](https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx#UniversalSortable) which is a touch more human readable, but still sortable.

RoosterDragon: Realised why you did this - colon is not a valid character in filenames.

RoosterDragon: You should specify invariant culture here though.

RoosterDragon (changes requested):

OpenRA.Game/Network/SyncReport.cs
pchote: Done here and everywhere else.

obrakmann (approved):
:+1:

RoosterDragon (approved):

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/7ccd4ce0d9e5bc42433bdf0d5099149c59548ad1)

PR #12100, opened 2016-09-25
Updated Mono to version 4.6.0

Mailaender:
Compiles and runs fine in my tests. See http://www.mono-project.com/docs/about-mono/releases/4.6.0/ for a changelog.


phrohdoh (approved):
Travis passed and I've been running 4.6.0 for a few days now with no issues. 👍

pchote (approved):

PR #12102, opened 2016-09-25
Tweak aircraft resupply behaviour.

pchote:
This implements two changes to hopefully wrap up our aircraft resupply ordeal (for this release cycle, at least):
1.  When the "return to base" hotkey is used all aircraft will return to base, but only those that can be resupplied will land on helipads.  This avoids the annoying dance (particularly in TD, where helipads are only used for repairing) where loaded/repaired helicopters will touch-and-go on the pad, wasting time that could be used resupplying those that actually need it.
2.  @e1d1s1 has been arguing (#11979, #11982) that aircraft continuing their attacks after resupplying is a gameplay regression, and i'm now inclined to agree.  This adds an `AbortOnResupply` flag to disable this feature, and enables it by default.

### DIFF
--- a/OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
+++ b/OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
@@ -20,11 +20,13 @@ namespace OpenRA.Mods.Common.Activities
 	{
 		readonly Aircraft heli;
 		readonly bool alwaysLand;
+		readonly bool abortOnResupply;
 
-		public HeliReturnToBase(Actor self, bool alwaysLand = true)
+		public HeliReturnToBase(Actor self, bool abortOnResupply, bool alwaysLand = true)
 		{
 			heli = self.Trait<Aircraft>();
 			this.alwaysLand = alwaysLand;
+			this.abortOnResupply = abortOnResupply;
 		}
 
 		public Actor ChooseHelipad(Actor self)
@@ -82,7 +84,7 @@ namespace OpenRA.Mods.Common.Activities
 					new Turn(self, initialFacing),
 					new HeliLand(self, false),
 					new ResupplyAircraft(self),
-					!heli.Info.AbortOnResupply ? NextActivity : null);
+					!abortOnResupply ? NextActivity : null);
 			}
 
 			return ActivityUtils.SequenceActivities(
--- a/OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
+++ b/OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
@@ -19,10 +19,12 @@ namespace OpenRA.Mods.Common.Activities
 	public class HeliReturnToBase : Activity
 	{
 		readonly Aircraft heli;
+		readonly bool alwaysLand;
 
-		public HeliReturnToBase(Actor self)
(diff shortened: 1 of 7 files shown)


pchote:
@obrakmann had the following to say about 1. in IRC:

>[20:29:32]	&lt;antares79&gt; I looked at your land-foo branch as well. It doesn't behave the way I feared it would (although I'm not sure why). I'd be ok with it being in the playtest 
> [20:32:51]	&lt;antares79&gt; pchote: I guess the extra ResupplyAircraft invocation can safely go as well. I don't see it being useful

reaperrr:
> This adds an AbortOnResupply flag to disable this feature, and enables it by default.

:+1: I was thinking about filing a ticket for that before you filed this PR.

reaperrr (approved):

OpenRA.Mods.Common/Activities/Air/ReturnToBase.cs
```diff
@@ -128,9 +140,17 @@ public override Activity Tick(Actor self)
...
+			if (ShouldLandAtBuilding(self, dest))
+			{
+				plane.MakeReservation(dest);
+
+				landingProcedures.Add(new Land(self, Target.FromActor(dest)));
+				landingProcedures.Add(new ResupplyAircraft(self));
+			}
+
+			if (!planeInfo.AbortOnResupply)
+				landingProcedures.Add(NextActivity);
```
obrakmann: Please re-write this in such a way that calling this activity from the Lua API won't throw away the queue.

OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
```diff
@@ -68,17 +70,36 @@ public override Activity Tick(Actor self)
...
+			if (ShouldLandAtBuilding(self, dest))
+			{
+				heli.MakeReservation(dest);
+
+				return ActivityUtils.SequenceActivities(
+					new HeliFly(self, Target.FromPos(dest.CenterPosition + offset)),
+					new Turn(self, initialFacing),
+					new HeliLand(self, false),
+					new ResupplyAircraft(self),
+					!heli.Info.AbortOnResupply ? NextActivity : null);
```
obrakmann: same here

obrakmann (changes requested):
The rest looks ok.

pchote:
Done.

obrakmann (approved):

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/c74f14a314b7bf2b1f11b7b5957b5bb7baf39742)

e1d1s1:
Mechanics still cannot heal helos unless they are rearming.

Mousing over a pad and issuing a "land" command (triple down arrow symbol) with a mouse click has the effect of a touch and go. At the very least, this command should lead to the helo landing in order to get mechanic repairs.

"F" command has the effect of all helos clumping to a point near a single helipad, and sometimes jittering/dancing about the point. If you have 3 pads and 3 helos, they should spread out accordingly. If you have more helos than pads, it might make more sense for them to line up at neatly at the rally point the same as when they spawn.

A compromising solution would be to continue your new behavior of rearm - takeoff if the helos return to base of their own accord (out of ammo), but a command of F or mouse click land command should force the helo down on the pad.

PR #12103, opened 2016-09-25
Add tooltips to map resources.

pchote:
This adds a small bit of polish from the original games (but oddly only RA and D2K).  Also fixes `ResourceType` trait field naming and adds [Desc]s, and changes the shroud tooltip to match the original games.

screen shot 2016-09-25 at 16 54 08

### DIFF
--- a/OpenRA.Game/Traits/World/ResourceType.cs
+++ b/OpenRA.Game/Traits/World/ResourceType.cs
@@ -17,11 +17,12 @@ namespace OpenRA.Traits
 	public class ResourceTypeInfo : ITraitInfo
 	{
 		[Desc("Sequence image that holds the different variants.")]
-		public readonly string Sequence = "resources";
+		public readonly string Image = "resources";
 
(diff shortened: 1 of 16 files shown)


Mailaender:
```csharp
		[Desc("Resource identifier used by other traits.")]
		public readonly string Type = null;
```

and 

```csharp
		[Desc("Terrain type used to determine unit movement and minimap colors.")]
		public readonly string TerrainType = null;
```

seem redundant.

Mailaender (approved):
Makes the resource layer a lot easier to understand and works as promised.

Micr0Bit:
:+1:

reaperrr (approved):

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/364c9a81046ac443436bfd807019a6967afae257...382adfd7795f814e0ca636b55475434d70c1bec2)

PR #12110, opened 2016-09-25
Fix LaserZap not being removed if HitAnim is not defined

### LINKED ISSUE #12108: OutOfMemoryException (external mod)
After letting a huge (estimated 200+) amount of actors fire at an empty cell for a while, I stopped them, let them fire again for a while, same for a third time, then selected all units of a specific type (only about 6 of them) and then ordered them to shoot at an empty cell again, which is when the OOM triggered.

It should be noted that after every time I let all 200+ actors fire for a while and then ordered them to stop, ms/Tick did not recover entirely, so the 'idle' tick load was building up over time. For comparison, before the 3rd all-unit firing phase, it was only ~15ms/tick, before the first shooting at that empty cell it was single-digit (don't remember the excact value).

based on bleed from Sept. 18th, right after #12014 was merged.

[exception.log](https://gist.github.com/reaperrr/a15daf3d23ba210b931d65fdd18c556f)
[perf.log](https://gist.github.com/reaperrr/f796279a71e0386fdc5e1ed946679153)

No custom warheads were used. Custom projectile code was used, I'll see if I can repro the OOME with bleed's projectiles and/or a specific projectile type.

custom code can be found here, if necessary: 
https://github.com/reaperrr/OpenRA/tree/mods-mw-code

reaperrr:
This avoids degrading performance over time as well as a possible (albeit unlikely) OutOfMemoryException.

Fixes #12108.

### DIFF
--- a/OpenRA.Mods.Common/Projectiles/LaserZap.cs
+++ b/OpenRA.Mods.Common/Projectiles/LaserZap.cs
@@ -86,6 +86,8 @@ namespace OpenRA.Mods.Common.Projectiles
 			{
 				if (hitanim != null)
 					hitanim.PlayThen(info.HitAnimSequence, () => animationComplete = true);
+				else
+					animationComplete = true;
 
 				args.Weapon.Impact(Target.FromPos(target), args.SourceActor, args.DamageModifiers);
 				doneDamage = true;


Mailaender (approved):

abcdefg30:

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/167c40fcae06ab73742d64d46fe757e9112a72f4...0d8f416ddd5d3a0d2c87bb5621f4fbefb615858b)

abcdefg30:
https://github.com/OpenRA/OpenRA/commit/c7adfd43a03787278c4209fa450727f150e75c0a

PR #12112, opened 2016-09-26
Add support for applying a target flash to actors through lua

abcdefg30:
Allows to code stuff like https://youtu.be/afzfD0AGTPQ?t=5m32s.

### DIFF
--- a/OpenRA.Mods.Common/Scripting/Properties/GeneralProperties.cs
+++ b/OpenRA.Mods.Common/Scripting/Properties/GeneralProperties.cs
@@ -11,6 +11,7 @@
 
 using Eluant;
 using OpenRA.Mods.Common.Activities;
+using OpenRA.Mods.Common.Effects;
 using OpenRA.Mods.Common.Traits;
 using OpenRA.Scripting;
 using OpenRA.Traits;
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Scripting/Properties/GeneralProperties.cs
```diff
@@ -72,6 +73,13 @@ public bool HasProperty(string name)
 		{
 			return Self.HasScriptProperty(name);
 		}
+
+		[Desc("Render a target flash on the actor. If set, 'asPlayer'",
```
Phrohdoh: Are these strings joined together with a space or directly concatenated?
If the latter you'll want to insert a space at the end of the first string or at the beginning of the second.

abcdefg30: They should be two different lines, iirc (as in ", == \n").

OpenRA.Mods.Common/Scripting/Properties/GeneralProperties.cs
```diff
@@ -72,6 +73,13 @@ public bool HasProperty(string name)
...
+			"defines which player palette to use. Duration is in ticks.")]
+		public void FlashTarget(int duration = 4, Player asPlayer = null)
```
pchote: This doesn't accept a target argument\ to flash, so "FlashTarget" isn't a great name.  How about just "Flash"?

pchote (changes requested):

abcdefg30:
Updated.

pchote (approved):
Works as advertised, and seems like a reasonable API addition to make.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/9a804f47725982845d9633d5b6fd92c1335ad5ae...66976c0cd3f116f3ed068b1581edd12f725e9903)

PR #12124, opened 2016-09-28
Add zoom scaling support to WithDecoration

reaperrr:
This allows modders to use it for things like sprite selection boxes, which can be used to improve performance (sprite decorations cost less render performance than system-drawn lines, at least on my system) or use entirely different art style for selection boxes.

Performance test case: https://github.com/reaperrr/OpenRA/tree/sprite-seldeco

In that test case, with ~390 Minigunners selected, ms/Render went down from ~14.5 to ~12.0 on my system when using a sprite selection box instead of the selection box renderable (local debug compile, performance gains in release would likely be lower).

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithDecoration.cs
@@ -52,6 +52,9 @@ namespace OpenRA.Mods.Common.Traits.Render
 		[Desc("Should this be visible only when selected?")]
 		public readonly bool RequiresSelection = false;
 
+		[Desc("Should this be scaled with zoom level?")]
+		public readonly bool ScaleToZoom = false;
+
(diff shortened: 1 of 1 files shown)


pchote:
It would be much better if we completed moving the selection box stuff into mod code, instead of hacking it.  All the hard parts have been done now.

reaperrr:
> It would be much better if we completed moving the selection box stuff into mod code, instead of hacking it. All the hard parts have been done now.

Valid point, of course. 
That being said, I think I just haven't chosen the best example for what this PR does, there might still be modders who would like to scale other decorations like rank chevrons and sprite group digits with zoom level.

pchote:
That will still conflict with #10382, where the selection decorations / other UI elements will be drawn independently from the world.

reaperrr:
Ok, no point in keeping this open then.

PR #12129, opened 2016-09-29
Change Aircraft.IsMoving to be true only when moving horizontally

reaperrr:
This gets rid of the overly simplistic "above ground = moving" assumption aircraft currently make, removing the need to work around it in other places.

In case we need a vertical movement check later, I added `IsMovingVertically` as a precaution.

Prerequisite for #11886.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
+++ b/OpenRA.Mods.Common/Traits/Air/Aircraft.cs
@@ -115,13 +115,13 @@ namespace OpenRA.Mods.Common.Traits
 		public int TurnSpeed { get { return Info.TurnSpeed; } }
 		public Actor ReservedActor { get; private set; }
 		public bool MayYieldReservation { get; private set; }
-		public bool IsMovingVertically { get; private set; }
 
 		bool airborne;
 		bool cruising;
 		bool firstTick = true;
 
 		bool isMoving;
+		bool isMovingVertically;
 		WPos cachedPosition;
 
 		public Aircraft(ActorInitializer init, AircraftInfo info)
@@ -182,7 +182,7 @@ namespace OpenRA.Mods.Common.Traits
 			var oldCachedPosition = cachedPosition;
 			cachedPosition = self.CenterPosition;
 			isMoving = (oldCachedPosition - cachedPosition).HorizontalLengthSquared != 0;
-			IsMovingVertically = (oldCachedPosition - cachedPosition).VerticalLengthSquared != 0;
+			isMovingVertically = (oldCachedPosition - cachedPosition).VerticalLengthSquared != 0;
 
 			Repulse();
 		}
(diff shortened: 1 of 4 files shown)


OpenRA.Mods.Common/Traits/Render/WithMoveAnimation.cs
```diff
@@ -43,8 +43,7 @@ public void Tick(Actor self)
 			var oldCachedPosition = cachedPosition;
 			cachedPosition = self.CenterPosition;
```
pchote: This is now unused and can be removed.

reaperrr:
Updated.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -266,12 +276,12 @@ public Actor GetActorBelow()
 
 		protected void ReserveSpawnBuilding()
 		{
-			/* HACK: not spawning in the air, so try to assoc. with our afld. */
-			var afld = GetActorBelow();
-			if (afld == null)
+			// HACK: Not spawning in the air, so try to associate with our spawner.
```
pchote: Idea for a future PR: check for `ParentActorInit` in the constructor and drop this hack.

GraionDilach:
This will break voxel helicopters - good example is the RA2 Blackhawk.

reaperrr:
> This will break voxel helicopters - good example is the RA2 Blackhawk.

How so?

pchote:
It looks like they are using `WithVoxelWalkerBody` for it, which seems like a horrible kludge.

GraionDilach:
Only `WithVoxelWalkerBody` allows for animated voxels. All the other voxel traits take the first hva frame of the voxel and keep it static.

reaperrr:
> Only `WithVoxelWalkerBody` allows for animated voxels. All the other voxel traits take the first hva frame of the voxel and keep it static.

Then make a copy/derivative of `WithVoxelWalkerBody`, name it `WithVoxelHelicopterBody`, adapt it to check for DAT > 0 instead of movement.

pchote:
Or add a more generic animated voxel trait and use upgrades to turn it on and off.

abcdefg30:
Why aren't you using `IsMovingVertically`?

reaperrr:
> Why aren't you using `IsMovingVertically`?

Assuming that was directed at me, that would make the TD Orca - and presumably TS aircraft voxels later - pitch forward when launching vertically from the helipad. Considering vertical-only movement as 'IMove.IsMoving' would cause more problems than it would solve.

That being said, once this is merged I could add a bool property to #11886 that allows modders to apply upgrades on vertical movement as well, if they want to (then `IsMovingVertically` would no longer be unused, either).

abcdefg30:
> That being said, once this is merged I could add a bool property to #11886 that allows modders to apply upgrades on vertical movement as well, if they want to (then  IsMovingVertically  would no longer be unused, either).

That sounds good to me.

reaperrr:
Updated, moved `IsMovingVertically` down a level to `IMove`, so it can be used by other locomotors and traits (and saves us an `Aircraft` look-up in #11886).

On that note, @pchote do you think splitting that into `IsMovingUp` and `IsMovingDown` could have any merit with regard to movement on slopes (TS had speed modifiers for moving up/down a slope, for example)?

abcdefg30:

pchote (approved):
:+1:

pchote:
This is just an internal refactoring, so doesn't need a changelog entry.

PR #12134, opened 2016-09-30
TS polishing pass: Overhauled missiles.yaml

reaperrr:
Reduces size and duplication by using a `^DefaultMissile`.
Overhauls the launch and flight behavior of missiles to be much closer to the original, while tweaking turn rates and `RangeLimit` to (mostly) maintain their ability to hit moving or very close targets.
- much lower launch speed and maximum speed
- much higher launch angle and cruise altitude, doubled VerticalRateOfTurn to make up for it
- increased RangeLimit, especially for AA-only missiles
- fixed Bike warhead Spread to match other missiles
- fixed SAM site missile explosion to match original
- fixed damage warheads of AA-only missile to be valid against ground as well (to avoid not damaging aircraft that just landed after the missile was launched)

### DIFF
--- a/mods/ts/weapons/missiles.yaml
+++ b/mods/ts/weapons/missiles.yaml
@@ -1,21 +1,25 @@
-Bazooka:
+^DefaultMissile:
 	ReloadDelay: 60
 	Range: 6c0
 	MinRange: 0c512
-	Report: rketinf1.aud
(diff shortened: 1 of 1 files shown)


pchote (approved):
:+1:

abcdefg30:
👍

Issue #12119, opened 2016-09-27
[TS] Selection Boxes of Buildings have wrong offsets


LipkeGu:
Map was converted and is the TS Map "Grand Canyon"

![sell_bounds](https://cloud.githubusercontent.com/assets/2057932/18876718/09291ed6-84cb-11e6-8a25-b0d2d7c62788.png)

abcdefg30:
This will be fixed when we get "3D" health bars (and selection boxes), I guess.

StexoO:
The long "undamaged" building seems to be cut by the ground.

Mailaender:
@reaperrr already tweaked the selection boxes for the base buildings, but we haven't corrected the civilian buildings. The automatically calculated offsets seem to be optimized for the RA/TD era SHP format.

Mailaender:
In fact: we have sprite offsets since https://github.com/OpenRA/OpenRA/pull/3407, but that pull request didn't adapt the surroundings.

https://github.com/OpenRA/OpenRA/pull/13926 fixes this issue.

PR #12135, opened 2016-09-30
TS polishing pass: More depth sorting and starting units

reaperrr:
- Fixes depth sorting of editor actors (spawnpoint, waypoint, camera)
- Fixes depth sorting and offsets of several civilian structures
- Increases number of starting units for medium by 1 and for heavy by 2

### DIFF
--- a/mods/ts/sequences/civilian.yaml
+++ b/mods/ts/sequences/civilian.yaml
@@ -405,7 +405,7 @@ capyr03:
 	Inherits: ^cabase
 	Defaults:
 		Offset: -24, -36, 36
-		DepthSpriteOffset: -24, 0
+		DepthSpriteOffset: -24, 48
 
 ^city4x3:
 	Inherits: ^cabase
@@ -415,6 +415,8 @@ capyr03:
 
 city01:
 	Inherits: ^city4x2
+	Defaults:
+		Offset: -12, -30, 36
 
 city02:
 	Inherits: ^city2x3
@@ -433,6 +435,8 @@ city06:
 
 city07:
 	Inherits: ^city4x2
+	Defaults:
+		Offset: -12, -30, 36
 
 city08:
 	Inherits: ^city2x2
@@ -461,7 +465,7 @@ city15:
 city16:
 	Inherits: ^city4x2
 	Defaults:
-		DepthSpriteOffset: -10, 0
+		Offset: -24, -30, 36
 
 city17:
 	Inherits: ^city4x3
--- a/mods/ts/sequences/civilian.yaml
+++ b/mods/ts/sequences/civilian.yaml
@@ -104,7 +104,7 @@ aban18:
 
 ^bboard1x1:
 	Defaults:
-		Offset: 0, -12
+		Offset: 0, -12, 6
 	idle:
 		ShadowStart: 2
 	damaged-idle:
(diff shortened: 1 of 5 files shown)


Mailaender (approved):

LipkeGu:
closes #12119

MustaphaTR:
I'm not sure if i'm happy with the starting units changes in this PR, i think it should be like following:

GDI Light: e1, e1, e2, e2, jumpjet, smech
GDI Heavy: e1, e1, e1, e2, e2, jumpjet, smech, smech, mmch
Nod Light: e1, e1, e3, e3, cyborg, bggy
Nod Heavy: e1, e1, e1, e3, e3, cyborg, bggy, bike, ttnk

reaperrr:
Updated.

Heavy starting forces now match what @MustaphaTR suggested, Medium now match what he suggested for Light, and on Light the e2 and e3 were replaced with a single Wolverine/Buggy respectively (Grenadiers are superior to Bazookas in direct combat, so these were quite unbalanced before).

mods/ts/sequences/civilian.yaml
```diff
@@ -368,17 +389,22 @@ capyr03:
 ^city2x3:
 	Inherits: ^cabase
 	Defaults:
-		Offset: 12, -30, 30
+		Offset: 12, -30, 36
```
pchote: Why are you setting a larger z offset than y offset in these cases?

reaperrr: The bottom-right edges were cut off (invisible) with the previous values. If there's a better way to fix that, I'm all ears/eyes.

pchote: The depth sprite is probably not centered relative to the artwork, try moving that to the right instead.  If ^ is the case then adding a uniform offset to Z may indeed fix the right hand side, but makes the imbalance on the left even larger, causing potential clipping problems there instead.

reaperrr:
Updated, used `DepthSpriteOffset` to further improve temperate civ building offsets.

pchote (changes requested):
city17 is missing a couple of px off the lamp posts on the side, and gakodk could do with a fix too.  Just in case you're doing this the hard way, you can use the map editor to quickly look for cut off pixels (compare the placement preview with the placed actor).

You should also try using 4d2aebe2b822de5afaa128ace12c91fbd922fab2 (ingame, won't work in the editor) to view the relative positions of the depth mask to the actor.  The bright stripe should be aligned with the front corner of the actor, which you'll see isn't the case for a few of civ structures.

reaperrr:
> city17 is missing a couple of px off the lamp posts on the side

Are you sure? I couldn't see a difference, and the depth sprite position looks right to me.

Anyway, improved the 4x2 buildings and added fixes for the Kodiak, Montauk, Hospital, deployed ICBM and the 5 larger `aban*` snow buildings.

reaperrr:
@pchote I gave city17 another good, hard look, but I really can't spot a single pixel difference between the preview and placed building. If it was really that building, I'll need a screenshot, otherwise I'll assume it was either a different building or I already fixed it by accident somehow.

I did spot some grave z-sorting issues with deployed sensor arrays, though, so I'll add a fix for that.

reaperrr:
Updated, added deployed sensor array z-sorting fix.

pchote:
> or I already fixed it by accident somehow.

I can easily reproduce it on your old branch (which is still available behind 4d2aebe), but not on the current version.

pchote (approved):

pchote (approved):
The depth changes look good enough to merge now. :+1:

A future PR might want to add the iso depth sprite to billboards and the dam.  Some of the civ structures with dirt and junk around them could also use some tweaks, but I expect the behaviour here to match the original game, so I won't ask for that here.

pchote:
I've attached a map with all the editor-placeable actors, which I used with a rebased  4d2aebe for testing.

[depthtest.zip](https://github.com/OpenRA/OpenRA/files/546445/depthtest.zip)

PR #12136, opened 2016-09-30
[Lint] Report when Maps contains invalid custom rules. 

LipkeGu:
This prevents the loading of default rules when loading (custom) Rules fails.

### DIFF
--- a/OpenRA.Mods.Common/Lint/CheckMapMetadata.cs
+++ b/OpenRA.Mods.Common/Lint/CheckMapMetadata.cs
@@ -31,6 +31,9 @@ namespace OpenRA.Mods.Common.Lint
 
 			if (!map.Categories.Any())
 				emitError("Map does not define any categories.");
+
+			if (map.InvalidCustomRules)
+				emitError("Map contains invalid custom rules.");
 		}
 	}
 }


pchote:
This is the wrong approach - we definitely do want the default rules to be loaded.  The errors just needs to be surfaced in the linter.

LipkeGu:
hmm can you give me a testcase where i can reproduce such errors?

LipkeGu:
updated

LipkeGu:
Added a Lobby Warning when selecting Maps with invalid Custom Rules :)

pchote:
The lobby already prevents you from playing games with invalid rules.

LipkeGu:
removed the lobby warning

OpenRA.Mods.Common/Lint/CheckMapMetadata.cs
```diff
@@ -31,6 +31,9 @@ public void Run(Action<string> emitError, Action<string> emitWarning, Map map)
 
 			if (!map.Categories.Any())
 				emitError("Map does not define any categories.");
+
+			if (map.InvalidCustomRules)
+				emitError("Map contains Invalid Custom Rules.");
```
pchote: Invalid, Custom, and Rules shouldn't be capitalized.

LipkeGu:
updated

pchote (approved):

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/2f920d8ba1eb752a5090352853c00c2f4598830c...e1ec15335cc54d7247e465d66347a52089ad18c2)

PR #12137, opened 2016-09-30
Search for the Next free closest Proc

LipkeGu:
This enables Harvesters to use the next closest and free Refinery as like in the originals

### DIFF
--- a/OpenRA.Mods.Common/Traits/Harvester.cs
+++ b/OpenRA.Mods.Common/Traits/Harvester.cs
@@ -165,7 +165,7 @@ namespace OpenRA.Mods.Common.Traits
 		{
 			// Find all refineries and their occupancy count:
 			var refs = self.World.ActorsWithTrait<IAcceptResources>()
-				.Where(r => r.Actor != ignore && r.Actor.Owner == self.Owner && IsAcceptableProcType(r.Actor))
+				.Where(r => r.Actor != ignore && r.Actor.Owner == self.Owner && r.Trait.AllowDocking && IsAcceptableProcType(r.Actor))
 				.Select(r => new {
 					Location = r.Actor.Location + r.Trait.DeliveryOffset,
 					Actor = r.Actor,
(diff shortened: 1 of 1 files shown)


pchote:
The current behavior was put in specifically to prevent harvesters traipsing across the map if their local refinery is blocked.  This looks like it will regress that.

LipkeGu:
> was put in specifically to prevent harvesters traipsing across the map if their local refinery is blocked.

I will take care of that.

OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -183,8 +183,8 @@ public Actor ClosestProc(Actor self, Actor ignore)
 
 					var occupancy = refs[loc].Occupancy;
 
-					// 4 harvesters clogs up the refinery's delivery location:
-					if (occupancy >= 3)
+					// 3 harvesters clogs up the refinery's delivery location:
+					if (occupancy >= 2)
```
GraionDilach: Wouldn't this alone fix https://github.com/OpenRA/OpenRA/issues/3089?

LipkeGu: Could be able ... this Pr was mainly the part of "Use all Refinerys in your base"  or "search another nearest free Refinery in your Base" so the value 2 allows at maximum 2 harvesters waiting at a refinery, as far i found out...

chrisforbes: The refinery /shouldn't/ get jammed with 3. If it does, there's another bug.

LipkeGu: @chrisforbes thats the problem what this PR tries to solve, normaly harvesters should search the next free Refinery not the nearest refinery, (regardless if blocked or not)

LipkeGu:
Superseeded by #12178

Issue #12139, opened 2016-10-01
Helicopters Can Not Be Commanded to Stay on Pad


e1d1s1:
This is a continuation/improvement upon #12102 that allows mechanics to more easily repair helo units.

Issuing an "Enter" command (mouse command to land) for a hind currently results in a "touch-and-go" behavior if it is fully armed. This is counter-intuitive.

Issuing a "Return to Base" command (F hotkey) results in a group of helos clustering around a single helipad.

I argued in #12102 that these situations should result in the helos landing and idling on the pad. I will submit a pull request with these modifications.

obrakmann:
Not a bug, see https://github.com/OpenRA/OpenRA/pull/12140#pullrequestreview-2435023

obrakmann:
Also, helicopters being constantly airborne (as opposed to aircraft) in RA is a design decision. You're suggesting to change that, which is very likely not going to happen.

pchote:
IMO the real bug here is that the return-to-base hotkey doesn't tell damaged aircraft to go to the repair depot.

obrakmann:
We don't have a 'go-repair-thyself' hotkey for regular units, either.

abcdefg30:
Using the repair depot costs money, using mechanics not.

abcdefg30:
You can also already use the repair wrench to order units to the service depot.

pchote:
> We don't have a 'go-repair-thyself' hotkey for regular units, either.

We kind of do, actually: Z (repair cursor) then click on the unit.

e1d1s1:
Do you really want to try to wrench click an airborn unit? Go have another look at the pull request, it works quite well now. Has no effect on your desired behaviors, they will normally still takeoff post-repair or post-reload. Just gives the user the option to override the takeoff behavior by issuing the "Enter" land command. Should keep everyone happy.

matjaeck:
#13975 is fixed and #12140 closed for a while now. Voting for closing as "helicopters being constantly airborne (as opposed to aircraft) in RA is a design decision" and the issue is too specific to serve for the generic question if there should be support for a command that makes helicopters land/take off.

PR #12140, opened 2016-10-01
Heli heal

e1d1s1:
Fix for #12139 

Notes:
I added an "idleOnPad" parameter to ResupplyAircraft and HeliReturnToBase, kind of shadows the changes previously done with "abortOnResupply". By default it is always false, except in the cases of the commands "Enter" and "Return to Base".

Changelog:
"Enter" and "Return to Base" commands result in helipads idling on the pad to allow mechanics access to perform repairs in the case the building is a "RearmBuilding". Otherwise, @pchote  behavior of takeoff after reload/repair is still maintained. This change basically only affects the RA mod and helo interaction with mechanics.


OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
```diff
@@ -48,12 +50,23 @@ public override Activity Tick(Actor self)
...
-					// Helicopters should take off from their helipad immediately after resupplying.
-					// HACK: Append NextActivity to TakeOff to avoid moving to the Rallypoint (if NextActivity is non-null).
-					inner = ActivityUtils.SequenceActivities(
-						aircraft.GetResupplyActivities(host)
-						.Append(new AllowYieldingReservation(self))
-						.Append(new TakeOff(self)).Append(NextActivity).ToArray());
+                    // Helicopters should take off from their helipad immediately after resupplying,
+                    // unless the "Enter" or "Return to Base" command was explictly given by user
+                    // HACK: Append NextActivity to TakeOff to avoid moving to the Rallypoint (if NextActivity is non-null).
+                    if (idleOnPad == false)
```
pchote: This makes helicopters remain on pads after repairing (heli + helipad in TD, heli + service depot in RA), which isn't desirable.

OpenRA.Mods.Common/Traits/Air/Aircraft.cs
```diff
@@ -623,7 +622,7 @@ public void ResolveOrder(Actor self, Order order)
 				if (IsPlane)
 					self.QueueActivity(new ReturnToBase(self, Info.AbortOnResupply, null, false));
 				else
-					self.QueueActivity(new HeliReturnToBase(self, Info.AbortOnResupply, false));
+					self.QueueActivity(new HeliReturnToBase(self, Info.AbortOnResupply, true, true));
```
pchote: This reintroduces the touch-and-go bug in TD.

pchote (changes requested):

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
LipkeGu: ```if (!idleOnPad)```

LipkeGu (changes requested):
The changes in the mod.yaml files in this commit shouldnt be here

e1d1s1:
Ok, changes noted. I am going to clean this up.

obrakmann (changes requested):
👎 

I vote to close this, it's based on a faulty assumption.

The problem of mechanics failing to repair helicopters only happens when the heli is on full ammo. Then they take off immediately, giving the mechanic no time to repair. When the heli needs to rearm, the mechanic can repair just fine, and is done much faster than rearming a full ammo load takes.

The problem here is that the heli is sent to the helipad (the rearm building) when it still has a full ammo load instead of to the repair pad (the designated repair building).

e1d1s1:
Sometimes helos take damage before you get a shot off. How can you use the mechanics in that case? An explicit "Enter" order given to a helo to land on a helipad should result in the helo landing, there is no other logic that can be justified. The gameplay mechanics actually play out quite well now.

GraionDilach:
Come to think of it, I have to agree with @obrakmann. Letting aircrafts idle on pads further than required is bad. This change just clogs up rearming.

A more understandable method could be to have an interface which can trigger the heli to land at specific actions (think Generals helicopters landing when a passenger wants to enter) and mechanic triggering this interface or giving helicopters a deploy command which lets them land and fly.

Also, since the mechanic already uses a nonmelee repair weapon, allowing this weapon to repair aircraft would already solve the problem in a more elegant way.

pchote:
Yeah, lets give mechanics electric pogo sticks :grin:

e1d1s1:
Let me stress again, the latest changes do not affect the behavior @pchote created in which helos takeoff after arming or repairing. You have give the Enter order to get it to stay on the pad. If you give the order on a building that is a RepairBuilding, it will still takeoff after repair.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
e1d1s1: Fixed in the latest code.

OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
```diff
@@ -20,12 +20,14 @@ public class HeliReturnToBase : Activity
...
 		readonly bool alwaysLand;
+		readonly bool idleOnPad;
 		readonly bool abortOnResupply;
 
-		public HeliReturnToBase(Actor self, bool abortOnResupply, bool alwaysLand = true)
+		public HeliReturnToBase(Actor self, bool abortOnResupply, bool alwaysLand = true, bool idleOnPad = false)
 		{
 			heli = self.Trait<Aircraft>();
 			this.alwaysLand = alwaysLand;
+            this.idleOnPad = idleOnPad;
```
GraionDilach: This still have spaces.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
```diff
@@ -19,12 +19,14 @@ namespace OpenRA.Mods.Common.Activities
...
 		readonly Aircraft aircraft;
+        readonly bool idleOnPad;
 		Activity inner;
 
-		public ResupplyAircraft(Actor self)
+		public ResupplyAircraft(Actor self, bool idleOnPad = false)
 		{
 			aircraft = self.Trait<Aircraft>();
-		}
+            this.idleOnPad = idleOnPad;
```
GraionDilach: Spaces.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
GraionDilach: Spaces.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
```diff
@@ -48,12 +50,23 @@ public override Activity Tick(Actor self)
...
 				else
 				{
-					// Helicopters should take off from their helipad immediately after resupplying.
-					// HACK: Append NextActivity to TakeOff to avoid moving to the Rallypoint (if NextActivity is non-null).
-					inner = ActivityUtils.SequenceActivities(
-						aircraft.GetResupplyActivities(host)
-						.Append(new AllowYieldingReservation(self))
-						.Append(new TakeOff(self)).Append(NextActivity).ToArray());
+                    // Helicopters should take off from their helipad immediately after resupplying,
+                    // unless the "Enter" or "Return to Base" command was explictly given by user
```
GraionDilach: And this is where your logic fails: just because the player selected the the dock manually - just because he has more docks and wants to micro where each heli should resupply - doesn't mean the heli should stay there.

GraionDilach:
>Yeah, lets give mechanics electric super pogo sticks :grin:

They are already throwing spanners further than people with Olympic medals. If that's an Acceptable Break From Reality, throwing it that high should be fine as well.

GraionDilach:
There are also more places with spaces btw but I'm not gonna comment on the other lines.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
GraionDilach: Going a step further, this change actually adds the warning that the player accidentally clogs up his/her own air production if (s)he get helicopters docked to all the pads manually. This is really unintuitive as-is.

e1d1s1: @GraionDilach this does not affect your production queue to have a helo on the pad. When a helo is in IDLE mode on the pad, new units just pop up from the same pad. This might not be what you expect, but it probably has something to do with the reservation system.

Your build queue is only blocked while a helo is in transit to the pad, or during the resupply animation, but once its idle, they just pop up. I'd be happy to take a video for you, but you should try it.

In TD, the helos will not stay on the pad regardless of what you do, since they are RepairBuilding types. In TS, they will only stay on the pad if they have taken damage, to allow your mechanics to do their work. I have accounted for the use cases to allow the takeoff behavior when desired.

OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
e1d1s1: Noted, and fixed.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
GraionDilach: I won't bother testing PRs which are plane wrong. If you're right with landed planes not preventing another to spawn, then that's a bug and one more reason why to reject this PR.

pchote: That is a feature that we went out of our way to include: planes idle on the airfield, but will take off if another plane wants to reload or a new plane is ready to be built.

e1d1s1: Well it breaks mechanics and 20 years of precedent. If you are going to make them takeoff by themselves you should have had them queue up and land automatically for reload or repair the way tanks did on the service depot.

LipkeGu:
@e1d1s1 : you can run ```make check``` to check for code style error (formating) also you need to squash the comments to one commit without the unwanted changes (style errors)

OpenRA.Mods.Common/Activities/Air/HeliReturnToBase.cs
```diff
@@ -100,8 +102,9 @@ bool ShouldLandAtBuilding(Actor self, Actor dest)
 			if (heli.Info.RepairBuildings.Contains(dest.Info.Name) && self.GetDamageState() != DamageState.Undamaged)
 				return true;
 
+			// allow mechanics to heal damaged units at their rearming pad when "Return to Base" order is issued
```
abcdefg30: Comments start with a capital letter, please.

OpenRA.Mods.Common/Activities/Air/ResupplyAircraft.cs
```diff
@@ -48,12 +50,23 @@ public override Activity Tick(Actor self)
 				}
 				else
 				{
-					// Helicopters should take off from their helipad immediately after resupplying.
+					// Helicopters should take off from their helipad immediately after resupplying,
+					// unless the "Enter" or "Return to Base" command was explictly given by user
 					// HACK: Append NextActivity to TakeOff to avoid moving to the Rallypoint (if NextActivity is non-null).
+					if (idleOnPad == false)
```
abcdefg30: `if (!idleOnPad)`

abcdefg30:
Two things I noticed: If you give a fully loaded heli the order to return to base (per hotkey, not by clicking on the pad) it flies back and hovers above the pad.
It works as expected if you got multiple fully loaded planes, but not with only one.

(If it is not fully loaded it will land, reload and not take off again as intended.)

reaperrr:
To clean up the PR queue, we're closing PRs with unadressed issues that haven't been updated for several weeks or more.

If you resume work on this, simply re-open it when you update your branch (or file a new PR).

PR #12158, opened 2016-10-03
Add AreaBeam target tracking

### LINKED ISSUE #10181: Homing AreaBeam weapon
Followup feature request to #9357 which introduced an area beam A.K.A. line impact projectile.

The existing area beam weapon has a static target, not re-adjusting its aim if its target is on the move. Instead it just keep firing at the spot the target was when it started shooting.

This is fine for Railgun weapons (Ghost + Mammoth Mk2) but the original TS disruptor tanks tracked their targets with their beam even if the target moved.

The following behavior should apply for a `HomingAreaBeam`:
- Should have a `RangeLimit` similar to missiles, in that it can continue tracking the target until this many ticks have passed since it started being emitted.
- Otherwise function as the existing `AreaBeam`.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28880963-homing-areabeam-weapon?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Adds support for target tracking to `AreaBeam` and enables it on TS Sonic tank weapon.

Closes #10181.

### DIFF
--- a/OpenRA.Mods.Common/Projectiles/AreaBeam.cs
+++ b/OpenRA.Mods.Common/Projectiles/AreaBeam.cs
@@ -54,6 +54,9 @@ namespace OpenRA.Mods.Common.Projectiles
 		[Desc("Can this projectile be blocked when hitting actors with an IBlocksProjectiles trait.")]
 		public readonly bool Blockable = false;
 
+		[Desc("Does the beam follow the target.")]
+		public readonly bool TrackTarget = false;
+
 		[Desc("Extra search radius beyond beam width. Required to ensure affecting actors with large health radius.")]
 		public readonly WDist TargetExtraSearchRadius = new WDist(1536);
 
@@ -93,6 +96,7 @@ namespace OpenRA.Mods.Common.Projectiles
 		int tailTicks;
 		bool isHeadTravelling = true;
 		bool isTailTravelling;
+		bool continueTracking = true;
 
 		bool IsBeamComplete { get { return !isHeadTravelling && headTicks >= length &&
 			!isTailTravelling && tailTicks >= length; } }
@@ -132,8 +136,44 @@ namespace OpenRA.Mods.Common.Projectiles
 			length = Math.Max((target - headPos).Length / speed.Length, 1);
 		}
 
+		void TrackTarget()
+		{
+			if (!continueTracking)
+				return;
+
+			if (args.GuidedTarget.IsValidFor(args.SourceActor))
+			{
+				var guidedTargetPos = args.GuidedTarget.CenterPosition;
+				var targetDistance = new WDist((guidedTargetPos - args.Source).Length);
+
+				// Only continue tracking target if it's within weapon range +
+				// BeyondTargetRange to avoid edge case stuttering (start firing and immediately stop again).
+				if (targetDistance > args.Weapon.Range + info.BeyondTargetRange)
+					StopTargeting();
+				else
+				{
+					target = guidedTargetPos;
+					towardsTargetFacing = (target - args.Source).Yaw.Facing;
+
+					// Update the target position with the range we shoot beyond the target by
+					// I.e. we can deliberately overshoot, so aim for that position
+					var dir = new WVec(0, -1024, 0).Rotate(WRot.FromFacing(towardsTargetFacing));
+					target += dir * info.BeyondTargetRange.Length / 1024;
+				}
+			}
+		}
+
+		void StopTargeting()
+		{
+			continueTracking = false;
+			isTailTravelling = true;
+		}
+
 		public void Tick(World world)
 		{
+			if (info.TrackTarget)
+				TrackTarget();
+
 			if (++headTicks >= length)
 			{
 				headPos = target;
@@ -148,14 +188,14 @@ namespace OpenRA.Mods.Common.Projectiles
 				tailPos = args.Source;
 			}
 
-			// Allow for 1 cell (1024) leniency to avoid edge case stuttering (start firing and immediately stop again).
-			var outOfWeaponRange = args.Weapon.Range.Length + 1024 < (args.PassiveTarget - args.Source).Length;
+			// Allow for leniency to avoid edge case stuttering (start firing and immediately stop again).
+			var outOfWeaponRange = args.Weapon.Range + info.BeyondTargetRange < new WDist((args.PassiveTarget - args.Source).Length);
 
 			// While the head is travelling, the tail must start to follow Duration ticks later.
 			// Alternatively, also stop emitting the beam if source actor dies or is ordered to stop.
 			if ((headTicks >= info.Duration && !isTailTravelling) || args.SourceActor.IsDead ||
 				!actorAttackBase.IsAttacking || outOfWeaponRange)
-				isTailTravelling = true;
+				StopTargeting();
 
 			if (isTailTravelling)
 			{
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -751,14 +751,14 @@ namespace OpenRA.Mods.Common.UtilityCommands
 				}
 
 				// Rename LaserZap BeamDuration to just Duration
-				if (engineVersion < 20161009)
+				if (engineVersion < 20161020)
 				{
 					if (node.Key == "BeamDuration")
 						node.Key = "Duration";
 				}
 
 				// Rename Bullet Angle to LaunchAngle
-				if (engineVersion < 20161016)
+				if (engineVersion < 20161020)
 				{
 					if (node.Key == "Angle")
 						node.Key = "LaunchAngle";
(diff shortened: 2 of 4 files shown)


OpenRA.Mods.Common/Projectiles/AreaBeam.cs
```diff
@@ -132,8 +135,24 @@ public AreaBeam(AreaBeamInfo info, ProjectileArgs args, Color color)
 			length = Math.Max((target - headPos).Length / speed.Length, 1);
 		}
 
+		void UpdateTarget()
+		{
+			if (args.GuidedTarget.IsValidFor(args.SourceActor))
+				target = args.GuidedTarget.CenterPosition;
```
pchote: This ignores the range limit, which looks rather odd when you target a fast moving vehicle that is running away.  What was the behaviour of the original game in that case?

pchote (dismissed):

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
reaperrr: I think in the original it would stop shooting (not suddenly, fading out) and probably stop tracking and target the last actor position that was in range until the tail reached that position.

I'll check later.

reaperrr:
Updated, tracking takes weapon range into consideration now.

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
```diff
@@ -54,6 +54,9 @@ public class AreaBeamInfo : IProjectileInfo
 		[Desc("Can this projectile be blocked when hitting actors with an IBlocksProjectiles trait.")]
 		public readonly bool Blockable = false;
 
+		[Desc("Does the beam follow the target.")]
+		public readonly bool TracksTarget = false;
```
pchote: `TrackTarget` (no `s`) would be more consistent with the other fields: `Render(s)Beam`, `Use(s)PlayerColor`.

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
```diff
@@ -132,8 +136,38 @@ public AreaBeam(AreaBeamInfo info, ProjectileArgs args, Color color)
...
+		void UpdateTarget()
+		{
+			if (args.GuidedTarget.IsValidFor(args.SourceActor))
+			{
+				var guidedTargetPos = args.GuidedTarget.CenterPosition;
+				var targetDistance = (guidedTargetPos - args.Source).Length;
+
+				// Only continue tracking target if it's within weapon range +
+				// 1 cell (1024) leniency to avoid edge case stuttering (start firing and immediately stop again).
+				if (targetDistance > args.Weapon.Range.Length + 1024)
```
pchote: Would `BeyondTargetRange` be more appropriate than hardcoding 1 cell?

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
```diff
@@ -132,8 +136,38 @@ public AreaBeam(AreaBeamInfo info, ProjectileArgs args, Color color)
...
+
+				target = guidedTargetPos;
+			}
+
+			towardsTargetFacing = (target - args.Source).Yaw.Facing;
+
+			// Update the target position with the range we shoot beyond the target by
+			// I.e. we can deliberately overshoot, so aim for that position
+			var dir = new WVec(0, -1024, 0).Rotate(WRot.FromFacing(towardsTargetFacing));
+			target += dir * info.BeyondTargetRange.Length / 1024;
```
pchote: If yes, then we'd want to account for that here so that we don't end up with 2 * `BeyondTargetRange` overshoot.

pchote (changes requested):
Looks fine once we sort out the following:

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
reaperrr: Are you sure? The check above only checks the distance, it doesn't add `BeyondTargetRange` to the target, and we return if the distance exceeds weapon range + `BeyondTargetRange` so we wouldn't even arrive here if the distance was larger.

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
reaperrr: I'm inclined to say 'yes'.

reaperrr:
Updated and rebased.

Additionally, for consistency I renamed `LaserZap`'s `TracksTarget` to `TrackTarget` as well and fixed some projectile upgrade rule dates (they were merged post-prep).

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
pchote: I'm not sure, and you appear to be correct.

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
```diff
@@ -132,8 +136,38 @@ public AreaBeam(AreaBeamInfo info, ProjectileArgs args, Color color)
 			length = Math.Max((target - headPos).Length / speed.Length, 1);
 		}
 
+		void UpdateTarget()
+		{
+			if (args.GuidedTarget.IsValidFor(args.SourceActor))
+			{
+				var guidedTargetPos = args.GuidedTarget.CenterPosition;
+				var targetDistance = (guidedTargetPos - args.Source).Length;
```
pchote: You're comparing this against other `WDist`s, so you can/should avoid the `.Length` in these.

reaperrr:
Updated.

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
```diff
@@ -132,8 +136,38 @@ public AreaBeam(AreaBeamInfo info, ProjectileArgs args, Color color)
...
+			{
+				var guidedTargetPos = args.GuidedTarget.CenterPosition;
+				var targetDistance = new WDist((guidedTargetPos - args.Source).Length);
+
+				// Only continue tracking target if it's within weapon range +
+				// BeyondTargetRange to avoid edge case stuttering (start firing and immediately stop again).
+				if (targetDistance > args.Weapon.Range + info.BeyondTargetRange)
+				{
+					continueTracking = false;
+					isTailTravelling = true;
```
pchote: Can you please pull out a `StopTargeting` method that is used here and when the target dies?  This will cut down on the number of places that directly manipulate these flags.

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
pchote: if `!args.GuidedTarget.IsValidFor(args.SourceActor)` then this delta will be added to the already-offset position from the last tick.  I don't think this is what we want?

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
```diff
@@ -132,8 +136,38 @@ public AreaBeam(AreaBeamInfo info, ProjectileArgs args, Color color)
...
+
+			// Update the target position with the range we shoot beyond the target by
+			// I.e. we can deliberately overshoot, so aim for that position
+			var dir = new WVec(0, -1024, 0).Rotate(WRot.FromFacing(towardsTargetFacing));
+			target += dir * info.BeyondTargetRange.Length / 1024;
+		}
+
 		public void Tick(World world)
 		{
+			if (info.TrackTarget && continueTracking)
```
pchote: You can make this a bit cleaner by moving the `continueTracking` check inside `UpdateTarget` (it can tie in with the fix for the issue just above) and then rename the method to `TrackTarget`.  This then keeps all the logic related to target tracking in one place.

pchote (dismissed):

reaperrr:
Updated.

OpenRA.Mods.Common/Projectiles/AreaBeam.cs
```diff
@@ -149,13 +189,13 @@ public void Tick(World world)
 			}
 
 			// Allow for 1 cell (1024) leniency to avoid edge case stuttering (start firing and immediately stop again).
```
pchote: This is no longer hardcoded to 1 cell.

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -491,19 +491,26 @@ internal static void UpgradeWeaponRules(ModData modData, int engineVersion, ref
...
 				// Rename Bullet Angle to LaunchAngle
-				if (engineVersion < 20161016)
+				if (engineVersion < 20161020)
 				{
 					if (node.Key == "Angle")
 						node.Key = "LaunchAngle";
 				}
 
+				// Rename LaserZap TracksTarget to TrackTarget
+				if (engineVersion < 20161102)
```
pchote: Please update this date.

pchote (dismissed):
Two minor code nits:

pchote (commented):
The beam head disconnecting from the target when the tail disconnects from the source looks weird, but I just confirmed that this is consistent with the original game.

pchote:
:+1: after fixing those minor nits and rebasing.

reaperrr:
Updated and rebased.

pchote (approved):

obrakmann (approved):
👍

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/93b5e94506f44f002e3e6012e775ae0fac5f1d52)

PR #12167, opened 2016-10-03
Marshal results in QueryDevices correctly

### LINKED ISSUE #12150: Settings - Audio - Audio Device showing unknown characters
Seen in Windows binaries of "playtest-20161001", and "playtest-20160904", did not appear on "release-20160508": from main menu open "Settings", click on tab "Audio", and open drop down "Audio Device".

The "Default Output" option is shown as expected, but any other audio device option is showing only a mixture of small rectangles and a few extra signs.
Trying to select any of these "crapped" audio device options, and restarting OpenRA will not change audio output away from default.
I have seen this on all three Mods in both "playtest-" Windows builts. PC is Win8.1-64bit. I did not test on any other platform.

RoosterDragon:
Fixes #12150.

On Windows systems, `Marshal.PtrToStringAuto` marshals to Unicode on "recent" (2000, XP, and later) Windows OSes. This doesn't appear to be the encoding of the text and therefore results in sadness.

Rather than using both `Marshal.PtrToStringX` and `Encoding.Y` and risk having the two be out of sync we can make things simpler by marshalling the buffer as null-separated bytes. Then all we have to do is guess the right encoding. I'm hoping our current choice of the system code page (`Encoding.Default`) is correct. It depends what `alcGetString` is using as the encoding (the documentation does not clarify).

Works on my machine - if anybody has sound devices that have non-ASCII names it would be great if you could verify this!

### DIFF
--- a/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
+++ b/OpenRA.Platforms.Default/OpenAlSoundEngine.cs
@@ -55,20 +55,35 @@ namespace OpenRA.Platforms.Default
 			// Clear error bit
 			AL10.alGetError();
 
-			var devices = new List<string>();
-			var next = ALC10.alcGetString(IntPtr.Zero, type);
-			if (next == IntPtr.Zero || AL10.alGetError() != AL10.AL_NO_ERROR)
+			// Returns a null separated list of strings, terminated by two nulls.
+			var devicesPtr = ALC10.alcGetString(IntPtr.Zero, type);
+			if (devicesPtr == IntPtr.Zero || AL10.alGetError() != AL10.AL_NO_ERROR)
 			{
 				Log.Write("sound", "Failed to query OpenAL device list using {0}", label);
-				return new string[] { };
+				return new string[0];
 			}
 
-			do
+			var devices = new List<string>();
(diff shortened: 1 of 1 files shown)


pchote (commented):
This doesn't break anything under OSX.

abcdefg30:
Resolves the problem. 👍

obrakmann:
@KevinYeti: would you be able test whether this pull request still works correctly for your sound card?

LipkeGu:
I installed Windows 10 (de-DE) on a VM where i could test the Issue ... This PR fixes the Error on the VM :)
:+1:

KevinYeti:
@obrakmann yes,that works fine on my windows10.

abcdefg30:
Thanks for testing this @KevinYeti.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/e4d8e515e0e028a6cd7cad35ece4a6c956796ec9...d9bae541ba70e1d61291ae1ae6a4ef3586a5cfd4)

https://github.com/OpenRA/OpenRA/commit/5b815ca2cff4daa4ea80f349d2151cf6ea2bb390

PR #12169, opened 2016-10-04
Fix non-selectable support actor health bars.

### LINKED ISSUE #12145: Health bars are rendered for paradrop planes
... and probably other actors that have health but not selection decorations.

pchote:
#11862 had an unintended side-effect of enabling health bar rendering for actors that have the `SelectionDecorations` trait but not `Selectable` (previously it required both).

This goes through and explicitly disables the health bars for _most_ of the actors that didn't show them before, either by removing `SelectionDecorations` completely (fields, flags, and that nasty `Colt` actor) or keeping `SelectionDecorations` but disabling health bars (badgers, C&C landing craft, KOTH capture points, barrels, ammo boxes, oil derrick husks).

This keeps the accidentally enabled health bars for the support chinooks and landing craft in missions: these are normal actors that you just can't control, so I figured it made sense to keep the bars for these.

Fixes #12145.

### DIFF
--- a/mods/cnc/maps/cnc64gdi01/rules.yaml
+++ b/mods/cnc/maps/cnc64gdi01/rules.yaml
@@ -41,6 +41,8 @@ OLDLST:
 	Inherits: LST
 	-WithRoof:
 	-Selectable:
+	SelectionDecorations:
+		RenderSelectionBars: False
 	RejectsOrders:
 	Cargo:
 		Types: disabled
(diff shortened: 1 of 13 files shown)


cjshmyr (approved):
Couldn't spot any problems

obrakmann (approved):
👍

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/3dc088330b1f20ddac7c7801ddd2819d05a3b5c9)

PR #12174, opened 2016-10-04
Replaced civillians with Technicians when building sold

### LINKED ISSUE #12122: Technicians
When you sell a building ... you actually dont get a civilian in the original ... but "Technicians"... that use civilian-sprites ... 

technician

according to http://cnc.wikia.com/wiki/Technician_%28Red_Alert%29 :
```
... technicians in Red Alert 1 could do mild damage to units with no armour whatsoever, like the V2 Rocket Launcher ...
```

tysonliddell:
Fixes #12122.

### DIFF
--- a/mods/ra/rules/defaults.yaml
+++ b/mods/ra/rules/defaults.yaml
@@ -466,7 +466,7 @@
 	ExternalCapturable:
 	ExternalCapturableBar:
 	EmitInfantryOnSell:
-		ActorTypes: e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,c1,c2,c3,c4,c5,c6,c7,c8,c9,c10,e6,e6,e6,e6,e6
+		ActorTypes: e1,e1,e1,tecn,tecn
 	MustBeDestroyed:
 		RequiredForShortGame: true
 	GpsDot:
--- a/mods/ra/rules/defaults.yaml
+++ b/mods/ra/rules/defaults.yaml
@@ -477,7 +477,7 @@
 ^ScienceBuilding:
 	Inherits: ^Building
 	EmitInfantryOnSell:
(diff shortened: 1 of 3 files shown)


Micr0Bit:
^ScienceBuilding should also emit "technicians" instead of "civilians" ... 

and we could make this a bit more interesting by having ^CivBuilding emit "civs" ... when they are destroyed ... with "EjectOnDeath:"

so we can actually still see "Civilians" ingame

mods/ra/rules/defaults.yaml
```diff
@@ -466,7 +466,7 @@
 	ExternalCapturable:
 	ExternalCapturableBar:
 	EmitInfantryOnSell:
-		ActorTypes: e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,c1,c2,c3,c4,c5,c6,c7,c8,c9,c10,e6,e6,e6,e6,e6
+		ActorTypes: e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,e1,tecn,tecn,tecn,tecn,tecn,tecn,tecn,tecn,tecn,tecn,e6,e6,e6,e6,e6
```
reaperrr: This list now contains 15x e1, 10x tecn and 5x e6, so you can actually shorten it drastically to `e1,e1,e1,tecn,tecn,e6`.

Although on that note, only construction yards should emit `e6`, so maybe we should remove it here and only add it on FACT.

tysonliddell:
Changes made. I think that the ^CivBuildings emitting civs on destruction should be raised as a separate issue.

reaperrr (approved):

Micr0Bit:
:+1:

abcdefg30:
👍 / ✅

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/5541930fd7a3e78038ba72b97337d6520e27bd8e...04ccbee67d5b40db2d347254d2f50e8ef24a1406)

PR #12175, opened 2016-10-04
Remove duplicated chrome yaml files from d2k

abcdefg30:
`color-picker.yaml` and `ingame.yaml` are the same as the ra ones.


Mailaender (approved):
Yes, indeed these are now completely redundant. The parts where a slightly different layout was required are over as the color preview unit parameter is stored in chrome metrics now and the sidebar can't overlap the perf graph anymore which I think was the reason for a forked ingame.yaml

reaperrr (approved):

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c6f744f2f6ee124e1c951d3da8a4b4eb441ffe41...7b54ecdeb1fad1478108e85588be21b75e32d32e)

PR #12176, opened 2016-10-04
[TS] Polish the Unit Movement

LipkeGu:
This Pr changes the Speed of Units
- Damaged Infantry moves now slower.
- Units gets slower on Sand and Tiberium

### DIFF
--- a/mods/ts/rules/defaults.yaml
+++ b/mods/ts/rules/defaults.yaml
@@ -104,7 +104,7 @@
 		Dimensions: 1,1
 		Footprint: x
 		BuildSounds: place2.aud
-		TerrainTypes: Clear, Road, DirtRoad, Rough
+		TerrainTypes: Clear, Road, Green, Pavement, DirtRoad, Rough
 		Adjacent: 4
 	FrozenUnderFog:
 	SoundOnDamageTransition:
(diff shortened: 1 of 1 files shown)


pchote:
@reaperrr does this look correct to you?

GraionDilach:
>Damaged Infantry moves now slower.

This was introduced in Generals.

>Units gets slower on Sand

The Sand terraintype doesn't exist.

>Units gets slower on [...] Tiberium

They already were and already had the respective TS setup. https://github.com/LipkeGu/TSIniFiles/blob/master/rules.ini#L10401

LipkeGu:
@GraionDilach  As far i know sand does exist in TS and wasn a valid buildTarget for buildings

reaperrr:
@LipkeGu The 'sand' in TS didn't have its own terrain type. It used the 'Rough' terrain type instead. 

And infantry didn't slow down on damage in the original TS, like @GraionDilach already pointed out.

> @reaperrr does this look correct to you?

No. Bleed already uses the correct original values.

GraionDilach:
@LipkeGu Beach exists as a terraintype and not Sand. However, this Beach terraintype was never used either. You could look up the tilesets or the original TMP data - they are indeed using Rough and Clear terrain.

LipkeGu:
These TileTypes are taken from temperate.yaml :/ and they works :/

GraionDilach:
Please reread https://github.com/OpenRA/OpenRA/blob/bleed/mods/ts/tilesets/temperate.yaml#L12315.

In any case, this PR is completely arbitrary and should be closed as it goes against the current TS plan.

PR #12178, opened 2016-10-04
harvester.cs: Refactor ClosestProc logic

LipkeGu:
Allow only 2 Harvesters at a refinery But now All other Harvesters will wait on the Tiberium field. maybe we should give them a Order to move to the nearest free proc and wait there... In TS harvesters was waiting behind the refinery

### DIFF
--- a/OpenRA.Mods.Common/Traits/Harvester.cs
+++ b/OpenRA.Mods.Common/Traits/Harvester.cs
@@ -35,6 +35,18 @@ namespace OpenRA.Mods.Common.Traits
 
 		public readonly int BaleLoadDelay = 4;
 
+		[Desc("How many Harvesters can wait at blocked refineries.")]
+		public readonly int MaxOccupancy = 2;
+
+		[Desc("Color of targetline when harvesters move back to refineries.")]
+		public readonly Color DeliveringColor = Color.Green;
+
+		[Desc("Color of targetline when harvesters move out to harvest.")]
+		public readonly Color HarvestingColor = Color.Red;
+
+		[Desc("Prefer less occupancy, this multiplier is to offset distance cost (Occupancy * OccupancyCostModifier)")]
+		public readonly int OccupancyCostModifier = 12;
+
 		[Desc("How fast it can dump it's carryage.")]
 		public readonly int BaleUnloadDelay = 4;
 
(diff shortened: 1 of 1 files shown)


OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -183,8 +183,8 @@ public Actor ClosestProc(Actor self, Actor ignore)
 
 					var occupancy = refs[loc].Occupancy;
 
-					// 4 harvesters clogs up the refinery's delivery location:
-					if (occupancy >= 3)
+					// 3 harvesters clogs up the refinery's delivery location:
+					if (occupancy >= 2)
```
pchote: This really should be pulled out to a trait parameter, it will be different for different mods.

pchote (changes requested):

OpenRA.Mods.Common/Traits/Harvester.cs
GraionDilach: #3089 will disagree when this goes above than 2. I couldn't find the cause of that issue, but using xx== xxx= xx== foundations on the TS/RA2 refineries couldn't help. Although I guess it's related to the fact that sometimes when two harvesters arrive to the same refinery, one docks in to dump, then the other nudges it aside and the jamming maybe happens when a latecomer forces the already-at-docking position one to nudge away to a position which is already taken by the other latecomer.

LipkeGu: When you have tested it you will note that Harvesters will wait on the Field until the refinery is free for the next Harvester. the changes i have made minutes ago will filter out blocked refineries.

LipkeGu:
updated

LipkeGu:
updated

* Unhardcode occupancy and document it
* Unhardcode Targetline colors for harvesting / delivering
* Unhardcode Distance Cost and document it

* ClosestProc: Select closest free refinery

OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -35,6 +35,18 @@ public class HarvesterInfo : ITraitInfo, Requires<MobileInfo>
 
 		public readonly int BaleLoadDelay = 4;
 
+		[Desc("How many Harvesters can wait at blocked refineries.")]
+		public readonly int MaxOccupancy = 2;
+
+		[Desc("Color of targetline when harvesters move back to refineries.")]
+		public readonly string DeliveringColor = "Gold";
```
pchote: FieldLoader understands `Color`, so there is no need for these to be strings.

OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -35,6 +35,18 @@ public class HarvesterInfo : ITraitInfo, Requires<MobileInfo>
...
+
+		[Desc("Color of targetline when harvesters move out to harvest.")]
+		public readonly string HarvestingColor = "Red";
+
+		[Desc("Prefer less occupancy, this multiplier is to offset distance cost (Occupancy * DistanceCost)")]
+		public readonly int DistanceCost = 12;
```
pchote: This is an *occupancy* cost *modifier*, and not a distance or a raw cost.

OpenRA.Mods.Common/Traits/Harvester.cs
```diff
@@ -161,15 +173,25 @@ bool IsAcceptableProcType(Actor proc)
 				Info.DeliveryBuildings.Contains(proc.Info.Name);
 		}
 
+		internal bool ProcHasSpace(int occupancy)
```
pchote: Why are these `internal`?

pchote (dismissed):

OpenRA.Mods.Common/Traits/Harvester.cs
LipkeGu: I took the Scheme from HackyAI should i make them private?

OpenRA.Mods.Common/Traits/Harvester.cs
LipkeGu: In the Line commit it was named "distance multipler" i will rename it

OpenRA.Mods.Common/Traits/Harvester.cs
LipkeGu: Ok then i will Set mit to Color

LipkeGu:
updated

reaperrr:
@pchote I'm 'dismissing' your review since the PR was updated in the meantime, and it might put off other reviewers thinking there are unadressed requests.

pchote:
Superseded by #12645.

PR #12179, opened 2016-10-05
Log an error when no cells are available to spawn starting units

abcdefg30:
Instead of crashing the game, we now just omit actors we can't spawn.

Easily testable with http://resource.openra.net/maps/16553/.

### DIFF
--- a/OpenRA.Mods.Common/Traits/World/SpawnMPUnits.cs
+++ b/OpenRA.Mods.Common/Traits/World/SpawnMPUnits.cs
@@ -89,7 +89,10 @@ namespace OpenRA.Mods.Common.Traits
 				var mi = w.Map.Rules.Actors[s.ToLowerInvariant()].TraitInfo<MobileInfo>();
 				var validCells = supportSpawnCells.Where(c => mi.CanEnterCell(w, null, c));
 				if (!validCells.Any())
-					throw new InvalidOperationException("No cells available to spawn starting unit {0}".F(s));
+				{
+					Log.Write("debug", "No cells available to spawn starting unit {0} for player {1}".F(s, p));
+					continue;
+				}
 
 				var cell = validCells.Random(w.SharedRandom);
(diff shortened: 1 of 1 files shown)


LipkeGu:
Would be nice when Lint also throws a warning or error ... is this able?

abcdefg30:
> is this able?

Looks like it isn't.

GraionDilach:
I'd still crash if no actor of a player could be placed tbh.

OpenRA.Mods.Common/Traits/World/SpawnMPUnits.cs
```diff
@@ -89,7 +89,10 @@ void SpawnUnitsForPlayer(World w, Player p, CPos sp)
 				var mi = w.Map.Rules.Actors[s.ToLowerInvariant()].TraitInfo<MobileInfo>();
 				var validCells = supportSpawnCells.Where(c => mi.CanEnterCell(w, null, c));
 				if (!validCells.Any())
-					throw new InvalidOperationException("No cells available to spawn starting unit {0}".F(s));
+				{
+					Log.Write("debug", "No cells available to spawn starting unit {0}".F(s));
```
pchote: Can you please add "for player {1}" to the end of this message?

pchote (dismissed):
Looks good, :+1: after one minor addition.

GraionDilach:
Oh, this is only about the support units and not the base one - ignore what I said then. :+1: when pchote's fixup is addressed.

abcdefg30:
Sorry for the delay. Updated.

GraionDilach (approved):
:+1:

obrakmann (approved):
👍

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/27ccc3088cd7346ff7c8d425014e0de61df3fb5b)

PR #12181, opened 2016-10-05
Revert "Update Mono to version 4.6.0"

### LINKED ISSUE #12166: Fatal error after downgrading to "release"
I am using Opensuse linux 42.1 and my openra version is the newest playtest. First I had the previous playtest, then downgraded to release. After first downgrading it I have not been able to launch the game and I am getting a "Openra has encountered a fatal error".

[exception-2016-10-03T192051Z.txt](https://github.com/OpenRA/OpenRA/files/507012/exception-2016-10-03T192051Z.txt)

I tried deleting  ~/.openra/ with all of its contents without any luck.

pchote:
The mono 4.6 compiler breaks compatibility with mono < 4.4 runtimes, and the upstream [bug report](https://bugzilla.xamarin.com/show_bug.cgi?id=36942) has been closed as "RESOLVED ANSWERED" so this appears to be a WONTFIX.

Reverting back to mono 4.4 restores compatibility with older runtimes until we decide to increase our minimum requirements to mono 4.4 / .NET 4.6.

Fixes #12166.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 # see travis-ci.org for details
 
 language: csharp
-mono: 4.6.0
+mono: 4.4.2
 
 # http://docs.travis-ci.com/user/migrating-from-legacy
 sudo: false


pchote:
Looks like this was a false alarm, I had forgotten that we now explicitly pass the sdk flag from travis, instead of having it in the default makefile.  Building on mono 4.6 explicitly specifying the 4.5 sdk works fine for me locally.

PR #12182, opened 2016-10-06
Updated to Mono version 4.6.1

Mailaender:
This appears to be a hotfix release http://www.mono-project.com/docs/about-mono/releases/4.6.1/ although I don't understand the urgency of the problem nor if we are effected by it. I guess not yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 # see travis-ci.org for details
 
 language: csharp
-mono: 4.6.0
+mono: 4.6.1
 
 # http://docs.travis-ci.com/user/migrating-from-legacy
 sudo: false


LipkeGu:
```
Created attachment 17526 [details]
repro project

# Steps to reproduce

Compile attached project.

# Expected behaviour

It compiles

# Actual behaviour

Compiler error: Error CS0029: Cannot implicitly convert type `int[]' to `System.Collections.IEnumerable'

# Supplemental information

Seems to work OK if you change any of the following
* Remove nuget package Rx-Linq
* Change PCL profile (The attached project is profile 78, it compiled for me on 136)
* Change project type to native class library (worked for me in an iOS class library, which was relevant to the project where I discovered the issue)
* Use Visual Studio 2015

Reproduces on 2 Macs in our office.
```

pchote:
:+1:

pchote (approved):

PR #12194, opened 2016-10-09
Uncloak on damage by default

### LINKED ISSUE #12192: RA Phase Transport, TD Stealth tank & D2k Stealth Raider don't uncloak on damage
Aside from `Move`, which is obviously intentional, `Damage` is the only trigger not listed in the internal default of `Cloak.UncloakOn`.

`UncloakType.Damage` was most likely ommitted by accident in
https://github.com/OpenRA/OpenRA/commit/f46185b85f458f44f32787f481983d91575d6808
because unlike the other 5 possible triggers, damage was previously hardcoded to trigger an uncloak and did not have a boolean field.

This also means this is a regression and should probably go into stable, as it affects at least 1 unit in each shipping mod + actors collecting stealth crates in TD.

reaperrr:
Not doing so is a regression compared to pre-20160508 releases and probably unexpected behaviour for the majority of modders.

All actors in the shipping mods which are _not_ supposed to uncloak on damage are already overriding the default at yaml level, so there aren't any regressions to expect on that end.

Fixes #12192.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Cloak.cs
+++ b/OpenRA.Mods.Common/Traits/Cloak.cs
@@ -43,7 +43,7 @@ namespace OpenRA.Mods.Common.Traits
 
 		[Desc("Events leading to the actor getting uncloaked. Possible values are: Attack, Move, Unload, Infiltrate, Demolish, Dock and Damage")]
 		public readonly UncloakType UncloakOn = UncloakType.Attack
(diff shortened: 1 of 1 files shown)


pchote:
From #12192:

> This isn't a regression: RA/TD/D2K have never uncloaked on damage.  The `UncloakType.Damage` flag was added for TS specifically in #9808.

IMO this doesn't belong in stable, but should instead be discussed as a balance change for each of the affected mods for Next+1.

reaperrr:
Yeah, this was a misunderstanding on my end. Closing.

PR #12198, opened 2016-10-09
Fix soviet units not stopping to hunt after being captured in Exodus

### LINKED ISSUE #12183: Hijacked units carry on with computers orders RA missions
Much like captured war factories and barracks in C&C, hijacked tanks etc. carry on with their old orders once taken over by the player.  This means for example in exodus, if I hijack the enemy mammoth tanks they keep driving off to where the computer last sent them.

abcdefg30:
and makes hijackers unbuildable in Evacuation. I made them unbuildable there because we don't allow building mechanics (which are somewhat the counter parts to hijackers) in this mission as well.
Fixes #12183.

This PR also fixes soviet units not attack moving to their rally point in Exodus, and also removes some code duplication (we don't need to check for idling units inside the Tick function if we can just assign them a hunt trigger directly after they're created).

### DIFF
--- a/mods/ra/maps/evacuation/rules.yaml
+++ b/mods/ra/maps/evacuation/rules.yaml
@@ -72,6 +72,10 @@ MECH:
 	Buildable:
 		Prerequisites: ~disabled
 
+HIJACKER:
+	Buildable:
+		Prerequisites: ~disabled
+
 SPEN:
 	Buildable:
 		Prerequisites: ~disabled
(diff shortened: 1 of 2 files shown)


Mailaender (approved):
Code looks good to me.

obrakmann:
> because we don't allow building mechanics

err, mechanics are buildable in Exodus. They're kind of my MO in that mission.

How about we disable both hijackers and mechanics in the campaign-rules.yaml file?

abcdefg30:
> err, mechanics are buildable in Exodus. They're kind of my MO in that mission.

Yes. I was talking about Evacuation there, not Exodus. 😃

obrakmann (approved):
Seems to work fine, 👍

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/12425d1b02b1411a60d81db55db4ab795969a834)

PR #12186, opened 2016-10-07
Overhaul the HackyAI logic

LipkeGu:
This PR overhauls the HackyAI logic in some Cases:
-  Replace `EnoughWaterToBuildNaval()` and `CloseEnoughToWater()` with `ProviderHasEnoughCells<T>()`
- Add `Radar` and `TechCenter` to `BuildingCommonNames`
- Replace `FindEnemyConstructionYards()` with `FindEnemyBuildings()`
- Replace `FindAndDeployBackupMcv()` with `FindAndDeployMcv()` and allow random locations on the map for the deployment when the AI has already a Construction Yard.

deploymcv

ai_search_a_location

test_normal_mcv

PR #12214, opened 2016-10-11
Make sure AI can deploy an MCV when it lacks a base.

### LINKED ISSUE #12159: OpenRA: AI will not deploy new MCV
After killing the MCV of an AI opponent, the AI will make a new MCV (if it has the service depot and war factory intact), but let the MCV idle around without deploying. 

This problem was first introduced in **Playtest 20160904.** 

Something in the codebase change between **release 20160508** and the previously mentioned playtest caused this. As I do not have any programming skills (I am learning some basic stuff in my little spare time) I can not fix it.

Reproducing the bug is easy, just let the AI make a warfactory and service depot and kill of the MCV.

RoosterDragon:
Fixes #12159.

The `RestrictMCVDeploymentFallbackToBase` flag restricts AI to deploying new MCVs near to the existing conyard (and is true by default). However, if there is no conyard then it obviously cannot find any suitable locations!

The fix is to allow the MCV to deploy anywhere if there is no conyard it can be built near to.

### DIFF
--- a/OpenRA.Mods.Common/AI/HackyAI.cs
+++ b/OpenRA.Mods.Common/AI/HackyAI.cs
@@ -964,8 +964,12 @@ namespace OpenRA.Mods.Common.AI
 				if (!mcv.IsIdle)
 					continue;
 
+				// If we lack a base, we need to make sure we don't restrict deployment of the MCV to the base!
+				var restrictToBase =
+					Info.RestrictMCVDeploymentFallbackToBase &&
+					CountBuildingByCommonName(Info.BuildingCommonNames.ConstructionYard, Player) > 0;
 				var factType = mcv.Info.TraitInfo<TransformsInfo>().IntoActor;
-				var desiredLocation = ChooseBuildLocation(factType, Info.RestrictMCVDeploymentFallbackToBase, BuildingType.Building);
+				var desiredLocation = ChooseBuildLocation(factType, restrictToBase, BuildingType.Building);
 				if (desiredLocation == null)
 					continue;
 


obrakmann:
I think this will fix it, but for the wrong reason. I think the real problem is that `RequiresBaseProvider` (and `Adjacency`) is set for `FACT`. Both are checked for in `BuildingInfo.IsCloseEnoughToBase`. By setting the `distanceToBaseIsImportant` argument to false, you're just avoiding this check.

The real fix will probably involve not re-using the `ChooseBuildLocation` method to get the MCV deploy location.

RoosterDragon:
I agree an overhaul of the deployment logic would be worthwhile in the longer term (e.g. #12186 or some other rework). However, I feel this is a sensible short-term way to ensure MCVs can at least deploy without introducing riskier changes which we may wish to avoid for stable.

pchote:
@obrakmann mentioned in IRC that this still fails in many cases because `ChooseBuildLocation` is returning a null location even when `restrictToBase` is false.

obrakmann:
No, I was testing on bleed, not this branch.

obrakmann (approved):
OK, tested this branch now. As expected, the bug that AIs don't deploy their MCVs is fixed, but of course they deploy the new MCV somewhere on the map and abandon their old base.

But yeah, for stable it's probably a better solution than coming up with new untested logic for the deployment, so 👍

reaperrr:
:+1:

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/2fb0b924e47eb17726cbf55c0ed2288647ef532f...b6b9b4cb6cfbb7974635b113821390e2e2028119)

PR #12217, opened 2016-10-12
Fix transport husks revealing more/less than the normal actor

### LINKED ISSUE #12213: [RA & TD] Vision range of Chinook helicopter husks.
The Chinook helicopter used to have a vision range of 12 cells, but this has since been reduced to 10 cells. However its husk still has 12 cells of vision, so it actually has better vision upon getting shot down which doesn't make sense. This should be changed to 10 cells.
In TD the opposite is true, the Chinook has a vision range of 10 cells, while its husk has 8 cells of vision.

[RA]
TRAN:
+RevealsShroud:
++Range: 10c0

TRAN.Husk:
+RevealsShroud:
++Range: 12c0

[TD]
TRAN:
+RevealsShroud:
++Range: 10c0

TRAN.Husk:
+RevealsShroud:
++Range: 8c0

abcdefg30:
Fixes #12213.

### DIFF
--- a/mods/cnc/rules/aircraft.yaml
+++ b/mods/cnc/rules/aircraft.yaml
@@ -245,7 +245,7 @@ TRAN.Husk:
 		TurnSpeed: 5
 		Speed: 140
 	RevealsShroud:
-		Range: 8c0
+		Range: 10c0
 		Type: CenterPosition
 	WithIdleOverlay@ROTOR1:
 		Offset: 597,0,85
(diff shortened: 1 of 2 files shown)


MustaphaTR:
:+1:

obrakmann (approved):
👍

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/2ce741d521ad6097a1ffd7dd3e67cbfe143ca928)

PR #12160, opened 2016-10-03
Fix yaks trying to attack invalid targets

abcdefg30:
Fixes #12128.
You can test this by placing a lot of camo pillboxes.

I removed the `yak.Wait(DateTime.Seconds(5))` as it just seems to make the yak stay in air without moving.

PR #12220, opened 2016-10-13
Fix yak attack in Evacuation

obrakmann:
Depends on #12160.

### DIFF
--- a/mods/ra/maps/evacuation/evacuation.lua
+++ b/mods/ra/maps/evacuation/evacuation.lua
@@ -73,16 +73,16 @@ SpawnAlliedReinforcements = function()
 end
 
 Yak = nil
-YakAttack = function(yak)
+YakAttack = function()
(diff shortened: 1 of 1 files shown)


reaperrr:
You can rebase now.

obrakmann:
Rebased.

reaperrr:
It seems the Yak doesn't attack Tanya if she's on her own, is that intentional?
Otherwise the trigger seems to work, though.

abcdefg30:
> It seems the Yak doesn't attack Tanya if she's on her own, is that intentional?

Yes. It would be kind of unfair otherwise, as the mission fails if Tanya is killed (and you can't really do much against those yak attacks, at least not before they killed their target).

Looks good to me as well. 👍 / ✅

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog/_compare/edb23e307db51538c7cb3258cd6223679215b4f7...f7a41cd98e8cf73cdd3a5bedf35c2131913e90a0)

dd50004d91e6afd21183226527aa857f62f0f9c4

PR #12223, opened 2016-10-14
Add sprite placeholder for TS ClusterMissile

reaperrr:
Extracted from crate animation (frame 16, flipped vertically for `down` sequence).

In my opinion it's better to have a visible placeholder.


GraionDilach:
I have/had(?) a 32-faced SHP version made from screenshotting the voxel in OpenRA in a pink dummy terrain posted in the cluster missile PR though. Might be better to use that one.

reaperrr:
> I have/had(?) a 32-faced SHP version made from screenshotting the voxel in OpenRA in a pink dummy terrain posted in the cluster missile PR though. Might be better to use that one.

Couldn't find it there, but either way, that would require to implement at least part of #11624, I already have too many other things on my roadmap I want to get in before Next+1.

This is supposed to be just a simple, temporary cosmetical fix.

pchote (approved):
:+1:

abcdefg30:
Not really a fan, but won't hurt that much either.

Issue #11505, opened 2016-06-22
Engineer can repair enemy building.


epicelite:
STEPS TO REPRODUCE:
1)Capture an oil derrick
2)Have it damaged a little bit somehow
3)Have the derrick captured by a enemy player (The derrick should go underneath the fog of war)
4)Select a engineer and try to tell them to capture the derrick, it will give the repair option
5)Capture option will be restored once the derrick is revealed, but it you use the repair command and don't re-issue commands to the engineer they will just repair it.

I say oil derrick because that's what happened to me, but I bet it would work with any building.

Phrohdoh:
Once the target is revealed we'll need to reevaluate what order should be given, and decide whether or not human interaction is necessary (can we automatically give the new order?).

pchote:
The simpler and more correct fix (as a baseline, at least) is to properly validate the target before applying the repair, and have the engineer leave again if it fails.  This is what we do in most other unit -> building cases.

epicelite:
Well double check that it actually does because I stopped the engineer just before he would of done it because it was the middle of a game and I needed him to capture it, but it looked like he was gonna fix it.

obrakmann:
FWIW, there's a real bug here. The engineer enters the enemy building and is lost. Didn't check whether any repairing actually happened.

TheRaffy:
Looks like my reported bug #12121 was somehow already known. Thanks to abcdefg30 for connecting the dots. That means I can prove this issue here with a replay. Feel free to check #12121.
Somebody also might wanna close my issue report, because of dupplication.  Thanks!

PR #12232, opened 2016-10-15
Fix for #11505

entropai:
Fix for #11505 checks for valid stance now. Updated OnInside for additional disposing validation (othervise engineer would just enter and dispose)

### DIFF
--- a/OpenRA.Mods.Common/Activities/CaptureActor.cs
+++ b/OpenRA.Mods.Common/Activities/CaptureActor.cs
@@ -37,13 +37,13 @@ namespace OpenRA.Mods.Common.Activities
 			return !capturable.BeingCaptured && capturable.Info.CanBeTargetedBy(self, actor.Owner);
 		}
 
-		protected override void OnInside(Actor self)
+		protected override bool OnInside(Actor self)
 		{
-			if (actor.IsDead || capturable.BeingCaptured)
-				return;
+            if (actor.IsDead || capturable.BeingCaptured)
+				return false;
 
 			if (building != null && !building.Lock())
-				return;
+				return false;
 
 			self.World.AddFrameEndTask(w =>
 			{
@@ -81,6 +81,7 @@ namespace OpenRA.Mods.Common.Activities
 
 				self.Dispose();
 			});
+			return false;
 		}
 	}
 }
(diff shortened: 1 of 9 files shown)


OpenRA.Mods.Common/Activities/Enter.cs
```diff
@@ -217,11 +217,11 @@ State FindAndTransitionToNextState(Actor self)
 						return State.ApproachingOrEntering;
 					}
 
-					OnInside(self);
+					var shouldDispose = OnInside(self);
```
pchote: This will also need to cover the suicide case: I suggest calling it `completed` or `valid` or similar.

OpenRA.Mods.Common/Activities/CaptureActor.cs
```diff
@@ -81,6 +81,7 @@ protected override void OnInside(Actor self)
 
 				self.Dispose();
 			});
+			return false;
```
pchote: This should then return true.

pchote: style nit: newline above,  since that is the end of a block.

OpenRA.Mods.Common/Activities/Demolish.cs
```diff
@@ -79,6 +79,7 @@ protected override void OnInside(Actor self)
 						demolishables.Do(d => d.Demolish(target, self));
 				}));
 			});
+            return false;
```
pchote: and likewise here.

OpenRA.Mods.Common/Activities/EnterTransport.cs
```diff
@@ -55,6 +55,7 @@ protected override void OnInside(Actor self)
 			});
 
 			Done(self);
+            return false;
```
pchote: and here.

pchote (dismissed):

pchote (changes requested):
Sorry for taking so long to get back to this.  The code changes look correct, although I haven't yet tested them.  Can you please use `git rebase -i` to squash all the commits together?  Ask in IRC if you need help on how to do that.

reaperrr:
Superseded by #12535.

PR #12234, opened 2016-10-15
Let the Lonestar AI use the new Sell function instead of Destroy

abcdefg30:
By the time I wrote the AI we didn't have that function. But now (as we have it) we should actually use it, I think. (As the AI "tells" the player it "sold" the walls anyway.)


Mailaender (approved):

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/13d9557032d068cada727efbbede68f3a0ee673e...9a804f47725982845d9633d5b6fd92c1335ad5ae)

PR #12244, opened 2016-10-16
Fixed map import for all d2k missions

CH4Code:
Added 58 Templates to d2k tileset (including few neccessary duplicates) to fix OpenRA.Utility map importer.

OpenRA.Utility Map importer now imports all 39 d2k missions without tile errors!
Also changed properties of few existing tiles to represent original game behaviour.

### DIFF
--- a/mods/d2k/tilesets/arrakis.yaml
+++ b/mods/d2k/tilesets/arrakis.yaml
@@ -1993,10 +1993,10 @@ Templates:
 		Size: 2,2
 		Category: Rotten-Base
 		Tiles:
-			0: Cliff
+			0: Rough
 			1: Cliff
-			2: Cliff
-			3: Cliff
+			2: Rough
+			3: Rough
 	Template@252:
 		Id: 252
 		Images: BLOXBASE.R8
@@ -2475,10 +2475,10 @@ Templates:
 		Size: 2,2
 		Category: Rotten-Base
 		Tiles:
-			0: Cliff
+			0: Rough
 			1: Cliff
-			2: Cliff
-			3: Cliff
+			2: Rough
+			3: Rough
 	Template@307:
 		Id: 307
 		Images: BLOXBAT.R8
@@ -3776,9 +3776,9 @@ Templates:
 		Size: 3,1
 		Category: Sand-Detail
 		Tiles:
-			0: Rough
+			0: Cliff
 			1: Rough
-			2: Rough
+			2: Cliff
 	Template@436:
 		Id: 436
 		Images: BLOXTREE.R8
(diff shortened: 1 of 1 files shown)


pchote:
The Arrakis tileset merges all the original d2k tilesets into one, and intentionally leaves out all these duplicated tiles that are defined in more than one tilesets.  I think it would be a mistake to add these duplicates in now.

It would be better if you could either modify the tile id translation that the map importer already does to account for these duplicates, or, if people really want to commit to working on the d2k misssions, the mod could include "vanilla" blox* tilesets for the missions to use directly.

CH4Code:
Yes, you are right. I thought initially about replacing the Tile-ID while converting with a simple `If(blox****):switch:case` statement in `GetTile(int tileIndex)`, replacing the few exceptions where the "lazy" method fails. (it fails because the frame-IDs for the same terrain feature can vary significantly for different tilesets)

But there were tiles missing. This is also a reason why vanilla blox* wont work without new tiles.  So the tileset has to change anyway. Also often tiles were falesly integrated into bigger templates (like what should have been a 1x1 tile was actually on the 2,2 position of a 2x2 tile). Would be very tedious to solve this if it is at all solvable. Also, there were allready(!) duplicates (i.e. ID 386 and 359) in the tileset.

Alternatively, we could make a tag that hides the (manually identified) duplicate tiles from the user in the editor. 
Of course to find the duplicates ( for ID converting) i first had to put them into the editor, so I did just the yaml edit which allready fixes everything...

What do you think? which course of action would be right, and why?

pchote:
> But there were tiles missing. This is also a reason why vanilla blox* wont work without new tiles.

I'm not sure what you mean by this, but if you're saying that there are bugged / bad / missing tile definitions in our arrakis.yaml then we're in agreement that these should be fixed.

My only objection is to the idea of adding tiles that are visual duplicates to our arrakis.yaml tilesets.   I don't have a strong opinion right now whether it will be better to solve this with a big switch case in the d2k importer, or to define new blox*.yaml tilesets that are direct ports of the original d2k versions (for missions only).

CH4Code:
> I'm not sure what you mean by this
In order to get an impression how the map should actually look I repalced all occurences of BLOXBASE in arrakis.yaml with BLOX*, so the tile-getter-function would select from BLOX* tileset first.( So i had 6 tilesets). Only on 14 of 42 maps this leads to a flawless map. To completely fix all maps, all "missing tiles" had to be included.

I will try the switch:case solution then, removing identical duplicates. I would prefer this because it allows the added non-duplicate assets into the map editor, advancing variety of tiles which can only be good for creativity ;)

pchote:
> Only on 14 of 42 maps this leads to a flawless map. To completely fix all maps, all "missing tiles" had to be included.

Right.  This is because there are visual duplicates *with different frame numbers* across the tilesets.  The "missing" tiles are just ones where arrakis.yaml took the version defined in one of the other BLOX* sets.

CH4Code:
But what about (single) tiles - sometimes they were "embedded" into arrakis tiles. Still displayed correctly after conversion with blox* but only because a fraction of an arrakis tile is then used.
I checked it, and will upload later - not all my added tiles are true duplicates. You'll see.
Cheers.

pchote:
Those will be legitimate errors in our arrakis.yaml (which was assembled by hand long before anyone considered importing the original d2k maps).  Can you please file a separate PR to fix those cases?

CH4Code:
This fix should be considered temporary, better than nothing.

This should be replaced by (as discussed with pchote) 
1) an openra tileset for each mission tileset (Blox*), which will only be used in the d2k map-importer to map the mission tilesets onto the arrakis.yaml (which is a reduced tileset containing all tiles only once)
2) a proper arrakis tileset without errors
3) a proper map-importer using openra mission tilesets
(See issue ...)

This temporary fix does:
- allow all original missions to be ported without error (see [here](https://github.com/CH4Code/D2k-Campaign-Resources))

This temporary fix doesn't:
- fix the arrakis tileset properly or the map-importer

pchote:
Sorry, but I'm not a fan of this even as a temporary thing – I see this as taking on technical debt with no obvious payoff.  If we had people lined up waiting to script all the d2k missions then it would be one thing, but right now we don't have the scripting manpower to cover even our two most popular mods.

CH4Code:
For one, I am interested in d2k map scripting! Not directly the mission maps though, but as a side product (i want to make cooperative versus play (playing missions human v human ). The first map is nearly done.). 
I wanted to contribute to openra as a side product, too. I could have just fixed the neccessary maps in the map editor not getting this project forward an inch, but decided to try a fix for ALL maps instead. If you tell me now that won't consider this even though there is no better solution right now, i wonder if it is worth to put any more time into contributing. All potential campaign scripters first have to face the problem that there are few flawless maps to work with!!

pchote:
The other tile fixes were pulled into and merged with #12653.  Nobody else has given an opinion on the importing problem, so unfortunately I think this has run its course.

abcdefg30:
I tried to complete this in #12784.

PR #12245, opened 2016-10-16
Split a WithAimAnimation from WithAttackAnimation and make both upgradable

### LINKED ISSUE #11884: WithAttackAnimation and WithMoveAnimation can't handle conditions
Meaning neither are the traits upgradable themselves, nor can they handle multiple sprite bodies.
- [x] Make `WithAttackAnimation` capable of handling multiple sprite bodies. #13350
- [x] Make `WithAttackAnimation` conditional. #13350
- [x] Make `WithMoveAnimation` capable of handling multiple sprite bodies. #13294
- [ ] Make `WithMoveAnimation` conditional.

reaperrr:
While the `With*SpriteBody`/`With*Animation` interaction might need a refactor to prevent or at least reduce conflicts between multiple animations, for now making `WithAttackAnimation` upgradable + able to handle multiple sprite bodies should be good enough for most mods.

Additionally, I split the aiming animation stuff off to its own trait, because these two aspects didn't really work properly together anyway. This also made the code cleaner and more straightforward.

A nice side-effect of the split is that `WithAttackAnimation` now only plays when the specified armament is used for attacking.

Last but not least, due to their current incompatibility with each other, a lint warning and a yaml exception when both are enabled at actor creation have been added to prevent modders from using them at the same time.

Upgrade rules, lint check and yaml exception have all been tested and worked when I tested them.

Closes #11884 (assuming #11886 is merged before this).

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
@@ -14,7 +14,7 @@ using OpenRA.Traits;
 
 namespace OpenRA.Mods.Common.Traits.Render
 {
-	public class WithAttackAnimationInfo : ITraitInfo, Requires<WithSpriteBodyInfo>, Requires<ArmamentInfo>, Requires<AttackBaseInfo>
+	public class WithAttackAnimationInfo : UpgradableTraitInfo, Requires<WithSpriteBodyInfo>, Requires<ArmamentInfo>, Requires<AttackBaseInfo>
 	{
 		[Desc("Armament name")]
 		public readonly string Armament = "primary";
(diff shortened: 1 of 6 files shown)


OpenRA.Mods.Common/Traits/Render/WithAttackAnimation.cs
```diff
@@ -14,93 +14,80 @@
...
+			wsbs = init.Self.TraitsImplementing<WithSpriteBody>().ToArray();
 		}
 
 		void PlayAttackAnimation(Actor self)
 		{
-			if (!string.IsNullOrEmpty(info.AttackSequence))
-				wsb.PlayCustomAnimation(self, info.AttackSequence, () => wsb.CancelCustomAnimation(self));
+			foreach (var wsb in wsbs)
+				if (!wsb.IsTraitDisabled && !string.IsNullOrEmpty(Info.AttackSequence))
+					wsb.PlayCustomAnimation(self, Info.AttackSequence, () => wsb.CancelCustomAnimation(self));
```
pchote: What happens if you switch bodies in the middle of an animation?  I suspect that the disabled traits won't be ticking the animation, meaning that you end up with the animations being out of sync.

reaperrr: Thinking aloud, might replacing this `IsTraitDisabled` check with an identifier/`Name` lookup (kind of like what we do with armaments already) help, or does it depend on the body itself rather than the modifying trait?

pchote: Unfortunately it does depend on the body itself.
Even more unfortunately, the only way I can see to fix this is the full-blown rewrite that I speculated about in #11886.

reaperrr:
> Unfortunately it does depend on the body itself.
Even more unfortunately, the only way I can see to fix this is the full-blown rewrite that I speculated about in #11886.

I don't think I'll be able to tackle that (not this year, at least), so I'm closing this for now.

PR #12258, opened 2016-10-18
Move production icon sequence from Tooltip to Buildable

reaperrr:
Makes more sense, and saves us some `Tooltip` look-ups which also allows to make the latter upgradable more easily and regression-risk-free.

All yaml changes were done via upgrade rule, but because _all TD husk and barb wire/wood fence icon definitions were bogus_ (never caught due to lack of `SequenceReference`), only the removals were staged for the C&C mod.

Prerequisite for #11984.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Buildable.cs
+++ b/OpenRA.Mods.Common/Traits/Buildable.cs
@@ -34,6 +34,9 @@ namespace OpenRA.Mods.Common.Traits
 		[Desc("Force a specific faction variant, overriding the faction of the producing actor.")]
 		public readonly string ForceFaction = null;
 
+		[Desc("Sequence of the actor that contains the icon.")]
+		[SequenceReference] public readonly string Icon = "icon";
+
 		[Desc("Palette used for the production icon.")]
 		[PaletteReference] public readonly string IconPalette = "chrome";
 
--- a/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
+++ b/OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
@@ -400,8 +400,8 @@ namespace OpenRA.Mods.Common.UtilityCommands
 				// Move production description from Tooltip to Buildable
 				if (engineVersion < 20161016)
 				{
-					var tooltipChild = node.Value.Nodes.FirstOrDefault(n => n.Key == "Tooltip");
-					if (tooltipChild != null || (tooltipChild = node.Value.Nodes.FirstOrDefault(n => n.Key == "DisguiseToolTip")) != null)
+					var tooltipChild = node.Value.Nodes.FirstOrDefault(n => n.Key == "Tooltip" || n.Key == "DisguiseToolTip");
+					if (tooltipChild != null)
 					{
 						var descNode = tooltipChild.Value.Nodes.FirstOrDefault(n => n.Key == "Description");
 						if (descNode != null)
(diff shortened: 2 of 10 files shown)


OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -416,6 +416,25 @@ internal static void UpgradeActorRules(ModData modData, int engineVersion, ref L
 					}
 				}
 
+				// Move production icon sequence from Tooltip to Buildable
+				if (engineVersion < 20161018)
+				{
+					var tooltipChild = node.Value.Nodes.FirstOrDefault(n => n.Key == "Tooltip");
+					if (tooltipChild != null || (tooltipChild = node.Value.Nodes.FirstOrDefault(n => n.Key == "DisguiseToolTip")) != null)
```
pchote: Reassigning `tooltipChild` inside the condition check is a bit dirty.  Can you change the line above to the following instead?
```csharp
var tooltipChild = node.Value.Nodes.FirstOrDefault(n => n.Key == "Tooltip")
    ?? node.Value.Nodes.FirstOrDefault(n => n.Key == "DisguiseToolTip");
```

OpenRA.Mods.Common/Traits/Buildable.cs
```diff
@@ -34,6 +34,9 @@ public class BuildableInfo : TraitInfo<Buildable>
 		[Desc("Force a specific faction variant, overriding the faction of the producing actor.")]
 		public readonly string ForceFaction = null;
 
+		[Desc("Sequence of the actor that contains the cameo.")]
```
pchote: The original games called the icons "cameo"s, but we don't except for here and the legacy sequence importer (where we can't really avoid it).  Can you please change "cameo" &rarr; "production icon" or "sidebar icon" here?

pchote (changes requested):

reaperrr:
Updated.

pchote (approved):

OpenRA.Mods.Common/UtilityCommands/UpgradeRules.cs
```diff
@@ -400,8 +400,10 @@ internal static void UpgradeActorRules(ModData modData, int engineVersion, ref L
 				// Move production description from Tooltip to Buildable
 				if (engineVersion < 20161016)
 				{
-					var tooltipChild = node.Value.Nodes.FirstOrDefault(n => n.Key == "Tooltip");
-					if (tooltipChild != null || (tooltipChild = node.Value.Nodes.FirstOrDefault(n => n.Key == "DisguiseToolTip")) != null)
+					var tooltipChild = node.Value.Nodes.FirstOrDefault(n => n.Key == "Tooltip")
```
abcdefg30: Wouldn't it be simpler to use `var tooltipChild = node.Value.Nodes.FirstOrDefault(n => n.Key == "Tooltip" || n.Key == "DisguiseToolTip")`?

abcdefg30:

reaperrr:
Updated upgrade rule as suggested and updated engineVersion to today.

OpenRA.Mods.Common/Traits/Buildable.cs
```diff
@@ -34,6 +34,9 @@ public class BuildableInfo : TraitInfo<Buildable>
 		[Desc("Force a specific faction variant, overriding the faction of the producing actor.")]
 		public readonly string ForceFaction = null;
 
+		[Desc("Sequence of the actor that contains the icon.")]
+		[SequenceReference] public readonly string Icon = "icon";
```
Phrohdoh: `IconSequence` maybe.
This doesn't have to be changed because you're just moving code around, obviously.

reaperrr: Basically I agree, but yeah I'd like to leave that for a separate PR, to get this dependency for #11984 out of the way asap.

reaperrr:
Updated, fixed the missing `;`.

abcdefg30:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/0d377ec6f40172897f9476162c0ff0b5a3b17a69...13d9557032d068cada727efbbede68f3a0ee673e)

PR #12267, opened 2016-10-20
Reduce aircraft ground reveal

### LINKED ISSUE #2535: Airplanes should not reveal that much while on the ground
While in the air, airplanes have large reveal range due to their altitude, however on the ground it should be limited - their reveal range on the ground is currently much beyond anything else which should not be the case.

Probably the best would be either to linearly increase their reveal range depending on the altitude or have a separate reveal range for when they are landed.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/100235-airplanes-should-not-reveal-that-much-on-the-ground?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Uses the upgradability to reduce aircraft sight range to 3 cells when landed in RA and TD.

Closes #2535.

### DIFF
--- a/OpenRA.Game/Traits/World/Shroud.cs
+++ b/OpenRA.Game/Traits/World/Shroud.cs
@@ -54,8 +54,8 @@ namespace OpenRA.Traits
 
 		// Cache of visibility that was added, so no matter what crazy trait code does, it
 		// can't make us invalid.
-		readonly Dictionary<Actor, PPos[]> visibility = new Dictionary<Actor, PPos[]>();
-		readonly Dictionary<Actor, PPos[]> generation = new Dictionary<Actor, PPos[]>();
+		readonly Dictionary<object, PPos[]> visibility = new Dictionary<object, PPos[]>();
+		readonly Dictionary<object, PPos[]> generation = new Dictionary<object, PPos[]>();
 
 		[Sync] bool disabled;
 		public bool Disabled
@@ -135,11 +135,8 @@ namespace OpenRA.Traits
 			return ProjectedCellsInRange(map, map.CenterOfCell(cell), range);
 		}
 
-		public void AddProjectedVisibility(Actor a, PPos[] visible)
+		public void AddProjectedVisibility(object key, PPos[] visible)
 		{
-			if (!a.Owner.IsAlliedWith(self.Owner))
-				return;
-
 			foreach (var puv in visible)
 			{
 				// Force cells outside the visible bounds invisible
@@ -151,17 +148,17 @@ namespace OpenRA.Traits
 				explored[uv] = true;
 			}
 
-			if (visibility.ContainsKey(a))
+			if (visibility.ContainsKey(key))
 				throw new InvalidOperationException("Attempting to add duplicate actor visibility");
 
-			visibility[a] = visible;
+			visibility[key] = visible;
 			Invalidate(visible);
 		}
 
-		public void RemoveVisibility(Actor a)
+		public void RemoveVisibility(object key)
 		{
 			PPos[] visible;
-			if (!visibility.TryGetValue(a, out visible))
+			if (!visibility.TryGetValue(key, out visible))
 				return;
 
 			foreach (var puv in visible)
@@ -171,62 +168,33 @@ namespace OpenRA.Traits
 					visibleCount[(MPos)puv]--;
 			}
 
-			visibility.Remove(a);
+			visibility.Remove(key);
 			Invalidate(visible);
 		}
 
-		public void AddProjectedShroudGeneration(Actor a, PPos[] shrouded)
+		public void AddProjectedShroudGeneration(object key, PPos[] shrouded)
 		{
-			if (a.Owner.IsAlliedWith(self.Owner))
-				return;
-
 			foreach (var uv in shrouded)
 				generatedShroudCount[(MPos)uv]++;
 
-			if (generation.ContainsKey(a))
+			if (generation.ContainsKey(key))
 				throw new InvalidOperationException("Attempting to add duplicate shroud generation");
 
-			generation[a] = shrouded;
+			generation[key] = shrouded;
 			Invalidate(shrouded);
 		}
 
-		public void RemoveShroudGeneration(Actor a)
+		public void RemoveShroudGeneration(object key)
 		{
 			PPos[] shrouded;
-			if (!generation.TryGetValue(a, out shrouded))
+			if (!generation.TryGetValue(key, out shrouded))
 				return;
 
 			foreach (var uv in shrouded)
 				generatedShroudCount[(MPos)uv]--;
 
-			generation.Remove(a);
+			generation.Remove(key);
 			Invalidate(shrouded);
-		}
-
-		public void UpdatePlayerStance(World w, Player player, Stance oldStance, Stance newStance)
-		{
-			if (oldStance == newStance)
-				return;
-
-			foreach (var a in w.Actors.Where(a => a.Owner == player))
-			{
-				PPos[] visible = null;
-				PPos[] shrouded = null;
-				foreach (var p in self.World.Players)
-				{
-					if (p.Shroud.visibility.TryGetValue(self, out visible))
-					{
-						p.Shroud.RemoveVisibility(self);
-						p.Shroud.AddProjectedVisibility(self, visible);
-					}
-
-					if (p.Shroud.generation.TryGetValue(self, out shrouded))
-					{
-						p.Shroud.RemoveShroudGeneration(self);
-						p.Shroud.AddProjectedShroudGeneration(self, shrouded);
-					}
-				}
-			}
 		}
 
 		public void ExploreProjectedCells(World world, IEnumerable<PPos> cells)
--- a/OpenRA.Mods.Common/Traits/AffectsShroud.cs
+++ b/OpenRA.Mods.Common/Traits/AffectsShroud.cs
@@ -55,7 +55,7 @@ namespace OpenRA.Mods.Common.Traits
 				.ToArray();
 		}
 
-		public void Tick(Actor self)
+		void ITick.Tick(Actor self)
 		{
 			if (!self.IsInWorld)
 				return;
@@ -79,7 +79,7 @@ namespace OpenRA.Mods.Common.Traits
 			}
 		}
 
-		public void AddedToWorld(Actor self)
+		void INotifyAddedToWorld.AddedToWorld(Actor self)
 		{
 			var centerPosition = self.CenterPosition;
 			var projectedPos = centerPosition - new WVec(0, centerPosition.Z, centerPosition.Z);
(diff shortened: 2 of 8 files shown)


MustaphaTR:
I'm OK with the balance changes this PR will cause. :+1: Not eated tho.

OpenRA.Mods.Common/Traits/CreatesShroud.cs
```diff
@@ -20,8 +20,19 @@ public class CreatesShroud : AffectsShroud
 	{
 		public CreatesShroud(Actor self, CreatesShroudInfo info)
 			: base(self, info) { }
-		protected override void AddCellsToPlayerShroud(Actor self, Player p, PPos[] uv) { p.Shroud.AddProjectedShroudGeneration(self, uv); }
-		protected override void RemoveCellsFromPlayerShroud(Actor self, Player p) { p.Shroud.RemoveShroudGeneration(self); }
-		protected override bool IsDisabled(Actor self) { return self.IsDisabled(); }
+
+		protected override void AddCellsToPlayerShroud(Actor self, Player p, PPos[] uv)
+		{
+			if (!IsDisabled(self))
```
pchote: Declaring `UpgradableTrait` on the parent `AffectsShroud` but then implementing all the bits that make it work in the subclasses is a bit dirty.  Can you please move the `IsTraitDisabled` check up into `AffectsShroud.Tick` instead?

pchote (changes requested):

reaperrr:
Updated.

OpenRA.Mods.Common/Traits/AffectsShroud.cs
```diff
@@ -64,12 +65,20 @@ public void Tick(Actor self)
...
 			var disabled = IsDisabled(self);
+			var traitDisabled = IsTraitDisabled;
+
+			if (traitDisabled)
+			{
+				cachedTraitDisabled = traitDisabled;
+				return;
+			}
 
 			if (cachedLocation == projectedLocation && cachedDisabled == disabled)
```
pchote: We'll need to update the state when the `cachedTraitDisabled` changes, so this will want to look something like
```csharp
if (cachedTraitDisabled != traitDisabled || (cachedLocation == projectedLocation && cachedDisabled == disabled))
```

OpenRA.Mods.Common/Traits/AffectsShroud.cs
```diff
@@ -64,12 +65,20 @@ public void Tick(Actor self)
 			var projectedPos = centerPosition - new WVec(0, centerPosition.Z, centerPosition.Z);
 			var projectedLocation = self.World.Map.CellContaining(projectedPos);
 			var disabled = IsDisabled(self);
+			var traitDisabled = IsTraitDisabled;
+
+			if (traitDisabled)
+			{
+				cachedTraitDisabled = traitDisabled;
+				return;
```
pchote: We need to `RemoveCellsFromPlayerShroud` when changing from enabled &rarr; disabled too!

pchote (changes requested):

OpenRA.Mods.Common/Traits/AffectsShroud.cs
reaperrr: Uh, but if `traitDisabled` is `false` and `cachedTraitDisabled` is `true`, we do *not* want to return early, but rather update, right?

reaperrr:
Updated.

abcdefg30:

OpenRA.Mods.Common/Traits/AffectsShroud.cs
```diff
@@ -64,12 +65,14 @@ public void Tick(Actor self)
...
-			if (cachedLocation == projectedLocation && cachedDisabled == disabled)
+			if (cachedLocation == projectedLocation && traitDisabled == cachedTraitDisabled && cachedDisabled == disabled)
 				return;
 
 			cachedLocation = projectedLocation;
 			cachedDisabled = disabled;
+			cachedTraitDisabled = traitDisabled;
 
 			var cells = ProjectedCells(self);
 			foreach (var p in self.World.Players)
```
pchote: You don't appear to be checking `traitDisabled` anywhere here?

OpenRA.Mods.Common/Traits/AffectsShroud.cs
```diff
@@ -79,23 +82,28 @@ public void Tick(Actor self)
...
+		void INotifyAddedToWorld.AddedToWorld(Actor self)
 		{
 			var centerPosition = self.CenterPosition;
 			var projectedPos = centerPosition - new WVec(0, centerPosition.Z, centerPosition.Z);
 			cachedLocation = self.World.Map.CellContaining(projectedPos);
 			cachedDisabled = IsDisabled(self);
+			cachedTraitDisabled = IsTraitDisabled;
 			var cells = ProjectedCells(self);
+
+			if (cachedTraitDisabled)
```
pchote: Shouldn't this case already be handled by setting the range to 0?

OpenRA.Mods.Common/Traits/AffectsShroud.cs
reaperrr: It's checked in line 70, otherwise it just serves to update `cachedTraitDisabled` in line 75.

reaperrr:
Updated.

I'm afraid due to the first commit @abcdefg30's +1 is now invalid.

OpenRA.Mods.Common/Traits/CreatesShroud.cs
```diff
@@ -18,10 +18,24 @@ public class CreatesShroudInfo : AffectsShroudInfo
...
 	{
+		readonly Actor self;
+
 		public CreatesShroud(Actor self, CreatesShroudInfo info)
-			: base(self, info) { }
-		protected override void AddCellsToPlayerShroud(Actor self, Player p, PPos[] uv) { p.Shroud.AddProjectedShroudGeneration(self, uv); }
-		protected override void RemoveCellsFromPlayerShroud(Actor self, Player p) { p.Shroud.RemoveShroudGeneration(self); }
+			: base(self, info) { this.self = self; }
+
+		protected override void AddCellsToPlayerShroud(object o, Player p, PPos[] uv)
```
pchote: You can avoid most of the changes in these two files by keeping this as `(Actor self, Player p, PPos[] uv)` and then passing doing `AddProjectedShroudGeneration(this, uv)`

OpenRA.Game/Traits/World/Shroud.cs
```diff
@@ -135,11 +135,8 @@ public static IEnumerable<PPos> ProjectedCellsInRange(Map map, CPos cell, WDist
 			return ProjectedCellsInRange(map, map.CenterOfCell(cell), range);
 		}
 
-		public void AddProjectedVisibility(Actor a, PPos[] visible)
+		public void AddProjectedVisibility(object o, PPos[] visible)
```
pchote: Please rename `o` to `key` through these methods?

pchote (changes requested):
Nearly there, I think!

OpenRA.Mods.Common/Traits/RevealsShroud.cs
```diff
@@ -18,9 +18,22 @@ public class RevealsShroudInfo : AffectsShroudInfo
...
 		public RevealsShroud(Actor self, RevealsShroudInfo info)
-			: base(self, info) { }
-		protected override void AddCellsToPlayerShroud(Actor self, Player p, PPos[] uv) { p.Shroud.AddProjectedVisibility(self, uv); }
-		protected override void RemoveCellsFromPlayerShroud(Actor self, Player p) { p.Shroud.RemoveVisibility(self); }
+			: base(self, info) { this.self = self; }
+
+		protected override void AddCellsToPlayerShroud(object o, Player p, PPos[] uv)
+		{
+			o = this;
+			if (self.Owner.IsAlliedWith(p))
```
pchote: It would be great if you could change this and `CreatesShroud` to use a stance check here with a TraitInfo property for who gets the reveal/hide.  This would be useful for scripted maps like allies-05, which currently teleports a camera actor over an enemy truck every tick.

pchote (dismissed):

pchote:
The changes in this PR make it much easier to implement the classic RA gap generator behaviour (actively blocking unit sight).  I'll have a PR for that after this one is finished and merged.

reaperrr:
Updated.

The stance support might turn out to be very useful for singleplayer.

mods/ra/rules/aircraft.yaml
```diff
@@ -97,9 +97,16 @@ MIG:
 		HP: 70
 	Armor:
 		Type: Light
+	RevealsShroud@GROUNDED:
```
pchote: Why aren't you defining these on `^Plane`, like you did for `^Helicopter`?

pchote:
Big :+1: for the code changes, but I don't really like the aircraft reveal change.  Having separate landed vs flying ranges feels like unnecessary complexity (is it *really* that much of a problem?), and I find the instant transition between the two while the aircraft is effectively still on the ground to be quite jarring.

reaperrr:
Due to the misleading branch name, I've split the code changes to #12308.

This PR will stay as test-case for now, after #12308 is merged we can decide whether we want the yaml changes or not.

obrakmann:
> I don't really like the aircraft reveal change. Having separate landed vs flying ranges feels like unnecessary complexity (is it really that much of a problem?), and I find the instant transition between the two while the aircraft is effectively still on the ground to be quite jarring.

When I was thinking about the problem a few months back, my idea was to have a `ScaleRangeWithAltitude` option. I think that would at least avoid the transition issue.

pchote:
I am thinking the same thing, but on a grander scale.  We need to introduce altitude-dependent sight ranges for the heightmap-enabled mods, so it makes sense to implement a system that covers both that and this.

I can't offer a more concrete suggestion yet because I don't have a feel for how TS/RA2 implemented their sight ranges (and chances are it was a total hack).

mods/ra/rules/aircraft.yaml
reaperrr: Because 3 out of 5 RA planes are not supposed to have any sight at all (both badgers and the Spy Plane), having to disable all these traits on those 3 would be uglier than defining them on the MiG and Yak directly, in my opinion.

pchote: Ah right, I see.  You could perhaps split out a separate `^PlaneVisibility` template?

reaperrr:
> I can't offer a more concrete suggestion yet because I don't have a feel for how TS/RA2 implemented their cliff-sight modifiers (and chances are it was a total hack).

At least in TS, aircraft had very low sight ranges and couldn't move into shroud either, so I'm not sure cliffs made much of a difference for them. Dunno how it was in RA2, though.

My guess would be that actors in those games only revealed terrain for height levels less than 4 levels (= cliff height) above them.

reaperrr:
I don't think I'll work on this for the next few weeks and the code changes are already in, so it'll be easy to revive if needed. Closing for now.

PR #12268, opened 2016-10-20
Several internal interface cleanups

reaperrr:
This PR 
- moves the check for explicit interface implementation violation to the start of `make check` so you don't have to wait for other style checks to finish
- moves several trait interfaces with no remaining ties to `Game` to `Mods.Common`
- makes two of the moved interfaces require explicit implementation
- explicitly implements interfaces where possible for the traits I had to touch
- moves warhead `ImpactTypes` to `Warhead`
- performs some additional minor style cleanups

All in all, the numbers of changed files and changed lines look far worse than they really are, because pretty much all of the changes are about moving stuff around or making interfaces explicit and would get caught at compile or `make check` if I messed up somewhere.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Buildings/Reservable.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/Reservable.cs
@@ -23,7 +23,7 @@ namespace OpenRA.Mods.Common.Traits
 		Actor reservedFor;
 		Aircraft reservedForAircraft;
 
-		public void Tick(Actor self)
+		void ITick.Tick(Actor self)
 		{
 			if (reservedFor == null)
 				return;		/* nothing to do */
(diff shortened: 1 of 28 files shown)


OpenRA.Mods.Common/Traits/Buildings/Reservable.cs
```diff
@@ -64,21 +65,25 @@ public static bool IsReserved(Actor a)
...
+		void INotifyOwnerChanged.OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
 		{
 			if (reservedForAircraft != null)
 				reservedForAircraft.UnReserve();
 		}
 
-		public void Selling(Actor self) { Sold(self); }
+		void INotifySold.Selling(Actor self)
+		{
+			if (reservedForAircraft != null)
```
pchote: This block is now duplicated four times.  Can you please define a private `UnReserve()` method they all call?

pchote (commented):
Looks good otherwise

reaperrr:
Updated.

pchote (approved):

phrohdoh (approved):

reaperrr:
This already has +2 and a follow-up is waiting, so I'll merge it myself now.

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/9b8f395134255ff05e70cef83b722754fd94d9d7...339724cd56b7939ef8b215164f6e2d0c88b467e1)

PR #12274, opened 2016-10-21
Fully support unhardcoded sequences in the voxel render traits

abcdefg30:
Seems like someone missed unhardcoding the strings in `RenderPreviewVoxels`.
Also unhardcoded the sequence of `WithVoxelWalkerBody`.


reaperrr (approved):

OpenRA.Mods.TS/Traits/Render/WithVoxelWalkerBody.cs
```diff
@@ -24,6 +24,8 @@ namespace OpenRA.Mods.TS.Traits.Render
 {
 	public class WithVoxelWalkerBodyInfo : ITraitInfo, IRenderActorPreviewVoxelsInfo,  Requires<RenderVoxelsInfo>, Requires<IMoveInfo>, Requires<IFacingInfo>
 	{
+		public readonly string Sequence = "idle";
```
pchote: If we are going to unhardcode these then we really should have a lint check for them.

pchote (approved):

PR #12275, opened 2016-10-21
Fix the search for alternative transports not caring about passenger types

### LINKED ISSUE #12273: Any tank type can be loaded into ANY transportation unit by using a trick
I found out that you can load ANY tank type into choppers, Phase tanks, APC's, rangers, just by using a simple trick. Normally you should only be able to do this in a naval transport, But i found out that by selecting an over-amount of units that the transports u have at the shoreline which the naval transports can hol, using the Alt-shift combination on 2 naval transports(so thats 10 tanks of any kind, And besides that you can remove the second naval transport before units reach them to load themselves in), In combination with ANY other transport unit like a chopper, phase transport, apc, or a ranger, setup close to the same shoreline will make the tankgroup which is left after the naval transport is fully loaded, load into the other transport units that are placed close to the same shoreline. I tested this over and over in single player and multiplayer. And it works. You can basicly make a chinook drop with 16 artilleries behind the enemy base without any trouble. which is much faster and than a chronoshift, and lasts indefinately. and the amount can be stacked as high as you want to. I really think this is a bug that needs to be fixed. Waiting for a response, And before that, CHEERS!!! :D Ohmk Torque (ingame player name)

abcdefg30:
Fixes #12273.

### DIFF
--- a/OpenRA.Mods.Common/Activities/EnterTransport.cs
+++ b/OpenRA.Mods.Common/Activities/EnterTransport.cs
@@ -65,7 +65,7 @@ namespace OpenRA.Mods.Common.Activities
 			return TryGetAlternateTargetInCircle(
 				self, passenger.Info.AlternateTransportScanRange,
 				t => { transport = t.Actor; cargo = t.Actor.Trait<Cargo>(); }, // update transport and cargo
-				a => { var c = a.TraitOrDefault<Cargo>(); return c != null && (c.Unloading || c.CanLoad(a, self)); },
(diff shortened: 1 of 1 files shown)


Micr0Bit:
:+1:

phrohdoh (approved):

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/04ccbee67d5b40db2d347254d2f50e8ef24a1406...cc49e8f285e03bc001b20b8ed232326fd4957748)

PR #12284, opened 2016-10-22
More explicit interface cleanups

reaperrr:
Follow-up to #12268.

This one might need at least a bit of testing for potential regressions on sprite bodies, sprite turrets and armaments, though in my own testing I didn't encounter any issues.

To make up for it, the diff is a lot smaller this time.
- moves `INotifyDamageStateChanged` to mod code
- makes explicit interface implementation mandatory for the above and `INotifyBuildComplete`
- adresses a few more cases that previously made implicit interfaces overridable, preventing explicit implementations
- some more misc 'collateral' explicit implementations

### DIFF
--- a/OpenRA.Mods.Common/Traits/Armament.cs
+++ b/OpenRA.Mods.Common/Traits/Armament.cs
@@ -139,7 +139,7 @@ namespace OpenRA.Mods.Common.Traits
 			return new WDist(Util.ApplyPercentageModifiers(Weapon.Range.Length, rangeModifiers));
 		}
 
-		public virtual void Created(Actor self)
+		protected virtual void Created(Actor self)
 		{
 			turret = self.TraitsImplementing<Turreted>().FirstOrDefault(t => t.Name == Info.Turret);
 			ammoPool = self.TraitsImplementing<AmmoPool>().FirstOrDefault(la => la.Info.Name == Info.AmmoPoolName);
@@ -147,7 +147,7 @@ namespace OpenRA.Mods.Common.Traits
 			rangeModifiers = self.TraitsImplementing<IRangeModifier>().ToArray().Select(m => m.GetRangeModifier());
 		}
 
-		public virtual void Tick(Actor self)
+		protected virtual void Tick(Actor self)
 		{
 			if (IsTraitDisabled)
 				return;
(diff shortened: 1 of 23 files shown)


obrakmann (approved):
Looks fine to me, 👍

OpenRA.Mods.Common/Traits/Armament.cs
```diff
@@ -139,15 +139,15 @@ public virtual WDist MaxRange()
 			return new WDist(Util.ApplyPercentageModifiers(Weapon.Range.Length, rangeModifiers));
 		}
 
-		public virtual void Created(Actor self)
+		protected virtual void Initialize(Actor self)
```
pchote: Can we please keep the original names here?  Renaming them just adds confusion.

OpenRA.Mods.Common/Traits/Armament.cs
```diff
@@ -168,6 +168,16 @@ public virtual void Tick(Actor self)
 			delayedActions.RemoveAll(a => a.First <= 0);
 		}
 
+		void INotifyCreated.Created(Actor self)
+		{
+			Initialize(self);
```
pchote: It would be worth a comment here explaining that the implementation is split into a protected method to allow subclassing.

OpenRA.Mods.Common/Traits/Render/WithSpriteTurret.cs
```diff
@@ -113,13 +113,13 @@ public string NormalizeSequence(Actor self, string sequence)
...
 
-		public virtual void DamageStateChanged(Actor self, AttackInfo e)
+		protected virtual void DamageStateChanged(Actor self)
 		{
 			if (DefaultAnimation.CurrentSequence != null)
 				DefaultAnimation.ReplaceAnim(NormalizeSequence(self, DefaultAnimation.CurrentSequence.Name));
 		}
 
-		public virtual void Tick(Actor self)
+		protected virtual void ReplaceSequenceWhenAiming(Actor self)
```
pchote: Here too (in subclasses it may have nothing to do with aiming).

pchote (changes requested):

reaperrr:
Updated as requested.

pchote (approved):

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/8351f8b5961e0157872325f1a4e30a69a3a12342

PR #12288, opened 2016-10-22
Added SpawnDelay for SpawnActorOnDeath trait

reaperrr:
> This new property will delay actor spawn on death (useful when you need to play unit's death animation first without overlapping dead actor and new actor).

Supersedes #11752.

Basically the important part of #11752 + some style fixes by me.

### DIFF
--- a/OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs
+++ b/OpenRA.Mods.Common/Traits/SpawnActorOnDeath.cs
@@ -9,8 +9,9 @@
  */
 #endregion
 
+using System;
 using System.Linq;
-using OpenRA.Mods.Common.Warheads;
+using OpenRA.Effects;
 using OpenRA.Primitives;
 using OpenRA.Traits;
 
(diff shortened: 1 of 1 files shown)


reaperrr:
Actually, now I remember we wanted to get rid of `DelayedAction`, so I guess I'll have to rewrite this at least partially.

pchote:
> This new property will delay actor spawn on death (useful when you need to play unit's death animation first without overlapping dead actor and new actor).

This is already handled, in principle, by the `SpawnActorOnDeath` trait.

GraionDilach:
`SkipMakeAnimations` skips the spawned actor's buildup - while the quote is about the dying actor's `WithDeathAnimation`.

pchote:
Based on the quote their use case is to have a transform animation between one actor type and another.  Our Transforms code, for better or worse, does this using a make animation.

GraionDilach:
So basically, YR Brutes should have buildup animations for the Genetic Mutator?

pchote:
How you implement RA2/YR mod code is up to you.  The current concern here is avoiding queuing a `DelayedAction` on the world (which is incompatible with save games), and using our existing code support is cleaner than having to write new redundant activity types.

reaperrr:
Another possibility would be to spawn a husk that uses the transition animation as sprite body sequence.

In any case, turns out the desired effect can actually be implemented already, so I'll close this.

PR #12290, opened 2016-10-23
[D2k] Added the 'Damage' UncloakType to several units

### LINKED ISSUE #12192: RA Phase Transport, TD Stealth tank & D2k Stealth Raider don't uncloak on damage

abc013:
Changed: Saboteur, Fremen and Stealth Raider. Closes #12192.

### DIFF
--- a/mods/d2k/rules/infantry.yaml
+++ b/mods/d2k/rules/infantry.yaml
@@ -158,6 +158,7 @@ fremen:
 	Cloak:
 		InitialDelay: 85
 		CloakDelay: 85
+		UncloakOn: Attack, Unload, Infiltrate, Demolish, Dock, Damage
 		CloakSound: STEALTH1.WAV
 		UncloakSound: STEALTH2.WAV
 		IsPlayerPalette: true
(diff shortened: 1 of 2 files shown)


pchote:
Is this consistent with the behaviour in the original game?

abc013:
@reaperrr wrote :
> IIRC, we want at least D2k to stay as true to the original as possible, so we might want to change it at least on the Stealth Raider, Fremen and possibly Saboteur.

So, should be.

reaperrr:
> Is this consistent with the behaviour in the original game?

I just checked, it is.

obrakmann (approved):
👍

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/3600dd0192ca45d300bc32fb3e44bc7c66a2eed0)

PR #12293, opened 2016-10-23
Fix null being passed through the INotifyAttack interface

abcdefg30:
[`SwallowActor` explicitly passes a `null` `Armament` to `INotifyAttack.PreparingAttack`](https://github.com/OpenRA/OpenRA/blob/634faa31deb9e03ed7b49b06c9b6ade0225bcb01/OpenRA.Mods.D2k/Activities/SwallowActor.cs#L96) which then gets redirected in [`WithInfantryBody`](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/Render/WithInfantryBody.cs#L128) to its `Attacking` method causing a crash.

I need the explicit `string sequence = null` as I otherwise get a `CS0165` compiler error.

Detected by Coverity.

### DIFF
--- a/OpenRA.Mods.D2k/Activities/SwallowActor.cs
+++ b/OpenRA.Mods.D2k/Activities/SwallowActor.cs
@@ -31,18 +31,22 @@ namespace OpenRA.Mods.D2k.Activities
 		readonly Sandworm sandworm;
 		readonly ConditionManager conditionManager;
 		readonly WeaponInfo weapon;
+		readonly Armament armament;
 		readonly AttackSwallow swallow;
 		readonly IPositionable positionable;
+		readonly IFacing facing;
 
 		int countdown;
 		CPos burrowLocation;
 		AttackState stance;
 		int attackingToken = ConditionManager.InvalidConditionToken;
 
-		public SwallowActor(Actor self, Target target, WeaponInfo weapon)
+		public SwallowActor(Actor self, Target target, Armament a, IFacing facing)
 		{
 			this.target = target;
-			this.weapon = weapon;
+			this.facing = facing;
+			armament = a;
+			weapon = a.Weapon;
 			sandworm = self.Trait<Sandworm>();
 			positionable = self.Trait<Mobile>();
 			swallow = self.Trait<AttackSwallow>();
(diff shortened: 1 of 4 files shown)


pchote:
Calling `INotifyAttack` with a null `Armament` is the real issue here.   Can we please fix the worm code instead?

If we really want to allow null-armaments then we're going to need to check and change a lot more than just this.

GraionDilach:
The worm code is a mess - it has RunAfterDelay in it even (borking ion nonnormal gamespeeds iow).

abcdefg30:
Updated. Now `null` isn't passed at all.

OpenRA.Mods.D2k/Activities/SwallowActor.cs
```diff
@@ -91,11 +95,13 @@ bool AttackTargets(Actor self, IEnumerable<Actor> targets)
 				}
 			});
 
+			var barrel = armament.CheckFire(self, facing, target);
```
pchote: The `WormJaw` defines a `SpreadDamage` warhead that will now be applied.  We'll want to remove that.

pchote (dismissed):

OpenRA.Mods.D2k/Activities/SwallowActor.cs
reaperrr: Removing the warhead makes the worm circle around targeted actors endlessly without ever swallowing them.

reaperrr:
Looks good to me, but I'm leaving the Fixup requested label in case I've misunderstood @pchote.

OpenRA.Mods.D2k/Activities/SwallowActor.cs
GraionDilach: Yes, the warhead is still used for targeting. So IMO it only needs the damage value removed (and maybe Falloff changed to 0,0).

abcdefg30:
Updated. (Wow, not so fast...)

mods/d2k/weapons/other.yaml
```diff
@@ -60,11 +60,11 @@ WormJaw:
...
 	Range: 1c512
-	Warhead@1Dam: SpreadDamage
+	Warhead@1Dam: SpreadDamage # HACK: The warhead is needed for targeting
 		InvalidTargets: Structure, Infantry
-		Spread: 768
-		Falloff: 100, 100, 0
-		Damage: 10000
+		Spread: 0
+		Falloff: 0,0
+		Damage: 0
```
reaperrr: Damage defaults to `0` anyway, so this line is redundant.

If we wait for #12605 to get merged, you should be able to use the `TargetDamage` warhead here and remove Spread + Falloff as well.

abcdefg30: Removed it. If we want to wait for #12605 is up to the reviewers. (I'm fine with either.)

reaperrr (approved):

mods/d2k/weapons/other.yaml
```diff
@@ -60,11 +60,10 @@ WormJaw:
 	ReloadDelay: 10
 	InvalidTargets: Structure, Infantry
 	Range: 1c512
-	Warhead@1Dam: SpreadDamage
+	Warhead@1Dam: SpreadDamage # HACK: The warhead is needed for targeting
```
pchote: It would be slightly less hacky if this used `TargetDamageWarhead`, but that can be done in #12605.

pchote (approved):

PR #12296, opened 2016-10-23
Fix a possible crash in the AssetsBrowserLogic

abcdefg30:
`errorLabelWidget` might be `null` (we check for `null` on all other occurences as well).

Detected by Coverity.


phrohdoh (approved):

abcdefg30:
Superseded by #12297.

PR #12297, opened 2016-10-23
Fix error label not updating and split message into multiple labels

phrohdoh:
~~Depends on #12296 being merged, or this can be merged and that closed.~~

`{filename}` was being replaced in the string upon first error so it did not exist later to be replaced again by any following errors.

screen shot 2016-10-23 at 3 50 32 pm

### DIFF
--- a/OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
@@ -356,12 +356,14 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 			{
 				if (errorLabel != null)
 				{
-					errorLabelFilename.GetText = () => WidgetUtils.TruncateText(filename, errorLabel.Bounds.Width, errorFont);
 					currentSprites = new Sprite[0];
 					errorLabel.Visible = true;
 
 					if (errorLabelFilename != null)
+					{
+						errorLabelFilename.GetText = () => WidgetUtils.TruncateText(filename, errorLabel.Bounds.Width, errorFont);
 						errorLabelFilename.Visible = true;
+					}
 				}
 
 				Log.AddChannel("assetbrowser", "assetbrowser.log");
--- a/OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
+++ b/OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
@@ -47,19 +47,10 @@ namespace OpenRA.Mods.Common.Widgets.Logic
 		bool isVideoLoaded = false;
 		int currentFrame;
 
-		LabelWidget errorLabelWidget;
+		LabelWidget errorLabel;
 		LabelWidget errorLabelFilename;
 		SpriteFont errorFont;
-		bool isErrorLabelVisible
-		{
-			get
-			{
-				return errorLabelWidget != null
-					&& errorLabelWidget.Visible
-					&& errorLabelFilename != null
-					&& errorLabelFilename.Visible;
-			}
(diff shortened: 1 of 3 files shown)


OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -48,8 +48,18 @@ public class AssetBrowserLogic : ChromeLogic
 		int currentFrame;
 
 		LabelWidget errorLabelWidget;
+		LabelWidget errorLabelFilename;
 		SpriteFont errorFont;
-		bool IsErrorLabelVisible { get { return errorLabelWidget != null && errorLabelWidget.Visible; } }
+		bool IsErrorLabelVisible
```
Phrohdoh: This isn't public so should be `isErrorLabelVisible`.
Adding `changes requested` label because of this.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -348,11 +364,10 @@ bool LoadAsset(string filename)
 			{
 				if (errorLabelWidget != null)
```
abcdefg30: You also need to `null` check `errorLabelFilename` somewhere around here.

Phrohdoh:
Do we actually need `ERROR_FILENAME` at this point?
This is only visible on the interface where the currently selected file is quite obvious.

Phrohdoh:
> OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs:L53: [SA1300] property names begin with an upper-case letter: isErrorLabelVisible.

😠

Phrohdoh:
@OpenRA/engine-hackers Opinions on changing the stylecop warning above (SA1300)?
I see no reason for a private member to be PascalCase, regardless of property, field, etc status.

obrakmann:
> Do we actually need ERROR_FILENAME at this point?

Not necessarily. I'd be fine with removing it. There are filenames that are too long to fit in the selection box on the left, though, as seen on your screenshot.

> I see no reason for a private member to be PascalCase, regardless of property, field, etc status.

Most of our properties usually are at least partly publicly accessible. That said, I'm fine with either.

Phrohdoh:
Fixed the case of the identifier.

OpenRA.Mods.Common/Widgets/Logic/AssetBrowserLogic.cs
```diff
@@ -346,13 +354,14 @@ bool LoadAsset(string filename)
 			}
 			catch (Exception ex)
 			{
-				if (errorLabelWidget != null)
+				if (errorLabel != null)
 				{
-					errorLabelWidget.Text = WidgetUtils.TruncateText(errorLabelWidget.Text.Replace("{filename}", filename),
-						errorLabelWidget.Bounds.Width, errorFont);
-
+					errorLabelFilename.GetText = () => WidgetUtils.TruncateText(filename, errorLabel.Bounds.Width, errorFont);
```
abcdefg30: You check for `null` everywhere else, so why not here as well?

abcdefg30:
Looks good to me. :+1: / :white_check_mark: after the nit has been addressed and the commits have been squashed.

reaperrr:
>  and the commits have been squashed.

pchote (changes requested):
A simpler way to do this is to put both LabelWidgets inside a ContainerWidget in the yaml, then you separate out your concerns of visibility (on the container) vs label content (on the label itself) and avoid having to link visibility of multiple widgets together.

pchote:
Superseded by #12646.

PR #12302, opened 2016-10-24
Restructuring make script

Jiyuu:
In addiction to restructuring I have also made the "all" command fetch dependencies.

I believe this structure is more readable and obvious, but if you think this structure is not proper for powershell LMK and I'll make a separate PR just for adding the dependencies to the "all" command.

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -1,52 +1,11 @@
-function FindMSBuild
-{
-	$msBuildVersions = @("4.0")
-	foreach ($msBuildVersion in $msBuildVersions)
-	{
-		$key = "HKLM:\SOFTWARE\Microsoft\MSBuild\ToolsVersions\{0}" -f $msBuildVersion
-		$property = Get-ItemProperty $key -ErrorAction SilentlyContinue
-		if ($property -eq $null -or $property.MSBuildToolsPath -eq $null)
-		{
-			continue
-		}
-		$path = Join-Path $property.MSBuildToolsPath -ChildPath "MSBuild.exe"
-		if (Test-Path $path)
-		{
-			return $path
-		}
-	}
-	return $null
-}
+####### The starting point for the script is the bottom #######
 
-function UtilityNotFound
+###############################################################
+########################## FUNCTIONS ##########################
+###############################################################
+function All-Command 
 {
-	echo "OpenRA.Utility.exe could not be found. Build the project first using the `"all`" command."
-}
-
-if ($args.Length -eq 0)
-{
-	echo "Command list:"
-	echo ""
-	echo "  all             Builds the game and its development tools."
-	echo "  dependencies    Copies the game's dependencies into the main game folder."
-	echo "  version         Sets the version strings for the default mods to the latest"
-	echo "                  version for the current Git branch."
-	echo "  clean           Removes all built and copied files. Use the 'all' and"
-	echo "                  'dependencies' commands to restore removed files."
-	echo "  test            Tests the default mods for errors."
-	echo "  check           Checks .cs files for StyleCop violations."
-	echo "  check-scripts   Checks .lua files for syntax errors."
-	echo "  docs            Generates the trait and Lua API documentation."
-	echo ""
-	$command = (Read-Host "Enter command").Split(' ', 2)
-}
-else
-{
-	$command = $args
-}
-
-if ($command -eq "all")
-{
+	Dependencies-Command 
 	$msBuild = FindMSBuild
 	$msBuildArguments = "/t:Rebuild /nr:false"
 	if ($msBuild -eq $null)
@@ -66,7 +25,8 @@ if ($command -eq "all")
 		}
 	}
 }
-elseif ($command -eq "clean")
+
+function Clean-Command 
 {
 	$msBuild = FindMSBuild
 	$msBuildArguments = "/t:Clean /nr:false"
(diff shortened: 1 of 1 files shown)


pchote:
Can some of our windows-based people please review this?

make.ps1
```diff
@@ -1,27 +1,5 @@
...
-		}
-	}
-	return $null
-}
-
-function UtilityNotFound
-{
-	echo "OpenRA.Utility.exe could not be found. Build the project first using the `"all`" command."
-}
+$functions = "$($MyInvocation.MyCommand.path | split-path)\make-functions.ps1"
```
pchote: I'm not convinced that the (minor, IMO?) readability improvement for shipping these into their own file outweighs the (minor) complexity of having a separate function file.  Wouldn't we get the same benefits by having all the functions at the top of the file?

Jiyuu: You will get most of the benefit, but it would take a few more seconds figuring out that the heart of the code is at the bottom (unless this is some kind of an agreed practice in powershell i am simply unaware of)

Basically I tried making the file as accessible as possible for someone opening it for the first time.


Though if there's strong agreement that its better not to have 2 files I don't mind combining them.

abcdefg30: I'd like to have them stay in one file.

Jiyuu:
I've combined the functions back into the original file and added some comments in the hope this will be clear enough.

Any comments?

make.ps1
```diff
@@ -225,20 +186,68 @@ elseif ($command -eq "docs")
...
+		{
+			return $path
 		}
-		Start-Sleep -Milliseconds 50
+	}
+	return $null
+}
+
+
+
```
abcdefg30: I think we can reduce this three new lines to one, and also add one new line below the `UtilityNotFound` function.

make.ps1
```diff
@@ -186,13 +143,16 @@ elseif ($command -eq "check")
 		./OpenRA.Utility.exe cnc --check-code-style OpenRA.Utility
 		echo "Checking for code style violations in OpenRA.Test..."
 		./OpenRA.Utility.exe cnc --check-code-style OpenRA.Test
+		echo "Checking for explicit interface violations..."
+		./OpenRA.Utility.exe all --check-explicit-interfaces
```
abcdefg30: Why is this moved around?

make.ps1
```diff
@@ -225,20 +186,68 @@ elseif ($command -eq "docs")
...
+	echo "  docs            Generates the trait and Lua API documentation."
+	echo ""
+	$command = (Read-Host "Enter command").Split(' ', 2)
+}
+else
+{
+	$command = $args
+}
+
+switch ($command) {
```
abcdefg30: Put the brace on a new line, please.

make.ps1
```diff
@@ -225,20 +186,68 @@ elseif ($command -eq "docs")
...
+	"all" { All-Command }
+	"dependencies" { Dependencies-Command }
+	"version" { Version-Command }
+	"clean" { Clean-Command }
+	"test" { Test-Command }
+	"check" { Check-Command }
+	"check-scripts" { Check-Scripts-Command }
+	"docs" { Docs-Command }
+	Default {
```
abcdefg30: Same here (or put everything on one line: `Default { echo (...) }`).

make.ps1
```diff
@@ -1,52 +1,10 @@
...
-	echo "  check-scripts   Checks .lua files for syntax errors."
-	echo "  docs            Generates the trait and Lua API documentation."
-	echo ""
-	$command = (Read-Host "Enter command").Split(' ', 2)
-}
-else
-{
-	$command = $args
-}
+####### The starting point for the script is the bottom #######
```
abcdefg30: Why actually? Couldn't we just have the main part at the top?

make.ps1
```diff
@@ -1,52 +1,10 @@
...
 
-if ($command -eq "all")
-{
+###############################################################
+########################## FUNCTIONS ##########################
+###############################################################
```
abcdefg30: This looks like a bit of overkill to me, but I don't really mind that much either. Maybe others have stronger opinions.

make.ps1
Jiyuu: Nope
Apparently Powershell doesn't support forward declaration, this is the reason I went with the separate file in the first place.

abcdefg30: Ok. I already feared that. This is fine then.

make.ps1
Jiyuu: not sure, i see that this is the exact opposite of the last commit before me, so either i used an old version by mistake, or alternatively i remember seeing some errors when i was testing all the commands so i may have fixed something and forgot about it. 

will return back for now

make.ps1
```diff
@@ -1,52 +1,10 @@
...
-{
-	$command = $args
-}
-
-if ($command -eq "all")
-{
+###############################################################
+########################## FUNCTIONS ##########################
+###############################################################
+function All-Command {
```
abcdefg30: The brace here also wants to go on a new line (likewise for `Clean-Command`, `Check-Command` and `Version-Command` below).

abcdefg30:
Looks good to me now (I'm even ok with adding the `dependencies` step to `all`). Just two things:
- The window now immediately closes instead of waiting for key input like it did before
- Please squash the commits

Jiyuu:
I restored the piece for keeping the window open when the script is called without any parameters

abcdefg30:
Thanks. 👍 / ✅

reaperrr:
> Looks good to me now (I'm even ok with adding the dependencies step to all)

Same here 👍

PR #12305, opened 2016-10-25
Replace DelayedAction in GpsWatcher with Tick-based solution

reaperrr:
This is the first of multiple `DelayedAction`-removing PRs.

GPS is probably amongst the most fragile pieces of code in OpenRA, so this gets its own PR in order to not block easier cases.

### DIFF
--- a/OpenRA.Mods.RA/Effects/GpsSatellite.cs
+++ b/OpenRA.Mods.RA/Effects/GpsSatellite.cs
@@ -18,14 +18,19 @@ namespace OpenRA.Mods.RA.Effects
 {
 	class GpsSatellite : IEffect
 	{
+		readonly Player launcher;
 		readonly Animation anim;
 		readonly string palette;
+		readonly int revealDelay;
 		WPos pos;
+		int tick;
 
-		public GpsSatellite(World world, WPos pos, string image, string sequence, string palette)
+		public GpsSatellite(World world, WPos pos, string image, string sequence, string palette, int revealDelay, Player launcher)
 		{
 			this.palette = palette;
 			this.pos = pos;
+			this.launcher = launcher;
+			this.revealDelay = revealDelay;
 
 			anim = new Animation(world, image);
 			anim.PlayRepeating(sequence);
@@ -36,8 +41,12 @@ namespace OpenRA.Mods.RA.Effects
 			anim.Tick();
 			pos += new WVec(0, 0, 427);
 
-			if (pos.Z > pos.Y)
+			if (++tick > revealDelay)
+			{
+				var watcher = launcher.PlayerActor.Trait<GpsWatcher>();
+				watcher.ReachedOrbit(launcher);
 				world.AddFrameEndTask(w => w.Remove(this));
+			}
 		}
 
 		public IEnumerable<IRenderable> Render(WorldRenderer wr)
--- a/OpenRA.Mods.RA/Effects/SatelliteLaunch.cs
+++ b/OpenRA.Mods.RA/Effects/SatelliteLaunch.cs
@@ -43,7 +43,7 @@ namespace OpenRA.Mods.RA.Effects
 			if (++frame == 19)
 			{
 				var palette = info.SatellitePaletteIsPlayerPalette ? info.SatellitePalette + launcher.Owner.InternalName : info.SatellitePalette;
-				world.AddFrameEndTask(w => w.Add(new GpsSatellite(world, pos, info.SatelliteImage, info.SatelliteSequence, palette)));
+				world.AddFrameEndTask(w => w.Add(new GpsSatellite(world, pos, info.SatelliteImage, info.SatelliteSequence, palette, info.RevealDelay * 25, launcher.Owner)));
 			}
 		}
 
--- a/OpenRA.Mods.RA/Traits/GpsWatcher.cs
+++ b/OpenRA.Mods.RA/Traits/GpsWatcher.cs
@@ -48,30 +48,26 @@ namespace OpenRA.Mods.RA.Traits
 		public void GpsRemove(Actor atek)
 		{
 			actors.Remove(atek);
-			RefreshGps(atek);
+			RefreshGps(atek.Owner);
 		}
 
 		public void GpsAdd(Actor atek)
 		{
 			actors.Add(atek);
-			RefreshGps(atek);
+			RefreshGps(atek.Owner);
 		}
 
-		public void Launch(Actor atek, GpsPowerInfo info)
+		public void ReachedOrbit(Player launcher)
 		{
-			atek.World.Add(new DelayedAction(info.RevealDelay * 25,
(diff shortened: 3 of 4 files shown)


OpenRA.Mods.RA/Traits/GpsWatcher.cs
```diff
@@ -26,14 +26,17 @@ class GpsWatcherInfo : ITraitInfo
 
 	interface IOnGpsRefreshed { void OnGpsRefresh(Actor self, Player player); }
 
-	class GpsWatcher : ISync, IFogVisibilityModifier
+	class GpsWatcher : ITick, ISync, IFogVisibilityModifier
 	{
 		[Sync] public bool Launched { get; private set; }
+		[Sync] public bool MissileFlying { get; private set; }
```
Phrohdoh: Please prefix this with `Is`

pchote:
A better fix would be to trigger this from the missile effect, which right now is only used for visuals.

reaperrr:
Updated.

GraionDilach (approved):

OpenRA.Mods.RA/Effects/GpsSatellite.cs
```diff
@@ -36,8 +41,12 @@ public void Tick(World world)
 			anim.Tick();
 			pos += new WVec(0, 0, 427);
 
-			if (pos.Z > pos.Y)
+			if (++tick >= revealDelay)
+			{
+				var owner = launcher.Owner.PlayerActor.Trait<GpsWatcher>();
```
obrakmann: If I'm not mistaken, this will change the behaviour subtly in the case where an engineer captures an atek while the GPS satellite is being launched. In this PR, the new owner will get the GPS, previously the old owner would have gotten it. Moving this line up into the constructor would restore the old behaviour.

obrakmann (commented):
Looks good to me otherwise, 👍

OpenRA.Mods.RA/Effects/GpsSatellite.cs
pchote: It would be even better if we could replace the `Actor launcher` ctor arg with a `Player owner`.  It looks like `GpsWatcher` doesn't care about the launcher for anything important, so this should hopefully be a simple change.

reaperrr:
Updated. I changed it from actor to player as @pchote suggested, so I guess this needs at least a partial re-review.

OpenRA.Mods.RA/Effects/GpsSatellite.cs
```diff
@@ -36,8 +41,12 @@ public void Tick(World world)
 			anim.Tick();
 			pos += new WVec(0, 0, 427);
 
-			if (pos.Z > pos.Y)
+			if (++tick >= revealDelay)
```
pchote: `++foo` increments the value before returning it, so this will run immediately if `revealDelay` is 1.  Is this what you intended?  I would normally expect this to look like either `if (tick++ >= revealDelay)` or `if (++tick > revealDelay)`.

OpenRA.Mods.RA/Effects/GpsSatellite.cs
```diff
@@ -36,8 +41,12 @@ public void Tick(World world)
...
+			{
+				var owner = launcher.PlayerActor.Trait<GpsWatcher>();
```
pchote: Naming nit: can we please call this `watcher`?

OpenRA.Mods.RA/Effects/SatelliteLaunch.cs
```diff
@@ -43,7 +43,7 @@ public void Tick(World world)
 			if (++frame == 19)
 			{
 				var palette = info.SatellitePaletteIsPlayerPalette ? info.SatellitePalette + launcher.Owner.InternalName : info.SatellitePalette;
-				world.AddFrameEndTask(w => w.Add(new GpsSatellite(world, pos, info.SatelliteImage, info.SatelliteSequence, palette)));
+				world.AddFrameEndTask(w => w.Add(new GpsSatellite(world, pos, info.SatelliteImage, info.SatelliteSequence, palette, info.RevealDelay * 25, launcher.Owner)));
```
pchote: Can we please change `info.RevealDelay` to be specified in ticks instead of "seconds measured at the default game speed"?  This isn't a merge blocker, but I'd really like to see it happen.

OpenRA.Mods.RA/Traits/GpsWatcher.cs
```diff
@@ -48,30 +48,26 @@ public GpsWatcher(Player owner)
 		public void GpsRemove(Actor atek)
 		{
 			actors.Remove(atek);
-			RefreshGps(atek);
+			RefreshGps(atek.Owner);
```
pchote: Can we please move this `.Owner` up a level? (i.e. pass the `Player launcher` into the method instead of `Actor atek`).

OpenRA.Mods.RA/Traits/GpsWatcher.cs
```diff
@@ -48,30 +48,26 @@ public GpsWatcher(Player owner)
...
 		}
 
 		public void GpsAdd(Actor atek)
 		{
 			actors.Add(atek);
-			RefreshGps(atek);
+			RefreshGps(atek.Owner);
```
pchote: Here too.

OpenRA.Mods.RA/Effects/GpsSatellite.cs
reaperrr: > Is this what you intended?

No, good catch.

pchote (dismissed):
A few minor nits:

OpenRA.Mods.RA/Traits/GpsWatcher.cs
reaperrr: How do I do the `actors.Remove` then?

pchote: Ah, good point.  I missed that, so nevermind.

OpenRA.Mods.RA/Effects/SatelliteLaunch.cs
reaperrr: Basically I agree and was thinking about doing that, but in my opinion it would be better to do *all* support power "seconds > ticks" conversions in one dedicated PR, I don't like the thought of having both cases mixed on bleed.

pchote: Ok, that's fair.

reaperrr:
Updated.

pchote (approved):

GraionDilach (approved):
:+1:

pchote:
https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/28c81b69be44535ce46c8778598f6a6216d8883a

PR #12307, opened 2016-10-25
Add plumbing for improved RA gap generator behaviour.

pchote:
The RA gap generators have never worked very well in our RA mod – our additions of the fog of war, initially explored map, and (until recently) GPS satellite dots made them relatively useless.

We had planned to make them more useful by actively blocking unit sight instead of just turning a circle of fog into shroud.  Not surprisingly... nobody wanted to actually code the feature, so it fell off the radar and other easier changes were made instead (changing "Disable Shroud" &rarr; "Explore Map", hiding GPS dots, etc).

~~#12267~~ #12308 added the last real prerequisite for this (supporting multiple `RevealsShroud` traits on an actor), so I feel this is the right time to try and push this feature through.

This reorganizes the internal `Shroud` plumbing to reduce duplication and adds a new "passive" reveal type that has lower precedence than generated shroud.  ~~The RA mod rules are then updated to give all actors half their normal vision range for gap-generated cells.~~

The effect ingame looks like this (gap generator is to the left of the tank):
gap2


AoAGeneral:
I actually like the idea. The Gaps in RA96 were sorta meh over all as well but had some usefulness. I think this is a good idea but I do have a few questions:

Does this also affect aircraft? If so how would AA guns work with them near Gaps? Would defense structures become stronger or should another unit be able to extend range for situations as this? (IE: Frances mobile gap generator ignoring gap shroud affects. Radar unit ignore gap shroud affects while also useful of its redirective missile mechanic.)

pchote:
Right now aircraft also lose half their vision range, but this can be tweaked for each actor based on future testing and balancing.  I think it would be an interesting throwback to the original game if gap generators blocked *all* aircraft vision, but I doubt we could make that balanced.

AoAGeneral:
Personally with how aircraft are used in RA this could work actually blocking vision. It would enable situations to use other units that are somewhat less popular (IE: MAD tanks, mobile gaps, etc). As you said tweaking will be needing but it does have promise. Just that concern on the defenses ^_^

pchote:
Rebased.

Micr0Bit:
:+1:  this is ready for the next playtest

SoScared:
Great work! It's very exciting to have a functional tweak for the redundant gap generator in a way that pays homage to the original game. 

Balance wise I think it will take quite some time to figure things out. Vision range is the most important property for engaging battle and for making quick decitions and we've seen the effect of even small vision range nerfs or buffs of units with some balance changes. As it stands above my best guess is it would break the game in a lot of situations given that target firing and planned approach would more or less go out the window with the gap generator. Customizing unit specific effects seems like an absolute must. In this regard I think it's worth posting this topic on the forum and have a proper brainstorming on it.

obrakmann (approved):
👍 The code looks fine, and I didn't find any regressions. The effect looks a bit weird with actors whose reveal range is larger than the gapgen's range:

screenshot from 2016-10-31 19 51 01

SoScared:
I've been fiddling a bit with the values and wanna pitch an early soft alternative setup:


Infantry: No effect.

Vehicles: Max 4 cells vision range, except no effect on Mobile Radar Jammer, Mammoth Tank.

Ships: Max 4/5 cells vision range.

Aircraft: 2 cells reduction on all aircrafts.

Structures/Defence: Max 5 cells vision range.


Might not look like much at first glance but given the advantage of shooting first in any engagement on RA my guess is the effect will play out significantly. France's Mobile Gap Generator also comes to mind. A few cells reduction of vision range for defensive structures would make base pushes harder to pull off. As for aircraft, cutting the vision range in half would probably be brutal, giving the opponent 0 chances to identify targets withing the sphere of the Gap Generator. 2 cells vision range reduction (10, down from 12) enables MiGs/Longbows to barely get a glimpse of an AA Gun placed within it. Yak/Hinds still stand no chance with the reduction (8, down form 10) but ought to have some way of peeking into the gap shroud. The last points hinges somewhat on #12330

Micr0Bit:
we didnt even have a single playtest with it , as it is ... thinking about balance at this point (when nobody even tested it) is kinda of prematurely deforming a feature which could add potentially new strategies

SoScared:
I'm not gonna stand in the way of developing new strategies. Since this is primerily my own concern I'd love to help starting up a few rogue playtest cycles (bleed) on this feature 'as is'.

abcdefg30:
> The effect looks a bit weird with actors whose reveal range is larger than the gapgen's range:

Imho we should sort that out.

pchote:
I'm not sure how we would sort it out: it's kind of the whole point of this feature!

abcdefg30:
Why should actors be able to look beyond shroud? That doesn't make much sense for me...

pchote:
You're asking to have active shroud generation block visibility outside the range of the shroud generation.  We would need to rewrite a lot of our shroud code (way more than just this pr) to make that work.

OpenRA.Game/Traits/World/Shroud.cs
```diff
@@ -104,6 +124,17 @@ void INotifyCreated.Created(Actor self)
 
 		void Invalidate(IEnumerable<PPos> changed)
 		{
+			foreach (var puv in changed)
+			{
+				var uv = (MPos)puv;
+				var type = ShroudCellType.Shroud;
+
+				if (explored[uv] && (!shroudGenerationEnabled || generatedShroudCount[uv] == 0 || visibleCount[uv] > 0))
+					type = visibleCount[uv] > 0 ? ShroudCellType.Visible: ShroudCellType.Fog;
```
abcdefg30: Missing space between `Visible` and `:`.

abcdefg30: Nvm, that get's reverted in following commit.

abcdefg30: Nvm, that gets reverted in following commit.

abcdefg30:
Unfortunately I don't have any good answers to this either (except working things around using yaml "hacks"). My problem is that we got a (somewhat) working state before and this PR (although it fixes things) leaves a bogus one behind. I'd probably be stupid to block it because of this though.

pchote:
@abcdefg30: If the final balance sets the active reveal range the same as the generation range then this won't be a problem.

pchote:
I don't have the enthusiasm to ~~herd cats~~ deal with balance arguments, so I have removed the commit that enables this in RA.  Someone else can deal with the balance issues.

pchote:
9aa77932ba1909d84a1d94428fea358aa730f449 is the commit with the rule changes for anyone who wants to work on the balance implementation or test this PR.

reaperrr (approved):

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/c7a3996beabf0ec671857a615f83d4c161d2df2d...5f5c633ac4204b8b85698d4118924a40000dc6d5)

hypercube33:
Sooo...why doesnt this just hide whats under the gap gen, and only reveal terrain? Why the game breaking bugs induced by this change that didnt improve any gameplay? Maybe fix the horribly broken stuff like planes landing instead?

reaperrr:
@hypercube33 "Add plumbing = add support for it, but don't enable it".

If you noticed any regressions from this change, file a bug report.

PR #12156, opened 2016-10-02
Support subtraction of two CPos in Lua

cjshmyr:
Fixes #11667, based off of [WPos already supporting this](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/WPos.cs#L93). I also tweaked the exception text as well (if this is fine, WPos could get updated too).

For testing I dropped this in the WorldLoaded function on a scripted map:

```
local pos1 = CPos.New(10, 10)
local vec1 = CVec.New(5, 5)
local a = pos1 - vec1
Media.DisplayMessage("CPos{10,10} - CVec{5,5} >> expected {5,5}, actual {" .. a.X .. "," .. a.Y .. "}")

local pos2 = CPos.New(10, 10)
local pos3 = CPos.New(5, 5)
local b = pos2 - pos3
Media.DisplayMessage("CPos{10,10} - CPos{5,5} >> expected {5,5}, actual {" .. b.X .. "," .. b.Y .. "}")
```

PR #12317, opened 2016-10-29
Fix Lua error message for WPos subtraction

cjshmyr:
This updates the Lua error message shown if a script screws up subtracting a WPos, and also fixes one of the `string.Format` parameters. Is similar to how the CPos message was changed in #12156.


abcdefg30:
👍

pchote (approved):

PR #12318, opened 2016-10-29
ScrollPanelWidget - Fix frame end task race condition

### LINKED ISSUE #12271: Crash when another player joins a server lobby as a game is starting.
Reported in the ingame IRC channel:

```
04:40 < pchote> what were you doing when it crashed?
04:40 < OMnom> besides clicking the start button?
04:41 < pchote> ah, so this happened when starting a game from the lobby?
04:41 < OMnom> yeah i was in teh lobby, i pressed ready, game crashed
04:41 < OMnom> i did see that someone snuck into the game, and then the game started immediately afterwards
```

```
Red Alert Mod at Version release-20161019
on map 5f58c7df7d0c9f3650a204a31a2243c85c2b8971 (Northwest Passage by Raymundo).
Date: 2016-10-21 03:32:59Z
Exception of type `System.NullReferenceException`: Object reference not set to an instance of an object.
   at OpenRA.Mods.Common.Widgets.ScrollPanelWidget.<BindingRefresh>m__1()
   at OpenRA.Primitives.ActionQueue.PerformActions(Int64 currentTime)
   at OpenRA.Game.LogicTick()
   at OpenRA.Game.Loop()
   at OpenRA.Game.Run()
```

cjshmyr:
This fixes #12271, which seems to be caused by a race condition. The global chat history or nickname panel has a chance to pass in `null` for `Game.GlobalChat.History` or `Game.GlobalChat.Users` during binding. What this will do is set `collection = null`, and `BindingRefresh()` would get caught with an NRE. What's interesting to me is it seemed like:
- Event bindings on `collection` are being cleared.
- `Collection` is being set null.
- For some reason, the event `BindingRefresh` still gets called, even though this event is not bound if `c == null` (I am assuming RunAfterTick causes this behavior).

I didn't go with adding a null check inside of `BindingRefresh` since it felt like a symptom of the problem, and I'm starting to get familiar with the scroll panel widget.

### DIFF
--- a/OpenRA.Mods.Common/Widgets/ScrollPanelWidget.cs
+++ b/OpenRA.Mods.Common/Widgets/ScrollPanelWidget.cs
@@ -374,9 +374,15 @@ namespace OpenRA.Mods.Common.Widgets
 			});
 		}
 
-		void BindingAdd(object item)
+		void BindingAdd(IObservableCollection col, object item)
 		{
-			Game.RunAfterTick(() => BindingAddImpl(item));
+			Game.RunAfterTick(() =>
+			{
+				if (collection != col)
+					return;
+
+				BindingAddImpl(item);
+			});
 		}
 
 		void BindingAddImpl(object item)
@@ -393,30 +399,40 @@ namespace OpenRA.Mods.Common.Widgets
 				ScrollToBottom();
 		}
 
-		void BindingRemove(object item)
+		void BindingRemove(IObservableCollection col, object item)
 		{
 			Game.RunAfterTick(() =>
 			{
+				if (collection != col)
+					return;
+
 				var widget = Children.FirstOrDefault(w => widgetItemEquals(w, item));
 				if (widget != null)
 					RemoveChild(widget);
 			});
 		}
 
-		void BindingRemoveAt(int index)
+		void BindingRemoveAt(IObservableCollection col, int index)
 		{
 			Game.RunAfterTick(() =>
 			{
+				if (collection != col)
+					return;
+
 				if (index < 0 || index >= Children.Count)
 					return;
+
 				RemoveChild(Children[index]);
 			});
 		}
 
-		void BindingSet(object oldItem, object newItem)
+		void BindingSet(IObservableCollection col, object oldItem, object newItem)
 		{
 			Game.RunAfterTick(() =>
 			{
+				if (collection != col)
+					return;
+
 				var newWidget = makeWidget(newItem);
 				newWidget.Parent = this;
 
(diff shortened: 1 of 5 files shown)


pchote:
The null collection will be coming from the `Unbind` method just above.
This change means that, whatever the real bug is, will be running on a stale collection instead of the unbound (null) collection.  This has a good chance of pushing the bug somewhere else.

I'd like to try and identify the race condition that is happening here, because that will hopefully inform a safer fix.

pchote:
I've left a comment in #12271 explaining the race condition and the crash.  This fix hides the crash, but unfortunately doesn't fix the problem.

> I suspect that the best fix here will be to add `collection` arguments to the `ScrollPanelWidget.Binding*` methods, and then check that `this.collection == collection` at the start of the frame end tasks that they queue.  If they don't match then someone has changes the bindings out from under them and they should immediately return.

cjshmyr:
With the changes requested (if I'd done them right) this isn't working quite right yet. It's somehow keeping the old widgets around with `null` collections and firing events on those.

pchote:
The changes look like what I had in mind, so I don't know off the top of my head what is going wrong here.

OpenRA.Mods.Common/Widgets/ScrollPanelWidget.cs
```diff
@@ -347,11 +347,11 @@ public void Bind(IObservableCollection c, Func<object, Widget> makeWidget, Func<
 			{
 				if (collection != null)
 				{
-					collection.OnAdd -= BindingAdd;
-					collection.OnRemove -= BindingRemove;
-					collection.OnRemoveAt -= BindingRemoveAt;
-					collection.OnSet -= BindingSet;
-					collection.OnRefresh -= BindingRefresh;
+					collection.OnAdd -= (item) => BindingAdd(item, collection);
```
pchote: Derp!  This creates a *new* anonymous delegate and then trys (and failed) to remove it.  The original bindings are therefore never removed.

pchote: Fixing this should be easy.  Change the `IObservableCollection` implementations to pass the second argument (`this` in that context) instead of DIYing it here.

pchote:
Would you mind if I took this over?

cjshmyr:
Nope, feel free

pchote:
I have pushed a fixed version of this to `pchote/fix-12271`, but I don't remember how to reproduce either of the original problem or the problem with this PR (I don't see anything wrong when leaving and rejoining the global lobby).  If someone could remind me of the repro cases then i'll test them and force-push my version on top of this PR.

pchote:
Repro case for #12271 is to add `Game.GlobalChat.Disconnect();` immediately before the `startGame()` call in the `START_GAME_BUTTON` handler in LobbyLogic.cs.  Start a skirmish, join the global chat, then press the start game button.

I've confirmed that this PR with my event binding fix fixes the crash.

pchote (approved):

RoosterDragon (approved):

RoosterDragon:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/be5f86738d6ce3bf886d7be06a1bf902abb5a110)

Issue #11265, opened 2016-05-11
Sometimes units will move rather than attack a camo pill under fog


AMHOL:
You can see this on [this replay](http://www.gamereplays.org/openra/replays.php?game=92&show=details&id=311530) just after 11:15.

This only seems to happen sporadically and apparently happens in other instances too (not just with camo pills)

Micr0Bit:
i have i noticed that i cant just attack "frozen" camopillboxes under fog ... the attack-cursos doesnt appear ... but still i see the frozen camo ... and "force fire" is also not allowed ... 

the only thing i can do , is shoot on a cell near by and hope to kill it with damage "spreading" over

obrakmann:
@Micr0Bit: with what kind of unit does that happen? Worked fine for e1 infantry and yaks in my tests:
![screenshot from 2016-05-14 00 41 27](https://cloud.githubusercontent.com/assets/4331210/15263715/b8a499e6-196c-11e6-8a9d-8ac7a829e2ac.png)

A difference is that the tooltip isn't shown for a cloakable frozen actor, which is a bug.

Micr0Bit:
@obrakmann : do they actually attack when you right-click on it ?

Artillery cant attack a frozen-camo ... you see the attack-cursor and the attack-line ... but it doesnt attack (i just checked it again)

obrakmann:
@Micr0Bit: yup, you're right, they don't. Sorry, I was focusing on you saying that the attack cursor wouldn't appear.

Micr0Bit:
@obrakmann : yeah my bad

pchote:
The problem will be that the frozen actor targeting will be handing off to the real underlying actor.  If the real actor is currently cloaked then the targeting will be rejected.

To fix this properly we will need to implement proper targeting handling for frozen actors, which is something i've been pushing for for years.  We've had at least one attempt, from @obrakmann, to implement this, but I couldn't find the pr/branch.

AMHOL:
I think a possible solution to this would be for structures not to re-cloak once revealed, the replay didn't save but there was an instance where someone had a tanya pillbox and I had to keep running infantry at it to reveal and kill it before I could move in. I guess this wouldn't be so much of a problem if force-fire worked but I think it makes sense for structures to not re-cloak anyway.

pchote:
There doesn't seem to be much interest in fixing this, so i'm not prepared to delay the next release over it.  Removing from milestone.

[...]

PR #12333, opened 2016-11-01
Prevent attacking units from knowing that a frozen building is dead

tohava:
Prevent units attacking a frozen building from magically knowing whether it is destroyed or not. (Note: I'm aware that this patch is not perfect yet. If you will tell me what else needs to be fixed, I will fix it).

### DIFF
--- a/OpenRA.Game/Traits/Target.cs
+++ b/OpenRA.Game/Traits/Target.cs
@@ -72,6 +72,10 @@ namespace OpenRA.Traits
 					if (actor.Generation != generation)
 						return TargetType.Invalid;
 				}
+				else if (type == TargetType.FrozenActor && !frozen.Visible)
+				{
+					return TargetType.Invalid;
+				}
 
 				return type;
 			}
--- a/OpenRA.Mods.Common/Activities/Attack.cs
+++ b/OpenRA.Mods.Common/Activities/Attack.cs
@@ -78,10 +78,11 @@ namespace OpenRA.Mods.Common.Activities
 				return NextActivity;
 
 			var type = Target.Type;
-			if (!Target.IsValidFor(self) || type == TargetType.FrozenActor)
+			if (!Target.IsValidFor(self))
 				return NextActivity;
 
-			if (attack.Info.AttackRequiresEnteringCell && !positionable.CanEnterCell(Target.Actor.Location, null, false))
+			var positionableLoc = self.World.Map.CellContaining(Target.CenterPosition);
+			if (attack.Info.AttackRequiresEnteringCell && !positionable.CanEnterCell(positionableLoc, null, false))
 				return NextActivity;
 
 			// Drop the target if it moves under the shroud / fog.
--- a/OpenRA.Mods.Common/ActorExts.cs
+++ b/OpenRA.Mods.Common/ActorExts.cs
@@ -80,24 +80,7 @@ namespace OpenRA.Mods.Common
 			// Flashes the frozen proxy
 			self.SetTargetLine(frozen, targetLine, true);
 
-			// Target is still alive - resolve the real order
-			if (frozen.Actor != null && frozen.Actor.IsInWorld)
-				return Target.FromActor(frozen.Actor);
-
-			if (!order.Queued)
-				self.CancelActivity();
-
-			var move = self.TraitOrDefault<IMove>();
-			if (move != null)
-			{
-				// Move within sight range of the frozen actor
-				var sight = self.TraitOrDefault<RevealsShroud>();
-				var range = sight != null ? sight.Range : WDist.FromCells(2);
-
-				self.QueueActivity(move.MoveWithinRange(Target.FromPos(frozen.CenterPosition), range));
-			}
-
-			return Target.Invalid;
+			return Target.FromFrozenActor(frozen);
 		}
 
 		public static void NotifyBlocker(this Actor self, IEnumerable<Actor> blockers)
--- a/AUTHORS
+++ b/AUTHORS
@@ -103,6 +103,7 @@ Also thanks to:
     * Nukem
     * Okunev Yu Dmitry (xaionaro)
     * Olaf van der Spek
+    * Ori Bar (tohava)
     * Paolo Chiodi (paolochiodi)
     * Paul Dovydaitis (pdovy)
     * Pavlos Touboulidis (pav)


tohava:
This bug is referenced in forum post http://www.sleipnirstuff.com/forum/viewtopic.php?p=298100&highlight=#298100 . If there's an issue for it as well, I'd be glad if someone can post it.

OpenRA.Mods.Common/ActorExts.cs
```diff
@@ -73,31 +80,15 @@ public static Target ResolveFrozenActorOrder(this Actor self, Order order, Color
...
-			{
-				// Move within sight range of the frozen actor
-				var sight = self.TraitOrDefault<RevealsShroud>();
-				var range = sight != null ? sight.Range : WDist.FromCells(2);
-
-				self.QueueActivity(move.MoveWithinRange(Target.FromPos(frozen.CenterPosition), range));
-			}
-
-			return Target.Invalid;
+			return Target.FromPos(frozen.CenterPosition);
```
pchote: You can avoid all the `uint frozenId` argument changes by using the FrozenActor target type.
The simplest way to do that is to use `Target.FromFrozenActor` here, but you could in principle fold this logic into the `Target.FromOrder` above.

This had been the plan all along, but it wasn't implemented originally because the attack code of the time was a mess that couldn't easily be adapted to work with frozen actors.  We've rewritten most of the problematic parts since then, so you should just need to change your `frozenId` checks to `t.Type == TargetType.FrozenActor` now.

pchote (changes requested):
This looks like a promising start!  I'd like to see this folded in to our existing plumbing as outlined below:

pchote:
This should fix #11265.

tohava:
@pchote - I have updated the pull request like you asked.

OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
```diff
@@ -296,6 +296,14 @@ public WDist GetMaximumRangeVersusTarget(Target target)
 		// Enumerates all armaments, that this actor possesses, that can be used against Target t
 		public IEnumerable<Armament> ChooseArmamentsForTarget(Target t, bool forceAttack)
 		{
+			// If the target is frozen, make sure it is still valid
+			if (t.Type == TargetType.FrozenActor)
+			{
+				var frozenLayer = self.Owner.PlayerActor.TraitOrDefault<FrozenActorLayer>();
+				if (frozenLayer == null || frozenLayer.FromID(t.FrozenActor.ID) == null)
```
pchote: You should be able to use `t.FrozenActor` directly here, instead of querying it again from the layer?

OpenRA.Mods.Common/Activities/Attack.cs
```diff
@@ -78,7 +78,7 @@ protected virtual Activity InnerTick(Actor self, AttackBase attack)
 				return NextActivity;
 
 			var type = Target.Type;
-			if (!Target.IsValidFor(self) || type == TargetType.FrozenActor)
+			if (!Target.IsValidFor(self))
 				return NextActivity;
 
 			if (attack.Info.AttackRequiresEnteringCell && !positionable.CanEnterCell(Target.Actor.Location, null, false))
```
pchote: I haven't tested it, but I expect this will crash because `Target.Actor` is null for `TargetType.FrozenActor`.  The rest of this method will need to be updated to deal with the frozen actor's `ActorInfo`.

OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
tohava: I am querying it to make sure that it is still alive. When the units explored the fog and removed the frozen actor, it's object still exists, but is no longer in the layer, this state is what I'm trying to get with the "if".

tohava: Your request is problematic. Let's say that target belongs to player B, and it is attacked by players A and C (they are not allies, it's ffa). Let's say that A managed to expose the shroud and see that the target already died, but C didn't, then the target should be `TargetType.Invalid` for A but `TargetType.FrozenActor` for C. I do not think that Target objects should also contain information about who is attacking them, am I correct? 

Do you have another solution or can I leave this code as it is?

tohava: @pchote - I believe I fixed this (it seems like there was only one place that needed fixing in the method, the rest either check the type before accessing the actor field, or use `CenterPosition`)

tohava:
@pchote - I fixed your second comment, but I have problems with the first.

OpenRA.Mods.Common/Activities/Attack.cs
tohava: @pchote - I believe I fixed this (it seems like there was only one place that needed fixing in the method, the rest either check the type before accessing the actor field, or use CenterPosition)

OpenRA.Game/Traits/Target.cs
```diff
@@ -133,6 +133,24 @@ public WPos CenterPosition
...
+				switch (Type)
+				{
+					case TargetType.Actor:
+						return actor.Location;
+					case TargetType.FrozenActor:
+						return frozen.Location;
+					default:
+					case TargetType.Invalid:
+					case TargetType.Terrain:
+						throw new InvalidOperationException("Attempting to query the location of a Target which is neither Actor nor FrozenActor");
```
pchote: The target code was specifically set up so that you could safely query the target `CenterPosition` even if the original target was killed (`TargetType.Actor` &rarr; `TargetType.Invalid`).  I'd like to avoid introducing different behaviour for this.

pchote: `AttackRequiresEnteringCell` is a hack (one of many) for the D2K sand worm.  We already know that the sandworm code needs a complete rewrite, so it would be better to keep the hacks there than to add (hopefully temporary) workarounds into the engine code.

OpenRA.Mods.Common/Activities/Attack.cs
```diff
@@ -78,10 +78,10 @@ protected virtual Activity InnerTick(Actor self, AttackBase attack)
 				return NextActivity;
 
 			var type = Target.Type;
-			if (!Target.IsValidFor(self) || type == TargetType.FrozenActor)
+			if (!Target.IsValidFor(self))
 				return NextActivity;
 
-			if (attack.Info.AttackRequiresEnteringCell && !positionable.CanEnterCell(Target.Actor.Location, null, false))
+			if (attack.Info.AttackRequiresEnteringCell && !positionable.CanEnterCell(Target.Location, null, false))
```
pchote: A simpler fix to the sand worm issue will be to use `positionable.CanEnterCell(self.World.Map.CellContaining(Target.CenterPosition), null, false)`.

OpenRA.Mods.Common/Traits/Attack/AttackBase.cs
pchote: The `Target` is created in `ResolveFrozenActorOrder` and then used for this actor only through all the attack code.  The `FrozenActorLayer` used here and in `ResolveFrozenActorOrder` will be the same so long as the owner hasn't changed between the target creation and now.

We are *supposed* to be invalidating the actor's target when its owner changes, so don't need to deal with the more complicated case.  I can see where we do this for `AttackFollow` and subclasses, but not for the others.  If we really aren't cancelling targets when the owner changes for e.g. an infantry in any mod (`AttackFrontal`), or the TD obelisk (`AttackOmni`) then that's a bug that will need to be fixed (in its own pr, not here!).

tohava:
@pchote - I have done as you asked, but I realized there is another problem with the code. When an attack is made against a visible actor, it will remain so, even when the actor has become frozen. Also, if an attack is made against a frozen actor, it will be stopped once the actor becomes visible. 

My idea on how to fix this is by modifying `Attack.InnerTick`:
1) Remove the hack code `if (!attack.Info.IgnoresVisibility && type == TargetType.Actor && !Target.Actor.Info.HasTraitInfo<FrozenUnderFogInfo>() && !self.Owner.CanTargetActor(Target.Actor)`
2) When a target becomes invalid, do one of the following:
2.1) If the target was an `Actor` with `FrozenUnderFog` and the attacked area is now under fog and there is a new `FrozenActor` at the position where the original `Target` was, create a new `AttackActivity` for it instead.
2.2) If the target was a `FrozenActor` and the attacked area is now NOT under fog and there is a new `Actor` at the position where the original `Target` was and the `Actor` also has `FrozenUnderFog` and is of the same building type as the `FrozenActor`, attack it.

Note: I am not sure if I will be able to continue on this since I am entering a busy schedule irl now. I'm leaving this information here for either me or someone else to continue at a later time.

pchote:
Ah right, that is indeed a problem.  Your idea sounds like the best way forward on this.

pchote:
I'll close this for now to clean up the PR queue, but i'd like to pick this up again in the medium-term (#11265 is a serious release blocker for the TS mod).  Adding to the future milestone so that I can find it again.

PR #12334, opened 2016-11-01
Added distinction between RA minelayers

roundowl:
Changed tooltip for allied and soviet minelayer, making clear which one lays anti-personnel or anti-tank mines. In "release-20161019", two minelayers look the same in the menu, which creates problems if you capture a factory from opposing faction.

### DIFF
--- a/mods/ra/maps/allies-03a/rules.yaml
+++ b/mods/ra/maps/allies-03a/rules.yaml
@@ -136,7 +136,7 @@ MCV:
 	Buildable:
 		Prerequisites: ~disabled
 
-MNLY.AP:
+MNLY:
 	Buildable:
 		Prerequisites: ~disabled
 
--- a/mods/ra/maps/allies-05a/rules.yaml
+++ b/mods/ra/maps/allies-05a/rules.yaml
@@ -161,7 +161,7 @@ MCV:
 	Buildable:
 		Prerequisites: ~disabled
 
-MNLY.AP:
+MNLY:
 	Buildable:
 		Prerequisites: ~disabled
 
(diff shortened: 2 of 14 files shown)


pchote:
I thought they both laid the same kind of mine (except for artwork)?

roundowl:
One lays ATMine (Explosions: large_explosion; DamageTypes: Prone50Percent, TriggerProne, DefaultDeath), another lays APMine (Explosions: napalm; DamageTypes: Prone50Percent, TriggerProne, SmallExplosionDeath). Damage value is the same. Maybe they are the same - then both minelayers can be merged into one.
Original problem is duplicate minelayer in production menu.

MustaphaTR:
In OpenRA they are almost same. In original game there is an AP-AT difference.

roundowl:
https://github.com/OpenRA/OpenRA/pull/4827 is the reason mines are equal. If it is still valid, then this PR needs to be closed. Then, maybe, minelayers (and mines) should be merged into one unit, to remove duplicate if both war factories are under control.

Phrohdoh:
This PR could do the merging, if you are up for that @roundowl.

MustaphaTR:
I think we would need a way to make it lay different mines (artwork wise) by side. Also looks like there is now duplicate tags for MNYL under missions which had both disabled.

roundowl:
Just woke up and fixed it, didn't notice it yesterday.
About different mines: there was talk about adding features to an Engineer. Maybe he can place true AP mines (low damage, low effect against armor).

GraionDilach:
This will need squashing and I'd also prefer the side-specific mine images - look at D2k buildings to see how can that be coded.

reaperrr:
Superseded by #12540.

PR #12337, opened 2016-11-01
Make friendly units aware of friendly mines

roundowl:
In release-20161019, in RA, friendly units don't mind going over mines placed by friendly minelayers.
This change allows friendly units to route around friendly mines, while still going over invisible enemy mines and blowing up.

### DIFF
--- a/OpenRA.Mods.RA/Traits/Mine.cs
+++ b/OpenRA.Mods.RA/Traits/Mine.cs
@@ -19,6 +19,7 @@ namespace OpenRA.Mods.RA.Traits
 	{
 		public readonly HashSet<string> CrushClasses = new HashSet<string>();
 		public readonly bool AvoidFriendly = true;
+		public readonly bool BlockFriendly = true;
 		public readonly HashSet<string> DetonateClasses = new HashSet<string>();
 
 		public object Create(ActorInitializer init) { return new Mine(this); }
@@ -52,6 +53,9 @@ namespace OpenRA.Mods.RA.Traits
 
 		bool ICrushable.CrushableBy(Actor self, Actor crusher, HashSet<string> crushClasses)
 		{
+			if (info.BlockFriendly && !crusher.Info.HasTraitInfo<MineImmuneInfo>() && self.Owner.Stances[crusher.Owner] == Stance.Ally)
+				return false;
+
 			return info.CrushClasses.Overlaps(crushClasses);
 		}
 	}


pchote:
We used to do this, but then changed to the current system for reasons that I no longer remember (but which were good at the time).  I suggest somebody revisits that history and makes sure it is no longer relevant as part of reviewing this.

roundowl:
I see, minelayers do not go over friendly mines, either. Plus there's no way to force units to enter a tile with a mine (not even with Alt pressed down).

roundowl:
Reverted that change, added change based on Mine trait from RA. Minelayer passes over it's mines, friendly units route around them, enemy soldiers blow up on them.

Also, I remember that original RA had a way to force movement over friendly crushables (including soldiers) by holding ALT key while issuing command. Is that planned?

mods/ra/rules/misc.yaml
```diff
@@ -1,5 +1,5 @@
 MINP:
-	Inherits: ^Mine
+	Inherits: ^Mine
```
abcdefg30: This should be reverted.

OpenRA.Mods.RA/Traits/Mine.cs
```diff
@@ -52,6 +52,9 @@ void INotifyCrushed.OnCrush(Actor self, Actor crusher, HashSet<string> crushClas
 
 		bool ICrushable.CrushableBy(Actor self, Actor crusher, HashSet<string> crushClasses)
 		{
+			if (!crusher.Info.HasTraitInfo<MineImmuneInfo>() && self.Owner.Stances[crusher.Owner] == Stance.Ally && info.AvoidFriendly)
```
abcdefg30: This shouldn't be needed because of https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.RA/Traits/Mine.cs#L43 ?

roundowl: If I understand correctly, INotifyCrushed.OnCrush happens when unit is already on the tile with a mine. ICrushable.CrushableBy is checked prior to unit entering a tile, preventing friendly units from moving over mines - that's why I added it. One of the lines should be removed, but I think it needs to be the one in OnCrush - to avoid situations where one unit chases another over a minefield, and only one of them blows up, just because mines are friendly.
Should I remove that one, to complete PR?

abcdefg30: Depends on wether we want friendly units to be able to enter the cell containing a mine or not.
However, this should use another bool than `AvoidFriendly` (maybe `BlockFriendly`?) if we want it.

mods/ra/rules/defaults.yaml
```diff
@@ -831,6 +831,7 @@
 		CrushClasses: mine
 		DetonateClasses: mine
 		AvoidFriendly: false
+		BlockFriendly: true
```
abcdefg30: You won't need to explicitely define this, as `true` is the default value (you set [here](https://github.com/OpenRA/OpenRA/pull/12337/files#diff-4fa252f4c4f2398df3d9099b8b8ee37bR22)).

roundowl: I thought it would be useful to have booleans related to mine behaviour in one place...

abcdefg30: Yes, please remove the yaml.

abcdefg30:
Looks good to me now. Please squash the commits.

roundowl:
Yay, I did a thing.

abcdefg30:
👍 / ✅

reaperrr:
👍

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/ad60e659907c548723302a211877c5e30a660595...85681c8eddb5d04ffac40140c4840e45d958221a)

Micr0Bit:
#12938

SoScared:
I just wanna plant a red flag on this issue. So far players have engaged the playtest minimally but from what little I've seen so far (very little) the path-around friendly mines fills me with profound apprehension regarding balance. As of this time I don't have the data to back anything up but for what it's worth all about me tells me this is a potential deal breaker. Right off the bat I have difficulty imagining less than arena-open maps not being sealed off cheap and fast once players realize how fast they can eliminate flanks and passages with invisible mine walls.

Just for the record I'd be more comfortable with _positive proof_ that this specific feature doesn't break balance. We might have that in the coming weeks but there's no guarantee. As discussed earlier regarding balancing, official play-tests generally functions sub-optimally as even the most dedicated ones can't be bothered with switching mods and organizing playtest games. Once the switch mods feature is implemented between releases and play-tests that might change.

Anyways as stated before I generally want to avoid feature specific inclusions that affects balance but this seem like a minor feature/performance tweak that may have more than reasonable ramifications for RA's balance.

pchote:
This sounds like a reasonable concern to me.  We can disable this for the next release by either toggling the default value for the new flag or by explicitly overriding it in yaml.  Balance test maps can then investigate the consequences of having it enabled.

Issue #12310, opened 2016-10-26
Get rid of DelayedAction effect


reaperrr:
One of the prerequisites for savegame support.

OpenRA.Game / OpenRA.Mods.Common
- [ ] WeaponInfo/Warhead

OpenRA.Mods.Common
- [ ] Demolish
- [ ] Beacon #12309
- [ ] ReinforcementsGlobal
- [ ] TriggerGlobal
- [ ] Bridge
- [ ] PlaceBuilding #12309
- [ ] WithBuildingExplosion
- [ ] WithDockedOverlay
- [ ] WithRepairOverlay
- [ ] NukePower

OpenRA.Mods.RA
- [ ] GpsWatcher #12305

OpenRA.Mods.D2k
- [ ] SpiceBloom
- [ ] WithDeliveryOverlay

OpenRA.Mods.TS
- [ ] AnimatedBeacon #12309

GraionDilach:
Is changing https://github.com/OpenRA/OpenRA/blob/634faa31deb9e03ed7b49b06c9b6ade0225bcb01/OpenRA.Mods.D2k/Activities/SwallowActor.cs#L80 related?

reaperrr:
> Is changing https://github.com/OpenRA/OpenRA/blob/634faa31deb9e03ed7b49b06c9b6ade0225bcb01/OpenRA.Mods.D2k/Activities/SwallowActor.cs#L80 related?

I'm rather sure RunAfterDelay is just as bad, possibly worse for savegames, as it ignores the tick rate.

tohava:
WeaponInfo fixed in https://github.com/OpenRA/OpenRA/pull/12347

reaperrr:
We're getting there, but lua / our scripting engine is not my strong point, so the two global script cases will probably need to be adressed by some other @OpenRA/engine-hackers.

Issue #12188, opened 2016-10-07
Add an "unstable" parameter to the [Desc] attribute.


pchote:
Our expectation for features to be complete and stable before merging them upstream makes life difficult for people who want to work on small parts of bigger features or for niche traits that we don't have test cases for in our default mods.

I believe that we can improve this situation, without hurting the stability or maintainability of our default mods, by allowing trait fields or entire traits to be marked as unstable using `[Desc("foo", unstable=true)]`.

These would be marked in the trait docs as unstable, indicating that modders should use these properties at their own risk.  We can then relax our review standards on these features, e.g. not requiring upgrade rules for incompatible changes, or allowing them to be merged with missing features or bugs if the benefits of the feature still outweigh the issues.

The main technical benefit comes from adding a `--disallow-unstable` flag to the `--check-yaml` utility command.  This would let us actively prevent unstable features from being accidentally or intentionally included in our shipping mods via the `make test` rule.

pchote:
Since somebody asked on IRC I want to clarify that I see this as a compromise to cover the cases where our current review system fails.  It is *not* a golden ticket to people to be lazy and file a half-done PR because they've lost interest or coded themselves into a corner.

reaperrr:
`WithSpriteBarrel` is a good example where this would help, it has 'lagged behind' the other sprite/voxel traits at least twice in terms of regressions due to lack of test cases.

Phrohdoh:
We are in a sore spot of C# when implementing this on top of `Desc`.

```csharp
public sealed class DescAttribute : Attribute
{
	public readonly string[] Lines;
	public DescAttribute(params string[] lines) { Lines = lines; }
}
```

`params` arguments _must_ be at the end of the argument list which means the boolean would need to be the first argument.

We cannot simply change this like so,
```csharp
public sealed class DescAttribute : Attribute
{
	public readonly bool IsUnstable;
	public readonly string[] Lines;
	public DescAttribute(bool isUnstable = false, params string[] lines) { Lines = lines; }
}
```

because a value needs to be given everywhere (the default does us no good here).

```
OpenRA.Game//Traits/World/Shroud.cs(29,9): error CS1503: Argument `#1' cannot convert `string' expression to type `bool'
```

We may be better off implementing this as a separate attribute.

```csharp
[AttributeUsage(AttributeTargets.All)]
public sealed class UnstableAttribute : Attribute
{
	public readonly string Reason;
	public readonly string GitHubTicketUrl;

[...]

PR #12338, opened 2016-11-01
Add footprint support to Explodes, remove WithBuildingExplosion

### LINKED ISSUE #4197: Building destruction should leave crater smudges
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1337907-building-destruction-should-leave-crater-smudges?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F36085&utm_medium=issues&utm_source=github).
</bountysource-plugin>

reaperrr:
Using `Explodes` instead of `WithBuildingExplosion` allows us to spawn smudges under buildings when they explode and use warhead delay instead of WBE's (now removed) `Delay`.
While weapons/warheads currently use `DelayedAction` themselves, this will be fixed at some point, making a direct fix for `WithBuildingExplosion` obsolete.

I'm not sure whether `WithBuildingExplosion` should be killed off entirely at this point, though, as it's now harmless, a bit easier to use, and keeping it would maintain backwards-compatibility. Possibly a candidate for #12188.

Closes #4197.

Contributes a little to #12310.


pchote:
I vote for killing `WithBuildingExplosion` if it has been completely superseded.

reaperrr:
Updated.

mods/ra/weapons/explosions.yaml
```diff
@@ -245,6 +245,13 @@ V2Explode:
 	Inherits: SCUD
 	-Report:
 
+BuildingExplode:
+	Warhead@1Eff: CreateEffect
+		Explosions: building, building_napalm, large_explosion, self_destruct
```
GraionDilach: Why was large_napalm removed?

GraionDilach (changes requested):
It doesn't sit well with me that TD has 5 explosions while RA has 4, while the explosion artworks are technically shared (just use different palettes and filenames).

Also needs a rebase.

mods/ra/weapons/explosions.yaml
reaperrr: Because a lot of buildings didn't list that before.
I thought that might be intentional, but didn't feel like adding a second weapon for that, so I removed it from the default because we already use the 'medium' napalm.

That being said, I don't really mind adding it back.

reaperrr:
Updated and rebased.

GraionDilach (approved):
:+1:

reaperrr:
Actually, I just noticed I need to account for civ building explosions.

reaperrr:
Updated.
- `CivBuildingExplosion` now does both the civ scatter triggering and visual explosion.
- civ buildings and defenses (except flame turret) don't use `large_napalm` when exploding.

TD didn't need any changes, because civ buildings don't spawn civilians there and building explosions include the small fireball instead of large napalm.

obrakmann (approved):
👍 Looks good to me, but needs a rebase.

reaperrr:
Rebased.

obrakmann:
[Changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/f21e4ed385778b667393091f824f080212b98538)

PR #12341, opened 2016-11-02
Remove hardcoded buildComplete delay in some overlays

reaperrr:
There certainly was and possibly still is some issue this was trying to work around, but unless it's a deal-breaker, we should rather try to fix that issue directly instead of keeping some uncommented magical work-around.

I couldn't spot any regressions in a D2k skirmish, all 3 overlays worked as intended there.

Contributes to #12310.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Render/WithDockedOverlay.cs
+++ b/OpenRA.Mods.Common/Traits/Render/WithDockedOverlay.cs
@@ -67,8 +67,7 @@ namespace OpenRA.Mods.Common.Traits.Render
 
 		void INotifyBuildComplete.BuildingComplete(Actor self)
 		{
-			self.World.AddFrameEndTask(w => w.Add(new DelayedAction(120, () =>
-					buildComplete = true)));
+			buildComplete = true;
 		}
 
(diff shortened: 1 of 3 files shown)


pchote:
None of #5119, #5373, #5376 mention why the delay was there.  I assume @Mailaender copy-pasted them from somewhere without asking too many questions.

Mailaender:
The reason was to avoid running the animation before the crumble was fallen down. I agree that it shouldn't be hardcoded. It is also possible that this was a workaround for a polish bug that looked more severe than now. After all the code is pretty old.

pchote (approved):
I couldn't reproduce any animation conflicts with the starport lights and the crumble, and the other actions aren't going to happen within the first couple of seconds of the building's life

chrisforbes (approved):

reaperrr:
[changelog](https://github.com/OpenRA/OpenRA/wiki/Changelog-(bleed)/_compare/70ab30f5af997388064a25cf42f3bb54a3c6ea19...824e529f088de10690d1282b0865308ed3f11d2a)

PR #12344, opened 2016-11-03
Make Radar Dome lose Detect Cloak ability when powered down 

### LINKED ISSUE #12272: RA - Make Radar Dome lose Stealth Detection ability if Powered Down/on Low Power
Was watching FiveAces' video and as a comment at here says, i agree that it doesn't make any sense that Stealth Detection ability of Radar still work if it is Powered Down. Also doing this would make players another reason to not Power Down the Radar.

https://www.youtube.com/watch?v=oPJvGgy-ZHs&t=1529s

Also maybe same should go for Defences.

EspressoDan:
Fixes #12272 

I gave Radar Domes an upgrade "powered" that is revoked when the building is powered down. The linter is complaining because it thinks there's nothing granting the upgrade, which is true. However, it is enabled by default and is disabled by revoking the upgrade.

### DIFF
--- a/OpenRA.Mods.Common/Traits/Power/CanPowerDown.cs
+++ b/OpenRA.Mods.Common/Traits/Power/CanPowerDown.cs
@@ -15,15 +15,17 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.Common.Traits
 {
 	[Desc("The player can disable the power individually on this actor.")]
-	public class CanPowerDownInfo : UpgradableTraitInfo, Requires<PowerInfo>
+	public class CanPowerDownInfo : UpgradableTraitInfo, Requires<PowerInfo>, Requires<UpgradeManagerInfo>
 	{
 		[Desc("Restore power when this trait is disabled.")]
 		public readonly bool CancelWhenDisabled = false;
 
 		public readonly string IndicatorImage = "poweroff";
-		[SequenceReference("IndicatorImage")] public readonly string IndicatorSequence = "offline";
+		[SequenceReference("IndicatorImage")]
+		public readonly string IndicatorSequence = "offline";
 
-		[PaletteReference] public readonly string IndicatorPalette = "chrome";
+		[PaletteReference]
+		public readonly string IndicatorPalette = "chrome";
 
 		public readonly string PowerupSound = null;
 		public readonly string PowerdownSound = null;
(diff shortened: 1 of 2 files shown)


abcdefg30:
Please use tabs instead of spaces.

OpenRA.Mods.Common/Traits/Power/CanPowerDown.cs
```diff
@@ -53,13 +55,19 @@ public void ResolveOrder(Actor self, Order order)
 			{
 				disabled = !disabled;
 
-				if (Info.PowerupSound != null && disabled)
-					Game.Sound.PlayNotification(self.World.Map.Rules, self.Owner, "Sounds", Info.PowerupSound, self.Owner.Faction.InternalName);
+			    if (Info.PowerupSound != null && disabled)
+			    {
+                    upgradeManager.RevokeUpgrade(self, "powered", this);
```
abcdefg30: This should not be hardcoded. Please expose it through a field in the info class (like we do with e.g. the `PowerupSound`).

mods/ra/rules/structures.yaml
```diff
@@ -520,6 +520,9 @@ DOME:
 	InfiltrateForExploration:
 	DetectCloaked:
 		Range: 10c0
+		UpgradeTypes: powered
+		UpgradeMaxEnabledLevel: 0
+		UpgradeMinEnabledLevel: -1
```
pchote: Upgrade types are additive, and you cannot revoke upgrades granted with one trait by another trait.  Negative upgrade levels can't happen.

OpenRA.Mods.Common/Traits/Power/CanPowerDown.cs
Phrohdoh: This will want to be a `string[]` that you iterate over and grant/revoke, too.

pchote:
You will need to make `Power` grant upgrades too, otherwise this only applies to manual powering down.

mods/ra/rules/structures.yaml
EspressoDan: Since that's the case, how do I handle DetectCloaked requiring an upgrade granted from a singular source?

pchote: `Power` and `CanPowerDown` both grant 1 level while "active" (for `CanPowerDown` this means when it *isn't* powered down), then `DetectCloaked` sets a `MinEnabledLevel` of 2.

EspressoDan:
Closing as this will require a heavy rewrite of `RequiresPower` to handle upgrade support, moving away from IDisable.

pchote:
My comment about `IDisable` was just for context about how this pr helps towards one of our longer term goals.  I wasn't asking for you to remove that here!

